PR #250, opened 2019-05-05
remove example data utils file

sofroniewn:
# Description
This PR removes the `utils/example_data_utils.py` file which is no longer being used

## Type of change

# How has this been tested?

- [x] run examples

## Final checklist:


royerloic (commented):
We could have an example that uses one of these large images...

sofroniewn:
@royerloic - we should definitely improve our example images to include more scientific ones, and the data that's hosted up there could be great, but let's revisit when we are next working on our examples

kne42 (approved):

PR #251, opened 2019-05-06
add keybindings framework 

kne42:
# Description

Add machinery for the adding, validation, and dispatch of keybindings on the viewer and its layers.

### Changes
- remove `on_key_press` and `on_key_release` methods from all layers
- on the viewer and its layers, store all initial keybinding information under the `default_keybindings` class variable, which is used to initialize the `keybindings` instance attribute
- add convenience methods for adding, removing, and rebinding keybindings

### key sequences
Key sequences are represented in the form `[modifier-]key`, e.g. `a`, `Control-c`, or `Control-Alt-Delete`. Valid modifiers are `Control`, `Alt`, `Shift`, and `Meta`, and can be represented as in Emacs with the shorthands `C`, `M`, `S`, and `s`, respectively.

Shift pressed with any letter will be consumed
to transform it to upper case, e.g. Shift + w => W.
Otherwise, the letter will be lower case.
Shift will also be consumed without transformation when pressed
with any other non-special key unless Control is also a modifier.

Special keys include Shift, Control, Alt, Meta, Up, Down, Left, Right,
PageUp, PageDown, Insert, Delete, Home, End, Escape, Backspace, F1,
F2, F3, F4, F5, F6, F7, F8, F9, F10, F11, F12, Space, Enter, and Tab

### function format
Functions take in only one argument: the parent that the function was attached to. This is the viewer or layer.

By default, all functions are assumed to work on key presses only, but can be denoted to work on release too by separating the function into two statements with the `yield` keyword:
```python
def hello_world(viewer):
    # on key press
    viewer.status = 'hello world!'

    yield

    # on key release
```

### bind a key
```python
viewer.bind_key('n', load_next_image)

@viewer.bind_key('p')
def load_prev_image(viewer): ...
```

### unbind a key
```python
viewer.unbind_keys('n')
viewer.bind_key('p', None)
```

### rebind a key
```python
viewer.bind_key('Left', 'p')
```

### bind multiple keys
```python
viewer.bind_key(('n', load_next_image),
                ('p', load_prev_image))

keybindings = [('Right', 'n'),
               ('Left', 'p')]
viewer.bind_key(keybindings)

keybindings = {
    'Control-s': save,
    'Control-u', undo
}
viewer.bind_key(keybindings)

viewer.bind_key(a=None)
```

### unbind multiple keys
```python
viewer.unbind_keys('a', 'b', 'c')
viewer.unbind_keys(['a', 'b', 'c'])
viewer.unbind_keys()
```

## Type of change

# References

# How has this been tested?

- [x] the test suite for my feature passes

## Final checklist:

### DIFF
--- a/napari/components/_viewer/model.py
+++ b/napari/components/_viewer/model.py
@@ -8,8 +8,7 @@ from ...util.event import EmitterGroup, Event
 from ...util.theme import palettes
 from ...util.misc import has_clims
 from .._dims import Dims
-from ...util.keybindings import (bind_key_method, unbind_key_method,
-                                 rebind_key_method, bind_keys_method, bind_keys)
+from ...util.keybindings import bind_key_method, unbind_keys_method, bind_keys
 
 
 class ViewerModel:
@@ -187,14 +186,8 @@ class ViewerModel:
     bind_key = bind_key_method(keybindings='_keybindings',
                                class_keybindings='default_keybindings')
 
-    unbind_key = unbind_key_method(keybindings='_keybindings',
-                                   class_keybindings='default_keybindings')
-
-    rebind_key = rebind_key_method(keybindings='_keybindings',
-                                   class_keybindings='default_keybindings')
-
-    bind_keys = bind_keys_method(keybindings='_keybindings',
-                                 class_keybindings='default_keybindings')
+    unbind_keys = unbind_keys_method(keybindings='_keybindings',
+                                     class_keybindings='default_keybindings')
 
     def reset_keybindings(self):
         self.keybindings = self.default_keybindings
--- a/napari/components/_viewer/model.py
+++ b/napari/components/_viewer/model.py
@@ -8,6 +8,8 @@ from ...util.event import EmitterGroup, Event
 from ...util.theme import palettes
 from ...util.misc import has_clims
 from .._dims import Dims
+from ...util.keybindings import (bind_key_method, unbind_key_method,
+                                 rebind_key_method, bind_keys_method, bind_keys)
 
 
 class ViewerModel:
@@ -22,13 +24,20 @@ class ViewerModel:
         List of contained layers.
     dims : Dimensions
         Contains axes, indices, dimensions and sliders.
+<<<<<<< HEAD
     key_bindings : dict of string: callable
         Custom key bindings. The dictionary key is a string containing the key
         pressed and the value is the function to be bound to the key event.
         The function should accept the viewer object as an input argument.
         These key bindings are executed instead of any layer specific key
         bindings.
+=======
+    camera : vispy.scene.Camera
+        ViewerWidget camera.
+>>>>>>> add keybinding convenience functions; convert existing keybindings
     """
+    default_keybindings = {}
+
     def __init__(self, title='napari'):
         super().__init__()
         from .._layers_list import LayersList
@@ -57,7 +66,8 @@ class ViewerModel:
         self._cursor_size = None
         self._interactive = True
         self._active_layer = None
-        self.key_bindings = {}
+        self._keybindings = {}
+        self.keybindings = self.default_keybindings
 
         # TODO: this should be eventually removed!
         # attached by QtViewer when it is constructed by the model
@@ -161,6 +171,33 @@ class ViewerModel:
             return
         self._active_layer = active_layer
         self.events.active_layer(item=self._active_layer)
+
+    @property
+    def keybindings(self):
+        return self._keybindings.copy()
+
+    @keybindings.setter
+    def keybindings(self, keybindings):
+        if keybindings == self.keybindings:
+            return
+        kb = {}
+        bind_keys(kb, keybindings.items())
+        self._keybindings = kb
+
+    bind_key = bind_key_method(keybindings='_keybindings',
+                               class_keybindings='default_keybindings')
+
+    unbind_key = unbind_key_method(keybindings='_keybindings',
+                                   class_keybindings='default_keybindings')
+
+    rebind_key = rebind_key_method(keybindings='_keybindings',
+                                   class_keybindings='default_keybindings')
+
+    bind_keys = bind_keys_method(keybindings='_keybindings',
+                                 class_keybindings='default_keybindings')
+
+    def reset_keybindings(self):
+        self.keybindings = self.default_keybindings
 
     def reset_view(self):
         """Resets the camera's view.
--- a/napari/components/_viewer/view/main.py
+++ b/napari/components/_viewer/view/main.py
@@ -1,3 +1,5 @@
+import inspect
+
 from qtpy.QtCore import QCoreApplication, Qt, QSize
 from qtpy.QtWidgets import QWidget, QSlider, QVBoxLayout, QSplitter
 from qtpy.QtGui import QCursor, QPixmap
@@ -8,6 +10,13 @@ from ..._layers_list.view import QtLayersList
 from ....resources import resources_dir
 from .controls import QtControls
 from .buttons import QtLayersButtons
+
+import os.path as osp
+from ....resources import resources_dir
+from ....util.theme import template, palettes
+from ....util.keybindings import components_to_seq
+palette = palettes['dark']
+
 
 class QtViewer(QSplitter):
 
@@ -76,6 +85,8 @@ class QtViewer(QSplitter):
                 'standard': QCursor()
             }
 
+        self._key_release_generators = {}
+
         self.viewer.events.interactive.connect(self._on_interactive)
         self.viewer.events.cursor.connect(self._on_cursor)
         self.viewer.events.reset_view.connect(self._on_reset_view)
@@ -112,6 +123,8 @@ class QtViewer(QSplitter):
     def _on_cursor(self, event):
         cursor = self.viewer.cursor
         size = self.viewer.cursor_size
+
+    def set_cursor(self, cursor, size=10):
         if cursor == 'square':
             if size < 10 or size > 300:
                 q_cursor = self._cursors['cross']
@@ -156,22 +169,37 @@ class QtViewer(QSplitter):
     def on_key_press(self, event):
         """Called whenever key pressed in canvas.
         """
-        if (event.text in self.viewer.key_bindings and not
-                event.native.isAutoRepeat()):
-            self.viewer.key_bindings[event.text](self.viewer)
+        if event.native.isAutoRepeat() or event.key is None:
             return
 
+        seq = components_to_seq(event.key.name, event.modifiers)
+
         layer = self.viewer.active_layer
-        if layer is not None:
-            layer.on_key_press(event)
+        if layer is not None and seq in layer.keybindings:
+            parent = layer
+        elif seq in self.viewer.keybindings:
+            parent = self.viewer
+        else:
+            return
+
+        func = parent.keybindings[seq]
+        gen = func(parent)
+
+        if inspect.isgeneratorfunction(func):
+            try:
+                next(gen)
+            except StopIteration:  # only one statement
+                pass
+            else:
+                self._key_release_generators[event.key] = gen
 
     def on_key_release(self, event):
         """Called whenever key released in canvas.
         """
-        layer = self.viewer.active_layer
-        if layer is not None:
-            layer.on_key_release(event)
-
+        try:
+            next(self._key_release_generators[event.key])
+        except (KeyError, StopIteration):
+            pass
 
 def viewbox_key_event(event):
     """ViewBox key event handler
--- a/napari/layers/_labels_layer/model.py
+++ b/napari/layers/_labels_layer/model.py
@@ -599,40 +599,35 @@ class Labels(Layer):
         """
         self._last_cursor_coord = None
 
-    def on_key_press(self, event):
-        """Called whenever key pressed in canvas.
+    def _activate_paint_mode(self):
+        self.mode = Mode.PAINT
 
-        Parameters
-        ----------
-        event : Event
-            Vispy event
-        """
-        if event.native.isAutoRepeat():
-            return
-        else:
-            if event.key == ' ':
-                if self.mode != Mode.PAN_ZOOM:
-                    self._mode_history = self.mode
-                    self.mode = Mode.PAN_ZOOM
-                else:
-                    self._mode_history = Mode.PAN_ZOOM
-            elif event.key == 'p':
-                self.mode = Mode.PAINT
-            elif event.key == 'f':
-                self.mode = Mode.FILL
-            elif event.key == 'z':
-                self.mode = Mode.PAN_ZOOM
-            elif event.key == 'l':
-                self.mode = Mode.PICKER
+    def _activate_fill_mode(self):
+        self.mode = Mode.FILL
 
-    def on_key_release(self, event):
-        """Called whenever key released in canvas.
+    def _activate_pan_zoom_mode(self):
+        self.mode = Mode.PAN_ZOOM
 
-        Parameters
-        ----------
-        event : Event
-            Vispy event
-        """
-        if event.key == ' ':
-            if self._mode_history != Mode.PAN_ZOOM:
-                self.mode = self._mode_history
+    def _activate_picker_mode(self):
+        self.mode = Mode.PICKER
+
+    def _hold_to_pan_zoom(self):
+        # on press
+        prev_mode = self.mode
+
+        if self.mode != Mode.PAN_ZOOM:
+            self.mode = Mode.PAN_ZOOM
+
+        yield
+
+        # on release
+        if prev_mode != Mode.PAN_ZOOM:
+            self.mode = prev_mode
+
+    default_keybindings = {
+        'Space': _hold_to_pan_zoom,
+        'p': _activate_paint_mode,
+        'f': _activate_fill_mode,
+        'z': _activate_pan_zoom_mode,
+        'l': _activate_picker_mode
+    }
--- a/napari/layers/_markers_layer/model.py
+++ b/napari/layers/_markers_layer/model.py
@@ -499,34 +499,41 @@ class Markers(Layer):
                 self._add(coord)
         self.status = self.get_message(coord, self._selected_markers)
 
-    def on_key_press(self, event):
-        """Called whenever key pressed in canvas.
-        """
-        if event.native.isAutoRepeat():
-            return
-        else:
-            if event.key == ' ':
-                if self.mode != 'pan/zoom':
-                    self._mode_history = self.mode
-                    self.mode = 'pan/zoom'
-                else:
-                    self._mode_history = 'pan/zoom'
-            elif event.key == 'Shift':
-                if self.mode == 'add':
-                    self.cursor = 'forbidden'
-            elif event.key == 'a':
-                self.mode = 'add'
-            elif event.key == 's':
-                self.mode = 'select'
-            elif event.key == 'z':
-                self.mode = 'pan/zoom'
+    def _activate_add_mode(self):
+        self.mode = 'add'
 
-    def on_key_release(self, event):
-        """Called whenever key released in canvas.
-        """
-        if event.key == ' ':
-            if self._mode_history != 'pan/zoom':
-                self.mode = self._mode_history
-        elif event.key == 'Shift':
-            if self.mode == 'add':
-                self.cursor = 'cross'
+    def _activate_select_mode(self):
+        self.mode = 'select'
+
+    def _activate_pan_zoom_mode(self):
+        self.mode = 'pan/zoom'
+
+    def _hold_to_pan_zoom(self):
+        if self.mode != 'pan/zoom':
+            # on key press
+            prev_mode = self.mode
+            self.mode = 'pan/zoom'
+
+            yield
+
+            # on key release
+            self.mode = prev_mode
+
+    def _hold_to_delete(self):
+        # on key press
+        if self.mode == 'add':
+            self.cursor = 'forbidden'
+
+        yield
+
+        # on key release
+        if self.mode == 'add':  # check again in case modes were switched
+            self.cursor = 'cross'
+
+    default_keybindings = {
+        'Space': _hold_to_pan_zoom,
+        'Shift': _hold_to_delete,
+        'a': _activate_add_mode,
+        's': _activate_select_mode,
+        'z': _activate_pan_zoom_mode
+    }
--- a/napari/layers/_shapes_layer/model.py
+++ b/napari/layers/_shapes_layer/model.py
@@ -748,46 +748,6 @@ class Shapes(Layer):
         self._node._subvisuals[1].set_data(pos=pos, color=edge_color,
                                            width=width)
 
-    def _finish_drawing(self):
-        """Reset properties used in shape drawing so new shapes can be drawn.
-        """
-        index = copy(self._moving_shape)
-        self._is_moving = False
-        self.selected_shapes = []
-        self._drag_start = None
-        self._drag_box = None
-        self._is_selecting = False
-        self._fixed_vertex = None
-        self._moving_shape = None
-        self._moving_vertex = None
-        self._hover_shape = None
-        self._hover_vertex = None
-        if self._is_creating is True and self.mode == Mode.ADD_PATH:
-            vertices = self.data._vertices[self.data._index == index]
-            if len(vertices) <= 2:
-                self.data.remove(index)
-            else:
-                self.data.edit(index, vertices[:-1])
-        if self._is_creating is True and self.mode == Mode.ADD_POLYGON:
-            vertices = self.data._vertices[self.data._index == index]
-            if len(vertices) <= 2:
-                self.data.remove(index)
-        self._is_creating = False
-        self.refresh()
-
-    def remove_selected(self):
-        """Remove any selected shapes.
-        """
-        to_remove = sorted(self.selected_shapes, reverse=True)
-        for index in to_remove:
-            self.data.remove(index)
-        self.selected_shapes = []
-        shape, vertex = self._shape_at(self.coordinates[-2:])
-        self._hover_shape = shape
-        self._hover_vertex = vertex
-        self.status = self.get_message(self.coordinates[-2:], shape, vertex)
-        self.refresh()
-
     def _rotate_box(self, angle, center=[0, 0]):
         """Perfrom a rotation on the selected box.
 
@@ -942,23 +902,6 @@ class Shapes(Layer):
         for index in self.selected_shapes:
             self.data.update_z_index(index, new_z_index)
         self.refresh()
-
-    def _copy_shapes(self):
-        """Copy selected shapes to clipboard.
-        """
-        self._clipboard = ([deepcopy(self.data.shapes[i]) for i in
-                           self._selected_shapes])
-
-    def _paste_shapes(self):
-        """Paste any shapes from clipboard and then selects them.
-        """
-        cur_shapes = len(self.data.shapes)
-        for s in self._clipboard:
-            self.data.add(s)
-        self.selected_shapes = list(range(cur_shapes,
-                                    cur_shapes+len(self._clipboard)))
-        self.move_to_front()
-        self._copy_shapes()
 
     def _move(self, coord):
         """Moves object at given mouse position and set of indices.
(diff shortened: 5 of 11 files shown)


sofroniewn:
This is looking great - I love the syntax.

Can you update the `custom_key_bindings` example (maybe rename it to `custom_keybindings` too)

Also as we discussed earlier can you add a `reset_keybindings` method that will remove any custom keybindings and reset to the default ones

napari/components/_viewer/view/main.py
```diff
@@ -71,6 +74,8 @@ def __init__(self, viewer):
                 'standard': QCursor()
             }
 
+        self._generators = {}
```
sofroniewn: can this have a more specific name like `key_generators` or something else

napari/layers/_shapes_layer/model.py
```diff
@@ -1571,86 +1514,153 @@ def on_mouse_release(self, event):
...
+        'd': _activate_direct_mode,
+        's': _activate_select_mode,
+        'z': _activate_pan_zoom_mode,
+        'i': _activate_vertex_insert_mode,
+        'x': _activate_vertex_remove_mode,
+        'Control-c': _copy_shapes,
+        'Control-v': _paste_shapes,
+        'a': _select_all,
+        'Backspace': remove_selected,
+        'Escape': _finish_drawing
```
sofroniewn, line 1595: The "Escape" doesn't seem to be working for me - i.e. select the polygon tool, start drawing and then click escape - I just keep drawing the polygon and have to quit

napari/layers/_shapes_layer/model.py
```diff
@@ -1571,86 +1514,153 @@ def on_mouse_release(self, event):
...
+        'e': _activate_add_elipse_mode,
+        'l': _activate_add_line_mode,
+        't': _activate_add_path_mode,
+        'p': _activate_add_polygon_mode,
+        'd': _activate_direct_mode,
+        's': _activate_select_mode,
+        'z': _activate_pan_zoom_mode,
+        'i': _activate_vertex_insert_mode,
+        'x': _activate_vertex_remove_mode,
+        'Control-c': _copy_shapes,
```
sofroniewn, line 1591: `Control-c` and `Control-v` don't seem to be working

napari/layers/_shapes_layer/model.py
```diff
@@ -1571,86 +1514,153 @@ def on_mouse_release(self, event):
...
+        yield
+
+        # on key release
+        self._fixed_aspect = False
+        if self._is_moving:
+            self._move(self._cursor_coord)
+
+    default_keybindings = {
+        'Space': _hold_to_pan_zoom,
+        'Shift': _hold_to_move_shape,
```
sofroniewn, line 1580: `Shift` doesn't seem to be working, i.e. select a shape and start resizing - the aspect ratio is no longer fixed

napari/layers/_shapes_layer/model.py
```diff
@@ -1571,86 +1514,153 @@ def on_mouse_release(self, event):
...
+        'p': _activate_add_polygon_mode,
+        'd': _activate_direct_mode,
+        's': _activate_select_mode,
+        'z': _activate_pan_zoom_mode,
+        'i': _activate_vertex_insert_mode,
+        'x': _activate_vertex_remove_mode,
+        'Control-c': _copy_shapes,
+        'Control-v': _paste_shapes,
+        'a': _select_all,
+        'Backspace': remove_selected,
```
sofroniewn, line 1594: Backspace / delete isn't working either

sofroniewn (changes requested):
Looking good, have requested some changes in my other comments

napari/components/_viewer/view/main.py
kne42: I've renamed it to `_key_release_generators`

kne42:
All reported bugs have been fixed. I will note that there is now a bug with Qt key handlers themselves when it comes to the modifier keys `Control` (this is actually the Command key on Mac) and `Shift` being used together on OSX.

When you press `Shift + 2` you get `@` on QWERTY keyboards, which is what will show up in most cases with Qt e.g. `Alt + Shift + 2` -> `Alt-@`. However, attempting the same thing with `Control` yields a different result: `Control + Shift + 2` -> `Control-Shift-2`.

Please test again and tell me what other bugs you discover :)

sofroniewn:
In `examples/custom_key_bindings.py` can you change `viewer.key_bindings` to `viewer.keybindings`

sofroniewn:
I just gave this a spin, all looks good to go after you fix the example. Maybe @jni wants to give it a review before merging.

The bug that you mention with control and shift is strange, I'd ignore for now though, but well found!

sofroniewn (approved):

napari/util/keybindings.py
```diff
@@ -0,0 +1,250 @@
...
+
+def _bind_keys_is_valid_pair(pair):
+    try:
+        _bind_keys_validate_pair(pair)
+    except TypeError:
+        return False
+    return True
+
+
+def bind_keys_normalize_input(args, kwargs):
```
jni: Public functions need docstrings. Come on, @kne42, you know this! =P

jni: (And it helps if private functions have docstrings also!)

napari/util/keybindings.py
```diff
@@ -0,0 +1,250 @@
...
+        def inner(func):
+            bind_key(seq, func, keybindings)
+            return func
+
+        return inner
+
+    bind_key(seq, func, keybindings)
+
+
+class bind_key_method:
```
jni: All this logic is super complicated. I know you like metaprogramming but you should add clear and extensive comments that tell the rest of us mortals what exactly this is doing. :stuck_out_tongue_winking_eye: A docstring with a tiny example would be super useful.

napari/layers/_shapes_layer/model.py
```diff
@@ -1571,86 +1514,153 @@ def on_mouse_release(self, event):
...
+        'r': _activate_add_rectangle_mode,
+        'e': _activate_add_elipse_mode,
+        'l': _activate_add_line_mode,
+        't': _activate_add_path_mode,
+        'p': _activate_add_polygon_mode,
+        'd': _activate_direct_mode,
+        's': _activate_select_mode,
+        'z': _activate_pan_zoom_mode,
+        'i': _activate_vertex_insert_mode,
+        'x': _activate_vertex_remove_mode,
```
jni, line 1590: I find it confusing that the functions are methods, but without `self` or `cls`... I also don't understand what to do if I want to set a keybinding to call a non-class function.

napari/layers/_shapes_layer/model.py
```diff
@@ -1571,86 +1514,153 @@ def on_mouse_release(self, event):
...
+        self.mode = Mode.DIRECT
+
+    def _activate_select_mode(self):
+        self.mode = Mode.SELECT
+
+    def _activate_pan_zoom_mode(self):
+        self.mode = Mode.PAN_ZOOM
+
+    def _activate_vertex_insert_mode(self):
+        self.mode = Mode.VERTEX_INSERT
```
jni, line 1472: these all feel like they should be lambdas...?

napari/layers/_markers_layer/model.py
```diff
@@ -499,34 +499,41 @@ def on_mouse_press(self, event):
...
+    def _activate_pan_zoom_mode(self):
+        self.mode = 'pan/zoom'
+
+    def _hold_to_pan_zoom(self):
+        if self.mode != 'pan/zoom':
+            # on key press
+            prev_mode = self.mode
+            self.mode = 'pan/zoom'
+
+            yield
```
jni, line 517: Super cool! =)

jni (changes requested):
@kne42 at a minimum, I think this needs a lot more documentation, including docstrings, code comments, and an example of how to change the keybindings... Sorry that I haven't had time to test yet, will do soon!

kne42:
This is just a WIP, comprehensive docs, examples, and tests as well as a cleaner implementation will all be added prior to merge. Please consult the TODO section on the PR for progress :)

napari/layers/_shapes_layer/model.py
kne42: All functions take in their parent as an argument i.e. if you're adding a keybinding to the viewer you would get passed a viewer, if you're adding a keybinding to an image layer, you'll get that image layer.

Since on the methods, `self` refers back to the parent anyways, it just works out. If you have a preexisting function, you can just wrap it with a lambda: `lambda v: my_func()`.

napari/layers/_shapes_layer/model.py
kne42: I was considering this but I think in the future we may want to have access to these functions in for example a YAML file (via some parsing):
```yaml
Shapes:
  Control-i: activate_vertex_insert_mode
```

jni:
LOL Ok, I was just thrown off by @sofroniewn's approval. =) Ping me when ready for a re-review.

napari/layers/_shapes_layer/model.py
jni: Side note: I'm not sure about separate sections in the yaml for each layer type, but in any case I would want it to include the full path to that class:

```yaml
napari.layers.Shapes:
    Ctrl-i: activate_vertex_insert_mode
```

calls `napari.layers.Shapes.activate_vertex_inside_mode`. (Note: the use of underscores should be consistent — which is a conundrum.)

I think we need to think long and hard about what callbacks accessible from the key config should get as input, *and this should always be consistent*. Is it the topmost layer? Is it the list of selected layers? Is it the Layers object itself? Is it the whole viewer?

I think it's ok for the bindings to change depending on the state of the viewer, but do we want the *inputs* to change depending on the state of the viewer? That could make it really difficult to keep everything straight as a plugin developer.

If we do define different callback inputs for different states, then we need to define what those states *can* be, and this should be a finite list.

I guess what you're going for here is that each layer can define its own shortcuts, and these shortcuts would be enabled in the case *and only in the case* where that is the only selected layer. In that case, the config file could look like:

```yaml
- base:
  - [base keybindings for all of napari]
- single-selection:
  - napari.layers.Shapes:
    - ctrl-i: path.to.my_function
- multi-selection:
  - crtl-up: napari.layers.shift_selected_layers_up
```

In that case, each top-level section could have different input arguments to the callbacks.

I really like the idea of full path specs on the functions, because then adding all kinds of crazy functionality could be added with a pip install and a .yaml file.

napari/components/_viewer/view/main.py
```diff
@@ -9,6 +11,13 @@
 from .controls import QtControls
 from .buttons import QtLayersButtons
 
+import os.path as osp
+from ....resources import resources_dir
+from ....util.theme import template, palettes
```
sofroniewn, line 16: I this `theme` stuff needed for the `keybindings` PR, or is it compensating for other refactoring, or can it be removed till later?

napari/layers/_labels_layer/model.py
```diff
@@ -601,40 +601,35 @@ def on_mouse_release(self, event):
...
+        if self.mode != Mode.PAN_ZOOM:
+            self.mode = Mode.PAN_ZOOM
+
+        yield
+
+        # on release
+        if prev_mode != Mode.PAN_ZOOM:
+            self.mode = prev_mode
+
+    default_keybindings = {
```
sofroniewn, line 627: I guess the reason for creating the dictionary instead of just using decorators here is so that we can reset to defaults later if we want, right?

napari/util/tests/test_keybindings.py
```diff
@@ -0,0 +1,230 @@
+import pytest
```
sofroniewn, line 1: Will these tests now get automatically run as part of our CI?

napari/components/_viewer/view/main.py
kne42: That must've been my mistake when rebasing, will fix.

napari/layers/_labels_layer/model.py
kne42: No, using decorators would also add it to the default keybindings anyways (as long as it is initialized as an empty dictionary). However, you cannot use the decorators in this situation because you don't have access to them – `bind_key` et al have not been defined in this scope; they are inherited from the base `Layer` class.

napari/util/tests/test_keybindings.py
kne42: Yes, any file starting with `test_` will have all `test_*` methods run automatically by `pytest`.

napari/layers/_labels_layer/model.py
sofroniewn: ah ok so the decorators are to be used by our users / plugin developers, but internally we will use the `default_keybindings` dictionary. that's fine

napari/util/tests/test_keybindings.py
sofroniewn: oh great!

napari/layers/_labels_layer/model.py
kne42: Not quite – I would say that we want to use the functional/decorator form wherever possible; it's just not possible to add these keybindings to a class within its own definition. I do know how to make it work but I'm not sure how useful it would really be.

kne42:
Proper tests and docstrings have been added and this is ready for review. Helpful comments to include would be on the clarity of the documentation, intuitiveness of the API, and whether the API is too overloaded or not.

cc @gregjohnso

napari/components/_viewer/model.py
```diff
@@ -22,13 +23,20 @@ class ViewerModel:
         List of contained layers.
     dims : Dimensions
         Contains axes, indices, dimensions and sliders.
+<<<<<<< HEAD
```
jni, line 26: Yeah, this sorta needs to go. =P

napari/components/_viewer/model.py
```diff
@@ -162,6 +171,27 @@ def active_layer(self, active_layer):
         self._active_layer = active_layer
         self.events.active_layer(item=self._active_layer)
 
+    @property
+    def keybindings(self):
+        return self._keybindings.copy()
+
+    @keybindings.setter
+    def keybindings(self, keybindings):
```
jni, line 179: What's the input here? please add a docstring.

napari/components/_viewer/model.py
```diff
@@ -162,6 +171,27 @@ def active_layer(self, active_layer):
...
+            return
+        kb = {}
+        bind_keys(kb, keybindings.items())
+        self._keybindings = kb
+
+    bind_key = bind_key_method(keybindings='_keybindings',
+                               class_keybindings='default_keybindings')
+
+    unbind_keys = unbind_keys_method(keybindings='_keybindings',
+                                     class_keybindings='default_keybindings')
```
jni, line 190: Can you add some comments to these definitions? Needless to say, this is not a standard way to define methods, so some comments explaining the rationale here would be great.

Some thoughts that may or may not be relevant:
- should `keybindings=` be renamed `instance_keybindings=`?
- could this be more cleanly implemented using multiple inheritance? (ie make a `AcceptsKeybindings` class?)

napari/components/_viewer/view/main.py
jni: This doesn't seem fixed yet. =P

Also, the dreaded quadruple-dot has reared its ugly head. Perhaps this is not the PR to fix it but just want to flag again that I think our hierarchy is too deep and complex. `components` seems redundant, actually... Isn't everything a component?

jni: (Note, please don't fix this in this PR, it'll be a mess! :joy:)

napari/util/keybindings.py
```diff
@@ -0,0 +1,582 @@
...
+Special keys include Shift, Control, Alt, Meta, Up, Down, Left, Right,
+PageUp, PageDown, Insert, Delete, Home, End, Escape, Backspace, F1,
+F2, F3, F4, F5, F6, F7, F8, F9, F10, F11, F12, Space, Enter, and Tab
+
+Functions take in only one argument: the parent that the function
+was bound to. This is the viewer or layer.
+
+By default, all functions are assumed to work on key presses only,
+but can be denoted to work on release too by separating the function
+into two statements with the yield keyword::
```
jni, line 21: I think I've said this before, but this is badass. =D

napari/util/keybindings.py
```diff
@@ -0,0 +1,582 @@
...
+
+    @viewer.bind_key('h')
+    def hello_world(viewer):
+        # on key press
+        viewer.status = 'hello world!'
+
+        yield
+
+        # on key release
+        viewer.status = 'goodbye world :('
```
jni, line 31: :joy:

napari/util/keybindings.py
```diff
@@ -0,0 +1,582 @@
...
+                keys.F11,
+                keys.F12,
+                keys.SPACE,
+                keys.ENTER,
+                keys.TAB]
+
+MODIFIER_KEYS = OrderedDict(C=keys.CONTROL,
+                            M=keys.ALT,
+                            S=keys.SHIFT,
+                            s=keys.META)
```
jni, line 86: Are these shortcuts standards somewhere?

napari/util/keybindings.py
```diff
@@ -0,0 +1,582 @@
...
+
+    Returns
+    -------
+    key : str
+        Base key of the sequence.
+    modifiers : set of str
+        Modifier keys of the sequence.
+    """
+    parsed = re.split('-(?=.+)', seq)
+    modifiers, key = parsed[:-1], parsed[-1]
```
jni, line 127:
```suggestion
    *modifiers, key = parsed
```

Python 3, baby!

```python
In [1]: *foo, bar = [1, 2, 3]
In [2]: foo
Out[2]: [1, 2]
In [3]: bar
Out[3]: 3
```

napari/util/keybindings.py
```diff
@@ -0,0 +1,582 @@
...
+
+    return key, set(modifiers)
+
+
+def components_to_seq(key, modifiers):
```
jni, line 132: a) some doctest examples would be super useful
b) as discussed in the call, let's rename sequence to combination

In some sense it's hard to distinguish between components and a key combination. My suggestion would be something like `normalize_key_combo`.

napari/util/keybindings.py
```diff
@@ -0,0 +1,582 @@
...
+            cond = lambda m: m != 'Shift'  # noqa: E731
+    elif key in SPECIAL_KEYS:
+        # remove redundant information i.e. an output of 'Shift-Shift'
+        cond = lambda m: m != key  # noqa: E731
+    else:
+        # Shift is consumed to transform key
+
+        # bug found on OSX: Command will cause Shift to not
+        # transform the key so do not consume it
+        # note: 'Control' is OSX Command key
```
jni, line 172: This note should probably be in the module-level docstring

napari/util/keybindings.py
jni: LMAO I just saw that that's the next function... Maybe I should stop using "Add single comment" =P

napari/util/keybindings.py
```diff
@@ -0,0 +1,582 @@
...
+    All aliases are converted and modifier orders are fixed to:
+    Control, Alt, Shift, Meta
+
+    Parameters
+    ----------
+    seq : str
+        Key sequence.
+
+    Shift pressed with any letter will be consumed
+    to transform it to upper case, e.g. Shift + w => W.
```
jni, line 194: Can you write a bit more about the shift-consuming choice? Is that standard? I kind of dislike it because some people might read `W` as just the key `w`...

napari/util/keybindings.py
```diff
@@ -0,0 +1,582 @@
...
+    return components_to_seq(key, modifiers)
+
+
+def _determine_binding_op(arg):
+    """Determine the binding operation based on the argument passed.
+
+    None     -> BindOperation.UNBIND
+    str      -> BindOperation.REBIND
+    callable -> BindOperation.BIND
+    """
```
jni, line 224: Can you elaborate on why this function exists? Where is it used? (I know I can go and look for usage but in general *code should be local*, ie you should not need to have global context in your head to understand the piece of code you're reading.

napari/util/keybindings.py
```diff
@@ -0,0 +1,582 @@
...
+    elif isinstance(arg, Callable):
+        return BindOperation.BIND
+    raise TypeError('expected arg to be a string, callable, or None; '
+                    f'got {type(arg)}')
+
+
+def _bind_key(keybindings, seq, func):
+    """Bind a key sequence to a keymap.
+    """
+    keybindings[seq] = func
```
jni, line 238: As mentioned in the discussion, here and elsewhere, replace `seq` with `combo`

napari/util/keybindings.py
```diff
@@ -0,0 +1,582 @@
...
+    if bind_op == BindOperation.UNBIND:
+        try:
+            _unbind_key(keybindings, seq)
+        except KeyError:
+            pass
+    elif bind_op == BindOperation.REBIND:
+        orig_seq = normalize_key_sequence(func)
+        _rebind_key(keybindings, seq, orig_seq)
+    else:
+        _bind_key(keybindings, seq, func)
```
jni, line 281: I don't love `_rebind`... Can we do without it?

I'm also not sure that `bind_key` should ever unbind, but I can see how `bind_key(dict, 'r', None)` might be an intuitive way to deal with this. BUT, not with `unbind_key` around! There should be one way to do this, I think.

napari/util/keybindings.py
```diff
@@ -0,0 +1,582 @@
...
+    for seq in seqs:
+        try:
+            funcs.append(unbind_key(keybindings, seq))
+        except KeyError:
+            if error:
+                # rebind unbound functions
+                bind_keys(keybindings, zip(seqs[:len(funcs)], funcs))
+                raise
+
+    return funcs
```
jni, line 383: This is a bit of a crazy API... I'm wondering if `unbind_keys` should not coexist with `unbind_key` — just have one function that deals with both use cases. I'd prefer if we didn't need to return `funcs` either... I can see where it might be useful, but we should try to keep our API minimal to begin with, and grow it where needed, rather than start with an enormous API.

napari/util/keybindings.py
```diff
@@ -0,0 +1,582 @@
...
+            pass
+        else:
+            args = (args,)
+
+    args += tuple(kwargs.items())
+
+    return args
+
+
+@formatdoc
```
jni, line 503: Wait what is this? I don't see the import for it...

napari/util/keybindings.py
```diff
@@ -0,0 +1,582 @@
...
+        return types.MethodType(self._method, keybindings)
+
+
+def _keybinding_method(name, method):
+    return type(name, (KeybindingDescriptor,),
+                {'_method': staticmethod(method)})
+
+
+bind_key_method = _keybinding_method('bind_key', _bind_key_method)
+unbind_keys_method = _keybinding_method('unbind_keys', unbind_keys)
```
jni, line 582: This is super funky logic... Can this not be simplified?

jni (changes requested):
@kne42 this still feels too complex for me. I don't have a good sense at all of how the framework works. The example is super great, but then there's all this complexity behind the scenes that I think will be very hard to maintain. At the very least it needs a lot more documentation about how it all fits together. But my instinct is that all this stuff can be achieved with significantly less code. I've left comments in the specific places I thought were confusing.

jni:
Also, are we leaving the yaml file for a future PR?

napari/util/keybindings.py
sofroniewn: this is tricky, most of the menu bars that I see just use capital letters and then the up arrow if the shift key is to be pressed, but i think using `w` and `W` also makes sense

napari/util/keybindings.py
sofroniewn: @jni does that mean that if you try and bind to an already bound key the effect would be additive, not to replace the existing binding? I'm not sure what is most intuitive, but best practice would probably be to let the user know somehow that there is already something there

sofroniewn:
@jni I think `yaml` can wait to a future PR.

I think having `bind_key` which allows us to use the decorators such as `@viewer.bind_key('a')` is really nice, and having `unbind_keys` exposed on the main object also makes sense to me.

I'd love to see the `example/custom_keybindings.py` get fleshed out a bit more, to include adding custom keybindings to different layer instances (i.e. `'q'` on one image layer does something different then `'q'` on another image layer), and adding to the `Markers` class, i.e. `'w'` will do the same thing on all markers layers, even ones that havn't been created yet. It would be nice to see an example of `unbind_keys` in there too to unbind a single key and multiple keys. Maybe also an example using some of the modifier keys like `Shift` or `Command`, and then maybe an example using some of the special keys like `SpaceBar`, `LeftArrow`, and `RightArrow`. For all of these cases you can just have the function do something like print the layer name / viewer title and the key that was pressed.

napari/components/_viewer/model.py
kne42: Ah yeah, I think I may go with your multiple inheritance idea.

napari/util/keybindings.py
kne42: WHAT

kne42: mind officially blown. python best lang

napari/util/keybindings.py
kne42: Emacs teehee

napari/util/keybindings.py
kne42: it is already. did i not make it clear enough?

jni: @kne42 I don't know if I'm blind but [here's](https://github.com/napari/napari/pull/251/files#diff-266e2fe0a33dbefd726fb2bf1e11055bR4) the module-level docstring and I don't see anything about Ctrl being macOS Cmd...

napari/util/keybindings.py
jni: Obviously I agree, but I would be remiss not to point out that this kind of pattern-matching was copied from functional languages such as Haskell. =)

napari/util/keybindings.py
jni: > most of the menu bars that I see just use capital letters and then the up arrow if the shift key is to be pressed

Exactly. I think expressing the shortcut with the capital is kinda non-standard and therefore bad...

sofroniewn: @freeman-lab want to weigh in here?

napari/util/keybindings.py
jni: I'm happy for `bind` to always replace, OR for it to be additive (with a warning) and having to do unbind/bind in sequence. I just think we should keep this API small. Maybe even having just `bind(..., None)` being `unbind`, then the API is really just one function.

napari/util/keybindings.py
kne42: this was more to play it in-line with how Qt handles other key combinations

for example, pressing `Shift+2` actually gets read in by Qt as `Shift+@`

jni: I think you mean `Shift-2` gets read-in as `@`?

At any rate, we should aim for our user-facing documentation to be as simple and intuitive as possible, and not have it bound to implementation details. In this case, I think writing "Shift-2" as the shortcut is way more intuitive than writing `@` as the shortcut.

kne42: no, `Shift-2` gets read in as `Shift-@`. our cleaner functions then remove the `Shift` because it is applied to a non-special key so it appears to the user as `@`

there is absolutely no way for us to know that `2` was ever pressed due to regional keyboard differences

which is why `Shift-2` as the shortcut would be invalid; also if we did want that we would have to create a table of all of the different key combos and their unicode and convert them back (which again would only work on US keyboards)

jni: Fascinating! Yes, `@` is AltGr + 2 on ISO keyboards. What a mess.

jni: @kne42 I think the "right" way to do this is to set/display the key combo we want, e.g. "shift-2", and then process the key based on the OS/system keyboard setting. This might be super hard but there might be pre-existing libraries to deal with all this stuff?

jni:
@kne42 will you get a chance to have another crack at this sometime soon?

napari/util/keybindings.py
kne42: seems way too complex to me. i think it's best that we leave this alone for now

kne42:
@jni i am currently working on a new version at https://github.com/kne42/napari/blob/better-keybindings/napari/util/keybindings.py

kne42:
closed in favor of #389

PR #253, opened 2019-05-07
theme setting

freeman-lab:
# Description
This PR makes it possible to set the theme based on a new `theme` property available on the viewer.

The API is simply to set the value of `viewer.theme` to either `light` or `dark`. This can eventually include any of the other themes we make available. Upon setting, the current color palette of the viewer will be adjusted.

settheme

I added an example under `examples/set_theme.py` that views an image and sets the theme to `light` 
rather than the default `dark`.

Implementation notes
- I was able to consolidate almost all the style handling inside `viewer` (where the `theme` property is defined).
- It was a bit ugly to handle correct styling on the clim slider for new image layers created *after* the theme has been set. To do this, I added a check to the `add_layer` method to set the clims slider colors based on the current theme *if* the new layer is an image layer. Unless we find a way to make the clims styled via `qss`, I don't see a way around this.
- I modified the range slider to have a `setColor` method and to decouple its color changes from the `enabled` / `disabled` flag.
- I fixed it so the `QtDivider` stylings come from the `qss`, making for one less place we needed to awkwardly patch the stylings.
- If an invalid theme is provided, it'll raise a KeyError with an informative error message `"Theme 'foo' not found, options are ['dark', 'light']."`

## Type of change

## References
As usual, lots of help from @sofroniewn @kne42 and @shanaxel42 !

## How has this been tested?
- [x] Added an example

## Final checklist:

### DIFF
--- a/napari/components/_layers/view/main.py
+++ b/napari/components/_layers/view/main.py
@@ -130,8 +130,8 @@ class QtDivider(QFrame):
         self.setSelected(False)
         self.setFixedSize(50, 2)
 
-    def setSelected(self, bool):
-        if bool:
+    def setSelected(self, state):
+        if state:
             self.setProperty('selected', True)
             self.style().polish(self)
         else:
--- a/napari/components/_layers/view/main.py
+++ b/napari/components/_layers/view/main.py
@@ -130,8 +130,8 @@ class QtDivider(QFrame):
         self.setSelected(False)
         self.setFixedSize(50, 2)
 
-    def setSelected(self, state):
-        if state:
+    def setSelected(self, selected):
+        if selected:
             self.setProperty('selected', True)
             self.style().polish(self)
         else:
--- a/napari/components/_layers/view/main.py
+++ b/napari/components/_layers/view/main.py
@@ -2,9 +2,6 @@ from qtpy.QtCore import Qt
 from qtpy.QtWidgets import (QWidget, QVBoxLayout, QHBoxLayout, QPushButton,
                             QFrame, QCheckBox, QScrollArea)
 
-from ....util.theme import palettes
-palette = palettes['dark']
-
 
 class QtLayers(QScrollArea):
 
@@ -128,10 +125,6 @@ class QtLayers(QScrollArea):
 
 
 class QtDivider(QFrame):
-    unselectedStlyesheet = "QFrame {;}"
-    selectedStlyesheet = """QFrame {border: 2px solid %s;
-        background-color: white; border-radius: 3px;}""" % palette['text']
-
     def __init__(self):
         super().__init__()
         self.setSelected(False)
@@ -139,6 +132,8 @@ class QtDivider(QFrame):
 
     def setSelected(self, bool):
         if bool:
-            self.setStyleSheet(self.selectedStlyesheet)
+            self.setProperty('selected', 'true')
+            self.style().polish(self)
         else:
-            self.setStyleSheet(self.unselectedStlyesheet)
+            self.setProperty('selected', 'false')
+            self.style().polish(self)
--- a/napari/components/_layers/view/main.py
+++ b/napari/components/_layers/view/main.py
@@ -135,5 +135,5 @@ class QtDivider(QFrame):
             self.setProperty('selected', 'true')
             self.style().polish(self)
         else:
-            self.setProperty('selected', 'false')
+            self.setProperty('selected', False)
             self.style().polish(self)
--- a/napari/components/_layers/view/main.py
+++ b/napari/components/_layers/view/main.py
@@ -132,7 +132,7 @@ class QtDivider(QFrame):
 
     def setSelected(self, bool):
         if bool:
-            self.setProperty('selected', 'true')
+            self.setProperty('selected', True)
             self.style().polish(self)
         else:
             self.setProperty('selected', False)
--- a/napari/util/misc.py
+++ b/napari/util/misc.py
@@ -24,6 +24,18 @@ def ensure_iterable(arg, color=False):
         return itertools.repeat(arg)
 
 
+def has_clims(arg):
+    """Check if a layer has clims.
+    """
+    if hasattr(arg, '_qt_controls'):
+        if hasattr(arg._qt_controls, 'climSlider'):
+            return True
+        else:
+            return False
+    else:
+        return False
+
+
 def is_iterable(arg, color=False):
     """Determine if a single argument is an iterable. If a color is being
     provided and the argument is a 1-D array of length 3 or 4 then the input
--- a/napari/viewer.py
+++ b/napari/viewer.py
@@ -1,5 +1,11 @@
+import os.path as osp
+
 from .components._viewer.view import QtViewer
 from .components import Window, Viewer
+from .layers._image_layer.model import Image
+from .resources import resources_dir
+from .util.theme import template, palettes
+from .util.misc import has_clims
 
 
 class ViewerApp(Viewer):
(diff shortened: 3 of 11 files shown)


napari/components/_layers/view/main.py
```diff
@@ -128,17 +125,15 @@ def dropEvent(self, event):
...
-
     def __init__(self):
         super().__init__()
         self.setSelected(False)
         self.setFixedSize(50, 2)
 
     def setSelected(self, bool):
         if bool:
-            self.setStyleSheet(self.selectedStlyesheet)
+            self.setProperty('selected', 'true')
```
kne42: would prefer if you used Python literals here:
```suggestion
            self.setProperty('selected', True)
```

napari/components/_layers/view/main.py
```diff
@@ -128,17 +125,15 @@ def dropEvent(self, event):
...
+            self.style().polish(self)
         else:
-            self.setStyleSheet(self.unselectedStlyesheet)
+            self.setProperty('selected', 'false')
```
kne42:
```suggestion
            self.setProperty('selected', False)
```

napari/util/misc.py
```diff
@@ -24,6 +24,18 @@ def ensure_iterable(arg, color=False):
         return itertools.repeat(arg)
 
 
+def has_clims(arg):
```
kne42, line 27: what's this for? seems a bit hacky to me

napari/viewer.py
```diff
@@ -38,3 +44,47 @@ def __init__(self, *images, meta=None, multichannel=None, clim_range=None,
         for name, image in named_images.items():
             self.add_image(image, meta=meta, multichannel=multichannel,
                            clim_range=clim_range, name=name)
+        self.theme = 'dark'
+
+    @property
+    def theme(self):
+        """string: Color theme.
+        """
+        if hasattr(self, '_theme'):
```
kne42: it's more pythonic to use the following:
```python
try:
    return self._theme
except AttributeError:
    return None
```

napari/viewer.py
```diff
@@ -38,3 +44,47 @@ def __init__(self, *images, meta=None, multichannel=None, clim_range=None,
...
+        palette = palettes[theme]
+
+        # template and apply the primary stylesheet
+        with open(osp.join(resources_dir, 'stylesheet.qss'), 'r') as f:
+            raw_stylesheet = f.read()
+            themed_stylesheet = template(raw_stylesheet, **palette)
+        self._qtviewer.setStyleSheet(themed_stylesheet)
+
+        # set window styles which don't use the primary stylesheet
+        self.window._status_bar.setStyleSheet("""QStatusBar { background: %s;
```
kne42: you should never use triple-quoted strings in-line

the following is a valid multi-line string:
```python
print('i just keep going '
      'and going! '
      'and going!')
```

napari/viewer.py
```diff
@@ -38,3 +44,47 @@ def __init__(self, *images, meta=None, multichannel=None, clim_range=None,
...
+            color: %s}""" % (palette['background'], palette['text']))
+        self.window._qt_center.setStyleSheet(
+            'QWidget { background: %s;}' % palette['background'])
+
+        # set styles on clim slider
+        for layer in self.layers:
+            if has_clims(layer):
```
kne42, line 86: Oh I see now why you have this. I think the fact that you have to do this means that we need to rethink this design.

kne42 (commented):
Good job! Code looks pretty clean. However, I disagree with the design.

Specifically, I think that:
- palettes should live on the viewer as a class variable called `themes` or something
- all of the logic in the current `theme` setter should live in a new `palette` setter
- the `theme` property merely wraps the `palette` property
- both of these are moved to `Viewer` itself and are given a corresponding update event
- all logic pertaining to specific components (e.g. window, sliders) should be moved elsewhere and triggered when the update event notifies them

I also think that the raw stylesheet should only be loaded once and stored as a class variable.

napari/components/_layers/view/main.py
```diff
@@ -128,17 +125,15 @@ def dropEvent(self, event):
...
-    selectedStlyesheet = """QFrame {border: 2px solid %s;
-        background-color: white; border-radius: 3px;}""" % palette['text']
-
     def __init__(self):
         super().__init__()
         self.setSelected(False)
         self.setFixedSize(50, 2)
 
     def setSelected(self, bool):
         if bool:
```
kevinyamauchi: `bool` is the built in boolean type. I think we should use a different variable name here.

sofroniewn:
Reading @kne42 suggestion on moving the `theme` / `palette` setting to the `Viewer`, I agree. I'd rather see it there and then use an event to update the `Window`.

It would also be pretty cool if we could get our custom range slider to style itself from the QSS so we didn't need to constantly special case it. Maybe @AhmetCanSolak has some ideas on how to do this?

Finally I think there's a bigger discussion about the `Viewer` / `Window` / `ViewerApp` refactor that needs to happen as I currently find the state of things there very confusing, but maybe we can have that after this PR gets merged. Does @jni have thoughts?

napari/components/_layers/view/main.py
kevinyamauchi: I think we should use a different variable name here (to avoid collision with the built-in `bool` type).

jni:
> Does @jni have thoughts?

He does! :blush: 

I suggest we rename `ViewerApp` to `View` and call the widget `ViewWidget` or `ViewerWidget`. The window I think has to stay unfortunately? @kne42 knows better than me. But then the syntax in IPython becomes:

```python
import napari
v = napari.View(image)
```

which I think is pretty nice! I'm of half a mind to ignore PEP8 and call it `view`. ;)

napari/components/_layers/view/main.py
sofroniewn: yeah that should be fixed, it isn't something that's really part of this PR, but we can make a short follow up one after

napari/viewer.py
freeman-lab: nice, fixed!

napari/components/_layers/view/main.py
freeman-lab: happy to just go ahead and fix here, @kevinyamauchi do you have a preferred name? `is_selected` `state` or... ?

freeman-lab:
thanks so much for the review! to capture an in person discussion with @kne42 and @sofroniewn , we agreed to fix the small stuff and merge this as is, and we'll then use a followup PR to refactor as @kne42 suggested (possibly alongside broader refactoring of the `Viewer`) -- completely agree that will be a better design :)

napari/components/_layers/view/main.py
freeman-lab: went with `state` but happy to change again :)

napari/viewer.py
freeman-lab: totally acknowledge this is suboptimal :) as discussed, we'll press on for now and refactor later

napari/viewer.py
freeman-lab: fixed, thanks!

napari/components/_layers/view/main.py
freeman-lab: nevermind went with `selected`

napari/viewer.py
```diff
@@ -38,3 +47,45 @@ def __init__(self, *images, meta=None, multichannel=None, clim_range=None,
...
+
+        palette = palettes[theme]
+
+        # template and apply the primary stylesheet
+        themed_stylesheet = template(self.raw_stylesheet, **palette)
+        self._qtviewer.setStyleSheet(themed_stylesheet)
+
+        # set window styles which don't use the primary stylesheet
+        self.window._status_bar.setStyleSheet('QStatusBar { background: %s;'
+            'color: %s}' % (palette['background'], palette['text']))
```
kne42: fix this pep8 error and we'll be ready to merge :)

kne42 (approved):

napari/viewer.py
freeman-lab: fixed! merge away!

PR #257, opened 2019-05-14
layer active when only one selected

sofroniewn:
# Description
This PR forces the layer controls on the left hand side to only directly modify the layer to which they belong. This will make removing the `viewer` easier from an individual `layer` as now there is no way for a layer to go in and modify many layers when those elements are changed. If multiple layers are now selected the control panel on the left disappears, even if those layers are all of the same type. This PR also removes the functionality that allowed multiple clims to be simultaneously adjusted when multiple layers were selected. In practice I have found this functionality confusing and not very helpful. If others (@royerloic) feel strongly that we should support this in some way we can discuss how best to, but the current solution of allowing one layer direct access to all the other layers runs very much against the refactoring that we were trying to achieve.

I may still need to edit / improve some doc strings, but otherwise this is good to go. 

## Type of change

# How has this been tested?

- [x] run `example/layers.py` now when multiple layers selected the control panel goes away and it is no longer possible to simultaneously adjust the clims of multiple layers

## Final checklist:

### DIFF
--- a/napari/layers/_image_layer/view/controls.py
+++ b/napari/layers/_image_layer/view/controls.py
@@ -25,14 +25,10 @@ class QtImageControls(QWidget):
         self.climSlider.rangeChanged.connect(self.clim_slider_changed)
 
     def clim_slider_changed(self, slidermin, slidermax):
-        from ..model import Image
-        for layer in self.layer.viewer.layers:
-            if isinstance(layer, Image) and layer.selected:
-                valmin, valmax = layer._clim_range
-                cmin = valmin+slidermin*(valmax-valmin)
-                cmax = valmin+slidermax*(valmax-valmin)
-                layer.clim = [cmin, cmax]
-                layer._qt_controls.clim_slider_update(None)
+        valmin, valmax = self.layer._clim_range
+        cmin = valmin+slidermin*(valmax-valmin)
+        cmax = valmin+slidermax*(valmax-valmin)
+        self.layer.clim = [cmin, cmax]
 
     def clim_slider_update(self, event):
         valmin, valmax = self.layer._clim_range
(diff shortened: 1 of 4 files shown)


napari/layers/_image_layer/view/controls.py
```diff
@@ -25,14 +25,10 @@ def __init__(self, layer):
...
-        from ..model import Image
-        for layer in self.layer.viewer.layers:
-            if isinstance(layer, Image) and layer.selected:
-                valmin, valmax = layer._clim_range
-                cmin = valmin+slidermin*(valmax-valmin)
-                cmax = valmin+slidermax*(valmax-valmin)
-                layer.clim = [cmin, cmax]
-                layer._qt_controls.clim_slider_update(None)
+        valmin, valmax = self.layer._clim_range
+        cmin = valmin+slidermin*(valmax-valmin)
```
kne42: PEP8 here and below

napari/layers/_image_layer/view/controls.py
```diff
@@ -25,14 +25,10 @@ def __init__(self, layer):
...
+        cmax = valmin+slidermax*(valmax-valmin)
+        self.layer.clim = [cmin, cmax]
```
kne42: the list brackets are probably unnecessary here

```suggestion
        self.layer.clim = cmin, cmax
```

kne42 (approved):
LGTM! Approved pending PEP8 fix.

napari/layers/_image_layer/view/controls.py
sofroniewn: fixed

napari/layers/_image_layer/view/controls.py
sofroniewn: fixed

sofroniewn:
fixed pep issues, i will wait for @royerloic or @jni to review before merging

jni (approved):
I like it!

#juansshortesteverreview

Issue #179, opened 2019-03-20
Examples should only have dependencies that are in the requirements


kevinyamauchi:
## 📚 Documentation
I think the scripts in the examples directory should be guaranteed to run immediately after a fresh installation (i.e., they should only have dependencies that are in the requirements). It's a little frustrating to do the installation, go to try out napari and then get a `ModuleNotFoundError`. What do you all think?

kne42:
please

kne42:
this is largely due to our dependency on scikit-image for images

if we get our own data module this should be easy

kevinyamauchi:
I took a look and I think you're right - we would just need to have our own data module. In any case, I ran into this while checking out the shapes PR and the wait was worth it!

jni:
I propose that we depend on scikit-image and we spend our energy improving the scikit-image data module rather than rolling our own. ;)

Another, probably better option is to add metapackages. `napari[core]` would contain only the bare requirements, `napari[full]` would contain bare requirements plus some extra convenience packages, ie stuff in `requirements/docs.txt`, such as skimage.

Issue #178, opened 2019-03-16
export to svg


sofroniewn:
## 🚀 Feature
Add support for export to svg

From @freeman-lab, mentioned in #100: 

> High-level agreement with @jni that this [napari] need not become illustrator :) with my user hat on, if we eventually have “export to svg” then I can do all fine tuning in other programs, making many features overkill to do within Napari. But without that, I may end up reproducing in Matplotlib what I already have in Napari just to fine tune the plot and/or get that SVG — not a dealbreaker but nice to avoid if possible.

freeman-lab:
this library might prove useful https://github.com/mozman/svgwrite

possible solution: programmatically assemble the SVG object using this library given the image layer contents (represented as an `Image` element) and all of the shape data (built using `Polyline`, `Polygon`, etc.), and dump the whole thing to an SVG file alongside a PNG linked to in the `Image` field

royerloic:
From Slack:
One approach is to have each layer know how to export itself as svg. The viewer can then ask each layer to produce a SVG group. Actually, we could have a SVG view for each layer type that takes a layer type and produces a SVG group. That would be a continuation of the Model-View paradigm.

PR #259, opened 2019-05-14
Viewer to svg

sofroniewn:
# Description
This PR adds full support for a to svg conversion for all of our layer types and the viewer as discussed in #178.

Each layer has the ability to generate an xml list of all the objects (i.e. shapes / images) inside it according to the svg spec. Each layer contains its own `to_svg` that can then turn that into an svg string. Additionally the viewer contains a `to_svg` that will loop through all the visible layers in order, generating the `xml_list` for each of them and combining them into a single svg string. I'm open to deleting the `to_svg` on the layers as you don't really need it if you have the option on the viewer (just make all the other layers not visible for example).

On generating the `svg` the layer opacity will be taken into account, as will the colors, thicknesses of any of the shapes, and the clims and colormaps of any of the images / labels. One thing that is not taken into account right now is the blending mode. I couldn't get the `mix-blend-mode` tag of the svg to work properly, but we can revisit that in the future. I'll also not that when illustrator saves an svg it looses the blend-mode settings too, so something maybe tricky about this in general.

Right now it is also possible to write an svg to a file by passing the `to_svg` an optional path-like object. An svg string will be returned regardless. Curious if people like this behavior. @kne42 might have thoughts.

Also to do the conversion of an image to the svg spec I first clip and scale the image according to the clim, then pass it through the colormap, then have to convert it to a png and then put the png into the svg as a base64 string. For doing the numpy -> png conversion I required the `imageio`. This would now add `imageio` as a dependency, which we could also use for loading data for our examples as this library can dynamically pull and cache these example images. This would address #179, but can be a followup discussion and PR.

One more thing is that I take the `_image_view` property for the image and labels layer as the input array. This array is already 2D and corresponds to what is being displayed by vispy. We have this array mainly so that for very large datasets we wouldn't be trying to grab a really big thing. I think this works fine for now (it wouldn't make sense to have 3D data in the svg) and when we get to image pyramids it will correspond to the displayed tile, which I think will also be fine, though we may have to apply some coordinate transforms.

Right now the markers layer will only generate circles in the svg regardless of which symbol is selected in the viewer. I think this is ok for now and we can expand support here in the future if critical. Support for the vectors layer has been added too.

Finally I've added a new example `examples/to_svg.py` so you can see the svg. Right now the line that would save the svg is commented out, though this example will still generate the svg string so it is useful for testing. It has all our layer types in there so you can see how they all work. You should be able to open the resulting file in illustrator and start editing it. Here's a screenshot of the resulting svg in illustrator (I can't paste svg directly in the PR)

image

## Type of change

# How has this been tested?

- [x] `example/to_svg.py`

## Final checklist:


kne42 (approved):
lgtm

jni:
I'm very happy with all your design decisions. Why is there some white space at the right of the image though, and can it be avoided? Perhaps the more relevant version of this question is, how do you figure out the clip coordinates?

sofroniewn:
Thanks for the catch @jni I had the width / height the wrong way around somewhere. Right now it will make the canvas go from (0,0) to the make value across any of the layers - so in this case it sticks out a little farther at the bottom because of those red lines. If the red lines had been on the top of the image (so going into negative space they actually would get clipped off, which isn't so ideal. They'd still be there if you opened up the svg in illustrator but they'd stick off the edge of the canvas. Fixing this requires improving the shape estimation stuff on each layer, which I could do in this PR or could do in a follow-up. I made the svg now use the `viewBox` so this should just work once it is implemented on the individual layers. Here's how things look now. Notice the bottom is correct, but the left-hand side is not

image

jni:
Right-oh, looks good to me now and makes sense. =) I'm going to merge, proper estimation can come later, this is already super useful!

sofroniewn:
@jni - just realized I had a git fail - hadn't pushed those changes yet :-(

sofroniewn:
I will fix tomorrow

PR #263, opened 2019-05-16
[WIP] Functional plugins infrastructure

jni:
# Description

This is a skeleton for what I envision functional plugins to look like, as a first pass. Things like live update etc I think we should leave till later.

For reference, here's what happens when I crawl the `test_validation` [module](https://github.com/napari/image-types/blob/eb2fecab25f02f0c5c3cb9b47baa8c7f1b978437/tests/test_validation.py) from image-types:

```python
In [6]: plug.crawl(tv)                                                    
Out[6]: {'gaussian_filter': <function test_validation.gaussian_filter(image: 'Image', sigma: 'Sigma') -> 'Image'>}
```

As a side bonus this PR includes a menu bar and a close-window shortcut. :tada:

This isn't even functional yet but wanted to submit to discuss during meeting.

Q: wasn't sure whether this should live in the window or the viewer, guidance here would be much appreciated. I kinda think the window.

## Type of change

# References

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/_qt/qt_main_window.py
+++ b/napari/_qt/qt_main_window.py
@@ -61,6 +61,7 @@ class Window:
 
         self._add_file_menu()
         self._add_view_menu()
+        self._add_plugins_menu()
         self._add_window_menu()
         self._add_help_menu()
 
--- a/napari/_qt/qt_main_window.py
+++ b/napari/_qt/qt_main_window.py
@@ -148,7 +148,8 @@ class Window:
         On-demand installation of remote packages is not yet implemented.
         """
         module_hierarchy = plugin_engine.crawl(module)
-        self._add_hierarchy_to_menu(module_hierarchy, self.plugins_menu)
+        menu = self.plugins_menu.addMenu('&' + module)
+        self._add_hierarchy_to_menu(module_hierarchy, menu)
         # - Create function popup/drawer windows with appropriate parameter
         #   selection
         # - Output of function linked to layer types *or* pure Python values
--- a/napari/_qt/qt_main_window.py
+++ b/napari/_qt/qt_main_window.py
@@ -22,9 +22,11 @@ from qtpy.QtWidgets import (  # noqa: E402
     QLabel,
     QAction,
     QShortcut,
+    QInputDialog,
 )
 from qtpy.QtGui import QKeySequence  # noqa: E402
 from ..util.theme import template  # noqa: E402
+from ..plugins import plugin_engine  # noqa: E402
 
 
 class Window:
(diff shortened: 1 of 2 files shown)


sofroniewn:
@jni this is really cool to see - definitely helps me understand how you're thinking about all this.

I'd say this should all live on `Window` for now, and that at some point in time we might want to pull it out of this repo entirely and into the `napari-app` repo. Having it on `Window` will make that transition easier. I don't think `Viewer` should know about plugins at all.

Looking forward to more discussion later!

jni:
@sofroniewn I've picked this up again finally, fyi. It's looking pretty cool right now. Just need to hook up actions to the functions. =) I've temporarily given up on typing and we'll just have to rely on users to give sane inputs to the functions, which is The Python Way anyway. It should be easy to add typing after an initial version is working.

The next stage is to have a popup menu for each action that lets you select either existing layers or variables in the global workspace or Python literals as input. Again, this can be replaced with a drawer with e.g. preview mode in the future.

Just keeping everyone updated... ;)

jni:
btw, I do get this warning when adding the scikit-image module: "The gui lock can only be acquired by one toolkit per session.             The lock is already acquired by qt". I presume it has something to do with adding menu items at runtime, but I have no idea... If anyone knows what it's about, pointers are appreciated!

sofroniewn:
@jni this looks so cool, great that you're picking it up again! i think this is a great time for us to make a push for the plugins :-)

when ready I'm very happy to reimplement the project @marshuang80 did this summer around image segmentation as a plugin - https://github.com/transformify-plugins/segmentify - right now it uses napari as a dependency and subclasses the Viewer.

We'll have to think about some additional things like how is a plugin allowed to define custom keybindings, and eventually how is a plugin allowed to request UI elements.

jni:
@sofroniewn indeed! To clarify, my goal for this PR is to only provide support for "functional" plugins (things that take in layer data and spit out other layer data). I feel like other stuff can come in later PRs.

I haven't thought very much about interactive plugins, but one idea is to have them be Python functions that take the viewer as input. Then the world is their oyster. =)

Something to discuss once this is ready for review. Hold please. =)

jni:
😂 I think we are going to have to curate a NumPy plugin rather than recommend that people add NumPy directly (note that the list below is in alphabetical order):

Screen Shot 2019-09-23 at 4 29 12 pm

goanpeca:
Is this something we can close now @jni? Thanks!

sofroniewn:
ToDo here is to convert this into an issue containing the code we want to preserve/ reuse in the future - cc @jni

jni:
Closing as outdated, discussion to continue in #3525.

PR #264, opened 2019-05-16
Fix svg canvas

sofroniewn:
# Description
This PR is a follow-on from #259 that includes the uncommited changes described at the end of that PR (my git mistake) and additional changes to support the canvas properly containing shapes with negative coordinates. The results are now as follows:

image

## Type of change

# References
#259, #178 

# How has this been tested?

- [x] `example/to_svg.py`

## Final checklist:

### DIFF
--- a/napari/components/_viewer/model.py
+++ b/napari/components/_viewer/model.py
@@ -243,8 +243,7 @@ class Viewer:
         xml = Element('svg', height=f'{shape[0]}', width=f'{shape[1]}',
                       version='1.1', **props)
 
-        transform = ("translate(" + str(-min_shape[1]) + " " +
-                     str(-min_shape[0]) + ")")
+        transform = f'translate({-min_shape[1]} {-min_shape[0]})'
         xml_transform = Element('g', transform=transform)
 
         for layer in self.layers:
@@ -367,8 +366,11 @@ class Viewer:
         This assumes that all layers are stacked.
         """
 
-        min_shape = [min for min, max, step in self._calc_layers_ranges()]
-        max_shape = [max for min, max, step in self._calc_layers_ranges()]
+        min_shape = []
+        max_shape = []
+        for min, max, step in self._calc_layers_ranges():
+            min_shape.append(min)
+            max_shape.append(max)
 
         return min_shape, max_shape
 
--- a/napari/components/_viewer/model.py
+++ b/napari/components/_viewer/model.py
@@ -206,7 +206,7 @@ class Viewer:
         """
         return self.canvas.render(region, size, bgcolor)
 
-    def to_svg(self, file=None, canvas_shape=None):
+    def to_svg(self, file=None, view_box=None):
         """Returns an svg string with all the currently viewed image as a png
         or writes to svg to a file.
 
@@ -217,9 +217,10 @@ class Viewer:
             either a str or bytes object representing a path, or an object
             implementing the `os.PathLike` protocol. If passed the svg will be
             written to this file
-        canvas_shape : 2-tuple, optional
-            Shape of SVG canvas to be generated. If not specified, takes the
-            shape of the last two dimensions of the view
+        view_box : 4-tuple, optional
+            View box of SVG canvas to be generated specified as `min-x`,
+            `min-y`, `width` and `height`. If not specified, calculated
+            from the last two dimensions of the view.
 
         Returns
         ----------
@@ -227,13 +228,16 @@ class Viewer:
             String with the svg specification of the currently viewed layers
         """
 
-        if canvas_shape is None:
-            canvas_shape = self._calc_max_shape()[-2:]
+        if view_box is None:
+            min_shape = self._calc_min_shape()[-2:]
+            max_shape = self._calc_max_shape()[-2:]
+            range = np.array(max_shape) - np.array(min_shape)
+            view_box = min_shape[::-1] + list(range)[::-1]
 
         props = {'xmlns': 'http://www.w3.org/2000/svg',
                  'xmlns:xlink': 'http://www.w3.org/1999/xlink'}
-        xml = Element('svg', width=f'{canvas_shape[0]}',
-                      height=f'{canvas_shape[1]}', version='1.1',
+
+        xml = Element('svg', viewBox=f'{view_box}', version='1.1',
                       **props)
 
         for layer in self.layers:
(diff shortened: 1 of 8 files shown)


napari/components/_viewer/model.py
```diff
@@ -206,7 +206,7 @@ def screenshot(self, region=None, size=None, bgcolor=None):
         """
         return self.canvas.render(region, size, bgcolor)
 
-    def to_svg(self, file=None, canvas_shape=None):
+    def to_svg(self, file=None, view_box=None):
         """Returns an svg string with all the currently viewed image as a png
```
kne42: Try to refrain from making the method summary a clone of the signature. E.g. avoid language like "return" or "takes in."

I would say something like "convert viewer state to an SVG" makes more sense here.

It's also unclear to me what you mean by "with all the currently viewed image as a png." Do you mean they're all in a single PNG (I'm guessing the SVG background)?

napari/components/_viewer/model.py
```diff
@@ -217,30 +217,38 @@ def to_svg(self, file=None, canvas_shape=None):
...
-            shape of the last two dimensions of the view
+        view_box : 4-tuple, optional
+            View box of SVG canvas to be generated specified as `min-x`,
+            `min-y`, `width` and `height`. If not specified, calculated
+            from the last two dimensions of the view.
 
         Returns
         ----------
         svg : string
             String with the svg specification of the currently viewed layers
```
kne42:
```suggestion
            SVG representation of the currently viewed layers.
```

napari/components/_viewer/model.py
```diff
@@ -217,30 +217,38 @@ def to_svg(self, file=None, canvas_shape=None):
...
         """
 
-        if canvas_shape is None:
-            canvas_shape = self._calc_max_shape()[-2:]
+        if view_box is None:
```
kne42, line 231: you define `min_shape`, `max_shape`, and `range` when `view_box` is `None` but do not define them when `view_box` is not `None`

napari/components/_viewer/model.py
```diff
@@ -217,30 +217,38 @@ def to_svg(self, file=None, canvas_shape=None):
...
+            min_shape = np.array(self._calc_min_shape()[-2:])
+            max_shape = np.array(self._calc_max_shape()[-2:])
+            range = max_shape - min_shape
```
kne42: since you're only converting the above to arrays to leverage broadcasted subtraction, i would just use `np.subtract(max_shape, min_shape)` instead

napari/components/_viewer/model.py
kne42: why overwrite the built-in `range` when that doesn't really capture what it represents here anyways? i think `shape` makes more sense

napari/components/_viewer/model.py
```diff
@@ -353,12 +361,18 @@ def _calc_layers_ranges(self):
...
-        this method should not be used but instead '_calc_layers_ranges' should
-        be called.
         """
 
-        max_shape = [max-min for min, max, step in self._calc_layers_ranges()]
+        max_shape = [max for min, max, step in self._calc_layers_ranges()]
+
+        return max_shape
+
+    def _calc_min_shape(self):
```
kne42: i think that min_shape and max_shape are poor names for what is going on here. from what i can tell, `min_shape` and `max_shape` really just describe the top-left and bottom-right corners of the viewer's bounding box

i also don't think that they should be in separate methods since they're related and calculated from the same method (`_calc_layers_ranges`)

i would just rename this to something like `_calc_bbox` and return a 2-tuple of 2-tuples equivalent to what is now `(min_shape, max_shape)`

napari/components/_viewer/model.py
```diff
@@ -353,12 +361,18 @@ def _calc_layers_ranges(self):
...
+        """Calculates the min shape of all displayed layers.
+        This assumes that all layers are stacked.
+        """
+
+        max_shape = [min for min, max, step in self._calc_layers_ranges()]
```
kne42: this should be `min_shape`

napari/layers/_base_layer/model.py
```diff
@@ -350,26 +349,36 @@ def to_svg(self, file=None, canvas_shape=None):
...
+        view_box : 4-tuple, optional
+            View box of SVG canvas to be generated specified as `min-x`,
+            `min-y`, `width` and `height`. If not specified, calculated
+            from the last two dimensions of the layer.
 
         Returns
         ----------
         svg : string
-            String with the svg specification of the currently viewed image
+            String with the svg specification of the currently viewed layers
```
kne42: Fix this and summary.

napari/layers/_base_layer/model.py
```diff
@@ -350,26 +349,36 @@ def to_svg(self, file=None, canvas_shape=None):
...
         """
 
-        if canvas_shape is None:
-            canvas_shape = self.shape[-2:]
+        if view_box is None:
```
kne42, line 363: see above

napari/layers/_base_layer/model.py
```diff
@@ -350,26 +349,36 @@ def to_svg(self, file=None, canvas_shape=None):
...
+            min_shape = np.array([r[0] for r in self.range[-2:]])
+            max_shape = np.array([r[1] for f in self.range[-2:]])
+            range = max_shape - min_shape
```
kne42: see above

kne42 (changes requested):
Good work! Left a lot of comments on design choices and some things I may have overlooked on the previous PR.

napari/components/_viewer/model.py
sofroniewn: fixed!

napari/components/_viewer/model.py
sofroniewn: fixed

napari/components/_viewer/model.py
sofroniewn: fixed

napari/components/_viewer/model.py
sofroniewn: fixed

napari/components/_viewer/model.py
sofroniewn: fixed

napari/components/_viewer/model.py
sofroniewn: no longer needed

napari/layers/_base_layer/model.py
sofroniewn: fixed

napari/layers/_base_layer/model.py
sofroniewn: fixed

napari/layers/_base_layer/model.py
sofroniewn: fixed

kne42 (approved):

PR #266, opened 2019-05-16
replace skimage in examples

sofroniewn:
# Description
This PR partially addresses #179 which proposes that examples should only depend on things that `napari` actually depends on. As we now depend on `imageio` for conversion to png in #259 we can now use imageio for loading example images. imageio doesn't download those images automatically, but only when they are requested. They are cached so that this only needs to happen once. [See here for more info and a list of standard images](https://imageio.readthedocs.io/en/stable/standardimages.html).

There are still examples that depend on `skimage` for computation - like `labels-0-2d.py`, and `labels-2d.py`. Also we are still using `skimage.data.binary_blobs` in a couple places like `nD_image.py`. Finally there are our `dask_nD_image.py` and `zarr_nD_image.py` examples. Curious what we want to do about all these? For now I think it's fine to just fix the ones that this PR fixes and we can revisit more in the future.

## Type of change

# References
#179 

# How has this been tested?

- [x] run examples

## Final checklist:

### DIFF
--- a/examples/add_image.py
+++ b/examples/add_image.py
@@ -3,16 +3,16 @@ Displays one image using the add_image API and then adjust some of its
 properties
 """
 
-from skimage import data
-from skimage.color import rgb2gray
+from imageio import imread
 from napari import ViewerApp
 from napari.util import app_context
 
 
+astro = imread('imageio:astronaut.png').mean(axis=2) / 255
+
 with app_context():
     # create the viewer with an image
-    viewer = ViewerApp(astronaut=rgb2gray(data.astronaut()),
-                       title='napari example')
+    viewer = ViewerApp(astronaut=astro, title='napari example')
 
     # adjust some of the layer properties
     layer = viewer.layers[0]
--- a/examples/add_markers.py
+++ b/examples/add_markers.py
@@ -4,18 +4,19 @@ add_image APIs
 """
 
 import numpy as np
-from skimage import data
-from skimage.color import rgb2gray
+from imageio import imread
 from napari import ViewerApp
 from napari.util import app_context
 
+
+astro = imread('imageio:astronaut.png').mean(axis=2) / 255
 
 with app_context():
     # create the viewer and window
     viewer = ViewerApp()
 
     # add the image
-    viewer.add_image(rgb2gray(data.astronaut()))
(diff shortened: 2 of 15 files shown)


examples/add_image.py
```diff
@@ -3,16 +3,16 @@
...
 """
 
-from skimage import data
-from skimage.color import rgb2gray
+from imageio import imread
 from napari import ViewerApp
 from napari.util import app_context
 
 
+astro = imread('imageio:astronaut.png').mean(axis=2) / 255
```
jni, line 11: Oh god, I take back the support I expressed in the meeting, this is just terrible LOL. I think this refactor should include some test data. We can use [pooch](https://www.leouieda.com/blog/introducing-pooch.html) to use some fluorescence or EM data, and avoid this kind of "unnatural" loading.

sofroniewn: @jni I don't have super strong opinions on this one, but can you explain what is "unnatural" about this loading and how pooch is different (from a loading standpoint). Is it the dynamic downloading / caching of the data? I feel like pooch does that too.

I get that pooch would allow us to use our own more scientifically interesting datasets - and maybe that would be sufficient justification for not doing this refactor, but I sense you have other objections!

jni: > can you explain what is "unnatural" about this loading and how pooch is different (from a loading standpoint)

With pooch you can use a different process/function (we can do this explicitly or with some magic — a file that would always be run when the examples are run) and then get *an actual filename with a real path in the filesystem*, which would match how people would normally load a file. This `imageio:` gives me the heebie-jeebies. =P

> pooch would allow us to use our own more scientifically interesting datasets

yeah that's another advantage. We could use EM data for grayscale examples and fluorescence data for multichannel ones.

sofroniewn: Ah so it's really about `imageio:astronaut.png` vs some more conventional path like `examples/astronaut.png`. I guess I don't have quite the aversion to the first one, but don't see this as so critical either. Curious how others feel. @kne42 @kevinyamauchi

sofroniewn:
I am going to close this PR as it seems like something like `pooch` is the preferred solution, but we will need to address this at some point

Issue #243, opened 2019-05-02
clim slider disappears when image layer is renamed


d-v-b:
## 🐛 Bug

When I rename an image layer, the clim slider disappears.

## To Reproduce

run `examples/nD_image.py` and rename the image layer to "Tane", or a name of your choice.

## Expected behavior

clim slider shouldn't have layer-name-dependent behavior

## Environment

 - napari : 0.0.6+33.g7b8750f
 - OS: OSX 10.12.6:
 - Python version: 3.7.3

d-v-b:
Update! there actually is no layer-name-dependent behavior. _Deselecting the image layer_ made the clim slider disappear (probably as it should). After renaming the layer, I clicked on the image layer itself (to escape the focus of the layer name textbox), and this also deselected the layer, which made me _think_ the layer rename was making the clim slider disappear. I'm not sure what kind of issue this is -- maybe escaping focus of the layer name textbox shouldn't also deselect the layer?

sofroniewn:
hmm, we've just updated some of our stylings. Can you try how it feels on the latest (v0.0.8 from pip is fine). I'm curious if it is still confusing / deselecting inappropriately - I think probably it is, but worth checking. Also you might like the new stylings! (Though note that some of the dims ordering has also changed if you have 3D data)

d-v-b:
Yes, this phenomenon still occurs with v0.0.8. I still think it's still a bit confusing -- my instinct when clicking outside of a textbox to deselect it is that _only_ that textbox will be deselected. Current behavior is to deselect the textbox and the entire layer with the same action. 

(the new stylings look good, and I'm much happier with the new dims ordering!)

sofroniewn:
yup, thanks @d-v-b we'll fix this

sofroniewn:
@d-v-b this should now be fixed on master by #271. Can you give it a try and let me know if you like it

d-v-b:
I'm still seeing that same issue on the latest version

sofroniewn:
Just discussed this with @d-v-b - the functionality is now such that clicking within the layer properties when editing text will deselect the text and leave the layer selected. However, clicking elsewhere in the layers panel will both deselect the text and unselect the layer. I don't find this particularly unintuitive, but a majority of users might disagree and think that only the text should be deselected and the layer should remain selected.

We plan to do more through a more thorough review of our interactivity for our `0.2` release, and as there are now plans to work on this further right now I will close this issue.

Issue #94, opened 2019-01-20
Fix circular referencing


sofroniewn:
## ❓ Questions and Help

Right now we have a lot of circular referencing of our main objects, i.e. you can type `viewer.control_bars.viewer` or `viewer.layers.viewer`. This was raised at our last meeting and was seen as undesirable as right now you can basically get anywhere from anywhere and read / write as you wish. Unfortunately it was not clear how to fix it. Note that this question doesn't directly connect to any of the qt / UI / non-UI stuff we were talking about, but is really just about how different parts of our "model" or "state" reference each other.

I think we'd like to move to a more tree / DAG like structure, but a problem then becomes - how do you share information across different parts of the graph. Maybe our struggles indicate we have not found the right abstractions and modularity.

As a concrete example I'd like to focus on the `ControlBars` class (but similar arguments would apply to `Dimensions` and `LayersList` too). On construction `viewer` gets passed to the object and appended as a property. When the clims slider values change we go inside `viewer` and inside `layers` to find the layer whose clim values need updating. Similarly when the layer selection changes we go inside `viewer` and inside `layers` to find the layer whose clim values now need to be sent to the slider. Notice how the communication is bidirectional.

I could imagine a world where we passed the `viewer` to `ControlBars` on construction, connected up some events but then did not append it as a property. This might make it more clear how different parts of our model interact with each other -  it's just that I can't really see how to make this work (without say also requiring that the LayersList object digs into the `ControlBars` object too). It could mean that we need to pull more of this functionality that involved both the `LayersList` and the `ControlBars` back into the viewer which can see both - but then maybe some of the modularity and motivation for having the separation is lost too.

Curious about any thoughts?
@kne42, @AhmetCanSolak, @jni, @royerloic

sofroniewn:
I feel like for read events it's ok - just have the `viewer` pass the `ControlBars` the information it needs when it needs it (could be all of the `LayersList` could just be some computed values).

The trickier bits for me to imagine are the write events - i.e. something is happening inside `ControlBars` and the `viewer` or `LayersList` needs to be updated. I see two potential solutions - either these objects are there as properties which can be directly written on, or have update methods directly called - OR - the `ControlBars` emits update events which the `viewer` and `LayersList` now need to listen too. Is this second option what we want? Will it make things cleaner / more modular or just more obfuscated and confusing?

PR #271, opened 2019-05-18
refactor draggable layers

sofroniewn:
# Description
This PR moves the draggable layers functionality from the individual `QtLayer` object to the `QtLayers` object. This means that the layer objects themselves no longer need to introspect into the layers list to edit other layers - i.e. calls like `layer.viewer.layers[ind]` are now gone. This will help remove our circular referencing as desired in #94. Now the individual `QtLayer` just passes through its mouse events (using `event.ignore`) to its parent the `QtLayers` which has access to everything it needs to do the rearrangement. Thanks to @shanaxel42 for the tips on how to do this.

As an added benefit - this PR should now fix the layer selection bug in #243. 

## Type of change

# References
#94, #243
 
# How has this been tested?

- [x] run `example/layers.py` and rearrange layers

## Final checklist:

### DIFF
--- a/napari/components/_layers/view/main.py
+++ b/napari/components/_layers/view/main.py
@@ -92,6 +92,8 @@ class QtLayers(QScrollArea):
         modifiers = event.modifiers()
         layer = self.layers[self.drag_name]
         if modifiers == Qt.ShiftModifier:
+            # If shift select all layers in between currently selected one and
+            # clicked one
             index = self.layers.index(layer)
             lastSelected = None
             for i in range(len(self.layers)):
@@ -102,8 +104,10 @@ class QtLayers(QScrollArea):
             for i in range(r[0], r[1] + 1):
                 self.layers[i].selected = True
         elif modifiers == Qt.ControlModifier:
+            # If control click toggle selected state
             layer.selected = not layer.selected
         else:
+            # If otherwise unselect all and leave clicked one selected
             self.layers.unselect_all(ignore=layer)
             layer.selected = True
 
--- a/napari/components/_layers/view/main.py
+++ b/napari/components/_layers/view/main.py
@@ -108,7 +108,9 @@ class QtLayers(QScrollArea):
 
     def mouseMoveEvent(self, event):
         widget = self.childAt(event.pos())
-        if hasattr(widget, 'layer'):
+        if widget is None:
+            return
+        elif hasattr(widget, 'layer'):
             layer = widget.layer
         elif hasattr(widget.parentWidget(), 'layer'):
             layer = widget.parentWidget().layer
@@ -123,14 +125,7 @@ class QtLayers(QScrollArea):
         drag = QDrag(self)
         drag.setMimeData(mimeData)
         drag.setHotSpot(event.pos() - self.rect().topLeft())
-        dropAction = drag.exec_(Qt.MoveAction | Qt.CopyAction)
-
-        if dropAction == Qt.CopyAction:
-            if not layer.selected:
-                index = self.layers.index(layer)
-                self.layers.pop(index)
-            else:
-                self.layers.remove_selected()
+        dropAction = drag.exec_()
         event.accept()
 
     def dragLeaveEvent(self, event):
--- a/napari/components/_layers/view/main.py
+++ b/napari/components/_layers/view/main.py
@@ -1,6 +1,7 @@
-from qtpy.QtCore import Qt
+from qtpy.QtCore import Qt, QMimeData
 from qtpy.QtWidgets import (QWidget, QVBoxLayout, QHBoxLayout, QPushButton,
-                            QFrame, QCheckBox, QScrollArea)
+                            QFrame, QCheckBox, QScrollArea, QApplication)
+from qtpy.QtGui import QDrag
 
 
 class QtLayers(QScrollArea):
@@ -24,6 +25,8 @@ class QtLayers(QScrollArea):
         self.layers.events.added.connect(self._add)
         self.layers.events.removed.connect(self._remove)
         self.layers.events.reordered.connect(self._reorder)
+
+        self.dragStartPosition = (0, 0)
 
     def _add(self, event):
         """Insert `event.widget` at index `event.index`."""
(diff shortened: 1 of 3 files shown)


sofroniewn:
I've also fixed the drag and drop on the trash icon to delete functionality which was broken

napari/components/_layers/view/main.py
```diff
@@ -65,9 +68,65 @@ def _reorder(self, event):
                                               layer._qt_properties)
                 self.vbox_layout.insertWidget(2*(total - i), divider)
 
+    def mousePressEvent(self, event):
+        widget = self.childAt(event.pos())
+        if hasattr(widget, 'layer') or hasattr(widget.parentWidget(), 'layer'):
```
kne42: when would the widget not have a `layer` attribute? what widgets have it?

napari/components/_layers/view/main.py
```diff
@@ -65,9 +68,65 @@ def _reorder(self, event):
...
+            self.dragStartPosition = event.pos()
+        event.accept()
+
     def mouseReleaseEvent(self, event):
         """Unselects all layer widgets."""
-        self.layers.unselect_all()
+
+        widget = self.childAt(event.pos())
+        if hasattr(widget, 'layer'):
```
kne42: this clause can be replaced by
```python
layer = getattr(widget, 'layer', None) or getattr(widget.parentWidget(), 'layer', None)
```
although it's potentially less clear in the new form so it's really up to you

napari/components/_layers/view/main.py
```diff
@@ -65,9 +68,65 @@ def _reorder(self, event):
...
+        elif hasattr(widget.parentWidget(), 'layer'):
+            layer = widget.parentWidget().layer
+        else:
+            layer = None
+
+        if layer is None:
+            self.layers.unselect_all()
+        else:
+            modifiers = event.modifiers()
+            if modifiers == Qt.ShiftModifier:
```
kne42: you do not need nested `if`-statements here – I would recommend moving `modifier = ...` outside and making this one giant `if-elif-...-else` statement

napari/components/_layers/view/main.py
```diff
@@ -65,9 +68,65 @@ def _reorder(self, event):
...
+                index = self.layers.index(layer)
+                lastSelected = None
+                for i in range(len(self.layers)):
+                    if self.layers[i].selected:
+                        lastSelected = i
+                r = [index, lastSelected]
+                r.sort()
+                for i in range(r[0], r[1]+1):
```
kne42: pep8

napari/components/_layers/view/main.py
kne42: could you also add comments as to what each modifier (or lack thereof) does?

napari/components/_layers/view/main.py
```diff
@@ -76,12 +135,12 @@ def dragLeaveEvent(self, event):
             self.vbox_layout.itemAt(i).widget().setSelected(False)
 
     def dragEnterEvent(self, event):
-        event.accept()
         divs = []
         for i in range(0, self.vbox_layout.count(), 2):
             widget = self.vbox_layout.itemAt(i).widget()
             divs.append(widget.y()+widget.frameGeometry().height()/2)
         self.centers = [(divs[i+1]+divs[i])/2 for i in range(len(divs)-1)]
+        event.accept()
```
kne42, line 137: just out of curiosity, does this change have any actual effect on the program?

napari/components/_layers/view/main.py
```diff
@@ -92,10 +151,11 @@ def dragMoveEvent(self, event):
         center_list = (i for i, x in enumerate(self.centers) if x > cord)
         divider_index = next(center_list, len(self.centers))
         # Determine the current location of the widget being dragged
-        layerWidget = event.source()
         total = self.vbox_layout.count()//2 - 1
-        index = total - self.vbox_layout.indexOf(layerWidget)//2 - 1
         insert = total - divider_index
+        layer_name = event.mimeData().text()
+        layer = self.layers[layer_name]
```
kne42: if you don't need `layer` you can just get the index directly from the layer name

napari/components/_layers/view/main.py
```diff
@@ -112,10 +172,11 @@ def dropEvent(self, event):
         cord = event.pos().y()
         center_list = (i for i, x in enumerate(self.centers) if x > cord)
         divider_index = next(center_list, len(self.centers))
-        layerWidget = event.source()
         total = self.vbox_layout.count()//2 - 1
-        index = total - self.vbox_layout.indexOf(layerWidget)//2 - 1
         insert = total - divider_index
+        layer_name = event.mimeData().text()
+        layer = self.layers[layer_name]
```
kne42: see above

napari/components/_viewer/view/buttons.py
```diff
@@ -48,8 +48,14 @@ def dragLeaveEvent(self, event):
         self.update()
 
     def dropEvent(self, event):
-        event.setDropAction(Qt.CopyAction)
         event.accept()
+        layer_name = event.mimeData().text()
+        layer = self.viewer.layers[layer_name]
+        if not layer.selected:
+            index = self.viewer.layers.index(layer)
+            self.viewer.layers.pop(index)
```
kne42: you can just `self.viewer.layers.remove(layer)`

kne42 (approved):
LGTM! Left some comments on design choices that you may or may not address.

napari/components/_layers/view/main.py
sofroniewn: added comments

napari/components/_layers/view/main.py
sofroniewn: i've reduced the number of places this logic occurred and used this version instead

napari/components/_layers/view/main.py
sofroniewn: fixed

napari/components/_layers/view/main.py
sofroniewn: I have removed the unnecessary `event.accept` that was in the mouse move / press / release events, but this `event.accept` which is inside the drag enter event is necessary to allow the drag to register. Without it the dividers between the layers don't change color

napari/components/_viewer/view/buttons.py
sofroniewn: fixed

napari/components/_layers/view/main.py
sofroniewn: fixed

napari/components/_layers/view/main.py
sofroniewn: fixed

napari/components/_layers/view/main.py
sofroniewn: fixed, added comments

sofroniewn:
Went through and made changes - thanks again @kne42 for another amazing PR review! Will merge once the tests are done

PR #274, opened 2019-05-19
add layer viewer update events

sofroniewn:
# Description
This short PR switches the layer to viewer updates to using events rather than a layer directly setting properties on the viewer. This steps us further in the direction desired in the refactor described in #94 by making it easier in the future to remove the `viewer` property from the `layer` entirely and remove our circular dependencies.

## Type of change

# References
#94

# How has this been tested?

- [x] run examples

## Final checklist:

### DIFF
--- a/napari/layers/_base_layer/model.py
+++ b/napari/layers/_base_layer/model.py
@@ -64,7 +64,12 @@ class Layer(VisualWrapper, ABC):
         self._cursor_position = (0, 0)
         self.events.add(select=Event,
                         deselect=Event,
-                        name=Event)
+                        name=Event,
+                        status=Event,
+                        help=Event,
+                        interactive=Event,
+                        cursor=Event,
+                        cursor_size=Event)
         self.name = name
 
     def __str__(self):
@@ -215,7 +220,7 @@ class Layer(VisualWrapper, ABC):
     def status(self, status):
         if status == self.status:
             return
-        self.viewer.status = status
+        self.events.status(status=status)
         self._status = status
 
     @property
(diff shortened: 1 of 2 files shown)


sofroniewn:
I have now also moved the name coercing onto the viewer - which can check all the layers to see if a duplicate name has been used and do the incrementing

sofroniewn:
I have also moved the setting of the vispy node parent to the viewer, which will mean one less reason the viewer needs to be on the layer

napari/layers/_base_layer/model.py
```diff
@@ -281,17 +280,6 @@ def scale_factor(self):
...
-    def _after_set_viewer(self, prev):
-        """Triggered after a new viewer is set.
-
-        Parameters
-        ----------
-        prev : Viewer
-            Previous viewer.
-        """
-        if self.viewer is not None:
-            self.refresh()
```
kne42, line 293: I don't think that this will break anything now but it might be good to keep track of in case anyone wants to juggle layers across multiple viewers!

kne42 (approved):

napari/layers/_base_layer/model.py
sofroniewn: In a soon to come PR the `viewer` property will be removed entirely from the `Layer` so as to prevent circular referencing. I'm not quite sure at that point how moving layers from one viewer to another would work. `viewer.add_layer(layer)` seems like find syntax, but we'd have to make sure that worked. I'm not sure it that is supported now

Issue #193, opened 2019-03-31
Make `_get_coord` a method on the `base_layer`


sofroniewn:
Right now each layer has to implement the conversion from canvas coordinates to image coordinates itself. This is something that should be handled by the base_layer

sofroniewn:
This issue is closed with #249

PR #115, opened 2019-02-19
Dims refactor

sofroniewn:
# Description

This is a pretty simple PR that removes the need to append the `dimensions` object to its `qt` counterpart. As we try and refactor into more of an MVC like framework these changes move us in the right direction. The PR also renames the `dimensions` to `dims` for ease of access.

## Type of change

- [x] Bug-fix (non-breaking change which fixes an issue)

# How has this been tested?

- [x] test using `examples\layers.ipynb`

## Final checklist:
- [x] My PR is the minimum possible work for the desired functionality

PR #275, opened 2019-05-19
flip markers

sofroniewn:
# Description
This PR fixes a bug in the markers layer introduced during either the dims refactor #115 or moving the get_coords and indices #193, where the interactivity for the markers layer was flipped relative to expected - i.e. clicking on the point (x, y) put a marker at (y, x). Also a marker at (300, 100) will now display at row 300 column 100, which is what conforms to our other dims ordering. I had to fix the svg generation too. Can someone please check this is all consistent for the markers. It probably isn't a bad time to double check all our layer types for this too.

## Type of change

# How has this been tested?

- [x] `examples/add_markers.py`

## Final checklist:

### DIFF
--- a/napari/layers/_markers_layer/model.py
+++ b/napari/layers/_markers_layer/model.py
@@ -419,7 +419,7 @@ class Markers(Layer):
         self._sizes_view = sizes
 
         self._node.set_data(
-            data, size=sizes, edge_width=self.edge_width,
+            data[:, [1, 0]], size=sizes, edge_width=self.edge_width,
             symbol=self.symbol, edge_width_rel=self.edge_width_rel,
             edge_color=self.edge_color, face_color=self.face_color,
             scaling=self.scaling)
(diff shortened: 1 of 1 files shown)


napari/layers/_markers_layer/model.py
```diff
@@ -419,7 +419,7 @@ def _set_view_slice(self):
         self._sizes_view = sizes
 
         self._node.set_data(
-            data, size=sizes, edge_width=self.edge_width,
+            data[:, [1, 0]], size=sizes, edge_width=self.edge_width,
```
jni, line 422: My only complaint here is that this breaks my idea that coordinate transforms should happen in one place and one place only. Do you think that principle is workable in this PR?

sofroniewn: I fear this is unavoidable as vispy represents the order differently from how we do it - there will always need to be a flip in there somewhere. Soon we might be able to move the vispy `node` stuff to a separate view file, but we aren't there yet!

jni: I'm aware that there will always need to be a flip, but that flip should *always* look the same — a call to `data2vispy` or `vispy2data` or similar. Happy to postpone to a later PR, just wanted to flag it again so it stays near the top of our consciousness. Will merge now as I confirmed that it works correctly for me on all annotate/marker examples.

jni (approved):
Works here.

Issue #261, opened 2019-05-14
Drop support for edge_width_rel and scaling in markers


sofroniewn:
## 🐛 Bug

I think we should drop support for exposing the `edge_width_rel` property and set the scaling property to always be `True`.

Only one of `edge_width_rel` and `edge_width` can ever be used and the rest of our mesh like layers only use `edge_width`. When making the `to_svg` conversion only `edge_width` is supported too. I think the api is cleaner and simpler without it.

Similarly in all our other mesh-like layers the equivalent of the `scaling` property is `True`. We should make sure though that the size of the markers is in units of the image and not of the screen, again similar to our other mesh-like layers.

@kevinyamauchi what is your opinion on this?

kevinyamauchi:
I think this makes sense to me!

One additional thing we can sneak into this refactor: I think we should change how we set individual marker sizes. I propose that we keep the `marker_size` property, but change the size slider in the layer properties to modulate a `marker_scale_factor` layer property. Thus when rendered, each marker has size `marker_size * marker_scale_factor` and when the size slider is used, relative scaling is preserved.

Thoughts?

sofroniewn:
@kevinyamauchi relative scaling makes sense, happy to explore

sofroniewn:
Closed by #276

PR #276, opened 2019-05-19
fix markers sizing

sofroniewn:
# Description
This PR addresses #261 by dropping support for relative edge widths and non-scaling markers. This moves the markers layer more in line with the shapes layers. It also makes the markers size slider preserve the relative size differences between markers as suggested by @kevinyamauchi

## Type of change

# References
#261

# How has this been tested?

- [x] run `examples/add_markers.py`. Note that adding markers is currently flipped, but #275 fixes that.

## Final checklist:

### DIFF
--- a/napari/layers/_markers_layer/view/properties.py
+++ b/napari/layers/_markers_layer/view/properties.py
@@ -93,6 +93,7 @@ class QtMarkersLayer(QtLayer):
         self.layer.symbol = text
 
     def changeSize(self, value):
+        """Rescale marker sizes according to the value of the size slider."""
         avg = np.mean(self.layer.size) or 1
         self.layer.size = self.layer.size / avg * value
 
--- a/napari/layers/_markers_layer/view/properties.py
+++ b/napari/layers/_markers_layer/view/properties.py
@@ -20,7 +20,7 @@ class QtMarkersLayer(QtLayer):
         sld = QSlider(Qt.Horizontal, self)
         sld.setFocusPolicy(Qt.NoFocus)
         sld.setFixedWidth(110)
-        sld.setMinimum(0)
+        sld.setMinimum(1)
         sld.setMaximum(100)
         sld.setSingleStep(1)
         value = self.layer.size
@@ -93,7 +93,10 @@ class QtMarkersLayer(QtLayer):
         self.layer.symbol = text
 
     def changeSize(self, value):
-        self.layer.size = value
+        avg = np.asarray(self.layer.size).mean()
+        if avg == 0:
+            avg = 1
+        self.layer.size = self.layer.size/avg*value
 
     def change_ndim(self, state):
         if state == Qt.Checked:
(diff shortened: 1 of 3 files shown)


napari/layers/_markers_layer/view/properties.py
```diff
@@ -20,7 +20,7 @@ def __init__(self, layer):
         sld = QSlider(Qt.Horizontal, self)
         sld.setFocusPolicy(Qt.NoFocus)
         sld.setFixedWidth(110)
```
kevinyamauchi, line 22: An issue for a separate PR, but I think constants like these should be defined in some sort of config or style file.

napari/layers/_markers_layer/view/properties.py
```diff
@@ -93,7 +93,10 @@ def changeSymbol(self, text):
         self.layer.symbol = text
 
     def changeSize(self, value):
-        self.layer.size = value
+        avg = np.asarray(self.layer.size).mean()
```
kne42:
```python
avg = np.mean(self.layer.size) or 1
```

napari/layers/_markers_layer/view/properties.py
```diff
@@ -93,7 +93,10 @@ def changeSymbol(self, text):
...
+        if avg == 0:
+            avg = 1
+        self.layer.size = self.layer.size/avg*value
```
kne42: pep8 and maybe some parentheses might be nice

kne42 (approved):
approved pending changes

napari/layers/_markers_layer/view/properties.py
sofroniewn: fixed

napari/layers/_markers_layer/view/properties.py
sofroniewn: fixed

napari/layers/_markers_layer/view/properties.py
```diff
@@ -93,7 +93,8 @@ def changeSymbol(self, text):
         self.layer.symbol = text
 
     def changeSize(self, value):
-        self.layer.size = value
+        avg = np.mean(self.layer.size) or 1
```
kevinyamauchi, line 97: I think it would be nice to have a comment explaining what is being done here.

kevinyamauchi (approved):
Awesome! My main comment is that I think there should be an explanation of how the size slider alters the marker sizes in the main docstring since the actual values of the size property are being changed. Other than that, looks good to me!

napari/layers/_markers_layer/view/properties.py
sofroniewn: agreed - good to note, can you make a followup issue?

napari/layers/_markers_layer/view/properties.py
sofroniewn: doc string added

sofroniewn:
Thanks for the reviews @kevinyamauchi and @kne42!

Issue #93, opened 2019-01-15
Figure out best printing format for status bar


jni:
Over in #90, @sofroniewn, @kne42, and I were discussing the formatting of numbers in the Napari status bar. No format string produces satisfactory results for all input types, so we need to write a function that picks the right format string for the right situation. See https://github.com/Napari/napari-gui/pull/90#discussion_r247782298

sofroniewn:
As noted by @royerloic in #372 - when changing the dims sliders 

> The status bar should display something like:
> 
> [3, 5, :, :], layer_name, value ...
>
> where 3 or 5 are changing depending on the dim slider being moved.
the ':' are there because we are not hovering over a pixel, so in some sense
we have nothing relevant to give there.

sofroniewn:
With #485 this has been dramatically improved. We now have a consolidated all our status formatting to one function inside the utils folder that can be tested / improved independently.

PR #278, opened 2019-05-19
fix int clim value

### LINKED ISSUE #154: Clims cannot be set to integer values
## 🐛 Bug

Clims cannot be set to integer values

## To Reproduce

```python
from napari import Window, Viewer
viewer = Viewer()
window = Window(viewer)
viewer.add_image(data.camera())
viewer.layers[0].clim = [0, 255]

ValueError                                Traceback (most recent call last)
<ipython-input-7-4d36bc0c2b58> in <module>()
----> 1 viewer.layers[0].clim = [0, 255]

~/Github/napari/napari/layers/_image_layer/model.py in clim(self, clim)
    265     @clim.setter
    266     def clim(self, clim):
--> 267         self._clim_msg = f'{clim[0]: 0.3}, {clim[1]: 0.3}'
    268         self.status = self._clim_msg
    269         self._node.clim = clim

ValueError: Precision not allowed in integer format specifier
```

## Expected behavior

This should be allowed. The problem is due to how we print the status bar, see #93 for a discussion around that. Resolving #93 should resolve this bug too.

## Environment

 - OS (e.g., Linux): OSX
 - Python version: 3.6.5

sofroniewn:
# Description
This PR fixes #154 by casting a clim value to a float before displaying it. This will result in pretty ugly values though like `255.000` when ints are needed. There's a broader thread about how to display numbers in #93, but for now I just wanted to fix this bug.

## Type of change

# How has this been tested?

- [x] run `example/add_image.py`

## Final checklist:

### DIFF
--- a/examples/add_image.py
+++ b/examples/add_image.py
@@ -40,7 +40,7 @@ with app_context():
 
     # change the layer colormap and color limits
     layer.colormap = 'gray'
-    layer.clim = (0.1, 0.9)
+    layer.clim = (0, 0.9)
 
     # change the layer interpolation mode
     layer.interpolation = 'bicubic'
--- a/napari/layers/_image_layer/model.py
+++ b/napari/layers/_image_layer/model.py
@@ -275,7 +275,7 @@ class Image(Layer):
 
     @clim.setter
     def clim(self, clim):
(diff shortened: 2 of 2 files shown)


jni (approved):
:+1: I agree that the problems can be addressed together in a future PR.

PR #281, opened 2019-05-21
rearrange layers folder structure

sofroniewn:
# Description
This PR rearranges the position of the individual `Layer` folders within our folder structure moving them into the `components._layers` folder. The hierarchical folder organization now mirrors the hierarchical organization of the objects themselves and I think is more intuitive. In follow-up PRs we might rename the `model.py` files to have more clear names, which might make what is the `layers_model.py` and `layers_view` vs other models and views more clear, but that can come later. No functionality is changed in this PR, only import path statements.

## Type of change

# How has this been tested?

- [x] run examples

## Final checklist:


napari/components/_layers/_base_layer/model.py
```diff
@@ -5,7 +5,7 @@
 
 import weakref
 
-from ...util.event import Event
+from ....util.event import Event
```
jni, line 8: Honestly, to me this says we have gone in too deep, and makes me :-1: on this PR. I don't think I've ever seen four leading dots in a Python import before.

sofroniewn: Agreed I don't love this either, but then I'd rather see us fix our `util.event` import (and other weird imports). To me the new organization really makes much more sense, and I think will be easier for new-comers to grasp. @kne42 what is your opinion?

sofroniewn:
Ok, after discussing with @kne42 in person, I recognize this level of nesting is undesirable and I'll be closing this PR. It is also the case that the folder organization didn't quite respect our object hierarchy either (`_layers` wasn't inside `_viewer` for example) so maybe this was just more confusing.

I will though make a very short PR after this to revert the name of the `Layers` object back to `LayersList`. I had changed it so that this PR would make more sense, but now it just leads to more confusion.

PR #282, opened 2019-05-21
Black formatter PR

AhmetCanSolak:
# Description

This PR is to see how `black` formatter would change some of our code.
Please check the details and let each other know about your comments.

## Type of change

# References

https://github.com/python/black

# How has this been tested?

Will be tested on review manually.

## Final checklist:


napari/components/_viewer/model.py
```diff
@@ -31,16 +31,19 @@ class Viewer:
         These key bindings are executed instead of any layer specific key
         bindings.
     """
-    def __init__(self, title='napari'):
+
+    def __init__(self, title="napari"):
```
sofroniewn: Is the any way to configure black to use `'` instead of `"`?

AhmetCanSolak: No as far as i know. This is something `black` has a strong preference. To me also it makes sense, as double `"` is used for strings as a widely popular convention. On top of that, it is consistent with C/OpenCL and so. IMO we are very likely to depend such languages/tools in the long run.

sofroniewn: yeah - i just found a super long [hacker news post](https://news.ycombinator.com/item?id=17151813) on this one! Let's just say some very strong opinions on either side

AhmetCanSolak: Everybody(including me) has an opinion on everything(especially in python world) :D that's why maybe better to decide according to priorities instead of opinions.

AhmetCanSolak: btw, good news it is possible to stop `black` on this with: `--skip-string-normalization`

sofroniewn: yes - very interesting, I just read this thread here https://github.com/python/black/issues/118, and the main developer very reasonably added that skip, which can be achieved with the short-cut `black . -S`

sofroniewn (changes requested):
I'm pretty happy with how `black` has done the new line spacing, and I love how it has inserted white space between operators, I think that will cut down the review time on my PRs quite a bit!

One thing I'd be curious to see though is the output of running with the `-S` (`--skip-string-normalization`) command. I prefer the ability to mix `'` and `"`, using `'` for things like dictionary names as I find them easier to read. If we like the results then I'd vote for `black -S`

jni:
Yeah I gotta say I'm pretty happy with the result! Thanks for running this experiment, @AhmetCanSolak!

- For *sure* I want `-S` on in our project. I'm firmly on one side of the debate here. =D
- I often prefer for operators to be spaced but "intelligently" grouped by priority, e.g. `length/2 - max_length/2` vs `length / 2 - max_length / 2`...
- but on the whole I think it's done a good job *and* if it clears up mental space for never thinking about formatting again, then that's a significant win.
- also, I'm surprised by the *low* number of changes it's proposed, which speaks to our generally good consistency in code style till now. Go team! =P

It'd be awesome if black could also format numpy-style docstrings, but that's a bit much to ask. =)

I'm moving my position from :-1: to neutral. =)

CC @stefanv (as we have considered `black` for skimage also)

AhmetCanSolak:
Thank you @sofroniewn and @jni !

Glad you guys like it and yes having `-S` will be okay too.

AhmetCanSolak:
all `pre-commit` and `black` integration business available now in this PR. Would you guys prefer to have a separate PR for all formatted version of codebase or prefer it also within this PR? @kne42  @sofroniewn @jni ?

sofroniewn:
I'm fine with one PR that does all the reformatting too and add the pre-commit. It looks like the build is failing now as it is not finding the right version of `pre-commit`

AhmetCanSolak:
eerh pypi and conda versions are different :( fixing it

.pre-commit-config.yaml
```diff
@@ -0,0 +1,6 @@
+repos:
+-   repo: https://github.com/python/black
+    rev: stable
+    hooks:
+    - id: black
+      language_version: python3.6
```
kne42, line 6: what happens when you only have python 3.7 installed and try to run this?

CONTRIBUTING.md
```diff
@@ -32,6 +32,11 @@ Make the development version available globally:
 $ pip install -e .
 ```
 
+Also do not forget to install black formatter hook with:
```
kne42:
```suggestion
Install the `black` formatter hook with:
```

.pre-commit-config.yaml
kne42: you may also want to look into python/black#675 and add `require_serial: true`

requirements/default.txt
```diff
@@ -6,3 +6,4 @@ vispy>=0.5.3
 PyOpenGL>=3.1.0
 PyQt5>=5.10.1
 SIP>=0.1.5
+pre-commit>=1.16.1
```
kne42: this is a dev-only tool so it doesn't belong in the default requirements

kne42 (commented):
I'd prefer to see the pre-commit stuff as a separate PR split into three commits:
1. actual pre-commit code/setup
2. documentation (README banner/CONTRIBUTING section)
3. the results of `pre-commit run --all-files`

We could also make this PR just the results of `black -S` and the pre-commit code as a separate PR.

requirements/default.txt
AhmetCanSolak: okay removing

.pre-commit-config.yaml
AhmetCanSolak: seems reasonable, i will be adding for now. we can update later if we observe non-serial is indeed better

requirements/default.txt
jni: I would say, let's add `requirements/development.txt` which includes pre-commit and black.

AhmetCanSolak:
Updated this PR. Will be happy if you guys @jni @sofroniewn @kne42 can review so i can continue with formatting the entire repo with black.

CONTRIBUTING.md
```diff
@@ -32,6 +32,11 @@ Make the development version available globally:
 $ pip install -e .
 ```
 
+Install the `black` formatter hook with:
```
sofroniewn, line 35: Can you maybe add something a bit more explicit here like, we use `black -S` to format our repo. More info can be found ... and put in a link to the `black` docs

sofroniewn (approved):
I'm approving as it looks close, though I did request some more detail in the contributing guidelines. Maybe @kne42 and @jni have further thoughts though

pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[tool.black]
+target-version = ['py36', 'py37']
+skip-string-normalization = true
```
jni, line 3: Can we add the max line length config to 79?

jni (changes requested):
@AhmetCanSolak I think @sofroniewn's comment is worth addressing in this PR, as well as line length. I think both should be quick work!

CONTRIBUTING.md
AhmetCanSolak: do we want to mention about command line usage `black -S` or enforce using pre-commit hook? i will be giving a link to black docs.

pyproject.toml
AhmetCanSolak: haha sure thing :D

CONTRIBUTING.md
sofroniewn: I think talking about command line usage for those that are interested / want to know, but then enforcing using pre-commit makes sense. Could just say this pre-commit equivalent to running `black -S`

AhmetCanSolak: I can't tell that pre-commit equivalent to `black -S` as it is not true. We have our line length and so. But i think enforcing pre-commit and giving a link to black repo could be satisfactory. Those interested can browse the link and find out more. How does this sound to you? @sofroniewn

sofroniewn: Ok - no need to say equivalent, but can you at least say that we have `skip-string-normalization = true` and `max-line-length = 79`

sofroniewn: i think it will just be easier for people to see this information here rather than have to dig around the repo

AhmetCanSolak: Done @sofroniewn please check

CONTRIBUTING.md
```diff
@@ -32,6 +32,15 @@ Make the development version available globally:
...
+```sh
+pre-commit install
+```
+
+With help of pre-commit, your future commits will be reformatted with our black configuration
+which includes settings like `skip-string-normalization = true` and `max-line-length = 79`.
+If you like to learn more please check [black](https://github.com/python/black).
```
sofroniewn: Can you change it to "If you would like to learn more", slightly better grammatically. Otherwise looks ready to merge

jni (approved):

jni:
Boom! =)

Issue #270, opened 2019-05-18
Using a formatter


AhmetCanSolak:
## 🚀 Feature

I am proposing to use `black` python formatter to decrease number of convention decisions to make and sustain consistency on code convention over repository. To me defaults of black seems great and i think you guys might like it as well.
Black formatter: https://github.com/python/black

## Motivation
Using a strong formatter decrease number of convention decisions to make. Reduces frustration on newcomers.

## Pitch
Black formatter is so cool. I compare it with `yapf` from google and autopep8 formatter.
Definitely black has an edge.

## Alternatives
Strongest alternatives: `yapf` from google: https://github.com/google/yapf 
and autopep8: https://github.com/hhatto/autopep8

jni:
I like the idea of formatters in general, but I dislike the rigidity of black (as I understand it it is not very configurable).

The number one concern is whether code formatted with black is actually nice and readable. Do you know any scientific packages that use it? Only the web-world appears to have taken it on.

I guess another way to propose this is to submit a PR with all the code formatted. Then we can see if we like its changes. =)

sofroniewn:
I'm open to a formatter, and `black` has come up at CZI recently, but I'm not sure any of our projects actually started using it. @kne42 might know more and I can follow-up.

A PR seeing what the new code would look like would be helpful and might be a good next step if it is not too much work to make. If the new code appears more readable and understandable then I'd be all for it, if we are constantly having to use `# fmt: off` and `# fmt: on` to make the code look nice then I'd probably want to pass

kne42:
Summarizing my comments from a different thread, I'm open but not thrilled to having an autoformatter. While it'll be great to finally have our codebase cleanly formatted, I don't think it's a proper substitution for people just actually writing well-formatted code.

At the end of the day, readability is something that a simple program cannot quantify; all it's doing is following a strict set of rules. Except styling doesn't consist of rules, it consists of guidelines, which means that following style guides to the letter does not necessarily make your code more readable, sometimes it can make your code _less_ readable.

Ideally, contributors should be following a style guide (*cough cough* PEP8) and otherwise format sensibly (which, if it isn't clear enough, we can make a guide for too).

However, given the current state of the codebase, I will admit that I have my doubts of letting contributors just go around willy-nilly without some sort of enforcement since I'd prefer not to have to leave 1000 comments on every PR about styling.

[...]

PR #283, opened 2019-05-21
demo yapf with default settings

kne42:
demo of `yapf` with default settings

see also: #282

### DIFF
--- a/napari/_qt/range_slider/range_slider.py
+++ b/napari/_qt/range_slider/range_slider.py
@@ -7,6 +7,7 @@ from qtpy.QtWidgets import QWidget
 
 from ...util.misc import str_to_rgb
 
+
 class QRangeSlider(QWidget):
     """
     QRangeSlider class, super class for QVRangeSlider and QHRangeSlider.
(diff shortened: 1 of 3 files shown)


napari/util/misc.py
```diff
@@ -129,7 +130,9 @@ def compute_max_shape(shapes, max_dims=None):
     if max_dims is None:
         max_dims = max(len(shape) for shape in shapes)
 
-    max_shape = [0, ] * max_dims
+    max_shape = [
+        0,
+    ] * max_dims
```
AhmetCanSolak, line 136: To me this looks awkward.

napari/util/misc.py
```diff
@@ -9,8 +9,9 @@
 def str_to_rgb(arg):
     """Convert an rgb string 'rgb(x,y,z)' to a list of ints [x,y,z].
     """
-    return list(map(int, re.match(r'rgb\((\d+),\s*(\d+),\s*(\d+)\)',
-                                  arg).groups()))
+    return list(
+        map(int,
+            re.match(r'rgb\((\d+),\s*(\d+),\s*(\d+)\)', arg).groups()))
```
AhmetCanSolak, line 15: I prefer this as one line

napari/components/_viewer/model.py
```diff
@@ -360,13 +366,13 @@ def _calc_layers_ranges(self):
...
+        ranges = [(inf, -inf, inf)] * ndims
 
         for layer in self.layers:
             layer_range = layer.range[::-1]
-            ranges = [(min(a, b), max(c, d), min(e, f)) for
-                      (a, c, e), (b, d, f) in zip_longest(ranges, layer_range,
-                      fillvalue=(inf, -inf, inf))]
+            ranges = [(min(a, b), max(c, d), min(e, f))
+                      for (a, c, e), (b, d, f) in zip_longest(
+                          ranges, layer_range, fillvalue=(inf, -inf, inf))]
```
AhmetCanSolak, line 376: for this section `yapf` seems more pythonic but new line in zip_longest argument is cringe

napari/_qt/range_slider/range_slider.py
```diff
@@ -405,29 +404,29 @@ def paintEvent(self, event):
         # Background
         painter.setPen(self.background_color)
         painter.setBrush(self.background_color)
-        painter.drawRect(w/2-self.slider_width/2, self.bar_width/2,
-                         self.slider_width, h-self.bar_width)
+        painter.drawRect(w / 2 - self.slider_width / 2, self.bar_width / 2,
+                         self.slider_width, h - self.bar_width)
```
AhmetCanSolak, line 409: on breaking arguments in newlines, black is much more readable to me

AhmetCanSolak (commented):
yapf is good. A strong option but i would prefer black due to points i mentioned in this review and previously in the #270 issue. Also enforcing 'a' for single characters and "string" for strings is widely accepted convention. I like that too.

napari/util/misc.py
kne42: This is a feature – lists, argument lists, etc. will be split apart in the same style as `black` when the last element ends with a comma.

napari/_qt/range_slider/range_slider.py
```diff
@@ -113,28 +114,28 @@ def mouseMoveEvent(self, event):
             size = self.rangeSliderSize()
             pos = self.getPos(event)
             if self.moving == "min":
```
sofroniewn, line 116: Does `yapf` not force `'` instead of `"`. I prefer `'`, but is seems strange to me that the auto formatter allows both

AhmetCanSolak: `yapf` does not force any by default

sofroniewn (commented):
I'm not so sure I find the new line breaks as readable as `black` in #282, but I do like that it preserved our `'`

napari/util/misc.py
jni: Yeah not a fan, feature or not. In this specific case it's *much* worse than the one-line version, because it is conceptually a one-line thing.

napari/_qt/range_slider/range_slider.py
jni: Yeah, this is one case where I have to revise my strongly-held opinion. I previously favoured "as many arguments on one line as possible", but seeing the two versions side by side, I think black's version (if some are in a new line, *all* are in a new line) is more readable.

napari/util/misc.py
```diff
@@ -147,7 +150,10 @@ def formatdoc(obj):
     """Substitute globals and locals into an object's docstring."""
     frame = inspect.currentframe().f_back
     try:
-        obj.__doc__ = obj.__doc__.format(**{**frame.f_globals, **frame.f_locals})
```
jni, line 150: Was this line too long? I don't know why it split it.

jni:
Since we now know that black can preserve our beloved single-quoted strings, based on the results, I prefer black over yapf.

kne42:
closed in favor of `black --skip-string-normalization` as demoed in #282

Issue #287, opened 2019-05-22
Dims slider can go to one too many slots


sofroniewn:
## 🐛 Bug

The dims slider can currently go to one too many slots, i.e. load a shape `[4, 512, 512]` image and the dims slider will have 5 possible positions, with the last one no longer changing the data

sofroniewn:
This has been fixed by #362

Issue #286, opened 2019-05-22
Changing dims slider no longer updates status message


sofroniewn:
## 🐛 Bug

Changing the dims slider no longer updates status message in the bottom left corner

PR #288, opened 2019-05-22
fix image dims update

### LINKED ISSUE #280: Z-slider doesn't change when changing image content
## 🐛 Bug

When a layer of the viewer is updated with an image with a different number of Z layers (say I go from 64 z-layers to 1 z-layer), the Z-slider does not adjust to that same range using `viewer.layers[name].image = new_image`

## To Reproduce

1. Load the viewer with an image that is 1 z-layer deep.
2. Load the viewer with an image that is 64 z-layers deep.
3. Try to move the z-slider.

The z-slider can be "reset" by calling `viewer._on_layers_change(None)` after every update.

## Environment

 - napari Version: recent??
 - MacOS
 - Python version: 3.7.3

sofroniewn:
# Description
This PR fixes #280 where changing the `image` data wouldn't trigger an update to the dims. This has now been fixed for an `Image` layer and the other layer types by adding a `data` event that gets triggered by changing this data and that in turn triggers a call to `_on_layers_change`. I have added an example `change_image_dims.py` that shows this behavior. Note that both changing the shape of the image data say `[64, 512, 512]` to `[4, 512, 512]` and the number of dims say `3D` to `2D` all should now work.

@gregjohnso if you can give this a spin and let us know if this now has the expected behviour that would be great!

In the process I have noticed two other bugs though that I will fix separately and I have filed separate issues for - #286, changing the dims slider no longer updates status message in the bottom left corner - and #287, the dims slider can currently go to one too many slots, i.e. load a shape `[4, 512, 512]` image and the dims slider will have 5 possible positions, with the last one no longer changing the data.

## Type of change

# How has this been tested?

- [x] run the new `examples/change_image_dims.py`

## Final checklist:

### DIFF
--- a/examples/change_image_dims.py
+++ b/examples/change_image_dims.py
@@ -1,0 +1,22 @@
+"""
+Change image shape directly on the layer
+"""
+
+import numpy as np
+from skimage import data
+from napari import ViewerApp
+from napari.util import app_context
+
+
+with app_context():
+    viewer = ViewerApp()
+
+    # add 3D data
+    blobs = data.binary_blobs(length=128, blob_size_fraction=0.05,
+                              n_dim=3, volume_fraction=.25).astype(float)
+
+    layer = viewer.add_image(blobs[:64])
+
+    # switch to 3D data
+    layer.image = blobs[:3]
+    #viewer._on_layers_change(None)
--- a/examples/change_image_dims.py
+++ b/examples/change_image_dims.py
@@ -1,5 +1,5 @@
 """
-Change image shape directly on the layer
+Change image shape and dims directly on the layer
 """
 
 import numpy as np
@@ -18,4 +18,10 @@ with app_context():
     layer = viewer.add_image(blobs[:64])
 
     # switch number of displayed dimensions
+    layer.image = blobs[0]
+
+    # switch number of displayed dimensions
+    layer.image = blobs[:64]
+
+    # switch the shape of the displayed data
     layer.image = blobs[:3]
--- a/examples/change_image_dims.py
+++ b/examples/change_image_dims.py
@@ -11,12 +11,11 @@ from napari.util import app_context
 with app_context():
     viewer = ViewerApp()
 
-    # add 3D data
+    # add data
     blobs = data.binary_blobs(length=128, blob_size_fraction=0.05,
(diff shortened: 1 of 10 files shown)


sofroniewn:
hmm - unclear to me why our linux tests are failing with the `examples/add_image.py` - it runs just fine for me locally. Maybe @kne42 has an idea?

jni:
@sofroniewn the failing example is `examples/set_theme.py`

jni:
(Or so it seems from the [log](https://github.com/napari/napari/pull/288/checks?check_run_id=133720411)...)

sofroniewn:
the `set_themes.py` example was failing on the other PR ;-) I think this one is failing on `test_example[examples/layers.py]`! Unfortunately if I do `pytest .` locally everything passes!

sofroniewn:
There was a bug where a qt signal was expecting an int but got a list, which failed silently on my laptop, but caused an error in our test. It has now been fixed, and this is ready for review

examples/change_image_dims.py
```diff
@@ -0,0 +1,27 @@
...
+"""
+
+import numpy as np
+from skimage import data
+from napari import ViewerApp
+from napari.util import app_context
+
+
+with app_context():
+    viewer = ViewerApp()
```
jni: This needs a rebase... =)

napari/components/_dims/view.py
```diff
@@ -170,10 +170,11 @@ def _trim_sliders(self, number_of_sliders):
         ----------
         number_of_sliders : new number of sliders
         """
-        for slider_num in range(self.nsliders, number_of_sliders):
+        for slider_num in range(number_of_sliders + 1):
```
jni: Can you explain this? I don't understand why there is a +1 here...

jni (approved):
@sofroniewn just a couple of comments but apart from that LGTM. The example itself needs to change or things won't be pretty on merge... =)

examples/change_image_dims.py
sofroniewn: done

napari/components/_dims/view.py
sofroniewn: this has now been fixed, comment has been added too

PR #290, opened 2019-05-23
Refactor viewer syntax

### LINKED ISSUE #134: Better syntax for launching a GUI programmatically
## 📚 Documentation

@royerloic pointed out that it's kinda clunky that the `QApplication` startup and shutdown logic are embedded at the start and end of each example. The whole thing certainly smells of a context manager! Suggestion:

```python
from skimage import data
from skimage.color import rgb2gray
from napari_gui import imshow, application

with application():
    image = rgb2gray(data.astronaut())
    viewer = imshow(image, {})
```

You could then nicely embed Napari within a bigger processing script:

```python
with application():
    viewer = imshow(image)
    viewer.add_markers()  # add clicks, but see #120  =P
    markers = viewer.layers[-1].markers

count = len(markers)
# ... more downstream processing...
```

If we are going for the OO API, I think the above would still work, but we might add a convenience method within the viewer?

```python
viewer = Viewer()

with viewer.start_application():
    # more good stuff

# downstream processing
```

jni:
# Description
Fixes #134.

I leave the extended discussion about LayerWidget/LayerModel etc for a later discussion and refactor. (I don't like that model because there is not necessarily a view for every model — views might instead combine information from different models.)

I removed the extra constructor arguments in the final commit, and that could be reverted. Specifically, I wonder what arguments make sense for an empty viewer. If there are not many, then we might as well allow the constructor to take in some images that mirror the `view` signature. But, as a counterpoint, there should be one obvious way to do things. So I'm happy with the way this PR stands. The examples now include a mix of using `napari.view(images)` and `napari.Viewer()`. 

## Type of change

# How has this been tested?

- [x] example: the test suite for my feature covers `napari.view` and `napari.Viewer`, and `ViewerWidget` by extension.

## Final checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ From inside an IPython shell or Jupyter notebook you can open up an interactive 
 %gui qt5
 from skimage import data
 from napari import ViewerApp
-viewer = ViewerApp(data.astronaut())
+viewer = Viewer(data.astronaut())
 ```
 
 ![image](resources/screenshot-add-image.png)
@@ -66,7 +66,7 @@ from napari import ViewerApp
 from napari.util import app_context
 
 with app_context():
-    viewer = ViewerApp(data.astronaut())
+    viewer = Viewer(data.astronaut())
 ```
 
 ## features
@@ -83,7 +83,7 @@ from napari.util import app_context
 
 with app_context():
     # create the viewer with four layers
-    viewer = ViewerApp(astronaut=rgb2gray(data.astronaut()),
+    viewer = Viewer(astronaut=rgb2gray(data.astronaut()),
                        photographer=data.camera(),
                        coins=data.coins(),
                        moon=data.moon())
@@ -105,7 +105,7 @@ from napari.util import app_context
 
 with app_context():
     # setup viewer
-    viewer = ViewerApp()
+    viewer = Viewer()
     viewer.add_image(rgb2gray(data.astronaut()))
     # create three xy coordinates
     points = np.array([[100, 100], [200, 200], [333, 111]])
@@ -141,7 +141,7 @@ with app_context():
                                         n_dim=3, volume_fraction=f)
                      for f in np.linspace(0.05, 0.5, 10)], axis=-1)
     # add data to the viewer
-    viewer = ViewerApp(blobs.astype(float))
+    viewer = Viewer(blobs.astype(float))
 ```
 
 ![image](resources/screenshot-nD-image.png)
--- a/README.md
+++ b/README.md
@@ -52,8 +52,8 @@ From inside an IPython shell or Jupyter notebook you can open up an interactive 
 ```python
 %gui qt5
 from skimage import data
-from napari import ViewerApp
-viewer = Viewer(data.astronaut())
+import napari
+viewer = napari.view(data.astronaut(), multichannel=True)
 ```
 
 ![image](resources/screenshot-add-image.png)
@@ -62,11 +62,11 @@ To do the same thing inside a script call
 
 ```python
 from skimage import data
-from napari import ViewerApp
+import napari
 from napari.util import app_context
 
 with app_context():
-    viewer = Viewer(data.astronaut())
+    viewer = napari.view(data.astronaut())
 ```
 
 ## features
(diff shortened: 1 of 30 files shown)


README.md
```diff
@@ -53,7 +53,7 @@ From inside an IPython shell or Jupyter notebook you can open up an interactive
 %gui qt5
 from skimage import data
 from napari import ViewerApp
```
sofroniewn: This should no longer be `ViewerApp` but `Viewer` or `view`

README.md
```diff
@@ -53,7 +53,7 @@ From inside an IPython shell or Jupyter notebook you can open up an interactive
...
-viewer = ViewerApp(data.astronaut())
+viewer = Viewer(data.astronaut())
```
sofroniewn: If the image is being passed directly then this should be `view`. I'll note that I like the clear distinction between `view` and `Viewer` now.

README.md
```diff
@@ -66,7 +66,7 @@ from napari import ViewerApp
 from napari.util import app_context
 
 with app_context():
-    viewer = ViewerApp(data.astronaut())
+    viewer = Viewer(data.astronaut())
```
sofroniewn: `view`

README.md
```diff
@@ -83,7 +83,7 @@ from napari.util import app_context
 
 with app_context():
     # create the viewer with four layers
-    viewer = ViewerApp(astronaut=rgb2gray(data.astronaut()),
+    viewer = Viewer(astronaut=rgb2gray(data.astronaut()),
```
sofroniewn: `view`

sofroniewn:
I like this. One thing that I'm now wondering though is if
`viewer = napari.view()` is equivalent to `viewer = napari.Viewer()`
then maybe all our examples should just use `napari.view()` (and either pass images or not).

It would still be possible to do `viewer = napari.Viewer()` if someone wanted to, but all our docs and examples would just have `napari.view()` or `napari.view(image)` and we would encourage `napari.view` as the canonical way to use `napari`. The reason for this is it then seems much simpler to a beginner - "oh you want to use napari, great - use `napari.view`" and there isn't this idea of two ways to do things. Thoughts?

jni:
@sofroniewn thanks for the review! Sorry I missed the README! Fixed now. Also, today I learned that PyCharm renders markdown while you're editing it *and* that it inspects code blocks for correctness! Mind officially blown.

> One thing that I'm now wondering though is if
> viewer = napari.view() is equivalent to viewer = napari.Viewer()
> then maybe all our examples should just use napari.view() (and either pass images or not).

> seems much simpler to a beginner - "oh you want to use napari, great - use napari.view" and there isn't this idea of two ways to do things. Thoughts?

I'm :-1: on this, though mildly. To me it's unpythonic to create classes with a zero-input function. The other reason is that `napari.view(image)` makes it clear that view is a transitive verb here. `napari.view()` now starts to sound like a noun, which makes the API slightly uglier.

Having said this, there's a perfect parallel with `plt.subplots()`, so I don't discard this idea out of hand by any means. Happy to hear input from others.

I also wouldn't necessarily block this PR to discuss this ;) — I'd like to get started on the singleton colormaps list.

kne42:
I don't like using `napari.view()` for just the viewer either. I'd say we should either raise a TypeError when no images are passed or not promote its usage.

examples/notebook.ipynb
```diff
@@ -43,7 +43,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "viewer = ViewerApp(data.moon())"
+    "viewer = Viewer(data.moon())"
```
sofroniewn: This should be `view`

examples/notebook.ipynb
```diff
@@ -27,7 +27,7 @@
     "# called\n",
     "\n",
     "from skimage import data\n",
-    "from napari import ViewerApp"
+    "from napari import Viewer"
```
sofroniewn: this should be view or the syntax below should change

sofroniewn (approved):
I'm approving this (subject to making the notebook example run) as I don't want to block everything unnecessarily, but I will note I find the use of `ViewerWidget` confusing as the thing being called `ViewerWidget` isn't a `Qt` element at all, but instead is the object inside our `model.py` file. Right now this still has the `_qtviewer` on it as a property, and so I am fine with it, but we are planning to remove that soon as part of the MVC refactor. At that point we should rename it to `ViewerModel`.  Then, if a developer wants to embed the napari viewer in their own application the will have to import both the `QtViewer` and `ViewerModel` - instantiate the model, then pass it to the `QtViewer` constructor and then embed the `QtViewer` (which I'm fine with being called `ViewerWidget`!).

If @jni is open to renaming `ViewerWidget` to `ViewerModel` right now then we can avoid one more name change, but we can also wait till the MVC refactor is complete.

I'll also note that I'd like to see some of the theme setting move out of what is now called `Viewer` and into what is now called `Window` and `QtViewer` but that can come in follow-up PRs.

The most important thing now is that we all come to agreement on `napari.view(image)` and `napari.Viewer()` as the two canonical ways to use `napari` and that we don't plan on changing that again before the `0.1` release and SciPy (and ideally after). If we can do that at the meeting today then I think we're making progress.

AhmetCanSolak (approved):

Issue #268, opened 2019-05-17
Smart Dropdown with text input search


AhmetCanSolak:
## 🚀 Feature

There are so many color and colormap options and we want to keep them. Also we want to provide a better experience on GUI. Hence, here i am proposing to have a dropdown with a text search as seen here: https://www.w3schools.com/howto/howto_js_filter_dropdown.asp

## Motivation

There are so many colors and colormaps. Right now it is not easy to grab you like.

## Pitch

This will make your life easier :)

## Alternatives
I can not think any better but i am open to suggestions.

## Additional context
-

jni:
To clarify the additional suggestion we had during the meeting: the actual list being searched could be bigger than the list being displayed without any search. In the CSS/js example I think you could do this with a `.initially-hidden` class that you apply to hidden elements on the list, and that you remove/ignore when you start filtering.

AhmetCanSolak:
Hi @jni , yes, i put css example to give a feeling. I will implement it in the way we discussed during the meeting.

royerloic:
This might be an interesting starting point:
https://stackoverflow.com/questions/4827207/how-do-i-filter-the-pyqt-qcombobox-items-based-on-the-text-input

PR #291, opened 2019-05-24
Simplify the colormaps list and add the single-color colormaps

### LINKED ISSUE #97: Add basic colormaps
## 🚀 Feature

As discussed in #89 it would be nice to add some standard colormaps that people might want to use (particularly in biology) like the following:

> Blues - black to blue
Reds - black to red
Greens - black to green
Magenta - black to magenta
Yellow - black to yellow
Cyan - black to cyan

It might also be nice to add some standard diverging ones too like red to blue through white or red to blue through black - though I don't think we want to try and make an exhaustive list

Similarly looking through the list of colormaps now it seems like there are some that could be removed as they are less popular

jni:
# Description

Fixes #97 

## Type of change

# References

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/layers/_image_layer/model.py
+++ b/napari/layers/_image_layer/model.py
@@ -67,8 +67,7 @@ def vispy_or_mpl_colormap(name):
 
 
 AVAILABLE_COLORMAPS = {k: vispy_or_mpl_colormap(k)
-                       for k in matplotlib_colormaps +
-                       list(vispy.color.get_colormaps())}
+                       for k in matplotlib_colormaps}
 
 
 @add_to_viewer
--- a/napari/layers/_image_layer/model.py
+++ b/napari/layers/_image_layer/model.py
@@ -66,10 +66,12 @@ def vispy_or_mpl_colormap(name):
     return cmap
 
 
-AVAILABLE_COLORMAPS = {k: vispy_or_mpl_colormap(k)
-                       for k in matplotlib_colormaps}
-AVAILABLE_COLORMAPS.update(simple_colormaps)
+# A dictionary mapping names to VisPy colormap objects
+ALL_COLORMAPS = {k: vispy_or_mpl_colormap(k) for k in matplotlib_colormaps}
+ALL_COLORMAPS.update(simple_colormaps)
 
+# ... sorted alphabetically by name
+AVAILABLE_COLORMAPS = {k: v for k, v in sorted(ALL_COLORMAPS.values())}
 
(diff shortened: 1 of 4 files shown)


sofroniewn (approved):
Do you want to fix the `examples/set_colormaps.py` so that it now uses the built in colormaps - or is it nice to have an example of setting custom colormaps? I could go either way, but if we want the latter we should probably have an example that no longer just uses the built-ins

I'm also still not completely sure we need all the remaining ones like `twilight` and `twilight-shifted` - did you intentionally leave those ones in there as we know they are useful?

I also find the order of the list confusing now, can you maybe make it alphabetical?

Finally, how would you feel about adding some two color diverging ones, like RdBl (or RdWBl) that start at a color, pass through black / white, then go to a color. Do we have a theory on how we want to handle these?

jni:
> Do you want to fix the examples/set_colormaps.py so that it now uses the built in colormaps - or is it nice to have an example of setting custom colormaps? I could go either way, but if we want the latter we should probably have an example that no longer just uses the built-ins

I still want that example. I agree that different cmaps from defaults might be useful but then the example should be totally reworked — I don't think it should hold that PR. It still illustrates the use case.

> I'm also still not completely sure we need all the remaining ones like twilight and twilight-shifted - did you intentionally leave those ones in there as we know they are useful?

These are cyclical colormaps that have been designed to be perceptually uniform. I think that's a very useful alternative to hsv (the usual cyclical colormap). Think of a "phase" or "orientation" image where black is down, white is up, blue is right, and red is left. Much more intuitive than hsv.

> I also find the order of the list confusing now, can you maybe make it alphabetical?

Yes, good point.

> Finally, how would you feel about adding some two color diverging ones, like RdBl (or RdWBl) that start at a color, pass through black / white, then go to a color. Do we have a theory on how we want to handle these?

Yeah Ok, I'll add back RdBl, I think that's the best there is currently, though it's not a "modern" one like viridis and twilight.

jni:
@sofroniewn please have a look at the diverging colormaps [here](https://matplotlib.org/3.1.0/tutorials/colors/colormaps.html). I think the best one is `PiYG`, which is what I've left on there. Let me know if you have a different preference. None of them go through black, interestingly.

jni:
This should be ready for merging — leaving the example for future work

sofroniewn:
`PiYG` is fine for now, we can always add more in the future after #268 is implemented. I'll merge

Issue #92, opened 2018-12-19
Where do qt elements get connected to functions


sofroniewn:
## ❓ Questions and Help

My question has two main parts to it. The first is where to qt elements get connected to functions? What I mean by this is right now take one of the buttons like out `QAddLayerButton` this needs to get connected to the function that actually adds the layer. It was previously done inside the `_layer_buttons` file inside the `qt` folder, but now in the latest on #90 I have moved it to inside the `_viewer` file which is not inside the `qt` folder. Basically - should this sort of logic be happening on files inside the `qt` folder or on files outside the `qt` folder. I lean towards the latter - I think it keeps the files in the `qt` folder then much simpler, and not having to do with the logic of the gui but just the actual elements that are getting rendered.

As a second part of this question - should the `viewer` or higher level object be passed as an input to the classes in the `qt` folder. I.e. should `dimensions` be passed to the `QtDimensions` class in the `qt` folder. Right now we do it in a whole bunch of places, but maybe as part of the refactor we should prevent this. Preventing this then forces us to put all the connection logic into the files outside the `qt` folder and means the files in the `qt` folder are just responsible for laying out the widgets and any self contained styling and logic.

What are peoples thoughts about this? @AhmetCanSolak @kne42 @royerloic @jni

jni:
@sofroniewn I have literally no thoughts on this. =P

One thing did strike me:

> Right now we do it in a whole bunch of places, but maybe as part of the refactor we should prevent this.

Without fully grasping the architecture now or what you're proposing, this sounds good to me. Single points of control are a Honking Good Idea.

sofroniewn:
It's really a question about what lives in `QtDimensions` vs `Dimensions` or in `QtLayer` vs `Layer` or in `QtLayerList` vs `LayerList` and whether the non-qt object should be passed to the qt object on constructions. To be more explicit - if you look at the current state of #90

In the `_dimensions` file in the `qt` folder is the following
```python
class QtDimensions(QWidget):
    def __init__(self, viewer):
```
and within that class `viewer` gets used in a couple of places. Instead should those usages be moved out of that file and into the `Dimensions` class?

In comparison in `_control_bars.py` within the `qt` folder it is just the following
```python
class QtControlBars(QWidget):
    def __init__(self):
```
and the file is very simple, creating the `qt` elements and doing the layouts and stylings. All our custom logic is inside the `ControlBars` class.

[...]

PR #292, opened 2019-05-24
Remove qtviewer from viewer

sofroniewn:
# Description
This PR would address more of our MVC refactor for example #92 by removing the `_qtviewer` from the `ViewerWidget` which will be renamed `ViewerModel`. It is currently not running as I need to fix some imports

## Type of change

# How has this been tested?

- [x] run examples

## Final checklist:

### DIFF
--- a/napari/components/_viewer/model.py
+++ b/napari/components/_viewer/model.py
@@ -60,7 +60,7 @@ class ViewerModel:
         self.key_bindings = {}
 
         # TODO: this should be eventually removed!
-        # attaced by QtViewer when it is constructed by the model
+        # attached by QtViewer when it is constructed by the model
         self._scene = None
 
         self.dims.events.axis.connect(lambda e: self._update_layers())
--- a/napari/components/_viewer/model.py
+++ b/napari/components/_viewer/model.py
@@ -10,7 +10,7 @@ from ...util.misc import has_clims
 from .._dims import Dims
 
 
-class ViewerWidget:
+class ViewerModel:
     """Viewer containing the rendered scene, layers, and controlling elements
     including dimension sliders, and control bars for color limits.
 
@@ -22,8 +22,6 @@ class ViewerWidget:
         List of contained layers.
     dims : Dimensions
         Contains axes, indices, dimensions and sliders.
-    camera : vispy.scene.Camera
-        ViewerWidget camera.
     key_bindings : dict of string: callable
         Custom key bindings. The dictionary key is a string containing the key
         pressed and the value is the function to be bound to the key event.
@@ -42,7 +40,6 @@ class ViewerWidget:
                                    title=Event,
                                    interactive=Event,
                                    cursor=Event,
-                                   cursor_size=Event,
                                    reset_view=Event,
                                    active_layer=Event)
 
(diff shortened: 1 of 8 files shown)


sofroniewn:
Bugs have now been fixed and this is ready for review. One thing I'll note is that `ViewerModel` still has a property on it `_scene` that gets set after it is passed to `QtViewer` which is needed for the individual `Layer` model files to access the scenecanvas. Once these have been refactored too this can be removed

napari/components/_viewer/model.py
```diff
@@ -59,30 +60,15 @@ def __init__(self, title='napari'):
         self.key_bindings = {}
 
         # TODO: this should be eventually removed!
-        # initialised by QtViewer when it is constructed by the model
-        self._qtviewer = None
+        # attaced by QtViewer when it is constructed by the model
```
kne42:
```suggestion
        # attached by QtViewer when it is constructed by the model
```
also is it necessary to have this on here? if it's `None`, none of the related code should run anyways

napari/components/_viewer/view/main.py
```diff
@@ -73,7 +76,42 @@ def __init__(self, viewer):
                 'standard': QCursor()
             }
 
-    def set_cursor(self, cursor, size=10):
+        self.viewer.events.interactive.connect(self._on_interactive)
+        self.viewer.events.cursor.connect(self._on_cursor)
+        self.viewer.events.reset_view.connect(self._on_reset_view)
+        self.viewer.layers.events.reordered.connect(self._update_canvas)
+
+    def screenshot(self, region=None, size=None, bgcolor=None):
```
kne42, line 84: What's the point of moving this here if no one is going to be able to use it?

kne42: I think this goes back to what @royerloic was hinting at during yesterday's meeting, that we need a controller class that combines the model and the given view (depending on backend) while also providing convenience functionality for the user.

kne42 (approved):
LGTM

napari/components/_viewer/model.py
sofroniewn: typo fixed. this is only initialized to `None`, it gets set to be the `scene` inside the `QtViewer`, but we will remove this in the future

napari/components/_viewer/view/main.py
sofroniewn: for now I forwarded the `screenshot` method inside the `Viewer` object so it's still accessible. Agreed a controller might be needed. Also now it's a little strange that `to_svg` is still on `ViewerModel`, but `screenshot` is on `QtViewer`. Maybe we can fix in future PRs though

napari/components/_viewer/model.py
kne42: Yes, but why is it initialized at all? It's going to break regardless; I don't think it belongs here.

sofroniewn: What do you mean by "break"? It runs fine now. The reason I put it here was that it gets used again below, line 248 `layer._parent = self._scene`. I'm not sure where else that could be

napari/components/_viewer/view/main.py
sofroniewn: `screenshot` is also able to be called from `Viewer`, there are no issues with `self`

napari/components/_viewer/model.py
kne42: ah, ok I didn't notice that. this was more a callback to the initial dims refactor PR which introduced this (and wasn't necessary at the time)

napari/components/_viewer/view/main.py
kne42: yes, when you first told me this i thought you had forwarded it on the class level but the instance level works fine

PR #293, opened 2019-05-24
refactor theme setting

kne42:
# Description
- add more general `palette` property and make `theme` a wrapper around it (as a preset)
- include all themes as a class variable
- move theme/palette setting to ViewerModel and make event-based

## Type of change

# References
see #253

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/components/_viewer/model.py
+++ b/napari/components/_viewer/model.py
@@ -6,7 +6,6 @@ from xml.etree.ElementTree import Element, tostring
 
 from ...util.event import EmitterGroup, Event
 from ...util.theme import palettes
-from ...util.misc import has_clims
 from .._dims import Dims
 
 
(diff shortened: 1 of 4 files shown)


sofroniewn (changes requested):
The code looks fine, but the range slider is getting the wrong color in the dark theme. If you run `add_image` on master you get this

image

If you run it on this PR you get this 

image

See how the color is lighter - not sure what's causing that. Maybe @freeman-lab has an idea

sofroniewn:
I really like how simple `Viewer` is now, nice job. If you can fix the range slider coloring then we can merge. Maybe @freeman-lab also wants to double check the rest of the stylings are ok too

kne42:
I've fixed the clims but the dimensions sliders are still not up-to-date. However, this is a bug present on master and this PR is not meant as a bug-fix.

Aforementioned fix is non-trivial without disrupting what this PR meant to clean up. Would recommend that we fix this by making the range slider style based on a stylesheet.

sofroniewn (approved):
This all looks good to me, confirmed that it is currently working. Fixing the dims sliders in a followup PR makes sense too

PR #211, opened 2019-04-10
2D image pyramids (WIP)

sofroniewn:
# Description
This PR adds support for massive 2D images that need to be accessed in a tiled fashion, including image pyramids. It does this by creating a new layer type called `pyramid` that inherits the `image` layer. It takes a list of images that form the pyramid with the base level at `0`. Right now I need to improve the switching between pyramid levels (assumes a 2x pyramid). It also has a `max_tile_shape`
that prevents too large images being sent to the gpu. There is not explicit tiling, but passing dask arrays or zarr arrays that have chunks in the 2D dimensions allows for very large images that cannot fit in RAM. Using dask opportunistic caching can prevent unnecessarily reloading often requested data. I have not explored computing image pyramids on the fly with dask `coarsen` but can investigate that too. Some of this was discussed in #103.

See bellow for an example gif with data from the [CAMELYON16](https://camelyon17.grand-challenge.org/Data/) pathology challenge. I converted the raw data to a precomputed image pyramid stored as a zarr file with 10 levels, with the following dimensions:
`[(97792, 221184), (48896, 110592), (24448, 55296), (12224, 27648), (6112, 13824), (3056, 6912), (1528, 3456), (764, 1728), (382, 864), (191, 432)]`.
The base of the pyramid is about 16GB. Each level was chunked with a `(300, 300)` chunking.

image_pyramid9

Note that drawing shapes / markers just works - it occurs in the coordinate system of the base layer of the pyramid regardless of what tile or level is displayed. If you wanted to get the shapes in the coordinates of on of the levels you'd just need to divide by the appropriate scale factor.

Right now I think there are some parts of this that might assume so 2D stuff - though we should make sure it supports nD too before a merge. I'm also open to the idea that this functionality is incorporated into the image layer directly and we don't need a new layer type, but maybe just a flag around if a list of image pyramids is passed.

## Type of change

- [x] New feature (non-breaking change which adds functionality)

# How has this been tested?

- [ ] there is no example yet :-( need to add one with small data just to show proof of concept

## Final checklist:
- [x] My PR is the minimum possible work for the desired functionality
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

Issue #103, opened 2019-02-05
Support visualization of lazy ndarrays


d-v-b:
## 🚀 Feature
Natively support lazily-defined arrays, e.g. memory-mapped arrays, distributed dask arrays, etc.

## Motivation
For large arrays that don't fit in memory (e.g., light sheet microscopy data), it's very convenient to use an interface that represents the data _as if_ it were an in-memory array. In python: numpy arrays via `np.memmap` or `dask.array`, hdf5 files via `h5py`; imagej has virtual stacks for the same application. 

It would be awesome for a python-native array visualization tool like Napari to recognize and support this kind of data / interface.

Disclaimer: I don't know the napari internals at all, so you might already have this functionally baked in, or it's on your roadmap. But I didn't see a public discussion about this topic, and I haven't seen such a discussion anywhere else, so this seemed the best place to start it :)

## Pitch
Suppose I have a 4D `dask` array distributed along axis 0 (a typical scheme for volumetric microscopy data, which might be stored 1 file : 1 3D timepoint ) and  I want to visualize 2D slices from this dataset. I would like to run code like this:
```python
viewer = win.viewer
layer = viewer.add_image(dask_array, meta) 
```
...and adjust sliders to see different planes and different timepoints, _all without having to load the entire dataset into memory_. In my example the data happens to be distributed along the 0th (time) axis, so seeking through time involves loading data from disk, while seeking through space (axes 1,2,3) involves displaying data that are already in memory. Alternatively, each image might be really really big and thus chunked in space AND time (i.e., all axes are distributed), and so just displaying a full plane for a single timepoint requires loading multiple chunks of data from disk. 

Working with lazy, distributed data adds lots of new complications: to reduce latency, the display software might asynchronously cache data along the distributed axis; the display code will need to know about the data's chunking scheme, the display code may need gui elements to indicate distributed vs local axes of the data, etc. 

I know that this scheme is standard for visualization of EM data; but for EM data, most interactive operations (like tracing neurites) are _local_, in that only a single, small, contiguous block of data needs to be in memory (i.e., displayable at low latency). But for live imaging data -- data with a temporal dimension -- analysts often want the entire temporal context of multiple spatial regions, and the ability to quickly "span" the dataset from start to stop, and the ability to visualize a post-processed version of the data, and a lot of other things that analysts working with dead tissues rarely, if ever, need to do. 

[...]

PR #295, opened 2019-05-26
Pyramid layer

sofroniewn:
# Description
This PR supersedes #211 and address loading multiresolution data as stored in an image pyramid, and discussed in #103.

It does this by creating a new layer type called pyramid that inherits the image layer. It takes a list of images that form the pyramid with the base level at 0. The pyramid need not be based on 2x downsampling, though that is the usecase I have most throughly tested.

I've added an example called `examples/add_pyramid.py` which creates some pyramid data to view in this way. The data created there is quite small, for ease of testing, and must fit in memory. I've also verified that the layer works using zarr / dask arrays and dynamic loading. I'll note that my `zarr` examples with a 100,000 x 200,000 base image has much better performance than the `add_pyramid.py` example though I'm not quite sure why.

This PR is ready for review and I'd like to get it in before more refactoring happens. Also right now it only supports 2D images, but in the future we can make it support nD.

Here's a gif of the included example:

pyramid_astro

## Type of change

# How has this been tested?

- [x] `examples/add_pyramid.py`

## Final checklist:

### DIFF
--- a/examples/add_pyramid.py
+++ b/examples/add_pyramid.py
@@ -28,5 +28,10 @@ with app_context():
     viewer.add_pyramid(pyramid)
 
     camera = viewer.window.qt_viewer.view.camera
-    camera.rect = (-.1*pyramid[0].shape[1], 0, 1.2*pyramid[0].shape[1],
-                   pyramid[0].shape[0])
+
+    # Set the view box of the camera to include the whole base image of the
+    # pyramid with a little padding. The view box is a 4-tuple of the x, y
+    # corner position followed by width and height
+    base_shape = pyramid[0].shape
+    camera.rect = (-0.1 * base_shape[1], -0.1 * base_shape[0],
+                   1.2 * base_shape[1], 1.2 * base_shape[0])
--- a/examples/add_pyramid.py
+++ b/examples/add_pyramid.py
@@ -1,0 +1,32 @@
+"""
+Displays an image pyramid
+"""
+
+from skimage import data
+from skimage.color import rgb2gray
+from skimage.transform import pyramid_gaussian
+import napari
+from napari.util import app_context
+import numpy as np
+
+
+image = data.astronaut()
+rows, cols, dim = image.shape
+
+
+# create pyramid from astronaut image
+astronaut=rgb2gray(data.astronaut())
+base = np.tile(astronaut, (16, 16))
+pyramid = list(pyramid_gaussian(base, downscale=2, multichannel=False))[:-8]
+print([p.shape[:2] for p in pyramid])
+
+with app_context():
+    # create the viewer
+    viewer = napari.Viewer()
+
+    # add image pyramid
+    viewer.add_pyramid(pyramid)
+
+    camera = viewer.window.qt_viewer.view.camera
+    camera.rect = (-.1*pyramid[0].shape[1], 0, 1.2*pyramid[0].shape[1],
+                   pyramid[0].shape[0])
--- a/examples/add_pyramid.py
+++ b/examples/add_pyramid.py
@@ -25,7 +25,7 @@ with app_context():
     viewer = napari.Viewer()
 
     # add image pyramid
-    viewer.add_pyramid(pyramid)
+    viewer.add_pyramid(pyramid, clim_range=[0, 255])
 
     camera = viewer.window.qt_viewer.view.camera
 
--- a/examples/add_pyramid.py
+++ b/examples/add_pyramid.py
@@ -15,7 +15,7 @@ rows, cols, dim = image.shape
 
 
 # create pyramid from astronaut image
-astronaut=rgb2gray(data.astronaut())
+astronaut = rgb2gray(data.astronaut())
 base = np.tile(astronaut, (16, 16))
 pyramid = list(pyramid_gaussian(base, downscale=2, multichannel=False))[:-8]
 print([p.shape[:2] for p in pyramid])
--- a/examples/add_pyramid.py
+++ b/examples/add_pyramid.py
@@ -26,11 +26,9 @@ with app_context():
     # add image pyramid
     viewer.add_pyramid(pyramid, clim_range=[0, 255])
 
-    camera = viewer.window.qt_viewer.view.camera
-
     # Set the view box of the camera to include the whole base image of the
     # pyramid with a little padding. The view box is a 4-tuple of the x, y
     # corner position followed by width and height
     base_shape = pyramid[0].shape
-    camera.rect = (-0.1 * base_shape[1], -0.1 * base_shape[0],
+    viewer.camera.rect = (-0.1 * base_shape[1], -0.1 * base_shape[0],
                    1.2 * base_shape[1], 1.2 * base_shape[0])
--- a/examples/add_pyramid.py
+++ b/examples/add_pyramid.py
@@ -3,22 +3,21 @@ Displays an image pyramid
 """
 
 from skimage import data
+from skimage.util import img_as_ubyte
+from skimage.color import rgb2gray
 from skimage.transform import pyramid_gaussian
 import napari
 from napari.util import app_context
 import numpy as np
 
 
-image = data.astronaut()
-rows, cols, dim = image.shape
-
 # create pyramid from astronaut image
-astronaut = data.astronaut().mean(axis=2) / 255
+astronaut = rgb2gray(data.astronaut())
 base = np.tile(astronaut, (16, 16))
 pyramid = list(pyramid_gaussian(base, downscale=2, max_layer=5,
                                 multichannel=False))
-pyramid = [(255*p).astype('uint8') for p in pyramid]
-print([p.shape[:2] for p in pyramid])
+pyramid = [img_as_ubyte(p) for p in pyramid]
+print('pyramid level shapes: ', [p.shape for p in pyramid])
 
 with app_context():
     # create the viewer
--- a/examples/add_pyramid.py
+++ b/examples/add_pyramid.py
@@ -3,7 +3,6 @@ Displays an image pyramid
 """
 
 from skimage import data
-from skimage.color import rgb2gray
 from skimage.transform import pyramid_gaussian
 import napari
 from napari.util import app_context
@@ -13,11 +12,12 @@ import numpy as np
 image = data.astronaut()
 rows, cols, dim = image.shape
 
-
 # create pyramid from astronaut image
-astronaut = rgb2gray(data.astronaut())
+astronaut = data.astronaut().mean(axis=2) / 255
 base = np.tile(astronaut, (16, 16))
-pyramid = list(pyramid_gaussian(base, downscale=2, multichannel=False))[:-8]
+pyramid = list(pyramid_gaussian(base, downscale=2, max_layer=5,
+                                multichannel=False))
+pyramid = [(255*p).astype('uint8') for p in pyramid]
 print([p.shape[:2] for p in pyramid])
 
 with app_context():
--- a/napari/components/_viewer/model.py
+++ b/napari/components/_viewer/model.py
@@ -61,7 +61,7 @@ class ViewerModel:
 
         # TODO: this should be eventually removed!
         # attached by QtViewer when it is constructed by the model
-        self._scene = None
+        self._view = None
 
         self.dims.events.axis.connect(lambda e: self._update_layers())
         self.layers.events.added.connect(self._on_layers_change)
@@ -244,8 +244,8 @@ class ViewerModel:
         layer.events.name.connect(self._update_name)
 
         self.layers.append(layer)
+        layer.parent = self._view
         layer.indices = self.dims.indices
-        layer._parent = self._scene
 
         if self.theme is not None and has_clims(layer):
             palette = palettes[self.theme]
--- a/napari/components/_viewer/model.py
+++ b/napari/components/_viewer/model.py
@@ -284,8 +284,8 @@ class ViewerModel:
         layer.events.name.connect(self._update_name)
         layer.events.data.connect(self._on_layers_change)
 
-        self.layers.append(layer)
         layer.parent = self._view
+        self.layers.append(layer)
         layer.indices = self.dims.indices
 
         if len(self.layers) == 1:
--- a/napari/components/_viewer/view/main.py
+++ b/napari/components/_viewer/view/main.py
@@ -28,19 +28,19 @@ class QtViewer(QSplitter):
         self.canvas.connect(self.on_mouse_release)
         self.canvas.connect(self.on_key_press)
         self.canvas.connect(self.on_key_release)
+        self.canvas.connect(self.on_draw)
 
         self.view = self.canvas.central_widget.add_view()
-
-        # TO DO: Remove
-        self.viewer._scene = self.view.scene
 
         # Set 2D camera (the camera will scale to the contents in the scene)
         self.view.camera = PanZoomCamera(aspect=1)
         # flip y-axis to have correct aligment
         self.view.camera.flip = (0, 1, 0)
         self.view.camera.set_range()
-
         self.view.camera.viewbox_key_event = viewbox_key_event
+
+        # TO DO: Remove
+        self.viewer._view = self.view
 
         center = QWidget()
         center_layout = QVBoxLayout()
@@ -172,6 +172,11 @@ class QtViewer(QSplitter):
         if layer is not None:
             layer.on_key_release(event)
 
+    def on_draw(self, event):
+        """Called whenever drawn in canvas. Called for all layers, not just top
+        """
+        for layer in self.viewer.layers:
+            layer.on_draw(event)
 
 def viewbox_key_event(event):
     """ViewBox key event handler
--- a/napari/layers/_base_layer/model.py
+++ b/napari/layers/_base_layer/model.py
@@ -57,8 +57,8 @@ class Layer(VisualWrapper, ABC):
         self._cursor = 'standard'
         self._cursor_size = None
         self._interactive = True
-        self._indices = ()
-        self._cursor_position = (0, 0)
+        self._indices = (0, 0)
+        self._position = (0, 0)
         self._coordinates = (0, 0)
         self._name = ''
         self.events.add(select=Event,
@@ -110,33 +110,45 @@ class Layer(VisualWrapper, ABC):
         if indices == self.indices:
             return
         self._indices = indices[-self.ndim:]
-        self.coordinates = None
+        self._update_coordinates()
         self._set_view_slice()
 
     @property
     def coordinates(self):
-        """Tuple of float: Coordinates of the cursor in the respective image
-        space of each layer.
-
-        The setter expects the a 2-tuple of coordinates in canvas space
-        ordered (x, y) and then transforms them to image space and inserts
-        them into the correct position of the layer indices. The length of the
-        tuple is equal to the number of dimensions of the layer. If ``None`` is
-        passed then (0, 0) will replace necessary the layer indices.
+        """Tuple of float: Coordinates of the cursor in the image space of each
+        layer. The length of the tuple is equal to the number of dimensions of
+        the layer.
         """
         return self._coordinates
 
     @coordinates.setter
-    def coordinates(self, cursor_position):
-        if cursor_position is not None:
-            transform = self._node.canvas.scene.node_transform(self._node)
-            position = tuple(transform.map(cursor_position)[:2])
-        else:
-            position = (0, 0)
+    def coordinates(self, coordinates):
+        if self._coordinates == coordinates:
+            return
+        self._coordinates = coordinates
+
+    @property
+    def position(self):
+        """2-tuple of int: Cursor position in canvas ordered (x, y)."""
+        return self._position
+
+    @position.setter
+    def position(self, position):
+        if self._position == position:
+            return
+        self._position = position
+        self._update_coordinates()
+
+    def _update_coordinates(self):
+        """Insert the cursor position (x, y) into the correct position in the
+        tuple of indices and update the cursor coordinates.
+        """
+        transform = self._node.canvas.scene.node_transform(self._node)
+        position = transform.map(list(self.position))[:2]
         coords = list(self.indices)
         coords[-2] = position[1]
         coords[-1] = position[0]
-        self._coordinates = tuple(coords)
+        self.coordinates = tuple(coords)
 
     @property
     @abstractmethod
@@ -389,5 +401,10 @@ class Layer(VisualWrapper, ABC):
 
     def on_key_release(self, event):
         """Called whenever key released in canvas.
+        """
+        return
+
+    def on_draw(self, event):
+        """Called whenever the canvas is drawn.
         """
         return
--- a/napari/layers/_base_layer/model.py
+++ b/napari/layers/_base_layer/model.py
@@ -40,6 +40,9 @@ class Layer(VisualWrapper, ABC):
     shape
     selected
     indices
+    coordinates : tuple of float
+        Coordinates of the cursor in the image space of each layer. The length
+        of the tuple is equal to the number of dimensions of the layer.
 
     Methods
     -------
@@ -59,7 +62,7 @@ class Layer(VisualWrapper, ABC):
         self._interactive = True
         self._indices = (0, 0)
         self._position = (0, 0)
-        self._coordinates = (0, 0)
+        self.coordinates = (0, 0)
         self._name = ''
         self.events.add(select=Event,
                         deselect=Event,
@@ -112,20 +115,6 @@ class Layer(VisualWrapper, ABC):
         self._indices = indices[-self.ndim:]
         self._update_coordinates()
         self._set_view_slice()
-
-    @property
-    def coordinates(self):
-        """Tuple of float: Coordinates of the cursor in the image space of each
-        layer. The length of the tuple is equal to the number of dimensions of
-        the layer.
-        """
-        return self._coordinates
-
-    @coordinates.setter
-    def coordinates(self, coordinates):
-        if self._coordinates == coordinates:
-            return
-        self._coordinates = coordinates
 
     @property
     def position(self):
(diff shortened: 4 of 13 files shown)


royerloic:
Looks great!
The performance hit you observe might be due to the number 
of visuals in the scene graph...  Do you draw images that are outside
of the field of view? How is the chunking done? It is better to have
as large as possible images (per visual) than have lots of smaller visuals
in the scene graph...

sofroniewn:
@royerloic the performance hit i'm referring too is for an in-memory example vs a zarr example, which surprises me (zarr is faster?). I need to do some `timeit` and profiling with identical data to really see if it's real though, for now I'd just say it's anecdotal.

To answer your other question though, there's only ever one visual in the scene graph, there's a `max_tile_shape` amount of data that ever gets sent over to `vispy`, and that gets swapped out depending on the zoom level or if there has been so much translation that another one needs to be viewed. I'm really happy with the performance of the `zarr` example (which isn't included as the data is so big), so I was a little disappointed to see some lag with the `in-memory` example I have provided, but as I noted above there are a couple other differences that could be causing the lag and I havn't investigated yet

examples/add_pyramid.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+image = data.astronaut()
+rows, cols, dim = image.shape
+
+
+# create pyramid from astronaut image
+astronaut = rgb2gray(data.astronaut())
+base = np.tile(astronaut, (16, 16))
+pyramid = list(pyramid_gaussian(base, downscale=2, multichannel=False))[:-8]
```
jni: I think it's nicer to use the `max_layer` kwarg here?

examples/add_pyramid.py
```diff
@@ -0,0 +1,32 @@
...
+with app_context():
+    # create the viewer
+    viewer = napari.Viewer()
+
+    # add image pyramid
+    viewer.add_pyramid(pyramid)
+
+    camera = viewer.window.qt_viewer.view.camera
+    camera.rect = (-.1*pyramid[0].shape[1], 0, 1.2*pyramid[0].shape[1],
+                   pyramid[0].shape[0])
```
jni: This logic is crazy and should be explained in comments. =)

examples/add_pyramid.py
```diff
@@ -0,0 +1,32 @@
+"""
+Displays an image pyramid
```
jni, line 2: I think an in-memory pyramid is an unlikely scenario, so if performance is better with zarr, why not create a zarr file with this dataset as the first part of the example?

napari/components/_viewer/view/main.py
```diff
@@ -28,20 +28,20 @@ def __init__(self, viewer):
...
-
         # Set 2D camera (the camera will scale to the contents in the scene)
         self.view.camera = PanZoomCamera(aspect=1)
         # flip y-axis to have correct aligment
         self.view.camera.flip = (0, 1, 0)
         self.view.camera.set_range()
-
         self.view.camera.viewbox_key_event = viewbox_key_event
 
+        # TO DO: Remove
```
jni, line 49: Hmmmmm...

napari/components/_viewer/model.py
```diff
@@ -244,8 +244,8 @@ def add_layer(self, layer):
         layer.events.name.connect(self._update_name)
 
         self.layers.append(layer)
+        layer.parent = self._view
```
jni, line 287: Do we really want this to be public?

napari/components/_viewer/view/main.py
```diff
@@ -172,6 +172,11 @@ def on_key_release(self, event):
         if layer is not None:
             layer.on_key_release(event)
 
+    def on_draw(self, event):
+        """Called whenever drawn in canvas. Called for all layers, not just top
+        """
+        for layer in self.viewer.layers:
+            layer.on_draw(event)
```
jni, line 196: How often is this called? It seems like it would be running hundreds of times a second?

jni (commented):
@sofroniewn sorry haven't got time to finish my review right now, but I've made a couple of suggestions that might help already. Will do more thorough review later.

napari/components/_viewer/view/main.py
sofroniewn: on draw is called whenever there is a cursor movement, or the scene graph changes. It's not so frequent. i.e. if nothing is happening it doesn't get called, and if something happens it only gets called for each something

napari/components/_viewer/model.py
sofroniewn: probably not, but this will hopefully get removed soon. is it ok to leave for now?

napari/components/_viewer/view/main.py
sofroniewn: yeah, this is needed until we pull the vispy stuff out of each layer's model file and into it's own view file. i'll work on that after you remove `add_to_viewer`

napari/components/_viewer/model.py
sofroniewn: it will get removed when we pull out the vispy stuff from each layer's model file, as mentioned bellow

examples/add_pyramid.py
sofroniewn: done

examples/add_pyramid.py
sofroniewn: So I've done some testing with zarr, identical data, and using timeit and the difference doesn't seem to be that great, though there were a couple places where it was a bit faster, and the times are right around what is noticeable for a human - i.e. a factor of 2x can be huge in this context, as can a difference of 5-10 ms. It's still unclear to me why the one example I have on my laptop seems less laggy than this one, and unfortunately `cProfile` isn't being very helpful, it's not really catching the layer updating, which `timeit` does catch. Not sure if thats something with how `cProfile` does events / multiple threads. Anyway, for now I'd say we just live with the performance, as it will be hard for me to debug without `cProfile` working properly

examples/add_pyramid.py
sofroniewn: a comment explaining this how now been added

examples/add_pyramid.py
sofroniewn: https://github.com/sumerc/yappi this might be a solution for profiling multiple threads

sofroniewn: Hmm just gave `yappi` a try too and it's only reporting one thread and still missing the functional calls on the layers that are triggered by the event and that `timeit` is including when it measures the total duration of the calls. I'm at a loss

henrypinkard:
@sofroniewn continuing our discussion from #211: I think the add_pyramid example you have put up is sufficient for testing on my end. The datasets I'm using are stored in unstitched 2D tiles that can be loaded and stitched as needed. So I think maybe I will need to implement a way to do this on demand.

napari/components/_viewer/view/main.py
kne42: you mean camera, not cursor movement, right? on cursor seems a bit extreme...

sofroniewn: yes, camera - sorry!

napari/layers/_base_layer/model.py
```diff
@@ -110,33 +110,45 @@ def indices(self, indices):
...
         return self._coordinates
 
     @coordinates.setter
-    def coordinates(self, cursor_position):
-        if cursor_position is not None:
-            transform = self._node.canvas.scene.node_transform(self._node)
-            position = tuple(transform.map(cursor_position)[:2])
-        else:
-            position = (0, 0)
+    def coordinates(self, coordinates):
```
kne42: what's the point of this being a property if nothing special happens? I would just turn this into a regular instance attribute

napari/layers/_pyramid_layer/model.py
```diff
@@ -0,0 +1,249 @@
...
+        self._update_coordinates()
+        self.refresh()
+
+    def _update_image_from_pyramid(self):
+        """Updates the image data from the pyramid based on requested tile
+        and pyramid level
+        """
+        if np.any(self._image_shapes[self.pyramid_level] >
+                  self._max_tile_shape):
+            slices = [slice(self._top_left[i], self._top_left[i] +
```
kne42: since you immediately turn it into a tuple, this should just be `tuple(generator expression)`

napari/layers/_pyramid_layer/model.py
```diff
@@ -0,0 +1,249 @@
...
+            Position of mouse cursor in data.
+        value : int or float or sequence of int or float
+            Value of the data at the coord.
+        msg : string
+            String containing a message that can be used as
+            a status update.
+        """
+        coord = np.round(self.coordinates).astype(int)
+        if self.multichannel:
+            shape = self._image_view.shape[:-1]
```
kne42, line 138: this code seems to be repeated a lot. maybe you should have a helper function perform this cropping?

napari/layers/_pyramid_layer/model.py
```diff
@@ -0,0 +1,249 @@
...
+        else:
+            shape = self._image_view.shape
+        coord[-2:] = np.clip(coord[-2:], 0, np.asarray(shape) - 1)
```
kne42: `shape` will always be a tuple, so you should use `np.array` or even better, since you're purely using this to achieve broadcasting, I would use either a list comprehension or `np.subtract`

napari/layers/_pyramid_layer/model.py
```diff
@@ -0,0 +1,249 @@
...
+        value = self._image_view[tuple(coord[-2:])]
+
+        pos_in_slice = (self.coordinates[-2:] +
```
kne42: operators should be leading in multiline expressions:
```python
 pos_in_slice = (self.coordinates[-2:]
                 + self.translate[[1, 0]] / self.scale[:2])

File: napari/layers/_pyramid_layer/model.py [hunk 12]
```diff
@@ -0,0 +1,249 @@
...
+            String containing a message that can be used as
+            a status update.
+        """
+        coord = np.round(self.coordinates).astype(int)
+        if self.multichannel:
+            shape = self._image_view.shape[:-1]
+        else:
+            shape = self._image_view.shape
+        coord[-2:] = np.clip(coord[-2:], 0, np.asarray(shape) - 1)
+        value = self._image_view[tuple(coord[-2:])]
```
kne42, line 142: is it actually necessary to convert `coord` to a tuple?

napari/layers/_pyramid_layer/model.py
```diff
@@ -0,0 +1,249 @@
...
+
+        pos_in_slice = (self.coordinates[-2:] +
+                        self.translate[[1, 0]] / self.scale[:2])
+
+        # Make sure pos in slice doesn't go off edge
+        shape = self._image_shapes[self._pyramid_level]
+        coord = np.clip(pos_in_slice, 0, np.asarray(shape) - 1)
```
kne42: see earlier comment

napari/layers/_pyramid_layer/model.py
```diff
@@ -0,0 +1,249 @@
...
+        coord = np.round(coord*self.scale[:2]).astype(int)
```
kne42: pep8

napari/layers/_pyramid_layer/model.py
```diff
@@ -0,0 +1,249 @@
...
+
+        return coord, value
+
+    def compute_pyramid_level(self, camera):
+        """Computed what level of the pyramid should be viewed given the
+        current camera position.
+
+        Parameters
+        ----------
+        camera : Camera
```
kne42: this should be more like `camera : path.to.vispy.PanZoomCamera`

napari/layers/_pyramid_layer/model.py
```diff
@@ -0,0 +1,249 @@
...
+            Vispy PanZoomCamera object.
```
kne42: better description please

napari/layers/_pyramid_layer/model.py
```diff
@@ -0,0 +1,249 @@
...
+
+        Returns
+        ----------
+        level : int
+            Level of the pyramid to be viewing.
+        """
+        # Requested field of view from the camera in log units
+        size = np.log2(np.array(camera.rect.size).max())
```
kne42:
```suggestion
        size = np.log2(np.max(camera.rect.size))
```

napari/layers/_pyramid_layer/model.py
```diff
@@ -0,0 +1,249 @@
...
+
+        # Max allowed tile in log units
+        max_size = np.log2(self._max_tile_shape.max())
+
+        # Allow for 2x coverage of field of view with max tile
+        diff = size - max_size + 1
+
+        # Find closed downsample level to diff
+        level = np.argmin(abs(np.log2(self._image_downsamples)-diff))
```
kne42: pep8

napari/layers/_pyramid_layer/model.py
```diff
@@ -0,0 +1,249 @@
...
+
+        return level
+
+    def find_top_left(self):
+        """Finds the top left pixel of the canvas. Depends on the current
+        pan and zoom position
+
+        Returns
+        ----------
+        top_left : np.ndarry
```
kne42:
```suggestion
        top_left : (2,) int array
```

napari/layers/_pyramid_layer/model.py
```diff
@@ -0,0 +1,249 @@
...
+            Length two array with coordinates of top left pixel.
```
kne42:
```suggestion
            Coordinates of top left pixel.
```

napari/layers/_pyramid_layer/model.py
```diff
@@ -0,0 +1,249 @@
...
+        """
+
+        # Find image coordinate of top left canvas pixel
+        transform = self._node.canvas.scene.node_transform(self._node)
+        pos = transform.map([0, 0])[:2] + self.translate[:2]/self.scale[:2]
```
kne42: pep8

napari/layers/_pyramid_layer/model.py
```diff
@@ -0,0 +1,249 @@
...
+
+        shape = self._image_shapes[0]
+
+        # Clip according to the max image shape
+        pos = [np.clip(pos[1], 0, shape[0]-1),
```
kne42: pep8

napari/layers/_pyramid_layer/model.py
```diff
@@ -0,0 +1,249 @@
...
+
+    def get_message(self, coord, value):
+        """Generate a status message based on the coordinates and information
+        about what shapes are hovered over
+
+        Parameters
+        ----------
+        coord : sequence of int
+            Position of mouse cursor in image coordinates.
+        value : int or float or sequence of int or float
```
kne42:
```suggestion
        value : int, float, or sequence of int or float
```

napari/layers/_pyramid_layer/model.py
```diff
@@ -0,0 +1,249 @@
...
+            Value of the data at the coord.
+
+        Returns
+        ----------
+        msg : string
+            String containing a message that can be used as a status update.
+        """
+
+        msg = f'{coord}, {self.pyramid_level}, {self.name}' + ', value '
```
kne42:
```suggestion
        msg = f'{coord}, {self.pyramid_level}, {self.name}, value '
```

kne42: would also much prefer you to determine the value string first and then
```python
return msg = f'{coord}, {self.pyramid_level}, {self.name}, value {value}'
```

napari/layers/_pyramid_layer/model.py
```diff
@@ -0,0 +1,249 @@
...
+        if isinstance(value, np.ndarray):
+            if isinstance(value[0], np.integer):
```
kne42: uhhh don't you mean `value.dtype == np.int`?

napari/layers/_pyramid_layer/model.py
```diff
@@ -0,0 +1,249 @@
...
+                msg = msg + str(value)
+            else:
+                v_str = '[' + str.join(', ', [f'{v:0.3}' for v in value]) + ']'
```
kne42, line 229:
```suggestion
                v_str = '[' + ', '.join(f'{v:0.3}' for v in value) + ']'
```

napari/layers/_pyramid_layer/model.py
```diff
@@ -0,0 +1,249 @@
...
+                msg = msg + v_str
+        else:
+            if isinstance(value, np.integer):
```
kne42:
```suggestion
            if isinstance(value, int):
```

napari/layers/_pyramid_layer/model.py
kne42: potentially better (have not investigated): `np.array_str(value, precision=3)` or `np.array_repr`

kne42 (changes requested):
Looks good! Left a lot of comments on small NumPy stuff, documentation, pep8, and some design choices.

napari/layers/_base_layer/model.py
sofroniewn: made a regular instance attribute

napari/layers/_pyramid_layer/model.py
sofroniewn: fixed

napari/layers/_pyramid_layer/model.py
sofroniewn: I'm not quite sure which bit you're referring to or if you mean repeated across different layers or within this layer?

napari/layers/_pyramid_layer/model.py
sofroniewn: fixed with `np.subtract`

napari/layers/_pyramid_layer/model.py
sofroniewn: fixed

napari/layers/_pyramid_layer/model.py
sofroniewn: yes, i get errors if I don't, something about how it's actually doing the slicing seems different

napari/layers/_pyramid_layer/model.py
sofroniewn: fixed

napari/layers/_pyramid_layer/model.py
sofroniewn: fixed

napari/layers/_pyramid_layer/model.py
sofroniewn: I'm having this function accept a tuple of the requested field of view size instead, which I think is much cleaner

napari/layers/_pyramid_layer/model.py
sofroniewn: As noted above, I'm having this function accept a tuple of the requested field of view size instead, which I think is much cleaner

napari/layers/_pyramid_layer/model.py
sofroniewn: fixed

napari/layers/_pyramid_layer/model.py
sofroniewn: fixed

napari/layers/_pyramid_layer/model.py
sofroniewn: fixed

napari/layers/_pyramid_layer/model.py
sofroniewn: fixed

napari/layers/_pyramid_layer/model.py
sofroniewn: fixed

napari/layers/_pyramid_layer/model.py
sofroniewn: fixed

napari/layers/_pyramid_layer/model.py
sofroniewn: fixed

napari/layers/_pyramid_layer/model.py
sofroniewn: have tried to make this better. @jni will also be working on generally improving our string generation

napari/layers/_pyramid_layer/model.py
sofroniewn: I'd say this is for #93 and @jni to address

napari/layers/_pyramid_layer/model.py
sofroniewn: this get's it wrong if I just have `int`, I need `np.integer`, again I'd say for #93 to fix

napari/layers/_pyramid_layer/model.py
sofroniewn: fixed

sofroniewn:
I have gone through and made most of the changes requested by @kne42. There are some around printing strings for the status bar that I didn't address, but those are for #93 and @jni to look it.

If @jni could review this too it would be much appreciated!

examples/add_pyramid.py
```diff
@@ -0,0 +1,37 @@
...
+
+from skimage import data
+from skimage.transform import pyramid_gaussian
+import napari
+from napari.util import app_context
+import numpy as np
+
+
+image = data.astronaut()
+rows, cols, dim = image.shape
```
jni: dim? I suggest ch or channels. But anyway, you don't use it later, so just delete this line!

examples/add_pyramid.py
```diff
@@ -0,0 +1,37 @@
...
+
+# create pyramid from astronaut image
+astronaut = data.astronaut().mean(axis=2) / 255
```
jni: `skimage.color.rgb2gray`? (This weights the color channels properly.)

examples/add_pyramid.py
```diff
@@ -0,0 +1,37 @@
...
+base = np.tile(astronaut, (16, 16))
+pyramid = list(pyramid_gaussian(base, downscale=2, max_layer=5,
+                                multichannel=False))
+pyramid = [(255*p).astype('uint8') for p in pyramid]
+print([p.shape[:2] for p in pyramid])
```
jni: Users running the demos will see this, so it's probably good to print a message with this also: `print('pyramid level shapes: ', ...)`

You also don't need the `:2` anymore.

examples/add_pyramid.py
```diff
@@ -0,0 +1,37 @@
...
+
+image = data.astronaut()
+rows, cols, dim = image.shape
+
+# create pyramid from astronaut image
+astronaut = data.astronaut().mean(axis=2) / 255
+base = np.tile(astronaut, (16, 16))
+pyramid = list(pyramid_gaussian(base, downscale=2, max_layer=5,
+                                multichannel=False))
+pyramid = [(255*p).astype('uint8') for p in pyramid]
```
jni: `skimage.util.img_as_ubyte`?

examples/add_pyramid.py
```diff
@@ -0,0 +1,37 @@
...
+    viewer = napari.Viewer()
+
+    # add image pyramid
+    viewer.add_pyramid(pyramid, clim_range=[0, 255])
+
+    camera = viewer.window.qt_viewer.view.camera
+
+    # Set the view box of the camera to include the whole base image of the
+    # pyramid with a little padding. The view box is a 4-tuple of the x, y
+    # corner position followed by width and height
```
jni, line 31: Why doesn't this happen automatically?

napari/components/_viewer/view/main.py
jni: Nevertheless, does this happen continuously while the zoom level or pan is moving? It still seems like it might be triggered extremely often in those periods? Could this be the source of sluggishness?

I also wonder whether there is a better name for it. `on_camera_update`? At any rate, probably these things could/should be explained in the docstring.

napari/layers/_pyramid_layer/model.py
jni: Yep, `image[[2, 5]]` means (*) "the second and fifth row of `image`, while `image[(2, 5)]` is the same as `image[2, 5]`, ie the element at coordinate 2, 5.

(*): in a future version of NumPy, for consistency with what happens in the current version with `image[array([2, 5])]`.

jni (approved):
I've added some minor comments to the example, otherwise this seems ready imho

napari/layers/_pyramid_layer/model.py
kne42: ah, I forgot that this was used in the base image layer too. yeah, maybe a function in `util` might be worth looking into now that this logic is popping up everywhere

napari/layers/_pyramid_layer/model.py
kne42: huh, I guess that makes sense thinking back to how wonky argument handling is with regards to `__getitem__`:
```python
In [1]: class Foo:
   ...:     def __getitem__(self, key):
   ...:         return key
   ...:

In [2]: f = Foo()

In [3]: f[1, 2]
Out[3]: (1, 2)

In [4]: f[(1, 2)]
Out[4]: (1, 2)
```

kne42 (approved):

examples/add_pyramid.py
sofroniewn: removed

examples/add_pyramid.py
sofroniewn: fixed

examples/add_pyramid.py
sofroniewn: fixed

examples/add_pyramid.py
sofroniewn: fixed

examples/add_pyramid.py
sofroniewn: We can fix this after removing `add_to_viewer` as we'll then be able to place the call to the camera inside `viewer.add_pryamid`

sofroniewn: Actually I'm wary of doing this automatically as there could be times when you add a layer and you don't want the camera to change

sofroniewn:
I have made those changes, and updated from master. I moved the `camera` to be explicitly on the `Viewer` so that the camera update is now a little cleaner. I'll merge once the tests pass.

Issue #296, opened 2019-05-26
API for nD shapes


sofroniewn:
## 🚀 Feature
As discussed in #112 and other places, it will de desirable to extend support for our shapes layer to be `nD`.

One plan is to just support `nD` shapes were each shape is confined to a single (x, y) slice. For example the raw data maybe 4D, but a given shape will reside entirely within one timepoint and one z plane. If we restrict to supporting this case, how should be represent the data? Will we run into problems later when we imagine allowing rapid switching of which dimensions are the displayed ones (i.e. to see the `xz` projection?). Do we want to support a more `nD` like data model but somehow use projections? (it is not clear to me how to do this, but i'd be open to suggestions).

If we go with the above plan what should our API look like. Say you want to have a rectangle on z-slice 0 and a rectangle on z-slice 1. Do you pass
```python
[[[0, y0, x0], [0, y1, x1]],
 [[1, y2, x2],  [1, y3, x3]]]
```
where `[y0, x0]` and  `[y1, x1]` are the two corner points of the rectangle in the zeroth z-slice.
What happens though if someone tries to pass
```python
[[[0, y0, x0], [1, y1, x1]]]
```
Should we reject this rectangle as not supported? Or is now where we use some of our projection magic?

Internally do we want to have a mapping from z-slice -> 2D shape representation that allows us to use the `ShapeList` that we have now unchanged? I'd heavily favor this option is we're only supporting `nD` shapes in the `2D` way described above. If we want to support projections, then maybe it's more complicated.

Do @jni or @d-v-b have any thoughts?

sofroniewn:
This is now implemented in #297 with an example at `examples/nD_shapes.py`

sofroniewn:
This is now merged

PR #297, opened 2019-05-27
Nd shapes

sofroniewn:
# Description
Following on from #296 this PR adds support from nD shapes according to the following API where each shape is specified by an `N x D` array. If all `N` of the first `D-2` values don't agree the shape is just ignored.

Internally the data is represented as a dict of tuples where the `D-2` coords of the tuple are the keys of the `ShapesList` object for each layer. This works fine, but we should probably add some convenience methods to get all shapes back regardless of which slice is being displayed. I can add that in soon.

I've also added `examples/nD_shapes.py` if you want to test this. All other examples should work as before. Here's a new gif:

nD_shapes

## Type of change

# How has this been tested?

- [x] `examples/nD_shapes.py`

## Final checklist:


napari/components/_viewer/model.py
```diff
@@ -265,7 +265,9 @@ def _new_markers(self):
         self.add_markers(empty_markers)
 
     def _new_shapes(self):
-        self.add_shapes([])
+        layer = self.add_shapes([])
+        layer.nd_data_dim = len(self.dims.indices)
+        layer.indices = self.dims.indices
```
jni: I feel like this is more fundamental than something you fiddle with after layer creation. I suggest something like `self.add_shapes([], dimensions=self.ndims)`. If I understand the dims model correctly, the idea is that *all* layers have access to it, either directly or through notifications, and this is how the shapes layer should know its dimensionality and current location.

napari/layers/_shapes_layer/model.py
```diff
@@ -183,7 +183,9 @@ def __init__(self, data, *, shape_type='rectangle', edge_width=1,
         # Freeze refreshes to prevent drawing before the layer is constructed
         with self.freeze_refresh():
             # Add the shape data
-            self.data = ShapeList()
+            self.nd_data = {'2D': ShapeList()}
+            self.nd_data_dim = 2
+            self._data = self.nd_data['2D']
```
jni: As mentioned above, I don't think we should default to 2D and then modify — the dimensionality should be in the constructor.

napari/layers/_shapes_layer/model.py
```diff
@@ -441,9 +445,17 @@ def _get_shape(self):
...
+            slice_shape = [1, 1]
+        else:
+            slice_shape = list(np.max(self.data._vertices, axis=0) + 1)
+
+        if len(self.nd_data.keys()) == 1:
+            return [1] * (self.nd_data_dim-2) + slice_shape
         else:
-            return np.max(self.data._vertices, axis=0) + 1
+            keys = list(self.nd_data.keys())
+            keys.remove('2D')
```
jni: This feels super hacky. The dictionary should only have one type of thing in it.

napari/layers/_shapes_layer/shape_util.py
```diff
@@ -3,6 +3,32 @@
...
+    ----------
+    data : np.ndarray
+        NxD array of vertices specifying the shape.
+
+    Returns
+    ----------
+    key : str | tuple of int | bool
+        If shape is 2D return the string '2D', otherwise return the tuple
+        of int defining the 2D plane if the shape is entirely definied
+        within a single 2D plane. Otherwise return False
```
jni: Why not return an empty tuple for 2D?

napari/layers/_shapes_layer/shape_util.py
```diff
@@ -3,6 +3,32 @@
...
+    value : np.ndarray | False
+        Nx2 array of vertices specifying the shape in the 2D plane or False
+        if no such specification exists.
+    """
+    if data.shape[1] == 2:
+        return '2D', data
+    elif (data[:, :-2] == data[0, :-2]).all():
+        return tuple(data[0, :-2]), data[:, -2:]
+    else:
+        return False, False
```
jni, line 34: I don't know where we are at with the dims model. Does it already encapsulate the concept of which dimensions are drawn? If it does, then this function should directly look at that information, and pick out the right dimensions to look at (rather than assume it's the leading dimension). If it doesn't, then a `# TODO` here would be most useful!

jni: The key could then be a tuple: `(slice_dims, coords)`.

napari/layers/_shapes_layer/shapes/rectangle.py
```diff
@@ -10,7 +10,9 @@ class Rectangle(Shape):
     Parameters
     ----------
     data : np.ndarray
-        Nx2 array of vertices specifying the shape.
+        Either a 2x2 array specifying the two corners of an axis aligned
+        rectangle, or a 4x2 array specifying the four corners of a boudning box
+        that contains the rectangle. These need not be axis aligned.
```
jni: :+1: nice catches in the docstrings

jni (dismissed):
@sofroniewn the functionality is exactly what I wanted! I think the implementation can be cleaned up a little, and I've left comments to that effect.

I think there is probably a "broadcast" mode of operation where people might want to have a 2D shapes layer overlaid on an 3D array. Is this feasible with this codebase? My impression is that it isn't. I'm happy to punt on this, just wanted to flag it to check whether there's any API design choices we can make now that will make it easier in the future.

sofroniewn:
@jni unfortunately "broadcast" mode and "slice" mode don't play nicely with each other in this implementation. It would be pretty trivial to switch between the two of them (we could add a checkbox if that would help), but it would be hard to make it so that you could have both. If you want the checkbox I can add that quickly, but when you're in the one mode you won't see shapes from the other and new shapes will be added to your current mode.

jni:
I'm not too concerned about the checkbox, just the ability to *launch* one or the other. And if I had to choose, for sure I would choose the nD version — it would be easy to slice out the broadcast version of the data after doing nD annotation.

sofroniewn:
@jni I'll make a property (and probably a checkbox too) that lets you switch back and forth.

royerloic:
Very nice comments @jni ! I agree with most if not all.

napari/components/_viewer/model.py
sofroniewn: yes, this should be passed to the constructor. layers shouldn't really be changing their dimensions without actually changing their data

napari/layers/_shapes_layer/model.py
sofroniewn: yes, replacing with empty tuple

napari/layers/_shapes_layer/shape_util.py
sofroniewn: yes, replacing with empty tuple

napari/layers/_shapes_layer/model.py
sofroniewn: done

napari/layers/_shapes_layer/shape_util.py
sofroniewn: agreed, that is not currently supported in the dims model, but I have added a `#TODO` with all this information and we can revisit once it is supported

sofroniewn:
I have made the changes suggested by @jni around `ndims` being a property passed to the constructor of the shapes layer. I have also added a `broadcast` property and checkbox that allows you to switch back and forth between broadcast mode and slice mode. It's pretty cool. You can also copy and paste a shape from broadcast mode to slice mode, but you cannot see shapes in both at the same time as mentioned above. I think this is pretty good for now.

This PR is ready for re-review

sofroniewn:
Ah I just remembered I have to fix `layer.data.to_masks` and `layer.data.to_list()` to make them `nD` I will do this before this is ready to merge

sofroniewn:
This has been updated again to include the nD list, labels, and mask generation. I've also updated the examples to include this. One other thing I'll note is that what used to be called `data` is now `_data_view` which is the currently viewed `ShapeList` object, and `data` is now the dictionary of `ShapeList` objects.

napari/layers/_shapes_layer/model.py
```diff
@@ -259,10 +283,41 @@ def data(self):
...
+        self.events.broadcast()
+        self.refresh()
+
+    @property
+    def nshapes(self):
+        """int: total number of shapes.
+        """
+        nshapes = 0
+        for key, data in self.data.items():
+            nshapes = nshapes + len(data.shapes)
```
jni: I'd refactor this whole function as `nshapes = sum(len(data.shapes) for data in self.data.values())`

I actually don't understand why this isn't just `len(data)`, or rather,

```python
nshapes = sum(map(len, self.data.values()))
```

Unless I'm misreading the code, each element is a ShapeList, which should have its own `__len__` method and return the right number, right?

napari/layers/_shapes_layer/model.py
```diff
@@ -444,22 +499,30 @@ def block_update_properties(self):
...
+            slice_shape = list(np.max(self._data_view._vertices, axis=0) + 1)
+
+        if len(self.data.keys()) == 1:
+            return [1] * (self._input_ndim-2) + slice_shape
         else:
-            return np.max(self.data._vertices, axis=0) + 1
+            keys = list(self.data.keys())
+            keys.remove(())
+            max_val = np.array(keys).max(axis=0)
+            return list(max_val) + slice_shape
```
jni: As a minor comment, it kind of makes me tick a little when a shape is a list and not a tuple. I think everything you've done here can be done with tuples, e.g. `(1,) * (self._input_ndim - 2) + slice_shape`, so I would prefer that. This is very minor though and can be left till later.

napari/layers/_shapes_layer/model.py
```diff
@@ -444,22 +499,30 @@ def block_update_properties(self):
...
         """Determines the shape of the vertex data.
         """
-        if len(self.data._vertices) == 0:
-            return [1, 1]
+        if len(self._data_view._vertices) == 0:
+            slice_shape = [1, 1]
+        else:
+            slice_shape = list(np.max(self._data_view._vertices, axis=0) + 1)
+
+        if len(self.data.keys()) == 1:
```
jni: I suggest adding some comments here for what it means for `len(self.data.keys()) == 1`. It's not immediately obvious what this translates to in terms of the shape of the shapes layer.

napari/layers/_shapes_layer/model.py
```diff
@@ -444,22 +499,30 @@ def block_update_properties(self):
...
+        if len(self._data_view._vertices) == 0:
             maxs = [1, 1]
             mins = [0, 0]
         else:
-            maxs = np.max(self.data._vertices, axis=0) + 1
-            mins = np.min(self.data._vertices, axis=0)
+            maxs = np.max(self._data_view._vertices, axis=0) + 1
+            mins = np.min(self._data_view._vertices, axis=0)
 
         return [(min, max, 1) for min, max in zip(mins, maxs)]
```
jni: If this is only used for slicing, this could be `tuple(slice(min, max) for min, max in zip(mins, maxs))`.

Or if ~~you're~~ I'm really showing off, `tuple(itertools.starmap(slice, zip(mins, maxs)))`. =P

napari/layers/_shapes_layer/model.py
```diff
@@ -510,41 +573,59 @@ def add_shapes(self, data, *, shape_type='rectangle', edge_width=1,
...
+        face_colors = ensure_iterable(face_color, color=True)
+
+        for d, st, ew, ec, fc, o, z, in zip(data, shape_types, edge_widths,
+                                            edge_colors, face_colors,
+                                            opacities, z_indices):
+            shape_cls = self._data_view._types[st]
+
+            # Slice data by 2D plane.
+            key, data_2D = slice_by_plane(d)
+            if key is not False:
```
jni: Maybe use `None` as the value for no keys?

jni: I'm also unsure what `key` means here. Maybe rename to `slice_key`? Then you could say:

```python
if slice_key is not None:  # None implies broadcast mode
```

jni: (is that what it implies? I'm not sure.)

napari/layers/_shapes_layer/model.py
```diff
@@ -1148,7 +1233,142 @@ def to_xml_list(self, shape_type=None):
             List of xml elements defining each shape according to the
             svg specification
         """
-        return self.data.to_xml_list(shape_type=shape_type)
+        return self._data_view.to_xml_list(shape_type=shape_type)
```
jni: After reading a bunch of this code, I am slightly wondering whether `_data_view` should (a) be made public, and (b) be renamed `slice_data`. The latter is more descriptive, whereas "data view" could be anything really. And I also am weary of using *lots* of private attributes and APIs internally, which makes it hard for e.g. plugins to access the same kind of functionality that we do.

napari/layers/_shapes_layer/model.py
```diff
@@ -1148,7 +1233,142 @@ def to_xml_list(self, shape_type=None):
...
+                                             shape_type=shape_type)
+        elif self.broadcast:
+            # For broadcast shapes convert current view to masks and
+            # broadcast across sliced dimensions
+            slices = self._data_view.to_masks(mask_shape=mask_shape[-2:],
+                                              shape_type=shape_type)
+            masks = []
+            for m in slices:
+                masks.append(np.broadcast_to(m, mask_shape))
+            masks = np.stack(masks, axis=0)
```
jni: Is this functionality tested anywhere? It seems pretty brittle to me...

napari/layers/_shapes_layer/model.py
```diff
@@ -1148,7 +1233,142 @@ def to_xml_list(self, shape_type=None):
...
+        else:
+            # For nD insert each keyed slice into correctd place in volume
```
jni:
```suggestion
            # For nD insert each keyed slice into correct place in volume
```

napari/layers/_shapes_layer/model.py
```diff
@@ -1148,7 +1233,142 @@ def to_xml_list(self, shape_type=None):
...
+            masks = []
+            for key, data in self.data.items():
+                if len(key) > 0:
+                    slices = data.to_masks(mask_shape=mask_shape[-2:],
+                                           shape_type=shape_type)
+                    for m in slices:
+                        vol = np.zeros(mask_shape)
+                        vol[key] = m
+                        masks.append(vol)
+            masks = np.stack(masks, axis=0)
```
jni, line 1389: This line is repeated across both conditions, so it can be pulled out.

napari/layers/_shapes_layer/model.py
```diff
@@ -1148,7 +1233,142 @@ def to_xml_list(self, shape_type=None):
...
+            labels = self._data_view.to_labels(labels_shape=labels_shape,
+                                               shape_type=shape_type)
+        elif self.broadcast:
+            # For broadcast shapes convert current view to labels and
+            # broadcast across sliced dimensions
+            labels = self._data_view.to_labels(labels_shape=labels_shape[-2:],
+                                               shape_type=shape_type)
+            labels = np.broadcast_to(labels, labels_shape)
+        else:
+            # For nD insert each keyed slice into correctd place in volume
```
jni:
```suggestion
            # For nD insert each keyed slice into correct place in volume
```

napari/layers/_shapes_layer/model.py
```diff
@@ -1148,7 +1233,142 @@ def to_xml_list(self, shape_type=None):
...
+            # and increment integer label of shape
+            labels = np.zeros(labels_shape)
+            nshapes = 0
+            for key, data in self.data.items():
+                if len(key) > 0:
+                    slices = data.to_labels(labels_shape=labels_shape[-2:],
+                                            shape_type=shape_type)
+                    slices[slices > 0] = slices[slices > 0] + nshapes
```
jni: `+=` ?

napari/layers/_shapes_layer/model.py
```diff
@@ -1148,7 +1233,142 @@ def to_xml_list(self, shape_type=None):
...
+
+        Returns
+        ----------
+        data : list
+            List of shape data where each element of the list is an
+            `np.ndarray` corresponding to one shape
+        """
+        if self.ndim == 2:
+            # For 2D shapes just convert current view to a list
+            data = self._data_view.to_list(shape_type=shape_type)
```
jni: I'm kinda surprised that 2D requires a special case here... Doesn't the nd code with the `()` key just work?

napari/layers/_shapes_layer/shape_util.py
```diff
@@ -3,6 +3,36 @@
...
+    # TODO make which coordinate are the displayed 2D coordinates changable
+    # instead of assuming they are the last two dimensions. This feature
+    # will be implemented here after it is implemented on the main dims model.
+    # The key can then be a tuple: (slice_dims, coords)
+    if data.shape[1] == 2:
+        return (), data
+    elif (data[:, :-2] == data[0, :-2]).all():
+        return tuple(data[0, :-2].astype('int')), data[:, -2:]
+    else:
+        return False, False
```
jni, line 34: Right, sorry I misunderstood `False` above. It's worth writing in a comment where this function is used, because it's really not intuitive. `# False means invalid shape`

jni (commented):
@sofroniewn I've made a few more comments, mostly minor. The biggest one is that, as far as I can tell, the example doesn't exercise all the lines of code that it could. An example with broadcast mode shapes would be very useful to make sure that everything works (both now and as we refactor this code). 

I also wonder why 2D needs to be a special case — I think it *should* just work with the same code as nD. If it doesn't, it's worth exploring in detail why/where.

napari/layers/_shapes_layer/model.py
sofroniewn: Ah so a `ShapeList` isn't actually a `List`, if this is confusing and misleading I can try fixing with help from @kne42. It contains a property `shapes` which is the `List`. We can fix this, but it should be a followup PR

sofroniewn: I have replace this with `nshapes = sum(len(data.shapes) for data in self.data.values())` for now

napari/layers/_shapes_layer/model.py
sofroniewn: changed to tuple

napari/layers/_shapes_layer/model.py
sofroniewn: added

napari/layers/_shapes_layer/model.py
sofroniewn: I switched this to be a tuple, but the change you suggested caused errors in other places

napari/layers/_shapes_layer/model.py
sofroniewn: renamed `slice_key` and added a note about `False` being invalid

napari/layers/_shapes_layer/model.py
sofroniewn: done

napari/layers/_shapes_layer/model.py
sofroniewn: fixed

napari/layers/_shapes_layer/model.py
sofroniewn: fixed

napari/layers/_shapes_layer/model.py
sofroniewn: fixed

napari/layers/_shapes_layer/shape_util.py
sofroniewn: added notes

napari/layers/_shapes_layer/model.py
sofroniewn: this is tested in the examples, gives outputs of the right shape. What's the concern?

napari/layers/_shapes_layer/model.py
sofroniewn: done

napari/layers/_shapes_layer/model.py
kne42: probably to do with the usage of `broadcast_to` on a `slice`

also the above statement should be a list comprehension

napari/layers/_shapes_layer/model.py
sofroniewn: I've reduced one less if statement here, but lumped the 2D case in with the broadcast code as it is simpler to understand I think and less complicated

napari/layers/_shapes_layer/model.py
sofroniewn: ah found an error here with no masks, will fix

sofroniewn: i've now fixed this I think

sofroniewn:
I have added an example of broadcasting to the `examples/nD_shapes.py` which should increase the code coverage of our tests. This is now ready for re-review

sofroniewn:
This PR has been updated with the changes from black and is ready for re-review. I'll not that already in the meeting @kne42 and @kevinyamauchi raised the point that they find switching back and forth between "slice" and "broadcast" mode confusing and maybe that should be a setting that cannot be changed once made. Unfortunately it's a little harder to allow the user to set this from the gui when creating a new layer so I'm not sure what to do

napari/layers/_shapes_layer/shape_util.py
```diff
@@ -3,6 +3,37 @@
 from vispy.geometry import PolygonData
 
 
+def slice_by_plane(data):
+    """Determine if shape is entirely definied within a single 2D plane
```
kne42: typo

napari/layers/_shapes_layer/shape_util.py
```diff
@@ -3,6 +3,37 @@
...
+
+    Parameters
+    ----------
+    data : np.ndarray
+        NxD array of vertices specifying the shape.
```
kne42: shape should go in type

napari/layers/_shapes_layer/shape_util.py
```diff
@@ -3,6 +3,37 @@
...
+
+    Returns
+    ----------
+    key : tuple of int | bool
```
kne42: use `or`, not `|`

napari/layers/_shapes_layer/shape_util.py
```diff
@@ -3,6 +3,37 @@
...
+        If shape is 2D returns an empty tuple, otherwise return the tuple
+        of int defining the 2D plane if the shape is entirely definied
+        within a single 2D plane. Otherwise return False as the shape is
+        invalid.
+    value : np.ndarray | False
+        Nx2 array of vertices specifying the shape in the 2D plane or False
```
kne42: put shape in type

napari/layers/_shapes_layer/shapes/ellipse.py
```diff
@@ -61,7 +61,7 @@ def __init__(
 
     @property
     def data(self):
-        """np.ndarray: Nx2 array of vertices.
+        """np.ndarray: 4x2 array of vertices.
```
kne42: shape in type

napari/layers/_shapes_layer/shapes/line.py
```diff
@@ -10,7 +10,7 @@ class Line(Shape):
     Parameters
     ----------
     data : np.ndarray
-        Nx2 array of vertices specifying the shape.
+        2x2 array of vertices specifying the shape.
```
kne42: same

napari/layers/_shapes_layer/shapes/line.py
```diff
@@ -51,7 +51,7 @@ def __init__(
 
     @property
     def data(self):
-        """np.ndarray: Nx2 array of vertices.
+        """np.ndarray: 2x2 array of vertices.
```
kne42: same

napari/layers/_shapes_layer/shapes/rectangle.py
```diff
@@ -10,7 +10,9 @@ class Rectangle(Shape):
     Parameters
     ----------
     data : np.ndarray
-        Nx2 array of vertices specifying the shape.
+        Either a 2x2 array specifying the two corners of an axis aligned
```
kne42: same

napari/layers/_shapes_layer/shapes/rectangle.py
```diff
@@ -53,7 +55,7 @@ def __init__(
 
     @property
     def data(self):
-        """np.ndarray: Nx2 array of vertices.
+        """np.ndarray: 4x2 array of vertices.
```
kne42: same

napari/layers/_shapes_layer/view/properties.py
```diff
@@ -82,6 +94,12 @@ def changeEdgeColor(self, text):
     def changeWidth(self, value):
         self.layer.edge_width = float(value) / 2
 
+    def change_broadcast(self, state):
+        if state == Qt.Checked:
```
kne42: `self.layer.broadcast = state == Qt.Checked`

napari/layers/_shapes_layer/shape_util.py
sofroniewn: fixed

napari/layers/_shapes_layer/shape_util.py
sofroniewn: should it be `value : np.ndarray or False, Nx2` or `value : np.ndarray or False, shape(N,2)` or something else?

napari/layers/_shapes_layer/shapes/ellipse.py
sofroniewn: so have no shape here, only `np.ndarray: array of vertices` and then shape up top 
```
Parameters
    ----------
    data : np.ndarray, 2x2 or 4x2
```
?

napari/layers/_shapes_layer/shape_util.py
kne42: `value : (N, 2) array or False`

napari/layers/_shapes_layer/view/properties.py
sofroniewn: fixed

napari/layers/_shapes_layer/shape_util.py
sofroniewn: perfect! thanks for the help. I need the autoformatter for our docs!

kne42: actually is `value` even used at all there? it doesn't seem to be

if it's used as a default when it's not passed, however, it shouldn't be defaulted to `False`, but rather `None`, in which case the type would be `value : (N, 2) array, optional`

napari/layers/_shapes_layer/shapes/ellipse.py
kne42: `4x2` is a shape and it should be formatted as `(4, 2) array: ellipse vertices`

napari/layers/_shapes_layer/shape_util.py
sofroniewn: `value` is in the return. It could return as `None` if that was better

napari/layers/_shapes_layer/shape_util.py
```diff
@@ -3,6 +3,37 @@
...
+        if no such specification exists.
+    """
+    # TODO make which coordinate are the displayed 2D coordinates changable
+    # instead of assuming they are the last two dimensions. This feature
+    # will be implemented here after it is implemented on the main dims model.
+    # The key can then be a tuple: (slice_dims, coords)
+    if data.shape[1] == 2:
+        return (), data
+    elif (data[:, :-2] == data[0, :-2]).all():
```
kne42: it's cleaner to use `np.all` instead of the method approach

napari/layers/_shapes_layer/shape_util.py
sofroniewn: fixed

napari/layers/_shapes_layer/shape_util.py
sofroniewn: fixed

napari/layers/_shapes_layer/shape_util.py
kne42: oh my bad didn't catch that; however `None` is preferable when something exists

although now looking at the other statement, why are you returning two `False`s instead of just raising an exception?

sofroniewn: I guess I was letting the thing doing the calling decide what to do if no such specification existed (which was to just ignore it) - see in the shapes model file for it being used

napari/layers/_shapes_layer/shapes/ellipse.py
sofroniewn: fixed

napari/layers/_shapes_layer/shapes/line.py
sofroniewn: fixed

napari/layers/_shapes_layer/shapes/line.py
sofroniewn: fixed

napari/layers/_shapes_layer/shapes/rectangle.py
sofroniewn: fixed

napari/layers/_shapes_layer/shapes/rectangle.py
sofroniewn: fixed

napari/layers/_shapes_layer/shape_util.py
sofroniewn: done

sofroniewn:
I have fixed some doc strings at the request of @kne42. I started fixing more of them, but in functions not originally touched by this PR, but there are just so many that need to be fixed. We should instead have a followup PR to fix them all, but I agree that we shouldn't be introducing more errors in this PR!

examples/nD_shapes.py
```diff
@@ -0,0 +1,73 @@
...
+        face_color=colors,
+        name='sliced',
+    )
+
+    masks = layer.to_masks(mask_shape=(128, 128, 128))
+    labels = layer.to_labels(labels_shape=(128, 128, 128))
+    shape_array = np.array(layer.to_list())
+
+    print(
+        'sliced nshapes',
```
kne42: given how `black` is formatting this, it would probably be cleaner if you used f-strings

examples/nD_shapes.py
```diff
@@ -0,0 +1,73 @@
...
+        broadcast=True,
+        name='broadcasted',
+    )
+
+    masks = layer.to_masks(mask_shape=(128, 128, 128))
+    labels = layer.to_labels(labels_shape=(128, 128, 128))
+    shape_array = np.array(layer.to_list())
+
+    print(
+        'broadcast nshapes',
```
kne42: ditto

napari/layers/_shapes_layer/shape_util.py
```diff
@@ -3,19 +3,49 @@
...
+        if no such specification exists.
+    """
+    # TODO make which coordinate are the displayed 2D coordinates changable
+    # instead of assuming they are the last two dimensions. This feature
+    # will be implemented here after it is implemented on the main dims model.
+    # The key can then be a tuple: (slice_dims, coords)
+    if data.shape[1] == 2:
+        return (), data
+    elif np.all(data[:, :-2] == data[0, :-2]):
+        return tuple(data[0, :-2].astype('int')), data[:, -2:]
```
kne42, line 32: use `np.int` instead

napari/layers/_shapes_layer/shape_util.py
```diff
@@ -36,13 +66,13 @@ def inside_boxes(boxes):
...
-        Nx8x2 array of N boxes that should be checked
+    boxes : (N, 8, 2) array
+        Array of N boxes that should be checked
 
     Returns
     -------
-    inside : np.ndarray
-        Length N boolean array with `True` values for boxes containg the origin
+    inside : (N,) array of bool
+        Array with `True` values for boxes containg the origin
```
kne42: I would take a page out of scikit-image's book:
```suggestion
        True if corresponding box contains the origin.
```

napari/layers/_shapes_layer/shapes/rectangle.py
```diff
@@ -9,8 +9,10 @@ class Rectangle(Shape):
 
     Parameters
     ----------
-    data : np.ndarray
-        Nx2 array of vertices specifying the shape.
+    data : (4, 2) array
+        Either a (2, 2) array specifying the two corners of an axis aligned
```
kne42, line 13: if you can have two different shapes, you should specify two different shapes in the type

napari/layers/_shapes_layer/shapes/rectangle.py
```diff
@@ -9,8 +9,10 @@ class Rectangle(Shape):
...
+        rectangle, or a (4, 2) array specifying the four corners of a boudning
```
kne42: typo: bounding

kne42 (approved):
approved pending changes

examples/nD_shapes.py
```diff
@@ -0,0 +1,73 @@
...
+    viewer = napari.view(blobs.astype(float))
+
+    # Create random rectangles
+    shapes = [
+        [
+            [i] + list(128 * np.random.random(2)),
+            [i] + list(128 * np.random.random(2)),
+        ]
+        for i in range(128)
+    ]
```
jni: Man, I'm afraid `black` has just switched one source of nitpicking for another: when black does something super-unpythonic like this, I just want to rework the whole statement to something that black will tolerate better! :joy:

Let me know if you like this, or some variant, happy to ignore it:

```python
# create one random rectangle per "plane"
planes = np.arange(128).reshape((128, 1))
corners0 = np.random.uniform(0, 128, size=(128, 2))
corners1 = np.random.uniform(0, 128, size=(128, 2))
pl_corners0 = np.concatenate((planes, corners0), axis=1)
pl_corners1 = np.concatenate((planes, corners1), axis=1)
shapes = np.stack((pl_corners0, pl_corners1), axis=0)
```

It's more wordy, but I like that it specifies what the expected format is.

examples/nD_shapes.py
```diff
@@ -0,0 +1,73 @@
...
+
+    base_cols = ['red', 'green', 'blue', 'white', 'yellow', 'magenta', 'cyan']
+    colors = [np.random.choice(base_cols) for i in range(128)]
```
jni:
```suggestion
    colors = np.random.choice(base_cols, size=128)
```

examples/nD_shapes.py
```diff
@@ -0,0 +1,73 @@
...
+
+    layer = viewer.add_shapes(
+        np.array(shapes),
+        shape_type='rectangle',
+        face_color=colors,
+        name='sliced',
+    )
+
+    masks = layer.to_masks(mask_shape=(128, 128, 128))
```
jni, line 33: Sorry, I've forgotten what `to_masks` does???

examples/nD_shapes.py
```diff
@@ -0,0 +1,73 @@
...
+        shape_array.shape,
+    )
+
+    layer = viewer.add_shapes(
+        np.array(shapes[0])[:, 1:],
+        shape_type='rectangle',
+        ndim=3,
+        broadcast=True,
+        name='broadcasted',
+    )
```
jni, line 49: Given others' justified confusion about broadcast and nd in the same layer, how about just making this two separate examples, and generally discouraging the mixing of the two styles?

napari/layers/_shapes_layer/model.py
```diff
@@ -277,10 +305,39 @@ def data(self):
...
     def data(self, data):
+        if self._data == data:
+            return
         self._data = data
         self.events.data()
         self.refresh()
 
+    @property
+    def broadcast(self):
+        """Bool: if shapes are broadcast across all dimensions.
```
jni: I'm quite annoyed that Black doesn't move the triple quotes up, as PEP257 specifies for one-line docstrings...

napari/layers/_shapes_layer/model.py
```diff
@@ -478,24 +543,34 @@ def block_update_properties(self):
...
-            return [1, 1]
+        if len(self.slice_data._vertices) == 0:
+            slice_shape = tuple([1, 1])
+        else:
+            slice_shape = tuple(np.max(self.slice_data._vertices, axis=0) + 1)
+
+        if len(self.data.keys()) == 1:  # or self.broadcast is True:
+            # If in broadcast mode or only broadcast shapes are present
+            # return the shape padded to the necessary dimensions
+            return (1,) * (self._input_ndim - 2) + slice_shape
```
jni, line 546: Nice. =)

napari/layers/_shapes_layer/model.py
```diff
@@ -552,60 +627,74 @@ def add_shapes(
...
+            # If a single array for a shape has been passed turn into list
+            data = [data]
+
+        # Turn input arguments into iterables
+        shape_types = ensure_iterable(shape_type)
+        edge_widths = ensure_iterable(edge_width)
+        opacities = ensure_iterable(opacity)
+        z_indices = ensure_iterable(z_index)
+        edge_colors = ensure_iterable(edge_color, color=True)
+        face_colors = ensure_iterable(face_color, color=True)
```
jni: Another black-related nitpick: since the iterables are all being placed on their own line in the for-loop, maybe just move these all to the zip()?

napari/layers/_shapes_layer/model.py
```diff
@@ -552,60 +627,74 @@ def add_shapes(
...
+
+        for d, st, ew, ec, fc, o, z in zip(
+            data,
+            shape_types,
+            edge_widths,
+            edge_colors,
+            face_colors,
+            opacities,
+            z_indices,
```
jni: This is killing me... I feel very strongly that the statement of the for-loop should not be 4-indented, so that it is not confused for the body. Considering forking black into `grey`. =P

napari/layers/_shapes_layer/model.py
```diff
@@ -552,60 +627,74 @@ def add_shapes(
...
+                    if not self.slice_data == self.data[slice_key]:
+                        self.slice_data = self.data[slice_key]
+                        # self._finish_drawing()
+                else:
+                    slice_key = self.indices[:-2]
+                    if slice_key not in self.data:
+                        self.data[slice_key] = ShapeList()
+                    if not self.slice_data == self.data[slice_key]:
+                        self.slice_data = self.data[slice_key]
+                        # self._finish_drawing()
```
jni: Either remove these or add a comment about why they are there

napari/layers/_shapes_layer/model.py
```diff
@@ -1251,7 +1348,141 @@ def to_xml_list(self, shape_type=None):
             List of xml elements defining each shape according to the
             svg specification
         """
-        return self.data.to_xml_list(shape_type=shape_type)
+        return self.slice_data.to_xml_list(shape_type=shape_type)
+
+    def to_masks(self, mask_shape=None, shape_type=None):
```
jni, line 1342: I'm not in love with this function... Perhaps we can leave it for a later PR?

napari/layers/_shapes_layer/model.py
```diff
@@ -1251,7 +1348,141 @@ def to_xml_list(self, shape_type=None):
...
+                    slices[slices > 0] += nshapes
+                    labels[slice_key] = slices
+                    nshapes += len(data.shapes)
+        return labels
+
+    def to_list(self, shape_type=None):
+        """Returns the vertex data assoicated with the shapes as a list
+        where each element of the list corresponds to one shape. Passing a
+        `shape_type` argument leads to only that particular `shape_type`
+        being returned.
```
jni: please follow PEP257 format, which is a single line summary (present imperative, "Do this" rather than "Does this"), followed by a blank line, followed by longer summary optionally

napari/layers/_shapes_layer/shape_util.py
jni: no, just `int`, for which `np.int` is an alias.

jni (commented):
@sofroniewn I'm pretty happy with this, excepting the comments below, some of which are minor or optional.

examples/nD_shapes.py
sofroniewn: improved

examples/nD_shapes.py
sofroniewn: improved

napari/layers/_shapes_layer/shape_util.py
sofroniewn: ok sticking with `int`

napari/layers/_shapes_layer/shape_util.py
sofroniewn: fixed

napari/layers/_shapes_layer/shapes/rectangle.py
sofroniewn: done

napari/layers/_shapes_layer/shapes/rectangle.py
sofroniewn: fixed

examples/nD_shapes.py
sofroniewn: simplified a little, but fixed to make it more clear

examples/nD_shapes.py
sofroniewn: fixed

examples/nD_shapes.py
sofroniewn: For a layer with N shapes, `to_masks` generates an NxDims array where each of the shape is its own binary mask of shape image Dims

examples/nD_shapes.py
sofroniewn: I feel like here they are already separated into two different layers (which makes sense to me), but having them in the same layer would be too much. I'd rather leave as is unless strong desires to make an entirely new example. I think it's pretty reasonable that people would want to use sliced and broadcasted together in separate layers with the same underlying image data

napari/layers/_shapes_layer/model.py
sofroniewn: fixed. could be because we have some of the string stuff disabled

napari/layers/_shapes_layer/model.py
sofroniewn: fixed

napari/layers/_shapes_layer/model.py
sofroniewn: fixed by making the zip output a variable on the lines above

napari/layers/_shapes_layer/model.py
sofroniewn: These should actually not be commented, uncommented and added a comment on what they are doing

napari/layers/_shapes_layer/model.py
sofroniewn: Is that because of the special casing for the `dims=2`? I'm sure we could remove that, but I'd really much rather see the whole function go in now and we refactor with a followup PR then have to pull it out and maintain it somewhere else in the interim

napari/layers/_shapes_layer/model.py
sofroniewn: fixed

sofroniewn:
I have made most of the changes requested by @jni and @kne42. There are maybe one or two things around the broadcast / sliced or the `to_masks` that can be improved, but I'd like to merge this once the tests pass as I have a couple other PRs in mind (like thumbnails for the shapes layer) that are waiting on this one

PR #303, opened 2019-05-28
WIP: Stop using add_to_viewer syntax for basic layers

jni:
I've kept the register module for now because I think it is probably useful for plugins.

Having worked with this more closely now, I see more clearly all the advantages of @kne42's `add_to_viewer` decorator. Specifically, neither copying the full docstring and signature of the class (thus resulting in duplication) nor setting it dynamically (which makes it unusable by IDEs) is desirable. I'm thus open to the following three-part alternative to this PR.

1. keep `add_to_viewer` *but* add type-accurate stub functions that get overwritten by `add_to_viewer`
2. make the `name` keyword argument of `add_to_viewer` a required argument.
3. add an `overwrite=False` keyword argument to `add_to_viewer`, which will result in a `ValueError` if the specified name is already taken.

Point 1 will address the problem of IDEs (though it does not solve the problem of duplicated function signature/docstrings), while 2-3 will address the problem of namespace clashes.

Very open to discussion here!

## Type of change

# References

# How has this been tested?

## Final checklist:


napari/layers/_register.py
```diff
@@ -134,6 +132,7 @@ def create_func(cls, name=None, doc=None):
 
 
 def _register(cls, *, name=None, doc=None):
+    from ..components import ViewerModel
```
sofroniewn, line 137: Why did you move the import statement in here?

sofroniewn:
If we are no longer using the `_register` model explicitly for the layers, but you want to keep it around, do you want to put it in the `components/_viewer` folder? Or maybe the util folder? it's a little unclear to me where it should live as it imports `ViewerModel`

sofroniewn:
I'm also open to the alternative you propose - in terms of guiding principles though, I'd really like to be able to instantiate all of the layers independent of the `ViewerModel` for testing purposes. Would that still be possible if we took the alternative approach? Currently on master it is not as each layer has
`from .._register import add_to_viewer`  and `_register.py` has `from ..components import ViewerModel`. Aside from being a problem for testing this can result in problems with circular imports, which is why `ViewerModel` imports `LayersList` inside it's constructor. After this refactor I'd really like to see that import move to the top of `ViewerModel.py`

kne42:
Ideally we shouldn't have to do anything; this is PyCharm's fault for statically providing completions and documentation in a dynamic language 🙄.

I feel like every situation here is a loss, so I'll just go with whatever is best for the user and I think that's this PR, but with the signatures/docstrings properly copied from their respective classes, not the whole "please see X documentation."

Addressing @sofroniewn's concerns, this PR actually fixes the circular references so that the layer list import can be moved back up top (see the line you commented on). You can already instantiate the layers without the `ViewerModel` and it is unavoidable to import it since it is in the top-level `__init__.py`.

jni:
Ok, this is ready for re-review. Many thanks to @kne42 for the test suggestion!

sofroniewn (approved):
Looks good to me, can you add the extra blank line to the test (pep) and can you move `_register.py` to the `utils` folder and make sure all the paths are still correct. Then I'd say this is good to merge if @kne42 approves too

napari/components/_viewer/model.py
```diff
@@ -293,6 +293,230 @@ def add_layer(self, layer):
...
+        -------
+        layer : :class:`napari.layers.Image`
+            The newly-created image layer.
+        """
+        layer = layers.Image(image, *args, **kwargs)
+        self.add_layer(layer)
+        return layer
+
+    def add_pyramid(self, pyramid, *args, **kwargs):
+        """Add image pyramid layer.
```
kne42: this looks different from the other summaries

napari/tests/test_class_docstrings.py
```diff
@@ -0,0 +1,27 @@
...
+)
+from napari import Viewer
+
+def test_docstrings():
+    for obj in (Image, Pyramid, Markers, Labels, Shapes, Vectors):
+        method = f'add_{obj.__name__.lower()}'
+        obj_doc = obj.__doc__
+        i = [idx for idx, section in enumerate(obj.__doc__.split('\n\n'))
+             if section.strip().startswith('Parameters')][0]
+        # We only check the parameters section, which is the first
```
kne42, line 14: since the summary and return sections seem to follow a pattern, do we want to test these as well?

kne42 (approved):
lgtm

napari/components/_viewer/model.py
sofroniewn: what do you mean here?

napari/tests/test_class_docstrings.py
sofroniewn: I think this is fine for now as not all Layers docs have all of these, but in the future, we might standardize all of them so that they all have an `Attributes` and `Extended Summary` too at which point I agree. But if easy to add now and it doesn't hurt can do it now as well

napari/components/_viewer/model.py
kne42: compare them and you'll see – they all follow a pattern/template that this does not

sofroniewn: I feel like this one is almost identical to the `Image` summary, apologies for completely missing this, but I'm not seeing it. Can you help me out

sofroniewn: ah I see it!

sofroniewn: oh wait no i don't :-(

sofroniewn: do you mean that that line should be
`Add a pyramid layer to the layers list`?
If so I'm happy to fix!

napari/tests/test_class_docstrings.py
sofroniewn: after speaking with @kne42 she made it clear that this was more about testing conformity of the summary and return sections in the docs in the viewer and not about the consistency between the docs in the viewer and the docs on the layers, so we said we'd wait to address in a follow-up PR

napari/components/_viewer/model.py
sofroniewn: done

PR #306, opened 2019-05-28
fix click on layer list

### LINKED ISSUE #304: error when clicking in layers list
## 🐛 Bug

program crashes when i click in the layers list

## To Reproduce

1. open any example
2. click in the layers list

```python
  File "/Users/kiraevans/Projects/napari/napari/components/_layers_list/view/main.py", line 115, in mouseMoveEvent
    distance = (event.pos() - self.drag_start_position).manhattanLength()
TypeError: unsupported operand type(s) for -: 'QPoint' and 'tuple'
```

## Expected behavior

it doesnt crash

## Environment

 - OS (e.g., Linux): OSX
 - Python version: 3.7

`conda list | grep qt` output:
```
pyqt                      5.9.2            py37h655552a_2
pyqt5                     5.12.1                   pypi_0    pypi
pyqt5-sip                 4.19.15                  pypi_0    pypi
qt                        5.9.7                h468cd18_1
qtawesome                 0.5.7                    pypi_0    pypi
qtconsole                 4.4.3                    pypi_0    pypi
qtpy                      1.7.0                    pypi_0    pypi
sphinxcontrib-qthelp      1.0.2                    pypi_0    pypi
```

sofroniewn:
# Description
This PR fixes #304 (hopefully). It's a bit hard for me to know as I don't have this problem on my machine, but I think this may do it. Maybe @kne42 can test

## Type of change

# How has this been tested?

- [x] run `example/add_image.py` and click on layer list

## Final checklist:


kne42:
new error:
```python
  File "/Users/kiraevans/Projects/napari/napari/components/_layers_list/view/main.py", line 151, in dragMoveEvent
    index = self.layers.index(layer_name)
  File "/Users/kiraevans/Projects/napari/napari/util/list/_typed.py", line 76, in index
KeyError: 'could not find element  was referencing'
```

sofroniewn:
@kne42 this should be fixed now, can you check again, thanks!

kne42:
still broken

kne42:
ok so i investigated and the error is different now – due to a bug in the implementation of custom lists

however, if you look more closely at the error message, you'll notice that the layer name is nothing, just an empty string

this is because your fix does not address the root of the problem, which is that this code is getting called at all despite the fact that I am not clicking on any layers. i am clicking in an area inside the layer list but it is empty space, so the related code should not be run at all

sofroniewn:
I was finally able to reproduce this bug, it required initiating a drag outside the layer properties not just clicking. I have now fixed the root cause by preventing the drag from initiating in this situation. This is ready for re-review from @kne42

napari/components/_layers_list/view/main.py
```diff
@@ -148,16 +152,17 @@ def dragMoveEvent(self, event):
...
-        # If the widget being dragged hasn't moved above or below any other
-        # widgets then don't highlight any dividers
-        selected = (not (insert == index) and not (insert-1 == index))
-        # Set the selected state of all the dividers
-        for i in range(0, self.vbox_layout.count(), 2):
-            if i == 2*divider_index:
-                self.vbox_layout.itemAt(i).widget().setSelected(selected)
-            else:
-                self.vbox_layout.itemAt(i).widget().setSelected(False)
+        if layer_name in self.layers:
```
kne42: this should be removed – we should always check if what's being dragged is actually a layer, in which case, its name should _always_ be in the layers list anyways

kne42 (approved):
looks great! :)

napari/components/_layers_list/view/main.py
sofroniewn: done

Issue #265, opened 2019-05-16
Unify state variable type and interface


kevinyamauchi:
## Overview
Currently, layer properties with categorical values (e.g., layer mode, blending mode) are stored and set in different ways both within layers and across layers. I think we should unify this both for consistency now, but also looking forward I think it will be helpful for recording and resuming viewer state (e.g., #234, #199).

## Proposed approach
Mirroring our approach in #233, I propose that we use Enums internally to store the state and provide a string-based interface for the users. I think this can be broken up into a couple of PRs:

- [x] Change layer modes to Enum and add string-based interface (#311)

    - `Labels`: Change the `Labels.mode` setter and property to  accept/return strings.
    - `Markers`: Change `Markers.mode` to an Enum. Keep the string-based property and setter.
    - `Shapes`: Change the `Shapes.mode` setter and property to  accept/return strings.

- [x] Update the base layer blending mode variable

- [ ] Pare down the `Image` layer color maps and use an Enum with a string-based setter/property as with the others (I suspect this should be its own issue).

- [ ] For each layer, change the remaining layer properties with categorical values to Enum with a string-based setter/property.
    - `Image`: interpolation
    - `Markers`: face color, edge color
    - `Vectors`:  color
    - `Shapes`: face color, edge color

sofroniewn:
This sounds great @kevinyamauchi - I really like the decomposition into a series of PRs

jni:
@kevinyamauchi I'm super into this!

One suggestion I had in the meeting is that it might be worth creating our own `StringyEnum` object that supports this kind of setting of values natively, so that you don't have to repeat the same kind of code everywhere. I'm not sure how feasible this is though. I think you still need to manually set the properties on each object that uses these enums.

kne42:
> it might be worth creating our own `StringyEnum` object that supports this kind of setting of values natively

They have a copy-pastable example of this in the [Python documentation](https://docs.python.org/3/library/enum.html#using-automatic-values).

kevinyamauchi:
Thanks for the feedback, @jni and @kne42 ! I need to read more about the autonaming, but that seems like a good approach.

sofroniewn:
@kevinyamauchi another place that I think will need the Enum refactor are all the uses of `shape_type` in the shapes layer (like 'rectangle', 'ellipse', etc.) which then refer to shape classes. happy to discuss thursday

sofroniewn:
[...]

PR #311, opened 2019-05-30
Unify layer mode Enums

kevinyamauchi:
# Description
As discussed in #265, this PR unifies the layer mode variables. All layer mode variables are now Enums. The mode property can now be set with a string and returns a string. I made the following changes:
- Created a `StringEnum` class that automatically assigns members the string of their name as a value per the [example recommended by @kne42](https://docs.python.org/3/library/enum.html#using-automatic-values) .
- `Labels`: Change the `Labels.mode` setter and property to  accept/return strings.
- `Markers`: Change `Markers.mode` to an Enum. Keep the string-based property and setter.
- `Shapes`: Change the `Shapes.mode` setter and property to  accept/return strings.

## Type of change

# References
This is the first PR in #265 .

# How has this been tested?
- [x] labels: ran `labels-2d.py` and added mode setting with a string.
- [x] shapes: ran `add_shapes.py` and added mode setting with a string.
- [x] markers: ran `add_markers.py` and added mode setting with a string.

## Final checklist:

### DIFF
--- a/napari/layers/_labels_layer/_constants.py
+++ b/napari/layers/_labels_layer/_constants.py
@@ -1,8 +1,10 @@
-from enum import Enum
+from enum import auto
 import sys
 
+from ...util.misc import StringEnum
 
-class Mode(Enum):
+
+class Mode(StringEnum):
     """MODE: Interactive mode. The normal, default mode is PAN_ZOOM, which
     allows for normal interactivity with the canvas.
 
@@ -22,10 +24,10 @@ class Mode(Enum):
     pixel. If the background label `0` is selected than any pixels will be
     changed to background and this tool functions like an eraser.
     """
-    PAN_ZOOM = 0
-    PICKER = 1
-    PAINT = 2
-    FILL = 3
+    PAN_ZOOM = auto()
+    PICKER = auto()
+    PAINT = auto()
+    FILL = auto()
 
 
 BACKSPACE = 'delete' if sys.platform == 'darwin' else 'backspace'
--- a/napari/util/misc.py
+++ b/napari/util/misc.py
@@ -47,7 +47,7 @@ def is_iterable(arg, color=False):
     elif np.isscalar(arg):
         return False
     elif color and isinstance(arg, (list, np.ndarray)):
-        if np.array(arg).ndim == 1 and (len(arg) == 3 or len(arg)==4):
+        if np.array(arg).ndim == 1 and (len(arg) == 3 or len(arg) == 4):
             return False
         else:
             return True
@@ -218,6 +218,13 @@ class StringEnum(Enum):
         """ autonaming function assigns each value its own name as a value
         """
         return name
+
+    @classmethod
+    def _missing_(cls, value):
+        """ function called with provided value does not match any of the class
+           member values. This function tries again with an upper case string.
+        """
+        return cls(value.upper())
 
     def __str__(self):
         """String representation: The string method returns the
--- a/napari/util/misc.py
+++ b/napari/util/misc.py
@@ -212,11 +212,13 @@ def segment_normal_vector(a, b):
         unit_norm = normal/norm
     return unit_norm
 
+
 class StringEnum(Enum):
     def _generate_next_value_(name, start, count, last_values):
         """ autonaming function assigns each value its own name as a value
         """
         return name
+
     def __str__(self):
         """String representation: The string method returns the
         valid vispy symbol string for the Markers visual.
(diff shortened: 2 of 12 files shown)


napari/layers/_labels_layer/_constants.py
```diff
@@ -1,8 +1,10 @@
-from enum import Enum
+from enum import auto
```
jni, line 1: mind. blown.

napari/util/misc.py
```diff
@@ -210,3 +211,16 @@ def segment_normal_vector(a, b):
     else:
         unit_norm = normal/norm
     return unit_norm
+
+
+class StringEnum(Enum):
+    def _generate_next_value_(name, start, count, last_values):
+        """ autonaming function assigns each value its own name as a value
+        """
+        return name
```
jni: This is so great! But how about using `name.lower()`? Then you can set it with any case input.

```
In [1]: from enum import Enum, auto

In [2]: class LowerName(Enum):
   ...:     def _generate_next_value_(name, start, count, last_values):
   ...:         return name.lower()
   ...:     

In [3]: class Mode(LowerName):
   ...:     PANZOOM = auto()
   ...:     SELECT = auto()
   ...:     

In [4]: x = Mode('panzoom')

In [5]: x
Out[5]: <Mode.PANZOOM: 'panzoom'>
```

jni: Sorry, I take that back, then it only works in lowercase...

kne42: i like lowercase names more because when we're spitting the mode back out to the user, lowercase names look nicer (e.g. `'PICKER'` vs `'picker'`)

kevinyamauchi: I totally agree that lowercase looks better to the user. However, the style in [the docs](https://docs.python.org/3/library/enum.html) is to use upper case names (like a constant I guess?). To harmonize these two things, the Enum names are capitol, but in the layer setters, I  call `upper()` on the string (similar to @jni's suggestion, e.g., [`Markers` layer](https://github.com/kevinyamauchi/napari/blob/a7346c9712973420304a72f9d3c4ddbf73d565c8/napari/layers/_markers_layer/model.py#L262)). Thus, our internal representation complies with the style guide and the user can use lower case strings.

I suppose we can also overload the `_missing_` method of `StringEnum` to capitalize strings so `Mode` enums can be instantiated with lower case strings (Python 3.6+ only). For example:
```python
class StringEnum(Enum):
    def _generate_next_value_(name, start, count, last_values):
        """ autonaming function assigns each value its own name as a value
        """
        return name

    def _missing_(cls, value):
       """ function called with provided value does not match any of the class
           member values. This function tries again with an upper case string.
       """
        return cls(value.upper())

    def __str__(self):
        """String representation: The string method returns the
        valid vispy symbol string for the Markers visual.
        """
        return self.value.lower()
```

Thoughts?

kne42: I don't understand – where are you getting the idea that the enum values have to be uppercase too? That was just an example of a `_generate_next_value_` method.

Also big fan of the `_missing_` method over the `.upper()` everywhere.

kevinyamauchi: Ah, sorry for the confusion. With regards to upper case names, I meant that the name (as opposed to value) of the Enum member should be upper case. I thought you saying that you preferred lowercase names over upper case names. If you'd prefer the values to be lowercase, I can modify `_generate_next_value_`.

I can implement the _missing_ method and update the setters!

kevinyamauchi: I just added a `__missing__` method to `StringEnum` so now the user can instantiate `StringEnum` with any case strings (i.e., `Mode('pan_zoom') == Mode('PAN_ZOOM'`). Also, I updated the `__str__` method such that the Enum string is returned as lower case.

The remaining question is should the `StringEnum` values be upper or lower case strings :P. Currently, they are upper case to match the name. I am happy either way. Any strong preferences @kne42 , @jni, @sofroniewn ?

kne42: lower

napari/util/misc.py
```diff
@@ -210,3 +211,23 @@ def segment_normal_vector(a, b):
...
+
+    @classmethod
```
kne42: this doesn't need to be a classmethod from what i can tell in the docs

kevinyamauchi: Good point. I changed `_missing_()` to an instance method.

sofroniewn (approved):
This all looks good to me, nice job @kevinyamauchi!

jni (approved):
Very good stuff! I love it! =D

Issue #294, opened 2019-05-25
Image thumbnails


sofroniewn:
## 🚀 Feature
It would be nice to see mini image thumbnails of the layers on the right-hand side in the layers list. This will help keep track of which layer is which when we have many layers. For image / labels layers it would be pretty easy to downsample the array and pass it through the colormap and display it using a qt widget. For shapes we'd need to rasterize the layer with the appropriate colors, this shouldn't be too much of a modification of our additional rasterize code. For vectors and markers, we'd have to do something similar. My main concern there is refreshing on interactivity as we might not want to have to rasterize each time there is a refresh, though maybe if we do the downsampling before the rasterization it wouldn't be too bad.

I'm happy to work on this feature, both the thumbnail generation and integration with the layer properties

sofroniewn:
@AhmetCanSolak suggested that we embed a colormap as the bottom row, separated by a black border into the thumbnail. This sounds like a good idea.

Also @kne42 suggested that we add a timer delay to avoid continually refreshing the thumbnail on every edit

sofroniewn:
Support for image thumbnails has been added in #314. Support for vectors, shapes, and markers thumbnails is still coming

sofroniewn:
Thumbnails for all layers are now supported, including vectors #377, points #352, and shapes #364

Issue #285, opened 2019-05-22
Generate layer Qt widgets automatically


sofroniewn:
## 🚀 Feature
In conversation today, @kne42 proposed that we autogenerate the `QtLayerProperties` and `QtLayerControls` files from the `Layer` model files for our different layer types, (i.e `Image`, `Labels`, `Shapes`, etc.). This would dramatically simplify what it takes to add new layer types and unify parts of the code base that feel heavily duplicated. It would further enforce separation between our model and view code, as now we'd be passing our model to the view constructor to automatically generate the needed view components. We'd need a way to translate from the properties on the `Layer` to the `QtWidget` for example, the `size` property on the markers would need to translate to a slider with `range(1, 100)`. For the properties it seems pretty straightforward. For the controls with all the buttons it seems a little harder (not quite sure how we'd get the clim histogram in there, but that could be extra). I also feel like this allows us to explore and throughly vet our system for autogenerating `Qt` that is going to be necessary for plugins (in some way now each layer is thought of as a special type of plugin, something we've discussed in the past).

Curious if @kne42 thinks I've captured this idea accurately and what @jni and @royerloic think.

sofroniewn:
@tlambert03 recently mentioned possibly doing something like this based on his previous experience [here](https://github.com/tlambert03/MOSAICpy/blob/1b51431d9c411e8d4ad3f6bff93b976a6b533097/mosaicpy/gui/helpers.py#L148-L202)

tlambert03:
for a more readable example of a use case, see the documentation i was starting to build that leverages that functionality here:
https://mosaicpy.readthedocs.io/en/latest/pages/imgprocessors.html

While I wouldn't exactly propose the same exact structure for napari, the basic idea there is that we introspect the `ImgProcessor.__init__` signature to generate the appropriate `QWidget` types to put into a GUI, and return the corresponding getter/setter function handles to the user to connect as they like.  The ultimate point there is that the user can add GUI elements by writing a simple function.

sofroniewn:
Very cool @tlambert03, curious what @jni and @kne42 make of this

jni:
I totally love the idea of autogenerating UI elements, especially for user plugins in dockable widgets as discussed in [zulip](https://napari.zulipchat.com/#narrow/stream/212875-general/topic/dock.20widgets), but if we can use our own framework on our own layers, indeed, all the better.

[...]

PR #314, opened 2019-06-03
Thumbnails

sofroniewn:
# Description
This PR adds support for image thumbnails to the `Image` and `Labels` layers as requested in #294. Support for the `Shapes`, `Markers`, and `Vectors` layers can come in follow up PRs, but I want to see that we all agree we are on the right track here before implementing them.

The `_image_view` is downsample, and then passed through the appropriate clims, colormap, and opacity settings. Right now I do the downsampling with `scipy.ndimage.zoom`, but this can lead to some weird artifacts for the labels layer, or a binary image due to the nature of the downsampling. I'm curious what others think. Should we do a `mean` or `median` downsample instead to avoid this?

Also the images right now are max 32x2x pixels, which is still pretty small, but maybe ok. I didn't implement yet the color bar suggestion from @AhmetCanSolak and @freeman-lab as I was worried the pictures would then look even smaller. It is not clear to me that it is needed, but I'm open to doing it in a follow up PR.

Here are some screen shots:
from `examples/layers.py`
image
and from `examples/labels-0-2d.py`
image

## Type of change

# How has this been tested?

- [x] run examples

## Final checklist:

### DIFF
--- a/napari/layers/_base_layer/model.py
+++ b/napari/layers/_base_layer/model.py
@@ -63,11 +63,14 @@ class Layer(VisualWrapper, ABC):
         self._indices = (0, 0)
         self._position = (0, 0)
         self.coordinates = (0, 0)
+        self._thumbnail_shape = (28, 28, 4)
+        self._thumbnail = np.zeros(self._thumbnail_shape, dtype=np.uint8)
         self._name = ''
         self.events.add(select=Event,
                         deselect=Event,
                         data=Event,
                         name=Event,
+                        thumbnail=Event,
                         status=Event,
                         help=Event,
                         interactive=Event,
@@ -154,6 +157,19 @@ class Layer(VisualWrapper, ABC):
     @abstractmethod
     def _get_shape(self):
         raise NotImplementedError()
+
+    @property
+    def thumbnail(self):
+        """np.ndarray: Integer array of thumbnail for the layer
+        """
+        return self._thumbnail
+
+    @thumbnail.setter
+    def thumbnail(self, thumbnail):
+        if thumbnail == self.thumbnail:
+            return
+        self._thumbnail = thumbnail
+        self.events.thumbnail()
 
     @property
     def ndim(self):
--- a/napari/layers/_base_layer/model.py
+++ b/napari/layers/_base_layer/model.py
@@ -63,7 +63,7 @@ class Layer(VisualWrapper, ABC):
         self._indices = (0, 0)
         self._position = (0, 0)
         self.coordinates = (0, 0)
-        self._thumbnail_shape = (28, 28, 4)
+        self._thumbnail_shape = (28, 28, 3)
         self._thumbnail = np.zeros(self._thumbnail_shape, dtype=np.uint8)
         self._name = ''
         self.events.add(select=Event,
@@ -166,8 +166,6 @@ class Layer(VisualWrapper, ABC):
 
     @thumbnail.setter
     def thumbnail(self, thumbnail):
-        if thumbnail == self.thumbnail:
-            return
         self._thumbnail = thumbnail
         self.events.thumbnail()
 
--- a/napari/layers/_base_layer/model.py
+++ b/napari/layers/_base_layer/model.py
@@ -63,7 +63,7 @@ class Layer(VisualWrapper, ABC):
         self._indices = (0, 0)
         self._position = (0, 0)
         self.coordinates = (0, 0)
-        self._thumbnail_shape = (28, 28, 3)
+        self._thumbnail_shape = (32, 32, 4)
         self._thumbnail = np.zeros(self._thumbnail_shape, dtype=np.uint8)
         self._name = ''
         self.events.add(select=Event,
--- a/napari/layers/_base_layer/view/properties.py
+++ b/napari/layers/_base_layer/view/properties.py
@@ -33,8 +33,8 @@ class QtLayer(QFrame):
         tb = QLabel(self)
         tb.setObjectName('thumbmnail')
         tb.setToolTip('Layer thumbmnail')
-        image = QImage(self.layer.thumbnail, self.layer._thumbnail_shape[0],
-                       self.layer._thumbnail_shape[1], QImage.Format_RGB888)
+        image = QImage(self.layer.thumbnail, self.layer.thumbnail.shape[0],
+                       self.layer.thumbnail.shape[1], QImage.Format_RGB888)
         tb.setPixmap(QPixmap.fromImage(image))
         self.thumbnail_label = tb
         self.grid_layout.addWidget(tb, 0, 1, 1, 1)
@@ -162,6 +162,6 @@ class QtLayer(QFrame):
             self.visibleCheckBox.setChecked(self.layer.visible)
 
     def _on_thumbnail_change(self, event):
-        image = QImage(self.layer.thumbnail, self.layer._thumbnail_shape[0],
-                       self.layer._thumbnail_shape[1], QImage.Format_RGB888)
+        image = QImage(self.layer.thumbnail, self.layer.thumbnail.shape[0],
+                       self.layer.thumbnail.shape[1], QImage.Format_RGB888)
         self.thumbnail_label.setPixmap(QPixmap.fromImage(image))
--- a/napari/layers/_base_layer/view/properties.py
+++ b/napari/layers/_base_layer/view/properties.py
@@ -1,6 +1,7 @@
 from qtpy.QtWidgets import (QSlider, QLineEdit, QGridLayout, QFrame, QLabel,
                             QVBoxLayout, QCheckBox, QWidget, QComboBox)
 from qtpy.QtCore import Qt
+from qtpy.QtGui import QImage, QPixmap
 
 
 class QtLayer(QFrame):
@@ -14,6 +15,7 @@ class QtLayer(QFrame):
         layer.events.blending.connect(self._on_blending_change)
         layer.events.opacity.connect(self._on_opacity_change)
         layer.events.visible.connect(self._on_visible_change)
+        layer.events.thumbnail.connect(self._on_thumbnail_change)
 
         self.setObjectName('layer')
 
@@ -26,7 +28,16 @@ class QtLayer(QFrame):
         cb.setProperty('mode', 'visibility')
         cb.stateChanged.connect(lambda state=cb: self.changeVisible(state))
         self.visibleCheckBox = cb
-        self.grid_layout.addWidget(cb, 0, 0)
+        self.grid_layout.addWidget(cb, 0, 0, 1, 1)
+
+        tb = QLabel(self)
+        tb.setObjectName('thumbmnail')
+        tb.setToolTip('Layer thumbmnail')
+        image = QImage(self.layer.thumbnail, self.layer._thumbnail_shape[0],
+                       self.layer._thumbnail_shape[1], QImage.Format_RGB888)
+        tb.setPixmap(QPixmap.fromImage(image))
+        self.thumbnail_label = tb
+        self.grid_layout.addWidget(tb, 0, 1, 1, 1)
 
         textbox = QLineEdit(self)
         textbox.setText(layer.name)
@@ -37,9 +48,9 @@ class QtLayer(QFrame):
         textbox.setEnabled(True)
         textbox.editingFinished.connect(self.changeText)
         self.nameTextBox = textbox
-        self.grid_layout.addWidget(textbox, 0, 1)
+        self.grid_layout.addWidget(textbox, 0, 2, 1, 2)
 
-        self.grid_layout.addWidget(QLabel('opacity:'), 1, 0)
+        self.grid_layout.addWidget(QLabel('opacity:'), 1, 0, 1, 2)
         sld = QSlider(Qt.Horizontal, self)
         sld.setFocusPolicy(Qt.NoFocus)
         sld.setFixedWidth(110)
@@ -50,7 +61,7 @@ class QtLayer(QFrame):
         sld.valueChanged[int].connect(
             lambda value=sld: self.changeOpacity(value))
         self.opacitySilder = sld
-        self.grid_layout.addWidget(sld, 1, 1)
+        self.grid_layout.addWidget(sld, 1, 2, 1, 2)
 
         blend_comboBox = QComboBox()
         for blend in self.layer._blending_modes:
@@ -61,8 +72,8 @@ class QtLayer(QFrame):
         blend_comboBox.activated[str].connect(
             lambda text=blend_comboBox: self.changeBlending(text))
         self.blendComboBox = blend_comboBox
-        self.grid_layout.addWidget(QLabel('blending:'), 2, 0)
-        self.grid_layout.addWidget(blend_comboBox, 2, 1)
+        self.grid_layout.addWidget(QLabel('blending:'), 2, 0, 1, 2)
+        self.grid_layout.addWidget(blend_comboBox, 2, 2, 1, 2)
 
         self.setLayout(self.grid_layout)
         msg = 'Click to select\nDrag to rearrange\nDouble click to expand'
@@ -123,11 +134,13 @@ class QtLayer(QFrame):
             self.setFixedHeight(55)
         rows = self.grid_layout.rowCount()
         for i in range(1, rows):
-            for j in range(2):
-                if self.expanded:
-                    self.grid_layout.itemAtPosition(i, j).widget().show()
-                else:
-                    self.grid_layout.itemAtPosition(i, j).widget().hide()
+            for j in range(4):
+                item = self.grid_layout.itemAtPosition(i, j)
+                if item is not None:
+                    if self.expanded:
+                        item.widget().show()
+                    else:
+                        item.widget().hide()
 
     def _on_layer_name_change(self, event):
         with self.layer.events.name.blocker():
@@ -147,3 +160,8 @@ class QtLayer(QFrame):
     def _on_visible_change(self, event):
         with self.layer.events.visible.blocker():
             self.visibleCheckBox.setChecked(self.layer.visible)
+
+    def _on_thumbnail_change(self, event):
+        image = QImage(self.layer.thumbnail, self.layer._thumbnail_shape[0],
+                       self.layer._thumbnail_shape[1], QImage.Format_RGB888)
+        self.thumbnail_label.setPixmap(QPixmap.fromImage(image))
--- a/napari/layers/_base_layer/view/properties.py
+++ b/napari/layers/_base_layer/view/properties.py
@@ -1,5 +1,6 @@
 from qtpy.QtWidgets import (QSlider, QLineEdit, QGridLayout, QFrame, QLabel,
-                            QVBoxLayout, QCheckBox, QWidget, QComboBox)
+                            QVBoxLayout, QCheckBox, QWidget, QComboBox,
+                            QHBoxLayout)
 from qtpy.QtCore import Qt
 from qtpy.QtGui import QImage, QPixmap
 
@@ -19,8 +20,20 @@ class QtLayer(QFrame):
 
         self.setObjectName('layer')
 
+        self.vbox_layout = QVBoxLayout()
+        self.top = QFrame()
+        self.top_layout = QHBoxLayout()
+        self.grid = QFrame()
         self.grid_layout = QGridLayout()
-        self.setContentsMargins(0, 0, 0, 0)
+        self.vbox_layout.addWidget(self.top)
+        self.vbox_layout.addWidget(self.grid)
+        self.vbox_layout.setSpacing(0)
+        self.top.setFixedHeight(38)
+        self.top_layout.setContentsMargins(0, 0, 0, 0)
+        self.grid_layout.setContentsMargins(0, 0, 0, 0)
+        self.top.setLayout(self.top_layout)
+        self.grid.setLayout(self.grid_layout)
+        self.setLayout(self.vbox_layout)
 
         cb = QCheckBox(self)
         cb.setObjectName('visibility')
@@ -29,30 +42,31 @@ class QtLayer(QFrame):
         cb.setProperty('mode', 'visibility')
         cb.stateChanged.connect(lambda state=cb: self.changeVisible(state))
         self.visibleCheckBox = cb
-        self.grid_layout.addWidget(cb, 0, 0, 1, 1)
+        self.top_layout.addWidget(cb)
 
         tb = QLabel(self)
         tb.setObjectName('thumbmnail')
         tb.setToolTip('Layer thumbmnail')
         self.thumbnail_label = tb
         self._on_thumbnail_change(None)
-        self.grid_layout.addWidget(tb, 0, 1, 1, 1, Qt.AlignLeft)
+        self.top_layout.addWidget(tb)
 
         textbox = QLineEdit(self)
         textbox.setText(layer.name)
         textbox.home(False)
         textbox.setToolTip('Layer name')
-        textbox.setFixedWidth(122)
+        #textbox.setFixedWidth(122)
         textbox.setAcceptDrops(False)
         textbox.setEnabled(True)
         textbox.editingFinished.connect(self.changeText)
         self.nameTextBox = textbox
-        self.grid_layout.addWidget(textbox, 0, 2, 1, 2)
+        self.top_layout.addWidget(textbox)
 
-        self.grid_layout.addWidget(QLabel('opacity:'), 1, 0, 1, 2)
+        row = self.grid_layout.rowCount()
+        self.grid_layout.addWidget(QLabel('opacity:'), row, 0)
         sld = QSlider(Qt.Horizontal, self)
         sld.setFocusPolicy(Qt.NoFocus)
-        sld.setFixedWidth(110)
+        #sld.setFixedWidth(110)
         sld.setMinimum(0)
         sld.setMaximum(100)
         sld.setSingleStep(1)
@@ -60,8 +74,9 @@ class QtLayer(QFrame):
         sld.valueChanged[int].connect(
             lambda value=sld: self.changeOpacity(value))
         self.opacitySilder = sld
-        self.grid_layout.addWidget(sld, 1, 2, 1, 2)
+        self.grid_layout.addWidget(sld, row, 1)
 
+        row = self.grid_layout.rowCount()
         blend_comboBox = QComboBox()
         for blend in self.layer._blending_modes:
             blend_comboBox.addItem(blend)
@@ -71,10 +86,9 @@ class QtLayer(QFrame):
         blend_comboBox.activated[str].connect(
             lambda text=blend_comboBox: self.changeBlending(text))
         self.blendComboBox = blend_comboBox
-        self.grid_layout.addWidget(QLabel('blending:'), 2, 0, 1, 2)
-        self.grid_layout.addWidget(blend_comboBox, 2, 2, 1, 2)
+        self.grid_layout.addWidget(QLabel('blending:'), row, 0)
+        self.grid_layout.addWidget(blend_comboBox, row, 1)
 
-        self.setLayout(self.grid_layout)
         msg = 'Click to select\nDrag to rearrange\nDouble click to expand'
         self.setToolTip(msg)
         self.setExpanded(False)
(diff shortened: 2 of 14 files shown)


napari/layers/_base_layer/model.py
```diff
@@ -63,11 +63,14 @@ def __init__(self, central_node, name=None):
         self._indices = (0, 0)
         self._position = (0, 0)
         self.coordinates = (0, 0)
+        self._thumbnail_shape = (32, 32, 4)
+        self._thumbnail = np.zeros(self._thumbnail_shape, dtype=np.uint8)
```
jni, line 67: I'm not sure this should go in the base layer model as opposed to a view... Views don't necessarily have thumbnails?

napari/layers/_base_layer/model.py
```diff
@@ -155,6 +158,17 @@ def data(self, data):
...
+    @property
+    def thumbnail(self):
+        """np.ndarray: Integer array of thumbnail for the layer
+        """
+        return self._thumbnail
+
+    @thumbnail.setter
+    def thumbnail(self, thumbnail):
+        self._thumbnail = thumbnail
+        self.events.thumbnail()
```
jni, line 170: ... But, this is useful — I'm certainly open to being talked into thumbnails-in-the-model.

napari/layers/_base_layer/view/properties.py
```diff
@@ -117,17 +129,19 @@ def setExpanded(self, bool):
...
-            self.setFixedHeight(55)
+            self.setFixedHeight(60)
         rows = self.grid_layout.rowCount()
         for i in range(1, rows):
-            for j in range(2):
-                if self.expanded:
-                    self.grid_layout.itemAtPosition(i, j).widget().show()
-                else:
-                    self.grid_layout.itemAtPosition(i, j).widget().hide()
+            for j in range(4):
```
jni: What's "4" here?

napari/layers/_image_layer/model.py
```diff
@@ -326,6 +332,34 @@ def interpolation_functions(self):
...
+        if self.multichannel:
+            thumbnail = ndi.zoom(self._image_view,
+                                 (zoom_factor, zoom_factor, 1))
+            if self._image_view.shape[2] == 4:
+                mapped = thumbnail
+            else:
+                mapped = np.pad(thumbnail, (0, 0, 255), 'constant',
+                                constant_values=1)
+        else:
+            thumbnail = ndi.zoom(self._image_view, zoom_factor)
```
jni: Oh yeah I totally think this is the wrong approach, and would use simple subsampling here. Otherwise it's going to be sluggish with many layers and in general it's mostly wasted computation. Currently you're doing bicubic interpolation which I think is insane for this use. :joy: My suggestion is `zoom(image, zoom_factor, order=0, prefilter=False)`, which *I believe* is roughly equivalent to `image[::zoom_factor, ::zoom_factor]`.

napari/layers/_image_layer/model.py
```diff
@@ -326,6 +332,34 @@ def interpolation_functions(self):
...
+        zoom_factor = np.divide(self._thumbnail_shape[:2],
+                                self._image_view.shape[:2]).min()
+        if self.multichannel:
+            thumbnail = ndi.zoom(self._image_view,
+                                 (zoom_factor, zoom_factor, 1))
+            if self._image_view.shape[2] == 4:
+                mapped = thumbnail
+            else:
+                mapped = np.pad(thumbnail, (0, 0, 255), 'constant',
+                                constant_values=1)
```
jni: i think there's probably a better way to express this. Currently I actually can't for the life of me figure out what this expression does. (The [docs](https://docs.scipy.org/doc/numpy/reference/generated/numpy.pad.html) for numpy.pad don't seem to explain what a 3-tuple would do here...)

napari/layers/_image_layer/model.py
```diff
@@ -326,6 +332,34 @@ def interpolation_functions(self):
...
+            thumbnail = ndi.zoom(self._image_view, zoom_factor)
+            image = np.clip(thumbnail, self.clim[0], self.clim[1])
+            image = image - self.clim[0]
+            color_range = self.clim[1] - self.clim[0]
+            if color_range != 0:
+                image = image/color_range
+            mapped = self.colormap[1].map(image) * 255
+            mapped = mapped.reshape(list(thumbnail.shape) + [4])
+        total_padding = np.subtract(self.thumbnail.shape, mapped.shape)
+        padding = [(p//2, (p+1)//2) for p in total_padding]
```
jni: I presume this is the correct way to handle both odd and even axis shapes, but it's not obvious to me. Perhaps you could add a comment with some example paddings for given image shapes?

napari/layers/_image_layer/model.py
```diff
@@ -326,6 +332,34 @@ def interpolation_functions(self):
     def _clim_range_default(self):
         return [float(self.image.min()), float(self.image.max())]
 
+    def _update_thumbnail(self):
+        """Update thumbnail with current image data and colormap.
+        """
```
jni, line 338: In this function, I don't understand the conceptual difference between `mapped` and `thumbnail`. Could you explain it?

napari/layers/_image_layer/model.py
```diff
@@ -391,6 +425,7 @@ def to_xml_list(self):
             as a png according to the svg specification.
         """
         image = np.clip(self._image_view, self.clim[0], self.clim[1])
+        image = image - self.clim[0]
```
jni, line 432: Hmmm, hardcoding the clims... I'm not sure I love it, but either way, I think I would prefer not to repeat variable names, keeping everything functional:

```python
low, high = self.clim
image_clipped = np.clip(self._image_view, low, high)
if color_range != 0:
    image_normalized = (image_clipped - low) / (high - low)
else:
    image_normalized = image_clipped
```

(I don't see the need to subtract the clim if the range is 0, and indeed I think it might be the wrong thing to do.)

napari/layers/_image_layer/view/properties.py
```diff
@@ -18,8 +18,8 @@ def __init__(self, layer):
         comboBox.setCurrentIndex(index)
         comboBox.activated[str].connect(
             lambda text=comboBox: self.changeColor(text))
-        self.grid_layout.addWidget(QLabel('colormap:'), 3, 0)
-        self.grid_layout.addWidget(comboBox, 3, 1)
+        self.grid_layout.addWidget(QLabel('colormap:'), 3, 0, 1, 2)
+        self.grid_layout.addWidget(comboBox, 3, 2, 1, 2)
```
jni: I find all the magic numbers quite offputting. Perhaps they can be defined, or at least used as kwargs so that newcomers can see what they all mean?

napari/layers/_labels_layer/model.py
```diff
@@ -507,6 +501,21 @@ def get_message(self, coord, value):
 
         return msg
 
+    def _update_thumbnail(self):
+        """Update thumbnail with current image data and colors.
+        """
+        zoom_factor = np.divide(self._thumbnail_shape[:2],
+                                self._image_view.shape[:2]).min()
+        thumbnail = np.round(ndi.zoom(self._image_view, zoom_factor))
```
jni: see my comments above for alternative parameters to `ndi.zoom` that I think we should use in both layers but definitely we should use in this one!

napari/layers/_markers_layer/view/properties.py
```diff
@@ -31,8 +31,8 @@ def __init__(self, layer):
         sld.setValue(int(value))
         sld.valueChanged[int].connect(lambda value=sld: self.changeSize(value))
         self.sizeSlider = sld
-        self.grid_layout.addWidget(QLabel('size:'), 3, 0)
-        self.grid_layout.addWidget(sld, 3, 1)
+        self.grid_layout.addWidget(QLabel('size:'), 3, 0, 1, 2)
```
jni: See above — let's use kwargs in this situation to make the meaning more obvious here.

jni (commented):
@sofroniewn the screenshots look great! The only change I would make is to move the thumbnails right up against the eye icon. Not sure if that's feasible. I also wonder if we should stop forcing alignment between the top line of the expanded widget (the icons and name) and the remaining ones (options). There's a lot of wasted space in the title line that could go to the layer name.

Good stuff!

napari/layers/_base_layer/model.py
sofroniewn: I like having the actual data of the thumbnail, and the calculation of it in the model, as I think it is most consistent with other things we have in there. All the display stuff should agreed be in the View. Could revisit in the future if our refactoring takes us there

napari/layers/_base_layer/model.py
sofroniewn: see above

napari/layers/_base_layer/view/properties.py
sofroniewn: fixed - by calculating number of columns instead and putting it in a variable

napari/layers/_image_layer/model.py
sofroniewn: this works great, much better for the labels layer especially, thanks!

napari/layers/_labels_layer/model.py
sofroniewn: fixed, definitely an improvement, especially here

napari/layers/_image_layer/view/properties.py
sofroniewn: I think we need a follow up PR fixing this across the whole code base, I'd rather not do this partially here. In the long run I #285 will be part of the solution, but there is lots we could do before then

napari/layers/_markers_layer/view/properties.py
sofroniewn: as noted above I'd prefer to have a follow up PR that refactors all these numbers for all the properties widgets

napari/layers/_image_layer/model.py
sofroniewn: turns out padding is not necessary, so I have removed

napari/layers/_image_layer/model.py
sofroniewn: turns out padding is not necessary, so I have removed

sofroniewn: yes, this is likely a mistake and wouldn't have run, have fixed

napari/layers/_image_layer/model.py
sofroniewn: I've renamed that first one `zoomed` instead of `thumbnail` as it was a little too soon to use that word, it corresponds to the downsampled image. `mapped` corresponds to the result after mapping through colormaps. i can improve if still confusing

napari/layers/_image_layer/model.py
sofroniewn: fixed the repetition of variable names and subtraction of clim for range `0`

kne42:
definitely left-align the thumbnails :)

sofroniewn:
I have made fixes requested by @jni including left aligning the thumbnails as much as possible - see improved screenshot:

image

If @kne42 could give a review too / weigh in on some of the comments I left unresolved that would be great

napari/layers/_base_layer/model.py
kne42: Agree that the thumbnail should be calculated in the model.

napari/layers/_image_layer/model.py
kne42: Isn't a more accurate name "downsampled?"

sofroniewn: yup, better choice - will fix

napari/layers/_base_layer/view/properties.py
```diff
@@ -26,7 +29,16 @@ def __init__(self, layer):
         cb.setProperty('mode', 'visibility')
         cb.stateChanged.connect(lambda state=cb: self.changeVisible(state))
         self.visibleCheckBox = cb
-        self.grid_layout.addWidget(cb, 0, 0)
+        self.grid_layout.addWidget(cb, 0, 0, 1, 1)
+
+        tb = QLabel(self)
+        tb.setObjectName('thumbmnail')
+        tb.setToolTip('Layer thumbmnail')
+        image = QImage(self.layer.thumbnail, self.layer.thumbnail.shape[1],
```
kne42: kinda confused – could you explain to me why the shape is reordered here when inserting as an image?

kne42: you're also using a pretty long name here – `self.layer.thumbnail` – since it's repeated a lot it might be best to reassign it to the shorter `thumbnail`

napari/layers/_base_layer/view/properties.py
```diff
@@ -117,17 +129,20 @@ def setExpanded(self, bool):
...
         for i in range(1, rows):
-            for j in range(2):
-                if self.expanded:
-                    self.grid_layout.itemAtPosition(i, j).widget().show()
-                else:
-                    self.grid_layout.itemAtPosition(i, j).widget().hide()
+            for j in range(columns):
+                item = self.grid_layout.itemAtPosition(i, j)
+                if item is not None:
+                    if self.expanded:
```
kne42: looking at this code now, and taking into consideration @jni's comments about left-alignment and layer name space, I would structure the layout like so:
```
VBoxLayout
    layer visibility | thumbnail | name
    FormLayout
        property name | property widget
```
then you can hide the entire form layout at once instead of iterating over and hiding each individual property like this

napari/layers/_base_layer/view/properties.py
```diff
@@ -147,3 +162,8 @@ def _on_blending_change(self, event):
     def _on_visible_change(self, event):
         with self.layer.events.visible.blocker():
             self.visibleCheckBox.setChecked(self.layer.visible)
+
+    def _on_thumbnail_change(self, event):
+        image = QImage(self.layer.thumbnail, self.layer.thumbnail.shape[1],
```
kne42: this is repeated – may want to look into putting this logic into its own function and then connect it into an event, wrapping it with a lambda

napari/layers/_image_layer/model.py
```diff
@@ -326,6 +332,32 @@ def interpolation_functions(self):
...
+        """
+        zoom_factor = np.divide(self._thumbnail_shape[:2],
+                                self._image_view.shape[:2]).min()
+        if self.multichannel:
+            zoomed = ndi.zoom(self._image_view, (zoom_factor, zoom_factor, 1),
+                              prefilter=False, order=0)
+            if self._image_view.shape[2] == 4:
+                mapped = zoomed
+            else:
+                ones = 255 * np.ones(list(zoomed.shape[:2]) + [1])
```
kne42:
```suggestion
                ones = 255 * np.ones(zoomed.shape[:2] + (1,))
```

napari/layers/_labels_layer/model.py
```diff
@@ -507,6 +501,19 @@ def get_message(self, coord, value):
...
+    def _update_thumbnail(self):
+        """Update thumbnail with current image data and colors.
+        """
+        zoom_factor = np.divide(self._thumbnail_shape[:2],
+                                self._image_view.shape[:2]).min()
+        zoomed = np.round(ndi.zoom(self._image_view, zoom_factor,
+                                   prefilter=False, order=0))
+        zoomed = self.raw_to_displayed(zoomed)
+        mapped = self.colormap.map(zoomed) * 255
+        mapped = mapped.reshape(list(zoomed.shape) + [4])
```
kne42:
```suggestion
        mapped = mapped.reshape(zoomed.shape + (4,))
```

napari/layers/_labels_layer/model.py
```diff
@@ -507,6 +501,19 @@ def get_message(self, coord, value):
...
+        mapped[:, :, 3] = mapped[:, :, 3] * self.opacity
```
kne42: is this different if you did
```python
mapped[:, :, 3] *= self.opacity
```
?

kne42 (commented):
Looking good! Agree with @jni that we should somehow document these "magic numbers." While @sofroniewn wants to hold off for a mega PR to fix this, I think that this PR should fix whatever it introduces. It is not guaranteed that whoever does the big fix will actually know what these mean (even if it is @sofroniewn – he may have forgot!), so it's best to clarify these while we still can.

napari/layers/_image_layer/model.py
jni:
```suggestion
                ones = np.full(zoomed.shape[:2] + (1,), 255, dtype=np.uint8)
```

... raising the stakes in the suggestion game. =D

jni:
@sofroniewn looking good! I agree with @kne42 that it's best to make sure the code you introduce (in this case, the QLabels) is formatted as best we can now, as they will assist a future refactor.

napari/layers/_labels_layer/model.py
jni: even though we know this is a 2D image, I think it's nicer to use the nD idiom:

```suggestion
        mapped[..., 3] *= self.opacity
```

napari/layers/_labels_layer/model.py
```diff
@@ -507,6 +501,19 @@ def get_message(self, coord, value):
...
+    def _update_thumbnail(self):
+        """Update thumbnail with current image data and colors.
+        """
+        zoom_factor = np.divide(self._thumbnail_shape[:2],
+                                self._image_view.shape[:2]).min()
+        downsampled = np.round(ndi.zoom(self._image_view, zoom_factor,
+                                        prefilter=False, order=0))
+        downsampled = self.raw_to_displayed(downsampled)
+        mapped = self.colormap.map(downsampled) * 255
+        mapped = mapped.reshape(list(downsampled.shape) + [4])
```
jni:
```suggestion
        mapped = mapped.reshape(downsampled.shape + (4,))
```

I also think you should rename `mapped` to `colormapped`.

napari/layers/_labels_layer/model.py
```diff
@@ -507,6 +501,19 @@ def get_message(self, coord, value):
...
 
+    def _update_thumbnail(self):
+        """Update thumbnail with current image data and colors.
+        """
+        zoom_factor = np.divide(self._thumbnail_shape[:2],
+                                self._image_view.shape[:2]).min()
+        downsampled = np.round(ndi.zoom(self._image_view, zoom_factor,
+                                        prefilter=False, order=0))
+        downsampled = self.raw_to_displayed(downsampled)
+        mapped = self.colormap.map(downsampled) * 255
```
jni: I suggest removing the 255 here and then using `skimage.util.img_as_ubyte` to convert to uint8 and rescale in one call.

napari/layers/_image_layer/model.py
```diff
@@ -326,6 +332,33 @@ def interpolation_functions(self):
...
+                mapped = np.concatenate([downsampled, ones], axis=2)
+        else:
+            downsampled = ndi.zoom(self._image_view, zoom_factor,
+                                   prefilter=False, order=0)
+            low, high = self.clim
+            downsampled = np.clip(downsampled, low, high)
+            color_range = high - low
+            if color_range != 0:
+                downsampled = (downsampled - low) / color_range
+            mapped = self.colormap[1].map(downsampled) * 255
```
jni: Same comment as below — use `skimage.util.img_as_ubyte`

napari/layers/_image_layer/model.py
```diff
@@ -326,6 +332,33 @@ def interpolation_functions(self):
...
+            mapped = mapped.reshape(list(downsampled.shape) + [4])
```
jni:
```suggestion
            mapped = mapped.reshape(downsampled.shape + (4,))
```

napari/layers/_image_layer/model.py
```diff
@@ -326,6 +332,33 @@ def interpolation_functions(self):
...
+        mapped[:, :, 3] = mapped[:, :, 3] * self.opacity
```
jni:
```suggestion
        mapped[..., 3] *= self.opacity
```

napari/layers/_image_layer/model.py
```diff
@@ -326,6 +332,33 @@ def interpolation_functions(self):
...
+        zoom_factor = np.divide(self._thumbnail_shape[:2],
+                                self._image_view.shape[:2]).min()
+        if self.multichannel:
+            downsampled = ndi.zoom(self._image_view,
+                                   (zoom_factor, zoom_factor, 1),
+                                   prefilter=False, order=0)
+            if self._image_view.shape[2] == 4:
+                mapped = downsampled
+            else:
+                ones = 255 * np.ones(list(downsampled.shape[:2]) + [1])
```
jni:
```suggestion
                ones = np.full(downsampled.shape[:-1] + (1,), 255, dtype=np.uint8)
```

napari/layers/_image_layer/model.py
```diff
@@ -326,6 +332,33 @@ def interpolation_functions(self):
...
+    def _update_thumbnail(self):
+        """Update thumbnail with current image data and colormap.
+        """
+        zoom_factor = np.divide(self._thumbnail_shape[:2],
+                                self._image_view.shape[:2]).min()
+        if self.multichannel:
+            downsampled = ndi.zoom(self._image_view,
+                                   (zoom_factor, zoom_factor, 1),
+                                   prefilter=False, order=0)
+            if self._image_view.shape[2] == 4:
```
jni:
```suggestion
            if self._image_view.shape[2] == 4:  # image is RGBA
```

napari/layers/_image_layer/model.py
```diff
@@ -326,6 +332,33 @@ def interpolation_functions(self):
...
+                mapped = downsampled
+            else:
+                ones = 255 * np.ones(list(downsampled.shape[:2]) + [1])
+                mapped = np.concatenate([downsampled, ones], axis=2)
```
jni: This code appears to assume that `_image_view` is uint8, or `ones` should be actually full of 1.0s. Are we sure this is true?

jni (commented):
@sofroniewn there's a couple of places where the code seems to assume uint8 0-255 range of the *input* image, where it could be a 0-1 float or even maybe a 0-65535 uint16 (maybe?), so it's worth double checking that the logic works in all situations.

napari/layers/_image_layer/model.py
kne42: oooh yes i missed that one!

napari/layers/_image_layer/model.py
kne42: also don't think that `ones` is a good name for this since it doesn't actually consist of ones

napari/layers/_base_layer/view/properties.py
sofroniewn: QImage expects the image width followed by height. I've added a note and done the reassignment

napari/layers/_base_layer/view/properties.py
sofroniewn: removed the extra function call

napari/layers/_image_layer/model.py
sofroniewn: fixed

napari/layers/_labels_layer/model.py
sofroniewn: fixed

napari/layers/_labels_layer/model.py
sofroniewn: fixed

napari/layers/_image_layer/model.py
sofroniewn: fixed

napari/layers/_image_layer/model.py
sofroniewn: fixed and renamed alpha

napari/layers/_image_layer/model.py
sofroniewn: added

napari/layers/_image_layer/model.py
sofroniewn: done

napari/layers/_labels_layer/model.py
sofroniewn: done

napari/layers/_image_layer/model.py
sofroniewn: done

napari/layers/_labels_layer/model.py
sofroniewn: done

napari/layers/_image_layer/model.py
sofroniewn: fixed

sofroniewn:
Following suggestions from @jni I am now using skimage `img_as_ubyte` for converting images, which is probably a good thing. This adds `skimage` now as a dependency, which I have changed in the requirements.txt

Still have a couple things to do though!

napari/layers/_base_layer/view/properties.py
sofroniewn: fixed

sofroniewn:
I have now fixed the layout of the thumbnail making a vbox with an hbox and grid like @kne42 suggested, and I have removed the magic numbers, replacing them with an auto calculation of the appropriate row to add the next elements in.

napari/layers/_base_layer/view/properties.py
```diff
@@ -26,32 +42,41 @@ def __init__(self, layer):
...
+        self.thumbnail_label = tb
+        self._on_thumbnail_change(None)
+        self.top_layout.addWidget(tb)
 
         textbox = QLineEdit(self)
         textbox.setText(layer.name)
         textbox.home(False)
         textbox.setToolTip('Layer name')
-        textbox.setFixedWidth(122)
+        #textbox.setFixedWidth(122)
```
jni: Should this line be deleted?

napari/layers/_base_layer/view/properties.py
```diff
@@ -26,32 +42,41 @@ def __init__(self, layer):
...
-        self.grid_layout.addWidget(textbox, 0, 1)
+        self.top_layout.addWidget(textbox)
 
-        self.grid_layout.addWidget(QLabel('opacity:'), 1, 0)
+        row = self.grid_layout.rowCount()
+        self.grid_layout.addWidget(QLabel('opacity:'), row, 0)
         sld = QSlider(Qt.Horizontal, self)
         sld.setFocusPolicy(Qt.NoFocus)
-        sld.setFixedWidth(110)
+        #sld.setFixedWidth(110)
```
jni: Should this line be deleted?

napari/layers/_base_layer/view/properties.py
```diff
@@ -26,32 +42,41 @@ def __init__(self, layer):
...
         textbox.setAcceptDrops(False)
         textbox.setEnabled(True)
         textbox.editingFinished.connect(self.changeText)
         self.nameTextBox = textbox
-        self.grid_layout.addWidget(textbox, 0, 1)
+        self.top_layout.addWidget(textbox)
 
-        self.grid_layout.addWidget(QLabel('opacity:'), 1, 0)
+        row = self.grid_layout.rowCount()
+        self.grid_layout.addWidget(QLabel('opacity:'), row, 0)
```
jni: I would suggest: 

```suggestion
        self.grid_layout.addWidget(QLabel('opacity:'), row=row, column=0)
```

but happy to drop it here. =)

napari/layers/_base_layer/view/properties.py
```diff
@@ -26,32 +42,41 @@ def __init__(self, layer):
...
+        #sld.setFixedWidth(110)
         sld.setMinimum(0)
         sld.setMaximum(100)
         sld.setSingleStep(1)
         sld.setValue(self.layer.opacity*100)
         sld.valueChanged[int].connect(
             lambda value=sld: self.changeOpacity(value))
         self.opacitySilder = sld
-        self.grid_layout.addWidget(sld, 1, 1)
+        self.grid_layout.addWidget(sld, row, 1)
```
jni: as above

jni (approved):
Good stuff! I've added some minor quibbles that may or may not be addressed as preferred. =)

napari/layers/_base_layer/view/properties.py
sofroniewn: done

napari/layers/_base_layer/view/properties.py
sofroniewn: done

napari/layers/_base_layer/view/properties.py
sofroniewn: unfortunately I then get an error `row` is not a valid keyword argument. I will change it to
```python
row = self.grid_layout.rowCount()
column = 0
self.grid_layout.addWidget(QLabel('opacity:'), row, column)
```
and call it a day as I think it's then clear enough

sofroniewn: actually, I'm defining the following two variables
```python
self.name_column = 0
self.property_column = 1
```
and using them. can always fix in a followup PR

napari/layers/_base_layer/view/properties.py
sofroniewn: see above

PR #319, opened 2019-06-05
Change Image.interpolation to Enum

### LINKED ISSUE #132: Remove buggy interpolation modes
## 🐛 Bug

Some of Vispy interpolation modes for images are very buggy and lead to unsightly
visual artefacts such as black or white vertical lines.
Here is a list of buggy interpolation modes:

- quadric (white lines, sometimes fails entirely)
- sinc (black lines)

We should remove any buggy interpolation scheme,
we are more interpolation than is healthy to have.

## To Reproduce

1. run 'add_images.py'
2. change interpolation schemes and check eac one.

image

For quadric, the white lines are faint but there:
image

## Expected behavior

no weird lines.

## Environment

 - napari-gui Version (e.g., 1.0): master
 - OS (e.g., Linux): OSX
 - Python version: 3.7
 - Any other relevant information: None

## Additional context

None

kevinyamauchi:
# Description
As discussed in #265, this PR introduces an Enum to represent the Image layer interpolation options (`Interpolation`). The `Interpolation` is inherits from the `StringEnum` class introduced in  #311 and thus has string values and is implemented such that the `Layer.blending` property can be set with a string and returns a string. Further, we have removed the ability to set the buggy quadric and sinc interpolation modes (per #132).

Previously, `Image.interpolation` was stored as an index (`Image.interpolation_index`) and the `Image.interpolation` property and setter methods converted between the index and a string name via some helper functions in `/napari/util/interpolation.py'. I checked the code base and ran the examples listed below and it appeared that nothing else referenced that file, so I deleted it. Let me know if that was a mistake!

Similarly, `Image.interpolation_names` (tuples of interpolation mode names) appeared unused so I deleted it. However, if people think it's important to have available, I am happy to add it back in.

## Type of change

# Reference
This is PR#3 in #265 and closes #132.

# How has this been tested?
To test compatibility with layers, I ran the following examples
- [x] `Layer.Shapes`: `/examples/add_shapes.py`
- [x] `Layer.Markers`: `/examples/add_markers.py`
- [x] `Layer.Labels`: `/examples/labels-2d.py`
- [x] `Layer.Vectors`: `/examples/add_vectors_coords.py`
- [x] `Layer.Pyramid`: `/examples/add_pyramids.py`

To test the ability to set `Image.interpolation` with a string, I ran:
- [x] `/examples/add_image.py`

## Final checklist:


sofroniewn (changes requested):
This looks great! Can we also address #132 right now too and drop support for `quadric` and `sinc` modes due to the bugs described in that issue.

napari/layers/_image_layer/_constants.py
```diff
@@ -0,0 +1,30 @@
...
+    CATROM = auto()
+    GAUSSIAN = auto()
+    HAMMING = auto()
+    HANNING = auto()
+    HERMITE = auto()
+    KAISER = auto()
+    LANCZOS = auto()
+    MITCHELL = auto()
+    NEAREST = auto()
+    QUADRIC = auto()
```
sofroniewn: drop quadric and sinc as described in #132

kevinyamauchi: Great point. Done!

kevinyamauchi:
@sofroniewn thanks for the comments. I dropped the quadric and sinc interpolation modes as you suggested. I am happy to wait to merge until #322 (`black` PR) is landed.

napari/layers/_image_layer/model.py
```diff
@@ -98,7 +97,7 @@ class Image(Layer):
     _colormaps = AVAILABLE_COLORMAPS
 
     default_colormap = 'magma'
-    default_interpolation = 'nearest'
+    default_interpolation = Interpolation.NEAREST
```
kne42: I would consider this "public" so this should be the string version:
```suggestion
    default_interpolation = str(Interpolation.NEAREST)
```

napari/layers/_image_layer/model.py
```diff
@@ -312,17 +298,16 @@ def method(self, method):
     def interpolation(self):
         """string: Equipped interpolation method's name.
         """
-        return _index_to_name(self.interpolation_index)
+        return str(self._interpolation)
```
kne42, line 314: IMO this should just be a wrapper around `self._node.interpolation` which calls into question if we really need an Enum at all.

napari/layers/_image_layer/model.py
```diff
@@ -312,17 +298,16 @@ def method(self, method):
     def interpolation(self):
         """string: Equipped interpolation method's name.
```
kne42: this docstring should include all possible methods listed in set format as the type: `{'nearest', 'bicubic', ...}`

napari/layers/_image_layer/_constants.py
```diff
@@ -0,0 +1,28 @@
+from enum import auto
+
+from ...util.misc import StringEnum
+
+
+class Interpolation(StringEnum):
```
kne42, line 6: do you think this Enum needs its own file?

sofroniewn: I like the constants having their own files for consistencies sake, and the fact that we may add more constants here too. Maybe some of the colormap processing stuff should be in here too instead of the model file. Thoughts @jni?

napari/layers/_image_layer/model.py
sofroniewn: I like that the Enum allows us more control over which modes we expose and don't (i.e. it lets us drop `sinc`)

napari/layers/_image_layer/model.py
kevinyamauchi: Good point. Fixed!

napari/layers/_image_layer/model.py
kevinyamauchi: Fixed. Thanks!

napari/layers/_image_layer/_constants.py
kevinyamauchi: I agree that we should be consistent across the layer types. Given that the other layers store things like Enum's in a `_constants.py` file, I think we should do the same here. As @sofroniewn mentioned, we may add other things here in the future anyway.

kevinyamauchi:
Based on our conversation at the sync meeting last Thursday, I think we decided to move forward with the `Interpolation` Enum and that it should live in the `_constants.py` file. I think I have addressed the remaining comments. Please let me know if there's anything else you would like me to change!

kne42 (approved):

sofroniewn:
I have update this with the `black` formatting and will merge once all the tests pass

sofroniewn (approved):
LGTM will merge

PR #322, opened 2019-06-05
Reformatting whole repo with Black

AhmetCanSolak:
# Description

We are adapting `black` formatter and this is an urgent PR for reformatting of whole repo.

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Final checklist:


sofroniewn:
This looks good to me, but a couple questions
- Do we want to also format the `_vispy` folder as at some point we might want to merge that code back in and smaller diffs will be easier?
- Do we want to format the `vendored` folder in colormaps as isn't there some stuff we've taken from other places there that we might want to keep the same?
- Does this PR also add a test to check if the repo has been formatted correctly or is that planned in a followup PR? I think I'd rather see it in this PR too if possible as then we know we won't accidentally start breaking our formatting, but I'm open.

Would be great for @kne42 and @jni to weigh in and then we can move forward

kne42:
I would say:
1) yes, also this should be refactored into a `compat` folder anyways in a later PR
2) no, we want easier diffs if this is really just copy/pasted from `matplotlib`
3) it shouldn't be too difficult to add but i dont think it should go in this PR since it has such a large diff

sofroniewn:
Thanks @kne42, I'll let @jni weigh in on `vendored` as he made that folder.

If @AhmetCanSolak can be ready to make a follow up PR that adds a test for compliance with `black -S` that would be great, as then we can get both of those in

AhmetCanSolak:
I think we should add that format check to CI in a separate PR. 

for `vendored` i am open to both, let's see what @jni will say.

examples/dask_nD_image.py
```diff
@@ -10,7 +10,13 @@
...
-                     for f in np.linspace(0.05, 0.5, 10)], axis=0)
+    blobs = da.stack(
+        [
+            data.binary_blobs(
+                length=128, blob_size_fraction=0.05, n_dim=3, volume_fraction=f
+            )
+            for f in np.linspace(0.05, 0.5, 10)
+        ],
+        axis=0,
+    )
```
jni, line 21: Sheesh... This one is horrible for me. But it's ok, I'll hold my nose, I still think black is a net win...

sofroniewn (approved):
I'm approving this, maybe it'll spur @jni on to review and merge if he approves too ;-)

jni (commented):
I'm not strongly opposed to blackening `vendored` and `_vispy` (is `_vispy` vendored code? I can't tell, the git history is already a black box since I renamed `napari-gui` to `napari`. =\), but I think leaving that code alone is the right move. As @sofroniewn points out, it's not our code, and if we make changes and want to submit a PR to upstream projects, we don't want some enormous diff. The converse is also true.

AhmetCanSolak:
Thanks @jni @sofroniewn @kne42 . As far as i understand we are good to merge if i take reformatting on `vendored` back. Will do and merge today.

sofroniewn:
sounds good to me @AhmetCanSolak

kne42:
ah, i totally forgot about the upstream case, thanks @jni for reminding me!

with that in mind i think that both `vendored` and `_vispy` (and maybe even `util.event`) should be ignored

Issue #318, opened 2019-06-05
Even lazier visualization of lazy ndarrays


d-v-b:
## 🚀 Feature
Lazily construct planar views of data

## Motivation
The current handling of lazy ndarrays in napari is lazy across planes, but eager within a plane. When displaying data that is chunked in x and y axes, Napari blocks until the entire plane is available as a numpy array. For very large datasets this is less than ideal.

An alternative approach is to display each chunk as it arrives. Once this mechanism is in place, it becomes possible to do fancy stuff like defer loading / rendering chunks conditioned on the current field of view (i.e., don't load stuff the user isn't looking at), etc. 

More concretely, I'm working with isotropic EM data where typical dataset dimensions are `[thousands, thousands, thousands]`, and the data are isotropically chunked. Loading a single plane at full res can take a long time, but we avoid this problem by using [bigdataviewer](https://github.com/bigdataviewer)-derived [tools](https://github.com/saalfeldlab/n5-utils) for looking at the data. These tools render each chunk of the data indpendently, so images start out looking patchy and get filled in over time, and the program basically never blocks.

(Less concretely, I'm imagining using napari for visualizing stuff like lazily defined synthetic data or simulations that might have insane dimensions. Chunked rendering of single planes / views would be essential for this kind of thing)

sofroniewn:
We've got some support recently added for stuff like this in the `Pyramid` layer, see the `examples/add_pyramid.py` example, which will leverage a multiresolution list of arrays (could be from zarr files) to just grab the "tile" you need to see given the resolution and fov of the camera. Right now its only been tested for `2D` but it should be pretty easy to extend to `nD` (though still only looking at slices). You'll have to have the multiresolution pyramid precomputed. Is that ok?

Here's an example zooming in on a 100,000 x 200,000 pathology slice:
![image_pyramid9](https://user-images.githubusercontent.com/6531703/58971550-8ccf6600-8770-11e9-9309-e7256948680c.gif)

sofroniewn:
I've extended our pyramids layer so it will work now with data that is `[thousands, thousands, thousands]` if you can precompute the image pyramids. Let me know if this is helpful or if you need something else

d-v-b:
Thanks! I'll see how well this works for our data (some of our datasets have a precomputed pyramid). But even for data with a pyramid, it might still be helpful to render planar views in a lazy way. 

[...]

PR #323, opened 2019-06-06
Nd pyramids

sofroniewn:
# Description
This PR adds support for nD pyramids to the `Pyramid` layer as requested in #318 by @d-v-b. This layer can now take a list of nD array like objects with the base of the pyramid the first element of the list. The downsampling can uniform or not across the different axes, i.e. it could go

```python
[(32, 2048, 2048), (16, 1024, 1024), (8, 512, 512), (4, 256, 256), (2, 128, 128)]
```
which is uniform downsampling across all axes by a factor of 2
or it could go 
```python
[(10, 2048, 2048), (10, 1024, 1024), (10, 512, 512), (10, 256, 256), (10, 128, 128)]
```
where only the displayed image dimensions are downsamples.
Examples of this are provided in `examples/nD_pyramid.py` and `examples/nD_pyramid_non_uniform.py`.

This PR also cleans up one or two minor things with the thumbnails that were causing errors. Now the thumbnails, and status bar updates etc. should all be working fine for both image and pyramid layers. For a pyramid layer the thumbnail will show the lowest resolution image at the current slice.

Here's the behavior in `examples/nD_pyramid.py` where the z-slice has been mapped to "intensity" so the image gets darker further down in z.

nD_pyramid
 
## Type of change

# How has this been tested?

- [x] `examples/nD_pyramid.py` and `examples/nD_pyramid_non_uniform.py`

## Final checklist:


napari/layers/_image_layer/model.py
```diff
@@ -187,17 +187,10 @@ def _slice_image(self):
...
-            try:
-                if indices[dim] > max_dim_index:
-                    indices[dim] = max_dim_index
-            except TypeError:
-                pass
-
+        indices[:-2] = np.clip(indices[:-2], 0,
+                               np.subtract(self.shape[:-2], 1))
         self._image_view = np.asarray(self.image[tuple(indices)])
+        self._image_thumbnail = self._image_view
```
jni, line 205: I find this to be a surprising bit of code! Why did you alias this?

examples/nD_pyramid.py
```diff
@@ -0,0 +1,37 @@
...
+
+    # add image pyramid
+    viewer.add_pyramid(pyramid, clim_range=[0, 255])
+
+    # Set the view box of the camera to include the whole base image of the
+    # pyramid with a little padding. The view box is a 4-tuple of the x, y
+    # corner position followed by width and height
+    base_shape = pyramid[0].shape[1:]
+    viewer.camera.rect = (-0.1 * base_shape[1], -0.1 * base_shape[0],
+                          1.2 * base_shape[1], 1.2 * base_shape[0])
```
jni: Continuing the [discussion](https://github.com/napari/napari/pull/295/files#r289509749) from the original pyramid PR:

> Actually I'm wary of doing this automatically as there could be times when you add a layer and you don't want the camera to change

I think this is a general problem for napari that we have somehow avoided thus-far. In the case of adding a 2D image, you probably don't want the camera to change *if you've already interacted with the camera*, or at least *if there is already a previous layer on the canvas*. I think *in general*, `add_layer` should set the camera to fully contain the first layer it gets.

That's my proposal anyway. Either way, I think the code above has to go — it's clunky to begin with, but the fact that the coordinates are in xy order makes it even worse! I don't think this is something we want in our examples!

... Although there is a middle road: add a function `viewer.camera.set_view_corners`, which uses data coordinates.

napari/layers/_pyramid_layer/model.py
```diff
@@ -89,23 +91,36 @@ def top_left(self, top_left):
...
-        """Updates the image data from the pyramid based on requested tile
-        and pyramid level
-        """
-        if np.any(self._image_shapes[self.pyramid_level] >
-                  self._max_tile_shape):
-            slices = tuple(slice(self._top_left[i], self._top_left[i] +
-                           self._max_tile_shape[i], 1) for i in range(2))
-            self._image = self.pyramid[self.pyramid_level][slices]
+    def _slice_image(self):
+        """Determines the slice of image from the indices."""
```
jni:
```suggestion
        """Determine the slice of image from the indices."""
```

napari/layers/_pyramid_layer/model.py
```diff
@@ -138,16 +153,17 @@ def get_value(self):
...
         value = self._image_view[tuple(coord[-2:])]
 
         pos_in_slice = (self.coordinates[-2:]
                         + self.translate[[1, 0]] / self.scale[:2])
 
         # Make sure pos in slice doesn't go off edge
-        shape = self._image_shapes[self._pyramid_level]
-        coord = np.clip(pos_in_slice, 0, np.subtract(shape, 1))
-        coord = np.round(coord * self.scale[:2]).astype(int)
+        shape = self._image_shapes[self.pyramid_level][-2:]
```
jni, line 192: Can you add a note above this line that this needs to change with the dims model?

napari/layers/_pyramid_layer/model.py
```diff
@@ -193,7 +210,7 @@ def find_top_left(self):
         transform = self._node.canvas.scene.node_transform(self._node)
         pos = transform.map([0, 0])[:2] + self.translate[:2] / self.scale[:2]
 
-        shape = self._image_shapes[0]
+        shape = self._image_shapes[0][-2:]
```
jni, line 243: As above.

jni (commented):
@sofroniewn this is awesome work! I'm really quite shocked that you've got us to nD pyramids before SciPy. Back in March I envisioned that this would happen in like, a year! :tada: 

There's a few places where I think the code very lackadaisically assumes that the last two dimensions are the ones displayed, which is the case only right now. This is fine, but I suggest marking those with comments to make it easier to find them when the dims model gets fleshed out. 

I've also flagged the camera setting in the examples. Hopefully this can be fixed within this PR.

napari/layers/_image_layer/model.py
sofroniewn: I needed to alias this in the `Image` layer so that it could be defined differently in the `Pyramid` layer (which subclasses the `Image` layer). While in the `Image` layer it makes sense for this to be the same as the `_image_view` in the `Pyramid` layer you want this to correspond to the view for the lowest resolution of the pyramid

napari/layers/_pyramid_layer/model.py
sofroniewn: fixed

napari/layers/_pyramid_layer/model.py
sofroniewn: I have added the following note where it is needed
`# TODO: Change dims selection when dims model changes`

napari/layers/_pyramid_layer/model.py
sofroniewn: see above

examples/nD_pyramid.py
sofroniewn: I can fix this, I will add a `camera_viewbox` property on the base layer and viewer and use events to propogate any changes from it up to the canvas. The `Pyramid` layer will then be able to set the camera when it is added (the viewer could then do this too if preferred)

sofroniewn: actually, that wouldn't have been so good as the property would have gotten out of sync pretty quickly. instead i've adapted the `reset_view` event to now include the shape of the `viewbox` that will make everything look good. the unpleasant lines in all the examples have been removed

sofroniewn:
Thanks @jni. Making it work for 2D slices in nD wasn't so bad, but we might have to revisit all this when it comes to 3D!

I've fixed the camera setting, and this is now ready for re-review

sofroniewn:
This has been updated with the current black formats and is ready for re-review

kne42 (approved):
lgtm

PR #324, opened 2019-06-06
Expand button

### LINKED ISSUE #321: Button to expand layer properties
## 🚀 Feature
Right now you need to double click the layer properties to expand then, but instead we could have a button that either pointed up or down to expand them. I think people would find this a more intuitive UI. Now that we've made the top part of the layer properties independent (in #314) we can easily add the button next to the layer name, and drop support for double clicking. Thoughts?

sofroniewn:
# Description
This PR resolves #321 by adding a button to explicitly expand the layer properties. As requested by @jni it preserves the option to doubleclick elsewhere inside the the layer properties to toggle the expansion.

Right now the PR is still in progress as I need help from our style and icon guru @freeman-lab to improve the icon. I have a "sandwhich bar" in there right now but I think I might switch to up / down triangles. Open to other suggestions too if people have them.

image

## Type of change

# How has this been tested?

- [x] run examples and toggle button

## Final checklist:


sofroniewn:
I've updated the style of the icon based on a design from @freeman-lab. See below in the gif. This PR is now ready for review.

expand_button

napari/resources/build_icons.py
```diff
@@ -51,9 +52,9 @@
         file = icon + '.svg'
         if icon == 'visibility':
             css = insert.replace('{{ color }}', palette['text'])
-        elif icon == 'visibility_off':
+        elif icon in ['visibility_off', 'menu']:
```
kne42, line 56: like a true pythonista :P

kne42 (approved):
the sandwich icon typically signals that this is a modal menu or a drag handle

would prefer to see an arrow that goes from left `<` to down `v`

bonus points if you animate it :)

napari/resources/build_icons.py
sofroniewn: i'm learning!

sofroniewn:
the black formatting has cause this PR to explode :-(

napari/layers/_base_layer/view/properties.py
```diff
@@ -74,16 +75,21 @@ def __init__(self, layer):
...
         sld = QSlider(Qt.Horizontal, self)
         sld.setFocusPolicy(Qt.NoFocus)
         sld.setMinimum(0)
         sld.setMaximum(100)
         sld.setSingleStep(1)
         sld.setValue(self.layer.opacity * 100)
-        sld.valueChanged[int].connect(
-            lambda value=sld: self.changeOpacity(value)
-        )
+        sld.valueChanged[int].connect(lambda value=sld: self.changeOpacity(value))
```
sofroniewn: not sure why the line length changed here @AhmetCanSolak

sofroniewn:
Line length issue with black has been fixed by replacing `max-line-length` with `line-length` in the `pyproject.toml` and the python 3.7 compatibility mentioned in #328 by using `python3` in our `pre-commit-config.yaml`. This is already to go and I will merge once the tests pass

napari/layers/_base_layer/view/properties.py
sofroniewn: fixed by replacing `max-line-length` with `line-length` in the pyproject.toml

PR #329, opened 2019-06-07
Black Format CI task

AhmetCanSolak:
# Description

Black format checking task added to CI configuration, little update made on CONTRIBUTING guideline, excludes moved to `pyproject.toml`.

## Type of change

# References

# How has this been tested?

## Final checklist:


.cirrus.yml
```diff
@@ -1,3 +1,35 @@
+task:
```
kne42, line 1: you should name these tests:
`lint_task` and `test_task`

.cirrus.yml
```diff
@@ -1,3 +1,35 @@
...
+  auto_cancellation: true
+
+  env:
+    matrix:
+      PY_VER: "3.6"
+      PY_VER: "3.7"
+
+  matrix:
```
kne42: this doesn't need a matrix

.cirrus.yml
```diff
@@ -1,3 +1,35 @@
...
+    - name: Black Formatting
+      container:
+        image: gcc:latest
```
kne42: I would use a python image: https://cirrus-ci.org/examples/#python

.cirrus.yml
```diff
@@ -1,3 +1,35 @@
...
+      env:
+        PATH: $HOME/conda/bin:$PATH
+      system_script:
```
kne42: don't need this

.cirrus.yml
```diff
@@ -1,3 +1,35 @@
+task:
+  auto_cancellation: true
+
+  env:
+    matrix:
+      PY_VER: "3.6"
```
kne42: do we really need to check against multiple python versions?

.cirrus.yml
AhmetCanSolak: but that will be reduced same we have right according to docs. Still want?

.cirrus.yml
AhmetCanSolak: maybe not

.cirrus.yml
AhmetCanSolak: true, fixing

.cirrus.yml
AhmetCanSolak: didn't know about, thanks!

.cirrus.yml
AhmetCanSolak: okay removing.

.cirrus.yml
```diff
@@ -1,3 +1,29 @@
+task:
+  auto_cancellation: true
+
+  env:
+      PY_VER: "3.6"
```
kne42: this has no effect on the task and should be removed

.cirrus.yml
```diff
@@ -1,3 +1,29 @@
...
+
+  name: Black Formatting
+  container:
+      image: python:latest
```
kne42: this is where you'd specify which language version we're testing against so maybe you want to pin it to 3.6 as you did above?

.cirrus.yml
```diff
@@ -1,3 +1,29 @@
...
+  env:
+      PATH: $HOME/conda/bin:$PATH
```
kne42: don't need this anymore

.cirrus.yml
```diff
@@ -1,3 +1,29 @@
...
+  conda_script:
```
kne42: this was only here to install python but we're just using a container version so it's no longer necessary and should be removed

.cirrus.yml
```diff
@@ -1,3 +1,29 @@
...
+      - curl https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh > install.sh
+      - bash install.sh -b -p $HOME/conda
+      - conda update -yn base conda
+      - conda install -y python=$PY_VER
+      - rm install.sh
+
+  install_script:
+    # pip >= 19.0 crashes on Linux with Python 3.7
+    - python -m pip install --retries 3 --upgrade "pip<19.0"
+    - pip install --retries 3 -r requirements.txt
```
kne42: we really only need `` pip install `cat requirements/development.txt | grep black` ``

.cirrus.yml
```diff
@@ -1,3 +1,29 @@
...
+
+  test_script:
+    - black -S -l 79 -t py36 --exclude vendored\|_vispy --check .
```
kne42: shouldn't all of these command line options (except `--check`) already be set by the `pyproject.toml` (and therefore autorun when we `black --check`)?

.cirrus.yml
kne42: ah, you're right, nvm

.cirrus.yml
AhmetCanSolak: nope, those are for pre-commit

kne42: https://github.com/python/black#pyprojecttoml

AhmetCanSolak: yeah but we have excludes and stuff, also that line matches our contributing guideline?

kne42: our contributing guidelines are there to tell people what it's equivalent to but i would prefer to cut down on redundancy wherever possible

AhmetCanSolak:
@kne42 i addressed your concern, can you provide a new review?

.cirrus.yml
```diff
@@ -1,3 +1,18 @@
+task:
+  auto_cancellation: true
+
+  name: Black Formatting
+  container:
+      image: python3.6:latest
```
kne42:
```suggestion
      image: python:3.6
```

.cirrus.yml
```diff
@@ -1,3 +1,18 @@
...
+
+  install_script:
+    # pip >= 19.0 crashes on Linux with Python 3.7
+    - python -m pip install --retries 3 --upgrade "pip<19.0"
+    - pip install `cat requirements/development.txt | grep black`
+
+  test_script:
+    - black --exclude vendored\|_vispy --check .
```
kne42: could you add this `--exclude` stuff to `pyproject.toml` and remove it from here? you can find an example setup [here](https://github.com/python/black/blob/master/pyproject.toml#L12)

pyproject.toml
```diff
@@ -2,3 +2,4 @@
 target-version = ['py36', 'py37']
 skip-string-normalization = true
 line-length = 79
+exclude = 'vendored\|_vispy'
```
kne42: test is failing because this somehow did not properly exclude the desired directories

AhmetCanSolak: yeah same command line string not working

kne42 (approved):
Sweet, looks great! 🎉

PR #330, opened 2019-06-09
Refactor layer qt properties and controls

sofroniewn:
# Description
This PR continues our MVC refactoring, resolving #92 by removing the layer `qt_properties` and `qt_controls` from the individual layer model files. These now get created and added to the `Qt` files inside view files. This change steps us closer to being able to have model files that are independent of any `Qt` and can be tested in isolation. The only remaining link now from non-model to model is the `vispy` `SceneCanvas` `node` which is still a property on the layer model files. Once this is removed we will have completed our MVC refactor, and be able to add tests for all our model files, but this will come in follow up PRs.

## Type of change

# How has this been tested?

- [x] run examples

## Final checklist:


napari/components/_layers_list/view.py
```diff
@@ -38,46 +39,62 @@ def __init__(self, layers):
...
         total = len(self.layers)
+
+        # Create list of the current property and divider widgets
+        widgets = [
+            self.vbox_layout.itemAt(i + 1).widget() for i in range(2 * total)
+        ]
+        indices = [
+            self.layers.index(w.layer)
+            for i, w in enumerate(widgets)
+            if i % 2 == 0
```
kne42, line 79: could you explain why you're only taking out evenly-indexed widgets? does this have to do with ignoring the dividers?

napari/layers/_pyramid_layer/__init__.py
```diff
@@ -1 +1,3 @@
 from .model import Pyramid
+from .._image_layer.view import QtImageProperties as QtPyramidProperties
```
kne42, line 2: is this really necessary?

napari/layers/util.py
```diff
@@ -0,0 +1,55 @@
...
+    ----------
+        layer : napari.layers._base_layer.Layer
+            Layer that needs its propetry widget created.
+
+    Returns
+    ----------
+        properties : napari.layers._base_layer.QtLayerProperties
+            Qt propetry widget
+    """
+    name = 'Qt' + type(layer).__name__ + 'Properties'
```
kne42, line 24: ok now I see why...

napari/layers/util.py
```diff
@@ -0,0 +1,55 @@
...
+    if name in globals().keys():
```
kne42: I would just use a `try` statement here

napari/layers/util.py
```diff
@@ -0,0 +1,55 @@
...
+        layer : napari.layers._base_layer.Layer
+            Layer that needs its controls widget created.
+
+    Returns
+    ----------
+        controls : napari.layers._base_layer.QtLayerControls
+            Qt controls widget
+    """
+    name = 'Qt' + type(layer).__name__ + 'Controls'
+    if name in globals().keys():
```
kne42: see above

napari/components/_layers_list/view.py
sofroniewn: yes, have added a comment

napari/layers/util.py
sofroniewn: changed

napari/layers/util.py
sofroniewn: changed

sofroniewn:
I have updated from master and made the fixes requested by @kne42 - this is now ready for rereview

napari/components/_layers_list/view.py
```diff
@@ -38,46 +39,64 @@ def __init__(self, layers):
...
+        """Remove widget for layer at index `event.index`."""
+        layer_index = event.index
+        total = len(self.layers)
+        # Find property widget and divider for layer to be removed
+        index = 2 * (total - layer_index) + 1
+        widget = self.vbox_layout.itemAt(index).widget()
+        divider = self.vbox_layout.itemAt(index + 1).widget()
+        self.vbox_layout.removeWidget(widget)
+        widget.deleteLater()
+        widget = None
```
kne42: why do this now? they will be garbage collected at the end of the function anyways

kne42: speaking of which, `del widget` would be more correct

napari/components/_layers_list/view.py
```diff
@@ -38,46 +39,64 @@ def __init__(self, layers):
...
-            layer = self.layers[i]
-            if layer._qt_properties is not None:
-                index = self.vbox_layout.indexOf(layer._qt_properties)
-                divider = self.vbox_layout.itemAt(index + 1).widget()
-                self.vbox_layout.removeWidget(layer._qt_properties)
+            # Find index of property widget in list of the current layer
+            index = 2 * indices.index(i)
+            widget = widgets[index]
+            divider = widgets[index + 1]
+            # Check if current index does not matche new index
```
kne42: typo: match

napari/layers/_base_layer/model.py
```diff
@@ -50,9 +48,7 @@ class Layer(VisualWrapper, ABC):
 
     def __init__(self, central_node, name=None):
         super().__init__(central_node)
-        self._selected = False
-        self._qt_properties = None
-        self._qt_controls = None
+        self._selected = True
```
kne42, line 51: what prompted this change?

napari/layers/_base_layer/view/properties.py
```diff
@@ -119,7 +119,6 @@ def __init__(self, layer):
         self.setFixedWidth(250)
         self.grid_layout.setColumnMinimumWidth(0, 100)
         self.grid_layout.setColumnMinimumWidth(1, 100)
-        self.layer.selected = True
```
kne42, line 122: oh ok that makes more sense, but why access it via `_selected` instead of `selected`, which would prompt events?

napari/layers/util.py
```diff
@@ -0,0 +1,55 @@
...
+    Returns
+    ----------
+        properties : napari.layers._base_layer.QtLayerProperties
+            Qt propetry widget
+    """
+    name = 'Qt' + type(layer).__name__ + 'Properties'
+    try:
+        qt_props = globals()[name]
+        properties = qt_props(layer)
+    except:
```
kne42: be explicit with what exception you're catching:
```suggestion
    except KeyError:
```

napari/layers/util.py
```diff
@@ -0,0 +1,55 @@
...
+    Returns
+    ----------
+        controls : napari.layers._base_layer.QtLayerControls
+            Qt controls widget
+    """
+    name = 'Qt' + type(layer).__name__ + 'Controls'
+    try:
+        qt_controls = globals()[name]
+        controls = qt_controls(layer)
+    except:
```
kne42: see above

kne42 (approved):
nicely work! a few small questions I have/tiny things to iron out but approved pending resolution

napari/components/_layers_list/view.py
sofroniewn: removed

napari/components/_layers_list/view.py
sofroniewn: fixed

napari/layers/_base_layer/model.py
sofroniewn: it used to be set in the view, but should have been set here instead

napari/layers/_base_layer/view/properties.py
sofroniewn: for most things like this we only use the `_` property, but it might be worth revisiting in general as it means we might be able to clean up some of the setters in the properties files. I'd rather deal with this in follow up PRs though

napari/layers/util.py
sofroniewn: done

napari/layers/util.py
sofroniewn: done

sofroniewn:
Have made fixes, will merge after tests pass

PR #331, opened 2019-06-11
Vispy qtpy and pyside2

sofroniewn:
# Description
This PR allows us to use `PySide2` instead of `PyQt5` for our python qt bindings, which will allow us to release a bundled app under an LGPL instead of GPL license. There should be no other change towards functionality.

Right now you are forced to use `PySide2` by a couple lines in `napari.__init__` that set some environment variables, but it would be pretty easy to let someone use `PyQt5` if they wanted, which might be nice for compatibility with other apps, but we still wouldn't ever want to distribute a bundled app that contained `PyQt5` due to the license issues.

Unfortunately adding a new backend to vispy required copying a whole bunch more files to our `_vispy` folder where, excluding the files in the actual backend folder, all the other changes are just to the import statements so that the files in the backend folder get used. It was hard for me to see how to avoid this, but maybe @kne42 knows a way.

Also major thanks to @kne42 for showing me how to get `PySide2` to find the right `Qt` binaries, that was a huge blocker for a while!

## Type of change

# How has this been tested?

run examples

## Final checklist:


napari/util/app.py
```diff
@@ -1,7 +1,6 @@
 import os
 import sys
 from contextlib import contextmanager
-
```
kne42: why delete this?

napari/_vispy/app/backends/_qt.py
```diff
@@ -0,0 +1,765 @@
...
+
+    QWidget, QApplication = QtGui.QWidget, QtGui.QApplication  # Compat
+elif qt_lib == 'pyqt5':
+    _check_imports('PyQt5')
+    if not USE_EGL:
+        from PyQt5.QtOpenGL import QGLWidget, QGLFormat
+    from PyQt5 import QtGui, QtCore, QtWidgets, QtTest
+
+    QWidget, QApplication = QtWidgets.QWidget, QtWidgets.QApplication  # Compat
+elif qt_lib == 'qtpy':
```
kne42: from what I can tell, the only real changes to vispy are here and line 57, along with the additional qtpy backend file, correct?

napari/components/_viewer/view/main.py
```diff
@@ -26,9 +28,10 @@ def __init__(self, viewer):
         )
 
         self.viewer = viewer
-
-        self.canvas = SceneCanvas(keys=None, vsync=True)
-        self.canvas.native.setMinimumSize(QSize(100, 100))
+        # set vispy application to the qtpy backend
+        app = Application('qtpy')
```
kne42: seems kinda weird to just initialize this here and never use it again. would prefer `use_app` instead

examples/add_image.py
```diff
@@ -10,6 +10,7 @@
 
 
 with app_context():
+
```
kne42: why this

requirements/default.txt
```diff
@@ -3,7 +3,7 @@ numpy>=1.10.0
 qtpy>=1.7.0
 scipy>=1.2.0
 scikit-image>=0.15.0
-vispy>=0.5.3
+vispy==0.5.3
```
kne42, line 6: yikes. I guess we don't really have a choice right now due to all of the mixing of stuff, but the ideal workflow would be:
1. implement the changes in the `vispy` main repo
2. backport it to be compatible with the `0.5.3` code
3. monkeypatch it in `napari` iff `vispy==0.5.3`

kne42 (commented):
Yes, the implementation of this could be a lot cleaner, but it's not exactly super trivial either, so I think it's fine to leave this be for now and clean it up in a later PR that addresses all of these issues. I think it would also be appropriate to make "mini-vispy" with this set of changes

please also rebase with master and revert all changes made by `black` to the vispy code base

I also don't think that this is a high priority issue so I'm fine with letting it sit here for a while

kne42:
also, please verify that the `%gui` magic commands are working properly with `PySide2`

napari/util/app.py
sofroniewn: undone, wasn't super intentional, had made some edits to this file and then undid them, this got mixed up along the way

examples/add_image.py
sofroniewn: not intentional, undone

napari/components/_viewer/view/main.py
sofroniewn: switched to `use_app`

napari/_vispy/app/backends/_qt.py
sofroniewn: yes that's correct, it is a pretty minor change, but we just needed all the other code so that these files would get read

sofroniewn:
I've confirmed the `%gui` magic still works, and updated with the latest changes from master. It's not clear to me how to easily undo the black formatting of the `_vispy` folder without just copying over everything again. Any suggestions?

sofroniewn:
I have changed the PR following suggestions by @kne42 to make it really easy to switch between `PySide2` and `PyQt5` now using just the environment variable `QT_API` which is the same used by `qtpy`. I also now follow the defaults of `qtpy` which will take `PyQt5` first if both libraries are installed and the environment variable isn't working. I've verified both work fine on my machine even when both are installed. I've changed our requirements though to include only `PySide2` so we will be able to make an LGPL app

Finally as noted in #335 there were some problems with black reformating our vispy code when using pre-commit, but this has now been fixed

jni:
@sofroniewn just to be clear, vispy are working on this already? Or should we submit a PR to vispy to add this compatibility upstream?

sofroniewn:
@jni vispy added support for `PySide2` in a PR since their last release. Here I add support for `qtpy` instead (which itself can figure out the difference between `PySide2` & `PyQt5`). To me this is much cleaner and if I were to make a PR to vispy I'd suggest replacing their support for `PySide / PyQt4 / PySide2 / PyQt5` with support for just `qtpy`. Curious what their take would be and if they are considering it already. I am not sure

napari/__init__.py
```diff
@@ -1,16 +1,28 @@
-import warnings
+import os
+from pathlib import Path
+from qtpy import API_NAME
 
-vispy_warning = "VisPy is not yet compatible with matplotlib 2.2+"
+if API_NAME == 'PySide2':
+    # Set plugin path appropriately if using PySide2
```
kne42: this should specify that this is merely a bugfix for when _both_ pyqt and pyside are installed

napari/__init__.py
```diff
@@ -1,16 +1,28 @@
...
+        Path(PySide2.__file__).parent / 'Qt' / 'plugins'
     )
-    from .viewer import Viewer
+elif API_NAME == 'PyQt5':
+    # Set plugin path appropriately if using PyQt5
+    import PyQt5
 
-from ._view import view
+    os.environ['QT_PLUGIN_PATH'] = str(
+        Path(PyQt5.__file__).parent / 'Qt' / 'plugins'
```
kne42: this entire clause is unnecessary; we only needed the one above because it was to resolve a bug

napari/__init__.py
```diff
@@ -1,16 +1,28 @@
...
+    )
+else:
+    raise (ValueError('qtpy backend must either be PyQt5 or PySide2'))
```
kne42: i dont think we should enforce which backend people are using; at most give a warning

we should, however, enforce a minimum version requirement so this should be done here

napari/_view.py
```diff
@@ -47,7 +47,9 @@ def view(
     :class:`napari.Viewer` class (preferred), or update the layers directly on
     the returned :class:`napari.Viewer` object.
     """
+
```
kne42: why this?

.pre-commit-config.yaml
```diff
@@ -5,3 +5,4 @@ repos:
     - id: black
       language_version: python3
       require_serial: true
+      exclude: 'vendored|\_vispy'
```
kne42: this line should be deleted

pyproject.toml
```diff
@@ -2,4 +2,4 @@
 target-version = ['py36', 'py37']
 skip-string-normalization = true
 line-length = 79
-exclude = 'vendored|_vispy'
+exclude = 'vendored|\_vispy'
```
kne42: this should go in a separate PR and also include other commonly ignored files e.g. `build`, `dist`, `.ipynb_checkpoints`, etc.

.pre-commit-config.yaml
sofroniewn: I needed to add this line because otherwise the folders would get reformatted as detailed in #335 - I am understanding this wrong?

I'm not sure why its `\_vispy` instead of `_vispy` I don't remember doing that

pyproject.toml
sofroniewn: As above, I'm not sure why its `\_vispy` instead of `_vispy` I don't remember doing that and can remove it.

I think black might already be ignoring files in our `.gitignore`?

napari/__init__.py
sofroniewn: done

napari/__init__.py
sofroniewn: removed

napari/__init__.py
sofroniewn: I don't understand `enforce a minimum version requirement` but not `which backend`? Should I just remove this error entirely?

.pre-commit-config.yaml
sofroniewn: removed from this PR

pyproject.toml
sofroniewn: removed from this PR

napari/_view.py
sofroniewn: unintentional, removed

napari/__init__.py
kne42: the error itself should be removed

we _should not_ error if a backend other than PyQt5 or PySide2 is used

we _should_ error if the Qt version does not meet our minimum requirements

sofroniewn: sounds good, I have removed this error

sofroniewn:
I have made the fixes requested by @kne42 this is ready for re-review

README.md
```diff
@@ -12,13 +12,13 @@
 [![Development Status](https://img.shields.io/pypi/status/napari.svg)](https://github.com/napari/napari)
 [![Code style: black](https://img.shields.io/badge/code%20style-black-000000.svg)](https://github.com/python/black)
 
-**napari** is a fast, interactive, multi-dimensional image viewer for Python. It's designed for browsing, annotating, and analyzing large multi-dimensional images. It's built on top of `PyQt` (for the GUI), `vispy` (for performant GPU-based rendering), and the scientific Python stack (`numpy`, `scipy`).
+**napari** is a fast, interactive, multi-dimensional image viewer for Python. It's designed for browsing, annotating, and analyzing large multi-dimensional images. It's built on top of `PySide2` (for the GUI), `vispy` (for performant GPU-based rendering), and the scientific Python stack (`numpy`, `scipy`).
```
kne42: this should be generalized to just `Qt`

napari/__init__.py
```diff
@@ -1,16 +1,20 @@
...
+    # for when both PyQt5 and Pyside2 are installed
+    import PySide2
 
-with warnings.catch_warnings():
-    warnings.filterwarnings(
-        "ignore", category=UserWarning, message=vispy_warning
+    os.environ['QT_PLUGIN_PATH'] = str(
+        Path(PySide2.__file__).parent / 'Qt' / 'plugins'
     )
-    from .viewer import Viewer
```
kne42, line 14: still would like to see something checking that the imported library's version is `>=5.12.3`

README.md
sofroniewn: fixed

napari/__init__.py
sofroniewn: have now added a check

sofroniewn:
not sure why we're now getting the segfaults on linux :-(

kne42:
strange – but it seems to consistently fail on `nD_pyramid`

sofroniewn:
This PR is superseded by #412 following the vispy `0.6` release

PR #340, opened 2019-06-14
fix labels colormap

### LINKED ISSUE #339: Labels colormap wrong
## 🐛 Bug

The labels colormap is now broken - I think from merging #330 - as it now renders with a different colormap and not the desired one. It's pretty unclear to be how this PR broke the functionality (or if it broke somewhere else), but it should be fixed!

Maybe @jni can look into it too

sofroniewn:
# Description
This PR fixes #339 by directly setting the labels colormap instead of relying on an event triggering an update in the layer properties (which are now constructed after the layer, since #330) 

## Type of change

# How has this been tested?

- [x] run `examples/labels-0-2d.py` and see colormap is correct again

## Final checklist:

### DIFF
--- a/napari/layers/_labels_layer/model.py
+++ b/napari/layers/_labels_layer/model.py
@@ -88,10 +88,10 @@ class Labels(Layer):
         self._need_display_update = False
         self._need_visual_update = False
 
-        self.events.opacity.connect(lambda e: self._update_thumbnail())
-
         self._node.clim = [0.0, 1.0]
-        self.events.colormap()
+        self._node.cmap = self.colormap
+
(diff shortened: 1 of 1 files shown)


kne42 (approved):
lgtm – only question is if this should be propagated back to the image layer

kne42:
merging anyways so we can make a `0.09` release and if the answer is a yes, we can make a follow up PR

sofroniewn:
I don't think anything on the Image Layer is broken right now and we're going to refactor how the `_node` is on the Layer model files asap anyway so I think it's fine to leave till then. Thanks for merging!

PR #341, opened 2019-06-14
fix black ignore

### LINKED ISSUE #335: black --exclude in precommit being ignored
## 🐛 Bug

Our precommit hook is ignoring our excludes in our `pyproject.toml` and reformatting our `_vispy` files. This is known behaviour as described here https://github.com/python/black/issues/395 (and other places referenced there in) and it seems like the solution is to move the excludes to the `.pre-commit-config.yaml` which I will try, but I wanted to flag here in case @AhmetCanSolak or @kne42 have any further thoughts

sofroniewn:
# Description
This PR fixes #335 by adding an extra exclude to our `.pre-commit-config.yaml`

## Type of change

# How has this been tested?

- [x] run examples

## Final checklist:

### DIFF
--- a/napari/_vispy/__init__.py
+++ b/napari/_vispy/__init__.py
@@ -1,5 +1,3 @@
 """
 Backported/custom features of VisPy.
-"""
-
-3    
+""" 
--- a/napari/_vispy/__init__.py
+++ b/napari/_vispy/__init__.py
@@ -1,3 +1,5 @@
 """
 Backported/custom features of VisPy.
 """
+
+3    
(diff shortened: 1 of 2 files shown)


napari/_vispy/__init__.py
```diff
@@ -1,3 +1,3 @@
 """
 Backported/custom features of VisPy.
-"""
+"""
```
kne42: remove this

kne42 (approved):

napari/_vispy/__init__.py
sofroniewn: done

Issue #310, opened 2019-05-29
Cursor coordinates are off by 0.5


jni:
## 🐛 Bug

The coordinates at the *center* of pixel [0, 0] should be [0, 0]. When looking at a tiny image, hovering over the pixels and looking at the value in the status bar, it is clear that the cursor coordinates are offset by 0.5 along each axis relative to the image:

![pix2](https://user-images.githubusercontent.com/492549/58534781-86e8fa80-822f-11e9-800c-2730620f9418.png)
![pix1](https://user-images.githubusercontent.com/492549/58534784-89e3eb00-822f-11e9-95ca-01aedb4a6003.png)
![pix3](https://user-images.githubusercontent.com/492549/58534805-9b2cf780-822f-11e9-822a-6826b20137a4.png)
![pix5](https://user-images.githubusercontent.com/492549/58534815-9ec07e80-822f-11e9-8e0b-b7ddee19a7ed.png)

## To Reproduce

```python
import numpy as np
import napari
import napari.util


image = np.array([[0, 0, 0, 1],
                  [0, 0, 1, 1],
                  [1, 0, 0, 0],
                  [0, 1, 0, 0]], dtype=float)

with napari.util.app_context():
    napari.view(image, multichannel=False)  # check that the intensity matches the cursor pos
```

## Expected behavior

With nearest-neighbor / no interpolation, the cursor should display the coordinates of the nearest pixel and the value should be what is displayed under the cursor.

## Environment

 - napari Version (e.g., 1.0): master
 - OS (e.g., Linux): Linux
 - Python version: 3.6
 - Any other relevant information:

## Additional context

We had some debates about whether to offset by 0.5 or not. I think there are many reasons why the centers of pixels should be at integer coordinates. My two leading ones are:

- if I put a marker smack in the middle of a pixel, I expect the coordinates of that marker to be integer values, not offset by 0.5.
- interpolation reduces to indexing when doing so on an integer grid.

sofroniewn:
Sounds reasonable, I can look into fixing this and making sure it has no unintended consequences for the markers and shapes layer, where we're already doing some `0.5` offsetting

sofroniewn:
Note how this appears in matplotlib. We will want to align with this convention across all our layer types.

![image](https://user-images.githubusercontent.com/6531703/60206819-5c2b9b00-9809-11e9-948b-ea791ca9cae8.png)

jni:
[...]

PR #343, opened 2019-06-18
Nd vectors

### LINKED ISSUE #262: Support for nD Vectors
## 🚀 Feature
For the vectors layer we want to support a fully nD model with projections occurring over non-displayed dimensions. For this we will switch to an internal representation of an N x 2 x D array which displays N vectors with start and end point in D dimensions. This representation could be passed to the `Shapes` layer, but within the `Vectors` layer special computations can be used to accelerate display as the type of data is know.

The `Vectors` layer will accept either an N x 2 x D array or an (N1 x N2 ... x ND) x D array where each "pixel" contains a length D vector with the D components of the projection of a line onto the D axes. In the later case (N1 x N2 ... x ND) vectors will be drawn.

Does this sound reasonable @bryantChhun?

sofroniewn:
# Description
This PR resolves #262 by adding support for nD vectors. After looking at implementations and using it a little I decided to follow the original proposition of @jni and take projections onto the displayed 2D plane and display a vector of that length only from the original starting point. It's quite simple from an implementation perspective and does the right thing if all vectors are defined inside 2D planes. As the vectors are now specified by a starting point and projection this feels right to me. Length and width are in fact variables that can be adjusted in the gui.

The add_vectors now accepts either `(N, 2, D)` or `(N1, N2, ..., ND, D)` arrays where an `(N, 2, D)` array is interpreted as "coordinate-like" data and a list of N vectors with start point and projections of the vector in D dimensions. An `(N1, N2, ..., ND, D)` array is interpreted as "image-like" data where there is a length D vector of the projections at each pixel. Internally the `(N, 2, D)` representation is used. This is a breaking change from the previous API that accepted `(N, 4)` lists of coords.

I've also changed how the averaging code worked under the hood, now using `ndimage.convolve`, and I no longer do the downsampling, though if @bryantChhun feels strongly about this we can make that an option.

I've also added two examples, `nD_vectors.py` and `nD_vectors_image.py`.
Here's how `nD_vectors.py` looks:

nD_vectors

## Type of change

# How has this been tested?

- [x] `example/add_vectors.py`, `example/add_vectors_image.py`, `example/nD_vectors.py`, and `example/nD_vectors_image.py`

## Final checklist:


examples/nD_vectors.py
```diff
@@ -0,0 +1,49 @@
+"""
+Display one 4-D image layer using the add_image API
```
jni: Update this description

examples/nD_vectors_image.py
```diff
@@ -0,0 +1,41 @@
+"""
+This example generates an image of vectors
+Vector data is an array of shape (N, M, 2)
```
jni: Update to nD

examples/nD_vectors_image.py
```diff
@@ -0,0 +1,41 @@
...
+
+import numpy as np
+
+with app_context():
+    # create the viewer and window
+    viewer = napari.Viewer()
+
+    n = 40
+    m = 60
+    p = 80
```
jni, line 19: I suggest "m, n, p".

examples/to_svg.py
```diff
@@ -104,17 +104,19 @@
...
-    pos[:, 0] = radius_space * np.cos(phi_space) + 256
-    pos[:, 1] = radius_space * np.sin(phi_space) + 256
+    pos[:, 0, 0] = radius_space * np.cos(phi_space) + 350
+    pos[:, 0, 1] = radius_space * np.sin(phi_space) + 256
+
     # assign x-y projection
-    pos[:, 2] = 2 * radius_space * np.cos(phi_space)
-    pos[:, 3] = 2 * radius_space * np.sin(phi_space)
+    pos[:, 1, 0] = 2 * radius_space * np.cos(phi_space)
+    pos[:, 1, 1] = 2 * radius_space * np.sin(phi_space)
```
jni, line 119: Perhaps we should add a `napari.data` module that includes stuff like this? It's kind of annoying to see it repeated (and thus needing updating) everywhere.

napari/layers/_vectors_layer/model.py
```diff
@@ -66,62 +64,32 @@ def __init__(
...
         Parameters
         ----------
-        vectors : np.ndarray
+        vectors : (N, 2, D) array
         """
-        self._original_data = vectors
-        self._current_data = vectors
 
-        self._vectors = self._convert_to_vector_type(self._current_data)
+        self._data = self._convert_to_vector_type(vectors)
```
jni: Could you pull this out as a function, e.g. `vector_image_to_collection`? I don't see a reason why it should live in the class.

napari/layers/_vectors_layer/model.py
```diff
@@ -66,62 +64,32 @@ def __init__(
...
 
-        vertices, triangles = self._generate_meshes(self._vectors, self.width)
+        vertices, triangles = self._generate_meshes(
+            self._data, self.width, self.length
+        )
```
jni, line 81: Total side note: how hard would it be to make arrows instead of lines? It's kinda weird that the visual is not oriented in any way.

napari/layers/_vectors_layer/model.py
```diff
@@ -133,109 +101,87 @@ def _convert_to_vector_type(self, vectors):
...
         ----------
-        vectors : np.ndarray
+        vectors : (N, 2, D) or (N1, N2, ..., ND, D) array
+            A (N, 2, D) array is interpreted as "coordinate-like" data and a list
+            of N vectors with start point and projections of the vector in D
+            dimensions. A (N1, N2, ..., ND, D) array is interpreted as
+            "image-like" data where there is a length D vector of the
+            projections at each pixel.
+
+        Returns
```
jni: Add underline

napari/layers/_vectors_layer/model.py
```diff
@@ -261,40 +207,17 @@ def _update_avg(self):
...
-                x_offset : range_x - x_offset : x,
-                y_offset : range_y - y_offset : y,
-            ]
+            if self.averaging == 1:
+                # use original data
+                self.data = self._raw_data
+            else:
+                # average original data
+                size = (self.averaging,) * self.ndim + (1,)
+                kernal = np.ones(size) / np.product(size)
```
jni: kernel

jni: `kernel = np.full(size, 1 / self.averaging ** self.ndim)`

napari/layers/_vectors_layer/model.py
```diff
@@ -443,13 +332,26 @@ def _set_view_slice(self):
         """Sets the view given the indices to slice with."""
 
         vertices = self._mesh_vertices
-        faces = self._mesh_triangles
+
+        if len(self.data) == 0:
+            faces = []
+        elif self.ndim > 2:
+            matches = np.all(self.indices[:-2] == self.data[:, 0, :-2], axis=1)
+            matches = np.where(matches)[0]
```
jni, line 226: what are "matches?"

napari/layers/_vectors_layer/model.py
```diff
@@ -443,13 +332,26 @@ def _set_view_slice(self):
...
+                faces = self._mesh_triangles[keep_inds]
+        else:
+            faces = self._mesh_triangles
 
         if len(faces) == 0:
             self._node.set_data(vertices=None, faces=None)
         else:
             self._node.set_data(
-                vertices=vertices[:, ::-1], faces=faces, color=self.color
+                vertices=vertices[:, ::-1] + 0.5, faces=faces, color=self.color
```
jni, line 240: I think this is related to the cursor status bug I reported in #310, and probably the wrong way to deal with it...? ie you are curing the symptom here instead of the cause (that the image pixels are offset).

jni (commented):
@sofroniewn I've requested a few changes, one of which is serious.

napari/layers/_vectors_layer/model.py
jni: :point_up: this is the serious one LOL, sorry for the vague comment!

examples/nD_vectors.py
sofroniewn: done

examples/nD_vectors_image.py
sofroniewn: done

examples/nD_vectors_image.py
sofroniewn: fixed

examples/to_svg.py
sofroniewn: agreed - but can that come in a separate PR

napari/layers/_vectors_layer/model.py
sofroniewn: it's easy to make arrowheads (that can be toggled on / off), but lets wait for a followup PR

napari/layers/_vectors_layer/model.py
sofroniewn: done

napari/layers/_vectors_layer/model.py
sofroniewn: fixed

napari/layers/_vectors_layer/model.py
sofroniewn: Vectors in the currently viewed plane. I've added a comment

napari/layers/_vectors_layer/model.py
sofroniewn: Yeah, so this is also happening in the `Points` layer here https://github.com/napari/napari/blob/961cb2cad10aab7228aa1cec8a443a7e5bd004df/napari/layers/_points_layer/model.py#L386
and is likely related to the issue in #310, but I'd much rather systematically fix this in a follow-up PR addressing just this stuff then try and add this on here. It will also affect the painting in the labels layer and the addition of polygons in the shapes layer

napari/layers/_vectors_layer/model.py
sofroniewn: I have pulled this out into a util folder

sofroniewn:
I have updated this PR with the fixes requested from @jni, though see my comment about the "hard one"

I've also removed the averaging functionality as discussed in meeting, which makes things quite a bit simpler for the layer.

Can @jni and @bryantChhun re-review

jni:
@sofroniewn the build is failing, but other than that I'm happy with the changes/comments!

jni (approved):
Approved conditional on tests passing!

sofroniewn:
thanks, failed test was just the doc string matching, have fixed now

jni:
Thanks to @kne42 again for pushing for that test! =)

PR #345, opened 2019-06-18
remove broadcast from shapes

### LINKED ISSUE #344: remove broadcast mode from nD shapes
## 🐛 Bug

#297 introduced nD shapes containing both `broadcast` and `sliced` shapes, with the option to toggle between them, but really looking back on it that was a mistake. As some people pointed out at the time - being able to switch back and forth between the two modes was strange. In reality it was unnecessary - `broadcast` mode is achieved by adding a `2D` shapes layer to the viewer. There was no need to have the concept of and `nD` shape that appeared everywhere.

For a really simple example of how this works see the nD vectors example in #343. 

I'd like to remove the explicit broadcast mode, which will preserve all functionality and reduce confusion

sofroniewn:
# Description
This PR fixes #344 by removing the broadcast mode from shapes. 2D shape layers are automatically broadcast across the nD viewer and so the extra mode was confusing at best. The only downside, as pointed out by @jni is that now it is no longer possible to add a 2D shapes layer from the GUI when an nD image is present. We should probably fix that another way though

## Type of change

# How has this been tested?

- [x] run `example/add_shapes.py` and `example/nD_shapes.py`

## Final checklist:

### DIFF
--- a/napari/layers/_shapes_layer/model.py
+++ b/napari/layers/_shapes_layer/model.py
@@ -218,6 +218,8 @@ class Shapes(Layer):
             )
 
             if len(self._data) > 1:
+                # If multiple ShapeLists present and 2D slice key still there
+                # from initialization then remove 2D slice key
                 if () in self._data:
                     del self._data[()]
 
@@ -525,8 +527,6 @@ class Shapes(Layer):
             return (1,) * (self._input_ndim - 2) + slice_shape
         else:
             slice_keys = list(self.data.keys())
-            if () in slice_keys:
-                slice_keys.remove(())
             max_val = np.array(slice_keys).max(axis=0)
             return tuple(max_val) + slice_shape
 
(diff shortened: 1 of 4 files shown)


napari/layers/_shapes_layer/model.py
```diff
@@ -227,6 +217,12 @@ def __init__(
                 z_index=z_index,
             )
 
+            if len(self._data) > 1:
+                # If multiple ShapeLists present and 2D slice key still there
+                # from initialization then remove 2D slice key
```
jni: I don't understand this comment at all, could you elaborate?

jni (approved):
Very nice! I love it when you can clean up so many lines of code!

napari/layers/_shapes_layer/model.py
sofroniewn: i fixed this, had to do with how things were initialized, but i've made it much better now i think

PR #347, opened 2019-06-18
fix layer select styling

### LINKED ISSUE #346: styling on layer widgets on new layer
## 🐛 Bug

When a new layer gets added the styling of the new layer does not reflect that it is selected, but the control panels display and functionality indicates that it is

sofroniewn:
# Description
This PR fixes #346 by setting the stylings on newly created layers to be selected. This was broken in #330

## Type of change

# How has this been tested?

- [x] run examples and look at layer selection

## Final checklist:

### DIFF
--- a/napari/layers/_base_layer/view/properties.py
+++ b/napari/layers/_base_layer/view/properties.py
@@ -120,6 +120,8 @@ class QtLayerProperties(QFrame):
         self.grid_layout.setColumnMinimumWidth(0, 100)
         self.grid_layout.setColumnMinimumWidth(1, 100)
 
+        self._on_select(None)
+
     def _on_select(self, event):
         self.setProperty('selected', True)
         self.nameTextBox.setEnabled(True)
--- a/napari/layers/_base_layer/view/properties.py
+++ b/napari/layers/_base_layer/view/properties.py
@@ -21,8 +21,8 @@ class QtLayerProperties(QFrame):
         super().__init__()
 
         self.layer = layer
-        layer.events.select.connect(self._on_select)
-        layer.events.deselect.connect(self._on_deselect)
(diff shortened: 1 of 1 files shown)


napari/layers/_base_layer/view/properties.py
```diff
@@ -120,6 +120,8 @@ def __init__(self, layer):
         self.grid_layout.setColumnMinimumWidth(0, 100)
         self.grid_layout.setColumnMinimumWidth(1, 100)
 
+        self._on_select(None)
```
kne42: isn't the real problem caused by _all_ of the events not triggering? (see https://github.com/napari/napari/pull/330#discussion_r293176737)

kne42 (changes requested):

napari/layers/_base_layer/view/properties.py
sofroniewn: as discussed in person, the problem is due to the `view` being constructed after the `model` and not being fully initialized properly. adding the event back in doesn't fix that.

I do appreciate that this might not be the nicest fix so an open to alternative suggestions, but it does get the job done

kne42: really dont like the passing `None` as an argument thing. i would just make it a function that takes no parameters and is passed to the event system as `lambda v: self._on_select`

furthermore, this assumes that the layer will always be selected at initialization

i would instead make a setter-type function that replaces `on_select` and `on_deselect` and plug that back into the event system e.g. `self.events.select.connect(lambda v: self.setSelected(True))`

sofroniewn: I have now done this

sofroniewn:
@kne42 this is ready for re-review, with a new implementation as described in your comment

kne42 (approved):
nice!

Issue #349, opened 2019-06-19
Precision loss warning from thumbnails


kevinyamauchi:
## 🐛 Bug
Converting image to uint8 for the thumbnails raises a precision loss warning:

```python
/Users/kevin.yamauchi/Documents/MyNapari/napari/.venv/lib/python3.6/site-packages/skimage/util/dtype.py:135: UserWarning: Possible precision loss when converting from float64 to uint8
  .format(dtypeobj_in, dtypeobj_out))
```

## To Reproduce

1. Run `/examples/add_image.py`
2. The following warning is printed to the terminal
```python
/Users/kevin.yamauchi/Documents/MyNapari/napari/.venv/lib/python3.6/site-packages/skimage/util/dtype.py:135: UserWarning: Possible precision loss when converting from float64 to uint8
  .format(dtypeobj_in, dtypeobj_out))
```

## Expected behavior
The warning should be suppressed

## Environment

 - napari Version (latest commit): 244c818
 - OS (e.g., Linux): OS X
 - Python version: 3.6

kevinyamauchi:
It looks like this will be resolved by https://github.com/scikit-image/scikit-image/pull/3575. @jni, do you know if this will be merged soon?

Otherwise, we can just catch/ignore it with `warnings`.

jni:
@kevinyamauchi I'll make another push on it.

kevinyamauchi:
Awesome! Thanks, @jni!

jni:
It's merged! =D If you start working with skimage master you won't see the warnings anymore. Unfortunately it'll probably be a couple of months before 0.16 is out though.

kevinyamauchi:
Amazing! Thanks for the quick turnaround, @jni!

sofroniewn:
This is fixed in #370, and I'm closing the issue, though once skimage `0.16` comes out we could remove our warnings catch

PR #352, opened 2019-06-20
add points thumbnail

sofroniewn:
# Description
This PR adds a thumbnail to the points layer. It does this my first downsampling the points and placing them in a grid the size of the thumbnail that is determined by their location within the points layer. This means that each point maps to at least one pixel in the 32x32 thumbnail - so no points are missing, but nearby points might get mapped to the same pixel and appear as one dot. Also the placement within the thumbnail is determined by the min / max of the points layer, so adding points at the extremes of the points layer can cause all the points to shift position, and there is always one point located in the corner of the thumbnail if there are more that one point present. In practice I find this behavior ok.

Also the `size` parameter of the points is ignored when rendering the thumbnail, they are all treated equally as exactly one pixel in size. Finally the edge color of the point is also ignored, only the face_color is used, which I think is fine too, though I will note that dark colors (even blue) are hard to see.

Here is what it looks like in practice:

points_thumbnails

Thoughts @kevinyamauchi?
 
## Type of change

# How has this been tested?

- [x] `examples/add_points.py`, `examples/nD_points.py`

## Final checklist:

### DIFF
--- a/napari/layers/_points_layer/model.py
+++ b/napari/layers/_points_layer/model.py
@@ -3,7 +3,6 @@ from xml.etree.ElementTree import Element
 import numpy as np
 from scipy import ndimage as ndi
 from skimage.util import img_as_ubyte
-from copy import copy
 from .._base_layer import Layer
 from ..._vispy.scene.visuals import Markers as MarkersNode
 from ...util.event import Event
@@ -436,6 +435,7 @@ class Points(Layer):
         """Update thumbnail with current points and colors.
         """
         colormapped = np.zeros(self._thumbnail_shape)
+        colormapped[..., 3] = 1
         if len(self._points_view) > 0:
             min_vals = [self.range[-2][0], self.range[-1][0]]
             shape = np.ceil(
--- a/napari/layers/_points_layer/model.py
+++ b/napari/layers/_points_layer/model.py
@@ -448,6 +448,9 @@ class Points(Layer):
             coords = np.floor(
                 (self._points_view - min_vals + 0.5) * zoom_factor
             ).astype(int)
+            coords = np.clip(
+                coords, 0, np.subtract(self._thumbnail_shape[:2], 1)
+            )
             for c in coords:
                 colormapped[c[0], c[1], :] = Color(self.face_color).rgba
         colormapped[..., 3] *= self.opacity
--- a/napari/layers/_points_layer/model.py
+++ b/napari/layers/_points_layer/model.py
@@ -1,6 +1,9 @@
 from typing import Union
 from xml.etree.ElementTree import Element
 import numpy as np
+from scipy import ndimage as ndi
+from skimage.util import img_as_ubyte
+from copy import copy
 from .._base_layer import Layer
 from ..._vispy.scene.visuals import Markers as MarkersNode
 from ...util.event import Event
@@ -90,6 +93,8 @@ class Points(Layer):
             # update flags
             self._need_display_update = False
             self._need_visual_update = False
+
+        self.events.opacity.connect(lambda e: self._update_thumbnail())
 
     @property
     def coords(self) -> np.ndarray:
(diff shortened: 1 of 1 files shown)


napari/layers/_points_layer/model.py
```diff
@@ -426,6 +431,29 @@ def get_message(self, coord, value):
...
+            min_vals = [self.range[-2][0], self.range[-1][0]]
+            shape = np.ceil(
+                [
+                    self.range[-2][1] - self.range[-2][0] + 1,
+                    self.range[-1][1] - self.range[-1][0] + 1,
+                ]
+            ).astype(int)
+            zoom_factor = np.divide(self._thumbnail_shape[:2], shape).min()
+            coords = np.floor(
+                (self._points_view - min_vals + 0.5) * zoom_factor
```
jni, line 449: Why use np.floor instead of np.round?

jni (approved):
Just made one minor comment

napari/layers/_points_layer/model.py
sofroniewn: I'm worried about the coords of a point getting rounded up and being equal to the shape of the thumbnail, which means I'll get an indexing error when I try and set that value

jni: This feels like a job for np.clip *and* a test. ;)

sofroniewn: I've added the `np.clip` to be sure, not sure what additional purpose the test serves!

jni: It means that when I come around 6 months from now and go "Hmm, this np.clip seems pointless, imma remove it to make the code prettier", I'll know immediately that I'm an idiot. =)

I think we're noticing with our bug whack-a-mole that we are at the stage where extensive testing is sorely needed — seemingly unrelated changes can end up breaking the code.

sofroniewn: agreed we should add more tests. once we've pulled out the vispy code from our layer model files we'll be able to more easily test our models and the functions inside them

napari/layers/_points_layer/model.py
```diff
@@ -426,6 +431,32 @@ def get_message(self, coord, value):
...
+            ).astype(int)
+            coords = np.clip(
+                coords, 0, np.subtract(self._thumbnail_shape[:2], 1)
+            )
+            for c in coords:
+                colormapped[c[0], c[1], :] = Color(self.face_color).rgba
+        colormapped[..., 3] *= self.opacity
```
kevinyamauchi, line 456: Do you think it's necessary to modify the opacity of the markers in the thumbnail? Since the thumbnail is for the layer they aren't blended with anything else. I think there would be some opacity/face color/display brightness combinations that would make the markers hard to see and the thumbnail would look black. 

On the other hand, it would be kind of neat to see the opacity change in the thumbnail...

napari/layers/_points_layer/model.py
```diff
@@ -426,6 +431,32 @@ def get_message(self, coord, value):
...
+        colormapped = img_as_ubyte(colormapped)
```
kevinyamauchi, line 457: I think we should suppress the precision loss warnings for now until skimage 0.16 is released (#349).

kevinyamauchi (commented):
This looks good to me! Just a couple of small comments below.

napari/layers/_points_layer/model.py
sofroniewn: agreed - fine to do in a followup PR - I see you just made #355

napari/layers/_points_layer/model.py
sofroniewn: The reason I did this was for the labels layer it's less confusing to have the opacity of the thumbnail match the opacity of the layer as the colors themselves then look identical. Otherwise the colors in the thumbnail look very saturated. We could change, but I think it should be standard across all layer types

PR #353, opened 2019-06-20
fix resource compiling instructions

### LINKED ISSUE #350: update contributing guidelines for compiling resources
## 📚 Documentation

As pointed out by @csweaver following the instructions in our contributing guidelines for installing `pyrcc5` does not actually lead to the install of `pyrcc5`. It seems like this comes with `PyQt5`. There is also the option to use `pyside2-rcc`  which comes with `PySide2`. We should document both these usage options.

sofroniewn:
# Description
This PR fixes #350 by updating our contributing guidelines explains how to compile resources if you are using either `PySide2` or `PyQt5`

Any thoughts @csweaver @jni? 

## Type of change


jni (approved):

CONTRIBUTING.md
```diff
@@ -70,15 +70,22 @@ This creates separate folders in `napari/resources/icons` with full colored icon
 
 Make sure that the file `napari/resources/icons/res.qrc` has a list of all the icons with the correct prefix for each of our themes, e.g. `<file>icons/light/select.svg</file>`, `<file>icons/dark/select.svg</file>`, etc.
 
-Finally, compile all the icons for use in the app by calling.
+Finally, compile all the icons for use in the app by calling either
+
+```sh
```
csweaver, line 75: Suggest adding comment to the example calls to show which is pyqt5 and which is pyside2 for clarity.

sofroniewn: will fix in a follow-up PR

PR #355, opened 2019-06-20
Ignore precision loss warnings from thumbnails

kevinyamauchi:
# Description
This PR ignores the precision loss warnings from skimage when converting the thumbnail image to uint8.

## Type of change

# References
This is a temporary fix for #349 until skimage 0.16 is released.

# How has this been tested?
- [ ] `/examples/add_image.py`
- [ ] `/examples/labels-2d.py`

## Final checklist:

### DIFF
--- a/examples/add_pyramid.py
+++ b/examples/add_pyramid.py
@@ -1,6 +1,7 @@
 """
 Displays an image pyramid
 """
+import warnings
 
 from skimage import data
 from skimage.util import img_as_ubyte
@@ -17,7 +18,10 @@ base = np.tile(astronaut, (16, 16))
 pyramid = list(
     pyramid_gaussian(base, downscale=2, max_layer=5, multichannel=False)
 )
-pyramid = [img_as_ubyte(p) for p in pyramid]
+# ignore skimage precision loss warning
+with warnings.catch_warnings():
+    warnings.simplefilter("ignore")
+    pyramid = [img_as_ubyte(p) for p in pyramid]
 print('pyramid level shapes: ', [p.shape for p in pyramid])
 
 with app_context():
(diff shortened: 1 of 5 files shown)


sofroniewn (changes requested):
I think I'd rather do this when we import the `Viewer` but just catch this specific warning as otherwise we'll need to do this in all the individual layers.

See how we catch the vispy warning in the main init right now https://github.com/napari/napari/blob/764b7473a05c8545fb28b233df48938d256b489b/napari/__init__.py#L5

does @kne42 have thoughts?

kevinyamauchi:
I am totally fine to do it that way @sofroniewn as long as there aren't going to be cases where we want to let the warning through.  I can't think of any at the moment, so I am happy to catch the warning at `Viewer` import unless anybody has other suggestions.

jni:
> as long as there aren't going to be cases where we want to let the warning through

those warnings are going to go away altogether very soon, so I don't think any such cases are relevant. =)

kevinyamauchi:
I looked into this a bit more and I don't think we can suppress the warning on import as we do with the vispy error in `Viewer` because the warning is generated when the thumbnail is generated. Thus, I propose the following:

1. Move the `img_as_ubyte()` conversion present in each individual layer's thumbnail into the setter method of the `thumbnail` property in the base `Layer`. 
2. Catch the warning in the `Layer.thumbnail` setter.

I think this satisfies your intent @sofroniewn, as we don't have to chase down the warnings in each layer. Thoughts?

sofroniewn:
That's a great suggestion @kevinyamauchi - much cleaner to have all the conversions in the base class anyway. I say go ahead and implement

kevinyamauchi:
Sounds good. I think I'll make a clean branch then implement.

kevinyamauchi:
Implemented the above suggestion in #370.

Issue #239, opened 2019-04-30
Extra clicking needed to change focus


sofroniewn:
## 🐛 Bug

Extra clicking is now needed to change focus when moving between different parts of the Qt application since the new styles #236. This occurs on my mac OSX, but doesn't seem to happen on @freeman-lab mac OSX - we will need to see which versions of `Qt`, `PyQt5` etc we each have.

sofroniewn:
To reproduce run `examples/add_shapes.py`. You should see the magnifying glass selected on the left in the control panel. Now click on the cursor symbol. On my machine the first click does nothing. If I then click again the cursor gets selected.

For an even stranger example - also run `examples/add_shapes.py`, but now click on the move_to_front button. If you then move the mouse away the button will remain highlighted as if you are hovering when you are not. If you click again it will return to normal.

Similar behavior can be seen if clicking into the layers list on the right or into the viewer into the center. Sometimes clicking in the viewer in the center can lock me into a weird pan mode.

It's really not clear to me what changed in #236 to cause this, nor how to fix it. I've tried a bunch of things including `focusPolicy`, `setWindowModality`, `setMouseTracking`, but none work.

Also I heard from @freeman-lab that he doesn't experience this on his mac. I tried `pip install --upgrade pyqt5` but I am already up to date.

Does anyone else experience this bug?

sofroniewn:
I will note that I also have a double title-bar when I run napari and @freeman-lab does not. These could be unrelated, but it is also strange that there are any differences for us.

See the extra title-bar
![image](https://user-images.githubusercontent.com/6531703/57006229-e4881980-6b93-11e9-98c4-2bb79bb76e82.png)

I will note that I've always had that extra toolbar line, but it wasn't so noticeable before, nor was it causing any problems

kne42:
Unable to reproduce on my machine. Output of `conda list | grep qt` is
```
pyqt5                     5.12                     pypi_0    pypi
pyqt5-sip                 4.19.14                  pypi_0    pypi
qtpy                      1.7.0                    pypi_0    pypi
sphinxcontrib-qthelp      1.0.2                    pypi_0    pypi
```

kne42:
is `PySide2` installed? what is the result of `python -c "import qtpy; print(qtpy.API_NAME)"`

sofroniewn:
I've uninstalled `PySide2` - but I don't think that was the issue as I noticed this before the `qtpy` PR merged and I get `PyQt5` when I run the command above.

[...]

PR #356, opened 2019-06-21
Add menubar to napari main window

jni:
# Description
Eventually, this menu will contain all the options we want out of napari, including file opening, adding plugins, and running plugins. For now, I leave it turned off for the programmatically-launched viewer, so this shouldn't affect functionality in any way, except adding the ability to toggle it on.

This is motivated by parallel work on https://github.com/napari/napari-app, where I'd like napari to behave like a native app, with a proper menubar.

As you'll see in the code, when the menubar is hidden, none of its shortcuts are active, which is annoying, but I didn't find a workaround. Suggestions most welcome!

I also don't know what this does on macOS (menubars shouldn't be hidden on mac, since the main menubar is always visible), so testing would also be appreciated! If the menubar items disappear when setting the menubar to invisible, I might favour adding a check to keep it always on on mac.

## Type of change

# References

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/main.py
+++ b/napari/main.py
@@ -25,7 +25,7 @@ def main():
     )
     args = parser.parse_args()
     with app_context():
-        v = Viewer(hide_menubar=False)
+        v = Viewer()
         images = io.ImageCollection(args.images, conserve_memory=False)
         if args.layers:
             for image in images:
@@ -35,6 +35,4 @@ def main():
                 image = images[0]
             else:
                 image = np.stack(images, axis=0)
-            v.add_image(
-                image, multichannel=args.multichannel, hide_menubar=False
-            )
+            v.add_image(image, multichannel=args.multichannel)
--- a/napari/main.py
+++ b/napari/main.py
@@ -11,7 +11,7 @@ from . import Viewer
 
 def main():
     parser = argparse.ArgumentParser(usage=__doc__)
-    parser.add_argument('images', nargs='+', help='Images to view.')
+    parser.add_argument('images', nargs='*', help='Images to view.')
     parser.add_argument(
         '--layers',
         action='store_true',
@@ -31,8 +31,9 @@ def main():
             for image in images:
                 v.add_image(image, multichannel=args.multichannel)
         else:
-            if len(images) == 1:
-                image = images[0]
-            else:
-                image = np.stack(images, axis=0)
-            v.add_image(image, multichannel=args.multichannel)
+            if len(images) > 0:
+                if len(images) == 1:
+                    image = images[0]
+                else:
+                    image = np.stack(images, axis=0)
+                v.add_image(image, multichannel=args.multichannel)
--- a/napari/main.py
+++ b/napari/main.py
@@ -1,0 +1,40 @@
+"""
+napari command line viewer.
+"""
+import argparse
+import numpy as np
+from skimage import io
+
+from .util import app_context
+from . import Viewer
+
+
+def main():
+    parser = argparse.ArgumentParser(usage=__doc__)
+    parser.add_argument('images', nargs='+', help='Images to view.')
+    parser.add_argument(
+        '--layers',
+        action='store_true',
+        help='Treat multiple input images as layers.',
+    )
+    parser.add_argument(
+        '-m',
+        '--multichannel',
+        help='Treat images as RGB.',
+        action='store_true',
+    )
+    args = parser.parse_args()
+    with app_context():
+        v = Viewer(hide_menubar=False)
+        images = io.ImageCollection(args.images, conserve_memory=False)
+        if args.layers:
+            for image in images:
+                v.add_image(image, multichannel=args.multichannel)
+        else:
+            if len(images) == 1:
+                image = images[0]
+            else:
+                image = np.stack(images, axis=0)
+            v.add_image(
+                image, multichannel=args.multichannel, hide_menubar=False
+            )
--- a/napari/_view.py
+++ b/napari/_view.py
@@ -3,10 +3,11 @@ from .viewer import Viewer
 
 def view(
     *images,
+    title='napari',
+    hide_menubar=True,
     meta=None,
     multichannel=None,
     clim_range=None,
-    title='napari',
     **named_images,
 ):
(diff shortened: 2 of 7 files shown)


sofroniewn:
Nice explorations @jni

I'm going to create an issue where we can discuss what belongs in the napari-app repo vs what belongs in the main napari repo and on the programaticly created Viewer. I know we've gone back and forth on some this stuff (like CLI) etc, and I was recently talking about this with @kne42, so having one central place to discuss it will probably be worth it, and while I see this as firmly `0.2` and after SciPy, it could become a high priority after the conference, so developing a plan now is probably worth it too

jni:
@sofroniewn would you be able to test what happens on mac with this branch when the menubar is hidden?

sofroniewn (changes requested):
Somethings not quite working right on the mac. I can't tell the difference between toggling `hide_menubar` and I'm not quite sure what the expected behavior there is / why we'd ever need / want to hide the menubar. When I launch the viewer now I get the screen bellow

image

The worst problem though is right now I'm unable to open any of the drop downs after I launch napari, my clicks are just ignored on the menubar. However if I toggle focus to another application, like chrome, and then come back to napari the menubar works. This was true for the very limited menubar that appeared before this PR as well (though it was less important then). I havn't looked into debugging yet, but can poke around

jni:
Thanks for testing, @sofroniewn!

> I can't tell the difference between toggling hide_menubar 

That is kinda sorta what I expected on Mac. The point is that macOS has a permanent menubar that changes based on application focus. (Side note: macOS is the only OS that does this right, based on sound UI theory. Specifically, the borders of the screen are effectively "infinite-size" targets for moving the mouse, because you can move the mouse "past" the border and it just sits there; This makes it the most useful real estate on your screen, so something useful should live there. On Windows and Linux, most of it is wasted on window chrome.) On Windows and Linux, though, the menubar sits *inside* the application, so toggling it makes sense. Here's what this looks like on Linux:

napari-menubar

> The worst problem though is right now I'm unable to open any of the drop downs after I launch napari

That sounds unrelated to this PR since it happened earlier. A quick search turned up e.g. [this](https://github.com/robotology/yarp/issues/457) among many other hits. I haven't yet read through the related threads to figure out the solution.

Of course we also want to set the menubar application menu on mac to say "napari" instead of "python".

Side note: did you fix your double titlebar, @sofroniewn???

jni:
Also, incidentally, given the above look (which will be mimicked by Windows), what do people think about having it there by default? I think it makes sense when napari is launched as a standalone, but less so when it's launched in a script, which is why I've designed it the way I have (default hidden, but shown when launched in `main()`), but am totally open to different views.

napari/main.py
```diff
@@ -0,0 +1,40 @@
+"""
+napari command line viewer.
+"""
+import argparse
+import numpy as np
+from skimage import io
```
sofroniewn, line 6: do we want to use `imageio` here instead? we currently depend on both libraries so could go either way

napari/main.py
```diff
@@ -0,0 +1,40 @@
...
+    parser.add_argument(
+        '-m',
+        '--multichannel',
+        help='Treat images as RGB.',
+        action='store_true',
+    )
+    args = parser.parse_args()
+    with app_context():
+        v = Viewer(hide_menubar=False)
+        images = io.ImageCollection(args.images, conserve_memory=False)
```
sofroniewn, line 29: if someone passes no images can you launch the empty viewer? Fine if you want to wait on this till future PRs (say when we add drag / drop capability), but also maybe nice just to get it in now

sofroniewn:
@jni - we got rid of the double titlebar with this line which is ignored if not on mac https://github.com/napari/napari/blob/db969b976ec34b18915ee49081aa2a8440cf225d/napari/components/_window/view.py#L26

If you can set the menubar application menu on mac to say "napari" instead of "python" can you do that in this PR?

I feel like the menubar should always be there by default - even on windows / linux regardless of how the app is launched. I'd rather people see it so they know the functionality is there and then hide it if they don't want to see it anymore vs. maybe missing it entirely and not knowing about it. I also think its a little strange to see in the code something this `hide_menu = True` on the mac but then still see the menubar at the top of my screen - though your explanation made sense. Finally I don't really like the idea of people seeing different "versions" of napari depending on how napari gets launched, even if it's just what the default settings are.

CLI works great - I'm looking forward to adding drag-drop!!!

jni:
@sofroniewn 

> we got rid of the double titlebar with this line which is ignored if not on mac 

Great! But did you ever figure out why you suffered from this and no one else did? Some default setting in your Qt was different for some reason? I wonder if some other qt software you installed modified some global config variable somewhere?

sofroniewn:
@jni - no didn't figure it out. as discussed in #239 the furthest progress i was able to make was figuring out that creating a new user got rid of that problem.

For this PR am I right in thinking the remaining things to do are
- [x]  remove the `hide_menu = True` keyword arg and make the menubar shown by default
- [x] make sure `napari` can be launched from the CLI with no keyword args
- [x] add file open to the menubar that opens an image / collection of images the same way that the CLI does

sofroniewn:
I have made a simple PR #378 that adds drag and drop functionality for files / folders to the viewer. Once this PR is merged we can pass the file / folders to the `open` functionality being added in this PR so that the same file / folder types that can be read with file open & the CLI can be dragged and dropped

jni:
@sofroniewn the other thing that I'm wanting help with is that the standard keyboard shortcut for hiding/showing the menubar, on Linux at least, is just tapping "alt" by itself, but I haven't found a way to make this work. If you or @kne42 have any ideas, let me know. For now we can go with Ctrl-M, or maybe Ctrl-B since Ctrl-M is usually maximise or minimise (I forget which LOL) on macOS.

jni:
@sofroniewn I've finished the items on your to-do list! =)

sofroniewn (approved):
Amazing work @jni - I love this so much. Works great, it's so much fun just to browse around opening images!! Choice of `ctrl-M` is fine with me. I'm approving and will merge today

PR #362, opened 2019-06-22
Add dims test and fix 5D images

### LINKED ISSUE #358: Incorrect number of sliders for particular shapes
## 🐛 Bug

This might be due to trying to be too clever with detecting multichannel, but maybe not... If I try to visualise an array of shape (51, 5, 2, 196, 171), I only get two sliders, corresponding to dimensions 0 and 2. I have no way to scroll through dimension 1. Additionally, there are too many stops for dimension 2, which can be set to three positions, 0, 1, and 2. The status happily tells me that I'm looking at coordinates [0, 0, 2, <whatever>, <whatever>]. 

Note: this was discovered while trying to visualise `mitosis.tif` from the Fiji examples, but random data does the trick (see below)

## To Reproduce

In IPython:

```python
import numpy as np
import napari

image = np.random.random((51, 5, 2, 196, 171))
napari.view(image)
```

As a script:

```python
import numpy as np
import napari
import napari.util

image = np.random.random((51, 5, 2, 196, 171))
with napari.util.app_context():
    napari.view(image)
```

## Expected behavior

I get three sliders, and they range from 0-50, 0-4, and 0-1.

## Environment

 - OS (e.g., Linux): Ubuntu 18.04
 - Python version: 3.6

sofroniewn:
# Description
This PR fixes #358, but does not fix the bug in #287.

It also adds comprehensive tests for all the code in the dims model and view files and makes sure they run with pytest.

## Type of change

# How has this been tested?

- [x] added `examples/5D_image.py` and add tests for dims model and view

## Final checklist:


sofroniewn:
I've now gone ahead and fixed #287 too, it was simple enough and now it's all in one PR. This is ready for review

napari/components/_dims/model.py
```diff
@@ -125,9 +125,9 @@ def ndim(self):
     def ndim(self, ndim):
         if ndim > self.ndim:
             for i in range(self.ndim, ndim):
-                self._range.insert(0, (0.0, 1.0, 0.01))
-                self._point.insert(0, 0.0)
-                self._interval.insert(0, (0.3, 0.7))
+                self._range.insert(0, (0, 2, 1))
+                self._point.insert(0, 0)
+                self._interval.insert(0, (0, 1))
```
jni, line 137: this looks less crazy than what was there before, but it's still confusing to me! Would you mind adding comments about the insert syntax and what is happening for each of these? I particularly don't understand range vs interval...

jni (approved):
@sofroniewn wonderful PR! Love all the tests. Can't wait for all our PRs to look like this. =)

napari/components/_dims/model.py
sofroniewn: have added comments to clarify / improved doc strings

PR #364, opened 2019-06-24
Shape thumbnails

sofroniewn:
# Description
This PR adds support for thumbnails for the shape layer. It colors rectangles, polygons, and ellipses with their face color and traces lines and paths with their edge color. All downsampling happens before the thumbnail is generated using the same process as the in the points layer. The positions of the shapes in the thumbnail also follow the shape of the entire layer, and can be resized when new shapes are added, as happens in the points layer. The thumbnail works in nD too. Right now the thumbnail is updated when a color changes, when drawing is finished, a new shape is added, or the mouse is released. It will not update during the drawing process as that would slow everything down too much.

Here's an example:

shapes_thumbnails

## Type of change

# How has this been tested?

- [x] run `examples/add_shapes.py` and `examples/nD_shapes.py`

## Final checklist:


sofroniewn:
I have updated this PR with the warning suppression from #370. It is ready for review

napari/layers/_shapes_layer/model.py
```diff
@@ -944,6 +952,28 @@ def _finish_drawing(self):
                 self.slice_data.remove(index)
         self._is_creating = False
         self.refresh()
+        self._update_thumbnail()
+
+    def _update_thumbnail(self):
+        """Update thumbnail with current points and colors.
+        """
+        offset = np.array([self.range[-2][0], self.range[-1][0]]) - 0.5
```
jni, line 962: Can you add comments here about what `self.range` holds? I'm confused by both the offset and the +1 in the lines below.

napari/layers/_shapes_layer/shape_list.py
```diff
@@ -668,6 +681,12 @@ def to_labels(self, labels_shape=None, shape_type=None):
         labels_shape : np.ndarray | tuple | None
             2-tuple defining shape of labels image to be generated. If non
             specified, takes the max of all the vertiecs
+        zoom_factor : float
+            Premultiplier applied to coordinates before generating mask. Used
+            for generating as downsampled mask.
+        offset : 2-tuple
+            Offset preapplied to coordinates before multiplying by the
+            zoom_factor. Used for putting negative coordinates into the mask.
```
jni, line 689: (a) used how?, and (b) what happens if there are negative coordinates and offset remains [0, 0]?

jni (changes requested):
I think that the `zoom_factor` keyword argument feels a bit out of place. An alternative implementation would be to write `def scale(factor=1):` which returns a copy of the layer with all the point coordinates scaled appropriately. Then to make the thumbnail you'd use `self.scale(32/self.shape).to_colors()`. What do you think?

napari/layers/_shapes_layer/shape_list.py
sofroniewn: I've changed this to `added`, and added `np.clip` to the code, and changed the doc string to reflect this

napari/layers/_shapes_layer/model.py
sofroniewn: have added comments - range holds the min / max values of the shape data. offset holds the min values, and the +1 is there for padding. all this is explained in the comments

sofroniewn:
At @jni - agreed the `zoom_factor` and `offset` keywords are a little ugly and maybe overly specific for generating the thumbnails, but I worry that copying the data and then using the scale / transform properties will incur a lot of unnecessary computation - like generating the meshes - and while it would be possible to decouple that stuff it would be a lot for right now. I'd rather see this merged and then we can focus on the refactoring / tests.

napari/layers/_shapes_layer/shape_list.py
sofroniewn: I meant `subtracted`!! have fixed :-)

napari/layers/_labels_layer/model.py
```diff
@@ -544,6 +544,7 @@ def _update_thumbnail(self):
         downsampled = self.raw_to_displayed(downsampled)
         colormapped = self.colormap.map(downsampled)
         colormapped = colormapped.reshape(downsampled.shape + (4,))
+        colormapped[..., 3] = 1
```
jni, line 548: what was happening here before?

napari/layers/_shapes_layer/model.py
```diff
@@ -944,6 +952,34 @@ def _finish_drawing(self):
                 self.slice_data.remove(index)
         self._is_creating = False
         self.refresh()
+        self._update_thumbnail()
+
+    def _update_thumbnail(self):
+        """Update thumbnail with current points and colors.
+        """
```
jni: PEP257

napari/layers/_shapes_layer/shape_list.py
```diff
@@ -701,11 +722,82 @@ def to_labels(self, labels_shape=None, shape_type=None):
...
         return labels
 
+    def to_colors(
+        self, colors_shape=None, zoom_factor=1, offset=[0, 0], shape_type=None
+    ):
+        """Returns an rgba image, where each shape is embedded in an
+        array of shape colors_shape with the rgba value of the shape,
+        and 0 for background. Passing a `shape_type` argument leads to only
+        colors from that particular `shape_type` being returned. For
+        overlapping shapes z-ordering will be respected.
```
jni: PEP257. I suggest:

```python
"""Rasterize shapes to an RGBA image array.

Each shape is embedded in an array of shape `colors_shape` with the
RGBA value of the shape...
"""

File: napari/layers/_shapes_layer/shapes/ellipse.py [hunk 6]
```diff
@@ -104,15 +104,22 @@ def transform(self, transform):
         self._edge_offsets = offsets
         self._edge_triangles = triangles
 
-    def to_mask(self, mask_shape=None):
+    def to_mask(self, mask_shape=None, zoom_factor=1, offset=[0, 0]):
         """Converts the shape vertices to a boolean mask with `True` for points
-        lying inside the shape.
+        lying inside the shape. Negative points or points outside the
+        mask_shape after the zoom and offset are clipped.
```
jni: PEP257

jni (approved):
@sofroniewn well, imho you're making it harder on yourself doing this before the vispy refactor, but who am I to judge. ;) Code lgtm other than the already-discussed concerns!

napari/layers/_labels_layer/model.py
sofroniewn: space where there were no shapes was being rendered as transparent instead of black, i think rendering as black makes it more clear what is going on. i've added a comment

napari/layers/_shapes_layer/shape_list.py
sofroniewn: fixed

napari/layers/_shapes_layer/model.py
sofroniewn: fixed

napari/layers/_shapes_layer/shapes/ellipse.py
sofroniewn: fixed here and elsewhere

PR #370, opened 2019-06-27
Refactor thumbnail type conversion

kevinyamauchi:
# Description
This PR moves the image type conversion when generating thumbnails to the base layer. Also, it ignores the precision loss warnings from skimage when converting the thumbnail image to uint8.

## Type of change

# References
This is the solution discussed in #355  and is a temporary fix for #349 until skimage 0.16 is released.

# How has this been tested?
I ran examples for each layer type altered
- [x] `/examples/add_image.py`
- [x] `/examples/add_points.py`
- [x] `/examples/labels-2d.py`

## Final checklist:

### DIFF
--- a/napari/layers/_base_layer/model.py
+++ b/napari/layers/_base_layer/model.py
@@ -1,7 +1,9 @@
+import warnings
 from abc import ABC, abstractmethod
 from contextlib import contextmanager
 from xml.etree.ElementTree import Element, tostring
 import numpy as np
+from skimage import img_as_ubyte
 
 from ...util.event import Event
 from ._visual_wrapper import VisualWrapper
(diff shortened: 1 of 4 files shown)


napari/layers/_base_layer/model.py
```diff
@@ -163,6 +165,10 @@ def thumbnail(self):
 
     @thumbnail.setter
     def thumbnail(self, thumbnail):
+        if thumbnail.dtype != 'uint8':
```
kne42: prefer you to compare to `np.uint8`

napari/layers/_image_layer/model.py
```diff
@@ -332,9 +331,9 @@ def _update_thumbnail(self):
             )
             if image.shape[2] == 4:  # image is RGBA
                 downsampled[..., 3] = downsampled[..., 3] * self.opacity
-                colormapped = img_as_ubyte(downsampled)
+                colormapped = downsampled
```
kne42: if these are now equivalent, should we refactor too?

napari/layers/_base_layer/model.py
kevinyamauchi: Ah right. Thanks!

napari/layers/_image_layer/model.py
kevinyamauchi: Good call. I took a crack at it Let me know if this is what you had in mind!

kne42 (approved):
LGTM! approved if tests pass

PR #371, opened 2019-06-29
Selectable points

sofroniewn:
# Description
This PR adds support for selectable points, moving the points layer more in line with the shapes layer. Points can now be selected independently or as a group and be resized, colored, and moved accordingly. They can also be copied and pasted. All functionality works in nD.

Here's interactivity in `examples/nD_points.py`:

selectable_points

I'll note the the z-order of points cannot be changed once set, but I think that is ok. There is also no way to resize points and maintain their relative sizing as @kevinyamauchi has requested. We could implement that, but unless everyone agrees that that is the behavior they want to see I feel like it could wait for a follow-up PR. I am open to changing that though if people feel strongly, @jni? 

## Type of change

# How has this been tested?

- [x] run `examples/add_points.py` and `examples/nD_points.py`

## Final checklist:


napari/layers/_points_layer/model.py
```diff
@@ -76,18 +91,67 @@ def __init__(
...
+                    range(len(coords)), ensure_iterable(face_color, color=True)
+                )
+            ]
+
+            # The following point properties are for the new points that will
+            # be added. Each point also has a corresponding property with the
+            # value for itself
+            if np.isscalar(size):
+                self._size = size
+            else:
```
jni, line 116: Thanks for all the comments everywhere, they are great! I'm still unclear what else "size" can be if not a scalar. Could you add a comment here, or maybe even an `elif size is None: ... else: raise TypeError`?

napari/layers/_points_layer/model.py
```diff
@@ -76,18 +91,67 @@ def __init__(
...
-            self._selected_points = None
+            self.edge_width = edge_width
+
+            self.size_array = size
+            self._edge_color_list = [
+                s
+                for i, s in zip(
+                    range(len(coords)), ensure_iterable(edge_color, color=True)
+                )
+            ]
```
jni: This took me a few seconds to parse... I think two alternatives that I would prefer are:

```python
import itertools

...

= list(itertools.islice(ensure_iterable(edge_color, color=True), 0, len(coords)))
```

or

```python
[s for s in itertools.islice(ensure_iterable(edge_color, color=True), 0, len(coords))]
```

A final option would be to add `toolz` as a dependency:

```python
import toolz as tz

= list(tz.take(len(coords), ensure_iterable(edge_color, color=True)))
```

But I think that is premature (though it is a very light and stable dependency).

napari/layers/_points_layer/model.py
```diff
@@ -103,26 +167,40 @@ def coords(self) -> np.ndarray:
...
             # If there are now less points, remove the sizes of the missing
             # ones
             with self.freeze_refresh():
-                self.size = self._size[: len(coords)]
-        elif len(coords) > len(self._size):
+                self.size_array = self._size_array[: len(coords)]
+                self._edge_color_list = self._edge_color_list[: len(coords)]
+                self._face_color_list = self._face_color_list[: len(coords)]
+
+        elif len(coords) > cur_npoints:
```
jni, line 182: Wait, does this mean that `points.coords = <10x2 array>`, followed by `points.coords = <12x2 array>`, will ignore the first 10 points of the second array?

napari/layers/_points_layer/model.py
```diff
@@ -230,19 +330,92 @@ def face_color(self) -> str:
...
+        Parameters
+        ----------
+        index : list
+            List of points around which to construct the interaction box.
+
+        Returns
+        ----------
+        box : np.ndarray
+            5x2 array of corners of the interaction box in clockwise order
+            starting in the upper-left corner and duplicating the first corner
```
jni, line 388: Is the duplication necessary?

napari/layers/_points_layer/model.py
```diff
@@ -324,47 +497,54 @@ def _slice_points(self, indices):
             Indices to slice with.
         """
         # Get a list of the coords for the points in this slice
-        coords = self.coords
+        coords = self.data
```
jni, line 511: Can you add `Returns` to the docstring?

napari/layers/_points_layer/model.py
```diff
@@ -375,37 +555,133 @@ def _select_point(self, indices):
...
+        else:
+            edge_color = 'white'
+            face_color = 'white'
 
-        self._node.set_data(
-            data[:, [1, 0]],
-            size=sizes,
+        self._node._subvisuals[2].set_data(
+            data[::-1, [1, 0]],
+            size=sizes[::-1],
```
jni, line 616: Can you add a comment here that this is to convert data coordinates (row/col) to Vispy canvas coordinates (x/y)? I presume that's what it's for... I also don't know why there's two inversions, `::-1` for axis 0 of `data` and then `[1, 0]` for axis 1 of data.

napari/layers/_points_layer/model.py
```diff
@@ -375,37 +555,133 @@ def _select_point(self, indices):
...
             scaling=True,
         )
         self._need_visual_update = True
+        self._set_highlight(force=True)
         self._update()
-        self.status = self.get_message(self.coordinates, self._selected_points)
+        self.status = self.get_message(self.coordinates, self._hover_point)
         self._update_thumbnail()
 
+    def _set_highlight(self, force=False):
```
jni, line 629: Should this not be part of the view rather than the model?

napari/layers/_points_layer/model.py
```diff
@@ -428,7 +704,8 @@ def get_message(self, coord, value):
         if value is None:
             pass
         else:
-            msg = msg + ', index ' + str(value)
+            index = self._indices_view[value]
+            msg = msg + ', index ' + str(index)
```
jni, line 724: Can you explain this? I'm confused about what `value` is in this context, as well as `index`.

napari/layers/_points_layer/model.py
```diff
@@ -451,42 +728,108 @@ def _update_thumbnail(self):
...
     def _add(self, coord):
-        """Adds object at given mouse position
+        """Adds point at given mouse position
         and set of indices.
         Parameters
         ----------
         coord : sequence of indices to add point at
         """
         self.data = np.append(self.data, [coord], axis=0)
-        self._selected_points = len(self.data) - 1
```
jni, line 468: Are we sure we want to remove this? Does a shape not get automatically selected after you're done making it?

napari/layers/_points_layer/model.py
```diff
@@ -451,42 +728,108 @@ def _update_thumbnail(self):
...
+                self._drag_start = coord[-2:]
+            self._drag_box = np.array([self._drag_start, coord[-2:]])
+            self._set_highlight()
+
+    def _copy_points(self):
+        """Copy selected points to clipboard.
+        """
+        if len(self.selected_points) > 0:
+            index = self._indices_view[self.selected_points]
+            self._clipboard = {
```
jni, line 805: whoa. =D

napari/layers/_points_layer/model.py
```diff
@@ -451,42 +728,108 @@ def _update_thumbnail(self):
...
+                self._face_color_list + self._clipboard['face_color']
+            )
+            self._selected_points = list(
+                range(npoints, npoints + len(self._clipboard['coord']))
+            )
+            self._cur_selected = list(
+                range(totpoints, totpoints + len(self._clipboard['coord']))
+            )
             self.refresh()
+            self._copy_points()
```
jni: Why do they need to be copied again?

napari/layers/_points_layer/model.py
```diff
@@ -550,25 +937,112 @@ def on_key_press(self, event):
...
+    ----------
+    data : (N, 2) array
+        Points around which the interaction box is created
+
+    Returns
+    -------
+    box : (5, 2) array
+        Vertices of the interaction box with the top left corner duplicated
+    """
+    min_val = [data[:, 0].min(axis=0), data[:, 1].min(axis=0)]
```
jni: `min_val = data[:, :2].min(axis=0)`?

jni: This all seems like work that should happen in the view rather than the model...

napari/layers/_points_layer/view/properties.py
```diff
@@ -133,10 +127,6 @@ def _on_symbol_change(self, event):
     def _on_size_change(self, event):
         with self.layer.events.size.blocker():
             value = self.layer.size
-            if isinstance(value, Iterable):
-                if isinstance(value, list):
-                    value = np.asarray(value)
-                value = value.mean()
```
jni, line 139: Just for future reference, for this pattern imho it's better to use `value = np.mean(value)`, which will work equally well with lists and arrays.

jni (approved):
@sofroniewn some minor comments, as always. In general, I'm concerned that a lot of "view-y" stuff is ending up in the model...

napari/layers/_points_layer/model.py
sofroniewn: i've added some comments to explain. it was more about the case when a list / array is passed instead and then just using the default value here instead.

napari/layers/_points_layer/model.py
sofroniewn: i've gone with the top suggestions

napari/layers/_points_layer/model.py
sofroniewn: no this will then pad the `size_array`, `_edge_color_list`, and `_face_color_list` so that they go from length 10 to length 12 using the new sizes / colors. I've tried to improve the comments. Let me know if it is still unclear or if there is something else confusing

napari/layers/_points_layer/model.py
sofroniewn: nope, fixed

napari/layers/_points_layer/model.py
sofroniewn: added

napari/layers/_points_layer/model.py
sofroniewn: agreed - we havn't seperated out the vispy views yet from the individual layers. It's the last remaining thing to do in our MVC refactor and we can start getting to it soon after this PR merges. At that point this will probably go along into the view file

napari/layers/_points_layer/model.py
sofroniewn: as noted above, we havn't separated out the vispy view yet, but one we do the highlighting stuff will move out there. The box creation might remain here as that is useful for determining which points are inside the box etc. But definitely the highlight itself will go over.

sofroniewn: also fixed the min / max values

napari/layers/_points_layer/model.py
sofroniewn: fixed, had to add another `deepcopy` in above

napari/layers/_points_layer/model.py
sofroniewn: when in `add` mode it's now not actually possible to select points at all. this is similar to the shapes layer and i think works fine

napari/layers/_points_layer/model.py
sofroniewn: added a comment - `value` is the index of the point in the slice and passing it through that other list maps it to the index of the point in the overall data, which is what we'll want to display to the user

napari/layers/_points_layer/model.py
sofroniewn: comment added. the reversing of the list is about the z_order of the points, ensuring the most recently added point appears on top. the `[1, 0]` is for the row / col -> x / y

sofroniewn:
I have made the fixes requested by @jni. The concerns about the vispy model / view separation are well founded. We havn't pulled out the vispy stuff for any of the individual layers yet, but will get to it very soon. We will then revisit that here too.

This PR  is ready for re-review

sofroniewn:
I'm going to merge - we can revisit the model / view separation when we pull out the vispy components from all the layers

PR #378, opened 2019-06-30
add drag and drop

sofroniewn:
# Description
This PR allows the viewer to load image files from drag and drop events for local files and directories. It creates a new file in our `util` folder that can contain some of our code for file parsing. I've created an other issue for discussion our file IO more broadly #379

## Type of change

# How has this been tested?

- [x] drag and drop files / folders / text (not accepted) into `examples/add_image.py`

## Final checklist:


napari/components/_viewer/view/main.py
```diff
@@ -123,26 +126,32 @@ def screenshot(self, region=None, size=None, bgcolor=None):
         """
         return self.canvas.render(region, size, bgcolor)
 
-    def _open_images(self):
-        """Add an image layer to the viewer.
+    def _open_files(self):
+        """Add an files from the menubar.
+        """
```
jni: I never mention anything because I hate nitpicking (no really I do! =P), but thought I'd bring it up maybe once and for all: [PEP257](https://www.python.org/dev/peps/pep-0257/) specifies that one-liner docstrings have the closing triple-quote on the same line.

Also might as well fix the grammar, "Add an files". =P

Finally, I'm not sure I agree with the renaming to "files", because I hope to eventually add "open segmentation", "open points", etc...

napari/components/_viewer/view/main.py
```diff
@@ -230,6 +239,24 @@ def on_draw(self, event):
...
 
+    def dragEnterEvent(self, event):
+        if event.mimeData().hasUrls():
+            event.accept()
+        else:
+            event.ignore()
+
+    def dropEvent(self, event):
+        """Add an files from the drag and drop.
+        """
```
jni: See above

napari/components/_viewer/view/main.py
```diff
@@ -230,6 +239,24 @@ def on_draw(self, event):
...
+        filenames = []
+        for url in event.mimeData().urls():
```
jni, line 250: Can you specify in a comment whether web URLs are allowed here? `skimage.io.imread` opens URLs so they totally could be!

jni (approved):
@sofroniewn left some comments, approved regardless, good stuff!

jni:
btw great job unifying the read function!

napari/components/_viewer/view/main.py
sofroniewn: I don't mind the nitpick, i'm just sorry to create extra work for you cause I always forget! where is black up for doc strings!!!

I'll rename to `_open_images` for now as that is what it is doing. We'll have to discuss in #379 what is feasible for one "open files" function vs what need separate "open segmentation", "open points" etc. I prefer as much of the former as possible, but it depends on us figuring out what we're opening and there could be ambiguous cases

napari/components/_viewer/view/main.py
sofroniewn: fixed

napari/components/_viewer/view/main.py
sofroniewn: i've made it so we urls are now supported and added a comment. they work fine as along as its just an image you drag in, otherwise you can an `skimage.io.imread` error and it crashes. we'll have to work on error handling / testing etc. but that can come in a followup PR I think

Issue #210, opened 2019-04-10
Unified framework for keybindings


jni:
Currently, keybindings are implemented as a hodgepodge of handcrafted if-else statements, see e.g. https://github.com/napari/napari/pull/195/files#diff-347152977a8bc6500ffce60fc0e90a0dR653

What I envision is a yaml config file that maps keys to functions, as well as a description. Each layer comes with its own file, as well as the base viewer. When a layer is selected, its keybindings override base layer keybindings. I think probably viewer keybindings should be universal and not allowed to be overwritten, but that can be discussed. We should *not* allow multiple keybindings. mpl does this and gives no indication about what is happening (the keybinding triggers *all* associated events!).

When the keybinding '?' is pressed, we use some kind of temporary, transparent popup display of the keybindings, with the layer keybindings on top and other keybindings following it. The popup displays each bound key and its description from the config file.

Config files can be customised. Custom copies live in .config/napari and are loaded *after* the default copy.

sofroniewn:
Hopefully this will then make it easier to support #200

gregjohnso:
> Hopefully this will then make it easier to support #200

and #212

sofroniewn:
Yes our framework should also allow attaching custom functions for mouse click, mouse press, mouse release, and mouse doubleclick

kevinyamauchi:
@jni , I think this is a great idea. In terms of features, I think this is a pretty high priority after #171 lands. I would even propose this is one of the last feature to add before circling back for testing/clean up of all of the awesome features that have been added thus far.

I like the idea of a mapping between events and functions. Were you thinking the main `Viewer` and each layer would have a dictionary where the keys are events and the values are the associated function? If we want to prevent overriding of the base `Viewer` key/event bindings, we can just check the `Viewer` dictionary first. We can also check for conflicts upon loading layers.

Do you think it's useful to allow the user to add/remove keybindings at run time (e.g., `Layer.add_binding()`? One potential use case is if the user wants to add behavior around a common layer for a specific task (e.g., painting labels on data) and does not want to write over the layer key bindings config file. If we store the keybinding/function mapping in a dictionary, it should be fairly straightforward to add/change bindings. Of course the challenge is validating them.

sofroniewn:
@kevinyamauchi i think you mean #171 not #115 (also named dims refactor ...)

I really like the `add_binding` idea where a custom dictionary is passed - I think it is similar to the proposal from @gregjohnso described in #200

[...]

PR #389, opened 2019-07-03
standardize keybindings framework

kne42:
# Description

Add machinery for the adding, validation, and dispatch of keybindings on the viewer and its layers.

Supersedes #251.

### key combinations
Keys are represented in the form `[modifier-]key`, e.g. `a`,
`Control-c`, or `Control-Alt-Delete`.
Valid modifiers are Control, Alt, Shift, and Meta.

Letters will always be read as upper-case.
Due to the native implementation of the key system, Shift pressed in certain
key combinations may yield inconsistent or unexpected results.
Therefore, it is not recommended to use Shift with non-letter keys.
On OSX, Control is swapped with Meta such that pressing Command reads as
Control.

Special keys include Shift, Control, Alt, Meta, Up, Down, Left, Right,
PageUp, PageDown, Insert, Delete, Home, End, Escape, Backspace, F1,
F2, F3, F4, F5, F6, F7, F8, F9, F10, F11, F12, Space, Enter, and Tab

### function format
Functions take in only one argument: the parent that the function was bound to. This is the viewer or layer.

By default, all functions are assumed to work on key presses only, but can be denoted to work on release too by separating the function into two statements with the `yield` keyword:
```python
def hello_world(viewer):
    # on key press
    viewer.status = 'hello world!'

    yield

    # on key release
```

### bind a key
```python
viewer.bind_key('n', load_next_image)

@viewer.bind_key('p')
def load_prev_image(viewer): ...
```

### unbind a key
```python
viewer.bind_key('p', None)
```

## Type of change

# References

see #210, #251

# How has this been tested?

- [x] the test suite for my feature passes

## Final checklist:

### DIFF
--- a/napari/util/keybindings.py
+++ b/napari/util/keybindings.py
@@ -1,0 +1,380 @@
+"""Key combinations are represented in the form ``[modifier-]key``,
+e.g. ``a``, ``Control-c``, or ``Control-Alt-Delete``.
+Valid modifiers are Control, Alt, Shift, and Meta.
+
+Letters will always be read as upper-case.
+Due to the native implementation of the key system, Shift pressed in certain
+key combinations may yield inconsistent or unexpected results.
+Therefore, it is not recommended to use Shift with non-letter keys.
+On OSX, Control is swapped with Meta such that pressing Command reads as
+Control.
+
+Special keys include Shift, Control, Alt, Meta, Up, Down, Left, Right,
+PageUp, PageDown, Insert, Delete, Home, End, Escape, Backspace, F1,
+F2, F3, F4, F5, F6, F7, F8, F9, F10, F11, F12, Space, Enter, and Tab
+
+Functions take in only one argument: the parent that the function
+was bound to. This is the viewer or layer.
+
+By default, all functions are assumed to work on key presses only,
+but can be denoted to work on release too by separating the function
+into two statements with the yield keyword::
+
+    @viewer.bind_key('h')
+    def hello_world(viewer):
+        # on key press
+        viewer.status = 'hello world!'
+
+        yield
+
+        # on key release
+        viewer.status = 'goodbye world :('
+
+"""
+
+import re
+import types
+from collections import OrderedDict, UserDict
+
+from vispy.util import keys
+
+
+SPECIAL_KEYS = [
+    keys.SHIFT,
+    keys.CONTROL,
+    keys.ALT,
+    keys.META,
+    keys.UP,
+    keys.DOWN,
+    keys.LEFT,
+    keys.RIGHT,
+    keys.PAGEUP,
+    keys.PAGEDOWN,
+    keys.INSERT,
+    keys.DELETE,
+    keys.HOME,
+    keys.END,
+    keys.ESCAPE,
+    keys.BACKSPACE,
+    keys.F1,
+    keys.F2,
+    keys.F3,
+    keys.F4,
+    keys.F5,
+    keys.F6,
+    keys.F7,
+    keys.F8,
+    keys.F9,
+    keys.F10,
+    keys.F11,
+    keys.F12,
+    keys.SPACE,
+    keys.ENTER,
+    keys.TAB,
+]
+
+MODIFIER_KEYS = [keys.CONTROL, keys.ALT, keys.SHIFT, keys.META]
+
+
+def parse_key_combo(key_combo):
+    """Parse a key combination into its components in a comparable format.
+
+    Parameters
+    ----------
+    key_combo : str
+        Key combination.
+
+    Returns
+    -------
+    key : str
+        Base key of the combination.
+    modifiers : set of str
+        Modifier keys of the combination.
+    """
+    parsed = re.split('-(?=.+)', key_combo)
+    *modifiers, key = parsed
+
+    return key, set(modifiers)
+
+
+def components_to_key_combo(key, modifiers):
+    """Combine components to become a key combination.
+
+    Modifier keys will always be combined in the same order:
+    Control, Alt, Shift, Meta
+
+    Letters will always be read as upper-case.
+    Due to the native implementation of the key system, Shift pressed in certain
+    key combinations may yield inconsistent or unexpected results.
+    Therefore, it is not recommended to use Shift with non-letter keys.
+    On OSX, Control is swapped with Meta such that pressing Command reads as
+    Control.
+
+    Parameters
+    ----------
+    key : str or vispy.app.Key
+        Base key.
+    modifiers : combination of str or vispy.app.Key
+        Modifier keys.
+
+    Returns
+    -------
+    key_combo : str
+        Generated key combination.
+    """
+    if len(key) == 1 and key.isalpha():  # it's a letter
+        key = key.upper()
+        cond = lambda m: True  # noqa: E731
+    elif key in SPECIAL_KEYS:
+        # remove redundant information i.e. an output of 'Shift-Shift'
+        cond = lambda m: m != key  # noqa: E731
+    else:
+        # Shift is consumed to transform key
+
+        # bug found on OSX: Command will cause Shift to not
+        # transform the key so do not consume it
+        # note: 'Control' is OSX Command key
+        cond = lambda m: m != 'Shift' or 'Control' in modifiers  # noqa: E731
+
+    modifiers = tuple(
+        key.name
+        for key in filter(
+            lambda key: key in modifiers and cond(key), MODIFIER_KEYS
+        )
+    )
+
+    return '-'.join(modifiers + (key,))
+
+
+def normalize_key_combo(key_combo):
+    """Normalize key combination to make it easily comparable.
+
+    All aliases are converted and modifier orders are fixed to:
+    Control, Alt, Shift, Meta
+
+    Letters will always be read as upper-case.
+    Due to the native implementation of the key system, Shift pressed in certain
+    key combinations may yield inconsistent or unexpected results.
+    Therefore, it is not recommended to use Shift with non-letter keys.
+    On OSX, Control is swapped with Meta such that pressing Command reads as
+    Control.
+
+    Parameters
+    ----------
+    key_combo : str
+        Key combination.
+
+    Returns
+    -------
+    normalized_key_combo : str
+        Normalized key combination.
+    """
+    key, modifiers = parse_key_combo(key_combo)
+
+    if len(key) != 1 and key not in SPECIAL_KEYS:
+        raise TypeError(f'invalid key {key}')
+
+    for modifier in modifiers:
+        if modifier not in MODIFIER_KEYS:
+            raise TypeError(f'invalid modifier key {modifier}')
+
+    return components_to_key_combo(key, modifiers)
+
+
+UNDEFINED = object()
+
+
+def bind_key(keymap, key, func=UNDEFINED, *, overwrite=False):
+    """Bind a key combination to a keymap.
+
+    Parameters
+    ----------
+    keymap : dict of str: callable
+        Keymap to modify.
+    key : str
+        Key combination.
+    func : callable or None
+        Callable to bind to the key combination.
+        If ``None`` is passed, unbind instead.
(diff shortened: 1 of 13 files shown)


sofroniewn (approved):
This looks great - I gave it a try. I was unable to bind to a Class though, i.e. I tried
```python
from napari.layers import Points
@Points.bind_key('w')
def func(layer):
    print('hello')
```
in `custom_keybindings.py` and then added a new points layer from the gui but the `w` key didn't seem to be bound. Maybe you could expand the example a little to include this use case. Maybe also modify one of the functions to print something on release too in that example. Otherwise I have no comments on the implementation, though I'll let @jni review before merging.

royerloic:
Beautiful feature, really like it.

napari/components/_viewer/view/main.py
```diff
@@ -214,20 +218,44 @@ def on_mouse_release(self, event):
...
-        ):
-            self.viewer.key_bindings[event.text](self.viewer)
+        if event.native.isAutoRepeat() or event.key is None:
             return
 
+        comb = components_to_comb(event.key.name, event.modifiers)
+
         layer = self.viewer.active_layer
+
+        # TODO: remove me once keybinding system converted
```
jni: What does this mean? Meaning, reworking all existing keybindings to use the new system?

Incidentally, how does this system interact with the Qt menubar system etc?

napari/components/_viewer/view/main.py
```diff
@@ -214,20 +218,44 @@ def on_mouse_release(self, event):
...
         if layer is not None:
             layer.on_key_press(event)
 
+        if layer is not None and comb in layer.keymap:
```
jni: Could you expand the names of things? `comb` -> `key_combo`, `parent` -> `key_receiver` (or somesuch, something more descriptive of what it is), `func` -> `function`, `gen` -> `result`?

napari/layers/_base_layer/model.py
```diff
@@ -6,10 +6,11 @@
 from skimage import img_as_ubyte
 
 from ...util.event import Event
+from ...util.keybindings import KeymapMixin
 from ._visual_wrapper import VisualWrapper
 
 
-class Layer(VisualWrapper, ABC):
+class Layer(VisualWrapper, KeymapMixin, ABC):
```
jni: Total side note: I'm also unclear from the name `VisualWrapper` what it is or what it's doing, so that might be another thing to rename in the near future.

napari/layers/_image_layer/model.py
```diff
@@ -94,6 +94,8 @@ class Image(Layer):
 
     _colormaps = AVAILABLE_COLORMAPS
 
+    class_keymap = {}
```
jni: Can this variable not be provided by the Mixin directly? It's a bit noisy that it has to be added to any subclass in the hierarchy...

napari/util/keybindings.py
```diff
@@ -0,0 +1,372 @@
...
+
+    @viewer.bind_key('h')
+    def hello_world(viewer):
+        # on key press
+        viewer.status = 'hello world!'
+
+        yield
+
+        # on key release
+        viewer.status = 'goodbye world :('
```
jni, line 31: I love this example! :joy:

napari/util/keybindings.py
```diff
@@ -0,0 +1,372 @@
...
+
+    Returns
+    -------
+    key : str
+        Base key of the combination.
+    modifiers : set of str
+        Modifier keys of the combination.
+    """
+    parsed = re.split('-(?=.+)', comb)
+    modifiers, key = parsed[:-1], parsed[-1]
```
jni: To reiterate:

```python
*modifiers, key = parsed
```

:grinning:

napari/util/keybindings.py
```diff
@@ -0,0 +1,372 @@
...
+            viewer.status = 'goodbye world :('
+
+    """
+    if func is UNDEFINED:
+
+        def inner(func):
+            bind_key(keymap, key, func, overwrite=overwrite)
+            return func
+
+        return inner
```
jni, line 248: What's happening here? Can you explain it in a comment?

napari/util/keybindings.py
```diff
@@ -0,0 +1,372 @@
...
+
+class KeybindingDescriptor:
+    """Descriptor which transforms ``func`` into a method with the first argument bound
+    to ``class_keymap`` or ``_keymap`` depending on if it was called
+    from the class or the instance, respectively.
+
+    Parameters
+    ----------
+    func : callable
+        Function to bind.
```
jni, line 276: Maybe add a usage example to the docstring? I'm not sure how this works, how to use it, or what problem this solves, based on just reading the description.

napari/util/keybindings.py
```diff
@@ -0,0 +1,372 @@
...
+    """Dictionary which inherits from another.
+
+    Values of ``None`` are treated as though the key doesn't exist.
+
+    Parameters
+    ----------
+    parent : callable() -> dict
+        Parent keymap.
+    contents : dict, optional
+        Contents with which to initialize.
```
jni, line 304: As above: some usage examples would be very handy here.

napari/util/keybindings.py
```diff
@@ -0,0 +1,372 @@
...
+                )
+
+        return object.__new__(cls)
+
+    def __init__(self):
+        self.keymap = {}
+
+    @property
+    def keymap(self):
+        """InhertiedKeymap : Keymap used for shortcuts.
```
jni: typo: inherited

napari/util/keybindings.py
```diff
@@ -0,0 +1,372 @@
...
+                raise
+        else:
+            del self[name]
+
+        return default
+
+
+class KeymapMixin:
+    """Mix-in to add keymap functionality. Must still define ``class_keymap``
+    in every subclass.
```
jni, line 350: Is there really no way around this?

jni (commented):
@kne42 the functionality is super awesome, and so much simpler than the last attempt! Thank you! =) I've left a few comments, but most are optional/requests for clarification. The main issue now is rebasing on master, which should be fun! =P

napari/layers/_base_layer/model.py
sofroniewn: `VisualWrapper` relates to our handling of the `vispy` scene and will get pulled out when we separate our vispy code from our layer models

napari/components/_viewer/view/main.py
kne42: Precisely.

Getting this to work with Qt's built-in shortcuts will be tricky if we want to do it natively. Otherwise, it would be pretty simple to just do something like `Viewer.bind_key('Control-W', lambda v: v.close())`

napari/components/_viewer/view/main.py
kne42: sure, but do you really think `function` is that much more descriptive than `func`? I thought that's pretty standard

I also chose `gen` because sometimes it returns a generator. still want to go with `result` or can you think of something that better captures this?

napari/layers/_image_layer/model.py
kne42: Dictionaries are mutable so it would create quite the mess if we did that. We could resolve this by using a custom metaclass but I'm not so keen on that either.

napari/util/keybindings.py
kne42: right, sorry I forgot this last time

napari/util/keybindings.py
kne42: it's similar to using `functools.partial` except it returns a function, which allows this to be used as a decorator with arguments and also as a regular function

napari/util/keybindings.py
kne42: Reiterating for those not following along the full conversation: we could use a metaclass which would be pretty gross too.

kne42:
@sofroniewn this is because I forgot to update `__contains__`. I think I'd like to eventually re-implement a custom dictionary class similar to what I did with `List` because there are like 1000 more bugs with the current implementation (e.g. repr, etc.).

napari/components/_viewer/view/main.py
jni: @kne42 func is the least worrisome. `gen` being *sometimes* a generator is a huge problem! LOL `value_or_generator` is the only thing I can think of otherwise, but `result` works in any case. maybe add a comment, "can be a single value or a generator"?

napari/layers/_image_layer/model.py
jni: I was thinking something like overriding `__new__` so that if the class has a dictionary, use that, but if it doesn't, add one? Is that possible? I'm not too clear on class variables in Python actually!

napari/util/keybindings.py
jni: I'm confused. functools.partial also returns a function?

napari/layers/_image_layer/model.py
kne42: That's a clever way to do it but it unfortunately won't work as expected when applied to subclasses; e.g. `Pyramid` would share the same bindings as `Image`. I think we could additionally do something hacky and iterate through a class's bases, checking if one of their `class_keybindings` is the same as its own via the `is` operator. Would you like me to try that?

napari/util/keybindings.py
kne42: Ah yes, I should've used better language ;P

So I'm sure you're familiar with decorators, but to explain this aspect, it's time to go on a deep-dive *queue theme music* 😎 (let's ignore that I'm just generally bad at concisely explaining things)

So I'm sure you're familiar with simple decorators which take in a single argument, the function, do something with it, and either return the same function or a modified version. The type we're using is registration so I won't go into modifying them but you can find examples at projects like [`wrapt`](https://wrapt.readthedocs.io).

For example, here is a basic registration decorator
```python
special_func_registry = []

def special_func(func):
    special_func_registry.append(func)
    return func

@special_func
def foo(): ...
```

However, say you wanted to allow users to specify a name for the function:
```python
special_func_registry = {}

def register_special_func(name):
    def inner(func):  # this is the actual decorator
        special_func_registry[name] = func
        return func
    return inner

@register_special_func('bar')
def foo(): ...
```

What happens here is that `register_special_func('bar')` resolves to `inner` which is then applied as a decorator.

Now, say that you wanted to also use this as a regular function instead of as a decorator. You'd have to do something like `register_special_func('bar')(foo)`, which is pretty nasty.

So we revise it again:
```python
def register_special_func(name, func=None):
    if func is not None:  # used as a function
        special_func_registry[name] = func
    else:                 # used as a decorator
        def inner(func):
            register_special_func(name, func)
            return func
        return inner
```

In cases where the functional version of the decorator does not return anything, you can use `functools.partial` and have it return the decorated function:
```python
def register_special_func(name, func=None):
    if func is None:
        return functools.partial(register_special_func, name)
    special_func_registry[name] = func
    return func
```

However, if you were to try this while returning anything other than the decorated function, you would be met with disaster:
```python
def register_special_func(name, func=None):
    if func is None:
        return functools.partial(register_special_func, name)
    special_func_registry[name] = func
    return name

@register_special_func('bar')
def foo(): ...



>>> foo()
TypeError: 'str' object is not callable

>>> foo
'bar'
```

This is why I implemented it the way I did: it allows you to both use `bind_key` as a function and a decorator. As a function, `bind_key` will return any function unbound by the operation performed (whether it was by explicitly unbinding the function or overwriting it) while also not terrifyingly deleting all your functions by accident.

jni: ooooooh!

I *think* what you want is `toolz.curry`. SOOOO much simpler:

```python
import toolz as tz

@tz.curry
def register_special_func(name, func):
    special_func_registry[name] = func
    return func
```

kne42: `toolz.curry` has the same problem as using `functools.partial` as explained in the last section above, but it looks really handy, will have to try it out in future decorators!

sofroniewn:
nice job getting this finished @kne42! our critical tests are passing, i will merge

Issue #369, opened 2019-06-26
Rename app_context


sofroniewn:
## ❓ Questions and Help

It was floated at the last meeting that we should rename `app_context` to something without the underscore like just `context` and expose it at the top level of napari, not just from `napari.util`. Putting this here to foster discussion / keep track of this change

jni:
imho it should be called app or even qtapp (since it's qt-specific):

```python
import napari
from skimage import data

# Not needed in IPython when launched with --gui=qt or using %qt magic
with napari.qtapp():
    napari.view(data.coins())
```

`context` by itself isn't right. It's the equivalent of calling `layerslist` just `list`.

sofroniewn:
Fair - how about `qtcontext` instead of `qtapp`?

Or maybe `qtgui` or `guiqt` or `qt_gui` or `gui_qt`- in someways that last one mirrors the IPython usage the best which is `%gui qt`

@freeman-lab, @kne42 ?

jni:
So, I feel like there's two conflicting issues:

`with napari.qtgui():` or `with qtgui():` *sounds* better in English than `with gui_qt()`, at least to me (?). The ordering is now different from IPython but I think that's not a big deal — `gui=qt` is already different from `gui_qt`. In one you're saying "give me a Qt GUI" and in the other you're saying "make the GUI be of type Qt". From a readability perspective, I think this is better, and I think that it is clear *enough* that they are doing the same thing.

However, I'm open to `gui_qt` if people feel strongly that it's better.

Regarding @kne42's suggestion that this should be `with qt_eventloop()`, I mean, sure, that's technically what we're doing, but in the end I think that this is an implementation detail that users shouldn't need to know about. They just need to know that we are "enabling" the use of a Qt GUI. I see the appeal, but I think it's more important to match IPython's terminology than it is to be "correct" here.

sofroniewn:
I'm open to either `qtgui` or `gui_qt`. I'd probably vote for `gui_qt`, but one thing to consider though is the file that contains this might make more sense as `qtgui.py` instead of `gui_qt.py` as the latter naming convention is similar to how @jni plans to refactor our qt views in #368

sofroniewn:
[...]

PR #390, opened 2019-07-04
Refactor directory structure

### LINKED ISSUE #368: Proposal: refactor repo
This is a major refactor so I wanted to check with everyone before going ahead.

I think the napari repo is far too nested, which makes navigation a chore. Even for skimage, which is a much bigger library, we only use two levels of nesting (`skimage/package/module.py`), with only the ~~module~~ **Edit:** package level actually publicly used. I think that is the right way to do it.

Additionally, the current directory structure needlessly pairs models with views. As discussed in #115, models should have no knowledge of a view and should be able to live quite happy existences without corresponding views. Therefore, putting a `_view` folder right next to each model is a bit jarring. In my opinion, we should completely divorce our models from our views — all the views should live in their own subdirectory.

Finally, it's not really necessary to name model files `model` — it can essentially be inferred by the path in the proposal below. Additionally, it's annoying in many/most IDEs to have all the names be `model.py` or `qt_properties.py` because most IDEs don't provide the full path, or most of the path is hidden anyway because it is too long.

Ok, on to my proposed repo structure:

```
setup.py
napari/
  layers/
    __init__.py
    base.py  # formerly layers/_base_layer/model.py
    image.py  # formerly layers/_image_layer/model.py
    vectors.py
    ...
  _qt/
    __init__.py
    app_context.py
    base_qt.py
    dims_qt.py  # formerly components/_dims/view.py
    layerlist_qt.py  # formerly components/_layers_list/view.py
    image_qt.py  # formerly layers/_base_layer/view/{controls/properties}.py
    ...
    window_qt.py  # formerly components/_window/view.py
  components/
    dims.py  # formerly components/_dims/model.py
    layers_list.py  # formerly components/_layers_list/model.py
    viewer_model.py  # formerly components/_viewer/model.py
  view_function.py
  viewer.py
```

Things I like here:

- the core napari data model is very easy to extract.
- if we decide we hate Qt one day that's very easy to cast aside
- if we want to a web viewer that gets added as a directory and nothing else
- the repo is easier to navigate e.g. on github because there's a more limited nesting level.

If others agree I'm happy to give this a whirl.

jni:
# Description

Resolves #368 

Whew! That was a slog. BUT tests are passing! There's about a million more things that could be done, but for now, all non-vispy Qt stuff is in `_qt` *except* for the main Viewer (ie QtViewer + Window) class, and the view function, and the qtgui ;) context manager. The last of these I'll leave for immediately after this PR, as it's big enough already. The other two... Well, I can move them to `_qt`, but they feel integral enough to napari that right now I kinda like them there. It's the "default view", if you will. Happy to change this though, if people feel strongly about it.

I imagine that reviewing this using the GitHub view will be a nightmare, and the best way would be to follow the link to my branch and navigate the directory structure there.

## Type of change

# References

# How has this been tested?

## Final checklist:


sofroniewn (changes requested):
@jni - great job getting this going. I have a couple requests and comments bellow

- [x] why do we have a mix of things named `qtdims.py` and `image_qt.py`, is it because the later file contains two `qt` classes? I find the difference strange though and would rather settle on one - for example `qt_dims.py` and `qt_image.py`. Right now `qt_main_window.py` also has the `_` but `qtlayerslist` doesn't. I'll also note in the components folder we have an `_` between `layers_list`, and so would rather see `qt_layers_list.py`. I'd also prefer to see `qt_range_slider.py` and get rid of the `range_slider` folder. Similarly I think we can drop the `qt_viewer` folder move the three files inside to be called `main.py` -> `qt_viewer.py`, `controls.py` -> `qt_controls.py`, and `buttons.py` -> `qt_layer_buttons.py`. At that point every file in that folder would be `qt_*.py` apart from the `utils.py` and our `tests` folder and `__init__.py`

- I will post in more detail on the app_context thread #369 , but will note here as well in case it influences how you're thinking about the above, that I'm now leaning pretty strongly to `with gui_qt` instead of `with qtgui` for the sole reason that it will be less confusing for beginners to keep `with gui_qt` and `%gui qt` straight instead of having to remember two different orders of the words.

- [x] inside our `components` folder I think we said we can drop the extra layer of folders, so move `_layers_list/model.py` -> `layers_list.py`, `_viewer/viewer_model.py` -> `viewer_model.py`, `_dims/model.py` -> `dims.py`, `_dims/_constants.py` -> `_dims_constants.py` and then create one `tests` folder with all the tests.

- I like having one folder per layer in the `layers` folder, but we can always revisit in the long run if we want to make even more flat.

- I'm also ok with `Viewer` and `view_function` living on the top directory now. That can also get revisited in the long run if necessary, especially now that we're just having the single repo for our "app" and "viewer"

jni:
@sofroniewn thanks for the review! Much smaller to-do list than I expected. =D

The difference between `image_qt` and `qt_dims` is that the `_qt` ones are layers. But I agree, how were you supposed to know that? :joy: I'll just normalise the name and we can decide later whether we want to stuff the layers into a layers folder.

jni:
btw I feel like we had this conversation already, but I don't remember why we ended with this... why is it layerslist instead of layerlist? I feel like generally it's "item-container" not "items-container", e.g. apple cart, matchbox, file cabinet... So I wouldn't mind renaming layerslist to layerlist if you and others agree.

sofroniewn:
Renaming to layerlist is fine, go ahead and include in this PR

jni:
ahhhh. It's so nice now. =) =) =)

sofroniewn (approved):
Looks great - will merge!

PR #392, opened 2019-07-05
Rename app_context gui_qt

### LINKED ISSUE #369: Rename app_context

jni:
# Description

Fixes #369 

## Type of change

# References

# How has this been tested?

## Final checklist:


sofroniewn (changes requested):
Nice @jni, one minor fix and then one request.

- [x] The minor fix is for some reason `app_context` didn't get replaced inside `examples/set_colormaps.py` and so that test is failing

- [x] The request is to expose `gui_qt` at the `napari` level and not just `napari.util` level and make all the usage in our examples either
```python
from napari import Viewer, gui_qt
with gui_qt():
    Viewer()
```
 or 
```python
import napari
with napari.gui_qt():
    napari.Viewer()
```
I think a mix of those two is fine, or you can just pick one, but regardless I think this is cleaner and simpler than having to go into `napari.util` every single time

sofroniewn:
If I remember correctly @jni you are on the airplane today (and tomorrow ?!) so I took the liberty of making the changes I requested above and moving the `event_loop.py` into our `_qt` folder. I also updated the readme.md. I think it's quite nice now and will merge after the tests pass, if there's anything you don't like we can easily revisit it this week

sofroniewn (approved):

jni:
Thanks for finishing this up @sofroniewn! Indeed I'm currently in the air between LAX and AUS! See you when you get in! =) I'm +/-1 about putting gui_qt in there (it is definitely a utility rather than core functionality), but I can live with it. =)

PR #393, opened 2019-07-07
Test labels layer

sofroniewn:
# Description
This PR adds tests and doc strings for the labels layer. It tries to clean up some functions a little too and adds a couple convenience key bindings, but it doesn't change any functionality.

## Type of change

# How has this been tested?

- [x] adds `napari/layers/labels/tests/test_labels.py`

## Final checklist:


napari/layers/image/image.py
```diff
@@ -282,9 +282,14 @@ def _update_thumbnail(self):
             self._thumbnail_shape[:2], image.shape[:2]
         ).min()
         if self.multichannel:
-            downsampled = ndi.zoom(
-                image, (zoom_factor, zoom_factor, 1), prefilter=False, order=0
-            )
+            with warnings.catch_warnings():
+                warnings.simplefilter("ignore")
```
jni, line 286: Just fyi, I fixed this in scipy master just a couple of days ago. ;)

napari/layers/image/image.py
```diff
@@ -115,7 +115,10 @@ def __init__(
         with self.freeze_refresh():
             # Set data
             self._data = image
-            self.metadata = metadata
+            if metadata is None:
+                self.metadata = {}
+            else:
+                self.metadata = metadata
```
jni: `self.metadata = metadata or {}` =)

napari/layers/image/image.py
jni: https://github.com/scipy/scipy/pull/10396

napari/layers/labels/labels.py
```diff
@@ -9,46 +10,97 @@
...
-        label_image,
-        meta=None,
-        *,
-        name=None,
+        labels,
+        metadata=None,
         num_colors=50,
+        seed=0.5,
         opacity=0.7,
+        *,
```
jni: imho this `*` should be much earlier in the list. The rule of thumb I use is "If I saw a call with just the positional arguments, would I know immediately what they mean?" So, `copyfile('path/to/src', 'path/to/dst')`, sure, but `add_labels(label_image, None, 72, 0.8)` is just terrible. I think it should come immediately after the label_image.

napari/layers/labels/labels.py
```diff
@@ -61,13 +113,22 @@ def __init__(
...
-        self._meta = meta
-        self.interpolation = 'nearest'
-        self.colormap_name = 'random'
-        self.colormap = colormaps.label_colormap(num_colors)
+        self._data = labels
+        self._data_view = np.zeros((1, 1))
+        if metadata is None:
+            self.metadata = {}
+        else:
+            self.metadata = metadata
```
jni: As above, `self.metadata = metadata or {}`

jni: Note: the above expression is somewhat controversial, but I really like it. Happy to be overruled if it is deemed too arcane.

napari/layers/labels/labels.py
```diff
@@ -534,13 +491,17 @@ def _update_thumbnail(self):
...
-            self._thumbnail_shape[:2], self._image_view.shape[:2]
+            self._thumbnail_shape[:2], self._data_view.shape[:2]
         ).min()
-        downsampled = np.round(
-            ndi.zoom(self._image_view, zoom_factor, prefilter=False, order=0)
-        )
-        downsampled = self.raw_to_displayed(downsampled)
-        colormapped = self.colormap.map(downsampled)
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore")
```
jni, line 500: Maybe add a comment that this warning filter can be removed with scipy 1.4?

napari/layers/labels/tests/test_labels.py
```diff
@@ -0,0 +1,280 @@
+import numpy as np
+from xml.etree.ElementTree import Element
+from vispy.color import Colormap
+from napari.layers import Labels
+
+
+def test_random_labels():
+    """Test instantiating Labels layer with random 2D data."""
+    shape = (10, 15)
+    data = np.round(20 * np.random.random(shape)).astype(int)
```
jni: Use np.random.randint?

napari/layers/labels/tests/test_labels.py
```diff
@@ -0,0 +1,280 @@
...
+    assert layer.range == tuple((0, m, 1) for m in shape_b)
+    assert layer._data_view.shape == shape_b[-2:]
+
+
+def test_changing_labels_dims():
+    """Test changing Labels data including dimensionality."""
+    shape_a = (10, 15)
+    shape_b = (20, 12, 6)
+    data_a = np.round(20 * np.random.random(shape_a)).astype(int)
+    data_b = np.round(20 * np.random.random(shape_b)).astype(int)
```
jni: as above, np.random.randint

napari/util/misc.py
```diff
@@ -227,21 +227,58 @@ def slice_image(data, indices, multichannel=False):
         Indices to slice array at.
     multichannel : bool, optional
         Flag if image is multichannel, for example RGB or RGBA.
+    return_indices : bool, optional
+        Flag if returning sliced image data or a slice tuple.
```
jni: I prefer using a different function here. It's generally considered bad practice to change the return type based on input values. It's also unclear from the title why you would use a function to put in indices and get out indices, so probably what you want is a helper function, `clip_indices`, that gets called by `slice_image`?

napari/util/misc.py
```diff
@@ -227,21 +227,58 @@ def slice_image(data, indices, multichannel=False):
...
+    else:
+        sliced = np.asarray(data[tuple(indices)])
+
+    return sliced
+
 
-    return data_view
+def interpolate_coordinates(old_coord, new_coord, brush_size):
+    """Interpolates coordinates between old and new, useful for ensuring
+    painting is continous. Depends on the current brush size
```
jni: PEP257...

jni: also, continUous

jni (commented):
@sofroniewn man I love this so much. =D I've left a few comments that I hope should be easy to fix!

napari/layers/labels/labels.py
sofroniewn: added

napari/layers/image/image.py
sofroniewn: comment added about removal in `0.14`

napari/layers/image/image.py

napari/layers/labels/labels.py
sofroniewn: agreed, fixed here and for images

napari/layers/labels/labels.py
sofroniewn: it's fine for me, i like that it is one line. fixed

napari/layers/labels/tests/test_labels.py
sofroniewn: fixed

napari/layers/labels/tests/test_labels.py
sofroniewn: fixed

napari/util/misc.py

napari/util/misc.py
sofroniewn: i've dropped this function, wasn't really needed

sofroniewn:
@jni I've made the fixes you have requested. If you approve and the tests pass can you merge. I'm working on a similar PR for the `Points` layer right now

napari/components/viewer_model.py
```diff
@@ -320,30 +320,30 @@ def add_image(self, image, *args, **kwargs):
...
             luminance images. If not passed is calculated as the min and max of
             the image.
-        clim_range : list (2,), optional
+        clim_range : list (2,), keyword-only
             Range for the color limits. If not passed is be calculated as the
             min and max of the image. Passing a value prevents this calculation
             which can be useful when working with very large datasets that are
             dynamically loaded.
-        interpolation : str, optional
+        interpolation : str, keyword-only
```
jni, line 346: note that optional and keyword-only are orthogonal concerns.

jni (approved):

PR #396, opened 2019-07-08
Test vectors

sofroniewn:
# Description
This PR adds tests for the vectors layer, improves the doc strings and cleans up the code a tiny bit, but doesn't change any functionality.

One question is - should the input parameters be 
`edge_width`, `edge_color`, and `length` for the width, color, and the multiplicative length factor for the vectors or should they be something else. They used to just be `width`, `color`, and `length` but I added `edge_` to make the parameters the same as for the `points` and `shapes` layer, though you could argue that for the `points` layer the parameters do different things and that in the vectors layer we don't have a `face` and an `edge` so it is just confusing. I'm open to suggestions - personally I like the consistency - but we can change it. Thoughts @bryantChhun @kevinyamauchi @jni?
 
## Type of change

# How has this been tested?

- [x] adds `napari/layers/vectors/tests/test_vectors.py`

## Final checklist:

### DIFF
--- a/examples/add_vectors.py
+++ b/examples/add_vectors.py
@@ -20,13 +20,13 @@ with napari.gui_qt():
     layer.colormap = 'gray'
 
     # sample vector coord-like data
-    n = 1000
+    n = 200
     pos = np.zeros((n, 2, 2), dtype=np.float32)
     phi_space = np.linspace(0, 4 * np.pi, n)
     radius_space = np.linspace(0, 100, n)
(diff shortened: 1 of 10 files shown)


AhmetCanSolak (approved):
it is working as expected and to me looks good.

bryantChhun:
I'm ok with "edge_".  It's clear enough to me that this shape isn't exactly like the others.

PR #400, opened 2019-07-09
Test shapes

sofroniewn:
# Description
This PR adds tests for the shape layer, improves doc strings, standardizes some things. One big change is that `layer.data` now returns a list of arrays instead of the complicated dictionary of shape objects. All of that is kept behind the scenes. I've still got more tests to write, so this isn't ready yet, but wanted to put it here for now.

## Type of change

# How has this been tested?

- [x] adds `napari/layers/shapes/tests/test_shapes.py`

## Final checklist:


sofroniewn:
I've made more updates, added more tests including for all shapes, 3d, all properties and some more - this PR is now ready for review

jni (approved):
With the caveat that I'm rapidly scanning this code while listening to talks and it's not the most rigorous review, LGTM! :joy:

jni:
Whoa, Linux 3.7 build has failed...?

sofroniewn:
ah - ok what is happening with the linux build failure is that i'm generating random polygons for testing and occasionally that causes a triangulation error with vispy code in our tests.

jni:
That seems... serious?

btw have you met David Hoese yet? He's around but not sprinting, so you should probably sit with him tonight to chat about vispy release and future directions!

jni:
also, short term fix: set random seed before generating those shapes. It would be extra awesome if you found a *failing* seed and marked that as an allowed failure, so we can figure it out later.

jni:
https://docs.pytest.org/en/latest/skipping.html

sofroniewn:
I've added a setting of the seed, hopefully this will fix it for now (just put one line at the top of the test - is that right @jni?)

sofroniewn:
Critical tests are now passing - i will merge

PR #401, opened 2019-07-10
add viewer model tests

sofroniewn:
# Description
This PR adds tests to the model viewer. Some are failing right now - but they should have been fixed in #400 which should be merged first

## Type of change

# How has this been tested?

- [x] add `napari/components/tests/test_viewer_model.py`

## Final checklist:

### DIFF
--- a/napari/layers/pyramid/tests/test_pyramid.py
+++ b/napari/layers/pyramid/tests/test_pyramid.py
@@ -4,6 +4,10 @@ from vispy.color import Colormap
 from napari.layers import Pyramid
 
 
+# Set random seed for testing
+np.random.seed(0)
+
+
 def test_random_pyramid():
     """Test instantiating Pyramid layer with random 2D data."""
     shapes = [(40, 20), (20, 10), (10, 5)]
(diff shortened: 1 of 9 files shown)


sofroniewn:
merge #400 in and fixed tests - can you review @jni

napari/layers/pyramid/tests/test_pyramid.py
```diff
@@ -4,6 +4,10 @@
 from napari.layers import Pyramid
 
 
+# Set random seed for testing
+np.random.seed(0)
```
jni, line 8: Typically these get set within each function, because things like parallel running of tests will break this. I think there's a way to set it once using a fixture but I've never used that.

sofroniewn: ah should i move to every example?

jni: ideally yes, sorry :grimacing:

jni: I fear if you don't we will end up with more weird stochastic failures, and we have enough of those! :joy:

sofroniewn: it's ok - will do now

freeman-lab (approved):
@sofroniewn  made me do it :) (apparently @jni  approved verbally but didn't click)

Issue #379, opened 2019-07-01
File IO


sofroniewn:
## 🚀 Feature
With #356 and #378 we will have the capacity for File IO from the CLI, the menubar, and drag and drop. I’m creating this issue for us to start discussing file opening and saving in more detail.

Ideally I think file IO from the CLI, the menubar, and drag and drop should all function in as similar way as possible (though from the menubar / CLI you might have a little more flexibility from the exact command you pick or any extra flags you provide). Again, ideally I think we want support as many file types / layer types from a single command and infer what the user is expecting without additional input - though if that is not practical we might need to pop up a dialog box allowing the user to appropriately set some additional input metadata (especially for drag / drop files).

Right now we only support adding images and make no use of metadata beyond guessing if it is `multichannel` or not (i.e. rgb / rgba). We also only read things that are supported by skimage.

For file opening things to think about are:

- [x] Supporting lazy loading with dask and zarr - by making them optional dependencies. This seems highly valuable.

- [ ] Supporting more complex (proprietary) image filetypes. I think we’d do this through optional dependencies on things like `python-bioformats`. For me this is a lower priority right now, but it’s good to be thinking about and maybe worth doing if easy.

- [ ] Parsing metadata from standard image file types, such as OME-tif to return things like colormaps. This also seems valuable, and something which we can grow into as we become able to do more with metadata (like label axes or apply scale bars)

- [x] Determine whether an input image should be rendered as a labels_layer and connect it to the `add_labels` function if appropriate. This seems very useful, but right now we’d have to guess and it’s unclear how to completely disambiguate an int geryscale image from a labels layer 

- [x] Support opening image pyramids through zarr and other pyramid filetypes and connect them to the `add_pryamid` function. This seems desirable, though the representation of multi resolution in zarr right now is pretty ad-hoc.

- [x] Support opening `points` using a custom csv representation, np.array, and svg representation. Seems valuable, disambiguation with other shapes, and capturing all metadata might be hard.

- [ ] Support opening `vectors` using a custom csv representation, np.array, and svg representation. Seems valuable, disambiguation with other shapes, and capturing all metadata might be hard.

- [x] Support opening `shapes` using a custom csv representation, np.array, and svg representation. Seems valuable. Seems valuable, disambiguation with other shapes, and capturing all metadata might be hard.

- [ ] Look into supporting reading geo-JSON for shape objects (likely through an optional dependency)

[...]

PR #405, opened 2019-07-11
3D volume viewer - volume layer

pranathivemuri:
# Description

<!- would like to add . a number of features like different camera, axis on off, pixel size, colorbar, threshold - but right now stuck on figuring each one out-->

## Type of change

# References
example resource used from vispy - https://github.com/vispy/vispy/blob/master/examples/basics/scene/volume.py

# How has this been tested?

- [x] example: the test suite for my feature covers all test cases like 3D volume with zeros, ones, random array, 4 channel with rgb and with rgba, colormaps etc

## Final checklist:


pranathivemuri:
Screen Shot 2019-07-11 at 4 45 54 PM

sofroniewn (changes requested):
@pranathivemuri this is really cool to see. Looking the code over and reflecting on it a bit yesterday I think we are going to want to try and integrate in the volume viewing into the image layer. The main reason for this is to allow flexibility between rapidly switching between 2D and 3D views - i.e. if someone loads some 3D image data and does add_volume and wants to look at it in 2D then now it will be very hard. My apologies for now advocating this approach when you asked me the other day, but seeing it now makes it more clear to me. Thoughts @royerloic?

I have a few more questions - one is what is special about your `best_3d_colormaps`? Why don't our normal colormaps work, with changing the `blending` mode (exposed in the drop down, might not be the right thing) or the render `method` (exposed on the vispy Volume visual). Also why would the `clim` code (and then clim sliders) work from the `Image` layer with the volume visual? It seems from the vispy docs it should all work fine http://vispy.org/visuals.html#vispy.visuals.VolumeVisual.

As to how to integrate with the Volume viewing with the `Image` layer. Ideally all the current parameters in the image layer like clim / colormap would still function as before. We'd need to change the thumbnail method, get status, and xml_to_list, but those can be ignored for now. Soon we will have the Vispy Image node pulled out from the image model file, at that point the image layer shouldn't require any modification to handle the volume, and only the vispy view needs to be able to support both. For now though you can import both the Volumes Visual and the Image Visual in the image layer.

The other thing that will need to happen is changing how the dims model gets interpreted. I'll make some comments in the code walking you how to through this.

Again, really impressive to see this come so far so quickly. Does @jni or @kne42 have anything they'd like to add?

napari/_qt/qt_viewer.py
```diff
@@ -285,6 +284,61 @@ def dropEvent(self, event):
...
+    def _add_files(self, filenames):
+        """Adds an image/volume layer to the viewer.
+
+        Whether the image is multichannel is determined by
+        :func:`napari.util.misc.is_multichannel`.
+
+        If multiple images are selected, they are stacked along the 0th
+        axis.
+
+        If there is only 1 filename opened and it is a 3D numpy array saved
```
sofroniewn: We need to figure out our file loading in general, see #379 for discussion, but I don't like here how the array is being forced to be viewed as 3D when the user may want to view it sliced and right now we have no way to toggle between the two options. Can you remove all the file loading code and revert file loading to how it was. We will have to make followup PRs to both allow loading of npy, npz files in general and then we'll also need a way for the user to decide either during or after loading if they want the 2D view or 3D view.

In general we've added support for features through the programatic API before adding support for file IO, for example there's also no current way to load shapes or points data from disk, but we'll be working on that too in `0.2`

napari/_qt/qt_viewer.py
```diff
@@ -285,6 +284,61 @@ def dropEvent(self, event):
...
+
+        if len(filenames) == 1 and (
+            filenames[0].endswith(".npy") or filenames[0].endswith(".npz")
+        ):
+            volume = load_numpy_array(filenames[0])
+
+            # Create an XYZaxis visual
+            self.axis = XYZAxis(parent=self.view)
+            self.axis.transform = STTransform(
+                translate=(50, 50), scale=(50, 50, 50, 1)
```
sofroniewn: why the magic numbers of translate (50, 50) and scale (50, 50, 50, 1)? Are these related to the shape of the 3D volume that you're trying to view, in which case they should be computed based on it, or are they independent of it and will work for all volumes? Either way can you add a comment

napari/_qt/qt_viewer.py
```diff
@@ -285,6 +284,61 @@ def dropEvent(self, event):
...
+            # flip y-axis to have correct alignment
+            self.view.camera.flip = (0, 1, 0)
+            self.view.camera.set_range()
+            self.view.camera.viewbox_key_event = viewbox_key_event
+
+            # TO DO: Remove
+            self.viewer._view = self.view
+
+            self.viewer.add_volume(
+                volume, multichannel=is_multichannel(volume.shape)
```
sofroniewn: multichannel will always be `FALSE` for volumetric data as vispy doesn't support rendering RGB or RGBA data.

napari/components/viewer_model.py
```diff
@@ -409,6 +409,43 @@ def add_pyramid(self, pyramid, *args, **kwargs):
...
+        name : str, keyword-only
+            Name of the layer.
+
+        Returns
+        -------
+        layer : :class:`napari.layers.Volume`
+            The newly-created volume layer.
+        """
+        layer = layers.Volume(volume, *args, **kwargs)
+        self.dims = Dims(3)
```
sofroniewn: you should need this - instead for now I'd try
```python
layer = layers.Volume(volume, *args, **kwargs)
self.add_layer(layer)
self.dims.set_display(-2) = `True`
```
After you call the `set_display` the dims.indices should be
```python
(..., slice(None, None, None), slice(None, None, None), slice(None, None, None))
```
which will allow you to do the 3D slicing on the `Image` layer

napari/layers/volume/volume.py
```diff
@@ -0,0 +1,224 @@
...
+            # Set colormap
+            self._colormap_name = ''
+            self.colormap = colormap
+
+            # Set update flags
+            self._need_display_update = False
+            self._need_visual_update = False
+
+            # Re intitialize indices
+            self._indices = (
```
sofroniewn: This should still have the ndims code as before to handle 4D datasets where we want the last 3 to be as a volume

napari/layers/volume/volume.py
```diff
@@ -0,0 +1,224 @@
...
+        self.events.colormap()
+
+    @property
+    def colormaps(self):
+        """tuple of str: names of available colormaps."""
+        return tuple(self._colormaps.keys())
+
+    def _set_view_slice(self):
+        """Set the view given the indices to slice with."""
+        indices = (
```
sofroniewn: this shouldn't be needed anymore, you should be able to use `self.indices` after having done `dims.set_display` in my comment above

sofroniewn:
I've gone through and made comments now, there's really a fork in the road ahead though where either you start making the changes above to the `Volume` layer or you switch to adding functionality to the `Image` layer. I know that it will require some refactoring when we pull out the vispy nodes but I think at this point you'll already be able to get it integrated into the `Image` layer which is where we ultimately want to be. What are your thoughts on this @pranathivemuri? 

To integrate with the `Image` layer I'd first start a new example in our `examples` folder called `3D_volume.py` we'll have to find a nice 3D dataset to look at, as it's not clear to me that skimage has one - @jni? for testing maybe you can make something fake in numpy or use the blobs 3D from skimage. You'll also want to make `nD_volume.py` which will have 4D data where the last 3 are a 3D volume and the first one is a slicing axis (imaging like time) and you'll be able to move the slider and show the different volumes. Here using a list of 3D blobs from skimage as we use in the `nD_image.py` example is probably the way to go.

Within both those examples after successful integration you should be able to call the following two lines of code
```python
viewer.add_image(data)
viewer.dims.set_display(-2) = True
```
and then everything should switch to 3D rendering mode. If you later typed `viewer.dims.set_display(-2) = False` you should switch back to only 2D slicing mode. Ideally for these examples you could create a custom key binding that when you called it switched back and forth between those modes.

To achieve this you'll still need to create the TurntableCamera as you did before, but maybe now you'll want to create a dictionary on the `QtViewer` called `cameras` where you instantiate and store both the `PanZoomCamera` and the `TurntableCamera` so that you'll be able to switch back and forth between them.

The way you can tell which one to use inside the `QtViewer` is by calling `np.sum(self.viewer.dims.display)`

Inside the `Image` layer you'll need to now import both the `Image` and `Volume` visual. It's a little unclear to me right now what the best way to enable toggling back and forth between them will be like. Maybe we should let @kne42 and @royerloic weigh in here before going on, as they might have reasons for wanting to actually keep the `Volume` layer separate.

napari/_vispy/visuals/volume.py
```diff
@@ -501,6 +502,11 @@ def clim(self):
     def cmap(self):
         return self._cmap
 
+    @clim.setter
```
sofroniewn: ah i didn't realize this wasn't on our backport of the vispy volume, as you might see in #406 we're in process of updating to `0.6` which seems to have some clim support through it's `set_data` - https://github.com/vispy/vispy/blob/cb5706f3ec0a7298a91692f2c73edd3ea3b9dcdc/vispy/visuals/volume.py#L451

pranathivemuri: cool! I might have to refactor then after its merged

pranathivemuri:
colormaps


maybe we can keep all the colormaps, but I think fire and grays colormaps have the best contrast for 3D rendering method "translucent"

sofroniewn:
@pranathivemuri looking at the colormaps, `gray` and `red` don't look as good, but they don't look terrible and adjusting the `clim` values might help. Does the slider on the left work?

I'd rename `grays` to `gray_enhanced` or something else that makes it a little more clear and then just have the one list.

What happens if you apply `gray_enhanced` or `fire` to a simple 2D image (say the `add_image.py` example). Does that work normally? (Doesn't have to look great, just has to not crash).

pranathivemuri:
seems like for the "mip" rendering mode since its a maximum intensity projection the "red", "green" and "gray" channels are looking better
I want to keep all the colormaps including "fire" and "grays"

The defaults for the colormap is "grays" and the rendering method is "translucent".  Let me know what is preferred

colormaps

pranathivemuri:
fire and grays work on 2D grayscale images. no colormap changes work on 2D color images.

sofroniewn:
yup colormaps don't work on `RGB` or `RGBA` data (i.e. multichannel), that's known. Thanks for testing these colormaps!

pranathivemuri:
cool, just saw an error on the console regarding the 2 new colormaps "fire" and "grays". it changes the colormaps but while updating thumbnail it raises a NotImplemented error.

NotImplementedError                       Traceback (most recent call last)
/Users/pranathivemuri/czbiohub/napari/napari/_qt/qt_image_layer.py in <lambda>(text)
     63         comboBox.setCurrentIndex(index)
     64         comboBox.activated[str].connect(
---> 65             lambda text=comboBox: self.changeColor(text)
     66         )
     67         self.grid_layout.addWidget(QLabel('colormap:'), row, self.name_column)

/Users/pranathivemuri/czbiohub/napari/napari/_qt/qt_image_layer.py in changeColor(self, text)
     89 
     90     def changeColor(self, text):
---> 91         self.layer.colormap = text
     92 
     93     def changeInterpolation(self, text):

/Users/pranathivemuri/czbiohub/napari/napari/layers/image/image.py in colormap(self, colormap)
    215         self._colormap_name = name
    216         self._node.cmap = self._colormaps[name]
--> 217         self._update_thumbnail()
    218         self.events.colormap()
    219 

/Users/pranathivemuri/czbiohub/napari/napari/layers/image/image.py in _update_thumbnail(self)
    321             if color_range != 0:
    322                 downsampled = (downsampled - low) / color_range
--> 323             colormapped = self.colormap[1].map(downsampled)
    324             colormapped = colormapped.reshape(downsampled.shape + (4,))
    325             colormapped[..., 3] *= self.opacity

/usr/local/lib/python3.6/site-packages/vispy/color/colormap.py in map(self, item)
    233 
    234         """
--> 235         raise NotImplementedError()
    236 
    237     def __getitem__(self, item):

NotImplementedError: 

pranathivemuri:
figuring out how to debug it right now

pranathivemuri:
https://github.com/vispy/vispy/blob/master/vispy/color/colormap.py#L1058 Fire and Grays colormap are already included in the current master in vispy . maybe updating to 0.6 version also might prevent this error

sofroniewn:
@pranathivemuri that's nice they are in `0.6` we should have #406 merged tomorrow so I wouldn't waste time on this then. also when pasting in code snippets if you put them in a block of  "\```python" and then "```" they will render better

royerloic:
@pranathivemuri , so cool!
Looks great already!

I am using nanapri every day, and 3D support is going to be just amazing.

A few points:

i) We should ideally rotate the volume around its center, not around a corner...

ii) I think a volume layer is fine for a first pass, but I agree that the best solution long term
    is to integrate the volume rendering in the image layer. All the reasons provided by @sofroniewn 
   are valid and important... Waiting for the image layer to be fully refactored is a good approach. 

iii) Wwe will need to think hatrd about the UI for switching between 2D and 3D, we had some basic ideas early on. The dims model can accommodate that on the model side of things.

iv) About having both the image and volume visual in the scene graph: First, there might be a flag somewhere that lets us activate or deactivate a visual in the graph. If not, we can simply detach/attach on demand while keep references of the visual that is detached. the volume viewer should only be instanced the first time 3D viewing has been requested but not otherwise -- 3D viewing can consume a lot of GPU ressources... 

v) A good reason for having 2D and 3D viewing tightly integrated is that we can support slice&volume viewing. See how it is done in Spimagine: https://vimeo.com/126597994 at 1:28. Something similar would be easy to implement with vispy's scene graph.   Overall, I like a lot Spimagine design (author: Martin Weigert) and we can definitely learn a few things there... 

v) 3D viewing will not work for very large 3D slices, already 1024x1024x1024 might be two much for most graphics cards. So we might want to downscale on the CPU before sending the array for viewing when we exceed the available GPU memory (which can be queried). The problem of constrained GPU ressources is barely noticeable  when doing 2D rendering, but  very dramatic as sson as we start allocating 3D tectures on the GPU (which is what vispy does via OpenGL). This will rapidly cause crashes or unresponsive UIs and so must be eventually addressed...  
In ClearVolume we had tons of tricks to make the rendering very smooth. We can revisit some of these techniques down the line.   I am particularly excited that with Vispy's scene graph we can relatively easily do a octree like adaptive LOD volume rendering. All probably premature right now... but very exciting. 

All in all, pretty amazing stuff already!
Looking forward to see more!

napari/_qt/qt_viewer.py
```diff
@@ -278,6 +280,15 @@ def dragEnterEvent(self, event):
         else:
             event.ignore()
 
+    def _open_files(self):
```
sofroniewn: can you move `_open_files` and `_add_files` back up to where they were before, it will make the diff smaller and easier to understand the changes

napari/components/tests/test_viewer_model.py
```diff
@@ -23,6 +23,16 @@ def test_add_image():
     assert np.all(viewer.layers[0].data == data)
 
 
+def test_add_volume():
```
sofroniewn, line 26: love it!

napari/view_function.py
```diff
@@ -48,10 +48,15 @@ def view(
     the returned :class:`napari.Viewer` object.
     """
     viewer = Viewer(title=title)
-    for image in images:
+    if len(*images) > 1:
```
sofroniewn: I don't think we want to sometimes create a volume layer using the `view` function, this is really just a convenience function for rapidly creating an image layer, and instead people should instantiate a viewer and then call `add_volume` if the want a volume, or - hopefully in not too long we'll be able to support a volume through the image layer. Can you revert this.

pranathivemuri:
we can add a button to go to menu switch camera to switch to turntable, arcball or other kind of 3D camera is my idea (I have seen this on a proprietary software called arivis I used back in my previous job). let me know what you guys think

volume layer displays as pan zoom camera with 2D slices that one can slide through by default. This functionality isn't working yet. It is opening a volume as scrollable layers but as I slide through nothing is changing and I think this is because my Dims is wrong according to the error below

``` python   File "/Users/pranathivemuri/czbiohub/napari/napari/_qt/qt_dims.py", line 115, in _update_slider
    range = self.dims.range[axis]
IndexError: list index out of range

If you suspect this is an IPython bug, please report it at:
    https://github.com/ipython/ipython/issues
or send an email to the mailing list at ipython-dev@scipy.org

You can print a more detailed traceback right now with "%tb", or use "%debug"
to interactively debug it.

Extra-detailed tracebacks for bug-reporting purposes can be enabled via:
    %config Application.verbose_crash=True ```

**Comment by pranathivemuri** [comment]
I am still not clear and would like help/suggestions on where/what I can do with the below after looking at where I set them currently in the code. I might have set them in a hacky way to make it work, not necessarily the right way. But the viewermodel, qtviewer, viewer class inheritance, attributes is currently slightly confusing and making the below attributes harder to access within other class instance

and I repeatedly see this todo - https://github.com/napari/napari/blob/master/napari/_qt/qt_viewer.py#L61 which I think correctly setting might make things easier/clearer

1) Why just 2 Colormaps are working now and the above GIFS I generated showing how the volume worked to display in red, green etc colormaps doesn't hold true anymore after upgrading to vispy 0.6.0
2) Switching cameras
3) Setting Dims

File: napari/_qt/qt_main_window.py [hunk 11]
```diff
@@ -104,12 +104,12 @@ def _toggle_menubar_visible(self):
             self._main_menu_shortcut.setEnabled(False)
 
     def _add_file_menu(self):
-        open_images = QAction('Open', self._qt_window)
-        open_images.setShortcut('Ctrl+O')
-        open_images.setStatusTip('Open image file(s)')
-        open_images.triggered.connect(self.qt_viewer._open_images)
+        open_files = QAction('Open', self._qt_window)
```
sofroniewn: can you revert all these changes

napari/_qt/qt_viewer.py
```diff
@@ -6,20 +6,23 @@
 from qtpy.QtCore import QCoreApplication, Qt, QSize
 from qtpy.QtWidgets import QWidget, QVBoxLayout, QSplitter, QFileDialog
 from qtpy.QtGui import QCursor, QPixmap
+from vispy.scene import SceneCanvas, PanZoomCamera, TurntableCamera
 from qtpy import API_NAME
-from vispy.scene import SceneCanvas, PanZoomCamera
 from vispy.app import use_app
 
+from napari.components import Dims
```
sofroniewn: you shouldn't need to import `Dims` to get this working, I'll put more detail below

napari/_qt/qt_viewer.py
```diff
@@ -50,22 +55,13 @@ def __init__(self, viewer):
...
-        self.view.camera = PanZoomCamera(aspect=1)
-        # flip y-axis to have correct aligment
-        self.view.camera.flip = (0, 1, 0)
-        self.view.camera.set_range()
-        self.view.camera.viewbox_key_event = viewbox_key_event
-
-        # TO DO: Remove
-        self.viewer._view = self.view
+        self.set_camera()
+        viewer.camera = self.view.camera
```
sofroniewn: I'd prefer to avoid adding the `camera` attribute directly to the viewer. We're going to be refactoring to remove the `view` from the `Viewer` too, but this should not be needed. If you need something from the `Viewer` to communicate with the `camera` we should be using events

napari/_qt/qt_viewer.py
```diff
@@ -111,6 +106,31 @@ def __init__(self, viewer):
 
         self.setAcceptDrops(True)
 
+    def set_camera(self):
+        print(self.viewer.dims.display)
+        print(sum(self.viewer.dims.display))
+        if sum(self.viewer.dims.display) > 2:
```
sofroniewn: Is this working as expected? It should be correct.

For future proofing, can you made the statement
```python
if ... == 2:
   # Set a 2D camera
   ....
elif ... == 3:
   # Set a 3D camera
   ....
else:
   raise an error
```

napari/_qt/qt_viewer.py
```diff
@@ -136,35 +156,40 @@ def screenshot(self, region=None, size=None, bgcolor=None):
...
+        """Adds files from the menubar."""
         filenames, _ = QFileDialog.getOpenFileNames(
             parent=self,
-            caption='Select image(s)...',
+            caption='Select image(s).../volume(npy or npz files)',
             directory=self._last_visited_dir,  # home dir by default
         )
         self._add_files(filenames)
 
     def _add_files(self, filenames):
```
sofroniewn: Can you revert all these changes, we're not ready yet to support automatically adding a volume through the drop downs. For now we should only support the `add_volume` api but make sure that works as expected

napari/layers/volume/tests/test_volume.py
```diff
@@ -0,0 +1,155 @@
...
+    np.random.seed(0)
+    data = np.round(10 * np.random.random(shape)).astype(int)
+    layer = Volume(data)
+    assert np.all(layer.data == data)
+    assert layer.ndim == len(shape)
+    assert layer.shape == shape
+    assert layer._data_view.shape == shape[-3:]
+
+
+def test_3D_volume():
```
sofroniewn, line 43: maybe make this test a `4D` volume

napari/components/viewer_model.py
```diff
@@ -409,6 +409,42 @@ def add_pyramid(self, pyramid, *args, **kwargs):
...
+        name : str, keyword-only
+            Name of the layer.
+
+        Returns
+        -------
+        layer : :class:`napari.layers.Volume`
+            The newly-created volume layer.
+        """
+        layer = layers.Volume(volume, *args, **kwargs)
+        self.add_layer(layer)
```
sofroniewn, line 450: You need to add 
```python
self.dims.set_display(-2) = True
``` 
here to enable the 3D camera

napari/layers/volume/volume.py
sofroniewn: these need to be dropped

sofroniewn:
Hi @pranathivemuri I've added some comments that will hopefully make the setting of the camera, and dims work properly now, you're still missing the call inside the `add_volume` method on the viewer where you set the 3rd axis (`-2`) to be displayed. That should solve lots of your problems hopefully.

Let's not worry about buttons to change the camera or file loading etc. right now, those can easily be added in future PRs. In general we try and keep our PRs as short and modular as possible. I'd rather review two small PRs than one really long one. I think the goal for this PR is to enable a call to `add_volume` to work with 3D / 4D data in a way that is maximally compatible with the rest of the viewer.

I'm not sure about the 2D vs 3D colormaps with the new vispy. For now then having separate 3D colormaps is fine, but we might want to work on unifying these in the future.

sofroniewn:
Oh - also can you fix the doc string the in Pyramids layer - you made a change to it in the viewer_model `date` -> `data` and they need to match exactly for the tests to pass.

In general it's much more preferable to fix bugs / errors like this that are independent of the current PR in separate a PR (especially if the current PR is a big one and will take a while to merge) as if someone else working on the pyrmaids layer now edits that doc string for another reason while we're still trying to get 3D volumes working it will create unnecessary merge conflicts.

napari/components/viewer_model.py
pranathivemuri: this does not enable the 3D camera since it the camera is initialized in QTViewer __init__

napari/_qt/qt_main_window.py
pranathivemuri: done, will push soon

napari/_qt/qt_viewer.py
pranathivemuri: changed

napari/_qt/qt_viewer.py
pranathivemuri: removed

napari/_qt/qt_viewer.py
pranathivemuri: done! camera won't work though

napari/_qt/qt_viewer.py
```diff
@@ -136,35 +156,40 @@ def screenshot(self, region=None, size=None, bgcolor=None):
...
         """
-        if len(filenames) > 0:
-            image = read(filenames)
+        image = read(filenames)
+        if len(filenames) == 1:
             self.viewer.add_image(
                 image, multichannel=is_multichannel(image.shape)
             )
-            self._last_visited_dir = os.path.dirname(filenames[0])
+        elif len(filenames) > 1:
```
pranathivemuri: @sofroniewn should I revert these changes too?

napari/_qt/qt_viewer.py
```diff
@@ -202,8 +227,24 @@ def _update_palette(self, palette):
 
     def on_mouse_move(self, event):
         """Called whenever mouse moves over canvas.
+        when axis is not None and the camera is turntable, on moving the mouse
+        one can rotate around the vertical axes.
         """
         layer = self.viewer.active_layer
+        if (
```
pranathivemuri: should i revert these changes too @sofroniewn

napari/components/viewer_model.py
pranathivemuri: it will not enable the camera

napari/layers/volume/tests/test_volume.py
pranathivemuri: why? vispy doesn't allow visualization of rgb, rgba volumes, right?

pranathivemuri: if I test a 4D volume I will get the below error

``` python     self.set_data(vol, clim)
  File "/usr/local/lib/python3.6/site-packages/vispy/visuals/volume.py", line 465, in set_data
    raise ValueError('Volume visual needs a 3D image.')
ValueError: Volume visual needs a 3D image.```

File: napari/_qt/qt_viewer.py (hunk 18 above)
Comment (sofroniewn) [reply napari/_qt/qt_viewer.py]: yes please

File: napari/_qt/qt_viewer.py (hunk 19 above)
Comment (sofroniewn) [reply napari/_qt/qt_viewer.py]: these changes are needed for the camera to work properly, right(?), and should stay - you should just revert the file IO changes

File: napari/components/viewer_model.py (hunk 17 above)
Comment (sofroniewn) [reply napari/components/viewer_model.py:450]: yes - you'll need to trigger the camera update on the dims displayed change in the QtViewer too. you can do thing using events - i'll add a comment in there

File: napari/layers/volume/tests/test_volume.py (hunk 16 above)
Comment (sofroniewn) [reply napari/layers/volume/tests/test_volume.py:43]: You need to slice the array so that you only pass a 3D array to vispy - I've left comments about this in the volume layer. I'll explain more there. This is to support volumetric timeseries data for example

File: napari/layers/volume/volume.py (hunk 5 above)
Comment (sofroniewn) [reply napari/layers/volume/volume.py]: please fix this to enable the `4D` example to work 
```python
            self._indices = (0,) * (self.ndim - 3) + (
                slice(None, None, None),
                slice(None, None, None),
                slice(None, None, None),
            )
```

napari/layers/volume/volume.py
sofroniewn: please make this match the image layer, but with the right number of dims
```python
        indices = list(self.indices)
        indices[:-3] = np.clip(
            indices[:-3], 0, np.subtract(self.shape[:-3], 1)
        )
```
this should make your `4D` test example pass

sofroniewn:
I've made some more comments inline - as you note the camera now isn't getting set - in order to do this you'll need to hook-up an event from the dims model to the function in the QtViewer

you'll first need to add a dedicated `display` event to the Dims model - see where the `axis` event and the `ndim` events are, and then modify the display setter to emit the `display` event instead of the `axis` event (which is used in other contexts). you'll then need to connect the `dims.events.display` to the `set_camera` method. Ideally you would not create and destroy the camera everytime that gets called, but check if the correct camera is being used, and if not create one.

After this all the tests should be passing and the functionality should behave as expected - calling `add_volume` with 3D or more arrays should lead to a volume being rendered and you should be able to add multiple volumes without reseting the camera every time.

napari/_qt/qt_viewer.py
pranathivemuri: reverted

napari/layers/volume/tests/test_volume.py
pranathivemuri: added that and pushing

pranathivemuri:
mostly done seems like tests are passing too 

1) For volumetric time frames display, do we wanna display different volumes as one slides through using the slider scale in the bottom - If so, the current code isn't doing that. It is just adding layers tot the same view. I am not sure how to correct that behaviour yet
scale_translate_multiple_volumes


2) And the scale, translate transform I am setting isn't being captured by volume visual and displaying the volume correctly, I had to use my zoom button to see something it was all 255 pixels first like in the above gif and then I zoomed in with my touchpad to see the actual volume. I am trying to correct this now.

Screen Shot 2019-07-16 at 3 18 31 PM

3. Remove axis and slider on delete of a layer

4. The clim updates the color limits on the left slider, but it doesn't change/scale the volume

examples/add_volume_series.py
```diff
@@ -0,0 +1,18 @@
+"""
+Display one 3-D volume layer using the add_volume API
```
sofroniewn: can you rename this example `nD_volume` and adjust the doc string to say you are using `4D` data representing a volumetric timeseries.

examples/add_volume_series.py
```diff
@@ -0,0 +1,18 @@
...
+
+from skimage import data
+import napari
+
+
+with napari.gui_qt():
+    blobs = data.binary_blobs(length=128, blob_size_fraction=0.05, n_dim=4)
+    blobs = blobs.astype(float) * 255
+    viewer = napari.Viewer()
+    for i in range(blobs.shape[0]):
```
sofroniewn: you should remove the for loop and just make one call 
```python
layer = viewer.add_volume(blobs)
```
This should then generate a single layer that has a slider at the bottom of the screen. Using the slider should move you through the data

napari/components/viewer_model.py
```diff
@@ -53,7 +53,7 @@ def __init__(self, title='napari'):
 
         # Initial dimension must be set to at least the number of visible
         # dimensions of the viewer
-        self.dims = Dims(2)
+        self.dims = Dims(3)
```
sofroniewn: you should revert this and will instead need to increase the dims in the call to the volume layer. i will comment below

napari/components/viewer_model.py
```diff
@@ -409,6 +410,43 @@ def add_pyramid(self, pyramid, *args, **kwargs):
...
+
+        name : str, keyword-only
+            Name of the layer.
+
+        Returns
+        -------
+        layer : :class:`napari.layers.Volume`
+            The newly-created volume layer.
+        """
+        layer = layers.Volume(volume, *args, **kwargs)
```
sofroniewn, line 446: here you need to add
```python
if self.dims.ndim == 2:
    self.dims.ndim = 3
```
to boost the number of dimensions so that setting the display `-3` is correct

napari/_qt/qt_viewer.py
```diff
@@ -108,9 +101,36 @@ def __init__(self, viewer):
...
         self.setAcceptDrops(True)
 
+    def _update_camera(self, event):
+        if sum(self.viewer.dims.display) == 3:
+            # Set a 3D camera
+            self.view.camera = TurntableCamera(name="TurntableCamera")
+            # Create an XYZaxis visual
+            self.axis = XYZAxis(parent=self.view)
+            self.axis.transform = STTransform(
+                translate=(50, 50), scale=(50, 50, 50, 1)
```
sofroniewn: where do these `(50, 50)` and `(50, 50, 50, 1)` numbers come from? do they relate the shape of the data, are they causing the zoom problems you mention?

napari/layers/volume/volume.py
```diff
@@ -0,0 +1,245 @@
...
+        self.events.colormap()
+
+    @property
+    def colormaps(self):
+        """tuple of str: names of available colormaps."""
+        return tuple(self._colormaps.keys())
+
+    def _set_view_slice(self):
+        """Set the view given the indices to slice with."""
+        self._data_view = np.asarray(self.data[tuple(self._indices)])
```
sofroniewn: This needs to be
```python
 indices = list(self.indices)
        indices[:-3] = np.clip(
            indices[:-3], 0, np.subtract(self.shape[:-3], 1)
        )
self._data_view = np.asarray(self.data[tuple(indices)])
```
otherwise you may get errors for data > 3D

sofroniewn: if you try and slice into data that is larger than the shape

napari/layers/volume/volume.py
```diff
@@ -0,0 +1,245 @@
...
+        clim=None,
+        clim_range=None,
+        name=None,
+        **kwargs,
+    ):
+
+        visual = VolumeNode(volume, threshold=0.225, emulate_texture=False)
+        print(self._translate)
+        print(self._scale)
+        visual.transform = STTransform(
```
sofroniewn: why are you setting the `visual.transform` here? this is done automatically in the `_visual_wrapper.py`. can you delete this, the print statements, and the `_scale` and `_translate` properties on this layer. If the camera position is not right we should fix it in the QtViewer rather than here

napari/layers/volume/volume.py
```diff
@@ -0,0 +1,245 @@
...
+        self._need_visual_update = True
+        self._update()
+
+        self._data_thumbnail = self._data_view
+
+    def _update_thumbnail(self):
+        pass
+
+    @property
+    def clim(self):
```
sofroniewn, line 179: in the `add_volume.py` example the clim slider does nothing, this could be because the data is binary, can you confirm that it works in other cases / maybe figure out a way to construct blobs that arn't quite binary, say by multiplying the blobs by some random noise before sending them to napari?

napari/layers/base/base.py
```diff
@@ -66,6 +66,7 @@ def __init__(self, central_node, name=None):
         self._thumbnail = np.zeros(self._thumbnail_shape, dtype=np.uint8)
         self._update_properties = True
         self._name = ''
+        self.axis = None
```
sofroniewn: is this needed?

napari/_qt/qt_viewer.py
```diff
@@ -202,8 +219,24 @@ def _update_palette(self, palette):
...
+            self.axis is not None
+            and self.view.camera.name == "TurntableCamera"
+        ):
+            self.axis.transform.reset()
+
+            self.axis.transform.rotate(self.view.camera.roll, (0, 0, 1))
+            self.axis.transform.rotate(self.view.camera.elevation, (1, 0, 0))
+            self.axis.transform.rotate(self.view.camera.azimuth, (0, 1, 0))
+
+            self.axis.transform.scale((50, 50, 0.001))
```
sofroniewn: what are these `50` numbers?

napari/_qt/qt_dims.py
```diff
@@ -251,3 +262,9 @@ def collapse_change_listener(collapsed):
         slider.collapsedChanged.connect(collapse_change_listener)
 
         return slider
+
+    def _update_display(self):
```
sofroniewn: i presume your intention here is to trim the number of sliders based on the number of displayed axes. To do this you'll want to check if the `-3` slider is present in the list of currently displayed sliders and if so call `_remove_slider(index)` with the `index`. I can help with the details, especially if we want to make sure it can be added back in

napari/_qt/qt_viewer.py
pranathivemuri: this just transforms the xyz  axis in red, green, blue color that is displayed on the right top
instead of showing at the origin i.e starting the axis from the top corner it translates it to 50, 50.
This is not causing any zoom problems. But this doesn't get deleted when the layer is deleted. This is a different kind of problem

napari/layers/volume/volume.py
pranathivemuri: I have opened a grayscale npz file before and I think it didn't change. but good catch, I will test it with simulated grayscale

napari/layers/base/base.py
pranathivemuri: removed

napari/layers/volume/volume.py
pranathivemuri: I think it is the scale with which the volume is displayed and is linked to the volume node how it should be scaled not the camera. I am removing the print statements but the problem still exists in the display

napari/_qt/qt_viewer.py
pranathivemuri: the length of x, y, z axis that is displayed in green. Turntable camera only rotates in z hence z is scaled but x, y remain same as what it was initialized with

examples/add_volume_series.py
pranathivemuri: volume node that is initialized when volume layer is called can only take 3D data. If I remove the loop it would fail. 

``` python ----------------------------------------------------------------------
  File "/usr/local/lib/python3.6/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/Users/pranathivemuri/czbiohub/napari/napari/layers/volume/tests/test_volume.py", line 59, in test_4D_volume
    layer = Volume(data)
  File "/Users/pranathivemuri/czbiohub/napari/napari/layers/volume/volume.py", line 83, in __init__
    visual = VolumeNode(volume, threshold=0.225, emulate_texture=False)
  File "/usr/local/lib/python3.6/site-packages/vispy/scene/visuals.py", line 127, in __init__
    subclass.__init__(self, *args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/vispy/visuals/volume.py", line 443, in __init__
    self.set_data(vol, clim)
  File "/usr/local/lib/python3.6/site-packages/vispy/visuals/volume.py", line 465, in set_data
    raise ValueError('Volume visual needs a 3D image.')
ValueError: Volume visual needs a 3D image.```

napari/layers/volume/volume.py
pranathivemuri: STTransform (we are actually setting STTransform in VisualWrapper class that is inherited by layers) by default doesn't do any scaling or transformation - https://github.com/vispy/vispy/blob/master/vispy/visuals/transforms/linear.py#L82  as our 2D images don't need that. But we have to find an optimal value or a formula for optimal value from the shape of 3D volume so it all fits inside the window.  hence the numbers 50, 50, 1 etc inside the volume layer

napari/_qt/qt_viewer.py
sofroniewn: ah i get it, can discuss the value of that thing with @royerloic - i always find them slightly off putting - in the long run would be nice to have an option to turn them on / off but that can come later

examples/add_volume_series.py
sofroniewn: hmm ok I can look at this, the slicing should have fixed this now

sofroniewn:
this is all great work @pranathivemuri.

Setting the zoom can be done with the camera viewbox, and I think will be a better approach than scaling and translating the layer but i'll have to look into it to be sure -  i can try taking a look later this afternoon at trying to fix the zoom issue and the 4D issue (I'll pull down your PR and then if I fix it make a commit directly back to it so all the work stays on this PR)

I think you're pretty close to having a first pass at a fully functional volume layer!!

royerloic:
Really cool stuff going on here...

I would recommend using an ArcBall camera instead of a Turntable.
Much more intuitive and easy to use -- that's what most 3D rendering
tools use...

https://github.com/vispy/vispy/blob/master/vispy/scene/cameras/arcball.py

pranathivemuri:
thanks @sofroniewn I just pushed all my changes. The display update function in qt_dims is still failing

napari/components/viewer_model.py
sofroniewn: fixed

napari/_qt/qt_dims.py
sofroniewn: fixed

sofroniewn:
@pranathivemuri I've now fixed the 4D example, the display of the dims slider, and the camera initialization. Things might not be perfect, but the volume layer seems pretty usable at this point.

Maybe the last thing to understand before merging are the `clims`, the `colormaps` and some of the other parameters on the volume layer like `threshold=0.225` which we have hard-coded but might want to make at least an input parameter.

If you can see how things look with some more real world examples that would be great.

Here's a gif of the `nD_volume.py` example

volumetric_3D

sofroniewn:
@pranathivemuri - i just merged in the updates from `0.1.2` - will make it easier to avoid merge conflicts. Let me know when you want me to checkout the colormaps / thumbnail stuff. One thing I'm thinking now is that we might need to wait till after we pass the volume through the colormap before we take the max projection, but we'll see - when I looked at things now I got some weird colormap artifacts

sofroniewn (approved):
I think this PR is now complete and ready for final review. Amazing work to get us here so quickly @pranathivemuri. I've added support now for all our old colormaps, had to add an extra line of code to deal with a vispy bug. But everything now works as I think it should.

Here's a gif of the `stent.npz` example from @pranathivemuri 

volumetric_3D_cmap

If @jni, @royerloic, or @kne42 want to give it a review - then I think we can merge

sofroniewn:
I'll also add that I've looked a tiny bit about integration between our 2D and 3D layers, and it actually looks like because of the scenegraph from vispy it will work pretty well. I think it is definitely stuff for a followup PR - but I've already got a proof of concept of overlaying a 2D slice in the 3D volume

jni:
@sofroniewn Wow, this is so amazing! I agree that proper integration can wait.

Can you investigate that Linux failure?

Also, what's your take on how hard it will be to do multicolor? Can you maybe show two 3D layers on top of each other with different colors? Will that just work out of the box now?

jni:
btw sorry, I don't have time for a proper review — at airport between planes, long trek back home now.

sofroniewn:
@jni - checkout the slack there are 2-color gifs in there, works out of the box. I'll add one example here too:

starFISH_3D

There's a 4 color example in the slack that's too big to post here.

I'll checkout the linux failure

Have a good trip back home, definitely was a productive visit!

jni:
:astonished: :heart_eyes:

sofroniewn:
All needed tests are now passing, I am going to merge

jni:
:fireworks:

PR #406, opened 2019-07-12
remove vispy backport

sofroniewn:
# Description
This PR removes our vispy backport and updates us for 0.6.0 of vispy. It currently still has one difference i don't understand - `add_labels.py` now looks like this 

image

I'll work on fixing, then maybe we can merge and make a new release on PyPi

## Type of change

# How has this been tested?

- [x] run examples

## Final checklist:


sofroniewn:
It seems like there is a difference between the rgb values we are getting from using `.map` on the colormap vs setting the colormap on the image visual, and that the correct behaviour is coming from the `.map` function

sofroniewn:
Note this is likely caused by this PR https://github.com/vispy/vispy/pull/1457 that has been merged in the vispy `0.6.0` release. Maybe @perlman has some ideas?

jni (approved):

PR #411, opened 2019-07-14
fix clim setter

sofroniewn:
# Description
This PR fixes a bug where the clim slider wasn't being updated on loading

## Type of change

# How has this been tested?

- [x] run examples

## Final checklist:

### DIFF
--- a/napari/_qt/qt_image_layer.py
+++ b/napari/_qt/qt_image_layer.py
@@ -10,7 +10,7 @@ class QtImageControls(QtLayerControls):
     def __init__(self, layer):
         super().__init__(layer)
 
-        self.layer.events.clim.connect(self.clim_slider_update)
+        self.layer.events.clim.connect(lambda e: self.clim_slider_update())
 
         # Create clim slider
         self.climSlider = QVRangeSlider(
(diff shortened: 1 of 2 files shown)


napari/_qt/qt_image_layer.py
```diff
@@ -10,7 +10,7 @@ class QtImageControls(QtLayerControls):
     def __init__(self, layer):
         super().__init__(layer)
 
-        self.layer.events.clim.connect(self.clim_slider_update)
+        self.layer.events.clim.connect(lambda e: self.clim_slider_update())
```
jni, line 13: Can you explain why this is needed?

sofroniewn: I dropped the 2nd `event` argument for `clim_slider_update` as it wasn't used, but the callback expects it. we use this pattern frequently in the code base

jni: Maybe add "event=None" to the signature instead?

sofroniewn: I prefer having the function call be simpler (i.e. no unnecessary arg). I think @kne42 has advocated for this in the past too. Thoughts @jni?

jni: @sofroniewn I prefer having code that can run without lambdas, *and* unified callback signatures, but I'll leave the discussion for later.

jni (approved):

jni:
@sofroniewn I've approved so you can merge, but I think I would like all of our callbacks to have the same signature, which allows us to write a simple guide for registering callbacks.

napari/_qt/qt_image_layer.py
sofroniewn: ok that's fine, we'll standardize in a follow up PR

Issue #376, opened 2019-06-30
Bundling napari with pyinstaller


sofroniewn:
## 🚀 Feature
With the decisions outlined in #359 and PR #356 where we are adding file I/O, menubar, and CLI to this main repo - I think it makes further sense to go ahead and add the machinery to create the `napari-app` to this repo as well, instead of making an entirely separate repo.

@csweaver has offered to help us with this process as she has been recently working on bundling a `Qt` app using `pyinstaller` for https://github.com/chanzuckerberg/cellxgene

One approach would be to add a `requirements/app-dev.txt` that contained `pyinstaller` and anything else needed to make the app. We might also want a `requirements/app.txt` for anything that is explicitly required inside the `app`, but might be optional for the script usage - but we can deal with that later.

Whenever we make releases we can then generate the binaries and upload them to the releases tab on github, so the binaries itself would never live in the repo, but would be easily available (including for different versions) in a place where it was simple to find them. I find this quite simple and appealing, and I think it makes the extra overhead and confusion of an additional repo unnecessary, though I know some projects have chosen to go down that road.

Maybe @csweaver can add some perspective on what lead the cellXgene team to take the approach they did.

Finally before we actually distribute a bundled app we need to merge #331 that switches us to depend on `PySide2`, but we need to make sure our tests are passing there first. We can get working on a PR adding `pyinstaller` functionality independently of that process too though - I think it we be great to have this for our `0.1` release if possible and I know @royerloic agrees!!

sofroniewn:
With #1289 we decided to use briefcase for our bundling and so this issue can be closed

PR #412, opened 2019-07-14
switch to pyside2

sofroniewn:
# Description
This PR supersedes #331  now that vispy `0.6` has been released by allowing us to use either `PySide2` or `PyQt5`. Right now we follow the conventions of [`qtpt`](https://github.com/spyder-ide/qtpy#requirements) such that you need one of `PySide2` or `PyQt5` installed. If both are installed `PyQt5` will be used first unless you set the `QT_API` environment variable to `PySide2` - which is worth noting for our existing developers that already have `PyQt5` installed.

I have dropped `PyQt5` and `SIP` from our requirements though and replaced it with `PySide2`. @csweaver - once this merges we can press full steam ahead with creating our bundled app as discussed in #376.

## Type of change

# How has this been tested?

- [x] run examples

## Final checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
 [![Development Status](https://img.shields.io/pypi/status/napari.svg)](https://github.com/napari/napari)
 [![Code style: black](https://img.shields.io/badge/code%20style-black-000000.svg)](https://github.com/python/black)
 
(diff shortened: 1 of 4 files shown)


jni (approved):
Very nice!

csweaver:
👍 Don't forget to change the project description too
Screen Shot 2019-07-13 at 10 21 17 PM

I'll look at #376 at the conference

Issue #414, opened 2019-07-14
Paint color doesn't match color indicator after shuffle color button is pressed


marshuang80:
## 🐛 Bug

When the "shuffle colors" button is pressed, the paint brush's color is changed as well. However, the paint brush color indicator on the left still displays the previous color. 

## To Reproduce

1. Make a label layer
2. Click shuffle colors
3. Paint on layer

![image](https://user-images.githubusercontent.com/15271135/61186046-6aedbc80-a615-11e9-8f15-e380a59f075d.png)

## Expected behavior

The paint color indicator should match the actual paint color. 

## Environment

 - napari 1.0
 - Linux 18.04 LTS
 - Python version: 3.7

sofroniewn:
This will hopefully be fixed after fixing our vispy `0.6` update - as discussed here #407

sofroniewn:
fixed by #416

PR #415, opened 2019-07-14
[FIX] paint color inidicator update when shuffle color

marshuang80:
# Description
This PR fixes the issue as described in #414: after the "shuffle color" button is pressed, the paint brush color is changed but the color indicator still displays the original color. The issue is fixed by updating selected_label when new_colormap() function is called.

## Type of change

# References
close issue #414

# How has this been tested?
Tested on local computer. Checked that the color indicator is changed when "shuffle color" is pressed.

## Final checklist:
- [x ] My PR is the minimum possible work for the desired functionality


marshuang80:
Updated on wrong version

PR #416, opened 2019-07-14
[FIX] paint color inidicator update when shuffle color

marshuang80:
# Description
This PR fixes the issue as described in #414: after the "shuffle color" button is pressed, the paint brush color is changed but the color indicator still displays the original color. The issue is fixed by updating selected_label when new_colormap() function is called.

## Type of change

# References
close issue #414

# How has this been tested?
Tested on local computer. Checked that the color indicator is changed when "shuffle color" is pressed.

## Final checklist:
- [x ] My PR is the minimum possible work for the desired functionality

### DIFF
--- a/napari/layers/labels/labels.py
+++ b/napari/layers/labels/labels.py
@@ -322,6 +322,7 @@ class Labels(Layer):
 
     def new_colormap(self):
         self._seed = np.random.rand()
+        self.selected_label = self._selected_label
         self.refresh()
 
     def get_color(self, label):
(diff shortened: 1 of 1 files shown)


sofroniewn (changes requested):
can you instead call
`self._selected_color = self.get_color(self.selected_label)`
it should also fix and is more transparent.
also can you delete
```python
        if selected_label == 0:
            # If background
            self._selected_color = None
        else:
```
in the selected_label setter as they aren't need as it is already done inside `get_color`
good catch

sofroniewn (approved):
Looks good to me

PR #421, opened 2019-07-15
[WIP] Python code generation and execution in IPython from the GUI

jni:
# Description

When this is finished, this will provide an embedded IPython console running on the same Python process as napari. The only currently implemented feature is to capture the IPython shell that napari was launched from, add text to it, and run that text. Todo:

- [ ] a context manager that captures any *current* text that the user has typed, clears it, adds ours, runs ours, then restores the contents. This way if the user types something then uses a napari UI function, their work isn't unceremoniously jettisoned.
- [ ] the napari.viewer code should *not* be added to the history in general as it is in the current example. The goal is to save any *processing* steps. We can see whether to add an option to the recorder that also records viewer state, but imho that should not be the default.
- [ ] the code will be different depending on whether we are running in the notebook, or in a Python process. I'll experiment with those once @sofroniewn gets me an embedded terminal in the GUI.

@sofroniewn the GlueViz code for a Qt embedded terminal is here:

https://github.com/glue-viz/glue/blob/master/glue/app/qt/terminal.py

## Type of change

# References

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/components/console.py
+++ b/napari/components/console.py
@@ -1,0 +1,13 @@
+from IPython.core.interactiveshell import InteractiveShell
+
+
+def call_ipython_code(code='print("hello world")'):
+    # get current running instance or create new instance
+    shell = InteractiveShell.instance()
+    # note: shell.extract_input_lines will return existing input lines.
+    # we should save them and restore them after running our command
+    # shell.set_next_input
+    shell.set_next_input(code)
+    shell.history_manager.store_inputs(shell.execution_count + 1, code)
+    shell.ex(code)
+    shell.execution_count += 1
--- a/napari/components/console.py
+++ b/napari/components/console.py
@@ -1,7 +1,13 @@
 from IPython.core.interactiveshell import InteractiveShell
 
 
-def call_ipython_code(code='print("hello world")'):
+CODE = '''from skimage import filters
+output = filters.gaussian(viewer.layers[0].data, sigma=10)
+viewer.add_image(output)
+'''
+
+
+def call_ipython_code(code=CODE):
     # get current running instance or create new instance
     shell = InteractiveShell.instance()
     # note: shell.extract_input_lines will return existing input lines.
--- a/napari/components/console.py
+++ b/napari/components/console.py
@@ -7,7 +7,6 @@ def call_ipython_code(code='print("hello world")'):
     # note: shell.extract_input_lines will return existing input lines.
     # we should save them and restore them after running our command
     # shell.set_next_input
-    shell.set_next_input(code)
-    shell.history_manager.store_inputs(shell.execution_count + 1, code)
-    shell.ex(code)
-    shell.execution_count += 1
+    # shell.set_next_input(code)
+    shell.pt_app.app.current_buffer.insert_text(code)
+    shell.pt_app.app.current_buffer.validate_and_handle()
--- a/napari/components/console.py
+++ b/napari/components/console.py
@@ -1,4 +1,4 @@
-from IPython.core.interactiveshell import InteractiveShell
+from IPython import get_ipython
 
 
 CODE = '''from skimage import filters
@@ -9,10 +9,16 @@ viewer.add_image(output)
 
 def call_ipython_code(code=CODE):
     # get current running instance or create new instance
-    shell = InteractiveShell.instance()
+    shell = get_ipython()
     # note: shell.extract_input_lines will return existing input lines.
     # we should save them and restore them after running our command
-    # shell.set_next_input
-    # shell.set_next_input(code)
-    shell.pt_app.app.current_buffer.insert_text(code)
-    shell.pt_app.app.current_buffer.validate_and_handle()
+    if shell is None:  # pure-python interpreter
+        pass
+    elif hasattr(shell, 'pt_app'):  # ipython terminal
+        shell.pt_app.app.current_buffer.insert_text(code)
+        shell.pt_app.app.current_buffer.validate_and_handle()
+    else:  # ipython kernel
+        shell.kernel._publish_execute_input(
+            code, None, shell.kernel.execution_count
+        )
+        shell.run_cell(code)
--- a/napari/_qt/qt_layer_buttons.py
+++ b/napari/_qt/qt_layer_buttons.py
@@ -1,5 +1,7 @@
 from qtpy.QtWidgets import QHBoxLayout, QPushButton, QFrame
 
+from ..components.console import call_ipython_code
+
 
 class QtLayersButtons(QFrame):
     def __init__(self, viewer):
@@ -10,6 +12,7 @@ class QtLayersButtons(QFrame):
         self.newPointsButton = QtNewPointsButton(self.viewer)
         self.newShapesButton = QtNewShapesButton(self.viewer)
         self.newLabelsButton = QtNewLabelsButton(self.viewer)
+        self.consoleButton = QtConsoleButton(self.viewer)
 
         layout = QHBoxLayout()
         layout.addStretch(0)
@@ -18,6 +21,7 @@ class QtLayersButtons(QFrame):
         layout.addWidget(self.newShapesButton)
         layout.addWidget(self.newLabelsButton)
         layout.addWidget(self.deleteButton)
+        layout.addWidget(self.consoleButton)
         self.setLayout(layout)
 
 
@@ -83,3 +87,14 @@ class QtNewLabelsButton(QPushButton):
         self.setFixedHeight(28)
         self.setToolTip('New labels layer')
         self.clicked.connect(lambda: self.viewer._new_labels())
+
+
+class QtConsoleButton(QPushButton):
+    def __init__(self, viewer):
+        super().__init__()
+
+        self.viewer = viewer
+        self.setFixedWidth(28)
+        self.setFixedHeight(28)
+        self.setToolTip('Open IPython terminal')
+        self.clicked.connect(lambda: call_ipython_code())
--- a/napari/resources/stylesheet.qss
+++ b/napari/resources/stylesheet.qss
@@ -348,6 +348,10 @@ QtNewLabelsButton {
    image: url(":/icons/{{ folder }}/new_labels.svg");
 }
 
+QtConsoleButton {
+  image: url(":/icons/{{ folder }}/properties_expand.svg");
+}
+
 /* ------------------------------------------------------ */
 
 QtModeButton {
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -1,3 +1,4 @@
+IPython>=7.6
 imageio>=2.5.0
 numpy>=1.10.0
 qtpy>=1.7.0


jni:
I also think the terminal drawer icon should resemble `>_` rather than just `>`. Active/inactive can be indicated by "lighting up" the button, adding shadow, etc.

jni:
Note: my takeaway from discussions with @Carreau and various browsings:

It's impossible or at least heavily discouraged to know whether we are running in notebook, jupyter lab, jupyter console, etc — the kernel should not know what kind of front-end is connected, by design. See [this SO answer](https://stackoverflow.com/a/15412661/224254) by @Carreau.

Instead, we have by my count three options:

  1. We are running on an IPython kernel. If so, connect our IPython terminal to that kernel and run our code on the kernel. As far as I can tell, in this situation we can't actually edit text on any console except our built in one, but the history etc. will just work. I think `ip = IPython.get_ipython()` will be sufficient to detect this here, and will be `None` if there is no IPython.
  2. We are running on IPython terminal. If so, use the `pt_app` machinery as in the earliest version of this PR. (@Carreau: can we reliably distinguish between options 1 and 2?)
  3. We are running in pure Python. If so, launch an IPython in-process console (?) as Glue does. *I think* that we can automatically have access to all the variables in the Python global namespace in our script.

jni:
Just to clarify: we need to detect and support all three options.

napari/components/console.py
```diff
@@ -0,0 +1,18 @@
...
+
+CODE = '''from skimage import filters
+output = filters.gaussian(viewer.layers[0].data, sigma=10)
+viewer.add_image(output)
+'''
+
+
+def call_ipython_code(code=CODE):
+    # get current running instance or create new instance
+    shell = InteractiveShell.instance()
```
Carreau: just FYI, the type of `shell` here, will depends on wether you are in a kernel of terminal IPython.
You could use this – or hasattr – to know what to do.

Carreau:
```python
from IPython import get_ipython


CODE = '''from skimage import filters
output = filters.gaussian(viewer.layers[0].data, sigma=10)
viewer.add_image(output)
'''


def call_ipython_code(code=CODE):
    # get current running instance or create new instance
    shell = get_ipython()
    # note: shell.extract_input_lines will return existing input lines.
    # we should save them and restore them after running our command
    # shell.set_next_input
    # shell.set_next_input(code)
    if shell is None:
        pass #pure python
    if hasattr(shell, 'pt_app'):
        shell.pt_app.app.current_buffer.insert_text(code)
        shell.pt_app.app.current_buffer.validate_and_handle()
    else:
        # likely terminal, we could check the type of shell, and it would by
        # ipykernel.interactiveshell.ZMQInteractiveShell
        shell.run_cell(code)
        # should run code and add to history. 

        # I don't see any easy way to populate the _interface_ with the code
        # in other kernel-based clients.
```

Carreau:
add a ` shell.kernel._publish_execute_input(code, None, shell.kernel.execution_count)` above `run_cell`, is ugly but seem to work on frontend that support it.

Carreau:
The "JupyterLab console" have a "Show all kernel activity" option, which when enabled shows code executed from other clients.

You can see below that after clicking @jni button the code appear in thee console (result is above, but that's fixable), but it will not show up in the notebook. 
Screen Shot 2019-07-15 at 3 21 45 PM

jni:
Wow, thanks @Carreau! This is super great! 🎉 

I'm a bit nervous using a function with a leading underscore. Should we send a PR to IPython kernel (is that ipython/ipython?) to propose making it public?

Carreau:
> I'm a bit nervous using a function with a leading underscore. Should we send a PR to IPython kernel (is that ipython/ipython?) to propose making it public?

I would suggest you play a bit with it, see how it goes, and if making it public is enough then we can do it.

 But we can also see how your usecase goes and provide a better api for embedded app like napari.
There are definitively things that I believe current code will not handle; in particular errors, and cleanup some internal logic/state once the errors are caught.


While I hav you, genuine "first time user" feedback, and I know this is still relatively at prototype stage;
Right now clicking apply a gaussian filter and create a new image. Is that the end goal; or do you plan on attaching a sequence of filter to the given image that you can modify after the fact ?

jni:
> While I have you, genuine "first time user" feedback, and I know this is still relatively at prototype stage;
> Right now clicking apply a gaussian filter and create a new image. Is that the end goal; or do you plan on attaching a sequence of filter to the given image that you can modify after the fact ?

@Carreau it makes me very happy that you are trying this out! I'm not sure I understand the question, but I'll try to express the vision in multiple stages.

1. First, napari is a viewer and annotator tool for multidimensional arrays. In this capacity, almost everyone on the current team is using it quasi-daily. It's so much easier to rapidly scroll through a bunch of images, or overlay an image and a segmentation, or somesuch, than it has been before.
2. In the next phase, we want napari to make the image analysis tools of the SciPy ecosystem available to biologists, and other researchers who don't (yet) know how to code. In this scenario, users would navigate the menus: plugins > scikit-image > filters > gaussian, or plugins > numpy > log10, and they would get some way of setting the parameters, and out would come the output. In this part, two things are important: (1) yes, it creates a new image and overlays it on the original image. Thus the editing is non-destructive and you can immediately see the result. And (2), the code that is run when clicking on this stuff would be proper Python (as in the example). Thus users could quickly grab it later and create their own functions which are combinations of other functions, or stick them in a for-loop and batch-process a whole bunch of images.

The embedded IPython terminal / IPython integration is our back-door to get people to code after using simple commands. Hopefully, after a few clicks, users will just be typing stuff into IPython.

I'm not sure whether you are asking whether we should modify the image in-place (no), or whether we should define a computational pipeline that isn't actually run until everything is complete, avoiding temporary intermediate steps. For this second bit, at least in the very short term, the answer is also no, but if you want this feature, let us know. =)

I hope this helps!

Carreau:
> I'm not sure whether you are asking whether we should modify the image in-place (no)

Yeah, that's not what I meant. 

> I'm not sure whether you are asking whether we should modify the image in-place (no), or whether we should define a computational pipeline that isn't actually run until everything is complete, avoiding temporary intermediate steps. For this second bit, at least in the very short term, the answer is also no, but if you want this feature, let us know. =)

I was thinking about two things. 
First:
  - let's say I gaussian blur of radius 10,  binarize,  then skeletonize (using skan :-) ).
  - Can I change the blur radius w/o rerunning the steps manually ?
Second
   - Know ing first, in the UI; I'd love to have a difference between "source image" and "derived image"

Apologies for the crude screenshot, but something where "derived" imaged, are for example indented wrt the origin source..
ex

Anyway when I deal with images, I'll be happy to try it out, break it and send you feedback :-)

sofroniewn:
@Carreau I think a long term goal of the plugin integration would be to allow things like both (1) and (2). i.e. you'd add a plugin that took in some layers and some parameters and outputted some layers, and you could either run it just once with one set of parameter values, or have it continuously update as you adjust parameters that have been added to the gui as interactive elements. We might need to have something more like a "preview" mode for the continuous updates where some downsampling / acceleration strategy to ensure smooth interactivity before you click "apply", but we'll see.

How you'd select input layers / denote output layers, and where we'd put the interactive parameter elements are all things that we'll need to figure out.

So glad you're excited about trying it out, and thanks for helping @jni here with the console integration. I think having a console will be a huge value add for the tool!

jni:
> or whether we should define a computational pipeline that isn't actually run until everything is complete, avoiding temporary intermediate steps. For this second bit, at least in the very short term, the answer is also no, but if you want this feature, let us know. =)

Thinking more about this, it might actually work out of the box on dask arrays, because we only instantiate the 2D view of the array when we display it. The rest of the array is not computed. So, with appropriate chunking, we would apply a transformation, and see the result of that transformation in that slice, but the rest of the array would not be created. We could apply a bunch of steps in sequence this way. Then we could call "save to zarr", and have the computation happen disk-to-disk.

jni:
Oops, GitHub had not refreshed to show me your earlier responses! :man_facepalming: 

Your suggestion is very magical, @Carreau. =) Part of the reason I want things to go in the IPython history is so that users could quickly promote multiple steps to a single function that does all the things they just did, and apply it in one step. But that's not exactly the same thing as you're asking.

Keeping track of provenance through all the steps, and displaying it intelligently is even more long term as far as I'm concerned. Many functions might require multiple inputs (e.g. an image, a segmentation, and some coordinates) — how do you indent things then? I think you need KNIME or some other computational graph drawing for this, and we are not trying to be KNIME. (For now anyway.)

@sofroniewn perhaps we can start an issue, "features for 2.0", and add @Carreau's comments to it? =P

Carreau:
> Keeping track of provenance through all the steps, and displaying it intelligently is even more long term as far as I'm concerned. Many functions might require multiple inputs (e.g. an image, a segmentation, and some coordinates) — how do you indent things then?

I know tough questions :-) Ithink that Photoshop (and GIMP) have really good UI in that sens, but do not go far enough in their usage of it; likely because of backward compatibility.You've probaly seen it; the "layer" view is a tree, which can be seen as a computation graph (IMO):

```
└── f(h,k)
    ├── h(a,b)
    │   ├── a
    │   └── b
    └── k(c,d)
        ├── c
        └── d
```
Except the "function applications" you can do are limited, and it's more kinda `k(c(d))` than `k(c,d)`

If ever you need to go from a tree to a DAG you can "collapse" part of the tree, and make it an object you can reference in multiple locations.

But let's not plan 2.0 too early :-) 

+1 for something like dask.

sofroniewn:
@jni so I started #442 as a largely complementary effort to explore integration of the actual Qt widget for the console, but during the process I have come to consider another approach and I'm curious about your opinion of it.

We could create an entirely separate IPython process inside the console and pass it a single variable `viewer` which was the `ViewerModel` and have it be independent of any variables defined in the python script, or the a jupyter notebook, or the `IPython` console where you launched napari. The nice thing about this approach is that the user only ever has the `viewer` variable predefined, and so we can document things really nicely. I'm not quite sure what the use case looks like for people who want to go back and forth between the jupyter notebook and this console, or if people launching from a script need all the other imports they have done.

If this decision dramatically simplifies things I think we should consider it. Thoughts @kne42?

jni:
I'm very very 👎 on this. Just like we worry about the consistency of launching napari from the command line vs the app vs IPython/notebook, we should worry about the consistency of IPython on napari vs existing instances.

Of course the idea of not being able to launch napari from an existing IPython console is a non-starter. But if you manage to fix that, then the fact that IPython in napari is completely independent of the IPython that launched napari is, to me, just terrible.

At an absolute minimum, the IPython terminal embedded in napari should mirror the global environment in which napari was launched. It's (to me) insane that a user might have all kinds of stuff defined in Python but they can't access them. If you have seven images in the namespace you can't even add them to the viewer with this scheme, unless I misunderstand? Just manipulating the viewer is insufficient reason to embed IPython, imho. If that's our goal then we should just scrap the whole approach.

I've mentioned my long term goals above — using napari as a backdoor to teach Python is a place where we could have massive impact, and imho the main reason why we would want to embed. Fiddling with the viewer is very low value in comparison.

Anyway, this seems like the perfect thing to discuss at the meeting tomorrow. =)

sofroniewn:
@jni - i hear you about the value of having all the global environment variables available in the console and shared state with the iPython Console / Jupyter Notebook - but I'm not convinced it's a complete deal breaker not to have it - for people launching from the CLI / app there is no difference. I also do think there is simplicity in having the console independent with just the `viewer` variable.

I also wouldn't underestimate the value of having just the `viewer` - and I don't see the disconnect between having access to the `viewer` and beginners learning some python. The viewer is the main object they will want to interact with, and I don't see why a beginner would be going back and forth between the embedded console and their main iPython notebook / Jupyter notebook.

Looking forward to discussing more

sofroniewn:
Also these comments on how glue does things might be informative too - https://github.com/glue-viz/glue/issues/1209 in terms of expectations on shared namespaces

similarly some comments shared namespaces with iPython that might now be out of date - https://github.com/ipython/ipykernel/issues/228

sofroniewn:
@jni with the merging of the qt_console in #443 what should we do with this PR? My impression is that this PR had different functionality, allowing us to inject code we executed into the ipython console which is relevant both for launching from ipython and from using our embedded console I think? But the amount of duplication / integration between the two is unclear, should we leave this PR open or close to start a new one when we're more ready for this aspect of things?

jni:
Let's leave it open. When I get back from holidays / teaching summer school, I'll look into rebasing and then using it as a basis for plugin interaction.

tlambert03:
@jni, still want to keep this PR open?

jni:
@tlambert03 Ha, a blast from the past. =) Erm, sort of? It's got some code that doesn't exist anywhere else, relating to code generation and the idea of having napari buttons show code execution in IPython. Perhaps the title needs to be updated. I'll give that a go.

sofroniewn:
@jni can i give this a bump. Should we close it or maybe get @Carreau to take it over at some point (not sure though if we should open a new issue for it first and then close this)?

Or is it something we should push to https://github.com/napari/napari-console?

Just trying to do some housekeeping!!

sofroniewn:
ToDo - convert to issue, and preserve any code that we might want to reuse - cc @jni

jni:
Closed as outdated — discussion continues in #3526.

Issue #308, opened 2019-05-29
Int images are not viewable


jni:
## 🐛 Bug

If an image has dtype `int` (int64 on most platforms), napari crashes.

## To Reproduce

Run the following code:

```python
import numpy as np
from skimage import data
import napari
from napari.util import app_context

with app_context():
    blobs = data.binary_blobs(length=128, volume_fraction=0.1, n_dim=3)
    v = napari.view(blobs=blobs)
```

Error:

```pytb
Traceback (most recent call last):
  File "image-int.py", line 8, in <module>
    v = napari.view(blobs=blobs)
  File "/home/jni/projects/napari/napari/_view.py", line 50, in view
    clim_range=clim_range, name=name)
  File "<string>", line 3, in add_image
  File "/home/jni/projects/napari/napari/components/_viewer/model.py", line 287, in add_layer
    layer.indices = self.dims.indices
  File "/home/jni/projects/napari/napari/layers/_base_layer/model.py", line 114, in indices
    self._set_view_slice()
  File "/home/jni/projects/napari/napari/layers/_image_layer/model.py", line 209, in _set_view_slice
    self.status = self.get_message(coord, value)
  File "/home/jni/projects/napari/napari/layers/_image_layer/model.py", line 378, in get_message
    msg = msg + f'{value:0.3}'
ValueError: Precision not allowed in integer format specifier
```

## Expected behavior

The image is displayed correctly.

## Environment

 - napari Version (e.g., 1.0): master
 - OS (e.g., Linux): Linux
 - Python version: 3.6
 - Any other relevant information:

sofroniewn:
:-( looks like another problem with our string formatting, as mentioned in #93 and has been cropping up elsewhere.

@jni do you want to take a stab at fixing this across the entire codebase. It's pretty important we get it right and I think you have a good sense for the desired string outputs and the best ways to get them!!

jni:
Yes, I'll work on this, just wanted to add it here for tracking. Another 1-2 issues incoming. =\

sofroniewn:
this is now fixed by #422

PR #422, opened 2019-07-15
Fix Crash with stacked binary tiffs.

Carreau:
Apparently the dtypes can also be booleans.

# Description
Some tiff in the wild appear to be binay (skimage.io.read return an nd-array of booleans), 
I unfortunately was not able to generate a boolean-tiff, and can't share my current sample publicly.

## Type of change

# How has this been tested?

`napari <file>` does not crash with the following anymore

```
  File "/Users/mbussonnier/dev/napari/napari/_qt/qt_viewer.py", line 146, in _open_images
    self._add_files(filenames)
  File "/Users/mbussonnier/dev/napari/napari/_qt/qt_viewer.py", line 165, in _add_files
    image, multichannel=is_multichannel(image.shape)
  File "/Users/mbussonnier/dev/napari/napari/components/viewer_model.py", line 360, in add_image
    layer = layers.Image(image, *args, **kwargs)
  File "/Users/mbussonnier/dev/napari/napari/layers/image/image.py", line 155, in __init__
    self._set_view_slice()
  File "/Users/mbussonnier/dev/napari/napari/layers/image/image.py", line 275, in _set_view_slice
    self.status = self.get_message(coord, value)
  File "/Users/mbussonnier/dev/napari/napari/layers/image/image.py", line 378, in get_message
    msg = msg + f'{value:0.3}'
ValueError: Precision not allowed in integer format specifier
```

## Final checklist:

-- 

I'm not entirely sure this is the right fix... should the image maybe always be casts as integers even if boolean ?

### DIFF
--- a/napari/layers/image/image.py
+++ b/napari/layers/image/image.py
@@ -372,7 +372,7 @@ class Image(Layer):
                 v_str = '[' + str.join(', ', [f'{v:0.3}' for v in value]) + ']'
                 msg = msg + v_str
         else:
-            if isinstance(value, np.integer):
+            if isinstance(value, (np.integer, np.bool_)):
                 msg = msg + str(value)
             else:
                 msg = msg + f'{value:0.3}'
--- a/napari/layers/image/tests/test_image.py
+++ b/napari/layers/image/tests/test_image.py
@@ -43,6 +43,18 @@ def test_integer_image():
     assert layer._data_view.shape == shape[-2:]
 
 
+def test_bool_image():
+    """Test instantiating Image layer with bool data."""
+    shape = (10, 15)
+    data = np.zeros(shape, dtype=bool)
+    layer = Image(data)
+    assert np.all(layer.data == data)
+    assert layer.ndim == len(shape)
+    assert layer.shape == shape
+    assert layer.multichannel == False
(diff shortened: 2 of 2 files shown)


sofroniewn (changes requested):
I think this fix is fine for now - I can confirm it fixes #308 too. We'll be refactoring our string printing pretty soon as discussed in #93.

Can you add a simple test to `napari/layers/image/tests/test_image.py`.

Something like
```python
def test_bool_image():
    """Test instantiating Image layer with bool data."""
    shape = (10, 15)
    data = np.zeros(shape, dtype=bool)
    layer = Image(data)
    assert np.all(layer.data == data)
    assert layer.ndim == len(shape)
    assert layer.shape == shape
    assert layer.multichannel == False
    assert layer._data_view.shape == shape[-2:]
```
which should now pass but would have failed before. Then I'll merge. Thanks for jumping in and fixing!

jni:
@Carreau thanks for the fix + stress test!

@sofroniewn I've added the requested test.

sofroniewn (approved):
Looks good to me, thanks for jumping in @jni

Carreau:
Thanks !

PR #423, opened 2019-07-16
cleanup shape classes

sofroniewn:
# Description
This PR consolidates the dictionary of shape classes in the `_constants.py` file. It is unclear to me if I should also be using Enums as described in #265. @kevinyamauchi @kne42 do you have thoughts?

## Type of change

# How has this been tested?

- [x] run tests

## Final checklist:

### DIFF
--- a/napari/layers/shapes/shape_list.py
+++ b/napari/layers/shapes/shape_list.py
@@ -2,7 +2,7 @@ import numpy as np
 from .shape_models import Shape, Line, Path
 from .shape_util import inside_triangles, triangles_intersect_box
 from .mesh import Mesh
-from ._constants import shape_classes
+from ._constants import shape_classes, ShapeType
 
 
 class ShapeList:
@@ -313,13 +313,12 @@ class ShapeList:
         if new_type is not None:
             cur_shape = self.shapes[index]
             if type(new_type) == str:
-                if new_type in shape_classes.keys():
-                    shape_cls = shape_classes[new_type]
+                shape_type = ShapeType(new_type)
+                if shape_type in shape_classes.keys():
+                    shape_cls = shape_classes[shape_type]
                 else:
                     raise ValueError(
-                        """shape_type not recognized. Must be one of
-                                 "{'line', 'rectangle', 'ellipse', 'path',
-                                 'polygon'}"."""
+                        f'{shape_type} must be one of {set(shape_classes)}'
                     )
             else:
                 shape_cls = new_type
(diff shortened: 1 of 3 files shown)


napari/layers/shapes/shape_list.py
```diff
@@ -681,15 +672,15 @@ def to_masks(
                 s.to_mask(mask_shape, zoom_factor=zoom_factor, offset=offset)
                 for s in self.shapes
             ]
-        elif shape_type not in self._types.keys():
+        elif shape_type not in shape_classes.keys():
             raise ValueError(
                 """shape_type not recognized, must be one of
                          "{'line', 'rectangle', 'ellipse', 'path',
```
kne42: this shouldn't be hardcoded

napari/layers/shapes/shapes.py
```diff
@@ -202,13 +202,6 @@ class Shapes(Layer):
     _rotation_handle_length = 20
     _highlight_color = (0, 0.6, 1)
     _highlight_width = 1.5
-    _types = {
```
kne42, line 205: tbh I did kind of like having the types of shapes attached as a class variable but it's fine to remove

kne42 (approved):
yes, I think having enums would also be good here but they don't work as a replacement to the shape dictionary, rather, it would be changed to something like this:
```python
shape_classes = {
    Shape.Rectangle: Rectangle,
    Shape.Ellipse: Ellipse,
    ...
}
```

it can come as a later PR

napari/layers/shapes/shape_list.py
sofroniewn: you mean the error message shouldn't have the types in them? removed

kne42: no, I mean you should include it as something like ```f'must be one of {set(shape_classes)}```

on another note I just realized that this is a literal string. this should also be changed since literal strings should almost never be used in-line as they will capture indentation which is undesirable.

sofroniewn: fixed

sofroniewn:
I've switched over to using an enum instead of string for the dictionary and have switch to using f-strings for the error print statements. will merge once tests pass unless there are any problems

napari/layers/shapes/shape_list.py
```diff
@@ -676,20 +666,20 @@ def to_masks(
         if mask_shape is None:
             mask_shape = self._vertices.max(axis=0).astype('int')
 
+        if type(shape_type) == str:
```
kne42, line 669: use `isinstance instead`

Issue #407, opened 2019-07-12
vispy 0.6 release


sofroniewn:
## 🐛 Bug

We currently have a bug with the new `0.6.0` vispy release - see errors in `add_image.py`

![image](https://user-images.githubusercontent.com/6531703/61162374-e5abc000-a4cd-11e9-90ad-ce1b111d3c1c.png)

and `add_labels.py`

![image](https://user-images.githubusercontent.com/6531703/61162400-ff4d0780-a4cd-11e9-95c7-65ab735d5206.png)

The first bug is already fixed in #406 by removing our vispy backport. we need to still fix the second one

sofroniewn:
The above colormap issue can be fixed by reverting this function back to its `0.5.6` version in the `vispy.color.colormaps.py` file

```python
def _glsl_step(controls=None):
    assert (controls[0], controls[-1]) == (0., 1.)
    ncolors = len(controls) - 1
    assert ncolors >= 2
    s = ""
    for i in range(ncolors-1):
        if i == 0:
            ifs = 'if (t < %.6f)' % (controls[i+1])
        elif i == (ncolors-2):
            ifs = 'else'
        else:
            ifs = 'else if (t < %.6f)' % (controls[i+1])
        s += """%s {\n    return $color_%d;\n} """ % (ifs, i)
    return """vec4 colormap(float t) {\n%s\n}""" % s
```
i'll look into a better fix, though it would be easy to back port this if needed @jni @kne42 @perlman

sofroniewn:
Something to try for @perlman is reversing the colors list inside the `_glsl_step` function
```python
def _glsl_step(controls=None, colors=None, texture_map_data=None):
    assert (controls[0], controls[-1]) == (0., 1.)
    ncolors = len(controls) - 1
    assert ncolors >= 2
    assert (texture_map_data is not None)
    # Reverse color
    # print(colors[::-1])
    colors = colors[::-1]
```
I'll check it out when I get the chance, for now I've merged our bug fix #426

perlman:
I think this is a bug in vispy. The colors in the generated LUT appear to be in inverse order from how they are specified.

I can add
>  colors = colors[::-1]

to _glsl_step (https://github.com/vispy/vispy/blob/master/vispy/color/colormap.py#L168) to bring back the prior behavior.

(To be clear, this isn't a proper fix. I think there is a logic flaw with how it's assigning the LUT bucket. e.g., one can invert LUT_tex_idx as well.)

[...]

PR #426, opened 2019-07-17
fix vispy 0.6.0 colormap bug

sofroniewn:
# Description
This PR reverts the colormaps we use to be the `0.5.6` code for the `_glsl_step` function as described in #407 and #406. Everything else is identical all functionality is preserved. this is not a long term solution but at least now brings functionality back to the labels layer 

## Type of change

# How has this been tested?

- [x] `add_labels.py` and `add_image.py` including changing colormaps and clims

## Final checklist:

### DIFF
--- a/napari/_vispy/color/__init__.py
+++ b/napari/_vispy/color/__init__.py
@@ -1,0 +1,1 @@
+from .colormap import Colormap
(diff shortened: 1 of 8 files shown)


kne42 (approved):
you mean `0.5.3` right? regardless lgtm

perlman:
I think this is a bug in vispy. The colors in the generated LUT appear to be in inverse order from how they are specified.

I can add
>  colors = colors[::-1]
to _glsl_step (https://github.com/vispy/vispy/blob/master/vispy/color/colormap.py#L168) to bring back the prior behavior.

PR #434, opened 2019-07-20
Expose args for blending, visible, opacity

sofroniewn:
# Description
This PR performs a little refactoring around the visual wrapper for the layers, including exposing `opacity`, `blending`, and `visible` as three new keyword arguments for every layer. I found myself often wanting to pass these directly to the `add_image`, `add_points` etc. functions and now you can. I've also added doc strings and tests for all these properties. No functionality should change.

## Type of change

# How has this been tested?

- [x] add new tests to each layer model, and add `examples/pass_colormaps.py`

## Final checklist:


napari/layers/base/base.py
```diff
@@ -20,6 +20,85 @@ class Layer(VisualWrapper, KeymapMixin, ABC):
...
         from `cls._basename()`
+    opacity : float
+        Opacity of the layer visual, between 0.0 and 1.0.
+    blending : str
+        One of a list of preset blending modes that determines how RGB and
+        alpha values of the layer visual get mixed. Allowed values are
+        {'opaque', 'translucent', and 'additive'}.
+    visible : bool
+        Whether the layer visual is currently being displayed.
+
```
kne42, line 31: you should only have 1 newline here

napari/layers/base/base.py
```diff
@@ -20,6 +20,85 @@ class Layer(VisualWrapper, KeymapMixin, ABC):
...
+        Determine if canvas pan/zoom interactivity is enabled.
+    cursor : str
+        String identifying which cursor displayed over canvas.
+    cursor_size : int | None
+        Size of cursor if custom. None yields default size
+    scale_factor : float
+        Conversion factor from canvas coordinates to image coordinates, which
+        depends on the current zoom level.
+
+    Extended Summary
```
kne42, line 93: there is no section in the numpy doc format for an "extended summary." see "notes" instead

napari/layers/base/base.py
```diff
@@ -20,6 +20,85 @@ class Layer(VisualWrapper, KeymapMixin, ABC):
...
+    ----------
```
kne42: "underline" should match length of corresponding header

napari/layers/base/base.py
```diff
@@ -20,6 +20,85 @@ class Layer(VisualWrapper, KeymapMixin, ABC):
...
+    _master_transform : vispy.visuals.transforms.STTransform
```
kne42, line 95: private attributes/methods should not be included in the docstring

napari/layers/image/image.py
```diff
@@ -105,12 +113,21 @@ def __init__(
         clim=None,
         clim_range=None,
         interpolation='nearest',
+        opacity=1,
```
kne42, line 116: could we make these keyword-only arguments by adding the `*` as a separate argument?

napari/layers/labels/labels.py
```diff
@@ -100,12 +106,20 @@ def __init__(
         num_colors=50,
         seed=0.5,
         opacity=0.7,
```
kne42, line 108: see above

napari/layers/points/points.py
```diff
@@ -113,6 +121,9 @@ def __init__(
         edge_color='black',
         face_color='white',
         n_dimensional=False,
+        opacity=1,
```
kne42, line 124: see above

napari/layers/shapes/shapes.py
```diff
@@ -213,8 +219,10 @@ def __init__(
         edge_width=1,
         edge_color='black',
         face_color='white',
-        opacity=0.7,
         z_index=0,
+        opacity=0.7,
```
kne42, line 223: see above

napari/layers/vectors/vectors.py
```diff
@@ -65,10 +73,25 @@ class Vectors(Layer):
...
 
     def __init__(
-        self, vectors, *, edge_width=1, edge_color='red', length=1, name=None
+        self,
+        vectors,
+        *,
+        edge_width=1,
+        edge_color='red',
+        length=1,
+        opacity=1,
```
kne42, line 82: see above

napari/layers/volume/volume.py
```diff
@@ -74,12 +82,21 @@ def __init__(
         colormap='gray',
         clim=None,
         clim_range=None,
+        opacity=1,
```
kne42, line 85: see above

kne42 (approved):
would like to see all arguments shared between all layer types as keyword-only

napari/layers/base/base.py
sofroniewn: fixed

napari/layers/base/base.py
sofroniewn: hmm, we've been using "extended summary" in other places, i think at the suggestion of @jni - happy to refactor, but would rather have a separate PR that fixes everywhere

napari/layers/base/base.py
sofroniewn: we'll need to fix this on our other layers too then - i'd rather wait for now and have a separate PR do all the doc strings for all the layers

napari/layers/base/base.py
sofroniewn: fixed, but we'll need to fix elsewhere too

napari/layers/image/image.py
sofroniewn: we already have a `*` at line 109. why do we need another?

```python
    def __init__(
        self,
        image,
        *,
        metadata=None,
        multichannel=None,
        colormap='gray',
        clim=None,
        clim_range=None,
        interpolation='nearest',
        opacity=1,
        blending='translucent',
        visible=True,
        name=None,
        **kwargs,
    ):
```

kne42: didn't see that, mb

PR #435, opened 2019-07-21
more dims fixes

sofroniewn:
# Description
Quick fix for some dims sliders problems. This will still need to be improved more to allow for better creating / destroying sliders, but I'll tackle that in a follow-up PR

## Type of change

## Final checklist:

### DIFF
--- a/napari/components/viewer_model.py
+++ b/napari/components/viewer_model.py
@@ -668,7 +668,11 @@ class ViewerModel(KeymapMixin):
             self.active_layer = active_layer
 
     def _on_layers_change(self, event):
-        self.dims.range = self._calc_layers_ranges()
+        range = self._calc_layers_ranges()
+        self.dims.ndim = len(range)
+        for i, r in enumerate(range):
+            self.dims.set_range(i, r)
+        self.dims.events.ndim()
 
     def _calc_layers_ranges(self):
         """Calculates the range along each axis from all present layers.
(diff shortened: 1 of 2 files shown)


napari/components/viewer_model.py
```diff
@@ -668,7 +668,11 @@ def _update_active_layer(self, event):
             self.active_layer = active_layer
 
     def _on_layers_change(self, event):
-        self.dims.range = self._calc_layers_ranges()
+        range = self._calc_layers_ranges()
```
jni: Don't overwrite the `range` builtin here. `layer_range`?

jni (changes requested):
@sofroniewn could you elaborate either in this PR or in a commit message why you made these changes? I have zero understanding of the consequences of the removal, why it's needed, and what problem you are trying to solve. It's useful to have a reference discussion e.g. if the same problem surfaces in the future.

sofroniewn:
This PR removes the ability to directly set the `range` property on the `Dims` model. In the previous range setter you were able to simultaneously set new range values and change the dimensions, this triggered events that were then causing updates before all the new values had been set. I.e. you could trigger the `ndims` event before the new range values were set or you couldn't properly set the new range values because you hadn't expanded the dims yet.

This PR decouples that process by dropping the `range` setter and forcing you to use the `set_range` method instead. Because the dims sometimes do need to change, this is now done outside the set_range method. 

Is this clear @jni?

jni (approved):

jni:
@sofroniewn clear enough. =)

PR #437, opened 2019-07-22
fix screenshot

### LINKED ISSUE #181: Viewer screenshot method returns only the first layer.
## 🐛 Bug
When screenshot method is used on a viewer with multiple layers, only the first layer is returned.

## To Reproduce

1. import

%gui qt5 
from napari import ViewerApp
import vispy.io as io

2. Use ViewerApp() and display multiple layers.

3. save screenshot of current view.

screen=viewer2D.screenshot()
io.write_png('vectorexport.png',screen)

vectorexport

## Expected behavior
Return an RGB image of the current view a numpy array.
image (2)

## Environment

 - napari Version (e.g., 1.0): vector branch, March 20, 2019
 - OS (e.g., Linux): Mac OS Mojave 10.4
 - Python version: 3.7

sofroniewn:
# Description
This PR fixes #181 ultimately by extracting the pixels sent to the screen from the `Qt` element directly. I was unable to fix the behavior of vispy's `render` and I had trouble getting vispy `_screenshot` to work directly, but we don't actually need either of these as we have the `Qt` objects and so can just grab the `QImage` that is actually being displayed and convert it back to a numpy array. 

The syntax is slightly different for `PySide2` vs `PyQt5`, but it's easy to handle. I've verified now this method works with both backends and with scripting usage and notebook usage. I added a line to the `to_svg` example to capture a screenshot. See here for a saved png

screenshot

I'm pretty excited about this feature as it will now allow us to make awesome animations and movies!

## Type of change

# How has this been tested?

- [x] `examples/to_svg`

## Final checklist:

### DIFF
--- a/examples/to_svg.py
+++ b/examples/to_svg.py
@@ -6,6 +6,7 @@ your shapes.
 
 import numpy as np
 from skimage import data
+from skimage.io import imsave
 import napari
 from vispy.color import Colormap
 
@@ -122,3 +123,8 @@ with napari.gui_qt():
 
     svg = viewer.to_svg()
     # svg = viewer.to_svg(file='viewer.svg')
+
+    # take screenshot
+    screenshot = viewer.screenshot()
+    viewer.add_image(screenshot, multichannel=True, name='screenshot')
+    # imsave('screenshot.png', screenshot)
--- a/examples/to_screenshot.py
+++ b/examples/to_screenshot.py
@@ -6,6 +6,7 @@ your shapes.
 
 import numpy as np
 from skimage import data
+from skimage.io import imsave
 import napari
 from vispy.color import Colormap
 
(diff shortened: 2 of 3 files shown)


examples/to_svg.py
```diff
@@ -122,3 +123,8 @@
 
     svg = viewer.to_svg()
     # svg = viewer.to_svg(file='viewer.svg')
+
+    # take screenshot
+    screenshot = viewer.screenshot()
```
kne42: not sure if this belongs under something titled `to_svg`... maybe rename it or make this its own example?

kne42 (approved):
very clever fix!

examples/to_svg.py
sofroniewn: make new example, thanks

PR #439, opened 2019-07-23
test add_* signatures and improve docstring testing

kne42:
## Changes
- rename docstring test
- improve coverage & implementation of docstring test for layers and their convenience methods
- add signature test for layers and their convenience methods
- remove `__new__` method from `KeymapMixin` to stop it from overriding class signatures

### DIFF
--- a/napari/components/viewer_model.py
+++ b/napari/components/viewer_model.py
@@ -292,7 +292,7 @@ class ViewerModel(KeymapMixin):
         return svg
 
     def add_layer(self, layer):
-        """Adds a layer to the viewer.
+        """Add a layer to the viewer.
 
         Parameters
         ----------
@@ -399,7 +399,7 @@ class ViewerModel(KeymapMixin):
         return layer
 
     def add_pyramid(self, pyramid, *args, **kwargs):
-        """Adds an image pyramid layer to the layers list.
+        """Add an image pyramid layer to the layers list.
 
         Parameters
         ----------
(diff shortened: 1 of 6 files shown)


napari/components/viewer_model.py
```diff
@@ -366,12 +381,25 @@ def add_image(self, image, *args, **kwargs):
...
+            multichannel=multichannel,
+            colormap=colormap,
+            clim=clim,
+            clim_range=clim_range,
+            interpolation=interpolation,
+            opacity=opacity,
+            blending=blending,
+            visible=visible,
+            name=name,
+            **kwargs,
```
sofroniewn, line 396: sometimes we have `**kwargs` - sometimes we don't - see `add_points`. are they just missing in `add_points`? do we want them everywhere? i think we can drop them, but curious why we'd want them

napari/tests/test_viewer_layer_parity.py
```diff
@@ -0,0 +1,115 @@
...
+    method_returns = *method_returns[:-1], description
+
+    assert method_returns == (
+        'layer',
+        f':class:`napari.layers.{name}`',
+        f'The newly-created {name.lower()} layer.',
+    ), f"improper 'Returns' section of '{method_name}'"
+
+
+@pytest.mark.parametrize('layer', layers, ids=lambda layer: layer.__name__)
```
sofroniewn, line 83: very nice

sofroniewn (approved):
approved - though see my minor comment on the `kwrd` ags

napari/components/viewer_model.py
kne42: This PR merely ensures that layers and their respective convenience methods have the same signatures and parameter docs. As to why the `Image`, `Labels`, and `Volume` layers contain unused, undocumented `**kwargs` should be addressed elsewhere.

sofroniewn: makes sense

PR #440, opened 2019-07-23
swappable dims

sofroniewn:
# Description
This PR adds support for swappable dims for the image layer (so far) - need to propagate changes to all layers. It adds a `swap_display` method to the `dims` object that will swap the display state of the dimensions - see here for usage:

swap_dims

@pranathivemuri - this will have implications for our support of 3D rendering for all layer types. we can work together on anything you're doing around the points layer.

## Type of change

# How has this been tested?

- [x] add new `example/swap_dims_image.py`

## Final checklist:


sofroniewn:
I'll note that right now you can't swap `XxY` to appear as `YxX` but that is probably desirable too, might need to think of a different way to do this using a vispy transform, but that can come later

sofroniewn:
Swappable dims now works with the `Points` layer too:

swap_dims_points

Note I've added a custom keybinding to this example that iterates through the dims

sofroniewn:
Changes in this PR also add compatibility for 3D points, but there are some clear vispy issues we need to solve (not how the size of the points changes depending on the view angle), and the selection tools cause errors, and the add tool might not behave as expected. Fixing these problems isn't a goal of this PR, but I wanted to make sure any changes to enable swappable dims were compatible with this usage mode:

3D_points

sofroniewn:
This PR now adds swappable dims to all layers apart from the `Shapes` layer (note this is different from 3D viewing).

For the `Shapes` layer we need to think about what our shapes should do when we swap dimensions as right now shapes are fundamentally 2D. The more I think about this the more I wonder if we shouldn't do some projection magic like we do in the `Vectors` layer - this will may make our path towards `3D` shapes easier too. Thoughts @jni?

For now though I think we should review this PR as is and we can the focus on the shapes layer and other `3D` stuff in the future. Thoughts @royerloic @kne42 @pranathivemuri ?

royerloic:
> This PR now adds swappable dims to all layers apart from the `Shapes` layer (note this is different from 3D viewing).

That's great!

> For the `Shapes` layer we need to think about what our shapes should do when we swap dimensions as right now shapes are fundamentally 2D. The more I think about this the more I wonder if we shouldn't do some projection magic like we do in the `Vectors` layer - this will may make our path towards `3D` shapes easier too. Thoughts @jni?

Even if shapes are only drawn  in 2D, they can still be thought as 3D objects.
When changing the displayed dimensions -- in any case really -- we always show
a 2D or 3D projection or slice of these objects. Slicing is preferred  of course because
that's what we do for images. Of course much easier said than done.
For points it is rather simple, for polygons it gets more difficult. But, we could for example
show the 1D section of a polygon which just means drawing a series of lines.
Projections make sense for vector fields because slicing a vector is a bit uninteresting...

> For now though I think we should review this PR as is and we can the focus on the shapes layer and other `3D` stuff in the future. Thoughts @royerloic @kne42 @pranathivemuri ?
I agree.
Shapes would have to be extended to meshes, and that becomes very difficult in arbitrary 
dimensions. A possible approach is to build all nD shapes by nD simplices (plural of simplex)
which are essentially: triangles (2D),tetrahedra (3D), n-simplex (nD). We might want to take a more pragmatic path and just support adhoc shapes in 2D and 3D. What does the 'Guardian of  Dimensions' 
@jni think about that ?

examples/nD_vectors_image.py
```diff
@@ -24,15 +24,13 @@
...
     pos = np.zeros(shape=(m, n, p, 3), dtype=np.float32)
-    rand1 = 2 * (np.random.random_sample(n * m * p) - 0.5)
-    rand2 = 2 * (np.random.random_sample(n * m * p) - 0.5)
-
-    # assign projections for each vector
-    pos[:, :, :, 0] = 0
-    pos[:, :, :, 1] = rand1.reshape((m, n, p))
-    pos[:, :, :, 2] = rand2.reshape((m, n, p))
+    for i in range(3):
+        pos[:, :, :, i] = (
```
kne42: usually we would just
```suggestion
        pos[..., i] = (
```
but it doesn't really matter

examples/swap_dims.py
```diff
@@ -0,0 +1,36 @@
...
+            for f in np.linspace(0.05, 0.5, 10)
+        ],
+        axis=0,
+    )
+    viewer = napari.view(blobs.astype(float))
+
+    # add the points
+    points = np.array(
+        [
+            [0, 0, 0, 100],
```
kne42, line 25: sidenote: can we disable `black` on the examples? this looks really atrocious

napari/_qt/qt_viewer.py
```diff
@@ -108,35 +108,35 @@ def __init__(self, viewer):
...
-            self.view.camera.flip = (0, 1, 0)
-            min_shape, max_shape = self.viewer._calc_bbox()
-            centroid = np.add(max_shape, min_shape) / 2
-            size = np.subtract(max_shape, min_shape)
-            # Scale the camera to the contents in the scene
-            if len(centroid) > 0:
-                centroid = centroid[-3:]
-                self.view.camera.center = centroid[::-1]
-                self.view.camera.scale_factor = 1.5 * np.mean(size[-3:])
+            if type(self.view.camera) != ArcballCamera:
```
kne42:
```suggestion
            if not isinstance(self.view.camera, ArcballCamera):
```

pranathivemuri:
Great job getting the 3D points to work, I was making similar changes in my branch seems like you beat me to it - https://github.com/napari/napari/compare/master...pranathivemuri:pranathi-points

We should discuss the polygon shapes in 3D layers for sure. Do we want to support ROI selection in 3D based on the shapes, it might be cool. but I am not sure how exactly shapes layer should be modified to support that just yet. The selection tool error is related this as when we select now, it draws a square in 2D, but in 3D we should be doing a cube.

Either way, we should merge this PR as is and write notes/todos to fix stuff in points layer to support 3D stuff so we can approach them in the next PR.

napari/_qt/qt_viewer.py
```diff
@@ -108,35 +108,35 @@ def __init__(self, viewer):
...
+                self.view.camera = ArcballCamera(name="ArcballCamera")
+                # flip y-axis to have correct alignment
+                self.view.camera.flip = (0, 1, 0)
+
+                self.view.camera.viewbox_key_event = viewbox_key_event
+                # TO DO: Remove
```
kne42: remove what, why, when?

napari/_qt/qt_viewer.py
```diff
@@ -108,35 +108,35 @@ def __init__(self, viewer):
...
+                self.viewer._view = self.view
+                self.viewer.reset_view()
         elif np.sum(self.viewer.dims.display) == 2:
             # Set 2D camera
-            self.view.camera = PanZoomCamera(aspect=1, name="PanZoomCamera")
-            # flip y-axis to have correct alignment
-            self.view.camera.flip = (0, 1, 0)
-            # Scale the camera to the contents in the scene
-            self.view.camera.set_range()
+            if type(self.view.camera) != PanZoomCamera:
```
kne42:
```suggestion
            if not isinstance(self.view.camera, PanZoomCamera):
```

napari/_qt/qt_viewer.py
```diff
@@ -108,35 +108,35 @@ def __init__(self, viewer):
...
+                self.view.camera = PanZoomCamera(
+                    aspect=1, name="PanZoomCamera"
+                )
+                # flip y-axis to have correct alignment
+                self.view.camera.flip = (0, 1, 0)
+
+                self.view.camera.viewbox_key_event = viewbox_key_event
+                # TO DO: Remove
```
kne42: remove what, why, when?

napari/_qt/qt_viewer.py
```diff
@@ -214,7 +214,15 @@ def _on_cursor(self, event):
         self.canvas.native.setCursor(q_cursor)
 
     def _on_reset_view(self, event):
-        self.view.camera.rect = event.viewbox
+        if type(self.view.camera) == PanZoomCamera:
```
kne42: isinstance

napari/_qt/qt_viewer.py
```diff
@@ -214,7 +214,15 @@ def _on_cursor(self, event):
...
+            self.view.camera.rect = event.viewbox
+        elif type(self.view.camera) == ArcballCamera:
```
kne42: isinstance

napari/_qt/qt_viewer.py
```diff
@@ -214,7 +214,15 @@ def _on_cursor(self, event):
...
+            self.view.camera.center = event.center
+            self.view.camera.scale_factor = event.scale_factor
+        else:
+            raise ValueError(
+                "Invalid camera type {}".format(type(self.view.camera))
```
kne42: use an f-string

napari/components/dims.py
```diff
@@ -252,6 +252,28 @@ def set_display(self, axis: int, display: bool):
...
+        axis_a : int
+            Dimension index
+        axis_b : int
+            Dimension index.
+        """
+        if axis_a < 0:
+            axis_a += self.ndim
+        if axis_b < 0:
+            axis_b += self.ndim
+
```
kne42: add a check if they are still negative after the wraparound and error accordingly

napari/components/dims.py
```diff
@@ -252,6 +252,28 @@ def set_display(self, axis: int, display: bool):
...
+        if self.display[axis_a] != self.display[axis_b]:
+            state = self.display[axis_a]
+            self._display[axis_a] = self.display[axis_b]
+            self._display[axis_b] = state
```
kne42: due to how python's assignment expressions work, you can actually do the following:
```python
self._display[axis_a], self._display[axis_b] = self.display[axis_b], self.display[axis_a]
```

napari/components/viewer_model.py
```diff
@@ -213,15 +214,24 @@ def reset_view(self):
...
-        shape = 1.1 * shape
-        rect = (min_shape[1], min_shape[0], shape[1], shape[0])
-        self.events.reset_view(viewbox=rect)
+        centroid = np.add(min_shape, max_shape) / 2
+        centroid = [centroid[i] for i in np.where(self.dims.display)[0]]
+        size = np.subtract(max_shape, min_shape)
+        size = [size[i] for i in np.where(self.dims.display)[0]]
+        corner = [min_shape[i] for i in np.where(self.dims.display)[0]]
+
+        if np.sum(self.dims.display) == 2:
```
kne42: im confused what does this do? also you should be able to use the built-in `sum` here

napari/layers/base/base.py
```diff
@@ -199,6 +201,11 @@ def indices(self, indices):
         self._update_coordinates()
         self._set_view_slice()
 
+    @property
+    def displayed(self):
+        """Tuple of bool: If dimension is diplayed or not."""
+        return tuple(type(i) == slice for i in self.indices)
```
kne42: isinstance

napari/layers/base/base.py
```diff
@@ -215,13 +222,29 @@ def _update_coordinates(self):
         """Insert the cursor position (x, y) into the correct position in the
         tuple of indices and update the cursor coordinates.
         """
+        displayed = np.where(self.displayed)[0]
```
kne42: personally i feel like this is much more elegant (note that `np.where(arr)` == `np.nonzero(arr)`)
```suggestion
        displayed, = np.nonzero(self.displayed)
```

but maybe what it does is less immediately obvious

napari/layers/image/image.py
```diff
@@ -416,8 +421,9 @@ def to_xml_list(self):
         image_str = imwrite('<bytes>', mapped_image, format='png')
         image_str = "data:image/png;base64," + str(b64encode(image_str))[2:-1]
         props = {'xlink:href': image_str}
-        width = str(self.shape[-1])
-        height = str(self.shape[-2])
+        disp = np.where(self.displayed)[0]
```
kne42: see above

napari/layers/labels/labels.py
```diff
@@ -386,7 +382,8 @@ def fill(self, coord, old_label, new_label):
         else:
             # work with just the sliced image
             labels = self._data_view
-            slice_coord = tuple(int_coord[-2:])
+            disp = np.where(self.displayed)[0]
```
kne42: ''

napari/layers/labels/labels.py
```diff
@@ -539,8 +548,9 @@ def to_xml_list(self):
         image_str = imwrite('<bytes>', mapped_image, format='png')
         image_str = "data:image/png;base64," + str(b64encode(image_str))[2:-1]
         props = {'xlink:href': image_str}
-        width = str(self.shape[-1])
-        height = str(self.shape[-2])
+        disp = np.where(self.displayed)[0]
```
kne42: ''

napari/layers/points/points.py
```diff
@@ -407,7 +403,8 @@ def selected_data(self, selected_data):
             with self.block_update_properties():
                 self.face_color = face_color
 
-        size = list(set([self.sizes[i, -2:].mean() for i in index]))
+        displayed = np.where(self.displayed)[0]
```
kne42: ''

napari/layers/points/points.py
```diff
@@ -407,7 +403,8 @@ def selected_data(self, selected_data):
...
+        size = list(set([self.sizes[i, displayed].mean() for i in index]))
```
kne42: does this really need to be a list? if so you should look at updating the `size` property since it should ideally take in any sequence type

napari/layers/points/points.py
```diff
@@ -526,12 +523,15 @@ def _slice_data(self, indices):
             less than 1 correspond to points located in neighboring slices.
         """
         # Get a list of the data for the points in this slice
+        not_disp = np.where([not x for x in self.displayed])[0]
+        disp = np.where(self.displayed)[0]
```
kne42: ''

napari/layers/points/points.py
```diff
@@ -526,12 +523,15 @@ def _slice_data(self, indices):
             less than 1 correspond to points located in neighboring slices.
         """
         # Get a list of the data for the points in this slice
+        not_disp = np.where([not x for x in self.displayed])[0]
```
kne42:
```suggestion
        not_disp, = np.invert(self.displayed).nonzero()
```

napari/layers/base/base.py
kne42: honestly this is repeated so much you should just turn it into a convenience property

napari/layers/points/points.py
```diff
@@ -897,10 +909,11 @@ def on_mouse_move(self, event):
                     self._move(self.selected_data, self.coordinates)
                 else:
                     self._is_selecting = True
+                    disp = np.where(self.displayed)[0]
```
kne42: ''

napari/layers/pyramid/pyramid.py
```diff
@@ -135,8 +131,8 @@ def data_level(self, level):
         if self._data_level == level:
             return
         self._data_level = level
-        # TODO: Change dims selection when dims model changes
-        self.scale = self.level_downsamples[self.data_level, [-1, -2]]
+        disp = np.where(self.displayed)[0]
```
kne42: ''

napari/layers/pyramid/pyramid.py
```diff
@@ -170,36 +166,38 @@ def _slice_data(self):
...
+        for i, d in enumerate(self.displayed):
+            if not d:
+                ind = np.round(
+                    indices[i] / self.level_downsamples[self.data_level, i]
+                ).astype(int)
+                indices[i] = np.clip(
+                    ind, 0, self.level_shapes[self.data_level][i] - 1
+                )
+
+        disp = np.where(self.displayed)[0]
```
kne42: ''

napari/layers/pyramid/pyramid.py
```diff
@@ -253,19 +251,26 @@ def get_value(self):
...
+                coord[i] = np.clip(coord[i], 0, shape[j] - 1)
+                slice_coord.append(coord[i])
+                j += 1
 
-        pos_in_slice = (
-            self.coordinates[-2:] + self.translate[[1, 0]] / self.scale[:2]
-        )
+        value = self._data_view[tuple(slice_coord)]
+
+        disp = np.where(self.displayed)[0]
```
kne42: ''

napari/layers/pyramid/pyramid.py
```diff
@@ -293,8 +298,8 @@ def compute_data_level(self, size):
         diff = size - max_size + 1
 
         # Find closed downsample level to diff
-        # TODO: Change dims selection when dims model changes
-        ds = self.level_downsamples[:, -2:].max(axis=1)
+        disp = np.where(self.displayed)[0]
```
kne42: ''

napari/layers/pyramid/pyramid.py
```diff
@@ -314,7 +319,8 @@ def find_top_left(self):
         pos = transform.map([0, 0])[:2] + self.translate[:2] / self.scale[:2]
 
         # TODO: Change dims selection when dims model changes
-        shape = self.level_shapes[0][-2:]
+        disp = np.where(self.displayed)[0]
```
kne42: ''

napari/layers/vectors/vectors.py
```diff
@@ -137,12 +133,15 @@ def data(self, vectors: np.ndarray):
 
         self._data = vectors_to_coordinates(vectors)
 
+        disp = np.where(self.displayed)[0]
```
kne42: ''

napari/layers/vectors/vectors.py
```diff
@@ -137,12 +133,15 @@ def data(self, vectors: np.ndarray):
...
         vertices, triangles = generate_vector_meshes(
-            self._data, self.edge_width, self.length
+            self._data[:, :, disp], self.edge_width, self.length
         )
         self._mesh_vertices = vertices
         self._mesh_triangles = triangles
+        self._displayed_stored = copy(self.displayed)
```
kne42, line 143:
```suggestion
        self._displayed_stored = self.displayed.copy()
```

napari/layers/vectors/vectors.py
```diff
@@ -213,15 +216,28 @@ def edge_color(self, edge_color: str):
...
+            disp = np.where(self.displayed)[0]
+            vertices, triangles = generate_vector_meshes(
+                self.data[:, :, disp], self.edge_width, self.length
+            )
+            self._mesh_vertices = vertices
+            self._mesh_triangles = triangles
+            self._displayed_stored = copy(self.displayed)
+
         vertices = self._mesh_vertices
+        not_disp = np.where([not x for x in self.displayed])[0]
```
kne42: ''

napari/layers/vectors/vectors.py
```diff
@@ -213,15 +216,28 @@ def edge_color(self, edge_color: str):
     def _set_view_slice(self):
         """Sets the view given the indices to slice with."""
 
+        if not np.all(self.displayed == self._displayed_stored):
+            disp = np.where(self.displayed)[0]
```
kne42: ''

napari/layers/vectors/vectors.py
```diff
@@ -213,15 +216,28 @@ def edge_color(self, edge_color: str):
...
+            vertices, triangles = generate_vector_meshes(
+                self.data[:, :, disp], self.edge_width, self.length
+            )
+            self._mesh_vertices = vertices
+            self._mesh_triangles = triangles
+            self._displayed_stored = copy(self.displayed)
+
         vertices = self._mesh_vertices
+        not_disp = np.where([not x for x in self.displayed])[0]
+        disp = np.where(self.displayed)[0]
```
kne42: ''

napari/layers/vectors/vectors.py
```diff
@@ -250,15 +266,13 @@ def _update_thumbnail(self):
         # calculate min vals for the vertices and pad with 0.5
         # the offset is needed to ensure that the top left corner of the
         # vectors corresponds to the top left corner of the thumbnail
-        offset = np.array([self.range[-2][0], self.range[-1][0]]) + 0.5
+        disp = np.where(self.displayed)[0]
```
kne42: ''

napari/layers/vectors/vectors.py
```diff
@@ -250,15 +266,13 @@ def _update_thumbnail(self):
...
+        offset = np.array([self.range[d][0] for d in disp]) + 0.5
```
kne42: if this is a 1d array
```suggestion
        offset = np.fromiter(self.range[d][0] for d in disp) + 0.5
```

examples/nD_vectors_image.py

examples/swap_dims.py
sofroniewn: agreed it doesn't look great - but I'd prefer to leave it an maybe find another way to save / generate these arrays especially as they are often used across multiple examples

napari/_qt/qt_viewer.py
sofroniewn: fixed

napari/_qt/qt_viewer.py
sofroniewn: added explanation

napari/_qt/qt_viewer.py
sofroniewn: changed

napari/_qt/qt_viewer.py
sofroniewn: added explanation

napari/_qt/qt_viewer.py
sofroniewn: fixed

napari/_qt/qt_viewer.py
sofroniewn: fixed

napari/_qt/qt_viewer.py
sofroniewn: fixed

napari/components/dims.py
sofroniewn: i feel like we should be better about input checking in general (i.e. negatives, or longer than the list, etc. I could just take a modulus and it will work every time - `axis_b = axis_b % self.ndim`?

sofroniewn: fixed with `%`

napari/components/dims.py
sofroniewn: done

napari/components/viewer_model.py
sofroniewn: explained what i'm doing. i like the consistency of the `np.sum` if that's ok

napari/layers/base/base.py
sofroniewn: fixed

napari/layers/vectors/vectors.py
sofroniewn: fixed

napari/layers/vectors/vectors.py
sofroniewn: fixed

napari/layers/base/base.py
sofroniewn: fixed all of them!

napari/layers/points/points.py
sofroniewn: fixed

sofroniewn:
I've addressed the changes requested by @kne42 and this PR is now ready for re-review. It now supports transposing axes within the displayed axes by switching the dims model to using `(None, None, 0, 1)` like representations as suggested by @royerloic. To change the displayed axes you should use the `dims.swap` method that will swap two axes.

An alternative `api` would be to have two variables, `ndisplay` which is either `2` or `3` depending on `2D` or `3D` mode and `order` which is a tuple of length `ndim` that contains the order of the dimensions where the last `2` or `3` are displayed and the order of the others corresponds to the order of the sliders, so if you had `order = (2, 0, 1, 3)` and `ndisplay = 2` then we'd display the 1st and 3rd axis as Y x X. This might actually be a lot cleaner ... maybe I'll do one more refactor!! But can @royerloic, @jni or @kne42 weigh in first?

sofroniewn:
superseeded by #450

royerloic:
> I've addressed the changes requested by @kne42 and this PR is now ready for re-review. It now supports transposing axes within the displayed axes by switching the dims model to using `(None, None, 0, 1)` like representations as suggested by @royerloic. To change the displayed axes you should use the `dims.swap` method that will swap two axes.
> 
> An alternative `api` would be to have two variables, `ndisplay` which is either `2` or `3` depending on `2D` or `3D` mode and `order` which is a tuple of length `ndim` that contains the order of the dimensions where the last `2` or `3` are displayed and the order of the others corresponds to the order of the sliders, so if you had `order = (2, 0, 1, 3)` and `ndisplay = 2` then we'd display the 1st and 3rd axis as Y x X. This might actually be a lot cleaner ... maybe I'll do one more refactor!! But can @royerloic, @jni or @kne42 weigh in first?

So actually I like your proposal, right now the scheme [None, None, 1, 0] mixes two things: (i) what is displayed, (ii) axis ordering. That's bad. The dims model design principle is to avoid mixing things because, for example, the 'None' entries obfuscate the order of the underlying axis which means that you can't change what axis are displayed without 'erasing' information. So yes, I like the idea of the number of axis displayed separately from the ordering of axis.  Great idea.  
A slight variation would be to have instead of a number of displayed axis an actual _set_ of axis. This provides maybe an even clearer separation between the two aspects, you can reorder the axis however you want, and then pick a _set_ of two axis to display (e.g. {2,3}), the set does not carry order, that is the job of the permutation list (e.g. [2,0,1,3] ). This gives the freedom to reorder displayed or not displayed axis (think slider order too) independently of which are chosen for display.  So you could for example, permutate slider axis without messing with which axis are displayed. What do  you think?

sofroniewn:
Thanks. I’ve gone ahead and implemented the new scheme on #450 and closed this PR. I think the number of displayed is working fine, but we’ll see comments after review

examples/swap_dims.py
kne42: I think examples should be self-contained so unless you're loading arrays from the cloud/are included in a pip-installable package you shouldn't be referencing any external images.

napari/components/dims.py
kne42: it would work every time but it doesn't make sense – if someone provides an index of `-5` when there are 2 axes that should raise an exception. you can see with lists, numpy, etc. it does not "wrap around" more than once and shouldn't even be thought about as wrapping around really, it's more like the index from the last element instead of the first

PR #442, opened 2019-07-25
[WIP] Console

sofroniewn:
# Description
This PR is complementary to #421 and focuses on the getting the Qt aspects of embedding the console. Right now though it creates a new ipython console and provides the current `ViewerModel` as a variable called `viewer`. Nothing else is provided. See here:

console

Right now it can't be launch from `IPython` as it gets the following error
```python
MultipleInstanceError: Multiple incompatible subclass instances of InProcessInteractiveShell are being created.
```
But we can work on that

I'm also still having some problems with the sytlesheets, but I'll work on those

## Type of change

- [x] New feature (non-breaking change which adds functionality)

# How has this been tested?

- [ ] example: the test suite for my feature covers cases x, y, and z
- [ ] example: all tests pass with my change

## Final checklist:
- [ ] My PR is the minimum possible work for the desired functionality
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #443, opened 2019-07-25
add qt console

sofroniewn:
# Description
This PR supersedes #442 and adds a console to the viewer. When launching from a python script or the CLI the console just has the instantiated Viewer as a variable called `viewer`.

When launching from a jupyter notebook the console has access to the same kernal as the notebook and everything is shared 100% between them.

Launching from `IPython` is currently not implemented.

There are a couple of things still to implement, like improving the stylings, cleaning up the code - but we should be close to getting an initial pass in.

console

## Type of change

# How has this been tested?

## Final checklist:


sofroniewn:
This PR is moving along well - i've improved the stylings for the console - which includes both its general stylings and it's syntax stylings. for the possible supported syntax stylings see here - https://help.farbox.com/pygments.html @freeman-lab might have some suggestions!

I still have to capture the globals state, and the jupyter notebook requires shift clicking.

Also if you give this a try make sure to update to our latest requirements, as the versions for `IPython` and `ipykernel` have to be the latest

sofroniewn:
This PR is now ready for review - it adds a console to napari.

When launched from a script or the CLI a new console gets created and instantiated with the variable `viewer` for the `Viewer`. If you at any point in your script run `viewer.update_console(locals())` all the local variables defined up to that point will be put in the namespace of the console. (see `examples/update_console.py` for usage). You can also pass `update_console` a dictionary of variable names / variables to customize what you pass over too. @jni does this functionality meet your needs? You have to choose to run the command, but I think that is a small ask of people to make for a lot of functionality.

When launching from `iPython` the console is not connected and the button to display it has no effect - this is because `iPython` doesn't actually create a kernel for us to connect to, and you have the `iPython` console that you launched from right there - so the same functionality is available to you.

When launching from a `jupyter notebook` you get a console with exactly the same shared state as the jupyter notebook and creating variables in either one will immediately be shared with the other. Unfortunately to execute commands on the napari console in this usage mode you need to press `shift-enter` instead of just `enter`. I cannot figure out how to overwrite this - the suggested `execute_on_complete_input` property still appears to be `True` but is having no effect. We might need to ask the `qtconsole` team about this at some point, but I think that can come after this PR merges.

As mentioned in comments above the stylings of the console can be done using our themes, including a syntax styling. @freeman-lab I'm not sure I've done the best job with the colorings, but checkout the gif below for examples

console_full

Finally it seems like our tests are failing - but it could be because our CI is getting overwhelmed as all tests pass locally for me and the error messages look like this:
```
OSError: [Errno 24] Too many open files
```
maybe @kne42 has an idea on how to fix.

napari/_qt/qt_console.py
```diff
@@ -0,0 +1,84 @@
...
+            # kernel
+            kernel_client = QtKernelClient(
+                connection_file=get_connection_file()
+            )
+            kernel_client.load_connection_file()
+            kernel_client.start_channels()
+
+            self.kernel_client = kernel_client
+            self.shell = shell
+            self.push = self.shell.push
```
jni, line 87: @sofroniewn regarding the shift enter shortcut, this comment looks relevant:

https://github.com/jupyterlab/jupyterlab/pull/3410#issuecomment-356818449

jni: Found from here: https://github.com/jupyterlab/jupyterlab/issues/3643#issuecomment-357744372

jni: ... But all of that stuff is from jupyterlab config... Super weird.

napari/_qt/qt_console.py
```diff
@@ -0,0 +1,84 @@
...
+    def __init__(self, user_variables=None):
+        super().__init__()
+
+        # get current running instance or create new instance
+        shell = get_ipython()
+
+        if shell is None or type(shell) == InProcessInteractiveShell:
+            # If there is no currently running instance create an in-process
+            # kernel or if there is an old running InProcessInteractiveShell
+            # then just create a new one - necessary for our tests to pass
```
jni: Can you elaborate on this point?

napari/_qt/qt_console.py
```diff
@@ -0,0 +1,84 @@
...
+        else:
+            raise ValueError(
+                'ipython shell not recognized; ' f'got {type(shell)}'
+            )
+
+        # Add any user variables
+        user_variables = user_variables or {}
+        self.push(user_variables)
+
+        self.enable_calltips = False
```
jni, line 97: Hahaha :+1:

napari/_qt/qt_layerlist.py
```diff
@@ -27,6 +27,7 @@ def __init__(self, layers):
         self.centers = []
         self.setAcceptDrops(True)
         self.setToolTip('Layer list')
+        self.setFixedWidth(275)
```
jni: I'm always wary of magic numbers like this one... Should this be coming from some stylings?

napari/_qt/tests/test_qt_console.py
```diff
@@ -0,0 +1,21 @@
...
+        console = QtConsole()
+        assert console.kernel_client is not None
+
+
+def test_console_user_variables():
+    """Test creating the console with user variables."""
+    with gui_qt():
+
+        console = QtConsole({'var': 3})
+        assert console.kernel_client is not None
```
jni: Could you test accessing the variable within the kernel somehow?

napari/resources/stylesheet.qss
```diff
@@ -348,6 +365,14 @@ QtNewLabelsButton {
    image: url(":/icons/{{ folder }}/new_labels.svg");
 }
 
+QtConsoleButton[expanded=true] {
+  image: url(":/icons/{{ folder }}/properties_expand.svg");
+}
+
+ QtConsoleButton[expanded=false] {
+  image: url(":/icons/{{ folder }}/properties_contract.svg");
```
jni: Could we add a `>_` icon instead of reusing properties? Maybe even `>_` -> `V_`, although that makes less sense, but still kinda cute.

jni (approved):
@sofroniewn very nice PR, and surprisingly small! A testament to the great work of the IPython team.

1. Could we add locals when instantiating the viewer by default?
2. More contentiously, could they somehow be added with a delay so that you (with high probability) capture the state of the namespace at the end of the context?

It might be informative to check the code for `IPython.embed()` to get this to work.

Having said all this, I am recognising as I write the difficulty of this stuff (I can even envision ast parsing), so we can potentially push this to a later PR.

jni:
Here's our test running code. It opens the files in a context manager so the files *should* close:

https://github.com/napari/napari/blob/master/test_examples.py#L27

I might consider changing it to:

```python
with open(path, 'r') as f:
    code = f.read()
exec(code)
```

to ensure that the file is closed *before* we are executing code.

jni:
Having said that, another source of this could be reading of style sheets by the qt console...?

napari/_qt/qt_console.py
sofroniewn: thanks! I tried reading some those threads, but can't figure it out. It should be doable, but let's contact them after we merge this one

napari/_qt/qt_console.py
sofroniewn: :-)

napari/_qt/qt_layerlist.py
sofroniewn: good catch - yes these can go in the sytlesheet - moving all of them now

sofroniewn:
@jni I think the issue with the tests is that the actual GUI doesn't get closed before the next one runs - it is on a timer - and so if i just run `pytest .` locally and start moving the windows around I can see many many tests open and the same time before the next test finishes. I tried moving the `exec` out of the file reading but it didn't help. Maybe @kne42 has ideas?

napari/_qt/tests/test_qt_console.py
sofroniewn: done

napari/_qt/qt_console.py
sofroniewn: have changed this - the problem comes when we have multiple viewers launched from the same python script - could be because of our tests, or could not be. Right now it doesn't work so well. I think we can bump fixing this to followup PRs as I don't think it is a top priority use case

napari/resources/stylesheet.qss
sofroniewn: using `>_` - but it doesn't change anymore

sofroniewn:
ok - made some updates requested by @jni 

- Still not sure about the tests passing and all the open files.

- Still have to `shift-enter` in the jupyter notebook

- Launching multiple viewers from the same script isn't so great for the consoles apart from the last launched one. Not sure if there are implications for starFISH display @kne42?

I think if we can get the tests sorted out then we should merge

kne42:
> * Still not sure about the tests passing and all the open files.

yeah i think it definitely has to do with multiple tests running at once but im not sure what aspect – since you close each file after reading them and all file loading within each example should be closed as well

> * Launching multiple viewers from the same script isn't so great for the consoles apart from the last launched one. Not sure if there are implications for starFISH display @kne42?

this is definitely not ideal. starfish may have to deal with some issues if you want to open e.g. your imagestack and your nuclei in different windows but I'll defer to @ambrosejcarr to give the final verdict

freeman-lab:
this is looking really cool!

quick styling comment: at least for now i think i would go with a darker background (more like the first gif)

once this is merged in i'm happy to do a quick pass for proper styling consistency

kne42:
## OSX failures

the maximum number of files we can have open at once is 256 on the OSX machines so it's still unclear to me how we're reaching this limit unless the apps don't actually close, but then we would have noticed this issue a while ago

along with all of the OSError being thrown there seems to be some error where the traitlets library somehow is unable to access `iopub`-related attributes. from what research ive done, `iopub` is used to display any messages from the kernel, kind of like what you see in your terminal when you launch `jupyter notebook`

my guess is that maybe on this remote machine, jupyter cannot find the stdout to send messages to and falls back on writing logs to disk, opening tons of files in the process

## linux failures

this error is much more consistent but gives us almost no hint as to why it's failing. all i can tell is that is has something to do with nD_shapes

sofroniewn:
@kne42 thanks for looking into this. as i mentioned in the meeting maybe it's time to look into using a qt testing suite. this will also open the door for more rigorous testing of the gui too. here is the one which i had seen online that i thought looked good [pytest-qt](https://pypi.org/project/pytest-qt/).

royerloic:
> quick styling comment: at least for now i think i would go with a darker background (more like the first gif)

I agree!

sofroniewn:
So i've update things with our new `pytest-qt` system from #453 and all our linux tests now pass, but we're still getting failures on our osx tests possibly due to `OSError: [Errno 24] Too many open files`. It's unclear if `pytest-qt` has actually made a significant change here or if the number and complexity of the tests has decreased now that we aren't testing the entire examples folder but are testing in a more targeted way and with much smaller dataset sizes.

I'll note that the tests that are failing are the ones that are importing and launching the whole `Viewer` as opposed to pieces, and when I run `pytest .` locally a new viewer window will open for each test. Maybe we should add a line to the end of each test explicitly closing the window? I'll investigate. @kne42 might have ideas too

sofroniewn:
so calling the `close` method on the main window at the end of a test does cause the viewer to close before the next one starts (though it seems like that functionality should have been provided by the [qtbot](https://pytest-qt.readthedocs.io/en/latest/reference.html)). Unfortunately this has not fixed our tests. Possibly because the console itself isn't shutting down properly. I've actually seen some docs on how to shut the console down properly - hopefully adding this will fix the tests.

sofroniewn:
I'm now shutting the console down properly triggered off the close event of the main window. This causes all our tests to pass, and is probably the right way of doing things. I'm going to merge this now, there might be a few things around this to improve (like making sure qtbot closes the viewer), but they can come in follow up PRs.

jni:
Awesome work @sofroniewn!!!

PR #444, opened 2019-07-25
adapt existing keybindings to use new system

kne42:
please check that the function names make sense and that i didnt miss any!


sofroniewn (approved):
This looks great - i've checked everything and it all works as expected. I did find a bug when "shift" dragging, but the bug is on master and in a different part of the code so will be fixed in a followup PR.

I will also make followup PRs to add some keybindings to the viewer and will follow the conventions established here.

I will also look into what it will take to solve our keybindings focus issue with Qt / vispy

PR #450, opened 2019-07-27
Swappable dims order

sofroniewn:
# Description
This PR supersedes #440 in favor of a better implementation that now relies on an `dims.order` property and a `dims.ndisplay` property. This can be treated like a transpose, enabling rapid switching between any configuration of `2D` and `3D` viewing as seen below:

swap_dims2

Functionality is not yet extended to that shapes layer, but that can come in a followup PR.

## Type of change

# How has this been tested?

- [x] add `examples/swap_dims.py`

## Final checklist:


examples/add_image.py
```diff
@@ -42,3 +42,7 @@
     # change the layer interpolation mode
     layer.interpolation = 'bicubic'
     layer.interpolation = 'nearest'
+
+    # swap dimension order
+    viewer.dims.swap(0, 1)
+    viewer.dims.order = (0, 1)
```
jni: @sofroniewn can we try to keep examples semi-realistic? Rather than a big set of all supported APIs everywhere with non-sensical switches, each example should illustrate one or a handful of features...

examples/add_points.py
```diff
@@ -62,3 +62,7 @@
 
     # change the layer mode
     layer.mode = 'add'
+
+    # swap dimension order
+    viewer.dims.swap(0, 1)
+    viewer.dims.order = (0, 1)
```
jni: And the above comment goes doubly so for *repeating* non-sensical switches across multiple examples.

examples/nD_image.py
```diff
@@ -18,3 +18,4 @@
         axis=0,
     )
     viewer = napari.view(blobs.astype(float))
+    viewer.dims.swap(1, 2)
```
jni: I wonder whether we really need a "swap" method vs only using the `order` attribute... Or, probably, just set the `.visible` list?

examples/nD_points.py
```diff
@@ -34,3 +34,4 @@
     viewer.add_points(
         points, size=[0, 6, 10, 10], face_color='blue', n_dimensional=True
     )
+    viewer.dims.swap(1, 2)
```
jni: Again, let's keep examples distinct... I don't think we need to reorder dims in *every* example. On the one hand, I understand the appeal of showing off all capabilities in each example. But, showing off the simplicity of napari should also be a target for these examples.

examples/nD_vectors_image.py
```diff
@@ -24,15 +24,13 @@
...
     pos = np.zeros(shape=(m, n, p, 3), dtype=np.float32)
-    rand1 = 2 * (np.random.random_sample(n * m * p) - 0.5)
-    rand2 = 2 * (np.random.random_sample(n * m * p) - 0.5)
-
-    # assign projections for each vector
-    pos[:, :, :, 0] = 0
-    pos[:, :, :, 1] = rand1.reshape((m, n, p))
-    pos[:, :, :, 2] = rand2.reshape((m, n, p))
+    for i in range(3):
+        pos[..., i] = (2 * (np.random.random_sample(n * m * p) - 0.5)).reshape(
```
jni: This can just be `random_sample((m, n, p) - 0.5` (and the letters should be in that order), or `uniform(-0.5, 0.5, size=(m, n, p))`

examples/nD_vectors_image.py
```diff
@@ -24,15 +24,13 @@
...
+            (m, n, p)
+        )
 
     print(image.shape, pos.shape)
 
     # add the vectors
     vect = viewer.add_vectors(pos, edge_width=0.2, length=2.5)
+    viewer.dims.swap(0, 1)
```
jni: as above.

examples/swap_dims.py
```diff
@@ -0,0 +1,36 @@
+"""
+Display one 4-D image layer using the add_image API
```
jni: Please update the comment of this example to explain what it is meant to show.

napari/_qt/qt_dims.py
```diff
@@ -141,7 +141,11 @@ def _update_range(self, axis: int):
                 self._displayed[axis] = False
                 slider.hide()
             else:
-                if not self._displayed[axis]:
+                if (
+                    not self._displayed[axis]
+                    and self.dims.ndim - self.dims.order.index(axis)
+                    > self.dims.ndisplay
+                ):
```
jni, line 147: I don't understand this logic at all...?

napari/_qt/qt_dims.py
```diff
@@ -152,27 +156,18 @@ def _update_range(self, axis: int):
...
-        else:
-            self._displayed[axis] = True
-            slider.show()
-
+    def _update_display(self):
+        """Updates display for all sliders."""
+        for axis, slider in enumerate(self.sliders):
+            if (
+                self.dims.ndim - self.dims.order.index(axis)
+                <= self.dims.ndisplay
```
jni: This logic is very cumbersome and to me suggests that the data model needs updating. Perhaps we need `displayed` to be a list, which includes which axes are displayed and in which order...

napari/_qt/qt_dims.py
```diff
@@ -152,27 +156,18 @@ def _update_range(self, axis: int):
...
+            ):
+                self._displayed[axis] = False
+                slider.hide()
```
jni, line 164: I'm also confused... If an axis is removed from `displayed`, doesn't that mean that its slider should be *shown*?

napari/_qt/qt_dims.py
```diff
@@ -181,9 +176,9 @@ def _update_nsliders(self):
         Updates the number of sliders based on the number of dimensions
         """
         self._trim_sliders(0)
-        self._create_sliders(self.dims.ndim - 2)
-        for i in list(range(self.dims.ndim - 2)):
-            self._update_display(i)
+        self._create_sliders(self.dims.ndim)
+        self._update_display()
+        for i in list(range(self.dims.ndim)):
```
jni: remove `list()`

napari/_qt/qt_viewer.py
```diff
@@ -106,37 +106,37 @@ def __init__(self, viewer):
...
+                # TO DO: Remove once we have removed vispy node from individual
+                # layers as then model / view separate
+                self.viewer._view = self.view
+                self.viewer.reset_view()
         else:
             raise ValueError(
-                "Invalid display flags set in dimensions {}".format(
-                    self.viewer.dims.display
-                )
+                f"Invalid display flags set in dimensions {self.viewer.dims.ndisplay}"
```
jni: Should this error be caught earlier, in a property setter? ie I feel like this code should never be executed.

napari/_qt/tests/test_qt_dims.py
```diff
@@ -19,8 +19,10 @@ def test_creating_view():
         # Check that the dims model has been appended to the dims view
         assert view.dims == dims
 
-        # Check the number of sliders is two less than the number of dimensions
-        assert view.nsliders == ndim - 2
+        # Check the number of displayed sliders is two less than the number of
+        # dimensions
+        assert view.nsliders == ndim
+        assert np.sum(view._displayed) == ndim - 2
```
jni: ah, I see, displayed is displayed sliders, not displayed dimensions... I think it's worth making that explicit in the attribute name.

napari/components/dims.py
```diff
@@ -28,11 +29,14 @@ class Dims:
         slider when in INTERVAL mode, one for each dimension
     mode : list of DimsMode
         List of DimsMode, one for each dimension
-    display : list of bool
-        List of bool indicating if dimension displayed or not, one for each
-        dimension
+    order : tuple of int
+        Order in which dimensions are displayed where the last two or last
+        three dimensions correspond to Y x X or Z x Y x X if ndisplay is
+        two or three respectively.
```
jni: Maybe add a note here "or plane x row x column" for the scikit-image sticklers among us? =D

napari/components/dims.py
```diff
@@ -230,33 +228,27 @@ def set_mode(self, axis: int, mode: DimsMode):
...
-            self.set_display(-2, True)
+        axis_a = axis_a % self.ndim
+        axis_b = axis_b % self.ndim
+        index_a = self.order.index(axis_a)
+        index_b = self.order.index(axis_b)
+        self._order[index_a], self._order[index_b] = (
+            self.order[index_b],
+            self.order[index_a],
+        )
+        self.events.display()
```
jni: As mentioned above, I'm not convinced that a swap method is an elegant way to expose ordering.

napari/components/tests/test_viewer_model.py
```diff
@@ -136,3 +136,32 @@ def test_new_shapes():
...
+        viewer.layers['Labels']._data_view == labels_data[0, 0, :, :]
+    )
+
+    # Swap dims
+    viewer.dims.swap(1, 2)
+    assert viewer.dims.order == [0, 2, 1, 3]
+    assert np.all(viewer.layers['Image']._data_view == image_data[0, :, 0, :])
+    assert np.all(
+        viewer.layers['Labels']._data_view == labels_data[0, :, 0, :]
+    )
```
jni, line 167: Good stuff, love these tests.

napari/components/viewer_model.py
```diff
@@ -684,6 +698,8 @@ def _update_layers(self):
         """Updates the contained layers.
         """
         for layer in self.layers:
+            layer.ndisplay = self.dims.ndisplay
+            layer.dims_order = self.dims.order
```
jni, line 702: Again, the fact that these two properties always need to be updated in tandem suggests that there should be a different parameterisation of this property...

napari/layers/base/base.py
```diff
@@ -217,11 +249,32 @@ def _update_coordinates(self):
...
+        else:
+            position = [0] * len(self.displayed)
+
+        coords = list(self.indices)
+        for d, p in zip(self.displayed, position):
+            coords[d] = p
+        self.coordinates = tuple(coords)
+
+    def _reset_indices(self):
+        """Reset indicies, which is useful after data has been changed."""
```
jni: indices

napari/layers/image/image.py
```diff
@@ -340,6 +342,9 @@ def _update_thumbnail(self):
             colormapped = self.colormap[1].map(downsampled)
             colormapped = colormapped.reshape(downsampled.shape + (4,))
             colormapped[..., 3] *= self.opacity
+        if 0 in colormapped.shape:
+            colormapped = np.zeros(self._thumbnail_shape)
+            colormapped[..., 3] = self.opacity
```
jni: what's the story here? It's a strange check.

napari/layers/image/image.py
```diff
@@ -358,9 +363,14 @@ def get_value(self):
             shape = self._data_view.shape[:-1]
         else:
             shape = self._data_view.shape
-        coord[-2:] = np.clip(coord[-2:], 0, np.asarray(shape) - 1)
+        j = 0
+        slice_coord = []
+        for i in self.displayed:
+            coord[i] = np.clip(coord[i], 0, shape[j] - 1)
+            slice_coord.append(coord[i])
+            j += 1
```
jni: `for j, i in enumerate(self.displayed):`

jni: (or even swap the i and j)

napari/layers/labels/labels.py
```diff
@@ -439,21 +435,24 @@ def paint(self, coord, new_label):
...
+                    np.round(
+                        np.clip(
+                            coord[i] - self.brush_size / 2, 0, self.shape[i]
+                        )
+                    ).astype(int),
+                    np.round(
+                        np.clip(
+                            coord[i] + self.brush_size / 2, 0, self.shape[i]
+                        )
+                    ).astype(int),
```
jni, line 450: Probably worth writing a function, `coord2index`, that does clipping, rounding, and converting to int.

napari/layers/points/points.py
```diff
@@ -526,12 +522,15 @@ def _slice_data(self, indices):
...
+        indices = np.array(indices)
         if len(self.data) > 0:
             if self.n_dimensional is True and self.ndim > 2:
-                distances = abs(self.data[:, :-2] - indices[:-2])
-                sizes = self.sizes[:, :-2] / 2
+                distances = abs(self.data[:, not_disp] - indices[not_disp])
+                sizes = self.sizes[:, not_disp] / 2
                 matches = np.all(distances <= sizes, axis=1)
-                in_slice_data = self.data[matches, -2:]
+                in_slice_data = self.data[np.ix_(matches, disp)]
```
jni, line 533: Can you perhaps express this with numpy.meshgrid instead? (Don't forget indexing='ij'). I've never seen `numpy.ix_` in the wild which makes me worry about using it here. meshgrid at least has some indication of its uses in the name.

jni: (But, maybe don't stress too much about this. I just had to look it up., and it concisely expresses what you need. It could also be that numpy.ogrid does the same work, again being a bit more explicit in the name...)

napari/layers/pyramid/pyramid.py
```diff
@@ -170,44 +168,48 @@ def _slice_data(self):
...
         indices = list(self.indices)
-        # TODO: Change dims selection when dims model changes
-        rescale = self.level_downsamples[self.data_level, :-2]
-        indices[:-2] = np.round(indices[:-2] / rescale).astype(int)
-        indices[:-2] = np.clip(
-            indices[:-2],
-            0,
-            np.subtract(self.level_shapes[self.data_level][:-2], 1),
-        )
+        for i in self.not_displayed:
```
jni, line 180: I worry that the information in the dims object is getting replicated in lots of places, which increases the complexity of keeping them in sync. It might be worth considering adding a reference to the dims model in each layer.

napari/layers/pyramid/pyramid.py
```diff
@@ -253,19 +255,23 @@ def get_value(self):
...
 
-        # TODO: Change dims selection when dims model changes
-        coord[-2:] = np.clip(coord[-2:], 0, np.subtract(shape, 1))
-        value = self._data_view[tuple(coord[-2:])]
+        j = 0
+        slice_coord = []
+        for i in self.displayed:
+            coord[i] = np.clip(coord[i], 0, shape[j] - 1)
+            slice_coord.append(coord[i])
+            j += 1
```
jni: as above: use enumerate

jni (changes requested):
@sofroniewn I've left some comments and suggestions for improvements. You mentioned that shapes currently ignores this — what does that mean for the behavior of the shapes layer when dimensions are swapped?

jni:
Incidentally, just looked at the linux failures, and indeed it's nd-shapes that is failing! I don't think that's a coincidence. =)

sofroniewn:
@jni I've looked over your comments - I completely agree with your points with the examples - that approach of throwing everything in there is legacy from when we lacked good unit tests. I'll clean-up the examples and put the same stuff into specific tests - but I'll probably do that in a separate PR. For now I'll just remove the extra stuff so as to not make the problem worse.

The comment though that really resonated was 
> I worry that the information in the dims object is getting replicated in lots of places, which increases the complexity of keeping them in sync. It might be worth considering adding a reference to the dims model in each layer.

I completely agree - and as I think about the 3D stuff that @pranathivemuri is working on and still to come it really makes me want each layer to have its own independent dims model - and then a global dims model that is responsible for coordinating them all and the qt sliders. This will allow us to very cool visualizations like combining 2D slicing and 3D rendering. It will also make things in this PR substantially cleaner I think. What are your thoughts on this @jni, @royerloic? I'm happy to give it a go.

examples/add_image.py
sofroniewn: removing

examples/add_points.py
sofroniewn: removed

examples/nD_image.py
sofroniewn: yup - can drop `swap`

examples/nD_points.py
sofroniewn: removing

examples/nD_vectors_image.py
sofroniewn: removed

examples/nD_vectors_image.py

examples/swap_dims.py
sofroniewn: done

napari/_qt/qt_dims.py
sofroniewn: done

napari/_qt/qt_viewer.py
sofroniewn: yup, moved to model file

napari/components/dims.py
sofroniewn: dropped swap

napari/components/dims.py
sofroniewn: done

napari/components/tests/test_viewer_model.py
sofroniewn: thanks

napari/layers/base/base.py
sofroniewn: fixed

napari/layers/image/image.py
sofroniewn: yeah - this is out of place - the issue is that the downsampling can create an array of shape (0, 32, 4) if the aspect ratio is > 32 (which will be very rare), but which causes an error. Probably best to catch this in the thumbnail setter. I'll delete for now, but should probably make an issue so we don't forget

napari/layers/image/image.py
sofroniewn: done

napari/layers/pyramid/pyramid.py
sofroniewn: fixed

napari/_qt/qt_dims.py
sofroniewn: fixed

napari/_qt/qt_dims.py
sofroniewn: fixed by adding this property to the `Dims` model

napari/_qt/qt_dims.py
sofroniewn: yup - poor naming, adding a note and renaming

napari/_qt/tests/test_qt_dims.py
sofroniewn: fixed by changing name to `_displayed_sliders`

sofroniewn:
I have made some improvements as suggested by @jni - but I am going to try making a new PR building off this one where each layer gets it's own dims model and then they are appropriately synchronized to the global dims model / sliders. I think it will work well and future proof us as we support more complex slicing and 3D features

jni:
@sofroniewn perhaps a stepping stone to that approach would be to reference the global dims object first, get that done and working, and then you can worry about individual dims models with syncing. What do you think?

I also think it's worth throwing shapes into this PR, and higher priority than multi-dims. The simple thing to do would be to hide shapes that are not co-planar with the axes you're viewing, as you are doing now with broadcast mode. Viewing "flat" shapes as lines could be left for another time.

sofroniewn:
@jni agreed on supporting shapes as being a higher priority - i'll work on adding those and hide non-planar ones.

I was thinking about how to have to global dims as a stepping stone, but unfortunately it won't help because there is a conversion that happens from the global dims to all those layer dims variables (because the global dims can be 4 say, but a layer only 2) - the only way to cleanly get around this is to allow each layer to have it's own dims and then synchronize them in the viewer model. I'm working on this bit now and I think it's going ok

napari/_qt/qt_dims.py
```diff
@@ -42,7 +42,7 @@ def __init__(self, dims: Dims, parent=None):
 
         # list of sliders
         self.sliders = []
-        self._displayed = []
+        self._displayed_sliders = []
```
kne42, line 45: 👌

napari/components/dims.py
```diff
@@ -95,16 +98,22 @@ def interval(self):
...
         return copy(self._mode)
 
     @property
-    def display(self):
-        """list: List of bool indicating if dimension displayed or not, one for
-        each dimension
-        """
-        return copy(self._display)
+    def order(self):
+        """list: List of int indicating display order of dimensions."""
```
kne42, line 106:
```suggestion
        """list of int: Display order of dimensions."""
```

napari/components/dims.py
```diff
@@ -145,7 +154,9 @@ def ndim(self, ndim):
             self._point = self._point[-ndim:]
             self._interval = self._interval[-ndim:]
             self._mode = self._mode[-ndim:]
-            self._display = self._display[-ndim:]
+            order = np.array(self._order[-ndim:])
+            order[np.argsort(order)] = list(range(len(order)))
```
kne42, line 158: whoa what does this do?

napari/components/dims.py
```diff
@@ -154,24 +165,37 @@ def ndim(self, ndim):
...
+        for axis in range(self.ndim):
+            if axis in self.displayed:
+                slice_list.append(slice(None))
+            else:
+                slice_list.append(int(round(self.point[axis])))
         return tuple(slice_list)
 
+    @property
+    def ndisplay(self):
+        """Int: Number of displayed dimensions."""
```
kne42, line 177:
```suggestion
        """int: Number of displayed dimensions."""
```

napari/components/dims.py
```diff
@@ -154,24 +165,37 @@ def ndim(self, ndim):
...
+        if ndisplay not in (2, 3):
+            raise ValueError(
+                f"Invalid number of dimensions to be displayed {ndisplay}"
+            )
+        self._ndisplay = ndisplay
+        self.events.display()
+
+    @property
+    def displayed(self):
+        """tuple: list of displayed dimensions."""
```
kne42, line 191:
```suggestion
        """tuple: Displayed dimensions."""
```

napari/components/dims.py
```diff
@@ -154,24 +165,37 @@ def ndim(self, ndim):
...
+        return tuple(self.order[-self.ndisplay :])
+
+    @property
+    def not_displayed(self):
+        """tuple: list of dimensions not displayed."""
```
kne42, line 196:
```suggestion
        """tuple: Dimensions that are not displayed."""
```

napari/components/dims.py
```diff
@@ -182,8 +206,7 @@ def set_range(self, axis: int, range: Sequence[Union[int, float]]):
         range : tuple
             Range specified as (min, max, step)
         """
-        if axis < 0:
-            axis += self.ndim
+        axis = axis % self.ndim
```
kne42, line 209: do not use modulus here (and everywhere else) for negative indexing. instead do what you had before and raise an exception if it's still negative afterwards

```python
In [1]: l = [0, 1]

In [2]: l[-3]
IndexError                                Traceback (most recent call last)
<ipython-input-2-55afb2c9d60b> in <module>
----> 1 l[-3]

IndexError: list index out of range
```

napari/layers/base/base.py
```diff
@@ -193,12 +199,38 @@ def indices(self):
...
     def indices(self, indices):
-        if indices == self.indices:
-            return
         self._indices = indices[-self.ndim :]
         self._update_coordinates()
         self._set_view_slice()
 
+    @property
+    def displayed(self):
+        """Tuple: Displayed dimensions."""
```
kne42, line 208:
```suggestion
        """tuple: Displayed dimensions."""
```

kne42: (and below)

napari/components/dims.py
```diff
@@ -154,24 +165,37 @@ def ndim(self, ndim):
...
+
+    @property
+    def displayed(self):
+        """tuple: list of displayed dimensions."""
+        return tuple(self.order[-self.ndisplay :])
+
+    @property
+    def not_displayed(self):
+        """tuple: list of dimensions not displayed."""
+        return tuple(self.order[: -self.ndisplay])
```
kne42, line 197: also since this does not really need to be calculated on the fly you should set it whenever you update `self.ndisplay` and `self.order`

royerloic:
> @jni I've looked over your comments - I completely agree with your points with the examples - that approach of throwing everything in there is legacy from when we lacked good unit tests. I'll clean-up the examples and put the same stuff into specific tests - but I'll probably do that in a separate PR. For now I'll just remove the extra stuff so as to not make the problem worse.
> 
> The comment though that really resonated was
> 
> > I worry that the information in the dims object is getting replicated in lots of places, which increases the complexity of keeping them in sync. It might be worth considering adding a reference to the dims model in each layer.
> 
> I completely agree - and as I think about the 3D stuff that @pranathivemuri is working on and still to come it really makes me want each layer to have its own independent dims model - and then a global dims model that is responsible for coordinating them all and the qt sliders. This will allow us to very cool visualizations like combining 2D slicing and 3D rendering. It will also make things in this PR substantially cleaner I think. What are your thoughts on this @jni, @royerloic? I'm happy to give it a go.

I don't think we need a full dims model per layer. This would unnecessarily blow up the complexity and change the global nature of the dims model. What makes more sense is that each layer has additional information that is specific to that layer. For example in the case of 3D viewing and showing a 2D slice, you simply have the global dims layer plus the two axis {0,1} that are sliced. The dims model 'point' has already enough information to know which slice to show. Each layer has flexibility in terms of how it uses the global dims information + layer specific info for selecting and rendering. the global dims model has a lot of info already, some of which is not always used (think point when 3d rendering) we should be carefull not to complexify without having thought hard about how we can do it with the current infrastructure.

We can talk about it...

sofroniewn:
@royerloic - I agree we want to avoid complexity and unnecessarily using classes, but I think re-using the dims model is warranted in this case.

The implementation currently on this PR is as you described: 

> global dims information + layer specific info for selecting and rendering

but that lead to at least 7 layer specific dims properties that needed to be added to the layer, for example `layer.dims_order`, `layer.displayed`,  `layer.not_displayed`,  `layer.ndisplay`, `layer.range`, `layer.indices`, `layer.ndim`. ALL of these properties already exist inside the dims model, and in the implementation in #451 we just have `layer.dims` which provides these properties provided, i.e. `layer.dims.order`, `layer.dims.displayed`, `layer.dims.ndisplay`, `layer.dims.range`, etc. To me this is much cleaner and it is much more clear how to synchronize the global dims model with these layer dims models than keeping track of and synchronizing all those properties individually.

At this point we've got both implemented approaches implemented for the `Image` layer on #450 and #451, so we can just look at them and compare (note the changes in the `Base` layer files too). I'll also note that we were able to drop a lot of `np.clip` commands in #451 too because now the layer dims model exactly matches that layer.

@jni what are your thoughts?

royerloic:
Let's chat about it on Thu, I need to better understand your concerns, and maybe better communicate what i have in mind -- assuming that I have a complete understanding of the problem. Quite an interesting topic...

jni:
I agree with @royerloic, this is probably a discussion best had in a high-bandwidth setting.

napari/components/dims.py
sofroniewn: The modulus will force everything to be positive though, why is this undesirable? In the example above you'd have
```python
l[-3 % 2]
```
which is equivalent to
```python
l[1]
```
I'm happy to change though as if someone passes `-17` and the list is of length 3 the probably deserve an error ....

kne42: my point is that it is standard practice to only perform this wrap once instead of forcing it to be positive always

sofroniewn: got it - will fix

sofroniewn:
This is being closed in favour of #451 as discussed in the team meeting, where we concluded that it was simpler for each layer to have its own dims model rather than just forward on specific properties like is being done in this PR

PR #451, opened 2019-07-29
Swappable dimensions

sofroniewn:
# Description
This PR superseeds #450 by attaching a `Dims` model to each layer and then synchronizing them with a global `Dims` model in the `ViewerModel`. This will result in less duplicate code and a much easier time keeping track of all our dims variables. It will ultimately enable much more complex features like combining 2D and 3D layers, which the vispy scene graph supports - but that is not planned for this PR.

Right now support has been extended to `Image` and `Labels` layers. It will be quick to add support for `Points`, `Pyramid`, `Volume`, and `Vectors`. This PR will also enable `Volume` to be merged with `Image` in the not too distant future.

It will be a little more effort to support `Shapes` but it is probably worth doing before merging as otherwise we'll have only partially implemented functionality.

## Type of change

# How has this been tested?

- [x] `examples/swap_dims.py`

## Final checklist:


kne42:
could you elaborate on how this makes things simpler compared to #450 and #440? it's not immediately obvious to me

sidenote: i feel like we're really bloating the layers models nowadays. @shanaxel42 do you have any ideas on how to use the MVC paradigm to separate all of the gunk/internal machinery so that users don't have to deal with it when they access layers in the notebook/scripts?

sofroniewn:
@kne42 - the reason I find this cleaner is that now there is an exact correspondence between the layer and the dims information - including number of dimensions, allowed range of the dimensions, etc. - before we had the case where the overall dims model might have a large number of dims or larger allowed ranges than the layer could accommodate and so we needed to do things like shorten the indices and clip their values. This removes a lot of those clip statements.

The other reason I like it is that it will support much more functionality that is in our `0.2` roadmap including more flexible `3D` rendering and mixing `2D` and `3D` using the vispy scene graph.


As to the layer model files getting polluted - I completely agree. Separating out the MVC of our layer models is really one of the last `0.1` features we still need to do and then we can declare `0.1` really closed. @shanaxel42 any advise on how to best do this would be appreciated. We should get to this after this PR lands.

sofroniewn:
I've now updated this PR with the new tests / console features and made sure all the tests continue to pass. I will implement the swappable dims feature for the shapes layer and then this will be ready for review

sofroniewn:
@kne42, @jni - this PR is now ready for review

It provides full support for swappable dimensions for all layer types, enabling behviour like in the gif below.

flip_dims

The most serious changes happened to the shapes layer were we moved away from the dictionary of 2D shapes approach to full nD shapes.

Also each layer now has it's own dims model that gets synchronized with the global dims model.

royerloic:
Very cool!

examples/nD_shapes.py
```diff
@@ -6,6 +6,8 @@
 from skimage import data
 import napari
 
+np.random.seed(0)
```
jni: Personally I would move this to right before the call to `np.random.uniform`, as it seems forlorn and out of place here.

examples/nD_shapes.py
```diff
@@ -43,8 +45,8 @@
         corners, shape_type='rectangle', name='broadcasted'
     )
 
-    masks = layer.to_masks(mask_shape=(128, 128, 128))
-    labels = layer.to_labels(labels_shape=(128, 128, 128))
+    masks = layer.to_masks(mask_shape=(128, 128))
+    labels = layer.to_labels(labels_shape=(128, 128))
```
jni, line 48: I'm confused... What's happened here?

examples/nD_vectors_image.py
```diff
@@ -24,13 +24,8 @@
...
     pos = np.zeros(shape=(m, n, p, 3), dtype=np.float32)
-    rand1 = 2 * (np.random.random_sample(n * m * p) - 0.5)
-    rand2 = 2 * (np.random.random_sample(n * m * p) - 0.5)
-
-    # assign projections for each vector
-    pos[:, :, :, 0] = 0
-    pos[:, :, :, 1] = rand1.reshape((m, n, p))
-    pos[:, :, :, 2] = rand2.reshape((m, n, p))
+    for i in range(3):
+        pos[..., i] = 2 * np.random.random_sample((m, n, p)) - 0.5
```
jni: This was too complicated, and the suggested change simplifies it but not enough imho. I suggest:

```
pos = np.random.uniform(-1, 1, size=(m, n, p, 3))
```

and be done with it.

(Your current code is incorrect due to missing parentheses btw.)

examples/swap_dims.py
```diff
@@ -0,0 +1,36 @@
...
+        [
+            [0, 0, 0, 100],
+            [0, 0, 50, 120],
+            [1, 0, 100, 40],
+            [2, 10, 110, 100],
+            [9, 8, 80, 100],
+        ]
+    )
+    viewer.add_points(
+        points, size=[0, 6, 10, 10], face_color='blue', n_dimensional=True
```
jni, line 33: Ah, so this was confusing to me for a second, as I thought this was different sizes for different points, and that you'd given the wrong-size array. Perhaps we should do like sklearn features here, meaning the input *has* to be 2D and the sizes are broadcast against the points coordinates. So you would need to pass a *column* 2D vector if you want different sizes for each dot but with the same size across all dimensions. (And here you would need to pass `size=[[0, 6, 10, 10]]`.) What do you think?

napari/_qt/qt_viewer.py
```diff
@@ -244,7 +240,13 @@ def _on_cursor(self, event):
...
 
     def _on_reset_view(self, event):
-        self.view.camera.rect = event.viewbox
+        if isinstance(self.view.camera, PanZoomCamera):
+            self.view.camera.rect = event.viewbox
+        elif isinstance(self.view.camera, ArcballCamera):
+            self.view.camera.center = event.center
+            self.view.camera.scale_factor = event.scale_factor
+        else:
+            raise ValueError(f"Invalid camera type {type(self.view.camera)}")
```
jni: I find this code pretty weird... Don't we have full control of the camera settings? And, if an advanced user goes in and changes the camera, don't we want to allow them to do what they want?

imho this should maybe warn ("we don't know how to reset this camera type"), but not raise.

jni: An indication that this shouldn't be here is that there's no test using `pytest.assert_raises` for this.

napari/components/dims.py
```diff
@@ -28,14 +29,23 @@ class Dims:
         slider when in INTERVAL mode, one for each dimension
     mode : list of DimsMode
         List of DimsMode, one for each dimension
-    display : list of bool
-        List of bool indicating if dimension displayed or not, one for each
-        dimension
+    order : tuple of int
+        Order in which dimensions are displayed where the last two or last
+        three dimensions correspond to row x column or plane x row x column if
+        ndisplay is two or three respectively.
```
jni: 2 or 3.

napari/components/dims.py
```diff
@@ -145,33 +156,61 @@ def ndim(self, ndim):
...
+            order[np.argsort(order)] = list(range(len(order)))
+            self._order = list(order)
 
             # Notify listeners that the number of dimensions have changed
             self.events.ndim()
 
     @property
     def indices(self):
-        """Tuple of slice objects for slicing arrays on each dimension."""
+        """tuple of slice objects for slicing arrays on each dimension."""
```
jni: But... Why?

napari/components/dims.py
```diff
@@ -145,33 +156,61 @@ def ndim(self, ndim):
...
+            if axis in self.displayed:
+                slice_list.append(slice(None))
+            else:
+                p = np.clip(
+                    self.point[axis],
+                    np.round(self.range[axis][0]),
+                    np.round(self.range[axis][1]) - 1,
+                )
+                p = np.round(p).astype(int)
+                slice_list.append(p)
```
jni, line 188: Do we not want to keep the dims mode logic around for the future? I'm also concerned that the clipping is happening now but wasn't before. Clipping is a great idea when it *intentionally* makes the code cleaner, but in other cases it can actually mask logic bugs, where out-of-range indices are getting propagated that shouldn't be.

napari/components/dims.py
```diff
@@ -182,8 +221,7 @@ def set_range(self, axis: int, range: Sequence[Union[int, float]]):
         range : tuple
             Range specified as (min, max, step)
         """
-        if axis < 0:
-            axis += self.ndim
+        axis = axis % self.ndim
```
jni: As above: I'm worried that this will mask bugs. if someone tries to set_range on axis 5 in a 3D dims object, that should be an error, rather than silently set the range on axis 2...

napari/components/viewer_model.py
```diff
@@ -683,11 +693,32 @@ def _new_labels(self):
...
+            # respectively
+            offset = self.dims.ndim - layer.dims.ndim
+            order = np.array(self.dims.order)
+            layer.dims.order = list(order[order >= offset] - offset)
+
+            # Update the point values of the layers for the dimensions that
+            # the layer has
+            for axis in range(layer.dims.ndim):
+                point = self.dims.point[axis + offset]
+                layer.dims.set_point(axis, point)
```
jni, line 859: This is a bit hairy, but I don't quickly see a way to improve it. Thanks very much for the comment above, which clarifies things immensely. =) I just want to make sure: are these lines of code explicitly tested?

napari/layers/base/base.py
```diff
@@ -103,7 +99,7 @@ class Layer(VisualWrapper, KeymapMixin, ABC):
     Notes
     -----
     Must define the following:
-        * `_get_shape()`: called by `shape` property
+        * `_get_range()`: called by `range` property
```
jni, line 102: I'm a little confused by this name change?

napari/layers/base/base.py
```diff
@@ -130,16 +126,17 @@ def __init__(
...
         self._freeze = False
         self._status = 'Ready'
         self._help = ''
         self._cursor = 'standard'
         self._cursor_size = None
         self._interactive = True
-        self._indices = (slice(None, None, None), slice(None, None, None))
         self._position = (0, 0)
         self.coordinates = (0, 0)
+        self.dims = Dims(2)
```
jni, line 138: I like this. =)

napari/layers/base/base.py
```diff
@@ -255,19 +256,14 @@ def thumbnail(self, thumbnail):
...
-
-    @property
-    def range(self):
-        """list of 3-tuple of int: ranges of data for slicing specifed by
-        (min, max, step).
-        """
-        return tuple((0, max, 1) for max in self.shape)
+        return tuple(
+            np.round(r[1] - r[0]).astype(int) for r in self.dims.range
+        )
```
jni, line 265: Ok maybe this answers my question above — this allows for ranges that don't have [0,] * ndim at the origin, correct?

napari/layers/image/image.py
```diff
@@ -275,11 +273,14 @@ def interpolation(self, interpolation):
 
     def _set_view_slice(self):
         """Set the view given the indices to slice with."""
-        indices = list(self.indices)
-        indices[:-2] = np.clip(
-            indices[:-2], 0, np.subtract(self.shape[:-2], 1)
+        if self.multichannel:
+            order = self.dims.displayed_order + (self.dims.ndisplay,)
```
jni, line 280: I'm confused about this...? What does `ndisplay` have to do with the number of channels?

napari/layers/labels/labels.py
```diff
@@ -384,7 +378,7 @@ def fill(self, coord, old_label, new_label):
         else:
             # work with just the sliced image
             labels = self._data_view
-            slice_coord = tuple(int_coord[-2:])
+            slice_coord = tuple(int_coord[d] for d in self.dims.displayed)
```
jni, line 381: So nice. =)

napari/layers/points/points.py
```diff
@@ -524,12 +513,15 @@ def _slice_data(self, indices):
...
+        indices = np.array(indices)
         if len(self.data) > 0:
             if self.n_dimensional is True and self.ndim > 2:
-                distances = abs(self.data[:, :-2] - indices[:-2])
-                sizes = self.sizes[:, :-2] / 2
+                distances = abs(self.data[:, not_disp] - indices[not_disp])
+                sizes = self.sizes[:, not_disp] / 2
                 matches = np.all(distances <= sizes, axis=1)
-                in_slice_data = self.data[matches, -2:]
+                in_slice_data = self.data[np.ix_(matches, disp)]
```
jni, line 525: Lovely! This generalised quite nicely to the new model!

napari/layers/pyramid/pyramid.py
```diff
@@ -166,46 +174,47 @@ def top_left(self, top_left):
...
-        )
-        if np.any(
-            self.level_shapes[self.data_level][-2:] > self._max_tile_shape
-        ):
-            slices = [
-                slice(
+        nd = self.dims.not_displayed
+
+        if self.multichannel:
+            order = self.dims.displayed_order + (self.dims.ndisplay,)
```
jni, line 183: As above: I find this quite confusing!

jni: Side question: I'm kinda thinking of deprecating multichannel data (which anyway is poorly named in this case, it's really rgb), and instead having convenience functions to add channels as separate layers in additive mode. This would simplify the internal data model. What do you think? If you agree (at least enough to have a dedicated discussion) I'll open an issue to keep track of this.

examples/nD_shapes.py
sofroniewn: this is actually fixing something that was inappropriate - this layer was actually 2D and so it should not have been possible to broadcast the shape into a 3D mask. note that the other layer was actually 3D and it is still possible to get out a 3D mask.

examples/nD_shapes.py
sofroniewn: moved

napari/layers/pyramid/pyramid.py
```diff
@@ -166,46 +174,47 @@ def top_left(self, top_left):
...
+        else:
+            order = self.dims.displayed_order
+
+        # Slice thumbnail
+        indices = np.array(self.dims.indices)
+        downsampled = indices[nd] / self.level_downsamples[-1, nd]
+        downsampled = np.rint(downsampled.astype(float)).astype(int)
```
jni: Why `rint` instead of `round`? I've never heard of `rint` before!

napari/layers/shapes/shape_list.py
```diff
@@ -17,6 +17,12 @@ class ShapeList:
     ----------
     shapes : (N, ) list
         Shape objects.
+    data : (N, ) list of (M, D) array
+        Data arrays for each shape.
+    slice_keys : (N, 2, P) array
+        Array of slice keys for each shape. Each slice key has the min and max
+        values of the P non-displayed dimensions, useful for slicing
+        multidimensional shapes.
```
jni: I'm a little confused about this. Could you maybe add an example usage of this in the docstring?

napari/layers/shapes/shape_list.py
```diff
@@ -90,6 +110,45 @@ def z_indices(self):
...
+        if not np.all(self._slice_key == slice_key):
+            self._slice_key = slice_key
+            self._update_displayed()
+
+    def _update_displayed(self):
+        """Update the displayed data based on the slice key."""
+        slice_key = np.array([self.slice_key, self.slice_key])
+
+        # Slice key must exactly match mins and maxs of shape
+        self._displayed = np.all(self.slice_keys == slice_key, axis=(1, 2))
```
jni: Yeah this is super confusing...

napari/layers/shapes/shape_list.py
```diff
@@ -555,7 +635,8 @@ def outline(self, indices):
         return centers, offsets, triangles
 
     def shapes_in_box(self, corners):
-        """Determines which shapes, if any, are inside an axis aligned box
+        """Determines which shapes, if any, are inside an axis aligned box.
+        Looks only at displayed shapes
```
jni, line 647:
```suggestion

        Looks only at displayed shapes
```

jni (commented):
@sofroniewn sorry, I'm only about halfway through the review, but since it's bedtime here (European time for the next 4 weeks), I thought I'd get my comments to you before resuming tomorrow. Goodnight!

jni:
At any rate, great PR, I think most of my comments are very minor!

Note that CI is failing quite reliably so we probably need that investigated before merging!

examples/nD_vectors_image.py

examples/swap_dims.py
sofroniewn: that's fine, but that should be addressed in a different PR, can you make an issue?

napari/_qt/qt_viewer.py
sofroniewn: we do have full control, it was more that if in the future someone adds a new camera type they'll have to adjust this part appropriately as I can't assume it should be one way or the other. is it fine to leave?

napari/components/dims.py
sofroniewn: fixed

napari/components/dims.py
sofroniewn: everything else is lowercase and tuple is lowercase in other places? i'm fine with uppercase but we should go through and systematically make everything the same rather than having a mix

napari/components/dims.py
sofroniewn: so we actually want the clipping only to happen sometimes, and it depends on what type of layer it is. i've had to actually add a clip / not clip flag, is that ok?

As to the dims mode stuff, it's so easy to add back in and was just strange to have when not implemented. i'd rather leave it out until we get there

napari/components/dims.py
sofroniewn: fixed - as requested by @kne42 too

napari/layers/base/base.py
sofroniewn: we're now getting the range on each layer - which has min / max instead of just the shape. this was needed as layers like `Points` and `Shapes` need to be described by their range

napari/layers/base/base.py
sofroniewn: it's mostly for the `Points` / `Shapes` layers, but yes

napari/layers/image/image.py
sofroniewn: i've added a comment to explain - it's really about keeping the final axis fixed if the data is multichannel, and the final axis is `2` if the display data is 2D and `3` if the display data is 3D, though we can't actually do 3D displays with rgb right now, and the image layer only supports ndisplay=2 right now, so I could just hard code this to be `2`, but this seemed a little more flexible

napari/layers/pyramid/pyramid.py
sofroniewn: yes - I'd prefer that, getting rid of multichannel would be great - will also make compatibility of the image layer with the volume layer much easier as the volume layer cannot be rgb. would also fix some clim range slider issues, so i'm all around in favor.

if someone passes rgba though hard to get pixelwise opacity values without `multichannel` though :-( to be discussed ....

napari/layers/pyramid/pyramid.py
sofroniewn: no reason, changing

napari/layers/shapes/shape_list.py
sofroniewn: have added some to the doc string

napari/layers/shapes/shape_list.py
sofroniewn: have added some comments

napari/layers/shapes/shape_list.py
sofroniewn: fixed

napari/components/viewer_model.py
sofroniewn: somewhat - i've added an extra test that at least mixes layers of different dimensionality. maybe could be improved

sofroniewn:
@jni @kne42 I've made updates based on comments from the first half of review from @jni - this is ready for more review

napari/components/dims.py
jni: The first letter of the docstring should be uppercase. Everywhere in the Python ecosystem, docstrings are capitalised as in normal English prose.

napari/_qt/qt_viewer.py
jni: imho we shouldn't have code paths that we can't test. If you can somehow trigger this line in a test, I'm ok with leaving it. Otherwise I think it should be removed, and maybe replaced with a comment about how to update it for new camera types.

napari/components/dims.py
jni: I like a new flag, as it makes things explicit.

napari/components/viewer_model.py
jni: You should be able to check with `coverage.py` that this line is covered by tests. Docs here:

https://coverage.readthedocs.io/en/v4.5.x/

examples/swap_dims.py
sofroniewn: thanks for making the issue

napari/_qt/qt_viewer.py
sofroniewn: ok removed and added a comment

napari/components/dims.py
sofroniewn: ok fixed in this file, but there will be other places we need to fix at another time. I've talked with @kne42 about her doing a pass through all our doc strings sometime soon

napari/layers/pyramid/pyramid.py
sofroniewn: thanks for making an issue #481

napari/components/viewer_model.py
sofroniewn: the lines get run, including in cases with layers of different dimensions, it's a little harder to tell that they are doing the right thing, i.e. there isn't an `assert` that directly corresponds to what their output should be

napari/layers/pyramid/pyramid.py
```diff
@@ -166,46 +174,50 @@ def top_left(self, top_left):
...
+        indices[nd] = downsampled
+        self._data_thumbnail = np.asarray(
+            self.data[-1][tuple(indices)]
+        ).transpose(order)
+
+        # Slice currently viewed level
+        indices = np.array(self.dims.indices)
+        level = self.data_level
+        downsampled = indices[nd] / self.level_downsamples[level, nd]
+        downsampled = np.rint(downsampled.astype(float)).astype(int)
```
jni:
```suggestion
        downsampled = np.round(downsampled.astype(float)).astype(int)
```

napari/layers/pyramid/pyramid.py
jni: Since we seem to be happy about deprecating multichannel, this might not be relevant, but typically in this situation I try to use a meaningful variable name to make things easier to digest: `channel_axis = self.dims.ndisplay  # 2 if 2D, 3 if 3D; order = self.dims.displayed_order + (channel_axis,)`

napari/layers/shapes/shape_list.py
```diff
@@ -17,6 +17,14 @@ class ShapeList:
     ----------
     shapes : (N, ) list
         Shape objects.
+    data : (N, ) list of (M, D) array
+        Data arrays for each shape.
+    slice_keys : (N, 2, P) array
+        Array of slice keys for each shape. Each slice key has the min and max
+        values of the P non-displayed dimensions, useful for slicing
+        multidimensional shapes. If the both min and max values of shape are
+        equal than the shape is entirely contained within the slice specified
```
jni:
```suggestion
        equal then the shape is entirely contained within the slice specified
```

napari/layers/shapes/shape_list.py
```diff
@@ -17,6 +17,14 @@ class ShapeList:
...
+        by those values.
```
jni, line 27: Does this mean that you can define prism shapes now?

jni: if the min and max values are not the same?

napari/layers/shapes/shape_list.py
```diff
@@ -90,6 +112,48 @@ def z_indices(self):
...
+    @slice_key.setter
+    def slice_key(self, slice_key):
+        slice_key = list(slice_key)
+        if not np.all(self._slice_key == slice_key):
+            self._slice_key = slice_key
+            self._update_displayed()
+
+    def _update_displayed(self):
+        """Update the displayed data based on the slice key."""
+        slice_key = np.array([self.slice_key, self.slice_key])
```
jni, line 131: I don't understand the repetition here...

napari/layers/shapes/shape_models/rectangle.py
```diff
@@ -1,17 +1,17 @@
...
     """Class for a single rectangle
 
     Parameters
     ----------
-    data : (4, 2) or (2, 2) array
-        Either a (2, 2) array specifying the two corners of an axis aligned
-        rectangle, or a (4, 2) array specifying the four corners of a bounding
+    data : (4, D) or (2, D) array
+        Either a (2, D) array specifying the two corners of an axis aligned
+        rectangle, or a (4, D) array specifying the four corners of a bounding
```
jni, line 14: Is an nD axis aligned rectangle uniquely defined by two points? I think actually not?

jni (approved):
@sofroniewn I haven't tested but the rest of the code LGTM! As before, I suggest you get to the bottom of the failing CI before merging. A lot of this functionality is pretty crazy (I mean that in a good way!), so the more can be tested, the better!

napari/layers/pyramid/pyramid.py
sofroniewn: fixed

napari/layers/shapes/shape_list.py
sofroniewn: fixed

napari/layers/shapes/shape_list.py
sofroniewn: it is to check against both the min and max value, i'll add a comment - maybe there's a better way to do this that we can implement in the future

napari/layers/shapes/shape_models/rectangle.py
sofroniewn: you are correct it is weird, and the code actually only supports the two point version for 2D shapes, I have updated the doc string to reflect this

sofroniewn:
critical tests are passing, i'm going to merge now - thanks for your review @jni

PR #452, opened 2019-07-29
use __init_subclass__ in keymap mixin to create empty class keymap

kne42:
# Description
automatically create an empty class keymap in all classes that inherit from `KeymapMixin` if it is not defined within that class

# References
see https://github.com/napari/napari/pull/389#discussion_r300839195
[PEP 487](https://www.python.org/dev/peps/pep-0487/) (new in Python 3.6)

### DIFF
--- a/napari/layers/image/image.py
+++ b/napari/layers/image/image.py
@@ -99,8 +99,6 @@ class Image(Layer):
 
     _colormaps = AVAILABLE_COLORMAPS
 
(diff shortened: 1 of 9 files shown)


sofroniewn (approved):
Looks good to me, thanks!

jni (approved):
So much :heart: :heart: :heart:

PR #453, opened 2019-07-30
use pytest-qt

kne42:
# Description
use `pytest-qt` for our tests and stop testing examples


sofroniewn (approved):
This looks good to me and will hopefully solve some of the file opening / memory issues we've been having with our CI.

It means we're no longer testing the examples folder - but as discussed with @kne42 - I think we should aim to have comprehensive coverage of our tests so that we know the examples will run instead of relying on the examples folder.

This will also allow us to make the examples more useful to new-comers rather than be about testing, which I think @jni will appreciate.

Thoughts @jni?

napari/_qt/tests/test_qt_dims.py
```diff
@@ -1,140 +1,138 @@
-import os
 import numpy as np
-from napari.components import Dims
-from napari._qt.qt_dims import QtDims
-from napari import gui_qt
 
-os.environ['NAPARI_TEST'] = '1'
+from ...components import Dims
+from ..qt_dims import QtDims
```
jni: Typically we use absolute imports in tests. This allows testing of installed packages e.g. if we ever decide not to ship the testing code with the application.

jni:
I'm kinda -1 or removing the examples now. Perhaps we should wait until we actually have that full coverage?

Also, as mentioned in the comment, tests should use absolute imports. (I'm not 100% sure about that being the best approach, but it's certainly the one used in scikit-image, SciPy, and dask, so I take it to be a good model.)

sofroniewn:
If we get this merged I can immediately make a follow up PR porting over functionality from the examples to comprehensive tests. 

The reason to do this now is that all our examples are causing our CI to break on #443 even though all the tests pass locally and i’d like to get that one merged as soon as possible

jni:
There's no reason to merge then fix. You can fetch @kne42's branch and push directly back to it, updating the PR. At any rate, the PR that moves the tests should be the PR that disables the example-based testing, not this one.

jni:
(And there is still the issue of relative vs absolute imports)

sofroniewn:
I'm happy to add the example -> test migration into this PR. I guess it sort of needs to be the same one otherwise I won't be able to do it using the nice new framework and I'll just have to update everything after this PR gets in.

I'll let @kne42 comment on the absolute / relative paths and then I'll get to work on it

kne42:
@jni yeah i forgot about that. i also discovered that installing `pytest-qt` will break our current examples tests anyways

sofroniewn:
@kne42 @jni I have updated this PR to port over a lot of the testing that was done in the examples into our proper unit tests, including adding all layer types to the viewer, adding nD layers to the viewer, taking screenshots, generating svg etc.

I've removed some of the examples that were really just functioning as tests, but have erred on the side of keeping examples if they might be useful.

This PR is now ready for re-review

.cirrus.yml
```diff
@@ -30,15 +30,17 @@ task:
         DISPLAY: ":99"
       system_script:
         - apt-get update
-        - apt-get install -y libgl1-mesa-glx xvfb libqt5x11extras5
+        - apt-get install -y libgl1-mesa-glx xvfb libqt5x11extras5 herbstluftwm
```
jni, line 33: This looks like such a typo. :joy:

napari/_qt/tests/test_advanced.py
```diff
@@ -0,0 +1,139 @@
...
+from napari.components import ViewerModel
+from napari._qt.qt_viewer import QtViewer
+
+
+def test_4D_5D_images(qtbot):
+    """Test adding 4D followed by 5D image layers to the viewer.
+
+    Intially only 2 sliders should be present, then a third slider should be
+    created.
+    ."""
```
jni:
```suggestion
    """
```

napari/_qt/tests/test_advanced.py
```diff
@@ -0,0 +1,139 @@
...
+    np.random.seed(0)
+    viewer = ViewerModel()
+    view = QtViewer(viewer)
+    qtbot.addWidget(view)
```
jni: If I understand correctly, you're using two steps here because you need the `view` for the qtbot. However, I wonder if it's better to do

```python
viewer = napari.Viewer()
qtbot.addWidget(viewer.window.qt_viewer)
```

because that is the type of instantiation that we want to test... I'm :+0.5: on that so happy to hear some pushback, just a suggestion.

jni (approved):
This is a wonderful PR! Thank you @kne42 and @sofroniewn! I've left an optional comment to address and also a tiny docstring fix, but I'd be perfectly happy if it got merged with just the fix.

napari/_qt/tests/test_advanced.py
sofroniewn: fixed

napari/_qt/tests/test_advanced.py
sofroniewn: yup - agreed, i'll switch this over for the `advanced` tests. It also makes it more clear these are testing the whole app as opposed to just the qt viewer and I might move and rename accordingly

sofroniewn:
I've added even more tests to explicitly test the `Viewer` too. I'll note that there's a funny pattern we have that results in a lot of duplication right now where we say test the `add_image` function on both the `ViewerModel`, the `QtViewer` and the `Viewer`. I think it's important to test all three as there are real reasons why one could fail and not the other, but we might be able to streamline this process. I leave that to future PRs though - @kne42 might be interested in thinking about that.

All the needed tests are passing now so I will merge this.

jni:
:+1: so great. =)

Issue #301, opened 2019-05-27
Features to 0.2 release


sofroniewn:
## 🚀 Feature
This issue is to discuss features for our `0.2` release, aimed for Oct 1 2019 and the Bioimage Informatics meeting in Seattle at the Allen Institute.

At a high level this release marks the addition of new features that will enable a large number of workflows within napari, including support for rapidly switching the displayed dimensions and projection type, support for 3D rendering, and an integrated console. We will also aim to have usable public examples, docs, and tutorials by this release.

**Dims features**:
- [x] Make which dims displayed changeable so can rapidly switch dims

**3D rendering**:
- [x] 3D rendering of Images, Vectors, Points, Labels layers
- [x] 3D rendering of Shapes layer

**Layers**:
- [x] Allow `viewer.add_image` to expand an array along a designated axis to add one layer per axis with a different colormap - `add_multichannel`
- [x] Merge pyramid layer into image layer
- [x] Support pyramid layer for labels (viewing at least)

**QSS, stylings**:
- [x] Unify layer properties / controls
- [x] fixes from design audit
- [x] GUI button to toggle 2D / 3D rendering

**interactivity**:
- [x] support custom mouse functions
- [x] basic undo / redo for labels
- [x] disable editing when in 3D rendering
- [x] support for interactive scripting

**File I/O**:
- [x] Loading basic image file types (i.e. using `imageio`)
- [x] Saving basic image file types (i.e. using `imageio`)
- [x] Loading labels layers
- [x] Saving labels layers
- [x] Saving screenshots
- [x] Loading / saving svg

**Basic grid mode**
- [x] Expose a to_grid / to_stack on the layers list which will show all the layer in a grid or in 		overlay. Can also take a shape = [N, M], positions = [0, 1, … ] that will place images in a location in a grid.

**Console**
- [x] A console with shared memory / state with the viewer

**Menu bars**
- [x] File IO (load / save)
- [x] Show keybindings
- [x] help with useful version info

**Tutorials / docs / install / tests**:
- [x] better test coverage
- [x] make docs readable & hosted
- [x] make tutorials, include gifs
- [x] Document windows install
- [x] documenting launch with gui magic as people forget

hectormz:
@sofroniewn you might guess I'm going through old issues 😁Should we reassess the unfulfilled features and add them/make an issue for a future release?

sofroniewn:
[...]

PR #460, opened 2019-07-31
Initial work on updating status bar

jni:
# Description

This PR makes the following changes:

- active layer is now the topmost visible layer, regardless of selection. (See below for discussion.)
- printing of status messages based on value is now handled centrally, and much improved. I haven't changed over the obvious (integer) prints. In a future PR, I think this machinery can be used to print shape info, rather than just the shape index.
- if the cursor is out of range of the data, no value is shown, rather than showing the value at the nearest edge.
- a new keyboard shortcut is added, toggling visibility of the topmost layer.

## Type of change

# How has this been tested?

## Final checklist:

# Discussion points

I'm still not sure about the active layer behavior. Now, it's a bit more cumbersome to mess with a layer's settings: for example, if I have a points layer and an image layer underneath it, I have to hide the points layer before I can adjust the clims on the image. However, this does solve the problem of adjusting settings on an invisible layer.

My suggestion, which I can implement here or in a followup PR depending on preferences, is that the active layer be defined as:
- if no layers are selected, the topmost visible layer.
- if a layer is selected, it is the active layer *if it is at all visible*, which means that only the following layers are in front of it:
  * image, labels, and pyramid with opacity < 1
  * points, shapes, or vectors, with any opacity (since these are generally sparse, visually)

I'll wait for thoughts before working on that logic.

Regarding the keyboard shortcut, topmost is kinda clunky, it turns out. My suggested improvement, which I can implement here or in a followup, is:

- Ctrl-V: hide the next visible layer (ie the active layer)
- Ctrl-Shift-V: show the next not-visible layer (ie the layer above the active layer), if any
- V: undo the last visibility change

Thoughts? And I just realised the shortcuts will have to change because Ctrl-V is taken. Suggestions welcome!

### DIFF
--- a/napari/layers/labels/labels.py
+++ b/napari/layers/labels/labels.py
@@ -8,7 +8,6 @@ from imageio import imwrite
 
 from ..base import Layer
 from vispy.scene.visuals import Image as ImageNode
-from ...util.status_messages import status_format
 from ...util.colormaps import colormaps
 from ...util.event import Event
 from ...util.misc import interpolate_coordinates
@@ -495,7 +494,11 @@ class Labels(Layer):
             String containing a message that can be used as a status update.
         """
         int_coord = np.round(coord).astype(int)
-        msg = f'{self.name} {int_coord} label {value}'
+        if value is not None:
+            value_str = ' label {value}'
+        else:
+            value_str = ''
+        msg = f'{self.name} {int_coord}{value_str}'
 
         return msg
 
--- a/napari/layers/labels/labels.py
+++ b/napari/layers/labels/labels.py
@@ -8,6 +8,7 @@ from imageio import imwrite
 
 from ..base import Layer
 from vispy.scene.visuals import Image as ImageNode
+from ...util.status_messages import status_format
 from ...util.colormaps import colormaps
 from ...util.event import Event
 from ...util.misc import interpolate_coordinates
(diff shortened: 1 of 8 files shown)


jni:
Also, after this PR is resolved, I do want the status bar to show the status from *all* visible layers in a sequence. But that can come in a followup, and I could definitely use some guidance from @sofroniewn about how to go about this!

napari/layers/pyramid/pyramid.py
```diff
@@ -1,4 +1,6 @@
 import numpy as np
+
+from napari.util.status_messages import status_format
```
sofroniewn: relative path?

napari/layers/labels/labels.py
```diff
@@ -495,7 +495,7 @@ def get_message(self, coord, value):
             String containing a message that can be used as a status update.
         """
         int_coord = np.round(coord).astype(int)
-        msg = f'{int_coord}, {self.name}, label {value}'
+        msg = f'{self.name} {int_coord} label {value}'
```
sofroniewn: do we need special handling for `None`?

napari/viewer.py
```diff
@@ -18,3 +18,10 @@ def __init__(self, title='napari'):
         self.window = Window(qt_viewer)
         self.screenshot = self.window.qt_viewer.screenshot
         self.camera = self.window.qt_viewer.view.camera
+
+
+@Viewer.bind_key('v')
```
sofroniewn, line 24: i feel like this should go in the `qt_viewer.py` file, not here - right @kne42?

sofroniewn (commented):
I really like that you've consolidated the `status_format` into one function. Do you think we could go one step further and move the `get_message` function to the utils too? i.e. something like the following:

```python
def get_message(name, coord, values=None):
     # concatenate string name + convert coord to ints + concatenate values, where
     # if values is None then ignore, or if a dictionary then use string keys
     # and pass values through formatter, so could have
     # {'value' : (230, 221, 10)} or {'shape' : 10, 'vertex' : 2} for image / shapes
     # shapes layer respectively all turn out the way we want.
```

This could really cut back on duplicated code and make our layer model files much simpler.

I'll give the comments around visible / selected etc. more thought but my initial reaction is to like your suggestion around treating image-like and shape-like layers differently as there are inherent differences in their sparsity

napari/layers/pyramid/pyramid.py
jni: Ah yes, good point, PyCharm does these imports automatically and I forgot to update this one.

Drives me nuts that they can't do relative imports!!! I might write a support ticket.

napari/layers/labels/labels.py
jni: ... Yes, yes we do.

napari/viewer.py
jni: The layer shortcuts are all on the models... But you might be right!

jni:
This is still WIP, but @sofroniewn @kne42, note that when I move my `@Viewer.bind_key` to `@ViewerModel.bind_key`, I can't get the keybindings to trigger at all... Is this expected?

jni:
Also, @sofroniewn, I believe you were going to give me some pointers about where to start for concatenating status messages from multiple layers together? Apologies if you already did and I have lost track of it.

sofroniewn:
@jni i would do the concatenation of status messages inside this for loop - https://github.com/napari/napari/blob/ebc5474a59a89496c99c43129b65e63afa7cc662/napari/components/viewer_model.py#L705

and then do `self.status = concat_status` at the end instead of `self.status = active_layer.status` and that will trigger all the updates needed

sofroniewn:
@jni with the merging of #485 which incorporates some of the code from this PR I recommend we close this PR. Also I added similar keybindings from this PR in #472.

You can then make a new PR that makes all the status bar printing from many layers happen at the same time.

jni:
Yep, that was the impression I had. Will deal with unifying status bars mid-september.

sofroniewn:
Sounds good. I bumped unifying status bars to `0.2` so it's on our list in #301 now :-)

PR #461, opened 2019-08-01
Use dask to load many images lazily from CLI

jni:
# Description
What the title says. =P

Recently I was helping a student who needed to annotate thousands of images. Annotation is quick with the right tool, and I thought napari could do the trick. I think it will, but we needed a way to load the images lazily.

This still won't solve the problem directly because we don't have a save button for the points layer (or any layer!), but at least it provides some sample code for how one could do it.

## Type of change

# References

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/main.py
+++ b/napari/main.py
@@ -3,6 +3,8 @@ napari command line viewer.
 """
 import argparse
 import numpy as np
+from dask import delayed
+import dask.array as da
 from skimage import io
 
 from . import Viewer, gui_qt
@@ -19,13 +21,28 @@ def main():
     parser.add_argument(
         '-m',
         '--multichannel',
-        help='Treat images as RGB.',
         action='store_true',
+        help='Treat images as RGB.',
+    )
+    parser.add_argument(
+        '-d',
+        '--dask',
+        action='store_true',
+        help='Load images lazily when given multiple images. Ignored with '
+        '--layers.',
+    )
+    parser.add_argument(
+        '-c',
+        '--clim-range',
+        type=lambda s: tuple(map(float, s.split(','))),
+        default=None,
+        help='Define the contrast limits for the input image. Highly '
+        'recommended when using --dask.',
     )
     args = parser.parse_args()
     with gui_qt():
         v = Viewer()
-        images = io.ImageCollection(args.images, conserve_memory=False)
+        images = io.ImageCollection(args.images, conserve_memory=True)
         if args.layers:
             for image in images:
                 v.add_image(image, multichannel=args.multichannel)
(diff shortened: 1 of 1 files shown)


sofroniewn:
A couple things:

1) adding dask to the CLI means adding it to our full dependencies - how do we want to think about optional dependencies for file-io? do you want to make a separate list of these dependencies, like zarr / dask / xarray? right now we have them in our "test" dependencies. @kne42 might have ideas

2) as you said for now there is no saving of labels, adding some in will be fine, but right now the labels layer doesn't work with a dask array as you can't "paint" into them - i.e. you can't edit the data in place. not sure how to solve that one yet, but might be easier if we actually have file saving enabled and we can just write our changes right back to the file

kne42:
agreed that we shouldn't have dask as a requirement. this code should be structured such that it will only work/appear as an option if dask is installed. then, we add an optional dependency package as `napari[cli]`, `napari[io]`, etc. to direct people to install if they want all I/O features

jmetz:
Is there a reason the `skimage.io.ImageCollection` (which is already being used) can't be switched to using the `conserve_memory` flag to lazily load the images instead? Are there performance improvements when using `dask`?

jni:
No performance improvements, but yes API improvements, because Napari expects an array like object, with things like ndim and shape, which ImageCollection lacks.

jmetz:
But `skimage.io.ImageCollection` is already being used in `napari` - is the intention to remove that then? 
What aspects of the array like object are needed other `ndim` and `shape`?
There  might be an argument for adding a lightweight adaptor on top of the `ImageCollection` to give it some of these (easy to implement) array-like attributes - rather than pulling in a large project like dask as a core dependency.

sofroniewn:
@jmetz we'll be working on our file IO quite a bit over the next 6 weeks. we're got a couple issues #379 and #490 where we've been discussing these things. In general I'd definitely like to standardize our file IO across the different ways of loading images (CLI, drag / drop, menubar) and use optional dependencies to allow people to take advantage of more advanced file IO options without forcing everyone to depend on them.

I think we'll probably use `imageio` for the simple stuff (which already is a dependency), and then we might push the lazy stuff onto dask / dask image, but we'll need to see.

For us it's really nice to have an array like object that you can index into as you index into a numpy array, and which when you call `np.asarray` on actually gives you a numpy array. I think adding that sort of indexing onto the `ImageCollection` would be a lot.

jmetz:
@sofroniewn  ah got it - thanks for the clarification. Looking forward to seeing how things evolve!

sofroniewn:
@jmetz sounds good, if you have any more requests around io feel free to add them to those issues or make a new issue!

sofroniewn:
@jni can this be closed as superceeded by #560?

jni:
@sofroniewn well, yes and no, since as you pointed out there, there is an error in my implementation there!

sofroniewn:
:-) we can make some follow up PRs to #560

sofroniewn:
This has now been superceeded by #590 right @jni? I'm closing

Issue #436, opened 2019-07-21
Proper suport for scales in 2D and 3D layers


jni:
## 🚀 Feature

Currently, scale on a layer can be set with `layer.scale = <tuple>`, which is great. However, this method currently sets vispy scale directly (?) and so does not follow row/column aka zyx coordinates as we do. For example, in the following code:

```python
from skimage import data
import napari
v = napari.Viewer()
image_layer = v.add_image(data.camera())
image_layer.scale = (1, 2)
```

The image is stretched along the *vertical* dimension, when NumPy axis conventions would lead us to expect that the horizontal dimension is stretched.

The scale parameter should be set with a property + setter to properly transpose the coordinate scales.

pranathivemuri:
we can allow it to be a parameter that can be tuned on the layer. I can add that it in. But I was wondering if we could set the whole 3  tuple of scale at once or do we want slider buttons for each of the scales in x y and z. We transpose/reverse the scales set on the layer from xyz to zyx and xy to yx in the property setter

jni:
@pranathivemuri imho we don't want a slider: the scale is something that is fixed for each image. Eventually we'll want to adjust it based on metadata, but an input parameter works great. I agree about transposing the scales in the property setter. =) Thanks!

constantinpape:
I agree with @jni that this should be a fixed parameter (and that it should respect the same axis-convention as napari). IMHO, it should be a keyword argument for `add_image` and other relevant functions.

Also, is this supported for the pyramid layer already? I.e. the scale of each level should be something like `scale_of_level = [sc1 * sc2 for sc1, sc2 in zip(scale, downsample_level)]`, where `scale` is the scale factor of level 0 and `downsample_level` the downsampling factor of the current level.

sofroniewn:
This has now been mostly implemented in #493 - each layer accepts a `scale` keyword arg that is in the axis conventions of napari.

inside the pyramid layer the `scale` right now gets autocomputed based on the downsampling, but the proposal by @constantinpape to have both a `scale` and `scale_of_level` would be preferable. I'll implement that shortly

Similarly the `translate` property is now a keyword arg on every layer, but in pyramids is getting overloaded and an additional `translate_of_view` property would be much cleaner.

PR #463, opened 2019-08-04
expose scaling factor for volume

sofroniewn:
# Description
This PR exposing a scaling factor of length equal to number of dimensions on the volume layer - it's a little bit of a quick fix as right now updating the scaling factor won't update the camera rotation center, but that functionality is implemented in #451 which will hopefully merge soon.

This partially addresses #436. I think though I'd rather expose scaling on all layers and have it work with the range slider etc. too. I think that can go into #451 or even later PRs though, and this will at least get something better than what we have previous.

spacing

Note this is a cube with a `[3, 1, 1]` spacing

## Type of change

# How has this been tested?

- [x] modified `examples/add_volume.py`

## Final checklist:

### DIFF
--- a/napari/layers/volume/tests/test_volume.py
+++ b/napari/layers/volume/tests/test_volume.py
@@ -94,6 +94,21 @@ def test_name():
     assert layer.name == 'img'
 
 
+def test_spacing():
+    """Test instantiating anisotropic 3D volume."""
+    shape = (10, 15, 20)
+    spacing = [3, 1, 1]
+    full_shape = tuple(np.multiply(shape, spacing))
+    np.random.seed(0)
+    data = np.random.random(shape)
+    layer = Volume(data, spacing=spacing)
+    assert np.all(layer.data == data)
+    assert layer.ndim == len(shape)
+    assert layer.shape == full_shape
+    assert layer.range == tuple((0, m, 1) for m in full_shape)
+    assert layer._data_view.shape == shape[-3:]
+
+
 def test_visiblity():
     """Test setting layer visiblity."""
     np.random.seed(0)
--- a/napari/layers/volume/volume.py
+++ b/napari/layers/volume/volume.py
@@ -113,7 +113,7 @@ class Volume(Layer):
             self._data = volume
             self.metadata = metadata or {}
 
-            self.spacing = spacing or [1] * self.ndim
+            self.spacing = spacing or [1] * len(volume.shape)
 
             # Intitialize volume views and thumbnails with zeros
             self._data_view = np.zeros((1, 1, 1))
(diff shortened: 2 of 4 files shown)


napari/layers/volume/tests/test_volume.py
```diff
@@ -94,6 +94,21 @@ def test_name():
     assert layer.name == 'img'
 
 
+def test_spacing():
+    """Test instantiating anisotropic 3D volume."""
+    shape = (10, 15, 20)
+    spacing = [3, 1, 1]
+    full_shape = tuple(np.multiply(shape, spacing))
```
jni, line 101: I'm really confused about this... Does this mean that the data is getting multiplied?

napari/layers/volume/volume.py
```diff
@@ -150,7 +159,17 @@ def data(self, data):
...
+
+    @property
+    def spacing(self):
+        """list: Anisotropy factors to scale the volume by."""
+        return self._spacing
+
+    @spacing.setter
+    def spacing(self, spacing):
+        self._spacing = spacing
+        self.scale = [self._spacing[s] for s in [-1, -2, -3]]
```
jni: I *think* this is better?

```python
visible_dims = ...
self.scale = [self._spacing[s] for s in visible_dims[::-1]]
```

Then you might be able to put that code in the base layer instead of Volume...?

jni (approved):
@sofroniewn I've left a few comments, but approved either way! Thank you for getting this fixed!

napari/layers/volume/tests/test_volume.py
sofroniewn: yeah - this is maybe not so good, the data is not getting multiplied, but i needed to increase the "shape" so that the camera setting now gets the right centroid and field of view. It's not ideal, I need to think about how this can maybe get better ...

napari/layers/volume/volume.py
sofroniewn: yeah - that will also make the integration with #451 easier as that PR creates an explicit parameter of that sort. I'll still probably keep in the volume for now though, as i'm not sure how we want to think about this sort of scaling yet - for example would be support this sort of scaling to convert things from pixels to microns?

napari/layers/volume/tests/test_volume.py
jni: Ooohhh... Hm. Ok I'll think about it also...

jni: First thought: this is the difference between our data coordinates and our "real world" coordinates, so we now need to start thinking about those, and maybe use `shape` and `real_shape`, or somesuch?

sofroniewn: yup agreed - i think that's the long term solution. for now this will be ok on just the volume layer

napari/layers/volume/volume.py
sofroniewn: fixed

sofroniewn:
merging now - will likely refactor in #451 and beyond

Issue #424, opened 2019-07-16
More selection rules discussion + a keyboard shortcut


jni:
After deleting a layer, the layer underneath it should be selected. This allows rapid deletion of layers by clicking the trashcan icon, rather than the current behavior, which requires going back and forth between the trash icon and the next layer. When deleting the bottommost layer, the layer above it should be selected.

Finally, we should add a shortcut, say Shift-Backspace, for deleting layers.

sofroniewn:
This is easy to do. @kne42 is going to move over the current keybindings to the new style soon. once that is done i can do this

jni:
CC @stefanv who just pinged me privately to ask for exactly this. ;)

sofroniewn:
fixed by #472

jni:
This is amazing @sofroniewn! I just tried it and it's soooo much nicer! :clap: 

I agree wholeheartedly with the decision to merge and revisit default bindings later, btw. =) Thank you!

Issue #260, opened 2019-05-14
Hotkey focus


sofroniewn:
## 🐛 Bug

Right now hotkeys require the main canvas (the inset where the images are seen) to be have focus to be active. This leads to confusing user interactions and is very frustrating when using the paintbrush in the labels layer as when clicking back into the part where the image is you accidentally do some painting.

## Expected behavior

The hotkeys should always be functional regardless of where the user has clicked and which part of the viewer has focus.

## Additional context

This was flagged by @gregjohnso

sofroniewn:
fixed by #472

Issue #227, opened 2019-04-20
Support left/right keys to move slider pointer


royerloic:
## 🚀 Feature

Would be good to be able to move the currently focused dim slider with left/right keybindings by one step. Pressing <Shift> would make it 10 steps or so. 

## Motivation

Sometimes one wants to scroll back and forth between two or several slices in a reproducible manner. This is easier by pressing keys.

sofroniewn:
Once we merge #171 and start using our custom `rangeslider` we have to implement this ourselves (it is currently supported on master which uses the built in sliders from PyQt)

royerloic:
I encountered this issue again today while browsing a dataset, 
turns out to be quite useful in practice as we know.
Just adding some weight to this issue here :-)

sofroniewn:
fixed by #472

Issue #136, opened 2019-03-01
Bring back keyboard shortcut for Fullscreen mode


royerloic:
## 🚀 Feature

Bring back keyboard shortcut for Fullscreen mode.
Press 'F' and you enter fullscreen. 'ESC' or 'F' again to leave fullscreen.

## Motivation

It's the best feature. So good. 

## Pitch

I just explained that this is the best feature.
People tell me they love this feature --  it is true.

## Alternatives

None. 

## Additional context

Fullscreen is all there is to see.

sofroniewn:
Ah I had not realized this had disappeared but I will look into it (I have an idea why). Maybe it's also time to add ctrl+W, ctrl+Q to close / quit? Need to think about where these live. Do they go on window or do the go on viewer or do they live somewhere else? My thoughts are on Window right now.

jni:
I think it disappeared in a very early refactor. I agree that shortcuts should go on the window, (w in ctrl-w is for window), but f is possibly for the viewer.

sofroniewn:
fixed by #472

PR #472, opened 2019-08-10
add viewer keybindings

sofroniewn:
# Description
This PR adds standard keybindings to the viewer and improves the keybindings focus, fixing #260

It adds 

`ctrl-F` to toggle fullscreen mode - fixing #136. Note that `escape` will not exit fullscreen mode as we are using escape to end polygon drawing on the shapes layer, but that can be changed if desired @royerloic. 

`ctrl-D` to delete any selected layers. also after selected layers are deleted the next layer is selected - fixing #424

`crtl-R` to reset the camera zoom

`ctrl-A` to select all layers

`ctrl-[` and `ctrl-]` to select the previous or next selected layer, superseeding some of the keybindings that were to be added in #460 (I'd remove the keybindings from that PR now @jni)

`ctrl-T` to toggle the theme (which now includes changing of the vispy canvas background color).

`Left` and `Right` to move the dimensions of the last used or focused axis, fixing #227 and a request from @d-v-b.

`Up` and `Down` to swap which is the focused axis.

I'll note that I havn't looked at menubar integration for any of these keybindings, and that I had to add the keybindings to the `Viewer` as adding them to the `ViewerModel` didn't seem to work. @kne42 do you think this is being done properly in this PR?

Finally @kne42 I had to use `Control-Backspace` instead of `Control-Delete` for it to work properly on my mac - are we handling differences between backspace and delete properly on a mac? should these be aliased? i know on windows they can be different.

## Type of change

## Final checklist:

### DIFF
--- a/napari/_qt/qt_viewer.py
+++ b/napari/_qt/qt_viewer.py
@@ -259,6 +259,7 @@ class QtViewer(QSplitter):
         self.console.style_sheet = themed_stylesheet
         self.console.syntax_style = palette['syntax_style']
         self.setStyleSheet(themed_stylesheet)
+        self.canvas.bgcolor = palette['canvas']
 
     def _toggle_console(self):
         """Toggle console visible and not visible."""
@@ -293,7 +294,10 @@ class QtViewer(QSplitter):
     def on_key_press(self, event):
         """Called whenever key pressed in canvas.
         """
-        if event.native.isAutoRepeat() or event.key is None:
+        if (
+            event.native.isAutoRepeat()
+            and event.key.name not in ['Up', 'Down', 'Left', 'Right']
+        ) or event.key is None:
             return
 
         comb = components_to_key_combo(event.key.name, event.modifiers)
@@ -331,6 +335,14 @@ class QtViewer(QSplitter):
         """
         for layer in self.viewer.layers:
             layer.on_draw(event)
+
+    def keyPressEvent(self, event):
+        self.canvas._backend._keyEvent(self.canvas.events.key_press, event)
+        event.accept()
+
+    def keyReleaseEvent(self, event):
+        self.canvas._backend._keyEvent(self.canvas.events.key_release, event)
+        event.accept()
 
     def dragEnterEvent(self, event):
         if event.mimeData().hasUrls():
--- a/napari/_qt/qt_viewer.py
+++ b/napari/_qt/qt_viewer.py
@@ -298,6 +298,8 @@ class QtViewer(QSplitter):
             event.native.isAutoRepeat()
             and event.key.name not in ['Up', 'Down', 'Left', 'Right']
         ) or event.key is None:
+            # pass is no key is present or if key is held down, unless the
+            # key being held down is one of the navigation keys
             return
 
         comb = components_to_key_combo(event.key.name, event.modifiers)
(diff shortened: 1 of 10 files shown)


d-v-b:
These are great! You could do this in a different PR, but the up/down arrow to change axis should have some corresponding visual indication of which axis is "active".

jni:
@sofroniewn at airport currently so forgive my brevity, but what happened to minimising use of "Ctrl" modifiers?

(With the exception of changing themes which I think should be harder actually... Ctrl-Shift-T?)

I think Esc should definitely be able to exit full screen mode... Perhaps we need our shortcuts to be context-dependent...?

sofroniewn:
@jni I was imaging something like we'd make the keybindings on the Viewer use `Control` as they are more like application keybindings, `Control-Q` to quit etc. and leave the keybindings on the layers as the single letters to prevent collisions, but I don't have strong feelings.

We should look across all our keybindings now and think about if they are good assignments. 

Maybe something @kne42 can do is think about how to we want to surface the keybinding info to the user in the app? i.e. they click the question mark and it prints to the console some help dictionary with info about the keybindings? could be a place to start. thoughts @jni? 

I don't like `escape` having double meaning - probably best to make it exit fullscreen mode and then have another way to finish drawing a polygon.

royerloic:
> `ctrl-F` to toggle fullscreen mode - fixing #136. Note that `escape` will not exit fullscreen mode as we are using escape to end polygon drawing on the shapes layer, but that can be changed if desired @royerloic.

Overall I like the bindings!
I am not sure about the debate on CTRL versus non-CTRL, we can discuss that next thu @jni 
For full screen, it would be great to be able to double-click on the canvas, but haa... I remember we don't have double click...

royerloic (approved):
I like it :-)

napari/_qt/qt_viewer.py
```diff
@@ -293,7 +294,10 @@ def on_mouse_release(self, event):
     def on_key_press(self, event):
         """Called whenever key pressed in canvas.
         """
-        if event.native.isAutoRepeat() or event.key is None:
+        if (
+            event.native.isAutoRepeat()
+            and event.key.name not in ['Up', 'Down', 'Left', 'Right']
```
kne42, line 299: could you include a comment explaining this?

kne42 (approved):
Yes, you did handle it right – `Viewer` would overwrite the class keybindings defined in `ViewerModel`. I have been considering creating a new implementation sometime based off of [`ChainMap`s](https://docs.python.org/3/library/collections.html#collections.ChainMap), which would allow multiple-inheritance via both classes and, if useful, context (for example, fullscreen mode provides all the normal viewer keybindings but additionally defines `Escape` to exit the mode).

As for the OSX stuff, this was a choice on Qt's part since Windows's `Backspace` and OSX's `Delete` perform the same operation, along with how `Control` is treated as `Command`. We can address this issue and others in a PR handling key-aliasing,

napari/_qt/qt_viewer.py
sofroniewn: added

sofroniewn:
I've added the note requested by @kne42

I completely get the desire by @jni to keep the shortcuts simple, but for now I'm just going to merge this. @kne42 is going to work on adding a function to print all our default keybindings to the console (we can ultimately display this graphically some other way, but this is ok for now). We can then go through that list and decide how things should be named

jni:
So many closed issues! :heart_eyes: btw, did you notice the sad state of our CI here? Since I noticed it in other PRs I guess it is an existing issue on master...

PR #479, opened 2019-08-15
fix windows ci build

kne42:
# Description
main fix involves working around a bug reported in cirruslabs/cirrus-ci-docs#423

the build originally broke because `CIRRUS_WORKING_DIR` was changed and introduced a bug where the default was not set prior to environment variable definition

this means that our environment variable, `VISPY_GL_LIB`, which was set to `$CIRRUS_WORKING_DIR\opengl32.dll`, came out equivalent to `\opengl32.dll`

since this environment variable was used by VisPy to determine where to load the OpenGL library from, it could not find it

conda installation was also an issue so we're now using `choco`, a windows package manager, to install it (comes pre-installed in cirrus-ci builds)

thanks to @csweaver for her help with sleuthing around the possible causes :)

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -44,28 +44,29 @@ task:
 
     - name: windows
       windows_container:
-        image: cirrusci/windowsservercore:2016
+        image: cirrusci/windowsservercore:2019
+        os_version: 2019
       env:
-        PATH: $PATH;$USERPROFILE\anaconda\Scripts;$USERPROFILE\anaconda;$ALLUSERSPROFILE\chocolately\bin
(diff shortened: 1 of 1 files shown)


kne42:
since cirruslabs/cirrus-ci-docs#423 was resolved, the build on master should be fixed

most of the other changes are related to cleanup/readability so I'm closing this and will include them in another PR that cleans up our implementation and fixes the OSX build

kne42:
actually I forgot that conda installation was an issue too so this is still relevant

sofroniewn (approved):
Looks good to me!

PR #483, opened 2019-08-16
remove vispy backport

kne42:
closes #477

PR #484, opened 2019-08-16
clean up black pre-commit hook & exclusion pattern

kne42:
# Description
clean up how black is used in napari:
- expand exclusion pattern to include build-related files
- exclude examples and free them from `black`'s iron grip (let's be real, they look hideous)
- force `black` to run on entire repo in the pre-commit hook. while it will take a little bit longer (realistically almost no time at all), it will ensure that `black` conforms to the exclusion patterns as defined in `pyproject.toml`. we can alternatively define a meta hook to auto-update the pre-commit config so that the exclusion patterns match, but that's probably not necessary

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,6 +3,6 @@ repos:
     rev: stable
     hooks:
     - id: black
-      language_version: python3
-      require_serial: true
-      exclude: 'vendored'
+      # force black to run on whole repo & keep settings from pyproject.toml
+      pass_filenames: false
+      args: [.]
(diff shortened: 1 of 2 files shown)


sofroniewn (approved):
I'd like to merge this right after we merge #483, but at that point can you remove `_vispy` from the `-exclude` as we will no longer have a back port, and I'm planning to use `_vispy` for our vispy view files

AhmetCanSolak (approved):
agreed with @sofroniewn about merging order, seems good

kne42:
I've rebased this on the latest master and removed `_vispy` from the exclude pattern

Issue #462, opened 2019-08-01
Turn off ndim for painting by default


marshuang80:
## 🚀 Feature
Requesting to turn off n-dim paint by default. 

## Motivation
I think cases where users will paint on multiple dimensions by default is small. Having paint default to ndim will confuse users when they see their paintings leak to other dimensions. 

## Pitch

Simply default ndim painting to False. 

## Alternatives

Better indication that paint bush is painting on multiple dimensions.

sofroniewn:
Yup, makes sense - happy to change that default

sofroniewn:
fixed in #485

Issue #459, opened 2019-07-31
Add hover status to vectors layer


jni:
## 🚀 Feature
Currently, hovering on any layer except the vectors layer produces a little status message about what's under the cursor in the currently visible layer. It would be nice, especially for nD vectors, if hovering over a vector produced a message about that vector's projection.

Given that our internal data format is (?) [coord -> projection], we probably need a cKDTree to efficiently find the nearest vector.

sofroniewn:
we can reuse code from the shapes layer for this - but I worry about the efficiency with millions of vectors. For now you can just get it to print the name / coords and nothing about the vectors. I think this is better than printing absolutely nothing, and something you could even put in #460

jni:
The efficiency is why I was suggesting cKDTree — I don't think you use that at all?

https://jakevdp.github.io/blog/2013/04/29/benchmarking-nearest-neighbor-searches-in-python/

The numbers there I'll admit are not super encouraging, but the examples are also quite different from what we would need. 2D seems to be pretty speedy.

Anyway, we could consider *not* homogenising the input data to list-of-vectors format. Millions of vectors is likely to only be an issue with dense data.

jni:
(And I think K=1 will also be reasonably speedy.)

sofroniewn:
the vectors layer now returns its coordinates when hovering. we could additionally return if we are over a vector or not, or maybe something like the length / direction of the current vector? I'm leaving this open but view these as more advanced status updates and bumping this issue to `0.2`.

@bryantChhun @jni do you have opinions? if it is fine as is then i will close

jni:
I'd suggest leaving it open, and maybe even bumping to 0.3.

Issue #458, opened 2019-07-31
Incorrect clipping at bottom edge of pyramid


jni:
## 🐛 Bug

(Sorry, no gif right now, will produce if you can't reproduce it.)

When I run `examples/add_pyramid.py`, if I zoom near the very bottom of the image, the image is cut off inside the viewport. This happens right at the very beginning of a level change. (ie when I go from 2 to 1, or 1 to 0.)

sofroniewn:
fixed in #485

jni:
@sofroniewn still seeing this on master on Linux... =\

sofroniewn:
Note this example here https://forum.image.sc/t/efficiently-open-qptiff-images-with-napari/33237/4?u=sofroniewn

> However, I’m now facing a weird problem. Napari opens with only the top left quarter of the image:

Some of this could be caused by their canvas being larger than we expect (say fullscreen mode on a big monitor). Right now we hard code this max tile shape https://github.com/napari/napari/blob/master/napari/layers/image/image.py#L136 which is currently in the model, but as has been pointed out by @d-v-b maybe this should be living in more of the view code and should be dependent on the canvas size, something which only the view knows ...

PR #485, opened 2019-08-17
remove vispy code from layer models

sofroniewn:
# Description
This PR removes vispy code from the layer models and is very much a WIP


sofroniewn:
@kne42 @shanaxel42 @jni - this is still pretty preliminary but I now have a working version of pulling out the vispy code from the base layer and the image layer. A lot of what used to be in the `_visual_wrapper.py` file has now been moved to `_vispy/vispy_base_layer.py` and a new `_vispy/vispy_image_layer.py` has been created to handle the image specific stuff.

Only the `add_image.py` and `nD_image.py` examples will work right now as I havn't migrated any of the other layers - but I'm curious if people think that I'm on the right track here or if there's another way I should be achieving our model / view separation

sofroniewn:
I've now started exploring converting the `Labels` layer. One thing that's coming up is where to put the mouse functions? Right now I'm moving them over the the `_vispy` folder but there's a lot of code that interacts with the models inside them - maybe it would be nice for them to have their own folder like the `keybindings.py`, we could have `mousebindings.py` @kne42, but that can wait till a followup PR.

I've also moved the `_vispy` folder inside `layers` as it's really a folder with the layer specific vispy code and not our backport or anything else

jni:
I think it would be good to ping @royerloic here too as he has a lot of experience with this model/view separation stuff.

I'm going to have a quick look but may not have time to think about this much in the next few days. I hope to have a bit more time in a week.

napari/layers/_vispy/util.py
```diff
@@ -0,0 +1,29 @@
...
+            Layer that needs its propetry widget created.
+
+    Returns
+    ----------
+        node : vispy.scene.visuals.VisualNode
+            Vispy visual node
+    """
+    name = 'Vispy' + type(layer).__name__ + 'Layer'
+    try:
+        VispyLayer = globals()[name]
```
jni: This seems brittle...? What's happening here? I think you're accessing the module namespace... I suggest instead to create a dictionary mapping names to objects.

napari/layers/_vispy/vispy_image_layer.py
```diff
@@ -0,0 +1,35 @@
+from vispy.scene.visuals import Image as ImageNode
```
jni: The `_vispy` folder should not be in `layers`, as it is a view. Only models should be in `layers`, at least as I understand the purpose of this refactor...?

jni:
Other than my comment about the location of the files, this looks ok to me. I agree that it would be great if model-specific keybindings could be on the models, and then they would be translated to our Qt view or (in the far future) web view accordingly. I think @kne42 was thinking about how to achieve that?

sofroniewn:
@kne42 @royerloic @shanaxel42 @jni - this PR is now ready for review

It separates out the vispy view code into their own files, currently living inside a folder called `layers/_vispy` though @jni has suggested moving elsewhere. The purpose of the refactor is definitely to pull out the code from the model files, but I don't mind where this folder lives.

I've also done some consolidation and cleanup of the layer model files at the same time, including pulling out the status bar formatting code from #460 and including it here - I did not tackle making all the layer status bar updates appear together, but that can be done in follow up PRs. I've also consolidated some of the `get_value` code.

I've also hopefully fixed the pyramid clipping observed in #458, made the labels `n_dimensionsal` property to be a keyword arg and `False` by default as requesting in #462.

I've also made the vectors layer print its name and coordinates when hovering, partially addressing #459 - is that sufficient? or do you want to know if the mouse is over a vector or not? that seems like not very useful information for a lot of compute. thoughts @bryantChhun?

there are a couple things not ideal with the refactor like `scale_factor` getting recomputed everytime there is a draw event, but it's pretty fast and i couldn't think of a better way to do it.

examples/add_volume.py
```diff
@@ -12,4 +12,4 @@
     )
     viewer = napari.Viewer()
     # add the volume
-    viewer.add_volume(blobs, spacing=[3, 1, 1])
+    viewer.add_volume(blobs, scale=[3, 1, 1])
```
jni, line 15: What's the dealio here? =)

napari/_qt/qt_viewer.py
```diff
@@ -54,6 +56,8 @@ def __init__(self, viewer):
         self.buttons = QtLayersButtons(self.viewer)
         self.console = QtConsole({'viewer': self.viewer})
 
+        self.vispy_layers = {}
```
jni: Can you add a comment here about what this dictionary is meant to hold? I'm not sure about `vispy_layers`, either... Maybe `layer_to_visuals`, `layer_visual`, `layer_visual_nodes`, `layer_scenegraph`...?

napari/components/viewer_model.py
```diff
@@ -467,7 +460,7 @@ def add_volume(
         opacity=1,
         blending='translucent',
         visible=True,
-        spacing=None,
+        scale=None,
```
jni, line 463: I'm not sure we want to rename this...?

napari/components/viewer_model.py
```diff
@@ -503,7 +496,7 @@ def add_volume(
             {'opaque', 'translucent', and 'additive'}.
         visible : bool
             Whether the layer visual is currently being displayed.
-        spacing : list, optional
+        scale : list, optional
             List of anisotropy factors to scale the volume by. Must be one for
             each dimension.
```
jni: "Must be one for each dimension" -> "The length should be equal to the number of dimensions". Otherwise it sounds like it must literally be [1, 1, ...].

napari/layers/_vispy/util.py
jni: This remains unaddressed... =)

napari/layers/_vispy/vispy_image_layer.py
jni: I know you've acknowledged this comment, but you haven't really discussed it or done anything about it... =)

napari/layers/base/base.py
```diff
@@ -197,28 +260,14 @@ def position(self, position):
...
-        coords = list(self.dims.indices)
-        for d, p in zip(self.dims.displayed, position):
-            coords[d] = p
-        self.coordinates = tuple(coords)
-
     def _update_dims(self):
         """Updates dims model, which is useful after data has been changed."""
         range = self._get_range()
-        self.dims.ndim = len(range)
+        if len(range) != self.dims.ndim:
```
jni: Can you add a comment explaining this condition? Under what circumstances will this be true?

Additionally, `range` shadows the built-in — it's not a *huge* deal, but I would prefer if we avoided that. `curr_range`? `self_range`?

napari/layers/points/points.py
```diff
@@ -125,107 +121,93 @@ def __init__(
...
+            self._size = 10
 
-            if type(edge_color) is str:
-                self._edge_color = edge_color
-            else:
-                self._edge_color = 'black'
+        if type(edge_color) is str:
+            self._edge_color = edge_color
+        else:
+            self._edge_color = 'black'
```
jni, line 161: This seems dangerous...? What if someone says `edge_color=(0, 0, 1)`? I suggest we use matplotlib's universal color handling here... (I believe I vendored it.)

jni: Having said that, I recognise that it was already like this, so we can leave this fix till later... But maybe add a TODO comment? Thanks!

jni (commented):
@sofroniewn it's wonderful how clean this is! I expected this would be a much messier job! As far as I can tell, @royerloic's commandments are fulfilled here, which is so great. It opens the road to really clarifying what the napari data model is, and thus how to build totally different viewers on top of it. I love it.

I've left a couple of minor comments, and, as I mentioned, I really think the `_vispy` folder should be moved out of `layers`. That really cleans up the model and allows others to build on it independently of vispy.

Great stuff though! And I love how you've taken the chance to fix so many other things! :joy:

napari/_qt/qt_viewer.py
sofroniewn: renamed `layer_to_visual` and comment added

napari/components/viewer_model.py

napari/components/viewer_model.py
sofroniewn: the reason i renamed is that `scale` is a property of all our layers - i'll probably move the keyword arg to the base layer (along with exposing `translate`), in prepartation for merging the volume layer with the image layer. at that point i felt like it was confusing to have `spacing` for non-volumetric stuff, whereas `scale` was clear regardless

examples/add_volume.py
sofroniewn: see comment bellow - 

> the reason i renamed is that scale is a property of all our layers - i'll probably move the keyword arg to the base layer (along with exposing translate), in prepartation for merging the volume layer with the image layer. at that point i felt like it was confusing to have spacing for non-volumetric stuff, whereas scale was clear regardless

napari/layers/base/base.py
sofroniewn: fixed with `curr_range`

napari/layers/_vispy/vispy_image_layer.py
sofroniewn: I've moved this back to the top level - I don't have strong feelings about it. We've already pulled qt layer files from the layers folder into the `_qt` folder so it's fine to have these separate too. We could have a separate folder called `_views` that contains both `_qt` and `_vispy`, but maybe that's making things too nested. Some of this stuff I think we can revisit when we start improving our docs and thinking about where everything lives and is grouped there

napari/layers/_vispy/util.py
sofroniewn: replaced with a dictionary - both here and in the `_qt/utils.py` where a similar pattern appeared - let me know if that's what you had in mind

sofroniewn:
@jni I have made your requested changes including moving the `_vispy` folder and replacing the globals with a dict. I have not renamed `spacing` for reasons in my comment above.

This is ready for re-review

napari/layers/points/points.py
sofroniewn: yeah, let's fix in another PR

napari/_qt/util.py
```diff
@@ -25,12 +43,8 @@ def create_qt_properties(layer):
...
             Qt propetry widget
     """
-    name = 'Qt' + type(layer).__name__ + 'Properties'
-    try:
-        qt_props = globals()[name]
-        properties = qt_props(layer)
-    except KeyError:
-        properties = QtLayerProperties(layer)
+    name = type(layer).__name__
+    properties = layer_to_properties[name](layer)
```
kne42: I would make the keys layers instead of their names

napari/_qt/util.py
```diff
@@ -49,11 +63,7 @@ def create_qt_controls(layer):
...
             Qt controls widget
     """
-    name = 'Qt' + type(layer).__name__ + 'Controls'
-    try:
-        qt_controls = globals()[name]
-        controls = qt_controls(layer)
-    except KeyError:
-        controls = QtLayerControls(layer)
+    name = type(layer).__name__
+    controls = layer_to_controls[name](layer)
```
kne42: see above

napari/_vispy/util.py
```diff
@@ -0,0 +1,37 @@
...
+        layer : napari.layers._base_layer.Layer
+            Layer that needs its propetry widget created.
+
+    Returns
+    ----------
+        visual : vispy.scene.visuals.VisualNode
+            Vispy visual node
+    """
+    name = type(layer).__name__
+    visual = layer_to_visual[name](layer)
```
kne42: see above

napari/_qt/util.py
kne42: additionally, you seem to re-use this functionality many times so it may be better just to create a more general function

kne42: something that could be cool that you might consider is using `__init_subclass__` in combination with `__new__`:
```python
class QtLayerProperty:
    _layer_properties = {}

    def __new__(cls, layer):
        if cls is QtLayerProperty and type(layer) in QtLayerProperty._layer_properties:
            cls = QtLayerProperty._layer_properties[type(layer)]
        return object.__new__(cls)

    def __init_subclass__(cls, layer=None, **kwargs):
        if issubclass(layer, Layer):
            QtLayerProperty._layer_properties[layer] = cls
        
        super().__init_subclass__(**kwargs)

class QtImageProperty(QtLayerProperty, layer=Image):
    ...

>>> QtLayerProperty(viewer.layers['Image 0'])
QtImageProperty(...)
```

@sofroniewn @jni thoughts?

napari/layers/base/base.py
```diff
@@ -4,20 +4,19 @@
...
 from ...components import Dims
-from ...util.event import Event
+from ...util.event import EmitterGroup, Event
 from ...util.keybindings import KeymapMixin
-from ._visual_wrapper import VisualWrapper
+from ...util.status_messages import status_format
 
 
-class Layer(VisualWrapper, KeymapMixin, ABC):
+class Layer(KeymapMixin, ABC):
```
kne42, line 15: check if we use abstract methods any more. if not, you can remove `ABC`

kne42 (commented):
only got thru about half before train arrived, but here you go

napari/_qt/util.py
sofroniewn: this look very cool - but can you make this in a follow up PR instead?

napari/layers/base/base.py
sofroniewn: we do - `data` is still an abstract method

napari/_qt/util.py
kne42: could you at least address the first comment then?

sofroniewn: i've switched over to making the keys layers - good suggestion. i also do like the proposed scheme a lot too - it would mean we don't need these weird convertors in utils. if @jni wants i can try and add now too, but also waiting to a follow-up PR is fine for me too as I said above.

jni (approved):
All good from me! =)

napari/_qt/util.py
sofroniewn: @kne42 I tired making this change myself, but ran into errors (probably easily fixable) but I don't fully understand it, so I'll leave it to you!

royerloic:
> I've now started exploring converting the `Labels` layer. One thing that's coming up is where to put the mouse functions? Right now I'm moving them over the the `_vispy` folder but there's a lot of code that interacts with the models inside them - maybe it would be nice for them to have their own folder like the `keybindings.py`, we could have `mousebindings.py` @kne42, but that can wait till a followup PR.
> 
> I've also moved the `_vispy` folder inside `layers` as it's really a folder with the layer specific vispy code and not our backport or anything else

Very good idea to apply the same ideas from keybindings to a new mousebindings.
This can proceed step by step. Most important is first the separation... 
and then having a good centralised abstraction layer for dealing with mouse events.

PR #487, opened 2019-08-20
Fix keybindings

sofroniewn:
# Description
This PR fixes a bug in the up / down keybindings introduced with the dims refactor. It also adds some new keybindings for features created in the dims refactor.

In general it would be good to write tests to test all our keybindings. @kne42 is there a nice automated way to do this? If so can you do that in a follow-up PR, otherwise we'll just have to test every one. Note here that I'm just suggesting testing the function defined within the keybinding, not the actual registering of the keypress event, but I'm open to anything.

### DIFF
--- a/napari/_qt/qt_dims.py
+++ b/napari/_qt/qt_dims.py
@@ -42,6 +42,7 @@ class QtDims(QWidget):
 
         # list of sliders
         self.sliders = []
+        # True / False if slider is or is not displayed
         self._displayed_sliders = []
 
         self.last_used = None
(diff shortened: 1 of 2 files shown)


kne42:
@sofroniewn What kind of automation are you suggesting? If you're just checking that the function does what it's supposed to do, you'll have to create a normal test suite.

sofroniewn:
@kne42 I'm not sure, maybe something that went through, looked at all our keybindings and then just tried executing each one of them to confirm they could run, but you're right much better to just write a test for each function where we confirm that the actual changes we expected to happen happened. maybe those can come in follow up PRs though ....

kne42 (approved):
yeah you should probably do that instead

PR #488, opened 2019-08-21
layer views

kne42:
# Description
Non-instantiable mix-in class for views corresponding to an aspect of a layer.

Direct subclasses of ``LayerView`` (ones that explicitly list it as a base) are considered **base classes**.

Classes that inherit from **base classes** can specify the ``layer`` class definition keyword argument to associate themselves with a specific layer type.

When a **base class** is initialized with a layer of that type, it will return an instance of the class associated with that layer type.

```python
>>> class Foo(LayerView): ...
>>> class ImageFoo(Foo, layer=Image): ...
>>> foo = Foo(Image(...))
>>> type(foo).__name__
'ImageFoo'
```

## Changes
- create layer view class
- adapt vispy views to use layer view class
- adapt qt properties to use layer view class
- adapt qt controls to use layer view class
- replace all usages of `create_*` functions with layer view base classes directly

# References

see https://github.com/napari/napari/pull/485#discussion_r315848620


sofroniewn (approved):
Looks great to me! Thanks. I'll leave for @jni to have a look before merging.

jni (approved):
Jeez. Working with @kne42 is a great way to learn about crazy Python features! :sweat_smile: Very cool stuff! @kne42, do you have a good source for this pattern? This is all I could find about it, it's not quite detailed enough for my tastes:

http://martyalchin.com/2011/jan/20/class-level-keyword-arguments/

kne42:
No, I haven't run into any patterns like this.

I guess the closest is as defined in [PEP 487](https://www.python.org/dev/peps/pep-0487/#subclass-registration):
```python
class PluginBase:
    subclasses = []

    def __init_subclass__(cls, **kwargs):
        super().__init_subclass__(**kwargs)
        cls.subclasses.append(cls)
```

in combination with [this blog's examples](https://howto.lintel.in/python-__new__-magic-method-explained/) (see [Python's documentation on `__new__`](https://docs.python.org/3/reference/datamodel.html#object.__new__) for more information).

PR #22, opened 2018-10-27
Implement workaround for opacity-order bug

kne42:
Hacky workaround to fix opacity until the bug as detailed in vispy/vispy#1541 is fixed.

See:
- #12 
- https://github.com/Napari/napari-gui/pull/21#issuecomment-433547202

PR #493, opened 2019-08-23
Integrate 3D rendering with layers

sofroniewn:
# Description
This PR integrates the image and volume layer so you can toggle easily between the two of them without changing layer type. It does this via the setting of the `layer.dims.ndisplay` property, for example see the gif below on a 4D dataset:

change_display

To add a volume now you use `add_image(data, ndisplay=3)`. We could add back in a convenience method `add_volume` that does this automatically.

Right now volume display functionality has not been ported to the labels layer or pyramid layer. Functionality is also untested for the points, vectors, shapes layers.

Support for mixed rendering of 2D slices and 3D volumes together is also not supported yet, though I may work on that soon

@pranathivemuri @royerloic soon you should give this a go and let me know if everything still works as expected, but you should probably wait till the issue mentioned below is fixed.

This PR also tried to incorporate a [fix](https://github.com/kne42/napari/commit/26dd02f30b35ab7dbe6f0b0f87cd52a20db424d3) from @kne42 for a current bug on master, but i'm not able to have both napari and the tests working on my computer. Not sure how we'll figure that out or what effect it will have on this PR, but I needed to do something to make it all work.

## Type of change

# How has this been tested?

- [x] moved over tests from volume to image.

## Final checklist:


sofroniewn:
This PR is still in progress, but coming along well. It now extends integrated 3D support to labels, points, and vectors. Not yet supported are pyramid and shapes.

It also adds a custom keypress `ctrl-Y` that toggles `2D` / `3D` mode. there are a bunch of outstanding issues, like its probably worth disabling the paintbrush / point selection / addition tools in 3D right now as they don't work.

Also right now only one global 3D or 2D viewing mode is supported and you cannot mix 2D and 3D viewing (or unsupported things will happen). If you have a fundamentally dims=2 layer and try to enter 3D mode now it will crash. We should probably intercept that with a warning instead and prevent that transition, and say something helpful like with 2D layers present 3D viewing is not supported. Also given that maybe `ndisplay` shouldn't be a keyword arg to any of our layers, and you have to set it on the viewer only to enter 3D mode.

Finally supporting the pyramid and shapes layers can probably come in follow-up PRs.

Here's what it looks like to switch back and forth between 2D and 3D viewing for vectors and points:

3D_vectors_display

3D_points_display

Note there is still a vispy bug present with the shapes of the points in 3D, I know i've seen issues present on that one, but I havn't looked into it yet

sofroniewn:
@jni @kne42 this PR is now ready for review

It merges the volume and image layers and adds support for 3D rendering to image, labels, vectors, and points. 3D rendering is a global property that applies to all layers settable using `viewer.dims.ndisplay = 3` or on construction with `Viewer(ndisplay=3)`. Right now you cannot mix 2D and 3D rendering, and if you have a layer that is fundamentally two dimensional trying to enter 3D viewing will cause a crash.

Right now in 3D rendering mode the interactivity of the labels and points layers are not disabled, but have also not been tested or are expected to perform normally. There is a known vispy issue with the sizes of the 3D points, which I can look into fixing.

This PR does not extend 3D rendering support to the shapes or pyramid layers, but I think that can come later

examples/nD_vectors.py
```diff
@@ -37,7 +37,7 @@
     pos[:, 1, 1] = 2 * radius_space * np.sin(phi_space)
 
     # add the broadcast vectors
-    layer = viewer.add_vectors(pos, edge_width=0.4, name='broadcast vectors')
+    #layer = viewer.add_vectors(pos, edge_width=0.4, name='broadcast vectors')
```
jni: ?

napari/_qt/layers/qt_image_layer.py
```diff
@@ -1,9 +1,9 @@
 from qtpy.QtWidgets import QHBoxLayout
-from . import QVRangeSlider
+from .. import QVRangeSlider
 from .qt_base_layer import QtLayerControls, QtLayerProperties
 from qtpy.QtCore import Qt
 from qtpy.QtWidgets import QLabel, QComboBox
-from ..layers.image._constants import Interpolation
+from ...layers.image._constants import Interpolation, Rendering
```
jni, line 6: Not super in love with the extra nesting, but could be fine...

napari/_qt/qt_viewer.py
```diff
@@ -149,24 +149,28 @@ def _add_layer(self, event):
         vispy_layer = create_vispy_visual(layer)
         vispy_layer.camera = self.view.camera
         vispy_layer.node.parent = self.view.scene
-        vispy_layer._order = -len(layers)
+        if self.viewer.dims.ndisplay == 2:
+            vispy_layer.order = -len(layers)
+        else:
+            vispy_layer.order = len(layers)
```
jni, line 157: What's the story here...? I don't understand why it's different for different dimensionality.

napari/_vispy/vispy_image_layer.py
```diff
@@ -1,35 +1,92 @@
...
+        )
 
-        self._on_interpolation_change()
-        self._on_colormap_change()
-        self._on_clim_change()
-        self._on_data_change()
+        self.reset()
+
+    def _on_display_change(self):
+        if self.layer.dims.ndisplay == 2 and type(self.node) == VolumeNode:
```
jni:
```suggestion
        if self.layer.dims.ndisplay == 2 and isinstance(self.node, VolumeNode):
```

napari/_vispy/vispy_image_layer.py
```diff
@@ -1,35 +1,92 @@
...
+            self.node.parent = None
+
+            self.node = ImageNode(None, method='auto')
+            self.node.parent = parent
+            self.order = order
+            self.layer._update_dims()
+            self.layer._set_view_slice()
+            self.reset()
+
+        elif self.layer.dims.ndisplay == 3 and type(self.node) == ImageNode:
```
jni: isinstance

napari/_vispy/vispy_points_layer.py
```diff
@@ -24,8 +29,38 @@ def __init__(self, layer):
...
-        self._on_data_change()
-        self._on_highlight_change()
+        self.layer.dims.events.display.connect(
+            lambda e: self._on_display_change()
+        )
+
+        self.reset()
+
+    def _on_display_change(self):
+        if self.layer.dims.ndisplay == 2 and type(self.node) == Markers:
```
jni: isinstance

napari/_vispy/vispy_points_layer.py
```diff
@@ -24,8 +29,38 @@ def __init__(self, layer):
...
+            self.node.parent = None
+
+            self.node = Compound([Line(), Markers(), Markers()])
+            self.node.parent = parent
+            self.order = order
+            self.layer._update_dims()
+            self.layer._set_view_slice()
+            self.reset()
+
+        elif self.layer.dims.ndisplay == 3 and type(self.node) == Compound:
```
jni: isinstance

napari/_vispy/vispy_points_layer.py
```diff
@@ -24,8 +29,38 @@ def __init__(self, layer):
...
+            order = abs(copy(self.node.order))
+            self.node.transforms = ChainTransform()
+            self.node.parent = None
+
+            self.node = Markers()
+            self.node.parent = parent
+            self.order = order
+            self.layer._update_dims()
+            self.layer._set_view_slice()
+            self.reset()
```
jni: For these two blocks, is it possible to reduce code duplication by pulling the common code out of the if statements? e.g. the first two lines (`parent =` and `order =`) are the same

jni: And so are the last few, `self.node.parent = ` onwards

napari/_vispy/vispy_points_layer.py
```diff
@@ -44,18 +79,39 @@ def _on_data_change(self):
...
+        else:
+            self.node.set_data(
+                data[::-1, ::-1] + 0.5,
+                size=size[::-1],
+                edge_width=self.layer.edge_width,
+                symbol=self.layer.symbol,
+                edge_color=edge_color,
+                face_color=face_color,
+                scaling=True,
+            )
```
jni: again, a lot of shared code... Maybe:

```python
    set_data = self.node._subvisuals[2].set_data
else:
    set_data = self.node.set_data
set_data(...)
```

jni: Or:

```python
kwargs = {...}
    self.node._subvisuals[2].set_data(**kwargs)
else:
    self.node.set_data(**kwargs)
```

examples/nD_vectors.py
kne42: ?

napari/_qt/qt_viewer.py
kne42, line 157: what's the reasoning behind this?

napari/_vispy/vispy_image_layer.py
```diff
@@ -1,35 +1,92 @@
...
-        self._on_interpolation_change()
-        self._on_colormap_change()
-        self._on_clim_change()
-        self._on_data_change()
+        self.reset()
+
+    def _on_display_change(self):
+        if self.layer.dims.ndisplay == 2 and type(self.node) == VolumeNode:
+            parent = self.node.parent
+            order = abs(copy(self.node.order))
```
kne42: why the copy here? isn't `order` an integer?

napari/_vispy/vispy_image_layer.py
kne42: generally use `isinstance` instead

napari/_vispy/vispy_labels_layer.py
```diff
@@ -1,18 +1,6 @@
...
-        self.node.cmap = self.layer.colormap[1]
-        self.node.clim = [0.0, 1.0]
-        self._on_data_change()
-
-    def _on_data_change(self):
-        self.node._need_colortransform_update = True
-        image = self.layer._raw_to_displayed(self.layer._data_view)
-        self.node.set_data(image)
-        self.node.update()
+class VispyLabelsLayer(VispyImageLayer):
```
kne42: what's the point to this? why not just delete it all?

napari/layers/labels/labels.py
```diff
@@ -491,8 +514,9 @@ def to_xml_list(self):
             List of a single xml element specifying the currently viewed image
             as a png according to the svg specification.
         """
-        image = self._raw_to_displayed(self._data_view)
-        mapped_image = (self.colormap[1].map(image) * 255).astype('uint8')
+        mapped_image = (self.colormap[1].map(self._data_view) * 255).astype(
+            'uint8'
```
kne42: i recommend to use `np.uint8` instead of a string

napari/components/viewer_model.py
```diff
@@ -28,7 +28,7 @@ class ViewerModel(KeymapMixin):
 
     themes = palettes
 
-    def __init__(self, title='napari'):
+    def __init__(self, title='napari', ndisplay=2):
```
kne42: could we maybe make this take in the specific dimensions we want displayed too? (e.g. `display=2` is equivalent to `display=(-2, -1)`)

kne42 (approved):

napari/components/viewer_model.py
```diff
@@ -368,126 +377,73 @@ def add_image(
...
 
         Returns
         -------
         layer : :class:`napari.layers.Image`
             The newly-created image layer.
         """
         layer = layers.Image(
-            image,
-            metadata=metadata,
+            data,
```
jni, line 407: Can we add a `*,` after `data`?

napari/components/viewer_model.py
```diff
@@ -738,31 +700,37 @@ def add_shapes(
...
             visible=visible,
-            name=name,
         )
         self.add_layer(layer)
         return layer
 
     def add_vectors(
         self,
-        vectors,
+        data,
```
jni, line 736: I'm not sure I love renaming all the input arguments, but I don't feel super strongly about it...

jni: Having said that, in many cases, the parameter name is self-documenting (`image` vs `coords`), which is nice, rather than having to scroll down to the argument description.

napari/layers/base/base.py
```diff
@@ -264,13 +283,34 @@ def _update_dims(self):
...
+            self._position = self._position[-ndisplay:]
+        elif len(self.position) < ndisplay:
+            self._position = (0,) * (ndisplay - len(self.position)) + tuple(
+                self.position
+            )
+
+        if len(self.scale) > ndim:
+            self._scale = self._scale[-ndim:]
+        elif len(self.scale) < ndim:
+            self._scale = (1,) * (ndim - len(self.scale)) + tuple(self.scale)
```
jni, line 301: Can you add comments explaining the logic here? It seems finicky...

napari/layers/image/image.py
```diff
@@ -175,15 +192,20 @@ def data(self, data):
...
         if self.multichannel:
             shape = self.data.shape[:-1]
         else:
             shape = self.data.shape
-        return tuple((0, m, 1) for m in shape)
+
+        if self.dims.ndisplay == 3:
+            scale = self.scale
+        else:
+            scale = 1
```
jni: I don't think this code is right...? Shouldn't scale be set correctly regardless of dimension?

jni: (This is important eg for xz views?)

napari/layers/labels/labels.py
```diff
@@ -99,33 +103,49 @@ class Labels(Layer):
...
 
-        self._data = labels
-        self._data_view = np.zeros((1, 1))
-        self.metadata = metadata or {}
+        self._data = data
+        self._data_labels = np.zeros((1,) * self.dims.ndisplay)
+        self._data_view = np.zeros((1,) * self.dims.ndisplay)
+        self.clim = [0.0, 1.0]
+        self.interpolation = 'nearest'
+        self.rendering = 'mip'
```
jni, line 148: How does this look in 3D? Is it usable?

napari/layers/vectors/vectors_util.py
```diff
@@ -73,27 +73,67 @@ def generate_vector_meshes(vectors, width, length):
...
+    ----------
+    vectors : (N, 2, D) array
+        A list of N vectors with start point and projections of the vector
+        in D dimensions, where D is 2 or 3.
+    width : float
+        width of the line to be drawn
+    length : float
+        length multiplier of the line to be drawn
+    p : 3-tuple, optional
+        orthogonal vector for segment calculation in 3D.
```
jni, line 120: wow. =)

napari/tests/test_nD.py
```diff
@@ -41,7 +41,8 @@ def test_nD_volume(qtbot):
 
     np.random.seed(0)
     data = np.random.random((6, 10, 15, 20))
-    viewer.add_volume(data)
+    viewer.add_image(data)
+    viewer.dims.ndisplay = 3
```
jni, line 45: Can we maybe add another test where the viewer is directly instantiated with ndim=3?

jni (approved):
Very cool stuff! I've raised issues in a couple of places but in general this is super great! =D

examples/nD_vectors.py
sofroniewn: dropping from example as it was a 2D layer (code still runs fine, just less useful)

napari/_qt/qt_viewer.py
sofroniewn: unfortunately the vispy bug mentioned in #22 and effects the volume and image layers differently - we can't use the `z_index` trick there too as the `z` dimension has actual meaning, it would be really good if vispy fixed that bug, would simplify things for us. I'll add a comment

napari/_vispy/vispy_image_layer.py
sofroniewn: removed

napari/_vispy/vispy_image_layer.py
sofroniewn: fixed

napari/_vispy/vispy_labels_layer.py
sofroniewn: done, was leftover from when i was trying out the new class based approach for generating views instead of the dictionary approach

napari/layers/labels/labels.py
sofroniewn: done

napari/components/viewer_model.py
sofroniewn: i've added a new keyword argument `order` that does this so you can now use them both together

examples/nD_vectors.py
sofroniewn: removed

sofroniewn: see note above

napari/_qt/layers/qt_image_layer.py
sofroniewn: ah yes ... can revert in the future if you like, i like the grouping of the qt layer stuff together, but i appreciate its undesirability

napari/_qt/qt_viewer.py
sofroniewn: yup, added a comment, see also response to @kne42 it's due to a vispy bug described in #22 that we should really try and fix. maybe @kne42 want's to look into that again, I know she started an issue on vispy but it didnt get resolved (instead we found a workaround that isn't so great for the 3D rendering case)

napari/_vispy/vispy_image_layer.py
sofroniewn: done

napari/_vispy/vispy_image_layer.py
sofroniewn: done

napari/_vispy/vispy_points_layer.py
sofroniewn: done

napari/_vispy/vispy_points_layer.py
sofroniewn: done

napari/_vispy/vispy_points_layer.py
sofroniewn: the issue is i only want that code to run if either of the if statements is true - otherwise just ignore, so we'd have to add in another if statement. if you know of a clean way to do this, i'm happy to make the change as i agree it's not nice to have the duplication

napari/_vispy/vispy_points_layer.py
sofroniewn: i went with the first suggestion, nice one

napari/components/viewer_model.py
sofroniewn: we could name everything according to it's layer type - so parameters would be
`image`, `labels`, `points`, `pyramid`, `vectors`, and `shapes` instead of `data` for all of them?

that's fine with me, i didn't like `coords` when we use `points` in most places. if you prefer this let me know and we can fix

napari/layers/labels/labels.py
sofroniewn: yeah, looks pretty good

napari/layers/vectors/vectors_util.py
sofroniewn: :-)

napari/tests/test_nD.py
sofroniewn: test added with `ndisplay=3` and a custom order

napari/layers/base/base.py
sofroniewn: i've added a comment describing it - it's about how to trim / pad the properties when the dimensionality shrinks / grows

napari/layers/image/image.py
sofroniewn: this was not needed, i've removed it

napari/components/viewer_model.py
sofroniewn: isn't that done in the def not the function call? i feel like we have it above and don't need it here, but maybe i don't understand

sofroniewn:
@jni @kne42 I've addressed your requested changes. @jni there are one or two minor ones that i've left comments on that i can change if you respond (for example the input parameters being called things like `image` and `labels` instead of `data`), that tricky duplicate code in the `if` statements, and the nestedness of the qt directory.

None of these are major and all can be addressed in follow up PRs, so i'll probably just merge at some point soonish

napari/_vispy/vispy_points_layer.py
jni: @sofroniewn doesn't the node type change in sync with the ndisplay, *and* don't these two clauses encompass all possibilities for the points layer? I'm just trying to understand what simple clause could be added for the whole function to run...

napari/_qt/qt_viewer.py
jni: :+1:

napari/components/viewer_model.py
jni: Erm, yes, sorry, totally misread this. :man_facepalming:

jni:
@sofroniewn left a response about the if blocks, otherwise all good from me, feel free to merge at your convenience! =)

napari/_vispy/vispy_points_layer.py
sofroniewn: this is fixed, agreed they shouldn't have been running

sofroniewn:
I've made the final requested changes, and will merge when the tests pass

PR #494, opened 2019-08-26
revert "layer views (#488)"

### LINKED ISSUE #492: master broken: missing node argument after automatic layer view creation
## 🐛 Bug

All the examples on Master are currently broken - the error I get on `add_image.py` initially is 

```python
  File "/Users/nicholassofroniew/Github/napari/napari/_qt/qt_viewer.py", line 149, in _add_layer
    vispy_layer = VispyBaseLayer(layer)
  File "/Users/nicholassofroniew/Github/napari/napari/util/layer_view.py", line 52, in __new__
    return super().__new__(cls)
TypeError: Can't instantiate abstract class VispyBaseLayer with abstract methods _on_data_change
```

if i make this no longer an abstract method, the error i get is  

```python
  File "/Users/nicholassofroniew/Github/napari/napari/_qt/qt_viewer.py", line 149, in _add_layer
    vispy_layer = VispyBaseLayer(layer)
TypeError: __init__() missing 1 required positional argument: 'node'
```
which is because the `VispyBaseLayer` takes two arguments the layer and the node which doesn't seem to work out well with the approach of #488. We could drop the `node` argument and append it inside the layer specific `Vispy` views, but I defer to @kne42 on how to fix this. 

Unfortunately this was not caught by any of our existing tests, which I find really confusing (they all pass locally for me even though I get this error). Maybe @kne42 can also figure out some tests to add that will catch this too!

kne42:
reverts #488

fixes #492


sofroniewn (approved):
Looks good to me, I have verified this works

PR #496, opened 2019-08-28
[WIP] add pyinstaller bundled app

sofroniewn:
# Description
This PR attempts to add support for generating a bundled app with pyinstaller. Thanks very much for @csweaver for getting us started as discussed in #376. 

This PR adds a spec file (that can be autogenerate with `gui-spec`) but needs to be edited to add `vispy` stuff and some of our resources.

I've also had to change a couple of our imports of `.qss` and `.txt` files to make sure the paths work both normally and with pyinstaller.

Unfortunately when I try launching the bundled app I get the following error: 

```
WARNING: Could not find the Qt platform plugin "cocoa" in ""
WARNING: This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.
```

I know @AhmetCanSolak you've done a lot of work with pyinstaller lately, can you jump in and give me a hand. Any help would be much appreciated.

We can also rename `gui-spec`, `gui-build`, `install-gui` as appropriate.


csweaver:
Possibly related to the cocoa error: 
https://github.com/pyinstaller/pyinstaller/pull/3233

AhmetCanSolak:
Hi @sofroniewn , i will check the branch and try to reproduce your observations later this week. My initial concern is using makefile, i think we should be using `CMake`. We will need to run pyinstaller on different operating systems to get apps. Related to this, we can have a build folder structure where we have separated folders for different operating systems as our bundling setup possibly will become OS specific. Like:
```
- build\
-- linux\
-- osx\
-- win\
```

jni:
I vaguely feel that this should go in a separate repo, e.g. `napari-build` or `napari-distribution`, as the CellProfiler team does. Not fixated on this opinion though, but it just feels like this adds a lot of "infrastructure" files.

AhmetCanSolak:
@jni i am not disturbed with having those under a build folder in root but also open to having a separate repo. What others think on this?

sofroniewn:
I prefer keeping it in the same repo (maybe putting it in some folder like @AhmetCanSolak suggests) as those who don't know about such things will easily ignore it, and for those that are interested it will be much easier to keep track of / find it if it is here

csweaver:
Re Cocoa bug:
So my exploration has identified that the bug happens here `app = QApplication.instance() or QApplication(sys.argv)` specifically this clause`QApplication(sys.argv)`

The reason cellxgene (the project I'm using pyinstaller for) was not affected is that our application is a cefpython application not a QApplication and so doesn't hit the same bug.

It looks like many others have encountered this problem, but that most of them are using pyqt5. (https://github.com/pyinstaller/pyinstaller/issues/2924)

It looks like the warning is being generated in the qt library as I couldn't find anything matching the text in the pyside2 library. The plugin path is getting set correctly. I added `print("environ", os.environ['QT_PLUGIN_PATH'])` on line 19 of event_loop.py to check. There may be another env variable though. You can see this being set at runtime via `PyInstaller/loader/rthooks/pyi_rth_pyside2.py`

The `libqcocoa.dylib` is being included in PySide2 folder inside the built app, so everything seems like it should be correct. 

Anyways, I've done what I could. I hope this gives you some ideas as to where to go next in this bug busting.

AhmetCanSolak:
Made a small commit to demonstrate folder structure i had in mind. Please provide feedback, accordingly we can move with this or another way.

PS: right now it bundles but not running properly, we are getting a dask error. We probably need to discuss further on which dependencies we definitely want or not.

AhmetCanSolak:
Now working on mac, will do windows and linux next.

tlambert03:
Hey  @sofroniewn and @AhmetCanSolak,  I have working pyinstaller builds for windows and linux (and fixed some hidden issues in the osx build) here: https://github.com/tlambert03/napari/tree/pr/sofroniewn/496
there are also multiple updates to the pipeline (updated data finding and multiple hooks).  I don't want to push it directly to this PR though without your go-ahead...
will add more detail soon

sofroniewn:
@tlambert03 this is great, thanks for picking this up. Fine by me to push directly to the branch. @AhmetCanSolak do you have any updates here? I think it would be nice to start providing a bundled app alongside our releases

AhmetCanSolak:
It is great to hear from @tlambert03 about this. Yes @sofroniewn indeed i was silently working on this :D i have osx and linux working but haven't made it work on windows yet. @tlambert03 feel free to push directly if you have the specs and hooks working for both. If not i will push it once im done. Thanks!

tlambert03:
I figured I would just push to this branch because this is working pretty nicely now, and I'm confident you're going to want a lot of this stuff.  If there are objections, we can always easily revert back to 3b0646d, which is where this was before I started working on it.

Here's what's been added:

- merged all changes from master since the last activity
- builds are now working on all three platforms
- unified cross-platform builds into a single file at `bundle/napari.spec`
- standardized resource locations (including splash screen, colormaps, stylesheets, etc...) across PyInstaller and dev environments with a new `utils.misc.absolute_resource` function following [this strategy](https://stackoverflow.com/a/13790741/1631624).
- created ico/icns files to add app icons for windows and mac, and version info (when getting app properties in the context menu)
- updated build.sh and build.bat scripts for idiosyncrasies of each environment
- created module-specific hooks (instead of adding `datas` for 3rd party packages directly in our spec file), for things like vispy, dask, skimage.  Ultimately, we should consider submitting these hooks to pyinstaller.  For instance, [the `skimage.io` hook](https://github.com/pyinstaller/pyinstaller/blob/develop/PyInstaller/hooks/hook-skimage.io.py) doesn't actually add all the plugins needed (at least in my tests),  so the existing mac build, while "successful", couldn't actually load Tiffs.  That's fixed here.
- added continuous integration builds including packaging into a mac DMG.  ultimately, we'll make this only happen on tagged commits on master (need to uncomment that line before merging), but here are examples for [mac](https://cirrus-ci.com/task/5674072234524672), [linux](https://cirrus-ci.com/task/5111122281103360), and [windows](https://cirrus-ci.com/task/6237022187945984).  (windows and linux need a little polish on the installer still, and I think there may be a portability issue with linux, but the mac dmg file works well)
- I also made everything portable, so it doesn't matter what directory you're in when you call it. but from the root directory you can just call `make bundle` for mac/linux or `call .\bundle\build.bat` for windows 
- other miscellaneous changes that were preventing successful functioning

pyinstaller 3.5 generally works everywhere for PyQt5, but their support for pyside2 is kinda crappy... so there are few OS-specific workarounds or `pip installs` in here.  (this is presumably why it was previously pinned to 3.4 for mac?)

@AhmetCanSolak, if you've continued to branch off of your last commit without pushing here, you will likely have a lot of conflicts.  So if there is any functionality you have that is missing here, I would probably branch your commits before trying to merge this with your work.

tlambert03:
would be good to start having people try the various different builds on different computers.  I definitely found a few unexpected things when taking a bundle from one computer to another.  And the various `hiddenimports` of our dependencies make it possible that we will be catching small bugs for quite a while as we use them . You can get them in the "Bundle linux/osx/windows" tasks in the CI above.

tlambert03:
added a windows installer exe using Inno Setup.  It's working pretty well for me (and gives a nice native feeling, adding optional start menu items, desktop shortcuts, and ability to uninstall in "Programs and Features").  If anyone has windows, you can download an example installer from this build:
https://cirrus-ci.com/task/6095440604299264

tlambert03:
I just added a `snapcraft.yaml` for bundling napari in [snap format](https://snapcraft.io/docs/snap-format) which I think should allow linux users to install via the ubuntu software center.  @jni, or other linux users, if you have a minute you should be able to install (a version of) napari with `sudo snap install --edge napari --devmode`.  If that works for you, we can polish out the edges, add icons, etc... and make it look nice in the store.

for what it's worth, the snap distribution doesn't use pyinstaller at all, it was easier to just do it directly [as described here](https://snapcraft.io/docs/python-apps).  I still get a couple warnings in the terminal at runtime, but no obvious errors yet.

With that, we now have bundled packages with platform-native distribution methods (.dmg for mac, `setup.exe` installer for windows, `snap install` for linux) for all three platforms

jni:
@tlambert03 all of this is amazing! I have only stolen moments for the next few days but I'll try to test all this more thoroughly next week.

AhmetCanSolak:
Hi @tlambert03. I think it is great what you have done on this PR. There seem to be small conflict. I love to review and merge this and update our documentation for users soon.

tlambert03:
I can clean up the conflict soon.  Would be nice to have a couple people try the last bundles on a few different systems.

I think there are a few downstream questions that remain (which can, perhaps, be left for another PR?).  Specifically... while the CI is currently building artifacts, it would be nice to figure out how to get those automatically added to github releases.  The API allows for it, but I need to know a bit more about the sequence of steps that @sofroniewn does for releases.  Until then we can always manually add them.

it would be icing (though obviously not strictly necessary), to have napari in the mac app store... as it would help encourage and facilitate updating.  That requires purchasing an apple developer account (I think $99/year).
for linux, getting it into the store is easy as long as our snapcraft build is working (which, currently, I think it's not working as smoothly as the pyinstaller build, but i'd like for people to try the `snap install` as described above)
for windows, the installer works great... the only unsatisfying thing is that the code is not signed, so users get that frightening message when they try to install it.  I know most people are used to that sort of thing, but there too, if someone wanted to buy a code signing cert for us, we could add that to give it a more polished feel.  I think those also run about $100/yr, and you can't use the OSX cert on windows.

sofroniewn:
> Would be nice to have a couple people try the last bundles on a few different systems.

yes - I'll try on mine, sorry havn't yet!!

> it would be nice to figure out how to get those automatically added to github releases

I'd love that, we can chat about that if easier - though I try to follow the steps in https://github.com/napari/napari/blob/master/docs/RELEASE.md as best I can

> requires purchasing an apple developer account

CZI might already have one / if not I can try and get and expense to them

> buy a code signing cert for us

similarly - I can look into it

sofroniewn:
Trying this now. The mac download works fine on my laptop - though the unverified developer warning as very bad - you really have to jump through hoops in your security settings to get napari to open. As noted above we should look into this further.

One thing is in the napari info tab is see `napari: 0+unknown`, would be good if that showed the correct version number corresponding to the latest release.

I'm trying the windows build on a fresh ec2 machine that I am accessing via remote desktop, and I get the following error message
image
I havn't installed python or anything else on this machine, just downloaded the windows exe from the first link here https://cirrus-ci.com/task/5674072234524672

The windows installer linked to above just hangs when I try and open it - could be because a weird `htm` got appended to the `exe` and I don't know how to remove (clearly I'm not much of a windows user)

image

I at least know how to remote desktop into ec2 machines, so I can do some more cross platform testing

tlambert03:
yeah I still need to figure out exactly how versioneer works to fix that version info problem :)

regarding windows on ec2... i don't expect remote desktop to work at all unfortunately, for all of the same reasons that remote vispy/openGL never works.

to directly download the exe installer, use [this link](https://api.cirrus-ci.com/v1/artifact/task/6095440604299264/napari/Output/napari_installer.exe) ... (just in case).  I tried that on a few windows machines and it always worked (though again, not with remote desktop).

sofroniewn:
We have decided to move away from pyinstaller instead favouring briefcase and other approaches, see #1289 and #1074 for discussion, as they will ship a "non-frozen" app that allows for pip install after bundling

PR #499, opened 2019-08-30
rename clim

### LINKED ISSUE #174: Change name of clims
From @jni in #172

> Side note, clims needs a name change

Can you elaborate here / do you have ideas on what name should be?

sofroniewn:
# Description
This PR renames `clim` to `contrast_limits` and closes #174. Does it look good @jni?

## Type of change

### DIFF
--- a/examples/add_pyramid.py
+++ b/examples/add_pyramid.py
@@ -22,4 +22,4 @@ with napari.gui_qt():
     viewer = napari.Viewer()
 
     # add image pyramid
-    viewer.add_pyramid(pyramid, contrast_limits_range=[0, 255])
+    viewer.add_pyramid(pyramid, contrast_limits=[0, 255])
--- a/examples/add_pyramid.py
+++ b/examples/add_pyramid.py
@@ -22,4 +22,4 @@ with napari.gui_qt():
     viewer = napari.Viewer()
 
     # add image pyramid
-    viewer.add_pyramid(pyramid, clim_range=[0, 255])
+    viewer.add_pyramid(pyramid, contrast_limits_range=[0, 255])
--- a/napari/_qt/layers/qt_image_layer.py
+++ b/napari/_qt/layers/qt_image_layer.py
@@ -15,20 +15,20 @@ class QtImageControls(QtLayerControls):
         )
 
         # Create contrast_limits slider
-        self.contrast_limitsSlider = QVRangeSlider(
+        self.contrastLimitsSlider = QVRangeSlider(
             slider_range=[0, 1, 0.0001], values=[0, 1], parent=self
         )
-        self.contrast_limitsSlider.setEmitWhileMoving(True)
-        self.contrast_limitsSlider.collapsable = False
-        self.contrast_limitsSlider.setEnabled(True)
+        self.contrastLimitsSlider.setEmitWhileMoving(True)
+        self.contrastLimitsSlider.collapsable = False
+        self.contrastLimitsSlider.setEnabled(True)
 
         layout = QHBoxLayout()
-        layout.addWidget(self.contrast_limitsSlider)
+        layout.addWidget(self.contrastLimitsSlider)
         layout.setContentsMargins(12, 15, 10, 10)
         self.setLayout(layout)
         self.setMouseTracking(True)
 
-        self.contrast_limitsSlider.rangeChanged.connect(
+        self.contrastLimitsSlider.rangeChanged.connect(
             self.contrast_limits_slider_changed
         )
         self.contrast_limits_slider_update()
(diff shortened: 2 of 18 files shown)


examples/add_pyramid.py
```diff
@@ -22,4 +22,4 @@
     viewer = napari.Viewer()
 
     # add image pyramid
-    viewer.add_pyramid(pyramid, clim_range=[0, 255])
+    viewer.add_pyramid(pyramid, contrast_limits_range=[0, 255])
```
jni: Should we perhaps set this to be just "contrast_limits" and not support setting them in the `add_*` calls? (ie by default they are always at the broadest extent)

napari/_qt/layers/qt_image_layer.py
```diff
@@ -10,42 +10,46 @@ class QtImageControls(QtLayerControls):
...
+        # Create contrast_limits slider
+        self.contrast_limitsSlider = QVRangeSlider(
             slider_range=[0, 1, 0.0001], values=[0, 1], parent=self
         )
-        self.climSlider.setEmitWhileMoving(True)
-        self.climSlider.collapsable = False
-        self.climSlider.setEnabled(True)
+        self.contrast_limitsSlider.setEmitWhileMoving(True)
+        self.contrast_limitsSlider.collapsable = False
+        self.contrast_limitsSlider.setEnabled(True)
```
jni: I'd call this `contrastLimitsSlider`. If there's anything worse than camelCase =P it's mixed_underscore_andCamelCase. =P

jni (approved):
Looks good to me! I just wonder whether we should just have `contrast_limits=` as the keyword argument, and have that define the range as well as the initial setting. I also made a style comment that I think you can push quickly and merge, so I won't hold up approval. =) Good stuff!

examples/add_pyramid.py
sofroniewn: @jni I dislike have both the contrast_limits and contrast_limits_range as keyword arguments - I think it will be confusing to new commers, so figuring a way to get rid of the range option sounds great.

Are you proposing that instead of using min / max we take the dtype range? So (0, 255) for uint8, (-255, 255) for int8, (0, 2^16-1) for uint16 etc and then 0-1 for floats? I just worry there will be too many times this does something weird, and that taking the min / max of the image right now is nicer.

The problem is just that for lazy loading of large arrays the min/max is expensive to compute. We had just the min/max of the first displayed frame at one point, but that can also be confusing.

Also what should the behavior be if someone does pass a contrasts_limits keyword arg? One option would just be to use these values as the range of the sliders too, but then you couldn’t make the clims more extreme ever. Another option would be to set them at the 30% and 70% point of the slider, possibly with the exception of locking 0 to 0. Thoughts?

sofroniewn: @jni I dislike have both the contrast_limits and contrast_limits_range as keyword arguments - I think it will be confusing to new commers, so figuring a way to get rid of the range option sounds great.

Are you proposing that instead of using min / max we take the dtype range? So (0, 255) for uint8, (-255, 255) for int8, (0, 2^16-1) for uint16 etc and then 0-1 for floats? I just worry there will be too many times this does something weird, and that taking the min / max of the image right now is nicer.

The problem is just that for lazy loading of large arrays the min/max is expensive to compute. We had just the min/max of the first displayed frame at one point, but that can also be confusing.

Also what should the behavior be if someone does pass a contrasts_limits keyword arg? One option would just be to use these values as the range of the sliders too, but then you couldn’t make the clims more extreme ever. Another option would be to set them at the 30% and 70% point of the slider, possibly with the exception of locking 0 to 0. Thoughts?

jni: @sofroniewn my proposal is to have a single `contrast_limits` keyword argument set to None by default. If it is set to anything else, then *that is the min and max*, and the sliders are at those positions to begin with. The user needs to move them manually or set them in the next line. It would be unsupported to set both `contrast_limits_range` and `contrast_limits` in that same line.

Does this make sense?

Incidentally, I do think we should deprecate using the min/max and use instead the data ranges and instead use the scikit-image ranges https://scikit-image.org/docs/dev/user_guide/data_types.html, with a way to override them manually. It's too easy to add a dask array and have the computation take forever, with no feedback to the user.

napari/_qt/layers/qt_image_layer.py
sofroniewn: done

examples/add_pyramid.py
sofroniewn: ok sounds reasonable - i will drop the `contrast_limits_range` keyword arg from this PR and make the default behavior `min/max` - we can decide about default behaviour later

you'll still be able to change `contrast_limits_range` by directly setting the property

PR #501, opened 2019-09-01
fix cursor position

### LINKED ISSUE #310: Cursor coordinates are off by 0.5

sofroniewn:
# Description
This PR fixes #310 where the cursor positions are off by 0.5, and extends support to all our layer types. See the issue for a detailed description of the current and desired behavior. The PR also adds a simple example `examples/cursor_position.py` that allows you to quickly visually check the verified behavior. When we get to adding tests with screenshots / gui interactivity we can verify this stuff better.

Here is a gif of the current desired behaviour (where you may need to look at the example input data to verify all is correct):

cursor_position

The labels layer paintbrush / fill also still works correctly, as does editing / adding shapes and points.

@jni can you please confirm that all functionality is as expected

(I also slipped in a bug fix for the contrast_limits_range that hadn't been caught from #499 )

## Type of change

# How has this been tested?

- [x] add `examples/cursor_position.py`

## Final checklist:

### DIFF
--- a/napari/_vispy/vispy_base_layer.py
+++ b/napari/_vispy/vispy_base_layer.py
@@ -173,12 +173,12 @@ class VispyBaseLayer(ABC):
         """
         if self.node.canvas is not None:
             transform = self.node.canvas.scene.node_transform(self.node)
-            position = transform.map(list(position))[
-                : len(self.layer.dims.displayed)
-            ]
-            # Offset position so that pixel center is at 0
-            position = position - 0.5
-            coords = tuple(position[::-1])
+            # Map and offset position so that pixel center is at 0
+            mapped_position = (
+                transform.map(list(position))[: len(self.layer.dims.displayed)]
+                - 0.5
+            )
+            coords = tuple(mapped_position[::-1])
         else:
             coords = (0,) * len(self.layer.dims.displayed)
         return coords
--- a/napari/_vispy/vispy_base_layer.py
+++ b/napari/_vispy/vispy_base_layer.py
@@ -176,6 +176,8 @@ class VispyBaseLayer(ABC):
             position = transform.map(list(position))[
                 : len(self.layer.dims.displayed)
             ]
+            # Offset position so that pixel center is at 0
+            position = position - 0.5
             coords = tuple(position[::-1])
         else:
             coords = (0,) * len(self.layer.dims.displayed)
--- a/napari/_vispy/vispy_shapes_layer.py
+++ b/napari/_vispy/vispy_shapes_layer.py
@@ -29,6 +29,8 @@ class VispyShapesLayer(VispyBaseLayer):
         faces = self.layer._data_view._mesh.displayed_triangles
         colors = self.layer._data_view._mesh.displayed_triangles_colors
         vertices = self.layer._data_view._mesh.vertices
+        if vertices is not None:
+            vertices = vertices + 0.5
 
         if len(faces) == 0:
             self.node._subvisuals[3].set_data(vertices=None, faces=None)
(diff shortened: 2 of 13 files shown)


napari/_vispy/vispy_base_layer.py
```diff
@@ -176,6 +176,8 @@ def _transform_position(self, position):
             position = transform.map(list(position))[
                 : len(self.layer.dims.displayed)
             ]
+            # Offset position so that pixel center is at 0
+            position = position - 0.5
```
jni: Maybe just add `- 0.5` to the line above? I generally dislike overwriting existing names.

napari/_vispy/vispy_shapes_layer.py
```diff
@@ -29,6 +29,8 @@ def _on_data_change(self):
         faces = self.layer._data_view._mesh.displayed_triangles
         colors = self.layer._data_view._mesh.displayed_triangles_colors
         vertices = self.layer._data_view._mesh.vertices
+        if vertices is not None:
+            vertices = vertices + 0.5
```
jni, line 33: I'm concerned that this is repeated in many places... Should this not be fixed higher upstream?

jni (approved):
@sofroniewn awesome! I'm so happy. =) I've left a couple of comments. Also, while playing with it locally, I noticed that I don't always see a brush size when painting labels, depending on the brush size. I don't know whether that is related to this PR or not, but worth checking out...

napari/_vispy/vispy_base_layer.py
sofroniewn: done

napari/_vispy/vispy_shapes_layer.py
sofroniewn: it would be a minor code change to vispy, but could cause some backwards compatibility issues for them. There might be a better way to do it that involves offsetting the image visual instead of all the mesh visuals, but this works for now

sofroniewn:
@jni I've made the minor fixes. The brush icon is working fine for me in the labels layer. I'll note that when you zoom out very far the brush disappears and a little plus shaped cross hairs appears. Is that what is happening to you, or are you seeing a normal cursor instead? Either way worth investigating, but likely independent of this PR.

I will merge once the tests pass

jni:
I was indeed getting the cross, but that also happens when you zoom *in* and the brush is too large, which is confusing. I suggest looking closely at the logic there. I also think that brush size should be an integer rather than a slider.

sofroniewn:
zooming in and the brush too large means that the brush edges are would be bigger than the canvas / screen and so we have to display the + otherwise you wouldn't see anything

i agree on the integer vs slider, can fix that at some point soon

napari/_vispy/vispy_shapes_layer.py
royerloic: This is such an ingrained way of doing things in OpenGl, I doubt that this is something that would be called a 'fix'. As we discussed, it is a convention. For our purpose it makes sense to have it 'fixed'. Please keep in mind that the drawback of this convention is that some parts of the image have negative coordinates... (for example, 75% of pixel at (0,0) has negative coordinates)

PR #503, opened 2019-09-02
add surface layer

### LINKED ISSUE #502: add Surface layer for easy 2D / 3D mesh rendering
## 🚀 Feature
Following this discussion on twitter - https://twitter.com/TrackingActions/status/1168040048664530949

I propose we add a new layer type `Surface` with a corresponding `add_surface` method that allows you to easily render a 2D or 3D mesh with certain colorings / shadings.

We can take inspiration for this from the MRI neuroimaging community that frequently wants to render surface plots of the brain - see here for an example API from a tool in that space https://nilearn.github.io/modules/generated/nilearn.plotting.plot_surf.html

Here there are 3 core input args - `surf_mesh` which is a pair or arrays, one with vertices and one with faces. This maps identically to vispy mesh visual. Then theres `surf_map` and `bg_map` which contain value information for coloring (either with a custom colormap / grayscale) - at the vertex level, which can be mapped to either `vertex_colors` or `vertex_values` in a vispy mesh visual (which also has `face_colors` too).

Using that tool one can generate plots like this:
image

This layer type is similar to our `Shapes` layer in some respect, but will be much simpler as there won't be interactivity, and should scale well to very large meshes. I think it would be awkward also to force the user to use a `Shapes` layer when all they want to do are display surfaces.

As to api - how do people feel about `data` being a tuple of (vertices, faces), or is that just confusing, and you'd rather have separate keyword args for both? Also do we want to expose both coloring by vertex / coloring by face or just one? I think we might need to just play around to see what feels right.

Thoughts @royerloic @jni?

sofroniewn:
# Description
This PR will close #502 by adding a surface layer for easy rendering of meshes. It is still a work in progress as it does not yet support nD slicing (i.e. for timeseries) and still has a non ideal UI elements. I've also notices a vispy blending bug that we'll have to fix, but to get a sense of what this new layer type can be used for see the gif below:

add_surface

@danlurie - does this look like some of the functionality you were looking for?

## Type of change

# How has this been tested?

## Final checklist:


sofroniewn:
I've fix some of the minor ordering bugs on our end, but I've now come up against some pretty big transparency / blending problems that may to be on the vispy end. Vispy is working on some of this stuff - see here for example https://github.com/vispy/vispy/issues/1665 - but I'm not sure exactly what's causing our problems yet.

For example see this gif here, as we reduce the opacity of the top colored visual you start seeing this weird line artifact until the screen goes entirely black, even though there is another black and white visual underneath it. Also if I enter the `additive` blending mode as currently defined then the top visual goes away. I can change some of this behavior using the `gloo.set_state` (for example toggling the `depth_test` has a large effect) but right now I was mainly just trying things and seeing what worked. Maybe @royerloic has a better idea, or @pranathivemuri is interested in trying to understand this better:

surface_blending

This PR still needs some work to support the nD stuff, and I saw the comment by @royerloic in #502 around the api design. It also needs tests. All that should be pretty straight-forward, and I'm ok merging this without having the opacity / blending stuff fixed, but we should definitely look into it at some point, and probably engage with the vispy team as they are working on improving the 3D mesh on their end

sofroniewn:
So it looks like some of opacity / z-ordering problems are the same as caused by #22 (the fact that when you bring opacity to 0 you see black) and have nothing to do with the 3D mesh. It does still feel like the weird patterning artifact as you reduce opacity is due to something with the 3D meshes  though

royerloic:
@sofroniewn I think the problem that you are seeing might be related to face-culling:
https://learnopengl.com/Advanced-OpenGL/Face-culling
I can't be sure just from the GIF, but it does seem that random back faces get hidden,
thus creating this strange pattern when transparency is activated. I would try to deactivate
face culling with gloo.set_state (or however it can be set). It might help. 
Note: face-culling is an important optimisation, here it seems that the faces have random winding orders... (hence the randomness of the pattern), so it should probably be active in the absence of transparency. However, this should assume data with correct winding orders... which can be difficult to set right in general...

sofroniewn:
@royerloic `cull_face=True` definitely helps the transparency problem, but does not make it perfect, you still see some areas with the pattern, but there are far fewer.

sofroniewn:
This PR is now ready for review.

There is one test that is currently failing which is around the `add_surface` signature. I tried to make `faces` and `values` required keyword args as @kne42 suggested to me, but doing so seems to cause the test to fail.

The three required input arguments are `add_surface(data, faces=, values=)` where `data` is the `(N, D)` array of vertices, `faces` in an `(M, 3)` array of triangle indices, and `values` is an `(N,)` array of the values to color each vertex with according to the `contrast_limits` and `colormap`

The layer `(N, D)` array of vertices is sliced using our dims model where only faces entirely within a slice are shown. You can also switch back and forth between 2D and 3D rendering as with our other layers.

I've added some tests for the surface layer, and some very simple examples (though the much better examples use real data like in the gifs above)

There are still some unresolved issues with vispy transparency in 3D, but these apply equally to our 3D volume layers and likely need to be fixed on the vispy end.

kne42:
@sofroniewn Ah yes, you discovered a bug, good catch! I'll follow up with a fix + tests soon but I need to get through airport security first ;)

If you want a quick fix to test that things work though try adding `or kind == inspect._KEYWORD_ONLY` [here](https://github.com/napari/napari/blob/master/napari/util/misc.py#L283).

sofroniewn:
Following continued discussion on https://github.com/vispy/vispy/issues/1541 I've identified a change to our `depth_testing` settings - `set_depth_func('lequal')` that has solved a lot of the blending issues described above. It now allows you to blend together two mesh layers, which is very nice functionality, see this gif below:

surface_blending_working

@jni can you review. I'd like to get this in our next release `0.1.5` which @kne42 will probably make in the next couple of days before her talk at imageXD next week

napari/_qt/layers/qt_image_layer.py
```diff
@@ -56,55 +56,69 @@ class QtImageProperties(QtLayerProperties):
...
-            self.layer.rendering, Qt.MatchFixedString
-        )
-        renderComboBox.setCurrentIndex(index)
-        renderComboBox.activated[str].connect(
-            lambda text=renderComboBox: self.changeRendering(text)
-        )
-        self.renderComboBox = renderComboBox
-        self.grid_layout.addWidget(QLabel('rendering:'), row, self.name_column)
-        self.grid_layout.addWidget(renderComboBox, row, self.property_column)
+        if hasattr(self.layer, 'colormap'):
```
kne42: i think all of this `hasattr` checking is really strange

if you are having trouble with inheriting things, just take the base code you want shared everywhere and create a new class that both `Image` and `Surface` can inherit from

sofroniewn: ok, makes sense - i can actually reduce some code duplication between with the labels layer properties here too

sofroniewn: actually the labels layer properties doesn't have the duplicate code. I just worry that we'll now have extra classes. I'll think about how to do this nicely

jni:
@sofroniewn this is so gorgeous! :clap: :clap: Should I wait for you to address @kne42's base class comment first before I review? It seems like it'll touch quite a bit of code, right?

sofroniewn:
It shouldn’t actually touch too much @jni - so either way - but I take that as your agreement with @kne42 that you don’t like the attr checking and would prefer to see more classes?

It will only effect that file

sofroniewn:
I have now added the extra class as requested by @kne42 and @jni - this PR is ready for re-review

napari/_qt/layers/qt_image_layer.py
sofroniewn: done

napari/components/viewer_model.py
```diff
@@ -719,6 +719,82 @@ def add_shapes(
         self.add_layer(layer)
         return layer
 
+    def add_surface(
+        self,
+        data,
```
jni, line 724: I can't remember the outcome of the discussion about `data` vs arguments named after what they are (in this case, `vertices`)? I find this API confusing, because `data` doesn't actually capture the data we need to plot. If anything, `data` in this case should be the `vertices, faces, values` tuple (optionally, NamedTuple). Also, perhaps we can make values optional, and render a monochrome surface if desired?

Another (possibly headed straight to the bin) idea is to rename `faces` with `triangle_idxs` or something similarly descriptive. For non computer graphics people such as myself, this is more descriptive of what the value should be.

Similarly, `values` could be renamed to `face_values` or `triangle_values`.

napari/layers/surface/surface.py
```diff
@@ -0,0 +1,309 @@
...
+    contrast_limits : list (2,)
+        Color limits to be used for determining the colormap bounds for
+        luminance images. If not passed is calculated as the min and max of
+        the image.
+
+    Extended Summary
+    ----------
+    _data_view : (M, 2, 2) array
+        The start point and projections of N vectors in 2D for vectors whose
+        start point is in the currently viewed slice.
```
jni: Huh?

napari/layers/surface/surface.py
```diff
@@ -0,0 +1,309 @@
...
+    Surface layer renders meshes onto the canvas.
+
+    Parameters
+    ----------
+    data : (N, D) array
+        Vertices of mesh triangles.
+    faces : (M, 3) array of int
+        Indices of mesh triangles.
+    values : (N,) array
+        Values used to color vertices.
```
jni: As above: either `vertices, faces, values`, or `data` should contain the three elements.

napari/layers/surface/surface.py
```diff
@@ -0,0 +1,309 @@
...
+    translate : tuple of float
+        Translation values for the layer.
+    opacity : float
+        Opacity of the layer visual, between 0.0 and 1.0.
+    blending : str
+        One of a list of preset blending modes that determines how RGB and
+        alpha values of the layer visual get mixed. Allowed values are
+        {'opaque', 'translucent', and 'additive'}.
+    visible : bool
+        Whether the layer visual is currently being displayed.
```
jni, line 51: Do we have a concept of lighting in this layer? (ie if you render something all the same colour, is it going to be visible as anything other than a homogeneous blob?) And if so, do we want to be able to control it?

examples/add_surface.py
```diff
@@ -0,0 +1,19 @@
...
+with napari.gui_qt():
+    # create the viewer and window
+    viewer = napari.Viewer()
+
+    data = np.array([[0, 0], [0, 20], [10, 0], [10, 10]])
+    faces = np.array([[0, 1, 2], [1, 2, 3]])
+    values = np.linspace(0, 1, len(data))
+
+    # add the surface
+    layer = viewer.add_surface(data, faces=faces, values=values)
```
jni: I'd call this example "add_surface_2d"

napari/layers/surface/surface.py
```diff
@@ -0,0 +1,309 @@
...
+    def _set_view_slice(self):
+        """Sets the view given the indices to slice with."""
+
+        not_disp = list(self.dims.not_displayed)
+        disp = list(self.dims.displayed)
+        indices = np.array(self.dims.indices)
+
+        self._data_view = self.data[:, disp]
+        if len(self.data) == 0:
+            self._view_faces = np.zeros((0, 3))
```
jni, line 269: This is super weird. I would say we either remove nD support or we need an nD example, because I don't really get how to use this. I think @royerloic was right that we probably need to support different number of vertices for different slices. At any rate, I don't understand how a slider would interact with this stuff?

jni (commented):
@sofroniewn As mentioned, this is very very cool, but I don't think the nD support is intuitive — or even tested? I think it should either be ripped out or tested thoroughly...

jni:
I'm also concerned about the API and (now that I see it) about the failing windows build — is that new???

napari/components/viewer_model.py
sofroniewn: We can make data be a tuple - I'm not familiar with NamedTuples, so maybe not require that.

We could have values be a string for a color say `white` or `red` if someone wants monochrome, but then I think the api becomes more confusing and I'm not sure it's warranted. Similarly right now the `values` are actually `vertex_values` - there is one per vertex, we could instead also support `face_values` - one per face (I think) - or `vertex_colors` or `face_colors` which are `Nx4` arrays with actual colors instead of values to pass through a colormap. I feel like having `values` is probably sufficient, but I could rename as `vertex_values`.

Similarly I prefers `triangles` to `faces` and could do `triangle_idxs` if it was more clear.

We might need to talk this over, but I'm open to the suggestions

napari/layers/surface/surface.py
sofroniewn: copy and paste error ;-) fixed

napari/layers/surface/surface.py
sofroniewn: hmm, i'll give it some thought ...

sofroniewn: on the one hand i like having `data` the standard across all our layers, but on the other hand I like `data` only having an array inside it as a tuple might be a bit weird to find, i'm not sure right now :-(

napari/layers/surface/surface.py
sofroniewn: there are some lighting controls easily accessible with the vispy mesh object - see here http://vispy.org/visuals.html#vispy.visuals.MeshVisual - but I'm not sure about how much we want to expose. I worry about too much customizability and too many options

napari/layers/surface/surface.py
sofroniewn: :-) I'm pretty sure the full nD functionality works as expected, but it might not be intuitive / documented / tested properly - i can try improving that, but the way it works right now is to keep the full list of vertices, but trim the list of triangles to only include those triangles that are contained entirely inside the slice.

See this gif where we switch between 2D and 3D display mode where the slider appears:

nD_surface

jni: Is there a test for this? I suggest:

- create 3D mesh and set ndisplay to 3
- set ndisplay to 2
- iterate over all possible values of the non-displayed dimensions and set the dims to that position

napari/layers/surface/surface.py
jni: Sure, absolutely, but what is it currently set to? I guess it should at least be documented somewhere what the default is.

napari/layers/surface/surface.py
jni: Yep. I don't think having it always be a tuple is good, it will make a lot of code more complicated. It should be a tuple when it makes sense.

napari/components/viewer_model.py
jni: NamedTuples are great! =)

```python
In [2]: from collections import namedtuple                                             
In [3]: namedtuple                                                                     
Out[3]: <function collections.namedtuple(typename, field_names, *, verbose=False, rename=False, module=None)>
In [5]: SurfaceMesh = namedtuple('SurfaceMesh', ['vertices', 'faces', 'values'])       
In [6]: data = SurfaceMesh(vertices=np.array([[0, 1], [1, 0], [0, 0]]), 
   ...:                    faces=np.array([[0, 1, 2]]), 
   ...:                    values=np.array([8.9]))                                     

In [7]: data[0]                                                                        
Out[7]: 
array([[0, 1],
       [1, 0],
       [0, 0]])

In [8]: data[1]                                                                        
Out[8]: array([[0, 1, 2]])

In [9]: data[2]                                                                        
Out[9]: array([8.9])

In [10]: data.vertices                                                                 
Out[10]: 
array([[0, 1],
       [1, 0],
       [0, 0]])

In [11]: data.faces                                                                    
Out[11]: array([[0, 1, 2]])

In [12]: data.values                                                                   
Out[12]: array([8.9])
```

jni: vertex_values is super confusing! Surely it makes more sense to color a face (which has extent) than a vertex (which does not)?

sofroniewn: A named tuple looks great, but i worry that making it the only input would be confusing to newcommers who have never used one before. We could convert from a normal tuple to a named tuple internally, but I still have a concern - about the standardization of the api.

For example in the #511 to zarr PR it's much easier if we have named keyword args that are simple objects like arrays instead of tuples

napari/layers/surface/surface.py
sofroniewn: I can make better tests

napari/components/viewer_model.py
sofroniewn: Ok I have switched this over to accepting a tuple where within the layer model, the data gets split up into the different attributes. @kne42 I think might lean towards the names keyword arguments.

Also I hear you that coloring the faces might be more intuitive, but at least this dataset / file format came with the vertex values.

i've also renamed `values` as `vertex_values`

napari/layers/surface/surface.py
sofroniewn: we now have the three separate attributes

napari/layers/surface/surface.py
sofroniewn: i've added a comment on the `VispySurfaceLayer` class

examples/add_surface.py
sofroniewn: done

napari/layers/surface/surface.py
sofroniewn: i've added a test doing exactly this - inside `test_advanced.py` called `test_changing_display_surface`

sofroniewn:
@jni I've made the changes requested including adding a complex test and making the data input argument a tuple. We can discuss tomorrow during team meeting, but it's otherwise ready for re-review

jni:
@sofroniewn still need to give this one another look, but we discussed in the team meeting chatting with @arokem about data formats for surface values + time, to see what the best internal representation might be for higher-than-3D data.

sofroniewn:
We didn't end up talking about it today, but I can look into that. I still think we can get this layer merged without that though

jni:
@sofroniewn another comment we discussed during the meeting is, does this look ok when rendering both a surface and a volumetric image?

examples/nD_labels.py
```diff
@@ -10,6 +10,7 @@
 
 with napari.gui_qt():
     blobs = data.binary_blobs(length=128, volume_fraction=0.1, n_dim=3)
-    viewer = napari.view(blobs=blobs.astype(float))
+    viewer = napari.Viewer()
+    viewer.add_image(blobs.astype(float))
```
jni: Why change this?

napari/_vispy/vispy_surface_layer.py
```diff
@@ -0,0 +1,68 @@
...
+        for b in range(self.layer.dims.ndisplay):
+            self.node.bounds(b)
+
+    def _on_data_change(self):
+        if len(self.layer._data_view) == 0 or len(self.layer._view_faces) == 0:
+            vertices = None
+            faces = None
+            vertex_values = np.array([0])
+        else:
+            vertices = self.layer._data_view[:, ::-1] + 0.5
```
jni, line 45: Can you add a note here about the offset? eg "vispy places pixel coordinates at the corners, with pixel centers offset at 0.5, but for our data we place values at the pixel centers. Here we convert from data space to vispy space"

Maybe there's a way to make that more concise. =P

napari/tests/test_advanced.py
```diff
@@ -171,3 +171,48 @@ def test_update_console(qtbot):
 
     # Close the viewer
     viewer.window.close()
+
+
+def test_changing_display_surface(qtbot):
+    """Test adding 3D surface and chaning its display."""
```
jni:
```suggestion
    """Test adding 3D surface and changing its display."""
```

jni (approved):
An impressive new function for such little code! :clap: I've left some very minor comments. The more interesting one is the data format discussion that you were supposed to have with @arokem. =P Specifically, when you have an MRI dataset I presume (?) the surface stays the same but the values are ever-changing. So we should be able to support that use case. But given that this is working now, I'll be happy if we create an issue to track that use case and merge this as is. :+1: ! Great stuff! :tada:

royerloic:
Cool stuff!

examples/nD_labels.py
sofroniewn: wasn't intentional, was playing around with the file during development, and didn't revert. i'm reverting now

napari/_vispy/vispy_surface_layer.py
sofroniewn: added a shortened version, we might want to put this message on the other layers too, but i'll just put it here for now

napari/tests/test_advanced.py
sofroniewn: fixed

sofroniewn:
I have made an issue #540 around better support for surface timeseries data and have made the changes requested by @jni. I will merge when the tests pass

jni:
:tada: :rocket:

PR #506, opened 2019-09-03
fix contributing guidelines

### LINKED ISSUE #357: Should we add versioneer to development.txt or INSTALL_REQUIRES in setup.py?
I'm trying to copy our version infrastructure for napari-app, and I don't quite grok it. Where is the base version string stored? (ie where does 0.0.9 vs 0.1.0 live, whether it's dev or not) Also, since versioneer is required to run setup.py, does it need to go in the requirements somewhere?

CC @kne42

kne42:
fix errors in contributing guidelines

closes #357

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,28 +32,16 @@ Make the development version available globally:
 $ pip install -e .
 ```
 
-Install the `black` formatter hook with:
+We use [`black`](https://github.com/psf/black) to format our code.
+Please enable autoformatting for the repository with
+[`pre-commit`](https://pre-commit.com):
 ```sh
 pre-commit install
 ```
 
-With help of pre-commit, your future commits to this project will be reformatted with our black configuration,
-which includes settings `skip-string-normalization = true` and `max-line-length = 79`.
-To learn more, see the [black README](https://github.com/python/black).
-
-## Building the dev-docs
-
-Go to documentation source
-```sh
-$ cd docs/source
-```
-
-Build the docs with help of given Makefile
-```sh
-$ make html
-```
-
-Then you can find the built docs in `docs/source/_build/html`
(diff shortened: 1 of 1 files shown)


sofroniewn (approved):
Looks good to me! I'll leave this for a little if @jni wants to comment, then I'll merge

CONTRIBUTING.md
```diff
@@ -39,7 +39,6 @@ pre-commit install
 
 With help of pre-commit, your future commits to this project will be reformatted with our black configuration,
 which includes settings `skip-string-normalization = true` and `max-line-length = 79`.
-On the command-line, our configuration would be equivalent to `black -S -l 79 -t py36 --exclude vendored\|_vispy .`.
```
jni, line 42: Why remove this? I think it's very useful for other libraries considering adding black. (such a thing was proposed at scikit-image a few weeks ago)

jni (commented):
@kne42 I'd revert the removal of the black config explanation, as I think it's relevant.

CONTRIBUTING.md
kne42: I removed it because we have to update it every time we change our `black` config and right now our config is super long (we exclude like 10 different things)

kne42: How about I add a reference to the file where these settings are stored instead?

kne42: Besides, it's not our job to document black; if people are interested in using it, they should look at the official documentation instead.

sofroniewn: adding a reference to the file where this info is stored sounds like a reasonable compromise to me, especially as the `exclude` list might continue to change

jni: Great, I'm into the reference idea.

PR #509, opened 2019-09-04
fix node ordering

sofroniewn:
# Description
This PR fixes the z-ordering problems encountered in #22 and dramatically simplifies the ordering code and switching between 2D and 3D rendering. It does this by using setting `depth_test=False` within our blending code. This hasn't caused any problems with our current layers. We'll have to see how the `Surface` layer does with it after #503 merges, but I'm confident it won't make the behavior there worse.

I'll also note that since #22 was identified there have been a number of improvements to vispy, and so it is very possible that this fix wouldn't have worked then. Once we merge this, I'll go back to https://github.com/vispy/vispy/issues/1541 and close that issue with a comment, sound good @kne42? 

## Type of change

## Final checklist:

### DIFF
--- a/napari/_qt/qt_viewer.py
+++ b/napari/_qt/qt_viewer.py
@@ -149,12 +149,7 @@ class QtViewer(QSplitter):
         vispy_layer = create_vispy_visual(layer)
         vispy_layer.camera = self.view.camera
         vispy_layer.node.parent = self.view.scene
-        # Workaround for bug in #22, and different handling of ordering
-        # for 2D and 3D rendering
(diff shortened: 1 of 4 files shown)


jni (approved):
Wow! =)

jni:
@sofroniewn are you sure that the vispy issue should be closed? Is it still not unexpected behaviour from vispy?

sofroniewn:
@jni - fair enough about closing of the vispy issue, but I'll at least comment that this is now our preferred solution and let them decide what to do

PR #510, opened 2019-09-04
fix call signature to work with keyword-only arguments

kne42:
# Description
fix `callsignature` to work with keyword-only arguments and add tests

see https://github.com/napari/napari/pull/503#issuecomment-528046842


sofroniewn (approved):
Looks good to me!

Issue #490, opened 2019-08-22
save napari to zarr file


sofroniewn:
## 🚀 Feature
It would be nice to do "file save" and save the entire state + data of the napari viewer to a zarr file. This is related to #234 which wanted the ability to serialize the state of the viewer (without the data), and #379 which talked about saving out individual layers (i.e. save an image / labels layer to a tif).

This feature would capture the meta data of the viewer (current camera position, dimensions positions, theme, etc.) and then both the meta data and data of each layer (i.e. for an image layer the clims, colormap, and the pixel data, for a shapes layer, the colors, the coordinates etc.). Each layer would correspond to a [zarr group](https://zarr.readthedocs.io/en/stable/api/hierarchy.html#). If you then opened this `.napari.zarr` file you'd get back to exactly where you were, and what you were looking at. Doing this will duplicate the data if you started from somewhere else, but for small files that would be fine, and it would be done in a loss-less way (without corruption).

A key thing would also be to be able to load in this file using just xarray and zarr, independent of napari, where xarray would get the metadata.

Thoughts @jni @kne42 @jakirkham?

sofroniewn:
@kne42 @jni I've been trying to make custom file icons for our saved zarr files based on napari screenshots, but i'm not able to figure out how to do it. do either of you have any experience with this?

I created an issue on the zarr repo to see if anyone else has any ideas but havn't got any engagement yet - https://github.com/zarr-developers/zarr-python/issues/477

sofroniewn:
So after some pointers from @freeman-lab to get me in the right direction I have figured out how to do this on a mac.

The first step is to generate a `.icns` file (for mac, `.ico` file linux/windows). Fortunately imageio.imwrite can already do this from a numpy array and so that step is very easy for us.

Then given a `.icns` file - in this example called `icon.icns` located inside a folder, to set the folder to have that icon on my mac you just need the following commands:

```python
base = 'path/to/folder'
icon_file = 'icon.icns'

resource_file = os.path.join(base, 'Icon.rsrc')
hidden_icon_file = os.path.join(base, 'Icon\r')

# Create a temporary resource file
with open(resource_file, "w") as resource:
    resource.write(f"read 'icns' (-16455)\"{icon_file}\";")
    
cmds = [['Rez', '-a', resource_file,  '-o', hidden_icon_file], # Create an Icon file
        ['SetFile', '-a', 'C', base], # Set icon of the folder
[...]

PR #511, opened 2019-09-05
[WIP] save / load viewer with zarr

sofroniewn:
# Description
This PR begins to add preliminary support for serializing the viewer state and saving it to a zarr file, and then being able to load the zarr file and auto-populate the viewer, as discussed in #490. It adds zarr as a dependency to napari.

The implementation is a little crude right now and a some corner cases are not tested - particularly for the `Shapes` and `Points` layer where there are some properties that are lists / arrays in addition to the main data properties, and where the data property can be a ragged array.

Also the camera state is not captured in the file yet. Capturing the camera state could be useful if you wanted someone to be able to return to the exact zoom you were at when you saved the file. Not sure though if that's strictly needed.

Right now you can use this functionality with
```python
root = viewer.to_zarr()
```
which returns the zarr object, here called `root` or with
```python
viewer.to_zarr(path)
```
where `path` should be a string path

You can then load a zarr file with
```python
viewer.from_zarr(root)
```
or
```python
viewer.from_zarr(path)
```
to load directly from a zarr object or path.  

Although it is still early some feedback on whether people think this is a good direction or not would be appreciated @royerloic @jni @kne42 @jakirkham. Note I see this functionality as complementary to saving individual layers (say as tif or jpeg) as discussed in #379, but really as way to capture the entire viewer state and all its data in one file.

Here is an example gif of the functionality where I load some data / make some edits, save a zarr file, clear the viewer, and then load in that zarr file:

to_zarr

Some future things to do are to see if there is a way to set the icon of the zarr file to be a thumbnail of the whole viewer.

Also I'll note that right now the zarr file can just be read in using `zarr.open(path)` and all the raw and meta data accessed through normal zarr access patterns. There's a special attribute at the top level of the zarr file `napari` that is meant to indicate that the zarr file is a napari zarr file. We should probably add some version information to the file as well - like generated with napari version ... etc.

### DIFF
--- a/napari/layers/image/image.py
+++ b/napari/layers/image/image.py
@@ -448,3 +448,18 @@ class Image(Layer):
             'image', width=width, height=height, opacity=opacity, **props
         )
         return [xml]
+
+    def _add_zarr_data(self, gp):
+        """Add layer data to zarr group."""
+        gp.array(
+            'data',
+            self.data,
+            shape=self.data.shape,
+            chunks=(128,) * len(self.data.shape),
+            dtype=self.data.dtype,
+        )
+        gp.attrs['multichannel'] = self.multichannel
+        gp.attrs['colormap'] = self.colormap[0]
+        gp.attrs['contrast_limits'] = self.contrast_limits
+        gp.attrs['interpolation'] = self.interpolation
+        gp.attrs['rendering'] = self.rendering
--- a/napari/layers/labels/labels.py
+++ b/napari/layers/labels/labels.py
@@ -533,6 +533,19 @@ class Labels(Layer):
         )
         return [xml]
 
+    def _add_zarr_data(self, gp):
+        """Add layer data to zarr group."""
+        gp.array(
+            'data',
+            self.data,
+            shape=self.data.shape,
+            chunks=(128,) * len(self.data.shape),
+            dtype=self.data.dtype,
+        )
+        gp.attrs['num_colors'] = self.num_colors
+        gp.attrs['seed'] = self.seed
+        gp.attrs['n_dimensional'] = self.n_dimensional
+
     def on_mouse_press(self, event):
         """Called whenever mouse pressed in canvas.
 
--- a/napari/layers/shapes/shapes.py
+++ b/napari/layers/shapes/shapes.py
@@ -1387,6 +1387,21 @@ class Shapes(Layer):
 
         return labels
 
+    def _add_zarr_data(self, gp):
+        """Add layer data to zarr group."""
+        gp.array(
+            'data',
+            self.data,
+            shape=self.data.shape,
+            chunks=(128,) * len(self.data.shape),
+            dtype=self.data.dtype,
+        )
+        gp.attrs['shape_type'] = 'polygon'
+        gp.attrs['edge_width'] = self.edge_width
+        gp.attrs['size'] = self.size
+        gp.attrs['edge_color'] = self.edge_color
+        gp.attrs['z_index'] = 0
+
     def on_mouse_press(self, event):
         """Called whenever mouse pressed in canvas.
 
--- a/napari/layers/shapes/shapes.py
+++ b/napari/layers/shapes/shapes.py
@@ -1389,18 +1389,35 @@ class Shapes(Layer):
 
     def _add_zarr_data(self, gp):
         """Add layer data to zarr group."""
+        gp.array('data', self.data, dtype='array:f8')
+        ew = np.array(self.edge_widths)
         gp.array(
-            'data',
-            self.data,
-            shape=self.data.shape,
-            chunks=(128,) * len(self.data.shape),
-            dtype=self.data.dtype,
+            'edge_width',
+            ew,
+            shape=ew.shape,
+            chunks=(128,) * len(ew.shape),
+            dtype=ew.dtype,
         )
-        gp.attrs['shape_type'] = 'polygon'
-        gp.attrs['edge_width'] = self.edge_width
-        gp.attrs['size'] = self.size
-        gp.attrs['edge_color'] = self.edge_color
-        gp.attrs['z_index'] = 0
+        op = np.array(self.opacities)
+        gp.array(
+            'opacity',
+            op,
+            shape=op.shape,
+            chunks=(128,) * len(op.shape),
+            dtype=op.dtype,
+        )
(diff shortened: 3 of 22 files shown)


napari/layers/image/image.py
```diff
@@ -448,3 +448,18 @@ def to_xml_list(self):
...
         )
         return [xml]
+
+    def _add_zarr_data(self, gp):
+        """Add layer data to zarr group."""
+        gp.array(
+            'data',
+            self.data,
+            shape=self.data.shape,
+            chunks=(128,) * len(self.data.shape),
```
jni, line 458: imho this should definitely be settable... Also, we should use *some* compression scheme, right?

napari/layers/labels/labels.py
```diff
@@ -533,6 +533,19 @@ def to_xml_list(self):
...
 
+    def _add_zarr_data(self, gp):
+        """Add layer data to zarr group."""
+        gp.array(
+            'data',
+            self.data,
+            shape=self.data.shape,
+            chunks=(128,) * len(self.data.shape),
+            dtype=self.data.dtype,
+        )
```
jni, line 544: Given that this is identical to image, it might be worth sharing some code between them.

napari/layers/shapes/shapes.py
```diff
@@ -1387,6 +1387,21 @@ def to_labels(self, labels_shape=None):
 
         return labels
 
+    def _add_zarr_data(self, gp):
+        """Add layer data to zarr group."""
+        gp.array(
+            'data',
+            self.data,
+            shape=self.data.shape,
+            chunks=(128,) * len(self.data.shape),
```
jni: btw, the most common rule of thumb I've heard for chunking sizes is that they should be ~1MB. So it's probably worth writing a function that returns ~1MB chunks depending on the dimensions.

jni (commented):
@sofroniewn very cool! We should probably start thinking about what it takes to write a new napari layer type. Will this PR mean that all layers should provide their own serialisation to zarr? It seems like a pretty heavyweight requirement. Can zarr store arbitrary binary data? If so, maybe we should add a fallback pickle option in the base layer?

jni:
Also, I think storing the camera state would be amazing.

sofroniewn:
The latest changes now allow for icons to be generated and added to our `.napari` files automatically on saving when using a mac. This won't work cross platform yet, but that can be investigated. I think the functionality is quite useful though - here is an example of me browsing some `.napari` files generated from our examples in the mac finder:

browse_files

It would also be nice if right now when you double clicked one of those folders it opened it using napari. It is possible to open a file from the command line with `napari file_name.napari` so it should be pretty straight forward

sofroniewn:
Thanks to some help from @kne42 I'm now on my mac able to double click to open a `.napari` file using napari. This depends right now on having a custom launch script that calls our cli, but once we distribute a bundled app it will be possible to use the bundled app to open napari files.

Also the zarr folder is now set to be a `Package` on a mac which means that you no longer see inside it's contents like a folder when you click on it (instead you need to click show package) and when you double click on it, it opens like a file instead of a folder. I think a lot of biologist will be very happy to see something that looks like a file instead of a folder with a thumbnail and everything, but these "files" can still just be read like plain old zarr files with all the advantages around scale and parallel read / write that comes with it.

Here's a gif of the new launching behavior:
launch_files

I will note that right now launching napari is quite slow, I think a lot of that comes with the stylings, we should as some point try and speed that up, or at least display some graphics that lets you know it is launching, but that can come later.

Right now we have it set such that you cannot re-write over the same file, but I'm going to look into that next

royerloic:
> Thanks to some help from @kne42 I'm now on my mac able to double click to open a `.napari` file using napari. This depends right now on having a custom launch script that calls our cli, but once we distribute a bundled app it will be possible to use the bundled app to open napari files.
> 
> Also the zarr folder is now set to be a `Package` on a mac which means that you no longer see inside it's contents like a folder when you click on it (instead you need to click show package) and when you double click on it, it opens like a file instead of a folder. I think a lot of biologist will be very happy to see something that looks like a file instead of a folder with a thumbnail and everything, but these "files" can still just be read like plain old zarr files with all the advantages around scale and parallel read / write that comes with it.
> 
> Here's a gif of the new launching behavior:
> launch_files
> 
> I will note that right now launching napari is quite slow, I think a lot of that comes with the stylings, we should as some point try and speed that up, or at least display some graphics that lets you know it is launching, but that can come later.
> 
> Right now we have it set such that you cannot re-write over the same file, but I'm going to look into that next

That's beyond amazing... 
This is going to be extremely practical.

sofroniewn:
@joshmoore this is still very much in progress, but was wondering your thoughts on it. You can see #490 for more discussion, but basically this PR is a way to serialize the entire viewer state, attributes, and raw data to zarr file which we then pop an icon on and call a `.napari` file.

joshmoore:
> a way to serialize the entire viewer state, attributes, and raw data to zarr file

Ah, metadata! I'm very excited.

joshmoore:
Few random thoughts / questions:

> can using zarr.open(path) and all the raw and meta data

Assuming the _original_ data came for a zarr, do you foresee writing back to it?

> There's a special attribute at the top level of the zarr file napari that is meant to indicate that the zarr file is a napari zarr file.

An extension mechanism is under way for Zarr v3. It would likely make sense to converge on a mechanism there.

> where there are some properties that are lists / arrays in addition to the main data properties

i.e. metadata spec in general: I certainly foresee having a zarr spec for _some_ of this. That would likely look a lot like having a json exchange format rather than individual attributes. It would be interesting to hear if something like that would work for napari.

d-v-b:
FYI zarr creates files with very restrictive access permissions -- only the file owner can read or write. Until that issue is resolved (see https://github.com/zarr-developers/zarr-python/issues/325 and https://github.com/zarr-developers/zarr-python/pull/445) be careful when using zarr to save data you want to share with other people, or manually change the permissions after creating (which is what i'm currently doing 😡)

sofroniewn:
@ttung @shanaxel42 here is the super preliminary `napari` zarr file

sofroniewn:
We've come a long way since this PR was initially made, and have a new approach to serialization based on pydantic. At some point it will still be nice to connect that serialization to a zarr file for a full round trip, but we can come back and find this PR at that point. No need for us to leave this open anymore.

PR #512, opened 2019-09-05
prevent selected label from being reduced below 0

### LINKED ISSUE #441: label layer can have its label id reduced below 0
## 🐛 Bug

not via the UI but via the property/keybindings

side-effect: the color preview thingy on the left disappears when this happens

kne42:
prevent selected label from being reduced below 0 by raising a `ValueError` in the `Labels.selected_label` property

fixes #441


sofroniewn (approved):
Looks good to me!

Issue #481, opened 2019-08-15
Deprecate add_image(multichannel=True)


jni:
I originally proposed that we use `multichannel=` as a keyword argument for an RGB image layer to match scikit-image. However, in scikit-image, the meaning is more broad: a multichannel array is one where the last dimension is not a spatial dimension but rather different channels of information for the same spatial coordinates. In napari, we use it only to *specifically* mean RGB, so it's a bit of a misnomer.

I instead propose that we deprecate the RGB image layer *or* make it a different, separate layer. When `multichannel` is passed to add_image, what I suggest we do instead is split the channels axis into separate layers, with the mode automatically set to "additive", and different primary colors selected as the colormaps.

There's a lot of API surface to consider here, though. First, this would change the return signature, because we would need to return multiple layers instead of one. (One possible solution here that I haven't explored is the Composite pattern, so that a collection of layers can behave as a single layer.) So perhaps a different method to `add_image` needs to be created (e.g. `add_channels`).

Additionally, @sofroniewn raised the point that RGBA gives us the possibility of pixel-wise transparency, which would be hard to replicate without the RGBA vispy node. I don't have a good solution for this yet but I think we should think hard about it: perhaps something like `alpha_mask=` would be a good parameter to expose for *all* images, not just RGBA ones.

sofroniewn:
@jni I like the concept of `add_channels` that is a convenience wrapper around multiple calls to `add_image` that allows you to add and `rgb` image as 3 color channels, or would allow you have add a `C x ....` array and expand along that C dimension with a provided or default list of colormaps.

I'm really not sure about the `rgba` support. unfortunately the `alpha_mask` is hard as that will require weird interactions with vispy (i.e. generating an rgba image even when we just want to use colormaps!!). i can poke around vispy a bit more, but there might not be a clean way to support this.

sofroniewn:
@royerloic @jni @kevinyamauchi have we had further thoughts on this one? I'm still leaning towards dropping support from `add_image`, providing a convenience method `add_channels` that makes multiple calls to `add_image` looping over a list of colormaps when provided an axis to expand over, and doing some clever guessing / looking at metadata to decide is someone drags / drops an image if its rgb / rgba and if and how we should use `add_channels`.

[...]

PR #528, opened 2019-09-12
add multichannel

sofroniewn:
# Description
Following on discussions #481 and with @bioimage-analysis this PR adds a convenience method for displaying multichannel image data. It also renames `multichannel` to `rgb` on the image layer. This PR still needs tests.

## Type of change

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/components/viewer_model.py
+++ b/napari/components/viewer_model.py
@@ -336,7 +336,7 @@ class ViewerModel(KeymapMixin):
         self,
         data,
         *,
-        multichannel=None,
+        rgb=None,
         colormap='gray',
         contrast_limits=None,
         interpolation='nearest',
@@ -355,9 +355,9 @@ class ViewerModel(KeymapMixin):
         ----------
         data : array
             Image data. Can be N dimensional. If the last dimension has length
-            3 or 4 can be interpreted as RGB or RGBA if multichannel is `True`.
-        multichannel : bool
-            Whether the image is multichannel RGB or RGBA if multichannel. If
+            3 or 4 can be interpreted as RGB or RGBA if rgb is `True`.
+        rgb : bool
+            Whether the image is rgb RGB or RGBA if rgb. If
             not specified by user and the last dimension of the data has length
             3 or 4 it will be set as `True`. If `False` the image is
             interpreted as a luminance image.
@@ -399,7 +399,7 @@ class ViewerModel(KeymapMixin):
         """
         layer = layers.Image(
             data,
-            multichannel=multichannel,
+            rgb=rgb,
             colormap=colormap,
             contrast_limits=contrast_limits,
             interpolation=interpolation,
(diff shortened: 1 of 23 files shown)


napari/components/viewer_model.py
```diff
@@ -415,6 +415,117 @@ def add_image(
...
+        -------
+        layers : list of :class:`napari.layers.Image`
+            The newly-created image layers.
+        """
+        n_images = data.shape[channel]
+
+        base_colormaps = ['red', 'green', 'blue', 'cyan', 'yellow', 'magenta']
+        if colormap is None:
+            indices = np.arange(n_images) % len(base_colormaps)
+            expanded_colormaps = [base_colormaps[i] for i in indices]
```
jni:
```python
expanded_colormaps = itertools.cycle(base_colormaps)
```

=)

napari/components/viewer_model.py
```diff
@@ -415,6 +415,117 @@ def add_image(
...
+            Whether the layer visual is currently being displayed.
+
+        Returns
+        -------
+        layers : list of :class:`napari.layers.Image`
+            The newly-created image layers.
+        """
+        n_images = data.shape[channel]
+
+        base_colormaps = ['red', 'green', 'blue', 'cyan', 'yellow', 'magenta']
```
jni: (a) I would put CYM before RGB, but
(b) before either, I might put BOP: https://twitter.com/christlet/status/1133726940127477761

Thoughts?

napari/components/viewer_model.py
```diff
@@ -415,6 +415,117 @@ def add_image(
...
+        else:
+            expanded_colormaps = [colormap] * n_images
+
+        if contrast_limits is None:
+            expanded_contrast_limits = [None] * n_images
+        else:
+            if isinstance(contrast_limits, list):
+                expanded_contrast_limits = contrast_limits
+            else:
+                expanded_contrast_limits = [contrast_limits] * n_images
```
jni: For all of these, I would say, (1) make sure it's iterable, then (2) use itertools.cycle on the iterable (regardless of length — could be an iterator with no `__len__`). I think that will make the code more readable.

napari/components/viewer_model.py
```diff
@@ -415,6 +415,117 @@ def add_image(
...
+
+        if name is None:
+            expanded_names = [None] * n_images
+        elif isinstance(name, list):
+            expanded_names = name
+        else:
+            expanded_names = [name] * n_images
+
+        layers = []
+        for i in range(data.shape[channel]):
```
jni: Then with cycle you'd need `for i, cmap, clims, name in zip(range(data.shape[channel]), colormaps, ...`. Though you could also use `itertools.islice`.

napari/layers/image/image.py
```diff
@@ -24,9 +20,9 @@ class Image(Layer):
     ----------
     data : array
         Image data. Can be N dimensional. If the last dimension has length
-        3 or 4 can be interpreted as RGB or RGBA if multichannel is `True`.
-    multichannel : bool
-        Whether the image is multichannel RGB or RGBA if multichannel. If
+        3 or 4 can be interpreted as RGB or RGBA if rgb is `True`.
+    rgb : bool
+        Whether the image is rgb RGB or RGBA if rgb. If
```
jni: "if rgb" at the end can be removed...

napari/layers/image/image.py
```diff
@@ -66,26 +62,26 @@ class Image(Layer):
...
     data : array
         Image data. Can be N dimensional. If the last dimension has length 3
-        or 4 can be interpreted as RGB or RGBA if multichannel is `True`.
+        or 4 can be interpreted as RGB or RGBA if rgb is `True`.
     metadata : dict
         Image metadata.
-    multichannel : bool
-        Whether the image is multichannel RGB or RGBA if multichannel. If not
+    rgb : bool
+        Whether the image is rgb RGB or RGBA if rgb. If not
```
jni, line 68: as above

jni (approved):
Love it! Left some minor comments, and of course tests would be nice! =)

napari/components/viewer_model.py
sofroniewn: done, thanks!

napari/components/viewer_model.py
sofroniewn: For now I'll switch CYM with RGB - BOP looks great, but we don't have `orange` and `purple` as defaults. We should add those first and then we can revisit this

napari/layers/image/image.py

napari/layers/image/image.py
sofroniewn: fixed

napari/components/viewer_model.py
sofroniewn: done

napari/components/viewer_model.py
sofroniewn: done

Issue #474, opened 2019-08-12
Every GUI action in napari should be undoable. 


shanaxel42:
## 🚀 Feature

- [ ]  Create Action base class that holds all information needed to execute and undo any given GUI action 
- [ ] Create AttributeGetter, AttributeSetter, and AttributeCaller classes that capture provenance of any given action in user readable format (i.e for macro generation)
- [ ] Create subclass InvertableAction that contains the inverse function of the action (the inverse takes the same params as the action) 
- [ ] Create subclass UnInvertable that saves the input data before the action is executed
- [ ] Wrap simple example actions inside corresponding subclasses 
- [ ] Create a combined undo stack / redo stack data structure with a divider
- [ ] Add functionality to viewer that adds executed action to stack and moves divider when undoing
- [ ] Wrap all existing actions inside corresponding subclasses
- [ ] Add ability to group similar actions together on the undo/redo stacks so that they execute together (i.e paintbrush grouping on mouse press / release instead of all moves) 
- [ ] Each wrapped action should have corresponding unit tests

PR #533, opened 2019-09-14
add undo/redo to labels layer

kne42:
add undo/redo feature to the labels layer. this is separate from the larger undo/redo feature and meant mainly cuz i felt like it should be at least essential on this layer since we do not plan to implement the main feature for a few months

undo/redo is bound to `Ctrl+Z`/`Ctrl+Shift+Z` respectively (remember that `Control` is swapped with `Command` on OSX)

the undo/redo stack is reset whenever you change the viewed dimensions


sofroniewn (approved):
@kne42 this is amazing!

I think reseting when the dimension sliders change is a reasonable choice for now.

I have two questions, one is do we want to impose some limit on the number of undos that are allowed? maybe 20? I worry that otherwise we might end up filling up too much memory

Second, do we want to put some of this machinery on the base layer and then have it save history on `Points` and `Shapes` too, as I think it would be very useful to have their too? We could also add undo / redo to the drop down menus.

Great job overall though, I really like it!!!

napari/layers/labels/labels.py
```diff
@@ -366,6 +370,41 @@ def _set_view_slice(self):
...
+        self._set_view_slice_keep_history()
+        self._undo_history = []
+        self._redo_history = []
+
+    def _trim_history(self):
+        while (
+            len(self._undo_history) + len(self._redo_history)
+            > self._history_limit
+        ):
+            del self._undo_history[0]
```
jni: Maybe use a `deque` instead of a list? You get some nice things (such as `maxlen`) for free:

https://docs.python.org/3.7/library/collections.html#collections.deque

jni (commented):
So exciting!

My main suggestion is to switch from lists to deques, which should simplify the code a fair bit. 

In terms of architecture, I think it's going to be a bit trickier than this to implement. For example, painting with a large brush requires not just clearing painted pixels, but rather replacing the value that was there before the paint action. What's the plan for that sort of thing?

napari/layers/labels/labels.py
kne42: great suggestions, deques look super awesome! since we're checking the length of two stacks, `maxlen` probably won't work atm, but will be super useful when we implement #474 (along with `rotate`)!

sofroniewn:
Thanks for adding the test @kne42! I've just fixed the merge conflict and can take over the PR from here.

This is ready for review by @jni, I can try and make changes if you request any. I can confirm that it works as expected and looks good to merge from my point of view

sofroniewn:
I have verified that this works as expected with the deques and fixed the merge conflicts. I will merge once the tests pass.

tinku99:
Undo only undoes 1 slice, when n-dim is selected.  It should undo all slices that were painted last while n-dim was selected.

sofroniewn:
@tinku99 that's a good point. Right now this undo / redo is a bit of a special case for the labels layer, hence it's limited functionality. We're planning to introduce a more general undo / redo here #474 that would be able to do this correctly

Issue #541, opened 2019-09-18
main api for launching


sofroniewn:
## ❓ Questions and Help

Last week @kne42 and I were able to onboard a bunch of new users at imageXD, and I couldn't help but feel some of our main api was still a little confusing. I know we talked about this a lot a couple months ago, but I wanted to see if anyone else was feeling similarly after getting napari out in the wild a little more.

Specifically I felt that people found confusing the distinction between `view` and `Viewer`, the jump to the `add_*` functions, and then the call signature of `view` which is slightly different then `add_image` (for example you can pass images with names as keyword arguments - `astronaut=image` in view but not in `add_image`).

Another thing that's come up is we've started adding more convenience wrapping - like `add_multichannel`, and more layer types like `add_surface`, but they are a little harder to get to / use then `view` as it requires instantiating that empty `Viewer`, which just isn't so nice.

So - my proposal, which I don't think we've discussed before is to drop `view` and bump all our `add_*` methods up to the top `napari` level and have them return the viewer, so you can now do
```python
from napari import add_image
viewer = add_image(astro)
```
and in one line you've shown an image
or you could do 
```python
from napari import add_surface
viewer = add_surface(mri_surf)
```
and in another line you've added a surface etc.
or you could do 
```python
import napari
viewer = napari.add_multichannel(multi_img)
viewer.add_point(pts)
```
and in two lines you've got multichannel images and points.

I think this will provide much easier access to our napari.

I think I'd leave `Viewer` on `napari` too so you could still do
```python
from napari import Viewer, add_surface
viewer = Viewer(title='new title')
viewer.add_surface(mri_surf)
```
for example, but we'd drop `view` entirely.

One final question then is what should `viewer.add_*` return? Right now it returns the layer. We could leave it like that, but then `napari.add_*` returns a viewer and `viewer.add_*` returns a layer which might be weird. We could have it return the viewer too, but why would you need the viewer back if you just did `viewer.add_*`, though it is the most consistent, or it could just return nothing.

What do people think about this proposal? @jni @kne42 @royerloic @AhmetCanSolak @kevinyamauchi @bryantChhun @shanaxel42 @freeman-lab

AhmetCanSolak:
> 
> So - my proposal, which I don't think we've discussed before is to drop `view` and bump all our `add_*` methods up to the top `napari` level and have them return the viewer, so you can now do
> 
> ```python
> from napari import add_image
> viewer = add_image(astro)
> ```

I am absolutely up for dropping `view`.
[...]

PR #542, opened 2019-09-19
add view_* methods at napari level

sofroniewn:
# Description
This PR simplifies the main api as discussed in #541 by dropping the `napari.view` method and promoting all our `add_*` methods so they can be used as `napari.add_*`. In doing so they will create a new viewer that they will return. You can still do `napari.Viewer()` to create a new empty viewer with the necessary keyword args there. Finally the `viewer.add_*` methods are left as is, they still return the layer or list of layers which was added. I see no harm in that actually, but we can drop.

It is no longer possible to add multiple images at the same time as it was with `view` where the keyword then becomes the name of the layer. Personally I always found that confusing, but if @jni you want functionality like that then how about we make something like `add_collection` or even `add_directory` if you want to point at a directory of tifs and use dask etc. I like the consistency of having these `add_*` methods, some of which can place additional wrapping around our layer primitives.

One downside though is that we now have three copies of essentially the same doc strings. Any ideas on how to improve this would be appreciated @kne42. Or we should at least add similar tests for this third set of docstrings to ensure they are the same as the others. That is not done in this PR yet.

## Type of change

## Final checklist:


sofroniewn:
Based on the feedback during team meeting I have made this now expose `napari.view_*` methods that create and return and viewer at the same time as adding a layer.

This is now ready for review @jni

README.md
```diff
@@ -51,7 +51,7 @@ From inside an IPython shell (started with `ipython --gui=qt`) or Jupyter notebo
 %gui qt5
 from skimage import data
 import napari
-viewer = napari.view(data.astronaut(), rgb=True)
+viewer = napari.view_image(data.astronaut())
```
jni: Why is `rgb=True` removed?

jni: Clarification: in general, I would like us to move towards less magic, not more, so I think it's handy to keep this here.

examples/add_pyramid.py
```diff
@@ -13,13 +13,10 @@
 # create pyramid from astronaut image
 base = np.tile(data.astronaut(), (8, 8, 1))
 pyramid = list(
-    pyramid_gaussian(base, downscale=2, max_layer=4, rgb=True)
+    pyramid_gaussian(base, downscale=2, max_layer=4, multichannel=True)
```
jni, line 16: Whoa, crazy that we let this through unnoticed. Methinks we should go back to running the examples automatically... Or at least have a script, ignored by travis, that we can run locally...

napari/__init__.py
```diff
@@ -28,7 +28,7 @@
 
 from .viewer import Viewer
 from . import keybindings
-from .view_function import view
+from .add_layers import *
```
jni: Never import `*`!!! As just one example why not, you know have `ensure_iterable` and `is_iterable` in the `napari.<TAB>` namespace. It also messes with tools like linters.

napari/add_layers.py
```diff
@@ -0,0 +1,633 @@
+import numpy as np
```
jni: Can we rename this file to e.g. `view.py`? ;) Or `view_layers.py`.

napari/add_layers.py
```diff
@@ -0,0 +1,633 @@
...
+            scale=scale,
+            translate=translate,
+            opacity=opacity,
+            blending=blending,
+            visible=visible,
+        )
+    return viewer
+
+
+def view_pyramid(data, *args, **kwargs):
```
jni: For the similar reasons that `import *` are discouraged, `*args` and `**kwargs` are also best avoided, especially when the parameter list is itself very clear (see the doc).

jni: And this is more true still given that the rest of the file does include all the parameters!

napari/add_layers.py
```diff
@@ -0,0 +1,633 @@
...
+        length=length,
+        name=name,
+        metadata=metadata,
+        scale=scale,
+        translate=translate,
+        opacity=opacity,
+        blending=blending,
+        visible=visible,
+    )
+    return viewer
```
jni: Consider this comment entirely optional, just wanted to write it down somewhere.

Depending on how tempted you are by extra work that makes the code more elegant (but it is extra work), it occurs to me that this whole file could be replaced by something like:

```python
import functools
from .viewer import Viewer


def view_layer(layer_method):
    def view_func(*args, **kwargs):
        v = Viewer()
        getattr(v, layer_method.__name__)(*args, **kwargs)
        return v
    view_func.__doc__ = fix_docstring(layer_method.__doc__)
    return view_func

view_image = view_layer(Viewer.add_image)
view_pyramid = view_layer(Viewer.add_pyramid)
...
```

Of course, you'd need to write `fix_docstring`, and I think then there's no way around `*args` and `**kwargs`, which is annoying. (Perhaps the signature can be altered the same way that the doc is altered, not sure.) For fix docstring, you probably want to use numpydoc, which I've been playing with for the plugins PR. It's pretty nifty!

```
In [13]: numpydoc.get_doc_object(feature.hessian_matrix)                                         
Out[13]: <numpydoc.docscrape_sphinx.SphinxFunctionDoc at 0x7f6593757908>

In [14]: _13.get_func                                                                            
Out[14]: <bound method FunctionDoc.get_func of <numpydoc.docscrape_sphinx.SphinxFunctionDoc object at 0x7f6593757908>>

In [15]: _13.get_func()                                                                          
Out[15]: 
(<function skimage.feature.corner.hessian_matrix(image, sigma=1, mode='constant', cval=0, order='rc')>,
 'hessian_matrix')

In [16]: _13.xref_param_type                                                                     
Out[16]: False

In [17]: _13.items()                                                                             
Out[17]: ItemsView(<numpydoc.docscrape_sphinx.SphinxFunctionDoc object at 0x7f6593757908>)

In [18]: list(_13.items())                                                                       
Out[18]: 
[('Signature',
  "hessian_matrix(image, sigma=1, mode='constant', cval=0, order='rc')"),
 ('Summary', ['Compute Hessian matrix.']),
 ('Extended Summary',
  ['The Hessian matrix is defined as::',
   '',
   '    H = [Hrr Hrc]',
   '        [Hrc Hcc]',
   '',
   'which is computed by convolving the image with the second derivatives',
   'of the Gaussian kernel in the respective r- and c-directions.']),
 ('Parameters',
  [Parameter(name='image', type='ndarray', desc=['Input image.']),
   Parameter(name='sigma', type='float', desc=['Standard deviation used for the Gaussian kernel, which is used as', 'weighting function for the auto-correlation matrix.']),
   Parameter(name='mode', type="{'constant', 'reflect', 'wrap', 'nearest', 'mirror'}, optional", desc=['How to handle values outside the image borders.']),
   Parameter(name='cval', type='float, optional', desc=["Used in conjunction with mode 'constant', the value outside", 'the image boundaries.']),
   Parameter(name='order', type="{'rc', 'xy'}, optional", desc=['This parameter allows for the use of reverse or forward order of', "the image axes in gradient computation. 'rc' indicates the use of", "the first axis initially (Hrr, Hrc, Hcc), whilst 'xy' indicates the", 'usage of the last axis initially (Hxx, Hxy, Hyy)'])]),
 ('Returns',
  [Parameter(name='Hrr', type='ndarray', desc=['Element of the Hessian matrix for each pixel in the input image.']),
   Parameter(name='Hrc', type='ndarray', desc=['Element of the Hessian matrix for each pixel in the input image.']),
   Parameter(name='Hcc', type='ndarray', desc=['Element of the Hessian matrix for each pixel in the input image.'])]),
 ('Yields', []),
 ('Receives', []),
 ('Raises', []),
 ('Warns', []),
 ('Other Parameters', []),
 ('Attributes', []),
 ('Methods', []),
 ('See Also', []),
 ('Notes', []),
 ('Warnings', []),
 ('References', ''),
 ('Examples',
  ['>>> from skimage.feature import hessian_matrix',
   '>>> square = np.zeros((5, 5))',
   '>>> square[2, 2] = 4',
   ">>> Hrr, Hrc, Hcc = hessian_matrix(square, sigma=0.1, order='rc')",
   '>>> Hrc',
   'array([[ 0.,  0.,  0.,  0.,  0.],',
   '       [ 0.,  1.,  0., -1.,  0.],',
   '       [ 0.,  0.,  0.,  0.,  0.],',
   '       [ 0., -1.,  0.,  1.,  0.],',
   '       [ 0.,  0.,  0.,  0.,  0.]])']),
 ('index', {})]

In [19]: _13['Parameters']                                                                       
Out[19]: 
[Parameter(name='image', type='ndarray', desc=['Input image.']),
 Parameter(name='sigma', type='float', desc=['Standard deviation used for the Gaussian kernel, which is used as', 'weighting function for the auto-correlation matrix.']),
 Parameter(name='mode', type="{'constant', 'reflect', 'wrap', 'nearest', 'mirror'}, optional", desc=['How to handle values outside the image borders.']),
 Parameter(name='cval', type='float, optional', desc=["Used in conjunction with mode 'constant', the value outside", 'the image boundaries.']),
 Parameter(name='order', type="{'rc', 'xy'}, optional", desc=['This parameter allows for the use of reverse or forward order of', "the image axes in gradient computation. 'rc' indicates the use of", "the first axis initially (Hrr, Hrc, Hcc), whilst 'xy' indicates the", 'usage of the last axis initially (Hxx, Hxy, Hyy)'])]
```

napari/tests/test_napari.py
```diff
@@ -0,0 +1,178 @@
+import numpy as np
+import napari
+
+
+def test_add_image(qtbot):
```
jni: Rename these functions?

jni (approved):
@sofroniewn approved conditional on removing `import *`. All other comments are optional. =)

README.md
sofroniewn: good point, added back in

examples/add_pyramid.py
sofroniewn: i know :-( I think a script that we commit to the repo and can run locally is a good start. can you make it or maybe @AhmetCanSolak?

napari/add_layers.py
sofroniewn: renaming to `view_layers.py`

napari/add_layers.py
sofroniewn: this is going to get refactored when we merge pyramid into image so i'll ignore for now

napari/__init__.py
sofroniewn: good points, fixed

napari/add_layers.py
sofroniewn: sounds like a great follow-up PR ;-)

napari/tests/test_napari.py
sofroniewn: done

sofroniewn:
I've removed the `import *` and made the major changes requested. I will merge after the tests pass

Issue #531, opened 2019-09-13
callback hook to provide additional information on hover over image


VolkerH:
## 🚀 Feature
Allow the user to provide the viewer with a callback function, that receives the current x,y coordinate and intensity and returns data to either be shown somewhere in the viewer or as a tooltip.

## Motivation / Pitch

One possible use case:

When dealing with segmented objects (label images) it is desirable to query information about these objects, e.g. object measurements such as those calculated with `skimage.measure.regionprops_table`.

```python
def example_callback(coord, intensity):
     ... # code to access table 
     if intensity > 0:
          measurements = obj_props_table[intensity]
          return f"<b>Area</b> {measurements["area"]}"
    else:
        return "background" 
```

There are probably also some uses for intensity images (e.g. providing the mean over a neighbourhood around the pixel, or the label number (from a label layer). Be allowing the user to provide a callback they have the flexibility to tailor this to their needs.

## Alternatives

There would be various ways to visualize the data, either as a hover tooltip, somewhere where currently the intensity values are shown or in a separate inspector window that can render HTML and which would offer more space. One could also consider having multiple such options with a callback for each.

sofroniewn:
thanks @VolkerH for making an issue. I think it would be addressed by #252 which we havn't worked on yet, but good to know another example use case, @kne42.

sofroniewn:
We've now added custom callbacks for mouse operations, we havn't added an additional display or html rendering capabilities, but for now you could print to the terminal. Curious if this is helpful to you.

We also need to add a tutorial on using the custom mouse functions, but for now you can maybe get a sense by looking at the top of #544

VolkerH:
Thanks, I will give this a try and see whether I can come up with a short demo.

VolkerH:
This is a very quick & dirty demo of what I had in mind:

![napari_label_inspector](https://user-images.githubusercontent.com/2210044/67252983-31083e00-f4c1-11e9-988c-03fd40318312.gif)

Code is here:
https://github.com/VolkerH/NapariPlayground/blob/master/CustomMouseFunctions/custom_mouse_functions.py

I experinted with a QTableView (which doesn't make too much sense as we can only hover over one object at a time, so there is only one row.  A table view would make sense for selections, e.g. rectangular selections containing multiple objects). The best thing might be a tooltip.

I will try and make a more "tidy" demo in the coming weeks.

[...]

Issue #527, opened 2019-09-12
windows problem


dany-fu:
## 🐛 Bug
Napari launches and a thumbnail of my image can be seen in the top right corner, but the actual viewer is just black

![image](https://user-images.githubusercontent.com/7750862/64814859-49a44080-d572-11e9-8c8d-95a01a09fbc6.png)

## To Reproduce
Jupyter notebook
https://github.com/hicsail/neuro-circuit/blob/df-dev/starfish-notebooks/sample2Round2.ipynb

## Expected behavior
See an enlarged version of the thumbnail in the viewer

## Environment
see notebook

## Additional context
I have run the example notebooks in the past on Napari v0.0.6 on a Mac and not did encounter this bug. Not sure if it's Windows specific or if it's my images.

sofroniewn:
hi @dany-fu this example is using napari `0.1.3` can you update to `0.1.5` and see how it goes?

kne42:
@sofroniewn starfish currently enforces napari's version to be `0.1.3`

dany-fu:
I can change that constant locally just for testing

kne42:
@dany-fu yeah you could try that. however, looking at the errors, it looks like it has to do with OpenGL/maybe VisPy. could you check if you have OpenGL installed?

dany-fu:
I have pyopengl

dany-fu:
Interestingly, when I upgraded to 0.1.5 the openGL errors went away, but the initial 'LooseVersion' error still persists

kne42:
That's the python bindings, but it does not guarantee that OpenGL is installed on your system.

Could you send me the output of the following?
```python
from vispy.gloo.gl2 import _lib, _get_gl_version

print(_lib)
print(_get_gl_version(_lib))
```

dany-fu:
`ModuleNotFoundError: No module named 'vispy.gloo.gl2'` 

My errors look similar to this thread here https://github.com/vispy/vispy/issues/1503, is it worth to try and install pyqt4? Would that work with Napari?

kne42:
my bad, it was `vispy.gloo.gl.gl2`, but regardless, if it was fixed by the latest napari version then it probably wasn't OpenGL

we are not compatible with PyQt4

your LooseVersion error may have to do with conda doing funny stuff. could you send me the output of `conda list`?

sofroniewn:
This could be related to some problems with windows people were having in #428 with the labels layer on windows

dany-fu:
[...]

Issue #428, opened 2019-07-18
Labels layer not visible on Windows


joe-albert:
## 🐛 Bug

If you add a labels layer on Windows you can see the labels in the thumbnail but not in the viewer.

I also get a vispy warning: 
INFO: Program has unset variables: ('texture2D_LUT')

## To Reproduce

```python
image = io.imread('image.png')
segmentation = io.imread('segments.png')
glom = (segmentation == 1).astype(int)
v = napari.Viewer()
v.add_image(image)
v.add_labels(glom)
```
~~@jni will provide the example data asap~~ **Note from @jni:** Added below

## Environment

 - napari Version (e.g., 1.0): 0.1.1+5g1a87b2b
 - OS (e.g., Linux): Windows 7
 - Python version:3.7
 - Any other relevant information:

jni:
test images: (though I suspect this would work fine with random data)

image.png
![image](https://user-images.githubusercontent.com/492549/61493092-678d6480-a978-11e9-9f0c-e10f13022acb.png)

segments.png
![segments](https://user-images.githubusercontent.com/492549/61493098-6b20eb80-a978-11e9-902f-7058f3779f67.png)

sofroniewn:
interesting, we've been running into some problems with our labels colormap - i have a PR in right now #430 that might fix it. if not we'll follow up

sofroniewn:
@DowneastFreaks do you mind upgrading napari to `0.2.0` and trying this again, we've made a number of improvements that might fix this, and we've also been trying to understand our windows problems a little more systematically - see for example this thread #558.

If it still doesn't work, do you mind running
```python
import vispy
print(vispy.sys_info())
```
and sending us the results.

Thanks!!

joe-albert:
@sofroniewn 

I was able to get this running using some of the comments in #558 

Specifically this was the code that worked for me:
```
from skimage import io
import napari

image = io.imread('image.png')
segments = io.imread('segments.png')
glom = (segments == 1).astype(int)

with napari.gui_qt():
    viewer = napari.view_image(image)
    viewer.add_labels(glom)
```

![image](https://user-images.githubusercontent.com/51097457/66314110-864d2880-e8e1-11e9-82f5-d937c08ae9bc.png)

Thanks for looking into this and let me know if there are any other questions I can help with! :)

sofroniewn:
That's great. So glad it is working. Do you mind actually posting the results of 

```python
import vispy
print(vispy.sys_info())
```

[...]

Issue #252, opened 2019-05-06
custom mouse functions


kne42:
as discussed in the meeting today, we may want to support user-defined functions that get executed with mouse events

following is the proposed format of the function a user would provide:
```python
def status(layer):
    # on mouse press
    print('pressed!')

    yield

    while layer.dragging:
        # on mouse move
        print(layer.pos)
        yield

    # on mouse release
    print('released!')
```

sofroniewn:
Looking promising!
 
We need to be able to get both an `is_moving` and an `is_dragging` (which is a subset of `is_moving`)

And we need to get any modifier keys (like if `Shift` being pressed)

Also we should make sure that we are ok not supporting `double_click` events for now. One of the reasons is that `double_click` events can be handled different on different OS - read this disclaimer from the vispy docs

> The ordering of the mouse_double_click, mouse_press, and mouse_release events are not guaranteed to be consistent between backends. Only certain backends natively support double-clicking (currently Qt and WX); on other backends, they are detected manually with a fixed time delay. This can cause problems with accessibility, as increasing the OS detection time or using a dedicated double-click button will not be respected.

Any further thoughts from @royerloic @jni?

sofroniewn:
Another idea was to have an extra mode / button on the left for each layer that allows the "custom cursor" to be selected. When in this mode any custom mouse events get executed if they have been defined. What this means is that the normal cursor is always the normal cursor with pan/zoom, the paintbrush is always the paintbrush, and the "custom" is always user defined.

If the user wants multiple user defined cursors, they can all live under the custom cursor mode / button and they can then use custom hotkeys to switch between them.

@gregjohnso does this sound like useful functionality to complement the custom keybindings?

royerloic:
I think we can definitely live without double clicking, at least for now.
It might even be a good thing, double clicking is not a necessity and
can be unnecessarily strenuous if needed many times -- think
annotations, etc...

jni:
Well, as I mentioned originally, I think double click to zoom is very nice and matches what Google Maps does. From your documentation link, @sofroniewn, it sounds like double click *would* work on Qt?

I don't love the custom cursor button... I'd rather a warning when your settings shadow existing settings. That's what every app I've seen does. There should also be a fast way to resolve these conflicts one way or the other.

[...]

PR #544, opened 2019-09-20
implement custom mouse function framework

kne42:
# Description

For on press, on drag, and on release events:
```python
@layer.mouse_drag_callbacks.append
def hello_world(layer, event):
    "dragging"
    # on press
    print('hello world!')
    yield

    # on move
    while event.type == 'mouse_move':
        print(event.pos)
        yield

    # on release
    print('goodbye world ;(')
```

For only an on press event:
```python
@layer.mouse_drag_callbacks.append
def press(layer, event):
    print('hi')
```

For whenever the mouse is moved - no pressing needed:
```python
@layer.mouse_move_callbacks.append
def move(layer, event):
    print(event.pos)
```

double click is not supported


sofroniewn:
I've added an example of a custom mouse function that does a dilation on the clicked region in a labels layer and prints to the status bar the size of the region after dilation. It's in `examples/custom_mouse_functions.py`. Here's a gif:

custom_mouse_functions

Pretty cool stuff I think! We still have to think about the API / naming a little, and I still need to add tests but thanks to Kira this is coming along very nicely!!

On the API side, we need to think about if it is possible to bind custom mouse functions to layer classes, not just instances, and if it is possible to bind to the viewer and the Viewer class.

Right now we've made the mouse functions be additive which I think is the right choice.

I'm not 100% sure about the `@layer.mouse_drag_callbacks.append` syntax, and if it's clear that that's where mouse press / release get added too, and if it's clear what the difference with `@layer.mouse_move_callbacks.append` is, but we can discuss that next meeting.

Maybe @royerloic @jni @AhmetCanSolak have thoughts

sofroniewn:
Note that this will being to address #252 and #531 from @VolkerH

sofroniewn:
This is now ready for review @jni. I've added what I hope are pretty comprehensive tests, including simulating mouse press, drag, release, and move using `from unittest.mock import Mock`. Thanks for the tip @shanaxel42! (I will probably make a new PR adding similar tests for our keybindings if people like them, and other aspects of the code - `Mock` seems like it will be very useful for testing a lot of the UI interactivity stuff)

I've made it so that a mouse binding added to `mouse_move_callbacks` will NOT get called during a drag event even though the mouse is technically moving. How do people feel about that? I thought it was cleaner to have it completely separate from anything that might get called in `mouse_drag_callbacks`.

I've also made it so you can now bind custom mouse functions to the `viewer` not just the layers. I havn't made them bindable to the class `Viewer`, but to the instance, but I think that's ok. We might even want to remove the class bindings from the keys at some point but that can wait.

I've also update the example now so that clicking will cause a dilation but alt clicking will cause an erosion to illustrate how you can use modifier keys inside custom mouse functions.

Finally if people like this, then I will move the custom mouse functions inside all the layers to a separate file called `mousebindings.py` that will look like the `keybindings.py` file in some senses. I'll do that in a new PR though. That will make it easier to overwrite any mousebindings that we've added instead of it just being additive.

sofroniewn:
So the I'm pretty confident the reason why our windows build is failing in this PR and this PR alone is I added the line `self.canvas.events.ignore_callback_errors = False`. This wasn't something needed for this PR, but is probably good practice to have as otherwise there could be errors that we aren't catching. I think we should make a new PR adding that line in, and my suspicion is that the windows tests that are failing could be related to the windows bugs that people have been reporting in #527 and and #428 but which we were missing.

examples/custom_mouse_functions.py
```diff
@@ -0,0 +1,46 @@
...
+            if 'Shift' in event.modifiers:
+                binary_new = binary_erosion(binary)
+                data[binary] = 0
+                data[binary_new] = val
+            else:
+                binary_new = binary_dilation(binary)
+                data[binary_new] = val
+            size = np.sum(binary_new)
+            layer.data = data
+            msg = f'clicked at {cords} on blob {val} which is {size} now pixels'
```
jni:
```suggestion
            msg = f'clicked at {cords} on blob {val} which is now {size} pixels'
```

examples/custom_mouse_functions.py
```diff
@@ -0,0 +1,46 @@
...
+        else:
+            msg = f'clicked at {cords} on blob {val} which is ignored'
```
jni:
```suggestion
            msg = f'clicked at {cords} on background ({val}) which is ignored'
```

napari/_qt/qt_viewer.py
```diff
@@ -289,26 +289,90 @@ def _toggle_console(self):
...
 
-    def on_mouse_move(self, event):
-        """Called whenever mouse moves over canvas.
+    def on_mouse_press(self, event):
+        """Called whenever mouse pressed in canvas.
         """
+        if event.pos is None:
+            return
+
+        event = ReadOnlyWrapper(event)
```
jni, line 304: Why is this needed?

napari/_qt/qt_viewer.py
```diff
@@ -289,26 +289,90 @@ def _toggle_console(self):
...
+            except StopIteration:
+                pass
+            del self.viewer._mouse_drag_gen[func]
+            del self.viewer._persisted_mouse_event[gen]
+
         layer = self.viewer.active_layer
         if layer is not None:
             self.layer_to_visual[layer].on_mouse_release(event)
+            for func, gen in tuple(layer._mouse_drag_gen.items()):
+                layer._persisted_mouse_event[gen].__wrapped__ = event
```
jni: All this stuff is very confusing... I think comments would be very welcome in code like this, explaining what's happening.

napari/_qt/qt_viewer.py
```diff
@@ -289,26 +289,90 @@ def _toggle_console(self):
...
+                try:
+                    next(gen)
+                except StopIteration:
+                    pass
+                del layer._mouse_drag_gen[func]
+                del layer._persisted_mouse_event[gen]
```
jni: Should these be in a `finally`? ie do they need to be activated whether or not an uncaught exception happens in the try clause?

napari/_qt/qt_viewer.py
```diff
@@ -289,26 +289,90 @@ def _toggle_console(self):
...
         if layer is not None:
-            self.layer_to_visual[layer].on_mouse_move(event)
+            self.layer_to_visual[layer].on_mouse_press(event)
+            for mouse_drag_func in layer.mouse_drag_callbacks:
+                gen = mouse_drag_func(layer, event)
+                if inspect.isgeneratorfunction(mouse_drag_func):
+                    try:
+                        next(gen)
+                        layer._mouse_drag_gen[mouse_drag_func] = gen
+                        layer._persisted_mouse_event[gen] = event
```
jni: This code is repeated in all the `on_mouse_*` methods... Would it be possible to unify it somewhere?

jni (approved):
@sofroniewn I don't have time for an in depth review right now. tbh I would prefer if you could hold off on merging this, so that we could get it to a simpler, more easily grokkable state, but if you want it for your talk, go for broke. ;)

royerloic:
This is very cool!
Happy with how it is implemented now,
usage will dictate if changes are needed.
 
I think the dilation/erosion feature is super useful, 
and could become a 'tool' in the label layer, perhaps.

sofroniewn:
@jni @royerloic I still need to make these changes - will do so now

napari/_qt/qt_viewer.py
sofroniewn: I asked kira about this, and she said it is needed to make sure the custom function don't edit the events. While this seems unlikely I guess it is safer.

examples/custom_mouse_functions.py
sofroniewn: fixed

examples/custom_mouse_functions.py
sofroniewn: fixed

napari/_qt/qt_viewer.py
sofroniewn: possibly, i'd be very open to suggestions. maybe can come in a refactor

napari/_qt/qt_viewer.py
sofroniewn: i'm not 100% sure

jni: I find it scary to merge code that we don't understand fully... There are parts of scikit-image that were merged with only a superficial understanding by the reviewer, and this code has ended up being very hard to maintain — essentially untouchable. So if you can live without this in 0.2 I'd love to wait for us to be able to review this in detail and understand deeply what's happening. (But if you can't... Well, we'll deal with it. ;)

sofroniewn:
Ok, I'm going to wait on this for `0.2.0`, but would like to add it to `0.2.1` so if we could do a through job reviewing this relatively soon @jni that would be great. I can look at it in more detail again next week.

napari/tests/test_mouse_bindings.py
```diff
@@ -0,0 +1,229 @@
+import numpy as np
+from unittest.mock import Mock
```
sofroniewn, line 2: @jni you mentioned using `pytest` mock, but there doesn't seem to be one built in - it looks like a separate package from here - https://github.com/pytest-dev/pytest-mock - but maybe I'm getting this wrong. Otherwise I think leaving the `unittest` one is fine

napari/_qt/qt_viewer.py
sofroniewn: unified in `utils.misc`

napari/_qt/qt_viewer.py
sofroniewn: i think they are in the right place as these objects need to be deleted at the end of the mouse release regardless of if there was an error or not

sofroniewn:
@jni I have refactored this to reduce duplicate code and added a couple comments. I'm pretty happy with it now and think we should merge, I leave it to you to do a deep dive though. It would be nice to get it in soon though

napari/components/viewer_model.py
```diff
@@ -89,6 +89,11 @@ def __init__(self, title='napari', ndisplay=2, order=None):
         self.layers.events.removed.connect(self._update_active_layer)
         self.layers.events.reordered.connect(self._update_active_layer)
 
+        self.mouse_move_callbacks = []
+        self.mouse_drag_callbacks = []
+        self._persisted_mouse_event = {}
```
jni, line 102: Can you add comments for each of these elements about what they are supposed to hold? It's not totally obvious from the name.

napari/tests/test_mouse_bindings.py
jni: Probably the package is what I was thinking about. Happy to stay with unittest for now since we are not using complex features within it (that might require less pytest-friendly syntax), and it's built in.

napari/util/misc.py
```diff
@@ -351,3 +353,80 @@ def __str__(self):
...
+    """
+    for mouse_drag_func in obj.mouse_drag_callbacks:
+        gen = mouse_drag_func(obj, event)
+        if inspect.isgeneratorfunction(mouse_drag_func):
+            try:
+                next(gen)
+                obj._mouse_drag_gen[mouse_drag_func] = gen
+                obj._persisted_mouse_event[gen] = event
+            except StopIteration:
+                pass
```
jni, line 599: So imho all this logic needs comments explaining what is being done, though I sort of get it:

- run the function, getting a result
- if the result is a generator, then get the next value from the generator, and
- ??? What is being added to those dictionaries? And where are the dictionaries used?

In general, one of the principles of writing bug free code is that code should be as local as possible, meaning that I shouldn't have to keep in my head all the possible accesses and lifetimes of stuff that lives outside of this function. I'm not sure how to improve this implementation but at a minimum, there should be comments here about what those dictionaries do and where they are "cleared" and why. (I see some dels in the next function but it's unclear to me whether they refer to the same object (ie are they supposed to happen in quick succession)

jni (commented):
@sofroniewn I sort of get the implementation now, but I think it needs some comments around those dictionary objects to make it clear to future generations what's happening where. At that point is should be mergeable.

napari/components/viewer_model.py
sofroniewn: added

napari/util/misc.py
sofroniewn: done

sofroniewn:
added more comments and the example code for defining one of these callbacks to the doc strings - i feel pretty good about this now

I will merge once the tests pass

Issue #538, opened 2019-09-16
Help with starfish.display for large images & mask collections


ambrosejcarr:
## 🐛 Bug

I would like to load the segmented Allen data in napari. These data consist of ~2200 regions of interest over a field of view of shape (27084, 20825)

When I attempt to use napari to load these data, the images are too large and the mask collections throw errors. You can view my attempts [here](https://github.com/spacetx/starfish/blob/2d61bdd0490f22db2837aa52249baaf66d40f18b/notebooks/manual_segmentation_example.ipynb), beginning under the "GeoJSON to SegmentationMaskCollection" heading. Please forgive the messy notebook. 

## To Reproduce

unzip the file downloaded from s3: `s3://starfish.data.spacetx/smFISH/mouse/contributor_results/napari_large_image.zip`
```
70167526 Sep 11 16:40 mFISH__001_registered-downsampled.png
3390209  Sep 16 14:56 mFISH_labels.png
```
To reproduce, load the label image and attempt to show the resulting data in napari. 

```
import skimage.io
labels = skimage.io.imread('mFISH_labels.png')
```
## Expected behavior

I'd like to be able to view my images & masks without size constraints. 

## Environment

 - napari Version (e.g., 1.0): 0.1.3 and 0.1.5 both throw the errors
 - OS (e.g., Linux): OS X 10.14
 - Python version: 3.7

sofroniewn:
As noted in our slack exchanges a napari image layer will choke if you try and pass it an image > 16k x 16k, you instead need to use the pyramid layer and pass a list of precomputed image pyramids. There is some discussion right now in #537 around autocreating the pyramid when we need, which could be the long term solution.

ambrosejcarr:
Does this also explain the issue with the masks? If that's the case, it sounds like the fix for this could be relatively simple and I'm happy to close this in favor of #537.

ambrosejcarr:
I tried to display the mask as an image using the pyramid and wasn't able to get it to work. Am I doing it incorrectly? 

```python3
>>> import napari
>>> print(napari.__version__)
0.1.5

>>> from skimage.transform import pyramid_gaussian
>>> pyramid = tuple(pyramid_gaussian(labels, downscale=2, multichannel=False))
>>> len(pyramid)
16

>>> with napari.gui_qt():
>>>     # set up viewer
>>>     viewer = napari.Viewer()
>>>     viewer.add_pyramid(pyramid)
---------------------------------------------------------------------------
[...]

Issue #537, opened 2019-09-16
Auto-create pyramid when image exceeds expected capabilities


ambrosejcarr:
## 🚀 Feature
I would like napari to create pyramids for me if I pass an image that's too large for it to handle. 

## Motivation

Right now, napari throws a difficult-to-parse exception when I pass it an image to display that's ~ 20k x 20k. The solution is known, and is simple, but it's not declared in the exception: 

### Suggested Solution:
> should be one line with from `skimage.transform import pyramid_gaussian` - https://scikit-image.org/docs/dev/auto_examples/transform/plot_pyramid.html

## Pitch

Given the simplicity of the solution, Napari should just solve this problem, optionally emitting a warning to tell the user that it's doing something in the background.

sofroniewn:
So one option here would be to drop the `add_pyramid` method entirely and integrate all the `Pyramid` layer functionality into the `Image` layer and then allow for `add_image` to accept a list of arrays too (like `add_pyramid`) does or add a flag that allows us to autocompute pyramids if we think we need to. I think this may be cleaner than having multiple layer types. Thoughts @jni?

We could then subclass this new layer for the `Labels` layer, thereby giving `Labels` multiresolution support largely for free (though we have to think about lazy painting in a labels layer as discussed in #470

sofroniewn:
After discussion at team meeting we will merge pyramid layer into image layer, adding keyword arg that says `is_pyramid`, and guess otherwise, and maybe adding a keyword arg that allows us to try and compute a pyramid if we think we need to (say an image >20k x 20k or something)

We'll also try to extend so that labels now supports pyramid - but will have to think about painting in the pyramid a little

sofroniewn:
Oh and also 3D rendering isn't supported for pyramids yet, so might need to think about that too.

Also if people request 3D rending on too large images we might need to do some pyramid stuff too ...

For now we can just show the top layer of the pyramid in 3D

ambrosejcarr:
Fantastic, thanks for the rapid discussion and design.

sofroniewn:
This is now done for images in #545. We still need to make it work for labels, but that is discussed in #538 and will be started soon

PR #545, opened 2019-09-24
Merge pyramid layer into image

### LINKED ISSUE #436: Proper suport for scales in 2D and 3D layers

sofroniewn:
# Description
This PR begins to address #537 by merging the pyramid layer into the image layer. It removes the layer called `Pyramid` and the `add_pyramid` and `view_pyramid` methods in favor of just using `Image`. Right now you can pass `add_image(data, pyramid=True)` to force the data to be considered as a pyramid. If you just pass `add_image(data)` it will assume that the data is a pyramid if it is a list (that may be too strong an assumption in case someone wants to pass a list of images an consider them as an array, but it's an easy fix).

Right now we will not convert an image to a pyramid when necessary, but as that was the main request behind #537 from @ambrosejcarr I will work on that next. I'll probably make it so that this is conversion happens on the inside of napari i.e. if you do `layer.data` you don't get a pyramid back you get the original array back. Also if someone passes a large array and `pyramid` is set to `True` should that force the computation of the pyramid if it is not a pyramid?

This PR also cleans up some of the scale stuff and closes #436 by now including a `_translate_view` and `_scale_view` tuples too.

Finally this does not yet address #538 which also request the labels layer support pyramids, but that will come soon. Maybe in this PR or more likely in a follow-up PR.

Probably the remaining two things for this PR are to improve our guessing on if something already is a pyramid, and autogenerate a pyramid for a non-pyramid when we need to or are requested to.

@constantinpape do you have any comments on this PR?

## Type of change

# How has this been tested?

- [x] moved over the `add_pyramid.py` tests

## Final checklist:


sofroniewn:
Oh I also the `contrast_limits` calculation for pyramids now only uses the top layer, so will be much faster if you don't supply `contrast_limits` explicitly. We still have a bit of an issue for large dask arrays and `contrast_limits` calculations, but we can address that in a different PR

sofroniewn:
This PR is now ready for review @jni. @ambrosejcarr would you like to test it too. 

I've added a more of an advanced `is_pyramid` method, that will ask if the data is a list and if the shapes are decreasing in size in some way across the list, though right now we don't support lists of arrays of image-like inputs either so maybe this was too much.

I've also added auto-generation of the image pyramid from a single array when needed. You can still access the original array with the `layer.data` property, but the pyramid will be at `layer._data_pyramid`. I decided to make the call on whether to pyramid or not based on a single axis, at the fairly conservative value of 2^13 or 8192 as this approach scales well to nD data, especially for cases where there are some very large dimensions and some very small dimensions. There might be some weird corner cases where there are a bunch of very large dimensions with very different aspect ratios, but at that point I really think people should be providing their own pyramids. This should work in all the basic cases though. I've also added a bunch of tests / doc strings etc.

I havn't extended support of pyramids to the labels layer, but I'd like to do that after this get's merged.

examples/nD_pyramid.py
```diff
@@ -2,26 +2,20 @@
...
 
-# create pyramid from astronaut image
-astronaut = data.astronaut()
-base = np.tile(astronaut, (3, 3, 1)).astype('float')
-base = np.round(np.array([base * (16 - i) / 16 for i in range(16)])).astype(
-    np.uint8
-)
+# create pyramid from random data
+base = np.random.random((1536, 1536))
+base = np.array([base * (8 - i) / 8 for i in range(8)])
```
jni, line 12: what's going on here???

napari/_vispy/vispy_image_layer.py
```diff
@@ -93,6 +93,102 @@ def _on_contrast_limits_change(self):
...
+            top_left, 0, np.subtract(self.layer.level_shapes[0], 1)
+        )
+
+        # Convert to offset for image array
+        rounding_factor = self.layer._max_tile_shape / 4
+        top_left = rounding_factor * np.floor(top_left / rounding_factor)
+
+        return top_left.astype(int)
+
+    def on_draw(self, event):
```
jni, line 178: Could this be renamed `on_draw_pyramid` and only get connected when the image is a pyramid? It's a lot of extra calls when displaying a random image, which won't be super expensive but still feels wasteful, especially one of the main design goals of napari is to be blazing fast.

napari/layers/image/image.py
```diff
@@ -18,13 +22,21 @@ class Image(Layer):
 
     Parameters
     ----------
-    data : array
+    data : array or list of array
         Image data. Can be N dimensional. If the last dimension has length
-        3 or 4 can be interpreted as RGB or RGBA if rgb is `True`.
+        3 or 4 can be interpreted as RGB or RGBA if rgb is `True`. If a
+        list and arrays are decreaing in shape then the data is treated as
```
jni: decreasing

napari/components/viewer_model.py
```diff
@@ -354,13 +355,21 @@ def add_image(
 
         Parameters
         ----------
-        data : array
+        data : array or list of array
             Image data. Can be N dimensional. If the last dimension has length
-            3 or 4 can be interpreted as RGB or RGBA if rgb is `True`.
+            3 or 4 can be interpreted as RGB or RGBA if rgb is `True`. If a
+            list and arrays are decreaing in shape then the data is treated as
```
jni: decreasing

napari/layers/image/image.py
```diff
@@ -93,12 +111,14 @@ class Image(Layer):
     """
 
     _colormaps = AVAILABLE_COLORMAPS
+    _max_tile_shape = 1600
```
jni, line 117: seems small?

jni: for a hardcoded value I mean.

napari/layers/image/image.py
```diff
@@ -143,7 +154,15 @@ def __init__(
         )
 
         # Set data
+        self.pyramid = pyramid
```
jni: I'd call this `is_pyramid`, as otherwise it seems like it holds the pyramid as opposed to e.g. image/data holding the full-res image.

napari/layers/image/image.py
```diff
@@ -314,6 +360,8 @@ def rendering(self, rendering):
 
     def _set_view_slice(self):
         """Set the view given the indices to slice with."""
+        nd = self.dims.not_displayed
```
jni: I don't like this abbreviation, as it usually means "n-dimensional"

napari/layers/image/image.py
```diff
@@ -322,16 +370,74 @@ def _set_view_slice(self):
...
-        image = np.asarray(self.data[self.dims.indices]).transpose(order)
+        if self.pyramid:
+            # If 3d redering just show lowest level of pyramid
+            if self.dims.ndisplay == 3:
+                self.data_level = len(self._data_pyramid) - 1
+
+            # Slice currently viewed level
+            level = self.data_level
+            indices = np.array(self.dims.indices)
+            downsampled = indices[nd] / self.level_downsamples[level, nd]
```
jni: Maybe call this `downsampled_indices`, `downsampled_idxs`, or `curr_level_indices`, or similar? Downsampled is the name we gave to downsampled images elsewhere, so this whole section was super-confusing to me for a bit.

napari/layers/image/tests/test_pyramid.py
```diff
@@ -292,8 +292,24 @@ def test_xml_list():
...
+    data = np.random.random(shape)
+    layer = Image(data)
+    assert np.all(layer.data == data)
+    assert layer.pyramid == True
+    assert layer._data_pyramid[0].shape == shape
+    assert layer._data_pyramid[1].shape == (shape[0] / 2, shape[1])
+    assert layer.ndim == len(shape)
+    assert layer.shape == shape
+    assert layer.rgb == False
+    assert layer._data_view.ndim == 2
```
jni, line 330: maybe add a test for `test_infer_pyramid` that you pass in a list of decreasing arrays to, without `pyramid=True`, and it should correctly set `layer.[is_]pyramid`?

napari/util/misc.py
```diff
@@ -58,6 +59,135 @@ def is_rgb(shape):
         return False
 
 
+def is_pyramid(data):
+    """If data is a list of arrays of decreaing size.
```
jni: decreasing. Is this a butterfly keyboard thing? =P

napari/util/misc.py
```diff
@@ -58,6 +59,135 @@ def is_rgb(shape):
...
+    """
+    if isinstance(data, list):
+        size = [np.prod(d.shape) for d in data]
+        return np.all(size[:-1] > size[1:])
+    else:
+        return False
+
+
+def should_be_pyramid(shape):
+    """Check if any data axes needs to be pyramided
```
jni: I think the correct verb is pyramidified. =P

napari/util/misc.py
```diff
@@ -58,6 +59,135 @@ def is_rgb(shape):
...
+    ----------
+    shape : tuple of int
+        Shape of data to be tested
+
+    Returns
+    -------
+    pyr_axes : tuple of bool
+        If any axis is above pyramiding threshold.
+    """
+    return np.log2(shape) > 13
```
jni: This function appears to be untested: testing `if should_be_pyramid((45, 435))` will raise an error. Should be `np.any(np.log2(shape) > 13)`. At any rate, `max_size` or something should be a parameter. Doesn't vispy allow up to 16k x 16k, ie log2(shape) == 14?

jni: Oh, I see, just saw its use later. Then change the doc to read, "True wherever an axis exceeds the pyramid threshold." And the type returned is an array of bool.

Finally, shouldn't there also be a check for the total size? ie 8k x 8k x 8k will give "should be pyramided" False everywhere, but vispy won't like it?

napari/util/misc.py
```diff
@@ -58,6 +59,135 @@ def is_rgb(shape):
...
+        else:
+            data_pyramid = None
+    else:
+        data_pyramid = data
+
+    return ndim, rgb, pyramid, data_pyramid
+
+
+def fast_pyramid(data, downscale=2, max_layer=None):
+    """Compute fast image pyramid.
```
jni, line 160: Cool! =) Could you add a note in a paragraph below this and before parameters, that this simply subsamples, rather than downsamples, the input image?

napari/view_layers.py
```diff
@@ -24,13 +25,21 @@ def view_image(
 
     Parameters
     ----------
-    data : array
+    data : array or list of array
         Image data. Can be N dimensional. If the last dimension has length
-        3 or 4 can be interpreted as RGB or RGBA if rgb is `True`.
+        3 or 4 can be interpreted as RGB or RGBA if rgb is `True`. If a
+        list and arrays are decreaing in shape then the data is treated as
```
jni: decreasing

jni (approved):
@sofroniewn very cool! I've left some very minor comments, feel free to merge once they have been suitably addressed and/or ignored. =P

examples/nD_pyramid.py
sofroniewn: I wanted to introduce some variation into the random image, which I think is better than the tiled astronaut. People found it confusing what that represented. We can still improve this further, but I think it is fine for now

napari/_vispy/vispy_image_layer.py
sofroniewn: Checking that if statement is really negligible I think. This only gets called when the image changes or the mouse moves, and will make it easier if a layer switches from being a pyramid to a non-pyramid, or visa versa. Otherwise we need to connect / disconnect. I can still change if there is some measurable performance difference, but I think it will be in the microseconds and not happening always.

napari/layers/image/image.py

napari/components/viewer_model.py

napari/layers/image/image.py
sofroniewn: if the tiles are larger it can be more expensive to fetch them, and also at 1-1 resolution matching with the screen 1600 x 1600 fills our viewport, but maybe I'm missing something. Why would you expect it to be larger?

napari/layers/image/image.py
sofroniewn: should the keyword arg be `is_pyramid` too then? I think so.

Also should it be `is_rgb` instead of `rgb`? I copied that convention here.

napari/layers/image/image.py
sofroniewn: replaced with `not_disp`

napari/util/misc.py
sofroniewn: :-) fixed - i copy and paste a lot??

napari/util/misc.py
sofroniewn: love it - changed

napari/util/misc.py
sofroniewn: if we do 2D rendering then 8k x 8k x 8k will always be fine - but yes if we go to 3D mode it will choke. I'm not sure how to think about 3D mode

napari/util/misc.py
sofroniewn: done

napari/view_layers.py
sofroniewn: done

napari/layers/image/image.py
sofroniewn: yeah, agreed - changed to `downsampled_indices`

napari/layers/image/tests/test_pyramid.py
sofroniewn: yup, good catch, done

napari/util/misc.py
sofroniewn: I think the cut off for 3D rendering might be more like 1k x 1k x 1k, and I don't want to autopyramid everything that's so big as people might not ever want to look at it in 3D. I could do a downsample on the vispy side just before sending to the graphics card so we don't crash, but I can do that in a new PR. Thoughts @jni?

napari/layers/image/image.py
sofroniewn: i've replaced `is_pyramid` everywhere including keyword arg (i like the keyword arg and the property being the same), but i've left `rgb` but can switch to `is_rgb` in a follow up PR. Thoughts @jni?

sofroniewn:
I have now made most of the changes requested. I might ask @jni to take one more look to see if you like that I made `is_pyramid` the keyword arg too. Also to get your thoughts on the `is_rgb` as I could change that in this PR too. I also left a couple of your requests unresolved with reasons, so if you feel strongly about any let me know

examples/nD_pyramid.py
jni: What I don’t understand is why this improves on just the random image in any way. Why go through the trouble of multiplying a random image by random numbers?

jni: Oh, wait, — does this add a third dimension? And if so, why not just add a third dimension to the random call?

jni: (I agree that the tiling was confusing, btw)

napari/_vispy/vispy_image_layer.py
jni: I worry that all the stuff happening during mouse moves is adding up and will continue to do so. When will a layer switch “live” rather than at creation time?

napari/layers/image/image.py
jni: Does it fill the viewport in a Retina display? What about a 4K monitor? Anyway I’m fine with it but I would just think it should be configurable...

napari/layers/image/image.py
jni: I think RGB is ok. I found self.pyramid confusing because it seems like that should point to the actual pyramid. But I don’t think one would expect self.rgb to hold the rgb image.

napari/util/misc.py
jni: Unclear, let’s punt on this. My intuition is to add a displayed_dims keyword argument that changes the computation, but it wouldn’t be fun to hook that up. We can rely on users for now.

jni:
All good! I’m still confused about the random image but that’s not a big deal.

PR #546, opened 2019-09-26
Add release notes

sofroniewn:
# Description
This PR adds release notes for our previous `0.1.0` release and some intermediate release. It adds instructions to our release guide on how to generate the release notes and a file `generate_release_notes.py` for doing so. In general the approach we have taken mirrors that taken by scikit-image.

We also now have a `release_template.rst` that is a blank template for future release notes, and a 
`release_dev.rst` that is an in progress copy of the notes for the next release.

The approach I took was for minor releases (something like `0.1`) to generate all the release notes since the last minor release (say `0.0`). Then for normal releases, like `0.1.4` to generate release notes since the last time release notes were generated (which should be `0.1.3` in this case).

How does this sound @AhmetCanSolak @royerloic @jni?

I'm less concerned with retroactively making all these release notes super accurate (though I'm glad we have something), but I'm more interested in establishing good practices going forward.

### DIFF
--- a/docs/release/release_0_1.rst
+++ b/docs/release/release_0_1.rst
@@ -1,0 +1,160 @@
+Announcement: napari 0.1.0
+==========================
+
+We're happy to announce the release of napari v0.1.0!
+napari is a fast, interactive, multi-dimensional image viewer for Python.
+It's designed for browsing, annotating, and analyzing large multi-dimensional
+images. It's built on top of Qt (for the GUI), vispy (for performant GPU-base
+rendering), and the scientific Python stack (numpy, scipy).
+
+This is our first major release, timed for the 2019 SciPy Conference in Austin.
+It marks our transition from pre-alpha to alpha, and establishes a reasonable
+API for adding images, shapes, and other basic layer types to an interactive
+viewer. It supports launching a viewer with python scripting or from Jupyter
+notebooks.
+
+For more information, examples, and documentation, please visit our website:
+https://github.com/napari/napari
+
+Other Pull Requests
+*******************
+- Add shapes (#100)
+- Vectors Layer (#129)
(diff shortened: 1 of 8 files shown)


docs/release/release_0_1.rst
```diff
@@ -0,0 +1,160 @@
...
+This is our first minor release, timed for the 2019 SciPy Conference in Austin.
+It marks our transition from pre-alpha to alpha, and establishes a reasonable
+API for adding images, shapes, and other basic layer types to an interactive
+viewer. It supports launching a viewer with python scripting or from Jupyter
+notebooks.
+
+For more information, examples, and documentation, please visit our website:
+https://github.com/napari/napari
+
+Other Pull Requests
```
jni: If you're not going to do some cherry picking here, I suggest changing this heading to just "Pull Requests"

jni (approved):
@sofroniewn I've left some minor comments. Bigger comments:

- in skimage we typically keep all the 0.x.y in the same 0.x file, with newer ones on top. Not a huge deal if you prefer to keep things separate
- in skimage we are now having a discussion about whether to follow NumPy in switching to Towncrier. But we can cross that bridge later, just thought I'd flag it.

docs/release/release_0_1.rst
sofroniewn: will change

sofroniewn:
@jni I thought for now because we're making so many 0.x.y releases and some of them might be breaking / quite big, better to give them their own files. Maybe as we get more stable we can move over to the practice where each 0.x gets its own.

Towncrier could be good, thanks for the flag. I hadn't heard of it.

I will make the change you requested and then merge.

I will then add similar release notes for `0.2` to #557 before merging that one

PR #548, opened 2019-09-28
Labels pyramid

### LINKED ISSUE #538: Help with starfish.display for large images & mask collections

sofroniewn:
# Description
This PR extends pyramid support to the Labels layer following the same conventions as the Image. It does this by making Labels subclass Image which actually reduces a lot of duplicate code.

I also make the thumbnails now always be blended into a black background.

It also introduces a new concept of layers being "editable" or not and sets the Labels layer to not being editable when it is a pyramid or being rendered in 3D (which would have caused bugs). One day we may extend painting to the pyramid or 3D but for now those are more advanced features. People have also requested the ability to disable editing for datasets where it doesn't make sense. When disabled the paint / fill / pick buttons now have a different background color to indicate they are no longer selectable. If people are ok with this I will extend the editable property to the shapes and points layers where we also need to disable editing for 3D rendering.

Here is what this looks like for the 3D rendering case:

editable_labels

This closes #538. @ambrosejcarr does this look good to you? 

## Type of change

# How has this been tested?

- [x] added pyramid test to the Labels tests

## Final checklist:

### DIFF
--- a/napari/layers/image/tests/test_pyramid.py
+++ b/napari/layers/image/tests/test_pyramid.py
@@ -273,13 +273,13 @@ def test_metadata():
 
 def test_value():
     """Test getting the value of the data at the current coordinates."""
-    shapes = [(40, 20), (20, 10), (10, 5)]
+    shapes = [(40, 20), (20, 10)]
     np.random.seed(0)
     data = [np.random.random(s) for s in shapes]
     layer = Image(data, is_pyramid=True)
     value = layer.get_value()
     assert layer.coordinates == (0, 0)
-    assert value == (2, data[-1][0, 0])
+    assert value == (1, data[-1][0, 0])
 
 
 def test_message():
--- a/napari/layers/image/tests/test_pyramid.py
+++ b/napari/layers/image/tests/test_pyramid.py
@@ -273,13 +273,16 @@ def test_metadata():
 
 def test_value():
     """Test getting the value of the data at the current coordinates."""
-    shapes = [(40, 20), (20, 10)]
+    shapes = [(40, 20), (20, 10), (10, 5)]
     np.random.seed(0)
     data = [np.random.random(s) for s in shapes]
     layer = Image(data, is_pyramid=True)
     value = layer.get_value()
     assert layer.coordinates == (0, 0)
-    assert value == (1, data[-1][0, 0])
+    # Note that here, because the shapes of the pyramid are all very small
+    # data that will be rendered will only ever come from the bottom two
+    # levels of the pyramid.
+    assert value == (1, data[1][0, 0])
 
 
 def test_message():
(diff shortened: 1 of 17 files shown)


sofroniewn:
I'v gone ahead and extended "editable" mode to Points and Shapes within this PR.

napari/layers/image/tests/test_pyramid.py
```diff
@@ -273,13 +273,13 @@ def test_metadata():
...
     """Test getting the value of the data at the current coordinates."""
-    shapes = [(40, 20), (20, 10), (10, 5)]
+    shapes = [(40, 20), (20, 10)]
     np.random.seed(0)
     data = [np.random.random(s) for s in shapes]
     layer = Image(data, is_pyramid=True)
     value = layer.get_value()
     assert layer.coordinates == (0, 0)
-    assert value == (2, data[-1][0, 0])
+    assert value == (1, data[-1][0, 0])
```
jni: Why this change?

jni (approved):
Brilliant! I love how much the code is simplified by this change! :tada:

My only concern is that the buttons styling to me suggests "pressed" rather than "disabled". Typically this is shown by buttons having lower contrast, but currently that would make it confusing compared to the styling for "selected". We can leave button styling details for a later PR though.

napari/layers/image/tests/test_pyramid.py
sofroniewn: I added the ability of the layer to "trim" the pyramid so that if you pass it say a pyramid that keeps going until the final layer is `1x1` the smallest resolution it would actually show is `63x63` (which is also the resolution it would use for the thumbnail). If all the layers are very small then the pyramid just takes the bottom two, which it is doing here. I've added a comment inside the test.

sofroniewn:
fair enough on the stylings, but let's wait till a second review with @liaprins-czi to improve

royerloic:
Very cool @sofroniewn !

PR #549, opened 2019-09-28
fix 3d point rendering

sofroniewn:
# Description
This PR fixes a known vispy bug in the size of points during 3D rendering. I've made a PR to vispy https://github.com/vispy/vispy/pull/1702 to fix this, but it might be a little before it is merged and they make a new release. At that point we can remove this fix. 

## Type of change


jni (approved):
Nice!

Issue #469, opened 2019-08-08
napari design suggestions


sofroniewn:
## ❓ Questions and Help

napari has the opportunity to get some design input from a couple designers at CZI. I'm creating this issue for us to discuss current design issues, suggestions, feedback, etc.

kevinyamauchi:
What an awesome opportunity! I have listed a few things below that I think we could use some feedback on:

- **Usage pattern for the sidebars**: Currently, we have a sidebar on either side of the the canvas and we are not consistent in how each layer utilizes the UI elements in each one. For example, should one sidebar be dedicated for parameters that are applied to the entire layer and the other sidebar is for adjusting parameters that can be specific to elements in a layer? 
- **Labeling for the dim sliders**:  the dim sliders are confusing because don't have any labels. I think it would be useful to get some advice on the best way to label them (e.g., where to denote the axis, should we use ticks)
- **Standardization of UI elements**: I think it would be useful to set some "rules" for when to use what type of UI element (e.g., when to use a dropdown vs. radio buttons or slider vs. text box).

sofroniewn:
@liaprins - has completed her design audit of napari and produced a truly incredible document with the feedback, included here as a pdf.

[Napari design audit.pdf](https://github.com/napari/napari/files/3598847/Napari.design.audit.pdf)

@kne42, @freeman-lab and I went over some of the main takeaways from the report today with Lia and we'll be going through it in more detail over the next week or two as we think about next steps etc. We can talk about it at our next team meeting and maybe we can get @liaprins to present on it next time she's at Biohub on a Thursday, as there's a lot to digest, but I encourage everyone (@royerloic @jni @kevinyamauchi @AhmetCanSolak @bryantChhun) to take a look.

I think we can then use a combination of discussion on this issue / creating new issues as needed to work through the feedback. During the meeting we had some good ideas about ways to unify the layer control buttons / sliders on the left and the layer properties that appear when you double click on a layer that I will create a new issue to discuss.

Finally I'll add that I'm incredibly excited to have this sort of feedback from a design expert as it can be rare to get it in a lot of academic / open-source projects and I think will make an enormous difference to the usability and ultimate success of napari.

liaprins-czi:
@liaprins-czi

liaprins:
Mentioning my CZI Github, so I can follow from it @liaprins-czi

sofroniewn:
[...]

PR #550, opened 2019-09-29
make dims sliders bars

sofroniewn:
# Description
This PR follows on making improvements suggested in our design audit, see #469 by @liaprins-czi and by others.

Here we're improving some tooltips over our buttons and making the dimension sliders be like scroll bars instead of sliders. This means for now we are moving away from the range slider, but we havn't been using the range slider functionality. If we want to start using the interval mode with max projections etc. like @royerloic has requested in the past then we can think about how to enable switching between that mode and the scroll bar, but for now I think the scroll bar works pretty well.

See here for a gif. One consequence is that right away you get a sense for which dimension you might be interacting with based on the width of the scroll bar:

scroll_bar_dims

## Type of change

## Final checklist:


jni (approved):
:heart_eyes: wow that looks so amazing! :tada: awesome work!

I only have one suggestion: the bar should have a minimum thinness, just like scroll elements or tabs in browsers have a minimum, otherwise they become unusable.

sofroniewn:
Thanks. The scroll bar has a minimum width right now of 26 pixels, we can change that in the stylesheet if people want, but it seemed pretty good for now. I will merge

PR #551, opened 2019-09-29
stop vispy error catching

sofroniewn:
# Description
As had been previously observed in our own event system which is based on the one from vispy, the default behavior of the event system is to ignore errors. We want this behavior turned off so that we know about errors. This PR now does this for the vispy event system on the canvas. Doing so exposed a couple errors on my local tests that I then fixed. Doing so in #544 exposed some windows failures that might actually be causing real user errors that we were previously missing - see #527 and #428.

## Type of change

### DIFF
--- a/napari/_qt/qt_viewer.py
+++ b/napari/_qt/qt_viewer.py
@@ -72,6 +72,7 @@ class QtViewer(QSplitter):
             self.viewerButtons.consoleButton.setEnabled(False)
 
         self.canvas = SceneCanvas(keys=None, vsync=True)
+        self.canvas.events.ignore_callback_errors = False
         self.canvas.native.setMinimumSize(QSize(200, 200))
         self.canvas.context.set_depth_func('lequal')
 
@@ -350,8 +351,8 @@ class QtViewer(QSplitter):
     def on_draw(self, event):
         """Called whenever drawn in canvas. Called for all layers, not just top
         """
-        for layer in self.viewer.layers:
-            self.layer_to_visual[layer].on_draw(event)
+        for visual in self.layer_to_visual.values():
+            visual.on_draw(event)
 
     def keyPressEvent(self, event):
         self.canvas._backend._keyEvent(self.canvas.events.key_press, event)


sofroniewn:
Our windows build now fails, which is great news as this error was happening in the wild - see #527 and we weren't really aware of it. Now we can have some hope that if our windows tests start passing again napari will actually work for our windows users.

Here's the stack trace that we get when our tests fail:
```python
---------------------------- Captured stderr call -----------------------------
WARNING: Error drawing visual <Image at 0x1200da71da0>
Qt exceptions in virtual methods:
________________________________________________________________________________
  File "c:\users\containeradministrator\anaconda\lib\site-packages\vispy\app\backends\_qt.py", line 818, in paintGL
    self._vispy_canvas.events.draw(region=None)
  File "c:\users\containeradministrator\anaconda\lib\site-packages\vispy\util\event.py", line 455, in __call__
  File "c:\users\containeradministrator\anaconda\lib\site-packages\vispy\util\event.py", line 475, in _invoke_callback
    self, cb_event=(cb, event))
  File "c:\users\containeradministrator\anaconda\lib\site-packages\vispy\util\event.py", line 471, in _invoke_callback
    cb(event)
  File "c:\users\containeradministrator\anaconda\lib\site-packages\vispy\scene\canvas.py", line 217, in on_draw
    self._draw_scene()
  File "c:\users\containeradministrator\anaconda\lib\site-packages\vispy\scene\canvas.py", line 266, in _draw_scene
    self.draw_visual(self.scene)
  File "c:\users\containeradministrator\anaconda\lib\site-packages\vispy\scene\canvas.py", line 304, in draw_visual
    node.draw()
  File "c:\users\containeradministrator\anaconda\lib\site-packages\vispy\scene\visuals.py", line 99, in draw
    self._visual_superclass.draw(self)
  File "c:\users\containeradministrator\anaconda\lib\site-packages\vispy\visuals\visual.py", line 443, in draw
    self._vshare.index_buffer)
  File "c:\users\containeradministrator\anaconda\lib\site-packages\vispy\visuals\shaders\program.py", line 101, in draw
    Program.draw(self, *args, **kwargs)
  File "c:\users\containeradministrator\anaconda\lib\site-packages\vispy\gloo\program.py", line 533, in draw
    canvas.context.flush_commands()
  File "c:\users\containeradministrator\anaconda\lib\site-packages\vispy\gloo\context.py", line 175, in flush_commands
    self.shared.parser.parse([('CURRENT', 0, fbo)])
  File "c:\users\containeradministrator\anaconda\lib\site-packages\vispy\gloo\glir.py", line 819, in parse
    self._parse(command)
  File "c:\users\containeradministrator\anaconda\lib\site-packages\vispy\gloo\glir.py", line 743, in _parse
    self._gl_initialize()
  File "c:\users\containeradministrator\anaconda\lib\site-packages\vispy\gloo\glir.py", line 851, in _gl_initialize
    if this_version < '2.1':
  File "c:\users\containeradministrator\anaconda\lib\distutils\version.py", line 52, in __lt__
    c = self._cmp(other)
  File "c:\users\containeradministrator\anaconda\lib\distutils\version.py", line 335, in _cmp
    if self.version == other.version:
AttributeError: 'LooseVersion' object has no attribute 'version'
``` 

This trace has been discussed in a number of vispy issues, including https://github.com/vispy/vispy/issues/1503, https://github.com/vispy/vispy/issues/1342, https://github.com/vispy/vispy/issues/1450, and most recently https://github.com/vispy/vispy/issues/1688, but never seems to be have been fully resolved.

jni (approved):
It's such terrible distutils behaviour... Searching for that exception string finds all kinds of unrelated errors, which are maybe distutils silencing a real stacktrace and then replacing it with this terrible one?

Anyway, @sofroniewn are you suggesting that we should merge and accept our broken windows build for now? I'll approve to let you get on with your work, but it makes my OCD unhappy. =P

sofroniewn:
Thanks for approving, but I think I'd rather hold off on the merge - it's such a trivial change we can make it at any point, and it will only cause all our windows builds to break and code-cov not to run. I will try and get my hands on a windows machine next week and investigate there.

jni:
@sofroniewn just reviewing open PRs and found this... Which I thought was the reason our windows builds were failing! :man_facepalming: Should we just merge, since windows isn't working anyway?

sofroniewn:
yup, will merge

tlambert03:
Hey folks, sorry if I'm a little late to the game here and this is all known... but I was just playing a bit with windows builds both on a local Win10 machine (where the tests in the master branch always pass) and on Cirrus, where they of course fail.  I found that enabling the pytest setting `qt_no_exception_capture = 1` allows the builds to pass on Cirrus ([example](https://cirrus-ci.com/task/5098963358711808)).  I'm not that familiar with this [pytest-qt virtual methods issue](https://pytest-qt.readthedocs.io/en/latest/virtual_methods.html), and it's still not clear to me whether `qt_no_exception_capture` is tantamount to just skipping all the important tests anyway (as this PR was trying to avoid in the first place), or whether those exceptions would be unlikely to be napari bugs (making it a potentially useful option).  But since I didn't see that pytest option discussed anywhere in these issues or in the vispy ones, i just wanted to check

jni:
I don't really understand what they're saying, nor why *disabling this* would actually cause the tests to pass rather than the interpreter to crash in a different way. It could be that their exception capture is actually broken, and that's the thing causing our crashes. For me that's the only reasonable explanation, since, according to that documentation, exceptions raised in our own custom methods will crash the interpreter during testing, and that's clearly not what we're observing.

All this to say: I think it's probably ok to disable it! @sofroniewn?

PR #552, opened 2019-09-30
Add anisotropy factors to points

### LINKED ISSUE #480: Specifying nD point sizes
Currently, point sizes can be specified with a 1D array of one uniform size for each point, or a 1D array of anisotropic sizes (one size per dimension), or a 2D array specifying a different anisotropic size per point. This convention can be convenient but it can also be confusing: if I have 5 5D points, and I pass `size=[5, 8, 2, 2, 1]`, am I specifying five isotropic point sizes, or a single anisotropic size for all points?

scikit-learn once had this problem with feature vectors, and decided to enforce 2D-only input for their features, to eliminate ambiguities and subtle bugs.

Another alternative is to have point sizes be scalars an anisotropy be a separate attribute that is universal for a layer.

xref: https://github.com/napari/napari/pull/451#discussion_r312808367

sofroniewn:
# Description
This PR closes #480 by introducing an `anisotropy` keyword arg to the Points layer and making the points sizes a 1-D array to resolve potential broadcasting ambiguity. The `anisotropy` factor is global for the layer (i.e. applies to all Points), but I think that is a good abstraction.

## Type of change

# How has this been tested?

- [x] added anisotropy test to `test_points.py`

## Final checklist:

### DIFF
--- a/examples/nD_points.py
+++ b/examples/nD_points.py
@@ -32,5 +32,5 @@ with napari.gui_qt():
         ]
     )
     viewer.add_points(
-        points, size=[0, 6, 10, 10], face_color='blue', n_dimensional=True
+        points, anisotropy=[0, 0.5, 1, 1], size=10, face_color='blue', n_dimensional=True
     )
--- a/napari/components/viewer_model.py
+++ b/napari/components/viewer_model.py
@@ -557,14 +557,15 @@ class ViewerModel(KeymapMixin):
             Symbol to be used for the point markers. Must be one of the
             following: arrow, clobber, cross, diamond, disc, hbar, ring,
             square, star, tailed_arrow, triangle_down, triangle_up, vbar, x.
-        size : float, tuple
+        size : float, tuple, list, array (N,)
             Size of the point marker. If given as a scalar, all points are made
-            the same size. If given as a tuple, size must be the same length
-            as the number of points.
-        anisotropy : tuple, optional
+            the same size. If given as a tuple, list, or 1-D array size must be
+            the same length as the number of points.
+        anisotropy : tuple, list, array (D,), optional
             List of anisotropy factors, must be one per dimension. These act as
-            pre-multipliers on the size of each point. If not provided,
-            defaults as 1 for each dimension.
+            pre-multipliers on the size of each point. Note that the anisotropy
+            is only displayed for dimensions out of plane for the view. If not
+            provided, defaults as 1 for each dimension.
         edge_width : float
             Width of the symbol edge in pixels.
         edge_color : str
--- a/napari/components/viewer_model.py
+++ b/napari/components/viewer_model.py
@@ -534,6 +534,7 @@ class ViewerModel(KeymapMixin):
         *,
         symbol='o',
         size=10,
+        anisotropy=None,
         edge_width=1,
         edge_color='black',
         face_color='white',
@@ -556,10 +557,14 @@ class ViewerModel(KeymapMixin):
             Symbol to be used for the point markers. Must be one of the
             following: arrow, clobber, cross, diamond, disc, hbar, ring,
             square, star, tailed_arrow, triangle_down, triangle_up, vbar, x.
-        size : float, array
+        size : float, tuple
             Size of the point marker. If given as a scalar, all points are made
-            the same size. If given as an array, size must be the same
-            broadcastable to the same shape as the data.
+            the same size. If given as a tuple, size must be the same length
+            as the number of points.
+        anisotropy : tuple, optional
+            List of anisotropy factors, must be one per dimension. These act as
+            pre-multipliers on the size of each point. If not provided,
+            defaults as 1 for each dimension.
         edge_width : float
             Width of the symbol edge in pixels.
         edge_color : str
@@ -600,6 +605,7 @@ class ViewerModel(KeymapMixin):
             data,
             symbol=symbol,
             size=size,
+            anisotropy=anisotropy,
             edge_width=edge_width,
             edge_color=edge_color,
             face_color=face_color,
--- a/napari/layers/points/points.py
+++ b/napari/layers/points/points.py
@@ -23,10 +23,14 @@ class Points(Layer):
         Symbol to be used for the point markers. Must be one of the
         following: arrow, clobber, cross, diamond, disc, hbar, ring,
         square, star, tailed_arrow, triangle_down, triangle_up, vbar, x.
-    size : float, array
+    size : float, tuple
         Size of the point marker. If given as a scalar, all points are made
-        the same size. If given as an array, size must be the same
-        broadcastable to the same shape as the data.
+        the same size. If given as a tuple, size must be the same length
+        as the number of points.
+    anisotropy : tuple, optional
+        List of anisotropy factors, must be one per dimension. These act as
+        pre-multipliers on the size of each point. If not provided,
+        defaults as 1 for each dimension.
     edge_width : float
         Width of the symbol edge in pixels.
     edge_color : str
@@ -62,6 +66,9 @@ class Points(Layer):
     size : float
         Size of the marker for the next point to be added or the currently
         selected point.
+    anisotropy : tuple
+        List of anisotropy factors, must be one per dimension. These act as
+        pre-multipliers on the size of each point.
     edge_width : float
         Width of the marker edges in pixels for all points
     edge_color : str
@@ -79,9 +86,10 @@ class Points(Layer):
         n-dimensions according to specified point marker size.
     selected_data : list
         Integer indices of any selected points.
-    sizes : array (N, D)
-        Array of sizes for each point in each dimension. Must have the same
-        shape as the layer `data`.
+    sizes : array (N,)
+        Array of sizes for each point. Get multiplied with the anisotropy
+        factor for each dimension to create the size for each point in each
+        dimension.
     mode : str
         Interactive mode. The normal, default mode is PAN_ZOOM, which
         allows for normal interactivity with the canvas.
@@ -118,6 +126,7 @@ class Points(Layer):
         *,
         symbol='o',
         size=10,
+        anisotropy=None,
         edge_width=1,
         edge_color='black',
         face_color='white',
@@ -163,6 +172,8 @@ class Points(Layer):
         self.symbol = symbol
         self._n_dimensional = n_dimensional
         self.edge_width = edge_width
+
+        self._anisotropy = anisotropy or np.ones(self.dims.ndim)
 
         # The following point properties are for the new points that will
         # be added. For any given property, if a list is passed to the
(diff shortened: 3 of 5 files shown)


napari/components/viewer_model.py
```diff
@@ -556,10 +557,14 @@ def add_points(
             Symbol to be used for the point markers. Must be one of the
             following: arrow, clobber, cross, diamond, disc, hbar, ring,
             square, star, tailed_arrow, triangle_down, triangle_up, vbar, x.
-        size : float, array
+        size : float, tuple
```
kevinyamauchi: The type is listed as tuple here, but the example (nD_points.py) uses a list.

napari/components/viewer_model.py
```diff
@@ -556,10 +557,14 @@ def add_points(
...
             Size of the point marker. If given as a scalar, all points are made
-            the same size. If given as an array, size must be the same
-            broadcastable to the same shape as the data.
+            the same size. If given as a tuple, size must be the same length
+            as the number of points.
+        anisotropy : tuple, optional
+            List of anisotropy factors, must be one per dimension. These act as
+            pre-multipliers on the size of each point. If not provided,
```
kevinyamauchi: I think it's a bit confusing how the anisotropy factor really only applies to the dimensions out of the viewer plane even though they are defined for all planes. Perhaps we should make a note about that here.

```suggestion
            pre-multipliers on the size of each point. Note that the anisotropy is only displayed for dimensions out of plane for the view. If not provided,
```

napari/layers/points/points.py
```diff
@@ -335,11 +340,30 @@ def size(self, size: Union[None, float]) -> None:
...
             for i in self.selected_data:
-                self.sizes[i, :] = (self.sizes[i, :] > 0) * size
+                self.sizes[i] = size
+                self._actual_sizes[i] = np.multiply(size, self.anisotropy)
             self._set_view_slice()
         self.status = format_float(self.size)
         self.events.size()
 
+    @property
+    def anisotropy(self) -> Union[tuple]:
```
kevinyamauchi: I don't think we need `Union` here if it only returns a `tuple`

napari/layers/points/points.py
```diff
@@ -335,11 +340,30 @@ def size(self, size: Union[None, float]) -> None:
...
+        """tuple: anisotropy factor for each dimension."""
+
+        return self._anisotropy
+
+    @anisotropy.setter
+    def anisotropy(self, anisotropy: Union[None, tuple]) -> None:
+        if anisotropy is None:
+            anisotropy = np.ones(self.dims.ndim)
```
kevinyamauchi, line 362: The type hint is for `None` or `tuple`, but `anisotropy` is an `ndarray` here.

kevinyamauchi (commented):
I think this looks good. However, I do think the expected types for `anisotropy`should be clarified. Please let me know if my comments are confusing.

napari/components/viewer_model.py
sofroniewn: yup, fixing - in general we havn't been so great about this stuff, probably need to do a sweep of the repo to catch them all - but no point making it worse!

napari/components/viewer_model.py
sofroniewn: agreed, added

napari/layers/points/points.py
sofroniewn: should be `np.ndarray`

napari/layers/points/points.py
sofroniewn: added `np.ndarray`

sofroniewn:
Thanks for catching this stuff, I've never been so good with the typing stuff. As noted above we'll have to go through the repo some time and type everything, but for now there's no point in making it worse by having the wrong types!!!

examples/nD_points.py
```diff
@@ -32,5 +32,5 @@
         ]
     )
     viewer.add_points(
-        points, size=[0, 6, 10, 10], face_color='blue', n_dimensional=True
+        points, anisotropy=[0, 0.5, 1, 1], size=10, face_color='blue', n_dimensional=True
```
jni, line 35: (a) I would put size before anisotropy, because it is the more intuitive and universal argument, in that most people will deal with size but not anisotropy; and
(b) I strongly suggest renaming anisotropy to scale, to make it consistent with the image layer. Anisotropy is not a name that most people will find intuitive!

napari/components/viewer_model.py
```diff
@@ -556,10 +557,15 @@ def add_points(
             Symbol to be used for the point markers. Must be one of the
             following: arrow, clobber, cross, diamond, disc, hbar, ring,
             square, star, tailed_arrow, triangle_down, triangle_up, vbar, x.
-        size : float, array
+        size : float, tuple, list, array (N,)
```
jni, line 560: I call all the sequence types "array-like".

napari/components/viewer_model.py
```diff
@@ -556,10 +557,15 @@ def add_points(
...
             Size of the point marker. If given as a scalar, all points are made
-            the same size. If given as an array, size must be the same
-            broadcastable to the same shape as the data.
+            the same size. If given as a tuple, list, or 1-D array size must be
+            the same length as the number of points.
+        anisotropy : tuple, list, array (D,), optional
+            List of anisotropy factors, must be one per dimension. These act as
+            pre-multipliers on the size of each point. Note that the anisotropy
+            is only displayed for dimensions out of plane for the view. If not
```
jni, line 567: displayed -> used? The anisotropy is not really "displayed".

A clarification here about *how* it's used would be handy. Higher anisotropy means that the point will "leak" into more planes? Or fewer?

napari/layers/points/points.py
```diff
@@ -335,11 +342,32 @@ def size(self, size: Union[None, float]) -> None:
...
 
+    @property
+    def anisotropy(self) -> np.ndarray:
+        """tuple: anisotropy factor for each dimension."""
+
+        return self._anisotropy
+
+    @anisotropy.setter
+    def anisotropy(
+        self, anisotropy: Union[None, list, tuple, np.ndarray]
```
jni, line 359: This is nice! I would refactor as follows though:

```python
from typing import Optional as Opt
ArrayLike = Union[list, tuple, np.ndarray]
... anisotropy: Opt[ArrayLike]
```

Optional marks something as "or None".

napari/layers/points/points.py
```diff
@@ -517,7 +543,7 @@ def _slice_data(self, indices):
         if len(self.data) > 0:
             if self.n_dimensional is True and self.ndim > 2:
                 distances = abs(self.data[:, not_disp] - indices[not_disp])
-                sizes = self.sizes[:, not_disp] / 2
+                sizes = self._actual_sizes[:, not_disp] / 2
```
jni, line 546: I'd change "actual sizes" to "displayed sizes", maybe? "scaled sizes"?

jni (approved):
@sofroniewn as per usual, approved for your convenience, but I would strongly recommend renaming the argument before merging. anisotropy is not a good kwarg name to add on a major release!

sofroniewn:
Hmm ok @jni I share your concern that `anisotropy` is a bad keyword argument as not as many people will know it.

Also - `scale` is already used as a keyword argument, it's used as a keyword argument for every layer, but it will be applied to the entire layer, not just the radii, i.e. if you have a point at `[50, 50]` with radii `[10, 10]` and apply scale `[2, 2]` it will move your point to `[100, 100]` and the radii will now be `[20, 20]`. This is desirable for some things, but different then rescaling the size of the just the radii.

Alternative proposal (as @jni posted on the issue) - how about we just keep the size keyword argument and force it to be broadcastable into data.shape (N, D)? So a scalar ->(N, D), vector of shape (D,) -> (N, D), vector of shape (N,) -> crashes, (N, 1) -> (N, D), (1, D) -> (N, D), and (N, D) -> (N, D).

The one thing I don't like about this is I feel that a lot of the time people will just have a length N list of sizes and a length D list of scale factors, but then maybe they just combine them ahead of time. This option is also strictly more general as it allows for different points to have different anisotropy.

napari/layers/points/points.py
```diff
@@ -309,20 +315,21 @@ def symbol(self, symbol: Union[str, Symbol]) -> None:
...
-        """(N, D) array: sizes of all N points in D dimensions."""
+        """(N,) array: sizes of all N points."""
         return self._sizes
 
     @sizes.setter
     def sizes(self, size: Union[int, float, np.ndarray, list]) -> None:
-        try:
-            self._sizes = np.broadcast_to(size, self.data.shape).copy()
-        except:
-            try:
```
sofroniewn, line 320: Note @jni if we go with the strict broadcasting approach I will close this PR and make a new one that only drops this extra try and changes the tests / examples appropriately. Does this sound good to you?

jni:
>  i.e. if you have a point at [50, 50] with radii [10, 10] and apply scale [2, 2] it will move your point to [100, 100] and the radii will now be [20, 20].

fwiw I'm not positive this is the correct behaviour... If I give you coordinates at [50, 50], I can mean two things:

- these are pixel coordinates, and therefore don't move them
- these are data coordinates, in which case, with scale [2, 2], you should move them to [25, 25] in pixel space, not [100, 100]?

But, the problem remains: how can we adjust radii and coordinates independently?

I kinda sorta like your broadcasting idea, but I agree with you that it's not very user-friendly. Additionally, maybe this should be the time where we separate data from display. imho, points should not have a "size", as they are points. =) However, they can have a "value". Then, the UI can display that value as a size property or as a colormapped color, for example. So, a suggestion for a deeper refactor:

- points have two attributes, coordinates and values. Coordinates have a fixed dimensionality, but values can be (N, F), where F is the "number of features" and can be arbitrary
- in the UI or programmatically, you can map those values to attributes such as size, color, or shape (for categorical values).
- we support providing feature names either by making values a pandas dataframe or by a separate "feature_names" kwarg that takes in a list of strings.

Thoughts?

sofroniewn:
I'll have to sleep on this one. Maybe we'll table this till after the `0.2.0` release. Color mapping with a value would be nice too though ...

napari/layers/points/points.py
jni: You could also force-push to this branch, which contains a lot of useful discussion. But, either way, I don't think the workflow is the primary concern.

sofroniewn:
@kevinyamauchi i wonder if we should revisit this and close #480 in a new PR before the `0.3.0` release? Now that we have the `properties` dictionary there are more options available to us for mapping values to size

goanpeca:
Is this PR still something to work on, or can it be closed @sofroniewn?

Thanks!

sofroniewn:
I think we can probably close this PR as we don't need to use this approach, but we still probably want to address #480 as that is a real issue. I think @jni might have better ideas for how to do this now too (thinking about what we decided to do for the labels painting anistropicaly too). I'm not sure though

PR #553, opened 2019-09-30
fix menubar focus mac

### LINKED ISSUE #380: Menu bar focus on Mac
## 🐛 Bug
We've now added a menubar, but you need to toggle focus in and out of napari before it becomes active on the mac. This bug has been encountered in other Qt apps, but we still need to look into fixing.

See here - https://github.com/robotology/yarp/issues/457

sofroniewn:
# Description
This closes #380 where I needed to leave the focus of napari to another application and then come back to napari before the menubar worked.

I'm not quite sure why the `raise_` call was in there or if there are any negative issues of dropping it on other platforms, but I can confirm that on a mac everything still functions as expected, including usage from the notebook.

Maybe some other people can verify performance on different platforms, but this bug fix seems pretty desirable.

## Type of change

### DIFF
--- a/napari/_qt/qt_main_window.py
+++ b/napari/_qt/qt_main_window.py
@@ -146,7 +146,6 @@ class Window:
         """
         self._qt_window.resize(self._qt_window.layout().sizeHint())
         self._qt_window.show()
-        self._qt_window.raise_()
 
     def _update_palette(self, palette):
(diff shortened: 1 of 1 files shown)


jni (approved):
Works perfectly for me on both macOS and Linux. Merging! :tada:

Issue #554, opened 2019-09-30
auto pyramid creation sometimes is not wanted


sofroniewn:
## 🐛 Bug

The auto pyramid creation that was enabled in #545 can sometimes lead to bad unexpected behavior - for example I just tried to pass a `100_000 x 100 x 100` zarr array which I will only ever want to view as a 2D timeseries and it started trying to compute a pyrmaid on that first dimension which it couldn't do.

I'm not sure how to both block this behavior, but then also allow a pyramid to be calculated if someone passes in a `20_000 x 20_000` image.

We could wait until we know what dimensions they actually want to view and so push this over to more of the view code, but I'm not so sure about that either.

## To Reproduce
run `examples/zarr_nD_image.py`

PR #555, opened 2019-09-30
move zarr, xarray, dask from examples to tests

### LINKED ISSUE #179: Examples should only have dependencies that are in the requirements

sofroniewn:
# Description
This PR closes #179 by moving our examples of zarr, xarray, and dask to our tests. I think that as we don't explicitly depend on these packages its more confusing to have them in our examples as they won't run when a new-commer installs napari. We now though do test for integration.

There are two problems that I noted, one is documented in #554 around autocreation of pyramids sometimes being undesirable. Another is that right now if using a zarr array without wrapping in dask you must sent `contrast_limits` explicitly otherwise things will crash.

## Type of change

# How has this been tested?

- [x] add `test_numpy_like.py`

## Final checklist:


napari/tests/test_numpy_like.py
```diff
@@ -0,0 +1,131 @@
+import numpy as np
```
jni, line 1: Should this file not include a `pytest.importorskip`?

napari/util/misc.py
```diff
@@ -246,12 +246,17 @@ def calc_data_range(data):
...
-    min = data.min()
-    max = data.max()
-    if min == max:
-        min = 0
-        max = 1
-    return [float(min), float(max)]
+    if np.prod(data.shape) > 1e6:
+        reduced_data = np.random.choice(data.flatten(), size=1e6)
+    else:
+        reduced_data = data
```
jni, line 260: Not in love with this change... Accessing 1M random points from a dask array could be extremely slow depending on backing (e.g. remote data). I still prefer dtype, but if you don't want that, I suggest taking e.g. the middle plane.

jni (commented):
@sofroniewn why move instead of copy? I think they are valuable as examples.

I also had some deeper concerns about the new additions... What do you think?

napari/util/misc.py
sofroniewn: yup, it's a concern for me too, if each of those is in a different chunk .... I just wasnt quite sure what to do. How about top, middle, and bottom planes? Seems like an ok compromise for now!!

sofroniewn:
The move instead of copy was because the examples won't run unless you have those packages installed. Maybe we need to add a try and catch with an error message saying "Dask not installed, in order to run dask example you must install dask with pip install dask" or something like that?

I can work on making these two changes and merge if you approve :-)

jni (approved):
Yep, I like all your proposed solutions, go for it! =)

napari/tests/test_numpy_like.py
sofroniewn: These packages are listed in our test dependencies so I don't think we need the `importorskip`, or am I mistaken?

sofroniewn:
I've also made an improvement to the multichannel api here (maybe should been it's own PR, sorry) that makes the keyword arg `axis` instead of `channel` and removes some duplicate code

i will merge this once the tests pass

PR #560, opened 2019-10-01
WIP Unify IO between different modalities

jni:
# Description

I didn't get a chance to finish this, but thought I'd submit in case you have time to improve it or would still want to merge it as is. I've tested locally with tiff, png, (command-line, drag and drop, and file>open) and zarr files (command line).

One issue I'm having is that zarr files are actually directories. This means that:

(a) ~~they don't work with drag and drop, because the default behavior of drag and drop is to list directory contents~~ **Edit:** this was an easy fix; and
(b) they don't work with File > Open, because the behavior of that dialog is to drill down into directories rather than let you select one as input.

I know there's solutions to (b) in macOS, but I'm not sure they can be expanded to Linux and Windows.

## Type of change

# References

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/util/io.py
+++ b/napari/util/io.py
@@ -1,10 +1,17 @@
+import os
+
 import numpy as np
 from skimage import io
-from glob import glob
+
+dask_available = True
+try:
+    from dask import array as da
+except ImportError:
+    dask_available = False
 
 
-def read(filenames):
-    """ Read image files and return an array.
+def imread(filenames, *, use_dask=None, stack=True):
+    """Read image files and return an array.
 
     If multiple images are selected, they are stacked along the 0th axis.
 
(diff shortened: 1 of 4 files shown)


sofroniewn:
@jni - does `QFileDialog.getExistingDirectory` instead of `QFileDialog.getOpenFileNames` work for (b) to open the zarr folders? I think it might, will give it a check now too

jni:
> does QFileDialog.getExistingDirectory instead of QFileDialog.getOpenFileNames work for (b) to open the zarr folders?

yes, but then normal files are not selectable or even visible...

napari/util/io.py
```diff
@@ -1,27 +1,88 @@
...
     Returns
     -------
     image : array
         Array of images
     """
+    if dask_available and use_dask is None:
+        use_dask = True
+    if not dask_available and use_dask:
+        raise ValueError('Dask array requested but dask is not installed.')
     images = [io.imread(filename) for filename in filenames]
```
sofroniewn, line 39: Isn't this just reading the images directly in a non-lazy way? In which case I don't understand what value dask is providing. For the use dask case don't we want to use something like dask-image

I'd like to be able to do the following on the command line
```
napari my_dir_of_images/*.png -D
```
and have that use `data = dask_image.imread(my_dir_of_images/*.png)` which I know works from a script and is super cool because those pngs don't need to be all the same size

sofroniewn (approved):
I've tested and can confirm the basics work as expected, though I have some requests for the dask file io. I am going to merge as it though for the `0.2.0` release and we can then work on a big file io push for `0.2.1`

Issue #476, opened 2019-08-13
Add benchmarks


constantinpape:
## 🚀 Feature

Add benchmarks to monitor the performance of napari.

## Motivation

Keep track of performance to catch regressions and judge performance improvements for planned features like lazy rendering, better tiling or caching.
Moving the discussion from #318 here.

## Pitch

Benchmarks should be easy to run, maybe be part of the CLI and should cover
relevant use-cases like 2d/3d image rendering, pyramids, lazy sources and volume rendering.
Could also consider using [asv](https://github.com/airspeed-velocity/asv) for monitoring the performance.

## Additional context

cc @sofroniewn @bryantChhun

sofroniewn:
Thanks for creating this issue @constantinpape. In the long run I'd like us to have full coverage for all our layer types for the benchmarks - shapes / vectors / points can all run into problems with 1000s or millions of objects depending on the details and we will want to understand those limits.

I also see a couple different things we'll want to measure.

- Each layer has a `.refresh` method that gets called whenever anything has changed that will effect what is displayed on the screen. This can trigger the thumbnail generation machinery which we want to make sure isn't slowing things down.

- There's also machinery for the dimensions - so going from looking at one z-plane to another

- There's just a camera zoom in / zoom out. This is done with vispy's machinery but probably good for us to monitor too. (For the pyramid layer it is more complex, so also good to have this in here).

- There's a `get_value` function that is used for cursor hovering - this is fast for image-like layers as you just have to grab the value of the image at the coordinate where we are hovering, but it can be slow for the shapes layer where you have to check is this coordinate inside any of the shapes.

- Changing the `layer.data` property, this was explored in #465 as it was very slow, and ultimately a work around was found by directly setting the vispy data property.

@jni @royerloic are there other high priority things we want to benchmark?

As to how to do this, I like something that integrates with our CI, but I'm very open to approaches / implementations. Maybe @kne42 has some ideas? I do think vispy does some benchmarking, so getting inspired from them makes sense, but I havn't looked into their approach

jni:
> I like something that integrates with our CI, but I'm very open to approaches / implementations.

[...]

PR #564, opened 2019-10-02
faster points thumbnails

### LINKED ISSUE #562: Plot lots of spots
## 🚀 Plot lots of spots
I would like to be able to plot millions of spots using napari.

## Motivation

Starfish users will often want to plot spots _after_ merging fields of view. In these cases, there will be millions of spots to plot. 

## Pitch

Right now I can plot 20k spots, which takes a while. _After_ the spots populate napari, navigating them is lightning fast. Larger groups of spots take longer than I'm willing to wait (a few minutes for 250k before I interrupt the kernel).

## Alternatives

I tried Vaex, and mayavi, and matplotlib, and could only get mayavi to work, but it ground my computer to a halt, and I had to play some tricks to kill it.

sofroniewn:
# Description
This PR closes #562 by speeding up our points plotting. It gives a max number of points to be randomly sampled for thumbnail generation, similar to how there was a maximum number of vectors in the vectors layer.

This speeds up plotting a large number of points considerably, but things are still too slow for ideal usage. Plotting ~100_000 points takes ~20 seconds on my macbook pro, plotting ~1_000_000 points takes ~3 minutes 25 seconds (roughly 10x as long). Not sure what @ambrosejcarr you were seeing before. I havn't done any profiling yet to see where the bottle necks are, but this still feels much slower than what I think vispy can handle.

Once vispy has these points it's pretty fast to pan / zoom, though changing color still appears slow - see here for interacting with 1M points:

1_M_points

I think in general it's time for us to consider performance benchmarking a little more seriously and I'd like us to set up some infrastructure for doing that more seriously before we start to chase down all our potential performance gains. Some of this is discussed in #476 

## Type of change

# How has this been tested?

## Final checklist:


napari/layers/points/points.py
```diff
@@ -112,6 +112,10 @@ class Points(Layer):
         None after dragging is done.
     """
 
+    # The max number of points that will ever be used to render the thumbnail
+    # If more vectors are present then they are randomly subsampled
```
```suggestion
    # If more are present then they are randomly subsampled
```

jni (approved):
LGTM, just a very minor comment.

Regarding your points about benchmarking, yes, very true. How's that provisioning of a benchmark machine at CZI going? =)

jni:
@sofroniewn I wonder if the cursor movement callbacks are a bottleneck now?

ambrosejcarr:
@sofroniewn since you're considering benchmarking, it might be worth looping in @ttung who has been thinking about what it would take to do this for starfish.

ambrosejcarr:
For my current use case, I can probably sub-sample down to ~ 100k points for demonstration purposes, thanks for the quick work @sofroniewn! If it guides additional exploration, eventually I'd hope to find something that could plot <10m points in a few seconds.

sofroniewn:
Would love to chat with @ttung about benchmarks - will follow up with him. As discussed in #476 going with the practices established in https://scikit-image.org/docs/dev/contribute.html#benchmarks and using https://asv.readthedocs.io/en/stable/ seems like a good place for us to start.

@jni no progress on getting a dedicated machine, but that shouldn't hold us back at this point. Lots of things to do to get going.

I don't think that it's the cursor movement callbacks holding us back as the latency is on loading and once the 1M points are in then its all very fast, but the cursor movement callbacks are still getting called then. I'll dig in a bit more before merging this, if I find anything easy to fix I'll add it, otherwise we can follow up later. This PR should at least make things a little better.

napari/layers/points/points.py
sofroniewn: fixed

sofroniewn:
I've merged the benchmarking PR in here - the changes in this PR wrt to the thumbnail generation along cause the following benchmark changes

Before:

```
[ 75.00%] ··· benchmark_points_layer.Points2DSuite.time_create_layer                                                                                                                                      ok
[ 75.00%] ··· ======== ============
               param1
              -------- ------------
                 16     4.27±0.4ms
                 64     9.24±0.2ms
                256      29.1±1ms
                1024     108±3ms
                4096     423±20ms
               16384    1.83±0.2s
               65536    6.92±0.03s
              ======== ============
```

After:
```
[ 25.00%] ··· benchmark_points_layer.Points2DSuite.time_create_layer                                                                                                                                      ok
[ 25.00%] ··· ======== ==========
               param1
              -------- ----------
                 16     5.13±0ms
                 64     9.92±0ms
                256     29.9±0ms
                1024    100±0ms
                4096    124±0ms
               16384    123±0ms
               65536    131±0ms
              ======== ==========
```

Which is a big improvement for basically no functionality loss. We can probably make the thumbnail stuff even better with the kdtree, but we can wait on that for a future PR.

Note the benchmarks are only are the points layer model right now, so don't include the vispy stuff

jni:
@sofroniewn super super cool! It's so gratifying to see these numbers rather than just "it feels much faster now". :clap: :clap: :clap:

Issue #561, opened 2019-10-01
multicanvas grid display for layers in Napari


kellman:
## 🚀 Feature
It would awesome if Napari could have a viewing option that would allow the users to view a stack of layers side by side in a grid or tile display. The users could then jointly zoom on each layer in the stack to compare high resolution features over a large field of view.

## Motivation

One of the most important ways to evaluate an algorithm or an image reconstruction's performance is to visualize its output side by side with many other methods (e.g. competing methods, base line method, naive approaches). The ability to jointly zoom on images will allow for quick comparison of high resolution details over a large field of view. 

## Alternatives

While it is possible in matplotlib to share axes on subplots for joint zoom, this solution is only accessible by python experts with extensive implementation and documentation knowledge and not general users.

sofroniewn:
Absolutely @kellman. We've got a note about this in our features for `0.3.0` issue #420, but now's a good time to make a dedicated issue for this.

One thing we can do very easily is just tile the layers but only have one camera. What that means is that when you zoom out you see all the tiles, but when you zoom in your just zooming in on one place and eventually you only see that layer. I'm not sure though that doing it this way is really that valuable though, but I'm curious from your perspective if you'd ever want functionality like this.

What I think you really want is 

> users could then jointly zoom on each layer

which we'll be able to support this but it requires us having multiple canvases with linked cameras under the hood. Maybe we just do this part and skip the simple tiling I mentioned above entirely.

Thoughts @royerloic?

royerloic:
I have had multiple people request this feature, I think it would extremely usefull for people that process images and want to compare them side-by-side. Essentially everybody doing machine learning on images (image restoration etc...). Having the possibility to cycle through different layouts ( stacked, horiz, grid) in a single key-binding would be super practical... 

Michael suggestion of being able to zoom in simultaneously would require deeper engineering, and takes us to the whole multi-view topic that should be addressed in 0.3 and beyond (simultaneous orthogonal displays, synced views, 3D in one view and slice in another, etc...)

I think starting with a simple grid toggling as suggested by @sofroniewn would already be great and address something that a lot of the ML community working on images would want to use.

sofroniewn:
@royerloic if we had the version of linked zooms, would you ever use the version without it? i.e. is that just a temporary feature because it might be easier, or is it something intrinsically valuable?

[...]

Issue #122, opened 2019-02-22
Inconsistency in ordering of layers visually and in the actual list object


kne42:
## ❓ Questions and Help

So I noticed something today that confused me a lot regarding how things in the `LayersList` are ordered.

It's easiest to provide an example – say your layers list is ordered like so: `[astronaut, camera, spots]`. On screen, the list is ordered like so:
```
spots
camera
astronaut
```
and the `spots` layer is displayed at the top of the stack.

I find this confusing since I would expect `astronaut` to appear at the top of the list and `spots` at the bottom.

So I wanted to get everyone's opinion on what ordering would make the most sense:
- should the ordering match?
- should new layers get appended or inserted to the list?
- should the topmost or the bottommost layer be displayed at the top of the stack?

sofroniewn:
This was extensively discussed in #565 and we eventually decided it is more intuitive to have the most visible, and most recently added list on top in the viewer, even though it is the reverse order from the code. Hopefully people will be indexing into the layers list using the layer name rather than the index so it won't be too confusing

PR #565, opened 2019-10-02
basic grid vs stack views

sofroniewn:
# Description
This PR follows up on #561 by implementing a basic grid vs stack view. Right now there are 3 methods that have gotten added to the viewer - `subplot(layer, position, size)` that will place a layer at a particular position in a grid of a particular size, and then two convenience methods `grid_view` and `stack_view` that will place layers in those respective configurations. `grid_view` also takes some parameters in case you want to set the grid to be a specific size or move through the layers with a particular stride, so that you could for example place layer 0, 1 on top of eachother in one grid cell, and layer 2, 3 on top of eachother in the next grid cell etc.

This is still WIP, I need to test it works in all configs and add some tests, but I wanted to quickly follow up on the discussion we were having in #561 with @jni @royerloic and @kellman

Here's a gif of the current performance:

grid_stack_views

It might be nice to add a button to the toggle grid vs stack view. Also should rearranging the order of the layers in the layers list rearrange the position of the layers in the grid once the grid has been made? If so in what way? Should there be away to rearrange the position of the layers in the grid just by clicking and dragging on them?

## Type of change

# How has this been tested?

## Final checklist:

### DIFF
--- a/examples/layers.py
+++ b/examples/layers.py
@@ -5,6 +5,7 @@ using the layers swap method and remove one
 
 from skimage import data
 from skimage.color import rgb2gray
+import numpy as np
 import napari
 
 
@@ -14,9 +15,5 @@ with napari.gui_qt():
     viewer.add_image(data.camera(), name='photographer')
     viewer.add_image(data.coins(), name='coins')
     viewer.add_image(data.moon(), name='moon')
-
-    # remove the coins layer
-    viewer.layers.remove('coins')
-
-    # swap the order of astronaut and moon
-    viewer.layers['astronaut', 'moon'] = viewer.layers['moon', 'astronaut']
+    viewer.add_image(np.random.random((512, 512)), name='random')
+    viewer.add_image(data.binary_blobs(length=512, volume_fraction=0.2, n_dim=2), name='blobs')
--- a/examples/layers.py
+++ b/examples/layers.py
@@ -17,3 +17,4 @@ with napari.gui_qt():
     viewer.add_image(data.moon(), name='moon')
     viewer.add_image(np.random.random((512, 512)), name='random')
     viewer.add_image(data.binary_blobs(length=512, volume_fraction=0.2, n_dim=2), name='blobs')
+    viewer.grid_view()
--- a/napari/_qt/qt_viewer_buttons.py
+++ b/napari/_qt/qt_viewer_buttons.py
@@ -1,5 +1,6 @@
 from qtpy.QtWidgets import QHBoxLayout, QPushButton, QFrame, QCheckBox
 from qtpy.QtCore import Qt
+import numpy as np
 
 
 class QtLayerButtons(QFrame):
@@ -31,6 +32,7 @@ class QtViewerButtons(QFrame):
         self.ndisplayButton = QtNDisplayButton(self.viewer)
         self.rollDimsButton = QtRollDimsButton(self.viewer)
         self.transposeDimsButton = QtTransposeDimsButton(self.viewer)
+        self.gridViewButton = QtGridViewButton(self.viewer)
         self.resetViewButton = QtResetViewButton(self.viewer)
 
         layout = QHBoxLayout()
@@ -39,6 +41,7 @@ class QtViewerButtons(QFrame):
         layout.addWidget(self.ndisplayButton)
         layout.addWidget(self.rollDimsButton)
         layout.addWidget(self.transposeDimsButton)
+        layout.addWidget(self.gridViewButton)
         layout.addWidget(self.resetViewButton)
         layout.addStretch(0)
         self.setLayout(layout)
@@ -134,6 +137,27 @@ class QtConsoleButton(QPushButton):
         self.viewer = viewer
         self.setToolTip('Open IPython terminal')
         self.setProperty('expanded', False)
+
+
+class QtGridViewButton(QCheckBox):
+    def __init__(self, viewer):
+        super().__init__()
+
+        self.viewer = viewer
+        self.setToolTip('Toggle grid view view')
+        self.viewer.events.grid.connect(self._on_grid_change)
+        self.stateChanged.connect(lambda state=self: self.change_grid(state))
+        self._on_grid_change(None)
+
+    def change_grid(self, state):
+        if state == Qt.Checked:
+            self.viewer.stack_view()
+        else:
+            self.viewer.grid_view()
+
+    def _on_grid_change(self, event):
+        with self.viewer.events.grid.blocker():
+            self.setChecked(np.all(self.viewer.grid_size == (1, 1)))
 
 
 class QtNDisplayButton(QCheckBox):
--- a/napari/_vispy/vispy_base_layer.py
+++ b/napari/_vispy/vispy_base_layer.py
@@ -130,7 +130,8 @@ class VispyBaseLayer(ABC):
 
     def _on_translate_change(self):
         self.translate = [
-            self.layer.translate[d] for d in self.layer.dims.displayed[::-1]
+            self.layer.translate[d] + self.layer.translate_grid[d]
+            for d in self.layer.dims.displayed[::-1]
         ]
         self.layer.position = self._transform_position(self._position)
 
--- a/napari/_vispy/vispy_image_layer.py
+++ b/napari/_vispy/vispy_image_layer.py
@@ -108,7 +108,9 @@ class VispyImageLayer(VispyBaseLayer):
 
     def _on_translate_change(self):
         self.translate = [
-            self.layer.translate[d] + self.layer._translate_view[d]
+            self.layer.translate[d]
+            + self.layer._translate_view[d]
+            + self.layer.translate_grid[d]
             for d in self.layer.dims.displayed[::-1]
         ]
         self.layer.position = self._transform_position(self._position)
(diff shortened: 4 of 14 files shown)


sofroniewn:
I'll also note that this implementation will work with all our different layer types

royerloic:
Perfect, this is going to be super useful. Probably going to use it everyday...  As is anybody that needs to process images and compare them...

sofroniewn:
This PR is now ready for review @jni @royerloic 

It is implemented as we discussed including a new button to toggle between grid and stacked mode. See here for its basic usage:

grid_mode

You can actually use the grid mode with nD data too, and with 3D rendering if you choose, see here:

nD_grid_mode

Finally if you want to use the stride parameter or set a custom grid size you have to use

```python
viewer.grid_view(n_row, n_column, stride)
viewer.stack_view()
```

sofroniewn:
Seems like our CI is failing to even install windows right now

jni:
@sofroniewn you are blowing my mind :heart_eyes:

sofroniewn:
ah - and one other thing that I'm curious for @jni @royerloic opinion on is right now the layers list is ordered such that the most recently added layer goes on top and the oldest layer goes on the bottom. This is different than the natural ordering of the list, and Kira raised this issue #122 to discuss this a while back. At the time I liked the ordering with the new layers ontop, reasoning that those were the ones you were most recently adding were the ones you wanted to see the most. Now actually the layers list would automatically scroll if we grew it from the bottom, so you would see the most recently added at the bottom anyway.

I think it might make sense with grid mode to have the list ordered in the other way now with oldest layer at the top, and grow the layer going down. The reason for that is that the oldest layer gets added to the (0, 0) spot in the grid, and it's weird when the (0, 0) spot and the layer at the top of the grid don't match.

Thoughts?

royerloic:
Wow, this is so cool. I love how the 3D grid just works...

royerloic:
> ah - and one other thing that I'm curious for @jni @royerloic opinion on is right now the layers list is ordered such that the most recently added layer goes on top and the oldest layer goes on the bottom. This is different than the natural ordering of the list, and Kira raised this issue #122 to discuss this a while back. At the time I liked the ordering with the new layers ontop, reasoning that those were the ones you were most recently adding were the ones you wanted to see the most. Now actually the layers list would automatically scroll if we grew it from the bottom, so you would see the most recently added at the bottom anyway. 
> I think it might make sense with grid mode to have the list ordered in the other way now with oldest layer at the top, and grow the layer going down. The reason for that is that the oldest layer gets added to the (0, 0) spot in the grid, and it's weird when the (0, 0) spot and the layer at the top of the grid don't match.

Here we can think about the model/view separation to inform us in the best way.
I think that the order of the list should not change, that's the model, how we choose 
to display it is a different thing.  This could cause all sorts of trouble
where scripts would expect the last computed  image at the top, etc... think about 'stack' 
based computation for example. I am perfectly happy with having the top (first) layer
be always at (0,0). The grid would be reorganised. Maybe that's confusing. 

Oh wait, you might be suggesting simply to display the other way around, not to
reorder the list itself... Yes. if that's what you mean, then having the top of the list 
displayed at the end of the grid. It might be confusing though... If you just display
two layers, it is not obvious which layer corresponds to what... One possibility is to
display a 'border' (highlight) that is synced with the layer list selection. I am not sure
that we want to go this path now, though.

Not obvious what is best. We need to think about it more.

jni:
I don't really get the confusion... If I remember correctly, we *append* to the layer list (?), and the list grows towards the top? This is actually fine and if you think about a 2D images + a right-handed reference frame, this is the direction you expect to grow in when you have (row, col) coordinates.

Now, if you take the linear list and reshape it using `np.reshape`, then you end up with the list growing on the bottom-right edge. So that remains consistent as far as I can tell...

So, my vote:

- layer list grows by appending
- latest layer on top in the layer list
- latest layer on the bottom-right in the grid view

I do see that most users would read left-right then top-bottom, but I don't think that is our primary concern...

sofroniewn:
Yeah @royerloic i’m not thinking of changing the model, which would be very complex, just the view.

Should new layers appear at the bottom or top of the view. Right now it is the top, maybe it should be the bottom? That’s all I am asking

royerloic:
Top seems reasonable to me, new layers get added at the top of the list,  and at the left-top corner of the grid. I am afraid we  have a tie with @jni . @sofroniewn feel free to pick your preference if Juan agrees :-)

sofroniewn:
Hmm ok - I think I will consult @liaprins-czi for a 3rd opinion on monday. I think we're all in agreement that the layers list gets appended too though as otherwise all the indexing is weird.

As to growing the layers list view at the top or bottom, and growing the grid view at the top left / bottom right. I think I actually lean towards grow the list view at the bottom and grow the grid view at the bottom right.

For growing the list view at the bottom it is then much easier to count down from the top and know that the top layer is the 0th element in the list, the next layer is the 1st etc, whereas now you need to count up from the bottom. Purely from a code standpoint now I have to do all sorts of weird reversals to maintain the correspondence between the layer list and its view, which is fine as they happen under the hood, but maybe an indication something isn't natural either.

The rest of this PR is ready for review

royerloic:
ok, makes sense, everything grows at the bottom (right). In that case it is all very intuitive and consistent. It was late yesterday... ;-)

sofroniewn:
What does @jni think? I'll add with this screenshot, in the current view it just looks confusing that the two lists run in opposite orders:

layers_order

jni:
I'm pretty happy to think of the layerlist view as a "stack", because the images do appear like they are stacked together and the topmost image is hiding all the other ones. Then, think about putting your palm on the topmost image and sliding it out, splaying all the "pictures" underneath across the "desk" towards the bottom right. (In fact this could be a neat animation to add. ;) It is that one metaphor that breaks down if you invert the order of the layerlist. But otherwise, your arguments make sense, and I definitely prefer inverting the list to inverting the order of the montage.

I'm very curious what @liaprins-czi will say. =)

examples/layers.py
```diff
@@ -14,9 +15,5 @@
...
     viewer.add_image(data.coins(), name='coins')
     viewer.add_image(data.moon(), name='moon')
-
-    # remove the coins layer
-    viewer.layers.remove('coins')
-
-    # swap the order of astronaut and moon
-    viewer.layers['astronaut', 'moon'] = viewer.layers['moon', 'astronaut']
+    viewer.add_image(np.random.random((512, 512)), name='random')
+    viewer.add_image(data.binary_blobs(length=512, volume_fraction=0.2, n_dim=2), name='blobs')
```
jni, line 19: Do you want to set the layout mode to grid here?

jni (approved):
@sofroniewn the implementation is wonderfully simple! And the current layout matches my preference, so feel free to merge immediately! ;)

sofroniewn:
During discussion with @liaprins-czi today, her preference was for the layers in the layers list view to remain as they are with most visible on top, and therefore most recent on top, as you want the most recently added layer also to be most visible. Given that, she felt it was natural to have the most recent layer in the top left corner (which is different then our current implementation where it is bottom right). She felt it was important to have that consistency between the top left corner of the grid view and the top of the layers list.

The above implementation still leaves us with a mismatch between the order of the layers list view and the order of the layers in the actual python list, but she felt this was ok - it was more important to have consistency within the GUI then between how things are done within the GUI and within the code, which I agree with, it isn't that big a deal that the ordering is different, and we should really be using the layer name to index into the layers list rather than the location in the list.

Finally she thought it was strange and confusing that the layers could be edited in grid mode, and that you could enter 3D rendering in grid mode, and that you could create new layers in grid mode. All of these options actually led us to weird confusing non-intuitive behavior and she proposed to disable as much of this functionality (at least from the GUI) when in grid mode so as users don't click the wrong stuff and end up doing something weird. I'll think about the best way to implement this, but broadly agree.

If this sounds good to @jni and @royerloic I'll probably make these changes tomorrow

jni:
@sofroniewn 

> She felt it was important to have that consistency between the top left corner of the grid view and the top of the layers list.

*sigh* fine. =P Actually I'd very much like it if we could make this a configurable option somewhere, so I could play around with both modes. I have a strong intuition that having the next added image on the bottom right would work better, at least for me, someone experienced in the NumPy ecosystem. For biologists who haven't necessarily worked with arrays, I'll admit it might be a different story...

> she thought it was strange and confusing that the layers could be edited in grid mode, and that you could enter 3D rendering in grid mode, and that you could create new layers in grid mode.

Here I'm actually in favour of keeping the current behaviour. One common use case that would be enabled basically as soon as this PR is merged is to pick correspondences between two images you want to co-register. This would actually be quite easy to do with the two images, two interleaved points layers, and a stride of 2.

And both @royerloic and I were very impressed with the side-by-side volume rendering, and I think it would be a very nice way to see certain datasets.

sofroniewn:
@jni ok your points are well taken - i wont disable anything now, we can always disable for the `0.2.1` release, but merging the PR now will give people a change to play around with it.

I've made the default order in the grid such that the most visible layer goes in the top left corner, but if you pass a negative stride (say `stride=-1`) then the order will be reversed and it will go in the bottom right corner. This seems like a pretty reasonable compromise to me.

I've also added tests. I will merge once they all pass.

examples/layers.py
sofroniewn: done

Issue #558, opened 2019-10-01
Napari on Windows


haesleinhuepf:
## 📚 Documentation
Hey @sofroniewn ,

I ❤️ your announcement on the image.sc forum and thus had to try Napari on my Windows. So the installation works fine apparently, but the example script from the readme doesn't:

![image](https://user-images.githubusercontent.com/12660498/65944442-fa935200-e431-11e9-8358-f60728fda7bf.png)

![image](https://user-images.githubusercontent.com/12660498/65944455-03842380-e432-11e9-8b9b-d7a56b9dafbf.png)

```
AttributeError: module 'napari' has no attribute 'view_image'
```

Are maybe just the examples outdated?

Thanks in advnace!

Cheers,
Robet

jni:
@haesleinhuepf :wave: great to see you here! =) You should install the version from GitHub or wait ~12h before pip installing with --update. =) @sofroniewn is about to release v0.2 which includes the new functions.

haesleinhuepf:
Waiting 12h sounds like an amaaaazing solution 😃 

Thanks guys! Looking forward to explore the second island on the other side of the planet ;-)

jni:
> Looking forward to explore the second island on the other side of the planet ;-)

Nice pickup. :wink:

sofroniewn:
@haesleinhuepf this should now be fixed - `0.2.0` has been release on pypi, ~12 hrs later!!

great to here that your windows worked ok, i think we might be having some issues with openGL on older systems, but we havn't figured it all out yet

sofroniewn:
oh actually now that i think about it, you havn't seen an image yet - so the windows problem might still come up :-(
If you get a black screen after adding the image and the error message described here https://github.com/napari/napari/pull/551#issuecomment-536347310 then we'll need to fix it

haesleinhuepf:
Unfortunately, it still comlpains about a missing method: 
![image](https://user-images.githubusercontent.com/12660498/66024501-54f4e700-e4f4-11e9-83ee-6dcf31097941.png)

```
AttributeError: module 'napari' has no attribute 'view_image'
```

To be honest, to me this doesn't look like a windows problem....

Any hints would be helpful!

Thanks!

Cheers,
Robert

jni:
@haesleinhuepf good morning! =D

[...]

PR #580, opened 2019-10-10
WIP: About Menubar

AhmetCanSolak:
# Description

This PR attempts to add about menubar to napari.
Need to update PR with version and other desired information, please give feedback on things you want to see in about menubar.

## Type of change

# References

# How has this been tested?

Tested on mac osx, and not adding about menubar to mainmenu, any tips are welcome.
edit: now properly working on mac osx.

## Final checklist:


jni:
Thanks for quickly tackling this, @AhmetCanSolak! Two comments:

1) The menu layout should be "Help (top level menu) - About napari (menu item)"
2) I would change the description to "napari: a fast n-dimensional image viewer"

For now, that and the version numbers for:
- napari
- Qt
- NumPy
- SciPy
- VisPy
- scikit-image

should be sufficient, imho. We can always expand with more info (and a logo) in the future.

jni:
I also would not add a keyboard shortcut for this.

napari/_qt/qt_main_window.py
```diff
@@ -129,6 +132,14 @@ def _add_window_menu(self):
         self.window_menu = self.main_menu.addMenu('&Window')
         self.window_menu.addAction(exit_action)
 
+    def _add_about_menu(self):
+        about_action = QAction('About', self._qt_window)
```
AhmetCanSolak: Having 'About' as an action name without having `self.main_menu.setNativeMenuBar(False)` prevents adding help menu on mac osx. For now will be putting something else.

AhmetCanSolak:
Hi @jni !

I addressed your comments and made it work properly on mac osx too. Ready for review.

napari/_qt/qt_main_window.py
```diff
@@ -129,6 +132,13 @@ def _add_window_menu(self):
         self.window_menu = self.main_menu.addMenu('&Window')
         self.window_menu.addAction(exit_action)
 
+    def _add_help_menu(self):
+        about_action = QAction("napari info", self._qt_window)
```
sofroniewn, line 136: can this be "about napari" instead of "napari info"

napari/_qt/qt_about.py
```diff
@@ -0,0 +1,48 @@
...
+        )
+
+        # Horizontal Line Break
+        self.hline_break1 = QFrame()
+        self.hline_break1.setFrameShape(QFrame.HLine)
+        self.hline_break1.setFrameShadow(QFrame.Sunken)
+        self.layout.addWidget(self.hline_break1)
+
+        # Versions
+        self.layout.addWidget(QLabel("napari, " + napari.__version__))
```
sofroniewn: Is there any way to make this more selectable text then just the qlabels. i.e. when it pops open I'd like to be able to drag and select all the text so i could easily paste it into a github issue, now I need to take a screenshot

napari/_qt/qt_about.py
```diff
@@ -0,0 +1,48 @@
...
+        self.layout.addWidget(QLabel("Qt, " + qtpy.__version__))
+        self.layout.addWidget(QLabel("NumPy, " + numpy.__version__))
+        self.layout.addWidget(QLabel("SciPy, " + scipy.__version__))
+        self.layout.addWidget(QLabel("VisPy, " + vispy.__version__))
+        self.layout.addWidget(QLabel("scikit-image, " + skimage.__version__))
```
sofroniewn: Can we include some of the results of calling  `vispy.sys_info()` - we've been asking our windows users to send us this info when they are having problems - see here https://github.com/napari/napari/issues/558#issuecomment-537569931
and it might be easier to just ask them to open this dialog box, particularly as we start reaching more users that aren't familiar with the command line

The results from there I want are 
```
Python:   3.7.3 (default, Apr 24 2019, 15:29:51) [MSC v.1915 64 bit (AMD64)]
Backend:  PySide2
GL version:  '4.6.0 - Build 25.20.100.7011'
MAX_TEXTURE_SIZE: 16384
```

I think also the `Qt` right now is reporting the qtpy version not the actual Qt version which could be confusing.
```python
from qtpy import QtCore
QtCore.__version__
```
will give the actual Qt version

napari/_qt/qt_main_window.py
AhmetCanSolak: that was my initial intention. On mac osx with native menubar 'about' word in action name prevents creation of help menu. That's why i changed to napari info, if you can find a way to make it work i will be happy to change.

napari/_qt/qt_about.py
AhmetCanSolak: this is easy, im changing as you requested

napari/_qt/qt_about.py
AhmetCanSolak: i am adding `vispy.sys_info()` but line length gets really long on some lines

sofroniewn: can you just pull out the those 5 fields from above?

AhmetCanSolak: i did :)

sofroniewn (approved):
Looks good to me, thanks for getting this done with the new info. If you can also edit the Bug report Issue template with instructions to copy and paste both sections into the bottom of the issue that would be great

We'll want to update with the logo at some point and maybe a link to the tutorials etc, but we can add that stuff later

sofroniewn:
I'm merging now as I want to use the about menu in another PR I'm working on now for keybindings, we can make follow-ups later

AhmetCanSolak:
Opened a follow up for issue template here #591

Issue #518, opened 2019-09-07
Report and test all keybindings


sofroniewn:
## 🚀 Feature
Right now we've been adding keybindings to the window menubar, the viewer, all our layers.

We need a way to document and expose all these keybindings to the user, and test that all of them work and remain stable over time.

I expect that our plugin developers will want to leverage these keybindings, or provide their own to overwrite or complement them, and we will need to know what keybindings we have and make sure they remain stable for this to be possible.

sofroniewn:
This is closed by #583

PR #583, opened 2019-10-13
Test and display keybindings

sofroniewn:
# Description
This PR follows on from #580 by adding the option to display all our keybindings in the pop-up from the help menu. It also adds more comprehensive tests for the keybindings themselves, including some tests similar to those in #544.

It also adjusts the stylings of both the `napari info` and `keybindings` pop-ups to use our style sheets.

The actual strings that are displayed for our keybindings clearly still need some work - particularly those associated with functions that don't have good doc strings or use a lambda

Finally I'm not quite so sure about the dialog box completely blocking the threads on the main viewer - say for keybindings it might be nice to have the keybindings box open and then be working with the viewer at the same time.

Here's what it looks like (note the actual menubar is cutoff from my gif!) - 
keybindings_display

Curious to get feedback from @AhmetCanSolak, @jni, and @royerloic at this stage though 

## Type of change

# How has this been tested?

- [x] added a custom keybindings test and tests of all our keybindings in our viewer / layers.

## Final checklist:

### DIFF
--- a/napari/_qt/qt_about.py
+++ b/napari/_qt/qt_about.py
@@ -10,9 +10,9 @@ from qtpy.QtWidgets import QWidget, QVBoxLayout, QLabel, QDialog, QFrame
 import napari
 
 
-class AboutPage(QWidget):
+class QtAbout(QWidget):
     def __init__(self, parent):
-        super(AboutPage, self).__init__(parent)
+        super(QtAbout, self).__init__(parent)
 
         self.layout = QVBoxLayout()
 
@@ -72,11 +72,11 @@ class AboutPage(QWidget):
         self.setLayout(self.layout)
 
     @staticmethod
-    def showAbout():
+    def showAbout(qt_viewer):
         d = QDialog()
-        d.setGeometry(150, 150, 350, 400)
-        d.setFixedSize(600, 700)
-        AboutPage(d)
-        d.setWindowTitle("About")
+        d.setObjectName('QtAbout')
+        d.setStyleSheet(qt_viewer.styleSheet())
+        QtAbout(d)
+        d.setWindowTitle('About')
         d.setWindowModality(Qt.ApplicationModal)
         d.exec_()
--- a/napari/_qt/qt_about_keybindings.py
+++ b/napari/_qt/qt_about_keybindings.py
@@ -1,0 +1,82 @@
+from qtpy import QtCore
+from qtpy.QtCore import Qt
+from qtpy.QtWidgets import (
+    QTabWidget,
+    QWidget,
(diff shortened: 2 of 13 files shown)


sofroniewn:
Note this will close #518

jni (approved):
Gorgeous! I love the nice use of numpydoc. =D

jni:
Ah, just looked at the screenshot LOL. I reckon the lambdas vindicate my earlier advocacy for proper functions taking in an event=None even if they don't use the event info. =)

sofroniewn:
@jni - many of those lambdas are actually about how to get the `Viewer` in as a keyword argument - for example 
```python
Viewer.bind_key('Control-E', lambda v: v.dims._roll())
```
not sure how to make this better.

We've definitely still got more to do here with our about menus / doc strings etc. but do you want to merge this PR now or are there more changes I should make before it goes in?

I'd probably say lets just merge and then we can keep going

jni:
@sofroniewn 

```python
def roll_axes(viewer):
    """Change the order of the visible axes, e.g. [0, 1, 2] -> [2, 0, 1]."""
    viewer.dims._roll()


Viewer.bind_key('Control-E', roll_axes)
```

?

sofroniewn:
that looks nicer! I can change all our lambdas over to that / or use the event trick so that we get nice doc strings for all of them. i'll do that before merging

royerloic:
A Popup window...
But, ok, as long as this window is not modal (blocking the main window)
that should be fine...  ;-))
Actually super useful!

jni:
@royerloic pop-ups are the standard way to do help menus. In the future, we can use vispy text to overlay the current shortcuts on the image when pressing '?'. Pressing any key would make the overlay disappear and activate the shortcut if the key was a shortcut. This is how Gmail keyboard shortcuts work and it's super handy.

sofroniewn:
Right now the pop is blocking unfortunately @royerloic - we can work on fixing that in the future, will have to ensure that it remains up to date even when aspects of the viewer change for the keybindings window

royerloic:
A floating -- non-blocking -- window that is visible and on the side, available to consult while working in napari, is not a bad thing I think... @sofroniewn indeed, syncing with potentially changing shortcuts could be done if we have some basic notification to that floating window that something change would be important in that case. I am happy with this as long as this is just a first prototype.... :-)

sofroniewn:
@royerloic - i've made the window non-blocking, syncing so it updates when needed, and resizable - (though the resizing isn't quite perfect, it's pretty close)

I've also improved all the doc strings - I think it's pretty cool right now

I will merge after tests pass

PR #584, opened 2019-10-13
viewer args to view_* methods

### LINKED ISSUE #581: Incorrect interpretation of dimensions for viewing multichannel data
## 🐛 Bug

When loading data with multiple channels, napari selects the first dimension as the channels dimension even when it represents rows, not channels

## To Reproduce

```
randoms = np.random.randint(0, 20, 1024*1024*5)
channels_first = randoms.reshape((5, 1024, 1024))
channels_last = randoms.reshape((1024, 1024, 5))

# example code for trying out napari functionality
with napari.gui_qt():
    viewer = napari.view_image(channels_last, rgb=False, name='channels_last')

```
## Expected behavior

Instead of displaying the data as 1024 distinct slices of size 5x1024, it should be displayed as 5 stacks fo 1024x1024. I imagine there is going to be a huge range of different shapes and data formats supplied to napari from different users, and setting up the exact logic to guess correctly is infeasible. Perhaps supplying a channels idx as an argument?

## Environment

 - napari Version (e.g., 1.0): 0.2

## Additional context

image

image

sofroniewn:
# Description
This PR closes #581 by making it possible to pass all keyword arguments that go to `napari.Viewer` to all of our `napari.view_*` methods.

For example, it is now possible to do
```python
import numpy as np
import napari


randoms = np.random.randint(0, 20, 1024*1024*5)
channels_first = randoms.reshape((5, 1024, 1024))
channels_last = randoms.reshape((1024, 1024, 5))

# example code for trying out napari functionality
with napari.gui_qt():
    viewer = napari.view_image(channels_last, rgb=False, name='channels_last', order=[2, 0, 1])
```

and observe the desired 1024x1024 slice immediately on display, rather than having to make a second function call.

## Type of change

## Final checklist:

### DIFF
--- a/napari/view_layers.py
+++ b/napari/view_layers.py
@@ -20,6 +20,9 @@ def view_image(
     opacity=1,
     blending='translucent',
     visible=True,
+    title='napari',
+    ndisplay=2,
+    order=None,
 ):
     """Create a viewer and add an image layer.
 
@@ -70,13 +73,21 @@ def view_image(
         {'opaque', 'translucent', and 'additive'}.
     visible : bool
         Whether the layer visual is currently being displayed.
+    title : string
+        The title of the viewer window.
+    ndisplay : int
+        Number of displayed dimensions.
+    tuple of int
+        Order in which dimensions are displayed where the last two or last
+        three dimensions correspond to row x column or plane x row x column if
+        ndisplay is 2 or 3.
 
     Returns
     -------
     viewer : :class:`napari.Viewer`
         The newly-created viewer.
     """
-    viewer = Viewer()
+    viewer = Viewer(title=title, ndisplay=ndisplay, order=order)
     viewer.add_image(
         data,
         rgb=rgb,
@@ -111,6 +122,9 @@ def view_multichannel(
     opacity=1,
     blending='additive',
     visible=True,
+    title='napari',
+    ndisplay=2,
+    order=None,
 ):
     """Create a viewer and add images layers expanding along one axis.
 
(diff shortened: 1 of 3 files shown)


napari/view_layers.py
```diff
@@ -153,13 +167,21 @@ def view_multichannel(
         {'opaque', 'translucent', and 'additive'}.
     visible : bool
         Whether the layer visual is currently being displayed.
+    title : string
+        The title of the viewer window.
+    ndisplay : int
```
```suggestion
    ndisplay : {2, 3}
```

napari/view_layers.py
```diff
@@ -235,6 +260,14 @@ def view_points(
...
```
```suggestion
    ndisplay : {2, 3}
```

napari/view_layers.py
```diff
@@ -318,13 +354,21 @@ def view_labels(
...
```
```suggestion
    ndisplay : {2, 3}
```

napari/view_layers.py
```diff
@@ -411,13 +458,21 @@ def view_shapes(
...
```
```suggestion
    ndisplay : {2, 3}
```

napari/view_layers.py
```diff
@@ -485,13 +543,21 @@ def view_surface(
...
```
```suggestion
    ndisplay : {2, 3}
```

napari/view_layers.py
```diff
@@ -553,13 +622,21 @@ def view_vectors(
         {'opaque', 'translucent', and 'additive'}.
     visible : bool
         Whether the layer visual is currently being displayed.
-
+    title : string
+        The title of the viewer window.
+    ndisplay : int
```
```suggestion
    ndisplay : {2, 3}
```

jni (approved):
@sofroniewn it's funny, when I first read your comment proposing this, I was like, ugh, that's gonna make the signature too complicated/confusing, not sure I like it. But seeing the implementation, it's actually very clear, and I agree that it's super useful. It's also super nice to be able to immediately get a 3D viewer with `napari.view_image(image, ndisplay=3)`. 

I've left some minor comments about the docstring, but otherwise LGTM.

napari/view_layers.py
sofroniewn: fixed

napari/view_layers.py
sofroniewn: fixed

napari/view_layers.py
sofroniewn: fixed

napari/view_layers.py
sofroniewn: fixed

napari/view_layers.py
sofroniewn: fixed

napari/view_layers.py
sofroniewn: fixed

sofroniewn:
Fixed all the doc strings as requested, this can be merged once the tests pass

PR #585, opened 2019-10-13
fix pyramid guessing

### LINKED ISSUE #563: 2D Image Pyramids not working with add_image (0.2.0)
## 🐛 Bug

Napari is not displaying an image pyramid that I've constructed and passed to `add_image()`

## To Reproduce

```python
import numpy as np
from skimage.transform import pyramid_gaussian

import napari

img = np.random.randint(0, 255, (1000, 1000))
pyramid_test = np.asarray(tuple(pyramid_gaussian(img)))
viewer = napari.Viewer()
viewer.add_image(pyramid_test, is_pyramid=True)
```
What I did to trigger the error: 
```python
>>> regions
<xarray.Regions 'array-baa123768d7ef2a6d2cdfa885e8f3eaa' (x_region: 31725, y_region: 51669)>
array([[0, 0, 0, ..., 0, 0, 0],
       [0, 0, 0, ..., 0, 0, 0],
       [0, 0, 0, ..., 0, 0, 0],
       ...,
       [0, 0, 0, ..., 0, 0, 0],
       [0, 0, 0, ..., 0, 0, 0],
       [0, 0, 0, ..., 0, 0, 0]], dtype=int16)
Dimensions without coordinates: x_region, y_region
Attributes:
    assay:             osmFISH
    authors:           ['Simone Codeluppi', 'Lars E. Borm', 'Amit Zeisel', 'G...
    organism:          mouse
    publication_name:  Spatial organization of the somatosensory cortex revea...
    publication_url:   https://www.nature.com/articles/s41592-018-0175-z
    sample_type:       somatosensory cortex
    year:              2018

>>> pyramid = tuple(pyramid_gaussian(regions), multichannel=False)

>>> [i.shape for i in pyramid]
[(31725, 51669),
 (15863, 25835),
 (7932, 12918),
 (3966, 6459),
 (1983, 3230),
 (992, 1615),
 (496, 808),
 (248, 404),
 (124, 202),
 (62, 101),
 (31, 51),
 (16, 26),
 (8, 13),
 (4, 7),
 (2, 4),
 (1, 2),
 (1, 1)]

>>> viewer = napari.Viewer()

>>> viewer.add_image(np.asarray(pyramid), is_pyramid=True)
ValueError                                Traceback (most recent call last)
<ipython-input-35-665dfea0cb05> in <module>
----> 1 viewer.add_image(np.asarray(pyramid), is_pyramid=True)

/usr/local/lib/python3.7/site-packages/napari/components/viewer_model.py in add_image(self, data, rgb, is_pyramid, colormap, contrast_limits, interpolation, rendering, name, metadata, scale, translate, opacity, blending, visible)
    426             opacity=opacity,
    427             blending=blending,
--> 428             visible=visible,
    429         )
    430         self.add_layer(layer)

/usr/local/lib/python3.7/site-packages/napari/layers/image/image.py in __init__(self, data, rgb, is_pyramid, colormap, contrast_limits, interpolation, rendering, name, metadata, scale, translate, opacity, blending, visible)
    185             else:
    186                 input_data = self.data
--> 187             self._contrast_limits_range = calc_data_range(input_data)
    188         else:
    189             self._contrast_limits_range = contrast_limits

/usr/local/lib/python3.7/site-packages/napari/util/misc.py in calc_data_range(data)
    268         reduced_data = data
    269 
--> 270     min_val = reduced_data.min()
    271     max_val = reduced_data.max()
    272 

/usr/local/lib/python3.7/site-packages/numpy/core/_methods.py in _amin(a, axis, out, keepdims, initial)
     30 def _amin(a, axis=None, out=None, keepdims=False,
     31           initial=_NoValue):
---> 32     return umr_minimum(a, axis, None, out, keepdims, initial)
     33 
     34 def _sum(a, axis=None, dtype=None, out=None, keepdims=False,

ValueError: operands could not be broadcast together with shapes (31725,51669) (15863,25835) 
```
## Environment

 - napari Version: 0.2.0
 - OS: OS X 10.14
 - Python version: 3.7

sofroniewn:
# Description
This PR closes #563 as desired by @jni - it reduces our guessing, raising errors when there are conflicts instead of coercing a user setting. It also adds more tests to catch the edge cases in #563

## Type of change

# How has this been tested?

- [x] added tests to `utils/tests/test_misc.py` and `layers/image/tests/test_pyramid.py`

## Final checklist:

### DIFF
--- a/napari/layers/image/tests/test_pyramid.py
+++ b/napari/layers/image/tests/test_pyramid.py
@@ -14,6 +14,7 @@ def test_random_pyramid():
     layer = Image(data, is_pyramid=True)
     assert layer.data == data
     assert layer.is_pyramid == True
+    assert len(layer._data_pyramid) > 0
     assert layer.ndim == len(shapes[0])
     assert layer.shape == shapes[0]
     assert layer.rgb == False
@@ -28,6 +29,7 @@ def test_infer_pyramid():
     layer = Image(data)
     assert layer.data == data
     assert layer.is_pyramid == True
+    assert len(layer._data_pyramid) > 0
     assert layer.ndim == len(shapes[0])
     assert layer.shape == shapes[0]
     assert layer.rgb == False
@@ -65,6 +67,7 @@ def test_forcing_pyramid():
     layer = Image(data, is_pyramid=True)
     assert np.all(layer.data == data)
     assert layer.is_pyramid == True
+    assert len(layer._data_pyramid) > 0
     assert layer.ndim == len(shape)
     assert layer.shape == shape
     assert layer.rgb == False
@@ -79,6 +82,7 @@ def test_blocking_pyramid():
     layer = Image(data, is_pyramid=False)
     assert np.all(layer.data == data)
     assert layer.is_pyramid == False
+    assert layer._data_pyramid == None
     assert layer.ndim == len(shape)
     assert layer.shape == shape
     assert layer.rgb == False
--- a/napari/layers/image/tests/test_pyramid.py
+++ b/napari/layers/image/tests/test_pyramid.py
@@ -3,6 +3,7 @@ from skimage.transform import pyramid_gaussian
 from xml.etree.ElementTree import Element
 from vispy.color import Colormap
 from napari.layers import Image
+import pytest
 
 
 def test_random_pyramid():
@@ -31,6 +32,15 @@ def test_infer_pyramid():
     assert layer.shape == shapes[0]
     assert layer.rgb == False
     assert layer._data_view.ndim == 2
+
+
+def test_error_pyramid():
+    """Test error on forcing non pyramid."""
+    shapes = [(40, 20), (20, 10), (10, 5)]
+    np.random.seed(0)
+    data = [np.random.random(s) for s in shapes]
+    with pytest.raises(ValueError):
+        layer = Image(data, is_pyramid=False)
 
 
 def test_infer_tuple_pyramid():
--- a/napari/layers/image/tests/test_pyramid.py
+++ b/napari/layers/image/tests/test_pyramid.py
@@ -1,4 +1,5 @@
 import numpy as np
+from skimage.transform import pyramid_gaussian
 from xml.etree.ElementTree import Element
 from vispy.color import Colormap
 from napari.layers import Image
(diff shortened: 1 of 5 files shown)


napari/layers/image/tests/test_pyramid.py
```diff
@@ -32,6 +33,63 @@ def test_infer_pyramid():
...
+    assert layer.data == data
+    assert layer.is_pyramid == True
+    assert layer.ndim == len(shapes[0])
+    assert layer.shape == shapes[0]
+    assert layer.rgb == False
+    assert layer._data_view.ndim == 2
+
+
+def test_forcing_pyramid():
+    """Test instantiating Image layer forcing 2D pyramid data."""
```
jni, line 63: This particular test is confusing to me... `is_pyramid=True` has no effect here? The resulting layer tests are the same for this test as for the one following it...

napari/util/misc.py
```diff
@@ -60,13 +60,10 @@ def is_rgb(shape):
...
-    """If data is a list of arrays of decreasing size.
+    """If shape of arrays along first axis is strictly decreasing.
     """
-    if isinstance(data, list):
-        size = [np.prod(d.shape) for d in data]
-        return np.all(size[:-1] > size[1:])
-    else:
-        return False
+    size = [np.prod(d.shape) for d in data]
+    return np.all(size[:-1] > size[1:])
```
jni: This doesn't do what you think it does:

```python
In [1]: size = [500, 1, 500]                                                                        

In [2]: size[:-1] > size[1:]                                                                        
Out[2]: True
```

The simplest fix is probably to cast `size` to `np.array` in the preceding line.

jni: maybe worth adding a test that fails with a list of sizes like the one above.

napari/util/misc.py
```diff
@@ -144,7 +142,14 @@ def get_pyramid_and_rgb(data, pyramid=None, rgb=None):
     else:
         # If rgb is True or None then guess if rgb
         # allowed or not, and if allowed set it to be True
-        rgb = is_rgb(init_shape)
+        rgb_guess = is_rgb(init_shape)
+        if rgb and rgb_guess is False:
+            raise ValueError(
+                """Non rgb or rgba data was passed, but rgb data was
+                requested."""
```
jni: This shouldn't be a triple-quoted string, (which contains the newline), but two single-quoted strings (which will get automatically concatenated). Don't forget to put a space at the end of the first string or the start of the second string. =)

jni: Also, could you add a test for this value error? You can test for raising with:

```python
    layer = Image(np.random.random((512, 512, 512)), rgb=True)
```

napari/util/misc.py
```diff
@@ -164,8 +169,12 @@ def get_pyramid_and_rgb(data, pyramid=None, rgb=None):
             data_pyramid = trim_pyramid(data)
             pyramid = True
         else:
-            # Guess if data should be pyramid
-            pyr_axes = should_be_pyramid(data.shape)
+            # Guess if data should be pyramid or if a pyramid was requested
+            if pyramid:
+                pyr_axes = [1] * ndim
```
jni:
```suggestion
                pyr_axes = [True] * ndim
```

jni (approved):
Looks good! There's a couple of required fixes but they should be easy to implement.

napari/util/misc.py
sofroniewn: fixing doc string, and test added.

there was actually a test for this that's actually now causing our tests to fail. Using this instead.

napari/util/misc.py

napari/util/misc.py
sofroniewn: added the `np.array`, and the special handling of length 1 lists so this now works. There was already a test that caught it after the addition.

sofroniewn:
Thanks @jni, I've made the fixes - will merge once the tests pass

napari/layers/image/tests/test_pyramid.py
sofroniewn: the `assert layer.is_pyramid == True` is different, but I've also added some extra asserts about the `layer._data_pyramid` property that distinguish the True vs False state, the other properties were the same

PR #482, opened 2019-08-15
fix OSX CI

kne42:
# Description
fix OSX CI by pinning `ipykernel` to `5.1.1`. for some reason `5.1.2` was causing errors with `pluggy` and the `asyncio` event loop

PR #587, opened 2019-10-14
[WIP] Don't pin ipykernel

jni:
# Description

Thanks to updates in pip, napari would no longer run in an environment where ipykernel 5.1.2 was installed. This addresses that issue.

## Type of change

# References

# How has this been tested?

## Final checklist:


sofroniewn:
The pin was needed because I think we broke on `5.1.2` - potentially only on mac osx. I'll look into it, but unless they have fixed things the console won't work

jni:
@sofroniewn (a) can you check, (b) can you check whether pip can do platform-specific pins? The console seems to be working fine on Linux, and (c) even then, you can do `ipykernel>=5.1.1,!=5.1.2` which allows upgrades through, should the issue be fixed. Indeed the CI seems to back (a), though. Have you raised an issue with ipykernel?

sofroniewn:
I will look into it - here is the PR #482 and discussion where we did the pinning. I'm not sure an issue has been raised, or if it just just a mac OSX CI issue. I imagine so as I probably had it working on my laptop before we pinned. The issue does still seem to be present though judging from our failing tests

sofroniewn (approved):
I've now dropped support for multiple kernels with multiple viewers, which was buggy anyway (you can still have multiple viewers, the later ones just won't get a console) and should allow our tests to pass on osx - but without pinning ipykernel. We shall see, if the tests pass then I will merge.

I saw at least one person at the course i was teaching have install problems because of the pinning!!

sofroniewn:
Our tests are still failing - now likely due to our kernels not actually shutting down in the way I expect. I have made an issue here - https://github.com/jupyter/qtconsole/issues/385 to discuss

jni:
@sofroniewn new release 5.1.3 of ipykernel, on first glance some of those updates might be relevant:

https://github.com/ipython/ipykernel/blob/master/docs/changelog.rst

sofroniewn:
I will had tried the latest master of ipykernel recently (after some of those updates) and it still didn't fix it, but I will try with the new release too

kne42:
5.1.2 and 5.1.3 are still causing too many open files errors for me. if we want to stop pinning it, we could set it as an inclusive range between 5.0.0 and 5.1.1

sofroniewn:
@kne42 the pinning has been causing install problems for some users. I'm not sure that range will fix them. I think we only run into the problem during our tests and it seems a shame to cause install problems just for our tests. I'm not quite sure what's best

kne42:
@sofroniewn ah, right, I completely forgot that this only causes errors in our tests! In that case, we should unpin `ipykernel` from the default requirements and add it to the dev requirements and our 
CI instead with a comment explaining why.

sofroniewn:
That’s a great idea @kne42! - do you want to create a fresh PR doing that and then I will close this one.

sofroniewn:
Even better would be to only change the version for our osx tests as then at least Linux will still be tested against latest. What does @jni think

kne42:
@sofroniewn sure! I recently reran the tests on this PR and they are failing even with `ipykernel==5.1.3` on every platform so I'm guessing it's more widespread than MacOS but maybe @jni can verify as an actual Linux user

kne42:
Since @jni has not responded yet, I will go ahead and create a new PR.

sofroniewn:
@kne42 as we just discussed on the video chat, I think the work @shanaxel42 is doing in #769 might fix this bug in our tests and then we won't need to worry about pinning anywhere

sofroniewn:
This has now been superseded by #784 and can be closed

PR #588, opened 2019-10-14
open file or folder from a QFileDialog

sofroniewn:
# Description
Allow you to open a file or folder from a `QFileDialog` as requested by @jni. Note that if you open a folder now things will crash as we havn't connected the code to open the folder


jni:
Thanks very much for the PR, @sofroniewn! I've rolled these changes into #590.

PR #589, opened 2019-10-14
bracketed numbers for autoincrementing

### LINKED ISSUE #586: novel gene discovery in Napari  (a.k.a. imperfect layer name incrementing)
## 🐛 Bug
napari does a smart thing and adds an integer to the name of a layer if the name is already in use in the viewer. It then increments that number when another layer is added that fits the pattern.  This is great until you have layer names that end in numbers 
Screenshot from 2019-10-13 13-28-55
which in this case results in different names that are sometimes real genes and sometimes not.

## To Reproduce

```
v = napari.Viewer()
for i in range(3):
    v.add_points(1000*np.random.rand(10,2), name = "no_problem")
for i in range(3):
    v.add_points(1000*np.random.rand(10,2), name = "problem:GAD1")
```

## Expected behavior
consistent use of a space or other character to separate the auto-increment integer from the original string

sofroniewn:
# Description
This PR closes #586 by switching to bracketed numbers for our auto incrementing - i.e. `layer [1]`, `layer [2]`. See screenshot:

autoincrement

It should match the behavior discussed in the issue, and there are tests for various edge cases

## Type of change

### DIFF
--- a/napari/components/tests/test_layers_list.py
+++ b/napari/components/tests/test_layers_list.py
@@ -111,13 +111,13 @@ def test_naming():
     layers.append(layer_a)
     layers.append(layer_b)
 
-    assert [l.name for l in layers] == ['img', 'img 1']
+    assert [l.name for l in layers] == ['img', 'img [1]']
 
     layer_b.name = 'chg'
     assert [l.name for l in layers] == ['img', 'chg']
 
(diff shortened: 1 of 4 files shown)


jni (approved):
Love all the tests!

PR #590, opened 2019-10-14
Further unify IO and use dask for multiple files

jni:
# Description

This PR fixes dask usage behaviour when reading in folders of files. It also further unifies the behaviour of the file > open dialog, the command line file IO, and the drag and drop functionality.

This PR does not:
- allow lazy loading of single files e.g. multi-page tiffs.
- work correctly with mixed dtypes or mixed shapes if the user doesn't use `use_dask=False` (because we need to assume that all dtypes are the same, or read in all files in sequence to check the dtype).
- allow the user to set `use_dask` or `stack` options with `File > open` or drag and drop.

I haven't had time to test this but it could definitely have a first round of preliminary review to check the design.

## Type of change

- [ ] Bug-fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

PR #592, opened 2019-10-15
0.2.1 release

sofroniewn:
# Description
These are release notes for `0.2.1` including still to be merged #590. My plan is to make the release after that gets merged, but can @jni you give this an approval already so that we can do that asynchronously if we need to as I'd like to get the new release out today / tomorrow

### DIFF
--- a/docs/release/release_0_2_1.rst
+++ b/docs/release/release_0_2_1.rst
@@ -11,55 +11,41 @@ rendering), and the scientific Python stack (numpy, scipy).
 For more information, examples, and documentation, please visit our website:
 https://github.com/napari/napari
 
+
 New Features
 ------------
-- added a code of conduct
-- added grid mode that allows for looking at layers side by side
-- added the ability to bind custom mouse functions
-- added a benchmarking suite
-- added help menu with about dialog and keybindings dialog
+- added a code of conduct (#374)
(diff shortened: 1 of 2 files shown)


docs/release/release_0_2_1.rst
```diff
@@ -0,0 +1,77 @@
...
+images. It's built on top of Qt (for the GUI), vispy (for performant GPU-base
+rendering), and the scientific Python stack (numpy, scipy).
+
+
+For more information, examples, and documentation, please visit our website:
+https://github.com/napari/napari
+
+New Features
+------------
+- added a code of conduct
```
jni: We typically put PR numbers in the scikit-image release notes... Are you not using the script for these?

jni: Oh, I see that they are below. Typically I move rather than copy PRs from the list to the sections. Not a big deal though.

jni (approved):
@sofroniewn it would be great to have #594 in this release also!

royerloic:
Great! Time to update!

docs/release/release_0_2_1.rst
sofroniewn: that's a good idea, have changed

PR #594, opened 2019-10-15
Allow add_points() to be called without an argument to create empty points layer

jni:
# Description
Similarly for add_shapes(). This allows easier annotation of volumes when working in an interactive IPython session.

## Type of change

# References

# How has this been tested?

- [x] example: the test suite for my feature covers all new code paths

## Final checklist:

### DIFF
--- a/napari/components/viewer_model.py
+++ b/napari/components/viewer_model.py
@@ -648,7 +648,7 @@ class ViewerModel(KeymapMixin):
             data = np.empty([0, ndim])
 
         layer = layers.Points(
-            data,
+            data=data,
             symbol=symbol,
             size=size,
             edge_width=edge_width,
@@ -823,7 +823,7 @@ class ViewerModel(KeymapMixin):
             data = np.empty((0, 0, ndim))
 
         layer = layers.Shapes(
-            data,
+            data=data,
             shape_type=shape_type,
             edge_width=edge_width,
             edge_color=edge_color,
--- a/napari/components/viewer_model.py
+++ b/napari/components/viewer_model.py
@@ -980,12 +980,6 @@ class ViewerModel(KeymapMixin):
         self.add_layer(layer)
         return layer
 
-    def _new_points(self):
-        self.add_points()
-
-    def _new_shapes(self):
-        self.add_shapes()
-
     def _new_labels(self):
         if self.dims.ndim == 0:
             dims = (512, 512)
--- a/napari/components/viewer_model.py
+++ b/napari/components/viewer_model.py
@@ -577,7 +577,7 @@ class ViewerModel(KeymapMixin):
 
     def add_points(
         self,
-        data,
+        data=None,
         *,
         symbol='o',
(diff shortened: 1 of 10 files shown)


jni:
@sofroniewn it would be awesome to get this added to 0.2.1 also...!

sofroniewn (approved):
Can you change the function signatures for `view_points` and `view_shapes` too?

How about doing the same for `add_labels` and `view_labels`?

napari/components/viewer_model.py
```diff
@@ -973,18 +981,10 @@ def add_vectors(
         return layer
 
     def _new_points(self):
```
sofroniewn: can we just drop the `_new_points` method entirely and connect the button to `self.add_points()` instead - one less function to worry about. similar for `_new_shapes` too

jni: Yes but I worried about existing tests and having to replicate/modify them. It's almost 1am here. ;) I'd prefer to leave that for a later PR.

jni:
Hmm looks like I have some fixing to do already anyway...!

jni:
Re labels, it was a bit more complicated logic, which I thought I might want to tweak, so I decided to leave that for later... I can add it if you feel strongly about it.

Re `view_`, it seems to me a bit weird to launch an empty viewer + an empty points layer... What do you think?

jni:
Ah, the class signature no longer matches the layer signature... HMMM. I guess I can add the `=None` to the layer, but it won't be quite the same, since there will be no way to know what the dimensionality of an empty layer should be. It'll have to be 2 by necessity.

jni:
Hmmm, ok, I'm going to bed, as I don't know whether I can properly think about out whether it's prudent to remove Kira's very rigorous test :joy: If you figure it out, feel free to push to this branch and merge! I'll keep at it tomorrow. Goodnight!

jni:
@sofroniewn how's the jet lag? ;) Anyway, this should work now! I'm pretty busy for the next couple of days (Lou Scheffer is actually in Melb! =), so I suggest merging this and leaving your other suggestions for a later release.

jni:
... Aaaaand conda is broken. Anyway, tests pass locally...

jni:
This is the cause of the conda failure:

https://github.com/conda/conda/pull/9014
https://github.com/conda/conda/issues/8149

Will attempt to solve with something similar to this:

https://github.com/umccr/woofr/commit/0b6b3633f095f8000302cc3eb60e98380a1dc33c

although maybe I can just delete our caches.

jni:
Woot! :tada:

sofroniewn:
I'd rather you can do `view_points()` and `view_shapes()` and have it add an empty layer too - I see no reason for the api to diverge there, and if people want to do it they then can.

I'll quickly make this change and maybe drop `_new_shapes` and `_new_points` is easy and then merge

sofroniewn:
Changes made, will merge once tests pass

Issue #367, opened 2019-06-25
Window scaling issue on Linux and Windows


marshuang80:
## 🐛 Bug

The drop down menu has scaling issues on linux
![image](https://user-images.githubusercontent.com/15271135/60139287-35665980-9762-11e9-8b3a-100b43d7fe51.png)

## To Reproduce
Run `python examples/labels-0-2d.py` and expand drop down menu for segmentation

## Environment

 - napari: 0.9
 - Linux Ubuntu 18.04LTS
 - Python version: 3.7
 - Laptop: Lenovo X1 Carbon 6th gen

sofroniewn:
Nice catch @marshuang80 - this is probably because we are hard coding a size for the `expanded` properties widget, but we might need to make it depend on things like screen resolution etc. I can look into it further.

In general we need to do a bit more testing of our stylings on linux and windows!

jni:
Note that I haven't run into these issues, and probably has to do with GNOME scaling on 2x (retina/HiDPI) displays (I don't have that turned on because I'm on i3).

It's *very* interesting that the layer name text is not scaled but all the other text is — that might be a hint about what "correct" styling looks like. Incidentally, I think the rest of the chrome should be doubled, rather than the text be made smaller — the window title text is a good guide.

marshuang80:
Seems like this issue only happens when the display on linux is scaled to anything larger than 100%. 

![image](https://user-images.githubusercontent.com/15271135/60553931-60b0f180-9cea-11e9-8e85-3bf8ee5ba655.png)

jni:
As noted by @hmaarrfk in #530, his library might be a good way to deal with this issue:

https://github.com/hmaarrfk/qtpy-tools

sofroniewn:
@hmaarrfk any PRs around this would be great, @marshuang80 and I found some settings that might have helped but couldn't quite get it working

hmaarrfk:
I've never coded up qt, just converted the UI files into py files with the qt tools. I typically do that step at build time.

sofroniewn:
Ok, I'll try and look into this again soon

jni:
For Windows, an image.sc user found a fix:

https://forum.image.sc/t/napari-issues-it-works-but-no-gpu-and-the-gui-is-all-squashed/31411/13

It would be great if we could figure out how to find that setting programmatically...

sofroniewn:
[...]

PR #600, opened 2019-10-18
Add Text layer

### LINKED ISSUE #593: Text layer
## 🚀 Feature
It would be nice to have a text layer built using the [vispy text visual](http://vispy.org/visuals.html#vispy.visuals.TextVisual) so that you can display a string or list of stings overlaid on the canvas

kevinyamauchi:
# Description
As discussed in #593, this PR adds a layer type for adding text to the viewer. To do so, I implemented a vispy [`Text` visual](http://vispy.org/scene.html#vispy.scene.visuals.Text). As suggested by @sofroniewn, the data are provided as a `(coords, text)` tuple where `coords` is a (NxD) array of N, D-dimensional coordinates for text positions and `text` is a list of strings to display.

From my read of the vispy docs and some experimentation, it looks like the `Text.font_size` and `Text.rotation` only take scalars, so adding per-text-element sizing and rotation doesn't look super straightforward.

For a demo, see `/examples/add_text.py`.

# To do:
- [x] Update the "hit box" for selecting text elements
- [x] Either include a way to add text from the UI or remove the add mode

## Type of change

# References
This PR closes #593.

# How has this been tested?
- [x] tests: /napari/layers/text/tests/test_text.py

## Final checklist:


kevinyamauchi:
Some open questions about the API and UI:

- The "hitbox" for selecting points is currently just an constant radius from the centroid. I couldn't find an easy way to retrieve the size of the text element. In the absence of that, my plan is to approximate the extent based on some function of the font size and string length, as suggested by @sofroniewn.
- Initially, I had an offset parameter that offsets the text location from the assigned coordinate. This was nice at the time because I had included markers to denote where the text referred to, but now that we have eliminated the markers, I am not sure how useful it is. Thus, I have removed the offset parameter. Thoughts?
- Adding annotations would be super nice, but the UI could be a little challenging. One idea would be a little text edit box that pops up where you click for specifying the text. Alternatively, there could be a text box in the controls that becomes visible in adding mode (and maybe selecting mode?). As discussed in #593, it may be best to push forward without the ability to add/edit text for this PR and revisit the feature later.

napari/_vispy/vispy_text_layer.py
```diff
@@ -0,0 +1,148 @@
...
+        # Update the text
+        if self.layer.dims.ndisplay == 2:
+            positions = np.flip(data, axis=1)
+            text_node = self.node._subvisuals[self._TEXT_NODE_INDEX]
+            self._update_text_node(
+                text_node,
+                text=text,
+                pos=positions,
+                color=self.layer.text_color,
+                font_size=self.layer.font_size,
```
sofroniewn: Right now it seems like the font size remains fixed on my screen regardless of zoom position, is there a way to make it rescale with zoom so it remains fixed relative to the image? On the markers layer I think this was called "rescale=True"? Do you agree that would be preferred behavior or do you like it as is? If we want to change it we might need to make a PR to vispy

napari/layers/text/_constants.py
```diff
@@ -0,0 +1,61 @@
...
+
+    SELECT allows the user to select points by clicking on them
+    """
+
+    ADD = auto()
+    SELECT = auto()
+    PAN_ZOOM = auto()
+
+
+class Symbol(Enum):
```
sofroniewn: Don't need these any more

napari/layers/text/_constants.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def __str__(self):
+        """String representation: The string method returns the
+        valid vispy symbol string for the Markers visual.
+        """
+        return self.value
+
+
+# Mapping of symbol alias names to the deduplicated name
+SYMBOL_ALIAS = {
```
sofroniewn: don't need these anymore

napari/layers/text/text.py
```diff
@@ -0,0 +1,841 @@
...
+        return str(self._mode)
+
+    @mode.setter
+    def mode(self, mode):
+        if isinstance(mode, str):
+            mode = Mode(mode)
+
+        if not self.editable:
+            mode = Mode.PAN_ZOOM
+
```
sofroniewn, line 333: I'd maybe put a line in here like
```python
if mode==Mode.ADD:
    mode = Mode.PAN_ZOOM
```
to prevent entering add mode for now, as i'd say that functionality can come when we have the ability to edit text too. Might also be nice to drop the add button from the GUI so its not clickable, but otherwise I'd try and keep as much as the add machinery in here that we might want to reuse.

unless we want to brainstorm supporting both adding and editing in this PR already - but I wouldn't do one without the other. I'd love it if you wanted to do both in this PR!!

napari/_vispy/vispy_text_layer.py
sofroniewn: see discussion here that looks out of data, but might be helpful - https://github.com/vispy/vispy/issues/1072

sofroniewn:
Coming along nicely - as to the selection box, yeah it's not clear to be if there's a way to grab the size of the text from vispy - there might be, and if you look at by rescaling comment above it might make it easier if the text does rescale, but otherwise estimating is probably fine.

I had kinda liked the offsets - but maybe it's more complicated than needed and pretty redundant. i think @jni didn't think it was needed, so fine to drop

for editing / adding new text, ideally we'd be able to type in the text box over the canvas, like illustrator, but that will require additionally rendering a line for where the typing is happening and updating everything really fast. Maybe for now add an edit box in the bottom of the layer controls panel, which when nothing is selected is empty, when one text box is selected has the text of that box in it, and if you type in that box it then updates the text over the visual - it's not quite as perfect, but that should be much easier to implement. Then you can keep the add button which will just add new text with an empty string. If multiple text boxes are selected then show the text if they all have the same label, or show empty if they have different (same as we do for colors) and then if you start editing then change all text labels - this would allow you to relabel many points

Have you tried sending lists of colors / lists of font sizes, sometimes they are supported by vispy even if the docs don't say so. I don't think we need to support rotated text at this point.

Can you also add an n-D example `nD_text.py` where we can see the behaviour both with slicing and 3D rendering

napari/layers/text/_constants.py
kevinyamauchi: Yikes. Thanks for catching that!

kevinyamauchi:
@sofroniewn, thanks for the feedback! In terms of adding/editing, I'd like to try adding the text box to the controls for editing/adding text just to see how it feels.

I have tried sending lists of font sizes and unfortunately, vispy expects a scalar value. I can try colors as well, but it be the case the we can only set those parameters on a per-layer basis.

I've been digging through vispy to see the best way to determine the size of the hitbox for selecting the text elements.

kevinyamauchi:
@sofroniewn , thanks again for the feedback. I've added text editing and the ability to add new next. I will work on scaling the hitbox next. In the meantime, let me know what you think!

kevinyamauchi:
Current editing behavior:
editing

Current new text behavior:
new

sofroniewn:
editing is looking great @kevinyamauchi!!

sofroniewn:
@kevinyamauchi I just saw the last commit is "remove ndim", I havn't looked at the code, but I'm not sure we want to drop that functionality, just thinking about the points layer it's maybe nice to have the ability to turn it on so that when you slice through layers text appears not only on that exact slice but the neighboring ones above and bellow. Could be a little weird if you have volumetric timeseries data. I'm not 100% we need to keep it, but curious you rational for removing it if that's what the commit did

kevinyamauchi:
I have a first pass at the hitbox scaling working. I think to be robust, I will have to get the viewer DPI since the point -> px conversion relies on that.

Hitbox scales with zoom and font size:
zooming

Hitbox expands/contracts with the text:
expanding

kevinyamauchi:
@sofroniewn, regarding the `ndim`option, I removed it because to me, the advantage of the `ndim` option in the points layer is that we can scale the point size along an axis. It didn't seem like we would want to do that with text and if the user wants the text in adjacent slices, they can just have data for each index the text should appear at. To me, that seemed more intuitive than a "size-like" parameter that describes the extent of the text in out-of-plane axes. Thoughts?

sofroniewn:
Yeah - fair enough @kevinyamauchi it will simplify things, fine to drop - we can always add back in after using it we really miss it.

sofroniewn:
love the hitbox scaling. the px / dpi conversion factors might be available in the Qt stuff somewhere, not quite sure, but I think we looked into it a while back in #367

kevinyamauchi:
A couple of updates:
- I've added a DPI property to the base layer, so we now use that to estimate the size of the text elements.
- I've added copy/paste to the text elements. Currently, when pasting, the pasted items have the same coords as their source and are highlighed. I believe this is how pasting works in other layers. Is this the desired pattern? Would it be better if they were pasted wherever the mouse is when the past command is issued? This would be convenient in some ways, but potentially problematic if the paste function were ever called programatically. Thoughts @sofroniewn , @jni , @royerloic ?

sofroniewn:
Hmm I'd leave the paste to be where the current is for now, can always revisit across all layer types in the future, but I don't have super strong opinions

kevinyamauchi:
I think this PR is mostly there. The linux and OS X tests are passing. Strangely, `black` is showing that `/napari/tests/test_viewer_layer_parity.py` needs to be reformatted, even though this PR doesn't touch that file and it is passing the check on my local machine. Any thoughts (cc @sofroniewn, @jni, @AhmetCanSolak)?

jni:
@kevinyamauchi annoyingly, black just released a new version and "changed its mind" about certain syntax. :man_facepalming: see #619. And the change is [dubious at best](https://github.com/napari/napari/pull/619/commits/7951ba6a8dde4bd5acf5636780ad26a4188c090b).

Anyway, once #619 gets merged you can merge master and black will be happy again. You will need to `python -m pip install -U black` to get the latest version on your machine.

kevinyamauchi:
Ah got it. Thanks for the quick response, @jni!

kevinyamauchi:
Sorry for the delay. I think it is finally ready for initial review!

napari/_vispy/vispy_text_layer.py
```diff
@@ -0,0 +1,152 @@
...
+            face_color = 'white'
+            text = []
+
+        highlight_text_node = self.node._subvisuals[
+            self._HIGHLIGHT_TEXT_NODE_INDEX
+        ]
+        self._update_text_node(
+            highlight_text_node,
+            text=text,
+            pos=np.flip(data, axis=1),
```
sofroniewn: this needs a rotation parameter too, right now the highlight is not rotated

sofroniewn: you might also now have to take rotation into account inside the selection box calculation - it doesn't seem to have been done yet

napari/_vispy/vispy_text_layer.py
```diff
@@ -0,0 +1,152 @@
...
+            self._update_text_node(
+                text_node,
+                text=text,
+                pos=positions,
+                rotation=self.layer.rotation,
+                color=self.layer.text_color,
+                font_size=self.layer.font_size,
+            )
+
+    def _on_highlight_change(self):
```
sofroniewn, line 104: right now when I deselect the layer the `_on_hightligh_change` isn't getting triggered so you can still see the highlighted text and selection box. Ideally if the layer isn't selected anymore nothing should be highlighted. This is done in the shapes layer, where I think something like `finsh_drawing` gets called

napari/layers/text/text.py
```diff
@@ -0,0 +1,820 @@
...
+        ----------
+        coord : sequence of indices to add point at
+        """
+        coords = np.append(self.text_coords, [coord], axis=0)
+        text = self.text
+
+        # Set the new text to 'EditMe' if none has been specified
+        new_text = self.new_text
+        if self.new_text == '':
+            new_text = 'EditMe'
```
sofroniewn: Not quite clear to me when you see `EditMe` vs what is already inside the text window. I'd also seems to save state in a way I don't understand. I'd like the newly created next to be whatever is currently displayed inside the textbox. It's possible for this not to be the case if you are just selecting text and then move to the text box.

Also right now if you just have a bunch of space like `  ` then it seems to behave a little strange.

Finally if we need a default - I prefer `edit` over `EditMe`

napari/_vispy/vispy_text_layer.py
```diff
@@ -0,0 +1,152 @@
...
+            lambda e: self._on_highlight_change()
+        )
+
+        self.layer.dims.events.ndisplay.connect(
+            lambda e: self._on_display_change()
+        )
+
+        self._on_display_change()
+
+    def _on_display_change(self):
```
sofroniewn, line 36: i've improved how some of this is done recently to remove repeated calls to `_on_data_change`, see edits to Points #629, can you adapt those here too

napari/_vispy/vispy_text_layer.py
```diff
@@ -0,0 +1,152 @@
...
+        parent = self.node.parent
+        self.node.transforms = ChainTransform()
+        self.node.parent = None
+
+        if self.layer.dims.ndisplay == 2:
```
sofroniewn, line 41: Right now entering 3D mode gives me errors for both the `add_text.py` and `nD_text.py`, there should be tests inside `tests/test_viewer.py` and `tests/test_napari.py` and `tests/components/test_viewer_model.py` that include both 2D and nD examples and `viewer.dims.ndim = 2` and `viewer.dims.ndim = 3` as is done for our other layers

For the `nD_text.py` when switching to 3D rendering I think you'll need to turn off some of the get_value / hitbox stuff, which I think is turned off in the shapes layer.

napari/components/viewer_model.py
```diff
@@ -571,6 +571,99 @@ def add_image(
...
+            text contains a list of the strings to be displayed as text.
+        rotation : float
+            Angle of the text elements
+        text_color : str
+            The color of the text font.
+        font_size : float
+            Size of the text font in points.
+        font : str
+            Text font. OpenSans is the default.
+        anchor_x : str
```
sofroniewn: Can anchor be a 2-tuple of str `('center', 'left')`. Can you also list here what are the acceptable string values (if there are only a few)

napari/components/viewer_model.py
```diff
@@ -571,6 +571,99 @@ def add_image(
                 layer_list.append(layer)
             return layer_list
 
+    def add_text(
```
sofroniewn, line 573: this is great, but we're you should add a `view_text` method to `view_layers.py` following the conventions there

napari/layers/text/keybindings.py
```diff
@@ -0,0 +1,66 @@
...
+
+        # on key release
+        layer.mode = prev_mode
+        layer.selected_data = prev_selected
+        layer._set_highlight()
+
+
+@Text.bind_key('P')
+def activate_add_mode(layer):
+    """Activate add points tool."""
```
sofroniewn: add "text"

napari/layers/text/keybindings.py
```diff
@@ -0,0 +1,66 @@
...
+    layer.mode = Mode.ADD
+
+
+@Text.bind_key('S')
+def activate_select_mode(layer):
+    """Activate select points tool."""
```
sofroniewn: select "text"

napari/layers/text/_constants.py
```diff
@@ -0,0 +1,18 @@
...
+
+from ...util.misc import StringEnum
+
+
+class Mode(StringEnum):
+    """
+    Mode: Interactive mode. The normal, default mode is PAN_ZOOM, which
+    allows for normal interactivity with the canvas.
+
+    ADD allows points to be added by clicking
```
sofroniewn, line 11: allows "text"

napari/layers/text/_constants.py
```diff
@@ -0,0 +1,18 @@
...
+
+    SELECT allows the user to select points by clicking on them
```
sofroniewn, line 13: select "text"

napari/layers/text/keybindings.py
```diff
@@ -0,0 +1,66 @@
...
+
+@Text.bind_key('Z')
+def activate_pan_zoom_mode(layer):
+    """Activate pan and zoom mode."""
+    layer.mode = Mode.PAN_ZOOM
+
+
+@Text.bind_key('Control-C')
+def copy(layer):
+    """Copy any selected points."""
```
sofroniewn: selected "text"

napari/layers/text/keybindings.py
```diff
@@ -0,0 +1,66 @@
...
+    if layer._mode == Mode.SELECT:
+        layer._copy_data()
+
+
+@Text.bind_key('Control-V')
+def paste(layer):
+    """Paste any copied points."""
```
sofroniewn: copied "text"

napari/layers/text/keybindings.py
```diff
@@ -0,0 +1,66 @@
...
+    if layer._mode == Mode.SELECT:
+        layer._paste_data()
+
+
+@Text.bind_key('A')
+def select_all(layer):
+    """Select all points in the current view slice."""
```
sofroniewn: all "text"

napari/layers/text/keybindings.py
```diff
@@ -0,0 +1,66 @@
...
+    if layer._mode == Mode.SELECT:
+        layer.selected_data = layer._indices_view[: len(layer._data_view)]
+        layer._set_highlight()
+
+
+@Text.bind_key('Backspace')
+def delete_selected(layer):
+    """Delet all selected points."""
```
sofroniewn: selected "text"

napari/layers/text/text.py
```diff
@@ -0,0 +1,820 @@
...
+        The color of the text font.
+    font_size : float
+        Size of the text font in points.
+    font : str
+        Text font. OpenSans is the default.
+    anchor_x : str
+        Positioning of the text relative to the coordinate. Default is 'center'.
+    anchor_y : str
+        Positioning of the text relative to the coordinate. Default is 'center'.
+    render_method : str
```
sofroniewn: I don't think we should expose this to our users - let's make a choice internally and hard code it as a private attribute. I'm worried about too many parameters that aren't critical for users to work with

sofroniewn (commented):
This is looking great, we're getting close. I left some comments - the most important ones are about 3D rendering and making sure we've got tests and methods for `napari.view_text` and `viewer.add_text`. It looks like the 3D rendering problems are mostly related to our model (I've seen those errors before) or maybe even fixed by some improvements in #629.

kevinyamauchi:
@sofroniewn , thanks for the review! I am heading out of town, but I will address your comments this weekend.

napari/layers/text/keybindings.py
kevinyamauchi: Sorry, I fixed these and other copy paste errors.

napari/_vispy/vispy_text_layer.py
kevinyamauchi: Got it. I'll work on this next.

napari/layers/text/text.py
kevinyamauchi: My intent here was to make new text items have some text value if the user hasn't entered anything into the text box. If there is anything in the text box, it should use that value for the new text element. The rationale is that it's really difficult to select text that is empty for future editing. Does this make sense?

Can you clarify what the issues you're seeing where state doesn't seem to be preserved correctly? I am probably missing a call to an update function somewhere.

I can change `EditMe` to `edit`!

napari/_vispy/vispy_text_layer.py
kevinyamauchi: I added the rotation parameter. I will add hitbox rotation next.

napari/_vispy/vispy_text_layer.py
kevinyamauchi: I fixed the 3D rendering. Would you mind taking a look and letting me know if you like the behavior? It seems like the text is presented at a fixed orientation to the viewer. This seems like the best way to handle 3D text to me, but I am happy to turn it off if we think it is strange. 

Great point about the lack of 3D tests. I will add those.

napari/layers/text/text.py
kevinyamauchi: Good point. Centering the anchors seems to be the way to go, as it makes calculating hitboxes, etc. easy.

Regarding the render mode, here's a snipped from the vispy docs:

> The ‘cpu’ method
        should perform better on remote backends like those based on WebGL.
        The ‘gpu’ method should produce higher quality results.

I made `cpu` default for max compatibility, but it seems like for our current use cases, `gpu` is safe.

What do you all think?

napari/_vispy/vispy_text_layer.py
kevinyamauchi: Thanks for catching this. I will look into it

kevinyamauchi:
Here is how the 3D rendering currently works. The text elements maintain their orientation relative to the viewer as the volume is rotated.

3D

napari/layers/text/text.py
jni: Let's go with GPU unless someone complains. We don't have plans to support WebGL for a long time. Maybe you can add that docs snippet where you set the parameter, so that we can find it easily in the future.

napari/layers/text/text.py
sofroniewn: Definitely agree - the empty text is hard to edit. The update was being missed if you go into the selection tool, then select new text, that text appears in the top left control panel, but then if you go back into the add tool and create a new text box it won't use that recently selected and previously visible text, it will use the last "saved" text. I was suggesting dropping that concept of "last saved" and use what ever is visible, unless it is an empty string, and then use the default. I don't have super strong feelings on this, so if this doesn't make sense don't worry / I can try making a gif / give you a demo sometime

napari/layers/text/text.py
kevinyamauchi: Sounds good. Done!

napari/_vispy/vispy_text_layer.py
kevinyamauchi: Tests added!

napari/_vispy/vispy_text_layer.py
kevinyamauchi: Fixed!

napari/_vispy/vispy_text_layer.py
kevinyamauchi: Done!

napari/_vispy/vispy_text_layer.py
kevinyamauchi: Done!

napari/layers/text/text.py
kevinyamauchi: I have updated the behavior such that the last selected text is preserved. For example, if you select an element with text "hi" and then switch to adding mode, the text of the element you add will also be "hi". This is also preserved if you switch to pan/zoom and then back to add. I think this is what you were describing. Please let me know what you think!

kevinyamauchi:
@sofroniewn , I think I have addressed your review points and this is ready for re-review!

edit: oops some tests are failing...fixing now
edit 2: tests fixed

Selection outline and hitbox rotation:
rotation

The new text adding behavior (saving last new text):
text add

kevinyamauchi:
Good morning! I just merged master and the Windows and OS X tests are still failing because `numcodecs` won't install. Is anybody else having this issue (cc @sofroniewn , @tlambert03, @jni )?

Edit: here's a relevant snippet of the log

```
  1 warning and 1 error generated.
  [numcodecs] command 'gcc' failed with exit status 1
    File "/Users/anka/conda/lib/python3.7/distutils/unixccompiler.py", line 118, in _compile
      extra_postargs)
    File "/Users/anka/conda/lib/python3.7/distutils/ccompiler.py", line 910, in spawn
      spawn(cmd, dry_run=self.dry_run)
    File "/Users/anka/conda/lib/python3.7/distutils/spawn.py", line 36, in spawn
      _spawn_posix(cmd, search_path, dry_run=dry_run)
    File "/Users/anka/conda/lib/python3.7/distutils/spawn.py", line 159, in _spawn_posix
      % (cmd, exit_status))
  distutils.errors.DistutilsExecError: command 'gcc' failed with exit status 1
  During handling of the above exception, another exception occurred:
    File "/private/var/folders/sy/2x5qvs0n4lg18fry9jz4y21m0000gn/T/pip-install-tlzrvkay/numcodecs/setup.py", line 286, in build_extension
      build_ext.build_extension(self, ext)
    File "/Users/anka/conda/lib/python3.7/distutils/command/build_ext.py", line 534, in build_extension
      depends=ext.depends)
    File "/Users/anka/conda/lib/python3.7/distutils/ccompiler.py", line 574, in compile
      self._compile(obj, src, ext, cc_args, extra_postargs, pp_opts)
    File "/Users/anka/conda/lib/python3.7/distutils/unixccompiler.py", line 120, in _compile
      raise CompileError(msg)
  distutils.errors.CompileError: command 'gcc' failed with exit status 1
  During handling of the above exception, another exception occurred:
    File "<string>", line 1, in <module>
    File "/private/var/folders/sy/2x5qvs0n4lg18fry9jz4y21m0000gn/T/pip-install-tlzrvkay/numcodecs/setup.py", line 362, in <module>
      run_setup(with_extensions)
    File "/private/var/folders/sy/2x5qvs0n4lg18fry9jz4y21m0000gn/T/pip-install-tlzrvkay/numcodecs/setup.py", line 355, in run_setup
      license='MIT',
    File "/Users/anka/conda/lib/python3.7/site-packages/setuptools/__init__.py", line 145, in setup
      return distutils.core.setup(**attrs)
    File "/Users/anka/conda/lib/python3.7/distutils/core.py", line 148, in setup
      dist.run_commands()
    File "/Users/anka/conda/lib/python3.7/distutils/dist.py", line 966, in run_commands
      self.run_command(cmd)
    File "/Users/anka/conda/lib/python3.7/distutils/dist.py", line 985, in run_command
      cmd_obj.run()
    File "/Users/anka/conda/lib/python3.7/site-packages/wheel/bdist_wheel.py", line 192, in run
      self.run_command('build')
    File "/Users/anka/conda/lib/python3.7/distutils/cmd.py", line 313, in run_command
      self.distribution.run_command(command)
    File "/Users/anka/conda/lib/python3.7/distutils/dist.py", line 985, in run_command
      cmd_obj.run()
    File "/Users/anka/conda/lib/python3.7/distutils/command/build.py", line 135, in run
      self.run_command(cmd_name)
    File "/Users/anka/conda/lib/python3.7/distutils/cmd.py", line 313, in run_command
      self.distribution.run_command(command)
    File "/Users/anka/conda/lib/python3.7/distutils/dist.py", line 985, in run_command
      cmd_obj.run()
    File "/private/var/folders/sy/2x5qvs0n4lg18fry9jz4y21m0000gn/T/pip-install-tlzrvkay/numcodecs/setup.py", line 279, in run
      build_ext.run(self)
    File "/Users/anka/conda/lib/python3.7/distutils/command/build_ext.py", line 340, in run
      self.build_extensions()
    File "/Users/anka/conda/lib/python3.7/distutils/command/build_ext.py", line 449, in build_extensions
      self._build_extensions_serial()
    File "/Users/anka/conda/lib/python3.7/distutils/command/build_ext.py", line 474, in _build_extensions_serial
      self.build_extension(ext)
    File "/private/var/folders/sy/2x5qvs0n4lg18fry9jz4y21m0000gn/T/pip-install-tlzrvkay/numcodecs/setup.py", line 289, in build_extension
      raise BuildFailed()
  __main__.BuildFailed
  ----------------------------------------
```

sofroniewn:
yes, the CI issues are being discussed in #665

kevinyamauchi:
> yes, the CI issues are being discussed in #665
Ah just saw that.

The tests are passing in Linux, so I think this PR is ready for re-review!

royerloic:
Looks awesome @kevinyamauchi !

jni:
LMAO I was playing around with the layer options and I almost spit my coffee out when I set the blending to "opaque":

Screen Shot 2019-11-15 at 10 57 20 am

😂 

I'm noticing that the text looks much smoother when the mode is set to "additive". On translucent, I see weird and ugly fringes around the text. Should we maybe set additive as the default for this layer? I'm not entirely sure of all the consequences, though...

Full review in progress...

jni:
Ok so let me comment here on my API proposal for points/annotations. I think the ideas below will work with both points and higher-dimensional (lines, shapes, volumes?) layers.

Ignore my names for the variables here, they can be debated later. Just the concepts.

(1) Points are *required* to have a data point, `coords=`, that is conceptually an npoints x ndim array.
(2) They can also have other properties, including color, size, marker, and annotations. These properties are per-point, and they can be straight up lists: `add_points(coords, color=colorlist, size=sizelist, annotations=strlist)`.
(3) They can, however, also be columns in an dataframe (or dataframe-like object), and they can then be addressed by column name. We can call the dataframe `properties`. Let's say properties has column names 'annotation', 'vesicle density', 'pre/post-synaptic', 'neurotransmitter'. Then you could call `add_points(coords, properties=props, size='vesicle density', marker='pre/post-synaptic', color='neurotransmitter')`. The properties would then be mapped appropriately using color cycles and shape cycles, just like matplotlib/seaborn do.
(4) annotations by default would be centered on their point, but we can add an offset= argument that contains individual offsets, and the argument could also be set to "auto" to do something like [ggrepel](https://cran.r-project.org/web/packages/ggrepel/vignettes/ggrepel.html).

As mentioned, the above paradigm could also be used to annotate lines, shapes, etc, and take the handling of mapping data properties to visuals away from users and into our layers, which, as the various declarative data visualization libraries have shown, is a honking good idea. All of the mappings would of course be configurable, but they would remain automatic, and the list input will always remain an option for everyone who likes doing stuff by hand.

I hope the above clarifies my thinking on this!

sofroniewn:
@jni those points make sense - and I think are more towards the approach 2 described here https://github.com/napari/napari/issues/651#issuecomment-552056627. Looking at the issue, I think the last post from @kevinyamauchi was leaning more towards approach 1 where there is still an explicit `Text` layer and `add_text` method, so I'm curious what he makes of this and the following.

Maybe I'm coming around to the @jni viewpoint here, this would be very clean from an API standpoint in certain ways and make us conform with the next generation of plotting libraries (this is really a scatter plot method in many ways), but @kevinyamauchi raised some good concerns about the API from other perspectives - we'd now have `edge_color`, `face_color`, `text_color`, `size`, and `text_size` etc. - which is maybe ok, but also maybe bloated. We could though reduce some of the UI complexity by having the concept of selecting the marker and text be one by default (i.e. you couldn't independently drag them by default, but maybe only from a special mode).

I'm also reconsidering my previous statements around text being part of `Shapes` too. I think this new layer would be the only way to add text and you could easily do with with / without a marker present `symbol=None` say.

I don't want to rat hole on names, but maybe we'd drop the name `Points` and we'd rename it `Annotations`, which makes it much more clear on how we view this layer and what it will be used for. If you want to have an image and some text you'd need to have an `Image` layer and an `Annotations` layer, which I think is better than say just having an `Image` layer with `Annotations` baked in. If you wanted to have shapes with text labels you'd have to have a `Shapes` layer and an `Annotations` layer too, which keeps those concepts nicely separable. If you just want some points and text you'd only need an `Annotations` layer.

I guess the main issue right now for this PR is that we still havn't resolved the question of if there will ever be a `viewer.add_text` method. If not, than I guess I have some concern about merging this PR and adding one, only to take it away shortly. I think these are the types of breaking changes people find very confusing. I know though that @kevinyamauchi put a lot of work into this PR and it has been sitting here for a while though so I want to be respectful of that. I think a lot of the concepts though would be reused in the `Annotations` layer.

I also worry though the longer we wait to decide the more it holds up changes in other parts of the repo. For example #686, #782, #772, and #694.

In summary, I guess my current thoughts are that I lean towards a single `add_annotations` method which has both a `Markers` and `Text` visual (`symbol=None` and/or `text=None` we allow you to use one without the other) and that should ultimately not merge this PR which adds an `add_text` method. I think we should get #694, #782, #686 and #772 done first - #772 has the possibility to dramatically simplify the new layer by moving a lot of the interactivity (like bounding box drawing, highlighting etc. to the base layer) and then we embark on this. 

@jni does this capture how you feel too? What are your thoughts on all this @kevinyamauchi?

kevinyamauchi:
Thanks for the comments @jni and @sofroniewn ! I haven't had a chance to review them yet and probably won't until this weekend. I just wanted to drop by to quickly mention that if we decide a different configuration/API is preferred (e.g., no standalone `Text` layer and/or `viewer.add_text()`), I don't mind and do not consider this wasted work. I think it's totally fine to take what we learned here and implement it in a way that makes the most sense.

pablooriol2:
Hi all,

I was getting excited when I saw this PR as I've been wishing to be able to add text to a label layer (particularly useful for 2-D time series tracking/curation).

I was wondering if there's been any progress on this?

Unfortunately, I couldn't find the mentioned `/examples/add_text.py`...

Is there an easy/straightforward-ish way to add text to a labels layer? I just want to be able to see the cells' 'names' 😆

FYI I know the label ID can be checked by hovering over the region/mask of interest (I don't know what I'd do without the bottom left info panel!), but having text is much more efficient in my opinion when it comes to looking for a particular label ID

kevinyamauchi:
Hey @pablooriol2 ! Funny timing - I was just thinking about this PR earlier today. In working on this PR, we realized that we needed to do some work refactoring the data model for our points layer. Unfortunately, I haven't been able to push this any farther. However, that refactor is nearly complete, so I will be ready to get back to this in the coming weeks.

> Is there an easy/straightforward-ish way to add text to a labels layer? I just want to be able to see the cells' 'names' 😆

There isn't an easy to way to add text at the moment, but you may be interested in this PR, which was recently merged that adds properties to labels #1281 .

If there are specific features you would like or ideas you have, we would love it if you opened an issue describing it!

pablooriol2:
Hi @kevinyamauchi ,

Thanks for getting back so quickly.

Ooh looks like it might be a feature for the 0.3.3 release? In any case, I'm really looking forward to it!

No specific features in mind, just being able to display the label's ID as text. Just as this [example](https://github.com/quantumjot/arboretum/blob/master/examples/napari.png) taken from quantumjot's napari tracking [plugin](https://github.com/quantumjot/arboretum).

kevinyamauchi:
> Thanks for getting back so quickly.

No problem!

> Ooh looks like it might be a feature for the 0.3.3 release? In any case, I'm really looking forward to it!

I think it is unlikely to make it into 0.3.3, but I will likely start working on it next week. Hopefully all goes smooth :)

> No specific features in mind, just being able to display the label's ID as text. Just as this example taken from quantumjot's napari tracking plugin.

Okay, well don't hesitate to drop an issue if anything comes up! By the way, @quantumjot's tracking plugin looks awesome!

pablooriol2:
> I think it is unlikely to make it into 0.3.3, but I will likely start working on it next week. Hopefully all goes smooth :)

Finger crossed!

> Okay, well don't hesitate to drop an issue if anything comes up! By the way, @quantumjot's tracking plugin looks awesome!

Yes, indeed it does! Very pumped for that plugin to be fully implemented & functional 😃

kevinyamauchi:
FYI @pablooriol2 , I have picked up adding text in #1374 . Feel free to chime in if you have any comments and/or suggestions!

Issue #596, opened 2019-10-16
add_* methods accept file paths / urls to data


sofroniewn:
## 🚀 Feature
As we start adding more fileIO should our `napari.view_*` methods and `viewer.add_*` methods start accepting paths and lists of paths to data as well as the data objects themselves.

We'd use the same machinery for fileIO from the dropdown menu, but now someone could leverage that functionality from a script. 

```python
napari.view_image('path_to_astro.png', rgb=True)
```

Would be pretty cool (note that this file IO would not go inside the Image Model files it would get called from inside the viewer file, keeping our models clean). I really like that this would unify fileIO for each of our layer types with its layer type, and it then makes it more clear where to add fileIO functionality when you add a layer type, and simplify our dropdown menu / drag and drop functionality as they only now need to be responsible to providing the correct path / list of paths to the correct `add_*` method and the rest will get handled from inside that method exactly the same regardless of where the data came from.

This will also make it really easy to support viewing of remote datasets, for example I want to be able to run

```python
napari.view_image('s3://my_bucket/path_to_image.zarr', colormap='red)
```

to just view my remote data on s3. We can then also keep track of this path and that the data was remote so that we serialize the viewer we just save this path, so if I send you the serialized `.napari` file it will be really small and you can just open it and see exactly what I saw, which is what was requested in #234 and will be pretty cool functionality for people working with remote data.

What are your thoughts on this @jni?

jni:
I'm not super in love with this idea... In the end in scikit-image we found that allowing the same parameter to take lots of different types is generally a bad idea, as it leads to a lot of increased complexity, and consequent brittleness. When we do want to do this for convenience, I think it's best to contain it, so I would try to keep all this magic within `io.magic_read`. 

So, I'd need to be convinced that the benefits outweigh the long-term maintenance costs, which are substantial.

The viewer serialization application is very cool, but I wonder if it can be implemented in some less magical way, e.g. `layer.data_path = path` (Which is what you would do internally anyway I believe...?

sofroniewn:
Ok, I appreciate your concerns about the complexity and maintenance burden - for example right now with all our existing layer types it is fairly unambiguous if a string is an image array or a file path, but as soon as we have a text layer is that a string that is the string of text we want to display, or is it a path to a csv file of text that we want to display. Or maybe even in the image case someone has an image as a string of bytes that needs to be converted etc.

[...]

PR #601, opened 2019-10-19
add loading from path for images

sofroniewn:
# Description
As discussed in #596 this PR adds loading of images (including multichannel) directly from path using the new `path` keyword argument.

The following now all work:
```python
viewer.add_image(path='astronaut.png', rgb=True) # load simple rgb image
viewer.add_multichannel(path='astronaut.png', axis=2) # load rgb image as multichannel
viewer.add_image(path='./data/LLSM/AOLLSM_m4_560nm.zarr') # load 100 GB 4D data
viewer.add_image(path='./data/camelyon16/tumor_001.zarr', is_pyramid=True) # load 60 GB pyramid
viewer.add_image(path='s3://sofroniewn/image-data/smFISH/raw.zarr', contrast_limits=(140.0, 1300.0)) # load remote zarr file
```
which is pretty cool I think, and previously discussed as useful with @freeman-lab 

A couple questions are - right now `magic_imread` also has keyword arguments around using dask, and stacking layers which I havn't exposed to the `add_image` and `view_image` methods, largely because I'm worried about keyword explosion, but we could if @jni thought they were appropriate.

Also right now we don't have any dask caching turned on - this can be done with
```python
from dask.cache import Cache
cache = Cache(2e9)  # Leverage two gigabytes of memory
cache.register()
```
(though i think dask cache is an optional dask dependency) and we probably only want to do if people are using dask. Do you think this should be done automatically inside magic read if dask gets used @jni and we should have some reasonable defaults? Or should it be settable? I'm open, but for the s3 example, it's almost certainly worth it.

Finally - I left the image reading inside `__main__.py` as is right now, I'd sort of prefer that to use the same code as the `add_image` etc, but there is a bit more logic going on there. I'm happy to leave it but wary of the CLI diverging to far from drop downs etc.

## Type of change

## Final checklist:


sofroniewn:
Oh also note that the doc stings for the `add_image` method and `Image` are now diverging because we have added a `path` keyword arg to the former, but not the later, so it might be time to weaken our doc string tests a little too

napari/util/io.py
```diff
@@ -62,9 +63,19 @@ def magic_read(filenames, *, use_dask=None, stack=True):
...
+            if isinstance(zr, zarr.core.Array):
+                # load zarr array
+                image = da.from_zarr(filename)
+                if shape is None:
+                    shape = image.shape
+            else:
+                # else load zarr all arrays inside file
+                image = [
+                    da.from_zarr(filename, component=c) for c, a in zr.arrays()
+                ]
```
jni: What's the story with this change? This is a zarr dataset?

jni: If so... This clause is getting kinda big and clunky, maybe let's add sub-function here, `read_zarr_dataset`?

sofroniewn: this now supports out image pyramid format - it's getting at the difference between a zarr array and a zarr group - the first clause of the if statement applies when the top level object in the zarr file is just a zarr array, the second clause comes into play when it is a zarr group. dask_array.from_zarr can only handle dask arrays

Example of zarr array at top:
image

Example of zarr group at top:
image

sofroniewn: i've pulled this out into it's own function

sofroniewn:
I've updated the PR - it is still needs tests.

For tests - should I start adding small jpg / png / tiff / zarr / zarr-pyramid / folder of png etc. to the repo so we can test against these? probably a good idea? i'll do that next

jni:
For small things, I guess it's ok. But it might be better to generate them from scikit-image data. Note that you can access the scikit-image files directly in `skimage.data.datadir`. I think I used a couple of those in my io tests.

It shouldn't be too hard to make zarr files etc from skimage or random data. Random data is ok as you can test round trips etc.

sofroniewn:
ah yes - exactly - i see how you've done it in your io tests, i will just re-use that approach

sofroniewn:
I've now updated the test and this is ready for review - the tests still fail though as the docstrings no longer match.

I'd like feedback around the following things, and whatever else people catch:

- [x] what to do about the non-matching docstrings between `add_image` and `Image`. I'm tempted just to make the tests slightly less strict so they pass.
- [ ] How to do any dask cacheing? I think this is important and we should add it in this PR
- [ ] Do we want to make the CLI loading more in-line with the other io - right now it will make some assumptions about Label etc. - I'm also fine with leaving it for now, especially if @jni is using the functionality as is
- [ ] do we need to expose more of the `magic_imread` keyword args like `use_dask` and `stack`? I lean no as I think our use_dask convention is good, especially if we add a cache, and i don't think stack should be true if you are using `add_image` - that is when you should use `add_multichannel` instead, but maybe those apis need to be better integrated. It's not the main focus of this PR, but I'm wondering now actually if we want to drop `add_multichannel` from our api and instead just have a keyword argument `add_image(data, multichannel_axis=None)` to add_image that does the same thing, and let add_image take lists of colormaps etc for that usecase ... We can have a separate thread about that though

jni:
@sofroniewn I don't have time for a full review in the next couple of days but I'll try to answer/opine on (😜) your questions so you can keep moving:

> what to do about the non-matching docstrings between add_image and Image. I'm tempted just to make the tests slightly less strict so they pass.

Yes, from what I could tell here, the way Kira did the tests is to create a string from both signatures and then compare the strings. Instead we could have a test to ensure that all the arguments in `Image.__init__` are present in `add_image`, *and* that the defaults and documentation match. You can do this with numpydoc, inspect, or a combination of both. 

> How to do any dask cacheing? I think this is important and we should add it in this PR

How important? Have you compared the performance at all? I'd be very curious. At any rate I think this could be done with user instructions for now (maybe a note in the docstring under 'path'). In the future, we will want to create a proper configuration system for napari, and then we can configure the dask cache. We could/should also provide a context manager to ensure that the cache is freed after the user is done with napari.

> Do we want to make the CLI loading more in-line with the other io - right now it will make some assumptions about Label etc. - I'm also fine with leaving it for now, especially if @jni is using the functionality as is

I'm not directly using it, so you could go wild, but I'd recommend holding off on this till another PR. PRs are easier to get through if they are small and self contained.

> do we need to expose more of the magic_imread keyword args like use_dask and stack?

Maybe. I think this will actually be a prerequisite for allowing different size image collections and stuff. It's unclear how to both expose those *and* harmonise functionality like drag-and-drop, where you don't want any interaction.

> It's not the main focus of this PR, but I'm wondering now actually if we want to drop add_multichannel from our api and instead just have a keyword argument add_image(data, multichannel_axis=None) to add_image that does the same thing, and let add_image take lists of colormaps etc for that usecase

I'm very much not a fan of this, because the API gets super complex then, with broadcasting logic that is hard to follow (see e.g. the `size` discussion for points). So I'd like to leave this be for now until there is a clear need.

sofroniewn:
ok - what I take away from these comments are that cacheing can come in a followup PR. changing the CLI loading can also come in a followup PR. no need to expose the `magic_imread` keyword args immediately, but maybe in the future, and that we like `add_multichannel`.

I will fix our doc string test, and then I think we're good to go

sofroniewn:
I've now fixed the viewer / layer tests by dropping the path parameter from the comparisons. 

This is now ready for review @jni. I'd like to get it in so I can press on with #616

napari/util/tests/test_io.py
```diff
@@ -66,41 +79,63 @@ def test_irregular_images(irregular_images):
...
+        z[:] = image
+        image_in = io.magic_imread([fout])
+        # Note: due to lazy loading, the next line needs to happen within
+        # the context manager. Alternatively, we could convert to NumPy here.
+        np.testing.assert_array_equal(image, image_in)
+
+
+@pytest.mark.skipif(not zarr_available, reason='zarr not installed')
+def test_zarr_pyramid():
+    np.random.seed(0)
```
jni: Why are you including the seed here?

jni (approved):
Looks good! I left a very minor comment. I'm also not convinced about renaming `magic_read` to `magic_imread` (I had ambitions about expanding it's magical powers to other layers), but I don't feel strongly about this.

napari/util/tests/test_io.py
sofroniewn: i've generally put one in whenever i call np.random.random, which happens in the next line, even if it shouldn't make a difference. nice to have that known reproducibility i think. i just realized i missed the seed in the test above so i will add it in there for completeness too

jni: No don't do that! LOL

jni: We do that in scikit-image because it's easy and for historical reasons. But the correct way to do this is to report the current seed (or choose a random one and then display it) so that your input is actually random. It helps to catch bugs that depend on random input.

sofroniewn:
i added the seed back in - for `magic_read` let's see how the reading stuff starts to progress as we add IO for more layer types - if that's how it happens happy to rename, but we might also end up with a function that is `magic_read` that contains `magic_imread`, `magic_shape_read` etc etc., although then i might want to rename things to be `read_image`, `read_shapes`, `read_multichannel` etc. as then it conforms nicely to out `add_*` and `view_*` naming conventions.

it looks like our windows test still failed here, so not sure why yours magically passed!!

i will merge this once our other tests pass though

napari/util/tests/test_io.py
sofroniewn: hmm ok - we do this in many places in the codebase right now, so we should probably go through and systematically fix if needed - remember when we did this at scipy, it was because occasionally the vispy triangulation code was failing. i'll pull the seed from these two functions and make a new commit, but followup will be needed elsewhere

PR #602, opened 2019-10-19
Move docs

### LINKED ISSUE #595: Repo cleanup on github and pypi
## 🐛 Not a bug but a cleanup request

We have some stall repos and packages on github and pypi.
pypi: napari-gui
github: image-types, plugin-cookiecutter,  ...

I think it would be great to clean that up and for example delete
napari-gui so as to avoid confusion and in general clutter... 
 
We can discuss here what can be deleted or somehow archived.
Perhaps we could have an archive repo on github  to keep old
unused code, i.e. some of Kira's  original plugin machinery for 
reference and future use. This is better than having repos that
appear legit to newcomers but are in fact completely stale. An 
archive repo would be a very explicit place to retire such old
stuff while keeping them for future reference.

sofroniewn:
# Description
This PR closes #595 by moving some more of our markdown files from the top level of the repo into the docs folder. Can people please check that all the links still work. I think we'll also have to change a link of two on the tutorials repo.


AhmetCanSolak (approved):
Thank you @sofroniewn for this cleaning PR.

PR #603, opened 2019-10-19
fix dim change when no layers

sofroniewn:
# Description
This PR fixes a bug that prevented the dims model from reseting to its default values when no layers are present. It lead to weird / confusing behaviour where you could add a 4D image, get 2 dims sliders present, deleted the 4D image, still have the two sliders present, add a new 2D image and still have the two sliders present, which I think was quite strange.

Now when there are no layers present, the `ndim` will return to 2 and no sliders will be present.

## Type of change

# How has this been tested?

- [x] I've added a test for this inside the viewer model tests

## Final checklist:

### DIFF
--- a/napari/components/dims.py
+++ b/napari/components/dims.py
@@ -147,18 +147,7 @@ class Dims:
             return
         elif self.ndim < ndim:
             for i in range(self.ndim, ndim):
-                # Insert default values for the new axis at the beginning of
-                # the lists
-                # Range value is (min, max, step) for the entire slider
-                self._range.insert(0, (0, 2, 1))
-                # Point is the slider value if in point mode
-                self._point.insert(0, 0)
-                # Interval value is the (min, max) of the slider selction
-                # if in interval mode
-                self._interval.insert(0, (0, 1))
-                self._mode.insert(0, DimsMode.POINT)
-                self._order = [0] + list(np.add(self.order, 1))
-
+                self.set_initial_dims(0, insert=True)
             # Notify listeners that the number of dimensions have changed
             self.events.ndim()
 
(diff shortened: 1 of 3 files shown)


jni (approved):
Very nice change! This wasn't on my radar but I agree that it's the right behaviour. The only thing that I would change is to make the `len(layers) == 0` clause call out to a method, `set_initial_dims`, that would do exactly the same thing as the initialisation (as in: initialisation would also call this function). This way we guarantee to always have the same behaviour when launching an empty viewer as when deleting all layers, and we reduce the code a little bit also. =)

royerloic:
Good catch @sofroniewn !

sofroniewn:
@jni I have consolidate this function - it didn't end up saving too much code as I had to deal with the special case of inserting a new axis and reseting an existing one separately. Let me know if you think the implementation can improve, otherwise feel free to merge once the tests pass

jni:
Once again: nice pre-emptive debugging! I just ran into this bug:

```python
import napari
import numpy as np


with napari.gui_qt():
    viewer = napari.view_image(np.random.random((51, 5, 2, 196, 171)),
                               name='mitosis')
    viewer.layers.remove('mitosis')
    viewer.add_image(np.random.random((10, 256, 256, 256)), name='blobs')
```

results in:

```pytb
 $ python test-remove.py
  File "test-remove.py", line 9, in <module>
    viewer.add_image(np.random.random((10, 256, 256, 256)), name='blobs')
  File "/Users/jni/projects/napari/napari/components/viewer_model.py", line 479, in add_image
    self.add_layer(layer)
  File "/Users/jni/projects/napari/napari/components/viewer_model.py", line 384, in add_layer
    self.layers.append(layer)
  File "/Users/jni/projects/napari/napari/util/list/_model.py", line 56, in append
    self.events.added(item=object, index=len(self) - 1)
  File "/Users/jni/projects/napari/napari/util/event.py", line 508, in __call__
  File "/Users/jni/projects/napari/napari/util/event.py", line 529, in _invoke_callback
    cb_event=(cb, event),
  File "/Users/jni/projects/napari/napari/util/event.py", line 523, in _invoke_callback
    cb(event)
  File "/Users/jni/projects/napari/napari/components/viewer_model.py", line 94, in <lambda>
    self.layers.events.added.connect(lambda e: self._update_grid())
  File "/Users/jni/projects/napari/napari/components/viewer_model.py", line 1193, in _update_grid
    stride=self.grid_stride,
  File "/Users/jni/projects/napari/napari/components/viewer_model.py", line 1180, in grid_view
    self._subplot(layer, (i_row, i_column))
  File "/Users/jni/projects/napari/napari/components/viewer_model.py", line 1208, in _subplot
    scene_size, corner = self._scene_shape()
  File "/Users/jni/projects/napari/napari/components/viewer_model.py", line 263, in _scene_shape
    size = [size[i] for i in self.dims.displayed]
  File "/Users/jni/projects/napari/napari/components/viewer_model.py", line 263, in <listcomp>
    size = [size[i] for i in self.dims.displayed]
IndexError: index 4 is out of bounds for axis 0 with size 4
```

This PR fixes it. I think there's probably improvements to be had in this code, but since I'm a little swamped and it fixes a significant bug (currently, a viewer cannot be safely "reused"), I'm just going to merge, and leave improvements for later PRs.

jni:
Also: man, writing documentation is the *best* way to fix bugs. I've been writing this blog post for the best part of a month and keep running into stuff I want to fix before publishing it. :joy: Almost there! 🤞

Issue #420, opened 2019-07-15
Features to 0.3 release


sofroniewn:
## 🚀 Feature
This issue has some features moved over from our `0.2` release #301. They are considered more advanced topics that we'll likely be aiming to have explored or supported for our `0.3` release, which might be around Jun 1 2020.

**Dims features**:
- [ ] Enable range mode for dims including slice / max_projection / crop (S/P/C)
- [x] Add play button to dimension sliders.
- [ ] Support units in status update and scaler bar overlays
- [x] Add axes labels
- [x] Add axes shapes

**File I/O**:
- [x] Loading / saving shapes, points, vectors
- [x] Loading / saving surface
- [x] Loading custom image file types
- [x] Loading / saving entire viewer state to zarr

**Rendering**
- [ ] 3D rendering of big data + image pyramids
- [ ] 3D interactivity / drawing
- [ ] Support mixed 2D and 3D rendering
- [ ] Editing in labels pyramid
- [ ] Lazy rendering

**1D plots**
- [ ] Scatter, line, histogram plots
- [ ] Live updating, with links

**Colors / clims**:
- [ ] For image layer add a clims histogram 
- [ ] Add a colorbar

**Animations**:
- [x] Support for making animations.

**Layers**:
- [ ] Support layer borders #298
- [ ] Support draggable / resizable layers #299
- [x] Add text layer #593 
- [ ] add events on finishing adding shapes / markers / labels / refresh etc

**Multicanvas**:
- [ ] Support multiple canvas with linked cameras
- [ ] For grid mode + coupled zoom
- [ ] Minimap when highly zoomed in
- [ ] For XY, XZ, YZ + 3D views for people doing volumetric imaging

**Provenance logging**
- [ ] Undo / redo buttons
- [ ] History feature
- [ ] Macro generation

**Qt / stylings**
- [ ] autogenerate `_qt_properties` /  `_qt_controls` from function annotations
- [x] display loading icon during start-up
- [x] Make side panels draggable, dockable
- [ ] Support title bar (with editable title and proper stylings)
- [ ] Status bar shows values from multiple layers
- [ ] Add edit boxes next to sliders to set min / max

**Typing / schema**
- [ ] Add typing system for images / shapes
- [ ] Adopt a metadata schema compatible with xarray for images / shapes

**Remote access**
- [ ] remote notebooks
- [ ] web viewing
- [ ] remote data with dask
- [ ] ability to launch from jupyterhub with jupyterlab

**Standalone app**
[...]

Issue #14, opened 2018-10-18
Pass xarray to napari-gui for autolabeling sliders


sofroniewn:
xarray is a library for N-D labeled arrays and datasets in Python http://xarray.pydata.org. If we allow xarrays to be passed to napari-gui then we could autopopulate labels of sliders etc. based on the names of the dimensions in the xarray object

jni:
Just to clarify, I think xarray support would be brilliant, but I just want to make sure that we don't become *dependent* on it. Bare NumPy arrays should be first-class citizens.

AhmetCanSolak:
@jni can you share why you think in that way? Just curious.

kne42:
This sounds like a nice convenience functionality; I'll keep it in mind!

Also, I'd like to note that you can already use `DataArray`s and other `numpy.ndarray`-like objects since we never explicitly call any NumPy functions on them (which would autoconvert them to strict `ndarray`s).

jni:
@AhmetCanSolak the entire SciPy ecosystem has thrived by using NumPy arrays as a lingua franca between packages. We are currently seeing some fragmentation as people test the boundaries of what can be done with arrays. I don't want to contribute to this fragmentation, and NumPy itself is taking steps to address it (see [NEP18](https://www.numpy.org/neps/nep-0018-array-function-protocol.html)). There is a good chance that supporting NumPy arrays will eventually let us support any array in the future "for free". The converse is not true.

AhmetCanSolak:
Thank you @jni !
It is nice to see people are considering this in NEP18. 

For our use, I dont think we will have issues with either labeled or unlabeled arrays, as soon as we document it well and be strict on our implementations.

ambrosejcarr:
> Just to clarify, I think xarray support would be brilliant, but I just want to make sure that we don't become dependent on it. Bare NumPy arrays should be first-class citizens.

I agree here. Numpy should always be first class. As a napari user, I want labeled axes in Napari so I know what axis of my 5-d tensor I am toggling (these tensors _happen_ to be in xarray). I would suggest allowing napari to take an iterable of string labels whose order match the sliders produced by showing an image with dimensionality > 2. 

Later, xarray support could be added such that when an xarray is passed, it auto-populates these labels. If you point me at the relevant place in the codebase, I would be happy to make these contributions. It's a pain point in our workflows (see: https://github.com/spacetx/starfish/blob/master/starfish/display/stack.py#L194).

HagaiHargil:
Hi,

[...]

PR #607, opened 2019-10-20
Basic "play" functionality (animate a dimension)

tlambert03:
# Description
This PR adds a "play" method to `qt_dims.QtDims` to animate changing the slice setpoint (e.g. for a time axis).  I'm sure this was something you were going to implement eventually but I didn't see any discussion yet in the issue tracker...  This was just how I implemented it, so feel free to "freeze" or disregard this PR if there were reasons for holding off on this functionality until later.  Currently there is no button in the GUI Window, just a keybinding (`ctrl-alt-P`), and a couple new methods on `QtDims`.  Similarly, changing playback speed or the axis being animated requires using the method directly.

Some other things that should be addressed before any sort of merge...
- The `QTimer` driving the animation appears to halt on any mouseover event in the main window causing a delay in animation.  That might be fine (might even be the desired compromise for performance), but I thought it was worth mentioning.  I suppose it could run in a different thread to prevent that? 
- currently, only the `QtDims.stop()` method halts the animation.  So if the user changes the number of displayed dims, the "animated" axis may no longer be visible, yet the timer and the corresponding computation continue.  this can create a performance/lag issue.

## Type of change

# References
https://doc.qt.io/qt-5/qtimer.html

# How has this been tested?

- [x] the test suite (`napari._qt.tests.test_qt_viewer.test_play_axis`) for my feature covers playing the 0th axis in a 3D image, asserting that the play method works (i.e. the `dims.point` has changed the expected number of frames after some wait interval), and asserting that the stop method actually works to prevent additional `set_point` calls

## Final checklist:


sofroniewn (commented):
This looks great @tlambert03! Thanks for the contribution - a play button is on our `0.3.0` roadmap #420 but no one was working on it yet, so great timing!

A couple quick comments - using another thread so the QTimer is not blocking might be a good idea, but I can also see reasons why it might be best to just let it play without being able to change anything. How do @royerloic @jni feel about the expected behavior?

If we do want to use another thread @AhmetCanSolak might have ideas on how to do this best too.

I'll also play around with some example datasets and see how it feels too.

Your comment about `QtDims.stop` is also interesting - we might want to hook up an event from the viewer to trigger stopping when the layers data changes. Right now we have a function `viewer._on_layers_change` that gets called when changes like that happen, we might want to make the viewer emit an event `layers_change` that will stop an animation, but again maybe that relates to what expectations we have about interactivity during playing.

If you want to add play buttons in this PR too that would be great - you'll need to look at our contributing guidelines for adding icons, but we can also merge without it and then make a followup PR later if not.

napari/keybindings.py
```diff
@@ -138,3 +138,13 @@ def toggle_ndisplay(viewer):
...
+
+
+@Viewer.bind_key('Control-Alt-P')
+def play(viewer):
+    """Toggle animation on the first axis"""
+    if viewer.window.qt_viewer.dims.is_playing:
+        viewer.window.qt_viewer.dims.stop()
+    else:
+        # TODO: need to allow control over FPS in gui
+        viewer.window.qt_viewer.dims.play_dim(0, 10)
```
sofroniewn: can this use the `last_used` axis instead of being hardcoded to `0` - 
```python
axis = viewer.window.qt_viewer.dims.last_used
if axis is not None:
    viewer.window.qt_viewer.dims.play_dim(axis, 10)
```

napari/_qt/qt_dims.py
```diff
@@ -328,3 +328,40 @@ def focus_down(self):
         else:
             index = (displayed.index(self.last_used) - 1) % len(displayed)
             self.last_used = displayed[index]
+
+    def play_dim(self, axis: int = 0, fps: float = 10):
```
sofroniewn: how do we feel about allowing multiple axes to be playing at the same time? it seems pretty strange to me, in which case we should probably add a `self.stop()` to stop any existing playing inside this function

tlambert03: yeah, I agree it should be only a single axis at a time (at least until some real need otherwise has been demonstrated).  The current code already accomplishes that though, since there is only a single `self.play_timer`... so if play gets called again on another axis, the older `QTimer` gets garbage collected and playback on that axis stops.  Do you think this should be made more explicit?

sofroniewn: ah ok nice I missed that - yeah, if you don't mind, I think including the explicit call to `stop` and a comment saying only one axis can be playing at a time makes it more clear what's going on

tlambert03:
OK, made the changes mentioned above, and I moved the animation timer to it's own `QThread` and I think it works much better (relatively smooth animation even during 3D rotation for instance).  I also added a new `layers_change` emitter to the viewer, call it on the `viewer._on_layers_change`, connected it to the `QtDims.stop()` method where all the other `viewer.layers.events` were getting connected in `QtViewer`.  It works to stop the animation when adding layers or changing dims displays.  Please let me know if I did that correctly, I definitely don't want to mess up any of the lower level model.

@AhmetCanSolak, I realize now that [you were going to take on the play button thing](https://github.com/napari/napari/issues/420#issuecomment-531940494), (sorry!) so please have a look when you get a chance and let me know what you would do differently... In particular, you might have thoughts on the issues below...

Remaining problems:

- If I add a lot of 3D image layers and the per-timepoint rendering burden gets higher, triggering playback can bring the whole thing to a grinding halt, sometimes requiring a force quit.  I don't expect it to keep up with the frame rate (in this case 10 fps was requested) but it should at least fail gracefully.  I guess we might need to debounce the animation timer's call to `set_point` somehow, so that the timer doesn't spam the viewer with requests that it can't keep up with
- I chose to create and destroy the animation QThread each time playback is started and stopped (rather than keeping a thread open the whole time just in case someone plays).  One issue is that if the user quits the program _while_ the animation is running, the QThread doesn't have time to call the `timer.deleteLater` cleanup, a warning is thrown (`WARNING: QObject::killTimer: Timers cannot be stopped from another thread`) and the program exits with a segfault.  I guess we need to hook into the QT [`aboutToQuit()`](https://doc.qt.io/qt-5/qcoreapplication.html#aboutToQuit) signal... but I haven't figured that out yet.

I'll try to take a look at adding icons and play buttons soon...

sofroniewn:
for the `aboutToQuit` you should be able to put the shutdown call inside `closeEvent(self, event)` which is at the bottom of the `qt_main_window.py` file and where we shutdown the console.

Gracefully failing / debouncing the animation timer sounds good.

The `layers_change` event looks good.

tlambert03:
I made some debouncing/performance changes that I think worked pretty well.  The `QtDims` now has a `_play_ready` attribute that gets set to false right before sending a `set_point` change to the `Dims` object ... and then it waits for that to cascade all the way to the `SceneCanvas.event.draw` event before re-enabling `QtDims._play_ready`.  Any frame update requests sent in the meantime by the `AnimationThread` get ignored, (but the desired frame is not lost).  The net effect is to keep up the "effective" requested frame rate, even if the graphics view can't deliver it, without choking the main thread.

The gif below shows the GUI remains responsive to start/stop requests and layer events even with 6 3D stacks "playing" as fast as 1000fps (obviously vispy isn't keeping up, but the gui remains responsive).

ezgif-4-49543dd5c047

unfortunately, I broke my tests, since now it's waiting on a vispy draw event that never occurs in the test and so it appears to advance only a single frame.  I could either mock the set_point function, or just relax the test...

sofroniewn:
That's amazing!! If you can try using `mock` that would be appreciated. You can see how I did it in #629 for an example using the `patch.object`

```python
    @patch.object(visual, '_on_data_change', wraps=visual._on_data_change)
    def test_ndisplay_change(mocked_method, ndisplay=3):
        viewer.dims.ndisplay = ndisplay
        mocked_method.assert_called_once()
```

tlambert03:
OK, I mocked the `Dims.set_point` method to make the test work (and learned about the cool mock `side_effect` parameter in the process :)  The `closeEvent` is also hooked now so it's safe to quit even with the animation running. 

I feel like a little bit of a "heads up" is in order: presumably due to unpredictability in clock timing, at first this test would fail like 5% of the time.  I tuned the parameters (mostly the interval and the buffer allowed on [this line](https://github.com/napari/napari/blob/07acd8e68b0c4428cabb18619f723ed14733b060/napari/_qt/tests/test_qt_viewer.py#L358)), and it just passed ~100 times successfully (and it looks like CI is ok with it)... so I think it's robust now.   but **if `test_qt_viewer.py ` ever fails for anyone in the future, you're all allowed to curse me and ping me to fix it**.

I'm inclined to keep the GUI integration in a separate PR for now.  It still needs a play button on each axis slider (perhaps in combination with the axis labels mentioned in https://github.com/napari/napari/issues/14#issuecomment-546993923), and the user needs a way to control the playback fps (maybe a right click on the play button as is done in Fiji?).  I think it would also be nice to add `first_frame`/`last_frame` parameters (like Fiji) to the play() method, to restrict the playback to a subset of the dimension range.

but I think it's ready to have you guys play with this on a few datasets to get your impressions. (`cmd-alt-P` shortcut to start & stop or `viewer.window.qt_viewer.dims.play_dim(axis=0, fps=30)` in the console)

sofroniewn:
that's great, i'll take a look at this again shortly. thanks for the note about the clock timing, do you want to add it in a comment in the line above here so that we don't forget.
```python
# the 0.5 allows for generous clock jitter...
```

Waiting to add the button is a good idea, I'd like to get this PR in without it and then we can revisit depending on how #14 is going

tlambert03:
as I was starting to implement the first/last frame looping idea here, I realized that there's actually a bug-like issue with the way that I'm incrementing frames here when `layer.scale != 1`.  I started issue #640 for general discussion about scale and how it relates to image indices... but it has implications here too.  Currently, the effective fps is actually `fps/layer.scale[axis]`.  I can definitely change that by taking scale into account when requesting a frame update during animation, but I'll wait to get your thoughts on #640.

sofroniewn:
I will review this properly today, but if you see my comment in https://github.com/napari/napari/issues/640#issuecomment-547528054, I think for now you should rescale `fps*layer.scale[axis]` so that `fps` is accurate, and add a note referencing issue #640 and explaining the rescaling that we can come back to if we need to

sofroniewn:
also @tlambert03 our osx tests seem to be failing - is that because of the clock problem? if it fails consistently on osx we'll want to figure out a solution before merging

tlambert03:
gdi.  yeah, that's the assertion error that happens when the count is wrong.  OK, I'll come up with a better testing strategy and resubmit.

tlambert03:
OK, I relaxed the tests to simply make sure that playing advances the frame roughly the amount of times requested (allowing for +/- 1 frame) which has never failed on me yet.  And I took scale into account so that it plays at the requested fps regardless of the scale.  I also added a `range` parameter to the `play_dim()` method that allows you to restrict the animation to a subset of frames, and added a few tests surrounding that functionality.

in addition to having you test the basic functionality, I would also welcome any comments on the naming of public facing methods or variables.  (e.g.  we could change `play_dim` to something else if you want)

napari/_qt/qt_dims.py
```diff
@@ -328,3 +329,128 @@ def focus_down(self):
         else:
             index = (displayed.index(self.last_used) - 1) % len(displayed)
             self.last_used = displayed[index]
+
+    def play_dim(self, axis: int = 0, fps: float = 10, range=None):
```
sofroniewn: I would rather this be named `play` - its already located on the `dims` view, and matches nicer with `stop`

napari/_qt/qt_dims.py
```diff
@@ -328,3 +329,128 @@ def focus_down(self):
...
+        """
+        Animate (play) axis
+
+        Parameters
+        ----------
+        axis: int
+            Index of axis to play
+        fps: float
```
sofroniewn, line 349: What do you think about if `fps < 0` can we move through the axis in reverse, and if `fps=0` we call `stop`?

napari/_qt/qt_dims.py
```diff
@@ -328,3 +329,128 @@ def focus_down(self):
...
+        ----------
+        IndexError
+            If ``axis`` requested is out of the range of the dims
+        IndexError
+            If ``range`` is provided and out of the range of the dims
+        ValueError
+            If ``range`` is provided and range[0] >= range[1]
+        """
+        # TODO: No access in the GUI yet. Just keybinding.
+        if axis >= len(self.dims.range):
```
sofroniewn: I think we should also check if `self._displayed_sliders[axis]` is `True`, as these are the axes that actually correspond to sliders, and only proceed if it is `True` and otherwise just stop everything.

Reason is right now I can "play" an axis for a 2D image, nothing bad happens but `is_playing` becomes True and we'll have unnecessary loops in the background

napari/_qt/qt_dims.py
```diff
@@ -328,3 +329,128 @@ def focus_down(self):
...
+    @property
+    def is_playing(self):
+        """Returns True if any axis is currently animated"""
+        return (
+            hasattr(self, 'animation_thread')
+            # this is repetive, since we delete the thread each time, but safer
+            and self.animation_thread.isRunning()
+        )
+
+    def _set_frame(self, axis, frame):
```
sofroniewn, line 413: I love how this is working, but one interaction pattern I'm curious if you have played with is what happens when you are playing an axis and you click in the slider for that axis. It looks like right now the animation actually stops. If you click again then it seems to start. The ideal interactivity would be if you just jumped to the frame where the mouse was clicked and then you continued to animate. Do you think this is possible?

napari/_qt/tests/test_qt_viewer.py
```diff
@@ -316,3 +317,84 @@ def test_qt_viewer_data_integrity(qtbot, dtype):
...
+    axis, interval, nframes = 0, 50, 5
+    view.dims.counter = 0
+    view.dims.play_dim(axis, 1000 / interval)
+    # the 0.5 allows for a little clock jitter...
+    qtbot.wait(interval * (nframes + 0.5))
+    view.dims.stop()
+    # really don't want this to fail due to timing, so we're only making sure it
+    # advanced about the right amount... precise timing will depend on the machine
+    c = int(view.dims.counter)
+    assert c >= nframes - 1
```
sofroniewn: unfortunately I think this has still failed for one of our osx builds but passed for the other, indicating that there is still some non-determinancy here. Maybe you have to weaken the test even further? Sorry about that, we can always try and improve the testing after this goes in

sofroniewn (approved):
Ok this looks great, I've left a few comments that I'd like addressed before we merge, probably the most critical ones are around clicking on the slider while the animation is playing and the test that still seems to sometimes fail, but I am approving this now, and hopefully we can then merge soon.

This functionality is really so cool, I love using it already, and can't wait to get it out to everyone else!!

napari/_qt/qt_dims.py
tlambert03: yeah that should be possible.  lemme play with it.

napari/_qt/tests/test_qt_viewer.py
tlambert03: yeah, I'll weaken further.  I don't think it's too big of a compromise at this point.  There may come a day when we need to be able to assert some very specific platform-independent timing, but for now it's probably enough to make sure that the play method advances at least a few frames, and the stop method stops it.  If I give it window of +/- 3 frames, i doubt the tests will ever fail, which is obviously more important for now, and we can leave cross-machine timing precision for another PR (if its even possible).

tlambert03:
I think this is working pretty well now.  The forward/backwards fps works, as well as restricting it to a particular range of frames.  I also added a `pingpong` argument to the `play` method that will loop the requested frames back and forth when `True`.  The tests are also passing.

I added a connection to the `dims.events.axis()` event to update the current frame of the animation, so regardless of whether someone sets the current frame using the slider, or directly uses the `set_point` method on the `Dims` object, the animation will update the current frame.=
_Note:_ if you simply click on the slider somewhere else while it's playing, it doesn't jump directly to that frame, but rather moves one frame in that direction.  If you are able to actually click and drag the slider, then it will resume playing wherever you drop it.  That is consistent with the regular (non-animated) behavior... and while I think most people are used to a slider behaving that way, it feels a bit weirder when the slider is actually animated.  If we want the slider to fully jump to the point that has been clicked, I think that should be addressed in a separate PR, but the current animation code should follow suit if we make that change elsewhere. 

We had discussed the possibility of using @AhmetCanSolak's closure API (#650) under the hood to drive the animation.  After thinking about it, I think it would be _possible_, but not necessarily the most elegant way to achieve the current functionality.  While it would be nice to have one API for everything, the `AnimationThread` needs to have a bit more cross-thread signals/slots to stay in synch with events going on elsewhere.  If you like, I can try to make a closure that would capture all the logic in [AnimationThread](https://github.com/napari/napari/blob/0c3f28766eaf37f600306e19c77f74c4a3129f67/napari/_qt/qt_dims.py#L418), and then you can decide which style you like better.

napari/_qt/qt_dims.py
tlambert03: changed

napari/_qt/qt_dims.py
tlambert03: fixed

napari/_qt/qt_dims.py
tlambert03: see https://github.com/napari/napari/pull/607#issuecomment-549182306  It doesn't jump straight to the frame... but it doesn't do that even when the animation is not playing.  So if we want to jump to the frame, we should fix that elsewhere. but if we do, this will also have that behavior.

sofroniewn:
This sounds amazing @tlambert03. I think at this point @jni should give it one more review and then we can merge.

I don't it's necessary to switch over to the thread style of #650, we can always revisit this with a refactoring PR, similarly we can revisit the behavior of clicking the slider in a followup.

napari/_qt/qt_dims.py
```diff
@@ -328,3 +330,174 @@ def focus_down(self):
...
+
+        if axis >= len(self.dims.range):
+            raise IndexError('axis argument out of range')
+        # we want to avoid playing a dimension that does not have a slider
+        # (like X or Y, or a third dimension in volume view.)
+        if not self._displayed_sliders[axis]:
+            return
+
+        self.animation_thread = AnimationThread(
+            self.dims, axis, fps, range, pingpong
```
jni: Let's not shadow the built-in `range`. I'm also not in love with `pingpong`. Perhaps we can have `playback_mode={'once', 'loop', 'loop_back_and_forth'}`? (Once being the default.)

napari/_qt/qt_dims.py
```diff
@@ -328,3 +330,174 @@ def focus_down(self):
         else:
             index = (displayed.index(self.last_used) - 1) % len(displayed)
             self.last_used = displayed[index]
+
+    def play(self, axis: int = 0, fps: float = 10, range=None, pingpong=False):
```
jni: Love the type annotations!

napari/_qt/qt_dims.py
```diff
@@ -328,3 +330,174 @@ def focus_down(self):
...
+        )
+        # when the thread timer increments, update the current frame
+        self.animation_thread.incremented.connect(self._set_frame)
+        self.animation_thread.start()
+
+    def stop(self):
+        """Stop axis animation"""
+        if self.is_playing:
+            self.animation_thread.quit()
+            self.animation_thread.wait()
```
jni: I think `animation_thread` should be "private" (ie leading underscore)

napari/_qt/qt_dims.py
```diff
@@ -328,3 +330,174 @@ def focus_down(self):
...
+            # this is repetive, since we delete the thread each time, but safer
+            and self.animation_thread.isRunning()
+        )
+
+    def _set_frame(self, axis, frame):
+        """Safely tries to set `axis` to the requested `point`
+        This function is debounced: if the previous frame has not yet drawn to
+        the canvas, it will simply do nothing.  If the timer plays faster than
+        the canvas can draw, this will drop the intermediate frames, keeping
+        the effective frame rate constant even if the canvas cannot keep up.
```
jni, line 419: Very nice comment!

napari/_qt/qt_dims.py
```diff
@@ -328,3 +330,174 @@ def focus_down(self):
...
+        """Returns True if any axis is currently animated"""
+        return (
+            hasattr(self, 'animation_thread')
+            # this is repetive, since we delete the thread each time, but safer
+            and self.animation_thread.isRunning()
+        )
+
+    def _set_frame(self, axis, frame):
+        """Safely tries to set `axis` to the requested `point`
+        This function is debounced: if the previous frame has not yet drawn to
```
jni, line 416:
```suggestion

        This function is debounced: if the previous frame has not yet drawn to
```

napari/_qt/qt_dims.py
```diff
@@ -328,3 +330,174 @@ def focus_down(self):
...
+    def is_playing(self):
+        """Returns True if any axis is currently animated"""
+        return (
+            hasattr(self, 'animation_thread')
+            # this is repetive, since we delete the thread each time, but safer
+            and self.animation_thread.isRunning()
+        )
+
+    def _set_frame(self, axis, frame):
+        """Safely tries to set `axis` to the requested `point`
```
jni:
```suggestion
        """Safely tries to set `axis` to the requested `point`.
```

napari/_qt/qt_dims.py
```diff
@@ -328,3 +330,174 @@ def focus_down(self):
...
+        """Stop axis animation"""
+        if self.is_playing:
+            self.animation_thread.quit()
+            self.animation_thread.wait()
+            del self.animation_thread
+            self.enable_play()
+
+    @property
+    def is_playing(self):
+        """Returns True if any axis is currently animated"""
```
jni:
```suggestion
        """Return True if any axis is currently animated"""
```

napari/_qt/qt_dims.py
```diff
@@ -328,3 +330,174 @@ def focus_down(self):
...
+
+    def enable_play(self, *args):
+        # this is mostly here to connect to the main SceneCanvas.events.draw
+        # event in the qt_viewer
+        self._play_ready = True
+
+
+class AnimationThread(QThread):
+    """A thread to keep the animation timer independent of the main event loop.
+    This prevents mouseovers and other events from causing the animation lag"""
```
jni:
```suggestion

    This prevents mouseovers and other events from causing animation lag
    """
```

napari/_qt/tests/test_qt_viewer.py
```diff
@@ -316,3 +316,83 @@ def test_qt_viewer_data_integrity(qtbot, dtype):
...
+    data = np.random.random((10, 10, 15))
+    viewer.add_image(data)
+
+    def increment(e):
+        view.dims.counter += 1
+        # if we don't "enable play" again, view.dims won't request a new frame
+        view.dims._play_ready = True
+
+    view.dims.dims.events.axis.connect(increment)
+    view.dims.counter = 0
```
jni: Could you set this before the definition of `increment`? Having it under it breaks my internal parser. =P

napari/tests/test_viewer.py
```diff
@@ -28,6 +28,12 @@ def test_viewer(qtbot):
     # Run all class keybindings
     for func in viewer.class_keymap.values():
         func(viewer)
+        # the `play` keybinding calls QtDims.play_dim(), which then creates a new QThread.
+        # we must then run the keybinding a second time, which will call QtDims.stop(),
+        # otherwise the thread will be killed at the end of the test without cleanup,
+        # causing a segmentation fault.  (though the tests still pass)
+        if func.__name__ == 'play':
+            func(viewer)
```
jni, line 37: This feels kinda messy and brittle... Could there perhaps be a context manager that cleanly deletes running threads? I can imagine more and more keybindings needing special treatment like this, which would be a nightmare without a unified way to handle them...

But, I'm reasonably ok with this workaround. Just want to make sure we've explored all the options.

jni (dismissed):
@tlambert03 very cool! A very clean PR. I've left some mostly easy comments to address I hope!

napari/_qt/qt_dims.py
tlambert03: you know, it's funny, I almost avoided shadowing `range` there, and then saw a [few lines ](https://github.com/napari/napari/blob/0c3f28766eaf37f600306e19c77f74c4a3129f67/napari/_qt/qt_dims.py#L164-L170) in the same file that also shadowed it and thought, "hey I guess if it's only in a small scope, maybe it is a nice word to use..." and so I followed that lead :)
I almost used `boomerang` in honor of australia, instead of `pingpong`... I guess I was feeling whimsical.  `playback_mode` is fine (soo predictable 😉).  Only thing I would add is that I prefer `'loop'` to be the default rather than `'once'`... @sofroniewn ?

jni: Oh no... What have I become...? (@stefanv often shoots me down in my whimsical moments, now here I am doing the same... =P)

I'm not strongly opposed to loop being the default, your call!

jni: (With Napari being in the middle of the Pacific, maybe we should go for `enso=`, ie El Niño Southern Oscillation. The meaning *might* be obvious to our climate science community... =P)

sofroniewn: Loop as default is fine for me

napari/tests/test_viewer.py
tlambert03: looked into it for a while, and while I agree this is ugly, I didn't like the alternatives I found.
The only way I could figure out to "find all QThreads" is to start at some high level object (like `QtWidgets.QApplication.instance().topLevelWidgets()`) and recurse through all `children()` looking for instances of `QThreads` to kill.  That might work, but may also eventually lead to unintended side effects of its own if we go killing all threads with reckless abandon.  And if we only try to kill certain threads (like `AnimationThread`), we're right back to checking for something specific (like `func.__name__ == 'play'`).  I agree it seems like a handy trick to have though, so let me know if you'd like me to push  harder on it.

tlambert03:
OK, I think I need some help on testing this... at this point I feel like I'm breaking the principles of testing and simply testing the performance of the QTimer driving the thread... which shouldn't be the point.  All I want to do is (roughly) assert that the anticipated number of frames should have passed. The timing of the OSX test varies _wildly_.  I've tried account for lag in starting the QThread as much as possible by hooking directly into the `AnimationThread.QTimer` signals where possible:

https://github.com/napari/napari/blob/c5249ed845564536b8ab34f1c974e1ba6f851803/napari/_qt/tests/test_qt_viewer.py#L339-L351

... but CI OSX is _still_ off sometimes by as much as 4-5 frames.  There seems to be a big variability of thread timing on Cirrus.  I can continue to relax the tests, but it starts to feel like the test itself is useless.  The tests _never_ fail on my local computer (a mac), and they never fail on the linux CI... but the OSX CI is all over the place.

any suggestions for how to redesign this test are very welcome.  It's more or less the only thing this PR is waiting on

sofroniewn:
@tlambert03 just gave this a try - everything is working great!

One question though - when you click in the slider during an animation it only skips backward / forward one point instead of jumping to that point. You can click and drag on the slider to get back to that point, but I feel like the functionality would be more useful if it just automatically jumped. Was there a reason you made it just skip one instead of jump?

One of the usage modes I am imagining is that you're playing and then you see something cool and you want to just click the slider and have it automatically jump back to that point so you can watch it again. You can't quite do that so easily right now.

I also seem to remember seeing a comment asking for help thinking about the tests, but I can't find it so you may have deleted it. Sorry that's such a pain right now. My recommendation is to relax so we can merge - maybe leave the more strict test in there as a comment though so we can always revisit in the future if something breaks.

tlambert03:
Thanks a lot @sofroniewn 

> One question though - when you click in the slider during an animation it only skips backward / forward one point instead of jumping to that point. You can click and drag on the slider to get back to that point, but I feel like the functionality would be more useful if it just automatically jumped. Was there a reason you made it just skip one instead of jump?

see the "_note_" in the second paragraph of my comment above: https://github.com/napari/napari/pull/607#issuecomment-549182306
I totally agree: that is probably what the user expects/wants when clicking on the slider.  But that's actually not what currently happens regardless of whether the slider is animated or not (i.e. that's just how the program works at the moment).  Since it's not specific to play behavior, I was thinking it might be cleaner to address that behavior in a separate PR, and I'm happy to open it immediately.  I guess the main question is whether we want to keep the current behavior (clicking on slider moves one frame) when the slider is _not_ animated, but change the behavior (clicking on slider moves to specified frame) when `QtDims.is_playing == True`, or whether you want it changed globally. 

yeah, I deleted the request for help; I thought about a new approach and didn't want anyone wasting their time.  The new approach is working a bit better, but still failing.  I'm determined to get it working and will post when ready for final review.

sofroniewn:
Ah - totally miss understood that comment about the current behavior - I got it now. I actually thought the current behavior was to jump, not just increment left / right by one, even with no animation. I completely agree this should be a separate PR, and I'd wait until this one merges to open it. I think it would be nice if jumped for both non-animation and animation usage.

Good luck with the tests. Let us know when you are satisfied with them and then I will merge

napari/_qt/tests/test_qt_play.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+# A lot of this code looks messy and unnecessary (and some of it may be!)
+# most of the weird bits are an attempt to circumvent thread timing
+# non-determinism on Cirrus CI OSX tests.
+# see https://github.com/napari/napari/pull/607 for more info
+@contextmanager
+def make_thread(
+    qtbot, nframes=8, fps=20, frame_range=None, playback_mode='loop'
+):
```
AhmetCanSolak, line 20: I would love you to revisit after #650 . Possibly will make your life easier @tlambert03

tlambert03:
I think the tests are stable now. I re-triggered the cirrus tests a few more times on my account and they work (excluding windows).  In case anyone is curious, I focused on testing `qt_dims.AnimationThread` in isolation, in particular the `advance()` method which holds most of the logic...  with just a couple loose tests to make sure that QtDims.play() triggers and stops the thread appropriately.  I'm happy with the functionality and tests at this point and leave it to your review.

@jni may want to comment again on https://github.com/napari/napari/pull/607#discussion_r341877876

sorry everyone for the many commits and emails this PR probably caused!

_things to do in another PR:_
- add play buttons to GUI sliders (related to #644)... will also need a way to control options like fps and range
- change slider-clicking behavior

sofroniewn:
This looks finished to me @tlambert03 - I think that seems like a good strategy on the tests.

I will leave this until end of day today in case @jni wants to weigh in once more and resolve his requested changes, but judging from his comment I think he trusts this is the right direction, and I will merge at that point

Issue #69, opened 2018-11-21
Intensities in are normalized by values in first frame


batson:
## 🐛 Bug

If the first slice of a z-stack is all zeroes, then subsequent slices appear all white. I believe that intensities are being normalized to the range of the first slice.

## To Reproduce

```
from napari_gui.elements import Window

img = np.random.randn(100, 100, 2)
img[:,:,0] = 0
# img[:,:,0] = img[:,:,0]*0.1
# img[:,:,1] = 0

win = Window()
viewer = win.add_viewer()
layer = viewer.add_image(img, dict(itype='mono'))
layer.cmap = 'grays'
win.show()
```

## Expected behavior

Intensity scaling should be done for the whole stack at once (or perhaps as a per-slice option), not based on the first slice.

## Environment

 - napari-gui Version (e.g., 1.0): 0.0.4+14.g9a08159
 - OS X
 - Python version: 3.6.4

kne42:
Thanks for reporting this bug @batson! Reproduced on [`9a08159`](https://github.com/Napari/napari-gui/tree/9a081595b3d51ed2e85e5119b58e1b9137b6bb77).

Summary of bug:
- `clim` is defaulted to `'auto'`
- first slice is prepared for texture upload
- `clim` is resolved from `'auto'` to the `min, max` of the first slice
- subsequent slices have this `clim` carried over (in this case `0.0, 0.0`)

We can fix this by intercepting `clim` before `vispy` gets to it and checking if `'auto'` is an option, then setting it to `np.min(image), np.max(image)`. This should additionally be reset upon `image` change.

@AhmetCanSolak is working on a PR to integrate `clim` and `gamma` into the sliders. If it's close enough to being done, we can shove this onto it, otherwise this _should_ be a simple enough PR to get done in the next few days.

sofroniewn:
This bug has now been fixed with #85. We now calculate and set clims using the max and min of the whole image volume on loading.

Issue #570, opened 2019-10-04
contrast limits range equal to contras limits on setting


sofroniewn:
## 🐛 Bug

If you explicitly set the contrast limits as a keyword argument to the layers list right now the range of the contrast limits slider gets set to the same value, which is quite annoying if you then need to adjust the sliders to be more extreme. I suggest that we set the range to be 10% smaller and 10% larger to give you a little extra room to adjust things.

This change will only affect the display and will be backwards compatible.

jni:
@sofroniewn imho we should be quite a bit smarter about contrast limits. For one, I think for uint8 images the limits should always be set to 0-255 and the image should not be introspected in any way. In this case, adding 10% margins would be totally nonsensical.

Another idea would be something like `bit_depth` which could be set to e.g. 12 to then automatically adjust the sliders to 0-4095 without needing to type those numbers in.

jni:
Also, I would not call this a bug at all, but a feature request. =)

berl:
just found this issue and wanted to chime in:
the current space in the clims slider is pretty finicky for my data because (like a lot of fluorescence imaging) a ton of our information is down in the low intensity values. With the current 2-dots-and-bar slider, I don't have the flexibility to interactively adjust the low end without the 2 dots overlapping and becoming useless.   

I don't have a great solution here without going to the full zoomable histogram+slider widget similar to what's in pyqtgraph or micromanager

jni:
In Quicktime, you can hard-press (or long press) when scrubbing on the movie position to "expand" the slider and have movement of the slider translate to finer movement of the time. We should definitely steal this idea.

sofroniewn:
The larger pop-up slider feature and editable range labels has now been added in #837 so closing this issue

PR #610, opened 2019-10-21
Add gamma slider

tlambert03:
# Description
This PR adds a functional gamma slider to the Image layer controls.  I know this has been discussed for a while, e.g. in #2, and it was mentioned that @AhmetCanSolak might be working on it [here](https://github.com/napari/napari/issues/69#issuecomment-440771272), but I didn't see any recent activity, so here's an implementation that works for me, for you to consider.

gif

besides just adding sliders and the appropriate fields to the image model, the way I chose to actually apply the gamma to the vispy image was to instantiate a new `vispy.Colormap` (with the gamma already applied) whenever `layer.gamma != 1`.  This seems to work well and was pretty performant in my tests

https://github.com/napari/napari/blob/ac25a0d41815b83d8e0eef7f1e30d96d51c570c3/napari/_vispy/vispy_image_layer.py#L86-L93

current problems:
-  This strategy doesn't currently work with the `TransFire` and `TransGrays` colormaps.  However, those two colormaps actually break the `vispy.Colormap` API slightly and I think the change should be made to those colormaps rather than this PR. See #611
- `QtSurfaceControl` tests are currently failing because I added gamma only to the image layer model (and not yet the surface model) but I added the gamma event to `QtBaseImageControls`.  So the surface layer is trying to connect to a gamma change event that does not exist on the `Surface` model.  This is easily changed, but I just wanted to make sure you would want this on the surface layer as well... or whether I should move the logic up from `QtBaseImageControls` to `QtImageControls`, and leave adding it to surfaces for some other time.

## Type of change

# How has this been tested?
- [x] still need to write tests
- [x] all existing tests pass with my change (not yet... see note above)

## Final checklist:


sofroniewn:
This looks great to me. I think @royerloic will be happy. He may have had a long-term vision of doing more of the gamma correction on the graphics card, and it maybe being more of a global parameter on the viewer rather than a layer specific parameter - so maybe he can weigh too.

I'd also extend support to the `QtSurfaceControl` layer too - it could be useful there too. You'll also need to add the `gamma` parameter to the `viewer.add_image` and `napari.view_image` methods too

sofroniewn:
for the broken colormaps, agreed they should be fixed outside of this PR. We might want to catch the error and reset gamma to one for those colormaps for now though

tlambert03:
Thanks.  OK, I'll wait for @royerloic or others to weigh in on this implementation before working on it too much more. (e.g. adding it to the QtSurfaceControl layer or `add_image/view_image` methods.

royerloic:
This is great!
Would it be possible @sofroniewn to resurrect the global slider and use it for gamma?
We could still have a gamma per channel and a global gamma (that gets premultiplied to
the gamma per channel). To explain, @tlambert03,  the motivation for a global gamma is that right now, if you open 10 layers and they are all dim, you have to adjust each layer, and if you wish to compare these images you have to set all the sliders in the exact same way which is extremely hard if not impossible. We discussed with the rest of the napari team how to have global contrast limits, 
but that turns out to be ill defined in general. Gamma is perhaps a much better candidate because it is a single number, and in most cases is enough to adjust the display range. This could also be introduced in a separate PR.

Regenerating the lookup table is a nice trick, it is indeed important that performance not suffer when gamma==1, and it is great if it does not suffer even for gamma!=1. I hope that long term we can add gamma to the image visual shader (as well as contrast limits) -- which would be the right way of doing it --  albeit a bit more complex. In the mean time that works for me!

jni:
I'd prefer not having a global slider, but rather (and I realise this is more complicated to implement), keep adjustments active in the widgets box that make sense for all selected layers (and this would include gamma).

sofroniewn:
Thanks @jni and @royerloic for weighing in - seems like the most sensible thing for this PR now is to press on with gamma on each Image and Surface layer. We can then think more about how to enable more global like contrast control (which I know @berl has also requested too). I lean towards the solution proposed by @jni where layers of the same type can be selected and edited together, but that is new functionality that can come in a new PR.

I'd say press on now @tlambert03 with adding gamma to the QtSurfaceControl layer and add_image / view_image methods. Thanks!

royerloic:
> I'd prefer not having a global slider, but rather (and I realise this is more complicated to implement), keep adjustments active in the widgets box that make sense for all selected layers (and this would include gamma).

I think that is a good approach, I am just wondering if a bigger slider which is easier to control might be better for something so fundamental. I think usage will tell us if we  need something bigger, or not...

jni:
@royerloic 

> I am just wondering if a bigger slider which is easier to control might be better for something so fundamental.

Note [this comment](https://github.com/napari/napari/issues/570#issuecomment-544385722) from #570: you could expand the slider upon long-press, to allow for finer control. This is a very useful interaction in general.

tlambert03:
- Added gamma to surface layers and controls.
- Added `gamma` parameter to all appropriate functions (`add_image`, `add_multichannel`, `add_surface`, `view_image`, etc...)
- Updated docstrings
- added gamma tests for image and surface layers

with #622 merged, this is ready for your evaluation... short of the "global slider" idea mentioned above, which I think is a good idea, but might be best for a seperate PR.

sofroniewn (approved):
This looks great. I just tried it out and it works perfectly. I was at first unsure if the slider should go to a value larger than 2, but after trying it on a bunch of different images I think it's a pretty good max value, and it also means that the default value is in the middle of the range which I like.

Thanks for adding tests too. They've all passed. I will merge shortly.

tlambert03:
i had the same internal debate about the max value 😄.  but the aesthetic value of having the default be in the middle (and the general usability of max=2) won me over.

Issue #612, opened 2019-10-21
float32 image data getting clobbered after changing ndisplay


tlambert03:
## 🐛 Bug

While viewing 32 bit arrays, I saw some unexpected behavior, and I'm curious whether this is something you all have observed before or indeed a bug.  The gist is that when I add a 3D `float32` array to the viewer, then either press the "toggle displayed dims" button or directly change `viewer.dims.ndisplay`, the image appears to disappear, and it doesn't come back when going back to a 2D view.  Looking directly at the data in `viewer.layers[0].data` shows that it's all been permanently changed to extremely small numbers.  This is not the case for any other data type. 

It has something to do with vispy... and one way I could fix the problem is just making a copy of the `_data_view` in the `VispyImageLayer._on_data_change` method prior to calling `self.node.set_data` ... for instance by adding 
```python
 if dtype == 'float32':
    data = data.copy()
```
to the else statement in following section: 
https://github.com/napari/napari/blob/85e6bd46a2b1be230cceef429316e41405aa6356/napari/_vispy/vispy_image_layer.py#L70-L74

I feel like I'm spamming you recently with PRs, so here's a test that you can add to `test_qt_viewer.py` to try it out locally (for me, it only fails on the `float32` parameter). If you see the same behavior and you'd rather me submit a PR, let me know.  (and I'd also be curious whether you can think of a better fix than making a copy of the array) 

```python
import pytest

@pytest.mark.parametrize(
    "dtype", ['int8', 'uint8', 'int16', 'uint16', 'float32']
)
def test_qt_viewer_data_integrity(qtbot, dtype):
    """Test that the viewer doesn't change the underlying array."""

    image = np.random.rand(10, 32, 32)
    image *= 200 if dtype.endswith('8') else 2 ** 14
    image = image.astype(dtype)
    imean = image.mean()

    viewer = ViewerModel()
    view = QtViewer(viewer)
    qtbot.addWidget(view)

    viewer.add_image(image.copy())
    datamean = viewer.layers[0].data.mean()
    assert datamean == imean
    # toggle dimensions
    viewer.dims.ndisplay = 3
    datamean = viewer.layers[0].data.mean()
    assert datamean == imean
    # back to 2D
    viewer.dims.ndisplay = 2
    datamean = viewer.layers[0].data.mean()
    assert datamean == imean
```

sofroniewn:
[...]

Issue #575, opened 2019-10-08
3D rendering seems to overwrite data array


li-li-github:
## To Reproduce

1. Display a 3D data on napari.
2. Adjust contrast.
3. Contrast does not come back even when you tune the contrast bar back to original position.

## Environment

 - napari Version (e.g., 1.0): 0.2.0
 - OS (e.g., Linux): Mac OS 10.14.6
 - Python version: 3.6.7
 - Any other relevant information:

jni:
@li-li-github I can't reproduce this on my machine. At any rate I think and hope that the error might be in the display, not in modifying the data.

Could you provide a complete example with data? In the code below, I can adjust the contrast back and forth perfectly in the 3D display:

```python
import numpy as np
from skimage import data
from skimage.util import random_noise
import napari

blobs_raw = np.stack([data.binary_blobs(length=64, n_dim=3, volume_fraction=f)
                      for f in np.linspace(0.05, 0.5, 10)])
blobs_sp = random_noise(blobs_raw, mode='s&p')
blobs_g = random_noise(blobs_sp, mode='gaussian')
blobs = random_noise(blobs_g, mode='poisson')
with napari.gui_qt():
    viewer = napari.view_image(blobs)
    viewer.dims.ndisplay = 3
```

li-li-github:
Hi @jni , thank you for the prompt reply.
I found that it doesn't work properly when datatype is `np.float32`.
I have tested this with other np.floats and found only float32 doesn't work properly.

sofroniewn:
@jni @li-li-github I am able to reproduce this bug by modifying the code from @jni to be
`viewer = napari.view_image(blobs.astype(np.float32))`
and then moving the left clim handle all the way to the right - my screen becomes black, and the data will never reappear. It appears to be a divide by 0 error within vispy that vispy can never recover from. Here is the error message I see
```python
/Users/nicholassofroniew/Github/vispy/vispy/visuals/volume.py:483: RuntimeWarning: overflow encountered in true_divide
  vol /= self._clim[1] - self._clim[0]
```

Ultimately this will be fixed by a PR to vispy which we are happy to make, in the mean time we can look into some workarounds / fixes on our end, and I recommend trying to avoid shrinking the clim range to be too small.

Thanks for catching this @li-li-github!!

jni:
[...]

PR #613, opened 2019-10-22
fix for data array overwriting during 3D rendering

tlambert03:
a fix for #575 and #612, and a test.
explanation [here](https://github.com/napari/napari/issues/575#issuecomment-544753893)

This does not solve the repeated calls to `VispyImageLayer._on_data_change`, but prevents the diminishing data problem.

### DIFF
--- a/napari/_qt/tests/test_qt_viewer.py
+++ b/napari/_qt/tests/test_qt_viewer.py
@@ -1,4 +1,5 @@
 import numpy as np
+import pytest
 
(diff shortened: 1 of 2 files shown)


jni (approved):
Looks great! I'll merge when the CI passes. (Our Windows CI is broken, fyi, so don't worry about it, unless you want to go down that rabbit hole, in which case we would be super grateful. =P But unrelated to this PR.)

Issue #578, opened 2019-10-09
Performance with invisible layers


jni:
## 🐛 Bug

When many layers are added to napari, many operations are extremely sluggish, *even if all but one* of the layers are set to invisible. We should instead immediately render only visible layers and turn off most/all data handling for layers with visibility 0.

Here are some things that are very slow when many (~100) layers are added to the display:

- zoom
- changing ndisplay (this is incredibly slow)
- changing the visibility of a layer

In general, it would be great if we didn't use `set_view_slice` at all on invisible layers, so that if the layers are referring to data on disk (e.g. hdf5 array), it is not loaded at all unless it is visible. Additionally, the effect of changing ndisplay should be minimal if the layer is not visible.

sofroniewn:
Ok - makes sense, but we just need to make sure things don't get in weird states where you expect things to be one way, but because of the visible stuff they aren't. Adding in the benchmarking and profiling stuff will help us figure out which are the bits that are really slow

if zoom is slow it's really a vispy thing that would be harder to fix on the napari side

eburling:
I'm experiencing this in a multiplex immunofluorescence imaging use case illustrated [here](https://www.loom.com/share/8af375e297274189bbbb7d2bd34596ea). For reference, each layer is a multiresolution zarr array being read in by dask, e.g.:

```python
cache = Cache(2e9)  # Leverage two gigabytes of memory
cache.register()

viewer = napari.Viewer()

base = imread('stain1.tif')
pyramid = list(pyramid_gaussian(base, downscale=2, max_layer=3, multichannel=False))
root = zarr.open_group('pyramid.zarr', mode='a')

for i in range(0, len(pyramid)):
    shape = pyramid[i].shape
    z = root.create_dataset(str(i), shape=shape, chunks=(300, 300), dtype='float')
    z[:] = pyramid[i]
    
pyramid = [da.from_zarr(file_name + '/' + str(i)) for i in range(len(pyramid))]

viewer.add_image(pyramid, name='stain1', visible=False, opacity=0.5, is_pyramid=True)
```
When I am browsing at the highest resolution of the pyramid (8059 x 9454 pixels) I notice that the layer index at the lower left of the viewer iterates over each layer, presumably loading the chunks of the invisible layers. At most, I expect to be visualizing 5 layers at any one time, with manual switching between layer combinations, so lazy loading at this level would be great.

jni:
[...]

PR #616, opened 2019-10-24
Open dialog

### LINKED ISSUE #604: customizing fileIO dialog
## 🚀 Feature
Right now we just use the native fileIO dialog as bellow
image

but can can potentially customize our own fileIO dialog that would allow us to include extra toggle buttons, and drop-downs. I think this will be increasingly important as we try and deal with more complex layer and file types, and I'd rather us just have one pop-up menu where it all happens, instead of multiple pop-up menus, or many different open functions inside the file menu. The goal would be to unify as much as possible.

I imagine that we'd have one drop-down for layer type, and then depending on what layer type you had selected you would have different checkboxes / dropdowns available to you. We could expose all the keyword arguments for the layer pre-populated with their defaults, which I sort of like as people at least get to see them too, even if they don't edit them.

A common problem I saw introducing some biologist to napari the other day was they had jpegs of data that were rgb but they wanted to add using `add_multichannel(data, axis=2)` so they got the channels separated. Or they had multichannel tiff of different color channels which they wanted to add using `add_multichannel(data, axis=0)` but there was no way for them to do that from the fileOpen dialog, and it was quite cumbersome to do the conversion once added.

Thoughts @jni @royerloic?

sofroniewn:
# Description
This PR will close #604 by adding support for a custom dialog where you can set keyword arguments before opening a native file dialog to select some files.

I'm very curious what @jni, @royerloic, and others make of this and what keyword arguments should be exposed. See the gif here for an example using a couple of parameters from the image layer:

fileopen_dialog

One argument might be we should expose every single keyword argument for the layer creation, another is we expose the ones we think are most important, or we just expose the ones that are essential - in so far as you cannot change them after the layer has been added - for example if you want your image to add as `multichannel` or you want to say `rgb` or `is_pyramid` is `True`.

I'm not sure the best option yet. Also you'll see in the dialog the tabs for the different layers which will change which of the keyword arguments are available and what layer gets added. Right now we only support adding images.

I'd like to get #601 in before we finish this one up too as it will make some of the path stuff simpler and ultimately more unified across our different layer types, but I wanted to get this out here now so people can see how we're thinking about it.

## Type of change

# How has this been tested?

## Final checklist:


jni:
Ah, I see the double entendre of the PR title... Nicely done. =P

I'd like to keep a simple "File -> Open image(s)" and "File -> Open folder" that immediately goes to the filesystem and uses default settings. For this menu, I would call it, "File -> Advanced open".

I think exposing every single kwarg is a good approximation for now. We can start from that and trim the ones we think are superfluous.

Other than that, I think the implementation is basically correct.

From an aesthetic perspective, I would like it if the selected layer tab was "continuous" with the panel underneath.

I also think the file selection should be the very first thing, not way at the bottom.

sofroniewn:
This is still very much in progress, but I was able to make some updates. There is now one `QLineEdit` at the top which you can type into. pressing enter will move the text into the `QListWidget` below. You can select and delete lines from the `QListWidget` by then pressing delete or clicking the `remove` button. You cannot rearrange the files, though maybe this is easily enabled with some flag.

Rather annoyingly right now when I have lots of files or files with very long names the `QScrollBars`
appear on the "inside" of the `QListWidget` and so cover-up the bottom file name entirely, which is not so good. It wasn't obvious to me how to get them on the "outside", so if anyone has any ideas that would be great.

You can also add files by clicking the select folder or select files buttons, these will add and not replace whatever is already there. This can be easily changed to be replace if people prefer.

Below is the list of tabs for all the layer types, right now the only one enabled is the `Image` layer, I'll figure out how to disable / show not implemented warnings for the others, but I want to leave them there for now so people can get a sense of what it might look like.

Within the tab are the properties that can then be set before loading. I think having the `name` there is good, as is having `visible` as one day we might have some optimizations around adding invisible layers (e.g. see #578). For things like `rgb` and `is_pyramid` I might need to explore tristate checkboxes which I think exist in Qt so that I can represent the `None` state and not just `True` and `False`. Similarly for something like `channel_axis` we need to think about representing both the `None` state or an integer state. I'm still leaning against exposing all the possible keyword args, particularly those that can be easily modified from our normal UI after the layer has been added.

Here is a gif:
fileopen_dialog2

Any general comments / guidance now would be appreciated, though this is definitely still too early for review

royerloic:
Looks great @sofroniewn !
The sliders look ok to me, 
you might want to make the whole thing a tad bigger 
to accommodate long file names. File dialogs ar hard to 
make for some reason.  The best ones start with a lot
of real-estate to avoid a cramped feeling....

sofroniewn:
I've made some more progress on this PR including using tristate checkboxes for keyword args that can be `None`, `True`, or `False`. I think this works quite nicely.

I've also added checkboxes that enable setting of the `channel_axis` and `contrast_limits` optional properties, if unchecked default to `None` which then triggers proper default setting.

See the gif here:

more_open_dialog

I'm still not sure we made the right choice in https://github.com/napari/napari/pull/619#issuecomment-547055984 wrt to `CYM` vs `RGB` for colormap default ordering for 3 channel images (see what happens in this example ...) but that convo is for another PR

also note the stylings for the QDialog still need some work but that can come later.

Probably next things to do are finish up adding necessary keyword args for the image layer, make it clear the other layers aren't supported right now. I might move away from the tabs to a QComboBox, might make scaling to more supported layer types easier and make the stylings easier too, curious what people prefer.

Then I need to add more tests / clean up all the remaining stylings, and then it should be good to go.

Follow-up PRs can then start to add File Open support for other layer types, and even further away follow-up PRs can add support for File Save for all layer types

jni:
@sofroniewn nice!

> CYM vs RGB

I still think for microscopy this is a better choice. The example is working just fine with RGB! =)

> I've made some more progress on this PR including using tristate checkboxes for keyword args that can be None, True, or False. I think this works quite nicely.

I like the idea, but I don't love the horizontal line... I'd have no idea what that means if I didn't know what the three options are in the code. Suggestions:

- a question mark
- a half-filled (diagonally?) box
- the text "AUTO" (is there a symbol for automatic?)
- ... something else?

royerloic:
Looks great!

sofroniewn:
Ok this PR is coming along nicely now, and I think could use another pair of eyes. The stylings definitely leave something to be desired right now - the very tall / thin approach looks a little weird, and I still need to improve the 3-choice box, but all the major functionality is there now. I also need to add comprehensive tests, and this should not be merged until that is done.

Here is a gif of the current state:

open_dialog_3

Note only adding `Image` layers is currently supported, but there is machinery in place to start supporting more layer types once this goes in

sofroniewn:
This PR is now ready for some review. Maybe @tlambert03 can check it out. It probably still needs some tests before merge, but would be good to see what others think at this point. I've now added support for adding labels too. See here:

open_labels

One question is - would it be nice to see a thumbnail in the display. The way I might be tempted to do it is to hijack the `add_*` functionality - try to "add the layer" as soon as there is data (and refresh whenever params change) and leverage the thumbnails that are auto generated. What's nice is that this becomes a little preview mode and so you know if the data is going to give an error or actually load before you hit "load", if it will error we can just show an "error" warning or something. I think it's probably best to do that in a new follow-up PR once this merges as this PR is already complex enough. Similarly I will not be adding fileIO for any of our other layer types to this PR but having two layer types in there now gives you a feel for how this will go

tlambert03:
sorry for the absence this week!  hectic schedule.  will try to review this tonight

sofroniewn:
no rush @tlambert03!

tlambert03:
not gonna leave a formal review yet... but after playing with it for about 30 minutes, I'm realizing just how big of a challenge this really is, and how many "gotchas" it opens up.  styles are the minor points, and as you said: would be nice to have 2 columns instead of tall/thin and the tristate buttons could be tweaked.

I think the bigger issue here is just how open-ended the user interaction can be with this dialog, and all of the potential ways a user can "break" this (I just tried doing a bunch of stupid things to check).  I recognize this is aimed towards a slightly more adept user, but if it's being added to the GUI I'm not sure we can really make that assumption.  Here's a couple things:

- if you enter a non-exist path (e.g. typo), it gets added to the list just fine, but then throws an error on opening.  More importantly, _any_ errors that get thrown are shown in the console (which may be hidden depending on how they launched napari) so a lot of people may not see it at all, and wonder why the open button isn't doing anything.  We haven't done this in napari yet, but in LLSpy [I ended up overriding `sys.excepthook`](https://github.com/tlambert03/LLSpy/blob/94d7ee6ff6505c22cf891a59d8f378d1ec7beed7/llspy/bin/llspy_gui.py#L116-L119) to catch certain exception classes and show an informative error QDialog (otherwise they get printed to the console).  With something as prone to user-error as this, it might be time to consider something like that.  (That of course, is a whole different PR... but it's what came to mind the most here)
- related to your thumbnail suggestion above, I think it would be a good idea to do some greedy header reading at least.  I don't know that I think you actually need to read the image data, but grabbing image or folder shape could be very valuable.  For instance, the `channel_axis` argument is _very_ easy to get in trouble with.  If you pick the wrong axis (like the x or y axis)... it will likely hang.  Knowing/showing the shape of the dataset would be very helpful there.  in LLSpy when a user drops a folder of files, I put the available metadata in a table (of course not all of that info is available with arbitrary data here, but if we could at least anticipate whether `channel_axis` would crash it, that would be great.):

llspy

- the meaning of many of these parameters (for instance `channel_axis`) gets a little tricker when adding multiple datasets of different shape.   There too, I don't really have a great answer, but just suspect that a lot of people will try to do something we didn't anticipate.  

so in general I guess I'm wondering how you feel about putting some effort into validation (i.e. anticipating possible problems) and more immediate & interpretable error messages...

sofroniewn:
So I started playing around with the "mock viewer" and "add_layer" idea that @jni mentioned for thumbnail generation / data validation, but the naive way I was planning on doing it basically involves making many many reads of the data, which just seems like a terrible idea, right? I can think about ways that involve only making one read but they still might be a little complex from a code standpoint. Thoughts @jni @tlambert03? Is that worth it?

AhmetCanSolak:
Hi @sofroniewn . I got similar requests from various people about forming a hyperstack while opening a folder of images. I think we might address this within this PR possibly. Or do you prefer to have it in a separate PR? Let me know. I will be interested taking it right away if you think before dialog PR we might be able to provide this functionality.

sofroniewn:
@AhmetCanSolak by opening hyperstack - do you mean something like point to folder of pngs with the open folder method and all pngs get stacked to form a 3D array then then gets added with one slider? I think the open folder button should do that now without this PR (similar if tif, or already 3D etc. you just get one extra dim added)

AhmetCanSolak:
@sofroniewn that's good, im happy as soon as it works with folder of nD images and make a n+1D stack with added slider by default.

sofroniewn:
@AhmetCanSolak have you tried that behavior on master? I just confirmed it works for me right now. I added a folder of four 5D tiffs and got a 6D image with first axis four. Support for this was added by @jni in #590 and earlier PRs.

sofroniewn:
We've come a long way since this PR too. Although there were some good ideas here for the channel axis/ preview stuff, I think we should focus on letting people transform data in the GUI after it is loaded (see #2556 for example) rather than push on this approach of a more complex file open menu. We might want to add an "open with hookspec" option from the GUI, but that is a separate piece of work

tlambert03:
we barely knew ye, open dialog.

PR #617, opened 2019-10-24
Add view_multichannel function

### LINKED ISSUE #615: Add `view_multichannel` method
## 🚀 Feature

I'm not sure whether we made a conscious decision not to add the `view_multichannel` function corresponding to the `add_multichannel` Viewer method, but if so I now believe that this decision was in error. =P

## Motivation

While working to add documentation, I found that I could meet many use cases of napari with a single `view_` call. However, for `add_multichannel`, I needed to first create an empty viewer, which is kind of ugly. I think it's pretty nice functionality, to, e.g. for the Mitosis Fiji sample, which is order TZCYX, to go:

```python
napari.view_multichannel(mitosis,
    axis=2, colormap=['magenta', 'green'], scale=[10, 1, 1]
)
```
and be done.

jni:
Fixes #615

I haven't added the viewer keyword arguments because they have so far
not been added to any of the other layers. It should be added in a
different PR.

### DIFF
--- a/napari/view_layers.py
+++ b/napari/view_layers.py
@@ -574,3 +574,84 @@ def view_vectors(
         visible=visible,
     )
     return viewer
+
+
+def view_multichannel(
+    data,
+    *,
+    axis=-1,
+    colormap=None,
+    contrast_limits=None,
+    interpolation='nearest',
+    rendering='mip',
+    name=None,
+    metadata=None,
+    scale=None,
+    translate=None,
+    opacity=1,
+    blending='additive',
+    visible=True,
+):
+    """Add image layers to the layers list expanding along axis.
+
+    Parameters
+    ----------
+    data : array
+        Image data. Can be N dimensional.
+    axis : int
+        Axis to expand colors along.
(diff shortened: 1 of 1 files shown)


sofroniewn:
i just looked at the current codebase on master and it seems like view_multichannel is already there 
https://github.com/napari/napari/blob/849d3e1582a7c7dbe22180c03e3fe275e785146f/napari/view_layers.py#L110
and the viewer args have already been passed to it. Not quite why #615 was raised - you might need to update from master, but I don't think we need this PR anymore.

side note - it seems like our windows tests are now passing again which is great

jni:
> i just looked at the current codebase on master and it seems like view_multichannel is already there

Ooooooh, I git cloned from my fork — but my fork was not updated to latest master. 🤦‍♂ 

I had memories of your PR and the viewer kwargs but thought I must have dreamed them! 😅 

> side note - it seems like our windows tests are now passing again which is great

Wow! That's super-excellent! Thanks Cirrus-CI! 🎉

PR #618, opened 2019-10-24
[WIP] Delay importing RichJupyterClient until use

jni:
Workaround for https://github.com/jupyter/qtconsole/issues/388

### DIFF
--- a/napari/_qt/qt_console.py
+++ b/napari/_qt/qt_console.py
@@ -1,106 +1,115 @@
-from qtconsole.rich_jupyter_widget import RichJupyterWidget
-from qtconsole.inprocess import QtInProcessKernelManager
-from qtconsole.manager import QtKernelManager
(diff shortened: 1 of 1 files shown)


jni:
Not sure this is the prettiest way to do this, but it works! 😬

sofroniewn:
hmm yeah - i'm so not sure about this I think I'd rather move the `shell = get_ipython` into `QtViewer` and then only instantiate QtConsole if we need to and send the shell as one of its input parameters. Thoughts?

jni:
Incidentally, a whole bunch of stuff is broken in this PR, but the tests pass, so when I get around to fixing this, I'll add a bunch of tests for this stuff. =)

sofroniewn:
This has now been superseded by #1055 which makes loading the whole console lazy

PR #619, opened 2019-10-25
drop multichannel

sofroniewn:
# Description
As discussed during meeting today, this PR drops the `add_multichannel` and `view_multichannel` methods in favor of adding the `channel_axis` keyword argument to the `view_image` and `add_image` methods. This argument defaults to `None` in which case all functionality of `add_image` is as before. If provided then it will use the similar code to what used to be in `add_mulitchannel`.  To migrate over old code do as follows.

```python
add_multichannel(data) => add_image(data, channel_axis=-1)
add_multichannel(data, axis=2) => add_image(data, channel_axis=2)
```

I had to drop the doc string test for the Image layer now though as the `add_image` method diverges from the `Image.__init__`

## Type of change

# How has this been tested?

- [x] tests have been adapted too

## Final checklist:


napari/components/tests/test_multichannel.py
```diff
@@ -1,15 +1,15 @@
 import numpy as np
 from napari.components import ViewerModel
 
-base_colormaps = ['cyan', 'yellow', 'magenta', 'red', 'green', 'blue']
+base_colormaps = ['red', 'green', 'blue', 'cyan', 'yellow', 'magenta']
```
jni: @sofroniewn there's been a lot of fuss about RGB not being colorblind friendly compared to CYM. On the other hand, I understand that it's more surprising to default to that. So, I don't feel strongly about undoing this change but I still lean towards CYM. What do you think?

napari/components/viewer_model.py
```diff
@@ -600,8 +498,60 @@ def add_multichannel(
...
+                contrast_limits[0]
+            ):
+                contrast_limits = itertools.repeat(contrast_limits)
+            else:
+                contrast_limits = ensure_iterable(contrast_limits)
+
+            layer_list = []
+            zipped_args = zip(range(n_images), colormap, contrast_limits, name)
+            for i, cmap, clims, name in zipped_args:
+                image = data.take(i, axis=channel_axis)
```
jni: This code probably breaks for pyramids... =\

napari/tests/test_viewer_layer_parity.py
```diff
@@ -29,6 +29,10 @@
 def test_docstring(layer):
     name = layer.__name__
 
+    # Stop testing Image
```
jni: This is a little too drastic. Can we check that all arguments in Image are in add_image and view_image?

tlambert03:
As a your local colorblind representative 😎, I would definitely vote for green/magenta being the first two colors.  Once you get to 3 colors, it gets a bit harder.  CYM is fine, as is GMB... but if you don’t want to calculate the number of channels to modify the cmap, then I would vote GMBCRY.  Us color challenged folks would be appreciative :)

jni (approved):
@sofroniewn I've left a few comments. I think fixing pyramid support is critical before merging. The naive way is probably fine:

```python
if is_pyramid:
    image = [data[j].take(i, axis=channel_axis) for j in range(len(data))]
else:
    image = data.take(i, axis=channel_axis)
```

Other than my comments, the code looks great and we should indeed merge ASAP.

jni:
@tlambert03 fascinating! =) Btw did you see [BOP (blue orange purple)](https://twitter.com/christlet/status/1133726940127477761) and what's your take on it? I've been wanting to add it to napari for a while. =)

tlambert03:
I think BOP is aesthetically appealing, particularly in Leterrier’s hands, but it’s not my personal fave.  The blue/purple is a bit similar, and neither are that good for human intensity perception.  But, like I said, three colors is a bitch, so I give everyone a pass at that point to use whatever they please.  (I also recognize that fighting the colorblind friendly fight with green magenta puts you in a position of battling people’s expectations, which sucks too). maybe a good thing for the “preferences” tab... which I assume is a discussion to be had in the near future :)

royerloic:
> I think BOP is aesthetically appealing, particularly in Leterrier’s hands, but it’s not my personal fave. The blue/purple is a bit similar, and neither are that good for human intensity perception. But, like I said, three colors is a bitch, so I give everyone a pass at that point to use whatever they please. (I also recognize that fighting the colorblind friendly fight with green magenta puts you in a position of battling people’s expectations, which sucks too). maybe a good thing for the “preferences” tab... which I assume is a discussion to be had in the near future :)

I agree. We should curate such nice triplets and quartets luts for multi-channel purposes.
RGB is sooooo boring and unelegant. Napari is all about elegance. I would love to have Leterrier’s input on this more generally... What would be other nice color combinations? And good inverted-lut  color combinations...

napari/components/tests/test_multichannel.py
sofroniewn: I should have left a note explaining this, the reason for the change is that with this ordering if someone opens an rgb image (say jpeg or png) and adds it with `axis=2` they will get an image that looks the same as the original but with the layers separated.

For example now
```python
napari.view_image(data.astronaut(), channel_axis=2)
```
will yield:
multichannel_rgb

If we change to CYM we get:
multichannel_cym

which I think will be quite confusing to people.

I appreciate that in all other cases (say fluorescent microscopy data) CYM or GMBCRY as suggested by @tlambert03 is a better option.

On option could be to try and do some special casing say is axis is len(3) use RGB, otherwise go with GMBCRY, but that also might be confusing.

Another point of view is to say that we primarily expect people to use the channel_axis keyword arg with data that is not rgb to begin with (or where those colors don't have intrinsic meaning)

What are your thoughts @royerloic @jni @tlambert03

sofroniewn: After reading more of the comments below again, I think I'm leaning towards `RGB` if length 3 and `GMBCRY` otherwise, thereby getting GM if only two colors, which I think is really nice, preserving expectations for the length 3 special case, and then trying to do our best as we add more and more colors

jni: @sofroniewn well, we haven't removed `rgb=` here yet, have we? So I suggest that users with rgb images should use that keyword argument. Then if we have `rgb=True, channel_axis=2` we use multichannel rgb colormap, and if `channel_axis=None` we use the rgb display.

napari is primarily for scientific images, so I would very much prefer to cater to microscopy display than to natural image display, as a general rule, especially since we already provide a keyword argument for RGB.

sofroniewn: I’d rather keep the rgb parameter as is so that someone could use channel_axis to expand a list of rgb images if they wanted, though maybe then it should be called something like layer_axis, or expand_axis

As to scientific use case first, I agree, I can go with the GMBCRY order as default for now, we can always give people the option to change that default somewhere else in the future. Thoughts @jni?

jni: I think it's quite natural that if someone passes `rgb=True` and `channel_axis=[anything other than None]`, they mean to expand the RGB channels into layers. Currently what would happen? Would it break?

Whether we want that to be the recommended way of doing things is another matter. We can always add a docs example with `colormap=['red', 'green', 'blue']`, and have that be the "canonical" way of doing it, until a global config is defined.

cleterrier:
Thanks for the BOP appreciation. I'd like to add something that has not been discussed I think.

I think an important aspect of LUT combinations is to keep the three colors as an orthogonal base in the color space. This means that 255,255,255 is white and that any combination of the three colors is unique (I think?). Also this means that a combined image can be split into it's components unambiguously. This is the case for RGB, CMY, BOP but not MGB (as M is R+B). MG is great as a 2-color combination though as it's sum is white.

When you have overlapping signal (colocalization) is where the importance of having an orthogonal decomposition of the space matters most. When signal are well segregated, you can go crazy with the combinations and favor aesthetics (these are usually the most pleasing images like the ones of cells with distinct cytoskeleton components).

The compliance to the different types of color blindness is another important aspect to consider. In the end, I agree that the best thing is to have the largest possible range of options (this is why I made the NeuroCyto LUTs update site for Fiji) and let users decide which aspect is the more important depending on image or use type. I'd favor having the classical color-space consistent combination featured first though (RGB, CMY).

cleterrier:
I think RGB should be used as the default decomposition method as it's consistent between dealing with 3-channels data and with RGB (!) images. The ImageJ order for multichannel (Image/Colors/Make Multichannel) makes sense in this regard (RGBCMYGray). It could be intersting to add a Preference to switch to a colorblind-friendly default such as CMYRGBGray. Sorry if I didn't understand the point of the OP, it wasn't very clear to me.

jni:
@cleterrier thank you so much for chiming in!

> Sorry if I didn't understand the point of the OP, it wasn't very clear to me.

You seem to have grasped it quite well:

> The ImageJ order for multichannel (Image/Colors/Make Multichannel)

This is effectively what this PR is about, and this discussion more specifically!

> This is the case for RGB, CMY, BOP but not MGB (as M is R+B). MG is great as a 2-color combination though as it's sum is white.

I don't think this is strictly true? Since the B channel is independent of R, you should be able to unambiguously separate the M and B signal? This does not account for clipping at white, but by that standard, CMY also fails? (Unless inverted.)

At any rate, I don't think RGB is a good default for this purpose because we are talking about *visualisation*, which means that the magnitude of the different channels from low end to high end should be perceptually similar. This is wildly inaccurate for RGB, even for non-colorblind people.

Having said all that, I *do* think it's important that the displayed channels should maximally sum to white, rather than overflow, so CMY and MGB fail this test. This is my main motivation for proposing BOP. (Your description on Twitter of this process as "sudoku" was brilliant. ;)

Although it adds complexity, I'm kinda leaning now towards different colormaps for different numbers of channels. 2-channel: MG, 3-channel: BOP?

To reiterate: my feeling is that perceptually, RGB fails on so many levels that it is not a sensible default. My feeling is that it was chosen as the default in ImageJ because it's the easiest choice, rather than the *best* choice from some perceptual optimality argument. I'm really quite sick of barely being able to see nuclei in microscopy images. I think BOP, and also OPF from later in that thread, make much more sense for microscopy (or any other non-natural images) than RGB.

tlambert03:
> Although it adds complexity, I'm kinda leaning now towards different colormaps for different numbers of channels. 2-channel: MG, 3-channel: BOP?

I agree, and it's really not that much complexity.  The answer feels pretty obvious for 1 (gray) & 2 (GM) colors.  For 3+ colors, I don't really have a strong opinion, I feel like they all fail (me) in one way or another;  BOP, RGB, and CMY each have strengths.  I like @royerloic's suggestion of a curated list; and it would be cool if the user could pick their own 3-color and 4-color behavior (even if that just meant picking between a few preselected options).

Speaking of preferences, I don't see any obvious threads about adding `QSettings` and persisting behavior across sessions.  I have some experience with that from other projects, should I start an issue for that, or is that something you all have a plan for already?

jni:
> Speaking of preferences, I don't see any obvious threads about adding QSettings and persisting behavior across sessions. I have some experience with that from other projects, should I start an issue for that, or is that something you all have a plan for already?

Please start an issue, but I'm not sure I like QSettings as the framework — I'd like to have preferences remain as free of Qt as possible, in case we ever want to add another frontend e.g. something web-based.

tlambert03:
yeah, I agree.  We can just wrap some more fundamental settings object in QSettings for the view layer, but design it as a Qt free key/value pair.  Will open shortly

napari/components/viewer_model.py
sofroniewn: yup, fixed, thanks for your suggestion

napari/components/tests/test_multichannel.py
sofroniewn: for `rgb=True`, if you have data that is `[100, 100, 5, 3]` and you split along `channel_axis=2` you'll get 5 rgb images that are 100x100. I think this is the expected behavior, and I've added a test to confirm it works

I am going to go with `GM` if len < 3, and `CYMRGB` otherwise. we can expose more default settings another time

napari/tests/test_viewer_layer_parity.py
sofroniewn: I have done this for add_image.

We don't actually have any doc string tests for any `view_*` methods, but should add them in a new PR

sofroniewn:
Thanks for the great discussion all, I think giving people more flexibility around setting defaults sounds great, and making a new issue for that @tlambert03 would be much appreciated.

For the colormaps, I've gone with `GM` if the length of the axis is < 3, and `CYMRGB` otherwise, which I think strikes a good balance. To add split an rgb image and have it preserve colors you can now do
```python
napari.view_image(astronaut, channel_axis=-1, colormap=['red', 'green', 'blue'])
```
I also made the pyramid fix @jni pointed out and have added tests to catch it.

I'd like to merge this today. This will be a deprecation so have to make sure we update docs / releases appropriately when the time is right.

Issue #611, opened 2019-10-21
TransFire & TransGrays colormaps don't work on scalars


tlambert03:
## 🐛 Bug

The `TransFire` and `TransGrays` colormaps slightly break the vispy colormap API when `vispy.BaseColormap.__getitem__()` is called with a scalar for example:

```python
In [1]: from napari.util.colormaps import colormaps

In [2]: inferno = colormaps.vispy_or_mpl_colormap('inferno')

# colormaps should be able to take a scalar and return a ColorArray
In [3]: inferno[.5]
Out[3]: <ColorArray: 1 color ((0.732796, 0.21433249, 0.332053, 1.0))>

# but transfire throws an error
In [4]: tfire = colormaps.TransFire()                                                                                                                                               
In [5]: tfire[0.5]
```
```                                                                                                                                                       
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-17-af8bd54a5e2c> in <module>
----> 1 tfire[0.5]

~/miniconda3/envs/napdev/lib/python3.7/site-packages/vispy/color/colormap.py in __getitem__(self, item)
    291         item = np.clip(item, 0., 1.)
    292         colors = self.map(item)
--> 293         return ColorArray(colors)
    294 
    295     def __setitem__(self, item, value):

~/miniconda3/envs/napdev/lib/python3.7/site-packages/vispy/color/color_array.py in __init__(self, color, alpha, clip, color_space)
    146 
    147         # Parse input type, and set attribute"""
--> 148         rgba = _user_to_rgba(color, clip=clip)
    149 
    150         if alpha is not None:

~/miniconda3/envs/napdev/lib/python3.7/site-packages/vispy/color/color_array.py in _user_to_rgba(color, expand, clip)
     55     color = np.atleast_2d(color).astype(np.float32)
     56     if color.shape[1] not in (3, 4):
---> 57         raise ValueError('color must have three or four elements')
     58     if expand and color.shape[1] == 3:  # only expand if requested
     59         color = np.concatenate((color, np.ones((color.shape[0], 1))),

ValueError: color must have three or four elements

```

I think the issue is related to the [`np.dstack`  in `TransFire.map()`](https://github.com/napari/napari/blob/85e6bd46a2b1be230cceef429316e41405aa6356/napari/util/colormaps/colormaps.py#L257) that returns an array with an extra dimension when called on a scalar, causing the ValueError shown above:

```python
[...]

PR #622, opened 2019-10-25
Standardize usage of cmaps and add cmap tests

tlambert03:
# Description
In adding the gamma slider (PR #610), I realized that a couple of the colormaps didn't behave like the others (#611).  After some digging, I found a number of places in napari in which colormaps were being used in a way that was inconsistent with the `vispy.colors.Colormap` API described at: http://vispy.org/color.html . This PR adds tests to make sure all `AVAILABLE_COLORMAPS` abide by the vispy API, and corrects a few places in the code where the `cmap.map()` method was being called directly in a way that led to different results depending on the cmap being used.

It's easy to write tests that assert existing cmaps follow the API, but it's hard to make sure they get used this way in the code... So just to summarize the API (for later reference), `vispy.color.Colormap`s should be used in one of two ways:

1. `cmap.__getitem__(item)` (i.e. `cmap[item]`)

   This is the preferred API (from vispy: "Users are expected to use a colormap with ``__getitem__()`` rather than ``map()`` (which implements a lower-level API)."  This API expects `item` to be either a scalar value, or a 1D numpy array (it does not handle 2D+ arrays).  It returns a `vispy.color.ColorArray` which has a few convenient methods such as `rgba()` or `RGBA()` that return normalized data in various dtypes that can then be reshaped as desired.  An example usage of this API (as in `make_colorbar`)
   ```python
   color_array = cmap[image.ravel()]
   # the ColorArray.RGBA method returns a normalized uint8 array
   cbar = color_array.RGBA.reshape(image.shape + (4,))
   ```
2. `cmap.map(item)`

   This is the lower-level API. Data provided to `map()` **MUST** be of shape `(N, 1)`, not `(N,)`, with values of [0,1]. (this was the issue that cause the inconsistent behavior).  So, for example:
this is ok: `cmap.map(image2d.reshape(-1, 1))`
but this is not: `cmap.map(image2d.reshape(-1))`
any new custom `cmaps` implementing a `map()` method must return an rgba array of size `(N, 4)`

This PR will be necessary for the gamma PR #610 to work consistently across cmaps.

## Type of change

# References
http://vispy.org/color.html [& [source](https://github.com/vispy/vispy/blob/876a0d37116f5b29c7b190c8e67cbf9247c45f04/vispy/color/colormap.py#L256-L293)]

# How has this been tested?
- [x] added tests to make sure that all available colormaps adhere to the vispy API assumptions

## Final checklist:

### DIFF
--- a/napari/layers/image/image.py
+++ b/napari/layers/image/image.py
@@ -533,8 +533,8 @@ class Image(Layer):
             color_range = high - low
             if color_range != 0:
                 downsampled = (downsampled - low) / color_range
-            colormapped = self.colormap[1].map(downsampled)
-            colormapped = colormapped.reshape(downsampled.shape + (4,))
+            color_array = self.colormap[1][downsampled.ravel()]
+            colormapped = color_array.rgba.reshape(downsampled.shape + (4,))
             colormapped[..., 3] *= self.opacity
         self.thumbnail = colormapped
 
@@ -584,8 +584,8 @@ class Image(Layer):
         color_range = self.contrast_limits[1] - self.contrast_limits[0]
         if color_range != 0:
             image = image / color_range
-        mapped_image = (self.colormap[1].map(image) * 255).astype('uint8')
-        mapped_image = mapped_image.reshape(list(self._data_view.shape) + [4])
+        mapped_image = self.colormap[1][image.ravel()]
+        mapped_image = mapped_image.RGBA.reshape(image.shape + (4,))
         image_str = imwrite('<bytes>', mapped_image, format='png')
         image_str = "data:image/png;base64," + str(b64encode(image_str))[2:-1]
         props = {'xlink:href': image_str}
--- a/napari/layers/labels/labels.py
+++ b/napari/layers/labels/labels.py
@@ -368,7 +368,7 @@ class Labels(Image):
             col = None
         else:
             val = self._raw_to_displayed(np.array([label]))
-            col = self.colormap[1].map(val)[0]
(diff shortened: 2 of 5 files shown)


sofroniewn (approved):
This looks great @tlambert03, great catch - I didn't realized that about the preferred vispy api, so that's nice standardization.

I'll let this sit briefly in case @jni wants to comment and then I'll merge so that you're not blocked on   #610

Great work!

napari/layers/image/image.py
```diff
@@ -533,8 +533,8 @@ def _update_thumbnail(self):
             color_range = high - low
             if color_range != 0:
                 downsampled = (downsampled - low) / color_range
-            colormapped = self.colormap[1].map(downsampled)
-            colormapped = colormapped.reshape(downsampled.shape + (4,))
+            color_array = self.colormap[1][downsampled.ravel()]
+            colormapped = color_array.rgba.reshape(downsampled.shape + (4,))
```
jni, line 537: why lowercase here and uppercase elsewhere? (I prefer lowercase if they are equivalent)

jni:
@tlambert03 I left a minor question, but it can be resolved in a followup PR. This is an improvement already! Thanks for digging into it!

napari/layers/image/image.py
tlambert03: That’s just the [ColorArray API](http://vispy.org/color.html#vispy.color.ColorArray).
Uppercase gives you a unit8 array of [0, 255], lowercase gives you float of [0, 1].

PR #629, opened 2019-10-27
Fix repeated calls to _on_data_change

### LINKED ISSUE #625: redundant calls to VispyImageLayer._on_data_change
## 🐛 Bug

This is just to start an issue to discuss the redundant calls to `VispyImageLayer._on_data_change()` as disccused in https://github.com/napari/napari/issues/575#issuecomment-544797848 ... which will likely be a drag on performance

I know this will be _much_ easier for @sofroniewn and others to solve, but I just wanted to use it as an opportunity to understand the napari event sequence a little better, and I'm adding some observations here in case it's useful for anyone else's debugging process.  I was able to track down the 6 paths leading to `_on_data_change`.  In the image below, you can trace the unique ways to get to `_on_data_change` from the `Dims.events.ndisplay()` event... all 6 of them are traversed when `ndisplay = 3`, only 5 are traversed when changing `ndisplay = 2`.

Screen Shot 2019-10-25 at 4 59 20 PM

and if you log events in the order they happen ... here is the (nested) sequence of calls 
```
components.Dims @ndisplay.setter
    <Dims.events.ndisplay event emitted>
        VispyImageLayer._on_display_change
            Layer._update_dims
                Image._set_view_slice
                    <Image.events.set_data event emitted>
                        ** VispyImageLayer._on_data_change **
            Image._set_view_slice
                <Image.events.set_data event emitted>
                    ** VispyImageLayer._on_data_change **
            VispyImageLayer.reset
                VispyImageLayer._on_contrast_limits_change
                    ** VispyImageLayer._on_data_change **
                ** VispyImageLayer._on_data_change **

        Layer._update_dims
            Image._set_view_slice
                <Image.events.set_data event emitted>
                    ** VispyImageLayer._on_data_change **
        Image._set_view_slice
            <Image.events.set_data event emitted>
                ** VispyImageLayer._on_data_change **
```

sofroniewn:
# Description
This will close #625. I still want to add some tests and check it works in all cases, but the basics should be done.

## Type of change

# How has this been tested?

## Final checklist:


tlambert03:
nice!  works for me... couldn't find any obvious issues, and switching between from 2D->3D feels noticeably snappier for larger datasets

sofroniewn:
Great, thanks for checking it out @tlambert03 - this is now ready for review @jni.

Only change in the last commit is adding tests using mock that check `_on_data_change` only gets called once. Right now I have 6 separate tests, one for each of our layer types, but we should probably start cleaning that stuff up so we have one parameterized test and reduce the amount of duplicate code. That applies across the repo in many places and come come later though

Also in the benchmark that I added for a `128 x 128 x 128` volume the 2D -> 3D transition went down from 74ms to 35ms, so thanks again @tlambert03 for catching this and diagnosing the problem so well in that other issue, it make tackling it much easier!

There are probably other places in the repo where similarly unnecessary duplicate calls are happening and we should start ironing those out and adding more benchmarks / tests

tlambert03:
nice work.  you're very welcome, going through the diagnosis was really helpful for me too... i feel much more familiar with the general napari structure now.  I don't think I've said it explicitly yet but I _love_ this project.  I'm really impressed with the framework you all have constructed here, and I look forward to seeing it grow and contributing more.

sofroniewn:
thanks @tlambert03, we've loved your contributions - if there are particular features you need or would be excited to work on, more performance optimizations, or anything else we can do to make contributing a better experience for you just let us know. You've been doing a great job so far.

royerloic (approved):
Love the use of mock to catch excess calls to methods.
Very nice!

sofroniewn:
I've just update this PR to remove merge conflicts, will merge after the tests pass

PR #633, opened 2019-10-28
fix undefined variables, remove unused imports

tlambert03:
# Description
feel free to reject or ignore this PR if it is annoying to merge or for any other reason... It refactors a number of different files, but changes no functionality and affects no tests.  It is mostly to remove unused imports, but there were a few syntax errors and undeclared variables that I had seen (including the `'cmap' is not defined` bug that @GenevieveBuckley mentioned in #632).

If you don't wish to merge, you can use it as a list of things to consider... but the actual "bugs" (syntax or undeclared variables) are in these files:
`napari/view_layers.py ` (cmap & clims undefined)
`napari/util/list/_typed.py ` (self undefined)
`napari/layers/surface/surface.py ` (warnings not imported)
`napari/layers/shapes/keybindings.py ` (self undefined)

if you don't want anything but the strict bug fixes, I can update the PR accordingly.

## Type of change

- [x] Bug-fix (non-breaking change which fixes an issue)

# How has this been tested?
- [x] example: all tests pass with my change

PR #638, opened 2019-10-28
Add linter pre-commit hook 

tlambert03:
# Description
picking up the conversation started in https://github.com/napari/napari/pull/633#issuecomment-546956411, this PR would add a pre-commit lint hook to prevent things like syntax or logical errors from being committed.

This PR is just a stub for now, to give everyone a chance to play with various linters.  My (very weak) personal preference has been for flake8.  While it's true that it covers a bit less than pylint I find that it's got more than enough to prevent me from accidentally introducing bugs, and can be expanded with plugins.  Ultimately, I found that if I just wanted to prevent actual errors from creeping in, and assert a little more conventions, flake8 was sufficient... whereas pylint took more time to tune so as not to get in the way.  On the flip side, there's obviously more momentum behind pylint.

I'm definitely not attached to flake8, and it's your call!  if you come to have a preference one way or the other, feel free to directly edit my branch, or leave a comment.

## Type of change
development environment

# References
https://github.com/PyCQA/pylint
https://github.com/PyCQA/flake8

## Final checklist:
- [x] add precommit hooks
- [x] agree upon linter configuration and put settings in `setup.cfg`
- [x] (optional) - run the linter on the entire repo and make sure it passes, so that we don't have a lot of small unrelated changes each time someone edits a file.  This has the potential to cause a lot of merge conflicts in the short term, but will help prevent commit failures in the long term.


tlambert03:
After playing a bit more with both flake8 and pylint, I decided to just push forward with flake8.  It was faster than pylint and should prevent any logical bugs from slipping in as well as many stylistic ones.  If in the future we'd like to move to pylint, it should be straightforward.

These are the errors I chose to ignore for now. 
See https://lintlyci.github.io/Flake8Rules for details
```
# E203  Whitespace before ':'  (sometimes conflicts with black)
# E501  line too long (> 79 characters)  (sometimes too annoying)
# W503	Line break occurred before a binary operator (conflicts with black)
# E711	Comparison to none should be 'if cond is none:' (there are tons of these)
# E712	Comparison to true should be 'if cond is true:' or 'if cond:' (lots)
# C901  McCabe complexity test.
```
The two main offenders of the [complexity test](https://lintlyci.github.io/Flake8Rules/rules/C901.html) were `Shapes._move` and `Shapes.on_mouse_press`, so if we can simplify those, it might be nice to re-enable that.

To bring the whole repo into flake8 compatibility with the remaining checks (e.g. `flake8 napari` at the command line), I fixed the following errors.  Note, many of these were fixed simply by running black... which would have happened eventually.

```
E114 indentation is not a multiple of four (comment)
E116 unexpected indentation (comment)
E123 closing bracket does not match indentation of opening bracket's line
E126 continuation line over-indented for hanging indent
E127 continuation line over-indented for visual indent
E131 continuation line unaligned for hanging indent
E123 closing bracket does not match indentation of opening bracket's line
E201 whitespace after '['
E202 whitespace before ']'
E241 multiple spaces after ':'
E302 expected 2 blank lines
E305 expected 2 blank lines after class or function definition
E704 multiple statements on one line (def)
F841 local variable assigned to but never used
W293 blank line contains whitespace
W504 line break after binary operator
```

I'm also currently ignoring all `__init__.py` files, since they have a lot of unused imports.  The "recommended" way to deal with those seems to be either to add a `# noqa` comment:
```python
from .my_class import MyClass  # noqa
```
or to define `__all__`
```python
from .my_class import MyClass
__all__ = ['MyClass',]
```
but it's up to you whether you'd like to enable that, or avoid the boilerplate.
This is all set for review/feedback.

.pre-commit-config.yaml
```diff
@@ -6,3 +6,8 @@ repos:
       # force black to run on whole repo & keep settings from pyproject.toml
       pass_filenames: false
       args: [.]
+-   repo: https://gitlab.com/pycqa/flake8
+    rev: '3.7.8'
```
bjlittle: @tlambert03 `3.7.9` just landed a few days ago

.pre-commit-config.yaml
```diff
@@ -6,3 +6,8 @@ repos:
       # force black to run on whole repo & keep settings from pyproject.toml
       pass_filenames: false
       args: [.]
+-   repo: https://gitlab.com/pycqa/flake8
+    rev: '3.7.9'
+    hooks:
+    - id: flake8
+      pass_filenames: true
```
bjlittle, line 13: @tlambert03 Apologies, I'm pretty new to pre-commit hooks... but isn't the `pass_filenames` by default `true`, see [here-ish](https://pre-commit.com/#new-hooks).

If that's the case, then is there a reason to set this explicitly? Just curious, thanks :smile:

tlambert03: I thought being explicit would make it easier for people to quickly glance at the config and know what was going on (without searching the docs)

tlambert03: I thought being explicit would make it easier for people to quickly glance at the config and know what was going on (without searching the docs)

bjlittle: Nice one, thanks @tlambert03 

Makes sense to me. I just wanted to clarify (for my understanding) that's what you were doing and not working around some other issue. Awesome!

bjlittle:
@tlambert03 At the moment I'm attempting to `black`ify and `flake8` the `iris` code base, see https://github.com/SciTools/iris/pull/3502/files. So I'm facing similar challenges.

I've also opted to ignore a handful of warnings and errors, which when compared with other scientific python packages, seems reasonable; particularly since `black` is involved.

IMHO the warnings and errors that you're ignoring here in this PR seems totally reasonable to me. However, I'd suggest that we should aim to resolve the `E711` and `E712` in a separate single follow on PR. I'm guessing that to resolve  `C901` may be harder and require more  concerted effort, so as a strategy going forward I'd suggest also resolving that separately.

If you buy into that approach, then perhaps raise associated issues to cover those two pieces of work i.e., `E711`+`E712` and `C901`, so that they are not forgotten about, and we just bank the goodness of this PR "as is"?

tlambert03:
Hi @bjlittle, it would be easy enough to fix `E711`&`E712` in this PR.  I was mostly waiting to give @jni and @sofroniewn an opportunity to look at the existing changes and weigh in before editing a lot more lines of the existing code base.

napari/_qt/qt_main_window.py
```diff
@@ -12,17 +12,17 @@
 app.use_app(API_NAME)
 del app
 
-from qtpy.QtWidgets import (
+from qtpy.QtWidgets import (  # noqa
```
sofroniewn: why do we end up with a `# noqa` here?

sofroniewn (commented):
Looking over it, it seems ok to me - though we had been ignoring the `vendored` colormaps stuff I think for `black` so it was kept the same as the code from other places, right @jni? Might want to ignore here too.

I also don't really like the appearance of `# noqa`. To me they do more harm then good (confuse a beginner, whereas only provide minimal clarity to an expert). Can we disable them?

In general though I don't have super strong opinions about this stuff (though I like having as much automated as possible) and defer to @jni.

Also your comment about the complexity of some code in the shapes layer makes sense, we should probably improve that, but that can be done in new PRs

napari/tests/test_viewer.py
```diff
@@ -45,7 +45,7 @@ def test_no_qt_loop():
     magic.
     """
     with pytest.raises(RuntimeError):
-        viewer = Viewer()
+        viewer = Viewer()  # noqa
```
bjlittle: @tlambert03 Can you drop the `# noqa` here and just have:
```python
with pytest.raises(RuntimeError):
    Viewer()
```

jni: Another alternative with me hoping against hope: does `_ = Viewer()` get ignored?

napari/_qt/qt_main_window.py
```diff
@@ -12,17 +12,17 @@
...
     QMainWindow,
     QWidget,
     QHBoxLayout,
     QLabel,
     QAction,
     QShortcut,
 )
-from qtpy.QtGui import QKeySequence
+from qtpy.QtGui import QKeySequence  # noqa
```
jni: What is the "violation" here? (and in the line below)

napari/_qt/qt_viewer.py
```diff
@@ -262,9 +262,10 @@ def _on_cursor(self, event):
 
     def _on_reset_view(self, event):
         if isinstance(self.view.camera, ArcballCamera):
-            self.view.camera._quaternion = self.view.camera._quaternion.create_from_axis_angle(
+            quat = self.view.camera._quaternion.create_from_axis_angle(
                 *event.quaternion
             )
+            self.view.camera._quaternion = quat
```
jni, line 268: Why is this change needed? (sorry to be "nosy" — I want to know for my own understanding but also to decide whether I want to add this to the ignore list...)

napari/components/tests/test_viewer_model.py
```diff
@@ -81,7 +81,7 @@ def test_add_empty_points_to_empty_viewer():
 def test_add_empty_points_on_top_of_image():
     viewer = ViewerModel()
     image = np.random.random((8, 64, 64))
-    image_layer = viewer.add_image(image)
+    viewer.add_image(image)
```
jni: Yeah, these are annoying to me, mainly because `*_layer = viewer.add_*(data)` is a "canonical" form for napari so we would like to keep to it as much as possible. Perhaps we can test some aspect of the added layer, e.g. `assert image_layer.dims.ndim == 3`?

napari/layers/points/tests/test_points.py
```diff
@@ -46,7 +46,7 @@ def test_negative_points():
     assert len(layer.data) == 10
 
 
-def test_empty_points():
+def test_empty_points_array():
```
jni, line 49: nice one! :sweat_smile:

tlambert03:
> though we had been ignoring the vendored colormaps stuff I think for black so it was kept the same as the code from other places, right @jni? Might want to ignore here too.

sorry, i missed that in pyproject... I'll ignore it in linting too.

> I also don't really like the appearance of # noqa. To me they do more harm then good (confuse a beginner, whereas only provide minimal clarity to an expert). Can we disable them?

yeah I hear ya... We can't disable them per se.  If you have code that doesn't follow one of the linting rules, you can either 
1.  ignore the specific error globally 
2. ignore the entire file
3. ignore the offending line with `# noqa`
http://flake8.pycqa.org/en/3.1.1/user/ignoring-errors.html

So every so often you come across something you want to ignore locally, but not globally or across the whole file.  One option is to agree to be more specific about the comment, e.g. `# noqa: E731,E123`, which provides more clarity to experts, though still probably confuses beginners. Fixing the error completely is preferable, but not always possible (though the few `noqa's` here are not necessarily that case, I was just being "gentle" with corrections).

napari/_qt/qt_main_window.py
tlambert03: that code breaks  [Module level import not at top of file (E402) ](https://lintlyci.github.io/Flake8Rules/rules/E402.html) ... but I didn't want to just move the imports, since that can sometimes alter behavior.  so rather than globally ignoring `E402`, I copped out with the  `# noqa`.  see my comment below about `#noqa` ... but I can also take a closer look to make sure moving the imports wouldn't break anything here.

napari/util/colormaps/vendored/_color_data.py
```diff
@@ -43,7 +43,8 @@
 
 # Normalize name to "tab:<name>" to avoid name collisions.
 TABLEAU_COLORS = OrderedDict(
-    ('tab:' + name, value) for name, value in TABLEAU_COLOR_TUPLES)
+    ('tab:' + name, value) for name, value in TABLEAU_COLOR_TUPLES
+)
```
jni: Yeah, @sofroniewn is right: let's ignore everything in `vendored/`

napari/tests/test_viewer.py
tlambert03: @bjlittle, yes you could.  But i was again opting for the more explicit option there, since @jni wrote that code and I didn't risk messing up the test.  @jni, you're in luck: flake8 will ignore unused underscore variables.  I think that's the best solution here.

napari/_qt/qt_main_window.py
tlambert03: same as above: [E402](https://lintlyci.github.io/Flake8Rules/rules/E402.html).  the problem is the `app.use_app(API_NAME)` above it.  options are to ignore E402 globally, skip this file, or use `# noqa`

napari/util/tests/test_keybindings.py
```diff
@@ -61,15 +61,15 @@ def test_bind_key():
     kb = {}
 
     # bind
-    f = lambda: 42
+    f = lambda: 42  # noqa
```
jni: Yeah, like @sofroniewn, I'm a little allergic to the `# noqa`'s everywhere. If there is a way to configure this to instead be `#flake8ignore` or somesuch (explicit is better than implicit), I'd be more inclined to ignore (heh) them. As it is, I think it's better to do as it says, and replace this with an explicit function definition.

napari/_qt/qt_main_window.py
bjlittle: @jni I'm guessing that it's caused by [lines 12-13](https://github.com/napari/napari/blob/master/napari/_qt/qt_main_window.py#L12-L13) coming between the `import` statements.

I don't know `napari` well enough yet (working on that, ♥️ it btw), but could:
```python
app.use_app(API_NAME)
del app
```
be moved after all the `import` statements?

setup.cfg
```diff
@@ -5,3 +5,16 @@ versionfile_source = napari/_version.py
...
 parentdir_prefix = napari-
+
+[flake8]
+# Ignores - https://lintlyci.github.io/Flake8Rules
+# E203  Whitespace before ':'  (sometimes conflicts with black)
+# E501 line too long (84 > 79 characters)  (sometimes too annoying)
+# W503	Line break occurred before a binary operator
+# E711	Comparison to none should be 'if cond is none:'
+# E712	Comparison to true should be 'if cond is true:' or 'if cond:'
+# C901 McCabe complexity test. Would be nice to reenable, but takes work
```
jni, line 14: Never heard of this one! I quite like it! Yes, I envision making multiple (many?) smaller PRs to get this fixed. I wouldn't look forward to reviewing a huge one that tackles it for the whole library!

setup.cfg
```diff
@@ -5,3 +5,16 @@ versionfile_source = napari/_version.py
...
 tag_prefix = v
 parentdir_prefix = napari-
+
+[flake8]
+# Ignores - https://lintlyci.github.io/Flake8Rules
+# E203  Whitespace before ':'  (sometimes conflicts with black)
+# E501 line too long (84 > 79 characters)  (sometimes too annoying)
+# W503	Line break occurred before a binary operator
+# E711	Comparison to none should be 'if cond is none:'
+# E712	Comparison to true should be 'if cond is true:' or 'if cond:'
```
jni: Good call waiting on the first round of review to fix these, but I agree with @bjlittle (:wave:!) that we should definitely tackle them!

jni (changes requested):
@tlambert03 thanks! Man, enabling a linter takes work! :sweat_smile: I've left a few comments, but overall I'm pretty happy with how this is shaping up!

napari/_qt/qt_main_window.py
jni: @bjlittle :tada: so great to see you here. =D If I remember correctly, it needs to come before the PyQt imports, but I'm not positive about this. The basic idea is to get VisPy and QtPy speaking the same "dialect" of Qt. Trying to figure this out for our [conda-forge recipe](https://github.com/conda-forge/staged-recipes/pull/9983), in case you might have input there! :grimacing:

napari/_qt/qt_viewer.py
tlambert03: lol @ the concept of _you_ being "nosy" on a napari PR.
The issue there was `E501` (line too long).  Sometimes, black is unable to come up with a satisfying solution (like in that case with a few long chained attributes).  Black will fail gracefully but flake8 won't, so that was just a way to decrease the line length.

napari/util/tests/test_keybindings.py
jni: @tlambert03 I definitely prefer the more explicit `#noqa: E<whatever>` form, thanks for pointing that out. It's a strong hint about e.g. what to Google to figure out what that comment is doing.

napari/tests/test_viewer.py
jni: YASSSS! Let's do that where needed!

napari/_qt/qt_viewer.py
jni: Ah good! It hadn't occurred to me that black could be unable to fix that. Happy with this workaround. Thanks!

jni: Also: I try to be mindful to minimize annoying nitpicks in PRs, but sometimes I can't help myself! :joy:

napari/components/tests/test_viewer_model.py
tlambert03: yeah, i guess these are the annoyances of linting 😕
how do you feel about `_ = viewer.add_image(image)` here?  It's just a test after all (but I recognize that tests are a great place for people to learn to use it, and we'd like to establish conventions there)

jni: Yep, let's go for it! Indicating at least that something is returned is good enough for me!

napari/_qt/qt_main_window.py
sofroniewn: those imports might need to be there, I can't fully remember, but I do think there were a couple places like that due to qt stuff

napari/_qt/qt_main_window.py
sofroniewn: that is my memory too @jni - i don't think it can be moved

tlambert03:
Just getting off train, will address all of these and resubmit soon

setup.cfg
tlambert03: if by "this one" you mean the mccabe test, I think we're actually in amazingly good shape that there are only two overly complex methods (on `Shapes`).  probably won't take long for someone to split those into a few methods and pass that test.

napari/util/tests/test_keybindings.py
tlambert03: looked into briefly but could not find any way to change the syntax anything other than `# noqa`.... will opt for explicit error codes

tlambert03:
- undid all changes within `vendored` and added it to `excludes` (note, the excludes are duplicated in `setup.cfg` and `pre-commit-config.yaml` ... I found some threads a similar requirement and went with it).
- fixed all `E711` and `E712`, (Comparison to `None`/`True`/`False`) and removed corresponding ignores from `setup.cfg`
- minimized `noqa`s.  All remaining ones have explicit flake8 codes
- ready to ignore `E501` (all lines ≥79 are either fixed or have `# noqa: E501`), but `E501` remains in `setup.cfg` ignores until everyone is ready for that level of linting strictness.

napari/util/tests/test_keybindings.py
bjlittle: @tlambert03 You should be able to tell `flake8` to ignore `E731` to allow assignment of `lambda`s... could you give that a try? If that works, then you can drop the `# noqa`

tlambert03: I already replaced it with an explicit function definition as @jni suggested above.  I don’t personally feel that assignment to lambdas provides any particular additional clarity besides simply fitting on one line... so hesitant to add E731 to ignores

sofroniewn (changes requested):
Thanks for the changes - they all look good to me, but I'll let @jni weigh in for final approval

Do we have to do anything extra to enable the pre-commit hook (especially if we already have napari installed)? Can you maybe leave a comment on this PR containing the info for current devs.

Can you add a line or two about our linting to our contributing guide and possibly include some helpful links - it can go next to the part about setting up `black` here - https://github.com/napari/napari/blob/master/docs/CONTRIBUTING.md#setting-up-a-development-installation

napari/layers/image/tests/test_image.py
```diff
@@ -206,7 +206,7 @@ def test_error_non_rgb_image():
     np.random.seed(0)
     data = np.random.random(shape)
     with pytest.raises(ValueError):
-        layer = Image(data, rgb=True)
+        Image(data, rgb=True)
```
jni, line 209: _ = 
```suggestion
        _ = Image(data, rgb=True)
```

jni: Eh, ignore me, just saw that it's not supposed to even get to the assignment step! :joy:

napari/util/list/_typed.py
```diff
@@ -6,7 +6,8 @@ def cpprint(obj):
 
 
 class TypedList(List):
-    """Enforce list elements to be of a specific type and allow indexing with their unique properties.
+    """Enforce list elements to be of a specific type and allow indexing with
+    their unique properties.
```
jni, line 10: This violates PEP257 (one-liner should be one line :joy:), but we can leave that for a later PR, as there's lots of those to clean up.

jni (approved):

tlambert03:
> Do we have to do anything extra to enable the pre-commit hook (especially if we already have napari installed)? Can you maybe leave a comment on this PR containing the info for current devs.

Assuming one has already installed pre-commit with `pre-commit install` as described in [CONTRIBUTING.md](https://github.com/napari/napari/blob/master/docs/CONTRIBUTING.md), the hook will be enabled once it gets merged, (but you may need to `pip install -r requirements.txt` if you don't already have flake8 installed)

> Can you add a line or two about our linting to our contributing guide and possibly include some helpful links - it can go next to the part about setting up `black` here - https://github.com/napari/napari/blob/master/docs/CONTRIBUTING.md#setting-up-a-development-installation

done!

jni:
@sofroniewn all clear from me, ball's in your court! =)

napari/util/list/_typed.py
tlambert03: yeah, that's a bit of a pickle isn't it!

sofroniewn (approved):
Looks great, thanks for adding those links. I will merge now

Issue #634, opened 2019-10-28
Be smarter about launching without a Qt event loop


jni:
Several new users have reported issues with the `%gui=qt` / `--gui=qt` / `with gui_qt()` stuff. It's actually a pretty big gotcha with napari right now. At a minimum, we should catch these exceptions in the following entry-points:

- `Viewer()`
- `view_*()`

and provide an informative error message. Better, we should detect when there is no event loop and create our own, which would be blocking in IPython — but this is fine, this is how matplotlib works.

jni:
CC @GenevieveBuckley @DrLachie

sofroniewn:
Agreed - minimum fail gracefully, 2nd option is to warn and wait an extra couple seconds, sometimes it may be coming, finally we should make our own.

Fixing this is a high priority

sofroniewn:
This has now been closed by #642

Issue #520, opened 2019-09-09
Providing viewer interactivity through scripting


AhmetCanSolak:
## 🚀 Feature

Right now due to nature of qt and our implementation it is not possible to interact with viewer programmatically through scripting. I want to propose providing a way to users to interact with viewer during runtime.

## Motivation

A script would be executed then rendered with the current implementation. But there are many case where one might want to interact with viewer in async fashion.

## Pitch

We can provide a method maybe for users to pass their code and we can call their code. Having a singleton QThreadPool in **napari** could be a nice basis. I have been already using a worker pattern to update viewer in async way. Hence, i am suggesting this. Open to other ways of providing such feature.

## Alternatives

I wont be providing alternatives on this initial comment but i am open to hear alternatives! 

## Additional context

An example script which is not working in the way we want by now:
```python
import numpy as np
from time import sleep
import napari
​
​
def main():
    with napari.gui_qt():
​
        # create an empty viewer
        viewer = napari.Viewer()
​
        # add the array
        layer = viewer.add_image(np.random.random((20, 40)), name='randArray')
​
        for i in range(10):
            layer.data = np.random.random((20, 40))
            print('Updating data...')
            sleep(.5)
​
​
if __name__ == "__main__":
   main()
```

Edit: Current solution for now, thanks to @bryantChhun:
```python
# bchhun, {2019-08-01}
import numpy as np
import napari
from PyQt5.QtCore import QThread
import time
from datetime import datetime
def send_random(layer_):
    # number of images to stream
    num_images = 10
    
    # time between image assignment
    update_interval = 0.5
    # pause so we can observe beginning/end of visual update
    time.sleep(2)
    print('from another thread')
    start = datetime.now()
    for k in range(num_images):
        time.sleep(update_interval)
        dat = np.random.random((512, 512))
        layer_.data = dat
        
        # check that data layer is properly assigned and not blocked?
        while layer.data.all() != dat.all():
            layer_.data = dat
    stop = datetime.now()
    print("time per update = "+
          str((stop-start).microseconds/num_images))
class ThreadRandom(QThread):
    def __init__(self, layer_):
        QThread.__init__(self)
        self.layer_ = layer_
    def run(self):
        send_random(self.layer_)
with napari.gui_qt():
    # create the viewer with an image
    data = np.random.random((512,512))
    viewer = napari.Viewer()
    layer = viewer.add_image(data)
    t = ThreadRandom(layer)
    t.start()
```

[...]

PR #642, opened 2019-10-29
Raise informative warning when no Qt loop is detected

jni:
# Description

Currently, trying `napari.Viewer()` or `napari.view_*()` outside of a Qt loop raises a nasty error. (#634). In this PR, I try to detect whether there is a Qt loop active, and if there isn't, I make one. This results in blocking behaviour when there is no loop, ie `view_image()` will pause the console until the user is done with napari.

Problems:
- I'm not sure how to test it, since once it blocks there's nothing really for us to do. And we can't launch a QTimer before starting this since the whole point is to not have a QApplication running.
- Given this, this has not been as thoroughly tested as I would like!
- It all seems a bit clunky. However, my [other attempt](https://github.com/jni/napari/commit/716c460c463fd318cce5899560cd35a6839145b7) to improve `gui_qt` so that it can be called during viewer creation doesn't work, and ends up with an empty napari window in some circumstances.

Comments very welcome! But let's get this in ASAP! =)

## Type of change

# References

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/viewer.py
+++ b/napari/viewer.py
@@ -1,3 +1,5 @@
+from time import sleep
+
 from qtpy.QtWidgets import QApplication
 from ._qt.qt_viewer import QtViewer
 from ._qt.qt_main_window import Window
@@ -22,19 +24,29 @@ class Viewer(ViewerModel):
     def __init__(self, title='napari', ndisplay=2, order=None):
         # instance() returns the singleton instance if it exists, or None
         app = QApplication.instance()
-        # if None, we create our own and we execute after creating our object
+        # if None, wait two seconds, as it may still be instantiating when
+        # using %gui qt inside IPython
         if app is None:
-            app = QApplication([])
-            self_execute = True
-        # otherwise, we don't execute and we delete our reference to app
-        else:
-            self_execute = False
-            # appears to be required to not conflict with IPython Qt loop
-            del app
+            sleep(2)
+            app = QApplication.instance()
+            # if still none, raise a RuntimeError with the appropriate message
+            if app is None:
+                message = (
+                    "napari requires a Qt event loop to run. To create one, "
+                    "try one of the following: \n"
+                    "  - use the `napari.gui_qt()` context manager. See "
+                    "https://github.com/napari/napari/tree/master/examples for"
+                    " usage examples.\n"
+                    "  - In IPython or a local Jupyter instance, use the "
+                    "`%gui qt` magic command.\n"
+                    "  - Launch IPython with the option `--gui=qt`.\n"
+                    "  - (recommended) in your IPython configuration file, add"
+                    " or uncomment the line `c.TerminalIPythonApp.gui = 'qt'`."
+                    " Then, restart IPython."
+                )
+                raise RuntimeError(message)
         super().__init__(title=title, ndisplay=ndisplay, order=order)
         qt_viewer = QtViewer(self)
         self.window = Window(qt_viewer)
         self.screenshot = self.window.qt_viewer.screenshot
         self.update_console = self.window.qt_viewer.console.push
-        if self_execute:
-            app.exec_()
--- a/napari/viewer.py
+++ b/napari/viewer.py
@@ -24,27 +24,22 @@ class Viewer(ViewerModel):
     def __init__(self, title='napari', ndisplay=2, order=None):
         # instance() returns the singleton instance if it exists, or None
         app = QApplication.instance()
-        # if None, wait two seconds, as it may still be instantiating when
-        # using %gui qt inside IPython
+        # if None, raise a RuntimeError with the appropriate message
         if app is None:
-            sleep(2)
-            app = QApplication.instance()
-            # if still none, raise a RuntimeError with the appropriate message
-            if app is None:
-                message = (
-                    "napari requires a Qt event loop to run. To create one, "
-                    "try one of the following: \n"
-                    "  - use the `napari.gui_qt()` context manager. See "
-                    "https://github.com/napari/napari/tree/master/examples for"
-                    " usage examples.\n"
-                    "  - In IPython or a local Jupyter instance, use the "
-                    "`%gui qt` magic command.\n"
-                    "  - Launch IPython with the option `--gui=qt`.\n"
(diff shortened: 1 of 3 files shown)


napari/viewer.py
```diff
@@ -19,8 +20,21 @@ class Viewer(ViewerModel):
     """
 
     def __init__(self, title='napari', ndisplay=2, order=None):
+        # instance() returns the singleton instance if it exists, or None
+        app = QApplication.instance()
+        # if None, we create our own and we execute after creating our object
+        if app is None:
```
sofroniewn, line 28: Do we want to print a warning that this is happening?

sofroniewn:
I'm very glad we're doing something to address #634 but I'm not 100% sure yet about whether we should be creating the `QApplication` instead of just displaying a very helpful error message, or even just waiting for 2 seconds and then trying again.

I feel like if we create the `QApplication` then yes that first call to say `view_image` will work from a script, but a subsequent call to `add_points` will not happen and then people will be really confused as to why they don't see the points. For me actually the following run from a script
```python
import numpy as np
from skimage import data
import napari

astro = data.astronaut()
# add the image
viewer = napari.view_image(astro, rgb=True)
# add the points
points = np.array([[100, 100], [200, 200], [300, 100]])
viewer.add_points(points, size=30)
```
just yields a black canvas with no astronaut and no points

Similarly running 
```python
viewer = napari.view_image(np.random.random((100, 100)))
```
from a jupyter notebook yields a blank canvas, and then blocks the notebook which is super confusing. As does
```python
%gui qt5
viewer = napari.view_image(np.random.random((100, 100)))
```
all from the same cell (the %gui magic hasn't had time to take effect).

I think I'd much rather us raise a really clear Error message like -

> napari has not detected an existing Qt event loop. You must first create an event loop either using `with napari.gui_qt():` from a script or `%gui qt5` from iPython or jupyter.

We could potential add a wait in there for the usage where someone has run the `%gui qt5` from inside the same cell as the `view_image` command and we just need them to wait for it to finish making it. Basically we could test if app is None. If None we could raise a warning that we will wait, then try again, and if still None we could fail again with that error message.

What are your thoughts on this @jni?

napari/viewer.py
AhmetCanSolak: possibly yes

AhmetCanSolak:
@sofroniewn what you are experiencing is a similar story with trying to update UI from script at the moment. IMHO, we can take this a bit more general and refactor our contextmanager choice too. This issue strongly related to #520 . If we can come up with a different way, possibly scripting users can get advantage on updating UI. How do you guys feel about removing contextmanager usage for napari viewer? and invisibly we take care of QApplication business?

sofroniewn:
@AhmetCanSolak if we can remove the need for a context manager / gui magic, and solve #520 that would be amazing. If you have a solution that you're working on feel free to start a PR. I think in the mean time though just catching this problem and adding a better error message will make napari a lot less frustrating for a lot of people.

AhmetCanSolak:
we can remove it and make it work, but have to introduce new behavior on notebooks :) we can take this offline and discuss further.

jni:
Thanks for testing @sofroniewn! Indeed it doesn't work when you want to do anything more than launch a viewer, and indeed from a scripting perspective that's a big problem.

As a first step I'm going to replace this attempt with just a warning message, as you suggest.

Regarding the model, we could do something like matplotlib's `plt.show()` — calls to napari don't result in a visible viewer until `napari.show()` is called, and this is blocking in scripts.

@AhmetCanSolak it seems I almost completely missed the discussion in #520! I have a very vague memory of it. I guess I was very busy at that time, sorry! :grimacing: I like some of the proposed solutions there. Both an internal context and a decorator seem like good solutions. Certainly, though, users should not have to deal with QThreads or any such nonsense. =P

jni:
@sofroniewn what do you think of this? Bonus: this can actually be tested! (Test on the way, just switching trains now!)

jni:
Hmmm, the test passes locally... Investigating, but ideas most welcome!

jni:
Oh wow: so once the qtbot has been instantiated once, the QApplication is available everywhere. I can make the test fail locally by moving it the second test instead of the first test. This is why I like context managers vs global state!

So, somehow we need to make sure that this test runs before any other tests involving qtbot. Any ideas?

jni:
btw this is what the error looks like. I quite like it:

napari-qt-error

sofroniewn:
This looks good, not sure how to help with the tests though, I can look into it tomorrow if you want

jni:
Woot, tests are passing! :tada:

AhmetCanSolak:
> 
> @AhmetCanSolak it seems I almost completely missed the discussion in #520! I have a very vague memory of it. I guess I was very busy at that time, sorry! 😬 I like some of the proposed solutions there. Both an internal context and a decorator seem like good solutions. Certainly, though, users should not have to deal with QThreads or any such nonsense. =P

No worries @jni :) feel free to weigh in #520 . I already played around little bit and ready to make a small PR about it.

napari/viewer.py
```diff
@@ -19,6 +22,29 @@ class Viewer(ViewerModel):
     """
 
     def __init__(self, title='napari', ndisplay=2, order=None):
+        # instance() returns the singleton instance if it exists, or None
+        app = QApplication.instance()
+        # if None, wait two seconds, as it may still be instantiating when
+        # using %gui qt inside IPython
+        if app is None:
+            sleep(2)
```
sofroniewn: This trick with `sleep(2)` is not working for me if I have one jupyter cell with both the magic and the view_image command in it. Even if I make it `sleep(20)` it still doesn't work, so I guess it is blocking the construction of the event loop. Similarly when launching from ipython I couldn't find a condition where having two checks and the sleep actually helped anything. I think we should remove the sleep and just raise the error - which is SO much better than the kernel dying like it used to. Unless you were able to find a condition when having the sleep in there prevented an error.

sofroniewn (changes requested):
I don't think our `sleep(2)` trick is working and so we should drop it, otherwise this is good to merge and I'll be so happy to get this in!!

sofroniewn (approved):
Ok great, looks good to me. I will merge after the tests pass

PR #644, opened 2019-10-29
Labeled axes

HagaiHargil:
# Description
This PR adds support for labeled dimensions, or axes, in `napari`. Each dimension can potentially have a name which will be displayed next to its corresponding slider.

## Type of change

# References
This is the first step in resolving #14. Once labeled axes exist we can read `xarray.DataArray`s and parse their dimension names into these labels.

# How has this been tested?
A single unit test has been added, I'll need to change a couple of other and add more before this is merged.

I squashed my commits into two fairly large ones due to a couple of issues I had with `pre-commit`, I hope it's not too bad.

## Final checklist:


kevinyamauchi:
I just wanted to drop by and say I am excited for this feature!

napari/components/dims.py
```diff
@@ -33,6 +33,8 @@ class Dims:
         Flag if to clip indices based on range. Needed for image-like
         layers, but prevents shape-like layers from adding new shapes
         outside their range.
+    axis_labels : list of str
```
sofroniewn: Can you also add `axis_labels=None` as a keyword argument to the Dims, and also `ndisplay=2`, and `order=None`, so the function signature is now
```python
def __init__(self, init_ndim=0, ndisplay=2, order=None, axis_labels=None)
```
and the initialize those variables as appropriate if `None` is passed, or use the passed values otherwise. Can you also add a check the `len(order)` and `len(axis_labels)` is equal to `init_ndim` if neither of them are `None`

napari/components/dims.py
```diff
@@ -110,6 +121,26 @@ def mode(self):
...
+
+        if len(labels) != self.ndim:
+            raise ValueError(
+                f"Number of labels doesn't match number of dimensions. Number"
+                f" of given labels was {len(labels)}, number of dimensions is"
+                f" {self.ndim}. Note: If you wish to keep some of the"
+                "dimensions unlabeled, use '' instead."
+            )
+        self._axis_labels = labels
+        self.events.axis_labels()
```
sofroniewn: do you think this event should emit as
```python
self.events.axis_labels(axis=list(range(self.ndim))))
```
for consistency with the other times it emits? I'm open

napari/components/dims.py
```diff
@@ -325,15 +364,45 @@ def set_mode(self, axis: int, mode: DimsMode):
...
+            Dimension index
+        label : str
+            Given label
+        """
+        axis = self._assert_axis_inbound(axis)
+        if self.axis_labels[axis] != str(label):
+            self.axis_labels[axis] = str(label)
+            self.events.axis_labels(axis=axis)
+
+    def _assert_axis_inbound(self, axis: int) -> int:
```
sofroniewn: love this, thanks!

napari/_qt/qt_dims.py
```diff
@@ -275,19 +303,19 @@ def _create_range_slider_widget(self, axis):
         -------
         slider : range slider
         """
-        range = self.dims.range[axis]
+        _range = self.dims.range[axis]
         # Set the maximum values of the range slider to be one step less than
         # the range of the layer as otherwise the slider can move beyond the
         # shape of the layer as the endpoint is included
-        range = (range[0], range[1] - range[2], range[2])
+        _range = (_range[0], _range[1] - _range[2], _range[2])
```
sofroniewn, line 321: thanks for all this, agreed we shouldn't be overwriting `range`, good catches!

napari/_qt/qt_dims.py
```diff
@@ -1,5 +1,11 @@
 from qtpy.QtCore import Qt, Signal
-from qtpy.QtWidgets import QWidget, QGridLayout, QSizePolicy, QScrollBar
+from qtpy.QtWidgets import (
+    QWidget,
+    QGridLayout,
```
sofroniewn: Here I'm not sure if we should be using a QGridLayout. I think we want to use a single top `QVBoxLayout` that contains a bunch of widgets with `QHBoxLayout`, one for each displayed axis. Each of these widgets with a `QHBoxLayout` can then have the label, and the slider (and then soon the play button after #607 goes in). Does that make sense?

napari/components/viewer_model.py
```diff
@@ -70,6 +70,8 @@ def __init__(self, title='napari', ndisplay=2, order=None):
 
         self.layers = LayerList()
 
+        self.axis_labels = [''] * ndim  # labels will be empty if data is 2D
```
sofroniewn: Here I don't want to see us make an extra property on the viewer model with the `axis_labels`, instead we should be passing this variable to the `Dims` constructor a couple lines above, so it now looks like 
```python
self.dims = Dims(ndim, ndisplay=ndisplay, order=order, axis_label=axis_label)
```

napari/components/viewer_model.py
```diff
@@ -407,6 +409,7 @@ def add_image(
         blending='translucent',
         visible=True,
         path=None,
+        axis_labels=None,
```
sofroniewn: as noted above I don't think `axis_labels` should pass through `add_image` at this point, we might want to revisit this stuff when we do the xarray stuff, but for this PR let's keep all the `axis_labels` stuff happening within the `Dims` on the viewer_model

napari/view_layers.py
```diff
@@ -25,6 +25,7 @@ def view_image(
     title='napari',
     ndisplay=2,
     order=None,
+    axis_labels=None,
```
sofroniewn, line 26: you'll ultimately want to add `axis_labels` as a keyword arg to all our `view_*` methods

napari/view_layers.py
```diff
@@ -111,6 +114,7 @@ def view_image(
         blending=blending,
         visible=visible,
         path=path,
+        axis_labels=axis_labels,
```
sofroniewn: as noted above I don't think this should go on the `add_image` method, but it should go a few lines above 
```python
viewer = Viewer(title=title, ndisplay=ndisplay, order=order, axis_labels=axis_labels)
```
You'll also need to change the function signature of `Viewer` in `viewer.py` to be
```python
    def __init__(self, title='napari', ndisplay=2, order=None, axis_labels=None):
```
and inside the `ViewerModel` in `components/viewer_model.py` to be
```python
    def __init__(self, title='napari', ndisplay=2, order=None, axis_labels=None):
```

napari/components/tests/test_dims.py
```diff
@@ -104,6 +104,11 @@ def test_indices():
     assert dims.indices == (1, 2) + (slice(None, None, None),) * 2
 
 
+def test_axis_labels():
```
sofroniewn, line 134: this test is great, can you also add tests with the new `Dims` function signature

napari/_qt/qt_dims.py
```diff
@@ -1,5 +1,11 @@
...
+    QSizePolicy,
+    QScrollBar,
+    QLabel,
```
sofroniewn: I think we want to make this QLineEdits so that you could edit the labels if you want. You can see examples of how we've done text edit boxes inside our layer widget - here https://github.com/napari/napari/blob/708271c47549e9de9b63316dfec6e381fcb0c7d2/napari/_qt/qt_layerlist.py#L344
you'll have to propogate any changes in the text back to the `axis_labels` property on the `Dims` model, but that shouldn't be too bad

sofroniewn: One of the advantages of the QLineEdit is that we can give it a fixed width and you then get scrolling left / right for free if someone uses a really long name

sofroniewn (commented):
This is great progress @HagaiHargil! I've left some comments that will hopefully help, the major ones are around the function signature for Dims, adding the axis_labels through our Viewer rather than `add_image`, using `QVBoxLayout` and `QHBoxLayout` instead of `QGridLayout` for the Qt stuff, and making the labels editable.

You'll need to add some tests for the `QtDims`, and our `view_*` methods, maybe adjust the code inside `examples/nD_image.py` so that it now takes in axis labels so people can see what it looks like (and include a screenshot in this PR).

We can then review again once those are done. Really great job though, a lot of people really want this feature!!

HagaiHargil:
Thanks for the comments, I hope this doesn't take too much of your time, I'm not too well versed with Qt apps. 

I was a bit reluctant to write extensive testing because I wasn't sure how much of this PR will survive your review. My next batch of commits will include many more. Also, I'm having troubles running `pytest` locally. Most likely it's my CentOS machine's fault, but just in case, is there any specific argument I should give `pytest`, or a folder to run it from?

I'll have more time to devote to this work after the weekend.

napari/components/dims.py
tlambert03: yeah, this emitter needs an axis kwarg, cause it is connected to
```python
def update_axis_listener(event):
    self.update_axis.emit(event.axis)
```

otherwise, the setter errs:
```python
v = napari.view_image(numpy.random.rand(10,10,10)) 
v.dims.axis_labels = list('xyz')
# throws:  AttributeError: 'Event' object has no attribute 'axis'
```

tlambert03:
Hey @HagaiHargil, thanks for the initiative on this!  I think a lot of people will appreciate this feature.  I haven't had time to dig deeper yet to figure out what might be going on, but in a quick test I noticed a couple things and I thought I'd just mention them in case you don't see them in your tests:

```python
import napari, numpy
v = napari.view_image(numpy.random.rand(10,10,10,10,10,10))
# viewer shows up, but there is only a single axis slider

v = napari.view_image(numpy.random.rand(10,10,10), axis_labels=list('zyx'))
# label doesn't make it to the QLabel
```
the second issue is because view_layers.add_image receives the axis_labels argument but doesn't do anything with it.  that will get fixed when you pass it to the Viewer() constructor instead as @sofroniewn suggested above, but just wanted to make sure you saw it.
looking forward to your updates!

napari/components/tests/test_dims.py
jni: I feel like this has been mentioned somewhere, already but haven't seen it in this PR... How about we make the default labels `[str(ax) for ax in range(ndim)]`? That will make it nicer to work with unlabeled axes, too!

jni:
@HagaiHargil this is looking pretty good already! I don't have time to do an in-depth review right now but I agree with all of @sofroniewn's comments, so that's a good place to pick up. I'm totally on board with waiting until we've fully settled the API to write the tests.

HagaiHargil:
Thanks for the warm welcomes and encouraging comments, sorry to keep you waiting :)

I resolved many of the aforementioned issues, but there are still a few remaining, mostly around setting and updating the axis labels on the go. My inexperience with Qt apps is really showing, so unfortunately the PR is not ready yet for merging. One "issue" I had is the pipeline surrounding `napari.view_image()` and the way it updates the GUI. Right now the data given to `view_image()` isn't passed to any of the class constructors (`Dims`, `QtDims`, `Viewer`, etc.) at first. Instead these classes create their "default" version and pass it around. The data only comes into play when the `Layer.append()` method is called on each of the data dimensions - and the GUI is updated accordingly. 

This instantiation method makes it difficult for me to pass the `axis_labels` with the constructor, since the code detects that I've given 5 labels (if my data has 5 dimensions), but it only sees the default data size of 2, raising an error as a result. This forces me to also supply an `order` argument to `view_image()`, even if I want to keep the default order. I guess I can hack around this issue, but I'm not sure I (we) should.

As I said, other than that my main issue is with the signaling of the Qt app to the `dims` variable, and actually setting and updating the labels. I'll need more help from you guys on this matter.

.gitignore
```diff
@@ -112,3 +113,6 @@ venv.bak/
 
 # Benchmarking results
 .asv/
+
+# VSCode
+.vscode/
```
HagaiHargil, line 118: I can undo this addition, as well as the `pip-wheel-metadata` one before this PR is merged. It's easier for me to just `git add .` in the mean time.

napari/_qt/qt_dims.py
```diff
@@ -213,6 +233,11 @@ def _update_nsliders(self):
             if self._displayed_sliders[i]:
                 self._update_slider(i)
 
+    def _update_axis_labels(self, axis):
```
HagaiHargil, line 237: This is one of the bigger issues I have, I'm not exactly sure what to write here, and it's a bit harder for me to "trial-and-error" it due to the larger issue I described in the conversation on this PR.

.gitignore
tlambert03: fwiw, you can also locally ignore files by editing your `.git/info/exclude` file
https://stackoverflow.com/questions/1753070/how-do-i-configure-git-to-ignore-some-files-locally

HagaiHargil: TIL :)

HagaiHargil:
Also, there are still quite a few issues which I haven't fixed. I decided to push my changes so far to get input from you on how to resolve the signaling mechanism.

napari/components/dims.py
```diff
@@ -110,6 +125,26 @@ def mode(self):
...
+    @axis_labels.setter
+    def axis_labels(self, labels):
+        if self._axis_labels == labels:
+            return
+
+        if len(labels) != self.ndim:
+            raise ValueError(
+                f"Number of labels doesn't match number of dimensions. Number"
+                f" of given labels was {len(labels)}, number of dimensions is"
+                f" {self.ndim}. Note: If you wish to keep some of the"
```
tlambert03:
```suggestion
                f" {self.ndim}. Note: If you wish to keep some of the "
```

napari/components/viewer_model.py
```diff
@@ -6,7 +6,7 @@
 from .dims import Dims
 from .layerlist import LayerList
 from .. import layers
-from ..util import colormaps
+from ..util import colormaps  # noqa F401
```
tlambert03: do you get a flake8 warning if you omit the `# noqa` here?  (that would be concerning since it is indeed used!)

tlambert03:
Nice updates @HagaiHargil!

> This instantiation method makes it difficult for me to pass the axis_labels with the constructor, since the code detects that I've given 5 labels (if my data has 5 dimensions), but it only sees the default data size of 2, raising an error as a result.

Good point.  One idea that might be worth discussing with everyone is whether the global `Viewer.dims` object that ultimately controls the slider labels really needs to have strict checking on the `axis_labels` at all.  I haven't thought my way through this entirely, but it's certainly conceivable that someone might want to create a viewer that will eventually have more dimensions that it does to begin with, and so `napari.Viewer(axis_labels='tzyx')` could make sense even without adding an image.  (and only the labels that correspond to visible dimensions would be shown).  So if the user then added a 3D image, only the "z" label would be shown, and if they then updated that 3D volume with a new timepoint, the "t" label would be shown on the new slider.  That would certainly make your life easier with the label validation.  Maybe that transfers the burden a bit to the user to remember the original `axis_labels` they used... but of course, as long as the `@axis_labels.setter` method and events are set up well (and as long as the labels are editable) that can be corrected at any point.  Might make your current task easier ... assuming others agree with that idea.

HagaiHargil:
Another semi-related issue to that is the fact that the `Dims` object currently has many implicit relations between its parameters. For example, we have `range` and `points` as separate parameters, even though they have a strict relation (size, value of `points` in `range`, etc.) that needs to be maintained. `interval` is also interconnected, along with `order` and the new `axis_labels`.

Perhaps we need a new `DimsMetadata` object that handles the state of these parameters internally, and asserts at all times that they're not in conflict. This new object can implement the above logic suggested by @tlambert03, but since its implementation is non of `Dims` business, we can potentially decide on a new logic whenever we wish. Another option is to handle this logic inside `Dims`, making it less of the "`dataclass`" that it's kinda is today.

napari/components/dims.py
```diff
@@ -110,6 +125,26 @@ def mode(self):
         """List of DimsMode: List of DimsMode, one for each dimension."""
         return copy(self._mode)
 
+    @property
+    def axis_labels(self):
```
HagaiHargil, line 141: The getter-setter approach here makes it difficult for me to allow code such as `my_dims_instance.axis_labels[3] = 'time'`. Not sure how to resolve it, perhaps making a class for it will help.

sofroniewn:
@HagaiHargil I'm going through some of this now, and seeing some of the messiness around internal conflicts with the lengths of some of these things like `order`, `axis_labels`, and `ndim` etc. that you are pointing out.

Do you mind if I make an attempt to clean some of this up and then push a couple changes directly to this PR? It might be the simplest way to get it done, and the clean up is slightly independent (but needed) for adding the main feature. I might also resolve the current merge conflict too.

sofroniewn:
@HagaiHargil I've been able to get the basics of the dims model working with axis labels on my copy of your PR. There's definitely still stuff to do around the Qt stuff and the tests, but I can push my changes to your branch if you like. You might need to enable edits by maintainers though - see here https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork.

If I now run:
```python
viewer = napari.view_image(np.random.random((10, 6, 30, 40)),
                           axis_labels=['t', 'z', 'y', 'x'])
```
you get the results bellow - this is going to be a very useful feature, so thanks for all your work so far getting us here!!
Screen Shot 2019-11-09 at 1 57 32 PM

napari/_qt/qt_dims.py
sofroniewn: This is where the new text from the dims model gets set to be the text on the corresponding QLineEdit box that gets displayed - something like
```python
        self.axis_labels[axis].setText(self.dims.axis_labels[axis])
```

napari/components/dims.py
sofroniewn: yes, we have that issue with all the properties where we use the setter approach, so that issue is definitely bigger than this PR.

unfortunately directly updating a list like that doesn't trigger the event emission that we need. it's definitely not ideal, but i'm not sure how to fix it

HagaiHargil:
Happy it works! You should be able to push them.

Regarding the way to construct `Dims` - the more I think of it, the more I'm inclined towards a table-like solution, where the rows of the tables represent the dimensions, their order being the current `self.order`, and the table's columns are the attributes (`point`, `axis_label`, etc.). It will also let you add more attributes along the way.

sofroniewn:
Hmm, I'm still getting
```
remote: Permission to PBLab/napari.git denied to sofroniewn.
fatal: unable to access 'https://github.com/PBLab/napari/': The requested URL returned error: 403
```
when I do `git push`. @jni does that make sense to you? I will look into this more

HagaiHargil:
Not sure why, it's a very simple operation. I've unchecked and re-checked it, you can try again I guess.

sofroniewn:
I may have set something weird up on my end so it's not pushing to the branch but somewhere else, thanks for checking the box though!

HagaiHargil:
Don't trouble yourself too much, if the changes are limited to a few files you can send them to me directly.

sofroniewn:
@HagaiHargil I just made a PR to your branch, it seems like I still don't have write access - so if you merge that PR the updates should appear here. Sorry about this.

sofroniewn:
Thanks @HagaiHargil for merging my PR to your branch.

To give you an update on where things are now - I've now updated the PR so the dims model can properly accept `axis_labels` and `order` and check that initial lengths are correct. The suggestion to use a table for some of the properties is worth thinking about, but I think we'll be ok now with these changes.

I also fixed the merge conflicts with the qt file and hooked up both the setter and editing events of qt to modify our dims model.

If you look in the gif below you can now see how the bidirectional communication works:

axis_labels

I had to comment out your tests containing `_assert_valid_init` as I dropped that function, but it would be nice if we added some tests back in to check similar stuff using the init of the `Dims` instead.

We should also think about the stylings on the QLineEdit - do we want to give it some background color? I can't figure out how to change that blue highlight color when it is being edited, so we might have to wait on that (I tried a while back in the context of another PR).

Otherwise I think we're pretty close to having this done. Do you want to go over it @HagaiHargil and see if there are any further edits you'd like to make, and then maybe it will be time for @tlambert03 and @jni to review?

napari/components/viewer_model.py
HagaiHargil: Not sure what was it about. It's gone, in any case.

HagaiHargil:
This looks great. Thanks for your help, couldn't have done it otherwise. 

I added a few tests and I think things look fine. The discussion about the table and other `Dims` stuff can go to a new issue, if it you think it's important enough.

sofroniewn:
Thanks for adding those tests! I think this is now ready for @jni and @tlambert03 to review. We can hold on making an issue for the dims model for now and see if it comes up again. Thanks for getting this going!!

After this goes in we'll probably make a follow up PR to add a play button next to the slider to give GUI access to the play feature @tlambert03 added in #607. We can also think about the second part of #14 where we autopopulate the axes labels based on an xarray - but I think I'd like to wait on that for a little bit to see how people use this feature before diving in. Is that ok @HagaiHargil - or is the autopopulation from xarray a high priority feature for you?

napari/_qt/qt_dims.py
```diff
@@ -310,6 +350,34 @@ def slider_focused_listener():
 
         return slider
 
+    def _create_axis_label_widget(self, axis):
+        """
+        Create the axis label widget which accompanies its slider.
+        # TODO - min and max width, word wrapping, text alignmemt.
```
jni: A minor nitpick: This and other docstrings should conform to [PEP257](https://www.python.org/dev/peps/pep-0257/) in addition to [NumPy's docstring conventions](https://numpydoc.readthedocs.io/en/latest/format.html) (which are a superset of PEP257):

```python
def my_function(arg1):
    """Provide a one-line description that ends in a full stop.

    Further explanation separated by a blank line.

    Parameters
    ----------
    arg1 : type
        Description of arg 1

    Returns
    -------
    label : type
        A description of the return value.
    """
```

napari/components/dims.py
```diff
@@ -143,29 +194,66 @@ def ndim(self):
...
+            # if in interval mode
+            self._interval = [(0, 1)] * (ndim - cur_ndim) + self._interval
+            self._mode = [DimsMode.POINT] * (ndim - cur_ndim) + self._mode
+            self._order = list(range(ndim - cur_ndim)) + [
+                o + ndim - cur_ndim for o in self.order
+            ]
+            # Default axis labels go from "-ndim" to "-1" so new axes can easily be added
+            self._axis_labels = [
+                str(i - ndim) for i in range(ndim - cur_ndim)
+            ] + self._axis_labels
```
jni: I don't think negative indices is the right choice here... Why not replace the whole set if new axes are added? [0, 1] -> [0, 1, 2] -> [0, 1, 2, 3]?

napari/components/dims.py
```diff
@@ -143,29 +194,66 @@ def ndim(self):
...
+            self._axis_labels = self._axis_labels[-ndim:]
 
             # Notify listeners that the number of dimensions have changed
             self.events.ndim()
 
+    def _reorder_after_dim_reduction(self, order):
+        """
+        When the user reduces the dimensionality of the array,
+        make sure to preserve the current ordering of the dimensions
+        while throwing away the unneeded dimensions.
```
jni: PEP257

napari/components/dims.py
```diff
@@ -325,15 +380,50 @@ def set_mode(self, axis: int, mode: DimsMode):
...
+
+        Raises
+        ------
+        ValueError
+            The given axis index is out of bounds
+        """
         if axis < 0:
             axis += self.ndim
         if axis < 0:
+            raise ValueError(f'Axis is too negative, got {axis}.')
```
jni: Heh. "too negative" is funny but imho doesn't quite capture it. And anyway the axis value printed is not what the function got. I suggest:

```python
if axis not in range(-self.ndim, self.ndim):
    msg = (f'Axis {axis} not defined for dimensionality {self.ndim}. '
           f'Must be in [{-self.ndim}, {self.ndim}).')
    raise ValueError(msg)

return axis % self.ndim
```

I also think it should be `in_bounds` instead of `inbound`, otherwise I think of "assert axis is incoming" 😄

napari/components/tests/test_dims.py
```diff
@@ -114,8 +146,60 @@ def test_order_when_changing_ndim():
...
+#     "ndim, propname, propvalue, default",
+#     [
+#         (3, 'order', range(4), []),
+#         (3, 'axis_labels', ('ax',), [str(ax) for ax in range(3)]),
+#         (2, 'order', [1], []),
+#     ],
+# )
+# def test_assert_invalid_init_raises(ndim, propname, propvalue, default):
+#     with pytest.raises(ValueError):
+#         Dims._assert_valid_init(ndim, propname, propvalue, default)
```
jni: Either keep these tests or don't. ;)

napari/components/viewer_model.py
```diff
@@ -43,7 +45,9 @@ class ViewerModel(KeymapMixin):
 
     themes = palettes
 
-    def __init__(self, title='napari', ndisplay=2, order=None):
+    def __init__(
+        self, title='napari', ndisplay=2, order=None, axis_labels=None
+    ):
```
jni, line 50: I don't think this should hold this PR, it's more of a long term discussion, but I feel like axis labels should come with the data, not the viewer. Putting axis labels on the viewer before sending in the data requires users to keep track of axes at *two* places: viewer initialisation, and data input. And if your datasets have different axis ordering, then you really have to do a lot of gymnastics.

So, medium term, I think we should update this to be an attribute on the data/layer dims, rather than on the viewer. All this is of course making me think of xarray...

jni (commented):
This has ended up surprisingly simple! 👍 ! I've made some minor comments around formatting and default axis names, and a major one that I think can wait for a later PR. But anyway, I think this is very close!

napari/components/tests/test_dims.py
sofroniewn: ah that was me - these can now be deleted that the new tests have been added

jni:
btw, regarding the problems with allowing edits from maintainers... Maybe GitHub only allows that through ssh rather than http? But I have no idea.

napari/components/viewer_model.py
sofroniewn: yup, fair points, though will need more discussion and agreed can come in followup PRs

benchmarks/benchmark_qt_viewer_image.py
```diff
@@ -13,7 +13,7 @@ class QtViewerViewImageSuite:
     params = [2 ** i for i in range(4, 13)]
 
     def setup(self, n):
-        app = QApplication.instance() or QApplication([])
+        app = QApplication.instance() or QApplication([])  # noqa F841
```
tlambert03: since we're trying to minimize the `#noqa`s, this could be:
```python
_ = QApplication.instance() or QApplication([])
```

napari/components/dims.py
```diff
@@ -110,10 +141,31 @@ def mode(self):
         """List of DimsMode: List of DimsMode, one for each dimension."""
         return copy(self._mode)
 
+    @property
+    def axis_labels(self):
+        """List of labels for each axis."""
+        return copy(self._axis_labels)
+
+    @axis_labels.setter
+    def axis_labels(self, labels):
```
tlambert03, line 146: it would be really nice if we could do `labels = list(labels)`, as that would then allow:
```python
viewer.dims.axis_labels = 'TZYX'
```
instead of 
```python
viewer.dims.axis_labels = ['T', 'Z', 'Y', 'X']
```

also... you could allow setting to `None` to return to the number indices.

napari/components/dims.py
tlambert03: I also don't love the negative indices

tlambert03 (commented):
This looks great and i look forward to having it merged!

napari/components/dims.py
sofroniewn: that was me - reason was that new axes add at the beginning not the end, so let's say you add a 2D image you'd have [0, 1], then if you add a 4D image afterward you'd really get [2, 3, 0, 1], whereas if you just added a 4D image it would be [0, 1, 2, 3]. To get around this I went with `-` for the default. Now maybe this just exposes a deeper problem, but adding new axes at the beginning was done to be consistent with broadcasting behavior

napari/components/dims.py
sofroniewn: both sound good, we'd need to add that `list` call in the `init` too as we do `self._axis_labels = axis_labels` there so don't trigger this setter

napari/components/dims.py
jni: @sofroniewn why can't you get [0, 1, 2, 3] when you add a 4D image? ie rename [0, 1] to [2, 3], then prepend [0, 1]?

sofroniewn: do we always rename? what if someone had set `['y', 'x]`. Maybe we should? It's not clear that `['-3', '-2', 'y', 'x']` or `['0', '1', 'y', 'x']` is preferable, so maybe forcing it to `['0', '1', '2', '3']` makes the most sense

jni: That I don't know, it's a bit of a corner case. My instinct is to only rename if the existing axes are the default, otherwise, add [0, 1, ..., `however_many_you_need`] to the front. But I don't think it's a big deal, the main Q is what to name prepended axes, and I think [0, 1, ...] makes the most sense, much more than negative values. I'd rather have empty strings than negative values tbh!

sofroniewn: ok - let's make this here
```python
self._axis_labels = [str(i) for i in range(ndim - cur_ndim)] + self._axis_labels
```
and
```python
self._axis_labels[axis] = str(axis)
```
down below. We'll also need to change the values in the test. I think these are a more sensible starting point, and we can examine the corner-case behavior as it comes up

jni: Wait, wouldn't that code result in `[0, 1, 0, 1]` being the labels?

sofroniewn: yes if someone went from 2D -> 4D it would, i thought we could accept that for now and revisit in the future though as I don't think that is a very common workflow at this point

jni: What? It's literally one of my examples in the blog post 😂, make a 2D image then overlay a 3D+ array for parameter scanning. I don't think it's too onerous to use instead:

```python
if self._axis_labels == list(map(str, range(cur_ndim))):
    self._axis_labels = list(map(str, range(ndim)))
else:
    self._axis_labels = list(map(str, range(ndim - cur_ndim))) + self._axis_labels
```

sofroniewn: yes - that looks like a better suggestion!!

HagaiHargil:
Thanks for all of your comments, I believe I addressed all of them.

`xarray`s are important in the long term. This PR goes 80% of the way for our current use cases. We can set a reminder (maybe using the "Projects" tab) to implement the automatic `xarray` parsing in a few months, after the dust of this PR settles.

napari/components/viewer_model.py
HagaiHargil: I left it as is currently.

sofroniewn (approved):
Looks great!! I will give @jni one more opportunity to weigh in, but all the changes were made and it is working great for me, so we should be able to merge very soon.

I'll leave the issue mentioning `xarray` open for now. I think we should be able to revisit this within weeks rather than months and get it finished up. Thanks so much for contributing!!

napari/view_layers.py
```diff
@@ -89,17 +90,21 @@ def view_image(
         The title of the viewer window.
     ndisplay : {2, 3}
         Number of displayed dimensions.
-    tuple of int
+    orider : tuple of int
```
```suggestion
    order : tuple of int
```

napari/_qt/qt_dims.py
```diff
@@ -365,7 +428,7 @@ def play(
                     stopped
 
         Raises
-        ----------
+        -----
```
jni:
```suggestion
        ------
```

napari/_qt/qt_dims.py
```diff
@@ -218,41 +234,57 @@ def _update_nsliders(self):
...
         # remove extra sliders so that only number_of_sliders are left
         # remove from the beginning of the list
         for slider_num in range(number_of_sliders, self.nsliders):
             self._remove_slider(0)
 
     def _remove_slider(self, index):
-        """
-        Remove slider at index
+        """Remove slider at index. Should also remove all accompanying widgets,
+        like the axis label.
```
jni: Still not compliant with PEP257: should be one line, with a longer description if needed after a blank line.

napari/components/dims.py
jni: Nope, still not PEP257 compliant :joy: Suggestion:

```
"""Ensure current dimension order is preserved after dims are dropped.

Parameters
----------
```

jni (commented):
@HagaiHargil I just left a few more very tiny comments / typo fixes. This is basically there!

HagaiHargil:
I guess this is it :)

sofroniewn:
@HagaiHargil @jni - yes I think it's all done now! I went through and directly edited some final pep stuff on the PR directly (seems like I was able to do that through the github UI, I just wasn't able to do it via a PR).

I will merge once the tests pass. @jni if you spot anymore pep stuff it's my fault and I think we can get to it in a refactor, I'd like to get this in today :-)

Thanks again for all your hard work @HagaiHargil and congrats on your first napari contribution!!

jni:
🎉

HagaiHargil:
Thank you for your technical support and encouraging words :)

PR #649, opened 2019-10-31
Fix np.pad() usage

### LINKED ISSUE #645: numpy.pad() issues on version 1.16
## 🐛 Bug

I had numpy version 1.16 and the `test_viewer.py` tests were failing because `np.pad()` was missing the `mode` positional argument (see below). I noticed that in [v1.17, `mode` is optional](https://docs.scipy.org/doc/numpy/reference/generated/numpy.pad.html) and when I upgraded numpy (1.17.3), the tests passed.

## To Reproduce

1. Install numpy version 1.16
2. run the tests in `/napari/tests/test_viewer.py`
3. Receive the following error

```python
self = <napari._vispy.vispy_shapes_layer.VispyShapesLayer object at 0x149450f28>

    def _on_data_change(self):
        faces = self.layer._data_view._mesh.displayed_triangles
        colors = self.layer._data_view._mesh.displayed_triangles_colors
        vertices = self.layer._data_view._mesh.vertices

        # Note that the indicies of the vertices need to be resversed to
        # go from numpy style to xyz
        if vertices is not None:
            vertices = vertices[:, ::-1] + 0.5

        if len(vertices) == 0 or len(faces) == 0:
            vertices = np.zeros((3, self.layer.dims.ndisplay))
            faces = np.array([[0, 1, 2]])
            colors = np.array([[0, 0, 0, 0]])

        if self.layer.dims.ndisplay == 3 and self.layer.dims.ndim == 2:
>           vertices = np.pad(vertices, ((0, 0), (0, 1)))
E           TypeError: pad() missing 1 required positional argument: 'mode'

../_vispy/vispy_shapes_layer.py:47: TypeError
```

## Expected behavior

Currently, the requirements specifty numpy >= 1.10.0, so I think we should bump to 1.17 (or the oldest version that passes).

kevinyamauchi:
# Description
As discussed in #645, this PR adds the `mode='constant'` keyword argument to `numpy.pad()` to maintain compatibility with `numpy < 1.17`.

## Type of change

# References
This PR closes #645 

# How has this been tested?
- [x] ran the tests in `/napari/tests

## Final checklist:

### DIFF
--- a/napari/_vispy/vispy_shapes_layer.py
+++ b/napari/_vispy/vispy_shapes_layer.py
@@ -44,7 +44,7 @@ class VispyShapesLayer(VispyBaseLayer):
             colors = np.array([[0, 0, 0, 0]])
 
         if self.layer.dims.ndisplay == 3 and self.layer.dims.ndim == 2:
-            vertices = np.pad(vertices, ((0, 0), (0, 1)))
+            vertices = np.pad(vertices, ((0, 0), (0, 1)), mode='constant')
 
         self.node._subvisuals[0].set_data(
             vertices=vertices, faces=faces, face_colors=colors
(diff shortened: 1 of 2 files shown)


sofroniewn (approved):
Looks good to me!

PR #650, opened 2019-10-31
Interactive Scripting

AhmetCanSolak:
# Description

This PR is an attempt to provide interactivity with **napari** viewer while scripting as easy as possible. This is a work in progress, please provide feedback to make it better. Main idea is letting API user write a python closure and pass it to `viewer.update_viewer()` method to create a separate thread without killing the main thread. Hence we can update UI without disrupting Qt event loop.
 

## Type of change

# References

#520 here we discussed about this feature and this is one of the best ways we can provide. Feel free to give feedback and offer different ways.

# How has this been tested?

- [X] a little example is written to test implementation 

## Final checklist:


AhmetCanSolak:
Hi @jni ! I updated example with arg and kwarg usage. It seems ready to review to me. Please provide feedback.

@sofroniewn i changed method name and right now the API is like `viewer.update()`. Please provide any other feedback you have.

sofroniewn (commented):
Looks good to me, @jni do you want to weigh in before merge?

AhmetCanSolak:
@sofroniewn I also want to add docstrings before merge. Please wait for that

examples/interactive_scripting.py
```diff
@@ -0,0 +1,27 @@
...
+
+with napari.gui_qt():
+    # create the viewer with an image
+    data = np.random.random((512, 512))
+    viewer = napari.Viewer()
+    layer = viewer.add_image(data)
+
+    update_period = 0.1
+
+    def layer_update(update_interval, kwargs):
```
jni:
```suggestion
    def layer_update(*, update_interval, num_updates):
```

examples/interactive_scripting.py
```diff
@@ -0,0 +1,27 @@
...
+            time.sleep(update_interval[0])  # Usage of ordinary arguments
+
+            dat = np.random.random((512, 512))
+            layer.data = dat
+
+            # check that data layer is properly assigned and not blocked?
+            while layer.data.all() != dat.all():
+                layer.data = dat
+
+    update_thread = viewer.update(layer_update, update_period, number_of_times=100)
```
jni:
```suggestion
    update_thread = viewer.update(layer_update, update_period=0.1, num_updates=100)
```

napari/viewer.py
```diff
@@ -45,3 +45,9 @@ def __init__(self, title='napari', ndisplay=2, order=None):
         self.window = Window(qt_viewer)
         self.screenshot = self.window.qt_viewer.screenshot
         self.update_console = self.window.qt_viewer.console.push
+
+    @staticmethod
+    def update(func, *args, **kwargs):
+        t = QtUpdateUI(func, args, kwargs)
```
jni:
```suggestion
        t = QtUpdateUI(func, *args, **kwargs)
```

jni:
This indeed needs some docstrings but also some tests!

jni:
To reiterate my comments from the meeting, I love this design!

examples/interactive_scripting.py
AhmetCanSolak: why do we accept any argument? Also i put `update_interval` and `kwargs` on purpose to demonstrate that both possible, do you still want me change @jni ?

examples/interactive_scripting.py
AhmetCanSolak: again i was trying to create variety here

napari/viewer.py
AhmetCanSolak: I am willing to accept this suggestion but what is the rational behind? I want to learn. As far as i observed it is not needed indeed?

examples/interactive_scripting.py
jni: Typically, the best practice in Python is to use explicit argument names wherever possible, and only use `**kwargs` where necessary. When defining small functions like this one, we should follow the best practices and use named arguments.

examples/interactive_scripting.py
jni: Understood. Generally, though, I think we should write our examples how we would like them to be written, rather than try to maximise variety. imho using named arguments makes the code significantly clearer, and that's a win. I don't feel like we need to explain closures, or positional vs named arguments in Python, but rather our update API, in this example.

napari/viewer.py
jni: I think it's a coincidence that it worked because your current code didn't have keyword-only arguments, and had exactly two arguments. `QtUpdateUI(func, args, kwargs)` was passing two positional arguments to `__init__`, both of which would have ended up in the `.args` attribute. Since this particular function could take two positional arguments, it was working fine, but it wasn't doing what you thought it was doing.

examples/interactive_scripting.py
jni: Oh, btw, `*,` doesn't mean accept any argument, it means, all arguments after these can only be specified as named arguments. (aka keyword-only arguments.)

examples/interactive_scripting.py
AhmetCanSolak: you are right @jni . updating example

AhmetCanSolak:
I made couple updates and it is ready for another round of reviews. implemented a test. It passes but not totally happy with the warnings. and i failed black formatting task? :D what. let me check and fix that

tlambert03:
Those aborts in the test reminded of a thought I had here.  Since the new QtUpdateUI threads introduce the possibility that a thread may be running when the user quits the program, you might want to clean those up in the `qt_main_window` `closeEvent` (otherwise users will see abort errors and segmentation faults upon closing).  Setting a parent `Qobject` on those `QThreads` would make finding them easier.

AhmetCanSolak:
> hen the user quit

You are right, but while scripting you won't be seeing anything(see the example i created). That's why API returns thread object to caller. So thread object will not be garbage collected while thread is running. 

To make it more clear @tlambert03 , with what i can achieve same functionality in close event of qt_main_window?

tlambert03:
> You are right, but while scripting you won't be seeing anything

I'm not sure I'm following... are you saying that the intended use case for scripting is such that the user wouldn't ever be able to quit while a script was still running?   Garbage collection of the thread is not the problem I was referring to, but rather this:  If you run your example with `ipython --gui=qt` instead of using the `napari.gui_qt()` context manager, and then run `exit()` while the thread is still running, you'll get this:
```
In [2]: exit()
WARNING: QThread: Destroyed while thread is still running
Abort trap: 6
```

One way to prevent that would be to assign all `QtUpdateUI` threads a parent `QObject`.  Then, in the `closeEvent`, you can search that object for all `children()` of type `QtUpdateUI` and call their `quit()` and `wait()` (or just `terminate`) methods to gracefully shut down the thread before closing napari.  Being a static method though (and lacking `self`), I'm not sure how to give the newly created thread a known parent behind the scenes without the end-user being required to provide it (which would be confusing to a lot of people).

Does having it as a static method provide a lot of freedom?  or is it always going to be associated with a `viewer` in one way or another?  If so, you change `update` slightly to be:

```python
    def update(self, func, *args, **kwargs):
        t = QtUpdateUI(self.window._qt_window, func, *args, **kwargs)
        t.start()
        return t  # returns thread object as it has to live on calling script
```
and then add some logic to the `qt_main_window.Window.closeEvent()` to look through its `self._qt_window.children()` for `QtUpdateUI` threads to quit.  Let me know if you want me to make that clearer in comments... or if I'm way off base here and shouldn't worry about it 😄

examples/interactive_scripting.py
```diff
@@ -0,0 +1,25 @@
...
+            time.sleep(update_period)
+
+            dat = np.random.random((512, 512))
+            layer.data = dat
+
+            # check that data layer is properly assigned and not blocked?
+            while layer.data.all() != dat.all():
+                layer.data = dat
+
+    update_thread = viewer.update(layer_update, update_period=0.1, num_updates=100)
```
jni: As I understand it, this will now fail our linting from #638, because `update_thread` is never used. You can rename it `_update_thread` and I think it will pass, or maybe it needs to be just `_`. @tlambert03?

jni (approved):
@AhmetCanSolak this looks really good to me now! I'm super excited to see this used for live updates from the microscope (I guess this is what you're developing it for?). I've left one comment that I think is important to prevent merge conflicts in the next few days (and the comment applies to the test also) but otherwise LGTM! :+1:

jni:
... Except I just saw that tests are failing... 🤦‍♂ LOL

napari/tests/test_viewer.py
```diff
@@ -344,3 +344,32 @@ def test_screenshot(qtbot):
...
+    view = viewer.window.qt_viewer
+    qtbot.addWidget(view)
+
+    layer = viewer.add_image(data)
+
+    def layer_update(*, update_period, num_updates):
+        # number of times to update
+
+        for k in range(num_updates):  # Usage of keyword arguments
+            time.sleep(update_period)  # Usage of ordinary arguments
```
jni: These comments need updating now. 😬

jni:
Longer term, I wonder whether the same thing can be achieved independently of Qt. I also wonder whether there is some way to ensure that the threads are killed eventually, see e.g. https://vorpus.org/blog/notes-on-structured-concurrency-or-go-statement-considered-harmful/

Right now, it seems that in a long-running IPython session, you could have many "phantom threads" continuously doing work if you forget to kill them... Am I right here?

jni:
... Perhaps I should read @tlambert03's helpful comments before providing my own. 😂 What he said! Adding the threads to a viewer and killing them when the viewer is closed seems like the right approach. The viewer then does the job of the "nursery" in the trio blog post.

tlambert03:
> Right now, it seems that in a long-running IPython session, you could have many "phantom threads" continuously doing work if you forget to kill them... Am I right here?

yeah, more generally: what I really like about this PR is that it's going to allow people to achieve some amount of concurrency without having to know anything about QThreads.  But by giving them the thread handle, we are leaving them responsible for (quitting) it.  We should go the extra mile and take care of that for them

napari/tests/test_viewer.py
AhmetCanSolak: yep :D

AhmetCanSolak:
> > You are right, but while scripting you won't be seeing anything
> 
> I'm not sure I'm following... are you saying that the intended use case for scripting is such that the user wouldn't ever be able to quit while a script was still running? Garbage collection of the thread is not the problem I was referring to, but rather this: If you run your example with `ipython --gui=qt` instead of using the `napari.gui_qt()` context manager, and then run `exit()` while the thread is still running, you'll get this:
> 
> ```
> In [2]: exit()
> WARNING: QThread: Destroyed while thread is still running
> Abort trap: 6
> ```
> 

This is happening fundamentally due to garbage collector when you run `exit()`.

> One way to prevent that would be to assign all `QtUpdateUI` threads a parent `QObject`. Then, in the `closeEvent`, you can search that object for all `children()` of type `QtUpdateUI` and call their `quit()` and `wait()` (or just `terminate`) methods to gracefully shut down the thread before closing napari. Being a static method though (and lacking `self`), I'm not sure how to give the newly created thread a known parent behind the scenes without the end-user being required to provide it (which would be confusing to a lot of people).
> 
> Does having it as a static method provide a lot of freedom? or is it always going to be associated with a `viewer` in one way or another? If so, you change `update` slightly to be:
> 
> ```python
>     def update(self, func, *args, **kwargs):
>         t = QtUpdateUI(self.window._qt_window, func, *args, **kwargs)
>         t.start()
>         return t  # returns thread object as it has to live on calling script
> ```
> 
> and then add some logic to the `qt_main_window.Window.closeEvent()` to look through its `self._qt_window.children()` for `QtUpdateUI` threads to quit. Let me know if you want me to make that clearer in comments... or if I'm way off base here and shouldn't worry about it 😄

Now i got what you offer better. Thank you for bearing with me and giving detailed reply @tlambert03 :D Being a static method gives a lot of freedom but i liked what you are offering here. Let me update the PR and will ask for another round of review. Thank you too @jni !

tlambert03:
one more thing I forgot to add ... in my example above with the modified `update` function, you would also need to modify the QtUpdateUI as follows:
```python
class QtUpdateUI(QThread):
    def __init__(self, parent, fn, *args, **kwargs):
        super().__init__(parent)
        ...
```

napari/_qt/qt_main_window.py
```diff
@@ -49,7 +50,10 @@ def __init__(self, qt_viewer, *, show=True):
         self._qt_window.setCentralWidget(self._qt_center)
         self._qt_window.setWindowTitle(self.qt_viewer.viewer.title)
         self._qt_center.setLayout(QHBoxLayout())
-        self._status_bar = self._qt_window.statusBar()
+        # self._status_bar = self._qt_window.statusBar()
+        # if you guys approve the change i will clean this line
+        self._status_bar = QStatusBar()
+        self._qt_window.setStatusBar(self._status_bar)
```
AhmetCanSolak, line 56: these changes not necessary after adding `sleep()` call to test but wanted to show basically what is needed to happen if we want to save all QWidget hierarchy from runtime gc.

AhmetCanSolak:
Updated PR, cleaned API further and NOT returning thread object back to user thanks to `QThreadPool`. Also instead of subclassing QThread now extending QRunnable class and let QThreadPool create thread accordingly for us. Due to this now our threads are _daemon_ like threads as @royerloic requested. 

Unfortunately, i introduced a `sleep()` call in test. Dropped a comment about what we need to do around view components to get rid of such sleep call.

Looking forward to hear more feedbacks from you guys!

napari/_qt/qt_main_window.py
sofroniewn: this change looks fine to me

napari/viewer.py
```diff
@@ -45,3 +45,8 @@ def __init__(self, title='napari', ndisplay=2, order=None):
         self.window = Window(qt_viewer)
         self.screenshot = self.window.qt_viewer.screenshot
         self.update_console = self.window.qt_viewer.console.push
+
+    def update(self, func, *args, **kwargs):
+        t = QtUpdateUI(func, *args, **kwargs)
+        self.window.qt_viewer.pool.start(t)
+        return self.window.qt_viewer.pool  # returns threadpool object
```
tlambert03, line 52: don't have an opinion, just curious: why did you choose to return the full threadpool instead of the new thread?

napari/_qt/qt_main_window.py
AhmetCanSolak: @sofroniewn we can merge this little change with this PR but if we want to keep alive all Qt objects from a single object, we need to do these kind of changes all over in qt folder. Would be a huge refactor.

tlambert03 (commented):
This looks awesome!

Maybe once this is merged I will submit an update to #607 to move the `QtDims._animation_thread` to your `threadpool` instead...

napari/_qt/qt_main_window.py
sofroniewn: ah i misunderstood, ok will need to think more then if needed. thoughts @tlambert03 @jni?

napari/viewer.py
AhmetCanSolak: We don't need to return threadpool. With current implementation threadpool is the one who creates and run thread from our runnable. I do not know an intuitive way to access created thread object. This pattern is good for run and forget type of threads. They are _like_ daemon threads, will not be affected by termination of the origin thread. More meaningful methods are provided on threadpool object, hence i returned that. Like there are methods to get information about pool and prevent execution of queued runnables. I am open to suggestions, wanna return something else? @tlambert03

tlambert03: no no, it makes sense to me.  Really was just curious

royerloic:
The use of a threadpool is a good move.

AhmetCanSolak:
Seems everybody okay, i am merging this now. We can continue discussion about QWidget objects lifetime somewhere else.

napari/_qt/qt_main_window.py
```diff
@@ -49,7 +50,10 @@ def __init__(self, qt_viewer, *, show=True):
         self._qt_window.setCentralWidget(self._qt_center)
         self._qt_window.setWindowTitle(self.qt_viewer.viewer.title)
         self._qt_center.setLayout(QHBoxLayout())
-        self._status_bar = self._qt_window.statusBar()
+        # self._status_bar = self._qt_window.statusBar()
+        # if you guys approve the change i will clean this line
```
sofroniewn, line 54: Can you remove these comments (a short followup PR will do, I can quickly merge)

sofroniewn:
@AhmetCanSolak thanks for merging, do you want to make a new issue to discuss the QWidget objects lifetime? Also do you mind making a quick PR to remove those commented lines, just keeps things a little cleaner

napari/_qt/qt_main_window.py
AhmetCanSolak: i was thinking to make an issue and referring there but i can remove since now we all following that discussion. let me make a small follow up PR.

AhmetCanSolak:
Commented lines cleaning PR merged and i dropped an issue for discussion here #689 . Thank you @sofroniewn !

Issue #631, opened 2019-10-27
error installing on ec2 deep learning ami


sofroniewn:
## 🐛 Bug

Following up on #495 I'm trying to install napari on an amazon ec2 instance with the ubuntu deep learning ami am but running into this error:

```python
ERROR: sparkmagic 0.12.5 has requirement ipykernel<5,>=4.2.2, but you'll have ipykernel 5.1.1 which is incompatible.
ERROR: sparkmagic 0.12.5 has requirement ipython<7,>=4.0.2, but you'll have ipython 7.9.0 which is incompatible.
ERROR: jupyter-console 5.2.0 has requirement prompt_toolkit<2.0.0,>=1.0.0, but you'll have prompt-toolkit 2.0.10 which is incompatible.
```

Once we get #587 in this should be improved. I don't know what `sparkmagic` is

sofroniewn:
If i `pip uninstall sparkmagic` this now goes away (we don't need that package), but I now run into

```python
  Found existing installation: qtconsole 4.3.1
ERROR: Cannot uninstall 'qtconsole'. It is a distutils installed project and thus we cannot accurately determine which files belong to it which would lead to only a partial uninstall.
```

We have a requirement of `>4.5.1` not sure how to proceed now, will keep trying

sofroniewn:
navigating to `/anaconda3/lib/python3.6/site-packages` and then doing `rm -rf` on the offending packages and then pip installing again seems to have worked

sofroniewn:
As to be expected I can now run `import napari` on in ipython or a jupyter notebook running from my ec2 instance, but I cannot actually view any images as it isn't possible to launch the `QApplication`. I want to now start investigating though if I can connect this python kernel to the qtconsole running inside napari on my local machine

jni:
@sofroniewn *sigh* the pip folks have done a lot of work to try to ensure that your distro remains consistent, but now it results in these annoying errors *that have no workaround*. It's insane that you have to `rm -rf` packages but I think that is sadly where we are at right now. 🤦‍♂ I'm not sure what the right repo is to complain about this.

sofroniewn:
Note the error message on trying the `%gui qt` magic is 
```
qt.qpa.screen: QXcbConnection: Could not connect to display
Could not connect to any X display.
```

YubinXie:
Hi, @sofroniewn  
I am curious that if you run it on AWS, how do you see the windows? Thanks!

sofroniewn:
[...]

PR #654, opened 2019-11-03
[WIP] remote viewer with dask

sofroniewn:
# Description
This PR is very work in progress right now, but it attempts to demonstrate how to use dask to potentially sync a remote viewer and a local viewer.

Before getting into the details, the use case I am imaging is someone who has a remote cluster that they are doing image analysis on, and they want to connect that cluster to a local napari running on their laptop and view images that are big and far away in a highly performant way. They will also have the ability to run code on the cluster, and want to continue to interact with napari and the viewer in the same way as when everything is local, for example running commands like `viewer.add_image(big_data)` and have that image appear in a napari viewer on their laptop screen

Following some guidance here https://github.com/dask/distributed/issues/3188 I use a [`Pub` / `Sub` paradigm](https://docs.dask.org/en/latest/futures.html#publish-subscribe) to synchronize one property of the viewer - in this example, the viewer.title - which is just a string. One of the nice things though about going with dask though, it is is very easy to also share `dask arrays` in a similar fashion, I know they work well with out nD slicing code.

To try this example you can start two local jupyter notebooks. One you will pretend is your "remote" ipython, the other will be your "local"

In the "remote" notebook:
```python
# remote process
from dask.distributed import Client, LocalCluster
cluster = LocalCluster(n_workers=0)  # Make a scheduler, but we don't need any workers for this
print(cluster.scheduler.address)
client = Client(cluster)

import napari
viewer = napari.Viewer(is_remote=True, address=cluster.scheduler.address)
viewer.remote.connect()
```
Notice here how we are using a new keyword arg `is_remote` that will prevent the QApp from being run and leave you just with the `ViewerModel` code, which is nice as this means it can now be run on an ec2 machine (see #631) for more discussion.

In the "local" notebook:
```python
# local process
%gui qt5
import napari

viewer = napari.Viewer(address='tcp://127.0.0.1:49804')
viewer.remote.connect()
```
Note that we need to provide the address to let dask know where to connect to, and that here we need to use our gui magic to make sure we actually launch the qt app and so have something to look at.

In the ideal world if you were to now do
```python
viewer.title = 'new_title'
 ```
In either process the `ViewerModel` in the other process would get automatically updated, and the GUI would get updated in local process if the change happened in the remote.

Unfortunately it doesn't seem to be working perfectly due to interactions between some of the `asyncio` event loop stuff, `Qt` event loops, and `jupyter` event loops, which I don't fully understand.

I think that we somehow want to replace the following line
```python
asyncio.get_event_loop().create_task(self.sub())
```
with something that more explicitly leverages the `QtEventLoop` for the case where it exists. That line seems to work fine for the `is_remote=True` viewer. @AhmetCanSolak do you have any ideas how to do this better?

Overall I'm imagining that we'd want to sync up all the properties in the viewer and on the layers using this `napari._dask.util.RemoteTopic` approach. Does that sound reasonable? That then means that each property can be updated independently (otherwise we'd be sending back and forth a lot of unchanged information each time something updated). It does mean that as we create and destroy layers we will need to create and destroy topics, and that we'll have to make sure we keep our naming conventions clear.

Syntactically there might be a much better way to achieve the `RemoteTopic` functionality than I am doing right now, so curious if @jni has any ideas. Also right now each process (local and remote) has both a pub and a sub to the same topic, so whenever one side does a put it also triggers a get on the same side. That won't then actually trigger much more, but it might not really be how people normally use pub and sub.

Finally we'll also want to make it possible to connect our console to the remote console (or potentially expose both the local and remote console as tabs) if someone provides an address like that too, but that is independent from how we synchronize the viewers.


sofroniewn:
I'll note that [Quamash](https://github.com/harvimt/quamash) provides a `QEventLoop` compatible with asyncio. I tried it out, but also couldn't get it to fix all the problems. I'm also reluctant to add a new dependency if we don't absolutely have to (and it seems like some of this stuff can be done without Quamash), but I wanted to put it here for reference as it certainly is relevant.

jni:
There's a fork of Quamash that is updated for more recent Qt/PySide: https://github.com/gmarull/asyncqt
   
Found here: https://bugreports.qt.io/browse/PYSIDE-668

jni:
@danielballan 👋 

@sofroniewn is having a bit of trouble integrating Qt and Jupyter event loops with dask distributed's async/await code. None of us have coded with async stuff before so I was wondering if you could offer some help/guidance? I remember you were doing cool async stuff and iirc you were also doing this with a GUI?

btw @sofroniewn I forgot to mention in the meeting, the *only* time async stuff has made sense for me was reading Nathaniel Smith's Trio tutorial: https://trio.readthedocs.io/en/stable/tutorial.html But I read it a long time ago and I've forgotten it all! 😅 But it might be useful for you now.

Also, I only just now remember, @Carreau did some very cool work with async/await in the IPython REPL and that might be the source of some of the weird behaviour you were seeing in Jupyter, see:
https://ipython.readthedocs.io/en/stable/interactive/autoawait.html
and
https://blog.jupyter.org/ipython-7-0-async-repl-a35ce050f7f7

sofroniewn:
So looking into this a bit more having just one master `pub` / `sub` called `viewer` should be possible and then syncing that state. It is also possible to wrap dask arrays in a `scatter` call https://docs.dask.org/en/latest/futures.html#distributed.Client.scatter which then needs to be `gathered` on the other end. For example if I create the following "fake viewer"

```python
img = np.random.random((20, 20, 5))
big_img = da.random.random((1000, 1000, 1000))
future_big_img = client.scatter(big_img)
shapes = [np.random.random((6, 2)), np.random.random((5, 2)), np.random.random((7, 2))]
layers = [{'name': 'image', 'colormap': 'red', 'ndim': 3, 'data': img},
          {'name': 'big-image', 'colormap': 'blue', 'ndim': 3, 'data': future_big_img},
          {'name': 'shapes', 'colormap': 'gray', 'ndim': 2, 'data': shapes}]
viewer = {'title': 'napari', 'ndim': 3, 'layers':layers}
```

I can then send that to my other client using `pub.put(viewer)`. On the other end I then have the viewer dictionary, and can get access to the big dask array by
```python
results = client.gather(viewer['layers'][1]['data'])
```

I need to make it simple to generate the serialized version of the viewer on the remote, and I need to make it easy to update the actual viewer locally when receiving such a serialized version. I think we will also need to do some diffing as I don't want to send the whole viewer - arrays included(!) - every time a property like colormap changes.

The event-loop problems as described above still remain

tlambert03:
this is gonna be so cool.  🆒

sofroniewn:
yeah, will have to think carefully about how this model extends to new UI additions, like the histograms in #675 or any plugin stuff though and make sure it is not too brittle

sofroniewn:
Another advantage of using the `scatter` call is that now all arrays are hashed and so I can just use [`jsondiff`](https://github.com/xlwings/jsondiff) do to the diffing - which works really well.

If we are only sending the diff then I'm still not 100% sure though that I will want to generate the full viewer state dictionary and take its diff against the previously sent viewer each time a button is pressed - I know which button got pressed, so I might as well just send the update for that button. If I'm sending the whole viewer then doing the diff after it has been received makes sense because I definitely don't want to update everything (although most of our properties return immediately if you try and set them to be the same value, so maybe it is faster just to run through all of them and set them rather than do the diff in this case too)

Carreau:
> Also, I only just now remember, @Carreau did some very cool work with async/await in the IPython

Yeah, there is weird stuff in IPython. Also note that IPython cli and Ipykernel will behave quite differently with eventloops; so it may be worth testing with both when you have a weird edge case. 

The reason is ipykernel will also have an eventloop running (tornado) while IPython CLI may start/stop eventloop in between actual code execution.

Also a remote napari would be supercool, and I can already see the use case.

sofroniewn:
For anyone wanting to learn more about the dask pub/sub system I'm trying to use here @mrocklin just made this great tutorial video that I highly recommend checking out! https://www.youtube.com/embed/Q-Y3BR1u7c0

henrypinkard:
@sofroniewn I'm very interested in using this functionality and would be happy to do some testing

sofroniewn:
> @sofroniewn I'm very interested in using this functionality and would be happy to do some testing

Sorry I seem to have missed this!!

It's been a long time since this PR was created and our thinking on how to do "headless" mode in napari has improved (we now have better qt separation) and we have some additional async work now that we didn't before. I'm going to close this as we're unlikely to pick up this PR as is, though we do still have work to do to improve the overall remote use case

PR #658, opened 2019-11-05
Update Cirrus OSX image and patch windows builds

tlambert03:
# Description
This PR bumps the Cirrus OSX image from `high-sierra-xcode-9.4.1` (which is now deprecated) to `mojave-xcode-10.2`.  The tests are all fine, but @kne42 may have additional thoughts on whether we should avoid that update?

Also, as mentioned in https://github.com/napari/napari/pull/551#issuecomment-549526680, this adds `qt_no_exception_capture = 1` to the pytest settings in `setup.cfg`, which fixes the Windows builds.  I tested to make sure that I could break the windows builds with some bad napari logic, and I could... so I think it's probably better than having always fully-broken tests, but we should be vigilant about the possibility that it's masking errors.  (an alternate possibility, as @jni pointed out, is just that [pytest-qt exception capturing](https://pytest-qt.readthedocs.io/en/latest/virtual_methods.html) is broken for virtual methods).

## Type of change

- [x] CI-fix (non-breaking change which fixes an issue)

# References
https://pytest-qt.readthedocs.io/en/latest/virtual_methods.html
https://cirrus-ci.org/guide/macOS/

# How has this been tested?

## Final checklist:


sofroniewn:
So exciting to see all our tests pass again and get code cov back again!!!. Is there anyway to disable this only on windows though?

I'm worried about what bugs might be missed by this, can you explain a little about what you think will / won't get caught within the napari codebase (like our view files?) or qt / openGL / vispy by setting this flag?

Does this also have implications for the tests in your animation PR - #607?

The OSX update sounds fine to me, but we can always revert if we run into problems.

tlambert03:
> Is there anyway to disable this only on windows though?

I changed it so that only windows has `qt_no_exception_capture = 1`, and the rest of the tests are untouched.  The cost was a little bit of duplication of `install_script` and `check_script` in  `.cirrus.yml`, (I checked at cirrus to make sure there was no other way: cirruslabs/cirrus-ci-docs/issues/506)... but now windows works 🎉

>can you explain a little about what you think will / won't get caught within the napari codebase (like our view files?) or qt / openGL / vispy by setting this flag?

I don't have a complete answer to what will or won't get caught, but from the [pytest-qt docs](https://pytest-qt.readthedocs.io/en/latest/virtual_methods.html), it sounds like the risk is missing some errors in Qt virtual methods (like `mouseReleaseEvent`) that we have overridden:

> if python code in Qt virtual methods raise an exception PyQt4 and PySide will just print the exception traceback to standard error, since this method is called deep within Qt’s event loop handling and exceptions are not allowed at that point. In PyQt5.5+, exceptions in virtual methods will by default call abort(), which will crash the interpreter... To make testing Qt code less surprising, pytest-qt automatically installs an exception hook which captures errors and fails tests when exceptions are raised inside virtual methods.

(It's possible their exception hook itself is broken).

In my opinion, it's worth it: we are likely to see those errors in virtual methods anyway in the 'nix tests, and currently we basically miss _all_ of the windows tests since they always fail.  This at least has the potential to catch a few more windows bugs, we get code coverage back, and the whole system of github ✅❌icons will be useful again in our PRs

> Does this also have implications for the tests in your animation PR - #607?

It's conceivable.  I haven't tried those tests again with these changes, but they are Good & Fixed ™ now anyway so it shouldn't matter either way.

jni (approved):
I ❤️ it!

sofroniewn (approved):
This is amazing, great job isolating the change to the windows test only. I will merge now. ✅forever!!!

sofroniewn:
@tlambert03 - do you have thoughts on the discussion happening here https://github.com/vispy/vispy/pull/1746 seems very relevant to our windows / CI testing

tlambert03:
oh actually, I do sort of have thoughts... but I'm not that versed in GL contexts and stuff, so some of this is above my head still.  What I would add is that I recently tried to remote into a linux box using x11 forwarding to run napari, and I got the exact same error that we saw in the windows tests (but running it locally on that computer worked fine).  So I no longer think it has anything to do with pytest-qt virtual methods exception handling, and I do think it has more to do with virtualization.  It seems likely that the virtual display with x11 forwarding could be similar to the VI on cirrus, and lack a GL version as is described here.  ... but that's about where my helpfulness ends 😜
from some of older commit notes I've come across here it seems like @kne42 might have some experience with this sort of thing?

PR #664, opened 2019-11-08
modify dims QScrollbar with scroll-to-click behavior

tlambert03:
# Description
As discussed in #607, this PR changes the behavior of the dims QScrollbars such that when the user clicks on a position in the background area, the scrollbar moves all the way to the clicked position (rather than one "page" in the direction of the click).  The default behavior is [baked into Qt](https://doc.qt.io/qt-5/qscrollbar.html#details), so I borrowed an approach from [this (relatively old and not very popular) stack overflow post](https://stackoverflow.com/questions/29710327/how-to-override-qscrollbar-onclick-default-behaviour) to extract the position from the `mousePressEvent` event and convert it into a `slider.setValue`

ezgif-3-954ab2da6bb0

## Type of change

# References
https://doc.qt.io/qt-5/qscrollbar.html
https://stackoverflow.com/questions/29710327/how-to-override-qscrollbar-onclick-default-behaviour

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/_qt/qt_dims.py
+++ b/napari/_qt/qt_dims.py
@@ -1,11 +1,13 @@
+from typing import Optional, Tuple
+
 import numpy as np
 from qtpy.QtCore import QEventLoop, Qt, QThread, QTimer, Signal, Slot
-from qtpy.QtWidgets import QGridLayout, QScrollBar, QSizePolicy, QWidget
+from qtpy.QtWidgets import QGridLayout, QSizePolicy, QWidget
 
 from ..components.dims import Dims
 from ..components.dims_constants import DimsMode
 from ..util.event import Event
-from typing import Tuple, Optional
+from .qt_scrollbar import ModifiedScrollBar
 
 
 class QtDims(QWidget):
@@ -285,7 +287,7 @@ class QtDims(QWidget):
         range = (range[0], range[1] - range[2], range[2])
         point = self.dims.point[axis]
 
-        slider = QScrollBar(Qt.Horizontal)
+        slider = ModifiedScrollBar(Qt.Horizontal)
         slider.setFocusPolicy(Qt.NoFocus)
         slider.setMinimum(range[0])
         slider.setMaximum(range[1])
--- a/napari/_qt/qt_scrollbar.py
+++ b/napari/_qt/qt_scrollbar.py
@@ -1,0 +1,58 @@
+from qtpy.QtWidgets import QScrollBar, QStyleOptionSlider, QStyle
+from qtpy.QtCore import Qt
+
+
+# https://stackoverflow.com/questions/29710327/how-to-override-qscrollbar-onclick-default-behaviour
+class ModifiedScrollBar(QScrollBar):
+    """Modified QScrollBar that moves fully to the clicked position.
+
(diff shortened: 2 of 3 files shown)


sofroniewn (commented):
Looks amazing - checked it works in both normal and play mode.

Minor question - right now I can't click and drag to continue moving the slider from the new position. The additional drag has no effect. If that is easy it might be nice to support that, if it is very hard then might not be worth it. Does that make sense? What are your thoughts @tlambert03?

tlambert03:
> Minor question - right now I can't click and drag to continue moving the slider from the new position. The additional drag has no effect. If that is easy it might be nice to support that...

Yeah, I noticed that too, and that was exactly what the [original poster of that SO topic demanded](https://stackoverflow.com/questions/29710327/how-to-override-qscrollbar-onclick-default-behaviour#comment47559337_29711676) in order to accept the answer!

I too would prefer that behavior, but it might be tricky.... need to think about it

tlambert03:
I changed it so that a `mouseMoveEvent` while left mouse button is down will also move the slider, which mostly accomplishes the desired effect.  The one slightly annoying thing remaining is that if your drag event leaves the bounds of the slider, it no longer triggers the `mouseMoveEvent` and therefore stops updating the slider (which is a counter to what people expect on, for instance, OSX)... see gif below:

ezgif-3-5a6f7c133fd3

to fix that, the first thing that comes to mind is somewhat invasive and would amount to raising the `mouseMoveEvent` to the whole window and basically saying "if the mouse is down, and it started with a click on some slider, then move that slider to the horizontal component of the current mouse position"

sofroniewn (approved):
Amazing @tlambert03 - i did indeed notice that with moving the mouse outside the slider while dragging, but I consider that I minor inconvenience, that is relatively easy to understand. I wouldn't rearchitect all the Qt interactions at this point to solve it, though if people seem to really be struggling with this we can revisit.

I will let @jni have a look and weigh in if he wants. Otherwise I will merge tomorrow.

Thanks for the great work again!

tlambert03:
actually... I just realized that the normal QSlider _does_ display the desired behavior on a click and drag (i.e. it doesn't care if the mouse leaves the bounds of the slider), but it of course assumes that you started the click on the slider handle, and not somewhere else on the slider background.  Given that, I see the behavior of this PR as an unacceptable loss in behavior, but it also gives me hope that it shouldn't be too hard to fix it.  So don't merge until I fix that.

tlambert03:
figured it out, just needed to return the `super().mouseMoveEvent(event)` after intercepting it.  It's quite fun to use now!  Added a test, ready for re-review.

sofroniewn:
Nice, yeah, this is great to use. Good to know about that `super` trick. I will merge soon unless I hear otherwise.

sofroniewn:
I'm going to merge this now so it is in the `0.2.5` release

jni:
@tlambert03 @sofroniewn thanks for this change! It's beautiful!

jni:
btw @tlambert03 are those images open-license/available somewhere? They are gorgeous. 😍 They would be nice to use in demos/tutorials.

Issue #665, opened 2019-11-08
numcodecs update breaking builds


tlambert03:
Just noticed our OSX and Windows builds are failing again... At first I was worried it was related to our recent CI updates in #658, but it looks like it's a failure to compile `numcodecs` v0.6.4 which was released today.  We could pin to 0.6.3, or just keep an eye on it

sofroniewn:
It looks like 0.6.4 numcodecs is [failing to build on OSX](https://cirrus-ci.com/task/5355480448237568?command=install#L435) and windows. I have now raised an issue on the numcodes repo https://github.com/zarr-developers/numcodecs/issues/210. @jakirkham might also have ideas.

jakirkham:
Do you have a C++ compiler on the CI? It appears that the build is complaining that it cannot find one.

Alternatively if you want to avoid building Numcodecs, you are welcome to use our binaries from conda-forge.

sofroniewn:
I'm afraid I don't know. Thanks for weighing in. I defer to @jni for next steps.

Does it make sense to you that the change from `0.6.3` to `0.6.4` would trigger this build error?

jakirkham:
We upgraded Numcodecs' vendored copy of Blosc in this release. Blosc vendors many more compression libraries, which it updated as well. So the odds for some sort of surprise like this are quite high. Interestingly the Blosc upgrade was made to fix some compilation issues.

Note: The CI appears to have failed while building Snappy.

AhmetCanSolak:
btw `numcodecs` also a problematic package for bundling with pyinstaller(I know due to Blosc it is there). I appreciate at crazy levels if any of you guys manage to drop it at some point :)

jakirkham:
@AhmetCanSolak, can you please raise an issue on the Numcodecs issue tracker with more details about why it is difficult to include in PyInstaller?

tlambert03:
Just had someone someone email me directly because they tried to `pip install napari` but it didn't work.  So anyone who tries to install napari as described in the readme (and who doesn't have the appropriate build tools on their computer) is going to have this issue at the moment.  For this person, they had to either `pip install numcodecs==0.6.3 napari` or `conda install numcodecs; pip install napari`

AhmetCanSolak:
@jakirkham i will drop and issue once i get back to bundling again.

@tlambert03 you are right, tested `pip install napari` and crashed again related to numcodecs. Tested in a fresh conda environment for py3.7

sofroniewn:
[...]

PR #666, opened 2019-11-09
Avoid numcodecs 0.6.4 for now

jni:
It appears to not be trivial to compile numcodecs 0.6.4 on Mac or
Windows, which results in failed `pip install napari` for users. See
https://github.com/napari/napari/issues/665 for more details.


sofroniewn (approved):
Looks good to me, install is passing again. I will merge and make `0.2.5` release

PR #670, opened 2019-11-09
allow all animation thread tests to be +/- 1 frame

tlambert03:
# Description
I had hoped that my revised `play_axis` tests in #607 would allow slightly stricter testing on the animation function.  But another failure [leaked through](https://cirrus-ci.com/task/6145843798736896?command=test#L452), so this PR relaxes the expectations to +/- 1 frame.  That should be enough, I but can relax more in the future if need be.

## Type of change

- [x] Test-fix (non-breaking change which fixes an issue)


sofroniewn (approved):
This has all passed, so let's get it in and we can then revisit again if needed, thanks!

Issue #516, opened 2019-09-07
1D rendering


sofroniewn:
## 🚀 Feature
As originally laid out in #9 it would be nice if the viewer can move between different rendering modes - 1D, 2D, and 3D. We've made good progress on 2D and 3D and so I'm making this issue to discuss 1D rendering.

I think the goal with 1D rendering is not to become a full 1D plotting library, there are many of those, but it is more to be able to expose functionality that people may often want at the same time as they are viewing images.

For example when given a 3D volumertic image you might want to look at the a 1D z-profile through a single pixel or the mean or the max of each slice.

Or if given a 2D timeseries you might want to look at the timeseries as a mean of an ROI defined through a label as requested in #112 and #326.

I don't think napari should necessarily be providing means to do all this re-slicing etc itself, that could be the job of plugins, but it would need to provide the ability to display the 1D trace, which is something vispy can easily do.

Plotting 1D traces though could quickly explode in complexity. Would we support a new layer type called `Trace(s)` that could support one or many lines given as `x` and `y` coordinates either overlaid of exploded? Would we support only the ability to take our existing layers and move them to a 1D view, possibly with some processing along the way?

What are thoughts @royerloic @d-v-b @jni?

sofroniewn:
Also note #326

royerloic:
We could start witha  1D mode that is just a natural extension of 2D and 3D: we just show one dimension and have sliders for the rest. this can be useful when considering for example 2D kimograph data to go back to see the 1D profile, etc...

But, of course, what would be interesting would be able to link multiple canvas that display different aspect of a dataset. For example, we could draw a line and display a line profile through all layers. or we could set a plane that slices a 3D volume, etc... This is quite advanced, and it is not clear what is really needed or how to best build that functionality. So probably the safest approach is to build the features verry slowly in that respect -- hence the point above.

I think the multi-canvas issue and linking displays of different (or same dimensions) -- for example linked orthogonal 2D views (XY, XZ, YZ) etc... should probably a different issue separate from 1D rendering...

What do you  guys think?

PR #675, opened 2019-11-10
[WIP] Histogram with 2-way LUT control

tlambert03:
# Description
Still a lot that needs to be done, but I wanted to show a quick example of a working histogram widget.  The widget is implemented using only vispy and Qt, and has 2-way data-binding with the `contrast_limits` and `gamma` sliders of a given layer.  I know there's been discussion about how/where this sort of thing would go (drawer? dockable? etc...), and this widget is agnostic to those implementations.

ezgif-4-78395626efaa

## Type of change

# How has this been tested?

- [ ] need to write tests

## Final checklist:


sofroniewn:
This is phenomenal @tlambert03!! Very nice to see the histograms / line plots done with vispy, this makes me very excited about #516 and supporting more general 1D plotting in the main canvas.

We're also getting more and more requests for supporting multiple vispy canvases simultaneously, either for linked ortho-views, or a linked grid mode, or link image + timeseries displays. If we go down that road does it sense then to consider this histogram view as a special case of some of that stuff and display it in a second canvas next to / under the main image, rather then say in some entirely separate window.

I wouldn't want to hold this PR up unnecessarily though, and starting with it in a separate pop-up window might be the easiest, but curious about your thoughts around multi-canvas mode

jni:
Just dropping by to say 😍

tlambert03:
> This is phenomenal @tlambert03!! Very nice to see the histograms / line plots done with vispy, this makes me very excited about #516 and supporting more general 1D plotting in the main canvas.

yeah, the canvas is pretty much a thin subclass on vispy's [`PlotWidget`](https://github.com/vispy/vispy/blob/6f4e52aeecc40c29bf2fde3d217c0a3e6ea6b438/vispy/plot/plotwidget.py#L12), which provides an example of linking the axes domains to the data and pan/zoom everything together.  So it's nicely set up for 1D data.  Part of what I want to do is refactor what I have here to make the histogram a subclass of a broader plot widget that could be easily reused.

> does it sense then to consider this histogram view as a special case of [multiple-canvases] and display it in a second canvas next to / under the main image, rather then say in some entirely separate window.

possibly... One of the reasons I made it as a Qt widget instead of just a vispy canvas is to add some interactivity features (like the mouse cursor changing when you hover over the gamma or clim handles).  I don't know how to do that in vispy alone (but it might be possible... or maybe the couple things requiring Qt aren't worth it and we can refactor).  That said, the multi-canvas thing could potentially just be implemented as a QGridLayout,  and it wouldn't really matter whether each item in the grid was a vispy canvas, or a QWidget (provided the widget had a black background and blended in seamlessly to the larger view)

> I wouldn't want to hold this PR up unnecessarily though, and starting with it in a separate pop-up window might be the easiest, but curious about your thoughts around multi-canvas mode

I do think it's worth thinking about up front.  If you guys feel like depending on Qt limits the flexibility of the 1D plotting "widget", let me know and I'll look into keeping it strictly as a vispy canvas. (at the possible cost of a bit of interactivity)

jni:
> If you guys feel like depending on Qt limits the flexibility of the 1D plotting "widget", let me know and I'll look into keeping it strictly as a vispy canvas. (at the possible cost of a bit of interactivity)

I don't have a *strong* opinion about this, but imho having it be a Qt widget makes sense. I think @royerloic would prefer a drawer to a pop-up, but other than that it makes sense to have it be a Qt Widget right now. Let's just be careful to have a separate model with the required data, and the view/controller Qt stuff is fine.

tlambert03:
> I think @royerloic would prefer a drawer to a pop-up, but other than that it makes sense to have it be a Qt Widget right now.

yeah I'd much prefer a drawer as well.  I think napari should generally avoid pop-up glut.

jni:
> I think napari should generally avoid pop-up glut.

It's one of our founding principles. =D

tlambert03:
> It's one of our founding principles. =D

you have all the best founding principles 😉

sofroniewn:
I've got no problem with using the `QWidget` if that's a better way to get more interactivity than just the vispy canvas as I know people will want interactivity with this stuff (e.g. link up rectangle from a shapes layer over a 3D image layer to a 1-D line plot to show the average of the rectangle in the line plot as you move the rectangle, requested by @d-v-b). 

I think we will also want to make sure that we keep some sense of model / view separation though, and maybe having some model of the histogram that has no `vispy` or `Qt` in it would be good, like we have done with our layer models.

Using the `QGridLayout` for the multicanvas stuff sounds good too, as does using a draw.

royerloic:
Great work @tlambert03 !

second @jni that we want to avoid popups and/or extra windows,
but this is of course a first attempt and so it is ok to temporarily
violate a founding principle ;-))

A few comments  and ideas:
i) Our earlier thoughts for the histogram was indeed to have a drawer, the question is where... Most screens have strong landscape aspect ratios, so I would suggest to have the drawer on the right side, with the histogram rotated 90 deg.
ii) The histogram could slide open automatically as you change contrast and gamma parameters.
    (just throwing the idea out, not sure about it). 
iii)  The LUT colors could be shown as a colored bar that spans the contrast interval on the Y axis, giving a clear readout from LUT value to values on the axis.
iv) Linked views can be a way to implement this that would help us set up more generic
machinery for linking 2D, 3D, line profiles, histograms and more...
v) I think the vispy multi-canvas might be too brittle and limited, we will have more freedom
and flexibility with multiple widgets.
vi) The axis and margins of the histogram might be a bit wastefull long-term, we might
have to tune that to have a more compact layout.
vii) Gamma and contrast-limits are a good part of the 'model' here, the histogram being 'linked-data' that should be updated at appropriate occasions.

tlambert03:
Thanks @royerloic for your helpful input!  I added a vertical layout flag to the histogram... and will definitely add a LUT bar soon.  I tried to tighten up the margins as much as possible; the layouts are pretty flexible and resize well to whatever space we put them in (the vertical orientation still feels a bit weird but that can change).  I also added some logic to link the histogram model to the `viewer.active_layer` so that it links and unlinks from the layer events upon switching.  You can also have it calculate the histogram just on the current view slice (which will obviously be necessary for very large data): you'll see that towards the end of the gif, where the bottom widget is in `view` mode and the vertical histogram is in full stack `data` mode.

ezgif-4-7f15f7ee92f0

I also started to abstract some of the ideas into a more general `QtPlotWidget` today - trying to follow your general MVC pattern - and I think it's got potential, but it's also possible that I've over-complicated things and/or started wandering into territory for which you already have a clear vision (like your plugin architecture), so I would love your thoughts/criticisms/warnings on it when you have time...  Here's what I'm currently thinking/doing for "plottable" things:

1. some underlying analysis model (here, `components.histogram.Histogram`) has a `set_data` method and does something interesting with it.
2. some `VispyPlotLayer` is responsible for linking a napari layer with the analysis model, (feeding it either some/or all of the data) and turning the output into a vispy node that can be plotted in the `NapariPlotWidget` (which is a gridded vispy scene that has `AxisWidgets` and a main `ViewBox` ready for plot nodes).
3. The `QtPlotWidget` wraps the `PlotWidget`, and has a convenience methods (like converting between plot coordinates and window coordinates on mouseover or resize).   Subclasses of it like `QtHistogramWidget` add `VispyPlotLayer`-specific functionality.  (It's also possible that we don't subclass the `QtPlotWidget`, but rather use composition somehow to deliver and remove plot nodes from `QtPlotWidget.plot.view`).

tlambert03:
cross-posting this gif here to keep the progress in one thread. 
the hard part was definitely done with the histogram, and creating the line profile widget was relatively trivial

ezgif-4-b18e9bf869c8

sofroniewn:
This is really looking incredible!!! Lots of people are going to find functionality like this very useful.

I have a relatively high-level question - do we want to think about the generic `Histogram` and `LinePlot` (name?) as layers that can be added to the viewer rather than components? I get that we might want to treat the contrast_limits histogram as a slightly special case, but for something like the gif with the cat above does it make sense to have 3 layers - `Image`, `Shapes`, `LinePlot` and then 2 canvases in the viewer and have `Image` and `Shapes` in canvas 1 and `LinePlot` in canvas 2 and then link them as needed?

This PR definitely shouldn't take on the multicanvas stuff, but I wonder if we shouldn't be thinking of these things more like `Layers` than just `Components`. Thoughts @tlambert03?

tlambert03:
yeah, there's lots to discuss here, and I still haven't hit on a strong opinion about how best to incorporate these things.  I'd like to try to distill what I think is actually being added in the this PR, and then get all of your opinions about where it belongs and whether they are layers or components.  And honestly, I wouldn't mind getting a little review about how you've been using those terms.

I think the core addition here is probably the [`NapariPlotWidget`](https://github.com/tlambert03/napari/blob/histogram/napari/_vispy/vispy_plot.py), which is really just a modified version of the `vispy.PlotWidget` (I will explain rationale for seeming duplication elsewhere/later).  That's most like a (strictly 2D) layer that could be dropped anywhere. My [`QtPlotWidget`](https://github.com/tlambert03/napari/blob/histogram/napari/_qt/qt_plot_widget.py) is an example of creating a fully independent `QWidget` with it's own canvas that holds the `NapariPlotWidget`, and the [`QtHistogramWidget`](https://github.com/tlambert03/napari/blob/histogram/napari/_qt/qt_histogram.py) is an example of subclassing the `QtPlotWidget` with additional events and interactions... but we could just as easily treat that `NapariPlotWidget` as a layer and put it directly into a canvas adjacent to the main canvas in the viewer.  (though I definitely do think it needs an independent vispy scene and camera.)

We might want to share a screen to discuss this a bit more interactively, but it's safe to say that nothing I've done here is making any great assumptions about how it might be incorporated into napari.  All possibilities are definitely very much on the table and I'd follow whatever you advise.

jni:
Side note: @sofroniewn, I feel like drawing a rectangle for the line selection isn't great. Would you want to look at that to make line selection points always lie just along the line?

Regarding the widgets: I don't think these things belong on the canvas, but as we think about doing arbitrary numbers of these things, I think we need to add standard areas for them (drawers along left, right, and bottom of the window?), and then different widgets (plugins!?) can take them over for their purposes. I *don't* want to promote an "ecosystem" where you just have this tumorous mass of widgets growing on top of each other.

It doesn't make sense to me to think of them as layers, but maybe I'm missing something.

This seems like a great topic to discuss and brainstorm during the meeting!

Anyway, the whole thing is very exciting. =D

sofroniewn:
I think this comment by @tlambert03 

> (though I definitely do think it needs an independent vispy scene and camera)

makes me less inclined to think about these 2D plots as layers because you fundamentally cannot overlay them with our other layers and do blending etc / 3D rendering and have everything just work. This suggests that they are more like "components" that deserve their own treatment.

A different view would be though that the line plot is really just a 1D image layer, and that points could have a 1D version, as could surface / shapes etc, so that you could click a `1D` button and the view would switch to `1D` rendering, rather similar to how you can now click a `3D` button and switch to `3D` rendering

royerloic:
> This is really looking incredible!!! Lots of people are going to find functionality like this very useful.

I agree, fantastic work @tlambert03 !

> 
> I have a relatively high-level question - do we want to think about the generic `Histogram` and `LinePlot` (name?) as layers that can be added to the viewer rather than components? I get that we might want to treat the contrast_limits histogram as a slightly special case, but for something like the gif with the cat above does it make sense to have 3 layers - `Image`, `Shapes`, `LinePlot` and then 2 canvases in the viewer and have `Image` and `Shapes` in canvas 1 and `LinePlot` in canvas 2 and then link them as needed?
> 
> This PR definitely shouldn't take on the multicanvas stuff, but I wonder if we shouldn't be thinking of these things more like `Layers` than just `Components`. Thoughts @tlambert03?

I actually like the idea of drawing histograms and line plots as layers:
Using the grid view you can nicely show the image and the histogram and lineplots
all together side-by-side and and have everything be updated.
Bute, the issue perhaps is that then histograms and lineplots are not first class citizens
and it might feel a bit as a cheap hack... 

Another possibility following some proposals already mentioned here, is to have _one_ 'drawer' for 1D plots, that is used in a context-dependent manner to display either a histogram or a line plot, or anything else that is 1D or histogram-like. I think it would be important to have reusability of that component for different purposes because otherwise we are quickly going to run out of space and it is going to get messy with tons of drawers popping out from all sides...

@tlambert03 : to answer your question, I think having a generic MVC architecture for handling histograms and lineplots is a great idea, your suggestions seem reasonable. A few thoughts:
(I am dumping thoughts here, brainstorming freely...)

We need to think hard about how does line profiles know about which layer to look at for both the image and the line... We might want to think about a generic notion of an 'interactive_analysis' module that can be attached to a viewer and that can analyse the layers in a viewer, do something and then populate a 'plot-drawer'...  Another possibility, in the case of line profiles, is to say that 'line profile' is a 'special kind of layer' with just one line, that can send lineplots to the 'line-plot-drawer'. You can add multiple line profile layers (move the corresponding lines) and inspect each profile by selecting it. In this case it makes sense to have a layer because we anyway need to display a line shape, and so why not reuse what we have... This would suggest that all layers (views) have the possibility to draw in the standard 1D plot drawer.

Of course it might not make sense to show multiple line plots in the 'drawer' so everytime you select a new 'line-profile-layer' it shows a different profile.
Note: if you have 10 images layers and 1 line profile layer, when you selected the line profiler layer the 1D plot gets updated with _10_ line plots for all image layers... Same thing with the FRC, you could see the correlation profile for all 10 images and compare, for example resolution between the images. Extremely powerful...

In the case of histograms, as soon as you play (or click, or hover) on the contrast limits the drawer shows the histogram (if the drawer is open). Perhaps we can have a 'stack' structure for what is shown in the drawer: clicking on a 'pop' button in the drawer would delete the last (interactive) plot added to the drawer, revealing what was there before: Imagine you are doing some line profiling and then you adjust the contrast, you can find it again. but ok, you can also simply click in the layers list on the 'line profile layer' -- maybe more explicit.... 

There  are other cool things that we could do apart from displaying an histogram, and that would benefit from a modular and generic framework: displaying the Fourier Ring Correlation of an image as a 1D plot in the same 'drawer'. I can imagine quite a few other applications in which we want to have 1D line plots or even scatter plots  interactively synced with the layers. 

Having the possibility to 'tear-up' the line plot drawer from the viewer window to have it as a _single_ separate window (if you close it goes back as closed drawer...) would be cool in the case that one needs to look at the line plots with greater detail and space...

At this point we have to identify a reasonable not too complex path to get most of the functionality we need.

tlambert03:
I think I agree with pretty much everything being said here (sounds like everyone is more or less hovering around the same ideas... which is nice 😄 ).  I also like the _one_ special drawer for now that can be shared in a context dependent way for plotting 1D data.

I looked around for a while for a way to slide out a nice drawer in Qt, but I didn't find anything I loved.  So I played a bit with the QDockWidget idea, which accomplishes the "tear-away" behavior nicely, and lets you drag and drop the plot window on either side (later, we can add [QPropertyAnimations](https://doc.qt.io/qt-5/qpropertyanimation.html) if desired to add "slide" behavior that a drawer would have).  Here's an example with a histogram, and the histogram changing to a vertical orientation when you move it to the right side.  (it's a bit slow when switching at the moment, and i recognize the orientation of the histogram is a bit weird, but just a proof of principle for you to evaluate)

Untitled

royerloic:
> I think I agree with pretty much everything being said here (sounds like everyone is more or less hovering around the same ideas... which is nice 😄 ). I also like the _one_ special drawer for now that can be shared in a context dependent way for plotting 1D data.
> 
> I looked around for a while for a way to slide out a nice drawer in Qt, but I didn't find anything I loved. So I played a bit with the QDockWidget idea, which accomplishes the "tear-away" behavior nicely, and lets you drag and drop the plot window on either side (later, we can add [QPropertyAnimations](https://doc.qt.io/qt-5/qpropertyanimation.html) if desired to add "slide" behavior that a drawer would have). Here's an example with a histogram, and the histogram changing to a vertical orientation when you move it to the right side. (it's a bit slow when switching at the moment, and i recognize the orientation of the histogram is a bit weird, but just a proof of principle for you to evaluate)
> 
> Untitled

This is great. Love it. As long as we have only one of these around that's great.
We would need a button in the viewer to  reveal it  etc...
Great direction!

sofroniewn:
@tlambert03 awesome demo - as I noted above, we had previous discussed making the layers list / layer control QDockWidgets too, potentially separate ones, so you could have a layer list that was the full height of the screen if you wanted.

I'm not 100% about only limiting to one of these, or if we do limit to one - do we support the concept of multiple widgets inside it, for example if I want to plot multiple 1D lines are these multiple plots inside our one "draw" or are they multiple moveable widgets like this?

I probably lean away from multiple dockable widgets / pop-ups and instead favor more of a multicanvas approach, but then I think we should have two slots on the viewer one horizontal only under the dims slider, and one vertical, pull length and you can how a 1D plots in either / or both of them and they might not even have to be dockable - for example a very crude mock of having both enabled:

image

We'd make places for both of them in the viewer QGridLayout https://github.com/napari/napari/blob/8ce944a1dff680f4d06b238f7556211550ba1d4b/napari/_qt/qt_viewer.py#L90 The console if turned on would go underneath all of this.

tlambert03:
just an update here...  I haven't stopped thinking about this, but I'm struggling with a strange & difficult-to-debug problem as described in vispy/vispy/issues/1759.  Briefly, when the dockwidget is floated, visuals added to a `vispy.SceneCanvas` are not drawing and you get a bunch of GL errors.  Doesn't mean we can't do the histogram logic outside of a dockwidget, but that's the current holdup.

dongyaoli10x:
Hi @tlambert03, I am wondering if there are plans to further some of the good works you did here. Plotting infrastructure, including live plotting, would be very useful for the dock_widget plugins.

tlambert03:
hi @dongyaoli10x,
This is a tough one.  On the one hand, I really liked the functionality here (fast, opengl-based basic plotting provided by vispy, without any new dependencies)... but the real challenge, and the thing that eventually led us to delay this indefinitely, is the API.  I believe there is a slight fear that if we start providing convenience functions to build plots, then it's not clear where that "ends", and we don't want to duplicate plotting libraries.
That said, a lot of time has passed, I can still see a lot of usefulness coming out of provided some wrapper around the vispy PlotWidget, (like the [NapariPlotWidget](https://github.com/napari/napari/pull/675/files#diff-c1e44ab8f10d471bff02191e74d28e3116b3c4f4c9c0285d24489239173cdabbR7)) in this PR.  And with the magicgui-backed functional widgets now, I can also imagine supporting basic line, bar & scatter plotting based on function return annotations.
Anyway, it's a huge topic with a lot of potential (and a lot of work!).  Thanks for bringing it up again, curious what @sofroniewn and @jni think about this general area these days.

sofroniewn:
> Anyway, it's a huge topic with a lot of potential (and a lot of work!). Thanks for bringing it up again, curious what @sofroniewn and @jni think about this general area these days.

One option that we have available to us as of `0.4.3` it to start making a `napari-plots` plugin. It could be its own repo, could be backed by either vispy and/or matplotlib and/or other plotting libraries and leverage the `provide_dock_widget` hookspec. This approach would allow us to rapidly iterate (we could have low expectations on API stability there for a couple of releases even), would keep the plotting functionality separate from the image functionality.

If we love the direction its going then maybe we fold the repo into master or maybe just make it a required dependency.

> And with the magicgui-backed functional widgets now, I can also imagine supporting basic line, bar & scatter plotting based on function return annotations.

I love this idea too!!! Could be done in this new repo also.

jni:
Yep, I love love love the idea of a napari-plots plugin.

PR #683, opened 2019-11-11
ArcballCamera fov default

sofroniewn:
# Description
This PR hard codes the `fov=0` of the ArcballCamera camera, which is its current setting as this will change on an upcoming vispy release - see discussion here https://github.com/vispy/vispy/issues/1723 and here https://github.com/vispy/vispy/pull/1702#issuecomment-552609734. We might want to change to `fov=45` - I'm curious if it might make https://github.com/napari/napari/issues/673 better, but we will need to check some other things and I'd rather us explicitly define a value than depend on a default that may change.

## Type of change


jni:
@sofroniewn would you be able to test fov=45 to see if it helps, without making anything worse? Might as well check it now?

sofroniewn:
I just did some more checking and it doesn't make things better for #673 - though I'm not sure I understand #673 perfectly, but I now realize that it is more about transparency than ordering.

The problem with `fov=45` right now is our points sizes become really weird. We might want to revisit this in the future, but for now I think it is best to merge, especially as napari behavior will be different for people depending on what vispy they are on right now

jni (approved):
In light of #697, gonna merge this!

Issue #569, opened 2019-10-03
Labels painting slow for big images


sofroniewn:
## 🐛 Bug

The labels painting is now quite slow for big images, say 2048 x 2048. It's not clear if it has degraded recently - the undo / redo does seem to slow it down a little - but it may have always been that slow. We might need to think about how to make it faster. Another thing to add to our benchmarking list. #476.

sofroniewn:
We might have had a recent regression when we changes some of our openGL options to improve mesh rendering for the shapes / surface layer

sofroniewn:
Note that labels painting will likely be significantly speeded up by #204

sofroniewn:
Painting has been significantly sped up by #684, so I will close this now, but it can still be further improved by #204

VolkerH:
Can this be reopened? 

I just tried creating a labels layer with a size of `(1143, 1069)` on top of an RGB image and drawing annotations with napari from the current master.

When drawing with the brush I get a lag on the order of a second or even more.  I can't compare to the performance before merging #684 but it still feels very slow.  At this level it is too tedious to  label a significant part of an image for machine learning annotation.

Environment used

```
napari: 0.2.6+23.g8735eb2.dirty 
Platform: Windows-10-10.0.18362-SP0 
Python: 3.7.5 (default, Oct 31 2019, 15:18:51) [MSC v.1916 64 bit (AMD64)] 
PySide2: 5.13.2 
Qt: 5.13.2 
VisPy: 0.6.3 
NumPy: 1.17.3 
SciPy: 1.3.1 
scikit-image: 0.16.2 
Dask: 2.8.0 

GL version: '4.5.0 - Build 25.20.100.6472'
MAX_TEXTURE_SIZE: 16384 
```

sofroniewn:
We can re-open, I think this will hopefully be closed by #713 which will remove an expensive int -> float conversion step

liu-ziyang:
cc @VolkerH I wonder if you are still seeing performance issue with large image labeling? I believe 0.3/0.4 releases introduced some improvements to our labelling performance

VolkerH:
I haven't done extensive labelling recently, but I certainly noticed various impreovements since 2019.
I still think it is not quite as responsive as I would hope.

[...]

Issue #204, opened 2019-04-04
Render labels using a shader based on hash(label, salt)


perlman:
## 🚀 Feature
Implement the rendering of label layers as a gl shader. Use a hash based on the label value and a salt to pick a color at random.

## Motivation

The shader generated by vispy.color.colormap for a colormap with a large number of intervals would fail to compile due to an excessive code size. This is not an issue with the shaders for linear colormaps, but those do not clearly differentiate labels with close values.

By using a hash, _any_ label value an be mapped to a random color. A salt is added to shuffle where labels hash to. Also, this allows display of arbitrary labels with complete disregard for the set of global labels.

## Alternatives

We have already implemented a quick-fix to render labels with a large number of segments (#203) by mapping colors to a random label that is mapped to a random color. One downside of the current approach is two label values that map to very close each other will be assigned the same color

## Additional context

An example implementation of using a shader for colorization is in [this branch](https://github.com/perlman/napari/tree/eric-label-hash).

Neuroglancer uses this hashing approach to [assign segment colors](https://github.com/google/neuroglancer/blob/master/src/neuroglancer/segment_color.ts).

Predefined color maps, loaded as textures, could be used to provide discrete sets of colors for the random colors to be drawn from.

sofroniewn:
@perlman @jni - to follow on here there are a couple problems with the current fix.

The first one is that sometimes there can be a discrepancy between the displayed color and the color returned from `label_color` - these are likely do to in-precisions with floating points. For example if you run the `labels-2d.py` example and then set the labels layer to `0` I get a slightly off-green color, that is the color generated by passing `0.00001` to the colormap. Similarly the color corresponding to label `2` appears on the screen as a dark gray - but gives an orange when passed to the function. 

One side question I have about the layer in general is where does the colormap actually get passed to the vispy node? i.e. I am expecting a line like `self._node.cmap=colormap`, but all I can find is `self.colormap = colormap`

Finally the interactivity in the paintbrush in #195 is now very slow. If you use a simple colormap or set `num_colors=10` it becomes fast again.

sofroniewn:
I have been trying about visualizing some labeled EM data from [CREMI](https://cremi.org/data/) - one of the unpadded training sets - and have noticed that switching z-planes is very slow even when the data is fully loaded in memory. See the example gif below.
 
![CREMI](https://user-images.githubusercontent.com/6531703/56714678-6686c700-66ea-11e9-9cbc-be1a36ea45d0.gif)

[...]

PR #684, opened 2019-11-11
improve labels painting speed

sofroniewn:
# Description
This PR begins to address https://github.com/napari/napari/issues/569 by improving the labels painting speed - it does this by removing unnecessary calls to `_set_view_slice`. Fixes like #204 will improve this further, but this is already a large improvement, try it out when painting on a `(2048, 2048)` image, it goes from unusable, to not very good ;-) `(1024, 1024)` is now not too bad.
Note that this PR doesn't affect the speed of one individual `Paint` call, it just bundles the refreshing of many paint calls. It also adds a dedicated benchmark for the individual `Paint` call

### DIFF
--- a/napari/layers/labels/labels.py
+++ b/napari/layers/labels/labels.py
@@ -549,7 +549,6 @@ class Labels(Image):
                 interp_coord = interpolate_coordinates(
                     self._last_cursor_coord, self.coordinates, self.brush_size
                 )
-            self._save_history()
             with self.events.set_data.blocker():
                 for c in interp_coord:
                     self.paint(c, new_label, refresh=False)
--- a/napari/layers/labels/labels.py
+++ b/napari/layers/labels/labels.py
@@ -545,7 +545,6 @@ class Labels(Image):
             Vispy event
         """
         if self._mode == Mode.PAINT and event.is_dragging:
-            new_label = self.selected_label
             if self._last_cursor_coord is None:
                 interp_coord = [self.coordinates]
             else:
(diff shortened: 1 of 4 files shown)


napari/layers/labels/labels.py
```diff
@@ -547,9 +552,8 @@ def on_mouse_move(self, event):
                 interp_coord = interpolate_coordinates(
                     self._last_cursor_coord, self.coordinates, self.brush_size
                 )
-            with self.events.set_data.blocker():
-                for c in interp_coord:
-                    self.paint(c, new_label)
+            for c in interp_coord:
+                self.paint(c, new_label, refresh=False)
```
jni: @sofroniewn I'm having a bit of trouble interpreting the codecov UI, but, if I'm not mistaken, it shows that this loop is not covered by any tests:

https://codecov.io/gh/napari/napari/compare/6c5bb037e191bce3c05c5bf23f4f143ecc89a240...89db15b0491fc3f0322a4541eba773be441c12d1/diff

Could you verify that?

jni:
@sofroniewn this is super great! Could you post the benchmark results from this PR compared to master?

napari/layers/labels/labels.py
sofroniewn: that is likely correct, as it is inside our mouse callback. I can try adding test + benchmark before we merge

sofroniewn:
@jni I have now added a test and a benchmark `time_on_mouse_move`. I want to make a dedicated PR refactoring all our mouse events to use the new framework in #544 and tests for all of them, but that can come later (soon!)

To run this benchmark:
```bash
asv run -b benchmark_qt_viewer_labels.QtViewerSingleLabels.time_on_mouse_move --python=same
```

Before this PR / actually run with `refresh=True` using the new code the benchmark to paint a full line from top left to bottom right corner of a (512, 512) image is `2.63±0.04s`. Now with `refresh=False` that drops down to `33.4±2ms`, almost two orders of magnitude!

This PR should make a big difference, but probably still more on the table to grab!!

jni (approved):
Brilliant!!!

Issue #651, opened 2019-11-01
Introduce the concept of points having a "class" that can be changed in the GUI.


adamltyson:
## 🚀 Feature

Introduce the concept of points having a "class". Rather than calling `viewer.add_points` multiple times, points should have a label, which means they are rendered in different colors, and can be moved from one class to the other in the GUI.
## Motivation

I want to use napari to curate points, and generate training data. Assuming I have an image, and N classes of detected objects, I want to load them all and visualize them in different colors (maybe with a legend, `class1=blue` etc). Ideally I would then be able to curate these points, by selecting them, and converting them to a different class (or generate a new class).

I can already do this by calling `viewer.add_points` and adding and deleting points, but the concept of changing a class (rather than delete and add) would streamline the process (especially in conjunction with #474.

## Pitch

Assuming there is an NxD `points` array, I would like to be able to pass an Nx1 `classes` array to `viewer.add_points`, `viewer.add_points(points, classes=classes)`. These would appear separated in the GUI, perhaps as an expandable list within a single `Points` section. 

When selecting a point, there would be an additional option to `convert to class X`, and ideally a keyboard shortcut to the number of that class.

Even more ideally, the arguments to `viewer.add_points` should accept lists of length N (or Nx1 arrays) so we could use `viewer.add_points(points, classes=classes, names=["Cell type 1", "Cell type 2", "Cell type 3"])`.

## Alternatives

This idea may have already been proposed (but I couldn't see it).

The only alternative I can think of is to call `viewer.add_points` in a loop, and then deal with adding/deleting points outside of napari (which isn't ideal), and lacks a true "convert" option.

## Additional context

I think a point-based data curation interface like this could be very useful to the community. There are lots of point-based deep learning applications coming out that require training data (e.g. [DeepLabCut](https://github.com/AlexEMG/DeepLabCut), although they have their own training data generation interface).

sofroniewn:
@adamltyson I like this idea a lot, I want to think about it in the context of https://github.com/napari/napari/pull/552#issuecomment-536878528 where the concept of `coordinates and values` has been proposed. I like the idea of `values` / `class` just being a continuous or categorical descriptor.

I think @berl might have opinions on improvements to this api too.

Making sure use case like this are well supported is definitely a priority for us.

sofroniewn:
[...]

Issue #448, opened 2019-07-26
Support for 3D animation 


pranathivemuri:
## 🚀 Feature
3D animation

## Motivation

 It would be great to have an open-source viewer that displays 3D volume and you can fly through it and record as you do so or just have a 360-degree rotation of volume. 

## Pitch

1. Scripted animation - http://zulko.github.io/blog/2014/11/29/data-animations-with-python-and-moviepy/
2. Screen Record animation - idea from @kevinyamauchi To do this we save camera position, colorbars etc in a json and generate the 3D animation later 

## Details

We can either use ffmpeg or gif to record frames and generate the video.

bscott711:
Adding some sort of storyboarding to the animation process would be very beneficial especially as there are multiple frames and vantage points wanted for timeseries data.

sofroniewn:
I think it would be nice to be compatible with - https://www.nature.com/articles/s41592-019-0359-1 - there is a FIJI plugin https://github.com/bene51/3Dscript which we could maybe port to python. If anyone has any experience with this tool or knows the original developers please let us know. I have not used it, but it seems like people really like it, for example https://twitter.com/DougPShepherd/status/1190365654311792641

sofroniewn:
Cross posting a link to an amazing [naparimove repo](https://github.com/guiwitz/naparimovie) built ontop of napari by @guiwitz - originally posted in this [image.sc post](https://forum.image.sc/t/movies-with-napari/31588)

I was able to successfully use it to make movies right out of the box, and I’m curious from @guiwitz if there’s more stuff we need to expose better to make this easier for you, if you have feedback on our apis or napari in general, or if you’re interested in contributing any of this into the main napari repo.

I think we see support for high quality animations as something pretty fundamental and something we want to provide to all users.

guiwitz:
Hi @sofroniewn, I would definitely be enthusiastic at trying to integrate naparimovie directly into napari! I guess my main question is what form this integration should take. If I remember correctly there's the plan to have a sort of plugin system in napari. Should that be the way or should it be a standard feature of napari?

In both cases, I'd need some pointers on "where to put things", e.g where should I add entire classes as for example the one replicating 3Dscript, or where's the best place to add list of interpolated camera states etc. In the contribution guide I couldn't find guidelines on the general organisation philosophy of the code. Is that info available somewhere or can you summarise general principles?

On a more detailed level, I guess the most difficult part was handling getting and setting the camera state, in particular its rotation. A few comments:

[...]

PR #686, opened 2019-11-12
layer properties to dictionary

sofroniewn:
# Description
This PR will need to go in before either #654 or #511 and will provide the necessary machinery to rapidly serialize (and deserialize?) the viewer.

It is still in progress, but right now it generates an `attrs` dictionary of basic `list` / `str` / `numeric` data for a layer, and an `arrays` dictionary of `array` or long `list` of `str` or `array`, or tuple of `array` data as these will likely be handled differently downstream.

I still need to add machinery to the viewer to assemble these layer dictionaries and any larger viewer dictionaries into a big dictionary.

Once assembled in #511 this could be passed to a function that wrote it out to zarr, or in #654 a function that shared it through dask distributed (doing the correct things for `arrays` vs `attrs`)

I also want to add to this PR a `deserializer` so that in something like #654 when a subscribers receives a new dictionary it can update its viewer model easily. For the special case #511 and reading a zarr frile from disk, the viewer dictionary could be passed to an empty viewer to populate a brand new one from scratch (i.e. create layers where needed vs update them).

I also want to add tests for the individual layers that the `attrs` + `arrays` match the function signature of the layer model `__init__` and test various round trips / updates.

Once all this is done then #654 and #511 can be continued (or superseded and begun again as appropriate)

## Type of change

# How has this been tested?

## Final checklist:


sofroniewn:
@kevinyamauchi @jni can I get your take on something.

As part of this PR I'm trying to make sure that round trips to and back from our json representation of a layer are really nice. One problem I'm encountering though is on both the `Points` and `Shapes` layer for the properties that can vary one for each shape. This may relate to discussions in #651 and ultimately get refactored, but for now if you take a property like `edge_color` on `Points`, then in the `Points` init you can pass either a single string, or a list of strings (one color for each point). Inside the layer though `edge_color` is forced to remain a single string, corresponding to the displayed text of the dropdown widget, and the list of strings is put under the `edge_colors`, this makes the round trip hard as the property name has changed and will require lots of special casing. What I do here is to move the single string to a new property `current_edge_color` and put our list of strings back under `edge_color`. Strictly speaking this will be a breaking change for anyone working with properties on our models, but it will preserve the main api of `add_points` and bring about a lot more symmetry. Ideally we'd approach the `Text` layer #600 in the same way.

Does this choice sound good to you @kevinyamauchi @jni? If so I will continue to refactor `Shapes` in the same way. Regardless of what we choose to do for the serialization stuff I think this change is in the right direction

jni:
tbh it's a bit early in the morning and I don't have much time to think deeply about this, BUT, to me your proposal seems like a reasonable approach, and we are still in alpha so I think it's reasonable to have such a (rather minor) API change. I would suggest though that we hold off on merging this until we are close to releasing 0.3, as we probably will have a few more 0.2.x releases with bug and minor usability fixes. We should reserve this API break for 0.3. Does that sound reasonable?

sofroniewn:
@jni we can also wait until #651 and the revamp of `Points`, `Text`, and maybe `Annotations` :-). @kevinyamauchi suggested similar recently.

I will press on with this draft, to keep exploring what this looks like in the mean time.

kevinyamauchi:
@sofroniewn , I agree that it is reasonable to create a new property that tracks the value of a corresponding element in the UI. I did something similar for the "new text" field in the `Text` layer in #600 and I think it works well. We should come up with a standardized way to name this type of property for all layers. I vote for something like `current_<property_name>` or `selected_<property_name>`. 

I can incorporate these changes when I do the `Points` and `Text` refactor being discussed in #651 .

Also, +1 for waiting to merge this change until 0.3.0 since it is a breaking change on the API.

sofroniewn:
This PR now adds a `_get_state` method to each layer that returns a dictionary with all the layer properties needed to recreate the layer perfectly. This will be useful during remote sharing of the viewer, saving and sharing viewer state, and maybe making animations. I'm keeping the method private for now as this is an experimental feature we might want to keep iterating on. I also made it a function so that in the future we can pass optional keyword arguments that might do things like wrap the data in dask futures calls or prepare arrays for zarr serialization etc. All this can come later I think. I also don't do serialization of the viewer right now, that can come up in a small follow up PR afterwards.

The thing this PR adds which is most important is the `test_serialize.py` test which actually tests the round trips for all our individual layers. In the future similar tests will be added to test round trips at the viewer level.

In order to make this round trip possible though we had do a lot of renaming `edge_color->current_edge_color` and `edge_colors->edge_color`. This will cause merge with #782 and some breaking changes that will need to be explicitly documented - but note non of those breaking changes are in the main API, all are in the layer properties and are simple renames
`X-> current_X` and `Xs->X`.

I'd like to get this in as soon as possible so we can press on with more changes. Does this sound reasonable @kevinyamauchi @jni?

kevinyamauchi:
@sofroniewn, sorry! I somehow missed this! I will take a look first thing tomorrow.

napari/layers/points/points.py
```diff
@@ -58,29 +58,29 @@ class Points(Layer):
         Coordinates for N points in D dimensions.
     symbol : str
         Symbol used for all point markers.
-    size : float
-        Size of the marker for the next point to be added or the currently
-        selected point.
+    size : array (N, D)
+        Array of sizes for each point in each dimension. Must have the same
```
kevinyamauchi, line 62: Thanks for fixing this! I just noticed this yesterday and hadn't gotten around to making an issue.

napari/layers/points/tests/test_points.py
```diff
@@ -385,55 +385,55 @@ def test_face_color():
...
     data = 20 * np.random.random(shape)
     layer = Points(data)
-    assert layer.face_color == 'white'
-    assert len(layer.face_colors) == shape[0]
-    assert np.all([col == 'white' for col in layer.face_colors])
+    assert layer.current_face_color == 'white'
+    assert len(layer.face_color) == shape[0]
+    assert np.all([col == 'white' for col in layer.face_color])
 
     # With no data selected chaning face color has no effect
```
kevinyamauchi:
```suggestion
    # With no data selected changing face color has no effect
```

napari/layers/tests/test_serialize.py
```diff
@@ -0,0 +1,41 @@
+import numpy as np
+import pytest
+from napari.utils.misc import callsignature
+from napari.tests.utils import layer_test_data
+
+
+@pytest.mark.parametrize('Layer, data, ndim', layer_test_data)
```
kevinyamauchi, line 7: Nice!

kevinyamauchi (approved):
Thanks, @sofroniewn! This looks good to me! Unfortunately, as you mentioned, I think this will create merge conflicts with #782, but it shouldn't be too bad.

kevinyamauchi:
Also, happy new year!

napari/layers/points/tests/test_points.py
sofroniewn: fixed

sofroniewn:
Thanks for taking a look and the approve @kevinyamauchi. I have fixed the typos. I will let sit a little for @jni to take a look, but I'd like to merge by Monday if possible :-)

napari/layers/image/image.py
```diff
@@ -337,6 +337,26 @@ def rendering(self, rendering):
...
+        Returns
+        -------
+        state : dict
+            Dictionary of layer state.
+        """
+        state = self._get_base_state()
+        state['rgb'] = self.rgb
+        state['is_pyramid'] = self.is_pyramid
+        state['colormap'] = self.colormap[0]
+        state['contrast_limits'] = self.contrast_limits
```
tlambert03: maybe add `contrast_limits_range` now that it's not private?

napari/layers/surface/surface.py
```diff
@@ -220,6 +220,21 @@ def _get_extent(self):
...
+        """Get dictionary of layer state.
+
+        Returns
+        -------
+        state : dict
+            Dictionary of layer state.
+        """
+        state = self._get_base_state()
+        state['colormap'] = self.colormap[0]
+        state['contrast_limits'] = self.contrast_limits
```
tlambert03: same as above `contrast_limits_range`?  (in both cases... since it will not affect the visual representation, it's not critical, but it also probably doesn't hurt)

napari/layers/tests/test_serialize.py
tlambert03: yeah this is great.

tlambert03 (approved):
this is great, and I can definitely see it being useful.  I don't mind the API change for the points layers.  Once you merge this, I'll open a small PR to make layers hashable for easy equivalence checking with `layer == layer`?

I think my main comment would be syntactical ... I slightly prefer `state.update()` over the more verbose `state['key'] = value`, for example in labels.py, it would be:
```python
    def _get_state(self):
        state = self._get_base_state()
        state.update(
            {
                'is_pyramid': self.is_pyramid,
                'num_colors': self.num_colors,
                'seed': self.seed,
                'data': self.data,
            }
        )
        return state
```

napari/layers/image/image.py
sofroniewn: The rule I am following for adding here or not is if it is present as a keyword arg to the layer. If we add here then we should add to the function signature of `Image` and `Surface`. I'm not sure if we want to do that or not, I think we did at one point, but that can happen later

napari/layers/surface/surface.py
sofroniewn: see comment above about `init` of layer

sofroniewn:
Following the request from @tlambert03 I've updated to use the `update` syntax. I will merge tomorrow morning, unless hearing otherwise.

A layers hashable PR would be a great followup PR.

I'll then start work on a viewer model to dictionary PR too - and address any of the remaining viewer-model separation concerns raised by @royerloic here - https://github.com/napari/napari/issues/448#issuecomment-570745596

napari/layers/labels/labels.py
```diff
@@ -241,6 +241,25 @@ def num_colors(self, num_colors):
...
+        """
+        state = self._get_base_state()
+        state.update(
+            {
+                'is_pyramid': self.is_pyramid,
+                'num_colors': self.num_colors,
+                'seed': self.seed,
+                'data': self.data,
+            }
+        )
```
jni, line 260: These dictionaries feel quite redundant and brittle. I would suggest that we have some kind of global/class variable, e.g. `_SERIALIZED_ATTRS`, and then this becomes:

```python
state.update({name: getattr(self, name) for name in self._SERIALIZED_ATTRS})
```

napari/layers/points/points.py
```diff
@@ -361,35 +367,57 @@ def edge_width(self, edge_width: Union[None, float]) -> None:
...
+            {
+                'symbol': self.symbol,
+                'edge_width': self.edge_width,
+                'face_color': self.face_color,
+                'edge_color': self.edge_color,
+                'n_dimensional': self.n_dimensional,
+                'size': self.size,
+                'data': self.data,
+            }
+        )
```
jni, line 418: As above

napari/layers/tests/test_serialize.py
```diff
@@ -0,0 +1,41 @@
...
+    # Check every property is in call signature
+    signature = callsignature(Layer)
+    for prop in properties.keys():
+        assert prop in signature.parameters
+
+    # Check number of properties is same as number in signature
+    assert len(properties) == len(signature.parameters)
+
+    # Check new layer can be created
+    new_layer = Layer(**properties)
```
jni, line 26: Love this bit. =D

jni (commented):
@sofroniewn sorry for the delay in reviewing, but maybe my comments can be implemented as a followup PR.  They are quite minimal — great work here. =)

napari/layers/labels/labels.py
tlambert03: unfortunately, there's at least [one case with colormaps](https://github.com/sofroniewn/napari/blob/65a275a168ae55fef38e4568bbf6178d7822d724/napari/layers/image/image.py#L353) where this won't work because it's not just getting the attribute.  could just change the colormap attribute though

sofroniewn: ooh nice, I like that idea. i feel like maybe we should change the colormap attribute for that reason, I've always found the tuple a little unclear. maybe I'll make a short follow-up PR

jni: sounds good. Well, the colormap itself should be saved anyway, since the name doesn't always contain all the info. Do vispy colormaps serialize, though?

sofroniewn: I don't know if vispy colormaps serialize - we might want to move to some numpy / list of numpy like representation if we can - like in #782 - then our model files would be vispy independent which I think would be really nice

PR #690, opened 2019-11-12
Add an example showing mouse drag callbacks

jni:
Inspired by this question on image.sc:

https://forum.image.sc/t/how-can-i-receive-mouse-move-drag-events-as-user-drags-a-line-shape/31247

### DIFF
--- a/examples/mouse_drag_callback.py
+++ b/examples/mouse_drag_callback.py
@@ -1,0 +1,45 @@
+"""
+Example updating the status bar with line profile info while dragging
+lines around in a shapes layer.
+"""
+
+from skimage import data
+from skimage import measure
+import numpy as np
+import napari
+
+
+def profile_lines(image, shape_layer):
+    profile_data = []
+    for line in shape_layer.data:
+        profile_data.append(
+            measure.profile_line(image, line[0], line[1]).mean()
+        )
+    msg = ('profile means: ['
+            + ', '.join([f'{d:.2f}' for d in profile_data])
+            + ']')
+    shape_layer.status = msg
+
+
+
+with napari.gui_qt():
+    np.random.seed(1)
+    viewer = napari.Viewer()
+    blobs = data.binary_blobs(length=512, volume_fraction=0.1, n_dim=2)
+    viewer.add_image(blobs, name='blobs')
+    line1 = np.array([[11, 13], [111, 113]])
+    line2 = np.array([[200, 200], [400, 300]])
+    lines = [line1, line2]
+    shapes_layer = viewer.add_shapes(lines, shape_type='line',
+            edge_width=5, edge_color='coral', face_color='royalblue')
+    shapes_layer.mode = 'select'
+
+    @shapes_layer.mouse_drag_callbacks.append
+    def profile_lines_drag(layer, event):
+        profile_lines(blobs, layer)
+        yield
+        while event.type == 'mouse_move':
+            profile_lines(blobs, layer)
+            # the yield statement allows the mouse UI to keep working while
+            # this loop is executed repeatedly
+            yield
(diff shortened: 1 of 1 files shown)


examples/mouse_drag_callback.py
```diff
@@ -0,0 +1,45 @@
...
+        profile_data.append(
+            measure.profile_line(image, line[0], line[1]).mean()
+        )
+    msg = ('profile means: ['
+            + ', '.join([f'{d:.2f}' for d in profile_data])
+            + ']')
+    shape_layer.status = msg
+
+
+
```
sofroniewn: not sure why black is failing as I thought we weren't testing the examples anymore, regardless maybe extra line here

sofroniewn:
This example is so cool!!! Deserves a sweet gif to go with it, I'm happy to make it if you like @jni

examples/mouse_drag_callback.py
```diff
@@ -0,0 +1,45 @@
...
+    blobs = data.binary_blobs(length=512, volume_fraction=0.1, n_dim=2)
+    viewer.add_image(blobs, name='blobs')
+    line1 = np.array([[11, 13], [111, 113]])
+    line2 = np.array([[200, 200], [400, 300]])
+    lines = [line1, line2]
+    shapes_layer = viewer.add_shapes(lines, shape_type='line',
+            edge_width=5, edge_color='coral', face_color='royalblue')
+    shapes_layer.mode = 'select'
+
+    @shapes_layer.mouse_drag_callbacks.append
```
sofroniewn, line 36: Curious why not define the callback at the viewer level and then work with the viewer object to get both the `Shape` and `Image` layers - we might have to use layer names / order to grab them though which isn't so nice as these are both changeable ...

Worth considering as we think about the best api's here

sofroniewn:
We're getting numcodecs fails again https://github.com/napari/napari/issues/665 - still on `0.6.4`, not sure what's going on with your branch @jni

jni:
@sofroniewn ah, I think I haven't updated my local master for a while, and I branched from there. Will need to rebase and force push.

sofroniewn:
@jni - hope you don't mind I made a gif - it's just so cool :-)
Also I think it was @cudmore who posted the question on stack overflow and might be interested in this:

line_callback

Not how as the lines move the status bar updates with the mean of the profile along the line

jni:
Thanks for the GIF! Sorry I haven't yet installed LICECap on this machine. But yep it's pretty great, if I do say so myself. =D This together with @tlambert03's histogram widget example will give us a line-profile plugin in no time. =D

Regarding your API Q, I prefer the closure to the viewer binding. To me this feels like it belongs on that shapes layer. But I agree both approaches might be useful...

sofroniewn (approved):
Sounds good with respect to the layer binding.

I will merge :-)

Definitely excited to hook this up to some 1D line plots!!!!!

tlambert03:
> This together with @tlambert03's histogram widget example will give us a line-profile plugin in no time. =D

you mean like this? 🙂 

ezgif-4-b18e9bf869c8

the relevant code here looks like this:
```python
    viewer = napari.Viewer()
    viewer.add_image(data.chelsea().mean(-1))
    shapes_layer = viewer.add_shapes(
        [np.array([[11, 13], [250, 313]])],
        shape_type='line',
        edge_width=5,
    )
    shapes_layer.mode = 'select'
    w = QtLineProfile(viewer=viewer)  # from #675 
    w.show()

    def profile_lines(image, shape_layer):
        # only a single line for this example
        for line in shape_layer.data:
            w.set_data(measure.profile_line(image, line[0], line[1]))

    @shapes_layer.mouse_drag_callbacks.append
    def profile_lines_drag(layer, event):
        profile_lines(chelsea, layer)
        yield
        while event.type == 'mouse_move':
            profile_lines(chelsea, layer)
            yield
```

PR #692, opened 2019-11-13
change no. of pixels calculation from 32 to 64-bit

### LINKED ISSUE #691: is_pyramid check has an integer overflow with large arrays
## 🐛 Bug

The pyramid size check uses np.prod which defaults to np.int32 data type and there is integer overflow when calculating size on images with dimension products greater than 2^32. 

The function below has the current function in, but commented out.
```python
def is_pyramid(data):
    """If shape of arrays along first axis is strictly decreasing.
    """
    #size = np.array([np.prod(d.shape) for d in data]) 
    size = np.array([np.prod(d.shape,dtype=np.uint64) for d in data])  #should be 64-bit integer

    if len(size) > 1:
        return np.all(size[:-1] > size[1:])
    else:
        return False
```

## To Reproduce
```python
import napari
import numpy as np

#pyramid of zeros
pyramid = [np.zeros((50000,50000,3),dtype=np.uint8),
 np.zeros((25000,25000,3),dtype=np.uint8),
np.zeros((12500,12500,3),dtype=np.uint8)]

size_default = np.array([np.prod(d.shape) for d in pyramid])
```
```
Out[43]: array([-1089934592,  1875000000,   468750000])
```
```python
size_uint64 = np.array([np.prod(d.shape,dtype=np.uint64) for d in pyramid])
```
```
Out[45]: array([7500000000, 1875000000,  468750000], dtype=uint64)
```
```python
with napari.gui_qt():
    # add image pyramid
    napari.view_image(pyramid, is_pyramid=True)

```
Traceback:
```

  File "<ipython-input-48-ce4a0e063e17>", line 3, in <module>
    napari.view_image(pyramid, is_pyramid=True)

  File "C:\anaconda3\envs\nap\lib\site-packages\napari\view_layers.py", line 120, in view_image
    path=path,

  File "C:\anaconda3\envs\nap\lib\site-packages\napari\components\viewer_model.py", line 509, in add_image
    visible=visible,

  File "C:\anaconda3\envs\nap\lib\site-packages\napari\layers\image\image.py", line 147, in __init__
    data, pyramid=is_pyramid, rgb=rgb

  File "C:\anaconda3\envs\nap\lib\site-packages\napari\util\misc.py", line 143, in get_pyramid_and_rgb
    init_shape = data.shape

AttributeError: 'list' object has no attribute 'shape'
```

Pyramidal checks are based on decreasing numbers of points per resolution. From the previous size estimate with 32 bit integers:
```python
np.all(size_default[:-1] > size_default[1:])
```
```
Out[49]: False
```

```python
np.all(size_uint64[:-1] > size_uint64[1:])
```
```
Out[51]: True
```

## Expected behavior

Should load a pyramidal image.

## Environment

napari, 0.2.5
Qt, 5.12.5
NumPy, 1.17.4
SciPy, 1.3.2
VisPy, 0.6.2
scikit-image, 0.16.2

Platform: Windows-10-10.0.10240-SP0
Python:   3.7.3 | packaged by conda-forge | (default, Jul  1 2019, 22:01:29) [MSC v.1900 64 bit (AMD64)]
Backend:  PyQt5
GL version:  '4.4.0 - Build 20.19.15.4531'
MAX_TEXTURE_SIZE: 16384

NHPatterson:
# Description
Fixes integer overflow bug when numpy calculates number of pixels in a series of pyramidal images using ```np.prod```, which defaulted to a ```np.int32``` type rather than the needed ```np.uint64``` type.

## Type of change

# References
closes #691 

# How has this been tested?
Large pyramids now work.

## Final checklist:


sofroniewn (commented):
Thanks so much @NHPatterson. Do you mind also adding short test to our [util/tests/test_misc.py](https://github.com/napari/napari/blob/master/napari/util/tests/test_misc.py) that captures the now working behavior. You can reuse your example from #691.

If this looks good to @jni too we can then get it merged!

jni:
This looks good to me, but I agree that there should be a test. It should be pretty simple, something like:

```python
import dask.array as da

def test_data_size_calc():
    s = 8192
    data = [da.ones((s,) * 3), da.ones((s//2,) * 3), da.ones((s//4,) * 3)]
    assert is_pyramid(data)
```

Thank you @NHPatterson!

NHPatterson:
I've added the following test to ```test_is_pyramid()```:
```python
    s = int(np.sqrt(2**32) + 2)
    data = [np.zeros((s,) * 2,dtype=np.bool_),np.zeros((s//2,) * 2,dtype=np.bool_),np.zeros((s//4,) * 2,dtype=np.bool_)]
    assert is_pyramid(data)
```
I avoided using dask arrays as @jni suggested because it would be another import in the test sequence but unfortunately with this test, the numpy array does involve making a large matrix, which I've given a ```np.bool_``` data type to use fewer bytes. Hope it is useful.

sofroniewn:
@NHPatterson I think using numpy here is fine, though we do try and keep our tests to using small arrays where possible, and adding a dask import in the test isn't a big deal as we already depend on the package. @jni is this ok with you or would you prefer to see the test with `dask`?

@NHPatterson if possible can you install our precommit hooks using `pre-commit install` and then reformat those two files as the are failing our autoformatting tests of `black` and `flake8`. You can see more info in this part of our [contributing guide](https://github.com/napari/napari/blob/master/docs/CONTRIBUTING.md#setting-up-a-development-installation). If that doesn't work out then we should be able to make a reformatting commit to your PR before merging

jni:
I agree with @sofroniewn. We depend on dask and I think it makes a lot of sense to use it here. Making a >2GB object during our test sequence seems wasteful. It's a substantial chunk of available RAM in many computers out there. (Note that `np.bool_` arrays still take up one byte per element.)

NHPatterson:
Yes, no problem, I'll reconfigure the pull request when I have a moment. As another topic, I'm curious if just checking the number of pixels is the best test for pyramidal images.. Would it not also be wise to check each image's dimensions to make sure they decrease sequentially by a factor of 2 or 3 (not sure how many don't fall within this regime) and possibly warn if they don't?

sofroniewn:
Thanks! We can actually support decreases that are not factor of 2 or uniform, estimating the best place to swap the resolution based on the size of the tiles and current zoom level, though in practice factors of 2x are probably best and switching for other factors might not work so great

sofroniewn:
@NHPatterson we're close to making our next release and I'd love to get this in for that, if possible can you finish it up, otherwise I'm happy to take it over if you enable edits by maintainers on the PR. Thanks so much again for your contribution!

NHPatterson:
Ok I've added the dask based pyramid check and corrected the flake8 / black issues. I think editing by maintainers is allowed if there are any supplemental changes you'd like. Really excited to see where this project can go. After seeing #736, it's clear people are very hungry for a great microscopy data viewer in Python, myself included.

sofroniewn:
Hi @NHPatterson - thanks so much for your change - I was just fixing some of the `"` formatting changes, but have inadvertently updated your PR will all the recent changes from master too! I will revert my commit, but and then just fix the formatting, apologies and thanks so much again for the change!

sofroniewn:
Ok, following some git jiu jitsu I've now restored this PR to it's proper state and removed the `"` that had crept in through the auto-formatting. I will wait till tests pass and then approve and merge!! The git squash will forgive all my git sins, and only the diff in this PR will be captured as we want

sofroniewn (approved):
This is now ready for merge

PR #695, opened 2019-11-15
add_dock_widget method on main Window class

tlambert03:
# Description
Adds a basic API for adding & removing `QDockWidgets` from the `QMainWindow` (`Window._qt_window`).  This will allow for something like this (using #675 as an example):

```python
viewer = napari.Viewer()
widg = QtHistogramWidget(viewer=viewer)  # or any other QWidget
viewer.window.add_dock_widget('bottom', widg)
```

Some new styles are added (because docked widgets inherit styles from `QMainWindow` and we didn't have any explicit styles on that).  

A couple things to discuss:
- I import 'stylesheet.qss' and apply it to the full `Window._qt_window` ... in order to have styles apply to the `QMainWindow`.  You may have thoughts about that.
- At first glance, turning the layers/controls/console into a dockable widget may require a bit more surgery.  Only `QMainWindows` can have docked widgets, but the controls and stuff are part of the grid in the central `QtViewer` widget.  So it may be a bit more invasive to extract them.  @AhmetCanSolak, I'll let you think/comment on that and feel free to push edits directly to my branch.

## Type of change

# References
https://doc.qt.io/qt-5/qdockwidget.html

# How has this been tested?

- [ ] need to write tests

## Final checklist:


sofroniewn:
@tlambert03 thanks for kicking this off. I think it's worth trying to get the dock widgets working with the QtLayerList, QtControls, and QtConsole still in this PR as if those don't work nicely then I'm not sure about going down the dockable path and would prefer not to have to then refactor out.

@AhmetCanSolak do you want to give this a shot on this PR?

@tlambert03 will this hold you up on #675? Or maybe we try and get that in without the dockable stuff as there is a lot in there we already know we want?

tlambert03:
I'm not so sure I see it all as linearly dependent as that.  I do agree that it's worth trying to get QtLayerList, QtControls, and QtConsole into this PR, but I don't think that having dockable widgets is useful only insofar as those things work out nicely.  There's really not a ton of flexibility here, since `QDockWidgets` have to live on a `QMainWindow`, there is only one of those in napari.  So it will be great if @AhmetCanSolak can dissect those widgets out of the `QtViewer` grid and raise them up to the level of the `Window._qt_window`... but this PR is at the `MainWindow` already. It could be used (or altered), if desired, in the process of extracting those and making them dockable, but it is agnostic as to whether those widgets ultimately become dockable or not.

I guess it mostly comes down to whether you all think you're going to want dockable widgets in general, or not.

It doesn't hold up #675, specifically... but if I know that those plots/histograms are destined for a QDockWidget eventually, it will be easier to design the visuals (for instance, to auto-orient the histogram following the `QDockWidget.dockLocationChanged` signal as I showed in the example at #675).

napari/resources/stylesheet.qss
```diff
@@ -925,3 +929,47 @@ QtDeleteShapeButton {
...
+    width: 18px;
+}
+
+QDockWidget::float-button {
+    subcontrol-position: top left;
+    subcontrol-origin: margin;
+    position: absolute;
+    top: 0px; left: 24px; bottom: 0px;
+    width: 18px;
+}
```
AhmetCanSolak:
```suggestion
}

```

sofroniewn:
> I guess it mostly comes down to whether you all think you're going to want dockable widgets in general, or not.

agreed - it's not obvious to me we do, but i'm also not fundamentally opposed. part of my concern is around how many of our features will use the dockable functionality - including those existing ones, or planned ones like plugin UIs. I think there is some tradeoff in terms of the flexibility that you get for having them vs the added UI complexity / variability.

I might try and get thoughts from @liaprins-czi who did our design audit - see here for a gif of how a dockable widget can be used with a 1D plot - https://github.com/napari/napari/pull/675#issuecomment-554102325

The alternative I see is dedicated space in the viewer where elements can be made to appear / not appear - but without the drag / drop aspect, without the option to rearrange them, and without the option to have them be free standing main windows - so like our Console.

Now @jni did say that it is nice to have that extra flexibility of free standing main windows for people who use multiple monitors, so if people want to make more of a case that dockable widgets are definitely in our future and we all agree, then merging this PR as it is now also makes sense to me too

tlambert03:
Fine by me either way.   Sorry if I misunderstood the takeaway from the last meeting.  we can let this one sit until everyone chimes in.  It's definitely not blocking anything important and can just be closed if it's off-target.

sofroniewn:
Takeaway from last meeting was definitely to make the PR so we can start playing with it, so I much appreciate it getting made so quickly!

I can also try promoting the QtConsole to be dockable too to see what that is like.

sofroniewn:
I did some more playing around with multiple dockable widgets using this minimal example:
```python
import numpy as np
import napari
from skimage import data
from qtpy.QtWidgets import QLabel


blobs_raw = data.binary_blobs(length=64, n_dim=2, volume_fraction=0.1)

with napari.gui_qt():
    viewer = napari.view_image(blobs_raw)
    widg1 = QLabel('TEST WIDGET 1')
    widg2 = QLabel('TEST WIDGET 2')
    widg3 = QLabel('TEST WIDGET 3')
    viewer.window.add_dock_widget('bottom', widg1)
    viewer.window.add_dock_widget('left', widg2)
    viewer.window.add_dock_widget('right', widg3)
```

and I'm pretty sold that this is the right direction for us - see this gif of what the interactivity looks like:

dockable_widgets

It really does give us lots of flexibility (which I think will be key for the plugins), while still being quite intuitive.

What are your thoughts @jni? @royerloic? If we know we like this then I think we should just merge and we can followup with the refactoring of the rest of the viewer afterwards - which I definitely think we will want to do

sofroniewn (approved):
Based on my comment above I'm approving this, but will wait to merge until others have had a chance to weigh in

AhmetCanSolak (approved):
I like the way @tlambert03 implemented. Which widgets should or shouldn't be dockable is a good discussion. Love to have it on a separate issue. For the main windows class i think this will do for the meantime. Thank you @tlambert03 !

royerloic:
> I did some more playing around with multiple dockable widgets using this minimal example:
> 
> ```python
> import numpy as np
> import napari
> from skimage import data
> from qtpy.QtWidgets import QLabel
> 
> 
> blobs_raw = data.binary_blobs(length=64, n_dim=2, volume_fraction=0.1)
> 
> with napari.gui_qt():
>     viewer = napari.view_image(blobs_raw)
>     widg1 = QLabel('TEST WIDGET 1')
>     widg2 = QLabel('TEST WIDGET 2')
>     widg3 = QLabel('TEST WIDGET 3')
>     viewer.window.add_dock_widget('bottom', widg1)
>     viewer.window.add_dock_widget('left', widg2)
>     viewer.window.add_dock_widget('right', widg3)
> ```
> 
> and I'm pretty sold that this is the right direction for us - see this gif of what the interactivity looks like:
> 
> dockable_widgets
> 
> It really does give us lots of flexibility (which I think will be key for the plugins), while still being quite intuitive.
> 
> What are your thoughts @jni? @royerloic? If we know we like this then I think we should just merge and we can followup with the refactoring of the rest of the viewer afterwards - which I definitely think we will want to do

Looks great. The added margins required by the close and expand buttons do add 'clutter' a bit to what is otherwise a beautiful design. I wonder if there is a possibility to do something else for that. Also, not completely sure about being able to put anything anywhere... But I guess the price of flexibility is just that...

tlambert03:
> Looks great. The added margins required by the close and expand buttons do add 'clutter' a bit to what is otherwise a beautiful design. I wonder if there is a possibility to do something else for that. 

I spent some time looking into that and came up empty without also killing the buttons.  The best I could do was make the title bar (as it's called) the same color as the background and give it a hover effect, as you see above.
 
>Also, not completely sure about being able to put anything anywhere... But I guess the price of flexibility is just that...

That's pretty easy to restrict here:
https://github.com/napari/napari/blob/156cae1c5cd07ae13771d8e2c2d0aedfa8567a7e/napari/_qt/qt_main_window.py#L201-L209

I put them all in there so the syntax was clear, but we can comment out, for instance `Qt.TopDockWidgetArea` to exclude that zone.  I think that would be a good default (and possible exclude the left too?)

napari/resources/stylesheet.qss
sofroniewn: @tlambert03 can you make this small change and then I think we can merge.

sofroniewn: @tlambert03 can you make this small change and then I think we can merge.

sofroniewn:
I think for now it's fine to have all the options enabled, we can then begin to refine them as we support more dockable widgets. Thanks for all the comments on this. I will merge once the tests pass

tlambert03:
i was just updating that :)  but  it can be always be controlled when instantiating the dockable.

jni:
This is awesome! Thanks everyone!

FWIW I would definitely remove the top area from the allowable spaces, but I guess I'm ok with waiting and seeing.

PR #701, opened 2019-11-17
support multichannel dask array

### LINKED ISSUE #700: channel_axis argument fails with dask.array
## 🐛 Bug
Just wanted to post this here so I don't forget.
Dask arrays do not have a `take` method, so if you try to call `add/view_image` on a dask array with the `channel_axis` argument, you get: `AttributeError: 'Array' object has no attribute 'take'` at line 552 below:

https://github.com/napari/napari/blob/15deef08f1aea5e6a6951f7a83d4713f16070ed4/napari/components/viewer_model.py#L545-L552

```python
import napari
import numpy as np
import dask.array as da 

napari.view_image(np.random.random((2,10,10,10)), channel_axis=0)  # works
napari.view_image(da.random.random((2,10,10,10)), channel_axis=0)  # Attribute Error
```
Can't immediately think of an elegant fix, other than checking if `isinstance(data, da.Array)` and then calling `da.take`

sofroniewn:
# Description
This PR closes #700 by switching to `np.take` calls, which still return dask arrays. It also adds a test to check dask arrays are supported. More generally we should probably add tests for dask arrays to more places in the code base, but that can come in followup PRs.

@tlambert03 can you confirm this works for you / isn't actually triggering compute on the dask array but doing lazy indexing.

## Type of change

# How has this been tested?

- [x] add test to `test_multichannel.py`

## Final checklist:

### DIFF
--- a/napari/components/tests/test_multichannel.py
+++ b/napari/components/tests/test_multichannel.py
@@ -1,4 +1,5 @@
 import numpy as np
+import dask.array as da
 from napari.components import ViewerModel
 from napari.util import colormaps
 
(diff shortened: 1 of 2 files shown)


tlambert03:
yup, looks good to me!  and doesn't trigger compute.

sofroniewn:
Thanks for checking @tlambert03 and thanks for catching the original bug, we'll get this in asap

jni (approved):

jni:
Medium term, we should add a battery of tests for not triggering compute inappropriately on dask arrays. One idea would be to build an array with delayed where the last slice triggers an exception. (Or second to last, since right now we are using the last slice to compute contrast limits.)

Issue #624, opened 2019-10-25
contrast slider only goes to 195


wiebkejahr:
## 🐛 Bug
The contrast slider can only be adjusted to [min, max] = [0, 195].

## To Reproduce
1. load image stack (I tried with a tiff stack written by ImageJ: 16 bit, 4 dimensions)
2. data displays correctly; hovering with the mouse over a pixel displays the correct value
3. however, any pixels brighter than 195 are "white" (ie brightest color of the current colormap)
4. the contrast slider cannot be adjusted to any larger max values

## Environment
napari, 0.2.2
Qt, 5.9.6
NumPy, 1.17.2
SciPy, 1.3.1
VisPy, 0.6.1
scikit-image, 0.15.0
Platform: Darwin-18.7.0-x86_64-i386-64bit
Python:   3.7.4 (default, Aug 13 2019, 15:17:50)  [Clang 4.0.1 (tags/RELEASE_401/final)]
Backend:  PyQt5
GL version:  '2.1 INTEL-12.10.12'
MAX_TEXTURE_SIZE: 16384

tlambert03:
I was actually considering making an issue about this earlier today, since I was confused about it as well.  The issue here is likely the line in `napari.util.misc.calc_data_range` that resorts to a rough estimate of the contrast limits if the image has more than 1e6 pixels (@wiebkejahr, what was the exact shape of your volume?):
https://github.com/napari/napari/blob/f747495fb45fb1d5c9eec339641d4be6a676e737/napari/util/misc.py#L272-L274

I think the 1e6 threshold might be a little bit strict.  I did some tests, and on my computer `calc_data_range` only took 2ms on cubic 100x100x100 volume.  If I relaxed it to 2e7, it took 20ms which still seems reasonable for a one-time call.  Of course that strategy only goes so far...

For very large data, perhaps there could be an attribute that stores whether clims were auto-calculated, and some event that recalculates if new data has been loaded into the view?  I know that could be a little tricky to do smoothly, but otherwise we're left with a clim range slider that will never contain some of the data.

wiebkejahr:
>> that resorts to a rough estimate of the contrast limits if the image has more than 1e6 pixels

I was surprised why nobody had found that before, but that explains it :)

Dimensions are:
64x3x350x300 px
and especially the first planes are very dim.

I'm providing the dataset here:
https://seafile.ist.ac.at/f/c39ed737f4e24c0eb63e/
(don't want to clutter it directly into github since it's rather large)

[...]

PR #705, opened 2019-11-18
Don't calc_data_range on uint8 data

jni:
# Description
We've had a few issues with `calc_data_range` being too brittle. In some cases, for example with bad chunking, even on lazy arrays it is too slow (see e.g. [this image.sc post](https://forum.image.sc/t/napari-issues-it-works-but-no-gpu-and-the-gui-is-all-squashed/31411), while, in other cases, e.g. #624, sampling a few planes does not result in correct estimation of the min and max.

Doing the right thing in all cases is tricky, but for uint8 images (both of the above cases), I think that 0-255 is a sufficiently narrow range that we can just set it to that.

As a bonus I've added a test to ensure that we don't instantiate big dask arrays when using calc_data_range. 

The one case where I see this failing is if someone has a uint8 image with values 0 and 1, when you will have an image that is completely black. I think this is less critical and should be solved in a future PR.

### DIFF
--- a/napari/util/misc.py
+++ b/napari/util/misc.py
@@ -268,7 +268,14 @@ def calc_data_range(data):
     -------
     values : list of float
         Range of values.
+
+    Notes
+    -----
+    If the data type is uint8, no calculation is performed, and 0-255 is
+    returned.
     """
+    if data.dtype == np.uint8:
+        return [0, 255]
     if np.prod(data.shape) > 1e6:
         # If data is very large take the average of the top, bottom, and
         # middle slices
--- a/napari/util/tests/test_misc.py
+++ b/napari/util/tests/test_misc.py
@@ -1,4 +1,6 @@
+import time
 import numpy as np
+import dask.array as da
 from skimage.transform import pyramid_gaussian
 from napari.util.misc import (
     is_rgb,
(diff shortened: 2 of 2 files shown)


sofroniewn:
@jni having the special case for `uint8` is fine. Your tests are failing on windows though. Can you investigate and we can then merge

jni:
@sofroniewn 

```
assert (1574074556.0555258 - 1574074555.0057974) < 1
```

(it's ~1.05)

Sheesh. LOL I guess our Windows machines are pretty slow! I'll change it to 2.

sofroniewn:
I wonder @jni if there isn't a better way to test the instantiation of the dask array instead of doing a time test? Would we start adding other time tests like this for our benchmarking suite? Should this be in our benchmarking suite and should we just run that during our tests? I get that this is a slightly different usage, but curious to get your take

jni:
@sofroniewn possibly. I was thinking about something like making a dask "bomb": using `dask.from_delayed` such that one of the delayed functions actually raises an error. However, I couldn't decide which slice to use for this (since currently we use the top, middle, and last slice, and that could change), and anyway the thing we *really* want to test here is that calc data range doesn't take too long. How we achieve this could change so I didn't want to hard-code too much about the specifics.

Benchmarks can and should take a long time to run, so I don't want to make them specifically part of our test suite. They are a different thing and as this test proves, it's hard to do benchmarking on VMs, so I don't think we should go overboard with this sort of test. It's useful for quick sanity testing, not for real benchmarking.

sofroniewn (approved):
@jni makes sense to me. I am approving and will merge

PR #709, opened 2019-11-18
allows Path in io.magic_imread, fixes #708

### LINKED ISSUE #708: Allow Path object in `magic_read`
## Motivation
I tend to use `pahtlib.Path` objects a lot for path manipulation, and passing such an object to `io.magic_read` fails due to:
```py
    if len(filenames) == 0:
        return None
```

`Path` object should be accepted as input to `magic_read`

glyg:
# Description

Avoids io.magic_read to raise a `TypeError` when passed a `Path` object, casts `Path` objects to string.

## Type of change

# References

closes #708 

# How has this been tested?
I added three tests in `util/tests/test_io.py` to test the input

## Final checklist:
> actually I'm not sure the documentation needs to be changed

### DIFF
--- a/napari/util/io.py
+++ b/napari/util/io.py
@@ -20,7 +20,9 @@ def magic_imread(filenames, *, use_dask=None, stack=True):
     Parameters
     -------
     filenames : list
-        List of filenames or directories to be opened
+        List of filenames or directories to be opened.
+        A list of `pathlib.Path` objects and a single filename or `Path` object
+        are also accepted.
     use_dask : bool
         Whether to use dask to create a lazy array, rather than NumPy.
         Default of None will resolve to True if filenames contains more than
--- a/napari/util/io.py
+++ b/napari/util/io.py
@@ -1,5 +1,7 @@
 import os
+
 from glob import glob
+from pathlib import Path
 
 import numpy as np
 from skimage import io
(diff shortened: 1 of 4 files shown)


napari/util/io.py
```diff
@@ -32,6 +34,10 @@ def magic_imread(filenames, *, use_dask=None, stack=True):
     image : array-like
         Array or list of images
     """
+    # cast Path to string
```
sofroniewn, line 39: Can you edit the doc string of `filenames` in `magic_imread` to indicate `Path` objects are now accepted

sofroniewn (approved):
This looks great, made a minor request on doc strings above. We might also want to update the doc strings of `path` in `add_image` in `components/viewer_model.py` and `path` in `view_image` in `view_layers.py` to indicate that Path objects can now be passed.

I will also wait to see if @jni has any comments before merging, but looks good! Thanks for your contribution!!

napari/util/io.py
```diff
@@ -32,6 +36,10 @@ def magic_imread(filenames, *, use_dask=None, stack=True):
...
+    if isinstance(filenames, Path):
+        filenames = filenames.as_posix()
```
jni, line 41: Q: does `as_posix()` work on Windows?

jni (approved):
@glyg great, thanks! I asked a question about Windows but I just realised it's silly, since your tests and our CI together answered it. =) Merging! Thank you for this improvement! (And great to see you here! 😊)

napari/util/io.py
jni: Never mind, I just realised that your tests and our Windows CI answered yes to this question. =D

glyg:
thanks all, this soft looks really nice, hope I can help later on!

PR #712, opened 2019-11-19
expose iso surface threshold

### LINKED ISSUE #572: isosurface with napari
Can napari generate isosurfaces from a regular 3D array? That would be very cool (and useful)

sofroniewn:
# Description
This PR exposes the iso surface threshold as a slider on image layers (it was previously hard coded to 0 making it not useful). It will close #572. 

The range of the iso surface threshold is always 0 to 1, with 0 rendering just the surface of the volume normally and 1 rendering nothing, I find it slightly counter-intuitive that moving the slider from left to right reduces what you see on the screen, but maybe it's just me and everyone else thinks its ok.

Right now I also get a vispy warning when first setting the iso surface threshold variable
```
INFO: Not setting value for variable float u_threshold; uniform is not active.
```
I should probably figure that out before we merge this.

Finally I don't love the colormap of the iso surface, it is hardcoded to max out on this slightly odd green. Maybe we can fix this before merging this PR.

Despite these caveats I think the PR is ready for review

iso_surface2

## Type of change

# How has this been tested?

- [x] added tests to the image model

## Final checklist:

### DIFF
--- a/napari/_qt/layers/qt_image_layer.py
+++ b/napari/_qt/layers/qt_image_layer.py
@@ -11,6 +11,7 @@ class QtImageControls(QtBaseImageControls):
         self.layer.events.interpolation.connect(self._on_interpolation_change)
         self.layer.events.rendering.connect(self._on_rendering_change)
         self.layer.events.iso_threshold.connect(self._on_iso_threshold_change)
+        self.layer.dims.events.ndisplay.connect(self._on_ndisplay_change)
 
         interp_comboBox = QComboBox()
         for interp in Interpolation:
@@ -23,6 +24,7 @@ class QtImageControls(QtBaseImageControls):
             lambda text=interp_comboBox: self.changeInterpolation(text)
         )
         self.interpComboBox = interp_comboBox
+        self.interpLabel = QLabel('interpolation:')
 
         renderComboBox = QComboBox()
         for render in Rendering:
@@ -35,6 +37,7 @@ class QtImageControls(QtBaseImageControls):
             lambda text=renderComboBox: self.changeRendering(text)
         )
         self.renderComboBox = renderComboBox
+        self.renderLabel = QLabel('rendering:')
 
         sld = QSlider(Qt.Horizontal)
         sld.setFocusPolicy(Qt.NoFocus)
@@ -46,6 +49,7 @@ class QtImageControls(QtBaseImageControls):
             lambda value=sld: self.changeIsoTheshold(value)
         )
         self.isoThesholdSilder = sld
+        self.isoThesholdLabel = QLabel('iso threshold:')
 
         # grid_layout created in QtLayerControls
         # addWidget(widget, row, column, [row_span, column_span])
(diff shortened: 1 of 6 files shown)


tlambert03:
hooray!  I had wondered why the isosurfaces looked funny.  these are gorgeous!
Couple things I noticed:
1. if you change the iso threshold, then go back to 2D mode, when you come back to 3D mode, the threshold doesn't seem to have persisted. 
2. is there ever a reason to change the iso threshold when not in iso rendering mode?  If not, can that slider be hidden when rendering mode is not iso?
3. i agree about the ugly colormap on the isosurface... i wish it obeyed the chosen colormap (but not strictly necessary if that's hardcoded in vispy somewhere)

jni:
fwiw I slightly favour removing the slider when not in iso mode... I think isosurfaces are much less common than volume rendering and having that slider cluttering the interface in the 90% (say) of the time when it is useless really bothers me...

sofroniewn:
Here is the code for the isosurface fragment shader https://github.com/vispy/vispy/blob/612776584a195aefd6369b5bde0a1799f3cfad27/vispy/visuals/volume.py#L335 and improvements here will improve the greenish color we find slightly odd. This will be a separate piece of work possibly involving changes in both vispy and napari.

I will fix the threshold persistence problem and hiding the slider when not in iso rendering mode. i will also hide rendering mode when not in 3D rendering as these not meaningful then

sofroniewn:
I have fixed the render persistence bug noted by @tlambert03 but I still get the following warning from vispy occasionally
```
INFO: Not setting value for variable float u_threshold; uniform is not active.
INFO:vispy:Not setting value for variable float u_threshold; uniform is not active.
```
and had to include an explicit setting of something pretty deep in vispy
```python
            self.node.shared_program['u_threshold'] = self.node.threshold
```
This may related to this vispy line:
https://github.com/vispy/vispy/blob/612776584a195aefd6369b5bde0a1799f3cfad27/vispy/visuals/volume.py#L546
and the statement here and elsewhere that a warning is emitted when setting is attempted on inactive variables - https://github.com/vispy/vispy/blob/612776584a195aefd6369b5bde0a1799f3cfad27/vispy/gloo/program.py#L24. We might want to followup with a vispy PR on this, but I think it's ok for now.

I will make it so that the Qt elements get shown and hidden appropriately - then I think this will be ready for re-review

sofroniewn:
I have now updated this so controls are shown / hidden appropriately. I think this is ready for re-review

napari/_qt/layers/qt_image_layer.py
```diff
@@ -74,3 +101,30 @@ def _on_rendering_change(self, event):
...
+            self.isoThesholdLabel.hide()
+            self.renderComboBox.hide()
+            self.renderLabel.hide()
+            self.interpComboBox.show()
+            self.interpLabel.show()
+        else:
+            self.renderComboBox.show()
+            self.renderLabel.show()
+            self.interpComboBox.hide()
+            self.interpLabel.hide()
```
tlambert03, line 129: I had actually been wondering... does vispy not have different interpolation methods for 3D?  I actually don't hate having a nearest neighbors interpolation for volumetric renderings...

tlambert03 (approved):
looks and works great.  I guess since vispy doesn't have volumeVisual interpolation methods, no reason for us to show it here.  I like hiding things that can't actually do anything :)
fine by me if you want to merge or wait for additional review.

napari/_qt/layers/qt_image_layer.py
sofroniewn: i don't think it does, i know i hard coded nearest neighbors a while back, i think for that reason

jni: It doesn't have NN as far as I can tell, but rather linear (hence the fringes on labels layers, for example). I agree that NN would be very useful, but we need to modify the vispy shaders for that.

sofroniewn: ah right, sorry, i was confused

sofroniewn:
I will merge this now, thanks for the review @tlambert03!

PR #713, opened 2019-11-20
[WIP] Random colormap in shader for labels layer

jni:
This isn't working yet, it's just my attempt to port @perlman's changes linked in #204 to present-day napari.

Currently this fails because vispy expects floats to be in [0, 1].

# Description

## Type of change

# References

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/layers/labels/labels.py
+++ b/napari/layers/labels/labels.py
@@ -132,7 +132,7 @@ class Labels(Image):
 
         self._seed = seed
         self._num_colors = num_colors
-        colormap = ('random', colormaps.label_colormap(self.num_colors))
+        colormap = ('random', colormaps.LabelColormap(seed))
 
         super().__init__(
             data,
@@ -224,6 +224,7 @@ class Labels(Image):
     @seed.setter
     def seed(self, seed):
         self._seed = seed
+        self.colormap.update_shader(seed)
         self._selected_color = self.get_color(self.selected_label)
         self._set_view_slice()
         self.events.selected_label()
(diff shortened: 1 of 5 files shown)


jni:
It is now running but not drawing anything on the labels layer. Additionally, it seems most values map to some shade of blue. :joy:

jni:
So, status update, this is now running! It's not quite correct (dividing by the max is the wrong thing to do since that can change from slice to slice), and the colors are a bit crappy, and the colors appear different when computed on the cpu (because of floating point precision issues I presume), BUT! It's working, looks more or less like it's supposed to, and is blazing fast! Slicing through CREMI is buttery smooth now! :tada: 

I think all the issues with it are totally fixable. Will be working on this in the coming days.

jni:
Thanks to @perlman who of course started this but also gave me the sage advice to just get *something* working on the shader and refine from there.

perlman:
The next step is to implement something along the lines of [neuroglancer's gpu hash](https://github.com/google/neuroglancer/tree/bf3029844f732ded12105e7ed64f6cdfe3bd407f/src/neuroglancer/gpu_hash).

[EDIT: This is overkill for what you are looking for -- this is aliasing various IDs together.]

jni:
@perlman I don't actually understand what the neuroglancer code is doing...

jni:
@perlman and I had a nice and meandering conversation about this, that I'll try to summarise here.

First, the goal of this PR is to offload as much as possible to the GPU. I originally thought that we could get away with zero-copy for int32 or uint32 labels, but actually that is not possible, since vispy copies all volumes after #575 and vispy/vispy#1728 (and rightly so). Long term, we might want to allow ints in vispy (they are allowed in GL, at least GL4), and provide a no-copy, no-modify visual, but for now, that's not an option.

So, we want to get from whatever type the labels have to float32 in as few steps as possible, and in a way that minimises information loss. Specifically, float32 only has 23 bits in the mantissa, so doing int64 -> float32 or even int32 -> float32 results in lossy label handling very quickly:

```python
In [31]: np.float32(2**24)
Out[31]: 16777216.0

In [32]: np.float32(2**24+1)
Out[32]: 16777216.0
```

So, we need to somehow map int32, uint32, int64, and uint64 labels to float32 values in [0, 2**24). A simple approach is `%16777213` (the largest prime number under that value), or another largeish prime if we re-seed.

Then, we can do .astype(float32) on the result or directly on any lower-integer-type labels.

Finally, we need to somehow map the labels to colors in the shader. The current approach is brittle because differences in floating point operations or math function implementations lead to different results in the CPU and GPU, which is the current problem. @perlman's idea for the correct way to do this is to hash the value to an index in a texture. This allows us to select a bunch of maximally-disparate colours, and even create a whole bunch of different label colormaps with different characteristics, which is great. And, we can use very simple hash functions such as mod() that will reproducibly map the first n labels to the first n colors in the texture.

I haven't implemented any of this yet, but I did fix our test. =)

sofroniewn:
@tlambert03 - i'm looking into tackling this now. based on your recent vispy experiences, any advice? Also see comment here https://github.com/jni/napari/pull/2#issuecomment-568159610.

Crux of the matter seems being forced to pass something 0-1. If someone points me in the right direction I can implement the hash

PR #714, opened 2019-11-20
dockable console

sofroniewn:
# Description
This PR makes the console dockable. Curious what your thoughts are @tlambert03 @AhmetCanSolak. The separation between the `QtViewer` and the `Window` class is breaking down here a little, particularly the line below blends them quite a bit
```python
        self.console.parentWidget().setVisible(viz)
```
Curious if you have advice on implementation. Overall I like the functionality.

I will note that the dock widget doesn't seem to preserve memory of the size on hide / show when it is docked on the left, but does when it is docked on the bottom. Does that happen to you too @tlambert03, do you know why that is?

Here is a gif of the new behavior:

dockable_console

If we like this then after we merge I can do something similar for the layers list / layer control panel 
 
## Type of change


sofroniewn:
Oh also, if you provide a `name` when adding the QDockWidget that now gets put in the `window` drop down menu and you can use that line to toggle on and off the widget visibility too

napari/_qt/qt_viewer.py
```diff
@@ -101,8 +98,8 @@ def __init__(self, viewer):
 
         self.setOrientation(Qt.Vertical)
         self.addWidget(main_widget)
-        if self.console.shell is not None:
-            self.addWidget(self.console)
+        # if self.console.shell is not None:
+        #     self.addWidget(self.console)
```
AhmetCanSolak: do we want to keep these lines?

AhmetCanSolak (approved):
Looks simple and clean. I only dropped a comment to clean up lines we are not using anymore.

Btw, I agree with @jni that putting a dockable widget on top of canvas looks not so pleasant. We can have a little follow up PR to disable that section maybe? If more people agree i can drop a quick PR about that.

napari/_qt/qt_viewer.py
sofroniewn: dropped

sofroniewn:
Thanks for the review @AhmetCanSolak I dropped those lines. I'll wait a bit if @tlambert03 or @jni want to comment and will then merge.

I actually like having the option to have widgets at the top - the usage I imagine is that I have the console open at the bottom, the image in the middle, and then a 1-D line plot at the top - I think it will still look quite nice, and feel like we should give users that flexibility.

After this goes in I can make a similar PR pulling out the layers list / layer controls

tlambert03:
i think this is great!
One problem I noticed: if you float the console DockWidget, then close it, clicking the `QtConsoleButton` _will_ make the console visible again, but it has lost the ability to be (re)-dockable... leaving you with a forever floating console.

one thought I had would be to subclass QDockWidget, and override it's closeEvent method, to hook into into the `_toggle_console()`... but that would then require adding more arguments to the `add_dock_widget` method to provide a custom `dockwidget`.  Can you think of a more elegant way to handle this?  perhaps by modifying `Window.docked_console` after instantiation and intercepting the close event?

tlambert03:
also, I've wondered for a while whether there was a keybinding to toggle the console visibility... but I don't see one.  Can that be included in this PR, to make sure that it also works well with the dockwidget?

napari/_qt/qt_main_window.py
```diff
@@ -187,6 +197,8 @@ def add_dock_widget(
             If provided, `widget` will be added as QDockWidget's main widget
         allowed_areas : Qt.DockWidgetArea, optional
             Areas, relative to main window, that the new dock is allowed to go.
+        name : str, optional
+            Name of dock widget to appear in window menu.
```
tlambert03: don't know how to comment on non-PR lines... but as long as we have a names argument (which I like!), you could name the dockwidget on line 212:
```python
dock_widget = QDockWidget(str(name or ''), self._qt_window)
```
and then tabs (if visible) will have names: Screen Shot 2019-11-20 at 1 07 59 PM

adding a name to the DockWidget, however, will then show the word "console" in the title bar ... which I think is unattractive (actually, it's even worse than this image because it's off-centered):

Screen Shot 2019-11-20 at 1 13 58 PM

so to hide it, you can modify `stylesheet.qss` as follows:

```css
QDockWidget::title {
  background-color: {{ background }};
  text-align: center; /* otherwise it overlaps with the buttons */
  margin-top:-50px
}
```

sofroniewn: great idea, done

sofroniewn: > don't know how to comment on non-PR lines

me neither - would be a good feature request for GitHub if it isn't possible!!

sofroniewn:
I have added `shift-cmd-C` as a keyboard shortcut for the console. One thing that is a bit odd right now is that we have a slight separation between keyboard shortcuts that we add via our own bindings or that appear in the drop-down menus, we should probably work to unify that stuff, but that is a separate PR.

I've also added the `name` with the appropriate stylings so that they don't normally appear.

> if you float the console DockWidget, then close it, clicking the QtConsoleButton will make the console visible again, but it has lost the ability to be (re)-dockable... leaving you with a forever floating console.

Yikes, yes this is not very desirable. It's not quite clear to me the best path forward either, but agree we should try and fix this before merging. I don't actually understand why it happens this way @tlambert03 only when the dockwidget is floating, i.e. i'm suprised that we can ever click the x and then still recover the dock widget - i thought i was always going to have to intercept that event.

would we ever consider dropping the title bar entirely? so no `x` and no `pop out`? is that even possible / would that mean that you lacked a "handle" to use to drag the widgets around?

tlambert03:
>Yikes, yes this is not very desirable. It's not quite clear to me the best path forward either, but agree we should try and fix this before merging. I don't actually understand why it happens this way @tlambert03 only when the dockwidget is floating, i.e. i'm suprised that we can ever click the x and then still recover the dock widget - i thought i was always going to have to intercept that event.

my guess is that the close **icon** when the widget is docked is calling `.hide()`, whereas the close **button** on the floated window is actually a different system button that calls `.close()`... or something like that.  I think it will be easy enough to fix, just need to play with it a bit.  i'll look around tonight (if you haven't already found a solution)

>would we ever consider dropping the title bar entirely? so no x and no pop out? is that even possible / would that mean that you lacked a "handle" to use to drag the widgets around?

we _can_ consider hiding those buttons, and alternatively we may find some use in [overriding the title bar widget](https://doc.qt.io/qt-5/qdockwidget.html#setTitleBarWidget)... but i think it's probably of little more than aesthetic use.  I do think the title bar is the handle for dragging widgets.  but regardless of _what_ the handle is... as long as there is any handle (which we of course need) it means that you can undock the widget by simply not dropping it back on the edges.  and once it's undocked, the system window close button will be visible, so i think we _still_ need to intercept the close event.  

more soon...

tlambert03:
ok, I found one more issue, but have a possible general approach for you to consider.

the additional issue is that when the `QDockWidget` has focus, then _none_ of the keybindings are passed through to the `QtViewer`, so no closing of the console with `cmd-shift-c` or any other of the keybindings.  There may be other ways to fix that, but the only way I've found so far is to just create our own QDockWidget subclass that has a viewer attribute to remember which viewer it's actually associated with, and pass keypress events to it:

```python
class ViewerDockWidget(QDockWidget):
    def __init__(self, name, parent, viewer):
        self.viewer = viewer
        super().__init__(name, parent)

    def keyPressEvent(self, event):
        return self.viewer.keyPressEvent(event)
```

nice thing there is you can hide/show the console with the keyboard shortcut even if it's floated...

then change line ~212 to use our own class:
```python
dock_widget = ViewerDockWidget(str(name or ''), self._qt_window, self.qt_viewer)
```

we can then either override the `closeEvent` on the `ViewerDockWidget` (which may be too general), or use [`QDockWidget.setFeatures`](https://doc.qt.io/qt-5/qdockwidget.html#features-prop) to limit what can and can't be done with the `DockWidget` (list of possible features [here](https://doc.qt.io/qt-5/qdockwidget.html#DockWidgetFeature-enum)).  I'm not sure we want to restrict those globally, but it might make sense to restrict the features for the `docked_console` widget, and perhaps the layers/controls... so after you instantiate self.docked_console... set features with:

```python
        if self.qt_viewer.console.shell is not None:
            self.docked_console = self.add_dock_widget(
                'bottom', self.qt_viewer.console, name='console'
            )
            self.docked_console.setFeatures(
                QDockWidget.DockWidgetMovable | QDockWidget.DockWidgetFloatable
            )
            self.docked_console.setVisible(False)
```

now the console can _only_ be shown/hidden using either `QtConsoleButton` in the GUI or the keyboard shortcut.  However, that will then leave you with a float icon that hovers in space:

Screen Shot 2019-11-20 at 5 24 46 PM

That's mostly because of the style overrides I added for `QDockWidget::close-button` and `QDockWidget::float-button`, which I originally included just to put some padding on the left.  For now though, I think you can just delete my styles:
```css
\* delete these *\
QDockWidget::close-button {
    subcontrol-position: top left;
    subcontrol-origin: margin;
    position: z;
    top: 0px; left: 6px; bottom: 0px;
    width: 18px;
}

QDockWidget::float-button {
    subcontrol-position: top left;
    subcontrol-origin: margin;
    position: absolute;
    top: 0px; left: 24px; bottom: 0px;
    width: 18px;
}
```
which will leave you with buttons that auto adjust themselves... but have no padding on the left (not a huge deal... but I'll try to figure out how to add back the padding in a nice way):

Screen Shot 2019-11-20 at 5 39 58 PM

sofroniewn:
Ok interesting @tlambert03 I need to digest that for a couple minutes - I also found a fix by adding the following to line 291 in `qt_viewer.py` that will now work just for the console
```python
        if self.console.parentWidget().isFloating():
            self.console.parentWidget().setFloating(True)
```
This came from this post - https://www.qtcentre.org/threads/28562-Floating-QDockWidget-not-re-docked-by-QMainWindow-restoreState

Note my solution will only work for our console, and I really don't like the calls to `console.parentWidget()` inside `qt_viewer.py` - its really violating some separation of concerns between `QtViewer` and the `Window` and making assumptions about there being a parent window container.

Note that post also has some useful tips about having unique object names for our dock widgets for remembering size / state.
I can do this
```python
        dock_widget.setObjectName(str(name or ''))
```
quite simply, but it doesn't force uniqueness

I will think on how to integrate this with what you posted above.

sofroniewn:
@tlambert03 if we create a custom class I think I'd rather it not take in `parent` as an argument so that we could create and control these things at the `QtViewer` level and then in the `Window` just set the parent and attach them - does that sound reasonable?

I think my solution of the `setFloating(True)` could mean that we can leave the close button there, which I think I prefer.

Forwarding the keypress events sounds good, but if someone has defined their own custom keypress events on the widget then we shouldn't overwrite them. There is a part of me that feels like maybe for the console we should just propagate our custom toggle keypress event, but not others. I'm open to either.

Finally are we absolutely sure @jni @royerloic that we want to support floating dockable widgets, because another option is that we just don't allow floating widgets and then all of this goes away!

tlambert03:
sounds good, I agree with everything you're saying here.  It sounds like there are multiple ways to achieve the behavior we're ultimately after, so it's definitely worth taking some time to consider the consequences of each strategy.  I also need to digest this a bit... and we'll see whether @jni or @royerloic have thoughts.

jni:
Sorry, no time to give proper comments on this but I see nothing here that makes me *not* want the dockable widgets! What are you worried about as the main problem? Code complexity? But, at the same time, I feel like dockable widgets are going to be a requirement (?) once we have plugins, and it's great (?) if we eat our own dog food and use them also.

Maybe we can discuss more during the meeting. Good night!

sofroniewn:
@jni I'm now 100% for dockable widgets, glad to hear you're supportive too, and your comment about plugins and us using in the main app, again 100% agree. The thing I was asking at the end now was if we should allow them to be "floating" or require them to always be in a dock. I feel like for people with multiple screens etc "floating" is pretty nice and we should probably bite the bullet and figure out how to support them properly.

@tlambert03 any thoughts on how to best move forward from where we are now?

royerloic:
I think dockable widgets are great, and I agree with @sofroniewn that multiple screen setups would benefit enormously from that. We might want to restrict a bit where things can dock, for example, no docking at the top. Otherwise I think it is great. My only concern perhaps is the added margins and buttons that are always there even if one does not use docking...
One possibility is to have a window item in the menu bar and have a tick-option for 'dockability' which would activate the docking buttons. Not sure about what I just wrote but just throwing ideas here...

tlambert03:
Great.  Sounds like everyone is pro-dockable-widgets in general.  Broadly speaking, I like the idea of putting some thought in to it upfront, even if that requires a bit more code complexity, and "eating our own dogfood/API" (lol) that we intend others to use.  While I'm all for restricting _certain_ widgets to certain areas, it being python, I think it's sort of a fools errand to restrict where some other developer docks things... unless we eventually create `DockablePlugin` framework that has it's own area restrictions, which is a conversation for another time.  So for now, I think the area conversation is a per-widget discussion. For the console specifically, I'm not terribly opposed to a top-docked arrangement, but don't feel strongly either way.

Regarding the added margins, I haven't given up that we can do a little better.  @royerloic's idea of toggling dockability would certainly eliminate the space-offending title bar (but also toggle all moving/floating functionality). Another option is to use the `DockWidgetVerticalTitleBar` Feature for horizontally oriented widgets... which puts the padding on the left instead (don't worry about the bottom aligned icon, that can change):

Screen Shot 2019-11-21 at 12 17 17 PM

@sofroniewn, with regards to moving forward... I still haven't had time to to play around with the various approaches that you mentioned last night.   but I guess here are the minimal bullet points we need to address "somehow":
1. keyboard events must be passed to `Window.qt_viewer`, even when the dock has focus (this should not prevent a developer from declaring their own keyPressEvents... but we _may_ just need to stipulate that they also call super().keyPressEvents() if they override the method).
2. closing a floated dockwidget should not prevent the ability to redock later
3. we may wish to solve 2 on a widget-by-widget basis by modifying widget features after instantiation, or we may want to make a few "prebuilt" classes that are good for certain use cases.  Either way, we need to make sure the styles (colors, margins, button positions) looks good for all cases.

I have about an hour free this afternoon and will try to work on it (but of course, let me know if you get something you like)

sofroniewn:
@tlambert03 I have now update this PR so that there is a new `QtViewerDockWidget` class and the  console gets put inside a `QtViewerDockWidget` inside `QtViewer` so that the dock widget just needs to get added in the `Window`. I think this is much cleaner.

I've also solved the keypress pass through according to your suggestion, and solved the re-floating bug inside the console hide / show code.

I think this is all ready to go apart from the stylings. Do you want to put the title bar vertically and get the button in the right place? You should now have commit rights to the repo and can push that change straight to my branch :-) Please also leave any other suggestions for edits elsewhere as part of code review too

tlambert03 (approved):
i love it.  I think the new class is great, and i like the private `_add_viewer_dock_widget` method.
i just made some minor changes that I figured you wouldn't mind (docstrings, and making `allowed_areas` behave a bit more like `area`).  And I just added method on the `QtViewerDockWidget` class to automatically set the titlebar orientation.
Note: one "catch" here is that if someone changes the features of the dockwidget after instantiation, then the way i coded it will clobber those changes when the dock is repositioned.  However, because the features are stored as bitwise flags, it's easy to add a feature, but I couldn't figure out how to say "current features but not vertical".  Maybe play around with it one more time and see what you think.  

Also make sure you like the vertical title bar: another "compromise" there is that the styles for the titlebars are hard-coded again.  So if we make a DockWidget that cannot be closed but _can_ be floated, then the float icon is going to be hovering awkwardly in space.  It's a fringe case to be sure... and I've run out of patience for the qss styles documentation.

tlambert03:
one more very minor thing: currently it is possible to create a `QtViewerDockWidget` with a starting `area` that is _not_ in the `allowed_areas`.  It's not really a big deal, and I can't really see the merit in throwing a warning or error, as it's more or less up to the developer to choose where they put the widget.  but just wanted to point that out.

sofroniewn:
@tlambert03 thanks for those additions and the vertical title bar - it looks good to me. I think that stuff with the set features all sounds pretty edge case and we can get this merged. I added a new line to the stylesheet (totally minor) and will merge this after the tests pass.

I'll probably then start work on a similar PR that leverages this same functionality to make the layers list and layer controls their own dock widgets.

Very excited to see how people use this new functionality!!

PR #721, opened 2019-11-22
raise main window when showing

tlambert03:
# Description
When using napari in, for example, a jupyter notebook.  Only the first call to `napari.view_image()` will raise the window above the browser.  In later calls, the window is opened in the background, but the user may not know it if their browser window is occluding the napari window (see gif).

This one-line PR adds a `_qt_window.raise_()` call to the `Window.show()` method to ensure that the window is visible when show() has been called

Untitled

## Type of change

# How has this been tested?

## Final checklist:


sofroniewn (approved):
This is great! Thanks for fixing, I've been annoyed by this before. Will merge promptly

PR #723, opened 2019-11-23
Refactor layer tests to be more parametrized

### LINKED ISSUE #685: More parametrized tests for testing across layer types
## 🚀 Feature
Use more parametrized tests for testing across layer types.

We often want to test the same thing across all of our different layer types, right now in many places in the code we have a different test for each layer. Instead we should reuse the same test but parameterize the different layers, including different input data.

The four places where this can currently be improved are:

- [ ] our `test_add_*` functions inside `tests/test_viewer.py` https://github.com/napari/napari/blob/6c5bb037e191bce3c05c5bf23f4f143ecc89a240/napari/tests/test_viewer.py#L57 (this file should also be given a better name)

- [ ] our `test_view_*` functions inside `tests/test_napari.py` https://github.com/napari/napari/blob/6c5bb037e191bce3c05c5bf23f4f143ecc89a240/napari/tests/test_napari.py#L5 (this file should also be given a better name)

- [ ] our `test_nD_*` functions inside `tests/test_nD.py` 
https://github.com/napari/napari/blob/6c5bb037e191bce3c05c5bf23f4f143ecc89a240/napari/tests/test_nD.py#L6

- [ ] our `test_add_*` functions inside `_qt/tests/test_qt_viewer.py` https://github.com/napari/napari/blob/6c5bb037e191bce3c05c5bf23f4f143ecc89a240/napari/_qt/tests/test_qt_viewer.py#L25

shanaxel42:
# Description

Refactors test_viewer, test_napari, and test_qt_viewer to have one parametrized test that iterates through layer types instead of a separate test for each type. Fixes #685, depends on #769 

I added a LayerTypes Enum to napari.layers.base._constants, but happy to remove if you guys don't like it. 

## Type of change

# References

# How has this been tested?

- I just modified existing tests

## Final checklist:


napari/layers/base/_constants.py
```diff
@@ -25,3 +25,14 @@ class Blending(StringEnum):
...
+
+
+class LayerTypes(StringEnum):
+    LABELS = auto()
+    POINTS = auto()
+    VECTORS = auto()
+    SHAPES = auto()
+    SURFACE = auto()
+    IMAGE = auto()
+    PYRAMID = auto()
```
sofroniewn: I know that we had `Pyramid` and `Volume` as types at one point - and they appeared in tests in that way, but in some ways that was a historical artifact.

I think we should definitely drop `Volume` from this list and I'm inclined to drop them both and just use `Image` in both cases above. For `Pyramid` we should be able to guess correctly that the data provided is a pyramid and do the right thing even if the `is_pyramid` keyword arg isn't passed

napari/components/viewer_model.py
```diff
@@ -364,6 +365,32 @@ def to_svg(self, file=None, view_box=None):
...
+            return self.add_labels(data, **kwargs)
+        if layer_type is LayerTypes.POINTS:
+            return self.add_points(data, **kwargs)
+        if layer_type is LayerTypes.VECTORS:
+            return self.add_vectors(data, **kwargs)
+        if layer_type is LayerTypes.SHAPES:
+            return self.add_shapes(data, **kwargs)
+        if layer_type is LayerTypes.SURFACE:
+            return self.add_surface(data, **kwargs)
+        if layer_type is LayerTypes.IMAGE or layer_type is LayerTypes.VOLUME:
```
sofroniewn: As noted by the Enum, I'd drop `Volume`

napari/components/viewer_model.py
```diff
@@ -364,6 +365,32 @@ def to_svg(self, file=None, view_box=None):
...
+            return self.add_image(data, **kwargs)
+        if layer_type == LayerTypes.PYRAMID:
+            return self.add_image(data, is_pyramid=True, **kwargs)
```
sofroniewn: As noted by the Enum, I'd drop `Pyramid` - in most cases we should do the right thing even if `is_pyramid` isn't set and I don't really think of it conceptually as it's own layer type anymore

napari/components/viewer_model.py
```diff
@@ -364,6 +365,32 @@ def to_svg(self, file=None, view_box=None):
 
         return svg
 
+    def add_layer_by_type(self, layer_type, data, **kwargs):
```
sofroniewn: How do people feel about this being a private method `_ add_layer_by_type` or making this `add_layer` and rename our current a `add_layer` something like `_add_layer`, or do we like it as proposed @jni?

I know this will be useful when doing things for us internally like loading a `napari.zarr` file (see #511) or for the pub/sub syncing with dask (see #686), but I'm not sure we want this to be a major usage mode for users (thinking about what the doc string says etc) which is why I lean towards making it private

napari/tests/test_napari.py
```diff
@@ -1,46 +1,70 @@
...
-    assert len(viewer.layers) == 1
-    assert view.layers.vbox_layout.count() == 2 * len(viewer.layers) + 2
-
-    assert viewer.dims.ndim == 3
-    assert view.dims.nsliders == viewer.dims.ndim
-    assert np.sum(view.dims._displayed_sliders) == 0
 
-    # Close the viewer
-    viewer.window.close()
+def view_layer_type(layer_type, data, ndisplay):
```
sofroniewn: Similar to my comment on `add_layer_type` I think we might want to make this private, but we should see how to convo there resolves. Regardless I think it should be more like `view_layer_by_type` and also take `**kwargs`, no need for special handle on `ndisplay`

napari/tests/test_napari.py
```diff
@@ -1,46 +1,70 @@
...
+    if layer_type is LayerTypes.LABELS:
+        return napari.view_labels(data)
+    if layer_type is LayerTypes.POINTS:
+        return napari.view_points(data)
+    if layer_type is LayerTypes.VECTORS:
+        return napari.view_vectors(data)
+    if layer_type is LayerTypes.SHAPES:
+        return napari.view_shapes(data)
+    if layer_type is LayerTypes.SURFACE:
+        return napari.view_surface(data, ndisplay=ndisplay)
```
sofroniewn: no need for special case ndisplay setting here - but we should have **kwargs on all the `view_*` methods

napari/tests/test_napari.py
```diff
@@ -1,46 +1,70 @@
...
+    layer_type : LayerTypes
+        Layer type to view
+    data :
+        The layer data to view
+    """
+    if layer_type is LayerTypes.IMAGE:
+        return napari.view_image(data)
+    if layer_type is LayerTypes.VOLUME:
+        return napari.view_image(data)
+    if layer_type is LayerTypes.PYRAMID:
```
sofroniewn: As noted elsewhere I lean towards dropping the `Pyramid` type

napari/tests/test_napari.py
```diff
@@ -1,46 +1,70 @@
...
+    Parameters
+    ----------
+    layer_type : LayerTypes
+        Layer type to view
+    data :
+        The layer data to view
+    """
+    if layer_type is LayerTypes.IMAGE:
+        return napari.view_image(data)
+    if layer_type is LayerTypes.VOLUME:
```
sofroniewn: As noted elsewhere I lean towards dropping the `Volume` type

napari/tests/test_viewer.py
```diff
@@ -54,254 +55,41 @@ def test_no_qt_loop():
...
-    """Test adding labels image."""
-    viewer = Viewer()
-    view = viewer.window.qt_viewer
-    qtbot.addWidget(view)
-
-    np.random.seed(0)
-    data = np.random.randint(20, size=(10, 15))
-    layer = viewer.add_labels(data)
+    layer = viewer.add_layer_by_type(layer_type=layer_type, data=data)
+    if layer_type is layer_type.VOLUME:
```
sofroniewn: We shouldn't need the special casing for `Volume`. At the beginning of the test we should check `viewer.dims.ndisplay == 2` the default value, check all the other things are good, then switch to `viewer.dims.ndisplay=3`, check everything again, then switch back to `viewer.dims.ndisplay=2` and check everything again

napari/tests/test_viewer.py
```diff
@@ -54,254 +55,41 @@ def test_no_qt_loop():
...
-        [np.all(vd == d) for vd, d in zip(viewer.layers[0].data, data)]
-    )
-
-    assert len(viewer.layers) == 1
+    if layer_type is LayerTypes.SURFACE:
+        assert np.all(
+            [np.all(vd == d) for vd, d in zip(viewer.layers[0].data, data)]
+        )
+    else:
+        assert len(viewer.layers) == 1
```
sofroniewn: This line should be True for all layer types, so doesn't need to be in the `else` clause

napari/tests/test_viewer.py
```diff
@@ -54,254 +55,41 @@ def test_no_qt_loop():
...
-    faces = np.random.randint(10, size=(6, 3))
-    values = np.random.random(10)
-    data = (vertices, faces, values)
-    layer = viewer.add_surface(data)
-    assert np.all(
-        [np.all(vd == d) for vd, d in zip(viewer.layers[0].data, data)]
-    )
-
-    assert len(viewer.layers) == 1
+    if layer_type is LayerTypes.SURFACE:
```
sofroniewn: We can actually drop this special case for the `Surface`, the line above should capture what we need

napari/tests/test_viewer.py
```diff
@@ -54,254 +55,41 @@ def test_no_qt_loop():
...
+        )
+    else:
+        assert len(viewer.layers) == 1
     assert view.layers.vbox_layout.count() == 2 * len(viewer.layers) + 2
 
-    assert viewer.dims.ndim == 3
+    assert viewer.dims.ndim == ndim
     assert view.dims.nsliders == viewer.dims.ndim
-    assert np.sum(view.dims._displayed_sliders) == 1
+    if layer_type is LayerTypes.SURFACE:
```
sofroniewn: we can drop the special case / if statement and instead do
```python
assert np.sum(view.dims._displayed_sliders) == ndim - ndisplay
```
where ndisplay is a variable we are changing in the test, default value == 2, see my comment above

napari/tests/test_napari.py
```diff
@@ -1,46 +1,70 @@
...
 import napari
 
+from napari.layers import LayerTypes
 
-def test_view_image(qtbot):
-    """Test adding image."""
 
-    np.random.seed(0)
-    data = np.random.random((10, 15))
+@pytest.mark.parametrize('layer_type, data, ndim, ndisplay', [
```
sofroniewn: This list of functions / data can actually be the same for both `test_napari.py` and `test_viewer.py` and these can be combined into one file to further reduce duplicate code. The only difference between them really is the use of 

```python
viewer = view_layer_type(layer_type, data, ndisplay)
```
and
```python
viewer = Viewer()
layer = viewer.add_layer_by_type(layer_type=layer_type, data=data)
```
i.e. how the viewer gets instantiated - as you can see lots of duplicate code!! Even the inside bits of the tests can be the same, and I left a bunch of comments in `test_viewer.py` on how it can be improved

napari/_qt/tests/test_qt_viewer.py
```diff
@@ -22,136 +23,31 @@ def test_qt_viewer(qtbot):
...
-
-    assert viewer.dims.ndim == 2
-    assert view.dims.nsliders == viewer.dims.ndim
-    assert np.sum(view.dims._displayed_sliders) == 0
-
-
-def test_add_vectors(qtbot):
-    """Test adding vectors."""
-    viewer = ViewerModel()
+@pytest.mark.parametrize('layer_type, data, ndim', [
```
sofroniewn: The data used here can actually be the same as in `test_viewer.py` and `test_napari.py`, but maybe its still better to keep this as separate file. Is there a way to define this data tuple / dict thing once in some utils file - maybe living in `napari/layers/tests/util.py` (might have to make this file) and then we can just import it once for use in multiple places?

napari/_qt/tests/test_qt_viewer.py
```diff
@@ -22,136 +23,31 @@ def test_qt_viewer(qtbot):
...
+    (LayerTypes.VOLUME, np.random.random((10, 15, 20)), 3),
+    (LayerTypes.PYRAMID, [np.random.random(s) for s in [(40, 20), (20, 10), (10, 5)]], 2),
+    (LayerTypes.LABELS, np.random.randint(20, size=(10, 15)), 2),
+    (LayerTypes.POINTS, 20 * np.random.random((10, 2)), 2),
+    (LayerTypes.VECTORS, 20 * np.random.random((10, 2, 2)), 2),
+    (LayerTypes.SHAPES, 20 * np.random.random((10, 4, 2)), 2),
+    (LayerTypes.SURFACE, (np.random.random((10, 3)), np.random.randint(10, size=(6, 3)),
+                 np.random.random(10)), 3)
+])
+def test_add_layer(qtbot, layer_type, data, ndim):
```
sofroniewn: the stuff that gets checked here can also be identical to the stuff that gets checked in the tests in `test_viewer.py` and `test_napari.py` the only difference is now
```python
viewer = ViewerModel()
view = QtViewer(viewer)
```
and the use of `view` in the tests. We can do the same tricks with `viewer.dims.ndisplay` going from 2 -> 3 -> 2 as I describe in some of my comments below

sofroniewn (changes requested):
Nice work @shanaxel42, I love how much redundant code this PR is removing!! I've left some comments below about how we can get rid of even more, possibly just defining that big list of layer types + data just once in some util file and then re-using that in multiple tests. I've also suggested dropping the `Pyramid` and `Volume` types and just using `Image`, along with some other more minor comments.

Before you start making changes though I want @jni to weigh in on two things:
- The layer type Enum - I think this will be very useful for us internally in other places too, for example loading `napari.zarr` files or remote syncing of the napari viewer where the `layer_type` info is being passed as a string, but we need to then generate a layer based off of it. We can also use it here - https://github.com/napari/napari/blob/84e05dbbb82b575c8c8949850171ff876644fffe/napari/_vispy/util.py#L9 and here https://github.com/napari/napari/blob/84e05dbbb82b575c8c8949850171ff876644fffe/napari/_qt/layers/util.py#L10 in the code base (though that can come in follow-up PRs), so I am very supportive of its addition but I want to make sure @jni agrees

- The `add_layer_by_type` and `view_layer_by_type` methods added in this PR. As noted in my comments around them I think `add_layer_by_type` can be useful to us internally, but I'm not sure we want to promote its external usage, in which case I lean towards `_add_layer_by_type` and `_view_layer_by_type` or if we really like these methods then I think they should be `view_layer` and `add_layer` and we should rename some other stuff, but curious how @jni sees it.

All in all very exciting to be getting your first napari PR in!!!

sofroniewn:
@shanaxel42 you should also checkout some of the bundling of asserts, and the concepts of a setup_viewer method that @HagaiHargil has implemented in #724. We might want to incorporate those changes into this PR or end up with two complementary PRs - curious what your take is, but we probably still want to wait for @jni to weigh in on the enums / `add_layer_by_type` methods.

also i noticed the flake8 / black stuff isn't passing and recall you mentioning maybe running into some problems - curious if you could expand on that / if there is anything we need to improve in our contributing guide / setup that is misleading or wrong

napari/components/viewer_model.py
jni: (a) I don't understand why there is an Enum being used instead of the classes directly, (b) I definitely think this should be private, and (c) if we use the classes, then this method becomes trivial:

```python
def _add_layer(self, layer_type, data, **kwargs):
    return self.add_layer(layer_type(data, **kwargs))
```

... and in general I prefer that public-facing functions (and most other functions) don't have free `**kwargs` arguments, as they are much more difficult to introspect.

napari/layers/base/_constants.py
jni: As mentioned above, I very much don't see the need at all for this enum. As demonstrated by your comment, @sofroniewn, it's just one more place where things can get out of sync.

napari/tests/test_napari.py
jni: Assuming this parametrization is the only reason for `add_layer_by_type`'s existence, we can parametrize with:

```python
(Image, np.random.random((10, 15)), 2, None),
(Image, np.random.random((10, 15, 20)), 3, None),
...
viewer.add_layer(layer_type(data, ndisplay=ndisplay))
```

@sofroniewn given your comment about QtViewer and ViewerModel tests being sometimes identical, we can then write a utility function, in `napari/util/_testing.py`:

```python
def test_viewer(viewer, layer_type, data, **kwargs):
    viewer.add_layer(layer_type(data), **kwargs)
```

How does that sound?
and the

jni:
@shanaxel42 

> I added a LayerTypes Enum to napari.layers.base._constants, but happy to remove if you guys don't like it.

I don't like it. =P

I left some more specific comments — I think it should be possible to use the classes directly for this.

Thanks for this work btw! It'll make our tests much easier to extend!

jni:
Ah, I only just saw the comment by @sofroniewn about the remote viewer and the enum. Even there, I think we should implement something else that is directly tied to the class name. For example, you could use the fully-qualified class path `napari.layers.Image` to store it, and eval it accordingly to instantiate it.

napari/tests/test_napari.py
sofroniewn: While that change would work in some ways, it isn't quite identical - as say there is some divergence between say `viewer.add_image` and `viewer.add_layer(Image(...))`, with how the multichannel information is being handled for example, and the test should really be capturing the full `add_*` methods.

We could instead just parameterize by the actual functions themselves instead, no?:

```python
(viewer.add_image, np.random.random((10, 15)), 2, None),
(viewer.add_image, np.random.random((10, 15, 20)), 3, None),
(viewer.add_labels, np.random.randint(20, size=(10, 15, 20)), 3, None),
...
add_method(data)
```

and 

```python
(napari.view_image, np.random.random((10, 15)), 2, None),
(napari.view_image, np.random.random((10, 15, 20)), 3, None),
(napari.view_labels, np.random.randint(20, size=(10, 15, 20)), 3, None),
...
view_method(data, ndisplay=ndisplay)
```

In this implementation, this PR wouldn't need to add the enum or change our API in anyway and we'd still have full coverage on all our tests with much less duplicate code.

How does this approach sound @shanaxel42 @jni?

sofroniewn:
@jni @shanaxel42 I left a more detailed comment in-line above, but I think we can parametrize on the `add_*` and `view_*` methods themselves and so test exactly what we want with no additional enums or api changes - my other comments should be re-interpreted in light of this suggestion if we go down this path. If we don't need the enum for this PR then we can wait on the enum for the remote stuff and deal with that at a later time

napari/tests/test_napari.py
jni: @sofroniewn For the first section, you would need to instantiate a single viewer for all the methods, which I think makes the tests subtly different...

sofroniewn: hmm ok fair point, maybe at the end of the test we delete the added layer and check we have an empty viewer `assert len(viewer.layers) == 0`. That's probably a good piece of functionality to have tested regardless!!

napari/components/viewer_model.py
shanaxel42: Does add layer work for every layer type? There seems to be a different method for each layer type (add_volume, add_surface, etc) each doing a different set of things

napari/tests/test_napari.py
shanaxel42: Things would be cleaner if they worked as @jni describes where `viewer.add_layer(Image(...)) == viewer.add_image()`, and actually even better if there was only the `add_layer` method. But since that's not true there needs to be some sort of mapping from a layer type to the correct add method. I'll try and work something out with the classes instead of the enum.

napari/components/viewer_model.py
sofroniewn: No, @shanaxel42, I wouldn't use `add_layer` here as there is a bunch of custom processing inside some of our add_* methods

napari/tests/test_napari.py
shanaxel42: how does one delete a layer?

sofroniewn: you should be able to treat `viewer.layers` like a normal list, so `pop` and `remove`

napari/tests/util.py
```diff
@@ -0,0 +1,89 @@
...
+from napari.layers import (
+    Image,
+    Labels,
+    Points,
+    Shapes,
+    Vectors,
+    Surface
+)
+
+layer_test_data = [
```
sofroniewn, line 9: Can you add a comment on what the elements of the tuple are

napari/tests/util.py
```diff
@@ -0,0 +1,89 @@
...
+    (Surface, (np.random.random((10, 3)), np.random.randint(10, size=(6, 3)),
+                 np.random.random(10)), 3)
+]
+
+
+def add_layer_by_type(viewer, layer_type, data):
+    """
+    Convenience method that maps a LayerType to it's add_layer method.
+
+    Parameters
```
sofroniewn, line 47: doc string out of date

napari/tests/util.py
```diff
@@ -0,0 +1,89 @@
...
+
+    assert viewer.dims.ndim == ndim
+    assert view.dims.nsliders == viewer.dims.ndim
+    assert np.sum(view.dims._displayed_sliders) == ndim - 2
+
+    # Switch to 3D rendering mode and back to 2D rendering mode
+    viewer.dims.ndisplay = 3
+    assert viewer.dims.ndisplay == 3
+    viewer.dims.ndisplay = 2
+    assert viewer.dims.ndisplay == 2
```
sofroniewn: add blank line at end

napari/tests/util.py
```diff
@@ -0,0 +1,89 @@
...
+        return napari.view_points(data)
+    if layer_type is Vectors:
+        return napari.view_vectors(data)
+    if layer_type is Shapes:
+        return napari.view_shapes(data)
+    if layer_type is Surface:
+        return napari.view_surface(data)
+
+
+def check_viewer_functioning(viewer, view, data, ndim):
```
sofroniewn: can we make `view=None` optional and at end of parameter list and if not provided then we don't make those checks.

napari/tests/util.py
```diff
@@ -0,0 +1,89 @@
...
+    viewer.dims.ndisplay = 2
+    assert np.all(viewer.layers[0].data == data)
+    assert len(viewer.layers) == 1
+    assert view.layers.vbox_layout.count() == 2 * len(viewer.layers) + 2
+
+    assert viewer.dims.ndim == ndim
+    assert view.dims.nsliders == viewer.dims.ndim
+    assert np.sum(view.dims._displayed_sliders) == ndim - 2
+
```
sofroniewn, line 79: can you also add some dims reordering in here - see https://github.com/napari/napari/blob/6c5bb037e191bce3c05c5bf23f4f143ecc89a240/napari/tests/test_nD.py#L229 for example, code will depend on ndim

napari/tests/util.py
```diff
@@ -0,0 +1,89 @@
...
+    (Image, np.random.random((10, 15)),  2),
+    (Image, np.random.random((10, 15, 20)), 3),
+    (Image, [np.random.random(s) for s in [(40, 20), (20, 10), (10, 5)]], 2),
+    (Labels, np.random.randint(20, size=(10, 15)), 2),
+    (Points, 20 * np.random.random((10, 2)), 2),
+    (Vectors, 20 * np.random.random((10, 2, 2)), 2),
+    (Shapes, 20 * np.random.random((10, 4, 2)), 2),
+    (Surface, (np.random.random((10, 3)), np.random.randint(10, size=(6, 3)),
+                 np.random.random(10)), 3)
+]
```
sofroniewn, line 29: can you add more ndim=3 tests where you can use the data from https://github.com/napari/napari/blob/6c5bb037e191bce3c05c5bf23f4f143ecc89a240/napari/tests/test_nD.py and then we can delete `test_nD.py` entirely

sofroniewn: for every layer type there should be both a 2D and 3D example, for Image there should also be a 4D example

.pre-commit-config.yaml
```diff
@@ -1,15 +0,0 @@
...
-      # force black to run on whole repo & keep settings from pyproject.toml
-      pass_filenames: false
-      args: [.]
--   repo: https://gitlab.com/pycqa/flake8
-    rev: 3.7.9
-    hooks:
-    - id: flake8
-      pass_filenames: true
-      # this seems to need to be here in addition to setup.cfg
-      exclude: vendored|__init__.py|examples
```
jni: Why is this removed?

napari/tests/util.py
```diff
@@ -0,0 +1,120 @@
...
+            np.random.random(10),
+        ),
+        3,
+    ),
+]
+
+
+def add_layer_by_type(viewer, layer_type, data):
+    """
+    Convenience method that maps a LayerType to it's add_layer method.
```
jni: its

napari/tests/util.py
```diff
@@ -0,0 +1,120 @@
...
+    if layer_type is Points:
+        return viewer.add_points(data)
+    if layer_type is Vectors:
+        return viewer.add_vectors(data)
+    if layer_type is Shapes:
+        return viewer.add_shapes(data)
+    if layer_type is Surface:
+        return viewer.add_surface(data)
+    if layer_type is Image:
+        return viewer.add_image(data)
```
jni: I *think* I would make this a dictionary:

```python
classes = [Labels, Points, Vectors, Shapes, Surface, Image]
names = [cls.__name__.lower() for cls in classes]
layer2method = {
    cls: getattr(Viewer, 'add_' + name) for cls, name in zip(classes, names)
}
return layer2method[layer_type](viewer, data)
```

A giant list of if... returns feels like an antipattern to me.

napari/tests/util.py
```diff
@@ -0,0 +1,120 @@
...
+    if layer_type is Labels:
+        return napari.view_labels(data)
+    if layer_type is Points:
+        return napari.view_points(data)
+    if layer_type is Vectors:
+        return napari.view_vectors(data)
+    if layer_type is Shapes:
+        return napari.view_shapes(data)
+    if layer_type is Surface:
+        return napari.view_surface(data)
```
jni: I would use a dictionary pattern as above. You could reuse the same classes and names lists.

napari/tests/util.py
jni: Yeah, in fact, I might make this whole function keyword-only, since it's unclear in what order these parameters *should* appear. Well, maybe viewer can be positional.

But, btw, I love this function! So much functionality tested for all our layer types now! :rocket:

jni (commented):
@shanaxel42 I've made some comments that should be easy enough to address I hope! This already feels way better. =D

.pre-commit-config.yaml
shanaxel42: A mistake! In my struggles to get pre-commit working with my virtual env

sofroniewn:
This looks great! There will probably be more refactoring of our tests to use these functions in other places, but that can come in follow-up PRs. My only concern now is the failing osx test which is because "too many files are open". We've seen this before - it happens when the Qt windows don't get closed which makes me think we are missing one of the following calls
```python
    # Close the viewer
    viewer.window.close()
```
but I checked and they seem to be in all the places we need, which makes me think that adding the `check_viewer_functioning` is somehow causing this. Unfortunately trying to figure this out will probably be a bit tricky - i.e. commenting lines of code pushing to trigger re-running osx tests - and figuring out where problem is. This will take a long time though as we don't have much osx CI resources. Maybe @kne42 can take a look and she'll know where the problem is

sofroniewn:
Also @shanaxel42 are there any other changes you'd still like to make?

shanaxel42:
> Also @shanaxel42 are there any other changes you'd still like to make?

nope!

jni:
@sofroniewn @shanaxel42 looks like we need to figure this osx failure... See #765 😭

shanaxel42:
> @sofroniewn @shanaxel42 looks like we need to figure this osx failure... See #765 😭

I think I solved it for this PR! See #769. But actually in #765 it looks like he just forgot the `viewer.window.close()` at the end of his paramertized test.

jni:
> looks like he just forgot

*she 😉

shanaxel42:
> > looks like he just forgot
> 
> *she 😉

My bad!!

sofroniewn:
@shanaxel42 I have merged #769 and #784. Can you fix the merge conflict and check this still works.

sofroniewn (approved):
This is great @shanaxel42 all tests are now passing with almost 800 lines of code removed and more comprehensive test coverage!!!! I'll wait till tomorrow morning in case @jni has any additional comments and will then merge

PR #724, opened 2019-11-23
Parametrize some tests

HagaiHargil:
# Description
Addresses #685.

I changed only a single file to see if my approach is what you had in mind. If not I'll just scratch these changes. The code is currently not well-documented, but I believe the idea comes through. I also haven't added any tests to the existing ones yet.

## Type of change

## Final checklist:


sofroniewn (commented):
Hi @HagaiHargil - great to see you take a stab at this - I like the grouping of the `_asserts_per_layer` into something that can be re-used and the `test_viewer` as something that can take different `setup_viewer` methods, that should reduce a lot of duplication.

@shanaxel42 actually also just started work on this yesterday in #723 - apologies for not marking that, and has actually worked on a quite complementary part to what you've approached around some layer type enums and layer creation methods. I think there are some good ideas here that can be incorporated into that PR. I will add a note over there too, but maybe best to hold off on continuing this one for now until we figure out what is overlapping. Maybe we'll try and end up with one PR that has all the stuff in it or we'll end up with two complementary ones

HagaiHargil:
Hi, 

I didn't see his PR. I'll close this one and perhaps work on something else. You can obviously reuse everything I did here if you need it.

Issue #725, opened 2019-11-23
add play button next to dimensions label and slider


sofroniewn:
## 🚀 Feature
To finish up the work done in #607 adding `play` functionality we need to add play / pause icons to next to the dimensions sliders. These can be the traditional triangle / parallel lines. To add icons one has to follow the instructions in our [contributing guide](https://github.com/napari/napari/blob/master/docs/CONTRIBUTING.md#building-the-icons).

We'll also probably want to make a widget for the whole row of elements at this point:
https://github.com/napari/napari/blob/09bf79ead28c80d879d76b80769aceecbfa28208/napari/_qt/qt_dims.py#L259
and then put the axis name, play button, and slider inside the QHBoxLayout of that widget and then keep that top widget in a list so that it we can call `show` / `hide` on the whole widget rather than all the components as is done here - https://github.com/napari/napari/blob/09bf79ead28c80d879d76b80769aceecbfa28208/napari/_qt/qt_dims.py#L216

For examples of how to toggle styles on button presses one can see how some of our other buttons like `QtNDisplayButton` are styled.

sofroniewn:
This is now underway in #726

tlambert03:
Hah, wow, I hadn’t even seen this.  Guess we were thinking the same thing

sofroniewn:
This has been added in #726

PR #726, opened 2019-11-23
Add play button to gui

tlambert03:
# Description
Following up on #607, this PR exposes the `QtDims.play()` method in the gui as play buttons (both forward and backwards) next to each axis.  The existing left/right arrows have been changed to "step" icons, the play button turns to a stop button with reddish background when playing.  I needed to make a change (`QtDims.layout.setSpacing(3)`) to make it work with a lot of dimensions.  Added some new signals to QtDims to make sure the buttons stay synchronized regardless of how the axes are played or stopped.

Screen Shot 2019-11-23 at 6 30 03 PM

I'm quite happy with the way it turned out, but not quite done... as I'd like to try to add a right-click feature that additionally exposes the playback speed and the option to restrict playback to a subset of frames.  Feel free to take a look at the existing functionality though.

## Type of change

# How has this been tested?

## Final checklist:
 ❌ My PR is the minimum possible work for the desired functionality (nope)


sofroniewn:
Did you see I just posted issue #725 :-) I havn't checked out the code or repo yet - will do so soon, but can comment on the screenshot.

I like the red background for the stop button, and the new step icons, but I'm a little confused as by the  play buttons in forward and reverse on both sides of the slider. I guess the one on the left plays in reverse, but I don't imagine that functionality will be used so frequently that it needs to exposed in the GUI in such a prominent way (play direction could be an advanced option like play speed). We could then just have a single play button on the left hand side of the slider immediately after the label name and before the reverse step icon, which I think then be very clear. Unless I'm missing something, and there is something else driving this layout.

As to the right click to expose additional options, that could be very useful in many places in the codebase so I'm curious how it will look and feel here!!

Overall very exciting stuff!!!

jni:
The buttons are beautiful! Just chiming in to agree with @sofroniewn that the reverse play button doesn't need to be there. I'm wary of overcomplicating the sliders, since we do have plans for even more stuff (specifically, projection vs slicing mode). Reverse play can be a right-click option as well.

tlambert03:
Great. I've made a lot of progress today, but am now getting a bit slowed down at the last step of  synching playback settings as set in the popup/context menu with those in the dims model (things like fps, start/stop frame, playback_mode...) so that each axis remembers its own settings.

but so far, I've refactored out the dims sliders into their own widget (keeping coverage at ~100%), added some smart width to the axis_labels so they are visible but everything is still aligned, and created a simple popup modal that should be reusable.  here's what it's currently looking like:

Screen Shot 2019-11-24 at 9 01 16 PM

jni:
:heart_eyes: That's gorgeous!

sofroniewn:
Loving the context menu and `ModalPopup`!!! That's going to be super nice to reuse in other parts of the codebase :-)

Also good choice to pull out `DimSliderWidget` as its own thing.

Pretty minor comment - we've been tending to name our own custom Qt widgets with names beginning `Qt...`, so `QtModalPopup` and `QtDimSliderWidget`, not sure if we conform 100% though ...

tlambert03:
>  we've been tending to name our own custom Qt widgets with names beginning `Qt...`

oh right!  will change.  I'll ping you guys when it's closer to being ready for review.

tlambert03:
OK,  I think this is getting really close!  I'm pretty happy with how it's shaping up at the moment and I think it's probably time for you guys to start looking at the code and weigh in.  As I said, there's a lot of changes, here (I apologize this is _not_ the minimum amount of code required) ... but things kept depending on other things.  That said, the majority of the work is confined to `qt_dims`, and `qt_dim_slider` (and the stylesheet).  Here's the visual overview:

ezgif-1-a8409d72be5b


I decided to leave out the frame-restriction for now, and will add in a later PR.  Here is an overview of what this now contains, to guide your review:

- `QtDims.sliders` and `QtDims.axis_labels` has become `QtDims.slider_widgets`, which is now a list of `QtDimSliderWidget`s... much of the axis-specific logic that was on the `QtDims` object is now on the `QtDimsSliderWidget`, and `QtDims` is leaner.
- `axis_labels` now resize their widths to the longest of all the labels (to a limit), and are right aligned.
- `QtDims.play` has become a mostly front-facing API, while `QtDimSliderWidget._play` is actually responsible for starting the `AnimationWorker`.  Calling `QtDims.play` with parameters will update the model and GUI for the corresponding slider widget.  Some `playback_mode` has turned into `loop_mode` and now wants integers instead of strings (0='once', 1='loop', 2='back_and_forth')... which was easier for the `QComboBox`. 
- `AnimationThread` is no longer subclassing `QThread`, and has become `AnimationWorker` (a `QObject`) following advice [here](https://www.qt.io/blog/2010/06/17/youre-doing-it-wrong) and in [the docs](https://doc.qt.io/qt-5/qthread.html#details).  This is a pretty nice pattern, so a [convenience function](https://github.com/tlambert03/napari/blob/34f0fbd922400dd193e90f5d37cbc8609aabb2d1/napari/_qt/util.py#L38) has been added that takes care of moving a `QObject` with a `work()` method to a new thread, and automatically cleaning up.
- a play button is added with a popup context menu (on right click) to change FPS, direction, and loop mode (frame subset will come in a separate PR).  The context menu comes from a new reusable `QtModalPopup` class that has a convenient `show_above_mouse()` method that will raise the popup just above where the mouse is (and as a modal, the user must close it to continue).
- each `QtDimSliderWidget` now serves as the model that remembers the play settings for the corresponding axis, so regardless of whether the user changes play settings with python or with the gui, the GUI is updated and each axis remembers the last settings used (but axes can be different from one another). 
- The new play button takes the `right_arrow.svg` and the older steppers are replaced with `step_right/left.svg` icons.
- test coverage has been slightly increased for related files.

In addition to looking through the code, I'd appreciate if someone could download this and try it out.  I tried to break it just now, and it seems pretty robust, but it's always possible I've missed something.  And of course, suggestions on code re-arrangement are quite welcome.  I'm taking this out of "draft" mode :)

napari/resources/stylesheet.qss
```diff
@@ -1000,22 +1000,147 @@ QScrollBar::sub-page {
...
+    image: url(":/icons/{{ folder }}/left_arrow.svg");
+}
+
+QtPlayButton[reverse=False] {
+    image: url(":/icons/{{ folder }}/right_arrow.svg");
+}
+
+QtPlayButton[playing=True] {
+    image: url(":/icons/{{ folder }}/new_image.svg");
+    background-color: rgb(153, 18, 31);
```
sofroniewn: do we want to make this color part of our palette, or do we want to hard code it? i'd probably just make it part of the palette, it is pretty easy to do and then no colors are hard coded, but maybe it's a special case

napari/resources/stylesheet.qss
```diff
@@ -1000,22 +1000,147 @@ QScrollBar::sub-page {
...
+}
+
+#QtPopupFrame > QCheckBox::indicator {
+   image: url(":/icons/{{ folder }}/right_arrow.svg");
+   width: 10px;
+   height: 14px;
+   padding-right: 2px;
+   padding-left: 3px;
+}
+#QtPopupFrame > QCheckBox::indicator:checked {
```
sofroniewn, line 1138: This also causes the background color to change as it is inheriting from other stylesheets i think - see screenshot, 
image
you can overwrite background color here

napari/_qt/qt_modal.py
```diff
@@ -0,0 +1,53 @@
...
+        self.setWindowFlags(Qt.Popup | Qt.FramelessWindowHint)
+        self.setLayout(QVBoxLayout())
+
+        self.frame = QFrame()
+        self.frame.setObjectName("QtPopupFrame")
+        self.layout().addWidget(self.frame)
+        self.layout().setContentsMargins(0, 0, 0, 0)
+
+        self.form_layout = QFormLayout()
+        closebutton = QPushButton("Close")
```
sofroniewn: How do you feel about dropping the explicit close button, but then if you click anywhere outside the pop-up it closes, sort of like right clicking in chrome or other applications which bring up those dropdown like menus.

napari/_qt/qt_dims_slider.py
```diff
@@ -0,0 +1,471 @@
...
+        label = QLineEdit(self)
+        label.setObjectName('axis_label')  # needed for _update_label
+        label.setText(self.dims.axis_labels[self.axis])
+        label.home(False)
+        label.setToolTip('Edit to change axis label')
+        label.setAcceptDrops(False)
+        label.setEnabled(True)
+        label.setAlignment(Qt.AlignRight)
+        label.setContentsMargins(0, 0, 2, 0)
+        label.editingFinished.connect(self._update_label)
```
sofroniewn: Curious your thoughts on splitting out the `editingFinished` callbacks which we just want triggered once at the end and also a `textEdited` callback that does the resizing so the resizing will happen as you type. This could be nice as right now if i start typing something long it actually disappears part way through my typing and is only revealed at the end

napari/_qt/qt_dims_slider.py
```diff
@@ -0,0 +1,471 @@
...
+
+
+class QtDimSliderWidget(QWidget):
+    """Compound widget to hold the label, slider and play button for an axis.
+
+    These will usually be instantiated in the QtDims._create_sliders method.
+    This widget *must* be instantiated with a parent QtDims.
+    """
+
+    label_changed = Signal(int, str)  # axis, label
```
sofroniewn, line 31: so I guess this relates to #730, and if not necessary would simplify the code right? We should figure this out from @royerloic because it is quite a bit more overhead if not needed right @tlambert03 and now would probably be a good time to make that change? or could it wait?

napari/_qt/qt_dims.py
```diff
@@ -234,9 +195,22 @@ def _update_nsliders(self):
             if self._displayed_sliders[i]:
                 self._update_slider(i)
 
-    def _update_axis_labels(self, axis):
-        """Updates the label for the given axis."""
-        self.axis_labels[axis].setText(self.dims.axis_labels[axis])
+    def _resize_labels(self):
```
sofroniewn, line 171: love this!

napari/_qt/qt_dims.py
```diff
@@ -280,106 +245,33 @@ def _trim_sliders(self, number_of_sliders):
...
-        return label
-
     def focus_up(self):
         """Shift focused dimension slider to be the next slider above."""
         displayed = list(np.nonzero(self._displayed_sliders)[0])
         if len(displayed) == 0:
             return
 
         if self.last_used is None:
+            # this code may be unreachable?
```
sofroniewn: `last_used` is initialized as `None`, but maybe better initialize it as `0` or something? I'm not sure if it is ever set to `None` again or if it is overwritten right away. If we change the initialization then fine to drop these

napari/_qt/qt_dims.py
```diff
@@ -392,6 +284,7 @@ def focus_down(self):
...
```
sofroniewn: see comment above

napari/_qt/qt_dims.py
```diff
@@ -413,18 +306,18 @@ def play(
         fps: float
             Frames per second for playback.  Negative values will play in
             reverse.  fps == 0 will stop the animation. The view is not
-            guaranteed to keep up with the requested fps, and may drop frames
+            guaranteed to keep up with the requested fps, aƒnd may drop frames
```
sofroniewn: typo - "and"

napari/_qt/qt_dims_slider.py
```diff
@@ -0,0 +1,471 @@
...
+    QCheckBox,
+)
+
+from ..util.event import Event
+from .qt_modal import QtModalPopup
+from .qt_scrollbar import ModifiedScrollBar
+from .util import new_worker_qthread
+
+
+class QtDimSliderWidget(QWidget):
```
sofroniewn, line 24: love this!

napari/_qt/qt_dims_slider.py
```diff
@@ -0,0 +1,471 @@
...
+        )
+        worker.finished.connect(self.qt_dims.stop)
+        thread.finished.connect(self.play_stopped.emit)
+        self.play_started.emit()
+        self.thread = thread
+        return worker, thread
+
+
+class QtCustomDoubleSpinBox(QDoubleSpinBox):
+    """ Custom Spinbox that emits an editingFinished event both on the regular
```
sofroniewn: extra " " typo - """Custom ....

napari/_qt/qt_dims_slider.py
```diff
@@ -0,0 +1,471 @@
...
+    editing event, as well as on the valueChanged event IF the left mouse
+    button is down.  On the standard class, the valueChanged event is too
+    spammy for the lineEdit, but the editingFinished does not pay attention
```
sofroniewn: I have to say I don't quite get this - is it for the frame rate setting? are you getting too many events that things crash / slow down?

napari/_qt/qt_dims_slider.py
```diff
@@ -0,0 +1,471 @@
...
+        layout = QHBoxLayout()
+        layout.addWidget(fpsspin)
+        layout.addWidget(revcheck)
+        self.popup.form_layout.insertRow(
+            0, QLabel('frames per sec:', parent=self.popup), layout
+        )
+        self.fpsspin = fpsspin
+        self.reverse_check = revcheck
+
+        # dimsrange = dims.dims.range[axis]
```
sofroniewn, line 329: I take it this is for the frame range stuff, fine to wait on that for another PR and fine to leave this in here commented, but can you also add another comment so we remember what it is and why it is commented

napari/_qt/qt_dims_slider.py
```diff
@@ -0,0 +1,471 @@
...
+        # rightMouseButton release event.
+        if event.button() == Qt.RightButton:
+            self.popup.show_above_mouse()
+        elif event.button() == Qt.LeftButton:
+            self._on_click()
+
+    def _on_click(self):
+        if self.property('playing') == "True":
+            return self.dims.stop()
+        # TODO: link this to DimSliderWidget not QtDims
```
sofroniewn: What is this TODO? Does it need to be done before merge? Can you add more detail in the comment

sofroniewn (commented):
@tlambert03 this is amazing!! i gave it a download and pushed on the functionality and nothing broke, all performed as expected. I've gone through the PR too and left a bunch of comments. Nothing too major, one about closing the popup box by clicking outside it, one about the impact of #730, one about resizing the line edit while typing, and a few more minor ones.

I think we're pretty close to getting this in!!!

napari/resources/stylesheet.qss
tlambert03: sure.  I'll add a "warning" color to the palette.

napari/_qt/qt_modal.py
tlambert03: I like that idea... I actually tried briefly to add it.  failed on the first attempt and then never got back to it.  I can take another look though.  (return key and escape will also close the menu though)

napari/_qt/qt_dims_slider.py
tlambert03: It’s as reasonable a time as any to trim it, but it could definitely wait.  I don’t know whether it costs us much (other than making it slightly less clear which signals actually communicates with other processes).  My very slight preference would be to omit those signals until used (premature optimization?), but would absolutely defer to @royerloic on that

napari/resources/stylesheet.qss
tlambert03: not sure i follow.  which lines and item background are you specifically referring to?  which part did you want to look different?

napari/_qt/qt_dims_slider.py
tlambert03: could do.  I avoided it at first to prevent a ton of redraws getting sent, but it's not like they'll happen unless someone is actively editing. so might work nicely.  Did you try it?  I'll try and change if it feels better.

napari/_qt/qt_dims.py
tlambert03: I had noticed those lines when trying to get 100% test coverage: there didn't seem to be any way I could get to them. I think it must be overwritten right away.  I'll look at the initialization and trim if possible.

napari/resources/stylesheet.qss
sofroniewn: if you compare the screenshot above with the play direction button reversed it has a light gray, whereas the play direction button forward has dark gray:
image
I think this is the relevant part of the sytlesheet, but maybe wrong, and i think it's because the reverse direction is also inheriting some stylesheet from elsewhere

napari/_qt/qt_dims_slider.py
tlambert03: sorry that wasn't clear. The behavior I wanted was for it _not_ to trigger on valueChanged when typing, because each keypress is going to cause a dramatic visual change (for instance, the "1" in 10 is going to cause a 1 second pause before it resumes when you hit the "0"... typing something like "0.5" is also problematic).  So I went for editingFinished... however, then the spinbox buttons don't work.  Turns out there are no signals for "spinBox.clicked"... Similarly, if you just use the mousePress event, then you can't hold down the mouse on the spinBox and increase the FPS.  It's all minor stuff, but this was the best I came up with to achieve the behavior I wanted.  If the extra code is not worth that slight behavioral "nicety", we can go back to valueChanged.

napari/resources/stylesheet.qss
sofroniewn: do you also want to style the background of the QDoubleSpinBox to have that intermediate gray too?

tlambert03:
thanks for the prompt and thorough review!  I'll fix up all these little details hopefully tomorrow.

napari/_qt/qt_dims_slider.py
sofroniewn: Just tried it - it feels great. I'd just swap `textEdited` for `editingFinished` here and then move
```python
        self.label.clearFocus()
        self.qt_dims.setFocus()
```
to its own function and connect that on `editingFinished` or just drop them entirely. I like them though as they get rid of the flashing cursor when you are done.

napari/_qt/qt_dims_slider.py
sofroniewn: ah that makes more sense, because this can be edited during playing which might trigger all sorts of unpleasant behavior. this is good, maybe make that more clear in the comment and then it's good to stay, it's a nice touch

napari/resources/stylesheet.qss
tlambert03: oh, I actually added that change in background on the direction checkbox to make it pop a bit more when in reverse... but I don't think it's all that intuitive!  good enough just to have it switch directions?  I don't mind the look of the fps, but probably shouldn't "go rogue" from the rest of the theme :) will change

napari/_qt/qt_dims_slider.py
tlambert03: yeah, I like it!

royerloic:
Looks fantastic @tlambert03 ! I like the visual design of the new sliders.

There is a popup there though... I am not a huge fan of popups, modal or not... Long term, I would prefer to see a 'drawer/docked panel' where such  parameters would be presented -- something that could be used for plugin parameters too... Once thing I really don't want to see is popups for (many) parameters of plugins and such... The rational is that popups are very intrusive, especially modal popups -- unless they are very small and appear right next to the place you clicked (which is the case here). The meta-comment really is: the more the popup feels an integral part of the main window and not yet another window that appears are a random position in the screen and requires a lot of click-management for moving, closing, params, etc...  the better it is imho... 

I second @sofroniewn that it would be great if clicking outside of the popup could close the popup. 

Concerning the playback mode, I am not sure about the integers.... You could use an enum with a corresponding integer and accept strings. You can decouple the internal representation from the API representation, enums can encapsulate code for that...

jni:
@royerloic I don't consider this a popup, but rather a context menu. It's an important distinction and it's important that it is not required to *access* the functionality, only to *refine* it. As you say, this is next to the click point and does not require window management, especially if @tlambert03 can figure out the click-out-to-dismiss. But honestly with Esc, Enter, and a fat close button, I'm not too concerned.

In this case, it is a drawer that would be cumbersome, because the drawer would necessarily be far away from the play button. In the scheme of things, I think this is a very good solution for parameter refinement. Without something like this, we will end up with the powerpoint/word banner UI, which is actually decent, but for me, overly complicated for napari. imho this interaction strikes the right balance between simplicity and being able to drill down into the details.

napari/resources/stylesheet.qss
jni: I personally would add an extra line to the dialog, "play direction", with that icon. It would have taken me a very long time to figure out that that's not another play/pause button.

royerloic:
Don't get me wrong, I like it as is... I agree that as implemented this is a context menu and looks great as is. Just wanted to restate my fear for popups :-)  Thinking again, the 'parameter drawer' only concerns plugins and such, and here indeed we have a context menu. This is all great :-)

napari/resources/stylesheet.qss
sofroniewn: I agree with @jni - a dedicated line is probably most clear

napari/_qt/qt_dims_slider.py
sofroniewn: Based on the comment from @royerloic in #730 it seems like this can be simplified here. I’d probably do this before we merge

sofroniewn: I’d also simplify the dims view too

tlambert03:
> There is a popup there though... I am not a huge fan of popups, modal or not... Long term, I would prefer to see a 'drawer/docked panel' where such parameters would be presented -- something that could be used for plugin parameters too... Once thing I really don't want to see is popups for (many) parameters of plugins and such... The rational is that popups are very intrusive, especially modal popups -- unless they are very small and appear right next to the place you clicked (which is the case here). The meta-comment really is: the more the popup feels an integral part of the main window and not yet another window that appears are a random position in the screen and requires a lot of click-management for moving, closing, params, etc... the better it is imho...

I know an agreement has already been reached and you've given the go-ahead on the current design... but just cause I don't want to be "the guy who added popups", I want to enthusiastically agree with @jni's take here:

> it's important that [this "popup"] is not required to access the functionality, only to refine it.. this is next to the click point and does not require window management...  A drawer that would be cumbersome, because the drawer would necessarily be far away from the play button.

I also feel like a generic parameter drawer (that could be used by plugins for instance), while useful in many cases, can also quickly get overloaded with functionality, making it slower to actually find the parameter you want (microsoft banner).  I like that is a very minimal set of strictly relevant parameters, with a non-window appearance, refining the behavior of the thing that is actually clicked on (the play button), and also is incapable of tear-out or other window-management demands.

napari/_qt/qt_dims_slider.py
tlambert03: changed my mind on this... works best as is, removing the TODO

napari/_qt/qt_modal.py
tlambert03: figured this out.  actually just needed to set the QDialog to _not_ be modal!
also manually added a close event on the return/enter keys (except if the spinbox has focus... in which case we simply want return to finishEditing and blur the spinbox but not yet close the window)

napari/resources/stylesheet.qss
tlambert03: ok, got a new line for play direction, made it look more like an arrow and less like a play button, and made the styles for the spinbox more like the other buttons.  is everyone ok with the label "`frames per sec:`" ?

Screen Shot 2019-11-29 at 10 09 34 PM

napari/_qt/qt_dims.py
tlambert03: I'm just gonna remove those comments.  I can't be 100% sure we have all tests cases covered and those conditionals aren't not hurting anything.

tlambert03:
OK, I think I've addressed everything!

The context menu now closes with escape, return/enter (except when editing the spinbox, in which case enter just triggers `editingFinished`) or a click anywhere outside of the popup, and the close button has been removed.  Let me know if you like the new styles, and if you have any remaining suggestions requests!

tlambert03:
oh wait there was one thing I didn't address which was using integers for `loop_mode` in the `play()` function.  @royerloic didn't like that.  I feel like an enum is a bit overkill here (and I don't want a user to have to use an enum as an argument...) but if people prefer the original string-based API for the `QtDims.play()` method (i.e. 'loop', 'once', or 'back_and_forth'), I can return it to that, and just convert to `int` internally for use in the gui `ComboBox`

napari/resources/stylesheet.qss
sofroniewn: looks great to me, both the arrow and the "frames per sec:" I guess the question is would we prefer "frames per second:" to be more explicit? I don't have strong feelings, either is fine

sofroniewn: maybe i lean towards "frames per second:" - might as well spell it out, it won't make the box that much bigger, and it might save someone some confusion

napari/_qt/qt_dims.py
```diff
@@ -415,16 +278,16 @@ def play(
             reverse.  fps == 0 will stop the animation. The view is not
             guaranteed to keep up with the requested fps, and may drop frames
             at higher fps.
-        frame_range: tuple | list
-            If specified, will constrain animation to loop [first, last] frames
-        playback_mode: str
+        loop_mode: int
```
sofroniewn: I agree with @royerloic I'd prefer to use our StringEnum class here where you can pass in strings, `play_once`, `loop_continuous`, `loop_back_and_forth` which get converted to their StringEnums - like this StringEnum https://github.com/napari/napari/blob/c28f7316fa8e68c2ba75ee99a5ae3e075833fa38/napari/layers/base/_constants.py#L6 and this conversion https://github.com/napari/napari/blob/c28f7316fa8e68c2ba75ee99a5ae3e075833fa38/napari/layers/base/base.py#L120

It's more consistent with the approach we have taken in the rest of the repo, and I see now major reason to deviate from it here

napari/_qt/qt_dims_slider.py
sofroniewn: I saw you were able to drop the signals from `QtDims`, which I think simplified it quite a bit. Did you need to keep these ones because they also are used to communicate with `QtDims` whereas the other ones were entirely internal to that class?

napari/_qt/qt_dims_slider.py
```diff
@@ -0,0 +1,492 @@
...
+            fps *= -1 if self.play_button.reverse_check.isChecked() else 1
+            self.__class__.fps.fset(self, fps)
+
+        self.play_button.fpsspin.editingFinished.connect(fps_listener)
+        self.play_button.reverse_check.stateChanged.connect(fps_listener)
+        self.play_stopped.connect(self.play_button._handle_stop)
+        self.play_started.connect(self.play_button._handle_start)
+
+    def _pull_label(self, event):
+        """updates the label LineEdit from the dims model"""
```
sofroniewn: `"""Updates` and `model."""`

sofroniewn (commented):
Truly incredible @tlambert03 I just looked over it again and tested it. It's working great. I left a couple pretty minor comments above, probably the most significant one is agreeing with @royerloic about the enum. I put in some links to our custom StringEnum class we've been using in other places.

So excited for this to go in!!

napari/_qt/qt_dims.py
sofroniewn: Also for populating the combo box, see how it is done here, iterating of `str()` calls to the Enum - https://github.com/napari/napari/blob/c28f7316fa8e68c2ba75ee99a5ae3e075833fa38/napari/_qt/layers/qt_base_layer.py#L35

it all should be pretty transferable to this use case, with no need to go to `int`.

also feel as you noted below, maybe `once`, `loop`, and `back_and_forth` are better string names

tlambert03: I was initially thinking that i liked this... but as I'm implementing this, I find myself having to jump through tons of hoops and boilerplate.  More imports (and new files to avoid circular imports), more lines, less readable code, more opaque signal connections, possibility for bugs (if you forget to convert between enum and string types), and it's still unclear to me exactly what I'm gaining in terms of robustness or even in terms of clarity for the user (that I couldn't get by simply converting one of those strings to an int in `QtDims.play()` which is the only front-facing API where anyone would change these things).  

This really feels like an anti-pattern to me... Other than being consistent with the rest of the repo, what's the reasoning here?

tlambert03: I pushed a commit so you can evaluate it... but I still think I have a preference for reverting it.  Another thing I ran into is that `Qt.Signal()` actually threw a segmentation fault when I declared it using the custom StringEnum type... which forces me to use `Signal(str)` and then make sure to convert back and forth between the custom type and `str` wherever that signal is connected.

sofroniewn: I'd put the Enum definition in a new file `_constants.py` which should avoid any circular import problems. I guess it's nice then to have a single place where constants like this are defined and can be tracked, edited independent of one particular look up table usage, and I prefer the api of `QtDimsSlider` now where mode is the enum instead on an integer which I'm unlikely to remember.

Probably the main difference here vs other places in the code is the `Signal(str)` which means you have an additional conversion. Maybe @jni who is generally more conservative about adding enums want's to weigh in.

tlambert03: yeah, I certainly understand wanting to allow the public api to accept strings rather than hard-to-remember ints.  That's easily done without enums, and one way or another we'll definitely have that be the final public API in the `QtDims.play` method.

For me the enum makes the most sense if a) we can imagine wanting to add additional choices in the future and b) we need those updates to be reflected in many places in the code.  In that case an enum lets you make the change in one place.  Here, though it's hard to think of what another `loop_mode` would even be, and it's also not reflected in multiple places throughout the code (it's basically just an argument to the `play` function and the comboBox).  Given the additional difficulty of using a custom enum type in signals and slots, I would probably be further inclined to immediately cast that enum to a string, which more or less means we've created an enum simply to immediately cast to a string after the play function is called.

All that said, I'm happy to just go with whatever the group decision is... so let's give @jni and others time to weigh in and then I'll implement whatever the general opinion is.

napari/resources/stylesheet.qss
sofroniewn: @tlambert03 did you have thoughts on this change? as i said not huge deal either way

tlambert03: i agree, just updated

napari/_qt/qt_dims_slider.py
tlambert03: Yeah, those remaining signals are used by other `QObjects` to communicate with `QtDims`, that's the distinction that I generally think determines whether we want to use a signal or call an internal method.

sofroniewn: makes sense, thanks for helping me understand!

napari/_qt/qt_dims.py
jni: 1) it definitely shouldn't be an int! That's yucky syntax from C.

2) it *could* be a string. From an API and community expectation perspective, that is Pythonic. But,

3) my rationale here and in all the other places where we use enums is: syntactically, an enum is the right structure: the loop mode can be one of several options. A string provides endless possibilities — and endless *erroneous* possibilities. From a type validation perspective, a string is almost completely uninformative in this case. Yes, enums feel unpythonic, but I think that's only because they are a very recent addition: they appeared in Python 3.4. I think in the future, now that enums are in the standard library and that typing is becoming a much bigger part of Python (type annotations, jits, etc), enums will become more common.

For me, really the thing that convinces me to use enums is that they are *the* data type to represent a value from a defined set of options. ints and strings are both kludges to hack around an incomplete language/typing spec. Now that we *have* enums, we should use them when there are multiple, non-numeric, non-boolean options.

napari/_qt/util.py
```diff
@@ -32,3 +35,102 @@ def QImg2array(img):
...
+
+    if start:
+        thread.start()  # sometimes need to connect stuff before starting
+    return worker, thread
+
+
+class LoopMode(StringEnum):
+    """Looping mode for animating an axis.
+
+    LoopMode.OPAQUE
```
jni: Opaque? 😂

tlambert03: :) yeah I saw that... there is some cleanup necessary with the Enums since I was waiting to hear what the consensus was.  tests are broken as well.  I'll go with the enum though and re-push

napari/_qt/qt_dims.py
tlambert03: OK, I'll leave it as is, and just clean it up a bit.

tlambert03:
OK, cleaned up things with the `LoopMode` enum.  Note, there is a bug with the current `StringEnum` implementation (see #754), and so I am still manually checking whether the provided `loop_mode` is valid and raising a `ValueError`.  We can revisit those lines in PR when that gets fixed.

Hopefully, this is now all set and ready for review.

sofroniewn (approved):
Ok this looks done. I'm approving, agreed we can push a fix after #754 get's fixed, it will be a very small one. I will let this sit one more day in case @jni or anyone else wants to weigh in again, then I will merge

jni:
:+1: from me! Awesome work @tlambert03, everything here just looks so gorgeous.

PR #727, opened 2019-11-24
dockable layers list

sofroniewn:
# Description
This PR builds on #695 and #714 to now make the layers list and controls dockable. I put the layer creation and viewer buttons on the layer list widget - i think it would get too confusing if we had them all be separate dockable ones. I also restrict these widgets to be on the `left` and `right`, and now restrict the console to be on the `top` and `bottom` otherwise it's very easy to accidentally drop them in the wrong place.

The styles might need a little bit of work but aren't terrible. We might want to revisit the foreground colouring of the layer controls, I'm very open to that.

Finally if you close either of these dock widgets then you can re-open it from the window, but they aren't redockable!! I really think that is a Qt bug, that can be fixed by triggering a call like
```python
        if self.dockConsole.isFloating():
            self.dockConsole.setFloating(True)
```
but we'll probably have to intercept the events on our custom `QtViewerDockWidget` class to get this done.

Curious for feedback @tlambert03 @AhmetCanSolak 

## Type of change

# How has this been tested?

- [ ] no tests have been added

## Final checklist:


napari/resources/stylesheet.qss
```diff
@@ -433,8 +443,8 @@ QtControls {
     padding: 4px;
     min-height: 210px;
     max-height: 210px;
-    min-width: 220px;
-    max-width: 220px;
+    min-width: 240px;
+    max-width: 240px;
```
tlambert03: one of the nice things about being able to float the controls would be the ability to gain a bit more horizontal accuracy on the range sliders... but the max-width limit here prevents you from resizing the window after floating.  What would you think about relaxing that?

tlambert03 (dismissed):
I really like the code and the functionality.  It's particularly nice to be able to pop out the layers when the list gets a bit long.  Popping out the controls might become even more useful if you could resize the floated window horizontally (see comment).

I agree the styles need some attention.  I know it's finicky work... but just to point out a couple things (that I'm sure you also noticed): 

- the background on the left edge of the controls used to align with the left edge on the individual layers, now it extends right up to the edge, and needs some padding. (or, as you said, a different background altogether?)
- the right edge of the trash can used to line up with the right edge of the layers.
- the titlebar and float/close icons were not so bad with the console... but i think the general aesthetic takes a bigger hit in this case.  I wonder if we should have an option to make them floatable, but default to a hidden dockwidget title bar otherwise?  Alternatively (or perhaps additionally), we could work on a [custom titlebar widget](https://doc.qt.io/qt-5/qdockwidget.html#setTitleBarWidget) for these more prominent widgets, that would perhaps be smaller in height, and make the hover effects more subtle, but also make it clearer that the titlebar is a handle. (could also set the cursor to an [openhand](https://doc.qt.io/qt-5/qcursor.html) on hover)

napari/resources/stylesheet.qss
sofroniewn: dropping that would be fine, but I can't figure out right now how to get the start-up width to be minimal (now 228), but still allow for it to be expanded. If you know can you suggest here / push the change to my PR

sofroniewn:
@tlambert03 I've just tried to do a round of style improvements, but I can't figure out the minimum but expandable width for the dock widget. Also not all the alignments are fixed. Can you take this PR over and bring it to the finish line with better styles, shouldn't be too much more to go at this point.

tlambert03:
sure, I'll take a look later today and let you know if I have any questions

tlambert03:
@sofroniewn, did you have an idea about where to apply the fix for closed floated windows not redocking?

sofroniewn:
not really to be honest. Should we also add keybindings to open / close these? Or maybe these ones are so important they shouldn't be closable! I know it punts on the issue, but maybe that's ok here.

tlambert03:
How anxious are you to get this in?  I'm running into issues getting the titlebar styles to be acceptable that require me to dig into the docs a bit more than I expected.  I was hoping to wrap up some stuff on #726 while it's fresh in my mind.  OK with you to wait on this a day?

sofroniewn:
Definitely finish up #726, didn’t mean to pull you away from that. I think we’ll want this to be part of our next release, but there are still a couple more other things to get in too. Thanks for picking this up!!

tlambert03:
Just hit on some things that I think will work here.  Will update in a day or so

tlambert03:
just pushed some updated styles for you to consider.  Basically, made a new "QtMinimalDockWidget" class (happy to change the name) that has a slim titlebar, with a 1px line, a slight hover effect, and a hand cursor when hovered.  Currently, even when floated it keeps that style (no native window even when floated).  There is a close button that avoids the "permanently-floating" bug by triggering the `toggleViewAction()` instead of the `closeEvent` (which strangely enough does the trick).  To slim it down even further, we could remove that close button altogether (or just remove when docked).  Double-clicking on the title bar toggles its floating status.  Here's what it looks like:

ezgif-1-a93ab3612dd5


Edit: forgot to add.  The layer controls can also be widened when floating, but not when docked.

sofroniewn:
@tlambert03 this looks amazing. What was the thinking behind not allowing the controls widget to widen when docked? I was unsure at first, but actually think it makes sense if the layer widget can't widen, though you could imaging allowing that to widen at the name edit box level so you could see longer names. I don't have super strong opinions on this though.

Also can you make the widget with the thumbnail + layer name + icon a tiny bit wider so the right edge is aligned with the right edge of the trashcan and controls - see here:
image

Then I think it's good for @jni to take a look over and hopefully merge :-)

napari/_qt/qt_viewer_dock_widget.py
```diff
@@ -76,7 +87,69 @@ def keyPressEvent(self, event):
...
+        self.setCursor(Qt.OpenHandCursor)
+
+    def sizeHint(self):
+        # this seems to be the correct way to set the height of the titlebar
+        szh = super().sizeHint()
+        szh.setHeight(18)
+        return szh
+
+
+class QtMinimalDockWidget(QtViewerDockWidget):
```
sofroniewn: I might call it `QtMinimalDockWidget` - though actually what does it look like it we just always use this one, i.e. even for the console. I might prefer it

sofroniewn: would simplify the code base and uniformity of stylings just to have one `QtViewerDockWidget` which had those minimal styles (either with title bar horizontal or vertical)

sofroniewn: would also remove the re-docking after close problem for all our added widgets now right? apologies if you already planned to just replace `QtViewerDockWidget` and I'm slow to catch on

tlambert03:
> @tlambert03 this looks amazing. What was the thinking behind not allowing the controls widget to widen when docked? I was unsure at first, but actually think it makes sense if the layer widget can't widen, though you could imaging allowing that to widen at the name edit box level so you could see longer names. I don't have super strong opinions on this though.

Some additional resizing handles show up in that case that I think are ugly.  I think my preference at the moment would be to address more flexible layer-list width in another PR if we find ourselves wishing for it.

Screen Shot 2019-12-01 at 6 53 21 PM


> Also can you make the widget with the thumbnail + layer name + icon a tiny bit wider so the right edge is aligned with the right edge of the trashcan and controls

That space is there for the scrollbar.  I think the best case scenario would be that the scrollbar steals the space from the layers list,  but could be a bit complicated and/or computationally busy (would need to check whether the scrollbar was visible on both window `resize` and `layer_change` events).
 
Screen Shot 2019-12-01 at 6 40 40 PM

napari/_qt/qt_viewer_dock_widget.py
tlambert03: well I hadn't necessarily... but I think it probably makes sense.  I don't mind losing that little "float" icon.  The functionality is triply duplicated, since you can already tear it out with a click and drag, and a double click on the title bar does what the float icon did.  The one thing I don't always love is the non-native window look on floated windows.  I wish we could have it both ways... but the native title bar brings back the re-docking problem.  (i even tried to intercept and ignore the `closeEvent`, but couldn't get it to work).  But let me work on it a little more from the perspective of making it the global DockWidget style

sofroniewn:
Ah forgot about the scroll bar, makes total sense, let’s leave the width as is.

Also makes sense for the widening when docked, let’s leave as is.

Great that you’re going to explore moving all dock widgets over to the new style, I do think that will be cleaner

napari/_qt/qt_viewer_dock_widget.py
sofroniewn: Losing that undock button is good by me, as you say very redundant functionality, especially now that the cursor turns to a hand when you hover. I don’t mind losing the native titlebar - I find the switch to a gray styling very off putting in the Mac. In fact after this I might want us to make a follow up PR where we drop the native titlebar on the main viewer window too, but that can come later

tlambert03: >I might want us to make a follow up PR where we drop the native titlebar on the main viewer window too, but that can come later

giphy

jni: I strongly oppose getting rid of the native titlebar. I actually quite liked the mini titlebars on the undocked windows, but I'm ok with them being special — most people expect that and most people wouldn't want to e.g. maximise them or move them to the dock. For the main window, not so much. I think it's very important that napari looks and acts like any other application in the user's system. The familiarity is super important. Everyone knows that pushing the little green button on napari makes it full-screen, (most people know alt-pushing it maximises it), pushing the yellow button puts in it the dock, etc. This is super important.

jni:
I'd love for the width to be adjustable, but it's not a hill I'm willing to die on (unlike the native titlebar =P), and if it makes things uglier that can certainly come in a later PR.

napari/_qt/qt_viewer_dock_widget.py
sofroniewn: We can also make the title bar try and obey the appropriate conventions with the different color buttons while styling it ourselves. Chrome and Illustrator both do that for me - i.e. illustrator looks as follows 
image
and chrome dark mode looks similar. I'd just want to do something similar, but we can discuss this in its own issue

tlambert03:
Hey @sofroniewn, when you have a moment take another look at this.  I made some changes that should allow the `QtMinimalDockWidget` to just be the general dock widget  (with the close button and 1px solid line handle in the titlebar).  It switches orientation depending on the aspect ratio of the dockwidget and handles the closeEvent well without preventing redocking later:

Screen Shot 2019-12-07 at 3 42 20 PM

I went back to the native menubar for floated windows, and was able prevent the permanently-floating problem there too.  I think I might be with @jni on this one that I prefer the native feeling when possible.  Additionally, window resizing seems to work better with the native windows, for some reason...

Screen Shot 2019-12-07 at 3 42 57 PM

lastly, I made the vertical scrollbar on the layers list a bit thinner, so that alignment issue with the trash bin is slightly less noticeable when the scrollbar is not visible:

Screen Shot 2019-12-07 at 3 44 16 PM

Don't worry too much about the exact code at the moment, but if you like the general behavior feel free to make suggestions for a little refactoring to clean it up.  Note also that the console is temporarily allowed to dock anywhere (just to have something to be able to move around freely).

I think it's pretty close otherwise.  The main questions are: 1) do we want the "minimal widget" here to just be the "always" dockwidget style. and 2) does anyone have any suggestions for style improvements (nit-picky is fine, a pixel here or there certainly can matter).

sofroniewn:
Loving it. I think it's very close. I think we should only support the "minimal widget" and that should be renamed to be just the `QtViewerDockWidget`.

Is there any way to get the yellow and green buttons for redock / expand working here:
image

I am with @jni too, I prefer the native styling here too. Nice job figuring that all out!

I actually wouldn't mind having the undock button back here again, next to the `x`, the icon was quite nice and it will be small. i think i'll find it easier than grabbing and dragging and nicer just to click:
image


Otherwise looking great. I havn't gone through the code but will wait till the next round of refactors

tlambert03:
I can add an additional icon,  but note also that you can double click the title bar to toggle dock state, both when docked and when floating.  Which, though it is less obvious, would make for three different ways to float/dock

sofroniewn:
Hmm, I did not realize the double click. I think I’d prefer the button as more explicit, but if looks bad we could always drop again. I’d also be fine dropping the double click, but if it comes for free then maybe people will be expecting it

tlambert03:
ok I incorporated all the requested changes.  unified a single DockWidget class, added a float button and removed the background from the buttons just to keep them a bit more low-key (would like to make them a bit dimmer... but that seems linked to the svg itself).  Fixed the missing colorbar.  here's how it looks.

Screen Shot 2019-12-08 at 10 24 56 AM

> Is there any way to get the yellow and green buttons for redock / expand working here

haven't figured out how to enable those...  double-clicking on the title bar still redocks the window, though it's not immediately obvious.

I'm about to jump on a plane.  I think this is ready for a "regular" review whenever anyone is able.

napari/_qt/layers/qt_labels_layer.py
```diff
@@ -87,28 +87,40 @@ def __init__(self, layer):
...
-        self.grid_layout.addWidget(QLabel('brush size:'), 3, 0, 1, 3)
-        self.grid_layout.addWidget(self.brushSizeSlider, 3, 3, 1, 4)
-        self.grid_layout.addWidget(QLabel('blending:'), 4, 0, 1, 3)
-        self.grid_layout.addWidget(self.blendComboBox, 4, 3, 1, 4)
-        self.grid_layout.addWidget(QLabel('contiguous:'), 5, 0, 1, 3)
-        self.grid_layout.addWidget(self.contigCheckBox, 5, 3)
-        self.grid_layout.addWidget(QLabel('n-dim:'), 6, 0, 1, 3)
-        self.grid_layout.addWidget(self.ndimCheckBox, 6, 3)
-        self.grid_layout.addWidget(self.colormapUpdate, 0, 0, 1, 3)
+        # self.grid_layout.addWidget(self.panzoom_button, 0, 6)
```
sofroniewn: Probably my fault for leaving these in here, but we should delete the commented lines

napari/_qt/layers/qt_points_layer.py
```diff
@@ -92,30 +93,37 @@ def __init__(self, layer):
...
-        self.grid_layout.addWidget(self.symbolComboBox, 4, 3, 1, 4)
-        self.grid_layout.addWidget(QLabel('face color:'), 5, 0, 1, 3)
-        self.grid_layout.addWidget(self.faceComboBox, 5, 3, 1, 3)
-        self.grid_layout.addWidget(self.faceColorSwatch, 5, 6)
-        self.grid_layout.addWidget(QLabel('edge color:'), 6, 0, 1, 3)
-        self.grid_layout.addWidget(self.edgeComboBox, 6, 3, 1, 3)
-        self.grid_layout.addWidget(self.edgeColorSwatch, 6, 6)
-        self.grid_layout.addWidget(QLabel('n-dim:'), 7, 0, 1, 3)
-        self.grid_layout.addWidget(self.ndimCheckBox, 7, 3)
+        # self.grid_layout.addWidget(QLabel('controls:'), 0, 0)
```
sofroniewn: delete commented line

napari/_qt/qt_viewer.py
```diff
@@ -58,10 +58,41 @@ def __init__(self, viewer):
...
+            area='left',
+            allowed_areas=['left', 'right'],
+        )
         self.dockConsole = QtViewerDockWidget(
-            self, self.console, name='console', shortcut='Ctrl+Shift+C'
+            self,
+            self.console,
+            name='console',
+            area='bottom',
+            # allowed_areas=['top', 'bottom'],
```
sofroniewn: i'm fine letting it be everywhere - don't have strong feelings, but we should either restrict it or not, no need for the comment

napari/_qt/qt_viewer_dock_widget.py
```diff
@@ -80,7 +97,94 @@ def keyPressEvent(self, event):
 
     def _set_title_orientation(self, area):
         if area in (Qt.LeftDockWidgetArea, Qt.RightDockWidgetArea):
-            features = self._initial_features
+            features = self._features
+            if features & self.DockWidgetVerticalTitleBar:
```
sofroniewn, line 101: syntactically can we use `and` and `or` instead of `&` and `^`? Curious what @jni and @kne42 think, but I find the explicit words easier to understand (particularly for new comers) and preferable

napari/util/theme.py
```diff
@@ -5,6 +5,7 @@
     'dark': {
         'folder': 'dark',
         'background': 'rgb(38, 41, 48)',
+        'background_dark2': 'rgb(33, 36, 42)',
```
sofroniewn: not sure I love the name `background_dark2`, maybe `darker_background` or something else is preferable

napari/_qt/layers/qt_labels_layer.py
sofroniewn: this maybe my fault here for leaving them in, but we can drop the commented `# self.grid_layout` lines

napari/_qt/layers/qt_points_layer.py
sofroniewn: as above, can drop the commented `# self.grid_layout` lines

sofroniewn (commented):
Looks great to me, love the functionality and stylings. I left some minor comments that maybe @tlambert03 you can address, but I think its good to go. I actually can't approve the PR as I started it so maybe you can or @kne42 or @jni can give this a review and approve too

royerloic:
Looks amazing guys! Congrats! Really, wow...

napari/_qt/qt_viewer_dock_widget.py
tlambert03: no, those are [bitwise operators](https://docs.python.org/3/reference/expressions.html?highlight=bitwise#binary-bitwise-operations) and are syntactically distinct from `and` and `or` ... Qt uses them a lot when representing combinations of constants.

napari/util/theme.py
tlambert03: yeah, I don't love it either.  `dark2` there means "2 percent darker" and I considered adding different perecentages, but leaving at just `dark2` made it look like it was just "option #2".  until we have additional ones, I can just make it `darker_background` with a comment that it is 2% so we don't forget

napari/_qt/qt_viewer.py
tlambert03: that was just for testing... will undo it now

napari/_qt/qt_viewer_dock_widget.py
sofroniewn: ah ok great, thanks for the clarification - fine to leave as it

napari/util/theme.py
sofroniewn: hmm yeah - i think `background_darker` + comment is a fine choice for now, and we might revisit this stuff when we do a global pass over the styling sometime in the future with a designer's help

tlambert03 (approved):
I think it's ready to go as well, so I'll approve since @sofroniewn opened the PR.  @sofroniewn you can merge it when you want, or we can wait until tomorrow in case someone who hasn't been staring at the code so much wants to review.

sofroniewn:
Thanks for taking this to the finish line @tlambert03 - it looks good to go from my perspective, but I'll leave it till tomorrow in case @jni or @kne42 want to weigh in and request changes.

Note that the windows fail is due to a `openssl.light` problem that is independent from this PR

jni:
I don't have time to do a proper code review but the latest status on the interactivity and look makes me super happy. I like the double click, even if it is not very discoverable, just like I liked 3D touch until Apple decided to do away with it. 😭

Currently the Windows tests are failing, not sure if it is related to this PR... But, as I said, functionally, imho this is ready to go, and super awesome.

tlambert03:
> I like the double click, even if it is not very discoverable

I added a tooltip, "drag to move. double-click to float" that might eventually help some people discover it if they rest their mouse long enough

sofroniewn:
Thanks for taking a quick look @jni, and glad you like the new functionality. I am going to merge now, @tlambert03 and I have been through this one quite thoroughly. The windows failures are unrelated.

tlambert03:
here's hoping for some slightly more "surgical" PRs in the near future :)

PR #728, opened 2019-11-24
Fix vertical scrollbars

tlambert03:
# Description
In working on some styles in #726, I realized the vertical scroll bars were messed up (even in `master`, probably in the recent dockable widget style changes.  This PR standardizes the stylesheet for all scrollbars, and makes sure the vertical scroll bars match the styles of the horizontal ones (including new `up_arrow` and `down_arrow` icons).  Previously, I think it was just using the native scroll for vertical bars.  Notably, this entailed removing the `QtDims >` prefix for many of the scrollbar styles, and only specifying `QtDims` for the dimslider specific styles.  It also moves the scrollbar styles to the bottom because some of the styles were cascading strangely.,

Additionally, the PR ignores `napari/resources/qt.py` from `flake8` and `black`, since it is autogenerated.  Merging this PR prior to #726 and possibly #727 will be useful, as there will likely be some conflicts to sort out.

Screen Shot 2019-11-24 at 10 29 11 AM

## Type of change

- [x] expanded styles

## Final checklist:


napari/resources/qt.py
```diff
@@ -2,21 +2,262 @@
...
 #
-# Created: Sat Oct 5 21:19:34 2019
-#      by: The Resource Compiler for PySide2 (Qt v5.12.3)
+# Created: Sun Nov 24 10:05:59 2019
+#      by: The Resource Compiler for PySide2 (Qt v5.13.2)
 #
 # WARNING! All changes made in this file will be lost!
 
-from qtpy import QtCore
+from PySide2 import QtCore
```
sofroniewn: you need to manually change this to `qtpy` for cross compatibility

napari/resources/stylesheet.qss
```diff
@@ -991,3 +915,121 @@ QDockWidget::float-button:horizontal {
...
+    background: {{ highlight }};
+}
+
+QtDims > QScrollBar::handle[last_used=true]:horizontal {
+    background: {{ secondary }};
+}
+
+QScrollBar::add-page:horizontal, QScrollBar::sub-page:horizontal {
+    background: none;
+}
```
sofroniewn: when you edit the stylesheet there never seems to be a new line added at the end, I know we aren't autolinting here, but how do you feel about it? Should we add?

sofroniewn (approved):
This all looks good to me, I'm approving so you can merge after this - but left two comments - one is critical and needs to be addressed (need to add `qtpy` instead of `PySide2` in the autogenerated file), the other is super minor and up to you.

It will be good to get this in before #727 and #726 as you note. Also probably good at some point soon to do some more clean-up / standardization of our stylesheets, at least just organizing where everything is in the file!

napari/resources/qt.py
tlambert03: ah!  thanks for catching that.  I'll add a test to make sure it gets caught in the future.

napari/resources/stylesheet.qss
tlambert03: arg sorry!  I've grown so used to auto-formatting that I miss it on that file every time.  might as well add it.

tlambert03:
added a test to make sure `resources/qt.py` imports from `qtpy` and not `PySide2`
Can't say I'm not a tiny bit nervous about merging into master  😬
I'll leave this for the rest of the evening here in case @jni or anyone else wants to comment.

sofroniewn:
This looks great @tlambert03 - just confirmed all works for me, and stylings look so much nicer now! Thanks for adding that test too. You can do the honors and merge whenever you are ready!!

tlambert03:
always wanted to say it.... LGTM! :)

Issue #736, opened 2019-11-25
big data + pyramid + contrast limits meta issue


sofroniewn:
## 🐛 Bug
We have a number of open issues right now related to big data + pyramid + contrast limits calculations, we've struggled with our api here for a while and it's clearly a major driver of current bug reports. I'm going collect references to all the issues below and I'd like us to spend a little time thinking about what the best overall solution is

- [x] #624 #570 contrast limits and range guessing
- [x] #691 integer overflow in `is_pyramid` check, solution in progress #692 
- [x] #699 and #710 - call to `is_pyramid` forces zarr decompression, shouldn't call when `is_pyramid=False`
- [ ] #718 - request for better dask caching
- [x] #731 - big 3D data failed because `contrast_limits` not provided
- [x] #733 - possible duplicate of #699 and #733

Looking over this is seems that we should only call `is_pyramid` if the `is_pyramid=None` keyword is None (I might also rename the function as it is confusing to share names like this). I will make this PR right now.

We should also do better is `contrast_limits` is not provided.

jni:
Thanks @sofroniewn! This has been causing me some anguish but I haven't found a moment to work on it. imho another tweak for the pyramid checking code is checking `hasattr(data, shape)`. If it has a shape, it must be regular and cannot be a pyramid. That should allow things to work with `is_pyramid=None`.

jni:
On the plus side, it's very cool that so many people are using napari to look at big microscopy data!

sofroniewn:
Unfortunately @jni we did have someone pass pyramid data as a 1-d numpy array with dtype object - see one of our old issues

sofroniewn:
I am working on a PR adding unit tests with the timing like you did for the your recent dask uint8 PR

jni:
@sofroniewn 

> Unfortunately @jni we did have someone pass pyramid data as a 1-d numpy array with dtype object - see one of our old issues

Ach! Users have such a way to find bugs. 😂 

How about:

`hasattr(data, 'shape') and hasattr(data, 'dtype') and data.dtype != 'object':`

?

sofroniewn:
@jni checkout my solution here - https://github.com/napari/napari/pull/737#issuecomment-558416914

sofroniewn:
Note also discussions in this issue - https://forum.image.sc/t/efficiently-open-qptiff-images-with-napari/33237/4?u=sofroniewn

[...]

PR #737, opened 2019-11-26
[WIP] redo pyramid guessing

### LINKED ISSUE #699: pyramid checking seems to be breaking lazy loading
## 🐛 Bug
I was just trying to familiarize myself with napari's handling of zarr datasets using one of my big light sheet datasets... but it would hang indefinitely each time I added it.

I went back to the example in `zarr_nD_image.py`, but there too, as i increase the size of the zarr store, napari starts to hang upon `napari.view_image(data)`... regardless of whether I provide the `contrast_limits`, `rgb`  or  `is_pyramid=False` arguments.  Digging a bit more, it seems that, as of #556, it is impossible to avoid the callchain from `Image.__init__()` -> `get_pyramid_and_rgb()` -> `is_pyramid()` ... and that `is_pyramid()` call takes forever on a big zarr store.

Is this a regression or is `zarr_nD_image.py` now invoking `napari.view_image` incorrectly?

### LINKED ISSUE #710: Handling of *compressed* in-memory zarr-arrays?
I have not found documentation on this issue, so please excuse this question.

I came across [this example](https://github.com/napari/napari/blob/4fbbca0dcdf9529ea55e08224e9962bf6f754bdf/examples/zarr_nD_image.py), which shows that `napari` can directly display in-memory zarr-arrays. In the example, the array is iniziallied like this: 

```python
data = zarr.zeros((102_0, 200, 210), chunks=(100, 200, 210))
```

`zarr` does in fact also allow compression for in-memory arrays (not only for on-disk storage). They can be initialized like this, for example:

```python
from numcodecs import Blosc
compressor = Blosc(cname = 'zstd', clevel = 3, shuffle = Blosc.BITSHUFFLE)
data = zarr.zeros((102_0, 200, 210), chunks = (100, 200, 210), compressor = compressor)
```

I am heavily using this feature in my day to day work for arrays that would, if handled uncompressed, out-grow a machine's memory by a factor of 2 to 10 (compression ratios of about 5x to 20x).

How does `napari` handle such datasets? Does it make sense to visualize them with `napari` or should I assume that most / all data is being decompressed in the process?

### LINKED ISSUE #733: Weird issue with dask/zarr datasets
I am experiencing a weird issue when trying to use napery to explore dask arrays stored in zarr files. I would like to integrate napari as the viewer for a calcium imaging data pipeline that I am building using dask (x)arrays stored in zarr datasets. Up to some size of the dask array, when I open the napari viewer everything runs smoothly. After increasing a dimension of the dask array, keeping the same chunking, when starting the viewer it tries to load the entire dataset in memory and crashes.

## To Reproduce
This creates dummy data similar to what I am using for my real case scenario:
```python
# Generate dummy dataset:
def get_dummy_dset(shape, chunks):
    dims = ["t", "z", "y", "x"]
    dask_data = da.random.random(shape, chunks=chunks)
    xdataset = xa.Dataset({"fluorescence": (dims, dask_data)}, 
                          coords={k: np.arange(dask_data.shape[i]) 
                                  for i, k in enumerate(dims)})
    return xdataset

# Save zarr groups with slightly different 1st dimension:
data_path = ".../somefolder/"
ds1 = get_dummy_dset((8100, 8, 100, 200), chunks=(50, 8, 100, 200))
ds1.to_zarr(data_path + "/example_dataset_large_good.zarr", 
            encoding=dict(fluorescence=dict(compressor=None)))

ds2 = get_dummy_dset((8500, 8, 100, 200), chunks=(50, 8, 100, 200))
ds2.to_zarr("{}/example_dataset_large_bad.zarr".format(data_path), 
            encoding=dict(fluorescence=dict(compressor=None)))
```

Then, I try to load them in napari with the following code:
```python
import napari
import xarray as xa

with napari.gui_qt():
    ds1 = xa.open_zarr(data_path + "/example_dataset_large_good.zarr")
    stack1 = ds1["fluorescence"]
    viewer = napari.view_image(stack1)
```

When loading the first dataset, everything run nicely. On the second one, only 5% bigger, napari tries to load everything in memory and it crashes. To my preliminary testing it seems that:
 - this is not related to absolute size: increasing of the same amount other dimensions does not look critical;
 - this is not related to chunk number, as different chunking does not fix it.

I experience this both on a macBook and on a Windows 10 desktop computer with different specs in memory and processor. 

Do you have any pointer on why/where this might be happening?

By the way, amazing project! I have been waiting a long time for something like this to happen, I would be happy to contribute with examples from our data analysis pipeline once I get more into it!

sofroniewn:
# Description
This PR aims to fix some of our recent big data problems - see #736 for a longer list.

It aims to close #699, close #710, and possibly close #733, by causing `guess_pyramid` (renamed from `is_pyramid`) only to run if the `is_pyramid` keyword arg is `None`. We'll still have problems if that keyword arg isn't set and someone is using big data, which is a bit of a shame, so we should probably improve that `guess_pyramid` too.

One option is to ask about `guess_pyramid` shape and dtype. If it has a shape and a non-object dtype then it cannot be a pyramid. This would solve the vast majority of simple cases, so I think I want to add that to this PR too.

Other suggestions welcome also.

## Type of change

# How has this been tested?

- [x] new tests added in `napari/layers/image/tests/test_big_images.py`

## Final checklist:


sofroniewn:
Ok - so I have improved the `guess_pyramid` so that it doens't choke with a big dask array by doing
```python
    # If the data has ndim and is not one-dimensional then cannot be pyramid
    if hasattr(data, 'ndim') and data.ndim > 1:
        return False
```

Unfortunately we now hit another problem - we added the ability for us to autogenerate a pyramid on axes when it was needed. This is nice, and necessary when looking at 2D data, but a bit weird if looking at a long timeseries axis. The currently problematic code comes after
```python
            pyr_axes = should_be_pyramid(data.shape)
```

We could try and wait to do that pyramid conversion only for displayed axes, but that starts becoming really weird for switching displayed dims.

What I don't like right now is that the main code path `add_image(big_data)` will break without extra keyword arguments, but a relatively low use case feature will be coerced to working. not sure how to fix, will think more

jni:
> We could try and wait to do that pyramid conversion only for displayed axes, but that starts becoming really weird for switching displayed dims.

Honestly autogeneration always seemed brittle. If needed (`should_be_pyramid` evaluates to True), I suggest we raise a warning or error for now, that describes how to generate one, including selecting which axes to use?

sofroniewn:
@jni how about we emit a warning and then downsample the only the view, this will cause no errors. Can include in that warning suggestion for how to make pyramid. I prefer this to autogeneration too which, agreed was a bit too much “magic”

sofroniewn:
The more I think about this actually the more I like the idea of adding a new keyword argument ‘force_pyramid=False’ that allows a non-coding user to leverage this functionality, but removes the magic of guessing. We can then downsample if needed and emit a warning to use this keyword arg to access full resolution data

jni:
> this will cause no errors. 

😂 the optimism! =P

Ok let me review this properly...

napari/layers/image/tests/test_pyramid.py
```diff
@@ -41,7 +41,7 @@ def test_error_pyramid():
     shapes = [(40, 20), (20, 10), (10, 5)]
     np.random.seed(0)
     data = [np.random.random(s) for s in shapes]
-    with pytest.raises(ValueError):
+    with pytest.raises(Exception):
```
jni, line 44: Why a bare exception? What is actually raised?

napari/util/misc.py
```diff
@@ -128,27 +132,57 @@ def get_pyramid_and_rgb(data, pyramid=None, rgb=None):
...
     data_pyramid : list or None
         If None then data is not and does not need to be a pyramid. Otherwise
         is a list of arrays where each array is a level of the pyramid.
     """
-    # Determine if data currently is a pyramid
-    currently_pyramid = is_pyramid(data)
-    if currently_pyramid:
-        shapes = [d.shape for d in data]
-        init_shape = shapes[0]
+    # If is_pyramd status is not set guess if data currently is a pyramid
```
jni:
```suggestion
    # If is_pyramid status is not set guess if data currently is a pyramid
```

napari/util/misc.py
```diff
@@ -128,27 +132,57 @@ def get_pyramid_and_rgb(data, pyramid=None, rgb=None):
...
+    if is_pyramid is None:
+        current_pyramid = guess_pyramid(data)
+        # current_pyramid = False
+        # If not currently a pyramid ask if needs to be
+        if current_pyramid is False:
```
jni: Everything in this clause is so messy. I suggest we turn off automatic pyramid generation for now, raise an informative error, and update to do it *properly* in a separate PR.

By properly, I mean, we should check what will actually fit in a texture (is this introspectable?) and limit there. One of the features I dislike about matplotlib image display is that it very aggressively downscales the input, and often entire lines disappear. This makes it quite annoying to teach about modifying images, for example, since often the modifications are single-pixel thick.

In short, I'd rather we error-out with an informative message about what to do about it as a user, instead of having a hacky first-pass solution that doesn't always work properly.

jni (commented):
@sofroniewn I've made one major comment. I think it fits in with your thoughts about how to handle this going forward, as by my understanding this code does not implement the optional switch, but let me know if not...

sofroniewn:
@NHPatterson we ran into some git problems with #692 that had to be reverted in #738 - all my fault, sorry. Given that we've got this PR in progress addressing similar things I've just added your fix from #692 directly to it, I hope this is ok. #692 did get merged so you're still listed as a contributor the the repo, but this seems like the cleanest fix for now. Sorry again about this, and thanks @AhmetCanSolak for helping sort this out!!

napari/util/misc.py

napari/layers/image/tests/test_pyramid.py
sofroniewn: Uninformative errors are raised - here we're basically pretending a pyramid data set is not a pyramid, but we removed our guess_pyramid from being called as it was too expensive. I'm not quite sure how to improve

sofroniewn:
Ok @jni I admit I may be turning this PR into a monster, which I want to avoid, but I've done some of the stuff we discussed above but not all of it.

Good things I think I have done here are pull out a lot of this image shape guessing / computation code into it's own file `image_shape.py` instead of being in `misc.py` and added its own set of tests which should make maintaining this stuff much easier.

The main function in pyramid function - `image_shape.py` now looks as follows:
```python
get_pyramid(data, is_pyramid=None, force_pyramid=False)
```
Where `force_pyramid` is an option. I **have not** exposed `force_pyramid` all the way up at `add_image` or `view_image` (though that change is the easiest one to make, I want to let you weigh in first as your last comment was that we should wait on that).

I have done the grabbing of `MAX_TEXTURE_SIZE` from vispy, but as you can see in the code right now it might be a little ugly as I have to wait until we actually show the window with the canvas for it to be provided, and I then have to propagate it through the vispy code.

What that also means is that a function like `should_be_pyramid` inside `util.image_shape.py` which is getting called from our model code doesn't know what that value is, and in many ways it is questionable if it should (i.e. why should a view parameter impact what we do with our models).

I do now emit a warning and downsample if the max texture size is exceeded. Otherwise vispy will just crash with the following:
```
RuntimeError: OpenGL got errors (Check before draw): GL_INVALID_VALUE
```
I think downsample + custom warning is preferable to crash, but we could also do custom error message with clear instructions.

I do have really mixed feelings about ever providing a `force_pyramid` keyword on `add_image` and `Image` - unclear if it is better to live on `add_image` maybe and not `Image` - it might make `Image` a bit cleaner, but at the expense of `add_image` and consequently `add_labels` being more complicated.

I think for programming users we can always just say - do this pyramid conversion outside napari, but for people using the stand alone app who have a 20k x 20k image right now they can load that into memory but we'd either crash, downsample, or pyramid generate.

@tlambert03 I'm really curious what your take on all this is too and would love you to weigh in.

@jni if you want me to cull functionality from this PR so we get this in really clean and then add in functionality in follow-up PRs that's fine, but I think we need to have something we feel really good about in this area before we can make the next release.

I also want to add more comprehensive tests around all this stuff too to avoid all the bug reports we have been getting

sofroniewn:
ah our windows tests are failing because they can't capture the max_texture_size and I coded the catch wrong (assumed `None` but should be `()` when no texture is found) that's an easy fix

napari/_qt/qt_main_window.py
```diff
@@ -265,6 +265,10 @@ def show(self):
         self._qt_window.show()
         # make sure window is not hidden, e.g. by browser window in Jupyter
         self._qt_window.raise_()
+        # Update to MAX_TEXTURE_SIZE needs to come after window is shown
+        self.qt_viewer.MAX_TEXTURE_SIZE = self.qt_viewer.canvas.context.shared.parser.capabilities[
+            'max_texture_size'
+        ]
```
jni: @sofroniewn why does this need to come after the window? It seems that vispy has some ways to introspect this value, see:

https://github.com/vispy/vispy/issues/1141

(And note from that thread that this value is different for 2D and 3D...)

napari/_vispy/vispy_image_layer.py
```diff
@@ -74,6 +76,24 @@ def _on_data_change(self):
...
+                    f"data shape {data.shape} exceeds GL_MAX_TEXTURE_SIZE"
+                    f"{self.MAX_TEXTURE_SIZE} in at least one axis and will be "
+                    f"downsampled."
+                )
+                downsample = np.ceil(
+                    np.divide(data.shape, self.MAX_TEXTURE_SIZE)
+                ).astype(int)
+                scale = np.ones(self.layer.ndim)
+                for i, d in enumerate(self.layer.dims.displayed):
+                    scale[d] = downsample[i]
```
jni: This seems like it would break the existing scale that a user might have set? Perhaps multiply by downsample[i]?

jni: (And this is the sort of thing that I meant when I said that doing this correctly is difficult.)

napari/_vispy/vispy_vectors_layer.py
```diff
@@ -4,9 +4,9 @@
 
 
 class VispyVectorsLayer(VispyBaseLayer):
-    def __init__(self, layer):
+    def __init__(self, layer, MAX_TEXTURE_SIZE=None):
         node = MeshNode()
-        super().__init__(layer, node)
+        super().__init__(layer, node, MAX_TEXTURE_SIZE)
```
jni: I'm not a fan of being able to set the value here (and elsewhere)! Is this necessary? Isn't it a limitation of the graphics card?

napari/layers/image/tests/test_pyramid.py
jni: What is the *class* of error? We should check for that class rather than "Exception", as checking for Exception can mask unexpected errors.

napari/layers/image/tests/test_pyramid.py
```diff
@@ -59,19 +59,13 @@ def test_infer_tuple_pyramid():
...
     data = np.random.random(shape)
-    layer = Image(data, is_pyramid=True)
-    assert np.all(layer.data == data)
-    assert layer.is_pyramid is True
-    assert len(layer._data_pyramid) > 0
-    assert layer.ndim == len(shape)
-    assert layer.shape == shape
-    assert layer.rgb is False
-    assert layer._data_view.ndim == 2
+    with pytest.raises(Exception):
```
jni, line 67: As above.

napari/_vispy/vispy_image_layer.py
sofroniewn: Totally agree on the complexity here, this will be ok as long as the data is not also a pyrmaid!!! `_scale_view` is a separate scale factor that we use the pyramid stuff that I'm leveraging here too. I could add an explicit step which throws an error if you are a pyramid and your tile size is too big!!!

napari/_vispy/vispy_vectors_layer.py
sofroniewn: yeah agreed - I don't like this either, will try and fix

napari/_qt/qt_main_window.py
sofroniewn: `sys_info` actually creates a little fake canvas in order to do this - https://github.com/vispy/vispy/blob/a6ec64bb7b56739b463e3c16245a31940ca4f5f7/vispy/util/config.py#L431 which we could do too once on start-up - though probably don't want to put inside our "model" code. I have an idea on how to at least improve what we have now significantly.

Thanks for catching `GL_MAX_3D_TEXTURE_SIZE` I can work with that too

napari/layers/image/tests/test_pyramid.py
sofroniewn: ok can improve - this might change by the end of the PR - lol

napari/_qt/qt_main_window.py
sofroniewn: so vispy does not actually expose `GL_MAX_3D_TEXTURE_SIZE` - I'm not quite sure why, I need to do some investigating before asking

napari/_vispy/vispy_image_layer.py
sofroniewn: I've added a catch in case a pyramid ever fails this test - but that is something entirely within our hands - as we control `layer._max_tile_shape` currently `1600`. To see why we don't need the multiplication here, it happens bellow:
https://github.com/napari/napari/blob/67789fba557ae58436437f4c8e3c123c5fbab22c/napari/_vispy/vispy_image_layer.py#L117
where the user provided `scale` interacts with our `_scale_view` that can handle additional offsets. It should work fine, but agree this could have hidden bugs in edge cases

napari/_qt/qt_main_window.py
sofroniewn: fixed, though see note about `GL_MAX_3D_TEXTURE_SIZE` not being exposed

napari/_vispy/vispy_vectors_layer.py
sofroniewn: fixed

sofroniewn:
So @jni I've made some improvements to this PR including adding a test with numpy data that exceeds the max texture sizes. Without the catches and downsampling those tests fail.

The 2D max texture size is now automatically grabbed by vispy. I havn't yet been able to grab the 3D max texture size, but i can look into that.

I very much like the downsampling now on the vispy side of things. It only affects the view not the model, it emits an informative warning, and it prevents crashes.

What I'm not sure about is the `force_pyramid` functionality. In many ways it is independent, but in other ways it is useful if people want to view their full resolution data. One thing right now is that its strange for us to get our model code to read the max texture size, and how one might want to do `force_pyramid` might depend on how they want to do the rendering. One option is to drop the `should_be_pyramid` method and have `force_pyramid` be an integer cut off if you want to force pyramiding, or a list of integers cut offs, so by default `force_pyramid=False`, but you could do `force_pyramid = 16_834` or `force_pyramid = [False, 16_834, 16_834]` etc.  because of the downsampling on the vispy side we'll never actually crash, but this could just be a nice way for people to see their big high resolution data without them having to worry about it. I might work on implementing this now, but curious what others think

jni:
Ok so part of the problem I have with this PR is that I myself am very confused about what is happening, what the flags and functions actually mean, and what will happen with each combination of options — there's 6, not counting axes!

In fact, as far as I can tell, literally the only two lines of this PR than are needed to fix all three issues that this PR closes are:

```
    if hasattr(data, 'ndim') and data.ndim > 1:
         return False
```

IMHO, we should merge a separate PR with those two lines and a test, and then we can return to all the fancy auto-generation, and take the time required to do it properly. The above, though, needs a release ASAP.

Regarding the rest of the functionality, I am very confused by the name `force_pyramid`. Does it mean that a pyramid *must* be provided? Does it mean that it *must* be generated? What if it is already generated? Would `auto_pyramid` be a better name? Or should we make a new string-enum: `pyramid={'pre-computed', 'guess', 'no', 'auto-generate'}`, defaulting to no?

My final concern is that our pyramid performance is not great. There can be noticeable pauses when crossing tile boundaries at the highest-resolution level, for example. We need to improve how tiles are loaded asynchronously. So, I wonder how many people are actually using our pyramids in production, compared to per-slice lazy loading, which seems to be very popular. So I would make all of the auto-generation ideas part of a bigger overhaul of our pyramid handling.

sofroniewn:
> In fact, as far as I can tell, literally the only two lines of this PR than are needed to fix all three issues that this PR closes are:

lol :-) yes that is probably true. I do like the refactoring to group all this code together outside of `util/misc.py` into its own file. That `misc.py` file is pretty scary right now and I think having it all mixed in there contributes to us struggling with this functionality. I also like the tests that I have added.

I also like the vispy texture changes which are preventing bugs that could occur on master right now (if someone does a 3D image with an axis > 2048 it crashes right now).

I think you're main source of concern is `force_pyramid` and that kinda crazy 6 clause if statement in `get_pyramid`. How about we drop the `force_pyramid` arg, and I clean that up. I can also then drop the `should_be_pyramid` function, `make_pyramid` function, and `fast_pyramid` which will no longer be used, or I can add a comment that they aren't being used and let them sit there.

Sound reasonable?

tlambert03:
> IMHO, we should merge a separate PR with those two lines and a test, and then we can return to all the fancy auto-generation, and take the time required to do it properly. The above, though, needs a release ASAP.

this is how I feel too.  part of the reason I haven't chimed in yet is that I wanted to take the time to understand what exactly was going on... but I haven't found the time yet.  The autogeneration stuff based on texture memory does seem pretty promising though!  but if we can get a stop-gap in that prevents 90% of the issues, that would be great

jni:
@sofroniewn 

> How about we drop the force_pyramid arg, and I clean that up.

I don't mean to discard all the work you've done here! I think it's valuable *long term*, but all those issues need to be fixed ASAP because many people are running into them. We've had at least three independent reports, and there's probably more — people who are trying it, having it fail, and giving up on napari altogether! So I think a separate, tiny PR is warranted, and let's take time to discuss everything else in this PR properly in the coming weeks.

sofroniewn:
I've updated this PR with changes from master and it is ready for more discussion. This is still a critical issue, as noted in https://github.com/napari/napari/issues/733#issuecomment-559936312 our short-term fix in #740 has not resolved all issues.

This PR currently does three things though, and it might be easier for everyone to review is we separate them out a bit, or if people are ok with it we can leave it like this:

- adds MAX_TEXTURE_SIZE limits and downsampling, and adds test. This could easily be turned into it's own PR.
- removes all shape / pyramid guessing stuff from `util/misc.py` and puts it in a new file `util/image_shape.py`. Does the same for tests. This could also be done in it's own PR with no changes to functionality. It will make the diff much easier to understand when we then do want to change functionality. Let me know if this is desirable.
- adds a `force_pyramid` keyword argument to our `get_pyramid` method. This argument is always set to `False`, but it could be exposed and gives people a sense of what the code would look like with it in.

We definitely shouldn't rush this, but getting this right I think is incredibly important, which is why taking a modular / stepwise approach is fine with me. @tlambert03 @jni what would you advise next?

sofroniewn:
This has been superseded by #1159 and other PRs

PR #740, opened 2019-11-28
improve pyramid guessing

### LINKED ISSUE #699: pyramid checking seems to be breaking lazy loading

sofroniewn:
# Description
This PR will close #699 and #710 and help with #733. See #736 for a meta issue describing our pyramid guessing and other big data problems right now.

It still won't solve all problems, for example those in #733 related to the auto-pyramid generation and `contrast_limits`, but it is a relatively minimal change for fixing a lot of bugs.

I've also added a bunch of tests, including 2 which currently fail but I skip because I think they will be important for us to ultimately support

## Type of change

### DIFF
--- a/napari/layers/image/tests/test_big_image_timing.py
+++ b/napari/layers/image/tests/test_big_image_timing.py
@@ -1,32 +1,40 @@
 import dask.array as da
+import zarr
 from napari.layers import Image
 import pytest
 
 
-data = da.random.random(size=(100_000, 1000, 1000), chunks=(1, 1000, 1000))
+data_dask = da.random.random(
+    size=(100_000, 1000, 1000), chunks=(1, 1000, 1000)
+)
+data_zarr = zarr.zeros((100_000, 1000, 1000))
 
 
 @pytest.mark.timeout(2)
-def test_timing_fast_big_dask_all_specified_():
+@pytest.mark.parametrize('data', [data_dask, data_zarr])
+def test_timing_fast_big_dask_all_specified_(data):
     layer = Image(data, is_pyramid=False, contrast_limits=[0, 1])
     assert layer.data.shape == data.shape
 
 
 @pytest.mark.timeout(2)
-def test_timing_fast_big_dask_is_pyramid_specified():
+@pytest.mark.parametrize('data', [data_dask, data_zarr])
+def test_timing_fast_big_dask_is_pyramid_specified(data):
     layer = Image(data, is_pyramid=False)
     assert layer.data.shape == data.shape
 
 
 @pytest.mark.skip(reason="currently fails as we try and generate pyramid")
 @pytest.mark.timeout(2)
-def test_timing_fast_big_dask_contrast_limits_specified():
+@pytest.mark.parametrize('data', [data_dask, data_zarr])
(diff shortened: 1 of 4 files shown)


napari/layers/image/tests/test_big_image_timing.py
```diff
@@ -0,0 +1,32 @@
+import dask.array as da
+from napari.layers import Image
+import pytest
+
+
+data = da.random.random(size=(100_000, 1000, 1000), chunks=(1, 1000, 1000))
```
jni: @sofroniewn the array here should actually be a zarr array and not dask: dask always worked fine with `is_pyramid()`. If you use `zarr.zeros` it should not take up any memory.

napari/util/tests/test_misc.py
```diff
@@ -119,6 +119,22 @@ def test_is_pyramid():
     )
     assert is_pyramid(data)
 
+    s = 8192
+    data = [da.ones((s,) * 3), da.ones((s // 2,) * 3), da.ones((s // 4,) * 3)]
+    t0 = time.time()
+    val = is_pyramid(data)
+    t1 = time.time()
+    assert val
+    assert t1 - t0 < 2
```
jni: Can you make these separate tests?

jni (approved):
@sofroniewn I'm approving because I'm about to go AFK, but I made what I think is a major comment: the tests should use zarr arrays, rather than dask arrays, because dask always worked: it's zarr arrays that were instantiated by `[d.shape for d in data]`.

jni:
I recommend making sure that whatever tests you add don't pass on the current master before merging.

napari/layers/image/tests/test_big_image_timing.py
sofroniewn: hmm I do rememeber that, I had just checked that both those fail with a big dask array too. I'll add zarr.zeros too. Nice to check both

napari/util/tests/test_misc.py
sofroniewn: will do

sofroniewn: will use the timing functionality too

sofroniewn: improved

sofroniewn:
@jni this has now been improved and includes the changes from #692. The added tests previously failed, and now test zarr and dask. I will merge once the tests pass

sofroniewn:
merging now

PR #674, opened 2019-11-10
add logo

sofroniewn:
# Description
This PR adds a png of the napari logo to the repo

Issue #567, opened 2019-10-03
Show logo during startup


sofroniewn:
## 🚀 Feature
Right now napari takes a long time to launch the viewer. People can get discouraged as they wait, unsure if napari is loading, or if their data is loading, or if it is just hanging. Showing the napari logo, possibly with / without some text during that time might make it feel better.

jni:
Just to clarify here: I think we should only do this in `__main__` or even only in the bundled app. It would be very jarring to have this happen when launching from an IPython session. The correct solution for IPython is to dramatically reduce the startup time. To reiterate what I mentioned in the meeting, startup time from the command-line is about 2s on Linux, which is reasonably fast. There's nothing fundamentally slow about our startup, and we need to check what's happening on mac to make it so slow.

AhmetCanSolak:
We can do it in bundled app with a runtime-hook which will be executed right before execution of our main entry point @jni

tlambert03:
@sofroniewn pointed me to this from #744.  Just wanted to hear exactly when people feel like it's slow?  I think the majority of that time might be qt itself spinning up or imports (which is harder to do anything about).  On my system, when running from `__main__`, the majority of time is in `__init__`:
```
start init:    0.000
finish init:   1.413
start main:    1.416
start gui_qt:  1.417
app ready:     1.601
image shown:   2.281
```

So even if we wanted to do something with, say, [QSplashScreen](https://doc.qt.io/qt-5/qsplashscreen.html), that splash wouldn't show up until we instantiate the app in main, and then only very briefly... unless we tried to do something completely independent of Qt, at the very beginning of `__init__`.  I really haven't minded the startup time much, @sofroniewn, is it still slow for you, and under what specific startup procedure?

sofroniewn:
That's great profiling @tlambert03 - how did you do it? I get about `3.2` seconds between hitting enter on the command line after typing `napari` and seeing something on my screen (timed using the stop watch on my phone and me pressing buttons). That becomes a little longer if I try to load a small image at the same time too.

If we could get something up at the end right after app ready it still might feel nicer. There are also times when people are launching from scripts doing more fileIO etc inside our `gui_qt` event loop so it might be nicer there too. Putting it there would also mean people launching from jupyter or IPython wouldn't see it too, which was a concern from @jni, and makes sense to me - I think expectations are different when launching from the command line vs an interactive session.

[...]

PR #744, opened 2019-11-28
Add dock icon

tlambert03:
# Description
adds the napari logo to the running application icon... cause who doesn't want this?

Screen Shot 2019-11-28 at 7 13 11 AM

I added the logo.png to `napari/resources/` ... so if this is to be merged, we should discuss whether that's an ok location for that.

side note:  in previous applications, I have used the following code in windows... but it's not strictly necessary for the icon to show up.  I can't remember now why I had to do it.  but here's the [windows docs](https://docs.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-setcurrentprocessexplicitappusermodelid)
```python        
if sys.platform.startswith("win32"):
    import ctypes

    id_ = f"napari.napari.{__version__}"
    ctypes.windll.shell32.SetCurrentProcessExplicitAppUserModelID(id_)
```

## Type of change

# How has this been tested?
Tested on mac and windows... not yet on linux

## Final checklist:


sofroniewn:
@tlambert03 love this functionality, will give it a try now - but I think I already added the logo to the repo in #674 to the exact location you mentioned. Does it not show up for you / is there a reason we need to add it again?

tlambert03:
in my repo, the logo is in `resources` at the top level... and I'm not sure that gets shipped with the pypi package?  I put it in `napari/resources`

sofroniewn (approved):
Ah thanks!!! Let's drop the copy at the top level then. No need for it to be there twice, and shipping with the repo is what we'll want when we make it an icon for the application etc.

I can confirm the functionality works great for me on mac. I am approving, but we should let @jni test on linux before we merge. This is so nice to have

tlambert03:
Just tested successfully on Ubuntu 18.04

sofroniewn:
Ok nice, this has been tested on all major systems - will merge

sofroniewn:
Ah wait - I just saw @tlambert03 you also moved `screenshot-add-image.png` into the main repo. I think this can remain outside the repo - no need for us to distribute that image with the main package. Can you move it back and revert the path change in the readme. Then I will merge

sofroniewn:
Also @tlambert03 curious if you want to weigh in on #567 at some point in time

sofroniewn (changes requested):
see request above for only moving `logo.png` and not `screenshot-add-image.png` into the main repo

tlambert03:
all set!  will weigh in on 567 over there.

sofroniewn:
Thanks so much, will merge once the tests pass!

sofroniewn (approved):
Thanks so much, will merge once the tests pass!

sofroniewn:
The windows fails here are independent of this PR so I am going to merge

PR #745, opened 2019-11-30
show splash screen

### LINKED ISSUE #567: Show logo during startup

sofroniewn:
# Description
This PR closes #567 by showing the napari logo during startup on a `QSplashScreen`. It's fairly unobtrusive and I kinda like it. The splash screen only appears when launching from script / command line - and not from IPython or jupyter (not 100% sure why but that is the observed behaviour, and seems to be what @jni and @tlambert03 want).

We can also add a message to the splash screen using `showMessage` but not sure if we need that yet.

Here is launching from a script with splash screen:
splash_screen_script

Here is launching from IPython, no splash screen:
splash_screen_ipython

What do people think about this? @AhmetCanSolak @royerloic? 

## Type of change

### DIFF
--- a/napari/__main__.py
+++ b/napari/__main__.py
@@ -50,7 +50,7 @@ def main():
         'only a single image is given.',
     )
     args = parser.parse_args()
-    with gui_qt():
+    with gui_qt(startup_logo=True):
         v = Viewer()
         if len(args.images) > 0:
             images = io.magic_imread(
--- a/napari/_qt/event_loop.py
+++ b/napari/_qt/event_loop.py
@@ -7,12 +7,12 @@ from qtpy.QtWidgets import QApplication, QSplashScreen
 
 
 @contextmanager
-def gui_qt(splash_screen=True):
+def gui_qt(*, startup_logo=False):
     """Start a Qt event loop in which to run the application.
 
     Parameters
     ----------
-    splash_screen : bool
+    startup_logo : bool
         Show a splash screen with the napari logo during startup.
 
     Notes
@@ -23,11 +23,11 @@ def gui_qt(splash_screen=True):
     ``ipython --gui=qt``.
     """
     app = QApplication.instance() or QApplication(sys.argv)
-    if splash_screen:
+    if startup_logo:
         logopath = join(dirname(__file__), '..', 'resources', 'logo.png')
         splash_widget = QSplashScreen(QPixmap(logopath).scaled(400, 400))
         splash_widget.show()
     yield
-    if splash_screen:
+    if startup_logo:
         splash_widget.close()
     app.exec_()
--- a/napari/_qt/event_loop.py
+++ b/napari/_qt/event_loop.py
@@ -1,7 +1,9 @@
 import sys
 from contextlib import contextmanager
+from os.path import dirname, join
 
-from qtpy.QtWidgets import QApplication
+from qtpy.QtGui import QPixmap
+from qtpy.QtWidgets import QApplication, QSplashScreen
 
 
 @contextmanager
@@ -16,5 +18,9 @@ def gui_qt():
     ``ipython --gui=qt``.
     """
     app = QApplication.instance() or QApplication(sys.argv)
+    logopath = join(dirname(__file__), '..', 'resources', 'logo.png')
+    splash_screen = QSplashScreen(QPixmap(logopath).scaled(400, 400))
+    splash_screen.show()
     yield
+    splash_screen.close()
     app.exec_()
--- a/napari/_qt/event_loop.py
+++ b/napari/_qt/event_loop.py
@@ -7,8 +7,13 @@ from qtpy.QtWidgets import QApplication, QSplashScreen
 
 
 @contextmanager
-def gui_qt():
+def gui_qt(splash_screen=True):
     """Start a Qt event loop in which to run the application.
+
+    Parameters
+    ----------
+    splash_screen : bool
+        Show a splash screen with the napari logo during startup.
 
     Notes
     -----
@@ -18,9 +23,11 @@ def gui_qt():
     ``ipython --gui=qt``.
     """
     app = QApplication.instance() or QApplication(sys.argv)
-    logopath = join(dirname(__file__), '..', 'resources', 'logo.png')
-    splash_screen = QSplashScreen(QPixmap(logopath).scaled(400, 400))
-    splash_screen.show()
+    if splash_screen:
+        logopath = join(dirname(__file__), '..', 'resources', 'logo.png')
+        splash_widget = QSplashScreen(QPixmap(logopath).scaled(400, 400))
+        splash_widget.show()
     yield
-    splash_screen.close()
+    if splash_screen:
+        splash_widget.close()
     app.exec_()
--- a/napari/viewer.py
+++ b/napari/viewer.py
@@ -1,7 +1,7 @@
 from os.path import dirname, join
 
-from qtpy.QtGui import QIcon, QPixmap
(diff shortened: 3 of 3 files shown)


sofroniewn:
Ultimately I think we will want this for the standalone app for sure, the question is then do we also want it when launching from CLI / script - I'd say yes, but as noted above curious what others think too and if people don't like it / find it too much

tlambert03:
I do like the visual effect, but for me, it adds about 1 second to loading (with splash: 3.3 sec from hitting enter to seeing the image; without splash: 2.2 sec), which I don't think is a worthwhile tradeoff.

Also, note, putting the splash screen in `viewer.Viewer` will show the splash even in ipython (and presumably notebook) the first time someone uses the `gui_qt()` context manager (rather than `ipython --gui=qt`)... which is also not great.

I think the reason for the variable behavior in showing the screen is in this line from the QSplashScreen docs:
>the splash screen is typically displayed before the event loop has started running

So maybe that eventloop has already started by the time the napari viewer is instantiated when using notebook or `ipython --gui=qt`, but when using `napari.gui_qt():`, you catch that moment between creating `QApplication()` in `event_loop.py`, but before the Qt loop has begun.

sofroniewn:
Hmm adding 1 sec just to see the splash screen doesn't seem great! If people use `gui_qt` in the notebook / IPython it don't really mind also showing the screen, but fine if others prefer not to see it. We could add an optional keyword argument to show it and then just set that `True` when launching from the cli (i.e. so not even show it when launching from scripts). This would then allow us to have it for the standalone app, but maybe we don't even want it there.

I'm happy to let this sit for a little as people decide if they like it or not, doesn't seem top priority

AhmetCanSolak (approved):
I like how simple it is. In terms behavior I agree with @sofroniewn . We are already promoting only one API for notebooks, if users choose to not follow it sounds okay to show splash screen to me.

jni:
As a strict opt-in I like it, otherwise, I don't love it, and as @tlambert03 notes, it's not worth adding a full second of extra launch time — even in the app. We should aim for load times approaching 0, though I realise that appears hard/beyond our control on macOS right now. If we can figure out how to make this approximately instant, I'm ok with having it with the packaged app. 

I'll also note that I kinda like Ps/Ai's splash screens with photo/artwork in the splash screen. We could find a picture to "represent" the napari brand.

sofroniewn:
> As a strict opt-in I like it, otherwise, I don't love it

Not sure if this is what you meant, but I'm not sure I really see adding a keyword argument the `Viewer` as a good choice, that seems like cluttering the api for little gain.

We can also wait to revisit this with the packaged app. I definitely don't feel super strongly about this

tlambert03:
whatever the final implementation ends up being, I think the splash screen should only be used if it actually comes up during loading, and not in addition to loading.  Currently, I think it's appearing more or less when the viewer would have appeared (minus file reading time), and then just showing for about 1 second before yielding to the viewer.  To get something that actually shows quickly _while_ the app is loading, we might even need to do it outside of Qt unfortunately (though it may not even be possible without resorting to some other gui kit which would be silly)

sofroniewn:
those are pretty compelling points @tlambert03 - I don't really like the idea of going outside Qt for this, seems like a lot of overhead for something not that critical. I might close this and the issue then, no point in having things cluttered and we can always reopen or revisit when we finish up #496 and the bundled app

tlambert03:
was definitely nice to see how easy the QSplashScreen is though!  When revisiting, it's possible that if we instantiate the `QApplication()` before some of the heavier imports, we could get that splash screen up nearly immediately?

sofroniewn:
And we're back! That was a good suggestion @tlambert03 - I've now added the splash screen directly to the `gui_qt` event loop, so no chances it gets called if not using that event loop and it now appears I think much faster, though maybe you could give it a try. I've also added an optional keyword argument to disable showing the splash screen. Because it goes to `gui_qt` I'm actually no longer worried about cluttering the API. I've made it default to on, so if people really don't like it they can turn it off

jni:
A few suggestions and observations!

## 0: the script

I used the following script to time launch with and without splash and measure how long each took:

```python
from skimage import data
import napari
import time

t2 = time.time()
with napari.gui_qt(False):
    napari.view_image(data.astronaut(), rgb=True)
    t3 = time.time()
    print(f'no splash took {t3 - t2} seconds')

t0 = time.time()
with napari.gui_qt(True):
    napari.view_image(data.astronaut(), rgb=True)
    t1 = time.time()
    print(f'splash took {t1 - t0} seconds')
```

## 1: speed

The splash screen now takes very little time, on Linux at least. At first I thought it was ~2x slower, than launching without, but it turns out that the first launch of napari is significantly slower than the second launch, so swapping the order makes the first call slower, regardless of splash screen. That's worth noting for future investigation.

```
 ~/projects/napari/examples (pr/745)
 $ python splash-timer.py 
splash took 0.49181127548217773 seconds
no splash took 0.26149511337280273 seconds
(f7) 
jni@jni-x1c6 Sun Dec 01 16:29 
 ~/projects/napari/examples (pr/745)
 $ vim splash-timer.py 
(f7) 
jni@jni-x1c6 Sun Dec 01 16:29 
 ~/projects/napari/examples (pr/745)
 $ python splash-timer.py 
no splash took 0.44128870964050293 seconds
splash took 0.2785491943359375 seconds
```

Also, compare those times to macOS...! It's worth figuring out why macOS is so much slower... It's really quite a different experience on Linux!

## 2: the code

This should be a keyword-only argument. Compare the following two calls for a newbie:

```python
with napari.gui_qt(True):
```

and

```python
with napari.gui_qt(show_splash=True):
```

With the first one, there is just no way you could deduce what `True` means without looking at the docs. The second call is completely self-documenting.

Anyway, I want to reiterate that I would strongly prefer for this to be off by default, but on with the bundled app. At best, we should do a platform check and have it off for Linux. Boot up is fast so it just looks kind of glitchy to have the splash screen pop up for a fraction of a second. On mac, where boot up is slow, maybe it makes sense. I'll let @tlambert03 weigh in on which experience is better, because I don't have my mac on me right now.

sofroniewn:
agreed on the keyword only arg aspect. I've also renamed it `startup_logo` as I think splash screen is a little jargony (I had never heard the word before @tlambert03 pointed me to the Qt function).

I've also made the default `False`, but explicitly set it on in `__main__.py` this means that launching from the CLI with `napari` will show it, but otherwise you won't see it in a python script unless you ask for it.

As to speed - on my mac I get:
```
no splash took 1.33648681640625 seconds
splash took 0.31774091720581055 seconds

splash took 1.3581428527832031 seconds
no splash took 0.28098607063293457 seconds
```
so ~ 1 frame at 30Hz difference which I think is a tolerable lag.

@jni thoughts?

jni:
Those are my thoughts. =P

More seriously, I still don't love it as default from the CLI, but I'm reasonably ok with it, so let's just get on with it! If people are uncomfortable with it after some use, it is easy enough to revert that one line. Thanks!!!

royerloic:
A few thoughts:
i) I would limit the display to once per process. So after the first time a process starts a napari instance, the next time there is no splash screen (set global flag)
ii) I would have the logo with napari written below so it is clear what that's about :-)

sofroniewn:
@royerloic (i) should now be handled with placing the call in `gui_qt`, and we can do (ii) in a short followup PR sometime

Issue #470, opened 2019-08-09
Clicking the "add labels" button when viewing a lazy (zarr/dask) image should create a zarr-backed labels array.


jni:
## 🚀 Feature

Currently, add_labels will try to create an in-memory labels array of the same shape as the currently viewed image. This will break if the image layer is a bigger-than-memory dask array. A possible solution is to create an out-of-core dataset backed by zarr.

sofroniewn:
@manzt do you have any thoughts on if you could make a zarr "plugin" for labels painting? maybe in the first case you'd want to just explore the feature directly before seeing if it could be funneled through our current plugin API.

manzt:
I think I would need to take a closer look at the hook specification (since I've only implemented `napari_get_reader`), but zarr would be a very efficient way to back the labels array, especially if labels are sparse.

> Currently, add_labels will try to create an in-memory labels array of the same shape as the currently viewed image.

You could by default create an in-memory `zarr.Array` which is very light-weight:

```python
>>> import zarr
>>> labels = zarr.zeros(image.shape, dtype=??) # let zarr determine chunks
>>> labels
<zarr.core.Array (10000, 10000) int32
>>> isinstance(labels.store, dict)
True
>>> labels.store.keys()
dict_keys(['.zarray'])
```

By design, `zarr.zeros` is simply a creation method where only metadata for the array is saved to the underlying store, in this case a python `dict`. Chunks are only created on write, and if you try to access a missing chunks, zarr will fill using the `fill_value` (`0`) on read. 

```python
>>> labels[:]
array([[0, 0, 0, ..., 0, 0, 0],
       [0, 0, 0, ..., 0, 0, 0],
       [0, 0, 0, ..., 0, 0, 0],
       ...,
       [0, 0, 0, ..., 0, 0, 0],
       [0, 0, 0, ..., 0, 0, 0],
       [0, 0, 0, ..., 0, 0, 0]], dtype=int32)
```

This means you could create a label array with a massive shape and only run into issues with memory when you start to write lots of chunks. Fortunately with zarr, you can actually keep track of how many chunks have been initialized as well as the number of bytes currently stored. 

```python
>>> labels.info
Type               : zarr.core.Array
Data type          : int32
Shape              : (10000, 10000)
Chunk shape        : (625, 625)
Order              : C
Read-only          : False
Compressor         : Blosc(cname='lz4', clevel=5, shuffle=SHUFFLE, blocksize=0)
Store type         : builtins.dict
[...]

PR #748, opened 2019-12-01
labels file opening

sofroniewn:
# Description
This path begins to add file opening functionality to our labels layer by exposing the `path` keyword arg on `view_labels` and `add_labels`. This will then get exposed in #616 or follow up PRs through our open dialog box.

Right now it just reuses the same file loading as with the image layer - which I think is fine. A question though is do we additionally want to do some type checking / casting of data into integers >= 0 (which is what our labels layer requires). If so, probably best to put it inside the init of `Labels` as no reason not to do that checking if someone just does `add_labels(data)`, but I also don't want us to do unnecessary things that force computation when someone is say passing a massive dask array or zarr pyramid etc.

I know @jni and @perlman are currently working on #713 so curious how they think about this? If someone passes floats or negative numbers is that ok / their problem? If so then all this PR needs are tests (which could now be identical to the image IO tests / very simple).

@jni I also couldn't find an example of a labels like `png` or `tif` in `skimage.data`, do you know of a good one?

## Type of change


jni:
@sofroniewn no, I don't know that we have saved a labels image in skimage! One option would be to apply `morphology.watershed(filters.sobel(data.coins()), markers=200)` and then try a round trip.

I don't have a very strong opinion about forcing labels to be a specific type. Casting should only happen when viewing (ie in view_slice rather than in data itself) anyway, and should be fast. Negative labels is probably something someone will want at some point, and I'm ok with that. I'll less ok with float labels.

perlman:
Since you asked for my 2¢, I suggest supporting _any_ image type as a label channel. Labels are about uniqueness, and you are likely encounter {1,8,16,32,64)-bit (u)ints. If we can come up with a clean strategy for hashing a given data type, then it should be usable as a label.

Also, the current assumption that "0 is None" should be changed such that "arbitrary set of values are None", e.g., -1, MAX_INT, etc.

sofroniewn:
Thanks for weighing in @jni @perlman. I have left `Labels` such that it can still take anything, and agree we should behave gracefully with non-canonical input types, making sure that if we need to do casting we do it in the view or the vispy code can handle it.

@perlman I also agree moving towards "arbitrary set of values are None" is a good idea too, is that being enabled / make easier with #713 (feel free to respond over there / make a separate issue for this).

@jni I've added two simple round trip tests like what you describe but using `blobs`. I think this is ready for review now

napari/components/tests/test_viewer_labels_io.py
```diff
@@ -0,0 +1,28 @@
...
+    blobs = binary_blobs(length=128, volume_fraction=0.1, n_dim=2)
+    labeled = ndi.label(blobs)[0].astype(np.uint8)
+    with NamedTemporaryFile(suffix='.png') as fout:
+        imwrite(fout, labeled, format='png')
+        viewer.add_labels(path=fout.name)
+        assert len(viewer.layers) == 1
+        assert np.all(labeled == viewer.layers[0].data)
+
+
+def test_tiff_labels():
```
jni: Want to use pytest.mark.parametrize here on the suffix? =)

napari/components/viewer_model.py
```diff
@@ -719,12 +720,22 @@ def add_labels(
...
+        path : str or list of str
+            Path or list of paths to image data. Paths can be passed as strings
+            or `pathlib.Path` instances.
 
         Returns
         -------
         layer : :class:`napari.layers.Labels`
             The newly-created labels layer.
         """
+        if data is None and path is None:
```
jni, line 732: How do we feel about automatically creating a labels layer of the same shape and type (np, dask, zarr) as the currently selected layer in the viewer? Much like add_points. dask probably requires a lot of engineering, but zarr probably works out of the box and would allow labeling big image volumes.

jni: (Though I’m happy to leave this for a future pr)

jni (approved):
@sofroniewn nice and simple PR! Love it. I’ve made a couple of optional comments. =)

napari/components/viewer_model.py
sofroniewn: yes, we should do this (i know you requested in #470, but let's make it a new PR)

napari/components/tests/test_viewer_labels_io.py
sofroniewn: done :-)

sofroniewn:
I have now parameterized the test with `suffix` and will merge after the tests pass. I will follow-up on #470 in its own PR I can do that asap

sofroniewn:
The windows tests are failing because of how I use `NamedTemporaryFile` and permissions - see this warning here https://docs.python.org/3.6/library/tempfile.html#tempfile.NamedTemporaryFile

@jni @kne42 do either or you have any suggestions for how to fix this? stack overflow wasn't so helpful

[edit] maybe I try `delete=False` as suggested here https://stackoverflow.com/questions/23212435/permission-denied-to-write-to-my-temporary-file?

jni:
@sofroniewn I think this issue is the reason why we created this in skimage:

https://github.com/scikit-image/scikit-image/blob/master/skimage/_shared/_tempfile.py

perhaps just use that? Even though it has an underscore so is theoretically private... It's been around for a long time and anyway, it's easy enough to roll our own if skimage's goes away.

sofroniewn:
I've added `temporary_file.py` - felt like something that better belonged in the repo, if you prefer it not there I can remove and just import the skimage one. We'll see if the tests now pass.

If you approve and the tests all pass can you then merge

napari/util/temporary_file.py
```diff
@@ -0,0 +1,26 @@
+from tempfile import NamedTemporaryFile
+from contextlib import contextmanager
+import os
+
+
+@contextmanager
+def temporary_file(suffix=''):
+    """Yield a writeable temporary filename that is deleted on context exit.
```
jni:
```suggestion
    """Yield a writeable temporary filename that is deleted on context exit.

```

napari/util/__init__.py
```diff
@@ -1 +1,2 @@
 from .misc import is_rgb, segment_normal
+from .temporary_file import temporary_file
```
jni: I'm not positive that we want this in our public API...? As opposed to having it be e.g. a testing utility.

jni (approved):
@sofroniewn I've made two very minor comments, but otherwise LGTM!

They are minor and could be fixed in follow-up PRs if you want to get on with life. =)

napari/util/temporary_file.py
kne42: also the correct spelling is "writable" ;)

napari/util/__init__.py
sofroniewn: I'm fine changing it, but what do you have in mind. Leaving file where it is, but dropping this import, or maybe putting it inside of `tests`. We might want to end up using it in multiple places

sofroniewn:
I have dropped the public import and fixed the typo. If this looks good @jni you can merge after tests pass

napari/util/temporary_file.py
sofroniewn: done

PR #749, opened 2019-12-01
simplify keybindings info and make dockable

sofroniewn:
# Description
This PR dramatically simplifies the keybindings info widget (removes two classes), improves its stylings (using QTextEdit instead of our own equivalent) and makes the widget dockable which is nice as you might want to have the keybindings info open during other usage. It also adds the ability to add custom shortcuts to show / hide our dock widgets. Curious what @tlambert03 thinks.

Here is a gif:

dockable_keybinding_info

## Type of change


sofroniewn:
Maybe I should change the label "Layer type:" as "Currently Active" is a bit weird there ...

tlambert03:
really like this!  Was just playing around a bit with some styles.  What would you think of simplifying it even more by getting rid of the function signature (which is repetitive and not _that_ useful in this context) and presenting it with platform-specific key icons, and aligned like this:

Screen Shot 2019-12-01 at 8 31 35 PM

I can push that code if you're into it.

sofroniewn:
Yeah, wow @tlambert03, love that! Please push!!

What are your thoughts on “currently active” and then all the layer types, does this make sense? Could it be better?

tlambert03:
> What are your thoughts on “currently active” and then all the layer types, does this make sense? Could it be better?

I was a little confused at first that it meant "currently active **layer**" ... which it sort of does, but maybe if it said "all active keybindings" it would be clearer?  I like having the dedicated layer-types tabs in case that's all someone wants to see.

sofroniewn:
Maybe “all active keybindings” and then “image layer keybindings” etc. maybe we also just drop extra QLabel and put this all in the QComboBox string? Can you make this or similar change and update the PR?

tlambert03:
updated.  We should discuss how the unicode characters for non-OSX systems should look.  We could use everything except keep `Alt` and `Ctrl` for windows, not sure about linux...

jni:
Great work here!

One thing I want to suggest: I really don't like having this be a regular pop-up or a docked window. I really like the overlay that e.g. gmail provides. Have you tried it? Here's the workflow:

- push '?' (shift-/) to show help as an overlay
- scroll etc to see all the shortcuts
- if you press any of the shortcuts, the overlay disappears and the shortcut is activated. It's super handy for learning because it's so fast. There is no fudging around with the mouse to get rid of the help. And to do the next command you repeat gladly because it's so unobtrusive — and often much faster than digging around some menus to find the right command.

tlambert03:
> if you press any of the shortcuts, the overlay disappears and the shortcut is activated. It's super handy for learning because it's so fast.

gmail is actually not doing that for me! :) .  I can only make it disappear with `Escape`.  I do like the look of it, and at least on mac, I would say it's becoming the expected style for a keybinding guide.  To accomplish, I think we'd want to turn the whole thing into a `QDialog`, with `setModal(True)`.  Getting it to disappear with any of the specific shortcuts might be tricker.

jni:
> gmail is actually not doing that for me! :)

Oh man! They've changed it! Well, it was the correct behaviour, and Google went and ruined it, as they love to do. =P

Yes, I acknowledge 100% that the implementation might be pretty difficult/impossible! 😬 Just pointing out a target to aspire to.

sofroniewn:
@jni my gmail implementation is like @tlambert03. I do like it though. We can revert the making of this dockable and make it a `QDialog` again. @tlambert03 I'm happy to do that in this PR, but what do you think?

I can also try the "show them all at once just scroll" approach if people want, but worry we might have a lot!

We could also do the overlay appear / disappear with press hold.

I think the click to execute can wait for followup PRs as that will be somewhat involved, though not too bad, you'd have to make every line it's own pushbutton with the text be the button and then style them to look nice

Finally this PR and #750 also changes the keyboard shortcuts to make the info box appear - now command+? for the system info box, and command+alt+? for the keybindings. I can change this if @jni you want, but most show / hide of our windows now had `alt` in it, and i thought I'd use the simpler keybinding for the system info box

tlambert03:
I’m a bit conflicted. On the one hand, I like the key binding modal that fills the screen, has some opacity, and feels like all the rest of them, but I also kinda like the ability to have it off to the side so I can refer to it _as_ I use the program, which the modal dialog would prevent.  The counter argument, I guess, is that if you make that dialog easy enough to get to, and easy enough to close, then bringing it up repeatedly is no more difficult.

napari/_qt/qt_about_keybindings.py
```diff
@@ -1,105 +1,80 @@
...
+        self.setLayout(self.layout)
 
-        scroll_widget = QWidget()
-        scroll_widget.setSizePolicy(
-            QSizePolicy.Expanding, QSizePolicy.Expanding
+        # stacked keybindings widgets
+        self.textEditBox = QTextEdit()
+        self.textEditBox.setTextInteractionFlags(Qt.TextSelectableByMouse)
+        self.textEditBox.setMinimumWidth(360)
+        self.keybindings_strs = {'All active keybindings': ''}
```
kne42: prior to Python 3.7, dictionaries being ordered by default is merely an implementation detail

please use an `OrderedDict` instead and add a comment marking that this can be changed to an ordinary dict when our minimum Python version is 3.7

napari/_qt/qt_about_keybindings.py
```diff
@@ -1,105 +1,80 @@
...
+        for layer in layers:
+            if len(layer.class_keymap) == 0:
+                text = 'No keybindings'
+            else:
+                text = get_keybindings_summary(layer.class_keymap)
+            self.keybindings_strs[f"{layer.__name__} layer"] = text
+
+        # layer type selection
+        self.layerTypeComboBox = QComboBox()
+        for name in self.keybindings_strs.keys():
```
kne42: the `.keys()` part is implicit when iterating over a dict

napari/_qt/qt_about_keybindings.py
```diff
@@ -1,105 +1,80 @@
...
+    def change_layer_type(self, text):
+        self.textEditBox.setHtml(self.keybindings_strs[text])
+
+    def update_active_layer(self, event):
+        text = ''
         # Add class and instance viewer keybindings
-        keybindings_str += get_keybindings_summary(self.viewer.class_keymap)
-        keybindings_str += get_keybindings_summary(self.viewer.keymap)
+        text += get_keybindings_summary(self.viewer.class_keymap)
+        text += get_keybindings_summary(self.viewer.keymap)
```
kne42: what happens if a class keymap and a layer keymap share the same key combination?

napari/_qt/qt_about_keybindings.py
```diff
@@ -1,105 +1,80 @@
...
-        layer_label = QLabel(keybindings_str)
-        layer_label.setTextInteractionFlags(Qt.TextSelectableByMouse)
-        # layer_label.setAlignment(Qt.AlignLeft)
-        layer_label.setSizePolicy(QSizePolicy.Expanding, QSizePolicy.Expanding)
-        self.setSizePolicy(QSizePolicy.Expanding, QSizePolicy.Expanding)
-        self.layout.addWidget(layer_label)
-        self.setLayout(self.layout)
+        # Do updates
+        self.keybindings_strs['All active keybindings'] = text
+        if self.layerTypeComboBox.currentText() == 'All active keybindings':
```
kne42: what's this check for?

napari/_qt/qt_about_keybindings.py
```diff
@@ -1,105 +1,80 @@
...
-
-        layer_label = QLabel(keybindings_str)
-        layer_label.setTextInteractionFlags(Qt.TextSelectableByMouse)
-        # layer_label.setAlignment(Qt.AlignLeft)
-        layer_label.setSizePolicy(QSizePolicy.Expanding, QSizePolicy.Expanding)
-        self.setSizePolicy(QSizePolicy.Expanding, QSizePolicy.Expanding)
-        self.layout.addWidget(layer_label)
-        self.setLayout(self.layout)
+        # Do updates
+        self.keybindings_strs['All active keybindings'] = text
```
kne42: with the amount that you use `'All active keybindings'`, I think a constant would be appropriate

napari/util/misc.py
```diff
@@ -692,6 +693,23 @@ def hello_world(layer, event):
...
+            'Shift': '⇧',
+            'Alt': '⌥',
+            'Option': '⌥',
+            'Left': '←',
+            'Right': '→',
+            'Up': '↑',
+            'Down': '↓',
+            'Backspace': '⌫',
+        }
+    )
```
kne42: beautiful! 😍

napari/util/misc.py
```diff
@@ -705,19 +723,28 @@ def get_keybindings_summary(keymap):
...
+    keybindings_str += '</table>'
     return keybindings_str
 
 
 def get_function_summary(func):
     """Get summary of doc string of function."""
     doc = FunctionDoc(func)
     summary = ''
-    summary += doc['Signature']
+    # summary += doc['Signature']
```
kne42: why the comment?

sofroniewn:
I agree - I'm also conflicted. @jni we could just merge this in now and then revisit later after we see how this feels. I think there are more important things for us to focus on now (like remote viewer, more file IO, plugins etc) and we've done a lot of work on the stylings the past couple weeks that I want to finish up.

kne42 (approved):
looks great nick!

i've left some comments on nitpicky stuff but overall, it looks ready to merge :)

tlambert03 (approved):
merging this as is would be my preference.  I'll approve and let @jni make the final call. 
@sofroniewn, you might want to add a couple windows or linux specific key codes to `KEY_SYMBOLS` before merging.  At the very least maybe make "Control" -> "Ctrl" ... You could put everything but the overtly mac-specific stuff in the main dict.

napari/util/misc.py
tlambert03: Hi @kne42!  Welcome back, I've heard so much about you and read so many of your commits :) Nice to meet you.  The comment was to remove the signature from the function summary as mentioned briefly [above](https://github.com/napari/napari/pull/749#issuecomment-560189455).  I know it's a bit of a change, but I felt like for the public facing window like this the signature was unnecessary (almost all of the signatures are duplicates of the docstring).  We could just delete the line fully, but i commented so as not to forget how it worked if we wanted to change our minds.

kne42: Nice to meet you too! I'm so happy that you're passionate about this project too, and am excited to see you join the core development team :)

I'm not a big fan of commenting instead of deleting, but if you add another comment explaining why it's commented out, it should be fine :)

napari/_qt/qt_about_keybindings.py
sofroniewn: done

napari/_qt/qt_about_keybindings.py
sofroniewn: done

napari/_qt/qt_about_keybindings.py
sofroniewn: hmm, right now both will appear (which is same as before this PR). What happens in the code again? We should make sure that the same thing prints as happens, but maybe we do that in a followup PR

napari/_qt/qt_about_keybindings.py
sofroniewn: If `All active keybindings` are displayed then when changing the active layer we need to update the actual display too. I've added a more informative comment

napari/_qt/qt_about_keybindings.py
sofroniewn: good idea, done

royerloic:
Beautifull.

napari/util/misc.py
jni: @kne42 :wave::wave::wave::wave::wave: so good to see you back. :grin: 

@tlambert03 I'm with @kne42, I would say remove it, but it's true that uncommenting a line is less effort than researching how to do this in the first place. An alternative would be:

 ```suggestion
    # if we want to add the function signature to the summary, use:
    # summary += doc['Signature']
```

Comment (sofroniewn) [reply napari/util/misc.py]: I'm ok with dropping, isn't too hard to get the signature back. I also really like how much cleaner it is without

**Comment by jni** [comment]
> The counter argument, I guess, is that if you make that dialog easy enough to get to, and easy enough to close, then bringing it up repeatedly is no more difficult.

:point_up: this thing! I'm telling you, guys, the Gmail team broke a great thing, and that way of interacting with shortcuts was *the best*, because when you are looking for keyboard shortcuts is precisely when you least want to go reach for a mouse to move the shortcuts frame from an awkward docked position to side by side. That I didn't notice them change the behaviour tells you how well it did its job of teaching me all the shortcuts. =)

How about the following: use a modal, that is trivially dismissed, *and* at the top of the modal, there is a keyboard shortcut for popping it out. This (a) removes the docked view which honestly I find *really* awkward for this use, while (b) not providing a pop-up, which will make @royerloic unhappy, and (c) allowing users to *opt in* on a pop-up, if they have enough screen real estate that they want to do this.

**Comment by sofroniewn** [comment]
I've made most of the updates requested by @kne42 - I think the comment about keybinding conflicts should be resolved in a separate PR.

After playing with this for a bit though I agree with @jni that the dock is a bit awkward, especially now that @tlambert03 uses the symbols instead of the function signatures it's possible to fit much more on a single screen.

I'll give this a go right now.

**Comment by sofroniewn** [comment]
@jni @kne42 I need your help with 

> At the very least maybe make "Control" -> "Ctrl" ... You could put everything but the overtly mac-specific stuff in the main dict

I never remember whats command, whats control etc

**Comment by jni** [comment]
Oh, another thing: dismissing with Esc is also encouraged *if* that is displayed prominently — at the top, next to the "pop-out" shortcut.

Also, regarding all the shortcuts: my suggestion here is to only show the currently active shortcuts in the modal, and add the drop down, optionally in the modal, possibly only in the pop-out. I don't feel strongly about this, but I do feel strongly that the default behaviour should show only currently-active shortcuts. The exception to this is if there are shortcuts that could *never* be simultaneously active with the help. For example, in Gmail, they show the shortcuts that are active during editing, since you couldn't access the help while editing. (Although I guess that's not strictly necessary and I don't think we will have this issue...)

**Comment by jni** [comment]
> I never remember whats command, whats control etc

Sorry, what do you mean? ie how can we help?

**Comment by sofroniewn** [comment]
I have just made it use a QDialog again - see here:

image

the only keypress to exit the dialog is `esc`. I couldn't figure out how to get the other shortcuts to be global - I thought 
```python
about_keybindings.setShortcutContext(Qt.ApplicationShortcut)
```
would do it, but it doesn't.

@jni can you take this PR over and get the QDialog behaving exactly as you want (it should be pretty close). Can you also handle any of the cross platform symbol dictionary compatibility stuff @tlambert03 mentioned

jni:
Yep, happy to take this over, although fyi I'll pick it up in a couple of days, as I am teaching a workshop tomorrow.

sofroniewn:
@jni maybe we should then just merge and revisit. @tlambert03 how bad is the keybindings shortcut display situation for Linux / windows? If it is ok or if you can quickly make it ok, maybe you can do this and then we merge. The QDialog we have now is fully functional and strictly better than what is currently there.

tlambert03:
I can make it ok quickly, but the question is when.  Just got to Germany this morning.  Might find time in the next 12 hours

tlambert03:
ok, updated keycodes

sofroniewn:
Thanks for updating those strings @tlambert03! Have a great time in Germany.

@jni I have now updated this PR so the dialog will toggle in / out with the `Command+Alt+/` keypress. I think it is quite nice, but you can definitely work on in a followup PR to make more like you imagine it (i.e. only appears on hold). Can you give this a final review and approve and then I’d like to merge.

sofroniewn:
I am merging. Can always make improvements in follow-up PRs

PR #655, opened 2019-11-04
add governance model, mission and values, core dev guide

sofroniewn:
# Description

This PR will add a governance model, mission and values statement, and core developer guide to `napari` as was discussed at the last team meeting. It is largely based off the equivalent docs from skimage [governance](https://scikit-image.org/docs/dev/skips/1-governance.html), [core dev guide](https://scikit-image.org/docs/dev/core_developer.html#core-dev), and [values](https://scikit-image.org/docs/dev/values.html#values). It will certainly need to adapted more to napari's specific case, but I wanted to get the ball rolling with this. I also want to make sure all current team members get a chance to weigh in before this merges.

I think we should work on the `VALUES.md` and make sure we are aligned on that.

Also we should make sure `CORE_DEV_GUIDE.md` accurately reflects what we expect, and if we are comfortable with the mapping between core devs and [members of the napari GitHub org](https://github.com/orgs/napari/people) or any of our [teams](https://github.com/orgs/napari/teams)

Right now I've cut the concept of a SKIP (scikit-image enhancement proposal), I don't think we are there yet with our community size, but curious for your thoughts @jni?

PR #750, opened 2019-12-01
clean up info box

sofroniewn:
# Description
This PR continues in the vein of #749 by cleaning up some of the info dialog box and adding citation information to it. It also changes our citation from `team` -> `contributors` which was proposed in #655 and liked by @jni - will be nice to get this in sooner though.

Here is the new look - not this dialog is not dockable as i think that doesn't make sense in this context:

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ Contributions are encouraged! Please read our [contributing guide](./docs/CONTRI
 
 If you find `napari` useful please cite this repository using its DOI as follows:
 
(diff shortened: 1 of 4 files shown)


AhmetCanSolak (approved):
Looks good on mac osx but i would be happy if someone can check on windows and linux before merging too. Thanks @sofroniewn !

jni:
Love it! The code is much nicer too.

PR #756, opened 2019-12-03
[WIP] Add repr_png method to Viewer

jni:
I was getting kind of annoyed that notebooks would only display
"napari.Viewer object at <0xGIBBERISH>" at the end of cells
calling napari. This instead makes a png and displays it.

The disadvantage is that users might be confused and think that
this is the interactive napari viewer. But, on the whole, I think
it's a pretty cool way to save notebook state!

napari-repr

### DIFF
--- a/napari/viewer.py
+++ b/napari/viewer.py
@@ -1,7 +1,9 @@
+from io import BytesIO
 from os.path import dirname, join
 
 from qtpy.QtGui import QIcon
 from qtpy.QtWidgets import QApplication
+from skimage import io
 
 from napari._qt.qt_update_ui import QtUpdateUI
 
@@ -82,6 +84,14 @@ class Viewer(ViewerModel):
         else:
             image = self.window.qt_viewer.screenshot()
         return image
+
+    def _repr_png_(self):
+        """PNG representation of the viewer object for IPython."""
+        image = self.screenshot(with_viewer=True)
+        file_obj = BytesIO()
+        io.imsave(file_obj, image, format='png', check_contrast=False)
+        file_obj.seek(0)
+        return file_obj.read()
 
     def update(self, func, *args, **kwargs):
         t = QtUpdateUI(func, *args, **kwargs)


jni:
Some alternatives:

- set the dims to the middle of each axis before screenshotting, then set them back. But that might cause some flashing in the actual viewer that would be undesirable
- use to_svg() instead and lose the window chrome; but the advantage is that all layers could now have reprs. Indeed, this might be a good idea independently of what we do with the viewer.
- discard the whole thing. 
- use ipywidgets+bokeh to live update the screenshot on mouse release as demonstrated by Matt Rocklin [here](https://matthewrocklin.com/blog/work/2017/10/16/streaming-dataframes-1). I quite like the sound of this but didn't have time to take this that far yet.

sofroniewn:
I'm torn. I just tried it out, it's pretty cool, and it's easily suppressed with `;` if people don't want this. If we go down this path having something for each `Layer` makes sense too, can think about what that is, screenshot / svg - or just informative text.

I'm not sure about

> use ipywidgets+bokeh to live update the screenshot on mouse release

I think that is too heavy and might encourage people to do strange things.
I'm also worries about 

> disadvantage is that users might be confused and think that
this is the interactive napari viewer

maybe we can add some extra message / something to make this clear. I guess now the main window is raised it's not so bad, but when that wasn't happening it could have been very confusing.

I think what ultimately tips me in favor of this is that people are starting to share cool notebooks of napari visualization, and unfortunately until you run that notebook you have no idea what they were actually looking at. Having this functionality will allow that, which I see is a pretty major usecase

AhmetCanSolak:
I would be okay if we do something less than whole screenshot. I think entire window screenshot is a really bad idea. So confusing. I like @sofroniewn 's comments. It will be much better to have something `Layer` specific and for viewer we can think of something else. We can even just put napari logo for viewer. I do not mind.

sofroniewn:
There is possible a middle ground where the viewer screenshot takes up say the right half of the output space and then we have some text / other visuals about say the layers and their shapes that takes up the left half - more like this:
image

kevinyamauchi:
@sofroniewn, I really like that!

jni:
> what ultimately tips me in favor of this is that people are starting to share cool notebooks of napari visualization, and unfortunately until you run that notebook you have no idea what they were actually looking at. Having this functionality will allow that, which I see is a pretty major usecase

Yeah, that's exactly right, and a big part of the motivation for this.

> I think that is too heavy and might encourage people to do strange things.

Such as? I actually kinda like that the notebook would then render the *last* thing the user looked at.

> There is possible a middle ground

Yep. So how about:

- layers info on the left,
- live-ish screenshot on the right

My main reason for wanting the screenshot to be live is that in many datasets, like the one shown above, the bottom layer is totally uninteresting, and that means that we will see tons of notebooks with super-boring napari screenshots.

sofroniewn:
I'm not sure about the "live screenshot", it depends exactly what you mean and I might not quite understand. Depending on how it works the notebook sharing case could be really weird. I see these as more static snapshots of the viewer state the moment the cell was run, capturing the state at the end of the cell. It's kinda of like printing the viewer and if you did it once near the top of your notebook and there were only 3 layers the screenshot should have 3 layers if you do it again at the bottom of the notebook and the viewer now has 5 layers you should now have 5 layers and it should look different than the 3 layer example from above. The notebook can then tell the story of what you were doing, whereas if the updates were live I guess that you'd just see 5 layers everywhere, but maybe I am not understanding.

I'm also tempted that we should only have one of these and when you do `Viewer()`, `napari.view_*`, or `viewer.add_*` you'd see the same thing. The last case is maybe a bit different because `viewer.add_*` returns a layer, but I'm not sure do layers get there own one of these? Seems like we're adding a lot of complexity to require each layer to define their own special thing here. I'd be curious to know the plans for `layer` before going too far for the plans for `viewer`

sofroniewn:
While this is not identical to #1269, I think this has been superseded by #1269 and this can be closed. @jni feel free to re-open or post if you have further ideas for this/ think otherwise.

Issue #640, opened 2019-10-28
layer.scale and Image [indices]


tlambert03:
##  🐛? 
not sure if this is a feature request or a bug.  but I was confused by the behavior of the sliders when `layer.scale != [1, 1, 1]` ... I haven't had time to look at the appropriate code yet, but here was the behavior I was "expecting" to happen from, I guess, previous experience:

- when in "2D" mode (with a 3D image), the dims slider should have as many "stops" as there are planes in the image.  That is: I shouldn't be able to move the slider without updating either the Image index or the visual. And regardless of any voxel anisotropy, the min/max bounds of the `Image [index]` status at the bottom left should reflect the image dimensions as sampled/captured.
- when in "3D" mode, I expect it to work like it currently does, wherein the image is resampled according to the scale factor I have set to maintain the proper "realspace" proportions.

however, these two things seem to be at odds.  If I set `layer.scale[0] != 1`, then the 3D image looks correct, but the max slider range becomes `nz * layer.scale[0]`.  Relatedly, the first time I set the parameter, I used my physical pixel sizes (something like `[0.25, 0.1, 0.1]`) and was similarly surprised by what it did to the apparent image shape.  Is this just something that still needs to be ironed out?  or was there a "long-view" vision for `layer.scale` that I'm not yet appreciating?

sofroniewn:
I'd say this is something that needs more thought, right now we have the concept of a "step size" in our range, which will influence the indices that we slice at - see here https://github.com/napari/napari/blob/f27b90dab2032c2540d50070f18afdbf060a4b11/napari/components/dims.py#L185

part of the motivation for this was to handle image pyramids where an axis has been downsampled but you still want the slider to have the same total range as the base of the pyramid, the code that was added at that time might be causing any counter intuitive behavior that you are noting.

I'm quite open to improving this, but we need to make sure the right thing still happens in the pyramid layer, the normal 2D case, and the 3D case as you note.

In general we also havn't thought too much about physical coordinates, units, and how they might get incorporated, so now might be a good time to think about some of that too

sofroniewn:
So testing this out more now. I can see how the current implementation might be undesirable, especially for the case where scale < 1 and you are looking in 2D, as your slider now will not give you the ability to look at every slice - on the other hand if you are applying a scale I think it makes sense that the indices in the bottom left status bar contain the values after the scale.

[...]

PR #759, opened 2019-12-04
Display slice info on right of slider

### LINKED ISSUE #456: Display total number of slices on right end of corresponding slider
## 🚀 Feature

From my colleague @VolkerH who was trying out napari on his (oddly-shaped due to IO issues) data: it would be nice if the max value of each slider was shown e.g. at the right end of that slider, so that you could at a glance tell the shape of your data.

tlambert03:
# Description
This would close #456 by displaying `current / total` slice info on the right of each dims slider

Untitled

still want to tighten up some styles... but just wanted to make sure this was what everyone was generally picturing.

## Type of change

# How has this been tested?

- [x] need to add tests

## Final checklist:

### DIFF
--- a/napari/_qt/qt_dims_slider.py
+++ b/napari/_qt/qt_dims_slider.py
@@ -3,7 +3,6 @@ from typing import Optional, Tuple
 import numpy as np
 from qtpy.QtCore import Qt, QTimer, Signal, Slot, QObject
 from qtpy.QtWidgets import QApplication
-from qtpy.QtGui import QFont, QFontMetrics
 from qtpy.QtWidgets import (
     QComboBox,
     QHBoxLayout,
@@ -46,7 +45,7 @@ class QtDimSliderWidget(QWidget):
         self.slider = None
         self.play_button = None
         self.slice_label = QLabel(self)
-        self.slice_label.setObjectName('sliceLabel')
+        self.slice_label.setObjectName('slice_label')
 
         self._fps = 10
         self._minframe = None
@@ -169,9 +168,6 @@ class QtDimSliderWidget(QWidget):
                 self.slider.setMaximum(_range[1])
                 self.slider.setSingleStep(_range[2])
                 self.slider.setPageStep(_range[2])
-                fm = QFontMetrics(QFont("", 0))
-                longest = str(_range[1] // _range[2]) * 2 + '   /   '
-                self.slice_label.setFixedWidth(fm.width(longest))
                 self._update_slice_labels()
         else:
             displayed_sliders[self.axis] = False
--- a/napari/_qt/qt_dims_slider.py
+++ b/napari/_qt/qt_dims_slider.py
@@ -3,6 +3,7 @@ from typing import Optional, Tuple
 import numpy as np
 from qtpy.QtCore import Qt, QTimer, Signal, Slot, QObject
 from qtpy.QtWidgets import QApplication
+from qtpy.QtGui import QFont, QFontMetrics
 from qtpy.QtWidgets import (
     QComboBox,
     QHBoxLayout,
@@ -44,10 +45,8 @@ class QtDimSliderWidget(QWidget):
         self.axis_label = None
         self.slider = None
         self.play_button = None
-        self.curslice_label = QLabel(self)
-        self.totslice_label = QLabel(self)
-        self.curslice_label.setObjectName('curSliceLabel')
-        self.totslice_label.setObjectName('totSliceLabel')
+        self.slice_label = QLabel(self)
+        self.slice_label.setObjectName('sliceLabel')
 
         self._fps = 10
         self._minframe = None
@@ -62,8 +61,7 @@ class QtDimSliderWidget(QWidget):
         layout.addWidget(self.axis_label)
         layout.addWidget(self.play_button)
         layout.addWidget(self.slider, stretch=1)
-        layout.addWidget(self.curslice_label)
-        layout.addWidget(self.totslice_label)
+        layout.addWidget(self.slice_label)
         layout.setContentsMargins(0, 0, 0, 0)
         layout.setSpacing(2)
         self.setLayout(layout)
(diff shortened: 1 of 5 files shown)


tlambert03:
one question, regarding the width of the text: do you guys want to maximize the width of the sliders (i.e. minimize width of labels) at all times, such as this:

Screen Shot 2019-12-04 at 1 59 17 PM

or set the width of all labels to be equal to the largest width so that the sliders are always right aligned, but allow for there to be some unused space (similar to what I did for the axis labels)

sofroniewn:
@tlambert03 looking good - I'd like the sliders to be always right aligned, so not like that last photo where one is longer than the others. I think that looks strange

tlambert03:
ok, looking like this now:

Screen Shot 2019-12-04 at 4 04 18 PM

(If you're thinking about the right vs. left alignment of the labels, I chose right here to avoid the denominator "jumping" around as the slice moves from single to multiple digits...).  
will add some tests

kevinyamauchi:
Looks awesome! I think the slice labels also help make it clear that the labels on the left are the axis labels.

I agree with @sofroniewn. I think all of the sliders should be aligned and the same length. I also prefer the text alignment in the first photo (i.e., digits aligned, “/“ aligned). 

Minor nitpick: the amount of space to the left and right of the “/“ looks different in the first photo and I’d prefer it to be the same.

tlambert03:
> I agree with @sofroniewn. I think all of the sliders should be aligned and the same length. I also prefer the text alignment in the first photo (i.e., digits aligned, “/“ aligned).
> Minor nitpick: the amount of space to the left and right of the “/“ looks different in the first photo and I’d prefer it to be the same.

you've hit on a challenge!  That first photo used two separate `Qlabels`, which makes aligning the "/" easy, regardless of the width of numerators/denominators.  The last photo uses a single `Qlabel`, which makes it easy to keep the space to the left and right of the "/" constant (very fine spacing details between individual `QWidgets` in a `QLayout` is a bit challenging)...  but it's hard to achieve both at the same time.  I could make 3 labels (one dedicated to the separator) but then it's hard to keep the extra dead space to a minimum...

tlambert03:
here's another option:
Screen Shot 2019-12-04 at 4 34 03 PM

kne42:
I like the last option the best, with all of the `/` aligned. Would also be nice in maybe a follow-up PR (or this one if it's trivial enough) to have the "current slice" be editable.

tlambert03:
current look & feel: 

Untitled

let me know if you approve and i'll add tests.

kne42:
Looks beautiful! My only quip is to maybe make the separator a bit lighter so that it will stand out more :)

napari/_qt/qt_dims_slider.py
```diff
@@ -40,9 +43,32 @@ def __init__(self, parent: QWidget, axis: int):
         self.axis = axis
         self.qt_dims = parent
         self.dims = parent.dims
-        self.label = None
+        self.axis_label = None
         self.slider = None
         self.play_button = None
+        self.curslice_label = QLineEdit(self)
```
kevinyamauchi, line 49: What do you think about adding a tooltip to the slice labels? I think this helps make it clear what they are when people are new to `napari`. It might be nice to do the same for the axis labels as well.

kevinyamauchi: The context for this comment is that while I like the spacing with the vertical separator between the current and total slice index elements, I think the vertical line is less intuitive than the slash.

tlambert03:
this is more than I want to address in this PR... but just wanted to note that the internal representation of data-space, slider-space, indexing and `scale` (i.e. #640) is popping up here again.  For instance, when `scale` is an integer value here (try `viewer.layers[0].scale = [1.2, 1, 1]`), using the `(min, max, step)` values from `dims.range` sometimes rounds the slice labels to the wrong integer value.   Or, for some `scales`, you cannot reach the actual max index:

Screen Shot 2019-12-05 at 10 22 49 AM


I find myself wishing there was a more direct API  to access the data-index represented by the current slider position.

tlambert03:
This is ready for review.  All comments have been addressed (more visible separator, tooltips), tests are added, styles and code cleaned up a bit more.

napari/resources/stylesheet.qss
```diff
@@ -1152,3 +1152,22 @@ QtPlayButton[playing=True] {
...
+  min-height: 12px;
+  border: 0px;
+  padding-left: 1px;
+}
+
+
+#slice_label_sep{
+  background-color: {{ background }};
+  border: 1px solid {{ primary }};
+}
```
sofroniewn: line missing

sofroniewn:
@tlambert03 did you make the numbers brighter? I think @kne42 asked for the separator brighter. I actually have a hard time seeing, but it looks the wrong way around to me. See here if I reverse the colors: 
image
what do people prefer?

Otherwise looks good and ready to go. Also agree on not doing #640 now, but that is maybe next to the priority list and we should start working on soon. It will be nice to have this functionality to make sure we are doing the right things

tlambert03:
i made the separator brighter... I like keeping it all a little bit muted, but don't have a strong preference.  Feel free to play with colors from the palette and tell me if you have preferences for any of the elements.  currently (and in the post above) the numbers are `{{ secondary }}` and the separator is `{{ primary }}`

> Screen Shot 2019-12-05 at 10 22 49 AM
>

napari/resources/stylesheet.qss
tlambert03: 🤬

sofroniewn (approved):
This looks great to me! Stylings look nice, we can always tweak when we do a more global revisit, probably not a bad idea after we've finished this recent wave of changes. I'll let this sit for a little and then merge unless any further requests for changes are made

kevinyamauchi:
Awesome work, @tlambert03! I really like it! Thanks for patiently considering all of the styling nitpicks :p

Also, I totally agree there should be a way to directly access the current data-index from the dims model.

sofroniewn:
I'm going to merge. @jni I think the styling is pretty good now, and as noted above if we want to make more tweaks probably best to bundle them across the GUI and make a follow-up styling only PR. This functionality should also help as we go through #763 to make sure we're showing what we want to be etc

PR #765, opened 2019-12-06
Napari "scale" keyword argument may accept array-like (numpy, dsak, xarray) input

### LINKED ISSUE #751: Passing a numpy array as the "scale" argument crashes napari
## 🐛 Bug

This one is probably my fault for promoting the `x = y or default_value` syntax for None arguments: it crashes when `y` is a NumPy array. :grimacing: Now, when you pass a NumPy array as the `scale=` parameter, it crashes with `ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()`

## To Reproduce

```python
import numpy as np
from skimage import data, util
import napari


camera = util.img_as_float(data.camera())

with napari.gui_qt():
    napari.view_image(camera, scale=np.array([1, 1]))
```

```pytb
  File "test_array_spacing.py", line 9, in <module>
    napari.view_image(camera, scale=np.array([1, 1]))
  File "/home/jni/projects/napari/napari/view_layers.py", line 130, in view_image
    path=path,
  File "/home/jni/projects/napari/napari/components/viewer_model.py", line 512, in add_image
    visible=visible,
  File "/home/jni/projects/napari/napari/layers/image/image.py", line 163, in __init__
    visible=visible,
  File "/home/jni/projects/napari/napari/layers/base/base.py", line 133, in __init__
    self._scale = scale or [1] * ndim
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

## Expected behavior

It should work the same as if the input was a list.

GenevieveBuckley:
# Description

This pull request addresses the bug report in issue https://github.com/napari/napari/issues/751.

The problem is if you pass an array-like object as the `scale` keyword argument to the napari image viewer (instead of the expected list or None type) an error results. This PR checks whether `scale` is None, and if not converts it to list type. Napari now handles `scale` input types of None, list, and array-like (eg: numpy array, dask array, and xarray).

## Type of change

# References
Closes https://github.com/napari/napari/issues/751

# How has this been tested?

- [x] the test suite now has a new test covering cases with `scale` keyword argument types of:
* None
* list
* numpy array
* dask array
* xarray

## Final checklist:
- [ ] I have made corresponding changes to the documentation <-- I don't think this needs a documentation update, my understanding is inputs like this are expected to work anyway. If I'm wrong, let me know and I'll add extra detail to the docs.

### DIFF
--- a/napari/layers/base/base.py
+++ b/napari/layers/base/base.py
@@ -130,7 +130,10 @@ class Layer(KeymapMixin, ABC):
         self.scale_factor = 1
 
         self.dims = Dims(ndim)
-        self._scale = scale or [1] * ndim
+        if scale is None:
+            self._scale = [1] * ndim
+        else:
+            self._scale = list(scale)
         self._translate = translate or [0] * ndim
         self._scale_view = np.ones(ndim)
         self._translate_view = np.zeros(ndim)
--- a/napari/layers/base/base.py
+++ b/napari/layers/base/base.py
@@ -132,7 +132,7 @@ class Layer(KeymapMixin, ABC):
         self.dims = Dims(ndim)
         if scale is None:
             self._scale = [1] * ndim
-        else:
+        else:  # covers list and array-like inputs
             self._scale = list(scale)
         self._translate = translate or [0] * ndim
         self._scale_view = np.ones(ndim)
--- a/napari/layers/base/base.py
+++ b/napari/layers/base/base.py
@@ -134,7 +134,10 @@ class Layer(KeymapMixin, ABC):
             self._scale = [1] * ndim
         else:  # covers list and array-like inputs
             self._scale = list(scale)
-        self._translate = translate or [0] * ndim
+        if translate is None:
+            self._translate = [0] * ndim
+        else:  # covers list and array-like inputs
+            self._translate = list(translate)
         self._scale_view = np.ones(ndim)
         self._translate_view = np.zeros(ndim)
         self._translate_grid = np.zeros(ndim)
--- a/napari/layers/image/tests/test_image.py
+++ b/napari/layers/image/tests/test_image.py
@@ -1,5 +1,7 @@
 import numpy as np
 from xml.etree.ElementTree import Element
+import dask.array as da
+import xarray as xr
 
 import pytest
 from vispy.color import Colormap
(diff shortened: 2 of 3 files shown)


jni:
@GenevieveBuckley Looks great! Right now it looks like your dev environment is not set up correctly for napari. We use black for formatting and Flake8 for linting, and both need to pass before merging. Please see https://github.com/napari/napari/blob/master/docs/CONTRIBUTING.md#setting-up-a-development-installation, searching for "pre-commit". 

The OSX failures are an annoying spectre from the past... =( I don't think they are related to this PR.

Although, the other PR we have seen this in recently is #724, so maybe pytest.parametrize and pytest qtbot don't play well together! 😱

sofroniewn (commented):
For the failing osx tests you might need to add an explicit call to `viewer.window.close()` as is done here https://github.com/napari/napari/blob/c0556a6a161ee4b6faa85f93c37408512395614d/napari/tests/test_viewer.py#L86

But I think we might want to move this test inside our `Image` layer model tests in [test_image.py](https://github.com/napari/napari/blob/master/napari/layers/image/tests/test_image.py) and have the test work directly on `Image` as the failure was happening inside the model code and without any of the vispy / Qt stuff being called. In general if we can test things at the model level I'd prefer to as tests which pull in the vispy / Qt stuff have a lot more overhead.

Test itself looks great though. I love the use of `parameterize`. 

It also looks like `translate` keyword arg is affected by a similar problem https://github.com/napari/napari/blob/c0556a6a161ee4b6faa85f93c37408512395614d/napari/layers/base/base.py#L134

Can you add the same fix for `translate` and a similar test into this PR too.

GenevieveBuckley:
> Although, the other PR we have seen this in recently is #724, so maybe pytest.parametrize and pytest qtbot don't play well together! 😱

I've been using pytest-qtbot and pytest parametrize together quite a bit recently, and haven't noticed any weirdness there. I do usually set up my window with qtbot as a pytest fixture, and pass that into other parameterized tests. Maybe that makes a difference? Happy to talk more about it if you like.

GenevieveBuckley:
Ok, so...
- My development environment is now set up properly with pre-commit hooks for black & flake8
- I've moved the tests into `napari\image\tests\test_image.py` at Nick's suggestion
- I've added a similar fix & test for the `translate` keyword argument as well.

sofroniewn (approved):
Looks great, thanks for the changes. I will let sit till tomorrow in case anyone has any additional comments and then merge

jni (approved):
Love it!

Issue #766, opened 2019-12-06
Python 3.8 not currently supported with PySide2


sofroniewn:
## 🐛 Bug

As noted by @kne42 in https://github.com/napari/napari/issues/729#issuecomment-561836861 it seems like we currently do not have support for Python 3.8 and get the following error:

```
[I 21:03:54.961 NotebookApp] Adapting from protocol version 5.1 (kernel 74092921-9466-477e-b15a-3074468ac75b) to 5.3 (client).
[IPKernelApp] ERROR | Exception in message handler:
Traceback (most recent call last):
  File "/home/sebi06/anaconda3/envs/napari/lib/python3.8/site-packages/IPython/core/interactiveshell.py", line 3319, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-1-e580eceeb872>", line 2, in <module>
    import napari
  File "/home/sebi06/anaconda3/envs/napari/lib/python3.8/site-packages/napari/__init__.py", line 29, in <module>
    from .viewer import Viewer
  File "/home/sebi06/anaconda3/envs/napari/lib/python3.8/site-packages/napari/viewer.py", line 3, in <module>
    from napari._qt.qt_update_ui import QtUpdateUI
  File "/home/sebi06/anaconda3/envs/napari/lib/python3.8/site-packages/napari/_qt/__init__.py", line 1, in <module>
    from .qt_range_slider import QHRangeSlider, QVRangeSlider
  File "/home/sebi06/anaconda3/envs/napari/lib/python3.8/site-packages/napari/_qt/qt_range_slider.py", line 316, in <module>
    class QHRangeSlider(QRangeSlider):
TypeError: 'Shiboken.ObjectType' object is not iterable
```

what's interesting is this in one of our custom Qt elements. Maybe @AhmetCanSolak has an idea on how to fix and then we'd be `3.8` compatible

kne42:
See [this reply](https://bugreports.qt.io/browse/PYSIDE-1140?focusedCommentId=486074&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-486074) from a Qt developer that says that they plan to provide Python 3.8 compatibility in Qt 5.14.

sofroniewn:
Python 3.8 support has been added with #878

PR #767, opened 2019-12-06
fix set_fps type to float

kne42:
# Description
fixes a small discrepancy between the signature of `set_fps` and the `fps_changed` emitter by ensuring that both are using the ~~`int`~~`float` data type (one was `float` and one was `int` previously)

this fixes compatibility with PyQt5 and consequently, allows napari to be run on Python 3.8 if PyQt5 is used as the backend (see #766 as to why PySide2 cannot be used)

the reason that this was only erroring with PyQt5 is because it checks that the emitters and callbacks share compatible signatures while PySide2 does not

~~@tlambert03 please verify that `int` is the correct data type to use for this emitter since I was just guessing as to which one to use~~

## Type of change

# How has this been tested?

- [x] all tests pass with my change using both PySide2 on Python 3.7 and PyQt5 on Python 3.8

### DIFF
--- a/napari/_qt/qt_dims_slider.py
+++ b/napari/_qt/qt_dims_slider.py
@@ -32,7 +32,7 @@ class QtDimSliderWidget(QWidget):
     """
 
     axis_label_changed = Signal(int, str)  # axis, label
-    fps_changed = Signal(int)
+    fps_changed = Signal(float)
     mode_changed = Signal(str)
     range_changed = Signal(tuple)
     play_started = Signal()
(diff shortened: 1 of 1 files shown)


tlambert03:
no it should be float, since we allow non-integer fps settings.  Sorry about that (thought I had caught all of those when switching from int to float), the slot should be changed to float: 

https://github.com/napari/napari/blob/cf563b512ec7ede97654a15385ea58ba574f9d8a/napari/_qt/qt_dims_slider.py#L477-L482

tlambert03 (approved):
thanks for catching!

kne42:
@tlambert03 thanks for the quick response!

sofroniewn:
LTGM merging

PR #769, opened 2019-12-07
Adding shutdown method to QtViewer that closes all resources

shanaxel42:
# Description
Adds a shutdown method to the QtViewer class that closes all resources. Specifically not shutting down the console was causing the "Too many open files" kernal crash in #723. The new shutdown method should be used in any future tests that instantiate a QtViewer. Should be merged before #723 

## Type of change

# References

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/_qt/qt_viewer.py
+++ b/napari/_qt/qt_viewer.py
@@ -418,6 +418,11 @@ class QtViewer(QSplitter):
             self.pool.clear()
         event.accept()
 
+    def shutdown(self):
+        self.pool.clear()
+        self.canvas.close()
+        self.console.shutdown()
+
 
 def viewbox_key_event(event):
     """ViewBox key event handler
--- a/napari/_qt/tests/test_qt_viewer.py
+++ b/napari/_qt/tests/test_qt_viewer.py
@@ -20,6 +20,7 @@ def test_qt_viewer(qtbot):
     assert viewer.dims.ndim == 2
     assert view.dims.nsliders == viewer.dims.ndim
     assert np.sum(view.dims._displayed_sliders) == 0
+    view.shutdown()
 
 
 def test_add_image(qtbot):
@@ -39,6 +40,7 @@ def test_add_image(qtbot):
     assert viewer.dims.ndim == 2
     assert view.dims.nsliders == viewer.dims.ndim
     assert np.sum(view.dims._displayed_sliders) == 0
+    view.shutdown()
 
 
 def test_add_volume(qtbot):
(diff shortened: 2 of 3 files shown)


napari/_qt/qt_viewer.py
```diff
@@ -418,6 +418,11 @@ def closeEvent(self, event):
             self.pool.clear()
         event.accept()
 
+    def shutdown(self):
```
jni, line 421: LOL I love that there’s “clear”, “close”, and “shutdown” inside here. Lots to choose. =D 

FWIW I think you picked the right one.

I wonder if it’s necessary to put this in `__del__()` also? Or will the below methods be called automatically?

napari/_qt/tests/test_qt_viewer.py
```diff
@@ -20,6 +20,7 @@ def test_qt_viewer(qtbot):
     assert viewer.dims.ndim == 2
     assert view.dims.nsliders == viewer.dims.ndim
     assert np.sum(view.dims._displayed_sliders) == 0
+    view.shutdown()
```
jni, line 23: Hmmm, it’s mildly annoying to have these in every test. Is there a way to use the view fixture to have the teardown after the yield statement happen automatically?

jni: If we can make the syntax nice enough, we might even be able to start testing our examples again! 🎉

jni (approved):
@shanaxel42 🙌🙌🙌🙌 brilliant! This PR makes me so happy. As mentioned in the review comments, I have a feeling we could improve the syntax even more, but perhaps you are leaving that for the parametrised tests anyway. Thank you!!!

napari/_qt/tests/test_qt_viewer.py
shanaxel42: Oh yea these test actually all getting deleted in #723 so theres only one view.shutdown() in the parametrized test. I used the fixture yeild teardown model in `test_qt_play.py` so that should work for the examples!

napari/_qt/qt_viewer.py
shanaxel42: ooo good question, I don't think it would automatically happen. Should I add a __del__() override?

sofroniewn (approved):
This looks great to me, thanks @shanaxel42! I'll leave for @jni to have one more look and he can merge if good

sofroniewn:
This all looks great. If we need to add in `__del__` we can do so in a follow up PR. I will merge now

PR #773, opened 2019-12-07
drop ndim keyword from labels

sofroniewn:
# Description
This PR drops the `n_dimensional` keyword from the `Labels` layer. It will actually be basically backwards compatible because the `n_dimensional` keyword doesn't actually affect any of the processing or rendering, it just sets the value of a checkbox in the Qt which affects how new data gets added with the paintbrush and fill bucket. I don't this parameters like this should not be settable as keyword args and other similar parameters are not. I am fine merging this and still keeping the version change as a "minor release" when that time comes

## Type of change

### DIFF
--- a/napari/components/viewer_model.py
+++ b/napari/components/viewer_model.py
@@ -673,7 +673,6 @@ class ViewerModel(KeymapMixin):
         is_pyramid=None,
         num_colors=50,
         seed=0.5,
-        n_dimensional=False,
         name=None,
         metadata=None,
         scale=None,
@@ -702,8 +701,6 @@ class ViewerModel(KeymapMixin):
             Number of unique colors to use in colormap.
         seed : float
             Seed for colormap random generator.
-        n_dimensional : bool
-            If `True`, paint and fill edit labels across all dimensions.
         name : str
             Name of the layer.
         metadata : dict
@@ -741,7 +738,6 @@ class ViewerModel(KeymapMixin):
             is_pyramid=is_pyramid,
             num_colors=num_colors,
             seed=seed,
-            n_dimensional=n_dimensional,
             name=name,
             metadata=metadata,
             scale=scale,
--- a/napari/layers/labels/labels.py
+++ b/napari/layers/labels/labels.py
@@ -33,8 +33,6 @@ class Labels(Image):
         Number of unique colors to use in colormap.
     seed : float
         Seed for colormap random generator.
-    n_dimensional : bool
-        If `True`, paint and fill edit labels across all dimensions.
     name : str
         Name of the layer.
     metadata : dict
@@ -120,7 +118,6 @@ class Labels(Image):
         is_pyramid=None,
         num_colors=50,
         seed=0.5,
-        n_dimensional=False,
         name=None,
         metadata=None,
         scale=None,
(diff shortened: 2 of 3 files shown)


kevinyamauchi (commented):
This looks good to me and the rationale for dropping the `n_dimensional` keyword makes sense to me.

By "minor release", do you mean it would still be part of 0.2.x? If so, just to play devil's advocate, this is a breaking change if a user is explicitly setting the `n_dimensional` argument. I suppose on the other hand, we do have a disclaimer on the readme about the possibility of breaking changes. I don't have strong feelings either way because I am guessing the `n_dimensional` keyword argument isn't used too often.

sofroniewn:
I meant we could merge now and still release again under `0.2.x` without a deprecation warning,  but you're right they will get an error if they had used that keyword, though nothing else will be changed by dropping it.

We should have a dedicated discussion around our attitude to release names, release cycles, and deprecation warning policy pretty soon. We can either start that here or move the discussion somewhere else, depending on how @napari/core-devs feel.

I'll put down a few raw thoughts that are certainly open to changing with discussion, but maybe reflect how I'm thinking about it right now. I think changes like this - and even #686 (which I'd really like to get in soon) - are still fine for now, and it's probably most important to provide very clear instructions on how to fix deprecations in our PRs and clear links to those PRs with understandable titles in our release notes so that someone who encounters a breaking change could consult our release notes, come to the PR and immediately understand what they have to do to make things work again. Ultimately that will be required anyway and I think it is more important than the expectations of semantic versioning which I don't think too many of our more computational biology users care too much about at this point (and conventions and expectations can differ before `1.0.0` releases in different communities too).

We might also want to start a policy around deprecation warnings before we worry about super strict semantic versioning too. Certainly when napari is a more mature and depended on package this will be absolutely essential for the survival of the community and a plugin ecosystem, but I also worry that a significant slowdown now would have worse consequences than having some breaking changes in minor releases.

If we do go down a path of wanting a stricter policy around deprecations then i think we need to start maintaining branches like `dev0.3.0` so we can keep working on complex features that will break the api. See my comment here https://github.com/napari/napari/pull/655#discussion_r354531369

As I said above, those thoughts should just be taken as current thoughts, and I'm very open to other points of view and coming to a different conclusion too. I think it's important to get right.

kevinyamauchi:
Thanks for the clarification and for your thoughts, @sofroniewn! I think it could be worth having a discussion in a separate issue, because as you said, it's important to get this right.

I definitely agree with your point that clear documentation of changes is much more important than semantic versioning. I personally don't have a strong preference for strict semantic versioning at this point. I was mostly making the comment because I recall us deciding that API-breaking changes should be a 0.x bump, but I may be imagining that conversation :P

sofroniewn:
Based on discussion today it was decided that changes of this relatively small magnitude are still allowed for our `0.2.x` releases despite being breaking changes, but should be clearly marked as in our release notes, with instructions on how to fix in the PR. I've added a new `breaking change` label that we can use to tag PRs too.

This is now ready for re-review.

Note the breaking change is to drop `n_dimensional` as a keyword argument from `add_labels` etc. as it only affected interactivity. If you want to change the `n_dimensional` property you can still do so directly on the layer, say `layer.n_dimensional = True`.

jni (approved):

PR #779, opened 2019-12-09
change language around windows in readme

sofroniewn:
# Description
This PR changes the language around our windows support in the readme, which is actually pretty good. we have the occasional failure with old systems, but most newer systems work and this language is overly pessimistic. See this note https://github.com/napari/napari/issues/558#issuecomment-537587561 for details of windows systems we know we support


AhmetCanSolak (approved):
Thanks @sofroniewn !

PR #780, opened 2019-12-10
[WIP] creating movies in napari

guiwitz:
# Description
This pull request implements directly in napari the package [naparimovie](https://github.com/guiwitz/naparimovie) allowing to create movies based on a selection of key-frames. Currently the package supports capturing both 2D and 3D views (they can be mixed also in a movie), all sliders (time, z), visibility of layers, camera rotation/zoom/displacement. The functioning is the following. First create a view:
```python
import napari
view = napri.View()
view.add_image(image)
```
Then create a movie object that takes the view as a parameter:
```
movie = napari.Movie(myviewer=viewer)
```
From this moment a set of key-bindings are activated to capture key-frames (see the naparimovie repo for details). Then the movie can be saved as an mp4 (if ffmpeg is available) or as a gif:
```
movie.make_movie(name = 'movie.mp4', resolution = 300, fps = 20)
movie.make_gif('gifmovie.gif')
```
Finally the key-bindings can be released:
```
movie.finish_movie()
```

## Type of change

# References
[naparimovie](https://github.com/guiwitz/naparimovie)

# How has this been tested?
No tests

## Final checklist:


GenevieveBuckley:
I'm excited about this pull request, looking forward to trying out this feature!

@guiwitz, it seems like your development environment isn't quite set up right, so here's what Juan told me when I had the same problem earlier this week:
> We use black for formatting and Flake8 for linting, and both need to pass before merging. Please see https://github.com/napari/napari/blob/master/docs/CONTRIBUTING.md#setting-up-a-development-installation, searching for "pre-commit".

It also looks like this PR depends on `pyquaternion`, which isn't (yet?) an explicit dependency of napari, which is why the CI builds are failing right now.

jni:
Thanks @guiwitz! Very excited to see this happening! I don't have time to do a full review right this second, but

> It also looks like this PR depends on pyquaternion, which isn't (yet?) an explicit dependency of napari

Not to mention matplotlib!

I've also been rolling the name in my head for a few days and have come to like "Animation" or "Animate" over "Movie". What do people think?

I'll work on a full review in the coming days. Thanks again @guiwitz!

guiwitz:
Thanks for the hints @GenevieveBuckley, all tests are passing now! This allowed be to clean many more other things... 

@jni I'm not sure what you mean with the matplotlib comment. Is it a problem that matplotlib is a dependency? I always assume that it's such a basic part of Python, like numpy, that this dependency shouldn't be a problem. Also, I have no strong feelings about the name, but I agree that animate sounds more general than movie.

napari/movie/naparimovie.py
```diff
@@ -0,0 +1,364 @@
...
+
+
+class Movie:
+    def __init__(self, myviewer=None, inter_steps=15):
+
+        """Standard __init__ method.
+
+        Parameters
+        ----------
+        myviewer : napari viewer
```
sofroniewn: Can we go with the name `viewer` here instead of `myviewer`, it is more consistent with naming with the rest of the repo

napari/__init__.py
```diff
@@ -36,6 +36,7 @@
     view_points,
     view_vectors,
 )
+from .movie.naparimovie import Movie
```
sofroniewn: I'd put an `__init__.py` inside the folder `movie` and rename `naparimovie.py` to just `movie.py` (no need for the extra napari now this will live in the main repo. Inside the `__init__.py` you can do from `from .movie import Movie`, and then inside this file you can also do `from .movie import Movie` both of which are a little cleaner.

I'm also fine with switching `Movie` to `Animate`. Maybe then I'd use `animation` for the folder and `animate.py` and `Animate` for the file and class, so in this file you'd do `from .animation import Animate` and inside the `__init__.py` in the folder `from .animate import Animate`. What does @jni think?

napari/movie/naparimovie.py
```diff
@@ -0,0 +1,364 @@
...
+    def get_new_state(self):
+        """Capture current napari state
+
+        Returns
+        -------
+        new_state : dict
+            description of state
+        """
+
+        new_state = {
```
sofroniewn: As noted in the discussion in https://github.com/napari/napari/issues/448#issuecomment-563994868 once #686 goes in we can start using the serialization of the viewer state, but for now this is fine for me, though mayb @jni has preferrences

napari/movie/naparimovie.py
guiwitz: Sure I'll change it. As we are on this topic, I have to admit I'm not sure about the mechanism of the viewer key-binding functions. I think they all HAVE to have "viewer" as parameter, but I'm not sure which viewer it refers to. Currently I have this kind of definition ```def capture_keyframe_callback(self, viewer)```  where I'm not using ```viewer``` at all since I'm using ```self.myviewer```. Any recommendation on making this cleaner ?

napari/movie/naparimovie.py
```diff
@@ -0,0 +1,364 @@
...
+        # establish key bindings
+        self.add_callback()
+
+    def finish_movie(self):
+        self.release_callbacks()
+
+    def add_callback(self):
+        """Bind keys"""
+
+        self.myviewer.bind_key("f", self.capture_keyframe_callback)
```
sofroniewn: I think I'd like to make these keybindings more complex than just presses, like "command+shift+f" so they don't conflict with inbuilt ones / ones likely to be in regular use inside plugins. I'm open to suggestions though

guiwitz: One needs a lot of these key-presses when creating the key-frames and going back and forth within them, so this is why I'd prefer to keep them simple. Note that the bindings are released when movie recording is over so it should not conflict too much.

napari/movie/naparimovie.py
```diff
@@ -0,0 +1,364 @@
...
+        self.anim = FuncAnimation(
+            self.fig,
+            self.update,
+            frames=np.arange(len(self.states_dict)),
+            init_func=self.movie_init,
+            blit=False,
+        )
+        plt.show()
+
+        self.anim.save(name, dpi=resolution, fps=fps)
```
sofroniewn: I don't think we need the matplotlib dependency here and these lines of code and instead we can just use [an imageio mp4 writer](https://imageio.readthedocs.io/en/stable/examples.html#convert-a-movie) here instead along with the `collect_images` function, similar to the approach in `make_gif`.

Maybe something like
```python
stack = self.collect_images()
name = 'name.mp4'
fps=15
writer = imageio.get_writer(name, fps=fps)

for im in stack:
    writer.append_data(im)
writer.close()
```
Is there a reason @guiwitz you didn't go with this approach and went with matplotlib instead? I havn't tried it so am not 100% sure it will work, but I think it will.

The reason we want to avoid matplotlib is that it is a rather large and complex dependency. We don't use it anywhere else in the repo right now and have no plans to, so I think we want to avoid adding it here.

I think this approach will use only one function instead of four and reduce the overall complexity of the PR

napari/movie/naparimovie.py
```diff
@@ -0,0 +1,364 @@
...
+        -------
+        name : str
+            name to use for saving the movie (can also be a path)
+        """
+
+        # create the image stack with all snapshots
+        stack = self.collect_images()
+
+        imageio.mimsave(
+            name, [stack[i, :, :, :] for i in range(stack.shape[0])]
```
sofroniewn: you should be able to just use `[im for im in stack]` for the list comprehension

napari/movie/naparimovie.py
guiwitz: To be honest I have only used matplotlib to make movies until now and didn't know imageio could be used for that. I see the point of not depending on matplotlib. I'll switch to imageio then.

sofroniewn (commented):
Looking great @guiwitz I've made a first pass through and made some fairly minor comments around naming conventions and put in a proposal on how to use [an imageio mp4 writer](https://imageio.readthedocs.io/en/stable/examples.html#convert-a-movie) instead of matplotlib, as for reasons explained in my comment about the size and complexity of the matplotlib dependency I think we'd like to avoid it in this case. My apologies for not catching that when I looked at your repo.

If you can make those changes then I can start testing the functionality and we can start thinking about how to best add unit tests. Can you create a folder called `tests` inside this one. Then inside that can you create a file called `test_util.py` and add tests for each of the functions inside `util.py`. If you look at some of our [other test files](https://github.com/napari/napari/blob/master/napari/layers/shapes/tests/test_shape_list.py) you can see the conventions we follow inside them. Functions are named `test_*` and generally feature minimal example data and some `assert` statements that verifies that not only does the function being tested run but it also produces the correct output. Sometimes we use [parameterized tests](https://github.com/napari/napari/blob/5d7623d94a630f88dded5e9682f72cabfac58b93/napari/util/tests/test_colormaps.py#L8) to test many input parameters for the same test. We'll also then want to have a file `test_movie.py` where we test the movie making class, with a mock viewer, including potentially saving out mock data (though that can maybe come later). 

We can help you with all of this, and thanks again for starting this PR. I think a lot of people are going to be really excited about making movies in napari!

napari/movie/naparimovie.py
sofroniewn: fair enough, further down the line we'll maybe want to add buttons to the GUI to create one of these movie objects and then visually indicate in the GUI that keybindings have changed, and then add a button to finish movie making, we'll also want to add them to our keybindings info dialog (future PR). I'm happy to leave as is for now. We might also do a pretty major relook at all our keybindings at some point and try and establish standard patterns across them and so we can revisit at that point. What do @jni and @royerloic think?

napari/movie/naparimovie.py
sofroniewn: @kne42 can you answer this one

napari/movie/naparimovie.py
guiwitz: I replaced matplotlib with imageio. Making mp4 movies requires now ```pip install imageio-ffmpeg```. Not sure whether we want to include this in the dependencies or let people decide to install it on their own. Also the code is simpler now, as the same function can generate mp4 and gif, and I could get rid of several matplotlib related functions.

One thing that we might need to look into is how to adjust the resolution of the movies. I don't know how to set the resolution when capturing the napari view. The ```viewer.screenshot()``` function doesn't seem to offer options.

napari/movie/naparimovie.py
guiwitz: This is gone now.

napari/movie/naparimovie.py
kne42: @guiwitz The name of the parameter in `capture_keyframe_callback`  doesn't actually matter. It is bound to the viewer that calls `bind_key` on `capture_keyframe_callback`, as in `self.viewer.bind_key("f", capture_keyframe_callback)`. Since, as you said, `viewer` is not used, you can ignore it, but it is required to have that signature to be registered as a callback.

I think, in the future, we'll want to rework the keybinding system to be more general than just applying to viewers and layers so that it supports additional modes/plugins like yours in which it would make more sense :) I'll make an issue for it shortly.

GenevieveBuckley:
> I've also been rolling the name in my head for a few days and have come to like "Animation" or "Animate" over "Movie". What do people think?

If you don't like "movie" then I'd prefer "animation" over "animate" (mostly because I think I could remember that better).

napari/movie/naparimovie.py
GenevieveBuckley: > the same function can generate mp4 and gif

This is great! I hadn't realised easy to make gifs were a possibility here, and they're great for putting in README files, etc.

VolkerH:
Regarding the naming:  what about the longish but descriptive "KeyframeAnimation" and "keyframe_animate" ?

If I were to read the source code and look at file names without the background knowledge from these discussions I might mistake "Movie"  for something that is related to nD datasets with a time dimension and I might mistake "Animate" as the play button functionality that was recently introduced.

napari/movie/naparimovie.py
jni: I like single stroke keybindings in general (again, mostly people will not be typing text into napari), and even more so in specific, short-lived modes.

jni:
imho `Animation` is a good class name and `animate` a good function name. I'm not sure yet whether we can make this be purely functional. I had no idea what "keyframes" were until quite recently so I would not like that in the class name.

napari/movie/naparimovie.py
guiwitz: Thanks for the infos @kne42, I'm glad I didn't make a complete mess with these key-bindings!

napari/animation/animate.py
```diff
@@ -0,0 +1,303 @@
...
+        self.create_steps()
+        for i in range(len(self.states_dict)):
+
+            self.update_napari_state(i)
+            images.append(self.viewer.screenshot())
+
+        image_stack = np.stack(images, axis=0)
+        return image_stack
+
+    def update_napari_state(self, frame):
```
sofroniewn: can this be `update_viewer_from_state`

napari/animation/animate.py
```diff
@@ -0,0 +1,303 @@
...
+        """Set the napari viewer to a given interpolated frame
```
sofroniewn: can you drop the word "napari" as not needed if inside the main repo

napari/animation/animate.py
```diff
@@ -0,0 +1,303 @@
...
+        """
+
+        self.current_frame = frame
+
+        for i in range(len(self.key_frames[frame]["sliders"])):
+            self.viewer.dims.set_point(i, self.key_frames[frame]["sliders"][i])
+
+        # set visibility of layers
+        for j in range(len(self.viewer.layers)):
+            # if self.key_frames[frame]["vis"]:
```
sofroniewn: can this commented if statement be deleted?

napari/animation/animate.py
```diff
@@ -0,0 +1,303 @@
...
+        self.set_to_keyframe(new_frame)
+
+    def key_back_frame(self, viewer):
+        """Go backwards in key-frame list"""
+
+        new_frame = (self.current_frame - 1) % len(self.key_frames)
+        self.set_to_keyframe(new_frame)
+
+    def set_to_keyframe(self, frame):
+        """Set the napari viewer to a given key-frame
```
sofroniewn: Can you drop the word "napari" here and elsewhere as not needed in main repo

napari/animation/animate.py
```diff
@@ -0,0 +1,303 @@
...
+        self.viewer.bind_key("r", None)
+        self.viewer.bind_key("d", None)
+
+        self.viewer.bind_key("a", None)
+        self.viewer.bind_key("b", None)
+
+        self.viewer.bind_key("w", None)
+
+    def get_new_state(self):
+        """Capture current napari state
```
sofroniewn: Can you drop the word "napari" here and elsewhere as not needed in main repo

napari/animation/animate.py
```diff
@@ -0,0 +1,303 @@
...
+        # adjust slider positions
+        for i in range(self.interpolated_states["sliders"].shape[1]):
+            self.viewer.dims.set_point(
+                i, self.interpolated_states["sliders"][frame][i]
+            )
+
+        # update view
+        self.viewer.window.qt_viewer.view.camera.view_changed()
+
+    def make_movie(self, name="movie.mp4", fps=20, quality=5):
```
sofroniewn: can you also expose a keyword arg `format=None` that get's passed on to `get_writer` - see [function signature here](https://imageio.readthedocs.io/en/stable/userapi.html#imageio.get_writer).

napari/animation/animate.py
```diff
@@ -0,0 +1,303 @@
...
+
+        # creat all states
+        self.create_steps()
+
+        # create imageio writer and add all frames
+        _, extension = os.path.splitext(name)
+        if extension == '.mp4':
+            writer = imageio.get_writer(name, fps=fps, quality=quality)
+        else:
+            writer = imageio.get_writer(name, fps=fps)
```
sofroniewn: if you pass `quality` to a non-mp4 `get_writer` does it error or just get ignored? If ignored can we always pass, as it will make our code a little cleaner

napari/animation/animate.py
```diff
@@ -0,0 +1,303 @@
+import numpy as np
+import imageio
+import copy
+import os
+
+from . import util
+
+
+class Animate:
```
sofroniewn: I think @jni had in mind this class be called `Animation`, maybe he can confirm

sofroniewn (commented):
This is coming along really nicely @guiwitz. I left some more minor comments around naming and the imageio writing. I'll do some thorough testing and review tomorrow.

napari/animation/animate.py
guiwitz: I had tried this and it generates an error, which is why I have this condition. But I can again split this into two functions if you think that's cleaner.

guiwitz:
I don't know why but suddenly the captured frames had a different size between the 2D and 3D views (I never had that issue before). I think the number of sliders changes the size of the canvas. So now I capture the size of the first frame and force the size of the next ones.

As I had to investigate those size issues, I seized the opportunity to add an option to change the movie resolution. I think it can be quite important for people who want to make high-res movie, but let me know if there's any issue with this and I'll roll back.

napari/animation/animate.py
sofroniewn: keeping as one function is better, but I'd like to mirror the api of `get_writer` as much as possible, and I don't really like splitting based on the extension, which seems brittle and special case like.

What happens if you try and make an mp4 without setting quality? Is `5` the default value inside imageIO too?

If so we could make `quality=None` be default and then splitting on `if quality is not None:` that means if there is a non mp4 writer that uses `quality` we now have access to that too, if someone tries to make an mp4 without setting quality things will still be ok, and if they want to set quality they can too.

How does this sound?

napari/animation/animate.py
```diff
@@ -0,0 +1,336 @@
...
+    def key_interpolframe(self, viewer):
+        """Progress through interpolated frames"""
+
+        self.create_steps()
+
+        new_frame = (self.current_interpolframe + 1) % len(self.states_dict)
+        self.update_viewer_from_state(new_frame)
+        self.current_interpolframe = new_frame
+
+    def collect_images(self):
```
sofroniewn: Am I right in noting `collect_images` is no longer called, but you now do all the screenshooting inside `make_movie`? That's great, I think there should only be one function here that calls the screenshot method, but then `make_movie` should have the option to also just collect all the images into an array and return the array too.

So maybe if make `name=None` be the default argument to `make_movie` and `if name==None` then `make_movie` returns `image_stack` the 3D numpy array of screenshots without writing a file, if `make_movie` is actually writing a file I'd just return the path to the movie file. Curious what @jni thinks, but it is definitely nice that the movie writing now happens frame-by-frame, so if someone wants a movie that is 10000s of frames you don't have to create a numpy array with all that data in it in memory

napari/animation/animate.py
```diff
@@ -0,0 +1,336 @@
...
+        # adjust slider positions
+        for i in range(self.interpolated_states["sliders"].shape[1]):
+            self.viewer.dims.set_point(
+                i, self.interpolated_states["sliders"][frame][i]
+            )
+
+        # update view
+        self.viewer.window.qt_viewer.view.camera.view_changed()
+
+    def make_movie(
```
sofroniewn: I think @jni imagined `make_movie` -> `animate`

napari/animation/animate.py
```diff
@@ -0,0 +1,336 @@
...
+        self,
+        name="movie.mp4",
+        fps=20,
+        resolution=1,
+        compression_quality=5,
+        format=None,
```
sofroniewn, line 296: can we add a keyword arg `with_viewer=False` that gets passed to the screenshot method and optionally includes the napari window when making the screenshot - see here for more detail https://github.com/napari/napari/blob/41d8617f9957fef8c2bd3f45ea04c3efaef1a8ef/napari/viewer.py#L65

napari/animation/animate.py
```diff
@@ -0,0 +1,336 @@
...
+                fps=fps,
+                quality=compression_quality,
+                format=format,
+                output_params=['-timeout', '3'],
+            )
+        else:
+            writer = imageio.get_writer(name, fps=fps, format=format)
+        for frame in range(len(self.interpolated_states["ndisplay"])):
+            self.update_viewer_from_state(frame)
+            self.viewer.window.qt_viewer.canvas.size = (
```
sofroniewn: I think we want to push this logic into the [viewer.screenshot](https://github.com/napari/napari/blob/master/napari/viewer.py#L65) potentially including adding keyword args to set desired screenshot shape / height / aspect ratio / resolution etc. inside that function that this function can then just also set. I'm not sure if we need all of those to be settable, but maybe having something like `max_shape=None` be a keyword argument so that if `None` is set then it just returns the screenshot as is, and if `max_shape` is an `int` then it will rescale the image so that the largest axis has that length (preserving the aspect ratio).

This function should not ever have to look at the `viewer.window` property or anything inside it, and so if someone passes `max_shape=None` then maybe I'd take a single screenshot, determine the current max shape and then call `self.viewer.screenshot` with `max_shape` set to that value to ensure that the screenshots from that point on always have the right size.

If we go with `max_shape` I'm not sure we then also need a `resolution` keyword argument as I think they are redundant right?

guiwitz: Very good idea. This will unclutter the movie function and at the same time offer the resolution option to ```screenshot()``` in general.

napari/animation/animate.py
guiwitz: Very good idea. ```collect_images``` was just a left-over from previous iterations but I kept it around in case people need the stack, but what you propose removes a lot of redundancy.

guiwitz:
I'm busy with my actual job :wink: these next days, so the next iterations might take a few days, but it's not forgotten!

napari/animation/animate.py
jni: I'm not sure a 3D stack is a great idea, as it'll tend to be huge. I might prefer a *generator* of images that the user can then consume as they please — perhaps even using `np.fromiter` if they do want the giant array! I haven't looked at the rest of the code closely enough to know whether this is feasible...

napari/animation/animate.py
jni: He did! :smile:

jni:
@guiwitz don't worry, we are all aware of the pitfalls of open source development! Thank you for all your work so far! :heart:

sofroniewn:
Yes, no worries @guiwitz, this PR is coming a long at a great pace and I think we're close to converging :-)

napari/animation/animate.py
sofroniewn: @jni it should be, but that maybe suggests a slightly different way of thinking about `collect_images` and `make_movie` and you might need to provide a little guidance on how that should be done. My current proposal was to just drop `collect_images` and put all the logic inside `make_movie` so that you didn't have to instantiate the full 3D array inside `make_movie`, but now with the idea of using a *generator* instead maybe we make the `collect_images` function return a generator, and then `make_movie` just does `np.fromiter` with the generator to save out each frame.

I actually like this idea the best as it means the `make_movie` logic is really only about the saving of the actual movie file, and the `collect_images` logic is only about the screenshoting and state updates of the viewer. Does this make sense @guiwitz?

napari/animation/animate.py
guiwitz: Small problem here: as mentioned before, when switching from 2D to 3D view, the canvas size is changing. But it's not changing isotropically: additional sliders at the bottom of the viewer are reducing the canvas size *only vertically*. Therefore rescaling is not sufficient. To make a movie all the images need to have the same size, so I actually *need* the ```viewer.window.qt_viewer.canvas.size``` call  in the movie making function to be able to set all frames to the exact same dimensions.

We could of course avoid that if the 2D/3D switch left the canvas size unchanged and rather the viewer itself would change.
Any thoughts on this ?

sofroniewn: ah hmm, there's not an obvious way to make it work it all cases without forcing the canvas size to be the same, which is something we might want to do. Curious if @tlambert03 has thoughts on that. Regardless as much of this logic as possible should be inside the `screenshot` method.

napari/animation/animate.py
guiwitz: I now have a python generator function ```frame_generator()``` which creates all the frames as 2D numpy arrays. I left the option to generate a single specific frame if a ```frame``` argument is passed. The ```animate()``` function now only takes care of actually making the movie.

napari/animation/animate.py
guiwitz: As suggested the writer is now create differently depending on whether a ```quality``` option is passed and not on the extension type.

guiwitz:
I think I have updated everything as discussed above. One thing remains a bit unsatisfying: we have this ```max_shape``` option now to create larger movies. That works fine when creating movies of only the canvas, i.e. ```with_frame = False``` but I didn't manage to get it to work properly for the entire frame. I don't think that it's a real problem as the ```with_frame = True``` case will only be used for demos, a case where one would probably not care too much about the image size, but it would be nice to have consistent options (now ```max_shape``` is just ignored if ```with_frame=True```).

There's multiple ways of setting the viewer size, but not of them precisely gives the expected result. ```viewer.window.resize()``` is the closest thing to a working solution that I found, but it doesn't set the viewer exactly to the size obtaining when doing ```viewer.screenshot()```. My guess is that ```viewer.window.resize()``` does not include the option panel and the sliders, but I didn't investigate much. Is there any function to exactly set the size of the whole window, i.e. the size returned by ```viewer.screenshot()``` ?

guiwitz:
Also, no idea what this test fail for windows PY_VER_3.6 and 3.7 means. Something to do with chocolatey windows update ? I don't think I touched any of that.

napari/viewer.py
```diff
@@ -79,8 +82,26 @@ def screenshot(self, with_viewer=False):
         """
         if with_viewer:
             image = self.window.screenshot()
+
         else:
+            # capture initial canvas size to reset after screenshot
```
sofroniewn: I think here I'd rather take the screenshot as is and then force the resizing using the [skimage resize function](https://scikit-image.org/docs/stable/api/skimage.transform.html#skimage.transform.resize). This should avoid triggering the canvas to actually change size and should be much faster (I guess).

Using `resize` will also allow us to solve our aspect ratio problem as it allows you to pass a tuple for shape and specify things like zero padding using the `mode` keyword argument. We should then replace the `max_shape` keyword argument with one just called `shape` which is a 2-tuple (which you may need to add a [3] onto as the screenshots give back rgb images I think)

This approach should now guarantee that if we request say a `[300, 400]` image we'll always get one and should work regardless of using `with_viewer` or not as it only operates on the image after the screenshot.

If you can get this working can you also add some tests to `napari/tests/test_screenshot.py` which test both `with_viewer` and `shape`, and maybe move our existing screenshot test in there too - https://github.com/napari/napari/blob/fffd2f2a4090f773f2aec904bee9ac28a4b1890d/napari/tests/test_viewer.py#L79.

napari/animation/animate.py
```diff
@@ -0,0 +1,344 @@
...
+
+        # create states
+        self.create_steps()
+
+        # capture SceneCanvas size of frist frame to set size of next ones
+        self.update_viewer_from_state(0)
+        frame_size = self.viewer.window.qt_viewer.canvas.size
+        # if no max size provided, take current size as default
+        if max_shape is None:
+            max_shape = max(frame_size)
```
sofroniewn: Following my comment below on using [skimage resize](https://scikit-image.org/docs/stable/api/skimage.transform.html#skimage.transform.resize) in the `screenshot` method I think we can drop the `frame_size = self.viewer.window.qt_viewer.canvas.size` and also drop the `if max_shape is None` from inside the generator.

I think we will want to add an extra line into `animate` though

napari/animation/animate.py
```diff
@@ -0,0 +1,344 @@
...
+        format: str
+            The format to use to write the file. By default imageio selects the appropriate for you based on the filename.
+        with_viewer : bool
+            If True includes the napari viewer, otherwise just includes the
+            canvas.
+        max_shape : int
+            Size to which to rescale the largest axis. Only used without
+            viewer (with_viewer = False).
+        """
+
```
sofroniewn, line 322: I think here we might need to add the lines
```python
if shape is None:
    shape = self.viewer.screenshot(with_viewer=with_viewer)[:2]
```
and then use `shape` inside the `frame_generator`.
Where not since my comment about using [skimage resize](https://scikit-image.org/docs/stable/api/skimage.transform.html#skimage.transform.resize) in the screenshot method I'm now talking `shape` instead of just `max_shape`

napari/viewer.py
guiwitz: I'm not convinced that this is a good solution. Resizing the images will deteriorate the output. Increasing the canvas size effectively captures a larger image, while ```skimage.transfrom.resize()``` just interpolates. As an example I add here two images which in the end have the same size, but one is resized with skimage (top) and for the other the canvas size is adjusted (bottom). 
with_skimage
with_canvas

Maybe @jni has an opinion here ?

jni: I don't, sorry! 😂 @sofroniewn is speed the only consideration with canvas resizing? Or is it also because the viewer itself will now change size? I agree that the latter is not ideal. But, the image quality problem is significant. Above, it's subtle, but visible. In other cases it might make a major difference.

guiwitz: I just want to add here that the whole function is written in such a way that the canvas is automatically reset after movie making, so this is transparent to the user. So the only issue might be speed. I didn't do extensive tests, but it never seemed to be an issue to me.

sofroniewn: @guiwitz your point about the high quality rendering is well taken, but if people don't provide a scale-like argument here they will still get back the full quality stuff and I imagine that if they are providing a scale-like argument they might be doing so to get something much smaller anyway. I will add though that @royerloic and @jni did seem to prefer the high quality renderings though, so I don't think this is settled yet.

I've also only had time to test the functionality in a cursory way myself, but I am noting some unexpected behaviors when the rescaling is done this way, potentially caused by limits Qt places on how it does the rescaling - for example if I run `examples/add_image.py` to display the standard 2D astronaut picture and then run commands like the following from the console I get
```python
viewer.screenshot(max_shape=None).shape # (796, 1600, 4)
viewer.screenshot(max_shape=1000).shape # (994, 2000, 4)
viewer.screenshot(max_shape=100).shape # (400, 400, 4)
viewer.screenshot(max_shape=10).shape # (400, 400, 4)
```
whereas i would have expected to get
```python
viewer.screenshot(max_shape=None).shape # (796, 1600, 4)
viewer.screenshot(max_shape=1000).shape # (498, 1000, 4)
viewer.screenshot(max_shape=100).shape # (50, 100, 4)
viewer.screenshot(max_shape=10).shape # (5, 10, 4)
```

I still lean towards using `skimage.transfrom.resize` but maybe figuring out how to make sure people don't call it when unnecessarily - when they want the full resolution stuff, and to make sure it does the best things possible when it is doing downsampling etc.

I'll try and have a more detailed look at this again soon, but will be traveling for the holidays now

sofroniewn: following up, it seems like there is a factor of 2 problem for intermediate sizes, small sizes are limited by our Qt limits, and big sizes (say >20_000) cause vispy crashed. if there is a way to get the behaviour working with `skimage.transfrom.resize` as best as possible I think it would be worth trying

guiwitz:
I'm finally going to start working on this again in the next days. Regarding the discussion about size/resolution I agree that at the moment the simplest and stablest solution is to just use the skimage resize function, and so I'll just skip all the canvas resizings currently in the code.

As as separate question, @sofroniewn I saw that you implemented a camera model in #854. This will be useful to capture camera states to make movies. What do you recommend: should I just ignore that PR for the moment, build directly on top of it, or wait until its full integrated in napari ?

sofroniewn:
> I'm finally going to start working on this again in the next days.

That's great!! Thanks

> I agree that at the moment the simplest and stablest solution is to just use the skimage resize function, and so I'll just skip all the canvas resizings currently in the code.

Thanks - the resize functionality should only be called if someone adds dims sliders / switches 2D to 3D / or resizes the canvas themselves during the movie making right? I think those will be pretty minor use cases for now, so we can always revisit with a higher quality solution in the future.

> I saw that you implemented a camera model in #854. This will be useful to capture camera states to make movies. What do you recommend: should I just ignore that PR for the moment, build directly on top of it, or wait until its full integrated in napari ?

I think you can make the above mentioned canvas resizing change now, but we might want to wait until a couple other PRs merge before we doing much more with this one. Having #854 and #851 in will be really nice because then we could use those for the state dictionary and then you'd be able to animate basically any change in napari, which I think is ultimately where we want to be. I can make a push on those soon.

Updating this PR after #791 goes in will also be nice too as it will make the keybinding stuff a little more elegant (right now I can trigger some unexpected behavior on our end that will be fixed by this). Maybe @jni can give that one a look soon too.

royerloic:
> I'm finally going to start working on this again in the next days. Regarding the discussion about size/resolution I agree that at the moment the simplest and stablest solution is to just use the skimage resize function, and so I'll just skip all the canvas resizings currently in the code.

My opinion on this: allow resizing the canvas to anything that fits in the screen, and if that's bigger, resize canvas to the biggest size that fits and 'divides' the desired size, and has the correct aspect ratio, and then rescale the image to get to what is asked. For example if your screen only accommodates 1024x768, and what is asked is 2048x2048, I would set the canvas to 512x512, and then do a good lanczos rescale to 2048x2048. That's not perfect but that's the best we can do right now given the limits of using GUIs as render engines... Important: limiting ourselves to integer multiple rescales would reduce aliasing artefacts and other annoying artifacts. It should be noted that if the window is hidden or across different monitors, weird stuff will happen... So maybe we should warn the user to leave things alone while rendering happens...

sofroniewn:
> My opinion on this: allow resizing the canvas to anything that fits in the screen, and if that's bigger, resize canvas to the biggest size that fits and 'divides' the desired size, and has the correct aspect ratio, and then rescale the image to get to what is asked. For example if your screen only accommodates 1024x768, and what is asked is 2048x2048, I would set the canvas to 512x512, and then do a good lanczos rescale to 2048x2048. That's not perfect but that's the best we can do right now given the limits of using GUIs as render engines.

I think that was the preference of @guiwitz and @jni too, which is fine with me (apologies that we seem to be oscillating on this one @guiwtz!!). I do like the possibility of a higher quality solution, but we just have to make sure that we aren't clipping or padding with black when we do those canvas resizes and that the zoom on the object is preserved correctly.

I'll also note that the camera model PR #854 is waiting on the coordinate system update in #885

guiwitz:
I went the opposite way and now think that for the moment it's just simpler to rescale things with skimage. In the end we probably want something that allows for "true" resizing" but at the moment it just makes things confusing for the user, and there's too many uncontrolled aspect (like size limits imposed by Qt or size issues having to do with presence/absence of sliders) that make things very clumsy. With just the skimage resize function, things are simple. The user can just use a scale_factor in the animate function if he wants a larger/smaller movie and we don't even have to touch the screen capture function in viewer.py.

I still noticed a few glitches for weird use cases (unexpected dimension switches) that I will try to fix. One open question is how to handle the movie making itself. I think at some point @sofroniewn said he didn't want ffmpeg to be an installation requirement. I agree, but I think people who do want it (or install it) should be able to directly create movies without calling themselves ffmpeg. We could add a check in the animation function to verify if ffmpeg is installed. If not a warning appears and images are stored in a folder. Any opinions on this ? 

Finally, there's still no tests on the functions that deal with the viewer/animation. I'm honestly not sure how to write them, as I'm not sure what we would check for as I wrote tests for the camera movements implemented in utils and the viewer already has its own general tests. Can someone give me directions on how to proceed ?

royerloic:
> I went the opposite way and now think that for the moment it's just simpler to rescale things with skimage. In the end we probably want something that allows for "true" resizing" but at the moment it just makes things confusing for the user, and there's too many uncontrolled aspect (like size limits imposed by Qt or size issues having to do with presence/absence of sliders) that make things very clumsy. With just the skimage resize function, things are simple. The user can just use a scale_factor in the animate function if he wants a larger/smaller movie and we don't even have to touch the screen capture function in viewer.py.

Ok, given the brittleness around Qt it might be better for now, indeed, to let people scale the canvas by hand, and do rescaling of the image. As long as it is possible to have no rescaling at all and then use whatever resolution the canvas has, then that's ok -- otherwise we will always have weird rescaling artefacts. 

> 
> I still noticed a few glitches for weird use cases (unexpected dimension switches) that I will try to fix. One open question is how to handle the movie making itself. I think at some point @sofroniewn said he didn't want ffmpeg to be an installation requirement. I agree, but I think people who do want it (or install it) should be able to directly create movies without calling themselves ffmpeg. We could add a check in the animation function to verify if ffmpeg is installed. If not a warning appears and images are stored in a folder. Any opinions on this ?

I agree, that's a great approach, I think this is what we had discussed with @sofroniewn... You can let the user know with a warning that if they install ffmpeg (and we provide the pip line for it) then they get a nice shiny movie instead of a folder. In some cases some people will want to have individual png files to be able to make their own movie, so I would suggest to make sure there is a way that a sequence of files can be returned. Also, some more _basic_ access to ffmpeg quality presets for the mp4 would be good (kwargs perhaps?). This can be addressed in a follow-up PR, just want to mention it here for discussion. What does @jni and @sofroniewn think?   

> 
> Finally, there's still no tests on the functions that deal with the viewer/animation. I'm honestly not sure how to write them, as I'm not sure what we would check for as I wrote tests for the camera movements implemented in utils and the viewer already has its own general tests. Can someone give me directions on how to proceed ?

vigji:
Hey guys, I would like to know it this is moving forward or being replaced by other PRs!
I was using the `naparimovie` package and it was great, I think that this functionality would be very neat to have. I can also try to work on this feature if it can help!

guiwitz:
Hey @vigji, I cleaned-up this PR at the end of February, and it has stalled for a bit now but I don't think there's another PR in the works about making movies with napari. I think the maintainers have just been busy with other topics. It could be that they are waiting on PRs #851 and #854 which should implement a clean way to serialize camera properties to simplify some of the code of this PR but I'm not sure. Maybe @sofroniewn knows more.

As mentioned in previous posts, something that is still missing are tests for the viewer/animation as I'm not sure exactly how to write these. So if you have any experience with that, you are welcome to help!

sofroniewn:
> Hey @vigji, I cleaned-up this PR at the end of February, and it has stalled for a bit now but I don't think there's another PR in the works about making movies with napari. I think the maintainers have just been busy with other topics. It could be that they are waiting on PRs #851 and #854 which should implement a clean way to serialize camera properties to simplify some of the code of this PR but I'm not sure. Maybe @sofroniewn knows more.

That's correct. My apologies that this has all taken so much longer, at the time I thoughts those changes would be fairly quick and then we could get this in. I do think we are close now to working on those other issues and then we can revisit this.

In other parts of the codebase we have dropped our scikit-image dependency, which we were using for the rescaling. I wonder if we shouldn't simply crop/ pad with 0 @guiwitz rather than rescale. We've also now switched to using `from imageio import imsave` for the saving, which we want to import lazily (i.e. in the function that uses it) so as to keep our startup time fast.

guiwitz:
> @guiwitz this PR is finally ready for review!! I'm so sorry that it has taken us so long to get back to this and your incredible napari movie PR #780 has been sitting there the whole time. We really are trying to improve our development process so that PRs don't get stranded like that anymore, and once this merges I'd love us to pick that up one up again!!

@sofroniewn Great, looking forward to see #854 merged to work again on this! Even at my small scale I see how difficult it is to manage projects in development and it's actually interesting to follow how the napari-team is managing this. I haven't done much on napari recently (except as a user of course!). I hope I can familiarise myself with all the changes ASAP and update my pull request. As always, progress speed will depend on how much "spare time" I find (unfortunately I didn't get the CZI imaging scientist grant I had hoped for and that would have allowed me to justify working more on this).

sofroniewn:
> @sofroniewn Great, looking forward to see #854 merged to work again on this! Even at my small scale I see how difficult it is to manage projects in development and it's actually interesting to follow how the napari-team is managing this. I haven't done much on napari recently (except as a user of course!). I hope I can familiarise myself with all the changes ASAP and update my pull request. As always, progress speed will depend on how much "spare time" I find (unfortunately I didn't get the CZI imaging scientist grant I had hoped for and that would have allowed me to justify working more on this).

Yeah, I understand, the CZI call was incredibly competitive.

#854 has merged now though, and thanks for being understanding with us as we've been getting that and other things done. We'll want to move this PR over to using that now, so instead of `self.viewer.window.qt_viewer.view.camera.get_state()` we just need to track `viewer.camera.zoom`, `viewer.camera.center`, `viewer.camera.angles` and `viewer.camera.ndisplay`, and that will work for both 2D and 3D rendering.

We've also dropped our dependency on skimage, but can use `imageio.imwrite` instead of `skimage.io.imsave` and maybe `scipy.ndimage.zoom` instead of `skimage.transform.rescale`.

We might also want to think about the keybindings in light of #1321, though clearly we have work to do there to think about a good policy and how to enable/ disable keybindings, which might be something we accomplish through our plugin interface eventually. Not sure for now though if we might want to place add a modifier to the keybindings. Curious what @jni thinks.

We've also in places started using an "experimental" name space to allow us to add features more rapidly with the understanding that the API might change as we improve them (see discussion in #1564). I'm not sure about @jni but we could try and get this in ASAP under the experimental namespace and then be able to more quickly iterate on it? i.e.

```python
from napari.experimental import Animation

anim = Animation(viewer)
anim.animate('my_movie.mp4')
```

Once API stabilizes and we expand the feature capabilities we could then move this out of experimental. Any additional thoughts @jni?

sofroniewn:
Also cc @GenevieveBuckley - it would be cool to one day think about autogenerating gifs/ mp4s for our docs too - not quite sure how to think about that yet, but just wanted to mention!!

sofroniewn:
Note also that I think we should merge this without waiting for #851 as I had once mentioned, that's taking a long time and we can always use that functionality when it is ready after this has merged

jni:
@guiwitz ditto, sorry to hear about the Imaging Scientist grant, but I second @sofroniewn's point that it was incredibly competitive. I know three people who applied *just in Melbourne!*

At any rate, in case it helps with motivation, I've had several researchers tell me that making movies is the last thing keeping them from using napari, so I would very much like to get this in, and I would strongly prioritise reviewing and merging quickly if you could get it working again on current master. Especially if you put the feature in the experimental namespace, I think we could merge it very quickly. Once it's in we can more effectively distribute the load of maintaining it!

Anyway, please let us know if there is anything we can do to help.

sofroniewn:
After discussion with @jni at team meeting we've decided that this functionality does after all belong in its own repo as a plugin. I've created a new `napari-animation` repo and made this PR https://github.com/sofroniewn/napari-animation/pull/1 that will refactor this code to use some of the new dataclasses on the viewermodel. You can read more about the approach being taken in that PR, but now because it is in it's own repo we'll be able to iterate much faster to get this released.

I want to thank @guiwitz again for his original work on naparimovie over a year ago that got all of this started!!

Svetlana1st42:
Version 99.0.1153.0 (Official build) canary (64-bit)
There is a new Canary build, 100.0.1154.0. If you are still having this issue after updating to the latest version, I would like you to reset your settings. This will reset your startup page, new tab page, search engine, and pinned tabs. It will also disable all extensions and clear temporary data like cookies. Your favorites, history, and saved passwords will not be cleared.
Click on the three dots menu button.
Click on Settings.
Click on Reset settings.
Click on Restore settings to their default values.
Click on Reset.
Eric

        
 
29 people found this reply helpful Was this reply helpful? Yes   No
Replies (22) 
Question Info
Last updated February 28, 2023
Views 14,448
Applies to:
Microsoft Edge  / Website issues  / Windows 11

You’re invited to try Microsoft 365 for free
Unlock now
What's new
Surface Pro 9
Surface Laptop 5
Surface Studio 2+
Surface Laptop Go 2
Surface Laptop Studio
Surface Go 3
Microsoft 365
Windows 11 apps
Microsoft Store
Account profile
Download Center
Microsoft Store support
Returns
Order tracking
Virtual workshops and training
Microsoft Store Promise
Flexible Payments
Education
Microsoft in education
Devices for education
Microsoft Teams for Education
Microsoft 365 Education
Education consultation appointment
Educator training and development
Deals for students and parents
Azure for students
Business
Microsoft Cloud
Microsoft Security
Dynamics 365
Microsoft 365
Microsoft Power Platform
Microsoft Teams
Microsoft Industry
Small Business
Developer & IT
Azure
Developer Center
Documentation
Microsoft Learn
Microsoft Tech Community
Azure Marketplace
AppSource
Visual Studio
Company
Careers
About Microsoft
Company news
Privacy at Microsoft
Investors
Diversity and inclusion
Accessibility
Sustainability

PR #784, opened 2019-12-10
Updating qt_console with better resource management. 

shanaxel42:
# Description
So the issue with the existing code was that if a QtConsole was instantiated with an existing shell (InProcessInteractiveShell) self. kernel_manager and self. kernel_client were not getting properly set, meaning the shutdown code wasn't doing anything. This PR updates the QtConsole to create a QtInProcessKernelManager with an existing kernel if one exists and then sets both attributes. It also adds the shutdown() calls to the test_qt.py tests. Everything works with `ipykernel 5.1.3` so I also removed the pinned dependency.  

## Type of change

# References

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/_qt/qt_console.py
+++ b/napari/_qt/qt_console.py
@@ -50,12 +50,9 @@ class QtConsole(RichJupyterWidget):
             # If there is an existing running InProcessInteractiveShell
             # it is likely because multiple viewers have been launched from
             # the same process. In that case create a new kernel.
-            kernel_manager = QtInProcessKernelManager()
-            kernel_manager.start_kernel(show_banner=False)
-            kernel_manager.kernel.gui = 'qt'
-
+            # Connect existing kernel
+            kernel_manager = QtInProcessKernelManager(kernel=shell.kernel)
             kernel_client = kernel_manager.client()
-            kernel_client.start_channels()
 
             self.kernel_manager = kernel_manager
             self.kernel_client = kernel_client
@@ -87,7 +84,6 @@ class QtConsole(RichJupyterWidget):
             raise ValueError(
                 'ipython shell not recognized; ' f'got {type(shell)}'
             )
-
         # Add any user variables
         user_variables = user_variables or {}
         self.push(user_variables)
@@ -98,7 +94,5 @@ class QtConsole(RichJupyterWidget):
         # self.execute_on_complete_input = True
 
     def shutdown(self):
-        if self.kernel_client is not None:
-            self.kernel_client.stop_channels()
-        if self.kernel_manager is not None:
-            self.kernel_manager.shutdown_kernel()
+        self.kernel_client.stop_channels()
+        self.kernel_manager.shutdown_kernel()
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -3,7 +3,7 @@ zarr>=2.3.0
 numcodecs!=0.6.4
 fsspec>=0.3.3
 imageio>=2.5.0
-ipykernel==5.1.1
+ipykernel>=5.1.1
 numpy>=1.10.0
 qtpy>=1.7.0
 qtconsole>=4.5.1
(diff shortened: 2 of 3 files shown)


kne42:
Awesome! Thanks for getting to the bottom of this :)

I have no idea what any of this code does (or did) so I'll leave @sofroniewn to review this.

requirements/default.txt
```diff
@@ -3,7 +3,7 @@ zarr>=2.3.0
 numcodecs!=0.6.4
 fsspec>=0.3.3
 imageio>=2.5.0
-ipykernel==5.1.1
+ipykernel>=5.1.3
```
jni: Do we need 5.1.3 as opposed to 5.1.1 or earlier?

napari/_qt/qt_console.py
```diff
@@ -98,7 +94,5 @@ def __init__(self, user_variables=None):
         # self.execute_on_complete_input = True
 
     def shutdown(self):
-        if self.kernel_client is not None:
-            self.kernel_client.stop_channels()
-        if self.kernel_manager is not None:
-            self.kernel_manager.shutdown_kernel()
+        self.kernel_client.stop_channels()
+        self.kernel_manager.shutdown_kernel()
```
jni, line 98: Are you sure this change is correct? both of these can be None if the viewer is run in an IPython shell (see line 61):

```python
        elif isinstance(shell, TerminalInteractiveShell):
            # if launching from an ipython terminal then adding a console is
            # not supported. Instead users should use the ipython terminal for
            # the same functionality.
            self.kernel_client = None
            self.kernel_manager = None
            self.shell = None
            self.push = lambda var: None
```

Having said that, I just tested this, and I didn't manage to make it crash (I couldn't access the QtConsole object to use this method)...

jni (commented):
@shanaxel42 this is so cool. All your PRs are making me very happy. =) I've left a couple of questions, but in principle this looks great!

napari/_qt/qt_console.py
shanaxel42: Yea it won't crash but ideally we should also be shutting down the kernal with both the TerminalInteractiveShell and the ZMQInteractiveShell, so that's my next investigation. The updates here will fix the issues with the ipykernel but I don't think it's a full solution yet.

requirements/default.txt
shanaxel42: ooo you're right no `>= 5.1.1` should be fine. I'll change.

napari/_qt/qt_console.py
jni: No, the point with the TerminalInteractiveShell is that there is no kernel! And with ZMQ there is one but I don't think we should shut it down: napari might be one of many clients connecting to that kernel.

jni (approved):
CI is looking so happy! =D

sofroniewn (approved):
LGTM! Thanks, I will merge now.

napari/_qt/qt_console.py
shanaxel42: oh great then!

tlambert03: I was just about to open a PR about this and then found this conversation.  I'm not sure i follow what the rationale was behind removing the `if self.kernel_client is not None` conditional.
I'm getting an error now in the master branch whenever I close a viewer (when starting in ipython):

```python
In [1]: import napari, numpy as np

In [2]: v = napari.view_image(np.random.rand(10,10,10,10))

# upon closing the window

  File "/Users/talley/Dropbox (HMS)/Python/forks/napari/napari/_qt/qt_main_window.py", line 324, in closeEvent
    self.qt_viewer.console.shutdown()
  File "/Users/talley/Dropbox (HMS)/Python/forks/napari/napari/_qt/qt_console.py", line 97, in shutdown
    self.kernel_client.stop_channels()
AttributeError: 'NoneType' object has no attribute 'stop_channels'
```

PR #788, opened 2019-12-12
respect vispy max texture limits

sofroniewn:
# Description
This PR contains parts of #737 that prevent vispy segmentation faults when the array sent over is too large by downsampling the array an emitting a warning that downsampling is happening. For 2D rendering it will read the max texture limit from your GPU, i havn't figured out how to read the value for 3D so have hard coded it to 2048, the value on my machine.

To test you can use
```python
img = np.random.random((20_000, 20_000))
napari.view_image(img, is_pyramid=False) 
```
note we need `is_pyramid=False` otherwise we will try and autogenerate a pyramid (something that maybe fixed in a future PR but is out of scope for this one)

## Type of change

# How has this been tested?

- [x] This PR adds two tests in `test_vispy_big_images.py` (one 2D one 3D) both of which currently fail on master

## Final checklist:

### DIFF
--- a/napari/_vispy/tests/test_vispy_big_images.py
+++ b/napari/_vispy/tests/test_vispy_big_images.py
@@ -1,0 +1,43 @@
+import numpy as np
+from napari import Viewer
+
+
+def test_big_2D_image(qtbot):
+    """Test big 2D image with axis exceeding max texture size."""
+    viewer = Viewer()
+    view = viewer.window.qt_viewer
+    qtbot.addWidget(view)
+
+    shape = (20_000, 10)
+    data = np.random.random(shape)
+    layer = viewer.add_image(data, is_pyramid=False)
+    visual = view.layer_to_visual[layer]
+    assert visual.node is not None
+    if visual.MAX_TEXTURE_SIZE_2D is not None:
+        print(
+            'aasdfasdf', visual.MAX_TEXTURE_SIZE_2D, visual.MAX_TEXTURE_SIZE_3D
+        )
+        ds = np.ceil(np.divide(shape, visual.MAX_TEXTURE_SIZE_2D)).astype(int)
+        assert np.all(layer._scale_view == ds)
+
+    # Close the viewer
+    viewer.window.close()
+
+
+def test_big_3D_image(qtbot):
+    """Test big 3D image with axis exceeding max texture size."""
+    viewer = Viewer(ndisplay=3)
+    view = viewer.window.qt_viewer
+    qtbot.addWidget(view)
+
+    shape = (5, 10, 3_000)
+    data = np.random.random(shape)
+    layer = viewer.add_image(data, is_pyramid=False)
+    visual = view.layer_to_visual[layer]
+    assert visual.node is not None
+    if visual.MAX_TEXTURE_SIZE_3D is not None:
+        ds = np.ceil(np.divide(shape, visual.MAX_TEXTURE_SIZE_3D)).astype(int)
+        assert np.all(layer._scale_view == ds)
+
+    # Close the viewer
+    viewer.window.close()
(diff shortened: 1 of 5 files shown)


napari/_vispy/tests/test_vispy_big_images.py
```diff
@@ -0,0 +1,43 @@
...
+    qtbot.addWidget(view)
+
+    shape = (20_000, 10)
+    data = np.random.random(shape)
+    layer = viewer.add_image(data, is_pyramid=False)
+    visual = view.layer_to_visual[layer]
+    assert visual.node is not None
+    if visual.MAX_TEXTURE_SIZE_2D is not None:
+        print(
+            'aasdfasdf', visual.MAX_TEXTURE_SIZE_2D, visual.MAX_TEXTURE_SIZE_3D
```
jni: print "aasdfasdf"?

napari/_vispy/vispy_image_layer.py
```diff
@@ -233,3 +246,44 @@ def reset(self):
         self._on_rendering_change()
         if self.layer.dims.ndisplay == 2:
             self._on_contrast_limits_change()
+
+    def downsample_texture(self, data, MAX_TEXTURE_SIZE):
+        """Downsample data based on maximum allowed texture size.
+        Parameters
```
jni, line 253:
```suggestion

        Parameters
```

napari/_vispy/vispy_base_layer.py
```diff
@@ -195,3 +206,26 @@ def on_draw(self, event):
...
+    # A canvas must be created to access gl values
+    _ = Canvas(show=False)
+    MAX_TEXTURE_SIZE_2D = gl.glGetParameter(gl.GL_MAX_TEXTURE_SIZE)
+    if MAX_TEXTURE_SIZE_2D == ():
+        MAX_TEXTURE_SIZE_2D = None
+    # vispy doesn't seem to expose this so hard coding for now ....
+    # MAX_TEXTURE_SIZE_3D = gl.glGetParameter(gl.GL_MAX_3D_TEXTURE_SIZE)
+    MAX_TEXTURE_SIZE_3D = 2048
+    if MAX_TEXTURE_SIZE_3D == ():
+        MAX_TEXTURE_SIZE_3D = None
```
jni: This logic seems weird... Did you mean to replace None with 2048 here? Or...?

jni (approved):
@sofroniewn I guess I don't utterly despise this. =P But I'm still wary that it might look pretty bad and aliased to do this and I might consider using e.g. scikit-image's `downscale_local_mean` so that we don't lose information... Let's discuss during the meeting.

napari/_vispy/vispy_image_layer.py
sofroniewn: added empty lines here and elsewhere

napari/_vispy/vispy_base_layer.py
sofroniewn: No, the `MAX_TEXTURE_SIZE_3D = 2048` is meant to replace call to the `gl.glGetParameter` that is commented above until I can figure out how to make that work. The `== ()` is only needed in the case that we can get that working, but it seemed ok to have in there. Maybe a little weird though.

I changed the comments slightly so maybe it makes more sense now, but happy to do different if you recommend

napari/_vispy/tests/test_vispy_big_images.py
sofroniewn: removed :-)

napari/_vispy/vispy_base_layer.py
sofroniewn: i have now improved commented region so the code that won't execute is commented

sofroniewn:
I will merge this after the tests pass. I think in the future we could consider doing better downsampling, so using a mean instead of just a stride, and we can also consider whether this limiting is done inside out vispy view code or any `DataSource` like object that @jni or @d-v-b might be considering adding.

For now though this will prevent seg faults with a warning which is an improvement

Issue #694, opened 2019-11-14
Use new mouse function bindings and test mouse functions on layers


sofroniewn:
We need to refactor the current mouse functions on the layers to use the new mouse bindings framework in #544 and add tests for all of them, as was done for one in #684

Issue #517, opened 2019-09-07
Standardize events


sofroniewn:
## 🚀 Feature
Right now there are many events that layers generate for use within napari. There are also interesting things that happen within a layer where we might not need an event, but where a user might want an event, for example finishing drawing of a polygon or rectangle.

We should standardize our events and expose useful ones across all our layer types and viewer. I see plugin developers wanting to connect to these events as useful sources of input.

It would then be nice in our documentation if we had all the events for each of our objects clearly listed so that plugin developers could find them. We would also need to add tests for each of our events to ensure they remain stable over time.

sofroniewn:
Note this [discussion on image.sc](https://forum.image.sc/t/registering-callback-functions-for-events-in-napari/32210) from @VolkerH around the syntax for event connecting. This is something @kne42 and I have discussed before, but worth thinking about when doing any refactoring, including maybe of #785

willingc:
I'm triaging all open test issues. I'm going to close this issue as not planned in the near future. As we look at event handling in the future, the tests will undergo changes to suit the event handling library or libraries.

PR #791, opened 2019-12-13
generalize keybindings

### LINKED ISSUE #785: Better generalize keybindings
The current keybindings system could be made more general. Right now, it's mainly optimized for the viewer and layers. This creates awkward situations for different modes (see https://github.com/napari/napari/pull/780#discussion_r356111992) and will create awkward situations in the future for plugin developers.

This can be accomplished by:
- making the keybinding handling system more generalized (for example, right now, all of it lives in the `QtViewer` class)
- allowing multiple keymaps to be chained together. for example, say you have the movie mode activated while having a points layer selected. then the functions called when the keybinding is pressed will check in the following order: movie instance keymap -> movie class keymap -> points instance keymap -> points class keymap -> viewer instance keymap -> viewer class keymap

Thoughts?

kne42:
# Description
Improve keybindings via the following changes:
- move keybinding pressing & releasing logic to a new `KeymapHandler` class
- make `ViewerModel` inherit from `KeymapHandler`
- add property to auto-compute currently active keybindings
- any class inheriting from `KeymapProvider` can prepend itself to a `KeymapHandler`'s `keymap_providers` to active its keymap
- delete `InheritedKeymap` and instead use a keymap lookup chain (**breaking change**)
- passing `...` as the function to `bind_key` will serve as a blocker to ignore this keybinding provided by any keymap further down the lookup chain
- passing `None` as the function to `bind_key` only unbinds from that keymap (**breaking change**); to achieve the previous functionality, pass `...` instead
- passing `...` as the key combination to `bind_key` behaves as a wildcard, and therefore causes any keymap further down the lookup chain to be ignored

#### press any key to exit
```python
Mode.bind_key(..., Mode.exit)
```

#### blocking mode
```python
Mode.bind_key(..., ...)  # ignore all "inherited" keybindings
Mode.bind_key('Left', Mode.left)
Mode.bind_key('Right', Mode.right)
```

#### re-map
```python
Mode.bind_key('Control-S', Mode.save)
...
mode.bind_key('Control-S', ...)  # block binding "inherited" from class
mode.bind_key('S', Mode.save)    # re-map to 'S'
```

This is meant to make it easier for plugin developers and modes, such as the movie mode in #780, to set keybindings. It also adds useful functionality in allowing primary modes to overwrite all "inherited" keybindings thanks to the `...` wildcard.

I'm sure, as always, that the way I explained things probably doesn't make sense so please give suggestions on how to better explain this in the docs :)

## Type of change

# References
closes #785 

# How has this been tested?

### DIFF
--- a/napari/_qt/qt_about_keybindings.py
+++ b/napari/_qt/qt_about_keybindings.py
@@ -121,16 +121,8 @@ class QtAboutKeybindings(QDialog):
             Event from the Qt context, by default None.
         """
         col = self.viewer.palette['secondary']
-        text = ''
         # Add class and instance viewer keybindings
-        text += get_keybindings_summary(self.viewer.class_keymap, col=col)
-        text += get_keybindings_summary(self.viewer.keymap, col=col)
-
-        layer = self.viewer.active_layer
-        if layer is not None:
-            # Add class and instance layer keybindings for the active layer
-            text += get_keybindings_summary(layer.class_keymap, col=col)
-            text += get_keybindings_summary(layer.keymap, col=col)
+        text = get_keybindings_summary(self.viewer.active_keymap, col=col)
 
         # Update layer speficic keybindings if all active are displayed
         self.keybindings_strs[self.ALL_ACTIVE_KEYBINDINGS] = text
--- a/napari/_qt/qt_viewer.py
+++ b/napari/_qt/qt_viewer.py
@@ -1,4 +1,4 @@
-import inspect
+import os.path
 from pathlib import Path
 
 from qtpy import QtGui
@@ -178,8 +178,6 @@ class QtViewer(QSplitter):
         }
 
         self._update_palette(viewer.palette)
-
-        self._key_release_generators = {}
 
         self.viewer.events.interactive.connect(self._on_interactive)
         self.viewer.events.cursor.connect(self._on_cursor)
@@ -505,31 +503,13 @@ class QtViewer(QSplitter):
             and event.native.isAutoRepeat()
             and event.key.name not in ['Up', 'Down', 'Left', 'Right']
         ) or event.key is None:
-            # pass is no key is present or if key is held down, unless the
+            # pass if no key is present or if key is held down, unless the
             # key being held down is one of the navigation keys
+            # this helps for scrolling, etc.
             return
 
-        comb = components_to_key_combo(event.key.name, event.modifiers)
-
-        layer = self.viewer.active_layer
-
-        if layer is not None and comb in layer.keymap:
-            parent = layer
-        elif comb in self.viewer.keymap:
-            parent = self.viewer
-        else:
-            return
-
-        func = parent.keymap[comb]
-        gen = func(parent)
-
-        if inspect.isgenerator(gen):
-            try:
-                next(gen)
-            except StopIteration:  # only one statement
-                pass
-            else:
-                self._key_release_generators[event.key] = gen
+        combo = components_to_key_combo(event.key.name, event.modifiers)
+        self.viewer.press_key(combo)
 
     def on_key_release(self, event):
         """Called whenever key released in canvas.
@@ -539,10 +519,8 @@ class QtViewer(QSplitter):
         event : qtpy.QtCore.QEvent
             Event from the Qt context.
         """
-        try:
-            next(self._key_release_generators[event.key])
-        except (KeyError, StopIteration):
-            pass
+        combo = components_to_key_combo(event.key.name, event.modifiers)
+        self.viewer.release_key(combo)
 
     def on_draw(self, event):
         """Called whenever drawn in canvas. Called for all layers, not just top
--- a/napari/_qt/qt_viewer.py
+++ b/napari/_qt/qt_viewer.py
@@ -1,4 +1,3 @@
-import os.path
 from pathlib import Path
 
 from qtpy import QtGui
--- a/napari/components/viewer_model.py
+++ b/napari/components/viewer_model.py
@@ -8,11 +8,11 @@ from .add_layers_mixin import AddLayersMixin
 from .dims import Dims
 from .layerlist import LayerList
 from ..utils.event import EmitterGroup, Event
-from ..utils.keybindings import KeymapMixin
+from ..utils.keybindings import KeymapHandler, KeymapProvider
 from ..utils.theme import palettes
 
 
-class ViewerModel(AddLayersMixin, KeymapMixin):
+class ViewerModel(AddLayersMixin, KeymapHandler, KeymapProvider):
     """Viewer containing the rendered scene, layers, and controlling elements
     including dimension sliders, and control bars for color limits.
 
@@ -89,6 +89,8 @@ class ViewerModel(AddLayersMixin, KeymapMixin):
         self.layers.events.changed.connect(self._on_layers_change)
         self.layers.events.changed.connect(self._update_active_layer)
         self.layers.events.changed.connect(self._update_grid)
+
+        self.keymap_providers = [self]
 
         # Hold callbacks for when mouse moves with nothing pressed
         self.mouse_move_callbacks = []
@@ -235,7 +237,15 @@ class ViewerModel(AddLayersMixin, KeymapMixin):
     def active_layer(self, active_layer):
         if active_layer == self.active_layer:
             return
+
+        if self._active_layer is not None:
+            self.keymap_providers.remove(self._active_layer)
+
         self._active_layer = active_layer
+
+        if active_layer is not None:
+            self.keymap_providers.insert(0, active_layer)
+
         self.events.active_layer(item=self._active_layer)
 
     def _scene_shape(self):
--- a/napari/layers/base/base.py
+++ b/napari/layers/base/base.py
@@ -10,14 +10,14 @@ from ._constants import Blending
 
 from ...components import Dims
 from ...utils.event import EmitterGroup, Event
-from ...utils.keybindings import KeymapMixin
+from ...utils.keybindings import KeymapProvider
 from ...utils.misc import ROOT_DIR
 from ...utils.naming import magic_name
 from ...utils.status_messages import status_format, format_float
 from ..transforms import ScaleTranslate
 
 
-class Layer(KeymapMixin, ABC):
+class Layer(KeymapProvider, ABC):
     """Base layer class.
 
     Parameters
--- a/napari/utils/_tests/test_keybindings.py
+++ b/napari/utils/_tests/test_keybindings.py
@@ -1,9 +1,11 @@
 import inspect
+import types
 
 import pytest
 
 from .. import keybindings
 from ..keybindings import (
+    _bind_keymap,
     bind_key,
     components_to_key_combo,
     KeymapHandler,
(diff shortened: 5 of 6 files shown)


napari/_qt/qt_about_keybindings.py
```diff
@@ -74,16 +74,8 @@ def change_layer_type(self, text):
...
         # Add class and instance viewer keybindings
-        text += get_keybindings_summary(self.viewer.class_keymap, col=col)
-        text += get_keybindings_summary(self.viewer.keymap, col=col)
-
-        layer = self.viewer.active_layer
-        if layer is not None:
-            # Add class and instance layer keybindings for the active layer
-            text += get_keybindings_summary(layer.class_keymap, col=col)
-            text += get_keybindings_summary(layer.keymap, col=col)
+        text = get_keybindings_summary(self.viewer.active_keymap, col=col)
```
sofroniewn, line 125: side note - mainly for me to keep track of - I should have used `color` instead of `col` for the keyword argument here, as `col` is definitely confusing. No need to fix in this PR though

sofroniewn: side note - mainly for me to keep track of - I should have used `color` instead of `col` for the keyword argument here, as `col` is definitely confusing. No need to fix in this PR though

napari/_qt/qt_viewer.py
```diff
@@ -377,39 +374,19 @@ def on_key_press(self, event):
...
-
-        if inspect.isgeneratorfunction(func):
-            try:
-                next(gen)
-            except StopIteration:  # only one statement
-                pass
-            else:
-                self._key_release_generators[event.key] = gen
+        combo = components_to_key_combo(event.key.name, event.modifiers)
+        self.viewer.press_key(combo)
```
sofroniewn, line 511: I like how much code is removed!

sofroniewn (approved):
This looks great to me, I like how it reduces the code in `qt_viewer` and `ViewerModel` pushing more to the other classes.

I'll let @jni go through and do a more thorough review of those classes though as he might spot things I have not.

Also, I'm curious, if any of this influence how you might think about any of the mousebinding refactor stuff #694 or event standardization #517 even just from an API perspective.

napari/util/keybindings.py
```diff
@@ -234,6 +241,7 @@ def hello_world(viewer):
             # on key release
             viewer.status = 'goodbye world :('
 
+    To create a keymap that will block others, simply ``bind_key(..., ...)```.
```
jni: remove "simply". Also, does this mean that `bind_key(..., ...)` will prevent all keyboard actions? How does one undo this?

napari/util/keybindings.py
```diff
@@ -188,11 +190,16 @@ def bind_key(keymap, key, func=UNDEFINED, *, overwrite=False):
...
         Key combination.
-    func : callable or None
+        ``...`` acts as a wildcard if no key combinations can be matched
+        in the keymap (this will overwrite all key combinations
+        further down the lookup chain).
+    func : callable, None, or ...
         Callable to bind to the key combination.
         If ``None`` is passed, unbind instead.
+        ``...`` acts as a blocker, effectively unbinding the key
+        combination for all keymaps further down the lookup chain.
```
jni: It's unclear to me why one would want to block instead of unbind...?

napari/util/keybindings.py
```diff
@@ -275,98 +284,135 @@ class KeybindingDescriptor:
...
+    def active_keymap(self):
+        """dict: Active keymap, created by resolving the keymap chain."""
+        keymaps = self.keymap_chain.maps
 
-    def __init_subclass__(cls, **kwargs):
-        super().__init_subclass__(**kwargs)
+        for i, keymap in enumerate(keymaps):
+            if Ellipsis in keymap:  # catch-all
+                break
+        keymaps = keymaps[: i + 1]  # trim maps after a catch-all
```
jni: I dislike two things about this expression:

- reuse of the `keymaps` name — better to rename.
- I like to read the `i` returned by `enumerate` as belonging in the scope of that loop. The escape from the loop feels like "implicit" instead of "explicit" counting.

Sorry if this is handwavy... It is. But anyway, here's my alternative suggestion:

```python
# don't set `keymaps =` at the top
catch_all_index = len(self.keymap_chain.maps)
for i, keymap in enumerate(self.keymap_chain.maps):
    if Ellipsis in keymap:
        catch_all_index = i
        break
keymaps = self.keymap_chain.maps[: catch_all_index]
```

Just btw, I had never seen ChainMap, and now I feel like I can't live without it! LOL

napari/util/keybindings.py
```diff
@@ -275,98 +284,135 @@ class KeybindingDescriptor:
...
 
-    @property
-    def keymap(self):
-        """InheritedKeymap : Keymap used for shortcuts.
-        Inherits from ``class_keymap``.
+        for keymap in reversed(keymaps):
+            for key, func in keymap.items():
+                active_keymap[key] = func
+                if func is Ellipsis:  # blocker
+                    del active_keymap[key]
```
jni: Another suggested restructuring, first to use `dict.update` and second to avoid mutating a container while iterating it:

```python
active_keymap = {}
for keymap in reversed(keymaps):
    active_keymap.update(keymap)
keys_to_delete = [
    k for k, func in active_keymap.items() if func is Ellipsis
]
for key in keys_to_delete:
    del active_keymap[key]
```

jni: Actually, another approach that might not be to everyone's tastes, but that I really like:

```python
from functools import reduce

...

active_keymap = reduce(lambda a, b: {**b, **a}, keymaps, {})
active_keymap_final = {
    k: func for k, func in active_keymap.items() if func is not Ellipsis
}
```

napari/util/keybindings.py
```diff
@@ -275,98 +284,135 @@ class KeybindingDescriptor:
...
 
-    bind_key = KeybindingDescriptor(bind_key)
+        Parameters
+        ----------
+        key_combo : str
+            Key combination.
+        """
+        key_combo = normalize_key_combo(key_combo)
+
+        for parent in self.keymap_providers:
```
jni: Why are we using this iteration and checking instead of `self.active_keymap`? If we are concerned with creating the whole thing each time (which we are), this could be solved by caching + cache invalidation...

napari/util/keybindings.py
```diff
@@ -275,98 +284,135 @@ class KeybindingDescriptor:
...
+                    func = parent.class_keymap[combo]
+                    break
+            else:  # no binding in this provider
+                continue
+            break  # there was a binding in this provider; stop search
+        else:  # no bindings in any provider
+            return
+
+        if func is ...:  # blocker
+            return
```
jni: Let's go with `Ellipsis` rather than `...` here. It's more readable.

napari/util/keybindings.py
jni: Oh, I see: we need the parent. Can the parent be somehow bound to the function when the shortcut is created? e.g.:

`keymap[key_combo] = lambda: func(self)`

?

napari/util/keybindings.py
```diff
@@ -275,98 +284,135 @@ class KeybindingDescriptor:
...
+        Parameters
+        ----------
+        key_combo : str
+            Key combination.
+        """
+        key, _ = parse_key_combo(key_combo)
+        try:
+            next(self._key_release_generators[key])  # call function
+        except (KeyError, StopIteration):
+            pass
```
jni: Man, these are pretty clever. =D

napari/util/tests/test_keybindings.py
```diff
@@ -95,61 +104,183 @@ def foo():
     assert kb == dict(A=foo)
 
 
-def test_InheritedKeymap():
-    d = {}
-    k = InheritedKeymap(lambda: d, {'A': 0})
-    assert k['A'] == 0
+def test_keymap_provider():
```
jni: @kne42 could you split this into several tests that test different aspects of the keymap provider? The class definitions could be outside the test, or in a pytest fixture.

napari/util/tests/test_keybindings.py
```diff
@@ -95,61 +104,183 @@ def foo():
...
+    assert foo.A is ...
+
+    # keybinding blocker on class
+    handler.press_key('D')
+    assert not hasattr(foo, 'D')
+
+    # non-overwriting instance keybinding
+    assert not hasattr(foo, 'E')
+    handler.press_key('E')
+    assert foo.E is None
```
jni: For each of these, would you mind expanding the comment to explain the resolution path relative to the object and keymap creation above? e.g. "the keymap for E is defined in the Foo instance definition, which is the first keymap checked, before the class keymap."

jni (commented):
@kne42 sorry that it has taken me so long to review this! It's pretty cool! I've made a few code suggestions, but they are mostly easy. I'm a little confused about the user perspective: what's changed, other than the ellipsis support? Does `examples/custom_keybindings.py` need updating?

Either way, I think it would be super great to have an example that illustrates all of the features enabled by this PR, with lots of comments about how things are working...

napari/util/keybindings.py
kne42: `...` as the key stands in for any keybinding that is not found in that keymap and `...` as the binding prevents all actions for that key

so `bind_key(..., ...)` tells the key handler to ignore all keymaps further up the chain

to undo this, you can `bind_key(..., None)`

napari/util/keybindings.py
kne42: how about `active_keymap = dict(filter(lambda kv: kv[1] is not Ellipsis, keymaps.items()))`?

napari/util/keybindings.py
kne42: > Can the parent be somehow bound to the function when the shortcut is created?

Ohhhh I like this! I think I'll put this in the `keymap_chain` property :)

> If we are concerned with creating the whole thing each time (which we are), this could be solved by caching + cache invalidation...

Hmmm I'm not quite sure what the best way to do this would be. How about I just do the above and we can let someone more experienced work on the optimization?

napari/util/keybindings.py
jni: Eh. I tend to prefer comprehensions to filters unless they're part of a pipe, but I'm ok with that, with one modification:

```python
dict(filter(lambda k, func: func is not Ellipsis, keymaps.items()))
```

But the three closing parentheses make this harder to read than my dict comprehension, I think... But again, I don't mind it. =)

napari/util/keybindings.py
jni: Haha I'm sure you are experienced enough, but yes, "get it running, get it right, get it fast" is a great mantra. =)

jni: but, essentially, the idea would be `self._cached_active_keymap` would be created, and anything that touches the keymap_chain property resets that to `None`. Then the `active_keymap` property first checks to see whether the cache is None, if it isn't, returns it, if it is, it does its magic.

napari/util/keybindings.py
kne42: I wish I could do this but `dict.items` returns a list of tuples. And yeah you're right, it is harder to read, so I'll do the comprehension.

I also just realized instead of the reducing expression, we could use `dict(ChainMap(*keymaps))`, is this any more or less readable? I'm on the fence.

jni: > but dict.items returns a list of tuples

Oh right. 🤦‍♂ Well then I think the comprehension is nicer.

> is this any more or less readable?

Heh, nice. Despite the fun I had coming up with that reduction, I think using the chainmap is probably more readable, at least once you've encountered chainmaps a few times, but I'll let @sofroniewn comment.

sofroniewn: I'm not super familiar with chainmap, so defer to you both, but it's fine with me

jni: @sofroniewn but do you find the `reduce` call understandable or arcane? If you know that `keymaps` is a list of dictionaries?

napari/util/keybindings.py
kne42: if you want to rebind a key for an instance from its class (defaults), you would need to block on the instance keymap

kne42:
@jni this is ready for re-review :)

sofroniewn:
@kne42 looks like flake8 is failing

jni (approved):
🎉

Issue #787, opened 2019-12-12
Console refactoring / improvements / bug fixes


sofroniewn:
## ❓ Questions and Help

Following some work from @shanaxel42 in #784 to better manage our console resources it is worth raising some additional questions around the expected behavior of the console in interactive environments like when napari is launched from IPython or Jupyter.

Right now if launched from IPython the button to open the console is there but pressing it does nothing and the console is not created. I think we should remove the button and stick with the principle that if launching from IPython you do not need a built-in console in napari. How does @jni feel about this?

Right now if launched from Jupyter we actually create a console and sync it with the jupyter notebook, but there is a weird issue where you now have to press shift-enter in the console to get things to execute. This is very confusing I think, and I'm not sure that we should provide a console in this case either. I propose we remove both the console and the console button when launching from jupyter.

Finally when launching from a script that contains two `napari.Viewer()` like calls strange things happen where both consoles get connected to the same kernal, typing in either though only makes output appear in the 2nd created one and the `viewer` variable gets overwritten so only the 2nd create Viewer is referred too. This is very strange, but I'm not sure how to improve it and very open to suggestions

In general there also might be an opportunity for @shanaxel42 to remove some unneeded code from `QtConsole` depending on how we resolve these questions, and potentially come up with a cleaner fix for #618 at the same time

tlambert03:
for what it's worth, I agree with all these things.  no need for icon when launching from ipython or jupyter (or at the very least no shift-enter if possible when launching from jupyter... remove altogether if not possible).  and don't clober `viewer` variable, but do something like create a second variable name if `viewer` is in `globals()` and emit a warning...

sofroniewn:
@jni proposed the following - drop console + button from IPython but keep for Jupyter and try and fix shift-enter.

We should allow a remote console to be attached in all cases including #801, at which point for IPython the button should appear

sofroniewn:
We should also fix #687 which is related to our console

sofroniewn:
There have been a couple places where the console has now come up and I think we want to make a couple changes right away

[...]

PR #794, opened 2019-12-14
sys.excepthook override, and GUI error dialogs

tlambert03:
# Description
This PR is just a loose proposal for discussion, and results from thinking about #616 

With more complicated and open-ended features, such as the "advanced open dialog", the likelihood of errors increases (with, for instance, unexpected user input or parameter permutations).  Currently, all errors would be printed as tracebacks to the console.  But with a GUI application, it's often desirable to present a cleaner, more interpretable error message (as an example, I often hear Fiji beginners confused about the cryptic error console).  The idea here is that if there are predictable errors that a user is likely to commit (e.g. typo in filename, choosing a `channel_axis` with a huge size), then we raise a specific error which will be caught and shown to the user with obvious next steps.  This PR introduces two things

1. A start to internal exception classes with a base `napari.exceptions.NapariError` class
2. A modified `sys.excepthook` that shows a subset of exceptions to the user with a `QMessageBox`

Note that the presentation of errors in the gui is _opt-in_, it will currently only show a `QMessageBox` for certain exception classes (such as `NapariError`; first example in gif below))... otherwise it dumps to the console as usual (second example in gif below).  That behavior is modifiable in `_qt.exceptions. ExceptionHandler`.

Untitled

It's also worth noting that this will only currently work when napari is the main process (for instance, by running `napari` on the command line or using `python -m napari` in development).  Ipython traps all exceptions and does not use `sys.excepthook`.  But I think [these docs](https://ipython.readthedocs.io/en/stable/api/generated/IPython.core.interactiveshell.html#IPython.core.interactiveshell.InteractiveShell.excepthook) could help if we wanted it to show in ipython launched things as well.

## Type of change

# References
https://docs.python.org/3/library/sys.html#sys.excepthook

# How has this been tested?

## Final checklist:


tlambert03:
As a "counter-argument"... thinking specifically about the open dialog in #616... a more elegant (but much harder to implement) approach would be to do something like what is down with browser form validation, where an error message or indicator is shown right at the point of error:

Screen Shot 2019-12-14 at 10 54 40 AM

sofroniewn:
@tlambert03 I love how you're thinking about this. I think this stuff will be really important for the long term vision of a tool useable by non coders and users of a standalone app.

I think approach 1 with the pop-up is worth exploring and may make sense for some things, plugins etc. I also really like approach 2 for these complex dialogs and think in the long run we should do as much of that as possible i.e. - don't let people even hit `ok` when we can know that it will fail, and point them where to fix things right away.

Another comment is - should everything print to the built-in console as opposed to the main console say when launching from a script? Sort of relates to #787 a little.

tlambert03:
I'm reopening this PR.  I moved the instantiation to the `event_loop`, which is a better place for it (ipython has its own `excepthook` and requires different handling anyway).

from discussion in #1090, I wonder if it maybe wasn't clear exactly what this PR does... so here's a little more background:

- this PR has us _own_ the mechanism for handling uncaught exceptions.  That may mean showing an error dialog in the GUI (in a bundled or non-console situation).  Or that may mean dispatching to the standard ipython traceback printer.  That's the context-specific behavioral stuff we all need to discuss next.  The point is, by overriding `sys.excepthook`, we get to dispatch uncaught exceptions to the secondary handler of our choosing.

- it includes an tentative GUI dialog that we might use ... but that is not the main point of the PR.

- perhaps I should have been more explicit about this too: this PR solves the problem of total app crashes when napari is the main process.  We haven't discussed this much, but starting in Qt5.5, the default behavior for PyQt and PySide is to [completely exit the program with `qFatal()` following an uncaught exception](https://www.riverbankcomputing.com/static/Docs/PyQt5/incompatibilities.html#unhandled-python-exceptions).  That's not a "crash" per se... it's how PyQt chooses to handle uncaught exceptions.  It's of course, unexpected for most python users... since normally in python an uncaught exception simply prints a traceback.  (See [this SO post](https://stackoverflow.com/questions/33736819/pyqt-no-error-msg-traceback-on-exit), and [this email thread from Luke Campagnola](https://riverbankcomputing.com/pipermail/pyqt/2017-July/039411.html) for discussion).  This is not so much an issue when running in Ipython, since the [ipython `excepthook`](https://ipython.readthedocs.io/en/stable/api/generated/IPython.core.interactiveshell.html#IPython.core.interactiveshell.InteractiveShell.excepthook) interferes.  And this is why the Qt app _won't_ actually crash if you enter an error into the integrated ipython console, but it will if you introduce an error into a napari script.

tlambert03:
re-requesting review on this one.  probably should have brought this one up at meeting.  I don't think this needs to wait on the docs that @ziyangczi is working on.  and it fixes the urgent need of preventing total crashes when an uncaught exception is raised.

liu-ziyang:
fwiw, I tested this change with the windows bug of toggle volume: https://github.com/napari/napari/issues/1347 reproduced on nd_labels.py as before on a windows machine, the hard crashing is indeed avoided, however the applicataion ends up in "Not Responding" state without much information raised anywhere, I am seeing some warning message on the console: 
```
WARNING: Error drawing visual <vispy.visuals.mesh.MeshVisual object at 0x000002442ED4A820>
```
that seems related to the error

I am approving the changes regardless because this does seem to handle python errors well, so we will get more peace of mind of these leaked exceptions from underlying code, however, we need to be aware it is yet bulletproof on weird edge cases like the vispy issue above, which could be due to c code bugs

liu-ziyang (approved):

napari/_qt/exceptions.py
```diff
@@ -0,0 +1,107 @@
...
+            self.setStyleSheet(
+                """QTextEdit{
+                    min-width: 800px;
+                    font-size: 12px;
+                    font-weight: 400;
+                }"""
+            )
+
+
+class ExceptionHandler(QObject):
```
liu-ziyang, line 58: I realized that NapariError is removed from the PR, is it intended? if we are to wrap sys exception handler, I would think it does make sense to introduce our wrapper error, thoughts?

tlambert03: I see that as the last bullet point in https://github.com/napari/napari/issues/1395#issuecomment-657702637 ... and yes, I think that since the functionality here doesn't depend on that concept, and would be best left to a followup (as it implies a whole different discussion about what different error classes might imply for the downstream handling).

napari/_qt/exceptions.py
```diff
@@ -0,0 +1,107 @@
...
+            The traceback object associated with the error.
+        """
+        # etype.__module__ contains the module raising the error
+        # Custom exception classes can have different behavior
+        # can add custom exception handlers here ...
+        self.error.emit((etype, value, tb))
+
+        # NOTE: we don't *have* to always show the dialog... we can put
+        # additional logic here to dispatch to secondary handlers.
+        self._show_error_dialog((etype, value, tb))
```
liu-ziyang, line 97: @tlambert03 if you don't mind shall we comment out or remove this for now? though I do like this a lot, we may want to pat down it and add this in a later PR

tlambert03:
> however the applicataion ends up in "Not Responding" state without much information raised anywhere, ...., we need to be aware it is yet bulletproof on weird edge cases like the vispy issue above, which could be due to c code bugs

yep, there's a very important distinction to make here.  This is not solving the underlying problem that bugs "exist" (as mentioned before, I don't imagine that we will _ever_ solve that problem)... It is not bulletproofing anything with regards to _avoiding_ exceptions or inconsistent states.  It is simply equipping us with some machinery to handle them, and report them.  In this particular example of the vispy error, the "solution" in my opinion would be to either fix the vispy error itself, _or_ at least identify what causes the error and raise a hard exception _before_ that error occurs that would cede flow to the exception handler.  But again, none of that has to do with _how_ we handle it (which is this PR)

napari/_qt/exceptions.py
tlambert03: yep, see bullet point 2 in https://github.com/napari/napari/issues/1395#issuecomment-657702637:

> The end goal of #794 should only be to make sure that we never crash back to the command line. And the “default behavior” (until we hook up a GUI error message) should be to simply print the traceback to the console… effectively recovering what the “default python behavior” is for almost every other (non Qt) program... Later PRs should think about implementing a better version of the _show_error_dialog function in there

liu-ziyang:
opened up another PR since I can't update this branch: https://github.com/napari/napari/pull/1476

Issue #735, opened 2019-11-25
Allow napari to launch in foreground (when called from Jupyter lab)


s-m-e:
## 🚀 Feature

A simple option when calling `napari.view_image`: `foreground`, a bool. If `True`, the `napari` viewer window receives the focus automatically when being opened.

## Motivation

At least when launched from Jupyter, `napari` opens its windows in the background. I can easily switch to them using my desktop's task manager / task bar, but this quickly becomes annoying with many open tasks / windows, virtual desktops and multiple physical screens. When I call `napari`, I want to directly see it in foreground, interact with it and close it, getting back to where I came from.

## Pitch

Less looking for windows, quicker workflow.

## Alternatives

None.

## Additional context

None.

tlambert03:
hi @s-m-e 
that behavior was actually changed in #721 and merged into master in 84e05db (not yet released).

however, it came at the cost of a reversion of #553... causing the menubar to be inactive when launching _outside_ of a jupyter notebook.  See ongoing discussion in #732.  We're not immediately sure how to reconcile the two problems... but regardless of which way the main branch goes with it, if you want to fix it manually in your notebook, you can do it like this:
```python

with napari.gui_qt():
    viewer = napari.view_image(data)
    viewer.window._qt_window.raise_()
```

sofroniewn:
As noted above this was fixed in #721 and preserved by #795 which fixed some related menubar problems. I am closing now

PR #795, opened 2019-12-15
Fix window raise & inactive menubar conflict

### LINKED ISSUE #732: window raise / menubar focus at conflict with each other
## 🐛 Bug

So turns out #721 which "raised" the main window as actually a reversion of #553 which had deleted that line to fix a menubar focus issue.

In #553 I made the following statement

> I'm not quite sure why the raise_ call was in there or if there are any negative issues of dropping it on other platforms, but I can confirm that on a mac everything still functions as expected, including usage from the notebook.

Turns out `raise_` was needed to bring the window to the front which is quite nice.

Now I'm not sure how to resolve both #553 and #721 at the same time!!!
@tlambert03 do you have any ideas?

tlambert03:
# Description
I think this should fix #732 (see also #721 and #553 ).  [This](https://github.com/napari/napari/issues/732#issuecomment-565765389) comment and [this](https://github.com/napari/napari/issues/732#issuecomment-565766439) comment detail some of the observations that led to this fix.  

The summary is that we want to call `Window._qt_window.raise_()` in pretty much every case _except_ when instantiating a viewer within a `gui_qt()` context manager for the first time within the Qt app's lifecycle.  It looks a bit strange but by naming any newly instantiated `QApplication`s within `gui_qt()` and checking whether `_qt_window.isActiveWindow()` during `Window.show()`... we seem to be able to figure out whether `raise_` should be called or not.

@sofroniewn, see if you can break it!

## Type of change

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/_qt/qt_main_window.py
+++ b/napari/_qt/qt_main_window.py
@@ -274,16 +274,18 @@ class Window:
         """Resize, show, and bring forward the window."""
         self._qt_window.resize(self._qt_window.layout().sizeHint())
         self._qt_window.show()
-        # make sure window is not hidden, e.g. by browser window in Jupyter
-        # if the application was instantiated in gui_qt(), app_name will be
-        # "napari".  If it is the second time a _qt_window has been created,
-        # then isActiveWindow() will be True (otherwise false).
-        # this is the only combination that seems to fix #732
+
+        # We want to call Window._qt_window.raise_() in every case *except*
+        # when instantiating a viewer within a gui_qt() context for the
+        # _first_ time within the Qt app's lifecycle.
+        #
+        # `app_name` will be "napari" iff the application was instantiated in
+        # gui_qt(). isActiveWindow() will be True if it is the second time a
+        # _qt_window has been created. See #732
         app_name = QApplication.instance().applicationName()
-        if app_name == 'napari' and not self._qt_window.isActiveWindow():
-            return
-        self._qt_window.raise_()  # for MacOS
-        self._qt_window.activateWindow()  # for Windows
+        if app_name != 'napari' or self._qt_window.isActiveWindow():
+            self._qt_window.raise_()  # for macOS
+            self._qt_window.activateWindow()  # for Windows
 
     def _update_palette(self, palette):
         # set window styles which don't use the primary stylesheet
--- a/napari/_qt/qt_main_window.py
+++ b/napari/_qt/qt_main_window.py
@@ -27,6 +27,7 @@ from qtpy.QtWidgets import (  # noqa: E402
     QAction,
     QShortcut,
     QStatusBar,
+    QApplication,
 )
 from qtpy.QtCore import Qt  # noqa: E402
 from qtpy.QtGui import QKeySequence  # noqa: E402
@@ -270,12 +271,19 @@ class Window:
         self._qt_window.resize(width, height)
 
     def show(self):
-        """Resize, show, and bring forward the window.
-        """
+        """Resize, show, and bring forward the window."""
         self._qt_window.resize(self._qt_window.layout().sizeHint())
         self._qt_window.show()
         # make sure window is not hidden, e.g. by browser window in Jupyter
-        self._qt_window.raise_()
+        # if the application was instantiated in gui_qt(), app_name will be
+        # "napari".  If it is the second time a _qt_window has been created,
+        # then isActiveWindow() will be True (otherwise false).
+        # this is the only combination that seems to fix #732
+        app_name = QApplication.instance().applicationName()
+        if app_name == 'napari' and not self._qt_window.isActiveWindow():
+            return
+        self._qt_window.raise_()  # for MacOS
+        self._qt_window.activateWindow()  # for Windows
 
     def _update_palette(self, palette):
         # set window styles which don't use the primary stylesheet
--- a/napari/_qt/event_loop.py
+++ b/napari/_qt/event_loop.py
@@ -22,7 +22,12 @@ def gui_qt(*, startup_logo=False):
     IPython with the Qt GUI event loop enabled by default by using
     ``ipython --gui=qt``.
     """
-    app = QApplication.instance() or QApplication(sys.argv)
+    app = QApplication.instance()
+    if not app:
+        # if this is the first time the Qt app is being instantiated, we set
+        # the name, so that we know whether to raise_ in Window.show()
+        app = QApplication(sys.argv)
+        app.setApplicationName('napari')
     if startup_logo:
         logopath = join(dirname(__file__), '..', 'resources', 'logo.png')
         splash_widget = QSplashScreen(QPixmap(logopath).scaled(400, 400))


sofroniewn (approved):
Whoo! I think you did it @tlambert03!! This works for me in all cases on my mac and I couldn't break it. I will let others give it a try and weigh in on the implementation before merge, but I think it's great. Thanks so much for pushing on this one :-)

napari/_qt/qt_main_window.py
```diff
@@ -270,12 +271,19 @@ def resize(self, width, height):
...
-        self._qt_window.raise_()
+        # if the application was instantiated in gui_qt(), app_name will be
+        # "napari".  If it is the second time a _qt_window has been created,
+        # then isActiveWindow() will be True (otherwise false).
+        # this is the only combination that seems to fix #732
+        app_name = QApplication.instance().applicationName()
+        if app_name == 'napari' and not self._qt_window.isActiveWindow():
+            return
+        self._qt_window.raise_()  # for MacOS
+        self._qt_window.activateWindow()  # for Windows
```
jni: (a) can you clarify a bit when, in *usage* terms, the window will be brought forward? It's unclear to me.

(b) I prefer logic that executes code when passing, rather than logic that bails out of the function (kinda like a double negative). So I suggest:

```python
# raise when the window is launched from Jupyter or...
if app_name != 'napari' or self._qt_window.isActiveWindow():
    self._qt_window.raise_()  # for macOS
    self._qt_window.activateWindow()  # for Windows
```

jni (commented):
@tlambert03 thank you very much for this investigation... It's a *big* deal to have this behaviour nailed down!!! 🎉

napari/_qt/qt_main_window.py
tlambert03: (let me know if you wanted this in comments, as opposed to here).
There is a lot of mystery still going on here around the Qt windowing system and how it plays with the OS, and I really can't say that I understand why this works, but I think the best i know how to put it in terms of usage is as I put it above:
> `raise_()` will be called in pretty much every case except when instantiating a viewer within a `gui_qt()` context manager for the first time within the Qt app's lifecycle.

In terms of the user experience, the aim of this PR  is to never have a newly instantiated viewer window appear below any other window (i.e. below the terminal when i IPython or the browser when in a jupyter notebook).  In every case, calling `raise_()` accomplishes that; however when you `raise_` a `main_window` _immediately_ after `QApplication()` instantiation (such as in the `gui_qt()` context), for whatever reason, the menubar becomes inactive.  It's almost like the menubar belongs to a different OS process.  I feel like this is some sort of Qt race condition bug (given the prevalence of the observation around the internet).

So after some guess & check, the logic I settled on here just avoids calling `raise_` whenever the Qt App was created in the `gui_qt()` context (where we give the `QApplication().instance()` the name 'napari') and when it is the first time a viewer is being instantiated that session (in which case `_qt_window.isActiveWindow()` will be `False`).  _why_ is it false the second+ time around?  that was the guess and check part!

sorry I don't have a better explanation! :/

jni: @tlambert03 

> is as I put it above

This is precisely how I would like it to appear in the comment! It's useful to know what the *intent* of the if-clause is, rather than just the mechanics of it.

I'm not positive that this is always the best behaviour — I often launch a viewer and want to immediately type new things to add to it — but I'm happy to make this the default for now, and see what people think.

tlambert03: that's a fair point. terminal windows are usually small(ish), so it's quite obvious where the napari window is, and one might not want to lose terminal focus.  in jupyter however, it's not uncommon for the napari window to show up _behind_ the browser window completely, and the user wouldn't know it... which is where the inspiration for `raise_` came from in the first place. (see for example this user request https://github.com/napari/napari/issues/735)

I think the name of the `QApplication().instance()` _might_ be different in those cases... would you want me to set it to _not_ raise in ipython? (fwiw, the current behavior for the last month or so has been to always raise)

jni: > would you want me to set it to not raise in ipython

I have a slight preference for this, but I wouldn't stress too much about it.

> fwiw, the current behavior for the last month or so has been to always raise

My daily workhorse is Linux + i3, which is a tiling window manager, so I don't really notice these things that much. =P But I do notice them when I switch to my mac for whatever reason!

sofroniewn: I lean towards how it will be after this PR, which will be a significant improvement for me. I'd keep the IPython and Jupyter the same for now, but we can revisit in follow-up PRs if it becomes clear that they should be treated differently

tlambert03: I think I'm with @sofroniewn, I have a slight preference for raising the viewer when `napari.Viewer()` is called.  But since this seems kinda like something a user preference that might change in different circumstances, maybe we should just add a "raise" parameter to `Window.__init__`?

Currently, it looks like while the main `Window` class [does have a `show=True` parameter](https://github.com/napari/napari/blob/fffd2f2a4090f773f2aec904bee9ac28a4b1890d/napari/_qt/qt_main_window.py#L54), it's sort of impossible to access, since most people will be instantiating the window indirectly in `napari.Viewer()` [here](https://github.com/napari/napari/blob/fffd2f2a4090f773f2aec904bee9ac28a4b1890d/napari/viewer.py#L62).  Should we consider exposing some of those options a little more?  To capture all of the subtleties, we'd really need to expose both `show` and `raise_` ... since someone might want to show but not raise (like @jni here).

sofroniewn: hmm agreed the `show` parameter doesn't do much right now without being exposed on our top level `Viewer` / `view_*` and agreed we'd need to expose a `raise` too. I'm open to exposing, but I'm also not sure its needed (i.e. I'm not sure if anyone right now wants to use `show=False`). My preference would be to merge this PR now as is and then consider expose those additional keyword args in followup PRs

tlambert03: yeah, I agree probably no need for `show=False`, and I'm fine to merge as is...
but maybe @jni wants to weigh in on whether he's ok with having `raise=True` as the default behavior, or if he wants a `raise=False` option

jni: @tlambert03 I'm totally fine with the behaviour, but my two comments above (add the usage comment from the PR, and change the logic flow) haven't been addressed yet!

sofroniewn: Ok great @jni! I'll wait for @tlambert03 to address those comments and then we can merge. Thanks for weighing in

tlambert03: addressed the comments and changed the logic.  All set to merge once tests pass.

sofroniewn:
Thanks @tlambert03 tests have passed, I will merge this now.

sofroniewn:
@tlambert03 so turns out this is no longer working for me. I should have tested before merge, but when I run a script from my mac the menubar doesn't work. I'll start investigating now

sofroniewn:
So, the logic is correct and `raise_` is not getting called, but things still aren't working on my mac. This is actually a new laptop, so I guess something else may have changed for me ... seems like it is independent of this PR though :-(

jni:
> This is actually a new laptop

Ooh, did you get a shiny 16"? =P

> seems like it is independent of this PR though :-(

This is terrible! Let me try locally...

sofroniewn:
image

tlambert03:
fascinating and infuriating :(
it's broken menubars you're referring to right?  is it still only in the `gui_qt()` context that they don't work?

jni:
Ah, that's nowhere near as exciting. =P

So, it's all working fine for me on Mojave... =\

jni:
I vote that we ban @sofroniewn's computers from testing from now on... =P

tlambert03:
hah.  Well here's the good news for @sofroniewn ... I can confirm the same broken menubars on _my_ shiny new 16"! ... which happens to be my only computer on Catalina.  So it will at least be easier for me to tinker with it, for another PR

sofroniewn:
Yes, only problem @tlambert03 is menubar + `gui_qt` everything else works fine

PR #796, opened 2019-12-15
Revert #784 console shutdown conditionals

tlambert03:
# Description
I left a comment on https://github.com/napari/napari/pull/784#discussion_r357945390, but since it's a closed PR, I'll just a new one so it doesn't get missed.  As stated there, I'm a bit confused why the conditionals were removed in the console shutdown method (even after reading the conversation there).  I'm getting exceptions now every time I close the napari window (in ipython).  If I'm missing something, we can just delete or modify this PR.

## Type of change

- [x] Bug-fix (non-breaking change which fixes a regression)

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/_qt/qt_console.py
+++ b/napari/_qt/qt_console.py
@@ -94,5 +94,7 @@ class QtConsole(RichJupyterWidget):
         # self.execute_on_complete_input = True
 
     def shutdown(self):
-        self.kernel_client.stop_channels()
-        self.kernel_manager.shutdown_kernel()
+        if self.kernel_client is not None:
+            self.kernel_client.stop_channels()
+        if self.kernel_manager is not None:
+            self.kernel_manager.shutdown_kernel()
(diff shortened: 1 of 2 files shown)


sofroniewn:
I'm approving, but will let the others weigh in too before merging. I think there might have been a time where the PR had a more extensive refactor and so this wasn't needed. When we look into #787 we also might want to clean up.

@tlambert03 is there a way we can add a test for this? I know it might be tricky given the usage of IPython, but it might be good to keep problems like this from slipping through the cracks.

sofroniewn (approved):

shanaxel42:
You're right this check should still be there. I had initially removed them thinking there should never be a time when `self.kernel` is not set, but that's not true in the ipython or terminal start scenario. My apologies!

tlambert03:
no problem!  Just wanted to make sure I wasn't missing something.  I'll see if I can create a reasonably useful test.

tlambert03:
added a test mocking an iPython terminal, just to make sure it creates and shuts down a `QtConsole`

sofroniewn (commented):
That test looks great @tlambert03, thanks for adding!

jni:
Yeah, as mentioned there, I just have no understanding why I couldn't actually get it to crash when reviewing the PR... Maybe something funky with my IPython config? Anyway, thank you so much Talley for the quick fix and for the test!

PR #797, opened 2019-12-15
Add blocked_signals context manager

tlambert03:
# Description
Tiny little PR here that adds a `blocked_signals` context manager.  There are a few places (and probably will be more) where we temporarily block qt signals using:
```python
obj.blockSignals(True)
# do something
obj.blockSignals(False)
```
this turns that into
```python 
with blocked_signals(obj):
    # do something
```

(also, it's in #675 and I'm trying to break that down into smaller bits)

## Type of change

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/_qt/layers/qt_image_base_layer.py
+++ b/napari/_qt/layers/qt_image_base_layer.py
@@ -1,9 +1,10 @@
-from qtpy.QtWidgets import QSlider
-from .. import QHRangeSlider
-from .qt_base_layer import QtLayerControls
 from qtpy.QtCore import Qt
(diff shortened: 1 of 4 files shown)


sofroniewn (approved):
LGTM, on minor requested change - can this be renamed `qt_blocked_signals` or maybe `qt_signals_blocked` just to make it clear that is different from our own vispy based event system.

tlambert03:
of course.  Funny, I actually had it named `blocked_qt_signals` and removed the qt.  So back it goes!  I like `qt_signals_blocked`

PR #800, opened 2019-12-16
Fix (minor) deprecations

tlambert03:
# Description
this just fixes a couple deprecation warnings that show up in our tests.  perhaps most importantly `from collections import Iterable` will stop working in python3.8

@sofroniewn, the comparison in this line is deprecated ... but it wasn't immediately clear to me what the goal of the logic is:
https://github.com/napari/napari/blob/4096e71b7e1fa041a62f4ac2f6853fba60c93e52/napari/layers/base/base.py#L613
 
```
napari/layers/image/tests/test_image.py::test_negative_rgba_image
  /Users/talley/Dropbox (HMS)/Python/forks/napari/napari/layers/base/base.py:614
  DeprecationWarning: elementwise comparison failed; this will raise an error in the future.
    if value is not None and not np.all(value == (None, None)):
```

is it just making sure that, if `value` is an array, at least one value is not null?

### DIFF
--- a/napari/layers/base/base.py
+++ b/napari/layers/base/base.py
@@ -610,6 +610,7 @@ class Layer(KeymapMixin, ABC):
 
         value = self._value
 
+        # FIXME: deprecation warning... but logic unclear
         if value is not None and not np.all(value == (None, None)):
             msg += ': '
             if type(value) == tuple:
--- a/napari/layers/base/base.py
+++ b/napari/layers/base/base.py
@@ -609,18 +609,15 @@ class Layer(KeymapMixin, ABC):
         msg = f'{self.name} {full_coord}'
 
         value = self._value
-
-        # FIXME: deprecation warning... but logic unclear
-        if value is not None and not np.all(value == (None, None)):
-            msg += ': '
-            if type(value) == tuple:
-                msg += status_format(value[0])
+        if value is not None:
+            if isinstance(value, tuple) and value != (None, None):
+                # it's a pyramid -> value == (data_level, value)
+                msg += f': {status_format(value[0])}'
                 if value[1] is not None:
-                    msg += ', '
-                    msg += status_format(value[1])
+                    msg += f', {status_format(value[1])}'
             else:
-                msg += status_format(value)
-
+                # it's either a grayscale or rgb image (scalar or list)
+                msg += f': {status_format(value)}'
         return msg
 
     def to_xml_list(self):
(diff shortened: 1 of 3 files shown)


napari/layers/base/base.py
```diff
@@ -610,6 +610,7 @@ def get_message(self):
 
         value = self._value
 
+        # FIXME: deprecation warning... but logic unclear
```
sofroniewn: yeah, this isn't so nice. Some layers return value as a singleton, hence the `is not None`, but some layers like `Shapes` can return a tuple, I think the following would work

```python
if value is not None and not np.all([v == None for v in value])
```

sofroniewn (commented):
Thanks for doing this @tlambert03. I left a suggestion on that strange line. Maybe we can fix this before merge. I'm surprised there aren't more warnings, or does this only fix a subset. Would be good to have them all fixed at some point

napari/layers/base/base.py
tlambert03: actually, looks like `value` can be: `None`, a scalar, a `tuple`, and even a`tuple` with a nested numpy `array` (for rgb_pyramid).   `for v in value` will throw an error if `value` is not iterable, so that won't work.  Do we just want to check whether it's a non-null scalar, or an iterable with no null values?  (or an iterable with at least one non-null value?)

sofroniewn: that sounds reasonable for now, then we might want to refactor the layers so the output of value is a little more constrained

jni (approved):

jni:
I'm happy to merge as is and then constrain the possible return types of `value`. Even in the suggested fix, the logic is hard to follow... Additionally, that whole machinery might change when we add support for world coordinates...

tlambert03:
I tweaked the logic of `get_message` very slightly.  I verified that the behavior is still the same with RGB, grayscale, and RGB images.  Hopefully it's a little bit clearer, and the deprecation warning is gone.  If you are ok with it, feel free to merge.

tlambert03:
not sure what's [going on with black](https://github.com/napari/napari/pull/800/checks?check_run_id=351350474)... closing and reopening to force rebuild

kne42:
@tlambert03 are you able to force a rebuild by clicking on "details" next to the check?

tlambert03:
yeah, but neither that method nor the other methods are fixing it.  almost seems like a problem with the python install on cirrus?

kne42:
yeah, that's bizarre. the versions of the installed packages haven't changed either

tlambert03:
looks like it finally succeeded at around 6:05 EST...  🤷‍♂️merging!

sofroniewn:
Nice, I triggered a manual re-run at that time, and I guess it had been fixed by our CI :-)

PR #801, opened 2019-12-16
[WIP] connect to remote kernel in new console

sofroniewn:
# Description
This PR adds the ability to connect to a second remote kernel in our console space. In this demo I have already launched a remote IPython on ec2, copied the relevant connection file, and created an ssh tunnel according to the [steps here](https://github.com/ipython/ipython/wiki/Cookbook:-Connecting-to-a-remote-kernel-via-ssh).

When I do `viewer.window.connect_remote_console(path_to_connection_file)` it will then create a link to the new kernel and I can start interacting with it from the viewer. Note though that this kernel doesn't have access to the viewer and we still need to explore things like #654 to make that connection. You can now only execute remote code on that kernel. I thought that this was a relatively self contained piece of work though and will make working on things like #654 much easier.

There are still a bunch of problems here, some around the Qt styling and how the show / hide console buttons work, and then some inside the QtConsole file, which likely needs to be refactored anyway, see #787. I wanted to get this going though so we could at least take this use case into account during that process.

I'm also probably not managing any of these resources properly and will need @shanaxel42 to help make sure they get shut down in the right way. There are also no tests for this right now either. 

At a high level though do people think this is a feature we want / direction we want to go in? @tlambert03 @jni? 

Here is a gif of the two consoles:
connect_remote_console

## Type of change

## Final checklist:

### DIFF
--- a/napari/_qt/qt_console.py
+++ b/napari/_qt/qt_console.py
@@ -26,11 +26,14 @@ class QtConsole(RichJupyterWidget):
         Shell for the kernel if it exists, None otherwise.
     """
 
-    def __init__(self, user_variables=None):
+    def __init__(self, user_variables=None, *, connection_file=None):
         super().__init__()
 
         # get current running instance or create new instance
-        shell = get_ipython()
+        if connection_file is None:
+            shell = get_ipython()
+        else:
+            shell = 'connection_file'
 
         if shell is None:
             # If there is no currently running instance create an in-process
@@ -80,6 +83,15 @@ class QtConsole(RichJupyterWidget):
             self.kernel_client = kernel_client
             self.shell = shell
             self.push = self.shell.push
+        elif shell == 'connection_file':
+            kernel_client = QtKernelClient(connection_file=connection_file)
+            kernel_client.load_connection_file()
+            kernel_client.start_channels()
+
+            self.kernel_manager = None
+            self.kernel_client = kernel_client
+            self.shell = shell
+            self.push = lambda var: None  # self.shell.push
         else:
             raise ValueError(
                 'ipython shell not recognized; ' f'got {type(shell)}'
--- a/napari/_qt/qt_main_window.py
+++ b/napari/_qt/qt_main_window.py
@@ -319,6 +319,13 @@ class Window:
         img = self._qt_window.grab().toImage()
         return QImg2array(img)
 
+    def connect_remote_console(self, connection_file):
+        self.qt_viewer.make_remote_console(connection_file)
+        self._add_viewer_dock_widget(self.qt_viewer.dockRemoteConsole)
+        self._qt_window.tabifyDockWidget(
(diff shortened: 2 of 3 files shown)


tlambert03:
I'm probably not being imaginative enough... but what is the use case here?  When would you want/need to connect to a remote kernel (that does not have access to the viewer) from within napari, instead of just using a regular terminal?

sofroniewn:
@tlambert03 in the long run I want the remote kernel and the viewer to interact through something like #654 and thought that this would be a small self contained PR to review and get in before continuing work on #654. Although one could always have a separate IPython running in a regular terminal I think it will be much nicer to have the integrated console for those like like integrated consoles in general.

I also want to make `connect_remote_console` directly accept the json dictionary provided by jupyter magic `%connect_info` (rather than just the path to the saved connection_file). Then someone could be inside a remote jupyter notebook, run `%connect_info` and copy and paste the dictionary into a napari that was running on their local machine and be connected to that kernel inside our console.

In many ways this is the easy part, the hard part will then be "syncing" the viewer state / data (whatever that will mean), but I thought if we know we want to support this type of embedded remote console at some point we might as well get it in now, especially as we think about things like #787. How does this sound?

jni:
@sofroniewn honestly, I'm not in love with this. For now, it's a bunch of untested code without a clear use case. "It will make #654 easier" is a bit too vague, and I think I would rather see this as a separate commit in #654 where it is needed, rather than as a bit of "orphan" functionality...

It's unclear to me also that we will want to implement remote viewing as two kernels, rather than doing some magic to make the remote kernel "feel" local. (ie only have a single console rather than tabs.)

sofroniewn:
@jni it's fair that this may be a bit premature, though it is still WIP and tests etc. were going to be added (though it wasn't 100% clear how we'd ever test the full functionality within our CI).

The use case is connecting to remote kernels to pass data back to the viewer which I was going to explore using dask futures directly and can be done without needing to have a full "remote napari" option (i.e. independent of #654). Clearly I can do this from a non integrated console too, but I thought I might prefer working in an integrated console.

Regardless I mainly wanted to see how easy this was (answer - pretty easy) and get a feel for what two consoles was like vs one. I had no problem with two, but also agree that we still have to figure out the harder part of how we'll actually do the syncing, and depending on how it is done whether we want two consoles or one (or more?!).

I also thought it would be good to understand this use case before refactoring #787, but we can let it sit for now, and maybe we revisit when we make more progress with figuring out #654 and some of the harder questions still ahead of us.

jni:
> we can let it sit for now

as a skimage core dev, I have grown quite comfortable with long-lived PRs. =P

sofroniewn:
I think I'm going to close this now that we merged #1055. Even if we decided to allow multiple consoles in the future we'd probably want to start with a fresh PR, though there might be ideas here that could be lifted out.

PR #782, opened 2019-12-10
Improve multiple colors behavior in relevant layers 

HagaiHargil:
# Description

This PR is intended to correct napari's handling of Points, Shapes, and other relevant layers containing multiple colors, e.g. a unique color per point in the layer. It should also allow a user to request a transparent element.

My intended API for multiple colors is detailed inside `colors_data.py`, which serves as the input data for a new testing file. As you can see, the goal is to allow for maximum flexibility on the user side, by allowing them to provide a simple string, a hexadecimal triplet, `vispy` objects, lists, tuples and arrays. However, napari's internal representation should always be a `vispy.color.ColorArray`. 

This last demand isn't satisfied yet, since it requires larger changes than what you guys perhaps intended to make when thinking of a solution to #774. In the current state, the point colors are stored in a Python list of objects, which I think can be simplified to either an array or (preferably) a `ColorArray`. Technically this PR can be completed without changing the internal color representation, but I'd argue against it.

What I currently did is to modify the color stored in the layer (e.g. `Points.edge_colors`) with the new `transform_color` function just before sending it down to vispy. I do believe however that this transformation should occur upon instantiation of the `Points` layer itself, as I detailed above.

Another peculiarity I faced is seen (here)[https://github.com/napari/napari/compare/master...PBLab:colorarray?expand=1#diff-10d45528bccb3636e9d4fdf94884b128R58], namely the need for the `max_len` value, the purpose of which I'm not sure of. I currently left it there for "backwards compatibility", but I'm not quite sure what it does.

Lastly, the tests I wrote currently all fail due to the internal representation issue. When we decide what to do about it I'll change them accordingly.

## Type of change

- [x] Bug-fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

# References
Addresses both #774 and #762. Still WIP, but I need your input on a few topics. I didn't touch upon the "transparent" color for points. This will be done before this PR is merged.

# How has this been tested?
Tests are in `test_vispy_set_colors`. If we change more things then I'll add new ones and adjust the current ones accordingly.

## Final checklist:
- [ ] My PR is the minimum possible work for the desired functionality
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #814, opened 2019-12-19
Create independent sys_info function; add to CLI

tlambert03:
# Description
probably should have proposed this with the recent updates to the `napari info` window ... but this PR does a couple very minor things for fetching napari environment info:

- moves the system info text (and `citation_text`) out of the `QtAbout` object to `napari.utils.info.sys_info`, also accessible at `napari.sys_info()`
- adds `--version` and `--info` flags to the command line arguments

    ```bash
    $ napari --version
    napari version 0.2.6+42.g0721798c

    $ napari --info
    napari: 0.2.6+42.g0721798c
    Platform: Darwin-19.0.0-x86_64-i386-64bit
    Python: 3.7.5 (default, Oct 25 2019, 10:52:18)  [Clang 4.0.1 (tags/RELEASE_401/final)]
    PySide2: 5.13.2
    Qt: 5.13.2
    VisPy: 0.6.3
    NumPy: 1.17.4
    SciPy: 1.3.3
    scikit-image: 0.16.2
    Dask: 2.9.0

    GL version:  '2.1 ATI-3.2.32'
    MAX_TEXTURE_SIZE: 16384
    ```
- modifies the size of the `QtAbout` window slightly to remove scrollbars (it's not that much more text):
Screen Shot 2019-12-18 at 10 12 29 PM

## Type of change


sofroniewn (approved):
Love this PR! So nice to be able to tell users who are struggling to get napari running to just do `napari--version` and `napari --info`. Will those run even if `napari` won't launch? I.e. let's say there is a Qt bug, or the person doesn't have Qt installed it would be great to make sure that they both still run.

Maybe a minor addition, can you add `napari --citation` to print the citation text too?

tlambert03:
good point.  It would not be hard to make sure that it can run no matter what.  To make it bombproof, we could even put every import in `napari.utils.info`  in a try/catch so that we're at guaranteed to get at least some info.  I'll play with that, add `--citation`, and update

sofroniewn:
I like the idea of using `Try / catch` and delivering an Import Failed in the message we print

tlambert03:
well, this is actually a lot harder than I thought.  I updated the `sys_info()` function to work regardless of the environment, but it's actually not all that useful.  Problem is: if you run napari as `__main__`, as is done from the CLI, it first has to successfully import the entirety of `__init__` before running the contents of `__main__`... and `__init__` basically imports all the dependencies.  (So if someone is missing a dependency, we won't even get to the `sys_info` function to tell them so before an `ImportError` or `ModuleNotFoundError` is raised).

I can think of some workarounds, but they all require mucking with `__init__` in a way that might lead to some "unconventional" behavior... which is not worth the very minor gain (since the current `ModuleNotFoundError` is already pretty self-explanatory)

We can leave the harmless try/catches in the `sys_info` function if you want, or we can just revert to a previous commit here since they make the function a bit less readable for no (current) reason.

napari/__main__.py
```diff
@@ -6,13 +6,31 @@
...
-from .utils import io
-from . import Viewer, gui_qt
+from . import Viewer, __version__, gui_qt
+from .utils import io, sys_info
+
+
+class InfoAction(argparse.Action):
+    def __call__(self, *args, **kwargs):
+        print(sys_info())
+        sys.exit()
```
jni, line 16: Is there an advantage to making this a class instead of a function?

napari/utils/info.py
```diff
@@ -0,0 +1,66 @@
+import platform
+import sys
+
+import napari
+
+
+def sys_info(as_html=False):
+
```
jni, line 16: Add a docstring?

jni (commented):
@tlambert03 

> this is actually a lot harder than I thought

That was exactly my reaction when I first read "It would not be hard to..." 😂

I actually like the code as it is, it's not too complicated, and it opens the door to creating a completely *separate* entrypoint to setup.py, e.g. `napari-info` (`naparinfo`? 😜) that does bypass all our normal imports. It just needs a docstring. 😉 

`as_html` is very nice. 👌

napari/__main__.py
tlambert03: don't really "want" it to be a class, but looked into it briefly and actually not sure what the best way to run a function and quit would be without using the [`argparse.Action` protocol](https://docs.python.org/3/library/argparse.html#argparse.Action), which only described doing it with a subclass.  Do you have an alternate strategy?

tlambert03:
>  it opens the door to creating a completely _separate_ entrypoint to setup.py, e.g. `napari-info` that does bypass all our normal imports.

would I accomplish this by having a new top-level `napari_info` folder that had the `sys_info` function somewhere inside of it, with a new `console_scripts` entry in setup.py?  And if so, does that mean I would then have to `from napari_info import sys_info` to use it within `QtAbout`?

tlambert03:
well, I tried what I said above:  introduced a new top-level package `napari_info` and put the `sys_info` function in there.  Now running `napari-info` on the command line _should_ give some useful information even if napari itself can't be imported.  I'm not 100% sure the functionality is worth the whole new module... but you can have a look at the changes, and we can easily revert to a6903c2.

napari/utils/info.py
```diff
@@ -0,0 +1,71 @@
+import platform
+
+import sys
+
+import napari
+
+
+def sys_info(as_html=False):
```
sofroniewn, line 8: is this just being left in here for comparison purposes and will get deleted if we go with the `napari_info` approach?

tlambert03: ah, that goes if we go with `napari_info`

napari/__main__.py
```diff
@@ -6,13 +6,32 @@
...
 def main():
     parser = argparse.ArgumentParser(usage=__doc__)
     parser.add_argument('images', nargs='*', help='Images to view.')
+    parser.add_argument(
+        '-v',
+        '--version',
+        action='version',
+        version=f'napari version {__version__}',
+    )
+    parser.add_argument(
```
sofroniewn, line 34: I think I'd leave in the `napari --info` syntax here too - I prefer it to `napari-info` and see `napari-info` as more of a last resort for people that have install fails and document it as such

napari/utils/info.py
sofroniewn: yup, makes sense

napari/__main__.py
jni: @tlambert03 I think this works?

```python
def info_action(*args, **kwargs):
    print(sys_info())
    sys.exit()

...

    action=info_action)
```

napari/__main__.py
jni: I agree with this sentiment very much.

jni:
@tlambert03 regarding your question about auto-complete on zulip: if you have commands `napari` and `napari-info` in your path, then typing `nap<tab>` will autocomplete to `napari`, but with only `napari`, `nap<tab>` will autocomplete to `napari ` (space afterwards), allowing you to go straight to the next argument. It's a small difference, but personally I find it an irritating one when it gets me! Finger memory is very powerful.

tlambert03:
Let’s just rename it then.  `_napari_info` or `info_napari` or something.  No one really needs to access it directly, it will also be available at `napari.sys_info`.  Is that the only thing you dislike?  Or is the value If a separate folder still debatable?

sofroniewn:
‘_napari_info’ works for me if we also have ‘napari --info‘, but I do wonder if it is even worth it and justifies the additional code complexity. I leave that to you @tlambert03 and @jni to decide - I am fine with just dropping it too. Maybe it is nice though, I’m just imagining trying to debug virtually with someone that can’t launch anything, I don’t know, maybe it wouldn’t help ...

tlambert03:
> ‘_napari_info’ works for me if we also have ‘napari --info‘, but I do wonder if it is even worth it and justifies the additional code complexity. I leave that to you @tlambert03 and @jni to decide - I am fine with just dropping it too. Maybe it is nice though, I’m just imagining trying to debug virtually with someone that can’t launch anything, I don’t know, maybe it wouldn’t help ...

I think it's possible that it would be useful, but probably in a limited number of scenarios. In most cases, if someone is completely lacking a dependency, the `ModuleNotFoundError` is going to be sufficient to debug.  And as long as everything at least imports, we can throw it all inside of the `napari` folder (in which case `napari --info` is more just a quick way to get at some version and GL information ...).  And even then, `conda list` or `pip freeze` is likely to get us most of that (but with a lot of extra crap).  I think I'll leave the call to @jni.

sofroniewn:
Ok yeah, good points @tlambert03. I think it was a nice idea at the time, but we should drop in favor of simpler code organization. I’ll leave the final call with @jni though

jni:
I agree with

> it was a nice idea at the time, but we should drop in favor of simpler code organization

If `pip install napari` has somehow failed, which is the only way that I can imagine missing dependencies can really happen, then even `_napari_info` will probably fail. So I think the benefit will be extremely limited, for a large jump in code complexity.

tlambert03:
Agreed. Will clean up and ping again soon

napari/__main__.py
tlambert03: actually... this appears to work at first, but I think `info_action` _always_ gets called.  presumably, the `action` argument of `parser.add_argument` is expecting a class, not a function, and will call it to instantiate it (only later calling `instance.__call__` if the `--flag` is `True`) so giving it a function just calls the function immediately and exits the program regardless of other arguments...  I tried some other function strategies, but couldn't get it to work.   Will leave it as a class unless you have other ideas.

tlambert03:
ok, everything is back in the `napari` module.  all set for one more review

napari/utils/info.py
sofroniewn, line 8: Can you add a simple test for this, maybe just that it is callable both with `as_html=True` and without and returns a string in both cases

napari/utils/info.py
```diff
@@ -0,0 +1,71 @@
...
+        text += f'<br>{sys_info_text}'
+
+    if not as_html:
+        text = (
+            text.replace("<br>", "\n").replace("<b>", "").replace("</b>", "")
+        )
+    return text
+
+
+citation_text = (
```
sofroniewn, line 73: can you maybe add a test where you import this and just check it is a string. would catch any weird slipups

sofroniewn:
This looks great @tlambert03, I requested a couple really simple tests, but then I think it's good to go. I also checked an confirmed all functionality works as expected on my machine

sofroniewn:
Looks like this needs the numpy fixes from #827. Do we also need the fixes from https://github.com/napari/napari/pull/782/commits/f825cd4aa8da19b86ff584803bf73613c8fea362 and https://github.com/napari/napari/pull/782/commits/b1b69210d459db94573aaa0d589192577469cf9a inside `test_interpolate_coordinates`, which I don't think were in #827?

I think I'd like to merge this and then make our `0.2.8` release

sofroniewn:
Ok, I think this is now ready for merge. Windows fails look unrelated to this PR

PR #825, opened 2019-12-21
Basic remote control of napari

rpanderson:
# Description
Provides access to methods and attributes of viewer using ZeroMQ.

This might be useful where you want to programmatically add images/layers/shapes from a separate process rather than via the napari console or the interactive session that napari was launched with. This will allow existing software to interface with napari in a way that doesn't require you to launch napari from that software (or vice versa) or on the same host.

Start napari with a ZMQ server using:
`napari --remote-port 7341`

Then, from a separate process, you can call:
```python
In [1]: from napari.remote import Client

In [2]: client = Client()

In [3]: client.request('version')
Out[3]: '0.2.7rc1+0.g04b5ae5.dirty'

In [4]: client.request('title')
Out[4]: 'napari'

In [5]: client.request('status')
Out[5]: 'Ready'

In [6]: client.request('add_image', path=r'https://github.com/napari/napari/raw/master/napari/resources/logo.png')
Out[6]: True

In[7]: data = client.request('get_data', layer=0)
Out[7]:
array([[[255, 255, 255,   0],
        [255, 255, 255,   0],
        [255, 255, 255,   0],
        ...,
        [255, 255, 255,   0],
        [255, 255, 255,   0],
        [255, 255, 255,   0]]], dtype=uint8)

In [8]: client.reset_view()
Out[8]: True
```
Requests are handled in the main thread and the Qt lock is acquired using the `inmain_decorator` of the [qtutils package](https://qtutils.readthedocs.io/).

In this proof-of-concept, methods of the `napari.viewer.Viewer` singleton are accessed via `client.request('method_name', *args, **kwargs)`. The `get_data`, `close`, and `hello`  requests are handled separately, with [their own methods](https://github.com/rpanderson/napari/blob/remote-access/napari/remote.py#L40) of the server (prefixed by `handle_`).

## Type of change

## TODO
- Document and support the use of a shared secret file for use with [zprocess](https://github.com/chrisjbillington/zprocess) security mechanisms.

### DIFF
--- a/examples/remote_access.py
+++ b/examples/remote_access.py
@@ -1,0 +1,26 @@
+"""
+Example of making calls to the viewer from a separate script or process
+using a ZMQ server started with the --remote-port flag.
+
+Run this example after having started napari on localhost with:
+   napari --remote-port 7341
+"""
+
+from napari.remote import Client
+
+client = Client()
+print(
+    f"Started client to napari remote server on {client.host} with port {client.port}"
+)
+
+version = client.request('version')
+print(f"napari version: {version}")
+
+print("Adding image to napari")
+client.request(
+    'add_image',
+    path=r'https://github.com/napari/napari/raw/master/napari/resources/logo.png',
+)
+
+print("Getting data from layer 0")
+data = client.request('get_data', layer=0)
--- a/napari/__main__.py
+++ b/napari/__main__.py
@@ -49,9 +49,15 @@ def main():
         'dask if all the images fit in RAM. This option does nothing if '
         'only a single image is given.',
     )
+    parser.add_argument(
+        '-p',
+        '--remote-port',
+        help='Start a ZMQSserver listening on specified port for remote procedure calls.',
+        default=None,
+    )
     args = parser.parse_args()
     with gui_qt(startup_logo=True):
-        v = Viewer()
+        v = Viewer(port=args.remote_port)
         if len(args.images) > 0:
             images = io.magic_imread(
                 args.images, use_dask=args.use_dask, stack=not args.layers
--- a/napari/components/viewer_model.py
+++ b/napari/components/viewer_model.py
@@ -12,6 +12,10 @@ from ..utils.keybindings import KeymapMixin
 from ..utils.theme import palettes
 from ..utils.misc import ensure_iterable, is_iterable
 from ..utils import io
+from .._version import get_versions
+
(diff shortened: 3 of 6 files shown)


sofroniewn:
@rpanderson this is super exciting, and so cool that it seems to have come out of the hackday that @jni organized :-) Better support for remote interactions with napari is definitely going to be a big priority area for us for 2020, and I'd be curious to get your take on some slightly higher level questions before going through the details of this PR - which I know is still in draft stage.

We've got a bunch of open issues and in-progress PRs around remote stuff right now which I've tried to put together under the [remote and sharable napari](https://github.com/napari/napari/projects/9) github project, and I've added this one there too, feel free to weigh in on any of them as appropriate, most are super in progress.

I do think though that we have some known use-cases that we really want to support, like someone who is using a remote jupyter notebook but now wants to look at images in napari while continuing to do interactive analysis in the remote notebook. Because of our use of Qt this is not immediately possible. For that use case and we've been considering using things like dask pub/sub to synchronize viewer states - see #654, where one viewer is on the remote machine and "headless", so no Qt stuff, just our model files, but has access to all the big data and compute, and the viewer is a normal one running locally on your computer which has all the Qt, vispy, and rendering stuff. Once nice aspect of using dask here is they have nice machinery for sharing big arrays that works well with our lazy-compute approach for that we can make sure we're just piping the minimal array slice across when we need to, they also have some nice caching approaches built in. I'm curious how you see your approach with ZeroMQ relating to this data sharing with dask approach, and what the main use case you have for this PR is.

I'm also curious how you feel about and api that on the remote ("headless") machine would allow you to still use the same napari apis but would not trigger remote calls - for example you'd instantiate something like `viewer =Viewer(remote=True, port=7341, etc.)` and then do `viewer.add_image(data)` and that would trigger the data to get sent as a dask array and added to the local machine where it would be visible with a napari viewer actually running there and that was waiting for such calls. It's slightly different then the `client.request` style API, though I think you could build an API like that on top of `client.request` style calls.

All in all though very cool see, and I'm sure there will be a lot of people who want better remote interaction stuff that will be excited by all this!!

jni:
@rpanderson this PR is sadly now out of date with napari's main branch. Please get in touch if you want to push this feature forward, would be happy to pair up again to try to get it merged! 😊 In the meantime I've created #3553 as a feature request to track this idea.

Issue #540, opened 2019-09-18
Better support for surface timeseries data


sofroniewn:
## 🚀 Feature
Right now the surface layer, created in #503 can support multiple dimensional data by providing a `NxD` array of vertices, a list of triangles, and a length `N` value for each vertex. This allows for very general surfaces that change over any of the `D` dimensions.

There is potentially a very special case where we have timeseries of either a 2D or 3D surface and that surface remains fixed, but the values of the vertices change in time. Right now that could be represented as `Nx3` or `Nx4` arrays of vertices and a length `N` vector where N is now very large and there is a lot of redundant structure in those lists of vertices. A potentially more compact representation might be `Mx2` or `Mx3` representation of the surface and then an `MxT` representation of the vertex values over T timepoints. This representation could be reshaped into the other one if needed, but it could be nice to support it more directly.

I'm curious what the standard file formats, say from neuroimaging of surface timeseries measurements provide, as that could give us guidance on what to support.

sofroniewn:
This has been closed by #831 which adds better support for surface timeseries

PR #831, opened 2019-12-26
support surface timeseries

### LINKED ISSUE #829: Better support for surface layer timeseries
## 🚀 Feature
As discussed in this [image.sc post](https://forum.image.sc/t/slider-widget-to-control-surface-values/32482/6) it would be nice to provide better support for people with timeseries values of static surfaces. It would be quite easy to support an API where people could pass data in the form
```
vertices N x 3 (constant: doesn’t change with time)
faces M x 3 (constant: doesn’t change with time)
values L x N (each time has N values for the vertices )
```
and we could take care of the rest. Note that vertices could still be `N x D`, and values could now be `D'` dimensional and we'd just convert all the `D'-1` dimensions into additional slider like dimensions. Should be quite clean and conform to users expectations much more strongly.

I can work on this when I get the chance, shouldn't be too bad

sofroniewn:
# Description
This PR will close #829 by allowing timeseries values for surfaces (i.e. static surface, but values that vary in one (say time) or more dimensions. This code snippet will allow you to reproduce the example bellow

```python
# NKI resting state data from nilearn
from nilearn import datasets
from nilearn import surface
import napari
import numpy as np


nki_dataset = datasets.fetch_surf_nki_enhanced(n_subjects=1)

# Fsaverage5 surface template
fsaverage = datasets.fetch_surf_fsaverage()

# Load surface data and resting state time series from nilearn
pial_left = surface.load_surf_data(fsaverage['pial_left'])
sulc_left = surface.load_surf_data(fsaverage['sulc_left'])
timeseries = surface.load_surf_data(nki_dataset['func_left'][0])
timeseries = np.stack([timeseries, timeseries/2, timeseries/4, np.full(timeseries.shape, -1.5)], axis=2)
print(timeseries.shape)

with napari.gui_qt():
    # create an empty viewer

    # add the mri
    viewer.add_surface((pial_left[0], pial_left[1], sulc_left), name='base')
    viewer.add_surface((pial_left[0], pial_left[1], timeseries), colormap='red', opacity=0.9, name='timeseries')
```

surface_timeseries

Note some counterintuitive things may happen if you now try and "display" a timeseries like dimension, but it shouldn't crash. There might be some weird corner cases I havn't caught yet though.

Right now the change is backwards compatible with our current shapes API, this PR purely adds functionality. There was some discussion of changing the API in #829 which I am open too, but I think that can be separated from this PR, though I'm open to doing it here too!

## Type of change

# How has this been tested?

- [x] tests added inside the surface layer tests

## Final checklist:

### DIFF
--- a/examples/surface_timeseries.py
+++ b/examples/surface_timeseries.py
@@ -6,8 +6,9 @@ try:
     from nilearn import datasets
     from nilearn import surface
 except ImportError:
-    raise ImportError("""This example uses data and methods from nilearn but
-    nilearn installed. To install try 'pip install nilearn'.""")
+    raise ImportError(
+        "You must have nilearn installed to run this example."
+    )
 
 import numpy as np
 import napari
@@ -18,16 +19,19 @@ nki_dataset = datasets.fetch_surf_nki_enhanced(n_subjects=1)
 fsaverage = datasets.fetch_surf_fsaverage()
 
 # Load surface data and resting state time series from nilearn
-pial_left = surface.load_surf_data(fsaverage['pial_left'])
-sulc_left = surface.load_surf_data(fsaverage['sulc_left'])
-timeseries = surface.load_surf_data(nki_dataset['func_left'][0]).transpose((1, 0))
+brain_vertices, brain_faces = surface.load_surf_data(fsaverage['pial_left'])
+brain_vertex_depth = surface.load_surf_data(fsaverage['sulc_left'])
+timeseries = surface.load_surf_data(nki_dataset['func_left'][0])
+# nilearn provides data as n_vertices x n_timepoints, but napari requires the
+# vertices axis to be placed last to match NumPy broadcasting rules
+timeseries = timeseries.transpose((1, 0))
 
 with napari.gui_qt():
     # create an empty viewer
     viewer = napari.Viewer(ndisplay=3)
 
     # add the mri
-    viewer.add_surface((pial_left[0], pial_left[1], sulc_left), name='base')
-    viewer.add_surface((pial_left[0], pial_left[1], timeseries),
+    viewer.add_surface((brain_vertices, brain_faces, brain_vertex_depth), name='base')
+    viewer.add_surface((brain_vertices, brain_faces, timeseries),
                        colormap='turbo', opacity=0.9,
                        contrast_limits=[-1.5, 3.5], name='timeseries')
--- a/examples/surface_timeseries.py
+++ b/examples/surface_timeseries.py
@@ -1,0 +1,33 @@
+"""
+Display a surface timeseries using data from nilearn
+"""
+
+try:
+    from nilearn import datasets
+    from nilearn import surface
+except ImportError:
+    raise ImportError("""This example uses data and methods from nilearn but
+    nilearn installed. To install try 'pip install nilearn'.""")
+
+import numpy as np
+import napari
+
+
+# Fetch datasets - this will download dataset if datasets are not found
+nki_dataset = datasets.fetch_surf_nki_enhanced(n_subjects=1)
+fsaverage = datasets.fetch_surf_fsaverage()
+
+# Load surface data and resting state time series from nilearn
+pial_left = surface.load_surf_data(fsaverage['pial_left'])
+sulc_left = surface.load_surf_data(fsaverage['sulc_left'])
+timeseries = surface.load_surf_data(nki_dataset['func_left'][0]).transpose((1, 0))
+
+with napari.gui_qt():
+    # create an empty viewer
+    viewer = napari.Viewer(ndisplay=3)
+
+    # add the mri
+    viewer.add_surface((pial_left[0], pial_left[1], sulc_left), name='base')
+    viewer.add_surface((pial_left[0], pial_left[1], timeseries),
+                       colormap='turbo', opacity=0.9,
+                       contrast_limits=[-1.5, 3.5], name='timeseries')
--- a/napari/components/viewer_model.py
+++ b/napari/components/viewer_model.py
@@ -869,7 +869,7 @@ class ViewerModel(KeymapMixin):
         data : 3-tuple of array
             The first element of the tuple is an (N, D) array of vertices of
             mesh triangles. The second is an (M, 3) array of int of indices
-            of the mesh triangles. The third element is the (N, K0, ..., KL)
+            of the mesh triangles. The third element is the (K0, ..., KL, N)
             array of values used to color vertices where the additional L
             dimensions are used to color the same mesh with different values.
         colormap : str, vispy.Color.Colormap, tuple, dict
--- a/napari/components/viewer_model.py
+++ b/napari/components/viewer_model.py
@@ -869,8 +869,9 @@ class ViewerModel(KeymapMixin):
         data : 3-tuple of array
             The first element of the tuple is an (N, D) array of vertices of
             mesh triangles. The second is an (M, 3) array of int of indices
-            of the mesh triangles. The third element is the (N, ) array of
-            values used to color vertices.
+            of the mesh triangles. The third element is the (N, K0, ..., KL)
+            array of values used to color vertices where the additional L
+            dimensions are used to color the same mesh with different values.
         colormap : str, vispy.Color.Colormap, tuple, dict
             Colormap to use for luminance images. If a string must be the name
             of a supported colormap from vispy or matplotlib. If a tuple the
--- a/napari/layers/surface/surface.py
+++ b/napari/layers/surface/surface.py
@@ -18,8 +18,9 @@ class Surface(Layer):
     data : 3-tuple of array
         The first element of the tuple is an (N, D) array of vertices of
         mesh triangles. The second is an (M, 3) array of int of indices
-        of the mesh triangles. The third element is the (N, ) array of
-        values used to color vertices.
+        of the mesh triangles. The third element is the (N, K0, ..., KL)
+        array of values used to color vertices where the additional L
+        dimensions are used to color the same mesh with different values.
     colormap : str, vispy.Color.Colormap, tuple, dict
         Colormap to use for luminance images. If a string must be the name
         of a supported colormap from vispy or matplotlib. If a tuple the
@@ -55,13 +56,14 @@ class Surface(Layer):
     data : 3-tuple of array
         The first element of the tuple is an (N, D) array of vertices of
         mesh triangles. The second is an (M, 3) array of int of indices
-        of the mesh triangles. The third element is the (N, ) array of
-        values used to color vertices.
+        of the mesh triangles. The third element is the (N, K0, ..., KL)
+        array of values used to color vertices where the additional L
+        dimensions are used to color the same mesh with different values.
     vertices : (N, D) array
         Vertices of mesh triangles.
     faces : (M, 3) array of int
         Indices of mesh triangles.
-    vertex_values : (N,) array
+    vertex_values : (N, K0, ..., KL) array
         Values used to color vertices.
     colormap : str, vispy.Color.Colormap, tuple, dict
         Colormap to use for luminance images. If a string must be the name
@@ -143,6 +145,7 @@ class Surface(Layer):
         # Data containing vectors in the currently viewed slice
         self._data_view = np.zeros((0, self.dims.ndisplay))
         self._view_faces = np.zeros((0, 3))
+        self._view_vertex_values = []
 
         # assign mesh data and establish default behavior
         self._vertices = data[0]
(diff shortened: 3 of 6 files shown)


sofroniewn:
This will also close #540 a duplicate of #829

napari/components/viewer_model.py
```diff
@@ -869,8 +869,9 @@ def add_surface(
         data : 3-tuple of array
             The first element of the tuple is an (N, D) array of vertices of
             mesh triangles. The second is an (M, 3) array of int of indices
-            of the mesh triangles. The third element is the (N, ) array of
-            values used to color vertices.
+            of the mesh triangles. The third element is the (N, K0, ..., KL)
+            array of values used to color vertices where the additional L
+            dimensions are used to color the same mesh with different values.
```
jni, line 874: Oh, I think this shape should be (K0, ..., KL, N)... This would mirror NumPy broadcasting (implicit dimensions are prepended, not appended).

napari/layers/surface/surface.py
```diff
@@ -286,10 +293,29 @@ def _set_view_slice(self):
         disp = list(self.dims.displayed)
         indices = np.array(self.dims.indices)
 
+        # Take vertex_values dimensionality into account.
+        # Note that displayed dimensions corresponding to vertex_values will be
+        # ignored from the display and averaged over to get the used values
```
jni: I don't understand this comment, nor the code underneath it... Would you mind adding more comments to explain further what this is doing? I imagine "adj" means "adjusted", but adjusted for what? What averaging is going on?

jni (changes requested):
@sofroniewn I have two suggestions:

- I think the time (etc) dimensions should be prepended, not appended, to match NumPy indexing more closely. (It also means that a batch of values will be a contiguous array, which could help performance.)
- It would be great if you could put the example code from your gif in the examples folder, perhaps with some conditional download + unzip code if the file is not found, or at least a comment + exit if the file is not found, pointing to where the dataset can be downloaded.

napari/components/viewer_model.py
sofroniewn: changed

napari/layers/surface/surface.py
sofroniewn: I've added more comments, hope it helps, if not I can add more. Agreed it's a little confusing here

sofroniewn:
I have added an example which depends on `nilearn`. Similar to our `xarray` example it using a `try` / `catch` with an `ImportError` to give a helpful warning if `nilearn` isn't installed. The `nilearn` package takes care of the downloading / loading of the dataset. Curious if @tlambert03 and @jni think this is the right approach?

I've also made the minor fixes requested by @jni including adding more comments and switching the order of the timeseries data to be `(K, N)` instead of `(N, K)` say, if there are `N` vertices.

Here's a gif of the example data playing at 40fps - pretty cool I think!!

surface_timeseries2

examples/surface_timeseries.py
```diff
@@ -0,0 +1,33 @@
+"""
+Display a surface timeseries using data from nilearn
+"""
+
+try:
+    from nilearn import datasets
+    from nilearn import surface
+except ImportError:
+    raise ImportError("""This example uses data and methods from nilearn but
+    nilearn installed. To install try 'pip install nilearn'.""")
```
jni:
```suggestion
    nilearn is not installed.""")
```

(a) the string formatting is a bit weird. I think with my suggestion you might be able to fit it in its own line with:

```python
raise ImportError(
    'This example uses ... but nilearn is not installed.'
)
```

(b) I don't want to prescribe installation methods here. Currently mixing pip and conda is discouraged, so, I prefer that users use their best judgment/google how to install nilearn.

examples/surface_timeseries.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+# Fetch datasets - this will download dataset if datasets are not found
+nki_dataset = datasets.fetch_surf_nki_enhanced(n_subjects=1)
+fsaverage = datasets.fetch_surf_fsaverage()
+
+# Load surface data and resting state time series from nilearn
+pial_left = surface.load_surf_data(fsaverage['pial_left'])
+sulc_left = surface.load_surf_data(fsaverage['sulc_left'])
+timeseries = surface.load_surf_data(nki_dataset['func_left'][0]).transpose((1, 0))
```
jni: I *slightly* would prefer that we transpose when adding the data, so that users can see in the call itself that our format is transposed from nilearn's. Or, at least, put the transpose in its own line. You could also add a comment on top of the line with the transpose, "nilearn provides data as n_vertices x n_timepoints (x n_trials?), but napari requires the vertices axis to be placed last to match NumPy broadcasting rules."

examples/surface_timeseries.py
```diff
@@ -0,0 +1,33 @@
...
+import numpy as np
+import napari
+
+
+# Fetch datasets - this will download dataset if datasets are not found
+nki_dataset = datasets.fetch_surf_nki_enhanced(n_subjects=1)
+fsaverage = datasets.fetch_surf_fsaverage()
+
+# Load surface data and resting state time series from nilearn
+pial_left = surface.load_surf_data(fsaverage['pial_left'])
```
jni: Maybe unpack this?

```python
pial_left_vertices, pial_left_triangles = surface.load_surf_data(fsaverage['pial_left'])
```

jni: (with corresponding changes below)

napari/layers/surface/surface.py
```diff
@@ -209,6 +212,10 @@ def _get_extent(self):
             maxs = np.max(self.vertices, axis=0)
             mins = np.min(self.vertices, axis=0)
 
+        if self.vertex_values.ndim > 1:
+            mins = [0] * (self.vertex_values.ndim - 1) + list(mins)
+            maxs = list(self.vertex_values.shape[:-1]) + list(maxs)
```
jni, line 220: Can you add a comment about why this is necessary?

napari/layers/surface/surface.py
```diff
@@ -282,14 +289,37 @@ def gamma(self, value):
...
 
-        not_disp = list(self.dims.not_displayed)
-        disp = list(self.dims.displayed)
-        indices = np.array(self.dims.indices)
+        # Take vertex_values dimensionality into account if more than one value
+        # is provided per vertex.
+        if self.vertex_values.ndim > 1:
+            values = self.vertex_values[
+                self.dims.indices[: -self.vertices.shape[1]]
+            ]
```
jni: Ok so it took me quite a while to understand this, and I think it's key to the whole process. If I understand correctly, the dims model has L + 3 dimensions. (or L + D, rather). So now we want to grab the current values, so we get the index for axes [0, 1, ..., -D]. I suggest a comment to this effect!

napari/layers/surface/surface.py
```diff
@@ -282,14 +289,37 @@ def gamma(self, value):
...
+            if values.ndim > 1:
+                # If not all dimensions corresponding to the vertex_values
+                # are being sliced through, average over any remaining ones
+                # so that there is always only value per vertex.
+                values = values.mean(axis=tuple(range(0, values.ndim - 1)))
```
jni: This seems very brittle and undocumented. Should we not raise an error in this case?

napari/layers/surface/surface.py
```diff
@@ -282,14 +289,37 @@ def gamma(self, value):
...
+            self._view_vertex_values = values
+            # Determine which axes of the vertices data are being displayed
+            # and not displayed, but ignoring the additional dimensions
+            # corresponding to the vertex_values.
+            indices = np.array(self.dims.indices[-self.vertices.shape[1] :])
```
jni: Again, the `.shape[i]` syntax is a bit confusing. My suggestion is to replace these values with letters at the top of the function: `N, D = self.vertices.shape`. (or similar) Then use those names rather than indexing into a shape.

napari/layers/surface/surface.py
```diff
@@ -282,14 +289,37 @@ def gamma(self, value):
...
+            additional_ndim = self.vertex_values.ndim - 1
```
jni: maybe I'd call this values_ndim?

napari/layers/surface/surface.py
```diff
@@ -282,14 +289,37 @@ def gamma(self, value):
...
+            disp = np.subtract(self.dims.displayed, additional_ndim)
+            disp = [d for d in disp if d >= 0]
```
jni: I'd combine these two into one expression:

```python
disp = [
    d for d in np.subtract(self.dims.displayed, additional_ndim) if d >= 0
]
```

(or similar)

napari/layers/surface/surface.py
```diff
@@ -282,14 +289,37 @@ def gamma(self, value):
...
+            self._view_vertex_values = self.vertex_values
+            indices = np.array(self.dims.indices)
+            not_disp = list(self.dims.not_displayed)
+            disp = list(self.dims.displayed)
 
         self._data_view = self.vertices[:, disp]
         if len(self.vertices) == 0:
             self._view_faces = np.zeros((0, 3))
-        elif self.ndim > self.dims.ndisplay:
+        elif self.vertices.shape[1] > self.dims.ndisplay:
```
jni: As above: give a name to this quantity. spatial_ndim?

jni (commented):
@sofroniewn I've left a few more code suggestions, but this is basically ready imho! 🎉

examples/surface_timeseries.py
sofroniewn: have improved formatting and wording. have dropped pip reference

examples/surface_timeseries.py
sofroniewn: done

examples/surface_timeseries.py
sofroniewn: changes made

napari/layers/surface/surface.py
sofroniewn: comment added

napari/layers/surface/surface.py
sofroniewn: added

napari/layers/surface/surface.py
sofroniewn: hmm, i'd prefer not to error here, but agree it's a bit strange. Can we leave for now?

napari/layers/surface/surface.py
sofroniewn: done

napari/layers/surface/surface.py
sofroniewn: done

napari/layers/surface/surface.py
sofroniewn: done

napari/layers/surface/surface.py
sofroniewn: done

sofroniewn:
@jni I've made all the changes requested apart from raising an error if there values object is not 1-D. I can, or I can try and think of something better to do if you think it is important. also if you want more comments / better variable names keep asking. i know the logic here is intuitive in places / might be hard to debug in the future, so I'm happy to do more to make it more clear

napari/layers/surface/surface.py
jni: non-breaking, non-documented, deep-in-the-code behaviour tends to never be "for now". Unless we can come up with a reasonable default behaviour here, in my opinion, this should be an error: you are trying to display multiple values as a single one. For all you know, these could be measurements in different units! Without a clear use example, it just seems wrong to do this. An error with a *clear* message of how users could view their data would be perfectly acceptable imho: it's a simple `.mean()` call outside of napari, and outside napari is where it should be: we shouldn't do statistical processing for users.

What I *might* live with is a *warning* that the data is being munged and, maybe, grabbing only the 0th element among the remaining axes.

But my preference is still to error out. I don't see how anything else would be correct in the majority of cases. And, actually, how is it even possible to land in this bit of code?

jni:
@sofroniewn other than the `.mean` behaviour, which I still think should go, this is ready to go imho! `.mean` isn't even tested as far as I can tell?

napari/layers/surface/surface.py
sofroniewn: Ok, these are good points, and I am changing the current behavior. You can land in the code by changing which dimensions are displayed vs non-displayed (i.e. by clicking the roll axes button in the nilearn example).

Raising an error wasn't so great as the old data from the layer would stay on the screen, I now emit a warning and make all the data disappear, which I think is the right call as it doesn't really make sense to even grab the `0th` axis or something in this case. It really means nothing should be shown.

sofroniewn:
I have removed the `.mean` behavior, now emitting a warning and making no data visible. I've also added a test for this new behavior.

This should be ready to go. Thanks for the helpful review @jni, I think the PR is much better now

jni (approved):

jni:
@sofroniewn great! Yes, I guess if you are changing the displayed dimensions, then the surfaces themselves should disappear? Minus some small, 2D, possibly 1D or even 0D slice? Which, yeah, doesn't make much sense to me... Possibly we should have a multicoloured (multivalued) line orthogonal to the surface slice? But who knows. Certainly it would be of limited use at best!

Thanks for all your patient changes! 😊

jni:
I love that we have this in! 🎉

royerloic:
Love it...

Issue #762, opened 2019-12-05
Easier way to make points with no face color


henrypinkard:
## 🚀 Feature

Minor improvement: I'm using napari to visualize results of a cell finding algorithm. Would like to use points layer to draw circles around detections. My instinct was to try something like this:

```viewer.add_points(yx, edge_color='magenta', face_color=None)```

But this throws an error. As far as I can tell, the way to accomplish this is to use the ring marker, but this doesn't seem the best way to outline circular objects if you also want to provide shape information. Thoughts?

**Comment by kevinyamauchi** [comment]
Hey @henrypinkard ! I think this should be possible. The `Points` layer wraps the vispy [`MarkersVisual`](https://github.com/vispy/vispy/blob/master/vispy/visuals/markers.py#L506-L586). The `face_color` can be set with a [`ColorArray`](https://github.com/vispy/vispy/blob/5fa85e0e9e639ce4282d04718fcd4c17de4d8e8d/vispy/color/color_array.py#L83) object, which has a alpha argument. We can make it such that when the `Points` layer `face_color` is set to `None`, the face color alpha is set to 0.  I _think_, this will make the face of the marker transparent. What do you think?

With regards to the ring marker, can you expand on what you mean by "doesn't seem the best way to outline circular objects if you also want to provide shape information"? You can pass a size to change the radius (can either by a single value that gets broadcasted to all markers in the layer or an array with sizes for each point).

```python
viewer.add_points(yx, edge_color='magenta', symbol='ring', size=size_array)
```

henrypinkard:
Yeah sounds great!

Should have been a bit more clear on the second point. I just meant that if you want to 1) circle an object, 2) see the pixels within the circle, and 3) have the radius of the circle represent the object size, it is hard to do with the ring because by expanding the size in order to accomplish (2), you mess up (3) because both the inner and outer diameter enlarge simultaneously

kevinyamauchi:
Got it, that makes sense. I have another issue I’m working on, but I can get to this next!

sofroniewn:
This would be great. I'd prefer we use the explicit string `'transparent'` instead of the python variable `None` as we sometime use `None` to trigger a guess or a use another default setting - which is maybe not the best practice, but it would lead to ambiguity here

jni:
[...]

PR #832, opened 2019-12-27
clickable color swatch with QColorDialog

tlambert03:
# Description
This PR turns the color swatch (in, e.g., the points layer controls) into a clickable widget that raises a color picker dialog, (as proposed by @sofroniewn [here](https://github.com/napari/napari/issues/762#issuecomment-563048345)).  I would propose that we remove the QComboBox color picker altogether and either have just the clickable swatch, or a LineEdit that can convert arbitrary input into our internal color representation as in #782.  This PR is just a stub/example for now, and will wait for #782 to merge before final integration (if the general GUI idea is approved here).

ezgif-2-107ce933206f

## Type of change

# How has this been tested?

## Final checklist:


sofroniewn:
Dropping the `QComboBox` sounds good. I'm open to either just having the swatch like Photoshop / Illustrator does, or the `QLineEdit` which we can type into / will show the string representation after we change colors.

Waiting till after #782 makes sense too

kevinyamauchi:
This is cool! I am becoming more and more convinced that @tlambert03 is a wizard.

I agree that we should drop the `QComboBox`. Outside of the standard red, green, blue, etc., I don't think the string representations (e.g., hex codes or `chartruesse` 😛 ) are super informative and I think most people will prefer to pick visually.

sofroniewn:
Ok yeah - maybe we just have the swatch, especially if the string name is visible somewhere inside the color dialog (even if not though, probably still ok not to have the QLineEdit - keeps things simpler and cleaner) but I defer to the preference of @tlambert03.

And yes, another amazing addition!!!

jni:
(a) this is super brilliant
(b) I still like the ability to pick colours by name somehow, which might be useful where exact reproducibility is needed. But perhaps it's fine to leave such functionality to the console.
(c) this gave me the idea that we could possibly make the colormap picker equally visual, with a grid of gradients. 🤯

sofroniewn:
It doesn't look like this PR has conflicts with #686 which had the conversions like `edge_color `-> `current_edge_color` and `edge_colors` -> `edge_color` but just wanted to tag here in case

tlambert03:
Thanks to @HagaiHargil's work in #782, polishing up this PR was made much easier.  I updated it with a generalized `QColorSwatchEdit` widget, which combines a colorswatch that raises an interactive`QColorDialog` on click with a `QLineEdit` that will both tab-autocomplete "standard" color names (any key from `vispy.color.get_color_dict()`), as well as accept hex values or even rgb(a) values with or without parenthesis (accomplishing goal (b) from @jni above).

you can try it out by fetching this PR and using `examples/add_points.py`:

Untitled


**2 questions:**
1. do we prefer Qt Color dialog over an OS-native color picker?  (has the benefit of being consistent across platforms, and matching the rest of napari look and feel, but the downside of perhaps breaking user expectations)
2. I updated the points layer controls to use this.  Is there anywhere else we have color pickers?

sofroniewn:
> I updated the points layer controls to use this. Is there anywhere else we have color pickers?

We use them in shapes and vectors, but we still need to get the #898 refactor from @HagaiHargil in before it might be as easy to adapt that here. You can take a look @tlambert03

tlambert03:
> we still need to get the #898 refactor from @HagaiHargil in before it might be as easy to adapt that here

ah right.  Well there's no particular reason to wait on that for this PR (this one stands alone).  In fact, it might be a good idea to get this in so they can follow this pattern there if desired.

sofroniewn:
> ah right. Well there's no particular reason to wait on that for this PR (this one stands alone). In fact, it might be a good idea to get this in so they can follow this pattern there if desired.

Agreed. I'm playing with this now. I'm finding some unexpected behaviour if i enter something bad which leads to an "unknown" warning and then start clicking to add points it defaults to "transparent" which I don't think it should do. I wonder if instead of giving "unknown"  it should just go back to what it was before?

Also is there a way to make it easy to select transparent from the dialog? Like a white square with a red diagonal line through it? Transparent also doesn't seem to have an autocomplete like the other colors do.

I do have some concerns about the stylings of the QDialog, there are a lot of options so it might seem overwhelming. I'm ok with the non-native feel. In Illustrator it is non-native. I think we can always revisit the stylings if we get a full-time designer working on the project.

I'm going to keep looking this over, but I think it is close

napari/_qt/qt_color_dialog.py
```diff
@@ -0,0 +1,267 @@
...
+# matches any 3- or 4-tuple of int or float, with or without parens
+# captures the numbers into groups.
+# this is used to allow users to enter colors as e.g.: "(1, 0.7, 0)"
+rgba_regex = re.compile(
+    r"\(?([\d.]+),\s*([\d.]+),\s*([\d.]+),?\s*([\d.]+)?\)?"
+)
+
+TRANSPARENT = np.array([0, 0, 0, 0], np.float32)
+UNKNOWN = -1  # a constant to indicate that the user entered an invalid string
+FALLBACK = TRANSPARENT  # what will be shown if a user entry is UNKNOWN
```
sofroniewn: Ah looking at the code I see why what I mentioned in the main thread happened, but I did find it unexpected as a user

tlambert03: no problem.  we can either make it black, or go back to what it was before.

napari/_qt/qt_color_dialog.py
```diff
@@ -0,0 +1,267 @@
...
+        layout.addWidget(self.lineEdit)
+
+    def color(self):
+        """Return the current color."""
+        return self.colorSwatch.color()
+
+    def _on_lineedit_edited(self):
+        """When the user hits enter or loses focus on the LineEdit widget."""
+        text = self.lineEdit.text()
+        rgb_match = rgba_regex.match(text)
```
sofroniewn, line 109: Just so I understand if this where the autocomplete functionality is happening? I've not used autocomplete in Qt before, could you just say a little something about where / how that is working (and say why I don't seem to get transparent to autocomplete)

sofroniewn: ah I see it below - sorry I should be doing this in review mode!!

napari/_qt/qt_color_dialog.py
```diff
@@ -0,0 +1,267 @@
...
+        if emit or np.all(_color == TRANSPARENT):
+            self.color_changed.emit(_color)
+
+
+class QColorLineEdit(QLineEdit):
+    """A LineEdit that takes hex, rgb, or autocompletes common color names."""
+
+    def __init__(self, parent=None):
+        super().__init__(parent)
+        self._compl = QCompleter(list(get_color_dict()))
```
sofroniewn: Ah here is the autocomplete, which I think needs transparent in there

napari/_qt/qt_color_dialog.py
```diff
@@ -0,0 +1,267 @@
...
+
+    currentColorChanged = Signal(QColor)
+    colorSelected = Signal(QColor)
+
+    def __init__(self, parent=None, initial=None):
+        super().__init__(parent)
+        self.setObjectName('QtColorPopup')
+        self.color_dialog = CustomColorDialog(self)
+
+        # TODO: ask everyone how they feel about native vs QtColorDialog look
```
sofroniewn: I like QtColorDialog, gives us more flexibility in the long run and is how illustrator looks. As noted in the main thread, we could get help from a designer in the future too

napari/_qt/qt_color_dialog.py
sofroniewn: i'd probably have it go back to what it was before, seems like the least confusing. might be worth trying what does it look like to quickly flash an "unknown" before reseting it, but that might be weird too. I'm not so so worried about this as we can improve it in the future

tlambert03: easily done.. changing now

sofroniewn (approved):
This looks good to me, I am approving though I'd like the minor changes I mention above made if possible. Then maybe @kevinyamauchi and @jni want to have a look.

I love that this PR didn't actually effect the points layer itself. That's good modularity!!

sofroniewn:
Also @tlambert03 I see the 3.7 test is failing, which I think is independent of this PR - and I think has been discussed elsewhere, but can you link me to that discussion / give me an update on plans to fix? It looks like it is in in a `test_view` is it possible because we aren't using an np.random.seed and so sometimes are generating weird data?

tlambert03:
> Also @tlambert03 I see the 3.7 test is failing, which I think is independent of this PR - and I think has been discussed elsewhere, but can you link me to that discussion / give me an update on plans to fix? It looks like it is in in a `test_view` is it possible because we aren't using an np.random.seed and so sometimes are generating weird data?

I really don't know, and I'm not sure I can be the one working on that one one. I did try to figure it out at one point (i think it's a similar phenomenon to #925) but it's beyond me at the moment.  I think @ttung was starting to look into that bug too... maybe he has time to pick it up?

tlambert03:
ok. transparent is added as an option to the autocomplete choices, and entering an invalid string sends you back to the previous setting.

>Also is there a way to make it easy to select transparent from the dialog? Like a white square with a red diagonal line through it?

the built in `QColorDiag` is actually not that flexible... so I'm not sure how we would add this sort of thing (or generally adjust the look and feel even if we had a design in mind).  Note however, there is a `Alpha Channel:` option in the dialog that you can set to 0.

sofroniewn:
> I really don't know, and I'm not sure I can be the one working on that one one. I did try to figure it out at one point (i think it's a similar phenomenon to #925) but it's beyond me at the moment. I think @ttung was starting to look into that bug too... maybe he has time to pick it up?

Yup thats fine @tlambert03 your focus should be elsewhere, thanks for the link.

> the built in QColorDiag is actually not that flexible... so I'm not sure how we would add this sort of thing (or generally adjust the look and feel even if we had a design in mind). Note however, there is a Alpha Channel: option in the dialog that you can set to 0.

Ok that's fine for me.

I have pulled this and it all works great. I think this is good to go then, but as noted above I will let @kevinyamauchi and @jni have a look before merge

tlambert03:
> Yup thats fine @tlambert03 your focus should be elsewhere, thanks for the link.

actually... now you got me thinking about it :) I think one potential fix is to use @ttung's `viewermodel_factory()` more.  The `view_layer_type` test utility function might need to be adjusted.  taking a look now

tlambert03:
it's hard to use his factory while testing the `view_*` methods, but i think the key in tony's method was calling `view.shutdown()` (rather than `viewer.window.close()`). 
I'm trying stuff here and will PR if it works: https://github.com/tlambert03/napari/pull/7

napari/_qt/layers/qt_points_layer.py
```diff
@@ -173,43 +160,27 @@ def _on_size_change(self, event=None):
...
-            if index == -1:
-                self.edgeComboBox.addItem(color)
-                index = self.edgeComboBox.findText(color, Qt.MatchFixedString)
-            self.edgeComboBox.setCurrentIndex(index)
-        self.edgeColorSwatch.setStyleSheet(f"background-color: {color}")
+    @Slot(np.ndarray)
+    def changeFaceColor(self, color: np.ndarray):
+        """Update the layer model based on user input in the color picker."""
+        with self.layer.events.current_face_color.blocker():
+            self.layer.current_face_color = color
```
jni, line 167: Wow, yeah, this is so great now!

napari/_qt/qt_color_dialog.py
```diff
@@ -0,0 +1,257 @@
...
+        """Receive QColorSwatch change event, update the lineEdit, re-emit."""
+        self.lineEdit.setText(color)
+        self.color_changed.emit(color)
+
+
+class QColorSwatch(QFrame):
+    """A QFrame that displays a color.
+
+    The internal color representation (self._color) is always a 1x4 np.ndarray
+    """
```
jni, line 143: I think the convention, for better or worse, is to have the `__init__` parameters in this docstring, as well as the public attributes.

napari/_qt/qt_color_dialog.py
```diff
@@ -0,0 +1,257 @@
...
+        self.setCursor(Qt.PointingHandCursor)
+
+        self.color_changed.connect(self._update_swatch_style)
+        self._color: np.ndarray = TRANSPARENT
+        if initial_color is not None:
+            self.setColor(initial_color)
+
+    def color(self):
+        """Return the current color"""
+        return self._color
```
jni, line 166: Should this be implemented as a property?

napari/_qt/qt_color_dialog.py
```diff
@@ -0,0 +1,257 @@
...
+        event.ignore()
+
+
+class QColorPopup(QtPopup):
+    """A QColorDialog inside of our QtPopup.
+
+    Allows all of the show methods of QtPopup (like show relative to mouse).
+    Passes through signals from the ColorDialogm, and handles some keypress
+    events.
+    """
```
jni, line 264: Could you add Parameters and Attributes to this docstring? e.g. I don't know what `initial` does looking at the `__init__` signature.

jni: (It might benefit from a rename to "initial_color"

napari/_qt/qt_color_dialog.py
```diff
@@ -0,0 +1,257 @@
...
+
+    def _on_color_selected(self, color):
+        self.colorSelected.emit(color)
+        self.close()
+
+    def _on_rejected(self):
+        self.close()
+
+    def keyPressEvent(self, event):
+        """Accept current color on enter, cancel on escape."""
```
jni: Very nice. =)

Can you add a Parameters section here, specifying the class of `event`? ie if I want to modify the behaviour of this function, how do I see what I can and can't do with `event`?

napari/layers/utils/color_transformations.py
```diff
@@ -3,18 +3,18 @@
...
 import numpy as np
+from qtpy.QtGui import QColor
+from vispy.color import Color, ColorArray
 
 from ...utils.colormaps.standardize_color import transform_color
 
-
 # All parsable input datatypes that a user can provide
-ColorType = Union[List, Tuple, np.ndarray, AnyStr, Color, ColorArray]
+ColorType = Union[List, Tuple, np.ndarray, AnyStr, Color, ColorArray, QColor]
```
jni: Hmmm, I don't like this change... It sneaks Qt into our models code. It seems like the wrong solution. If you need a signature that takes in a QColor in addition to all of the above, I suggest defining a new type `QColorType = U[ColorType, QColor]` in our Qt files... If you want `standardize_color` to take in QColors, then I suggest modifying the dictionary from the view files.

jni (changes requested):
@tlambert03 I've made a few minor comments, but request a major change, which is to remove all Qt imports from our non-Qt color-handling code. I think it should be feasible to keep all this logic on the Qt side of things.

napari/layers/utils/color_transformations.py
tlambert03: don't need it to.  I like the idea of pulling out all the qt stuff... just wasn't thinking.  Will update.

napari/_qt/qt_color_dialog.py
tlambert03: I'm always torn when doing Qt stuff, whether it's better to mimic the Qt API or the pythonic API.

jni: I'm inclined to go with Pythonic in ambiguous cases. =) Also, I think in cases where we are creating a class that we will *mostly* interact with using our own methods, rather than the built-in Qt methods, we should use a Pythonic API.

napari/_qt/qt_color_dialog.py
tlambert03: really? I thought it was the other way around...  [pep 257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings) seems to suggest that public methods and instance vars can go in the class docstring, but the constructor parameters go in the `__init__` docstring.  in any case, i can add the public stuff to the class docstring.

jni: That's disappointing that it goes against PEP 257, but it's what I've seen everywhere else, e.g. sklearn:

https://github.com/scikit-learn/scikit-learn/blob/244d118ef77a513b487b95b721179c395cbc1660/sklearn/ensemble/_forest.py#L850

jni: (And it makes sense for sphinx rendering...)

tlambert03: yeah, that's a bummer... I don't love it actually.  but I guess the hoops that sphinx makes you jump through with `__init__` justifies it.

tlambert03:
thanks for the reviews... all set for another look

napari/_qt/qt_color_dialog.py
```diff
@@ -0,0 +1,312 @@
...
+
+    def mouseReleaseEvent(self, event: QEvent):
+        """Show QColorPopup picker when the user clicks on the swatch."""
+        if event.button() == Qt.LeftButton:
+            initial = QColor(*(255 * self._color).astype('int'))
+            popup = QColorPopup(self, initial)
+            popup.colorSelected.connect(self.setColor)
+            popup.show_right_of_mouse()
+
+    def setColor(self, color: AnyColorType) -> None:
```
jni, line 182: `@color.setter`?

jni (approved):
@tlambert03 just one more minor comment, but approving nonetheless.

napari/_qt/qt_color_dialog.py
tlambert03: please no, much hard to use to connect to signals.  i know we don't have to do everything the Qt way, but I think getting too far away from the standard qt patterns and conventions starts to just confuse anyone who _is_ familiar with it.

jni: Oh ok! Sorry, "anyone who is familiar with it" and myself is an empty set. =P The question mark was genuine! =)

sofroniewn:
If this is done and @tlambert03 you want to merge then go ahead and I'll update the release notes #994 accordingly and this can go in `0.2.12`. I don't plan to merge #994 until tomorrow anyway to let some more of the team look it over so you have time

napari/_qt/qt_color_dialog.py
tlambert03: lol.  ok :)  yeah... just cause it's funny, here's what using a `@setter` would do to line 179 just above it:
```python
popup.colorSelected.connect(self.setColor)

popup.colorSelected.connect(lambda color: self.__class__.color.fset(self, color))
```

PR #820, opened 2019-12-20
Automatically set high DPI scaling in Qt

gcadenazzi:
# Description
Fixes issue #367 (with slider bars not scaling correctly in high DPI screens)

## Type of change

- [x] Bug-fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# References

# How has this been tested?

- [ ] example: the test suite for my feature covers cases x, y, and z
- [x] example: all tests pass with my change

## Final checklist:
- [ ] My PR is the minimum possible work for the desired functionality
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #834, opened 2019-12-28
Settings

tlambert03:
# Description
This PR is an opportunity to discuss how a persistent settings/preferences framework would look in napari, and also an initial example of a few things implemented with `QSettings`.  One of the immediate things that has always bothered me is that the window geometry is not preserved across sessions, so you have to resize and move every time if you'd like to have it arranged a particular way.  As an initial example, this PR retains the window size/position, and state/position of the dockwidgets across sessions.  It also auto-generates a preferences window with type-appropriate widgets that update the settings immediately upon being clicked.

@jni said [here](https://github.com/napari/napari/pull/619#issuecomment-546953127) that he's not sure he wants to use `QSettings`, and I definitely get that.  But it actually implements quite a lot of good behavior in a native cross-platform way (e.g. plist files in the OSX ~/Library/Preferences, registry keys in Windows, gsettings in linux) that befits a full-fledged application in a way would be a bit laborious to implement natively using something like dotfiles in the home directory or environmental variables.  However, this by no means precludes those options (they could all be used together even), and we could certainly change the persistence layer in this PR later, using perhaps [`appdirs`](https://pypi.org/project/appdirs/), with `json`/`yml`/`xml`, maybe [shelve](https://docs.python.org/3/library/shelve.html), or any other suggestions that people have.  It doesn't seem terrible to me to use `QSettings` inasmuch as the setting relates specifically to the GUI (where we would never _not_ use Qt).  But I do also think some `~/.napari/config` style config file might be useful for non-gui settings that may arise later.

Untitled

## Type of change

# References
https://doc.qt.io/qt-5/qsettings.html
https://github.com/ActiveState/appdirs

# How has this been tested?

## Final checklist:


sofroniewn:
This is very nice @tlambert03. I've not used or explored the `QSettings` much, so if the following comments seem misguided / reflect some misunderstanding please point them out, but I think I may favour a hybrid approach were we have all our settings as a JSON like representation, so

``` python
{'default_play_speed': 20, 'preserve_window_state': True}
```

And this could be introspected and manipulated independently of Qt and live at least in code / memory outside of our `_qt` folder (i.e. maybe in `components`, like any of our other model files). If we then decide to leverage the `QSettings` to do the fileIO / handling of paths etc. that seems fine with me, because if we decided to drop Qt one day we'd still be able to preserve the JSON representation. We could also provide some convenience methods so someone could update settings from JSON inside the app, or save to JSON from the app (i.e. maybe you have one settings you use with 3D cryoEM data and on settings you use with 2D static images and you save them to different files).

Again not sure how much of this makes sense / is possible now, but I think having some separation for the `model` like representation of settings seems valuable even if we use Qt machinery for IO. I think @royerloic would approve of that approach ;-). Curious if this is similar to how @jni imagines it too

tlambert03:
> if the following comments seem misguided / reflect some misunderstanding please point them out, but I think I may favour a hybrid approach were we have all our settings as a JSON like representation, so
> 
> ```python
> {'default_play_speed': 20, 'preserve_window_state': True}
> ```

not misguided or misunderstood at all!  We can definitely do something like that.  In terms of the actual representation on disk, I generally prefer the simple JSON file as well.  `QSettings` mostly brings "higher level" stuff to the table like [logic for choosing where to put settings](https://doc.qt.io/qt-5/qsettings.html#platform-specific-notes) in a platform-specific way.  It is integrated into the event loop, periodically [synchronizing changes ](https://doc.qt.io/qt-5/qsettings.html#sync)made to the QSettings object with the disk representation. It is also [reentrant](https://doc.qt.io/qt-5/qsettings.html#accessing-settings-from-multiple-threads-or-processes-simultaneously), allowing for access across multiple threads.  It also has a nice concept of [settings groups](https://doc.qt.io/qt-5/qsettings.html#beginGroup), which could be useful as the number of settings grows, or with plugins.  We can absolutely implement some or all of that functionality ourselves, without relying on QSettings...  it just requires a little extra doing...

> I think having some separation for the `model` like representation of settings seems valuable even if we use Qt machinery for IO.

no problem.  The model itself is pretty much just a `dict` in this case.  QSettings, just provides methods for interacting with the `dict`, synchronizing to disk, and integrating into the event loop.

I'm working on a Qt-free version and will update shortly

tlambert03:
well, as usual, when I spend some time implementing your suggestions, I end up liking it much better 😊.  I now have a Qt-independent `Settings` class that behaves much like a `dict` (inherits from `abc.MutableMapping`), and a start to a API for how one would "register" settings and persistence models.  You can have a look through the changes... and I'll try to give an overview here (trying, but likely failing, to be brief):

- The main settings object lives at [`napari.settings.Settings`](https://github.com/napari/napari/blob/81ff3a44adf4bcadcdbfc9e121d0ecc6d2787d7c/napari/settings.py#L78), and currently is instantiated as a singleton available at `napari.settings.SETTINGS`.  It behaves like a `dict`.  It has some conveniences like `autosync=True`, which will synchronize the settings to disk immediately whenever changed.
- The persistence layer is handled by setting [`Settings.format`](https://github.com/napari/napari/blob/81ff3a44adf4bcadcdbfc9e121d0ecc6d2787d7c/napari/settings.py#L143-L148), which expects a 3-tuple of `(extension, readfunc, writefunc)`... where the functions handle reading and writing the settings to disk.  We can swap that out easily if we want to, and an example/default JSON backend can be seen at [`napari.utils.settings_formats.json`](https://github.com/napari/napari/blob/81ff3a44adf4bcadcdbfc9e121d0ecc6d2787d7c/napari/utils/settings_formats/json.py)... (Note, that file highlights some of the challenges different backends will face with serialization, and there is an example of how to [handle something like `Enums`](https://github.com/napari/napari/blob/81ff3a44adf4bcadcdbfc9e121d0ecc6d2787d7c/napari/utils/settings_formats/json.py#L6))
- I use [appdirs](https://github.com/ActiveState/appdirs) to get [platform-specific folders](https://github.com/napari/napari/blob/81ff3a44adf4bcadcdbfc9e121d0ecc6d2787d7c/napari/settings.py#L166-L179), but vendor it because they don't seem to have released an update in over 2 years, despite a lot of commits to master.
- While any user can simply set some key/value pairs in the `SETTINGS` instance, one thing I particularly like is the [`Settings.register_setting`](https://github.com/napari/napari/blob/81ff3a44adf4bcadcdbfc9e121d0ecc6d2787d7c/napari/settings.py#L270-L316) method, which allows us (and eventually plugin developers), to register a setting `key`, along with a `description`, a `default` value, a `dtype` (which will be gleaned from `default` if not provided), and a `callback` that will be called whenever it changes. Registered settings appear in the main GUI preferences window at [`_qt.qt_preferences.PreferencesWindow`](https://github.com/napari/napari/blob/81ff3a44adf4bcadcdbfc9e121d0ecc6d2787d7c/napari/_qt/qt_preferences.py#L24-L43), with type-appropriate widgets (from [`_qt.auto_widget`](https://github.com/napari/napari/blob/81ff3a44adf4bcadcdbfc9e121d0ecc6d2787d7c/napari/_qt/auto_widget.py))
- Simple settings that appear as [`SettingTuple`s at the top of `napari.settings`](https://github.com/napari/napari/blob/81ff3a44adf4bcadcdbfc9e121d0ecc6d2787d7c/napari/settings.py#L10-L18) will be automatically registered ... but a more complex example of `register_setting` that mimics what a plugin developer might do can be seen in the `ViewerModel`, where I [register theme options](https://github.com/napari/napari/blob/81ff3a44adf4bcadcdbfc9e121d0ecc6d2787d7c/napari/components/viewer_model.py#L87-L95).  Because the default value is provided as an `Enum`, the preferences window shows a `QComboBox` with all of the options, and because a `callback` is provided, changing that combo box immediately updates the theme (and persists it across sessions):

  Untitled


two features that would be nice to implement:
- settings "groups":  `QSettings` has a concept of [groups](https://doc.qt.io/qt-5/qsettings.html#beginGroup) that prefixes settings keys with a groupname, and something like an `enter_group/leave_group` context manager would let related settings (like for a plugin) be grouped under a nested key conveniently.
- the current model is not reentrant, and assumes there is only one process using the preferences file, and each time it writes it clobbers the file.  If we think it would be useful, I can improve the `Settings.sync` method to keep track of `updatedKeys`, `removedKeys`, etc... to actually synchronize with some external settings (knowing who has changed in the meantime).

current problems:

- the singleton `SETTINGS` approach might need to be modified, as it can lead to unexpected behavior when multiple viewers are instantiated.  While some settings truly are global, others might not be... so we might need multiple `Settings` instances within a session.
- serialization of non-standard types remains a sticking point.  As it stands, it's mostly up to the plugin developer to make sure that their value is serializable... so we minimally need to handle errors gracefully.

royerloic:
Very nice @tlambert03 !

@sofroniewn wrote:
> Again not sure how much of this makes sense / is possible now, but I think having some separation for the `model` like representation of settings seems valuable even if we use Qt machinery for IO. I think @royerloic would approve of that approach ;-). Curious if this is similar to how @jni imagines it too
--> Indeed, I think separation of the settings model from its view and control will be key, happy to see that @tlambert03 has gone towards that design.

A few thoughts:
i) @tlambert03 I agree that a SETTINGS singleton is going to bite us eventually, global settings often end up being a 'pita'... as there will be a time when they need to be local... or viewer specific. I am not sure what the right approach should be here. Perhaps, settings should be instantiated per viewer based on a 'global' template, saving the settings to global would need to be done explicitly -- in that case only the global setting object is synced to disk.
ii)  Love the idea to use a context manager to group settings, very nice.

jni:
> saving the settings to global would need to be done explicitly -- in that case only the global setting object is synced to disk.

:+1: 

> Love the idea to use a context manager to group settings, very nice.

Agreed! This is the approach taken for e.g. matplotlib styles and it works really nicely.

Sorry that this is a very quick pass-through response, but I thought I should share some quick thoughts even before my thinking fully crystallises here:

- Over time I've come to much prefer yaml over JSON, principally because it allows comments and it is much easier to write by hand.
- I agree that the features you've described from QSettings, specifically the native-style preferences storage, is very nice, but I very much like the sync proposed by @sofroniewn between model and view for the settings that are model-specific rather than gui-specific.
- We have to be mindful of environments, profiles, etc. I can imagine that people might want significantly different settings when working from the bundled app vs from the command-line vs from IPython vs from scripts. We should have multiple "profiles" by default (either the four above or some subset of them that gets reused), and support changing profiles on the fly as well as creating custom profiles.
- for plugin settings, we should require *namespaces*, to ensure that plugin settings don't clobber each other (or ours).

sofroniewn:
@tlambert03 do you think now is a good time to start revisiting this? We know have a better sense of how the plugin stuff is shaping up that could help.

It might also be nice to add a "high DPI" setting right away in this PR that people can use to manually fix #820. Just seeing napari out in the wild last week at Janelia hits home how common and annoying that bug is.

We could also add to this PR a concept of code that executes the very first time napari is launched on after install. There could be a hidden setting "first launch" that we keep track of, and that could do the high DPI check by creating a test QApp checking if high DPI is needed adjusting our setting and then destroying that QApp before creating the real one with the correct setting now in place.

The high DPI stuff can also happen in follow up PRs if you want to keep this one minimal

tlambert03:
definitely!  I've mostly been waiting for others to feel like it's time to revisit it 😅 
I frequently come across things that I feel could use persistence ... 
I do think we should leave most of the settings out of this PR (like DPI), but perhaps we should start a new issue, kinda like the one tracking plugins, where people can dump ideas for things that they think should be made into settings.

sofroniewn:
> definitely! I've mostly been waiting for others to feel like it's time to revisit it 😅

Great, I think now is that time, especially as we have a better view on plugins - which might want to be able to add settings (though that doesn't need to be enabled now)

> I do think we should leave most of the settings out of this PR (like DPI), but perhaps we should start a new issue, kinda like the one tracking plugins, where people can dump ideas for things that they think should be made into settings.

Love it, can you make that issue, I'm happy to do so as well / contribute ideas to it too.

Let's keep this PR as minimal as possible

PR #837, opened 2019-12-30
Public contrast_limits_range setter, and climSlider popup on right click

tlambert03:
# Description
Small PR here to make it easier for people to change the range of the contrast limits slider by setting `layer.contrast_limits_range`.  This came up [one at image.sc](https://forum.image.sc/t/napari-lazy-loading-from-omero/32292).  I am also preparing a PR for complex images, and there it becomes important to be able to reset the range as the data range can change dramatically...

This also includes a helper function (`utils.validators.validate_N_seq`) that could be used anywhere we need to make sure that incoming data is a sequence with a certain length and (optionally) datatype.

Will add tests if the general idea is ok with everyone.  This will need to merge before the complex images PR (#838).

## Type of change

- [x] New-Feature (non-breaking change which adds functionality)

# How has this been tested?

- [x] added tests for some qt_image_base_layer behavior

## Final checklist:


sofroniewn:
This looks great @tlambert03 - I am in favor, and I think with tests it will then be good to go

I'm also linking #624 and #570 here too - they are more about editing the contrast limits, but maybe now we can add two QLineEdits that can be used to set the contrast limits (and if values are outside the current range then the range will expand too). All this can wait for other PRs, but I just want us to be tracking it too.

tlambert03:
maybe we could add that right-click context menu to let users manually edit it?  (like the play button menu)

sofroniewn:
`context-menu` could work though discoverability might be hard if there wasn't a dedicated icon for it. I do like the way the numbers are shown on the right of the dims sliders right now, with the slice being editable. There was also maybe the [request that right clicking causes the slider to expand](https://github.com/napari/napari/issues/570#issuecomment-544385722) and we could then add editable numbers to be only in the expanded slider.

jni:
I'm into the right-click idea — the real estate up there is precious!

Anyway, I'm 👍 on this change!

tlambert03:
it would be pretty straightforward to add a right-click to bring up an expanded range slider (for more real-estate) along with LineEdits for changing the limits... so I'll work on adding that here.  I was also playing around with [`QTapAndHoldGesture`](https://doc.qt.io/qt-5/qtapandholdgesture.html), that accomplishes the long-press behavior that @jni, once mentioned.  I think it feels a bit out-of-place in a mouse-based environment though (touchpad maybe, touchscreen definitely)... but I do think we should remember the [gestures framework](https://doc.qt.io/qt-5/gestures-overview.html) for things like sidecar!

tlambert03:
the right-click to expand the clim slider is working now.  Here's a gif... stylistic or other suggestions welcome:

Untitled

sofroniewn:
Love this @tlambert03 - I like the color of the expanded slider more than the color of the slider in the docked control panel - we might want to change those colors over to be the same, but that is a dedicated styling PR.

I've noticed slightly strange things happen if I type in a clim_range upper value that is smaller than the clim lower value, it doesn't clamp properly until after you make one adjustment of the slider - i.e. say I have `[0, 20, 40, 100]` reading left to right. If I type in `10` for the `100`, then I get `[0, 20, 10, 10]` which is obviously strange. Then if I adjust the slider it clamps everything to `[0, 20, 20, 20]`. I'd probably say that when finishing typing it should clamp to `[0, 10, 10, 10]`

Once this is in, I think it makes an interesting idea for #675 and the histogram widget - should that appear above the elongated slider? Or is that really its own thing in a dock window? Will that supersede this / be combined with it? Having a right-click on the clim slider be the way to access the histogram makes sense to me too. Curious what your thoughts are on this.

tlambert03:
Thanks for catching that @sofroniewn.  Fixed the clamping behavior, added some tests, and hopefully wired up all the necessary connections correctly.  I also added a couple "reset" and "full range" buttons, along with some helper functions on the image layer to introspect data.  If you have another chance to bang on it a little bit, that'd be great.  Looks like I have some tests related to the surface layer that I need to fix in the meantime but will ping again when ready for review.

> Once this is in, I think it makes an interesting idea for #675 and the histogram widget - should that appear above the elongated slider? Or is that really its own thing in a dock window? Will that supersede this / be combined with it? Having a right-click on the clim slider be the way to access the histogram makes sense to me too. Curious what your thoughts are on this.

yeah, I'm not sure.  There's obviously some similar functionality, but don't see them as totally overlapping.  I see this as a way to refine the behavior of an otherwise cramped slider, and expose the ability to change the range of the slider easily.  The histogram is a full-blown data evaluation tool and I sort of see it as its own dock widget (I think it takes up too much real estate for a popup)...  though perhaps a small button in the popup (or next to the clim slider) could show the histogram widget.

tlambert03:
This is ready for review, but has expanded a fair amount in scope.  I don't think that anything here should be difficult to merge with other open PRs, (but sorry if so).  Here is the updated list of what is happening in this PR:

- `Surface`/`Image.contrast_limits_range` is now a public method, setting the range on the model will update all view sliders accordingly.
- I extracted duplicated code between the `Surface` and `Image` layers into `layers.mixins.ImageSurfaceMixin` that controls gamma, clims, and colormap logic.
- Added a validator for "sequences of length n" that might be handy elsewhere in `napari.utils.validators`
- Created a new modal window (`_qt.qt_range_slider_popup`) that has a range slider and linked   `LineEdits` for range min/max and current min/max.
- Made a `QRangeSliderPopup` appear when right clicking `contrastLimitsSlider`.  It additionally includes a "reset" button (returns it to the `calc_data_range` choices) and "full range" button (only appears on `np.unsignedinteger` types and will show full bit range).

   Screen Shot 2020-01-04 at 1 10 51 PM

- a couple more tweaks to `QRangeSliders` carried over from #842
- added docs and tests that should nearly completely cover all new features as well as some existing features like `QtBaseImageControls`

sofroniewn:
@tlambert03 I'm having some problems typing negative numbers and decimals into the clim range text edits, both if the actual image data is `int` or `float`. In general I think we should be a flexible as possible there.

The clamping is now working great, thanks. I will continue to review

napari/_qt/layers/tests/test_qt_image_base_layer_.py
```diff
@@ -0,0 +1,52 @@
+import numpy as np
+import pytest
+from qtpy.QtCore import Qt
+from qtpy.QtWidgets import QPushButton
+
+from ....layers import Image, Surface
```
sofroniewn: I feel like in our tests we have tended to use absolute imports. I'm not quite sure why, but I probably wouldn't change convention here

napari/layers/mixins.py
```diff
@@ -0,0 +1,148 @@
...
+from ..utils.event import Event
+from ..utils.status_messages import format_float
+from ..utils.validators import validate_n_seq
+from .layer_utils import increment_unnamed_colormap
+
+
+validate_2_tuple = validate_n_seq(2)
+
+
+class ImageSurfaceMixin:
```
sofroniewn: Should we name this whole file `image_surface_mixin.py` instead of just `mixins.py`. I feel if we add more mixins might be nicer to keep them separate rather than all in here

sofroniewn (commented):
This is looking good, I think it's very close. I like the use of the `mixin` to reduce duplicate code. We'll probably want to do similar when @kevinyamauchi thinks about the points / shapes / text / annotation refactor.

I've left some comments, mostly pretty minor, the most significant one though is around negative / decimal values in the contrast lims text edit

tlambert03:
Thanks for catching that @sofroniewn.  I made all these changes, fixed the validator, and added a little smart width adjustment so labels don't get hidden on (slightly) bigger numbers. 

Stylistic question:  it occured to me that the meaning of the LineEdits might be hard to glean without labels, but I don't really want to add a lot of words like "range min", "current value", etc...
I _could_ add those labels if you think... or i could do something like this mockup, where the current value is always above the handle (at the expense of a little vertical room):
mockup

thoughts?  @jni?

sofroniewn:
The hover is nice, curious what others think. We should have tool tips too - they might be enough. Not sure if we have them right now

sofroniewn:
Looking at it one more time I think numbers on top is nice and clear, my vote is to go for that, with tool tips too

What does @kevinyamauchi think?

sofroniewn:
Also would be nice to add similar pop up / edit labels / range control for our single sliders too - in which case having one number above the slider is the natural generalization

This would be particularly useful for things like the point size slider where the upper end of the slider range was picked pretty arbitrarily

For things like opacity it’s still nice to expand so you can type in a direct value, but there maybe we shouldn’t allow the range ends to be editable as 0, 1 is the natural choice

Maybe for point size only the upper range is editable as 0 is the natural minimum, or maybe it’s more appropriate to think of absolute bounds that you can’t edit beyond

Anyway, all of that would be for follow up PRs, but putting down here as it might influence our thinking about the design of the range slider pop up

tlambert03:
I have a slight preference for labels-above-handles as well.  So I'll see what it takes to implement...

sofroniewn:
How do you feel about labels below vs above? I seem to recall photoshop does something similar for some sliders but under with little triangles connecting like ^

Probably don’t need the ^ but under might be more conventional

tlambert03:
I think the labels ended up looking better on top.  I implemented it, you can try it out if you'd like, here's the gif:

Untitled

sofroniewn (approved):
This looks really good! Great job as always, I love the number above the slider. Feel free to merge when ready

jni:
This looks stunning @tlambert03! 😍! I'll try to have a look at the code a bit later today, but do merge if anything is being held back — things can always be fixed in later PRs.

examples/add_grayscale_image.py
```diff
@@ -4,8 +4,9 @@
 
 from skimage import data
 import napari
+import numpy as np
 
 
 with napari.gui_qt():
     # create the viewer with an image
-    viewer = napari.view_image(data.astronaut().mean(-1))
+    viewer = napari.view_image(data.astronaut().mean(-1).astype(np.uint16))
```
jni: Maybe add a comment here that this conversion is meant to test the contrast limits automatic adjustments?

napari/_qt/layers/qt_image_base_layer.py
```diff
@@ -56,13 +59,83 @@ def __init__(self, layer):
     def changeColor(self, text):
         self.layer.colormap = text
 
-    def _on_clim_change(self, event=None):
+    def _clim_mousepress(self, event):
+        """when right clicking the clim slider, we want a QHRangeSliderPopup
+        to appear that provides greater control over the values and range of
+        the contrast limits.
+        """
```
jni, line 68: Can you make this docstring PEP257 compliant? Suggestion:

```
"""Update the slider, or, on right-click, pop-up an expanded slider.

The expanded slider provides finer control, directly editable values, and
the ability to change the available range of the sliders.
"""

File: napari/_qt/layers/qt_image_base_layer.py [hunk 5]
```diff
@@ -56,13 +59,83 @@ def __init__(self, layer):
...
+                self.layer.reset_contrast_limits()
+                self.layer.contrast_limits_range = self.layer.contrast_limits
+
+            btn = QPushButton("reset")
+            btn.setObjectName("reset_clims_button")
+            btn.setToolTip("autoscale contrast to data range")
+            btn.setFixedWidth(40)
+            btn.clicked.connect(reset)
+            self.clim_pop.layout.addWidget(btn)
+            # the "full range" button doesn't do anything if it's not an
```
jni: it's not that it doesn't do anything, it's that it's not even there. I think this should be made clear in the comment.

Don't surface layers have a dtype for their values? Or is it that only float values are supported for surface vertices?

napari/_qt/layers/qt_image_base_layer.py
```diff
@@ -56,13 +59,83 @@ def __init__(self, layer):
...
+        to appear that provides greater control over the values and range of
+        the contrast limits.
+        """
+        if event.button() == Qt.RightButton:
+            precision = 2
+            try:
+                layer_is_int = np.issubdtype(self.layer.data.dtype, np.integer)
+                precision = 0 if layer_is_int else precision
+            except AttributeError:
+                layer_is_int = False
```
jni: Whew! This was a bit hard to follow. Firstly, when will there be an attributeerror here? But secondly:

```python
try:
    layer_is_int = np.issubdtype(...)
except AttributeError:
    layer_is_int = False
precision = 0 if layer_is_int else 2
```

napari/_qt/layers/qt_image_base_layer.py
```diff
@@ -56,13 +59,83 @@ def __init__(self, layer):
...
+                btn.setObjectName("full_clim_range_button")
+                btn.setToolTip("set contrast range to full bit-depth")
+                btn.setFixedWidth(65)
+                btn.clicked.connect(reset_range)
+                self.clim_pop.layout.addWidget(btn)
+                self.clim_pop.finished.connect(self.clim_pop.deleteLater)
+
+            self.clim_pop.show_at('top')
+            self.clim_pop.curmin_label.update_position()
+            self.clim_pop.curmax_label.update_position()
```
jni: This right-click clause is very long. I suggest breaking it up into several sub-functions, something like the below:

```python
if right_click:
    create_pop_up()
    create_reset_button()
    if integer_layer:  # btw maybe this should apply to signed integers also?
        create_full_range_button()
    display_popup()
else:
    ...
```

napari/_qt/layers/qt_image_base_layer.py
```diff
@@ -56,13 +59,83 @@ def __init__(self, layer):
...
 
+        try:
+            self.clim_pop.slider.setRange(self.layer.contrast_limits_range)
+            with qt_signals_blocked(self.clim_pop.slider):
+                clims = self.layer.contrast_limits
+                self.clim_pop.slider.setValues(clims)
+                self.clim_pop._on_values_change(clims)
+        # if popup has been deleted or doesn't exist
+        except (AttributeError, RuntimeError):
+            pass
```
jni: not to do in this PR, but this pattern is reasonably common and it makes me think of a general context manager:

```python
with ignore_errors([AttributeError, RuntimeError]):
    do_things()
```

I find that more readable than the try/except, since the try is far away from the code. Might be worth submitting to the stdlib... Thoughts?

napari/_qt/layers/tests/test_qt_image_base_layer_.py
```diff
@@ -0,0 +1,52 @@
...
+@pytest.mark.parametrize('model', [Image, Surface])
+def test_base_controls_contrast_limits(qtbot, model):
+    if model == Image:
+        data = np.arange(100).astype(np.uint16).reshape((10, 10))
+    elif model == Surface:
+        vertices = np.random.random((10, 2))
+        faces = np.random.randint(10, size=(6, 3))
+        values = np.arange(100)
+        data = (vertices, faces, values)
+    layer = model(data)
```
jni: I think these datasets should be parametrised outside the test function:

```python
_IMAGE = np.arange(100).astype(np.uint16).reshape((10, 10))
_SURF = (
    np.random.random((10, 2)),
    np.random.randint(10, size=(6, 3)),
    np.arange(100),
)

@pytest.mark.parametrize('layer', [Image(_IMAGE), Surface(_SURF)])
```

Possibly, it would be best if each sub-test was its own function, but only if there isn't too much linear dependency...

napari/_qt/qt_modal.py
```diff
@@ -33,16 +32,46 @@ def __init__(self, parent):
...
-        self.frame.setLayout(self.form_layout)
-
     def show_above_mouse(self, *args):
         pos = QCursor().pos()  # mouse position
         szhint = self.sizeHint()
         pos -= QPoint(szhint.width() / 2, szhint.height() + 14)
         self.move(pos)
         self.show()
 
+    def show_at(self, pos='top', width_ratio=0.9):
```
jni: Maybe rename "pos" to "position" (as indicated in the error message), and make width_ratio keyword-only...

napari/_qt/qt_range_slider.py
```diff
@@ -50,10 +74,12 @@ def __init__(
         self.setHandleBorderColor(QtGui.QColor(200, 200, 200))
 
         self.setEnabled(True)
+        if not parent:
+            self.setGeometry(200, 200, 200, 20)
```
jni: Is the identical call below (line 85) still needed, then?

napari/_qt/qt_range_slider_popup.py
```diff
@@ -0,0 +1,168 @@
...
+            function that returns the position of the appropriate slider handle
+            by default None
+        """
+        super().__init__(value, parent=parent)
+        self.fm = QFontMetrics(QFont("", 0))
+        self.setObjectName('slice_label')
+        self.min_width = 30
+        self.max_width = 200
+        self.setCursor(Qt.IBeamCursor)
+        self.setValidator(QDoubleValidator(-9999999, 9999999, 4))
```
jni: I don't love these magic values....?

napari/_qt/qt_range_slider_popup.py
```diff
@@ -0,0 +1,168 @@
...
+        self._on_text_changed(value)
+
+        self.get_pos = get_pos
+        if parent is not None:
+            self.min_width = 50
+            self.slider = parent.slider
+            self.setAlignment(Qt.AlignCenter)
+
+    def _on_text_changed(self, text):
+        width = self.fm.boundingRect('8' * len(text)).width() + 4
```
jni, line 46: This width calculation is a bit arcane also... Could you add a comment explaining it?

napari/_qt/qt_range_slider_popup.py
```diff
@@ -0,0 +1,168 @@
...
+        layout.addWidget(self.slider, 50)
+        layout.addWidget(self.range_max_label)
+        self.layout = layout
+        self.slider.setFocus()
+        self.curmin_label.editingFinished.connect(self._curmin_label_changed)
+        self.curmax_label.editingFinished.connect(self._curmax_label_changed)
+        self.range_min_label.editingFinished.connect(self._range_label_changed)
+        self.range_max_label.editingFinished.connect(self._range_label_changed)
+        self.slider.valuesChanged.connect(self._on_values_change)
+        self.slider.rangeChanged.connect(self._on_range_change)
```
jni: This giant block of text is hard to follow... Perhaps break it up into commented sections?

napari/_qt/tests/test_qt_range_slider_popup.py
```diff
@@ -0,0 +1,40 @@
...
+    assert float(pop.curmin_label.text()) == 10
+    assert float(pop.curmax_label.text()) == 20
+    # and vice versa
+    pop.curmin_label.setText('100')
+    pop.curmax_label.setText('300')
+    pop.curmin_label.editingFinished.emit()
+    pop.curmax_label.editingFinished.emit()
+    assert np.all(pop.slider.values() == (100, 300))
+
+    # make sure changing the range labels updates the slider
```
jni: As above: it's nicer if these tests can be broken up into separate test functions (then these comments become docstrings). You can use a fixture to create a repeated starting condition...

napari/layers/image_surface_mixin.py
```diff
@@ -0,0 +1,148 @@
...
+from vispy.color import Colormap
+
+from ..utils.colormaps import make_colorbar
+from ..utils.event import Event
+from ..utils.status_messages import format_float
+from ..utils.validators import validate_n_seq
+from .layer_utils import increment_unnamed_colormap
+
+
+validate_2_tuple = validate_n_seq(2)
```
jni: Sweet functional programming goodness. =D

napari/layers/image_surface_mixin.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+class ImageSurfaceMixin:
+    """A mixin that adds gamma, colormap, and contrast limits logic to Layers.
+
+    When usinging, this should come before the Layer in the inheritance, e.g.:
```
jni:
```suggestion
    When used, this should come before the Layer in the inheritance, e.g.:
```

napari/utils/tests/test_validators.py
```diff
@@ -0,0 +1,18 @@
...
+    validate = validators.validate_n_seq(2, int)
+
+    # this should work
+    validate([4, 5])
+
+    with pytest.raises(TypeError):
+        validate(8)  # raises TypeError
+
+    with pytest.raises(ValueError):
+        validate([1, 2, 3])  # raises TypeError
```
jni:
```suggestion
        validate([1, 2, 3])  # raises ValueError
```

jni (approved):
@tlambert03 I've left a bunch of code suggestions. Some typos, mostly minor stuff, and a little bit of refactoring. I'll approve so you can merge at will.

The ImageSurfaceMixin is wild! Very cool stuff. I'd suggest a more informative name, rather than one that enumerates the subclasses. Maybe 'IntensityVisualizationMixin', or 'ColormappingMixin'?

napari/_qt/layers/qt_image_base_layer.py
tlambert03: good catch, that comment was a little outdated (I later added the dtype @property to the surface layer).  I'll fix the comment

napari/_qt/layers/qt_image_base_layer.py
tlambert03: another good catch.  The surface layer used to throw an Attribute error there (since `Surface.data` is a tuple.  But I later added the dtype property to surface making this obsolete.  thanks!

napari/_qt/layers/qt_image_base_layer.py
tlambert03: thanks, broken up and applied the full range button to all integer types.  I made them as standalone functions rather than methods, just in case we want to use them elsewhere with "rangey" type data... but I left them in the same module for now.  sound ok?

napari/layers/image_surface_mixin.py
tlambert03: oh, you don't know the word "usinging"?  must be an australian/american english thing... ;)

napari/_qt/layers/qt_image_base_layer.py
tlambert03: how about [this lovely gem](https://docs.python.org/3.8/library/contextlib.html#contextlib.suppress)?
```python
with contextlib.suppress(AttributeError, RuntimeError):
    do_things()
```

napari/_qt/layers/tests/test_qt_image_base_layer_.py
tlambert03: fixed

napari/_qt/qt_range_slider_popup.py
tlambert03: silly me... can just instantiate a [QDoubleValidator](https://doc.qt.io/qt-5/qdoublevalidator.html#QDoubleValidator) without any values to achieve the desired effect.

napari/_qt/qt_range_slider_popup.py
tlambert03: will do.  The point there was that, in a non-monospaced font, numbers have different widths, which ended up causing a very slight jitter in the size of the label as the value changed. '8' is a wide constant width...  will add a comment

napari/_qt/qt_range_slider.py
tlambert03: (that's a thumbs down for, "nope" ... 😊)

tlambert03:
Thanks for the thorough review @jni!  So useful... sorry it turned into such a big one.

Caught a couple outdated exceptions for me.  I made all the changes, renamed the mixin, enabled the full range button on all integer types (not just unsigned), and broke up the tests.

Will wait for tests to pass and then merge

houpuli:
> the right-click to expand the clim slider is working now. Here's a gif... stylistic or other suggestions welcome:
> 
> Untitled

Hi! thanks for making this gif.  I tried to follow this yet in my case nothing happened when I right click on the slider. 
the small slider itself works fine, but I don't have this popup. 

my environment is:
napari: 0.2.11
Platform: Windows-10-10.0.17134-SP0
Python: 3.7.4 (default, Aug 9 2019, 18:34:13) [MSC v.1915 64 bit (AMD64)]
Qt: 5.9.6
PyQt5: 5.9.2
VisPy: 0.6.4
NumPy: 1.16.5
SciPy: 1.3.1
scikit-image: 0.15.0
Dask: 2.5.2

GL version: 4.0.0 - Build 10.18.10.4425
MAX_TEXTURE_SIZE: 16384

sofroniewn:
Thanks for reporting this @fenixnano can you try updating to napari `0.2.12`, though I think this should have still worked with `0.2.11`, and then updating your Qt / PyQt5 too. I'm now on `5.14.1`. I suspect that is the source of the problem. Maybe @tlambert03 knows if this behavior is expected or not with `5.9.x`

VolkerH:
I actually had some issues with this as well. For me with a dual screen set up the pop up showed up on the other screen (and I didn't notice). I meant to create an issue but too scatter-brained at the moment.

tlambert03:
I can confirm that the slider popup as currently implemented will _not_ work on Qt < 5.10.  Technically, we don't support Qt < 5.12.3, but it probably shouldn't be too hard to make this one work for older Qt installations.  I'm going to create a separate issue for this

Issue #792, opened 2019-12-13
Clarifying the interface between the image layer and data


d-v-b:
## 🚀 Feature
(This is a follow-up to discussions that occurred offline, and on [zulip](https://napari.zulipchat.com/#narrow/stream/212875-general/topic/Async.20Tiles.20.23318)) 

Currently there is a lot of logic in the image layer for handling different types of data, e.g. arrays that represent grayscale data (the default), arrays that represent RGB data, and image pyramids (lists of arrays). Accordingly, the image layer has `rgb` and `is_pyramid` properties, which the image layer uses when figuring out how get images from the user-supplied data, e.g. in [`Image._set_view_slice`](https://github.com/napari/napari/blob/master/napari/layers/image/image.py#L423).  
There are several standing issues that might further complicate the data wrangling in the image layer, namely adding functionality for lazy loading of planar views (https://github.com/napari/napari/issues/318) and adding support for a physical coordinate model (https://github.com/napari/napari/issues/763). I think these issues might be simpler to resolve if there's a cleaner separation of concerns between the _image layer_ and the source of the data that the image layer is displaying. I know @jni's blood pressure shoots up when he hears the O word, but I want to use this issue to explore the advantages and disadvantages of making an object (tentatively called a `datasource`) to take a lot of data logic out of the image layer class.

Schematically, I'm a imagining what life would be like if the image layer had a `datasource` object which could act like a locus for data properties and access methods, e.g.

* dimensional information / units (e.g., how the data grid maps to world coordinates)
* min and max values of the data (which are not the same as the layer's contrast limits)
* whether the data has changed (maybe important for streaming camera data)
* whether the data is RGB or not
* a method that takes some spatial parameters and returns a view of the data corresponding to those parameters. By making this a method call and not just using `array_like.__getitem__`, it becomes a lot easier to pass in arrays of (eventually) floats for resampling data off-grid. 

I'm still not super sure what exactly this object should look like, or what it should do. But I think adding a little bit of abstraction in this area could make solving some current issues a lot easier.

To help with https://github.com/napari/napari/issues/763, a `datasource` object could totally encapsulate the mapping from the data grid to a grid in the viewer space. That information (e.g., an affine matrix and a transform method) needs to live _somewhere_; a datasource-like object seems like a good fit for it. 

[...]

PR #838, opened 2019-12-30
Add support for complex valued images

tlambert03:
# Description
This PR adds support for complex valued images (and pyramids), by adding a `QComboBox` to the layer controls when an image is detected as complex.  The combo box lets the user select between `real`, `imaginary`, `magnitude`, and `phase` representations of the complex image.

Untitled

In the `Image` layer, this is represented as a new `layer.complex_func` attribute which can be set either with one of the above strings, as an enum (`layers.image._constants.ComplexRendering`), or (and curious to get your thoughts on this) as any callable that accepts a complex valued `np.array` and returns a real-valued array.  That last functionality would only be accessible by directly setting `layer.complex_func = my_custom_func` in python, and the `QComboBox` in the GUI would then show the name of the custom function.

The function itself is currently applied in `Layer._set_view_slice`.  I sort of see it as a way to simply convert the underlying data to a viewable (real) representation, kinda like a "view" does... so it seemed appropriate there.  But let me know if you would have that happen elsewhere.

Note that whenever the `complex_func` changes, i reset the contrast limits slider (or set it to `(-pi, pi)` when viewing `phase`), so I have already merged changes from #837 here.  I would wait for that to merge to master before merging this (and you can ignore the bits that change _contrast_limits_range).

There are two examples added at `add_image_complex.py` and `add_pyramid_complex.py`.  With your approval, I'll add some tests to make sure all the new code is covered.

## Type of change

# How has this been tested?

## Final checklist:


sofroniewn:
Very cool, I know a number of people have requested this, so will be nice to have. I'll wait till further along to give a detailed review, but approach looks good

jni:
Nice, @tlambert03! In principle I'm already on board with this, and the features below can come later, but just wanted to throw them out there: (they are my answer to your question on zulip ;)

- do we want to automatically change/subset to cyclic colormaps when viewing phase? Since you are changing the contrast limits, this doesn't seem like a big addition. We already provide twilight and twilight shifted for this purpose. There's a few other nice cyclic colormaps out there (though I do not include hsv in this category).
- Additionally, we could provide a special mode to view both phase (cycle around a color wheel) and magnitude (saturation) in the same view. 
- Do we want to provide some convenience argument/magic behavior to apply fftshift to the input, so that users can feed in the output of scipy.fft directly into napari?

It would be great to get @royerloic's input here also since native support for complex numbers was one of the reasons he was excited originally to switch to Python. =)

royerloic:
Totally love it @tlambert03 . I was definitely looking forward to see that implemented.
I think the current rendering modes are great. I would suggest adding one in which the magnitude of the complex number is encoded by the color luminance (and lightness within a range), and the phase is encoded by the hue (rainbow color wheel). This would render as an RGB image that would nicely convey the full vectorial nature of complex numbers. Could be useful to represent 2D vector fields too. I am sure that @bryantChhun and Shalin Metha's team and others would find that useful.  

This website goes through the HSL to RGB conversion:
https://www.rapidtables.com/convert/color/hsl-to-rgb.html

Complex numbers have 2 degrees of freedom but RGB and HSL have three, so we actually have more 'room'. One suggestion is to have lightness vary with complex amplitude too, but over a restricted range: [0%, 50%], because 100% lightness is white and then you can't see the information of the phase encoded by the hue. For small amplitudes, it becomes a feature that a small lightness reduces perceived hue differences because the complex plane is 'continuous' around the origin so it is good not to have rapidly varying hues. 

I would suggest using numexpr ( https://github.com/pydata/numexpr ) to speed up some of the numpy math here... In general I am interested to know what @jni and @sofroniewn and the others think about having this as a dependency and whether we want to benchmark this in some other parts of napari that have a lot of complex numpy expressions...
(I have been using it a lot and it works great)

sofroniewn:
> I would suggest using numexpr

Is there a reason to use this over numba?

I think at some point in time we're going to probably want to add support for some faster array operation approach (a lot of the shapes / vectors / mesh code would probably benefit from it) but we'll probably only want to support one approach

jni:
> is there a reason to use this over numba

For mathematical expressions, numexpr is hard to beat, but I really hate its string-based API. See this thread (including the link from Francesc) for more info:
https://mail.python.org/pipermail/numpy-discussion/2019-July/079867.html

I'm mildly :-1: on depending on numexpr. It's certainly not a trivial dependency, and as I mention above, I dislike the API, though it could conceivably be improved with recent Python technology. We already have hsv2rgb implemented in scikit-image so using that should be our first step. 

A caveat on the numba measurements above: they have not been optimised by the numba folks, so I imagine that you can actually do much better with numba than shown there, probably approaching numexpr performance.

Re colors, HSV has similar problems to jet/rainbow in terms of perceptual uniformity, so although I'm happy using it as a first step, I would suggest that we aim to use a perceptually uniform cyclic colormap as demonstrated here: https://peterkovesi.com/projects/colourmaps/ 

Having said all this, @tlambert03 as our resident colorblindness expert, do you have opinions on cyclic colormaps? How does twilight work for you? Do you know of any other colormaps that are cyclic, colorblind-friendly, *and* have a second degree of freedom? I know it's a big ask. :grimacing:

tlambert03:
>do we want to automatically change/subset to cyclic colormaps when viewing phase? do you have opinions on cyclic colormaps?

I like the idea of automatically switching, but I kinda wish it could be a setting (#834 😉), since someone may prefer something other than our "choice" and it could be annoying to have to switch it every time.  I'm personally a fan of "white-red-white-blue" cylic maps.  I know it comes at the expense of 0/180˚ ambiguity, but I guess I find that those are usually differentiable by context (phase changes are rarely discontinuous), and the simplification make the visual color challenge much easier for me.  So I wouldn't mind seeing that added (or even just non-cyclic bwr, since there are many things that often fall well within a 2-pi range, and don't actually need to cycle... like pupil/zernike plots).  But generally in support of any Koveski suggestions.

> @jni: Additionally, we could provide a special mode to view both phase (cycle around a color wheel) and magnitude (saturation) in the same view.
> @royerloic: I would suggest adding one in which the magnitude of the complex number is encoded by the color luminance and the phase is encoded by the hue .

love it.  working on it.

> Do we want to provide some convenience argument/magic behavior to apply fftshift to the input, so that users can feed in the output of scipy.fft directly into napari?

sounds great... can't exactly picture how the magic would work, unless we assumed that the DC component is always much higher?

> numba/numexpr

have worked with numba more than numexpr.  Definitely imagine we'll want something like that eventually, but agree it's a bigger conversation.

tlambert03:
got something kinda cool!  We may not end up using all of these variants, but here's an example (or you can pull the PR, since the gif kinda kills it) of converting complex to RGB ... with a somewhat [flexible function behind it](https://github.com/napari/napari/blob/d60789943ff818d88a185fb21f183afe6ea7bf78/napari/utils/complex.py#L17-L69) that allows for easy variation on which component of the image is mapped to which component in colorspace (like [here in the enum](https://github.com/napari/napari/blob/d60789943ff818d88a185fb21f183afe6ea7bf78/napari/layers/image/_constants.py#L68-L73)).

  The image here is a ramp of the phase component in the X dimension from `-pi` to `pi` and the Y dimensions is a magnitude ramp.  The combo box names represent the mappings, for instance: `map_p2h_m2s` means "map phase to hue and magnitude to saturation... value is always 1"

Untitled

tlambert03:
updated it with what I think is a nice, more flexible user interface ("map_colormap" in the combobox).  Here the user gets to pick a colormap with the existing dropdown, and phase is mapped to color, while magnitude is mapped to brightness.  When in this mode, the contrast_limits sliders are hijacked to control the phase range limits (defaults to -pi to pi), and the gamma slider is used to control the linearity of the magnitude -> brightness mapping.  It's basically the same as the HSV->RGB mapping we started out with, but not restricted to the HSV color-table, and with finer grained control over the lookups. 

i think it would be easy enough to implement as a shader, but I'll leave that for a later PR.  Thoughts @royerloic, @jni?

(same Xphase/Ymag ramp image as above)
(Untitled3)


[longer version @ YouTube](https://www.youtube.com/watch?v=rkxg_CQF61Q&feature=youtu.be&vq=hd1080)

examples/add_image_complex.py
```diff
@@ -0,0 +1,19 @@
+"""
+Display one image using the add_image API.
+"""
+
+import napari
+
+# from skimage import data
```
sofroniewn: any reason for leaving the commented code?

napari/layers/image/image.py
```diff
@@ -163,12 +164,19 @@ def __init__(
...
+            interpolation=Event,
+            rendering=Event,
+            iso_threshold=Event,
+            complex_func=Event,
         )
 
         # Set data
         self.is_pyramid = is_pyramid
         self.rgb = rgb
+        self.iscomplex = (
```
sofroniewn: can this be `is_complex` - I think we use `is_pyramid`, so seems more consistent

napari/layers/image/image.py
```diff
@@ -337,6 +350,64 @@ def rendering(self, rendering):
         self._rendering = rendering
         self.events.rendering()
 
+    @property
+    def complex_func(self):
```
sofroniewn: I don't love the name `complex_func`, is there something else that can be more explicit, maybe `complex_converter`? Not sure I love that either.

napari/utils/misc.py
```diff
@@ -46,6 +46,11 @@ def is_iterable(arg, color=False):
         return True
 
 
+def is_complex(obj):
```
sofroniewn: maybe this should be named `guess_complex` (I think we may have used that convention elsewhere)

sofroniewn (approved):
This looks great to me, I've left some mostly minor comments and given it a try, but I don't have any experience with complex data, so would be good if @jni can look at this too before merge

tlambert03:
Thanks @sofroniewn!  I'll make all these changes.  I should have clarified that it's not quite ready (sorry, should have had WIP in the title)... but it's very close!  I just wanted to confirm that the behavior is ok with everyone. Assuming everyone is on board, I'll clean it up, make sure I'm not breaking things, and add tests before we merge.

sofroniewn:
@tlambert03 I know you're pushing on many fronts right now, but I was wondering if you could bring this over the finish line. I think there were only very minor revisions requested and it won't conflict with much else development on the repo. Also I might want to use it for a visualization I want to make soon :-) :-)

jni:
Looks like I didn't comment to agree that the current UI is a good one! Go for it! :tada:

napari/layers/image/image.py
tlambert03: renamed to `complex_rendering`, which sticks with other terminology.

napari/_qt/layers/_tests/test_qt_image_base_layer_.py
```diff
@@ -94,3 +94,15 @@ def test_clim_slider_step_size_and_precision(qtbot, mag):
...
+@pytest.mark.skip(reason="broken test that needs to be fixed")
+def test_complex_colormap_mode_in_3d(viewermodel_factory):
+    """Test that complex rending in one of the RGB modes works in 3d views."""
+    view, viewer = viewermodel_factory()
+
+    data = np.random.random((10, 10)) + np.random.random((10, 10)) * 1j
+    layer = viewer.add_image(data)
+
+    layer.complex_rendering = 'colormap'
+    viewer.dims.ndisplay = 3
```
tlambert03, line 108: this line causes a problem in `_set_view_slice` here:
https://github.com/napari/napari/blob/bd538488c9a6f9d6721cec8268610a1ea5b0c7d0/napari/layers/image/image.py#L491

order is incorrect (`(0,)`)... it has to do with hijacking the RGB system for complex numbers... but haven't figured it out yet.

tlambert03:
@sofroniewn, this is pretty much ready now.  As mentioned in meeting, there are a few "janky" steps here that might be a little brittle, specifically for the modes that convert complex data to RGB images.  I don't think any of this should affect _non-complex_ data at all, and _most_ of the main use-cases for complex data will be fine... so it's probably fine to go with this and improve over time (including, as you mentioned, doing this all with shaders).

there is one bug that I can't yet figure out: if you open a complex image, change the complex mode to an RGB type (like "colormap"), then go into 3D, you get an error in the `transpose(order)` step of `_set_view_slice`.  Feel free to take a look if you want, you can try it with `examples/add_image_complex.py`.  I added a skipped test above to remind us about this, but I personally am ok with merging with that known (and probably relatively fringe) bug.

tlambert03:
also... the failing tests are unrelated... and unfortunately seem to harken back to #925

royerloic:
Love it. Great work.

A few points:

(i) If we start hijacking sliders from one purpose to another, we should probably change their names as exposed to the user. Since complex number viz is a advanced feature, and we have the explicit notion of mapping that's probably fine for now. I would also second @sofroniewn that it would be great to merge this. We should gather some experience using it so we know what to fix for the next version. 

(ii) @tlambert03: about shaders: I think we need a PR that would move computations for contrast limits, gamma, (log?), (invert lut?), and complex mapping, into the shader. This would bring lots of benefits. I think this is the main source of performance degradation at this point. I have observed that 'playing' the time dimension of a 2D+t stack is difficult (lost control of UI) at, more than 10 fps and one of my CPUs is fully utilised doing so, that's probably due to a lot of numpy calls happening...  Will make an issue.

(iii) Do we have support for 3D rendering complex data with this? How hard? Just a question. I am sure a follow-up PR would be the right place to do it.

tlambert03:
> (i) If we start hijacking sliders from one purpose to another, we should probably change their names as exposed to the user.

could do...  I didn't here because the sliders really are still controlling contrast_limits (for phase info) and gamma (for mag->intensity) lookup.  but I agree that even a subtle adjustment warrants some indication.

> (ii) @tlambert03: about shaders: I think we need a PR that would move computations for contrast limits, gamma, (log?), (invert lut?), and complex mapping, into the shader. This would bring lots of benefits. I think this is the main source of performance degradation at this point. 

absolutely agree.  so many benefits.  definitely the main slowdown.

> (iii) Do we have support for 3D rendering complex data with this? How hard? Just a question. I am sure a follow-up PR would be the right place to do it.

yep this works for 3D.  I tried some 3D PSF->OTF examples on my own and they worked great.  The caveat (as mentioned above) is that the RGB rendering modes (combining phase and mag in the same view) don't yet work in 3D.  But all the standard ones (`abs, phase, real, imag`) do.

tlambert03:
the slider labels now change to "phase limits" and "mag gamma" when the complex rendering mode is "colormap"

sofroniewn:
@tlambert03 I find `mag gamma` a bit jargony - I guess it stands for `magnitude` gamma, but I wonder if it will confuse people. Maybe it's ok. It's also strange to me that `mag gamma` and `phase limits` aren't used for the `p2h...` options, can they not be?

I also find the `colormap` option and the the `p2h...` etc options in the complex drop down and how they interact with the colormap drop down slightly confusing. When I have the `p2h...` etc options selected the colormap changes and then changing the colormap again has no further effect - this makes me wonder if those options couldn't actually be inside the `colormap` dropdown instead of the `complex` dropdown and be added / taken away when say the `complex` dropdown is set to `colormap`. 

Stepping back, as is see it there are 5 ways of treating the complex number - as a complex number, a real part, an imaginary part, as a phase, and a magnitude - and I want one dropdown to correspond to those options. Maybe named `complex`, `real`, `imaginary`, `phase`, and `magnitude`. Maybe these are in a dropdown named `representation:`.

Independently there are ways to do the coloring / colormaps. When we have real / imaginary / phase / magnitude representation selected we just have our normal colormaps and things proceed as expected. When we have `complex` representation selected we have our normal colormaps for which phase is mapped to color and magnitude is mapped to brightness using `complex2colormap` and we have the `p2h...` etc options which use `complex2rgb`.

When in the `complex` representation we can rename the sliders to `mag gamma` and `phase limits` either irrespective of colormap, or depending on colormap - you know better here. 

How does this sound? Do you think this is more intuitive? I'm trying to pull apart how we treat the complex number from the how we actually do the coloring.

tlambert03:
as a front-facing goal, I like this suggestion a lot.  I think it makes more sense for all the `p2h_*` type options to live in the colormap menu and I'll aim to implement that.  As I started to do it though, I started realizing just how invasive it was to put a string in the colormap QComboBox that isn't really a mapping to an actual colormap.  (i.e. it's really hard to do that without affecting mixins and other things that have nothing to do with the image layer).  So I need to think a bit about how best to achieve that goal.

One possibility that might make all of this much easier is if we actually had a `ComplexImage` Layer type that subclassed from `Image`.  Given all of the places where I've now mucked up the `Image` with "`if self.is_complex`" it would make the whole thing cleaner... should I pursue that approach?

sofroniewn:
Hmm. Initially I think I was leaning away from subclassing the image layer. I definitely want to avoid having to add an `add_complex` method, but maybe we could stick with just `add_image` and then inside that function do the guessing etc and then call `Image` or `Complex` or `RGB` or `Pyramid`, the last three of which all subclass `Image`. Maybe that would actually be much cleaner as our API stays minimal, but under the hood we use nice objects. Thoughts? You could try that approach for `Complex` in this PR, then if we like we can extend to `RGB` / `Pyramid`.

sofroniewn:
Maybe `Pyramid` should stay on Image actually, as we want `Labels` to be able to use `Pyramid`, also `RGB` might be something that gets used in `Complex` too - so have to think about this more. It might be simpler to avoid all the subclassing ...

d-v-b:
This functionality looks super cool, but I'm worried about bloat the image layer -- this PR adds yet another block of if/else evaluations in [set_view_slice](https://github.com/tlambert03/napari/blob/complex/napari/layers/image/image.py#L512), which makes me a little nervous, since it's hard enough to tell what that function actually does.
  
As someone who tried and failed to get traction on the async tiles issue because of arcane workings in the image layer, I'm strongly in favor of adding a layer of abstraction on top of the array-likes that go in to napari. This abstraction layer would also be a natural place for putting coordinate stuff, too. I'm happy to continue that discussion in https://github.com/napari/napari/issues/792

royerloic:
@tlambert03 @sofroniewn 
I am very much in favour of having subclassing internally while having a unified front facing API, that should take care of @tlambert03 observation that the non-complex image class gets polluted by if then else and keep the API clean and simple to the outside as @sofroniewn suggested.

tlambert03:
sounds good.  I'll take a stab at a subclass without changing the front-facing api

Issue #763, opened 2019-12-05
Physical coordinate systems


sofroniewn:
## 🚀 Feature

There have been a number of requests from the community around supporting physical coordinates. Right now we have `scale` and `translate` properties and a scene graph so it will be possible, but we need to do some work around our APIs and bug-fixing to make it work well. We also have to make our scaling in Z work appropriately with our sliders and dims range. There have been a number of issues raised around this already, but I wanted to collect some of them here #614, #659, #640

I'm not sure if the vispy scene graph can handle rotations, but we should look into that. I'm fine with delaying on that for a bit too.  

@axtimwalde proposed the addition of a new coordinate system so we now have something akin to `layer`, `world`, and `canvas` coordinates. I know @jni has also discussed this with him too.

We can collect implementation / API ideas here.

sofroniewn:
We should also fix #707 at this point too

GenevieveBuckley:
This is of particular interest to me, and I may be able to help with example use cases & demo data here. 

I didn't catch Juan's conversation with Stephan, so subscribing here to get a better sense of people's thoughts.

jni:
Yep so a little bit about how Big Data Viewer works, as I understood it/remember it.

Somewhat surprisingly, all of the rendering is CPU, not GPU-based. All of the speed is achieved through lazy rendering (as suggested by @d-v-b and @constantinpape) and async/non-blocking interactivity. There is a transformation from screen-pixel/canvas space to world space (typically, just a scaling, but in our case for e.g. grid mode, it could be more complicated, including a tiling), and from world space to the data space of each dataset/layer.

On each draw event, BDV steps through all the pixels, and for each pixel, it composes the transformations together to find the source location in the original data and grabs it. Here, Java has an advantage over Python in that the compiler inlines all of the transformations and it's reasonable to do them all for each individual screen pixel. In Python, we would need to find a way to vectorise this operation, or else move a chunk of code to numba. 

The other advantage they have is that the imglib2 data model is pretty excellent and again, efficient to access on a per-pixel level. Unlike a dask array, asking for a pixel value in an imglib2 array actually grabs it. The way they minimise IO is that the data source (which, like a dask array, can be virtual, backed by computation or IO) is transparently cached, so the first pixel grab is slow, but subsequent ones to the same block are fast.

[...]

PR #839, opened 2019-12-31
[WIP] support mixed 2D / 3D rendering

### LINKED ISSUE #639: Support mixed 2D and 3D rendering
There have been a couple requests for mixed 2D and 3D rendering - for example

_Originally posted by @GenevieveBuckley in https://github.com/napari/napari/issues/420#issuecomment-546795030_

>
>Does "Support mixed 2D and 3D rendering" have an issue I can follow?
>
>I often have a 3D volume (fluorescence microscopy), and a 2D image (scanning electron >microscopy) from the same sample. It'd be really awesome if I could overlay the two.
>
>They generally have very different angles of view from the detector to the sample, as well >as very different pixel sizes which makes things kinda tricky.

I've done some exploration of this from the vispy side and our scene graph can handle most of this complexity, we've just got to think about how to best expose the interactivity to the user

sofroniewn:
# Description
This PR will close #639 by supporting mixed 2D / 3D rendering for all layer types. I'm not quite sure about API yet, and I havn't added full documentation / tests, but I wanted to get this going to see what it is like. It'll also probably intersect with some of the work on physical coordinates - see #763, and also orthoviews, see #760, but I do like the functionality so far.

To make it work you must set the `viewer.dims.embedded = True`, this will pop up a third slider corresponding to the "embedded" or "sliced" dimension. You must then set the specific layer that is to be embedded to have `layer.dims.ndisplay=2` (while keeping the whole viewer in 3D rendering mode, i.e. `viewer.dims.ndisplday=3`).

Here are a couple gifs of the new functionality, note that the blending modes work really nicely here:  
mixed_2D_3D_stent

and mixed surface + image rendering with some cryoET data

mixed_2D_3D_cryoET

## Type of change

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/components/dims.py
+++ b/napari/components/dims.py
@@ -53,7 +53,15 @@ class Dims:
         Order of only displayed dimensions.
     """
 
-    def __init__(self, ndim=None, *, ndisplay=2, order=None, axis_labels=None):
+    def __init__(
+        self,
+        ndim=None,
+        *,
+        ndisplay=2,
+        order=None,
+        embedded=False,
+        axis_labels=None,
+    ):
         super().__init__()
 
         # Events:
@@ -64,6 +72,7 @@ class Dims:
             axis_labels=None,
             ndim=None,
             ndisplay=None,
+            embedded=None,
             order=None,
             range=None,
             camera=None,
@@ -76,6 +85,7 @@ class Dims:
         self._axis_labels = []
         self.clip = True
         self._ndisplay = 2 if ndisplay is None else ndisplay
+        self._embedded = embedded
 
         if ndim is None and order is None and axis_labels is None:
             ndim = self._ndisplay
@@ -297,6 +307,19 @@ class Dims:
         self.events.camera()
 
     @property
+    def embedded(self):
+        """Bool: Whether 2D display is embedded in 3D rendering."""
+        return self._embedded
+
+    @embedded.setter
+    def embedded(self, embedded):
+        if self._embedded == embedded:
+            return
+
+        self._embedded = embedded
+        self.events.embedded()
+
+    @property
     def displayed(self):
         """Tuple: Dimensions that are displayed."""
         return self.order[-self.ndisplay :]
@@ -305,6 +328,14 @@ class Dims:
     def not_displayed(self):
         """Tuple: Dimensions that are not displayed."""
         return self.order[: -self.ndisplay]
+
+    @property
+    def sliced(self):
+        """int: Dimension that is sliced if embedded."""
+        if self.embedded and self.ndim >= 3:
+            return self.order[-3]
+        else:
+            return None
 
     @property
     def displayed_order(self):
--- a/napari/_qt/qt_dims.py
+++ b/napari/_qt/qt_dims.py
@@ -61,6 +61,7 @@ class QtDims(QWidget):
         self.dims.events.range.connect(lambda ev: self._update_range(ev.axis))
         self.dims.events.ndisplay.connect(self._update_display)
         self.dims.events.order.connect(self._update_display)
+        self.dims.events.embedded.connect(self._update_display)
 
     @property
     def nsliders(self):
@@ -137,7 +138,7 @@ class QtDims(QWidget):
         """
         widgets = reversed(list(enumerate(self.slider_widgets)))
         for (axis, widget) in widgets:
-            if axis in self.dims.displayed:
+            if axis in self.dims.displayed and axis != self.dims.sliced:
                 # Displayed dimensions correspond to non displayed sliders
                 self._displayed_sliders[axis] = False
                 self.last_used = None
--- a/napari/_vispy/vispy_base_layer.py
+++ b/napari/_vispy/vispy_base_layer.py
@@ -138,10 +138,16 @@ class VispyBaseLayer(ABC):
         self.layer.position = self._transform_position(self._position)
 
     def _on_translate_change(self):
-        self.translate = [
-            self.layer.translate[d] + self.layer.translate_grid[d]
+        translate = [
+            self.layer.translate[d]
+            + self.layer._translate_view[d]
+            + self.layer.translate_grid[d]
             for d in self.layer.dims.displayed[::-1]
         ]
+        d = self.layer.dims.sliced
+        if d is not None and len(translate) == 2:
+            translate = translate + [self.layer._translate_view[d]]
+        self.translate = translate
         self.layer.position = self._transform_position(self._position)
 
     def _transform_position(self, position):
--- a/napari/_vispy/vispy_image_layer.py
+++ b/napari/_vispy/vispy_image_layer.py
@@ -149,15 +149,6 @@ class VispyImageLayer(VispyBaseLayer):
             self.layer.top_left = self.find_top_left()
         self.layer.position = self._transform_position(self._position)
 
-    def _on_translate_change(self):
-        self.translate = [
-            self.layer.translate[d]
-            + self.layer._translate_view[d]
-            + self.layer.translate_grid[d]
-            for d in self.layer.dims.displayed[::-1]
-        ]
-        self.layer.position = self._transform_position(self._position)
-
     def compute_data_level(self, size):
         """Computed what level of the pyramid should be viewed given the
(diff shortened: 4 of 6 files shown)


sofroniewn:
I think here we have to also consider the tradeoffs between this sort of functionality and supporting cropping of the data during 3D rendering - see comment here https://github.com/napari/napari/pull/846#issuecomment-570462328.

Once #846 is merged I might give cropping a try before we try and finish this PR. Maybe at first just single-ended cropping with our current sliders, though eventually we'll want to use the range slider - probably after the refactor in #844 - to do double ended cropping from both sides

GenevieveBuckley:
To clarify, can the 2D plane lie slightly off-axis or tilted relative to the volume array?

sofroniewn:
> To clarify, can the 2D plane lie slightly off-axis or tilted relative to the volume array?

Not really, this is mainly about slicing the array, and certainly doesn't have any of the concepts that are coming in #885. I think we'll want to wait on this until #885 goes in too, and I'm also still not sure that cropping functionality isn't more natural here too.

I do remember though you were interested in this functionality right? If so can you describe a little more about how you'd like to see this work / what functionality you need

GenevieveBuckley:
> > To clarify, can the 2D plane lie slightly off-axis or tilted relative to the volume array?
> 
> Not really, this is mainly about slicing the array, and certainly doesn't have any of the concepts that are coming in #885.

That's what I thought. Still a big step forward though!

> I do remember though you were interested in this functionality right? If so can you describe a little more about how you'd like to see this work / what functionality you need

Our lab combines scanning electron/ion beam imaging (2D images, often with some "perspective" tilt to the view) with fluorescence microscopy (3D volumes with colour channels) of the same samples. We want to:
* spatially relate the two datasets and display them together
* have support for control point matching, so users can do manual registration
* have better support for affine transforms (for manual or automated image alignment & registration)

GenevieveBuckley:
Here's a great example of something that I have now that I really want to use napari to make it better.

This is image alignment using manual control point matching of two 2d images of different modalities. I've added some image transforms on top of this [cpselect](https://github.com/adal02/cpselect) tool, and it's what we're using for now:
image
What I like about this is that it's very easy to add/delete matched control points. It *requires* you to click once in each image, and won't let you accidentally add unmatched points if you click several times in the same image.

How could napari make this better?
* A *huge* disadvantage right now is that users need to be able to adjust the image brightness/contrast while selecting matched control points, but need different display settings in the final figure for papers. We need to crank up the brightness in the fluorescence display to see the grid bars on the sample holder (used as reference markers) which are barely distinguishable from the auto-fluorescence background. 
* If napari gets good support for nD transforms (and mixing 2D/3D data), I might not have to use a maximum intensity projection to squash my 3D fluorescence volume into a 2D image. It would be more accurate overall, especially the two cameras are often at very different angles relative to the sample.
* I don't want to lose the ease of use of adding matched control points across images/layers, but I'm sure we can work on making sure that support is also added to napari. 

Longer term, I'd love to have our main microscope control GUI display be napari based too. I can launch napari instances from my own PyQt GUI, so that's great. But I'm not sure about best practices to get information back out of napari (so I'm following the pluggy discussions closely), or how to make data displayed in several separate napari instances play nicely together (napari doesn't have the right support for showing two workspaces side by side - grid mode isn't good for this).

sofroniewn:
This is very helpful @GenevieveBuckley, I really appreciate the screenshot - helps make everything a little more concrete for me. Once we get the physical coordinates, and basic plugin stuff in, we can put some time towards the multicanvas stuff, which I think will be important for your usage, as I see how grid mode isn't sufficient. Keeping this use case in mind then will be important. It's also great to see you weighing in and helping out with #885!!

> I don't want to lose the ease of use of adding matched control points across images/layers, but I'm sure we can work on making sure that support is also added to napari.

Absolutely, this will make a great plugin! I think many people interested in multimodal registration will be excited about this.

napari/components/dims.py
```diff
@@ -306,6 +329,14 @@ def not_displayed(self):
         """Tuple: Dimensions that are not displayed."""
         return self.order[: -self.ndisplay]
 
+    @property
+    def sliced(self):
+        """int: Dimension that is sliced if embedded."""
+        if self.embedded and self.ndim >= 3:
+            return self.order[-3]
```
alisterburt, line 336: this will need switching to a variable rather than hardcoding

alisterburt (commented):
@sofroniewn - I had a quick look at how this worked (I want to make animations with it) 

What do you think is the main thing stopping this from moving forward? (besides the presence of a million other things to work on, of course :laughing:)

I found one hardcoded variable `dims.sliced` which would need sorting and in general the functionality would need to be exposed in a useful/intuitive way - do we have any idea what this may look like? 

My immediate thought is 
- an checkbox connected to `layer.dims.embedded` 
- a spinbox/combobox for the `layer.dims.sliced` 
these would only be exposed in the 3D viewer mode

I'm assuming there isn't something you fundamentally don't like about what you've done here?

sofroniewn:
> What do you think is the main thing stopping this from moving forward? (besides the presence of a million other things to work on, of course 😆)
> ...
> I'm assuming there isn't something you fundamentally don't like about what you've done here?


I'd say nothing fundamental, though I think we've got better machinery now in place to deal with this (like our world coordinate system) then we did back when this started, and we've improved some of the slider handling.

I think we've also come further along in our ideas of multiscale projection like #1820, so I'd want to step back and think about the overall API and user interactivity, before just cleaning up this PR and merging it.

We've also removed `layer.dims` from each layer, and are in general trying to move to a world where an individual layer don't know how it is sliced (i.e. slicing is something that happens to a layer rather than a slice is a thing that lives on a layer, see #1353). 

For example, do we need an `embedded` property on each layer, or are the global controls we can use instead?

>  I had a quick look at how this worked (I want to make animations with it)

Can you share a little more detail about your particular need. Do you have multiple 3D volumes, 2D? etc. we might be able to get something more simple in faster

alisterburt:
thanks for the links - goldmine of info, especially #1353 !

Happy to take a step back and think about the API and interactivity, I'll add it as a discussion point for the next meeting if there aren't already too many other things :smiley: 

re: the particular use case - I often find myself showing 2D slices when presenting and think this hides the 3D nature of the data. To get around this, I often have a slide like this where I show 2D and 3D renderings side by side...
2d3d

I'd basically like to not have to explain this, having the 2D slice moving through the 3D volume then moving the camerato give the equivalent 2D view basically solves the problem - you immediately get a feel for the scale of the z-steps and the 3D-ness of the data.

jni:
@alisterburt side note: what software did you use to generate that isosurface? It's gorgeous!

alisterburt:
@jni more [ChimeraX](https://www.cgl.ucsf.edu/chimerax/) loveliness! 
specific settings: ambient occlusion (64 evenly distributed point sources) with black silhouettes around the isosurface

edit: just had a quick dig and it turns out ChimeraX is [open source](https://github.com/RBVI/ChimeraX)

jni:
> edit: just had a quick dig and it turns out ChimeraX is open source

Under a rather restrictive license, unfortunately, so we can't make use of the source code:

> Licensee agrees that it will use the Software, and any modifications, improvements, or derivatives to the Software that the Licensee may create (collectively, “Improvements”) solely for internal, non-commercial purposes and shall not distribute or transfer the Software or Improvements to any person or third parties without prior written permission from The Regents

I presume there's an OpenGL recipe somewhere for those rendering settings, though, so maybe we can add this kind of view to vispy/napari???

sofroniewn:
> I presume there's an OpenGL recipe somewhere for those rendering settings, though, so maybe we can add this kind of view to vispy/napari???

There are actually PRs already in progress at vispy that add better lighting with shading to the surface layer, see discussion in https://github.com/vispy/vispy/issues/1665 and https://github.com/vispy/vispy/pull/1463 so if we help get that finished we should get this nice functionality too in napari without having to look at chimeraX

sofroniewn:
It's been a while since this PR was made and we've now got some other vispy ones in progress like #1820 and won't merge this as is, so i will close now. If we want to pick up this conversation again we can go back to #639 the original issue

Issue #574, opened 2019-10-08
Bad 3D rendering of labels


jni:
## 🐛 Bug

I get some bad rendering of labels data in 3D, running the code in this gist:

https://gist.github.com/jni/587d8a946e5e86a8c0d3a7f8bee329cc

with `ipython --gui=qt` then `%run -i rendering.py`

<img width="1200" alt="Screen Shot 2019-10-08 at 12 45 45 pm" src="https://user-images.githubusercontent.com/492549/66361390-59c10b80-e9ca-11e9-9c19-a093758f613c.png">

The issue seems to be that the layer values are interpolated, which causes the crazy fringing with the random colormap. Note that I played around with the interpolation method in the image layers, but the problem does persist after restarting Python.

Additionally, "mip" rendering is a poor choice for labels, because high-value labels appear on top of low-value ones, even when the low labels are in front of the high ones. However, translucent rendering did not fix the fringing problem, so it still looks terrible for me.

## To Reproduce

1. Download https://gist.github.com/jni/587d8a946e5e86a8c0d3a7f8bee329cc to a file named `rendering.py`
2. Launch `ipython --gui=qt` in that directory.
3. Run `%run -i rendering.py`

## Environment

 - napari Version (e.g., 1.0): master
 - OS (e.g., Linux): macOS 10.14.6
 - Python version: 3.7.3
 - Any other relevant information: vispy version 0.6.1

sofroniewn:
I can reproduce this effect on my laptop - both the fringes and the poor order behavior of `mip`. If I switch to `translucent` the order problems go away, but the fringes are terrible, which is likely due to the interpolation @jni mentioned and problems with our colormap

jni:
I think we are probably going to need to mess with the shaders in vispy to fix this one, since as far as I can tell VisPy does not expose the ability to remove interpolation in the Volume Rendering view.

PR #840, opened 2019-12-31
better isosurface rendering with colormaps

sofroniewn:
# Description
This PR significantly improves our isosurface rendering by making it respect our colormaps and getting rid of the light green default. To do this I "vendored" the vispy volume shaders, which results in quite a bit of copied code, but which will give us significant flexibility as we continue to address issues like #574.

The changed lines are relatively minor
```bash
const vec4 u_ambient = vec4(0.2, 0.2, 0.2, 1.0);
```
 gets rid of the green, and 
```bash
                color = $sample(u_volumetex, iloc);
                if (color.g > u_threshold) {
                    color = calculateColor(color, iloc, dstep);
                    gl_FragColor = $cmap(color.g);
```
puts the colormap call in the right place

Here is a gif of the new performance, which I think is quite nice!!
better_iso_surface

Ultimately we will probably want to push these changes to vispy, but I imagine we'll be making a bunch of them over the next chunk of time, so might make sense to wait till we're done.

### DIFF
--- a/napari/_vispy/vispy_image_layer.py
+++ b/napari/_vispy/vispy_image_layer.py
@@ -1,6 +1,6 @@
 import warnings
 from vispy.scene.visuals import Image as ImageNode
-from vispy.scene.visuals import Volume as VolumeNode
+from .volume import Volume as VolumeNode
 from vispy.color import Colormap
 import numpy as np
 from .vispy_base_layer import VispyBaseLayer
--- a/napari/_vispy/volume.py
+++ b/napari/_vispy/volume.py
@@ -47,7 +47,7 @@ varying vec4 v_nearpos;
 varying vec4 v_farpos;
 
 // uniforms for lighting. Hard coded until we figure out how to do lights
-const vec4 u_ambient = vec4(0.2, 0.4, 0.2, 1.0);
+const vec4 u_ambient = vec4(0.2, 0.2, 0.2, 1.0);
 const vec4 u_diffuse = vec4(0.8, 0.2, 0.2, 1.0);
 const vec4 u_specular = vec4(1.0, 1.0, 1.0, 1.0);
 const float u_shininess = 40.0;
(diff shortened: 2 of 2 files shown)


sofroniewn:
I've figured out how to fix #574 by enabling nearest neighbor interpolation in the volume layer - something like 
```python
        self._tex = Texture3D((10, 10, 10), interpolation='nearest')
        self.shared_program['u_volumetex'] = self._tex
```
we'll need a little bit more wrapping to make that settable etc, but the labels now look great in 3D transluencent mode (so no more max intensity projecting)

I'll maybe make a new PR that builds off of this one to implement this

jni:
@sofroniewn 😍😍😍😍😍

This is amazing! And fixing #574... 🤯

I would suggest pushing these to vispy as they happen. On the vispy side it's a smallish change (?), which is much easier to review than an all-encompassing overhaul. Smaller is better with PRs!

jni:
@sofroniewn indeed, to help with *my* review, would it be hard for you to split this into two commits:

- vendoring _vispy volume at a particular git hash
- making the necessary changes to that file

?

I might even temporarily enable merge without squash for this purpose...

sofroniewn:
I'll make a vispy PR now - I think it is quite nice. I've added two commits - one which takes the PR back to the vispy vendored state, and one called `our_changes` which includes the 6 vispy lines changed.

jni:
@sofroniewn wowza! So simple, so great.

I'm happy to do this for you but in case you don't know this, you can use:

```
git rebase -i master better_isosurface_rendering
```

to launch an interactive rebase, and squash the first two commits together, then force push to update the PR. 

Or even simpler, just found through `tldr git-rebase`:

```
git rebase -i HEAD~3
```

to rebase/rearrange/squash the most recent three commits in-place.

sofroniewn:
@jni I didn't know, thanks for the git tips! If you want to do that go-ahead. I just made a PR https://github.com/vispy/vispy/pull/1802 with the change too

tlambert03:
Omg 😍😍

jni:
Good-oh! I've force-pushed, will merge (not squash) once CI passes.

jni (approved):

royerloic:
Fantastic @sofroniewn , I see you are starting to hack the shaders :-) 
I will find some time this year to do some of that myself :-)

PR #841, opened 2020-01-01
Add nearest interpolation mode to volume rendering

### LINKED ISSUE #574: Bad 3D rendering of labels

sofroniewn:
# Description
This PR follows on from #840 (which should be reviewed and merged first) and closes #574 by providing a `nearest` interpolation mode (in addition to default `linear` interpolation mode) for 3D volume rendering. Ultimately we may want to push these changes back to vispy, but for now as we are editing things frequently, I think it is fine to keep them here. Not that this improvement will be compatible with whatever colormap improvements come with #713 as it is only touching the interpolation code. Rendering of our image layers is also unchanged by this PR.

Here is what the labels layer now looks like:
better_3D_labels
which is much nicer!!

### DIFF
--- a/napari/_vispy/vispy_image_layer.py
+++ b/napari/_vispy/vispy_image_layer.py
@@ -5,6 +5,8 @@ from vispy.color import Colormap
 import numpy as np
 from .vispy_base_layer import VispyBaseLayer
 from ..layers.image._constants import Rendering
+from ..layers import Image, Labels
+
 
 texture_dtypes = [
     np.dtype(np.int8),
@@ -101,7 +103,11 @@ class VispyImageLayer(VispyBaseLayer):
         self.node.update()
 
     def _on_interpolation_change(self):
-        if self.layer.dims.ndisplay == 2:
+        if self.layer.dims.ndisplay == 3 and isinstance(self.layer, Labels):
+            self.node.interpolation = 'nearest'
+        elif self.layer.dims.ndisplay == 3 and isinstance(self.layer, Image):
+            self.node.interpolation = 'linear'
+        else:
             self.node.interpolation = self.layer.interpolation
 
     def _on_rendering_change(self):
--- a/napari/_vispy/volume.py
+++ b/napari/_vispy/volume.py
@@ -381,13 +381,13 @@ class Volume(BaseVolume):
         return self._interpolation
 
     @interpolation.setter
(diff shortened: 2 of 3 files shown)


sofroniewn:
I have made the corresponding PR here https://github.com/vispy/vispy/pull/1803

sofroniewn:
@jni I am trying to make this PR only have the changes needed for the interpolation functionality, but instead the diff seems to have everything!! I tried a rebase with master after merging #840, but it seems to have just polluted the git history :-( Can you maybe help? The PR should only be adding about 20 lines in the Volume file instead of the whole thing ....

jni:
@sofroniewn done! I'll merge once tests pass. For the record, what I did was:

- update my master
- fetch your branch
- `git rebase -i master` to do an interactive rebase
- delete the first two commits, which were redundant (they were the commits in #840)
- apply the rebase (which applied cleanly and only contained the small changes)
- force push to your branch.

Everything looks great now. I've renamed "i" to "interp" and you might want to do the same in your PR to VisPy, but other than that all good! :tada: This pair of PRs is just so fantastic for our 3D functionality!

jni (approved):

royerloic:
This is great @sofroniewn !

PR #844, opened 2020-01-02
refactor RangeSlider to accept data range & values.

### LINKED ISSUE #842: make RangeSlider.setValues accept data values (not 0-1)
## 🚀 Feature
In working on #837, I find myself needing to duplicate the logic for setting the values of the contrast limits slider that are defined in the `qt_image_base_layer`:

https://github.com/napari/napari/blob/53d95988fee000f5fa0a1c5e567e09ea05d80635/napari/_qt/layers/qt_image_base_layer.py#L79-L91

That code is due to the fact that the sliders themselves pretty much always have a range from 0-1, and the underlying data needs to get converted to "slider space".  I propose moving that logic inside the range sliders, so that consumers of the RangeSlider can simply set the data as they see it, and the slider figures out how to render itself.  Any objections?  If not, I will probably just do it in #837

tlambert03:
# Description
This would close #842 and make the right-click slider popup that I'm working on in #837 easier.
The behavior of the `RangeSlider` doesn't change at all.. but interacting with the slider from external objects changes such that `slider.setRange()`, `setValues()`, and `setStep()` expect values in "data space" and not in "slider space" (which was always from 0-1.).
 
This makes reusing the slider easier, because you don't need to duplicate the logic to convert from "data range" to "slider range" everywhere you implement the slider.

## Type of change

- [x] Refactor with a couple new methods

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/_qt/layers/qt_image_base_layer.py
+++ b/napari/_qt/layers/qt_image_base_layer.py
@@ -3,7 +3,7 @@ from qtpy.QtGui import QImage, QPixmap
 from qtpy.QtWidgets import QComboBox, QLabel, QSlider
 
 from .. import QHRangeSlider
-from ..utils import qt_signals_blocked
+from ..utils import qt_signals_blocked, connect_model_to_rangeslider
 from .qt_base_layer import QtLayerControls
 
 
@@ -12,9 +12,6 @@ class QtBaseImageControls(QtLayerControls):
         super().__init__(layer)
 
         self.layer.events.colormap.connect(self._on_colormap_change)
-        self.layer.events.contrast_limits.connect(
-            lambda e: self.contrast_limits_slider_update()
-        )
         self.layer.events.gamma.connect(lambda e: self.gamma_slider_update())
 
         comboBox = QComboBox()
@@ -28,16 +25,11 @@ class QtBaseImageControls(QtLayerControls):
 
         # Create contrast_limits slider
         self.contrastLimitsSlider = QHRangeSlider(
-            slider_range=[0, 1, 0.0001], values=[0, 1]
+            self.layer.contrast_limits, self.layer._contrast_limits_range
         )
-        self.contrastLimitsSlider.setEmitWhileMoving(True)
-        self.contrastLimitsSlider.collapsable = False
-        self.contrastLimitsSlider.setEnabled(True)
-
-        self.contrastLimitsSlider.rangeChanged.connect(
-            self.contrast_limits_slider_changed
+        connect_model_to_rangeslider(
+            self.layer, 'contrast_limits', self.contrastLimitsSlider
         )
-        self.contrast_limits_slider_update()
 
         # gamma slider
         sld = QSlider(Qt.Horizontal)
@@ -75,20 +67,6 @@ class QtBaseImageControls(QtLayerControls):
             QImage.Format_RGBA8888,
         )
         self.colorbarLabel.setPixmap(QPixmap.fromImage(image))
-
-    def contrast_limits_slider_changed(self, slidermin, slidermax):
-        valmin, valmax = self.layer._contrast_limits_range
-        cmin = valmin + slidermin * (valmax - valmin)
-        cmax = valmin + slidermax * (valmax - valmin)
-        self.layer.contrast_limits = cmin, cmax
-
-    def contrast_limits_slider_update(self):
-        valmin, valmax = self.layer._contrast_limits_range
-        cmin, cmax = self.layer.contrast_limits
-        slidermin = (cmin - valmin) / (valmax - valmin)
-        slidermax = (cmax - valmin) / (valmax - valmin)
-        with qt_signals_blocked(self.contrastLimitsSlider):
-            self.contrastLimitsSlider.setValues((slidermin, slidermax))
 
     def gamma_slider_changed(self, value):
         self.layer.gamma = value / 100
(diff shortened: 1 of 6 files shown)


napari/_qt/utils.py
```diff
@@ -123,3 +123,31 @@ def qt_signals_blocked(obj):
     obj.blockSignals(True)
     yield
     obj.blockSignals(False)
+
+
+def connect_model_to_rangeslider(layer, attr, slider, initialize=False):
```
sofroniewn: I'm not sure I love this function. I guess you want to reuse it more in #837 which is why you've added it here. I'm curious what @jni thinks?

On the one hand it reduces code duplication, but on the other it is another function to keep track of that doesn't do too much.

sofroniewn: We also don't us this paradigm (i.e functions like this) with our other Qt elements, say the normal slider - I'm curious though if you think we should? In which case I'd be more in favor of this, I'm just not sure it warrants a different approach. That being said I'm very open to keeping it if you think it makes things simpler, and maybe having more convenience functions like this for our other Qt elements would be good too.

sofroniewn (commented):
This looks good to me - though see my question about the new convenience function that has been added.

Also it looks like the range slider doesn't have its own tests. Can you add some that test both the horizontal and vertical range slider in isolation, including testing the new behavior - maybe a `test_qt_range_slider.py` file.

Also curious if @AhmetCanSolak wants to weigh in before merge.

napari/_qt/utils.py
tlambert03: I originally did that because the connection process was more laborious, and it would have made reusing the connections easier (I did want to reuse in #837), but actually now with this refactor, it reduces the connection code enough that i think it's unnecessary.  I can remove it.

tlambert03:
> Can you add some that test both the horizontal and vertical range slider in isolation, including testing the new behavior - maybe a `test_qt_range_slider.py` file.

yeah I'll add some tests.

Also curious to get @AhmetCanSolak's opinions.  One thing that wasn't totally clear to me was the expected behavior for the collapsible feature.  Since we don't seem to use that anywhere, might we want to remove it?  or is there a use case for which we might want to hang on to it?  Currently, I think it's a little broken (or maybe just in this PR)

sofroniewn:
The collapsibility was so that we could have easy transitions between a range and a normal slider that preserved some information so we could toggle between a "max projection" like mode and a slicing like mode. See #7 for more info on the max projection feature. We havn't gotten around to implementing it yet, and now I'm not sure that it wouldn't be easier to have the range slider only be a range slider - so not collapsable, but just swap out the entire slider for one of the default Qt sliders (we now use the scroll bar) when the time is right. It would dramatically simplify the range slider code, and we could keep track of any additional needed information inside the QtDims model, and we wouldn't be reimplementing something Qt does already.

I'd be in favor of a refactor to remove the "collapsable" functionality from the range slider as it would simply that code quite a bit - but curious what @AhmetCanSolak and @royerloic think before we do the refactor. If they agree then I'd do it in this PR too

tlambert03:
OK, added tests as best I could.  There seems to be a bug in `QTest` at the moment that prevents simulating drag/drop motions, so I had to kinda fake mouse moves.  but the main value changes should be tested.

thanks for explaining the collapsible idea.  makes sense.  I think, given how easy/cheap it is to instantiate new sliders, I'd personally just remove that logic and create a totally new QSlider if you wanted to collapse back to the single handle..., and then rebuild the RangeSlider from the model when expanding...  but will wait to hear what @AhmetCanSolak or @royerloic think

napari/_qt/layers/qt_image_base_layer.py
```diff
@@ -28,16 +28,13 @@ def __init__(self, layer):
...
         )
-        self.contrastLimitsSlider.setEmitWhileMoving(True)
-        self.contrastLimitsSlider.collapsable = False
-        self.contrastLimitsSlider.setEnabled(True)
 
-        self.contrastLimitsSlider.rangeChanged.connect(
-            self.contrast_limits_slider_changed
-        )
-        self.contrast_limits_slider_update()
+        set_clim = partial(setattr, self.layer, 'contrast_limits')
```
sofroniewn, line 34: Is there a reason to use the `partial` syntax here? Should we be using it elsewhere too? I'm not so familiar with it (had to do some googling to read more!!)

jni: I'm quite, ahem, partial to it myself! =P It's standard Python but for some reason (partly because Guido van Rossum doesn't much like functional programming, I'm sure), it's not very common.

> Should we be using it elsewhere

I don't think it's mandatory or worth homogenising, but I personally like it. My preference would actually be to use [`toolz.curry`](https://toolz.readthedocs.io/en/latest/curry.html#id1) on everything, but that is *non-*standard, so I restrain myself. =) I have been toying with the idea of authoring a PEP for this.

sofroniewn (approved):
I'm going to approve this. We can merge as is and then do a follow-up PR to refactor the collapsable stuff if we decide to do that, depending on what @royerloic and @AhmetCanSolak say.

napari/_qt/layers/qt_image_base_layer.py
tlambert03: no reason other than personal preference.  i find myself reaching for it a lot with Qt signals, because it pains me to do something like this:
```python
def setter(x):
    self.something = x

widget.signal.connect(setter)
```
(and also, here, there were multiple setters to declare)

tlambert03:
thanks @sofroniewn.  I'm gonna merge so I can move on with the other PRs.

PR #846, opened 2020-01-03
add attenuated MIP mode

### LINKED ISSUE #843: More advanced 3D rendering
## 🚀 Feature
Following up on https://github.com/napari/napari/pull/840#issuecomment-570379878 from @royerloic - right now for 3D arrays we have `translucent`, `mip`, `additive`, and `iso surface` rendering. @royerloic and @AhmetCanSolak mentioned another type of rendering that can be nicer than `mip` and that has one free parameter that you can use to control - i can't remember what it was though. If you could give the name / a link to the rendering equation then I can try implementing it as a new option, which should be easy now that we have a copy of the vispy volume shader code in our repo

sofroniewn:
# Description
This PR will close #843 by adding an attenuated maximum intensity projection mode, where the attenuation is controlled by a decay exponent. This PR still needs doc string edits, and right now I re-use the `iso_threshold` parameter. I should probably add a new parameter `attenuation_rate` (or something) to keep it clear whats going on. I'll do that shortly, but I wanted to let people see how it looks, which I think is quite nice :-)

attenuated_mip

I'm also getting
```bash
INFO: Not setting value for variable float u_threshold; uniform is not active.
INFO:vispy:Not setting value for variable float u_threshold; uniform is not active.
```
warning messages all over, that I would like to make disappear (problem is not new to this PR, but now there are many more of them)

### DIFF
--- a/napari/_qt/layers/qt_image_layer.py
+++ b/napari/_qt/layers/qt_image_layer.py
@@ -49,8 +49,8 @@ class QtImageControls(QtBaseImageControls):
         sld.valueChanged[int].connect(
             lambda value=sld: self.changeisoThreshold(value)
         )
-        self.isoThesholdSlider = sld
-        self.isoThesholdLabel = QLabel('iso threshold:')
+        self.isoThresholdSlider = sld
+        self.isoThresholdLabel = QLabel('iso threshold:')
 
         sld = QSlider(Qt.Horizontal)
         sld.setFocusPolicy(Qt.NoFocus)
@@ -73,8 +73,8 @@ class QtImageControls(QtBaseImageControls):
         self.grid_layout.addWidget(self.contrastLimitsSlider, 1, 1, 1, 2)
         self.grid_layout.addWidget(QLabel('gamma:'), 2, 0)
         self.grid_layout.addWidget(self.gammaSlider, 2, 1, 1, 2)
-        self.grid_layout.addWidget(self.isoThesholdLabel, 3, 0)
-        self.grid_layout.addWidget(self.isoThesholdSlider, 3, 1, 1, 2)
+        self.grid_layout.addWidget(self.isoThresholdLabel, 3, 0)
+        self.grid_layout.addWidget(self.isoThresholdSlider, 3, 1, 1, 2)
         self.grid_layout.addWidget(self.attenuationLabel, 3, 0)
         self.grid_layout.addWidget(self.attenuationSlider, 3, 1, 1, 2)
         self.grid_layout.addWidget(QLabel('colormap:'), 4, 0)
@@ -103,7 +103,7 @@ class QtImageControls(QtBaseImageControls):
 
     def _on_iso_threshold_change(self, event):
         with self.layer.events.iso_threshold.blocker():
-            self.isoThesholdSlider.setValue(self.layer.iso_threshold * 100)
+            self.isoThresholdSlider.setValue(self.layer.iso_threshold * 100)
 
     def changeAttenuation(self, value):
         with self.layer.events.blocker(self._on_attenuation_change):
@@ -133,11 +133,11 @@ class QtImageControls(QtBaseImageControls):
         if isinstance(rendering, str):
             rendering = Rendering(rendering)
         if rendering == Rendering.ISO:
-            self.isoThesholdSlider.show()
-            self.isoThesholdLabel.show()
+            self.isoThresholdSlider.show()
+            self.isoThresholdLabel.show()
         else:
-            self.isoThesholdSlider.hide()
-            self.isoThesholdLabel.hide()
+            self.isoThresholdSlider.hide()
+            self.isoThresholdLabel.hide()
         if rendering == Rendering.ATTENUATED_MIP:
             self.attenuationSlider.show()
             self.attenuationLabel.show()
@@ -147,8 +147,8 @@ class QtImageControls(QtBaseImageControls):
 
     def _on_ndisplay_change(self, event):
         if self.layer.dims.ndisplay == 2:
-            self.isoThesholdSlider.hide()
-            self.isoThesholdLabel.hide()
+            self.isoThresholdSlider.hide()
+            self.isoThresholdLabel.hide()
             self.attenuationSlider.hide()
             self.attenuationLabel.hide()
             self.renderComboBox.hide()
--- a/napari/_qt/layers/qt_image_layer.py
+++ b/napari/_qt/layers/qt_image_layer.py
@@ -47,7 +47,7 @@ class QtImageControls(QtBaseImageControls):
         sld.setSingleStep(1)
         sld.setValue(self.layer.iso_threshold * 100)
         sld.valueChanged[int].connect(
-            lambda value=sld: self.changeisoThreshold(value)
+            lambda value=sld: self.changeIsoThreshold(value)
         )
         self.isoThresholdSlider = sld
         self.isoThresholdLabel = QLabel('iso threshold:')
@@ -97,7 +97,7 @@ class QtImageControls(QtBaseImageControls):
         self.layer.rendering = text
         self._toggle_rendering_parameter_visbility()
 
-    def changeisoThreshold(self, value):
+    def changeIsoThreshold(self, value):
         with self.layer.events.blocker(self._on_iso_threshold_change):
             self.layer.iso_threshold = value / 100
 
--- a/napari/_qt/layers/qt_image_layer.py
+++ b/napari/_qt/layers/qt_image_layer.py
@@ -47,7 +47,7 @@ class QtImageControls(QtBaseImageControls):
         sld.setSingleStep(1)
         sld.setValue(self.layer.iso_threshold * 100)
         sld.valueChanged[int].connect(
-            lambda value=sld: self.changeIsoTheshold(value)
+            lambda value=sld: self.changeisoThreshold(value)
         )
         self.isoThesholdSlider = sld
         self.isoThesholdLabel = QLabel('iso threshold:')
@@ -97,7 +97,7 @@ class QtImageControls(QtBaseImageControls):
         self.layer.rendering = text
         self._toggle_rendering_parameter_visbility()
 
-    def changeIsoTheshold(self, value):
+    def changeisoThreshold(self, value):
         with self.layer.events.blocker(self._on_iso_threshold_change):
             self.layer.iso_threshold = value / 100
 
--- a/napari/_qt/layers/qt_image_layer.py
+++ b/napari/_qt/layers/qt_image_layer.py
@@ -109,7 +109,7 @@ class QtImageControls(QtBaseImageControls):
         rendering = self.layer.rendering
         if isinstance(rendering, str):
             rendering = Rendering(rendering)
-        if rendering == Rendering.ISO:
+        if rendering in [Rendering.ISO, Rendering.ATTENUATED_MIP]:
             self.isoThesholdSilder.show()
             self.isoThesholdLabel.show()
         else:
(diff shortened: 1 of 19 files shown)


sofroniewn:
Look how much better this example gets!!
attenuated_mip2

Note that for this one it would be really nice to have some crop sliders so we could then start seeing the brain!!

jni:
@sofroniewn can we see the AOLLSM-3D dataset like this? It's wonderful! =D

sofroniewn:
@jni @tlambert03 - I have an API question about this - I am torn. Right now we could renamed `iso_threshold` -> something generic like `3D rendering parameter` and us the same variable for both iso_surface rendering threshold and our attenuation rate, and any future single parameters that 3D rendering needs.

Alternatively we leave `iso_threshold` and we add `attenuation` or `attenuation_rate` as a new parameter with its own setters / sliders etc.

I lean towards the later option as I think it is clearer and safer, but I worry that if this keeps happening the API will expand rapidly.

Curious what people's thoughts are

royerloic:
Awesome @sofroniewn !

I think there is no remedy to have each parameter treated separately, because eventually we might need more than one 'generic parameter' and then the code will become obscure in terms of what these parameters are doing at any given time. Having the API expand is fine, perhaps we need to have the possibility to hide some of the sliders because having lots of sliders  do nothing depending on the mode will get very confusing quickly... 

Another note: it feels that the volume rendering gets quite a bit dimmer when toggling the attenuation mode. This might be caused by simply having the brighter regions deeper in the volume and thus inaccessible, but it could also be another issue.

sofroniewn:
@royerloic 
> each parameter treated separately

I will go with that approach - agree it is the better one. Thanks!

> volume rendering gets quite a bit dimmer when toggling the attenuation mode

I had noticed this too, but wasn't quite sure what to do about it. It should be fixed though. The problem is that as the attenuation rate becomes large everything in the image starts getting attenuated and so everything becomes a little darker and we never use the full dynamic range available to us. Adjust clims doesn't help because that applies before the attenuation. I'll have a look at your mathematics and see if there is a difference, or see if there is a clean way to quickly renormalize everything so there isn't this global dimming. Any advice appreciated

royerloic:
About the attenuation: one possibility is to use the attenuated intensity profile along the ray path only as an -indicator- of which value along the ray should be picked. This way, you separate the decision process of which position along the ray to sample based on the attenuated max, from the actual value picked (non-attenuated) that gets displayed. This would not introduce any attenuation on the final image. I am curious how this would look like...  To implement this, just replace the max with a if statement for sampling or better use a ternary operator (can often be optimised to avoid branching on GPU: condition ? a : b)

sofroniewn:
> one possibility is to use the attenuated intensity profile along the ray path only as an -indicator- of which value along the ray should be picked.

I like this, I think it looks quite nice
attenuated_mip3

I think this is almost ready to review - I'd like to fix the
```bash
INFO: Not setting value for variable float u_threshold; uniform is not active.
INFO:vispy:Not setting value for variable float u_threshold; uniform is not active.
```
and then think it's good to go. I also need to think about the best way to get this into vispy, but that can happen in followup PRs to vispy directly

royerloic:
Great! There are a few things we could still add, but first let's get this out. I would be interested in doing a  PR with some more features and tweaks... This is too much fun... So maybe it is best to wait a bit before we send ur changes upstream to vispy...

Some ideas of improvements going forward that I would love to play with:

i) Have gamma be computed on the fly in the shader instead of outside, much faster especially dfor time series (computing gamma in the shader is for free, can be done at the very end outside of the loop)
ii) dithering of ray sampling to avoid stepping artefacts
iii) variable 'render quality'/'render speed' trade-off -- which is essentially the amount of sampling.
iv) loop unrolling for speedup as done in CV (necessary when doing iii )
v) box / grid around volume 
vi) two-pass tracing, first pass is coarse, second pass refines it -- not useful for all modes. Gives better quality and better performance.
vii) more advanced render modes that introduce more physics to the rendering such as subsurface scattering, volumetric lighting, screen-space-ambient-occlusion (ssao). This article has a nice overview of advanced volume rendering techniques: https://developer.download.nvidia.com/books/HTML/gpugems/gpugems_ch39.html


Not everything needs to be exposed in the ui, but power- users could still play with all features...

tlambert03:
don't have any strong/important comments, but just took a look and really like it!  It's particularly good with medical type images as you've shown here.
As mentioned above, it would be nice to figure out why everything gets dimmer.

obviously, this is dataset dependent, but for a couple fluorescence images I tried (like volker's ovarioles) , I kinda prefer the original shader you used in the first commit to the revised "decision" shader.   the new shader ends up with very sharp cutoffs (even at low thresholds) that make for discontinuities in the rendering:

Screen Shot 2020-01-03 at 5 06 18 PM

whereas the previous shader had a true depth-dependent attenuation appearance:
Screen Shot 2020-01-03 at 5 03 38 PM

that of course makes the image much dimmer again (in this case I just added a multiplier to `after_loop` but I'm sure that's not a good general solution).  Anyway, feel free to ignore that if you feel like it's not really the intended use case.

sofroniewn:
I think the "true depth" appearance is important. I've gone back to the old solution with a multiplier (could be in main loop or after / doesn't really matter, but put it after). Exact choice was somewhat arbitrary though what I went for seems to work well for a bunch of stuff I just looked at. Curious how it looks if you try it now @tlambert03 

attenuated_mip4

sofroniewn:
This PR is now ready for review. I set the vispy logger to be at `Warning` level, it is at `Info` by default, and no need for our users to see those openGL related messages

@royerloic - those comments are great. I will open a new issue for them

royerloic:
> @royerloic - those comments are great. I will open a new issue for them

One thing to keep an eye on, and a source of inspiration:
https://www.allencell.org/pathtrace-rendering.html

I will study what they are doing there and see if we can do something
to bring a bit more photo-realism to our 3D volumetric rendering while 
retaining speed... -- yet in some cases, being able to make an explicit
trade-off between photo-realism and speed could be acceptable as long 
as it is not a default.

AhmetCanSolak:
> https://www.allencell.org/pathtrace-rendering.html

Hi @royerloic . I checked their implementation before, it is a PBR-like rendering. The quality comes really with time cost. I think fast interactivity we have with napari is amazing and it could be possibly a good idea to have pbr-like rendering selections optional? or as a plugin maybe?

napari/_qt/layers/qt_image_layer.py
```diff
@@ -46,22 +47,36 @@ def __init__(self, layer):
         sld.setSingleStep(1)
         sld.setValue(self.layer.iso_threshold * 100)
         sld.valueChanged[int].connect(
-            lambda value=sld: self.changeIsoTheshold(value)
+            lambda value=sld: self.changeisoThreshold(value)
```
tlambert03: while you're at it, is there any reason not to just make it this?
```python
sld.valueChanged.connect(self.changeisoThreshold)
```

napari/_qt/layers/qt_image_layer.py
```diff
@@ -46,22 +47,36 @@ def __init__(self, layer):
...
+
+        sld = QSlider(Qt.Horizontal)
+        sld.setFocusPolicy(Qt.NoFocus)
+        sld.setMinimum(0)
+        sld.setMaximum(200)
+        sld.setSingleStep(1)
+        sld.setValue(self.layer.attenuation * 100)
+        sld.valueChanged[int].connect(
+            lambda value=sld: self.changeAttenuation(value)
+        )
```
tlambert03, line 63: same here:  how about just `sld.valueChanged.connect(self.changeAttenuation)`?

napari/_vispy/vispy_image_layer.py
```diff
@@ -36,7 +36,10 @@ def __init__(self, layer):
         )
         self.layer.events.gamma.connect(lambda e: self._on_gamma_change())
         self.layer.events.iso_threshold.connect(
-            lambda e: self._on_iso_threshold_change()
+            lambda e: self._on_threshold_change()
```
tlambert03, line 39: along the lines of the previous comments, but slightly different... not sure what your personal preference here is (and I know this is older code), but if there's a method like this that is more or less designed to be a vispy event handler, even if it doesn't use the event object, I think it'd be nice if we added `event=None` to the method so that we don't have to catch the event and discard it with a lambda.  so this would be:
```python
self.layer.events.iso_threshold.connect(self._on_threshold_change)
...
def _on_threshold_change(self, event=None):
...
```
(same for `self.layer.events.attenuation.connect` below)

napari/_qt/layers/qt_surface_layer.py
```diff
@@ -9,7 +9,7 @@ def __init__(self, layer):
         # grid_layout created in QtLayerControls
         # addWidget(widget, row, column, [row_span, column_span])
         self.grid_layout.addWidget(QLabel('opacity:'), 0, 0)
-        self.grid_layout.addWidget(self.opacitySilder, 0, 1, 1, 2)
+        self.grid_layout.addWidget(self.opacitySlider, 0, 1, 1, 2)
```
tlambert03, line 12: nice catch on all these! 😂

napari/layers/image/_constants.py
```diff
@@ -36,6 +36,9 @@ class Rendering(StringEnum):
               the result is opaque.
             * mip: maxiumum intensity projection. Cast a ray and display the
               maximum value that was encountered.
+            * attenuated_mip: attenuated maxiumum intensity projection. Cast a
+              ray and attenuate values based on integral of encountered values,
+              display the maximum value that was encountered after attenuation.
```
tlambert03, line 41: would it be appropriate to add a "qualitative hint" here, like "(makes nearer objects appear more prominent)" or something like that?

tlambert03 (approved):
looks good to me @sofroniewn!  Really nice work here, and it was a good example for me on shaders.  I think what you settled on for the brightness fix in the shader is good and should work in most cases (can always revisit later).  I left a couple minor comments that don't really pertain directly to this PR, just code you were already editing...

napari/_vispy/vispy_image_layer.py
jni: 👍 on this change, which I campaigned for originally. =)

tlambert03: @sofroniewn, curious to hear what you think.  if you're for it, another option is to ignore the change here, and I can follow up with a PR to alter that pattern for uniformity across the repo.  On quick check, there are roughly 35 places where the `events.__.connect(lambda e: self._do_something())` pattern is used.  And another 25 or so more qt signal connections that could be simplified in one way or another

sofroniewn: That's a lot! I'm absolutely fine with the changes being made, I remember @jni mentioning this and I don't have objections, but given the magnitude of the change I think a follow-up dedicated PR is warranted

tlambert03: cool, may have a couple questions here and there, but will mark them inline if I do. feel free to ignore here

napari/layers/image/_constants.py
sofroniewn: done

sofroniewn:
I have added the qualitative comment that @tlambert03 suggested, but I have not done any of the `lambda` refactors. As noted above given the magnitude of the problem I'd like that to be a follow-up PR, that maybe @tlambert03 can work on after this gets merged.

I will merge after the tests pass. We can always revisit the details of the shader later, and it will be nice to have this in

napari/_qt/layers/qt_image_layer.py
sofroniewn: as noted below will defer for followup PR

napari/_qt/layers/qt_image_layer.py
sofroniewn: as noted below will defer for followup PR

royerloic:
Looks great! Let's get this in!

PR #850, opened 2020-01-07
Make layers hashable and comparable

tlambert03:
# Description
Building on #686, this PR adds `__hash__` and `__eq__` methods to the base layer, making it easier to do equivalence checks with `layer1 == layer2`?  All it really adds is a utility `recursive_hash` function for hashing nested dictionaries and numpy arrays, and calls that on @sofroniewn's new `_get_state()` dicts.  I guess the main questions this PR raises are:

- *how would we actually use it?*  If I'm not mistaken, one of the reasons for making dictionaries of layer properties was its potential for remote sync.  In that case, being able to quickly ask whether two layers are equal is a nice way to know whether update is necessary
- *how do we define equality?* Currently, two layers would need to have the same `id` to be equal, this PR relaxes that to simply having equivalent output from `get_state`.  Is that what we want?  do we need to add more properties for equivalence check?  I suppose that goes back to question number 1. (I'm fine with the current approach for now).
- *how to handle numpy arrays*.  If we are to use this to know whether a synced viewer needs updating, it seems like we'd need to include the data arrays in the hash.  That can get hairy though as the array gets bigger, and it's unclear how to do it with dask.  [xxhash](https://github.com/Cyan4973/xxHash) (via its [python binding](https://github.com/ifduyue/python-xxhash)) can hash the bytes of a moderate sized in-ram array really fast, but it brings a new dependency.  I implement it with a fallback in this PR, but if we want the hash to actually be useful, we kinda need to use the same hashing method on all installations, so it would need to be non-optional.  (the stdlib strategies either takes a small subset of the array or can quickly become many seconds long).  (see [this SO thread](https://stackoverflow.com/questions/16589791/most-efficient-property-to-hash-for-numpy-array)).

## Type of change

# References
https://stackoverflow.com/questions/16589791/most-efficient-property-to-hash-for-numpy-array
https://stackoverflow.com/a/8714242/1631624

# How has this been tested?

- [x] the test suite I added makes sure layers can be compared correctly and that arrays are included correctly in the hash.

## Final checklist:

### DIFF
--- a/napari/layers/base/base.py
+++ b/napari/layers/base/base.py
@@ -10,6 +10,7 @@ from ...components import Dims
 from ...utils.event import EmitterGroup, Event
 from ...utils.keybindings import KeymapMixin
 from ...utils.status_messages import status_format, format_float
+from ...utils.misc import recursive_hash
 
 
 class Layer(KeymapMixin, ABC):
@@ -434,6 +435,13 @@ class Layer(KeymapMixin, ABC):
     @abstractmethod
     def _get_state(self):
         raise NotImplementedError()
+
+    def __eq__(self, other):
+        # check ids first to avoid more expensive hash if possible
+        return (id(self) == id(other)) or (hash(self) == hash(other))
+
+    def __hash__(self):
+        return recursive_hash(self._get_state(), skip_arrays=False)
 
     @property
     def thumbnail(self):
--- a/napari/layers/tests/test_serialize.py
+++ b/napari/layers/tests/test_serialize.py
@@ -39,3 +39,40 @@ def test_attrs_arrays(Layer, data, ndim):
             )
         else:
             assert np.all(getattr(layer, prop) == getattr(new_layer, prop))
+
+
+@pytest.mark.parametrize('Layer, data, ndim', layer_test_data)
+def test_layer_hash(Layer, data, ndim):
+    """Test that layers can be compared.
+
+    Layers with the same properties evaluate as equal (even if they are
+    different objects).
+    """
+    layer1 = Layer(data)
+    layer2 = Layer(data)
+    assert id(layer1) != id(layer2)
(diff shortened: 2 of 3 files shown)


sofroniewn:
Thanks for getting this going, I think this will be useful for remote syncing / sharing state and diffing etc. I think equivalent get_state is good for equality, though curious what @jni thinks.

> how to handle numpy arrays.

I'm not familiar with the `xxhash`, but agree we're sort of all in or not. Looking at dask's [from_array](https://docs.dask.org/en/latest/array-api.html) it seems like they do some hashing that can be customized and name 3 potential choices. I'm not familiar with their relative trade-offs

> By default, hash uses python’s standard sha1. This behaviour can be changed by installing cityhash, xxhash or murmurhash. If installed, a large-factor speedup can be obtained in the tokenisation step. Use name=False to generate a random name instead of hashing (fast)

jni:
My instinct is that we should not hash numpy arrays. Let's use a random name and provide users the ability to update the "hash" when they modify the data. (e.g. "on_data_update" could update the "hash" also.)

tlambert03:
if we're not worried about being strict about hashing arrays, one quick option is to just use `hash(str(arr))`.    That little string representation is as good of a name as any, and hashing it takes about 1.2 ms on my system, regardless of the size of the array.

sofroniewn:
Curious if we might want to revisit this - see https://github.com/napari/napari/pull/1374#issuecomment-659565241

> if we're not worried about being strict about hashing arrays, one quick option is to just use hash(str(arr)). That little string representation is as good of a name as any, and hashing it takes about 1.2 ms on my system, regardless of the size of the array.

If we just go with this approach and not worry about the arrays for now then this should be pretty easy to wrap up. Could be a good first issue for a new contributor, in which case I'd flag it as such. Thoughts @tlambert03

PR #851, opened 2020-01-07
[WIP] serialize viewer

sofroniewn:
# Description
This PR will leverage #686 to serialize the viewer by adding a `viewer._get_state` command. It's a little unclear to me how to proceed and this is very much a WIP right now. One guiding principle is that this should state should capture all the information needed to ensure that what two people are looking at in the data canvas is identical.

One question is - should I add a `_get_state` to our Dims model that can be leveraged by the Viewer `_get_state`? I lean yes, but maybe it isn't needed.

One problem right now is that we need `camera` information to make the state be able to reproduce the view from one person to another, but that camera info isn't stored in our model, and the current separation between vispy and the viewer makes it a little hard, to put it there, though we could if we overwrote some of vispy's camera stuff. If @royerloic @jni think it's important I can.

I'm also not sure how to think about testing this / the API. For the layers there was the very simple guiding principle that if it was in the state dict then it was a documented keyword arg to the layer. Do we want that to be the case for the viewer? That maybe fine, but then what about all the dims properties? They are currently on the viewer so maybe its ok. If we did that then right now we'd add `dims_points`, `theme`, `metadata`, `camera`, all to the viewer api. Is that ok? Should `layers` also be a keyword arg to viewer where you can pass it a list of layer states? Should `dims` be a keyword arg where you can pass it a dims dictionary?

Do people like having `napari: True` and the napari version info in there? Maybe that was more needed when I was playing around with #511 and those don't belong in a model dict, but can be added elsewhere.

I'm leaning towards doing the necessary stuff to get the 2D and 3D camera info into our model in a nice way that is vispy independent, and then adding keyword args to Viewer, so that there is a 1-1 mapping between the state dictionary and the init and a full roundtrip is possible, i.e.

```python
state_1 = viewer_1._get_state()
viewer_2 = Viewer(**state_1)
state_2 = viewer_2._get_state()
assert state_1 == state_2
assert viewer_1 == viewer_2 # using the hashing like in #850  
```

How do people feel about that as a guiding principle?

## Type of change

### DIFF
--- a/napari/viewer.py
+++ b/napari/viewer.py
@@ -8,6 +8,10 @@ from napari._qt.qt_update_ui import QtUpdateUI
 from ._qt.qt_main_window import Window
 from ._qt.qt_viewer import QtViewer
 from .components import ViewerModel
+from ._version import get_versions
+
+__version__ = get_versions()['version']
+del get_versions
 
 
 class Viewer(ViewerModel):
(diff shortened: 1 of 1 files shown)


sofroniewn:
We're now using Pydantic for out viewermodel and so will get serialization through it!

PR #853, opened 2020-01-07
in bindings logic, check if generator, not generator function

kne42:
# Description
Change usages of `inspect.isgeneratorfunction` to `inspect.isgenerator` in the logic of mouse and key bindings.

This makes it significantly easier to use helper functions when writing bindings:
```python
def foo_helper(bar):
    print(bar)
    yield
    print(bar)


def foo_42():
    return foo_helper(42)


>>> inspect.isgeneratorfunction(foo_helper)
True
>>> inspect.isgeneratorfunction(foo_42)
False

>>> inspect.isgenerator(foo_helper(42))
True
>>> inspect.isgenerator(foo_42())
True
```

### DIFF
--- a/napari/_qt/qt_viewer.py
+++ b/napari/_qt/qt_viewer.py
@@ -395,7 +395,7 @@ class QtViewer(QSplitter):
         func = parent.keymap[comb]
         gen = func(parent)
 
-        if inspect.isgeneratorfunction(func):
+        if inspect.isgenerator(gen):
             try:
                 next(gen)
             except StopIteration:  # only one statement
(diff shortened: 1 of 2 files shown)


sofroniewn (approved):
Seems like a fine choice to me, will see if @jni wants to comment and if not will then merge tomorrow

jni (approved):

kne42:
TIL that this can actually be achieved in native Python with the [`yield from`](https://www.python.org/dev/peps/pep-0380/) syntax.

jni:
@kne42 can you elaborate on this intriguing statement? =D

PR #854, opened 2020-01-08
Add camera model

### LINKED ISSUE #852: camera model
## 🚀 Feature
It's becoming a problem in both #851 (viewer serialization) and #780 (animation making) that the camera state is not part of our model - see in particular this comment from @royerloic https://github.com/napari/napari/issues/448#issuecomment-570745596.

I'd like to fix this, but I am unsure if we should have an explicit `Camera` model which can either be a 2D or 3D camera or should we just have more properties on the viewer, maybe beginning `camera_`

One option would be
```python
viewer.camera_dim = 2 # or 3
viewer.camera_center = [0, 0] # or [0, 0, 0]
viewer.camera_size = [h, w] # right now for vispy in 3D it is a single number - the scale factor
viewer.camera_angle = [a, b, c, d] # right now for vispy in 3D it is a 4 vector quaternion
```
OR with a camera object
```python
viewer.camera.dim = 2 # or 3
viewer.camera.center = [0, 0] # or [0, 0, 0]
viewer.camera.size = [h, w] # right now for vispy in 3D it is a single number - the scale factor
viewer.camera.angle = [a, b, c, d] # right now for vispy in 3D it is a 4 vector quaternion
```

We'll need to intercept calls from the vispy camera to update these properties whenever the camera moves, which I think will be fast, and setup and event system to update the vispy camera when these get updated.

What do people think? @jni @royerloic @tlambert03. I lean towards us having our own very minimal camera object. I could also serve us well when we think about supporting multiple cameras, say for ortho views etc as is being discussed in #760

sofroniewn:
# Description
This PR will close #852 and complete our model view separation as desired by @royerloic, by dealing with the last remaining issue the camera!

It adds a very minimal camera model object and then sets up bidirectional synchronization between the camera model and the vispy cameras - now using the same model regardless of if we are doing 2D or 3D rendering, which I think is a big plus!

This will make things like the viewer serialization #851 and the animation support #780 much nicer too.

Happy for feedback. I might be able to improve some of the tests for our camera angles in 3D, but I tried to add pretty good tests at various levels.

## Type of change

# How has this been tested?

- [x] tests added for camera model, viewer model, and vispy stuff

## Final checklist:


napari/components/tests/test_camera.py
```diff
@@ -0,0 +1,38 @@
...
+
+    scale = 200
+    camera.scale = scale
+    assert camera.scale == scale
+
+    angles = (20, 90, 45)
+    camera.angles = angles
+    assert camera.angles == angles
+
+    center = (20, 45)
```
kevinyamauchi: I think it would be nice if there was a test for the `ValueError` when `len(center) not in [2, 3]`

napari/_qt/qt_viewer.py
```diff
@@ -414,6 +419,29 @@ def on_key_release(self, event):
     def on_draw(self, event):
         """Called whenever drawn in canvas. Called for all layers, not just top
         """
+        if isinstance(self.view.camera, ArcballCamera):
+            self.viewer.camera._center = self.view.camera.center[::-1]
+            self.viewer.camera._scale = self.view.camera.scale_factor
+
+            # Do conversion from quaternion representation to euler angles
+            q = self.view.camera._quaternion
+            angles = (
```
kevinyamauchi: Thematically, it feels like the conversion from quaternions to euler angles should be a util function. However, it doesn't seemed to be used anywhere else right now, so maybe it's fine as is. That being said, I think it would be nice if it had test coverage and that might be easier to do as a standalone function.

napari/_qt/qt_viewer.py
```diff
@@ -414,6 +419,29 @@ def on_key_release(self, event):
     def on_draw(self, event):
         """Called whenever drawn in canvas. Called for all layers, not just top
         """
+        if isinstance(self.view.camera, ArcballCamera):
+            self.viewer.camera._center = self.view.camera.center[::-1]
+            self.viewer.camera._scale = self.view.camera.scale_factor
+
+            # Do conversion from quaternion representation to euler angles
+            q = self.view.camera._quaternion
```
kevinyamauchi: This may be a nitpick, but I find it a bit confusing that we have both `self.viewer.camera` and `self.view.camera` in the same function.

napari/_vispy/tests/test_vispy_camera.py
```diff
@@ -0,0 +1,89 @@
+import numpy as np
+from napari import Viewer
+from vispy.scene import PanZoomCamera, ArcballCamera
+
+
+def test_camera(qtbot):
+    """Test vispy camera model interaction."""
+    viewer = Viewer()
+    view = viewer.window.qt_viewer
```
kevinyamauchi: I don't think this is a big deal, but if you wanted to get change all of the `view.view.*` to `view.*`, you could probably do 
```suggestion
    view = viewer.window.qt_viewer.view
```

napari/components/camera.py
```diff
@@ -0,0 +1,98 @@
...
+
+class Camera:
+    """Camera object modeling position and view of the camera.
+
+    Parameters
+    ----------
+    center : 2-tuple or 3-tuple
+        Center point of camera view for 2D or 3D viewing.
+    scale : float
+        Zoom level.
```
kevinyamauchi: Can you expand on this a bit? In general, I find the `scale` parameters in `napari` to be confusing.

kevinyamauchi (commented):
Nice job, @sofroniewn ! I made a few minor comments. I can take a closer look tomorrow.

royerloic:
Just had a look at the new camera class, seems right to me, pretty straightforward. @jni ?

napari/_qt/qt_viewer.py
sofroniewn: hmm, its not ideal, but not sure how to improve - maybe rename the `view` - this could come with the refactoring described in #855

napari/_qt/qt_viewer.py
sofroniewn: Ideally I'd actually put it + tests into the vispy and it's Quaternion object, which already has methods that generate rotation matrix / axis angle representations - https://github.com/vispy/vispy/blob/master/vispy/util/quaternion.py

I can do that in a vispy PR in the future, for now I will make standalone + add tests

sofroniewn: done

napari/_vispy/tests/test_vispy_camera.py
sofroniewn: did some renaming to try and improve

napari/components/camera.py
sofroniewn: yeah, this might need more attention, but have added something

sofroniewn: linking to #763 too

napari/components/tests/test_camera.py
sofroniewn: done

sofroniewn:
I have made the first round of requested changes from @kevinyamauchi and improved the testing of the euler angle stuff, including some nice round trip tests. I think this is now ready to go

kevinyamauchi (approved):
This looks great to me. Thanks, @sofroniewn !

sofroniewn:
Thanks @kevinyamauchi I will wait if @jni wants to take a look and then plan on merging sometime tomorrow

napari/_qt/qt_viewer.py
```diff
@@ -290,15 +291,19 @@ def _on_cursor(self, event):
...
+            self.view.camera.center = self.viewer.camera.center[::-1]
+            self.view.camera.scale_factor = self.viewer.camera.scale
         else:
             # Assumes default camera has the same properties as PanZoomCamera
-            self.view.camera.rect = event.rect
+            corner = np.subtract(
+                self.viewer.camera.center, self.viewer.camera.scale / 2
+            )
+            rectangle = tuple(corner[::-1]) + (self.viewer.camera.scale,) * 2
+            self.view.camera.rect = rectangle
```
jni: What's going on in the above lines? It's pretty complicated logic. Why are coordinates now being inverted that weren't before?

napari/_vispy/tests/test_vispy_camera.py
```diff
@@ -0,0 +1,100 @@
...
+    viewer.window.qt_viewer.on_draw(None)
+    assert viewer.camera.center == (8, -2, 12)
+    assert viewer.camera.scale == 20
+
+    # Update angle and check roundtrip is correct
+    angles = (12, 53, 92)
+    q = Quaternion.create_from_euler_angles(*angles, degrees=True)
+    vispy_view.camera._quaternion = q
+    viewer.window.qt_viewer.on_draw(None)
+    np.testing.assert_allclose(viewer.camera.angles, angles)
```
jni: Can this be broken down into smaller unit tests?

napari/components/tests/test_camera.py
```diff
@@ -0,0 +1,36 @@
...
+    scale = 300
+    angles = (-20, 10, -45)
+    camera.update(center=center, scale=scale, angles=angles)
+    assert camera.ndisplay == 2
+    assert camera.center == center
+    assert camera.scale == scale
+    assert camera.angles == angles
```
jni: As above: can we write smaller, more self-contained tests that test one or a few aspects of a class, rather than the full suite in one function? This makes it much easier to pinpoint when tests fail *why* they failed.

jni (changes requested):
I've made a few notes where the code was unclear to me. Overall, the purpose of this PR is unclear in the code. Perhaps the note from the PR about why the camera class exists should go in a Notes section in the class docstring?

The coordinate system used should also be explained, especially where the coordinate order is inverted. I kinda get the impression that this might need to wait on world coordinates, or that this will need modification in that PR, so I'd like to make sure we are very clear on *which* coordinate system is being used where...

sofroniewn:
I've just update this to include all the changes up to an including #1539. This still needs some update to use our newer nomenclature and some fixes. I'll likely wait until after #1539 merges, but just wanted to bump this along as it will be what I work on next.

napari/_qt/qt_viewer.py
sofroniewn: These are conversions from numpy to vispy ordering, i've added a comment.

sofroniewn:
This PR is coming along nicely, incorporating lots of recent changes to our coordinate system. There's one failing screenshot test, but before fixing I have some API questions.

This PR would introduce a `Camera` object that represents things in `WorldSlice` coordinates. We already have a `Dims` that actually represents things in what I might call a `WorldNotSlice` coordinates (but actually has a full `World` representation in places too).

There's definitely an opportunity now to simplify/ clarify some things that have been confusing for a long time. At the crux of it the `Dims` should represent the Sliders which corresponds to axes that are not displayed, and the `Camera` should represent the camera which corresponds to axes that are displayed. Given that our next release is going to be `0.4.0` we could make some breaking changes to clean things up, but I'm wondering what the API should look like. One option would be to add a new `Sliders` object inside of `Dims` to handle all the info for the sliders and put the `Camera` inside the `Dims` too. Some example API for the new system might be

```python
viewer.dims.camera.center # stores center of FOV for camera for displayed dimensions (world slice)
viewer.dims.sliders.position # stores position of sliders for not-displayed dimensions (world not sliced)
viewer.dims.position # combines position of sliders and center of FOV of camera into world coordinates
viewer.dims.displayed # indices of axes that are being displayed

# consider a 5D dataset being rendered in 3D
len(viewer.dims.camera.center) == 3
len(viewer.dims.sliders.position) == 2
len(viewer.dims.position) == 5
len(viewer.dims.displayed) == 3

# consider a 5D dataset being rendered in 2D
len(viewer.dims.camera.center) == 2
len(viewer.dims.sliders.position) == 3
len(viewer.dims.position) == 5
len(viewer.dims.displayed) == 2
```

This could actually be done in a relatively backwards compatible way and might really make things crystal clear. Thoughts @jni and others? If people like this I can do that for this PR :-)

d-v-b:
To me it seems weird to see a reference to a specific UI element (sliders) in an object that seems abstract (representing how data is sampled). 

I also don't understand why you need `dims.position` and `dims.sliders.position`, when there is literally one underlying object that you could describe instead -- the sampling grid / hyperplane. For ND data, you could express this as a set of N basis vectors or as a set of ND points. In either case, sliders and panning with the mouse correspond to translations of the sampling grid along along axes orthogonal to it. In addition to matching the underlying math of image generation, I think this approach would be very forward-looking -- how will napari support array reslicing if you rely on this concept of "displayed dims"?

Sorry if this isn't very specific to the code in this PR. I am just worried that no API for image reslicing will be nice unless it matches the underlying geometrical operations.

sofroniewn:
>  I think this approach would be very forward-looking -- how will napari support array reslicing if you rely on this concept of "displayed dims"?
>
> Sorry if this isn't very specific to the code in this PR. I am just worried that no API for image reslicing will be nice unless it matches the underlying geometrical operations.

Hmm ok, this point is well taken, thanks for brining it up @d-v-b. I've done some experimenting with reslicing in the context of #1616 (though non-orthogonal slicing is not supported yet in that PR). It would be great if you can start playing with #1616 and give some feedback there. My plan was for that to merge without non-orthogonal slicing (right now it prints a warning if you try that) and then add that in as a new PR. Thinking about this API in the context of re-slicing makes a lot of sense though. Maybe we try and get #1616 merged first and then revisit this. Thoughts @d-v-b and @jni? This is definitely our opportunity to get this right!!!

sofroniewn:
@kevinyamauchi @royerloic curious what both of you think about this discussion too https://github.com/napari/napari/pull/854#issuecomment-703324023 from an API design standpoint

sofroniewn:
Ok, I've made a number of improvements to this PR and it is now ready for review. I settled on making minimal changes to the `Dims` in the process and didn't go with the approach which had a new sliders object etc. Instead we just have a very simple camera object accessible at `viewer.camera` with

```python
viewer.camera.center #2-tuple or 3-tuple of center of FOV in world slice coordinates
viewe.camera.zoom # number of canvas pixels per world pixel
viewer.camera.angles # euler angles of camera in degrees during 3D viewing
```

these get automatically synced with the vispy camera when you zoom/ pan etc. and changing them will automatically update the vispy camera too.

I think this is actually going to be really nice when we start having multicanvas support and want to link the cameras etc. I can actually really imagine now how we might be able to do this and could start work on this as soon as we release `0.4.0` - cc @GenevieveBuckley 

I'd like to merge #1616 first and then do some final testing before this merges, but otherwise I'd say this is ready for review, both functionality, API, and code, @jni @GenevieveBuckley @kevinyamauchi @Czaki 

I'm particularly interested @Czaki if this breaks any of your workflows, I know you have a multicanvas/ camera system in your tool.

napari/components/viewer_model.py
```diff
@@ -55,7 +56,6 @@ def __init__(
             title=Event,
             interactive=Event,
             cursor=Event,
-            reset_view=Event,
```
Czaki: I use this to catch reset view when synchronize two instances of viewer. It should be deprecated, not removed.

Czaki (commented):
I found one place where this PR break my tool. 

I also found that reset view is not working and size of image is not properly adjusted to canvas.

sofroniewn:
I've added `reset_view` back in, though the args are now slightly different and are based off the camera. Hopefully this is ok. You can also now listen for the `camera.events.center`, `camera.events.zoom`, `camera.events.angles` instead.

I think it's ok not to deprecate `reset_view`. We're no longer using it internally, but it doesn't do harm there. I've added a comment to that effect. Let me know if this is ok @Czaki. Curious what @jni thinks too

sofroniewn:
@sfcaracciolo, following up on your post on image.sc https://forum.image.sc/t/change-origin-of-3d-canvas/43583, this PR has now been cleaned up and is ready for testing. You can move the origin of the camera in 3D now with `viewer.camera.center = (z, y, x)`. Any feedback before merge would be much appreciated!!

sofroniewn:
I have now updated this PR to include the the affine transforms from #1616 which just merged.

Note that right now this PR doesn't fix #1739, but it doesn't make it worse either. I will look into that tomorrow. This is now ready for review @jni

sofroniewn:
Note the zoom change from 2D to 3D noted in #1740 is fixed here, see gif below:

world_coordinates_volume_fix3

Czaki:
Stil problem with reset view. Both screenshots are done after press the reset button. Only thing done is resize window.  

Zrzut ekranu z 2020-10-19 17-09-19
Zrzut ekranu z 2020-10-19 17-09-23

sofroniewn:
> Still problem with reset view. Both screenshots are done after press the reset button. Only thing done is resize window.

@Czaki can you explain in a little more detail what's going wrong
- is it now that the default size has changed when you reset the view, i.e the images a smaller?
- is it an unexpected interaction between canvas size and zoom level?
- does reset view not actually do anything for you anymore?

You might need to point me to the code in your project to help me understand this one more. Thanks and sorry about the breakage!!

sofroniewn:
Ok @Czaki I can reproduce what I think your issue is - and I think I know what causes it. If I make the canvas really big and take a screenshot it looks like this with lots of black space around the edge.

Screen Shot 2020-10-19 at 8 29 50 AM

If I make the canvas really small and take a screenshot it looks like this

Screen Shot 2020-10-19 at 8 30 12 AM

The current cause of this is the hard coded value of `600` for the canvas size that is being used here https://github.com/napari/napari/blob/32f5690418f816fee266483ceef57c8a985c7b7e/napari/components/viewer_model.py#L300. I can change this so that the viewer knows about the size of the canvas and that this gets adjusted automatically.

How does that sound? Thanks for brining this to our attention

Czaki:
> . I can change this so that the viewer knows about the size of the canvas and that this gets adjusted automatically.
> 
> How does that sound? Thanks for brining this to our attention

It looks ok. I like previous behavior and using the size of canvas should be a proper decision. Thick frames around images are also good. I may need to use some scalar to get them also.

sofroniewn:
> It looks ok. I like previous behavior and using the size of canvas should be a proper decision. Thick frames around images are also good. I may need to use some scalar to get them also.

Ok this is fixed now, I've added a `_canvas_size` property to the viewer that can be used during the reset, and I've also adjusted things to take into account different aspect ratios of canvas and image such that now the most filled axis will be filled 95% of the canvas regardless of canvas size etc.

For a long thin image like a pathology slide, this is what the default now looks like 

Screen Shot 2020-10-19 at 9 20 48 AM

This is what the default astronaut looks like

Screen Shot 2020-10-19 at 9 22 54 AM

Let me know @Czaki if this sufficiently addresses your concerns and needs!!

sofroniewn:
oh and the magic number `600` which would have set of the spidey sense of @jni has been removed :-)

sofroniewn:
@guiwitz this PR is finally ready for review!! I'm so sorry that it has taken us so long to get back to this and your incredible napari movie PR #780 has been sitting there the whole time. We really are trying to improve our development process so that PRs don't get stranded like that anymore, and once this merges I'd love us to pick that up one up again!!

Czaki:
There is still a problem with the size of the image after resize: The left window is napari 0.3.8, right from this PR 
napari-scalig

Czaki:
I also get this error when set axis with size 1 as one of the visible axes (using `self.viewer.dims.order`) 

```
/home/czaki/.pyenv/versions/partseg3.7/lib/python3.7/site-packages/napari/components/viewer_model.py:301: RuntimeWarning: divide by zero encountered in true_divide
  np.divide(self._canvas_size[::-1], size[-2:])
```
This works in napari 0.3.8

I use this to implement *side view*. And it happens when use *side view* on 2D data.

sofroniewn:
> There is still a problem with the size of the image after resize: The left window is napari 0.3.8, right from this PR

Unfortunately I can't reproduce on my screen right now :-( Here's a 512x512 image in the gif below. Just in case it helps what's the shape of your image - I see that it's 4D, maybe something weird is happening with that too (I hope not!!)

camera_resize

> I also get this error when set axis with size 1 as one of the visible axes (using self.viewer.dims.order)

Ok yes, I can try and reproduce/ fix this - should be straight forward.

Czaki:
I think that the most important differences are the shape of the image and that I maximize the window by double click on the title bar. You use the square image and I use rectangle one.  

I attached the image which I use. 

[stack1_component4.zip](https://github.com/napari/napari/files/5405217/stack1_component4.zip)

sofroniewn:
> I think that the most important differences are the shape of the image and that I maximize the window by double click on the title bar. You use the square image and I use rectangle one.

Unfortunately I can't reproduce this @Czaki - see my attempt below, the resize works ok for me.

camera_resize2

One potentially funny thing I'll note is the asymmetry that exists when resizing the canvas with the zoom depending on aspect ratio of the canvas - see this gif, and note when resizing left/ right the zoom doesn't change but when resizing up/ down it does. Somehow this behavior seems odd.

camera_resize3

This behavior is present on master right now, so I'd rather fix in a follow-up PR, but I just wanted to note.

Curious @Czaki if you have further ideas.

I'll fix the set axis with size 1 issue before merge though.

sofroniewn:
> I'll fix the set axis with size 1 issue before merge though.

1D images have now been fixed

Czaki:
It looks like in you make the window narrow you do not push the reset button and I do this. Without pushing the reset button it looks the same as in your gif. 
I would like to debug this but could you suggest which functions called during resizing are good to investigate? It looks like a problem with the update of scale.

napari-scalig2

napari/_vispy/vispy_camera.py
```diff
@@ -0,0 +1,148 @@
...
+from vispy.scene import ArcballCamera, PanZoomCamera
+
+from .quaternion import quaternion2euler
+
+
+class VispyCamera:
+    """Vipsy camera for both 2D and 3D rendering.
+    """
+
+    def __init__(self, view, camera, dims):
```
jni: @sofroniewn could you add a docstring here? What is `view`? What is `camera`?

napari/components/camera.py
```diff
@@ -0,0 +1,102 @@
+from ..utils.events import EmitterGroup
+
+
+class Camera:
```
jni, line 4: @sofroniewn you don't want to dataclass this up? ;)

napari/components/camera.py
```diff
@@ -0,0 +1,102 @@
...
+
+    @property
+    def zoom(self):
+        """float: Scale from canvas pixels to world pixels."""
+        return self._zoom
+
+    @zoom.setter
+    def zoom(self, zoom):
+        # if self._zoom == zoom:
+        #     return
```
jni: ?

napari/components/camera.py
```diff
@@ -0,0 +1,102 @@
...
+        """3-tuple: Euler angles of camera in 3D viewing, in degrees."""
+        if self.ndisplay == 3:
+            return self._angles
+        else:
+            return (0, 0, 90)
+
+    @angles.setter
+    def angles(self, angles):
+        # if self._angles == tuple(angles):
+        #     return
```
jni: I guess this is the sort of stuff that you need when connecting events, but not in model files?

jni (approved):
@sofroniewn I'm approving so you aren't blocked, but would be good to get a docstring on the VispyCamera class. I've pointed out other minor fixes.

It would be good to address @Czaki's issue, which points to something a bit weird going on with the zoom levels, but that could be done in a follow-up PR I think. Certainly in reading the code I don't understand at all why it happens...

Czaki:
> It would be good to address @Czaki's issue, which points to something a bit weird going on with the zoom levels, but that could be done in a follow-up PR I think. Certainly in reading the code I don't understand at all why it happens...

I think it could be some interaction with old code.

sofroniewn:
> It looks like in you make the window narrow you do not push the reset button and I do this. Without pushing the reset button it looks the same as in your gif.
> I would like to debug this but could you suggest which functions called during resizing are good to investigate? It looks like a problem with the update of scale.

Ok i was able to reproduce now!! I'll try and get to the bottom of this before merge

napari/_vispy/vispy_camera.py
sofroniewn: done

napari/components/camera.py
sofroniewn: follow up PR!! but will be great

napari/components/camera.py
sofroniewn: commented when debugging, adding back

napari/components/camera.py
sofroniewn: commented when debugging, adding back

napari/_vispy/vispy_camera.py
```diff
@@ -6,10 +6,18 @@
 
 class VispyCamera:
     """Vipsy camera for both 2D and 3D rendering.
+
+    Paramters
```
jni: typo - Parameters

sofroniewn:
Ok, i've made the simple doc string requests from @jni, but I'm really struggling with the zoom resetting issue.

I'm using this as example
```python
with napari.gui_qt():
    viewer = napari.view_image(np.random.random((200, 50)))
```
What can happen is that when I shrink the FOV very small, the image can shrink in a funny way, when i then click reset view it returns to normal, but then when i expand view it starts clipping again. If i click reset view it then returns to normal. I guess the desired behavior is that it never shrinks in a funny way, and expanding from being the "right zoom" keeps it the right zoom.

See gif here

camera_resize4

sofroniewn:
Note the current behavior on master is as follows. When you hit that minimum size, instead of continuing to shrink the zoom then remains constant. This behavior is much nicer and I can try and make this PR match that behavior before merge.

camera_resize5

napari/_vispy/vispy_camera.py
sofroniewn: fixed

sofroniewn:
Ok, I'm heading to bed now. I can't figure out what has changed in this PR that has caused the change in behaviour. Maybe @Czaki will have better luck investigating. I tried rolling back various aspects of this PR, but havn't been able to succeed.

sofroniewn:
Ok after much digging the problem can be traced to something with the call to `camera.rect` that used to happen here https://github.com/napari/napari/blob/v0.3.8/napari/_qt/qt_viewer.py#L460 now going wrong.

It looks like the bad behavior also happens in 3D rendering on master. I think I've narrowed this down quite a lot though, so we should be able to fix it!!

sofroniewn:
Ok the offending lines are

```python
corner = np.subtract(self._view.camera.center[:2], scale / 2)
self._view.camera.rect = tuple(corner) + (scale, scale)
```

where it makes it clear that information about data aspect ratio has been lost. I can look into it more

sofroniewn:
Ok, I've fixed the clipping issue that @Czaki noted and now understand what was going wrong with images of non 1 aspect ratio. We were using a square bounding box to determine the max zoom level for the panzoomcamera on vispy but using an aspect ratio matched box to determine the max zoom level in our method on the viewer model. these could get out of sync when clicking the reset view button which would then cause the clipping.

i've changed things so we always use a square bounding box to determine the max zoom. This also matches the behavior for the 3D camera too.

You can see the behavior in the new gif below. It does mean that at a certain point you get what looks like a large shrinking, but that is to keep the square bounding box entirely within view. The nice thing now is that pressing the view reset button doesn't change anything like it did before and no clipping happens. Is this ok @Czaki?

camera_resize6

sofroniewn:
Ok, I'm going to merge!!! If we want to improve the zoom in the future we can revisit, but I'm pretty happy with this now, and it will be great to have it in!

jni:
@sofroniewn amazing sleuthing! :detective: :raised_hands:

Czaki:
@sofroniewn Thanks. I'm a little hurry so do not have time to check. When I will have more time, I will investigate this zoom issue.

PR #859, opened 2020-01-08
Qt/Vispy connection & lambda refactor

tlambert03:
# Description
Following up on the discussion in https://github.com/napari/napari/pull/846#discussion_r364016783, this PR is a broad sweep through the codebase cleaning up and standardizing our usage of connections (both in Qt and vispy), and lambda functions.    Here were the rules I followed:

- if there is a class method that is more or less designed to be a vispy event handler or a qt slot, even if the method doesn't use the event object, we should add `event=None` to the method signature so that we don't have to catch the event and discard it with a lambda every time it is connected:

  ```python
  self.events.something.connect(lambda x: self._on_something())
  
  #becomes
  self.events.something.connect(self._on_something)
  ...
  def _on_something(self, event=None):
      handle_something()
  ```

- `connect(lambda val: self.doSomething(val))` is the same as `connect(self.doSomething)` and has been changed where appropriate

- defaults in lambda parameters are used in a lot of places throughout the code, but often the defaults don't have the same type as the eventual signal.  So it wasn't clear to me why they were being provided.  Removing them seems to have no effect either on tests or functionality, but let me know if I'm missing something:
  ```python
  sld.valueChanged.connect(lambda value=sld: self.changeSize(value))

  # sld is never an integer, right? ... it becomes
  sld.valueChanged.connect(self.changeSize)
  ```

- the syntax `widget.signal[type].connect` is used for [overloaded signals](https://wiki.qt.io/Qt_for_Python_Signals_and_Slots), and it is unnecessary unless the signal you are handling is overloaded _and_ you want the non default signature.  For instance: 
  ```python
  # this is unnecessary because Qslider.valueChanged always returns an int
  QSlider.valueChanged[int].connect()
  # this is necessary because QCombBox.activated is overloaded, 
  # and the default is an int
  QComboBox.activated[str].connect()

I made a couple additional Qt refactors:

- Juan's `QtModeButton` was being used in some places but other places had lots of duplicated classes.  I used his button class everyone and removed the duplicates.  But I renamed it to `QtModeRadioButton` because I also added a `QtModePushButton` that serves a similar code-reduction purpose.
- wherever there was a `for` loop using `addItem` to a combobox, I turned it into `addItems`
- I added a `keys()` method to our `StringEnum` because we often just want to get the list of options for a combo box.

i think a lot of the code is much easier to follow now.  Hope you agree and hope it isn't a pain to merge into any open PRs.

## Type of change
<!-- Please delete options that are not relevant. -->
- [x] Refactor

# References
https://wiki.qt.io/Qt_for_Python_Signals_and_Slots

# How has this been tested?
<!-- Please describe the tests that you ran to verify your changes. -->

### DIFF
--- a/napari/_qt/layers/qt_labels_layer.py
+++ b/napari/_qt/layers/qt_labels_layer.py
@@ -127,7 +127,7 @@ class QtLabelsControls(QtLayerControls):
         elif mode == Mode.FILL:
             self.fill_button.setChecked(True)
         else:
-            raise ValueError("Mode not recongnized")
+            raise ValueError("Mode not recognized")
 
     def changeColor(self):
         self.layer.new_colormap()
--- a/napari/_qt/layers/qt_labels_layer.py
+++ b/napari/_qt/layers/qt_labels_layer.py
@@ -1,7 +1,6 @@
 from qtpy.QtGui import QPainter, QColor
 from qtpy.QtWidgets import (
     QButtonGroup,
-    QRadioButton,
     QWidget,
     QPushButton,
     QSlider,
@@ -15,6 +14,7 @@ from qtpy.QtCore import Qt
 import numpy as np
 from .qt_base_layer import QtLayerControls
 from ...layers.labels._constants import Mode
+from ..qt_mode_radio import QtModeRadio
 
 
 class QtLabelsControls(QtLayerControls):
@@ -64,23 +64,22 @@ class QtLabelsControls(QtLayerControls):
         self.ndimCheckBox = ndim_cb
         self._on_n_dim_change(None)
 
-        self.panzoom_button = QtModeButton(
-            layer, 'zoom', Mode.PAN_ZOOM, 'Pan/zoom mode'
+        self.panzoom_button = QtModeRadio(
+            layer, 'zoom', Mode.PAN_ZOOM, 'Pan/zoom mode', checked=True
         )
-        self.pick_button = QtModeButton(
+        self.pick_button = QtModeRadio(
             layer, 'picker', Mode.PICKER, 'Pick mode'
         )
-        self.paint_button = QtModeButton(
+        self.paint_button = QtModeRadio(
             layer, 'paint', Mode.PAINT, 'Paint mode'
         )
-        self.fill_button = QtModeButton(layer, 'fill', Mode.FILL, 'Fill mode')
+        self.fill_button = QtModeRadio(layer, 'fill', Mode.FILL, 'Fill mode')
 
         self.button_group = QButtonGroup(self)
         self.button_group.addButton(self.panzoom_button)
         self.button_group.addButton(self.paint_button)
         self.button_group.addButton(self.pick_button)
         self.button_group.addButton(self.fill_button)
-        self.panzoom_button.setChecked(True)
         self._on_editable_change(None)
 
         button_row = QHBoxLayout()
@@ -174,24 +173,6 @@ class QtLabelsControls(QtLayerControls):
         self.pick_button.setEnabled(self.layer.editable)
         self.paint_button.setEnabled(self.layer.editable)
         self.fill_button.setEnabled(self.layer.editable)
-
-
-class QtModeButton(QRadioButton):
-    def __init__(self, layer, button_name, mode, tool_tip):
-        super().__init__()
-
-        self.mode = mode
-        self.layer = layer
-        self.setToolTip(tool_tip)
-        self.setChecked(False)
-        self.setProperty('mode', button_name)
-        self.toggled.connect(lambda state=self: self._set_mode(state))
-        self.setFixedWidth(28)
-
-    def _set_mode(self, bool):
-        with self.layer.events.mode.blocker(self._set_mode):
-            if bool:
-                self.layer.mode = self.mode
 
 
 class QtColorBox(QWidget):
--- a/napari/_qt/layers/qt_labels_layer.py
+++ b/napari/_qt/layers/qt_labels_layer.py
@@ -65,16 +65,20 @@ class QtLabelsControls(QtLayerControls):
         self._on_n_dim_change()
 
         self.panzoom_button = QtModeRadioButton(
-            layer, 'zoom', Mode.PAN_ZOOM, 'Pan/zoom mode', checked=True
+            layer,
+            'zoom',
+            Mode.PAN_ZOOM,
+            tool_tip='Pan/zoom mode',
+            checked=True,
         )
         self.pick_button = QtModeRadioButton(
-            layer, 'picker', Mode.PICKER, 'Pick mode'
+            layer, 'picker', Mode.PICKER, tool_tip='Pick mode'
         )
         self.paint_button = QtModeRadioButton(
-            layer, 'paint', Mode.PAINT, 'Paint mode'
+            layer, 'paint', Mode.PAINT, tool_tip='Paint mode'
         )
         self.fill_button = QtModeRadioButton(
-            layer, 'fill', Mode.FILL, 'Fill mode'
+            layer, 'fill', Mode.FILL, tool_tip='Fill mode'
         )
 
         self.button_group = QButtonGroup(self)
--- a/napari/_qt/layers/qt_labels_layer.py
+++ b/napari/_qt/layers/qt_labels_layer.py
@@ -65,20 +65,16 @@ class QtLabelsControls(QtLayerControls):
         self._on_n_dim_change()
 
         self.panzoom_button = QtModeRadioButton(
-            layer,
-            'zoom',
-            Mode.PAN_ZOOM,
-            tool_tip='Pan/zoom mode',
-            checked=True,
+            layer, 'zoom', Mode.PAN_ZOOM, tooltip='Pan/zoom mode', checked=True
         )
         self.pick_button = QtModeRadioButton(
-            layer, 'picker', Mode.PICKER, tool_tip='Pick mode'
+            layer, 'picker', Mode.PICKER, tooltip='Pick mode'
         )
         self.paint_button = QtModeRadioButton(
-            layer, 'paint', Mode.PAINT, tool_tip='Paint mode'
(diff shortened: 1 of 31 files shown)


napari/_qt/layers/qt_labels_layer.py
```diff
@@ -48,43 +48,40 @@ def __init__(self, layer):
         sld.setMinimum(1)
         sld.setMaximum(40)
         sld.setSingleStep(1)
-        sld.valueChanged[int].connect(lambda value=sld: self.changeSize(value))
+        sld.valueChanged.connect(self.changeSize)
         self.brushSizeSlider = sld
         self._on_brush_size_change(None)
```
sofroniewn: can `self._on_brush_size_change(None)` -> `self._on_brush_size_change()` by adding `event=None` to the function def?

napari/_qt/layers/qt_labels_layer.py
```diff
@@ -48,43 +48,40 @@ def __init__(self, layer):
...
 
         contig_cb = QCheckBox()
         contig_cb.setToolTip('contiguous editing')
-        contig_cb.stateChanged.connect(
-            lambda state=contig_cb: self.change_contig(state)
-        )
+        contig_cb.stateChanged.connect(self.change_contig)
         self.contigCheckBox = contig_cb
         self._on_contig_change(None)
```
sofroniewn: can `self._on_contig_change(None)` -> `self._on_contig_change()` by adding `event=None` to the function def?

napari/_qt/layers/qt_labels_layer.py
```diff
@@ -48,43 +48,40 @@ def __init__(self, layer):
...
 
         ndim_cb = QCheckBox()
         ndim_cb.setToolTip('n-dimensional editing')
-        ndim_cb.stateChanged.connect(
-            lambda state=ndim_cb: self.change_ndim(state)
-        )
+        ndim_cb.stateChanged.connect(self.change_ndim)
         self.ndimCheckBox = ndim_cb
         self._on_n_dim_change(None)
```
sofroniewn: can `self._on_n_dim_change(None)` -> `self._on_n_dim_change()` by adding `event=None` to the function def?

napari/_qt/layers/qt_labels_layer.py
```diff
@@ -48,43 +48,40 @@ def __init__(self, layer):
...
+            layer, 'fill', Mode.FILL, 'Fill mode'
+        )
 
         self.button_group = QButtonGroup(self)
         self.button_group.addButton(self.panzoom_button)
         self.button_group.addButton(self.paint_button)
         self.button_group.addButton(self.pick_button)
         self.button_group.addButton(self.fill_button)
-        self.panzoom_button.setChecked(True)
         self._on_editable_change(None)
```
sofroniewn: can `self._on_editable_change(None)` -> `self._on_editable_change()` by adding `event=None` to the function def?

napari/_qt/qt_viewer_buttons.py
```diff
@@ -28,12 +43,27 @@ def __init__(self, viewer):
         super().__init__()
 
         self.viewer = viewer
-        self.consoleButton = QtConsoleButton(self.viewer)
-        self.ndisplayButton = QtNDisplayButton(self.viewer)
-        self.rollDimsButton = QtRollDimsButton(self.viewer)
-        self.transposeDimsButton = QtTransposeDimsButton(self.viewer)
+        self.consoleButton = QtViewerPushButton(
```
sofroniewn, line 46: love this

sofroniewn (approved):
Love this PR, so nice to see refactors pull the # of lines down :-)

I flagged some other places where we still have `(None)` calls that seem like they can be removed too - maybe worth doing a search across the whole repo to make sure you've caught them all

great work - code is much more readable now

napari/_qt/qt_mode_buttons.py
```diff
@@ -0,0 +1,32 @@
+from qtpy.QtWidgets import QRadioButton, QPushButton
+
+
+class QtModeRadioButton(QRadioButton):
+    def __init__(self, layer, button_name, mode, tool_tip, checked=False):
```
sofroniewn: should some of these args be keyword based? I worry they might be hard to keep track of / remember based on order along? thoughts @jni?

napari/_qt/qt_mode_buttons.py
```diff
@@ -0,0 +1,32 @@
...
+        self.setFixedWidth(28)
+
+    def _set_mode(self, bool):
+        with self.layer.events.mode.blocker(self._set_mode):
+            if bool:
+                self.layer.mode = self.mode
+
+
+class QtModePushButton(QPushButton):
+    def __init__(self, layer, button_name, tool_tip, slot=None):
```
sofroniewn: similar question about should more of these be keyword args @jni?

napari/_qt/qt_viewer_buttons.py
```diff
@@ -76,67 +106,15 @@ def dropEvent(self, event):
...
-
-        self.viewer = viewer
-        self.setToolTip('Transpose displayed dimensions')
-        self.clicked.connect(lambda: self.viewer.dims._transpose())
-
-
-class QtConsoleButton(QPushButton):
-    def __init__(self, viewer):
+class QtViewerPushButton(QPushButton):
+    def __init__(self, viewer, tooltip, button_name, slot=None):
```
sofroniewn: similar question @jni about more keyword arg? maybe not for `viewer` or `button_name`, but for `tooltip=None`, could then default on `button_name` if None

napari/_qt/qt_mode_buttons.py
sofroniewn: maybe just `tooltip=None` with default to `button_name` if not provided, and maybe after button name `*`

napari/_qt/qt_viewer_buttons.py
sofroniewn: maybe reorder `button_name` and `tooltip` and add `*` after `button_name`

napari/_qt/qt_mode_buttons.py
sofroniewn: maybe made `mode=None` and `tool_tip=None` and then have them default to `button_name`, see comments below - though maybe `mode` here is more important and we want to be more primary ...

napari/_qt/qt_mode_buttons.py
```diff
@@ -0,0 +1,35 @@
+from qtpy.QtWidgets import QRadioButton, QPushButton
+
+
+class QtModeRadioButton(QRadioButton):
+    def __init__(
+        self, layer, button_name, mode=None, tool_tip=None, checked=False
```
sofroniewn: Looking at this again I guess `mode=None` doesn't really make sense here as this button would never do anything without mode passed. Maybe ok to make that not keyword arg, but I still might be the `*` before `tool_tip` for all these so `tool_tip` had to be keyword arg, and then go back and add `tool_tip=` to all the places where the function is called, it might make them more readable. If you prefer otherwise @tlambert03 that's totally fine, but curious why?

tlambert03: yeah, the mode arg is pretty important here.  I can make it required again if you want.  I don't think it matters too much since it's only going to be used by GUI builders.  My inclination is to make it required.  
I don't care too much either way regarding the tool_tip.  I haven't personally used the `*` enforcement much in my own stuff, but I don't feel strongly about it.  I guess we're just saying we never want to see this used without someone labeling it as a tooltip for readability?

sofroniewn: making mode required sounds good.
yeah on `tool_tip` that matches how I think about it, @jni has good intuition for this stuff, maybe he  can weigh in, but I think he's tended to be in favor of at least using `tool_tip=` like stuff in our internal code instead of position stuff even if its not required for readability, so we could just do that without the `*`

napari/_qt/qt_mode_buttons.py
kne42: agreed, having these as keyword-only arguments would make it much clearer as to what each parameter does in the code

tlambert03: Everything that can be keyworded has been in the most recent commit, and things that are optional remain positional.  Non optional args remain positional

napari/_qt/qt_mode_buttons.py
jni: > I guess we're just saying we never want to see this used without someone labeling it as a tooltip for readability?

Yes, exactly. For example, all of the QWidget(parent, 0, 1, 1, 2, 75, whatever) calls drive me up the wall. Where arguments are not essentially self-documenting (e.g. out = gaussian_filter(arr)), keyword-only arguments should be used.

Also imho, `tooltip` over `tool_tip`. The joined form is a word now as far as I'm concerned.

tlambert03: sounds good, I can change that.  also, just cause I don't want you to think i'm hastily adding new classes without thinking about them 😂, that class was just transplanted from the qt_shapes layer in master (but I agree this is the time to reconsider it and I could/should have done that):
https://github.com/napari/napari/blob/0f2f8ef98b5f22983283cbd6072f92b267304fbe/napari/_qt/layers/qt_shapes_layer.py#L239-L241

sofroniewn:
Windows test is now failing - looks like one of the two many files open test that @shanaxel42 did a great job fixing - maybe she can look again. Is this failing on master too or is it something introduced by this PR? Might help her narrow it down

jni (approved):
@tlambert03 I love this PR! :heart_eyes:

tlambert03:
thanks.  now I just gotta find some time to figure out what's going on with tests... which were working at some point before.

jni:
:joy: commit messages :joy:

jni:
@tlambert03 this looks ready to merge!! :tada: :rocket: Is it ready???

tlambert03:
Yep, looks like it!  LGTM!

PR #866, opened 2020-01-10
remove calls to QPoint.toTuple which is invalid in PyQt5

kne42:
# Description
Remove calls to `QPoint.toTuple` which is part of `PySide2`'s API but not `PyQt5`'s.

## Type of change

- [x] Bug-fix (non-breaking change which fixes an issue)

# References
fixes #865

# How has this been tested?

- [x] the test suite passes for both `PyQt5` and `PySide2` installs

PR #867, opened 2020-01-11
Add PyQt5 tests to Linux and OSX CI

### LINKED ISSUE #768: Test both PySide2 and PyQt5
While we have chosen to ship with PySide2 rather than PyQt5 due to its more permissive license, we should still ensure that `napari` is compatible with PyQt5 (and any other Python bindings supported by `qtpy` where we can).

Does anyone have an opinion on whether we should run tests with both packages on only the CI or locally as well?

see #767 for an example of how not testing both packages has allowed bugs to slip through that break compatibility with one package

tlambert03:
# Description
sorry about missing that API difference in #866 !  Here's an update to our tests that will also test PyQt5 in Linux and OSX.  It fails on the current master until we merge @kne42's fix.
until it finishes testing here, here's an example failed test on cirrus:
https://cirrus-ci.com/task/4594543160983552

thoughts @kne42 ?

## Type of change

- [x] update to our CI

# References

closes #768

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -132,7 +132,12 @@ windows_task:
     - pip list
     - python -c "import napari; print(napari.__version__)"
 
-  test_script:
+  test_pyside2_script:
+    - pytest -v -o qt_no_exception_capture=1
+
+  test_pyqt5_script:
+    - pip uninstall -y pyside2
+    - pip install pyqt5
     - pytest -v -o qt_no_exception_capture=1
 
 codecov_task:
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -83,7 +83,7 @@ task:
     - pytest -v
 
   switch_backend_script:
-    - pip uninstall pyside2
+    - pip uninstall -y pyside2
     - pip install pyqt5
 
   testpyqt_script:
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,91 +1,91 @@
-task:
-  auto_cancellation: true
+# task:
+#   auto_cancellation: true
 
-  name: Black Formatting
-  container:
-      image: python:3.6
+#   name: Black Formatting
+#   container:
+#       image: python:3.6
 
-  install_script:
-    - python -m pip install --retries 3 --upgrade pip
-    - pip install `cat requirements/development.txt | grep black`
+#   install_script:
+#     - python -m pip install --retries 3 --upgrade pip
+#     - pip install `cat requirements/development.txt | grep black`
 
-  test_script:
-    - black --check .
+#   test_script:
+#     - black --check .
 
-nix_test_task:
-  auto_cancellation: true
+# nix_test_task:
+#   auto_cancellation: true
 
-  name: Flake8 Linting
-  container:
-      image: python:3.6
+#   name: Flake8 Linting
+#   container:
+#       image: python:3.6
 
-  install_script:
-    - python -m pip install --retries 3 --upgrade pip
-    - pip install `cat requirements/development.txt | grep flake8`
+#   install_script:
+#     - python -m pip install --retries 3 --upgrade pip
+#     - pip install `cat requirements/development.txt | grep flake8`
 
-  test_script:
-    - flake8 napari
+#   test_script:
+#     - flake8 napari
 
 
-task:
-  auto_cancellation: true
+# task:
+#   auto_cancellation: true
 
-  env:
-    matrix:
-      PY_VER: "3.6"
-      PY_VER: "3.7"
+#   env:
+#     matrix:
+#       PY_VER: "3.6"
+#       PY_VER: "3.7"
 
-  matrix:
-    - name: linux
-      container:
-        image: gcc:latest
-      env:
-        PATH: $HOME/conda/bin:$PATH
-        DISPLAY: ":99"
-      system_script:
-        - apt-get update
-        - apt-get install -y libgl1-mesa-glx xvfb libqt5x11extras5 herbstluftwm
-      conda_script:
-        - curl https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh > install.sh
-        - bash install.sh -b -p $HOME/conda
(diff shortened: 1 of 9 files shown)


kne42 (approved):
I love it!

sofroniewn:
Looks good to me too. Would be good to get #866 merged into master and then into this PR and re-run the tests to make sure they all pass before we merge this one.

tlambert03:
yeah I agree.  I'll merge #866 and pull into this

tlambert03:
this seems to be working now... but the osx tests are hung waiting to run due to concurrency limits.

sofroniewn:
@tlambert03 just so I understand this for each of our 6 configurations, (python 3.6, 3.7 across 3xOS) we will now in serial run all our tests first with pyside2 and then with pyqt5 following an uninstall of pyside2? That means all of our model tests which are Qt independent will run twice, which I think is fine, they are fast, and then all the Qt stuff will be tested once with PySide2 and once with PyQt5?

jni:
I guess this now needs master to check that it's passing?

tlambert03:
that's mostly correct except it's currently only testing PyQt5 on linux and osx (which are usually done before windows anyway), and _not_ testing pyqt5 on windows.  I figured the serial approach is probably better than parallel, since the environment setup is a significant part of the task time, and concurrency limits at cirrus prevent us from having that many parallel tasks.

>I guess this now needs master to check that it's passing?

yeah, we need to merge the zarr changes here to test them

tlambert03:
also, it would be trivial to only run `pytest napari/_qt` for PyQt5 testing if you prefer to limit that second test to Qt specific stuff

jni:
> only run pytest napari/_qt for PyQt5 testing if you prefer to limit that second test to Qt specific stuff

Clever! I think that's a very good idea!

I'm ok with only testing PyQt5 on some platforms. Ultimately, chasing a complete test matrix is impossible, and we should just get "as good as reasonably possible".

tlambert03:
ok, merged the master zarr changes, and change pyqt5 to only test napari/_qt.  will see how it goes here

tlambert03:
looking good!  merge?

sofroniewn:
wait, `napari/test` also contains lots of Qt tests too

tlambert03:
yeah, good point.  since it really doesn't add much time at all (maybe a couple minutes, and we're generally waiting on windows anyway during those minutes), let's just keep it testing the whole thing until we decide we need it to be faster.

sofroniewn:
> PyQt5 on linux and osx (which are usually done before windows anyway), and not testing pyqt5 on windows. 

Just so I understand, PySide2 is tested on all 6, and PyQt5 is now tested on four, everything apart from windows. Is there a reason we're not testing PyQt5 on windows? I agree with @jni might not be needed for complete the full matrix, but just curious why this choice.

Also rerunning PyQt5 tests only on tests with Qt stuff makes sense, but we actually have three places right now where that happens `napari/_qt`, `napari/_vispy`, and `napari/tests` so all three of those should be included before merge

sofroniewn:
Ok, testing whole repo is fine, the model tests are very fast

tlambert03:
correct.  the rationale was that we're almost always waiting on the windows tests, so it wouldn't really add any time to put it on linux and osx only... we would still catch all api differences (and I don't know what sorts of things are likely to be platform _and_ qt backend specific, not that can't exist).  i don't have a strong opinion though...

sofroniewn:
> waiting on the windows tests

agreed but I think most of that is waiting for the machines to be available / things to kick off, but then once it gets running I don't think our tests are that long such that it adds much more time. I'd probably prefer to just have the duplication and know everything is tested everywhere rather then be unpleasantly surprised about Qt binding and platform specific differences.

I also don't find myself waiting on tests to actually make a merge very often (i.e. I'll do something else after that final push with all the approvals in place) and think the extra time may help our culture of trying to give everyone a chance to weigh in before merge. Maybe it makes development a little slower if you don't see problem until much later. I don't have strong feelings though, so if it makes your life better to have it this way fine to leave too

tlambert03:
doesn't affect my life either way :) so if there's even a slight preference, let's go with the thorough testing and revisit later if we change our minds.  will update

sofroniewn:
Great, thanks! @tlambert03 you can merge once all the tests pass :-)

tlambert03:
well that's interesting!  both of the windows pyqt5 had failures in one of the animation tests, and they both seemed to have crashed in the screenshot test.  let me try to reproduce those locally to see what's going on

tlambert03:
well, I was excited when I thought maybe it had caught a new bug, but now I think this has more to do with virtualization and the CI environment.  I am unable to reproduce this on a real windows system with pyqt5 (or pyside2).  the screenshots always work, (all tests do).
but the screenshot tests are consistently failing on CI, with a stack overflow (`Exit status: 3221226505`)... so I'm not sure what to do about it.  and not sure why it's only pyqt5.  
signing off for the night though...  feel free to look into it, push to the PR, and/or tell me what you want to do tomorrow

sofroniewn:
Hmm ok, I think we can wait to tomorrow to look into this. Maybe @ttung, @shanaxel42, or @kne42 might have an idea

tlambert03:
I played around with this today and i'm pretty sure now these are just the same issues we've always had with windows testing, it just appears that pyqt5 and pyside2 fail in different ways:
- If I test on a remote windows computer using remote desktop, I get the same stack overflow crash that we see on CI, during the same tests
- If I test on windows with RDP using pyside2, it doesn't crash, but all of the tests fail (with that old `LooseVersion` error message) unless I add that `-o qt_no_exception_capture=1` that we have on there.  So either way, testing gl-dependent stuff on a virtualized windows machine isn't working.  It's just a hard crash on PyQt5 and a soft crash on Pyside2

given that this is really just a different manifestation of an old (and bigger) issue, I propose we roll back to just testing PyQt5 on linux and mac as this began.

sofroniewn:
ahhh yes, makes sense - let's roll back and then and get this in

tlambert03:
Ok, we’re back to the minimal change that @kne42 approved and I think @jni was ok with as well.  It will at least catch api differences between the two backends. Merging

Issue #868, opened 2020-01-11
new zarr release / numcodecs


sofroniewn:
## 🐛 Bug

Looks like zarr's new release requires numcodecs > 0.6.4, but we pinned to exclude it see discussion #666. I think we need to resolve this ASAP and then make the 0.2.10 release (which also includes the #866 bug fix). Thoughts @tlambert03 @jni? Has the 0.6.4 numcodecs install problem been resolved? You can see our failing tests in #867.

tlambert03:
that's funny... I was literally about to hit "submit" on a similar issue.  
[here is the commit](https://github.com/zarr-developers/zarr-python/pull/497) where they bump the dependency, and they do it " to use `ensure_text` text handling functionality from it".  I don't know zarr deeply enough to know how frequently that is likely to happen...

numcodecs 0.6.4 is still not working, (this actually was just [checked in the dependabot fork today](https://github.com/tlambert-forks/napari-dependabot/pull/4) 😊)...  so not sure what to do about that

tlambert03:
actually... looks like it was just the windows builds that [broke with numcodecs](https://github.com/tlambert-forks/napari-dependabot/pull/4/checks?check_run_id=384243596)... so maybe something has changed in the meantime?

sofroniewn:
ok - also if I remember https://github.com/zarr-developers/numcodecs/issues/210 correctly the issue is just with pip install on our CI, so people in the wild should still be installing ok. Maybe then this can wait till tomorrow

jni:
pip install works fine on my machine. imho the fix is to stop depending on zarr. We can use `dask.array.from_zarr` instead of `zarr.open`, and it should work whenever the file has standard compression, if I understand correctly.

jni:
Ah. I just saw why we started using zarr directly: we can tell whether there is a dataset instead of an array. Guh. =)

sofroniewn:
hmm - relevant part of the codebase is here https://github.com/napari/napari/blob/e2b90e96189cc18f4b5e5594b19f302ee305d380/napari/utils/io.py#L118

I agree that we should figure out a way to drop our dependency on both zarr and numcodecs, and just leverage dask, possibly with try catches to do everything, or say that additional functionality belongs in a plugin. We can still have zarr as one of our test dependencies (if that works with our CI) or just use dask functions for all our tests too. I'll think more about this tomorrow too

sofroniewn:
[...]

PR #871, opened 2020-01-11
fix aspect ratio press

### LINKED ISSUE #870: Error when locking shape aspect ratio with shift
## 🐛 Bug
haven't looked too far into yet, but just want to note it somewhere:
In `examples/mouse_drag_callback.py`, if I try to use the `Shift` key to constrain the aspect ratio of a shape, I get the following error in `hold_to_lock_aspect_ratio`:

```
  File "napdev/lib/python3.7/site-packages/vispy/app/backends/_qt.py", line 502, in keyPressEvent
    self._keyEvent(self._vispy_canvas.events.key_press, ev)
  File "napdev/lib/python3.7/site-packages/vispy/app/backends/_qt.py", line 551, in _keyEvent
    func(native=ev, key=key, text=text_type(ev.text()), modifiers=mod)
  File "napdev/lib/python3.7/site-packages/vispy/util/event.py", line 455, in __call__
  File "napdev/lib/python3.7/site-packages/vispy/util/event.py", line 475, in _invoke_callback
    self, cb_event=(cb, event))
  File "napdev/lib/python3.7/site-packages/vispy/util/event.py", line 471, in _invoke_callback
    cb(event)
  File "napari/napari/_qt/qt_viewer.py", line 394, in on_key_press
    next(gen)
  File "napari/napari/layers/shapes/keybindings.py", line 39, in hold_to_lock_aspect_ratio
    layer._move(layer.coordinates[layer.dims.displayed])
TypeError: tuple indices must be integers or slices, not list
```

sofroniewn:
# Description
This PR closes #870 which I was able to reproduce. It doesn't add a test, which would involve simulating both mouse movement and button pressing during the move. @tlambert03 do you have good ideas on how to test / feelings on if a test is needed here are this point. I'd like to get this out in `0.2.10`, but think we need to have a broader discussion about adding GUI tests with mouse move / presses across the repo.

## Type of change

# How has this been tested?
Not tested


tlambert03 (approved):

tlambert03:
I tried it out and this works for me!

Regarding testing, correct me if I'm wrong, but couldn't you test this without simulating mouse movements (using qtbot for instance) by just calling `hold_to_lock_aspect_ratio` directly on layers (with various preset states) until you hit all of the conditions required to get 100% coverage for that function?  There might be some conditions that occur with real mouse movements that I'm not anticipating, but something like this is sufficient to catch this particular `TypeError`:

```python
def test_lock_aspect_ratio():
    # Test a single four corner rectangle
    layer = Shapes(20 * np.random.random((1, 4, 2)))
    layer._is_moving = True
    # need to go through the generator
    _ = list(hold_to_lock_aspect_ratio(layer))
```

(you might already know/do this, but just in case)
For rapid local iteration, I like to use `pytest-cov`, write a test, and then call pytest on a small section of tests like this:
`$ pytest --cov-report=html --cov=napari/layers/shapes napari/layers/shapes`
and then open the html report:
`$ open htmlcov/index.html`
which gives you something like this:

Screen Shot 2020-01-11 at 5 19 01 PM

and then keep adding conditions until you've covered the lines in the function you're working on.

As a sidenote here, I've run into difficulties actually simulating click & drag events in pytest-qt and eventually stumbled onto [this thread](https://github.com/pytest-dev/pytest-qt/issues/260), that makes it sound like there might be a bug in the Qt QTest object itself.

sofroniewn:
Thanks for those great suggestions @tlambert03!! I actually wasn't familiar with looking at code-cov snippets like that locally - it is amazing! Definitely going to start trying to push the repo towards more test coverage. I ended up adding a file just for shape keybindings and brought it all the way to 100% coverage. Some of these tests would have failed without this PR. Let me know what you think!!

tlambert03:
hooray!  yeah, if you don't have to wait for CI-codecov, and you can look at those html reports, it can turn into a little game eh?  these tests are great and it's good you can do it without simulating mouse clicks.  As with #869, would be nice to figure out what we're going to do about zarr/CI before merging this.  Do you have thoughts on [this strategy](https://github.com/napari/napari/issues/868#issuecomment-573361473) for (perhaps even temporarily) removing the zarr dependency?

sofroniewn:
agreed it's kinda fun! also agreed on waiting to merge to fix our CI, I will look at #868 now

jni (approved):
This is great! @sofroniewn if you merge master in, I reckon this is ready to go!

sofroniewn:
I have just merged in master, I will merge if all tests pass

PR #872, opened 2020-01-12
fix None label

sofroniewn:
# Description
This PR fixes #857. I also havn't added tests, not sure if it might be easier to add comprehensive tests after @kne42 finishes the mouse function refactoring and tests in #694 

## Type of change

- [x] Bug-fix (non-breaking change which fixes an issue)

PR #873, opened 2020-01-12
Drop zarr and numcodecs dependency

### LINKED ISSUE #868: new zarr release / numcodecs

sofroniewn:
# Description
This PR closes #868 using the solution from @tlambert03 provided in https://github.com/napari/napari/issues/868#issuecomment-573361473 with minor modification. I added zarr to our test dependencies so we can still use it there. We'll see how our tests do and if we then need any followups. @tlambert03 feel free to take over the PR and make them directly if you like.

Once we have our CI fixed, we'll want to merge these changes into #867, #869, #871, #872 and then release I think.

## Type of change

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -114,12 +114,16 @@ windows_task:
   conda_script:
     - choco install -y miniconda3 --params="'/D:%ANACONDA_LOCATION%'"
     - conda install --yes -c conda-forge setuptools  # https://github.com/napari/napari/pull/594#issuecomment-542475164
+    - where python
+    - where conda
     - conda update -yn base conda
     - conda install -y python=%PY_VER%
     - pip install setuptools-scm
 
   install_script:
-    - pip install --retries 3 --upgrade pip
+    - where python
+    - echo %PATH%
+    - python -m pip install --retries 3 --upgrade pip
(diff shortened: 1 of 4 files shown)


tlambert03:
Ok, looks like CI is now working again with all our tests on all platforms, with zarr as a test dependency.  We might want to leave this overnight for @jni to look at the new zarr function, and I wouldn’t mind having one more think on it tomorrow morning.  But assuming everyone is 👍 we can merge this and make sure it works with all those others mentioned above

sofroniewn:
Thanks!! Agreed let’s revisit tomorrow, would be nice if @jni can give a thumbs up and then we can get the 0.2.10 PRs merged and start the release process tomorrow

tlambert03 (approved):

jni (approved):
Perfect!

PR #877, opened 2020-01-14
Fix invalid instructions in setup.py

### LINKED ISSUE #812: Setup.py displays invalid command
Not sure why, but setup.py says `pip install python==3.6` installs Python 3.6, which is not at all true?  pip is a dependency manager for Python, not a version manager like [RVM](https://rvm.io), I think you may be looking for Pyenv?

RDIL:
Bug fix, fixes #812


sofroniewn (commented):
Thanks for fixing this @RDIL! If we know that #878 works then this looks like a great fix, otherwise we might want to say `3.6` and `3.7` explicitly - let's resolve #878 first and then we can merge this one

jni:
I would say that python.org downloads are one of the less common ways to get Python these days, especially in scientific Python circles. I would point either to miniconda or to a less prescriptive list. Having said that, I haven't found in a quick search an up-to-date list of options (Anaconda, miniconda, Canopy, Python(x,y), python.org...), so maybe I need to write this...

tlambert03:
yeah I would just remove the specifics on how to install, just leave it at "please install python 3.6 or later".

sofroniewn (approved):
I have updated this PR to remove the specific install instructions as requested above and fixed some flake8 errors. I will merge after the tests pass. Note that this will serve some merge conflicts in #880 that will need to be resolved but that should be pretty straight forward. Thanks for getting this going @RDIL

PR #878, opened 2020-01-14
Bump tests to python 3.8, general tests fix

RDIL:
Tests on Python 3.8 since it is the newest version


sofroniewn:
Hi @RDIL thanks for submitting this. I know there were some PySide2 compatibility problems with 3.8 that were supposed to get resolved with their `5.14` release - see #766. Do we know if that has come out yet?

@jni @kne42 what do you imagine is our timeline for supporting 3.8 (i.e. bumping minimum Qt versions etc to accommodate). Also will we want to test on all three `3.6`, `3.7`, and `3.8`? That could take a long time for our CI to finish if we do it on all platforms. We could to all 3 on linux and then only 1 on windows and mac, giving us only 5 tests in total - which might be nicer. I don't think I've ever seen a bug that was platform + python version specific, though it doesn't mean it couldn't happen!!

jni:
I haven't looked at the test failures, but the 3 + 1 + 1 test matrix you suggest sounds good to me.

Yes, PySide2 and PyQt 5.14 are both out, so hopefully that means we can pass on 3.8 now...

.cirrus.yml
```diff
@@ -99,6 +99,7 @@ windows_task:
     matrix:
       PY_VER: "3.6"
       PY_VER: "3.7"
+      PY_VER: "3.8"
```
kne42: could you add this to the linux/osx tests too? (line 37)

.cirrus.yml
```diff
@@ -1,9 +1,10 @@
+container:
```
kne42, line 2: didn't know this, cool!

kne42 (approved):
Thanks for the contribution @RDIL! I've conditionally approved this :)

kne42:
oh, just saw the above discussion (oops!)

I'd prefer to test on all platforms. if you're concerned about too many tests being run, we can always set up [compute credits](https://cirrus-ci.org/pricing/#compute-credits) for priority queuing and guaranteed parallel tests (instead of being limited by the number of machines we can use at a time)

sofroniewn:
Testing the full grid is fine if we use compute credits - look like we got some for free right @kne42? Can you set it up so we use those first and then we get an estimate of how much we need to spend per month  and then we can figure out if Stephani can pay for it

RDIL:
Yes, you have received 60 credits for free.

kne42:
Ok, I can do that. @sofroniewn to what extent do we want to use compute credits?

- what tasks? [just the main tests, or every task, or specific tasks (e.g. linux doesn't take very long to run so maybe we don't need to use credits)]
- what users? [collaborators, core devs, or everyone]

sofroniewn:
Some back of the envelope calculations are - we get  1000 minutes / 10 windows credits (or 5 linux, 30 OS X), it's 1 credit per dollar - we might have 3 tests per platform, for 3 platforms, each one taking around 10 minutes, so say 60 per PR at the max rate => 1.35$ per PR, which isn't enormous but isn't 0$ either.

I feel right now I'm very liberal with pushing changes to a PR even if I know they aren't done because I want the code up there for people to see etc. and if we start using credits people might feel inhibited about doing that, which I don't think is great.

Maybe let's hear from @jni before deciding anything (and someone should maybe double check my numbers / assumptions / maths above)

sofroniewn:
Another approach is to have the full 3x3 grid run only on merge into master, and the 3 + 1 + 1 set run on every branch PR. This could have the advantage of keeping things fast, but making sure we catch errors if they do creep in.

For the 3 + 1 + 1 I'd go for 3.6, 3.7, 3.8 on Linux and 3.8 on OSX, 3.8 on windows

I'd then stay away from the credits

jni:
Yeah I'm into @sofroniewn's suggestion to only do the full matrix on master. I agree that paying per build can act as a disincentive to "push early, push often", which is important.

sofroniewn:
Ok great, @RDIL if you could update this PR so that we do the 3x3 grid run only on master and the  3 + 1 + 1 otherwise (with 3.6, 3.7, 3.8 on Linux and 3.8 on OSX, 3.8 on windows) that would be great.

@kne42 can give you a hand if you're unsure about anything.

Right now it looks like the 3.8 windows is failing, not sure why - it would be good to know that 3.8 tests pass locally with PySide2 as well - see #766 - and make sure our minimum version requirements are in line with this

.cirrus.yml
```diff
@@ -90,23 +94,28 @@ task:
...
-  
+
   windows_container:
     image: cirrusci/windowsservercore:cmake
     os_version: 2019
-    
+
+  matrix:
+    - only_if: $CIRRUS_PR != ""
+    - only_if: $CIRRUS_PR == ""
```
tlambert03, line 135: can you comment the desired behavior for these conditionals?

sofroniewn:
I think it would still be nice to have our checks in cirrus listed as `OS X 3.8`, `Windows X 3.8`, `Linux 3.8`, `Linux 3.7`, `Linux 3.6` and have 5 of them (excluding the flake, codecov, and black tests). That will really quickly let us see if the issue is platform specific or python version specific.

Also curious where we are landing with the PyQt5 testing @tlambert03? Seems like with #887 we are skipping PyQt5 tests on OSX, and we already skip them on windows right? Because we also skip all Qt tests on windows right? In which case I then might also make one `Linux 3.8 PyQt5` here instead of having both the PyQt5 + PySide2 inside all 3 linux tests? The disadvantage is a slight reduction in coverage (i.e. no 3.6 for PyQt5) but the advantage is that it is really clear why things are consistently failing just from looking at the top line items, clear what we are testing from the top line items, and all our tests will complete much faster. In that case this PR would supersede #887. Thoughts @tlambert03?

tlambert03:
> I think it would still be nice to have our checks in cirrus listed as `OS X 3.8`, `Windows X 3.8`, `Linux 3.8`, `Linux 3.7`, `Linux 3.6` and have 5 of them (excluding the flake, codecov, and black tests). That will really quickly let us see if the issue is platform specific or python version specific.

Yeah, I also prefer the old way that checks were reported.  @RDIL did you change that intentionally because of the discussion above about compute credits?  Or was that an unintended side effect of the syntax change you made with the `env matrix`

> Also curious where we are landing with the PyQt5 testing @tlambert03? Seems like with #887 we are skipping PyQt5 tests on OSX, and we already skip them on windows right? Because we also skip all Qt tests on windows right? In which case I then might also make one `Linux 3.8 PyQt5` here instead of having both the PyQt5 + PySide2 inside all 3 linux tests?

#887 and skipping tests on OSX was more of a bail-out than anything, because of the end-of-test crashes.  And after looking into it some more, it is a _dense_ issue that I think has to do with platform- and qt-backend-specific ways of dealing with threading and desktop virtualization.  (I have some more info and observations there, but nothing approaching a solution).  It would be absolutely fine with me if we wanted to have this PR supercede that one and only test `PyQt5` on linux for now.  The only thing I would mention is that if you pull it out into its own test, while you will gain the nice "PyQt5" label in the check title, it will add a significant amount of startup time to the test while it builds the environment, whereas if we just add one more script to the linux tests with a conditional that says "`if $OS == linux and $PY_VER == 3.8`", then the check is super fast (though you'd have to dig into the Linux test to see what was failing if anything).

tlambert03 (changes requested):
i'm posting a "request changes" review just so that the merge button is disabled for now.  some of this would change a pretty core part of our test pipeline and there's plenty more to be discussed here before merging.

RDIL:
I will try to do this later today.

.cirrus.yml
RDIL: done

.cirrus.yml
```diff
@@ -30,13 +31,21 @@ nix_test_task:
 task:
   auto_cancellation: true
 
-  env:
-    matrix:
-      PY_VER: "3.6"
-      PY_VER: "3.7"
+  matrix:
```
kne42, line 87: could you refactor this out of the windows/(linux+osx) tasks into one place using YAML aliases?

.cirrus.yml
```diff
@@ -30,13 +31,21 @@ nix_test_task:
...
+        matrix:
+          PY_VER: "3.6"
+          PY_VER: "3.7"
+          PY_VER: "3.8"
+    # is a pull request, only run python 3.8
+    - only_if: $CIRRUS_PR != ""
+      env:
+        PY_VER: "3.8"
 
   matrix:
```
kne42: wow, I'm surprised that [this worked](https://cirrus-ci.com/build/6008271744008192) since in normal YAML, it would've been overwritten. good to know!

.cirrus.yml
```diff
@@ -88,25 +97,35 @@ task:
...
-      PY_VER: "3.7"
-    ANACONDA_LOCATION: $USERPROFILE\anaconda
-    PATH: $ANACONDA_LOCATION\Scripts;$ANACONDA_LOCATION;$PATH
-    # must set this to its default since it doesn't work in env variables
-    # see https://github.com/cirruslabs/cirrus-ci-docs/issues/423
-    CIRRUS_WORKING_DIR: C:\Users\ContainerAdministrator\AppData\Local\Temp\cirrus-ci-build
-    # https://github.com/vispy/vispy/blob/v0.5.3/appveyor.yml#L44
-    VISPY_GL_LIB: $CIRRUS_WORKING_DIR\opengl32.dll
-    PYTHON_ARCH: 64
+      ANACONDA_LOCATION: $USERPROFILE\anaconda
```
kne42: could you fix the indentation here?

kne42 (approved):
fantastic job @RDIL! left a few small comments :)

@ others, what do you want to do about the 3.8 testing (since it fails)? do we wait until we fix all 3.8 bugs until we merge this?

we could also enable 3.7/3.8 for PRs temporarily until we do fix these issues, and then test on 3.8 only

tlambert03:
Thanks again @RDIL!

@sofroniewn said above:
>I think it would still be nice to have our checks in cirrus listed as OS X 3.8, Windows X 3.8, Linux 3.8, Linux 3.7, Linux 3.6 and have 5 of them (excluding the flake, codecov, and black tests). That will really quickly let us see if the issue is platform specific or python version specific.

So I thought the idea was to have at least one platform testing all three supported python versions, even during PRs.  Also, I love being able to minimize the test matrix during PRs, but what happens if we merge something to master, the full test suite is triggered, and we find a legitimate issue that didn't show in our minimal tests?  Does that mean we're merging bugs into master that then need to be fixed by a follow up PR?  Is my understanding correct on that?  If everything thinks that's likely to be super rare, I suppose it's ok... but curious to hear thoughts.

>do we wait until we fix all 3.8 bugs until we merge this?

Since the point of this PR was to bump to testing 3.8, I do think we should make sure that tests actually work on 3.8 before merging.  here's what's currently broken

- [linux](https://cirrus-ci.com/task/4875001270632448): this showed the old intermittent fail in the `test_play_api`... but that should be fixed in master, so merging back in the master branch should fix that.  I'm not sure what the `AttributeError` at the end of the test is from... but if that's a 3.8 issue, then we should figure it out before merging.
- [macos](https://cirrus-ci.com/task/6000901177475072): this one passed everything and then had the old abort trap at the end of the pyqt5 tests.  As mentioned in #887, we were simply going to no longer test PyQt5 on osx for now, and I'm ok with that... since that is a sticky problem that doesn't seem to actually affect anyone using the program in non-remote environment.
- [windows](https://cirrus-ci.com/task/5437951224053760): this one is more concerning, and has a lot of `NotImplemented` errors that seem to be coming from `asyncio`.  Not something I have any familiarity with... but may be indicating that at least one of our dependencies isn't ready to support python 3.8 on windows?  It also throws the json `AttributeError` that the linux test threw.

.cirrus.yml
RDIL: I can try

.cirrus.yml
RDIL: Before this, the file had really strange indentation, so do we want to go with 2 or 4 spaces?

tlambert03: 2 seems to be standard for YAML, so go ahead and clean up anywhere we had 4.

.cirrus.yml
RDIL: I ended up matrixing it all

sofroniewn:
> Also, I love being able to minimize the test matrix during PRs, but what happens if we merge something to master, the full test suite is triggered, and we find a legitimate issue that didn't show in our minimal tests? Does that mean we're merging bugs into master that then need to be fixed by a follow up PR? Is my understanding correct on that? If everything thinks that's likely to be super rare, I suppose it's ok... but curious to hear thoughts.

I think that is correct. We will have to then be somewhat vigilant about checking if tests fail after merge and why (which isn't a bad thing anyway), and make absolute sure we've checked all of this before a release.

> Since the point of this PR was to bump to testing 3.8, I do think we should make sure that tests actually work on 3.8 before merging

That would be nice @tlambert03 - can you work on that either in a PR to this PR / commit directly to this PR if @RDIL can enable edits from maintainers. Otherwise you could make a separate PR, but then there is a bit of a chicken and an egg thing. It will be fairly clear that those edits to the codebase are for the 3.8 compatibility.

> The only thing I would mention is that if you pull it out into its own test, while you will gain the nice "PyQt5" label in the check title, it will add a significant amount of startup time to the test while it builds the environment

Ok - let's do PyQt5 tests on all the linux tests 3.6, 3.7, 3.8 - it shouldn't be that big a deal to drop in and see what's broken when there are errors, and if we don't like that we can change in the future to pull them out separately.

RDIL:
I have it on, but now I'm confused - can you explain what you need me to do from here @sofroniewn?

RDIL:
You guys realize you can make pre-setup environments will all the tools you need installed, right?

Just making sure, and if not I would be more then happy to implement it.

tlambert03:
> You guys realize you can make pre-setup environments will all the tools you need installed, right? Just making sure, and if not I would be more then happy to implement it.

would be happy to see what you're proposing.  Our current strategy mimics what a user would experience when they `pip install napari` (including dependencies). If you mean perhaps a docker environment or a cirrus cache or something like that with a set of pinned dependencies pre-installed, then one of the concerns I would have is that we would miss breaking changes due to dependency updates (see for example #868 or #665).

RDIL:
Yeah, but you can update them every month or so.

tlambert03:
at this phase in the project, I don't think we want to allow the possibility that our `pip install` is broken for weeks, and the slightly longer build time in our tests is not such a big compromise for us to be confident that we are seeing what our users are seeing when they run `pip install`.

tlambert03:
> I have it on, but now I'm confused - can you explain what you need me to do from here @sofroniewn?

I think what we are looking for, is what @sofroniewn said here https://github.com/napari/napari/pull/878#issuecomment-575937081.  We want these 5 tests:
- Linux with pyside2 and pyqt5 on python 3.6, 3.7, 3.8
- osx with pyside2 (not pyqt5) on python 3.8
- windows with pyside2 (pyqt5 iff it is working) on python 3.8
- flake8, black, codecov

regarding 3.7 vs 3.8 ... I think we all liked the possibility of bumping our tests to 3.8 as you introduced it, provided we are actually capable of supporting 3.8.  And I think it's worth trying to debug that a bit at the moment.  So I would vote for leaving 3.8 as the default, and let's see if we can't debug it here in this PR.  If you can return it to 3.8, now that you've merged the master branch back in, I'd be happy to take a look to see what residual problems we are still having on 3.8.

tlambert03:
ok I have a bit more information on this.  We can't officially support python 3.8 yet on windows since tornado is still having issues (which is required for ipykernel and qtconsole).  See this thread: https://github.com/tornadoweb/tornado/issues/2608
and many of the references to that issue at the bottom are what we are also seeing with asyncio, like [this](https://github.com/jupyter/qtconsole/issues/405).  I'm preparing a modified test suite and will push it to this PR shortly.

It's unfortunately not as clean as the condensed syntax that @RDIL has nicely laid out here, but it works, and we need to get our tests back in working order ASAP.

we _could_ put in some try/catch statements that would allow everything but the console to work, but i'm not sure it's worth all the work just for windows on 3.8, with something that will likely get fixed soonish anyway.

RDIL:
I can set the tests to skip if you tell me which file its in

tlambert03:
there were a number of other things needing attention... i have a seperate fork I'm testing at the moment ... if these tests all pass, I will either PR to your PR, or just merge it here if you've set those permissions.
https://github.com/tlambert03/napari/pull/5/checks?check_run_id=411757924

tlambert03:
ok, all desired tests are now [working](https://cirrus-ci.com/build/5724010197549056).

I'm very aware that this is not a "pretty" file, and I had to break all of the lovely env matrices that you made to achieve what we need our tests to be.  If you have the time, feel free to try to reduce code again with matrices, but here are the things that cannot be compromised:

- linux should be tested on py3.6, py3.7, and py3.8, always with both pyside2 and pyqt5, regardless of whether it is a PR or not.  I tried a lot of things to get this to work starting with the matrices you had ... but couldn't get it and gave up.
- mac should be test on py3.8 for PRs, and 3.6/3.7/3.8 for master branch. always only on pyside2.
- windows should be tested on py3.7 for PRs, and 3.6/3.7 for master branch (until #915 merges).  always only on pyside2.  Note, I had to break the matrix again for this one because it's critical that the test script for windows be "`pytest -v -o qt_no_exception_capture=1`", and I wasn't able to figure out a cross-platform bash conditional (if $OS == 'windows')... feel free to try re-matrixing it, but that line cannot be removed.
- note I added a `pip_cache` with a platform-specific folder ... so if you want to rematrix things, make a new env var for that folder.
- to reduce the number of tasks, I combined black and flake8
- also note, I hard-coded conda=4.8.1, because 4.8.2 recently broke our environment building on python 3.6

let me know if you have time and want to try to clean any of this stuff up, or if we should just go with the ugly but functional version here.  thanks!

.cirrus.yml
```diff
@@ -1,104 +1,152 @@
...
-  auto_cancellation: true
+# default container
+container:
+  image: python:3.8
 
-  name: Black Formatting
-  container:
-      image: python:3.6
 
+black_task:
```
RDIL, line 6:
```suggestion
task:
```

Avoid doing this so the parser has less work to do

.cirrus.yml
```diff
@@ -1,104 +1,152 @@
...
     - python -m pip install --retries 3 --upgrade pip
     - pip install `cat requirements/development.txt | grep black`
-
   test_script:
     - black --check .
 
-nix_test_task:
-  auto_cancellation: true
 
+lint_task:
```
RDIL, line 16:
```suggestion
task:
```
same here

.cirrus.yml
```diff
@@ -1,104 +1,152 @@
...
   install_script:
     - python -m pip install --retries 3 --upgrade pip
     - pip install `cat requirements/development.txt | grep flake8`
-
   test_script:
     - flake8 napari
 
 
-task:
+linux_task:
```
RDIL, line 26: and here

.cirrus.yml
```diff
@@ -1,104 +1,152 @@
...
+    - conda install -y python=$PY_VER
+    - rm install.sh
+
+  # https://github.com/cirruslabs/cirrus-ci-docs/issues/97
+  xvfb_start_background_script: Xvfb $DISPLAY -ac -screen 0 1024x768x24
+  # https://pytest-qt.readthedocs.io/en/latest/troubleshooting.html#xvfb-assertionerror-timeouterror-when-using-waituntil-waitexposed-and-ui-events
+  herbstluftwm_start_background_script: herbstluftwm
+
+  pip_cache:
+      folder: ~/.cache/pip
```
RDIL: Indents

.cirrus.yml
```diff
@@ -1,104 +1,152 @@
...
+    - python -m pip install --retries 3 --upgrade pip
+    - pip install --retries 3 -r requirements.txt
+    - python setup.py develop
+
+  check_script: python -c "import napari; print(napari.__version__)"
+
+  test_script: pytest -v
+
+
+win_task:
```
RDIL, line 130: lastly, here

RDIL (commented):
@tlambert03 some feedback on your changes

.cirrus.yml
tlambert03: seems strange... in valid yaml, mapping keys are unique (even if it still works)... what is the "cost" of the parser having more work to do?  are we talking potential bugs?  or a tiny bit of time?  and is there docs somewhere describing this concept?

tlambert03: not to mention that's how cirrus encourages it [in their own docs](https://cirrus-ci.org/guide/writing-tasks/)... would need to see the reference for the downside of having unique keys like this

RDIL: Never mind

tlambert03:
> @tlambert03 some feedback on your changes

thank you, would you like a crack at the matrices again?  or shall we merge once tests pass?

tlambert03:
@napari/core-devs, I'm approving this so we can merge and get our tests/CI back in order... but i'll leave it and allow someone else to merge just in case anyone wants to comment.

tlambert03 (approved):

sofroniewn:
@tlambert03 if we merge #915 first would this change? Or do you want to merge this first? I'm fine with either order, or let me know if they are independent

sofroniewn:
I think we should merge this after all the tests pass, thanks for all the hard work @RDIL,  @tlambert03 and others!! Will be very nice to have this in :-)

royerloic:
Great!

PR #880, opened 2020-01-14
Clean up setup.py

ttung:
# Description
1. napari-gui is gone.
2. collapse comment removal from requirements.txt into one place.
3. [setup()](https://packaging.python.org/guides/distributing-packages-using-setuptools/#setup-args) doesn't have a `requires` kwarg.
4. use [literal string concatenation](https://docs.python.org/3/reference/lexical_analysis.html#string-literal-concatenation).  IDEs know how to deal with literal string concatenation better.
5. move the block of string constants after the imports for PEP8 compliance.

## Type of change

# How has this been tested?

```
mkdir /tmp/i
cd /tmp/i
python3.6 venv .venv
pip install -e ~/imaging/napari
```

## Final checklist:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,8 +1,17 @@
 #!/usr/bin/env python
-"""Napari GUI
-
-GUI component of Napari.
+"""Napari viewer is a fast, interactive, multi-dimensional image viewer for
+Python. It's designed for browsing, annotating, and analyzing large
+multi-dimensional images. It's built on top of `Qt` (for the GUI), `vispy`
+(for performant GPU-based rendering), and the scientific Python stack
+(`numpy`, `scipy`).
 """
+
+import os.path as osp
+import sys
+from setuptools import find_packages, setup
+
+import versioneer
+
 
 import os.path as osp
 import sys
@@ -10,7 +19,9 @@ from setuptools import setup, find_packages
 
 import versioneer
 
-MIN_PY_VER = '3.6'
+MIN_PY_MAJOR_VER = 3
+MIN_PY_MINOR_VER = 6
+MIN_PY_VER = f"{MIN_PY_MAJOR_VER}.{MIN_PY_MINOR_VER}"
 DISTNAME = 'napari'
 DESCRIPTION = 'n-dimensional array viewer in Python.'
 LONG_DESCRIPTION = __doc__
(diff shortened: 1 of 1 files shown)


sofroniewn (commented):
This looks good to me, though maybe @kne42 can take a look too. I think we'll still want to merge #888 and #887 first (which may cause a tiny merge conflict)

setup.py
```diff
@@ -4,6 +4,13 @@
 GUI component of Napari.
```
jni: Might as well update this docstring? =)

jni (approved):

setup.py
ttung: Done.

setup.py
```diff
@@ -1,9 +1,14 @@
 #!/usr/bin/env python
-"""Napari GUI
-
-GUI component of Napari.
+"""Napari Viewer
```
kne42: could you replace this with a subset of our README? it will be used as part of the pip description

setup.py
```diff
@@ -34,61 +39,33 @@
...
 
-
-INSTALL_REQUIRES = []
-REQUIRES = []
-
-for l in requirements:
-    sep = l.split(' #')
-    INSTALL_REQUIRES.append(sep[0].strip())
-    if len(sep) == 2:
-        REQUIRES.append(sep[1].strip())
```
kne42, line 71: idk why everyone removed the comments from the requirements file :(

kne42 (commented):
Thanks for the contribution @ttung!

I'm confused why you removed the `packages` kwarg; it's listed in the link you provided ([link to `packages` section](https://packaging.python.org/guides/distributing-packages-using-setuptools/#packages)). Has it been made optional or something?

setup.py
ttung: Done.

setup.py
ttung: I re-added the logic to strip comments, but integrated the comment-at-start-of-line and the comment-at-end-of-line logic into one thing.

ttung:
> I'm confused why you removed the packages kwarg; it's listed in the link you provided (link to packages section). Has it been made optional or something?

LOL i'm so bad at this.
I meant to remove the `requires` kwargs.  `packages` is legit.

kne42:
ooooh ok that makes sense!

sofroniewn:
I have now merged #877 with the desired language there. I had to move some of the imports around too. Can you fix the merge conflicts @ttung and then I think this will be ready to merge. Thanks for your patience

ttung:
> I have now merged #877 with the desired language there. I had to move some of the imports around too. Can you fix the merge conflicts @ttung and then I think this will be ready to merge. Thanks for your patience

Updated.

setup.py
```diff
@@ -39,52 +48,37 @@
     'Operating System :: MacOS',
 ]
 
+
 if sys.version_info < (3, 6):
```
tlambert03: could we get `MIN_PY_VER` in here?

ttung: Updated.

sofroniewn (approved):
Thanks for the updates @ttung - this looks good to me. I will let sit for a bit in case anyone else wants to weigh in and then merge.

jni (approved):

PR #884, opened 2020-01-16
improve slider step precision based on data range

### LINKED ISSUE #882: contrast limits slider increment is static and too large for some images
## 🐛 Bug

When I open images with intensities occupying the low end of [0,1] range, the contrast limits slider  increments of 0.01 are too large to be useful.  Like I only get 2-3 positions for some of my recent images.

## To Reproduce

1. open an image with intensity values _val_ s.t.   0 < _val_ < 0.05 
2. adjust the clims
3. frown

## Expected behavior
The user shouldn't see slider increments unless the underlying control is of something with a few discrete values. In this case, I suggest they should scale to ~5% of the range for the small sliders and < 1% for the big sliders.

## Environment
napari==0.2.10+4.gb45a26d

## In other news:

I like the expando-slider and support their further expansion across the landscape of napari!

tlambert03:
# Description
this fixes #882 by guessing an appropriate slider step size (and number of decimal places for the popup clims control) based on the current contrast_limits_range. 

## Type of change

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/_qt/layers/qt_image_base_layer.py
+++ b/napari/_qt/layers/qt_image_base_layer.py
@@ -161,7 +161,11 @@ def create_range_popup(layer, attr, parent=None):
         precision=(
             0
             if is_integer_type
-            # scale precision with the log of the data range
+            # scale precision with the log of the data range order of magnitude
+            # eg.   0 - 1   (0 order of mag)  -> 3 decimal places
+            #       0 - 10  (1 order of mag)  -> 2 decimals
+            #       0 - 100 (2 orders of mag) -> 1 decimal
+            #       ≥ 3 orders of mag -> no decimals
             else int(max(3 - np.log10(max(d_range[1] - d_range[0], 0.01)), 0))
         ),
         parent=parent,
--- a/napari/_qt/layers/qt_image_base_layer.py
+++ b/napari/_qt/layers/qt_image_base_layer.py
@@ -153,11 +153,17 @@ def create_range_popup(layer, attr, parent=None):
         )
     is_integer_type = np.issubdtype(layer.dtype, np.integer)
 
+    d_range = getattr(layer, range_attr)
     popup = QRangeSliderPopup(
         initial_values=getattr(layer, attr),
-        data_range=getattr(layer, range_attr),
+        data_range=d_range,
         collapsible=False,
-        precision=(0 if is_integer_type else 2),
+        precision=(
+            0
+            if is_integer_type
+            # scale precision with the log of the data range
+            else int(max(3 - np.log10(max(d_range[1] - d_range[0], 0.01)), 0))
+        ),
         parent=parent,
     )
 
--- a/napari/_qt/qt_range_slider.py
+++ b/napari/_qt/qt_range_slider.py
@@ -77,7 +77,13 @@ class QRangeSlider(QWidget):
 
         self.setRange((0, 100) if data_range is None else data_range)
         self.setValues((20, 80) if initial_values is None else initial_values)
-        self.setStep(0.01 if step_size is None else step_size)
+        if step_size is None:
+            # pick an appropriate slider step size based on the data range
+            if data_range is not None:
+                step_size = (data_range[1] - data_range[0]) / 1000
+            else:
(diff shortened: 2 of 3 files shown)


sofroniewn (approved):
This looks good to me! Thanks for getting this done so quickly @tlambert03 - does @berl want to give it a try before we merge?

napari/_qt/layers/qt_image_base_layer.py
```diff
@@ -153,11 +153,17 @@ def create_range_popup(layer, attr, parent=None):
...
         initial_values=getattr(layer, attr),
-        data_range=getattr(layer, range_attr),
+        data_range=d_range,
         collapsible=False,
-        precision=(0 if is_integer_type else 2),
+        precision=(
+            0
+            if is_integer_type
+            # scale precision with the log of the data range
+            else int(max(3 - np.log10(max(d_range[1] - d_range[0], 0.01)), 0))
```
jni, line 169: I'd love for the comment to explain this expression a bit more...!

napari/_qt/qt_range_slider.py
```diff
@@ -77,7 +77,13 @@ def __init__(
 
         self.setRange((0, 100) if data_range is None else data_range)
         self.setValues((20, 80) if initial_values is None else initial_values)
-        self.setStep(0.01 if step_size is None else step_size)
+        if step_size is None:
+            # pick an appropriate slider step size based on the data range
+            if data_range is not None:
+                step_size = (data_range[1] - data_range[0]) / 1000
+            else:
+                step_size = 0.001
```
jni, line 85: I feel like 100 steps was a better default than 1000...

jni (approved):
@tlambert03 I suppose this needs a couple of tests (with big floats and small floats), and I'd love the comments to explain the logic a bit better, but otherwise great, as @sofroniewn said, thanks for the :zap: fast turnaround!

napari/_qt/layers/qt_image_base_layer.py
tlambert03: added a comment, let me know if it's clear enough:
https://github.com/napari/napari/blob/3f391b6a36e002f4f7ecc4273d0de4e388110609/napari/_qt/layers/qt_image_base_layer.py#L164-L169

point is, we want the editable labels to have many decimal places if the data range is very small, and no decimal places if we have data than spans many orders of magnitude, but with upper (5) and lower (0) limits of decimal places.

napari/_qt/qt_range_slider.py
tlambert03: I actually tried a few values, and while 100 is sufficient for the contrast limits sliders in the main controls, it is noticeably choppy for the popup slider (try the `example/add_grayscale_image`), even 500 feels unsmooth.  And I couldn't think of any particular performance or other downside to giving the slider finer grained control.  sound ok?  or are you thinking of a downside?

jni: Ah, I forgot about the big slider. =D

Well, as I understand it, and correct me if I'm wrong, we're triggering some brief but nonzero computation on each tick of the slider, and currently we don't do this asynchronously, so having too many stops will result in choppiness?

But, you're the one actually playing around with this, so, ignore me! =P

jni:
All good from me now! 🎉 Thanks!

napari/_qt/qt_range_slider.py
tlambert03: let me try to raise it to a ridiculous amount and see if the handle seems to noticeably lag behind the mouse as a I raise it.  If I learn something, i'll update the PR, otherwise, I'll just merge

tlambert03: I upped it to one _nonillion_ steps (1e30 😜) and didn't notice any observable difference.  One possibility is that the pixel pitch of the monitor/QCursor position is limiting signals to something quite reasonable?

jni: 😆 😆 😆 😆  That's awesome!

berl:
whoa.  that was _crazy_ fast.  Thanks @tlambert03 !

Issue #614, opened 2019-10-22
Scale not respected in 2D display


jni:
## 🐛 Bug

When viewing an anisotropic volume in 2D, transposing the axes should result in a properly-scaled display of the cross-section.

## To Reproduce

1. Download the cropped sample A volume from cremi.org: https://cremi.org/static/data/sample_A_20160501.hdf
2. Load and display it as follows:

```python
import h5py
import napari

# volume is 125, 1250, 1250 with 10, 1, 1 anisotropy
raw = h5py.File('path/to/sample_A_20160501.hdf', mode='r')['/volumes/raw']
viewer = napari.view_image(raw, scale=[10, 1, 1])
```

3. Push the "roll dimensions order for display" button

<img width="934" alt="Screen Shot 2019-10-22 at 8 02 36 pm" src="https://user-images.githubusercontent.com/492549/67272040-79e2e580-f507-11e9-9aaf-971e380dd764.png">

4. Weep. =P

## Expected behavior

The side view should be expanded with the appropriate interpolation mode. (which I expect will break the labels layer with anything other than nearest-neighbour)

sofroniewn:
Strange to me that this didn't work - i would have expected it to. I'll take a look when I have time

sofroniewn:
Looks like `_on_scale_change` isn't getting triggered on the vispy view when the roll happens - should be easy to fix

sofroniewn:
Note this bug is independent of addition of world coordinates

sofroniewn:
This was fixed by #1102 and a test added. @jni would you like to confirm this is now fixed and close? (I just checked and it looks fixed to me)

sofroniewn:
ping @jni - want to confirm this is fixed and then I can close?

jni:
Yep! Sorry for the delay, works great! 😊

PR #885, opened 2020-01-16
WIP: Proper models for world coordinates

jni:
# Description

Opening this PR very prematurely so we have something to discuss in lab meeting. Currently, there is just initial work on the models.

## Type of change

# References

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/layers/transforms.py
+++ b/napari/layers/transforms.py
@@ -65,6 +65,7 @@ class Translate(Transform):
         self.vector = np.array(vector)
 
     def __call__(self, coords):
+        coords = np.atleast_2d(coords)
         vector = np.concatenate(
             ([0.0] * (coords.shape[1] - len(self.vector)), self.vector)
         )
@@ -88,6 +89,7 @@ class Scale(Transform):
         self.scale = np.array(scale)
 
     def __call__(self, coords):
+        coords = np.atleast_2d(coords)
         if coords.shape[1] > len(self.scale):
             scale = np.concatenate(
                 ([1.0] * (coords.shape[1] - len(self.scale)), self.scale)
--- a/napari/layers/transforms.py
+++ b/napari/layers/transforms.py
@@ -1,0 +1,95 @@
+from __future__ import annotations  # noqa: F407
+
+import toolz as tz
+from typing import Sequence
+import numpy as np
+
+from ..utils.list import ListModel
+
+
+class Transform:
+    """Base transform class.
+
+    Defaults to the identity transform.
+
+    Parameters
+    ----------
+    func : callable, Coords -> Coords
+        A function converting an NxD array of coordinates to NxD'.
+    """
+
+    def __init__(self, func=tz.identity, inverse=None):
+        self.func = func
+        self._inverse_func = inverse
+        if func is tz.identity:
+            self._inverse_func = tz.identity
+
+    def __call__(self, coords):
+        """Transform input coordinates to output."""
+        return self.func(coords)
+
+    def set_slice(self, axes: Sequence[int]) -> Transform:
+        """Return a transform subset to the visible dimensions."""
+        raise NotImplementedError('Cannot subset arbitrary transforms.')
+
+    @property
+    def inverse(self):
+        if self._inverse_func is not None:
+            return Transform(self._inverse_func, self.func)
+        else:
+            raise ValueError('Inverse function was not provided.')
+
+
+class TransformChain(Transform, ListModel):
+    def __init__(self, transforms=[]):
+        super().__init__(basetype=Transform, iterable=transforms, lookup=id)
+
+    def __call__(self, coords):
+        return tz.pipe(coords, *self)
+
+    def set_slice(self, axes: Sequence[int]) -> TransformChain:
+        return TransformChain([tf.set_slice(axes) for tf in self])
+
+
+class Translate(Transform):
+    """n-dimensional translation (shift) class.
+
+    An empty translation vector implies no translation.
+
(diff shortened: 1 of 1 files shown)


napari/layers/transforms.py
```diff
@@ -0,0 +1,95 @@
+from __future__ import annotations  # noqa: F407
```
jni, line 1: Note: this is valid in Python 3.7 but PyFlakes doesn't seem to think so... I don't really understand why:

https://github.com/PyCQA/pyflakes/pull/374

sofroniewn:
`Translate` and `Scale` look good. Let me know if there are things you want to discuss in particular here, but as said at team meeting pressing on getting `Translate` and `Scale` onto the base layer and then the vispy base layer seems best.

sofroniewn:
Note we have the current bug #614 that you may encounter along the way. (I can try and fix that first if you want too - I think it's just a missing event call, or something simple)

napari/layers/transforms.py
```diff
@@ -0,0 +1,95 @@
...
+    def inverse(self):
+        if self._inverse_func is not None:
+            return Transform(self._inverse_func, self.func)
+        else:
+            raise ValueError('Inverse function was not provided.')
+
+
+class TransformChain(Transform, ListModel):
+    def __init__(self, transforms=[]):
+        super().__init__(basetype=Transform, iterable=transforms, lookup=id)
```
kne42: i dont think you want to specify a lookup table in this case

it was mainly meant for the layer list so it can be indexed by the layer names

jni:
@sofroniewn yeah the main issue I wanted to discuss was what I mentioned, that it seemed kinda crazy to duplicate all the Vispy transforms machinery, but your spiel about the models hit home. =) And by starting with just scale and translate, the task becomes much more manageable. =)

napari/layers/transforms.py
jni: Thanks for the tip! I didn't know whether I needed one regardless. I think eventually indexing by name might be very useful, e.g. `grid` for all the transforms related to grid mode, so that they can be popped in and out as needed.

sofroniewn:
I want to put a link to #763 here, just so it is easy to get too. I'm happy to work on this @jni, but I'm not quite sure what comes next. Maybe we can put a plan down in the comments and then work towards it. I think getting this in before #854 (the camera model) makes sense too. If I can think of a plan I'll start putting it in comments. Having this in `0.2.12` would be very nice

sofroniewn:
I think I'd also like to get this in before fixing #458 (which will probably come after #854)

jni:
@sofroniewn I certainly won't get to this till the weekend at the earliest, so if you are itching to push it forward, you could. Sorry that I haven't managed to put in the sustained effort to get this done. The basic idea I have for this is:

- be clear on what coordinates refer to. Don't use `coords`, use `data_coords`, `world_coords`, `sliced_world_coords`, `canvas_coords` (this is where you add the 0.5 offset), etc.
- Create the scale and translate models. The skeletons I've put in aren't yet feature complete. Probably adding a `__getitem__` to slice the visible dims would be useful, in addition to `set_slice`.
- Replace the vispy STTransform with a ChainedTransform that mirrors our model. It needs to mirror the full transform model chain, although this is in the *sliced* space (this goes back to @d-v-b's comment in #763), since you can't have a 4D transform in Vispy.
- On scale or translate or displayed change, the transform model gets resliced and the vispy transform views get updated.

To be clear, I'd love to sink my teeth into this, but just haven't found the right block of time to do it in. Let me know how you want to move forward... As I've mentioned before, I think this is one of our highest priorities...

sofroniewn:
@jni thanks for that plan. I'll wait before doing anything here, but nice to think about how we might be able to work together on some of it

napari/layers/transforms.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def __call__(self, coords):
+        vector = np.concatenate(
+            ([0.0] * (coords.shape[1] - len(self.vector)), self.vector)
+        )
+        return coords + vector
+
+    @property
+    def inverse(self):
+        return Translate(-self.vector)
```
GenevieveBuckley: I think you need something like `return Translate(-np.array(self.vector))` here instead. 
Right now I get a "TypeError: bad operand type for unary -: 'tuple'" (which makes sense, `self.vector` is actually a tuple).

GenevieveBuckley:
@jni can you tell me how these new transform classes fit in with the vispy transforms? I'm having trouble seeing how it should all fit together.

FYI following up on our hack day: I tried replacing the `vispy.visuals` `STTransform` with `ChanTransform` instead, but `ChainTransform` does not have the same `.scale` and `.translate` attributes available. I think this means we'll need a lot more work on the `_master_transform()` method in `vispy_base_layer.py`.
Do you have any suggestions for what I should try next?

GenevieveBuckley:
Notes (mostly to myself) from the hack day today:
* We did some work on the transform classes, will probably be merged here at some stage: https://github.com/jni/napari/pull/3/files
* Next steps: Hook up emitters for events. Can use the examples in these files for reference:
    * `napari/_qt/qt_layerlist.py`
    * `layerlist.py`
    * `napari/utils/list/_model.py`
* Next steps: `TransformChain` class  needs an `append()` method (also with an emitter for events)
* Next steps: replace vispy `STTransform` with `ChainTransform` in `_master_transform` and update setting the scale and translate properties (so we don't break everything downstream).

sofroniewn:
This has been superseded by #1018 and #1042 which added Transform and ChainTransform classes respectively and can now be closed. Thanks for all the work on it @jni and @GenevieveBuckley

PR #887, opened 2020-01-17
skip pyqt5 tests on osx py3.7

tlambert03:
# Description
tests are consistently failing on OSX py3.7 with PyQt5.  They fail with an Abort trap after all tests are complete, which might suggests that a thread has not been completely cleaned up?  However, I've been unable to reproduce this locally, making it really hard to debug, and we're already testing pyqt5 on py3.7 on linux, and OSX on py3.6 ... so for now this backs off a bit.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -88,8 +88,6 @@ task:
       pip install pyqt5
       pytest -v
     fi
-  
-
 
 
 windows_task:
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -82,10 +82,14 @@ task:
   test_pyside2_script:
     - pytest -v
 
-  test_pyqt5_script:
-    - pip uninstall -y pyside2
-    - pip install pyqt5
-    - pytest -v
+  test_pyqt5_script: |
+    if [ "$PY_VER" == "3.6" ] || [ $OS == "linux" ]; then
+      pip uninstall -y pyside2
+      pip install pyqt5
+      pytest -v
+    fi
+  
+
 
 
 windows_task:


tlambert03:
Lol.  I think that’s the first time 3.6 has failed?  
Ok, I’ll try to figure out what threads are left dangling that might be crashing these tests at the very end. Happy to have any suggestions: @AhmetCanSolak, @kne42, @ttung?

ttung:
Is it possible to bisect which test(s) is causing the problem?

tlambert03:
by "bisect" do you mean like "git bisect"?  If so... I don't think so, since we haven't been testing pyqt5 on OSX until very recently... we've always only tested pyside2 (where the problem doesn't happen).

I was eventually going to get around to manually enabling/disabling certain tests to see if I could narrow it down...  But, as I mentioned, all of the tests "pass" as far as pytest is concerned, and I can't reproduce this locally... so it's slow-going since every test takes some 10-15 minutes. 

if you mean something else/specific by bisect, let me know

ttung:
> I was eventually going to get around to manually enabling/disabling certain tests to see if I could narrow it down... But, as I mentioned, all of the tests "pass" as far as pytest is concerned, and I can't reproduce this locally... so it's slow-going since every test takes some 10-15 minutes.

you can disable half the tests and see which side repros.
perhaps faster is generating N git branches, where each branch only contains one test file.  fire it all off on circle ci, and come back after a long coffee break. :)

jni:
@tlambert03, have you tried pytest-xdist and the `--forked` option?

https://stackoverflow.com/questions/48234032/run-py-test-test-in-different-process

🤞

tlambert03:
> you can disable half the tests and see which side repros.
> perhaps faster is generating N git branches, where each branch only contains one test file. fire it all off on circle ci, and come back after a long coffee break. :)

that's a good idea.  One additional detail here is that it's also non-deterministic.  There's some sort of race condition and it doesn't always fail. but I like the massive multi-branch idea.  I'll give it a shot.

@jni ... no haven't tried xdist `--forked`... will read up and try it too

tlambert03:
closing this in favor of the testing discussions going on in #878.  
for what it's worth, trying xdist `--forked` locally raised all hell.  Over 150 of our tests failed.  Some of those failures were probably because they needed data that was instantiated in some other process... but some of them were definitely not. For instance, simply creating a test that calls `vispy.sys_info()` fails with a segmentation fault (which originates in `vispy.util.dpi._quartz.get_dpi()` during the call to `CGDisplayScreenSize` for what it's worth).  I don't know for sure if that's along the same lines as the segfaults seen on CI, but it seems possible... I don't have the time to dig into that sort of thing right now.

PR #891, opened 2020-01-18
[WIP proposal] IO plugin

tlambert03:
# Description
This PR implements a basic IO plugin architecture that enables users to control how a filepath dropped into the napari viewer is handled.  I see that there was once a [napari-io](https://github.com/napari/napari-io) repo and I suspect there has been plenty of discussion around io plugins, so forgive the late-entry and please take this PR simply as a working proposal very much open for criticism.

### Rationale

I see input plugins as very high priority, since there may be many potential users who simply want to view their images but don't necessarily know how to construct the appropriate numpy/dask array from their image (or other) files.  They should just be able to drag and drop their folders onto napari.  Also, there are many stereotyped microscopy image folder structures (like lattice data, or metamorph data, or basically any folder of tiffs with a naming convention), that could easily be made to look better in napari when dropping the folder on to the viewer.  

IO plugins may also be easier for us figure out than more "general" plugins as in #263, since their usage is  much more defined in some ways (e.g. take a filepath, and add it to the viewer)

### Design of this PR

this PR introduces a couple things:

- `napari.plugins.PluginManager`: this class finds all available napari plugins in the environment.  This differs slightly (but is not incompatible) with the `plug.crawl` discovery method that @jni started in #263, and is loosely based on the pytest plugin manager.  The goal here was to make plugins discoverable by simply being available in the environment.  `PluginManager. discover_plugins` has the main discovery logic and uses two discovery methods:
  - [naming convention](https://packaging.python.org/guides/creating-and-discovering-plugins/#using-naming-convention) - where a plugin is simply named something like `napari_myplugin` or a subclass like `napari_io_myplugin`
  - [package metadata](https://packaging.python.org/guides/creating-and-discovering-plugins/#using-package-metadata) - where a package declares a special key (like "`napari.plugins`") in their `entry_points` in setup.py.  This is one way that [pytest does it](https://docs.pytest.org/en/latest/writing_plugins.html#making-your-plugin-installable-by-others).

  the plugin manager can also search pypi for packages following the naming convention (and something like github gists could also be added).  There are a couple notes there about plugin "subtypes" (e.g. IO plugins vs other plugins), but I haven't fleshed that out yet.  But for now, `PluginManager.readers` identifies and yields plugins that follow the IO plugin convention described below.

- `napari_io_lls` - this is an example package that gets discovered using the "naming convention" approach above (it would NOT be included in the final PR).  Note also: it's not a particularly "good" plugin, just an example of the API. It demonstrates a functional approach for an IO plugin module (could be easily switched to a class-based API).  The idea here is that all IO plugins declare their functionality with two module level `lists`:
  - `READERS` : a list of 2-tuples `(checker, reader)` where `checker` is a function that accepts a filepath and returns True if it recognizes it as something it can read, and `reader` is a function that accepts args `(filepath, viewer)` and adds layers to the `viewer` instance given a `filepath`.
  - `WRITERS` : not implemented yet... this is just focusing on getting images into napari for now.

- file-reader plugin functionality is added (for now) to the viewer in `QtViewer._add_files` as follows:  every filepath that is dropped onto the viewer is passed sequentially to each reader plugin discovered by `plugins_manager.readers` (an instance of `PluginManager`), and if the reader `checker` function returns `True`, the path is handled by the corresponding `reader` function.  Not implemented yet, but potentially important, would be a way for the user to re-order the reader plugins as desired (since the first one to "recognize" the folder wins).

If you'd like to try the `napari_io_lls` plugin, here's a [small example dataset](https://www.dropbox.com/s/d9lfdk9do0bfini/small_lls_example.zip?dl=0)

### DIFF
--- a/napari/plugins.py
+++ b/napari/plugins.py
@@ -1,0 +1,127 @@
+import importlib
+import pkgutil
+import re
+
+import pkg_resources
+import requests
+
+"""
+This proposal for plugin discovery follows two of the recommendations here:
+https://packaging.python.org/guides/creating-and-discovering-plugins/
+
+1) Using naming convention:
+    plugins installed in the environment that follow a naming convention
+    (e.g. "napari_fancy_plugin"), can be discovered using `pkgutil`.
+    This also enables easy discovery on pypi
+
+2) Using package metadata:
+    plugins that declare a special key (e.g. "napari.plugins") in their
+    setup.py `entry_points` can be discovered using `pkg_resources`.
+    (this is used by pytest, for example)
+"""
+
+
+class PluginManager:
+    PREFIX = 'napari_'  # for discovery using naming convention
+    PLUGIN_ENTRY_POINT = "napari.plugins"  # for discovery using entry_points
+    IO_PLUGINS = "napari_io_"  # plugin subtype... this couple be an enum?
+    PYPI_SIMPLE_API_URL = 'https://pypi.org/simple/'
+
+    def __init__(self, discover=True):
+        self.plugins = {}
+        if discover:
+            self.discover_plugins()
+
+    @property
+    def readers(self):
+        """only plugins that declare themselves as io plugins.
+
+        the super basic API example I'm using here is that io plugins may
+        declare:
+            plugin.READERS : a list of 2-tuples (checker, reader)
+                `checker` is a function that returns True if it recognizes a
+                directory as something it can handle
+                `reader` is a function that accepts args (path, viewer) and
+                adds layers to the viewer given a path.
+            plugin.WRITERS : not implemented...
+
+        Yields
+        -------
+        tuple
+            (name, module) for all plugins
+        """
+        for name, module in self.plugins.items():
+            if name.startswith(self.IO_PLUGINS):
+                for item in getattr(module, 'READERS', []):
+                    yield item
+
+    def discover_plugins(self):
+        # using naming convention: http://bit.ly/pynaming-convention
+        # looks for all installed packages starting with self.PREFIX
+        # propsed convention for plugins is: napari_<plugin-name>
+        # coupld potentially have subclasses like napari_io_readerplugin
+        self.plugins.update(
+            {
+                name: importlib.import_module(name)
+                for finder, name, ispkg in pkgutil.iter_modules()
+                if name.startswith(self.PREFIX)
+            }
+        )
+
+        # using package metadata: http://bit.ly/pypackage-meta
+        # installed packages can register themselves for discovery by providing
+        # the `entry_points` argument in setup.py using self.PLUGIN_ENTRY_POINT
+        self.plugins.update(
+            {
+                entry.name: entry.load()
+                for entry in pkg_resources.iter_entry_points(
+                    self.PLUGIN_ENTRY_POINT
+                )
+            }
+        )
+
+    @property
+    def pypi_plugins(self):
+        # packages using naming convention: http://bit.ly/pynaming-convention
+        # can be autodiscovered on pypi using the SIMPLE API:
+        # https://www.python.org/dev/peps/pep-0503/
+        if not hasattr(self, '_pypi_plugins'):
+            response = requests.get(self.PYPI_SIMPLE_API_URL)
+            if response.status_code == 200:
+                pattern = f'<a href="/simple/(.+)">({self.PREFIX}.*)</a>'
+                self._pypi_plugins = {
+                    name: self.PYPI_SIMPLE_API_URL + url
+                    for url, name in re.findall(pattern, response.text)
+                }
+        return self._pypi_plugins
+
+    def fetch_plugin_versions(self, name):
+        """fetch versions for a plugin.
+
+        Parameters
+        ----------
+        name : str
+            name of the plugin
+
+        Returns
+        -------
+        tuple of versions availabe on pypi
+
+        Raises
+        ------
+        KeyError
+            if the plugin name is not found on pypi
+        """
+        if name not in self.pypi_plugins and not name.startswith(self.PREFIX):
+            # also search for plugin preceeded by self.PREFIX
+            name = f'{self.PREFIX}-{name}'
+        if name not in self.pypi_plugins:
+            raise KeyError(f'"{name}"" is not a recognized plugin name')
+
+        response = requests.get(self.pypi_plugins.get(name))
+        if response.status_code == 200:
+            versions = re.findall(f'>{name}-(.+).tar', response.text)
+            return tuple(set(versions))
+
+
+plugin_manager = PluginManager()
--- a/napari/plugins.py
+++ b/napari/plugins.py
@@ -1,77 +1,76 @@
 import importlib
 import pkgutil
 import re
+import warnings
+from collections import abc
+from typing import Iterator
 
 import pkg_resources
 import requests
 
+
 """
-This proposal for plugin discovery follows two of the recommendations here:
+Plugin discovery follows two of the conventions here:
 https://packaging.python.org/guides/creating-and-discovering-plugins/
 
 1) Using naming convention:
     plugins installed in the environment that follow a naming convention
-    (e.g. "napari_fancy_plugin"), can be discovered using `pkgutil`.
+    (e.g. "napari_plugin"), can be discovered using `pkgutil`.
     This also enables easy discovery on pypi
 
 2) Using package metadata:
     plugins that declare a special key (e.g. "napari.plugins") in their
     setup.py `entry_points` can be discovered using `pkg_resources`.
-    (this is used by pytest, for example)
 """
 
 
-class PluginManager:
-    PREFIX = 'napari_'  # for discovery using naming convention
+class PluginManager(abc.Mapping):
+    PLUGIN_PREFIX = 'napari_'  # for discovery using naming convention
     PLUGIN_ENTRY_POINT = "napari.plugins"  # for discovery using entry_points
-    IO_PLUGINS = "napari_io_"  # plugin subtype... this couple be an enum?
-    PYPI_SIMPLE_API_URL = 'https://pypi.org/simple/'
 
-    def __init__(self, discover=True):
-        self.plugins = {}
-        if discover:
-            self.discover_plugins()
+    def __init__(self, autodiscover: bool = True):
+        """Mapping between plugin module names and imported modules.
 
-    @property
-    def readers(self):
-        """only plugins that declare themselves as io plugins.
-
-        the super basic API example I'm using here is that io plugins may
-        declare:
-            plugin.READERS : a list of 2-tuples (checker, reader)
-                `checker` is a function that returns True if it recognizes a
-                directory as something it can handle
-                `reader` is a function that accepts args (path, viewer) and
-                adds layers to the viewer given a path.
-            plugin.WRITERS : not implemented...
-
-        Yields
-        -------
-        tuple
-            (name, module) for all plugins
+        Parameters
+        ----------
+        autodiscover : bool, optional
+            Whether to autodiscover plugin modules on init, by default True
         """
-        for name, module in self.plugins.items():
-            if name.startswith(self.IO_PLUGINS):
-                for item in getattr(module, 'READERS', []):
-                    yield item
+        self._plugins = {}
+        if autodiscover:
+            self.discover()
 
-    def discover_plugins(self):
-        # using naming convention: http://bit.ly/pynaming-convention
-        # looks for all installed packages starting with self.PREFIX
-        # propsed convention for plugins is: napari_<plugin-name>
-        # coupld potentially have subclasses like napari_io_readerplugin
-        self.plugins.update(
+    def __getitem__(self, item):
+        return self._plugins[item]
+
+    def __iter__(self):
+        return iter(self._plugins)
+
+    def __len__(self):
+        return len(self._plugins)
+
+    def discover(self):
+        """Discover napari plugins in the environment.
+
+        This uses two methods:
+            1. naming convention: http://bit.ly/pynaming-convention
+                looks for all installed packages starting with self.PREFIX
+                convention for plugins is: napari_<pluginname>
+            2. package metadata: http://bit.ly/pypackage-meta
+                installed packages can register themselves for discovery by
+                providing the `entry_points` argument in setup.py using
+                self.PLUGIN_ENTRY_POINT
+        """
+        # discover plugins using naming convention
+        self._plugins.update(
             {
                 name: importlib.import_module(name)
                 for finder, name, ispkg in pkgutil.iter_modules()
-                if name.startswith(self.PREFIX)
+                if name.startswith(self.PLUGIN_PREFIX)
             }
         )
-
-        # using package metadata: http://bit.ly/pypackage-meta
-        # installed packages can register themselves for discovery by providing
-        # the `entry_points` argument in setup.py using self.PLUGIN_ENTRY_POINT
-        self.plugins.update(
+        # discover plugins using package metadata
+        self._plugins.update(
             {
                 entry.name: entry.load()
                 for entry in pkg_resources.iter_entry_points(
@@ -80,23 +79,73 @@ class PluginManager:
             }
         )
 
+    def register(self, name: str):
+        """Manually register a module name as a plugin
+
+        Parameters
+        ----------
+        name : str
+            name of an importable module
+        """
+        self._plugins.update({name: importlib.import_module(name)})
+
     @property
-    def pypi_plugins(self):
-        # packages using naming convention: http://bit.ly/pynaming-convention
-        # can be autodiscovered on pypi using the SIMPLE API:
-        # https://www.python.org/dev/peps/pep-0503/
+    def readers(self) -> Iterator[tuple]:
+        """Generator that yields all readers declared in plugins.
+
+        Plugins may declare a top level variable `READERS` which is a list of
+        2-tuples [(checker, reader) ...] where:
+            `checker` is a function that returns True if it recognizes a
+            directory as something it can handle
+            `reader` is a function that accepts args (path, viewer) and
+            adds layers to the viewer given a path.
+
+        Yields
+        -------
+        tuple
+            (name, module) for all plugins
+        """
+        for name, module in self.items():
+            for item in getattr(module, 'READERS', []):
+                if (
+                    isinstance(item, tuple)
+                    and len(item) == 2
+                    and all([callable(i) for i in item])
+                ):
+                    yield item
+                else:
+                    warnings.warn(
+                        'Reader {item} from plugin {name} was not loaded '
+                        'because it is not a 2-tuple of callables. If you did '
+                        'not write this plugin, please alert the developer.'
+                    )
+
+    def get_pypi_plugins(self) -> dict:
+        """Search for napari plugins on pypi.
+
+        Packages using naming convention: http://bit.ly/pynaming-convention
+        can be autodiscovered on pypi using the SIMPLE API:
+        https://www.python.org/dev/peps/pep-0503/
+
+        Returns
+        -------
+        dict
+            {name: url} for all modules at pypi that start with self.PREFIX
+        """
+        PYPI_SIMPLE_API_URL = 'https://pypi.org/simple/'
+
         if not hasattr(self, '_pypi_plugins'):
-            response = requests.get(self.PYPI_SIMPLE_API_URL)
+            response = requests.get(PYPI_SIMPLE_API_URL)
             if response.status_code == 200:
                 pattern = f'<a href="/simple/(.+)">({self.PREFIX}.*)</a>'
                 self._pypi_plugins = {
-                    name: self.PYPI_SIMPLE_API_URL + url
+                    name: PYPI_SIMPLE_API_URL + url
                     for url, name in re.findall(pattern, response.text)
                 }
         return self._pypi_plugins
 
-    def fetch_plugin_versions(self, name):
-        """fetch versions for a plugin.
+    def fetch_plugin_versions(self, name: str) -> tuple:
+        """Fetch available pypi versions for a plugin name
 
         Parameters
         ----------
@@ -112,16 +161,18 @@ class PluginManager:
         KeyError
             if the plugin name is not found on pypi
         """
-        if name not in self.pypi_plugins and not name.startswith(self.PREFIX):
+        if name not in self.pypi_plugins and not name.startswith(
+            self.PPLUGIN_PREFIXREFIX
+        ):
             # also search for plugin preceeded by self.PREFIX
             name = f'{self.PREFIX}-{name}'
         if name not in self.pypi_plugins:
             raise KeyError(f'"{name}"" is not a recognized plugin name')
 
         response = requests.get(self.pypi_plugins.get(name))
-        if response.status_code == 200:
-            versions = re.findall(f'>{name}-(.+).tar', response.text)
-            return tuple(set(versions))
+        response.raise_for_status()
+        versions = re.findall(f'>{name}-(.+).tar', response.text)
+        return tuple(set(versions))
 
 
-plugin_manager = PluginManager()
+plugins = PluginManager()
--- a/napari/plugins.py
+++ b/napari/plugins.py
@@ -3,7 +3,7 @@ import pkgutil
 import re
 import warnings
 from collections import abc
-from typing import Iterator
+from typing import Iterator, Tuple, Callable, List
 
 import pkg_resources
 import requests
@@ -22,6 +22,15 @@ https://packaging.python.org/guides/creating-and-discovering-plugins/
     plugins that declare a special key (e.g. "napari.plugins") in their
     setup.py `entry_points` can be discovered using `pkg_resources`.
 """
+
+# defining some type aliases
+Checker = Callable[[str], bool]
+Reader = Callable[[str], List[Tuple]]
+# numpy doesn't offer typing yet: https://github.com/numpy/numpy/issues/7370
+# but a return type from a reader plugin would really look something like this
+# (and someday some of this will be replaced with imtypes)
+# LayerInfo = Union[Tuple[np.ndarray], Tuple[np.ndarray, dict]]
+# ReaderReturn = List[LayerInfo]
 
 
 class PluginManager(abc.Mapping):
@@ -90,7 +99,7 @@ class PluginManager(abc.Mapping):
         self._plugins.update({name: importlib.import_module(name)})
 
     @property
-    def readers(self) -> Iterator[tuple]:
+    def readers(self) -> Iterator[Tuple[Checker, Reader]]:
         """Generator that yields all readers declared in plugins.
 
         Plugins may declare a top level variable `READERS` which is a list of
@@ -102,8 +111,8 @@ class PluginManager(abc.Mapping):
 
         Yields
         -------
-        tuple
-            (name, module) for all plugins
+        Generator of tuples
+            [(checker, reader) ... ] for all plugins
         """
         for name, module in self.items():
             for item in getattr(module, 'READERS', []):
--- a/napari/_qt/qt_viewer.py
+++ b/napari/_qt/qt_viewer.py
@@ -269,7 +269,8 @@ class QtViewer(QSplitter):
             for fname in filenames:
                 for check, read in plugin_manager.readers:
                     if check(fname):
-                        read(fname, self.viewer)
+                        data, meta = read(fname)
+                        self.viewer.add_image(data, **meta)
                         return
             self.viewer.add_image(path=filenames)
             self._last_visited_dir = os.path.dirname(filenames[0])
--- a/napari/_qt/qt_viewer.py
+++ b/napari/_qt/qt_viewer.py
@@ -1,17 +1,17 @@
-import os.path
 import inspect
+import os.path
 from pathlib import Path
 
 from qtpy import QtGui
-from qtpy.QtCore import QCoreApplication, Qt, QSize
-from qtpy.QtWidgets import QWidget, QVBoxLayout, QFileDialog, QSplitter
+from qtpy.QtCore import QCoreApplication, QSize, Qt, QThreadPool
 from qtpy.QtGui import QCursor, QPixmap
-from qtpy.QtCore import QThreadPool
-from vispy.scene import SceneCanvas, PanZoomCamera, ArcballCamera
+from qtpy.QtWidgets import QFileDialog, QSplitter, QVBoxLayout, QWidget
+from vispy.scene import ArcballCamera, PanZoomCamera, SceneCanvas
 from vispy.visuals.transforms import ChainTransform
 
 from .qt_dims import QtDims
 from .qt_layerlist import QtLayerList
+from ..plugins import plugins
 from ..resources import resources_dir
 from ..utils.theme import template
 from ..utils.misc import str_to_rgb
@@ -30,7 +30,15 @@ from .qt_console import QtConsole
 from .qt_viewer_dock_widget import QtViewerDockWidget
 from .qt_about_keybindings import QtAboutKeybindings
 from .._vispy import create_vispy_visual
-from ..plugins import plugin_manager
+
+
+def add_plugin_layer(viewer, data, meta=None, dims=None):
+    meta = meta or {}
+    layer_type = meta.pop('layer_type', 'image').lower()
+    # layer2addmethod from #723
+    layer = layer2addmethod[layer_type](viewer, data, **meta)
+    # could use dims argument here to modify dims properties
+    return layer
 
 
 class QtViewer(QSplitter):
@@ -267,11 +275,13 @@ class QtViewer(QSplitter):
             # change the order of reader plugins (because the first plugin
             # to claim a path wins)
             for fname in filenames:
-                for check, read in plugin_manager.readers:
+                for check, read in plugins.readers:
                     if check(fname):
-                        data, meta = read(fname)
-                        self.viewer.add_image(data, **meta)
-                        return
+                        new_layers = [
+                            add_plugin_layer(self.viewer, *layer_info)
+                            for layer_info in read(fname)
+                        ]
+                        return new_layers
             self.viewer.add_image(path=filenames)
             self._last_visited_dir = os.path.dirname(filenames[0])
 
--- a/napari/_qt/qt_viewer.py
+++ b/napari/_qt/qt_viewer.py
@@ -30,6 +30,7 @@ from .qt_console import QtConsole
 from .qt_viewer_dock_widget import QtViewerDockWidget
 from .qt_about_keybindings import QtAboutKeybindings
 from .._vispy import create_vispy_visual
+from ..plugins import plugin_manager
 
 
 class QtViewer(QSplitter):
@@ -260,6 +261,16 @@ class QtViewer(QSplitter):
             List of filenames to be opened
         """
         if len(filenames) > 0:
+            # FIXME: haven't decided how to handle lists of filenames yet in a
+            # way that also falls back gracefully to add_image(path=filenames)
+            # ALSO: need a GUI option (with persistence) to let the users
+            # change the order of reader plugins (because the first plugin
+            # to claim a path wins)
+            for fname in filenames:
+                for check, read in plugin_manager.readers:
+                    if check(fname):
+                        read(fname, self.viewer)
+                        return
             self.viewer.add_image(path=filenames)
             self._last_visited_dir = os.path.dirname(filenames[0])
 
--- a/napari_io_lls/__init__.py
+++ b/napari_io_lls/__init__.py
@@ -29,16 +29,21 @@ def is_lls_folder(path):
 # two options here... the reader can either accept the viewer instance, and
 # call viewer.add_image (or even other layers) itself,
 # or it can return args and kwargs that would be passed to viewer.add_image
-# the first case is implemented here
-def read_lls_folder(directory, viewer):
-    """Take a directory and load images in the viewer
+# the second case is implemented here
+def read_lls_folder(directory):
+    """Take a directory and return image data and metadata
     
     Parameters
     ----------
     directory : str
         Path to directory
-    viewer : napari.Viewer
-        viewer instance
+    
+    Returns
+    -------
+    data : array
+        Image array data
+    meta : dict
+        Image metadata dictionary.
     """
     i = 0
     channels = []
@@ -73,8 +78,13 @@ def read_lls_folder(directory, viewer):
     stack = da.stack(channels)
     scale = [1] * stack.ndim
     scale[-3] = dz / dx
-    viewer.add_image(stack, channel_axis=0, rgb=False, scale=scale, name=waves)
-    viewer.dims.ndisplay = 3
+    meta = {
+        'channel_axis': 0,
+        'rgb': False,
+        'scale': scale,
+        'name': waves,
+    }
+    return stack, meta
 
 
 """
--- a/napari_io_lls/__init__.py
+++ b/napari_io_lls/__init__.py
@@ -78,12 +78,7 @@ def read_lls_folder(directory):
     stack = da.stack(channels)
     scale = [1] * stack.ndim
     scale[-3] = dz / dx
-    meta = {
-        'channel_axis': 0,
-        'rgb': False,
-        'scale': scale,
-        'name': waves,
-    }
+    meta = {'channel_axis': 0, 'rgb': False, 'scale': scale, 'name': waves}
     return stack, meta
 
 
--- a/napari_io_lls/__init__.py
+++ b/napari_io_lls/__init__.py
@@ -1,0 +1,94 @@
+import os
+import dask.array as da
+from dask import delayed
+from skimage.io import imread
+from skimage.external import tifffile as tf
+from glob import glob
+import re
+
+delayed_read = delayed(imread)
+
+
+def is_lls_folder(path):
+    """this should be a fast function that doesn't actually read any data.
+    It should simply return True if it is capable of handling the path."""
+    if not os.path.isdir(path):
+        return False
+    tiffs = 0
+    for fname in os.listdir(path):
+        # ignore all files but tifs
+        if not fname.endswith('.tif'):
+            continue
+        if not 'msecAbs' in fname and 'stack' in fname:
+            return False
+        else:
+            tiffs += 1
+    return tiffs > 0
+
+
+# two options here... the reader can either accept the viewer instance, and
+# call viewer.add_image (or even other layers) itself,
+# or it can return args and kwargs that would be passed to viewer.add_image
+# the first case is implemented here
+def read_lls_folder(directory, viewer):
+    """Take a directory and load images in the viewer
+    
+    Parameters
+    ----------
+    directory : str
+        Path to directory
+    viewer : napari.Viewer
+        viewer instance
+    """
+    i = 0
+    channels = []
+    waves = []
+    shape = None
+    dtype = None
+    while True:
+        ch_files = glob(os.path.join(directory, f'*ch{i}*.tif'))
+        if not ch_files:
+            break
+        if not (shape and dtype):
+            tif = tf.TiffFile(ch_files[0])
+            shape = tif.series[0].shape
+            dtype = tif.series[0].dtype
+            dx = tif.series[0][0].x_resolution or (1, 1)
+            dx = dx[1] / dx[0]
+            try:
+                dz = tif.series[0][0].imagej_tags.get('spacing', 1)
+            except AttributeError:
+                dz = 1
+        channels.append(
+            da.stack(
+                [
+                    da.from_delayed(delayed_read(fn), shape, dtype)
+                    for fn in ch_files
+                ]
+            )
+        )
+        waves.append(re.search('(\d+)nm_', ch_files[0]).groups()[0])
+        i += 1
+
+    stack = da.stack(channels)
+    scale = [1] * stack.ndim
+    scale[-3] = dz / dx
+    viewer.add_image(stack, channel_axis=0, rgb=False, scale=scale, name=waves)
+    viewer.dims.ndisplay = 3
+
+
+"""
+For a class-free, functional API, all IO plugins declare their functionality
+with two module level variables:
+    READERS : a list of 2-tuples (checker, reader)
+        `checker` is a function that returns True if it recognizes a
+            directory as something it can handle
+        `reader` is a function that accepts args (path, viewer) and
+            adds layers to the viewer given a path.
+    WRITERS : a list of writing functions
+        not implemented
+
+Alternatively, we could define a class-based API.
+"""
+READERS = [(is_lls_folder, read_lls_folder)]
+WRITERS = []


napari/plugins.py
```diff
@@ -0,0 +1,127 @@
+import importlib
+import pkgutil
+import re
+
+import pkg_resources
+import requests
+
+"""
+This proposal for plugin discovery follows two of the recommendations here:
+https://packaging.python.org/guides/creating-and-discovering-plugins/
```
sofroniewn, line 14: Conforming to the recommendations from https://packaging.python.org/guides/creating-and-discovering-plugins/ certainly makes sense to me

sofroniewn:
Thanks for getting us going here @tlambert03 - I think it's so much easier to have discussions around concrete examples like this, and agree IO plugins are a great place to start - and maybe we should think about #616 in this context a bit too.

I'll put down a couple high level thoughts right away, but my thinking around all of this is definitely early and very open to change based on suggestions from the group.

### main concepts
I think philosophically I like the idea of providing infrastructure for plugin developers who are committed to our ecosystem (say developing specifically for napari) to use if they want, but then also providing reasonable defaults / guesses for people who aren't (or who are developing for multiple ecosystems) to still get their stuff in here too (i.e. around the concept of reducing developer burden)

So right off the bat, I like the idea of things like `io_plugins`, `functional_plugins`, `interactive_plugins`, and maybe others.

In your current code you mention:
> two options here... the reader can either accept the viewer instance, and
call viewer.add_image (or even other layers) itself,
or it can return args and kwargs that would be passed to viewer.add_image

I have a pretty strong preference for something like the later as it reduces the napari-specificness of the plugin, enabling reuse by many tools, and it might make direct integration with other loading libraries much easier - who might not want to go all in on the napari ecosystem, but don't mind conforming to our keyword arguments.

By default we could pass these params to `add_image`, but it would be nice if developers could also import from our yet to exist `image_types` repo the types of all our layers `Image`, `Vectors`, `Shapes` etc. and then we could parse to look for those can connected up to `add_*` appropriately.

A disadvantage of this approach is that it might be more error prone, if APIs diverge etc it could cause more breaking changes.

Similarly I think writers could be typed so we new which type of layer they were getting, and then you could pass a path and then the dict of layer properties from the `_get_state` method defined in #686 rather than accepting the whole layer.

This makes it clear what the API IO plugins are expected to conform to, both reading / writing, which we could also maintain tests for.

### reader and checker
Right now you propose READERS will contain a `list of 2-tuples (checker, reader)`. While I like this, I worry forcing developers to always provide checkers might be too much of a burden. Is there a clean way to make them optional? Also what is our strategy around using the `checker`. Do we run the checker on every path before running the reader to check that we can then run the reader? Why not just try the reader and fail gracefully?

### conclusions
All in all I love this, and would love to see us get this in in a timely fashion, so thanks again!!! :-)

tlambert03:
thanks for your thoughts!  couple responses to direct questions below:

> philosophically I like the idea of providing infrastructure for plugin developers who are committed to our ecosystem (say developing specifically for napari) to use if they want, but then also providing reasonable defaults / guesses for people who aren't (or who are developing for multiple ecosystems) to still get their stuff in here too

not entirely sure I'm following... we would certainly continue to offer the `magic_imread` function as a reasonable default.  This proposal just adds a way to extend that with additional readers that are "in line" before the default behavior.  As usual, it's up to the end-user to decide which plugins they have installed in their environment, and no `io_plugins` would be required to open basic images.  But just like pytest, we could develop additional "built-in" plugins for the most common use cases (if desired).

>> (@tlambert03): the reader can either accept the viewer instance, and call `viewer.add_image` (or even other layers), or it can return args and kwargs that would be passed to `viewer.add_image`

> (@sofroniewn): I have a pretty strong preference for something like the later as it reduces the napari-specificness of the plugin, enabling reuse by many tools, and it might make direct integration with other loading libraries much easier - who might not want to go all in on the napari ecosystem, but don't mind conforming to our keyword arguments.

Can you be more specific?  This may again be lack of imagination, but it's not entirely clear to me who we would be targeting with a napari io plugin infrastructure _besides_ people writing functions to read data into `napari`.  That is, i don't think there is nearly enough agreement in APIs (keyword arguments) across python packages that anyone would be writing a generic plugin to "take a filepath and load it into an arbitrary viewer".  Thinking also about plugins for other python packages (like pytest), I don't know of many examples where those plugins were written for multiple target programs.

i fear a little bit of "premature optimization" if we make something that is, as its core, a framework for napari plugins, but which is open enough for developers who "don't want to go all in on the napari ecosystem".

> By default we could pass these params to `add_image`, but it would be nice if developers could also import from our yet to exist image_types repo the types of all our layers Image, Vectors, Shapes etc. and then we could parse to look for those can connected up to add_* appropriately.

so what are you suggesting the plugin function would return?  layer objects?  If so, that requires a plugin to actually import from `napari.layers`, whereas handing them the viewer does not require any imports from napari.  And rather than calling `viewer.add_image` in the plugin, they have to construct an `Image` layer and then we just end up calling `viewer.add_image` internally anyway... so the benefit isn't totally clear.  Additionally, by giving them the viewer, plugin authors may decide to do more than simply add a layer.  For instance, it's conceivable that a data type might nearly always want to be opened in volumetric view, which would be possible if the plugin was handed the viewer instance, but not if they had to return layers.
I also agree that it's more prone to error or stale APIs.

> This makes it clear what the API IO plugins are expected to conform to, both reading / writing, which we could also maintain tests for.

even if we don't expect explicit layers to be returned, I don't think it is necessarily _unclear_ what API plugin developers are expected to conform to if we say "write a function that takes a filepath and the viewer, and adds data from that path to the viewer".  We can't test plugins for developers (but we can write additional internal code that makes sure io plugins do what we expect them to, and that it is done in a reasonable amount of time).

> Right now you propose `READERS` will contain a list of 2-tuples `(checker, reader)`. While I like this, I worry forcing developers to always provide checkers might be too much of a burden. Is there a clean way to make them optional? Why not just try the reader and fail gracefully?

The rationale there was that an io plugin should "abstain" _very_ fast if it knows it doesn't recognize the data.  If there are 30+ io plugins, it should still take almost no time for all of them to report whether they recognize the data (I can also imagine writing some internal code that would wrap that checker in a timeout that just warns the user about a misbehaving plugin and moves on to the next one if it didn't give a boolean response within some designated SHORT amount of time).  Reader functions however might need to take a little bit of time, so if we just try each reader sequentially and fail gracefully, then if the "8th" reader in line is the correct one, it still has to wait for all of the others to fail before getting to read the data.  By asking for a boolean-returning function (that we can wrap in a timeout) we signal to developers that that process is important.  An alternative is to request that they return `False` early in their reader function if they know the data is incompatible, but doesn't seem to be any less of a burden on developers, while being more error prone.  Note, the checker function can also be as simple as `return filepath.endswith('.nd2')`

an alternative to the `(checker, reader)` list API might be to ask plugin developers to subclass some abstract `napari.ReaderPlugin` class.  And then rather than a module level READERS list, we would search the module for subclasses of our special classes. There's lots we could do there, (for instance, add [abstractmethods](https://docs.python.org/3/library/abc.html#abc.abstractmethod) to indicate which methods plugin developers have to provide).  honestly I kinda prefer that approach, but I started with the function-based API to make @jni happy :)

> Also what is our strategy around using the checker. Do we run the checker on every path before running the reader to check that we can then run the reader? 

the strategy for the checker is to run the requested filepath on each plugin `checker` in sequence until one returns true, at which point the reader function is triggered, like here:
https://github.com/napari/napari/blob/9ee3f59a33653a9f256f018236d67273b03ebcc0/napari/_qt/qt_viewer.py#L270-L273

sofroniewn:
> so what are you suggesting the plugin function would return?

I was suggesting returning a dictionary containing basic types (including numpy and dask), and possibly use some additional typing from a more neutral light weight library of image types to get our layer type information - something which we'd been prototyping here https://github.com/napari/image-types - but note that wouldn't require importing `napari.layers` or instantiating any of our layer objects, so they wouldn't need `napari` as a dependency.

Another thing I might consider returning instead is either just a singleton object - which will get passed as the `data` object (so could be an array or list of arrays for pyramids, etc), or a tuple `(data, meta)` where `data` was the same as just described and `meta` was the dictionary of other keyword arguments (so things like `contrast_limits` etc. that will get unpacked to the `add_*` method). 

This is already potentially quite similar to many existing fileIO libraries, we'd really just be asking people to try and standardize around the `meta` object, which I think would be valuable for the community at large.

> Can you be more specific? This may again be lack of imagination, but it's not entirely clear to me who we would be targeting with a napari io plugin infrastructure besides people writing functions to read data into napari. That is, i don't think there is nearly enough agreement in APIs (keyword arguments) across python packages that anyone would be writing a generic plugin to "take a filepath and load it into an arbitrary viewer".

I'd like to see things like [python-bioformats](https://github.com/CellProfiler/python-bioformats) and the [PIL file open](https://pillow.readthedocs.io/en/stable/reference/open_files.html) become easy to use with napari. One option is to maintain things like `napari-python-bioformats` or `napari-PIL` which create an explicit bridge from their fileIO to the napari viewer. The other is to try and push towards standard APIs that the napari viewer and then other tools could accept, for example the `(data, meta)` approach mentioned above. If successful that could have large positive impact in the community. Curios what @jni thinks.

> i fear a little bit of "premature optimization" if we make something that is, as its core, a framework for napari plugins, but which is open enough for developers who "don't want to go all in on the napari ecosystem".

On the one hand I agree - but on the other, I worry that if we try and force people into napari they will be resistant and we'll lose them - whereas if we just say, "return a dictionary with these optional keyword args" then all of a sudden their tool becomes something the can be used by many tools.

> The rationale there was that an io plugin should "abstain" very fast if it knows it doesn't recognize the data. If there are 30+ io plugins, it should still take almost no time for all of them to report whether they recognize the data 

This could be a very good idea. I feel like bioformats has taken this approach, but I'm not sure, and I'm curious if they have any advice for us on that matter. I like it in some ways, but I'm also wary of many checks, and the idea that two plugins might pass a check but then return data in different ways, and forcing people to be explicit about what plugin loader they want might be good, but I can also see people might like that magic and it might be needed if we have many IO plugins.

Also curious what @joshmoore from the bioformats team makes of all this - it's still super early in our thought process though so be gentle ;-)

jni:
Hi everyone, apologies that I haven't fully read or digested everything mentioned here, but I would like to propose that as a *first* pass, I would like IO plugins to be similar to functional plugins, meaning something that takes in a PathLike and returns an {ArrayLike, Image, Coords, Paths, WhateversThatHasARegisteredLayer}. If we provide a good type hierarchy for this, I would hope that many in the community would adopt it, because it would be useful entirely outside of napari.

In contrast, things that take in a napari Viewer as input, or that subclass from a napari Reader, require immediate buy-in into the napari ecosystem. I still want to maintain "napari headless is just Python", which means that plugins should look like Python modules that napari uses, rather than plugins immediately being hooked into napari.

In terms of "Reader" hierarchies (thanks for the shout out, @tlambert03, appreciated! 😅), I would like us to work with the community here, meaning imageio, PIMS, or something else, but *not* rolling our own Reader thing. There's lots of Reader things out there, and I'd like us to work with and improve them if needed, rather than develop our own.

Sorry, no time for more specific thoughts right now, but I hope that provides some of the ideas of how I would like to think about IO plugins. Apologies if some of them have already been debunked above and I missed it in my quick skimming. More tomorrow!

tlambert03:
In principle I love the idea of a universal API that doesn't require buy-in to the napari ecosystem, but I guess I need some more concrete code examples to see what you guys have in mind.  I started thinking about this because I had at least 3 conversations in the last couple weeks with people who said "napari looks great but I haven't figured out how to use it yet with my data" ... and their data was nothing more than a folder of tiffs in a particular arrangement.  so the driving motivation here was user experience: someone should be able to (e.g.) `pip install napari napari_io_myplugin`, and then drag and drop their experiment folder onto the window.

It's possible that what I'm proposing here is a much smaller, lower-level, napari-focused "slot" for a more global io layer like you guys are envisioning, but I remain a bit confused as to how you would achieve certain outcomes _in napari_ and would need to see some code.


> I would like IO plugins to be similar to functional plugins, meaning something that takes in a PathLike and returns an {ArrayLike, Image, Coords, Paths, WhateversThatHasARegisteredLayer}. 

how would you then signal that the `ArrayLike` was actually multichannel and should be shown with three different channels with specific names, scales, colors, and possibly displayed immediately in volume view?
 
> In contrast, things that take in a napari Viewer as input, or that subclass from a napari Reader, require immediate buy-in into the napari ecosystem. I still want to maintain "napari headless is just Python", which means that plugins should look like Python modules that napari uses, rather than plugins immediately being hooked into napari.

Sounds good but I guess I need a minimum concrete example.  A main part of this proposal is not so much how a reader looks, but plugin **discovery** using 2 [standard python plugin discover methods](https://packaging.python.org/guides/creating-and-discovering-plugins/), which require the plugin developer to either name their package `napari_x` or to add a napari-specific `entry_point` to setup.py.  So I'm still unclear how we can both A) auto-detect a module that has been pip installed that declares itself to have additional napari functionality while also B) having a plugin just look like a generic module not being immediately being hooked into napari... (see also the next point)

> In terms of "Reader" hierarchies (thanks for the shout out, @tlambert03, appreciated! 😅), I would like us to work with the community here, meaning imageio, PIMS, or something else, but _not_ rolling our own Reader thing.

I want to be clear that I'm not proposing a whole new hierarchy for file io here, or what a reader should look like... and I'm definitely _not_ suggesting we write new file readers for arbitrary file types.  This is a proposal for how we get packages like PIMS or bioformats (or some developer's domain-specific code) to help users who can't open their stuff in napari right now.  for example, adding add PIMS or bioformats support to napari with this proposal would look something like this:

```python
# napari_io_pims.py

import pims

def is_pims(filepath):
    return filepath.endsith(pims_compatible_extensions)

def read(filepath, viewer):
    images = pims.open(filepath)
    viewer.add_image(images)  # or whatever we decide the return type should be

READERS = [(is_pims, read)]
```

> There's lots of Reader things out there, and I'd like us to work with and improve them if needed, rather than develop our own.

I guess I still need examples of how napari becomes **aware** of those things.  we can certainly pick projects like PIMS to team up with such that if someone does `pip install napari pims`, new functionality becomes available within napari, (with, eg. `try: import pims` somewhere inside napari)... but that's really not a plugin.  That's "napari supports pims"... and it doesn't prescribe a way that some random code-savvy user can help some other non-code-savvy user look at their specific, perhaps unusual, data format in napari (just by dragging and dropping).

I do think there's a very real and present need for user-assistance in opening data in napari (based on recent conversations), so if either of you have some time in the near future, I'd love to see some minimal code examples that demonstrate a bit more concretely what you're envisioning.

sofroniewn:
> Sounds good but I guess I need a minimum concrete example.

I can work on this asap, likely trying to make a PR to this PR with a reader that returns `(data, meta)` instead of taking in the viewer.

> "napari looks great but I haven't figured out how to use it yet with my data" ... and their data was nothing more than a folder of tiffs in a particular arrangement. 

I'm very sympathetic to this, and very sympathetic to guiding our development and prioritization to meet real user needs. I also agree there is a distinction between foundational libraries that say read tiffs and a convenience io-plugin that might use a foundational library and then parse the filename into some metadata that only makes sense for people doing exactly one thing and may have no adoption outside of people doing exactly that thing.

> A main part of this proposal is not so much how a reader looks, but plugin discovery using 2 standard python plugin discover methods, which require the plugin developer to either name their package napari_x or to add a napari-specific entry_point to setup.py.

I like this part of the proposal, and agree it does make sense to start doing things like autodetection for people that have explicitly chosen to be part of the ecosystem. To me it was separate from whether the plugin took in the viewer or not. Curious what @jni and @kne42 make of this part though. I think there should also be ways to explicitly register plugins that aren't named in this way, but that could come later.

Hopefully I can make a PR to this PR soon, I'm traveling for work right now - just about to land in London!!

tlambert03:
> I think there should also be ways to explicitly register plugins that aren't named in this way, but that could come later.

would be easy to add a `register_plugin` method to `PluginManager` that takes the name of a module and adds it to the plugins dict.  and then provide a convenience function (`napari.register_plugin()`)

tlambert03:
> I can work on this asap, likely trying to make a PR to this PR with a reader that returns `(data, meta)` instead of taking in the viewer.

just to list some desirable goals for this (as I see it):
- should be able to add (an arbitrary number of) different layer types to the viewer
- should be able to set name, scale, color, other props, etc... on each layer
- would be nice to control dims as well (e.g. label dims, set `ndisplay = 3`)

tlambert03:
> > The rationale there was that an io plugin should "abstain" very fast if it knows it doesn't recognize the data. If there are 30+ io plugins, it should still take almost no time for all of them to report whether they recognize the data
> 
> This could be a very good idea. I feel like bioformats has taken this approach, but I'm not sure, and I'm curious if they have any advice for us on that matter. I like it in some ways, but I'm also wary of many checks, and the idea that two plugins might pass a check but then return data in different ways, and forcing people to be explicit about what plugin loader they want might be good, but I can also see people might like that magic and it might be needed if we have many IO plugins.

I'd also be curious to hear what the bioformats team has to say, but to my understanding, this is more or less exactly what bioformats and python-bioformats does.  All [`loci.formats.FormatReader`](https://downloads.openmicroscopy.org/bio-formats/5.5.1/api/loci/formats/FormatReader.html) subclasses have a  method [`isThisType`](https://downloads.openmicroscopy.org/bio-formats/5.5.1/api/loci/formats/FormatReader.html#isThisType-java.lang.String-) that accepts a filename or bytes and returns a boolean (whether it matches the reader format).   In `python-bioformats` you can see an example [here](https://github.com/CellProfiler/python-bioformats/blob/22f5bcbd798779fb0425098c305fcf472685c008/bioformats/formatreader.py#L630-L668) of iterating through the all known reader classes, calling `isThisType()` until one of them returns `True`.

... and an example of such a checker function for a `napari_io_bioformats` plugin could be:
```python
from python-bioformats import READABLE_FORMATS

def is_readable(filepath):
    for ext in READABLE_FORMATS:
        if filepath.endswith(ext):
            return True
    return False
```

sofroniewn:
Pulling the convo from https://github.com/tlambert03/napari/pull/4 back in here

see - https://github.com/tlambert03/napari/pull/4#issue-364553987 and https://github.com/tlambert03/napari/pull/4#issuecomment-576027526 for full context

> a special layer_type keyword would be fine with me... and by keeping the actual layer creation logic within napari, I do think it will be easier to change the IO API over time with backwards compatibility. so I'm good with that if everyone else is (but I wonder if that gets closer to the napari-specific ecosystem that @jni was trying to avoid).

We could go with `type` instead of `layer_type` which feels very napari centric. I think supporting this will be really nice for us in general, even napari internal IO usage - like I was exploring in #511 

> For an API to be consumed by external developers like this, I'm personally in favor of keeping it strict, so I think readers should always return an iterable of [(data, meta, ...), (data, meta, ...), ...]. Pyramid data would be [([im0, im1, im2], meta,...), ...]. And on the napari side we would do something like:
> 
>```python
>    def add_plugin_layer(viewer, data, meta=None, dims=None):
>    meta = meta or {}
>    layer_type = meta.pop('layer_type', 'image').lower()
>    # layer2addmethod from #723
>    layer = layer2addmethod[layer_type](viewer, data, **meta)
>    # could use dims argument here to modify dims properties
>    return layer
>
># and then to actually consume the data returned from the plugin
>new layers = [add_plugin_layer(viewer, *layer_info) for layer_info in plugin_data]
>```
> with the asterisk on layer_info, meta becomes optional and we can also add additional optional things after that (like a dims-adjusting feature as shown here). But if we overload the plugin API by allowing them to return an array sometimes, a list other times, etc... then this all gets more complicated. (it also gets harder to check and emit a warning when a plugin is breaking the API).

I like the idea of keeping the API strict. I'm not 100% sure yet about the concept of multiple dictionaries - `meta`, `dims` etc. I do see how having more though would be useful to do things like change the viewer dims / camera etc. An alternative is to keep it forced at `(data, meta)` and then inside meta have more "reserved" keywords that napari interprets in special ways - for example the `type` or `layer_type`, and now maybe `dims` (or `viewer` etc). This could give us the flexibility to set those extra properties like `ndisplay` and `axis_labels`, but also make it simpler for plugin devs. (Note if non of the "reserved" keywords are provided we always degrade nicely - for `dims` it would mean not changing any of those properties. I am still not sure that you want your file IO plugin to take control of the dims model / camera position - maybe only the `axis_labels` seem critical and as I was saying above we might want to figure out a way to get them added within the context of `add_*` methods. I.e. maybe `axis_labels` should be a keyword arg to all our `add_*` methods - this would also make it clear where to implement #14.

Though I am sympathetic to

> your call, but I think it will probably come up relatively quickly ("how do I"... ) and users will figure out that they need to also do "x, y, z" whenever they use some plugin to achieve the desired behavior in napari. It's nice if we can control that "x, y, z" slightly and I think if we outline an API for plugins (like the optional dims argument above) that would enable a large subset of functionality like axis labeling and volumetric rendering which would satisfy the majority of users without them having to create their own workarounds.

Being prepared for the "reserved" keyword approach within the `meta` dict could allow us to add this stuff in later if we want.

tlambert03:
> I'm not 100% sure yet about the concept of multiple dictionaries - meta, dims etc.  An alternative is to keep it forced at `(data, meta)` and then inside meta have more "reserved" keywords that napari interprets in special ways - for example the `type` or `layer_type`, and now maybe `dims` (or `viewer` etc).

I'm not too attached to that third `dims` argument in the tuple (though I do think it would be nice).  The nice thing about the `add_plugin_layer(viewer, *layer_info)` strategy is that it wouldn't matter if plugin developers knew about `dims` or not: so `[(data, meta, dims), (data, meta), (data,)]` would _all_ be valid.  I'm less into the idea of added reserved keywords in the `meta` dict, because each one of them would need to be explicitly `pop`ed out of the `meta` dict before handing `**meta` to the layer creation (either that or we add a catchall `**kwargs` to all layer creation functions).

jni:
> I think there should also be ways to explicitly register plugins that aren't named in this way, but that could come later.

:+1: For me, actually, this should come *first*, and the discovery part can come later. I'm ok with, for now, telling users that you need to `pip install napari imageio` and then go to `Plugins -> register` and type or select imageio from a list.

The discovery part is just selecting from a big list and then calling the function above. ie, the function above is a necessary part of the process anyway, so we might as well write it, and then worry about discovery. Which brings me to:

> Sounds good but I guess I need a minimum concrete example. A main part of this proposal is not so much how a reader looks, but plugin discovery using 2 standard python plugin discover methods, which require the plugin developer to either name their package napari_x or to add a napari-specific entry_point to setup.py. So I'm still unclear how we can both A) auto-detect a module that has been pip installed that declares itself to have additional napari functionality while also B) having a plugin just look like a generic module not being immediately being hooked into napari... (see also the next point)

The idea would be to crawl installed modules for functions annotated with the type "PathLike -> (ImageLike, MetadataLike)" and use them. So I would write (abbreviated for clarity, all names in flux):

```python

from imtypes import Path, Image, Metadata, Coords, Paths

def imread(path: Path) -> Image, Metadata:
    ...

def coords_read(csvfile: Path) -> Coords:
    ...

def skeleton_read(swcfile: Path) -> Paths
    ...
```

publish the above as a package, and if napari comes across this in the install packages (inspecting the annotations), it'll register all of those functions as readers for those files producing the appropriate layers. This doesn't solve the fast checking problem, but I think that is a minor hurdle.

But, I am certainly getting the sense of urgency here and I understand that getting the typing info to propagate across the ecosystem is going to be a massive effort, possibly a year or two. So ok, we need something to work now. I can probably be convinced to support the named modules approach as long as we don't paint ourselves into a corner with the API. In particular,

> how would you then signal that the ArrayLike was actually multichannel and should be shown with three different channels with specific names, scales, colors, and possibly displayed immediately in volume view?

This to me seems like a secondary goal. Almost all of those things are fixable after reading, so let's not worry too much about having the plugins do *everything* immediately. Let's instead work on getting plugins that `add_image` with default things working, and let's add a "split channels" function to the menubar, and then worry about all of the metadata, how we format it without tying it to the API, etc, later. I *really* don't want people writing "napari-format" metadata when this is a problem that so many people are trying to tackle.

As just one example I want to point to this imageio issue:

https://github.com/imageio/imageio/issues/362

where we are discussing minimal metadata requirements, what the names of things should be, etc.

I'm certainly happy to contemplate a PR that allows people to define readers that provide all the metadata we need to "properly" present their data, but don't you think that this could come in *after* getting a basic reader in?

Sorry, I have again run out of time but I'll look at the code tomorrow, I promise! :grimacing:

tlambert03:
I can definitely see how nice a future would be with a community-agreed `imtypes` schema that requires a `(Path) -> Imtype:` type annotation for autodiscovery.  though I do agree it seems years away.  And would napari then have to import and inspect every submodule in every package in the environment?  Or does there need to be some agreed-upon convention for where such functions live (i.e. only in the top level namespace), or would users have to manually register the plugin for it to be crawled?

Currently, without any sort of settings persistence in napari, the `Plugins -> register` idea would require the user would to register the plugin every time they opened napari, which seems like an unnecessarily clunky way to workaround what is already a [community standard](https://packaging.python.org/guides/creating-and-discovering-plugins/) plugin detection scheme.  I guess I don't see _why_ the discovery part needs to come later if it's as simple and commonly used as it is.  (I also didn't know what you meant by "the function above" here... so perhaps we're having a slight misunderstanding).

anyway, I'm not attached to any of this, so I'd very much look forward to alternate proposals that lets us (potential plugin developers) help people get their data into napari.

sofroniewn:
> >how would you then signal that the ArrayLike was actually multichannel and should be shown with three different channels with specific names, scales, colors, and possibly displayed immediately in volume view?
>
>
> This to me seems like a secondary goal. Almost all of those things are fixable after reading, so let's not worry too much about having the plugins do everything immediately. Let's instead work on getting plugins that add_image with default things working, and let's add a "split channels" function to the menubar, and then worry about all of the metadata, how we format it without tying it to the API, etc, later. I really don't want people writing "napari-format" metadata when this is a problem that so many people are trying to tackle.

I'm not sure about this @jni - to me this sounds like the top line goal that people are asking for and something that would dramatically improve the usability of napari and its accessibility to non-coders (who have coder friends that have written them readers ;-)

I think we're not trying to write one generic IO library with standard metadata for all of imaging - which I think we agree would be a mistake, but I see it as more that we're trying to provide a way for developers to write code that could help non-coders look at their data more easily in napari which necessitates the usage of keywords that make sense in napari. Hopefully we've chosen those keywords to be fairly generic and make sense outside of napari, but I think we need to support them to allow people to set the things they want set when loading data, and expecting people to do all of that by hand whenever they load in new data sounds like a terrible user experience that people are already complaining about. If people provide different keyword args or no keyword args we shouldn't fail (I realize the proposed scheme may not have the checks to take that into account right now), and I don't want to do things that try and lock people into a "napari ecosystem" by say writing code that only works with napari imported, but the meta dict or multiple dicts approach with keyword args seems to me to strike the right balance between being a neutral object - a dictionary - but giving developers enough control to get the behaviour in napari they want.

> I'm certainly happy to contemplate a PR that allows people to define readers that provide all the metadata we need to "properly" present their data, but don't you think that this could come in after getting a basic reader in?

I appreciate that - and as noted above I'd argue that we should try and do this now at the same time as that's where I see the real need. I am curious what alternatives you can see to the dict approach described above or if you have ideas on how to improve the API there (particularly if one reader call is allowed to add multiple layers), as I do want us to do this in the least burdensome way possible.

I'll also note that I see this as quite different than the convos in https://github.com/imageio/imageio/issues/362 (which I think are still valuable) as even then we'd need a way to translate those keyword args into keyword args that make sense in the napari model, and there maybe would still be napari parameters not represented in that minimal set that users of napari would still wanted set on load, so we'd still have to let people provide their own keyword args for usage in napari anyway. 

Our approach should also be consistent across all layer types, and I'm not sure how else we'd do this for `Shapes` or `Surface` where aligning metadata in a cross-community way seems impossible or to not really mean much.

Maybe my summary of the above is that I don't think we're inventing or forcing a new "napari-format" here, but we're trying to allow developers to interact with the already defined napari `add_*` API in a convenient way, which necessitates letting them use napari keyword args when they want to.

Using the standard python recommendations around plugin discovery also makes sense to me, particularly as it will be a long time before the typing is ready, and for people that are less comfortable with types (which maybe many of our potential plugin developers) it is pretty simple, but agreed we should think carefully before deciding what exactly to support.

jni:
> I appreciate that - and as noted above I'd argue that we should try and do this now at the same time as that's where I see the real need.

My point above is that:

- directory of e.g. tiffs in some specific arrangement -> napari image layer is a clear goal with few contentious issues at this point, so we can get that in ~tomorrow.
- keyword arguments and layer type stuff poses all kinds of, at a minimum, API stability issues that I think warrant further consideration and discussion. To pick @tlambert03's example above, I've never really loved the `ndisplay` keyword, not that I have a wonderful alternative in my pocket, but I am apprehensive of seeing a proliferation of plugins using that keyword, possibly holding us back from a better alternative in the future.

Add to that the fact that smaller PRs are faster and easier to review, and to me it seems like a slam dunk to get the first stage merged now-ish and continue discussing in a subsequent PR.

> Currently, without any sort of settings persistence in napari, the Plugins -> register idea would require the user would to register the plugin every time they opened napari, which seems like an unnecessarily clunky way to workaround

Oh yeah, indeed that whole framework requires caching/persistence, as it is much too slow to work.

> what is already a community standard plugin detection scheme.

I really appreciate this, and I love trying to adhere to existing standards, but I caution that Python packaging is in a massive state of flux and I would argue that some of those standards are kind of brittle and not ideal. `setup.py` itself is slowly falling out of favour compared to new, declarative standards like pyproject.toml, while the naming convention is somewhat prone to namespace clashes.

Overall, I would say, let's go ahead with `napari-*` or, perhaps preferably, `napari-plugin-*`, and leave the function annotation approach for the future.

> even then we'd need a way to translate those keyword args into keyword args that make sense in the napari model, and there maybe would still be napari parameters not represented in that minimal set that users of napari would still wanted set on load, so we'd still have to let people provide their own keyword args for usage in napari anyway.

Sure, but I want to make sure that translation/expansion layer is as small as possible. Things like axis names and pixel/voxel spacing will exist in both namespaces, and it would be great if we converged on the same wording. (Worth looking at things like OME-TIFF/OMERO here.)

> I am curious what alternatives you can see to the dict approach described above or if you have ideas on how to improve the API there (particularly if one reader call is allowed to add multiple layers), as I do want us to do this in the least burdensome way possible.

Yep, good points, and I don't yet have a clear answer here, which is again why I think it might be good to get a "default" plugin infrastructure in more quickly. There's a clear divide between true metadata (stuff like spacing), and viewer parameters (like ndisplay or which axes to display). In fact, in viewer parameters, the user might want to override metadata in some way using keyword arguments, for example renaming `time since epoch` or some similarly annoying, microscope-written name to `t` for display purposes.

Anyway, it seems I'm coming around to most of the original ideas in this PR, thanks for being patient with me. =)

sofroniewn:
> Overall, I would say, let's go ahead with `napari-*` or, perhaps preferably, `napari-plugin-*`, and leave the function annotation approach for the future.

I'm fine with either, but maybe lean towards the simple `napari-*`. Can we imagine another usage for `napari-*` that wouldn't be a plugin or wanting to do something different with that in the future, in which case the more specific `napari-plugin-*` may be better at this point, but otherwise the simpler one will do.

> Add to that the fact that smaller PRs are faster and easier to review, and to me it seems like a slam dunk to get the first stage merged now-ish and continue discussing in a subsequent PR.

I'm fine with a simple PR to get in some of the basic machinery discussed above, but I don't think we should add support for an IO plugin that say just returns an array that we pass to add_image, when we still havn't decided that we will always support that behaviour (see discussion above about should IO plugins always return lists of things etc), because that then paints us into a corner we might not want to be in when we consider the more general case.

@tlambert03 do you have a clear idea on how to move this PR forward while keeping our options still open or are there more things you need to know?

> Anyway, it seems I'm coming around to most of the original ideas in this PR, thanks for being patient with me. =)

Always, I think these sorts of discussions are key to having a well designed API for napari and all of us approach them with an open mind, willing to change our views as we learn new things - which is almost always the case for me when you weigh in @jni!!

tlambert03:
great.  sounds like we're hovering around a near-consensus.  I think I have an idea on how to move forward but will spell it out here as a quick summary for those catching up on the whole thread (and to make sure we're on the same page):

- create a `PluginManager` class that registers napari plugins in one of three ways:
  - automatically, by looking for installed packages named `napari-*` _(I share @sofroniewn's preference for this over `napari-plugin-*`...  If we also want to name things `napari-*` later, I don't think it will be a problem, since we can also confirm internally that some package named `napari-*` is actually conforming to a particular known API, or just ignore it.)_
  - automatically, by looking for packages with a `napari.plugins` `entry_point` in their setup.py file.
  - manually, by calling a method on the PluginManager with the name of a module

- define a basic interface that a reader plugin needs to implement.
  - a reader function accepts a `PathLike` and returns a list of tuples in the form `[(data, [optional_meta]), ...]`. where the `meta` dict includes a special `layer_type` keyword along with keys to pass to `viewer.add_*`. _@jni, are you ok for now with that for now? or would you like to request a modification to that?_
  - a module top-level variable (e.g. `READERS = [(checker, reader)]`) that we look for.  The presence of this variable `napari_readerplugin.READERS` is what defines a package in the environment as a reader plugin (for now... until a better `imtypes` solution is established).

- update the `viewer._add_files` method to look for plugin readers.

variable names and stuff can be changed later, but let me know if you already think I've missed the point with the structure.


**future directions**:
- push for a community supported IO plugin API that looks for modules that declare functions with type annotations that accept a `PathLike` and return an `ImType` (once an `imtypes` library exists)

jni:
Re `napari-` vs `napari-plugins-`, my idea was simply to use a testbed to try these ideas out before creating too many `napari-` plugins with an API that we later want to revise, in particular regarding other types of plugins. I guess while we have full control over all the packages, this is not a big issue, but once we start getting packages created by third parties, it might be an issue.

Regarding the list of (data, meta), I wonder whether meta should be split in two (standard metadata vs keyword arguments for the viewer), but I'm not sure.

Finally, (sorry 😛), do we need a class vs a function here? From what I can tell, this can be accomplished by a function, similar to `add_plugins(module=napari, packages=None)` that does the work of crawling the packages (all installed packages if None) and adds them to some napari persistent variable. Or am I missing something?

At any rate, take the above as comments. I'm happy with the proposed solution for now.

tlambert03:
> Regarding the list of (data, meta), I wonder whether meta should be split in two (standard metadata vs keyword arguments for the viewer), but I'm not sure.

I was going to leave out arguments for the viewer or dims (like `ndisplay`) completely for now, due to the concerns you raised above.  And `meta` will be strictly defined as keywords that can be passed to an `add_layer` method (with the special/single exception of a `layer_type` keyword in the `meta` dict).  But if we did have viewer kwargs, I would make it a third dict, and I intend to consume these tuples with `*args` expansion in an internal function that looked something like this, which allows us to add items to that tuple (i.e. update the API) later:
```python
def add_plugin_layer(viewer, data, meta=None, dims=None):
    ...

# how that function will be used to handle `plugin_data` returned from the reader
new_layers = [add_plugin_layer(viewer, *layer_info) for layer_info in plugin_data]
```

> Finally, (sorry 😛), do we need a class vs a function here? From what I can tell, this can be accomplished by a function, similar to `add_plugins(module=napari, packages=None)` that does the work of crawling the packages (all installed packages if None) and adds them to some napari persistent variable. Or am I missing something?

were you bullied by classes as a kid or something? 😜
I'm making it as an `abc.Mapping` with `{module_name: imported_module}`.  let me submit the revised PR, then take a look at the class and tell me if you want me to change it all to top level module functions and variables.  I personally find it much easier to read/understand as a class... and need help understanding the problem with a class in this case.  I'm still not convinced that just because something _can_ be done with a function, it _should_

> Re napari- vs napari-plugins-, my idea was simply to use a testbed to try these ideas out before creating too many napari- plugins with an API that we later want to revise, in particular regarding other types of plugins.

let's revisit this later before merging anything.

jni:
> I'm still not convinced that just because something can be done with a function, it should

Ditto for classes. =) 

Re my personal history, I think it's a mixture of my first programming course was in Haskell, and I've found that when I want to contribute to an unfamiliar package, I have a much easier time if I'm dealing with functions than with classes, especially class hierarchies.

Now for my rationalisation of this. Functions are generally much easier to understand coming in: it takes some inputs, and produces some outputs or (worst case) side effects. In particular, *everything you need* to understand what a function does tends to be right there. In contrast, in OO programming you typically have a whole bunch of references to `self.whatever`s that are defined god-knows-where — `__init__` if you're lucky, but possibly some superclass, or it's a property, or it's a thing that gets added after running some *other* method that needs to be run before the method you're currently looking at.

In short, functional programming *tends* to be better at code and data locality, which are part of the ["Power of 10" rules](https://en.wikipedia.org/wiki/The_Power_of_10:_Rules_for_Developing_Safety-Critical_Code) from NASA/JPL for bug free code.

But, let's see what comes out of this. =)

tlambert03:
But none of that is happening here.  No complex inheritance, no complex __init__, no hidden attributes, and so far all deterministic functions... But by using a class we also get nicely grouped functions that all act on the same underlying `dict` and a little bit easier time using introspection tab completion in the console (and I prefer the usage syntax).  I definitely don’t think we should always use classes, but I take the decision on a case by case basis...  I happily defer to your experience, and I’m sure there’s stuff I’m not anticipating, but also wish it felt more logical and less dogmatic for the case at hand

ttung:
My very uninformed opinion is that the [consensus](https://github.com/napari/napari/pull/891#issuecomment-577399238) is a fairly good one, but way overkill for the application being described here (directory of tiffs).  Feel free to disregard, though, as I am still just learning. :)

jni:
@ttung directory of tiffs is just a throwaway example — indeed that already just works. More relevant here would be something like directory of tiffs + microscope parameters that determine some amount of deskewing to do with the tiffs. So, the checker would check for parameters in the right format in that folder, and then read the tiffs in a specific way (ie stacking along specific axes).

@tlambert03 well, I feel like dogmatic is a bit harsh, but sure, I suppose it is my policy to use functions *except* where I think classes are *necessary*, which usually for me means complex data grouping (as seen in our layers, for example). With regards to what you were mentioning,

- group related functions together: this is what modules do, including tab completion in the console
- act on the same underlying data: well, this is a global variable, unless you are going to have lots of different, long-lived instances of the data, which I think is *not* the case here (?), or it's a bunch of disparate data structures, in which case it makes a lot of sense to group them into a single object.
- (and I prefer the usage syntax) — well, that one we can't really resolve easily, only that I fall in the opposite camp. =P Happy to see where this goes, and maybe propose a functional alternative once the code is ready. I do agree that I feel like I'm getting into a language war when discussing these things in the abstract like this. Apologies for that.

tlambert03:
@jni, sorry I definitely don’t mean to come off as harsh or insulting.  I genuinely am just curious to dig into the topic and learn from your experience (which I really do respect a lot).  It don't think it comes through in my word choice, but I think more than trying to convince you, I’m trying to settle into and challenge my own understanding of the nuances (just want good reasoning). Apologies again.

@ttung, to add to @jni, consider also proprietary image formats (non TIFF) for which python readers exist. We need to somehow offer a hook for people to use those easily without explicitly supporting every possible format ourselves

sofroniewn:
Convo sounds like it's going great, and we can pick up again once @tlambert03 has a prototype based on the consensus views. I'll quickly note is in response to

> Regarding the list of (data, meta), I wonder whether meta should be split in two (standard metadata vs keyword arguments for the viewer), but I'm not sure.

all of our layers currently accept a keyword arg `metadata` so inside this `meta` dict there you could use the `metadata` keyword arg to propagate any standard metadata not needed for the layer explicitly.

Maybe one more API idea would be to have the return be
```python
([(data_1, meta_1), (data_2, meta_2) ... ], viewer_meta)
```

i.e. a tuple of a list and a viewer meta data dict, where the list is the object we've described above with the meta dict only containing layer data and layer meta data (which includes the `metadata` keyword arg for anything extra not explicitly on the layer), and the `viewer_meta` is a single dict with the dims / camera stuff (this could be optional, or forced to be empty).

I feel that if we're adding 10 layers it doesn't make sense for each layer to define its own dims metadata as only the last / first one will have any effect, but I'm also wary of this becoming too complex

tlambert03:
> I feel that if we're adding 10 layers it doesn't make sense for each layer to define its own dims metadata as only the last / first one will have any effect

yeah, I realized that too... it's definitely not ideal.  We can certainly "merge" non-conflicting keywords across all layers (so it doesn't just have to take first/last all-or-nothing), but ultimately it's up to the plugin developer to realize that they can't deliver conflicting directives within the same return value (which I don't think is too big of an ask).
the one thing I don't like about the `([(data_1, meta_1), (data_2, meta_2) ... ], viewer_meta)` return type is that we are now enforcing an additional nested iterator:  even if they didn't want to provide `viewer_meta` (and most probably won't), they still need to return a tuple... _or_ we need to do additional type checking on our end, which i'd like to avoid... (whereas the optional third argument strategy is opt-in and doesn't change the `list-of-tuples` return type).

I'll be using type annotations heavily here, so we can keep discussing it in the context of the PR:
https://github.com/napari/napari/blob/372595aabc9965e579b5a2081dd532531a3bbbd3/napari/plugins.py#L102
where...
https://github.com/napari/napari/blob/372595aabc9965e579b5a2081dd532531a3bbbd3/napari/plugins.py#L27-L28

kne42:
FWIW I'm also biased towards a function, assuming that this is a singleton.

I swing towards classes in the following cases:
- I'm using some sort of internal state that needs to be duplicated
- I'm defining a new data type
- I want to use custom syntactic sugar

If we were to give each viewer its own plugin manager (e.g. we want some plugins to be enabled on one viewer but not another), then I would say we should use a class. Otherwise, functions all the way!

tlambert03:
thanks @kne42!  Yeah, currently I'm making it a singleton, so all that seems relevant here.  had a quick chat with @jni about this offline as well.  I'll leave the class in the PR for a little bit for the sake of discussion, but will also try to add a functional approach modeled after the skimage plugins for the sake of discussion.  more soon(ish)

tlambert03:
just took a good look through [skimage.io.manage_plugins](https://github.com/scikit-image/scikit-image/blob/master/skimage/io/manage_plugins.py#L31), and I see no reason why we shouldn't just basically mimic it as our functional approach.  I will put it into code within the next day or two, but If anyone wants to review it  for the purpose of meeting discussion,  here would be my main questions and some points to discuss:

- do we want to start declaring string-based plugin 'types' (as is done there with `imread`, `imread_collection`, `imshow`, etc...), but just have a single (reader type) plugin type for now?
- do we want to distinguish between a reader plugin that knows how to handle folders  vs one that only accepts individual files?  or should we just let that distinction be handled by the plugin itself (for instance, in its "checker" function), and call them all `reader` or `imread`?
- I'd prefer to avoid the `.ini` file approach used there, which makes a bit less sense when using name convention discovery and/or setup.py `entry_points` for discovery.  But it's worth discussing how plugins will declare the interfaces that they provide.  Currently in my example, plugins identify themselves as providing a reader by declaring a top level `READERS` variable (list of `[(checker, reader) ... ]`), and they can internally name functions whatever they want (this also allows them to declare multiple readers in a single plugin).  Declaring other non-reader interfaces could work similarly.  Everyone ok with that?
- we haven't dwelled much on the `(checker, reader)` convention since https://github.com/napari/napari/pull/891#issuecomment-575959941.  I like requiring plugins to declare a `checker` function as I think it will let us fly through the plugin chain easily, but would welcome renewed input on that.

jni:
Just a minor point, the pyproject.toml way to describe plugins, maybe, just from a quick search "python pyproject.toml plugins": https://python-poetry.org/docs/pyproject/#plugins

For more background on pyproject.toml, see: https://www.python.org/dev/peps/pep-0517/

jni:
Minor point about what we settled on in the meeting, a list of `(data[[, meta], layer_type])` tuples. There was some concern that if 3-tuples somehow became 4-tuples and 5-tuples, this would be annoying and maybe we should use dicts and `**` unpack them. The nice thing is that if we ever decide to go that way, the tuples should be easily backward compatible:

```python
if type(layer_data) == tuple:
    layer_data = dict(zip(['data', 'meta', 'layer_type'], layer_data))
# ... proceed using new dict interface.
```

So I suggest we stick to tuples for now, they are nice and easy to construct, and switch to dicts if the need arises.

tlambert03:
> Just a minor point, the pyproject.toml way to describe plugins, maybe

Reading up on it now, will add support for it!

sofroniewn:
Something I want to raise that hasn't come up yet is - if we go with the naming convention (e.g. `napari-`) for plugins on PyPi, is there all of a sudden going to be a rush to register all sorts of names of standard ones and what is our attitude towards that going to be. I.e. should we the core team @napari/core-devs start registering some of these now? Should we as individuals register them if we want to make them? Or do we just let them be available first come first serve for the community.

For example the following all seem like pretty basic packages we might want to make or support at some time:
- napari-segmentation
- napari-registration
- napari-tracking
- napari-classification
- napari-denoising
- napari-spot-finding
- napari-io

and the community might have expectations that packages with such fundamental names are of a certain quality etc

joshmoore:
Hi all. Just getting caught up with this lovely thread :wink:

> https://github.com/napari/napari/pull/891#issuecomment-575963286 This could be a very good idea. I feel like bioformats has taken this approach, but I'm not sure, and I'm curious if they have any advice for us on that matter. I like it in some ways, but I'm also wary of many checks, and the idea that two plugins might pass a check but then return data in different ways, and forcing people to be explicit about what plugin loader they want might be good, but I can also see people might like that magic and it might be needed if we have many IO plugins.
> 
> Also curious what @joshmoore from the bioformats team makes of all this - it's still super early in our thought process though so be gentle ;-)

 - There have definitely been some gotchas in the "isThisType" determinations over the years. Both in terms of prioritization of the plugins (who gets a chance first) as well as broken assumptions on certain platforms (Don't call `isFile` more than you need to on Windows!)
 - Performance will eventually be an issue, _especially_ if one of the readers is also calling out to Bio-Formats which is then calling out to its list of readers....

cc: @dgault in case I've forgotten anything.

>  https://github.com/napari/napari/pull/891#issuecomment-575974754 I would like us to work with the community here, meaning imageio, PIMS, or something else,

From a conversation _somewhere_ ([slack!](https://cziscience.slack.com/archives/C91PBDKGS/p1560297968026200)) with @jni, I was certainly working under the impression that a target would be to have Bio-Formats as a plugin in imageio as a follow-on to the [PIMS reader](https://github.com/soft-matter/pims/pull/301), but that ...

> https://github.com/napari/napari/pull/891#issuecomment-575995174 how would you then signal that the ArrayLike was actually multichannel and should be shown with three different channels with specific names, scales, colors, and possibly displayed immediately in volume view?

... raises the question of whether or not the returned object will do everything we need it to (n-dim, metadata, etc.)

> https://github.com/napari/napari/pull/891#issuecomment-575995174 I guess I still need examples of how napari becomes aware of those things.

A naive question but: is the goal here more the GUI interaction? Seems like in the headless case the developer would import the library and use it, no?

> https://github.com/napari/napari/pull/891#issuecomment-576032134 I'd also be curious to hear what the bioformats team has to say, but to my understanding, this is more or less exactly what bioformats and python-bioformats does. All loci.formats.FormatReader subclasses have a method isThisType that accepts a filename or bytes and returns a boolean (whether it matches the reader format). In python-bioformats you can see an example here of iterating through the all known reader classes, calling isThisType() until one of them returns True.

This is all true, but in Bio-Formats' case "then return data in different way" isn't particular an issue. The data structures are all so typed & defined that you're sure of what you're going to get. The downside is that it's likely less flexible (and likely less exciting to implement).

> https://github.com/napari/napari/pull/891#issuecomment-576276344 But, I am certainly getting the sense of urgency here and I understand that getting the typing info to propagate across the ecosystem is going to be a massive effort, possibly a year or two.

✋ Concur on the timeline, but I'm _very_ interested in this aspect of the solution.

> https://github.com/napari/napari/pull/891#issuecomment-576471483 I think we're not trying to write one generic IO library with standard metadata for all of imaging - which I think we agree would be a mistake, but I see it as more that we're trying to provide a way for developers to write code that could help non-coders look at their data more easily in napari which necessitates the usage of keywords that make sense in napari.

If this is the consensus target here, :+1: for the summary and I can see how the likes of https://forum.image.sc/t/napari-loading-images-from-omero/30727 could be encapsulated for others. (cc: @will-moore)

> https://github.com/napari/napari/pull/891#issuecomment-577399238 future directions: push for a community supported IO plugin API that looks for modules that declare functions with type annotations that accept a PathLike and return an ImType (once an imtypes library exists)

:+1:

tlambert03:
> (@sofroniewn): if we go with the naming convention (e.g. napari-) for plugins on PyPi, is there all of a sudden going to be a rush to register all sorts of names of standard ones and what is our attitude towards that going to be

that's an excellent point.  PyPI namespaces will be rather valuable for the naming convention discovery.  I would support the idea of claiming high value names, but curious what others have to say.  This sort of phenomenon happens all over the place with extensions (I feel like it's particularly bad with rapidly-developed node/npm packages where it can be unclear whether the extension is supported by the official repo or not), and I really don't know whether there are established best practices.

Relatedly, I have been thinking that packages like PIMS (should they choose to support napari) might be better off using `pyproject.toml` or `setup.py`: (`entry_points={"napari.plugin": ["pims = pims.napari"]}`) instead creating a new `napari_pims` package ... as that would then allow  
`pip install napari pims`. (unless of course they'd rather keep napari-specific code out of the repo).

tlambert03:
@joshmoore:
thanks so much for your feedback!   it's really helpful to have someone with your experience in this domain paying attention and weighing in.  I will say that we've progressed a bit in our internal discussions since some of topics were introduced, and are beginning to identify some patterns that I think we're happy starting with.  I now need to find a decent stretch of time to put our ideas into some code, and it would be great to get your feedback again at that point!

couple direct responses/questions:

> Performance [of the `checker` pattern] will eventually be an issue, especially if one of the readers is also calling out to Bio-Formats which is then calling out to its list of readers....

I'm definitely sensitive to that possibility.  I have some (potentially naive) strategies for imposing internal time limits on the checker, but it very much remains an open question.  I guess I was sort of hoping that the end-user might play at least some role in this process: if they can set order/priority of reader plugins, and if we can provide useful warnings (or perhaps even timeouts) if any particular plugin seems to be slow, then i had hoped to foster a situation where the end user can at least a) tune their own experience or b) say "why is this plugin so slow" and disable it/contact the developer.  But again, as the number of potential plugins goes higher, I absolutely recognize that this is a potential gotcha... (and happy to hear suggestions!)

> is the goal here more the GUI interaction? Seems like in the headless case the developer would import the library and use it, no?

I think it's safe to say that the "urgent" need described here arose from watching people using the GUI... and wanting to help non-coders, or semi-coders load semi-steroetyped datasets and formats into the viewer, by simply being able to `pip install` something from another developer.  But we do want to design it in a general way that is extensible.  

>I was certainly working under the impression that a target would be to have Bio-Formats as a plugin in imageio as a follow-on to the PIMS reader...

If I am interpreting this correctly, I think this touches a bit on the "[future direction](https://github.com/napari/napari/pull/891#issuecomment-577399238)" goal.  Ultimately, the driving principle is "napari headless is just python", but we also want to provide an opt-in framework for developers to help out non-developers (that can simply be pip installed).   And, as you mention, there are some decidedly napari-specific concepts (layers) that don't yet have a community spec (like "`imtypes`")... So in the meantime, we propose that plugging something like PIMS into napari would look like one of two approaches:
1) someone creates a new package on PyPI called, e.g., `napari_pims`, that imports `pims`, and defines the `READERS = [(reader, checker)...]` interface that we are proposing.
Then `pip install napari napari_pims`

2) (as I mentioned in my response to @sofroniewn above) PIMS itself could support napari with a `pims.napari` module and an `entry_points={"napari.plugin": ["pims = pims.napari"]}` in their `setup.py`.
Then `pip install napari pims`

...
I do hope you'll keep an eye on our progress as we'd very much welcome your continued input!

tlambert03:
I had one more idea that I'd like @napari/core-devs input on:

Rather than requiring a napari reader plugin to define two functions (`checker`, `reader`), what do you think of having a reader function be a single generator function, where it first yields true or false if recognizes the path, and then yields the layer_data next. For instance, this would be a deltavision plugin:

```python
from mrc import bindFile

def reader(path: str):
    yield path.endswith(('.dv', '.mrc'))
    return bindFile(path)
```

internal usage in napari would be:
```python
for plugin in reader_plugins:
    generator = plugin(filepath)
    layer_data = next(generator) if next(generator) else None
    if layer_data is None:
        continue
    ...
```

Confirming that a plugin was conforming to the API would be:
```python
import time
from types import GeneratorType

output = reader_plugin('fake_path')
assert isinstance(output, GeneratorType), 'Wrong reader function format'
start = time.time()
assert next(output) == False, "Do you *really* know how to read 'fake_path'??"
end = time.time()
assert (end - start) < MAX_CHECK_TIME,  "Too slow, Joe."
```

is it maybe too much to ask of developers to write a generator?

sofroniewn:
Ooh interesting - is there a way that if the generator isn’t provided it will still work, (with a try / catch)? Or maybe that will encourage people not to provide it. Overall I like the idea. We’re also using generators in our mouse functions so I think it maybe a good idea. Curious what @kne42 thinks

tlambert03:
> is there a way that if the generator isn’t provided it will still work, (with a try / catch)? Or maybe that will encourage people not to provide it.

I can think of ways to quickly make sure they are returning a generator (shown above), but I can't think of how to make it still work if they _don't_ return a generator.  since now we'd be stuck without a `checker` method... and their function could take an arbitrary amount of time.  I think I would probably want to run the "confirmation" code above on plugin import, and exclude it with a warning if it failed.  Worst case scenario is that the plugin reader function is _not_ a generator and takes a long time to return, so we could put that assertion code in a [timeout context](https://stackoverflow.com/a/601168/1631624) to be safe.

jni:
Hi all! tbh I don't love the generator idea...! `next(generator) if next(generator)` looks terrible and unreadable to me, and the whole thing is much less "self-documenting". It's also pretty unconventional use of generators, so an extra thing for people to get their brains around. I don't see a huge advantage to writing one weird function instead of two simple functions! Is there another reason why you would think this would be a nicer API?

I guess the rationale was, can we do a lazy `imread` on the file that we only execute on demand? But even then, I would ask for a function that returns a (bool, generator) tuple. Ultimately, though, I think it adds complexity for very little gain...

tlambert03:
Good enough for me!  I agree on all accounts.  The only benefit here was that it is a single function instead of two...  but let’s kill it
I have one more consideration for you that I’ve been looking into today and will post soon.

kne42:
I am most in favor of the try-except style because it looks like an ordinary Python function and uses EAFP rather than the LBYL that the separate function style uses.

I'm not a fan of the generator style since it seems a bit unnatural and has too similar a structure to the try-except style.

I would prefer to use the try-except style and have the separate function style as a secondary method. However, I will note that the separate function style would be better if we want to filter out which files can be read in the "open" dialogue.

tlambert03:
this has been a fun PR/thread :) but i think it's time to thank it for its service and move further discussion to #908 or #936

Issue #858, opened 2020-01-08
Refactor Points data model


kevinyamauchi:
## 🚀 Feature
This issue proposes to allow the user to store point metadata (`Points` layer) and allow setting of point properties as a function of the point metadata. Additionally, this issue explores the possibility of combining the `Points` and `Text` layers.

**Definitions**

For the purpose of this issue, I will use the following definitions:

- _point annotions_: metadata associated with each point. These are numbers or strings that describe each point and are set by the user. Each point can have multiple annotations. Some examples of possible annotations: class labels or descriptions of the points.
- _point properties_: display attributes of each point (e.g., `face_color`, `size`, and `edge_color`) that get passed to the vispy `Markers` node. 

## Motivation 

This proposal draws from the following discussions:

- #651 : @adamltyson proposed to add a class label to points
- #600 : @jni proposed to store the point properties in a `DataFrame` and combine `Points` and `Text` into a single layer (both toggleable)

### Approach
**Store point annotations in a `DataFrame`**

As proposed by @jni in #600, the point annotations  will be stored a `DataFrame` (`Points.annotations`). Properties can be definied upon instantiation of the layer using keyword arguments. All keyword arguments outside of the defined parameters will be captured  (with the `**kwargs` syntax) and used to instantiate the `Points.annotations` `DataFrame`.

```python
with napari.gui_qt():
    # create the viewer with an image layer
    viewer = napari.view_image(rgb2gray(data.astronaut()))
    
    # create the point data
    points = np.array([[100, 100], [200, 200], [333, 111]])
    size = np.array([10, 20, 20])
    
    # associate a class with each point
    point_classes = ['type A', 'type B', 'type A']
    
    points_layer = viewer.add_points(points, size=size, point_class=point_classes)
    
```
If no point annotations are provided upon layer creation, `Points.annotations` is instantiated as an empty `DataFrame` with an index corresponding with the indices of the points.

The point annotations can be added, modified, and deleted using standard `DataFrame` syntax. For example, we can add a new column (annotation) as follows:

```python
points_layer.annotations['gene_target'] = ['ACTB', 'GAPDH', 'ACTB]

```

**Setting Point properties**

Currently there are two types of point properties: (1) uniform properties, which have to be constant across the node and (2) per-point properties that can be set on a per-point basis.

Uniform properties (`symbol`, `edge_width`) will be set via our current setter methods (e.g., `Points.symbol = 'ring'`)

[...]

Issue #862, opened 2020-01-09
Generic colormap representation


sofroniewn:
## 🚀 Feature
Right now in our model files we use the vispy colormap object. It would be nice to completely separate our model files from vispy as noted here https://github.com/napari/napari/pull/686#discussion_r363550535 in discussions around serializing the viewer and round-trips.

After #782 (which does a similar refactor but for colors) and #854 (which adds a camera model) both get merged our models will be vispy independent.

Note I'm not sure yet what this representation should be or if vispy colormaps are serializable as they are now, but am making this issue to discuss this.

jni:
I'd love for this contribution to be to vispy or matplotlib, rather than creating yet another colormap representation...

sofroniewn:
Agreed - I think if we use a tuple that is an Nx4 array of colors, an optional length N list of control points, and an interpolation mode string that is either `linear` or `zero` then we can recreate any vispy colormap simply from its API here - http://vispy.org/color.html#vispy.color.Colormap

We can revisit this once #782 lands and when we get further in discussion of #858

tlambert03:
Just want to put a few thoughts here from our conversation yesterday... and hopefully we can do something with this to simplify the work on the models.

I don't totally mind the idea of a new object `(colors, controls, interpolation, name)`, but I am also sensitive to Juan's comment about creating "yet another colormap representation".  If the counter argument is that we don't want our users to need vispy installed to work with a colormap object, then I guess I feel like we have "bigger fish to fry" when it comes to extracting that dependency, and i wonder whether we aren't just making our lives harder by having a simpler `cmap` object that we have still to convert to-and-from internally whenever we need a "real" vispy object.

It sounds like you've got a pretty clear vision on this and feel strongly about it, so I'd be interested to see some examples of what you'd propose, along with your concerns and driving principles... so we can have some more concrete conversations about it.

sofroniewn:
> It sounds like you've got a pretty clear vision on this and feel strongly about it, so I'd be interested to see some examples of what you'd propose, along with your concerns and driving principles... so we can have some more concrete conversations about it.

yup i totally recognize the burden here is on me to provide something concrete. I will give it a go, but it might not be till next week. I'm offline this weekend. I do feel pretty strongly about this one though so want to give it my best shot to convince everyone, recognizing your concerns too!!

tlambert03:
> I'm offline this weekend.

that's fine!  nothing urgent.  look forward to your thoughts

PR #895, opened 2020-01-21
Allow Points display properties to be set by point properties

kevinyamauchi:
# Description
This PR introduces the ability to define annotations for each point and set `face_color` and `edge_color` as a function of the annotations.

### Annotations
I added a property containing user-defined annotations for each point (`Points.annotations`). `Points.annotations` is a dictionary where the keys are the names of the annotations and the values are numpy arrays with an annotation for each point. For example, to create a `Points` layer with a "class" for each point, we would would do the following. 

```python
with napari.gui_qt():
    # create the viewer with an image layer
    viewer = napari.view_image(rgb2gray(data.astronaut()))
    
    # create the point data
    coords = np.array([[100, 100], [200, 200], [333, 111]])
    
    # annotate each point with a class
    annotations = {
        'point_class': np.array(['A', 'A', 'B']),
        'best_point': np.array([True, False, False])
    }
    points_layer = viewer.add_points(coords, annotations=annotations)
```
We can add and modify annotations via the `Points.annotations` property

```python
# add annotations with standard dictionary syntax
points_layer.annotations['gene_target'] = ['ACTB', 'GAPDH', 'ACTB]

# similarly, we can modify an existing annotation
points_layer.annotations['point_class'][1] =  'B'
```

### Setting colors with annotations
We can also set color properties as a function of the annotations. 

```python
with napari.gui_qt():
    # create the viewer with an image layer
    viewer = napari.view_image(rgb2gray(data.astronaut()))
    
    # create the point data
    coords = np.array([[100, 100], [200, 200], [333, 111]])
    
    # annotate each point with a class
    annotations = {
        'point_class': np.array(['A', 'A', 'B']),
        'best_point': np.array([True, False, False])
    }
    
    # the color cycle can be a list of color names
    face_color_cycle = ['green', 'blue']
    
    # make the points layer
        coords,
        annotations=annotations,
        face_color='point_class'
        face_color_cycle=face_color_cycle
        
    )
```

We can also directly set the `face_color`

```python
# set face_color to a single color by name
points_layer.face_color = 'red'

# set face_color to a single color by RGBA
points_layer.face_color = [1, 0, 0, 1]

# set face_color for each point (also works for RGBA as above)
points_layer.face_color = ['red', 'blue', 'blue']
```

Similarly, we can set the color cycle used to set the face_color

```python
# set face_color_cycle using an array of color names
# (gets converted to cycle of RGBA numpy arrays)
points_layer.face_color_cycle = ['red', 'blue', 'blue']

# set face_color_cycle using RGBA array
color_cycle = np.array([[1, 0, 0, 1], [0, 1, 0, 1]])
points_layer.face_color_cycle = color_cycle

# we can also directly pass a cycle
from itertools import cycle
points_layer.face_color_cycle = cycle(color_cycle)
```

### Adding points and copy/paste
Adding points works as before. The added point takes on the same properties (including annotation) as the last selected point. If the `face_color` is being set by an annotation, then the annotation is copied and then `face_color` is calculated based on the annotation.

Similarly, copy/paste works as before: all properties (including annotations) are transferred from the copied points to the pasted points.

### Example
I have added an example demonstrating setting the face color with annotations based on the use case provided by @adamltyson in #858 (`/examples/add_points_with_annotations.py`). In this example, the point face color is set by an annotation. Additionally, there is a function that toggles the annotation (and thus face color) of the selected point that is bound to the `t` key.

### Layer properties GUI
This PR is focuses on adding support for annotations and setting point properties with annotations to the model and view. Once we have solidified the API and data model, I will make an follow-up PR to modify the layer properties GUI to support the different property setting modes.

## Type of change

# References
This is the first PR of #858.

Also, shoutout to @HagaiHargil , as this build upon the awesome color handling added in #782! 

# How has this been tested?
- [x] the test suite for the `Points` layer passes
- [x] all tests pass

## Final checklist:


kevinyamauchi:
I have implemented point annotations and setting `face_color` with a color cycle applied to an annotation. It would be great to get some early feedback, @napari/core-devs . If you all think it is on the right track, I can propagate the changes to `edge_color` and then work on adding colormaps for continuous annotations.

I think it would be nice to have similar behavior for the point size, but I think that should go into a separate PR.

napari/components/viewer_model.py
```diff
@@ -607,6 +609,9 @@ def add_points(
         ----------
         data : array (N, D)
             Coordinates for N points in D dimensions.
+        annotations : dict
```
jni: (a) I would call this "dict {str: array (N,)}", and
(b) I would aim to support both dicts of columns and pandas dataframes, and, in the future, dask dataframes and cudf dataframes.

napari/layers/points/_constants.py
```diff
@@ -3,6 +3,22 @@
 from ...utils.misc import StringEnum
 
 
+class ColorMode(StringEnum):
+    """
+    ColorMode: Color setting mode.
+
+    DIRECT (default mode) allows each point to be set arbitrarily
```
jni, line 10: Ooh, clever. =)

adamltyson:
Hi @kevinyamauchi. Just a note to say that this is great from a user point of view. I tested this for our use case, and everything just works without issue. The API is clear, and implementing this for a custom application is trivial.

Thanks!

napari/components/viewer_model.py
kevinyamauchi: Thanks for taking a look, @jni !

> (a) I would call this "dict {str: array (N,)}", and

Good point. I will fix this.

> (b) I would aim to support both dicts of columns and pandas dataframes, and, in the future, dask dataframes and cudf dataframes.

I agree with keeping compatibility with DataFrame and DataFrame-like things in mind for the annotations. There was some discussion in #858 about trying to avoid picking up dependencies, especially for the model, so I opted not to implement conversion methods here (for now). In terms of `dask` and `cudf` `DataFrame` support, I _think_ those might be dropin if we make our internal annotations datatype a `pandas` `DataFrame` (in a similar way, we can use dask arrays in the `Image` layer). I have tried to implement this such that we don't touch annotations too much and the slicing is simple, so if/when we decide to migrate to a `DataFrame`, I think it's a pretty straightforward refactor.

Since our annotations are fairly simple for now, I propose the following:

- We keep `Points.annotations` as a dictionary of arrays.
- When this lands, I will update the `Points` layer tutorial to include these new features. I will also add a section on converting `DataFrame` to our `Points.annotations` data type. That way if users already have annotations in a `DataFrame`, they can still take advantage of the annotations.
- If there is demand for native `DataFrame`, we can refactor and make `Points.annotations` a `DataFrame`. That might also be an opportunity to make sure the coordinates are also implemented in a way that can utilize dask arrays.

How does this sound?

kevinyamauchi:
Thanks for the feedback, @adamltyson ! I am glad you like it so far!

kevinyamauchi:
I have finished adding the core features for this PR (i.e., annotations and setting colors with colormaps/cycles on annotations) and added some more tests and examples. I have described the biggest updates below. As implemented, I believe this is not a breaking change (all existing examples and tests continue to run).

**Choosing color modes**
I added a properties for setting the color mode (`Points.face_color_mode` and `Points.edge_color_mode`) that can be set to one of three modes:
- `ColorMode.DIRECT`: user can set colors either with a single color applied to all points or an array of colors specifying each point's color.
- `ColorMode.CYCLE`: the point colors are set by applying a color cycle to an annotation (used for categorical annotations).
- `ColorMode.CMAP`: the point colors are set by applying a color map to an annotation (used for continuous annotations)

When setting a color (e.g., `Points.face_color`), the `Points` layer tries to guess which mode to use based on the input: 
- If a string that corresponds to an annotation name, it then guesses whether `ColorMode.CYCLE` or `ColorMode.CMAP` is more appropriate based on the datatype of the annotation array.
- If the user provides a string that is not an annotation name, it tries to map it to a color and broadcasts it to all points (`ColorMode.DIRECT`). If that fails (i.e., the string was neither an annotation nor a color name), an error is raised.
- If the user provides a color array (i.e., Nx4 array of RGBA colors), it tries to broadcast it to all points (`ColorMode.DIRECT`). If that fails (i.e., the array is an incompatible shape), an error is raised.

The user can also directly set the color mode (e.g., `points_layer.face_color_mode = 'cycle'`).

**Setting colors with cmaps**
I have added the ability to set `face_color` and `edge_color` using color maps on annotations. See below for the API. For now, I am using the `vispy` colormaps, as we already use them in the image layer (i.e., no new dependencies). However, as noted in #862, we probably want to eventually move to our own colormap representation towards removing the `vispy` dependency from the model.

```python
with napari.gui_qt():
    # add the image
    viewer = napari.view_image(rgb2gray(data.astronaut()))
    # add the points

    # create annotations for each point
    annotations = {
        'confidence': np.array([1, 0.5, 0]),
        'good_point': np.array([True, False, False])
    }

    # create a points layer where the edge_color is set via
    # a color map (grayscale) on the confidence annotation.
        points,
        size=20,
        annotations=annotations,
        edge_color='confidence',
        edge_color_cmap='gray',
        edge_width=7
    )

    # set the face_color mode to cmap
    points_layer.edge_color_mode = 'cmap'

    # change the colormap
    points_layer.edge_color_cmap = 'viridis'

    # we can then change back to a color cycle
    # since we didn't specify a color cycle, it will use the default
    points_layer.edge_color_mode = 'cycle'
```

**Examples**
- I have updated the 2D example (`/examples/add_points_with_annotations.py`) with more descriptive comments and an example of applying a colormap to an annotation.
- I have added an nD example (`/examples/nD_points_with_annotations.py`)

kevinyamauchi:
All tests other than the os x python 3.7 pyqt5 tests pass, so I think this is ready for another round review, @napari/core-devs!

napari/components/viewer_model.py
jni: @kevinyamauchi I think we can support dataframes without depending on pandas, if you just get the data as `np.asarray(annotations[colname])`. This should work with annotations being a dict of arrays or a pandas dataframe.

kevinyamauchi: Oh cool, I didn't think of that! I'll try that out. Thanks, @jni!

If it works, I'd like to add a test. If I do, should I add `pandas` to /requirements/test.txt`?

kevinyamauchi: @jni, I just took this for a spin and it mostly works. The one wrinkle is adding and removing annotations (e.g., when adding or deleting a point). I am not super familiar with the pandas API, but I can't figure out a way to do so without the output being a dictionary. I think converting to a dictionary is fine if we want to simply support inputing a pandas dataframe, but I think we would lose the advantages of a dask or cudf DataFrame. 

For example, if we want to delete the the annotations at `index` (could be one index or a list of indices), we could use the following:

```python
self.annotations = {
    k: np.delete(v, index, axis=0) for k, v in self.annotations.items()
}
```

In pandas, I think we would usually use `df.drop()` to remove the whole row associated with the indices we would like to remove. We could modify the snippet above to accept a pandas DataFrame in `self.annotations`, but the output would be a dictionary. 

```python
self.annotations = {
    k: np.delete(np.asarray(self.annotations[k]), index, axis=0) \
        for k in self.annotations
}
```

Do you know of a better implementation? If not, perhaps a good middle ground is to convert if a DataFrame is provided, convert it into a dictionary when setting `Points.annotations`. We can then consider making `DataFrame` the internal representation later if there is demand. What do you think?

kevinyamauchi:
I have made the changes we discussed in the sync meeting today. In particular, I:
- changed `Points.annotations` to `Points.properties`
- added support for DataFrames properties by converting to our `Points.properties` dictionary format in the setter method.
- changed variable names to match their counterparts in the other layers
- changed the colormaps to return a tuple of ( colormap_name, colormap_object)

I think this is ready for review!

napari/components/viewer_model.py
```diff
@@ -587,11 +587,18 @@ def add_points(
         self,
         data=None,
         *,
+        properties=None,
         symbol='o',
         size=10,
         edge_width=1,
         edge_color='black',
+        edge_color_cycle=None,
+        edge_color_colormap='viridis',
```
sofroniewn: curious if you think this could just be `edge_colormap` and `edge_contrast_limits` - might be a little easier to read. same for `face_colormap` and `face_contrast_limits`. I'm also fine with keeping it more explicit.

sofroniewn (approved):
This looks great to me! I left a very minor comment on naming. This will cause merge conflicts with #897, but they will be easy to fix either way.

I'm approving, but maybe @kne42 and @jni want to review too before merge

napari/components/viewer_model.py
kevinyamauchi: I think this makes sense. I'll make the change now!

kevinyamauchi:
I have made @sofroniewn 's requested changes. I will wait to see if @jni and/or @kne42 have comments before merging. Thanks everyone!

kevinyamauchi:
Good morning! I have resolved the conflicts from #897 and the tests are passing other than the conda install failures. Does anybody else have comments or want to review (cc @napari/core-devs )? No worries if you need more time!

sofroniewn:
@kevinyamauchi can you change the title of this PR as we don't use the word annotations anymore

kevinyamauchi:
> @kevinyamauchi can you change the title of this PR as we don't use the word annotations anymore

Fixed!

kevinyamauchi:
i've merged in the test fixes and everything is passing. Unless I hear otherwise, I will merge tonight. Of course, I am happy to wait if anybody else wants to weigh in and needs more time!

examples/add_points_with_properties.py
```diff
@@ -0,0 +1,55 @@
...
+    points_layer = viewer.add_points(
+        points,
+        properties=properties,
+        size=20,
+        edge_width=7,
+        edge_color='confidence',
+        edge_colormap='gray',
+        face_color='good_point',
+        face_color_cycle=face_color_cycle
+    )
```
jni, line 38: I just love this. =D

examples/add_points_with_properties.py
```diff
@@ -0,0 +1,55 @@
...
+
+    # set the face_color mode to colormap
+    points_layer.edge_color_mode = 'colormap'
```
jni, line 41: The comment and statement are in conflict...

examples/add_points_with_properties.py
```diff
@@ -0,0 +1,55 @@
...
+        selected_points = viewer.layers[1].selected_data
+        if selected_points:
+            selected_annotations = viewer.layers[1].properties['good_point'][selected_points]
+            toggled_annotations = np.logical_not(selected_annotations)
+            viewer.layers[1].properties['good_point'][selected_points] = toggled_annotations
+
+            # we need to manually refresh since we did not use the Points.properties setter
+            # to avoid changing the color map if all points get toggled to the same class,
+            # we set update_colors=False (only re-colors the point using the previously-determined color mapping).
+            points_layer.refresh_face_color(update_color_mapping=False)
```
jni: I'd like to make a few changes to this. Specifically,

- `selected_points` is a list right now but it could one day be an array, and then checking the truthiness of it like this will fail. I suggest `if len(selected_points) > 0:`.
- the three lines could be more easily followed and understood if you grab the property directly:

```python
good_point = viewer.layers[1].properties['good_point']
good_point[selected_points] = ~good_point[selected_points]
```

Finally, the `refresh_face_color` method seems a bit clunky to me... I don't have a great solution for this right now, but it would be nice to brainstorm a bit about it... 

Oh and you should refer to the layer in one way only... either `viewer.layers[1]` or `points_layer`... I prefer the latter since it'll work totally fine even if the user rearranges the layers.

examples/nD_points_with_properties.py
```diff
@@ -0,0 +1,49 @@
...
+    for z in range(blobs.shape[0]):
+        points += [
+                [z, 43, 43],
+                [z, 43, 86],
+                [z, 86, 43],
+                [z, 86, 86]
+        ]
+
+    face_property = np.array([True, True, True, True, False, False, False, False] * int((blobs.shape[0] / 2)))
+    edge_property = np.array(['A', 'B', 'C', 'D', 'E'] * ceil(len(points) / 5))[0:len(points)]
```
jni: I have no understanding of what is happening here...

examples/nD_points_with_properties.py
```diff
@@ -0,0 +1,49 @@
...
+
+    properties = {
+        'face_property': face_property,
+        'edge_property': edge_property
+    }
+
+    points_layer = viewer.add_points(
+        points, properties=properties,
```
jni: I guess we're not running black on examples, which I like, but in this one case, I totally agree with it that `properties=properties` should be in its own line...

examples/nD_points_with_properties.py
```diff
@@ -0,0 +1,49 @@
...
+        size=3,
+        edge_width=5,
+        edge_color='edge_property',
+        face_color='face_property',
+        n_dimensional=False
```
jni: ... and that `n_dimensional=False` should get a trailing comma. =)

examples/nD_points_with_properties.py
```diff
@@ -0,0 +1,49 @@
...
+    )
+
+    # change the color cycles
+    points_layer.face_color_cycle = ['white', 'black']
+    points_layer.edge_color_cycle = ['c', 'm', 'y', 'k']
```
jni, line 56: Are you leaving this to have four values to demonstrate that `c` will be repeated? And is that indeed what will happen?

napari/layers/points/points.py
```diff
@@ -232,15 +305,65 @@ def __init__(
...
+        # set the edge color properties
+        self._edge_color_property = ''
+        if self._is_color_mapped(edge_color):
+            self._edge_color_mode = ColorMode.CYCLE
+        else:
+            self._edge_color_mode = ColorMode.DIRECT
+        if edge_color_cycle is None:
+            self._edge_color_cycle = cycle(
+                np.array([[1, 1, 0, 1], [0, 1, 0, 1]])
+            )
```
jni: Yellow green is the default color cycle? It's kinda terrible? LOL I would suggest:

- magenta/green
- black/yellow
- black/white
- ... Something else?

napari/layers/points/points.py
```diff
@@ -232,15 +305,65 @@ def __init__(
...
+        elif isinstance(edge_color_cycle, cycle):
+            self._edge_color_cycle = edge_color_cycle
+        else:
```
jni: Maybe add a comment here explaining what this catch-all clause is? It's actually the most common case, I would imagine...

Actually, I would say, would it be possible to modify `transform_color_cycle` so that you can unify this clause and the one above it, ie, if it's already a cycle, you just return it. That way, it becomes an easy case: if it's none, use this simple cycle, otherwise, use what was provided while making sure it's a cycle.

napari/layers/points/points.py
```diff
@@ -232,15 +305,65 @@ def __init__(
...
+        # set the face color properties
+        self._face_color_property = ''
+        if self._is_color_mapped(face_color):
+            self._face_color_mode = ColorMode.CYCLE
+        else:
+            self._face_color_mode = ColorMode.DIRECT
+        if face_color_cycle is None:
+            self._face_color_cycle = cycle(
+                np.array([[1, 1, 0, 1], [0, 1, 0, 1]])
+            )
```
jni: As above, I think this is a terrible default! Sorry. 😬

napari/layers/points/points.py
```diff
@@ -175,6 +254,13 @@ def __init__(
         self._data = np.asarray(data)
         self.dims.clip = False
 
+        # Save the properties
+        if properties is None:
+            properties = {}
+        elif not isinstance(properties, dict):
+            properties = dataframe_to_properties(properties)
```
jni, line 268: So, I'm not positive that *everything* will work, but most things should, if `dataframe_to_properties` is just `dict`. Then you get a dict of keys -> series, and boolean indexing still works fine:

```python
In [1]: df = pd.DataFrame({'hello': [1, 2], 'goodbye': [3, 4]})
In [2]: dict(df)
Out[2]:
{'hello': 0    1
 1    2
 Name: hello, dtype: int64, 'goodbye': 0    3
 1    4
 Name: goodbye, dtype: int64}

In [3]: d = dict(df)
In [4]: d['hello']
Out[4]:
0    1
1    2
Name: hello, dtype: int64

In [5]: type(d['hello'])
Out[5]: pandas.core.series.Series

In [6]: x = d['hello']
In [7]: x[[True, False]]
Out[7]:
0    1
Name: hello, dtype: int64
```

napari/layers/points/points.py
```diff
@@ -279,18 +405,100 @@ def data(self, data: np.ndarray):
...
+                    face_color_property_value = self.current_properties[
+                        self._face_color_property
+                    ][0]
+
+                    fc, _ = map_property(
+                        prop=face_color_property_value,
+                        colormap=self.face_colormap[1],
+                        contrast_limits=self._face_contrast_limits,
+                    )
+                    new_face_colors = np.tile(fc, (adding, 1))
```
jni, line 442: The repetition of edge color and face color code makes me think maybe we could write a function here that uses `getattr` to get the appropriate value. What do you think @kevinyamauchi?

napari/layers/points/points.py
```diff
@@ -279,18 +405,100 @@ def data(self, data: np.ndarray):
...
+    @properties.setter
+    def properties(self, properties: Dict[str, np.ndarray]):
+        if not isinstance(properties, dict):
+            properties = dataframe_to_properties(properties)
+        self._properties = self._validate_properties(properties)
+        if self._face_color_property and (
+            self._face_color_property not in self._properties
+        ):
+            self._face_color_property = ''
+            warnings.warn('property used for face_color dropped')
```
jni, line 466: this is very nice! Wouldn't have thought of this. =)

napari/layers/points/points.py
```diff
@@ -279,18 +405,100 @@ def data(self, data: np.ndarray):
...
+
+    def _validate_properties(self, properties: Dict[str, np.ndarray]):
+        """Validates the type and size of the properties"""
+        for v in properties.values():
+            if not isinstance(v, np.ndarray):
+                raise TypeError('all properties should numpy arrays')
```
jni: I think this check is too stringent. If someone wants to give us an ndarray like object here (e.g. a jax array), as long as everything works as expected, we should be ok with that. (`pandas.Series` is one relevant example).

napari/layers/points/points.py
```diff
@@ -380,6 +588,126 @@ def edge_width(self, edge_width: Union[None, float]) -> None:
...
+        self, edge_color_cycle: Union[list, np.ndarray, cycle]
+    ):
+        if isinstance(edge_color_cycle, cycle):
+            self._edge_color_cycle = edge_color_cycle
+        else:
+            self._edge_color_cycle = transform_color_cycle(
+                color_cycle=edge_color_cycle,
+                elem_name="edge_color_cycle",
+                default="white",
+            )
```
jni: Since this code is repeated, it would be nice to write a function for this (or, rather, just use this setter directly in the constructor!). If so, this setter probably needs to accept None (make the above type `Optional[]`).

napari/layers/points/points.py
```diff
@@ -380,6 +588,126 @@ def edge_width(self, edge_width: Union[None, float]) -> None:
...
+
+            Parameters
+            ----------
+            update_color_mapping : bool
+                If set to True, the function will recalculate the color cycle map or colormap
+                (whichever is being used). If set to False, the function will use the current color cycle map or
+                color map. For example, if you are adding/modifying points and want them to be colored with the same
+                mapping as the other points (i.e., the new points shouldn't affect the color cycle map or colormap),
+                set update_color_mapping=False. Default value is True.
+        """
```
jni: This docstring isn't properly formatted....

jni: Perhaps `refresh_edge_color` and `refresh_face_color` could be combined into a single function? At a minimum... It would be great if we could somehow avoid these functions, maybe using the `attrs` library or similar? I have no experience here, I just think it's not a great user experience.

Happy to punt on this, btw, as this PR is already great and insanely useful, but I just wanted to flag this as an issue. And I do think even in keeping the function(s) they can be simplified.

napari/layers/points/points.py
```diff
@@ -474,6 +1034,13 @@ def selected_data(self, selected_data):
             with self.block_update_properties():
                 self.current_size = size
 
+        properties = {
+            k: np.unique(v[index], axis=0) for k, v in self.properties.items()
+        }
+        n_unique_properties = np.array([len(v) for v in properties.values()])
+        if np.all(n_unique_properties == 1):
+            self.current_properties = properties
```
jni: This seems overly stringent... I suggest that we set the current properties to be the selected properties for every column where there is only one value, and ignore the other columns. ie something like:

```python
for n, prop in zip(n_unique_properties, properties):
    if n == 1:
        self.current_properties[prop] = properties[prop]
```

napari/layers/points/points_utils.py
```diff
@@ -0,0 +1,60 @@
+from typing import Dict, Tuple, Union
+
+import numpy as np
+from vispy.color.colormap import Colormap
+
+
+def dataframe_to_properties(dataframe) -> Dict[str, np.ndarray]:
+    """ Converts a dataframe to Points.properties formatted dictionary
```
jni:
```suggestion
    """Convert a dataframe to Points.properties formatted dictionary.
```

(Similar formatting suggestions apply to all docstrings. =)

napari/layers/points/points_utils.py
```diff
@@ -0,0 +1,60 @@
...
+
+    properties = {col: np.asarray(dataframe[col]) for col in dataframe}
+    return properties
+
+
+def guess_continuous(property: np.ndarray) -> bool:
+    """Guess if the property is continuous (return True) or categorical (return False)"""
+    # if the property is a floating type, guess continuous
+    if issubclass(property.dtype.type, np.floating):
+        return True
```
jni, line 30: There's probably some threshold number of integer values, let's say either 15 or 127, beyond which integer values can probably also be considered continuous... But I'm happy to ignore this for now...

jni (changes requested):
@kevinyamauchi spectacular work! I can't wait for this to be in.

I've left some comments, most of which should be straightforward to address. The main pickles are:

- do we want to allow using straight-up pandas data series as columns instead of ndarrays? I think everything should work for now but it might spell trouble down the track. But if it works now it might be nicer to not constrain things yet...
- I don't love `refresh_{face,edge}_color`... I don't have a good solution but for now at least they should be the same function, imho. It's not a lot of overhead to do both in one shot and it simplifies the code a fair bit I think, as well as the user experience.

Thanks!

napari/components/viewer_model.py
jni: Sorry, I forgot to respond here. I think conversion to our canonical form is the right move for now. The only question I have is whether it's necessary to directly convert to np.ndarray, rather than keeping the series in there.

examples/add_points_with_properties.py
kevinyamauchi: Fixed!

examples/nD_points_with_properties.py
kevinyamauchi: Fixed!

examples/nD_points_with_properties.py
kevinyamauchi: Fixed!

examples/nD_points_with_properties.py
kevinyamauchi: Yes and yes! I will add a comment.

examples/nD_points_with_properties.py
kevinyamauchi: Ah sorry, the number of the number of categories wasn't a factor of the number of points, so I had to trim the array. I changed the number of points, so it is simpler now.

napari/layers/points/points.py
kevinyamauchi: Haha, sorry. I changed it to magenta/green.

napari/layers/points/points.py
kevinyamauchi: Changed to magenta/green 😛

napari/layers/points/points.py
kevinyamauchi: That makes sense to me. I will remove the type check and leave the length check.

napari/layers/points/points.py
kevinyamauchi: > This docstring isn't properly formatted....

Fixed!

napari/layers/points/points.py
kevinyamauchi: I like it! I implemented your suggestion.

examples/add_points_with_properties.py
kevinyamauchi: I refactored the example as suggested. I will reply to your comments on the refresh methods below.

napari/layers/points/points_utils.py
kevinyamauchi: That's a good point. I will set it at 15 for now. It's just a guess and the user is free to switch back and forth between cycles and colormaps, so I don't think we need to get the threshold exactly right 😛

napari/layers/points/points.py
kevinyamauchi: Good point. I have modified `transform_color_cycle` to allow `cycle` objects to pass through.

napari/layers/points/points.py
kevinyamauchi: > Perhaps refresh_edge_color and refresh_face_color could be combined into a single function? At a minimum... It would be great if we could somehow avoid these functions, maybe using the attrs library or similar? I have no experience here, I just think it's not a great user experience.

I think this is a good point. I will respond to code duplication and attr type comments together at the bottom.

kevinyamauchi:
Thanks for the thorough review, @jni! I have implemented your more straight forward suggestions and responded to the larger comments below. Let me know what you think!

**Duplication between face and edge color parameters**
As you pointed out, there is a lot of duplication between the edge and face color parameters methods. Since we use a standard pattern, I think it would be fairly straightforward to write a single method that for each parameter that retrieves the appropriate variable via `getattr` and thus could be used by both the face and edge color parameter (e.g., `Points._set_color()` could be used in both the `Points.face_color` and `Points.edge_color` setters).

While this would reduce some of the code, I think one reason we may avoid using this type of pattern is that I think it is more brittle and harder to refactor (e.g., using Pycharm's renaming feature wouldn't work). That being said, the test coverage is pretty good for that area of the model, so I think we'd know pretty quickly if we introduced errors.

I propose that since there are a couple more parameters we would like to make settable via properties (e.g., size) and the reduction in code isn't too big, we hold off on using `getattr()` for now and revisit after it gets some use and feedback. However, we can reduce code now by switching to using the property setters in the `__init__()` method instead of performing the parsing, etc. with duplicate code.

**\*_color_refresh methods**
I agree that the color refresh methods are a bit clunky. I especially do not like that the user needs to call the function after altering `Points.properties` in a keybinding or plugin. I don't think it is an obvious pattern, so I made the example`add_points_with_properties.py` use that pattern so there is some documentation (I plan to add a tutorial as well following the tutorials website revamp). One alternative I had considered is to create a custom `Properties` object that would emit an event any time a property was changed, but that seemed like overkill. 

Another aspect is the code duplication between the edge and face color methods as described above. I think we can revisit writing methods that can serve both the face and edge color using `getattr()` in the future.

**Pandas series in the parameters dictionary**
I agree with your philosophy of holding off on constraining types as long as we can. As you mentioned, accessing/indexing a dict of Series can be the same as a dict of numpy arrays. However, when we need to add and remove properties (e.g., adding points, copy/paste), as I understand it, the datatypes diverge. I don't think we would be able to fully support a dict of Series without picking up `pandas` as a dependency.

I think our current method of converting the Dataframe into our dict of numpy arrays is a good middle ground for now, as it allows a lot of data types to be accepted without increasing our dependence on other libraries. However, I think we will want to revisit in the future, because I see a `Points` layer that can fully use `dask` arrays and dataframes for the coordinates, and properties, respectively as a great goal (e.g., remote data, lazy loading)!

**Summary**
In summary, I propose, we do the following:
- We switch to using the property setters in the `__init__()` method instead of performing the parsing, etc. with duplicate code.
- We refresh the face color and edge color in the same method (i.e., `refresh_face_color()` and `refresh_edge_color()` become `refresh_color`)
- Continue converting DataFrames into dicts of numpy arrays (i.e., do not support dicts of pandas `Series`)

jni:
@kevinyamauchi I approve of your plan. =) Thanks!

kevinyamauchi:
@jni , I have implemented your suggestions and I think this is ready for another round of review!

napari/layers/points/points.py
```diff
@@ -412,6 +793,162 @@ def current_face_color(self, face_color: ColorType) -> None:
...
+                If set to True, the function will recalculate the
+                color cycle map or colormap (whichever is being used).
+                If set to False, the function will use the current
+                color cycle map or color map. For example, if you
+                are adding/modifying points and want them to be
+                colored with the same mapping as the other
+                points (i.e., the new points shouldn't affect
+                the color cycle map or colormap), set update_color_mapping=False.
+                Default value is True.
+        """
```
jni, line 867: Incorrect docstring formatting again... :grimacing:

kevinyamauchi: Dang it! fixed.

jni (approved):

napari/layers/points/points.py
```diff
@@ -619,7 +619,7 @@ def edge_color_cycle(
 
     @property
     def edge_colormap(self):
-        """(str, vispy.color.Colormap) colormap to be applied to a property to set edge_color"""
+        """(str, vispy.color.Colormap): colormap to be applied to a property to set edge_color"""
```
jni: @kevinyamauchi I haven't seen any docstring convention documents deal with properties specifically, but I would follow the convention for methods, I think:

```python
"""Return the colormap to be applied to a property to get the edge color.

Returns
-------
colormap_name : str
    The name of the current colormap.
colormap : vispy.color.Colormap
    The actual colormap object.
"""

kevinyamauchi: I looked and couldn't find any either. I will update them to follow the convention for methods.

kevinyamauchi: To avoid departing from the style of the other layers, I changed the docstrings that didn't work as one-liners and left the rest as-is.

kevinyamauchi:
It looks like we have two approvals, so I will let this sit until morning and then merge. Thanks, everyone!

PR #897, opened 2020-01-22
refactor add_* methods

### LINKED ISSUE #892: Refactor add_* methods in ViewerModel into mixin class
## 🚀 Refactor

I'd like us to consider refactoring the `add_*` methods currently inside the [ViewerModel file](https://github.com/napari/napari/blob/master/napari/components/viewer_model.py#L393-L1001) into a mixin class `AddLayersMixin` in it's own file `add_layers.py`

The reason for this is I currently find the `ViewerModel` file very large and hard to find things in, and often we just have to update either the `add_*` methods or the actual `ViewerModel` stuff and they are a little different in nature. This organization would then mirror the existence of the [`view_layers.py` file](https://github.com/napari/napari/blob/master/napari/view_layers.py) and make things much easier to parse.

Note, that I'd leave the [`add_layer` method](https://github.com/napari/napari/blob/a10a88b2115160211863b47ee7d3012fc8c7beff/napari/components/viewer_model.py#L368) layer inside the main `ViewerModel` file, though I might make it private `_add_layer` as I don't imagine users explicitly making their own layers and adding them in this way.

Aside from that minor change the rest of this would be a completely backwards compatible refactor and only aid our internal organization of the codebase.

What do you think @jni @kne42?

sofroniewn:
# Description
This PR closes #892 by pulling out `add_*` methods out of the viewer_model file directly and putting them in their own mixin class. This is done purely to improve readability and maintainability. I did not make the `add_layer` method private - we can do that in a future PR if we want, so this PR is purely an internal reactor.

@kne42 I think the doc strings are working ok, but can you check them too? I know we use this pattern elsewhere in the codebase (like in the layers themselves right?) so it would be good to know that all the doc strings there would be ok too. Thanks.

## Type of change


kevinyamauchi (approved):
Thanks, @sofroniewn . This looks good to me! As far as I can tell, the docstrings work on my machine, but as you said, @kne42 is probably the best person to take a look :)

kne42 (approved):
No, it does not show inherited methods in the sphinx documentation. However, this was already the case because `Viewer` inherits from `ViewerModel`. Must be some setting somewhere...

Approved. One thing you should do though is change the viewer layer parity test to point to the new mixin class.

sofroniewn:
> No, it does not show inherited methods in the sphinx documentation. However, this was already the case because Viewer inherits from ViewerModel. Must be some setting somewhere...

Ah, I was maybe just looking at the doc string in the console. Is there something that needs to be done in this PR or can it be fixed elsewhere / later?

> One thing you should do though is change the viewer layer parity test to point to the new mixin class.

I'm not quite sure what you're recommending. Right now the import looks like this https://github.com/napari/napari/blob/a10a88b2115160211863b47ee7d3012fc8c7beff/napari/tests/test_viewer_layer_parity.py#L12
which I think is on the whole `Viewer` not even `ViewerModel`. How do you recommend we change it? Or should the change be somewhere else?

kne42:
> Is there something that needs to be done in this PR or can it be fixed elsewhere / later?

It should be fixed elsewhere.

> which I think is on the whole Viewer not even ViewerModel. How do you recommend we change it? Or should the change be somewhere else?

Instead of testing against `Viewer`, we should be testing against `AddLayersMixin`

jni:
@kne42 

> Instead of testing against Viewer, we should be testing against AddLayersMixin

I'm not positive about this. Ultimately, what we care about is the public-facing class, so that's what we should test. *How* the class gets those docstrings/methods is an implementation detail, and as such does not need to be tested. I think?

jni (approved):

kne42:
@jni I disagree. Since the viewer now inherits that class, it means that those methods were not actually defined in it. We should be testing against the class that the methods were defined in.

@sofroniewn asked me offline what exactly I wanted. Basically
```python
from napari import layers as module
from napari.components.add_layers_mixin import AddLayersMixin
```

and then replace all instances of `Viewer` with `AddLayersMixin`

I would also move the test to be under `napari/components/tests`

and regardless of whether you choose to do above, add a comment explaining that the reason why we copy/pasted everything instead of autogenerating it was because IDEs like PyCharm parse the source code for docs instead of pulling it up dynamically

jni:
@kne42 

> We should be testing against the class that the methods were defined in.

This isn't obvious to me? Users don't care where things are defined, they care that a thing works. If we test AddLayerMixin directly, and someone removes that inheritance from the ViewerModel, all those tests will pass but that functionality will be broken! At the very least, we should test *both*...

> add a comment explaining

This is a good idea. =) Your original dynamic generation was super cool! Stupid PyCharm... =P

kne42:
> Users don't care where things are defined, they care that a thing works.

That's true.

I won't die on this hill, so @sofroniewn I would say add the aforementioned comment and then merge unless anyone else has objections?

sofroniewn:
I will add the requested comment and then I think this is ready for merge

sofroniewn:
I have updated this PR with the comment requested by @kne42. Can @kne42 @jni check if it is good. I wanted to give us enough flexibility in case we start adding methods like `add_tracks` which are built on top of things like `Shapes`.

I've also left the test where it is as it involve both the Viewer and the layers. If this looks good then I'll merge.

jni:
@sofroniewn just a minor typo to fix, then all ok by me!

sofroniewn:
I have fixed the typo. I will wait for @kne42 to give it the ok and then merge

sofroniewn:
I am going to merge now - test failures are unrelated to this PR. We can improve the comments in follow up PRs if needed

Issue #702, opened 2019-11-17
Inconsistent behaviour when adding points (some options affect all points in layer, some don't)


VolkerH:
## 🐛 Bug

Not sure whether the following is a bug. Maybe it is just behaviour as envisaged, then it is inconsistenst interaction.

When adding markers in a points layer, changing the symbol (e.g. from "disk" to "diamond") affects all points in the current points layer (including exisiting ones). 
When changing the marker color or size, existing markers stay the same subsequently added markers are affected.

So it is possible to create multiple markers with different colors and sizes in a single point layer, but it is not possible to create markers with different symbols. 
![image](https://user-images.githubusercontent.com/2210044/69011663-30c86900-09c1-11ea-8530-1ba86555f527.png)

## To Reproduce

1. Add a points layer
1. add a marker as "disk"
2. change marker type to "diamond"
3. add another marker

## Expected behavior

I would expect that marker symbols in a point layer behave like marker color and size, i.e.
that I can create markers with different symbols in a single point layer.

## Environment

napari, 0.2.5+12.g15deef0
Qt, 5.13.2
NumPy, 1.17.3
SciPy, 1.3.1
VisPy, 0.6.2
scikit-image, 0.16.2

Platform: Windows-10-10.0.18362-SP0
Python:   3.7.5 (default, Oct 31 2019, 15:18:51) [MSC v.1916 64 bit (AMD64)]
Backend:  PySide2
GL version:  '4.5.0 - Build 25.20.100.6472'
MAX_TEXTURE_SIZE: 16384

 - Any other relevant information:

sofroniewn:
Ah yes - so this is a known issue that should at least be documented if not exposed better in the UI. Right now `symbol` and `opacity` in the points layer are global properties, while we could make opacity apply for each point (potentially at the cost of increasing latency), making symbol apply to each point would require some changes at the vispy level, and I'm not sure what performance hits we might take. In the mean time we should at least document this better as I have no doubt that it is confusing to people

VolkerH:
Opacity (and blending mode) I didn't test as  intuitively I would always think of  opacity as a layer property, which would be consistent with the other layer types (same for blending mode BTW). 

Now that you mention it, this could of course also be a per-point property. However, in that case you would lose the ability to dial back the opacity for the whole layer.

My suggestion is to decide which properties affect a layer and which properties affect a point and separate this out in the GUI. Maybe a vertical alignment.

```
point properties
------ <hline> ------
layer properties
````

[...]

PR #898, opened 2020-01-23
[WIP] Refactor color repr for Shapes layer

HagaiHargil:
# Description

(Following #782)

This PR changes the internal representation of colors in the Shapes and Vectors layers to match the one in Points - i.e. an Nx4 numpy array. Once it's merged `napari` should be using Nx4 RGBA arrays wherever its dealing with colors.

## Type of change

# How has this been tested?

## Final checklist:


HagaiHargil:
I still have a long way before merging, but I already wanted to point out a couple of changes which might have to be API-breaking:

1. `Points` layer uses `Points.face_color` for the Nx4 array and `Points.current_face_color` for the last chosen color(s). `Shapes` doesn't have that attribute. I think they should match after this PR, but it's obviously a breaking change.

2. `Shapes` has an `Shapes.opacity` attribute which is redundant once we always set the transparency of the layer when setting the edge or face colors. Should this PR remove (or deprecate) this API? I believe it should, again for the sake consistency across the different layers.

napari/layers/shapes/shape_models/shape.py
```diff
@@ -1,33 +1,38 @@
...
-        If string can be any color name recognized by vispy or hex value if
-        starting with `#`. If array-like must be 1-dimensional array with 3 or
-        4 elements.
-    face_color : str | tuple
-        If string can be any color name recognized by vispy or hex value if
-        starting with `#`. If array-like must be 1-dimensional array with 3 or
-        4 elements.
+    edge_color : str or array-like
+        Color of the shape marker border. Strings can be color names or hex
+        values (starting with `#`). Array-like should have 3 or 4 elements
```
sofroniewn: Shape color has always been global per shape - i.e. all the edges are "red" this now makes it sound like you could give a different color to each edge, I'm not sure we want to support that from an API complexity standpoint, though we certainly could. Was that change intentional?

Particularly I'm not sure if: `If array-like must be 1-dimensional array` -> `Array-like should have 3 or 4 elements per row` is a change we want to make

sofroniewn:
> Points layer uses Points.face_color for the Nx4 array and Points.current_face_color for the last chosen color(s). Shapes doesn't have that attribute. 

Shapes has those attributes here https://github.com/napari/napari/blob/master/napari/layers/shapes/shapes.py#L426 and https://github.com/napari/napari/blob/master/napari/layers/shapes/shapes.py#L382, and they are intended to be the same as in Points. Maybe I don't quite understand what you mean, can you explain more?

> Shapes has an Shapes.opacity attribute which is redundant once we always set the transparency of the layer when setting the edge or face colors. Should this PR remove (or deprecate) this API? I believe it should, again for the sake consistency across the different layers.

Probably yes. Opacity should then just behave the same as for Points, and the opacity slider should just be global. This has come up here https://github.com/napari/napari/issues/702#issuecomment-555106044 where I think the feeling was to remove the special handling of opacity for different shapes

Otherwise seems like you're on the right track

HagaiHargil:
@sofroniewn Thanks for the very helpful comments. I had the wrong mental model of the way the Shapes layer works internally, which lead to me asking pretty stupid questions, but your comments helped me understand what's going on.

I believe I'm now on the right track towards complete the first part of this PR.

napari/layers/shapes/shape_models/shape.py
```diff
@@ -1,33 +1,38 @@
...
-        starting with `#`. If array-like must be 1-dimensional array with 3 or
-        4 elements.
+    edge_color : str or array-like
+        A single color for all of the shape marker borders. A string can be
+        the color name or its hex values (starting with `#`). Array-like should
+        have 3 or 4 elements in a single row (can be 2D) in a RGB(A) format.
+    face_color : str or array-like
+        Color of the shape marker body. A string can be the color name or its
+        hex value (starting with `#`). Array-like should have 3 or 4 elements
+        in a single row (can be 2D) in a RGB(A) format.
```
HagaiHargil: @sofroniewn Do you like this phrasing better than the previous one?

napari/layers/shapes/shape_models/shape.py
```diff
@@ -208,39 +213,44 @@ def edge_width(self, edge_width):
...
+        """Change the edge color of the shape.
+
+        This method assumes that the incoming data passed through
+        transformation (``transform_color_with_defaults``) and
+        normalization (``normalize_and_broadcast_colors``) which
+        means it's already an 1x4 numpy array.
+        """
+        self._edge_color = edge_color
+        hexval = rgb_to_hex(self._edge_color)
+        self._edge_color_name = hex_to_name.get(hexval, hexval)
```
HagaiHargil: Since this class will not be instatiated by a user I allowed it to set the internal `_edge_color` attribute without any further checks. All checks happen upstream in the child classes.

napari/layers/shapes/shape_models/shape.py
sofroniewn: Much better, thanks! You can propagate that to the rest of the individual shapes

HagaiHargil:
@sofroniewn One more question I had was regarding the different `Rectangle`, `Ellipse` shapes instatiation mechanisms. 

Currently, as seen [here](https://github.com/napari/napari/blob/fa122dc87e10019b50bf0dfac4d273561b194ad5/napari/layers/shapes/shape_models/rectangle.py#L48_L56) for example, instatiating a Rectangle is pretty much instatiating a Shape with a unique name. However, while the Shape ABC is generally not user-facing, I believe that a Rectangle is. This raises the issue of data validation, since I generally don't trust the user to supply the color data in the exact format that the underlying Shape class expects, i.e. a 2D RGBA. Luckily we've already written these validation\transformation\normalization functions when we dealt with the Points layer - users creating Points can supply their data in a variety of formats, and the internal `__init__` should be able to convert them to our desired RGBA array. 

In principle I could simply add these validation functions to every `__init__` of every shape. This solution will work for 100% of the cases, but it's also extremely redundant, as in most cases Rectangles, Ellipses and such are created using the `add_shape(name='rectangle')` method, which instatiates Rectangles internally, and I've already added the validation step to this function. Hence when users create their shapes using the preferred (and I'm assuming common) approach of `add_shape()`, their data will be validated twice - once in `add_shape()`, and once during `__init__` of the the shape they requested.

One solution for this issue is to move the data validation step into the different classes, and remove it from the user facing functions such as `add_shape()`. This is pretty straight-forward, although it will probably force me to write a lot of duplicate code.

A different solution I had in mind was to make a new class method for each of the shapes termed `from_raw_data()`, which validates the incoming data and then instatiates the underlying class. Users who wish to manually create Rectangles (e.g plugin authors or other curious naparians [is there such a thing? :thinking:]) will be pointed towards the `from_raw_data()` class method, while the rest of the napari codebase can still instantiate Rectangles as usual provided that the data has undergone validation before hand. This might be less ergonomic than the previous method, but its explicitness might be valued on its own.

A third solution is to simply declare that Rectangles, Ellipses and such aren't user-facing, and document the fact that they need their data validated before instatiation.

What do you think would work best?

sofroniewn:
@HagaiHargil I don't understand why we can't just have the validation done in the setter of `edge_color` and `face_color` on the `Shape` class, which is the base class all the other shapes like `Ellipse`, `Rectangle` etc. inherit from. See https://github.com/napari/napari/blob/fa122dc87e10019b50bf0dfac4d273561b194ad5/napari/layers/shapes/shape_models/shape.py#L216 and https://github.com/napari/napari/blob/fa122dc87e10019b50bf0dfac4d273561b194ad5/napari/layers/shapes/shape_models/shape.py#L231.

Having the validation here ensures it will always work, because this code will always get called right? Or am I missing something?

sofroniewn:
@HagaiHargil I was wondering if you were able to pick this up again or if there are things blocking you that I can help with.

I think it would be nice to get in so we can finish standardizing some of our color stuff and get things like the color picker @tlambert03 implemented in #832 used more uniformly.

Thanks again for starting the contribution!!

sofroniewn:
I you can pick this up again, you also might want to merge in all the latest changes from master so you don't run into any merge conflicts down the road!!

HagaiHargil:
Hi @sofroniewn, sorry for the delay, I had a busy month in the office. I actually did work on this PR this week - I merged master and tried to implement the suggestion you made here https://github.com/napari/napari/pull/898#issuecomment-583007901, but got stuck trying to fix the initialization order of the different color attributes.

I noticed that this PR has started blocking a couple of new features, I'll make an effort to show considerable progress this week.

napari/layers/shapes/shapes.py
```diff
@@ -684,10 +681,20 @@ def add(
...
+
+            # If we just instantiated the layer then we can directly set
+            # new face and edge colors. Else we'll append to the existing
+            # ones the new values.
+            if len(self._data_view) == num_of_shapes:
+                self._face_color = new_face_colors
+                self._edge_color = new_edge_colors
+            else:
+                self._face_color = np.concatenate((self.face_color, new_face_colors), axis=0)
+                self._edge_color = np.concatenate((self.edge_color, new_edge_colors), axis=0)
```
HagaiHargil: @sofroniewn The current pain point I'm having is due to the interaction of the `Shapes` layer with the `ShapeList` object. I'm currently holding two 'ground truths' regarding the true values of face and edge color: The first is in each of the shapes themselves, as being kept in the `ShapeList`, while the other is in the `Shapes` layer (seen in this code snippet). As you can see - there's a `self._data_view.add(shape)` method (`ShapeList`) and the rest of the code here deals with the `Shapes` layer colors. If this was the only place that changes the underlying `ShapeList` then I wouldn't have cared too much, but I've discovered a few other places, such as the `paste` methods, which change the `ShapeList` but leave other properties (colors as well as opacity) unchanged, which is an issue.

Ideally I'd like to have one ground truth, and I believe that the better option is to only hold the color and opacity data in the `ShapeList` by modifying its existing `face_color`, `edge_color` and `opacity` attributes. When the `Shapes` layers needs this data it will simply ask the `ShapeList` for its colors and opacity.

While this change seems reasonable to me, I don't want to implement it before hearing from you - I'm afraid I'm breaking some model-view-controller-controllerview-modelcontroller-viewmodel rules which I honestly have very little experience with :)

sofroniewn: > I believe that the better option is to only hold the color and opacity data in the ShapeList by modifying its existing face_color, edge_color and opacity attributes. When the Shapes layers needs this data it will simply ask the ShapeList for its colors and opacity.

This sounds good to me too - the guiding philosophy of only having one source of ground truth is a good one. Here we have this distinction between `ShapeList` and the main `Shapes` layer, that hopefully can the code more modular, but maybe introduces additional complexities like what you mentioned above.

HagaiHargil:
Hi again @sofroniewn,

I merged your re-naming PR and also pushed my work from last week. I did make progress, but not too significant, since I'm missing a key point - how does one change an individual's shape edge or face color from the command line? What is the suggested API for that? I know it can be done from the GUI, although due to some issues with 0.2.12 I can't seem to create shapes using the GUI alone, so it's difficult for me to follow how is it done via the GUI.

Once I have an answer to this question I'll be able to take another step forward. Currently the data for each shape's colors should be stored inside that shape alone. The over-arching `ShapeList` will have a list of face_colors which contains a view to the color of each shape, and the `Shapes` layer will show a view of that view. The question I raised earlier came up when I tried to write the new setters for each of the objects.

sofroniewn:
> I know it can be done from the GUI, although due to some issues with 0.2.12 I can't seem to create shapes using the GUI alone, so it's difficult for me to follow how is it done via the GUI.

Uh oh, that doesn't sound so great. Can you maybe elaborate more on that one. Shapes drawing works fine on master and the `0.2.12` release. Would be good to understand what's broken.

> how does one change an individual's shape edge or face color from the command line? What is the suggested API for that? 

Right now you to change the `face_color` of the 1st and 2nd shape you'd have to do
```python
layer.selected_data = [1, 2]
layer.face_color = 'red'
```
(Though I just noticed @kevinyamauchi that this current approach fails on points, at least with `face_color_mode`) We probably want to fix that, I will investigate more / make a dedicated issue.

We can improve that API in the future / make it a one liner, but for now we should make sure at least that approach works

kevinyamauchi:
@sofroniewn, I believe, the current pattern for changing the color of selected points is to first select points and then change the value in `Points.current_*`. This is what is happening behind the scenes in the GUI as well. Would something similar work here? In any case, I think we should try to be consistent between points and shapes.

sofroniewn:
@HagaiHargil hope you're doing well (all things considered in the world right now).

We're planning to make our `0.3.0` release soon and it would be great to have this PR in for that release. I've also made a change in #1051 (making `selected_data` a `set` instead of `list`) that might cause conflicts here after it was merged, so it might be nice to get this one in first.

I was wondering if you could give us a status update on this PR, and maybe an estimate of when it will be ready for review? Also no worries if now is a busy time for you and you can't get to this for a while, then maybe I or @kevinyamauchi might take it over and finish it up. Thanks!!

HagaiHargil:
Hi @sofroniewn, I pushed my current version.

I'm sorry that it took me so long, I tried to prioritize it higher but honestly I had more important stuff to do recently, and every time I tried to get back to it the context switch was pretty costly.

I guess I'm passing it over to you guys, and when I'll have more time I'll try to get back to it, if it's still relevant. Feel free to discard my work if it's not good or clear enough.

Sorry again.

sofroniewn:
> I'm sorry that it took me so long, I tried to prioritize it higher but honestly I had more important stuff to do recently, and every time I tried to get back to it the context switch was pretty costly.

No worries @HagaiHargil you've been a valuable contributor, and this PR definitely pushes in a direction we want to go. @kevinyamauchi and I will look into what it takes to finish it up.

Do keep raising issues on the repo if you encounter problems or have new ideas for features, and PRs are always still welcome. Good luck with the rest of your work!!

kevinyamauchi:
Thank you, @HagaiHargil, for all of the work thus far! I am more than happy to take the baton on this one. As @sofroniewn said, I hope you will continue to make issues/feature requests and contributions when you have time.

In terms of getting this to the finish line, I think I will break this up into two PRs: one for `Vectors` and one for `Shapes`. Since it looks like so far, the work in this PR is  mostly on `Shapes`, I will open a second PR for `Vectors`. I think `Vectors` will be the more straightforward one, so I will start there.

Issue #901, opened 2020-01-23
Selecting points in 3D


HagaiHargil:
## 🐛 Bug

When I select a 3D point while `ndisplay == 2` and then try to change the viewer to a 3D one napari crashes.

## To Reproduce
```python
with napari.gui_qt():
    v = napari.view_points(np.array([[1, 0, 0], [0, 0, 1]]))
```
Then select a point with the cursor and try to change the display to a 3D one.

## Expected behavior

I expect napari to transition to a 3D view of the data without crashing. When I come back to the 2D view it would be nice if the points would still be highlighted. This is not currently the case with 2D points, though.

## Environment

Python 3.7, napari 0.2.10

## Additional context
```
WARNING: Traceback (most recent call last):
  File "/home/hagai/.conda/envs/ncd/lib/python3.7/site-packages/napari/_qt/qt_viewer_buttons.py", line 173, in <lambda>
    lambda state=self: self.change_ndisplay(state)
  File "/home/hagai/.conda/envs/ncd/lib/python3.7/site-packages/napari/_qt/qt_viewer_buttons.py", line 178, in change_ndisplay
    self.viewer.dims.ndisplay = 3
  File "/home/hagai/.conda/envs/ncd/lib/python3.7/site-packages/napari/components/dims.py", line 296, in ndisplay
    self.events.ndisplay()
  File "/home/hagai/.conda/envs/ncd/lib/python3.7/site-packages/napari/utils/event.py", line 508, in __call__
    self._invoke_callback(cb, event)
  File "/home/hagai/.conda/envs/ncd/lib/python3.7/site-packages/napari/utils/event.py", line 529, in _invoke_callback
    cb_event=(cb, event),
  File "/home/hagai/.conda/envs/ncd/lib/python3.7/site-packages/napari/utils/event.py", line 523, in _invoke_callback
    cb(event)
  File "/home/hagai/.conda/envs/ncd/lib/python3.7/site-packages/napari/components/viewer_model.py", line 88, in <lambda>
    self.dims.events.ndisplay.connect(lambda e: self._update_layers())
  File "/home/hagai/.conda/envs/ncd/lib/python3.7/site-packages/napari/components/viewer_model.py", line 1037, in _update_layers
    layer.dims.ndisplay = self.dims.ndisplay
  File "/home/hagai/.conda/envs/ncd/lib/python3.7/site-packages/napari/components/dims.py", line 296, in ndisplay
    self.events.ndisplay()
[...]

PR #902, opened 2020-01-23
Fix points selection

kevinyamauchi:
# Description
This PR fixes the `Points` layer bug where switching from 2D to 3D with a point selected causes a crash.

## Type of change
- [x] Bug-fix (non-breaking change which fixes an issue)expected)

# References
This addresses the bug described by @HagaiHargil in #901, but there are some nice feature suggestions in that issue we should follow up on.

# How has this been tested?
- [x] added a test in `test_points.py`

## Final checklist:

### DIFF
--- a/napari/layers/points/tests/test_points.py
+++ b/napari/layers/points/tests/test_points.py
@@ -105,8 +105,13 @@ def test_selecting_points():
     np.random.seed(0)
     data = 20 * np.random.random(shape)
     layer = Points(data)
+    layer.mode = 'select'
     layer.selected_data = [0, 1]
     assert layer.selected_data == [0, 1]
+
+    # test switching to 3D
+    layer.dims.ndisplay = 3
+    assert layer.selected_data == []
 
 
 def test_adding_points():
--- a/napari/layers/points/points.py
+++ b/napari/layers/points/points.py
@@ -667,9 +667,10 @@ class Points(Layer):
                 ind = list(self._indices_view).index(c)
                 selected.append(ind)
         self._selected_view = selected
-        if len(selected) == 0:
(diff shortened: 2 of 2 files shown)


napari/layers/points/tests/test_points.py
```diff
@@ -105,9 +105,14 @@ def test_selecting_points():
     np.random.seed(0)
     data = 20 * np.random.random(shape)
     layer = Points(data)
+    layer.mode = 'select'
     layer.selected_data = [0, 1]
     assert layer.selected_data == [0, 1]
 
+    # test switching to 3D
+    layer.dims.ndisplay = 3
+    assert layer.selected_data == []
```
jni, line 114: @kevinyamauchi I don't understand why the selected data is lost, rather than just the box not being drawn?

kevinyamauchi: It's because the way our modes are currently implemented, the selected points are cleared when switching from the `select` to another mode and when going from 2D to 3D , the mode is automatically switched to the `pan_zoom` mode (since selection isn't supported in 3D for now).

That being said, it's fairly straightforward to allow point selection in 3D from the command line/API (as requested in #901). I just thought would be better to make a small PR here to fix the bug that causes a crash and then make the larger PR as a follow-up.

sofroniewn (approved):
LGTM! Thanks, feel free to merge at your convenience @kevinyamauchi

napari/layers/points/tests/test_points.py
sofroniewn: approach sounds fine for now, but agreed a larger followup sounds good too

PR #904, opened 2020-01-23
[WIP] Add pop-ups to all sliders in our control widget

AhmetCanSolak:
# Description

This PR will bring pop-ups to all sliders in our control widget. Just a placeholder PR for now.

## Type of change

# References

# How has this been tested?

- [X] example: all tests pass with my change

## Final checklist:

### DIFF
--- a/napari/_qt/qt_qslider_popup.py
+++ b/napari/_qt/qt_qslider_popup.py
@@ -1,0 +1,148 @@
+from qtpy.QtCore import Qt, QPoint
+from qtpy.QtGui import QDoubleValidator, QFontMetrics, QFont
+from qtpy.QtWidgets import QHBoxLayout, QLineEdit, QSlider
+
+from .qt_modal import QtPopup
+from .utils import qt_signals_blocked
+
+
+class LabelEdit(QLineEdit):
+    def __init__(self, value='', parent=None, get_pos=None):
+        """Helper class to position LineEdits above the slider handle
+
+        Parameters
+        ----------
+        value : str, optional
+            starting value, by default ''
+        parent : QRangeSliderPopup, optional
+            required for proper label positioning above handle, by default None
+        get_pos : callable, optional
+            function that returns the position of the appropriate slider handle
+            by default None
+        """
+        super().__init__(value, parent=parent)
+        self.fm = QFontMetrics(QFont("", 0))
+        self.setObjectName('slice_label')
+        self.min_width = 30
+        self.max_width = 200
+        self.setCursor(Qt.IBeamCursor)
+        self.setValidator(QDoubleValidator())
+        self.textChanged.connect(self._on_text_changed)
+        self._on_text_changed(value)
+
+        self.get_pos = get_pos
+        if parent is not None:
+            self.min_width = 50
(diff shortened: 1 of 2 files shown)


napari/_qt/qt_qslider_popup.py
```diff
@@ -0,0 +1,129 @@
...
+            QSlider(Qt.Horizontal) if horizontal else QSlider(Qt.Vertical)
+        )
+        self.slider.setMinimum(2)
+        self.slider.setMaximum(200)
+        self.slider.setSingleStep(1)
+        self.slider.setValue(100)
+        self.slider.setMinimumHeight(18)
+        self.slider.setFocus()
+        self.slider.valueChanged.connect(self._on_value_change)
+        # self.slider.resized.connect(self._update_cur_label_positions) # no resized signal yet
```
tlambert03, line 90: rather than creating a resized signal like was originally done on the rangeslider, you can just override the [QWidget `resizeEvent()`](https://doc.qt.io/qt-5/qwidget.html#resizeEvent) method and put the necessary logic there.  (note, the window already has its new size when the event is called)

napari/_qt/qt_qslider_popup.py
```diff
@@ -0,0 +1,129 @@
...
+    def _update_cur_label_positions(self):
+        self.cval_label.update_position()
+
+    def _on_value_change(self, value):
+        with qt_signals_blocked(self.slider):
+            self.cval_label.setText(str(float(value) / 100.0))
+            self._update_cur_label_positions()
+
+    def _curval_label_changed(self):
+        cval = int(self.curmin_label.text())  # TODO: check int casting here
```
tlambert03, line 120: for maximum reusability, it might be worth considering implementing self.slider as a "double" slider behind the scenes, so that it can be used stuff like gamma as well.  (Then you would stick with casting to `float` in `_label_changed`, and use `_numformat` to make integer sliders "look" like integer sliders).

<details>
<summary>abbreviated QDoubleSlider Example</summary>

```python
class QDoubleSlider(QSlider):
    """A Slider Widget that can handle float values."""

    PRECISION = 1000

    def __init__(self, parent=None):
        super().__init__(Qt.Horizontal, parent=parent)
        self.setMaximum(10)

    def value(self):
        """Get the slider value and convert to float."""
        return super().value() / self.PRECISION

    def setValue(self, value):
        """Set integer slider position from float ``value``."""
        super().setValue(value * self.PRECISION)

    def setMaximum(self, value):
        """Set maximum position of slider in float units."""
        super().setMaximum(value * self.PRECISION)
```
</details>

AhmetCanSolak: i agree, i was thinking to do same

AhmetCanSolak: i agree, i was thinking to do same

sofroniewn:
hi @AhmetCanSolak its been a while since this was opened. i think we're probably going to take a new approach to this with some of the slider work @ppwadhwa and @tlambert03 have been doing recently, is it alright if we close? Thanks for kicking this off though

Issue #772, opened 2019-12-07
Add basic interactivity to the base layer


sofroniewn:
## 🚀 Feature
I'd like to add basic interactivity to the base layer so that it is available for all layers to use. Not all layers would necessarily have to expose all this functionality by default, but I think it would be good to have there for plugin developers to access if they need and to reduce duplication across our code base.

I am imagining having the concept of "pan/zoom mode", "select mode", "direct select", and "lasso mode" as at least 3 modes we'd like to support. Currently "pan/zoom" features on "Shapes", "Points", and "Labels", "select" on "Shapes", "Labels", "Points" (though on "Labels" it isn't fully functional), and "lasso" doesn't exist on any, but has been requested on "labels" - see #582.

- In "pan/zoom" mode you'd just be able to pan/zoom as normal.

- In "select mode" a single click would trigger a select callback that could be customized per layer with the coordinate of the selection, and dragging would trigger that callback but with a bounding box around the selection.

- "direct select" would be the same as "select" but with a different icon an callback

- "lasso" would allow you do draw a freehand lasso and then trigger a callback with the coordinates of the lasso.

Ideally things like "paintbrush" and "fill bucket" would also be able to reuse components of these things too. Same with some of the more advanced shapes mouse interactivity. Maybe this idea is still a little rough around the edges, but I think it will make make things a lot simpler

I'd like to wait until after #694 is done, though I think this scheme will also dramatically simplify

I'm also not quite sure about how to think about this in relation to #600 and #651 and whether I should wait for more work there. This might also make #651 simpler as again more code will have been moved to base and will be reused.

What do @kne42 @kevinyamauchi and @jni think of this?

sofroniewn:
Thinking about this more, in think I will pull the basic highlighting model code into the base layer and then pull the highlighting vispy code into its own node which will reduct the total number of visuals we have, reduce code complexity, and improve performance. I'm going to start working on this now. Although it will be a large refactor it should be backwards compatible from an API standpoint

PR #907, opened 2020-01-24
Allow 3D point selection via API

### LINKED ISSUE #901: Selecting points in 3D

kevinyamauchi:
# Description
This PR is implements @HagaiHargil 's suggestions in #901 to:

- allow points to be selected in 3D via the command line / API
- make point selections persistent (i.e., selected points will remain selected when change modes, even if they are not highlighted)

## Type of change

# References
This PR fixes #901 

# How has this been tested?
- [x] I added selection tests to `test_points.py`

## Final checklist:


kevinyamauchi:
@HagaiHargil , is this what you had in mind?

points_2 2020-01-24 12_54_10

HagaiHargil:
This seems great, thanks!

sofroniewn (dismissed):
> make point selections persistent (i.e., selected points will remain selected when change modes, even if they are not highlighted)

@kevinyamauchi - I find this behavior somewhat counter-intuitive - one of the ideas of select mode was that was the mode where points were editable and if you then went to pan zoom mode they wouldn't be. I find it very strange to be in pan zoom mode, see no highlighting but have a the subset of points that I previously had selected change when I adjust the size slider or face color. Now maybe those GUI elements shouldn't be adjustable in pan/zoom mode, but then in add mode we'd have to unselect things too.

I also find it strange when I am in 3D viewing in select mode and I can't see which points I have selected and then adjust some properties and some change (we can work on this in separate PRs)

Another option would be to use the `_selected_data_stored` property to store the selected points when transitioning to 3D / pan-zoom etc. which we could then refresh the selected property with when transitioning back to selected mode. We already use this approach when the user presses the `spacebar` to temporarily enable pan/zoom - we'd now want that to be more general.

This approach would enable the requested functionality, but with minimal changes to the existing functionality (i.e. when not in select mode nothing would be selected). What do you think? 

Note also we have a `_selected_data_history` property that never gets used and should be deleted, but that used to do this

HagaiHargil:
> I also find it strange when I am in 3D viewing in select mode and I can't see which points I have selected and then adjust some properties and some change (we can work on this in separate PRs)

I agree that this is weird, but besides this behavior I actually think the opposite - as someone who didn't know that changing to pan\zoom should deselect the points, it seemed very counter-intuitive for me at first. I see this as "highlighting" a few special points among a 3D cloud of non-interesting points.

For example, I might be looking for all points in a cloud with a distance `d < 10` from some main point. I'd then color all the `d < 10` points in a single color, perhaps change their size, and look at it in 3D. But once I'm looking at the 3D view I will probably want to continue changing the way I'm visualizing my points, i.e. change their color, size, etc. If they were de-selected when I transitioned into 3D I'd find _that_ behavior to be counter-intuitive.

sofroniewn:
> But once I'm looking at the 3D view I will probably want to continue changing the way I'm visualizing my points, i.e. change their color, size, etc. If they were de-selected when I transitioned into 3D I'd find that behavior to be counter-intuitive.

Sorry I wasn't clear, I agree with the 2D / 3D transition in select mode ultimately shouldn't unselect points.  I guess right now I consider that an "unsupported" / "needs work" feature because we need to get the bounding box working in 3D (which shouldn't be too hard). At that point I think we will have no problem having the 2D to 3D transition keep selection, but until we have the bounding box working in 3D I think it's  simpler to just make them not selected in 3D. @kevinyamauchi do you also want to work on a 3D bounding box at some point in time? You'll just need to find the full extents of the selected points in 3D not just 2D and then adjust the coordinates, and then all the line segments. Maybe that deserves it's own issue or you could just add to this PR

kevinyamauchi:
Thanks for the review @sofroniewn and the feedback @HagaiHargil !

@sofroniewn , my understanding of the main points in your comment is that you would do not want the selected points in `Points.selected_data` to persist when changing modes because the user should not be able to modify point properties via the layers control UI when not in selection mode. Is that right?

I definitely agree that when outside of the selection mode (both 2D and 3D), adjusting the point layer controls should not modify the point properties. That was an oversight on my part and I will fix it.

In terms of `Points.selected_data` persisting when switching modes, I am a little torn. I can see why users interacting with the selected points (e.g., plugin developers) would want to be able to continue to access which points were selected after the mode is changed (i.e., clearing selected points should be orthogonal to changing the mode). I think there are a couple of approaches we can take here:

1. As suggested by @sofroniewn , we can store the selected points in a property (e.g., `Points._selected_data_stored`) when transitioning from select mode to a different mode. A user that wants to retrieve or operate on selected points would then have to first check the `Points.mode` and then get the selected points from the appropriate property (i.e., `Points.selected_data` or `Points._selected_data_stored`),.
2. We can allow `Points.selected_data` to persist across modes, but deactivate updating the selected point display properties via the layer control UI (e.g., add some logic to the `Points.current_*` setters). 

What do you all think? I don't have a strong preference, but it seems like there are advantages to keeping the "current selection" in one property (approach 2). Of course this would require us to also change the behavior in the `Shapes` layer to match.

If we go with approach 1, I think we should still consider prevent point properties from being updated via the layer control UI when outside select mode by adding some logic to the `Points.current_*` setters.  Currently, the user can select points via the command line (`points_layer`.selected_data = [0]`) and then adjust point properties when outside select mode.
 

> I also find it strange when I am in 3D viewing in select mode and I can't see which points I have selected and then adjust some properties and some change (we can work on this in separate PRs)

When I test on my machine, I cannot switch to select mode (`Points.mode  = 'select'`). Is that that case for you? Or did you mean when points are selected in 2D and then switching to 3D with them still selected (i.e., the points are listed in `Points.selected_data`)? I just want to clarify in case I accidentally introduced another bug.

**3D bounding boxes**
In terms of the visualizing the 3D selection, I am happy to work on adding support for bounding boxes, etc. in 3D. I think that is it's own issue though.

sofroniewn:
> When I test on my machine, I cannot switch to select mode (Points.mode = 'select'). Is that that case for you? Or did you mean when points are selected in 2D and then switching to 3D with them still selected (i.e., the points are listed in Points.selected_data)?

I misspoke a little - that is the same for me, I cannot switch to `select`, I am forced into pan-zoom, but now I am able to adjust those properties that previously I couldn't.

> I can see why users interacting with the selected points (e.g., plugin developers) would want to be able to continue to access which points were selected after the mode is changed (i.e., clearing selected points should be orthogonal to changing the mode)

Ok, maybe I'm coming around to this point of view (thanks for your patience!). Looking at illustrator when you switch to the magnifying glass you don't unselect any of the currently selected shapes that you have selected, so maybe it's strange that we do that. Maybe then point properties (like face color etc) should continue to be editable in pan/zoom mode too.

The thing I really care about is that when we have selected points you see the visual representation of that inside the GUI - i.e. the bounding box and highlight. That should be true regardless of mode and everything else when the layer is selected. There is a little complexity with how we are currently doing the line styling to making it be the right size when zooming, but that will be fixable in the view code. If you unselect the layer I think you should then leave those points selected too, but you might not want to display those bounding boxes anymore - though maybe it's ok to also just leave them. How does this sound @kevinyamauchi @HagaiHargil?  

An additional comment (maybe orthogonal to this, but I might as well mention it now as I think it relates to how we think about what select mode is) is that I think right now it's still not clear how to best do things like make a subset of points red (or make all the points red!) from the command line. For example right now you'd have to first "select" the points by setting the `selected_data` property and then change the `current_face_color` property to red, and then you might want unselect them too. We might want to make some convenience methods like `set_edge_color(list_of_points, color)` that does this stuff for you. Or you could set the rgba values of the `edge_color` array directly and call `refresh` on the layer (but I think I prefer the other approach).

Note also that I had wanted to pull some of this mode stuff out of the individual layers into the base layer - see #772 but I think we're waiting on that till all of this other stuff is more clear

> In terms of the visualizing the 3D selection, I am happy to work on adding support for bounding boxes, etc. in 3D. I think that is it's own issue though.

Thanks!!

HagaiHargil:
> An additional comment (maybe orthogonal to this, but I might as well mention it now as I think it relates to how we think about what select mode is) is that I think right now it's still not clear how to best do things like make a subset of points red (or make all the points red!) from the command line. For example right now you'd have to first "select" the points by setting the selected_data property and then change the current_face_color property to red, and then you might want unselect them too. We might want to make some convenience methods like set_edge_color(list_of_points, color) that does this stuff for you. Or you could set the rgba values of the edge_color array directly and call refresh on the layer (but I think I prefer the other approach).

I feel like this point hits the nail on its head, and is very much related to this PR and to my original issue. When doing exploratory data viz work in napari you deal very little with the "command line interface" (what is the canonical name for it BTW? Scripting interface?), but in the past couple of weeks I've been trying to generate figure-worthy plots with napari, so my intent was to delegate as much work as possible to the code, so that when the viewer loads up I basically have the viz ready the way I want it. But this point\shape selection API is indeed not too user friendly quite yet, and it obviously still has some issues. It's almost as if it's a second-class citizen in napari, which I believe it shouldn't be.

For example, when I use the command line I don't think I should care about the mode (pan\zoom or selection) that the viewer is in, simply because I'm not using the viewer at all - I might even not have it available quite yet. From this POV it also doesn't make too much sense to move the selected points, when transitioning to 3D, from `Points.selected_data` to `Points._stored_selected_data`, since my script ideally (IMHO) shouldn't be affected by the viewing mode, as stated earlier. This is why I'd obviously welcome any change that improves the UX when handling selected data during scripting.

Lastly, regarding the 3D bounding box, maybe @kevinyamauchi can only keep the circles around the points, while discard the actual bounding box? This might be easier to implement, and these circles are usually pretty clear.

sofroniewn:
> For example, when I use the command line I don't think I should care about the mode (pan\zoom or selection) that the viewer is in, simply because I'm not using the viewer at all - I might even not have it available quite yet.

Yup, this is fair enough. How do you feel @HagaiHargil about the proposed convenience functions `set_edge_color(list_of_points, color)`? Or do you have other recommendations for the scripting API (which is the term I was looking for too!!). Thoughts on this @kevinyamauchi too?

> Lastly, regarding the 3D bounding box, maybe @kevinyamauchi can only keep the circles around the points, while discard the actual bounding box? This might be easier to implement, and these circles are usually pretty clear.

This could work well and certainly could be an easier to achieve intermediate - I leave it to @kevinyamauchi to decide

kevinyamauchi:
**Scripting API**
Thanks for the feedback, @HagaiHargil . As you mentioned, the "scripting API" hasn't been the main focus and thus is in a pretty nascent state. Given that the plugin architecture is starting to take shape, it is probably a good time to make a concerted effort to improve it. Please continue to let us know when there are changes/features that can make your interactions with napari smoother. Your comments, issues, and PRs have been instrumental in making napari better!

With regards to convenience functions such as `set_edge_color(point_indices, colors)`, I am definitely open to it (seems useful!). I would want to learn more about the use cases before I give a stronger answer. I think we need to balance convenience functions against creating a large API with many parallel paths to achieve a given task. For example, once #895 lands, colors will be able to be set by directly setting with an array of colors, setting colors with a property and modifying the properties and `set_edge_color()` would be a third way. I think having an API that is too granular will be harder to maintain and it will be harder for users to know which is the "best" or "right" way to use napari. Maybe it makes sense to open an issue where we collect some use cases and try to solidify what scripting features should be made available in core napari?

**Persistent point selection**
I agree that it should be clear to users which points are selected, irrespective of the mode. I like @HagaiHargil 's suggestion to leave the circles around the selected points when outside of the selection mode (both 2D and 3D). I also agree that the selection indicator should go away when deslecting the layer. I can incorporate that into this PR.

Are we in agreement that selected points should remain in `Points.selected_data` when transitioning across modes? If so, should the layer control UI still be able to modify the selected points (assuming there is some indication in the viewer which points are selected)? I think it makes sense to do so if there is an indication as to which points are selected, but I am open for discussion.

sofroniewn:
> I think having an API that is too granular will be harder to maintain and it will be harder for users to know which is the "best" or "right" way to use napari. Maybe it makes sense to open an issue where we collect some use cases and try to solidify what scripting features should be made available in core napari?

Yeah - a dedicated issue and discussion on this seems pretty important, and I appreciate you both weighing in there. Can you make one @kevinyamauchi. 

> Are we in agreement that selected points should remain in Points.selected_data when transitioning across modes? If so, should the layer control UI still be able to modify the selected points (assuming there is some indication in the viewer which points are selected)? I think it makes sense to do so if there is an indication as to which points are selected, but I am open for discussion.

Yes, let's go for it - selection remains as long as their is some visual indication of selection. For 2D that is hightlight + bounding box, for 3D that will just be highlight until we get bounding box support there too.

Ultimately we will want to standardize this stuff with between the Shapes and Points layer too, but we can solve first for points.

@kevinyamauchi I leave it to you to propose what to do in this PR vs what to do in follow up PRs

kevinyamauchi:
Sounds good, @sofroniewn. I propose the following:

For this PR:
- Selected points will always be highlighted when the points layer is active. In 2D, selected points will have a highlighted marker and a bounding box. In 3D, the selected points will only have a highlighted marker. If the points layer is inactive, no points are highlighted.
- Selected points persist in `Points.selected_data` when changing modes.
- The properties of the selected points can be modified via the layer controls UI in select and pan/zoom mode. They cannot be modified in the add mode (users may want to put down points with different properties and that should not affect existing points).

In future PRs:
- Bounding boxes for highlighted points in 3D
- Convenience functions for a "scripting API"

kevinyamauchi:
I have added the proposed behavior. It's still a work in progress, as one of the tests is intermittently failing and there's some code clean up to do. However, it would be great to get some feedback on the behavior when you have time, @sofroniewn and @HagaiHargil .

napari/layers/points/points.py
```diff
@@ -250,7 +250,9 @@ def __init__(
             size=Event,
             edge_width=Event,
             face_color=Event,
```
sofroniewn, line 252: why did you create the `current_face_color` event instead of re-using the `face_color` event?

sofroniewn (commented):
This is looking good. One piece of feedback, when you switch to add mode and start adding new points any already selected points should probably be unselected and each new point should be selected (only ever one at a time if many points are added sequentially).

I'll also note that moving to a different layer doesn't unselect the points, which might be ok, but it might make things a bit cluttered if this starts happening in many layers

napari/layers/points/points.py
kevinyamauchi: I added the current_* events to update the current value in the layer control dropdowns. Previously, the face_color and edge_color dropdown menus were triggered by the face_color event and updated from `Points.current_face_color` and `Points.current_edge_color`. `face_color` and `current_face_color` aren't always updated at the same time, so I thought it made sense to create an event specifically for the property that is being viewed. Parts of the view are refreshed by the `face_color` event, so thought it was better not to refresh them every time current_face_color is updated (e.g., in add mode and selecting a new color for the next point to be added). Do you think that makes sense?

kevinyamauchi:
> I'll also note that moving to a different layer doesn't unselect the points, which might be ok, but it might make things a bit cluttered if this starts happening in many layers

I don't think it would be too hard to implement that. I try to add it into this PR, as I agree it will make things cluttered.

napari/layers/points/points.py
sofroniewn: yeah it does, we might want to separate them for all `current_*` properties.

kevinyamauchi:
I just added highlight toggling on layer select/deselect! I will try to track down the cause of the failing test soon.

napari/_vispy/vispy_points_layer.py
```diff
@@ -85,43 +81,42 @@ def _on_data_change(self, event=None):
...
+                width = 0
+            else:
+                pos = self.layer._highlight_box
+                width = self._highlight_width
+
+            self.node._subvisuals[2].set_data(
+                pos=pos[:, ::-1] + 0.5,
+                color=self._highlight_color,
+                width=width,
+            )
```
jni, line 120: For what it's worth, I'd consider removing the interaction box entirely. I think just highlighting the points is sufficient.

jni (approved):
LGTM! I've commented that I think we could just remove the interaction box, but that can come later.

napari/_vispy/vispy_points_layer.py
sofroniewn: @kevinyamauchi if you want to pull the interaction box entirely (so even in 2D) that is fine with me, no real need to have it as there is no functionality associated with it beyond highlighting. If you remove if from this PR we are then good to merge

kevinyamauchi:
Thanks for the reviews @jni and @sofroniewn ! I have removed the section box, as suggested. I did leave the dragging box because I think it makes it a lot easier to see what you are selecting. The tests are passing and I will leave it up for a day or so to see if anybody has further comments before merging!

jni:
👍 !

sofroniewn (approved):
Love it, just tried it out - works great. I will merge

PR #908, opened 2020-01-26
Implement pluggy as plugin manager

tlambert03:
# Description
sorry to muddy the water a bit... but in the interest of being thorough, I'd like us to at least consider using [pluggy](https://pluggy.readthedocs.io/en/latest/) (the plugin backend for pytest and tox) as our plugin manager before moving forward..  While I was writing the code for #891 that would control plugin module discovery and how plugins declare the interfaces that they implement,  I was trying to create at least some code that might be reusable as we add different (non-reader) type plugins later... which eventually led me to pluggy.  This PR serves as a minimal example of what it might look like to use pluggy as our plugin system.  I actually like it a lot.

### nice things

What I really like about pluggy is how the host application declares [hook specifications](https://pluggy.readthedocs.io/en/latest/#specifications) that are basically just well-documented function signatures (with type annotations if desired) that plugins can implement. We then identify places in our code that these hooks/functions get called. These hookspecs lend themselves nicely to autodocs (see the [pytest hooks API](https://docs.pytest.org/en/latest/reference.html#hooks)).  We can then also eat our dogfood/write our own internal plugins for these hooks, and can modify the order in which the registered hook implementations are called.
It has a nice [namespace-based plugin organization](https://pluggy.readthedocs.io/en/latest/#define-and-collect-hooks), and it also allows for [easy hook deprecation](https://pluggy.readthedocs.io/en/latest/#warnings-on-hook-implementation). 

### one thing you might not like

the way pytest uses this (and the convention seems to be) that all hook functions that plugin modules may define must begin with the name of the main application (e.g. `pytest_`... or in our case `napari_`).  since ultimately this is about plugging in to `napari`, i don't see that as a big problem, but just wanted to point it out in case you think that is too "napari-centric"

**update**: but see https://github.com/pytest-dev/pluggy/pull/251, which could address this.
- - - 

The entirety of the recently proposed IO plugin architecture can basically be done in a few lines with pluggy (most of which is just docs).  In this PR:
- hook specifications are defined in [`napari/plugins/hookspecs.py`](https://github.com/napari/napari/compare/master...tlambert03:pluggy?expand=1#diff-0be73a10495ecfe2ae42b85734530cac)
- an internal "plugin" (implementation) of the one current hook is in [`napari/plugins/_internal.py`](https://github.com/napari/napari/compare/master...tlambert03:pluggy?expand=1#diff-da37e9711d62401c1c25ffdf7e3e0ad2)
[...]

PR #909, opened 2020-01-26
_add_layer_data method and tests

tlambert03:
# Description
I'm going to try to break up #891 / #908 into small PRs with isolated functionality inasmuch as possible, for easier review.
This PR adds an `_add_layer_data` to the `Viewer` model that will be useful for adding arbitrary layer data received from IO plugins in the form of `(data, [meta[, layer_type]])`.  (I chose to put this on the viewer model rather than pulling `layer2addmethod` out of the tests utils.)

This method should be moved to the new `AddLayersMixin` in #897 if this merges first
(or I will move it in this PR if #897 merges first).

## Type of change

# How has this been tested?
- [x] added tests for adding all layer types, and checking for badly formatted data

## Final checklist:


napari/components/tests/test_viewer_model.py
```diff
@@ -315,3 +317,54 @@ def test_add_remove_layer_dims_change():
...
+]
+
+
+@pytest.mark.parametrize('data', good_layer_data)
+def test_add_layer_data(data):
+    # make sure adding valid layer data calls the proper corresponding add_*
+    # method for all layer types
+    viewer = ViewerModel()
+    viewer._add_layer_data(*data)
+    assert len(viewer.layers) == 1
```
kevinyamauchi, line 331: I think it would be nice to include a check that the right type of layer is added.

napari/components/viewer_model.py
```diff
@@ -1000,6 +1000,69 @@ def add_vectors(
...
+        ...     'points',
+        ... )
+        >>> viewer._add_layer_data(*data)
+
+        """
+
+        try:
+            add_method = getattr(self, 'add_' + layer_type)
+        except AttributeError:
+            valid_types = {
```
kevinyamauchi: I think this would this include the `Viewer.add_layer()`. If so, I don't think `layer` would be a valid `layer_type` value.

napari/components/viewer_model.py
```diff
@@ -1000,6 +1000,69 @@ def add_vectors(
         self.add_layer(layer)
         return layer
 
+    def _add_layer_data(
```
kevinyamauchi: This is totally a nitpick, but I think the name `_add_layer_data()` is confusing. That being said, I'm not sure what is better. Maybe `_add_layer_from_data()` (too long?)?

kevinyamauchi (commented):
Thanks, @tlambert03 ! I left a couple of minor comments below, but this looks good to me!

napari/components/viewer_model.py
tlambert03: yeah, I meant to mention that in the PR.  I'm not in love with it either and definitely open to options.  I don't mind `_add_layer_from_data` ... maybe we can one more person to weigh in and I'll go with the consensus.  The other question is: does everyone agree this should be a _private method?

napari/components/viewer_model.py
tlambert03: ahh.  good catch.  If possible, I'd prefer to *not* have to specify every valid layer type, just because it's one more thing to have to remember if we add new `Layer` types later... but then again we may eventually add new public `add_*` methods to the viewer as well, so simply excluding `layer_type == " layer"` could also break in the future.  Do we have a "single source of truth" somewhere in the codebase for valid layer names?

napari/components/tests/test_viewer_model.py
tlambert03: can you clarify?  do you mean that the `data` format matches the `layer_type`? or that the `layer_type` is a valid string?

tlambert03: ohhh... sorry, I get it.  that the layer that got added to the viewer is the right type, right?

kevinyamauchi: Ah sorry for being unclear. Maybe it's overkill, but I was thinking it would be nice to make sure the default `layer_type` and method name lookup are working. Something like:

```python
assert isinstance(viewer.layers[0], correct_layer_type)
```

napari/components/viewer_model.py
tlambert03: how about something like this (in addition to making sure an `add_*` method is available)
```python
valid_layer_types = [subclass.__name__.lower() for subclass in Layer.__subclasses__()]
```

kevinyamauchi: I agree that we should avoid adding something else that needs to be manually curated when adding/modifying the layer classes. This would definitely not be the intended usage, but there is a dictionary containing valid layers in the [vispy utils](https://github.com/napari/napari/blob/b32501198c52e3145dca1fb2feeac284227013a7/napari/_vispy/utils.py#L9) (used to map layer models to views). Maybe somebody else knows a better way!

kevinyamauchi: > how about something like this (in addition to making sure an add_* method is available)

Sorry, I missed this! That looks good to me!

tlambert03: realized I actually need to make it recursive in order to catch things like `Labels` (which is subclassed from `Image`)...  so I made a convenience method in utils, and added a new module-level `layers.names` variable that can serve as a source-of-truth if everyone is ok with that.

napari/components/viewer_model.py
jni: I'm :+1: on `_add_layer_from_data`, and yes, let's keep it private for now, in case we want to change it as we use it more.

napari/components/viewer_model.py
jni: I love the recursive subclasses approach! Probably useful in other parts of the codebase!

tlambert03:
made all requested changes.  ready for re-review.  unrelatedly, it looks like conda 4.8.2 somehow broke our python 3.6 environment with unresolvable conflicts ...  :/

napari/layers/__init__.py
```diff
@@ -5,11 +5,14 @@
...
+from .labels import Labels
 from .points import Points
-from .vectors import Vectors
-from .surface import Surface
 from .shapes import Shapes
-from .labels import Labels
+from .surface import Surface
+from .vectors import Vectors
+
+names = [subclass.__name__.lower() for subclass in all_subclasses(Layer)]
```
kevinyamauchi: I think this should be all caps (i.e., a constant)?
```suggestion
NAMES = [subclass.__name__.lower() for subclass in all_subclasses(Layer)]
```

napari/utils/misc.py
```diff
@@ -165,3 +165,16 @@ def __str__(self):
...
+
+def all_subclasses(cls):
+    """Recursively find all subclasses of class ``cls``.
+
+    Returns
+    -------
+    set
+        the set of all classes that are subclassed from ``cls``
+    """
+    return set(cls.__subclasses__()).union(
```
kevinyamauchi, line 184: Awesome!

napari/utils/misc.py
```diff
@@ -165,3 +165,16 @@ def __str__(self):
 
 
 callsignature = CallSignature.from_callable
+
+
+def all_subclasses(cls):
```
kevinyamauchi: Is `cls` reserved for class methods in python?

napari/utils/misc.py
```diff
@@ -165,3 +165,16 @@ def __str__(self):
...
+    """Recursively find all subclasses of class ``cls``.
+
+    Returns
```
kevinyamauchi, line 179: For the sake of completeness, I think it would be nice to have a Parameters section, even though it's basically covered in the function description.

napari/components/viewer_model.py
```diff
@@ -1000,6 +1000,73 @@ def add_vectors(
...
+        """
+
+        if layer_type not in layers.names:
+            raise ValueError(
+                f"Unrecognized layer_type: '{layer_type}'. "
+                f"Must be one of: {layers.names}."
+            )
+
+        try:
+            add_method = getattr(self, 'add_' + layer_type)
```
kevinyamauchi: It might be nice to use `layer_type.lower()` in case people use a different case. I could see getting confused since the class names are capitalized.

kevinyamauchi: Although, I guess they would have received the very clear `ValueError` above. I'm happy with whatever you think is best!

napari/components/viewer_model.py
```diff
@@ -1000,6 +1000,73 @@ def add_vectors(
...
+        except AttributeError:
+            raise NotImplementedError(
+                f"Sorry! {layer_type} is a valid layer type, but there is no "
+                f"viewer.add_{layer_type} available yet."
+            )
+
+        try:
+            add_method(data, **(meta or {}))
+        except TypeError as exc:
```
kevinyamauchi: Nice!

kevinyamauchi (approved):
Thanks, @tlambert03 . This looks great to me! I left a few minor comments below.

napari/components/tests/test_viewer_model.py
```diff
@@ -315,3 +317,63 @@ def test_add_remove_layer_dims_change():
     viewer.layers.remove(layer)
     assert len(viewer.layers) == 0
     assert viewer.dims.ndim == 2
+
+
+good_layer_data = [
```
sofroniewn: how about putting this in our `napari/tests/utils.py` ?

napari/components/viewer_model.py
```diff
@@ -1000,6 +1000,73 @@ def add_vectors(
         self.add_layer(layer)
         return layer
 
+    def _add_layer_from_data(
```
sofroniewn: I don't like `_add_layer_from_data` as we use `data` in other ways in other places (i.e. just as the array). I don't think `_add_layer_from_tuple` or `_add_layer_from_dict` are any better, so I'm not quite sure right now. I will sleep on it and maybe have a better idea in the morning :-)

sofroniewn (commented):
Overall this is looking great, and I super appreciate getting this in before #891 / #908 - makes it so much easier to review!!!!

I have merged #897 now. I agree makes sense to move `_add_layer_from_data` over to the mixin too. Curious if you think it makes sense to move `add_layer` over there to the mixin too and make it private `_add_layer`. I think so given the we'll now have `_add_layer_from_data` over there too, but curious what you think (note there was some discussion of this in #892)

My main reservation with this PR is the name `_add_layer_from_data` - I find it confusing as noted in one of my other comments. I'm not sure what a better one is right now `_add_layer_from_*` could work, if we thought of a better `*`.

napari/utils/misc.py
tlambert03: `cls` is not reserved, it's just a convention (like `self`).  and as far as usage I have seen, it's a commonly used variable when referring to "some" class.

napari/utils/misc.py
tlambert03: ha! funnily-enough, my autoDocstring extension usually adds one, but must not have because of the `cls` variable name. 😅

tlambert03:
> Curious if you think it makes sense to move `add_layer` over there to the mixin too and make it private `_add_layer`. I think so given the we'll now have `_add_layer_from_data` over there too, but curious what you think (note there was some discussion of this in #892)

If we're adding this `_add_layer_from_data` method to the mixin, then I think it makes sense to also put `add_layer` in the mixin (as it would then have all of the "add_*" methods in the mixin).  If you wanted to keep the `AddMixin` as a strict mixin to add layer types, then actually I think neither `add_layer` nor this PR should be in the mixin.  But I favor putting them both there.  Also, I like `add_layer` as a public method.

I'll move `add_layer` over to the mixin on the assumption we agree, but if not let me know and I'll move them both back to the main class.

> My main reservation with this PR is the name `_add_layer_from_data`... I don't like `_add_layer_from_data `as we use `data` in other ways in other places (i.e. just as the array)

yeah, no one seems to have any good convictions on this one :/ 
I don't personally mind it, since I think it does capture what is being done and the `data` argument to this method actually _is_ the same data as would be handed to the corresponding `add_*` methods, like the array (so I i'm not as sure we use it in other ways in other places).

kevinyamauchi:
> yeah, no one seems to have any good convictions on this one :/
I don't personally mind it, since I think it does capture what is being done and the data argument to this method actually is the same data as would be handed to the corresponding add_* methods, like the array (so I i'm not as sure we use it in other ways in other places).

I agree that the usage of "data" here is consistent with the data parameter in the layers for the reasons mentioned by @tlambert03 . However, I am terrible at naming things, so I am happy to consider other options!

napari/utils/misc.py
kevinyamauchi: Got it, thanks for the explanation!

sofroniewn (approved):
This looks great and I've approved it too. `add_layer_tuple` or `add_layer_from_tuple` are the best I could come up with instead of `add_layer_from_data`. Maybe let's @jni look at the whole thing before we merge (he also might have ideas on the name!)

tlambert03:
> This looks great and I've approved it too. `add_layer_tuple` or `add_layer_from_tuple` are the best I could come up with instead of `add_layer_from_data`.

yeah... but since the function doesn't actually accept a tuple, that's not ideal either (nor `from_dict`).

sofroniewn:
How about `_add_layer_with_parameters`?

tlambert03:
> How about ‘_add_layer_with_parameters’?

that's fine with me too, i feel about the same as for `_add_layer_from_data`
@jni ?  can you put us out of our ambivalent misery here with a strong opinion?  that's all this is waiting on.

jni:
Haha I sadly don't have a strong opinion, and am a solid "meh" on both candidates. 😂 ultimately, it's a private method, so let's merge whatever is there, and we can rename it in the future if we think of something better.

sofroniewn:
Ok sounds good @jni. @tlambert03 if you feel good about this you can merge it as is when you are ready

Issue #875, opened 2020-01-13
Testing a pip install doesn't work


jni:
## 🐛 Bug

Running `pytest --pyargs napari` should run the napari test suite, but right now, this fails:

```pytb
 $ pytest --pyargs napari
============================= test session starts ==============================
platform darwin -- Python 3.7.6, pytest-5.3.2, py-1.8.1, pluggy-0.13.1
PySide2 5.14.0 -- Qt runtime 5.14.0 -- Qt compiled 5.14.0
rootdir: /Users/jni
plugins: ordering-0.6, qt-3.3.0, cov-2.8.1
collected 149 items / 4 errors / 145 selected

==================================== ERRORS ====================================
_ ERROR collecting conda/envs/napari-0.2.10/lib/python3.7/site-packages/napari/_qt/tests/test_qt_viewer.py _
ImportError while importing test module '/Users/jni/conda/envs/napari-0.2.10/lib/python3.7/site-packages/napari/_qt/tests/test_qt_viewer.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
conda/envs/napari-0.2.10/lib/python3.7/site-packages/napari/_qt/tests/test_qt_viewer.py:6: in <module>
    from napari.tests.utils import (
E   ModuleNotFoundError: No module named 'napari.tests'
_ ERROR collecting conda/envs/napari-0.2.10/lib/python3.7/site-packages/napari/layers/image/tests/test_big_image_timing.py _
ImportError while importing test module '/Users/jni/conda/envs/napari-0.2.10/lib/python3.7/site-packages/napari/layers/image/tests/test_big_image_timing.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
conda/envs/napari-0.2.10/lib/python3.7/site-packages/napari/layers/image/tests/test_big_image_timing.py:2: in <module>
    import zarr
E   ModuleNotFoundError: No module named 'zarr'
_ ERROR collecting conda/envs/napari-0.2.10/lib/python3.7/site-packages/napari/layers/image/tests/test_image.py _
ImportError while importing test module '/Users/jni/conda/envs/napari-0.2.10/lib/python3.7/site-packages/napari/layers/image/tests/test_image.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
conda/envs/napari-0.2.10/lib/python3.7/site-packages/napari/layers/image/tests/test_image.py:4: in <module>
    import xarray as xr
[...]

PR #911, opened 2020-01-27
Exclude `benchmarks` from the list of packages found

### LINKED ISSUE #910: napari should not install "benchmarks" as a toplevel package
## 🐛 Bug

pip-installing napari not only creates a toplevel "napari" importable package, but also a spurious "benchmarks" package.

## To Reproduce

In a clean venv, run `pip install napari`.  Check that "benchmarks" has been installed as a toplevel package with e.g. `python -c 'import benchmarks; print(benchmarks)'`

## Expected behavior

No such package installed.

## Environment

This occurs when installing napari==0.2.10.

ttung:
# Description
The benchmarks directory was not meant to be packaged for distribution.

## Type of change

# References
Fixes #910

# How has this been tested?
```
% python setup.py sdist
% mkdir /tmp/4
% cd /tmp/4
% pip install ~/imaging/napari/dist/napari-0+unknown.tar.gz
% python 
Python 3.7.5 (default, Oct 28 2019, 07:13:20) 
[Clang 11.0.0 (clang-1100.0.33.8)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import benchmarks
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'benchmarks'
>>> ^D
%
```

## Final checklist:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -54,7 +54,7 @@ if sys.version_info < (3, 6):
 
 
 PACKAGES = [
-    package for package in find_packages() if not package.startswith('gui')
+    package for package in find_packages(exclude=['benchmarks']) if not package.startswith('gui')
 ]
 
 


kne42 (approved):
thanks!

sofroniewn:
This will close #910. @ttung will the benchmarks will still get included in the pip install? i.e. if i want to run them i can, but they just won't be available as a top level import, or will that code not even get distributed now?

I don't know what the conventions / norms around this are so I'll ultimately defer to others who know more. It makes sense that we shouldn't be adding something that appears under `import benchmarks` but it might still be nice for people to run the benchmarking suites even if they got napari from a pip install? Or maybe that isn't needed? What does @jni think?

ttung:
I think it's far better to bundle these into `napari.benchmarks` if you want them distributed.

sofroniewn:
> I think it's far better to bundle these into napari.benchmarks if you want them distributed.

Makes sense - I do see this as quite analogous to #875 too, so we might want to mirror whatever solution we come up with there here - see #912

ttung:
Dumping this one...

PR #915, opened 2020-01-27
patch py3.8 on windows

tlambert03:
# Description
napari is broken on python 3.8 on windows, because a change in `asyncio` in python 3.8 poses a [problem for tornado](https://github.com/tornadoweb/tornado/issues/2608), which `ipykernel` in turn depends on, which our console mechanism in-turn depends on.

as of v5.1.4 ipykernel has fixed this for instances of their `IPKernelApp`.  but the patch doesn't seem to be triggered in the way we access the kernel (maybe lower down... haven't looked too far into it).
I borrowed their fix from [here](https://github.com/ipython/ipykernel/blob/7b2727d827a95d2059c9d9d9354899feabb642ee/ipykernel/kernelapp.py#L506-L535) and added it to `qt_console`, it fixes napari and tests on windows py 3.8.

## Type of change

# References
https://github.com/tornadoweb/tornado/issues/2608
https://github.com/ipython/ipykernel/pull/456

# How has this been tested?

### DIFF
--- a/napari/_qt/qt_console.py
+++ b/napari/_qt/qt_console.py
@@ -1,11 +1,33 @@
-from qtconsole.rich_jupyter_widget import RichJupyterWidget
-from qtconsole.inprocess import QtInProcessKernelManager
-from qtconsole.client import QtKernelClient
-from IPython import get_ipython
-from IPython.terminal.interactiveshell import TerminalInteractiveShell
+import sys
+
+from ipykernel.connect import get_connection_file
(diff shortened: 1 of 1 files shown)


sofroniewn (approved):
I'm ok with this, would be nice to get this in now and then round out #878 

@jni thoughts?

sofroniewn:
I am merging this now so we can merge get 3.8 support in with #878

jni:
LGTM, thanks for checking @sofroniewn and thanks for the quick fix @tlambert03! =)

PR #920, opened 2020-01-28
Fix nD Shapes.to_labels()

### LINKED ISSUE #914: to_labels broken in nD_shapes.py
>
> Hi Nicolas,
>
> Thanks for building such a great visualization tool for python!
>
> I have been trying to use (napari version `0.2.10+7.g47af135`) the `.to_labels` functionality on an image stack and not getting the behavior I expected. For example, if I add the line `print(np.unique(labels))` at line number 36 of the nD_shapes.py example, expecting this to reflect the labels of all unique shapes, the output that I get is `array([0])`, indicating no labels present.
>
> If I change line 34 to read:
`labels = layer.to_labels(labels_shape=(128, 128))`
I get an index for all 128 shapes, but of course they are now compressed into one image plane, but if line 34 reads:
`labels = layer.to_labels()`
I again get only zeros (but with shape `(127,128,128)`).
>
> It does seem that `to_labels` is meant to be compatible with n-dimensional images, am I expecting the wrong behavior, or otherwise misusing this functionality?
>
_Originally posted by @miketaormina in https://github.com/napari/napari-tutorials/issues/46#issuecomment-578882563_

kevinyamauchi:
# Description
This PR fixes `Shapes.to_labels()` for nD `Shapes` layers with nD data. 

## Type of change

# References
Fixes #914 

# How has this been tested?
- [ ] TODO: add test for nD Shapes.to_labels()

## Final checklist:


sofroniewn:
Nice catch @kevinyamauchi - should we maybe add some better tests for this stuff, like make sure we get an array of the right size back while you're at it, right number of labels if you call np.unique, or right bool data type. Would be good to have tests for `to_labels` and `to_masks` in 2D and nD.

kevinyamauchi:
@sofroniewn , I'd definitely like to take the opportunity to add some tests. I was trying to add a some tests with simple shapes (e.g., squares) that we can test resulting masks. However, I am running into strange `Shapes.shape` values when instantiating just the model with nD shapes. I will see if I can pin down what the root cause is. In the meantime, I can add the tests you mentioned!

kevinyamauchi:
@sofroniewn , I am curious on your thoughts about this PR. It does technically fix the issue in that the user can now generate valid label images via `Shapes.to_labels()`. The one wrinkle is that due to the shape/range issue described in #922. The output can be strange in the following condition: (1) there are no shapes at the origin and (2) the user doesn't provide an appropriate shape to `labels_shape`.

That is, `Shapes_layer.to_labels()` doesn't work very well with the default values. The main reason is the range/extent is calculated as the bounding box around shapes. Consider a `Shapes` layer with a square that has vertices:  [ [100, 100], [100, 200], [200, 100], [200, 200] ]. The shape will be (100, 100) or (101, 101) (depending on what we decide in #922). If no `labels_shape` is provided to `Shapes_layer.to_labels()`, it will guess the shape based on the layer shape, which means it will try to make the label image with shape (100, 100). While that is the correct shape, it is assuming that that array is starting at the origin. However, if the user provides a sensible `labels_shape` that would encompass the square when starting at the origin  (e.g., (250, 250) ), it works.

This doesn't pop up in the labels examples because the shapes are superposed on an image and the `labels_shape` is given as the shape of the image (and thus will include all of the shapes).

All of that being said, even though there is more work to be done, I think it's worth merging, as it at least removes a bug that prevents `Shapes.to_labels()` usage on nD labels and doesn't introduce any new bugs. What do you think?

sofroniewn:
Merging now to fix the bug is fine and then working on a larger fix for `0.2.12` is fine too. Do you want to add more tests to this PR, or should we just get it in? Maybe also can you create a dedicated follow-up issue, or should we discuss on #922 more.

kevinyamauchi:
@sofroniewn sounds good. I agree on holding off until `0.2.12` for the larger fix. I think we should discuss more in #922 first and then make a dedicated issue if necessary.

I think the current tests are sufficient for now, but we can definitely improve them in the larger fix.

sofroniewn (approved):
Ok I am approving and will merge. We can discuss more on #922 - great to have you looking at this part of the code base @kevinyamauchi - it was bundled with a lot of other stuff when it first went in.

PR #923, opened 2020-01-29
Fix singleton dims

### LINKED ISSUE #888: singleton dimensions are a problem
## 🐛 Bug
Was working in napari with someone yesterday and one of their images revealed an issue with napari's handling of singleton dimensions:

```python
import napari, numpy

with napari.gui_qt():
    napari.view_image(numpy.random.rand(10,1,10,10))
```

Screen Shot 2020-01-17 at 8 56 50 AM

(see dimensions sliders 🖕)
this would be solved visually easily enough by adding `np.squeeze` somewhere in `magic_imread`, but i'm not sure that's necessarily what we want to do here.  perhaps dealing with them in the dims model is more appropriate?  If anyone has a preference on how to tackle this, I'm happy to submit a PR

sofroniewn:
# Description
This PR fixes #888 and adds tests to catch if from happening again.

See before:

Screen Shot 2020-01-28 at 4 45 16 PM

and after fix:

Screen Shot 2020-01-28 at 4 46 08 PM

## Type of change


sofroniewn:
Interesting to see the test fails that I've introduced

```
TEARDOWN ERROR: Exceptions caught in Qt event loop:
________________________________________________________________________________
  File "/tmp/cirrus-ci-build/napari/_qt/qt_dims_slider.py", line 61, in change_slice
    val = int(self.curslice_label.text())
NameError: free variable 'self' referenced before assignment in enclosing scope
```

probably because I've added the `view.show()` calls (which were needed to test the visibility properly)

jni (approved):
I can't say that I understand these changes, but code looks good. =) CI isn't happy though. I'm approving so you're not held up by review.

sofroniewn:
Thanks @jni! I think I want @tlambert03 to weigh in on the test fails before merge as they are definitely related to changes in this PR!!

tlambert03:
Awesome. Will take a look tomorrow

tlambert03:
that ought to fix it.  (Thanks also for fixing what I obviously re-introduced in my `dims` widget rearrangement a while back!)

>Interesting to see the test fails that I've introduced.
NameError: free variable 'self' referenced before assignment in enclosing scope

Really quite an interesting error!  Basically this closure on `self` in this function defined in `QtDimSliderWidget.__init__` was causing an error... but only in PyQt5, not pyside2! 

https://github.com/napari/napari/blob/5ed20a05f2ea0d7b0e3bfca9a5fe0c0f54239e7c/napari/_qt/qt_dims_slider.py#L60-L61

just goes to show there's more differences between those two backends than meets the eye.  The fix was easy, I just made `change_slice` a proper method (not sure why I didn't before).  You can merge when tests pass.

tlambert03 (approved):

sofroniewn:
Very interesting @tlambert03 - good to know about the pyqt5 / pyside2 difference and good to see the value in more comprehensive testing. The linux 3.7 test just failed, but not just if that was sporadic, so i am re-running. thanks for the fix!!

tlambert03:
just ran it ~20 times on my local pyqt5 test environment and couldn't reproduce it... so I think it's safely unrelated (something with CI virtualization) and you can just merge whenever.

tlambert03:
Hmm, actually it’s the same dataset that is failing 3 times in a row on pyqt5... ugh.  So annoying, can’t make it happen locally

sofroniewn:
Ah ok, probably best to think about a follow-up PR at this point if possible.

sofroniewn:
@tlambert03 are you testing locally with python 3.7 - it seems we're only getting fails there

tlambert03:
yeah.  actually, I was eventually able to reproduce it locally.  It's super interesting, and I'll post an issue and fix soon

Issue #925, opened 2020-01-29
working on fixing tests


tlambert03:
## 🐛 Bug

here are some crazy observations that started from looking into the test failure in #923, and may have some implications for general test fixes that we want to look into in #921.  I'll patch it in a PR soon, but just want to document this somewhere, in case it becomes pertinent to #921

on a mac (but we observe it in linux tests too), with python 3.7, using pyqt5 (I'm on 5.14.1), running this code:
```python
import numpy as np
import pytest  # <-- this unused import is required for the bug!
import napari

with napari.gui_qt():
    data = np.random.random((10, 15, 20))
    viewer = napari.view_image(data)
    viewer = napari.view_image(data)
    viewer.dims.ndisplay = 3
    viewer.dims.ndisplay = 2
```
leads to an error deep in vispy that culminates in:
```
lib/python3.7/site-packages/vispy/gloo/glir.py in _validate(self)
   1299                                         gl.GL_VALIDATE_STATUS):
   1300             raise RuntimeError('Program validation error:\n%s'
-> 1301                                % gl.glGetProgramInfoLog(self._handle))
   1302 
   1303     def _post_draw(self):

RuntimeError: Program validation error:
```
observations:
- only happens with python 3.7; python 3.6 & 3.8 are fine
- only happens with pyqt5, not pyside2
- only happens with the **double** viewer instantiation (two calls to `view_image`), assigning both viewers to the same variable name.  If you name the second viewer `viewer2`, it's fine
- only happens when you switch from `ndisplay=` 2, then to 3, back to 2
- the error occurs only once per session.  running the same code a second time works fine.
- pytest is key.  if you comment out the `import pytest` line, you don't get the error...
- i thought it might have something to do with our adding of 'viewer' to console globals... but removing the `self.console = QtConsole({'viewer': self.viewer})` from our `QtConsole` instantiation doesn't change anything.

tlambert03:
another note... the following will create an error if you run it in an ipython session started with 
`--gui=qt` (But not if you run it in a `gui_qt` context):

```python
import numpy as np
import pytest  # (this is required for the bug)
import napari

data = np.random.random((10, 15, 20))
viewer = napari.view_image(data)
viewer = napari.view_image(data)  # they must have the same name
viewer.dims.ndisplay = 3
```

... and then all hell breaks loose culminating in `RuntimeError: OpenGL got errors (Check before draw): GL_INVALID_OPERATION`

sofroniewn:
[...]

PR #927, opened 2020-01-30
Import scipy stats to prevent strange bug.

tlambert03:
# Description
This PR fixes one of the strangest bugs i've seen (see #925 for details) ... and I still don't have a good explanation why.  Whatever the case may be, importing scipy.stats early prevents a bug that shows up (wait for it) .... only on py3.7, with pyqt5, using vispy >= 0.6.4, when pytest has been imported, when going from 2d -> 3d -> back to 2d, when a duplicate viewer has been instantiated with the same variable name as another.  (i know... 😭)

Seems relatively harmless to add for now until we learn more, since scipy.stats is almost certainly being imported eventually.

also worth mentioning.  the bug in our tests only started showing up in `test_napari/test_view` after [this](https://github.com/napari/napari/pull/923/files) seemingly innocuous commit.

edit: because pytest needs to be imported to see the bug I moved the import from `napari.__init__` to `napari._tests.utils`

### DIFF
--- a/napari/__init__.py
+++ b/napari/__init__.py
@@ -52,3 +52,11 @@ from .utils import sys_info
 
 __version__ = get_versions()['version']
 del get_versions
+
+# this unused import is here to fix a very strange bug.
+# there is some mysterious magical goodness in scipy stats that needs
+# to be imported early.
+# see: https://github.com/napari/napari/issues/925
+from scipy import stats  # noqa: F401
+
+del stats
--- a/napari/__init__.py
+++ b/napari/__init__.py
@@ -52,11 +52,3 @@ from .utils import sys_info
 
 __version__ = get_versions()['version']
 del get_versions
-
-# this unused import is here to fix a very strange bug.
-# there is some mysterious magical goodness in scipy stats that needs
-# to be imported early.
-# see: https://github.com/napari/napari/issues/925
-from scipy import stats  # noqa: F401
-
-del stats
--- a/napari/_tests/utils.py
+++ b/napari/_tests/utils.py
@@ -3,6 +3,14 @@ import numpy as np
 from napari import Viewer
 from napari.layers import Image, Labels, Points, Shapes, Vectors, Surface
 
+# this unused import is here to fix a very strange bug.
+# there is some mysterious magical goodness in scipy stats that needs
+# to be imported early.
+# see: https://github.com/napari/napari/issues/925
+from scipy import stats  # noqa: F401
+
+del stats
+
 """
 Used as pytest params for testing layer add and view functionality (Layer class, data, ndim)
 """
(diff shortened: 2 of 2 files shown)


sofroniewn (approved):
This is truly crazy. I am fine to merge this fix now, as chasing this down further doesn’t seem like a good use of @tlambert03 time. I will note that this fix only fixes the test and not the sequence of events a user could do to reproduce the bug, but that is fine for me. We can always revisit if this becomes a more common problem.

Does @jni have thoughts before merge?

kevinyamauchi:
Wow, amazing find, Detective Talley!

tlambert03:
@sofroniewn, I originally had the import in `napari.__init__` ... which would fix it even if the use had imported pytest.  Would you prefer that?  Seems fine to me to put it there as well

jni:
@tlambert03 if you could import it and delete it at the bottom of the file, that would be fine by me.

Having said this, I find this all very unsatisfying... At the very least, I feel like we should investigate enough to raise an issue at scipy, assuming it's a scipy thing... Sweeping this under the rug makes me rather nervous. Unfortunately the list of changes from 1.3 to 1.4 is huge, so a quick glance through that list didn't raise anything other than the the change in FFT library. The vispy 0.6.4 release notes might offer some clues...?

The changes from #923 that I find significant are the `.show` calls. @sofroniewn why did you need to add those exactly?

tlambert03:
> I find this all very unsatisfying... At the very least, I feel like we should investigate enough to raise an issue at scipy

totally agree, and don't intend to stop investigating and will submit something when I understand it better... the multi-factorial nature makes it hard (it literally depends on python 3.7 instead of 3.6/3.8, pyqt5 instead of pyside2, pytest being imported, _and_ a strange double import in our code).  this just fixes our tests in an harmless way while we can keep looking.  I don't really mean to sweep it under the rug, just let our other PRs pass.

@jni, in case you missed it, see also https://github.com/napari/napari/issues/925#issuecomment-580059473.
the change in vispy in 0.6.4 that causes this issue is simply moving `from skimage.measure import find_contours` from the top of `visuals.isocurve` to a try/catch somewhere in the body of a method in the file (a method that doesn't actually get used in our tests).  I chased the effect of that import as far as I could and it ultimately died at `scipy.stats` (but without any satisfying explanation) ...  I haven't tried other versions of scipy yet but I will.

napari/__init__.py
```diff
@@ -52,3 +52,11 @@
 
 __version__ = get_versions()['version']
 del get_versions
+
+# this unused import is here to fix a very strange bug.
+# there is some mysterious magical goodness in scipy stats that needs
```
jni, line 57: :joy: tbh I kinda love having this comment in our codebase forever. =D

jni (approved):
@tlambert03 got it, thank you for the explainer, good to go as far as I'm concerned!

tlambert03:
Great, will merge this to let our CI pass again, and continue the investigation in #925

Issue #930, opened 2020-01-31
Adding points with properties / empty properties


sofroniewn:
## 🐛 Bug

@kevinyamauchi I'm trying to create a basic demo where I load an empty points layer with a properties dict with an empty list - say named "class", which I can then populate as I add points - leveraging a combination of custom keybindings / mouse functions. The overall goal is to create a simple demo of how to annotate a movie for input into a tracking program like DeepLabCut. The reason I'm doing this is to try and learn about whether our Points / keybindings / mouse functions APIs are enabling real world use cases like that. I'm running into a couple problems though.

First is around initializing the empty points layer:

<details>

```python
"""
Displays empty points with properties
"""

import napari
import numpy as np


with napari.gui_qt():
    # create an empty viewer
    viewer = napari.Viewer()
    viewer.add_image(np.random.random((5, 400, 400)))

    # create properties for each point
    properties = {
        'class': np.array([]),
    }

    # define the color cycle for the face_color annotation
    face_color_cycle = ['red', 'blue', 'green', 'cyan', 'magenta', 'yellow']

    points_layer = viewer.add_points(
        properties=properties,
        size=10,
        edge_width=1,
        edge_color='white',
        face_color='class',
        face_color_cycle=face_color_cycle,
        name='points'
    )
```

```python-traceback
(base) czirwc1macos2701:image-demos nsofroniew$ python test-examples/points_properties_empty.py
Traceback (most recent call last):
  File "test-examples/points_properties_empty.py", line 29, in <module>
    name='points'
  File "/Users/nsofroniew/GitHub/napari/napari/components/add_layers_mixin.py", line 366, in add_points
    visible=visible,
  File "/Users/nsofroniew/GitHub/napari/napari/layers/points/points.py", line 336, in __init__
    self._current_face_color = self.face_color[-1]
IndexError: index -1 is out of bounds for axis 0 with size 0
```

The expected behavior is to be able to instantiate the layer.

</details>

The second error is around adding points when using properties

<details>

```python
"""
Displays points with properties
"""

import napari
import numpy as np


with napari.gui_qt():
    # create an empty viewer
    viewer = napari.Viewer()
    viewer.add_image(np.random.random((5, 400, 400)))

    points = np.array([[100, 100], [200, 200], [333, 111]])

    # create properties for each point
    properties = {
        'class': np.array([1, 2, 3]),
    }

    # define the color cycle for the face_color annotation
[...]

PR #933, opened 2020-01-31
Fix add points

kevinyamauchi:
# Description
This PR fixes the point added bug described in #930. Points were unable to be added without selecting a point first due to an bug in how `Points.current_properties` was initialized. I have fixed the issue and added a test.

## Type of change

# References
This PR addresses the point adding bug described in #930.

# How has this been tested?
- [ ] added a test for adding a point with properties and no points selected

## Final checklist:

### DIFF
--- a/napari/layers/points/points.py
+++ b/napari/layers/points/points.py
@@ -1034,11 +1034,8 @@ class Points(Layer):
             k: np.unique(v[index], axis=0) for k, v in self.properties.items()
         }
         n_unique_properties = np.array([len(v) for v in properties.values()])
-        self.current_properties = {}
-        # only include the properties that common across all selected points
-        for n, prop in zip(n_unique_properties, properties):
-            if n == 1:
-                self.current_properties[prop] = properties[prop]
+        if np.all(n_unique_properties == 1):
+            self.current_properties = properties
 
     def interaction_box(self, index):
         """Create the interaction box around a list of points in view.
--- a/napari/layers/points/points.py
+++ b/napari/layers/points/points.py
@@ -336,7 +336,7 @@ class Points(Layer):
         self._current_face_color = self.face_color[-1]
         self.size = size
         self.current_properties = {
-            k: v[-1] for k, v in self.properties.items()
+            k: np.asarray([v[-1]]) for k, v in self.properties.items()
         }
 
         # Trigger generation of view slice and thumbnail
--- a/napari/layers/points/_tests/test_points.py
+++ b/napari/layers/points/_tests/test_points.py
@@ -300,6 +300,9 @@ def test_properties():
     layer = Points(data, properties=copy(properties))
     assert layer.properties == properties
 
+    current_prop = {'point_type': np.array(['B'])}
+    assert layer.current_properties == current_prop
+
     # test removing points
     layer.selected_data = [0, 1]
     layer.remove_selected()
@@ -344,13 +347,13 @@ def test_adding_annotations():
     shape = (10, 2)
     np.random.seed(0)
     data = 20 * np.random.random(shape)
-    annotations = {'point_type': np.array(['A', 'B'] * int((shape[0] / 2)))}
+    properties = {'point_type': np.array(['A', 'B'] * int((shape[0] / 2)))}
     layer = Points(data)
     assert layer.properties == {}
 
     # add properties
-    layer.properties = copy(annotations)
-    assert layer.properties == annotations
+    layer.properties = copy(properties)
+    assert layer.properties == properties
 
     # change properties
     new_annotations = {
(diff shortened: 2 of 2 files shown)


sofroniewn:
First bug is fixed, but I can still find a new one, around adding points, deleting points, then trying to add new ones.

<details>

code snippet is

```python
"""
Displays points with properties
"""

import napari
import numpy as np


with napari.gui_qt():
    # create an empty viewer
    viewer = napari.Viewer()
    viewer.add_image(np.random.random((5, 400, 400)))

    # create properties for each point
    properties = {
        'class': np.array([1, 2, 3]),
    }

    # define the color cycle for the face_color annotation
    face_color_cycle = ['red', 'blue', 'green', 'cyan', 'magenta', 'yellow']

        points,
        properties=properties,
        size=10,
        edge_width=1,
        edge_color='white',
        face_color='class',
        face_color_cycle=face_color_cycle,
        name='points'
    )
```

Add some points, then select some points delete them, and then try adding some more and I get the following error

```python-traceback
(base) czirwc1macos2701:image-demos nsofroniew$ python test-examples/points_properties_add.py
  File "/Users/nsofroniew/GitHub/vispy/vispy/app/backends/_qt.py", line 447, in mousePressEvent
  File "/Users/nsofroniew/GitHub/vispy/vispy/app/base.py", line 181, in _vispy_mouse_press
    ev = self._vispy_canvas.events.mouse_press(**kwargs)
    self, cb_event=(cb, event))
    cb(event)
  File "/Users/nsofroniew/GitHub/napari/napari/_qt/qt_viewer.py", line 338, in on_mouse_press
    self.layer_to_visual[layer].on_mouse_press(event)
  File "/Users/nsofroniew/GitHub/napari/napari/_vispy/vispy_base_layer.py", line 189, in on_mouse_press
    self.layer.on_mouse_press(event)
  File "/Users/nsofroniew/GitHub/napari/napari/layers/points/points.py", line 1513, in on_mouse_press
    self.add(self.coordinates)
  File "/Users/nsofroniew/GitHub/napari/napari/layers/points/points.py", line 1334, in add
  File "/Users/nsofroniew/GitHub/napari/napari/layers/points/points.py", line 385, in data
    self.current_properties[k], adding, axis=0
KeyError: 'class'
```

</details>

sofroniewn:
I'll also note - now with the mapped variable, it's much harder to say adjust the class variable from the GUI - for example it would be nice to have a dropdown that would let me adjust which class I am adding points to. Changing the color could then change the color of the entire class or something? Not quite sure how to achieve this in the GUI in a generic way, but it would be nice!!!

GUI element could depend on the color mode or if a property is being used .... Would make the DeepLabCut annotation case that I describe in #930 much easier too. This would be a new PR though

kevinyamauchi:
> First bug is fixed, but I can still find a new one, around adding points, deleting points, then trying to add new ones.

Good catch, I fixed it. We had added a behavior where selecting points with different properties would set `Points.current_properties` to the properties that the points have in common. However, this can drop the property being used to set a color, so I reverted it back to the "only update when selected points have the same properties" behavior. I'll spend some more time playing around with it tonight to make sure there aren't any more lurking bugs.

> I'll also note - now with the mapped variable, it's much harder to say adjust the class variable from the GUI - for example it would be nice to have a dropdown that would let me adjust which class I am adding points to. Changing the color could then change the color of the entire class or something? Not quite sure how to achieve this in the GUI in a generic way, but it would be nice!!!

I totally agree. My plan was to get all of the features into the model and then figure out the best way to interact with them via the GUI. I was thinking along the same lines as you and I am definitely open to suggestions!

napari/layers/points/points.py
```diff
@@ -1034,11 +1034,8 @@ def selected_data(self, selected_data):
             k: np.unique(v[index], axis=0) for k, v in self.properties.items()
         }
         n_unique_properties = np.array([len(v) for v in properties.values()])
-        self.current_properties = {}
-        # only include the properties that common across all selected points
-        for n, prop in zip(n_unique_properties, properties):
-            if n == 1:
-                self.current_properties[prop] = properties[prop]
+        if np.all(n_unique_properties == 1):
+            self.current_properties = properties
```
jni, line 1038: Wait, can't we fix this in the same way as last time, but perhaps with more checks?

kevinyamauchi: The challenge is that if two points with different properties are selected, it isn't clear how to resolve the conflicting properties to set `current_properties`without causing an error when subsequently adding new points. For example, if we have the points layer below

```python
data = [[100, 200], [200, 200]]
properties = {
    'point_class': np.array(['A', 'B']),
    'radius': np.array([6, 6]),
}

layer = Points(
    data,
    face_color = 'point_class'
    properties=properties)
)

# select both points
layer.selected_data = [0, 1]
```
As we had it before, we give `current_properties`the properties that all of the selected points have in common. That is, 

```python
>>> layer.current_properties
{'radius': np.array([6])}
```
If we then add a point, we will get the error [described by Nick above](https://github.com/napari/napari/pull/933#issuecomment-580944251), because `layer.current_properties` does not have a `point_class` value (the selected points had different values for `point_class`).

I had considered resolving the conflicting properties by taking the first of the conflicting values (or something like that), but that didn't seem intuitive to the user (i.e., they wouldn't easily be able to figure out how the property is chosen).

I think another option could be to have the user define a fall back value for the properties, but I think we should avoid adding another property if we can.

It seemed like the most straightforward thing was to revert to the original behavior where it just doesn't overwrite the value of `Points.current_properties` if the selected points conflict (it's how the other properties work). All of that being said, I am open to other suggestions! What do you think, @jni?

sofroniewn (approved):
I've checked this out - all looks good to me, thanks @kevinyamauchi!

napari/layers/points/points.py
sofroniewn: I think not overwriting is fine for now. I think we can revisit this when the think about how to represent the properties in the GUI / change properties for new points

kevinyamauchi: That works for me. @jni, do you have any ideas or a strong preference against merging as is?

jni: Yeah, ok, I see the issue clearly now. We can revisit later.

PR #935, opened 2020-02-01
[WIP] Plugin hookimpl tests

tlambert03:
# Description
This will only go in after #908, (so at the moment, there's a lot of code here that is merged from #908). but it's distinct enough to keep it as a separate PR, and I wanted to have it up here for conversation.

This PR adds a framework for "hook implementation tests" or "validators".  While pluggy provides built-in checks to make sure that `hookimpls` have signatures that match their corresponding `hookspecs` (for instance a plugin cannot implement a `hookimpl` with more arguments than the corresponding `hookspec`), one might also want to confirm that `hookimpls` meet some _functional_ requirements as well: for instance, that they return the proper type, or that they execute within a certain amount of time.

As an example: the `napari_get_reader` hook gives the plugin a `path` string and expects it to _quickly_ return either `None` or a `callable`.  To check this, we 
1) declare [a test function in `napari/plugins/validators.py`](https://github.com/tlambert03/napari/blob/8377022d16455b6c4b5cc83437b41feb313b8c2e/napari/plugins/validators.py#L38-L39), and decorate the function it with `@validates('napari_get_reader')`

    > _all validator functions will accept a single argument with type_ `pluggy.hook.HookImpl`
2) we then register all of our validators with the [`PluginManager.add_hook_validators`](https://github.com/tlambert03/napari/blob/8377022d16455b6c4b5cc83437b41feb313b8c2e/napari/plugins/manager.py#L90-L107) method, passing it our `validators` module. (It will look for decorated functions).

3) The plugin manager will then either greedily validate new hook implementations upon registration (if `greedy_validation == True`), or on demand, using [`PluginManager.validate_hookimpls`](https://github.com/tlambert03/napari/blob/8377022d16455b6c4b5cc83437b41feb313b8c2e/napari/plugins/manager.py#L126)

4) currently, if a hookimpl fails validation, it is disabled for that particular hook (though other hooks from the same plugin are not affected).  We could also just emit a warning though...

Obviously, actually executing these validation functions incurs some time cost.  So it's an open question where best to implement them.  For hooks that we expect to be very fast, we may want to greedily validate registered hook implementations .  For other hooks, this may be the kind of thing that a plugin developer would import and use to make sure that their plugins are working as we _functionally_ intended the spec to perform (and not just that their signature is correct).

### todo:
- make `greedy_validation` accept a list of specnames rather than a boolean.

## Type of change

# How has this been tested?

- [x] added a test plugin fixture (`napari_bad_plugin`) that matches the `napari_get_reader` API, but performs too slowly.  The `test_manager.test_hookimpl_validation` test asserts that it is removed from the `hookimpls` for that `hookspec`.

## Final checklist:


jni:
@tlambert03 in the abstract this all sounds great, but could you elaborate a bit more on what the experience is as a user? How do I use this framework to test things if I'm writing my own plugin?

Re: greedy vs just in time, I guess as long as we have caching of the results, I don't think this is super important...?

tlambert03:
the end-user probably doesn't see/experience any of this... except when a badly-behaving plugin triggers a validation error and gets inactivated.  The "high-level" goal here is to protect the napari experience by establishing a place to put functions that can check how hook implementations are performing (_outside_ of the actual places in napari where the hooks are called).

from the perspective of plugin developers, these functions are sort of like tests that _we_ (napari) write _for_ them.  in code, that might look kind of like:

```python
from napari.plugins import NapariPluginManager, HookImplementationError
import my_plugin_module

# this will automatically include any of the validators we put in `plugins/validators.py`
pm = NapariPluginManager(autodiscover=False)
# developer registers their local module
pm.register(my_plugin_module)
try:
    pm.validate_hookimpls()
except HookImplementationError as e:
    print(f"Sorry, please play again! {e}")
```

sidenote: I'm not 100% sure this is necessary or the right way to implement it! 😅 so feel free to be extra critical on this one.  It's not essential for moving forward.

sofroniewn:
> The "high-level" goal here is to protect the napari experience by establishing a place to put functions that can check how hook implementations are performing (outside of the actual places in napari where the hooks are called).

I like this goal. I think testing as much as possible when someone first adds a plugin seems worth it to prevent unexpected surprises further down the line

tlambert03:
this doesn't negate the purpose/usefulness of this PR... but just also wanted to mention here that pytest provides an optional plugin called [pytester](https://docs.pytest.org/en/latest/writing_plugins.html#testing-plugins) (basically just some useful fixtures) that is intended to help developers test their plugins.

Issue #539, opened 2019-09-17
Tracks lines for representing trajectories in nD+t


royerloic:
## 🚀 Feature

A typical bioimage analysis task is to track the movement of objects of interest such as cells in their 2D or 3D movement (possibly more, @jni --guardian-of-dimensions-- would argue). Having the possibility to display such tracks both in the 2D and 3D view and show future and past positions (increasingly faint and transparent as you go away from the current time point) would be extremely useful.  The possibility to edit such tracks and adjust the positions of objects would be cool too to refine the output of automated algorithms. Dealing with object division can be done by ending a track and restarting two tracks. Basic styling of the thickness, opacity of the tracks, and how far back they extend in the past or future, would be great too. 

Tracks could be represented as a collection of n*k arrays where n is the length of the track, and k is the dimension of the space occupied by the objects.

sofroniewn:
@royerloic how do you feel this functionality intersects with the current `Path` object in the shapes layer? For example see this WIP PR #532 for examples of lines in 3D? You can pass a list of NxD arrays as input and a list of colors and it will do 2D and 3D rendering. It won't change transparency along the line, this could be done, but unclear if it is needed functionality. There's also only one global width for each line right now.

royerloic:
Yes this looks like a good starting point!
One possibility would be to subclass 'paths' to 'tracks'
adding features specific to spatio-temporal tracking as 
described above. This is something we could work on with
@pranathivemuri . Could be exciting feature.

sofroniewn:
There might be some general interest in more advanced coloring options for the paths from other communities too, I thinking about some of the track tracing people from diffusion tensor imaging for example, https://dipy.org/, and I've been chatting with @arokem about some example datasets from that space too, but starting with subclassing sounds good, and I'd be excited to get more contributions from @pranathivemuri!! The 3D stuff is really coming along now. 

Unfortunately one things that we're coming up against is that right now the way the Shapes layer is architected has a lots of generality but at the expense of some performance issues, so we struggled with 1000s of paths for example, but that could be improved with some profiling and better design decisions maybe

royerloic:
Sounds good. 
My guess is that you have a list of vertex/face data , one for each shape or so...
Whereas for performance you would want to have it all packed into single arrays...
Is that the case? We can also just tailor-made a layer just for tracks that could handle 
tens of thousands of tracks...  Subclassing might not be the right way then, because
it creates shallow dependencies and prevents performance optimisation....
We should think about it...

[...]

PR #937, opened 2020-02-02
Hook up reader plugins

tlambert03:
# Description
This PR requires #908, and hooks up `napari_get_reader` functionality offered by plugins and our own `builtins`.

Based on https://github.com/napari/napari/pull/908#discussion_r371938674 from @sofroniewn, I also created a new viewer mixin `add_files`, and a corresponding convenience `view_files` method.  That unifies the places in napari where file io happens, and makes for (currently) a single place where reader plugins need to be hooked up ([in `Viewer.add_files`](https://github.com/napari/napari/blob/2605715bf9abf2611746625f4166649b67b80bdb/napari/components/add_layers_mixin.py#L75)).

I also modified `__main__.py` to use `view_files` rather than directly calling magic_imread (this will allow reader plugins to more easily add different types of layers from the CLI.

**Note**: I modified the default file-stacking behavior to be opt-in.
- On the CLI, you now have to use `--stack` to concatenate multiple files into a stack.
- In the viewer menubar, to concatenate multiple individual files, you now have to use `File > Open image series as stack...`
`File > Open Image(s)...` will treat individual files as layers (as will dragging and dropping multiple files onto the viewer).
- Directories will be stacked as usual when possible (unless intercepted by some other plugin).  I believe @jni set up the previous behavior, so I definitely want his input on that.
- I also want @jni to comment on [the changes in `__main__.py`](https://github.com/napari/napari/blob/2605715bf9abf2611746625f4166649b67b80bdb/napari/__main__.py#L81-L101) where certain `dtypes` were autodetected as labels.  We can still do something like that, but I'd like to revisit how we determine whether something is a label.

Everything is working at the moment, but I still need to add some more stringent tests, and try to break it with bad plugins, etc...  Will probably need some more try/catch statements.  (and may also have implications/need for #935).  That said, it would be good to get some general input on the structure as it currently stands.

## Type of change
- [x] New feature - **this is a breaking change for the command line interface** and changes the stacking behavior of drag/drop in the viewer.

# How has this been tested?

- [x] want to add more tests.

## Final checklist:


tlambert03:
if you'd like to try the reader plugin in action, here's a working napari plugin that I made for reading deltavision/mrc files using my [mrc](https://github.com/tlambert03/mrc) package: https://github.com/tlambert03/napari-dv

to try it out:
- checkout this PR
- `pip install napari-dv`
- then download [this test file](https://github.com/tlambert03/napari-dv/blob/master/toxo.dv?raw=true), open napari, and drag and drop it onto the viewer.

sofroniewn:
Pretty cool @tlambert03 - I can confirm this works :-)
Screen Shot 2020-02-02 at 3 26 37 PM

sofroniewn:
I'm not 100% about the `add_files` API right now. It's a pretty different function then `add_image` or `add_surface` say, but it sounds like them. Right now you can pass a `path` to `add_image` too, which is maybe then confusing too - as that path won't pass through our io plugins. Maybe that's a mistake though, maybe it should, or maybe we should drop the path support there.

I'm also even less sure about `view_files` - I appreciate it was added for the symmetry with `add_files`, but I'm not really sure its needed or that we should be trying to make this command like our layer commands.

One option would be to rename `add_files` to `read_files` and drop `view_files`. I think that makes it clear that this is really a different way of interacting with the viewer. It might be nice to be able to specify an io-plugin by name `read_files(..., plugin=None)` or something like that.

I'm also not quite sure how to resolve the `stack` right now, but I don't particularly like it's special casing

tlambert03:
> I'm not 100% about the `add_files` API right now. It's a pretty different function then `add_image` or `add_surface` say, but it sounds like them. Right now you can pass a `path` to `add_image` too, which is maybe then confusing too - as that path won't pass through our io plugins. Maybe that's a mistake though, maybe it should, or maybe we should drop the path support there.

I would certainly go for removing `path` from `add_image`.  I'm not sure I totally agree that `add_files` is all that different from `add_image` or `surface`.  Each of these methods expects a certain data type (_not_ a layer: `add_image` expects a numpy array.  `add_surface` expects a 3-tuple, and `add_file` expects a string, filepath, URL, or any resource locator).  and the `add_*` methods just convert it into the appropriate layer type to add to the viewer.  That's exactly what `add_file` is doing here... just via the plugins, which in turn will call `add_layer_from_data`.  In other words, we don't really know what type of layer a file might represent... but external/internal plugins do.

> I'm also even less sure about `view_files` - I appreciate it was added for the symmetry with `add_files`, but I'm not really sure its needed or that we should be trying to make this command like our layer commands.

can remove.  But I think the same argument as above applies.  Down the road, this would actually allow someone to make a custom file type (with some random extension), and some plugin could know how to create N layers from that file and add them all to the viewer, all with `view_file()`.  

> One option would be to rename `add_files` to `read_files` and drop `view_files`. I think that makes it clear that this is really a different way of interacting with the viewer. It might be nice to be able to specify an io-plugin by name `read_files(..., plugin=None)` or something like that.

I'm open to this.  but I think `read_files` is less clear.  what do you do with them next if not view them somehow?  (and I am still not convinced this is really a different way of interacting with the viewer).

> I'm also not quite sure how to resolve the `stack` right now, but I don't particularly like it's special casing.

i actually felt like stacking multiple files on top of each other is _more_ of a special case then not as it assumes that all the files have the same shape, etc...  What if someone just wants to open two files at the same time?  (that currently works... but _only_ because it defaults to the shape-agnostic dask viewer, and it would not work if one was not an image type).  the argument for the the new arrangement is that it allows two independent files to be opened, it allows plugins a go at each file, and also enables stacking if that's the explicit desire (and of course, if they just drag a folder of files, stacking is still the default behavior).

sofroniewn:
> I would certainly go for removing path from add_image.

I have found this quite useful though - and was planning to add `path` support to all our layer types to give them some default IO. It's not quite the same functionality as provided by going through one of our reader plugins.

I guess we're also moving away from the 1-1 relationship between the `add_*` methods and the layers as we talk about things like `add_tracks` in #539 and have the `channel_axis` keyword arg in `add_image`, so maybe `add_files` is just one more step in that direction, but I still see it different because you don't know what layer type you will get out. I can see how `read_files` is maybe not so great because you do end up adding them.

If we go with `add_files` What do you think about?

> It might be nice to be able to specify an io-plugin by name add_files(..., plugin=None) or something like that.

Is that a reasonable concern, of just confusing even more.

> the argument for the the new arrangement is that it allows two independent files to be opened, it allows plugins a go at each file, and also enables stacking if that's the explicit desire (and of course, if they just drag a folder of files, stacking is still the default behavior).

sorry what do you mean by "new arrangement"? Do you mean the `if stack` and `io.magic_imread` combo or something about the new plugin architecture?

tlambert03:
> > I would certainly go for removing path from add_image.
> 
> I have found this quite useful though - and was planning to add `path` support to all our layer types to give them some default IO. It's not quite the same functionality as provided by going through one of our reader plugins.

lol, then why did you plant that idea in my head! :)  but i agree, we can't get rid of it just without a substitute: If someone _knew_ that a filepath represented an image, and wanted to provide keyword arguments to the layer creation, `add_image` with `path` would be the only way to do it (because otherwise plugins control the layer args...).  And along those lines, perhaps it _would_ be nice to add to other `add_*` methods.

Also, I think want to rename `add_files` to `add_path`, since it could be a file, directory, or URL (URLs are another thing I'd like to discuss... for another issue).  that ok?

> It might be nice to be able to specify an io-plugin by name add_files(..., plugin=None)

I _do_ like that idea, and was also thinking about how to enable forcing a plugin (in addition to allowing the user to reorder the plugins... which I've made progress on).  I think it fits nicely on `add_files`.  I would would have it raise an exception on failure...

> what do you mean by "new arrangement"?

sorry, this `stack` conversation might require a step back and a fresh start 😅.  what I meant is that the "old" behavior is that when multiple files are passed to napari (either on the CLI or a multi-file drop), then napari assumes that they are all part of the same stack.  by "new arrangement", I mean just assuming that multiple files are _not_ part of the same stack unless explicitly instructed (with the 
`--stack` flag on the CLI, `add_files(stack=True)` in the console, or `Open images as stack` in the gui).  but let me know if I just missed your original point when you said:
>I'm also not quite sure how to resolve the stack right now, but I don't particularly like it's special casing

sofroniewn:
> Also, I think want to rename add_files to add_path, since it could be a file, directory, or URL (URLs are another thing I'd like to discuss... for another issue). that ok?

I prefer `add_path` - it actually helps things be more consistent with words we've used elsewhere in the API. I also think it's then fine to keep `path` as a keyword arg to the other `add_*` methods too.

I'll have to think about the stack stuff more

tlambert03:
I added a widget that will generally let us rearrange the call order of plugins after registration.  Pluggy doesn't internally offer that sort of rearrangement, so I made a [new function for it](https://github.com/napari/napari/blob/f54d899ecd2495b6440617db3840cb3efae7dd00/napari/plugins/manager.py#L116), and hooked it up to a `QListWidget` so users can drag/drop their preferred plugin call order.  Here's the general idea (styles can/will definitely change and we should discuss the terminology we want to use for front-facing GUI stuff):

Untitled

jni:
Hey guys!

This is super cool! Sorry, I don't have time for a deep dive into everything, but I did want to say that I don't have a strong opposition to any of the views expressed here, so if you need to move ahead without my blessing please do so. (Though probably this would ideally wait for a discussion at lab meeting.)

On `add_path` vs `add_image(path=...)`, I suggest we keep the existing method for now. I'm agnostic about the new API, I see all the arguments in all the different directions, but am leaning towards `viewer.read_file[s]`. `view_path`... Eh. I don't love it. But I see the appeal.

Anyway, whatever you guys decide is ok with me, but again, perhaps it'd be good to get some input in group meeting. The whole `view`, `Viewer`, `ViewerModel` thing was decided in a brainstorming session and I'm honestly pretty happy about it. I think it's superior to all the ideas that were floated before it, even if we are now pushing on its boundaries. So I really like the whole group to weigh in on API naming issues.

@tlambert03 what exactly did you want to discuss about the labels by dtypes? Have you removed that behaviour by default? I just thought that int32 and higher are such rare dtypes for intensity images that it was a safe enough default assumption. Have you hit this limit? And what is the proposed alternative behaviour?

tlambert03:
> @tlambert03 what exactly did you want to discuss about the labels by dtypes? Have you removed that behaviour by default? I just thought that int32 and higher are such rare dtypes for intensity images that it was a safe enough default assumption. Have you hit this limit? And what is the proposed alternative behaviour?

i agree it's a safe enough assumption.  Mostly I wanted to discuss _where_ that code lived.  the only place I found that was in `__main__`, so opening `int32` files on the command line would generate labels, but dragging into the viewer wouldn't.  I've got no problem with the auto-label behavior, but I'd like to unify it.  I would propose moving that logic to `_add_layer_from_data`... where `layer_type` defaults to `image`... but would now be `labels` for those dtypes. Since that method sits at the end of our builtin reader plugin, it would be global behavior, rather than CLI behavior.

**edit**: 
I moved it to `_add_layer_from_data` and it looks like this: 
```python
        if layer_type == 'image':
            if data.dtype in (np.int32, np.uint32, np.int64, np.uint64):
                layer_type == 'labels'
```

napari/__main__.py
```diff
@@ -81,25 +81,6 @@ def main():
...
-        #                 np.uint32,
-        #                 np.int64,
-        #                 np.uint64,
-        #             ):
-        #                 v.add_labels(layer)
-        #             else:
-        #                 v.add_image(layer, rgb=args.rgb)
-        #     else:
-        #         v.add_image(images, rgb=args.rgb)
-
```
AhmetCanSolak, line 102: :D

napari/plugins/_hookexec.py
```diff
@@ -0,0 +1,110 @@
...
+        except BaseException:
+            excinfo = sys.exc_info()
+    finally:
+        if firstresult:  # first result hooks return a single value
+            outcome = _Result(results[0] if results else None, excinfo)
+        else:
+            outcome = _Result(results, excinfo)
+
+        # run all wrapper post-yield blocks
+        for gen in reversed(teardowns):
```
shanaxel42: Why do you reverse this list? And if so, why not make teardowns a stack instead of a list so you don't need to

napari/plugins/tests/test_reader_plugins.py
```diff
@@ -0,0 +1,64 @@
...
+    # the plugin loads fine, so there should be no exceptions yet.
+    assert 'napari_bad_plugin2' not in plugin_manager._exceptions
+
+    # but when we try to read an image path, it will raise an IOError.
+    # we want to catch and store that IOError, and then move on to give other
+    # plugins chance to return layer_data
+    layer_data = get_layer_data_from_plugins('image.ext', plugin_manager)
+
+    # the good plugins (like "napari_test_plugin") should return layer_data
+    assert layer_data
```
shanaxel42: It might get kind of confusing if we just silently abstract away error messages from the users. They may believe a particular plugin is working when it's not, or not fully understand which plugin did execute. Maybe we could add a Warning when this happens and let the user know that "XYZ" plugin failed to load their data but another was successful?

napari/plugins/utils.py
```diff
@@ -0,0 +1,153 @@
...
+    if attr:
+        hookattrs = [getattr(hookimpl, attr) for hookimpl in hookimpls]
+    else:
+        hookattrs = hookimpls
+
+    # find the current position of items specified in `order`
+    indices = []
+    seen: Set[HookOrderType] = set()
+    for i in order:
+        if i in seen:
```
shanaxel42: I find the extra `seen` check within this loop a little confusing. You could add an assert check at the beginning of this method below line 52 that raises a ValueError if `len(your_list) != len(set(your_list))`. This isn't really a optimization or anything though just my personal preference so feel free to ignore : )

napari/components/add_layers_mixin.py
```diff
@@ -695,8 +746,8 @@ def add_vectors(
         return layer
 
     def _add_layer_from_data(
-        self, data, meta: dict = None, layer_type: str = 'image'
-    ):
+        self, data, meta: dict = None, layer_type: Optional[str] = None
+    ) -> Union[layers.Layer, List[layers.Layer]]:
```
shanaxel42, line 788: Types! : D

shanaxel42 (commented):
Seems overall chill, nice! Had a couple questions... although the one thing I'm wondering is if all the added ability to reorder the plugins call order is needed? what prompted you to add that functionality, were you imagining a scenario in which someone would have multiple reader plugins installed and want to be specific about which to call first?

napari/plugins/_hookexec.py
tlambert03: I didn't write that... that bit of code is vendored [directly from pluggy](https://github.com/pytest-dev/pluggy/blob/6e8b6d4e8e0519e50af05b4e4b90b18b924bcbcc/src/pluggy/callers.py#L201-L206) while I [wait for them to consider a PR I submitted](https://github.com/pytest-dev/pluggy/pull/253) that allows you to know _which_ plugin actually returned the result.  (sorry, I thought I had put that in the module docstring, but I guess not!).   So I'm just trying to keep it as similar to their code as possible....
But, they reverse the list because they want wrappers to "sandwich" the call (wrpA, wrpB, hookA, hookB, wrpB, wrpA)

napari/plugins/tests/test_reader_plugins.py
tlambert03: sorry, by "catch and store" I didn't mean to imply it doesn't get reported. A warning _does_ get logged anytime there is a plugin error.  the logic is `plugin.manager. log_plugin_error` and an example of what the warning looks like is here: https://github.com/napari/napari/pull/908#discussion_r380187837
It _also_ gets stored for later retrieval so that they can easily submit a full bug report with traceback to the developer.  I'm working on a full interface for that in another PR.

napari/plugins/utils.py
tlambert03: yeah that's better. I think when I was writing this there was some reason why I was trying not to iterate that list too many times (and a greedy check like that causes one more loop right?) ... but now I can't remember why that was! 😅

tlambert03 (commented):
thanks for taking a look!

>what prompted you to add [the ability to reorder the plugins call order], were you imagining a scenario in which someone would have multiple reader plugins installed and want to be specific about which to call first?

exactly. I definitely imagine that a typical user might end up with many different reader plugins installed, and it's quite possible that more than one of them will "recognize" a given file-type/format.  Without that logic, the second plugin in line would be literally inaccessible.  It's also conceivable that one plugin would offer a subset of formats of another one that should be "tried first" and fall back to the second plugin.  But that requires the ability to reorganize call order after registration (which pluggy doesn't provide built in).

evamaxfield:
Thanks @sofroniewn for letting me know about this PR. From what I can tell from parsing through the code, I think it would be _decently easy_ to drop in an [aicsimageio](https://github.com/AllenCellModeling/aicsimageio) reader plugin once this is merged.

If [this bad plugin](https://github.com/tlambert03/napari/blob/plugin-reader/napari/plugins/tests/fixtures/invalid_plugin.py) was a spec to copy, we pretty much work as we already have a [`determine_reader`](https://github.com/AllenCellModeling/aicsimageio/blob/master/aicsimageio/aics_image.py#L130) and then subsequent readers `dask_data` and `data` attributes.

Question: Does the plugin need to return a `numpy.ndarray` or can it be both a `dask.Array` and `numpy.ndarray`? I would assume both but just curious. Either way, excited to give it a try.

I am not too familiar with the pluggy interface so if I could ask for anything it would be an example fake plugin be added as a part of documentation. I can probably figure it out given time but would help a lot if possible.

EDIT: Clarifying above, my question is basically, would we need to release `aicsimageio-napari-reader` as a python package for pluggy to pick it up? How does that process work?

tlambert03:
Hi @JacksonMaxfield, sorry that the documentation is lagging behind the PR, but I promise as soon as this merges @jni and I will coordinate on making extensive docs.  In the meantime here are some direct answers:

>If this `bad plugin` was a spec to copy, we pretty much work as we already have a determine_reader

the "source of truth" for the API that plugins need to follow will live in [`napari.plugins.hook_specifications`](https://github.com/napari/napari/blob/master/napari/plugins/hook_specifications.py).  There you will find a list of function signatures that plugins may implement.  In this case, you're looking for the `napari_get_reader` hook.  (more below...) 

>Question: Does the plugin need to return a numpy.ndarray or can it be both a dask.Array and numpy.ndarray? I would assume both but just curious.

The api is actually rather specific here:
- plugins must implement a function that satisfies the `napari_get_reader` hook specification, and decorate it with a pluggy hook implementation decorator (example below, pluggy docs [here](https://pluggy.readthedocs.io/en/latest/#implementations))
- looking at the hook specification for `napari_get_reader`, you'll see it's a function that takes a string (the path to read) and either return `None` (if the plugin does not recognize the file), or, if the plugin recognizes the path as supported, returns a _new_ function that could be called to actually read the file.
- _that_ function must be of type `ReaderFunction` shown below: it takes a `str` and returns a `list`, where each item in the list is a 1-, 2-, or 3-`tuple` of arguments that could be passed to [`Viewer._add_layer_from_data`](https://github.com/napari/napari/blob/323c24c1d06378dcf2f654033c90bcdcac8f1a58/napari/components/add_layers_mixin.py#L697)

... and to answer your question directly, yes, the first item in the tuple can be dask or numpy (or zarr), etc... as long as it is valid for the corresponding `Viewer.add_*` function.

https://github.com/napari/napari/blob/323c24c1d06378dcf2f654033c90bcdcac8f1a58/napari/plugins/hook_specifications.py#L29-L30


>I am not too familiar with the pluggy interface so if I could ask for anything it would be an example fake plugin be added as a part of documentation. I can probably figure it out given time but would help a lot if possible.

here is an example _real_ plugin that you can look at: https://github.com/tlambert03/napari-dv, or create your own from this cookiecutter template: https://github.com/napari/cookiecutter-napari-plugin  (there too, docs are still forthcoming)

but the bare minimum would look like this:
```python
import pluggy

napari_hook_implementation = pluggy.HookimplMarker("napari")

def reader(path):
    my_array = some_internal_reader_function(path)
    return [(my_array,)]

@napari_hook_implementation
def napari_get_reader(path):
    if i_can_read_this(path):
        return reader
```

>EDIT: Clarifying above, my question is basically, would we need to release aicsimageio-napari-reader as a python package for pluggy to pick it up? How does that process work?

napari will find plugins using either "naming convention" or "package metadata" described here.
https://packaging.python.org/guides/creating-and-discovering-plugins/

1. _naming convention:_ any modules in the environment that start with `napari_` will be searched for functions decorated as hook implementations
2. _package metadata_: all distributions in the environment that declare an entry point `napari.plugin` will have each of their declared entry points searched for hook implementations. 

again... no one is expected to know all this just yet, and we'll definitely make all this clearer very soon!

tlambert03:
oh... and your `determine_reader` staticmethod is basically ready to go as a `napari_get_reader` hook implementation.  Except that it needs to return `None` instead of raising an exception

evamaxfield:
Thanks for really clear response 🙂 

The cookiecutter is _massive_ help already so it's awesome to see you have thought about setting that up and will likely be the route we go. Thanks again!

Will keep an eye on this PR and once merged will schedule some time to use the cookiecutter and attempt to deploy a plugin.

tlambert03:
sounds good!  Just to point out an alternative to the cookiecutter for you... since you already have a pip installable package, if you wanted to allow someone to just `pip install napari aicsimageio`, and didn't want to have to maintain a separate pypi package, you could do so like this:
```python
# your setup.py

setup(
    ...
    entry_points: {'napari.plugin': ["aicsimageio_reader=aicsimageio.readers._napari"]}
    # where 'aicsimageio_reader' is the name of your plugin in napari ... 
    # you can name it whatever
    ...
)


# new file at aicsimageio/readers/_napari.py

from pluggy import HookimplMarker
from aicsimageio import AICSImage

napari_hook_implementation = HookimplMarker("napari")


@napari_hook_implementation
def napari_get_reader(path):
    # note, this function should return as fast as possible...
    try:
        reader_class = AICSImage.determine_reader(path)
        # ... convert reader_class into a function that takes a path
        # and returns the appropriate LayerData type
        return modified_reader
    except UnsupportedFileFormatError:
        return None
```

evamaxfield:
Now we're talking 🎉 

I _think_ because we have `napari` as an extra dep people could even do `pip install aicsimageio[interactive]` and they would get the plugin? But dang, that's a cool tip.

tlambert03:
> I _think_ because we have `napari` as an extra dep people could even do `pip install aicsimageio[interactive]` and they would get the plugin?

indeed!

napari/plugins/_builtins.py
```diff
@@ -23,4 +23,5 @@ def napari_get_reader(path: str):
     callable
         function that returns layer_data to be handed to viewer._add_layer_data
     """
-    return lambda path: [(None, {'path': path})]
+    if path.endswith(('tif', 'png', 'jpg')):
```
tlambert03: mostly just a placeholder... see my main review comment about handling "fallback behavior".  If we _do_ want to implement the "when-all-else-fails" behavior as our own internal plugin, then we would probably remove any conditional here and return to the way it is in `master` where this function simply returns `lambda path: [(None, {'path': path})]` (which just tells `add_layer_from_data` to call `add_image(path=path)`).

The alternative is to let `get_layer_data_from_plugins` just return nothing, and put the fallback logic (which would basically be our current `io.magic_read` logic) in line 95 in `add_layers_mixin` where the `print()` statement currently is.

napari/components/add_layers_mixin.py
```diff
@@ -45,6 +47,55 @@ def add_layer(self, layer):
...
+        # iterate through each path provided, looking for a suitable reader
+        for path in paths:
+            layer_data = get_layer_data_from_plugins(path)
+            if layer_data:
+                for data in layer_data:
+                    added = self._add_layer_from_data(*data)
+                    added = added if isinstance(added, list) else [added]
+                    added_layers.extend(added)
+            else:
+                print(f"No reader found for {path}")
```
tlambert03: this is where we would put fallback logic if we didn't implement it as a builtin plugin

tlambert03 (commented):
I'm taking the WIP out of the name so @napari/core-devs feel free to start reviewing.  some small things might change, but i'd like to start getting feedback so I can implement requested changes too.
There is one remaining open question for me though, which is how we want to handle "fallback" behavior (i.e. when no plugins claim the `path`) . I see two (only subtly different) options:
1. we implement our internal fallback behavior as just another `napari_get_reader` implementation... it or would simply perform the current logic which is just to call `add_image(path=path)`.  This is nice because then "literally every reader is a plugin".
2. we _don't_ implement the fallback as just another plugin (though we could definitely create other internal reader plugins)... but rather just put the fallback behavior outside of a hook implementation, with something like our current `add_image(path=path)`.  While the end result is quite similar to 1, this allows us to treat that very last fallback a little differently (perhaps with a try/catch that says "nothing could read it... maybe you'd like to get a plugin."...)

I'm struggling to explain myself succinctly here, so take a look at the code and see if it's clear what I mean.

tlambert03:
just fyi, i touched up the internal reader plugin (it's extremely minimal), so this just needs review at this point.

napari/_qt/qt_main_window.py
```diff
@@ -172,6 +180,17 @@ def _add_window_menu(self):
         self.window_menu = self.main_menu.addMenu('&Window')
         self.window_menu.addAction(exit_action)
 
+    def _add_plugins_menu(self):
+        order_plugin_action = QAction("Plugin call order...", self._qt_window)
```
sofroniewn: Is there a reason you put `...` in the dropdown menu label?

napari/_qt/qt_plugin_list.py
```diff
@@ -0,0 +1,199 @@
...
+        order : list
+            A list of str, hookimpls, or module_or_class, with the desired
+            CALL ORDER of the hook implementations.
+        """
+        if not self.hook:
+            return
+        permute_hook_implementations(self.hook, order)
+
+
+class QtPluginSorter(QDialog):
```
sofroniewn: This is great - as we add more hook specs / see how more plugins work, we might want to revisit, provide more options etc, but for now works well

napari/components/add_layers_mixin.py
```diff
@@ -45,6 +50,57 @@ def add_layer(self, layer):
...
+    ) -> List[layers.Layer]:
+        """Add a path to the viewer
+
+        Paths will be handed one-by-one to the napari_get_reader hook.
+
+        Parameters
+        ----------
+        path : str or list of str
+            path(s) to view.
+        stack : bool, optional
```
sofroniewn: Maybe worth noting here that is `stack` is true then no plugins will be used?

napari/components/add_layers_mixin.py
```diff
@@ -45,6 +50,57 @@ def add_layer(self, layer):
...
+            Concatenate multiple input files into a single stack,
+            by default False
+        use_dask : bool, optional
+            Whether to use dask to create a lazy array, rather than NumPy.
+            If ``None``, will resolve to True if filenames contains more than
+            one image, False otherwise.  by default None.
+        """
+        paths = [path] if isinstance(path, str) else path
```
sofroniewn: someone has before asked for pathobjects to be used in #708 so we added https://github.com/napari/napari/blob/master/napari/utils/io.py#L38-L40 to magic_imread, should we have those lines here too / update docs accordingly

napari/components/add_layers_mixin.py
```diff
@@ -737,7 +795,21 @@ def _add_layer_from_data(
...
+        if not layer_type:
+            if hasattr(data, 'dtype') and data.dtype in (
+                np.int32,
+                np.uint32,
+                np.int64,
+                np.uint64,
+            ):
+                layer_type = 'labels'
+            else:
+                layer_type = 'image'
```
sofroniewn, line 843: I know this is preserving previous behaviour, that I think @jni added, but I don't really like this special treatment of labels - i think fine to assume if `layer_type` not provided it is an image, but i don't like guessing that it is labels. We can add a `layer_type` flag to our CLI if that is needed.

sofroniewn: that might be tricky though as it means we need to add one to `add_path` and its unclear how to pass it in a nice way, only if it hasn't been provided by the reading

napari/components/add_layers_mixin.py
```diff
@@ -45,6 +50,57 @@ def add_layer(self, layer):
...
+                "'path' argument must be a string, list, or tuple"
+            )
+
+        added_layers: List[layers.Layer] = []
+        if stack:
+            images = io.magic_imread(paths, use_dask=use_dask, stack=stack)
+            added = self.add_image(images)
+            added = added if isinstance(added, list) else [added]
+            added_layers.extend(added)
+            return added
```
sofroniewn: I can't say I love this, I guess the main reason this can't be a plugin is that plugins can only accept paths instead of also accepting lists of paths. If we allowed lists of paths then this could be accomplished, but maybe that increases complexity of the plugins for something that isn't too bad here, so maybe this is the right choice

sofroniewn (approved):
Ok @tlambert03 I think this is basically good to go!!! Very excited to get this in. Most of my comments are really about how we're combining our previous IO with the new `napari_get_reader` approach. I guess the most important thing to me is that we are all 100% about the function signature for the hook spec. I think we're in agreement on `ReaderFunction` signature here https://github.com/napari/napari/blob/master/napari/plugins/hook_specifications.py#L29-L34 that was introduced in #908 but becomes available for the first time now.

One question is if we are sure we don't want to allow lists of paths as input (this may be the right choice, and i know there was some discussion before)?

Another question is would we consider ever providing more IO hook_specs, for example something like `napari_get_image_reader` which we would match the API of imageio.imread, maybe `array + optional dict` and assume that was is provided is a single `Image` or `Labels`? or `napari_get_surface_reader` that took `tuple + optional dict` where the tuple was the vertex, face, value tuple we use for data? This would make integration with https://github.com/nschloe/meshio really easy. We don't have to decide this now, but curious how you think about this?

napari/_qt/qt_main_window.py
jni: This is typical of actions that require more input. See e.g. in Safari "Open File..." vs "New Window"

napari/components/add_layers_mixin.py
jni: Well, on the one hand, the plugins allow `layer_type` to be declared now (this was not the case pre-plugins), so it's fair to make that the canonical mechanism. On the other, though, I don't recall literally *ever* seeing a normal (ie intensity, not label) image stored as one of those dtypes. Ever. Have you? So to me it seems like, if a plugin wants to do the weird thing and load such an image as intensities, it can return the `'image'` layer type, and otherwise, we do our inference...

napari/components/add_layers_mixin.py
jni: Yes this is a good idea. =)

napari/_qt/qt_main_window.py
```diff
@@ -132,6 +134,11 @@ def _add_file_menu(self):
         open_images.setStatusTip('Open image file(s)')
         open_images.triggered.connect(self.qt_viewer._open_images)
 
+        open_stack = QAction('Open image series as stack...', self._qt_window)
+        open_stack.setShortcut('Ctrl+Alt+O')
+        open_stack.setStatusTip('Open image files')
+        open_stack.triggered.connect(self.qt_viewer._open_images_as_stack)
```
jni, line 148: Here I think `open image file(s)` should maybe be renamed to `open image file(s) as layers...`, because to me, stack is the assumption...

napari/_qt/qt_plugin_list.py
```diff
@@ -0,0 +1,199 @@
...
+        Usually instantiated by QtPluginSorter.
+
+        Parameters
+        ----------
+        hook : pluggy.manager._HookCaller, optional
+            A pluggy HookCaller to show implementations for. by default None
+            (i.e. no hooks shown)
+        parent : QWidget, optional
+            Optional parent widget, by default None
+        """
```
jni: @tlambert03 refer to my comment on other PRs... =) ie this docstring should, tragically, be on the class docstring.

napari/_qt/qt_plugin_list.py
```diff
@@ -0,0 +1,199 @@
...
+        hook : _HookCaller, optional
+            A pluggy HookCaller to show implementations for. by default None
+            (i.e. no hooks shown)
+        """
+        self.clear()
+        self.hook = hook
+        if not hook:
+            return
+        for hookimpl in reversed(hook.get_hookimpls()):
+            self.addItem(QtHookImplItem(hookimpl))
```
jni: Why are these reversed? Could you add a comment? I presume this has to do with the display order of the widget vs the priority order of the hookimpls...

napari/_qt/qt_plugin_list.py
```diff
@@ -0,0 +1,199 @@
...
+        initial_hook : str, optional
+            If provided the QComboBox at the top of the dialog will be set to
+            this hook, by default None
+        firstresult_only : bool, optional
+            If True, only hook specifications that declare the "firstresult"
+            option will be included.  (these are hooks for which only the first
+            non None result is returned).  by default True (because it makes
+            less sense to sort hooks where we just collect all results anyway)
+            https://pluggy.readthedocs.io/en/latest/#first-result-only
+        """
```
jni: As above...

napari/_qt/qt_plugin_list.py
```diff
@@ -0,0 +1,199 @@
...
+        self.hookComboBox.addItems(hooks)
+
+        self.hookComboBox.activated[str].connect(self.change_hook)
+        self.hookList = QtHookImplListWidget()
+
+        title = QLabel('Plugin Sorter')
+        title.setStyleSheet(
+            "QLabel { background-color : rgb(38, 41, 47); "
+            "color : #ddd; font-size: 18pt; }"
+        )
```
jni: Do these stylings belong here rather than in our stylesheets directly?

jni: Indeed this seems like it'll just plain look terrible when switching themes...?

napari/_qt/qt_plugin_list.py
```diff
@@ -0,0 +1,199 @@
...
+
+        Parameters
+        ----------
+        hook : str
+            Name of the new hook specification to show.
+        """
+        if hook == self.NULL_OPTION:
+            self.hookList.set_hook(None)
+        else:
+            self.hookList.set_hook(getattr(self.plugin_manager.hook, hook))
```
jni: This line is confusing to me... Does `self.plugin_manager.hook` contain attributes for every possible hook name?

jni: This might make more sense to me:

```python
getattr(self.plugin_manager.hooks, hookname)
```
? Am I parsing this right?

napari/_qt/qt_plugin_list.py
```diff
@@ -0,0 +1,199 @@
...
+class QtHookImplItem(QListWidgetItem):
+    """Item in a QtHookImplListWidget."""
+
+    def __init__(self, hookimpl, parent=None):
+        self.plugin_name = hookimpl.plugin_name
+        if self.plugin_name == 'builtins':
+            self.plugin_name = 'napari built-in'
+        super().__init__(self.plugin_name, parent)
+        self.hookimpl = hookimpl
+        self.setBackground(Qt.black)
```
jni: Here also, the theming within the class seems to go against our existing design principles.

napari/components/add_layers_mixin.py
jni: I'm totally with you @sofroniewn. This seems almost worth removing. @tlambert03 out of the options you presented, I'm definitely leaning towards making our own implementation a plugin. This will help us figure out how plugins are limited.

napari/components/add_layers_mixin.py
```diff
@@ -45,6 +50,57 @@ def add_layer(self, layer):
...
+
+        # iterate through each path provided, looking for a suitable reader
```
jni: Two possibilities:

(1) `get_reader` can also accept a list of paths, and return a plugin that can read them all, or
(2) plugin returned images can also be stacked.

I kinda like the first one better.

napari/components/add_layers_mixin.py
```diff
@@ -67,7 +123,7 @@ def add_image(
         blending=None,
         visible=True,
         path=None,
-    ):
+    ) -> Union[layers.Image, List[layers.Image]]:
```
jni, line 75: Ha, I was like, "surely not!", and had to keep expanding the signature until I was proven wrong. 😂

napari/plugins/_tests/fixtures/napari_bad_plugin2.py
```diff
@@ -0,0 +1,16 @@
+"""
+Test plugin that is technically correct, but would cause performance problems
```
jni: Incorrect docstring?

napari/plugins/_tests/fixtures/napari_bad_plugin2.py
```diff
@@ -0,0 +1,16 @@
...
+import pluggy
+
+napari_hook_implementation = pluggy.HookimplMarker("napari")
+
+
+def reader_function(path):
+    raise IOError("whoops")
+
+
+@napari_hook_implementation
```
jni, line 16: @tlambert03 can you remind me... Is it possible to have two hooks in napari with the same signature, but that do different things? And if so, how are they resolved? By the function name? Is there a way to avoid the function name problem? Apologies if you've answered this already... I think this was your PR to pluggy, right?

napari/plugins/utils.py
```diff
@@ -0,0 +1,160 @@
...
+                msg += (
+                    " If all items in `order` "
+                    "argument are not strings, they are assumed to be an "
+                    "imported plugin module or class."
+                )
+            raise ValueError(msg) from e
+
+    # make new arrays for _wrappers and _nonwrappers
+    _wrappers = []
+    _nonwraps = []
```
jni: What are wrappers and non-wrappers?

napari/plugins/utils.py
```diff
@@ -0,0 +1,160 @@
...
+    if all(isinstance(o, HookImpl) for o in order):
+        attr = None
+    elif all(isinstance(o, str) for o in order):
+        attr = 'plugin_name'
+    elif any(isinstance(o, str) for o in order):
+        raise TypeError(
+            "order list must be either ALL strings, or ALL modules/classes"
+        )
+    else:
+        attr = 'plugin'
```
jni: What is attr?

napari/plugins/utils.py
```diff
@@ -0,0 +1,160 @@
...
+    # were not specified in `order` argument
+    for i in sorted(indices, reverse=True):
+        del hookimpls[i]
+
+    if hookimpls:
+        _wrappers = [x for x in hookimpls if x.hookwrapper] + _wrappers
+        _nonwraps = [x for x in hookimpls if not x.hookwrapper] + _nonwraps
+
+    hook_caller._wrappers = _wrappers
+    hook_caller._nonwrappers = _nonwraps
```
jni: @tlambert03 honestly I find this function extremely hard to parse. From the docstring, I expected something like:

```python
hook_caller.hooks = [_get_hook_implementation(hook_caller, hook_id) for hook_id in order]
```

with a few checks. Can you explain why it is so much more complicated?

napari/plugins/utils.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+def get_layer_data_from_plugins(
+    path: str, plugin_manager=None
+) -> Optional[LayerData]:
+    """Iterate reader hooks and return first non-None LayerData or None.
+
+    This function returns as soon as the path has been read successfully,
+    whilecatching any plugin exceptions, storing them for later retrievial,
```
jni:
```suggestion
    while catching any plugin exceptions, storing them for later retrieval,
```

jni (commented):
@tlambert03 probably I'm a bit too tired for this review, but the truth is that I found the code quite hard to get through. I'm actually super excited about getting this in, but I would like us to spend a bit more time thinking about how to simplify this code so that it is clearer to newcomers... I've left comments at specific bits that I found confusing...

tlambert03:
> @tlambert03 probably I'm a bit too tired for this review, but the truth is that I found the code quite hard to get through. I'm actually super excited about getting this in, but I would like us to spend a bit more time thinking about how to simplify this code so that it is clearer to newcomers... I've left comments at specific bits that I found confusing...

Thanks so much for your already very detailed review.  your "tired" reviews are still pretty amazing.  I'm very sorry it was a struggle... I definitely agree that I let this PR balloon a bit too much.  Based on your comments, I think that a good portion of the complexity here has to do with "reordering" the call order of plugins, which is itself made much more complicated by [pluggy's concept of hook wrappers](https://pluggy.readthedocs.io/en/latest/#wrappers).  I will respond to your comments inline here, but I think what I'm going to do is pull out all of the code relating to the reordering of plugins into another PR, so we can discuss that independently.  That will leave us with plugins that (for now) always get called in the order that they get registered... we can add the reordering feature later.

napari/components/add_layers_mixin.py
tlambert03: yes I agree... will do.  curious though, why did you use `path.as_posix()` instead of `str(path)`?  is that going to be as robust on windows?

napari/components/add_layers_mixin.py
tlambert03: I agree. I'd much prefer to let plugins have access to everything, and this prevents plugins from reading stacks of individual files.  I like @jni's suggestion of having `napari_get_reader` accept either a `str` or `Sequence[str]` and return the appropriate reader function accordingly.  I'll work on that.

napari/components/add_layers_mixin.py
tlambert03: Yeah, I'm not opposed to letting an un-specified uint32/64 layer be coerced to labels unless they specifically provide `image` layer_type.  I've never seen those dtypes used for anything else.

napari/_qt/qt_main_window.py
tlambert03: i need clarification, there are two items in the menu (ignoring the statusTip for a moment): 
1. `Open image(s)...`
2. `Open image series as stack...`

I think there should be one very simple `Open image(s)` option, which is the first one, which just takes a path, makes no assumptions about it, and adds it to the viewer.  It allows choosing multiple files, but no requirement/assumption about their shapes is made, they each make a new layer.  I don't love renaming that `open image file(s) as layer(s)` because that assumes some napari knowledge about what a layer _is_ ... there should just be an "open file" option and this is it.  The second option imposes requirements on the images (i.e. that they all be the same dimensions), and then builds a 3D stack from 2D images...

Are you saying that, to you, the assumption is that if you provide two file paths, they are usually meant to be stacked on top of each other?  And you want the second function to be the default one?

napari/plugins/_tests/fixtures/napari_bad_plugin2.py
tlambert03: Same signature (`(path: str)`) yes, same function name (`napari_get_reader`), no.  Hook specifications and implementations are registered and matched up by function name (until they release that PR of mine that got accepted)

napari/components/add_layers_mixin.py
jni: iirc we checked that `as_posix` works on Windows. 🤷‍♂ I'm old fashioned and don't use Pathlib, so I can't say too much about it... =)

napari/_qt/qt_main_window.py
jni: > Are you saying that, to you, the assumption is that if you provide two file paths, they are usually meant to be stacked on top of each other? And you want the second function to be the default one?

Yes, that is what I was saying, *but*, you've convinced me otherwise. =) Well, I wasn't making judgment on the default, just saying that whichever default you picked, it should be documented in the menu name. Anyway, carry on... This problem will be much smaller when we add the "combine layers into stack" function. =)

napari/components/add_layers_mixin.py
sofroniewn: I can't say for sure - if I wrote that line then I must have stackoverflow that made the suggestion, but i'm really willing to defer to others who know more than i do here

napari/components/add_layers_mixin.py
sofroniewn: fine with me

sofroniewn:
@tlambert03 asked
> I think I could still use a little help understanding your desired function signatures and outcomes for Viewer.add_path, napari_get_reader, as well as QtViewer.dropEvent and the QtViewer._open* methods. if you can just describe the behavior you'd like to see, I can work on implementing it.

I think `add_path` should accept both a list of paths / individual path, and a stack bool. If stack is True then the entire list gets passed to `get_layer_data_from_plugins` in one go and then the plugins march through trying the whole list (if plugins only knows how accept a single path and more are provided then that plugin just fails and a new one is tried), if stack is False the list of paths is broken up as is done now and each path is loaded individually.

For `QtViewer.dropEvent` by default I'd have `stack=False`. If `shift` drop then I'd have `stack=True` and pass the paths + stack bool to `add_path`.

For the drop down menus I'd then have one _open_file with `stack=False` and one `_open_stack` with `stack=True` both of which again just call `add_path` with the appropriate args.

How does this sound?

tlambert03:
> How does this sound?

sounds good to me!  I'd like to hear a 👍from @jni before starting to implement it, so will wait for a little bit.

jni:
I like @sofroniewn's proposal. =)

tlambert03:
ok, the stack functionality is working nicely now, I added the shift/drop feature that @sofroniewn  mentioned, and support for Path objects is in there.  I _think_ we're pretty much there, but let me know if you can think of anything else.

I did have to make a compromise with the `use_dask`, command line option for stacks.  There doesn't seem be a good way to propagate it through `add_path`.  So do have a look at that and let me know if you want to preserve that option somehow (and have an idea).

For your convenience, here's the comparison of what has happened since the last round of reviews: https://github.com/napari/napari/compare/5248784..21d5811

napari/components/add_layers_mixin.py
```diff
@@ -1,10 +1,16 @@
 import itertools
+from logging import getLogger
+from typing import List, Optional, Sequence, Union
+from os import fspath
```
jni, line 4: Fascinating! Great find! =)

napari/components/add_layers_mixin.py
```diff
@@ -44,6 +50,65 @@ def add_layer(self, layer):
...
+        paths = [path] if isinstance(path, str) else path
+        paths = [fspath(path) for path in paths]  # PathObjects -> str
+        if not isinstance(paths, (tuple, list)):
+            raise ValueError(
+                "'path' argument must be a string, list, or tuple"
+            )
+
+        added: List[layers.Layer] = []  # for layers that get added
+
+        def load(_path):
```
jni: I find this nested function a bit confusing/hard to follow. Some suggested changes:

- rename `_path` to `_path_or_paths`
- add comment to the section `if not layer_data`: when this happens (plugin can't read), and what the response is (nothing much, but it looks like a lot of code)

As a bonus, I don't see it clearly, but it feels to me like the nested function definition isn't necessary. It might be feasible to make the code clearer e.g. by having an *external* rather than nested function.

napari/plugins/_builtins.py
```diff
@@ -2,12 +2,21 @@
...
 """
 from pluggy import HookimplMarker
+from typing import Union, List
+from ..types import ReaderFunction, LayerData
+from ..utils import io
 
 napari_hook_implementation = HookimplMarker("napari")
 
 
+def _interal_reader_plugin(path: Union[str, List[str]]) -> List[LayerData]:
```
jni: internal?

napari/plugins/_builtins.py
```diff
@@ -2,12 +2,21 @@
...
+    """Pass ``path`` to our magic_imread function and return as LayerData."""
+    # magic_imread is already capable of accepting both a str or a list
+    return [(io.magic_imread(path),)]
```
jni: It seems so close to being able to return `io.imread` directly, rather than this thing... Could we make a decorator that converts `path -> ArrayLike` into `path -> [(ArrayLike,)]` and use that? It seems that more than one package would find this useful...

jni: ... And I also wonder whether we could somehow also support the `io.magic_read` signature directly anyway... Maybe call it SimpleReaderFunction?

napari/plugins/_hookexec.py
```diff
@@ -0,0 +1,131 @@
...
+# Vendored with slight modifications from pluggy:
+# https://github.com/pytest-dev/pluggy/blob/master/src/pluggy/callers.py#L157
+def _multicall(hook_impls, caller_kwargs, firstresult=False):
+    """Execute a call into multiple python functions/methods and return the
+    result(s).
+
+    ``caller_kwargs`` comes from _HookCaller.__call__().
+    If ``caller_kwargs`` contains a key ``with_impl`` that evaluates to true,
+    results will be returned as 2-tuples of (result, hook_impl) instead of the
+    bare result.
```
jni: I know this is vendored but it would definitely benefit from a more complete, SciPy-style docstring, including with examples. Perhaps some example uses of `_multicall` could go in the above module docstring?

napari/plugins/_hookexec.py
jni: @tlambert03 since the module docstring contains some "local" comments, maybe the above comment could also be incorporated. More more more explicit docs! =) Thanks!

napari/plugins/_hookexec.py
```diff
@@ -0,0 +1,131 @@
...
+                    outcome.get_result(),
+                    impl_hits[0] if impl_hits else None,
+                )
+            return list(zip(outcome.get_result(), impl_hits))
+        return outcome.get_result()
+
+
+def _hookexec(
+    hook, skip_imps: Optional[Sequence[HookImpl]] = None, **kwargs
+) -> Any:
```
jni: @tlambert03 I feel like this needs a docstring as well... Maybe it can be incorporated into your PR? =)

napari/plugins/_tests/fixtures/napari_bad_plugin2.py
jni: Once that PR is released, what then? How are implementations mapped to specs?

napari/plugins/_tests/test_reader_plugins.py
```diff
@@ -0,0 +1,88 @@
...
+
+    # the good plugins (like "napari_test_plugin") should return layer_data
+    assert layer_data
+
+    # but the exception from `bad_plugin2` should now be stored.
+    assert 'napari_bad_plugin2' in plugin_manager._exceptions
+    # we can print out a string that should have the explanation of the error.
+    exception_string = plugin_manager.format_exceptions('napari_bad_plugin2')
+    assert 'IOError' in exception_string
+    assert "napari_get_reader" in exception_string
```
jni, line 38: I love love love this test and the explanatory comments! :heart:

napari/plugins/utils.py
```diff
@@ -0,0 +1,67 @@
...
+    Returns
+    -------
+    LayerData or None
+        LayerData that can be *passed to _add_layer_from_data.  If no reader
+        plugins are (or they all error), returns None
+    """
+    plugin_manager = plugin_manager or napari_plugin_manager
+    skip_imps = []
+    while True:
+        (reader, imp) = _hookexec(
```
jni: `(reader, implementation) = `?

napari/plugins/utils.py
```diff
@@ -0,0 +1,67 @@
+from . import PluginError, log_plugin_error
+from . import plugin_manager as napari_plugin_manager
+from ._hookexec import _hookexec
+from typing import Optional
+from ..types import LayerData
+
+
+def get_layer_data_from_plugins(
```
jni: This kind of feels less generic than `utils.py`. Should this perhaps live in `plugins/io_plugins.py`? I might also rename it `read_data_with_plugins`.

napari/plugins/utils.py
jni: I'd kind of like a comment above this about what hookexec does. It seems very magical.

napari/plugins/utils.py
jni: I love the function though, quite easy to follow. =)

jni (commented):
@tlambert03 

This is almost ready, which is exciting! In addition to the inline comments below, which are mostly about adding documentation, could you add a tiny Reader plugin (e.g. for imageio.imread) in `examples/reader_plugin.py`?

jni:
Re: dask, I feel like this is quite a necessary feature, but I think it can be brought back via plugin ordering? `magic_read` already does some dasking, right? =)

tlambert03:
> Re: dask, I feel like this is quite a necessary feature, but I think it can be brought back via plugin ordering? `magic_read` already does some dasking, right? =)

yeah, the magic_read function still has the `use_dask` argument, and will automatically use dask in certain cases, that's not changing... what is gone is the propagation of that option all the way back to the CLI.  One of the challenges here: since we're treating `magic_imread` like an internal plugin, we'd need to add some sort of `use_dask` option to the hook specification ... but I don't think we should do that.  So `magic_imread` just needs to decide on its own now.

napari/plugins/_builtins.py
tlambert03: dumb name :) will change

napari/plugins/_hookexec.py
tlambert03: Sure.  With this last modification (wrapping `res = hook_impl.function(*args)` in a try/catch), I'm actually starting to think that we may well need to keep a fair amount of permanent vendored/modified pluggy code.  I'm not sure we'll be able to expect pluggy to integrate all the changes we want... so I'll just remove the "temporary" comments and flesh out our internal comments

napari/components/add_layers_mixin.py
tlambert03: Will do. I considered making it a separate function, and I'm fine with that.

napari/plugins/_builtins.py
tlambert03: sure, love that.

napari/plugins/_hookexec.py
tlambert03: will add.  I feel like that PR might sit there for quite a while, so let's just own it.

napari/plugins/_tests/fixtures/napari_bad_plugin2.py
tlambert03: now: hook implementation names _must_ match the hook specification name:
```python
@hook_implementation
def name_of_hookspec(): ...
```
after that PR: the decorator accepts a `specname` option:
```python
@hook_implementation(specname="name_of_hookspec")
def any_name_you_want(): ...
```

napari/plugins/utils.py
tlambert03: yeah, I have also felt like it didn't quite belong there.  I'd be happy to create a new module for it and rename it.

napari/plugins/utils.py
tlambert03: I will add a comment to the code, but here's the explanation... buckle up ;)

When using a pluggy `PluginManager`, when you actually call a hook using their syntax `plugin_manager.hook.hook_spec_name(**kwargs)` it goes through a relatively complicated chain of methods and aliases before ultimately landing on the basic function that loops through plugins:

[`hooks._HookCaller.__call__`](https://github.com/pytest-dev/pluggy/blob/6e8b6d4e8e0519e50af05b4e4b90b18b924bcbcc/src/pluggy/hooks.py#L278) → 
[`hooks._HookCaller._hookexec`](https://github.com/pytest-dev/pluggy/blob/6e8b6d4e8e0519e50af05b4e4b90b18b924bcbcc/src/pluggy/hooks.py#L292) → created during hook registration in
[`PluginManager.register`](https://github.com/pytest-dev/pluggy/blob/6e8b6d4e8e0519e50af05b4e4b90b18b924bcbcc/src/pluggy/manager.py#L124) →
[`PluginManager._hookexec`](https://github.com/pytest-dev/pluggy/blob/6e8b6d4e8e0519e50af05b4e4b90b18b924bcbcc/src/pluggy/manager.py#L90)  →
[`PluginManager._inner_hookexec`](https://github.com/pytest-dev/pluggy/blob/6e8b6d4e8e0519e50af05b4e4b90b18b924bcbcc/src/pluggy/manager.py#L84) → which really just calls
[`_HookCaller.multicall`](https://github.com/pytest-dev/pluggy/blob/6e8b6d4e8e0519e50af05b4e4b90b18b924bcbcc/src/pluggy/hooks.py#L215) → which is an alias for
[`pluggy.callers._multicall`](https://github.com/pytest-dev/pluggy/blob/6e8b6d4e8e0519e50af05b4e4b90b18b924bcbcc/src/pluggy/callers.py#L157) (_this is the the thing that actually loops through plugin hook_implementations_)

**point being**: if you want to make any modifications to the way plugins are looped through in a `PluginManager`, one can't really just override a method on our `PluginManager` class.  My `_hookexec` function here is my strategy for modifying the core plugin loop.  I will update the comments/name of that function to make that clearer, since I suspect it may be a permanent resident in napari (given the importance of that plugin loop).

tlambert03:
ok, new diff from the last review is [here](https://github.com/napari/napari/compare/21d5811..9ba1109).
- added a significant amount of documentation to `_hookexec.py`, in preparation for it being a perhaps-not-so-temporary part of the codebase.  hopefully, combined with my comment above, it's a little clearer now exactly what it's doing, and why it's necessary.
- broke out the nested `load()` function from `Viewer.add_path` into its own `_add_layers_with_plugins` method.
- created `napari.types.array_return_to_layerdata_return`: it takes a `PathLike -> ArrayLike` function and returns a `PathLike -> LayerData` function, and can be used as a decorator.  I use it to wrap `magic_imread` in our `_builtins.napari_get_reader`.  

@jni: I like the concept of providing direct support for `SimpleReaderFunction`s (`PathLike -> ArrayLike` functions like `magic_imread`)... but it's tricky.  Seems to me we either have to provide two different hookspecs (which has the downside of being a bit more confusing to developers, and then we need to figure out how to combine them wherever we actually call that hook), or we could offer a single hookspec with return type inspection: if the plugin annotates their hook implementation as returning an `ArrayLike` instead of `List[LayerData]`, _we_ would do the wrapping for them (but I don't love having a "fuzzy" type on our hook specification... feels like a "sloppy contract", plus they would need to import the `ArrayLike` type from `napari`, right?).  They could also just import `array_return_to_layerdata_return` and wrap their function.

napari/components/add_layers_mixin.py
```diff
@@ -736,7 +828,21 @@ def _add_layer_from_data(
 
         """
 
-        layer_type = layer_type.lower()
+        layer_type = (layer_type or '').lower()
+
+        # this came from __main__.py
```
sofroniewn: Probably don't need this comment `# this came from __main__.py` anymore

napari/plugins/_builtins.py
sofroniewn: I'm ok with that guessing / default behavior, but we might want to keep a close handle on this as we think about easy ways to add meshes, shapes etc too

sofroniewn: actually, i take that back - see more detailed comment bellow

sofroniewn (commented):
@tlambert03 

> broke out the nested load() function from Viewer.add_path into its own _add_layers_with_plugins method.

I love how this part of the code is looking now, really nice.

> @jni: I like the concept of providing direct support for SimpleReaderFunctions (PathLike -> ArrayLike functions like magic_imread)... but it's tricky. Seems to me we either have to provide two different hookspecs (which has the downside of being a bit more confusing to developers, and then we need to figure out how to combine them wherever we actually call that hook), or we could offer a single hookspec with return type inspection: if the plugin annotates their hook implementation as returning an ArrayLike instead of List[LayerData], we would do the wrapping for them (but I don't love having a "fuzzy" type on our hook specification... feels like a "sloppy contract", plus they would need to import the ArrayLike type from napari, right?). They could also just import array_return_to_layerdata_return and wrap their function.

This all just scares me - my comment here https://github.com/napari/napari/pull/937#pullrequestreview-369233214 was trying to get at that

> Another question is would we consider ever providing more IO hook_specs, for example something like napari_get_image_reader which we would match the API of imageio.imread, maybe array + optional dict and assume that was is provided is a single Image or Labels? or napari_get_surface_reader that took tuple + optional dict where the tuple was the vertex, face, value tuple we use for data? This would make integration with https://github.com/nschloe/meshio really easy. We don't have to decide this now, but curious how you think about this?

and I think we should wait on this for a follow-up PR. We're about to get in a really nice general purpose system. I think in the future we could add some convenience hookspecs that were dedicated for the purposes mentioned above, but adding guessing / magic on top of our general system seems like a bad idea to me

For the `dask` stuff:

> So magic_imread just needs to decide on its own now.

This seems like a good approach for me. We can hard code that to be True, but I think plugins themselves should decide about things like dask, not us.

All in all @tlambert03 this looks amazing - I think it is good to go from my perspective (I made one truly minor comment about removing a not needed comment). Are there more things that you'd like my take on?

We'll let @jni respond to the above too - but I think this is getting pretty close!!

tlambert03:
> This all just scares me ... and I think we should wait on this for a follow-up PR.

yeah that's my sentiment too.  Worst case scenario with this current PR is that it demands _slightly_ more from a plugin developer.  I'm more than happy to leave considerations of a simpler, restricted API for later.

> We can hard code that to be True, but I think plugins themselves should decide about things like dask, not us.
###  👍

> Are there more things that you'd like my take on?

nope, thanks!

napari/plugins/__init__.py
```diff
@@ -2,7 +2,7 @@
 
 from pluggy import HookimplMarker
 
-from ._hookexec import _hookexec
+from ._hook_callers import execute_hook
```
jni, line 5: :heart:

napari/plugins/_hook_callers.py
```diff
@@ -0,0 +1,272 @@
...
+...                     plugin_manager.hook.my_hook,
+...                     arg=arg,
+...                     return_impl=True,
+...                 )
+
+NOTE: A version code here was submitted to pluggy as PR #253:
+https://github.com/pytest-dev/pluggy/pull/253
+We may want to consider removing this code if that merges, but we may also just
+wish to maintain our internal hook looping logic.
+"""
```
jni, line 41: :heart: this docstring

napari/plugins/_hook_callers.py
```diff
@@ -0,0 +1,272 @@
...
+
+
+@overload
+def _multicall(
+    hook_impls: Sequence[HookImpl],
+    caller_kwargs: dict,
+    return_impl: Literal[True],
+    firstresult: Literal[False],
+) -> List[Tuple[Any, HookImpl]]:
+    ...
```
jni: I'm super-confused about these signatures, what's happening here?

napari/plugins/_hook_callers.py
```diff
@@ -0,0 +1,272 @@
...
+    caller_kwargs: dict,
+    return_impl: bool = False,
+    firstresult: bool = False,
+) -> Union[Any, List[Any], Tuple[Any, HookImpl], List[Tuple[Any, HookImpl]]]:
+    """Loop through ``hook_impls`` with ``**caller_kwargs`` and return results.
+
+    Parameters
+    ----------
+    hook_impls : list
+        A sequence of HookImpl objects
```
jni:
```suggestion
        A sequence of hook implementation (HookImpl) objects
```

napari/plugins/_hook_callers.py
```diff
@@ -0,0 +1,272 @@
...
+    **kwargs,
+) -> Union[Any, List[Any], Tuple[Any, HookImpl], List[Tuple[Any, HookImpl]]]:
+    """Return result(s) from all or some of the implementations for ``hook``.
+
+    Examples
+    --------
+    Calling hooks the normal way...
+    >>> result = plugin_manager.hook.my_hook(arg=arg)
+
+    ...is identical to calling them with this function as follows:
```
jni: identical is the wrong word here. =)

napari/plugins/_hook_callers.py
tlambert03: Totally optional, those are typing.overload hints.  Since the return type of _multicall changes depending on the arguments, I though it might be useful to be explicit about what output should be expected depending on the input.  Happy to remove!  Wouldn’t change any functionality

napari/plugins/_hook_callers.py
tlambert03: Hah,  “equivalent” yes?

napari/types.py
```diff
@@ -1,6 +1,50 @@
-from typing import Callable, List, Tuple, Union, Any, Dict
+from functools import wraps
+from typing import Any, Callable, Dict, List, Protocol, Tuple, Union
+
+import numpy as np
+
+
+class ArrayLike(Protocol):
+    """This is a WOEFULLY inadqueate stub for a duck-array type.
```
jni:
```suggestion
    """This is a WOEFULLY inadequate stub for a duck-array type.
```

napari/plugins/_builtins.py
```diff
@@ -32,4 +28,4 @@ def napari_get_reader(path: Union[str, List[str]]) -> ReaderFunction:
     callable
         function that returns layer_data to be handed to viewer._add_layer_data
     """
-    return _interal_reader_plugin
+    return array_return_to_layerdata_return(magic_imread)
```
jni: Just making sure I'm understanding: technically, shouldn't this return `None` if the file is not something that `magic_imread` can read? e.g. a `.csv` file? But you are leaving it like this because we might as well try it and have it crash rather than create a super complex checker?

napari/types.py
```diff
@@ -1,6 +1,50 @@
...
-# using "Any" for the data type for now
+# using "Any" for the data type until ArrayLike is more mature.
 LayerData = Union[Tuple[Any], Tuple[Any, Dict], Tuple[Any, Dict, str]]
-ReaderFunction = Callable[[str], List[LayerData]]
+
+PathLike = Union[str, List[str]]
+ReaderFunction = Callable[[PathLike], List[LayerData]]
+
+
+def array_return_to_layerdata_return(
```
jni: I love everything about this except the name. =) I might call it `image_reader_to_layerdata_reader`?

napari/types.py
```diff
@@ -1,20 +1,13 @@
...
-
-    shape: Tuple[int, ...]
-    ndim: int
-    dtype: np.dtype
-
+# This is a WOEFULLY inadqueate stub for a duck-array type.
+# Mostly, just a placeholder for the concept of needing an ArrayLike type.
+# It doesn't actually get used for type checking anywhere.
+# Ultimately, this should come from https://github.com/napari/image-types
+ArrayLike = NewType("ArrayLike", np.array)
```
jni: Where does it get used if not in type checking? This is even more inadequate than before. =P

napari/plugins/_hook_callers.py
jni: Oooooh, I see now that the input *values* are changing. That's super crazy, I didn't know typing could do that. But I guess `Literal[True]` would not actually work if a variable that evaluates as True is passed? Personally I think it is too confusing to make it worthwhile. But I don't feel super strongly about it, but if you want to keep it I would add a comment to the top of the section explaining that the type of the return is changing in response to the value of the inputs, and `@overload` is taking care of that.

napari/plugins/_hook_callers.py
jni: For example. =)

jni:
@tlambert03 This is very wonderful. =) I've left only comments about style and documentation. The change from `_hookexec()` to `execute_hook` is :ok_hand:!

Last thing, could you add that plugin example to the examples that I mentioned earlier? I think it would be useful as a preliminary user guide, especially since our current internal plugin doesn't really show all the features (return None if the path can't be read).

jni:
Regarding the SimpleImageReader, my proposal was to have it be a different hookspec, not an overloaded one, but I'm happy to scrap it for now with your new decorator, as long as we give it a friendlier name. =)

napari/plugins/_hook_callers.py
tlambert03: yeah, `@overload` is very cool.   I agree though, it's too confusing here and I'll take it out.
but fyi, the main reason I'm generally tempted to do that sort of thing is for linting in my IDE since I've started using `mypy`... it would, for instance, show me if I'm trying to unpack a tuple that doesn't exist (cause i forgot to use `return_impl=True` or something)

jni: That's very clever. =) At skimage though, we're starting to see changing the number of return values in response to a kwarg value as an anti-pattern, precisely for this reason. Our proposal for this in the future is:

- `function.extra()` is a function with extra return values.
- You can only return the "normal" return, or everything, but,
- "everything" will come bundled in a NamedTuple or Dataclass, and different attributes can be None if desired.

So you can have `func(Arr) -> Arr` and `func.extra(Arr) -> Arr, Extras`.

tlambert03: definitely agree with overloaded return types being an antipattern!  (even more so now since I'm type-"woke" and want to be able to trace types carefully 😉 ).  I _love_ that approach you have there.

napari/plugins/_builtins.py
tlambert03: Good question :) 
Yes: because our own `napari_get_reader` plugin is last in line, we let it be a little special and try to read anything.  This is partially because `skimage.io` is somewhat flexible (e.g. even URLs work) and I didn't want to accidentally exclude something it _could_ read.
Note too: because we _also_ use a try/catch in [this line](https://github.com/napari/napari/blob/9ba11095e8a90f6056194249d539df9eb10c612c/napari/plugins/io.py#L47-L49) of `read_data_with_plugins`, if `magic_imread` throws an exception when reading the file, it will fail gracefully: catching and storing the exception, and still showing the "no plugin reader found" error.  (Users could also report that to us in #1024)

napari/types.py
tlambert03: i was at a loss :). what should I use then for `ArrayLike` in the type annotation in 
```python
def image_reader_to_layerdata_reader(
    func: Callable[[PathLike], ArrayLike]
) -> ReaderFunction:
```

jni: Well, it should at least be a `Union[np.array, dask.Array, zarr.Array]` (or whatever those types actually are, I haven't checked. =)

jni: Also, there should be a note that this should be a Protocol once we move to 3.7, is that right?

tlambert03: > Well, it should at least be a Union[np.array, dask.Array, zarr.Array]

I skimped out on that because I would have to `try: import zarr` since we don't depend on it.  But I can do that

> there should be a note that this should be a Protocol once we move to 3.7, is that right?

`Protocol` is actually 3.8, but can be used in ≥3.5 if we depend on the [typing-extensions](https://github.com/python/typing/tree/master/typing_extensions) backport.

tlambert03:
OK, I made all the last requested changes, and I went back through all the conversations (resolved and unresolved) and I think i've hit everything!

> Last thing, could you add that plugin example to the examples that I mentioned earlier? I think it would be useful as a preliminary user guide, especially since our current internal plugin doesn't really show all the features (return None if the path can't be read).

added an example in `examples/`... also just want to point out that there's a pretty extensive user guide that is more or less ready to go in https://github.com/napari/napari/pull/1030

jni (approved):
:tada:

jni:
> also just want to point out that there's a pretty extensive user guide that is more or less ready to go in #1030

The downside of splitting PRs into digestible chunks! "Hey I ordered some fries with this!" "The fries come in a separate course, sir, due to past complaints of indigestion."

tlambert03:
> > also just want to point out that there's a pretty extensive user guide that is more or less ready to go in #1030
> 
> The downside of splitting PRs into digestible chunks! "Hey I ordered some fries with this!" "The fries come in a separate course, sir, due to past complaints of indigestion."

LOL

sofroniewn:
Ok, beautiful work @tlambert03 and thanks for the through reviews @jni - @tlambert03 you should do the honors with the merge whenever you are ready!! Excited to have this in :-)

jni:
Good morning! :joy: So I was closing tabs and I'd had this open in so many, eventually that big green button flashing over and over on my screen was too much and I couldn't resist clicking it. :joy: Bring on the IO plugins! :tada:

tlambert03:
Hah!  Yay I’m glad you did it!  Exciting!

Issue #382, opened 2019-07-02
Capability to exporting markers,regions coordinates into csv file or xlsx files


pranathivemuri:
## 🚀 Feature
<Capability to exporting markers,regions coordinates into csv file or xlsx files-->

## Motivation

Annotators/Pathologists/ in general people who don't have access to/know how to code in python can just go to in the menu on GUI File -> export markers -> select file type/filename (csv or xlsx) and save them.

GenevieveBuckley:
I think something like this might work well for points, and a similar method for polygons could be made pretty easily as well. I'll open a PR in a bit, unless anyone has comments now.

```python
def to_csv(self, filename):
    with open(filename, mode='w') as file:
        writer = csv.writer(file, delimiter=',', quotechar='"', quoting=csv.QUOTE_MINIMAL)
        n_dimensions = self.data.shape[1]
        header = ['id'] + ['dimension_' + str(n) for n in range(n_dimensions)]
        writer.writerow(header)
        for idx, row in enumerate(self.data):
            point_data = np.concatenate([np.array([idx]), row])
            writer.writerow(point_data)
```

sofroniewn:
This has now been implemented with our builtin writer plugin

PR #941, opened 2020-02-03
Save points/shapes coordinates to csv file

GenevieveBuckley:
# Description

This pull request adds functionality for saving points and shapes to csv, and adds a button to the GUI for this function (as requested in issue https://github.com/napari/napari/issues/382).

Feedback wanted:
1. How do I add a button to the GUI for exporting points/shapes to csv? What should it look like?
2. How should we handle the nested data structures for shapes layers (several shapes, each with their own coordinates)? What should the output csv file look like?
3. Is there a better row header for the points csv than "id, dimension_0, dimension_1, ..."? Is it possible/should we try to tie the csv headings to the actual dimension names (xyztc)?

Before merging, this PR will also:
- [x] Example for the napari docs. Includes explanation of how to convert to/from pandas dataframes/tables/csv.
- [x] Allow csv export of points/shapes
    - [x] points to_csv() method
    - [x] shapes to_csv() method
    - [x] test for points to_csv()
    - [x] test for shapes to_csv()
- [x] Allow csv export of only the *selected* points/shapes
    - [x] *selected* points only
    - [x] *selected* shapes only
    - [x] test for selected points only
    - [x] test for selected shapes only
- [ ] ~~Allow import of points/shapes from csv~~
*I think it's probably better to do this in a separate PR, there needs to be some discussion.*
    - [ ] ~~points from_csv() method~~
    - [ ] ~~test for points from_csv()~~
    - [ ] ~~shapes from_csv() method~~
    - [ ] ~~test for shapes from_csv()~~
    - [ ] ~~test for points round trip to_csv() -> from_csv()~~
    - [ ] ~~test for shapes round trip to_csv() -> from csv()~~
- [ ] ~~Add GUI buttons for exporting (selected or all) points/shapes to csv.~~ 
*(Edit: we'll leave this for a later PR, as suggested by Nick below).*

## Type of change

# References

Reference: https://github.com/napari/napari/issues/382

# How has this been tested?

- [x] new test added: napari/layers/points/_tests/test_points.py::test_points_to_csv
- [ ] new test added: napari/layers/shapes/_tests/test_shapes.py::test_shapes_to_csv
- [ ] the new tests cover the case where we save only the *selected* points/shapes.

## Final checklist:


sofroniewn:
@GenevieveBuckley great to see you picking this up - this is 100% functionality that we will want to support and now is a good time to be thinking about it. There's been a tiny bit of discussion of file IO in #379 which I want to link here just for completeness. I'll start with answering your questions as best I can and then maybe make some additional comments

### Answers to questions

> How do I add a button to the GUI for exporting points/shapes to csv? What should it look like?

I'd say we should probably try and get the first version of the PR in without a GUI button / possibly with an entry from the file dropdown menu, but even that can wait. We've often taken the strategy of trying to get the "model" (i.e. no viz / GUI) code in first and then added GUI code in follow-up PRs as it can make development and review a little easier. I'm also not sure if just having something in the file dropdown won't be sufficient.

> How should we handle the nested data structures for shapes layers (several shapes, each with their own coordinates)? What should the output csv file look like?

You could use the `ID, dim_0, dim_1 ... ` approach where each shape has been concatenated along the list axis, and the ID is duplicated as many times as needed. So if you have a `5x3`, a `7x3`, and a `6x3` polygon you'd get an `18x3` array, and an index of `[0, 0, 0, 0, 0, 1, 1, ...., 2, 2]`. Each shape always has to have the same number of coordinates so this should always work. You might want to also grab the `face_color`, `edge_color`, `edge_width`, and make them columns too. Generating this representation should be it's own method on Shapes, and we might want to make the Shapes layer able to directly load this sort of data too. It might be easier if we had pandas as a dependency now as what I'm describing really is a tidy data frame, but I at least had been resistant to adding that dependency. Curious what @kevinyamauchi thinks, and how this would fit in with any of the Points refactoring done in #858. Maybe now is the time to depend on pandas too.

At a high level I'd like there to be as much of a round trip possible with our fileIO - i.e. you have data in the GUI, you save it out, you load it in, and you should hopefully see the exact same thing :-)

> Is there a better row header for the points csv than "id, dimension_0, dimension_1, ..."? Is it possible/should we try to tie the csv headings to the actual dimension names (xyztc)?

I like the idea of taking the  actual dimension names which I think can be found in `viewer.dims.axis_labels`, but these don't live in the actual layer (i.e. `layer.dims.axis_labels` will just have default values), so we might need to give this more thought.

### Additional comments
 
I'm not 100% sure about where the `to_csv` functions should live, in general we've avoided explicit fileIO within the layer models and I think I'd like to keep it that way as it becomes easier to separate out concerns that way (for example the svg generation functionality just generates the svg - it doesn't actually write the file, and the non of the layers have a file open method themselves, though we did add fileIO to the `add_*` methods that live on the viewer). Similarly here we might want to just prep the data for file writing but not write the file within the layer. An advantage of this separation is that if you have 3 layers say, you might be able to prep each one independently and then combine them in a larger command at the viewer level. (This would also make it easier to include the viewer axis labels too). In general I think we should support both - saving a selected layer as an atomic unit and saving data grouped together across selected layers or the entire viewer.

We're also actively working on the IO plugin interface right now - see #937 for how we're thinking about file readers. I think we'll still want to support basic file saving like this for all our layer types without plugins, but it might be worth thinking a little about how any of this stuff would interact with say file "writers" and curious if @tlambert03 has thoughts on this. We might also want to think a little about what save for the image / labels layers looks like to ensure consistency of approach too.

Do these thoughts help @GenevieveBuckley? Curious what your take on them is, if they seem reasonable / different from what you had in mind.

GenevieveBuckley:
* I am anti-pandas as a dependency (and we don't need it for this anyway)
* Sure, we can leave actually saving the csv to a later PR and just return the table from `to_csv()`. 
(FYI: your suggestion is the same behaviour as `to_xml()`, but `to_svg()` includes an optional "file" kwarg which will save to disk if provided. Mind you, writing an svg file to disk doesn't need any decisions about *how* to write, whereas saving images definitely does need a lot of decisions so I agree with your point.)
* Sure, we can also leave adding a button or menu item to a later PR as well.

> At a high level I'd like there to be as much of a round trip possible with our fileIO - i.e. you have data in the GUI, you save it out, you load it in, and you should hopefully see the exact same thing :-)

I imagine this means we probably need a `from_csv()` method for both the points and shapes layers as well.

GenevieveBuckley:
I've used this [markdown table generator](http://www.tablesgenerator.com/markdown_tables#) to illustrate what's going on.

I won't include things like face_color, etc. in this pull request, but we can discuss whether to add that in a later PR. Here's what we currently have for shapes:

### Shapes csv
| shape_id | shape_type | coord_id | dim_0 | dim_1 |  dim_2 | ... | dim_n |
|----------|------------|----------|-------|-------|--------|-----|-------|
| 0        | Ellipse    | 0        |       |       |        |     |       |
| 0        | Ellipse    | 1        |       |       |        |     |       |
| 0        | Ellipse    | 2        |       |       |        |     |       |
| 0        | Ellipse    | 3        |       |       |        |     |       |
| 1        | Rectangle  | 0        |       |       |        |     |       |
| 1        | Rectangle  | 1        |       |       |        |     |       |
| ...      | ...        | ...      |       |       |        |     |       |
| n        | Path       | n-1      |       |       |        |     |       |
| n        | Path       | n        |       |       |        |     |       |

And this is what I'd like to have for the points...

### Points csv
| shape_id | shape_type | coord_id | dim_0 | dim_1 | ... | dim_n |
|----------|------------|----------|-------|-------|-----|-------|
| 0        | Points    | 0        |       |       |     |       |
| 0        | Points    | 1        |       |       |     |       |
| 0        | Points    | ...      |       |       |     |       |

jni:
A few thoughts:

- reading and writing CSVs gets very hairy very fast, so I might indeed be inclined to depend on pandas at that point, unless we decide to make this a plugin, in which case, the plugin should depend on pandas. :joy: For example, now that we have attributes with points, presumably saving to csv will save the attribute columns. Some of those might include text, and if it does, the text might include commas, and then you're in trouble without a proper library.
- I don't think having the type "Point" repeated in every column of the CSV is nice or useful, with the possible exception of multi-layer CSV files, which I'm mildly against as they will be non-standard. And at any rate, shouldn't the shape ids be incrementing in that table?
- I don't think CSV is the right format for shapes, though if we do provide it as a format, the table you suggested @GenevieveBuckley is very nice. For shapes, I think a format like GeoJSON might be the right approach. At any rate, we should look at other libraries that deal with shapes to find standard solutions. Off the top of my head: shapely, fiona, rasterio. It might also be interesting to see how Fiji saves and loads ROIs. It'd be pretty neat to be able to move ROIs between Fiji and napari!
- @sofroniewn, things like face color and edge color are the reason why I advocate the approach in #858 throughout. To a first approximation, color is not the thing that the researcher is interested in. Rather, it *encodes* some other variable of interest, such as "object type" (nucleus, cell, Golgi..., or human, bicycle, tree, ...). It is this variable that needs saving and loading, not color. This is quite analogous to views and models.
- @sofroniewn is there a reason that the dims names are not propagated from the viewer to the layers? This seems important. It would be ideal to propagate the dims names through the CSV.
- I agree that we should worry about the drop down and possible buttons in later PRs.

sofroniewn:
> @sofroniewn, things like face color and edge color are the reason why I advocate the approach in #858 throughout. To a first approximation, color is not the thing that the researcher is interested in. Rather, it encodes some other variable of interest, such as "object type" (nucleus, cell, Golgi..., or human, bicycle, tree, ...). It is this variable that needs saving and loading, not color. This is quite analogous to views and models.

I guess as much as possible I'd like us to save whatever is needed for a full round trip to get the result we want, so if people use the `properties` approach then `face_color='object_type'` would be nice, but if they have specific colors `face_color='red'` it would be nice to save them too. I'm also ok with us starting by adding some basic support that meets the needs of @GenevieveBuckley without colors etc. and then maybe try making a writer plugin with more of this stuff, and then if everyone wants it move that writer plugin into the core (or not!!)

> @sofroniewn is there a reason that the dims names are not propagated from the viewer to the layers? This seems important. It would be ideal to propagate the dims names through the CSV.

I think we were going to revisit this when working on finally closing out #14 and we havn't gotten there yet. Would be good to think about both at the same time

GenevieveBuckley:
> in general we've avoided explicit fileIO within the layer models and I think I'd like to keep it that way as it becomes easier to separate out concerns that way (for example the svg generation functionality just generates the svg - it doesn't actually write the file

This isn't actually true, @sofroniewn. In the base layer class, there is a `file` kwarg in `to_svg()` https://github.com/napari/napari/blob/7a04984472db9a0b74c4efc26d56a6f56b4ce9dd/napari/layers/base/base.py#L657
```python
        if file:
            # Save svg to file
            with open(file, 'w') as f:
                f.write(svg)

        return svg
```

Does this need to be arranged differently? I was considering repeating this kind of pattern in other places (see discussion https://github.com/napari/napari/pull/944/files#r374978156), so this is a good opportunity to say if it should be done another way.

jni:
> I guess as much as possible I'd like us to save whatever is needed for a full round trip to get the result we want

I don't think this is necessarily a goal for every save interaction. As an example, in every case *I've* cared about, I just wanted point coordinates, and would have been annoyed to have extra columns of face colour and edge colour, all of which were incidental to my data.

In the viewer save state interaction, it makes a lot of sense to provide round trip capability. In other contexts, it may or may not make sense.

GenevieveBuckley:
> At a high level I'd like there to be as much of a round trip possible with our fileIO - i.e. you have data in the GUI, you save it out, you load it in, and you should hopefully see the exact same thing :-)

I think there's a big conflict between the request in the original issue (for non-programmers to have an easy way to export points/shapes/etc to csv that they can use for analysis), and the impulse of the programmers here who want round-trip capable functions.

I'm going to argue against round-trip capable csv files. A csv containing every piece of information needed for round-trip reloading with no information loss is going to be ugly, overwhelming, and difficult to use for analysis in Excel or similar. That is not a kind thing to do to our non-programmer friends and users.

I would be ok with:
* Other round-trip capable file formats existing alongside the csv option (I'm sure there are more appropriate formats out there for this purpose)
* The option to include every single piece of information in the csv, as long as it's not the default. Maybe.

@jni is right, sometimes people will encode information in things like face_color instead of different layers. In that case, I think it's enough flexibility to also allow exporting only *selected* points/shapes to csv, and the user can control this.

sofroniewn:
> I would be ok with:
>
> - Other round-trip capable file formats existing alongside the csv option (I'm sure there are more appropriate formats out there for this purpose)
> - The option to include every single piece of information in the csv, as long as it's not the default. Maybe.

I am also ok with these options, I'm also ok with them coming later / separately, especially as we are close to having support for writer plugins when this can be explored to our hearts content :-)

I'm fine with this PR focusing on saving out just the minimal location / coordinate data to a csv file, ideally without adding any adding dependencies. It should also be easy for to get that minimal information back into napari, but we can also think about that in a followup PR. Maybe there I'd take the approach of adding a `path` argument to `add_points` and `add_shapes` that just loads the location / coordinate data and does minimal prepping of it, in the same way that for `add_image` the `path` argument just loads the image data. Though there has been some discussion about this API in #937.

As to the comment from @jni
> I don't think CSV is the right format for shapes, though if we do provide it as a format, the table you suggested @GenevieveBuckley is very nice. For shapes, I think a format like GeoJSON might be the right approach. At any rate, we should look at other libraries that deal with shapes to find standard solutions. Off the top of my head: shapely, fiona, rasterio. It might also be interesting to see how Fiji saves and loads ROIs. It'd be pretty neat to be able to move ROIs between Fiji and napari!

I love the idea of having readers and writers to both GeoJSON and FIJI ROI format, but I think these will be IO plugins with their own dependencies, and not something we want to put in the core yet (or maybe ever).

I think having the csv representation that as @GenevieveBuckley says can be read into excel and not be super intimidating seems really useful for everyone.

@GenevieveBuckley do you have a clear sense of what to do next? Are there still points of confusion? Sorry I side tracked us a little on the `face_color` stuff etc., but it was helpful for me at least to discuss it.

GenevieveBuckley:
> @GenevieveBuckley do you have a clear sense of what to do next? Are there still points of confusion?

I'm not entirely sure I have a good grasp of where you'd like stuff to live, @sofroniewn 
How do you like this proposal? https://gist.github.com/GenevieveBuckley/eeeae8a803dc0d955354dcd365eef7ad

napari/layers/points/points.py
```diff
@@ -1558,6 +1559,40 @@ def on_mouse_release(self, event):
                 self.selected_data = []
             self._set_highlight(force=True)
 
+    def to_table(self, path=None):
```
sofroniewn: love it!

napari/layers/points/points.py
```diff
@@ -1558,6 +1559,40 @@ def on_mouse_release(self, event):
...
+        Notes
+        -----
+        You can construct a pandas dataframe from the results like this:
+        ```
+        table, column_names = my_points.to_table()
+        points_dataframe = pandas.DataFrame(table, columns=column_names)
+        ```
+        """
+        n_dimensions = self.data.shape[1]
+        column_names = ['coord_id'] + [
```
sofroniewn: Is there a reason to have the `coord_id` field? I guess it maybe makes this table look more similar to the table from shapes which is nice, but it is always just `0, 1, 2, ... , N` right? I'm fine either way, but curious what the thinking was

napari/utils/io.py
```diff
@@ -127,3 +128,25 @@ def read_zarr_dataset(path):
     else:
         raise ValueError(f"Not a zarr dataset or group: {path}")
     return image, shape
+
+
+def write_csv(path, data, column_names=None):
```
sofroniewn, line 133: Out of curiosity if I did

```python
table, column_names = my_points.to_table(path='my_file_1.csv')
points_dataframe = pandas.DataFrame(table, columns=column_names)
points_dataframe.to_csv('my_file_2.csv')
```

Would 'my_file_1.csv' and 'my_file_2.csv' be identical? And if I did 
```python
my_new_dataframe = pandas.read_csv('my_file_1.csv')
```
Would `points_dataframe == my_new_dataframe`?

Seems like these would all be nice if True, but not essential

napari/layers/shapes/shapes.py
```diff
@@ -1768,3 +1769,39 @@ def on_mouse_release(self, event):
...
+        tuple
+            (table, column_names) where table is a list of lists, and
+            column_names is a list of strings.
+
+        Notes
+        -----
+        You can construct a pandas dataframe from the results like this:
+        ```
+        table, column_names = my_shapes.to_table()
+        points_dataframe = pandas.DataFrame(table, columns=column_names)
```
sofroniewn, line 1794: Might be nice to add `pandas` as a test dependency and explicitly test that this construction works.

sofroniewn (commented):
This is looking good. I do wonder if we should provide some support to load this data back into napari - I'd probably add a path argument to `add_points` and `add_shapes` that when pointed at a file will load the data back in and create a new layer using that data - similar to how it is done inside `add_image` - see https://github.com/napari/napari/blob/master/napari/components/add_layers_mixin.py#L146-L151
(You'd also need to propagate the path arg to `view_points` and `view_shapes`). I think I'd like to see this in this PR as it will make it clear how to work with our tabular format. We can also add some nice roundtrip tests that just check the data can make a round trip.

I think things like GeoJSON etc for shapes can come later.

napari/layers/points/points.py
GenevieveBuckley: Yes, I wanted to keep consistent column naming across tables produced by the points and shapes layers. I'm open to suggestions; the conventions I've picked are arbitrary.

napari/utils/io.py
GenevieveBuckley: The files and pandas dataframes are the same **if you use `index=None` when writing out the pandas dataframe to csv**. 

I think this way is easiest, and what we should recommend:

```python
table, column_names = my_points.to_table(path='my_file_1.csv')
points_dataframe = pandas.DataFrame(table, columns=column_names)
points_dataframe.to_csv('my_file_2.csv', index=None)
my_new_dataframe = pandas.read_csv('my_file_2.csv')
```

Alternatively, the files and pandas dataframes are the same **if you set the index column equal to "coord_id"**.

```python
table, column_names = my_points.to_table(path='my_file_1.csv')
points_dataframe = pandas.DataFrame(table, columns=column_names).set_index('coord_id')
points_dataframe.to_csv('my_file_2.csv')
my_new_dataframe = pandas.read_csv('my_file_2.csv', index_col='coord_id') 
```

Result in both cases:
* 'my_file_1.csv' and 'my_file_2.csv' have the same contents.
*  `np.allclose(points_dataframe, my_new_dataframe)` returns True.

The downside is, this isn't the pandas default so it might be a little hard for people to remember. But I really don't want to write out the useless pandas index to file, so I think we'll just have to live with it. 

We can add a file to the examples that describes essentially the same thing as this comment, if you like.

jni: Why do we have a thing called `coord_id`? I'd prefer an index-less dataframe without point ids...

GenevieveBuckley: I did that for consistency with the shapes csv, although I'm open to dropping that column.

I think we wanted to keep both the 'shape_id' and 'coord_id' for the shapes csv, particularly because the order of coordinates in a shape is important.

### Shapes csv
What we have now, what we plan to keep:

| shape_id | shape_type | coord_id | dim_0 | dim_1 |  dim_2 | ... | dim_n |
|----------|------------|----------|-------|-------|--------|-----|-------|
| 0        | Ellipse    | 0        |       |       |        |     |       |
| 0        | Ellipse    | 1        |       |       |        |     |       |
| 0        | Ellipse    | 2        |       |       |        |     |       |
| 0        | Ellipse    | 3        |       |       |        |     |       |
| 1        | Rectangle  | 0        |       |       |        |     |       |
| 1        | Rectangle  | 1        |       |       |        |     |       |
| ...      | ...        | ...      |       |       |        |     |       |
| n        | Path       | n-1      |       |       |        |     |       |
| n        | Path       | n        |       |       |        |     |       |

### Points csv
What we have now:

| coord_id | dim_0 | dim_1 | ... | dim_n |
|----------|-------|-------|-----|-------|
| 0        |       |       |     |       |
| 1        |       |       |     |       |
| ...      |       |       |     |       |

What @jni would like instead:

| dim_0 | dim_1 | ... | dim_n |
|-------|-------|-----|-------|
| .      |       |     |       |
| .      |       |     |       |
| .      |       |     |       |

GenevieveBuckley:
Ok this is everything I'm planning on doing in this PR, unless there are any other requests.

sofroniewn:
This has been superseded by #1193 - thanks @GenevieveBuckley for the great work getting this going!!

PR #942, opened 2020-02-03
Sphinx docs restructuring (create space for dev-focused prose)

tlambert03:
# Description
I was going to start writing up some developer-focused documentation to explain plugins in napari.  But with the current dev-docs being dedicated just to the auto-generated API docs, the main landing page and organization of the files is a bit hard to extend.  This PR restructures the docs folder to be a bit more like a standard sphinx install, and preps the way for expansion as follows:

- separates sphinx-docs specific stuff (`.rst` files and `conf.py`) into the `docs/source` folder, as the standard sphinx-quickstart does, so it is no longer mixed up with all of our markdown docs that do not get used for sphinx.  (I put those in `docs/md` but happy to move them).
- `docs/source` is now checked into version control, and rather than copying `conf.py` and `index.rst` into `source` before running `sphinx-autoapi`, `sphinx-autoapi` just puts docs into a subfolder `docs/source/api` that is added to .gitignore (the previous arrangement was going to require explicitly adding every new `.rst` file that we made to our makefile).
- ignored tests from `autoapi` generation.
- The autogenerated api docs are now just an "API Reference" link on the sidebar, so we can use index.rst as a proper docs landing page.
- starting a page for plugin explanations, and a developer-focused "concepts in napari" page to explain the main classes like the `Viewer` and `Layer` models.
- some other niceties, like a logo and proper version catching.

here's how it would look:

Screen Shot 2020-02-03 at 1 34 29 PM

happy to get feedback on the structure.  this is [WIP] until I finish writing the plugin docs (and if anyone else also wants to work on `concepts.rst` that would be great, feel free to push directly to this).  Specifically interested in feedback from @kne42 and #764


kne42 (approved):
🙇‍♀ Thank you! 🙇‍♀

 I've built the docs locally and I love how much cleaner they look!

I have been researching sphinx for the past few days since I've basically never touched it before so it helps a lot that you have experience and are tackling this :)

I don't know if you already have plans to do this part but I'd love to add a section under the plugins guide that pertains to published packages that want to register as a napari plugin but don't want to list napari as a requirement (basically a try-import statement that defines dummies if napari is not installed).

tlambert03:
> I've built the docs locally and I love how much cleaner they look!

yay!

> I don't know if you already have plans to do this part but I'd love to add a section under the plugins guide that pertains to published packages that want to register as a napari plugin but don't want to list napari as a requirement (basically a try-import statement that defines dummies if napari is not installed).

yeah that would be great.   I hadn't yet thought of a specific section on that, so feel free to start one! Actually, it will be pretty easy for packages to be napari plugins without depending on or even importing napari.
If we explore that idea for a moment, the only thing a package _really_ needs to do is to mark one or more of their functions as conforming to one of our `hookspecs` by using a pluggy `@hookimpl` decorator.  (well, at the moment, they actually _must_ name that function the same thing as one of our hookspecs, but if/when https://github.com/pytest-dev/pluggy/pull/251 gets merged, then they can just use `@hookimpl(specname='napari_hookspec_name')`

They can get that decorator directly from pluggy (which is effectively just stdlib python), without importing napari:
```python
from pluggy import HookimplMarker
hookimpl = HookimplMarker("napari")

@hookimpl(specname='napari_get_reader')
def their_function():
    ...
```
and then, for discovery, if the plugin isn't named something like "napari-plugin" ... they would just need to add an `entry_point` ("napari.plugin") to their setup.py and point it to the module that has the decorated function.

I will definitely put some of this in what I'm writing, but feel free to also push to this PR.

kne42:
> They can get that decorator directly from pluggy (which is effectively just stdlib python), without importing napari:
> 
> ```python
> from pluggy import HookimplMarker
> hookimpl = HookimplMarker("napari")
> 
> @hookimpl(specname='napari_get_reader')
> def their_function():
>     ...
> ```

Ohhhhh I didn't realize that, I thought that they would have to do something like `from napari import hook_impl` so this is much nicer!

You can disregard what I wrote above then, but I would still be happy to contribute pages/sections if you have any ideas :)

tlambert03:
yeah, I was happy when I realized that too :)

sofroniewn (approved):
This looks great @tlambert03 and sounds like it is good to go. Does the merge order here matter / is there other stuff to go in before merge (wasn't quite clear to me if there might be)

tlambert03:
order doesn't matter here.  and the only reason I had made this WIP was cause I intended to populate it with some more stuff.  But since the docs will always be a WIP, and since the dev docs aren't actively used yet, it would be fine to merge, and additional content can be added in later PRs.  The most glaring omission at the moment is [`concepts.rst`](https://github.com/napari/napari/pull/942/files#diff-57faa53de3150ca9f9215bff0c410a25), which trails off mid-sentence :)

docs/source/concepts.rst
```diff
@@ -0,0 +1,16 @@
...
+
+At the heart of napari is the viewer model.  For a tutorial introduction to the
+viewer, see the `napari viewer tutorial <https://napari.org/tutorials/viewer>`_.
+
+The :class:`~napari.viewer.Viewer` object contains, among other things:
+
+- the main canvas
+- an instance of :class:`~napari.components.layerlist.LayerList` - a list of
+  layer objects that have been added to the :class:`~napari.viewer.Viewer`
+- an instance of
```
sofroniewn: Can we remove this for now. I'd like to merge what we have which does a lot of reorg stuff and then make dedicated follow up PRs for content. Would be nice not to have something trailing off ...... :-)

tlambert03: cleaned it up.  Plugins.rst is still there, but I removed all references to it, so it's mostly undiscoverable.  I'll add to it soon, but we can merge this if you want.

jni:
A few comments (non-blocking):

- I don't love the `md` in the path for the markdown documents. I would prefer if the paths were reflective of the *content*, not the format.
- Does sphinx generate API docs for our private APIs? If so, is there a way to configure it to put those into a separate "developer API" or "private APIs" section?
- the logo is a bit huge imho.

Also, we should probably start to think about our "standard image" for napari screenshots. Nothing wrong with astronaut, but I think we can do better. =)

tlambert03:
> A few comments (non-blocking):
> 
> * I don't love the `md` in the path for the markdown documents. I would prefer if the paths were reflective of the _content_, not the format.

definitely agree.  was looking for some opinions on that... I don't care where it goes or what we name it as long as they're separate from the sphinx rst files.  most people put those in the top level, but we have so many of them :) . I'm not entirely sure they even belong in docs.  tell me where to move 'em :)

> * Does sphinx generate API docs for our private APIs? If so, is there a way to configure it to put those into a separate "developer API" or "private APIs" section?

it's an option ( `--private` in [sphinx-apidoc](https://www.sphinx-doc.org/en/master/man/sphinx-apidoc.html)) and it was enabled previously, so I left it enabled.  I can disable it, which would remove all the _qt, and _vispy stuff (but even stuff in there could be selectively/manually added with autodoc directives).  It would be harder to separate the two, since I don't think there's a "private only" option in `apidoc`.


> * the logo is a bit huge imho.

I agree! :)  I had made it 60% smaller since that first image. now looks like this,  ok?

Screen Shot 2020-02-04 at 6 10 22 PM


> Also, we should probably start to think about our "standard image" for napari screenshots. Nothing wrong with astronaut, but I think we can do better. =)

I totally agree :)  let's have a contest!

jni:
> tell me where to move 'em :)

`project/`? `developers/`? `governance/`? `management/`? Would be fine with any of those, or similar.

> so I left it enabled

Yeah I would disable for now. I thought it was extremely cluttered before. I would like people browsing those docs to have the same experience as browsing the skimage API reference, ie, short and to the point.

> there's not a private-only option

Well, we could have a whole new hierarchy with private *and* public APIs, and label that "private". I get a feeling that private stuff overwhelms the public stuff so it's not a big deal to have the public stuff in there also.

> I had made it 60% smaller since that first image. now looks like this, ok?

I'd go smaller still, but it's just a matter of taste at this point, that is a reasonable size. =) In my opinion it should probably be about the size of the icon in your OS dock.

tlambert03:
will do on all accounts.

>I would like people browsing those docs to have the same experience as browsing the skimage API reference, ie, short and to the point.

me too!  Soon after this PR, I'd really like to have a slightly more guided dev-focused docs that points to the API references, but with a little bit more curation and narrative.
I'll disable private stuff, and see if I can't add a few of them in their own section.

docs/source/conf.py
```diff
@@ -27,6 +28,26 @@
...
+                    'napari is a fast',
+                    "It's designed for",
+                    "images. It's built",
+                    "rendering), and ",
+                )
+            ):
+                print(line, end='')
+
+
+clean_release_notes()
```
jni, line 50: I wouldn't necessarily use this function... I'm pretty happy to leave all the announcement stuff in there...

tlambert03: the main reason I did that was because of the title saying "Announcement".  It makes the table of contents look ridiculous (each one says announcement).  I don't mind leaving the first paragraph.  Actually, I would recommend we just remove the word "announcement" from the heading of our release template.

tlambert03:
ok, cleaned all that stuff up.  The autogenerated API docs still need work.  It's a big improvement... but it will be a continual work in progress.  It's hard to make it look super clean and elegant while also using a tool like `sphinx-apidoc` to just automatically do everything.  I'm at my wits end for the moment with `sphinx-apidoc`.  I think the purpose of this PR has been served

docs/source/conf.py
jni: Eh, I'm quite happy to keep even the announcement heading — it just makes it look like a news feed, which is fine. But I'm ok with where we are now. =)

jni (approved):

jni:
@tlambert03 🎉 🙏

Issue #669, opened 2019-11-09
switching themes causes incomplete updates


jni:
## 🐛 Bug

The 'light' theme for napari is looking pretty terrible these days:

<img width="1200" alt="Screen Shot 2019-11-09 at 8 58 44 pm" src="https://user-images.githubusercontent.com/492549/68526810-c2860580-0333-11ea-8455-4bea32aedf01.png">

We should update it and make sure that any PRs that touch the UI are tested with both themes.

Also worth noting that the base layer needs to have blending mode "translucent" — with additive, you get a totally white canvas. Perhaps it's a chance to play with inverted LUTs or something, not sure.

Finally, I'm not sure whether there is a way to set the theme globally. I could set it on a Viewer instance, but I could not set it on the `napari.Viewer` class. It seems to me that it should be more of a global than an instance setting.

## To Reproduce

```python
import napari
v = napari.Viewer()
v.theme = 'light'
```

This crashes:

```python
import napari
napari.Viewer.theme = 'light'
v = napari.Viewer()
```

This makes sense because it replaces a property with a text string! But I couldn't find how else to set the theme.

jni:
Bonus issue: changing back to dark doesn't work fully:

```python
v.theme = 'dark'
```

<img width="1200" alt="Screen Shot 2019-11-09 at 9 17 38 pm" src="https://user-images.githubusercontent.com/492549/68527028-78525380-0336-11ea-9c7e-bbad8c98e504.png">

sofroniewn:
good to flag!! I'll take a look. our stylesheets need some love in general too

marciovm:
Any reason not to deprecate theme options, and make dark the only option? Having theme options makes UI iteration much more expensive.

sofroniewn:
@jni theme looks fine on my mac so this might be something cross platform :-( 

<img width="1200" alt="Screen Shot 2019-11-11 at 8 17 39 AM" src="https://user-images.githubusercontent.com/6531703/68602853-f52b1d80-045b-11ea-90bc-7b40282e4119.png">

@marciovm we've had comments from people who prefer `light` to `dark`. The way we've set it up should make it pretty easy to maintain both - we have color palettes and style using the palettes, but something funky seems to be happening above :-(

AhmetCanSolak:
@sofroniewn maybe not right place to discuss but isn't it better to keep canvas background black on both themes?

sofroniewn:
[...]

PR #943, opened 2020-02-04
Toggle theme

boonepeter:
# Description

I am finally getting around to issue #669. This does not fix that issue, but it enables the user to toggle between light and dark modes from the GUI View menu. 

 I tried to make sure that the check mark would get updated if the palette was updated elsewhere. That seemed a little hack-y to me so I welcome any other ideas if there is a better way. Still learning the Qt world.

I will borrow a mac and try to fix the weird updating bug that others are seeing.

## Type of change

# References

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/_qt/qt_main_window.py
+++ b/napari/_qt/qt_main_window.py
@@ -306,7 +306,9 @@ class Window:
             template('QWidget { background: {{ background }}; }', **palette)
         )
         self._qt_window.setStyleSheet(template(self.raw_stylesheet, **palette))
-
+        for action in self.view_menu.actions():
+            if action.text() == "Dark mode":
+                action.setChecked(self.qt_viewer.viewer.theme == "dark")
     def _status_changed(self, event):
         """Update status bar.
         """
--- a/napari/_qt/qt_main_window.py
+++ b/napari/_qt/qt_main_window.py
@@ -123,7 +123,12 @@ class Window:
         else:
             self.main_menu.setVisible(True)
             self._main_menu_shortcut.setEnabled(False)
-
+    
+    def _toggle_theme(self):
+        if self.qt_viewer.viewer.theme == "dark":
+            self.qt_viewer.viewer.theme = "light"
+        elif self.qt_viewer.viewer.theme == "light":
+            self.qt_viewer.viewer.theme = "dark"
     def _add_file_menu(self):
         open_images = QAction('Open image(s)...', self._qt_window)
         open_images.setShortcut('Ctrl+O')
@@ -146,8 +151,14 @@ class Window:
         toggle_visible.setShortcut('Ctrl+M')
         toggle_visible.setStatusTip('Hide Menubar')
         toggle_visible.triggered.connect(self._toggle_menubar_visible)
+        dark_mode = QAction("Dark mode", self._qt_window)
+        dark_mode.setStatusTip("Toggle dark/light modes")
+        dark_mode.setCheckable(True)
+        dark_mode.triggered.connect(self._toggle_theme)
+        dark_mode.setChecked(self.qt_viewer.viewer.theme == "dark")
         self.view_menu = self.main_menu.addMenu('&View')
         self.view_menu.addAction(toggle_visible)
+        self.view_menu.addAction(dark_mode)
 
     def _add_window_menu(self):
         exit_action = QAction("Close window", self._qt_window)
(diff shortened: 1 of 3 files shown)


tlambert03:
Hi @boonepeter.  thanks for the PR!  this looks/works good for me.  
Ultimately, I think I'd like to see this functionality in a preferences menu, with a dropdown selection instead of a boolean choice (in case we expand theme options, or allow plugins).  I implemented that in the settings PR #834 ([here is an example](https://github.com/napari/napari/pull/834#issuecomment-569544298))... but I think it might be a while until there is sufficient demand to get that in.  So I have no problem with this solution for now.

will leave it for @sofroniewn to also review.  And you'll need to fix the black/flake8 errors before it can merge.

thanks again!

boonepeter:
Thanks @tlambert03! Your #834 looks pretty slick and is definitely the way to go. I definitely won't be offended if this doesn't make the cut as it is just a small thing I wanted to add while looking into that other issue.

tlambert03:
yeah, but that could take a while, and this is minimal.  The only possible argument against it is that we might eventually remove it.  but I don't mind that.  I'll approve this but leave the final decision to @sofroniewn.

tlambert03 (approved):

napari/_qt/qt_main_window.py
```diff
@@ -124,6 +124,12 @@ def _toggle_menubar_visible(self):
             self.main_menu.setVisible(True)
             self._main_menu_shortcut.setEnabled(False)
 
+    def _toggle_theme(self):
```
sofroniewn: I'd prefer to see this function look like the toggle_theme function in our `napari/keybindings.py` https://github.com/napari/napari/blob/c01024f151546f7a2ada9727572efb3e65fa4f0f/napari/keybindings.py#L15
Can you replace the code here with that code (keeping the name `_toggle_theme` and remove the code around `light` and `dark`. The reason for this is that if people add more themes it will allow them to cycle through them instead of just having to alternate between our two fixed ones

napari/_qt/qt_main_window.py
```diff
@@ -146,8 +152,14 @@ def _add_view_menu(self):
         toggle_visible.setShortcut('Ctrl+M')
         toggle_visible.setStatusTip('Hide Menubar')
         toggle_visible.triggered.connect(self._toggle_menubar_visible)
+        dark_mode = QAction("Dark mode", self._qt_window)
```
sofroniewn: I think I'd rather see the `QAction` be called `Toggle theme`, with the status tip `Toggle theme` and the keybinding `Control+Shift+T` as is used https://github.com/napari/napari/blob/c01024f151546f7a2ada9727572efb3e65fa4f0f/napari/keybindings.py#L15. There shouldn't be any special references to light / dark here

Once this is done you can remove the code here https://github.com/napari/napari/blob/c01024f151546f7a2ada9727572efb3e65fa4f0f/napari/keybindings.py#L15

napari/_qt/qt_main_window.py
sofroniewn: Ah, also this function should maybe live inside the `ViewerModel` https://github.com/napari/napari/blob/master/napari/components/viewer_model.py instead of `qt_main_window` as it only needs to reference the Viewer

sofroniewn (changes requested):
Hi @boonepeter thanks for adding this! I've requested some changes as we have some functionality for toggling themes right now that allows you to cycle through our current themes here https://github.com/napari/napari/blob/c01024f151546f7a2ada9727572efb3e65fa4f0f/napari/keybindings.py#L15, and I think we can move that code inside the ViewerModel and then reuse it in a drop down menu.

The code allows for the cycling through themes - which maybe a little annoying if you have many themes added, but is also a little more flexible than if you just have the two default ones. I think in the long run we'll probably have a solution more along the lines of #834.

@tlambert03 does this approach seem reasonable too?

napari/_qt/qt_main_window.py
boonepeter: I like it. I copied this and moved it to `viewer_model`

napari/_qt/qt_main_window.py
boonepeter: I renamed this and removed the keybinding. The keybinding no longer shows up in the "keybindings help" window, but I guess that isn't an issue since it is visible in the main menu. 

I changed `Ctrl-Shift-O` to `Ctrl+Shift+O` for the open folder shortcut so that Qt recognizes this and displays the shortcut it in the menu.

sofroniewn (approved):
Thanks for making those changes so quickly @boonepeter. This looks great, I have confirmed it works for me. I will approve and then merge shortly.

PR #944, opened 2020-02-04
Add "Screenshot" option to File menu

GenevieveBuckley:
# Description

This PR adds a "Screenshot" item to the file menu, saving a screenshot of the viewer to disk. It makes the existing screenshot() method usable from the GUI.

There is no hotkey shortcut attached to this menu item, but we can add one if there is consensus on what the hotkey combination to use.

## Type of change

# References

# How has this been tested?

- [x] test added to `napari/_qt/_tests/test_qt_viewer.py::test__save_screenshot`

## Final checklist:


jni (approved):
Looks great! I'd be ok with adding Ctrl-Alt-S for the screenshot, but agree that it might be contentious. =)

napari/_qt/_tests/test_qt_viewer.py
```diff
@@ -177,6 +181,47 @@ def test_screenshot(qtbot):
     view.shutdown()
 
 
+def test__save_screenshot(qtbot, tmpdir):
```
kevinyamauchi:
```suggestion
def test_save_screenshot(qtbot, tmpdir):
```

napari/_qt/qt_viewer.py
```diff
@@ -228,6 +229,23 @@ def screenshot(self):
         img = self.canvas.native.grabFramebuffer()
         return QImg2array(img)
 
+    def _save_screenshot(self):
```
kevinyamauchi, line 305: What do you think about allowing users the option to pass in a file name as well (with the default behavior being as it is now)? I could see `_save_screenshot()` being a useful method for people who want to script screen shots (e.g., load in data set, move to a specific view, then take a screen shot) and I think being able to set the filename would be nice for that use case.

kevinyamauchi (commented):
This looks great to me, @GenevieveBuckley ! I left a couple of minor comments below.

I was also curious about what you think about adding a passthrough method to Viewer so that users can call `viewer.save_screenshot(filename)` rather than `viewer.qt_viewer._save_screenshot()`. As I mentioned below, I could see users wanting to use the functionality you added here from the command line or scripted in functions attached to keybindings. That being said, I think the work here is great as is, so it's totally fine with me to punt and think about adding it in a later PR.

napari/_qt/qt_viewer.py
GenevieveBuckley: Maybe it would be better to have the option of passing in a filename to `screenshot()` itself? Seems simpler overall for users to remember. What do you think about using the same structure @sofroniewn uses here with the `file` kwarg in `to_svg()` https://github.com/napari/napari/blob/7a04984472db9a0b74c4efc26d56a6f56b4ce9dd/napari/layers/base/base.py#L657
```python
        if file:
            # Save svg to file
            with open(file, 'w') as f:
                f.write(svg)

        return svg
```

kevinyamauchi: Adding a 'filename' kwarg to `Viewer.screenshot()` works for me!

sofroniewn: That works for me too. We might want to propagate that keyword up to the screenshot method defined here too https://github.com/napari/napari/blob/c01024f151546f7a2ada9727572efb3e65fa4f0f/napari/viewer.py#L64.

napari/_qt/qt_main_window.py
```diff
@@ -137,9 +137,14 @@ def _add_file_menu(self):
         )
         open_folder.triggered.connect(self.qt_viewer._open_folder)
 
+        screenshot = QAction('Screenshot', self._qt_window)
+        screenshot.setStatusTip('Take screenshot')
```
sofroniewn: Do you want to add a shortcut too, maybe `screenshot.setShortcut("Ctrl+Alt+S")`?

sofroniewn: Can the status tip maybe be `Save screenshot` and the QAction be `Save Screenshot` or do you think that is implied / redundant?

napari/_qt/qt_viewer.py
sofroniewn: We're starting to use `path` quite a bit in the codebase, maybe a better choice than `file` or `filename` and something we might want to change in our `to_svg` methods

sofroniewn (approved):
Made some minor comments, I like to proposed change around adding `path` to `Viewer.screenshot()`

Thanks for adding this @GenevieveBuckley. I'm not sure if this comment has been made elsewhere, but it might be nice to add a similar "Save svg" option to our File menu in a followup PR

napari/_qt/qt_main_window.py
sofroniewn: > Looks great! I'd be ok with adding Ctrl-Alt-S for the screenshot, but agree that it might be contentious. =)

lol - I just read this, I guess we're aligned ;-)

napari/viewer.py
```diff
@@ -61,11 +61,13 @@ def __init__(
         self.window = Window(qt_viewer)
         self.update_console = self.window.qt_viewer.console.push
 
-    def screenshot(self, with_viewer=False):
+    def screenshot(self, path=None, with_viewer=False):
```
GenevieveBuckley: There's an API change here as a result of adding the `path` kwarg we discussed.

I think `path` should go before the `with_viewer` kwarg because:
1. It keeps this function consistent with all the other screenshot related functions
2. I think use of `path` will be a lot more common than use of `with_viewer`, and I really want people to be able to write: `viewer.screenshot('path/to/screenshot.png')`

GenevieveBuckley: Are you happy with this, @sofroniewn & @jni?

jni: Works for me. Might as well make `with_viewer` keyword-only, which it should have been from the start.

sofroniewn: Yup, agreed - lets add make it `screenshot(self, path=None, *, with_viewer=False)` so `with_viewer` is keyword-only

napari/_qt/qt_viewer.py
```diff
@@ -216,7 +217,7 @@ def _update_camera(self):
                 self.view.camera.viewbox_key_event = viewbox_key_event
                 self.viewer.reset_view()
 
-    def screenshot(self):
+    def screenshot(self, path=None):
```
sofroniewn, line 286: Can you add a doc string for `path` here (same one you used in `Viewer.py` is fine.

napari/_qt/qt_main_window.py
```diff
@@ -311,7 +319,7 @@ def _help_changed(self, event):
         """
         self._help.setText(event.text)
 
-    def screenshot(self):
+    def screenshot(self, path=None):
```
sofroniewn, line 329: Can you add a doc string for `path` here (same one you used in `Viewer.py` is fine).

napari/_qt/qt_viewer.py
```diff
@@ -226,8 +227,26 @@ def screenshot(self):
...
 
+    def _save_screenshot(self):
+        """Save screenshot of currently displayed screen to file.
+
+        Returns
+        -------
+        filename : str
+            File path of the saved screenshot.
+        """
+        filename, _ = QFileDialog.getSaveFileName(
```
sofroniewn, line 307: Is there a way to autopopulate the save name space with something that has a `png` extension, or a clean way for us to automatically add a `png` extension if someone just types in a string.

I just tried to take a screenshot, typed in something like `astro` and then got this error

```python-traceback
(base) czirwc1macos2701:napari nsofroniew$ python examples/add_image.py
  File "/Users/nsofroniew/GitHub/napari/napari/_qt/qt_viewer.py", line 248, in _save_screenshot
    self.screenshot(path=filename)
  File "/Users/nsofroniew/GitHub/napari/napari/_qt/qt_viewer.py", line 231, in screenshot
    imsave(path, QImg2array(img))  # scikit-image imsave method
  File "/Users/nsofroniew/opt/anaconda3/lib/python3.7/site-packages/skimage/io/_io.py", line 131, in imsave
    return call_plugin('imsave', fname, arr, plugin=plugin, **plugin_args)
  File "/Users/nsofroniew/opt/anaconda3/lib/python3.7/site-packages/skimage/io/manage_plugins.py", line 210, in call_plugin
    return func(*args, **kwargs)
  File "/Users/nsofroniew/opt/anaconda3/lib/python3.7/site-packages/imageio/core/functions.py", line 302, in imwrite
    writer = get_writer(uri, format, "i", **kwargs)
  File "/Users/nsofroniew/opt/anaconda3/lib/python3.7/site-packages/imageio/core/functions.py", line 226, in get_writer
    "Could not find a format to write the specified file " "in mode %r" % mode
ValueError: Could not find a format to write the specified file in mode 'i'
```

If I typed `astro.png` it worked fine. Maybe @tlambert03 has a suggestion on how to do this well

napari/_qt/qt_viewer.py
```diff
@@ -226,8 +227,26 @@ def screenshot(self):
...
+            parent=self,
+            caption='',
+            directory=self._last_visited_dir,  # home dir by default
+        )
```
sofroniewn, line 313: I'm also curious here what would it be like to add a checkbox to the defualt `QFileDialog` that allowed people to save the screenshot using the `Viewer.screenshot(..., with_viewer=True)` option when that checkbox is selected. Personally it's quite nice for me to be able to take screenshots with the viewer, but I completely recognize it's a minority use case, so I would make it unchecked by default. But I'm curious if it is possible / how it would feel. We'd also need to move some of the functions around inside this PR to accommodate that, but it shouldn't be too hard. (I could even see a case then for moving `Viewer.screenshot` to `Viewer.window.screenshot` and putting that on the `window` as it is a bit of an outlier on `Viewer`, but this is a separate thread).

Curious what does @tlambert03 think of the checkbox idea?

sofroniewn (commented):
This is looking great. I've left a couple comments, two around doc strings, one about auto adding `png` extensions, one on the `with_viewer` keyword arg only, and then one question around the possibility of having a checkbox that allows for `with_viewer` screenshots

napari/_qt/qt_viewer.py
GenevieveBuckley: I don't really see the point of adding this feature, I assume anyone who wants to get that end result will just use their system screenshot/clipping tool.

napari/_qt/qt_viewer.py
GenevieveBuckley: If we do pathname manipulations, I'd like that to be something that happens as a convenience *only* to strings returned through GUI interaction. 

There are two main options:
1. Append ".png" to the end of whatever string is provided. Possible disadvantage: this does mean it is possible to wind up with filenames like "file.png.png" or "file.tif.png".

2. Use `os.path.splitext()` on whatever string is provided and replace the extension with ".png". Possible disadvantage: this means it's possible to lose some filename information you might have intended to keep, for instance if you type "file.name" expecting to get "file.name.png" instead you'll wind up with "file.png". 

**I'm leaning towards option 1, appending ".png" to whatever string was returned from the file menu "Screenshot" option.** 

I prefer this because:
* there's no possible loss of information that was typed by the user
* there is no pathname manipulation if you are using the API for scripting (I would find that super unintuitive)

GenevieveBuckley:
> This is looking great. I've left a couple comments, two around doc strings, one about auto adding `png` extensions, one on the `with_viewer` keyword arg only, and then one question around the possibility of having a checkbox that allows for `with_viewer` screenshots

* docstrings fixed, thanks!
* we now auto-append ".png" to whatever path is returned using the GUI ([my reasoning on that decision is here](https://github.com/napari/napari/pull/944/files#r375588938))
* `with_viewer` is now a keyword only argument, thanks!
* on the possibility of having a checkbox that allows for `with_viewer` screenshots: I'm not planning to do this, mostly because I believe people will probably just use their system screenshot/clipping tools if they want this end result. I'm leaving this suggestion open for someone else to pick up if they want to implement it :)

napari/_qt/qt_viewer.py
sofroniewn: It's certainly not essential. Maybe at best it's only mildly convenient for the core team / people that actually want the viewer to appear in their slides. Absolutely fine to ignore, part of what motivated the question was around understanding what it's like to add custom checkboxes / QWidgets to the default save menu as I could imagine a context where we want some additional options for people and maybe don't want another pop-up or entirely custom dialog

napari/_qt/qt_viewer.py
sofroniewn: I definitely agree with "convenience only to strings returned through GUI interaction". I'd prefer to use the built-in Qt Dialog methods for this though, sorry I didn't link to them earlier (I had to do a bit of reading up on them now before replying as I wasn't too familiar with them before). They have a concept called `NameFilters` - see here https://doc.qt.io/qt-5/qfiledialog.html#setNameFilters, that allows the desired behaviour. I think adding `filter="Images (*.png *tif *.tiff)",` to `QFileDialog.getSaveFileName` will achieve what we want. If no extension is provided then `.png` will be added. If any one of the extensions listed are provided they will be used, and if a different extension is provided you will get a standard warning and an option to save with both `.xyz.png` or just `.xyz` if that is what you had typed. You can then drop the `+.png`  you were adding.

Can you also adjust the tooltip to say something like `as png by default` or drop the mention of `png` entirely

sofroniewn:
Sounds good, I reopened the comment about the `png` and made a suggestion to use some built-in Qt machinery to essentially achieve that. I think their functionality is a little nicer and cleaner.

Tests are failing on `test_save_screenshot` right now, and we're getting some formatting / linting errors but after those are fixed this should be good to go. Thanks again!!

napari/_qt/qt_viewer.py
GenevieveBuckley: I can't make this work, sorry. I've tried several different permutations on this idea but it seems that the filters only affect what is visible to you in the dialog box, not the value of what is returned. This seems relevant - https://bugreports.qt.io/browse/QTBUG-27186

There is also supposed to be a `setDefaultSuffix`, but I can't get that working properly with getSaveFilename either :(

sofroniewn: No worries, there are sometimes bugs / weird things with Qt, and sometimes they vary depending on what platform you're on!!! I can take a look at this tomorrow, or maybe @tlambert03 has some experience here and can also take a look (If you can @tlambert03 it would be appreciated as I am quite busy tomorrow). If either of us can get it working we'll push the changes directly to this PR if that's ok, if we can't then we can go with your solution or one where we implement a little bit of this logic ourselves. Thanks for trying!!

sofroniewn: @GenevieveBuckley I've gotten this working for me now, I just pushed the changes directly to the branch. Can you check them out? There was a typo in what I pasted above `(*.png *tif *.tiff)` vs `(*.png *.tif *.tiff)` which will have caused all sorts of problems if not caught, my apologies. I've included a gif in the main thread of the behavior for me.

sofroniewn:
I have now adjusted this PR to use the Qt builtin file extension adding behavior which is quite nice - see here 

save_dialog

If this looks good to @GenevieveBuckley then I think it is good to go.

Thanks again @GenevieveBuckley. Also if you're interested in adding saving dialogs for images, labels, points and shapes (after #941) and svgs I think those would all be appreciated additions.

napari/_qt/qt_viewer.py
GenevieveBuckley: No, this fails for me. 

The problem must be platform dependent (running Linux Ubuntu here), it's the same as when I tried earlier. Essentially, if you give it a filename without an extension it won't automatically add one for you like it is supposed to do. The Qt dialog is definitely *supposed* to work the way you expect, but it seems too buggy to rely on here.

Here's my system and napari info:
```
napari: 0.2.11+17.g66fcdc79
Platform: Linux-5.3.0-28-generic-x86_64-with-debian-buster-sid
Python: 3.7.4 (default, Aug 13 2019, 20:35:49) [GCC 7.3.0]
Qt: 5.14.0
PySide2: 5.14.0
VisPy: 0.6.4
NumPy: 1.18.1
SciPy: 1.4.0.dev0+1566e8e
scikit-image: 0.16.2
Dask: 2.10.0

GL version: 4.6.0 NVIDIA 435.21
MAX_TEXTURE_SIZE: 32768
```

`pytest napari/_qt/_tests/test_qt_viewer.py::test_save_screenshot` fails on my system.
The relevant traceback is here (imageio fails because it wasn't given a filename with a file extension):

```python-traceback
  File "/home/genevieve/Documents/GitHub/napari/napari/_qt/qt_viewer.py", line 249, in _save_screenshot
    self.screenshot(path=filename)
  File "/home/genevieve/Documents/GitHub/napari/napari/_qt/qt_viewer.py", line 236, in screenshot
    imsave(path, QImg2array(img))  # scikit-image imsave method
  File "/home/genevieve/anaconda3/envs/napari-dev/lib/python3.7/site-packages/skimage/io/_io.py", line 131, in imsave
    return call_plugin('imsave', fname, arr, plugin=plugin, **plugin_args)
  File "/home/genevieve/anaconda3/envs/napari-dev/lib/python3.7/site-packages/skimage/io/manage_plugins.py", line 210, in call_plugin
    return func(*args, **kwargs)
  File "/home/genevieve/anaconda3/envs/napari-dev/lib/python3.7/site-packages/imageio/core/functions.py", line 302, in imwrite
    writer = get_writer(uri, format, "i", **kwargs)
  File "/home/genevieve/anaconda3/envs/napari-dev/lib/python3.7/site-packages/imageio/core/functions.py", line 226, in get_writer
    "Could not find a format to write the specified file " "in mode %r" % mode
ValueError: Could not find a format to write the specified file in mode 'i'
```

sofroniewn: ah, that's really no good - and that issue you link to above really does seem unresolved. Agreed we shouldn't have something that breaks on linux. Out of curiosity what is the behavior like if you put in a "bad" file extension, say `jpg` in the context of the current PR, do you get a warning and a request to add `png` or does it do nothing?

I'm inclined to keep the current behavior and then add in your previous catch in case it fails, that should make the best of this not perfect situation. i can submit a fix soon

GenevieveBuckley: With jpeg I get an error say pillow won't support an alpha channel (I just opened a png with the camera data in it, didn't think that should have had an alpha channel to begin with).

GenevieveBuckley:
> If this looks good to @GenevieveBuckley then I think it is good to go.

So this is more visible than the conversation in-line with the code - we're getting some test failures (especially on Linux), so this needs to stay open until that is resolved.

napari/_qt/qt_viewer.py
sofroniewn: Ok, what that means is that the dialog option is completely ignored for you on linux, for me on mac it catches it earlier. The screenshot does have an alpha value as it allows us to capture opacity from vispy in what I think is the best way possible, but I might be wrong - if the alpha doesn't get used we could drop.

GenevieveBuckley: > Ok, what that means is that the dialog option is completely ignored for you on linux, for me on mac it catches it earlier. 

No, I don't think this is strictly accurate. The "filters" dialog option does get used on Linux - in the dialog box that pops up I see only folders and files that match the filter (so no text files, etc. are visible to me there). What it doesn't do is automatically use the first file extension in that filters list if the user types in a filename without any file extension.

sofroniewn: Thanks for the clarification!

sofroniewn:
Ok @GenevieveBuckley I've added the catch code back in and I've left the filter code in too - hopefully this will mean we now get to leverage the filter code where we can (and the additional dialogs on mac that it provides), but we also catch and fix if it fails. Does this seam reasonable to you? Hopefully the tests now pass too. Let me know if this is good :-)

GenevieveBuckley:
It looks reasonable to me now @sofroniewn, and the tests pass on my machine.

However the Windows build test produces this DeprecationWarning, is that something we need to deal with?
```
napari/_tests/test_viewer.py::test_screenshot
  c:\users\containeradministrator\anaconda\lib\site-packages\qtconsole\console_widget.py:490: DeprecationWarning: QFontMetrics.width(const QString & arg__1, int len) const is deprecated
    width = font_metrics.width(' ') * self.console_width + margin
```

sofroniewn:
I do believe that deprecation warning was present before this PR and relates to the console code. We should probably address at some point, but it can be in future PRs

I will let this sit till tomorrow and then merge. Thanks again @GenevieveBuckley

PR #946, opened 2020-02-04
Fixes broken tutorials links

HagaiHargil:
# Description
Fixes broken links in the README.

## Type of change

- [x] Bug-fix (non-breaking change which fixes an issue)

## Final checklist:
- [x] My PR is the minimum possible work for the desired functionality

PR #948, opened 2020-02-05
Update tutorial and installation links

manzt:
# Description
Fixes broken links in README.md

# References
I was trying to point a friend to napari and they were having trouble getting started

# How has this been tested?
I did my best to update the URLs to those on `napari.org`

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@
 
 We're developing **napari** in the open! But the project is in an **alpha** stage, and there will still likely be **breaking changes** with each release. You can follow progress on this repository, test out new versions as we release them, and contribute ideas and code.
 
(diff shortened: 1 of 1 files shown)


manzt:
Perhaps "in-depth tutorials" are distinct from "tutorials", but in that case there is no link to the "tutorials" directly from the GitHub README.md.

sofroniewn:
Hi @manzt thanks so much for jumping in to fix those broken links, it's much appreciated, but someone else submitted a PR on this yesterday #946, that I just merged. Apologies that we didn't get that in sooner.

As to "in-depth tutorials" we should actually probably drop "in-depth" as it maybe is only confusing at this point. I think we used to have a "mini" tutorials section in the read me, but that's long gone at this point. If you want to change this PR to drop the word "in-depth" that's fine with me, otherwise I will close it now that #946 has gone in

Issue #922, opened 2020-01-29
Shapes shape, range, and extent


kevinyamauchi:
## ❓ Questions and Help

Hello everyone. While working on #920, I realized that I don't really understand how we use extent, range, and shape in the context of the dims, but I may have found a bug. In short, the shape calculation for the `Shapes` layer seems off by one.

## To Reproduce

For example, if I am making a `Shapes` layer with the following data (1 square each on z planes 0, 1, and 2):
```python
data = [
    [
        [0, 100, 100],
        [0, 100, 200],
        [0, 200, 200],
        [0, 200, 100]
    ],
    [
        [1, 125, 125],
        [1, 125, 175],
        [1, 175, 175],
        [1, 175, 125]
    ],
    [
        [2, 100, 100],
        [2, 100, 200],
        [2, 200, 200],
        [2, 200, 100]
    ],
]
layer = Shapes(np.array(data), shape_type='polygon')
```
I get the following:

```python
>>> layer.shape
(2, 100, 100)

>>> layer._get_range()
((0.0, 2.0, 1), (100.0, 200.0, 1), (100.0, 200.0, 1))

>>> layer._get_extent()
((0.0, 2.0, 1), (100.0, 200.0, 1), (100.0, 200.0, 1))
```
Further, when instantiating a viewer with just shapes (`napari.view_shapes(data, shape_type='polygon`), the z dim slider only has positions z=0 and z=1.

## Expected behavior
I am not sure what to expect for the range and extent, but I would have expected:
```python
>>> layer.shape
(3, 101, 101)
```
since there are 3 z slices and the squares are 101 pixels wide in z slices 0 and 1 (i.e., same as how numpy array shape works). I took a look at the `Image` layer and it looks like the shape is based on the shape of the image numpy array, which is consistent with my expectation.

## Environment
napari: 0.2.10+19.gfeb879f
Platform: Darwin-16.7.0-x86_64-i386-64bit
Python: 3.6.8 (default, Oct 31 2019, 11:52:31) [GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.39.2)]
Qt: 5.13.1
PySide2: 5.13.1
VisPy: 0.6.4
NumPy: 1.17.3
SciPy: 1.3.1
scikit-image: 0.16.2
Dask: 2.6.0

GL version: 2.1 INTEL-10.25.24
MAX_TEXTURE_SIZE: 16384

kevinyamauchi:
[...]

PR #950, opened 2020-02-05
fix 4D ellipses

sofroniewn:
# Description
This PR fixes the bug reported in https://forum.image.sc/t/adding-shapes-to-4d-data-in-napari/33600 on adding 4D ellipses and adds a test that would have previously failed. The problem was due to an incorrect initialization line.

## Type of change

### DIFF
--- a/napari/layers/shapes/_tests/test_shapes.py
+++ b/napari/layers/shapes/_tests/test_shapes.py
@@ -153,6 +153,25 @@ def test_ellipses():
     assert np.all([s == 'ellipse' for s in layer.shape_type])
 
 
+def test_4D_ellispse():
+    """Test instantiating Shapes layer with 4D planar ellipse."""
+    # Test a single 4D ellipse
+    np.random.seed(0)
+    data = [
+        [
+            [3, 5, 108, 108],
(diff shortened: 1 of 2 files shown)


kevinyamauchi (approved):
Thanks, @sofroniewn ! This works for me. In addition to your test, I tried adding an ellipse to 4D data using the linked image.sc example and it works great! I also tried adding ellipses with a 4D bounding box to a bare viewer (`napari.view_shapes()`) and the extent of the dims  (see below for a screenshot and the code snippet), but I think this is due to the shape and extent calculation as described in #922 and can be fixed in a follow up PR once we decide on how shape and extent should be defined.

Screen Shot 2020-02-06 at 9 35 10 AM

```python
import napari

with napari.gui_qt():

    bbox = [[3, 5, 108, 108], [3, 5, 108, 148], [3, 5, 148, 148], [3, 5, 148, 108]]
    viewer = napari.view_shapes(bbox, shape_type='ellipse')
```

sofroniewn:
Thanks for the review @kevinyamauchi I will merge this now. Those shape / size issues can be addressed in dedicated PRs

PR #951, opened 2020-02-05
Points view data refactor

kevinyamauchi:
# Description
As discussed in #858, this PR moves the view slicing for the points data and display properties to the model. The following properties were added:

- `Points.view_data`:  the in slice point data (`Points._data_view`). `Points._data_view` is calculated in Points._set_view_slice() because it is used for multiple operations per refresh.
- `Points.view_size`: the in slice point sizes (`Points._size_view`). `Points._size_view` is calculated in Points._set_view_slice() because it is used for multiple operations per refresh.
- `Points.view_face_color`: the in slice face colors. The slicing is performed when needed.
- `Points.view_edge_color`: the in slice. The slicing is performed when needed.

## Type of change

# References
This is the second PR in #858.

# How has this been tested?
- [x] added view tests in `test_points.py`

## Final checklist:


sofroniewn (commented):
This is looking good @kevinyamauchi - do we also want to move the slicing logic inside `view_data` and `view_size` and drop `_data_view` and `_size_view` entirely, and have `_set_view_slice` just set `_indices_view` only? In the future we might want to rename `_set_view_slice` -> `_set_view_indices` (this can only happen when we've refactored all the layers). I might also rename `_indices_view` -> `_view_indices` (this could happen now).

The one downside I see is that the slicing would always happen twice - once when requested in vispy, and once when requested for the thumbnail generation. This might be ok / we might want to think of another way to fix this.

Also why is the PR still [WIP], is there something still missing?

kevinyamauchi:
> This is looking good @kevinyamauchi - do we also want to move the slicing logic inside view_data and view_size and drop _data_view and _size_view entirely, and have _set_view_slice just set _indices_view only? 

The main reason I didn't move the data and size slicing into the property is because, as you noted, `_data_view` and `_size_view` would have to be calculated multiple times per refresh. I think the performance hit is small, so we can make the move if we think it would be clearer/cleaner. Alternatively, I can add a comment in the docstring to make it clear where `Points._data_view` and `Points._size_view` are calculated. What do you think?

> In the future we might want to rename _set_view_slice -> _set_view_indices (this can only happen when we've refactored all the layers).

I agree that we should consider this once we have unified the behavior across all layers.

> I might also rename _indices_view -> _view_indices (this could happen now).

Sounds good. Will do!

> Also why is the PR still [WIP], is there something still missing?

I was still debating whether or not the data view and size views should be calculated on-demand or saved (see above). Other than that, I wasn't planning on making any other changes.

sofroniewn:
> The main reason I didn't move the data and size slicing into the property is because, as you noted, _data_view and _size_view would have to be calculated multiple times per refresh. I think the performance hit is small, so we can make the move if we think it would be clearer/cleaner. Alternatively, I can add a comment in the docstring to make it clear where Points._data_view and Points._size_view are calculated. What do you think?

Personally I like the idea of moving this stuff into the property, it will make the `_set_view_slice` simpler and I think the code base more readable / better organized for a minor hit in performance (I think right now a 2x on the slicing for update thumbnails). But it might also allow us to improve some things, such as only trigger the setting of `_view_indices` when the dimensions actually change instead of whenever we need to refresh the view (for example when we add points we don't need to recalculate `_view_indices`, but we do now).

It might also make things a little simpler for the `Labels` layer where again we'll be able to update the data when people are painting without having to update `_view_indices`.

What does @jni think?

kevinyamauchi:
@sofroniewn, I think you make some good points and I have moved the data and size slicing into the `Points.view_data*` and `Points.view_size` methods. I added some tests and all tests pass, so I think this is ready for the first round of review.

napari/_vispy/vispy_points_layer.py
```diff
@@ -83,10 +83,10 @@ def _on_data_change(self, event=None):
     def _on_highlight_change(self, event=None):
         if len(self.layer._highlight_index) > 0:
             # Color the hovered or selected points
-            data = self.layer._data_view[self.layer._highlight_index]
+            data = self.layer.view_data[self.layer._highlight_index]
             if data.ndim == 1:
                 data = np.expand_dims(data, axis=0)
-            size = self.layer._size_view[self.layer._highlight_index]
+            size = self.layer.view_size[self.layer._highlight_index]
```
jni: Why did this get made public?

napari/layers/points/points.py
```diff
@@ -1127,6 +1124,72 @@ def mode(self, mode):
...
+            sizes = (
+                self.size[
+                    np.ix_(self._indices_view, self.dims.displayed)
+                ].mean(axis=1)
+                * self._point_scale_view
+            )
+
+        else:
+            # if no points in this slice send dummy data
+            sizes = [0]
```
jni: Why not an empty list?

sofroniewn:
Thanks @kevinyamauchi - I think @jni might have had some reservations about the slicing inside the `view_*` properties, but I'm not sure if that has been discussed further. I think he (and I) also have a preference for these to all be private attributes `_view_*`

napari/_vispy/vispy_points_layer.py
```diff
@@ -50,22 +50,22 @@ def _on_data_change(self, event=None):
             self._on_display_change()
             self._on_highlight_change()
 
-        if len(self.layer._data_view) > 0:
-            edge_color = self.layer.edge_color[self.layer._indices_view]
-            face_color = self.layer.face_color[self.layer._indices_view]
+        if len(self.layer._indices_view) > 0:
+            edge_color = self.layer.view_edge_color
+            face_color = self.layer.view_face_color
```
jni: As @sofroniewn suspected, I do prefer these to be private methods/properties

napari/layers/points/points.py
```diff
@@ -1159,29 +1220,26 @@ def _slice_data(self, indices):
...
-                in_slice_data = self.data[np.ix_(matches, disp)]
                 size_match = sizes[matches]
                 size_match[size_match == 0] = 1
                 scale_per_dim = (size_match - distances[matches]) / size_match
                 scale_per_dim[size_match == 0] = 1
                 scale = np.prod(scale_per_dim, axis=1)
-                indices = np.where(matches)[0].astype(int)
-                return in_slice_data, indices, scale
+                slice_indices = np.where(matches)[0].astype(int)
+                return slice_indices, scale
```
jni, line 1237: Could we take this opportunity to implement this slicing with scipy.spatial.cKDTree? We've used this in a couple of recent PRs in skimage with 1-2 orders of magnitude speedup relative to linear search. See eg: https://gist.github.com/clementkng/cf0321e2b62daed21189867187e36751

jni (changes requested):
@kevinyamauchi as @sofroniewn suggest, I don't see why these views should be publicly visible. I also had a couple of questions about implementation, and a suggestion for speeding up slicing.

napari/layers/points/points.py
sofroniewn: I suspect it's because of some weird stuff inside the corresponding `vispy_points_layer.py` code https://github.com/napari/napari/blob/master/napari/_vispy/vispy_points_layer.py#L65 as vispy can't take an empty list there, but I'd instead make this return an empty list and then do that catch and conversion inside `vispy_points_layer.py`

napari/layers/points/points.py
```diff
@@ -305,12 +305,9 @@ def __init__(
...
         self._drag_start = None
 
-        # Nx2 array of points in the currently viewed slice
-        self._data_view = np.empty((0, 2))
-        # Sizes of points in the currently viewed slice
-        self._size_view = 0
-        # Full data indices of points located in the currently viewed slice
+        # initialize view data
         self._indices_view = []
+        self._point_scale_view = []
```
sofroniewn: can this be `_view_size_scale` as it is really a scalar on the size of the points right?

napari/layers/points/points.py
kevinyamauchi: @sofroniewn is right. The `Points._view_size` now returns an empty list when there are no points in view.

napari/layers/points/points.py
kevinyamauchi: That's a good point. Changed!

napari/_vispy/vispy_points_layer.py
kevinyamauchi: I had initially done it to make it consistent with the other view properties I had made public. However, I think you and @sofroniewn made a good point about making them private, so I've made them all private.

napari/layers/points/points.py
sofroniewn: @jni I feel like that should be it's own PR

sofroniewn: very happy to use that instead though!!

sofroniewn (approved):
Thanks for those changes @kevinyamauchi - this looks good to me, I'm approving and think we should merge. I do wonder though how the changes proposed in #1020 will relate to this - we might be able to drop these explicit `_view*` properties or at least change how the think about them. Feel free to weigh in on #1020 as well @kevinyamauchi. Also @shanaxel42 we can talk about how you could incorporate this `_view*` stuff into #1020 too as it will come up soon

kevinyamauchi:
Thanks, @sofroniewn! I agree that it could make sense to convert the in-slice calculation to a kD-tree (or other faster search) in a different PR. I think that it's not super straight forward to implement with when `n_dimensional == True` because of the anisotropy (i.e., just filtering by euclidian distance isn't sufficient). I suppose we may be able to use a hamming distance if we shift the coordinates by 0.5 and then round down. In any case, I think it warrants a longer

#1020 looks awesome! I will take a look soon.

jni (approved):

PR #954, opened 2020-02-06
flatten auto-generated docs repo

### LINKED ISSUE #953: docs/dev URL
can we change https://napari.org/docs/dev/ to https://napari.org/docs/ ?
I think the /dev/ made sense as long as it was a strictly auto-generated apidoc, but if we merge #942, then I can't really think of what else we'll need to put in `/docs/` that would necessitate an additional level for that URL.
or, in any case ... can we change the "docs" link at https://napari.org/ to go straight to the sphinxdocs?

kne42:
resolves #953 

tested by running the commands locally

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -251,9 +251,10 @@ push_docs_task:
     - cd doc_repo
     - git config --global user.email "cirrus@cirrus-ci.org"
     - git config --global user.name "Cirrus Bot"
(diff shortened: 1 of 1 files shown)


tlambert03 (approved):
👍

sofroniewn (approved):
Looks good to me!

PR #955, opened 2020-02-07
[WIP] adds edge_color_mode and face_color_mode as kwargs

### LINKED ISSUE #931: Should color mode be a kwarg arg to Points?
## ❓ Questions

Should color mode be a kwarg arg to Points (i.e. `edge_color_mode` and `face_color_mode`)? The reason for this is that these properties will affect how things get rendered on the screen and so if we want to enable full round trips of our data through serialization and back again to have the exact same thing appear on the screen they need to be included. Thoughts @kevinyamauchi?

neuromusic:
# Description

adds `edge_color_mode` and `face_color_mode` as kwargs as described in #931 . I basically just did what @kevinyamauchi said to do 😛 

## Type of change

# References

closes #931 

# How has this been tested?

~I didn't dig deeply, but it's not clear to me how this part of the codebase is tested. `edge_color_cycle` doesn't appear in the test suite at all, for example.~ nevermind -jk

## Final checklist:
- [ ] ~I have commented my code, particularly in hard-to-understand areas~ - n/a
- [ ] I have made corresponding changes to the documentation - TODO
- [ ] I have added tests that prove my fix is effective or that my feature works - TODO


neuromusic:
aha! found the tests for `Points`

kevinyamauchi:
Hey, @neuromusic ! Thanks for working on this.

One wrinkle that I didn't mention when mocking this out in the issue is reconciling the default color mode with the other kwargs. For example, in the color cycle tests that are failing, no *_color_mode argument is given, so the layer is instantiated with in the `direct` mode (default), but the face_color is given has a property on which to apply the color cycle. Thus, the edge/face color is not set properly.

I think we can add a bit of kwarg validation and raise the appropriate errors to address this. For example:

- if a face/edge color is given as a property and the color mode is `direct` raise an error. 
- if a face/edge color is given as a property and no color mode is specified, do not explicitly set color mode, as the color mode is guessed in the [`face_color](https://github.com/napari/napari/blob/3950dba6ced8a7f477877149df1d920f3502a647/napari/layers/points/points.py#L733)` and [`edge_color`](https://github.com/napari/napari/blob/3950dba6ced8a7f477877149df1d920f3502a647/napari/layers/points/points.py#L585) setter (i.e., the color mode is already set)

How does this sound?

goanpeca:
Is this PR stale @sofroniewn, or something that still needs to be worked on :-) ?

sofroniewn:
We've changed the API around this with the addition of the colormanger and probably want to take a different approach and avoid a kwarg explosion. I will close now, but let's leave the original issue open as it is somewhat unresolved

PR #957, opened 2020-02-07
Abstract the construction of view/viewermodel

ttung:
# Description
This ensures that we properly clean up QtViewers.

Some views were not being cleaned up properly, and thus exposed a bug where qtconsoles shared a kernel.  see the change on `napari/_qt/qt_console.py`

## Type of change

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/_qt/_tests/test_qt_viewer.py
+++ b/napari/_qt/_tests/test_qt_viewer.py
@@ -1,8 +1,6 @@
 import numpy as np
 import pytest
 
-from napari.components import ViewerModel
-from napari._qt.qt_viewer import QtViewer
 from napari._tests.utils import (
     add_layer_by_type,
     check_viewer_functioning,
@@ -10,11 +8,9 @@ from napari._tests.utils import (
 )
 
 
-def test_qt_viewer(qtbot):
+def test_qt_viewer(viewermodel_factory):
     """Test instantiating viewer."""
-    viewer = ViewerModel()
-    view = QtViewer(viewer)
-    qtbot.addWidget(view)
+    view, viewer = viewermodel_factory()
 
     assert viewer.title == 'napari'
     assert view.viewer == viewer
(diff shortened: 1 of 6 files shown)


tlambert03:
🙌 thanks!  will look soon

napari/_tests/infra/viewer.py
```diff
@@ -0,0 +1,24 @@
+from typing import List
+
+import pytest
+
+from napari._qt.qt_viewer import QtViewer
+from napari.components import ViewerModel
+
+
+@pytest.fixture(scope="function")
```
tlambert03, line 9: this is the default scope... there just to be explicit?

napari/_qt/_tests/test_qt_viewer.py
```diff
@@ -56,9 +46,7 @@ def test_new_labels(qtbot):
     assert np.sum(view.dims._displayed_sliders) == 0
 
     # Add labels with image already present
-    viewer = ViewerModel()
-    view = QtViewer(viewer)
-    qtbot.addWidget(view)
+    view, viewer = viewermodel_factory()
```
tlambert03, line 49: just curious here: how does the `viewermodel_factory` handle two instantiations in a single function scope?  and actually... what was previously happening?  Did that first `view` variable just get left in limbo after becoming unbound when we created a second viewer with the same variable name... and therefore never get shutdown?

tlambert03 (approved):
this looks great @ttung.  I love how it simplifies the code and makes sure we're cleaning up.
I think that previous test failure was unrelated... I just restarted it and we can merge this if it passes

tlambert03:
looks like everything is passing now.  Will wait a day to see if any other @napari/core-devs have thoughts, and then merge

napari/_qt/_tests/test_qt_viewer.py
ttung: It saves both entries and shuts both down.  We save a list of `QtViewer`s, not just one.

And yes, the previous code just clobbered and left the other `QtViewer` hanging.

kevinyamauchi: > And yes, the previous code just clobbered and left the other QtViewer hanging.

Thanks for fixing this!

tlambert03:
I'm going to merge this... I'd like to see if it fixes some of the other test issues we're having at the moment.  thanks @ttung

Issue #945, opened 2020-02-04
Add docstrings to all our Qt classes


jni:
## 📚 Documentation

I was looking through our Qt classes to try to provide some guidance in napari/napari-tutorials#52, but found it hard to understand exactly what each Qt class is doing, because almost none of them have class docstrings, and many methods have no docstring or a one-liner rather than a complete docstring. See e.g. https://github.com/napari/napari/blob/master/napari/_qt/qt_layerlist.py#L19

PR #958, opened 2020-02-07
Add docstrings to qt_layerlist.py

GenevieveBuckley:
# Description
@jni suggests in https://github.com/napari/napari/issues/945 we need better docstrings for the Qt classes:
> I was looking through our Qt classes to try to provide some guidance in [napari/tutorials#52](https://github.com/napari/tutorials/issues/52), but found it hard to understand exactly what each Qt class is doing, because almost none of them have class docstrings, and many methods have no docstring or a one-liner rather than a complete docstring. See e.g. https://github.com/napari/napari/blob/master/napari/_qt/qt_layerlist.py#L19

This PR chips away at that problem a bit.

Note: it's entirely possible I've made some mistakes trying to decipher thing. Let me know if that's happened, since those are likely to be the places we need the most clarification. We might need to be especially careful about docstrings for things with similar names in both the napari models and qt widgets (stuff like "layer" might be super ambiguous). 

## Type of change

- [x] documentation update

# References
Reference https://github.com/napari/napari/issues/945

# How has this been tested?

N/A, docstring only changes

## Final checklist:

### DIFF
--- a/napari/_qt/qt_layerlist.py
+++ b/napari/_qt/qt_layerlist.py
@@ -415,7 +415,7 @@ class QtLayerList(QScrollArea):
 
 
 class QtDivider(QFrame):
-    """QtDivider class
+    """Qt divider used to separate Qt widgets visually.
 
     Attributes
     ----------
@@ -448,7 +448,7 @@ class QtDivider(QFrame):
 
 
 class QtLayerWidget(QFrame):
-    """QtLayerWidget class
+    """Qt view for Layer model.
 
     Attributes
     ----------
--- a/napari/_qt/qt_layerlist.py
+++ b/napari/_qt/qt_layerlist.py
@@ -114,7 +114,7 @@ class QtLayerList(QScrollArea):
         """Reorder list of layer widgets.
 
         Loops through all widgets in list, sequentially removing them
-        and inserting them into the correct place in final list.
+        and inserting them into the correct place in the final list.
 
         Parameters
         ----------
--- a/napari/_qt/qt_layerlist.py
+++ b/napari/_qt/qt_layerlist.py
@@ -182,7 +182,8 @@ class QtLayerList(QScrollArea):
 
         Parameters
         ----------
-        layer : napari.components.LayerList
+        layer : napari.layers.Layer
+            An instance of a napari layer.
 
         """
         total = len(self.layers)
@@ -452,8 +453,8 @@ class QtLayerWidget(QFrame):
 
     Attributes
     ----------
-    layer : [type]
-        [description]
+    layer : napari.layers.Layer
+        An instance of a napari layer.
     layout : QVBoxLayout
         Layout of the widget.
     nameTextBox : QLineEdit
--- a/napari/_qt/qt_layerlist.py
+++ b/napari/_qt/qt_layerlist.py
@@ -17,6 +17,30 @@ import numpy as np
 
 
 class QtLayerList(QScrollArea):
+    """QtLayerList class
+
+    Parameters
+    ----------
+    layers : napari.components.LayerList
+        Layer widget list class instance.
+
+    Attributes
+    ----------
+    centers : list
+        List of layer widgets center coordinates.
+    drag_name : str
+        Name attribute of the dragged layer.
+    drag_start_position : Two element array
+        Mouse coordinates at the beginning of the mouse drag event.
+    dragTimer : QTimer
+        Timer for autoscrolling layers list up/down when dragging a layer
+        near the end of the displayed area.
+    layers : napari.components.LayerList
+        Layer widget list class instance.
+    vbox_layout : QVBoxLayout
+        Layout of the layer list widgets.
+    """
+
     def __init__(self, layers):
         super().__init__()
 
@@ -52,7 +76,13 @@ class QtLayerList(QScrollArea):
         self.drag_name = None
 
     def _add(self, event):
-        """Insert widget for layer `event.item` at index `event.index`."""
+        """Insert widget for layer `event.item` at index `event.index`.
+
+        Parameters
+        ----------
+        event : qtpy.QtCore.QEvent
+            Event from the Qt context.
+        """
         layer = event.item
         total = len(self.layers)
         index = 2 * (total - event.index) - 1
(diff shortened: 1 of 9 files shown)


napari/_qt/qt_layerlist.py
```diff
@@ -17,6 +17,30 @@
 
 
 class QtLayerList(QScrollArea):
+    """QtLayerList class
```
jni: This (and in other commits) should not be a repeat of the class name, but rather something descriptive. e.g. "Widget to hold and display a list of all the layers present in the current window." (that's probably too long, but that's the idea)

napari/_qt/qt_layerlist.py
```diff
@@ -17,6 +17,30 @@
...
+
+    Parameters
+    ----------
+    layers : napari.components.LayerList
+        Layer widget list class instance.
```
jni: "The layer list to track and display." There is no "widget" in the components.

napari/_qt/qt_layerlist.py
```diff
@@ -17,6 +17,30 @@
...
+        List of layer widgets center coordinates.
+    drag_name : str
+        Name attribute of the dragged layer.
+    drag_start_position : Two element array
+        Mouse coordinates at the beginning of the mouse drag event.
+    dragTimer : QTimer
+        Timer for autoscrolling layers list up/down when dragging a layer
+        near the end of the displayed area.
+    layers : napari.components.LayerList
+        Layer widget list class instance.
```
jni: as above.

napari/_qt/qt_layerlist.py
```diff
@@ -17,6 +17,30 @@
...
+    vbox_layout : QVBoxLayout
+        Layout of the layer list widgets.
```
jni: The layout instance in which the layouts appear.

napari/_qt/qt_layerlist.py
```diff
@@ -17,6 +17,30 @@
...
+    layers : napari.components.LayerList
+        Layer widget list class instance.
+
+    Attributes
+    ----------
+    centers : list
+        List of layer widgets center coordinates.
+    drag_name : str
+        Name attribute of the dragged layer.
+    drag_start_position : Two element array
```
jni: Not that this is a mistake in the PR, but are we sure we want this to be a public attribute? ditto for drag name and drag timer. The inconsistent naming is also a concern.

Doc PRs are so great for figuring out API snafus!

jni: Aside, to follow the nomenclature used in skimage, this would be "array, shape (2,)".

napari/_qt/qt_layerlist.py
```diff
@@ -75,9 +111,15 @@ def _remove(self, event):
         divider.deleteLater()
 
     def _reorder(self, event=None):
-        """Reorders list of layer widgets by looping through all
-        widgets in list sequentially removing them and inserting
-        them into the correct place in final list.
+        """Reorders list of layer widgets.
```
jni:
```suggestion
        """Reorder list of layer widgets.
```

napari/_qt/qt_layerlist.py
```diff
@@ -75,9 +111,15 @@ def _remove(self, event):
...
+
+        Loops through all widgets in list, sequentially removing them
+        and inserting them into the correct place in final list.
```
jni:
```suggestion
        and inserting them into the correct place in the final list.
```

napari/_qt/qt_layerlist.py
```diff
@@ -139,15 +193,37 @@ def _ensure_visible(self, layer):
         self.ensureWidgetVisible(widget)
 
     def keyPressEvent(self, event):
+        """Ignores key press event.
```
jni:
```suggestion
        """Ignore a key press event.
```

btw why does this exist, @sofroniewn?

napari/_qt/qt_layerlist.py
```diff
@@ -139,15 +193,37 @@ def _ensure_visible(self, layer):
...
+
+        Parameters
+        ----------
+        event : qtpy.QtCore.QEvent
+            Event from the Qt context.
+        """
         event.ignore()
 
     def keyReleaseEvent(self, event):
+        """Ignores key relase event.
```
jni:
```suggestion
        """Ignore key relase event.
```

napari/_qt/qt_layerlist.py
```diff
@@ -292,6 +415,16 @@ def dropEvent(self, event):
 
 
 class QtDivider(QFrame):
+    """QtDivider class
```
jni: As above: this should be a description of what the class is for, not just its name.

jni (commented):
@GenevieveBuckley thanks! This goes a long way towards making these classes more friendly to use to newcomers. I've left some comments about the class docstrings, but otherwise all my suggestions are really minor.

napari/_qt/qt_layerlist.py
GenevieveBuckley: I don't think any of `drag_start_position`, `dragTimer`, or `drag_name` should be public attributes. I'm happy to open an issue and followup PR if there is general agreement here.

GenevieveBuckley:
I think these test failures are unrelated to this PR.
 I tried merging master in, like Talley suggested today, but that hasn't fixed it entirely.

tlambert03:
> I think these test failures are unrelated to this PR.
> I tried merging master in, like Talley suggested today, but that hasn't fixed it entirely.

Sorry, was wrong about that... but @kevinyamauchi has now fixed this bug in #972, and **now** you should be able to merge in master :)

sofroniewn (approved):
Wow, this is amazing @GenevieveBuckley, so great to see!! I'm approving now and it looks good to me, but I will wait to merge, as I'm not sure what order you want this and #974 to go in

GenevieveBuckley:
> Wow, this is amazing @GenevieveBuckley, so great to see!! I'm approving now and it looks good to me, but I will wait to merge, as I'm not sure what order you want this and #974 to go in

I don't mind what order they happen in, whichever one is second will need to be updated in any case. Either way, no big deal.

sofroniewn:
> I don't mind what order they happen in

Ok great, I will let this sit till tomorrow then in case @jni wants to weigh in one more time, and if not I will merge this one then. You can then update #974 as needed. Thanks again.

jni (approved):

jni:
Lovely!

napari/_qt/qt_layerlist.py
GenevieveBuckley: Nick answers this [in another comment](https://github.com/jni/napari/pull/5#issuecomment-589027107), he says:

> The `event.ignore` in the qt_layerlist allows the event to pass through a parent widget to its child widget without doing anything, otherwise the parent catches the event and it doesn't get passed on.

Issue #900, opened 2020-01-23
Bind the viewer in the napari.gui_qt() context manager


HagaiHargil:
## ❓ Questions and Help

I was wondering why can't I do the following:

```python
with napari.gui_qt() as viewer:
   viewer.add_image(...)
```

It seems l like a nice QOL improvement, and a good default behavior for the `gui_qt` context manager. Also, I find that one of the most confusing things for new users is the duplication between `add_image` and `view_image` - they're never sure when to use which and why. I believe that if the context manager can return a viewer instance, all of the `view_X` methods will become mostly redundant since most use cases involving napari use this context manager for starting the Qt app.

jni:
> most use cases involving napari use this context manager for starting the Qt app.

This is actually not true. I use napari mostly in IPython with `%gui qt`, and in that case, it's much faster to use `napari.view_image(my_array)` to look at it quickly, than to have to first create a `Viewer` object.

Regarding creating the Viewer and yielding it... Well, then I think the context manager should change its name. But, we did have discussions about this at one point, I can't remember why we steered away from yielding a fully created viewer. @sofroniewn @kne42 ?

tlambert03:
I agree: `gui_qt()` does exactly what it looks like it should do: it creates a qt event loop for the gui (really _any_ gui... not just napari), and if we were going to yield a viewer it should be a different context manager with a new name (or a perhaps a `with_viewer=True` argument to `gui_qt`)

sofroniewn:
> Also, I find that one of the most confusing things for new users is the duplication between add_image and view_image - they're never sure when to use which and why. I believe that if the context manager can return a viewer instance, all of the view_X methods will become mostly redundant since most use cases involving napari use this context manager for starting the Qt app.

I agree with this, I still don't love it - but as noted above for people that want a one liner we need it. I'm def open to more suggestions though, and having some input from new comers can be really helpful in these discussions.

> if we were going to yield a viewer it should be a different context manager with a new name (or a perhaps a with_viewer=True argument to gui_qt)

I'm open to this, but am wary of providing too many different options to all do the same thing. We'd also have to make sure it was clear then how to manipulate all the viewer properties that are now keyword arguments to the `Viewer`

tlambert03:
> I'm open to this, but am wary of providing too many different options to all do the same thing. We'd also have to make sure it was clear then how to manipulate all the viewer properties that are now keyword arguments to the `Viewer`

[...]

PR #961, opened 2020-02-08
add show param to Viewer

tlambert03:
# Description
anyone mind if we add a `show` argument to the `Viewer` to pass to `Window`?  (to allow `napari.Viewer(show=False)`).  I actually just had a case where I wished it was there (wanted to add layers and stuff but didn't actually want the gui up yet).  I'm not adding a show method to `Viewer` in this PR, so to show later would be `viewer.window.show()`

## Type of change

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/viewer.py
+++ b/napari/viewer.py
@@ -14,20 +14,27 @@ class Viewer(ViewerModel):
 
     Parameters
     ----------
-    title : string
-        The title of the viewer window.
-    ndisplay : {2, 3}
-        Number of displayed dimensions.
-    order : tuple of int
+    title : string, optional
+        The title of the viewer window. by default 'napari'.
+    ndisplay : {2, 3}, optional
+        Number of displayed dimensions.  by default 2.
+    order : tuple of int, optional
         Order in which dimensions are displayed where the last two or last
         three dimensions correspond to row x column or plane x row x column if
         ndisplay is 2 or 3.
-    axis_labels : list of str
-        Dimension names.
+    axis_labels : list of str, optional
+        Dimension names. by default they are labeled with sequential numbers
+    show : bool, optional
+        Whether to show the viewer after instantiation. by default True.
     """
 
     def __init__(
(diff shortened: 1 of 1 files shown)


AhmetCanSolak:
I like and need this too @tlambert03 ! Thank you for bringing up. Just a question: Can we come up with something like `viewer.show()`? it might be helpful for future direction with possible multiple viewers/windows i believe.

tlambert03:
we could certainly add a method to Viewer that just called the `self.window.show()` method.  I don't see anything wrong with that (unless we want to be super protective of methods on Viewer)

jni:
> unless we want to be super protective of methods on Viewer

We do! But this one makes sense to me, and it matches expectations from existing packages in the ecosystem, eg Figure.show in matplotlib.

kevinyamauchi:
I like this too! As you mentioned, being able to set up the viewer before showing it is useful. I also agree that it would be nice to have a `Viewer.show()` method.

tlambert03:
it's all set to go if someone wants to approve it! :) (failing test is unrelated)

AhmetCanSolak (approved):
Thank you @tlambert03 !

sofroniewn:
This is great, as is the `viewer.show` command. My only question is having added `show` to `Viewer` should we also add `show` to all our `view_*` commands as they have all the other Viewer args? If so maybe we should make an issue for it (could be a good first issue for someone new)

tlambert03:
Fine by me... if this is the first PR that breaks that relationship between `view_*` and `Viewer` args, perhaps it would be nice to fix just on principle (though I do wonder just how long those signatures will eventually get!)  But not super high priority (IMO), and would be fine to sit as a good first issue...

sofroniewn:
> (though I do wonder just how long those signatures will eventually get!)

Yeah that is a concern of mine, we might need to rethink this in light of other discussions like in #900, but for now it is the only one that breaks consistency. Also fine if you want to quickly get it done @tlambert03

PR #964, opened 2020-02-10
fix docs version tag

kne42:
update the git tags so that the version tag on the generated docs is correct

Screen Shot 2020-02-10 at 2 25 40 PM

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -245,7 +245,6 @@ push_docs_task:
 
   push_script:
     # adapted from https://github.com/scikit-image/scikit-image/blob/master/tools/travis/deploy_docs.sh
-    - git fetch --tags
     - VERSION=`python -c "from versioneer import get_version; print(get_version())"`
     - git clone --quiet https://${GH_REF} doc_repo
     - cd doc_repo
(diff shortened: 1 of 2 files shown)


jni:
@kne42 this is great! Is there any way to get rid of the `.dirty` tag at the end?

kne42:
@jni that's there because that's the development version

jni:
@kne42 no, if I'm not mistaken, the +[hash] is there because it's development, .dirty is there because there are build files in addition to the committed stuff.

kne42:
@jni oooooh you're right, I got confused. will fix now...

kne42:
@jni turns out that's just a result of my local version being dirty. if you look at https://napari.org/docs, you'll see that it's never labeled dirty

jni:
@kne42 ah, awesome! Except right now it shows up as "untagged" for me...? :joy: at least the hash is very useful. =)

kne42:
@jni this PR fixes the untagged part :)

jni:
:man_facepalming: LOL Got it :joy:

jni (approved):

PR #968, opened 2020-02-12
[WIP] napari.Viewer() as a context manager

### LINKED ISSUE #900: Bind the viewer in the napari.gui_qt() context manager

HagaiHargil:
# Description

This PR adds new functionality to `napari.Viewer()` by making it a context manager which can start a Qt application on its own, allowing a user to write

```python
with napari.Viewer(ndisplay=3) as viewer:
    viewer.add_points([])
```

This is useful since it removes a small amount of boilerplate which comes along with `napari.gui_qt()`, and is also more ergonomic since the viewer is returned from `__enter__` (the `as viewer` clause) which makes the context manager a bit more Pythonic and less surprising.

When instantiating a Viewer without a context manager, an event loop will be created (if it doesn't already exist), but napari itself won't start. The user would have to use a new `viewer.show()` method which blocks and spawns the event loop.

No other changes were made - `napari.gui_qt()` still works with the new Viewer.

## Type of change
- [ X ] This change requires a documentation update

# References
Closes #900.

# How has this been tested?

I've yet to write tests for this new feature, as I'm still waiting for confirmation from the core devs. Testing shouldn't be too hard, I believe. Lots of work should be put into the docs, though, assuming this becomes the preferred instatiation method for scripted napari.

## Final checklist:

### DIFF
--- a/napari/viewer.py
+++ b/napari/viewer.py
@@ -1,7 +1,8 @@
 from os.path import dirname, join
+import sys
 
-from qtpy.QtGui import QIcon
-from qtpy.QtWidgets import QApplication
+from qtpy.QtGui import QIcon, QPixmap
+from qtpy.QtWidgets import QApplication, QSplashScreen
 
 from ._qt.qt_update_ui import QtUpdateUI
 from ._qt.qt_main_window import Window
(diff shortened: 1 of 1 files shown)


tlambert03:
Hi @HagaiHargil,  I hate to be a continual naysayer here... and I want to re-emphasize that I appreciate your user-focused goals here!  
I just want to be very protective of what is pretty much our _main_ api.... and I'm afraid I'm a strong 👎 on this implementation:

- I dislike that now if someone creates a new viewer with `v = napari.Viewer()` without a Qt event loop running, _nothing happens_.  And they have to know to call `viewer.show()`.
- secondly `viewer.show()`, as implemented here, really has nothing to do with "showing" anything (it's poorly named).  It simply starts the Qt event loop.  A viewer instance _should_ be able to exist without being shown (and will after #961)... and showing of that viewer should have nothing to do with starting or stopping the Qt event loop.  And this conflates that process (showing/not showing the viewer) with the Qt event loop... an entirely separate thing.
- this also has some side effects: every time someone creates a new `napari.Viewer()` in an ipython session (on mac at least) the terminal window loses focus.  And there will probably be other things like that, that I haven't discovered yet.  Qt stuff can be tricky... and the previous `gui_qt` strategy of explicitly starting an even loop made it all, well, much more explicit.

semantically: I think we're getting farther and farther away from what is actually happening under the hood, and it's getting dangerously obfuscated.  I understand the goal is to "just make stuff work" for users who don't care about the Qt event loop.  But this is hiding it way too deeply for my tastes.  I would much rather just educate users: "hey, there's this thing called Qt that runs the gui ... doesn't really matter exactly what it's doing (and you'll see it in other Qt programs that you may use), but you have to make sure it's started before creating a viewer".  That's exactly what the previous error message did.

HagaiHargil:
Hi @tlambert03,

I really appreciate your effort and dedication on this matter. I'm by no means locked on this idea, and while I think it's a good one, I also think it's not crucial at all. In fact, the only reason I went through with this PR is because you :+1:'d @kne42 comment in the issue, which signaled to me, obviously wrongly so, that you support the `viewer.show()` idea. I have no intention of pushing the matter further against your advice.

tlambert03:
oh no... I hope I haven't caused you to waste time under the impression that this is what I wanted!  Very sorry for the confusion.  If it was this https://github.com/napari/napari/issues/900#issuecomment-582088056 that you were referring to... what I intended to "thumb up" there was more her main sentiment that "this will not be so trivial" ... and less that last parenthetical aside in which she mentioned a possible show method.

HagaiHargil:
All is good, this PR took me exactly 10 minutes to make.

sofroniewn:
I think we should let this and #900 sit for a little as we havn't reached consensus yet on how to move forward. Thanks though for the potential implementation and good discussion all around, we'll keep thinking about it.

PR #976, opened 2020-02-16
Viewer view_* func signature parity

tlambert03:
# Description
as discussed https://github.com/napari/napari/pull/961#issuecomment-586725447
this PR adds the `Viewer(show=)` argument introduced in #961 to all of the `view_*` methods to keep their signatures consistent.

## Type of change

## Final checklist:

### DIFF
--- a/napari/view_layers.py
+++ b/napari/view_layers.py
@@ -116,7 +116,11 @@ def view_image(
         The newly-created viewer.
     """
     viewer = Viewer(
-        title=title, ndisplay=ndisplay, order=order, axis_labels=axis_labels
+        title=title,
+        ndisplay=ndisplay,
+        order=order,
+        axis_labels=axis_labels,
+        show=show,
     )
     viewer.add_image(
         data=data,
@@ -259,7 +263,11 @@ def view_points(
     http://api.vispy.org/en/latest/visuals.html#vispy.visuals.MarkersVisual
     """
     viewer = Viewer(
-        title=title, ndisplay=ndisplay, order=order, axis_labels=axis_labels
+        title=title,
+        ndisplay=ndisplay,
+        order=order,
+        axis_labels=axis_labels,
+        show=show,
     )
     viewer.add_points(
         data=data,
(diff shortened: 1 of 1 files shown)


sofroniewn:
Thanks @tlambert03. Do you think we should add a test for the API parity - it would test that `view_*` looks like `add_*` and `Viewer` combined? Could test function signature / doc strings? Could be a followup PR.

Also looks like you still need to pass `show` to the `Viewer` inside those functions (apologies if this is still WIP)

tlambert03:
ha!  no it was not WIP... I just missed it.  🤦‍♂️thank you!
Testing for parity would be nice (at least as long as that is our intention!) ... but it'd be great if someone else wanted to tackle that in a follow up

sofroniewn:
> Testing for parity would be nice (at least as long as that is our intention!) ... but it'd be great if someone else wanted to tackle that in a follow up

Maybe @kne42 can do that as she wrote our original API testing functions

tlambert03:
also... I've been playing with [wrapt](https://wrapt.readthedocs.io/en/latest/) recently for tinkering with function signatures... and I'm pretty we could do the whole `view_layers.py` file automatically, while maintaining docs and tab completion.  Do you want me to make a draft of that?

sofroniewn:
> also... I've been playing with wrapt recently for tinkering with function signatures... and I'm pretty we could do the whole view_layers.py file automatically, while maintaining docs and tab completion. Do you want me to make a draft of that?

I'm maybe a little conservative when it comes to adding a new dependency like that. If we felt we'd use it significantly else where in the code base, then maybe, but otherwise having that the extra `view` file doesn't seem like such a big deal once we have the tests described in #977 in.

On the other hand it might be nice, especially if we could remove the duplication between the `add_*` methods and the `Layer.__init__` API too (and then only need to maintain one docstring inside the actual Layer and drop all our parity tests). We've broken some of the strict parity there, adding / changing some keyword args (particularly for allowing adding multichannel images as separate layers with a single `add_*` call), but it still might be worth it.

I really defer to @kne42 and @jni on these matters though

sofroniewn (approved):
LGTM!

tlambert03:
> I'm maybe a little conservative when it comes to adding a new dependency like that.

I agree.  I can probably do it without wrapt ... I guess what I really meant was that "in using wrapt recently... I've become aware of some cool tricks that could be applied without wrapt".  lemme see what I can do.  It would guarantee that these things stay in agreement...

PR #978, opened 2020-02-16
auto generate view_* methods

### LINKED ISSUE #977: Test `view_*` API
## 🚀 Feature
As noted in #976 it might be nice to add a test that our `view_*` API looks like our `add_*` and `Viewer` APIs combined. This should be true for all layers. We should test both the function signature and doc strings. We can test that the keyword args etc are even in the same order.

tlambert03:
# Description
Following conversation in #976 about keeping all of our `add_*`,`view_*`, and `Viewer.__init__` methods in sync, here's a crazy little bit of code that will autogenerate all of our `view_*` methods, while retaining their current function signatures, introspectability, tab completion in IPython, and sphinx autodoc.  It combines the `Viewer.__init__` and `Viewer.add_*` method signatures and docstring parameters.  I haven't yet been able to find any practical differences between this and the current code.  I tried to comment inline with explanations.

Let me know what you think...

This would close #977

## Type of change

# How has this been tested?

## Final checklist:


sofroniewn (approved):
Wow @tlambert03 this is crazy! Any PR that removes > 600 lines of code for the same functionality seems pretty cool to me!! I am approving this, but I'll let @jni and @kne42 weigh in before merge.

Out of curiosity can you extend any of this approach to our `add_*` methods (in follow up PRs)? It might be trickier there now that we allow for some API divergence around the multichannel stuff in `add_image` and some of the `path` file loading code. Still curious what your take is?

tlambert03:
It's conceptually possible to do it for the `add_*` methods, but much harder in some cases due to divergence between the `Layer.__init__` method and the corresponding `add_*` method.  For instance, in the `add_image` method we accept `lists` of values for a number of parameters like colormap, contrast_limits, gamma, etc...  So we can't just copy the docstrings and signature from `Image.__init__`.  As such, I think it's probably either too brittle or error prone to autogenerate, and/or won't actually be that much more DRY.

sofroniewn:
> It's conceptually possible to do it for the add_* methods, but much harder in some cases due to divergence between the Layer.__init__ method and the corresponding add_* method.

Yup, makes sense. Useful to think about though, as it might be a nice goal to keep in mind for the future

jni:
> I am approving this, but I'll let @jni and @kne42 weigh in before merge.

I am less gung-ho about this one. It removes a lot of code, which is very fancy and clever, but perhaps too clever. I certainly don't understand what's happening on a quick glance and would like for more than one of us to do so before merging. I also would like to try it out on pycharm to verify that it doesn't adversely affect introspection there.

Not saying at all I don't want this merge, just that it's important to make the code readable for when things go wrong.

tlambert03:
Yeah I share @jni’s view here.  This one is kinda freaky.  I think I’m generally in a favor of it, since it will guarantee parity between those methods and (reduce human error) but it’s pretty unconventional, so let’s let everyone sit with it for a bit. 

In terms of readable code, I tried to comment pretty well, but can definitely be even more verbose if you’d like, or add more to the docstring notes

napari/view_layers.py
```diff
@@ -1,703 +1,106 @@
...
+
+    inner_sig = ", ".join([f"{p}={p}" for p in combo_sig.parameters])
+
+    # compile a new function with the combined signature that calls real_func
+    fakefunc = f"def func{combo_sig}:\n    return real_func({inner_sig})\n"
+    view_func_code = compile(fakefunc, "fakesource", "exec")
+
+    # evaluate the new function in a fake namespace and extract it
+    fakeglobals = {}
+    eval(view_func_code, {"real_func": real_func}, fakeglobals)
```
tlambert03: This line is really the meat of it.  Here's how it works:

1. we define `real_func`, which is the (easier to understand) function that will do the work of creating a new viewer and adding a layer to it.
2. we create a **string** (`fakefunc`) that represents how we _would_ have typed out the original `view_*` method. 
    - `{combo_sig}` is an [inspect.Signature](https://docs.python.org/3/library/inspect.html#inspect.Signature) object (whose string representation is, conveniently, exactly how we would have typed the original function).
    - the inner `real_func({inner_sig})` part is basically how we were typing the body of the `view_*` functions before, e.g.: `(data=data, name=name, scale=scale ...)`
3. we compile that string into `view_func_code`
4. finally, we actually evaluate the compiled code in a (safe) empty namespace, and provide a `locals()` dict that tells python that the function name `real_func` in the `fakefunc` string actually corresponds to the `real_func` that we defined on line 66.   (Note: evaluation at this step is essentially exactly what was previously happening when python hit each `def view_*` declaration when importing `view_layers.py`)
5. we grab our newly evaluated function from the empty namespace that we supplied... and go on to add our documentation below

kne42 (commented):
awesome PR @tlambert03!

while I am a huge fan of autogenerating duplicated functions, I am :-1: on this for the same reason why we copy/pasted the layer signatures to the add_* methods: certain IDEs (*cough\* PyCharm *cough*) opt to statically parse documentation instead of dynamically fetching it for auto-complete and documentation which means that these functions would not show up for their users

tlambert03:
> while I am a huge fan of autogenerating duplicated functions, I am 👎 on this for the same reason why we copy/pasted the layer signatures to the add_* methods: certain IDEs (*cough* PyCharm _cough_) opt to statically parse documentation instead of dynamically fetching it for auto-complete and documentation which means that these functions would not show up for their users

wow, that sucks!  the code autocompletion and intellisense work just fine with this code in vscode ... come on pycharm, that's quite frustrating.

kne42:
@tlambert03 we can check again and pray that PyCharm updated this, as the last time this issue arose was ~1 year ago but yeah it's rly dumb :/

kne42:
or maybe it has a plugin for jedi? lol

tlambert03:
I just downloaded pycharm and can confirm that it still doesn't work. 😡. lame.

tlambert03:
I’m going to close this, since pycharm usage is common enough that this would be unacceptable.  If either pycharm capabilities improve (or we care less about supporting it) in the future, we can revisit

tlambert03:
reopening in case we consider not worrying about pycharm

sofroniewn:
> reopening in case we consider not worrying about pycharm

I'm supportive of merging this. Maybe @jni can review

napari/view_layers.py
```diff
@@ -1,867 +1,127 @@
+"""Methods to create a new viewer instance and add a particular layer type.
+
+This module autogenerates a number of convenience functions, such as
+"view_image", or "view_surface", that both instantiate a new viewer instance,
+and add a new layer of a specific type to the viewer.  Each convenience
+function signature is essentially just a merged version of one of the
+``Viewer.add_<layer_type>`` methods, along with the signature of the
+``Viewer.__init__`` method.  The final generated functions follow this pattern
```
```suggestion
function signature is a merged version of one of the ``Viewer.__init__`` method
and the ``Viewer.add_<layer_type>`` methods.  The final generated functions
follow this pattern
```

napari/view_layers.py
```diff
@@ -1,867 +1,127 @@
...
+          object (whose string representation is, conveniently, exactly how we
+          would have typed the original function).
+        - the inner `real_func({inner_sig})` part is basically how we were
+          typing the body of the `view_*` functions before, e.g.:
+          `(data=data, name=name, scale=scale ...)`
+    3. we compile that string into `view_func_code`
+    4. finally, we actually evaluate the compiled code in a (safe) empty
+       namespace, and provide a `locals()` dict that tells python that the
+       function name `real_func` in the `fakefunc` string actually corresponds
+       to the `real_func` that we defined on line 66.   (Note: evaluation at
```
jni: I'd remove the line number, that is just *asking* to be made instantly obsolete.

napari/view_layers.py
```diff
@@ -1,867 +1,127 @@
...
+        raise AttributeError(f"No Viewer method named '{add_string}'")
+
+    # get signatures of the add_* method and Viewer.__init__
+    add_sig = inspect.signature(add_method)
+    viewer_sig = inspect.signature(Viewer)
+
+    # create a new combined signature
+    new_params = list(add_sig.parameters.values())[1:]  # [1:] to remove self
+    new_params += [
+        p.replace(kind=p.KEYWORD_ONLY) for p in viewer_sig.parameters.values()
```
jni: Should we just make these keyword-only to begin with anyway?

napari/view_layers.py
```diff
@@ -1,867 +1,127 @@
...
+    ]
+    combo_sig = add_sig.replace(parameters=new_params)
+
+    # define the actual function that will create a new Viewer and add a layer
+    def real_func(*args, **kwargs):
+        view_kwargs = {}
+        for key in list(kwargs.keys()):
+            if key in viewer_sig.parameters:
+                view_kwargs[key] = kwargs.pop(key)
```
```suggestion
view_kwargs = {
    k: kwargs.pop(k) for k in kwargs if k in viewer_sig.parameters
}
```

napari/view_layers.py
```diff
@@ -1,867 +1,127 @@
...
+    add_method_doc = NumpyDocString(add_method.__doc__)
+    viewer_doc = NumpyDocString(Viewer.__doc__)
+    n = 'n' if layer_string.startswith(tuple('aeiou')) else ''
+    new_doc = f"Create a viewer and add a{n} {layer_string} layer.\n\n"
+    new_doc += "\n".join(add_method_doc._str_param_list('Parameters'))
+    new_doc += "    " + "\n".join(viewer_doc._str_param_list('Parameters')[2:])
+    new_doc += (
+        "\nReturns\n-------\n"
+        "viewer : :class:`napari.Viewer`\n"
+        "    The newly-created viewer."
```
jni: Can you make this a separate function, `_combine_docstrings` for example?

jni (approved):
Great stuff! I'm super happy about this. We indeed chose not to do this because it confused PyCharm, but at this point I think that is PyCharm's problem rather than ours. Having the functions repeated everywhere has been frustrating when adding/changing keyword arguments, and this solves that problem, and is perfectly Pythonic.

FWIW the `add_` methods can also be autogenerated after this... =)

napari/view_layers.py
tlambert03: sure... add the star here?
```python
class Viewer:
    def __init__(
        self,
        *,
        title='napari',
        ndisplay=2,
        order=None,
        axis_labels=None,
        show=True,
    ):
```

napari/view_layers.py
tlambert03: hah.  _this_ broke my test :). `RuntimeError: dictionary changed size during iteration`. had to put the `kwargs` iterator in a list

napari/view_layers.py
jni: Yep!

napari/view_layers.py
```diff
@@ -20,15 +20,16 @@ def view_<layer_type>(*args, **kwargs):
...
+
 
 from numpydoc.docscrape import NumpyDocString
 
 from .components.add_layers_mixin import AddLayersMixin
 from .viewer import Viewer
 
 
-def _build_view_function(layer_string: str) -> Callable:
+def _build_view_function(layer_string: str, method_name: str = None):
```
jni: Rename to `_generate_view_function` or similar now, so that it's clear that this is doing more than just building?

napari/view_layers.py
tlambert03: done

sofroniewn:
Ok, this looks all good to go, any PR that removes 700 lines of code is very exciting! I'll let sit a little more in case @jni wants to look on more time when he wakes up, but then I'll merge!!

PR #979, opened 2020-02-17
Allow LayerList to be initialized by passing in a list of layers

GenevieveBuckley:
# Description

## 🚀 Feature

This PR alllows you to initialize a `LayerList` by passing in a list of layers. 

## Motivation
Currently the LayerList class in `napari/components/layerlist.py` doesn't allow you to initialize it by passing in a list of layers. It would be convenient to be able to do this, instead of having to append multiple times for very long lists. 

```python
from napari.components.layerlist import LayerList
from napari.layers.image import Image
from skimage import data
```

Currently you need to do this to create a layerlist with image layers in it:
```python
my_layerlist = LayerList()
my_layerlist.append(Image(data.camera()))
my_layerlist.append(Image(data.astronaut()))
```

But with this PR, we can make that a one-line call:
```python
my_layerlist = LayerList([Image(data.camera()), Image(data.astronaut())])
```

## Type of change

- [x] This change requires a documentation update. The docstring for LayerList has been updated.

# References

# How has this been tested?

- [x] added a new test checking you can initialize LayerLIst with a list of layers

## Final checklist:

### DIFF
--- a/napari/components/_tests/test_layers_list.py
+++ b/napari/components/_tests/test_layers_list.py
@@ -12,6 +12,13 @@ def test_empty_layers_list():
     assert len(layers) == 0
 
 
+def test_initialize_from_list():
+    layers = LayerList(
+        [Image(np.random.random((10, 10))), Image(np.random.random((10, 10)))]
+    )
+    assert len(layers) == 2
+
+
 def test_adding_layer():
     """
     Test adding a Layer
--- a/napari/components/layerlist.py
+++ b/napari/components/layerlist.py
@@ -24,9 +24,11 @@ class LayerList(ListModel):
             * reordered(): whenever the list is reordered
     """
 
-    def __init__(self):
+    def __init__(self, iterable=()):
         super().__init__(
-            basetype=Layer, lookup={str: lambda q, e: q == e.name}
+            basetype=Layer,
+            iterable=iterable,
+            lookup={str: lambda q, e: q == e.name},
         )
 
         self.events.added.connect(_add)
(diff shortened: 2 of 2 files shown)


sofroniewn:
This seems like a reasonable addition @GenevieveBuckley - but I am curious where you are encountering this in practice.

Are there places within the repo that we could use the new syntax to cutdown on code? Do you find yourself instantiating `Layers` and a `LayerList` like this when you are using napari? I only ever create a Viewer and use `add_*` methods so curious to learn more about your usage practices

GenevieveBuckley:
> This seems like a reasonable addition @GenevieveBuckley - but I am curious where you are encountering this in practice.
> 
> Are there places within the repo that we could use the new syntax to cutdown on code? Do you find yourself instantiating `Layers` and a `LayerList` like this when you are using napari? I only ever create a Viewer and use `add_*` methods so curious to learn more about your usage practices

I think this is only really useful to have while debugging or coding new features, I don't see this happening much in general use of napari from an end-user. 

I was debugging something else at the time (for https://github.com/napari/napari/pull/885, although the context here doesn't really matter).

sofroniewn:
> I think this is only really useful to have while debugging or coding new features, I don't see this happening much in general use of napari from an end-user.

Ok fair enough. Curious how @jni feels about this.

jni:
> Curious how @jni feels about this.

I suggested the PR! :joy: @GenevieveBuckley and I were working on TransformChains and we were trying out/debugging our use of the ListModel indexing magic that @kne42 made, and we were pretty surprised that LayerList could not be instantiated from a list of layers. Even if we see little use for it right now, I think it's worthwhile to avoid surprising behavior in our classes.

tlambert03:
I like!  Seems like a very logical thing to be able to do :)

sofroniewn (approved):
Great that everyone is supportive. I will approve now, but will wait if @kne42 wants to weigh in before merge. Thanks @GenevieveBuckley!

PR #980, opened 2020-02-17
Use another name for input arguments, instead of the python built-in "object"

GenevieveBuckley:
# Description
The `insert()` and `remove()` methods in `napari.utils._base.List` have an input argument named "object". This was a little confusing to me at first, since`object` is a python built-in class.

I thought it might be clearer if the input argument was renamed `item`, what do other people think?

## Type of change

("Bug-fix" is putting it too strongly since it's debatable if this is even an issue, but this a non-breaking change so it was the closest option that kinda fit.)

# References

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/utils/list/_base.py
+++ b/napari/utils/list/_base.py
@@ -141,11 +141,11 @@ class List(list):
         for e in iterable:
             self.append(e)
 
-    def insert(self, index, object):
-        super().insert(self.__locitem__(index), object)
+    def insert(self, index, item):
+        super().insert(self.__locitem__(index), item)
 
     def pop(self, index):
         return super().pop(self.__locitem__(index))
 
-    def remove(self, object):
-        self.pop(self.__locitem__(object))
+    def remove(self, item):
+        self.pop(self.__locitem__(item))


sofroniewn (approved):
This change makes sense to me, thanks! I will approve but wait to merge in case @kne42 wants to weigh in first

jni (approved):

kne42:
I chose to do this so that the function signature would match that of the built-in `list`. if that's not desired feel free to merge, but that was my justification at the time

sofroniewn:
> I chose to do this so that the function signature would match that of the built-in list.

@kne42 that definitely makes sense to me too. I don't have strong feelings on this. I lean towards merging, but for me it 51% vs 49%

GenevieveBuckley:
I don't feel strongly about it, this was more just on my list to raise as an FYI. Given the discussion I'm inclined to close this unless @jni wants to weigh in with a different opinion.

jni:
Eh, I think there's cases where venerable libraries are stuck with a poor early choice, but we can do better these days. e.g. "input" as the argument name for images in `scipy.ndimage`. They've acknowledged that it was a poor choice, but decided that the pain of changing that name outweighs the benefits of "getting it right". I doubt that the standard library would choose `object` today if they could start from scratch. So I'm still +1.

sofroniewn:
Ok, seems like people are either pretty neutral or slightly +1 so let's go with it so I'll merge

PR #981, opened 2020-02-17
Support for magicgui

tlambert03:
# Description
This PR adds support for [magicgui](https://github.com/tlambert03/magicgui) (but does not require it or add a dependency).  If `magicgui` is importable, then it calls `register_type` to register a callback function (`get_layers`).  That function tells `magicgui` to make a `QComboBox` with current layers of a given type whenever the user calls `@magicgui` on a function that has an argument annotated with class/subclass`napari.layer.Layers`

```python
@magicgui
def get_layer_mean(layer: napari.layers.Image) -> float:
    return layer.data.mean()

# this gui will have one widget, a QComboBox with layers in the Viewer
gui = get_layer_mean.Gui()
# we want to keep the combobox in sync in case the parent Viewer changes.
gui.parentChanged.connect(gui.refresh_choices)
```

If you're generally ok with this addition, feel free to comment on where the `_magicgui.py` file should actually live and be imported.

## Type of change

# References
https://magicgui.readthedocs.io/en/latest/

# How has this been tested?

I have not added any tests, but could do so if we want to add `magicgui` to our test dependencies.

## Final checklist:


sofroniewn:
Just to understand this, if I did
```python
@magicgui
def get_layer_mean_sum(layer_1: napari.layers.Image, layer_2: napari.layers.Image) -> float:
    return layer_1.data.mean() + layer_2.data.mean()
```
would I get two combo boxes with the names of all the `Image` layers?

```python
@magicgui
def get_layer_opcaity(layer: napari.layers.Layer) -> float:
    return layer.opacity
```
would I get one combo box with the names of all the layers regardless of type?

I can see this as very convenient reusable functionality that we want many plugin developers to have access too, curious though for right now how hard is it to do this without having this PR merged? Is it possible but just so common we want to put convenience features in the main repo to make it easier, or is it really quite annoying without having this code in here? (Maybe this is a dumb question as I guess you could just run the code from this PR ... but just trying to understand a little more as I havn't dived into magicgui yet).

tlambert03:
yep, you have it exactly right in your examples.  it's similar to [this example](https://magicgui.readthedocs.io/en/latest/examples/napari_img_math/) in the docs (though note I changed the callback API and need to update the docs)

If someone wanted to do it without having this merged, they would just have to include the code in this PR manually in their code.  stripped to the bare minimum without comments, that looks like:

```python
from magicgui import register_type
from napari.layers import Layer

def get_layers(magicgui, layer_type):
    try:
        viewer = magicgui.parent().qt_viewer.viewer
        return tuple(l for l in viewer.layers if isinstance(l, layer_type))
    except AttributeError:
        return ()

register_type(Layer, choices=get_layers)
```

so this is just a convenience, not merging it doesn't prevent anyone from doing it on their own.

jni:
Gotta say I love this though! We definitely want to have this in at some point, the question is only whether now is that point.

Another note: @tlambert03 not sure if you ever saw all the discussion in #263 or whether that even captures the discussion, but the idea would be to also allow the *output* of a function to define behaviour, so that if the output of function was defined as `-> napari.layers.Image:`, it would add that layer upon execution. (Of course, then, things like parameter sweeps need to be carefully implemented so that they don't add a layer on every execution!) ie, with proper annotation, we could skip the `show_result` definition in your example. Is that something you've been considering?

tlambert03:
> also allow the _output_ of a function to define behaviour, so that if the output of function was defined as `-> napari.layers.Image:`, it would add that layer upon execution.

yes definitely something we should do!  Would be quite easy.  We could either add another argument `return_callback` to the existing `register_type`, or create a new `register_return` function.  I created a new issue napari/magicgui#3

sofroniewn (approved):
I'm fine merging this now and having it in `0.2.12`, though I'd like the file to be inside the `_qt` folder as I think it is fundamentally about Qt things. How does that sound @tlambert03, if you agree can you move the file over?

I'll still wait for @jni to merge

tlambert03:
sure that's fine with me.  I will update this as well to include a return callback as well for https://github.com/napari/magicgui/pull/4

tlambert03:
actually... it's slightly tricky for this to live in `_qt` due to circular imports.  And it doesn't actually interact with Qt stuff at all:  it needs the `layers` module...  Where this _really_ belongs is in some sort of `napari.typings` module... which is a conversation for another day.  It doesn't quite feel right to put it inside of `napari.layers`  or `napari.layers.utils`... i'm at a bit of a loss, thoughts?

sofroniewn:
Maybe `napari.utils`? `napari.layers.utils` I think is also fine. It can easily be moved in the future without changing any external APIs right?

tlambert03:
k, moved to `napari.utils`.  should be all set if tests pass.

napari/utils/_magicgui.py
```diff
@@ -0,0 +1,88 @@
...
+            adder = getattr(viewer, f'add_{return_type.__name__.lower()}')
+            adder(data=result, name=result_name)
+
+    # tell magicgui that whenever an argument is annotated as a
+    # napari.layers.Layer, then it should be rendered as a ComboBox,
+    # where the items are the current Viewer.layers of a certain type.
+    register_type(Layer, choices=get_layers, return_callback=show_result)
+
+except ImportError:
+    pass
```
jni: I generally dislike modules that affect global state on import. I also don't love that this whole module is wrapped in a try/except.

My suggestions would be:

- take the above function definitions out of the try/except
- use a try/except on the import, where `register_type` is a do-nothing function if the magicgui module isn't available.
- in the `napari/__init__.py`, after importing, run the function, rather than running it in this module.

I'm still a bit cautious about running the function automagically on importing napari, even, but at least we give people digging in a chance to quickly find out where this is happening.

jni (changes requested):
I've made a major suggestion for refactor... I hope you agree with it, @tlambert03. It's up for discussion, but I'd like to at least have the discussion before merge. =)

napari/utils/_magicgui.py
tlambert03: Thank @jni, these are great suggestions, and I incorporated them all.  I understand the concern about running the function on napari import... if it makes you feel better, all the `register_type` function is doing in `magicgui` is populating a `dict` in magicgui.  
for what it's worth, with magicgui installed in my environment adding this registration to `napari.__init__` did not add a detectable increase to napari import time using `time python -c "import napari"` (i.e. any increase was less than the standard deviation of import times)

jni: Thanks @tlambert03! My concern is not about import time, it's just that when I am debugging some behaviour, it never occurs to me that the bug might be coming from an *import* rather than a function running. So e.g. I typically will step right over the import lines and then be like "WTF HOW DID THIS DICTIONARY GET POPULATED", upend the table, etc. ;) Anyway, I think (before having run through the code again) that this will be in a satisfactory place at this point. =)

napari/__init__.py
```diff
@@ -48,7 +48,10 @@
 )
 from ._qt import gui_qt
 from ._version import get_versions
-from .utils import sys_info
+from .utils import sys_info, _magicgui
+
+# register napari object types with magicgui
```
jni:
```suggestion
# register napari object types with magicgui if it is installed
```

napari/utils/_magicgui.py
```diff
@@ -0,0 +1,136 @@
...
+    if result is None:
+        return
+
+    try:
+        viewer = gui.parent().qt_viewer.viewer
+    except AttributeError:
+        return
+
+    # if they have annotated the return type as layers.Layer,
+    # the function MUST return a list of LayerData tuples
```
jni: (a) can you define what a LayerData tuple is in this comment?

(b) can you clarify "if they have annotated the return type as a base layer (layers.Layer), NOT a subclass of it, then ...". I originally read this as including subclasses and was highly confused. =)

jni (approved):
@tlambert03 I've made two very minor suggestions. Thank you!

napari/utils/_magicgui.py
tlambert03: hah, such a temper!  well, before we merge this, I'll just throw out three more possibilities:
1. no automagic behavior: we just create the `_magicgui.register_types_with_magicgui` function but don't call it, and anyone who wants to use `napari` with `magicgui` will just have to import the function from napari and call it themselves.
2. we put the `try: import napari` behavior in magicgui instead (I don't love that, since ultimately it'd be nice for other packages to support their own types, but it's definitely an option).
3. we stick with the current plan, but if anyone opens an issue because something unexpected happened and running this function was the culprit, we banish it!

jni: Yeah, option 3 is what I'd settled for, and happy to proceed with it. =)

PR #982, opened 2020-02-17
Add layers.events.changed event

tlambert03:
# Description
This PR adds a `changed` event to the List model.  

```python
viewer.layers.events.added.connect(callback)
viewer.layers.events.removed.connect(callback)
viewer.layers.events.reordered.connect(callback)

# becomes
viewer.layers.events.changed.connect(callback)
```

I found myself doing the former a number of times, and also found it here in `master`:
https://github.com/napari/napari/blob/b25ba4828ceef903d7532fb9568f8e0908be1a30/napari/components/viewer_model.py#L89-L96

## Type of change

- [x] Refactor

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/components/viewer_model.py
+++ b/napari/components/viewer_model.py
@@ -86,14 +86,9 @@ class ViewerModel(AddLayersMixin, KeymapMixin):
         self.dims.events.ndisplay.connect(self._update_layers)
         self.dims.events.order.connect(self._update_layers)
         self.dims.events.axis.connect(self._update_layers)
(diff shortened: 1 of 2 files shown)


sofroniewn (approved):
This looks good to me, do we have any concern though about weird behavior if people are expecting the order of the `add` and `changed` events to happen first / second. In general with the event system people aren't supposed to be making assumptions about the order of callbacks, but I wanted to raise this point before merge.

I'll also let @kne42 weigh in before merge too as she wrote a lot of the list model code

PR #985, opened 2020-02-18
Update computation of low discrepancy images

### LINKED ISSUE #983: Black is still a color in the label colormap => Some labels are not displayed
## 🐛 Bug

When displaying label images with a label of value 305, that label is not displayed in the viewer. When colors are shuffled, it appears (see images below). It also is still visible in the thumbnail. It looks like 305 is black in the default colormap. Black with translucency is not displayed. The same happens for 915, 1525, 1902 etc. in the standard colormap

## To Reproduce

1. Create a label image with a label of value 305
2. Display this image in napari by using `add_labels`

When originally displayed:
OriginalLabelMap
After shuffle colors:
PostShuffleColors

Here's a simpler case with an image generated in code:
```lbl = np.zeros((20, 20))
lbl[:, :10] = 304
lbl[:, 10:20] = 305
viewer = napari.Viewer()
viewer.add_labels(lbl)
```

Notice how the thumbnail still shows both labels, but the 305 label is not visible in the viewer until the colormap is shuffled.
After a color shuffle:

## Expected behavior
All labels should be visible when used in the default mode, with translucent blending. Therefore, black should be removed from the label colormap except for value 0. Currently, they still occur at some pattern all over the colormap, relatively sparsely.

## Environment

napari: 0.2.11
Platform: Darwin-19.3.0-x86_64-i386-64bit
Python: 3.7.4 (default, Aug 13 2019, 15:17:50) [Clang 4.0.1 (tags/RELEASE_401/final)]
Qt: 5.13.2
PySide2: 5.13.2
VisPy: 0.6.4
NumPy: 1.17.4
SciPy: 1.3.3
scikit-image: 0.16.2
Dask: 2.9.0

GL version: 2.1 INTEL-14.4.23
MAX_TEXTURE_SIZE: 16384

We reproduced this on Mac & Windows, in napari 0.2.6, 0.2.10 & 0.2.11

(cc @HessMax who also helped with narrowing down this bug)

jni:
This new computation avoids clipping (which maps too many values
to the same value), and maps the values to an appropriate margin
away from 0 and 1.

Fixes #983

### DIFF
--- a/napari/_qt/qt_viewer.py
+++ b/napari/_qt/qt_viewer.py
@@ -315,13 +315,7 @@ class QtViewer(QSplitter):
             # double check that an appropriate extension has been added as the filter
             # filter option does not always add an extension on linux and windows
             # see https://bugreports.qt.io/browse/QTBUG-27186
-            image_extensions = (
-                '.bmp',
-                '.gif',
-                '.png',
-                '.tif',
-                '.tiff',
-            )
+            image_extensions = ('.bmp', '.gif', '.png', '.tif', '.tiff')
             if not filename.endswith(image_extensions):
                 filename = filename + '.png'
             self.screenshot(path=filename)
--- a/napari/utils/colormaps/colormaps.py
+++ b/napari/utils/colormaps/colormaps.py
@@ -98,8 +98,8 @@ def _low_discrepancy_image(image, seed=0.5, margin=1 / 256):
         The set of ``labels`` remapped to [0, 1] quasirandomly.
 
     """
-    phi_inv = 1 / 1.6180339887498948482
-    image_float = seed + image * phi_inv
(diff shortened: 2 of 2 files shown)


jni:
I've confirmed that this avoids getting black labels for the image `np.array([[305, 915, 1525, 1902]])`, but I'll note that this image does expose the problem raised by @perlman last year that changing the seed does not disperse values that originally mapped close to each other. The image often appears as a single label, even after shuffling. It takes many shuffles to separate the labels.

This will need to be fixed in #713, but in the meantime this solves an existing issue.

sofroniewn (approved):
I have confirmed this fixes the reported bug. Thanks @jni, I will let sit for a bit and then merge

Issue #299, opened 2019-05-27
Draggable and resizable image layers


sofroniewn:
## 🚀 Feature
It would be nice to be able to drag and resize image layers using their `translate`, `scale`, and potentially `rotate` options. This would allow people doing image registration or template matching to attempt their best guess manually. Image border functionality #298 and the interactive box like functionality from the shapes layer could be reused to make this possible

sofroniewn:
One further thought - should this functionality actually be on the base layer and so accessible to all layers?

jni:
Yes, I think it makes little sense to make it part of the image layer only. This way you could select e.g. an image and its markers layer together and transform them together.

sofroniewn:
Ok making it a generic property of the base layer sounds good, as it makes sense for all our different layer types. Note that it is different then selecting all the markers and moving them - as that  would affect their data coordinates, whereas this would only be affecting the layer translate and scale properties (i.e. the scene graph stuff). It's possible to apply these transformation stuff to our `svg` generation too, so that it remains consistent.

I'm not sure though about allowing multiple layers to be transformed at the same time. Right now if you have multiple layers selected the controls panel on the left goes away and I imagine this being an option in that panel. @jni do you feel strongly that simultaneously manipulating multiple layers needs to be supported?

jni:
> Note that it is different then selecting all the markers and moving them - as that would affect their data coordinates, whereas this would only be affecting the layer translate and scale properties (i.e. the scene graph stuff).

Yes, I understand this. Incidentally it would be nice to be able to get at this data programmatically, ie, load two images, move one or the other, then grab the transform data. Ditto for *setting* the transform — instant visualization of registrations!

> do you feel strongly that simultaneously manipulating multiple layers needs to be supported?

Well, if it is a property of the base-layer, then maybe this interaction can live on the layers/layerslist directly? I am only going by intuition, but I think it might be very useful to select multiple layers and drag them together. One might annotate an image, then add a new image and want to translate the image and annotations for consistency. If you have to translate them separately, you lose the original relationship.

An intermediate option would be the ability to apply the transformation of one layer to another.

sofroniewn:
[...]

Issue #970, opened 2020-02-12
Nesting image layers in groups for navigation


NHPatterson:
## 🚀 Feature
Following discussion [here](https://forum.image.sc/t/image-array-layer-grouping/33908). It could be beneficial to group image planes from the same image. For instance, group multi-channel images together and allow a collapsible view of the channels. Right now, the system assigns a numerical index to the name of a multi-channel image so that ```My MC image``` becomes ```My MC image(1)```, ```My MC image(2)```, etc. in the side bar. It would also be beneficial here to have naming in the arguments of a nested layers, so that one could have like below where both the modality and the channels can be optionally named. 
```
Modality 1
   - DAPI
   - Cool antibody 1
   - Cool antibody 2

Modality 2
   - DAPI
   - Cool antibody 3
   - Cool antibody 4

```

## Motivation
Improve navigation in large datasets with many image planes. One can imagine a scenario where there is a multi-channel image composed of 10s or 100s of images, perhaps different processing of the original plane. Loading them all as single entries in the sidebar without individual naming could be very cumbersome.

## Alternatives
Not sure.

## Additional context
Here is an image of what that may look like, out of context. Of course, maybe hierarchy can be limited to only so many depths. 
![image](https://user-images.githubusercontent.com/17855764/74354223-a5bcfc00-4d80-11ea-96bf-33fc306f1141.png)

kevinyamauchi:
I like this idea!

If layers are nested, should there be some layer properties that are synced between the layers? For example, it seems like it should be possible to show/hide all of the layers in the group with a single click/command. I could also see it being useful to set opacity for all layers in a group together as well. Or maybe it would be better if the `LayerGroup` (or whatever we call it) can set the visibility/opacity for all child layers, but the child layers cannot set the layer properties of their "siblings". What do you all think?

In terms of API, I think it could make sense to add a `group` kwarg to the `Viewer.add_*` methods. If a `group` name is provided, the layer is added to the group if it exists and if the group does not exist, the group gets created and the layer gets added to it. Thoughts?

GenevieveBuckley:
I think grouped layers will be very useful. I think it's also important to be able to support groups of groups as well (so people can have more complicated hierarchies if that's what they need as well).

sofroniewn:
[...]

PR #989, opened 2020-02-21
[WIP] Allow translate/scale/rotation alignment of images

jojoelfe:
# Description
This is a very simple implementation of image translation/scaling/rotation to allow for registration of images. To do this the STTransform is replaced with a ChainedTransform containing the origin STTransform to avoid breakin existing code and adding a MatrixTransform. 

The UI introduces a "Transforming" mode to the image layer that first allows for translation of the active image and after a fixed point is placed with a shift-click for rotation and scaling. I find this is a efficient way of aligning images from different modalities as it requires only knowledge of two common points

Below is a video demonstrating the UI:

astronautalignment

I am not sure if you would want to pull this in its current form, but I though it might be interesting as a prototype.  I understand that there a underlying changes coming that will allow for a cleaner implementation. I think the biggest issue are:

- The main logic is in the image layer, while it probably should be in the base layer. However, I could not figure out how to introduce modes/UI modifications in the base layer.

-  It only works for 2D data. I don't see a big reason why this should not easily scale to 3D, but I haven't thought about it much.

- It would be neat to add a symbol, like a pin, to the fixed point but I am not sure how to do this in an image layer.

## Type of change

# References
This is related to #763 and #299 

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/_qt/layers/qt_image_layer.py
+++ b/napari/_qt/layers/qt_image_layer.py
@@ -1,12 +1,21 @@
 from qtpy.QtCore import Qt
-from qtpy.QtWidgets import QLabel, QComboBox, QSlider
+from qtpy.QtWidgets import (
+    QLabel,
+    QComboBox,
+    QSlider,
+    QHBoxLayout,
+    QButtonGroup,
+)
 from .qt_image_base_layer import QtBaseImageControls
-from ...layers.image._constants import Interpolation, Rendering
+from ...layers.image._constants import Interpolation, Rendering, Mode
+from ..qt_mode_buttons import QtModeRadioButton
 
 
 class QtImageControls(QtBaseImageControls):
     def __init__(self, layer):
         super().__init__(layer)
+
+        self.layer.events.mode.connect(self.set_mode)
 
         self.layer.events.interpolation.connect(self._on_interpolation_change)
         self.layer.events.rendering.connect(self._on_rendering_change)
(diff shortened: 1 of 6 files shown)


sofroniewn:
Wow @jojoelfe this is so cool!! I love it, and its great to see you dive in with the PR. I think this will be functionality that a lot of people find very useful and that we definitely want to support.

> I am not sure if you would want to pull this in its current form, but I though it might be interesting as a prototype. I understand that there a underlying changes coming that will allow for a cleaner implementation. 

As you note there is a PR coming soon #885 that should make this implementation easier (though I think @jni and @GenevieveBuckley were going to hold off on rotation / general affine transforms for now, though maybe we will want to think about that a bit more now in light of this PR. We could still get #885 in first with just translate and scale and then make a new PR for rotation/affine right after).

I think I'd recommend we hold off on further work on this PR until those PRs go in, because then as you say

> The main logic is in the image layer, while it probably should be in the base layer. However, I could not figure out how to introduce modes/UI modifications in the base layer.

this should become much easier.

> It would be neat to add a symbol, like a pin, to the fixed point but I am not sure how to do this in an image layer.

I'd actually like us to reuse the same bounding box / highlight functionality that we have in the shapes layer for rotating / resizing boxes. As this resizing will now be possible for every layer the highlighting code will move out of the shapes layer and onto the base layer. There is some discussion of this in #772. This could come in a separate PR though after this PR is merged.

> It only works for 2D data. I don't see a big reason why this should not easily scale to 3D, but I haven't thought about it much.

I think once we have #885 in which, will allow for nD transforms / coordinate systems it will become easier to think about how to extend this to 3D.

Does this approach make sense to you?

jojoelfe:
Yes, waiting for #885 and #772 would make things a lot cleaner.

I think together with #851 and #970 this will be a super powerful way to keep multi-modality imaging data organized.

I would love to help in any way so let me know if there is anything I can do.

sofroniewn:
> I would love to help in any way so let me know if there is anything I can do.

I wonder if we should make an `InteractionBox` object that lives in `napari/layers/interaction_box.py` and has a simple model consisting of the edge / corner / center / rotation handles vertices. This object would also have the mouse functions for resizing. We could pull a lot of this code out of the shapes layer, which would hopefully make that easier to maintain. We could then have a corresponding visual that showed to lines and vertices of the highlight in the highlight color. We could then have one of the objects be attached to the base layer and it could be used to either resize selected objects like in the shapes layer right now or when resizing whole layers.

How does this sound @shanaxel42 @jni @kne42 @jojoelfe?

If the others agree this is a good approach do you want to take a stab at this @jojoelfe in a new PR which would only remove the interaction box from shapes, put it on the base layer while preserving the functionality currently in shapes (and hopefully reducing some of the complexity in shapes) while preparing to support the functionality in this PR in followup PRs. I can provide more guidance / detail if needed.

Note this is the interaction box I am referring too which right now lives inside shapes but really is more generally useful:

interaction_box

jni:
@sofroniewn 

- I think rotations are pretty far off, unfortunately. The main issue is that we are n-dimensional, and if you rotate in the 2D plane then roll the axes, now you are doing arbitrary reslicing and interpolation. We don't even have a concept of that right now anywhere in the code base. It would be feasible to restrict rotations to 2D planes and e.g. reset/ignore them when rolling the axes, but that seems icky, to say the least. I'm not sure it's desirable.

- other than that, I do think the interaction box should be reused for this kind of functionality, that makes perfect sense.

jojoelfe:
@jni 
Would it be possible to allow rotations after slicing? For example the base layer could expose a slice_transform property, that allows a 2D or 3D slice to be transformed before painting on the canvas. I can't really judge how confusing this would be if this behaves different than scale/translate and whether it is to icky in the long-term

@sofroniewn 
I will give the InteractionBox object a shot in a separate pull request.

sofroniewn:
> I will give the InteractionBox object a shot in a separate pull request.

Great, thanks for claiming that!! I want to give you a heads up that we're planning to refactor some of the mouse function code - see #694, which I think can wait till after you're done, but I wanted to put it on your radar. Also don't hesitate to start a WIP PR and reach out early if you hit stumbling blocks - maybe @shanaxel42 and I jump into help

> Would it be possible to allow rotations after slicing?

It's a possibility. I want to give all this some serious thought, which personally I havn't done yet.

GenevieveBuckley:
> I think rotations are pretty far off, unfortunately. The main issue is that we are n-dimensional, and if you rotate in the 2D plane then roll the axes, now you are doing arbitrary reslicing and interpolation. We don't even have a concept of that right now anywhere in the code base. It would be feasible to restrict rotations to 2D planes and e.g. reset/ignore them when rolling the axes, but that seems icky, to say the least. I'm not sure it's desirable.

I think we need to find an acceptable compromise, rotations are really important to have (even if only for 2D planes).

Is it *silent* reset/ignore behaviour that gives you icky feelings? I personally would feel ok with a pop-up warning that says something like: "Hey you are trying to roll the axes but your image has been rotated in 2D, so this is not supported. Do you want to undo the rotation and continue rolling the axes, or do you want to cancel?"

Additionally, we could also support image flips in nd, and therefore get rotations that are multiples of 90 degrees and won't require arbitrary reslicing. I'm sure somebody wants this as a feature, even if it doesn't solve the specific use case presented here.

d-v-b:
Speaking as someone who needs support for arbitrary reslicing (which should probably happen as a corollary of #885 ), I'd be happy seeing this PR blocked until napari can do it naturally. I don't think implementing partial support for this feature by writing code for user warnings and undo operations is a good plan, given that a lot of that code would be totally obviated once arbitrary reslicing is around.

GenevieveBuckley:
> Speaking as someone who needs support for arbitrary reslicing (which should probably happen as a corollary of #885 )

Just FYI, arbitrary reslicing is not a direct corollary of https://github.com/napari/napari/pull/885, as Nick says there needs to be another PR for rotation/affine transformations before arbitrary slicing can exist.

d-v-b:
> Just FYI, arbitrary reslicing is not a direct corollary of #885, as Nick says there needs to be another PR for rotation/affine transformations before arbitrary slicing can exist.

Yeah, maybe corollary is too strong of a word :) I should have said, I really hope the world coordinates implementation makes it super easy to implement arbitrary reslicing (by allowing transformations of the clipping plane... Edit: or by allowing transformations of the data's coordinate space, which is what this PR would require)

sofroniewn:
This will be superseeded by #3577

PR #996, opened 2020-02-24
0.2.12 release notes edit

sofroniewn:
# Description
This PR is a follow on to #994 that includes the changes requested by @kne42 before we make the `0.2.12` release

PR #997, opened 2020-02-25
pin jupyter-client

kne42:
something in `jupyter-client` 6.0.0 broke; this PR puts a bandaid on that by pinning it until a solution can be determined


sofroniewn (approved):
Looks good to me! I'd like to merge then than update our #996 release notes before the `0.2.12` release. Sound good @tlambert03 @jni

Issue #793, opened 2019-12-13
Better generalize mouse functions


sofroniewn:
## 🚀 Feature
Following up on #785 it might be a good idea to think of a better generalization for our custom mouse functions. We should probably do this before we refactor our current mouse bindings as requested in #694 and before the refactoring proposed in #772.

We also now have more examples of people using custom mouse functions, for example from @tlambert03 in https://github.com/napari/napari/pull/675#issuecomment-553672258.

Possibly from @cudmoe - see https://github.com/napari/napari/issues/790#issue-537378307, and @VolkerH, see https://forum.image.sc/t/registering-callback-functions-for-events-in-napari/32210, and requests from @berl in https://github.com/napari/napari/issues/646.

This issue is meant to collect any such general feedback and request before @kne42 implements any improvements and be a place for @kne42 to propose such improvements for feedback

sofroniewn:
Also see - https://github.com/napari/napari/issues/517

tinku99:
I wanted to rebind mouse click drag to change brightness and contrast and change mouse scroll to change displayed image slice index.

kne42:
Hi @tinku99, so sorry that we missed your comment!

Could you explain how exactly you would like to control brightness/contrast with dragging?

We do not currently support binding mouse scroll events but we plan to in the future. I've just proposed a new system for mouse bindings in #1048 and would love to hear your input on if it covers your use cases and is not confusing!

PR #999, opened 2020-02-26
[WIP] Move InteractionBox to Base Layer

jojoelfe:
# Description
This is very much a premature PR to clarify some things. Right now it only contains the model and some tests. The plan for how it should behave is like this:

The base layer creates an instance of this class:
```
self._interaction_box = InteractionBox()
```
A layer can then use it either for a dragging select behavior like this
```
self._interaction_box.points = [drag_start, drag_end]
self._interaction_box.show_handle = False
self._interaction_box.show = True
```
or use it for indicating selection and translate, scale, rotate behavior
```
self._interaction_box.points = [list of selected points]
self._interaction_box.show_handle = True
self._interaction_box.show = True
```

A few questions:
- My feeling is this class should handle mouse events and needs to listen to mouse_press, _move, etc.. . My initial feeling was to call corresponding methods from the mouse_* methods of the base layer, but these will be overwritten by the subclasses, right?
- If the box gets rotated or moved it probably should communicate this to the layer with events. The correct thing is to give it a .events EmitterGroup property and have layers connect to them?
- Rendering ideally should be done in vispy_base_layer.py? Then there needs to be a way of creating a Marker and Line visual in the base layer. We could do this with:
```
self.node = Compound([node, Line(), Markers()])
```
but this seems to create a lot of issues in existing code that expects self.node to be a specific visual.

## Type of change

# References
This is based on discussion in #989 

# How has this been tested?

## Final checklist:

### DIFF
--- a/examples/interaction_box.py
+++ b/examples/interaction_box.py
@@ -6,10 +6,15 @@ from skimage import data
 import napari
 import numpy as np
 
+def on_selection_box_drag(event):
+    sel_i = napari.layers.points.points.points_in_box(event.box,viewer.active_layer._data_view,viewer.active_layer._size_view)
+    
 
 with napari.gui_qt():
     # create the viewer with an image
-    viewer = napari.view_image(data.astronaut(), rgb=True)
+    X, Y = np.mgrid[-500:500:50, -500:500:50]
+    positions = np.dstack([X.ravel(), Y.ravel()])
+    viewer = napari.view_points(positions[0,:,:])
     viewer.active_layer.interactive = False
-    viewer.active_layer._interaction_box.points = np.array([[50,50],[300,302]])
-    viewer.active_layer._interaction_box.show = True
+    
+    viewer.active_layer._interaction_box.events.selection_box_changed_drag.connect(on_selection_box_drag)
--- a/examples/interaction_box.py
+++ b/examples/interaction_box.py
@@ -1,0 +1,15 @@
+"""
+Demonstrate interaction box on image layer
+"""
+
+from skimage import data
+import napari
+import numpy as np
+
+
+with napari.gui_qt():
+    # create the viewer with an image
+    viewer = napari.view_image(data.astronaut(), rgb=True)
+    viewer.active_layer.interactive = False
+    viewer.active_layer._interaction_box.points = np.array([[50,50],[300,302]])
+    viewer.active_layer._interaction_box.show = True
--- a/examples/interaction_box.py
+++ b/examples/interaction_box.py
@@ -7,7 +7,8 @@ import napari
 import numpy as np
 
 def on_selection_box_drag(event):
-    sel_i = napari.layers.points.points.points_in_box(event.box,viewer.active_layer._data_view,viewer.active_layer._size_view)
+    #sel_i = napari.layers.points.points.points_in_box(event.box,viewer.active_layer._data_view,viewer.active_layer._size_view)
+    pass
     
 
 with napari.gui_qt():
@@ -15,6 +16,6 @@ with napari.gui_qt():
     X, Y = np.mgrid[-500:500:50, -500:500:50]
     positions = np.dstack([X.ravel(), Y.ravel()])
     viewer = napari.view_points(positions[0,:,:])
-    viewer.active_layer.interactive = False
+    viewer.layers.selection.active.interactive = False
     
-    viewer.active_layer._interaction_box.events.selection_box_changed_drag.connect(on_selection_box_drag)
+    viewer.layers.selection.active._interaction_box.events.selection_box_changed_drag.connect(on_selection_box_drag)
--- a/examples/interaction_box.py
+++ b/examples/interaction_box.py
@@ -1,21 +1,51 @@
 """
-Demonstrate interaction box on image layer
+Demonstrate interaction box on points layer
 """
 
 from skimage import data
 import napari
 import numpy as np
+from napari.layers.points._points_utils import points_in_box
+
 
 def on_selection_box_drag(event):
-    #sel_i = napari.layers.points.points.points_in_box(event.box,viewer.active_layer._data_view,viewer.active_layer._size_view)
-    pass
+    sel_i = points_in_box(event.box,viewer.layers.selection.active._view_data,viewer.layers.selection.active._view_size)
+    viewer.layers.selection.active.selected_data = sel_i
     
+def on_selection_box_final(event):
+    sel_i = points_in_box(event.box,viewer.layers.selection.active._view_data,viewer.layers.selection.active._view_size)
+    viewer.layers.selection.active.selected_data = sel_i
+    viewer.layers.selection.active._interaction_box.points = np.array([viewer.layers.selection.active._view_data[i] for i in sel_i])
+    viewer.layers.selection.active._interaction_box.show = True
+    viewer.layers.selection.active._preselect_points = [viewer.layers.selection.active._view_data[i] for i in sel_i]
 
-with napari.gui_qt():
-    # create the viewer with an image
-    X, Y = np.mgrid[-500:500:50, -500:500:50]
-    positions = np.dstack([X.ravel(), Y.ravel()])
-    viewer = napari.view_points(positions[0,:,:])
-    viewer.layers.selection.active.interactive = False
+
+def on_transform_changed_drag(event):
+    sel_i = viewer.layers.selection.active.selected_data
+    points = viewer.layers.selection.active._preselect_points
+
+    for i, index in enumerate(sel_i):
+        viewer.layers.selection.active._data[index] = event.transform(points[i])
+    viewer.layers.selection.active._update_dims()
+    viewer.layers.selection.active.events.data(value=viewer.layers.selection.active.data)
+
+def on_transform_changed_final(event):
+    sel_i = viewer.layers.selection.active.selected_data
     
-    viewer.layers.selection.active._interaction_box.events.selection_box_changed_drag.connect(on_selection_box_drag)
+    viewer.layers.selection.active._preselect_points = [viewer.layers.selection.active._view_data[i] for i in sel_i]
+
+
+    
+ 
+
+X, Y = np.mgrid[-500:500:50, -500:500:50]
+positions = np.dstack([X.ravel(), Y.ravel()])
+viewer = napari.view_points(positions[0,:,:])
+viewer.layers.selection.active.interactive = False
+
+viewer.layers.selection.active._interaction_box.events.selection_box_changed_drag.connect(on_selection_box_drag)
+viewer.layers.selection.active._interaction_box.events.selection_box_changed_final.connect(on_selection_box_final)
+viewer.layers.selection.active._interaction_box.events.transform_changed_drag.connect(on_transform_changed_drag)
+viewer.layers.selection.active._interaction_box.events.transform_changed_final.connect(on_transform_changed_final)
+
+napari.run()
--- a/napari/_qt/qt_viewer.py
+++ b/napari/_qt/qt_viewer.py
@@ -54,9 +54,9 @@ from .._vispy import (  # isort:skip
 if TYPE_CHECKING:
     from ..viewer import Viewer
 
+from .._vispy.vispy_interaction_box import VispyInteractionBox
 from ..settings import get_settings
 from ..utils.io import imsave_extensions
-from .._vispy.vispy_interaction_box import VispyInteractionBox
 
 
 class QtViewer(QSplitter):
@@ -189,10 +189,6 @@ class QtViewer(QSplitter):
         self._canvas_overlay.set_welcome_visible(show_welcome_screen)
         self._canvas_overlay.sig_dropped.connect(self.dropEvent)
 
-        self._interaction_box_visual = VispyInteractionBox(viewer)
-        self._interaction_box_visual.node.parent = self.view.scene
-        self._interaction_box_visual.node.order = len(self.viewer.layers) + 1
-
         main_widget = QWidget()
         main_layout = QVBoxLayout()
         main_layout.setContentsMargins(10, 22, 10, 2)
@@ -237,6 +233,9 @@ class QtViewer(QSplitter):
 
         # Add axes, scale bar
         self._add_visuals()
+        self._interaction_box_visual = VispyInteractionBox(viewer)
+        self._interaction_box_visual.node.parent = self.view.scene
+        self._interaction_box_visual.node.order = len(self.viewer.layers) + 1
 
         # Create the experimental QtPool for octree and/or monitor.
         self._qt_poll = _create_qt_poll(self, self.viewer.camera)
--- a/napari/_qt/qt_viewer.py
+++ b/napari/_qt/qt_viewer.py
@@ -56,6 +56,7 @@ if TYPE_CHECKING:
 
 from ..settings import get_settings
 from ..utils.io import imsave_extensions
+from .._vispy.vispy_interaction_box import VispyInteractionBox
 
 
 class QtViewer(QSplitter):
@@ -187,6 +188,10 @@ class QtViewer(QSplitter):
         self._canvas_overlay = QtWidgetOverlay(self, self.canvas.native)
         self._canvas_overlay.set_welcome_visible(show_welcome_screen)
         self._canvas_overlay.sig_dropped.connect(self.dropEvent)
+
+        self._interaction_box_visual = VispyInteractionBox(viewer)
+        self._interaction_box_visual.node.parent = self.view.scene
+        self._interaction_box_visual.node.order = len(self.viewer.layers) + 1
 
         main_widget = QWidget()
         main_layout = QVBoxLayout()
@@ -427,6 +432,7 @@ class QtViewer(QSplitter):
         vispy_layer.node.parent = self.view.scene
         vispy_layer.order = len(self.viewer.layers) - 1
         self.layer_to_visual[layer] = vispy_layer
+        self._interaction_box_visual.node.order = len(self.viewer.layers) + 1
 
     def _remove_layer(self, event):
         """When a layer is removed, remove its parent.
--- a/napari/_vispy/_tests/test_vispy_interaction_box.py
+++ b/napari/_vispy/_tests/test_vispy_interaction_box.py
@@ -1,0 +1,34 @@
+from napari import Viewer
+import numpy as np
+
+
+def test_vispy_interacton_box(qtbot):
+
+    viewer = Viewer()
+    view = viewer.window.qt_viewer
+    qtbot.addWidget(view)
+
+    np.random.seed(0)
+    data = np.random.random((10, 15, 8))
+    layer = viewer.add_image(data)
+    layer._interaction_box.show = True
+    layer._interaction_box.points = np.array([[0, 0], [100, 100]])
+
+    assert (
+        view._interaction_box_visual.marker_node._data['a_position'].shape[0]
+        == 9
+    )
+
+    viewer.add_image(data)
+
+    assert (
+        view._interaction_box_visual.marker_node._data['a_position'].shape[0]
+        == 1
+    )
+
+    viewer.active_layer = None
+
+    assert (
+        view._interaction_box_visual.marker_node._data['a_position'].shape[0]
+        == 1
+    )
--- a/napari/_vispy/vispy_interaction_box.py
+++ b/napari/_vispy/vispy_interaction_box.py
@@ -9,6 +9,7 @@ class VispyInteractionBox:
         self.node = Compound([Line(), Markers()])
         self.layer = None
         self._viewer.events.active_layer.connect(self._on_active_layer_change)
+        self._on_interaction_box_change()
 
     @property
     def marker_node(self):
--- a/napari/_vispy/vispy_interaction_box.py
+++ b/napari/_vispy/vispy_interaction_box.py
@@ -36,7 +36,7 @@ class VispyInteractionBox:
         """Called whenever the interaction box changed.
         """
         if self.layer is None:
-            self.marker_node.set_data(np.empty((0, 2)))
+            self.marker_node.set_data(np.zeros((1, 2)))
             self.line_node.set_data(np.zeros((1, 2)))
             return
 
--- a/napari/_vispy/vispy_interaction_box.py
+++ b/napari/_vispy/vispy_interaction_box.py
@@ -1,0 +1,76 @@
+from vispy.scene.visuals import Line, Compound, Markers
+import numpy as np
+
+
+class VispyInteractionBox:
+    def __init__(self, viewer):
+
+        self._viewer = viewer
+        self.node = Compound([Line(), Markers()])
+        self.layer = None
+        self._viewer.events.active_layer.connect(self._on_active_layer_change)
+
+    @property
+    def marker_node(self):
+        """sequence of float: Scale factors."""
+        return self.node._subvisuals[1]
+
+    @property
+    def line_node(self):
+        """sequence of float: Scale factors."""
+        return self.node._subvisuals[0]
+
+    def _on_active_layer_change(self, event):
+        if self.layer is not None:
+            self.layer._interaction_box.events.points_changed.disconnect(
+                self._on_interaction_box_change
+            )
+        self.layer = event.item
+        if self.layer is not None:
+            self.layer._interaction_box.events.points_changed.connect(
+                self._on_interaction_box_change
+            )
+        self._on_interaction_box_change()
+
+    def _on_interaction_box_change(self, event=None):
+        """Called whenever the interaction box changed.
+        """
+        if self.layer is None:
+            self.marker_node.set_data(np.empty((0, 2)))
+            self.line_node.set_data(np.zeros((1, 2)))
+            return
+
+        # Compute the location and properties of the vertices and box that
+        # need to get rendered
+        (
+            vertices,
+            face_color,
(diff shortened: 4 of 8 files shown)


sofroniewn:
Thanks @jojoelfe - sorry for not getting to this yet. It looks very exciting and that you're on the right track. I only have time to give some quick pointers right now, but will try and do a deeper dive on Friday.

> My feeling is this class should handle mouse events and needs to listen to mouse_press, move, etc.. . My initial feeling was to call corresponding methods from the mouse* methods of the base layer, but these will be overwritten by the subclasses, right?

Yeah, we have some plans to do some refactoring to actually make this easier in #694. I can also try and work on this over the weekend, which I think will help here.

> If the box gets rotated or moved it probably should communicate this to the layer with events. The correct thing is to give it a .events EmitterGroup property and have layers connect to them?

This sounds about good, and consistent with our current practices.

> Rendering ideally should be done in vispy_base_layer.py? Then there needs to be a way of creating a Marker and Line visual in the base layer. We could do this with:
`self.node = Compound([node, Line(), Markers()])`
but this seems to create a lot of issues in existing code that expects self.node to be a specific visual.

Yeah, I've been a little worried about that. The main problem I see is that even after the refactor some layer visuals have a single visual, but some will have two, one for the main visual and one for highlighting. Another way to think about this would be to have the `interaction box` visual get added only once to the canvas in the qt_viewer and then each layer can just update that visual with its interaction box, instead of the base layer having the visual and each layer having one. This will be much more performant too as if you have 10 layers instead of having ~30 visuals you'll still have ~10 visuals. Does that make sense? Do you think that sounds reasonable?

jojoelfe:
@sofroniewn 

> Yeah, I've been a little worried about that. The main problem I see is that even after the refactor some layer visuals have a single visual, but some will have two, one for the main visual and one for highlighting. Another way to think about this would be to have the interaction box visual get added only once to the canvas in the qt_viewer and then each layer can just update that visual with its interaction box, instead of the base layer having the visual and each layer having one. This will be much more performant too as if you have 10 layers instead of having ~30 visuals you'll still have ~10 visuals. Does that make sense? Do you think that sounds reasonable?

That makes a lot of sense. I just pushed some code where the baselayer adds an additional node, which would results in the layer explosion you mentioned. It would be good to have a dedicated layer for this with an order of ```len(layers) + 1``` to keep it on top of everything else. 
The other thing I was worried about is that every layer would have to manually hide it's interaction box when it's not active. By having a single vispy visual that only renders the interaction box of the active layer we get that behaviour for free.

sofroniewn:
> The other thing I was worried about is that every layer would have to manually hide it's interaction box when it's not active. By having a single vispy visual that only renders the interaction box of the active layer we get that behaviour for free.

Yes exactly - I'm imagining there will now be a separate `vispy_interaction_box.py` file and that will get instantiated once when the canvas is created around here in the code base:

https://github.com/napari/napari/blob/723c937d47c266aa37916c3ab5f06ca4daf7ffaf/napari/_qt/qt_viewer.py#L146-L150

Then when the active layer is changed the interaction box visual will get cleared / hidden / swapped to the interaction box of the new layer. That will happen in the `qt_viewer.py` too, but you might now need to add a new function that gets connected to the `active_layer` event that gets emitted from the viewer model here https://github.com/napari/napari/blob/723c937d47c266aa37916c3ab5f06ca4daf7ffaf/napari/components/viewer_model.py#L239

Note each I think that the actual interaction box model code can still live on the base model layer code as you have right now with the `_interaction_box`.

How does this sound?

I also want to give you a heads up that there is a proposal for refactoring our model / view separation in #1006 by adding an explicit controller. That will take a bit of time for the team to discuss and if we do decide to do it will take a bit of time to implement, so for now I'd recommend keeping on going with the current architecture, but it might be worth looking #1006 over too just to keep those ideas in mind.

sofroniewn:
This is looking great @jojoelfe, I really like how the code is shaping up.

It looks like a big thing still missing at this point are the mouse functions. I think there we have some plans to generalize our mouse functions, see #793 at which point you could add mouse functions more easily to the interaction box (maybe using a mixin). For now though you might be best off adding them to the viewer and then having a quick way to determine if the interaction box is visible or not and whether they need to be called. We're moving towards the generator syntax described in #544 across the whole code base(see for this for the Labels layer for example in #1010, which is still in progress) and it would be nice to use that syntax if possible.

Thinking more about

> If the box gets rotated or moved it probably should communicate this to the layer with events. The correct thing is to give it a .events EmitterGroup property and have layers connect to them?

I think that making the connections directly inside the layer for now is fine, but that would be something that would go inside the controller if we move forward with #1006 I think (right @shanaxel42?) and could look much cleaner then. 

I'm curious how you're thinking about the scope of this PR and whether you'll actually start using the interaction box in the shapes layer in this PR or wait for a followup PR to hook it up there and remove the duplicate code? It might be easier for the rest of the team to review this PR if all it does is add the interaction box and we do the shapes stuff in a follow-up. If that is the case you might want to think about adding a small little example that forces the interaction box to display as it would be nice to have a way to see it work. You could add some simple callbacks in the example to print things out too.

jojoelfe:
> We're moving towards the generator syntax described in #544 across the whole code base(see for this for the Labels layer for example in #1010, which is still in progress) and it would be nice to use that syntax if possible.

Yes, my next plan was to get the mouse events hooked up. I'll try to do it with the generator functions. 

> I think that making the connections directly inside the layer for now is fine, but that would be something that would go inside the controller if we move forward with #1006 I think (right @shanaxel42?) and could look much cleaner then.

I'll keep an eye out for #1006 and see how things develop. 

> I'm curious how you're thinking about the scope of this PR and whether you'll actually start using the interaction box in the shapes layer in this PR or wait for a followup PR to hook it up there and remove the duplicate code? It might be easier for the rest of the team to review this PR if all it does is add the interaction box and we do the shapes stuff in a follow-up. If that is the case you might want to think about adding a small little example that forces the interaction box to display as it would be nice to have a way to see it work. You could add some simple callbacks in the example to print things out too.

I think it makes sense to keep this PR to the API generation for the interaction_box and then have a separate PR that makes the shapes (and maybe points) layer use it. The example sounds like a good idea to show how things behave.

napari/layers/interaction_box.py
```diff
@@ -1,9 +1,44 @@
 import numpy as np
+from skimage.transform import SimilarityTransform
 
 from ._constants import Box
 from ..utils.event import EmitterGroup, Event
 
 
+def inside_boxes(boxes):
```
jojoelfe, line 8: I copied this function from shape_utils.py, because I couldn't import it. As far as I can see this function is not used anywhere else.

sofroniewn: That's fine, we can remove from Shapes in follow up PR

sofroniewn:
Hi @jojoelfe - apologies it's been so long getting back to you on this one. Maybe @andy-sweet can pick this up and do some review of it and we can decide how to move forward. As for context @andy-sweet some of the original motivation was to enable functionality like #989 which will now be much easier as we have a transforms class.

This interaction box could also be the foundation for a highlight layer - see discussion in #2560, #2954. #772 also has additional context that might help - cc @nclack

andy-sweet:
Hey @jojoelfe - the idea and functionality look great here, so I'm happy to help out however I can.

I took a quick look tonight and also tried to get `examples/interaction_box.py` working.

At first, I attempted to create an environment based on the state of the repo at the last commit of this PR. But I got a vispy dependency error that didn't immediately make sense.

So I decided to merge main into this and handle that pain instead. There aren't many automatic conflicts, but there are plenty of updates needed to get that example working again. I made enough updates so that calling `napari examples/interaction_box.py`, but when I actually start interaction with viewer I get new ones.

I'm happy to keep working on resolve those until things seem fairly functional. But if you still have interest working on this, then it might be quicker for you to resolve some of those. Or I'm also happy to find some time together to combine our powers.

jojoelfe:
Hi @sofroniewn  and @andy-sweet ,

yes, sorry I also got distracted. But I am still very interested in getting this working. It might take me a few days to get up to speed to remember what I have done and combine it with changes since then, but if you are not in a super rush it's probably easier for me to merge the current main and get things working again.

sofroniewn:
> It might take me a few days to get up to speed to remember what I have done and combine it with changes since then, but if you are not in a super rush it's probably easier for me to merge the current main and get things working again.

Awesome, yes! Great if you can pick this back up again when convenient for you - no rush though. @andy-sweet can then help with the review and make sure we get it merged this time around.

Also as a note for #989 this should be much much easier now that we've added support for affine transforms on all our layers

jojoelfe:
Hi @andy-sweet, I merged with main and caught up with enough changes to the coordinate system and events to get the example running again. I haven't really thought about whether the design still fits after these changes, but you can give it a try.

The general idea was for the interaction box two have two modes:

- A selection mode where due to a mouse-drag a box is drawn and events in the active layer are triggered, which can then decide whether items should be selected

- A transformation mode where the box is always drawn and has handles for rotation+scaling. When the box is then transformed it maintains an affine transformation which then gets passed to the active layer with events. The active layer can then decide whether to apply the transform to the complete layer (for images) or only selected items (for points)

The example only covers the first mode for now.

andy-sweet:
That was quick @jojoelfe! I'll try to play with and review this on Monday, so expect feedback by the end of Monday or early Tuesday. Hope that's OK!

examples/interaction_box.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def on_selection_box_drag(event):
+    sel_i = points_in_box(event.box,viewer.layers.selection.active._view_data,viewer.layers.selection.active._view_size)
+    viewer.layers.selection.active.selected_data = sel_i
+    
+def on_selection_box_final(event):
+    sel_i = points_in_box(event.box,viewer.layers.selection.active._view_data,viewer.layers.selection.active._view_size)
+    viewer.layers.selection.active.selected_data = sel_i
+    viewer.layers.selection.active._interaction_box.points = np.array([viewer.layers.selection.active._view_data[i] for i in sel_i])
```
andy-sweet, line 18:
```suggestion
    viewer.layers.selection.active._interaction_box.points = viewer.layers.selection.active._view_data[sel_i]
```

If the interaction box needs a copy of the data, you can wrap the RHS in `np.array(...)`.

There are some other similar changes in this file, so if this looks good, please make those too.

napari/_qt/qt_viewer.py
```diff
@@ -232,6 +233,9 @@ def __init__(self, viewer: Viewer, show_welcome_screen: bool = False):
 
         # Add axes, scale bar
         self._add_visuals()
+        self._interaction_box_visual = VispyInteractionBox(viewer)
+        self._interaction_box_visual.node.parent = self.view.scene
+        self._interaction_box_visual.node.order = len(self.viewer.layers) + 1
```
andy-sweet, line 238: Is there a node with order `len(self.viewer.layers)`? If not, could we remove the `+ 1`?

napari/_vispy/_tests/test_vispy_interaction_box.py
```diff
@@ -0,0 +1,34 @@
...
+    qtbot.addWidget(view)
+
+    np.random.seed(0)
+    data = np.random.random((10, 15, 8))
+    layer = viewer.add_image(data)
+    layer._interaction_box.show = True
+    layer._interaction_box.points = np.array([[0, 0], [100, 100]])
+
+    assert (
+        view._interaction_box_visual.marker_node._data['a_position'].shape[0]
```
andy-sweet, line 18: Style/Nit: extract this as a local variable to make the assertions easier to read. E.g.

`marker_pos = view._interaction_box_visual.marker_node._data['a_position']`

napari/_vispy/_tests/test_vispy_interaction_box.py
```diff
@@ -0,0 +1,34 @@
...
+        == 9
+    )
+
+    viewer.add_image(data)
+
+    assert (
+        view._interaction_box_visual.marker_node._data['a_position'].shape[0]
```
andy-sweet, line 25: Can this be `marker_node.pos` instead?

napari/_vispy/_tests/test_vispy_interaction_box.py
```diff
@@ -0,0 +1,34 @@
...
+        == 1
```
andy-sweet, line 26: Why 1 instead of 9 here? I thought the 9 came from the 9 marker control points of the interaction box (rather than being specific to the points layer) - is that right? Did adding an image make the interaction box effectively hidden?

napari/_vispy/vispy_interaction_box.py
```diff
@@ -0,0 +1,79 @@
...
+            edge_color,
+            pos,
+            width,
+        ) = self.layer._interaction_box._compute_vertices_and_box()
+
+        if vertices is None or len(vertices) == 0:
+            vertices = np.zeros((1, self._viewer.dims.ndisplay))
+            size = 0
+        else:
+            vertices = vertices + 0.5
```
andy-sweet, line 60: Why is this 0.5 needed? Can you add an implementation comment to explain?

napari/layers/_constants.py
```diff
@@ -0,0 +1,16 @@
+class Box:
+    """Box: Constants associated with the vertices of the interaction box
+    """
+
+    WITH_HANDLE = [0, 1, 2, 3, 4, 5, 6, 7, 9]
+    LINE_HANDLE = [7, 6, 4, 2, 0, 7, 9]
```
andy-sweet, line 6: I assume that 7 is in the list twice for a deliberate repeat? Consider adding an implementation comment explain why that repeat is needed.

napari/layers/_constants.py
```diff
@@ -0,0 +1,16 @@
+class Box:
+    """Box: Constants associated with the vertices of the interaction box
```
andy-sweet, line 2: Would be good to clarify the integer values are indices of the vertices.

napari/layers/_constants.py
andy-sweet: Also, does order matter here? This seems to go clockwise around the box, whereas the others go anti-clockwise.

napari/layers/_constants.py
```diff
@@ -0,0 +1,16 @@
...
+    WITHOUT_HANDLE = [0, 2, 4, 6]
+    LINE = [0, 2, 4, 6, 0]
+    TOP_LEFT = 0
+    TOP_CENTER = 7
+    LEFT_CENTER = 1
+    BOTTOM_RIGHT = 4
+    BOTTOM_LEFT = 2
+    CENTER = 8
+    HANDLE = 9
+    LEN = 8
```
andy-sweet, line 16: Is this supposed to be the number of vertices (i.e. is `LEN` like `len()` and short for length)? If so, shouldn't it be 10? If I misunderstood it might be good to rename this or add an explanatory implementation comment.

But, if it is meant to be the number of vertices, consider defining an Enum like `BoxVertex` (or maybe `BoxVertexIndex`) with just the single indices. Then you can access that as `len(BoxVertex)`. Then maybe you could define `BoxVertexGroup` for the others.

napari/layers/_constants.py
```diff
@@ -0,0 +1,16 @@
+class Box:
+    """Box: Constants associated with the vertices of the interaction box
+    """
+
+    WITH_HANDLE = [0, 1, 2, 3, 4, 5, 6, 7, 9]
+    LINE_HANDLE = [7, 6, 4, 2, 0, 7, 9]
+    WITHOUT_HANDLE = [0, 2, 4, 6]
+    LINE = [0, 2, 4, 6, 0]
+    TOP_LEFT = 0
```
andy-sweet, line 9: It would be easier to understand the groups like `WITHOUT_HANDLE` if all single vertices were defined by name and integer index first. I.e. put the following first:

```python
TOP_LEFT = 0
LEFT_CENTER = 1
...
HANDLE = 9
```

then

```python
WITHOUT_HANDLE = [0, 2, 4, 6]
LINE = WITHOUT_HANDLE + [0]
```

You might also want to consider reusing the individual vertex names like
```python
WITHOUT_HANDLE = [TOP_LEFT, BOTTOM_LEFT, BOTTOM_RIGHT, TOP_RIGHT]
```

though that might be a bit verbose.

napari/layers/_tests/test_interaction_box.py
```diff
@@ -0,0 +1,99 @@
+import numpy as np
+
+from ..interaction_box import InteractionBox
+
+
+def test_interaction_box_initialization():
```
andy-sweet, line 6: I don't think we're consistent across the napari tests with stuff like this, but it would be good to rename this as the shorter `test_init` because it's already in a test module/suite/file that includes `interaction_box`.

Also, I prefer making each test capture one behavior, so that specific changes in behavior can be easily identified. So I would personally prefer to see this split up into a few tests with names like `test_init`, `test_init_without_handle`, `test_init_hide`, etc.

napari/layers/interaction_box.py
```diff
@@ -0,0 +1,444 @@
...
+        self._show_vertices = show_vertices
+        self.events.points_changed()
+
+    @property
+    def angle(self):
+        offset = self._box[Box.HANDLE] - self._box[Box.CENTER]
+        angle = -np.degrees(np.arctan2(offset[0], -offset[1])) - 90
+        return angle
+
+    def points_in_box(self, points):
```
andy-sweet, line 131: This implementation is empty. I guess we plan to use the one in `_points_utils.py` instead?

napari/layers/interaction_box.py
```diff
@@ -0,0 +1,444 @@
...
+            else:
+                box = self._box[Box.WITHOUT_HANDLE]
+
+            if self._selected_vertex is None:
+                face_color = 'white'
+            else:
+                face_color = self._highlight_color
+
+            edge_color = self._highlight_color
+            vertices = box[:, ::-1]
```
andy-sweet, line 173: Either remove this or the `if` statement. Or consider a ternary statement like

```python
vertices = box[:, ::-1] if self._show_vertices else np.empty((0, 2))
```

andy-sweet (commented):
I'm still looking through most of the guts of `vispy_interaction_box.py`, so I'm only half way through of my review, but I thought I'd leave my comments for now, so you can act on those if you want.

In general, the functionality here is really nice and I liked playing around with the example. I did run into a couple of general issues while doing that, but have not identified the cause yet.
1. If you set the points layer to have a rotation (i.e. run `viewer.layers[0].rotate = 45`) after initializing the example, the interaction does not behave correctly. I'm not sure if that also occurs if you set the rotation when adding the layer.
2. At some point I got in a bad state where it seemed like nothing was selected, but trying to make a new selection by dragging caused some points to transform. I think I saw two boxes, that weren't always synced. But I'm not sure how to reproduce this - sorry!

I'm also still trying to understand how this will generalize to other layers. I think it would be good to add an image to the example, to understand how well that works.

andy-sweet:
Also, I will try to finish this review tomorrow - sorry about the delay!

jojoelfe:
Hi @andy-sweet,

thanks for the review so far. My thoughts about you points are:

1. I think this is be due to the fact that I have to make a decision whether the interaction box internally operates in canvas, world, or layer coordinates and then where the ultimate conversion into layer coordinates happens. My initial guess is that the InteractionBox should operate in canvas coordinates, since like the canvas it fundamentally operates in 2D and not in nD. Then there would have to be layer and viewer-mode specific code to convert into layer coordinates.

2. I encountered 2 scenarios where weird things happen. (a) When a single point is selected which produces a box with size 0 (b) When the layer mode is changed, which should enable or disable the InteractionBox, but isn't yet.

Also I just noticed that there are now overlays in the vispy code (for scalebar and axes). It seems to me like this would be a good place for the InteractionBox too and would get rid of the "len(layers) + 1" order you wondered about.

I will go ahead and make an example for the image layer.

Johannes

napari/layers/interaction_box.py
```diff
@@ -0,0 +1,444 @@
...
+    c2 = ABAM <= ABAB
+    c3 = 0 <= BCBM
+    c4 = BCBM <= BCBC
+
+    inside = np.all(np.array([c1, c2, c3, c4]), axis=0)
+
+    return inside
+
+
+class InteractionBox:
```
sofroniewn, line 42: Could you make this a pydantic evented model like the ScaleBar & Axes. It means you wouldn't need to add the `EmitterGroup` yourself as that would come for free

sofroniewn (commented):
> Also I just noticed that there are now overlays in the vispy code (for scalebar and axes). It seems to me like this would be a good place for the InteractionBox too and would get rid of the "len(layers) + 1" order you wondered about.

Yes, I like the idea of treating the interaction box like an overlay. My apologies for not mentioning this above - but I think that might be a better design. In that case I guess the box doesn't live on the layer itself, but lives on the viewer. It might be good to make a "highlight" object on the viewer and then have the interaction box inside there, so `viewer.highlights.interaction_box` or `viewer.overlays.interaction_box` and move scale bar/ axes over there too. We could add `viewer.interaction_box` to start too (@tlambert03 @andy-sweet any preference here?)

napari/layers/interaction_box.py
sofroniewn: https://github.com/napari/napari/blob/3776819578ecf2785aaed4c0f464658609ea5272/napari/components/axes.py#L4

sofroniewn:
I'll add I like `viewer.highlights.interaction_box` or `viewer.overlays.interaction_box` as we could put a ruler here too cc @nclack, it just seems better than putting them all on the viewer directly

napari/layers/interaction_box.py
```diff
@@ -0,0 +1,444 @@
...
+        box = self._box
+
+        if box is not None:
+            rot = box[Box.TOP_CENTER]
+            length_box = np.linalg.norm(
+                box[Box.BOTTOM_LEFT] - box[Box.TOP_LEFT]
+            )
+            if length_box > 0:
+                r = self._rotation_handle_length
+                rot = (
```
andy-sweet, line 208: The formatting here makes the operation a little hard to parse.

Maybe the best thing to do here is to store a local variable and do something like

```python
left_edge = box[Box.BOTTOM_LEFT] - box[Box.TOP_LEFT]
height = np.linalg.norm(left_edge)
...
rot -= r * left_edge / height
```

napari/layers/interaction_box.py
```diff
@@ -0,0 +1,444 @@
...
+        self.show = False
+        self.show_handle = True
+        self.show_vertices = True
+        # self.points = None
+        if self._box is not None:
+            self.events.selection_box_changed_final(
+                box=self._box[Box.WITHOUT_HANDLE]
+            )
+
+    def initialize_mouse_events(self, layer):
```
andy-sweet, line 373: The only use of this is immediately after creating `InteractionBox` in the base layer.

Instead, could we make this private and just call this at the end of `__init__`. Or do we expect to use this elsewhere?

andy-sweet (commented):
Sorry, I'm still making my way through `interaction_box.py`, but I had a couple more comments that might be useful, so I added those.

> My initial guess is that the InteractionBox should operate in canvas coordinates, since like the canvas it fundamentally operates in 2D and not in nD.

Yes, this sounds reasonable to me. Given that we often use `Layer.world_to_data` (which includes any grid offset transforms) to convert from `event.position` in a mouse related callback, I'm assuming that `event.position` are in the world coordinate system, but restricted to the displayed dimensions. Is that what you mean by canvas coordinates, or is there another transform I'm missing?

If those are canvas coordinates, then it seems like the box coordinates should always defined in that space, so translation, scale, rotation of the box is straightforward. When we want to transform the actual points data, I think we can apply the same transform to the points in canvas coordinates, then apply the `world_to_data` transform to define the new data (in data coordinates). I guess the other situation we care about is defining which points are within the box as it's initially defined - I think that's most easily done in the canvas coordinates because the selection box will be axis-aligned in that coordinate space.

andy-sweet:
> It might be good to make a "highlight" object on the viewer and then have the interaction box inside there, so `viewer.highlights.interaction_box` or `viewer.overlays.interaction_box` and move scale bar/ axes over there too. We could add `viewer.interaction_box` to start too (@tlambert03 @andy-sweet any preference here?)

I don't have a strong preference, but I'm also not sure exactly what highlight means in this context. I think that highlights are used to describe selected visuals (e.g. a selected point has a visible `MarkerVisual` highlight). But maybe they are more generally used to describe things that also define those selections and are interactive? Whereas overlays are non-interactive?

sofroniewn:
> I don't have a strong preference, but I'm also not sure exactly what highlight means in this context. I think that highlights are used to describe selected visuals (e.g. a selected point has a visible MarkerVisual highlight). But maybe they are more generally used to describe things that also define those selections and are interactive? Whereas overlays are non-interactive?

One thought might be if the highlight/ overlay doesn't depend on the layer data (i.e. is a bounding box or a ruler etc) then it should live on the viewer, but if it depends on the layer data (e.g. outlines of the selected shapes) then it should live on the layer. i don't know though, that's just a thought

jojoelfe:
I have been refactoring this as an overlay and using an EventedModel from scratch in a separate branch. Would it be better to close this and open a new PR or should I force-push the commits into this branch?

andy-sweet:
> I have been refactoring this as an overlay and using an EventedModel from scratch in a separate branch. Would it be better to close this and open a new PR or should I force-push the commits into this branch?

If it's a fairly different approach (and it sounds like it is), then I would probably open a new PR and close this one. You can always refer back to this PR if you resolved some things mentioned here.

But feel free to do either.

jojoelfe:
I made a new pull-request for the refactor at #3577

PR #1002, opened 2020-02-26
Fix ability to test released distributions

tlambert03:
# Description
following the conversation on zulip.  This PR fixes tests in our released distributions as follows:

1. moves everything in `napari.plugin.tests` to `napari.plugins._tests`
2. adds `recursive-include napari/plugins/_tests/fixtures` to `MANIFEST`
3. removes the top level `conftest.py` file and moves the fixture that was in `_tests/infra/viewer.py` into `_qt/_tests/conftest.py` (`_qt` is currently the only subpackage that uses the `viewermodel_factory` fixture... so we don't need a global `conftest.py` yet)

## how i tested
i ran `python setup.py sdist`, then `cd`ed into newly created dist folder, confirmed that all the required files are included, and ran pytest successfully within the sdist folder.


sofroniewn (approved):
This looks good to me, but can you add a line to the `0.2.12` release note with (1002) in there too before merge @tlambert03

sofroniewn:
It also looks like this change might fix our linux 3.7 test

jni (approved):
Thank you @tlambert03! Installing that sdist in a fresh env would be a further useful test! But this sounds good to me

tlambert03:
> It also looks like this change might fix our linux 3.7 test

well that would be amazing.  I'll believe it when i (continue to) see it :)

> can you add a line to the 0.2.12 release note with (1002) in there too before merge @tlambert03

sure.  added to support

tlambert03:
just tested the sdist locally in a clean env and everything passed

PR #1006, opened 2020-02-27
RFC: Refactor proposal! Adding a controllers to napari 

shanaxel42:
The current flow of data in napari is through the use of events emitted. A user can modify a layer through the gui or directly in a jupyter notebook or ipython terminal. 

The following  diagram charts the flow of calls when a user modifies the interpolation attribute of an image in the second scenario. 

napari_data_flow_1

This system works alright but when a user modifies the interpolation attribute through the gui dropdown menu the flow is a bit more complicated, see below:

napari_data_flow_2

The current architecture:

- makes it hard to debug, and track everything that happens when changes are made
- causes repeat calls of methods 
- forces the qt, vispy connection through qt_viewer

This proposal adds a controllers to the codebase that are the central place to process data updates. A layer's controller has access to all three components (data, qt, and vispy). When changes are made the controller is the only component that receives events, it then updates the layer, qt_controls, and vispy_layer. See modified architecture below:

proposal_napari_data_flow

This simplifies the codebase by:

-  Only having one method to track down that shows everything that happens on a user change of any kind
- Neither the vispy layer or qt  controls will need to connect to events, they will only be responsible for updating themselves. 
- When something goes wrong the error message will be much easier to parse 
- There won't be any need to ever suppress events, or risk of causing an endless update loop
- eventually qt_viewer will not need to keep track of each layer's vispy mapping and can just call the controllers instead. 

The code in this PR is just a small example I made for feedback. I added everything needed for napari to process image interpolation changes from either the layer level or gui level through the new image_controller. The whole refactor would be significant but I think it would be really helpful longterm for readability and debugging. Also from a plugin standpoint we could add hooks to the controller methods and not have to worry about updating anything else. 

Looking for all the feedback, likes, dislikes, and concerns!! 

NOTE: The tests and build will definitely be broken since this is just a partial refactor meant as an example

### DIFF
--- a/napari/_qt/layers/qt_image_layer.py
+++ b/napari/_qt/layers/qt_image_layer.py
@@ -1,5 +1,6 @@
 from qtpy.QtCore import Qt
 from qtpy.QtWidgets import QLabel, QComboBox, QSlider
+from ...utils.event import Event
 from .qt_image_base_layer import QtBaseImageControls
 from ...layers.image._constants import Interpolation, Rendering
 
@@ -8,7 +9,8 @@ class QtImageControls(QtBaseImageControls):
     def __init__(self, layer):
         super().__init__(layer)
 
-        self.layer.events.interpolation.connect(self._on_interpolation_change)
+        self.events.add(interpolation=Event)
+
         self.layer.events.rendering.connect(self._on_rendering_change)
         self.layer.events.iso_threshold.connect(self._on_iso_threshold_change)
         self.layer.events.attenuation.connect(self._on_attenuation_change)
@@ -20,7 +22,7 @@ class QtImageControls(QtBaseImageControls):
             self.layer.interpolation, Qt.MatchFixedString
         )
         interp_comboBox.setCurrentIndex(index)
-        interp_comboBox.activated[str].connect(self.changeInterpolation)
+        interp_comboBox.activated[str].connect(self.emit_interpolation_event)
         self.interpComboBox = interp_comboBox
         self.interpLabel = QLabel('interpolation:')
 
@@ -80,8 +82,12 @@ class QtImageControls(QtBaseImageControls):
         self.grid_layout.setColumnStretch(1, 1)
         self.grid_layout.setVerticalSpacing(4)
 
-    def changeInterpolation(self, text):
-        self.layer.interpolation = text
+    def emit_interpolation_event(self, text):
+        self.events.interpolation(interpolation=text)
+
+    def set_interpolation(self, text):
+        index = self.interpComboBox.findText(text, Qt.MatchFixedString)
+        self.interpComboBox.setCurrentIndex(index)
 
     def changeRendering(self, text):
         self.layer.rendering = text
@@ -102,13 +108,6 @@ class QtImageControls(QtBaseImageControls):
     def _on_attenuation_change(self, event):
         with self.layer.events.attenuation.blocker():
             self.attenuationSlider.setValue(self.layer.attenuation * 100)
-
-    def _on_interpolation_change(self, event):
-        with self.layer.events.interpolation.blocker():
-            index = self.interpComboBox.findText(
-                self.layer.interpolation, Qt.MatchFixedString
-            )
-            self.interpComboBox.setCurrentIndex(index)
 
     def _on_rendering_change(self, event):
         with self.layer.events.rendering.blocker():
--- a/napari/layers/image/image.py
+++ b/napari/layers/image/image.py
@@ -213,7 +213,7 @@ class Image(IntensityVisualizationMixin, Layer):
         self._contrast_limits = tuple(self.contrast_limits_range)
         self.colormap = colormap
         self.contrast_limits = self._contrast_limits
-        self.interpolation = interpolation
+        self._set_interpolation(interpolation)
         self.rendering = rendering
 
         # Trigger generation of view slice and thumbnail
@@ -335,10 +335,12 @@ class Image(IntensityVisualizationMixin, Layer):
 
     @interpolation.setter
     def interpolation(self, interpolation):
+        self.events.interpolation(interpolation=interpolation)
+
+    def _set_interpolation(self, interpolation):
         if isinstance(interpolation, str):
             interpolation = Interpolation(interpolation)
         self._interpolation = interpolation
-        self.events.interpolation()
 
     @property
     def rendering(self):
--- a/napari/_controllers/__init__.py
+++ b/napari/_controllers/__init__.py
@@ -1,0 +1,6 @@
+from ..layers import Image
+from .image_controller import ImageController
+
+layer_to_controller = {
+    Image: ImageController,
+}
--- a/napari/_controllers/_base.py
+++ b/napari/_controllers/_base.py
@@ -1,0 +1,30 @@
+from napari._qt.layers.qt_image_base_layer import QtBaseImageControls
+from napari._vispy.vispy_base_layer import VispyBaseLayer
+from napari.layers import Layer
+
+
+class ControllerBase:
+    """
+    Base layer controller class responsible for the interactions between vispy, qt and the data
+    """
+
+    def __init__(
+        self,
+        layer: Layer,
+        qt_controls: QtBaseImageControls,
+        vispy_layer: VispyBaseLayer,
+    ):
+        """
+        Parameters
+        ----------
+        layer: Layer
+            The data layer
+        qt_controls: QtBaseImageControls
+            The qt_controls attached to the layer
+        vispy_layer: VispyBaseLayer
+            The vispy rendering of the layer
+        """
+
+        self.layer = layer
+        self.qt_controls = qt_controls
+        self.vispy_layer = vispy_layer
--- a/napari/_controllers/image_controller.py
+++ b/napari/_controllers/image_controller.py
@@ -1,0 +1,35 @@
+from ._base import ControllerBase
+
+from ..layers.image.image import Image
+from .._qt.layers.qt_image_layer import QtImageControls
+from .._vispy.vispy_image_layer import VispyImageLayer
+
+
+class ImageController(ControllerBase):
+    def __init__(
+        self,
+        layer: Image,
+        qt_controls: QtImageControls,
+        vispy_layer: VispyImageLayer,
+    ):
+        super().__init__(layer, qt_controls, vispy_layer)
+
+        # connect to specific image layer events
+        self.layer.events.interpolation.connect(self.on_interpolation_change)
+
+        # connect to qt events
+        self.qt_controls.events.interpolation.connect(
+            self.on_interpolation_change
+        )
+
+    def on_interpolation_change(self, event=None):
+        """
+        Process changes when the interpolation attribute is changed from any interface
+        """
+        value = event.interpolation
+        # update layer
+        self.layer._set_interpolation(value)
+        # update qt
+        self.qt_controls.set_interpolation(value)
+        # update vispy
+        self.vispy_layer.set_interpolation(value)
--- a/napari/_qt/layers/qt_image_base_layer.py
+++ b/napari/_qt/layers/qt_image_base_layer.py
@@ -6,6 +6,7 @@ from qtpy.QtCore import Qt
 from qtpy.QtGui import QImage, QPixmap
 from qtpy.QtWidgets import QComboBox, QLabel, QSlider, QPushButton
 
+from ...utils.event import EmitterGroup
 from ..qt_range_slider import QHRangeSlider
 from ..qt_range_slider_popup import QRangeSliderPopup
 from ..utils import qt_signals_blocked
@@ -15,6 +16,9 @@ from .qt_base_layer import QtLayerControls
 class QtBaseImageControls(QtLayerControls):
     def __init__(self, layer):
         super().__init__(layer)
+
+        # initialize qt events
+        self.events = EmitterGroup()
 
         self.layer.events.colormap.connect(self._on_colormap_change)
         self.layer.events.gamma.connect(self.gamma_slider_update)
--- a/napari/_qt/qt_controls.py
+++ b/napari/_qt/qt_controls.py
@@ -14,7 +14,6 @@ class QtControls(QStackedWidget):
         self.addWidget(self.empty_widget)
         self._display(None)
 
-        self.viewer.layers.events.added.connect(self._add)
         self.viewer.layers.events.removed.connect(self._remove)
         self.viewer.events.active_layer.connect(self._display)
 
@@ -49,6 +48,7 @@ class QtControls(QStackedWidget):
         controls = create_qt_controls(layer)
         self.addWidget(controls)
         self.widgets[layer] = controls
+        return controls
 
     def _remove(self, event):
         """Remove the controls target layer from the list of control widgets.
--- a/napari/_qt/qt_viewer.py
+++ b/napari/_qt/qt_viewer.py
@@ -32,6 +32,9 @@ from .qt_viewer_dock_widget import QtViewerDockWidget
 from .qt_about_keybindings import QtAboutKeybindings
 from .._vispy import create_vispy_visual
 
+from .._controllers import layer_to_controller
+from .layers import create_qt_controls
+
 
 class QtViewer(QSplitter):
     """Qt view for the napari Viewer model.
@@ -136,6 +139,9 @@ class QtViewer(QSplitter):
         # This dictionary holds the corresponding vispy visual for each layer
         self.layer_to_visual = {}
 
+        # This dict holds coresponding controllers for each layer
+        self.layer_controllers = {}
+
         if self.console.shell is not None:
             self.viewerButtons.consoleButton.clicked.connect(
                 lambda: self.toggle_console()
@@ -218,7 +224,6 @@ class QtViewer(QSplitter):
 
     def _add_layer(self, event):
         """When a layer is added, set its parent and order.
-
         Parameters
         ----------
         event : qtpy.QtCore.QEvent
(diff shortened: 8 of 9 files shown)


napari/layers/image/image.py
```diff
@@ -335,10 +335,12 @@ def interpolation(self):
 
     @interpolation.setter
```
sofroniewn, line 336: I find it counter intuitive that the `interpolation.setter` doesn't actually set the `interpolation` property and only emits an event. This means that if the controller isn't present then setting the property will have no effect, which makes it hard for me to imagine how to re-use the napari models without a controller - say in a "headless mode" if one wanted, or does the controller also get used in "headless mode"? I also find the existence of an explicit `set_` method in addition to being able to directly set the property confusing.

Could this instead be implemented as
```python
@interpolation.setter
def interpolation(self, interpolation):
    if isinstance(interpolation, str):
            interpolation = Interpolation(interpolation)
    if self._interpolation == interpolation:
           return
    self._interpolation = interpolation
    self.events.interpolation(interpolation=interpolation)
```
Here if you try and set the property to something it already is then nothing happens, otherwise it does the setting and then emits the event. Does this conflict with the MVC approach?

sofroniewn (commented):
Thanks so much for the great diagrams and example code @shanaxel42! I can see how the MVC diagram is cleaner at the diagram level then what we have now, and the centralization of the event handling makes sense to me, but I have some questions about some of the consequences of the refactor and what it will look like for the rest of the code base.

- After the refactor, if we wanted to swap out vispy / Qt or have multiple rendering options / backends would the we need new controllers for each one? And would we have more duplicate code after the refactor relative to if we tried to do this now.

- If we wanted to use the model files without any Qt / vispy imports would we still need to import the controller or do something special inside it - i.e. what does a "headless-mode" look like? (I left a more detailed question about that in the model code)

- Where do things like mouse functions / keybindings live - say the code in the labels layer around using the paintbrush, including both painting related functions in the model file and the mouse callbacks?

- For the viewer, where does code like `add_image` and `add_labels` live and how does it get called? Here I think it is really important that all that code can be run and interacted with without any vispy or Qt imports.

I'd also love @tlambert03, @jni and particularly @royerloic, who spent a lot of time thinking about our model and view refactor, to take their time to understand the proposed refactor in detail, both for this chunk of code and also how it will propagate to the rest of the code base, and then weigh in.

napari/layers/image/image.py
shanaxel42: Yea I agree this was the one thing I thought was really weird/isn't great. Ideally the setter method would _only_ be doing the setting and not emit the event, and then a user would use the `set_interpolation()` public method when they wanted to make changes and that would emit the event. But it seems like you guys want to maintain the ability for a user to set attributes directly/not through an api. The issue with your idea is that when a change get's made from the gui, this method will get called, update the data then re-emit the event. We don't want this since in that scenario a change event has already been emitted from the gui and re emitting causes a loop. So I couldn't find my way around have two methods, one that just sets the data and one that emits the event. One other way you could get around this is by setting `self._interpolation == interpolation` directly in the controller, but I thought that was a little hacky. Open to ideas though I agree it's weird.

shanaxel42:
> * After the refactor, if we wanted to swap out vispy / Qt or have multiple rendering options / backends would the we need new controllers for each one? And would we have more duplicate code after the refactor relative to if we tried to do this now.

No, we wouldn't need new controllers. We'd just need to edit the existing ones to modify the new viewers or backends. I actually think this get's easier after this refactor because the changes you'd have to make are all in one place rather then setting up event connections in each new view. 

> * If we wanted to use the model files without any Qt / vispy imports would we still need to import the controller or do something special inside it - i.e. what does a "headless-mode" look like? (I left a more detailed question about that in the model code)

The models shouldn't import the controllers. The controllers import the model. So everything there should work as it does now. 

> * Where do things like mouse functions / keybindings live - say the code in the labels layer around using the paintbrush, including both painting related functions in the model file and the mouse callbacks?

I think this should also live in the controller. So say a mouse event will get emitted from vispy_image.py, the image_controller will be listening to that event and process accordingly. 

> * For the viewer, where does code like `add_image` and `add_labels` live and how does it get called? Here I think it is really important that all that code can be run and interacted with without any vispy or Qt imports.

Nothing should get changes in `add_image` or `add_labels`. In my mind these are the methods that instatiate the data layer, so they get called, and then the separate methods get called for creating the vispy and qt components (but all three are separate), and then the controller gets created with access to all three. (See changes to qt_viewer)
 
> I'd also love @tlambert03, @jni and particularly @royerloic, who spent a lot of time thinking about our model and view refactor, to take their time to understand the proposed refactor in detail, both for this chunk of code and also how it will propagate to the rest of the code base, and then weigh in.
 
Yes please give me all the feedback!

napari/layers/image/image.py
kne42: This PR is super great! I love how clear you made everything :)

> One other way you could get around this is by setting `self._interpolation == interpolation` directly in the controller, but I thought that was a little hacky.

Building on our offline chat, what if we treated the front-facing API as another type of view? Then we could easily separate/differentiate between the setting of the actual data (model) and the triggering of event updates.

Thoughts? @shanaxel42 @sofroniewn @jni @tlambert03

shanaxel42: Nick and I brainstormed a new option. There's still a separate method called `_set_interpolation` that actually updates the model attribute, but it get's called from the setter method. We also added a check to prevent unnecessary extra calls

napari/_qt/layers/qt_image_layer.py
```diff
@@ -8,7 +9,8 @@ class QtImageControls(QtBaseImageControls):
     def __init__(self, layer):
         super().__init__(layer)
 
-        self.layer.events.interpolation.connect(self._on_interpolation_change)
```
shanaxel42, line 11: Eventually all these connection lines would get deleted and added to the controller

ttung:
Generally, this is a great direction!  Here's a few small concerns:

1. How does error handling work in this universe?  One could presume that settings driven by UI may need to be validated against other state, and need to be cancelled if the request is not valid.
2. The way attributes are set should be consistent between `qt_image_layer.py` and `image.py`.  It appears that setting the attribute in `image.py` updates the attribute and broadcasts the event and then receives yet another set attribute request from the controller.  `qt_image_layer.py` broadcasts the event, and then receives a set attribute from the controller.  I am fairly sure that the path for `qt_image_layer.py` is the better one because error handling could plausibly be done by the controller and cancelled.
3. I like @kne42's comment about the public python object being exposed as another "view".
4. There is a lot of boiler plate here.  Ideally, we should be codegen'ing a lot of it.

napari/layers/image/image.py
shanaxel42: Reverted to original implementation and re-openeing this convo after @ttung's comments, @ttung do you want to elaborate on your concerns?

ttung: Sure.  There should be one source of truth.  With this approach, there are multiple sources of truth, and they can diverge in a way that never converges.  The original approach allows for error handling inside the controller before state changes are made.  If there are multiple state changes where the order of the operations matter, this approach does not guarantee that you have the same outcome.  It adds confusion because it diverges unnecessarily.

sofroniewn: The downside I see with this revert is that now a person reading our model file code doesn't understand how writing `layer.interpolation = 'new_value'` actually leads to that value being changed without looking into our event system and other files, which I find makes our model file code much less readable.

Additionally with this approach our model code can no longer be run without the controller which does our vispy and Qt imports which makes understanding what "headless mode" much harder for me too. I think a top level requirement is that the model code can continue to run and function without any of the vispy and Qt code and that the separation be as clean as possible.

Note that I remain excited about the rest of the refactor.

sofroniewn: @shanaxel42 @ttung - I have a potential solutions. How about we put the setting call back inside the setter in the model, but then in the controller look at the source of the event and if the event comes from the model not call the setter again. This will prevent the double call that @ttung (and I!) didn't like, but allow our model to function independently. The changes would be as follows:

In the model
```python
def interpolation(self, interpolation):
    self._set_interpolation(self, interpolation)
    self.events.interpolation(interpolation=interpolation)
```

and in the controller
```python
def on_interpolation_change(self, event=None):
"""
Process changes when the interpolation attribute is changed from any interface
"""
value = event.interpolation
# update layer if event did not come from layer
if not isinstance(event.source, Layer):
    self.layer._set_interpolation(value)
# update qt
self.qt_controls.set_interpolation(value)
# update vispy
self.vispy_layer.set_interpolation(value)
```

We have to check that we're making the right choice passing the string vs enum around inside the `event.interpolation` property and we're always using the same one.

Thoughts?

tlambert03: Hi all, just getting up to speed here.  Definitely agree this is a great direction @shanaxel42!
from what I've read here, I think I mostly agree with @ttung's position (that the controller should be the one source of truth), even if it appears at first to impose some circuitous setters.  I agree with @sofroniewn that whatever direction we go in should not create tighter coupling between our model and vispy/qt code... but I don't necessarily think it needs to go that way (but it would take some slight changes to the current controller design).  Many thoughts:

1. @sofroniewn, I think your last proposal here has the multiple-source-of-truth problem that @ttung was mentioning:  if some error occurs in the controller method when trying to `set_interpolation` on the `qt_controls`, then the GUI and model states diverge (because the model has already been set).  I guess you might be saying at the end there that we would try real hard to prevent that by being extra careful with arguments... but I suspect errors are still inevitable.  Whereas if an error was caught in the controller, it could raise an exception before updating the model.

2. I'm not actually sure the controller would/should need to be so tightly coupled to Qt & vispy.  Currently in this PR, it's very tightly coupled: since the actual controller **instances** reside in `QtViewer.layer_controllers` and requires three instances (`Layer`, `QtControls`, and `VispyLayer`) to be instantiated.  As @shanaxel42 said, we could refactor that later to some other backend, but that would be a big refactor, due to the tight coupling.  Instead, would it be possible to have `ControllerBase` only take a `layer` instance, and live somewhere on its own (i.e. some `Controllers` object that maps layer instances to controller instances, or even on the layer itself) and have `QtImageControls` and `VispyImageLayer` instances (and indeed, _any_ other future listener object) register _themselves_ with the controller?  With that design, if the qt part or the vispy part disappear, the controller doesn't really care... it's just receiving and sending napari events out into the void for whoever is listening, and the listeners are responsible for hooking up events to their own set methods (not the controller)... In a headless mode, this _would_ still mean that the `@interpolation.setter` has to go through the controller to set the model, but it doesn't mean that vispy or Qt need to be involved.  Yes, that does add a little complexity for someone reading our model code trying to understand the flow of events, but I think to some extent we either need to decide if napari is an "event-driven ecosystem" or not.  If so, I like how this PR creates our own master controller (rather than piggy-backing on a mishmash of Qt/Vispy event loops).  Also, if we _really_ wanted to make the models have the option of standing alone, then the model setter methods could have logic that says "if I have a controller, set the value through the controller... if not, just set it directly".

(I'm starting to wonder whether I basically just went back to the old model there?...  aggh. I don't know.  But I do think that ultimately any view-like thing (vispy, qt) should be able to just disappear without changing the controller code).

3. lastly, given how often we talk about making sure we can operate independently of Qt or Vispy (which I wholeheartedly agree with!), I'm just curious whether anyone has actually looked into what it would take to refactor the code to even import without `Qt` or `vispy` present?  (Say, if we wanted to have a headless pipeline that uses io-plugins to create some layers, does some shit to them, and writes something to disk).

sofroniewn: @tlambert03 I have now added a headless mode in #1012, we can continue further discussion of that there.

Your other points are well taken - for (1) that was indeed the tradeoff. I'll let others weigh in on (2), but I'm not sure the increased complexity is worth it. Maybe though. Thanks for the review!!

jni: I'll caveat this by saying that I have not taken the time to get up to speed yet, just want to comment on @tlambert03's comments:

> (I'm starting to wonder whether I basically just went back to the old model there?... aggh. I don't know. But I do think that ultimately any view-like thing (vispy, qt) should be able to just disappear without changing the controller code).

:point_up: this thing! LOL As I was reading 2 I really felt like it was describing the current state of things. To me it feels really important that the single source of truth is not the controller, but the model. That's why it's a model. I'm happy adding an extra object to ensure (1) that everything is in sync, and (2) that setters and other methods are only ever called once. I think that is a worthwhile goal, and certainly a well-founded one since once upon a time Talley fixed our slice setting code being called SIX times. :joy: (@shanaxel42 knows that I didn't at first grok the point of this refactor, but I get it now! :grinning:) But to me it makes the most sense for the model to be the single source of truth, not the controller. Otherwise names are meaningless!

The "order of operations" problem is a bit thornier, but I think (?) it can be solved by having a reasonable API in the model class. In short, if the [voltsock has to be activated before the plectogrill](https://www.polygon.com/2013/3/18/4092094/shouting-crazy-science-fiction-terms-with-spaceteam), don't offer two methods to do that, just offer the one that sets both, and make the rest private.

> I'm just curious whether anyone has actually looked into what it would take to refactor the code to even import without Qt or vispy present?

It turns out that lazy import is finally easy in Python 3.7 ([PEP-562](https://www.python.org/dev/peps/pep-0562/)), so, actually, it would be trivial these days, by lazy-importing the Qt stuff in `napari/__init__.py`. Then you can do `from napari import ViewerModel as Viewer` and be off to the races. (I think!) And, by the way, @sofroniewn, I wonder whether that might be a better approach in the long term than the one you took in #1012... 

I'll go over this in more detail in the coming days, just wanted to voice my current thinking from the first-principles perspective, as well as respond to some of @tlambert03's specific comments...

Thanks everyone for the flurry of activity so far!

sofroniewn: If the single source of truth is an issue - how about this?

```python
def on_interpolation_change(self, event=None):
"""
Process changes when the interpolation attribute is changed from any interface
"""

# update layer if event did not come from layer
if not isinstance(event.source, Layer):
    self.layer._set_interpolation(event.interpolation)

# update qt from layer
self.qt_controls.set_interpolation(self.layer.interpolation)
# update vispy from layer
self.vispy_layer.set_interpolation(self.layer.interpolation)
```

ttung: A fairly common approach is to:
1. stick the data in the model
2. mutation operations are managed by the controller.  only the controller is supposed to write to the model.

The controller does not necessarily have to bind to vispy / qt.  One could build the following:

1. (the above two points)
2. standardized UI class that the controller interacts with.
2a. a vispy/QT implementation of that UI class.
2b. a console implementation of that UI class that has the nice attribute getters/setters.

Headless mode would just be 1 + 2b.  Normal mode would be 1 + 2a + 2b.

EDIT: Also, don't expose the controller or the data to the user.

shanaxel42: Thanks for all the feedback guys!! 

> But to me it makes the most sense for the model to be the single source of truth, not the controller. 

I'm with Juan here. I think the model should be the source of truth for the data. The controller should just be about updating the modal and views in a central location. 

>Currently in this PR, it's very tightly coupled: since the actual controller instances reside in QtViewer.layer_controllers and requires three instances (Layer, QtControls, and VispyLayer) to be instantiated. 

I get the concern here. One thing we could do is create really simple base classes for the QTControls and VispyLayer components that have the getters and setters and pass those into the controllers. That way if we wanted to change out the libraries down the line we'd just need to make sure they implement the methods in the base class. I personally don't like the model where the controller just emits events and the views update themselves because, as Talley said, that's essentially whats happening now and makes it hard to track changes.

> EDIT: Also, don't expose the controller or the data to the user.

I'm no sure if this is possible. Seems like the ability for a user to interact directly with the data layer is a requirement @sofroniewn ? 

I'll keep iterating on this idea but so far it seems the main concerns are:

- Wanting to make it easy to swap view libraries at some point down the line. 
- Making sure headless mode makes sense to a user
- Making sure the refactor doesn't allow for situations in which data is out of sync between the viewers and the modal

Let me know if I'm missing any main themes!

tlambert03: > I personally don't like the model where the controller just emits events and the views update themselves because that's essentially what's happening now and makes it hard to track changes.

I was thinking about this after my first post, and I think they each have merits.  I agree, this PR makes it extremely easy for someone _reading_ the code to see what's happening and track changes, because it takes all of the events and their connections, and physically puts them in one spot in the codebase.  But in doing so, it sort of "formalizes" all of those third-party players.

Whereas in the current model (and I _think_ generally in projects using Qt-style signals and slots), those connections and methods are necessarily spread across multiple objects because that makes it much easier for one or more of those objects to simply _disappear_ without breaking anything.  That is, there's no problem sending a signal out even if no one is listening.

With the controller as it currently is in this PR, we're saying "here's who _has_ to be listening and here's what they will do".

jni: Hi everyone!

@shanaxel42 I saw #1012 roll in, but thought I'd comment here for continuity. I wanted to add to:

> I'll keep iterating on this idea but so far it seems the main concerns are:

one more: the headless API should be Pythonic. Things like `set_interpolation` should not be visible in our public API. `image_layer_instance.interpolation = 'nearest'` should not just be possible, it should be the recommended way of doing things.

@shanaxel42 also, I couldn't see above whether you responded to @sofroniewn's suggestion of `if event.source == model: pass`?

@ttung 

> EDIT: Also, don't expose the controller or the data to the user.

How does the user set the interpolation mode from the REPL, then?

ttung: > How does the user set the interpolation mode from the REPL, then?

2b. a console implementation of that UI class that has the nice attribute getters/setters.

jni: > a console implementation of that UI class that has the nice attribute getters/setters.

Ah, right. So, in that case, if I have a UI displaying stuff and I want to mess with the parameters of the UI from that console, I have (a) the controller, (b) the data, (c) the UI instance, and (d) the console instance. I guess `viewer.add_image` returns the `console` view instance, but it creates all four objects behind the scenes. Is my interpretation correct?

sofroniewn: @jni #1020 is an alternate implementation that allows the model and a headless controller to be used together without vispy and qt - it is quite elegant once you see how it works

ttung: > Ah, right. So, in that case, if I have a UI displaying stuff and I want to mess with the parameters of the UI from that console, I have (a) the controller, (b) the data, (c) the UI instance, and (d) the console instance. I guess viewer.add_image returns the console view instance, but it creates all four objects behind the scenes. Is my interpretation correct?

Yes, that's correct.

royerloic:
@all we should definitely move this discussion to #1020 as it is imho a superior solution...

shanaxel42:
Gonna close this out now in favor of the new direction in #1020!

Issue #992, opened 2020-02-23
Ideas for Settings & Preferences


tlambert03:
## 🚀A list of things that might benefit from persistence
Following up on this https://github.com/napari/napari/pull/834#issuecomment-590112950, this issue is a place list ideas for things that might make good settings/preferences (once we have a persistence layer in place).  Feel free to edit this comment directly, or start your own lists.

- [x] theme (superseding #943)
- [ ] monitor DPI (#820)
- [x] call order of plugins (after #937)
- [x] window position/geometry
- [ ] default colormaps or color combinations (as discussed in #619)
- [ ] magic-naming for layer (off by default in #1008)
- [ ] PySide2 vs PyQt5 preference (if both installed)
- [x] highlight thickness for shapes / points layers
- [x] dask cache size maximum, and dask fusion settings (https://github.com/napari/napari/pull/1173#issuecomment-620231964)
- [ ] opt in for telemetry
- [ ] (need to know if it's the first launch)
- [ ] version updates (has user been asked)

GenevieveBuckley:
Font size in the integrated ipython console.

I would really like to be able to change this, I keep going to use the integrated ipython console & then remembering this is the reason I don't like using it. And yes, I do have terrible eyesight.

jni:
I just came across [omegaconf](https://omegaconf.readthedocs.io/en/2.0_branch/). Not that we need any more choices for napari config management, but I think part of the reason we haven't got this done is that nothing felt "satisfying", and OmegaConf kinda does! See the "creating" part of the user guide, lots of nice stuff in there.

It *is* a young project, which might be reason enough to avoid it, but I still thought it was worth considering.

tlambert03:
looks nice!  just went through their binder tutorial which is great.
For me, i feel like the biggest conflict wasn't necessarily the implementation or the API (I do like what I see in omegaconf, but I wasn't finding the dask approach particularly lacking either)... it was more the conflict between the concept of app settings and library settings, as described in https://github.com/napari/napari/pull/1183#issuecomment-694243607.  I feel like if we just said "sure, let's have _some_ sort of private persistence layer that the app can write to, and the user doesn't need to touch" ... we'd be off and running.  I don't think it's impossible to do both, but it's definitely the harder challenge and we're biting off a lot.

Carreau:
Hi all, from yesterday productive discussion about setting a couple of my mental notes

- Config can also be useful for remembering custom keyboard shortcuts.
[...]

PR #1008, opened 2020-02-28
add magic name guessing

kne42:
# Description
basically i had caffeine today and i cant sleep so i decided to try messing around with `skimage` + `napari` and noticed how amazing the walrus operator is so I added optional magic name guessing based on variable name because I'm lazy

e.g.
```python
viewer.add_labels(cells := some_func())
> 'cells' labels layer

selection = cells > 2
viewer.add_labels(selection)
> 'selection' labels layer
```

this is optional so you need to set `MAGICNAME` to a nun-null value in your environment (eg. `export MAGICNAME=1` or `import os; os.environ['MAGICNAME'] = '1'`)

example workflow:
Screen Shot 2020-02-28 at 8 49 26 AM

## Type of change

# References

# How has this been tested?

## Final checklist:


napari/view_layers.py
```diff
@@ -1,4 +1,11 @@
 from .viewer import Viewer
+from .utils.magic import magic_name
+
+
+def _guess_name(name, data):
```
sofroniewn: if `_guess_name` is getting called in the `AddLayersMixin` do we need a copy here?

napari/utils/magic.py
```diff
@@ -0,0 +1,41 @@
+"""Utilities that provide magic functionality.
```
sofroniewn: Can this code live inside `utils/naming.py` instead of its own file?

napari/view_layers.py
kne42: yes, as it will guess based on the calling function, which means that if we didn’t do this in `view_*`, it would always default to `'data'`

sofroniewn: ah, hmm - it doesn't seem so desirable to have the duplicate code. Maybe at least we can move `guess_name` to our utils too?

kne42: yeah I was thinking that too. also note to self: need to fix the assignment tests

kne42:
@sofroniewn this is ready for re-review

napari/view_layers.py
```diff
@@ -115,6 +116,7 @@ def view_image(
     viewer : :class:`napari.Viewer`
         The newly-created viewer.
     """
+    name = _guess_name(name, data)
```
jni: I dislike this line everywhere, because one cannot guess (heh) that guess-name is *not* guessing anything if name is None. My suggestion to remove the checking from the function and leave it to the caller:

```python
if name is None:
    name = _find_variable_name(data)
```

This would make these lines much clearer.

napari/utils/naming.py
```diff
@@ -42,3 +46,60 @@ def inc_name_count(name):
...
+        raise ValueError('cannot have a level lower than 1')
+
+    if not os.getenv('MAGICNAME'):
+        return
+
+    level += 1
+    frame = inspect.currentframe()
+
+    for i in range(level):
+        frame = frame.f_back
```
jni, line 69: It might be worth trying to make the function recursive (search *each* frame *up to* `level`), so that you don't have to `guess_name` both in `view_image` and `add_image`. This would pick the lowest frame for which the variable is found as long as the name is not `data`. *OR*, it could be the lowest frame that isn't inside the napari library...? Making this function only be called in `AddLayersMixin` would be a massive win for the codebase, and I think the problem is probably solvable...! What do you think?

jni (changes requested):
@kne42 thanks! I love this functionality! I'd like us to spend a bit more time thinking about the implementation to see if we can avoid the many calls to `_guess_name` scattered around the codebase currently. I've made a suggestion for how to do this, but I don't know yet if the suggestion is dumb. =P I've also made a more minor comment about code style.

kne42:
@sofroniewn @jni this is ready for re-review. I have made it so that it will make its guess at the lowest frame not under the napari root directory

napari/utils/_tests/test_naming.py
```diff
@@ -33,3 +43,76 @@ def test_inc_name_count():
...
+
+def test_basic():
+    """Check that name is guessed correctly."""
+    assert magic_name(42) is None
+
+    z = 5
+    assert magic_name(z) == 'z'
+
+    if walrus:
+        assert eval_with_filename("magic_name(y:='SPAM')") == 'y'
```
jni, line 68: This is clever! =D

jni (approved):
Wow. Very cool. =D

I'm cool with this, I'll let @sofroniewn or @tlambert03 do the merging honours. =)

sofroniewn:
This looks great @kne42. I was a little tripped up that it is off by default - I didn't see any change at first, but then realized I had to set the environment variable, but this is something that can be easily adjusted with #834 and is now noted in #992. I will let sit for a bit in case @tlambert03 or others want to weigh in too and will then merge. Thanks!!

napari/utils/naming.py
```diff
@@ -42,3 +45,33 @@ def inc_name_count(name):
         Numbered name incremented by ``1``.
     """
     return numbered_patt.sub(_inc_name_count_sub, name, count=1)
+
+
+def magic_name(value, *, path_prefix):
```
tlambert03, line 50: ❤️

tlambert03 (approved):

PR #1010, opened 2020-02-29
Refactor labels layer mouse bindings

sofroniewn:
# Description
This PR is a step towards closing out #694 first just refactoring the mouse functions for `Labels` layer. If successful we can merge, and then I will follow up with a PR that refactors the mouse functions for the `Shapes` and `Points` layer. I'll note that all it is really doing is pulling out the `on_mouse_*` code from the model files and making a new `mouse_functions.py` file for each layer. (I can name this `labels_mouse_functions.py` if @jni prefers) (In which case I should probably make a standalone PR to rename our `keybindings.py` to `*_keybindings.py` too).

Right now there is one problem though and tests are failing because of it. In order to get this to work I had to move `mouse_move_callbacks` and `mouse_drag_callbacks` (which are inside the base layer) to be at the class level instead of the instance level, because otherwise I couldn't make the `@Labels.mouse_drag_callbacks.append` call inside `mouse_functions.py`. Any advice from @kne42 on how to keep this at the instance level instead of the class level would be appreciated.

## Type of change


napari/_qt/qt_viewer.py
```diff
@@ -460,7 +460,7 @@ def on_mouse_press(self, event):
 
         layer = self.viewer.active_layer
         if layer is not None:
-            # Line bellow needed until layer mouse callbacks are refactored
+            # Line bellow needed until layer mouse callbacks are refactored!
```
sofroniewn: Just wanted to highlight here that at the end of the refactor those lines can be removed

napari/layers/base/base.py
```diff
@@ -181,11 +181,14 @@ def __init__(
...
         self.dims.events.axis.connect(self.refresh)
 
-        self.mouse_move_callbacks = []
-        self.mouse_drag_callbacks = []
+        # self.mouse_move_callbacks = []
+        # self.mouse_drag_callbacks = []
         self._persisted_mouse_event = {}
         self._mouse_drag_gen = {}
 
+    mouse_move_callbacks = []  # Note this is now at class level
```
sofroniewn: Here is the problem code that I need help fixing

ttung: If you want the behavior we discussion in person (i.e., decorator to set up the default callbacks but with exceptions on a per-layer basis), I think you're stuck with this strategy:

1. Create class callbacks like what you have.
2. Create a list of instance additions to the class callbacks.
3. Create a list of instance removals from the class callbacks.

Yeah, it's ugly.  :(

kne42: @sofroniewn I think we should hold off on class vs instance callbacks for now. How I did it in [my branch](https://github.com/kne42/napari/tree/refactor-internal-mouse-bindings) was by keeping the callbacks list at the instance level and instead of using decorators to register them as defaults, inside the `mode` setter you choose which one you want.

[example](https://github.com/kne42/napari/blob/339b1271aa8591913ebdb035293931a05c2ce5e1/napari/layers/points/points.py#L1094-L1118):
```python
        old_mode = self._mode


        if old_mode == Mode.ADD:
            self.mouse_drag_callbacks.remove(add)
        elif old_mode == Mode.SELECT:
            self.mouse_move_callbacks.remove(highlight)
            self.mouse_drag_callbacks.remove(select)


        if mode == Mode.ADD:
            self.cursor = 'pointing'
            self.interactive = False
            self.help = 'hold <space> to pan/zoom'
            self.mouse_drag_callbacks.append(add)
        elif mode == Mode.SELECT:
            self.cursor = 'standard'
            self.interactive = False
            self.help = 'hold <space> to pan/zoom'
            self.mouse_move_callbacks.append(highlight)
            self.mouse_drag_callbacks.append(select)
        elif mode == Mode.PAN_ZOOM:
            self.cursor = 'standard'
            self.interactive = True
            self.help = ''
        else:
            raise ValueError("Mode not recognized")
```

sofroniewn: Thanks for the guidance @ttung and the example of how you tried it earlier @kne42. As you mention, in your approach you put the binding inside the mode setter. I was trying to pull the binding out into its own file, but that is proving tricky. 

@shanaxel42 what do you make of this / how do you recommend we proceed given the refactor planned in #1020. The goal of this series of PRs would have been to pull the mouse stuff out a little, but maybe it's not worth it. It could be very useful for #1020 to think about what adding an example of with a mouse function / keybindings callback (maybe from the Labels layer) would look like and then determine if this refactor helps or not, or what it should look like.

sofroniewn: I have now taken the approach that @kne42 implemented in her branch. This PR is now almost identical to that one, but just for the labels layer, and with some additional naming changes.

sofroniewn:
This PR is now ready for re-review. It just changes the Labels layer, using the new generator syntax and the approach taken by @kne42 here https://github.com/kne42/napari/tree/refactor-internal-mouse-bindings which I think is very nice. It also makes some minor naming convention choices.

Once this is merged I'll make a follow-up PR that does the same thing for the `Points` layer, then `Shapes` etc.

napari/layers/labels/_tests/test_labels_mouse_bindings.py
```diff
@@ -0,0 +1,38 @@
...
+
+    # Simulate drag
+    event = ReadOnlyWrapper(Event(type='mouse_move', is_dragging=True))
+    mouse_move_callbacks(layer, event)
+
+    # Simulate release
+    event = ReadOnlyWrapper(Event(type='mouse_release', is_dragging=False))
+    mouse_release_callbacks(layer, event)
+
+    assert np.unique(layer.data[:5, :5]) == 3
```
kevinyamauchi, line 51: @sofroniewn , can you please explain why `np.unique(layer.data[:5, :5]) == 3`?

This is my understanding from [the paint callback](https://github.com/napari/napari/blob/6313ffc2dcc2408fb56eb5b132399af34c126708/napari/layers/labels/_labels_mouse_bindings.py#L21)

1. (line 26) On the first click, label 3 is painted on `layer.coordinates` (with brush size 10) and `layer._last_cursor_coord` is set to `layer.coordinates` (19, 19)
2. (line 30) The dragging event starts. `layer.coordinates` isn't updated, so no new additional pixels are painted
3. (line 34) The dragging and clicking ends.

If that is correct, it seems like `np.unique(layer.data[-5:, -5:]) == 3` should be true, but I can't tell how the pixels at `[:5, :5]` are painted.

napari/layers/labels/_labels_mouse_bindings.py
```diff
@@ -0,0 +1,61 @@
...
+    # on press
+    layer.selected_label = layer._value or 0
+    yield
+
+    # on move
+    while event.type == 'mouse_move':
+        yield
+
+    # on release
+    layer._last_cursor_coord = None
```
kne42: this is something that I wanted to ask you on the original branch. I converted it to this because I wanted to avoid running into bugs from trying to clean the code up so I created the exact equivalent of what it was before

from my guess, the code in the on release section is specific to painting and can be removed from here (along with the on move section)

napari/layers/labels/_labels_mouse_bindings.py
```diff
@@ -0,0 +1,61 @@
...
+    # on press
+    layer.fill(layer.coordinates, layer._value, layer.selected_label)
+    yield
+
+    # on move
+    while event.type == 'mouse_move':
+        yield
+
+    # on release
+    layer._last_cursor_coord = None
```
kne42: same as above

napari/layers/labels/_labels_mouse_bindings.py
```diff
@@ -0,0 +1,61 @@
...
+    layer._block_saving = False
+
+
+def paint(layer, event):
+    """Paint with the currently selected label."""
+    # on press
+    layer._save_history()
+    layer._block_saving = True
+    layer.paint(layer.coordinates, layer.selected_label)
+    layer._last_cursor_coord = copy(layer.coordinates)
```
kne42: which of these private variables assigned to the layer can be made local to this function instead? also I would generally go for the `obj.copy()` method over `copy.copy(obj)` in most situations

napari/layers/labels/_tests/test_labels_mouse_bindings.py
sofroniewn: `layer._last_cursor_coord = (0, 0)` so the painting goes from `(0, 0)` to `(19, 19)`. I will add a note explaining

napari/layers/labels/_labels_mouse_bindings.py
sofroniewn: great catch, yes this can be dropped

napari/layers/labels/_labels_mouse_bindings.py
sofroniewn: yes, dropped, thanks!

napari/layers/labels/_labels_mouse_bindings.py
sofroniewn: great observation - yes, this could be cleaned up - done now

sofroniewn:
I've added two more tests and put some more notes in one of them to answer the question from @kevinyamauchi. I've also made a bunch of improvements and simplifications based on ideas from @kne42.

napari/layers/labels/_tests/test_labels_mouse_bindings.py
```diff
@@ -0,0 +1,101 @@
...
+
+    # Simulate drag
+    event = ReadOnlyWrapper(Event(type='mouse_move', is_dragging=True))
+    mouse_move_callbacks(layer, event)
+
+    # Simulate release
+    event = ReadOnlyWrapper(Event(type='mouse_release', is_dragging=False))
+    mouse_release_callbacks(layer, event)
+
+    # Painting goes from (0, 0) to (19, 19) with a brush size of 10, chaning
```
kevinyamauchi:
```suggestion
    # Painting goes from (0, 0) to (19, 19) with a brush size of 10, changing
```

napari/layers/labels/_tests/test_labels_mouse_bindings.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+def test_paint():
+    """Test painting labels with different brush sizes."""
+    data = np.ones((20, 20))
+    layer = Labels(data)
+    layer.brush_size = 10
+    layer.mode = 'paint'
+    layer.selected_label = 3
+    layer._last_cursor_coord = (0, 0)
```
kevinyamauchi: Should this be deleted now that we are using `last_cursor_coord` that is local to `paint()`?

napari/layers/labels/_tests/test_labels_mouse_bindings.py
kevinyamauchi: Sorry to dwell on this - I am mostly curious for my own understanding.

I see that `layer._last_cursor_coord` is set to `(0, 0)` on [line 19](https://github.com/napari/napari/blob/1d5210e3b50a2746f432695976a02fa59b4e7250/napari/layers/labels/_tests/test_labels_mouse_bindings.py#L19). However, in the `paint()` callback, it looks like when the mouse is clicked, [the pixels within the brush radius of `layer.coordinates` (`(19, 19)`) are painted](https://github.com/napari/napari/blob/1d5210e3b50a2746f432695976a02fa59b4e7250/napari/layers/labels/_labels_mouse_bindings.py#L9) and then [`last_cursor_coord` (now local to `paint()`) is assigned `layer.coordinates`](https://github.com/napari/napari/blob/1d5210e3b50a2746f432695976a02fa59b4e7250/napari/layers/labels/_labels_mouse_bindings.py#L10) (i.e., `last_cursor_coord == (19, 19)`) before yielding. Therefore, to me, it seems like the dragging should start at (`(19, 19)`).

napari/layers/labels/_tests/test_labels_mouse_bindings.py
sofroniewn: fixed

napari/layers/labels/_tests/test_labels_mouse_bindings.py
sofroniewn: yes, dropped

sofroniewn:
@kevinyamauchi the drag test has now been fixed, it was a little funky before - and I now added two explicit sets of the `layer.position` variable which make it clear what is happening.

napari/layers/labels/_tests/test_labels_mouse_bindings.py
sofroniewn: Now that we dropped `_last_cursor_coord` it's a little weird that this worked, it was because not all the right updates were being called. I've fixed the test to have two explicit calls to `layer.position` (which is something like `worldslice coordinates` in our new (but yet to be implemented) terminology) which triggers the appropriate coordinate and value updates behind the scene. When I do the world coordinates PR we'll sort out the names of things like position and coordinates which is super confusing right now

kevinyamauchi: Thanks, @sofroniewn - that is helpful! I'm looking forward to world coordinates!

napari/layers/labels/_labels_mouse_bindings.py
```diff
@@ -0,0 +1,37 @@
...
+            last_cursor_coord, layer.coordinates, layer.brush_size
+        )
+        for c in interp_coord:
+            layer.paint(c, layer.selected_label, refresh=False)
+        layer.refresh()
+        last_cursor_coord = layer.coordinates
+        yield
+
+    # on release
+    layer._block_saving = False
```
ttung, line 25: This is fine now, but in the longer term, we will plausibly need to think about how to gracefully recover from errors.  As it is, this doesn't seem to have a way to ensure that the release code is called if something bombs earlier.

And since I had no idea how generators and `finally` clauses interact, I looked it up.  It looks like a [nightmare](https://amir.rachum.com/blog/2017/03/03/generator-cleanup/).

napari/layers/labels/_labels_mouse_bindings.py
```diff
@@ -0,0 +1,37 @@
...
+    """Paint with the currently selected label."""
+    # on press
+    layer._save_history()
+    layer._block_saving = True
+    layer.paint(layer.coordinates, layer.selected_label)
+    last_cursor_coord = layer.coordinates
+    yield
+
+    # on move
+    while event.type == 'mouse_move':
```
ttung, line 14: I found this **highly** confusing.  It looks like it's a vispy implementation detail that it modifies the original mouse press event to `mouse_move` and later to `mouse_release`.  I couldn't see anything in the documentation that details this behavior.  It makes me squeamish. :/

I would much much rather observe the current event than to depend on this quirk.  You could plausibly [send the new event back to the generator and save it](https://docs.python.org/3/reference/expressions.html#generator.send).

sofroniewn: Agreed, @kne42 and I were actually talking about this yesterday - she's planning a refactor to the mouse functions that will move us away from vispys event system into our own. Maybe she can talk about it at the dev meeting today, or can get your feedback on how she is thinking about it. As this will come later, I'd say we still merge as is and then fix during that refactor

napari/layers/labels/_labels_mouse_bindings.py
sofroniewn: Maybe this is also something to think about during the refactor @kne42 is thinking about that i mentioned in my other comment

napari/layers/labels/_labels_mouse_bindings.py
ttung: SGTM.

I would caution against using generators for this though.  It's not the most intuitive if you want developers to extend mouse clicking functionality, and error handling appears to be quite brittle.

napari/layers/labels/_labels_mouse_bindings.py
```diff
@@ -0,0 +1,37 @@
...
+        interp_coord = interpolate_coordinates(
+            last_cursor_coord, layer.coordinates, layer.brush_size
+        )
+        for c in interp_coord:
+            layer.paint(c, layer.selected_label, refresh=False)
+        layer.refresh()
```
jni, line 20: Is this fast enough now, or is this a performance regression?

napari/layers/labels/_tests/test_labels_mouse_bindings.py
```diff
@@ -0,0 +1,101 @@
...
+def test_paint():
+    """Test painting labels with different brush sizes."""
+    data = np.ones((20, 20))
+    layer = Labels(data)
+    layer.brush_size = 10
+    layer.mode = 'paint'
+    layer.selected_label = 3
+    layer.position = (0, 0)
+
+    Event = collections.namedtuple('Event', 'type is_dragging')
```
jni: I really don't understand what you're doing with this... Could you add a comment about how this is mocking a real event? `'type is_dragging'` is particularly confusing to me...

napari/layers/labels/_tests/test_labels_mouse_bindings.py
```diff
@@ -0,0 +1,101 @@
...
+    data[-5:, -5:] = 3
+    layer = Labels(data)
+    assert np.unique(layer.data[:5, :5]) == 2
+    assert np.unique(layer.data[-5:, -5:]) == 3
+    assert np.unique(layer.data[:5, -5:]) == 1
+    assert np.unique(layer.data[-5:, :5]) == 1
+
+    layer.mode = 'fill'
+
+    Event = collections.namedtuple('Event', 'type is_dragging')
```
jni: Can this be defined once in the whole file rather than in each test?

napari/layers/labels/_labels_mouse_bindings.py
```diff
@@ -0,0 +1,37 @@
+from ._labels_utils import interpolate_coordinates
+
+
+def paint(layer, event):
+    """Paint with the currently selected label."""
+    # on press
+    layer._save_history()
+    layer._block_saving = True
```
jni, line 8: This looks like it should be a context manager...

napari/layers/labels/labels.py
```diff
@@ -312,23 +307,33 @@ def mode(self, mode: Union[str, Mode]):
         if mode == self._mode:
             return
 
+        if self._mode == Mode.PICK:
+            self.mouse_drag_callbacks.remove(pick)
+        elif self._mode == Mode.PAINT:
+            self.mouse_drag_callbacks.remove(paint)
+        elif self._mode == Mode.FILL:
+            self.mouse_drag_callbacks.remove(fill)
```
jni, line 313: This feels brittle... Could we not have a stack-type API, so that we could just use `self.mouse_drag_callbacks.pop()` regardless of what is currently active?

napari/layers/labels/labels.py
```diff
@@ -312,23 +307,33 @@ def mode(self, mode: Union[str, Mode]):
...
             self.cursor_size = self.brush_size / self.scale_factor
             self.cursor = 'square'
             self.interactive = False
             self.help = 'hold <space> to pan/zoom, drag to paint a label'
+            self.mouse_drag_callbacks.append(paint)
         elif mode == Mode.FILL:
             self.cursor = 'cross'
             self.interactive = False
             self.help = 'hold <space> to pan/zoom, click to fill a label'
+            self.mouse_drag_callbacks.append(fill)
```
jni, line 334: Alternatively, a dictionary mapping modes to callbacks could be added. `self.mouse_drag_callbacks.append(callback_dict[mode])` at the very end, without all the repetitive calls in the `elif` cascade. You could also use `self.mouse_drag_callbacks.remove(callback_dict[self._mode])` at the top.

jni (changes requested):
@sofroniewn I've requested a few tweaks/clarifications. I'm also not sure how this fits in with @kne42's generalisations — I feel like the ChainMap stuff would simplify a lot of this code... Apologies if I've missed the discussions about this!

napari/layers/labels/_labels_mouse_bindings.py
sofroniewn: It is, this is not changed by this PR. It still needs to be faster, but that will come :-)

napari/layers/labels/_tests/test_labels_mouse_bindings.py
sofroniewn: yeah, that's the vispy event syntax. @kne42 proposal in #1048 will remove this, but it will be much easier to make those changes after this refactor than before. I'll add a comment.

napari/layers/labels/_tests/test_labels_mouse_bindings.py
sofroniewn: done

napari/layers/labels/_labels_mouse_bindings.py
sofroniewn: Agreed, that should be a follow-up PR, issue made #1065

napari/layers/labels/labels.py
sofroniewn: This will also change in #1048, but I think this matches our current API and approach and this can be fixed then

napari/layers/labels/labels.py
sofroniewn: This will change in #1048 too, so I'd probably rather just leave, but I can do this if needed

sofroniewn:
@jni I've made some minor fixes here, most of your comments are better addressed during the #1048 refactor / generalization proposal, which will be much easier to do on top of these PRs, so I'd rather get these in pretty much as they are (without breaking changes etc)

jni (approved):
:heart:

PR #1011, opened 2020-02-29
Fix pip install with older versions of pip

### LINKED ISSUE #1009: Problems with pip installation of the new napari version
Trying to pip install napari in anaconda 3 on a Windows produces an error.

## To Reproduce

1. From terminal, `pip install napari` gives me:

```
ERROR: Exception:
  File "C:\Users\lpetrucco\anaconda3\lib\site-packages\pip\_internal\cli\base_command.py", line 188, in main
    status = self.run(options, args)
  File "C:\Users\lpetrucco\anaconda3\lib\site-packages\pip\_internal\commands\install.py", line 407, in run
    use_user_site=options.use_user_site,
  File "C:\Users\lpetrucco\anaconda3\lib\site-packages\pip\_internal\req\__init__.py", line 58, in install_given_reqs
    **kwargs
  File "C:\Users\lpetrucco\anaconda3\lib\site-packages\pip\_internal\req\req_install.py", line 928, in install
    use_user_site=use_user_site, pycompile=pycompile,
  File "C:\Users\lpetrucco\anaconda3\lib\site-packages\pip\_internal\req\req_install.py", line 461, in move_wheel_files
    warn_script_location=warn_script_location,
  File "C:\Users\lpetrucco\anaconda3\lib\site-packages\pip\_internal\wheel.py", line 432, in move_wheel_files
    clobber(source, lib_dir, True)
  File "C:\Users\lpetrucco\anaconda3\lib\site-packages\pip\_internal\wheel.py", line 381, in clobber
    ', '.join(info_dir))
AssertionError: Multiple .dist-info directories: C:\Users\lpetrucco\anaconda3\Lib\site-packages\napari\plugins\_tests\fixtures\napari_working_plugin-0.1.0.dist-info, C:\Users\lpetrucco\anaconda3\Lib\site-packages\napari-0.2.12.dist-info
```

I have tried also in a completely new conda env and the same happens. I was able to install using conda-forge.

tlambert03:
# Description
This PR fixes #1009, in which pip installation of `napari-0.2.12` failed with older versions of pip.

A file-naming strategy I had used to test plugin discovery in #908 seems to have been incompatible with older versions of pip.  But it was testing a pretty fringe case (where a dist-info entry point points to another existing module), which apparently is so fringe that it used to actually raise an exception. I just renamed a couple files in the plugin test fixtures and modified one test slightly and it works again.

I tested locally by running `python setup.py sdist`, pip installing that `sdist` in a clean environment with pip versions as far back as `10.0.1`, and running `pytest --pyargs napari`

Sorry for the problems!

## Type of change

# How has this been tested?


sofroniewn (commented):
This looks good to me @tlambert03, but I'll let @jni and @kne42 actually approve

sofroniewn (commented):
This looks good to me, but @tlambert03 can you put in this PR exactly which naming strategy was causing problems so we don't make that mistake again. We also have to make it clear that people shouldn't use that naming strategy otherwise their plugins won't work right?

Maybe @jni can do the approve - or I can once I fully understand the problem & fix :-)

tlambert03:
The chances that someone else would have done this is next to nothing.  Basically, I created a module `napari_test_plugin` that was _in-itself_ a valid napari plugin via naming convention.  Then, to also test discovery of a working plugin via entry points, I created a mock dist-info package with an `entry_point` pointing to that other `napari_test_plugin` (rather than to it's _own_ module, which is how probably 100% of packages would be shipped).  Pip found it while installing `napari` and older versions of pip didn't like it.  So I created a new module `working_plugin.py`, and had the `entry_point` point to that instead.

I think the main challenge here is that we ship all of our tests with the release.  That's easy enough to do when you have a library that has easily generated test cases, but the goal here is to test plugin discovery with "real looking" mock plugin packages.  Apparently, pip also found those mock packages during installation and got confused due to the very unconventional entry point usage (which was never intended to be used for a real `pip` installation anyway).

sofroniewn (approved):
Ok, that makes sense to me! Thanks @tlambert03 I am approving. I will let this sit till tomorrow in case @jni wants to comment and will then merge.

PR #1012, opened 2020-02-29
Add headless mode

sofroniewn:
# Description
This PR follows on discussion in https://github.com/napari/napari/pull/1006#discussion_r386058302 by adding an explicit `headless` mode. It has an implementation that follows the `is_remote` option in #654, which will be superseded by this.

We might want to think about how to improve our testing so that the main viewer in headless mode undergoes more stringent testing, but there is at least some testing for now.

## Type of change

### DIFF
--- a/napari/_tests/test_viewer.py
+++ b/napari/_tests/test_viewer.py
@@ -158,3 +158,13 @@ def test_update(qtbot):
 
     # Close the viewer
     viewer.window.close()
+
+
+@pytest.mark.parametrize('layer_class, data, ndim', layer_test_data)
+def test_add_layer_headless(qtbot, layer_class, data, ndim):
+    viewer = Viewer(headless=True)
+    assert viewer.window is None
+    assert viewer.screenshot() is None
+
+    _ = add_layer_by_type(viewer, layer_class, data, visible=True)
+    check_viewer_functioning(viewer, None, data, ndim)
--- a/napari/_tests/utils.py
+++ b/napari/_tests/utils.py
@@ -91,11 +91,12 @@ def check_viewer_functioning(viewer, view=None, data=None, ndim=2):
     viewer.dims.ndisplay = 2
     assert np.all(viewer.layers[0].data == data)
     assert len(viewer.layers) == 1
-    assert view.layers.vbox_layout.count() == 2 * len(viewer.layers) + 2
-
     assert viewer.dims.ndim == ndim
-    assert view.dims.nsliders == viewer.dims.ndim
-    assert np.sum(view.dims._displayed_sliders) == ndim - 2
+
+    if view is not None:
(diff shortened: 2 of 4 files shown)


sofroniewn:
Actually, I guess we need to nest some imports here too.

It might also be nice to make is so that if you call `viewer.show` in headless mode it will try and create the GUI.

Thoughts @napari/core-devs?

jni:
@sofroniewn just wanted to cross-reference the last bit of my comment here: https://github.com/napari/napari/pull/1006#discussion_r386065994 , namely the idea that, rather than create a whole bunch of Qt-based objects without launching Qt, perhaps we can get to a place where napari provides the objects that are naturally headless, ie ViewerModel et al. What do you think? The main advantage is that we can eventually (Py 3.7+) do that without depending on Qt at all.

jni:
And btw I'll note that the [NEP-29](https://numpy.org/neps/nep-0029-deprecation_policy.html) deprecation policy has the ecosystem becoming Py3.7+ only this June!

sofroniewn:
@jni I'm very open to those ideas, though I can't really say I understand [PEP-562](https://www.python.org/dev/peps/pep-0562/) or what this actually looks like in practice.

I do wonder if it is easier to ask people to do

```python
from napari import Viewer
viewer = Viewer(headless=True)
```
vs
```python
from napari import ViewerModel as Viewer
viewer = Viewer()
```

We could consider renaming `ViewerModel` -> `HeadlessViewer` or `ViewerHeadless` though maybe those are bad names.

sumanthratna:
> I do wonder if it is easier to ask people to do
> 
> ```python
> from napari import Viewer
> viewer = Viewer(headless=True)
> ```
> 
> vs
> 
> ```python
> from napari import ViewerModel as Viewer
> viewer = Viewer()
> ```

not sure how much my opinion is valued but I personally prefer the first option :)) I think it would be a lot easier for cases when users might need to conditionally use headless napari (e.g., running a headless napari as a sort of smoke test or users adding a `--headless` option to their CLIs)

> We could consider renaming ViewerModel -> HeadlessViewer or ViewerHeadless

IMO `HeadlessViewer` is the best class name!

sofroniewn:
Thanks @sumanthratna for the input! This PR is on hold right now, but I'm sure we'll return to this at some point!!

sofroniewn:
We've moved on quite a bit since this PR was initially made. We've brought the `Viewer` closer in line with the `ViewerModel`, only the screenshot functionality and some console updates now separate them.

Following on from some ideas in https://github.com/napari/napari/pull/1799#issuecomment-720002919 I think we still want to make sure that workflows like

```python
window = Window(viewer_model)
```

work as expected, but these will come in other PRs. I will close this one shortly.

sofroniewn:
We've made great improvements in being able to using the viewer with qt in a headless mode, and so this PR is no longer needed

PR #1014, opened 2020-03-01
Refactor cleanup, prevent leaked widgets, add viewer.close method

tlambert03:
# Description
I was playing around with some things and realized that after a viewer has been created and closed within a session, a _lot_ of unused widgets remain in the QApplication.  For instance, try running the following code a few times in a row (in `master`) and you'll see that ~14 widgets are leaked for every new viewer that gets created.:
```python
import napari

with napari.gui_qt():
    print(QApplication.topLevelWidgets())
    v = napari.Viewer()
```
<details>
<summary>see output of second run</summary>

```
[<napari._qt.qt_viewer_dock_widget.QtViewerDockWidget(0x7f8f21c8da60, name="console") at 0x12e8a76c0>,
 <napari._qt.qt_modal.QtPopup(0x7f8f206f06f0, name="QtModalPopup") at 0x12e888900>,
 <PySide2.QtWidgets.QMenu(0x7f8f2377afb0) at 0x13475dac0>,
 <PySide2.QtWidgets.QMenu(0x7f8f23777090) at 0x13475d680>,
 <PySide2.QtWidgets.QMainWindow(0x7f8f23722250) at 0x13475d540>,
 <qtconsole.completion_html.CompletionHtml(0x7f8f21d52810) at 0x12e89f180>,
 <qtconsole.call_tip_widget.CallTipWidget(0x7f8f21d59080) at 0x12e89fb00>,
 <PySide2.QtWidgets.QWidget(0x7f8f23761280) at 0x10d574200>,
 <napari._qt.qt_modal.QtPopup(0x7f8f21f02510, name="QtModalPopup") at 0x12e88a580>,
 <PySide2.QtWidgets.QWidget(0x7f8f23711890) at 0x10d72a540>,
 <PySide2.QtWidgets.QMenu(0x7f8f2377df70) at 0x13475db80>,
 <napari._qt.qt_about_keybindings.QtAboutKeybindings(0x7f8f21c977b0) at 0x12e8a79c0>,
 <PySide2.QtWidgets.QWidget(0x7f8f236f18f0) at 0x1457d2d80>,
 <PySide2.QtWidgets.QMenu(0x7f8f2377b180) at 0x13475d8c0>]
```

</details>
some of these are related to the QtConsole and some may (or may not) be related to some of the bugs in our tests.  Others are dialogs (like the `about_keybindings` dialog which are created then hidden, rather than shown and destroyed on demand).

relatedly, try running the following code:
```python
with napari.gui_qt():
    napari.Viewer().window.close()
```
creating and destroying a viewer like that leaves something hanging such that (for me at least) a blank window shows up that cannot be cleared, and force quitting the entire ipython session is the only way to proceed.

**This PR overhauls the viewer shutdown process as follows**
- add a `viewer.close` method (which just calls `viewer.window.close`) to complement `viewer.show`
- make sure calling `viewer.window.close` actually closes every widget that got created for that viewer.
- standardize usage of `close()` method and `closeEvent()` for any cleanup on a QWidget class (rather than creating our own `shutdown` methods)
- change `QDialogs` to be created and shown on demand (rather than visibility toggled).  Unless it's really important to preserve dialog state over time, I think this should be preferred to avoid resource leakage.

**It also overhauls the tests as follows:**
- @ttung's `viewermodel_factory` has been elevated to `napari/conftest.py` (this is the only way that all submodules can use the factory), and has been changed to a `viewer_factory` to accept any viewer kwargs (not just `viewermodel`).
- all of duplicated viewer creation code has been replaced with `view, viewer = viewer_factory()` ... which cleans up after itself by calling `viewer.close()`.  Assuming this all works, that is how viewers should be created in tests.
- `qtbot.addWidget()` is _not_ used everywhere, since that method is just there as a [convenience](https://pytest-qt.readthedocs.io/en/latest/reference.html#pytestqt.qtbot.QtBot.addWidget) to call widget `close` and `deleteLater`, and the factory takes care of the cleanup.
  - similarly ... doing something like this will be an error because qtbot will have nothing to close if we do it ourselves:
```python
    console_a = QtConsole({'var_a': 3})
    qtbot.addWidget(console_a)
    console_a.close()
```

## Type of change

- [x] Refactor

# How has this been tested?

- [x] added a test that makes sure a bunch of widgets don't remain after creating & closing a viewer.

## Final checklist:


sofroniewn (commented):
I love how much duplicate code is removed!!

napari/viewer.py
```diff
@@ -99,3 +99,7 @@ def update(self, func, *args, **kwargs):
     def show(self):
         """Resize, show, and raise the viewer window."""
         self.window.show()
+
+    def close(self):
```
GenevieveBuckley, line 103: Thanks for this, you are removing a trivial inconvenience from my day :)

napari/_tests/test_napari.py
```diff
@@ -10,31 +11,20 @@
 
 
 @pytest.mark.parametrize('layer_type, data, ndim', layer_test_data)
-def test_view(qtbot, layer_type, data, ndim):
-
+def test_view(qapp, layer_type, data, ndim):
     np.random.seed(0)
     viewer = view_layer_type(layer_type, data)
```
ttung, line 16: suggestion: make `view_layer_type` a factory that cleans up the viewers that get created.

napari/_tests/test_viewer.py
```diff
@@ -153,8 +137,18 @@ def layer_update(*, update_period, num_updates):
...
+def test_viewer_cleanup(qapp):
+    """Test that closing the viewer doesn't leave any orphaned widgets."""
+    initial = len(qapp.topLevelWidgets())
+    viewer = Viewer()
+    viewer.close()
+    qapp.processEvents()
+
+    # somtimes unable to get the very last QMainWindow to clean up in pytest...
+    # but all the other widgets should be gone
+    assert initial <= len(qapp.topLevelWidgets()) <= initial + 1
```
ttung: does it make sense to stick this in the viewer_factory cleanup code?

ttung:
Generally, I'm concerned that there are too many ways to wire up the fixtures for the test.  Now there's qapp, qtbot, viewer_factory, and plausibly a `view_layer_type_factory`.  I'm not sure what the answer is, but as a relative newcomer to this codebase, I find that confusing and intimidating.

tlambert03:
I know it's a bit complicated, but there are a lot of moving parts in our tests, and the goal of this PR is to reduce that slightly.  all of those methods are _already_ being used in the current codebase (just with `viewermodel_factory` instead of `viewer_factory`, and `qtbot` is being used where `qapp` is sufficient).  `qtbot` and `qapp` are not ours, those are from pytest-qt (as you know) and [`qapp`](https://pytest-qt.readthedocs.io/en/latest/reference.html#pytestqt.plugin.qapp) is preferred when you don't actually need to send any events to QtObjects.  Here is what needs to be used, in increasing order of complexity, and if you can think of an easier way to package this to a newcomer while also making sure that nothing leaks through in our tests, i'm definitely open to suggestions:

1. `qapp`, the bare minimum fixture, if you simply need a `QApplication` running for your test.
2. `qtbot`, when you need both a `QApp` _and_ need to keep track of some objects. 
 [`qtbot.addWidget`](https://pytest-qt.readthedocs.io/en/latest/reference.html#pytestqt.qtbot.QtBot.addWidget), will ensure that the added widget is closed and deleted at the end of the test.

... if you wanted to use pytest-qt on its own without creating our own convenience fixtures for cleanup, then you would have to do this _every time_ you created a new viewer.
```python
viewer = Viewer()
view = viewer.window.qt_viewer
qtbot.addWidget(viewer.window)  # note, you cannot use viewer because it is not a QObject

# instead of qtbot.addWidget(), you could, at the end of every test
# viewer.close()
```
so...

3. `viewer_factory` creates a viewer for you, but cleans it up automatically.  it obviates the need for `qtbot` (in fact, `viewer.window` objects should _not_ be added to `qtbot` here, or else they will be doubly cleaned up, causing an error).  There may be cases when you _also_ use qtbot because you need to test widget interaction (i.e. sending clicks or awaiting signals).

sofroniewn:
This PR definitely steps us in the right direction @tlambert03 - we should probably add a section to our contributing guidelines around writing UI tests and which of the different options to use when. Could make the process much less intimidating.

Also if @ttung has ideas on how to simplify that would be great too!

GenevieveBuckley:
I would also like to see a short summary like the one Talley gives above added to the contributing docs.

tlambert03:
Can do!

tlambert03:
(thought I posted this yesterday but I guess not!)

I added a section to our contributor docs clarifying tests and cleanup [here](https://github.com/napari/napari/blob/c637abe39ba2f723e48b192ff665eddbe1b989ef/docs/developers/CONTRIBUTING.md#writing-tests).  If you want, I can also go through and replace all of the `qapp` fixtures with `qtbot`.  They do each have their purpose (as described in the blurb I wrote), but if it's easier just to always use qtbot we can do that and I'll update.

either way it'd be nice to get this in... I hope (but would like to know) that it will fix the residual pyqt5 test problems we've been having.  let me know what you think.

sofroniewn (approved):
This is great @tlambert03! Thanks for adding the docs on writing GUI tests. I'm approving, but maybe @ttung can take a quick look at the docs addition you linked to in your last comment and give feedback on them before merge.

napari/_tests/test_napari.py
tlambert03: since it's only used twice, and only in one module, I think i'd like to delay creating a new fixture for that.  they are cleaned with the viewer.close() at the end of the test

tlambert03:
Hold of on merging for now, just found a couple more leaked widgets in certain cases that I’d like to track down

docs/developers/CONTRIBUTING.md
```diff
@@ -102,6 +102,74 @@ Add and commit your changed files:
...
+## Running Tests
+
+To run our test suite locally, install test requirements and run pytest as follows:
+
+```sh
+pip install -r requirements/test.txt
+pytest
+```
+
+## Writing Tests
```
ttung, line 114: This is really great; nicely done.

napari/conftest.py
```diff
@@ -0,0 +1,59 @@
...
+
+import pytest
+from napari import Viewer
+from qtpy.QtWidgets import QApplication
+
+
+def pytest_addoption(parser):
+    """An option to show viewers during tests. (Hidden by default).
+
+    Showing viewers decreases tests by about %18.  Note, due to the placement
```
ttung: What does this mean?  Decreases test speed?

ttung (approved):
looks really great, pls see comments.

napari/conftest.py
tlambert03: whoops.  yeah, fixed.

tlambert03:
Thanks for looking @ttung.  Since your approval, I had to make a couple minor changes for testing on PyQt5 (related to your fullscreen shutdown, and @AhmetCanSolak's viewer.update() cleanup).
If want to double check, changes since your approval are [here](https://github.com/napari/napari/compare/cb9b397..9571288).

Otherwise, tests are passing and this is ready to go.

sofroniewn:
@tlambert03 this is looking great - i just tested again locally, and while the Viewer pops up far less than before, it still pops up a few times for me (when I run `pytest .`) I think during `napari/_tests/test_mouse_bindings.py` and `napari/_tests/test_viewer.py` maybe. Also is it easy to disable all the Qt popups during the `napari/_qt/_tests/test_qt_dims.py` etc. tests where just the sliders or some other Qt widget pops up? This can all wait to a follow up PR if you want, I know there's a lot in here already and it looks good to go otherwise

tlambert03:
There are a few tests where the test itself depends on the viewer being visible.  In order to make those tests pass without rewriting the tests (and their intentions) I used `viewer_factory(show=True)` (which overrides the default hidden behavior).  Same goes for some of the widget tests (many of which are testing virtual mouse clicks which required a visible widget).  So the short answer is, it can't _all_ be disabled without fundamentally changing what we are testing for in some cases.

(If whoever wrote those tests wants to consider rewriting them for hidden viewers, that could be something for another PR perhaps)

sofroniewn:
>  So the short answer is, it can't all be disabled without fundamentally changing what we are testing for in some cases.

Ok great, thanks for the explanation - we can revisit that stuff in followup PRs as you mention. I think this is good to go. I will let sit till tomorrow in case anyone wants to weigh in again, and then I will merge!!

PR #1015, opened 2020-03-02
Reduce code duplication: vispy _on_scale_change() & _on_translate_change()

GenevieveBuckley:
# Description

This PR is a refactor to reduce code duplication.

Currently, `VispyImageLayer` overrides the base layer methods `_on_scale_change()` and `_on_translate_change()`. But there's no reason this can't live in `VispyBaseLayer` instead, and this simplifies things a bit.

## Type of change

- [x] Refactor

# References

Original discussion with jni here: https://github.com/jni/napari/pull/6#discussion_r383197606

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/_vispy/vispy_base_layer.py
+++ b/napari/_vispy/vispy_base_layer.py
@@ -128,13 +128,18 @@ class VispyBaseLayer(ABC):
 
     def _on_scale_change(self, event=None):
         self.scale = [
-            self.layer.scale[d] for d in self.layer.dims.displayed[::-1]
+            self.layer.scale[d] * self.layer._scale_view[d]
+            for d in self.layer.dims.displayed[::-1]
         ]
+        if self.layer.is_pyramid:
+            self.layer.top_left = self.find_top_left()
         self.layer.position = self._transform_position(self._position)
 
     def _on_translate_change(self, event=None):
         self.translate = [
-            self.layer.translate[d] + self.layer.translate_grid[d]
+            self.layer.translate[d]
+            + self.layer._translate_view[d]
+            + self.layer.translate_grid[d]
             for d in self.layer.dims.displayed[::-1]
         ]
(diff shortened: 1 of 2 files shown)


jni (approved):
LGTM! Thanks @GenevieveBuckley! =)

sofroniewn:
Nice catch, looks good to me too - I will merge after all tests pass

tlambert03 (approved):

sofroniewn:
@GenevieveBuckley looking at this again, I notice that the `find_top_left` function which is in `VispyImageLayer` is now referenced inside `VispyBaseLayer`. Do you think it should be moved over too? Fine to be in a follow-up PR now that this is merged

sofroniewn:
Though that function depends on a `level_shapes` property on the Layer which I'm not sure is on the Base layer ... Hmm need to think more about it

GenevieveBuckley:
> @GenevieveBuckley looking at this again, I notice that the `find_top_left` function which is in `VispyImageLayer` is now referenced inside `VispyBaseLayer`. Do you think it should be moved over too?
> Though that function depends on a `level_shapes` property on the Layer which I'm not sure is on the Base layer ... Hmm need to think more about it

Ooh, well spotted! I don't like having this concept split across two classes like this... 
I'd prefer we picked a side: does `is_pyramid` and all related attributes/methods live completely in the base layer, or does it live completely in the image layer. I'll let you think about it some more.

GenevieveBuckley:
One question: what kind of support to pyramid *label layers* does napari have? That's probably something that will affect the decision here.

jni:
I don't remember where this concept came up last week, but we had the idea that `is_pyramid` would become `is_multiscale` and would apply across all layers (eg millions of points would get grouped into a single point if you zoom out enough, just like this happens with some mapping applications, for example). So, I think a lot of that logic should definitely go on the base layer.

Issue #1016, opened 2020-03-02
World coordinates API question


sofroniewn:
## ❓ Questions and Help

After discussions on Thursday I'm picking up work on #885 and https://github.com/jni/napari/pull/6 from @jni and @GenevieveBuckley to add support for world coordinates to the viewer (see previous discussions in #763 too and a number of comments from @d-v-b across these issues). I think I am fairly close to having a follow-on PR that has many of the features we want but I want to ask two questions before moving forward. I'm making a new issue to keep the discussion easy to follow and self-contained.

### Question 1:
**Can each layer just have a single nD affine transform that goes from data to world coordinates?** Ignore the complexity of the 3D rotation / arbitrary slicing stuff for now. I have a plan to fail gracefully and maybe one day support that if needed. Also ignore things like grid mode and image pyramids. What I really want to know is if people want more than that, i.e. piecewise affine or deformable transforms etc. and want our system to be constructed at a level of generality to allow that. I basically cannot see how to integrate those options with vispy and think that we should limit ourselves to affine, but want to check with @napari/core-devs before pushing further. I also think that each layer should only have one affine transform since affine transforms can be easily composed with each other and just having to keep track of one seems much easier to me.

### Question 2:
If the answer to Question 1 is yes, **what should the transform related API on the base layer look like?** Right now the base layer accepts `scale` and `translate`. Should it also accept `affine` and then compose transforms with `scale` and `translate`? If so what should it return when someone wants to do a roundtrip with layer? Should it only accept `affine` and nothing else? Should it accept `scale`, `translate`, a `rotation` matrix (and a `sheer` matrix?). Something else entirely?

I like layer accepting `scale`, `translate`, `rotation` and `sheer` the best as those 4 elements can be composed into an affine matrix which can then be decomposed to get the same elements back, even in nD (see https://github.com/matthew-brett/transforms3d/blob/master/transforms3d/affines.py for example). This allows us to maintain one source of ground truth internally, the affine matrix, but allows users to create it in an easy way.

We could also let a user pass an `affine` matrix directly (n+1 x n+1), but that would be an exclusive choice with the above, i.e. you couldn't also pass `scale`.

This would now mean we have 5 transform related keywords in the `add_*` methods, all optional, but one mutually exclusive with the other 4.

What do @napari/core-devs @d-v-b think?

jni:
@sofroniewn 

> Can each layer just have a single nD affine transform that goes from data to world coordinates?

[...]

PR #1018, opened 2020-03-02
Add ScaleTranslate transform

sofroniewn:
# Description
This PR would superceed #885 and follows on from work that @GenevieveBuckley and @jni have been doing around our world coordinates.

This PR does not actually add world coordinates though, it just adds the `Transform` base class and as `STTransform` class for transforms defined by a `scale` and `translate` and refactors the current scale / translate code to use it. It should be a non-breaking / non-functionality change that is purely a refactor.

Right now each layer still maintains three transforms `_transform`, `_transform_view`, and `_transform_grid`. We can maybe improve this for the future (I'd like to drop `_transform_grid` from the actual layer, and find a more elegant way to handle `_transform_view`).

Right now we also don't make use of the chain transform instead doing a compose of the STTransform. I know people want to chain transforms - see discussion in #1016, but note this , which says that if possible to compose it can be nice to do that too.

If people like this approach I will add tests to the `Transform` and `STTransform`. In a followup PR I will then introduce the concept of world coordinates which will affect how layers set their range, and how the dimension sliders interact with them, which right now is off. If that goes well it should be easy to adapt this approach to an `AffineTransform` too, and we can work from there. This PR is definitely just a baby step that at least gets us using Transforms.

Thoughts @jni @GenevieveBuckley?


napari/layers/transforms.py
```diff
@@ -0,0 +1,173 @@
+from __future__ import annotations  # noqa: F407
```
sofroniewn: @jni it looks from our tests that this line is not python 3.6 compatible

GenevieveBuckley:
@sofroniewn did you mean to include a link here? I'd like to read it.

>  but note this , which says that if possible to compose it can be nice to do that too.

On the off chance "compose" here means effectively the same thing as simplifying a transform chain down to one matrix, my comment here might be relevant: https://github.com/napari/napari/issues/1016#issuecomment-593666632

napari/layers/base/base.py
```diff
@@ -130,17 +131,16 @@ def __init__(
         self.scale_factor = 1
 
         self.dims = Dims(ndim)
+
         if scale is None:
-            self._scale = [1] * ndim
-        else:  # covers list and array-like inputs
-            self._scale = list(scale)
+            scale = [1] * ndim
```
GenevieveBuckley: Will this still work in cases where a (not None) scale argument is passed in? I don't see a line for that setting the `self._scale` attribute.

napari/layers/transforms.py
```diff
@@ -0,0 +1,173 @@
...
+        return STTransform(scale, translate)
+
+    def compose(self, transform: STTransform) -> STTransform:
+        scale = self.scale * transform.scale
+        translate = self.translate + self.scale * transform.translate
+        return STTransform(scale, translate)
+
+
+# class Translate(Transform):
+#     """n-dimensional translation (shift) class.
```
GenevieveBuckley: Why are these classes commented out?

napari/layers/transforms.py
```diff
@@ -0,0 +1,173 @@
...
+#
+#     def __newlike__(self, iterable):
+#         return ListModel(self._basetype, iterable, self._lookup)
+#
+#     def set_slice(self, axes: Sequence[int]) -> TransformChain:
+#         return TransformChain([tf.set_slice(axes) for tf in self])
+
+
+class STTransform(Transform):
+    """n-dimensional scale and translation (shift) class.
```
GenevieveBuckley, line 72: I like a lot of the implementation details here, but overall I feel like this is a clear sign our model is being influenced far too much by vispy's own structure.

sofroniewn:
@GenevieveBuckley yes - i meant to have a link to https://github.com/napari/napari/issues/1016#issuecomment-593417171, in particular:

>It'll take some work, but will be nice to avoid the overhead caused by chains of transforms if the code/developer knows how to simplify them. E.g. a chain consisting of entirely affines can be simplified to a single affine. Whether this is the dev's job or napari's job is a decision to make. I'd go with deferring to dev's for now, and putting this off to the future.

Your further comments there are well taken, including concerns of @VolkerH that you might not always want to collapse. Again what I'm taking away is increased flexibility is good.

napari/layers/transforms.py
sofroniewn: We can delete if people don't think we need anymore, but I left them there as @jni put them in. I am in favor of deleting as I think the `STTransform` supersedes this one.

napari/layers/transforms.py
sofroniewn: That's fair, though I do think a `Scale` then `Translate` transform was a pretty sensible choice on their part for an abstraction. I'd like to add `AffineTransform` in a follow-up PR, at which point we could either leave this or replace it. For things like grid mode, you might only ever need `STTransform`, but we have to have a good way of chaining / composing. In the future we might want to add `PiecewiseAffine` or `Deformable` too, but I want to work towards those slowly

napari/_vispy/vispy_base_layer.py
```diff
@@ -127,21 +127,25 @@ def _on_blending_change(self, event=None):
...
-        self.scale = [
-            self.layer.scale[d] * self.layer._scale_view[d]
-            for d in self.layer.dims.displayed[::-1]
-        ]
+        scale = (
+            self.layer._transform_view.compose(self.layer._transform)
+            .set_slice(self.layer.dims.displayed)
+            .scale
+        )
+        self.scale = scale[::-1]
```
jni, line 136: Can you add a comment here, "convert NumPy axis ordering to VisPy axis ordering"? Same below.

napari/_vispy/vispy_base_layer.py
```diff
@@ -127,21 +127,25 @@ def _on_blending_change(self, event=None):
...
     def _on_scale_change(self, event=None):
-        self.scale = [
-            self.layer.scale[d] * self.layer._scale_view[d]
-            for d in self.layer.dims.displayed[::-1]
-        ]
+        scale = (
+            self.layer._transform_view.compose(self.layer._transform)
+            .set_slice(self.layer.dims.displayed)
+            .scale
+        )
```
jni, line 134: I don't know if I was ever clear what scale view and scale were, but I'm definitely not clear now. =P Could you clarify in a comment? Are there maybe better names for these?

napari/layers/base/base.py
```diff
@@ -130,17 +131,16 @@ def __init__(
...
-        else:  # covers list and array-like inputs
-            self._translate = list(translate)
-        self._scale_view = np.ones(ndim)
-        self._translate_view = np.zeros(ndim)
-        self._translate_grid = np.zeros(ndim)
+            translate = [0] * ndim
+
+        self._transform = STTransform(scale, translate)
+        self._transform_view = STTransform([1] * ndim, [0] * ndim)
+        self._transform_grid = STTransform([1] * ndim, [0] * ndim)
```
jni: (1) why `[1]*` instead of NumPy?
(2) I'm not sure I like combining scale and translate into a single transform rather than chaining them, but we can forget about that for now... ;)
(3) I don't like exactly mirroring the name of the vispy transforms, as it makes it hard for me to tell what you're actually talking about here without going up to see the import. Why not call it ScaleTranslate for example? (The alternative is STTransformModel but I like that even less I think!)

napari/layers/transforms.py
jni: Oops! =) Should we drop 3.6? Note that this would only be slightly ahead of schedule, as NEP-29 has the community dropping 3.6 this June...

napari/layers/transforms.py
```diff
@@ -0,0 +1,173 @@
...
+    Scaling is always applied before translation.
+
+    An empty translation vector implies no translation, and an empty scale
+    implies no scaling.
+
+    Translation is broadcast to 0 in leading dimensions, so that, for example,
+    a translation of [4, 18, 34] in 3D can be used as a translation of
+    [0, 4, 18, 34] in 4D without modification.
+
+    Translation is broadcast to 1 in leading dimensions, so that, for example,
```
jni:
```suggestion
    Scale is broadcast to 1 in leading dimensions, so that, for example,
```

napari/layers/transforms.py
jni: Note that without this, it is an error to write `-> STTransform:` as the type annotation of a method of `STTransform`. See [PEP-563](https://www.python.org/dev/peps/pep-0563/)

napari/layers/transforms.py
```diff
@@ -0,0 +1,173 @@
...
+            ([1.0] * (coords.shape[1] - len(self.scale)), self.scale)
+        )
+        translate = np.concatenate(
+            ([0.0] * (coords.shape[1] - len(self.translate)), self.translate)
+        )
+        return np.squeeze(scale * coords + translate)
+
+    @property
+    def inverse(self):
+        return STTransform(1 / self.scale, -self.translate)
```
jni: So this turns out to be incorrect, because the inverse of doing a scaling followed by a translation, is a translation followed by a scaling, which cannot be represented by this class. Another reason why I liked implementing this as a chain. ;)

```python
In [2]: import numpy as np                                                         
In [3]: arr = np.array([[0.5, 0, 45], [0, 0.5, 89], [0, 0, 1]])                    
In [4]: np.linalg.inv(arr)                                                         
Out[4]: 
array([[   2.,    0.,  -90.],
       [   0.,    2., -178.],
       [   0.,    0.,    1.]])
```

You can see that if you want to implement it this way, you will have to divide translate by scale also.

napari/layers/transforms.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def set_slice(self, axes: Sequence[int]) -> STTransform:
+        return STTransform(self.scale[axes], self.translate[axes])
```
jni: Might this have issues when the `scale` is set at one dimensionality, but the viewer has since increased in ndims, and now axes is too big for the original `self.scale`?

napari/layers/transforms.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def set_pad(self, axes: Sequence[int]) -> STTransform:
```
jni: What does this do? Needs a docstring. =)

napari/layers/transforms.py
jni: Perhaps... See my comment above about STTransform and its inverse... =)

ST is a reasonably natural grouping but it seems to me to be more natural to leave them separate and manage them separately, as then you have a shot at reasoning about the order of operations...

jni (commented):
@sofroniewn I've given this the once-over. I'm still not sure about translate_grid, translate_view, and translate... In my imagining, these were all part of the transform chain, rather than created on the fly. Perhaps you just want to take things one PR at a time...?

napari/layers/transforms.py
sofroniewn: I'm also find holding back on those sorts of annotations and keeping 3.6 support till June. Can discuss at team meeting - curious how @kne42 feels

napari/_vispy/vispy_base_layer.py
sofroniewn: done

napari/layers/transforms.py
sofroniewn: done

napari/layers/transforms.py
sofroniewn: The code here is meant to prevent that https://github.com/napari/napari/pull/1018/files#diff-9ef2d7d09674e6ab30a1456a11fd74a8R350-R359

napari/layers/transforms.py
sofroniewn: yup absolutely, good catch - thanks - now the error is caught by a test too :-) Dividing by the translate by the scale is totally fine though to make the inverse, you have to do a similar option for an affine more generally and I think independent of the chain stuff. I'll respond in the main thread about why i like scale and translate going together, but this is at least now fixed and tested

napari/layers/base/base.py
GenevieveBuckley: I agree with (c)

sofroniewn: done (1) and (3). I still like ScaleTranslate combined, will comment in main thread on it

napari/_vispy/vispy_base_layer.py
sofroniewn: yeah - the whole `grid` / `view` / normal thing is a bit ugly - i will try and improve in the next PR that actually adds the world coordinates. For now I've added a comment directly on the base layer where those are created

GenevieveBuckley:
I'm also unclear about the differences & overlap between these, can you give us a bit of a breakdown?
* scale_view
* scale_factor
* scale
* translate_view
* translate_grid
* translate

Of course, I have my own theory about what they mean but I don't want to muddy the waters here when I could be wrong.

sofroniewn:
@GenevieveBuckley no worries - it's good to talk about this now. I'm hoping that the next PR will clean some of this up though the concepts need to be preserved. Here is are the comments that I am in the process of adding to this PR (I'm working on it right now, should have something to push soon)

```python
        # Create main transform mapping data to a world-like coordinate
        self._transform = ScaleTranslate(scale, translate)
        # Create additional transform mapping viewed data to data coordinate
        self._transform_view = ScaleTranslate(np.ones(ndim), np.zeros(ndim))
        # Create additional transform mapping world-coordinates into a grid
        # for looking at layers side-by-side
        self._transform_grid = ScaleTranslate(np.ones(ndim), np.zeros(ndim))
```

In more words - the `_view` part of the transform is really only needed for multiresolution data or data that is so big it can't fit on the graphics card so we need to downsample it and then apply a scale - say a 24k x 24k image.

The `_grid` part is really about looking at layers side-by-side and in the next PR I want to move it off the layer as it's really more of a viewer property / layer grouping property (I'm thinking about the #970 here)

`scale_factor` is a little different, it's something used to adjust things like edge thicknesses etc. and takes into account the vispy zoom level. It's really very visualization / rendering related. Maybe we can find a way to make that less confusing / put that elsewhere after some of the Model / View / Controller refactoring in #1020 as it also isn't a core layer thing

napari/layers/transforms.py
sofroniewn: doc string added - this pads the transform in the desired axes with scale 1 and translate 0 values, it's basically an inverse of set_slice with default values

sofroniewn:
I have updated this PR, including adding tests, doc strings, renaming to `ScaleTranslate`, dropping commented code, and maintaining python 3.6 compatibility. As you can see now this is really a pretty minor PR that doesn't change any functionality, but consolidates our scale and translate code into one object the `ScaleTranslate` transform. The reason I really like this is that it will make it much easier to reason about what transforms are what if we have less of them and it will be much easier to adopt affine transforms which lots of people want.

I've been following the discussion in #1016 closely and can see that people want lists of transforms, chaining of transforms, and more general transforms, in addition to world coordinates. All of that will start in the next PR - which will probably just be scoped at adding world coordinates. I'm trying for smaller more modular PRs as @jni like :-)

Let me know @GenevieveBuckley @jni if this looks good or if you have larger concerns that you don't think will be addressed by the followup PRs or that you'd like to still see in this PR

GenevieveBuckley:
@sofroniewn can we have a project for all these related issues and pull requests? I can't make one since I'm not a core dev, but it would be helpful to have.

I'm starting to lose track of all the moving parts, but I think this is everything:

### Pull requests:
* Add ScaleTranslate transform - https://github.com/napari/napari/pull/1018
* WIP: Proper models for world coordinates - https://github.com/napari/napari/pull/885

### Issues:
* World coordinates API question - https://github.com/napari/napari/issues/1016
* Clarifying the interface between the image layer and the data - https://github.com/napari/napari/issues/792
* Physical coordinate systems - https://github.com/napari/napari/issues/763

### Up next, planned pull requests:
* introduce the concept of world coordinates which will affect how layers set their range, and how the dimension sliders interact with them
* Move _transform_grid out from the layer
* chain transforms
* AffineTransform

GenevieveBuckley:
I re-read your PR @sofroniewn but I don't have any new comments to add.

napari/layers/base/base.py
```diff
@@ -130,17 +131,20 @@ def __init__(
...
-        else:  # covers list and array-like inputs
-            self._translate = list(translate)
-        self._scale_view = np.ones(ndim)
-        self._translate_view = np.zeros(ndim)
-        self._translate_grid = np.zeros(ndim)
+            translate = [0] * ndim
+
+        # Create main transform mapping data to a world-like coordinate
+        self._transform = ScaleTranslate(scale, translate)
+        # Create additional transform mapping viewed data to data coordinate
```
jni, line 150: This comment is still unclear to me, although you went over the outlines of it during group meeting. Based on our discussion in the meeting I see two scenarios where this is useful:

- you are using a pyramid, so the scale of the data at layer 1 (starting at 0) is (say) 2x the scale input by the user. Also, the coordinate of the 0th pixel is shifted slightly. (?) 
- you sent in an image that is bigger than the maximum texture size, so we downsample it 2x before sending it to vispy, so again, transform view will be 2x the scale.

It would be useful to put one of these worked examples in a comment here, just to clarify what "viewed data to data coordinate" means.

Another way of phrasing it that might make more sense to me:

"Additional transform for cases where the displayed data is *not* the base data given by the user. In such cases, we need to transform the coordinates from the displayed data space to the original input data space. One example of such a situation is viewing a higher level of an input pyramid, where `_transform` contains the transform only from the base level to world coordinates."

It's wordy, and might seem like overkill, but I really think it's worth erring on the side of wordiness here. One example I often refer to is the CPython implementation of dictionaries, which has about a 1-1 comment:code ratio. Which is great because I find the code incomprehensible! 😂 

https://github.com/python/cpython/blob/8bae21962bab2fac7630982abd73676b89930902/Objects/dictobject.c#L134-L224

napari/_vispy/vispy_base_layer.py
jni: @sofroniewn thanks! I think it's worth adding a comment here also, as it might not occur to someone reading this code to go find an explanation in the base layer.

napari/layers/transforms.py
```diff
@@ -0,0 +1,107 @@
...
+    def inverse(self):
+        """Return the inverse transform."""
+        return ScaleTranslate(1 / self.scale, -1 / self.scale * self.translate)
+
+    def set_slice(self, axes: Sequence[int]):
+        """Return a transform subset to the visible dimensions."""
+        return ScaleTranslate(self.scale[axes], self.translate[axes])
+
+    def set_pad(self, axes: Sequence[int]):
+        """Return a transform with added axes for non-visible dimensions."""
```
jni: Can you add Parameters and Returns sections? It's unclear from this signature and docstring what the correct input for this method is. I would have thought that `set_pad` would take in an integer, the number of total axes.

napari/layers/transforms.py
```diff
@@ -0,0 +1,107 @@
...
+        n = len(axes) + len(self.scale)
+        not_axes = [i for i in range(n) if i not in axes]
+        scale = np.ones(n)
+        scale[not_axes] = self.scale
+        translate = np.zeros(n)
+        translate[not_axes] = self.translate
+        return ScaleTranslate(scale, translate)
+
+    def compose(self, transform):
```
jni: You should probably specify that this requires an STTransform. You can do this with strings if you don't want to require Py3.7+. I'll note that I am personally ok with dropping 3.6 early, because we are still alpha.

jni (approved):
@sofroniewn I've approved but left a few comments that would be good to address before merging. Thanks!

napari/layers/base/base.py
sofroniewn: comment added, text slightly modified to include even more words - I agree with the sentiment and the practice

napari/layers/transforms.py
sofroniewn: Done

sofroniewn: Note `set_pad` takes a list so that it can act like an inverse to `set_slice` and pad not just at the beginning or end, but at desired locations

napari/layers/transforms.py
sofroniewn: I have gone with the string approach. Easy to switch over in the future if we want, but I'd rather wait to when we know we will use them (which could be very shortly). No need to make things unpleasant for our python 3.6 users until we need to.

sofroniewn:
I have made all the requested changes from @jni I will let this sit until tomorrow and then merge. Much more still to be done.

I've also made a `World Coordinates` project as requested by @GenevieveBuckley that will help us all keep track of this stuff better https://github.com/napari/napari/projects/10, thanks for the suggestion!!

PR #1020, opened 2020-03-03
RFC: Refactor proposal! Adding controllers to napari (TAKE 2)

shanaxel42:
Given the feedback provided in #1006 I created a modified proposal of the refactor proposal. It's different enough that I thought I'd make a separate PR so we can compare approaches and see which we like better (if any). The two main issue I tried to address with this new approach were:

- De-coupling the controllers from Qt and Vispy
- Making clear what would and would not be included in headless mode. 

In order to decouple the controllers from specific view libraries I introduced the concept of an "UpdateContract". The UpdateContract is an abstract class that defines the list of methods shared between all three components of the application (data, vispy, qt). Each component now imports the ImageUpdateContract and must implement the methods defined in it.

 So following on our `interpolation` example:

1.  a `_set_interpolation` abstract method is defined in the ImageUpdateContract class.
2. Image, QtImageControls, and VispyImageLayer all subclass ImageUpdateContract and implement their own version of `_set_interpolation` . (this bit was already happening they just maybe weren't all named the same). 
3. ImageController now only takes in a list of components that adhere to an ImageUpdateContract, it does not care what these components are made of (qt, vispy, etc)
4. When an interpolation event is triggered from any interface ImageController loops through the list of components it has and calls each's `_set_interpolation` respectively. 

In headless mode the ImageController would only get instantiated with a single component to update: the image layer. This means headless mode would include both the model files and their controllers. 

I added a small change in `test_image.py` just to show how you'd go about using the controller in a test. 

Some upsides to this approach:

- Controller is not tied to specific libraries. If we wanted to add a new front end it would be as simple as having that front end implement our UpdateContracts. 
- Same advantages as the last proposal (no more risk of endless update loops, one central location for changes, events all consolidated in one place)

Some downsides to this approach:

- a bit more boiler plate code
- the UpdateContract addition may not be immediately intuitive to someone not familiar with interfaces/abstract classes. 

Interested to see what the thoughts/questions/concerns are!


napari/layers/image/image_update_contract.py
```diff
@@ -0,0 +1,22 @@
+from abc import abstractmethod
+from ..base._base_update_contract import UpdateContractBase
+
+
+update_methods = {"interpolation": "_set_interpolation"}
```
royerloic: Why not just have the convention that we always prepend '_set_' to the method name? 
That would reduce boiler code here and, impoortantly I think,  avoid the need to maintain 
such maps that are not analysed and checks for errors by IDEs....

royerloic: That  would let you push that code up to the abstract class...

royerloic:
Very, very, nice @shanaxel42 ! I like it.

Definitely an improvement over the previous proposal!

One thought to make it even better: Perhaps things could be simplified 
further by merging the controller with the model. This would reduce the 
number of classes that need to be written and maintained and reduce boiler 
plate code. In that context, maybe the Controller could be renamed
to 'Syncroniser' or something similar... because, really, that's what it does. 
And why not have synchronizers be an integral part of the model? 

How this would work: 
The model would inherit from the controller all the sync machinery. 
Components (model, vispy, qt) would have to 'register' to the
controller machinery that would now be part of the model.
Here, registering just means passing it via a special method (in ControllerBase
but inherited by the model) that appends it to 'self.components_to_update'. 
To keep the existing behaviour, you do have to register the model itself as one of the 
components... that's perfectly fine.

One reason I think this simplification makes sense is because if you look at  'image_controller.py' 
there is _very_ little in there that is really image specific... Most of it can be pushed up to 
the base class, and the model (Image) can directly inherit that ControllerBase base class
and enjoy all its benefits. Everything else is unchanged in terms of how things work...
The essence of what you have done here @shanaxel42 is unchanged, just a bit compacted.

Then, headless is _just_ the model -- which does make sense to me. 
You want a view? register it... -- assuming that the view implements the contract.

Note for @jni : This does not break the rule of 'model should not know about its view', because,
the model only knows about objects that implement the update contract. UpdateContracts act as insulating interfaces, and ControllerBase simply keeps everything in sync and hides away the machinery for it.  
 
What do you think @shanaxel42 @sofroniewn ?  
Did I miss something that invalidates this proposal?

Finally, I think we should avoid using maps of strings to connect things together (see note in code), that makes automated refactoring impossible and will cause trouble down the line because one has to start tracking all the places where methods names could appear in strings -- something that IDE linters and python interpreters can't verify. Also, would be great to have @kne42 take on where we can pythonify things a bit... I am sure that we can leverage some of Python syntactic sugar here, but only if it makes sense and makes the code simpler. What do you think @jni ?

Curious to hear what @tlambert03 @ttung think too...

ok, going to sleep now...

napari/layers/image/image_update_contract.py
shanaxel42: yea that'd work! I like it

shanaxel42:
> One reason I think this simplification makes sense is because if you look at 'image_controller.py'
there is very little in there that is really image specific... 

This is true right now, but after the entire refactor I suspect there will be much more in the specific controllers. I'd also like to route things like mouse functions and model to model interactions through the controllers so making them separate helps there - not overloading the model class. It also means that the model only emits events and doesn't need to listen to them. Just some more things to consider. 

I do like the name Synchronizer...down to run with that : )

napari/layers/image/image_update_contract.py
shanaxel42: Updated with this change

sofroniewn (commented):
@shanaxel42 I really like how this looks, great job improving on #1006, and I like the addition of the suggestion from @royerloic to remove that dictionary. I'm actually ok with the controller being its own thing for now (and i prefer the word controller to synchronizer) rather than putting it directly on the model. I wonder as we extend support for the controller to the rest of the image layer / mousebindings / keybindings if this will be simpler.

kne42 (approved):
thumbs up from me :)

napari/_qt/layers/qt_image_base_layer.py
```diff
@@ -16,6 +17,9 @@ class QtBaseImageControls(QtLayerControls):
     def __init__(self, layer):
         super().__init__(layer)
 
+        # initialize qt events
+        self.events = EmitterGroup()
```
tlambert03, line 21:
```suggestion
        self.events = EmitterGroup(source=self)
```
so the event object will have a `.source` attribute that tracks what created the event.  Maybe also want to take advantage of `auto_connect` ... but haven't thought my way through that yet.

napari/_qt/layers/qt_image_layer.py
```diff
@@ -20,7 +23,7 @@ def __init__(self, layer):
             self.layer.interpolation, Qt.MatchFixedString
         )
         interp_comboBox.setCurrentIndex(index)
-        interp_comboBox.activated[str].connect(self.changeInterpolation)
+        interp_comboBox.activated[str].connect(self.emit_interpolation_event)
```
tlambert03, line 26: would be awesome if we could figure out how to adjust the `EventEmitter` objects to make this:
```
interp_comboBox.activated[str].connect(self.events.interpolation)
```
and dispense with the `emit_interpolation_event` altogether

napari/_qt/layers/qt_image_layer.py
```diff
@@ -80,8 +83,12 @@ def __init__(self, layer):
         self.grid_layout.setColumnStretch(1, 1)
         self.grid_layout.setVerticalSpacing(4)
 
-    def changeInterpolation(self, text):
-        self.layer.interpolation = text
+    def emit_interpolation_event(self, text):
+        self.events.interpolation(name="interpolation", value=text)
```
tlambert03, line 87: `name="interpolation"` is unnecessary here.  events have a `.type` property that contains the name of the `EventEmitter`.  
but I _love_ standardizing the `kwarg` to `value=text` instead of `interpolation=text`... I've wanted us to do that for a while

napari/_qt/qt_viewer.py
```diff
@@ -218,20 +224,34 @@ def _constrain_width(self, event):
...
 
+        # make qt_controls
+        controls = create_qt_controls(layer)
+        self.controls.addWidget(controls)
+        self.controls.widgets[layer] = controls
+
+        editable_components = [layer, vispy_layer, controls]
+
+        # make event controller
+        layer_controller = layer_to_controller[type(layer)](
```
tlambert03, line 250: as mentioned in my main comment, I don't think the controllers instances should live here. If they are only here because you need access to `vispy_layer` and `controls`, then the `controller.register()` method I mentioned would solve that.

napari/layers/base/_base_controller.py
```diff
@@ -0,0 +1,32 @@
...
+            ex.  [qt_image, vispy_image, image]
+        """
+
+        self.components_to_update = editable_components
+
+    def on_change(self, event=None):
+        """
+        Process changes when attribute is changed from any interface
+        """
+        name = event.name
```
tlambert03: `event.type` has this same info.  name not necessary

tlambert03 (commented):
I've been playing with this for the last couple hours and in general, I think it's the right direction to go in.   Definitely like it better than #1006.  There are things I like a lot and other things I'm less enthusiastic about:

I'm positively giddy about the possibility of removing all of the `self.layer.events.*.connect` statements from Qt and Vispy `__init__` methods.  And I _love_ enforcing standard `_set_` method naming convention... that definitely is long overdue.  I'm a fan of adding our own EmitterGroup to the Qt objects, instead of switching back and forth between event systems depending on the object type.

I still don't love the fact that the actual controller instances live in the `QtViewer.layer_controllers` dict.  I tend to agree with @royerloic that the controller instance could/should live on the layer (`Layer.controller`). I know it's possible that, with more stuff added to it, that might not look so good eventually, but i'm not convinced yet (but either way, I don't like it on the QtViewer object).  Relatedly, I still don't like that the controller gets instantiated up front with the three "listener" objects (the model layer, the vispy layer, and the QtController).  I'd much prefer each of these objects to register themselves with the controller after it is created.  That would make headless operation (or changing backends) as easy as not instantiating the other objects.  I made a modification of your PR with that sort of `controller.register` approach here: https://github.com/tlambert03/napari/tree/tlambert-1020

For me, the jury is still out on the `Contracts`.  I don't mind them, but need to think a bit more about them.  I don't see them as absolutely necessary here, and other than perhaps catching a poor concrete subclass, I'm not sure they add a ton. 

I have some additional small notes that I'll leave inline, but the last comment I would make is that I don't know that we're using the `EmitterGroups` and `EventEmitters` to their fullest.  They have lots of nice tricks on them (autoconnecting of methods to events, detection of the source of the event, etc...) that the controller here is kinda duplicating.  I'm not saying the controller doesn't belong, but I do think we shouldn't duplicate functionality that already exists in those Emitters.

ttung:
Re the comments in the meeting -- in many respects, this has evolved from being a controller to become like an event bus with sanity/error checking.  Given that, I don't find it to be too weird if the data layer instantiates the "controller".

napari/_qt/layers/qt_image_layer.py
shanaxel42: agreed, I was really sad when that didn't work

jni:
@ttung 

> in many respects, this has evolved from being a controller to become like an event bus

Event bus! I love that. To me it is much clearer than "Controller", which could mean anything for someone not intimately familiar with MVC (such as myself! 😅). Perhaps we can call these things ImageEventBus or ImageEventHandler or something similar before this is done? I'm not wedded to this idea, just throwing it out there that it might be a more intuitive/descriptive name for some folks.

shanaxel42:
> Perhaps we can call these things ImageEventBus or ImageEventHandler

I like ImageEventHandler! Pretty intuitive

royerloic:
I happy with the direction that this conversation is taking:
(i) I like the notion of 'event bus' that echos what I meant with 'synchroniser' 
(ii) I agree with @tlambert03 that we need to do explicit registration.
(iii) I am still very much in favor of having the _generic_ event bus be 'inherited' so that the model/data immediately gets the functions for registration and more, very useful. 
(iv) If we want to have keyboard control, mouse control, tablet control, network control, etc... These should all be implemented as ImageEventHandlers (following @jni  nomenclature), and certainly not be implemented in the 'controller' itself.   
(v) About contracts/interfaces, that is very much a Java/C#?... way of thinking, which I like, which means that IDEs have a chance to warn us that the contract is not respected before even running the code. However, that might not be the most pythonic way of doing things.  I thus tend to agree with  @tlambert03 that we need to think more about this. Do @jni and @kne42 have ideas on what would be a more pythonic approach? What is the standard way in python to enforce compliance to a given api? ABCs might be it really.

jni:
> Do @jni and @kne42 have ideas on what would be a more pythonic approach?

Well, protocols are the hot new thing: https://www.python.org/dev/peps/pep-0544/#using-protocols

They exist only in Python 3.8 though, though I think @tlambert03 found a backport? Anyway, personally, I think you are right @royerloic that interfaces (I like that name better than contracts) are the right way to think about this, so I would love to use the type machinery to do this. But we will be at the forefront of the Python language! :sweat_smile:

shanaxel42:
> Well, protocols are the hot new thing: https://www.python.org/dev/peps/pep-0544/#using-protocols

Just saw this after making https://github.com/napari/napari/pull/1040/files ! Hopefully it addresses @royerloic 's  concerns. But I'm open to making the interfaces protocols, it's pretty similar to how things are set up now except the actual components wouldn't need to inherit the Interface classes. But would we be ok with having napari be only Python 3.8 compatible?

tlambert03:
the backport for `Protocols` is [`typing-extensions`](https://pypi.org/project/typing-extensions/).  Works with python ≥3.5

shanaxel42:
> [`typing-extensions`](https://pypi.org/project/typing-extensions/)

Ok let me try this out

sofroniewn:
I am closing this in favor of #1040 which people have voice a preference for

PR #1023, opened 2020-03-05
Sorting and disabling of hook implementations

tlambert03:
# Description
Currently, hook implementations for a give hook specification are called based on the order that they are registered ("first in last out").  This poses a potential problem if two plugins recognize the same format, but the user would like one to have precedence.  For instance, plugin 1 reads .nd2 and .dv, whereas plugin 2 reads only .dv but does a better job of it.  We'd like to have both plugins installed, but we want plugin 2 to be in line first.

This PR adds a mechanism to sort plugin hooks _after_ registration with a `permute_hookimpls` function.  Sorting is made a bit more complicated by the presence of [hook wrappers](https://pluggy.readthedocs.io/en/latest/#wrappers) (a less common hook type that gets called before and after all the other hooks). 

## Type of change

# References

# How has this been tested?

## Final checklist:


napari/_qt/qt_plugin_list.py
```diff
@@ -0,0 +1,199 @@
...
+        order : list
+            A list of str, hookimpls, or module_or_class, with the desired
+            CALL ORDER of the hook implementations.
+        """
+        if not self.hook:
+            return
+        permute_hook_implementations(self.hook, order)
+
+
+class QtPluginSorter(QDialog):
```
tlambert03, line 205: > @sofroniewn: This is great - as we add more hook specs / see how more plugins work, we might want to revisit, provide more options etc, but for now works well

napari/_qt/qt_plugin_list.py
```diff
@@ -0,0 +1,199 @@
...
+        Usually instantiated by QtPluginSorter.
+
+        Parameters
+        ----------
+        hook : pluggy.manager._HookCaller, optional
+            A pluggy HookCaller to show implementations for. by default None
+            (i.e. no hooks shown)
+        parent : QWidget, optional
+            Optional parent widget, by default None
+        """
```
tlambert03, line 90: > @jni said: @tlambert03 refer to my comment on other PRs... =) ie this docstring should, tragically, be on the class docstring.

napari/_qt/qt_plugin_list.py
```diff
@@ -0,0 +1,199 @@
...
+        hook : _HookCaller, optional
+            A pluggy HookCaller to show implementations for. by default None
+            (i.e. no hooks shown)
+        """
+        self.clear()
+        self.hook = hook
+        if not hook:
+            return
+        for hookimpl in reversed(hook.get_hookimpls()):
+            self.addItem(QtHookImplItem(hookimpl))
```
tlambert03: > @jni said: Why are these reversed? Could you add a comment? I presume this has to do with the display order of the widget vs the priority order of the hookimpls...

napari/_qt/qt_plugin_list.py
```diff
@@ -0,0 +1,199 @@
...
+        initial_hook : str, optional
+            If provided the QComboBox at the top of the dialog will be set to
+            this hook, by default None
+        firstresult_only : bool, optional
+            If True, only hook specifications that declare the "firstresult"
+            option will be included.  (these are hooks for which only the first
+            non None result is returned).  by default True (because it makes
+            less sense to sort hooks where we just collect all results anyway)
+            https://pluggy.readthedocs.io/en/latest/#first-result-only
+        """
```
tlambert03: > @jni said: As above...

napari/_qt/qt_plugin_list.py
```diff
@@ -0,0 +1,199 @@
...
+        self.hookComboBox.addItems(hooks)
+
+        self.hookComboBox.activated[str].connect(self.change_hook)
+        self.hookList = QtHookImplListWidget()
+
+        title = QLabel('Plugin Sorter')
+        title.setStyleSheet(
+            "QLabel { background-color : rgb(38, 41, 47); "
+            "color : #ddd; font-size: 18pt; }"
+        )
```
tlambert03: > @jni said: Do these stylings belong here rather than in our stylesheets directly?  this seems like it'll just plain look terrible when switching themes...?

napari/_qt/qt_plugin_list.py
```diff
@@ -0,0 +1,199 @@
...
+
+        Parameters
+        ----------
+        hook : str
+            Name of the new hook specification to show.
+        """
+        if hook == self.NULL_OPTION:
+            self.hookList.set_hook(None)
+        else:
+            self.hookList.set_hook(getattr(self.plugin_manager.hook, hook))
```
tlambert03: > @jni said: This line is confusing to me... Does self.plugin_manager.hook contain attributes for every possible hook name?  This might make more sense to me:
``` getattr(self.plugin_manager.hooks, hookname)```
? Am I parsing this right?

File: napari/_qt/qt_plugin_list.py [hunk 7]
```diff
@@ -0,0 +1,199 @@
...
+class QtHookImplItem(QListWidgetItem):
+    """Item in a QtHookImplListWidget."""
+
+    def __init__(self, hookimpl, parent=None):
+        self.plugin_name = hookimpl.plugin_name
+        if self.plugin_name == 'builtins':
+            self.plugin_name = 'napari built-in'
+        super().__init__(self.plugin_name, parent)
+        self.hookimpl = hookimpl
+        self.setBackground(Qt.black)
```
tlambert03: > @jni said: Here also, the theming within the class seems to go against our existing design principles.

napari/plugins/utils.py
```diff
@@ -0,0 +1,160 @@
...
+                msg += (
+                    " If all items in `order` "
+                    "argument are not strings, they are assumed to be an "
+                    "imported plugin module or class."
+                )
+            raise ValueError(msg) from e
+
+    # make new arrays for _wrappers and _nonwrappers
+    _wrappers = []
+    _nonwraps = []
```
tlambert03: > @jni said: What are wrappers and non-wrappers?

napari/plugins/utils.py
```diff
@@ -0,0 +1,160 @@
...
+    if all(isinstance(o, HookImpl) for o in order):
+        attr = None
+    elif all(isinstance(o, str) for o in order):
+        attr = 'plugin_name'
+    elif any(isinstance(o, str) for o in order):
+        raise TypeError(
+            "order list must be either ALL strings, or ALL modules/classes"
+        )
+    else:
+        attr = 'plugin'
```
tlambert03: > @jni said: What is attr?

napari/plugins/utils.py
```diff
@@ -0,0 +1,160 @@
...
+    # were not specified in `order` argument
+    for i in sorted(indices, reverse=True):
+        del hookimpls[i]
+
+    if hookimpls:
+        _wrappers = [x for x in hookimpls if x.hookwrapper] + _wrappers
+        _nonwraps = [x for x in hookimpls if not x.hookwrapper] + _nonwraps
+
+    hook_caller._wrappers = _wrappers
+    hook_caller._nonwrappers = _nonwraps
```
tlambert03: > @jni said: honestly I find this function extremely hard to parse. From the docstring, I expected something like:
```python
hook_caller.hooks = [_get_hook_implementation(hook_caller, hook_id) for hook_id in order]
```
with a few checks. Can you explain why it is so much more complicated?

tlambert03 (commented):
adding line-specific comments provided by reviews in #937

GenevieveBuckley:
Will there be a way to force use of a specific plugin (maybe by opening the plugin through a menu instead of the default drag and drop)?

I'm wondering about what happens in cases where people might want different a preferred ordering depending on the circumstances. Eg: Plugins 1, 2, and 3 can all read tif, png, and jpeg files, but I have conflicting requirements & want plugin 2 to be used for tiff files, but plugin 1 should be used for any png files, etc. If you can select a specific plugin yourself there's no issue - you just set the default ordering to what's most common for you & force use of a specific plugin in the other cases. Is that how this is intended to work?

I hope this isn't too much of a side discussion; it's more of a user discussion than it is about this PR.

tlambert03:
yeah, I do imagine adding a "Open file with plugin..." menu option.  would be easy, and useful.

napari/_qt/qt_plugin_list.py
tlambert03: fixed

napari/_qt/qt_plugin_list.py
tlambert03: fixed

napari/_qt/qt_plugin_list.py
tlambert03: yeah, i agree, that would make more sense, but that's not how pluggy made it.  [`PluginManager.hook`](https://github.com/pytest-dev/pluggy/blob/6e8b6d4e8e0519e50af05b4e4b90b18b924bcbcc/src/pluggy/manager.py#L75) is just an [empty namespace object](https://github.com/pytest-dev/pluggy/blob/6e8b6d4e8e0519e50af05b4e4b90b18b924bcbcc/src/pluggy/hooks.py#L200-L204) that stores all of the "`_HookCaller`" objects for each hook specification ... and it's called "`hook`" not "`hooks`"

(if you're curious, the `_HookCaller` objects get added to to the `PluginManager.hook` object [here](https://github.com/pytest-dev/pluggy/blob/6e8b6d4e8e0519e50af05b4e4b90b18b924bcbcc/src/pluggy/manager.py#L122-L125)).

want a comment to that extent in our code?

napari/_qt/qt_plugin_list.py
tlambert03: moved

napari/plugins/utils.py
tlambert03: sorry again for the complexity, I think this was a case of premature optimization.  It's possible I should restrict the range of inputs this function accepts, which would simplify things quite a bit. But let me first explain what's going on here and what my goals were, and you can guide me as to what I should do.

First off, just a point of clarification: the "`hook_caller`" is the thing _after_ `PluginManger.hook` (e.g. in `plugin_manager.hook.napari_get_reader`, `napari_get_reader` is a `_HookCaller` instance).  It has no attribute called `hooks` (where you might expect to find implementations), so you can't just reorder them like that.  It actually has _multiple_ internal lists storing hook implementations.  One is `._nonwrappers`, where you find the "normal" kind of hook implementations that just get called in sequence... the other is `._wrappers`, where you find "[hookwrappers](https://pluggy.readthedocs.io/en/latest/#wrappers)" (implementations that were decorated with `@hookimpl(hookwrapper=True)`).  These are kind of like context managers that get called both before and after the nonwrappers (see pluggy docs for more).  You can get _all_ hook implementations, regardless of type, using `_HookCaller.get_hookimpls()` (which just returns the combined lists... but is not mutable).  This division of implementation types, and the fact that `_wrappers` must be sorted separately from `_nonwrappers`, makes for part of the difficulty.

Second, `HookImpl` instances have no "id" _per se_ (outside of their object id). They have a `plugin_name` attribute, and a `plugin` attribute (the actual module from which they were imported)... and here's where my premature optimization comes in:  I wanted this function to take in an `order` argument that was a list of _either_ `plugin_name`, `plugin` (module), _or_ the actual `HookImpl` instances themselves. (see the `HookOrderType` I declared above).

Furthermore, in my function, `order` needn't be a complete set of hook implementation pointers.  You can call `permute_hook_implementations(hook_caller, ["my_plugin"])` and it will move  the implementation from `my_plugin` to the front of the list (similar to the [`use_plugin` function in `skimage`](https://github.com/scikit-image/scikit-image/blob/8cf26a04cf20ac81c8233907a02bf0850e3961e0/skimage/io/manage_plugins.py#L212))... (while also being conscious of _hookwrappers_).  Or you can make `order` a (partial) list of `HookImpl` instances, and it will move the ones specified to the front of their respective (wrapper/nonwrapper) lists... this is how I use the function in `qt_plugin_list.py`.
(you can look at `test_permute_hook_implementations` for examples of all the cases I was accommodating).

The function _does_ work at the moment for all of these cases (with tests), but if you'd like to restrict the function signature in the interest of clarifying the logic of the body, let me know how you'd design it.

napari/plugins/utils.py
tlambert03: see comment below

napari/plugins/utils.py
tlambert03: see comment below: this bit is about determining the `type` of the members of the `orders` list.  (due to a perhaps overly permissive function signature).

napari/_qt/qt_plugin_list.py
tlambert03: These are reversed because `hook.get_hookimpls()` actually returns hook implementations in an order _reversed_ from their call order.  And I wanted the GUI to show it in call order, which I thought was more intuitive.  [Here is the line in pluggy](https://github.com/pytest-dev/pluggy/blob/6e8b6d4e8e0519e50af05b4e4b90b18b924bcbcc/src/pluggy/callers.py#L169) where the list of hook implementations (the output of `get_hookimpls()`) is reversed before calling them.

tlambert03:
OK, I'm gonna call this ready for review.  I moved the relevant comments from #937 over here to remind you.  I updated all the easy comments (resolved ones I didn't think you'd care about anymore), added lots of docs, added the ability to disable hook implementations (in addition to sorting them) and improved the styles and behavior of the GUI parts a lot.

I stopped short of refactoring `permute_hook_implementations` because i want some feedback before restricting functionality there.  @jni, let me know what you think after reading my clarification.

here's the visual on the GUI plugin sorter:
Untitled

sofroniewn:
Looking at the docs PR #1033 I see this line

> A single plugin package may implement more than one *hook specification*, but
may not declare more the one *hook implementation* for any given specification.

I guess I sort-of always new this @tlambert03, but I'm curious how we think about this now in practice as we start having reader plugins and how it relates to this PR.

Naively I would have liked the idea that one big `napari-io` plugin could define many readers (say one for `.mat`, one for `.npy`, one for `.nd2` etc) and then we could sort them all using this mechanism. While its still possible to have a single `napari-io` plugin that can read all those file types it will have to contain it's own if-else logic and won't be able to take advantage of the sorting done in this PR. Given that it seems like we might be better off encouraging many smaller plugins say a `napari-io-mat`, `napari-io-npy`, `napari-io-nd2` etc.

What are your thoughts on all this @tlambert03?

tlambert03:
> While its still possible to have a single `napari-io` plugin that can read all those file types it will have to contain it's own if-else logic and won't be able to take advantage of the sorting done in this PR. However it seems like we might be better off encouraging many smaller plugins say a `napari-io-mat`, `napari-io-npy`, `napari-io-nd2` etc.

yeah, that's exactly right.  A single plugin can read a bunch of different formats by putting logic in it's `napari_get_reader` implementation:

```python
@napari_hook_implementation
def napari_get_reader(path):
    if path.endswith('.nd2'):
        return nd2_reader
    if path.endswith('.mat'):
        return mat_reader
    ... 
```
but, as you state, that plugin will show up as a single plugin in this sorting mechanism.

However, I don't really see that as an issue just yet, since plugin sorting should only be required for two plugins that read the _same_ filetype.  That is, I don't really see a need to sort `napari-io-mat` relative to `napari-io-nd2` if their filetype support is strictly non-overlapping... and I haven't yet envisioned a scenario where a single plugin needs to have two conflicting readers for a given file-type.  And if they do, I think it's sufficient to leave that logic up to the plugin itself (i.e, i don't think we need to support "sub-plugin" implementation sorting).

In any case, if we really want this sort of behavior internally in `napari`, then we can just register different modules as different internal plugins (remember, even "packages" can provide more than one "plugin" via entry points... and we can too).  Currently, we have one internal "plugin" module at `napari.plugins._builtins` and it is registered [here](https://github.com/napari/napari/blob/0e8deac12b772afb4924549d5bf6e53015083224/napari/plugins/manager.py#L55).  but if need be, we can add more `self.register()` calls, with different modules... each of which would be its own "plugin" in the sorting dialog.

sofroniewn:
Yes, that all makes sense @tlambert03. Thanks!! I think we can press on with the current approach for now, but it might be worth continuing to thing about this when we add functional and interactive plugins.

I'll give this a review soon

tlambert03:
One additional thing perhaps worth noting: the sorting here is additionally only necessary for hook specifications that specify [`@hook_specification(firstresult=True)`](https://pluggy.readthedocs.io/en/latest/#first-result-only) (like `napari_get_reader`), but the default behavior for hook specifications is to actually run _all_ hook implementations and return all the results.  Remains to be seen how we’ll be implementing the functional plugins, but just wanted to point that out.

napari/_qt/qt_main_window.py
```diff
@@ -190,6 +192,17 @@ def _add_window_menu(self):
         self.window_menu = self.main_menu.addMenu('&Window')
         self.window_menu.addAction(exit_action)
 
+    def _add_plugins_menu(self):
+        order_plugin_action = QAction("Plugin call order...", self._qt_window)
```
sofroniewn, line 192: The sort plugin window is blocking which makes sense if I think about the code (as are our keybindings / info ones) but it tripped me up a little while testing as I was expecting to be able to toggle boxes on / off or rearrange things and drag and drop at the same time. At first i thought something was wrong with my plugin or file, then i noticed i wasn't seeing any error messages, and then finally i realized that no drag and drop was working at all. It might be nice to have the window be non-blocking to allow people to continue experimenting with drag and drop while the window is open

napari/_qt/utils.py
```diff
@@ -151,3 +151,45 @@ def square_pixmap(size):
     painter.drawRect(1, 1, size - 3, size - 3)
     painter.end()
     return pixmap
+
+
+def drag_with_pixmap(list_widget: QListWidget) -> QDrag:
```
sofroniewn, line 159: This is so nice - really makes me think we can drop the whole custom layer list widget and use a similar approach here - the only thing we need to make sure is that we can embed the thumbnails and shape icons inside the `QListWidgetItem` - that would be such a nice refactor @tlambert03 :-)

sofroniewn (approved):
This looks good to me, I've only left one comment about the window blocking drag / drop to the viewer, which I think should be fixed otherwise I think it is ready. I am approving, but hopefully the above can be addressed, and we should also wait for @jni to review. Great work again @tlambert03!!

napari/_qt/qt_main_window.py
tlambert03: ah good point.  Yeah, if you were confused momentarily that's no good.
I made it modal/blocking just to stay with the general aversion I've sensed here to avoid additional windows being open.  I can definitely see the argument for having it open while trying some io, and I'm fine with that.  Curious whether @jni or @royerloic have other preferences though?  If others _did_ want to keep it modal, we could definitely do more to make it clear that the window needs to be dismissed before proceeding.

napari/_qt/utils.py
tlambert03: yeah, I agree!  I feel like that little function there helped me feel like there might just be a clean path for that `LayerList` refactor.  Will think about that with all my newfound free time.

jni:
Not a full review yet but fwiw I prefer it to be non-blocking, people will want to reorder things then test the behaviour *all the time*. It'll be annoying to have to close and open again over and over. It goes from a pop-up (mildly annoying) to pop-up after pop-up (infuriating). =P

tlambert03:
I made the dialog non-blocking, and turned it into a dock widget when it opens.  so you can toggle and rearrange things while dragging/dropping files:
Screen Shot 2020-03-19 at 12 03 30 PM

sofroniewn:
Oh so nice, love the dock widget - we should use more of them!! This is so slick, I just tried it out. This is good to go as far as I am concerned, but I will let @jni review before merge

napari/_qt/qt_plugin_list.py
```diff
@@ -0,0 +1,301 @@
...
+        layout.addWidget(self.plugin_name_label)
+        layout.setStretch(2, 1)
+        layout.setContentsMargins(0, 0, 0, 0)
+
+    def _set_enabled(self, state: bool) -> None:
+        """Set the enabled state of this hook implementation to ``state``."""
+        # "hookimpl.enabled" is NOT a pluggy attribute... we are adding that to
+        # allow skipping of hook_implementations.
+        # see plugins.io.read_data_with_plugins() for an example
+        setattr(self.item.hookimpl, 'enabled', bool(state))
```
jni: Why use setattr though rather than setting it directly? Also, if `state` is a bool, why cast it to bool again?

jni: (thanks for the comment though, it is much appreciated! =)

napari/_qt/qt_plugin_list.py
```diff
@@ -0,0 +1,301 @@
...
+        by default None (i.e. no hooks shown)
+    """
+
+    order_changed = Signal(list)  # emitted when the user changes the order.
+
+    def __init__(
+        self,
+        parent: Optional[QWidget] = None,
+        hook: Optional[_HookCaller] = None,
+    ) -> None:
```
jni: I think the `-> None` can be omitted in an init, as it is expected...

napari/_qt/qt_plugin_list.py
```diff
@@ -0,0 +1,301 @@
...
+        self.setAcceptDrops(True)
+        self.setSpacing(1)
+        self.setMinimumHeight(1)
+        self.setSizePolicy(
+            QSizePolicy.MinimumExpanding, QSizePolicy.MinimumExpanding
+        )
+        self.order_changed.connect(self.permute_hook)
+        self.set_hook(hook)
+
+    def set_hook(self, hook: Optional[_HookCaller]) -> None:
```
jni: Can you clarify for me the difference between a hook and a hook specification?

napari/_qt/qt_plugin_list.py
```diff
@@ -0,0 +1,301 @@
...
+    def add_hook_implementation_to_list(self, hookimpl: HookImpl) -> None:
+        """Add a list item for ``hookimpl`` with a custom widget.
+
+        Parameters
+        ----------
+        hookimpl : HookImpl
+            The hook implementation object to add to the list.
+        """
+        # don't want users to be able to resort builtin plugins.
+        # this may change in the future, and might require hook-specific rules
```
jni: I'm really not sure about this...?

napari/_qt/qt_plugin_list.py
```diff
@@ -0,0 +1,301 @@
...
+    NULL_OPTION = 'select hook... '
+
+    def __init__(
+        self,
+        plugin_manager: Optional[PluginManager] = None,
+        parent: Optional[QWidget] = None,
+        *,
+        initial_hook: Optional[str] = None,
+        firstresult_only: bool = True,
+    ) -> None:
```
jni: As above

napari/_qt/qt_plugin_list.py
```diff
@@ -0,0 +1,301 @@
...
+        The list widget that displays (and allows sorting of) all of the hook
+        implementations for the currently selected hook.
+    """
+
+    NULL_OPTION = 'select hook... '
+
+    def __init__(
+        self,
+        plugin_manager: Optional[PluginManager] = None,
+        parent: Optional[QWidget] = None,
```
jni: I might make parent keyword-only also...

napari/plugins/manager.py
```diff
@@ -149,6 +149,48 @@ def _register_module(self, plugin_name: str, module_name: str) -> None:
...
+                for imp in hook_caller.get_hookimpls()
+                if imp.plugin_name == plugin_name
+            )
+        except StopIteration:
+            raise KeyError(
+                f"No implementation of {hook_spec} found "
+                f"for plugin {plugin_name}."
+            )
+
+        setattr(implementation, 'enabled', enabled)
```
jni: This could be simply `implementation.enabled = enabled`?

napari/plugins/utils.py
jni: @tlambert03 I still don't quite grasp all the complexity here, but at least some of the comments should be inline here, at the point that the variables are introduced. `attr` is such a generic name...

jni (commented):
@tlambert03 a partial review for now. I still don't understand a lot of the complexity, even after your comments, but it's late so maybe I need to try again later with fresh eyes.

Overall, I *would* prefer to limit the complexity at the cost of future expandability that we may or may not need.

The other part that isn't clear to me here is the model/view separation here. It *seems* to be there, in the plugin manager together with the permute_hook_implementation function, but overall I find the mapping a lot harder to follow than with our layers... Can you provide an example script with code that shows how I would reorder the IO plugins programmatically?

Finally, my distaste for the word `hookimpl` is well documented... There's quite a few places where it could be expanded, e.g.

```python
hookimpl = pluggy.HookimplMarker('dummy')
```
to

```python
dummy_hook_implementation = pluggy.HookimplMarker('dummy')
```

napari/_qt/qt_plugin_list.py
tlambert03: thanks.  that was me misinterpreting a [mypy example](https://mypy.readthedocs.io/en/stable/common_issues.html#no-errors-reported-for-obviously-wrong-code) stating "Must specify return type to type-check".  I realize now that means only if there are no arguments with type hints.

napari/_qt/qt_plugin_list.py
tlambert03: I'm not sure why I used `setattr` here.  brain fart?  state needs to be cast because the checkbox.stateChanged signal may return an int.  I'll update the type signature

napari/_qt/qt_plugin_list.py
tlambert03: removed, `builtins` now appears in the list and can be sorted.

napari/_qt/qt_plugin_list.py
tlambert03: by hook I meant "`_HookCaller`" instance.  (I renamed `hook` and `set_hook` to `hook_caller` and `set_hook_caller` to reduce confusion).  In "`plugin_manager.hook.napari_get_reader`", the 
"`napari_get_reader`" is an instance of `pluggy.manager._HookCaller`.  It is basically an object that maintains a list of (and knows how to call) all of the `HookImpl` objects that have been discovered for the `hook_specification` with the same name.

tlambert03:
> I still don't understand a lot of the complexity, even after your comments... Overall, I *would* prefer to limit the complexity at the cost of future expandability that we may or may not need.

let's chat on zulip or something then.  If looking at the test I mentioned in https://github.com/napari/napari/pull/1023#discussion_r392613808 didn't clarify the current capabilities, I'll try to explain better what it can currently do, and you can tell me what you'd prefer to cut.

> The other part that isn't clear to me here is the model/view separation here. It *seems* to be there, in the plugin manager together with the permute_hook_implementation function, but overall I find the mapping a lot harder to follow than with our layers... Can you provide an example script with code that shows how I would reorder the IO plugins programmatically?

the model and single source of truth for the call order is the `_HookCaller` instance: `plugin_manager.hook.napari_get_reader`.  To programmatically change it, for instance:

```python
hook_caller = napari_plugin_manager.hook.napari_get_reader
permute_hook_implementations(hook_caller, ["plugin_3", "plugin_1", "plugin_2"] )
```

> Finally, my distaste for the word `hookimpl` is well documented... There's quite a few places where it could be expanded

I went back and renamed everything I could.  
they snuck in there because, if I am writing the line "`for x in get_hookimpls()`, I'd like `x` to read as `hookimpl`... keeping with standard plural/singular python convention, and `get_hookimpls` can't be changed, that's a pluggy function.  this seems like one of the "downsides" of changing the convention internally.  in any case, they all now read `for hook_implementation in get_hookimpls()`

tlambert03:
I just pushed an update to `permute_hook_implementations`.  I removed one of the valid types in the `order` argument, which simplified some of the error checking, broke out a subfunction, and added more step-by-step clarification.

napari/_qt/qt_main_window.py
```diff
@@ -190,6 +192,22 @@ def _add_window_menu(self):
...
+        self.plugins_menu.addAction(order_plugin_action)
+
+    def _show_plugin_sorter(self):
+        plugin_sorter = QtPluginSorter(parent=self._qt_window)
+        dw = self.add_dock_widget(
+            plugin_sorter, name='Plugin Sorter', area="right"
+        )
+        plugin_sorter.finished.connect(dw.close)
+        plugin_sorter.finished.connect(plugin_sorter.deleteLater)
+        plugin_sorter.finished.connect(dw.deleteLater)
```
jni: I'd rename `dw` to `dock_widget`. It's only very local, but the line `plugin_sorter.finished.connect(dw.close)` would read a lot more nicely as `plugin_sorter.finished.connect(dock_widget.close)`

napari/_qt/qt_plugin_list.py
```diff
@@ -0,0 +1,303 @@
...
+        self.enabled_checkbox = QCheckBox(self)
+        self.enabled_checkbox.setToolTip("Uncheck to disable this plugin")
+        self.enabled_checkbox.stateChanged.connect(self._set_enabled)
+        self.enabled_checkbox.setChecked(
+            getattr(item.hook_implementation, 'enabled', True)
+        )
+        layout.addWidget(self.position_label)
+        layout.addWidget(self.enabled_checkbox)
+        layout.addWidget(self.plugin_name_label)
+        layout.setStretch(2, 1)
```
jni, line 71: Could you use keyword arguments here? It's unclear what "setStretch(2, 1)" could possibly mean.

napari/_qt/qt_plugin_list.py
```diff
@@ -0,0 +1,303 @@
...
+        layout.setContentsMargins(0, 0, 0, 0)
+
+    def _set_enabled(self, state: Union[bool, int]):
+        """Set the enabled state of this hook implementation to ``state``."""
+        # "hook_implementation.enabled" is NOT a pluggy attribute...
+        # we are adding that to allow skipping of hook_implementations.
+        # see plugins.io.read_data_with_plugins() for an example
+        self.item.hook_implementation.enabled = bool(state)
```
jni, line 79: Just a side comment. Is this something that you think the pluggy developers might be interested in adding?

jni: /persuaded to add. ;)

napari/_qt/qt_plugin_list.py
```diff
@@ -0,0 +1,303 @@
...
+            None (i.e. no hooks shown)
+        """
+        self.clear()
+        self.hook_caller = hook_caller
+        if not hook_caller:
+            return
+        for hook_implementation in reversed(hook_caller.get_hookimpls()):
+            self.add_hook_implementation_to_list(hook_implementation)
+
+    def add_hook_implementation_to_list(self, hook_implementation: HookImpl):
```
jni: Should we rename this to append instead of add? To me that makes it clearer that this is happening at the end of the list. What do you think? And `to_list` is probably redundant.

napari/_qt/qt_plugin_list.py
```diff
@@ -0,0 +1,303 @@
...
+        ----------
+        order : list, optional
+            A HookOrderType list ... unused by this function, but here for ease
+            of signal connection, by default None.
+        """
+        position = self.item.listWidget().indexFromItem(self.item).row() + 1
+        self.position_label.setText(str(position))
+
+
+class QtHookImplementationListWidget(QListWidget):
```
jni, line 95: Do we have have a policy on how we deal with camelCase vs under_scores when adding functions to Qt subclasses? Whatever it is, we should put it in the contributing guide.

napari/_qt/qt_plugin_list.py
```diff
@@ -0,0 +1,303 @@
...
+        self.setWindowModality(Qt.NonModal)
+        self.plugin_manager = plugin_manager
+        self.layout = QVBoxLayout()
+        self.setLayout(self.layout)
+        self.hook_combo_box = QComboBox()
+        self.hook_combo_box.addItem(self.NULL_OPTION)
+
+        # populate the combo box with all of the hooks known by the plugin mngr
+        hooks = []
+        for name, hook_caller in vars(plugin_manager.hook).items():
```
jni: Could we name this attribute `hooks` instead of `hook`? It's weird to get a *bunch* of hooks from an attribute called `hook`, don't you think?

napari/_qt/qt_plugin_list.py
```diff
@@ -0,0 +1,303 @@
...
+
+        Parameters
+        ----------
+        hook : str
+            Name of the new hook specification to show.
+        """
+        if hook == self.NULL_OPTION:
+            hook_caller = None
+        else:
+            hook_caller = getattr(self.plugin_manager.hook, hook)
```
jni: I think this line would also read better if the attribute was `.hooks`...

napari/plugins/_tests/test_pluggy_utils.py
```diff
@@ -0,0 +1,112 @@
...
+
+import pluggy
+from napari.plugins.utils import permute_hook_implementations
+
+hookimpl = pluggy.HookimplMarker("dummy")
+
+
+def _get_call_order(hook_caller):
+    hookimpls = list(
+        reversed([i.plugin_name for i in hook_caller.get_hookimpls()])
```
jni: I'm pretty sure I've asked this again, so apologies, but can you remind me again why there's so many `reversed` calls scattered around? Why is plugin call or traverse order not in the order of the list always?

napari/plugins/_tests/test_pluggy_utils.py
```diff
@@ -0,0 +1,112 @@
...
+    )
+    hookimpls.extend([i.plugin_name for i in hook_caller._wrappers])
```
jni: What does this line do? What are `hook_caller._wrappers`, and why do they go in the same list as the hook implementations? (Which should be called hook implementations, btw ;)

napari/plugins/_tests/test_pluggy_utils.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+class Plugin_3:
+    @hookimpl
+    def myhook(self):
+        return "p3"
+
+
+class Plugin_W1:
+    @hookimpl(hookwrapper=True)
```
jni: This! What does this do? LOL

napari/plugins/_tests/test_pluggy_utils.py
```diff
@@ -0,0 +1,112 @@
...
+    assert pm.hook.myhook() == expected
+
+
+def test_permute_hook_implementations_raises(pm):
+    """Test that invalid calls to permute_hook_implementations raise errors."""
+
+    with pytest.raises(TypeError):
+        # both implementations are valid, but cannot mix types.
+        imp0 = pm.hook.myhook.get_hookimpls()[0]
+        permute_hook_implementations(pm.hook.myhook, [imp0, 'p2'])
```
jni: Does the comment mean that you can reorder hook implementations by passing the actual objects rather than their names? I don't see that in any other test though?

napari/plugins/manager.py
```diff
@@ -149,6 +149,48 @@ def _register_module(self, plugin_name: str, module_name: str) -> None:
         except Exception as exc:
             raise PluginRegistrationError(plugin_name, module_name) from exc
 
+    def set_implementation_enabled(
+        self, hook_spec: str, plugin_name: str, enabled: bool
+    ) -> None:
```
jni: I kinda dislike this API... It feels very C++. I would prefer something like `enable_implementation` and `disable_implementation`, that both use this function (which would be private). What do you think?

napari/plugins/utils.py
```diff
@@ -0,0 +1,179 @@
...
+    # get a list of all hook implementations in hook_caller
+    hook_implementations = hook_caller.get_hookimpls()
+    if len(order) > len(hook_implementations):
+        raise ValueError(
+            "too many values in order: "
+            f"({len(order)} > {len(hook_implementations)})"
+        )
+
+    # figure out what list type was entered for ``order`` (or raise error)
+    order_type = _guess_order_type(order)
```
jni: I'd call this `_validate_order_type`.

But, to be honest, why not allow both types arbitrarily, but homogenise to the actual objects, then you can forget about the names below?

napari/plugins/utils.py
jni: OK! So I finally get this. Mostly. =P

So the thing I wasn't grokking is that wrappers and nonwrappers are in different lists, but here the API is unified, and, I want to say, worse?, it also allows you to mix wrappers and nonwrappers in the same order call? Although I don't know because actually I just looked back at the test and it doesn't look like you test reordering the wrappers?

I can't come up with a useful way to use wrappers for our existing plugins, nor how to even write one. I presume the wrapper signature is not the same as the hookspec signature? Or is it?

My suggestion is that we should (a) rename this function `bring_plugins_to_front`, because that more clearly describes what happens with a partial list, and, more controversially, (b) remove all the support for `_wrappers` for now, adding a note to the function about this. I know it's a bit annoying, but I do think the complexity isn't worth it here, not until we have people wanting to reorder wrappers.

An alternative, if you really want to keep support for wrappers in, might be to have arguments `order=` and `wrapper_order=` and use one or the other, maybe through sub-functions. Is there a specific reason why these things should be on the same list? To me it really feels like putting a list of apples and oranges together, that are never together in the model in the first place.

What do you think?

jni (changes requested):
Ok! This is almost ready! Apologies for holding things back. I've made a few minor API suggestions, plus a more major one — I think we need to deal with the different wrapper and nonwrapper lists up-front, rather than use the same `order=` list for both. I've left a couple of suggestions for how to do this, but happy to hear alternatives also. At least now I'm on the same page about what we're talking about! :sweat_smile:

napari/plugins/utils.py
tlambert03: > I can't come up with a useful way to use wrappers for our existing plugins, nor how to even write one. I presume the wrapper signature is not the same as the hookspec signature? Or is it?

have you looked at the [pluggy docs](https://pluggy.readthedocs.io/en/latest/#wrappers) yet?  they are explained pretty well there.  They use the same signature as the hookspec, they cannot return results, they can only modify results using a specific API (`get_result()`/`force_result()`) provided by pluggy.  We do not use them yet, and they would likely never be used in a [`firstresult=True`](https://pluggy.readthedocs.io/en/latest/#first-result-only) hook like `napari_get_reader`.


> My suggestion is that we should (a) rename this function `bring_plugins_to_front`, because that more clearly describes what happens with a partial list, and, more controversially, (b) remove all the support for `_wrappers` for now, adding a note to the function about this. I know it's a bit annoying, but I do think the complexity isn't worth it here, not until we have people wanting to reorder wrappers.

since we're moving farther away from something that could be generalized as a pluggy PR, we can do anything we want. I will rename the function and remove support for wrappers until we use them later.

napari/_qt/qt_plugin_list.py
tlambert03: yes, I agree, but that's just [what it's called in pluggy](https://github.com/pytest-dev/pluggy/blob/6e8b6d4e8e0519e50af05b4e4b90b18b924bcbcc/src/pluggy/manager.py#L75).  We _could_ probably make some alias in our own plugin manager subclass, but I'd rather not keep changing every part of their API that we don't love, or we'll soon have a seperate "napari pluggy API" that will confuse anyone who has any experience with pluggy

napari/_qt/qt_plugin_list.py
tlambert03: i do too, but see above, that's a pluggy attribute. I don't think we should change it.

napari/plugins/_tests/test_pluggy_utils.py
tlambert03: see [this comment]( https://github.com/napari/napari/pull/1023/files/cd19161ac9ea86f17963ead631aae6eb3927b767#r392614235).  Adding a comment to the code

tlambert03:
something weird is happening with your last comment... I can't respond in this thread so I'm copying my response again here:
> I can't come up with a useful way to use wrappers for our existing plugins, nor how to even write one. I presume the wrapper signature is not the same as the hookspec signature? Or is it?

have you looked at the [pluggy docs](https://pluggy.readthedocs.io/en/latest/#wrappers) yet?  they are explained pretty well there.  They use the same signature as the hookspec, they cannot return results, they can only modify results using a specific API (`get_result()`/`force_result()`) provided by pluggy.  We do not use them yet, and they would likely never be used in a [`firstresult=True`](https://pluggy.readthedocs.io/en/latest/#first-result-only) hook like `napari_get_reader`.


> My suggestion is that we should (a) rename this function `bring_plugins_to_front`, because that more clearly describes what happens with a partial list, and, more controversially, (b) remove all the support for `_wrappers` for now, adding a note to the function about this. I know it's a bit annoying, but I do think the complexity isn't worth it here, not until we have people wanting to reorder wrappers.

since we're moving farther away from something that could be generalized as a pluggy PR, we can do anything we want. I will rename the function and remove support for wrappers until we use them later.

napari/plugins/_tests/test_pluggy_utils.py
tlambert03: see https://github.com/napari/napari/pull/1023#issuecomment-604996010 and [pluggy docs](https://pluggy.readthedocs.io/en/latest/#wrappers).

napari/_qt/qt_plugin_list.py
tlambert03: # 🤷‍♂️

napari/_qt/qt_plugin_list.py
tlambert03: created https://github.com/napari/napari/issues/1091

napari/_qt/qt_plugin_list.py
tlambert03: (we had a zulip side chat that convinced me to not be afraid to go our own way).  I added an alias `PluginManager.hooks -> PluginManager.hook`, and changed these lines to read `plugin_manager.hooks` as you prefered

napari/_qt/qt_plugin_list.py
tlambert03: sorry, tried but got `TypeError: setStretch() takes no keyword arguments`:
see [`QBoxLayout.setStretch`](https://doc.qt.io/qt-5/qboxlayout.html#setStretch)

napari/plugins/_tests/test_pluggy_utils.py
tlambert03: refactored the sorting method to ignore wrappers... so removed this line too.

napari/plugins/_tests/test_pluggy_utils.py
tlambert03: https://pluggy.readthedocs.io/en/latest/#wrappers

napari/plugins/_tests/test_pluggy_utils.py
tlambert03: yes, added tests

napari/plugins/manager.py
tlambert03: i added those methods (and moved the whole thing to the _HookCaller class), but probably won't be using them internally yet, since this function is mostly there to receive Qt-style signals that (for instance) return the check state of a checkbox...
it's clunkier to do `enable_implementation() if checked else disable_implementation()`
rather than `set_enabled(checked)`

napari/plugins/utils.py
tlambert03: gone now

tlambert03:
ok @jni, in addition to all the smaller requests, I've made a few notable changes:
- rather than a `permute_hook_implementations` function that accepts a `_HookCaller` as its first argument, I decided to subclass our own `_HookCaller` in `napari.plugins.manager` and added a new method `_HookCaller.bring_to_front()` which serves the same purpose.  The new syntax for call reordering is `plugin_manager.hook.hook_spec_name.bring_to_front(['plugin_name', ...])`
- I got rid of `hookwrapper` support which simplifies the logic of looking for hook implementations and sorting them.
- I added an alias: `plugin_manager.hooks == plugin_manager.hook`

ready for re-review

napari/_qt/qt_plugin_list.py
jni: Missed this originally. Blows my mind. LOL Thanks for adding the code comments! I expect they will come in handy in the future. =)

jni (approved):

PR #1024, opened 2020-03-05
Plugin exception storage and developer notification GUI

tlambert03:
# Description
This PR improves the framework for gathering and reporting exceptions that result from plugins. 

- modifies `PluginError` class to automatically stores any errors that occur during the session. (anytime a `PluginError` or any of its subclasses gets instantiated, the instance will be added to the `napari.plugins.exceptions.PLUGIN_ERRORS` dict)
- centralizes convenience methods for formatting exceptions as text, or colored html
- provides a GUI that users can use to easily review and report helpful tracebacks to the developers (if a github URL is detected, a button is provided that, when clicked, will open the issue tracker at the corresponding plugin, pre-populated with traceback info).

Here's the basic idea:

Untitled

**one stylistic question:** Do you like the checkbox that gives the option of showing plugins that did _not_ generate errors (selecting them would simply show a blank textArea)?  Or will it never be used in this window, and should I just remove it to simplify things?

## Type of change

# How has this been tested?

If you'd like to test this locally, you can use the plugin fixtures to generate some errors:

```sh
$ export PYTHONPATH=napari/plugins/_tests/fixtures/
$ python -m napari
# then go to Plugins > Plugin Errors... "invalid" will have a button to open at github
```

## Final checklist:


tlambert03:
OK, updated this since #937, and filled out all the docstrings and stuff.  I edited the original description with a summary of the PR, one stylistic question, and a suggestion for local testing.  Ready for review.

sofroniewn:
This is pretty cool @tlambert03! I have a couple questions - one is how does it relate to any of the error handling planned in #794?

I love the styles, error message coloring etc, the copy to clip board button. 

> one stylistic question: Do you like the checkbox that gives the option of showing plugins that did not generate errors (selecting them would simply show a blank textArea)? Or will it never be used in this window, and should I just remove it to simplify things?

I'm ok with it - but then we should show something like `No errors detected` instead of just a blank textArea which is more confusing.

For plugins that don't have GitHub it seems like the open in github button doesn't appear, which is good. Can the url, and maybe email be displayed up at the top too in this block? Or is that a bit much? Would be nice to know the name of the github repo so I could go there myself etc.
```
=========================== Errors for plugin 'invalid' ===========================

napari version: 0.2.12+19.g660f9396
plugin name: invalid
version: 0.1.0
module: invalid_plugin
```
Also curious what happens if someone doesn't have a github account and they click the open in github button, does it error? tell them they need to open an account?

also how likely is this functionality breaks if github updates their website?

All-in-all though pretty cool! Maybe @ttung can give this a look over too and @jni can review also

tlambert03:
> This is pretty cool @tlambert03! I have a couple questions - one is how does it relate to any of the error handling planned in #794?

Nothing directly related, but the mechanism here for turning a python traceback into a nicely formatted html string could be used there as well.  That PR is more to provide GUI integration for error messages (so that someone not looking at the terminal would understand that an error happened, and _maybe_ be able to understand why).  This PR adds a _cache_ all exceptions of a specific type (`PluginError`), and a window dedicated for reviewing those errors.  It's conceivable we might also want some of that pattern in #794, to provide people with an easier mechanism for reporting bugs to us.  (note, this is _not_ automatic bug reporting, the user needs to click a button _every_ time they want to send an error).

> I'm ok with [the checkbox] - but then we should show something like `No errors detected` instead of just a blank textArea which is more confusing.

sounds good.  For the record, I'm also totally fine removing it if anyone finds in superfluous.

> For plugins that don't have GitHub it seems like the open in github button doesn't appear, which is good. Can the url, and maybe email be displayed up at the top too in this block? Or is that a bit much? Would be nice to know the name of the github repo so I could go there myself etc.

I like that idea, I'll add a bit more along those lines.

> Also curious what happens if someone doesn't have a github account and they click the open in github button, does it error? tell them they need to open an account?

good question!  I just tested it out: if someone isn't logged in, it will redirect them to a login page, and once they log in the body of the page is preserved.  (if they have to go all the way through the log in, who knows... but they can always come back).  _however_ there seems to be a character limit of exactly 5910 characters in the body of the message when not logged in (but if logged in, it's fine).  In any case, we probably need to have that email link as a backup plan just in case.

> also how likely is this functionality breaks if github updates their website?

always possible, but that GET request pattern `&body=....` is so common that I can't imagine them changing it.  They could change their URLs altogether.  But that all falls in the broad camp of depending on some external API, which can always change, and then we update...

napari/plugins/_tests/fixtures/invalid_plugin-0.1.0.dist-info/METADATA
```diff
@@ -2,6 +2,6 @@ Metadata-Version: 2.1
 Name: invalid
 Version: 0.1.0
 Maintainer-Email: example@invalid.com
-Home-Page: https://www.example.com
+Home-Page: https://github.com/napari/napari
```
jni: Should we maybe redirect to a nonsense URL? :grimacing:

napari/plugins/exceptions.py
```diff
@@ -1,23 +1,251 @@
+import sys
+from collections import defaultdict
+from types import TracebackType
+from typing import DefaultDict, Dict, List, Optional, Tuple, Type
+import re
+import IPython.core.ultratb
+
+
+PLUGIN_ERRORS: DefaultDict[str, List['PluginError']] = defaultdict(list)
```
jni, line 18: Could you add a comment about how this dictionary is structured? What is it supposed to hold exactly?

napari/plugins/exceptions.py
```diff
@@ -1,23 +1,251 @@
...
+        A formatted string with traceback information for every exception
+        raised by ``plugin_name`` during this session.
+    """
+    from napari import __version__
+
+    if not PLUGIN_ERRORS.get(plugin_name):
+        return ''
+
+    color = 'Linux' if as_html else 'NoColor'
+    vbtb = IPython.core.ultratb.VerboseTB(color_scheme=color)
```
jni: We are discussing in other issues making IPython an optional dependency, so I wonder if we can vendor this particular class to avoid depending on all of IPython just to colour tracebacks?

napari/plugins/exceptions.py
```diff
@@ -1,23 +1,251 @@
...
+        if package_meta:
+            msg.extend(
+                [
+                    f'{"plugin name": >16}: {package_meta["name"]}',
+                    f'{"version": >16}: {package_meta["version"]}',
+                    f'{"module": >16}: {err0.plugin_module}',
+                ]
+            )
+    except Exception:
+        pass
```
jni: What exactly are the things that can go wrong here?

napari/plugins/exceptions.py
```diff
@@ -1,23 +1,251 @@
...
+            html = (
+                "<span style='font-family: monaco,courier,monospace;'>"
+                + html
+                + "</span>"
+            )
+            msg.append(html)
+        else:
+            msg.append(vbtb.text(etype, value, tb))
+
+    msg += ['=' * 80]
```
jni: `msg.append('=' * 80)`?

jni: or better:

```
msg.append('=' * _linewidth)
```

napari/plugins/io.py
```diff
@@ -16,7 +18,7 @@ def read_data_with_plugins(
     returned, or no readers are found.
 
     Exceptions will be caught and stored as PluginErrors
-    (in plugin_manager._exceptions)
+    (in plugins.PLUGIN_ERRORS)
```
jni, line 21: I kinda prefer this being in the plugin manager... This is a mild preference, but could you tell us why you decided to make it a global variable instead?

jni (commented):
@tlambert03 made just a few requests for clarification, but this looks essentially ready to me. =)

napari/plugins/_tests/fixtures/invalid_plugin-0.1.0.dist-info/METADATA
tlambert03: i wanted at least one of them to be a valid github url so that we can test the "submit issue at github" button.  I'll change it to my personal fork so we don't get any accidental issues.

napari/plugins/exceptions.py
tlambert03: added comment
```
# This is a mapping of plugin_name -> PluginError instances
# all PluginErrors get added to this in PluginError.__init__
```

napari/plugins/exceptions.py
tlambert03: at first glance, it looks like `VerboseTB` uses quite a bit of other things in IPython.  so, can we leave that for a later PR when we actually remove the dependency?  worst case scenario, we can just have them be uncolored...

napari/plugins/exceptions.py
tlambert03: I never actually saw any exceptions, so it's hard to say.  The reason for this is that the _last_ thing I want to happen is that a real exception get buried due to some unexpected exception raised while trying to get package metadata just to make a nicer error message.

napari/plugins/io.py
tlambert03: I agree with your mild preference.  I did this mostly because I had lots of issues with circular imports...  also: anything can raise a `PluginError` (not just the plugin_manager), so it seemed ok to extract that concept.  I just briefly tried to put it back on the manager, and would need to rearrange a fair amount of things.  can figure it out though if you want it

tlambert03:
either addressed all issues, or commented inline.  have another look when ready.

napari/plugins/exceptions.py
sofroniewn: We made the import of IPython lazy in #1055 which sped up launch time quite a bit. Could we at least also make this import lazy here too? We could then make optional coloring in another PR

sofroniewn:
I'm trying to demo this again now but 
```
$ export PYTHONPATH=napari/plugins/_tests/fixtures/
$ python -m napari
```
doesn't show any errors. Is that expected?

tlambert03:
hmm... no, not expected, and it still works for me (when I type those things into the console, I see a lot of PluginError logs in the console, and I also see them in the Qt Window).  I'm not sure what to suggest... can you double check?  open a new terminal, make sure PYTHONPATH actually is getting set, and make sure `napari/plugins/_tests/fixtures/` has all the stuff in it?

sofroniewn:
@tlambert03 my bad - I was launching from the wrong directory. Everything looks good, my only remaining comment is the one I made above about making the IPython import lazy. Otherwise will be good to go

napari/plugins/io.py
jni: let's figure it out in a later PR if needed. =)

napari/_qt/qt_plugin_report.py
```diff
@@ -0,0 +1,186 @@
+"""Provides a QtPluginErrReporter that allows the user report plugin errors.
+"""
+import webbrowser
```
jni, line 3: Didn't know this existed! Awesome! =)

napari/_qt/qt_plugin_report.py
```diff
@@ -0,0 +1,186 @@
...
+
+        # Create plugin dropdown menu
+        self.plugin_combo = QComboBox()
+        self.plugin_combo.addItem(self.NULL_OPTION)
+        self.plugin_combo.addItems(list(sorted(set(PLUGIN_ERRORS))))
+        self.plugin_combo.currentTextChanged.connect(self.set_plugin)
+        self.plugin_combo.setCurrentText(self.NULL_OPTION)
+
+        # create github button (gets connected in self.set_plugin)
+        self.github_button = QPushButton('Open issue at github', self)
```
jni:
```suggestion
        self.github_button = QPushButton('Open issue on GitHub', self)
```

Incidentally, should we make this more platform-agnostic? The developer could be using GitLab or some other issue tracker...?

napari/_qt/qt_plugin_report.py
```diff
@@ -0,0 +1,186 @@
...
+        self.github_button.setToolTip(
+            "Open webrowser and submit this traceback\n"
```
jni:
```suggestion
            "Open a web browser to submit this error log\n"
```

(Users might not know what a "traceback" is...)

napari/_qt/qt_plugin_report.py
```diff
@@ -0,0 +1,186 @@
...
+            "to the developer's github issue tracker"
```
jni:
```suggestion
            "to the developer's GitHub issue tracker"
```

napari/_qt/qt_plugin_report.py
```diff
@@ -0,0 +1,186 @@
...
+        )
+        self.github_button.hide()
+
+        # create copy to clipboard button
+        self.clipboard_button = QPushButton()
+        self.clipboard_button.hide()
+        self.clipboard_button.setObjectName("QtCopyToClipboardButton")
+        self.clipboard_button.setToolTip("Copy traceback to clipboard")
```
jni:
```suggestion
        self.clipboard_button.setToolTip("Copy error log to clipboard")
```

napari/_qt/qt_plugin_report.py
jni: I guess, reading below (`.hide()`), that this is because each service will have a slightly different API, so it's hard for us to support all of them... Is that right? If so, I'm ok with that, but it might be interesting to allow plugin developers to provide a submit issue link template in the future.

napari/_qt/qt_plugin_report.py
```diff
@@ -0,0 +1,186 @@
...
+        self.plugin_meta.setText(meta_text)
+        if 'github.com' in meta.get('url', ''):
+
+            def onclick():
+                err = format_exceptions(plugin, as_html=False)
+                err = (
+                    "<!--Provide detail on the error here-->\n\n\n\n"
+                    "<details>\n<summary>Traceback from napari</summary>"
+                    f"\n\n```\n{err}\n```\n</details>"
+                )
```
jni, line 174: Not positive in this specific instance whether we should replace "traceback" with "error log" here. I think in this case it's ok to use traceback. But now I'm unsure about whether the nomenclature inconsistency becomes an issue.

napari/plugins/exceptions.py
```diff
@@ -1,23 +1,254 @@
...
         super().__init__(msg, plugin_name, plugin_module)
+
+
+def format_exceptions(plugin_name: str, as_html: bool = False) -> str:
+    """Return formatted tracebacks for all exceptions raised by plugin.
+
+    This method takes advantage of ``IPython.core.ultratb.VerboseTB`` to create
+    a nicely formatted traceback.  If ``as_html`` is True, is then converted
+    from ansi to html.  (This approach was used instead of cgitb because the
+    IPython ultratb tracebacks were preferred).
```
jni: What does this mean? Should it be rephrased as "because Talley preferred IPython ultratb tracebacks"? :joy:

napari/plugins/exceptions.py
jni: But that's exactly what's going to happen, no? All kinds of bugs might be hidden by having a bare except here. Personally, I think we should remove this until we actually observe such an exception. Then we can either fix the cause of that exception or make a specific catch, as appropriate.

napari/plugins/exceptions.py
jni: @tlambert03 can you also elaborate on why this is so much better than cgitb? Can one be swapped for the other reasonably easily? If so, then we could make IPython lazy and optional here and that would just be wonderful. =) But, you can choose to leave it as just lazy and we can leave the optional part for a later PR.

jni (changes requested):
@tlambert03 almost there!

napari/_qt/qt_plugin_report.py
tlambert03: yep.  this is all github specific.  we will need to figure out the API for any specific server we want to support. I'll open an issue after this merges.

napari/plugins/exceptions.py
tlambert03: hah, yeah I really didn't know how to write that one.   but I promise you, you prefer it too! ;)

napari/_qt/qt_plugin_report.py
tlambert03: this one is targeted at developers in their issue trackers...  i think it's fine, but I like swapping it out elsewhere.

tlambert03:
@jni I'm not sure why sometimes I can't respond to your comments inline... here they are:

> All kinds of bugs might be hidden by having a bare except here.

i'm not sure i follow.  there's basically just one function (`fetch_module_metadata`)  in that try block... it just adds a couple additional details to the log message.  so isn't the worst that can happen just that it fails to fetch a little additional detail?  That seems far less concerning to me than having the actual exception that we're actively trying to report _not_ be visible due to some silly additional error fetching the developer url, no?  That said, I also made `fetch_module_metadata` to catch known errors, so if you really want the try/catch out of there I'm ok with leaving it bare.

> can you also elaborate on why this is so much better than cgitb?

google image search shows a pretty accurate depiction of cgitb: 
https://www.google.com/search?q=cgitb&tbm=isch
color scheme doesn't match ours at all (and looks pretty abandoned), but more importantly, the ipython tb has lots of useful contextual info (like the value of variables).
It's not a simple swap out, but it's not _that_ hard either.  let me make ipython lazy here and let's deal with that later.

tlambert03:
I made the IPython import lazy, and also added a try/except to fall back to `cgitb` if `IPython` is not available... so this should be future proof.  I had to add additional code to make the cgitb output look presentable, and to handle exception chaining (e.g. "raise from")

here's the comparison, default on left, cgitb fallback on right:

combo

@jni the only remaining thing (I think) is for you to weigh in on the bare except after my last comments.  I think it should stay, but don't feel too strongly and happy to be overruled.

jni:
> so isn't the worst that can happen just that it fails to fetch a little additional detail?

The worst that can happen is that it fails *reliably and reproducibly* to fetch a little additional detail, and we *never know* because of the bare except. The worst that can happen if we remove it is that someone runs into such a bug, reports it, and then we fix it for everyone. (Ok that's a little bit disingenuous: they might not report it and stop using napari... But *someone* will report it!)

So, my preference is that we remove it — as you say, we don't have any expectation of anything breaking here, it's just paranoia. *IF* something breaks, we'll know about it and put in an *appropriate* exception handler there.

An alternative is to replace the pass with a print statement. I could live with that. =)

Awesome work re making IPython optional! 🎉 tbh I am not that bothered by the grey cgitb output — definitely glad it's not purple. =P So I am very very happy where it is now.

tlambert03:
yeah, fair enough.  I think the average napari user at this point will probably report it, or not care about it.  I'll remove the bare except.

tlambert03:
ok, bare `except` removed.  so i believe this is all set to go once we get an approving review.

sofroniewn (approved):
LGTM!! Thanks @tlambert03 - I will leave to @jni for the final review

jni (approved):

jni:
@tlambert03 now that #1023 is in, this needs a merge! But otherwise all good! 😁

PR #1027, opened 2020-03-06
Allow StringEnum to accept instances of self as argument

tlambert03:
# Description
I noticed a number of places where we want to coerce string values to a certain `StringEnum` type, so there was a lot of this pattern:
```python
if isinstance(value, str):
    value = StringEnum(value)
```
This PR modifies `StringEnumMeta` so that calling a subclass of a StringEnum with an instance of itself simply returns that instance, making the above equivalent to `value = StringEnum(value)`

This _might_ allow some of the hookups in #1020 to be even leaner.

## Type of change

# How has this been tested?

- [x] added tests for calling StringEnums with instances of themselves, and make sure it errors if the argument is neither a string nor an instance of that specific StringEnum.


kevinyamauchi:
Nice! This looks good to me! Were there more things you wanted to add? Otherwise, I'm happy to approve.

tlambert03:
nope, I think that's it for this one!  Just fixed the merge conflict...

kevinyamauchi (approved):
Looks good to me! Thanks for adding a test to demo the functionality!

sofroniewn (approved):
Beautiful - love it. Will let sit for a day and then merge

PR #1028, opened 2020-03-06
Close the canvas used to retrieve maximum texture size.

ttung:
# Description
Canvases start up an OpenGL context, which is shared since Qt 5.4 (see `Qt::AA_ShareOpenGLContexts` under https://doc.qt.io/qt-5/qt.html#ApplicationAttribute-enum).  Unfortunately, once the canvas is out of scope, the context could potentially be deallocated.  Subsequent code that fires up another canvas (i.e., any visualization) could potentially use the reference to the context's memory, but that memory could have been repurposed for other objects.  I've seen this happen fairly reliably when I add code that allocates a fair amount of memory, triggering the python garbage collector.

This fixes the problem by closing the canvas.  Subsequent canvases will spin up their own context.

## Type of change

# How has this been tested?

- [x] Test plan: Adding a dask client to each qt app fixture seems to trigger the crash fairly reliably (like 8 out of 10 times) with just `napari/_qt/_tests/test_qt_viewer.py`.  After this change, it doesn't seem to crash.

## Final checklist:

### DIFF
--- a/napari/_vispy/vispy_base_layer.py
+++ b/napari/_vispy/vispy_base_layer.py
@@ -1,7 +1,8 @@
-from vispy.gloo import gl
-from vispy.app import Canvas
-from vispy.visuals.transforms import STTransform
 from abc import ABC, abstractmethod
+from functools import lru_cache
+from vispy.app import Canvas
+from vispy.gloo import gl
+from vispy.visuals.transforms import STTransform
 
 
 class VispyBaseLayer(ABC):
@@ -208,6 +209,7 @@ class VispyBaseLayer(ABC):
         self.layer.scale_factor = self.scale_factor
 
 
+@lru_cache()
 def get_max_texture_sizes():
     """Get maximum texture sizes for 2D and 3D rendering.
 
@@ -219,8 +221,11 @@ def get_max_texture_sizes():
         Max texture size allowed by the vispy canvas during 2D rendering.
     """
     # A canvas must be created to access gl values
-    _ = Canvas(show=False)
-    MAX_TEXTURE_SIZE_2D = gl.glGetParameter(gl.GL_MAX_TEXTURE_SIZE)
+    c = Canvas(show=False)
+    try:
+        MAX_TEXTURE_SIZE_2D = gl.glGetParameter(gl.GL_MAX_TEXTURE_SIZE)
+    finally:
+        c.close()
     if MAX_TEXTURE_SIZE_2D == ():
         MAX_TEXTURE_SIZE_2D = None
     # vispy doesn't expose GL_MAX_3D_TEXTURE_SIZE so hard coding


sofroniewn:
Thanks @ttung! Looks like Flake8 is still failing, but otherwise this is great

ttung:
> Looks like Flake8 is still failing, but otherwise this is great

Sorry!  Fixed this.

sofroniewn (approved):
LGTM! thanks!!

sofroniewn:
@ttung the linux tests seem to be consistently failing. Can you tell if that is that something specific to this PR? I tried re-running them and they failed again

ttung:
Looks like it's getting OOM killed.  I don't know why that might happen, and there's no log... :(

tlambert03:
i think something funny is going on with cirrus at the moment. 
 getting strange stuff here too: https://github.com/napari/napari/pull/1014/checks?check_run_id=491484481

tlambert03:
> i think something funny is going on with cirrus at the moment.
> getting strange stuff here too: https://github.com/napari/napari/pull/1014/checks?check_run_id=491484481

i take it back... I had a legitimate issue in my pyqt5 tests that had nothing to do with cirrus.  So think the OOM error here is related to this PR somehow.

sofroniewn:
Yeah, this does now see specific to this PR as linux tests are passing elsewhere on the repo - I'm not sure about that `capabilities` syntax and what's going on there with the linux machines. Would be good to know if it passes locally on linux too. I think @tlambert03 that this is likely a vispy / qt test and we're also not running all of those on windows / macos right? (I can't remember), so it might be more of a vispy CI thing than a linux thing and the fact that windows and mac pass is only because we don't run the test there.

I can confirm all the tests pass locally for me on my mac

ttung:
I suspect that there's some memory leak with OpenGL contexts on Linux.  I wrapped the results for max texture size in a lru cache, and it seems to pass all CI.

tlambert03:
@ttung, with this new approach, `MAX_TEXTURE_SIZE_2D` always ends up being `None` for me.  (i.e. `c.context.shared.parser.capabilities["max_texture_size"]` is always empty).  Whereas with the older `gl.glGetParameter(gl.GL_MAX_TEXTURE_SIZE)` approach I always get some value around 16,384. Is that the expected result?

sofroniewn:
Unfortunately our test is basically skipped if the output is `None` - https://github.com/napari/napari/blob/d827636888d78b4d889008ec59524c4e68107481/napari/_vispy/_tests/test_vispy_big_images.py#L13 which isn't so great either, but I think was needed for the times when our CI wasn't able to make a canvas etc.

ttung:
> @ttung, with this new approach, MAX_TEXTURE_SIZE_2D always ends up being None for me. (i.e. c.context.shared.parser.capabilities["max_texture_size"] is always empty). Whereas with the older gl.glGetParameter(gl.GL_MAX_TEXTURE_SIZE) approach I always get some value around 16,384. Is that the expected result?

I moved back to the `glGetParameter` approach.  I could have sworn the other approach worked, but I can't reproduce it working.

sofroniewn:
Looks good to me now, thanks @ttung - I will merge promptly unless hearing otherwise

napari/_vispy/vispy_base_layer.py
```diff
@@ -208,6 +209,7 @@ def on_draw(self, event):
         self.layer.scale_factor = self.scale_factor
 
 
+@lru_cache()
```
tlambert03, line 212: TIL :)  nice to know this is in the standard lib!

tlambert03 (approved):

PR #1033, opened 2020-03-07
Refactor points layer mouse bindings

sofroniewn:
# Description
This PR is a continuation of #1010, though can be reviewed and merged independently. It does some similar renaming of files and refactoring of the mouse bindings using the very nice approach suggested by @kne42. It also adds fairly comprehensive tests for the mouse bindings. Maybe @kevinyamauchi can give this a spin and check everything behaves as expected. 

After this and #1010 goes in I'll extend the approach to Shapes, and I'll make a separate PR propagating the name changes across the rest of the repo

## Type of change

# References

Follow-on from #1010 a step towards closing #694.
Approach inspired heavily by https://github.com/kne42/napari/tree/refactor-internal-mouse-bindings

# How has this been tested?

- [x] `pytest napari/layers/points/_tests/test_points_mouse_bindings.py`

## Final checklist:


napari/layers/points/points.py
```diff
@@ -182,15 +185,15 @@ class Points(Layer):
 
     Extended Summary
     ----------
-    _data_view : array (M, 2)
+    _view_data : array (M, 2)
```
kevinyamauchi, line 188: Good catch, thanks!

napari/layers/points/_tests/test_points_mouse_bindings.py
```diff
@@ -0,0 +1,305 @@
...
+from napari.layers import Points
+from napari.utils.interactions import (
+    ReadOnlyWrapper,
+    mouse_press_callbacks,
+    mouse_move_callbacks,
+    mouse_release_callbacks,
+)
+
+
+def create_points_layer():
```
kevinyamauchi: What do you think about putting this in a util file? I think it would be nice to use this function in the other points tests as well (refactoring them can be a separate PR)?

napari/_vispy/vispy_points_layer.py
```diff
@@ -9,7 +9,7 @@
 
 class VispyPointsLayer(VispyBaseLayer):
     _highlight_color = (0, 0.6, 1)
-    _highlight_width = 5
+    _highlight_width = 2
```
kevinyamauchi, line 12: Probably a separate PR, but I think we should consider moving `_highlight_color` and `_highlight_width` somewhere else so it's easier for the user to change it "on the fly" (e.g., onto the model).

napari/layers/points/points.py
```diff
@@ -1098,16 +1101,27 @@ def mode(self, mode):
             return
         old_mode = self._mode
 
+        if old_mode == Mode.ADD:
+            self.mouse_drag_callbacks.remove(add)
+        elif old_mode == Mode.SELECT:
+            # add mouse drag and move callbacks
+            self.mouse_drag_callbacks.remove(select)
```
kevinyamauchi, line 1143: It might be nice to add a test that goes from `Mode.ADD` to `Mode.SELECT` (and vice versa) to make sure the callbacks are properly added/removed. I think the tests only go from `Mode.PAN_ZOOM` to `Mode.ADD` or `Mode.SELECT`.

kevinyamauchi (commented):
Thanks, @sofroniewn. Love the added tests! The callbacks seem to be working for me. I made a couple of small comments below, but otherwise this looks good to me.

napari/_vispy/vispy_points_layer.py
sofroniewn: Agreed - I've added `_highlight_width` to #992 ideas for setting. Highlight color should get automatically set by our theme. I've made a new issue for that #1035

napari/layers/points/_tests/test_points_mouse_bindings.py
sofroniewn: moved into `test_points_utils.py`, which already existed so it was easy. We can do more refactoring / reuse in followup PRs. Agreed that is something we can improve across the code base

napari/layers/points/points.py
sofroniewn: Good suggestion - two tests have been added, one for add and one for select. Thanks!

sofroniewn:
I have made the changes requested by @kevinyamauchi - this is ready for further review

napari/layers/points/_tests/test_points_utils.py
```diff
@@ -1,12 +1,29 @@
...
+    known_non_point = [20, 30]
+    n_points = len(data)
+
+    layer = Points(data, size=1)
+    assert np.all(layer.data == data)
+    assert layer.ndim == 2
+    assert len(layer.data) == n_points
+    assert len(layer.selected_data) == 0
+
+    return layer, n_points, first_point, known_non_point
```
kevinyamauchi: It might be more generally useful to return `data` instead of `first_point`. I think it's fine either way though.

napari/layers/points/_tests/test_points_utils.py
```diff
@@ -1,12 +1,29 @@
...
     dataframe_to_properties,
     guess_continuous,
 )
 
 
+def create_known_points_layer():
+    """Create points layer with known coordinates."""
+    data = [[1, 3], [8, 4], [10, 10], [15, 4]]
+    first_point = [1, 3]
+    known_non_point = [20, 30]
```
kevinyamauchi: I think it would be nice to have a comment explaining what `known_non_point` is. There could be enough in this function to warrant a full docstring explaining the returned variables.

kevinyamauchi (approved):
Thanks, @sofroniewn . This looks good to me! I just left a couple of minor nitpicks below.

napari/layers/points/_points_mouse_bindings.py
```diff
@@ -0,0 +1,95 @@
+import numpy as np
+
+from ._points_utils import points_in_box
+
+
+def select(layer, event):
```
kne42, line 6: can `layer._is_selecting`, `layer._drag_start`, and `layer._drag_box` be made local to this function?

napari/layers/points/_tests/test_points_utils.py
sofroniewn: i've dropped `first_point` and now use `layer.data` instead, it is more clear

napari/layers/points/_tests/test_points_utils.py
sofroniewn: good point, doc string added

napari/layers/points/_points_mouse_bindings.py
sofroniewn: unfortunately right now they are all also used inside the main model file, either in the `_move` function, which could probably be refactored into this file or inside some of the highlighting stuff. Maybe we leave this for now, and possibly revisit after we're done some of the interaction box stuff in #999

sofroniewn:
I have addressed the comments from @kevinyamauchi. I thing the further refactoring @kne42 mentions is best saved for future PRs. This is now ready for rereview @jni

kne42 (approved):

napari/layers/points/_points_mouse_bindings.py
```diff
@@ -0,0 +1,95 @@
...
+    Holding shift throughout the entirety of this process will add those points
+    to any existing selection, otherwise these will become the only selected
+    points.
+    """
+    # on press
+    shift = 'Shift' in event.modifiers
+
+    # if shift add / remove any from existing selection
+    if shift:
+        if layer._value is not None:
```
jni, line 32: How is `layer._value` created/populated? It's unclear in this context where it comes from. From the context, it's the index of a point? Should it be renamed `layer._point_index`, for example?

napari/layers/points/_points_mouse_bindings.py
```diff
@@ -0,0 +1,95 @@
...
+            if layer._value in layer.selected_data:
+                layer.selected_data = [
+                    x for x in layer.selected_data if x != layer._value
+                ]
+            else:
+                layer.selected_data += [layer._value]
+        else:
+            pass
```
jni: Also, this whole clause (`if shift:`) is quite hard to read for something simple: `_toggle_selected(layer.selected_data, layer._value)`. Also, since `layer.selected_data` is a list, the removing bit should be `layer.selected_data.remove(layer._value)`, and the adding bit should be `layer.append(layer._value)`. It might also be better to use a set for this, rather than a list...? Set add/remove is O(1), whereas list remove is O(n).

napari/layers/points/_points_mouse_bindings.py
```diff
@@ -0,0 +1,95 @@
...
+    currently selected points.
+
+    Clicking and dragging on an empty part of the canvas (i.e. not on a point)
+    will create a drag box that will select all points inside it when finished.
+    Holding shift throughout the entirety of this process will add those points
+    to any existing selection, otherwise these will become the only selected
+    points.
+    """
+    # on press
+    shift = 'Shift' in event.modifiers
```
jni: Q: should we allow *both* shift and cmd(mac)/ctrl(other) as "change selection" modifiers? Both would be expected by users.

napari/layers/points/_points_mouse_bindings.py
```diff
@@ -0,0 +1,95 @@
...
+        else:
+            pass
+    else:
+        if layer._value is not None:
+            if layer._value not in layer.selected_data:
+                layer.selected_data = [layer._value]
+        else:
+            layer.selected_data = []
+    layer._set_highlight()
+    yield
```
jni, line 47: Add a space above the yield?

napari/layers/points/_points_utils.py
```diff
@@ -4,6 +4,80 @@
...
+    -------
+    inside : list
+        Indices of points inside the box
+    """
+    box = create_box(corners)[[0, 2]]
+    rect = points_to_squares(points, sizes)
+    below_top = np.all(box[1] >= rect, axis=1)
+    above_bottom = np.all(rect >= box[0], axis=1)
+    inside = np.logical_and(below_top, above_bottom)
+    inside = np.unique(np.where(inside)[0] % len(points))
```
jni: Is the `% len(points)` required???

jni: (If so, please add a comment as to why!)

napari/layers/points/_points_utils.py
```diff
@@ -4,6 +4,80 @@
...
+    sizes : (N,) array
+        Size of each point
+
+    Returns
+    -------
+    inside : list
+        Indices of points inside the box
+    """
+    box = create_box(corners)[[0, 2]]
+    rect = points_to_squares(points, sizes)
```
jni: So do I have to drag around the full point to select it? Honestly I think I prefer for the centers to be the selection point...? Any comments on this?

napari/layers/points/_tests/test_points_mouse_bindings.py
```diff
@@ -0,0 +1,359 @@
...
+from napari.layers import Points
+from napari.utils.interactions import (
+    ReadOnlyWrapper,
+    mouse_press_callbacks,
+    mouse_move_callbacks,
+    mouse_release_callbacks,
+)
+
+
+def create_known_points_layer():
```
jni, line 29: Probably this should be a [pytest fixture](https://docs.pytest.org/en/latest/fixture.html)?

napari/layers/points/points.py
```diff
@@ -1098,16 +1101,27 @@ def mode(self, mode):
...
+            self.mouse_move_callbacks.remove(highlight)
```
jni, line 1144: Same comments as with the labels layer: it would be good to use a dictionary, or some kind of stack, so that you could remove the keybindings from the previous mode change, without an if/else cascade that goes through all the modes — which I expect will be super painful for Shapes!

jni (changes requested):
@sofroniewn I've left some requests for clarification, as well as a few suggestions for alternate behaviour/implementation.

napari/layers/points/_points_mouse_bindings.py
sofroniewn: done

napari/layers/points/_points_mouse_bindings.py
sofroniewn: This gets set inside the base layer here https://github.com/napari/napari/blob/51cedac63cb166c2851f10a28a86461aca43f635/napari/layers/base/base.py#L619. As it is in base it has a generic name, in the image layer the `_value` property is the pixel value. It get's used for things like the status bar. Agreed maybe this should be refactored in a future PR, but I wouldn't change here

napari/layers/points/_points_mouse_bindings.py
sofroniewn: interesting, this is not the case for me in illustrator when drawing a rectangle. Only shift locks aspect ratio. I'd probably rather not have two modifiers do the same thing, we might want to keep the other for different behavior, like resize from center. Is that ok?

sofroniewn: Unless I am missing something here

napari/layers/points/_points_utils.py
sofroniewn: yes, it is because the previous line refers to corner indices but we want things in points indices, I have added comments to explain

napari/layers/points/_points_utils.py
sofroniewn: No if any corner is inside then it is included, I find this behaviour much more visually intuitive then either must contain the center or must contain the whole shape and it becomes very clear when drawing the bounding box what will and what will not be highlighted. I have added comments to explain

napari/layers/points/_tests/test_points_mouse_bindings.py
sofroniewn: done

napari/layers/points/points.py
sofroniewn: As discussed in the labels review, I'd rather this wait for the refactor in #1048

napari/layers/points/_points_mouse_bindings.py
sofroniewn: so I made it a set, good idea, but if I use the `add` / `remove` methods of the set I still need a line like `layer.selected_data = layer._selected_data` to trigger the code inside the selected_data setter, which isn't so great, but maybe it's ok. Curious what you recommend

sofroniewn:
This has been updated and is now ready for re-review @jni

napari/layers/points/_points_mouse_bindings.py
sofroniewn: Curious what @kevinyamauchi thinks on all this too, both the `list` -> `set` transition and the `add` / `remove` stuff.

napari/layers/points/_points_mouse_bindings.py
jni: wait, this line is not about locking aspect ratio, it's about selecting points when clicking on them, right?

sofroniewn: ooooh right, yeah, i can fix then

napari/layers/points/_points_utils.py
jni: For reference, my preference is to have the centers dictate selection, because it may otherwise be impossible to select only some points if the point diameter is too large. But, this battle can wait for another time, thanks to your comment. =)

jni: My preference is doubly so since this way you may select a point with a circle marker without even touching the point. =\

napari/_vispy/vispy_points_layer.py
jni: I think theme is the right place for this, rather than the model

napari/layers/points/_points_mouse_bindings.py
sofroniewn: fixed

napari/layers/points/_points_mouse_bindings.py
jni: Maybe add a comment above this line, "layer._value is defined in the base layer and contains the value under the cursor. For points, this is the index of the highlighted point."

napari/layers/points/_points_utils.py
sofroniewn: this i feel we could settle with some usability testing - if most people want this i'd change it for sure, but i feel it could wait till then

napari/layers/points/_points_mouse_bindings.py
sofroniewn: comment added

napari/layers/points/_points_mouse_bindings.py
jni: `else: pass` is superfluous. So I would say:

```
if modify_selection:
    if layer._value is not None:
        _toggle_selected(layer.selected_data, layer._value)
```

and define the above function. What do you think?

napari/layers/points/_points_mouse_bindings.py
```diff
@@ -0,0 +1,100 @@
...
+            if layer._value in layer.selected_data:
+                layer.selected_data.remove(layer._value)
+            else:
+                layer.selected_data.add(layer._value)
+            layer.selected_data = layer._selected_data
+        else:
+            pass
+    else:
+        if layer._value is not None:
+            if layer._value not in layer.selected_data:
```
jni, line 41: Can we remove this if? I think if I click without shift I want to select that point and that point only. (This is what happens in Keynote at least.)

napari/layers/points/_points_mouse_bindings.py
sofroniewn: done

napari/layers/points/_points_mouse_bindings.py
sofroniewn: This is needed if you want to drag many points that are currently selected. If you take it away then you can only ever drag one at a time. It makes sense to me to leave, but I could drop if I understood the behavior we're trying to avoid better. Again this is something that could wait for more detailed usability testing

sofroniewn: I have added a comment to this effect

sofroniewn:
@jni I added the `_toggle_selected` function, which does clean up the code nicely I think. I havn't changed the click selection line you flagged as it is needed for dragging many points at the same time, but I did add a comment to that effect. This is ready for re-review

napari/layers/points/_points_mouse_bindings.py
jni: Great, thanks! I was thinking about click, not drag.

jni (approved):

PR #1034, opened 2020-03-08
Fix points color cycle refresh

### LINKED ISSUE #988: Toggle point class property fails if they are all in the same class
## 🐛 Bug

When running the example for adding points with properties. If the 'good_point' property in the [line](https://github.com/napari/napari/blob/f53d7448f2858a36ff7a57bcfc5176018f2af670/examples/add_points_with_properties.py#L21) is homogeneous for all points (all 'True' or all 'False'), then the toggling fails. Is that expected behavior?

## To Reproduce

1. Set `good_point': np.array([True, True, True])`
2. Run the example
3. Select point, press 't' and nothing happens

## Expected behavior

1. Set `good_point': np.array([True, True, True])`
2. Run the example
3. Select point, press 't' the point should be toggling from green to blue

## Environment

napari: 0.2.11
Platform: Darwin-18.6.0-x86_64-i386-64bit
Python: 3.7.6 (default, Dec 30 2019, 19:38:28) [Clang 11.0.0 (clang-1100.0.33.16)]
Qt: 5.13.2
PySide2: 5.13.2
VisPy: 0.6.4
NumPy: 1.18.1
SciPy: 1.4.1
scikit-image: 0.16.2
Dask: 2.11.0

GL version: 2.1 ATI-2.9.26
MAX_TEXTURE_SIZE: 16384

kevinyamauchi:
# Description
This PR fixes a bug where adding values to properties used to set an edge/face color cycle and then calling refresh caused a `KeyError` (described in #988). The bug is fixed by adding any property values not in the color cycle map when refreshing the edge/face colors.

## Type of change

# References
Fixes #988 

# How has this been tested?
- [x] added tests to verify property values can be added when refreshing color cycle properties

## Final checklist:

### DIFF
--- a/napari/layers/points/_tests/test_points.py
+++ b/napari/layers/points/_tests/test_points.py
@@ -598,6 +598,29 @@ def test_edge_color_cycle():
     )
 
 
+def test_adding_value_edge_color_cycle():
+    shape = (10, 2)
+    np.random.seed(0)
+    data = 20 * np.random.random(shape)
+    annotations = {'point_type': np.array(['A', 'B'] * int((shape[0] / 2)))}
+    color_cycle = ['red', 'blue']
+    layer = Points(
+        data,
+        properties=annotations,
+        edge_color='point_type',
+        edge_color_cycle=color_cycle,
+    )
+
+    # make point 0 point_type C
+    point_types = layer.properties['point_type']
+    point_types[0] = 'C'
+    layer.properties['point_type'] = point_types
+    layer.refresh_colors(update_color_mapping=False)
+
+    edge_color_map_keys = [*layer.edge_color_cycle_map]
+    assert 'C' in edge_color_map_keys
+
+
 def test_edge_color_colormap():
     # create Points using with face_color colormap
     shape = (10, 2)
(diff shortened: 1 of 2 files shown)


sofroniewn (approved):
I have tested this works, and the code looks good and test looks good - but I wonder @kevinyamauchi if it is now possible to grant the request of @jni in https://github.com/napari/napari/issues/988#issuecomment-589482806

> I like your solution, @kevinyamauchi. Bonus points, I feel like we can leave update_color_map as default False now!

Or is there a reason you still want the default to be `True`?

kevinyamauchi:
>> I like your solution, @kevinyamauchi. Bonus points, I feel like we can leave update_color_map as default False now!
>
> Or is there a reason you still want the default to be True?

Ah dang, I missed that. I think it makes sense to change and I can change it tomorrow.

kevinyamauchi:
I have updated the default args for `Points.refresh_colors()`, as requested by @jni!

sofroniewn:
Thanks @kevinyamauchi this looks good to me! I will let sit till tomorrow in case @jni or anyone else wants to weigh in, then I will merge

napari/layers/points/_tests/test_points.py
```diff
@@ -598,6 +598,35 @@ def test_edge_color_cycle():
     )
 
 
+def test_adding_value_edge_color_cycle():
+    """ Test that adding values to properties used to set an edge color cycle
+    and then calling Points.refresh_colors() performs the update and adds the
+    new value to the edge_color_cycle_map.
+
+    See: https://github.com/napari/napari/pull/1034
```
jni:
```suggestion
    See: https://github.com/napari/napari/issues/988
```

jni: I think it's good here to point to the issue that it's testing, rather than the PR that added the code — which will be visible with `git blame` anyway.

napari/layers/points/_tests/test_points.py
```diff
@@ -770,6 +799,35 @@ def test_face_color_cycle():
     )
 
 
+def test_adding_value_face_color_cycle():
+    """ Test that adding values to properties used to set an face color cycle
+    and then calling Points.refresh_colors() performs the update and adds the
+    new value to the face_color_cycle_map.
+
+    See: https://github.com/napari/napari/pull/1034
```
jni:
```suggestion
    See: https://github.com/napari/napari/issues/988
```
as above

jni (approved):
:+1:! Just left a couple of super minor doc suggestions. LGTM otherwise! =)

jni:
Ooh and I just saw that I could commit them all by myself. =) Just waiting on CI to merge. =)

PR #1017, opened 2020-03-02
Style refactor

tlambert03:
# Description
When creating new dialogs and widgets, I have often noticed the need to add widget-specific styles (e.g. padding around a `QPushButton`, background on a `QLabel` or something like that).  This makes it difficult to standardize the look & feel of widgets within napari, and will also make it hard for others to extend the napari GUI (e.g. with `magicgui`) without needing to tweak their own stylesheets.  This PR refactors our stylesheets (and other style related things) with the aim of standardizing styles without needing widget-specific styles (for instance, a "spinbox" in napari, should look like all the other ones).  I'm mostly done, and will remove [WIP] when it's ready for review.

**This PR:**
- adds a `SampleWidget` in `_qt/theme_sample.py` that contains most other widget types, which could be used to show a theme preview or test out a new theme.
- breaks up `stylesheet.qss` into multiple cascading sheets for easier organization.
    - base.qss refers only to built-in Qt types, and establishes the base themed style.
    - other stylesheets in there are used to customize napari-specific objects.
- removes as many duplicate/conflicting styles as possible.
- adds two new `darken`/`lighten` "directives" that can be used in stylesheets.  For instance the following will create a color that is 15 percent darker than `{{ background }}`.
```css
QLineEdit {
  background-color: {{ darken(background, 15) }};
}
```
- overhauls `build_icons.py` to include _all_ required steps for adding an icon.  Now adding an icon simply requires putting the svg file in `/icons/svgs` and running `python -m napari.resources.build_icons`.  `CONTRIBUTING.md` has been updated with new instructions.

before after comparison of the test Widget:
before_after

## Type of change

- [x] Bug-fix (non-breaking change which fixes an issue)

# How has this been tested?
very little functional code has been touched.

## Final checklist:
- [ ] My PR is the minimum possible work for the desired functionality
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #1037, opened 2020-03-08
Fix styles for Qt < 5.12.  Fix styles in manifest for pip install

### LINKED ISSUE #1031: QtModal.show_at doesn't work on Qt < 5.10
## 🐛 Bug
@fenixnano reported in https://github.com/napari/napari/pull/837#issuecomment-595508738 that the clims slider right click didn't work, using Qt 5.9.6.

the `QGuiApplication.screenAt` method used there was [introduced in 5.10.0](https://doc.qt.io/qt-5/qguiapplication.html#screenAt) so a different solution would be needed if we are to support older versions of Qt.

@Volker, perhaps you can also comment more here on the dual-monitor bug (I don't have a good setup to test that).  The code is _supposed_ to pick the screen where the mouse is at the moment of right click...

### LINKED ISSUE #1032: Style refactor destroyed styles on Qt 5.9.3
## 🐛 Bug
I just realized that #1017 seems to have totally broken styles on Qt 5.9.3.  I know we technically only support >= 5.12  ... but given that it use to at least _kind of_ work on 5.9.3, I think we should figure that out before including #1017 in the next release.

will work on it

tlambert03:
# Description
Fixes #1031 and Fixes #1032. Let me know if you want me to split this into two PRs.  It started as a fix for styles on older (≤5.12) versions of Qt, and during testing I also realized that MANIFEST.ini needed to be updated for #1017. 

This PR requires some scrutiny and feedback: as it changes the way we distribute resources and styles.

1. The main change here is that `res.qrc`, `qt.py` and all the theme specific icons that were generated in `build_icons` are no longer checked into source.  Nor is it ever required to run `build_icons` from the command line. Rather, they are built _after_ installation, the first time napari is run (and can be rebuilt on demand by setting an environmental variable).  The reasoning for this was that I found some cases where the `qt.py` autogenerated by a newer version of pyside2 5.14 was _not_ backwards compatible with an older version of pyqt5 5.9... so this just lets the user's environment dictate how the icons are built.  
  
    **my main question:**, this approach is nice for it's environmental flexibility, but it does assume that napari will be able to _write files_ to its final install location.  I have seen other packages do this sort of thing (`pims` does it for instance), but I'm not entirely sure how safe that is. @jni, @kne42 ?  We may also need to make special accomodations for bundled apps (#496).

2. I have updated `MANIFEST.ini` for pip distribution (and `.gitignore`) accordingly.  _NOTE:_ if we choose not to merge this PR, we still need to fix distribution of `qss` files in `MANIFEST.ini` in `master` before the next release.

3. `qlineargradient` is now only used in stylesheets if the `Qt` version is ≥ 5.12 (this is done using a new `vgradient`/`hgradient` directive in `theme.py`).

## Type of change

## Testing

I tested this approach using `python setup.py sdist` and installing the distribution into a fresh local environment with various pyqt and pyside2 installations all the way back to v5.7


.gitignore
```diff
@@ -117,3 +117,12 @@ venv.bak/
 
 # VSCode
 .vscode/
+
+# these will get autogenerated not
```
jni: not...? :joy:

jni:
@tlambert03 I very much like the gist of this PR. Still haven't gone through the whole code, some thoughts though:

(1) How expensive is this? Does it destroy the user experience to do it on every boot-up if the directory isn't writeable? In that case, temporary files could be used, for example.
(2) Could we write to `~/.config/napari/qt_resources` (or whatever) if the install directory isn't writeable?
(3) Should we match the resources to the napari version? ie if I work from my main git checkout, I probably want to rebuild every time that checkout changes, right?

A more minor comment, why is `_STYLESHEET` used as a global variable within `get_stylesheet`, instead of having `_STYLESHEET = _generate_stylesheet()` and then `get_stylesheet()` return that module variable?

tlambert03:
> (1) How expensive is this? Does it destroy the user experience to do it on every boot-up if the directory isn't writeable? In that case, temporary files could be used, for example.

On my computer, it takes about 280±20 ms  to rebuild all the resources.

> (2) Could we write to `~/.config/napari/qt_resources` (or whatever) if the install directory isn't writeable?

could certainly try.  one relevant detail here though: while doing this I tried to load the generated `qt.py` file from a path using `importlib.util.spec_from_file_location` as described in [this post](https://stackoverflow.com/questions/67631/how-to-import-a-module-given-the-full-path)... but I couldn't get it to work.  I always got a segfault.  So I resorted to writing the resource file to `napari/resources/_qt.py` and then just calling `from . import _qt`.  I don't understand yet why the `spec_from_file` approach didn't work, but that will probably be required for a tempfile not on `sys.path`

> (3) Should we match the resources to the napari version? ie if I work from my main git checkout, I probably want to rebuild every time that checkout changes, right?

might be a little overkill (but probably harmless).  if you're not actively changing either the icons themselves or the versions of Qt you're using, the same rendered file is probably fine.  But since it's also relatively cheap, it would be fine to put some sort of version hash in the autogenerated file name and re-render if missing.

> A more minor comment, why is `_STYLESHEET` used as a global variable within `get_stylesheet`, instead of having `_STYLESHEET = _generate_stylesheet()` and then `get_stylesheet()` return that module variable?

the goal was to only combine those files once... and I'm still developing my functional skills 😂.  However, I just learned about [`lru_cache`](https://docs.python.org/3/library/functools.html#functools.lru_cache) from Tony's recent PR... so I'll change it to that instead.

jni:
> it takes about 280±20 ms to rebuild all the resources

Paying this cost once per startup *only* when the install isn't writeable seems acceptable to me.

>  one relevant detail here though: while doing this I tried to load the generated qt.py file from a path using importlib.util.spec_from_file_location

this sounds like the "right" way of doing things, but I think in this case we could do it the "wrong" way with a context:

- create a tempfile with a temp name, e.g. TtnrLCt8920TRe.py
- add the directory of the file to sys.path
- import the file as _qt
- remove the directory from sys.path

Would that work?

> it would be fine to put some sort of version hash in the autogenerated file name and re-render if missing.

You are right that this might be overkill, but not doing so opens us up to hard-to-debug issues down the line, I feel, of the kind that this PR is trying to fix.

jni:
@tlambert03 is this ready for review? If so, do you want to merge master first?

tlambert03:
gimmie 5 minutes... just finishing up some docs and explanations.  I got the "right way" working and I think it's better now.

tlambert03:
ok.  I incorporated all of @jni's suggestions:
- resources are now built to a resource file with a version hash that includes napari version, Qt API, and Qt version, for example: `napari/resource/_qt_resources_0.2.12+29.g5b06b739.dirty_pyside2_5.14.1.py`
- If that path doesn't exist at boot, napari tries to create it, if it cannot write to that location, it tries to write to `~/.config/napari`. If it _still_ has a problem it creates a temporary file.
(by the way, the segfault I was having with `spec_from_file_location` was due to the imported module going out of scope when the function ended.  the solution was to add the new module to `sys.modules`)
- `python -m napari.resources.build_icons` is no longer necessary or used.  `CONTRIBUTING.ini` updated.

ready for review, and in the meantime, I will retest some local pip installs.

docs/developers/CONTRIBUTING.md
```diff
@@ -45,25 +45,28 @@ Code will also be linted to enforce the stylistic and logistical rules specified
 
 If you wish to tell the linter to ignore a specific line use the `# noqa` comment along with the specific error code (e.g. `import sys  # noqa: E402`) but please do not ignore errors lightly.
 
-## Building the icons
+## Adding icons
 
-Our icon build process starts with a base set of `svg` formatted icons stored in `napari/resources/icons/svg`. If you want to add a new icon to the app, make the icon in whatever program you like and add it to that folder.
+If you want to add a new icon to the app, make the icon in whatever program you
+like and add it to `napari/resources/icons/`.
```
sofroniewn: Do icons still need to be svg files?

napari/_qt/qt_viewer.py
```diff
@@ -168,10 +168,8 @@ def __init__(self, viewer):
         self._last_visited_dir = str(Path.home())
 
         self._cursors = {
-            'disabled': QCursor(
-                QPixmap(':/icons/cursor/cursor_disabled.png').scaled(20, 20)
-            ),
             'cross': Qt.CrossCursor,
+            'disabled': Qt.ForbiddenCursor,
```
sofroniewn: I would completely drop `disabled` as I don't think we use it at all, no need to keep it around

napari/_qt/qt_viewer.py
```diff
@@ -364,16 +362,8 @@ def _on_cursor(self, event):
...
         if cursor == 'square':
-            if size < 10 or size > 300:
-                q_cursor = self._cursors['cross']
-            else:
-                q_cursor = QCursor(
-                    QPixmap(':/icons/cursor/cursor_square.png').scaledToHeight(
-                        size
-                    )
-                )
+            q_cursor = QCursor(square_pixmap(self.viewer.cursor_size))
```
sofroniewn: I guess we still can't add `square` to the `_cursors` dictionary and remove this if / else clause, because we need to be able to do that resizing, but I've never loved this solution - curious if you have ideas on how it could be improved - maybe we make a square cursor that has a "resize" method on it, and then do if square and just call resize. Might still be more complex than it is worth

napari/_qt/utils.py
```diff
@@ -127,9 +130,24 @@ def qt_signals_blocked(obj):
...
         widget = getattr(obj, wdg)
         widget.setEnabled(obj.layer.editable)
         op = QGraphicsOpacityEffect(obj)
         op.setOpacity(1 if obj.layer.editable else 0.5)
         widget.setGraphicsEffect(op)
+
+
+@lru_cache(maxsize=64)
+def square_pixmap(size):
+    """Create a white/black hollow square pixmap. For use as labels cursor."""
```
sofroniewn, line 144: The current square is actually white / black / white which i found gave better contrast over many images - did you intentionally make this one just white / black? (It may be fine, just curious if it was done on purpose)

sofroniewn: Having spent more time with it, the new cursor with just the white / black is really quite nice, I like that the thickness of the border remains the same with size, something not possible with the old approach. One thing though is you need to add some clipping to size. In the old version I had `size < 10 or size > 300` and it transition to the plus symbol. This is needed if you are really zoomed out and make the brush very small, then now you can get this error message `WARNING: QPainter::setPen: Painter not active` and see nothing, or if you are very zoomed and make brush very large then the square becomes so big that it is larger than the whole screen and you can't see it anymore. Both those behaviours are undesirable

sofroniewn (commented):
This is looks really nice @tlambert03, I've left some comments, mostly minor - the only one that really needs addressing is the behaviour when the square cursor size is very small or very large.

docs/developers/CONTRIBUTING.md
sofroniewn: If so probably worth adding a note about that here

napari/_qt/qt_modal.py
```diff
@@ -123,7 +123,10 @@ def show_at(self, position='top', *, win_ratio=0.9, min_length=0):
         # necessary for transparent round corners
         self.resize(self.sizeHint())
         # make sure the popup is completely on the screen
-        screen_size = QGuiApplication.screenAt(QCursor.pos()).size()
+        if hasattr(QGuiApplication, 'screenAt'):
+            screen_size = QGuiApplication.screenAt(QCursor.pos()).size()
+        else:
+            screen_size = QGuiApplication.screens()[0].size()
```
jni, line 131: I presume this if/else accounts for different Qt versions... Could you explicitly note them?

napari/_qt/utils.py
```diff
@@ -39,7 +42,7 @@ def QImg2array(img):
 
 
 def new_worker_qthread(
-    Worker: type(QObject), *args, start=False, connections=None, **kwargs
+    Worker: Type[QObject], *args, start=False, connections=None, **kwargs
```
jni, line 45: What is this sorcery?

napari/resources/__init__.py
```diff
@@ -1,22 +1,149 @@
...
+    
+    Returns
+    -------
+    target : str or None
+        Returns the target if it is writeable, returns None if it is not.
+    """
+    target = Path(target)
+    if not target.exists():
+        try:
+            target.parent.mkdir(parents=True, exist_ok=True)
```
jni, line 41: I feel like this should go in the docstring explanation... By my reading, this will create all directories up to the target file, and it will not clean them up, correct?

napari/resources/__init__.py
```diff
@@ -1,22 +1,149 @@
...
+    Parameters
+    ----------
+    version : str, optional
+        Version string, by default the resources will be written to a file that
+        encodes the current napari version, as well as Qt backend and version:
+        ``_qt_resources_{napari.__version__}_{API}_{QT_VERSION}.py``
+
+    overwrite : bool, optional
+        Whether to recompile and overwrite the resources.
+        By default, resources will be rebuilt if any of the following are True:
```
jni:
```suggestion
        Resources will be rebuilt if any of the following are True:
```

napari/resources/__init__.py
```diff
@@ -1,22 +1,149 @@
...
+            "https://github.com/napari/napari/issues"
+        )
+    overwrite = (
+        overwrite
+        or bool(environ.get('NAPARI_REBUILD_RESOURCES'))
+        or tempfile is not None
+    )
+    # this actually builds the res.qrc Qt resources file, and then from that
+    # autogenerates the python resources file that needs to be imported
+    # if the file already exists and overwrite is False, it will not be
```
jni:
```suggestion
    # autogenerates the python resources file that needs to be imported.
    # If the file already exists and overwrite is False, it will not be
```

napari/resources/__init__.py
```diff
@@ -1,22 +1,149 @@
...
+    # regenerated.
+    respath = build_pyqt_resources(fspath(target_file), overwrite=overwrite)
+    # import the python resources file and add to sys.modules
+    spec = spec_from_file_location("napari._qt_resources", respath)
+    module = module_from_spec(spec)
+    # important to add to sys.modules! otherwise segfault when function ends.
+    sys.modules[spec.name] = module
+    spec.loader.exec_module(module)
```
jni, line 119: 🧙

napari/resources/__init__.py
```diff
@@ -1,22 +1,149 @@
...
+
+    Parameters
+    ----------
+    extra : list of str, optional
+        Additional qss files to include in stylesheet, by default None
+    
+    Returns
+    -------
+    css : str
+        the combined stylesheet.
```
jni:
```suggestion
        The combined stylesheet.
```

napari/resources/__init__.py
```diff
@@ -1,22 +1,149 @@
...
+    """Combine all qss files into single (cached) style string.
+    
+    Note, this string may still have {{ template_variables }} that need to be
+    replaced using the :func:`napari.utils.theme.template` function.  (i.e. the
+    output of this function serves as the input of ``template()``)
+
+    Parameters
+    ----------
+    extra : list of str, optional
+        Additional qss files to include in stylesheet, by default None
```
jni: file paths? It's slightly ambiguous here whether you are expecting paths or the text from the files.

napari/resources/build_icons.py
```diff
@@ -73,20 +84,64 @@ def build_icons():
...
+    # make sure we import from qtpy
+    with open(out_path, "rt") as fin:
         data = fin.read()
         data = data.replace('PySide2', 'qtpy').replace('PyQt5', 'qtpy')
-    with open(qtpy_file, "wt") as fin:
+    with open(out_path, "wt") as fin:
         fin.write(data)
+
+    # cleanup.
+    # we do this here because pip uninstall napari would not collect these
```
jni, line 184: I'm a little confused... Does this mean this would need regenerating on each boot? Or are these files used only once during creation of the .py, and then not looked at again?

jni (approved):
@tlambert03 this looks ready to me! I've left a couple of very minor comments/typo fixes. Thank you! :tada:

docs/developers/CONTRIBUTING.md
tlambert03: yes, must be svg. will leave a note.

kne42:
What do you think about doing this during setup instead, see for example what [scikit-image does](https://github.com/scikit-image/scikit-image/blob/master/skimage/filters/setup.py#L14)?

tlambert03:
> What do you think about doing this during setup instead, see for example what scikit-image does?

certainly open to that.  let me play with it a little bit

napari/_qt/qt_viewer.py
tlambert03: I think it's a totally reasonable approach.  since it's the only cursor that requires scaling... having a resize method on the cursor is more or less the same as redrawing the square each time (and that function is cached, so it's inexpensive on repeated calls).
I reinstated the size limits, but made the max size scale with the qt_viewer.canvas (rather than a fixed max of 300).

napari/_qt/utils.py
tlambert03: which, the function?  that's an old function (from the dims animation) that obviously needs better docs.  It's not related to this PR, I just couldn't keep my grimy hands off changing `type()` to `Type[]` ... since `type()` is not a valid type annotation... but I didn't know it when I wrote it.

napari/_qt/utils.py
tlambert03: I didn't intentionally exclude the inner white... and could definitely add it back if you want (though it sounds like maybe you've come around to just white/black?). I felt like two was clean and sufficient, but I'm not at all opposed to the additional white.

sorry!  I meant to re-add that... thanks for catching that.  I returned it, but made the max size `min(*self.viewer.window.qt_viewer.canvas.size)` rather than a hard-coded `300`.

napari/resources/__init__.py
tlambert03: yeah that's correct.  I'll put it in the docstring... are you otherwise ok with that?

napari/resources/__init__.py
tlambert03: good magic?  or scary magic?  (I added a link to the highly rated [SO post ](https://stackoverflow.com/a/67692/1631624) that provided the inspiration)

napari/resources/build_icons.py
tlambert03: the latter, and that was something I learned here too.  The file generated by `pyrcc5` or `pyside22-rcc` is actually a complete  bytes file with all necessary info ... neither the themed SVGs nor the res.qrc are necessary after creating (what we called) `qt.py`.

napari/_qt/utils.py
jni: No, I was wondering why the type of the thing is `Type[QObject]` rather than `QObject`.

napari/resources/__init__.py
jni: :man_shrugging: 

LOL It makes me uncomfortable, but I'm hoping that this will be used quite rarely, so let's go ahead with it for now.

napari/resources/__init__.py
jni: Oh no, thanks for the link, it's useful, I was only saying that you are a wizard. =D

napari/_qt/utils.py
tlambert03: oh good point... lol

napari/resources/build_icons.py
jni: :+1: !

napari/_qt/utils.py
sofroniewn: even better, current behavior works great - thanks!!

napari/_qt/utils.py
tlambert03: oh ... actually, sorry, not a good point 😅. This function accepts the actual class or subclasses of QObject, not an instance.  [here](https://github.com/napari/napari/blob/dec5b9965d4a3fcea2d8314064465af14a97fec7/napari/_qt/qt_dims_slider.py#L386-L391) is an example usage.

so `Type[QObject]` _is_ correct... anyway it's peripheral to this PR (#grimyhands)

napari/resources/__init__.py
tlambert03: yeah, and chances are, whatever folder it creates, would have been created on every napari launch anyway.

tlambert03:
OK, I haven't changed any of the behavior since the approving reviews, but I did refactor the generation of the themed SVGs (hadn't really reviewed that code before), and I also slightly changed the way the fallback tempfile is being created (to make it work on Windows).  I've now tested this on both mac and windows for a bunch of different Qt versions and backends... and I'm pretty confident in it now.  Can probably merge after tests pass, but will leave for a day in case anyone wants one more look

jni:
@kne42 we should mention that during the meeting we decided that doing it during setup.py gives us less flexibility and might have to be undone if we move to distribution using only pyproject.toml in the future. So we decided to stick with this approach.

tlambert03:
> @kne42 we should mention that during the meeting we decided that doing it during setup.py gives us less flexibility and might have to be undone if we move to distribution using only pyproject.toml in the future. So we decided to stick with this approach.

thanks!  forgot to mention that

sofroniewn:
> Can probably merge after tests pass

That's fine by me

PR #1039, opened 2020-03-10
Fix label styles in contrast limit popup 

tlambert03:
# Description
The labels in range slider popups (like the contrast limits popup) needed a tiny tweak after #1017.

### DIFF
--- a/napari/_qt/qt_range_slider_popup.py
+++ b/napari/_qt/qt_range_slider_popup.py
@@ -72,7 +72,7 @@ class LabelEdit(QLineEdit):
     def update_position(self):
         """Update slider position."""
         x = self.get_pos() - self.width() / 2
-        y = self.slider.handle_radius + 6
+        y = self.slider.handle_radius + 12
         self.move(QPoint(x, -y) + self.slider.pos())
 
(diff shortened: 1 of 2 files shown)


jni (approved):

napari/_qt/qt_range_slider_popup.py
```diff
@@ -72,7 +72,7 @@ def _on_text_changed(self, text):
     def update_position(self):
         """Update slider position."""
         x = self.get_pos() - self.width() / 2
-        y = self.slider.handle_radius + 6
+        y = self.slider.handle_radius + 12
```
jni, line 75: @tlambert03 I approved but now I find myself wondering whether this number is dependent on whether you have a HiDPI screen...

tlambert03: possibly.  But if so, then it falls into a big category of similar things that need fixing in some High DPI PR.  My gut feeling as that, with the HIGH_DPI env variable set... this works fine (I say that because I use a high DPI monitor at home, and when I boot into linux, it had been working with the previous number, which was similarly hard-coded)

PR #1040, opened 2020-03-10
RFC: Refactor proposal! Adding event handlers to napari (TAKE 3) 

shanaxel42:
Alright! Thanks so much for all the feedback and help on these proposals everyone! I feel like we have moved in a cool direction that works for all. Given the discussion in #1020 and in the Napari sync, I iterated on the refactor proposal with a few changes: 

- UpdateContract is now an Interface (a more accurate description of what it is)
- Controller's are now EventHandlers
- The base data layer initializes the event handler so it does not have to be done outside of a viewer.add_layer() call. 
- The vispy and qt components register themselves with the layer's event handler rather then everything happening in qt_viewer. 

I dug into the `emitter_group` code a bit more and realized it could be utilized in this refactor. I added the ability to pass a single callback method to an emitter group that automatically get's hooked up to any event added to the group. This way we can pass `layer.event_handler.on_change` into the EmitterGroup's created in the base qt and vispy layers and everything "just works". 

For reference this is what the new event architecture would look like: 

Refactor-take-3

I wanted to give more of any idea of what this approach would look like so this PR refactors more of the code then just the `interpolation` attribute. The entire `base_layer, base_qt_layer, base_vispy_layer` and `image_layer, qt_image_layer, vispy_image_layer` have been reformatted. 

Check it out at let me know all the questions/comments/concerns! 

Also I am going to start referring to this refactor as an "Event Handler Refactor" instead of a "Controller Refactor"  : )

### DIFF
--- a/napari/_qt/layers/qt_base_layer.py
+++ b/napari/_qt/layers/qt_base_layer.py
@@ -1,10 +1,12 @@
 from qtpy.QtCore import Qt
 from qtpy.QtWidgets import QSlider, QGridLayout, QFrame, QComboBox
 
+from ...layers.base._base_layer_interface import BaseLayerInterface
 from ...layers.base._base_constants import Blending
+from ...utils.event import Event, EmitterGroup
 
 
-class QtLayerControls(QFrame):
+class QtLayerControls(QFrame, BaseLayerInterface):
     """Superclass for all the other LayerControl classes.
 
     This class is never directly instantiated anywhere.
@@ -30,8 +32,14 @@ class QtLayerControls(QFrame):
         super().__init__()
 
         self.layer = layer
-        layer.events.blending.connect(self._on_blending_change)
-        layer.events.opacity.connect(self._on_opacity_change)
+
+        self.layer.event_handler.register_component_to_update(self)
+        self.events = EmitterGroup(
+            source=self,
+            blending=Event,
+            opacity=Event,
+            event_handler_callback=self.layer.event_handler.on_change,
+        )
         self.setObjectName('layer')
         self.setMouseTracking(True)
 
@@ -47,9 +55,13 @@ class QtLayerControls(QFrame):
         sld.setMinimum(0)
         sld.setMaximum(100)
         sld.setSingleStep(1)
-        sld.valueChanged.connect(self.changeOpacity)
+
+        self.emit_opacity_event = lambda value: self.events.opacity(
+            value=value / 100
+        )
+        sld.valueChanged.connect(self.emit_opacity_event)
         self.opacitySlider = sld
-        self._on_opacity_change()
+        self._on_opacity_change(self.layer.opacity)
 
         blend_comboBox = QComboBox(self)
         blend_comboBox.addItems(Blending.keys())
@@ -57,52 +69,26 @@ class QtLayerControls(QFrame):
             self.layer.blending, Qt.MatchFixedString
         )
         blend_comboBox.setCurrentIndex(index)
-        blend_comboBox.activated[str].connect(self.changeBlending)
+        blend_comboBox.activated[str].connect(self.events.blending)
         self.blendComboBox = blend_comboBox
 
-    def changeOpacity(self, value):
-        """Change opacity value on the layer model.
-
-        Parameters
-        ----------
-        value : float
-            Opacity value for shapes.
-            Input range 0 - 100 (transparent to fully opaque).
-        """
-        with self.layer.events.blocker(self._on_opacity_change):
-            self.layer.opacity = value / 100
-
-    def changeBlending(self, text):
-        """Change blending mode on the layer model.
-
-        Parameters
-        ----------
-        text : str
-            Name of blending mode, eg: 'translucent', 'additive', 'opaque'.
-        """
-        self.layer.blending = text
-
-    def _on_opacity_change(self, event=None):
+    def _on_opacity_change(self, value):
         """Receive layer model opacity change event and update opacity slider.
 
         Parameters
         ----------
-        event : qtpy.QtCore.QEvent, optional.
-            Event from the Qt context, by default None.
+        value : int
+            opacity value to change set the slider to.
         """
-        with self.layer.events.opacity.blocker():
-            self.opacitySlider.setValue(self.layer.opacity * 100)
+        self.opacitySlider.setValue(value * 100)
 
-    def _on_blending_change(self, event=None):
+    def _on_blending_change(self, value):
         """Receive layer model blending mode change event and update slider.
 
         Parameters
         ----------
-        event : qtpy.QtCore.QEvent, optional.
-            Event from the Qt context, by default None.
+        value : str
+           The value to set the blending element to
         """
-        with self.layer.events.blending.blocker():
-            index = self.blendComboBox.findText(
-                self.layer.blending, Qt.MatchFixedString
-            )
-            self.blendComboBox.setCurrentIndex(index)
+        index = self.blendComboBox.findText(value, Qt.MatchFixedString)
+        self.blendComboBox.setCurrentIndex(index)
--- a/napari/_qt/layers/qt_image_layer.py
+++ b/napari/_qt/layers/qt_image_layer.py
@@ -1,3 +1,4 @@
+from napari.layers.image.image_interface import ImageLayerInterface
 from qtpy.QtCore import Qt
 from qtpy.QtWidgets import QComboBox, QHBoxLayout, QLabel, QSlider
 
@@ -7,51 +8,53 @@ from ...layers.image._image_constants import (
     Rendering,
 )
 from .qt_image_base_layer import QtBaseImageControls
+from ...utils.event import Event
 
 
-class QtImageControls(QtBaseImageControls):
+class QtImageControls(QtBaseImageControls, ImageLayerInterface):
     """Qt view and controls for the napari Image layer.
 
-    Parameters
-    ----------
-    layer : napari.layers.Image
-        An instance of a napari Image layer.
+   Parameters
+   ----------
+   layer : napari.layers.Image
+       An instance of a napari Image layer.
 
-    Attributes
-    ----------
-    attenuationSlider : qtpy.QtWidgets.QSlider
-        Slider controlling attenuation rate for `attenuated_mip` mode.
-    attenuationLabel : qtpy.QtWidgets.QLabel
-        Label for the attenuation slider widget.
-    grid_layout : qtpy.QtWidgets.QGridLayout
-        Layout of Qt widget controls for the layer.
-    interpComboBox : qtpy.QtWidgets.QComboBox
-        Dropdown menu to select the interpolation mode for image display.
-    interpLabel : qtpy.QtWidgets.QLabel
-        Label for the interpolation dropdown menu.
-    isoThresholdSlider : qtpy.QtWidgets.QSlider
-        Slider controlling the isosurface threshold value for rendering.
-    isoThresholdLabel : qtpy.QtWidgets.QLabel
-        Label for the isosurface threshold slider widget.
-    layer : napari.layers.Image
-        An instance of a napari Image layer.
-    renderComboBox : qtpy.QtWidgets.QComboBox
-        Dropdown menu to select the rendering mode for image display.
-    renderLabel : qtpy.QtWidgets.QLabel
-        Label for the rendering mode dropdown menu.
-    """
+   Attributes
+   ----------
+   attenuationSlider : qtpy.QtWidgets.QSlider
+       Slider controlling attenuation rate for `attenuated_mip` mode.
+   attenuationLabel : qtpy.QtWidgets.QLabel
+       Label for the attenuation slider widget.
+   grid_layout : qtpy.QtWidgets.QGridLayout
+       Layout of Qt widget controls for the layer.
+   interpComboBox : qtpy.QtWidgets.QComboBox
+       Dropdown menu to select the interpolation mode for image display.
+   interpLabel : qtpy.QtWidgets.QLabel
+       Label for the interpolation dropdown menu.
+   isoThresholdSlider : qtpy.QtWidgets.QSlider
+       Slider controlling the isosurface threshold value for rendering.
+   isoThresholdLabel : qtpy.QtWidgets.QLabel
+       Label for the isosurface threshold slider widget.
+   layer : napari.layers.Image
+       An instance of a napari Image layer.
+   renderComboBox : qtpy.QtWidgets.QComboBox
+       Dropdown menu to select the rendering mode for image display.
+   renderLabel : qtpy.QtWidgets.QLabel
+       Label for the rendering mode dropdown menu.
+   """
 
     def __init__(self, layer):
         super().__init__(layer)
 
-        self.layer.events.interpolation.connect(self._on_interpolation_change)
-        self.layer.events.rendering.connect(self._on_rendering_change)
-        self.layer.events.iso_threshold.connect(self._on_iso_threshold_change)
-        self.layer.events.attenuation.connect(self._on_attenuation_change)
-        self.layer.dims.events.ndisplay.connect(self._on_ndisplay_change)
+        self.events.add(
+            interpolation=Event,
+            rendering=Event,
+            iso_threshold=Event,
+            attenuation=Event,
+        )
 
         self.interpComboBox = QComboBox(self)
-        self.interpComboBox.activated[str].connect(self.changeInterpolation)
+        self.interpComboBox.activated[str].connect(self.events.interpolation)
         self.interpLabel = QLabel('interpolation:')
 
         renderComboBox = QComboBox(self)
@@ -60,7 +63,7 @@ class QtImageControls(QtBaseImageControls):
             self.layer.rendering, Qt.MatchFixedString
         )
         renderComboBox.setCurrentIndex(index)
-        renderComboBox.activated[str].connect(self.changeRendering)
+        renderComboBox.activated[str].connect(self.events.rendering)
         self.renderComboBox = renderComboBox
         self.renderLabel = QLabel('rendering:')
 
@@ -70,7 +73,11 @@ class QtImageControls(QtBaseImageControls):
         sld.setMaximum(100)
         sld.setSingleStep(1)
         sld.setValue(self.layer.iso_threshold * 100)
-        sld.valueChanged.connect(self.changeIsoThreshold)
+        self.emit_iso_threshold_event = lambda value: self.events.iso_threshold(
+            value=value / 100
+        )
+
+        sld.valueChanged.connect(self.emit_iso_threshold_event)
         self.isoThresholdSlider = sld
         self.isoThresholdLabel = QLabel('iso threshold:')
 
@@ -80,7 +87,10 @@ class QtImageControls(QtBaseImageControls):
         sld.setMaximum(200)
         sld.setSingleStep(1)
         sld.setValue(self.layer.attenuation * 100)
-        sld.valueChanged.connect(self.changeAttenuation)
+        self.emit_attenuation_event = lambda value: self.events.attenuation(
+            value=value / 100
+        )
+        sld.valueChanged.connect(self.emit_attenuation_event)
         self.attenuationSlider = sld
         self.attenuationLabel = QLabel('attenuation:')
         self._on_ndisplay_change()
@@ -114,58 +124,22 @@ class QtImageControls(QtBaseImageControls):
         self.grid_layout.setColumnStretch(1, 1)
         self.grid_layout.setSpacing(4)
 
-    def changeInterpolation(self, text):
+    def _on_interpolation_change(self, text):
         """Change interpolation mode for image display.
 
-        Parameters
-        ----------
-        text : str
-            Interpolation mode used by vispy. Must be one of our supported
-            modes:
-            'bessel', 'bicubic', 'bilinear', 'blackman', 'catrom', 'gaussian',
-            'hamming', 'hanning', 'hermite', 'kaiser', 'lanczos', 'mitchell',
-            'nearest', 'spline16', 'spline36'
-        """
-        self.layer.interpolation = text
+       Parameters
+       ----------
+       text : str
+           Interpolation mode used by vispy. Must be one of our supported
+           modes:
+           'bessel', 'bicubic', 'bilinear', 'blackman', 'catrom', 'gaussian',
+           'hamming', 'hanning', 'hermite', 'kaiser', 'lanczos', 'mitchell',
+           'nearest', 'spline16', 'spline36'
+       """
+        index = self.interpComboBox.findText(text, Qt.MatchFixedString)
+        self.interpComboBox.setCurrentIndex(index)
 
-    def changeRendering(self, text):
-        """Change rendering mode for image display.
-
-        Parameters
-        ----------
-        text : str
-            Rendering mode used by vispy.
-            Selects a preset rendering mode in vispy that determines how
-            volume is displayed:
-            * translucent: voxel colors are blended along the view ray until
-              the result is opaque.
-            * mip: maxiumum intensity projection. Cast a ray and display the
-              maximum value that was encountered.
-            * additive: voxel colors are added along the view ray until
-              the result is saturated.
-            * iso: isosurface. Cast a ray until a certain threshold is
-              encountered. At that location, lighning calculations are
-              performed to give the visual appearance of a surface.
-            * attenuated_mip: attenuated maxiumum intensity projection. Cast a
-              ray and attenuate values based on integral of encountered values,
-              display the maximum value that was encountered after attenuation.
-              This will make nearer objects appear more prominent.
-        """
-        self.layer.rendering = text
-        self._toggle_rendering_parameter_visbility()
-
-    def changeIsoThreshold(self, value):
-        """Change isosurface threshold on the layer model.
-
-        Parameters
-        ----------
-        value : float
-            Threshold for isosurface.
-        """
-        with self.layer.events.blocker(self._on_iso_threshold_change):
-            self.layer.iso_threshold = value / 100
-
-    def _on_iso_threshold_change(self, event):
+    def _on_iso_threshold_change(self, value):
         """Receive layer model isosurface change event and update the slider.
 
         Parameters
@@ -173,21 +147,9 @@ class QtImageControls(QtBaseImageControls):
         event : qtpy.QtCore.QEvent
             Event from the Qt context.
         """
-        with self.layer.events.iso_threshold.blocker():
-            self.isoThresholdSlider.setValue(self.layer.iso_threshold * 100)
+        self.isoThresholdSlider.setValue(value * 100)
 
-    def changeAttenuation(self, value):
-        """Change attenuation rate for attenuated maximum intensity projection.
-
-        Parameters
-        ----------
-        value : Float
-            Attenuation rate for attenuated maximum intensity projection.
-        """
-        with self.layer.events.blocker(self._on_attenuation_change):
-            self.layer.attenuation = value / 100
-
-    def _on_attenuation_change(self, event):
+    def _on_attenuation_change(self, value):
         """Receive layer model attenuation change event and update the slider.
 
         Parameters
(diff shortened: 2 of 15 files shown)


napari/layers/image/image.py
```diff
@@ -337,7 +342,12 @@ def interpolation(self):
     @interpolation.setter
     def interpolation(self, interpolation):
         self._interpolation = Interpolation(interpolation)
```
sofroniewn: This line `self._interpolation = Interpolation(interpolation)` should be dropped right?

napari/layers/image/image.py
```diff
@@ -363,7 +373,12 @@ def rendering(self):
     @rendering.setter
     def rendering(self, rendering):
         self._rendering = Rendering(rendering)
```
sofroniewn: This line `self._rendering = Rendering(rendering)` should be dropped right?

napari/utils/event.py
```diff
@@ -226,8 +226,10 @@ class EventEmitter(object):
         The class of events that this emitter will generate.
     """
 
-    def __init__(self, source=None, type=None, event_class=Event):
-        self._callbacks = []
+    def __init__(
```
sofroniewn, line 229: As I mentioned in person, I think it's fine to start modifying the event class from vispy - but it would be good to have really detailed comments about what / why we're modifying. Also if it can be done in a backwards compatible way, then it might be nice to make a PR to vispy with those modifications too to at least keep things more in sync.

napari/layers/base/base.py
```diff
@@ -484,14 +509,18 @@ def selected(self):
 
     @selected.setter
     def selected(self, selected):
+        if selected:
+            self.events.select(value=selected)
```
sofroniewn, line 538: I know this is minor, but we could use this as an opportunity to remove the `select` and `deselect` events, and just have `selected` True / False - would be more keeping with the rest of the code base

napari/layers/base/base.py
```diff
@@ -334,7 +357,7 @@ def translate_grid(self, translate_grid):
         if np.all(self.translate_grid == translate_grid):
             return
         self._transform_grid.translate = np.array(translate_grid)
-        self.events.translate()
+        self.events.translate(value=self._translate)
```
sofroniewn: hmm, this is a bit of a funny one. i'm hoping to remove the `translate_grid` method soon, but maybe safest until then to think about adding an explicit `translate_grid` event. I know this doesn't relate to the core of the refactor, but just want to note it now

napari/layers/base/base.py
```diff
@@ -272,7 +280,13 @@ def blending(self):
     @blending.setter
     def blending(self, blending):
         self._blending = Blending(blending)
```
sofroniewn: `self._blending = Blending(blending)` should be dropped right?

napari/layers/base/base.py
```diff
@@ -169,7 +171,9 @@ def __init__(
         self._thumbnail_shape = (32, 32, 4)
         self._thumbnail = np.zeros(self._thumbnail_shape, dtype=np.uint8)
         self._update_properties = True
-        self._name = ''
+        self._name = name
+        self.event_handler = LayerEventHandler(component=self)
+
         self.events = EmitterGroup(
```
sofroniewn, line 199: Is there anyway the actual emitter group can be moved inside the `BaseLayerInterface`? I feel like there is now a 1-1 correspondence between the properties mentioned in the emitter group and the `_on_*_change` methods in the interface, so it could live there. This would then simplify our model files a bit.

sofroniewn (commented):
@shanaxel42 this is great! I'm really liking how this is going. I left a bunch of comments, mostly minor or notes around some places where we might need to remember some stuff during the refactor. The biggest question I had was around if the emitter groups themselves can be moved to the Interface as that could simplify things too, or is it weird to have that symmetry - if you can respond to something, you can emit something? I kinda like it :-)

napari/layers/base/base.py
shanaxel42: Yea so I went down this route. The issue becomes the order of initializations. By adding the EmitterGroup instantiation to the Interface, it no longer is "abstract" and needs to be instantiated. It also needs the event_handler callback method on instantiation (or it needs to add it after the fact). But LayerEventHandler get's instantiated with the Layer init method so you can't have access to it until that happens. And since BaseLayerInterface is a superclass of Layer things got weird. Basically I tired adding the event handler info after the init of all the events but then it was confusing how things worked down the line in ImageLayer.  I came to the conclusion that it was better to leave BaseLayerInterface and purely abstract and not have it actually instantiate anything.

sofroniewn: Ok, that's fine - thanks for looking into it. Maybe at one point it will be possible to say just have a list of properties and then autogenerate a lot of the stuff based on that list :-) but that can come later

napari/layers/image/image.py
shanaxel42: ahh yup good call! some rebasing weirdness let me clean all these up

napari/_qt/layers/qt_base_layer.py
```diff
@@ -47,62 +55,40 @@ def __init__(self, layer):
         sld.setMinimum(0)
         sld.setMaximum(100)
         sld.setSingleStep(1)
-        sld.valueChanged.connect(self.changeOpacity)
+
+        self.emit_opacity_event = lambda value: self.events.opacity(
+            value=value / 100
+        )
```
ttung, line 61: is it common that transformations to the input value have to occur?  i imagine this idiom could get tiresome.

napari/_qt/layers/qt_base_layer.py
```diff
@@ -47,62 +55,40 @@ def __init__(self, layer):
...
-            self.opacitySlider.setValue(self.layer.opacity * 100)
+        self.opacitySlider.setValue(value * 100)
 
-    def _on_blending_change(self, event=None):
+    def _on_blending_change(self, value):
         """Receive layer model blending mode change event and update slider.
 
         Parameters
         ----------
         event : qtpy.QtCore.QEvent, optional.
```
ttung: this needs to be updated.

napari/_qt/layers/qt_base_layer.py
```diff
@@ -47,62 +55,40 @@ def __init__(self, layer):
...
-        """
-        self.layer.blending = text
-
-    def _on_opacity_change(self, event=None):
+    def _on_opacity_change(self, value):
         """Receive layer model opacity change event and update opacity slider.
 
         Parameters
         ----------
         event : qtpy.QtCore.QEvent, optional.
```
ttung: this needs to be updated.

napari/_vispy/vispy_image_layer.py
```diff
@@ -95,21 +86,21 @@ def _on_data_change(self, event=None):
...
+    def _on_rendering_change(self, value):
         if self.layer.dims.ndisplay == 3:
-            self.node.method = self.layer.rendering
-            self._on_threshold_change()
+            self.node.method = value
+            self._on_iso_threshold_change(value=None)
 
-    def _on_colormap_change(self, event=None):
-        cmap = self.layer.colormap[1]
+    def _on_colormap_change(self, value=None):
```
ttung, line 100: While it isn't illegal to have a default value when the base interface doesn't, it may be incompatible with type checking.  Not relevant now, but may be one day.

napari/_vispy/vispy_image_layer.py
```diff
@@ -95,21 +86,21 @@ def _on_data_change(self, event=None):
...
+        cmap = value if value else self.layer.colormap[1]
```
ttung, line 101: This may illustrate a weakness of having the data and the data handler as two different constructs.  Ideally, one should be referencing the data layer in all the individual handler methods, **OR** using the value passed in on the handler.

napari/layers/base/layer_event_handler.py
```diff
@@ -0,0 +1,26 @@
...
+    def __init__(self, component: BaseInterface = None):
+        super().__init__(component)
+
+    def on_change(self, event=None):
+        """
+        Process changes made from any interface
+        """
+        name = event.type
+        value = event.value
+        print(f"event: {name}")
```
ttung: maybe remove?
even better would be to add this to a [logger](https://docs.python.org/3/library/logging.html)

napari/layers/base/layer_event_handler.py
```diff
@@ -0,0 +1,26 @@
...
+        for component in self.components_to_update:
+            update_method_name = f"_on_{name}_change"
+            update_method = getattr(component, update_method_name)
```
ttung, line 36: Might be good to leave some documentation around this area.  If someone adds a new event, and trips `AttributeError`, they'll come poking around and wonder what to do.  Alternative: catch the error and raise a more helpful message.

napari/_qt/layers/qt_base_layer.py
```diff
@@ -30,8 +32,14 @@ def __init__(self, layer):
...
         self.layer = layer
-        layer.events.blending.connect(self._on_blending_change)
-        layer.events.opacity.connect(self._on_opacity_change)
+
+        self.layer.event_handler.register_component_to_update(self)
+        self.events = EmitterGroup(
+            source=self,
+            blending=Event,
+            opacity=Event,
+            event_handler_callback=self.layer.event_handler.on_change,
```
ttung, line 41: I imagine this is something we need to document carefully.

napari/layers/base/_base_layer_interface.py
```diff
@@ -0,0 +1,86 @@
+from abc import abstractmethod
+
+from napari.utils.base_interface import BaseInterface
+
+
+class BaseLayerInterface(BaseInterface):
+    """
+    Base class for UpdateContract. Defines a set of methods shared between the data layer visual
```
ttung:
```suggestion
    Empty implementation of BaseInterface. Defines a set of methods shared between the data layer visual
```

napari/_qt/layers/qt_base_layer.py
sofroniewn: i think we can make the slider go from 0 to 1 (which is how we like to think of opacity) in steps of 0.01 rather than 0 to 100 in steps of 1, then this won't be needed. Similarly in other places where the variable is transformed for use with a slider

napari/layers/base/layer_event_handler.py
shanaxel42: Yea that's what I was thinking!

napari/_qt/layers/qt_base_layer.py
tlambert03: I think we should probably have our own "QtDoubleSlider" that receives and emits the values properly.  Something like the one I started in magicgui [here](https://github.com/napari/magicgui/blob/11838be48b92069abb2d9f74514d1fe355580714/magicgui/_qt.py#L260-L279).  As @sofroniewn said, that can come in another PR, but will probably simplify this one a lot

shanaxel42: Yea it seemed to be just these two sliders so if we could change the ranges that'd be great

sofroniewn: @tlambert03 or I should be able to do that quickly in a stand alone PR

napari/layers/base/layer_event_handler.py
```diff
@@ -0,0 +1,26 @@
...
+
+class LayerEventHandler(EventHandler):
+    """
+    Event handler for layer specific events. Receives change events made from the data/controls/
+    or visual interface and updates all associated components.
+
+    Ex. ImageLayer components = [qt_image.py, vispy_image.py, image.py]
+    """
+
+    def __init__(self, component: BaseInterface = None):
```
tlambert03, line 19: might as well remove until we need to do something other than just call super()

napari/_vispy/vispy_image_layer.py
```diff
@@ -95,21 +86,21 @@ def _on_data_change(self, event=None):
                 self.node.set_data(data, clim=self.layer.contrast_limits)
         self.node.update()
 
-    def _on_interpolation_change(self, event=None):
+    def _on_interpolation_change(self, interpolation):
```
tlambert03, line 92: while we're doing this, it might be nice to standardize even the argument naming.  so `interpolation` here (and all the other setters) could just be `value`

napari/layers/base/base.py
```diff
@@ -271,8 +279,12 @@ def blending(self):
 
     @blending.setter
     def blending(self, blending):
-        self._blending = Blending(blending)
-        self.events.blending()
+        self.events.blending(value=blending)
+
+    def _on_blending_change(self, blending):
+        if isinstance(blending, str):
```
tlambert03, line 306: this is a reversion of #1027.  no longer need to check `isinstance(str)` ... just call `Blending()`

napari/layers/image/image.py
```diff
@@ -336,8 +341,12 @@ def interpolation(self):
 
     @interpolation.setter
     def interpolation(self, interpolation):
-        self._interpolation = Interpolation(interpolation)
-        self.events.interpolation()
+        self.events.interpolation(value=interpolation)
+
+    def _on_interpolation_change(self, interpolation):
+        if isinstance(interpolation, str):
```
tlambert03: reversion

napari/layers/image/image.py
```diff
@@ -362,8 +371,12 @@ def rendering(self):
 
     @rendering.setter
     def rendering(self, rendering):
-        self._rendering = Rendering(rendering)
-        self.events.rendering()
+        self.events.rendering(value=rendering)
+
+    def _on_rendering_change(self, rendering):
+        if isinstance(rendering, str):
```
tlambert03: reversion

napari/utils/event.py
```diff
@@ -665,10 +674,11 @@ class EmitterGroup(EventEmitter):
         See the :func:`add <vispy.event.EmitterGroup.add>` method.
     """
 
-    def __init__(self, source=None, auto_connect=True, **emitters):
+    def __init__(self, source=None, event_handler_callback=None, **emitters):
```
tlambert03, line 677: why removing `auto_connect` here?  (also note, that kwarg is still being used in `Layer.__init__`)

napari/layers/image/image_interface.py
```diff
@@ -0,0 +1,36 @@
...
+from ..base._base_layer_interface import BaseLayerInterface
+
+
+class ImageLayerInterface(BaseLayerInterface):
+    """
+    Defines the getters/setters editable across components for an ImageLayer
+    """
+
+    @abstractmethod
+    def _on_interpolation_change(self, value):
```
tlambert03, line 11: still wondering about the need for this interface... do these serve any other purpose than to make sure that the `getattr` call in `EventHander.on_change` doesn't throw an error?  And if so, can't we just use a `try/except` or a default to `getattr`?
```python
method = getattr(component, update_method_name, None)
if method:
    ...
```

tlambert03 (commented):
Yep, still like the direction this is going in!  feel free to keep pushing.  Left some thoughts throughout.

I do wish I could actually run napari to test this a bit more, but I get an error when trying to add an image to the viewer:
<details>

```
  File "/Users/talley/miniconda3/envs/napdev/lib/python3.8/runpy.py", line 193, in _run_module_as_main
  File "/Users/talley/miniconda3/envs/napdev/lib/python3.8/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/Users/talley/Dropbox (HMS)/Python/forks/napari/napari/__main__.py", line 105, in <module>
    sys.exit(main())
  File "/Users/talley/Dropbox (HMS)/Python/forks/napari/napari/__main__.py", line 101, in main
    v.add_image(images, rgb=args.rgb)
  File "/Users/talley/Dropbox (HMS)/Python/forks/napari/napari/components/add_layers_mixin.py", line 157, in add_image
    layer = layers.Image(
  File "/Users/talley/Dropbox (HMS)/Python/forks/napari/napari/layers/image/image.py", line 208, in __init__
    self._on_gamma(gamma)
  File "/Users/talley/Dropbox (HMS)/Python/forks/napari/napari/layers/intensity_mixin.py", line 154, in _on_gamma
    self._update_thumbnail()
  File "/Users/talley/Dropbox (HMS)/Python/forks/napari/napari/layers/image/image.py", line 570, in _update_thumbnail
    downsampled = np.clip(downsampled, low, high)
  File "<__array_function__ internals>", line 5, in clip
  File "/Users/talley/miniconda3/envs/napdev/lib/python3.8/site-packages/numpy/core/fromnumeric.py", line 2084, in clip
    return _wrapfunc(a, 'clip', a_min, a_max, out=out, **kwargs)
  File "/Users/talley/miniconda3/envs/napdev/lib/python3.8/site-packages/numpy/core/fromnumeric.py", line 61, in _wrapfunc
    return bound(*args, **kwds)
  File "/Users/talley/miniconda3/envs/napdev/lib/python3.8/site-packages/numpy/core/_methods.py", line 99, in _clip
    raise ValueError("One of max or min must be given")
ValueError: One of max or min must be give
```
</details>

as noted, I'm still hazy on the need for the interface class.  Seems like it's simply there to avoid calling a method that doesn't exist, which is just as easily done with `method = getattr(object, name, None);  if method: ...`

napari/layers/image/image_interface.py
royerloic: This can be checked 'at compile time' which with python means that for example pycharm can tell you right away that you have non-compliant implementation even before you run the code. Also, interfaces are implicitly  documentation about which methods need to be implemented. it might be difficult to execute all code paths to figure out that 'all is there'. of course, tests can do a lot to help in this respect, but it is good to have this explicit declaration that can be checked 'offline'.

tlambert03: I do think simply having things explicitly documented and discoverable by IDEs is sufficient reason to include it ... and we discussed that too in the "atlantic" meeting yesterday.  But this is not exactly a traditional python interface (it doesn't impose anything on its subclasses because it doesn't inherit from ABC, and it is not a `typing.Protocol`), and while I too lean towards using it, I think it's important that we understand exactly _why_ it's there.
Just to play devils advocate for a moment, all of these `_on_*_change` actually _are_ defined elsewhere, so IDEs will be able to find them... but also, internally, in this PR, these methods are mostly being called in `LayerEventHandler.on_change`, where a `getattr(str)`  is already being used to look them up (which pycharm can't trace anyway)... so while what you're saying here is _usually_ the case for an interface like this... my point was that in this case, it's _kinda_ not.  (but again, simply having one place to document the "intentions" for each event is perhaps justification enough)

kne42:
As discussed in [this thread](https://github.com/napari/napari/pull/1040#discussion_r393260527), we are weighing the benefits of using an interface vs. protocols. I am leaning towards an interface due to where it catches errors: when a class is missing methods, a protocol will catch this only when the type checking occurs (which also means that anyone with type checking not turned on won't find it), whereas with an interface, it will catch it and display the errors directly in the source code of that class, which I think is much cleaner.

@tlambert03 brought up that this checking in the interface is currently not done, so I will add it.

Thoughts?

napari/_qt/layers/qt_base_layer.py
jni: @sofroniewn did you ever get this in? I couldn't find such a PR...

napari/_qt/layers/qt_image_layer.py
```diff
@@ -7,51 +8,53 @@
...
-        Label for the isosurface threshold slider widget.
-    layer : napari.layers.Image
-        An instance of a napari Image layer.
-    renderComboBox : qtpy.QtWidgets.QComboBox
-        Dropdown menu to select the rendering mode for image display.
-    renderLabel : qtpy.QtWidgets.QLabel
-        Label for the rendering mode dropdown menu.
-    """
+   Parameters
+   ----------
```
jni, line 18: This indent error needs fixing...

napari/_qt/layers/qt_image_layer.py
```diff
@@ -114,118 +124,52 @@ def __init__(self, layer):
...
-        with self.layer.events.blocker(self._on_iso_threshold_change):
-            self.layer.iso_threshold = value / 100
+       Parameters
+       ----------
+       text : str
+           Interpolation mode used by vispy. Must be one of our supported
+           modes:
+           'bessel', 'bicubic', 'bilinear', 'blackman', 'catrom', 'gaussian',
+           'hamming', 'hanning', 'hermite', 'kaiser', 'lanczos', 'mitchell',
+           'nearest', 'spline16', 'spline36'
```
jni, line 137: Side note: how do we feel about trimming this list way down? I feel like there's really no need for us to provide all these, and the list is confusing. Alternatively, we could sort into two lists with a [separator](https://doc.qt.io/qtforpython/PySide2/QtWidgets/QComboBox.html#PySide2.QtWidgets.PySide2.QtWidgets.QComboBox.insertSeparator). First "nearest", "bilinear", "bicubic", and the second containing the rest?

napari/layers/base/_base_layer_interface.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+class BaseLayerInterface(BaseInterface):
+    """
+    Empty implementation of BaseInterface. Defines a set of methods shared between the data layer visual
+    rendering and controls.
+    """
+
+    @abstractmethod
+    def _on_refresh_change(self):
```
jni, line 13: Can we add docstrings to these? It'd be good for subclass writers to know what they are meant to do when defining these.

napari/layers/base/layer_event_handler.py
```diff
@@ -0,0 +1,26 @@
+from ...utils.event_handler import EventHandler
+from napari.utils.base_interface import BaseInterface
+
+
+class LayerEventHandler(EventHandler):
```
jni, line 11: Holy crap, I was going through the code above this going, "Ok, so all the complexity is going away, I bet it's just been moved to the LayerEventHandler...." But this class is so dead simple! :exploding_head:

napari/layers/image/image.py
```diff
@@ -225,16 +226,15 @@ def __init__(
         self._data_thumbnail = self._data_view
 
         # Set contrast_limits and colormaps
-        self._gamma = gamma
-        self._iso_threshold = iso_threshold
-        self._attenuation = attenuation
+        self._on_gamma(gamma)
+        self._on_iso_threshold_change(iso_threshold)
+        self._on_attenuation_change(attenuation)
```
jni: Some update functions are called `on_attr_change` while others are `on_attr`... Should we unify these?

napari/layers/image/image_interface.py
jni: Ok so I'm a little in the middle here. I like the self-documenting aspect of having the interface, but also it's ugly. The way I would *want* the interface documented is something like:

```python
@dataclass
class ImageLayerInterface:
    interpolation : str = 'nearest'
    contrast_limits : Tuple[int] = (0, 1)
    rendering : str = 'mip'
    iso_threshold : float = 0.5
    attenuation : float = 1.0
    gamma : float = 1.0
    colormap : str = 'gray'
```

That's of course incorrect, but perhaps there's a class decorator we could come up with that would generate the abstract interface for us from a code snippet like the above?

jni: And btw, at a minimum, my comment above in the BaseLayerInterface applies: the methods should at least have docstrings describing what they're about.

napari/layers/intensity_mixin.py
```diff
@@ -121,7 +125,9 @@ def gamma(self):
 
     @gamma.setter
     def gamma(self, value):
+        self.events.gamma(value=value)
+
+    def _on_gamma(self, value):
```
jni: Again, I don't understand why some attributes have "on_attr_**change**" while others don't...

napari/utils/base_interface.py
```diff
@@ -0,0 +1,4 @@
+class BaseInterface:
+    """
+    Base interface for updateble components
+    """
```
jni, line 4: Now this one does seem useless....? :joy:

napari/utils/event.py
jni: Agreed, I very very strongly would like this to be backwards compatible, and push the updates upstream to vispy.

napari/utils/event.py
jni: Also concerned about this... Again, it would be best to make any changes be backwards compatible here.

jni:
@kne42 

> I am leaning towards an interface due to where it catches errors

See [my comment on the thread](https://github.com/napari/napari/pull/1040/files#r428545893). Generally I'm on board with either approach, but I don't really agree with the rationale... IDEs will catch typing errors these days, and we can add mypy to our CI, so catching errors is not really a huge deal. For me the biggest concern is the quality of the code, which in the current syntax is pretty ugly. I've suggested something like what I would *like* it to look like. I figure if anyone can make it work it's you. :joy: 

@shanaxel42 dunno if you're reading but I just want to say, while I wasn't looking you got this to a *beautiful* spot, I really love where it's landed. So thanks for this and I hope you're going great in your new projects! :blush:

napari/layers/intensity_mixin.py
kne42: it's a typo that I'm fixing

napari/layers/image/image_interface.py
kne42: while this looks a lot cleaner, I fear that it does not make it obvious that interfaces need to implement a `_on_[name]_change` method. also, due to the autogenerated nature of these classes, they won't work with PyCharm if we're checking using abstract methods, but it may work if it creates a `typing.Protocol` since I believe that PyCharm uses mypy behind the scenes, but I'd need to check it to find out

napari/layers/intensity_mixin.py
jni: Good to know I'm not crazy! =)

sofroniewn:
@kne42 is this now ready for me to pull down and play with? If so i'll work on it tomorrow/ over the weekend trying to fix a few more of the weird tests!!

napari/layers/base/base.py
sofroniewn: noted

napari/_vispy/vispy_image_layer.py
sofroniewn: I'm revisiting this PR now see https://github.com/shanaxel42/napari/pull/2, and agreed this is a bit strange. Curious how @napari/core-devs think we should resolve it?

It's hard for me to see how we'd only use the values passed in the handler, for example the next line requires the `layer.gamma` property even though this is the `_on_colormap_change` handler as you need to know both `gamma` and `colormap` to draw the right thing.

It's also a bit funny now because the type of the value passed to the event handler can be different now if it comes from the view side or the model side - here for example the model passes a tuple for the colormap whereas the view passes a string. Now maybe that's a symptom of a deeper problem, but it's hard for me to see how to resolve all this

napari/layers/image/image_interface.py
sofroniewn: Having now tried to get tests working with https://github.com/shanaxel42/napari/pull/2 I'm leaning away from this as we ended up needing to register components that don't implement the full interface (see QtLayerWidget) and I think there's nothing wrong with that. Given that this is all internally facing code as well (i.e. no plugin developer would ever be adding an interface here) I really think we can do without them.

sofroniewn: I went with the method proposed here https://github.com/napari/napari/pull/1040/files#r393260527

napari/utils/event.py
sofroniewn: Yeah, this is a mystery to me right now, not all events connections are working as expected and I can't figure out why - specifically `layer.events.name.connect(lambda e: layers._update_name(e))` in https://github.com/shanaxel42/napari/pull/2 which I unfortunately still need

napari/layers/base/layer_event_handler.py
```diff
@@ -0,0 +1,37 @@
...
+        try:  # until refactor on all layers is complete, not all events will have a value property
+            value = event.value
+        except AttributeError:
+            logger.debug(f"did not handle event {name}")
+            return
+
+        for component in self.components_to_update:
+            update_method_name = f"_on_{name}_change"
+            update_method = getattr(component, update_method_name)
+            update_method(value)
```
sofroniewn, line 37: One thing I'm wondering is - should the event handler be able to modify the value coming in from an event before calling the update methods?

This would maybe move the event handler towards being more of a controller where it actually did logic, but I'm having a hard time fitting in a couple of our more advanced use cases and it could simplify somethings elsewhere.

Take this example, the user types in a new name into the dialog box on the GUI, it triggers and event with that name, then the event arrives at the layer which sets the name, and at the layerlist/ viewer which has to check if the name is unique and if not coerces it to be unique, emitting a new event with the new name which then goes back to everyone. Here we now have two events and things could easily get out of sync. If the event handler itself did the checking then that might be simpler.

Basically I'm wondering if any place where we're putting logic in the `_on_*_change` method on the base layer that changes the value in a way that all the "listeners" will want to know, should probably go in the event handler. For example in the `_on_colormap_change` method we take the string and generate the actual colormap. This is needed in the vispy listener too (see @ttung comment here https://github.com/napari/napari/pull/1040/files#r391268444). If we did the transformation in the event handler then all our listeners could get the actual colormap - or maybe these transformations just shouldn't be living in the places they are living either!

Thoughts from @shanaxel42 @ttung @kne42 would be appreciated on this one

sofroniewn: I've given this a try here https://github.com/shanaxel42/napari/pull/2/commits/a9f182e886b360180f8ee619c117361e8fa3605f see comment here https://github.com/shanaxel42/napari/pull/2#issuecomment-647029262. I think it might work quite nicely, but curious what @napari/core-devs think

sofroniewn:
This can now be closed following the merge of #1376 which supersedes it. The amazing diagram is preserved in #1394. Thanks to all who worked on this, in particular @shanaxel42 and @kne42 !!

Issue #860, opened 2020-01-09
Points & polygons position inconsistent in grid view mode (needs spatial coordinates)


GenevieveBuckley:
## 🐛 Bug

Points and polygons do not share spatial coordinates with images they are selected from. This means when you switch grid view on/off points & polygons no longer overlay the image.

This is just one example of what we're trying to fix with the discussion around spatial coordinates at https://github.com/napari/napari/issues/763. I think the self-contained example in this issue could be useful as a test case.

## To Reproduce

1. Open napari with an image.
```
$ ipython --gui=qt
Python 3.7.5 (default, Oct 31 2019, 15:18:51) [MSC v.1916 64 bit (AMD64)]
Type 'copyright', 'credits' or 'license' for more information
IPython 7.10.1 -- An enhanced Interactive Python. Type '?' for help.
In [1]: import napari
In [2]: napari.__version__
Out[2]: '0.2.3+104.g0d1c812'
In [3]: from skimage.data import astronaut, camera
In [4]: viewer = napari.Viewer()
In [5]: viewer.add_image(camera())
```

2. Add a points layer, and select some points on the image.
<img width="477" alt="napari_points_image_1" src="https://user-images.githubusercontent.com/30920819/72029124-0e740d00-32d9-11ea-80af-bf5ce3b487a3.PNG">

3. Click the "grid mode" view.
<img width="473" alt="napari_points_image_2" src="https://user-images.githubusercontent.com/30920819/72029132-12a02a80-32d9-11ea-9506-c63fafb4dd27.PNG">

## Expected behavior

Points and polygons should share spatial coordinates with the underlying image data, and always be displayed overlaid regardless of which image mode (grid-view/2D/3D) is selected.

## Environment

 - Please copy and paste the information at napari info option in help menubar here:
```
napari: 0.2.3+104.g0d1c812
Platform: Windows-10-10.0.17763-SP0
Python: 3.7.5 (default, Oct 31 2019, 15:18:51) [MSC v.1916 64 bit (AMD64)]
Qt: 5.13.2
PySide2: 5.13.2
VisPy: 0.6.3
NumPy: 1.17.4
SciPy: 1.3.3
scikit-image: 0.16.2
Dask: 2.8.1

GL version: 4.5.0 - Build 25.20.100.6472
MAX_TEXTURE_SIZE: 16384
```
 - Any other relevant information:

## Additional context

Relevant issues: https://github.com/napari/napari/issues/763

jni:
[...]

PR #1042, opened 2020-03-11
Add chains of transforms

sofroniewn:
# Description
This PR follows on from #1018 and #885 and the work of @jni and @GenevieveBuckley by adding support for a `ChainTransform` (first described in #885) and consolidating our current 3 types of transform - view2data, data2world, and world2grid inside it.

This PR is not designed to change any functionality, but makes it easier to chain our transforms together. We still have to think about what does it look like when a user wants to provide many transforms of mixed types etc as described in #1016, this PR is only a small step in that direction.

The next step after this will be to introduce the concept of a world coordinate system that our dims sliders obey too. 

Would love your feedback @GenevieveBuckley @jni and others

## Type of change

# How has this been tested?

- [x] adds `napari/layers/_tests/test_transform_chain.py`

## Final checklist:

### DIFF
--- a/napari/_vispy/_tests/test_vispy_big_images.py
+++ b/napari/_vispy/_tests/test_vispy_big_images.py
@@ -11,8 +11,8 @@ def test_big_2D_image(viewer_factory):
     visual = view.layer_to_visual[layer]
     assert visual.node is not None
     if visual.MAX_TEXTURE_SIZE_2D is not None:
-        ds = np.ceil(np.divide(shape, visual.MAX_TEXTURE_SIZE_2D)).astype(int)
-        assert np.all(layer._transforms['view2data'].scale == ds)
+        s = np.ceil(np.divide(shape, visual.MAX_TEXTURE_SIZE_2D)).astype(int)
+        assert np.all(layer._transforms['view2data'].scale == s)
 
 
 def test_big_3D_image(viewer_factory):
@@ -25,5 +25,5 @@ def test_big_3D_image(viewer_factory):
     visual = view.layer_to_visual[layer]
     assert visual.node is not None
     if visual.MAX_TEXTURE_SIZE_3D is not None:
-        ds = np.ceil(np.divide(shape, visual.MAX_TEXTURE_SIZE_3D)).astype(int)
-        assert np.all(layer._transforms['view2data'].scale == ds)
+        s = np.ceil(np.divide(shape, visual.MAX_TEXTURE_SIZE_3D)).astype(int)
+        assert np.all(layer._transforms['view2data'].scale == s)
--- a/napari/_vispy/_tests/test_vispy_big_images.py
+++ b/napari/_vispy/_tests/test_vispy_big_images.py
@@ -12,7 +12,7 @@ def test_big_2D_image(viewer_factory):
     assert visual.node is not None
     if visual.MAX_TEXTURE_SIZE_2D is not None:
         ds = np.ceil(np.divide(shape, visual.MAX_TEXTURE_SIZE_2D)).astype(int)
-        assert np.all(layer._transform_view.scale == ds)
+        assert np.all(layer._transforms['view2data'].scale == ds)
 
 
 def test_big_3D_image(viewer_factory):
@@ -26,4 +26,4 @@ def test_big_3D_image(viewer_factory):
     assert visual.node is not None
     if visual.MAX_TEXTURE_SIZE_3D is not None:
         ds = np.ceil(np.divide(shape, visual.MAX_TEXTURE_SIZE_3D)).astype(int)
-        assert np.all(layer._transform_view.scale == ds)
+        assert np.all(layer._transforms['view2data'].scale == ds)
--- a/napari/_vispy/_tests/test_vispy_big_images.py
+++ b/napari/_vispy/_tests/test_vispy_big_images.py
@@ -12,7 +12,7 @@ def test_big_2D_image(viewer_factory):
     assert visual.node is not None
     if visual.MAX_TEXTURE_SIZE_2D is not None:
         s = np.ceil(np.divide(shape, visual.MAX_TEXTURE_SIZE_2D)).astype(int)
-        assert np.all(layer._transforms['view2data'].scale == s)
+        assert np.all(layer._transforms['tile2data'].scale == s)
 
 
 def test_big_3D_image(viewer_factory):
@@ -26,4 +26,4 @@ def test_big_3D_image(viewer_factory):
     assert visual.node is not None
     if visual.MAX_TEXTURE_SIZE_3D is not None:
         s = np.ceil(np.divide(shape, visual.MAX_TEXTURE_SIZE_3D)).astype(int)
-        assert np.all(layer._transforms['view2data'].scale == s)
+        assert np.all(layer._transforms['tile2data'].scale == s)
--- a/napari/_vispy/vispy_image_layer.py
+++ b/napari/_vispy/vispy_image_layer.py
@@ -266,7 +266,7 @@ class VispyImageLayer(VispyBaseLayer):
             scale = np.ones(self.layer.ndim)
             for i, d in enumerate(self.layer.dims.displayed):
                 scale[d] = downsample[i]
-            self.layer._transforms['view2data'].scale = scale
+            self.layer._transforms['tile2data'].scale = scale
             self._on_scale_change()
             slices = tuple(slice(None, None, ds) for ds in downsample)
             data = data[slices]
--- a/napari/_vispy/vispy_image_layer.py
+++ b/napari/_vispy/vispy_image_layer.py
@@ -266,7 +266,7 @@ class VispyImageLayer(VispyBaseLayer):
             scale = np.ones(self.layer.ndim)
             for i, d in enumerate(self.layer.dims.displayed):
                 scale[d] = downsample[i]
-            self.layer._transform_view.scale = scale
+            self.layer._transforms['view2data'].scale = scale
             self._on_scale_change()
             slices = tuple(slice(None, None, ds) for ds in downsample)
             data = data[slices]
--- a/napari/layers/_tests/test_transform_chain.py
+++ b/napari/layers/_tests/test_transform_chain.py
@@ -15,13 +15,13 @@ def test_transform_chain():
     npt.assert_allclose(new_coord_1, new_coord_2)
 
 
-def test_transform_chain_composite():
+def test_transform_chain_simplified():
     coord = [10, 13]
     transform_a = ScaleTranslate(scale=[2, 3], translate=[8, -5])
     transform_b = ScaleTranslate(scale=[0.3, 1.4], translate=[-2.2, 3])
 
     transform_chain = TransformChain([transform_a, transform_b])
-    transform_c = transform_chain.composite
+    transform_c = transform_chain.simplified
 
     new_coord_1 = transform_c(coord)
     new_coord_2 = transform_chain(coord)
@@ -58,7 +58,7 @@ def test_transform_chain_slice():
     npt.assert_allclose(new_coord_1, new_coord_2)
 
 
-def test_transform_chain_pad():
+def test_transform_chain_expanded():
     coord = [10, 3, 13]
     transform_a = ScaleTranslate(scale=[2, 1, 3], translate=[8, 0, -5])
     transform_b = ScaleTranslate(scale=[0.3, 1, 1.4], translate=[-2.2, 0, 3])
@@ -67,8 +67,8 @@ def test_transform_chain_pad():
 
     transform_chain_a = TransformChain([transform_a, transform_b])
     transform_chain_b = TransformChain([transform_c, transform_d])
-    transform_chain_padded = transform_chain_b.set_pad([1])
+    transform_chain_expandded = transform_chain_b.expand_dims([1])
 
     new_coord_2 = transform_chain_a(coord)
-    new_coord_1 = transform_chain_padded(coord)
+    new_coord_1 = transform_chain_expandded(coord)
     npt.assert_allclose(new_coord_1, new_coord_2)
--- a/napari/layers/_tests/test_transform_chain.py
+++ b/napari/layers/_tests/test_transform_chain.py
@@ -1,0 +1,74 @@
+import numpy.testing as npt
+from napari.layers.transforms import ScaleTranslate, TransformChain
+
+
+def test_transform_chain():
+    coord = [10, 13]
+    transform_a = ScaleTranslate(scale=[2, 3], translate=[8, -5])
+    transform_b = ScaleTranslate(scale=[0.3, 1.4], translate=[-2.2, 3])
+    transform_c = transform_b.compose(transform_a)
+
+    transform_chain = TransformChain([transform_a, transform_b])
+
+    new_coord_1 = transform_c(coord)
+    new_coord_2 = transform_chain(coord)
+    npt.assert_allclose(new_coord_1, new_coord_2)
+
+
+def test_transform_chain_composite():
+    coord = [10, 13]
+    transform_a = ScaleTranslate(scale=[2, 3], translate=[8, -5])
+    transform_b = ScaleTranslate(scale=[0.3, 1.4], translate=[-2.2, 3])
+
+    transform_chain = TransformChain([transform_a, transform_b])
+    transform_c = transform_chain.composite
+
+    new_coord_1 = transform_c(coord)
+    new_coord_2 = transform_chain(coord)
+    npt.assert_allclose(new_coord_1, new_coord_2)
+
+
+def test_transform_chain_inverse():
+    coord = [10, 13]
+    transform_a = ScaleTranslate(scale=[2, 3], translate=[8, -5])
+    transform_b = ScaleTranslate(scale=[0.3, 1.4], translate=[-2.2, 3])
+
+    transform_chain = TransformChain([transform_a, transform_b])
+    transform_chain_inverse = transform_chain.inverse
+
+    new_coord = transform_chain(coord)
+    orig_coord = transform_chain_inverse(new_coord)
+    npt.assert_allclose(coord, orig_coord)
+
+
+def test_transform_chain_slice():
+    coord = [10, 13]
+    transform_a = ScaleTranslate(scale=[2, 3, 3], translate=[8, 2, -5])
+    transform_b = ScaleTranslate(scale=[0.3, 1, 1.4], translate=[-2.2, 4, 3])
+    transform_c = ScaleTranslate(scale=[2, 3], translate=[8, -5])
+    transform_d = ScaleTranslate(scale=[0.3, 1.4], translate=[-2.2, 3])
+
+    transform_chain_a = TransformChain([transform_a, transform_b])
+    transform_chain_b = TransformChain([transform_c, transform_d])
+
+    transform_chain_sliced = transform_chain_a.set_slice([0, 2])
+
+    new_coord_1 = transform_chain_sliced(coord)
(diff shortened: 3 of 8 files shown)


napari/_vispy/_tests/test_vispy_big_images.py
```diff
@@ -12,7 +12,7 @@ def test_big_2D_image(viewer_factory):
     assert visual.node is not None
     if visual.MAX_TEXTURE_SIZE_2D is not None:
         ds = np.ceil(np.divide(shape, visual.MAX_TEXTURE_SIZE_2D)).astype(int)
-        assert np.all(layer._transform_view.scale == ds)
+        assert np.all(layer._transforms['view2data'].scale == ds)
```
GenevieveBuckley: This naming is so much clearer than it was, love that!

GenevieveBuckley: Although I'm not entirely sure why `ds` is called `ds` if it's a scale variable...

napari/layers/base/base.py
```diff
@@ -145,21 +145,30 @@ def __init__(
...
-        # level of an input pyramid, where _transform contains the transform
-        # only from the base level to world coordinates, another is when an
-        # image is larger than the maximum allowed texture size and has been
-        # downsampled
-        self._transform_view = ScaleTranslate(np.ones(ndim), np.zeros(ndim))
-        # Create additional transform mapping world-coordinates into a grid
-        # for looking at layers side-by-side
-        self._transform_grid = ScaleTranslate(np.ones(ndim), np.zeros(ndim))
+        # Create a transform chain consisting of three transforms:
+        # - An initial transform mapping viewed data to the data coordinates
```
GenevieveBuckley: Can you use a numbered list here instead of bullet points? 

I suggest it because there's a lot of explanation in the first point & it took me a second to remember I was actually in the middle of reading a list when I got to the next point.

napari/layers/base/base.py
```diff
@@ -145,21 +145,30 @@ def __init__(
...
-        # space. One example of such a situation is viewing a lower resolution
-        # level of an input pyramid, where _transform contains the transform
-        # only from the base level to world coordinates, another is when an
-        # image is larger than the maximum allowed texture size and has been
-        # downsampled
-        self._transform_view = ScaleTranslate(np.ones(ndim), np.zeros(ndim))
-        # Create additional transform mapping world-coordinates into a grid
-        # for looking at layers side-by-side
-        self._transform_grid = ScaleTranslate(np.ones(ndim), np.zeros(ndim))
+        # Create a transform chain consisting of three transforms:
```
GenevieveBuckley, line 148: In your comment can you also list the names used in the code:
* `view2data`
* `data2world`
* `world2grid`

napari/layers/base/base.py
```diff
@@ -305,35 +314,35 @@ def editable(self, editable):
...
     @scale.setter
     def scale(self, scale):
-        self._transform.scale = np.array(scale)
+        self._transforms['data2world'].scale = np.array(scale)
         self._update_dims()
         self.events.scale()
 
     @property
     def translate(self):
         """list: Factors to shift the layer by."""
```
GenevieveBuckley: I know this is an existing docstring, but can we clarify what this actually means. "Factors" is pretty vague, do we mean to say instead this is a shift in pixels applied to the layer? I think it's important the units.

napari/layers/base/base.py
```diff
@@ -305,35 +314,35 @@ def editable(self, editable):
...
     @translate.setter
     def translate(self, translate):
-        self._transform.translate = np.array(translate)
+        self._transforms['data2world'].translate = np.array(translate)
         self._update_dims()
         self.events.translate()
 
     @property
     def translate_grid(self):
         """list: Factors to shift the layer by."""
```
GenevieveBuckley, line 337: Ditto
> I know this is an existing docstring, but can we clarify what this actually means. "Factors" is pretty vague, do we mean to say instead this is a shift in pixels applied to the layer? I think it's important the units.

GenevieveBuckley (commented):
I've barely started looking at this and will check back later (but I did want to make sure the one or two comments I did write get saved for you)

napari/layers/base/base.py
sofroniewn: updated

napari/layers/base/base.py
sofroniewn: done

napari/layers/base/base.py
sofroniewn: updated - the next PR will really focus on the coordinates / units / names

napari/layers/base/base.py
sofroniewn: updated - the next PR will really focus on the coordinates / units / names

napari/_vispy/_tests/test_vispy_big_images.py
sofroniewn: I've renamed `s`

napari/layers/base/base.py
```diff
@@ -145,21 +145,30 @@ def __init__(
...
-        # level of an input pyramid, where _transform contains the transform
-        # only from the base level to world coordinates, another is when an
-        # image is larger than the maximum allowed texture size and has been
-        # downsampled
-        self._transform_view = ScaleTranslate(np.ones(ndim), np.zeros(ndim))
-        # Create additional transform mapping world-coordinates into a grid
-        # for looking at layers side-by-side
-        self._transform_grid = ScaleTranslate(np.ones(ndim), np.zeros(ndim))
+        # Create a transform chain consisting of three transforms:
+        # 1. `view2data`: An initial transform mapping the actually viewed data
```
GenevieveBuckley: What does "actually viewed data" mean here? Do we say `view` is the screen canvas? It would be helpful to clarify in the comment block

napari/layers/transforms.py
```diff
@@ -60,12 +72,67 @@ def set_pad(self, axes: Sequence[int]) -> 'Transform':
...
     @property
-    def inverse(self) -> 'Transform':
-        if self._inverse_func is not None:
-            return Transform(self._inverse_func, self.func)
+    def inverse(self) -> 'TransformChain':
+        """Return the inverse transform chain."""
+        return TransformChain([tf.inverse for tf in self[::-1]])
+
+    @property
+    def composite(self) -> 'Transform':
```
GenevieveBuckley: `compose` and `composite` are too similarly named & will lead to confusion. I'd suggest keeping `compose` and renaming `composite`. [Vispy's chain transform](http://vispy.org/visuals.html?highlight=chaintransform#vispy.visuals.transforms.ChainTransform) uses `simplified`, maybe we could call it something similar

napari/layers/base/base.py
```diff
@@ -145,21 +145,30 @@ def __init__(
...
+        #   cases, we need to  transform the coordinates from the displayed
+        #   data space to the original input data space. One example of such a
+        #   situation is viewing a lower resolution level of an input pyramid,
+        #   another is when an image is larger than the maximum allowed texture
+        #   size and has been downsampled so it can be viewed.
+        # 2. `data2world`: The main transform mapping data to a world-like
+        #   coordinate.
+        # 3. `world2grid`: An additional transform mapping world-coordinates
+        #   into a grid for looking at layers side-by-side.
+        self._transforms = TransformChain(
```
GenevieveBuckley, line 160: This is probably the most important point, but I've been stuck on what/how to say this.

I assume that transforms arranged in a chain will "flow" in a direction. Eg: from the data to what you see on screen, or vice versa. This section is confusing for me because it seems like each end of the transform chain (`view` and `grid`) relate to the canvas  you see on your monitor. The direction of the "flow" isn't as obvious as it should be.

It's possible I am confused because I'm mistaken and `grid` is not about the canvas view & instead is spatial. Is this the case? But if `grid` is spatial & changing the underlying position of the data in the world space, that would cause problems eg: picking points/shapes.

I mean, I guess that's exactly the behaviour we have now https://github.com/napari/napari/issues/860, and you do say "This PR is not designed to change any functionality" so maybe this is intended to be addressed in a later PR?

napari/layers/image/image.py
```diff
@@ -454,11 +454,13 @@ def _set_view_slice(self):
                         self._top_left[d] + self._max_tile_shape,
                         1,
                     )
-                self._transform_view.translate = (
-                    self._top_left * self.scale * self._transform_view.scale
+                self._transforms['view2data'].translate = (
+                    self._top_left
+                    * self.scale
```
GenevieveBuckley: I don't like that we still have a `self.scale` here, shouldn't this PR mean that all scales are tracked in transforms? This seems like information is being stored in several places (so more chance of things going wrong)

GenevieveBuckley (commented):
Sorry for the delay here, I sat down several times to do this but keep getting interrupted with other things. Comments added inline.

napari/_vispy/vispy_image_layer.py
```diff
@@ -266,7 +266,7 @@ def downsample_texture(self, data, MAX_TEXTURE_SIZE):
             scale = np.ones(self.layer.ndim)
             for i, d in enumerate(self.layer.dims.displayed):
                 scale[d] = downsample[i]
-            self.layer._transform_view.scale = scale
+            self.layer._transforms['view2data'].scale = scale
```
jni: I love this. =)

napari/layers/base/base.py
jni: I have to say I'm similarly confused... Why does the chain start at `view` and not `data`? (`grid` I get, it's spreading the world out on the canvas if desired. But, view?

jni: Ok I've re-read the paragraph and it's a little bit clearer... I don't think `view` is the best name here... `processed_data2data`? A bit too verbose...? Something we could brainstorm.

napari/layers/image/image.py
jni: I'm confused here also! What's `self._top_left`? Why is it different from `self.translate`?

napari/layers/transforms.py
```diff
@@ -28,6 +29,17 @@ def __call__(self, coords):
...
 
+    @property
+    def inverse(self) -> 'Transform':
+        if self._inverse_func is not None:
+            return Transform(self._inverse_func, self.func)
+        else:
+            raise ValueError('Inverse function was not provided.')
+
+    def compose(self, transform: 'Transform') -> 'Transform':
+        """Return the composite of this transform and the proivded one."""
```
jni:
```suggestion
        """Return the composite of this transform and the provided one."""
```

napari/layers/transforms.py
jni: I agree, this caught me out.

napari/layers/transforms.py
```diff
@@ -60,12 +72,67 @@ def set_pad(self, axes: Sequence[int]) -> 'Transform':
...
+            Axes to subset the current transform chain with.
+
+        Returns
+        -------
+        TransformChain
+            Resulting transform chain.
+        """
+        return TransformChain([tf.set_slice(axes) for tf in self])
+
+    def set_pad(self, axes: Sequence[int]) -> 'Transform':
```
jni: Ah! Can we rename this `expand_dims`? I was totally confused about pad because padding in NumPy land is of existing dims. `expand_dims` is exactly the analogous function in NumPy.

napari/layers/transforms.py
```diff
@@ -110,6 +177,12 @@ def inverse(self) -> 'ScaleTranslate':
         """Return the inverse transform."""
         return ScaleTranslate(1 / self.scale, -1 / self.scale * self.translate)
 
+    def compose(self, transform: 'ScaleTranslate') -> 'ScaleTranslate':
+        """Return the composite of this transform and the proivded one."""
```
jni:
```suggestion
        """Return the composite of this transform and the provided one."""
```

jni (changes requested):
Made some comments! I love where this is going, just needs a bit more experimenting with the names of things. I especially like @GenevieveBuckley's suggestion of `.simplified`, which generalises to more complex chains too (only the links that can be simplified would be simplified, similar to node fusion in task graphs).

napari/layers/image/image.py
GenevieveBuckley: Top left is some extra adjustment if you have pyramid data. This conversation indicates it also needs a better name :) Another one for the brainstorming list!

napari/layers/transforms.py
sofroniewn: will use `simplified`, great suggestion

napari/layers/base/base.py
sofroniewn: ok, yeah - this is showing how bad a word `view` is - maybe `processed` is better, I don't love `processed_data`, we can discuss. I need to think more, but i totally get how this is confusing and should change

napari/layers/image/image.py
sofroniewn: At this point `.scale` is just a convenience for `_transforms['data2world'].scale`. I think it's still nice to make that available for a user, but internally it's probably better to use the more detailed one.

`_top_left` is getting improved and could easily be renamed in #1053 where we fix a bunch of the pyramid stuff. Would be nice to get this one in without touching it and then make sure we fix that there. We can link this comment to that PR

napari/layers/transforms.py
sofroniewn: will do

napari/layers/base/base.py
GenevieveBuckley: Maybe `downsampled_data`?

sofroniewn: love it. Given that the way this will actually appear is `downsampled_data2data` could we get away with just `downsampled2data` or is that less clear?

sofroniewn: @jni suggested `tile2data` I will rename and add comment / explanation

napari/layers/image/image.py
sofroniewn: will add a comment about what top left is too

napari/layers/base/base.py
sofroniewn: switched to `tile2data`, explanation updated

napari/layers/base/base.py
sofroniewn: switched to `tile2data`, explanation updated

napari/layers/transforms.py
sofroniewn: switched to `simplified`

napari/layers/image/image.py
sofroniewn: added comment about top left, and put in `_transforms['data2world'].scale`. We will improve the `tile2data` / `top_left` stuff in #1053 too as soon as this merges

napari/layers/transforms.py
sofroniewn: done

napari/layers/transforms.py
sofroniewn: done

napari/layers/transforms.py
sofroniewn: done

sofroniewn:
@GenevieveBuckley @jni this has now been updated with all the requested changes and is ready for re-review. Once merged I will update #1053 and we can press on fixing all the pyramid stuff, at which point I think things will be much cleaner :-) Thanks both for all your help on this!!

napari/layers/base/base.py
```diff
@@ -146,22 +146,22 @@ def __init__(
...
-        #   to the data coordinates. This is useful for cases where the
-        #   displayed data is not the base data given by the user. In such
-        #   cases, we need to  transform the coordinates from the displayed
-        #   data space to the original input data space. One example of such a
-        #   situation is viewing a lower resolution level of an input pyramid,
-        #   another is when an image is larger than the maximum allowed texture
-        #   size and has been downsampled so it can be viewed.
+        # 1. `tile2data`: An initial transform only needed displaying tiles
+        #   of an image. It maps pixels of tile into the pixels of the tile
+        #   into the coordinate space of the full resolution data and can
```
GenevieveBuckley:
```suggestion
        #   of an image. It maps pixels of tile into
        #   the coordinate space of the full resolution data and can
```

napari/layers/_tests/test_transforms.py
```diff
@@ -34,20 +35,24 @@ def test_scale_translate_compose():
...
+        scale=[2, 1, 3], translate=[8, 3, -5], name='st'
+    )
     npt.assert_allclose(transform_b.set_slice([0, 2]).scale, transform_a.scale)
     npt.assert_allclose(
         transform_b.set_slice([0, 2]).translate, transform_a.translate
     )
+    assert transform_b.set_slice([0, 2]).name == 'st'
 
 
 def test_scale_translate_pad():
```
jni:
```suggestion
def test_scale_translate_expand_dims():
```

napari/layers/_tests/test_transform_chain.py
```diff
@@ -0,0 +1,74 @@
...
+    new_coord_2 = transform_chain_b(coord)
+    npt.assert_allclose(new_coord_1, new_coord_2)
+
+
+def test_transform_chain_expanded():
+    coord = [10, 3, 13]
+    transform_a = ScaleTranslate(scale=[2, 1, 3], translate=[8, 0, -5])
+    transform_b = ScaleTranslate(scale=[0.3, 1, 1.4], translate=[-2.2, 0, 3])
+    transform_c = ScaleTranslate(scale=[2, 3], translate=[8, -5])
+    transform_d = ScaleTranslate(scale=[0.3, 1.4], translate=[-2.2, 3])
```
jni, line 66: I feel like these transforms should be provided by a fixture rather than copy-pasted, but it's not a huge deal.

napari/layers/image/image.py
```diff
@@ -454,11 +454,15 @@ def _set_view_slice(self):
                         self._top_left[d] + self._max_tile_shape,
                         1,
                     )
-                self._transform_view.translate = (
-                    self._top_left * self.scale * self._transform_view.scale
+                # Note that top left marks the ocation of top left canvas
```
```suggestion
                # Note that top left marks the location of top left canvas
```

napari/layers/image/image.py
```diff
@@ -454,11 +454,15 @@ def _set_view_slice(self):
...
+                # pixel in image
```
```suggestion
                # pixel in image in data coordinates
```
?

napari/layers/transforms.py
```diff
@@ -60,12 +72,67 @@ def set_pad(self, axes: Sequence[int]) -> 'Transform':
...
+    @property
+    def simplified(self) -> 'Transform':
+        """Return the composite of the transforms inside the transform chain."""
+        if len(self) == 0:
+            return None
+        if len(self) == 1:
+            return self[0]
         else:
-            raise ValueError('Inverse function was not provided.')
+            return tz.pipe(self[0], *[tf.compose for tf in self[1:]])
```
jni, line 103: btw I love this. =D

jni (approved):
@sofroniewn I've made some very minor suggestions, nothing blocking. Approving! 🎉

napari/layers/_tests/test_transforms.py
sofroniewn: done

napari/layers/image/image.py
sofroniewn: done

napari/layers/image/image.py

napari/layers/_tests/test_transform_chain.py
sofroniewn: will leave as discussed as it might actually be more readable as is

napari/layers/base/base.py
sofroniewn: fixed!

sofroniewn:
Merging now, very excited to get this in, thanks for all the help @GenevieveBuckley @jni! Will merge this into #1053 and then press on with things there

PR #1046, opened 2020-03-13
Shapes layer internal renaming

sofroniewn:
# Description
This PR just does some internal renaming to make the Shapes layer consistent with the naming conventions adopted in #1010, #1033. It might cause some conflicts with #898 but they should be pretty easy to resolve. After this PR merges I'll work on actually doing the mouse bindings refactor, but that will be easier to review with all this renaming already done. Note this breaks nothing with our public API. I will make a new PR in parallel to propagate the renaming changes to the rest of the repo.


napari/layers/shapes/_shapes_list.py
```diff
@@ -1,11 +1,11 @@
...
-from .mesh import Mesh
-from ._constants import shape_classes, ShapeType
+from ._mesh import Mesh
+from ._shapes_models import Shape, Line, Path
+from ._shapes_utils import inside_triangles, triangles_intersect_box
+from ._shapes_constants import shape_classes, ShapeType
 
 
-class ShapeList:
+class ShapesList:
```
jni: Not in love with this change...? ShapeList matches LayerList which matches "car collection" which matches "card deck"...

jni (changes requested):
@sofroniewn honestly I generally prefer the the singular name, but I specifically prefer it in the NounList form — which matches English usage in general, "Item Collection" rather than "Items Collection". For the other stuff, it's internal so I don't mind so much.

napari/layers/shapes/_shapes_list.py
sofroniewn: I will revert

sofroniewn:
This has now been updated. `ShapeList` is singular, but everything else is plural, `add_shapes`, `Shapes` etc. This is keeping with the current API. Changing it would be a breaking change. Is this now ok @jni?

jni (approved):
Yes yes, perfect, a "ShapeList" contains "Shapes", this is totally fine and sensical and non-breaking changes are good. =)

PR #1047, opened 2020-03-13
Internal utils / keybindings / layer accessories renaming

sofroniewn:
# Description
This PR completes the renaming in #1010, #1033, #1046 by extending it to the remaining layers and any remaining places in the repo. It only affects internal code and causes no changes to our public API. It makes things more clear so `image/_constants.py` -> `image/_image_constants.py` and `napari/keybindings.py` -> `napari/_viewer_key_bindings.py` for example. The PR changes no functionality


kne42 (approved):
LGTM

napari/__init__.py
```diff
@@ -43,7 +43,7 @@
 vispy_logger.setLevel(logging.WARNING)
 
 from .viewer import Viewer
-from . import keybindings
+from . import _viewer_key_bindings
```
jni: Why does this need to be imported here if it's not used and it's private?

jni (approved):
@sofroniewn I've just raised a minor issue, but fix/ignore as appropriate then feel free to merge. Thank you! =)

napari/__init__.py
sofroniewn: The import is needed as inside that file the Viewer object get's decorated with the keybindings. Before the method was not private, but probably shouldn't have been used. We use this pattern with our layers quite frequently. Maybe I should add a `del _viewer_key_bindings` after the import so it doesn't pollute the namespace? What do you and @kne42 think?

sofroniewn: I will add a comment about the import doing the decoration too. I will propagate this change into all the keybindings files

sofroniewn: done

sofroniewn:
The requested `del _viewer_key_bindings` and comment have been added. This is ready for re-review @jni. If you like this comment and approach I will propagate it to the layer specific PRs before they merge.

jni (commented):
@sofroniewn I guess you need to add `  # noqa: F401` to the import now, same as `from scipy import stats`. I also would move the comment from the deletion to the import, again, as with the stats import.

sofroniewn:
merging now that tests have passed. thanks all

Issue #1048, opened 2020-03-13
mouse bindings generalization proposal


kne42:
# Introduction
As called for in #793, following is my proposal for the interface for the improvement of the registration & definition of custom mouse functions. This proposal streamlines registration into one style and better defines interactions between multiple custom mouse functions.

# Registration
There will be 4 types of mouse bindings: **click** (this includes press, release, and drag), **move**, **double click**, and **scroll**.

You bind similarly to how you bind keys, on to an instance or the class itself:
```python
@Image.bind_mouse('click')
def click_func(layer, event):
    ...
```

You can register multiple mouse bindings to be active at once. Whether they are active will be determined by a truth function.

```python
@Image.bind_mouse('scroll', predicate=lambda layer: layer.mode == 'pan/zoom')
def scroll_func(layer, event):
    ...
```

They are internally stored as a dictionary of `name: (predicate, func)`. The name is taken from `func.__name__` but can be provided in the event that two functions share the same name. It must be a valid python identifier.

```python
@image.bind_mouse('click', name='add_point')
def click_func(layer, event):
    ...
```

Thus, one can replace a function or its predicate by specifying the same name.

```python
image.bind_mouse('click', name='add_point', predicate=False)  # disable point adding
image.bind_mouse('scroll', name='scroll_func', func=zoom)  # change scroll function to zooming
```

To prevent confusion, if the `name` kwarg is not explicitly provided, replacing an existing function or predicate will result in an error.

# Function Definitions
Mouse bindings will receive two arguments: `layer` and `event`. Instead of using the `vispy` events, we'll pass our own as `namedtuple`s to add a layer of abstraction.

<details><summary><strong>event definitions</strong></summary>

### MouseButton Enum
1 = Left
2 = Right
3 = Middle
4 = Back
5 = Forward

### ClickEvent
- **screen_pos**: *(x, y) tuple* - position on the 2D view of the screen
- **worldview_pos**: *(a0, a1) or (a0, a1, a2) tuple* - position in 2- or 3-D in the current view of the canvas
- **modifiers**: *tuple of {'Shift', 'Control', 'Alt', 'Meta'}* - all modifier keys currently being pressed
- **button**: *MouseButton or None* - mouse button that triggered this event
- **buttons**: *tuple of MouseButton* - all mouse buttons currently being pressed
- **is_dragging**: *bool* - if this event is during the moving segment of a drag event
- **native**: *object* - native event

### MoveEvent
[...]

PR #1051, opened 2020-03-14
shapes layer mouse bindings refactor

sofroniewn:
# Description
This PR is a follow on from #1046 (which should merge first) that actually refactors the shapes mouse bindings. This PR still needs doc strings and tests, I will add shortly.

## Type of change

- [x] Refactor


sofroniewn:
Tests and doc strings have now been added, #1046 should still be merged first, and this PR should then be updated. I'll also note that a lot of the tests for shapes mouse bindings have some similarities with tests for points mouse bindings in #1033

sofroniewn:
This PR has now been update and is ready for review. If you like the use of a `set` instead of a `list` for the `selected_data` in the points layer (implemented in #1033) then I should do the same conversion here for `selected_data` here too for consistency. Curious what @kevinyamauchi thinks on this too.

napari/layers/shapes/_shapes_mouse_bindings.py
```diff
@@ -0,0 +1,257 @@
...
+    layer._is_moving = False
+    layer._drag_start = None
+    layer._drag_box = None
+    layer._fixed_vertex = None
+    layer._moving_value = (None, None)
+    layer._set_highlight()
+    layer._update_thumbnail()
+
+
+def add_line_rectangle_ellipse(layer, event):
```
kne42: I would create a function for each mode and use a helper function.

e.g.
```python
def add_line(layer, event):
    ...
    yield from _add_line_rectangle_ellipse(layer, event, data=np.array([corner, corner + size]), shape_type='line')
```

napari/layers/shapes/_shapes_mouse_bindings.py
```diff
@@ -0,0 +1,257 @@
...
+        layer._moving_value = copy(layer._value)
+        data_full = layer.expand_shape(vertices)
+        layer._data_view.edit(index, data_full, new_type=new_type)
+        layer._selected_box = layer.interaction_box(layer.selected_data)
+
+
+def add_path_polygon_creating(layer, event):
+    """While a path or polygon move next vertex to be added."""
+    if layer._is_creating:
+        coord = [layer.coordinates[i] for i in layer.dims.displayed]
```
kne42: this line is repeated a lot. it should be made into a helper function

napari/layers/shapes/_shapes_mouse_bindings.py
```diff
@@ -0,0 +1,257 @@
...
+            layer.selected_data = []
+    elif layer._is_selecting:
+        layer.selected_data = layer._data_view.shapes_in_box(layer._drag_box)
+        layer._is_selecting = False
+        layer._set_highlight()
+    layer._is_moving = False
+    layer._drag_start = None
+    layer._drag_box = None
+    layer._fixed_vertex = None
+    layer._moving_value = (None, None)
```
kne42, line 63: where possible, could you convert these to local variables instead of putting them on the layer?

kne42 (changes requested):
nice work! left a few comments on using helper functions / whatnot

kevinyamauchi:
> If you like the use of a set instead of a list for the selected_data in the points layer (implemented in #1033) then I should do the same conversion here for selected_data here too for consistency. Curious what @kevinyamauchi thinks on this too.

I think making `selected_data` a set makes sense!

sofroniewn:
I will make `selected_data` a set here too then as it looks like #1033 is close to merge

napari/layers/shapes/_shapes_mouse_bindings.py
sofroniewn: Unfortunately all of these occur in the main shapes model too right now. The biggest offender is the `layer._move` function. That could probably be pulled in here to be local too, but that would be a much bigger lift and I'd like to save that for a follow up PR

napari/layers/shapes/_shapes_mouse_bindings.py
sofroniewn: should I add a property to the layer `layer.displayed_coordinates`?

napari/layers/shapes/_shapes_mouse_bindings.py
sofroniewn: done, though this did mean I had to add some nested if statements inside the shapes model, curious if you prefer it, have ideas on how it can be improved

napari/layers/shapes/_shapes_mouse_bindings.py
sofroniewn: I can put the property on the base layer and use it where needed

sofroniewn: change made, thanks for the suggestion - it was present in a lot of places!!!

napari/layers/shapes/_shapes_mouse_bindings.py
sofroniewn: done

sofroniewn:
@kne42 changes have been made. This is now ready for re-review @jni.

I also made the `selected_data` property a set, which might have implications for #898 @HagaiHargil

sofroniewn:
Note this PR also closes #1116 by making the escape key on polygon no longer draw one more vertex.

napari/layers/shapes/_shapes_mouse_bindings.py
```diff
@@ -0,0 +1,272 @@
...
+    """Select shapes or vertices either in select or direct select mode.
+
+    Once selected shapes can be moved or resized, and vertices can be moved
+    depending on the mode. Holding shift when resizing a shape will preserve
+    the aspect ratio.
+    """
+    shift = 'Shift' in event.modifiers
+    # on press
+    layer._moving_value = copy(layer._value)
+    if layer._value[1] is None:
```
jni: Can you make an issue to rename _value to something else? It’s really confusing to read this code. 

Actually, in this particular instance, one way to clarify the code is to say, for example:

```python
shape_under_cursor = layer._value
```

And then proceed with the renamed variable.

jni: The rename should also include a note about what the two components are, or maybe even better, you can unpack them into two variables.

napari/layers/shapes/_shapes_mouse_bindings.py
```diff
@@ -0,0 +1,272 @@
...
+    vertices = np.insert(vertices, ind, [coord], axis=0)
+    with layer.events.set_data.blocker():
+        data_full = layer.expand_shape(vertices)
+        layer._data_view.edit(index, data_full, new_type=new_type)
+        layer._selected_box = layer.interaction_box(layer.selected_data)
+    layer.refresh()
+
+
+def vertex_remove(layer, event):
+    """Remove a vertex from a selected shape."""
```
jni: Can you flesh out this docstring? And above also?

napari/layers/shapes/_shapes_mouse_bindings.py
```diff
@@ -0,0 +1,272 @@
...
+    if layer._value[1] is not None:
+        # have clicked on a current vertex so remove
+        index = layer._value[0]
```
jni: Yeah! Exactly like this! Now do it everywhere layer._value appears.

napari/layers/shapes/_shapes_mouse_bindings.py
```diff
@@ -0,0 +1,272 @@
...
+    elif shape_type == Rectangle:
+        # Adding vertex to rectangle turns it into a polygon
+        new_type = Polygon
+    else:
+        new_type = None
+    closed = shape_type != Path
+    vertices = layer._data_view.displayed_vertices[
+        layer._data_view.displayed_index == index
+    ]
+    if closed is not True:
```
jni: `if not closed`

napari/layers/shapes/_shapes_mouse_bindings.py
jni: I particularly want to know the types of the arguments and what attributes they are expected to contain.

napari/layers/shapes/_shapes_mouse_bindings.py
```diff
@@ -0,0 +1,272 @@
...
+
+
+def add_path_polygon_creating(layer, event):
+    """While a path or polygon move next vertex to be added."""
+    if layer._is_creating:
+        layer._move(layer.displayed_coordinates)
+
+
+def vertex_insert(layer, event):
+    """Insert a vertex into a selected shape."""
```
jni: I found this function and the one below it quite hard to follow. I would suggest breaking it up into smaller functions eg for each shape type.

napari/layers/shapes/shapes.py
```diff
@@ -559,20 +586,35 @@ def mode(self, mode):
...
+            if mode == Mode.VERTEX_INSERT:
+                self.mouse_drag_callbacks.append(vertex_insert)
+            else:
+                self.mouse_drag_callbacks.append(vertex_remove)
+            self.mouse_move_callbacks.append(highlight)
         elif mode in [Mode.ADD_RECTANGLE, Mode.ADD_ELLIPSE, Mode.ADD_LINE]:
             self.cursor = 'cross'
             self.interactive = False
             self.help = 'hold <space> to pan/zoom'
+            if mode == Mode.ADD_RECTANGLE:
```
jni, line 604: For this list I would consider creating a mode2function dictionary and having a single call to append

jni (commented):
@sofroniewn I’ve suggested a few changes to code structure and docstrings that I think should go in before merging, but generally this is good. I’ll mention that the “mode2function” suggestion is optional.

kne42 (approved):

napari/layers/shapes/_shapes_mouse_bindings.py
sofroniewn: Issue made https://github.com/napari/napari/issues/1120

expanded as 
```python
    shape_under_cursor, vertex_under_cursor = layer._value
```

napari/layers/shapes/_shapes_mouse_bindings.py
sofroniewn: done

napari/layers/shapes/_shapes_mouse_bindings.py
sofroniewn: done

napari/layers/shapes/shapes.py
sofroniewn: this will get redone in #1048 so i'd rather leave it for now

napari/layers/shapes/_shapes_mouse_bindings.py
sofroniewn: I've cleaned up some code and added a bunch of comments. hopefully this makes it more clear. i'm not sure splitting would help as much

napari/layers/shapes/_shapes_mouse_bindings.py
sofroniewn: Doc strings expanded. the arguments `layer` and `event` are the same for all of our mouse bindings, but they will change in #1048. I'd like to just get this in as is

sofroniewn:
Changes made @jni can you rereview

napari/layers/shapes/_shapes_mouse_bindings.py
jni: I'm going to merge, but just to reiterate a rule of thumb I love: functions should fit in a screen, ie 25-50 lines max, so that you can "at a glance" tell what it's doing. Humans are not good at holding context over 100 lines of code, so by the time I'm looking at the end of the function, I have forgotten what was happening at the start, and I just have to pray that there are not any non-local effects at play.

(Of course, this is not helped if you call a function `_super_complex_helper(var1, var2, ..., var100)` in the middle. But typically you can break it down into *sensible* chunks that can be read easily, e.g. `validate_event(event, attributes_required)` or similar. Then I can instantly guess at what it's supposed to do, trust that it's doing it's job because I will review that bit of self-contained code by itself later, and move on to the rest of the function.)

sofroniewn: yes for sure, thanks for merging - definitely something to continue to improve

Issue #845, opened 2020-01-03
Make changing views non-blocking


sofroniewn:
## 🚀 Feature

In https://github.com/napari/napari/issues/763#issuecomment-562552657 @jni wrote:

> Rearchitecting how we update the viewer so that changing views isn't blocking. Right now this last one is a big one, and could kind of happen independent of the transforms thing. Like, if I move the slider quickly and it's a slow process (e.g. Talley's live deconv example), we should be smart enough to toss out the intermediate tasks. The slider itself should never block waiting for the display. This might be reasonably straightforward to achieve with a task queue: rather than just calling np.array() on a dataset, add the call to np.array to a queue, and have that queue hold at most nlayers tasks. Then, as soon as a new event is triggered, drop running queue jobs and start again (unless the hash of the job hasn't changed).

This seems like a nice feature that should be relatively self contained. Curious if people have more thoughts on how / where it should be implemented. Might give us some big usability improvements for relatively little effort. @tlambert03 @jni?

jni:
There's a bit of opportunity to make things generic here. We could build the queuing mechanism into the base layer, which takes as input three functions to do slicing, viewport cropping, and instantiating. These would vary between the layer types but all layer types would need to provide them (even if it's just an identity — which could be implemented in the base layer). Instantiating is `np.asarray` for Image layers, but `np.asarray(_raw_to_displayed())` for labels, etc. Slicing and viewport cropping are done by `__getitem__` and might include subsetting as we do now for big images, but for points it is done by selection.

All of the above would be done asynchronously and this would be managed in a single spot.

Additionally, after the framework is in place, the functions themselves could be made asynchronous, meaning they can yield results progressively as they compute them.

sofroniewn:
I like that idea a lot, some stuff could go inside the `refresh` function https://github.com/napari/napari/blob/53d95988fee000f5fa0a1c5e567e09ea05d80635/napari/layers/base/base.py#L573 now on the base layer, by breaking apart `_set_view_slice` so it doesn't actually do the instantiation

d-v-b:
Yall are probably already thinking this, but making a view change non-blocking is a special case / side effect of lazy tiled loading (#318), so hopefully resolving one of these issues helps resolve the other.

mrocklin:
Does Napari use any sort of event loop, async/await functions, or background threads? Looking online at docs it seems like Qt might have its own thing here.  

[...]

PR #1053, opened 2020-03-14
Fix pyramid clipping

### LINKED ISSUE #458: Incorrect clipping at bottom edge of pyramid

### LINKED ISSUE #1050: big image (68 megapixels) does not display entirely
## 🐛 Bug

I am trying to display a very large image (6898 x 9946 pixels), but only the part of it is displayed: the bottom part is black. If I display only a part of the image, everything works fine. From my experience, problems start to occur from image size > ~55 megapixels.

## To Reproduce

This is the code to reproduce the issue with a random image:

img = np.random.randint(0, 100, size=(6898, 9946))

with napari.gui_qt(): 
    ...:     v = napari.view_image(img)

This is the result that I get:

partly_displayed_image

Displaying the first 8000 pixels in x gives the result that I expect:

with napari.gui_qt(): 
    ...:     v = napari.view_image(img[:,:8000]) 

expected_result

The problem starts to occur starting from the x-size of ~8200 pixels

## Environment

napari: 0.2.12
Platform: Linux-3.10.0-1062.4.3.el7.x86_64-x86_64-with-redhat-7.7-Maipo
Python: 3.7.6 (default, Jan 8 2020, 19:59:22) [GCC 7.3.0]
Qt: 5.14.1
PySide2: 5.14.1
VisPy: 0.6.4
NumPy: 1.18.1
SciPy: 1.4.1
scikit-image: 0.16.2
Dask: 2.12.0

GL version: 4.6.0 NVIDIA 440.31
MAX_TEXTURE_SIZE: 32768

sofroniewn:
# Description
This PR follows on from #1042 (though strictly speaking it maybe could have come sooner) and fixes the pyramid clipping issues in #458 and #1050. I still need to add tests, improve doc strings and maybe add some comments - but it would be good to know if this fixes the problem for as many use cases as we know are having it (see https://forum.image.sc/t/efficiently-open-qptiff-images-with-napari/33237/4 too).

The approach is to pick a good zoom level and then find the locations of the top right and bottom left canvas pixel in world-slice coordinates and then make sure the viewed data always includes that data.

## Type of change

# References

Closes #458, closes #1050


sofroniewn:
Note context for this PR @jni was discussion here https://forum.image.sc/t/using-napari-as-imagepys-canvas/32579/10?u=sofroniewn

sofroniewn:
@GenevieveBuckley @jni I have merged in the latest chain transform and this is now ready for some preliminary review. I imagine we'll want to rename a bunch of things that are confusing, and I still need to put some doc strings in etc, but i'm also curious if the functionality at least works for you both

napari/_vispy/vispy_base_layer.py
```diff
@@ -47,6 +48,8 @@ def __init__(self, layer, node):
         self.layer = layer
         self.node = node
 
+        self.tile_rounding = 400
```
GenevieveBuckley: 1. Should this be a private attribute?

2. Are these tiles always square? Is that a good assumption to have here?

3. You mentioned the value here as a very arbitrary choice, can you add a comment describing:
a. What "tile_rounding" is (elsewhere in a comment you call it the "max allowed tile size" for the vispy canvas).
b. The fact that this value is an arbitrary choice.
c. Why you picked 400. (Even if the reasons are super vague along the lines of "it seemed like a good idea at the time" that's still helpful to future developers. It helps people have a sense of where this falls on the scale of: sure, you can change this ---> this number is magic & no-one really understands why things break if you set it to something different.)

napari/_vispy/vispy_base_layer.py
```diff
@@ -134,7 +137,7 @@ def _on_scale_change(self, event=None):
         # convert NumPy axis ordering to VisPy axis ordering
         self.scale = scale[::-1]
         if self.layer.is_pyramid:
-            self.layer.top_left = self.find_top_left()
+            self.layer.corner_pixels = self.find_extrema()
```
GenevieveBuckley: I'd say we're moving in the right direction with renaming functions, but we're not quite there yet. It's possible to confuse extrema with finding max/min of intensity values, or a bunch of other things.

I know this is nitpicky, if we can collectively think of something a bit better that'd be great, otherwise it's not a huge issue.

napari/_vispy/vispy_base_layer.py
```diff
@@ -204,9 +290,19 @@ def on_mouse_release(self, event):
...
+        data is sent to the canvas or the camera is moved.
         """
         self.layer.scale_factor = self.scale_factor
+        if self.layer.is_pyramid:
+            self.layer.scale_factor = self.scale_factor
+            size = self.camera.rect.size
+            data_level = self.compute_data_level(size)
+
+            if data_level != self.layer.data_level:
+                self.layer.data_level = data_level
```
GenevieveBuckley: Why doesn't `self.layer.corner_pixels = self.find_extrema()` get called if the `data_level` changes?

GenevieveBuckley (commented):
I've made some comments.
I think my most useful suggestion here is the one from this meeting - we should try and add a test using the `screenshot()` function and the reproducible example of the clipping bug from https://github.com/napari/napari/issues/1050

tlambert03:
Hey @sofroniewn, just playing with some tests for this, and I came across something I don't quite understand:  I was thinking it would make sense to make a pyramid of ones and test without a screenshot by setting layer.position = (x, y) and looking at layer._get_value(), like this:

```python
def test_pyramid_clipping(viewer_factory):
    img = np.ones((6898, 9946))
    layer = viewer.add_image(img)
    layer.position = (0, 0)
    assert layer._get_value() == (3, 1.0)

    # doesn't work
    layer.position = (6897, 9945)
    assert layer._get_value() == (3, 1.0)
   
    # but works with a non-pyramid
    layer = viewer.add_image(np.ones((256, 256)))
    layer.position = (255, 255)
    assert layer._get_value() == 1
```


However, the coords don't behave as I'd expect with pyramid layers (try `napari.view_image(np.ones((6898, 9946)))`).  When I move the mouse and look at layer.position, the X coord (row) is as expected, but the Y coord is about 8 times smaller than expected.  Furthermore, `layer.position` is sensitive to zoom in a pyramid.  If I keep my mouse on the bottom right pixel, but zoom in a lot, the `layer.position` changes a bunch (which is not the case for non-pyramids).  Is that the expected behavior?  I would have assumed that `_transform_position` would always output coordinates in image space... but not sure what it's supposed to do for pyramids... or why it should be different for X vs Y

sofroniewn:
hmm @tlambert03 - i need to dig in, but it looks like that might be related to the mysterious factor of 2 popping up here https://github.com/napari/napari/pull/1102/files#r404226219. i will investigate

napari/_vispy/vispy_base_layer.py
sofroniewn: this arbitrary parameter was dropped!! now all calculations are done using the canvas size, thanks @jni and @GenevieveBuckley for pushing me to really think about this one

napari/_vispy/vispy_base_layer.py
sofroniewn: agreed - still needs to be improved

napari/_vispy/vispy_base_layer.py
sofroniewn: comment added - it is because setting `data_level` will trigger computation that recalculates the corner pixels now that the `data_level` has changed.

sofroniewn:
> However, the coords don't behave as I'd expect with pyramid layers (try napari.view_image(np.ones((6898, 9946)))). When I move the mouse and look at layer.position, the X coord (row) is as expected, but the Y coord is about 8 times smaller than expected.

So part of the weirdness is that the autopyramid generation - which we want to drop after this PR has only kicked in for the second axis, above the 8095 threshold, so your data has only been pyramided in one axis!! Quite strange. That doesn't mean that what you did shouldn't work, but it does explain the asymmetry that you have observed.

I have made some fixes to this PR, including dropping the "arbitrary" rounding parameter. I will go through tomorrow and add more tests and think about the naming a little more carefully. Hopefully this is close to being done, and will be a lot cleaner when it is finished

tlambert03:
> so your data has only been pyramided in one axis!!

ah!  that explains it!  thanks.

napari/_vispy/vispy_base_layer.py
sofroniewn: name made more explicit

sofroniewn:
@jni @GenevieveBuckley - I have gone through now and added a series of tests for this including screenshot ones. It is now ready for review - all comments should have been addressed!

napari/_vispy/_tests/test_vispy_pyramid.py
```diff
@@ -0,0 +1,149 @@
...
+#
+#     # Test position at corner of image
+#     layer.position = target_position
+#     value = layer.get_value()
+#     np.testing.assert_allclose(value, (target_level, data[target_level][target_position]))
+#
+#     # Test position at outside image
+#     layer.position = (40, 20)
+#     value = layer.get_value()
+#     assert value[1] is None
```
jni: Should this be uncommented or deleted?

napari/_vispy/_tests/test_vispy_pyramid.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+def test_pyramid(viewer_factory):
+    """Test rendering of pyramid data."""
+    view, viewer = viewer_factory(show=True)
+
+    shapes = [(4000, 2000), (2000, 1000), (1000, 500), (500, 250)]
+    np.random.seed(0)
+    data = [np.random.random(s) for s in shapes]
+    _ = viewer.add_image(data, is_pyramid=True)
```
jni: should you set `contrast_limits=[0, 1]` since we know them?

napari/_vispy/vispy_base_layer.py
```diff
@@ -192,10 +190,108 @@ def _reset_base(self):
...
+        size = size[self.layer.dims.displayed]
+        ds = self.layer.level_downsamples[:, self.layer.dims.displayed]
+
+        # Consider largest dimension after scaling by downsamples
+        ratio = (size / ds).max(axis=1)
+
+        # Find the maximum allowed size
+        level = np.argmin(ratio > size_threshold)
+
+        return level
```
jni: I find this method quite confusing...

(a) why is this in the base layer? I understand that we want multiscale for layers other than `image`, but then the metaphors start to break down... `maximum size in pixels`... for points? `level_downsamples` isn't a great name imho... `downsample_factors`? 
(b) It also feels like it should be a function that takes in size, maximum size, and downsample factors and returns the level. The function should be called already with the displayed dims subsampled. One reason this bugs me is that it relies on values that are determined outside of the function (`self.layer.dims.displayed` and `self.layer.level_downsamples`), which makes it difficult to follow the logic — I have to go figure out what those are elsewhere.
(c) 'ratio' is also not a great name... effective_size?
(d) although documented, I think it's an implementation detail that argmin returns the first element that is 0 here. Perhaps `np.argwhere(effective_size < size_threshold)[0]`? With a note that it is assumed that the downsample factors are sorted? Or, at least, if using `argmin`, add a comment saying, "find the first element for which the effective size is lower than the size threshold."
(e) Finally, I *think* you actually want the *last* element for which the effective size is greater than the threshold, assuming your threshold is the canvas size. Specifically, you want there to be *at least* one data pixel per screen pixel. With the current method, you are always guaranteed to have a slightly blurrier dataset than you have pixels on your monitor.

napari/_vispy/vispy_base_layer.py
```diff
@@ -192,10 +190,108 @@ def _reset_base(self):
...
+        ----------
+        corner_pixels : array
+            Coordinates of top left and bottom right canvas pixel in the data.
+        size : array
+            Size of requested tile in data coordinates.
+        """
+        nd = self.layer.dims.ndisplay
+        # Find image coordinate of top left canvas pixel
+        if self.node.canvas is not None:
+            offset = self.translate[:nd] / self.scale[:nd]
```
jni, line 206: Does this logic depend on the displayed dimensions not being transposed? Or does the transposition already affect `self.translate` and `self.scale` directly?

napari/_vispy/vispy_base_layer.py
```diff
@@ -192,10 +190,108 @@ def _reset_base(self):
...
+            tl_raw = [0] * nd
+            br_raw = [1] * nd
+
+        top_left = np.zeros(self.layer.ndim)
+        bottom_right = np.zeros(self.layer.ndim)
+        for d, tl, br in zip(self.layer.dims.displayed, tl_raw, br_raw):
+            top_left[d] = tl
+            bottom_right[d] = br
+
+        # Clip according to the max data of the level shape
```
jni: Why is clipping necessary? I'd kinda rely on users of this function to do the clipping, since the function promises the coordinates of the canvas pixel, not the smaller of the canvas pixel and the bottom right canvas pixels...

napari/_vispy/vispy_base_layer.py
```diff
@@ -192,10 +190,108 @@ def _reset_base(self):
...
+        # Consider largest dimension after scaling by downsamples
+        ratio = (size / ds).max(axis=1)
+
+        # Find the maximum allowed size
+        level = np.argmin(ratio > size_threshold)
+
+        return level
+
+    def find_coordinates_of_corners_of_canvas(self):
+        """Find location of the corners of canvas in data coordinates.
```
jni, line 193: Does this function make sense if ndisplay = 3?

napari/_vispy/vispy_base_layer.py
```diff
@@ -192,10 +190,108 @@ def _reset_base(self):
...
+        size = (bottom_right - top_left) * self.layer.level_downsamples[
+            self.layer.data_level
+        ]
+
+        return np.array([top_left, bottom_right]).astype(int), size
+
     def on_draw(self, event):
-        """Called whenever the canvas is drawn.
+        """Called whenever the canvas is drawn, which happens whenever new
+        data is sent to the canvas or the camera is moved.
```
jni:
```suggestion
        """Called whenever the canvas is drawn.
        
         This is triggered whenever new data is sent to the canvas
         or the camera is moved.
```

how is this triggered? If it's an automagic thing from VisPy, it's worth pointing it out. If it's some place in our code, that is also worth pointing out.

napari/_vispy/vispy_base_layer.py
```diff
@@ -192,10 +190,108 @@ def _reset_base(self):
...
         """
         self.layer.scale_factor = self.scale_factor
+        if self.layer.is_pyramid:
+            corner_pixels, size = self.find_coordinates_of_corners_of_canvas()
+            data_level = self.compute_data_level(size)
+
+            if data_level != self.layer.data_level:
+                # Set the data level, which will trigger further updates
+                # including recaluation of the corner_pixels for the new
```
jni:
```suggestion
                # including recalculation of the corner_pixels for the new
```

napari/layers/image/image.py
```diff
@@ -475,51 +475,43 @@ def _set_view_slice(self):
...
-                ).astype(int)
-                downsampled_indices = np.clip(
-                    downsampled_indices, 0, self.level_shapes[-1, not_disp] - 1
-                )
-                indices[not_disp] = downsampled_indices
-                thumbnail = np.asarray(
-                    self._data_pyramid[-1][tuple(indices)]
-                ).transpose(order)
+            if np.product(image.shape) == 0:
+                image = np.zeros((1, 1))
```
jni: what's going on here exactly?

napari/layers/image/image.py
```diff
@@ -475,51 +475,43 @@ def _set_view_slice(self):
...
+
+            # Slice thumbnail
+            indices = np.array(self.dims.indices)
+            downsampled_indices = (
+                indices[not_disp] / self.level_downsamples[-1, not_disp]
+            )
+            downsampled_indices = np.round(
+                downsampled_indices.astype(float)
```
jni, line 496: `.astype(float)` is not needed here: division always results in float in Py3.

napari/layers/image/image.py
```diff
@@ -475,51 +475,43 @@ def _set_view_slice(self):
...
+            ).astype(int)
+            downsampled_indices = np.clip(
+                downsampled_indices, 0, self.level_shapes[-1, not_disp] - 1
+            )
+            indices[not_disp] = downsampled_indices
```
jni: This is a heck of a lot of work for a thumbnail, and I expect that this work is also done elsewhere? Could we not reuse that code/computation?

I also generally dislike reusing a variable name over and over during processing, and, given black formatting, I might prefer a single call:

```python
# find index in the current pyramid level
downsampled_indices = np.clip(
    np.round(
        indices[not_disp] / self.downsample_factors[-1, not_disp]
    ).astype(int),
    0,
    self.level_shapes[-1, not_disp] - 1,
)
```

napari/layers/image/image.py
```diff
@@ -475,51 +475,43 @@ def _set_view_slice(self):
...
+            thumbnail = np.asarray(
+                self._data_pyramid[-1][tuple(indices)]
+            ).transpose(order)
```
jni, line 504: I thought thumbnails had a standard size (32x32)? If so, then could we rename this e.g. `thumbnail_source`?

napari/layers/image/image.py
jni: Regarding reuse, I'm thinking something like, `def slice_index_convert_to_level`, then use that function with `level=-1`. I presume you would use it when actually rendering on the canvas? **Edit:** Yeah, I see that you do the exact same thing above but with the current level rather than the smallest level.

jni (commented):
@sofroniewn this is close! There little bits of the code where I left comments where I think things could be improved. In a couple of spots it would be good to write things as a function rather than relying on a bunch of `self.whatever` attributes — which make it harder to reason about what that function is doing, since there isn't a local definition of that attribute. At a minimum, it would be good to have an explanatory comment about what that variable contains, like you would have in a function docstring if it was an argument.

napari/_vispy/_tests/test_vispy_pyramid.py
sofroniewn: deleted

napari/_vispy/_tests/test_vispy_pyramid.py
sofroniewn: done

napari/_vispy/vispy_base_layer.py
sofroniewn: done

napari/_vispy/vispy_base_layer.py
sofroniewn: done

napari/_vispy/vispy_base_layer.py
sofroniewn: Clipping is needed for the size calculation that comes inside this function too and for all subsequent functions. I think it is better to keep it in the function. An explanation has been added to the doc string

napari/_vispy/vispy_base_layer.py
sofroniewn: This logic should be robust to transposes because that is captured in the dims ordering before vispy

napari/_vispy/vispy_base_layer.py
sofroniewn: The function will run fine, but 3D pyramids aren't supported yet, so it's not meaningful - I'll add a comment

napari/_vispy/vispy_base_layer.py
sofroniewn: all great points - particularly (e) - thanks in particular there - all addressed

napari/layers/image/image.py
sofroniewn: not needed, removed

napari/layers/image/image.py
sofroniewn: removed here and above

napari/layers/image/image.py
sofroniewn: good point, renamed `thumbnail_source`

napari/layers/image/image.py
sofroniewn: can i fix this in the next PR which will remove pyramid autogeneration and will make this stuff much easier to clean up as won't have this weird case where we were given an array but are using a pyramid. will be much nicer then

sofroniewn:
@jni I have addressed all your concerns - thanks in particular for your suggestions around the `1 data pixel per canvas pixel` I think that has really been cleaned up now and looks really nice for a user.

This is ready for final review!!

napari/layers/image/image.py
sofroniewn: actually this is needed - i got errors when removed

```python-traceback
napari/layers/image/_tests/test_pyramid.py:164:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
napari/layers/image/image.py:228: in __init__
    self._update_dims()
napari/layers/base/base.py:388: in _update_dims
    self.refresh()
napari/layers/base/base.py:605: in refresh
    self._set_view_slice()
napari/layers/image/image.py:470: in _set_view_slice
    downsampled_indices = np.round(downsampled_indices).astype(int)
<__array_function__ internals>:6: in round_
    ???
../../opt/anaconda3/lib/python3.7/site-packages/numpy/core/fromnumeric.py:3599: in round_
    return around(a, decimals=decimals, out=out)
<__array_function__ internals>:6: in around
    ???
../../opt/anaconda3/lib/python3.7/site-packages/numpy/core/fromnumeric.py:3224: in around
    return _wrapfunc(a, 'round', decimals=decimals, out=out)
../../opt/anaconda3/lib/python3.7/site-packages/numpy/core/fromnumeric.py:70: in _wrapfunc
    return _wrapit(obj, method, *args, **kwds)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

obj = array([0.0], dtype=object), method = 'round', args = (), kwds = {'decimals': 0, 'out': None}, wrap = <built-in method __array_wrap__ of numpy.ndarray object at 0x1a373a7940>

    def _wrapit(obj, method, *args, **kwds):
        try:
            wrap = obj.__array_wrap__
        except AttributeError:
            wrap = None
>       result = getattr(asarray(obj), method)(*args, **kwds)
E       TypeError: loop of ufunc does not support argument 0 of type numpy.float64 which has no callable rint method

../../opt/anaconda3/lib/python3.7/site-packages/numpy/core/fromnumeric.py:47: TypeError
```

sofroniewn: i have added it back in

napari/_vispy/vispy_base_layer.py
```diff
@@ -251,43 +220,58 @@ def find_coordinates_of_corners_of_canvas(self):
             tl_raw = [0] * nd
             br_raw = [1] * nd
 
+        # Perform rounding to prevent repeated triggering of requesting
+        # of new pyramid tiles for small camera movements
+        tl_raw = self._rounding * np.floor(np.divide(tl_raw, self._rounding))
+        br_raw = self._rounding * np.ceil(np.divide(br_raw, self._rounding))
```
jni: @sofroniewn I don't understand this logic at all... What does the rounding factor represent? How big does a camera movement have to be before we trigger? Possibly we should revert this and leave it for a future PR. It seems to me like there is a real possibility of glitches where we move the camera a little bit and get black canvas around the edges because we haven't reached this threshold.

napari/_vispy/vispy_base_layer.py
```diff
@@ -320,3 +304,38 @@ def get_max_texture_sizes():
     MAX_TEXTURE_SIZE_3D = 2048
 
     return MAX_TEXTURE_SIZE_2D, MAX_TEXTURE_SIZE_3D
+
+
+def compute_pyramid_level(
```
jni: Love it. =D

napari/layers/image/image.py
jni: Weird! 🤷‍♂️ sorry for the wild goose chase.

napari/_vispy/vispy_base_layer.py
```diff
@@ -190,54 +193,20 @@ def _reset_base(self):
...
-
-    def find_coordinates_of_corners_of_canvas(self):
+    def find_coordinates_of_canvas_corners(self):
         """Find location of the corners of canvas in data coordinates.
 
-        Depends on the current pan and zoom position.
+        This method should only be used during 2D image viewing. The result
+        depends on the current pan and zoom position. Note that the returned
+        coordinates have been clipped to be inside the data to reflect the
+        actual amount of data that would be needed to cover canvas.
```
jni: 👍

jni (approved):
@sofroniewn great!

I have suggested one more change. `self._rounding` is new and it's super confusing. Unless strictly necessary, I would remove it for now and implement in a future PR. Once we get async loading, unnecessary tile loading should not be a huge issue...

jni:
Also, the CI is failing for some reason. 😬

jni:
Here are the failures:

```pytb
_________________________________ test_pyramid _________________________________
viewer_factory = <function viewer_factory.<locals>.actual_factory at 0x7f240f884840>
    def test_pyramid(viewer_factory):
        """Test rendering of pyramid data."""
        view, viewer = viewer_factory(show=True)
        shapes = [(4000, 2000), (2000, 1000), (1000, 500), (500, 250)]
        np.random.seed(0)
        _ = viewer.add_image(data, is_pyramid=True, contrast_limits=[0, 1])
        layer = viewer.layers[0]
        # Set canvas size to target amount
        view.view.canvas.size = (800, 600)
        # Check that current level is top level of pyramid
>       assert layer.data_level == 3
E       assert 1 == 3
E         +1
E         -3
napari/_vispy/_tests/test_vispy_pyramid.py:18: AssertionError
________________________ test_pyramid_screenshot_zoomed ________________________
viewer_factory = <function viewer_factory.<locals>.actual_factory at 0x7f240f8847b8>
    def test_pyramid_screenshot_zoomed(viewer_factory):
        """Test rendering of pyramid data with screenshot after zoom."""
        view, viewer = viewer_factory(show=True)
        shapes = [(4000, 2000), (2000, 1000), (1000, 500), (500, 250)]
        data = [np.ones(s) for s in shapes]
        _ = viewer.add_image(data, is_pyramid=True, contrast_limits=[0, 1])
        # Set canvas size to target amount
        view.view.canvas.size = (800, 600)
        # Set zoom of camera to show highest resolution tile
        view.view.camera.rect = [1000, 1000, 100, 200]
>       view.on_draw(None)
E       AttributeError: 'QtViewer' object has no attribute 'on_draw'
napari/_vispy/_tests/test_vispy_pyramid.py:79: AttributeError
_________________________ test_image_screenshot_zoomed _________________________
viewer_factory = <function viewer_factory.<locals>.actual_factory at 0x7f240fd4b1e0>
    def test_image_screenshot_zoomed(viewer_factory):
        """Test rendering of image data with screenshot after zoom."""
        view, viewer = viewer_factory(show=True)
        data = np.ones((4000, 2000))
        _ = viewer.add_image(data, is_pyramid=True, contrast_limits=[0, 1])
        # Set canvas size to target amount
        view.view.canvas.size = (800, 600)
        # Set zoom of camera to show highest resolution tile
        view.view.camera.rect = [1000, 1000, 100, 200]
>       view.on_draw(None)
E       AttributeError: 'QtViewer' object has no attribute 'on_draw'
napari/_vispy/_tests/test_vispy_pyramid.py:110: AttributeError
```

For two of them you need to change the expected level now that we've gone for "smallest while still being larger than the canvas"

napari/_vispy/vispy_base_layer.py
sofroniewn: I've made `rounding_factor` an optional kwarg arg to this function and added a very detailed doc string around how it is used, it's a tradeoff - but i think a good one for performance. also the use of `floor` and `ceil` instead of pure rounding means that we always request more than we need and the issued described around the edges cannot happen

sofroniewn:
I have fixed the tests and make the `rounding_factor` much more explicit and optional. I will wait for a final look from @jni before merge

napari/_vispy/vispy_base_layer.py
```diff
@@ -224,3 +320,38 @@ def get_max_texture_sizes():
...
 
     return MAX_TEXTURE_SIZE_2D, MAX_TEXTURE_SIZE_3D
+
+
+def compute_pyramid_level(
+    requested_shape, shape_threshold, downsample_factors
+):
+    """Computed desired level of the pyramid given requested field of view.
+
+    The level of the pyramid should be the highest resolution such that
```
sofroniewn: lowest

napari/_vispy/vispy_base_layer.py
```diff
@@ -192,10 +192,106 @@ def _reset_base(self):
...
+            of new pyramid tiles for small camera movements. The rounding
+            factor is in pixel until of the currently viewed tile. Floor and
+            ceil rounding operations are used appropriately to ensure that the
+            rounded requested field always exceeds the needed field to fill the
+            canvas. There is a tradeoff between larger rounding factors causing
+            less frequent updates and tile fetches, but bigger tiles. A factor
+            of around 50 gives good results empically.
+
+        Returns
+        ----------
```
kne42, line 199: could you make this dash as long as returns such that it looks like an underline?

napari/_vispy/vispy_base_layer.py
```diff
@@ -192,10 +192,106 @@ def _reset_base(self):
...
+        rounding_factor : int, optional
+            Use a rounding factor to prevent repeated triggering of requesting
+            of new pyramid tiles for small camera movements. The rounding
+            factor is in pixel until of the currently viewed tile. Floor and
+            ceil rounding operations are used appropriately to ensure that the
+            rounded requested field always exceeds the needed field to fill the
+            canvas. There is a tradeoff between larger rounding factors causing
+            less frequent updates and tile fetches, but bigger tiles. A factor
+            of around 50 gives good results empically.
```
kne42:
```suggestion
            of around 50 gives good results empirically.
```

napari/_vispy/vispy_base_layer.py
```diff
@@ -192,10 +192,106 @@ def _reset_base(self):
...
+        depends on the current pan and zoom position. Note that the returned
+        coordinates have been clipped to be inside the data to reflect the
+        actual amount of data that would be needed to cover canvas.
+
+        Parameters
+        ----------
+        rounding_factor : int, optional
+            Use a rounding factor to prevent repeated triggering of requesting
+            of new pyramid tiles for small camera movements. The rounding
+            factor is in pixel until of the currently viewed tile. Floor and
```
kne42: I don't understand this sentence: "The rounding factor is in pixel until of the currently viewed tile."

kne42 (commented):
great job! just some small nitpicky comments...

napari/_vispy/vispy_base_layer.py
```diff
@@ -192,10 +192,106 @@ def _reset_base(self):
...
+        return corner_pixels, requested_shape
+
     def on_draw(self, event):
         """Called whenever the canvas is drawn.
+
+        This is triggered from vispy whenever new data is sent to the canvas or
+        the camera is moved and is connected in the `QtViewer`.
         """
         self.layer.scale_factor = self.scale_factor
+        if self.layer.is_pyramid:
```
sofroniewn: put corner_pixels on base layer, and move this out of `is_pyramid` for @AhmetCanSolak

napari/_vispy/vispy_base_layer.py
sofroniewn: done, good catch! thanks

napari/_vispy/vispy_base_layer.py
sofroniewn: done, thanks!

napari/_vispy/vispy_base_layer.py
sofroniewn: should be `units`! Great catch, thanks!!

napari/_vispy/vispy_base_layer.py
sofroniewn: so a potential problem now is that @AhmetCanSolak probably doesn't want the values to be clipped and rounded, but the pyramid code does. To solve this I will move the rounding, clipping code elsewhere - will clean some things up

sofroniewn:
This PR now exposes `corner_pixels` on all layers without clipping/ rounding as requested by @AhmetCanSolak and drops the rounding of tiles to prevent repeated updates. As noted by @jni once we have the ability to be async/ drop repeated calls from #845 this will be less of an issue.

This is now ready for a final review

napari/_vispy/_tests/test_vispy_pyramid.py
```diff
@@ -1,4 +1,6 @@
 import numpy as np
+import os
```
jni: LOL this should be `import sys` then just sys. `os.sys` is an alias for `sys`, and I think it's incidental.

sofroniewn: done

sofroniewn:
merging!!!

Issue #734, opened 2019-11-25
Activating matplotlib widget in jupyter lab causes subsequent use of napari to crash the kernel


s-m-e:
**Update from maintainers**: You can see the full thread below but it seems one workaround for napari 0.4.4 is to *first* activate matplotlib widgets and *then* use the `%gui qt` magic for napari.

## 🐛 Bug

Activating the matplotlib widget in Jupyter lab through the `%matplotlib widget` magic causes every subsequent call of `napari.view_image` to crash the kernel.

## To Reproduce

```python
%gui qt
import napari
import numpy as np
data = np.arange(1, 17, dtype = 'int32').reshape(4, 4)
viewer = napari.view_image(data) # does work
%matplotlib widget # root cause of crash
# import matplotlib.pyplot as plt # does not matter
viewer = napari.view_image(data) # triggers crash
```

No traceback or similar is provided (neither in the console nor in the notebook).

## Expected behavior

matplotlib and napari should work side by side.

## Environment

- i7-3740QM & 32 GB RAM
- NVidia K1000M (2GB)
- Linux 4.12 x86_64 (SUSE Leap 15.1)
- Proprietary NVidia driver G04 418.67
- FireFox 68
- Python 3.7.4 (GCC 7)
- **napari 0.2.5**
- PyOpenGL 3.1.0
- zarr 2.3.2
- numcodecs 0.6.3
- numpy 1.17.0
- matplotlib 3.1.1
- ipython 7.7.0
- ipykernel 5.1.1
- jupyter-client 5.3.1
- jupyter-core 4.5.0
- jupyterlab 1.0.4
- jupyterlab-server 1.0.0
- node 12.7.0

## Additional context

None.

s-m-e:
There is a variation to the above bug / example: 

```python
%gui qt
%matplotlib widget
import napari
# import matplotlib.pyplot as plt
import numpy as np
data = np.arange(1, 17, dtype = 'int32').reshape(4, 4)
viewer = napari.view_image(data) # FIRST USE OF NAPARI
```

If I activate `matplotlib widget` **before** I actually use `napari`, I do get a traceback in Jupyter lab ...

```python
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-1-d7b17dc32d77> in <module>
      5 import numpy as np
      6 data = np.arange(1, 17, dtype = 'int32').reshape(4, 4)
----> 7 viewer = napari.view_image(data)

[...]

Issue #687, opened 2019-11-12
Active napari viewer breaks ipywidgets events


jstutters:
## 🐛 Bug

If an instance of napari.Viewer() is present in a jupyter notebook, ipython widget events do not function properly.  The following error is sent to stderr:

`IPKernelApp] WARNING | No such comm: b161f0d81502401fab9931c48a91f974`

## To Reproduce

1. Run the following code in a notebook:

```python
%gui qt
```

Then:

```python
import napari
import ipywidgets as widgets
from IPython.display import display

viewer = napari.Viewer()

def print_it(ev):
    with outputter:
        if ev["name"] == "value":
            print(ev["new"])
            
select = widgets.Select(options=["image 1", "image 2"])
select.observe(print_it)
outputter = widgets.Output()
display(select, outputter)
```

## Expected behavior

Widget events should fire correctly.

## Environment

napari==0.2.5+2.g09a4117
PyQt5==5.13.2
jupyter-core==4.6.1
ipywidgets==7.5.1
ipython==7.9.0

sofroniewn:
Hmm, ok - I suspect it might be something to do with our console code - we connect our napari console to the existing jupyter kernel here https://github.com/napari/napari/blob/32e4533c210d69115fd9f1c9b28861bdd8ca5751/napari/_qt/qt_console.py#L73 and maybe that's causing something weird to happen

I can try reproducing on my machine, and then try seeing if the problem persists if I disable the console. That will at least narrow things down

guiwitz:
I ran several times into this same problem and stumbled across this issue. I tried to do what @sofroniewn suggested and modified the qt_console.py section into:
```python
elif isinstance(shell, ZMQInteractiveShell):
            # if launching from jupyter notebook, connect to the existing
            # kernel
            '''kernel_client = QtKernelClient(
                connection_file=get_connection_file()
            )
            kernel_client.load_connection_file()
            kernel_client.start_channels()

            self.kernel_manager = None
            self.kernel_client = kernel_client
            self.shell = shell
            self.push = self.shell.push'''
            self.shell = None
```
and also commented out this line:
https://github.com/napari/napari/blob/5c2848f298bbdb94ee9706b1cc280cc2c6d61e96/napari/viewer.py#L63
and it seems it did solve the problem. So now I can finally use ipywidgets with napari!

[...]

PR #1055, opened 2020-03-15
make console loading lazy

### LINKED ISSUE #758: slow imports
## 🐛 Bug

@cgohlke has identified some causes of the slowness of our vispy and other imports - see here https://github.com/vispy/vispy/issues/1779 and also this comment from @larsoner https://github.com/vispy/vispy/issues/1779#issuecomment-561305146 related to the use of `pkg_resources` and asking if we can maybe 

> nest the triangle, docutils, and sphinx imports

We should definitely look into doing what we can on our end to improve things too. We can collect ideas here.

sofroniewn:
# Description
This PR will make console loading lazy - i.e. we won't load any console until it is requested by the user. This will have the nice effect of making unintended bugs from our console - such as #687 and #734 less painful to users (though we should maybe also fix them an alternative way), and it will also make load time ~30% faster - see https://github.com/napari/napari/issues/758#issuecomment-573787996.

This was one of the proposed console improvements in #787. There are more proposed in #787, but this seemed like a reasonable one to start with. Note that if you launch from ipython you'll get a button which then pops open a black space when clicked. We could disable that or present a message saying no console present with ipython. Curious what people recommend.

This is ready for review. Maybe @shanaxel42 @ttung @jni and @tlambert03 can take a look.

## Type of change

# References

Closes #758 - see https://github.com/napari/napari/issues/758#issuecomment-573787996

# How has this been tested?

- [x] example: added an explicit test `test_qt_viewer_with_console` inside `napari/_qt/_tests/test_qt_viewer.py`

## Final checklist:

### DIFF
--- a/napari/_qt/_tests/test_qt_viewer.py
+++ b/napari/_qt/_tests/test_qt_viewer.py
@@ -34,7 +34,7 @@ def test_qt_viewer_with_console(viewer_factory):
     assert view.console is None
 
     # Test creation of console
-    view.toggle_console(None)
+    view.toggle_console_visibility(None)
     assert view.console is not None
     assert view.dockConsole.widget == view.console
 
--- a/napari/_qt/_tests/test_qt_viewer.py
+++ b/napari/_qt/_tests/test_qt_viewer.py
@@ -18,6 +18,7 @@ def test_qt_viewer(viewer_factory):
 
     assert viewer.title == 'napari'
     assert view.viewer == viewer
+    assert view.console is None
 
     assert len(viewer.layers) == 0
     assert view.layers.vbox_layout.count() == 2
@@ -25,6 +26,17 @@ def test_qt_viewer(viewer_factory):
     assert viewer.dims.ndim == 2
     assert view.dims.nsliders == viewer.dims.ndim
     assert np.sum(view.dims._displayed_sliders) == 0
+
+
+def test_qt_viewer_with_console(viewer_factory):
+    """Test instantiating console from viewer."""
+    view, viewer = viewer_factory()
+    assert view.console is None
+
+    # Test creation of console
+    view.toggle_console(None)
+    assert view.console is not None
+    assert view.dockConsole.widget == view.console
 
 
 @pytest.mark.parametrize('layer_class, data, ndim', layer_test_data)
--- a/napari/_qt/_tests/test_qt_viewer.py
+++ b/napari/_qt/_tests/test_qt_viewer.py
@@ -18,7 +18,8 @@ def test_qt_viewer(viewer_factory):
 
     assert viewer.title == 'napari'
     assert view.viewer == viewer
-    assert view.console is None
+    # Check no console is present before it is requested
+    assert view._console is None
 
     assert len(viewer.layers) == 0
     assert view.layers.vbox_layout.count() == 2
@@ -31,11 +32,21 @@ def test_qt_viewer(viewer_factory):
 def test_qt_viewer_with_console(viewer_factory):
     """Test instantiating console from viewer."""
     view, viewer = viewer_factory()
-    assert view.console is None
-
-    # Test creation of console
-    view.toggle_console_visibility(None)
+    # Check no console is present before it is requested
+    assert view._console is None
+    # Check console is created when requested
     assert view.console is not None
+    assert view.dockConsole.widget == view.console
+
+
+def test_qt_viewer_toggle_console(viewer_factory):
+    """Test instantiating console from viewer."""
+    view, viewer = viewer_factory()
+    # Check no console is present before it is requested
+    assert view._console is None
+    # Check console has been created when it is supposed to be shown
+    view.toggle_console_visibility(None)
+    assert view._console is not None
     assert view.dockConsole.widget == view.console
 
 
--- a/napari/_qt/qt_console.py
+++ b/napari/_qt/qt_console.py
@@ -141,7 +141,7 @@ class QtConsole(RichJupyterWidget):
         themed_stylesheet : str
             Stylesheet that has already been themed with the current pallete.
         """
-        # self.style_sheet = themed_stylesheet
+        self.style_sheet = themed_stylesheet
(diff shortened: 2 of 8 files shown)


napari/_tests/test_advanced.py
```diff
@@ -155,6 +155,11 @@ def test_update_console(viewer_factory):
     """Test updating the console with local variables."""
     view, viewer = viewer_factory()
 
+    # Create console
+    assert view.console is None
+    view.toggle_console(None)
+    assert view.console is not None
+
     # Check viewer in console
     assert view.console.kernel_client is not None
```
jni: I think this test points to something a bit broken right now: attempting `view.console.whatever` right now will do bad things if it hasn't been instantiated yet. How about making `.console` a property and have it instantiate then if not already created?

napari/_qt/qt_console.py
```diff
@@ -128,6 +130,22 @@ def __init__(self, user_variables=None):
...
+
+        Parameters
+        ----------
+        palette : dict of str: str
+            Color palette with which to style the viewer.
+            Property of napari.components.viewer_model.ViewerModel.
+        themed_stylesheet : str
+            Stylesheet that has already been themed with the current pallete.
+        """
+        self.style_sheet = themed_stylesheet
```
tlambert03, line 144: i'm not certain, but this may be unnecessary?  `QtWidgets` inherit their styles from their parents, so it's possible that if we just make sure parent-child relationships are all set, you don't need to do this.  (note, the empty `super().__init__()` call in the `__init__` function would need to be fixed with a parent).  Setting parents is good practice anyway for the purpose of cleanup, so it might be a good time to review that for the QtConsole.

napari/_qt/qt_viewer.py
```diff
@@ -117,7 +114,7 @@ def __init__(self, viewer):
         )
         self.dockConsole = QtViewerDockWidget(
             self,
-            self.console,
+            QWidget(),
```
tlambert03, line 117: we could also modify `QtViewerDockWidget` to accept `None` for widget

tlambert03 (commented):
this is nice!  Given how often the console is responsible for tricky bugs, I think this is a good direction.  I agree with @jni about a `.console` property that creates itself if it doesn't exist.  
Also, just want to mention that the console was one of the biggest problems when I looked into cleanup in #1014.  So it's possible that, if you do end up adding new parent/child relations as I mentioned in one comment, it might have implications for they way shutdown is handled in `QtViewer.closeEvent`

napari/_qt/qt_console.py
sofroniewn: @tlambert03 how do you recommend I set the parent as we don't pass the QtViewer to the console? Maybe we should change that, but then I think that should be its own PR.

napari/_qt/qt_viewer.py
sofroniewn: That sounds like a fine choice to me, but maybe something for a follow up PR if you could do it.

napari/_tests/test_advanced.py
sofroniewn: done, though please check my implementation and its consequences in `toggle_console_visibility`

sofroniewn:
I have made `console` a property as requested by @jni and @tlambert03, but I'm not sure it might have lead to some undesirable consequences in `toggle_console_visibility`, i.e. the line `_ = self.console` that is needed to force instantiation. Maybe this is ok.

I've also sort of punted on the two comments from @tlambert03, with guidance I could try and address now, or they could wait for follow up PRs.

This is ready for re-review

napari/_qt/_tests/test_qt_viewer.py
```diff
@@ -27,6 +29,27 @@ def test_qt_viewer(viewer_factory):
...
 
 
+def test_qt_viewer_with_console(viewer_factory):
+    """Test instantiating console from viewer."""
+    view, viewer = viewer_factory()
+    # Check no console is present before it is requested
+    assert view._console is None
+    # Check console is created when requested
+    assert view.console is not None
+    assert view.dockConsole.widget == view.console
```
jni, line 39: This is a very pretty test now. =)

napari/_qt/qt_viewer.py
jni: I agree that this should be a followup.

napari/_qt/qt_viewer.py
```diff
@@ -195,6 +186,22 @@ def __init__(self, viewer):
...
 
+    @property
+    def console(self):
+        """QtConsole: iPython console terminal integrated into the napari GUI.
+        """
+        if self._console is None:
+            from .qt_console import QtConsole
+
+            self.console = QtConsole({'viewer': self.viewer})
+        return self._console
```
jni, line 197: This is so nice and clean now!

napari/_qt/qt_viewer.py
```diff
@@ -195,6 +186,22 @@ def __init__(self, viewer):
...
         self.setAcceptDrops(True)
 
+    @property
+    def console(self):
+        """QtConsole: iPython console terminal integrated into the napari GUI.
+        """
+        if self._console is None:
+            from .qt_console import QtConsole
+
+            self.console = QtConsole({'viewer': self.viewer})
```
jni, line 196: I just want to flag that now *might* be a good time to update this behaviour, but that can wait for a follow-up PR, since this is not making things worse.

jni (approved):
Ah, this is so great now @sofroniewn! I do think @tlambert03's styling issues can wait for a new PR — I'm so excited to have this in!

tlambert03 (approved):

napari/_qt/qt_viewer.py
sofroniewn: Sounds good

sofroniewn:
Ok, merging now, excited to get this in too!

napari/_qt/qt_viewer.py
sofroniewn: Made a note in #787 which is still open

PR #1056, opened 2020-03-16
[WIP] Move contrast and gamma calculations to shader

tlambert03:
# Description
The vispy `VolumeVisual.set_data` method [performs contrast limits scaling on the CPU](https://github.com/vispy/vispy/blob/3ac28ff0daf8836b44c6eba8d3c98541e37c5702/vispy/visuals/volume.py#L487-L493) (normalizing the data to 0-1 based on clim sliders), prior to transfer to texture memory on the GPU.  This creates a lag that is particularly noticeable in 3d volume view (on my computer, @VolkerH's modestly-sized 136x592x790 drosophila dataset [here](https://www.dropbox.com/s/93ykwmonofq7171/droso_ovarioles.tif?dl=0) takes somewhere around 150ms to render each time the contrast is changed).

This PR puts the contrast limits and gamma logic in the shader, improving volume render time somewhere between **2-5 fold** when changing contrast, depending on the volume size.
benchmarks:
<table>
<thead>
<tr>
<th></th>
<th colspan="2">Change Contrast</th>
<th colspan="2">Change Contrast Range</th>
<th colspan="2">Change Gamma</th>
</tr>
<tr>
<th>XY size<br>(Z is n/2)</th>
<th>master</th>
<th>this PR</th>
<th>master</th>
<th>this PR</th>
<th>master</th>
<th>this PR</th>
</tr>
</thead>
<tbody>
<tr>
<td>16</td>
<td>3.06ms</td>
<td>2.79ms</td>
<td>3.36ms</td>
<td>3.13ms</td>
<td>6.69ms</td>
<td>2.79ms</td>
</tr>
<tr>
<td>32</td>
<td>3.60ms</td>
<td>3.29ms</td>
<td>3.51ms</td>
<td>3.51ms</td>
<td>6.94ms</td>
<td>3.29ms</td>
</tr>
<tr>
<td>64</td>
<td>5.43ms</td>
<td>2.74ms</td>
<td>5.88ms</td>
<td>4.71ms</td>
<td>5.83ms</td>
<td>2.74ms</td>
</tr>
<tr>
<td>128</td>
<td>16.7ms</td>
<td>6.26ms</td>
<td>26.6ms</td>
<td>16.6ms</td>
<td>9.30ms</td>
<td>6.26ms</td>
</tr>
<tr>
<td>256</td>
<td>118ms</td>
<td>34.8ms</td>
<td>186ms</td>
<td>116ms</td>
<td>38.6ms</td>
<td>34.8ms</td>
</tr>
<tr>
<td>512</td>
<td>1.21s</td>
<td>256ms</td>
<td>2.04s</td>
<td>1.20s</td>
<td>263ms</td>
<td>256ms</td>
</tr>
<tr>
<td>1024</td>
<td>9.21s</td>
<td>2.11s</td>
<td>16.5s</td>
<td>9.80s</td>
<td>2.14s</td>
<td>2.11s</td>
</tr>
</tbody>
</table>

In order to keep from overriding more of the vispy `Volume` code (in this case, the `set_data` method), I'm hijacking the semantics of their "clim" attribute for what we internally think of as `contrast_limits_range`.  The reasoning here is that the `glTexture3D` object expects data normalized from 0-1, so it seems at least some normalization needs to happen on the CPU.  However, that normalization need only restrict the _maximum_ contrast range, not the _current_ contrast range, so it really only needs to be renormalized (on the CPU) whenever `contrast_limits_range` changes... whereas regular contrast limit changes can all be done on the GPU.

I'll try to update with inline comments soon to clarify.

## Type of change

# How has this been tested?

## Final checklist:


sofroniewn:
@tlambert03 this is great. Using the contrast limits range to do an initial normalization seems like a good idea. It might be nice to think if this can be be made a vispy PR eventually. This file is starting to diverge quite a bit from theirs now (it will get better after https://github.com/vispy/vispy/pull/1802 merges, but we still need to make a PR for the attenuated_MIP mode).

Also it looks like you havn't added gamma and display lim calculations to all the rendering modes yet - i know this is WIP, but just pointing it out in case.

tlambert03:
Created a new PR at vispy with these changes, this either needs to be a breaking change, or the property names will end up "not really matching" their symantics... so we'll see what they say.  I definitely would like us to be able to use this internally though, as it very much improves routine usage of napari

sofroniewn:
This has been superseded by #1456 - great work on this @tlambert03!!

Issue #704, opened 2019-11-18
Don't depend on pyside


hmaarrfk:
## 🚀 Feature
Depending on Pyside2 makes things easy to install, but if you already depend on qtpy, then maybe let the users decide which backend is best for them?

Just my 2 cents.

tlambert03:
I kinda agree here.  I see that this creates a problem: how do you make an easy "one line install" for those who know nothing about Qt, while also preventing duplicate Qt backends for those who already depend on pyqt for other things.  

I _do_ think that it's really important for a beginner to just `pip install napari` and be off and running – even something like `pip install napari[pyside]` is undesirable – but when I went about trying to include support for napari in some of my other qt packages, I ran up against either needing to have both pyqt5 & pyside2 in the environment, or changing all my stuff over to pyside2.  @jni, do you have thoughts on how to approach that sort of thing?

AhmetCanSolak:
If you guys want to use pyqt5 or another backend, you should first import the backend you want to use, then import *napari*. Since napari uses qtpy if before importing napari you have another backend imported pyside2 will not be able to import successfully.

About the reason of pyside2 choice, mainly licensing AFAIK. Correct me if i'm wrong @sofroniewn . As soon as we are not depending non-LGPL parts of pyqt5 it would be okay to switch backend and i love to do that too, personally i like pyqt5 better than pyside2.

tlambert03:
I might be wrong... but I think the issue was not so much about _how_ to get qtpy (and therefore napari) to use the desired backend.  That much is straightforward.  and it's also not so much about using `pyside2` as the "go-to" default backend... that's also fine, particularly if it makes licensing easier.
I think it is an interesting discussion though as to whether installing napari will _force_ the user to have pyside2 installed in their environment (even if they then take steps to not use it).

hmaarrfk:
> I think it is an interesting discussion though as to whether installing napari will force the user to have pyside2

^this

AhmetCanSolak:
Good point @tlambert03 . But either way we need a backend by default. What could be an alternative? Do you have any suggestions?

tlambert03:
nope!  😄   that's why in my first comment I said it creates a problem and punted to @jni :)
[...]

PR #1059, opened 2020-03-18
check if PyQt5 is installed and then not install PySide2

Czaki:
# Description
If someone have installed PyQt5 then installing PySide2 next to may create problem with chosing Qt implementation. 

## Type of change
Filter Pyside2 from requirements if PyQt5 is installed 

# References

# How has this been tested?

- [x] run `python setup.py sdist` then on new interpeter (ex. virtualenv) `pip install -U pip; pip install PyQt5; pip install dist/napari*tar.gz` 
I do not know why, but when try to install from repository it still install PySide2 

## Final checklist:

Other option is to move PySide2 requirements from requirements/default.txt to setup.py and add it under condition.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -69,7 +69,7 @@ linux_task:
   install_script:
     - python -m pip install --retries 3 --upgrade pip
     - pip install --retries 3 -r requirements.txt
-    - python setup.py develop
+    - python -m pip install -e .
 
   check_script:
     - conda list
@@ -123,7 +123,7 @@ mac_task:
   install_script:
     - python -m pip install --retries 3 --upgrade pip
     - pip install --retries 3 -r requirements.txt
-    - python setup.py develop
+    - python -m pip install -e .
 
   check_script:
     - conda list
@@ -187,7 +187,7 @@ win_task:
   install_script:
     - python -m pip install --retries 3 --upgrade pip
     - pip install --retries 3 -r requirements.txt
-    - python setup.py develop
+    - python -m pip install -e .
 
   check_script:
     - conda list
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -220,6 +220,7 @@ codecov_task:
   install_script: |
     python -m pip install --retries 3 --upgrade pip
     pip install -r requirements.txt
+    pip install -e .
     pip install pytest-cov codecov
 
   coverage_script: |
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -15,4 +15,4 @@ PyOpenGL>=3.1.0
 wrapt>=1.11.1
 numpydoc>=0.9.2
 pluggy>=0.13.0
-# Qt version is managed form setup.py to allow Pyside2 installation omit whem PyQt5 is installed
+# Qt version is managed from setup.py to install PySide2 when no Qt libraries are found
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -12,7 +12,7 @@ vispy>=0.6.4
 IPython>=7.7.0
 backcall>=0.1.0
 PyOpenGL>=3.1.0
-PySide2>=5.12.3
 wrapt>=1.11.1
 numpydoc>=0.9.2
 pluggy>=0.13.0
+# Qt version is managed form setup.py to allow Pyside2 installation omit whem PyQt5 is installed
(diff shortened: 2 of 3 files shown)


kne42 (changes requested):
Thanks for the contribution @Czaki, this is super cool! Didn't realize we can change our requirements at installation time :)

Two things I'd like to request due to Qt versioning:
- if PyQt5 is found, it should be updated to the same minimum version as we have for PySide2
- if both PyQt5 and PySide2 are found, they both should be updated to the same minimum version as we have for PySide2

kne42:
> I do not know why, but when try to install from repository it still install PySide2

Did you run `pip install -r requirements.txt` first?

Czaki:
@kne42 Code will be cleaner if I can remove PySide from `requirements/default.txt`but in these case dependency will be stored in two places. 

What You prefer?

Czaki:
>>    I do not know why, but when try to install from repository it still install PySide2

> Did you run pip install -r requirements.txt first?

No,

kne42:
@Czaki Yeah, I think you should go ahead and remove it from `requirements.default.txt`, but please leave a comment in its place telling people that we will dynamically install the right Qt library in `setup.py`.

.cirrus.yml
```diff
@@ -69,7 +69,7 @@ linux_task:
   install_script:
     - python -m pip install --retries 3 --upgrade pip
     - pip install --retries 3 -r requirements.txt
-    - python setup.py develop
+    - python -m pip install -e .
```
kne42, line 72: hmm, I don't think we should make this editable

requirements/default.txt
```diff
@@ -12,7 +12,7 @@ vispy>=0.6.4
 IPython>=7.7.0
 backcall>=0.1.0
 PyOpenGL>=3.1.0
-PySide2>=5.12.3
 wrapt>=1.11.1
 numpydoc>=0.9.2
 pluggy>=0.13.0
+# Qt version is managed form setup.py to allow Pyside2 installation omit whem PyQt5 is installed
```
kne42:
```suggestion
# Qt version is managed from setup.py to install PySide2 when no Qt libraries are found
```

kne42 (approved):

.cirrus.yml
Czaki: I do not understand. `-e` means that package will use project structure as source root. So test will use same files which is installed as package. 

Maybe best solution is to use [tox](https://tox.readthedocs.io/en/latest/). Did You talk about this?

kne42: ah ok nevermind then

jni: I like this change as is.

setup.py
```diff
@@ -54,6 +54,22 @@
     )
     sys.exit(1)
 
+QT_MIN_VERSION = "5.12.3"
+
+try:
+    import PyQt5  # noqa: F401
+
```
jni, line 61:
```suggestion
```

setup.py
```diff
@@ -54,6 +54,22 @@
...
+    pyqt = True
+except ImportError:
+    pyqt = False
+
+try:
+    import PySide2  # noqa: F401
+
```
jni, line 68:
```suggestion
```

jni (approved):
Love it! I'm quite unhappy that we can't implement this either/or with declarative syntax, but this is a great workaround. Thanks @Czaki!

setup.py
Czaki: This line comes from black.

setup.py
Czaki: This line comes from black.

sofroniewn (approved):
Looks good to me! Thanks - I think this can close #704 too

Czaki:
@jni I do not know solution which is declarative and do not use meta package (like suggested in #704).

sofroniewn:
@Czaki this looks great, thanks for your contribution!! Merging now.

We'll update our installation tutorials appropriately (@GenevieveBuckley) :-)

GenevieveBuckley:
I've made a PR for the installation guide in the docs here: https://github.com/napari/tutorials/pull/61

ttung:
FYI this strategy may be incompatible with binary wheels.

Czaki:
@ttung Yes. For wheel napari should be changed to metapackage which depend on something like `napari-base[pyside]`

PR #1030, opened 2020-03-06
Plugin docs (for plugin developers)

tlambert03:
# Description
Just wanted to get on top of writing some plugin docs.  I believe @jni was planning on writing the docs for napari contributors and developers that would explain how the napari plugin architecture is implemented.  I created a stub file here for him to do that.  The file that I created is a "guide on creating plugins" aimed at potential plugin developers.

I'll leave this as WIP until at least #937 merges, in case anything changes about the hookspecs.

## Type of change

- [x] docs

PR #1061, opened 2020-03-18
remove scikit-image dependency

Czaki:
# Description
This PR remove scikit-image dependency. This remove also matplotlib dependency. Both are quite big so increase size of bundled (eg. with pyinstaller) app.

## Type of change

- [x] Replace scikit-image components with custom one. 
- [x] if scikit-image is installed then it is still used for reading data, so one can use scikit-image plug-in system to control it. (environment variable "NO_SKIMAGE" disable it. I'm not sure if it should be documented.)
- [x] side effect. Add support of lsm files. 
- [x] on linux need pip in version at least 19.0 to install manylinux2010 wheels. Compiling imagecodecs manually is time consuming. 
- [x] add check if all project dependecies are listed in requirements

# How has this been tested?

- [x] introduced new test
- [x] all previous tests pass with my change

## Final checklist:

### DIFF
--- a/napari/layers/_tests/test_base.py
+++ b/napari/layers/_tests/test_base.py
@@ -1,0 +1,39 @@
+import numpy as np
+import pytest
+from skimage import img_as_ubyte
+
+from ..base.base import convert_to_uint8
+
+
+@pytest.mark.parametrize("dtype", [np.uint8, np.uint16, np.uint32, np.uint64])
+def test_uint(dtype):
+    data = np.arange(50, dtype=dtype)
+    data_scaled = data * 256 ** (data.dtype.itemsize - 1)
+    assert np.all(data == convert_to_uint8(data_scaled))
+    assert np.all(img_as_ubyte(data) == convert_to_uint8(data))
+    assert np.all(img_as_ubyte(data_scaled) == convert_to_uint8(data_scaled))
+
+
+@pytest.mark.parametrize("dtype", [np.int8, np.int16, np.int32, np.int64])
+def test_int(dtype):
+    data = np.arange(50, dtype=dtype)
+    data_scaled = data * 256 ** (data.dtype.itemsize - 1)
+    assert np.all(img_as_ubyte(data) == convert_to_uint8(data))
+    assert np.all(2 * data == convert_to_uint8(data_scaled))
+    assert np.all(img_as_ubyte(data_scaled) == convert_to_uint8(data_scaled))
+    assert np.all(img_as_ubyte(data - 10) == convert_to_uint8(data - 10))
+    assert np.all(
+        img_as_ubyte(data_scaled - 10) == convert_to_uint8(data_scaled - 10)
+    )
+
+
+@pytest.mark.parametrize("dtype", [np.float64, np.float32, np.float])
+def test_float(dtype):
+    data = np.linspace(0, 0.5, 128, dtype=dtype, endpoint=False)
+    res = np.arange(128, dtype=np.uint8)
+    assert np.all(convert_to_uint8(data) == res)
+    data = np.linspace(0, 1, 256, dtype=dtype)
+    res = np.arange(256, dtype=np.uint8)
+    assert np.all(convert_to_uint8(data) == res)
+    assert np.all(img_as_ubyte(data) == convert_to_uint8(data))
+    assert np.all(img_as_ubyte(data - 0.5) == convert_to_uint8(data - 0.5))
--- a/napari/layers/_tests/test_base.py
+++ b/napari/layers/_tests/test_base.py
@@ -9,6 +9,7 @@ from ..base.base import convert_to_uint8
 def test_uint(dtype):
     data = np.arange(50, dtype=dtype)
     data_scaled = data * 256 ** (data.dtype.itemsize - 1)
+    assert convert_to_uint8(data_scaled).dtype == np.uint8
     assert np.all(data == convert_to_uint8(data_scaled))
     assert np.all(img_as_ubyte(data) == convert_to_uint8(data))
     assert np.all(img_as_ubyte(data_scaled) == convert_to_uint8(data_scaled))
@@ -18,6 +19,8 @@ def test_uint(dtype):
 def test_int(dtype):
     data = np.arange(50, dtype=dtype)
     data_scaled = data * 256 ** (data.dtype.itemsize - 1)
+    assert convert_to_uint8(data).dtype == np.uint8
+    assert convert_to_uint8(data_scaled).dtype == np.uint8
     assert np.all(img_as_ubyte(data) == convert_to_uint8(data))
     assert np.all(2 * data == convert_to_uint8(data_scaled))
     assert np.all(img_as_ubyte(data_scaled) == convert_to_uint8(data_scaled))
@@ -31,9 +34,18 @@ def test_int(dtype):
 def test_float(dtype):
     data = np.linspace(0, 0.5, 128, dtype=dtype, endpoint=False)
     res = np.arange(128, dtype=np.uint8)
+    assert convert_to_uint8(data).dtype == np.uint8
     assert np.all(convert_to_uint8(data) == res)
     data = np.linspace(0, 1, 256, dtype=dtype)
     res = np.arange(256, dtype=np.uint8)
     assert np.all(convert_to_uint8(data) == res)
     assert np.all(img_as_ubyte(data) == convert_to_uint8(data))
     assert np.all(img_as_ubyte(data - 0.5) == convert_to_uint8(data - 0.5))
+
+
+def test_bool():
+    data = np.zeros((10, 10), dtype=np.bool)
+    data[2:-2, 2:-2] = 1
+    converted = convert_to_uint8(data)
+    assert converted.dtype == np.uint8
+    assert np.all(img_as_ubyte(data) == converted)
--- a/napari/layers/_tests/test_base.py
+++ b/napari/layers/_tests/test_base.py
@@ -1,6 +1,6 @@
 import numpy as np
 import pytest
-from skimage import img_as_ubyte
+from skimage.util import img_as_ubyte
 
 from ..base.base import convert_to_uint8
 
--- a/napari/layers/base/base.py
+++ b/napari/layers/base/base.py
@@ -5,7 +5,6 @@ from abc import ABC, abstractmethod
 from contextlib import contextmanager
 from xml.etree.ElementTree import Element, tostring
 import numpy as np
-from skimage import img_as_ubyte
 from ._base_constants import Blending
 
 from ...components import Dims
@@ -452,7 +451,7 @@ class Layer(KeymapProvider, ABC):
         if thumbnail.dtype != np.uint8:
             with warnings.catch_warnings():
                 warnings.simplefilter("ignore")
-                thumbnail = img_as_ubyte(thumbnail)
+                thumbnail = convert_to_uint8(thumbnail)
 
         padding_needed = np.subtract(self._thumbnail_shape, thumbnail.shape)
         pad_amounts = [(p // 2, (p + 1) // 2) for p in padding_needed]
@@ -737,3 +736,33 @@ class Layer(KeymapProvider, ABC):
         """Called whenever mouse released in canvas.
         """
         return
+
+
+def convert_to_uint8(data: np.ndarray):
+    out_dtype = np.dtype(np.uint8)
+    out_max = np.iinfo(out_dtype).max
+    if data.dtype == out_dtype:
+        return data
+    in_kind = data.dtype.kind
+    if in_kind == "f":
+        image_out = np.multiply(data, out_max, dtype=data.dtype)
+        np.rint(image_out, out=image_out)
+        np.clip(image_out, 0, out_max, out=image_out)
+        return image_out.astype(out_dtype)
+
+    if in_kind in "ui":
+        if in_kind == "u":
+            if data.max() < out_max:
+                return data.astype(out_dtype)
+            return np.right_shift(data, (data.dtype.itemsize - 1) * 8).astype(
+                out_dtype
+            )
+        else:
+            np.maximum(data, 0, out=data, dtype=data.dtype)
+            if data.dtype == np.int8:
+                return (data * 2).astype(np.uint8)
+            if data.max() < out_max:
+                return data.astype(out_dtype)
+            return np.right_shift(
+                data, (data.dtype.itemsize - 1) * 8 - 1
+            ).astype(out_dtype)
--- a/napari/layers/base/base.py
+++ b/napari/layers/base/base.py
@@ -744,6 +744,8 @@ def convert_to_uint8(data: np.ndarray):
     if data.dtype == out_dtype:
         return data
     in_kind = data.dtype.kind
+    if in_kind == "b":
+        return data.astype(out_dtype) * 255
     if in_kind == "f":
         image_out = np.multiply(data, out_max, dtype=data.dtype)
         np.rint(image_out, out=image_out)
--- a/napari/layers/base/base.py
+++ b/napari/layers/base/base.py
@@ -10,6 +10,7 @@ from ._base_constants import Blending
 from ...components import Dims
 from ...utils.event import EmitterGroup, Event
 from ...utils.key_bindings import KeymapProvider
+from ..utils.layer_utils import convert_to_uint8
 from ...utils.misc import ROOT_DIR
 from ...utils.naming import magic_name
 from ...utils.status_messages import status_format, format_float
@@ -736,35 +737,3 @@ class Layer(KeymapProvider, ABC):
         """Called whenever mouse released in canvas.
         """
         return
-
-
-def convert_to_uint8(data: np.ndarray):
-    out_dtype = np.dtype(np.uint8)
-    out_max = np.iinfo(out_dtype).max
-    if data.dtype == out_dtype:
-        return data
-    in_kind = data.dtype.kind
-    if in_kind == "b":
-        return data.astype(out_dtype) * 255
-    if in_kind == "f":
-        image_out = np.multiply(data, out_max, dtype=data.dtype)
-        np.rint(image_out, out=image_out)
-        np.clip(image_out, 0, out_max, out=image_out)
-        return image_out.astype(out_dtype)
-
-    if in_kind in "ui":
-        if in_kind == "u":
-            if data.max() < out_max:
-                return data.astype(out_dtype)
-            return np.right_shift(data, (data.dtype.itemsize - 1) * 8).astype(
-                out_dtype
-            )
-        else:
-            np.maximum(data, 0, out=data, dtype=data.dtype)
-            if data.dtype == np.int8:
-                return (data * 2).astype(np.uint8)
-            if data.max() < out_max:
-                return data.astype(out_dtype)
-            return np.right_shift(
-                data, (data.dtype.itemsize - 1) * 8 - 1
-            ).astype(out_dtype)
--- a/napari/layers/image/image.py
+++ b/napari/layers/image/image.py
@@ -4,7 +4,6 @@ from base64 import b64encode
 from xml.etree.ElementTree import Element
 
 import numpy as np
-from imageio import imwrite
 from scipy import ndimage as ndi
 
 from ...utils.colormaps import AVAILABLE_COLORMAPS
@@ -623,6 +622,8 @@ class Image(IntensityVisualizationMixin, Layer):
             List of a single xml element specifying the currently viewed image
             as a png according to the svg specification.
         """
+        from imageio import imwrite
+
         if self.dims.ndisplay == 3:
             image = np.max(self._data_thumbnail, axis=0)
         else:
--- a/napari/layers/image/image.py
+++ b/napari/layers/image/image.py
@@ -622,6 +622,7 @@ class Image(IntensityVisualizationMixin, Layer):
             List of a single xml element specifying the currently viewed image
             as a png according to the svg specification.
         """
+        # we delay this import to minimize import time at launch
         from imageio import imwrite
 
         if self.dims.ndisplay == 3:
--- a/napari/utils/io.py
+++ b/napari/utils/io.py
@@ -1,14 +1,77 @@
 import os
+import re
 
 from glob import glob
 from pathlib import Path
 
 import numpy as np
-from skimage import io
-from skimage.io.collection import alphanumeric_key
+import tifffile
+import imageio
 
 from dask import delayed
 from dask import array as da
+
+
+def imsave(filename: str, data: np.ndarray):
+    """custom imaplementation of imread to avoid skimage dependecy"""
+    ext = os.path.splitext(filename)[1]
+    if ext in [".tif", "tiff"]:
+        tifffile.imsave(filename, data)
+    else:
+        imageio.imsave(filename, data)
+
+
+def imread(filename: str):
+    """custom imaplementation of imread to avoid skimage dependecy"""
+    ext = os.path.splitext(filename)[1]
+    if ext in [".tif", "tiff", ".lsm"]:
+        image = tifffile.imread(filename)
+    else:
+        image = imageio.imread(filename)
+
+    if not hasattr(image, 'ndim'):
+        return image
+
+    if image.ndim > 2:
+        if image.shape[-1] not in (3, 4) and image.shape[-3] in (3, 4):
+            image = np.swapaxes(image, -1, -3)
+            image = np.swapaxes(image, -2, -3)
+    return image
+
+
+if not os.environ.get("NO_SKIMAGE", False):
+    try:
+        from skimage.io import imread as _imread
+    except ImportError:
+        _imread = imread
+else:
+    _imread = imread
+
+
+def _alphanumeric_key(s):
+    """Convert string to list of strings and ints that gives intuitive sorting.
+
+    Parameters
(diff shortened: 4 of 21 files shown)


cgohlke:
On my system, `import skimage.io` amounts to ~17% of napari's startup time. See also https://github.com/napari/napari/issues/758#issuecomment-561493551

Unfortunately `import imageio` is also expensive. Maybe importing `imageio` and `tifffile` can be delayed, i.e. moved to the `imsave` and `imread` functions?

Czaki:
@cgohlke How do you measure loading time?

Czaki:
Tifffile impact on importing time is minimal.

Czaki:
Reffering to https://forum.image.sc/t/using-napari-as-imagepys-canvas/32579/13

imageio is backed used by scikit-image.

scikit-image also embed some old version of tifffile (0.15 if I good remember)  and use more modern if available (installed).

In last release of tifffile @cgohlke decide to stop using `imagecodecs-lite` and switch to imagecodecs which is generally better idea but I (I'm the person who create pipeline to build wheels for linux and macos) do not found way to build manylinux1 wheels for it. It is not a big problem because most of currently working system can use it.

Only problem is because python 3.6 is shipped by default with pip in version 9.0. And manylinux2010 is recognized as proper platform tag from pip 19.0.

Some workaround of pip 19.0 need for this problem is to move `tifffile` dependency resolving to `setup.py` file like in #1059) and if is not installed and system is linux then pin to 2019.7.26 version (last which works with  `imagecodecs-lite`)

Other is to check pip version in setup.py and fail installation if it is to low (allows to clearly inform user how to solve the problem).

Maybe @cgohlke has some other suggestion?

Czaki:
anyone knows why this happens time to time?
```
========== 694 passed, 6 skipped, 3535 warnings in 122.49s (0:02:02) ==========
Windows fatal exception: access violation
Current thread 0x00000acc (most recent call first):
  File "c:\users\containeradministrator\anaconda\lib\site-packages\IPython\core\displayhook.py", line 311 in flush
  File "c:\users\containeradministrator\anaconda\lib\site-packages\IPython\core\interactiveshell.py", line 1430 in reset
  File "c:\users\containeradministrator\anaconda\lib\site-packages\IPython\core\interactiveshell.py", line 3696 in atexit_operations
C:\Users\ContainerAdministrator\AppData\Local\Temp\cirrus-ci-build>if -1073741819 NEQ 0 exit /b -1073741819 
Exit status: 3221225477
```

sofroniewn:
Hi @Czaki thanks for making this PR and apologies for taking so long to respond. I think in principle this is something that we might be open too. There are certainly some details in the PR (where new code is put etc.) that we might have request changes on, but we can work through these. I think @jni will give a thorough review when he gets the chance.

To @cgohlke suggestion:
> Unfortunately import imageio is also expensive. Maybe importing imageio and tifffile can be delayed, i.e. moved to the imsave and imread functions?

I love this idea and it might probably make sense to do that as part of this PR too or in a follow up. Thanks again @cgohlke, you've really shaved a lot of time off of our startup with your suggestions. Once  #1055 goes in, its going to be even faster :-)

Czaki:
> I love this idea and it might probably make sense to do that as part of this PR too or in a follow up. Thanks again @cgohlke, you've really shaved a lot of time off of our startup with your suggestions. Once #1055 goes in, its going to be even faster :-)

Ok. I delay also tifffile import

napari/layers/base/base.py
```diff
@@ -732,3 +731,35 @@ def on_mouse_release(self, event):
         """Called whenever mouse released in canvas.
         """
         return
+
+
+def convert_to_uint8(data: np.ndarray):
```
jni: I would put this somewhere in utils and it needs a docstring.

napari/layers/_tests/test_base.py
```diff
@@ -0,0 +1,51 @@
+import numpy as np
+import pytest
+from skimage import img_as_ubyte
```
jni:
```suggestion
from skimage.util import img_as_ubyte
```

napari/layers/image/image.py
```diff
@@ -603,6 +602,8 @@ def to_xml_list(self):
             List of a single xml element specifying the currently viewed image
             as a png according to the svg specification.
         """
+        from imageio import imwrite
```
jni, line 626:
```suggestion
        # we delay this import to minimize import time at launch
        from imageio import imwrite
```

napari/utils/io.py
```diff
@@ -1,16 +1,85 @@
...
 import numpy as np
-from skimage import io
-from skimage.io.collection import alphanumeric_key
 
 from dask import delayed
 from dask import array as da
 
 
+def imsave(filename: str, data: np.ndarray):
+    """custom imaplementation of imread to avoid skimage dependecy"""
```
jni:
```suggestion
    """Custom implementation of imsave to avoid skimage dependency.

    Parameters
    ----------
    filename : string
        The path to write the file to.
    data : np.ndarray
        The image data.
    """
```

napari/utils/io.py
```diff
@@ -1,16 +1,85 @@
...
+    ext = os.path.splitext(filename)[1]
+    if ext in [".tif", "tiff"]:
```
jni:
```suggestion
    if ext in [".tif", ".tiff"]:
```

napari/utils/io.py
```diff
@@ -1,16 +1,85 @@
...
+        import tifffile
+
+        tifffile.imsave(filename, data)
+    else:
+        import imageio
+
+        imageio.imsave(filename, data)
+
+
+def imread(filename: str):
```
jni:
```suggestion
def imread(filename: str) -> np.ndarray:
```

napari/utils/io.py
```diff
@@ -1,16 +1,85 @@
...
+    """custom imaplementation of imread to avoid skimage dependecy"""
```
jni:
```suggestion
    """Custom implementation of imread to avoid skimage dependency.

    Parameters
    ----------
    filename : string
        The path from which to read the image.

    Returns
    -------
    data : np.ndarray
        The image data.
    """
```

napari/utils/io.py
```diff
@@ -1,16 +1,85 @@
...
+    ext = os.path.splitext(filename)[1]
+    if ext in [".tif", "tiff", ".lsm"]:
```
jni:
```suggestion
    if ext in [".tif", ".tiff", ".lsm"]:
```

napari/utils/io.py
```diff
@@ -1,16 +1,85 @@
...
+        import tifffile
+
+        image = tifffile.imread(filename)
+    else:
+        import imageio
+
+        image = imageio.imread(filename)
+
+    if not hasattr(image, 'ndim'):
+        return image
```
jni: What is this clause for?

napari/utils/io.py
```diff
@@ -1,16 +1,85 @@
...
+
+    if image.ndim > 2:
+        if image.shape[-1] not in (3, 4) and image.shape[-3] in (3, 4):
+            image = np.swapaxes(image, -1, -3)
+            image = np.swapaxes(image, -2, -3)
```
jni: Honestly, I would simply remove this logic. It is something archaic that we do in skimage and we will probably remove it in 1.0.

napari/utils/io.py
```diff
@@ -1,16 +1,85 @@
...
+    return image
+
+
+if not os.environ.get("NO_SKIMAGE", False):
+    try:
+        from skimage.io import imread as _imread
+    except ImportError:
+        _imread = imread
+else:
+    _imread = imread
```
jni: And then I would remove this logic as well, because otherwise, yikes. LOL Let's just only support tifffile and imageio, and if we miss out on something, then we can write an skimage io plugin.

jni (changes requested):
@Czaki thank you for this one! We've been trying to slim down skimage for a while, but none of those efforts have been pushed through to the end, so I can't blame you for wanting to rip it out right now! Overall, this ended up being cleaner than I expected, thanks!

One thing I would say is that it would be great to verify that everything in napari works fine without skimage, so I would consider removing it even from the test dependencies, and ripping out all the remaining references to it. What do you think?

napari/utils/io.py
Czaki: I'm not sure, but it is included in original code https://github.com/scikit-image/scikit-image/blob/master/skimage/io/_io.py#L50

I do not know why it is there present and I do not use imageio extensively to know if it is needed... 

Maybe beeter solution is to check if it is `np.ndarray` instance and convert otherwise?

napari/utils/io.py
Czaki: It is no so archaic as you think. I work with files that need this, but `tifffile` allow to do it better. Tifffile image format in metadata store axes order and do not force this order (The data with which I work has axes order ZCYX). 

In my program (PartSeg) I have separated module to read image with metadata (based on tifffile), which is prepared to extract to separate package. You can see logic to ensure axes order here https://github.com/4DNucleome/PartSeg/blob/master/package/PartSegImage/image_reader.py#L97

napari/utils/io.py
Czaki: I do not understand. How writing skimage io plugin can help if  skimage support was removed. Yo mean skimage plugin support? skimage plugin porting?

I removed this.

napari/layers/base/base.py
Czaki: done

Czaki:
> One thing I would say is that it would be great to verify that everything in napari works fine without skimage, so I would consider removing it even from the test dependencies, and ripping out all the remaining references to it. What do you think?

I remove all connected with io, but there still are:
```python
from skimage.data import data_dir
from skimage.data import binary_blobs
from skimage.transform import pyramid_gaussian
```
You prefer to copy them all to napari? 

The `from skimage.util import img_as_ubyte` can be removed from test but with cost of readability (or maintainers can assume that this part will not be changed and if it currently work then only simple tests are needed). 

If docstring from `utils/colormaps/vendored/colorconv.py` can be ignored?

What with examples?

napari/utils/io.py
jni: @Czaki what I mean is that napari lets you do `channel_axis=1` for your ZCYX image, so the channels don't need to be at the final position. The *assumption* of the position of the channel axis is not needed. Do you agree?

napari/utils/io.py
jni: I mean, right now it's saying, if it's not a numpy array, return it immediately, so it's doing the opposite of converting it. I would say that we should just remove it for now, and bring it back if it causes grief for someone. I doubt it's the right solution for the problem it's trying to solve anyway. Definitely seems like a bandaid.

napari/utils/io.py
jni: Sorry, I meant, we can write a napari IO plugin (see #1030) that uses skimage.io to read files.

jni:
Whoa, yeah, those are big dependencies. But we have a problem now: with skimage in our test dependencies, we have no way of knowing whether removing skimage breaks napari. e.g. if someone inadvertently adds an skimage dependency to some part of the production code, and we don't notice it, the tests will still pass, but napari will be broken on install.

I don't have a good solution for this... =\ Any ideas? @sofroniewn?

fwiw, for the data, we could (should?) start using pooch, as we have just done in skimage: https://github.com/scikit-image/scikit-image/pull/3945 but that is a significant undertaking.

napari/utils/io.py
Czaki: Hm. It depends. If one use napari only as viewer then it is not needed. If One would like to test general data processing algorithm (which does not need to be changed when microscope is changed) then order of axes is important

napari/utils/io.py
Czaki: ok

Czaki:
@jni Maybe this may be used to check? https://pypi.org/project/pip_check_reqs/ An then simple check on scikit image?

jni:
@Czaki I love it! I presume it can be configured to ignore certain files/directories? If so, then it would make a lot of sense to add it to the test requirements and add a test that skimage is not required in the main codebase.

napari/utils/io.py
jni: It *is* important, but my point is that it should be a parameter, not hardcoded at the end.

Czaki:
@jni I add this check. This need few changes. `requirements.txt` path is not configurable so all development releases need to be moved outside. So I create `requirements-dev.txt`. 

`pip-missing-reqs` need to have all packages installed (it import modules) and failed silently if package is missed in environment. 
Also this tool need pip in version bellow 19.3. See https://github.com/r1chardj0n3s/pip-check-reqs/pull/22

I test and current configuration fails on cirrus if some of module is missed. 

I add two missed entries, but I'm not sure which version choose:
```
toolz>=0.10.0
importlib-metadata>=1.5.0 ; python_version < '3.8'
```
They are installed as dependency of other package. 

This tool also report extra requirements. Only jupyter-client has information why is needed.
```
pip-extra-reqs napari                                                                                                                                                                                                                                         
Extra requirements:
fsspec in requirements.txt
backcall in requirements.txt
jupyter-client in requirements.txt
pyopengl in requirements.txt
```

sofroniewn:
@Czaki @jni - while I'm excited about how this is progressing, one concern I want to raise is that after this PR many of our examples won't be runnable after pip install they use `scikit-image` for data. We might need to wait until we have a data fetching system, maybe using [pooch](https://github.com/fatiando/pooch) or a similar tool before we can merge this PR.

Also I need to take a look at where in the tests we are using `scikit-image` as I also feel like if we aren't depending on it for the main package, and we aren't using it for example data, we should be able to rewrite the tests to avoid it

jni:
Ha! Richard Jones is a fellow Melbournite. =) (No, I don't know where he lives to go knock on his door, nor could I anyway under current COVID restrictions. =P)

I would also suggest waiting until the above issue is fixed, as I don't particularly want us to be broken under pip 20+!

@sofroniewn we could fix the examples problem by adding skimage to the doc requirements.

Czaki:
@sofroniewn test also needs package like pytest, pandas, zarr so it not looks like a problem when scikit_image is moved to test dependencies.

> Also I need to take a look at where in the tests we are using scikit-image as I also feel like if we aren't depending on it for the main package, and we aren't using it for example data, we should be able to rewrite the tests to avoid it

I take a look on  examples and test and rewriting such modules like `skimage.morphology` or `skimage.transform` not looks like a great idea.

here is list of test dependencies which are not satisfied after `pip install napari` with moved skimage to dev requirements 

```  
napari/utils/_tests/test_color_to_array.py:1 dist=pytest module=pytest
napari/utils/_tests/test_colormaps.py:1 dist=pytest module=pytest
napari/utils/_tests/test_io.py:9 dist=pytest module=pytest
napari/utils/_tests/test_misc.py:3 dist=pytest module=pytest
napari/utils/_tests/test_interactions.py:1 dist=pytest module=pytest
napari/utils/_tests/test_validators.py:2 dist=pytest module=pytest
napari/utils/_tests/test_key_bindings.py:4 dist=pytest module=pytest
napari/_qt/_tests/test_qt_dims.py:5 dist=pytest module=pytest
napari/_qt/_tests/test_qt_viewer.py:5 dist=pytest module=pytest
napari/_qt/_tests/test_qt_play.py:4 dist=pytest module=pytest
napari/_qt/_tests/test_qt_range_slider.py:2 dist=pytest module=pytest
napari/_qt/_tests/test_qt_range_slider_popup.py:2 dist=pytest module=pytest
napari/_qt/layers/_tests/test_qt_image_base_layer_.py:5 dist=pytest module=pytest
napari/_tests/test_dtypes.py:2 dist=pytest module=pytest
napari/_tests/test_viewer.py:4 dist=pytest module=pytest
napari/_tests/test_napari.py:2 dist=pytest module=pytest
napari/_tests/test_viewer_layer_parity.py:9 dist=pytest module=pytest
napari/plugins/_tests/test_manager.py:5 dist=pytest module=pytest
napari/plugins/_tests/conftest.py:1 dist=pytest module=pytest
napari/plugins/_tests/test_hook_specifications.py:4 dist=pytest module=pytest
napari/layers/utils/_tests/test_color_transforms.py:3 dist=pytest module=pytest
napari/layers/utils/_tests/test_layer_utils.py:2 dist=pytest module=pytest
napari/layers/image/_tests/test_image.py:6 dist=pytest module=pytest
napari/layers/image/_tests/test_pyramid.py:6 dist=pytest module=pytest
napari/layers/image/_tests/test_big_image_timing.py:4 dist=pytest module=pytest
napari/layers/image/_tests/test_image_utils.py:3 dist=pytest module=pytest
napari/layers/labels/_tests/test_labels_mouse_bindings.py:1 dist=pytest module=pytest
napari/layers/points/_tests/test_points.py:6 dist=pytest module=pytest
napari/layers/_tests/test_serialize.py:2 dist=pytest module=pytest
napari/layers/_tests/test_utils.py:2 dist=pytest module=pytest
napari/components/_tests/test_viewer_model.py:2 dist=pytest module=pytest
napari/components/_tests/test_viewer_image_io.py:6 dist=pytest module=pytest
napari/components/_tests/test_viewer_labels_io.py:1 dist=pytest module=pytest
napari/components/_tests/test_dims.py:1 dist=pytest module=pytest
napari/utils/_tests/test_io.py:13 dist=zarr module=zarr
napari/_tests/test_numpy_like.py:3 dist=zarr module=zarr
napari/layers/image/_tests/test_big_image_timing.py:2 dist=zarr module=zarr
napari/components/_tests/test_viewer_image_io.py:11 dist=zarr module=zarr
napari/utils/_tests/test_io.py:6 dist=scikit-image module=skimage.data
napari/components/_tests/test_viewer_image_io.py:4 dist=scikit-image module=skimage.data
napari/components/_tests/test_viewer_labels_io.py:3 dist=scikit-image module=skimage.data
napari/layers/image/_tests/test_pyramid.py:2 dist=scikit-image module=skimage.transform
napari/layers/image/_tests/test_image_utils.py:4 dist=scikit-image module=skimage.transform
napari/layers/_tests/test_utils.py:3 dist=scikit-image module=skimage.util
napari/_tests/test_numpy_like.py:4 dist=xarray module=xarray
napari/layers/image/_tests/test_image.py:4 dist=xarray module=xarray
napari/layers/points/_tests/test_points.py:5 dist=pandas module=pandas
napari/layers/points/_tests/test_points_utils.py:2 dist=pandas module=pandas
```

sofroniewn:
> @sofroniewn we could fix the examples problem by adding skimage to the doc requirements.

But someone who just did pip install napari still wouldn't be able to run to examples most of the examples.

> I take a look on examples and test and rewriting such modules like skimage.morphology or skimage.transform not looks like a great idea.

Yeah, we definitely would not rewrite those! We would either drop or modify those more advanced examples, or make it clear they need the additional pip install. What I don't like is that our most simple `add_image.py` example won't run

Czaki:
> Yeah, we definitely would not rewrite those! We would either drop or modify those more advanced examples, or make it clear they need the additional pip install. What I don't like is that our most simple add_image.py example won't run

`skimage.,data` is used 40 times in `examples` 
Most popular are binary_blobs which are generators (17 times) - and I think that pooch does not help with this.
Most popular image is `astronaut` used 13 times (792 kB)

Maybe some proxy in napari (napari.sample_data) which will call pip to install scikit-image?

Czaki:
`tifffile` is dependency of skimage https://github.com/scikit-image/scikit-image/pull/4235, so this PR does not introduce new dependencies.

Czaki:
@sofroniewn could you review my proposition of fix `skimage` problem? file `napari/utils/data.py`

It looks like napari test system fails (missed conda).

sofroniewn:
The napari test system has been fixed by #1119

napari/utils/data.py
```diff
@@ -0,0 +1,17 @@
+"""
+This is proxy file for examples to get data from scikit image
+"""
+
+try:
```
sofroniewn: I'll let @jni make the final call on this, but I think we don't want to trigger a pip install under the hood without someone explicitly doing it.

I know I had voiced concerns about the examples not being run-able after this PR merges, but maybe the best thing to do is just inside our tutorials make it very clear that to run our examples you'll also need to install scikit-image and give instructions on how to do that. If we do that then I'd be ok dropping this function and leave the examples unchanged.

@jni your opinion?

tlambert03: i agree, I don't mind a helpful message that says "scikit-image not found" (though perhaps the standard ModuleNotFound error is self-explanatory enough) ... but I don't think we should alter the users environment without requesting permission.

Czaki: I agree with this point of view, but my English is not enough fluent to suggest this documentation text.

jni: 100%, no auto-installs. I think it's totally fine for the examples to only run with the libraries specified in `requirements/docs.txt`, and as Talley suggests, I think ModuleNotFound is indeed explanatory enough.

jni:
@Czaki 

> path is not configurable so all development releases need to be moved outside. So I create requirements-dev.txt

We actually have `requirements/development.txt` for this purpose.

Czaki:
@jni 
> > path is not configurable so all development releases need to be moved outside. So I create requirements-dev.txt
> 
> We actually have `requirements/development.txt` for this purpose.

this is content of requirements_dev.txt
```
-r requirements/development.txt
-r requirements/test.txt
-r requirements/default.txt
```

If you think that `requirements_dev.txt ` is obsolete I will remove it. It is only shortcut.

napari/utils/data.py
Czaki: Hm. I see that examples are not included to `napari` package. So maybe other option is to copy eg. astronaut to examples dir and switch to `imageio`?

jni: No, I would like to keep the repo size small, and in the future we might switch away from astronaut anyway.

jni:
@Czaki I think you are making the argument that `requirements/development.txt` should *include* test and default. In that case, I would say that you should add `-r default.txt` and `-r test.txt` to `requirements/development.txt`. I would prefer that to `requirements_dev.txt`.

Czaki:
> @Czaki I think you are making the argument that `requirements/development.txt` should _include_ test and default. In that case, I would say that you should add `-r default.txt` and `-r test.txt` to `requirements/development.txt`. I would prefer that to `requirements_dev.txt`.

My only idea is to have one file which should be argument of pip to setup development environment.

jni:
Looks like the CI instructions need to be updated now. :grimacing: Also since there's one more change needed, I'll be nitpicky: I prefer a space after the `-r`.

napari/utils/io.py
jni: @Czaki I still would like to remove this whole clause. It is brittle and results in inconsistencies. For example, the famous `mitosis.tif` dataset has order TZCYX, *but* only two channels, so this code will not place the channel axis at the end even though that is the intention. Some other datasets might have a single channel but still have that dimension, or 5 channels. Another dataset might be simply ZYX and single-channel, but Z might only have length 4, and this code will incorrectly transpose that axis.

As you point out, the correct way to do this is with metadata, not with guesses about the shapes of things.

requirements/test.txt
```diff
@@ -6,3 +6,4 @@ pytest-ordering
 pytest-timeout
 xarray
 zarr
+scikit-image
```
jni, line 9: Could you also add scikit-image to `requirements/docs.txt`?

jni (changes requested):
@Czaki this is very close! I've asked for two changes, should be quick! Thank you for your patience with this PR!

requirements/test.txt
Czaki: done

jni (approved):

jni:
Thank you @Czaki!!!

Czaki:
@jni I check this now. I load lsm to imagej and save it as tiff.

```
im.series[0].axes
Out[12]: 'ZCYX'
```

And if I good remember it is imagej standard behaviour. So It looks like napari lost possibility to proper read data preprocessed in imagej (but as you good pointed this check may easy fail). 

As part of my Code I have reading data with metadata https://github.com/4DNucleome/PartSeg/tree/master/package/PartSegImage. It is wrapper around few @cgohlke package  

I have plan to extract it to separate package (I delay this because of additional work with synchronization between projects).

Use it as io backed allow to read also `czi` and `oif` but also add additional dependecy.

jni:
@Czaki 

> And if I good remember it is imagej standard behaviour. So It looks like napari lost possibility to proper read data preprocessed in imagej (but as you good pointed this check may easy fail).

By "lost the ability", do you mean, it no longer displays it as RGB? Indeed, as I mentioned, I don't think this was "correct" behaviour, it just approximated it. (e.g. there's nothing to say that the channel colours were RGB.) This problem would be fixed by #1103, correct?

We could also create a "imagej-tiff" reader plugin that reads channel order from the metadata and then creates the appropriate channels as layers in additive blending mode.

Czaki:
@jni For me it is only workaround, but maybe best option for now. For me best option is to read metadata and have deterministic way of presenting. Independent if it is imagej-tiff or ome-tiff.

tlambert03:
Hey @Czaki,

quick question: can you comment on why `"pip<19.3"` is in the CI `requirements_task` install step?  is there something about `pip_check_reqs` that needs an older version of pip?

Czaki:
> Hey @Czaki,
> 
> quick question: can you comment on why `"pip<19.3"` is in the CI `requirements_task` install step? is there something about `pip_check_reqs` that needs an older version of pip?

Yes. https://github.com/r1chardj0n3s/pip-check-reqs/pull/22

PR #1063, opened 2020-03-19
add error raise in Viewer._add_layer_from_data

tlambert03:
# Description
2-liner here... in `Viewer._add_layer_from_data`, I added a try/except to provide a more useful error message in one certain case, but forgot to re-raise the exception otherwise (so if the `add_method` fails with a different `TypeError`, it passes).


kevinyamauchi (approved):
Looks good to me!

jni (approved):

PR #1069, opened 2020-03-20
Fix empty points with properties

### LINKED ISSUE #930: Adding points with properties / empty properties

kevinyamauchi:
# Description
This PR fixes the `Points` layer behavior when instantiating an empty layer with properties (issue described in #930). We introduce a new property to store default values for properties (`Points.default_properties`). When a new point is added to an empty layer, new points are given property values from `Points.default_properties`.  

# Type of change

# References
This PR fixes #930.

# How has this been tested?
- [x] added tests for adding points to empty layers

## Final checklist:


kevinyamauchi:
For an example of how this can be used, see this [work in progress annotator GUI](https://github.com/kevinyamauchi/PointAnnotator/blob/master/pointannotator/gui.py).

tlambert03:
cool!  just took it for a spin :)

for the magicgui part, in case you're curious, here's an (equivalent) way to code what you did there without the Enums, and with slightly simpler events (that don't require using QWidget-specific methods):
```python
        # you can set 'choices' on an argument instead of creating an Enum if you want
        @magicgui(label={'choices': labels})
        def label_selection(label):
            return label

        label_menu = label_selection.Gui()

        # since label is no longer an Enum, it can just be used directly
        def label_changed(label):
            """Update the Points layer when the label menu selection changes"""
            points_layer.current_properties = {'label': np.asarray([label])}

        # a nicety within magicgui is that you can directly set properties
        # on the widget, and the GUI will update accordingly
        # so Qt-specific "setCurrentIndex(index)" is not needed.
        def update_label_menu(event):
            """Update the label menu when the point selection changes"""
            label_menu.label = points_layer.current_properties['label'][0]

        # connect the events and add the label menu widget
        points_layer.events.current_properties.connect(update_label_menu)
        label_menu.label_changed.connect(label_changed)
        viewer.window.add_dock_widget(label_menu)
```

kevinyamauchi:
Thank you for the tips, @tlambert03 ! Super cool. I think `magicgui` is going to be really great for people who want to extend napari with their own UI elements!

napari/layers/points/points.py
```diff
@@ -471,6 +532,44 @@ def properties(self, properties: Dict[str, np.ndarray]):
...
             warnings.warn('property used for face_color dropped')
 
+        if self._edge_color_property and (
+            self._edge_color_property not in self._properties
+        ):
+            self._edge_color_property = ''
+            warnings.warn('property used for edge_color dropped')
+
+    @property
+    def current_properties(self):
```
sofroniewn, line 578: Can you add a doc string

napari/layers/points/points.py
```diff
@@ -275,6 +279,17 @@ def __init__(
...
         self._properties = self._validate_properties(properties)
 
+        if default_properties is None:
+            default_properties = {}
+        elif not isinstance(default_properties, dict):
+            default_properties = dataframe_to_properties(default_properties)
+        self._default_properties = default_properties
+
+        if len(data) == 0 and self.properties and not default_properties:
+            raise ValueError(
```
sofroniewn: Do we also want to check the actual keys from `properties` are present in `default_properties`? Should you be forced to provide a default for every property in this case? I can trigger errors with this combo say
```python
    # create properties for each point
    properties = {
        'class': np.array([]),
    }
    default_properties = {'label': [0, 1, 2]}
```
might be better to catch them here first?

sofroniewn (approved):
@kevinyamauchi - works well. I left some minor comments. I'd like @jni to look too before merge as we are expanding the API here.

The annotator example with magicgui is amazing (I adapted your script [here to make this demo](https://github.com/sofroniewn/image-demos/blob/master/examples/annotator.py) it is so much fun!!!)

Having some problems with my gifs, will try and upload one soon

tlambert03:
sidenote about `magicgui` here (unrelated to this PR):
magicgui provides no alternative, but I find this type of usage unfortunate (where you really didn't need to make a function):
```python
@magicgui(label={'choices': labels})
def label_selection(label):
    return label
```
I created https://github.com/napari/magicgui/issues/7 to remind me to make it easier to directly create a widget in this type of scenario

sofroniewn:
annotator

napari/layers/points/points.py
kevinyamauchi: Sorry! Done.

napari/layers/points/points.py
kevinyamauchi: Good point! I added a check and a test for the check.

kevinyamauchi:
Thanks for the review and comments, @sofroniewn ! I have addressed Nick's comments, so I think it is ready for review, @jni .

napari/layers/points/_tests/test_points.py
```diff
@@ -15,6 +15,44 @@ def test_empty_points():
     assert pts.data.shape == (0, 2)
 
 
+def test_empty_points_with_properties():
+    """ Test instantiating an empty Points layer with properties
+        See: https://github.com/napari/napari/pull/1069
+    """
```
jni: Minor point: please make sure all docstrings conform to [PEP257](https://www.python.org/dev/peps/pep-0257/). In this case:
```suggestion
    """Test instantiating an empty Points layer with properties.

    See: https://github.com/napari/napari/pull/1069
    """
```

napari/layers/points/_tests/test_points.py
```diff
@@ -799,6 +859,28 @@ def test_face_color_cycle():
...
+    color_cycle = ['red', 'blue']
+    layer = Points(
+        properties=annotations,
+        default_properties=default_properties,
+        face_color='point_type',
+        face_color_cycle=color_cycle,
+    )
+
+    # add a point
+    layer.add([10, 10])
```
jni, line 989: Something that would be useful to add to this PR: allow adding a point together with its properties:

```python
layer.add([10, 10], {'point_type': 'B'})
```

jni: Note: if we pass this value through `dict`, we can also accept `pd.Series` objects!

```python
In [1]: import pandas as pd
In [2]: pt = pd.Series({'point_type': 'B'})
In [3]: dict(pt)
Out[3]: {'point_type': 'B'}
```

jni: (which are how individual rows (or columns) of pandas dataframes are represented.)

napari/layers/points/points.py
```diff
@@ -44,6 +44,8 @@ class Points(Layer):
     properties : dict {str: array (N,)}, DataFrame
         Properties for each point. Each property should be an array of length N,
         where N is the number of points.
+    default_properties : dict {str: array (N,)}
+        Default property values for each property in Points.properties
```
jni: A few things that discussed during the meeting that I think are really important:

1. I want us to be very cautious about expanding the API here. I see the necessity, but let's take a little bit of time to get it right. Given that the discussion has moved to allowing *choices* here rather than default values, I would at a minimum rename this to `properties_choices`.

2. Additionally, it's unclear that this is the right model — what about continuous properties? How do we represent those? One suggestion is to pass in a range of possible values, but even that is tricky: is `[0.2, 0.5]` a range of floating point values or two possible values that happen to be floats? We could repurpose `range(a, b)`, but that doesn't work for floating point a or b. We could make our own custom object, but that's clunky... Also, what if we want to allow *any* floating point number, not just numbers in a range? I don't want to stall this PR, it's critically important, but I do think floating point properties are a fundamental part of the API (ie not an edge case), and we should pick a solution that will work for those.

3. Finally, another use case that I think is important is using `add_points()` (no arguments), clicking on a bunch of points, and then annotating them after the fact. In this case, a user might say `points.properties = {'class': ['A', 'B', 'A', 'B']}`. Will the user be able to add points after that? If so, `default_properties` seems to have a pretty limited shelf life within the API, and it seems annoying to add this to the function signature for something that will be used only during an infinitesimally short time. Again, `properties_choices=` might actually have a longer shelf life, because we can use them to include categories that may not exist in the first n points, or even in that particular image, but do exist in later parts of the dataset.

4. Ultimately, I do think we want to include a table viewer/editor to deal with all this, but, as @sofroniewn pointed out during the meeting, this is a heavyweight option that shouldn't be the primary interaction mode.

5. The lightweight option, exemplified by the combobox in the latest example script, should ideally include (a) the ability to add a new category, (b) every column, which means different interaction modes for different column types.

napari/layers/points/points.py
```diff
@@ -275,6 +279,25 @@ def __init__(
...
+            default_properties = dataframe_to_properties(default_properties)
+        self._default_properties = default_properties
+
+        if len(data) == 0 and self.properties and not default_properties:
+            raise ValueError(
+                'default_properties must be set for empty Points layers with colors that depend on properties'
+            )
+
+        if (
+            self._default_properties
```
```suggestion
            self._default_properties is not None
```

jni: (Note that an empty dict evaluates to false! Explicit is better than implicit.)

napari/layers/points/points.py
```diff
@@ -275,6 +279,25 @@ def __init__(
...
 
+        if default_properties is None:
+            default_properties = {}
+        elif not isinstance(default_properties, dict):
+            default_properties = dataframe_to_properties(default_properties)
+        self._default_properties = default_properties
+
+        if len(data) == 0 and self.properties and not default_properties:
+            raise ValueError(
+                'default_properties must be set for empty Points layers with colors that depend on properties'
```
jni: Remember that size and symbol should eventually also be settable by properties, so "colors" isn't right here. Let's say, "properties_choices must be set for empty Points layers with properties".

napari/layers/points/points.py
jni: I just saw that you make it an empty dict, above. So, the question becomes, if I pass `properties={'labels': []}` and `default_properties={}`, do we want a KeyError or a ValueError?

napari/layers/points/points.py
```diff
@@ -336,13 +359,41 @@ def __init__(
...
+                    colors=face_color,
+                    elem_name="face_color",
+                    default="white",
+                )
+            else:
+                self._current_face_color = np.array([1, 1, 1, 1])
+        else:
+            self._current_edge_color = self.edge_color[-1]
+            self._current_face_color = self.face_color[-1]
+            self.current_properties = {}
```
jni, line 424: I'm quite wary of big if/else cascades with layers of nesting. At the very least, this should be abstracted away into its own function. But it would be worth spending some time to try to simplify it. For example, we follow the exact same code path for both edge and face color, with a few things changed, so it might be worth writing `self._set_current_color(element='face', color=face_color)`, then use `getattr` to grab the `f'_{element}_color_mode'` element as required, and set the `f'_current_{element}_color'` attribute.

napari/layers/points/points.py
```diff
@@ -336,13 +359,41 @@ def __init__(
...
+                self._current_edge_color = np.array([0, 0, 0, 1])
+            if self._face_color_mode == ColorMode.DIRECT:
+                self._current_face_color = transform_color_with_defaults(
+                    num_entries=1,
+                    colors=face_color,
+                    elem_name="face_color",
+                    default="white",
+                )
+            else:
+                self._current_face_color = np.array([1, 1, 1, 1])
```
jni: I don't understand why this is what happens when the color mode is *not* direct...?

napari/layers/points/points.py
```diff
@@ -427,11 +487,20 @@ def data(self, data: np.ndarray):
...
                         self._face_color_property
                     ][0]
+
+                    # check if the new edge color property is in the cycle map
+                    # and add it if it is not
+                    face_color_cycle_keys = [*self.face_color_cycle_map]
+                    if face_color_property not in face_color_cycle_keys:
+                        self.face_color_cycle_map[face_color_property] = next(
+                            self.face_color_cycle
+                        )
```
jni, line 528: As above... the amount of repeated code for face color and edge color is making me quite uncomfortable!

napari/layers/points/points.py
```diff
@@ -471,6 +540,45 @@ def properties(self, properties: Dict[str, np.ndarray]):
...
 
+        if self._edge_color_property and (
+            self._edge_color_property not in self._properties
+        ):
+            self._edge_color_property = ''
+            warnings.warn('property used for edge_color dropped')
+
+    @property
+    def current_properties(self):
+        """Dict[str: np.ndarray]: properties for the next added point."""
```
jni: The type annotation should be a type annotation rather than part of the docstring. =)

napari/layers/points/points.py
```diff
@@ -471,6 +540,45 @@ def properties(self, properties: Dict[str, np.ndarray]):
...
+            for k in props:
+                props[k][self.selected_data] = current_properties[k]
+            self.properties = props
+
+            self.refresh_colors()
+        self.events.current_properties()
+
+    @property
+    def default_properties(self):
+        """dict {str: array (N,)}, DataFrame: default annotations"""
```
jni: As above... Let's aim for consistency in typing and docstrings!

jni: As I understand it, we will never actually have a dataframe here, right? Also, dataframes can't have different-length columns, so it would not be the right model for this data.

jni (changes requested):
Hi @kevinyamauchi! Sorry that it's taken me a while to have a close look at this.

I've made quite a few comments. I think this PR is critical, but I think we haven't yet found the solution for the main problem, which is: how do we update point properties, both from an API and GUI perspective? `default_properties` seems to me like more of a placeholder rather than the final solution, because it doesn't let me add new choices, or express that a new property should be a floating point value, or an integer value. Ideally, we want to be able to add columns to our data properties *after* layer creation.

Additionally, I think the size and branching structure of the code is getting a bit out of hand, and this is probably a good time to look for simplifications, rather than adding lots of new branches. The codecov report shows a few missed ones:

https://codecov.io/gh/napari/napari/compare/fa1cd4029b30112d5c763d4147c6a1fddb23d75b...487a0fb89afcea9f1684b94d80a6a27e1f36a443/diff

(Though to be honest I was surprised by how many *were* covered! :sweat_smile:)

Let me know what you think of my comments/suggestions. I think the suggestions I've made are the minimum to get this mergeable, but even then I'm not totally satisfied that we've nailed the API. Happy to schedule a live discussion with you in the coming days @kevinyamauchi if you want to shortcut a lot of back and forth on this issue! But, happy to hear your thoughts on my feedback here in the meantime — please let me know if you think I'm being unreasonable/annoying! :grimacing:

kevinyamauchi:
@jni, thank you for your thorough review! We are definitely aligned in that I am also growing uneasy with the increasing size/branching of the points layer code. I am definitely open to going in a different direction with this. I think it would be great to have a call to discuss once I've had a chance to think about all of your comments. I will message you on slack to set up a time!

jni:
s/slack/zulip/

=P

kevinyamauchi:
@jni discussed this PR and I think came to a nice solution. I have summarized the plan below. Please feel free to comment or provide suggestions. In the meantime, I will start implementing the changes.

## Initializing empty Points layers with properties
We will rename the `Points.default_properties` to `Points.property_choices`. `Points.property_choices` will be used to both provide default values and in the future options for property values. We will not perform validation of property values against `Points.property_choices`. Also, `Points.property_choices` will not be synchronized with `Points.properties`. However, the union of `Points.property_choices` and the unique values in `Points.properties` can be used to get the current property value options (e.g., in the case of a property value picker UI).

We will eliminate the `default_properties` kwarg and instead populate `Points.property_choices` from the `properties` kwarg. This will work as follows:

- If the user does not provide properties, `Points.property_choices` is an empty dictionary.
- If the user provides data (i.e., the layer is not empty) and properties, `Points.property_choices` is populated with the unique value of each property.
- If the user does not provide data (i.e., the layer is empty), but does provide properties, the dictionary passed to properties will be interpreted as `Points.property_choices` and `Points.properties` will be initialized with a dictionary with the appropriate keys and empty numpy arrays for the values.

## Code branching/duplication
There is an increasing amount of code duplication and branching as we increase the interactivity of the layer. For this PR, we will focus on making sure all branches of the code are tested. We will explore using `getattr` to reduce duplication between `edge_color` and `face_color` code in a future PR.

sofroniewn:
Nice ideas @jni @kevinyamauchi green light from me!

napari/layers/points/points.py
kevinyamauchi: > As above... Let's aim for consistency in typing and docstrings!
Fixed!


> Also, dataframes can't have different-length columns, so it would not be the right model for this data.

Good point. Fixed.

kevinyamauchi:
I've implemented the things we discussed, @jni! I will take the opportunity to add a couple tests. In the meantime, if you have time to take a look, it would be great if you could see if we are on the same page.

sofroniewn (approved):
This looks really nice @kevinyamauchi good job to you and @jni for figuring out how to fix the bug without expanding the API!

napari/layers/points/points.py
```diff
@@ -273,10 +276,17 @@ def __init__(
...
             properties = dataframe_to_properties(properties)
-        self._properties = self._validate_properties(properties)
+            self._properties = self._validate_properties(properties)
+            self._property_choices = {
+                k: np.unique(v) for k, v in properties.items()
+            }
+        elif len(data) == 0:
+            self._property_choices = properties
+            empty_properties = {k: np.empty(0) for k in properties}
+            self._properties = empty_properties
```
jni, line 296: I’d like this to be a tiny bit more elaborate. Specifically, properties can be a choice among objects, where np.unique makes sense, but it can also be a float or an int. I’d like to support the following syntax:

```python
layer.add_points(properties={‘class’: [‘a’, ‘b’], ‘value’: float})
```

The empty arrays should then have the correct dtype for the given property choices.

napari/layers/points/points.py
jni: @kevinyamauchi did you see this question?

jni (commented):
@kevinyamauchi, from an API perspective I think there is one more change this PR needs, noted below. I think there are also some code branches that need testing. Finally, I had a question about one of the branches, that I’ve pinged below.

napari/layers/points/points.py
kevinyamauchi: I agree that we should make sure we are initializing the property arrays with the correct dtype. However, I think the user always needs to provide at least one value so that when a point gets added to the empty layer, it is given a value for all properties. I propose that we continue requiring that properties is a dictionary with values that are numpy arrays and that we use the `properties` dtype to populate the `empty_properties` with arrays of the correct dtype.

To do this, we can change lines 288 in the snippet above to

```python
empty_properties = {k: np.empty(0, dtype=v.dtype) for k, v in properties.items()}
```

Constructing the layer now looks like:

```python
layer.add_points(
    properties={
        ‘class’: np.array([‘a’, ‘b’]),
        ‘value’: np.array([0], dtype=np.float)
    }
)
```

How does that sound?

napari/layers/points/points.py
kevinyamauchi: Ah, sorry about that - missed this!

This situation is that we need to initialize with a `_current_*_color` value and when we have an empty layer where the colors are mapped to a property, we don't really have a "current" color. I had initially given a placeholder default value during development and forgot to circle back to fix it. I think we can now give it the color that would correspond to the default property value from `property_choices`

napari/layers/points/points.py
```diff
@@ -185,6 +185,8 @@ class Points(Layer):
 
     Extended Summary
     ----------
+    _property_choices : dict {str: array (N,)}
+        Possible values for the properties in Points.properties
```
tlambert03:
```suggestion
        Possible values for the properties in Points.properties.
        If ``properties`` is not provided at instantiation, will be undefined.
```

napari/layers/points/points.py
tlambert03: late to the party, but just curious whether it wouldn't be possible to call np.array() on whatever the user provides?  I definitely also like allowing the `Points(properties={'class': ['a', 'b']})` syntax if at all possible.

jni: @kevinyamauchi yeah, that works great for now! I still like `'value': float` as readable syntax, but we can leave that for a future PR.

I definitely agree with @tlambert03 that it would be great to not have to pass `np.array()`

napari/layers/points/points.py
kevinyamauchi: Thank you for catching this, @tlambert03! I didn't intend to make it undefined. I've fixed it!

napari/layers/points/points.py
kevinyamauchi: Sounds good. I've implemented it!

kevinyamauchi:
Thank you for the reviews, @jni and @tlambert03 ! I believe I have addressed your comments.

jni (approved):

jni:
:tada: :rocket:

tlambert03:
Great work!

PR #1076, opened 2020-03-23
fix image format

Czaki:
# Description

There are cases when `self._qt_window.grab().toImage()` produces image in QImage.Format_RGB16 (on azure pipelines linux for example)

## Type of change

# References

# How has this been tested?

test start passing on azure pipelines

## Final checklist:

### DIFF
--- a/napari/_qt/utils.py
+++ b/napari/_qt/utils.py
@@ -5,7 +5,7 @@ from typing import Type
 import numpy as np
 from qtpy import API_NAME
 from qtpy.QtCore import QObject, QSize, Qt, QThread
-from qtpy.QtGui import QPainter, QPixmap
+from qtpy.QtGui import QPainter, QPixmap, QImage
 from qtpy.QtWidgets import QGraphicsOpacityEffect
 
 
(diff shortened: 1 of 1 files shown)


sofroniewn (commented):
This looks good to me, thanks for the catch and fix @Czaki.

Before I approve can you confirm that `viewer.screenshot(with_viewer=True)` also works in your usage. It uses a slightly different approach to getting the image - https://github.com/napari/napari/blob/master/napari/_qt/qt_main_window.py#L382 and should be converted too if it needs to be. I don't think it will, but worth double checking.

Also, when you say azure pipelines linux, do I take it that you are then running this in the cloud on linux in azure and taking screenshots? After this fix, do you then see the images as you would see them if you had run napari on your laptop and taken a screenshot there?

We've had some discussions about using screenshots like this during our tests or for autogeneration of examples for our docs (cc. @kne42 @jni @tlambert03) but haven't gone down that path. We might want too!

Czaki:
But I do conversion in `QImg2array` function, not in place where image is created. 

In my program i use Azure pipelines as a CI service. So I decide to check which problems can occurred during test run after include napari as viewer, so I setup napari test on azure pipeline and this problem occurred (on your tests).  

I'm not sure if screen shots looks good, but tests pass. 

I still have problem with windows tests:
```
RuntimeError: Using glBindFramebuffer with no OpenGL context.
```

sofroniewn (approved):
Thanks for checking that with_viewer screenshot works fine too, I thought it would, just wanted to make sure.

I am approving this now, will let sit till tomorrow and then merge.

As to the windows on azure openGL warning - we've actually had a lot of those problems ourselves too. There are a number of vispy / napari issues around this. Maybe @tlambert03 or @kne42 can point you to some of them. I think our stratgey has just been to disable tests that require openGL on windows

PR #1080, opened 2020-03-23
pyenv with PyQt5 in other enviroment workaround

Czaki:
# Description
When using `pyenv` to manage python versions and have installed PyQt5 in any environment 
when user calls `prccc5` from environment which not contains this tool then got message like:

```
pyenv: pyrcc5: command not found

The `pyrcc5' command exists in these Python versions:
  3.6.9/envs/partseg3.6
  3.7.4/envs/partseg3.7
  partseg3.6
  partseg3.7
```

So I suggest to not only check existence of file but also return code and if fail then try ones from PySide2. 

## Type of change

# References

# How has this been tested?

can run without using PyQt5 tool to generate resources

## Final checklist:

### DIFF
--- a/napari/resources/build_icons.py
+++ b/napari/resources/build_icons.py
@@ -6,7 +6,7 @@ for styling svg elements using qss
 import os
 import re
 import shutil
-from subprocess import run
+from subprocess import check_call, CalledProcessError
 from typing import Dict, List, Tuple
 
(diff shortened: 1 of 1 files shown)


tlambert03 (approved):
looks good to me!  thanks for the fix.

Issue #1074, opened 2020-03-21
Installing plugins from bundled app


tlambert03:
## 🚀 Feature
As discussed in the last meeting, we'll need a way to extend bundled apps with plugins.  So this issue is here to propose and discuss ideas.

After playing with it a little bit, I have one possible solution:
As long as `pip` is included in the bundle (easily done by adding it to `hiddenimports` in the pyinstaller spec), people can pip install packages... but will _not_ of course, be able to install them to the frozen bundle itself (actually, I found that the default behavior was to install into the base python environment 🤮).  Pip provides both the `--target` and `--prefix` options that let you specify where things should be installed (the difference is that `--target` seems to install _all_ of the dependencies to that target directory, whereas `--prefix` first checks the existing PYTHONPATH and only installs what is missing to the prefix directory).
As a proof of principle, I just tried pip installing `napari-dv` to `~/.config/napari/plugins` and then used `napari.plugins.plugin_manager.autodiscover()`, and provided that path... and it worked (I could read `.dv` files from the bundled app).

I would propose that we have a function (& GUI) within napari that lets users install a pip or conda dependency (using `--prefix`), use [appdirs](https://github.com/ActiveState/appdirs) to get the appropriate platform-specific `user_data_dir`, then run `plugin_manager.autodiscover(path=user_data_dir)` on napari launch (or after plugin install, it wouldn't even require a relaunch after a plugin was installed 🎉).  We could also detect whether napari is being run from a bundle or not, and have the function just `pip install` into the regular current environment if not frozen.

thoughts?

(as a sidenote, i think `appdirs` would be useful for #834 as well... and we should just vendor it, it's a single file that hasn't had a pypi release in too long).

tlambert03:
some more info here.  I was able to make a nice "`install_plugin`" function that installs packages with pip to some platform-specific folder (outside of the environment), which then loads fine with our `plugin_manager.discover()` method.  
However, it's going to be rather tricky to "bundle" pip with pyinstaller for a couple reasons:
- pip _strongly_ discourages using anything but the CLI to install things... they don't support programmatic install with anything other than `subprocess.run(['python', '-m', 'pip', 'install'])`.  (Nonetheless, I tried bundling `pip`, but it didn't work... just hung when I tried to install).
- however, I don't yet know how to get at "`python`" within a bundled app.  normally, you could use `sys.executable` to access the current interpreter, but in our bundled app, `sys.executable` does not resolve to "python" but rather to `napari.__main__`
[...]

Issue #896, opened 2020-01-21
dbus_message_get_allow_interactive error after installation 


erjel:
Hi,

I would like to use napari on a compute cluster for vizualisation purposes (Connection via noVNC browser Interface).

About the system:
conda version: 4.6.14
OS: SUSE Linux Enterprise Server 12 SP4

I installed napari as folllows:
```
conda create -n napari python=3.6
conda activate napari
pip install napari
```

When want to start napari from the terminal
```
$ napari
```
 I get the error message:
```
***${HOME}***/conda-envs/napari/bin/python: relocation error: ***${HOME}***/
conda-envs/napari/lib/python3.6/site-packages/PySide2/Qt/plugins/platforms/../..
/lib/libQt5DBus.so.5: symbol dbus_message_get_allow_interactive_authorization,
 version LIBDBUS_1_3 not defined in file libdbus-1.so.3 with link time reference
```

Have you ever seen this error before?

kne42:
Hi @erjel!

No, I have not seen this before (I'm not a linux user). While you wait for one of our devs who are familiar with linux to respond, could you try installing the following packages?

```
$ apt-get install libgl1-mesa-glx libqt5x11extras5
```

erjel:
Hi @kne42!

Wow, that was a fast reply! Thank you for your quick response.
Unfortunately I have no admin rights on the cluster, so I can not install any new packages.

In order to install other packages than the provided ones, I have to compile them myself in my user space, which I would like to avoid.

kne42:
Ok! Hopefully one of the other devs will be able to resolve this for you :)

erjel:
Hi,

I double checked with the support team of the cluster. They figured out a solution, which worked for me:

```
module load <anaconda environment>
pip install --user napari
```
This installed napari in $HOME

For starting napari I run
```
vglrun ~/.local/bin/napari
```

In order to remove napari:

```
module load <anaconda environment>
pip uninstall napari
```

tested with anaconda 3 (2019.03)

tlambert03:
oh nice!  good to know about [VirtualGL](https://wiki.archlinux.org/index.php/VirtualGL).  we should check to see if that is a suitable workaround for some other remote access problems (like with X11 forwarding).  thanks for reporting back

tlambert03:
[...]

Issue #886, opened 2020-01-17
napari works over google chrome remote desktop!


tlambert03:
File this under potentially interesting rainy day information that doesn't really require any immediate action, but which has implications for remote GL-based rendering.

As has been mentioned [elsewhere](https://github.com/napari/napari/issues/799#issuecomment-565859004), openGL has issues with virtualized desktops, making things like microsoft remote desktop, x11 forwarding on *nix, and (sometimes) tests on our CI fail.  My limited understanding is that the way it is implemented generally limits openGL to older versions (e.g. ≤1.4) that are incompatible with vispy, and [QWindow](https://doc.qt.io/qt-5/qtgui-index.html#opengl-and-opengl-es-integration) rendering using openGL.

Interestingly, napari works just fine using google [chrome remote desktop](https://remotedesktop.google.com/).  I don't really understand why, and googling it hasn't really turned up anything obvious.

Perhaps related (but I'm not really sure), is the [ANGLE library](https://github.com/google/angle) from google which appears to be a layer between OpenGL ES API calls and platform-specific hardware APIs. there is also [discussion about ANGLE in the qt wiki](https://wiki.qt.io/Qt_5_on_Windows_ANGLE_and_OpenGL).

again.  not really sure what to do with that info.  It's also completely possible that other solutions (e.g. teamviewer) would work too and we just haven't tested them yet.  None of this really matters for our more immediate remote goals, but it's just interesting :) and potentially a short-term workaround for some users who really want napari to work in remote desktop environments.

jni:
Well, as much as I hate to create features that are Chrome-only (perhaps there is similar technology from Mozilla?), *ideally* the thing that would be great to do with this information is to figure out how to properly configure a JupyterHub where launching a napari viewer would open a new tab with a virtual napari window in it. Then teaching image analysis in Python using a JupyterHub would be easy!

https://forum.image.sc/t/image-viewer-for-python-notebooks/32976/

(Though note @guiwitz's last response, which I remembered but had never run so I didn't want to propose it as a solution.)

sofroniewn:
@choldgraf - whom I happen to be sitting next to right now :-) curious how you think of the above in terms of napari / jupyterHub integration

choldgraf:
Hey all - a quick thought here, I wonder if the Jupyter Server Proxy would be a way to accomplish this. Basically, if you have a web-app running on a port, then you can proxy that within a Jupyter session and allow users access to it.

A few examples:

[...]

PR #1082, opened 2020-03-24
[WIP] Create executable using pyinstaller and tests  on azure pipelines.

Czaki:
# Description
During prepare of embedding napari in my application I would like to identify all possible problem connected with napari. So I apply test scheme and pyinstaller build scheme to napari. 

From my projects I have bad experience with using pyinstaller on anaconda. Now I prefer to use CPython distribution. 

1. This PR can boost #496 or replace it (napari related things are bring from this PR). 
2. PR is to add information about problem connected with napari on azure pipelines (OpenGL on windows - I need help with this)
3. Introduce tox. 

result of build can be seen here https://dev.azure.com/bokota/open%20source%20contrib/_build?definitionId=8&_a=summary

PyInstaller support of PySide2 is worst than PyQt5. I need to change backend on macos and windows, because PySide2 base build fails during time of execution. It looks like it need additional hooks for PySide2. (Here you see why I look for PyQt5 support)

Even If it will not be finally merged to napari please keep it until all problems are fixed:

- [ ] OpenGl test on windows
- [ ] Icons problem on linux. 

## Type of change

# References

# How has this been tested?

Download result of pipeline on each system and test if it launch. 

## Final checklist:

@sofroniewn can you look on this. Did you know who can help with OpenGl?


tlambert03:
Thanks again for all the great PRs @Czaki,
I realize it doesn't look like there is much activity at #496, but I've actually been working on it a lot over the last few days in my branch: https://github.com/napari/napari/compare/master...tlambert03:pr/sofroniewn/496
I have it working with pyside2 in that branch, and I've been trying to come up with a strategy to allow people to extend it with plugins after bundling (see #1074).  It'd be nice if we could keep all of the pyinstaller activity on the #496 thread rather than creating a new one.  (It's a major feature that we're all very excited to include, but there's a lot of moving parts)

are you proposing here that we add azure pipelines to our tests?  or is this just something you're working on on the side for your project?

Regarding OpenGL in virtualized environments... there is a lot to be said there.  I would recommend starting with https://github.com/napari/napari/issues/799#issuecomment-565859004, https://github.com/napari/napari/issues/896#issuecomment-577321885, https://github.com/napari/napari/issues/886, and perhaps [this one](https://github.com/vispy/vispy/issues/78) at Vispy.  I can't claim I fully understand it... but after looking into it for a while it definitely seems like a hard problem due to limited support for openGL2.0 on many virtualized window servers.  I think  [virtualGL](https://www.virtualgl.org/) might be a potential solution ... but it will take some research.

Czaki:
> I realize it doesn't look like there is much activity at #496, but I've actually been working on it a lot over the last few days in my branch: https://github.com/napari/napari/compare/master...tlambert03:pr/sofroniewn/496
I have it working with pyside2 in that branch, and I've been trying to come up with a strategy to allow people to extend it with plugins after bundling (see #1074). It'd be nice if we could keep all of the pyinstaller activity on the #496 thread rather than creating a new one. (It's a major feature that we're all very excited to include, but there's a lot of moving parts)

I agree, but there are few reasons:
1) I'm not a maintainer of napari, so I only can create PR to PR or leave comment. First solution is weak for this moment because #496 need to be rebased. This PR have few important difference, which may be hard form me to explain in comments so I decide to create PR to show it.
2) This PR base on CPython, not Anaconda.
3) I switch to PyQt5 backend which is better supported by PyInstaller. 
4) I force to create `napari/resources/_qt_resources_*.py` during bundling time (versioner create some problems)

But I will be happy if both cases. If this PR is merged or if strong part of this PR will be moved to #496. 

> are you proposing here that we add azure pipelines to our tests? or is this just something you're working on on the side for your project?

Hard question.
Running on  azure pipelines allow to use CPython, current solution base on Anaconda. It can make difference. 
I would like to have at least information about deal with OpenGL in documentation. 

So I try to solve OpenGL problem basing on your links.

tlambert03:
@Czaki, now that we've closed #496 and #1289 is pretty much working, is it ok with you if we close this PR?

Czaki:
Briefcase looks promising. I do not understand why its ceeate installer on Windows.

But as i chcek pyinstaller is not compatible with napari plugin system. So close this PR.

If briefcase works well with napari I propably would ask few qestions on zulip because i still think if I could apply it to my software.

tlambert03:
>  I do not understand why its create installer on Windows.

as opposed to something like a portable executable?  I assume they went with the .msi installer as the default bundler because it lets you integrate nicely into the start menu, and adds an uninstaller, etc... but I don't know much about the various windows install methods.  I will point out that the code is pretty nicely organized such that that if you wanted a different type of bundle, you can probably create a new template for it (it would go in [this folder](https://github.com/beeware/briefcase/tree/master/src/briefcase/platforms/windows))

> If briefcase works well with napari I propably would ask few qestions on zulip because i still think if I could apply it to my software.

sure,  happy to share what we've learned.  There's some good discussion in #1074 and #1289, and you can try the [latest bundles](https://github.com/napari/napari/pull/1289#issuecomment-647062277) as an example if you'd like.

PR #1083, opened 2020-03-24
Fix function and signature to match with code

Czaki:
# Description
During prepare #1082 I found that signature of function inform that function return `None` but it actually return part to file. This PR contains changes in signature and in documentation.

## Type of change

## Final checklist:

### DIFF
--- a/napari/resources/__init__.py
+++ b/napari/resources/__init__.py
@@ -46,7 +46,7 @@ def _try_touch_file(target) -> Optional[Path]:
     return target
 
 
-def import_resources(version: str = '', overwrite: bool = False) -> None:
+def import_resources(version: str = '', overwrite: bool = False) -> str:
     """Build and import our icons as Qt resources.
 
     This function attempts to write that file to one of three locations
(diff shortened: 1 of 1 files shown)


napari/resources/__init__.py
```diff
@@ -71,7 +71,13 @@ def import_resources(version: str = '', overwrite: bool = False) -> None:
             - the resources file does not already exist.
             - ``overwrite`` argument is True
             - the ``NAPARI_REBUILD_RESOURCES`` environmental variable is set
-    
+
+    Returns
+    -------
+    out_path : str
+        Path to the python resource file.  Import this file to make the SVGs
```
tlambert03: can you change this docstring to indicate that the file _is_ imported by this function.  This makes it sound like you need to import the file (after running this func) to make the SVGs available.

tlambert03 (approved):
thanks for catching that @Czaki.  Looks good to me, with one small change requested to the docstring

napari/resources/__init__.py
Czaki: Please check new version.

PR #1088, opened 2020-03-27
Fix LayerList and ListModel type checking

tlambert03:
# Description
fixes #1086 and adds a test.
LayerList and ListModel were not inheriting the `TypedList.append` method

## Type of change

- [x] Bug-fix (non-breaking change which fixes an issue)

# How has this been tested?

- [x] new test added 
- [x] all tests pass with my change

## Final checklist:
- [x] My PR is the minimum possible work for the desired functionality
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

Issue #1086, opened 2020-03-26
ListModel.append does not check type


tlambert03:
## 🐛 Bug
in working on layer groups, I found a strange lack of type checking when appending to a `ListModel` (which inherits from `TypedList`).  [`ListModel.append`](https://github.com/napari/napari/blob/59ed366e9d492a2389c451468fd8b9f96508b4e2/napari/utils/list/_model.py#L59)  jumps right over `TypedList.append`
https://github.com/napari/napari/blob/59ed366e9d492a2389c451468fd8b9f96508b4e2/napari/utils/list/_model.py#L58-L60

... and if you try to something that is not a `Layer` to a `LayerList`, it works fine up until throwing an error (unrelated to typing) in `components.layerlist._add`.  Is that supposed to be `TypedList.append(self, obj)`?  or was that intentional?

sofroniewn:
Maybe @kne42 can comment on this one?

PR #1089, opened 2020-03-27
Layergroups

### LINKED ISSUE #970: Nesting image layers in groups for navigation

GenevieveBuckley:
# Description

This PR develops the concept of nested, hierarchical layergroups. 

## Type of change

# References

Closes https://github.com/napari/napari/issues/970

# How has this been tested?

## Final checklist:

## Try it out!

```python
import numpy as np
import skimage.data

from napari.layers import *


def small_layergroup():
    tree = Layergroup(name='small-tree')

    branch = Layergroup(name='branch')
    tree.append(branch)

    twig = Layergroup(name='twig')
    twig.append(Image(skimage.data.chelsea(), name='chelsea'))
    branch.append(twig)
    
    branch.append(Image(skimage.data.camera(), name='camera'))
    branch.append(Points(np.random.random((10, 3)) * 200, name='points'))  # 10 points in 3D, range scaled to 200 pixels
    
    tree.append(Image(skimage.data.astronaut(), name='astronaut'))
    
    return tree
```


GenevieveBuckley:
There has been a lot of discussion about this on the 'layer tree' topic on zulip, so I'll summarise the relevant parts here:

* There is consensus that layergroups should have a ordered tree-like model (nesting, hierarchical, ordered, etc.)
* @jni suggests we use the [Composite design pattern](https://en.m.wikipedia.org/wiki/Composite_pattern). There is a nice example of a python implementation here: https://refactoring.guru/design-patterns/composite/python/example
* @tlambert03 thinks we could build the GUI side using QTreeView. Links:
    * [Qt Model/View](https://doc.qt.io/qt-5/model-view-programming.html)
    * [QAbstractItemModel](https://doc.qt.io/qt-5/qabstractitemmodel.html)
    * [QTreeView](https://doc.qt.io/qt-5/qtreeview.html)
* @GenevieveBuckley has put some preliminary working notes in a gist here: https://gist.github.com/GenevieveBuckley/973a096ca70027f15bacf7c2c8643133

GenevieveBuckley:
I think it might go a little like this?
LayerGroup will eventually replace LayerList (on the napari model side). Then napari.Viewer().layers will be/hold a layer group. QLayerGroup (which we'll need to write) will replace QLayerList (on the Qt side).

Some of the next steps for layergroups will probably include:
* Use coerce_names() function to ensure unique names LayerGroup trees. Will probably need to modify the layergroup append method for this.
* ~~Make sure when adding new layers or layergroups to an existing layergroup that they respect the parent attributes state. Will probably need to modify the layergroup append method for this.~~ Edit: on second thought, I don't think this is actually necessary.
* Figure out if there's a nice way to set/apply LayerGroup attributes & methods to all LayerGroup children without duplicating large amounts of code.
* Figure out if there's a nice way to create the LayerGroup().events EmitterGroup without duplicating large amounts of code.

GenevieveBuckley:
@tlambert03 describes a problem:
> one issue I'm currently running into... once you fix the inheritance such that LayerGroup actually triggers the Layer.__init__ method, then the Layer.events EmitterGroup conflicts with the ListModel.events EmitterGroup, and appending new items to the list throws an error. we'll need to deal with that somehow. One possibility is to make LayerGroup not be a LayerList but rather have a layerlist: LayerGroup._children = LayerList()

Note from Gen: I did originally have LayerGroup as a class that contains `LayerGroup._children = LayerList()` (like the example [here](https://refactoring.guru/design-patterns/composite/python/example)). We could return to that if it makes things easier, I didn't really have a good reason for changing away from that in the first place.

tlambert03:
> We could return to that if it makes things easier, I didn't really have a good reason for changing away from that in the first place.

maybe let's try it for now, forwarding the corresponding append/remove methods as required, see what other issues it exposes, and then re-evaluate whether it can/should inherit directly later?

GenevieveBuckley:
One last note: there was a worry that the bugfix for https://github.com/napari/napari/issues/1086 in PR https://github.com/napari/napari/pull/1088 would break our ability to nest layergroups.

Fortunately, this is not a problem. So long as `LayerGroup._basetype == napari.layers.base.base.Layer` then appending layer groups works fine.

GenevieveBuckley:
> > We could return to that if it makes things easier, I didn't really have a good reason for changing away from that in the first place.
> 
> maybe let's try it for now, forwarding the corresponding append/remove methods as required, see what other issues it exposes, and then re-evaluate whether it can/should inherit directly later?

Oooh, now I remember why I did that! I wanted to have easier access to things like `len()`, indexing, appending, etc. and noticed I was often trying to work with `layergroup._children` from the outside.

I'll try changing it, see how that goes.

napari/components/layergroup.py
```diff
@@ -0,0 +1,75 @@
...
+    @property
+    def visible(self):
+        """bool: Whether the visual is currently being displayed."""
+        return self._visible
+
+    @visible.setter
+    def visible(self, visibility):
+        self._visible = visibility
+        for child in self:
+            child.visible = visibility
```
tlambert03: can you add this method for a nice `print(layer_group)` ascii tree?
```python
    def _render(self):
        """recursively return list of strings that can render ascii tree."""
        lines = []
        lines.append(self.name)

        for n, child in enumerate(self):
            try:
                child_tree = child._render()
            except AttributeError:
                child_tree = [child.name]
            lines.append('  +--' + child_tree.pop(0))
            spacer = '   ' if n == len(self) - 1 else '  |'
            lines.extend([spacer + l for l in child_tree])

        return lines

    def __str__(self):
        """Render ascii tree string representation of this layer group"""
        return "\n".join(self._render())
```

```
tree
  +--branch1
  |  +--image1_on_branch1
  |  +--image2_on_branch1
  +--branch2
  |  +--image1_on_branch2
  +--branch3
  |  +--branch4
  |  |  +--image1_on_branch4_on_branch3
  |  |  +--image2_on_branch4_on_branch3
  |  +--branch5
  |  |  +--image1_on_branch5_on_branch3
  |  +--image1_on_branch3
  +--single_img
```

napari/utils/list/_model.py
```diff
@@ -33,6 +33,7 @@ def __init__(self, basetype, iterable=(), lookup=None):
             removed=None,
             reordered=None,
             changed=None,
+            name=Event,
```
tlambert03: btw, `Event` is the default when adding a `kwarg=None` to an `EmitterGroup`, so it's not necessary to import here (I know that's inconsistent across the codebase).  It gets assigned  [here](https://github.com/napari/napari/blob/59ed366e9d492a2389c451468fd8b9f96508b4e2/napari/utils/event.py#L724-L726)

napari/components/layergroup.py
GenevieveBuckley: Much prettier!

GenevieveBuckley:
Notes from the meeting:
Attributes on layergroup nodes should *not* propagate downwards, all nodes & leaves should get/set attributes as normal. But, there needs to be a hidden "effective attribute" method that is calculated purely for display/rendering.

To do:
- [ ] Play with Talley's [`__new__` experiment](https://gist.github.com/GenevieveBuckley/973a096ca70027f15bacf7c2c8643133#file-new_for_layergroups-py)
- [ ] Try building off QTree for the Qt side of things. Ignore QtLayerList, this doesn't line up with QList for mostly historical reasons so we don't need to worry about that.

napari/components/add_layers_mixin.py
```diff
@@ -66,6 +66,26 @@ def add_layer(self, layer: layers.Layer) -> layers.Layer:
...
+        Returns
+        -------
+        layergroup : :class:`napari.layers.Layergroup` or list
+            The layergroup that was added (same as input).
+        """
+        # FIXME
+        for layer in layergroup:
+            self.add_layer(layer)
+        return layergroup
+
```
jni: Is this needed? I'm hoping that we will be able to encapsulate everything in add_layer.

GenevieveBuckley: There's a test checking for every Layer type, there is a corresponding `add_` method.
I can add an exception to that test in the case of layergroups, if you prefer.

Layergroups (as they stand now) break quite a few of these types of test.

jni: Failing tests are very much to be expected. But I guess there isn't one for Base, so however we prevent that, we should apply that to LayerGroup also.? I think?

GenevieveBuckley: Nah, the check is for consistency among all **subclasses** of the base Layer class, so by definition that won't include Layer itself. I've added a conditional that excludes Layergroup from those tests for now (we can always change this later on).

GenevieveBuckley:
I've edited my comment up top, and added example code you can use to generate a layergroup and play with the python model.

Questions:
* Do we need typed lists or something fancier than a regular list to hold the children?
* Do we want better ways of indexing into the layergroup?
* _get_value() of a layergroup currently returns a flat list containing all the get_value() results for individual layers in the layergroup. Should it be nested instead? Something other than a list?
* _get_state() of a layergroup currently returns a nested list of dictionaries, is this what we want?

Still to do:
- [ ] Add list movement methods to layergroup (right now we have only append)
- [ ] Need emittergroup on the layergroup
- [ ] Need to work out how we want to add layergroups to the viewer model
- [ ] Probably need "effective" positioning for layers, allowing layergroups to shift scale/translate/etc. On hold for now pending world coordinates discussion.

GenevieveBuckley:
This isn't merge-able yet (we need vispy and qt stuff added in as well), but I've marked it ready for review since we decided not to move forward until there was agreement on the python model.

napari/layers/base/base.py
```diff
@@ -311,6 +334,15 @@ def visible(self, visibility):
         else:
             self.editable = False
 
+    @property
+    def effective_visibility(self):
+        """bool: Whether the visual is displayed, taking layergroup
```
tlambert03: side question, out of curiosity... is this way of representing type in the docstring a standard somewhere?  I see that we have it in some other comments, but wonder whether we should try to unify on `(self) -> bool:` and/or numpy docstring types?

napari/layers/layergroup/layergroup.py
```diff
@@ -0,0 +1,183 @@
...
+
+        Returns
+        -------
+        tuple
+            Extent returned as tuple, ndim x 2 (min, max)
+        """
+        return combine_extents([c._get_extent() for c in self._children])
+
+    def _get_ndim(self):
+        return max([c._get_ndim() for c in self._children])
```
tlambert03: if there are no children this will be a `ValueError`

napari/layers/layergroup/layergroup.py
```diff
@@ -0,0 +1,183 @@
+from ..base import Layer
+from ..utils.layer_utils import combine_extents
+
+
+class Layergroup(Layer):
+    def __init__(
+        self, children=None, *, name='LayerGroup', ndim=2, visible=True
+    ) -> None:
+        if children is not None:
+            ndim = self._get_ndim()
```
tlambert03: there's an `AttributeError` here because self._get_ndim() requires self._children which is not yet defined.

But actually... having to assign ndim for a layergroup at all "smells bad" (feels like it should always be an `@property` of children).  This is more a limitation of the `Dims` model, which currently requires _something_ to be instantiated, but maybe that should be considered here too.  Alternatively ... just pass a stub of "2" to the `super().__init__()` call, and set `Layergroup.ndim` as an `@property` that returns the value of `_get_ndim` and (don't use a setter).  See also, the note below that `_get_ndim` will error if there are now children.

napari/layers/layergroup/layergroup.py
```diff
@@ -0,0 +1,183 @@
...
+        return lines
+
+    def __repr__(self):
+        """Render unambiguous tree string representation of this layer group"""
+        return "\n".join(self._render_repr())
+
+    def __iter__(self):
+        """Iteration yields every non-group layer with a depth first search."""
+        for child in self._children:
+            yield from child.__iter__()
```
tlambert03:
```suggestion
            yield from child
```

don't think the `__iter__()` is necessary there?

napari/layers/layergroup/layergroup.py
```diff
@@ -0,0 +1,183 @@
...
+        self, children=None, *, name='LayerGroup', ndim=2, visible=True
+    ) -> None:
+        if children is not None:
+            ndim = self._get_ndim()
+        super().__init__(None, ndim)
+        self._name = name
+        if children is None:
+            self._children = []
+        else:
+            self._children = children
```
tlambert03: I think this needs to use `self.append(c) for c in children` ... since we need to set the `parent` attribute on all of the children that are added to this group.  As is, i don't think any of the methods that use `_up_through_parents` (like `Layer. effective_opacity` or `Layer. effective_visibility`) are going to work if you create a LayerGroup with `Layergroup(children=[group2, point2, image1])`

```python
        super().__init__(None, ndim)
        self._children = []
        for child in (children or []):
            self.append(child)
```

napari/layers/layergroup/layergroup.py
```diff
@@ -0,0 +1,183 @@
...
+                yield child
+
+    def __getitem__(self, idx):
+        """Allows indexing into Layergroup, similar to a list of lists."""
+        return self._children.__getitem__(idx)
+
+    # We need more & better list handling methods
+    # moving a layer from one position to another
+    # moving a layer up/down heirarchy branches when moving its position
+    # removing a layer from a group (should remove the parent attribute)
```
tlambert03, line 81: agreed.  need a corresponding `remove()` method

napari/layers/layergroup/layergroup.py
```diff
@@ -0,0 +1,183 @@
...
+        state : list
+            List of layer state dictionaries.
+        """
+        state = []
+        state.append(self._get_base_state())
+        for child in self:
+            if child._children is not None:
+                state.append(child._get_state())
+            else:
+                state.append(child._get_state())
```
tlambert03: what's the conditional for?

napari/layers/layergroup/layergroup.py
```diff
@@ -0,0 +1,183 @@
...
+            for child in self._children:
+                child.refresh()
+
+    @property
+    def data(self):
+        return None
+
+    @data.setter
+    def data(self, value):
+        raise NotImplementedError  # user should not be able to overwrite this
```
tlambert03: I think you can just remove this.  By default `@properties` are not settable... and if you try to set `layer_group.data = ...` you'll get `AttributeError: can't set attribute`.

Note: I notice that the base class has the setter as an abstractmethod... but in this case, I don't think it's actually working as intended.  (That is, python doesn't complain at all about instantiating a subclass that doesn't define `data.setter` ... probably something to do with the @property)

tlambert03 (commented):
i like where this is headed!!  here are some initial thoughts

napari/layers/layergroup/layergroup.py
GenevieveBuckley: I really don't like having `ndim` on the layergroup either. But you have to have it if you want the spatial stuff (scale and extent depend on this) and a spatial description for the layergroup position is something we need, so we can eventually do grid view of grouped data. 

Maybe some of @sofroniewn's planned changes for world coordinates might untangle `ndim` from scale/extent/etc.?

tlambert03: I'm not saying remove `ndim` altogether ... just saying remove it from the `__init__`, and have it be a property that always looks at `_get_ndim` ...
```python
@property
def ndim(self):
    return self._get_ndim()
```
would that work for the spatial stuff you mentioned?

sofroniewn: Right now in #1276 I'm making `ndim` a property of the layers and layer list - see https://github.com/napari/napari/pull/1276/files#diff-9bc2d2879ff64351239220d9f2732cc5R288-R289

once that is ready for review I can bring it in line with how you're thinking about things here too

napari/layers/base/base.py
GenevieveBuckley: This type of notation is scattered around quite a few places, I copied that format directly from the `visible` property immediately above my new `effective_visibility` property. I don't think @sofroniewn will mind if this gets standardized. So, does everyone agree on what the new standard should be?

tlambert03: Type annotations! :)

napari/layers/layergroup/layergroup.py
GenevieveBuckley: I don't think that does remove the bad smell. 

What I don't understand is why the dims model doesn't always contain 3d spatial info, that we could pick and choose from depending on if the display is in 2 or 3 spatial dimensions at any given time.

napari/layers/layergroup/layergroup.py
GenevieveBuckley: > Note: I notice that the base class has the setter as an abstractmethod... but in this case, I don't think it's actually working as intended. (That is, python doesn't complain at all about instantiating a subclass that doesn't define `data.setter` ... probably something to do with the @Property)

Huh, you're right about that! I didn't try it without just the setter.

Note: I'm keeping the blending mode setter on the layergroup for now (because that does not raise NotImplementedError from the base class), but this is another code smell.

napari/layers/layergroup/layergroup.py
GenevieveBuckley: Thank you for catching this!

GenevieveBuckley:
Ping @0x00b1 - this is the layergroup proposal we discussed in the meeting, if you want to take a look

sofroniewn:
@GenevieveBuckley if i understand correctly we're going to be taking a slightly different approach here based on some of the more recent PRs from @tlambert03, is it alright if we close this one? Thanks for getting it going though!!

GenevieveBuckley:
Yes, this has been outdated for a while

PR #1092, opened 2020-03-27
All args to add_image() accept sequences when channel_axis != None

tlambert03:
# Description
as discussed in https://github.com/AllenCellModeling/napari-aicsimageio/issues/2, @JacksonMaxfield would benefit from being able to pass a list of booleans to the `visible`parameter in `add_image` for multichannel.  This PR adds that functionality, along with a test.

## Type of change

# How has this been tested?

- [x] the test suite for my feature covers adding a 5 channel image with some visible and some not.

## Final checklist:


evamaxfield:
:eyes: Huge PR thank you so much :tada:

sofroniewn (approved):
LGTM! Thanks @tlambert03 for getting this done so quickly.

@JacksonMaxfield @tlambert03 curious if there are other params you think we should make passable as a list too? Could it be all of them? We can also wait on that until another request comes up.

tlambert03:
i suppose they could all conceivably accept lists, shall we do that here?  I also just noticed that a couple arguments (`iso_threshold`, `attenuation`) aren't being passed in multichannel...

sofroniewn:
> I also just noticed that a couple arguments (iso_threshold, attenuation) aren't being passed in multichannel.

That seems like an oversight, would be good to fix.

> i suppose they could all conceivably accept lists, shall we do that here? 

I guess might as well, could have some nasty guessing then with lists of lists etc, but we already do some of that for contrast_limits.

Thoughts @jni?

jni:
Yeah the list of lists struggle is real. I'm almost thinking we should just break this into `add_images` or similar, but that would also be messy. I guess it all depends on the implementation. The Python gist on this is "in the face of ambiguity, refuse the temptation to guess", so the question is, can we avoid ambiguity? Can all arguments be unambiguously determined to be either a list to pass to the channels, or a single value to be broadcast across the channels? If so, then let's go for it, if not, then in the ones where broadcasting is not unambiguous, then we should not support a list.

tlambert03:
ok, I made the whole thing accept single values or a sequence.  It got tremendously verbose, so I ended up refactoring quite a bit, and I do like how it reads now... but this is definitely a "proposal" more than anything, so please critique.

regarding lists of lists, the main challenges were clims, scale, and translate, which are handled by the new `ensure_iterable_sequences`.  room for improvement... but let me know if you like the general pattern.

napari/components/add_layers_mixin.py
```diff
@@ -49,6 +49,7 @@ def add_layer(self, layer):
 
         if len(self.layers) == 1:
             self.reset_view()
+        return layer
```
jni, line 61: ... Interesting omission!

napari/components/add_layers_mixin.py
```diff
@@ -155,91 +159,77 @@ def add_image(
...
-                scale=scale,
-                translate=translate,
-                opacity=opacity,
-                blending=blending,
-                visible=visible,
-            )
-            self.add_layer(layer)
-            return layer
+            kwargs['colormap'] = kwargs.get('colormap', 'gray')
+            kwargs['blending'] = kwargs.get('blending', 'translucent')
```
jni: kwargs.get will return the default value when the key doesn't exist, not when it exists but the value is None. Since the key *does* exist, this is the wrong construct here. I think?? Ditto for other calls to `.get`

napari/components/add_layers_mixin.py
```diff
@@ -155,91 +159,77 @@ def add_image(
...
-            name = ensure_iterable(name)
-
-            if blending is None:
-                blending = 'additive'
-
-            if colormap is None:
-                if n_channels < 3:
-                    colormap = colormaps.MAGENTA_GREEN
+            n_channels = (data[0] if is_pyramid else data).shape[channel_axis]
+            kwargs['blending'] = kwargs.get('blending', 'additive')
```
jni: As above

napari/components/add_layers_mixin.py
```diff
@@ -155,91 +159,77 @@ def add_image(
...
+
+            # turn the kwargs dict into a mapping of {key: iterator}
+            # so that we can use {k: next(v) for k, v in kwargs.items()} below
+            for key, val in kwargs.items():
+                if key in ('scale', 'translate', 'contrast_limits'):
```
jni:
```suggestion
                if key in {'scale', 'translate', 'contrast_limits'}:
```

napari/components/add_layers_mixin.py
```diff
@@ -155,91 +159,77 @@ def add_image(
...
+                        isinstance(val, list)
+                        and len(val) == n_channels
+                        and all(isinstance(m, dict) for m in val)
+                    ):
+                        kwargs[key] = iter(val)
+                    else:
+                        raise ValueError(
+                            "metadata must either be a dict, or a list of dicts "
+                            "with a length equal to the number of channels."
+                        )
```
jni: Why is this not raised with the other maybe-sequences?

napari/components/add_layers_mixin.py
```diff
@@ -155,91 +159,77 @@ def add_image(
...
+            # turn the kwargs dict into a mapping of {key: iterator}
+            # so that we can use {k: next(v) for k, v in kwargs.items()} below
+            for key, val in kwargs.items():
+                if key in ('scale', 'translate', 'contrast_limits'):
+                    kwargs[key] = iter(
+                        ensure_iterable_sequences(val, n_channels)
+                    )
+                elif key == 'metadata':
+                    if val is None or isinstance(val, dict):
+                        kwargs[key] = itertools.repeat(metadata)
```
jni: Can ensure iterable not be made to work for dicts?

napari/components/add_layers_mixin.py
jni: Should a test be added for this?

jni (commented):
@tlambert03 I really like the approach actually! Tests aren't passing, but I am very happy with the direction.

napari/components/add_layers_mixin.py
tlambert03: it's funny... I recently made your exact same point to someone in an external PR, but somehow forgot it here!  I meant to do `kwargs['colormap'] or 'gray'`, that ok?  The only difference from the current code is that it also excludes the empty string, which in this context is as useless as `None`

napari/components/add_layers_mixin.py
tlambert03: changed `ensure_sequence_of_iterables` to handle dicts too.

tlambert03:
looks like the failing test was in `test_signature` where we were checking that the `add_image` method _literally_ had code that said `layer = layers.Image(<all of the arguments typed out>)`.  Since that test is definitely incompatible with this approach, (and since it would make for a terribly ugly method otherwise), I [modified](https://github.com/napari/napari/blob/783ac96c450af1d1011b0fd5d9821b2caba30438/napari/_tests/test_viewer_layer_parity.py#L130-L139) the signature test for `add_image`. Would be great if @kne42 has a moment to take a look at that.

napari/components/add_layers_mixin.py
```diff
@@ -155,91 +159,65 @@ def add_image(
...
+
+            # turn the kwargs dict into a mapping of {key: iterator}
+            # so that we can use {k: next(v) for k, v in kwargs.items()} below
+            for key, val in kwargs.items():
+                if key == 'colormap' and val is None:
+                    if n_channels < 3:
+                        kwargs[key] = iter(colormaps.MAGENTA_GREEN)
+                    else:
+                        kwargs[key] = itertools.cycle(colormaps.CYMRGB)
+                elif key in {
```
sofroniewn: Can you add a comment on why these ones get special treatment (I guess because they themselves can be iterable)

napari/components/add_layers_mixin.py
```diff
@@ -155,91 +159,65 @@ def add_image(
         elif data is None:
             data = io.magic_imread(path)
 
+        # doing this here for IDE/console autocompletion in add_image function.
```
sofroniewn, line 190: Just curious, how does this relate to IDE autocomplete? It looks to me like this is useful in the for loop below

napari/components/add_layers_mixin.py
```diff
@@ -123,8 +124,9 @@ def add_image(
...
         attenuation : float
             Attenuation rate for attenuated maximum intensity projection.
-        name : str
-            Name of the layer.
+        name : str or list of str
+            Name of the layer.  If a list then must be same length as the axis
+            that is being expanded as channels.
         metadata : dict
             Layer metadata.
         scale : tuple of float
```
sofroniewn: do we need to add more "or list of ..." to these doc strings?

napari/_tests/test_viewer_layer_parity.py
```diff
@@ -127,9 +127,20 @@ def test_signature(layer):
 
     code = inspect.getsource(method)
 
-    args = re.search(rf'layer = layers\.{name}\((.+?)\)', code, flags=re.S)
-    # get the arguments & normalize whitepsace
-    args = ' '.join(args.group(1).split())
+    if name == 'Image':
```
sofroniewn, line 134: I have to say I don't really understand this, or what the original code was doing. Can you maybe add a comment above the if statement about what the whole block is trying to achieve and then maybe make it more clear what is different with our current approach. It doesn't look to me like the actual function signature is changing (in terms of names of things etc.)

sofroniewn (commented):
@tlambert03 this is looking good, I left a couple comments, mostly minor and asking for more comments / explanations around a couple things that weren't so clear to me

napari/components/add_layers_mixin.py
tlambert03: that's just there for a newcomer that might look at that and say "why didn't you just put `**kwargs` in the function signature and remove all this extra code?  maybe it's either unnecessary or could be clearer

napari/components/add_layers_mixin.py
tlambert03: will do.  yeah, those are the 4 args that are already iterable even in the single channel case.

napari/components/add_layers_mixin.py
tlambert03: whoops.  def.

napari/_tests/test_viewer_layer_parity.py
tlambert03: will add comment.  this test was using `code = inspect.getsource(method)` to grab the source code of the (e.g.) `add_image` method and asserts that somewhere in the method, there _literally_ appears the line `layer = layers.Image(data, rgb=rgb, is_pyramid=is_pyramid, ...)` with every parameter typed out as it is in the corresponding `Layer.__init__` method.  That doesn't work with this approach, which uses `layers.Image(image, **i_kwargs)` to prevent needing to zip up and then unzip 16 different variables.

I think it'd eventually be nicer to eventually rewrite this such that the behavior is tested, rather than a specific implementation, but that's tough to do.

napari/components/add_layers_mixin.py
sofroniewn: oooh, i get it, lol. that's fine

napari/components/add_layers_mixin.py
tlambert03: I'm not sure,  it wasn't exactly an error that it didn't previously return a layer.  (since you already had the layer if you were calling this method in the first place)... And we didn't really need it to return one until I wanted it below in this PR :)

sofroniewn: I think no need

tlambert03:
just updated the multichannel tests to use `mark.parametrize` to make sure we're hitting pretty much every case.  @jni, if you can take one more look, that'd be great.

napari/components/add_layers_mixin.py
```diff
@@ -155,91 +187,76 @@ def add_image(
...
-
-            name = ensure_iterable(name)
-
-            if blending is None:
-                blending = 'additive'
-
-            if colormap is None:
-                if n_channels < 3:
-                    colormap = colormaps.MAGENTA_GREEN
+            n_channels = (data[0] if is_pyramid else data).shape[channel_axis]
```
jni, line 227: Won't this fail if we pass in an implicit pyramid? ie pass in a list of arrays and `is_pyramid=None`?

tlambert03: lemme see if I can write a test for that

will-moore:
Hi,
 This seems to break my usage of ```colormap=("from_omero", cmap),``` when NOT using channel_axis. The docs say:

"If a tuple the first value must be a string to assign as a name to a colormap and the second item must be a Colormap".

Thanks,
  Will.

tlambert03:
ah indeed!  sorry we missed that one as a multi-type argument.  I'll add a PR and a test

tlambert03:
just to check, are you getting `TypeError("Received sequence ... `

will-moore:
Yes:
``` 
File "/Users/wmoore/Desktop/NAPARI/napari/napari/components/add_layers_mixin.py", line 212, in add_image
    f"Received sequence for argument '{k}', "
TypeError: Received sequence for argument 'colormap', did you mean to specify a 'channel_axis'? 
```
Thanks!

PR #1094, opened 2020-03-28
Added documentation for label painting

### LINKED ISSUE #1041: Label painting modifies the input array in-place
## 📚 Documentation

It was brought to my attention by @VolkerH that not everyone will expect that painting/editing labels will modify the input array in-place. It might be worth documenting this fact and also suggesting that an array copy should be made if this behaviour is not desired (e.g. `viewer.add_labels(segmentation.copy())`).

BhavyaC16:
# Description
Added clarification that `add_labels` modifies the input array.
Provided notes at the end of docstring, suggesting that a copy can be made to avoid modification of the input array.
Fixes #1041 

## Type of change
- [x] Documentation

### DIFF
--- a/napari/components/add_layers_mixin.py
+++ b/napari/components/add_layers_mixin.py
@@ -394,6 +394,8 @@ class AddLayersMixin:
         An image-like layer where every pixel contains an integer ID
         corresponding to the region it belongs to.
 
+        Modifies the input array in-place.
+
         Parameters
         ----------
         data : array or list of array
@@ -432,6 +434,14 @@ class AddLayersMixin:
         -------
         layer : :class:`napari.layers.Labels`
             The newly-created labels layer.
+
+        Notes
+        -----
+        Pass deep copy of the input array as a parameter to avoid modification of input data array.
+
+        For example, call add_labels as follows::
+            viewer.add_labels(data.copy())
+        where data is the input array.
         """
         if data is None and path is None:
             raise ValueError("One of either data or path must be provided")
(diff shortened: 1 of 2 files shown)


sofroniewn (approved):
This is great @BhavyaC16, thanks for your first contribution! It looks ready to go for me, but I'll let @jni take a quick look before merge too

napari/components/add_layers_mixin.py
```diff
@@ -394,6 +394,8 @@ def add_labels(
         An image-like layer where every pixel contains an integer ID
         corresponding to the region it belongs to.
 
+        Modifies the input array in-place.
```
jni: This isn't necessarily true, if one is only looking at the layer. I would say "Using the viewer's label editing tools (painting, erasing) will modify the input-array in-place. To avoid this, pass a copy as follows:"

You might also specify that you can *get* the painted labels back with:

```python
layer = viewer.add_labels(data.copy())
# do some painting/editing
result = layer.data`. 
```

jni (changes requested):
@BhavyaC16 thank you for taking this on! I've made a suggestion to improve the text a little. I also think that the "modifies" and "how to avoid it" notes should be together. I think I prefer them at the top of the docstring rather than in a Notes section, because it is very important that users see it.

Additionally, could you please make the corresponding changes to `napari.view_labels` as well?

BhavyaC16:
Hi, thanks a lot for the feedback! I have made the suggested changes and request you to please review them.

jni (approved):

jni:
Great! @sofroniewn can you have another look over? The two docstrings are slightly different, but for good reasons, so I like where it is now.

PR #1095, opened 2020-03-31
Roadmap for 0.3 releases

sofroniewn:
# Description
This PR adds a roadmap for our upcoming 0.3 series of releases.

The goal is to capture current development priorities within the project, and provide clear information to users who are interested in seeing where napari is going, contributors who would like help out, the core team to ensure that we're staying on track, and potential plugin developers who'd like to build on top of napari. The document itself explains this out in more detail, so I won't comment here more.

I want to add thanks in particular to @jni who helped a lot with this first draft and others who weighed in too, @tlambert03 @royerloic @justinelarsen. It would be great if all the @napari/core-devs could now go through and give comments / suggestions / improvements.

@tlambert03 I also updated the statement around `plugins` in the mission and values doc, can you look at that bit in particular


docs/developers/ROADMAP_0_3.md
```diff
@@ -0,0 +1,76 @@
...
+
+- Make the **data viewing and annotation** capabilities **bug-free and fast**.
+
+- Make a **downloadable application** with reader / writer plugin management.
+
+- Make accessible documentation, tutorials, and demos.
+
+Once the above goals are met, we will develop napari's capabilities for image processing and analysis. We are prioritizing the robustness and polish of the core viewer before adding advanced features or support for functional or interactive plugins to ensure that plugin development will happen against a solid foundation, and because we have repeatedly encountered biologists and other scientists who have trouble even looking at their data, and annotating it. We are prioritizing the downloadable application with reader / writer plugin management so that we can start getting feedback from non-coding users to better understand their needs.
+
+### Bug-free and fast data viewing and annotation
```
tlambert03: heading levels should preferably only increment by one level at a time (this is sometimes problematic in sphinx)

docs/developers/ROADMAP_0_3.md
```diff
@@ -0,0 +1,76 @@
...
+
+- Add a unified **world coordinate system**. Scientists need to measure data that comes from a real space with physical dimensions. Currently, napari has no concept of the space in which data lives: everything is unitless. Further, it is unclear at various parts in the UI whether a coordinate has been transformed. And finally, some data are acquired with distortions, such as skew in lattice lightsheet data, and napari should be able to account for those distortions by chaining together transforms - including affine and ultimately deformable transforms. We are tracking progress in this area in the [World Coordinates project board](https://github.com/napari/napari/projects/10).
```
tlambert03: rather than "skew in lattice lightsheet data" i would say "skew in data collected on stage-scanning lightsheet microscopes".  There are a lot of scopes like that.

docs/developers/ROADMAP_0_3.md
```diff
@@ -0,0 +1,76 @@
...
+
+### Accessible documentation, tutorials, and demos
+
+- Improve our website [napari.org](https://napari.org) to provide easy access to all napari related materials, including the [**four types of documentation**](https://www.divio.com/blog/documentation/): learning-oriented tutorials, goal-oriented how-to-guides or galleries, understanding-oriented explanations (including developer resources), and a comprehensive API reference. See [#764](https://github.com/napari/napari/issues/764) and the [`documentation` label](https://github.com/napari/napari/labels/documentation) on the napari repository for more details.
+
+- Add a **napari human interface guide** for plugin developers, akin to [Apple's Human Interface Guidelines](https://developer.apple.com/design/human-interface-guidelines/). We want such a guide to promote best practices and ensure that plugins provide a consistent user experience.
```
tlambert03, line 47: 💚

tlambert03 (approved):
looking great!

justinelarsen (approved):
Looks great!!

docs/developers/ROADMAP_0_3.md
```diff
@@ -0,0 +1,76 @@
...
+
+
+### Other improvements
```
justinelarsen: I would eliminate this section as it just seems like leftover items and doesn't fit into the three goals above.
You could change the first goal to "Make the data viewing and annotation capabilities bug-free, fast _**and easy to use**_**." and then we can add the "Improve the **user interface and design** of the viewer." in there.

"Improve our unit testing" could also go into the first goal because of "bug-free".

"layer groups" is the odd one out, but you could argue that it could then fall under "easy to use".

docs/developers/ROADMAP_0_3.md
```diff
@@ -0,0 +1,76 @@
+# Roadmap for 0.3 series of releases - April 2020.
+
+The napari roadmap captures current development priorities within the project and should serve as a guide for napari core developers, to encourage and inspire contributors, and to provide insights to external developers who are interested in building for the napari ecosystem. For more details on what this document is and is not, see the [about this document section](#about-this-document).
+
+The [mission](MISSION_AND_VALUES.md#our-mission) of napari is to be a foundational multi-dimensional image viewer for Python and provide graphical user interface (GUI) access to image analysis tools from the Scientific Python ecosystem for scientists across domains and levels of coding experience. To work towards this mission, we have set the following three high-level priorities to guide us over the upcoming months:
+
+- Make the **data viewing and annotation** capabilities **bug-free and fast**.
+
```
justinelarsen, line 8: I didn't see headless mode or serialization items in these goals below.

sofroniewn: good catches, they should be captured somewhere. curious where @jni thinks they should go?

docs/developers/ROADMAP_0_3.md
sofroniewn: I agree, I like those suggestions. @jni?

jni: So, Nick and I debated about this for some time. To be honest, I would favour dropping CI, serialization, and headless. The other two, focused on user friendliness, I'm happy to move along with a relabel of the top-level goal. (Though together they might merit their own subsection.)

My perspective is that the document is supposed to outline priorities and focus effort. Its purpose becomes a bit more muddled if *everything* is a priority — which it's not too far off from already!

But, Nick wanted to include these items so we don't lose track of them. Fine. But then we should have some sort of indication that they are less critical/urgent/should come after the others. Hence the "other improvements".

Additionally, another idea I had for the document is that it should focus on the *user experience*. So, words like *refactor* and *software architecture* should be neatly buried, not front and centre — they should only serve the other, user-facing goals. I think "improve CI" falls in that category — most users don't even know what CI is.

Now, in terms of the items discussed:

- CI: not user facing at all. This is a problem we face as developers and we should expend effort getting it right, but I don't think it needs to be on the roadmap. As developers, we directly feel the pain, and we are (sometimes) motivated to go fix it in the course of pursuing the other goals.
- layer groups: I've often come across the issue of wanting to adjust settings simultaneously for many layers, so I think this can take priority from a user-friendliness perspective.
- ditto for the UI audit, though perhaps it should wait until everything else in this roadmap, so maybe it can be jettisoned for now.
- serialization: I think 90%+ of use cases involved data writing, and that should be the focus, rather than things that are napari-specific. I think the use case of "I want to write my data to a portable format for further analysis, publication, archival, or to use in different software" is much more common than "I want to share this entire viewer including state with someone else". Indeed, with the former, the latter becomes quite a bit easier ("here's the data, can you open it with this script and go to slice 47?"), and everything else is convenience rather than necessity.
- headless: well, I'm still behind my mantra of "napari headless is just Python", though I understand that connecting to a remote napari is interesting. But, we have accessing remote data in the roadmap already. There's lots of options for that and I don't see that headless should get its own heading (heh) here.

Anyway, this is just my loosely-held perspective. We can discuss more during the meeting! Happy to be overruled if others feel differently.

sofroniewn:
We should add #834 and persistent settings to the roadmap as i see it as a basic feature, especially for the bundled app - you don't want to have to re-order your plugins every time you launch napari

docs/developers/ROADMAP_0_3.md
sofroniewn: done

docs/developers/ROADMAP_0_3.md
sofroniewn: done

docs/developers/ROADMAP_0_3.md
sofroniewn: This additional section has been eliminated - UI and layer groups are now framed under making the viewer easier to use, serialization doesn't get mentioned, headless doesn't get mentioned, and CI has been moved under the eliminate all bugs.

I think this achieves the right balance between being user facing and motivated, but also letting developers know what is planned and prioritized

sofroniewn: This additional section has been eliminated - UI and layer groups are now framed under making the viewer easier to use, viewer serialization appears in "work for future roadmaps", headless only gets mentioned in the "viewing big datasets" , and CI has been moved under "eliminate all bugs".
Remote jupyter notebooks also now get mentioned in "work for future roadmaps"

I think this achieves the right balance between being user facing and motivated, but also letting developers know what is planned and prioritized and what will come later

docs/developers/ROADMAP_0_3.md
sofroniewn: addressed in comment above

sofroniewn:
I have made some changes reflecting the most recent comments. @jni can you re-review

jni (approved):
🚀

docs/developers/ROADMAP_0_3.md
```diff
@@ -0,0 +1,75 @@
...
+- Ensure **easy installation with a success rate close to 100%**. We can still struggle with installation of Qt in some cases, and have had problems with dependency conflicts with other Python packages. We have recently added a [conda-forge package](https://github.com/conda-forge/napari-feedstock) and are working towards distributing bundled applications in [#496](https://github.com/napari/napari/pull/496). See the [`installation` label](https://github.com/napari/napari/labels/installation) for a current list of issues related to installation.
```
kevinyamauchi, line 23: > easy installation with a success rate close to 100%

I think this is a great goal and I am curious how we would measure the success rate. I think as we expand more to users who are closer to "non-coders", they may be less likely to raise issues on github. Anyway, this is more of a curiosity than a critique.

docs/developers/ROADMAP_0_3.md
```diff
@@ -0,0 +1,75 @@
...
+
+- **Improve support for annotation** of points [#858](https://github.com/napari/napari/issues/858), shapes [#177](https://github.com/napari/napari/issues/177), labels, and images [#209](https://github.com/napari/napari/issues/209) including text rendering [#600](https://github.com/napari/napari/pull/600).
+
+- Add **linked multi-canvas support** ([#760](https://github.com/napari/napari/issues/760)) to allow orthogonal views, or linked views of two datasets, for example to select keypoints for image alignment, or simultaneous 2D slices with 3D rendering.
```
kevinyamauchi, line 29: I am psyched for this!

kevinyamauchi (approved):
This looks great. Thanks, everyone! I am stoked for 0.3!

docs/developers/ROADMAP_0_3.md
sofroniewn: As commented in meeting - i think for the GitHub, seeing the rate at which we use the installation label and bug label combined go to zero would be good. For non-GitHub users, pointing them towards image.sc and the napari tag might be the way to go

sofroniewn:
I'm merging now! Thanks for feedback all, excited to get this in!!

PR #1096, opened 2020-04-01
Add installed plugins to sys_info()

tlambert03:
# Description
This PR adds a list of currently installed plugins (and versions) to `napari.utils.info.sys_info()` so that they appear in the "napari info" dialog... and when using `napari --info` on the command line.  Looks like this:
```
napari: 0.2.12+35.g5924b456.dirty
Platform: macOS-10.13.6-x86_64-i386-64bit
Python: 3.8.2 | packaged by conda-forge | (default, Mar  5 2020, 16:54:44)  [Clang 9.0.1 ]
Qt: 5.14.1
PySide2: 5.14.1
NumPy: 1.18.1
SciPy: 1.4.1
scikit-image: 0.16.2
Dask: 2.10.1
VisPy: 0.6.5.dev83+g8c52a317

GL version:  2.1 NVIDIA-10.33.0 387.10.10.15.15.108
MAX_TEXTURE_SIZE: 16384

Plugins:
  - aicsimageio: 0.1.0
  - aicsimageio_delayed: 0.1.0
  - dv_reader: 0.2.3
```

## Type of change

# How has this been tested?
- [x] added line to test

## Final checklist:

### DIFF
--- a/napari/_qt/qt_main_window.py
+++ b/napari/_qt/qt_main_window.py
@@ -226,9 +226,16 @@ class Window:
         title = QLabel("Installed Plugins")
         title.setObjectName("h2")
         layout.addWidget(title)
-        table = QtDictTable(
+        # get metadata for successfully registered plugins
+        data = [
+            v
+            for k, v in plugin_manager._plugin_meta.items()
+            if k in plugin_manager._name2plugin
+        ]
+        # create a table for it
+        dialog.table = QtDictTable(
             self._qt_window,
-            list(plugin_manager._plugin_meta.values()),
+            data,
             headers=[
                 'plugin',
                 'package',
@@ -239,11 +246,13 @@ class Window:
             ],
             min_section_width=60,
         )
-        table.horizontalHeader().setObjectName('pluginTableHeader')
-        table.verticalHeader().setObjectName('pluginTableHeader')
-        table.setGridStyle(Qt.NoPen)
-        layout.addWidget(table)
+        dialog.table.horizontalHeader().setObjectName('pluginTableHeader')
+        dialog.table.verticalHeader().setObjectName('pluginTableHeader')
+        dialog.table.setGridStyle(Qt.NoPen)
+        layout.addWidget(dialog.table)
         dialog.setLayout(layout)
+        dialog.setAttribute(Qt.WA_DeleteOnClose)
+        self._plugin_list = dialog
         dialog.exec_()
 
     def _show_plugin_sorter(self):
--- a/napari/_qt/qt_main_window.py
+++ b/napari/_qt/qt_main_window.py
@@ -9,7 +9,7 @@ from skimage.io import imsave
 
 from .qt_about import QtAbout
 from .qt_plugin_report import QtPluginErrReporter
-from .qt_plugin_list import QtPluginSorter
+from .qt_plugin_sorter import QtPluginSorter
 from .qt_viewer_dock_widget import QtViewerDockWidget
 from ..resources import get_stylesheet
 
--- a/napari/_qt/qt_main_window.py
+++ b/napari/_qt/qt_main_window.py
@@ -228,7 +228,7 @@ class Window:
         layout.addWidget(title)
         table = QtDictTable(
             self._qt_window,
-            plugin_manager._plugin_meta.values(),
+            list(plugin_manager._plugin_meta.values()),
             headers=[
                 'plugin',
                 'package',
--- a/napari/_qt/qt_main_window.py
+++ b/napari/_qt/qt_main_window.py
@@ -224,6 +224,7 @@ class Window:
         dialog.setMaximumHeight(800)
         dialog.setMaximumWidth(1280)
         layout = QVBoxLayout()
+        # maybe someday add a search bar here?
         title = QLabel("Installed Plugins")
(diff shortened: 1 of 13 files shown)


sofroniewn:
This is great, but looking at things now I wonder if we shouldn't also have a separate menu bar line in the plugins tab that is just "show plugins". We can keep the info to have plugin info too, but I think the top line in the plugins tab should be "show plugins".

We could also add `napari --plugins` to the CLI to list the available plugins.

I also think we should take the citation panel out of the "info" box and just have a menu bar item in the help tab that called "citation" that pops that box up along. Thoughts?

tlambert03:
> I wonder if we shouldn't also have a separate menu bar line in the plugins tab that is just "show plugins". We can keep the info to have plugin info too, but I think the top line in the plugins tab should be "show plugins".

Sure, I can add that.  I definitely think it should be included in the regular `--info` though as well... since it’s relevant when we’re trying to help debug an issue for someone.


> We could also add `napari --plugins` to the CLI to list the available plugins.

By available do you mean “installed”, or “installable” from pip?  If the former, I kinda think that `--info` is sufficient for the CLI (not sure we want to bloat the CLI and I do think it should also be in `--info`), if the latter, let’s do that in another pr.

> I also think we should take the citation panel out of the "info" box and just have a menu bar item in the help tab that called "citation" that pops that box up along. Thoughts?

That’s fine with me, don’t feel strongly either way

sofroniewn:
> Sure, I can add that. I definitely think it should be included in the regular --info though as well... since it’s relevant when we’re trying to help debug an issue for someone.

Yup, let's have both from drop downs, I think they have different purposes

> By available do you mean “installed”, or “installable” from pip? If the former, I kinda think that --info is sufficient for the CLI (not sure we want to bloat the CLI and I do think it should also be in --info), if the latter, let’s do that in another pr.

I had meant installed, but agreed info is probably sufficient there. Available sounds really cool, but let's wait on that for a future PR and add to CLI and menubar at same time.

> I also think we should take the citation panel out of the "info" box and just have a menu bar item in the help tab that called "citation" that pops that box up along. Thoughts?
>
> >That’s fine with me, don’t feel strongly either way

Great, do you just want to that that in this PR or wait for a new one?

tlambert03:
OK @sofroniewn, I added a "List installed plugins..." menu item, and it brings up this table:

Screen Shot 2020-04-03 at 12 17 38 PM

This PR ended up touching a lot of other files because I wanted to improve the way package metadata was collected and stored at the moment of plugin discovery.  So the main changes are a new return type from `manager. iter_plugin_modules` ... which needed to be accomodated elsewhere, a new `PluginManager._plugin_meta` dict that stores info... and a `QtDictTable` class that builds a table from a list of dicts (used when showing the plugin info).  Sorry it grew in size... but hopefully it's for the better.

napari/_qt/qt_main_window.py
```diff
@@ -204,6 +215,46 @@ def _add_plugins_menu(self):
         report_plugin_action.triggered.connect(self._show_plugin_err_reporter)
         self.plugins_menu.addAction(report_plugin_action)
 
+    def _show_plugin_list(self):
```
sofroniewn, line 219: Not sure where to put this comment, but the grid text is editable, which I don't think it should be 
image

sofroniewn: Can you also make the url underlined so it is more clear that it is a link?

sofroniewn: Also do you see how in the top left there is a bright white vertical line? Can you make that the same as the others? 
image

napari/_qt/qt_main_window.py
```diff
@@ -204,6 +215,46 @@ def _add_plugins_menu(self):
...
+        """Show dialog with a table of installed plugins and metadata."""
+        from ..plugins import plugin_manager
+
+        dialog = QDialog(self._qt_window)
+        dialog.setMaximumHeight(800)
+        dialog.setMaximumWidth(1280)
+        layout = QVBoxLayout()
+        title = QLabel("Installed Plugins")
```
sofroniewn, line 228: Maybe a comment for the future - could we add a search bar on the top? Does Qt have built in "search" functionality that we can leverage? Or would we need to build ourselves? Not a priority now, but curious

sofroniewn (commented):
Looking great @tlambert03 I left some comments, main one is about text being editable, but its close

tlambert03:
thanks @sofroniewn, styling tables is a real PITA!  i managed to fix all those items though, and made it uneditable.  Not immediately sure about the searchbar, but it shouldn't be terribly hard to implement once the need arises.

jni:
> styling tables is a real PITA!

Well, just fyi, I think you are doing an amazing job of it @tlambert03! My immediate reaction when seeing @sofroniewn's screenshots was 😍 "Damn, Talley's got *style!*"

image

(And don't worry, we like you! The gif is just about the second part of that sentence! 😂)

tlambert03:
lol

sofroniewn (approved):
Looks great, just tried it again. Thanks for the fixes, we can think about the search bar or any tweaks to this after people start using it more. Nice job.

jni (approved):

PR #1097, opened 2020-04-01
Type annotations on Layers

tlambert03:
# Description
As I started to play around with LayerGroups in #1089 a little more, I found myself looking through layer methods trying to remember exactly what they accept and return, and really wishing we had type annotations on all of our layer methods.   So this PR will add type hints to all the layers (and maybe update some of the docstrings slightly too).  I will refrain from touching code as much as possible, unless it would be made much cleaner by changing a type (e.g. from a list to a tuple).   The image layer is done, so have a look at that if you want to weigh in on whether you're ok with this.

Another benefit here, if you're using an IDE that supports something like mypy, is that you can catch a lot more potential bugs in the code if it's annotated.

## Type of change

- [x] Refactor

# How has this been tested?

## Final checklist:


sofroniewn (commented):
@tlambert03 this is looking great, I'm no expert here so never used Type annotations for code I contributed, but I think @kevinyamauchi among others have. I think standardizing across all our layers (and eventually the whole repo makes a lot of sense). Curious what @kevinyamauchi @jni @kne42 think too.

We might also want to add a section about Type annotations to our contributing guide too - what they are, why we use them etc. to make it clear what our policy is and what reasons we have for them

GenevieveBuckley:
> We might also want to add a section about Type annotations to our contributing guide too - what they are, why we use them etc. to make it clear what our policy is and what reasons we have for them

This would be an excellent thing to include there, I agree

jni:
Well, my overarching feeling about this is that we should hold off on this until we get more of a grip on array types/image types. I think your frustration could have been relieved with NumPy-style docstrings, right @tlambert03?

Don't get me wrong, I'm all for type annotations in general and I'm looking forward to resolving this, I just feel that doing this now means double the effort, once we separate e.g. color arrays from numpy arrays from any-arrays from coordinates from images from...

Thoughts?

A good item for discussion at the meeting later. =)

tlambert03:
> Well, my overarching feeling about this is that we should hold off on this until we get more of a grip on array types/image types. I think your frustration could have been relieved with NumPy-style docstrings, right @tlambert03?
> Thoughts?

This was my first thought, but actually my main goal here is not really to define the types of _data_ underlying the layer (_à la_ `imtypes`), but rather the methods and API that layers and subclasses should adhere to.  When thinking about how we could have `LayerGroups` propagate their attributes to children without a lot of code duplication I had the following observations/thoughts:

- some of the methods are returning different types on different layers without any obvious reason.  With `_get_extent()`, `Points` return a `list` and all the others return a `tuple`; `Image` returns 2-tuples `((min, max), ...)`, and all the others return 3-tuples `((min, max, 1), ...)`
- yes, numpy style docstrings everywhere would definitely help, but they are, in my opinion, far less useful than type annotations that my IDE can parse allowing my linter to highlight potential problem areas.
- some of the docstrings are outdated, (and a linter would catch that with type annotations but not docstrings), or missing altogether.  so this effort would be a little like what @GenevieveBuckley did for our Qt classes...

Does that make sense?  Main point being: I'm not trying to nail down sub-types of arrays and data types, but rather all the _other_ stuff that will make layer groups easier.

jni:
Yes, it makes sense, but I still see a lot of ArrayLikes sprinkled around...

btw, why did we decide not to use the Protocol idea in the end?

Anyway, I'm not opposed to the idea! Let's discuss at meeting. =)

tlambert03:
> I still see a lot of ArrayLikes sprinkled around...

I guess my strategy was to narrow it down as much as possible... but no more.  So if we _know_ it's going to be an arraylike, we use the (rather unsatisfying) current `ArrayLike` from `napari.types`.  But we avoid clarification of stuff like color arrays, any arrays, coordinate arrays, etc... until we have a better way for those.

> why did we decide not to use the Protocol idea in the end?

I wouldn't say we decided that.  I still think it's the most promising (if perhaps still incomplete) strategy.  If you want to do it in napari though, we'd need to add that backport dependency for python 3.6 & 3.7.

GenevieveBuckley:
@tlambert03 I'm going to open an issue to track fixes for the inconsistencies you list here, since those changes probably shouldn't get mixed in with this PR

> some of the methods are returning different types on different layers without any obvious reason.  With `_get_extent()`, `Points` return a `list` and all the others return a `tuple`; `Image` returns 2-tuples `((min, max), ...)`, and all the others return 3-tuples `((min, max, 1), ...)`

tlambert03:
I finished up typing all of the attributes and methods on the layers (again, with the exception of `Layer.data`, which is important enough and difficult enough to leave to another PR).  I'm going to take this out of draft and get some feedback.

I wrote up a [little blurb in `contributing.md`](https://github.com/napari/napari/blob/8a5dc804f280ba728a27a2182eaf06930c967d0d/docs/developers/CONTRIBUTING.md#type-hints-in-napari) that outlines why I think these are valuable, and why I'd like to have them in master.  (for me, it's mostly about the IDE autocomplete and linting)

napari/layers/image/image.py
```diff
@@ -249,93 +253,96 @@ def __init__(
...
+        if self.multiscale:
+            return self.data[0].dtype
+        else:
+            return cast(ArrayLike, self.data).dtype
 
     @property
-    def data(self):
-        """array: Image data."""
+    def data(self) -> Union[ArrayLike, List[ArrayLike]]:
+        """Image data. Will be a list of arrays if self.multiscale"""
```
GenevieveBuckley:
```suggestion
        """Image data. Will be a list of arrays if self.multiscale is True."""
```

This maybe looks like the sentence has been accidentally truncated, this might help make it clear that's not the case.

napari/layers/image/image.py
```diff
@@ -249,93 +253,96 @@ def __init__(
...
         self._update_dims()
         self.events.data()
 
-    def _get_ndim(self):
+    def _get_ndim(self) -> int:
         """Determine number of dimensions of the layer."""
         return len(self.level_shapes[0])
 
-    def _get_extent(self):
+    def _get_extent(self) -> Tuple[Tuple[int, int], ...]:
```
GenevieveBuckley, line 285: The return type of `_get_extent()` is still inconsistent between different layer types. Is this intentional?

* Image `-> Tuple[Tuple[int, int], ...]`
* Labels - inherits `_get_extent()` from Image.
* Points `-> List[Tuple[int, int], ...]`
* Shapes - ?
* Surface `-> Tuple[Tuple[int, int], ...]`
* Vectors `-> List[Tuple[int, int], ...]`

tlambert03: not intentional, thanks for pointing it out!
So far with this PR, I've mostly tried _not_ to actually alter the types of anything... just to annotate them as they are.  But I also think it would be within scope to unify types like that (and this is a perfect example), agree?

(also, looks like I haven't done shapes yet!  thanks)

GenevieveBuckley: > So far with this PR, I've mostly tried _not_ to actually alter the types of anything... just to annotate them as they are. But I also think it would be within scope to unify types like that (and this is a perfect example), agree?

It certainly seems in scope, yes. Although if you do decide to keep them separate, it'd be good to open a tracking issue for it.

PR #1102, opened 2020-04-02
changing translate order when changing data dims

vigji:
# Description
I was building a little widget to change online the `translate` attribute of a layer and I noticed that the moment I was rolling dimension the translation was wrong until a new `translate` value was set. This happens b/c to my understanding in none of the calls triggered by the dimension roll button the translate values order is changed. To fixed that, I added a call to `_on_translate_change()` method of the `VispyImageLayer` inside the `on_data_change()` method. Not 100% this is the optimal place for this to happen - still not familiar with the entire program structure - but for my usage this fix the problem.

## Type of change

# How has this been tested?
- My bug has been fixed;
- all tests pass with my change (apart from an unrelated dask array test)

## Final checklist:


sofroniewn:
@vigji Thanks for your contribution!

This makes sense to me, but could you provide a minimal example of how to create the problem, I imagine it looks something like this
```python
viewer = napari.Viewer()
# Create 3D image layer
layer = viewer.add_image(np.random.random((30, 60, 70)))
# Set translate
layer.translate = [10, 50, 100]
# Roll dims
layer.dims.order = [2, 0, 1]
# Now problems has occurred as changing the order changed the underlaying display data in vispy, but it didn't change update the underlying translate property
```
But I havn't tested that.

If this is the case then your fix might be the best one, though I want to have a bit more of a think about it - we might also need to call the scale change method too for example, so there might be a better way to think about this.

Also once we have this minimal example we'll want to turn it into a unit test in this PR too.

vigji:
@sofroniewn thanks for the quick feedback! It's my first PR and I'll have to learn :) 
Your code actually seems to unravel another funny glitch: it transpose the image in the current view but the original axes order is reset when start scrolling the stack. I will look into this as well, and maybe I can also test the scale.
And sure, when everything's clearer I will try to wrap it in a test as well!

vigji:
Minimal example of how to reproduce the bug, for both scale and translation. (t occurred to me that the same glitch happens also with image transposing):

```
import napari
import numpy as np

with napari.gui_qt():
    viewer = napari.Viewer()

    block = np.random.randint(0, 255, (10, 100, 50))

    # Populate 2 layers with the same image:
    lyr1 = napari.layers.Image(block, colormap="green", blending="additive")
    lyr2 = napari.layers.Image(block, colormap="magenta", blending="additive")
    # Create 3D image layer
    viewer.add_layer(lyr1)
    viewer.add_layer(lyr2)
    # Here we translate and stretch on the 2nd axis (y, vertical in 2D view):
    lyr1.translate = [0, 40, 0]
    lyr1.scale = [1, 0.1, 1]
    # Roll dims:

    print(lyr1.translate)
    viewer.dims._roll()
    # this could also be viewer.dims._roll()
    print(lyr1.translate)

    # Now the stretch and the translation should be on the 3rd axis (x, horizontal)
    # but are still on the horizontal one unless we set again the translate/scale

    # Repeating these lines here produce the correct view *until a new roll is done*:
    # lyr1.translate = [0, 40, 0]
    # lyr1.scale = [1, 0.1, 1]

```

I updated the pull request invoking `self._reset_base()` instead of `_on_translate_change()`, in this way the scale is updated as well. The same could be done by calling also `on_scale_change()`.

I also added a test that seems to fail in the old version and pass with this additional line, where this is tested on the 2D view of a 4D image.

vigji:
A more general question: currently translation over the scrollbar axis is not supported. Is this something that can happen in the "real world coordinates" reorganisation?

napari/_vispy/vispy_image_layer.py
```diff
@@ -93,6 +93,9 @@ def _on_data_change(self, event=None):
                 self.node.set_data(data)
             else:
                 self.node.set_data(data, clim=self.layer.contrast_limits)
+
+        # Call to update order of translation values with new dims:
+        self._reset_base()
```
sofroniewn:
```suggestion
        self._on_scale_change()
        self._on_translate_change()
```
Looking at what is inside `_reset_base` I think you can just call these two methods and it will be fine

napari/_tests/test_viewer.py
```diff
@@ -135,5 +135,58 @@ def test_changing_theme(viewer_factory):
     viewer.theme = 'light'
     assert viewer.palette['folder'] == 'light'
 
+
+    # Close the viewer
```
sofroniewn: I don't think we need the explicit close anymore and it is causing our tests to fail. @tlambert03 can you comment / advise?

napari/_tests/test_viewer.py
```diff
@@ -135,5 +135,58 @@ def test_changing_theme(viewer_factory):
...
+    viewer.window.close()
+
     with pytest.raises(ValueError):
         viewer.theme = 'nonexistent_theme'
+        
+        
+def test_roll_traspose_update(qtbot):
```
sofroniewn: Similarly here I think you can use `viewer_factory` instead of `qtbot` - see [writing tests](https://github.com/napari/napari/blob/master/docs/developers/CONTRIBUTING.md#writing-tests) in our contributing guide. Right @tlambert03?

sofroniewn: You will then be able to drop the `viewer.window.close` bellow too

napari/_tests/test_viewer.py
```diff
@@ -135,5 +135,58 @@ def test_changing_theme(viewer_factory):
...
+    """ Controlling that translation and rolling preserve correct sequence
+    of translate and scaling values.
+    """
+
+    def check_view_consistency(viewer, transf_dict):
```
sofroniewn: This is great! Very nice test!!

sofroniewn (changes requested):
@vigji this is looking great, thanks for working on this so promptly. I've left a couple requests for changes in line, probably the large one being to use our `viewer_factory` for tests instead of `qtbot`, I left a link to documentation above.

Also can you add the `_on_scale_change`, `_on_translate_change` calls to the `_on_data_change` methods for the rest of the vispy layers, not just `image`.

Given that can you also make you layer parameterized over all our different layer types - see here https://github.com/napari/napari/blob/master/napari/_tests/test_viewer.py#L63 for an example on how to do that.

Thanks again for a great contribution!!

[EDIT] Also it looks like your Black Formatting and Flake8 is failing, can you try fixing those too. There are [some instructions here](https://github.com/napari/napari/blob/master/docs/developers/CONTRIBUTING.md#setting-up-a-development-installation).

sofroniewn:
> A more general question: currently translation over the scrollbar axis is not supported. Is this something that can happen in the "real world coordinates" reorganisation?

Yes, that will 100% be a goal of the world coordinates support

napari/_vispy/vispy_image_layer.py
sofroniewn: Will be faster not to unnecessarily call the other methods

sofroniewn: Curious also @tlambert03 if we might want to put some sort of check inside `_on_scale_change` etc. to see if the scale has actually changed and only make the call if so, otherwise we might take a real performance hit for two unnecessary calls each time data is updated, or does vispy already do such a check under the hood?

napari/_tests/test_viewer.py
tlambert03: yeah remove the `viewer.window.close()` call.  See #2 in the [section on writing tests in the contributing guide](https://napari.org/docs/developers/CONTRIBUTING.html#writing-tests) for details.

napari/_tests/test_viewer.py
tlambert03: yep

napari/_tests/test_viewer.py
```diff
@@ -135,5 +135,58 @@ def test_changing_theme(viewer_factory):
...
+            disp_dims = viewer.dims.displayed  # dimensions displayed in 2D
+            vis_vals = getattr(vis_lyr, transf_name)[
+                1::-1
+            ]  # values of visual layer
+            correct_vals = transf[disp_dims]  # expected translate/scale values
+            assert (vis_vals == correct_vals).all()
+
+    viewer = Viewer()
+    view = viewer.window.qt_viewer
+    qtbot.addWidget(view)
```
tlambert03: as described [here](https://napari.org/docs/developers/CONTRIBUTING.html#writing-tests) it's potentially problematic to _both_ add a widget to the `qtbot` and perform your own cleanup (`viewer.window.close()`) at the end of the test.  If at all possible, prefer the `viewer_factory()` to creating your own Viewer,

napari/_vispy/vispy_image_layer.py
tlambert03: not immediately sure.  couldn't quickly find how `_on_scale_change` ultimately makes its way back to the vispy code...

sofroniewn: hmm, i think it is pretty easy to add something like `if np.all(self.scale == scale): return` in the scale setter https://github.com/napari/napari/blob/master/napari/_vispy/vispy_base_layer.py#L94, and similar for translate, then we don't run the risk. @vigji can you give this a go too?

vigji:
I have now added the updating to all layers, parameterised the test and added a check on the property setter so that it does not call useless updates if nothing's changing. I noticed running the tests on the full battery of layers that I needed a funny factor 2 when making the comparison for a pyramid layer, but I have 0 experience with them so right now I just added the arbitrary division there.

vigji:
I also changed the `test_changing_theme` as it was trying to close the viewer and failing a test, it's not something I wrote though, don't know what's the etiquette for that... :)

napari/_tests/test_viewer.py
```diff
@@ -137,3 +137,54 @@ def test_changing_theme(viewer_factory):
...
+        vis_lyr = viewer.window.qt_viewer.layer_to_visual[layer]
+
+        # Visual layer attributes should match expected from viewer dims:
+        for transf_name, transf in transf_dict.items():
+            disp_dims = viewer.dims.displayed  # dimensions displayed in 2D
+            vis_vals = getattr(vis_lyr, transf_name)[
+                1::-1
+            ]  # values of visual layer
+
+            # If layer's a pyramid scale is multiplued by a factor 2:
```
sofroniewn: Interesting, I think it makes sense to let this go in now, as I am working on a fix for some pyramid stuff in #1053 that can address this in more detail

sofroniewn: Though maybe you could fix `multiplued ` to `multiplied`

tlambert03:
> I also changed the `test_changing_theme` as it was trying to close the viewer and failing a test, it's not something I wrote though, don't know what's the etiquette for that... :)

hmm... `test_changing_theme` is not trying to close the viewer [in the master branch](https://github.com/napari/napari/blob/ca06948e6039313e8a348a9a7762ee77fb4d478a/napari/_tests/test_viewer.py#L130-L139).  is it possible you're working off of an old branch?  have you merged master recently?

vigji:
> > I also changed the `test_changing_theme` as it was trying to close the viewer and failing a test, it's not something I wrote though, don't know what's the etiquette for that... :)
> 
> hmm... `test_changing_theme` is not trying to close the viewer [in the master branch](https://github.com/napari/napari/blob/ca06948e6039313e8a348a9a7762ee77fb4d478a/napari/_tests/test_viewer.py#L130-L139). is it possible you're working off of an old branch? have you merged master recently?

Oh drat, I totally am...apologies, I will update and merge it

napari/_tests/test_viewer.py
vigji: ok!

vigji:
I was seeing the old version when editing but now everything seems to be fine, the pull happened after merging the master branch

sofroniewn (approved):
@vigji I have updated this PR to fix the special case of the pyramid layer. The details there are that the pyramid layer has an additional transform that we call `tile2data` that needs to be taken into account to deal with the multiresolution nature of its data. For non pyramid data the transform is an identity transform and so could be merged. The value of `2` was arbitrary coming only from the choice of data in the pyramid (which was downsamples by factors of `2`). The test now should be robust to changes in data there.

Great work on this, I am approving. Would be nice if @jni could have a quick look before merge too.

napari/_tests/test_viewer.py
```diff
@@ -137,3 +137,61 @@ def test_changing_theme(viewer_factory):
 
     with pytest.raises(ValueError):
         viewer.theme = 'nonexistent_theme'
+
+
+@pytest.mark.parametrize('layer_class, data, ndim', layer_test_data)
+def test_roll_traspose_update(viewer_factory, layer_class, data, ndim):
+    """ Controlling that translation and rolling preserve correct sequence
+    of translate and scaling values.
+    """
```
jni:
```suggestion
def test_roll_transpose_update(viewer_factory, layer_class, data, ndim):
    """Check that transpose and roll preserve correct transform sequence."""
```

napari/_tests/test_viewer.py
```diff
@@ -137,3 +137,61 @@ def test_changing_theme(viewer_factory):
...
+
+    def check_view_consistency(layer, viewer, transf_dict):
+        """ Utility function for doing the checks.
+        """
```
jni:
```suggestion
        """Utility function for doing the checks."""
```

It would also be nice to describe *what* checks the function is doing.

jni: Also, does this need to be a nested function?

napari/_vispy/vispy_base_layer.py
```diff
@@ -93,6 +94,9 @@ def scale(self):
 
     @scale.setter
     def scale(self, scale):
+        # Avoid useless update if nothing changed in the displayed dims:
+        if self.scale is not None and np.all(self.scale[:2] == scale[:2]):
```
jni: I'm confused about why this is `:2` instead of `-2:`? And what about 3D?

jni: (ie, it would be good to add a comment here and on `translate` about what dimensions exactly this is comparing.)

jni (approved):
@vigji thank you for this and @sofroniewn thanks for the review! I've made some minor formatting suggestions and also asked for a clarification, but otherwise this LGTM!

sofroniewn:
Thanks @jni for the review. I will quickly fix the doc strings @vigji and the remaining comment

napari/_tests/test_viewer.py
sofroniewn: added doc string and moved to test utils

napari/_tests/test_viewer.py
sofroniewn: changed

napari/_vispy/vispy_base_layer.py
sofroniewn: it's because the vispy properties are always 4-vectors, i added a comment and padded the scale, translate so it works in all 2D and 3D cases too, good catch

sofroniewn:
This is ready to go in, I'll note that there are some double calls to the `scale` / `translate` setters that happen (some of them may already have been happening) which is why the comparisons and returns are needed. This is definitely something we'll want to fix in #1040!! But that can come later.

I'll merge later tonight

vigji:
Thanks a lot for fixing these last things, and sorry I was out these last days! Happy to see this merged :)

Issue #1068, opened 2020-03-20
Meta issue: roadmap for exporting output from napari


GenevieveBuckley:
## 🚀 Feature

Meta-issue for exporting data, annotations, etc. from napari in a consistent way.

## Motivation

The motivation here is pretty clear: in order for napari to be useful as part of an analysis pipeline it should support basic export capabilities. Users are already asking for these features: https://github.com/napari/napari/issues/382

## Discussion

**I want to hear from everyone about what export formats should be supported.**
I'll update the list below as people comment. 

At this stage, I'd like an emphasis on formats that should be easy for non-programmers to work with,
and formats that give us the greatest compatibility with other tools (eg: for further analysis).

Also, now that @tlambert03 has put in a bunch of work to enable plugins https://github.com/napari/napari/pull/937 (& other PRs),
I'd like to hear from him about how we can best use that infrastructure here as well.

### napari viewer
* serializable, round trip capable

### layer group
* layer groups don't exist yet, but would we also want them to be serializable too?

### image layer
* all imageio formats
* zarr

### labels layer
* all imageio formats
* zarr

### points layer
* csv table
* ImageJ compatible point ROIs
* SVG

### shapes layer
* csv table
* ImageJ compatible ROIs
* SVG
* geoJSON

### surface layer
* PLY
* STL
* OBJ (ASCII variant)

### vector layer
* csv table
* SVG

## Additional context

@danielballan has just started a discussion around a SciPy wide reader protocol
* https://github.com/napari/napari/issues/1066
* https://blog.danallan.com/posts/2020-03-07-reader-protocol/

This PR https://github.com/napari/napari/pull/941, intended to address https://github.com/napari/napari/issues/382 was made before the io plugin infrastructure was in place.
I expect that the `write_csv` function introduced here should probably become a writer plugin instead.

sofroniewn:
@GenevieveBuckley thanks so much for kicking off this issue with all those great examples. One thing I'd like to discuss (could be here or in an even more focused issue) is what the API for our plugin writer hookspec should be.

For me a guiding principle with the plugin writer should be symmetry with the plugin reader. So while the reader ultimately provides a function that takes a path or list of paths and returns an object (list of layer data tuples) I think the writer function should take a path or list of paths, that same object (the list of layer data tuples), and an optional format / extension.

If we change the object the reader returns we should change the object the writer accepts. There should always be symmetry.

How does this sound @GenevieveBuckley @tlambert03 @jni?

[...]

Issue #1090, opened 2020-03-27
Unified error console and error reporting


tlambert03:
## 🚀 Feature
As discussed in meetings, napari should have an embedded, unified error console.  The idea being that, for the most "bundled" like usage scenarios, a user should not have to look in the console that started the napari process to see errors that occured.  It should also allow easy review and reporting of errors that have occured in the session, along the lines of #1024, but for all errors raised in napari. This would probably supersede #794 

**UPDATED 5/28 with Links to relevant topics:** 

- #794 describes a mechanism in which we would override `sys.excepthook` with our own handler.  That handler would have some logic that decides when to pop up an error to the user (for instance, if the exception was a subclass of some `NapariError` class ... or perhaps we would want to be even more permissive about what got shown).  This is probably the main thing we need to figure out and get right here.
- #1024 has some code that shows how to store and show tracebacks from errors that were previously raised.  Note: a key idea there was that `napari-plugin-engine` has a base exception class (`PluginError`) that [stores _all_ instances that are created in the class `_record` attribute](https://github.com/napari/napari-plugin-engine/blob/7ce84475aa003bdcd3e181b755be0e2e61fb14e8/napari_plugin_engine/exceptions.py#L14-L48).  We could do something similar here that would let the user review past exceptions, and potentially send them to us.
- see also #1303, that has the good idea to have this sort of "error report window" default to showing the most recent error...
- #1301 is another request to make certain errors (in that case, when opening an image fails) more visible to the user.  That's a key idea here, but rather than having a lot of `QDialog` instances scattered through the code, we should define a general API that we use throughout the codebase to show an error message in the GUI.  (for instance, as shown in #794)
- some of this stuff might be relevant for telemetry and bug reporting #621

justinelarsen:
Yes, this would be helpful. 

For example, I dragged an image to the napari canvas, but nothing happened. I saw an error in the console “No plugin found capable of reading ... .tiff.” It would be nice to have a user friendly message that displayed within napari and not just the console.

jni:
@justinelarsen well that is already a worrisome bug since we should definitely be able to open tiffs with the default plugin!!!! Did we forget to check for .tiff in addition to .tif?

tlambert03:
[...]

PR #1104, opened 2020-04-03
Writer plugins

sofroniewn:
# Description
[EDIT] 
This PR now add supports for writer plugins as discussed at length in #1068. The PR provides functionality for saving individual layers using a `napari_write_<layer_type>` hook spec that accepts the layer data and metadata and a for saving multiple layers using a `napari_get_writer` hook spec that can determine based on a list of `layer_types` whether it can write out a list of layer data tuples.

The saving functionality is exposed to the user through `napari.save_layer`, `viewer.layers.save`, and `layer.save` methods that all internally use the same functionality.

Builtin hook_implementations have been provided for the `Points` and `Image` layer, other layers will come in follow-up PRs.

Comprehensive tests have been added too.

UI elements/ Qt dialogs will come in future PRs.


napari/plugins/io.py
```diff
@@ -74,3 +74,98 @@ def read_data_with_plugins(
...
+
+            skip_impls.append(implementation)  # don't try this impl again
+            if implementation.plugin_name != 'builtins':
+                # If builtins doesn't work, they will get a "no writer" found
+                # error anyway, so it looks a bit weird to show them that the
+                # "builtin plugin" didn't work.
+                logger.error(err.format_with_contact_info())
+
+
+def write_image_with_plugin(
```
sofroniewn: @tlambert03 - curious if you think this function should be `write_layer_with_plugin` and accept a `layer_type: str` arg and then call the appropriate hook based on that arg instead of having one of these for each of our layer types?

sofroniewn: Another way would be for this function to just accept the layer object and to do the getting of the data and metadata dictionary inside this function. Is that better?

napari/components/add_layers_mixin.py
```diff
@@ -896,3 +900,44 @@ def _add_layer_from_data(
                 raise exc
 
         return layer
+
+    def _save_individual_layers_with_plugin(
```
sofroniewn: @tlambert03 one problem here is that the same plugin has to save all the different layer types - might be nicer to have one plugin per layer type. Not sure how to think about that in terms of UI / API

napari/components/add_layers_mixin.py
```diff
@@ -896,3 +900,44 @@ def _add_layer_from_data(
...
+        for layer in layers:
+            write_layer_with_plugin(
+                plugin_name,
+                path,
+                data=layer.data,
+                meta=layer._get_state(),
+                layer_type=layer.__class__.__name__.lower(),
+            )
+
+    def _save_multiple_layers_with_plugins(
```
sofroniewn: @tlambert03 is it now strange that with this function you don't specify the plugin name but search through all possible ones?

Are we weirdly coupling the ability to search through and the ability to select an individual plugin and what we really want for all our hookspecs is an API where if we pass a plugin name it uses it and if we don't it searches through to try and find one that works? We could extend this to the readers too - i could imaging a dialog which let you pick which reader you wanted to use too

napari/plugins/io.py
```diff
@@ -74,3 +74,106 @@ def read_data_with_plugins(
...
+
+            skip_impls.append(implementation)  # don't try this impl again
+            if implementation.plugin_name != 'builtins':
+                # If builtins doesn't work, they will get a "no writer" found
+                # error anyway, so it looks a bit weird to show them that the
+                # "builtin plugin" didn't work.
+                logger.error(err.format_with_contact_info())
+
+
+def write_layer_with_plugin(
```
sofroniewn: @tlambert03 do you like the function signature here? the other way could be to have one function per layer type, but that seems highly redundant, or to just pass in the layer here and grab what is needed inside the function, but this seems like the right balance to me.

napari/plugins/io.py
tlambert03: I like the `write_layer_...` function that you've started... and I do actually like the idea of just passing a layer, and making sure all layers provide the same interface for retrieving meta

napari/components/add_layers_mixin.py
tlambert03: I'm actually not sure I see the need for this function at all in its current design (that is, a case in which the user is trying to save multiple layers, but we're also using the individual `napari_write_layer` hookspec _and_ specifying a certain plugin)

I imagined two cases for saving multiple layers: 

1) the case where we use `napari_get_writer` to give _single_ plugins a shot to save all selected layers on their own
2) the case where we dispatch individual `napari_write_layer` hooks from _multiple_  plugins to save each layer at a time... kinda like this function.  This is sort of a "try to save all of this stuff anyway you can".  but this is complicated and we need to do some more work to prepare for it (like creating a super-directory into which all plugins will be writing the output).

but in either case, I don't think we need to support the case where the user is trying to save multiple layers, using a single plugin, but _not_ using `napari_get_writer`.  let's leave that up to plugins.

tlambert03: additional thought here:  we could implement our own `napari_get_writer` "fallback plugin" that accepts _any_ combination of `layer_type` strings, and dispatches each layer to the plugins via `napari_write_layer`, putting it all into a directory.

napari/components/add_layers_mixin.py
tlambert03: > what we really want for all our hookspecs is an API where if we pass a plugin name it uses it and if we don't it searches through to try and find one that works

this is what I think we want, and yes I see it as analogous to reader plugins:  when they don't select a plugin, it loops through implementations and picks the first one that hits, as determined by the sort order.  When they _do_ select one, it tries that specific one and fails if it doesn't work (we still need to add that for readers, but it's always been [something we planned to do](https://github.com/napari/napari/pull/1023#issuecomment-595551417)).

i think part of the weirdness here is actually due to the previous `_save_individual_layers_with_plugin` function, which I don't think belongs.

napari/plugins/io.py
tlambert03: i like this function, but actually wouldn't mind passing the layer itself.  I don't see us using this function as part of any external API, and we'll probably never call it with any other pattern than the one used here:

https://github.com/napari/napari/blob/6ff9ab143db01001db0e053d70a5b8df85e1a0e8/napari/components/add_layers_mixin.py#L919-L925

... so why not just pass a layer?

tlambert03:
coming along nicely @sofroniewn ... though it's clear we still have some design challenges to figure out :)

one additional sidenote: I'd really like to pick _either_ the word "save" or "write" and use it consistently everywhere, rather than switching between the two (e.g. `_save_multiple_layers_with_plugins` currently calls `write_data_with_plugins`).  I don't have a strong preference, and I suspect this conversation has been had before somewhere else (what with all the `imsave` and `imwrite` functions out there... @jni?)

napari/plugins/io.py
sofroniewn: sounds good, will pass layer

jni:
> I'd really like to pick either the word "save" or "write" and use it consistently everywhere

LOL tbh I have never made up my mind on this. Right now I’m thinking that the natural pairings are read/write and open/save.

sofroniewn:
@tlambert03 I have now implemented the cases you described in https://github.com/napari/napari/issues/1068#issuecomment-612546793 - I think it feels quite nice. At the top there is now one `viewer.save_selected` method that takes a path and optionally the name of a plugin - the rest is figured out for them, including what to do if one or more layers are selected.

Next steps are
1. Better error handling for all the cases inside writer functions in `plugins/io.py`
2. Connecting the Qt dialogs for saving, including passing the path and picking the optional picking of a  plugin_name. When choosing what `plugin_names` to show we'll want to look at if one layer is selected, and if so what is its type, or if multiple layers are selected and choose appropriately
3. Adding the default `napari_get_writer` plugin which looks at all the individual layer type plugins and does the `folder from path` + `layer.name` trick for naming files.
4. Adding default `napari_write_<layer_type>` for all our layer_types.

If @tlambert03 you can take 1. and 2. then I can press on with 3. and 4.

sofroniewn:
> 3. Adding the default napari_get_writer plugin which looks at all the individual layer type plugins and does the folder from path + layer.name trick for naming files.

I have now added a first attempt at this, probably needs some work around how to handle the path, the making of the folder, the appending of the names (and extensions?, i.e. should we add `tif` somewhere etc, of do we thing plugins can do that if needed??). It probably also needs some error handling etc around folder creation and also some try catches inside the layer data list loop. Tests are also failing because of circular imports with the `plugin_manager`. Maybe @tlambert03 you can take this part over too.

I will then work on 
> 4. Adding default napari_write_<layer_type> for all our layer_types.
For `image` we can use `imageio.imwrite`. For `points` and `shapes` we can use things from #941. Similarly the approach of #941 can be used for `vectors`. For `surface` we can see if there is something simple we can do that is compatible with `meshio` without depending on it otherwise we can just write out 3 `csv` files for it

napari/components/add_layers_mixin.py
```diff
@@ -896,3 +900,98 @@ def _add_layer_from_data(
                 raise exc
 
         return layer
+
+    def _save_layers(
```
tlambert03: while `save_selected` belongs on the `Viewer` object, I'm not sure this should live on the viewer object.  It doesn't need access to the viewer, just a list of layers.  I see this more as a top level `napari.save_layers` convenience function.  And, for that matter, i think `napari.plugins.io.write_multiple_layers_with_plugin` already IS that convenience function... so do we really need this one at all?

tlambert03: the "if len(layers) > 1" logic does belong somewhere, but that should then be the main function we use:

```python
def save_layers(path, layers, plugin_name=None):
    if isinstance(layers, list):
        if len(layers) > 1:
            return write_multiple_layers(path, layers, plugin_name)
        elif len(layers) == 1:
            layers = layers[0]
    if isinstance(layers, Layer):
        return write_single_layer(path, layers, plugin_name)
```

sofroniewn: changes made, i think this moves in the right direction, let me know if you have more suggestions

napari/plugins/io.py
```diff
@@ -74,3 +68,194 @@ def read_data_with_plugins(
...
+
+    if plugin_name is None:
+        # Loop through all plugins using first successful one
+        skip_impls = []
+        while True:
+            (writer, implementation) = plugin_manager.hook.napari_get_writer(
+                path=path,
+                layer_types=layer_types,
+                _return_impl=True,
+                _skip_impls=skip_impls,
```
sofroniewn: @tlambert03  I now get the following warning when using this hookspec (or a similar one for the readers). Not sure how serious it is, but would be good to either fix or suppress
```
/Users/nsofroniew/GitHub/napari/napari/plugins/io.py:116: UserWarning: Argument(s) ('__multicall__',) which are declared in the hookspec can not be found in this hook call
  _skip_impls=skip_impls,
```

napari/plugins/manager.py
```diff
@@ -26,6 +26,17 @@
 
 pluggy.manager._HookCaller = _HookCaller
 
+# the main plugin manager instance for the `napari` plugin namespace.
+_plugin_manager = None
+
+
+# TODO: ask Juan if there's a better pattern than this for getting a lazy singleton
```
jni: :joy: I literally hunted down this specific commit going, "why did he do this? There *must* be a better way." :joy: I honestly hate this pattern. But I can't find the motivation for this... Is it to avoid instantiating the plugin manager until... Until what? Why don't we want to instantiate the plugin manager?

jni: This page has a lot of options for singleton objects, and I like all of them better than a global `get_...()` function:

https://python-3-patterns-idioms-test.readthedocs.io/en/latest/Singleton.html

jni: Probably my favourite is the class variable approach:

```python
# Singleton/ClassVariableSingleton.py
class SingleTone(object):
    __instance = None
    def __new__(cls, val):
        if SingleTone.__instance is None:
            SingleTone.__instance = object.__new__(cls)
        SingleTone.__instance.val = val
        return SingleTone.__instance
```

napari/_qt/qt_plugin_sorter.py
```diff
@@ -241,24 +239,21 @@ class QtPluginSorter(QDialog):
...
-        self.plugin_manager = plugin_manager
         self.layout = QVBoxLayout()
         self.setLayout(self.layout)
         self.hook_combo_box = QComboBox()
         self.hook_combo_box.addItem(self.NULL_OPTION)
 
         # populate comboBox with all of the hooks known by the plugin manager
         hooks = []
-        for name, hook_caller in vars(plugin_manager.hooks).items():
+        for name, hook_caller in vars(get_plugin_manager().hooks).items():
```
jni: In this specific spot I dislike that we have lost the ability to use this for arbitrary plugin managers... It's not like we expect to use it as such, but it seems unnecessary to remove it (?), and it makes it a bit more difficult to track down, from glancing at the code, where this magical object comes from.

I guess if we go down this route it should be crystal clear from the function/class name, "NapariGlobalPluginManager()` or somesuch.

I dunno. Possibly I need to sleep on this, but this solution leaves me uncomfortable. Feel free to convince me otherwise! :grimacing: 

You can blame my Computing Fundamentals B teacher... He taught the whole class to reflexively respond:

Teacher: "Macros are...?"
Class: "Baaaaaaaad"
Teacher: "Globals are...?"
Class: "Baaaaaaaad"

It's two decades later and it still sticks! :joy:

napari/components/add_layers_mixin.py
```diff
@@ -896,3 +896,56 @@ def _add_layer_from_data(
...
+
+    def save_layers(
+        self,
+        path: str,
+        selected: bool = True,
+        plugin_name: Optional[str] = None,
+    ):
+        """Save all or only selected layers to a path using writer plugins.
+
+        If no `plugin_name` is provided and only one layer is targted, then we
```
jni:
```suggestion
        If no `plugin_name` is provided and only one layer is targeted, then we
```

napari/components/add_layers_mixin.py
```diff
@@ -896,3 +896,56 @@ def _add_layer_from_data(
...
+        just directly call ``plugin_manager.hook.napari_write_<layer>()`` which
+        will loop through implementations and stop when the first one returns a
+        non-None result. The order in which implementations are called can be
+        changed with the implementation sorter/disabler.
+
+        If no `plugin_name` is provided and multple layers are targted, then
```
jni:
```suggestion
        If no `plugin_name` is provided and multiple layers are targeted, then
```

Having issues with the butterfly keyboard, @sofroniewn? :joy:

napari/components/add_layers_mixin.py
```diff
@@ -896,3 +896,56 @@ def _add_layer_from_data(
...
+        we call ``plugin_manager.hook.napari_get_writer()`` which loops through
+        plugins to find the first one that knows how to handle the combination
+        of layers and is able to write the file. If no plugins offer
+        `napari_get_writer` for that combination of layers then the default
+        `napari_get_writer` will create a folder and call
+        `napari_write_<layer>` for each layer using the `layer.name` variable
+        to modify the path such that the layers are written to unique files in
+        the folder.
+
+        If a `plugin_name` is provided and a single layer is targted, then
```
jni: (a) typo, (b) this is repeated from above...

napari/components/add_layers_mixin.py
```diff
@@ -896,3 +896,56 @@ def _add_layer_from_data(
...
+
+        Parameters
+        ----------
+        path : str
+            A filepath, directory, or URL (or a list of any) to open.
+        selected : bool
+            Optional flag to only save selected layers. True by default.
+        plugin_name : str, optional
+            Name of the plugin to use for saving. If None then all plugins
+            corresponding to appropriate hook specification will be loop
```
jni:
```suggestion
            corresponding to appropriate hook specification will be looped
```

napari/components/layerlist.py
```diff
@@ -69,6 +69,12 @@ def _update_name(self, event):
         layer = event.source
         layer.name = self._coerce_name(layer.name, layer)
 
+    @property
+    def selected(self):
+        """List: List of selected layers.
+        """
```
jni:
```suggestion
        """List of selected layers."""
```

napari/plugins/_builtins.py
```diff
@@ -29,3 +36,125 @@ def napari_get_reader(path: Union[str, List[str]]) -> ReaderFunction:
...
+def napari_write_image(path: str, data: Any, meta: dict) -> bool:
+    """Our internal fallback image writer at the end of the plugin chain.
+
+    Parameters
+    ----------
+    path : str
+        Path to file, directory, or resource (like a URL).
+    data : array or list of array
+        Image data. Can be N dimensional. If the last dimension has length
+        3 or 4 can be interpreted as RGB or RGBA if rgb is `True`. If a
```
jni:
```suggestion
        3 or 4 can be interpreted as RGB or RGBA if meta['rgb'] is `True`. If a
```

napari/plugins/_builtins.py
```diff
@@ -29,3 +36,125 @@ def napari_get_reader(path: Union[str, List[str]]) -> ReaderFunction:
...
+    """
+    imsave(path, data)
+    return True
+
+
+@napari_hook_implementation(trylast=True)
+def napari_write_points(path: str, data: Any, meta: dict) -> bool:
+    """Our internal fallback points writer at the end of the plugin chain.
+
+    Append `.csv` extension to the filename if it is not already there.
```
jni: I'd rather not do this. If someone writes `napari_write_points('file.xlsx', points_data)`, I want an error, not `file.xlsx.csv` to be created...

napari/plugins/_builtins.py
```diff
@@ -29,3 +36,125 @@ def napari_get_reader(path: Union[str, List[str]]) -> ReaderFunction:
...
+    bool : Return True if data is successfully written.
+    """
+    ext = os.path.splitext(path)[1]
+    if ext != '.csv':
+        path = path + '.csv'
+
+    # construct table from data
+    table = []
+    for row in data:
+        table.append(list(row))
```
jni:
```suggestion
    table = list(map(list, data))
```

But why isn't our own (?) `write_csv` capable of doing this?

jni: ... Indeed it is, so this whole thing can be removed?

jni: Finally, does this work with our properties layer? I think it should? Specifically:

- if the points layer has properties, they should be included in the table
- the coordinates data should include axis names as the column headers, now that we have them. It is fine if they are just integers as is done by default. I would also be ok with `coord-0`, `coord-1`, ..., or `coord-x`, `coord-y`, etc if they have provided names. Anyway that's a small decision, but the label names should be used.

napari/plugins/_builtins.py
```diff
@@ -29,3 +36,125 @@ def napari_get_reader(path: Union[str, List[str]]) -> ReaderFunction:
...
+
+    Returns
+    -------
+    callable
+        function that accepts the path and a list of layer_data (where
+        layer_data is (data, meta, layer_type)) and writes each layer.
+    """
+    if os.path.exists(path):
+        # If something exists at the current path return None
+        return None
```
jni: I find this surprising. I'd at least want a `exist_ok` flag to control this behaviour, e.g.:

https://docs.python.org/3/library/os.html#os.makedirs

I also *do* think we should check whether we can write all of the layers being requested. It would be annoying for napari to create a directory then realise it can't write a vectors layer.

jni: Could we somehow use a context manager here? @tlambert03 is this where wrappers would come in handy??? :grimacing: :joy: :grimacing: 

ie create a directory, pass to the writers, and then if anything goes wrong, delete the whole thing.

napari/plugins/manager.py
tlambert03: Well actually what we’re really trying to delay is the plugin discovery... but that’s definitely also worth discussing.  We were reasoning that since it imports all the plugin modules and might add to startup, perhaps we want to wait until someone tried to actually use a plugin.  But what do you think?

napari/plugins/_builtins.py
```diff
@@ -29,3 +36,125 @@ def napari_get_reader(path: Union[str, List[str]]) -> ReaderFunction:
...
+
+    plugin_manager = get_plugin_manager()
+    # Loop through data for each layer
+    for ld in layer_data:
+        # Get hook specification according to layer type
+        hook_specification = getattr(
+            plugin_manager.hook, 'napari_write_' + ld[2]
+        )
+        # Create full path using name of layer
+        full_path = os.path.join(path, ld[1]['name'])
```
jni: I like this idea! =)

napari/plugins/_hook_callers.py
```diff
@@ -16,43 +16,24 @@
...
->>> (result, imp) = execute_hook(
-...                     plugin_manager.hook.my_hook,
-...                     arg=arg,
-...                     return_impl=True,
-...                 )
-
-NOTE: A version code here was submitted to pluggy as PR #253:
-https://github.com/pytest-dev/pluggy/pull/253
-We may want to consider removing this code if that merges, but we may also just
-wish to maintain our internal hook looping logic.
```
jni: Why did you remove this? That PR is still languishing...

napari/plugins/_hook_callers.py
```diff
@@ -177,51 +158,240 @@ def _multicall(
...
+    # just for type annotation.  These are the lists that store HookImpls
+    _wrappers: List[HookImpl]
+    _nonwrappers: List[HookImpl]
 
-    The difference is that this function offers the ability to skip certain
-    implementations, and also calls our internal :func:`_multicall` function
-    above, which wraps each hook implentation call in a try/execpt, and offers
-    the option of returning the HookImpl responsible for the result along with
-    the result itself.
+    def get_hookimpl_for_plugin(self, plugin_name: str):
```
jni:
```suggestion
    def get_plugin_hook_implementation(self, plugin_name: str):
```
?

napari/plugins/_hook_callers.py
```diff
@@ -177,51 +158,240 @@ def _multicall(
...
+            raise ValueError("repeated item in order")
+
+        # make new lists for the rearranged _nonwrappers
+        # for details on the difference between wrappers and nonwrappers, see:
+        # https://pluggy.readthedocs.io/en/latest/#wrappers
+        _old_nonwrappers = self._nonwrappers.copy()
+        _new_nonwrappers: List[HookImpl] = []
+        indices = [self.index(elem) for elem in new_order]
+        for i in indices:
+            _new_nonwrappers.insert(0, _old_nonwrappers[i])
```
jni:
```suggestion
        indices = [self.index(elem) for elem in new_order]
        _new_nonwrappers: List[HookImpl] = [
            _old_nonwrappers[i] for i in indices
        ]
```

napari/plugins/_hook_callers.py
```diff
@@ -177,51 +158,240 @@ def _multicall(
...
+        # remove items that have been pulled, leaving only items that
+        # were not specified in ``new_order`` argument
+        # do this rather than using .pop() above to avoid changing indices
+        for i in sorted(indices, reverse=True):
+            del _old_nonwrappers[i]
+
+        # if there are any hook_implementations left over, add them to the
+        # beginning of their respective lists
+        if _old_nonwrappers:
+            _new_nonwrappers = [x for x in _old_nonwrappers] + _new_nonwrappers
```
jni: This probably came up last time I dealt with this code, but I'm confused again, which indicates that there should be a comment here... Why are the new_nonwrappers at the end rather than the beginning?

napari/plugins/_hook_callers.py
```diff
@@ -177,51 +158,240 @@ def _multicall(
...
+        ]
+        firstresult = self.spec.opts.get("firstresult") if self.spec else False
+        return _multicall(
+            hookimpls,
+            kwargs,
+            _return_impl=_return_impl,
+            firstresult=firstresult,
+        )
+
+        return self._hookexec(self, self.get_hookimpls(), kwargs)
```
jni: I'm confused about why this class is necessary now? Is this code lifted from pluggy? What's `not self.is_historic()`? Why does this method not have a docstring?

napari/plugins/_tests/test_hook_specifications.py
```diff
@@ -54,6 +54,10 @@ def test_annotation_on_hook_specification(name, func):
     sig = inspect.signature(func)
     if sig.parameters:
         for param in sig.parameters.values():
+            for forbidden in ('_plugin', '_skip_impls', '_return_impl'):
+                assert (
+                    param.name != forbidden
+                ), f'Must not name hook_specification argument "{forbidden}"!'
```
jni:
```suggestion
                ), f'Must not name hook_specification argument "{forbidden}".'
```

napari/plugins/io.py
```diff
@@ -74,3 +68,194 @@ def read_data_with_plugins(
...
+                    implementation.plugin_name,
+                    implementation.plugin.__name__,
+                )
+                err.__cause__ = exc  # like `raise PluginError() from exc`
+
+                skip_impls.append(implementation)  # don't try this impl again
+                if implementation.plugin_name != 'builtins':
+                    # If builtins doesn't work, they will get a "no writer"
+                    # found error anyway, so it looks a bit weird to show them
+                    # that the "builtin plugin" didn't work.
```
jni: I'm not sure I agree with this comment? I think it is useful to know that the builtin plugin failed, as well as that no working plugins were found.

jni (commented):
@sofroniewn @tlambert03 there's some design choices here that I don't really understand the rationale for. That doesn't mean they're wrong, I'd just like the decisions to be made explicit.

(1) Why do we need `get_plugin_manager` as opposed to having the plugin manager directly like we did before?
(2) Why did `execute_hook` get removed?
(3) If we are vendoring and modifying code, can we make it closer to what we want it to look like?

I had a private chat a week or two ago with @tlambert03 where we pondered whether pluggy was in fact the right choice for our plugin management. It's actually less popular than I would have expected by some measures — e.g. our repo already has more stars and more contributors. (!) Given the recent pytest fallout, I *am* now leaning more than ever towards making our own fork of it and doing with it what we please, including getting rid of all the confusing (to me) abbreviations.

I am just thinking out loud btw, I'm not asking you to reimplement pluggy in this PR! :joy: One answer to my concerns above could be, yep we are going to rip out pluggy but this is basically working, so let's get it in and worry about that later.

*However*, we *should* have this conversation because we don't want people writing a whole bunch of plugins using pluggy that we are then not compatible with! Should we, for example, recommend `@napari_plugin_implementation` as the right way to create plugins for now, which would allow us to seamlessly swap the, erm, implementation of our plugin manager?

napari/plugins/manager.py
jni: Could that be delayed by having a single object like we have now, and then having `.hooks` (or wherever the plugin discovery *needs* to have happened) be a property that triggers the plugin discovery?

tlambert03:
> @sofroniewn @tlambert03 there's some design choices here that I don't really understand the rationale for. That doesn't mean they're wrong, I'd just like the decisions to be made explicit.

yeah, I think we're still in the "fake it until you make it" phase of this PR, so all queries are welcome and nothing is set in stone.

> (1) Why do we need `get_plugin_manager` as opposed to having the plugin manager directly like we did before?

we just wanted to delay the discovery of plugins, anticipating that it might eventually add too much time to the startup.  not married to the concept at all, and more than happy to consider a different approach.  Since it's pretty hard to get anything into napari without our own reader plugins, maybe that's a fool's errand to begin with?


> (2) Why did `execute_hook` get removed?
> (3) If we are vendoring and modifying code, can we make it closer to what we want it to look like?

These 2 go together.  `execute_hook` got removed _because_ I'm less worried about vendoring and making it look exactly how we want (like you, due to the pytest fallout).  `execute_hook` was my first pass at overriding the behavior of directly calling the hook_caller (e.g. `plugin_manager.hook.hook_caller`), but now I'm just overriding the `_HookCaller.__call__` method itself and and making it exactly how we want it.  haven't added a docstring yet, but here are the use cases it supports: 

- `hook_caller(_plugin='name', ...)` - call a hook, requesting a specific plugin to be used, or fail.
- `hook_caller(_return_impl=True, ...)` - return the implementation responsible for the result (this was the original reason for execute_hook)
- `hook_caller(_skip_impls=True, ...)` - call as usual, but exclude certain implementations

> Given the recent pytest fallout, I _am_ now leaning more than ever towards making our own fork of it and doing with it what we please... Should we, for example, recommend `@napari_plugin_implementation` as the right way to create plugins for now?

I think we're all on exactly the same page here.  I'd like to avoid having everyone depend on pluggy as well... though i did like the ability to not depend directly on napari.  what do you think about making a `napari-plugins` package that just offers the `napari_plugin_implementation` decorator?  (or just keep it all in napari?). That is an extremely small amount of code, and, I think, a pattern that I would probably want to keep if we had our own plugin system.

napari/plugins/io.py
tlambert03: I agree here and was going to mention something like this.  Not knowing that the builtin failed leads to the (misleading) type of [message that Justine saw](https://github.com/napari/napari/issues/1090#issuecomment-612273408).  let's change it.

napari/_qt/qt_plugin_sorter.py
tlambert03: i don't need convincing!  just needed a better pattern :) hence the comment...
also, I'm happy to go back to the non-singleton approach if you liked it.

napari/plugins/_builtins.py
tlambert03: hah! ... I believe it might be :)

tlambert03:
thanks for taking a look @jni!  super helpful.  I won't put this on each comment, but I'll just mention that it's definitely still WIP, hence the lack of docstrings (didn't expect you to be on it so soon... but glad you are! hope it didn't waste too much time trying to reverse engineer).  definitely adding better docstrings soon

napari/plugins/manager.py
tlambert03: yep! was actually thinking the same thing

jni:
>  what do you think about making a napari-plugins package that just offers the napari_plugin_implementation decorator?

Yes, this is great, we can do this straight away, depending on pluggy, and then you get all the benefits of the "façade", which lets us change "backends", and of not having to depend on napari proper.

napari/_qt/qt_plugin_sorter.py
jni: I did. =)

napari/components/add_layers_mixin.py
sofroniewn: done

napari/components/add_layers_mixin.py
sofroniewn: done

napari/components/add_layers_mixin.py
sofroniewn: fixed typo, changed words to make clear how this is different

napari/components/add_layers_mixin.py
sofroniewn: fixed

napari/components/layerlist.py
sofroniewn: fixed

napari/plugins/_builtins.py
sofroniewn: fixed

napari/plugins/_builtins.py
sofroniewn: That's fine, but if no extension is provided then we should add one, i.e. `'file_name' -> 'file_name.csv'`. I can use `os.path.splitext(path)[1] == ''` to test if there is no extension, but the problem there is it will fail with paths like `"the.name.of.my.path.isnt.good"` which someone then wants a `csv` on. Curious what you recommend

napari/plugins/_builtins.py
sofroniewn: changed

napari/plugins/_builtins.py
sofroniewn: > the coordinates data should include axis names as the column headers, now that we have them. 

we don't have them - right now they reside on the viewer and not on the layer and so don't get passed in through the `meta` dict. This is true for both reading and writing.

@GenevieveBuckley took the approach of `dim-0` through `dim-N` in #941 which is where i took this from. I can copy over that approach here, it will make the transition when we have axis labels easier.

I can add the points `properties` too if the dict is not empty

napari/plugins/_tests/test_hook_specifications.py
sofroniewn: changed

tlambert03:
> > what do you think about making a napari-plugins package that just offers the napari_plugin_implementation decorator?
> 
> Yes, this is great, we can do this straight away

ok made it: https://github.com/napari/napari-plugins  
doesn't depend on pluggy and it's available on pypi

```python
from napari_plugins import napari_hook_implementation
```

napari/plugins/_builtins.py
```diff
@@ -29,3 +37,138 @@ def napari_get_reader(path: Union[str, List[str]]) -> ReaderFunction:
...
+    -------
+    bool : Return True if data is successfully written.
+    """
+    ext = os.path.splitext(path)[1]
+    if ext == '':
+        path = path + '.csv'
+    elif ext != '.csv':
+        raise ValueError('If an extension if provided then it must be `.csv`')
+
+    # construct table from data
```
sofroniewn, line 105: This has now been improved - the `examples/add_points_with_properties.py` now looks as followed when loaded into other software
image

I went with `index` and `axis-` as those words are intuitive for people not familiar with napari. Thoughts @jni @GenevieveBuckley @kevinyamauchi

napari/plugins/_builtins.py
sofroniewn: i've fixed this now, in that case i mention above, they can always add `.csv` at the end and then it's good

napari/plugins/_builtins.py
sofroniewn: Ok yeah - context manager sounds better, I forgot about `exist_ok` too

napari/_qt/qt_plugin_sorter.py
tlambert03: reverted and went with an approach where plugin discovery happens the first time an attribute of `plugin_manager.hook` is accessed _after_ hookspec and builtin plugin registration.

napari/plugins/_hook_callers.py
tlambert03: i've more or less given up on that PR... I added the text back, cause the link is good to have, but I don't expect we'll ever move away from our internal logic.  (and it's getting more firmly established here, but overwriting the _HookCaller.__call__ method).

napari/plugins/_hook_callers.py
tlambert03: because [they](https://github.com/napari/napari/pull/937/files/a7c7da78fa1f68bc6de9fe0e669d17fbb5762d11#diff-a034bf860a5ae6372e719ef0f973224f) [get called](https://github.com/napari/napari/pull/1023#discussion_r392614235) in [reverse order](https://github.com/napari/napari/blob/1ae61bbb428ddf4eeef57aebc89bc6474bcb7751/napari/_qt/qt_plugin_sorter.py#L152-L154) :)

will add a comment.

napari/plugins/_builtins.py
jni: I love it!

napari/plugins/_hook_callers.py
tlambert03: actually, I just committed this but then realized it doesn't work (breaks all the tests) and need to revert it.  were you trying to change the behavior or the syntax?  were you doing this conscious of the fact that they get called in reverse order?  I added another comment.

napari/plugins/_hook_callers.py
tlambert03: > I'm confused about why this class is necessary now?

the whole class??  or just the `__call__` method?  The class is absolutely necessary, it's how the hooks get called, and this `__call__` method is the main way to call the hooks.  I much prefer the syntax `plugin_manager.hooks.hook_spec(...)` to `execute_hook(plugin_manager.hooks.hook_spec, ...)`, so execute hook has gone away and now we just call the hookcaller directly, and the "features" that I had added via execute hook now no longer require using a different way of calling the hookcaller.

`is_historic` is there for now to keep pluggy compatibility.  See [historic hooks](https://pluggy.readthedocs.io/en/latest/#historic-hooks)

docstring coming, hadn't gotten to it before nick requested review.

napari/components/add_layers_mixin.py
```diff
@@ -896,3 +896,56 @@ def _add_layer_from_data(
...
+            A filepath, directory, or URL (or a list of any) to open.
+        selected : bool
+            Optional flag to only save selected layers. True by default.
+        plugin_name : str, optional
+            Name of the plugin to use for saving. If None then all plugins
+            corresponding to appropriate hook specification will be looped
+            through to find the first one that can save the data.
+        """
+        if selected:
+            layers = self.layers.selected
```
tlambert03: if selected is true but no layers are selected, should we do anything?

napari/plugins/_builtins.py
GenevieveBuckley: Me too!

napari/plugins/_builtins.py
GenevieveBuckley: > > the coordinates data should include axis names as the column headers, now that we have them.
> 
> we don't have them - right now they reside on the viewer and not on the layer and so don't get passed in through the `meta` dict. This is true for both reading and writing.

This is the second time this has been mentioned, so I've made an issue here so it can get picked up later on (not now, but eventually): https://github.com/napari/napari/issues/1144

napari/components/add_layers_mixin.py
```diff
@@ -917,6 +917,59 @@ def _add_layer_from_data(
 
         return layer
 
+    def save_layers(
```
tlambert03: if someone wanted to programmatically save a couple layers by name, index, or instance, they would need to call `write_layers_with_plugins` themselves directly. Should we make that easier by adding an optional layers argument here? It could take a list of layers, or indices/names for `LayerList.__getitem__`...  (or should we just make `write_layers_with_plugins` easier to get at)

additionally... should we add a `Layer.save(path)` method to the base layer? imagine:
```python
points_layer = viewer.add_points()
# interact with points layer
points_layer.save('path.csv')
```

napari/layers/points/points.py
```diff
@@ -512,7 +512,7 @@ def _add_point_color(self, adding: int, attribute: str):
                 setattr(self, f'{attribute}_color_cycle_map', color_cycle_map)
 
             new_colors = np.tile(
-                color_cycle_map[color_property_value], (adding, 1),
+                color_cycle_map[color_property_value], (adding, 1)
```
tlambert03: (pre-commit forced these changes...)

napari/components/add_layers_mixin.py
sofroniewn: I like making this function able to accept a list of layers or list of names - I wouldn't promote `write_layers_with_plugins`.

I'm not sure about the `Layer.save(path)` method. I see how it is convenient, but it would mean that the `Layer` itself would now need to know all about our plugin machinery which seems like mixing things across levels that shouldn't be mixed - unless you can think of a clean way to implement this

sofroniewn: Ah - just saw it implemented below - ok, I'm open

napari/plugins/_builtins.py
```diff
@@ -27,3 +35,144 @@ def napari_get_reader(path: Union[str, List[str]]) -> ReaderFunction:
...
+
+    Returns
+    -------
+    bool : Return True if data is successfully written.
+    """
+    ext = os.path.splitext(path)[1]
+    if ext == '':
+        path = path + '.csv'
+    elif ext != '.csv':
+        raise ValueError('If an extension is provided then it must be `.csv`')
```
sofroniewn: @tlambert03 should this `return False` instead of raising the ValueError? Should we emit a warning? Curious what you see as best practices that we should embody

napari/plugins/io.py
```diff
@@ -76,3 +78,191 @@ def read_data_with_plugins(
...
+    if plugin_name:
+        # if plugin has been specified we just directly call napari_get_writer
+        # with that plugin_name.
+        implementation = hook_caller.get_plugin_implementation(plugin_name)
+        writer_function = hook_caller(
+            _plugin=plugin_name, path=path, layer_types=layer_types
+        )
+    else:
+        (writer_function, implementation) = hook_caller(
+            path=path, layer_types=layer_types, _return_impl=True
```
sofroniewn, line 188: @tlambert03 I'm running into the following error here:

```python-traceback
viewer.save_layers('njs-test', selected=False)
TypeError                                 Traceback (most recent call last)
<ipython-input-1-ccf374d6cb14> in <module>
----> 1 viewer.save_layers('njs-test', selected=False)

~/GitHub/napari/napari/components/add_layers_mixin.py in save_layers(self, path, selected, plugin)
   1001             return
   1002 
-> 1003         write_layers_with_plugins(path=path, layers=layers, plugin_name=plugin)
   1004 
   1005 

~/GitHub/napari/napari/plugins/io.py in write_layers_with_plugins(path, layers, plugin_name, plugin_manager)
    140         if len(layers) > 1:
    141             return _write_multiple_layers_with_plugins(
--> 142                 path, layers, plugin_name, plugin_manager
    143             )
    144         elif len(layers) == 1:

~/GitHub/napari/napari/plugins/io.py in _write_multiple_layers_with_plugins(path, layers, plugin_name, plugin_manager)
    198     else:
    199         (writer_function, implementation) = hook_caller(
--> 200             path=path, layer_types=layer_types, _return_impl=True
    201         )
    202 

TypeError: cannot unpack non-iterable function object
```

sofroniewn:
I've started adding tests @tlambert03 - one fails and so is skipped for reasons noted above. Would be good to sync tomorrow on what additional tests we need

napari/plugins/_builtins.py
tlambert03: yeah, I think that should be the API (until we decide otherwise), changed it.
my reasoning is that if plugins _know_ they can't handle the request, they should gracefully exit without us having to handle the exceptions.  Let's leave exceptions for actual errors.

napari/plugins/io.py
tlambert03: fixed.  (The `_return_impls` argument is gone after #1153.). I unskipped your tests and they all pass now

tlambert03:
this is looking good @sofroniewn!  I think from an API perspective, I'm still slightly bothered that the `save_layers` method only lives on the `Viewer`.  The only thing we need the viewer for is if someone wants to save layers selected in a specific viewer.  I do kind of want our `write_layers_with_plugins` to be accessible (under a different name) without even instantiating a `Viewer`... this would be important for eventual headless stuff:

```python
from napari.layers import Image, Points
from napari import save_layers

img_layer = Image(img_data)
pts_layer = Point(point_data)

save_layers('path', [img_layer, pts_layer])
```

anyone else have API thoughts here?

sofroniewn:
> I think from an API perspective, I'm still slightly bothered that the save_layers method only lives on the Viewer. The only thing we need the viewer for is if someone wants to save layers selected in a specific viewer. I do kind of want our write_layers_with_plugins to be accessible (under a different name) without even instantiating a Viewer... this would be important for eventual headless stuff

I tend to agree - though for headless stuff we could promote working with the `ViewerModel` without the `Qt/ vispy` stuff which has the method on it. Also curious out how @jni would feel about having both a `napari.save_layers` and `viewer.save_layers` I seem to recall us not liking that for `add_image` hence it became `napari.view_image` and `viewer.add_image`, though maybe those we more different

tlambert03:
> Also curious out how @jni would feel about having both a `napari.save_layers` and `viewer.save_layers`

If we had `napari.save_layers` I would propose that we rename `viewer.save_layers` to `viewer.save_selected_layers` and/or `save_all_layers` (as it originally was). since the only thing we actually need the viewer for is the `LayerList` and it's current state.  I don't love the idea of promoting using the Viewer model if someone doesn't need to view anything.

jni:
Well, we have `viewer.open_path` so `viewer.save_layers` makes quite a bit of sense. I also think that people won't want to access/create layer lists when they access the viewer. I would allow:

- `napari.save_layers(path, layers)`
- `viewer.save_layers(path, selected=[True/False])  # False -> all` (I would make False the default)
- `viewer.save_layers(path, layers)`

Whether we allow `layers` to be a single layer is up in the air for me... I'd say yes.

sofroniewn:
@jni - @tlambert03 and I want to propose the following

```python
napari.save_layers(path, layers)
viewer.layers.save(path, selected=False) # False -> all (False the default)
layer.save(path)
```

What is nice about this is that it sets us up well for the composite design pattern with layer groups having a `save` method like individual layers, and positions us well in the future for the layer groups to be a nice object people can read and right too without ever thinking about the `Viewer`.

The current PR has been updated to reflect the pattern above. I'm working on some comprehensive tests now. This should be ready for a review soon

jni:
> `viewer.layers.save(path, selected=False) # False -> all (False the default)`

Hmm, although this is cute, it breaks the symmetry with open. So, I like it, but I have reservations. Would be good to hear from others about this before committing one way or the other.

Come to think of it, `viewer.open` seems a lot nicer than `viewer.open_path`, no?

sofroniewn:
>Hmm, although this is cute, it breaks the symmetry with open. So, I like it, but I have reservations. Would be good to hear from others about this before committing one way or the other.

It does, but I actually fell like we should have gone with `viewer.layers.open` and `viewer.layers.add_image`, so that people could use the `layers` object without the viewer. So I'm fine with the `save` asymmetry for now - but this could be a good swap to consider in `0.4.0` when we add support for the layer groups and composite design pattern.

> Come to think of it, viewer.open seems a lot nicer than viewer.open_path, no?

I prefer `open` too - we could update in #1149 before the 0.3.0 release too.

GenevieveBuckley:
Another +1 from me in favor of `open` vs `open_path`

sofroniewn:
This is basically ready from my end @tlambert03 - I ran into some problems with Flake8 and Pytest fixtures which meant I had to add all sorts of unpleasant skips to Flake8, can you maybe think about how to get around those.

Maybe one of the last things I'm not so sure about is whether our `builtins` `get_writer` should be forced to use `builtins` `write_image` etc rather than go through the plugin manager - it would make behavior much more deterministic when using `layers.save(path, plugin='builtin')` which right now could end up with anything. This tripped me up a bunch when writing tests. It would also make the implementation a bit cleaner as we could directly import those `write_image` methods. Thoughts @tlambert03?

tlambert03:
> I ran into some problems with Flake8 and Pytest fixtures which meant I had to add all sorts of unpleasant skips to Flake8, can you maybe think about how to get around those.

ah.  pytest fixtures (somewhat confusingly) aren't supposed to be imported explicitly.  Instead, you put them in a `conftest.py` file and then they are "magically" available to all tests that are in  that folder or any subfolder (so, for instance, the fixtures in `napari/conftest.py` are available anywhere, and those in `napari/plugins/_test/conftest.py` are only available for tests inside the plugins folder).  The word fixtures though is a general test term so (perhaps even more confusingly!) I used it for those test plugins in the plugins directory, but those are not `@pytest.fixture`s).  I moved some files around and removed the flake8 ignores.

> whether our builtins `get_writer` should be forced to use builtins `write_image` etc rather than go through the plugin manager

yes I like this a lot

napari/layers/points/points.py
```diff
@@ -1100,11 +1100,11 @@ def _get_state(self):
                 'symbol': self.symbol,
                 'edge_width': self.edge_width,
                 'face_color': self.face_color,
-                'face_color_cycle': self.face_color_cycle,
+                # 'face_color_cycle': self.face_color_cycle,
```
tlambert03: why are these commented out?  they are breaking `test_serialize.py` at `# Check number of properties is same as number in signature`

tlambert03: oh I see, cause otherwise `test_layer_save` fails at `# Compare layer metadata`...

tlambert03 (approved):
Looks good to me @sofroniewn! other than the test_serialize fail that I mentioned.
I'd like to refactor the plugin test a bit in a follow up PR.  but the actual behavior and coverage of the tests are great.

EDIT: It might also be nice to go back and modify the description of the first post in this thread summarizing what has been done here.

napari/layers/points/points.py
sofroniewn: ah yes - this is a problem. Maybe @kevinyamauchi or @jni can help me fix it. We can't return a `Cycle` here as that object doesn't serialize well - instead I want to do something like `list(self.face_color_cycle)` but that goes on for ever. One solution would be to not accept cycles for this property but only accept lists and then call `cycle` on the list elsewhere as a private attribute. Thoughts?

sofroniewn:
> > whether our builtins get_writer should be forced to use builtins write_image etc rather than go through the plugin manager
>
> yes I like this a lot

I've done this, but preserved the full capacity of the old function too - it might be nice in the future. Can you look over my implementation.

> Looks good to me @sofroniewn! other than the test_serialize fail that I mentioned.

As noted in the inline comment I need help here.

> EDIT: It might also be nice to go back and modify the description of the first post in this thread summarizing what has been done here.

I can do this.

This PR is now ready for review from @jni!!! (note despite the know test failure which i need help with)

docs/source/plugins/hook_specifications.rst
```diff
@@ -11,4 +11,25 @@ napari hook specification reference
...
\ No newline at end of file
+.. autofunction:: napari_get_reader
+.. autofunction:: napari_get_writer
+
+.. _write-single-layer-hookspecs:
+
+Single Layers IO
+''''''''''''''''
+
+The following hook specifications will be called when a user saves a single
```
jni, line 22: These are wonderful! Side note @tlambert03 : I remember you didn’t like napari.read_image(path), but to me it seems very naturally symmetric with the single layer writers... What do you think now?

napari/components/_tests/test_layers_list.py
```diff
@@ -424,3 +425,84 @@ def test_toggle_visibility():
...
+
+    # Check individual files still do not exist
+    for f in filenames:
+        assert not os.path.isfile(os.path.join(path, f))
+
+    # Check no additional files exist
+    assert set(os.listdir(tmpdir)) == set('')
+
+
+def test_layers_save_seleteced(tmpdir, layer_data_and_types):
```
jni, line 484: Selected

napari/components/layerlist.py
```diff
@@ -179,3 +185,74 @@ def toggle_selected_visibility(self):
...
+        self,
+        path: str,
+        *,
+        selected: bool = False,
+        plugin: Optional[str] = None,
+    ):
+        """Save all or only selected layers to a path using writer plugins.
+
+        If ``plugin`` is not provided and only one layer is targeted, then
+        we just directly call ``napari_write_<layer_type>`` hook specification
```
jni: ~~just~~

napari/components/layerlist.py
```diff
@@ -179,3 +185,74 @@ def toggle_selected_visibility(self):
...
+        (see :ref:`single layer writer hookspecs
+        <write-single-layer-hookspecs>`) which will loop through
+        implementations and stop when the first one returns a non-``None``
+        result. The order in which implementations are called can be changed
```
jni: This is incorrect, right? It’ll stop when it returns True?

napari/components/layerlist.py
```diff
@@ -179,3 +185,74 @@ def toggle_selected_visibility(self):
...
+        then we call
+        :meth:`~napari.plugins.hook_specifications.napari_get_writer` which
+        loops through plugins to find the first one that knows how to handle
+        the combination of layers and is able to write the file. If no plugins
+        offer :meth:`~napari.plugins.hook_specifications.napari_get_writer` for
+        that combination of layers then the default
+        :meth:`~napari.plugins.hook_specifications.napari_get_writer` will
+        create a folder and call ``napari_write_<layer_type>`` for each layer
+        using the ``Layer.name`` variable to modify the path such that the
+        layers are written to unique files in the folder.
```
jni, line 218: How are extensions handled in this case?

napari/conftest.py
```diff
@@ -57,3 +62,72 @@ def actual_factory(*model_args, **model_kwargs):
 
     for viewer in viewers:
         viewer.close()
+
+
+@pytest.fixture(params=['image', 'points', 'points-with-properties'])
+def layer_writer_and_data(request):
```
jni, line 68: This needs a docstring

napari/conftest.py
```diff
@@ -57,3 +62,72 @@ def actual_factory(*model_args, **model_kwargs):
...
+        data = np.random.rand(20, 2)
+        Layer = Points
+        layer = Points(data)
+        writer = napari_write_points
+        extension = '.csv'
+
+        def reader(path):
+            return (
+                io.read_csv(path)[0][:, 1:3],
+                {},
```
jni: Worth sprinkling comments here and in other def reader lines that this is the metadata.

napari/layers/points/points.py
jni:
```
from toolz import take

list(take(n, self.face_color_cycle))
```

Where n is the length of the desired list, which I guess we have access to here?

There is also `itertools.islice(cycle, None, n)` if you prefer to use the standard library. But we already depend on toolz and I find take more readable.

napari/plugins/_builtins.py
```diff
@@ -27,3 +36,198 @@ def napari_get_reader(path: Union[str, List[str]]) -> ReaderFunction:
...
+    elif ext != '.csv':
+        # If an extension is provided then it must be `.csv`
+        return False
+
+    if 'properties' in meta:
+        properties = meta['properties']
+    else:
+        properties = {}
+    # construct table from data
+    column_names = ['axis-' + str(n) for n in range(data.shape[1])]
```
jni, line 106: Can you add a todo or create an issue pointing to this line to say that we need to change this to the axis names once we get access to them?

napari/plugins/_builtins.py
```diff
@@ -27,3 +36,198 @@ def napari_get_reader(path: Union[str, List[str]]) -> ReaderFunction:
...
+    if bool(properties):
```
jni: `if properties` works if we are just going to rely on the truthiness of dictionaries...

napari/plugins/_builtins.py
```diff
@@ -27,3 +36,198 @@ def napari_get_reader(path: Union[str, List[str]]) -> ReaderFunction:
...
+        column_names += properties.keys()
+        prop_table = np.concatenate(
+            [np.expand_dims(p, axis=1) for p in properties.values()], axis=1,
+        )
```
jni: I *think* this can be replaced by `np.stack(properties.values(), axis=1)`

One disadvantage of this approach vs using pandas here is that we might lose type information. But I guess that that’s what plugins are for!

napari/plugins/_builtins.py
```diff
@@ -27,3 +36,198 @@ def napari_get_reader(path: Union[str, List[str]]) -> ReaderFunction:
...
+        table = np.concatenate([data, prop_table], axis=1)
+    else:
+        table = data
+
+    # add index of each point
+    column_names = ['index'] + column_names
+    indices = np.expand_dims(list(range(data.shape[0])), axis=1)
+    table = np.concatenate([indices, table], axis=1)
```
jni: concatenating and stacking are expensive operations that require copying the full dataset. It would be best to create a list of all the things we want to stack and then do it once.

jni: (When there are no properties, you can have an empty list, or an array of shape (n,0))

napari/plugins/_builtins.py
```diff
@@ -27,3 +36,198 @@ def napari_get_reader(path: Union[str, List[str]]) -> ReaderFunction:
...
+
+    # remember whether it was there to begin with
+    already_existed = os.path.exists(path)
+    # Try and make directory based on current path if it doesn't exist
+    if not already_existed:
+        os.makedirs(path)
+
+    try:
+        # build in a temporary directory and then move afterwards,
+        # it makes cleanup easier if an exception is raised inside.
```
jni, line 214: Awesome. I was thinking of this as I read the function docstring, and I’m so happy to find it already here! =D

napari/plugins/_tests/conftest.py
```diff
@@ -23,3 +27,40 @@ def builtin_plugin_manager(plugin_manager):
             plugin_manager.unregister(mod)
     assert plugin_manager.get_plugins() == set([napari.plugins._builtins])
     return plugin_manager
+
+
+@pytest.fixture
+def temporary_hookimpl(plugin_manager):
```
jni, line 33: Whoa. =D

napari/plugins/_tests/test_builtin_write_layers.py
```diff
@@ -0,0 +1,79 @@
+import os
+import numpy as np
+
+
+def test_write_layer_with_round_trip(tmpdir, layer_writer_and_data):
```
jni, line 6: Oh man. I hate the non-locality of fixtures in conftest. Can we at least have a comment that it comes from that file?

jni: Same everywhere else where this happens

napari/plugins/_tests/test_hook_specifications.py
```diff
@@ -80,18 +79,3 @@ def test_docs_match_signature(name, func):
...
-        name = getattr(sig_param.annotation, '_name', None)
-        name = name or getattr(sig_param.annotation, '__name__', None)
-        if doc_param.type != name:
-            msg = (
-                f'The type ({name}) for parameter '
-                f'"{sig_param.name}" in hook specification "{name}" does not '
-                'match the type specified in the docstring '
-                f'({doc_param.type})'
-            )
-            warnings.warn(msg)
```
jni, line 97: Why is this removed? Apologies if I missed an earlier discussion...

napari/plugins/exceptions.py
```diff
@@ -78,11 +78,13 @@ class PluginCallError(PluginError):
...
         plugin_name = hook_implementation.plugin_name
-        plugin_module = hook_implementation.plugin.__name__
-        specname = getattr(
-            hook_implementation,
-            'specname',
-            hook_implementation.function.__name__,
+        if hook_implementation.plugin:
+            plugin_module = hook_implementation.plugin.__name__
+        else:
+            plugin_module = None
```
jni, line 86: Can you add a comment about what this if/else block is about? I don’t get why it exists.

napari/plugins/io.py
```diff
@@ -54,25 +56,189 @@ def read_data_with_plugins(
...
-                f"hook 'napari_get_reader': {exc}"
-            )
-            # instantiating this PluginError stores it in
-            # plugins.exceptions.PLUGIN_ERRORS, where it can be retrieved later
-            err = PluginError(msg, plugin_name, plugin_module)
-            err.__cause__ = exc  # like ``raise PluginError() from exc``
-
-            skip_impls.append(hook_implementation)  # don't try this impl again
-            if plugin_name != 'builtins':
+            err = PluginCallError(result.implementation, cause=exc)
```
jni, line 59: This is very cute!

napari/plugins/io.py
```diff
@@ -54,25 +56,189 @@ def read_data_with_plugins(
...
                 logger.error(err.format_with_contact_info())
+
+
+def save_layers(
+    path: str, layers: List[Layer], *, plugin: Optional[str] = None,
+):
+    """Write list of layers or individual layer to a path using writer plugins.
+
+    If ``plugin`` is not provided and only one layer is passed, then we
+    just directly call ``plugin_manager.hook.napari_write_<layer>()`` which
```
jni: ~~just~~

napari/plugins/io.py
```diff
@@ -54,25 +56,189 @@ def read_data_with_plugins(
...
+            if result.implementation != 'builtins':
                 # If builtins doesn't work, they will get a "no reader" found
                 # error anyway, so it looks a bit weird to show them that the
                 # "builtin plugin" didn't work.
                 logger.error(err.format_with_contact_info())
+
+
+def save_layers(
+    path: str, layers: List[Layer], *, plugin: Optional[str] = None,
+):
```
jni: -> bool?

napari/plugins/io.py
```diff
@@ -54,25 +56,189 @@ def read_data_with_plugins(
...
+def _write_multiple_layers_with_plugins(
+    path: str,
+    layers: List[Layer],
+    *,
+    plugin_name: Optional[str] = None,
+    plugin_manager=napari_plugin_manager,
+):
+    """Write data from multiple layers data with a plugin.
+
+    If a ``plugin_name`` is not provided we loops through plugins to find the
```
jni: Loop

napari/plugins/io.py
jni: @tlambert03 Why does the call still have a `_return_impl` kwarg passed to it?

napari/utils/io.py
```diff
@@ -31,6 +32,19 @@ def imsave(filename: str, data: np.ndarray):
...
+    """Valid extensions of files that imsave can write to.
+
+    Returns
+    ----------
+    tuple
+        Valid extensions of files that imsave can write to.
+    """
+    import imageio
+
+    return tuple(set(x for f in imageio.formats for x in f.extensions))
```
jni: Are we sure this returns all valid writing extensions? I see .foobar and .nonexistentext in there... LOL

jni (commented):
I’ve made some code suggestions and requested some clarification in places, but this is very close!

jni (commented):
I’ve made some code suggestions and requested some clarification in places, but this is very close!

docs/source/plugins/hook_specifications.rst
tlambert03: you mean like as additional hook_specifications?  I agree that the syntactic symmetry is appealing... but what behavior are you specifically imagining vs our current arrangement?  I get tripped up when I start thinking about all the things that "`path`" could actually contain... like what does `napari_read_image('my.csv')` do?  what are the responsibilities of the plugins?  what does it give us over the current `napari_get_reader`?  
It's a bit easier for the writers since we _know_ what the input to the hooks is (it's one of our layers)... but we will never really know what the `path` being read contains.  (though perhaps the point here is to assert that it contains something and fail if not?)

napari/plugins/_tests/test_builtin_write_layers.py
tlambert03: yeah, it's extremely magical and hard to follow, I agree.  are you saying put a comment everywhere a fixture is used?

napari/plugins/_tests/test_hook_specifications.py
tlambert03: there wasn't any discussion.  it was just error prone and led to lots of warnings in the tests.  The goal was to make sure that docstrings matched type annotations, but since there's no standard for docstring syntax, it's _very_ hard to test equality... and it almost always warns that they are different.

docs/source/plugins/hook_specifications.rst
sofroniewn: I'm still open to this, but we can save for a future PR - I think if we hear from people that they struggle with the current `napari_get_reader` or if we start finding ourselves really wanting to use the `napari_read_image` then we can go down this road, but always easier to add rather than take away :-)

napari/components/layerlist.py
sofroniewn: It is wrong - needs to be fixed

napari/components/layerlist.py
sofroniewn: Our builtins have the ability to add a default extension in the case when none is provided

napari/layers/points/points.py
sofroniewn: > Where n is the length of the desired list, which I guess we have access to here?

I'm not actually sure if we do? I can't recall - maybe @kevinyamauchi can quickly jump in here

sofroniewn: Just had a good slack with @kevinyamauchi - he is going to make a small standalone PR to convert the `face_color_cycle` and `edge_color_cycle` to a list and change the doc strings/ typing appropriately (both in Points and Vectors) and then add a private property with the actual cycle on it that gets used. Once that PR is merged we will update this one accordingly

napari/components/layerlist.py
tlambert03: actually. it's correct, (in a `firstresult` hook, it goes until it gets a `not None` result... logic is [here](https://github.com/napari/napari/blob/e78647206109d915d9a9baa788a1e31f1e53d6eb/napari/plugins/_hook_callers.py#L214-L217)).  which means we need to modify our implementations slightly (this was an oversight in the first place):  As a general pattern, hooks should return `None` if they "abstain" or do nothing.  So I'll make all the `napari_write_layer` specs return `None` or a string (the path that was written), rather than a `bool`.

napari/components/layerlist.py
tlambert03: I added this docstring to all of the single layer writers specs:

>    It is the responsibility of the implementation to check any extension on
    ``path`` and return ``None`` if it is an unsupported extension.  If ``path``
    has no extension, implementations may append their preferred
    extension.

napari/plugins/exceptions.py
tlambert03: it's there for testing.  That `temporary_hookimpl` context manager above creates an implementation that wasn't imported from a module.  will add comment

napari/plugins/io.py
tlambert03: updated, will now be a list of strings (any filepaths that were written), or an empty list of nothing was written.

napari/plugins/io.py
tlambert03: No!  WE LOOPS IT!

NINTCHDBPICT0004777086082-2

napari/utils/io.py
tlambert03: i think those are intended more as exclusion criteria, not inclusion.  The vast majority of those extensions will also not work unless the data is appropriate, and sometimes also require additional imageio plugins to be installed.  Not sure how we should handle that though

sofroniewn: seems like having this is better than nothing, even if not ideal

napari/plugins/_builtins.py
```diff
@@ -27,3 +36,200 @@ def napari_get_reader(path: Union[str, List[str]]) -> ReaderFunction:
...
+        Name of the plugin to use for saving. If None then all plugins
+        corresponding to appropriate hook specification will be looped
+        through to find the first one that can save the data. By default,
+        only builtin napari implementations are used.
+    plugins.
+    plugin_manager : plugins.PluginManager, optional
+        Instance of a napari PluginManager.  by default the main napari
+        plugin_manager will be used.
+
+    Returns
```
sofroniewn, line 193: I know changes are still in progress but just flagging so as not to forget.
Doc string here needs update I think
```
    Returns
    -------
    path : str or None
        If data is successfully written, return the ``path`` that was written.
        Otherwise, if nothing was done, return ``None``.
    """
```

napari/plugins/_builtins.py
```diff
@@ -27,3 +36,200 @@ def napari_get_reader(path: Union[str, List[str]]) -> ReaderFunction:
...
+    Parameters
+    ----------
+    path : str
+        Path to file, directory, or resource (like a URL).
+    data : array (N, D)
+        Coordinates for N points in D dimensions.
+    meta : dict
+        Points metadata.
+
+    Returns
```
sofroniewn, line 87: I know changes are still in progress but just flagging so as not to forget.
Doc string here needs update I think
```
    Returns
    -------
    path : str or None
        If data is successfully written, return the ``path`` that was written.
        Otherwise, if nothing was done, return ``None``.
    """
```

napari/plugins/_builtins.py
```diff
@@ -27,3 +36,200 @@ def napari_get_reader(path: Union[str, List[str]]) -> ReaderFunction:
...
+    path : str
+        Path to file, directory, or resource (like a URL).
+    data : array or list of array
+        Image data. Can be N dimensional. If meta['rgb'] is ``True`` then the
+        data should be interpreted as RGB or RGBA. If ``meta['is_pyramid']`` is
+        ``True``, then the data should be interpreted as an image pyramid.
+    meta : dict
+        Image metadata.
+
+    Returns
```
sofroniewn, line 56: I know changes are still in progress but just flagging so as not to forget.
Doc string here needs update I think
```
    Returns
    -------
    path : str or None
        If data is successfully written, return the ``path`` that was written.
        Otherwise, if nothing was done, return ``None``.
    """
```

napari/plugins/io.py
```diff
@@ -54,25 +56,189 @@ def read_data_with_plugins(
...
+
+    If a ``plugin_name`` is not provided we loop through plugins to find the
+    first one that knows how to handle the combination of layers and is able to
+    write the file. If no plugins offer ``napari_get_writer`` for that
+    combination of layers then the default ``napari_get_writer`` will create a
+    folder and call ``napari_write_<layer>`` for each layer using the
+    ``layer.name`` variable to modify the path such that the layers are written
+    to unique files in the folder.
+
+    If a ``plugin_name`` is provided, then we call we call
```
kne42: typo

napari/plugins/io.py
```diff
@@ -54,25 +56,189 @@ def read_data_with_plugins(
...
+    (in plugins.exceptions.PLUGIN_ERRORS)
+
+    Parameters
+    ----------
+    path : str
+        The path (file, directory, url) to write.
+    layers : List of napari.layers.Layer
+        List of napari layers to write.
+    plugin_manager : plugins.PluginManager, optional
+        Instance of a napari PluginManager.  by default the main napari
```
kne42: typo

napari/plugins/_tests/conftest.py
```diff
@@ -23,3 +27,40 @@ def builtin_plugin_manager(plugin_manager):
...
+            with pytest.raises(PluginCallError):
+                writer(tmpdir, layer_data, plugin_manager)
+    """
+
+    @contextmanager
+    def inner(
+        func, specname, tryfirst=True, trylast=None, plugin_name="<temp>"
+    ):
+        caller = getattr(plugin_manager.hook, specname)
+        HookimplMarker('napari')(tryfirst=tryfirst, trylast=trylast)(func)
```
kne42, line 56: tbh this initially confused me because I was wondering why you would create a class without assigning it to anything

kne42 (approved):
looks solid, great work!

I left some comments on mostly docs stuff. I also noticed that you talked a lot in the 1st-person 
(e.g. "we", "our"). Idk if that's okay, maybe @jni can chime in.

napari/plugins/_tests/conftest.py
tlambert03: I can leave a comment if it would help?  `HookimplMarker` creates a decorator.  And when we call it on `func` it just sets an attribute on `func`

napari/plugins/_tests/test_builtin_write_layers.py
jni: Whenever a fixture is used that is not defined in that file. The comment should specify the file where the fixture is defined.

napari/utils/io.py
jni: I think imageio provides a bunch of writers classes. We should try to find the extensions for which such a class exists. Also, imho we should use tifffile directly for writing tiffs.

tlambert03: We are using tifffile directly for tiffs

tlambert03: We are using tifffile directly for tiffs

napari/plugins/_tests/test_builtin_write_layers.py
tlambert03: Did that.  I’ll also note that (in my IDE at least) it works to “go to definition” on the argument in the signature.  Takes you right to the def

docs/source/plugins/hook_specifications.rst
jni: Yeah, it's fine to save this for a future PR.

> like what does napari_read_image('my.csv') do?

It searches for a plugin that can read an image from a csv, if there is one it returns that image, if there isn't, it fails. Am I missing something?

> what does it give us over the current napari_get_reader?

A simple decorator that can be added to reader functions (path -> arraylike) as-is. Basically, we can start using Dan Allan's universal reader functions as napari plugins, with ~0 work from library authors.

I guess the concern is that the fast return is now not guaranteed?

napari/plugins/_tests/test_builtin_write_layers.py
jni: > Takes you right to the def

That's pretty insane. =D But, I am old fashioned and I think code should be readable without aids. =) Thanks for adding the comments!

napari/layers/points/points.py
kevinyamauchi: I have made the PR in #1163 . I need to do a bit more testing, but I think it is good to go.

tlambert03:
Took another look through and incorporated what I could.  @sofroniewn, the main things remaining now have to do with the implementations of the builtin points writer and tests (which is maybe fixed now by #1163 ?). can you take those?

tlambert03 (approved):

napari/plugins/_builtins.py
tlambert03: fixed with @jni's help

napari/utils/io.py
tlambert03: changed to a more realistic set of extensions, (by testing them all with a variety of image shapes/formats).  with a TODO to write a proper plugin

tlambert03:
ok I fixed the csv reader/writers on windows and handled all the remaining comments.
I think all tests are passing now and this is ready for final approval

sofroniewn:
> ok I fixed the csv reader/writers on windows and handled all the remaining comments.
I think all tests are passing now and this is ready for final approval

Thanks so much @tlambert03 for taking those. I have looked everything over again and this is now ready for merge in my opinion. We can wait for @jni to look one more time and then merge

Issue #1098, opened 2020-04-01
Pillow 7.1.0 breaking tests (and screenshots)


tlambert03:
## 🐛 Bug
Pillow 7.1.0 was released 3 hours ago and appears to be breaking tests.  https://cirrus-ci.com/task/5001684665499648?command=test#L800

changelog: https://github.com/python-pillow/Pillow/blob/master/CHANGES.rst

PR #1108, opened 2020-04-03
Pin pillow at <= 7.0.0

tlambert03:
# Description
#1098 was fixed in Pillow 7.1.1, but it looks like there's still an issue that is affecting us too: https://github.com/python-pillow/Pillow/issues/4518.
Given the imminent release, it seems like a safer bet to pin pillow at 7.0.0 rather than just avoiding specific versions.

## Type of change

- [x] dependency

### DIFF
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -1,7 +1,7 @@
 dask[array]>=2.1.0
 fsspec>=0.3.3
 imageio>=2.5.0
-Pillow!=7.1.0  # not a direct dependency; remove once fixed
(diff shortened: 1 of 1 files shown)


sofroniewn (approved):
LGTM!

jni:
FWIW I still prefer blocking individual versions until fixed, as it's easier to know when it's fixed...

tlambert03:
Sorry I merged it so fast.  want to change it back for the release? Or release with this and then change it after the release?

PR #1109, opened 2020-04-03
Use canvas.render instead of grabFramebuffer for screenshot

### LINKED ISSUE #1105: Screenshots taken during a pytest run appear entirely black
## 🐛 Bug

Screenshots taken during a pytest run appear entirely black.

## To Reproduce

1. In pytest (this test produces an AssertionError when it should not)

```python
import numpy as np
from skimage.data import camera


def test_screenshot_introspection(viewer_factory):

    layer = viewer.add_image(camera())
    screenshot = viewer.screenshot()
    # the screenshot should NOT be completely black
    assert not (screenshot == np.array([0, 0, 0, 255])).all()
```

2. With python (this script does not produce an AssertionError, which is good and what we expect)

```
import napari
import numpy as np
from skimage.data import camera


with napari.gui_qt():
    viewer = napari.Viewer()
    layer = viewer.add_image(camera())
    screenshot = viewer.screenshot()
    # the screenshot should NOT be completely black
    assert not (screenshot == np.array([0, 0, 0, 255])).all()

```

## Expected behavior

I expected the contents of a screenshot acquired during a pytest run to look the same as a screenshot taken when running napari normally.

## Environment

I'm running napari using the current master branch:

napari: 0.2.12+37.g7a831db.dirty
Platform: Windows-10-10.0.17134-SP0
Python: 3.8.1 (default, Mar 2 2020, 13:06:26) [MSC v.1916 64 bit (AMD64)]
Qt: 5.14.1
PySide2: 5.14.1
VisPy: 0.6.4
NumPy: 1.18.2
SciPy: 1.4.1
scikit-image: 0.16.2
Dask: 2.12.0

GL version: 4.4.0 - Build 20.19.15.5058
MAX_TEXTURE_SIZE: 16384

## Additional context

We talked about using screenshot introspection to write some tests for https://github.com/napari/napari/pull/1053 (that PR addresses issue https://github.com/napari/napari/issues/1050). Screenshot introspection is a better idea than `pytest-mpl` because then we won't clog up the repository with lots of big files with benchmark images.

tlambert03:
# Description
The current `qt_viewer.screenshot` strategy requires uses the Qt [`grabFramebuffer()`](https://doc.qt.io/qt-5/qopenglwidget.html#grabFramebuffer) method to get a copy of the canvas... but this requires that the widget actually be visible.
Vispy provides a [`SceneCanvas.render()`](https://github.com/vispy/vispy/blob/8387ea1aac961224488657ff57d1794f4e68ab32/vispy/scene/canvas.py#L219) method that renders the scene to an offscreen buffer and returns the image array.  This lets us take a screenshot of the canvas even when the viewer is not shown.

this would fix #1105 ... and allow screenshot tests to work without explicitly passing `viewer_factory(show=True)`.  Note... for screenshots that include the viewer window, it will still be necessary to pass `show=True`

## Type of change

# How has this been tested?

- [x] the example test in #1105 passes

## Final checklist:


sofroniewn:
@tlambert03 so funny, I was never able to get that to work - I always got error messages or there were problems with layer ordering / blending modes etc. I just tried your PR and it seems to work. Do you know if vispy recently made a fix? We should now add some pretty comprehensive introspection tests, like multiple layers on top of each other, make sure the blended pixel is what we expect etc.

Also it does seem like quite a lot of our tests are failing with the new approach

tlambert03:
huh, not sure if they changed anything.  (it's the first time I've played with it).

annoying on the tests... they don't fail locally or on mac CI, so it's gotta be that whole virtualized GL thing again, which i still don't really have a good grasp on.  so that would obviously need to be figured out too.

sofroniewn:
Yeah - see #181 for a previous screenshot bug. I can't seem to reproduce now, so maybe it is fixed / a different approach, but that is part of why I am wary. Can we also fix #1105 just by the `show=True` parameter or does that not work either?

tlambert03:
`show=True` also works... but I don't think we should give up on this or be wary... I definitely think it's the better approach to going through Qt.  but there's no hurry on figuring it out.

sofroniewn:
> show=True also works... but I don't think we should give up on this or be wary... I definitely think it's the better approach to going through Qt. but there's no hurry on figuring it out.

Agreed - am I right in thinking we need a quick PR to allow for `show=True` to be an option for people that want to do screenshot tests with introspection right now though? I could then use that in #1053. If so can you make a quick PR with a minimal introspection test like @GenevieveBuckley had in #1109 that would be great, then we can revisit the canvas render option at a longer timescale

tlambert03:
> am I right in thinking we need a quick PR to allow for show=True to be an option for people that want to do screenshot tests with introspection right now though?

no PR needed.. that's already an option.  Genevieve just needs to change the test as I described here https://github.com/napari/napari/issues/1105#issuecomment-608382286

goanpeca:
This PR is over a year old @tlambert03, wondering if this is something worth finishing :) ?

PR #1111, opened 2020-04-04
Allow add_path() to accept any layer-specific kwarg and rename to open_path()

### LINKED ISSUE #1107: Feature request: add_path() should accept kwargs like 'name', 'metadata', etc.
## 🚀 Feature

I would like the `add_path()` function to accept keyword arguments like `name`, `metadata`, etc.

## Motivation

I think the behaviour of `add_path()` should be as similar as possible to the other `add_` methods (`add_image()`, `add_points()`, `add_labels()`, etc.)

What do you think @tlambert03?

tlambert03:
# Description
This would close #1107 (allowing `Viewer.add_path` to accept kwargs like the other `add_*` methods) and would solve the conversation started in https://github.com/napari/napari/pull/937#issuecomment-581190625, in which we discussed deprecating the `path=` argument to specific `viewer.add_*` methods (while maintaining the option for layer-specific kwargs).  Actual deprecation warnings could be added here, or in a follow-=up PR.

Basic idea is to pass any received `kwargs` through to `_add_layers_with_plugins`, where they are used to overwrite the corresponding keys in the `meta` dict returned from plugins.  A new `prune_kwargs(kwargs, layer_type)` function ensures that the appropriate kwargs are passed to the `add_*` methods.

After this PR:
```python
viewer.add_path('mybundle.zip', image_name=['dapi', 'fitc'], points_name='spots', scale=(0.5, 1)) 
```

## Type of change

# How has this been tested?

- [x] test added for `prune_kwargs` function

## Final checklist:

### DIFF
--- a/napari/components/add_layers_mixin.py
+++ b/napari/components/add_layers_mixin.py
@@ -794,9 +794,16 @@ class AddLayersMixin:
         # add each layer to the viewer
         added: List[layers.Layer] = []  # for layers that get added
         for data in layer_data:
-            if kwargs and len(data) > 1:
-                assert isinstance(data[1], dict), '2nd item should be a dict'
-                data[1].update(kwargs)  # or should we update kwargs instead?
+            # if user provided kwargs, use to override the plugin meta dict
+            if kwargs:
+                if len(data) == 1:
+                    data = (data[0], kwargs)
+                elif len(data) > 1:
+                    assert isinstance(
+                        data[1], dict
+                    ), '2nd item should be a dict'
+                    # or should we update kwargs instead?
+                    data[1].update(kwargs)
             new = self._add_layer_from_data(*data)
             # some add_* methods return a List[Layer] others just a Layer
             # we want to always return a list
--- a/napari/components/add_layers_mixin.py
+++ b/napari/components/add_layers_mixin.py
@@ -822,10 +822,6 @@ class AddLayersMixin:
                 if len(data) == 1:
                     data = (data[0], valid_kwargs)
                 elif len(data) > 1:
-                    assert isinstance(
-                        data[1], dict
-                    ), '2nd item should be a dict'
-                    # or should we update kwargs instead?
                     data[1].update(valid_kwargs)
             # actually add the layer
             new = self._add_layer_from_data(*data)
--- a/napari/components/add_layers_mixin.py
+++ b/napari/components/add_layers_mixin.py
@@ -711,10 +711,10 @@ class AddLayersMixin:
         self.add_layer(layer)
         return layer
 
-    def add_path(
+    def open_path(
         self, path: Union[str, Sequence[str]], stack: bool = False, **kwargs
     ) -> List[layers.Layer]:
-        """Add a path or list of paths to the viewer.
+        """Open a path or list of paths with plugins, and add layers to viewer.
 
         A list of paths will be handed one-by-one to the napari_get_reader hook
         if stack is False, otherwise the full list is passed to each plugin
@@ -762,7 +762,7 @@ class AddLayersMixin:
     ) -> List[layers.Layer]:
         """Load a path or a list of paths into the viewer using plugins.
 
-        This function is mostly called from self.add_path, where the ``stack``
+        This function is mostly called from self.open_path, where the ``stack``
         argument determines whether a list of strings is handed to plugins one
         at a time, or en-masse.
 
--- a/napari/components/add_layers_mixin.py
+++ b/napari/components/add_layers_mixin.py
@@ -797,7 +797,8 @@ class AddLayersMixin:
         # add each layer to the viewer
         added: List[layers.Layer] = []  # for layers that get added
         for data in layer_data:
-            # if user provided kwargs, use to override the plugin meta dict
+            # if user provided kwargs, use to override any meta dict values
+            # that were returned by the plugin
             if kwargs:
                 layer_type = 'image' if len(data) < 3 else data[2]
                 valid_kwargs = prune_kwargs(kwargs, layer_type)
@@ -809,6 +810,7 @@ class AddLayersMixin:
                     ), '2nd item should be a dict'
                     # or should we update kwargs instead?
                     data[1].update(valid_kwargs)
+            # actually add the layer
             new = self._add_layer_from_data(*data)
             # some add_* methods return a List[Layer] others just a Layer
             # we want to always return a list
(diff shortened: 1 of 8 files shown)


GenevieveBuckley:
Hm, the update will need to happen earlier, before the layers get added to the viewer.

Here's an example where I pass the `name` kwarg (or you can also use `scale` and an image with non-isotropic voxels). You can see that the layer in the viewer doesn't take into account the keyword arguments.

```python
In [1]: %gui qt

In [2]: import napari

In [3]: viewer = napari.Viewer()

In [4]: viewer.add_image(path='/Users/genevieve/Documents/demo-data/confocal-ser
   ...: ies.tif', name='confocal')
Out[4]: <Image layer 'confocal' at 0x123ffafd0>
# This is what we expected, the name keyword argument has been respected

In [5]: viewer.add_path('/Users/genevieve/Documents/demo-data/confocal-series.ti
   ...: f', name='confocal 2')
Out[5]: [<Image layer 'Image' at 0x124a91710>]
# This is not what we want, you can see the name of the image layer is still the default
```

tlambert03:
oh woops... that's just an implementation error on my part.  The update _does_ happen before the layers get added, but I was only using the kwargs if the plugin _also_ returned a metadata dict (worked for the napari-dv example I tested) ... but the builtin reader doesn't use any meta dict, so it doesn't work.  will fix

GenevieveBuckley:
Other thoughts:
* Passing in kwargs that don't apply to a particular layer type should not raise an error. I think I'm leaning towards preferring continuing on silently instead of producing a warning, but it's not a strong preference.
* If you pass in a kwarg that doesn't apply to a particular layer type, it should not end up in the metadata dictionary. I'm pretty sure all kwargs have default values & so should already exist as keys in the dictionary, so it might work well to first check for the existence of a key & then update it. Otherwise this might be very confusing later.
* It'd be nice to include the same access to kwargs to the view_path convenience method too (I'm sure that's already on your list, you seem very thorough)

And if context is helpful for you: I ran into this while I was writing up an explanation of how to directly pass a URL to napari (because I forgot `viewer.add_image(path=url)` and tried `add_path(url)` first). It's not something that's causing a problem for anyone right now, but I did think it was worth mentioning.

tlambert03:
> * Passing in kwargs that don't apply to a particular layer type should not raise an error. I think I'm leaning towards preferring continuing on silently instead of producing a warning, but it's not a strong preference.
> * If you pass in a kwarg that doesn't apply to a particular layer type, it should not end up in the metadata dictionary. I'm pretty sure all kwargs have default values & so should already exist as keys in the dictionary, so it might work well to first check for the existence of a key & then update it. Otherwise this might be very confusing later.

this is also pretty easily done, we just need to change the logic in this [bit here](https://github.com/napari/napari/blob/8f9f3fb17ff92dc8d2f428d4300a036413bd4716/napari/components/add_layers_mixin.py#L797-L806).  For kwargs provided by the user (who doesn't technically know which types of layers will result from the path, since they are provided by plugins) I like the idea of pruning all the kwargs to layer-appropriate keys.  For plugins, I think we should leave the `unexpected keyword` error, since that actually implies an implementation problem.

> * It'd be nice to include the same access to kwargs to the view_path convenience method too (I'm sure that's already on your list, you seem very thorough)

yup, will do once we nail down the add_path API

> And if context is helpful for you: I ran into this while I was writing up an explanation of how to directly pass a URL to napari (because I forgot `viewer.add_image(path=url)` and tried `add_path(url)` first). It's not something that's causing a problem for anyone right now, but I did think it was worth mentioning.

yeah good to know, thanks.  FWIW, there has been discussion of deprecating the `add_image(path=...)` API, but we can't really do that until there is a way to pass all the other kwargs to the `add_image` method via add_path.  ... which this would solve!  so if this all makes it in, your explanation might need to change :)

tlambert03:
oh btw, I fixed your example if you want to try again

tlambert03:
ok, fixed all those,  so this example should work now, without erroring on the `invalid` kwarg:

```python
In [1]: import napari 
   ...: viewer = napari.Viewer() 
   ...: viewer.add_path('image_stack.tif', name='confocal', invalid='invalid')
Out[1]: [<Image layer 'confocal' at 0x132597c40>]
```
or, with `view_path`
```python
In [4]: napari.view_path('image_stack.tif', name='confocal', invalid='invalid', ndisplay=3)
Out[4]: <napari.viewer.Viewer at 0x1463e2e20>
```

tlambert03:
@sofroniewn, curious to get your thoughts on this too.  I think this would solve our conversation about deprecating the path argument to all the `add_*` methods in favor of using `add_path` with kwargs regardless of the final layer type: https://github.com/napari/napari/pull/937#issuecomment-581190625.

GenevieveBuckley:
Looking great! I like it a lot :)

Should I be seeing a warning (not an error) for the invalid keyword? I don't, but I think that's by design right now.

(It's super late so I'm going to go to sleep now, chat later)

tlambert03:
> Should I be seeing a warning (not an error) for the invalid keyword? I don't, but I think that's by design right now.

yeah, at the moment I just [prune layer-inappropriate kwargs without warning](https://github.com/napari/napari/blob/041c873d0109739f9ad7d001120aa6bcc03f310d/napari/components/add_layers_mixin.py#L912)

hard to say what the appropriate behavior is there... since one might conceivably add a path that has _both_ image and points layers (for example) while wanting to provide layer-specific kwargs for each, and we don't really want a bunch of warnings in that case.

actually, that also gives me an idea.  If you _do_ expect multiple layer types to result, and you want to, say, name the image and points layers separately, we could support optional kwargs with `<layer_type>_` prefix:

```python
viewer.add_path('mybundle.zip', image_name=['dapi', 'fitc'], points_name='spots') 
```

tlambert03:
after playing with it a bit.  I quite like this solution.  So I added tests, updated the origin description, and removed the draft status.  Ready for review.  Might be nice to include in `0.3.0` since it complements the plugins, but not urgent.

sofroniewn (approved):
This looks great to me! I'm approving - will wait for @jni to merge.

As discussed I think we can now remove the `path` arg in the individual layers ([like image](https://github.com/napari/napari/blob/master/napari/components/add_layers_mixin.py#L74)). We could either just remove it right away or add a deprecation warning for it to go in `0.3.1`. Curious what @jni thinks about the PR overall, removing the path arg in individual layers, and the need for a deprecation warning?

GenevieveBuckley:
> As discussed I think we can now remove the `path` arg in the individual layers ([like image](https://github.com/napari/napari/blob/master/napari/components/add_layers_mixin.py#L74)). We could either just remove it right away or add a deprecation warning for it to go in `0.3.1`. Curious what @jni thinks about the PR overall, removing the path arg in individual layers, and the need for a deprecation warning?

Changes here will affect PR 65 to the tutorials repository (which is ok, just flagging it here so we don't forget).

jni (changes requested):
Can we hold off on merging this? As far as I can tell there isn't an urgent need for it (?) before release, since we don't have reader plugins for other layers. I'm not convinced that this is the best course of action, and the changes here merit at least a discussion group meeting? I particularly dislike allowing invalid kwargs to pass through silently.

jni:
I just realised/remembered we *do* have reader plugins for any layer type. But the rest of my comment still stands I think... 😬

sofroniewn:
If we don't want to merge this immediately, but think we might want to eventually, then we might still want to deprecate or remove the `path` arg to `add_image` - I do think that's a big step in the right direction that this PR would allow, and I don't really like keeping something in the API unnecessarily longer that we know we will drop.

tlambert03:
Sure that’s fine, we were waiting for your input.  If you want to be stricter with invalid kwargs, I can see two options:
- most strict: track which kwargs get used in “prune_kwargs”, and raise if any are left over after all the layers are loaded
- less strict: make sure all kwargs are valid for at least one of the add_layer methods (even if they are not consumed)

The reason this is tricky is that we don’t know what types of layers we’re going to get when the user calls add_path

sofroniewn:
@tlambert03 thanks for making the function signature behave as we discussed in team meeting. Looking at this now I have three additional comments

1) I'd like to rename `add_path` to `open_path`. Looking at `save_layers` that is coming in #1104 I see a lot of symmetry between `open_path` and `save_layers`, and I've always found it confusing to have `add_path` sound like `add_image` and `add_shapes` etc. which I see as quite different.
2) I think we allow a `plugin_name` to be passed to `open_path`, like is done in #1104 `save_layers`. This can be added once #1104 has merged
3) I think we should drop the `path` argument from `add_image` and `add_labels`. First that `path` argument is only present on those two layers and not on all the others, and we have no plans to add it to the others, so it is quite strange that only those two have it (I can say this as I think it was my idea to add it there in the first place). When it was added it was necessary, but now that we have an `open_path` with the function signature above the exact functionality can be replicated by
```python
add_image(path='xyz.tif', colormap='red')
# is equivalent to
open_path(path='xyz.tif', colormap='red', plugin_name='builtin')
```
This means there are now two highly similar ways to do one thing, and the `0.3.0` release and this PR seems like the perfect time to fix this.

Thoughts @tlambert03 @jni?

GenevieveBuckley:
Just to check: `add_path` has been renamed to `open_path`, but the convenience method `view_path` still has the same name, right?

tlambert03:
yep, I left that as `view_path` since it's also opening a viewer instance.

tlambert03:
Ok, I changed `add_path` to `open_path`, and can add a `plugin=` argument in another PR after #1104.  Removing path from `add_image` and `add_labels` is a bit trickier since there's a subtle difference.  I started implementing this internally by having `add_image(path=...)` simply forward the path and kwargs to `open_path`, which works fine.  But for labels, since `open_path` doesn't have a way to _enforce_ that the layer type be something specific (`Labels`) or nothing at all (it is left up to the plugins), then users might often ask for `add_labels()` but get an `Image`.  which is nicht gut.  So I think we've got some more work to do before we can remove `path` from those two `add_*` methods.

napari/components/add_layers_mixin.py
```diff
@@ -804,6 +814,20 @@ def _add_layers_with_plugins(
...
+            # that were returned by the plugin
+            if kwargs:
+                layer_type = 'image' if len(data) < 3 else data[2]
+                valid_kwargs = prune_kwargs(kwargs, layer_type)
+                if len(data) == 1:
+                    data = (data[0], valid_kwargs)
+                elif len(data) > 1:
+                    assert isinstance(
+                        data[1], dict
+                    ), '2nd item should be a dict'
```
jni: Is this check necessary? Isn't it a guarantee of the type annotations that the 2nd item *will* be a dict?

jni: In fact, none of these lines inside `if kwargs` are covered by a test

jni (approved):
I think this is in a good spot! Just needs at least one more test. =)

napari/components/add_layers_mixin.py
tlambert03: since those are returned by plugins, seemed like there was no guarantee of anything, so I figured a helpful error message there would be better.  can remove though if you prefer.  and will add a test (actually, probably will be useful to have a general `mock_plugin_return` test pattern)

(the main logic is tested in `test_prune_kwargs.py` ... the rest comes down to what the plugins actually return)

tlambert03:
added tests.  merging now.  things for followup PRs:
- allow a plugin_name to be passed to open_path
- figure out how/whether to drop the path argument from `add_image` and `add_labels`

PR #1043, opened 2020-03-11
add github links to contributor list

boonepeter:
# Description

Addresses issue #995. I changed the script to keep track of users by login, then write out their name as the text with a link to their github profile.

## Type of change

- [x] documentation
# References

# How has this been tested?

## Final checklist:
- [x] My PR is the minimum possible work for the desired functionality
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #1113, opened 2020-04-05
Fix a number of sphinxdocs errors

tlambert03:
# Description
When you run `make docs`, sphinx will list any errors encountered during docs generation in red.  (they can come from anywhere: docstrings that are unparseable, errors in rst, etc).  Currently, we have a _lot_ of them.  This PR aims to fix as many as possible, and also adds a few niceties, like auto-linking of PR numbers in our [sphinx release notes](https://napari.org/docs/release/release_0_2_12.html), to the respective PRs.  

sidenote/reminder:  `codeblocks` in .rst require two backticks (`) instead of one. so I changed them in our release notes.

There are two remaining issues that are harder to fix, which I'm happy to leave till later:

- `hook_specifications.rst:2: WARNING: duplicate object description of napari.plugins.hook_specifications.napari_get_reader, other instance in api/napari.plugins, use :noindex: for one of them`
   - the problem here is that we'd like to have a special page for hook specifications, and apidocs don't give you granular control over indexing or not indexing specific functions.

- `CONTRIBUTING.md:40: WARNING: None:any reference target not found: ../pyproject.toml`
  - problem here is that we _move_ the devdocs when building them for sphinx, so relative links break. longer term solution is to pick whether we're writing docs for distribution on github or sphinx, and stick with one method.

### DIFF
--- a/docs/release/release_0_2_0.rst
+++ b/docs/release/release_0_2_0.rst
@@ -13,21 +13,20 @@ https://github.com/napari/napari
 
 New Features
 ------------
-
-- **Improved UI**, unifying buttons from controls, icons for layers, and more
-  understandable dimensions sliders
+- **Improved UI**, unifying buttons from controls, icons for layers,
+  and more understandable dimensions sliders
 - Add support for **3D rendering** for all our layer types
-- Add a ``Surface`` layer to render already generated meshes. Support nD meshes
+- Add a `Surface` layer to render already generated meshes. Support nD meshes
   rendered in 2D or 3D.
-- Add ``viewer.add_multichannel`` method to rapidly add expand a multichannel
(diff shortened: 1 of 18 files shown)


sofroniewn:
> sidenote/reminder: codeblocks in .rst require two backticks (`) instead of one. so I changed them in our release notes.

Given the merge on #1043 our release notes are going to be markdown moving forward. I was actually planning a dedicated PR just to convert our old ones to the new formatting so maybe best not to change them in this PR.

tlambert03:
undid those backtick changes (but I did leave a couple changes in there where newlines in a list were not indented)

sofroniewn (approved):
LGTM!

PR #1119, opened 2020-04-08
Fix miniconda download in CI

tlambert03:
# Description
noticed that tests are failing in #1069, see: https://github.com/napari/napari/pull/1069/checks?check_run_id=570525381

I believe that something might have changed with redirects on on the miniconda side of things.  By default, `curl` does not follow redirects, so this PR adds `curl -L` to follow url redirects.

## Type of change

- [x] test fix


sofroniewn (approved):
LGTM! Will merge so tests are fixed

PR #1122, opened 2020-04-09
make Qt component non private module

Czaki:
# Description
Move qt components to public module. Other software should not import from private modules, so ths changes simplify embedding napari viewer as widget, not as separate window.

This code still allow to import from `napar._qt` using custom Loader. 

## Type of change

- [ ] This change requires a documentation update (add qt modules documentation)

# References

# How has this been tested?

## Final checklist:

This PR still need changes in documentation which will be long. So I ask if this change is good for you?

### DIFF
--- a/napari/qt/__init__.py
+++ b/napari/qt/__init__.py
@@ -1,0 +1,4 @@
+from .._qt.qt_viewer import QtViewer
+from .._qt.qt_viewer_buttons import QtViewerButtons, QtNDisplayButton
+
+__all__ = ('QtViewer', 'QtViewerButtons', 'QtNDisplayButton')


sofroniewn:
@Czaki - which particular private Classes from the `_qt` folder are you using? If all you are using is 
```python
from napari._qt.qt_viewer import QtViewer
```
Then maybe we just want to expose `QtViewer` as an exception at the top level so you could do
```python
from napari import QtViewer
```
I'm not sure about exposing all the guts of the Qt, which I see as quite different from just allowing people easy access to the `QtViewer` so they can embed it somewhere else.

Also the further we go with the plugin infrastructure the more we will want to encourage people to add their own custom Qt widgets to the viewer as `DockWidgets`, rather than embedding the viewer in another app. See this answer here from @tlambert03 https://forum.image.sc/t/integration-of-napari-module-subclass-plugin/36018/2?u=sofroniewn

Would you be able to go down the `DockWidgets` approach to meet your needs instead of embedding the viewer? I'm curious what functionality you'd be missing and if we could work on adding that instead.

Czaki:
We talk about this already: https://forum.image.sc/t/using-napari-as-imagepys-canvas/32579/9?u=gbokota

In this moment QtViewer. but maybe more. 

My application has already designed layout, and few other assumption which conflict with napari layout. Also I need to have possibility to use at least 2 viewer simultaneously with possibility of synchronisation (for comparison of multiple variants)

Czaki:
One remark. I see that intersphinx is used in documentation. Some times ago I create plug-in to add references to qt documentation https://pypi.org/project/sphinx-qt-documentation/

tlambert03:
> One remark. I see that intersphinx is used in documentation. Some times ago I create plug-in to add references to qt documentation https://pypi.org/project/sphinx-qt-documentation/ 

Thanks!  Good to know!

Czaki:
Hm. It looks like I only need QtViewer. What do you think to export only this model? 

1. Import it to napari `__init__.py`
2. Create `qt` module which this class

?

Czaki:
> I'm not sure about exposing all the guts of the Qt, which I see as quite different from just allowing people easy access to the `QtViewer` so they can embed it somewhere else.

@sofroniewn Currently I need `QtViewer` and Buttons (`QtViewerPushButton` and `QtNDisplayButton`). I decide use buttons from napari to gave user same button for same action.

jni:
I'd be in favour of a top-level `napari.QtViewer`. `napari.QtViewerPushButton` is a bit messier though... So maybe yes, we should do `napari.qt.{WhateverWeWantToExposeFrom_qt}`. btw @Czaki have you seen [magicgui](https://magicgui.readthedocs.io/en/latest/). It might contain enough interaction elements for your needs to only use QtViewer...?

Czaki:
As you ca see even in magicgui  example it does not allow to add labels to control widget obraz so there is no information what control this widget (imagine few bool switch without labels) 

And my interfece is much more complicated 

Zrzut ekranu z 2020-05-13 09-44-51

And I have second part where I use two viewer: 

Zrzut ekranu z 2020-05-13 09-56-01

So I do not see other option than dock widget and control it from code.

tlambert03:
[this branch](https://github.com/napari/magicgui/tree/labels-forms) is adding labels to magicgui.  that said... I don't think you should be using magicgui, since you have plenty of Qt experience and a complicated app.

tlambert03:
Hey @Czaki, I think we're all in support of a new `napari.qt` module where we import the various objects from `napari._qt` that we want to expose.  Do you want to update this PR to undo all the changes that rename `napari._qt` to `napari.qt`, then just add a new `napari.qt` module that imports a couple things like the `QtViewer`?  e.g.:
```
# napari/qt.py
from ._qt.qt_viewer import QtViewer

__all__ = ['QtViewer']
```

Czaki:
flake8 fail looks unrelated to this PR. Maybe new version of flake.

tlambert03:
thanks, fixing in #1263

tlambert03 (approved):
thanks!  Anyone else want add anything?  maybe `qt_main_window.Window`?

sofroniewn (approved):
This is fine with me.

Probably no need to add window until someone request it @tlambert03  I don't think that is as likely though as window you can't really use to embed in something else like you can with the others.

Czaki:
Maybe This class should be somehow marked as public (in comment etc)? To other person know that it should not be changed drasticaly.

tlambert03:
thanks @Czaki, can you merge master to fix the flake8 stuff?

sofroniewn:
@tlambert03 I think I was wrong here, we should have added our `Window` class so that the doc string for https://github.com/napari/napari/blob/master/napari/_qt/qt_main_window.py#L353-L384 would appear on our docs website. My bad. Can you make a new PR to expose that, thanks!!

Issue #718, opened 2019-11-21
avoiding unnecessary io in dask + napari


tlambert03:
This really isn't a napari issue, but it will directly affect us and our consumption of `dask.delayed` arrays and/or dask arrays made by `dask-image`...  So just wanted to start a conversation here.

let's say you create a dask array from a folder of 600 3D tiff stacks:
`dask.array<stack, shape=(600, 65, 512, 512), dtype=uint16, chunksize=(1, 65, 512, 512)`

if you view that dask array in napari, things are as you would expect in the 3D viewer: it loads each file lazily as you scroll the 4th dim (e.g. time) slider.  However, if you go down to 2D view, and scroll the z-dimension slider, it loads the _full_ image for every plane.

I tried to fix that using [opportunistic caching](https://docs.dask.org/en/latest/caching.html), but that only fixes excess io when loading a full 3D stack `stack[0]` multiple times... but it is not smart enough to realize that if you've loaded `stack[0]` in the past, then you also already have `stack[0, 10]` in the cache.  I asked this question on stack overflow, but Matt Rocklin didn't have an immediate suggestion: https://stackoverflow.com/questions/58902595

so it might be worth considering creating our own [`cachey`](https://github.com/dask/cachey) `memoized`  reader internally...

sofroniewn:
I have noticed some of this stuff when I was doing some of the benchmarking here - https://github.com/napari/napari/issues/103#issuecomment-480640037 - I think there are some assumptions about usage that might be valid in an image viewer context that don't apply to dask generally and that we might want to bake in to our usage. Maybe it relates to the `Remote / Shareable` stuff more generally too

jni:
Thanks for raising this @tlambert03! It's kinda surprising to me that this doesn't work, and it certainly feels like we should be able to fix it. The idea is if an *internal* node of the graph is being frequently accessed, you should keep it in memory.

I guess `arr[0, 10]` is *not* evaluated as `arr[0][10]`, so it would result in a different computation from `arr[0, 11]`. I wonder if we changed our indexing to `[][][]` instead of `[,,]` it would make things better, even though it would pain me. =P

mrocklin:
Yeah, so what's happening is that you want to turn off how Dask fuses getitem optimisations like the following:

    x[0, :, :][:5] -> x[0, :5, :]

[...]

PR #1124, opened 2020-04-10
Monkeypatching dask to limit read events

tlambert03:
# Description
following the tips from @mrocklin in #718 this is mostly just a demonstration of how we could monkeypatch dask in order to prevent excess io with the way napari currently indexes nd data.  I'm not necessarily proposing we do this (it's a possibility I guess, but I'd of course like to hear from the dask folks whether this seems like a good short-term solution while we figure out whether this is an option that dask could support... or if there are hidden gotchas I'm missing here).

to test this, you can either make your own dask array of delayed array providers that "report" when they are called.  Or, you can use the llsfolder plugin I recently started making:  https://github.com/tlambert03/napari-llsfolder (follow instructions in the readme there to install the plugin, download sample data, and enable logging.DEBUG so you can watch the read/deskew tasks).

(there's probably a better way involving viewing the dask task graph too...)

### DIFF
--- a/napari/__init__.py
+++ b/napari/__init__.py
@@ -47,6 +47,7 @@ del app
 vispy_logger = logging.getLogger('vispy')
 vispy_logger.setLevel(logging.WARNING)
 
+from .utils import dask_monkeypatch  # must come before importing .viewer
 from .viewer import Viewer
 
 # Note that importing _viewer_key_bindings is needed as the Viewer gets
--- a/napari/layers/image/image.py
+++ b/napari/layers/image/image.py
@@ -508,7 +508,15 @@ class Image(IntensityVisualizationMixin, Layer):
                 ).transpose(order)
         else:
             self._transforms['tile2data'].scale = np.ones(self.dims.ndim)
-            image = np.asarray(self.data[self.dims.indices]).transpose(order)
+
+            if self.dims.ndim > 2:
+                outer_idx = tuple(self.dims.indices[: self.dims.ndim - 3])
+                image = np.asarray(self.data[outer_idx])
+                image = image[tuple(self.dims.indices[-3:])]
+            else:
+                image = np.asarray(self.data[self.dims.indices])
+
+            image = image.transpose(order)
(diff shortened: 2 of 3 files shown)


mrocklin:
Making Dask's optimization pipeline configurable is *definitely* something that is in-scope for the Dask project.  

I'll write up an issue with a loose plan and tag it for core maintenance.  Hopefully we can get someone working on it soonish.  If you all are around to help drive requirements (or do the actual work) then that would be awesome.

tlambert03:
Thank you @mrocklin!  I will definitely keep a close eye on whatever issue you write up, and will try to learn/contribute as much as possible

mrocklin:
https://github.com/dask/dask/issues/6083

jni:
I would 10x prefer for this to be solved on the dask side. Apart from my general and well-documented dislike of imports doing anything other than importing, actually modifying the behaviour of such a popular library as dask seems mighty risky to say the least! So, although I admire your chutzpah @tlambert03, I would rather take this as a wonderful demonstration rather than something that we would consider merging. :grimacing:

tlambert03:
yes, i agree!  just wanted a way to point to the specific bit of code that was responsible with a full working example.  we should all keep an eye on https://github.com/dask/dask/issues/6083 and https://github.com/dask/dask/pull/6084

mrocklin:
Here is another, perhaps simpler option.  https://github.com/dask/dask/pull/6087

tlambert03:
oh that looks perfect.  will pull and test locally with napari, and then comment on your PR.  thank you!

tlambert03:
yep.  dask/dask#6087 solves this (and, to state the obvious, infinitely more elegant 😂)

tlambert03:
going to close this and will open a new PR to take advantage of dask/dask#6087 once it merges

mrocklin:
That PR has merged.  You should be good to go.

tlambert03:
thanks again @mrocklin!

PR #1127, opened 2020-04-10
Enable volume rendering interpolation control

### LINKED ISSUE #1125: Different interpolation methods for 3D viewer
## 🚀 Feature
Just like for the 2d viewer it'd be nice to be able to choose interpolation methods for the 3d version

tlambert03:
# Description
This would close #1125 by exposing the 'nearest' interpolation option for volume visuals both on the model layer and in the Qt controls.  I changed the private `_interpolation` attribute to be a dict that stores the interpolation setting for both 2D and 3D rendering modes (so that it remembers settings when switching between `ndisplay`), but the external getter/setter API hasn't changed.

I haven't exposed the setting on the Labels Qt controls, since it looks so weird, but one can change it from the model (`labels_layer.interpolation = 'linear'`)

Untitled

## Type of change

# How has this been tested?

## Final checklist:


sofroniewn (approved):
LGTM! Thanks, will let sit for a day then merge

PR #1130, opened 2020-04-10
Refactor Vectors colors

kevinyamauchi:
# Description
This PR updates how the `edge_color` is stored in the `Vectors` layer model. Previously, the color was stored a single string of a color name and that color was applied to all vectors. Now, building on #782, we store the colors in an Nx4 numpy array with N, RGBA colors (one for each vector). On the view side, we are now setting colors via the `vertex_colors` and `face_colors`, which gives us the ability to color each vector independently (see below). 

**Creating Vectors layers**
As before, we can instantiate a layer with a single color name string

```python
layer = viewer.add_vectors(data, edge_color='red')
```

We can also give a single RGB array

```python
layer = viewer.add_vectors(data, edge_color=[1, 0, 0])
```

or an RGBA array
```python
layer = viewer.add_vectors(data, edge_color=[1, 0, 0, 1])
```

or an Nx4 array of RGBA colors for each vector. 

Apr-10-2020 21-04-21
<details><summary>Code</summary>
<p>

```python

import numpy as np
from skimage import data
from vispy.color import get_colormap
import napari


with napari.gui_qt():
    blobs = np.stack(
        [
            data.binary_blobs(
                length=128, blob_size_fraction=0.05, n_dim=3, volume_fraction=f
            )
            for f in np.linspace(0.05, 0.5, 10)
        ],
        axis=0,
    )
    viewer = napari.view_image(blobs.astype(float))

    # sample vector coord-like data
    n = 200
    pos = np.zeros((n, 2, 2), dtype=np.float32)
    phi_space = np.linspace(0, 4 * np.pi, n)
    radius_space = np.linspace(0, 20, n)

    # assign x-y position
    pos[:, 0, 0] = radius_space * np.cos(phi_space) + 64
    pos[:, 0, 1] = radius_space * np.sin(phi_space) + 64

    # assign x-y projection
    pos[:, 1, 0] = 2 * radius_space * np.cos(phi_space)
    pos[:, 1, 1] = 2 * radius_space * np.sin(phi_space)

    planes = np.round(np.linspace(0, 128, n)).astype(int)
    planes = np.concatenate(
        (planes.reshape((n, 1, 1)), np.zeros((n, 1, 1))), axis=1
    )
    vectors = np.concatenate((planes, pos), axis=2)

    # make colormap for edge colors
    cm = get_colormap('viridis')
    c_values = np.linspace(0, 1, n)
    edge_color = cm.map(c_values)

    # add the sliced vectors
    layer = viewer.add_vectors(
        vectors, edge_width=0.4, name='sliced vectors', edge_color=edge_color
    )

    viewer.dims.ndisplay = 3
```

</p>
</details>

## Type of change

# References

# How has this been tested?

## Final checklist:


kevinyamauchi:
I don't think I have the mapping of the face and vertex indices (`Vectors._view_faces`, `Vectors._view_vertices`) to the vector indices (i.e., `Vectors.data`) quite right. I am having trouble figure it out based on the code. I think the answer is [here](https://github.com/napari/napari/blob/391cfd5728bdf100f0c16deccb0c6df138e657aa/napari/layers/vectors/vectors.py#L273). @sofroniewn , I think this was introduced in #343, would you mind taking a look?

sofroniewn:
@kevinyamauchi some of the complexity comes here because there a three different representations of the data - one is length N, corresponding to vectors, another is length 2*N for the triangle faces, and the third is length 4*N for the triangle vertices when doing 2D imaging. For 3D imaging it is 4*N and 8*N respectively.

It looks like you are using both the `vertex_color` and `face_color` kwarg in vispy right now - that is not needed, you only need to use `face_color` if we only want to support one color per vector. If you wanted to have the ability for a single vector to go from red to blue say, then we'd need to use `vertex_color`, but you wouldn't actually ever need to use both at the same time (I think under-the-hood in vispy one will take precedent). So with that said I'd focus on `face_color`.

Now for the slicing you'll probably want to create a variable called something like `_view_indices` which will be used to slice both the `data` and the `edge_color` (like in the `Points` layer) - it shoukd be the arg to the list in this line https://github.com/napari/napari/blob/391cfd5728bdf100f0c16deccb0c6df138e657aa/napari/layers/vectors/vectors.py#L269

Once you have the sliced `edge_color` you'll need the repeat each color either twice or four times (if in 3D mode) and then use that result to set the `face_color` on the vispy mesh.

Let me know if this work / if you have further questions. Also, good idea to do this PR first and then come back to #898

kevinyamauchi:
Thanks for the quick response, @sofroniewn! I think we are mostly on the same page. I wanted to confirm with you the order of the faces for the 3D vectors. When I zoom in, it looks like there are two "lines" together with a "+" cross-section for each vector. Those two "lines" comprise 4 faces (as you mentioned). From [this line in the code](https://github.com/napari/napari/blob/391cfd5728bdf100f0c16deccb0c6df138e657aa/napari/layers/vectors/vectors.py#L273), it looks like if there are two vectors displayed, A and B, with indices i_A and i_B, respectively, the edge_colors (one color for each vector, Nx4) would indexed `[i_A, i_A, i_B, i_B, i_A, i_A, i_B, i_B]`. That is the two "lines" (each comprising two faces) are stacked on top of each other.

In the code, I implemented it as:

```python
# Get the face colors for the two faces comprising the vector
face_color = np.repeat(self.edge_color[self._view_indices], 2, axis=0)

if self.dims.ndisplay == 3:
    # if in 3D, there are two lines per vector, so we stack the indices    
    face_color = np.vstack([face_color, face_color])
```

Visually, it seems right, but I would like to make sure.
Screenshot 2020-04-11 at 08 24 26

Also, thanks for letting me know that the `MeshVisual` doesn't require both `vertex_color` and `face_color`! I didn't pick that up from reading the docs.

kevinyamauchi:
Sorry, one more question:

Why is the `MeshVisual` node given dummy data here? Is that just a vispy thing? If so, I think I will add a bit of code there to give a dummy face_color as well rather than have our model do that (i.e., our model will still give an array of length 0 when there is no data to display).

https://github.com/napari/napari/blob/391cfd5728bdf100f0c16deccb0c6df138e657aa/napari/_vispy/vispy_vectors_layer.py#L15-L22

sofroniewn:
> ```python
>    # if in 3D, there are two lines per vector, so we stack the indices    
>    face_color = np.vstack([face_color, face_color])
> ```

Yes that is correct, you can see the stacking here.
https://github.com/napari/napari/blob/391cfd5728bdf100f0c16deccb0c6df138e657aa/napari/layers/vectors/_vector_utils.py#L102
If you try a multicolor line example it should be clear, if you get it right then all four faces for each vector will have one color, if you get it wrong then you'll have vectors with half one color, half the other. In the future we want to make these 3D lines look nicer with more faces, but we can do that later.

> Why is the MeshVisual node given dummy data here? Is that just a vispy thing?

Yes this is a vispy thing, you can't pass vispy `None` here you have to pass it "empty arrays", feel free to solve in the model rather than here, it's a bit of a hack

Nice job with this, looks like it is coming together well

kevinyamauchi:
A few updates:

## Properties
I have added properties to the Vectors layer. If we are going to allow per-vector colors, I think we should have properties, because I think mapping values (e.g., angle) onto the color is probably a major use case. The API now looks like this:

As before, we can instantiate a layer with a single color name string

```python
layer = viewer.add_vectors(data, edge_color='red')
```

We can also give a single RGB array

```python
layer = viewer.add_vectors(data, edge_color=[1, 0, 0])
```

or an RGBA array
```python
layer = viewer.add_vectors(data, edge_color=[1, 0, 0, 1])
```

or an Nx4 array of RGBA colors for each vector. 

If we have a property we want to assign to each vector (e.g., the angle of the vector) and then map the vector color to it, we can do the following:
```python
properties = {'angle': vector_angles}
layer = viewer.add_vectors(data, edge_color='angles', properties=properties)
```

## GUI
I've updated the GUI so we can now select the edge color mode and the edge_color picker element changes depending on the mode. In direct mode, it's the standard color picker. In the cycle or colormap modes, it turns into a dropdown menu to select the property to map to the edge color. I made a small demo gif below where there is a property for the angle of the vectors (`angle`) and a property if angle > pi (`pos_angle`).

Apr-13-2020 17-46-15

@tlambert03, it would be great to get a bit of feedback on how I swap out the widgets in the grid layout ([here](https://github.com/napari/napari/blob/601df87c6c4c3978def82f393e5772e52186e6b3/napari/_qt/layers/qt_vectors_layer.py#L190)) when you have a chance. I am definitely not a Qt expert...

## selected_data
Do we want to expose an API for selecting sets of vectors and modifying their properties as we do in `Points` and `Shapes`? This adds a lot of complication and I am not sure how often people would want to do it. I see `Vectors` more as a view of data like an `Image` rather than an annotation of data like `Labels`, `Points`, and `Shapes`. What do you all think?


## To do
- discuss if we are happy with the API
- discuss if we are happy with the GUI update
- add tests

sofroniewn:
This is beautiful!! So nice @kevinyamauchi 

> I've updated the GUI so we can now select the edge color mode and the edge_color picker element changes depending on the mode. In direct mode, it's the standard color picker. In the cycle or colormap modes, it turns into a dropdown menu to select the property to map to the edge color.

This is great! Should we propagate this change back to `Points` once it is done merged here?

> Do we want to expose an API for selecting sets of vectors and modifying their properties as we do in Points and Shapes? This adds a lot of complication and I am not sure how often people would want to do it. I see Vectors more as a view of data like an Image rather than an annotation of data like Labels, Points, and Shapes. What do you all think?

I don't think we do - I agree this is more like image, there should be more of a push for performance rather than edit-ability, which can come in shapes.

Let me know when this is ready for full review! I like the API and GUI update functionality.

tlambert03:
this is super cool and also beautiful!  
without having looked too closely at the code, I have a question that might actually be relevant for points properties as well (and I apologize if it's already in there and I missed it):  would it be possible to support passing a callable as the value for a particular property (for instance, `vector_angles` as a function), which will take an item (vector) from the data and return a value (color)?

regarding the Qt usage.  the one thing I'm wondering is whether you need to call `replaceWidget` at all?  Isn't it enough just to hide/show whatever you want?  If that causes things to reorder in a way you don't like, you might be able to solve that by reordering them [here](https://github.com/napari/napari/blob/601df87c6c4c3978def82f393e5772e52186e6b3/napari/_qt/layers/qt_vectors_layer.py#L89-L103).

kevinyamauchi:
Thanks, @tlambert03 !
> without having looked too closely at the code, I have a question that might actually be relevant for points properties as well (and I apologize if it's already in there and I missed it): would it be possible to support passing a callable as the value for a particular property (for instance, vector_angles as a function), which will take an item (vector) from the data and return a value (color)?

We haven't implemented that yet and I think it's a nice suggestion! Especially if it could be done in a way that paved the way for lazy loading/evaluation of properties for large datasets. I think it's out of scope for this PR, but we should definitely consider it moving forward!

> regarding the Qt usage. the one thing I'm wondering is whether you need to call replaceWidget at all? Isn't it enough just to hide/show whatever you want? If that causes things to reorder in a way you don't like, you might be able to solve that by reordering them here.

Ah, I didn't realize the grid layout would "collapse" when the widgets in a row are hidden!  🤦 I've changed it to that and removed the calls to `replaceWidget()`. Thanks!

@sofroniewn , it sounds like we are aligned on not enabling selection/setting properties of subsets of vectors. I've dropped that functionality, which definitely reduces complexity!

> This is great! Should we propagate this change back to Points once it is done merged here?

Yes, I think we should! I plan to circle back to Points after this to do some of the code duplication reduction we've talked about since that will also be useful in the `Shapes` refactor and I think it will be easier to work it out in the `Points` layer.

> Let me know when this is ready for full review! I like the API and GUI update functionality.

I have a bit more cleanup to do before this is ready to review (mostly documentation). I think it should be ready sometime on Tuesday your time.

kevinyamauchi:
I think this is ready for the first round of review! If you'd like to take it for a spin, you can try `examples/add_vectors_color_by_angle.py`.

There is a particular part of the interactivity that I could use some feedback on:

- when the user switches the color mode to either cycle or colormap and no color property had been previously selected, the model automatically chooses the first property in `properties`. A warning is emitted to the terminal, but there's nothing in the GUI that says what was done. I think that is the right behavior, but it may be nice to somehow show that warning in a way that is more obvious. We could use the status bar at the bottom, but my eyes are usually elsewhere when interacting with the canvas/controls and I usually miss what is displayed there.
- when the Vectors layer is in the 'colormap' mode and the user selects a property that is not really "colormappable" (e.g., the property is has string values), an error is emitted and the colormode gets switched to the color cycle mode (which does work). An alternative would be to "reject" the color property change and remain in color map mode. Thoughts?

In both cases, it seems like it would be nice if we had a better way of displaying error messages on the GUI (I think this is inline with some of @justinelarsen's feedback). Obviously out of the scope of this PR, but maybe some sort of "pop down" (_definitely_ not pop up 😉  ) banner notification like on a phone?

edit: I just saw #1090. I think that would address my point above.

sofroniewn:
> when the Vectors layer is in the 'colormap' mode and the user selects a property that is not really "colormappable" (e.g., the property is has string values), an error is emitted and the colormode gets switched to the color cycle mode (which does work). An alternative would be to "reject" the color property change and remain in color map mode. Thoughts?

hmm yeah - this is a little confusing, i think the behaviour in the GUI is ok, in some senses things like this I think people just need to play around with to see what is natural, and i wouldn't overthink it now

napari/components/add_layers_mixin.py
```diff
@@ -682,12 +686,26 @@ def add_vectors(
...
+            where N is the number of vectors.
         edge_width : float
             Width for all vectors in pixels.
         length : float
              Multiplicative factor on projections for length of all vectors.
         edge_color : str
             Edge color of all the vectors.
+        edge_color_cycle : np.ndarray, list, cycle
+            Cycle of colors (provided as RGBA) to map to edge_color if a
+            categorical attribute is used to set face_color.
```
sofroniewn: maybe better to say `to color the vectors` - it is a little confusing to reference `face_color` here if one isn't familiar with the vispy api - which people shouldn't need to be when reading our model files

napari/view_layers.py
```diff
@@ -701,12 +705,26 @@ def view_vectors(
...
+        where N is the number of vectors.
     edge_width : float
         Width for all vectors in pixels.
     length : float
          Multiplicative factor on projections for length of all vectors.
     edge_color : str
         Edge color of all the vectors.
+    edge_color_cycle : np.ndarray, list, cycle
+        Cycle of colors (provided as RGBA) to map to edge_color if a
+        categorical attribute is used to set face_color.
```
sofroniewn: see comment above about `face_color`

napari/layers/vectors/vectors.py
```diff
@@ -21,12 +36,26 @@ class Vectors(Layer):
...
+        where N is the number of vectors.
     edge_width : float
         Width for all vectors in pixels.
     length : float
          Multiplicative factor on projections for length of all vectors.
     edge_color : str
         Edge color of all the vectors.
+    edge_color_cycle : np.ndarray, list, cycle
+        Cycle of colors (provided as RGBA) to map to edge_color if a
+        categorical attribute is used to set face_color.
```
sofroniewn: see comment above about `face_color`

sofroniewn (approved):
Looks great @kevinyamauchi!! Such nice functionality. I bet @bryantChhun will like it too!!

napari/components/add_layers_mixin.py
kevinyamauchi: That makes sense. I will update here and in the other places you flagged.

kevinyamauchi:
Thanks for the review, @sofroniewn ! I have made the documentation changes you suggested.

sofroniewn:
I'm going to merge now - looks good. I do think once we've refactored shapes we'll want to probably make a mixin for the "color" and "properties" code at least (across points, shapes, and vectors) but we can do that then. For now press on @kevinyamauchi!

PR #1135, opened 2020-04-11
convert old release notes to md

sofroniewn:
# Description
Following #1043 this PR converts all our old release notes from `rst` to `md`, adjusts headers to be the same, and puts contributors names in with the new formatting. It changes none of the content of the old release notes only fixes these formatting issues so that from now on release notes will be consistent.

I still have to update a few contributor names in some of our old releases, will do that before merge (hit a rate limit on the github api right now), but this is otherwise ready for review.

We might need to adjust our docs so that these render correctly. Maybe @kne42 or @tlambert03 can help with that.


sofroniewn:
Github handles have now been updated, this is ready for review

docs/release/generate_release_notes.py
```diff
@@ -125,7 +125,7 @@ def add_to_users(users, new_user):
 highlights['Highlight'] = {}
 highlights['New Feature'] = {}
 highlights['Improvement'] = {}
-highlights['Bugfix'] = {}
+highlights['Bug Fixe'] = {}
```
kne42: typo

kne42 (approved):

docs/release/generate_release_notes.py
sofroniewn: good catch, thanks!

sofroniewn:
Typo fixed! Will merge after tests pass

PR #1140, opened 2020-04-14
Reduce code duplication in Points

### LINKED ISSUE #1136: Uncovered lines in points properties code
In this code report:

https://codecov.io/gh/napari/napari/pull/1069/diff

these lines are not covered:

https://github.com/napari/napari/blob/6ea95a9fa2f9150a4dbb5ec1286b8ff2020c3957/napari/layers/points/points.py#L565-L575

and

https://github.com/napari/napari/blob/6ea95a9fa2f9150a4dbb5ec1286b8ff2020c3957/napari/layers/points/points.py#L606-L608

(The "ifs" are executed but never evaluate to True in tests.) It's possible that the second one in particular will *never* be hit, since we automatically convert input to arrays. So the code could possibly be removed.

kevinyamauchi:
# Description
This PR aims to reduce some of the code duplication in the Points layer that was introduced when adding properties and allowing colors to be mapped to properties. The general approach is to write general functions that can be used for both edge_color and face_color properties using `setattr` and `getattr`.

I also added some tests to increase coverage on the Points model (closes #1136).

## Type of change

# References

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/layers/points/_tests/test_points.py
+++ b/napari/layers/points/_tests/test_points.py
@@ -469,7 +469,7 @@ def test_adding_properties(attribute):
     assert isinstance(layer.properties['point_type'], np.ndarray)
 
     # removing a property that was the _edge_color_property should give a warning
-    setattr(layer, '_%s_color_property' % attribute, 'vector_type')
+    setattr(layer, f'_{attribute}_color_property', 'vector_type')
     properties_2 = {
         'not_vector_type': np.array(['A', 'B'] * int(shape[0] / 2))
     }
@@ -649,9 +649,9 @@ def test_switch_color_mode(attribute):
     properties = {'point_type': np.array([0, 1.5] * int((shape[0] / 2)))}
     initial_color = [1, 0, 0, 1]
     color_cycle = ['red', 'blue']
-    color_kwarg = '%s_color' % attribute
-    colormap_kwarg = '%s_colormap' % attribute
-    color_cycle_kwarg = '%s_color_cycle' % attribute
+    color_kwarg = f'{attribute}_color'
+    colormap_kwarg = f'{attribute}_colormap'
+    color_cycle_kwarg = f'{attribute}_color_cycle'
     args = {
         color_kwarg: initial_color,
         colormap_kwarg: 'gray',
@@ -659,36 +659,36 @@ def test_switch_color_mode(attribute):
     }
     layer = Points(data, properties=properties, **args,)
 
-    layer_color_mode = getattr(layer, '%s_color_mode' % attribute)
-    layer_color = getattr(layer, '%s_color' % attribute)
+    layer_color_mode = getattr(layer, f'{attribute}_color_mode')
+    layer_color = getattr(layer, f'{attribute}_color')
     assert layer_color_mode == 'direct'
     np.testing.assert_allclose(
         layer_color, np.repeat([initial_color], shape[0], axis=0)
     )
 
     # there should not be an edge_color_property
-    color_property = getattr(layer, '_%s_color_property' % attribute)
+    color_property = getattr(layer, f'_{attribute}_color_property')
     assert color_property == ''
 
     # transitioning to colormap should raise a warning
     # because there isn't an edge color property yet and
     # the first property in Vectors.properties is being automatically selected
     with pytest.warns(UserWarning):
-        setattr(layer, '%s_color_mode' % attribute, 'colormap')
-    color_property = getattr(layer, '_%s_color_property' % attribute)
+        setattr(layer, f'{attribute}_color_mode', 'colormap')
+    color_property = getattr(layer, f'_{attribute}_color_property')
     assert color_property == next(iter(properties))
-    layer_color = getattr(layer, '%s_color' % attribute)
+    layer_color = getattr(layer, f'{attribute}_color')
     np.testing.assert_allclose(layer_color[-1], [1, 1, 1, 1])
 
     # switch to color cycle
-    setattr(layer, '%s_color_mode' % attribute, 'cycle')
-    color = getattr(layer, '%s_color' % attribute)
+    setattr(layer, f'{attribute}_color_mode', 'cycle')
+    color = getattr(layer, f'{attribute}_color')
     layer_color = transform_color(color_cycle * int((shape[0] / 2)))
     np.testing.assert_allclose(color, layer_color)
 
     # switch back to direct, edge_colors shouldn't change
-    setattr(layer, '%s_color_mode' % attribute, 'direct')
-    new_edge_color = getattr(layer, '%s_color' % attribute)
+    setattr(layer, f'{attribute}_color_mode', 'direct')
+    new_edge_color = getattr(layer, f'{attribute}_color')
     np.testing.assert_allclose(new_edge_color, color)
 
 
@@ -701,7 +701,7 @@ def test_colormap_without_properties(attribute):
     layer = Points(data)
 
     with pytest.raises(ValueError):
-        setattr(layer, '%s_color_mode' % attribute, 'colormap')
+        setattr(layer, f'{attribute}_color_mode', 'colormap')
 
 
 @pytest.mark.parametrize("attribute", ['edge', 'face'])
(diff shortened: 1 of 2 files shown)


napari/layers/points/points.py
```diff
@@ -964,40 +893,121 @@ def face_color_mode(self):
...
+            setattr(self, '_%s_color' % attribute, colors)
+            setattr(self, '_%s_color_mode' % attribute, ColorMode.DIRECT)
+
+            color_event = getattr(self.events, '%s_color' % attribute)
+            color_event()
+
     def refresh_colors(self, update_color_mapping: bool = False):
         """Calculate and update face and edge colors if using a cycle or color map
 
+                Parameters
```
jni: This is over-indented.

jni: Twice over. :joy:

napari/layers/points/points.py
```diff
@@ -964,40 +893,121 @@ def face_color_mode(self):
...
-        if face_color_mode == ColorMode.DIRECT:
-            self._face_color_mode = face_color_mode
-        elif face_color_mode in (ColorMode.CYCLE, ColorMode.COLORMAP):
-            if self._face_color_property == '':
+    def _set_color_mode(
+        self, color_mode: Union[ColorMode, str], attribute: str
+    ):
+        """ Set the face_color_mode or edge_color_mode property
+
+        Paramters:
```
```suggestion
        Parameters
```

napari/layers/points/points.py
```diff
@@ -964,40 +893,121 @@ def face_color_mode(self):
...
                     'selected property must be numeric to use ColorMode.COLORMAP'
                 )
-            self._face_color_mode = face_color_mode
+            setattr(self, '_%s_color_mode', color_mode)
             self.refresh_colors()
 
+    def _set_color(self, color: ColorType, attribute: str):
+        """ Set the face_color or edge_color property
+
+        Paramters:
```
```suggestion
        Parameters
```

jni (commented):
@kevinyamauchi I've only been able to quickly skim but on first pass this looks great! Let us know when it's out of WIP. =)

napari/layers/points/points.py
kevinyamauchi: 🤦 !

kevinyamauchi:
Thanks for the initial comments, @jni! I think this is ready for full review!

Also, the tests need a similar refactor, but I will do that in a separate PR, so we aren't changing the tests and code at the same time.

sofroniewn (approved):
Looks good to me! Nice deduplication!!

napari/layers/points/_tests/test_points.py
```diff
@@ -445,6 +445,38 @@ def test_properties():
...
+    layer.properties = properties
+    np.testing.assert_equal(layer.properties, properties)
+
+    # add properties as a dataframe
+    properties_df = pd.DataFrame(properties)
+    layer.properties = properties_df
+    np.testing.assert_equal(layer.properties, properties)
+
+    # add properties as a dictionary with list values
+    properties_list = {'point_type': ['A', 'B'] * int(shape[0] / 2)}
```
jni:
```suggestion
    properties_list = {'point_type': ['A', 'B'] * (shape[0] // 2)}
```

Having said this, I feel this is a bit brittle, and will only work when `shape[0]` is even. I think the "correct" way here would be to use itertools:

```python
from itertools import cycle, islice

point_type = list(islice(cycle('AB'), None, shape[0]))
```

Or:

```python
from itertools import cycle
import toolz as tz

point_type = list(tz.take(shape[0], cycle('AB')))
```

jni: And you make this multiple times in this test so might as well make it once (as a list) and transform it to `np.array` where needed.

napari/layers/points/points.py
```diff
@@ -426,6 +379,52 @@ def __init__(
...
+        self, color: ColorType, attribute: str
+    ):
+        """Initialize the current_edge_color or current_face_color
+        property for the case when you are starting with an empty layer
+
+        Parameters:
+        -----------
+        color : (N, 4) array or str
+            The value for setting edge or face_color
+        attribute : str
```
```suggestion
        attribute : str in {'edge', 'face'}
```

napari/layers/points/points.py
```diff
@@ -426,6 +379,52 @@ def __init__(
         # Trigger generation of view slice and thumbnail
         self._update_dims()
 
+    def _initialize_current_color_for_empty_layer(
+        self, color: ColorType, attribute: str
+    ):
+        """Initialize the current_edge_color or current_face_color
+        property for the case when you are starting with an empty layer
```
```suggestion
        """Initialize current_{edge,face}_color when starting with empty layer.
```

napari/layers/points/points.py
```diff
@@ -473,85 +472,66 @@ def data(self, data: np.ndarray):
...
+                self._add_point_color(adding, 'face')
 
                 self.size = np.concatenate((self._size, size), axis=0)
                 self.selected_data = set(np.arange(cur_npoints, len(data)))
 
         self._update_dims()
         self.events.data()
 
+    def _add_point_color(self, adding: int, attribute: str):
+        """Add the edge or face colors for new points
```
```suggestion
        """Add the edge or face colors for new points.
```

napari/layers/points/points.py
```diff
@@ -473,85 +472,66 @@ def data(self, data: np.ndarray):
...
+
+        Parameters:
+        ----------
+        adding : int
+            the number of points that were added
+            (and thus the number of color entries to add)
+        attribute : str
```
```suggestion
        attribute : str in {'edge', 'face'}
```

napari/layers/points/points.py
```diff
@@ -964,35 +873,93 @@ def face_color_mode(self):
...
+        if color_mode == ColorMode.DIRECT:
+            setattr(self, '_%s_color_mode' % attribute, color_mode)
+        elif color_mode in (ColorMode.CYCLE, ColorMode.COLORMAP):
+            color_property = getattr(self, '_%s_color_property' % attribute)
+            if color_property == '':
                 if self.properties:
-                    self._face_color_property = next(iter(self.properties))
+                    setattr(
+                        self,
+                        '_%s_color_property' % attribute,
```
```suggestion
                        f'_{attribute}_color_property',
```

Also, since you use this multiple times in this function, it's probably worth creating it once, e.g. `color_property_attr = f'_{attribute}_color_property'` then using that each time. But maybe not worth it, as the string is more explicit. Either way, use f-strings. =)

napari/layers/points/_tests/test_points.py
```diff
@@ -608,6 +640,100 @@ def test_n_dimensional():
     assert layer.n_dimensional is True
 
 
+@pytest.mark.parametrize("attribute", ['edge', 'face'])
+def test_switch_color_mode(attribute):
+    """Test switching between color modes"""
+    shape = (10, 2)
+    np.random.seed(0)
+    data = 20 * np.random.random(shape)
+    properties = {'point_type': np.array([0, 1.5] * int((shape[0] / 2)))}
```
jni: Perhaps this should be `np.random.random`, since it's continuous values?

jni (approved):
@kevinyamauchi good stuff! I don't want to hold up merge, so I'm approving, but there are some code style changes I'd like to see before merging:

- the way the properties lists are defined in the tests is a bit brittle, I've suggested a more generic approach (bonus: works with any number of discrete choices, not just two)
- when an input can only have a small possible set of values, typically these are listed in the type annotation in the docstring.
- use f-strings rather than `%s`.

napari/layers/points/points.py
kevinyamauchi: My preference is to write the string each time, because as you said, it's more explicit. However, I will change to f-strings.

napari/layers/points/_tests/test_points.py
kevinyamauchi: Thanks for the suggestion. I've implemented it as a function using the islice method. That way we can make property arrays of different lengths, if needed.

napari/layers/points/_tests/test_points.py
kevinyamauchi: That's a good point. I've updated it so there is a continuous property (using `np.random.random`) and a categorical property.

kevinyamauchi:
Thank you for the reviews, @sofroniewn and @jni ! I have addressed your comments. I will let this sit until tomorrow to see if there are any other comments and then merge!

Next up: reduce duplication in points tests.

PR #1147, opened 2020-04-15
Add keybinding to toggle selected layer visibility

### LINKED ISSUE #1139: Toggle current layer visibility w/keyboard shortcut
## 🚀 Feature

I'd like to introduce a new keyboard shortcut to toggle visibility of current layer

## Motivation

When working with a few layers, I often find myself wanting to toggle the visibility of a particular layer to look clearly at one beneath. This has been coming up most recently when painting labels.

## Pitch

I'd like to propose a new keyboard shortcut to toggle the visibility of the current layer. 

This part is up for debate, but if this is very useful for people, it could be a single key, like `H`, or a combo like `Ctrl+H`, or some other key(s) entirely. 

From looking at the other key bindings, I assume this would be straightforward to do, unless there is some `Qt` issue I'm not familiar with.

## Alternatives

At the moment, I use the mouse to manually toggle the visibility.

hectormz:
# Description

This adds a function to toggle selected layer visibility along with a key binding (`Ctl+H`)

## Type of change

# References

This PR (partially) closes #1139

# How has this been tested?

- [X] Added test to select some of layers with heterogeneous visibility and toggle the visibility back and forth

## Final checklist:

### DIFF
--- a/napari/components/layerlist.py
+++ b/napari/components/layerlist.py
@@ -176,6 +176,6 @@ class LayerList(ListModel):
 
     def toggle_selected_visibility(self):
         """Toggle visibility of selected layers"""
-        for i in range(len(self)):
-            if self[i].selected:
-                self[i].visible = not self[i].visible
+        for layer in self:
+            if layer.selected:
+                layer.visible = not layer.visible
--- a/napari/components/layerlist.py
+++ b/napari/components/layerlist.py
@@ -173,3 +173,9 @@ class LayerList(ListModel):
                 self[selected[0] - 1].selected = True
         elif len(self) > 0:
             self[0].selected = True
+
+    def toggle_selected_visibility(self):
+        """Toggle visibility of selected layers"""
+        for i in range(len(self)):
+            if self[i].selected:
+                self[i].visible = not self[i].visible
--- a/napari/_viewer_key_bindings.py
+++ b/napari/_viewer_key_bindings.py
@@ -142,7 +142,7 @@ def play(viewer):
         viewer.window.qt_viewer.dims.play(axis)
 
 
-@Viewer.bind_key('Control-H')
+@Viewer.bind_key('V')
 def toggle_selected_visibility(viewer):
     """Toggle visibility of selected layers"""
     viewer.layers.toggle_selected_visibility()
--- a/napari/_viewer_key_bindings.py
+++ b/napari/_viewer_key_bindings.py
@@ -140,3 +140,9 @@ def play(viewer):
     else:
         axis = viewer.window.qt_viewer.dims.last_used or 0
         viewer.window.qt_viewer.dims.play(axis)
+
+
+@Viewer.bind_key('Control-H')
+def toggle_selected_visibility(viewer):
+    """Toggle visibility of selected layers"""
+    viewer.layers.toggle_selected_visibility()
--- a/napari/components/_tests/test_layers_list.py
+++ b/napari/components/_tests/test_layers_list.py
@@ -393,3 +393,34 @@ def test_move_selected():
         False,
         False,
     ]
+
+
+def test_toggle_visibility():
+    """
+    Test toggling layer visibility
+    """
+    layers = LayerList()
+    layer_a = Image(np.random.random((10, 10)))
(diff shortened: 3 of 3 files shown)


tlambert03:
just a quick note: `Control-H` (on a mac at least, and I thought also windows?) is reserved for hiding the entire napari application.  So we'll probably need to pick a slightly different combo.
("control" here actually maps to the "command" key on macs)

hectormz:
@tlambert03 Good to know, I wasn't aware of that. That's not the case on Windows (I think). From my own laziness, I think a one or two key shortcut works well. How do you feel about just `H`?

tlambert03:
I personally kinda like hotkeys without any modifiers, like just `H`, however those can be complicated in programs that also accept text (imagine you're editing the name of a layer and suddenly it disappears).  It's probably doable, but I believe that would be our first un-modified hotkey... so we'd be wading into new territory with lots of testing to do.

hectormz:
Good point. I didn't think about that. I did a really quick test and changed it to `H`. Nothing was triggered when editing layer names and typing `H` or in file dialog, but I'm not sure all the corner cases that `napari` might have.

Other options might be `Ctrl-V` (for visible) or `Ctrl-I` (for invisible), but I think those are better served with potential pasting, or showing layer info

tlambert03:
keypress events are handled by the active widget, and then (sometimes) [cascade up through the parent widgets](https://doc.qt.io/qt-5/eventsandfilters.html#event-handlers), depending on how the specific widget and its key event methods and event filters were implemented.  So I'm not surprised it worked on the LayerList widget... but like you say, we'd need to think about every widget (or someone would need to read up on event filters and figure out whether we're safe to have it on the main viewer).

I also like short key combos ... but unfortunately there are only so many letters we get :) i agree that ctrl-v and i are better served by paste and info.  so might need to go with an unintuitive letter, or add another modifier (like ctrl-shift-V)

hectormz:
@tlambert03 I agree. May be there should be a roadmap for future keybindings so we don't waste a good one now 😂. My last thought unless others have an idea is `Ctrl-D` for `Display/Hide`

tlambert03:
What about alt instead of ctrl?  Alt-V?

jni:
I'm +1 on unmodified shortcuts, and advocated a lot for them early on. Alt is annoying, not many commands (or any?) work that way on Mac, without Cmd also pressed.

Since visibility is easy to revert, I don't think it's a huge risk to accidentally hide a layer. It's obvious in the UI what has happened, and easy to toggle back.

sofroniewn:
After discussion with @jni, @tlambert03 we propose to just use the unmodified `V` key. If you make the change @hectormz then this PR should be good to go

You can see all the keybindings we currently have in the `Show key bindings` dialog in the help menu.

hectormz:
@sofroniewn I updated to `V`. Thanks to @tlambert03  and @jni for weighing in

napari/components/layerlist.py
```diff
@@ -173,3 +173,9 @@ def select_previous(self, shift=False):
                 self[selected[0] - 1].selected = True
         elif len(self) > 0:
             self[0].selected = True
+
+    def toggle_selected_visibility(self):
+        """Toggle visibility of selected layers"""
+        for i in range(len(self)):
+            if self[i].selected:
+                self[i].visible = not self[i].visible
```
sofroniewn:
```suggestion
        for layer in self:
            if layer.selected:
                layer.visible = not layer.visible
```
Minor request to change, but because this class subclasses a list you can actually just iterate right through it, making the code slightly more readable

hectormz: 👌 good point!

sofroniewn (approved):
Looks great to me!! Thanks @hectormz

jni (approved):

tlambert03 (approved):

jni:
Just played with this locally and oh man, I love it! 😃😃😃😃 Thanks @hectormz !

hectormz:
@jni awesome! `napari` made it very easy to add the functionality. I've been enjoying it more than I would too!

PR #1148, opened 2020-04-17
Only block in gui_qt or quit QApp when necessary

tlambert03:
# Description
This PR solves two issues that have bothered me for a long time!

### Problem `#1`

sometimes `gui_qt` blocks unnecessarily (for instance, if a qt event loop has already been started by IPython, then our context manager actually doesn't need to block at all).  Fixing this allows you to quickly load any of our examples in IPython, keeping all of the variables in scope.

```python
> ipython

In [1]: %gui qt

# in the current master, this command would block
# (unnecessarily, since we already have an event loop provided by IPython,
# AND you can't use the console in napari since it was started from IPython)
In [2]: %run examples/add_points_with_properties.py

# after this PR, you can keep working with all of the variables
# that were declared in the file/example.
In [3]: viewer
Out[3]: <napari.viewer.Viewer at 0x13ce1adf0>

In [4]: points_layer
Out[4]: <Points layer 'Points' at 0x143c54940>
```

### Problem `#2`

If you already have an event loop running (via IPython or otherwise) if you QUIT the application (rather than closing the main window) you actually can't create another window without restarting the app (which usually just means restarting the IPython session).

```python
> ipython

In [1]: %gui qt

In [2]: import napari

In [3]: v = napari.Viewer()
# if you close this viewer window using the close button on the main window
# or using command-W (on mac) ... then it's fine.

# but if you QUIT the program (command-Q or File -> Exit), then there is
# no more event loop and this command just hangs ...
In [4]: v = napari.Viewer()

# until you actually create a new event loop
In [5]: from qtpy.QtWidgets import QApplication

In [6]: QApplication.exec_()  # now it shows up.
```

after this PR, the second instantiation on line 4 would show the Viewer.

## Type of change

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/_qt/event_loop.py
+++ b/napari/_qt/event_loop.py
@@ -22,17 +22,20 @@ def gui_qt(*, startup_logo=False):
     IPython with the Qt GUI event loop enabled by default by using
     ``ipython --gui=qt``.
     """
+    had_app = True
     app = QApplication.instance()
     if not app:
+        had_app = False
         # if this is the first time the Qt app is being instantiated, we set
         # the name, so that we know whether to raise_ in Window.show()
         app = QApplication(sys.argv)
         app.setApplicationName('napari')
-    if startup_logo:
-        logopath = join(dirname(__file__), '..', 'resources', 'logo.png')
-        splash_widget = QSplashScreen(QPixmap(logopath).scaled(400, 400))
-        splash_widget.show()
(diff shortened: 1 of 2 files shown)


sofroniewn (approved):
This looks good to me! Will let @jni weigh in before merge though

PR #1149, opened 2020-04-17
Remove path arg from add_image and add_labels

tlambert03:
# Description
As proposed in https://github.com/napari/napari/pull/1111#issuecomment-612955351 and discussed in meeting, this removes the `path` arg from `add_image(path='path')` (and `add_labels`) in favor of `open_path('path', layer_type='image')`.  The ability to pass a plugin to `open_path` requires some of the work in #1104, so we'll either need to extract that into a different PR, or wait to merge this.

**UPDATE**: everything I need to keep working on this is now in #1153 

## Type of change

# References

# How has this been tested?

## Final checklist:


sofroniewn (approved):
Beautiful! I really like how clean this is. This looks good to go from me. Is there anything else it needs from your perspective @tlambert03 or this this good for a final review from @jni?

tlambert03:
I think it’s ready.  The one thing I wondered is how we feel about having `view_path` but `viewer.open` (no path).   Do we want to stick with that?

sofroniewn:
> The one thing I wondered is how we feel about having view_path but viewer.open (no path). Do we want to stick with that?

I like `viewer.open`, I agree `view_path` is not ideal, but I can live with it. I could also live with `open_path`, I could also live with dropping `view_path` entirely.

jni (approved):
This LGTM, but @sofroniewn we need to make sure that the release notes offer very specific instructions for how to replace `viewer.add_image(path=...)` and `viewer.add_labels(path=...)` calls in the new version.

jni:
> I could also live with dropping view_path entirely.

Me too.

tlambert03:
> > I could also live with dropping view_path entirely.
> 
> Me too.

really??
so to just open tiff with napari is always going to require a multistep process of first creating a viewer?

sofroniewn:
> so to just open tiff with napari is always going to require a multistep process of first creating a viewer?

Hmm this part of the API has long been an issue for us. Not sure if we want to try and fix it now for `0.3.0` maybe we just live with `view_path` for now and then fix all of this with `0.4.0` if we move to `viewer.layers.add_image` (though that has it's own complexity).

Do you have a recommendation @tlambert03 on how to proceed?

tlambert03:
Other than considering different names for `napari.view_path`, I don't have any suggestions... but I don't think we should remove it entirely.  I think it's pretty important to be able to open a file with one command.  (it forms the basis of our own CLI after all... so you can imagine others might want something like that).

sofroniewn:
We can wait on dropping/ renaming `view_path` - will merge once tests pass

PR #1152, opened 2020-04-17
Fix black consistency

tlambert03:
# Description
We've noticed a couple cases recently where CI black let some trailing commas through, but my (and @jni's) local pre-commit was unhappy and tried to fix them.  This PR just does a couple things that I hope will prevent that:

- pins the version of black used in our `.pre-commit-config.yaml` and `requirements/development.txt`
- changes `pass_filenames: false` to `true` in `pre-commit-config`, which should make it only look at staged files.

note, I don't think it should be necessary for our CI to maintain consistency, but you may want to run `pip install -r requirements/development.txt` after merging this PR into your local environments.

## Type of change

- [x] CI Bug-fix

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,32 +4,6 @@ repos:
     hooks:
     - id: black
       pass_filenames: true
-      args: [
-        '--target-version', 'py36',
-        '--target-version', 'py37',
-        '--target-version', 'py38',
-        '--skip-string-normalization',
-        '--line-length', '79',
-        '--exclude', '''
-        (
-          /(
-              \.eggs
-            | \.git
-            | \.hg
-            | \.mypy_cache
-            | \.tox
-            | \.venv
-            | _build
-            | buck-out
-            | build
-            | dist
-            | examples
-            | vendored
-          )/
-          | napari/resources/qt.py
-        )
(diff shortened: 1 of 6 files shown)


sofroniewn (approved):
Makes sense to me - i had noticed that with the trailing commas too

jni (approved):
Woot woot. =) Does this mean we should change those trailing commas now?

tlambert03:
Well, I'm still a little confused about that actually.  for consistency, I'll remove them... but here's what black says about [trailing commas](https://black.readthedocs.io/en/stable/the_black_code_style.html#trailing-commas):

> One exception to adding trailing commas is function signatures containing *, *args, or **kwargs. In this case a trailing comma is only safe to use on Python 3.6. Black will detect if your file is already 3.6+ only and use trailing commas in this situation. If you wonder how it knows, it looks for f-strings and existing use of trailing commas in function signatures that have stars in them. In other words, if you’d like a trailing comma in this situation and Black didn’t recognize it was safe to do so, put it there manually and Black will keep it.

jni:
I kinda hate them when inline...

jni:
Thinking of forking black... =P

tlambert03:
lol.  omg

PR #1153, opened 2020-04-17
Plugin _HookCaller and registration updates

tlambert03:
# Description
This PR pulls some stuff out of #1104 that was not specific to writer plugins, so we can evaluate it separately.  #1104 will depend on this, (and this will also make it easier to keep working on plugin exception handling without waiting on #1104).

The core change here is the removal of the `execute_hook` function, the rationale was explained in https://github.com/napari/napari/pull/1104#issuecomment-612880642, but to summarize:

- This simplifies things by making _all_ hook calls go through `plugin_manager.hook.name_of_hookspec(**kwargs)` rather than _sometimes_ using `execute_hook(plugin_manager.hook.name_of_hookspec, **kwargs)`, and adding a couple conveniences to the _HookCaller.__call__ method (like the ability to call a specific plugin or fail).  
- This also adds lazy discovery of plugins, which is now triggered the first time an attribute of `plugin_manager.hook` is accessed.
- I am overriding the `PluginManager.register` method from pluggy so we no longer need to wait for the next pluggy release to use the `@napari_hook_implementation(specnam=...)` feature (provided developers depend on `napari-plugins` and not pluggy).

A lot of the diff is due to moving code from `plugins.manager.py` to `_hook_callers.py`

## Type of change

- [x] Refactor
- [x] New feature (sortof, in the ability to use `specname` argument)

# References

# How has this been tested?

- [x] I pulled in some of the tests from pluggy (in test_hook_callers) so we can make sure these changes still work with their tests.  and edit their tests as needed.

## Final checklist:


napari/plugins/_hook_callers.py
```diff
@@ -177,51 +165,321 @@ def _multicall(
...
+        """
+        self.get_plugin_implementation(plugin_name).enabled = enabled
+
+    def enable_plugin(self, plugin_name: str):
+        """enable implementation for ``plugin_name``."""
+        self._set_plugin_enabled(plugin_name, True)
+
+    def disable_plugin(self, plugin_name: str):
+        """disable implementation for ``plugin_name``."""
+        self._set_plugin_enabled(plugin_name, False)
```
tlambert03: this whole stretch is just moved from manager.py

napari/plugins/_tests/test_hook_callers.py
```diff
@@ -0,0 +1,468 @@
...
+from pluggy import HookCallError, HookspecMarker, PluginValidationError
+from pluggy.hooks import HookImpl
+
+from napari.plugins import _hook_callers, PluginManager
+from napari_plugins import HookimplMarker
+
+hookspec = HookspecMarker("example")
+hookimpl = HookimplMarker("example")
+
+
```
tlambert03, line 11: these tests are mostly from pluggy, but import from our local overrides (`_hook_callers` and `PluginManager`) instead.

napari/plugins/manager.py
```diff
@@ -341,6 +214,67 @@ def _register_module(
         except Exception as exc:
             raise PluginRegistrationError(plugin_name, module_name) from exc
 
+    def register(self, plugin: ModuleType, name=None):
```
tlambert03, line 217: vendored from pluggy, but contains the changes in my PR that allow the `specname` argument to be used by developers.

sofroniewn (commented):
Good idea to pull this out directly. It looks good to me, but I'll let @jni do a more thorough review

napari/plugins/_hook_callers.py
```diff
@@ -177,51 +165,321 @@ def _multicall(
...
+            ) from exc
+
+    def __call__(
+        self,
+        *,
+        _skip_impls: Optional[Sequence[HookImpl]] = None,
+        _return_impl: bool = False,
+        _plugin: Optional[str] = None,
+        **kwargs,
+    ):
```
jni: I'm not in love with functions changing the return type based on kwargs. @tlambert03 and I discussed on Zulip and came up with:

- `.__call__` -> return only the result
- `.call_with_impl` -> return an object containing the result, the implementation used, and other metadata.

This echoes some discussions we had at scikit-image e.g. [here](https://github.com/scikit-image/scikit-image/issues/3892#issuecomment-591819705) and in these [meeting notes](https://github.com/scikit-image/meeting-notes/blob/master/2020/2020-02-27--BIDS-core-dev-meeting.markdown) (search for "API policies")

jni: `_skip_impl` and `_plugin` are fine.

jni: Also: I wonder whether we want to support returning of a list. This was a pluggy thing, but I don't know whether we will have hookspecs for which we want to call *all* implementations. What do you think @tlambert03? (Just trying to simplify the return type to always be consistent...)

napari/plugins/_hook_callers.py
```diff
@@ -177,51 +165,321 @@ def _multicall(
...
+              ``firstresult==True``, in which case a single result will be
+              returned.
+            - If ``_return_impl`` is ``True``, will return a list of 2-tuples
+              ``(result, HookImpl)``, unless the hookspec was declared with
+              ``firstresult==True``, in which case a single 2-tuple will be
+              returned.
+            - If ``_plugin`` is provided, will return the single result from
+              the specified plugin
+        """
+        assert not self.is_historic()
```
jni: what's the story here?

napari/plugins/_tests/test_hook_callers.py
```diff
@@ -0,0 +1,468 @@
+import pytest
+from pluggy import HookCallError, HookspecMarker, PluginValidationError
+from pluggy.hooks import HookImpl
+
+from napari.plugins import _hook_callers, PluginManager
+from napari_plugins import HookimplMarker
+
+hookspec = HookspecMarker("example")
```
jni: I'd rename these `example_hookspec` and `example_implementation`

napari/plugins/_tests/test_hook_callers.py
```diff
@@ -0,0 +1,468 @@
...
+hookimpl = HookimplMarker("example")
+
+
+def MC(methods, kwargs, firstresult=False):
```
jni: MC?

napari/plugins/_tests/test_hook_callers.py
```diff
@@ -0,0 +1,468 @@
...
+    caller = _hook_callers._multicall
+    hookfuncs = []
+    for method in methods:
+        f = HookImpl(None, "<temp>", method, method.example_impl)
+        hookfuncs.append(f)
+    return caller(hookfuncs, kwargs, firstresult=firstresult)
+
+
+def test_call_passing():
+    class P1(object):
```
jni: Plugin1 etc

napari/plugins/_tests/test_hook_callers.py
jni: We can leave these as-is for now, but if so, I'd like an issue created to naparify these tests at some point. I think the pluggy devs loved short names far too much.

jni (approved):
@tlambert03 I've only reviewed the non-vendored changes, and they look good, except for the `_return_impl` thing which you have under control I think. =) So I'm approving pending that change.

tlambert03:
> I wonder whether we want to support returning of a list. This was a pluggy thing, but I don't know whether we will have hookspecs for which we want to call all implementations. What do you think @tlambert03? (Just trying to simplify the return type to always be consistent...)

I can imagine cases in which we want to call all plugins (imagine a hook that can add something to the menubar, or "provide" something like a command line option) so I don't want to get rid of that entirely.  However, to avoid different returns, we could make the whole `firstresult` thing be its own `.call_firstresult_only` call:  perhaps the whole _multicall loop could actually be a generator, and `firstresult` is just `next(i for i in self())` or something like that.  Anyway,  I'd like to do that on a different PR... but I do think there are improvements that could be made.

tlambert03:
I made the main change requested from @jni (namely: to have a consistent return type from `_hook_callers._multicall`) by "owning" the `_Result` object that pluggy was previously using behind the scenes and making it a little more useful (it can now store the `HookImpl` instances responsible for each result _and_ any `PluginCallErrors` that occurred, which I will use in later PRs to clean up the exception handling).  Calling the hook_caller directly (`napari.hook.spec_name(**kwargs)`) still returns a result or list of results as always, but calling `napari.hook.spec_name.call_with_result_obj(**kwargs)` returns the `HookResult` object itself.  That keeps the return type independent of kwargs, and gives us a place to store metadata about the hook loop.

sofroniewn (approved):
This looks good to me, thanks for making those changes @tlambert03 - I will merge and press on with #1104

PR #1155, opened 2020-04-17
Fix sphinx

### LINKED ISSUE #1134: developer resources have disappeared from website
## 🐛 Bug

Developer resources have disappeared from website

image

tlambert03:
# Description
fixes #1134 
I also took the opportunity to update to sphinx 3.0.1 (and pin it), also clean formatting on all of the markdown (using a markdownlint).  Things like:

- there should be a line between headers and the body of a section
- fenced code should have a blank line separating it, and language specified.
- wrapped text to 80 chars

also:
- cleaned up look of TOCs on the left
- made sure releases are shown in order (reverse from newest to oldest)
- fixed releases for markdown by adding recommonmark extension.

NOTE: there are still a _lot_ of errors when building the docs, and it has to do with using `Attributes` in the docstring of a class for an `@property` that also has a setter.  It also renders incorrectly.  (see https://napari.org/docs/api/napari.layers.vectors.html)
This is not new, but the warnings are new.  there is an issue about this... https://github.com/sphinx-doc/sphinx/issues/7427  but we might also need to change our code (in a followup RP)

## Type of change

# References

# How has this been tested?

## Final checklist:


kne42 (approved):
LGTM

sofroniewn (approved):
LGTM!! Thanks

PR #1138, opened 2020-04-13
automate release process

kne42:
# Description
Automate the release process using GitHub Actions.

Upon pushing a tag beginning with `v` to `napari/napari`, the CI will:
- copy the release notes to the tag
- build the distribution
- upload the distribution to the appropriate release on GitHub
- upload the distribution to PyPI

PR #1156, opened 2020-04-17
change release notes source file to use .md ext

kne42:
fixes a bug in #1138 because I forgot to account for #1135


sofroniewn (approved):
Looks good to me! thanks!!

Issue #1052, opened 2020-03-14
Remaining pyramid bugs


sofroniewn:
## 🐛 Bug
I've been closing out some of our older pyramid and big data issues, (i.e. #736) to try and make it clear what issues still remain and what are solved. `contrast_limits` computation is now in good shape (please open fresh issues if it is not). Behavior with `is_pyramid=False` is in good shape, but if `is_pyramid` is not specified we can still auto-compute one (like in #1050) which can expose unsolved problems with the pyramid tiling clipping (see #458).

#737 was a WIP PR to address the pyramid guessing, but should be superseded by a new PR. I'll work on a fresh PR to fix the clipping in #458 now that using progress with our transforms / world coordinates.

sofroniewn:
This is closed by #1159

PR #1159, opened 2020-04-18
Drop pyramid autogeneration

sofroniewn:
# Description
This PR drops support for pyramid autogeneration. Which was causing unexpected behavior and bugs. 

There is still more refactoring that could be done with our pyramid code and a rename to `multiscale` that has been discussed, but that can come in follow-up PRs. We could also improve some of our error handling (make more clear error messages when people try and do things that don't make sense) but again I think that can come later.

## Type of change


sofroniewn:
This will close #1052

napari/layers/image/_image_utils.py
```diff
@@ -1,191 +1,51 @@
...
-    if max_layer is None:
-        max_layer = np.floor(np.log2(np.max(data.shape))).astype(int) + 1
+    ndim = len(shape)
+    last_dim = shape[-1]
 
-    zoom_factor = np.divide(1, downscale)
+    if ndim > 2 and last_dim < 5:
+        return True
+    else:
+        return False
```
jni:
```suggestion
    return ndim > 2 and last_dim < 5
```

napari/layers/image/image.py
```diff
@@ -184,11 +200,18 @@ def __init__(
...
         if self.is_pyramid:
-            self._data_level = len(data_pyramid) - 1
+            self._data_level = len(self.data) - 1
+            big_levels = [
+                np.any(np.greater_equal(p.shape, 2 ** 6 - 1)) for p in data
+            ]
+            if np.any(big_levels):
+                self._thumbnail_level = np.where(big_levels)[0][-1]
+            else:
+                self._thumbnail_level = 0
```
jni: can you explain this in a comment? I don't understand what this code is doing. What's a big level?

napari/layers/image/image.py
```diff
@@ -483,24 +499,27 @@ def _set_view_slice(self):
                 * self._transforms['tile2data'].scale
             )
 
-            image = np.asarray(
-                self._data_pyramid[level][tuple(indices)]
-            ).transpose(order)
+            image = np.asarray(self.data[level][tuple(indices)]).transpose(
+                order
+            )
```
```suggestion
            image = np.transpose(
                np.asarray(self.data[level][tuple(indices)]), order
            )
```

napari/layers/image/image.py
```diff
@@ -483,24 +499,27 @@ def _set_view_slice(self):
...
+                / self.downsample_factors[self._thumbnail_level, not_disp]
             )
             downsampled_indices = np.round(
                 downsampled_indices.astype(float)
             ).astype(int)
             downsampled_indices = np.clip(
-                downsampled_indices, 0, self.level_shapes[-1, not_disp] - 1
+                downsampled_indices,
+                0,
+                self.level_shapes[self._thumbnail_level, not_disp] - 1,
```
jni, line 522: I thought all this stuff was going to get simplified/unified in this PR? ;)

jni (commented):
@sofroniewn So great. =) I've made some code suggestions, they're quite minor, but also requested a clarification about "big levels".

One thing that would be nice would be to catch VisPy's error message if an image is too big to be displayed, and direct users to how to create an image pyramid using zarr. That can come later, though. We could make an issue for it both here and in `napari/tutorials`.

napari/layers/image/image.py
sofroniewn: Ah yeah, I'm saving that for another PR - turned out not to be necessary (I thought more would change in this one) so worth making that clean up a self contained thing

sofroniewn:
> One thing that would be nice would be to catch VisPy's error message if an image is too big to be displayed, and direct users to how to create an image pyramid using zarr. That can come later, though. We could make an issue for it both here and in napari/tutorials.

This will happen here https://github.com/napari/napari/blob/78b4e9413cb2c916d0f74ac56e14fb83534608f4/napari/_vispy/vispy_image_layer.py#L68-L78
and is unchanged by this PR - we will still emit a warning here https://github.com/napari/napari/blob/78b4e9413cb2c916d0f74ac56e14fb83534608f4/napari/_vispy/vispy_image_layer.py#L172-L177, we could add a tutorial on this, but that can come later.

I will add clarification around `big_levels` now

napari/layers/image/image.py
sofroniewn: comment made. This is to find a good level to use for thumbnail generation - not too small so that you just say have a 2x2 pixel image for the thumbnail (which might be the top of the pyramid)

jni: I might call that a goldilocks level. ;)

jni (approved):
@sofroniewn approving this so you can self-merge, but I would rename `big_levels` something like `big_enough_levels` or `goldilocks_levels`. Also the end of the comment is a bit garbled, "and so containting being very blurred", could use one more edit. =) But otherwise this is ready to go! 🎉

sofroniewn:
Renamed `big_enough_levels` and fixed comment - will merge after tests pass

PR #1160, opened 2020-04-18
Rename requirements test

sofroniewn:
# Description
Renames the test checking our requirements to be more in line with our other tests

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -225,7 +225,7 @@ codecov_task:
     pip install -e .
     pip install pytest-cov codecov
 
-  test_script:
+  check_requirements_script:
     pip-missing-reqs napari --ignore-file="**/_tests/*" --ignore-module=PySide2 --ignore-module=zarr  --ignore-file=napari/conftest.py
 
   coverage_script: |
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -23,7 +23,7 @@ lint_task:
     - flake8 napari
 
 requirements_task:
-  name: Check if all requirements are listed
+  name: Check Requirements
   install_script:
     - python -m pip install --retries 3 --upgrade "pip<19.3"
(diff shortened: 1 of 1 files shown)


sofroniewn:
Should this test also have `auto_cancellation: true` too? Ever since it has been added I see weird things in the tests rendering on github - for example now I see this 

image

instead of seeing all 9(?) checks. Does anyone know what is causing this?

tlambert03:
I do think it should have auto cancelation, but I can’t think why that would be causing that.  I’ll try to look into it tomorrow if it’s still happening

tlambert03:
Honestly I’d prefer it wasn’t its own test, but unfortunately it sounds like that check_reqs thing requires an older version of pip.  It’s probably fine to put it in the flake8 test anyway though.

sofroniewn:
Added `auto_cancellation: true`

tlambert03:
I moved the pip-check-reqs test into the coverage test.  Since it requires installing everything, it's not super fast, but it's also something we don't need to check up front everytime, and fits well in the coverage tests (i.e. make sure our requirements.txt covers our dependencies).
I also chose to install `pip-check-reqs` not from pypi, but rather from [this specific PR](https://github.com/r1chardj0n3s/pip-check-reqs/pull/22).  I know that's a little weird/brittle, but I think requiring and older pip is also brittle, and if that package doesn't merge that PR (3 months old without a comment) then I'm not sure it's a good long term solution anyway.  Since the coverage tests require all the others to pass, at the very least it won't bug us until the final step.

tlambert03 (approved):

sofroniewn:
That works for me - thanks for the fix!

sofroniewn:
I'm merging now as still getting some weird stuff with tests rendering in GitHub, maybe unrelated, but would be nice to see

PR #1163, opened 2020-04-22
Color cycles properties return array

kevinyamauchi:
# Description
As discussed in #1104 , this PR changes the color cycle properties to only be set by lists a numpy arrays of colors and the property returns the same way. This change is to enable serialized layers to be be written out in a human-readable format.

## Type of change

# References

# How has this been tested?

## Final checklist:


napari/layers/points/points.py
```diff
@@ -696,23 +696,15 @@ def edge_color(self, edge_color):
 
     @property
     def edge_color_cycle(self):
-        """Union[list, np.ndarray, cycle] :  Color cycle for edge_color.
-        Can be a list of colors or a cycle of colors
+        """Union[list, np.ndarray] :  Color cycle for edge_color.
```
tlambert03: maybe while we're at it we can turn these into proper return annotations?

```python
    def edge_color_cycle(self) -> Union[list, np.ndarray]:
        """Color cycle for edge_color.

        Can be a list of colors defined by name, RGB or RGBA.
        """
```

(same with other similar functions)

napari/layers/utils/color_transformations.py
```diff
@@ -74,20 +74,18 @@ def transform_color_cycle(
     -------
     transformed : cycle
         cycle of Nx4 numpy arrays with a dtype of np.float32
+    transformed_color_cycle: np.ndarray
```
tlambert03: does the return annotation need to be updated here?

tlambert03 (approved):
thanks @kevinyamauchi!

napari/layers/utils/color_transformations.py
kevinyamauchi: Good point. I also think the variable names were confusing so I updated them.

napari/layers/points/points.py
kevinyamauchi: I think it makes sense (it's been on the to do list). I can work on it later today!

kevinyamauchi: Done!

kevinyamauchi:
Thanks for the quick review, @tlambert03 ! @sofroniewn , would you mind taking a look to make sure this matches what you need for the #1104 ?

sofroniewn (approved):
LGTM! Thanks for the rapid turn around on this one @kevinyamauchi! Will give @jni the option to review and then merge!

jni (approved):

PR #1164, opened 2020-04-22
Allow last added point to be deleted with backspace keybinding

### LINKED ISSUE #1161: Consistent behavior when deleting points with hotkeys
## 🚀 Feature
When adding points, it would be nice to be able to use the 'delete' key to delete points. 

Behavior should be consistent. Current behavior when on a points layer:
- Selecting a point and using the delete(mac) or backspace(windows), deletes the selected point.
- If a point is selected and I click the 'Delete selected points' button (x), the selected point is deleted.

However, if I am in the adding points mode and I use the delete/backspace key, nothing happens even though the last created point is selected.

## Pitch

The last created point should be deleted after pressing the delete/backspace key.

kevinyamauchi:
# Description
This PR adds the ability to delete the last added point with backspace keybinding when in the point adding mode.

## Type of change

# References
This PR closes #1161 .

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/_viewer_key_bindings.py
+++ b/napari/_viewer_key_bindings.py
@@ -75,6 +75,7 @@ def focus_axes_down(viewer):
 
 
 @Viewer.bind_key('Control-Backspace')
+@Viewer.bind_key('Control-Delete')
 def remove_selected(viewer):
     """Remove selected layers."""
     viewer.layers.remove_selected()
@@ -87,6 +88,7 @@ def select_all(viewer):
 
 
 @Viewer.bind_key('Control-Shift-Backspace')
+@Viewer.bind_key('Control-Shift-Delete')
 def remove_all_layers(viewer):
     """Remove all layers."""
     viewer.layers.select_all()
--- a/napari/layers/points/_points_key_bindings.py
+++ b/napari/layers/points/_points_key_bindings.py
@@ -60,6 +60,7 @@ def select_all(layer):
 
 
 @Points.bind_key('Backspace')
+@Points.bind_key('Delete')
 def delete_selected(layer):
     """Delet all selected points."""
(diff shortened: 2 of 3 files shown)


kevinyamauchi:
@justinelarsen, would you mind taking a look and seeing if this is what you had in mind?

justinelarsen (approved):
Looks good to me!

GenevieveBuckley:
Does it make sense to also bind the `Delete` key here, as well as `Backspace`?

kevinyamauchi:
> Does it make sense to also bind the Delete key here, as well as Backspace?

I am definitely open to it! Just to make sure we're on the same page: when you say `Delete`, are you referring to the key that removes characters to the right of the cursor? I always get confused because mac keyboards call the key that most keyboards list as "backspace" as "delete" 😆

GenevieveBuckley:
> 
> 
> > Does it make sense to also bind the Delete key here, as well as Backspace?
> 
> I am definitely open to it! Just to make sure we're on the same page: when you say `Delete`, are you referring to the key that removes characters to the right of the cursor? I always get confused because mac keyboards call the key that most keyboards list as "backspace" as "delete" 😆

Yes, that's the key I'm talking about. 

imo, that (possible) confusion is an argument for making both delete/backspace do the same thing here, but if someone has an argument against it that would also be good to hear.

jni (approved):

justinelarsen:
Yes, I think it makes sense to bind the 'delete' (forward delete) key as well.

sofroniewn:
I have now bound the "forward-delete" key too in all places where we have bound "backspace" these function the same now. At some point we could add a test to ensure they are always used the same this if we felt it was important. I think we can wait on that for now.

I will merge after tests pass

Issue #1019, opened 2020-03-03
Support for multiresolution (pyramid) layers besides Image layers


GenevieveBuckley:
## 🚀 Feature

1. Move everything needed for multiresolution/pyramid layers into VispyBaseLayer. Currently we have things split across VispyImageLayer and VispyBaseLayer, eg: the `find_top_left()` method, which depends on the `level_shapes` property (which is only available in the image layer).
2. (Possibly) change the name of `is_pyramid` to `is_multiresolution`.

## Motivation

The motivation here is that we'd be in a better position to support multiresolution display for layer types besides Image layers. 
Eg: points layers with *lots* of points maybe don't need all of them rendered if you're zoomed out a lot, etc.

## Pitch

This isn't a new pitch, it's more documentation of an idea that's been floating around for a while.

https://github.com/napari/napari/pull/1015#issuecomment-593254516 by @jni

> I don't remember where this concept came up last week, but we had the idea that `is_pyramid` would become `is_multiscale` and would apply across all layers (eg millions of points would get grouped into a single point if you zoom out enough, just like this happens with some mapping applications, for example). So, I think a lot of that logic should definitely go on the base layer.

sofroniewn:
Thanks for making this issue!! I like this idea, it would maybe be good to have a concrete example of say many many points where napari currently cannot handle it, but where having a multiresolution approach would help. Maybe @kevinyamauchi has ideas / experience there

d-v-b:
This seems like a good idea; additionally, it might be nice to force all data to be represented internally as multiscale, even if only 1 scale level (full resolution) is used / calculated. This would remove a lot of the current `is_pyramid` checks.

sofroniewn:
Noting the discussion in https://github.com/zarr-developers/zarr-specs/issues/50 it looks like their preferred term would be `is_multiscale` and I see that is what @jni used above. Do we prefer that or `is_multiresolution` which I think I had used (without much thought)? I'm fine with either but would like us to match them if we can.

d-v-b:
to be super pedantic, "multiscale" is more implementation-agnostic than "multiresolution". The resolution of actual images can be decoupled from the sampling density -- two images can have the same resolution, but different sampling densities. "multiscale" more clearly describes the space in which the image is embedded. Also multiscale is a little shorter :)

sofroniewn:
[...]

PR #1170, opened 2020-04-25
replace pyramid with multiscale

sofroniewn:
# Description
This PR will follow on from https://github.com/napari/napari/issues/1019#issuecomment-598986933 and replace `pyramid` by `multiscale` and move the `is_multiscale` kwarg onto the base layer.

It is a breaking change due `is_pyramid` -> `is_multiscale` now inside `add_image` and `add_labels`, but probably good to bite the bullet now for `0.0.3`.

We've also dropped the pyramid autogeneration so hopefully less people will need to use `is_pyramid=False` as a safe guard now, and we've improved some of the lazy handling.

One question for @jni and others is if we want to add any warning/ error message if people try and pass `is_pyramid` that tells them to change to `is_multiscale` instead?

This is ready for review.

## Type of change


jni (commented):
@sofroniewn This LGTM, with the exception that now that we are not autogenerating, we might want to rename `is_multiscale` to simply `multiscale`. iirc we went with `is_pyramid` because we were offering autogeneration, so `pyramid=True` could be interpreted as "this is a pyramid" *or* "generate a pyramid for me". Currently we have just `rgb=True` so I think `multiscale=True` is more consistent, and nicer.

sofroniewn:
>  Currently we have just rgb=True so I think multiscale=True is more consistent, and nicer.

Love it - this is done! This is now ready for final review.

I'll also make clear `is_pryamid` -> `multiscale` in the release notes

jni (approved):

PR #1171, opened 2020-04-25
add napari-svg plugin support

sofroniewn:
# Description
This PR pulls out all of the svg generating code from napari and adds a dependency to [napari-svg](https://github.com/napari/napari-svg). Write now there are a couple changes to the previous behavior - now only 2D data will be saved instead of taking the currently viewed portion of nD data.

We don't support colors for shapes yet - waiting on #898 to merge

We don't support non-vispy colormaps for images, which means we also don't color labels properly.

We don't support surfaces, we didn't before this PR either though, so no change there.

These can all be addressed in time.

To use the plugin you can now do
```python
viewer.layers.save(path, plugin='svg')
```
or
```python
layer.save(path, plugin='svg')
```
To save all the layers or a single layer to an svg.

Tests have been added for all layer supported types.

## Type of change


jni (approved):
Love it!

jni:
@sofroniewn the only thing I would add is keeping the example, but using the plugin to save.

sofroniewn:
> @sofroniewn the only thing I would add is keeping the example, but using the plugin to save.

I've now put the plugin saving as a comment in the `add_shapes.py` example - makes more sense there I think. The line is also commented so that you don't write an svg every time you run the example

sofroniewn:
I have now updated this PR to use the `0.1.0` release of `napari-svg` which includes the use of the new `multiscale` kwarg. I will merge once tests pass.

tlambert03 (approved):
wowzers!  that's a lot of code removed :)

sofroniewn:
Hmm - @tlambert03 - I just noticed what should be a bug in the napari-svg code, there is still an import from pluggy https://github.com/napari/napari-svg/blob/master/napari_svg/hook_implementations.py#L3, which napari-svg doesn't depend on, but all these tests still passed. Curious if you understand what's going on here?

tlambert03:
if you have pytest in your environment then you’ll have pluggy too (even if you don’t put it in your requirements). And the marker itself is a tiny bit of code that is still cross-compatible between pluggy and napari-plugin-engine.  So it shouldn’t be a bug (i.e. devs can still use the pluggy marker if they want to)

PR #1146, opened 2020-04-15
Open up Pillow again after 7.1.*

tlambert03:
# Description
@jni mentioned that he [prefers blocking specific versions](https://github.com/napari/napari/pull/1108#issuecomment-608979036), and the Pillow issues are going to be fixed in version ≥7.2.0 (fixed in https://github.com/python-pillow/Pillow/pull/4528), so this PR just restricts Pillow to != 7.1.*

## Type of change

- [x] dependencies

# How has this been tested?

I ran our test suite and some examples on Pillow 7.2.0.dev and everything works

PR #1172, opened 2020-04-26
Update Pillow dependency pin

hugovk:
# Description

As Pillow is an indirect dependency, remove the pin added in https://github.com/napari/napari/pull/1146.

https://github.com/python-pillow/Pillow/pull/4528 was released in Pillow 7.1.2 yesterday: https://pillow.readthedocs.io/en/stable/releasenotes/7.1.2.html

Alternatively, you could keep the pin with:

`Pillow!=7.1.0,!=7.1.1`

or

`Pillow>=7.1.2`

## Type of change

# References

# How has this been tested?

## Final checklist:
- [n/a] I have commented my code, particularly in hard-to-understand areas
- [n/a] I have made corresponding changes to the documentation
- [n/a] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -2,7 +2,7 @@ napari-plugins>=0.1.0
 dask[array]>=2.1.0
 fsspec>=0.3.3
 imageio>=2.5.0
-Pillow!=7.1.*  # not a direct dependency, but 7.1.0 and 7.1.1 broke imageio
+Pillow!=7.1.0,!=7.1.1  # not a direct dependency, but 7.1.0 and 7.1.1 broke imageio
 ipykernel>=5.1.1
 numpy>=1.10.0
 qtpy>=1.7.0


tlambert03:
Thanks @hugovk!  I made the (incorrect) assumption that the next version was gonna be 7.2 (based on the `7.2.dev0` version I saw somewhere.). Since I'm fairly certain that a lot of napari functions will be broken if someone happens to end up with 7.1.0 or 7.1.1 (and so as to maintain backwargs compatibility), I think I'd prefer `Pillow!=7.1.0,!=7.1.1`.  Would you mind updating this PR, and then I can merge it?

hugovk:
> > Thanks @hugovk! I made the (incorrect) assumption that the next version was gonna be 7.2 (based on the `7.2.dev0` version I saw somewhere.). 

You made the correct assumption! 7.2.0 would be the next scheduled quarterly release (July 2020) and `7.2.dev0` is in the `master` branch.

Because this was affecting quite a few projects, we made a 7.1.2 point release.

> Since I'm fairly certain that a lot of napari functions will be broken if someone happens to end up with 7.1.0 or 7.1.1 (and so as to maintain backwards compatibility), I think I'd prefer `Pillow!=7.1.0,!=7.1.1`. Would you mind updating this PR, and then I can merge it?

Good idea, updated!

tlambert03 (approved):
Thanks!  and thanks for coming by to let us know! :)

PR #1173, opened 2020-04-27
Turn caching on and fusion off when adding a dask array 

tlambert03:
# Description
carrying on from the discussion in #718 and #1124, this PR detects when a layer is being created from a dask array, and spins up a dask cache for opportunistic caching and turns off task fusion (using the new feature from https://github.com/dask/dask/pull/6087, thanks @mrocklin).

It require dask > 2.15.0, but if they have an older version, it doesn't break anything, and they will see a warning to update to >= 2.15.0 for best performance.  To autodetect cache size, I try to import [psutil](https://github.com/giampaolo/psutil), and then show a message and fall back to a hard-coded 32GB if it's not available (I think it's fine to be larger than physical RAM?).

Just tried it with https://github.com/tlambert03/napari-llsfolder and it performs much better than the current master, (without having to add the stuff to the plugin).

## Type of change

# References
https://github.com/dask/dask/pull/6087
https://docs.dask.org/en/latest/caching.html

# How has this been tested?

only manual tests at this point.

## Final checklist:


napari/layers/base/base.py
```diff
@@ -130,6 +133,23 @@ def __init__(
         if name is None and data is not None and os.getenv('MAGICNAME'):
             name = magic_name(data, path_prefix=ROOT_DIR)
 
+        # if we have a dask array, setup some sane defaults for optimized
+        # indexing and opportunistic caching.
+        if data is not None and isinstance(data, da.Array):
```
sofroniewn: Can you maybe also check if data is a list of dask arrays - i'm thinking about a multiscale use case here

napari/utils/misc.py
```diff
@@ -260,3 +260,53 @@ def all_subclasses(cls: Type) -> set:
...
+
+    See `Dask oportunistic caching
+    <https://docs.dask.org/en/latest/caching.html>`_
+
+    Parameters
+    ----------
+    nbytes : int, optional
+        The desired size of the cache, in bytes, by default, napari wil attempt
+        to import ``psutil`` and use 80% of the available memory.  Otherwise it
+        will emit a warning and use 32GB.
```
sofroniewn: This seems reasonable to me, but curious what others think

tlambert03: Unless we want to avoid dependency creep, I'd also be quite happy to just depend on psutil.  as a "high performance" app, I can definitely imagine needing it in the future for system monitoring features.

tlambert03:
one other point to make here:  this PR turns of dask task fusion _globally_ and permanently for the session.  That is the simplest way to implement it, but if we wanted to be more granular, I think we could have an `_is_dask` attribute on the layer, and then wrap the `set_view_slice` methods in a [`dask.config.set`](https://docs.dask.org/en/latest/configuration.html#dask.config.set) context.

d-v-b:
Is there any way to avoid changing the global dask config here? Programs modifying the global dask config at runtime sounds like a potential source of nightmarish bugs.

tlambert03:
I'm not sure about "nightmarish bugs" in this case. The reason I did it this way is [this comment](https://github.com/napari/napari/issues/718#issuecomment-612062030), where @mrocklin said:
> Task fusion is not critical. Things will be mildly less efficient, but I wouldn't worry too much about it.

to do it just for dask layer indexing, I think we would need to either make sure all Layer subclasses call `super()._set_view_slice` in their methods, or give Layer a non-private `set_view_slice` that calls the child `_set_view_slice` in a dask context, and make sure it's always used (my preference).  Both approaches require the "[consenting adults](https://python-guide-chinese.readthedocs.io/zh_CN/latest/writing/style.html#we-are-all-consenting-adults)" principle, and can of course cause their _own_ types of nightmarish bugs if later developers don't adhere to the pattern.

edit: also, it's not going to modify it at runtime.  It's only going to do it if someone instantiates a napari Layer using a dask array as the data.

d-v-b:
OK, a mild drop in efficiency doesn't sound nightmarish :)

But to be clear -- this would modify the global dask config, right? So the following scenario could transpire: 
```
1. do dask stuff with user config
2. visualize dask array with napari
3. do more dask stuff, but now with a config that napari has silently modified
```
In practice I think it's fine if napari only modifies the dask config in a subtle way. But in principle, I don't like the idea of my visualization tool changing (let alone impairing, even mildly) how my distributed compute jobs run. Debugging distributed code is hard enough with a constant config.

sofroniewn:
@tlambert03 once we have persistant settings in place, this would probably be a good place for one - both cache size and turning this on/ off

tlambert03:
just pushed an update that keeps the config change local to setting the view slice.  The only catch is that it only works if we call `layer.set_view_slice()` instead of `layer._set_view_slice()` ... but since the _only_ place that function is currently called is in `Layer.refresh`, I think we'll be fine.

@sofroniewn, will add to #992

mrocklin:
Can you call `dask.compute.set(...)` around any time that you call `.compute()`  perhaps?  That might help you to isolate things.

tlambert03:
did you mean `dask.config.set()`?  If so, that's how I have it currently implemented after the last commit.  Our `Layer._set_view_slice()` is the main place we call `compute()` and I wrapped it in a config set [here](https://github.com/napari/napari/blob/2075429048d765f9f3abca77eb2fa6a3d7a58c04/napari/layers/base/base.py#L604-L609)

mrocklin:
Oh yes, that's exactly what I meant.  I should do my homework next time :)

tlambert03:
lol.  you're allowed

napari/utils/misc.py
sofroniewn: I'm happy for us to depend on psutil too - seems like a major and small package

sofroniewn: https://github.com/giampaolo/psutil

jni: I think 50% is a better target by default. I generally prefer having headroom with my RAM, rather than have a process grind my computer to a halt...

I'm ok with depending on psutil.

napari/utils/misc.py
```diff
@@ -260,3 +260,53 @@ def all_subclasses(cls: Type) -> set:
...
+    """
+
+    from dask.cache import Cache
+    from napari import utils
+
+    if not nbytes:
```
jni: Maybe we should allow `nbytes=0` to turn off caching, and `nbytes=None` to mean "guess".

napari/utils/misc.py
tlambert03: sounds good.  how about i add another keyword argument that lets you state either the absolute amount (the nbytes argument), or the percentage of available/total (since we'll have psutil)?

napari/layers/base/base.py
```diff
@@ -138,6 +142,30 @@ def __init__(
...
+
+            resize_dask_cache()
+            dask_version = tuple(map(int, dask.__version__.split(".")))
+            if dask_version < (2, 15, 0) and not Layer._have_warned_dask:
+                warnings.warn(
+                    'For best performance with Dask arrays in napari, please '
+                    'upgrade Dask to v2.15.0 or later. Current version is '
+                    f'{dask.__version__}'
+                )
+                Layer._have_warned_dask = True
```
jni: This is super-messy for our `__init__`. I'd suggest sweeping it under appropriately named method/rug, e.g.

```python
self._is_dask = self._config_dask()
```

Also, I don't think `Layer._have_warned_dask` is needed? The default is to only warn once per matching line?

https://docs.python.org/3/library/warnings.html#the-warnings-filter

napari/layers/base/base.py
```diff
@@ -581,6 +609,13 @@ def cursor_size(self, cursor_size):
         self.events.cursor_size(cursor_size=cursor_size)
         self._cursor_size = cursor_size
 
+    def set_view_slice(self):
+        if self._is_dask_data:
+            with dask.config.set({"optimization.fuse.active": False}):
+                self._set_view_slice()
+        else:
+            self._set_view_slice()
```
jni: @tlambert03 as a reminder from our conversation on Zulip, could you add a link to the dask code that this affects?

napari/utils/misc.py
```diff
@@ -260,3 +260,53 @@ def all_subclasses(cls: Type) -> set:
...
+            import warnings
+
+            if not utils.misc._have_warned_psutil:
+                warnings.warn(
+                    'Could not import psutil to get available memory for '
+                    'caching. Run "pip install psutil" to automatically detect'
+                    ' memory. Or resize cache manually with '
+                    'napari.utils.resize_dask_cache'
+                )
+                utils.misc._have_warned_psutil = True
```
jni: as above — I don't think this should be needed.

tlambert03: that would be great.  I did it cause I saw a double warning... but will go back and figure out what was going on there

napari/layers/base/base.py
jni: I also wonder if we should just do the config set regardless of `_is_dask_data`.

tlambert03: sure! then I guess we don't need to care if it's a dask array at all. should we still emit a one-time warning?  (And if so, should it still live on the base layer?)

jni:
@tlambert03 I've made a few comments about specifics of this approach. I'd like to think about this for another couple of days. It's kind of sad to me to add dask-specific code to the base napari components, when we had thus far been able to effectively treat dask arrays as any other numpy-coercible array, which was so clean.

Having said this, I do think on balance this is preferable to setting global configuration outside of napari, whether silently or explicitly via docs. So I'm leaning towards +1.

napari/layers/base/base.py
jni: that's a good question! =P Not sure.

tlambert03:
> It's kind of sad to me to add dask-specific code to the base napari components, when we had thus far been able to effectively treat dask arrays as any other numpy-coercible array, which was so clean.

yeah I get that.  just as a reminder... i _believe_ the desired result here could be also be achieved just changing the way we actually index into the data arrays, as you pointed out in https://github.com/napari/napari/issues/718#issuecomment-557513190

edit:
I also posted this workaround on zulip, but will put here again.  This code, placed in `_set_view_slice()` also avoids needing to change the dask config (this is just for reading 3D image files, whereas the dask config generalizes better):

```python
            if self.dims.ndim > 2:
                outer_idx = tuple(self.dims.indices[: self.dims.ndim - 3])
                image = np.asarray(self.data[outer_idx])
                image = image[tuple(self.dims.indices[-3:])]
            else:
                image = np.asarray(self.data[self.dims.indices])
            image = image.transpose(order)
```

napari/layers/base/base.py
jni: Another approach would be to create a context manager that is a no-op if there is no dask data within the `self._config_dask()` method proposed above. then here we would always do:

```python
with self.dask_optimized_slicing():
    self._set_view_slice()
```

napari/utils/misc.py
jni: Sure!

jni:
One more thought: @tlambert03 do you think you could create a benchmark for this? e.g. create a slow dask array and show that caching is working as expected when moving slices (slow first access, fast second access).

Incidentally, would this speed up `dask_arr[i + 1]` when we have `dark_arr[i]` if the chunking is such that `dask[i + 1]` *must* be fetched to fetch `dask[i]`?

sofroniewn:
If anyone was wondering, this doesn't seem to have sped up my PIMS lazy movie viewing, see https://github.com/dask/dask-image/issues/134, though it probably wasn't expected too.

@tlambert03 responded in another thread
> that's too bad :/ woulda been nice if it just worked

napari/layers/base/base.py
tlambert03: for some reason, it's warning every time. this gives 4 warnings:

```python
with napari.gui_qt():
    napari.view_image(dask.array.ones((10, 10)))
    napari.view_image(dask.array.ones((10, 10)))
    napari.view_image(dask.array.ones((10, 10)))
    napari.view_image(dask.array.ones((10, 10)))
```

for the life of me, I can't figure out why, or how to get the simplefilter("once") to work.  even if I use `python -W once` or explicitly put it in a context:
```
with warnings.catch_warnings():
    warnings.simplefilter("once")
```
I still see a warning every time I create a new layer.

napari/utils/misc.py
tlambert03: yeah, as mentioned above, I can't it to be a single warning no matter what I do (besides my explicit conditional), but since it's "supposed" to work like that (I even see the "once" filter in `warnings.filters`), maybe it's something in my environment.  I will just remove this code and let it potentially warn multiple times.

napari/layers/base/base.py
tlambert03: love this, did it. and I think it looks much better now!

tlambert03:
OK, thanks to all your input, I think this is _much_ better now!
I moved all the logic to `utils.misc.configure_dask` and `utils.misc.resize_dask_cache`, leaving the `Layer` module totally free of dask imports, and just a couple new lines.

@jni, I don't have a benchmark for you, but I added [two tests](https://github.com/napari/napari/pull/1173/files#diff-97ca8537ba2c907c03734c08881d8117R82-R151) that (I think at least) are better: it shows the exact number of io calls made when performing various indexing operations, and then monkeypatches the `configure_dask` function to break caching and fusion optimizations, and counts all the excess calls that get made.  It shows that, without this PR, we would make a full 3D read every time the Z plane is changed (due to task fusion), _and_ would re-read previously read timepoints when revisiting (due to no caching).

jni (approved):
Pure gold now! The tests are :ok_hand:

PR #1175, opened 2020-04-28
Change array to np.ndarray in type information

Czaki:
# Description
Some of IDE like PyCharm read types from docstring and use it in type hints/warnings

## Type of change

- [x] Update docstring
- [x] add ArrayInterface protocol 

# References

https://www.jetbrains.com/help/pycharm/type-hinting-in-product.html#

## Final checklist:

### DIFF
--- a/napari/components/add_layers_mixin.py
+++ b/napari/components/add_layers_mixin.py
@@ -1,12 +1,15 @@
 import inspect
 import itertools
+from abc import ABC, abstractmethod
 import os
 from functools import lru_cache
 from logging import getLogger
-from typing import Any, Dict, List, Optional, Sequence, Set, Union
+from typing import Any, Dict, List, Optional, Sequence, Union, TypeVar
+from ..utils.colormaps import ensure_colormap_tuple
 
 import numpy as np
 
+from . import Dims
 from .. import layers
 from ..layers.image._image_utils import guess_labels, guess_multiscale
 from ..plugins.io import read_data_with_plugins
@@ -21,8 +24,14 @@ from ..utils.misc import (
 
 logger = getLogger(__name__)
 
+LayerType = TypeVar("LayerType", bound=layers.Layer)
 
-class AddLayersMixin:
+
+class ArrayInterface(ABC):
+    pass
+
+
+class AddLayersMixin(ABC):
     """A mixin that adds add_* methods for adding layers to the ViewerModel.
 
     Each method corresponds to adding one or more layers to the viewer.
@@ -35,7 +44,12 @@ class AddLayersMixin:
     easier to read and make these methods easier to maintain.
     """
 
-    def add_layer(self, layer: layers.Layer) -> layers.Layer:
+    def __init__(self):
+        super().__init__()
+        self.layers = []
+        self.dims = Dims()
+
+    def add_layer(self, layer: LayerType) -> LayerType:
         """Add a layer to the viewer.
 
         Parameters
@@ -256,7 +270,7 @@ class AddLayersMixin:
                 else:
                     kwargs[key] = iter(ensure_iterable(val))
 
-            layer_list = []
+            layer_list: List[layers.Image] = []
             for i in range(n_channels):
                 if multiscale:
                     image = [
@@ -266,8 +280,9 @@ class AddLayersMixin:
                 else:
                     image = np.take(data, i, axis=channel_axis)
                 i_kwargs = {k: next(v) for k, v in kwargs.items()}
-                layer = self.add_layer(layers.Image(image, **i_kwargs))
-                layer_list.append(layer)
+                layer_list.append(
+                    self.add_layer(layers.Image(image, **i_kwargs))
+                )
             return layer_list
 
     def add_points(
@@ -397,8 +412,7 @@ class AddLayersMixin:
             blending=blending,
             visible=visible,
         )
-        self.add_layer(layer)
-        return layer
+        return self.add_layer(layer)
 
     def add_labels(
         self,
@@ -913,6 +927,7 @@ class AddLayersMixin:
         A typical use case might be to upack a tuple of layer data with a
         specified layer_type.
 
+        >>> import napari
         >>> viewer = napari.Viewer()
         >>> data = (
         ...     np.random.random((10, 2)) * 20,
@@ -956,6 +971,42 @@ class AddLayersMixin:
                 raise exc
 
         return layer
+
+    @abstractmethod
+    def _update_active_layer(self, event):
+        pass
+
+    @abstractmethod
+    def _update_status(self, event):
+        pass
+
+    @abstractmethod
+    def _update_help(self, event):
+        pass
+
+    @abstractmethod
+    def _update_interactive(self, event):
+        pass
+
+    @abstractmethod
+    def _update_cursor(self, event):
+        pass
+
+    @abstractmethod
+    def _update_cursor_size(self, event):
+        pass
+
+    @abstractmethod
+    def _on_layers_change(self, event):
+        pass
+
+    @abstractmethod
+    def _update_layers(self, event=None, layers=None):
+        pass
+
+    @abstractmethod
+    def reset_view(self, event=None):
+        pass
 
 
 @lru_cache(maxsize=1)
--- a/napari/components/add_layers_mixin.py
+++ b/napari/components/add_layers_mixin.py
@@ -92,7 +92,7 @@ class AddLayersMixin:
 
         Parameters
         ----------
-        data : array or list of array
+        data : np.ndarray or list of np.ndarray
             Image data. Can be N dimensional. If the last dimension has length
             3 or 4 can be interpreted as RGB or RGBA if rgb is `True`. If a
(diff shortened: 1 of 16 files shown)


napari/components/add_layers_mixin.py
```diff
@@ -88,7 +88,7 @@ def add_image(
 
         Parameters
         ----------
-        data : array or list of array
+        data : np.ndarray or list of np.ndarray
```
jni: @Czaki the problem is that this is inaccurate. We actually accept anything coercible to a NumPy array here, including dask arrays, pint Quantities, zarr arrays, pytorch CPU tensors...

Czaki: Hm. I see  [PEP 544](https://www.python.org/dev/peps/pep-0544/) which can be used to solve this. 

I found that the problem with is produced by existence of `array` module https://docs.python.org/3/library/array.html and name collision,

tlambert03: yes indeed, pep 544 holds promise... and has been a hot topic of conversation within the napari team for a while now! :)

it is definitely a long term goal to implement real type stubs for valid napari layer data types.  (see @jni's [imtypes](https://github.com/napari/image-types) module for a _very_ early start on this concept... though that was created long before we had started discussing Protocols).  After digging in a little deeper, it turns out to be a bit harder than it first appears... and so we have a temporary "copout" union [here](https://github.com/napari/napari/blob/6933f5bcc12608d66458987d69db2f7c535b21de/napari/types.py#L13-L20).

but if you wanted to look into creating a protocol type that would recognize all of the various types of data that napari supports, such that an IDE recognizes them, it is definitely something that is needed!

Czaki: But If I good understand Protocols are introduced in python 3.8. And It takes long time before `napari` drop support for older pythons? 

(PyCharm suggest that support this from python 3.7).

Also connected with this topisc. I do not see type annotation in napari code. But @tlambert03 message suggest that there is no contradictions to introduce them?

tlambert03: > But If I good understand Protocols are introduced in python 3.8. And It takes long time before `napari` drop support for older pythons?

there is a backport [typing-extensions](https://pypi.org/project/typing-extensions/) that covers all the versions we support.


> Also connected with this topisc. I do not see type annotation in napari code. But @tlambert03 message suggest that there is no contradictions to introduce them?

there are plenty of type annotations in napari code!  ([31 different files have type hints](https://github.com/napari/napari/search?q=%22from+typing+import%22)) though of course they are not everywhere, and in particular, they are not all over our layer classes (though, see https://github.com/napari/napari/pull/1097 for a start and some discussion on it)

by all means, feel free to use type annotations... but if you use one that is introduced after 3.6, we will need to depend on the backport (which we have not yet done)

jni: Note that `imtypes` was written before PEP-544 was accepted. It would probably look different now. =)

We will drop Py3.6 support in November at the latest, along with the rest of the scientific Python stack, see [NEP-29](https://numpy.org/neps/nep-0029-deprecation_policy.html). So we can definitely start to contemplate things that are available in Py3.7 only. But I'm 100% on board with depending on typing backports because I love typing. =)

napari/layers/utils/array_interface.py
```diff
@@ -0,0 +1,21 @@
...
+class ArrayInterface(Protocol):
+    shape: Tuple[int] = (0,)
+    dtype: type = np.uint8
+    ndim: int = 1
+
+    def __getitem__(self, item) -> Union['ArrayInterface', int, float, bool]:
+        ...
+
+    def __len__(self) -> int:
+        ...
```
Czaki: @tlambert03 @jni Something like this?

napari/layers/utils/array_interface.py
```diff
@@ -0,0 +1,21 @@
...
+import numpy as np
+
+if sys.version_info.minor < 8:
+    from typing_extensions import Protocol
+else:
+    from typing import Protocol
+
+
+class ArrayInterface(Protocol):
+    shape: Tuple[int] = (0,)
```
tlambert03:
```suggestion
    shape: Tuple[int, ...]
```

napari/layers/utils/array_interface.py
```diff
@@ -0,0 +1,21 @@
...
+    dtype: type = np.uint8
```
tlambert03:
```suggestion
    dtype: np.dtype
```

this might be too strict? but worth considering ... both dask and zarr use numpy dtypes.

napari/layers/utils/array_interface.py
```diff
@@ -0,0 +1,21 @@
...
+    ndim: int = 1
```
tlambert03:
```suggestion
    @property
    def ndim(self) -> int: ...
```
otherwise suggests that `ndim` is writable and will lead to this mypy error if someone implements as a property:
Screen Shot 2020-04-30 at 12 05 35 PM

napari/layers/utils/array_interface.py
tlambert03: yep! something like that!  will be fun/interesting to try to "break" it with various cases... that's where i've generally run into hurdles.  lack of type stubs in various libraries also makes this a little less "practically useful" sometimes, which is another thing that has cause me to drag my feet on it.  But this is exactly how we should start.

napari/layers/utils/array_interface.py
Czaki: np.uint8 is not in type np.dtype

napari/layers/utils/array_interface.py
Czaki: changed

napari/layers/utils/array_interface.py
tlambert03: It might also be worth considering [subprotocols](https://mypy.readthedocs.io/en/stable/protocols.html#defining-subprotocols-and-subclassing-protocols).  For instance, there are lot of places where we have restrictions on the shape/type of an array (like points data), and we had imagined the ability to hint that.  Subprotocols will let us construct specific combinations,  For instance:

```python
class SupportsShape(Protocol):
    shape: Tuple[int, ...]

class Is2D(Protocol):
    shape: Tuple[int, int]

class SupportsLen(Protocol):
    def __len__(self, item) -> int: ...

class ArrayInterface(SupportsShape, SupportsLen, Protocol): ...
class Array2DInterface(Is2D, SupportsLen, Protocol): ...
PointsLayer = Union[Array2Dinterface, ...]  # or something like that
```

Czaki: I implement only minimal signature because I'm not familliar with other numpy like packets. 

So I think that best is to know where also (other than `set_image`) could be passed non numpy array and check which properties are needed.

napari/layers/utils/array_interface.py
tlambert03: ok, perhaps that's not the right syntax... but you understand the idea, right?

tlambert03: maybe `Type[np.dtype]`

Czaki:
```
type(np.uint8)
Out[4]: type
```
but
```
a = np.arange(5, dtype=np.uint8)
type(a.dtype)
Out[6]: numpy.dtype
```

napari/layers/utils/array_interface.py
```diff
@@ -0,0 +1,29 @@
...
+    def __len__(self) -> int:
+        ...
+
+    @property
+    def shape(self) -> Tuple[int, ...]:
+        return (0,)
+
+    @property
+    def dtype(self) -> type:
+        return np.uint8
```
tlambert03: interestingly enough, these two actually _shouldn't_ be `@property` ... since they _are_ editable in a numpy array.  try it out:
```python
In [42]: x = np.ones((10,10))                                                                                                                                                                  

In [43]: x.shape                                                                                                                                                                               
Out[43]: (10, 10)

In [44]: x.shape = (25, 4)                                                                                                                                                                     

In [45]: x.shape                                                                                                                                                                               
Out[45]: (25, 4)

In [46]: x                                                                                                                                                                                     
Out[46]: 
array([[1., 1., 1., 1.],
       [1., 1., 1., 1.],
       [1., 1., 1., 1.],
       [1., 1., 1., 1.],
...
```

Czaki: Now I remember.. Only assumption is that new shape need to multiply to same size. And type overwrite change size.

napari/layers/utils/array_interface.py
Czaki: see now.

Czaki:
This PR becomes outdated. Should I rebase and fix it or protocol introduction will be delayed?

jni:
The NumPy, Torch, Tensorflow et al folks have been [making progress](https://mail.python.org/archives/list/numpy-discussion@python.org/message/55UHYK5JNYMYFNGT4VLLP27ZW6PQMXAK/) on proper type annotations for tensor-like data, so we will close this and wait for that ecosystem to mature. Thanks @Czaki!

Issue #1178, opened 2020-04-29
Pyramid rendering failing for ndisplay=3


jni:
## 🐛 Bug

Trying to view pyramids in 3D results in a hard crash.

## To Reproduce

```python
import napari
import numpy as np

pyr = [np.random.random((128,) * 3),
       np.random.random((64,) * 3)]

with napari.gui_qt():
    v = napari.view_image(pyr)
    v.dims.ndisplay = 3
```

gives:

```pytb
Traceback (most recent call last):
  File "/home/jni/miniconda3/envs/f7/lib/python3.7/site-packages/vispy/app/backends/_qt.py", line 825, in paintGL
    self._vispy_canvas.events.draw(region=None)
  File "/home/jni/miniconda3/envs/f7/lib/python3.7/site-packages/vispy/util/event.py", line 455, in __call__
    self._invoke_callback(cb, event)
  File "/home/jni/miniconda3/envs/f7/lib/python3.7/site-packages/vispy/util/event.py", line 475, in _invoke_callback
    self, cb_event=(cb, event))
  File "/home/jni/miniconda3/envs/f7/lib/python3.7/site-packages/vispy/util/event.py", line 471, in _invoke_callback
    cb(event)
  File "/home/jni/projects/napari/napari/_vispy/vispy_base_layer.py", line 241, in on_draw
    shape_threshold=self.node.canvas.size,
  File "/home/jni/projects/napari/napari/layers/base/base.py", line 675, in _update_multiscale
    downsample_factors,
  File "/home/jni/projects/napari/napari/layers/utils/layer_utils.py", line 249, in compute_multiscale_level
    locations = np.argwhere(np.all(scaled_shape > shape_threshold, axis=1))
ValueError: operands could not be broadcast together with shapes (2,3) (2,) 
```
## Expected behavior

The lowest resolution level of the pyramid should be displayed.

## Environment

 - Please copy and paste the information at napari info option in help menubar here:

```
napari: 0.2.12+78.g6933f5bc
Platform: Linux-4.15.0-29-generic-x86_64-with-debian-buster-sid
Python: 3.7.3 | packaged by conda-forge | (default, Jul  1 2019, 21:52:21)  [GCC 7.3.0]
Qt: 5.13.0
PyQt5: 5.13.1
NumPy: 1.18.3
SciPy: 1.3.1
Dask: 2.5.2
VisPy: 0.6.4

GL version:  3.0 Mesa 19.2.8
[...]

PR #1179, opened 2020-04-29
Prevent crash when viewing 3D pyramids

jni:
# Description

Fixes crash in #1178.

Note that the test proposed doesn't work, in that it also passes on master. I don't understand why, though. @sofroniewn? 

## Type of change

# References

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/_tests/test_viewer.py
+++ b/napari/_tests/test_viewer.py
@@ -61,6 +61,16 @@ def test_no_qt_loop():
         _ = Viewer()
 
 
+def test_3d_image_pyramid(viewer_factory):
+    view, viewer = viewer_factory()
+    data = [np.random.random((128,) * 3), np.random.random((64,) * 3)]
+    viewer.add_image(data)
+    # check that this doesn't crash
+    viewer.dims.ndisplay = 3
+    viewer.layers[0].refresh()
+    check_viewer_functioning(viewer, view=view, data=data, ndim=3)
+
+
 @pytest.mark.parametrize('layer_class, data, ndim', layer_test_data)
 @pytest.mark.parametrize('visible', [True, False])
 def test_add_layer(viewer_factory, layer_class, data, ndim, visible):
--- a/napari/_tests/test_viewer.py
+++ b/napari/_tests/test_viewer.py
@@ -61,16 +61,6 @@ def test_no_qt_loop():
         _ = Viewer()
 
 
-def test_3d_image_pyramid(viewer_factory):
-    view, viewer = viewer_factory()
-    data = [np.random.random((128,) * 3), np.random.random((64,) * 3)]
-    viewer.add_image(data)
-    # check that this doesn't crash
-    viewer.dims.ndisplay = 3
-    viewer.layers[0].refresh()
-    check_viewer_functioning(viewer, view=view, data=data, ndim=3)
-
-
 @pytest.mark.parametrize('layer_class, data, ndim', layer_test_data)
 @pytest.mark.parametrize('visible', [True, False])
 def test_add_layer(viewer_factory, layer_class, data, ndim, visible):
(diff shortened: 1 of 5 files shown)


napari/_tests/test_viewer.py
```diff
@@ -61,6 +61,16 @@ def test_no_qt_loop():
...
 
 
+def test_3d_image_pyramid(viewer_factory):
+    view, viewer = viewer_factory()
+    data = [np.random.random((128,) * 3), np.random.random((64,) * 3)]
+    viewer.add_image(data)
+    # check that this doesn't crash
+    viewer.dims.ndisplay = 3
+    viewer.layers[0].refresh()
+    check_viewer_functioning(viewer, view=view, data=data, ndim=3)
```
jni: This passes on master though it crashes when I try to run the viewer on my machine (and I expect all machines). Any ideas why?

napari/layers/image/image.py
```diff
@@ -483,21 +483,22 @@ def _set_view_slice(self):
                 scale[d] = self.downsample_factors[self.data_level][d]
             self._transforms['tile2data'].scale = scale
 
-            corner_pixels = np.clip(
-                self.corner_pixels,
-                0,
-                np.subtract(self.level_shapes[self.data_level], 1),
-            )
+            if self.dims.ndisplay == 2:
```
jni, line 486: Once I fixed `_update_multiscale`, I noticed that rotating a 3D cube resulted in the cube disappearing as it was rotated, until it was completely gone if looking at it side on. This change fixes it, though I don't have a test for it.

jni:
@sofroniewn I don't have a good enough understanding of the multiscale code path to figure out whether these changes are the best way of handling #1178 and the subsequent rendering problem. Feel free to take this PR over while I sleep. :slightly_smiling_face: 

I think this needs to be in before 0.3.0.

jni:
I'll also note that the performance is now atrocious when viewing the top level of a pyramid in 3D... Any idea how to prevent that? Since we are never switching levels in 3D, we should shortcut all refreshes with pan/zoom. Example code:

```python
import dask.array as da
import numpy as np
import napari                                                                                                                                                                         

high_res = da.random.random((199, 201, 192 * 4, 256 * 4), chunks=(1, 1, 192*4, 256*4))                                                                                                
low_res = da.random.random((199, 201, 192, 256), chunks=(1, 1, 192, 256))                                                                                                             
low_res_vol = np.array(low_res[0])                                                                                                                                                    

# this is unusable
with napari.gui_qt():
    v = napari.view_image([high_res, low_res])                                                                                                                                            

# this is perfectly smooth
with napari.gui_qt():
    v = napari.view_image(low_res_vol, ndisplay=3)                                                                                                                                        
```

napari/_tests/test_viewer.py
sofroniewn: The reason is because the vispy `on_draw` command isn't triggered so the code with the bug doesn't get called in our tests see here for an example https://github.com/napari/napari/blob/master/napari/_vispy/_tests/test_vispy_multiscale.py#L18

I will move the test into that folder and add the `on_draw` command

sofroniewn:
> I'll also note that the performance is now atrocious when viewing the top level of a pyramid in 3D... Any idea how to prevent that? Since we are never switching levels in 3D, we should shortcut all refreshes with pan/zoom

Done

Also test above fixed so will fail on master, but passes now. This is ready for review @jni

sofroniewn (approved):
This looks good to me. I'm approving, would be good to hear from @jni before merge as I made the last change

napari/_tests/test_viewer.py
sofroniewn: done

jni:
Looking over this now @sofroniewn but as a side note, do you know why the requirements checks are failing?

tlambert03:
> Looking over this now @sofroniewn but as a side note, do you know why the requirements checks are failing?

I fear that pip_check_reqs package might be a little buggy.

jni:
@sofroniewn thank you for the updates!!! :tada:

PR #1180, opened 2020-04-29
Add ensure_colormap utility function to standardize colormap getting/setting 

tlambert03:
# Description
This PR fixes the issue raised by @will-moore in https://github.com/napari/napari/pull/1092#issuecomment-621148849.  I introduce a new utility function `napari.utils.colormaps.ensure_colormap_tuple` that accepts any of the valid colormap argument types (string, `Colormap` instance, tuple of `(string, Colormap)`, dict of `{string: Colormap}`) and returns a `(name, Colormap)` tuple that is _guaranteed_ to be available in `AVAILABLE_COLORMAPS`, otherwises it raises various Exceptions depending on the specifics of the argument (it will also look up strings in `vispy_or_mpl_colormap` before raising).  This simplifies fixing https://github.com/napari/napari/pull/1092#issuecomment-621148849 because it means we can always normalize any type of colormap arg to a string, that can be looked up elsewhere.

I also replaced all of our `@colormap.setter` methods to use this function, and added a lot of tests for all of the various combinations of types with and without `channel_axis`.

@kevinyamauchi: I changed the "unknown_colormap" literal in the points and vectors layers to match the `increment_unnamed_colormap` pattern used in `Image`, let me know if that breaks any of your stuff (i.e. if you need the literal string `unknown_colormap` somewhere).

## Type of change

- [x] refactor

# References

# How has this been tested?

- [x] added tests

## Final checklist:

### DIFF
--- a/napari/components/_tests/test_multichannel.py
+++ b/napari/components/_tests/test_multichannel.py
@@ -1,12 +1,17 @@
 import numpy as np
 import dask.array as da
 from napari.components import ViewerModel
-from napari.utils.colormaps import colormaps
+from napari.utils.colormaps import colormaps, ensure_colormap_tuple
 from napari.utils.misc import ensure_sequence_of_iterables, ensure_iterable
 import pytest
 
 base_colormaps = colormaps.CYMRGB
 two_colormaps = colormaps.MAGENTA_GREEN
+green_cmap = colormaps.simple_colormaps['green']
+red_cmap = colormaps.simple_colormaps['red']
+fire = colormaps.AVAILABLE_COLORMAPS['fire']
+cmap_tuple = ("my_colormap", colormaps.Colormap(['g', 'm', 'y']))
+cmap_dict = {"your_colormap": colormaps.Colormap(['g', 'r', 'y'])}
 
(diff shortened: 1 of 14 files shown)


will-moore:
Thanks. That's working for me locally. 👍

tlambert03:
great!  thanks a lot for testing it out, and thanks for working off our master branch! 😁

kevinyamauchi:
Thanks, @tlambert03 ! 

>@kevinyamauchi: I changed the "unknown_colormap" literal in the points and vectors layers to match the increment_unnamed_colormap pattern used in Image, let me know if that breaks any of your stuff (i.e. if you need the literal string unknown_colormap somewhere).

That's totally fine with me!

sofroniewn (approved):
This looks great to me thanks for the fix @tlambert03! I will wait to see if there are any other comments and then merge.

One thing I'll note is that in `0.3.1` I'd like us to try and move away from passing around vispy colormap objects but instead to have a simple numpy colormap representation that we use instead - similar to have we have our `Nx4` array of colors for shapes etc. I imagine this will be much easier to do with this function in place.

PR #1183, opened 2020-04-30
Add napari.config with yaml-based persistence

tlambert03:
# Description
this PR adds a persistence layer, based on the `dask.config` module, and would supercede #834.

As I was working on a save dialog for plugins, it became too annoying that simply saving an image layer without an extension would default to an svg image and not a tiff! 😂So here's a proposal.  I really like the [dask.config](https://docs.dask.org/en/latest/configuration.html) pattern.  It's based on a simple python dictionary with yaml persistence, has really nice nested namespace support, works well with downstream libraries and packages, and has a nice temporary settings context manager pattern which allows things like this:

```python
viewer.open('/Users/talley/Desktop/toxo.dv')  # works

# no plugin found
with napari.config.set({"plugin_manager.hooks.napari_get_reader.disabled": ['dv_reader']}): 
    viewer.open('/Users/talley/Desktop/toxo.dv') 

viewer.open('/Users/talley/Desktop/toxo.dv')  # works again
```

I added a timer that syncs the config to yaml every 5 seconds so, nobody has to do anything extra to make sure it persists (even if there is a crash).  here's a gif showing the file being updated (1 second here):
settings

tests will fail for now, because this PR goes together with an update to `napari-plugin-engine`:
https://github.com/napari/napari-plugin-engine/pull/1

will follow up tomorrow, but just wanted to get this out there for consideration.

## Type of change

# References

# How has this been tested?

## Final checklist:


napari/_qt/qt_main_window.py
```diff
@@ -31,12 +33,35 @@
...
     QVBoxLayout,
 )
-from qtpy.QtCore import Qt  # noqa: E402
+from qtpy.QtCore import Qt, QTimer  # noqa: E402
 from qtpy.QtGui import QKeySequence, QIcon  # noqa: E402
 from .utils import QImg2array  # noqa: E402
 from ..utils.theme import template  # noqa: E402
 
 
+class NapariMainWindow(QMainWindow):
```
tlambert03: can definitely discuss this, just wanted access to the main_window `closeEvent` which is easier with a subclass.

napari/_tests/test_config.py
```diff
@@ -0,0 +1,490 @@
+import os
```
tlambert03: these are from dask ... but all are passing with the modified napari code.

sofroniewn:
I'm getting crashes with an old version of dask here `2.12.0`
```python-traceback
  File "examples/hand_demo-2D.py", line 90, in <module>
    viewer.add_image(m, name=n)
  File "/Users/nsofroniew/GitHub/napari/napari/components/add_layers_mixin.py", line 226, in add_image
    return self.add_layer(layers.Image(data, **kwargs))
  File "/Users/nsofroniew/GitHub/napari/napari/layers/image/image.py", line 189, in __init__
    multiscale=multiscale,
  File "/Users/nsofroniew/GitHub/napari/napari/layers/intensity_mixin.py", line 27, in __init__
    super().__init__(*args, **kwargs)
  File "/Users/nsofroniew/GitHub/napari/napari/layers/base/base.py", line 159, in __init__
    resize_dask_cache()
  File "/Users/nsofroniew/GitHub/napari/napari/utils/misc.py", line 312, in resize_dask_cache
    utils.dask_cache.cache.resize(nbytes)
AttributeError: 'Cache' object has no attribute 'resize'
```
not sure if plan is to pin to newer version or be backwards compatible this far.

[EDIT] Maybe my cachey was old too, was on `0.1.1`, now upgraded to `0.2.1` and upgraded dask to `2.15.0` and everything works fine

tlambert03:
> I'm getting crashes with an old version of dask here `2.12.0`
> [EDIT] Maybe my cachey was old too, was on `0.1.1`, now upgraded to `0.2.1` and upgraded dask to `2.15.0` and everything works fine

yep that's a `cachey` issue.  however, I think you're on the wrong PR here?? 🤔 that `resize_dask_cache` function is from #1173  ... and in that PR, I put `cachey` in the requirements.txt

sofroniewn:
If anyone was wondering, this doesn't seem to have sped up my PIMS lazy movie viewing, see https://github.com/dask/dask-image/issues/134, though it probably wasn't expected too.

tlambert03:
> If anyone was wondering, this doesn't seem to have sped up my PIMS lazy movie viewing, see [dask/dask-image#134](https://github.com/dask/dask-image/issues/134), though it probably wasn't expected too.

that's too bad :/ woulda been nice if it just worked.  PS. still wrong PR 😂

sofroniewn:
indeed @tlambert03 - too much GitHub, i'm on the wrong PR, my apologies. I'm going to move the comment about the movie scrubbing to the other PR :-)

jni:
@tlambert03 this is very cool! The gif is brilliant. =D

However, I think it's a deeper issue — I don't want our default behaviour for images to be svg! I wonder whether we should:
- remove svg from being a default plugin, and/or
- move svg to the back of the queue by default

Thoughts @sofroniewn?

tlambert03:
> - move svg to the back of the queue by default

How about instead we make our builtins “tryfirst” instead of “trylast”?  I think, for a first release, that will minimize unexpected behavior for people (regardless of what plugins they have), and it’s actually more in line with Fiji style reader plugins (I think).  Though I still persistence would be nice to have in there too :)

jni:
> How about instead we make our builtins “tryfirst” instead of “trylast”?

Mmm, I dunno, I feel like if someone installs a plugin, they have signaled their intention to use it. So I'm quite happy to have them be trylast. But between our builtins, we can do as we please. And I think that should be to have svg come after the other defaults.

tlambert03:
> Mmm, I dunno, I feel like if someone installs a plugin, they have signaled their intention to use it. So I'm quite happy to have them be trylast. But between our builtins, we can do as we please. And I think that should be to have svg come after the other defaults.

problem is, napari-svg is not in "builtins" anymore... and I need to check what the result is of having both as "trylast" (depends on order of regsitration) 
I feel like the solution here is this persistence PR with a default yaml...

sofroniewn:
> Mmm, I dunno, I feel like if someone installs a plugin, they have signaled their intention to use it. So I'm quite happy to have them be trylast. But between our builtins, we can do as we please. And I think that should be to have svg come after the other defaults.
>
> >problem is, napari-svg is not in "builtins" anymore... and I need to check what the result is of having both as "trylast" (depends on order of regsitration)
I feel like the solution here is this persistence PR with a default yaml...

That is correct, I agree with "I feel like the solution here is this persistence PR with a default yaml..."

I'd actually like us to consider removing our "builtins" to their own repos too in the future

napari/config.py
```diff
@@ -0,0 +1,689 @@
...
+    if destination is None:
+        destination = PATH
+
+    if os.path.isdir(destination):
+        destination = os.path.join(destination, 'last_session.yaml')
+
+    with lock:
+        # TODO: should we also read from the file and merge new keys?
+        with open(destination, 'w') as f:
+            yaml.safe_dump(config, f)
```
tlambert03: I added this, (not in dask).  Dask seems to only _read in_ config, but never write it out.  this function writes the current config to `last_session.yaml` in the default PATH.  (happy to change the name).  _all_ yaml files in PATH get read on launch.  we can refine this later.
Note, I also added the `clean=` parameter to `set()` and `pop()`, by default, when the config is altered, a `_dirty` key is added so we know whether we need to write to disk.

napari/config.py
```diff
@@ -0,0 +1,689 @@
...
+                    config['_dirty'] = True
+            else:
+                result = result[k]
+        except (TypeError, IndexError, KeyError):
+            if default is not no_default:
+                return default
+            else:
+                raise
+    return result
+
```
tlambert03: I added this method.

tlambert03:
I removed the couple plugin-sorting specific lines, so this PR sits on it's own now and is ready for review.  I noted the functions that I added, but the rest is pretty much as it was in dask (including the tests), except where commented.  The `NapariMainWindow` is my addition to get a Timer running that writes the config to file periodically (if changed) as long as any window is open but stop when the last one is closed.  Happy to discuss/move/rename that

napari/_tests/test_config.py
```diff
@@ -0,0 +1,490 @@
...
+import stat
+import sys
+import shutil
+import tempfile
+from collections import OrderedDict
+from contextlib import contextmanager
+
+import pytest
+
+import napari.config  # noqa
```
jni: what's the dealio here?

napari/_tests/test_config.py
```diff
@@ -0,0 +1,490 @@
...
+    get,
+    merge,
+    refresh,
+    rename,
+    set,
+    update,
+    update_defaults,
+)
+
+yaml = pytest.importorskip("yaml")
```
jni: Also here...

napari/_tests/test_config.py
```diff
@@ -0,0 +1,490 @@
...
+
+
+@contextmanager
+def ignoring(*exceptions):
+    try:
+        yield
+    except exceptions:
+        pass
```
jni: :exploding_head:

napari/_tests/test_config.py
```diff
@@ -0,0 +1,490 @@
...
+#                     "    dask-schema.yaml: {}\n\n"
+#                     "Examples in these files should be a good start, \n"
+#                     "even if you are not familiar with the jsonschema spec".format(
+#                         sorted(c), sorted(s["properties"])
+#                     )
+#                 )
+#             if isinstance(v, dict):
+#                 test_matches(c[k], s["properties"][k])
+
+#     test_matches(config, schema)
```
jni: shall we get rid of the comments?

napari/config.py
jni: Lovely. =) Is it tested?

napari/config.py
```diff
@@ -0,0 +1,689 @@
...
+from collections.abc import Mapping
+import os
+import sys
+import threading
+import warnings
+
+try:
+    import yaml
+except ImportError:
+    yaml = None
```
jni: Shouldn't we just depend on pyyaml?

jni (commented):
@tlambert03 I don't really understand where the file gets written/read from. Is it just `napari.yaml` inside the package? That seems... dangerous and non-standard?

napari/config.py
tlambert03: added test

napari/_tests/test_config.py
tlambert03: removed, since we depend on yaml.  But that's a nice pytest trick that lets you try to import something, but then just skip any tests that use the module if it's not available

napari/_tests/test_config.py
tlambert03: also removed... I'm not sure why my linter previously thought it wasn't used... cause it is used :)

napari/config.py
tlambert03: sounds good to me.  did it.

tlambert03:
> @tlambert03 I don't really understand where the file gets written/read from. Is it just `napari.yaml` inside the package? That seems... dangerous and non-standard?

yeah that part isn't really well documented in here. I'll try to clarify in the module docstring.

no, the `napari.yaml` inside the package is a place for us to put defaults.  This module is set to read _multiple_ yaml files and merge them into one glorious namespaced config. 

_I put this in the module docstring, but will copy here:_

When this module is imported, this sequence of events happens (at the end of the file):

1. the config is cleared and `refresh()` is called
2. `refresh()` updates the empty config with values from dicts in the
   `config.defaults` list (currently empty)
3. `refresh()` then updates the config by calling `collect()` ...
   1. `collect()` gathers settings from various places:
      1. `collect_yaml()` looks for *any* `.yaml` files in the `config.paths`
         list at the top of the file. This includes `~/.config/napari/` and `/etc/napari`.
         For now, I'm syncing session settings to a
         file named `~/.config/napari/_session.yaml`, but that is very much open
         for discussion (though perhaps best for a follow up PR?)
      2. `collec_env()` then looks for any environment variables that begin with
         `NAPARI_`
   2. finally, `collect()` calls `merge()` on all of those discovered `dicts`,
      to yield one merged result and returns it to the `refresh()` function (by
      default, dicts discovered later in the process take priority).
4. back in the `refresh()` function, that final merged result is used to update
   the global config dict.
5. At that point, the `napari.yaml` in the repo is read, if one doesn't already
   exist, a file is created at `~/.config/napari/napari.yaml` (with
   `ensure_file()`) that is an exact copy of the repo `napari.yaml` but with
   everything commented out, as a way to show users what they can change.
6. finally, `update_defaults()` is called using the settings in the
   `napari.yaml` file, which merges the configs again, but gives priority to
   already existing values in the config (i.e. ones that were overwritten by the
   user or the environement).  so our own `napari.yaml` is really just defaults
   for things that the user has not changed.

tlambert03:
this is ready for another round of review.  I removed the plugin sorting code, and will add that in a small follow up PR that also adds a default setting for napari-svg, and bumps the napari-plugin-engine version.

Other issues for follow up PRs (probably after 0.3.0) will be:

- determining conventions for how (whether?) we differentiate settings that were set/changed during a standard interactive napari session, vs something that a user _always_ wants to set on startup (for instance, if plugins call order is resorted during a session, it's "easiest" if we just immediately store it for the next session ... but one can imagine also wanting the ability to have some things reset at startup).
- add the ability to register a callback that gets called on a settings change (for instance, something that changes the theme when the `napari.theme` config changes)

napari/_qt/qt_main_window.py
sofroniewn: I'm ok with this

sofroniewn (approved):
Very exciting @tlambert03!! Looks good to me, will wait for @jni to review before merge

napari/config.py
```diff
@@ -0,0 +1,877 @@
...
+nested namespaces, backwards compatilibity of new settings, context management
+(temporary setting) of config values, choice of yaml for persistence, and
+environment variable options
+
+see documentation for dask.config at:
+https://docs.dask.org/en/latest/configuration.html
+
+Configuration is specified in one of the following ways:
+
+1. YAML files in ~/.config/napari/ or /etc/napari/
```
jni: Does this all work well (enough) in Windows?

napari/config.py
```diff
@@ -0,0 +1,877 @@
...
+                    pass
+            else:
+                file_paths.append(path)
+    configs = []
+
+    # Parse yaml files
+    for path in file_paths:
+        try:
+            with open(path) as f:
+                data = yaml.safe_load(f.read()) or {}
```
jni: Just double checking — yaml can load json?

napari/config.py
```diff
@@ -0,0 +1,877 @@
...
+        if os.path.exists(path):
+            if os.path.isdir(path):
+                try:
+                    file_paths.extend(
+                        sorted(
+                            [
+                                os.path.join(path, p)
+                                for p in os.listdir(path)
+                                if os.path.splitext(p)[1].lower()
+                                in (".json", ".yaml", ".yml")
```
jni: I'm not 100% sure that I like this pattern vs specific files in those directories. For example, someone might move `config.yaml` to `backup-config.yaml` and be confused that their old config is getting loaded. Do we have a specific use case for this?

Also, `os.listdir(path)` returns filenames in arbitrary order, which will make for fun debugging! =P

napari/config.py
```diff
@@ -0,0 +1,877 @@
...
+    for name, value in env.items():
+        if name.startswith("NAPARI_"):
+            varname = name[7:].lower().replace("__", ".")
+            try:
+                d[varname] = ast.literal_eval(value)
+            except (SyntaxError, ValueError):
+                d[varname] = value
+
+    result: dict = {}
+    set(d, config=result, clean=True)
```
jni: Shadowing a built-in is naughty. I suggest using `_set` within this module and aliasing `set = _set` at the bottom, so that other importers can do `config.set`.

napari/config.py
```diff
@@ -0,0 +1,877 @@
...
+
+            try:
+                os.rename(tmp, destination)
+            except OSError:
+                os.remove(tmp)
+    except (IOError, OSError):
+        pass
+
+
+class set:
```
jni: Oh. God. :joy:

napari/config.py
```diff
@@ -0,0 +1,877 @@
...
+
+no_default = "__no_default__"
+
+
+paths = [
+    os.getenv("NAPARI_ROOT_CONFIG", "/etc/napari"),
+    os.path.join(sys.prefix, "etc", "napari"),
+    os.path.join(os.path.expanduser("~"), ".config", "napari"),
+    os.path.join(os.path.expanduser("~"), ".napari"),
+]
```
jni: ah, this was the bit I'd missed on my first review. =)

napari/config.py
```diff
@@ -0,0 +1,877 @@
...
+
+if "NAPARI_CONFIG" in os.environ:
+    PATH = os.environ["NAPARI_CONFIG"]
+    paths.append(PATH)
+else:
+    PATH = os.path.join(os.path.expanduser("~"), ".config", "napari")
```
jni: :+1:

napari/config.py
```diff
@@ -0,0 +1,877 @@
...
+
+    # update the config with the time of last modification
+    config['_last_synced'] = os.path.getmtime(destination)
+    return True
+
+
+refresh()
+
+# read in the default settings from this directory
+fn = os.path.join(os.path.dirname(__file__), "napari.yaml")
```
jni: I kinda dislike having this file in `napari/napari.yaml`. How about making a directory `napari/config/{config.py,napari.yaml,...}`, and in `napari/__init__.py` doing `from .config import config`?

jni (commented):
Ok so I've made a few comments, they should be easy-ish to address. I think the main thing is I'd like to move this stuff to a config/ folder. I don't love `napari/napari.yaml`. We had to do a cleanup of that directory once before already, so let's protect it.

Also, rename `set` to `_set` internally, even if `config.set` remains.

napari/config.py
tlambert03: it can!

napari/config.py
tlambert03: I think it's probably not the "most" native behavior, but it's also common enough that I always end up with a `/Users/talley/.config/` folder on windows.  works well enough, and could be modified to a more native approach in the future

napari/config.py
tlambert03: I moved everything into a folder, but ran into an issue with the constant `napari.config.PATH`, which should be modifiable at that location.  If I put the whole config file in `napari/config/__init__.py` it's fine, but if I put it all in `napari/config/config.py` then if a user changes `napari.config.PATH` (rather than `napari.config.config.PATH`) it breaks.  how much do you hate putting it all in `napari/config/__init__.py` vs jumping through hoops to avoid circular imports?

napari/config.py
tlambert03: If I'm understanding the intended [pattern for downstream libraries](https://docs.dask.org/en/latest/configuration.html#downstream-libraries), the use case is this:

1. developers create their own `napari_plugin.yaml` in their source code, with defaults under their namespace.
2. in their own `config.py` module, they register those defaults using `napari.config.update_defaults(defaults)`, but they _also_ run `napari.config.ensure_file(source=fn, comment=True)` which puts a fully commented-out copy of their `napari_plugin.yaml` file at `~/.config/napari/napari_plugin.yaml` ... that the user can explore and modify.
3. The developer can then use `napari.config.get('napari-plugin.key')` anywhere in their library.

I like this, and it only works if we look for _all_ yaml files in `~/.config/napari/`.  But I also do finding myself wanting an escape hatch.  How about we don't read yaml files prefaced with an underscore?  This doesn't fix your `backup-config.yaml` scenario (would need to be `_backup-config.yaml`), but I think the benefits for downstream namespaces outweigh the cost

napari/config.py
tlambert03: took a quick look at implementing with `@contextmanager`, but I think the additional `_assign` method on the class makes it uglier to implement this as a function

tlambert03:
ok, I think i hit all the main points, like not shadowing `set` and moving into a folder.  @jni, if you don't like having all the code in `napari/config/__init__.py` (which I don't love either), then can you help me figure out how to deal with the changing of the config.PATH variable and circular imports? (see specifically the test `test_ensure_file_defaults_to_NAPARI_CONFIG_directory`)

I also made two improvements (I think at least 😅):

1. By not reading private `_something.yaml` files, it partially solves the problem of having session (auto-saved) settings that don't overwrite "manually set" settings in the users `~/.config/napari/napari.yaml` (or other yaml) file.  I can write to `_session.yaml`, but then only use variables from the last session that are not explicitly set in the user settings.
2. I didn't like muddying the `config.config` dict with metadata about syncing status, so I created a `SyncStatus` object that tracks changes to objects, lets us clean up the sync() function a lot and provides a nice API:
```python
    >>> from napari import config
    >>> config.status
    <config.status CLEAN :: last synced 5.18 min ago to ~/.config/napari/_session.yaml>

    >>> config.set(a=1)
    >>> config.status.needs_sync()
    True
    >>> config.sync()
    True
    >>> config.status
    <config.status CLEAN :: last synced 0.29 min ago to ~/.config/napari/_session.yaml>

    >>> config.status.needs_sync()
    False
```

Here's an (admittedly hard to read) gif that gives you a sense for the read/write events and settings priority that goes on in the event loop (not the "`status CLEAN/DIRTY`" in the terminal at the top):

Untitled

napari/config.py
sofroniewn: I'll let @jni weigh in on this more, but personally I don't love having everything in `napari/config/__init__.py`. Maybe I don't understand how the whole config works, but can we not break this out into more files like we do elsewhere? It's ok if some stuff still lives in the `__init__.py` if that really does make things simpler, but does everything have too?

For example could we have a `config_utils.py` with some of the `dict` -> `dict` functions like `update` in there or a `str->str` function `canonical_name`.

Could we also have the classes be their own files like `_set`  and `SyncStatus`?

tlambert03: > can we not break this out into more files like we do elsewhere

probably yes, but it's made trickier by the fact that this config module is very much a *stateful* module, and many of the functions in the module take, as their default values, variables that are declared within the module (for instance: `def get(key, config=config)` is using the module level `config` variable as its default).
Combine that with the desire to keep the user-facing API simple (by importing everything in the `__init__.py`) and it makes it harder to _both_ avoid circular imports _and_ keep the API nice while preventing scenarios like this:
```python
# config/__init__.py
from .inner import PATH, show_path

# config/inner.py
PATH = 'original value'

def show_path(path=PATH):
    print(path)


In [1]: from napari import config
In [2]: config.show_path()
original value

In [3]: config.PATH
Out[3]: 'original value'

In [4]: config.PATH = 'new value'

In [5]: config.show_path()
original value
```

This is the case with `test_ensure_file_defaults_to_NAPARI_CONFIG_directory` that I mentioned in my other comment.  Now, we can definitely just say "we're not going to support changing PATH that way"... but we need to consciously decide that.

napari/_qt/qt_main_window.py
```diff
@@ -33,12 +37,38 @@
...
+    """MainWindow, tracks open windows and can do things when last is closed.
+    """
+
+    open_windows: List[QMainWindow] = []
+    pref_sycn_timer = QTimer()
```
sofroniewn: Why is this called `perf_sync_timer`, makes me think of the `perf_mon` stuff but it is unrelated right? More like `config_sync_timer` no?

Can you expand the class doc string to include `open_windows` and the QTimer too

napari/components/viewer_model.py
```diff
@@ -40,6 +41,7 @@ class ViewerModel(AddLayersMixin, KeymapHandler, KeymapProvider):
     """
 
     themes = palettes
+    theme_key = "main_window.theme"
```
sofroniewn: is this more illustrative right now, or is the plan to define config keys like this more generally? i.e. will we have `_key` attributes or maybe we can autogenerate them? Regardless, probably for a later PR (see my launch error problem below)

napari/components/viewer_model.py
```diff
@@ -78,7 +80,7 @@ def __init__(
         self.grid_stride = 1
 
         self._palette = None
-        self.theme = 'dark'
+        self.theme = config.get(self.theme_key, 'dark')
```
sofroniewn: maybe we drop this from this PR for merge and then have a followup dedicated to all this stuff, but it is nice to see one example. I've gotten myself in a little bind where I've defined the value of the key to be a non-recognized theme, and so get a start-up error
```
Theme 'lighs' not found; options are ['dark', 'light'].
```
I'm sure I'll figure out how to fix, but maybe best to leave out for now

sofroniewn: As expected fixing was easy (just edited the yaml) but something to be aware of for the future!!

napari/config/_tests/test_config.py
```diff
@@ -0,0 +1,549 @@
...
+    with open(destination) as f:
+        text = f.read()
+    assert "123" in text
+
+    with open(destination) as f:
+        result = yaml.safe_load(f)
+    assert not result
+
+
+def test_set():
```
sofroniewn, line 267: When I do

```python
config.set(main_window.theme='dark')
```

in the console I get this

```python-traceback
config.set(main_window.theme='dark')
  File "<ipython-input-7-95730f83dd74>", line 1
    config.set(main_window.theme='dark')
              ^
SyntaxError: keyword can't be an expression
```
similarly for
```python
config.set('main_window.theme'='dark')
```
but 
```python
config.set({'main_window.theme':'dark'})
```
works just fine. Not sure if this is expected, or I am doing something wrong

napari/config/_tests/test_config.py
```diff
@@ -0,0 +1,549 @@
...
+        config.ensure_file(source=source)
+    finally:
+        config.core.PATH = PATH
+
+    assert os.path.isdir(destination)
+    [fn] = os.listdir(destination)
+    assert os.path.split(fn)[1] == os.path.split(source)[1]
+
+
+def test_rename():
```
sofroniewn, line 366: If I have the following config in the console
```python
config.config
Out[36]: {'perfmon': 1, 'main_window': {'theme': 'dark'}}
```
and I do
```python
config.rename({'main_window': 'new_window'})
config.config
Out[38]: {'perfmon': 1, 'new_window': {'theme': 'dark'}}
```
that works as expected, but if I then try and rename `theme` I get the following
```python-traceback
config.rename({'new_window.theme': 'new_window.color'})
KeyError                                  Traceback (most recent call last)
<ipython-input-39-cc1db23ba1be> in <module>
----> 1 config.rename({'new_window.theme': 'new_window.color'})

~/GitHub/napari/napari/config/core.py in rename(aliases, config)
    507 
    508     for k in old:
--> 509         del config[canonical_name(k, config)]  # TODO: support nested keys
    510 
    511     _set(new, config=config)

KeyError: 'new_window.theme'
```
Should that have worked? Or I am doing something wrong?

napari/config/core.py
```diff
@@ -0,0 +1,689 @@
...
+if "NAPARI_CONFIG" in os.environ:
+    PATH = os.environ["NAPARI_CONFIG"]
+    paths.append(PATH)
+
+#: a special yaml file to dump settings that are changed within a session
+#:
+#: this file will be loaded, but takes lower priority than all non-private
+#: yaml files.  So if a user has a specific setting in their napari.yaml file,
+#: it will override anything in _session.yaml
+_SESSION = os.path.join(PATH, '_session.yaml')
```
sofroniewn, line 115: The `_session.yaml` is still not quite clear to me, it seems like this is where things I changed during my session got saved, which is fine, but then would the "factory" defaults be in `napari.yaml`, or those are stored commented and eventually the user will make changes to `napari.yaml` too? Maybe when the explicitly hit some sort of "save preferences" or something? (i.e. this lets us capture both those save in the background things like window position AND the save when we ask sort of things too?) I certainly don't mind the file, just want to make sure I understand.

sofroniewn (approved):
@tlambert03 I just went through this again, it's looking great - I think we should make any final changes and then merge. Juan gave the ok for this while he's out too. I think it would be nice to have it in and then start using it at various places in the codebase.

I've left various clarifying questions/ times when I got errors for things I thought would have worked, and requested dropping the `theme` example, but it's all pretty minor stuff. Is there anything else you'd like to put in here before merging?

napari/config/_tests/test_config.py
clbarnes: Isn't this just python? Keyword argument names have the same constraints as variable names, i.e. unambiguously not expressions. They can't include dots, because that resolves to `pre_dot.__getattr__(post_dot)`. Putting quotes around it would be interpreted as a string literal, also not a valid variable name. The last one, being a dict literal from str to str, is fine. It looks like hierarchical keys in keyword arguments can be separated with a double underscore instead (as is done later in the tests): `config.set(main_window__theme="dark")`.

tlambert03: Yes, @clbarnes has it right. @sofroniewn, if you want  to use the dot notation, it must be a `dict` key, as you found.  Otherwise you have to use the double underscore notation for nested keys

napari/_qt/qt_main_window.py
tlambert03: it's not :) it's called `"pref"_sync_timer` (for "preferences")... I think you had perfmon on the brain.  actually, it's called `pref_"sycn"_timer`, lol...  
In any case.  we don't really need the timer anymore if we're not doing 2-way sync with a file on disk that can written to from outside of the napari process.  if it's one-way, we can just write to disk whenever the config gets touched, without having to poll.

napari/components/viewer_model.py
tlambert03: illustrative.  no plan at the moment

napari/components/viewer_model.py
tlambert03: yeah, very good point!  In that case... I _think_ it's going to have to be the job of anything that is pulling a value from the config like that to sanitize it?  definitely something to be aware of

napari/config/_tests/test_config.py
tlambert03: nested keys aren't supported for `rename` yet ... (it's in the stack trace there: `# TODO: support nested keys`).  I don't think rename will be used all that often though

napari/config/core.py
tlambert03: - `napari.yaml`: factory defaults
- `~/.config/napari/napari.yaml`: user defaults ... get's copied (commented out) from `napari.yaml` on first run.  May be modified both programmatically and by the user.
- `_session.yaml` was where I was putting values that we do want to save from session to session, that are based on user interactivity, but which we don't necessarily want to be in `~/.config/napari/napari.yaml`.  Shouldn't be modified by the user.

An argument could be made that we just put _everything_ in the user config... but then the point of making it a human-readable yaml file starts to get a little lost.  There are almost certainly going to be values that we want to save, that we don't want a curious user stumbling upon and messing up state.  So think of `_session.yaml` as "private" settings, and `~/.config/napari/napari.yaml` as "public" settings.  I don't see it as depending that much on _when_ the preference is saved (i.e. hitting "save preferences" vs some background task) ... but rather _what_ is being saved (private setting vs public setting).

In a strict app ... all of this would be private, and the user would never look at the `config.yaml` file on disk: they would only interact with those settings in a GUI.
In a strict library, the user would be all over the `config.yaml` file ... using it to tune the library as desired... and the library probably wouldn't have any private settings like window position or "have we asked the user ____"

here, it's a bit confusing. #834  was a very private settings approach.  This one tends towards a very public settings approach.  Neither one feels 100% correct to me

napari/_qt/qt_main_window.py
jni: @tlambert03 so if we forget about 2-way sync (I think the last time we discussed things that was the plan?), then we can get rid of this class? That would be nice... =)

tlambert03: yep, that's vestigial

napari/config/core.py
```diff
@@ -0,0 +1,708 @@
...
+         with ``NAPARI_``
+   b. finally, ``collect()`` calls `merge()` on all of those discovered dicts,
+      to yield one merged result and returns it to the ``refresh()`` function
+      (by default, dicts discovered later in the process take priority).
+4. back in the ``refresh()`` function, that final merged result is used to
+   update the global config dict.
+5. At that point, the ``napari.yaml`` in the repo is read, if one doesn't
+   already exist, a file is created at ``~/.config/napari/napari.yaml`` (with
+   ``ensure_file()``) that is an exact copy of the repo ``napari.yaml`` but
+   with everything commented out, as a way to show users what they can change.
```
jni, line 46: This seems to come really late in the game? Shouldn't this be the first thing that happens?

napari/config/core.py
```diff
@@ -0,0 +1,708 @@
...
+6. finally, ``update_defaults()`` is called using the settings in the
+   `napari.yaml` file, which merges the configs again, but gives priority to
+   already existing values in the config (i.e. ones that were overwritten by
+   the user or the environement)
```
jni:
```suggestion
   the user or the environment)
```

napari/config/core.py
```diff
@@ -0,0 +1,708 @@
...
+   b. finally, ``collect()`` calls `merge()` on all of those discovered dicts,
+      to yield one merged result and returns it to the ``refresh()`` function
+      (by default, dicts discovered later in the process take priority).
+4. back in the ``refresh()`` function, that final merged result is used to
+   update the global config dict.
+5. At that point, the ``napari.yaml`` in the repo is read, if one doesn't
+   already exist, a file is created at ``~/.config/napari/napari.yaml`` (with
+   ``ensure_file()``) that is an exact copy of the repo ``napari.yaml`` but
+   with everything commented out, as a way to show users what they can change.
+6. finally, ``update_defaults()`` is called using the settings in the
```
jni, line 47: Why is it called "update defaults"?

napari/config/core.py
tlambert03: it doesn't really matter where `ensure_file` is called.  it's just creating a commented-out copy of the `napari.yaml` that ships with the repo at `~/.config/napari/napari.yaml` if one doesn't already exist there (so curious users can see what options are available to them).  It doesn't have any effect on the settings that are loaded.

tlambert03:
I think someone should take this PR over... I don't really "believe" in it anymore.  feels like we're forcing a square peg (library-like config file) into a round hole (application settings).  It's also more complicated than it needs to be for a first step.  If it were me, I'd probably dramatically simplify, and/or probably just use QSettings until it becomes clearer how we're going to be using this in practice.

this whole approach is designed for a read-only `config.yaml` file... where the library doesn't ever _write out_ changes in settings during the session (instead, a user is expected to directly edit their `config.yaml` file).  if we're also going to be writing to the user's `config.yaml` file (instead of some private `_session.yaml` file) then the commonly-used scenario of having a "commented out" `config.yaml` file that the user can partially uncomment and set (like `matplotlibrc` or `dask.yaml`) doesn't make as much sense, since on the very first run, we will have to override it to store what was changed by the user via the GUI during the session.  One can imagine some sort of logic to compare the current settings to default, and only write out uncommented settings to `config.yaml` if they are not the defaults, but that's a different set of functionality than what is present here.

napari/config/core.py
```diff
@@ -0,0 +1,708 @@
...
+
+
+#: The primary directory for user config files
+#:
+#: Defaults to ~/.config/napari
+#: May be set with the environment variable "NAPARI_CONFIG"
+PATH = os.path.join(os.path.expanduser("~"), ".config", "napari")
+if "NAPARI_CONFIG" in os.environ:
+    PATH = os.environ["NAPARI_CONFIG"]
+    paths.append(PATH)
```
Czaki, line 108: There should be an option to omit standard directories. To not mix production with development.

napari/config/core.py
```diff
@@ -0,0 +1,708 @@
...
+    result = config
+    for k in keys:
+        k = canonical_name(k, result)
+        try:
+            result = result[k]
+        except (TypeError, IndexError, KeyError):
+            if default is not no_default:
+                return default
+            else:
+                raise
```
Czaki, line 169: I could imagine case where function is called in two places with two different default values. So when value is not defined in configuration then there will be inconsistency in viewer state. I suggest to recursive create when default is provided.

napari/config/core.py
```diff
@@ -0,0 +1,708 @@
...
+    """Collect configuration from yaml files.
+
+    This searches through a list of paths, expands to find all yaml or json
+    files, and then parses each file.
+    """
+    logger.debug("collect_yaml from %r", paths)
+
+    # Find all paths
+    file_paths = []
+    for path in paths:
```
Czaki, line 378: I see that `paths` variable is not modified here, but mutable object should not be used as default argument

```suggestion
def collect_yaml(paths_list: Optional[List[str]] = None) -> List[dict]:
    """Collect configuration from yaml files.
    This searches through a list of paths, expands to find all yaml or json
    files, and then parses each file.
    """
    if paths_list is None:
        paths_list = paths
    logger.debug("collect_yaml from %r", paths)

    # Find all paths
    file_paths = []
    for path in paths_list:

File: napari/config/core.py [hunk 30]
```diff
@@ -0,0 +1,708 @@
...
+    os.path.join(os.path.expanduser("~"), ".config", "napari"),
+    os.path.join(os.path.expanduser("~"), ".napari"),
+]
+
+
+#: The primary directory for user config files
+#:
+#: Defaults to ~/.config/napari
+#: May be set with the environment variable "NAPARI_CONFIG"
+PATH = os.path.join(os.path.expanduser("~"), ".config", "napari")
```
Czaki, line 105: What with versioning? By default there should be different locations for stable one and development one. Maybe even python version may be important?

napari/config/core.py
```diff
@@ -0,0 +1,708 @@
...
+    file_paths = []
+    for path in paths:
+        if os.path.exists(path):
+            if os.path.isdir(path):
+                try:
+                    for fname in sorted(os.listdir(path)):
+                        base, ext = os.path.splitext(fname)
+                        if base.startswith("_"):
+                            continue
+                        if ext.lower() in (".json", ".yaml", ".yml"):
```
Czaki, line 386: why `.json` here? I do not found information that YAML could parse json file?

Czaki (commented):
Looks really nice. 
But I have few doubts:

1. How it will work with more than one napari instance running at same time?
2. How It will interact with tools that embed naprai viewer? 
3. What is strategy for multiple version support (for example for debug user reported error)

I also left few comments in code.

sofroniewn:
I'm wondering if we might want to use pydantic for our settings, https://pydantic-docs.helpmanual.io/usage/settings/ it supports nice json serialization to/from files. Could make sense if use use pydantic elsewhere - see #2032 for example

sofroniewn:
This has been superseded by #2212 and can now be closed. Thanks all for the great work and discussion here!

PR #1185, opened 2020-04-30
Save all/selected layers Qt dialogs

tlambert03:
# Description
Does what it sounds like.  No GUI-facing error handling is added yet... so it's a bare-bones implementation.  Will work on it more tomorrow.

## Type of change

# How has this been tested?

no tests at the moment, but minimal code...

## Final checklist:


napari/_qt/qt_viewer.py
```diff
@@ -284,6 +284,16 @@ def _update_camera(self):
                 self.view.camera.viewbox_key_event = viewbox_key_event
                 self.viewer.reset_view()
 
+    def _save_layers(self, selected=False):
+        """Save screenshot of current display, default .png"""
```
jni: ... Is this accurate?

napari/_qt/qt_main_window.py
```diff
@@ -157,7 +158,26 @@ def _add_file_menu(self):
...
+        save_all_layers.setShortcut('Ctrl+Shift+S')
+        save_all_layers.setStatusTip('Save all layers')
+        save_all_layers.triggered.connect(
+            lambda: self.qt_viewer._save_layers(selected=False)
+        )
+
+        save_selected_layers = QAction(
+            'Save selected layer(s)...', self._qt_window
+        )
+        # Save ctrl+S for saving the entire viewer state?
```
jni: As I recall we decided Ctrl-S for save selected layers, Ctrl-Shift-S for save all layers, and Ctrl-Alt-Shift-S for saving the viewer state.

napari/_qt/qt_main_window.py
```diff
@@ -157,7 +158,26 @@ def _add_file_menu(self):
...
+        save_selected_layers.setShortcut('Ctrl+Alt+Shift+S')
+        save_selected_layers.setStatusTip('Save selected layers')
+        save_selected_layers.triggered.connect(
+            lambda: self.qt_viewer._save_layers(selected=True)
+        )
+
+        # kinda want this shortcut for other things...
```
jni: Hilarious. What things? I'm not highly attached to this shortcut, but perhaps @sofroniewn is...

jni (dismissed):
@tlambert03 I think just another round of changes is needed.

Do we test GUI shortcuts at all?

napari/_qt/qt_main_window.py
sofroniewn: You mean `'Ctrl+Alt+S'` should not be for saving screenshot but something else? That's fine - do we have a suggestion for what saving screenshot should be? I'm open

tlambert03: I mostly just want to discuss it.  That's a pretty nice key combo, and I wonder how often someone needs to save a screenshot.  However, since everyone was happy with using Ctrl+S for saving layers, we're not running out of save key-combos.  Maybe someday, Ctrl+Alt+S can be for saving viewer state :)

sofroniewn: :-) let's leave for now then

sofroniewn: maybe remove the comment :-)

tlambert03: will do! :)

napari/_qt/qt_main_window.py
sofroniewn: If we're making `ctrl-S` the simpler choice for save-selected, should we also make `selected=True` the default value for `viewer.layers.save`? Under the same assumption of that is what people will want to do more frequently?

tlambert03: this i like less. I liked that `layers.save()` behaved as a composite variant on `layer.save()`, it also sets us up for using layer groups as temporary disposable collection of layers.

sofroniewn: makes sense

napari/_qt/qt_viewer.py
tlambert03: sure isn't

tlambert03:
ok, updated shortcuts, and added a helpful warning dialog if there are no layers in the viewer, or nothing was selected when using save selected.  

> Do we test GUI shortcuts at all?

coverage on that is pretty poor at the moment, but it's definitely doable (though more challenging) and I'll create an issue to follow up on it.  This is uncomplicated enough that i think it's safe to pass for now.

I started working on a save-with-plugin dialog... but it quickly got in the weeds with issues like "do we ask for the filename first?" If not, how do we restrict the filename?  etc... so let's add that later after we can discuss it.

napari/_qt/qt_main_window.py
```diff
@@ -157,7 +158,23 @@ def _add_file_menu(self):
...
+        )
+
+        save_all_layers = QAction('Save all layers...', self._qt_window)
+        save_all_layers.setShortcut('Ctrl+Shift+S')
+        save_all_layers.setStatusTip('Save all layers')
+        save_all_layers.triggered.connect(
+            lambda: self.qt_viewer._save_layers(selected=False)
+        )
+
+        screenshot = QAction('Save Screenshot', self._qt_window)
```
sofroniewn, line 177: When I do save screen shot I don't get the helpful status tip at top of dialog like i do for save all layers and save selected - see

image

vs.

image

Not sure what the difference is

sofroniewn (commented):
LGTM! Left one minor comment

sofroniewn:
Also do you mind updating from master if you make that above fix so all tests will pass again

napari/_qt/qt_main_window.py
tlambert03: added caption to the save_screenshot method

sofroniewn:
One question - how would people feel if we made `ctrl-alt-shift-S` save screenshot with the Qt viewer? i.e. `viewer.screenshot(path, with_viewer=True)`

I think its nice right now when people are tweeting/ sharing screenshots of napari that include the viewer - it helps us build our brand and so we might as well make it easy for people. see here for tweets that I have in mind https://twitter.com/notjustmoore/status/1256232842755014656

tlambert03:
> how would people feel if we made `ctrl-alt-shift-S` save screenshot with the Qt viewer? i.e. `viewer.screenshot(path, with_viewer=True)`

fine by me

sofroniewn (approved):
LGTM! I'm approving, we can also hold-off on the `with_viewer` screenshot depending on what @jni thinks

jni:
Don't we want to reserve Ctrl-Alt-Shift-S for saving the viewer state?

Maybe we could do just Alt-S for screenshot? It's the VLC shortcut for grabbing a frame.

jni:
I guess the answer to my question will have to be decided in a future PR. :grinning:

sofroniewn:
I'm open to `Alt-S` for screenshot and `Atl-Shift-S` for screenshot with viewer. Or just adding `Alt-S`. We could do a quick PR with that right now? Agreed it's not urgent, but if we all agree ...

PR #1187, opened 2020-04-30
Update manifest.in for plugin code removal

tlambert03:
# Description
This just updates manifest.in and removes a couple vestigial blocks of plugin code after the `napari-plugin-engine` merge.


sofroniewn (approved):
LGTM! will merge

napari/plugins/_tests/test_builtin_get_writer.py
```diff
@@ -45,7 +46,7 @@ def test_get_writer_succeeds(
 
 
 # the layer_data_and_types fixture is defined in napari/conftest.py
-# the plugin_manager fixture is defined in napari/plugins/_tests/conftest.py
+# test_plugin_manager fixture is provided by napari_plugin_engine._testsupport
```
jni, line 49: How does this work now? Does pytest discover *all* installed fixtures???

tlambert03: It works because of the “pytest11” entry point in the setup.py file of napari-plugin-engine.  In this case, npe is acting like a pytest plugin (just like the qtbot fixture that we use from pytest-qt)

jni: Sheesh. Can you expand your comment to mention this?

God help us. =P

tlambert03: i can, but where?  every external fixture like qtbot too?  Can't this just be one of those "read the pytest docs" moments?  I know it's an unfortunate pattern... but I don't know where to comment on it.  you happened to find it in this particular function, but that phenomenon is going on all over the place with pytest

tlambert03: I'll just add a note here that pytest discovers plugins and fixtures as described [here](https://docs.pytest.org/en/2.7.3/plugins.html#plugin-discovery-order-at-tool-startup), and if you want to see all the fixtures used in your tests, and where they come from run `pytest --fixtures` this also works narrowed to a particular file or directory, for instance:

```shell
pytest napari/plugins --fixtures
```

<details>
<summary>Show Output</summary>

```
========================================== test session starts ===========================================
platform darwin -- Python 3.7.5, pytest-5.3.2, py-1.8.0, pluggy-0.13.1
benchmark: 3.2.3 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
PySide2 5.14.0 -- Qt runtime 5.14.0 -- Qt compiled 5.14.0
plugins: timeout-1.3.3, ordering-0.6, benchmark-3.2.3, qt-3.3.0, cov-2.8.1, napari-plugin-engine-0.1.5.dev3+g932cb49
collected 55 items                                                                                       
cache
    Return a cache object that can persist state between testing sessions.
    
    cache.get(key, default)
    cache.set(key, value)
    
    Keys must be a ``/`` separated value, where the first part is usually the
    name of your plugin or application to avoid clashes with other cache users.
    
    Values can be any object handled by the json stdlib module.

capsys
    Enable text capturing of writes to ``sys.stdout`` and ``sys.stderr``.
    
    The captured output is made available via ``capsys.readouterr()`` method
    calls, which return a ``(out, err)`` namedtuple.
    ``out`` and ``err`` will be ``text`` objects.

capsysbinary
    Enable bytes capturing of writes to ``sys.stdout`` and ``sys.stderr``.
    
    The captured output is made available via ``capsysbinary.readouterr()``
    method calls, which return a ``(out, err)`` namedtuple.
    ``out`` and ``err`` will be ``bytes`` objects.

capfd
    Enable text capturing of writes to file descriptors ``1`` and ``2``.
    
    The captured output is made available via ``capfd.readouterr()`` method
    calls, which return a ``(out, err)`` namedtuple.
    ``out`` and ``err`` will be ``text`` objects.

capfdbinary
    Enable bytes capturing of writes to file descriptors ``1`` and ``2``.
    
    The captured output is made available via ``capfd.readouterr()`` method
    calls, which return a ``(out, err)`` namedtuple.
    ``out`` and ``err`` will be ``byte`` objects.

doctest_namespace [session scope]
    Fixture that returns a :py:class:`dict` that will be injected into the namespace of doctests.

pytestconfig [session scope]
    Session-scoped fixture that returns the :class:`_pytest.config.Config` object.
    
    Example::
    
        def test_foo(pytestconfig):
            if pytestconfig.getoption("verbose") > 0:
                ...

record_property
    Add an extra properties the calling test.
    User properties become part of the test report and are available to the
    configured reporters, like JUnit XML.
    The fixture is callable with ``(name, value)``, with value being automatically
    xml-encoded.
    
    Example::
    
        def test_function(record_property):
            record_property("example_key", 1)

record_xml_attribute
    Add extra xml attributes to the tag for the calling test.
    The fixture is callable with ``(name, value)``, with value being
    automatically xml-encoded

record_testsuite_property [session scope]
    Records a new ``<property>`` tag as child of the root ``<testsuite>``. This is suitable to
    writing global information regarding the entire test suite, and is compatible with ``xunit2`` JUnit family.
    
    This is a ``session``-scoped fixture which is called with ``(name, value)``. Example:
    
    .. code-block:: python
    
        def test_foo(record_testsuite_property):
            record_testsuite_property("ARCH", "PPC")
            record_testsuite_property("STORAGE_TYPE", "CEPH")
    
    ``name`` must be a string, ``value`` will be converted to a string and properly xml-escaped.

caplog
    Access and control log capturing.
    
    Captured logs are available through the following properties/methods::
    
    * caplog.messages        -> list of format-interpolated log messages
    * caplog.text            -> string containing formatted log output
    * caplog.records         -> list of logging.LogRecord instances
    * caplog.record_tuples   -> list of (logger_name, level, message) tuples
    * caplog.clear()         -> clear captured records and formatted log output string

monkeypatch
    The returned ``monkeypatch`` fixture provides these
    helper methods to modify objects, dictionaries or os.environ::
    
        monkeypatch.setattr(obj, name, value, raising=True)
        monkeypatch.delattr(obj, name, raising=True)
        monkeypatch.setitem(mapping, name, value)
        monkeypatch.delitem(obj, name, raising=True)
        monkeypatch.setenv(name, value, prepend=False)
        monkeypatch.delenv(name, raising=True)
        monkeypatch.syspath_prepend(path)
        monkeypatch.chdir(path)
    
    All modifications will be undone after the requesting
    test function or fixture has finished. The ``raising``
    parameter determines if a KeyError or AttributeError
    will be raised if the set/deletion operation has no target.

recwarn
    Return a :class:`WarningsRecorder` instance that records all warnings emitted by test functions.
    
    See http://docs.python.org/library/warnings.html for information
    on warning categories.

tmpdir_factory [session scope]
    Return a :class:`_pytest.tmpdir.TempdirFactory` instance for the test session.

tmp_path_factory [session scope]
    Return a :class:`_pytest.tmpdir.TempPathFactory` instance for the test session.

tmpdir
    Return a temporary directory path object
    which is unique to each test function invocation,
    created as a sub directory of the base temporary
    directory.  The returned object is a `py.path.local`_
    path object.
    
    .. _`py.path.local`: https://py.readthedocs.io/en/latest/path.html

tmp_path
    Return a temporary directory path object
    which is unique to each test function invocation,
    created as a sub directory of the base temporary
    directory.  The returned object is a :class:`pathlib.Path`
    object.
    
    .. note::
    
        in python < 3.6 this is a pathlib2.Path


------------------------ fixtures defined from napari_plugin_engine._testsupport -------------------------
temporary_hookimpl
    A fixture that can be used to insert a HookImplementation in the hook call loop.
    
    Use as a context manager, which will return the hook_caller for the
    corresponding hook specification.
    
    Example
    -------
    .. code-block: python
    
        def my_hook_implementation(arg):
            raise ValueError("shoot!")
    
        with temporary_hookimpl(my_hook_implementation) as hook_caller:
            with pytest.raises(PluginCallError):
                hook_caller(arg=42)

test_plugin_manager
    A plugin manager fixture with the project name 'test'.

add_specification
    Return a decorator that adds a HookSpecification to test_plugin_manager.

add_implementation
    Return a decorator that adds a HookImplementation to test_plugin_manager.

caller_from_implementation
    Return hook caller with implementation as its own spec definition.
    
    Adds a specification and implementation to the test_plugin_manager based on
    a single function definition (e.g. assumes that the implementation has the
    correct signature).  Returns the hook caller instance prepopulated with the
    hook implementation.


----------------------------- fixtures defined from pytest_benchmark.plugin ------------------------------
benchmark
    /Users/talley/miniconda3/envs/napdev/lib/python3.7/site-packages/pytest_benchmark/plugin.py:391: no docstring available

benchmark_weave
    /Users/talley/miniconda3/envs/napdev/lib/python3.7/site-packages/pytest_benchmark/plugin.py:415: no docstring available


-------------------------------- fixtures defined from pytest_cov.plugin ---------------------------------
no_cover
    A pytest fixture to disable coverage.

cov
    A pytest fixture to provide access to the underlying coverage object.


--------------------------------- fixtures defined from pytestqt.plugin ----------------------------------
qapp_args [session scope]
    Fixture that provides QApplication arguments to use.
    
    You can override this fixture to pass different arguments to
    ``QApplication``:
    
    .. code-block:: python
    
       @pytest.fixture(scope='session')
       def qapp_args():
           return ['--arg']

qapp [session scope]
    Fixture that instantiates the QApplication instance that will be used by
    the tests.
    
    You can use the ``qapp`` fixture in tests which require a ``QApplication``
    to run, but where you don't need full ``qtbot`` functionality.

qtbot
    Fixture used to create a QtBot instance for using during testing.
    
    Make sure to call addWidget for each top-level widget you create to ensure
    that they are properly closed after the test ends.

qtlog
    Fixture that can access messages captured during testing

qtmodeltester
    Fixture used to create a ModelTester instance to test models.


--------------------------------- fixtures defined from napari.conftest ----------------------------------
layer_data_and_types
    Fixture that provides some layers and filenames
    
    Returns
    -------
    tuple
        ``layers, layer_data, layer_types, filenames``
    
        - layers: some image and points layers
        - layer_data: same as above but in LayerData form
        - layer_types: list of strings with type of layer
        - filenames: the expected filenames with extensions for the layers.

layer
    Parameterized fixture that supplies a layer for testing.
    
    Parameters
    ----------
    request : _pytest.fixtures.SubRequest
        The pytest request object
    
    Returns
    -------
    napari.layers.Layer
        The desired napari Layer.

layers
    Fixture that supplies a layers list for testing.
    
    Returns
    -------
    napari.components.LayerList
        The desired napari LayerList.

qtbot
    A modified qtbot fixture that makes sure no widgets have been leaked.

viewer_factory
    napari/conftest.py:50: no docstring available

layer_writer_and_data
    Fixture that supplies layer io utilities for tests.
    
    Parameters
    ----------
    request : _pytest.fixtures.SubRequest
        The pytest request object
    
    Returns
    -------
    tuple
        ``(writer, layer_data, extension, reader, Layer)``
    
        - writer: a function that can write layerdata to a path
        - layer_data: the layerdata tuple for this layer
        - extension: an appropriate extension for this layer type
        - reader: a function that can read this layer type from a path and
                  returns a ``(data, meta)`` tuple.
        - Layer: the Layer class


-------------------------- fixtures defined from napari.plugins._tests.conftest --------------------------
plugin_manager
    PluginManager fixture that loads some test plugins

builtin_plugin_manager
    napari/plugins/_tests/conftest.py:27: no docstring available


========================================= no tests ran in 0.05s ==========================================
```

</details>

sofroniewn: This works for me. I will merge shortly

PR #1188, opened 2020-04-30
Warn when nothing saved

tlambert03:
# Description
This adds a warning if `save_layers` is called but nothing is written.  Could make it an exception instead if you prefer.  (Also fixing a `PluginCallError()` signature)

## Type of change

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/plugins/io.py
+++ b/napari/plugins/io.py
@@ -141,6 +141,8 @@ def save_layers(
             path, layers[0], plugin_name=plugin
         )
         written = [written] if written else []
+    else:
+        written = []
 
     if not written:
         # if written is empty, it means no plugin could write the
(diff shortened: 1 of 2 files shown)


sofroniewn (approved):
LGTM!

napari/plugins/io.py
```diff
@@ -132,15 +133,25 @@ def save_layers(
...
-    if len(layers) == 1:
+    elif len(layers) == 1:
         written = _write_single_layer_with_plugins(
             path, layers[0], plugin_name=plugin
         )
-        return [written] if written else []
-    return []
+        written = [written] if written else []
+
+    if not written:
```
jni, line 147: this will be an NameError if len(layers) == 0

tlambert03: thank you!  fixed, and added a test

PR #1192, opened 2020-05-02
fix multichannel implicit multiscale

### LINKED ISSUE #1117: Implicit pyramids will cause AttributeError in add_image(channel_axis=...)
## 🐛 Bug
as @jni pointed out here https://github.com/napari/napari/pull/1092#discussion_r404757426, we have a bug in `Viewer.add_image()`:  If the user passes an implicit pyramid (a `list` of arrays _withouth_ explicitly using `is_pyramid=True`) and tries to use `channel_axis != None`, then there will be an `AttributeError` here, since a `list` does not have `.shape`:
https://github.com/napari/napari/blob/871825e9ca9f570575e0681132d638e03cb39708/napari/components/add_layers_mixin.py#L227

(note, the syntax of of that line changed in #1092, but the potential error was always there).

sofroniewn:
# Description
This PR closes #1117 by fixing the implicit multiscale with a multichannel by using the `guess_multiscale` function if none is provided.

Given the usage of `guess_multiscale` here we might want to move it out of the `image` layer utils into our more general utils, we could also rename it `guess_image_multiscale` to be more explicit.

A test has been added.

## Type of change

### DIFF
--- a/napari/components/_tests/test_multichannel.py
+++ b/napari/components/_tests/test_multichannel.py
@@ -147,6 +147,28 @@ def test_multichannel_multiscale():
         assert viewer.layers[i].colormap[0] == base_colormaps[i]
 
 
+def test_multichannel_implicit_multiscale():
+    """Test adding multichannel implicit multiscale."""
+    viewer = ViewerModel()
+    np.random.seed(0)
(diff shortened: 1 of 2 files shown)


tlambert03 (approved):
LGTM!

jni (approved):

PR #1193, opened 2020-05-02
Builtin shapes writer and reader

sofroniewn:
# Description
This PR follows on from work started in #941 by @GenevieveBuckley by adding support for writing shape files to csv as a builtin plugin following a similar pattern used for points.

It additionally creates a `read_points_csv` and `read_shapes_csv` that can be used for a roundtrip. This now opens the door for usage with our reader plugins, but curious @tlambert03 how you think we should go about this.

Right now each reader doesn't check they are getting a csv and only checks the data is correct after reading the entire file! The better approach would be to just read the first row and look at the column names and then decide. I think we could probably merge this as is and then make a follow-up PR that integrated this with the builtin and included things like fast checking, but I am open

## Type of change


napari/plugins/_builtins.py
```diff
@@ -122,6 +122,78 @@ def napari_write_points(path: str, data: Any, meta: dict) -> Optional[str]:
...
+    if ext == '':
+        path = path + '.csv'
+    elif ext != '.csv':
+        # If an extension is provided then it must be `.csv`
+        return
+
+    if 'shape_type' in meta:
+        shape_type = meta['shape_type']
+    else:
+        shape_type = ['rectangle'] * len(data)
```
tlambert03, line 204: just curious, will there ever be an `as_layer_data_tuple()` meta that does _not_ have `"shape_type"`?

napari/plugins/_builtins.py
```diff
@@ -122,6 +122,78 @@ def napari_write_points(path: str, data: Any, meta: dict) -> Optional[str]:
     return path
 
 
+@napari_hook_implementation(trylast=True)
+def napari_write_shapes(path: str, data: Any, meta: dict) -> Optional[str]:
+    """Our internal fallback points writer at the end of the plugin chain.
```
tlambert03, line 174: I know this is following the other docstrings, but In case these things change or get resorted, maybe we shouldn't specify exactly where in the call chain it's going to occur?

napari/utils/io.py
```diff
@@ -312,5 +312,85 @@ def read_csv(filename: str) -> Tuple[np.array, List[str]]:
...
+            try:
+                values = np.array(values).astype('float')
+            except ValueError:
+                pass
+            meta['properties'][column_names[ind]] = values
+
+    return data, meta
+
+
+def read_shapes_csv(filename: str) -> Tuple[np.array, dict]:
```
tlambert03: might as well have this return a `napari.types.FullLayerData`  of `(array, dict, 'shapes')`

napari/utils/io.py
```diff
@@ -312,5 +312,85 @@ def read_csv(filename: str) -> Tuple[np.array, List[str]]:
             if row_index == 0:
                 column_names = [str(i) for i in row]
             else:
-                output_data.append([float(i) for i in row])
+                output_data.append([i for i in row])
         return np.array(output_data), column_names
+
+
+def read_points_csv(filename: str) -> Tuple[np.array, dict]:
```
tlambert03: might as well have this return a `napari.types.FullLayerData` of `(array, dict, 'points')`

napari/utils/io.py
```diff
@@ -312,5 +312,85 @@ def read_csv(filename: str) -> Tuple[np.array, List[str]]:
...
+    if n_shapes != len(shape_boundaries) - 1:
+        raise ValueError('Expected number of shapes not found')
+
+    data = []
+    shape_type = []
+    for ind_a, ind_b in zip(shape_boundaries[:-1], shape_boundaries[1:]):
+        data.append(raw_data[ind_a:ind_b])
+        shape_type.append(table[ind_a, 1])
+
+    return data, {'shape_type': shape_type}
```
tlambert03: these are not hooked up for our builtin reader, right?  Might be nice to make a combined function:
```python
def csv_to_layer_data(filename: str) -> List[FullLayerData]:
    for reader in (read_shapes_csv, read_points_csv):
        try:
            # this is inefficient cause it will read each time, so you might want to
            # pull out the recognition logic from each function.
            return [reader(filename)]
        except ValueError:
            pass
    raise ValueError("unrecognized CSV format")
```

and then change our `_buitlins.napari_get_reader` to 
```python
    if isinstance(path, str) and path.endswith('csv'):
        return csv_to_layer_data
    return image_reader_to_layerdata_reader(magic_imread)
```

tlambert03 (commented):
looking good @sofroniewn,  left a few minor comments... Was there a specific reason that you didn't actually hook up the csv reader functions?  I don't think it would take much, and would be nice to be able to drop a shapes csv onto the viewer...  I left a suggestion.  Though, if you were specifically avoiding, let me know

napari/utils/io.py
tlambert03: following up on this, we are losing a lot of the nicety of the `csv.reader` by reading the full file just to check the headers in each function.  Here's an alternative:
```python
def iter_csv(filename: str):
    with open(filename, newline='') as csvfile:
        reader = csv.reader(csvfile, delimiter=',')
        yield from reader

# then you don't need to read the full file just to check the column names
gen = iter_csv('file.csv')
header = next(gen)
if i_dont_like_the(header):
    return
# otherwise continue iterating...
data = np.array(list(gen))
```

(note, you can also recreate the output of the current `read_csv` function as follows)
```python
def read_csv(filename):
    column_names, *data = iter_csv(filename)
    return np.array(data), column_names
```

then the `csv_to_layer_data` is more like:

```python
def csv_to_layer_data(filename: str) -> List[FullLayerData]:
    gen = iter_csv('file.csv')
    header = next(gen)
    layer_type = guess_layer_type_from_header(header)
    if layer_type:
        return [reader_functions[layer_type]('file.csv')]
```

napari/plugins/_builtins.py
sofroniewn: Agreed, but maybe we wait to change them all together. I think we might even to want to pull all these out into their own repo for `0.3.1`

napari/plugins/_builtins.py
sofroniewn: there won't be, but I like having functions here that can accept an empty meta dict and still behave, makes them more reusable

napari/utils/io.py
sofroniewn: good idea, done

napari/utils/io.py
sofroniewn: done

napari/utils/io.py
sofroniewn: nice suggestions - incorporated

sofroniewn:
@tlambert03 I have incorporated your suggestions of adding reading the header of the csv first and including the csv reading in our builtin

napari/utils/io.py
```diff
@@ -305,12 +352,121 @@ def read_csv(filename: str) -> Tuple[np.array, List[str]]:
...
+    ----------
+    column_names : list of str
+        List of the column names from the csv.
+
+    Returns
+    -------
+    str or None
+        Layer type if recognized, otherwise None.
+    """
+    if column_names[:2] == ['index', 'axis-0']:
```
tlambert03: will raise IndexError if `len(column_names) < 2`

napari/utils/io.py
```diff
@@ -5,6 +5,7 @@
 from glob import glob
 from pathlib import Path
 from typing import Union, List, Optional, Tuple
```
tlambert03:
```suggestion
from typing import Union, List, Optional, Tuple, Generator
```
(If you choose to use the type hint below)

napari/utils/io.py
```diff
@@ -290,13 +291,59 @@ def write_csv(
...
+    layer_type = guess_layer_type_from_column_names(column_names)
+    if layer_type:
+        print('asdf')
+        print(csv_reader_functions[layer_type])
+        return [csv_reader_functions[layer_type](filename)]
+    else:
+        return None
+
+
+def iter_csv(filename: str):
```
tlambert03:
```suggestion
def iter_csv(filename: str) -> Generator[List[str], None, None]:
```

napari/utils/io.py
```diff
@@ -290,13 +291,59 @@ def write_csv(
...
+
+    Parameters
+    ----------
+    filename : str
+        Filename for reading csv.
+
+    Returns
+    -------
+    generator
+        Generator for reading lines from a csv
```
tlambert03:
```suggestion
    Yields
    ------
    list of str
        List of strings for each line in the csv file
```

napari/utils/io.py
```diff
@@ -290,13 +291,59 @@ def write_csv(
...
+    if not filename.endswith('.csv'):
+        return None
+
+    gen = iter_csv(filename)
+    column_names = next(gen)
+    layer_type = guess_layer_type_from_column_names(column_names)
+    if layer_type:
+        print('asdf')
+        print(csv_reader_functions[layer_type])
+        return [csv_reader_functions[layer_type](filename)]
```
tlambert03:
```suggestion
    if layer_type in csv_reader_functions:
        return [csv_reader_functions[layer_type](filename)]
```

napari/utils/io.py
```diff
@@ -290,13 +291,59 @@ def write_csv(
...
+        Filename for reading csv.
+
+    Returns
+    -------
+    list of napari.types.FullLayerData or None
+        Layer full layer data or None.
+    """
+    # If extension is not csv then cannot read file
+    if not filename.endswith('.csv'):
+        return None
```
tlambert03:
```suggestion
```
since the only place this is called at the moment is `napari_get_reader`, where we have just checked this already.  I'm not sure it's necessary?  (but if we do keep it, should return an empty list instead of None)

napari/utils/io.py
```diff
@@ -290,13 +291,59 @@ def write_csv(
...
+
+    gen = iter_csv(filename)
+    column_names = next(gen)
+    layer_type = guess_layer_type_from_column_names(column_names)
+    if layer_type:
+        print('asdf')
+        print(csv_reader_functions[layer_type])
+        return [csv_reader_functions[layer_type](filename)]
+    else:
+        return None
```
tlambert03, line 318:
```suggestion
        return []
```
(`ReaderFunctions` don't return `Optional[List[LayerData]]`)

napari/utils/io.py
```diff
@@ -290,13 +291,59 @@ def write_csv(
...
+    -------
+    generator
+        Generator for reading lines from a csv
+    """
+    with open(filename, newline='') as csvfile:
+        reader = csv.reader(csvfile, delimiter=',')
+        yield from reader
+
+
 def read_csv(filename: str) -> Tuple[np.array, List[str]]:
```
tlambert03: could consider removing `read_csv` altogether, I think the only place it is currently used is in `test_read_csv`

napari/utils/io.py
```diff
@@ -305,12 +352,121 @@ def read_csv(filename: str) -> Tuple[np.array, List[str]]:
...
-            if row_index == 0:
-                column_names = [str(i) for i in row]
-            else:
-                output_data.append([float(i) for i in row])
-        return np.array(output_data), column_names
+    column_names, *data = iter_csv(filename)
+    return np.array(data), column_names
+
+
+def guess_layer_type_from_column_names(column_names):
```
tlambert03:
```suggestion
def guess_layer_type_from_column_names(column_names: List[str]) -> Optional[str]:
```
(If you're so inclined 😄 )

napari/utils/io.py
```diff
@@ -305,12 +352,121 @@ def read_csv(filename: str) -> Tuple[np.array, List[str]]:
...
+        Filename for reading csv.
+
+    Returns
+    -------
+    data : array
+        Points data.
+    meta : dict
+        Points metadata.
+    layer_type : str
+        Points layer type.
```
tlambert03:
```suggestion
    layer_data : tuple
        3-tuple ``(array, dict, str)`` (points data, metadata, 'points')
```

napari/utils/io.py
```diff
@@ -305,12 +352,121 @@ def read_csv(filename: str) -> Tuple[np.array, List[str]]:
...
+        Filename for reading csv.
+
+    Returns
+    -------
+    data : list of array
+        Shapes data.
+    meta : dict
+        Shapes metadata.
+    layer_type : str
+        Shapes layer type.
```
tlambert03:
```suggestion
```suggestion
    layer_data : tuple
        3-tuple ``(array, dict, str)`` (points data, metadata, 'shapes')
```

tlambert03 (approved):
LGTM!  It's fun to be able to drag a CSV file into the window and get some shapes 😄 

I left a lot of suggestions, but they are mostly niggly and/or type hints... feel free to take or leave as desired.

napari/utils/io.py
sofroniewn: Agreed - it might be useful though, I'm inclined to keep it for now

napari/utils/io.py
sofroniewn: Good catch - have fixed

sofroniewn:
Thanks for all the type/ docs suggestion - accepted all of them, have also added code to prevent the index error. I think this is ready for final review from @jni and then merge

sofroniewn:
For `points` I have also added `property-` in front of the names of the columns in the properties dict if they are provided, this will prevent weird name collisions and make it much easier for us to detect what are properties vs something else. I feel like being explicit here will be better for us in the long run. The read method strips these off appropriately.

One question is - should we add in colors etc to saving of the points and shapes layers?
I know it was discussed earlier and there was some hesitation - but I'm playing with the thing now and I just find it strange that when you save out the layer and drag it back you loose all the coloring information. Thoughts @jni? @GenevieveBuckley? Maybe this can be done in `0.3.1` though

napari/plugins/_builtins.py
```diff
@@ -113,7 +113,8 @@ def napari_write_points(path: str, data: Any, meta: dict) -> Optional[str]:
     # construct table from data
     column_names = ['axis-' + str(n) for n in range(data.shape[1])]
     if properties:
-        column_names += properties.keys()
+        prop_names = ['property-' + p for p in properties.keys()]
```
jni: I'm sorry, I super hate this.

jni (changes requested):
@sofroniewn Can this wait for 0.3.1? I'd like to have a discussion about the best way to deal with properties. I really dislike the current approach. People will want their column names preserved when e.g. analysing the data with pandas in downstream scripts.

jni:
> I feel like being explicit here will be better for us in the long run. 

I think whatever method we choose needs to be maximally compatible with downstream and upstream analysis tools. It's extremely unlikely that those tools will want to prefix things with `property-` or anything else.

The main issue is that csv is a very limited format, and you can't attach metadata to it. The equivalent to this method is wanting to save to `.npy` (bare array), and adding one or more rows/slices to the image where you encode metadata. It's just very unfriendly to any other tool. The solution is using some io format such as feather/parquet that has more features where we could attach napari-specific metadata. For csv or metadata-free arrays, the correct solution is not to mangle names, but to provide UI elements for e.g. assigning columns to facets.

I think this should wait for 0.3.1+ for a proper implementation, so we can have a discussion about things. We need to put axis labels in the saved data anyway, so it's not like we're one merge away from the perfect csv IO.

sofroniewn:
Ok that's fair enough @jni - I will roll back `property-` as discussed on zulip we want to minimize friction for people loading/ saving these things out of excel. Ultimately though we might need a cool widget that allows someone to specify what columns correspond to what parts of the layer model, but that can come later. It will be particularly important though when we let people have custom names axes as column names

napari/plugins/_builtins.py
sofroniewn: dropped ;-)

napari/utils/_tests/test_io.py
```diff
@@ -196,5 +196,18 @@ def test_read_csv(tmpdir):
...
     # Read csv file
     read_data, read_column_names = io.read_csv(expected_filename)
+    read_data = np.array(read_data).astype('float')
     np.testing.assert_allclose(expected_data, read_data)
+
     assert column_names == read_column_names
+
+
+def test_guess_layer_type_from_column_names():
+    points_names = ['index', 'axis-0', 'axis-1']
```
jni, line 206: Why do we need an index for points? It just makes the table more annoying to work with.

napari/utils/io.py
```diff
@@ -306,12 +348,121 @@ def read_csv(filename: str) -> Tuple[np.array, List[str]]:
...
+    prop_axes[0] = False
+    meta = {}
+    if np.any(prop_axes):
+        meta['properties'] = {}
+        for ind in np.nonzero(prop_axes)[0]:
+            values = table[:, ind]
+            try:
+                values = np.array(values).astype('float')
+            except ValueError:
+                pass
```
jni: This approach is so brittle... integer values are important to preserve. This is one reason why I mentioned we'd probably want to depend on pandas eventually for csvs.

I would suggest, alternatively, that we treat the csv file as a list of *columns*, each of which is separately cast to a numpy array — same as properties. Something like:

```python
columns = [[] for name in column_names]
for row in gen:
    for elem, col in zip(row, columns):
        col.append(elem)
table = {
    colname: np.array(col) for colname, col in zip(column_names, columns)
}
```

This will mean, at least, that our array type inference will match NumPy's, rather than "float if possible, object otherwise".

jni: We might also want to add a function to canonize column names, as duplicate column names will make this approach explode. We can mitigate this by e.g. separating column dicts between axis columns and property columns *before* iterating on the table. Then, on each row, you can send part of the row to the coordinates table and part to the properties table.

napari/utils/_tests/test_io.py
jni: I just saw that I missed this in #1104. My suggestion would be to get rid of the index altogether for points layers.

napari/utils/io.py
```diff
@@ -291,13 +292,54 @@ def write_csv(
...
+        Layer full layer data or None.
+    """
+
+    gen = iter_csv(filename)
+    column_names = next(gen)
+    layer_type = guess_layer_type_from_column_names(column_names)
+    if layer_type in csv_reader_functions:
+        return [csv_reader_functions[layer_type](filename)]
+    else:
+        return []
```
jni: Just for my own understanding, this function will conceivably be modified in the future to be able to read arbitrary csvs? It might be useful to e.g. be able to read a table of numbers (no headers whatsoever) as a coordinates table.

napari/utils/io.py
```diff
@@ -291,13 +292,54 @@ def write_csv(
...
+        Filename for reading csv.
+
+    Returns
+    -------
+    list of napari.types.FullLayerData or None
+        Layer full layer data or None.
+    """
+
+    gen = iter_csv(filename)
+    column_names = next(gen)
```
jni: Although I initially liked the generator approach, I'm concerned that this might leave the file open until Python quits. I don't know whether Python will be smart enough to garbage collect the generator when it falls out of scope.

See [here](http://scipy-lectures.org/advanced/advanced_python/index.html#bidirectional-communication) that generators have a `.close()` method, so it might be worth adding:

```suggestion
    gen = iter_csv(filename)
    column_names = next(gen)
    gen.close()
```

jni: For completeness' sake, here's the walrus equivalent, which we can't use because you need Py3.8, but it's still nice to note the utility of it :smiley: :

```python
column_names = next(gen := iter_csv(filename))
gen.close()
```

It might also be worth making `iter_csv` itself a context manager...

napari/utils/io.py
```diff
@@ -306,12 +348,121 @@ def read_csv(filename: str) -> Tuple[np.array, List[str]]:
...
-        reader = csv.reader(csvfile, delimiter=',')
-        output_data = []
-        for row_index, row in enumerate(reader):
-            if row_index == 0:
-                column_names = [str(i) for i in row]
-            else:
-                output_data.append([float(i) for i in row])
-        return np.array(output_data), column_names
+    column_names, *data = iter_csv(filename)
+    return np.array(data), column_names
```
jni: I kinda feel like `read_csv` should be what `csv_to_layer_data` is now...

jni (changes requested):
@sofroniewn this still needs at least one more round of revisions imho. The most urgent thing for me is that I think we should not write/read the `index` column from points. It is pointless (heh), since it can be trivially replaced with `np.arange(len(data))`. My bad for missing it in #1104, but at least it's not in any release.

So now I kind of agree that 0.3.0 should wait for this! :sweat_smile: 

I should have a bit of time tomorrow morning (ie starting 4pm PST) to work together to get this in.

napari/utils/io.py
tlambert03: > See [here](http://scipy-lectures.org/advanced/advanced_python/index.html#bidirectional-communication) that generators have a `.close()` method, so it might be worth adding:

funny, I had that in there at first and removed it.  Should have stuck with my gut :)

Another thing that I wanted to mention is that maybe we shouldn’t use this little `yield from` wrapper on csv.reader at all?  Maybe we should just use the `with open()... reader = csv.reader` each time?  Makes it clearer that this is just standard lib

napari/utils/io.py
sofroniewn: We can drop `read_csv` and rename `csv_to_layer_data` as `read_csv_layer_data` might be a bit more keeping with our other names

napari/utils/_tests/test_io.py
sofroniewn: > It just makes the table more annoying to work with.

Many downstream software packages assume the first column of a table is an "index" - see for example this when I load the current approach into applesheets 
image

vs the approach you want

image

personally I find the first option much nicer to work with.

napari/utils/io.py
sofroniewn: I've added the `gen.close`, but it does seem like the with open syntax here would be more appropriate - @tlambert03 could you maybe do that swap?

napari/utils/io.py
sofroniewn: actually keeping `read_csv` for now is nice, but i have renamed `csv_to_layer_data` to `read_csv_layer_data`

napari/utils/io.py
sofroniewn: yeah we could do that in a future PR

napari/utils/io.py
sofroniewn: > This approach is so brittle... integer values are important to preserve. This is one reason why I mentioned we'd probably want to depend on pandas eventually for csvs.

Agreed - we will likely want to pull this code out into its own dedicated plugin for `0.3.1` and maybe depending on pandas then is a good choice ... we can see.

> I would suggest, alternatively, that we treat the csv file as a list of columns, each of which is separately cast to a numpy array — same as properties.

This works for me - one option would be to put all this code inside `read_csv` so it doesn't get duplicated. I don't know @tlambert03 thoughts?

napari/utils/io.py
tlambert03: updated

napari/utils/io.py
sofroniewn: Ok, now after some improvements from @tlambert03 there are two places which does the `float` call, one is on the actual data table - to me that is fine, those are coordinates of the points, and are floats internally if you add then from clicking etc.

The second place is inside the properties dict - i think we can either just leave it for now or is it possible to try and cast to int, detect if that is "safe", if not try to cast to float, and it not ignore?

jni (approved):
We got there! :tada:

PR #1194, opened 2020-05-03
One more plugin error reporting  fix

tlambert03:
# Description
Caught one more lingering bug in the plugin error reporter since the move to `napari-plugin-engine`.  This PR fixes that one-liner [here](https://github.com/napari/napari/compare/master...tlambert03:error-reporter-fix?expand=1#diff-765ae9e85183a03498d33b5be8ff07cbR55), and adds more tests to ensure full coverage on all error reporting stuff.

## Type of change

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/_qt/_tests/test_qt_plugin_report.py
+++ b/napari/_qt/_tests/test_qt_plugin_report.py
@@ -1,12 +1,16 @@
-from napari._qt import qt_plugin_report
-from qtpy.QtCore import Qt
-from qtpy.QtGui import QGuiApplication
-from napari_plugin_engine import PluginError
-import pytest
 import webbrowser
 
+import pytest
+from qtpy.QtCore import Qt
+from qtpy.QtGui import QGuiApplication
 
+from napari._qt import qt_plugin_report
+from napari_plugin_engine import PluginError
+
+
+# qtbot fixture comes from pytest-qt
 # test_plugin_manager fixture is provided by napari_plugin_engine._testsupport
+# monkeypatch fixture is from pytest
 def test_error_reporter(qtbot, test_plugin_manager, monkeypatch):
     """test that QtPluginErrReporter shows any instantiated PluginErrors."""
 
--- a/napari/_qt/_tests/test_qt_plugin_report.py
+++ b/napari/_qt/_tests/test_qt_plugin_report.py
@@ -1,23 +1,53 @@
-from napari._qt.qt_plugin_report import QtPluginErrReporter
+from napari._qt import qt_plugin_report
+from qtpy.QtCore import Qt
+from qtpy.QtGui import QGuiApplication
 from napari_plugin_engine import PluginError
 import pytest
+import webbrowser
 
 
 # test_plugin_manager fixture is provided by napari_plugin_engine._testsupport
-def test_error_reporter(qtbot, test_plugin_manager):
+def test_error_reporter(qtbot, test_plugin_manager, monkeypatch):
     """test that QtPluginErrReporter shows any instantiated PluginErrors."""
+
+    monkeypatch.setattr(
+        qt_plugin_report,
+        'standard_metadata',
+        lambda x: {'url': 'https://github.com/example/example'},
+    )
+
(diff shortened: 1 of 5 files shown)


napari/_qt/_tests/test_qt_plugin_report.py
```diff
@@ -1,23 +1,53 @@
-from napari._qt.qt_plugin_report import QtPluginErrReporter
+from napari._qt import qt_plugin_report
+from qtpy.QtCore import Qt
+from qtpy.QtGui import QGuiApplication
 from napari_plugin_engine import PluginError
 import pytest
+import webbrowser
 
 
 # test_plugin_manager fixture is provided by napari_plugin_engine._testsupport
```
sofroniewn, line 12: Maybe a comment here that the monkeypatch fixture is provided by pytest?

napari/plugins/_tests/test_exceptions.py
```diff
@@ -0,0 +1,28 @@
+from napari_plugin_engine import PluginError
+from napari.plugins import exceptions
+import sys
+import pytest
+
+
+@pytest.mark.parametrize('as_html', (True, False), ids=['as_html', 'as_text'])
```
sofroniewn, line 10: Maybe a comment here that the monkeypatch fixture is provided by pytest?

sofroniewn (approved):
LGTM, left a minor comment

sofroniewn:
Note that added test is currently failing only on linux

tlambert03:
man I love tests :) so useful!  fixed and added comments

napari/_qt/qt_plugin_report.py
```diff
@@ -135,8 +136,12 @@ def set_plugin(self, plugin: str) -> None:
         plugin : str
             name of a plugin that has created an error this session.
         """
+        try:
+            self.github_button.clicked.disconnect()
+        except (RuntimeError, TypeError):
+            pass
```
jni: @tlambert03 can you add a comment about why this is needed? it's very mysterious.

jni: I see also that it's repeated below... Are both actually needed?

tlambert03: sure, disconnecting a signal that has no connected slots is an error... but this seems to be the best way to make sure you have nothing connected.  (It's a runtime error in pyside, and type error on pyqt5)

tlambert03: where is it repeated?

jni: Lines 153-158:

https://github.com/napari/napari/pull/1194/files#diff-9164dcf0ab67f9870f138832a24fdc1dR153-R158

jni: (They are not changed lines, they are already there.)

jni: In the comment, I meant, note down that we are disconnecting any *previous* plugin's GH button click before connecting to this plugin's GH button click.

tlambert03: lol.  sorry I missed that.  and that one has a comment!   will remove

jni: The fact that the disconnect and the reconnect are separated in the function makes it confusing, so if they can happen together, that would be ideal.

tlambert03: I removed the duplicate, but it's hard to put them together.  We always have to disconnect the last callback when we change the plugin we're viewing (so it makes sense to do that at the top) but we only reconnect it if the new plugin has a github link.

jni: @tlambert03 

> I meant, note down that we are disconnecting any previous plugin's GH button click before connecting to this plugin's GH button click.

Could you add a comment right before the disconnection that this is what's happening? Other than that this is :+1: from me!

jni: Actually, scratch that, life's too short. =P

PR #1195, opened 2020-05-03
Normalize paths handed to reader plugins

tlambert03:
# Description
Just found that using `viewer.open_path("~/Desktop/my.tiff")` fails, in this case, because tifffile is not expecting a relative path with an un-expanded user directory.  This PR normalizes paths prior to handing to reader plugins (and also in our `imsave` function) by using the `abspath_or_url` function that was introduced for writer plugins.

Also let `abspath_or_url` accept a sequence of paths, and added more docs and tests for 

## Type of change

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/utils/misc.py
+++ b/napari/utils/misc.py
@@ -7,7 +7,7 @@ import re
 import warnings
 from contextlib import contextmanager
 from enum import Enum, EnumMeta
-from os import fspath, path
+from os import fspath, path, PathLike
 from typing import ContextManager, Optional, Type, TypeVar, Sequence
 
 import dask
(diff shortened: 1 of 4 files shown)


sofroniewn:
Looking good, but windows specific test fail for a `/` vs `\` - see

```python-traceback
_____________________________ test_abspath_or_url _____________________________
    def test_abspath_or_url():
>       assert abspath_or_url('~/something') == expanduser('~/something')
E       AssertionError: assert 'C:\\Users\\C...or\\something' == 'C:\\Users\\C...tor/something'
E         - C:\Users\ContainerAdministrator/something
E         ?                                ^
E         + C:\Users\ContainerAdministrator\something
E         ?                                ^
napari\utils\_tests\test_misc.py:175: AssertionError
```

sofroniewn (approved):
Tests now all passing, and LGTM! thanks!!

napari/utils/misc.py
```diff
@@ -193,11 +193,36 @@ def camel_to_snake(name):
...
+        input).
+    """
+    if isinstance(relpath, (tuple, list)):
+        return type(relpath)(abspath_or_url(p) for p in relpath)
+
+    if isinstance(relpath, (str, PathLike)):
+        relpath = fspath(relpath)
+        if relpath.startswith(('http:', 'https:', 'ftp:', 'file:')):
+            return relpath
+        return path.abspath(path.expanduser(relpath))
```
jni, line 223: I won't hold up the PR for this, but for the record I prefer `os.path.funcname` to `path.funcname`, as I had to think for a second that we weren't using `pathlib` rather than `os.path`.

tlambert03: Agreed.  Will follow up later

PR #1197, opened 2020-05-03
resize axis labels on show

### LINKED ISSUE #1114: view_image with axis_labels doesn't resize labels
## 🐛 Bug
Just noticed that calling `napari.view_image` with `axis_labels` doesn't immediately "resize" the labels to fit long text.  will fix later, just didn't want to forget.

sofroniewn:
# Description
This PR will close #1114 - thought not sure if there is a nicer way to do it @tlambert03? I couldn't get calling the resize to work before the window is shown 

To reproduce I was using
```python
data = np.random.random((20, 100, 200))
viewer = napari.view_image(data, axis_labels=['a very long label', 'y', 'x'])
```

## Type of change

### DIFF
--- a/napari/_qt/qt_main_window.py
+++ b/napari/_qt/qt_main_window.py
@@ -409,6 +409,8 @@ class Window:
         """Resize, show, and bring forward the window."""
         self._qt_window.resize(self._qt_window.layout().sizeHint())
         self._qt_window.show()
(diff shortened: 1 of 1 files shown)


tlambert03 (approved):
works for me!  surprised me at first that it needed to go all the way up to the main window... but you're right!

PR #1200, opened 2020-05-04
Bump svg-dep, add builtin write_labels

tlambert03:
# Description
This bumps the napari-svg dependency to include the "trylast" marker from the last release on images and labels and the multiwriter.  So images will be written as tiff by default, instead of svg.  I added a test to confirm the load order.  We can revisit this in the future when we have persistence.

I also noticed that we didn't have a builtin `write_labels`... so I added one (it's the same as write_image except for dtype), added it to the tests, and  fixed a regression in the "guessing" of labels type with uint32/64

## Type of change

# References

# How has this been tested?

## Final checklist:


sofroniewn (approved):
LGTM! Nice additions

napari/plugins/_builtins.py
```diff
@@ -69,6 +69,30 @@ def napari_write_image(path: str, data: Any, meta: dict) -> Optional[str]:
...
+    meta : dict
+        Image metadata.
+
+    Returns
+    -------
+    path : str or None
+        If data is successfully written, return the ``path`` that was written.
+        Otherwise, if nothing was done, return ``None``.
+    """
+    return napari_write_image(path, np.asarray(data, dtype='uint32'), meta)
```
jni: Hmm, dunno about hardcoding the dtype, vs specifying a minimum of int32. e.g. if labels are uint64, I don't think we want to downcast!

jni: I suggest `dtype = data.dtype if data.dtype.itemsize >= 4 else np.uint32`. (You could be even fancier with testing for negative values, but I'm not as concerned about that.)

tlambert03: I did that after some trial and error with Fiji ... it seemed that even with bioformats, the uint64 wasn’t working (image was entirely 0s).  But I agree it’s not nice to downcast... but slightly worried about people trying to open in Fiji

tlambert03: implemented your suggestion

jni: Ah, interesting. Well, if you want, I'd be ok with checking that the max value fits in uint32. But that will be hard with dask-y things.

jni: anyway, happy with a merge as its stands

tlambert03: Yeah let’s worry about that once someone gets confused with cross compatibility

PR #1201, opened 2020-05-04
change screenshot hotkey and open menubar names

sofroniewn:
# Description
This PR changes the screenshot hotkey to `alt-S` and the screenshot with viewer hot-key to `alt-shift-S`.

It also renames the `Open image` to `Open file` in a couple places as now that we have support for reader plugins we can open more than images

### DIFF
--- a/napari/_qt/qt_main_window.py
+++ b/napari/_qt/qt_main_window.py
@@ -31,6 +31,7 @@ from qtpy.QtWidgets import (  # noqa: E402
     QShortcut,
     QStatusBar,
     QVBoxLayout,
+    QFileDialog,
 )
 from qtpy.QtCore import Qt  # noqa: E402
 from qtpy.QtGui import QKeySequence, QIcon  # noqa: E402
@@ -141,21 +142,19 @@ class Window:
 
     def _add_file_menu(self):
         """Add 'File' menu to app menubar."""
-        open_images = QAction('Open image(s)...', self._qt_window)
+        open_images = QAction('Open file(s)...', self._qt_window)
         open_images.setShortcut('Ctrl+O')
-        open_images.setStatusTip('Open image file(s)')
-        open_images.triggered.connect(self.qt_viewer._open_images)
+        open_images.setStatusTip('Open file(s)')
+        open_images.triggered.connect(self.qt_viewer._open_files)
 
-        open_stack = QAction('Open image series as stack...', self._qt_window)
+        open_stack = QAction('Open files as stack...', self._qt_window)
         open_stack.setShortcut('Ctrl+Alt+O')
-        open_stack.setStatusTip('Open image files')
-        open_stack.triggered.connect(self.qt_viewer._open_images_as_stack)
+        open_stack.setStatusTip('Open files')
+        open_stack.triggered.connect(self.qt_viewer._open_files_as_stack)
 
-        open_folder = QAction('Open Folder...', self._qt_window)
+        open_folder = QAction('Open folder...', self._qt_window)
         open_folder.setShortcut('Ctrl+Shift+O')
-        open_folder.setStatusTip(
-            'Open a folder of image file(s) or a zarr file'
-        )
+        open_folder.setStatusTip('Open a folder')
         open_folder.triggered.connect(self.qt_viewer._open_folder)
 
         save_selected_layers = QAction(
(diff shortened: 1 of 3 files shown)


napari/_qt/qt_main_window.py
```diff
@@ -492,6 +501,24 @@ def _help_changed(self, event):
         """
         self._help.setText(event.text)
 
+    def _save_screenshot(self):
```
tlambert03: i find it a little odd that we now have `screenshot` and `_save_screenshot`, where `_save_screenshot` doesn't actually do the saving, it just generates a dialog window, and `screenshot` does the saving, but only with the `path=` argument...

sofroniewn: This pattern exists in `QtViewer` too - I am happy to change it, and now is a fine time to do so, but then we should do it in both

napari/_qt/qt_main_window.py
```diff
@@ -211,6 +219,7 @@ def handle_exit():
         self.file_menu.addAction(save_selected_layers)
         self.file_menu.addAction(save_all_layers)
         self.file_menu.addAction(screenshot)
+        self.file_menu.addAction(screenshot_wv)
```
jni, line 224: I might expect to find this is the "Window" rather than "File" menu... A very soft suggestion though, I'm not sure.

napari/_qt/qt_main_window.py
```diff
@@ -492,6 +501,24 @@ def _help_changed(self, event):
...
+        filename, _ = QFileDialog.getSaveFileName(
+            parent=self.qt_viewer,
+            caption='Save screenshot with viewer',
+            directory=self.qt_viewer._last_visited_dir,  # home dir by default
+            filter="Image files (*.png *.bmp *.gif *.tif *.tiff)",  # first one used by default
+            # jpg and jpeg not included as they don't support an alpha channel
+        )
+        if (filename != '') and (filename is not None):
+            # double check that an appropriate extension has been added as the filter
+            # filter option does not always add an extension on linux and windows
```
jni:
```suggestion
            # double check that an appropriate extension has been added as the
            # filter option does not always add an extension on linux and windows
```

napari/_qt/qt_main_window.py
jni: Yeah I find it a bit weird also. I even wonder if we should remove it from the public API, and make it a private utility with menubar access. We should start to be careful with our growing public API footprint, lest our viewer become a wasteland like the numpy main namespace.

napari/_qt/qt_main_window.py
```diff
@@ -174,13 +173,22 @@ def _add_file_menu(self):
             lambda: self.qt_viewer._save_layers(selected=False)
         )
 
-        screenshot = QAction('Save Screenshot', self._qt_window)
-        screenshot.setShortcut('Ctrl+Alt+S')
+        screenshot = QAction('Save screenshot...', self._qt_window)
```
jni: We should pick a case convention and stick to it... But every app I have checked uses title case in their menubars...

jni (commented):
@sofroniewn I've made some minor comments. I don't have strong feelings here but I have slight leanings away from both lowercase menubar options and from a public `.screenshot` method overall. Feel free to merge one way or the other after due consideration, though. =)

napari/_qt/qt_main_window.py
sofroniewn: Will go back to uppercase and make consistent

napari/_qt/qt_main_window.py
sofroniewn: done

napari/_qt/qt_main_window.py
sofroniewn: I like having it near the other screenshot in `File` but we can move later if more people find confusing - easy to change later

napari/_qt/qt_main_window.py
sofroniewn: ok - will go with `_save_screenshot` -> `_screenshot_dialog`. I think having `screenshot` as a public api makes sense. Note if `path=None` nothing will get saved and in either case you'll get back the actual screenshot

sofroniewn:
I have made the requested changes, this is now ready for re-review @tlambert03 @jni

napari/_qt/qt_main_window.py
tlambert03: # 🚪🚪

tlambert03 (approved):
👍

PR #1202, opened 2020-05-05
only assert that config returns to original value

tlambert03:
# Description
fix for dask test on release candidate.  Changing test to just assert that the config returns to what it was before we touched it, and not what it's supposed to be to begin with.  Also skipping the two call-counting tests if dask version isn't high enough.  (Note, the cache tests should still pass on earlier versions of dask).


sofroniewn:
Can you also add a line to the release notes too

jni (approved):
LGTM!

Issue #1205, opened 2020-05-05
Dask cache tuning


tlambert03:
## 🐛 Tune dask cache

Chatting with @jni about the dask cache... we need to do a few things to tune performance:

- **think more about what the default size should be.**  If the cache is large, but still not as large as the dataset, then looping a dataset with the play button will just continuously swap the cache, and potentially cause more overhead than if it was off altogether
- **don't resize after created**: If a user creates a Layer with a dask array, the cache is created... if they then resize with `resize_dask_cache(100)`, it works, but if they create a new `Layer`, I think the cache will go back to the default size.  which it should not
- **possibly modify scoring algorithm**: [cachey](https://github.com/dask/cachey) lets you modify the scoring algorithm for cache eviction.  by default, it's `score += compute_time / num_bytes * (1 + eps) ** tick_time`.

jni:
I just want to add that (1) is definitely a problem I'm encountering now. A small or disabled cache makes play much much smoother than the default 1/2RAM cache, for a 15GB dataset (200 x 75MB timepoints).

tlambert03:
and just to be explicit (in case someone stumbles on to this), if you'd like to disable the cache, for now, you need to do it _after_ you add a layer with a dask array, and probably each time:

```python
import napari
from napari.utils import resize_dask_cache

napari.view_image(big_dask_array)
resize_dask_cache(0)
```

sofroniewn:
As noted in #1206 by @tlambert03 

> Let's maybe leave #1205 open for input on tuning the default cache size (which this PR does not change)

tlambert03:
let's maybe close it now :joy:

PR #1206, opened 2020-05-05
rework dask cache

tlambert03:
# Description
This addresses #1205, by making sure that if a user has resized the dask cache manually, then it will not be changed.  And once a cache has been created, it can only be resized using `utils.resize_dask_cache`.  See the tests `test_dask_cache_resizing` and `test_prevent_dask_cache` for examples of what this is fixing.  Let's maybe leave #1205 open for input on tuning the default cache size (which this PR does not change)

Summary:
- `utils.resize_dask_cache(0)`, called _either_ before or after Layer creation will (permanently) disable the cache.
- cache can be resized up or down using `utils.resize_dask_cache(nbytes)` and it will stick.
- if someone actually deletes the `utils.dask_cache` attribute, we will lose the "memory" of the cache, and it will therefore get spun up again on the next Layer addition (but it won't break) ... so using the `resize_dask_cache` API is the way to go.

## Type of change

# How has this been tested?

- [x] added a couple new well-commented tests that describe/test the use case

## Final checklist:

### DIFF
--- a/napari/layers/_tests/test_dask_layers.py
+++ b/napari/layers/_tests/test_dask_layers.py
@@ -48,6 +48,9 @@ def test_dask_array_creates_cache():
     layer2._set_view_slice = mock_set_view_slice2
     layer2.set_view_slice()
 
+    # clean up cache
+    utils.dask_cache = None
+
 
 def test_list_of_dask_arrays_creates_cache():
     """Test that adding a list of dask array also creates a dask cache."""
@@ -121,6 +124,7 @@ def test_dask_unoptimized_slicing(delayed_dask_stack, monkeypatch):
     # make sure we are not caching for this test, which also tests that we
     # can turn off caching
     utils.resize_dask_cache(0)
+    assert utils.dask_cache.cache.available_bytes == 0
 
     # mock the dask_configure function to return a no-op.
     def mock_dask_config(data):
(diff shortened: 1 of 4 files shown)


napari/utils/misc.py
```diff
@@ -292,8 +292,42 @@ def all_subclasses(cls: Type) -> set:
...
+        The fraction (from 0 to 1) of total memory to use for the dask cache.
+        by default, 50% of total memory is used.
+
+    Returns
+    -------
+    [type]
+        [description]
+    """
+    import psutil
+    from napari import utils
```
jni: Relative import?

napari/utils/misc.py
```diff
@@ -332,26 +366,31 @@ def resize_dask_cache(
     >>> cache.cache.total_bytes   # currently used bytes
     """
 
-    from napari.utils import dask_cache
+    from dask.cache import Cache
+    from napari import utils
```
jni: Relative import?

napari/layers/_tests/test_dask_layers.py
```diff
@@ -157,3 +161,66 @@ def dask_optimized_slicing(*args, **kwds):
...
+
+    v = viewer.ViewerModel()
+    dask_stack = delayed_dask_stack['stack']
+    # adding a new stack will not increase the cache size
+    v.add_image(dask_stack, multiscale=False, contrast_limits=(0, 1))
+    assert utils.dask_cache.cache.available_bytes == 0
+    # and the cache will not be populated
+    for i in range(3):
+        v.dims.set_point(0, i)
+    assert len(utils.dask_cache.cache.heap.heap) == 0
```
jni, line 226: 😂 this is quite a line but I’ll take it!

jni (approved):
Very nice @tlambert03 ! I think a couple of imports should be made relative, but otherwise this is good to go!

napari/layers/_tests/test_dask_layers.py
tlambert03: yeah 😄 ... I don't think we'll ever need to access that directly.  but that's where the actual hashes are stored, for a rainy day!

tlambert03:
cleaned up imports.  can be merged after tests pass

sofroniewn (approved):
LGTM!

PR #1208, opened 2020-05-06
Use napari logo as taskbar icon on windows when run as python script

### LINKED ISSUE #1168: Napari taskbar icon on Windows is generic python logo
## 🐛 Bug

On Windows, the windows icon is different from the taskbar icon, so Napari shows up as a generic Python application:

image

## Expected behavior

It would be nice to have the Napari logo show in the taskbar, and I assume this is a Windows specific issue. From this [Stack Overflow answer](https://stackoverflow.com/a/1552105/2337392), adding this to `Viewer()` class does this trick:

```python
import ctypes
myappid = 'mycompany.myproduct.subproduct.version' # arbitrary string
ctypes.windll.shell32.SetCurrentProcessExplicitAppUserModelID(myappid)
```
image

`myappid` seems to not really matter, and we could just set that to `"napari"`
## Environment

```
napari: 0.2.12+71.gb2a95d6.dirty
Python: 3.8.2 (tags/v3.8.2:7b3ab59, Feb 25 2020, 23:03:10) [MSC v.1916 64 bit (AMD64)]
Qt: 5.14.1
PySide2: 5.14.1
NumPy: 1.18.1
SciPy: 1.4.1
Dask: 2.12.0
VisPy: 0.6.4

GL version: 4.6.0 NVIDIA 441.45
MAX_TEXTURE_SIZE: 32768

Plugins: None
```

Czaki:
# Description
On windows use napari logo as taskbar icon when running from script. Fixes  #1168 

## Type of change

- [x] Bug-fix (non-breaking change which fixes an issue) #1168

# References
https://stackoverflow.com/questions/1551605/how-to-set-applications-taskbar-icon-in-windows-7/1552105#1552105

# How has this been tested?
Run on windows 10 and icon is changed from default python logo

## Final checklist:

### DIFF
--- a/napari/_qt/event_loop.py
+++ b/napari/_qt/event_loop.py
@@ -27,8 +27,13 @@ def gui_qt(*, startup_logo=False):
     """
     if platform.system() == "Windows" and not getattr(sys, 'frozen', False):
         import ctypes
-        napari_app_id = 'napari.napari.viewer.' + str(__version__)  # arbitrary string
-        ctypes.windll.shell32.SetCurrentProcessExplicitAppUserModelID(napari_app_id)
+
+        napari_app_id = 'napari.napari.viewer.' + str(
+            __version__
+        )  # arbitrary string
+        ctypes.windll.shell32.SetCurrentProcessExplicitAppUserModelID(
+            napari_app_id
+        )
 
     splash_widget = None
     app = QApplication.instance()
--- a/napari/_qt/event_loop.py
+++ b/napari/_qt/event_loop.py
@@ -1,12 +1,9 @@
-import platform
 import sys
 from contextlib import contextmanager
 from os.path import dirname, join
 
 from qtpy.QtGui import QPixmap
 from qtpy.QtWidgets import QApplication, QSplashScreen
-
-from napari import __version__
 
 
 @contextmanager
@@ -25,16 +22,6 @@ def gui_qt(*, startup_logo=False):
     IPython with the Qt GUI event loop enabled by default by using
     ``ipython --gui=qt``.
     """
-    if platform.system() == "Windows" and not getattr(sys, 'frozen', False):
-        import ctypes
-
-        napari_app_id = 'napari.napari.viewer.' + str(
-            __version__
-        )  # arbitrary string
-        ctypes.windll.shell32.SetCurrentProcessExplicitAppUserModelID(
-            napari_app_id
-        )
-
     splash_widget = None
     app = QApplication.instance()
     if not app:
--- a/napari/_qt/event_loop.py
+++ b/napari/_qt/event_loop.py
@@ -1,9 +1,12 @@
+import platform
 import sys
 from contextlib import contextmanager
 from os.path import dirname, join
 
 from qtpy.QtGui import QPixmap
 from qtpy.QtWidgets import QApplication, QSplashScreen
+
+from napari import __version__
 
 
 @contextmanager
@@ -22,6 +25,11 @@ def gui_qt(*, startup_logo=False):
     IPython with the Qt GUI event loop enabled by default by using
     ``ipython --gui=qt``.
     """
+    if platform.system() == "Windows" and not getattr(sys, 'frozen', False):
+        import ctypes
+        napari_app_id = 'napari.napari.viewer.' + str(__version__)  # arbitrary string
+        ctypes.windll.shell32.SetCurrentProcessExplicitAppUserModelID(napari_app_id)
+
     splash_widget = None
     app = QApplication.instance()
     if not app:
--- a/napari/viewer.py
+++ b/napari/viewer.py
@@ -29,10 +29,6 @@ class Viewer(ViewerModel):
         Dimension names. by default they are labeled with sequential numbers
     show : bool, optional
         Whether to show the viewer after instantiation. by default True.
-    icon_bugfix: bool
-        If use bugfix to show proper icon when running as script on Windows.
-        May want to disable when embed napari as viewer in own application
-
     """
 
     napari_app_id = 'napari.napari.viewer.' + str(__version__)
--- a/napari/viewer.py
+++ b/napari/viewer.py
@@ -1,3 +1,5 @@
+import platform
+import sys
 from os.path import dirname, join
 
 from qtpy.QtGui import QIcon
@@ -7,6 +9,7 @@ from ._qt.qt_update_ui import QtUpdateUI
 from ._qt.qt_main_window import Window
 from ._qt.qt_viewer import QtViewer
 from .components import ViewerModel
+from . import __version__
 
 
 class Viewer(ViewerModel):
@@ -26,6 +29,10 @@ class Viewer(ViewerModel):
         Dimension names. by default they are labeled with sequential numbers
     show : bool, optional
         Whether to show the viewer after instantiation. by default True.
+    icon_bugfix: bool
+        If use bugfix to show proper icon when running as script on Windows.
+        May want to disable when embed napari as viewer in own application
+
     """
 
     def __init__(
@@ -35,6 +42,7 @@ class Viewer(ViewerModel):
         order=None,
         axis_labels=None,
         show=True,
+        icon_bugfix=True,
     ):
         # instance() returns the singleton instance if it exists, or None
         app = QApplication.instance()
@@ -66,6 +74,19 @@ class Viewer(ViewerModel):
         )
         qt_viewer = QtViewer(self)
         self.window = Window(qt_viewer, show=show)
+        if (
+            icon_bugfix
+            and platform.system() == "Windows"
+            and not getattr(sys, 'frozen', False)
+        ):
+            import ctypes
+
+            napari_app_id = 'napari.napari.viewer.' + str(
+                __version__
+            )  # arbitrary string
+            ctypes.windll.shell32.SetCurrentProcessExplicitAppUserModelID(
+                napari_app_id
+            )
 
     def update_console(self, variables):
         """Update console's namespace with desired variables.
--- a/napari/viewer.py
+++ b/napari/viewer.py
@@ -35,6 +35,8 @@ class Viewer(ViewerModel):
 
     """
 
+    napari_app_id = 'napari.napari.viewer.' + str(__version__)
+
     def __init__(
         self,
         title='napari',
@@ -42,7 +44,6 @@ class Viewer(ViewerModel):
         order=None,
         axis_labels=None,
         show=True,
-        icon_bugfix=True,
     ):
         # instance() returns the singleton instance if it exists, or None
         app = QApplication.instance()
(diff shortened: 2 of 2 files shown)


napari/_qt/event_loop.py
```diff
@@ -1,10 +1,13 @@
+import platform
 import sys
 from contextlib import contextmanager
 from os.path import dirname, join
 
 from qtpy.QtGui import QPixmap
 from qtpy.QtWidgets import QApplication, QSplashScreen
 
+from napari import __version__
```
tlambert03: relative import?

napari/_qt/event_loop.py
```diff
@@ -22,6 +25,16 @@ def gui_qt(*, startup_logo=False):
     IPython with the Qt GUI event loop enabled by default by using
     ``ipython --gui=qt``.
     """
+    if platform.system() == "Windows" and not getattr(sys, 'frozen', False):
```
tlambert03: I'm not sure we plan on using pyinstaller anymore... is `frozen` used for anything else?

tlambert03 (commented):
@hectormz, can you try this?

@Czaki, by putting this in `gui_qt` it will not work if someone has started the qt loop outside of napari (such as by using `ipython --gui qt` or elsewhere) right?  Was that intentional?

napari/_qt/event_loop.py
Czaki: There is #496 work to create bundle app. I try it also in #1082 but fail on some openGL problem. I'm not on time with napari talks, so maybe you change plan. 

In my case I need to create bundle, to avoid setup python by some of users.

Czaki:
> @Czaki, by putting this in `gui_qt` it will not work if someone has started the qt loop outside of napari (such as by using `ipython --gui qt` or elsewhere) right? Was that intentional?

My intention is to not conflict with persons who would like to embed napari in their application. From issues which I read You suggest person to dock their widget instead of dock `Viewer` as widget in their window. So overriding this value on every `Viewer` start may break some usages.

napari/_qt/event_loop.py
tlambert03: yep I'm aware of #496 (look at the commit history there).  and yes, we're considering alternatives that will make it easier to extend the environment after bundling.  lots of discussions going on here https://github.com/napari/napari/issues/1074

if you are creating a bundle, I assume it is of your own application that wraps napari right?  In which case wouldn't you be using your own icon?

Czaki: When I create bundle there is no problem with icon, because program is not interpreted as python one. Problem is in this line:
https://github.com/napari/napari/blob/9ed0fda58d2dca7f14e67107a0fd5b3791116555/napari/_qt/event_loop.py#L34-L36

This may cause that two apps which using napari viewer will be treated as one because they have same SetCurrentProcessExplicitAppUserModelID

hectormz:
This works for me when launching from command line!

But does not work when launching from `ipython`, if that's the intent.

(not work=python logo instead of napari)

Czaki:
Maybe this version?

hectormz:
@Czaki Now launching from the commandline nor from ipython show the napari logo.

Czaki:
Ok. This ctypes change need to be done before windows creation. 
Last commit should work (I again login on windows machine for test)

hectormz:
Awesome @Czaki! Icon shows up when launching from commandline and from ipython now!

napari/viewer.py
```diff
@@ -26,6 +29,10 @@ class Viewer(ViewerModel):
         Dimension names. by default they are labeled with sequential numbers
     show : bool, optional
         Whether to show the viewer after instantiation. by default True.
+    icon_bugfix: bool
+        If use bugfix to show proper icon when running as script on Windows.
+        May want to disable when embed napari as viewer in own application
```
tlambert03: @Czaki, I know you're thinking ahead to embedding napari in your own application, but `napari.Viewer` is pretty much our _main_ offering, and I think we want to be very protective about adding parameters.  I'll defer to @jni and @sofroniewn, but I don't think this parameter is going to make the cut.

Czaki: But forcing this cptypes overwrite all time it may break behaviour of napari based (using docking system) applications on windows. 

In my case I use `ViewerModel` so there is no problem for me. 

You sell napari as viewer, not complex stand-alone application.

tlambert03: I'm not saying you have to force it, but I am politely requesting that you consider alternatives that don't require adding a parameter to our main product.  perhaps a module level variable?  etc...

I'm really hoping this is a lost-in-translation thing... but I have to tell you that it comes off as extremely rude for you to tell us how we "sell our application" as justification for rejecting a request.  We _greatly_ value your input, but please also respect that we have lots of different users and use cases, and we like to consider all of them before simply accepting a proposed change.  There are a _lot_ of people who will never embed napari, but may look at our Viewer parameters and say "What is this `icon_bugfix` thing?  do I need to use it?... What's the bug?"  In general, please try to be open to these discussions.

tlambert03: (I'm sorry if I overreacted or was too harsh there, I'd just really like us to consider all alternatives and would like you to help us in that process)

jni: @Czaki thanks for your input, both here and in the other PRs where you have contributed.

I think @tlambert03 overreacted a bit because the scope of napari might not be immediately clear from the README or existing uses. napari is many things and one of them is a Python viewer, indeed — and from that angle, we actually want to be *more* protective of the API signature, not less.

In the future, we do hope to release a napari standalone application on which other applications can be built as plugins. So napari will *also* be an application. You can see our public short-term roadmap [here](https://napari.org/docs/developers/ROADMAP_0_3.html).

At any rate, this is not to dismiss your concern — just to say that this is not the solution we would like to address that concern. Something like an environment variable, eg NAPARI_DISABLE_WIN_ICON, that we can check, would be much less obtrusive but still accessible for application developers. 

Again, thanks for working with us to come up with a solution that works for everyone! What do you think of the environment variable approach?

Czaki: I do not think that enviroment variable is good solution (it is global and rather for user than for programmer)

> I'm really hoping this is a lost-in-translation thing... but I have to tell you that it comes off as extremely rude for you to tell us how we "sell our application" as justification for rejecting a request.

I base on current description of napari. Maybe it is outdated 

> Napari viewer is a fast, interactive, multi-dimensional image viewer for Python. It’s designed for browsing, annotating, and analyzing large multi-dimensional images. It’s built on top of Qt (for the GUI), vispy (for performant GPU-based rendering), and the scientific Python stack (numpy, scipy).

But for me some part of code which will group different napari based solution in one group on windows which need code edition to disable is not a good idea. In first try I put it in qt loop initialisation to add programmer possibility to redefine _identifier_. But this solution does not work with napari launched from `ipython --gui qt`. 

I do not see other option than put it in Viewer constructor because it is always executed before create napari window, but for me this need to be easy to disable. Way which i see are:

 * Constructor argument - easiest to found, but may also some basic user may be not sure what it is. 
 * Environment variable - strongly against. As mentioned before it is rather for end user  than for programmer
 * Module level variable - in which module?
 * Class level variable  - not visible at firs moment but easy to control (overwrite before object creation, sublclass with new value) 
 * Check if current object instance of subclass and then not apply bugfix?  

One thing which can be done is to check if control variable is string and if is then use it value instead of predefined one.

Czaki: @tlambert03 @jni

jni: @Czaki 

> Environment variable - strongly against. As mentioned before it is rather for end user than for programmer

Well, note that a programmer can execute:

```python
import os

os.environ['NAPARI_DISABLE_WIN_ICON'] = '1'
```

before importing napari. So I don't see environment variables as a big barrier to generalisation here.

> Module level variable - in which module?

Could do. `napari._qt.WIN_ICON = True` by default, but app developers can set it to False if needed. That works for me. @sofroniewn?

The key thing for me is that this is a private, hidden from view thing, not something for our public API. It's an ugly hack and we don't expose ugly hacks — but maybe we allow ugly hacks to patch our ugly hacks. =P

sofroniewn: > But forcing this cptypes overwrite all time it may break behaviour of napari based (using docking system) applications on windows.
>
> In my case I use ViewerModel so there is no problem for me.

Sounds like disabling this functionality is not currently needed and this is more about potential bugs rather than current usage so I think we should remove the kwarg and just have that first if statement always run for now.

Can you make that change @Czaki?

We can always add in the ability to turn off the functionality in the future if many people are requesting it or it is clearly blocking a major usage mode. 

Stepping back I think that apps that are looking to consume the napari viewer, embed it directly in their own Qt application, and personalize the icon, should be importing the `QtViewer` directly - see #1122 for proposals to make that easier, rather than importing the `Viewer` directly, and so I don't see this functionality needed for that use case.

Hope this is an ok resolution.

Czaki: > Well, note that a programmer can execute:
> 
> ```python
> import os
> 
> os.environ['NAPARI_DISABLE_WIN_ICON'] = '1'
> ```
> 
> before importing napari. So I don't see environment variables as a big barrier to generalisation here.

This is not pythonic way. In exception to really special cases there should be no code execution before import ends. 


> Stepping back I think that apps that are looking to consume the napari viewer, embed it directly in their own Qt application, and personalize the icon, should be importing the `QtViewer` directly - see #1122 for proposals to make that easier, rather than importing the `Viewer` directly, and so I don't see this functionality needed for that use case.

I decide to embed viewer but You suggest to use docking instead of `QtViewer` and `ViewerModel` in same isue which you pointed  https://github.com/napari/napari/pull/1122#issuecomment-611595540.

Czaki: Maybe this? removed kwargs, but allow control by user

sofroniewn: Can you also drop the `icon_bugfix` doc string. I'll let @jni then review

Czaki:
Test fails because of scikit-image update. See #1249

Czaki:
rebased after skimge fix

napari/viewer.py
```diff
@@ -28,6 +31,8 @@ class Viewer(ViewerModel):
         Whether to show the viewer after instantiation. by default True.
     """
 
+    napari_app_id = 'napari.napari.viewer.' + str(__version__)
```
jni: Final change request: let's make this attribute semi-private (`_napari_app_id = `) so that people don't see it when tab-completing with `viewer.<TAB>`. Perhaps there should also be a comment above the attribute: `# set napari.Viewer._napari_app_id to None to avoid overwriting dock icon on Windows`? In fact, is it None or the empty string? I don't know the right way to deal with this, which somewhat defeats the purpose.

Czaki: Done. Also add information, that if you use custom string then it prevent grouping (tested on Windows). 
I add check that prevent setting ctypes if value cast to false.

jni (approved):
Great! Thanks for bearing with us. This is good to go as far as I'm concerned.

sofroniewn (approved):
Looks good to me too, I will merge later today

hectormz:
Thanks all for working to resolve my issue!

Issue #1110, opened 2020-04-03
Are some operations impossible via the Viewer.update() function?


AndrewGYork:
This code works how I expect: the screen updates with random "snow".
```
import numpy as np
import napari
import time

with napari.gui_qt():
    viewer = napari.Viewer()
    layer = viewer.add_image(np.random.random((512, 512)))

    def update(update_period):
        while True:
            time.sleep(update_period)
            layer.data = np.random.random((512, 512))

    viewer.update(update, update_period=0.05)
```

...but this code does not; it triggers several warnings like this:
`WARNING: QObject::setParent: Cannot set parent, new parent is in a different thread`
and also seems to break and undock the layer controls panel.

```
import numpy as np
import napari
import time

with napari.gui_qt():
    viewer = napari.Viewer()

    def update(update_period):
        layer = viewer.add_image(np.random.random((512, 512)))
        while True:
            time.sleep(update_period)
            layer.data = np.random.random((512, 512))

    viewer.update(update, update_period=0.05)
```

I guess Viewer.update() creates a QRunnable, which I guess I should think of as a thread?

https://github.com/napari/napari/blob/63b0a939b88e6b6ffaa76a026c69d2168afa87e2/napari/viewer.py#L111-L114

https://github.com/napari/napari/blob/63b0a939b88e6b6ffaa76a026c69d2168afa87e2/napari/_qt/qt_update_ui.py#L4

It seems to me that it's not possible to create a layer in any thread except the main thread, which suggests that I can't create layers interactively via the `update` function. Is that accurate? If so, I assume this applies to other napari operations, and maybe we should focus on a different approach (for example, scheduling interaction via a QTimer)

AhmetCanSolak:
Yes it is basically a new thread @AndrewGYork . These are run and forget type of threads to handle some side computation/data manipulation.

The way to interact with parent thread's objects would be using signals and slots from another thread. Creating a slot in parent thread, passing connected signal to closure can do what you want to achieve here. Maybe wrapping some `viewer` methods as slots can help, or do you think that will not be enough for your use?

tlambert03:
[...]

PR #1210, opened 2020-05-06
General multithreading API and @thread_worker decorator

tlambert03:
# Description
This is a proposal for a general purpose mechanism to put long-running functions written in simple python syntax into a separate thread (like `viewer.update`), while also maintaining two-way communication between the threads. The user provides a regular function or a generator function (at least one yield in the function), and this will abstract away all the signals/slots hookup, providing a standard set of signals (`yielded`, `returned`, `errored`, `finished`, etc...) and slots (`send`, `abort`, etc...) that can be used back in the main thread.  Might be useful for people like @AndrewGYork (see discussions starting around https://github.com/napari/napari/issues/1110#issuecomment-609063870), and @jni, who came up against issues trying to train a model in the same thread as napari.

**EDIT** the API below has since changed... the basic idea is similar... For an updated overview on how this is intended to be used, see the docs in #1258 

**at its simplest (no communication... but closures can be used):**

```python
@qthreaded
def simple_function():
    """Just a long running function, most like viewer.update."""
    time.sleep(5)  # long running, blocking function
    return 'value'

worker, thread = simple_function()  # it's running in another thread!
worker.returned.connect(lambda v: print(v))  # do something with result
```

**at its most complicated (two-way communication):**

```python
@qthreaded
def two_way_communication_with_args(begin, end):
    """Both sends and receives values to & from the main thread."""

    # optionally, do setup here... this happens in the main thread
    # if you yield a dict, keys will be added as attributes to the worker object
    yield {'begin_val': begin, 'end_val': end, 'some_other_val': 'hi'}

    # do computationally intensive work here that reports back to main thread
    i = begin
    while i <= end:
        time.sleep(0.5) # (blocking)
       
        # `yield i` sends values back to the main thread
        # while `incoming = yield` receives values from the main thread
        incoming = yield i 
        i = incoming if incoming is not None else i + 1

    # do optional teardown here
    return "done"

worker, thread = two_way_communication_with_args(0, 8, start_thread=False)  # not running yet
print(worker.end_val)  # 8
worker.yielded.connect(...)  # do something with yielded values
worker.returned.connect(...)  # do something with return values
worker.errored.connect(...)  # do something with exceptions
thread.start() # now it's running!

worker.send(4)  # put the count back to 4
worker.abort()  # eh, forget about it.
```

See complete examples in `examples/multithreading_work.py`

still some cleanup needed to make sure threads are cleaned if we quit the program.  can do if people like this idea.

## Type of change

# References

# How has this been tested?

## Final checklist:


tlambert03:
here's a [gist](https://gist.github.com/tlambert03/836cf83e1679fec649e86f9138b55f0b) that demonstrates usage of this `@qthreaded` decorator, along with [noise2self](https://github.com/czbiohub/noise2self), to update the viewer with model predictions as the model trains in another thread:

Untitled


the relevant part for this PR boils down to:

```python
@qthreaded
def train():
    """train a model"""
    for i, batch in enumerate(data_loader):
        if i % 10 == 0:
            ...
            yield model  # or whatever you want


worker, thread = train()
worker.yielded.connect(handle_model_update)
```

jni:
@tlambert03 this is so awesome! One Q:

> `worker.yielded.connect(handle_model_update)`

Does `handle_model_update` in this case mean refresh the layer?

tlambert03:
Yep, (or anything else you want to do). It’s any callback, and it receives the yielded results

jni:
so... is this ready to come out of draft? =)

tlambert03:
could be, do you want it in soon?  would probably be fine to sneak it in as an "experimental" feature.  I was going to try to add a processes backend as well, and probably rename to `@thread_worker` and `@process_worker` ... (leave the q out of it)

jni:
> do you want it in soon?

Yeh. =D I am super excited about this.

jni:
I'd leave processes out for now (can be added later), and yes, I would also make the name UI-agnostic.

tlambert03:
ok @jni!    I did lots of stuff today and this is ready for review.

In addition to looking through `_qt.threading` itself, I definitely recommend running both `examples/multithreading_simple.py` ... which is the bare minimum example of throwing a function into another thread, and `examples/multithreading_two_way.py` which is a complete example of controlling (starting, pausing & aborting) a function in another thread as shown below:

Untitled


Since this is a super-set of the `Viewer.update` functionality, I reimplemented `viewer.update` using this API and deprecated that function (my personal feeling is that the convenience function for starting another thread doesn't really belong on the viewer... but curious what @napari/core-devs think... Here it's a decorator).  I also updated `examples/interactive_scripting.py` by adding just a couple yields to the function... and if you compare it to the one on master, you'll notice that if you quit the viewer mid-stream, this PR lets you quit _immediately_ ... This is also due to the new global threadpool API.  Any worker started with `_qt.threading.start_worker` function will be tracked and cleaned up on exit.  And provided they are controlling a function that at least _sometimes_ yields, they can exit elegantly and immediately.

note: the gist from the example above is now broken because I changed some of the variable names.  Can update if you'd like.

examples/multithreading_simple.py
```diff
@@ -0,0 +1,45 @@
...
+    # Note that signals/slots are best connected *before* starting the worker.
+    worker.started.connect(lambda: wdg.status.setText("worker is running..."))
+    worker.returned.connect(
+        lambda x: wdg.status.setText(f"worker returned {x}")
+    )
+    worker.start()
+
+    # # The above syntax is equivalent to:
+    # worker = long_running_function(
+    #     connections={
```
jni: Hmm, could we use prefixed kwargs here to avoid potential name clashes? `_connections={`? Alternatively, `_worker_kwargs={'connections': {...`

napari/_qt/threading.py
```diff
@@ -0,0 +1,563 @@
+import inspect
+from functools import wraps
+from typing import Type, Dict, Callable, Any, Set, Optional
+from qtpy.QtCore import QObject, QThread, Signal, Slot, QRunnable, QThreadPool
+import time
+
+
+def as_generatorfunction(func: Callable) -> Callable:
```
jni:
```suggestion
def as_generator_function(func: Callable) -> Callable:
```

Surely...? =)

napari/_qt/threading.py
```diff
@@ -0,0 +1,563 @@
...
+        self._incoming_value = None
+        self._pause_requested = False
+        self._resume_requested = False
+        self._paused = False
+        self._pause_interval = 0.01
+        self._running = False
+
+        first_yield = next(self._gen)
+        if isinstance(first_yield, dict):
+            self.parse_first_yield(first_yield)
```
jni: This feels hacky... I don't like that yields can have different types. Happy to keep it if there's no other way, but have you thought of any alternatives? I don't have any myself...

jni: At any rate, there should be some kind of validation that this dictionary contains stuff we care about. I can definitely see people yielding json dicts at each step and us mishandling that...!

napari/_qt/threading.py
```diff
@@ -0,0 +1,563 @@
...
+
+    def parse_first_yield(self, first_yield):
+        self.__dict__.update(**first_yield)
+
+    def work(self) -> None:
+        """Core loop that calls the original function.  Enters a continual
+        loop, yielding and returning from the original function.  Checks for
+        various events (quit, pause, resume, etc...)
+        """
```
jni, line 275: Is it just me, or did you just implement an event loop from scratch? And if so, does that point to something like, maybe we should try to hook directly into Qt's main event loop rather than implement our own?

I have a very rudimentary understanding of event loops, so the above are genuine questions, not leading ones!

jni: semi-related question: does this implementation mean that we are limited to 100 yields per second in the worker function? Also, what if the worker function actually takes much longer than this? Is it fine because this loop will run in a new thread so blocking workers won't block the main UI?

jni: As always, I'd like to request that the answers to my questions are summarised in the function docstring, as they will surely be handy in the future!

napari/_qt/threading.py
```diff
@@ -0,0 +1,563 @@
...
+            if not isinstance(self._length, int) and self._length > 0:
+                raise ValueError(
+                    "If providing __len__, it must be a positive int"
+                )
+        super().parse_first_yield(result)
+
+    def pre_yield_hook(self):
+        self._counter += 1
+        if self._length:
+            self.progress.emit(round(100 * (self._counter - 1) / self._length))
```
jni: Could we instead make an object that is compatible with tqdm/progressbar? Namely, make the worker itself iterable and have a `__len__` attribute. We could then write stuff around tqdm and take advantage of its many features, including ETA. This issue has lots of useful links around it:

https://github.com/tqdm/tqdm/issues/482

napari/_qt/threading.py
```diff
@@ -0,0 +1,563 @@
...
+
+
+def worker_factory(
+    *args,
+    start_thread: bool = True,  # TODO: Get opinions on best default value
+    connections: Dict[str, Callable] = None,
+    worker_class: Type[WorkerBase] = Worker,
+    **kwargs,
+) -> WorkerBase:
+    """Convenience function to start a function in a .
```
jni: in a...? ...? ...???? :joy:

examples/multithreading_simple.py
```diff
@@ -0,0 +1,45 @@
...
+    widget.show()
+    return widget
+
+
+if __name__ == "__main__":
+    app = QApplication([])
+    wdg = create_widget()
+
+    # call decorated function
+    worker = long_running_function(start_thread=False)
```
jni: Same here — use prefixed kwargs?

napari/_qt/threading.py
```diff
@@ -0,0 +1,563 @@
...
+    """Return the number of active threads in the global ThreadPool."""
+    return QThreadPool.globalInstance().activeThreadCount()
+
+
+#############################################################################
+
+# convenience functions for creating Worker instances
+
+
+def worker_factory(
```
jni: I don't really like the name `factory` as a public API. It's a very programmery term. Are users meant to use this function anyway? If not, maybe prefix with `_`? If so, then I would rename it to something more descriptive, such as `create_worker_for_func(func, *args, _start_thread=True, _connections=None, _worker_class=Worker, **kwargs)`

jni: Also, shouldn't this function add the created worker to the workers set?

napari/_qt/threading.py
```diff
@@ -0,0 +1,563 @@
...
+            # do teardown
+            return 'anything'
+
+        # call the function to start running in another thread.
+        worker = long_function(start_thread=False)
+        # connect signals if desired
+        worker.start()
+    """
+
+    def inner_func(func: Callable) -> Callable:
```
jni: `inner_func` is certainly not descriptive enough... In some sense, this is the decorator, right? Also, why is `function` optional? It feels like one level of nesting too many...

napari/_qt/threading.py
```diff
@@ -0,0 +1,563 @@
...
+    the work. They can be moved to a QThread as is done here.
+
+    .. note:: Mostly ignorable detail
+
+        While the signals/slots syntax of the worker looks very similar to
+        standard "single-threaded" signals & slots, note that inter-thread
+        signals and slots (automatically) use an event-based QueuedConnection,
+        while intra-thread signals use a DirectConnection. See `Signals and
+        Slots Across Threads
+        <https://doc.qt.io/qt-5/threads-qobject.html#signals-and-slots-across-threads>`_
```
jni, line 692: I love this level of detail in documentation, thank you! :blush:

napari/_qt/threading.py
```diff
@@ -0,0 +1,563 @@
...
+############################################################################
+
+# This is a variant on the above pattern, it uses QThread instead of Qrunnable
+# see https://doc.qt.io/qt-5/threads-technologies.html#comparison-of-solutions
+# It provides more flexibility, but requires that the user manage the threads.
+# With the runnable pattern above, we leverage the QThreadPool.globalInstance()
+# which is wrapped in our own convenience method API.
+
+
+def new_worker_qthread(
```
jni: Could play be refactored to use the simpler convenience decorators? 

In other words, when you say this provides more flexibility, what are you actually talking about? Can you give an example in that comment about something you can do with this that you can't do with the decorators?

napari/viewer.py
```diff
@@ -58,6 +58,8 @@ def __init__(
         logopath = join(dirname(__file__), 'resources', 'logo.png')
         app.setWindowIcon(QIcon(logopath))
 
+        app.aboutToQuit.connect(wait_for_workers_to_quit)
```
jni, line 81: Does wait actually wait, or does it kill them after the timeout? I wonder if there's a better name for it, say, `shut_down_workers_by_hook_or_by_crook` or similar. :stuck_out_tongue:

jni (commented):
@tlambert03 very nice PR, thank you! :tada: I've left some comments about the API, but yeah, more or less it's ready, just things like which bits are public and what to name things left to discuss! Looking forward to a rundown with the team tomorrow. =)

jni:
@AhmetCanSolak might be interested to look at this in detail before the meeting also!

examples/multithreading_two_way.py
```diff
@@ -0,0 +1,111 @@
+import time
```
kevinyamauchi, line 1: I love these examples. Thanks for including them, @tlambert03 !

examples/multithreading_simple.py
tlambert03: agreed.  clashes bad.  see other comment

examples/multithreading_simple.py
tlambert03: yeah, that's a definite issue.  I like the underscore prefix just fine.  The other (not mutually exclusive) option is to accept non-prefixed optional kwargs on the decorator itself:

```python
@thread_worker(start_thread=False, connect={'errored': handle_error})
def myfunc(arg):
    ...

worker = myfunc('arg')  # will not start immediately
worker.returned.connect(my_callback)  # can still connect here too
worker.start()
```

I also wanted to discuss whether you think the default behavior should be to start the thread when the function is called, or to force the user to call `worker.start()` unless they
used:
```python
@thread_worker(start_thread=True)
def myfunc(arg):
    ...

# or 
worker = myfunc(_start_thread=True)
```
(the main issue here is that all connections should happen _before_ the thread is started.  If there's nothing at all to connect, then it's sort of moot.

napari/_qt/threading.py
tlambert03: hah.  that was sort based on `inspect.isgeneratorfunction` ... but it looks like I split the difference 😂.

napari/_qt/threading.py
tlambert03: yeah, it's a bit hacky... a couple general thoughts:
This part of the API is definitely the most open to user-error.  The goal is to create something like an `__init__` section that operates in the main thread and allows some setup.  It's not so much whether _we_ care about what's in there (with the exception of the `ProgressWorker` example ... it's more for the consumer of the decorator. So you could do something like this:

```python
@thread_worker(start_thread=False)
def my_func(a, b):
    yield {'product', b * a}
    for i in range(10:
        ....
        yield

worker = my_func(8, 8)
if worker.product > 60:
    worker.start()
else:
    ....
```

two problems (and this would be clearly documented in the docs I'm writing... but still):
1. if the user _doesn't_ yield once in the beginning of a generator function, but just goes straight into computation... then the first yield would happen in the main thread and choke the viewer.  The only way around that I can see is to remove this option for setup all together, and _not_ step once into the generator in the main thread.

2. If you do like the setup option and we keep that first yield... I'll note that if they yield anything other than a dict in that first yield, it is simply ignored... and if they do return a dict but never intended to access it on the `worker` object that's fine too.  Also, it's _only_ the first yield that we inspect... so if they are yielding json dicts at each step in the main body, those will just be emitted in the `worker.yielded` signal (for them to handle in the main thread)

napari/_qt/threading.py
tlambert03: > did you just implement an event loop from scratch?

yes, that's an event loop!  (Actually, in looking into this, I later realized that this pattern is extremely similar to [generator coroutines](https://docs.python.org/3/library/asyncio-task.html#generator-based-coroutines), used before we had the `async/await` keywords.). Anyway, to answer your question... there _is no_ "main Qt event loop" running in the worker thread where this `work()` method is being executed.  So there's nothing to hook into.  I guess it's conceivable that we could spin up a new [`QEventLoop`](https://doc.qt.io/qt-5/qeventloop.html) in that other thread (but I'm not sure) ... however, i'm not sure what the benefit would be over this syntax.  This is just a pure python event loop that handles all the events we need/declare.

> does this implementation mean that we are limited to 100 yields per second in the worker function?

No.  That `self._pause_interval = 0.01` only pertains to when the user has explicitly "paused" execution of the thread. In which case the thread will poll 100 times per second for `if self._resume_requested`.

> Also, what if the worker function actually takes much longer than this? Is it fine because this loop will run in a new thread so blocking workers won't block the main UI?

yeah, this 👆 . it's in another thread... it can take as long as it wants without blocking main UI.

> summarised in the function docstring

will do!

napari/_qt/threading.py
tlambert03: funny... I actually had a `__len__` method on there and then removed it.  Not sure why.  I like the idea!

... but as I'm writing this... I'm realizing: doesn't the object also need to implement the full iterator protocol to be useful for tqdm?  That _might_ be trickier... (but not impossible)

napari/_qt/threading.py
tlambert03: > I don't really like the name factory as a public API

also funny: I had it as `create_worker` for a while and then changed it.  Will go back to `create_worker` ... (do you definitely prefer the `_for_func` prefix?)

> shouldn't this function add the created worker to the workers set?

It does, because it either calls `start_worker(worker)` directly (and _that_ function adds it to the worker set), or puts that function as the `.start()` method on the worker object.

napari/_qt/threading.py
tlambert03: lol.  you'll have to tune in next week.

napari/_qt/threading.py
tlambert03: That there is the pattern for creating a decorator that accepts optional arguments! Strange huh?  Both of these are valid:

```python
@thread_worker
def my_function():
    ...

@thread_worker(worker_class=ProgressWorker)
def my_function():
    ...
```

> In some sense, `inner_func` is the decorator, right

it is, but only if the outer decorator is used with arguments... (in which case `function` will _not_ be `None`)

napari/_qt/threading.py
tlambert03: yes probably... but after quickly looking to see if I could unify them, I decided to wait because of how many _other_ methods there are on the `AnimationWorker` (it's very stateful...)

More broadly (this gets into the weeds but is important), i could definitely refactor this whole PR to be using [`QThreads`](https://doc.qt.io/qt-5/qthread.html) and [`QObjects`](https://doc.qt.io/qt-5/qobject.html) _without_ using [`QRunnable`](https://doc.qt.io/qt-5/qrunnable.html) at all (as was done for the `AnimationWorker`)... and it would be a relatively easy refactor in the future without changing anything about the front-facing API.  The reason I opted for `QRunnable` is because it is the simpler pattern, and because we can take advantage of Qts can thread management in the global [`QThreadPool`](https://doc.qt.io/qt-5/qthreadpool.html) instance (which has a few nice convenience methods).  In fact, [this chart](https://doc.qt.io/qt-5/threads-technologies.html#comparison-of-solutions) makes it _appear_ that `QRunnable` can't receive updates through signals, but it is perfectly doable with multiple inheritance from `QObject` as I did in `WorkerBase`.

If we went with QThreads instead of QRunnables, then it would immediately unify the `AnimationWorker` API... but it would require us to build our own global thread management logic.  In the future, we may actually _want_ our own `ThreadPool` model (i started, but pulled back on it) that would accept any type of threading backend... but I opted to leave that for another day.

napari/viewer.py
tlambert03: Short answer: the default behavior of `wait_for_workers_to_quit` is to request all threads to quit, then block _indefinitely_ until threads finish (the safest... leaving it up to the user to "force quit" if a thread is not gracefully closing).  You can call it with a timeout `wait_for_workers_to_quit(1000)` that will return False if 1 second passes and threads have still not successfully finished... and we could decide to try to kill threads ourselves at that point.

But this is another super interesting and important detail (and I'm definitely still learning!).  Turns out it is _extremely_ hard to kill a misbehaving thread that is not continually "asking" whether it should be killed.  ([This SO thread](https://stackoverflow.com/questions/323972/is-there-any-way-to-kill-a-thread) is super useful).  So that is one reason why the `start_worker` function demands that the worker be an instance of `WorkerBase` ... where we know we've added some periodic checks for `_abort_requested` in the event loop.  But here too, there is a catch here (which again, I will add to the user-facing docs).  If the user does _not_ put any yields in their function, and the function is super long, it will just hang...  For instance:
```python
@thread_worker
def ZZZzzz():
    time.sleep(10000000)
```

There's no graceful way to kill this thread in python.  This is why I added the yield to the `layer_update` function in `examples/interactive_scripting.py`.  My (preliminary) understanding from [that SO post](https://stackoverflow.com/a/325528/1631624) is that, should we choose to, we can might be able to use `ctypes` to leverage OS-specific mechanisms to forcefully (and dangerously) kill misbehaving threads.  But if I'm not mistaken, that's not dissimilar from the user having to force quit... and I opted (for now) to leave that last `by_hook_or_by_crook` step to the user...

Curious what you think.  I agree from a user perspective, if you say quit, you want **QUIT**.  I just don't yet know enough about this stuff to feel super comfortable baking in potentially dangerous thread-killing logic.  (maybe a future PR?)

tlambert03 (commented):
Thanks so much @jni.  Lots of really important points here, so forgive the long explanations!

tlambert03:
one more comment: I feel comfortable with the front-facing API (besides a few naming changes) and the Qt-based backend used here.  I think it will work and immediately help in certain cases... but I also think it would be useful to start working on our own pure-python backend using `concurrent.futures` or something.  maybe @pwinston wants to take a look at the patterns?  It boils down to some pretty standard stuff:

**Front end:**
- define generator based API and `@thread_worker` decorator that puts a simple function/generator in another thread

**Back end:**
- create `ThreadPool` that remembers all the workers with methods to change the size of the pool
- create function to start `Workers` in a thread from the pool.
- run the user-provided generator function in an event loop (e.g. `WorkerBase.run -> Worker.work`)
- provide methods for starting/stopping/pausing the event loop
- provied methods for cleaning up the pool.

AhmetCanSolak:
great effort @tlambert03 ! I like the the detail of documentation and (most of) the introduced API.

So far i was using signals and slots for two-way communication between QThreads too and i have chosen a relatively simpler( and possibly less extensible way). I am wondering if we can simplify our implementation here while keeping the functionality we want to provide. I'm not sure if we need to implement our own event loop as in this PR right now. We can handle basic default signals in the worker run/start slot and let API user handle specific signals in their closure/function they pass to worker. What do you think about that?

Regarding first yield and other yields or different types of yields, I can see how easily that can become a trap for API user. As far as i understood first yield runs on master thread and to setup some stuff if needed, but API user gonna create and start worker from the main thread right? so why don't we just let API user to handle any setup needs out of worker? Am I missing some other use cases of the first yield here?

Regarding yields and achieving two-way communication, I am thinking if we can come up with an abstraction like `WorkerSignal` in which we can let API user to create custom signals for their use before running the worker and emit those created signals in needed places in worker closure/function. This would be a great flexibility in my opinion for API users. This can help to reduce use of generators and possibly bring a leaner implementation. 

Looking forward to see this pr in action!

examples/multithreading_simple.py
AhmetCanSolak: I like `_start_thread=True` but another idea could be letting user create their worker objects from constructor if they need `worker.start()`. I would have easier time if i can do:
`worker = Worker(myfunc, *args)`
`worker.start()`

if i use the decorator on `myfunc` then i expect calling `myfunc` should run it in a separate thread immediately.

tlambert03:
> I am wondering if we can simplify our implementation here while keeping the functionality we want to provide. I'm not sure if we need to implement our own event loop as in this PR right now. We can handle basic default signals in the worker run/start slot and let API user handle specific signals in their closure/function they pass to worker. What do you think about that?

very open to suggestions!  But I think I'd need some more concrete examples in order to follow

> API user gonna create and start worker from the main thread right? so why don't we just let API user to handle any setup needs out of worker?

Yeah that's definitely possible.  I think the main reason was to allow, much like an __init__ function, some stuff to be added to the worker at the moment of instantiation, rather than the declaration of the setup in the function... as shown [here](https://github.com/napari/napari/pull/1210#discussion_r425165297).  But happy to just remove that first yield if no one wants that.  (also, it's _only_ problematic for users who decorate a generator, not a pure function ...  so in a sense, it's a self-selecting group)

> I am thinking if we can come up with an abstraction like WorkerSignal in which we can let API user to create custom signals for their use before running the worker and emit those created signals in needed places in worker closure/function.

I'm afraid I'm gonna need an example here too ...

examples/multithreading_simple.py
tlambert03: > I would have easier time if i can do `worker = Worker(myfunc, *args); worker.start()`

you can also do that.

> if i use the decorator on myfunc then i expect calling myfunc should run it in a separate thread immediately.

i can see that... but I also think some people will decorate a function, then try to connect some stuff... but have it fail cause the worker was already off and running.  so the "default" behavior is a bit tricky to decide

napari/_qt/threading.py
AhmetCanSolak: I like going for `QRunnable` and having them in a single `QThreadpool`. Since every workers ends up in a new thread they can access to global context and they subscribe to objects in main thread, via updating objects actually `QRunnable` _can_ get updates from outside context. Are there problems im missing with this approach?

AhmetCanSolak:
> very open to suggestions! But I think I'd need some more concrete examples in order to follow
> 

I meant something like:
```
class Worker(QRunnable):

    def __init__(self, fn, *args, **kwargs):
        super(Worker, self).__init__()

        # Store constructor arguments (re-used for processing)
        self.fn = fn
        self.args = args
        self.kwargs = kwargs

        self.kwargs['special_signal'] = signal.special
        self.kwargs['progress_signal'] = signal.progress
        self.kwargs['result_signal'] = signal.result

    @pyqtSlot()
    def run(self):
        try:
            result = self.fn(*self.args, **self.kwargs)
        except Exception as e:
            self.signals.error.emit(e)
        else:
            self.signals.result.emit(result)
        finally:
            self.signals.finished.emit()
```
we handle error finish and result/return signals and API user handles special signals(yielded or whatever). Could be hard to handle generators this way though. Depending on how we want API users to use generators.


> Yeah that's definitely possible. I think the main reason was to allow, much like an **init** function, some stuff to be added to the worker at the moment of instantiation, rather than the declaration of the setup in the function... as shown [here](https://github.com/napari/napari/pull/1210#discussion_r425165297). But happy to just remove that first yield if no one wants that. (also, it's _only_ problematic for users who decorate a generator, not a pure function ... so in a sense, it's a self-selecting group)
> 

It is fancy the way it is right now :smile: I am just not sure about the need. Love to hear others opinions too.

> 
> I'm afraid I'm gonna need an example here too ...

What i meant here is just a wrapper around pyqt signals for API user to define and use custom signals with Workers

examples/multithreading_simple.py
AhmetCanSolak: I agree it is tricky. I am thinking we give option to connect things with first usage and with decorator we can give call and run usage.

tlambert03:
I think I see ... that `Runnable` you show there is pretty much just like the `BaseWorker` in this PR, but without the signals  so it sounds like your main points/issues are that A) we should not provide default signals like `yielded`, `paused`, `aborted`, asking users to provide them instead, and B) we should not accept generators and run them in our event loop.  Is that a correct interpretation?  If so, (but correct me if I misinterpreted!) ... 

1.  I think that adding a few extra default signals (like `yielded` and `aborted`) allows us to begin with a standardized set of signals that will meet the vast majority of use cases, without requiring the user to know or import _anything_ from Qt... which was the original goal here.  It doesn't really cost us to have a couple more signals there, and they may definitely be useful.  (though I could certainly buy the argument that we don't need `paused` or `resumed` signals.

2. The generator part I actually do feel kinda strongly about though.  It's not required, you _can_ use a standard function.  But if you do use a generator, you A) buy into a richer set of features, with built-in support for things like progress signals, stopping and starting, aborting, etc... B) you get two-way communication with pure-python syntax (generator.send),  ... and C) perhaps most importantly, you let _us_ provide a graceful exit strategy (by checking `is_aborted` each time in our own event loop).  With the current viewer.update strategy, it's very hard for us to gracefully shut down.  For instance in the interactive_scripting example, if you quit before the runnable is done... it throws a runtime error.  This offers a solution to that problem (which would be made clear in the docs). 

    The pure-python generator syntax also allows us to swap out the backend (away from Qt to `concurrent.futures` for example) without changing the user API.

tlambert03:
made a couple changes:
- broke up `Worker` into `GeneratorWorker` and `FunctionWorker`, to keep most of the complexity for generators seperate.  The `FunctionWorker` class is now super-simple.
- added `init_yield` argument to `GeneratorWorker`, so providing that first yield as setup is now "opt-in" behavior.
- use `re` in `ProgressWorker` to actually enforce correct usage at runtime (i.e. it makes sure there is a `yield` expression that yields a dict with the key `"__len__"`), otherwise raises an error.
- `worker_factory` is now called `create_worker`
- `start_thread` is now `False` by default (can change back)

napari/_qt/threading.py
```diff
@@ -0,0 +1,663 @@
...
+        """
+        self.started.emit()
+        self._running = True
+        try:
+            self.work()
+        except Exception as exc:
+            self.errored.emit(exc)
+        self.finished.emit()
+
+    def work(self):
```
AhmetCanSolak, line 121: btw in which order slot and base implementation will be called once work implementation from child class is called @tlambert03 ?

tlambert03: yeah, that should be in the docstring!

1. `QThreadPool.globalInstance().start(worker)` calls `Worker.run()`
2. `Worker.run` calls `Worker.work` (implemented by subclasses).

(as a convenience... `Worker.start()` is there to calls that first QThreadPool method)

AhmetCanSolak: gotcha. thanks.

tlambert03:
@jni, I believe [these](https://github.com/napari/napari/pull/1210/files/311e7560f92c..3e894e23f3a) are all the changes since you last looked at it.  happy to do more API tweaks.  Will take a look at the `futures` API tomorrow for guidance...

napari/_qt/threading.py
pwinston: The "setup in the main thread before calling yield" is definitely confusing to me. Just having a function that runs in two different threads is already confusing. And having that transition be controlled by "the first yield" seems magic and unconventional? Or is this in fact a pattern from somewhere?

It's weird that line 1 of the function will start in a different thread based on something that comes later in the function (whether it's a generator). Once you get it it makes sense, but it is kind of hidden. I guess the way I'd illustrate that it's hidden is I imagine every function every user writes will probably contain a giant comment "this part runs in the main thread" and "now we are in the worker thread" to explain it. The need for those comments to me is an illustration things are kind of hidden.

I think it's very compact and clever. It's a clever way to fit in a lot of functionality using implicit built-in stuff, without anything explicit. But that's the drawback as well. I think we are very proud to say "this is really simple, you just define one function", but in fact it's not actually simple.

I really don't know a better way though. Not off hand I'd have to think about it. My guess is you'd get rid of the idea of having the user just define a function. And you'd have a WorkerThread base class people inherit from. Then you can make it clear which method runs in which thread, etc. Make everything way more explicit. But less compact/clever. Or maybe there is a model we can find of doing all this in a standard way?

This would work and it's totally learnable. I think people would figure it out from the documentation. But it always seem a bit magic/custom I think.

tlambert03: My comment above is a little out of date now: After @AhmetCanSolak's review, I made that setup strictly opt-in (with the `init_yield=True` option).  I agree it's too implicit to be the default.  I can also take it out all together as well and I wouldn't miss it too much.  Is this whole comment just about that first yield ?  or are you talking more broadly?

napari/_qt/threading.py
jni: it needs a len and an iter, yes. I feel like `__iter__` could be the same as pre_yield_hook, more or less?

tlambert03: I played with this for a while actually... and while i got it to "work" (in the sense that tqdm worked), I never got around having the iterator block in the main thread again when used with tqm.  (which kinda makes sense... since it's the main thread calling "next").  Maybe we can leave this elaboration for a later PR?

napari/_qt/threading.py
jni: I do prefer going with the simpler pattern! So, if I can summarise my vague understanding of your comments:

- for AnimationWorker you used QThreads and QObjects because reasons
- But it *could* be refactored to use these QRunnable workers, it just needs some work that should go in a new PR
- That leaves QThreads and QObjects for people who... Like pain? Or what? I'm still not clear on what the *advantages* would be of keeping this exposed. So I am wondering whether we should just make this thing private and then eventually get rid of it.

tlambert03: lol.  Yes, its for people who like pain :). My current understanding of the limitation of using `QRunnable` is that you never get the `QThread` object itself (it's managed by the threadpool).  Take a quick look at the methods on [`QThread`](https://doc.qt.io/qt-5/qthread.html) and you'll see there's a lot more there, like `terminate` and `wait`.  Those are things that might come in handy if we wanted to create our own custom `close_by_hook_or_by_crook` shutdown process... and things that you do _not_ get access to if you allow the QThreadPool to turn your QRunnables into threads.

so:
**QRunnable** = free thread management, but you _must_ build in these little "hooks" to your workers to listen for a "stop this thread" signal.
**QThread** = complete control and responsibility over your threads... much harder to reuse threads (which causes more overhead as you create and destroy them) but easier to destroy rogue threads (at least as far as I've found)

I used QThreads for my animation worker because that was what I was more familiar with at the time.  but yes I _think_ it could be refactored.  I can make this method private right now

napari/viewer.py
jni: Awesome, thank you for writing this, it's super useful. As long as this makes it into the docs in some form, I'm happy to postpone the decision on whether to go on a murderous thread rampage.

napari/_qt/threading.py
```diff
@@ -0,0 +1,688 @@
...
+            kwargs['_start_thread'] = kwargs.get('_start_thread', start_thread)
+            kwargs['_connect'] = kwargs.get('_connect', connect)
+            kwargs['_worker_class'] = kwargs.get('_worker_class', worker_class)
+            return create_worker(func, *args, **kwargs,)
+
+        return wrapper
+
+    if function is None:
+        return decorator
+    return decorator(function)
```
jni: Ooooooh, right. I use `toolz.curry` for this, I think it ends up way prettier:

```python
@tz.curry
def thread_worker(function, start_thread=False, connect=None, worker_class=None):
    @functools.wraps
    def worker_function(*args, **kwargs):
        kwargs['_start_thread'] = kwargs.get('_start_thread', start_thread)
        kwargs['_connect'] = kwargs.get('_connect', connect)
        kwargs['_worker_class'] = kwargs.get('_worker_class', worker_class)
        return create_worker(func, *args, **kwargs)
    return worker_function
```

napari/_qt/threading.py
jni: Yeah, sure, I just think then maybe we could have a len and a counter and leave the calculation to the caller?

I'd also emit first and increment the counter second so that you don't need the -1?

napari/_qt/threading.py
jni: Very cool, so I suggest let's make this private for now and add that link to QThread methods so people who go fishing in our source are clear on what this provides.

jni:
The Linux failure looks fun! =P

napari/_qt/threading.py
tlambert03: Omg 😍

napari/_qt/threading.py
pwinston: I guess `init_yield=True` helps. In that case though, when True, I just feel it's very hidden that the one function will be called from two different threads. Unless it's some sort of convention or pattern people do with generators?

Basically we're saying in the user's function yield doesn't just mean "yield" it means "yield and, if this is the first yield, call me in another thread from now on". To me that's just inherently hidden/implicit. In the user's function there's no indication that anything with threading happening, yet it is totally happening, and in fact it can be critical to know which thread is running, but that's basically only evident in comments.

I'm not sure what else I'd suggest. I feel like it was strong goal here that user provide a **function** and not have `WorkerThread` base class? Given that constraint, this is a clever solution I guess. If it were just internal code, not providing an API for users, I would never make that constraint, that it has to be just a function. I'd probably have a base class with a super explicit `setup()` method with a big comment "This runs in the main thread" and a `run()` method that "runs in the worker thread".

But that's way clunkier. So maybe the trade off here is reasonable. There's a bit of magic but it's compact. I'm not sure, I would just never have gone down this road myself, but I see the benefits.

tlambert03:
> The Linux failure looks fun! =P

I was able to repeat this if I connect remotely to a linux box... but I am unable to repeat it when I run the tests (and try the examples) on a local linux setup.  So this seems to be related to virtualization somehow... not entirely sure how yet.  But I'm tempted to skip that test for linux CI

napari/_qt/threading.py
tlambert03: add docstring to both `worker.run` and `worker.start`

examples/multithreading_simple.py
tlambert03: I'm going to resolve this conversation to clean things up.  Current behavior is:
- non-prefixed arguments may be used on the decorator: `@thread_worker(start_thread=True)`
- prefixed arguments may be used when calling the function (though I don't think we should "demonstrate" this in our docs): `decorated_function(_start_thread=True)`
- decorated functions do not start workers immediately unless:
   - the decorator is used with `@thread_worker(start_thread=True)`
   - the decorated function is called with `decorated_function(_start_thread=True)`
   - the `start_thread` is not specified (`None`) _and_ the `connect` argument _is_ specified (not `None`) on either the decorator or when calling the function.  For example: `@thread_worker(connect={'returned': print})` will start the thread immediately upon calling the decorated function, under the assumptions that all of the required connections have already been declared.

napari/_qt/threading.py
tlambert03: comments added

napari/_qt/threading.py
tlambert03: renamed.  Note also that the `start_worker` method (which adds the worker to the `_WORKERS` set) is now lower down on the `WorkerBase.start` method... so calling `worker.start()` from any subclass will add the worker to our global pool.

napari/_qt/threading.py
tlambert03: fixed with the lovely `toolz.curry`

napari/viewer.py
tlambert03: added to docs

napari/_qt/threading.py
tlambert03: > I feel like it was strong goal here that user provide a function and not have WorkerThread base class

Yeah, I think that is a goal here.  I don't know whether this is an "explicit" goal of napari, but I think we generally try to offer at least one API that relative python newcomers (e.g. scientists who don't have the time to learn about class inheritance) could take advantage of to accomplish the majority of their needs in a relatively efficient/safe way.  This certainly doesn't prevent people from subclassing `WorkerBase` and taking it in any direction... and some may even use the `init_yield` in their subclasses (as I did with the `ProgressWorker`).  And since the default is `init_yield==False`, it's as if it wasn't there for users who don't want it.

But your point is well taken!  There's nothing that couldn't be done here in a more explicit way in the main thread by the end-user.  I'm not too married to this pattern (i kinda like it, but I don't _love_ it), and happy to keep or remove it according to the group consensus.

tlambert03:
> > The Linux failure looks fun! =P
> 
> I was able to repeat this if I connect remotely to a linux box... but I am unable to repeat it when I run the tests (and try the examples) on a local linux setup. So this seems to be related to virtualization somehow... not entirely sure how yet. But I'm tempted to skip that test for linux CI

scratch that.  Turns out it wasn't a linux thing, but rather PyQt5 not liking a multiple inheritance strategy I used.  For those interested:  PyQt5 seems not to like creating python classes that inherit from multiple Qt classes (e.g. `class Worker(QRunnable, QObject)`) ... while PySide2 seems to have no problems with it.  Safest bet is always to use composition when combining Qt classes.

tlambert03:
ok, I was able to fix tests on all platforms and backends.  The worker object now has a `worker._signals` object where the `QObject` with the signals actually lives (like in @AhmetCanSolak's [example](https://github.com/napari/napari/pull/1210#issuecomment-628842809)), but I made them accessible on the worker object as well with a `__getattribute__` mod to keep the public worker API as clean as possible...  so: `worker._signals.finished == worker.finished`

I have also now made it so that, by default, errors raised in the other thread are re-raised in the main thread, unless `@thread_worker` is used with `ignore_errors=True`.  This should help debug mysterious errors in the other thread (which I was running up against while trying to fix this).

Outside of the conversation happening about the general design of the `GeneratorWorker` happening in [this thread](https://github.com/napari/napari/pull/1210#discussion_r425511680) (which we should continue having until everyone is happy...), this is ready for a final review.

tlambert03:
one other thing to discuss here is the private `_qt` module in something we want people to important use.  We could either expose this in `napari.threading` ...  importing some internal qt stuff there (we can make it a try/except if we ever offer different backends) or do something along the lines of #1122 ... but probably not by changing `_qt` everywhere ... but rather adding a `napari.qt` module that exposes the few things we want

sofroniewn:
> but rather adding a napari.qt module that exposes the few things we want

I lean towards this, it gives us more control

tlambert03:
> > but rather adding a napari.qt module that exposes the few things we want
> 
> I lean towards this, it gives us more control

(this actually will help our API docs a lot too... since it defaults to only public things... it would be a nice way to expose just what we want in the docs as well)

examples/interactive_scripting.py
```diff
@@ -1,5 +1,6 @@
 import numpy as np
 import napari
+from napari._qt.threading import thread_worker
```
sofroniewn: now that #1122 has merged will this go in `napari.qt`? Probably a good place for it

examples/multithreading_two_way.py
```diff
@@ -0,0 +1,113 @@
...
+class Controller(QWidget):
+    def __init__(self):
+        super().__init__()
+
+        layout = QGridLayout()
+        self.setLayout(layout)
+        self.status = QLabel('Click "Start"', self)
+        self.play_btn = QPushButton("Start", self)
+        self.abort_btn = QPushButton("Abort!", self)
+        self.reset_btn = QPushButton("Reset", self)
```
sofroniewn, line 48: Would be nice if pressing reset could send progress bar down to zero right away, instead it doesn't reset until continue is pressed. also reset doesn't seem to work after the whole example has finished and the "done" appears, i.e. i can't seem to run the whole thing twice which might be nice

amazing stuff though!

examples/multithreading_two_way.py
```diff
@@ -0,0 +1,113 @@
...
+        w.play_btn.clicked.connect(handle_pause)
+
+    worker.started.connect(on_start)
+    return w
+
+
+if __name__ == "__main__":
+
+    with napari.gui_qt():
+        viewer = napari.Viewer()
```
sofroniewn: Could we add a random image here too, would be nice to have something to look at to see how pan/ zoom keep working during this whole time

examples/interactive_scripting.py
tlambert03: will move

examples/multithreading_two_way.py
tlambert03: resetting to zero immediately is not technically possible with threading... (Clicking the "reset" button is more of a "request" to the thread that it may or may not honor). 
 We could just "put" the slider to zero, but until the thread "checks in" there's really no way to know where the thread is at.  (There's a concept in web design called "optimistic updates" in which you set the GUI to what you "hope" will happen, and after the asynchronous call returns, you fix the GUI if you were wrong ... but I think for this example, that would unnecessarily hide the fact that asynchronous work is definitely a bit different).

Reusing workers is possible... but adds a level of difficulty (which is why I disable the button on finished).  I will look into that in a future PR

sofroniewn: ah interesting, maybe then just disable the reset button when the example is done?

tlambert03: done

examples/multithreading_simple.py
```diff
@@ -0,0 +1,45 @@
...
+    # worker = long_running_function(
+    #     _connect={
+    #         'started': lambda: wdg.status.setText("worker is running..."),
+    #         'returned': lambda x: wdg.status.setText(f"returned {x!r}"),
+    #     }
+    # )
+
+    worker.start()
+
+    app.exec_()
```
jni, line 45: At first I was like, wait, why not use `napari.gui_qt`... Then I re-read the function and realised this has no napari in it! :joy:

tlambert03: _super_ minimal example :)

napari/_qt/threading.py
```diff
@@ -0,0 +1,834 @@
...
+        """Main method to execute the worker.
+
+        The end-user should never need to call this function.
+        But subclasses must implement this method (See
+        :meth:`GeneratorFunction.work` for an example implementation).
+        Minimally, it should check ``self.abort_requested`` periodically and
+        exit if True.
+
+        Example
+        -------
```
jni:
```suggestion
        Examples
        --------
```

jni: https://numpy.org/doc/stable/docs/howto_document.html#sections

napari/_qt/threading.py
```diff
@@ -0,0 +1,842 @@
...
+        return out
+
+    @property
+    def is_paused(self) -> bool:
+        """Whether the worker is currently paused."""
+        return self._paused
+
+    def toggle_pause(self) -> None:
+        """Send a request to pause the worker if playing or resumed if paused.
+        """
```
jni:
```suggestion
        """Request to pause the worker if playing or resume if paused."""
```

napari/_qt/threading.py
```diff
@@ -0,0 +1,842 @@
...
+        if self.is_paused:
+            self._resume_requested = True
+        else:
+            self._pause_requested = True
+
+    def pause(self) -> None:
+        """Send a request to pause the worker.
+        """
```
jni:
```suggestion
        """Request to pause the worker."""
```

napari/_qt/threading.py
```diff
@@ -0,0 +1,842 @@
...
+        if self.is_paused:
+            self._resume_requested = True
+
+    def pre_yield_hook(self):
+        """Hook for subclasses. Called just before yielding from generator"""
+        pass
+
+    def post_yield_hook(self):
+        """Hook for subclasses. Called just after yielding from generator"""
+        pass
```
jni: Are these used now that there is no progressworker? If not, I would also remove them from the PR.

napari/_qt/threading.py
```diff
@@ -0,0 +1,840 @@
...
+    _ignore_errors : bool, optional
+        If ``False`` (the default), errors raised in the other thread will be
+        reraised in the main thread (makes debugging significantly easier).
+    *args
+        will be passed to ``func``
+    **kwargs
+        will be passed to ``func``
+
+    Example
+    -------
```
jni:
```suggestion
    Examples
    --------
```

jni: Actually this should be after Returns anyway

napari/_qt/threading.py
```diff
@@ -0,0 +1,840 @@
...
+
+        def long_function(duration):
+            import time
+            time.sleep(duration)
+
+        worker = create_worker(long_function, 10)
+
+    Returns
+    -------
+    WorkerBase
```
jni:
```suggestion
    worker : WorkerBase
```

napari/_qt/threading.py
```diff
@@ -0,0 +1,840 @@
...
+    if not _worker_class:
+        if inspect.isgeneratorfunction(func):
+            _worker_class = GeneratorWorker
+        else:
+            _worker_class = FunctionWorker
+
+    if inspect.isclass(_worker_class) and issubclass(
+        WorkerBase, _worker_class
+    ):
+        raise TypeError(f'Worker {_worker_class} must be a subclass of Worker')
```
jni: should this be **not** issubclass???

jni: I'm very confused about this clause...

jni: it also seems like the classes are inverted. ie I would write `and not issubclass(_worker_class, WorkerBase)`. But I'm not sure whether that's the intent or I'm misunderstanding something.

napari/_qt/threading.py
```diff
@@ -0,0 +1,732 @@
...
+
+    # if the user has not provided a default connection for the "errored"
+    # signal... and they have not explicitly set ``ignore_errors=True``
+    # Then rereaise any errors from the thread.
+    if not _ignore_errors and not (_connect or {}).get('errored', False):
+
+        def reraise(e):
+            raise e
+
+        worker.errored.connect(reraise)
```
jni, line 523: Very cute.

jni: though actually here I would use a lambda, would make the whole thing more compact: `worker.errored.connect(lambda e: raise e)`. Though can lambdas raise?

.... TIL that lambdas can't raise!

napari/_qt/threading.py
```diff
@@ -0,0 +1,732 @@
...
+        reraised in the main thread (makes debugging significantly easier).
+
+    Returns
+    -------
+    callable
+        function that creates a worker, puts it in a new thread and returns
+        the worker instance.
+
+    Example
+    -------
```
jni:
```suggestion
    Examples
    --------
```

tlambert03:
> though actually here I would use a lambda, would make the whole thing more compact: worker.errored.connect(lambda e: raise e). Though can lambdas raise?
.... TIL that lambdas can't raise!

yeah... `raise` is a statement... and lambdas cant have statements.

napari/_qt/threading.py
```diff
@@ -0,0 +1,732 @@
...
+
+    .. note:: Mostly ignorable detail
+
+        While the signals/slots syntax of the worker looks very similar to
+        standard "single-threaded" signals & slots, note that inter-thread
+        signals and slots (automatically) use an event-based QueuedConnection,
+        while intra-thread signals use a DirectConnection. See `Signals and
+        Slots Across Threads
+        <https://doc.qt.io/qt-5/threads-qobject.html#signals-and-slots-across-threads>`_
+
```
jni, line 693:
```suggestion
```

napari/_qt/threading.py
```diff
@@ -0,0 +1,732 @@
...
+        be manually started with thread.start().
+    connections: dict, optional
+        Optional dictionary of {signal: function} to connect to the new worker.
+        for instance:  connections = {'incremented': myfunc} will result in:
+        worker.incremented.connect(myfunc)
+    *args
+        will be passed to the Worker class on instantiation.
+    **kwargs
+        will be passed to the Worker class on instantiation.
+
```
jni, line 710: Missing Returns section

napari/_qt/threading.py
tlambert03: yeah, that was all wrong.  thanks

napari/_qt/threading.py
```diff
@@ -171,13 +177,33 @@ def start(self):
...
+    ------
+    TypeError
+        If ``func`` is a generator function and not a regular function.
     """
 
     def __init__(self, func: Callable, *args, **kwargs):
+        """[summary]
+
+
+        """
```
jni: ?????

napari/_qt/threading.py
```diff
@@ -682,6 +712,12 @@ def _new_worker_qthread(
     **kwargs
         will be passed to the Worker class on instantiation.
 
+    Returns
+    -------
+    tuple
+        ``(worker, thread)``, where ``worker`` is an instance of ``Worker``,
+        and ``thread`` is an instance of ``QThread``.
```
jni: normally this would be:

```suggestion
worker : WorkerBase
    The created worker.
thread : QThread
    The thread on which the worker is running.

Comment (tlambert03) [reply napari/_qt/threading.py]: oh I wondered!  i didn't know you could do that.  thanks

File: napari/_qt/threading.py [hunk 33]
```diff
@@ -710,9 +710,10 @@ def _new_worker_qthread(
 
     Returns
     -------
-    tuple
-        ``(worker, thread)``, where ``worker`` is an instance of ``Worker``,
-        and ``thread`` is an instance of ``QThread``.
+worker : WorkerBase
+    The created worker.
+thread : QThread
+    The thread on which the worker is running.
```
jni: Ach, sorry, stuffed up the indent:

```suggestion
    worker : WorkerBase
        The created worker.
    thread : QThread
        The thread on which the worker is running.
```

jni (approved):
:fireworks:

PR #1217, opened 2020-05-07
Use grayscale when n_channels=1

### LINKED ISSUE #1216: Don't use magenta when `channel_axis` is given but nchannels is 1
## 🐛 Bug

The OME-zarr [plugin](https://github.com/ome/ome-zarr-py) nicely grabs the channel axis from the OME standard layout. However, for single channel images, this results in the selected colormap being magenta. It should be gray if there is only one channel.

```python
import numpy as np
import napari

with napari.gui_qt():
    _ = napari.view_image(
        np.random.random(1, 100, 100), channel_axis=0, name='I ❤️ purple!'
    )
```

Screen Shot 2020-05-07 at 5 00 29 pm

jni:
# Description

Closes #1216 

## Type of change

# References

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/components/add_layers_mixin.py
+++ b/napari/components/add_layers_mixin.py
@@ -237,7 +237,9 @@ class AddLayersMixin:
             # so that we can use {k: next(v) for k, v in kwargs.items()} below
             for key, val in kwargs.items():
                 if key == 'colormap' and val is None:
-                    if n_channels < 3:
+                    if n_channels == 1:
+                        kwargs[key] = iter(['gray'])
+                    elif n_channels == 2:
                         kwargs[key] = iter(colormaps.MAGENTA_GREEN)
                     else:
                         kwargs[key] = itertools.cycle(colormaps.CYMRGB)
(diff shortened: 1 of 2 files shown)


napari/components/add_layers_mixin.py
```diff
@@ -237,7 +237,9 @@ def add_image(
             # so that we can use {k: next(v) for k, v in kwargs.items()} below
             for key, val in kwargs.items():
                 if key == 'colormap' and val is None:
-                    if n_channels < 3:
+                    if n_channels == 1:
```
kevinyamauchi, line 240: This is more of a curiosity and if we do want to change this, it's definitely out of the scope of this PR, but why do we do these types of kwarg checks/cleaning in `add_image()` rather than the `Image` model init method? It seems like this would lead to a divergence in how the model behaves on its own versus how it effectively behaves with the viewer.

kevinyamauchi (approved):
This looks good to me! Good call - it's definitely better to default to grayscale for 1 channel images.

napari/components/add_layers_mixin.py
jni: Because `Image` doesn't have `channel_axis`, since that requires the creation of multiple Image layers

kevinyamauchi: Oh, right 🤦 Thanks for the explanation!

sofroniewn (approved):
LGTM!

tlambert03 (approved):
grayscale FTW!

PR #1219, opened 2020-05-07
Improve plugin error messages, bump napari-plugin-engine

tlambert03:
# Description
Improves error messages in the following cases:
- if `read_data_with_plugins` has errors, in addition to showing `"No plugin found capable of reading ... "`, also shows which plugins tried and failed... and a reminder to look in the Plugins > Plugins Errors... menu
- if `open()` or `save()` is called specifying an invalid plugin name, improves the error message by showing the available plugin names.
- bump `napari-plugin-engine`, which now has an improved error format that shows the exact file and line in the plugin where the error occurred... example:
```no-highlight
  Cause was: ValueError("No files found in ['/Users/talley/Desktop/covid-gut'] after removing subdirectories")
    in file: ~/python/forks/napari/napari/utils/io.py
    at line: 195
    package: napari
    version: 0.3.0+1.g565089a4.dirty
```

## Type of change

- [x] Bug-fix, sorta (non-breaking change which fixes an issue)

# How has this been tested?

(will try to add some `assert raises` soon)


sofroniewn (approved):
This is really nice @tlambert03 - i really like that `read_data_with_plugins` gives you back the empty list now and does all the error handling. very clean. LGTM!!

napari/plugins/io.py
```diff
@@ -208,6 +241,11 @@ def _write_multiple_layers_with_plugins(
     if plugin_name:
         # if plugin has been specified we just directly call napari_get_writer
         # with that plugin_name.
+        if plugin_name not in plugin_manager.plugins:
+            raise ValueError(
+                f"There is no registered plugin named {plugin_name!r}.\n"
+                f"Plugin names include: {set(plugin_manager.plugins)!r}"
```
jni: Include, or are? This looks like all available plugins? Also, should we only list the plugins that provide a writer hook spec here? ie “The available plugins for saving are:”

napari/plugins/io.py
```diff
@@ -278,6 +316,12 @@ def _write_single_layer_with_plugins(
         plugin_manager.hook, f'napari_write_{layer._type_string}'
     )
 
+    if plugin_name and (plugin_name not in plugin_manager.plugins):
+        raise ValueError(
+            f"There is no registered plugin named {plugin_name!r}.\n"
+            f"Plugin names include: {set(plugin_manager.plugins)!r}"
```
jni: Why is this segment repeated? I can’t follow the flow...

napari/components/_tests/test_add_layers.py
```diff
@@ -30,7 +30,7 @@ def test_add_layers_with_plugins(layer_datum):
 
 @patch(
     "napari.components.add_layers_mixin.read_data_with_plugins",
-    MagicMock(return_value=None),
+    MagicMock(return_value=[]),
```
jni, line 33: What if a plugin actually returns None?

jni (commented):
I have a few minor suggestions/questions @tlambert03 ! Would you mind having a look?

jni:
I love the new messages btw they are excellent

napari/components/_tests/test_add_layers.py
tlambert03: if a plugin returns None, then `read_data_with_plugins` will still return `[]`.

napari/plugins/io.py
tlambert03: one is for readers (coming from `viewer.open`) the other two are for writers (coming from `save_layers`)...  In both cases, I put the check as low as possible in the call chain, so it can't be missed.  For writers, I _could_ move the ones in `_write_multiple_layers_with_plugins` and `_write_single_layer_with_plugins` up into `save_layers` (and let it be a more cryptic error if someone calls the private functions incorrectly) but then it would also be best if I added a `plugin_manager` argument to that public function, which I was trying to avoid.  let me know what you prefer

tlambert03: perhaps also worth pointing out... this check is technically not necessary at all.  It would still be caught in the hookcaller, but in that case the error message would be something like `"No implementation of napari_get_reader found for plugin ___"` ... which makes it sound like there _might_ be a plugin with that name... (but the hookcaller doesn't have access to the plugin manager and doesn't know that for sure)

napari/plugins/io.py
tlambert03: fixed

napari/plugins/io.py
```diff
@@ -242,9 +243,10 @@ def _write_multiple_layers_with_plugins(
         # if plugin has been specified we just directly call napari_get_writer
         # with that plugin_name.
         if plugin_name not in plugin_manager.plugins:
+            names = {i.plugin_name for i in hook_caller.get_hookimpls()}
             raise ValueError(
-                f"There is no registered plugin named {plugin_name!r}.\n"
-                f"Plugin names include: {set(plugin_manager.plugins)!r}"
+                f"There is no registered plugin named '{plugin_name}'.\n"
+                f"Names of plugins offering readers are: {names}"
```
jni:
```suggestion
                f"Names of plugins offering writers are: {names}"
```
:grimacing:

jni (approved):

PR #1220, opened 2020-05-07
CLI accepts --plugin or any add_* kwargs

tlambert03:
# Description
This adds some features to the command line program:

- accept `--plugin` argument that uses a specific plugin to open a file.  An error will be thrown if this argument is used without specifying a file to open.
- add `-v`, `-vv`, verbosity flags that change the logging level from `WARNING` (default) to `INFO` (with `-v`) or `DEBUG` (with `-vv`)
- added the ability to pass "unrecognized" arguments to the command line, that will be parsed, validated, and passed to `viewer.open()`.  If any of the arguments are not valid for at least one of the `viewer.add_*` methods, the program quits and warns (so we don't lose the "unrecognized arguments" error.  Values are parsed with `literal_eval`.

All together, allows stuff like this:

```
napari -vv path/to/file.tif --plugin builtins --gamma 0.6 --contrast_limits '(0, 700)' --name confocal
```
while preserving helpful errors like this:

```bash
# should be "--plugin" not "--plugins"
napari path/to/file.tif --plugins builtins
error: unrecognized arguments: --plugins
```

and combined with #1219  ... will generally make testing reader plugin easier.

## Type of change

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/__main__.py
+++ b/napari/__main__.py
@@ -27,20 +27,6 @@ class CitationAction(argparse.Action):
         sys.exit()
 
 
-def get_valid_kwargs():
-    """Get a set of valid argument names for add_* methods."""
-    from .components.add_layers_mixin import AddLayersMixin
-    import inspect
-
-    valid = set()
-    for meth in dir(AddLayersMixin):
-        if not meth.startswith('add_'):
-            continue
-        params = inspect.signature(getattr(AddLayersMixin, meth)).parameters
-        valid.update(set(params) - {'self', 'kwargs'})
-    return valid
-
-
 def validate_unknown_args(unknown: List[str]) -> Dict[str, Any]:
     """Convert a list of strings into a dict of valid kwargs for add_* methods.
 
@@ -58,14 +44,17 @@ def validate_unknown_args(unknown: List[str]) -> Dict[str, Any]:
         {key: val} dict suitable for the viewer.add_* methods where ``val``
         is a ``literal_eval`` result, or string.
     """
+    from .components.add_layers_mixin import valid_add_kwargs
+
     out: Dict[str, Any] = dict()
-    valid = get_valid_kwargs()
+    valid = set.union(*valid_add_kwargs().values())
     for i, arg in enumerate(unknown):
         if not arg.startswith("--"):
             continue
         if "=" in arg:
             sys.exit(f"error: '=' in argument {arg}. (Use space instead)")
-        if arg[2:] not in valid:
+        key = arg[2:].replace("-", "_")
+        if key not in valid:
             sys.exit(f"error: unrecognized arguments: {arg}")
         try:
             next_arg = unknown[i + 1]
@@ -73,19 +62,24 @@ def validate_unknown_args(unknown: List[str]) -> Dict[str, Any]:
                 raise IndexError()
         except IndexError:
             sys.exit(f"error: argument {arg} expected one argument")
-            get_valid_kwargs
         try:
             val = literal_eval(next_arg)
         except Exception:
             val = next_arg
-        out[arg[2:]] = val
+        out[key] = val
     return out
 
 
 def main():
     parser = argparse.ArgumentParser(usage=__doc__)
     parser.add_argument('images', nargs='*', help='Images to view.')
-    parser.add_argument('-v', '--verbose', action='count', default=0)
+    parser.add_argument(
+        '-v',
+        '--verbose',
+        action='count',
+        default=0,
+        help="increase output verbosity",
+    )
     parser.add_argument(
         '--version', action='version', version=f'napari version {__version__}',
     )
@@ -108,20 +102,6 @@ def main():
         '--stack',
         action='store_true',
         help='Concatenate multiple input files into a single stack.',
-    )
-    parser.add_argument(
-        '-r',
-        '--rgb',
-        help='Treat images as RGB.',
-        action='store_true',
-        default=None,
-    )
-    parser.add_argument(
-        '-g',
-        '--grayscale',
-        dest='rgb',
-        action='store_false',
-        help='interpret all dimensions in the image as spatial',
     )
 
     args, unknown = parser.parse_known_args()
--- a/napari/__main__.py
+++ b/napari/__main__.py
@@ -105,7 +105,7 @@ def main():
     )
 
     args, unknown = parser.parse_known_args()
-    kwargs = validate_unknown_args(unknown)
+    kwargs = validate_unknown_args(unknown) if unknown else {}
 
     # parse -v flags and set the appropriate logging level
     levels = [logging.WARNING, logging.INFO, logging.DEBUG]
--- a/napari/__main__.py
+++ b/napari/__main__.py
@@ -52,22 +52,29 @@ def validate_unknown_args(unknown: List[str]) -> Dict[str, Any]:
     for i, arg in enumerate(unknown):
         if not arg.startswith("--"):
             continue
+
         if "=" in arg:
-            sys.exit(f"error: '=' in argument {arg}. (Use space instead)")
-        key = arg.lstrip('-').replace("-", "_")
+            key, value = arg.split("=", maxsplit=1)
+        else:
+            key = arg
+        key = key.lstrip('-').replace("-", "_")
+
         if key not in valid:
             sys.exit(f"error: unrecognized arguments: {arg}")
+
+        if "=" not in arg:
+            try:
+                value = unknown[i + 1]
+                if value.startswith("--"):
+                    raise IndexError()
+            except IndexError:
+                sys.exit(f"error: argument {arg} expected one argument")
         try:
-            next_arg = unknown[i + 1]
-            if next_arg.startswith("--"):
-                raise IndexError()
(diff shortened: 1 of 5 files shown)


jni:
@tlambert03 I need to play with this a bit more, but on first pass I love it!!! Great idea!

sofroniewn (approved):
This looks good to me, everyone was very excited at team meeting so I am approving

napari/__main__.py
```diff
@@ -2,32 +2,89 @@
...
+    """
+    from .components.add_layers_mixin import valid_add_kwargs
+
+    out: Dict[str, Any] = dict()
+    valid = set.union(*valid_add_kwargs().values())
+    for i, arg in enumerate(unknown):
+        if not arg.startswith("--"):
+            continue
+        if "=" in arg:
+            sys.exit(f"error: '=' in argument {arg}. (Use space instead)")
```
jni: I don’t love this. = is valid GNU syntax. Why not split the arg accordingly?

napari/__main__.py
```diff
@@ -2,32 +2,89 @@
...
+        key = arg[2:].replace("-", "_")
```
jni: 2?

napari/components/add_layers_mixin.py
```diff
@@ -968,6 +968,19 @@ def _add_layer_from_data(
         return layer
 
 
+@lru_cache(maxsize=1)
+def valid_add_kwargs() -> Dict[str, Set[str]]:
```
jni, line 962: I like making this lazy, I presume to minimise startup times in python, but it gets called every time with validate_unknown_kwargs. Maybe only call that function if there are any?

jni (commented):
@tlambert03 I’ve left some minor suggestions. Maybe also add a line to the help string with “other arguments: these will be passed to the corresponding layer or viewer”?

This is very exciting.

napari/components/add_layers_mixin.py
tlambert03: good idea, I'll make that call conditional. (but also, for the sake of completeness, here's how it performs on my laptop):

```python
In [3]: %time v = valid_add_kwargs()                                                                           
CPU times: user 387 µs, sys: 3 µs, total: 390 µs

In [4]: %time v = valid_add_kwargs()                                                                           
CPU times: user 3 µs, sys: 0 ns, total: 3 µs
```

napari/__main__.py
tlambert03: trimming the "--" prefix, which we asserted was present a couple lines above

napari/__main__.py
tlambert03: well, it was a bit of a cop out... because it makes parsing with literal_eval much harder.  For instance, these all work:
```
--contrast-limits "(1, 0)"
--contrast-limits "(1,0)"
--contrast-limits="(1,0)"
```
but this does not (because of the space in the tuple)
```
--contrast-limits="(1, 0)"
```

tlambert03:
> Maybe also add a line to the help string with “other arguments: these will be passed to the corresponding layer or viewer”?

It now looks like this, (note, I also added the `--layer-type` option). let me know if you have any formatting requests (or if you'd prefer the extended help to only show when requested)

```no-highlight
usage: 
napari command line viewer.

positional arguments:
  images             image(s) to view.

optional arguments:
  -h, --help         show this help message and exit
  -v, --verbose      increase output verbosity
  --version          show program's version number and exit
  --info             show system information and exit
  --citation         show citation information and exit
  --stack            concatenate multiple input files into a single stack.
  --plugin PLUGIN    specify plugin name when opening a file
  --layer-type TYPE  force file to be interpreted as a specific layer type.
                     one of {'image', 'points', 'labels', 'surface',
                     'vectors', 'shapes'}

optional layer-type-specific arguments (precede with '--'):
  Image:
    attenuation, blending, channel-axis, colormap, contrast-limits, data,
    gamma, interpolation, iso-threshold, metadata, multiscale, name,
    opacity, rendering, rgb, scale, translate, visible
  Labels:
    blending, data, metadata, multiscale, name, num-colors, opacity,
    scale, seed, translate, visible
  Points:
    blending, data, edge-color, edge-color-cycle, edge-colormap,
    edge-contrast-limits, edge-width, face-color, face-color-cycle,
    face-colormap, face-contrast-limits, metadata, n-dimensional, name,
    opacity, properties, scale, size, symbol, translate, visible
  Shapes:
    blending, data, edge-color, edge-width, face-color, metadata, name,
    opacity, scale, shape-type, translate, visible, z-index
  Surface:
    blending, colormap, contrast-limits, data, gamma, metadata, name,
    opacity, scale, translate, visible
  Vectors:
    blending, data, edge-color, edge-color-cycle, edge-colormap,
    edge-contrast-limits, edge-width, length, metadata, name, opacity,
    properties, scale, translate, visible
```

(the cost is that I have to call `valid_add_kwargs()` every time... but in my tests it takes less than 1ms...  which is a negligible portion of the time it already takes to import everything else:

napari/__main__.py
jni: I don't understand... use `split('=')` and it won't split on spaces?

```python
import sys
import argparse


parser = argparse.ArgumentParser()

known, unknown = parser.parse_known_args()

print(unknown)
```

```sh
$ python arg.py --contrast-limits="(0, 1)"
['--contrast-limits=(0, 1)']
```

And if I do `print(unknown[0].split('='))`:

```sh
 $ python argp.py --contrast-limits="(0, 1)"
['--contrast-limits', '(0, 1)']
```

napari/__main__.py
jni: Ah! :facepalm: 

Ok, this is nitpicky, but since I want you to fix the `=` anyway: how about:

```suggestion
        key = arg.lstrip('-').replace("-", "_")
```

napari/components/add_layers_mixin.py
jni: Hmm that might be approaching the cost of a conditional! :joy:

napari/__main__.py
tlambert03: ah, ok ... your example doesn't have the positional argument with `nargs='*'` (as we do for the `'images'` argument) ... that's what screws it all up:
```bash
# with  parser.add_argument('images', nargs='*', ...
# args, unknown = parser.parse_known_args()
# print(args,unknown)

$ napari path/to/image.tif --contrast-limits "(1, 0)"
Namespace(images=['path/to/image.tif']) ['--contrast-limits', '(1, 0)']

$ napari path/to/image.tif --contrast-limits "(1,0)"
Namespace(images=['path/to/image.tif']) ['--contrast-limits', '(1,0)']

$ napari path/to/image.tif --contrast-limits="(1,0)"
Namespace(images=['path/to/image.tif']) ['--contrast-limits=(1,0)']

$ napari path/to/image.tif --contrast-limits="(1, 0)"
Namespace(images=['path/to/image.tif', '--contrast-limits=(1, 0)']) []  # <- breaks

# with parser.add_argument('images', nargs=1...

$ napari path/to/image.tif --contrast-limits="(1, 0)"
Namespace(images=['path/to/image.tif']) ['--contrast-limits=(1, 0)']
```

jni: Damn that's so crap. It should not include things that have a `--`.

Well, this is probably not what you want to hear, but you *could* parse `args.images` and shuffle all the entries that start with `--` over to unknown...?

tlambert03: lol... fine.

jni: Well, it's that or submit a patch to argparse, your pick... =P

tlambert03: fixed!

napari/components/add_layers_mixin.py
```diff
@@ -1,8 +1,9 @@
+import inspect
 import itertools
 from functools import lru_cache
 from logging import getLogger
 from os import fspath
-from typing import Any, Dict, List, Optional, Sequence, Union, Set
+from typing import Any, Dict, List, Optional, Sequence, Set, Union
```
jni, line 6: Hahaha this change definitely reveals a personality type. :joy:

tlambert03: sorry :/ motor memory

napari/__main__.py
```diff
@@ -52,22 +52,29 @@ def validate_unknown_args(unknown: List[str]) -> Dict[str, Any]:
...
         if key not in valid:
             sys.exit(f"error: unrecognized arguments: {arg}")
+
+        if "=" not in arg:
+            try:
+                value = unknown[i + 1]
+                if value.startswith("--"):
+                    raise IndexError()
+            except IndexError:
+                sys.exit(f"error: argument {arg} expected one argument")
```
jni:
```suggestion
            key, value = arg.split("=", maxsplit=1)
        else:
            key = arg
            try:
                value = unknown[i + 1]
                if value.startswith("--"):
                    raise IndexError()
            except IndexError:
                sys.exit(f"error: argument {arg} expected one argument")
        key = key.lstrip('-').replace("-", "_")

        if key not in valid:
            sys.exit(f"error: unrecognized arguments: {arg}")
```

PR #1222, opened 2020-05-07
Specify viewer.open(plugins='builtins') for all tests

tlambert03:
# Description
Just saw some zarr tests failing locally that weren't failing on CI, and narrowed it down to different behavior between our `magic_imread` and the `ome_zarr` plugin, which I had recently installed in my env.  So this PR changes all of the `viewer.open()` calls in our tests to specify `plugin='builtins'` (which is the assumption in our tests anyway).  

as a sidenote: I haven't really looked into why they are behaving differently, but I think it has to do with stacking behavior in our function?  (their plugin added 20 layers and ours added 1).  This sort of thing is inevitable, but just wanted to point it out so people can start thinking about how/whether to eventually address that sort of issue where two plugins handle a similar file quite differently.  

## Type of change

# How has this been tested?

- [x] this makes tests more predictable regardless of environment.

## Final checklist:

### DIFF
--- a/napari/components/_tests/test_viewer_image_io.py
+++ b/napari/components/_tests/test_viewer_image_io.py
@@ -112,7 +112,7 @@ def test_add_zarr():
     with TemporaryDirectory(suffix='.zarr') as fout:
         z = zarr.open(fout, 'a', shape=image.shape)
         z[:] = image
-        viewer.open([fout])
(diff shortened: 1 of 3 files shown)


jni:
Yes they assume TZCYX iirc

jni (approved):

PR #1223, opened 2020-05-08
make skimage data fixtures compatible with 0.17.0

tlambert03:
# Description
With v0.17.0, scikit-image no longer bundles `multipage.tif`, which is causing our tests to break.
This fixes tests in a way that will continue to work for older versions of scikit-image.

## Type of change

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/components/_tests/test_viewer_image_io.py
+++ b/napari/components/_tests/test_viewer_image_io.py
@@ -1,7 +1,5 @@
-import os
 import numpy as np
 from dask import array as da
-from skimage.data import data_dir
 from tempfile import TemporaryDirectory
 import pytest
 from napari.components import ViewerModel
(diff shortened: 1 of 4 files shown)


tlambert03:
one more test still failing, and it seems to be due to an unexpected kwarg "rgb": https://cirrus-ci.com/task/6620521733292032?command=test_pyside2#L1024 ... and it only happens in `test_open_labels` when writing the `tiff` suffix

strangely... I can't yet replicate it locally.  @jni any thoughts on what's going on here?

jni:
Heh I was just looking at it. It's bizarre, cos it doesn't even have anything to do with the data dir does it? Let me try locally.

tlambert03:
yeah, I think it's unrelated... very confusing

jni:
Passes fine locally. 🤦

jni:
I tried to run the full test suite in case some tests were interacting, but, still fine...

jni:
minus the zarr error that you fixed in another PR

jni (approved):

jni:
The test failures are due to a new tifffile release basically *as these commits* were coming in (!). The solution was in #1222, so I'm going to merge this and hope for the best! 😬 🤞 

See discussion in Zulip:

https://napari.zulipchat.com/#narrow/stream/212875-general/topic/skimage.200.2E17

jni:
Yep, master is passing! 🎉 

https://cirrus-ci.com/build/4504978817548288

PR #1228, opened 2020-05-08
Add homepage information, move const part from setup.py to setup.cfg

Czaki:
# Description

This add information about napari homepage to package metadata and move most of constant information from setup.py to setup.cfg. In setup.py there should be only things which are calculated.

## Type of change

- [x] Refactor

# References

# How has this been tested?

## Final checklist:

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,3 +1,47 @@
+[metadata]
+name = napari
+description = n-dimensional array viewer in Python.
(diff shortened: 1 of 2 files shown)


Czaki:
It looks like this PR could be closed because files was changed in #1324 ?

tlambert03:
yep, thanks!

PR #1215, opened 2020-05-07
Don't build wheels with releases

jni:
# Description

Wheels break custom setup.py logic from #1059 to prevent installing PySide if PyQt5 is already present. Since napari is pure Python, this is not a big loss.

Closes #1212 

## Type of change

- [ ] Bug-fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# References

# How has this been tested?

- [ ] example: the test suite for my feature covers cases x, y, and z
- [ ] example: all tests pass with my change

## Final checklist:
- [ ] My PR is the minimum possible work for the desired functionality
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #1229, opened 2020-05-08
Better icon-building strategy

### LINKED ISSUE #1209: Icons not loading with napari 0.3.0
## 🐛 Bug

When trying to start napari, the interface loads, but all icons are absent.
grafik

## To Reproduce

`python -m napari`

In the console
```
WARNING: DirectWrite: CreateFontFaceFromHDC() failed (Indicates an error in an input file such as a font file.) for QFontDef(Family="8514oem", pointsize=12, pixelsize=20, styleHint=5, weight=50, stretch=100, hintingPreference=0) LOGFONT("8514oem", lfWidth=0, lfHeight=-20) dpi=144
WARNING: Cannot open file ':/themes/dark/right_arrow.svg', because: No such file or directory
WARNING: Cannot open file ':/themes/dark/right_arrow.svg', because: No such file or directory
WARNING: Cannot open file ':/themes/dark/delete_shape.svg', because: No such file or directory
... and so on for all the other icon files
```

## Expected behavior

The icons should load
## Environment
```
napari: 0.3.0
Qt: 5.12.5
PyQt5: 5.12.3
NumPy: 1.18.1
SciPy: 1.3.1
Dask: 2.13.0
VisPy: 0.6.4

GL version: 4.6.0 NVIDIA 441.22
MAX_TEXTURE_SIZE: 32768

Plugins:
- napari-plugin-engine: 0.1.4
- svg: 0.1.2
```
I got the same issue on another Windows installation, but not on OS X (where I also have an older QT version than officially required by napari)

### LINKED ISSUE #1221: Need consensus on options for delivering icons & resources
## ❓ Question

As shown in #1209, it looks like the strategy of building icons after installation is not going to work in every case.  Even though I tried to put in three levels of "fallback" behavior, in case of permissions errors and stuff, I think it's probably safer to ship a default resources file that is checked into source (or built just prior to uploading our package to PyPI)... 

This brings us back to cross pyside/pyqt resource compatibility and compatibility of resources built with various rcc versions.  I just tested a crapload of combinations of `_qt_resources.py` files built with various versions of `pyside2-rcc` or `pyrcc5`, and tested them on various installs of napari using pyside2 and qt5 installed with both pip and conda.

## tldr;
resources built with the `pyrcc5` binary from `pyqt5` have much broader compatibility than those built with `pyside2-rcc`.  So if we pre-build using the latest `pyrcc5`, we should be able to expect it to be supported on all version of pyside2 and pyqt5 available on either pip or conda.  (that is not the case with `pyside2-rcc`)

<details>
<summary>exact combinations I tested</summary>

### using pip to install pyside2/pyqt5

>PySide2==5.11.2 -> 'Shiboken.ObjectType' object is not iterable
>PySide2==5.12.6 -> 'Shiboken.ObjectType' object is not iterable
>PySide2==5.13.2 -> 'Shiboken.ObjectType' object is not iterable

PySide2==5.14.0 -> **works**
PySide2==5.14.2.1 -> **works**

(PyQt5 < 5.11.3 -> uninstallable on pip due to sip)
PyQt5==5.11.3 -> no icons
PyQt5==5.12.3 -> no icons
PyQt5==5.13.2 -> **works**
PyQt5==5.14.2 -> **works**

### using conda to install pyside2/pyqt5

pyside2=5.12.4.py37h9cca949_0      conda-forge -> no icons
pyside2=5.13.2.py37ha8f7116_2     conda-forge -> no icons
pyqt=5.9.2.py37h2a560b1_4      conda-forge -> no icons
pyqt=5.12.3.py37ha62fc16_3      conda-forge -> no icons

### using pip to install pyside2/pyqt5

PySide2==5.14.2.1 -> **works**
PyQt5==5.11.3 -> **works**
PyQt5==5.12.3 -> **works**
PyQt5==5.13.2 -> **works**
PyQt5==5.14.2 -> **works**

### using conda to install pyside2/pyqt5

pyside2=5.12.4.py37h9cca949_0      conda-forge -> **works**
pyside2=5.13.2.py37ha8f7116_2      conda-forge -> **works**
pyqt=5.9.2.py37h2a560b1_4      conda-forge -> **works**
pyqt=5.12.3.py37ha62fc16_3      conda-forge -> **works**

</details>

### question for @napari/core-devs 
I think we should bundle resources built with `pyrcc5`, rather than building them after install. (though end-users could still be able to rebuild icons using `python -m napari.resources.build_icons`).  How best to do this?
- Just have someone manually build it and check it into source whenever it needs updating?  making sure they are building with `pyrcc5` instead of `pyside2-rcc` (testing for this is possible)... 
- not check into source, and have it built on github actions prior to deployment, by installing `pyqt5` prior to running `setup.py`. This is made complicated by the fact that we actually check to see which backend is installed when writing our requirements.  So we'd basically have to pip install pyqt5, build the icons, uninstall, pip install pyside2, run setup.py... 🙄

tlambert03:
# Description
Fixes #1209 and closes #1221 
The issue in #1209 resulted from a fantastic convergence of a number of different factors!  and can be repeated like this:

1. create conda environment **on windows**
2. install pyqt with `conda install pyqt` (_not pip_.  conda is currently shipping v5.12.3)
3. Then install napari: `pip install napari >=0.3.0`
    - even with the napari wheel gone from pypi, looks like pip now pretty much always builds a wheel, unless the user explicitly asks it not to when installing with pip.  (see [SO post](https://stackoverflow.com/a/37693431/1631624)).  This means that the strategy in #1059 won't work, and _still_ won't be fixed by #1215.  As a result, everyone will [end up with pyside2](https://github.com/napari/napari/issues/1209#issuecomment-625824541) installed from pip, regardless of whether they have pyqt5 installed.  (I'll open another issue for this)
4. when `napari` is launched, it first tries to build icons with the `pyrrc5` binary ... however on windows, that file is actually called `pyrrc5.bat` (unlike with pyside2, which is consistent between platforms).
5. So that fails and then falls back to `pyside2-rcc`, which is now also in the environment because of step 3.
6. That successfully creates a resources file, but it fails to load the icons because the resource file built with pyside2-rcc (5.14.2) is [not backwards compatible with pyqt 5.12](https://github.com/napari/napari/issues/1221), which is what we now have because qtpy will prefer pyqt5 in the environment over pyside!!

# 💻🔨

The fix for icons in this PR is easy... we just look for `pyrrc5.bat` in addition to `pyrrc`... the wheel issue will still need to be adressed.

## Type of change

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/resources/build_icons.py
+++ b/napari/resources/build_icons.py
@@ -6,7 +6,7 @@ for styling svg elements using qss
 import os
 import re
 import shutil
-from subprocess import check_call, CalledProcessError
+from subprocess import check_call
 from typing import Dict, List, Tuple
 
 from ..utils.theme import palettes as _palettes
@@ -167,10 +167,24 @@ def build_pyqt_resources(out_path: str, overwrite: bool = False) -> str:
     )
 
     # then convert it to a python file
-    try:
-        check_call(['pyrcc5', '-o', out_path, qrc_path])
-    except (FileNotFoundError, CalledProcessError):
-        check_call(['pyside2-rcc', '-o', out_path, qrc_path])
+    # this looks a little funny, but rather than pick the exact binary to try
+    # once, we use this try/catch loop because
+    # 1. we always want to use pyrcc5 if it's available, regardless of API
+    # 2. it will sometimes, (if not always) be named pyrcc5.bat on windows...
+    # 3. only then do we try pyside2-rcc
+    # see https://github.com/napari/napari/issues/1221 for background
+    for binary in ('pyrcc5.bat', 'pyrcc5', 'pyside2-rcc'):
+        try:
+            check_call([binary, '-o', out_path, qrc_path])
+        except FileNotFoundError:
+            continue
+        break
+    else:
+        raise FileNotFoundError(
+            "Unable to find an executable to build Qt resources (icons).\n"
+            "Tried: 'pyrcc5.bat', 'pyrcc5', 'pyside2-rcc'.\n"
+            "Please open issue at https://github.com/napari/napari/issues/."
+        )
 
     # make sure we import from qtpy
     with open(out_path, "rt") as fin:


sofroniewn (approved):
Wow amazing detective work @tlambert03 - this solutions looks good to me, and we can discuss #1230 on #1230!

jni (approved):

PR #1232, opened 2020-05-09
Unify user/plugin kwargs.  Use filename for layer name

### LINKED ISSUE #1227: Handling of unexpected layer kwargs from plugins
## 🐛 How to handle mismatched kwargs from plugins?
When we introduced the option to force `viewer.open` to interpret layers as a specific type with the `layer_type` options, we had decided it would be an error if the user specified a specific layer type, but a plugin returned kwargs incompatible with that type.  So for instance:

if the user asks:  `viewer.open('path/to/image.tif', layer_type='labels')` and the plugin returns layer data of `[(data, {'rgb': False}, 'image')]` then we raise a `TypeError` because the `add_labels` method has no keyword `rgb`.  The plugin wasn't wrong, it just couldn't possibly anticipate what the user was going to do.

Yesterday, tifffile released version `2020.5.7`, which includes a tentative napari reader plugin (🎉), and before #1222 merged, it was causing an error in our `test_viewer_labels_io.test_open_labels` for exactly the scenario mentioned above.

1. modify where we apply the `prune_kwargs` logic, such that we simply strip any invalid kwargs returned from the plugins.  in other words "yes, that was a valid kwarg for the layer type you _thought_ you were returning, but the user has requested that we go in a different direction with this"...
2. pass the user-supplied `layer_type` argument to the plugins and make them deal with it (I dislike this option a lot... we shouldn't put that burden on the plugins)

if we go with 1... we might want to _only_ strip invalid kwargs when the user has specified `viewer.open(path, layer_type=...)`, and keep it as an error otherwise (for instance, if the plugin mistakenly returned a key in their `meta` dict that didn't match the `layer_type` that *they* specified.)

tlambert03:
# Description
This adds one of the action items from the last meeting: to use the filename as a fallback name for the layer name when opening a path with `viewer.open`.  Because there were similar issues here to #1227 around "conflicting info between the plugin and the user", this PR also fixes #1227 by cleaning up the process through which we "merge" keyword arguments and layer_type requests provided by the user, with those provided by the plugin.

Note: I also merged in #1219 here due to some overlapping sections... so that PR should be merged first.

## Type of change

# How has this been tested?

- [x] the test suite for my feature covers naming layers from filename, as long as plugin doesn't also provide a name.

## Final checklist:


napari/components/add_layers_mixin.py
```diff
@@ -849,40 +850,28 @@ def _add_layers_with_plugins(
...
+            if len(path_or_paths) == len(layer_data):
+                filenames = iter(path_or_paths)
             else:
-                path_repr = path_or_paths
-            msg = f'No plugin found capable of reading {path_repr}.'
-            logger.error(msg)
-            return []
+                # hard to say what to do if a list of paths have been returned
+                # as a list of layer data without a 1:1 relationship.
+                # here we just use the first name...
```
sofroniewn:
```suggestion
                # if a list of paths has been returned as a list of layer data
                # without a 1:1 relationship between the two lists we just use
                # iterate over the first name
```
Took me a while to parse the above - i think this comment is now more clear, i agree it's an odd case

tlambert03: yup I agree, that's clearer

napari/components/add_layers_mixin.py
```diff
@@ -849,40 +850,28 @@ def _add_layers_with_plugins(
...
+        elif is_sequence(path_or_paths):
+            if len(path_or_paths) == len(layer_data):
+                filenames = iter(path_or_paths)
             else:
-                path_repr = path_or_paths
-            msg = f'No plugin found capable of reading {path_repr}.'
-            logger.error(msg)
-            return []
+                # if a list of paths has been returned as a list of layer data
+                # without a 1:1 relationship between the two lists we just use
```
sofroniewn:
```suggestion
                # without a 1:1 relationship between the two lists we just
```
my bad typo ...

napari/components/add_layers_mixin.py
```diff
@@ -968,6 +957,111 @@ def _add_layer_from_data(
...
+        # were returned by the plugin. We only prune kwargs if the user did
+        # *not* specify the layer_type. This means that if a user specified
+        # both layer_type and additional keyword arguments to viewer.open(),
+        # it is their responsibility to make sure the kwargs match the
+        # layer_type.
+        _meta.update(prune_kwargs(kwargs, _type) if not layer_type else kwargs)
+
+    if not _meta.get('name') and fallback_name:
+        # trim to reasonable length
+        if len(fallback_name) > 16:
```
sofroniewn: Why the trim? Is it because of UI reasons and names not displaying well?
If we are 100% sure about a trim like this would we apply it always? Not just when applying the fallback? I remember @royerloic had liked the trim, but i'm not so sure ...

sofroniewn (approved):
LGTM! Left some minor comments, we can get #1219 merged and then pass this to @jni for review

napari/components/add_layers_mixin.py
tlambert03: all good questions. we should get some opinions.  I put it here only on the fallback, since that was taken from a filename, which tends to be long, and the user never "asked" for this to be the layer name... but I'm happy to change to whatever people want

napari/components/_tests/test_add_layers.py
```diff
@@ -44,14 +39,24 @@ def test_plugin_returns_nothing():
...
+    viewer.open('mock_path.tif')
     assert len(viewer.layers) == 1
+    # The name should be taken from the path name, stripped of extension
+    assert viewer.layers[0].name == 'mock_path'
 
-    viewer.open('mock_path', stack=True)
+    # stack=True also works... and very long names are truncated
+    viewer.open('mock_path_very_long_name.tif', stack=True)
     assert len(viewer.layers) == 2
+    assert viewer.layers[1].name == 'mock_path_very...'
```
jni: I don’t love this at all. It’s not a very long truncation length! I can see many people coming across this limit with their images.

napari/components/add_layers_mixin.py
```diff
@@ -849,41 +850,29 @@ def _add_layers_with_plugins(
...
+                filenames = iter(path_or_paths)
             else:
-                path_repr = path_or_paths
-            msg = f'No plugin found capable of reading {path_repr}.'
-            logger.error(msg)
-            return []
+                # if a list of paths has been returned as a list of layer data
+                # without a 1:1 relationship between the two lists we iterate
+                # over the first name
+                filenames = itertools.repeat(path_or_paths[0])
```
jni, line 864: Wow this wouldn’t have occurred to me to check for... do we really not have a way to map these correctly?

napari/components/add_layers_mixin.py
```diff
@@ -968,6 +957,111 @@ def _add_layer_from_data(
...
+            raise ValueError(
+                "The second item in a LayerData tuple must be a dict"
+            )
+    else:
+        _data.append(dict())
+    if len(_data) > 2:
+        if _data[2] not in layers.NAMES:
+            raise ValueError(
+                "The third item in a LayerData tuple must be one of: "
+                f"{layers.NAMES!r}."
```
jni, line 993: What does !r do btw?

napari/components/add_layers_mixin.py
```diff
@@ -968,6 +957,111 @@ def _add_layer_from_data(
...
+            )
+    else:
+        _data.append(guess_labels(_data[0]))
+    return tuple(_data)  # type: ignore
```
jni, line 997: Hmmm, interesting, how come you have to ignore typing here?

napari/components/add_layers_mixin.py
jni: Yep, long names for me. Editing a file name is just as easy long or short, and long involves no information loss. To clarify, this is the base name and not the full path, right?

jni (commented):
I’d definitely remove the trimming here. Otherwise just minor questions, feel free to ignore them

napari/components/_tests/test_add_layers.py
tlambert03: I'll remove it.  It was the length that fits in the LayerList without truncation... but probably no reason to make special accommodations for that

napari/components/add_layers_mixin.py
tlambert03: yes, basename without extension.  nixed the trimming.

napari/components/add_layers_mixin.py
tlambert03: not really, there's nothing (currently) stopping a plugin from returning a different number of layer_data tuples than there were paths in the list.  probably unlikely, but possible.  in most cases, I suspect it will be a listed of files stacked into a single image...

napari/components/add_layers_mixin.py
tlambert03: same as `{repr(layers.NAMES)}`

napari/components/add_layers_mixin.py
tlambert03: Yeah, that one is frustrating... looks like [mypy is sometimes unable to do type narrowing on tuple lengths](https://github.com/python/mypy/issues/8733)... so it thinks the return type is `Tuple[Any, ...]`

tlambert03:
ok fixed the name trimming... let me know if you want any of the other things changed, or were just curious

napari/components/add_layers_mixin.py
jni: Well, I don't see a way around this, but it's making me deeply uncomfortable, for the record. =P

PR #1239, opened 2020-05-11
add new performance doc in new "explanations" directory

pwinston:
## Description
Adds new file: `docs/explanations/performance.md`. It contains findings from my first week of reading about napari performance related issues and discussing performance with a number of people. It also contains some pretty generic background/philosophy about performance in general.

## Type of change
Documentation only change.

## Explanations
I created a new `docs/explanations` directory for this based on this comment  by @sofroniewn:
https://github.com/napari/napari/issues/764#issuecomment-605077118

The "4 types of documentation" idea is from this divio page:
https://documentation.divio.com/

### DIFF
--- a/docs/explanations/performance.md
+++ b/docs/explanations/performance.md
@@ -20,6 +20,8 @@ application can feel clunky or frustrating if not designed well.
 
 ## Objective Performance
 
+How to keep napari objectively fast:
+
 **Focus On Real Use Cases**
 
 * Focus on cases that matter to lots of people.
@@ -69,25 +71,25 @@ Napari should strive to have these properties:
 
 **Interruptible**
 * Modeless operations are best. They can interrupted by simply performing some
-  other action. For example is imagery is loading in the background, you can
+  other action. For example if imagery is loading in the background you can
   interrupt it just by navigating to somewhere else.
 * Modal operations that disable the UI should have a cancel button when possible
   unless they are very short.
 * The user should never feel “trapped”.
 
 **Progressive**
-* Show intermediate results as they become available, instead of showing nothing
+* Show intermediate results as they become available instead of showing nothing
   until the full result is ready.
 * Sometimes progressive results are better even if they slow things down a bit,
   which is not necessarily intuitive.
 
 **Informative**
 * Clearly show what controls are enabled or disabled.
-* If progressive display is not possible, show progress bars.
-* Show an busy animation as the last resort, never look totally locked up.
+* If progressive display is not possible, show a progress bar.
+* Show a busy animation as the last resort, never look totally locked up.
 * Show time estimates for super long operations.
 * Let power users see timings, bandwidth, FPS, etc.
-* Revealing internal state lets users appreciate why its taking time.
+* Revealing internal state that explains why it's taking time is helpful.
 
 ## Performance Is Never Done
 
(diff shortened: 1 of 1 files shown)


docs/explanations/performance.md
```diff
@@ -0,0 +1,124 @@
+# Napari Performance
+
+Performance is a core feature of napari. Performance is a concern for every
+feature that napari implements. Without adequate performance users will migrate
+to other visualization tools, even if napari has impressive functionality.
```
jni: I would frame this a bit differently. napari has been successful so far because it solves use cases that nobody had adequately solved in Python, even if some of its design is failing to scale so far. So here I would mention more intrinsic characteristics of performance as it relates to visualisation:

- Fundamental use cases such as data exploration depend on performance, because one needs to be able to swiftly find interesting parts of the data. If things are slow, one drowns in noise.
- With visualisation and UI in particular, and in contrast to analysis code, fast, responsive software is joyful to use, while slowdowns are frustrating and makes one want to seek alternatives or simply stop using visualisation altogether, opting instead for summary statistics. This results in loss of information and missed opportunities.

docs/explanations/performance.md
```diff
@@ -0,0 +1,124 @@
...
+
+How to keep napari objectively fast:
+
+**Focus On Real Use Cases**
+
+* Focus on cases that matter to lots of people.
+* It’s easy to waste time optimizing things no one cares about or no one will
+  notice.
+* If a dataset is unreasonable or out of scope or fringe, don’t waste time
+  trying to make it run fast.
```
jni:
```suggestion
* If a dataset is unreasonable or out of scope or fringe, don’t spend too
  many resources trying to make it run fast.
```
Almost every dataset matters to *someone.* =)

docs/explanations/performance.md
```diff
@@ -0,0 +1,124 @@
...
+
+**Always Be Timing**
+* Build timers into the software that always run.
```
jni, line 40: Love this.

docs/explanations/performance.md
```diff
@@ -0,0 +1,124 @@
...
+* If not always visible, power users and developers should be able to toggle them on.
+* This gives people an ambient awareness of how long things take.
+* Allows users to report concrete performance numbers:
+  *  *it seemed slow* &#8594; *it ran at 10Hz*.
+  *  *it took a long time* &#8594; *it took 2 minutes and 30 seconds*.
```
jni:
```suggestion
  *  *it seemed slow* → *it ran at 10Hz*.
  *  *it took a long time* → *it took 2 minutes and 30 seconds*.
```

docs/explanations/performance.md
```diff
@@ -0,0 +1,124 @@
...
+* Teaches users how different hardware impacts performance.
```
jni, line 46: I was surprised to find out the other day that random seek times are two orders of magnitude faster on SSDs vs HDDs. Obviously I knew it was faster but I didn't know it was 100x faster!

The context was testing performance on a dataset on a laptop's SSD, where 121 1MB chunks worked almost equally fast as 1 121MB chunk. But when we applied it to the full data on an external HDD, the performance was unusable!

jni: It might be worth noting this fact here as an example.

docs/explanations/performance.md
```diff
@@ -0,0 +1,124 @@
...
+**Performance Unit Tests**
+* Time one small operation to monitor for regressions.
+* Napari has some of these today as "benchmarks".
+* Interesting to see how different hardware performs as time goes on.
+
+**Run All Tests Every Merge**
+* Save results to a database or somewhere.
+* Catch a regression right when it happens and not weeks or
+  months later.
+* See how new features run on large datasets no one tested.
```
jni, line 63: These are great! Setting up continuous ASV testing would be wonderful.

jni (approved):
@pwinston thanks for this doc! I've made some very minor suggestions but generally it looks great. We've always wanted to set up continuous benchmarking with airspeed velocity, but never gotten around to it. The biggest challenge is that it requires a dedicated machine or a "bare metal" VM on the cloud, which most CI services don't provide (or any existing CI service?).

sofroniewn (approved):
This is great @pwinston, thanks so much for adding!!

docs/explanations/performance.md
pwinston: That intro wasn't in my original doc. I just tacked it on. Your suggestions were good I revised it to try and capture the idea that performance for "interactive" tools even super criticial.

docs/explanations/performance.md
pwinston: I added a line about HDD vs. SSD vs. network file systems.

Yes SSD seek times are so radically different there are algorithms today people run using SSD that aren't sped up versions of HDD algorithms, they are seek-intensive things that never would have been attempted on HDD. So it has opened some doors for having huge datastructures on disk that "hop around in" all over the place.

On the flip side if you do run one of those on HDD will performance so horrendously bad it will seem just broken and unusable.

It's also interesting the most common block storage for AWS system EBS is SSD based, but it's networked! You'd think if it's networked what does it matter if it's SSD or not, but you want/need the fast seek times even over the network.

docs/explanations/performance.md
pwinston: I added mention of ASV. In general the document is pretty aspirational without many concrete plans. Talking with @sofroniewn we kind of felt the order here is:
1. Write this grand performance vision document.
1. Put it aside and dive into obvious big problems like #845 and #204 
1. Come up for air after things are better and survey the performance landscape and prioritize what to do, including automated tests, timing, benchmarking, etc.
1. At that point maybe this doc can point to lots of things we are actually doing to support these goals.

The point being #845 is so glaring it's worth tackling first.

sofroniewn:
Thanks for making those changes - merging now. When we get further in the website reorganization we can see if this is the best home for our "explanations" or if we want to keep them in a different home

PR #1242, opened 2020-05-12
Allow just backspace to delete a layer

jni:
# Description

I often find that I want to delete a layer with just backspace. Since that key is not used for anything else, it might as well be used here?

## Type of change

# References

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/_viewer_key_bindings.py
+++ b/napari/_viewer_key_bindings.py
@@ -74,6 +74,7 @@ def focus_axes_down(viewer):
     viewer.window.qt_viewer.dims.focus_down()
 
 
+@Viewer.bind_key('Backspace')
 @Viewer.bind_key('Control-Backspace')
 @Viewer.bind_key('Control-Delete')
 def remove_selected(viewer):


sofroniewn:
Backspace is actually used in the points layer to delete selected points - see https://github.com/napari/napari/blob/master/napari/layers/points/_points_key_bindings.py#L62

We don't really have much of a rule anymore for what types keybindings live on layers (say modified/ unmodified) vs viewer.

I'm open to any changes, we should just try and be consistent

jni:
> Backspace is actually used in the points layer to delete selected points

That's great, but can the shortcut be dependent on what part of the UI is selected? ie, backspace deletes what I have just selected, be it a point, a shape, or a layer.

jni:
> I'm open to any changes, we should just try and be consistent

Well, I just added a key to the existing binding. Hard to be more consistent than that. =P

sofroniewn:
> That's great, but can the shortcut be dependent on what part of the UI is selected? ie, backspace deletes what I have just selected, be it a point, a shape, or a layer.

We could do that, but right now I think the keybindings are additive, so we'd need to make some changes - otherwise every time you deleted some points you'd also delete the whole layer

jni:
> otherwise every time you deleted some points you'd also delete the whole layer

Might save users some time??? :joy: 

Ok fair enough. I'll try to rework this a bit. Would be great to get some advice from @kne42 when she comes back about the best way to achieve this — or if it's even feasible with the current framework.

cwood1967:
If someone is changing the name of a layer, they will often need to backspace to to fix something. It might be too easy to hit tab or something to take you out of edit mode and have the layer deleted when that is not desired.

sofroniewn:
@jni i think we can close this now since we are moving to user configurable shortcuts, at which point you will be able to customize this as you wish!!

Issue #1230, opened 2020-05-08
pip will still install pyside2 even without wheel on pypi


tlambert03:
## 🐛 Bug
as mentioned in #1229, it seems that pip pretty much always builds a wheel before installing (even if it downloads source), unless the user specifies `--no-cache-dir` ... which means that #1059 will not prevent pyside2 from being installed, even after #1215 .  If you `pip install napari>=0.3.0`, you will probably see it still build a wheel for caching.

You can repeat this locally:
```
# in napari repo
python setup.py sdist

# in clean environment
pip install path/to/napari/dist/napari-*.tar.gz 
...
Building wheels for collected packages: napari
  Building wheel for napari (PEP 517) ... done
  Created wheel for napari: filename=napari-0.3.0+6.gabe2afd6-py3-none-any.whl...
  Stored in directory: ~/Library/Caches/pip/wheels/...
Successfully built napari
...
# and only then does it build
Installing collected packages: ... napari

sofroniewn:
I think @jni noticed this when testing the release candidate, but we weren't sure what was going on - this probably explains it

tlambert03:
I asked a question about this on the [python packaging forum](https://discuss.python.org/t/conditional-package-install-depending-on-other-packages-in-environment/4140).  Got a very helpful answer there.  Long and short of it is that (as far as the responder knew) there's currently no good way to achieve what we want to do, and he suggested that living with `pip install napari[pyside2]`was probably our best option at this point.

He did provide useful links to two other very similar discussions that are going on.  [This one](https://github.com/pypa/setuptools/issues/1139) is particularly similar to what we want... it was mentioned by @ttung back in #704, but bears another look if you missed it the first time.

Czaki:
Temporary fix is in #1246 (which disable isolated build), but until https://github.com/pypa/setuptools/issues/1139 is fixed there is no option for default extras.

Czaki:
Maybe napari should define `all` extras which install all optional/default depedencies. Then if there is some other problem which implementation use then install command will not change.

jni:
This was closed by #1324

PR #1246, opened 2020-05-12
exclude pyproject.toml from sdist

Czaki:
# Description

Enable #1059 by removing `pyproject.toml` from source bundle
#1215 is not enough because `pyproject.toml` introduce build isolation 

## Type of change

# References

https://github.com/napari/tutorials/pull/61#discussion_r400777453

# How has this been tested?

- [x] tested on my machine
- [x] add proper test task

## Final checklist:

@sofroniewn @jni

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -194,6 +194,27 @@ win_task:
 
   test_script: pytest -v -o qt_no_exception_capture=1
 
+pyqt_task:
+  auto_cancellation: true
+
+  name: pyqt_test_instalation
+
+  container:
+    image: python:3.7-slim
+
+  install_script:
+    - apt-get update
+    - apt-get install -y build-essential
+    - pip install --retries 3 -U pip setuptools
+    - pip install --retries 3 -r requirements/development.txt
+    - python setup.py sdist
+    - pip install PyQt5 numpy
+    - pip install dist/napari*.tar.gz
+
+  test_script:
+    - python -c "ff=open('test.py', 'w');ff.write('import sys\ntry:\n\timport PySide2\n\tsys.exit(1)\nexcept ImportError:\n\tpass\n');ff.close()"
+    - python test.py
+
 codecov_task:
   auto_cancellation: true
   name: Coverage & Requirements
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -5,3 +5,4 @@ recursive-include napari/resources *.png *.jpeg *.jpg *.svg *.qss
 exclude napari/resources/_qt_resources*.py
 include napari/utils/colormaps/matplotlib_cmaps.txt
 include requirements/default.txt
+exclude pyproject.toml


Czaki:
Ready

tlambert03:
I'd like to propose that we're generally going down the wrong path here...

In an attempt to avoid having our users run `pip install napari[pyside2]` or `pip install napari[pyqt5]`, we are jumping through a *lot* of hoops, and breaking a lot of standard practices.  The truth of the matter is that pip and setuptools [don't really support the use case we're after here](https://discuss.python.org/t/conditional-package-install-depending-on-other-packages-in-environment/4140). So we've introduced a few non-standard tricks like [adding logic to `setup.py`](https://github.com/napari/napari/pull/1059), and then we [avoided distributing wheels](https://github.com/napari/napari/pull/1215), but found that [that still doesn't work](https://github.com/napari/napari/issues/1230), and so now we're trying another hack (removing `pyproject.toml`) to see if we can force pip (which has [moved on to a new default behavior and standard](https://pip.pypa.io/en/stable/reference/pip/#pep-517-and-518-support)) to do for us something that it does not do.

Meanwhile, what we have now is a situation in which we very often have both pyside and pyqt installed in the environment, which has created challenges (e.g. for  resource building) and is likely to lead to other bugs... and I don't think avoiding isolated builds and the new PEP517/518 direction of things is going to be our best long-term solution.

I would propose that we just bite the bullet, promote `pip install napari[pyside2]` or `pip install napari[pyqt5]`, add a bomb-proof error message if neither backend is available, start building wheels, and simplify our setup.py logic again.  And if anything is done with pyproject.toml, we should expand it to [make explicit our build-backend](https://github.com/pypa/setuptools/issues/1698)(`build-backend = "setuptools.build-meta"`)

tlambert03:
This issue is largely due to the "split personality" that napari has as a library vs an application:
as an application, we want end-users to be able to "just install napari" and have things as we intended them, which often requires making rather specific dependency declarations.   As a library, we'd like to allow end-users some flexibility in the way they use napari, which necessitates loose dependency declarations.  It's really quite hard to have it both ways.

I'll also reraise the possibility that has been [mentioned in the past](https://github.com/napari/napari/issues/704#issuecomment-572900776), which is to have `napari` be a shell package that requires `napari-lib` and `pyside2` ... and tell those who don't want pyside2 to `pip install napari-lib` ...

sofroniewn:
One approach we could go with is

```
pip install napari[pyside2]
pip install napari[pyqt5]
pip install napari[all]
```

Where pip install napari[all] also pulls down `pyside2`, but also things like `napari-svg`, potentially `napari-console` if we pulled that code out, `napari-builtins` if we put all our builtin code somewhere else.

I don't think it is intimidating or confusing to new users to say

```
pip install napari[all]
```

it doesn't require them to understand the subtleties of our backend, and as new users they can feel confident they are getting everything.

Thoughts @jni @tlambert03?

Czaki:
I totally agree. This was previously rejected, but it is best solution until some default extras was introduced to python package managment.

jni:
I agree 100% with Talley and was about to propose the same thing. So I guess we have consensus!

@Czaki I'll close this PR since it has nothing in common with the proposed strategy we're going for in the end. Happy we found a solution we're all happy with!

PR #1248, opened 2020-05-13
Shapes colors refactor

kevinyamauchi:
# Description
This PR updates the representation of shape colors to RGBA arrays. Also, it updates the layer controls to include the color picker widget for the edge and face colors.

## Type of change

# References
Closes #898 

# How has this been tested?

## Final checklist:


kevinyamauchi:
As implemented, this PR stores the color values (1x4 numpy array) in the `Shape` object for each shape in the layer, as that is how color is currently stored. However, it would be nice to store all colors in a single Nx4 for each property where N is the number of shapes. 

Currently, the model data is stored in a couple of places:

- `ShapesList`: the shapes list contains a list of `Shape` objects (one for each shape) as well as some associated helper methods. Each `Shape` object contains the data and display properties (e.g., face and edge color) associated with a particular shape.
- `Mesh`: a thin class that groups the properties related to the mesh elements (e.g., coordinates, colors, in view). The `Mesh` object is stored in the `ShapesList`.

As discussed in #898 (see [this comment](https://github.com/napari/napari/pull/898#discussion_r389391645)), it probably makes sense to store the display properties in the `ShapesList` and it should be removed from the individual `Shape` objects (to avoid the data being stored in multiple locations). There was also some discussion of giving the `Shape` object access to the display property data (e.g., face and edge colors), but I don't think that is necessary. In this case, the `Shape` objects become classes to store validate and store the coordinate data. As far as I can tell, the individual shape models do not make use of the face and edge colors internally and only store those data. Is that correct, @sofroniew?

I think it is out of scope of this PR, but I think we should consider eliminating the `Mesh` class. It seems like all of the data stored in `Mesh`is public, so it seems like we could just attach the `Mesh` properties and methods to the `ShapesList` class (most of the properties are prepended with "mesh" already anyway).

kevinyamauchi:
Just to capture the thoughts from the dev sync today:

- We will pull `edge_color` and `face_color` from the `Shape` models in the the `ShapesList`
- We will eliminate the `opacity` property and use the alpha value in the RGBA color array
- In the future, we will consider integrating `Mesh` into `ShapesList`

Following this work, I will make a follow-up PR to add properties and setting colors by properties.

kevinyamauchi:
I have added a first pass at the core functionality of this PR. In particular, I:

- removed colors from the individual `Shape` models and face/edge color are now stored Nx4 RGBA arrays in `ShapesList._*_color`.
- removed opacity from the individual `Shape` models. Opacity on the per-shape level can now be set using the alpha value in the RGBA color array (dim 3). The opacity slider now controls the layer opacity, as is done in the other layers
- updated the UI to use @tlambert03 's nice color picker GUI for setting edge/face colors.

@napari/core-devs , it would be great to get some thoughts on the best way to expose this to the user. Thus far, I have tried to mirror how the shapes layer works as much as possible, but I think this is a good opportunity to bring the behavior more in line with the `Points` and `Vectors` layers, especially given the plan to add properties next. 

Current ways colors are set:

1. The user can specify colors when adding shapes (initializing the layer or the `Shapes.add()` method) using standard string names, RGB arrays, and RGBA arrays.
2. The user can modify the colors by selecting the shapes they wish to modify and using the `Shapes.current_*_color` property.
3. The user _cannot_ directly edit colors via the `Shapes.*color` property.

To bring `Shapes` more in line with `Points` and `Vectors`, I propose we add the ability to directly modify colors via the `Shapes.*_color` property. This is useful for globally setting a color across the layer, can be an easier way to directly modify colors, and will be necessary when adding properties.

### SVG writer
Since we are removing the opacity and face color from the shape model, I think we will need to update the svg writer plugin in parallel. What is the best way to approach doing that? Should we make a separate PR to pin the version of the SVG writer plugin to the current version, update the SVG writer plugin, release the SVG writer plugin, then update the requirements in this PR?

sofroniewn:
> To bring Shapes more in line with Points and Vectors, I propose we add the ability to directly modify colors via the Shapes.*_color property.

This sounds good to me!

> Should we make a separate PR to pin the version of the SVG writer plugin to the current version, update the SVG writer plugin, release the SVG writer plugin, then update the requirements in this PR?

Yikes! Sounds like that is the best way, if you can kick this off that would be great!!

kevinyamauchi:
> Yikes! Sounds like that is the best way, if you can kick this off that would be great!!

Haha yeah, it's a bit convoluted, but hopefully we won't have to make changes like this too often. At least it encourages us prioritize stability :P

I will get going on that soon.

kevinyamauchi:
I have now added the ability to directly set the shapes colors via `Shapes.*_color`. I think the PR is now feature complete. From here, I think the followup PR to add properties and setting colors via properties should be fairly straightforward.

As soon as napari-svg 0.1.3 goes up, I can bump the requirements here and the tests should pass (they pass locally for me).

sofroniewn (approved):
Code look good, functionality works great. All tests pass locally for me. I've just released `napari-svg` 0.1.3 and have requested all the tests to re-run. Once they all pass I think this is good to merge.

Next steps as a PR that adds support for properties and makes the API there match the points layer sounds good!!

kevinyamauchi:
Thanks, @sofroniewn ! I am going to add a few tests and then I think it will be good to go!

kevinyamauchi:
I have finished adding tests for the new color stuff. I will let it sit and then if there aren't further comments merge tomorrow.

sofroniewn:
@kevinyamauchi I just updated with a fix of the tests from master, want to see if all the tests now pass (we were having pyside2 issues)

kevinyamauchi:
@sofroniewn , I see the 0.3.3 PR. Should I wait to merge this until that lands?

sofroniewn:
I think we can merge, I don't think @jni needs to review, it's a pretty clear refactor and it's sat for a day. Having it in will also make my life easier (currently my local stuff is a bit confused about the different napari svg versions!).

@kevinyamauchi do you want to give it one final check, make sure my updating from master is fine and then merge

kevinyamauchi:
Looks good to me. Merging!

PR #1249, opened 2020-05-13
Add pooch to requirements/test.txt

Czaki:
# Description

bugfix after release of scikit-image 0.17.2

## Type of change

- [x] Bug-fix (non-breaking change which fixes an issue)

# References

pooch is optional dependecy for scikit-image 

https://github.com/scikit-image/scikit-image/blob/master/doc/release/release_0.17.rst
# How has this been tested?

- [x] example: all tests pass with my change

## Final checklist:
- [x] My PR is the minimum possible work for the desired functionality

PR #1250, opened 2020-05-14
remove scipy.stats import

### LINKED ISSUE #925: working on fixing tests

tlambert03:
# Description
closes #925 

Just tried to repeat the bug that I described in #925 and I wasn't able to, even without this mysterious import (on py 3.7 with qt5, etc... as described there).  So I think whatever that was has been solved.  We've changed a number of things about widget cleanup, console creation, etc... so I'm not terribly surprised!

### DIFF
--- a/napari/__init__.py
+++ b/napari/__init__.py
@@ -69,13 +69,5 @@ from .utils import sys_info, _magicgui
 # register napari object types with magicgui if it is installed
 _magicgui.register_types_with_magicgui()
 
-
-# this unused import is here to fix a very strange bug.
-# there is some mysterious magical goodness in scipy stats that needs
-# to be imported early.
-# see: https://github.com/napari/napari/issues/925
-from scipy import stats  # noqa: F401
-
 del _magicgui
-del stats
 del _viewer_key_bindings


sofroniewn:
Thanks, looks like this needs #1249 for tests to pass. Can you update and then once tests all pass i'll approve and merge

jni (approved):

PR #1260, opened 2020-05-18
Anti-aliasing on splash screen logo

tlambert03:
# Description
just adds anti-aliasing to the splash screen
with (left) and without:
Combined Stacks


sofroniewn (approved):
LGTM!

kevinyamauchi (approved):
So smooth!

PR #1261, opened 2020-05-18
Prefer rcc binary at front of path

### LINKED ISSUE #1254: Issue with using wrong pyrcc
## 🐛 Bug

I updated `napari` to post- #1229 and `napari` started failing when trying to build icons on Windows after importing napari:
https://github.com/napari/napari/blob/96e9722419a5b0adddbc14fe12c69a7d769f8758/napari/resources/build_icons.py#L178

## Cause

This may be a non-issue for others, but became an issue for me. Basically, each of these binaries were called:
`'pyrcc5.bat', 'pyrcc5', 'pyside2-rcc'`

and my vanilla python virtual environment didn't have `pyrcc5.bat`, but my base anaconda install was on path (bottom of list), and was the only source of `pyrcc5.bat`. This then failed, because it was found, but didn't have `pyqt5-sip` I believe. Rearranging so that `pyrcc5` was tried first fixed the issue for me.

This is a weird bug because I happened to have an Anaconda install on my path that happened to have `pyrcc5.bat`, and is kind of my fault. I don't know if it makes sense for you to rearrange the binaries to try, or ensure it's coming from the same environment. At least this is an explanation for anyone else getting a similar error:

```
subprocess.CalledProcessError: Command '['pyrcc5.bat', '-o', 'XXX\\lib\\site-packages\\napari\\resources\\_qt_resources_0.3.1_pyqt5_5.14.0.py', 'XXX\\lib\\site-packages\\napari\\resources\\res.qrc']' returned non-zero exit status 1.
```

@tlambert03

tlambert03:
# Description
Hopefully fixes #1254 
This PR uses `shutil.which` to look for `pyrcc5` and then `pyside2-rcc`.  `shutil.which` will find both `pyrcc5.bat` and `pyrcc5.exe` on windows, and will also prefer executables at the top of the PATH, so should prefer binaries in the immediate environment.

@hectormz @volkerH, if you could test, that would be great.

## Type of change

# How has this been tested?

- [x] manually on my windows and mac computers

## Final checklist:


hectormz:
@tlambert03 thanks for the fix!

napari/resources/build_icons.py
```diff
@@ -171,14 +171,16 @@ def build_pyqt_resources(out_path: str, overwrite: bool = False) -> str:
...
+    #    but shutil.which() will find that too
+    # 3. We also want to prefer binaries higher up on the path
+    # 4. only then do we try pyside2-rcc
+    # see https://github.com/napari/napari/issues/1221
+    # and https://github.com/napari/napari/issues/1254
+    for bin_name in ('pyrcc5', 'pyside2-rcc'):
+        rcc_binary = shutil.which(bin_name)
+        if rcc_binary:
+            check_call([rcc_binary, '-o', out_path, qrc_path])
+            break
```
jni: @tlambert03 This still seems risky. It's possible to call napari with `/path/to/conda/env/bin/python -m napari`, in which case the `shutil.which` will still, I believe, find `pyrcc` from a different environment. Why not check for pyrcc in the same environment as `sys.executable`?

tlambert03: good point, hadn't thought of that.
but when you say the "same environment as `sys.executable`" what exactly do you mean by "environment"?  do you mean "directory"?  or do you mean "figure out what environment that python executable is in"?  If the former, I guess maybe I'm (perhaps unreasonably) afraid of restricting it to a single directory like that.   Also, I think where these things live is perhaps platform specific, on windows they are not in the same directory: 

```shell
(napdev) C:\Users\tjl10\python\napari> where pyside2-rcc
C:\Users\tjl10\miniconda3\envs\napdev\Scripts\pyside2-rcc.exe

(napdev) C:\Users\tjl10\python\napari> where python
C:\Users\tjl10\miniconda3\envs\napdev\python.exe
```

if you mean the latter... then how do we figure out what environment that python executable is "supposed" to be in without making major assumptions about the environment manager and platform?

One possibility here is to keep using `shutil.which` but provide it with a `path` argument, based on `os.getenv('PATH')` ... but with various additional folders (like `sys.executable`, and maybe also that `Scripts` folder for windows) prepended?

jni: Wow, that's highly different from *nix envs! Yikes.

```
jni@jni-x1c6 Thu May 21 08:59 
 ~ 
 $ which pyside2-rcc
/home/jni/miniconda3/envs/all/bin/pyside2-rcc

jni@jni-x1c6 Thu May 21 08:59 
 ~ 
 $ which python
/home/jni/miniconda3/envs/all/bin/python
```

And of course there's no accounting for non-conda envs/Pythons...

My suggestion so far is:

- yes, as you say, try shutil.which(path=os.path.dirname(sys.executable)) on *nix, shutil.which(os.path.join(os.path.dirname, r'\Scripts')) on Windows
- if that fails, try your current strategy

That should cover a fair few use cases I think...

tlambert03: > Wow, that's highly different from *nix envs! Yikes.

yeah... why can't they all just agree to do the same thing?? 😤

> suggestion so far...

I was thinking combine all that into a single step:

```python
path = [os.path.dirname(sys.executable)] + os.getenv("PATH")
# maybe add another path or to on windows
shutil.which(bin_name, path=path)
```

jni: oh I was totally unaware that you could pass *multiple* paths. In that case, go for it, that's awesome! And yes you should add the Scripts dir on windows!

jni: `os.getenv("PATH")` probably needs splitting on `':'`?

jni: actually, no, on `os.pathsep`. :grimacing:

tlambert03: yeah, I'll make sure to look at the actual values on all platforms when I test.

jni (approved):
Thanks @tlambert03 this looks great!!

napari/resources/build_icons.py
```diff
@@ -132,6 +133,46 @@ def build_resources_qrc(
...
+    -------
+    path : str
+        Path to the located rcc binary, or None if not found
+
+    Raises
+    ------
+    FileNotFoundError
+        If no executable can be found.
+    """
+    python_dir = os.path.dirname(sys.executable)
```
jni, line 159: @tlambert03 re #1336, should this have been `os.path.abspath(os.path.join(os.path.dirname(sys.executable), '..'))`

jni: No, it shouldn't, never mind. :thinking:

PR #1262, opened 2020-05-18
Performance monitoring

pwinston:
## Type of change

## Description
This PR adds performance monitoring. We created performance monitoring to help in the investigation of #845, #1300 and #1320. However we want performance monitoring to be a permanent part of napari. There will always be performance issues cropping up that need to be investigated.

Care was taken so the performance monitoring code has zero impact on anything unless it is enabled with this `NAPARI_PERFMON` environment variable. We don't want regular users to take any sort of performance hit from the monitoring code. Although note that even when enabled performance monitoring is *much* lighter weight than comprehensive profiling.

The performance monitoring is based on timers than have a name and a duration. Timing data comes from 3 sources:

1. Our special `QApplication` that times all **Qt Events**.
1. Our`perf_timer` context object to time specific blocks of code.
1. Our `perf_func` decorator to time a specific functions.

All 3 types of timers call into `perf_counter_ns()` which itself takes about 100ns to run. However there are other costs to timers including writing the trace file.

The timing data is displayed in two ways:

1. Our QtPerformance widget:

perf-window

Right now this only prints "slow events". It does not give a graphical picture of anything or show the hierarchy of events.

2. From the debug menu you can record "trace files" which can be viewed in Chrome using the special `chrome://tracing` URL. This shows **every** timing event that was recorded in a nice graphical view that you can pan/zoom.

Create a trace file from the Debug Menu:

trace-menu

Drag-n-drop the trace file into `chrome://tracing` to view:

napari-long-paint

## How To Use

Typical investigation of a performance issue might go like this:

1. Run napari with `NAPARI_PERFMON=1`. This works from command line or from IPython/Jupyter.
2. View slow events in the QtPerformance widget.
3. Use the Debug menu to create a trace file, view it with `chrome://tracing`.
4. To dig in deeper add more timers by hand using our `perf_timer` and `perf_func` utilities. 

In theory `perf_timer` and `perf_func` calls can be kept in permanently. They should be 100% disabled and take no time if `NAPARI_PERFMON` is off. However I have not started doing that with this PR. I've been adding timers when exploring but then reverting them.

Note you *can* add perf timers to `vispy`. Just import  `napari.utils.perf as perf` it works even though it seems circular. It can be extremely helpful to see `napari` and `vispy` perf times combined together in one display.

# Trace Files

Chrome's `chrome://tracing` is a very nice way to view trace files. I used the `WASD` keys along with the mouse. The trace file format is JSON. Their file format has a lot features we aren't using yet.

You can also view the same trace files in:
* https://www.speedscope.app/
* https://doc.qt.io/qtcreator/creator-qml-performance-monitor.html

Speedscope can show a [flamegraph](http://www.brendangregg.com/flamegraphs.html) view that Chrome cannot. Speedscope calls the view "Left Heavy" instead of "flamegraph", but it's the same thing.

With flamegraphs the X axis is not time, it's more like "blame". If while timer A is running we make 50 separate calls into timer B, the flamegraph will show that A called B once, but the duration of the call will be the sum of those 50 times we called B. The benefit is it shows how expensive B was in total, which may not be clear in the view showing calls over time.

## Terminology
**Performance monitoring** is meant to be much lighter weight than profiling. Ideally performance timers can be left on at all time, or at least can be toggled on easily like we are doing. Profiling is great because it can time every single function at once, the downside is profiling can sometimes make interactive programs unusuably slow.

Both performance monitoring and profiling have their place.

## Overhead

Performance monitoring can slow things down. This is a form of the "observer effect". The root `perf_counter_ns()` call only takes around 100ns but other parts are probably a lot slower:

1. Updating the `Stat` objects in the `PerfTimers` dict.
1. Writing the JSON file when tracing is active.
1. Updating the QtPerformance widget.

We have not explored the overhead of timing yet in detail. That's one reason we are leaving it opt-in for now. Anecdotally the overhead seems low.

## How has this been tested?
- [x] Used a lot by hand to investigate performance issues.

## Final checklist:


tlambert03:
so cool!!  Just tried it out.  Works great so far.  excited to see where this goes!

sofroniewn:
Nice @pwinston just gave this a spin, works great! Lots that could be thought about in terms of feature addition to the performance widget - I think this will be great functionality that our plugin developers might want to leverage as well as they work to build their plugins. Here's a gif of me drawing on a 2048 x 2048 labels layer, where there is still quite some lag:

profiling

As to `QtCreator` and `QML` I'm not aware of us explicitly using that functionality

sofroniewn:
Also, when I go to `chrome://tracing` I don't see anything even if I have set the `NAPARI_PERFMON_TRACE_PATH` environment variable. Do I have to browse and load in a particular file or does the file get written out automatically?

I guess if `Qt Creator` can render these files and there is a simple Qt widget we can display inside napari as a dock widget with the tracing data that would be my preference even if the functionality is slightly more limited as it will be more accessible to plugin developers who want to access this information. A power user could maybe still open up the file in one of these other tools right?

pwinston:
Right now you have to manually load the trace file into Chrome. In Chrome go to the URL chrome://tracing. Drag-n-drop the trace file into that window, or hit "load" and navigate to the file.

I know of 4 things that can view these trace files:
1) Chrome using the URL chrome://tracing
2) Catapult is the library version of chrome://tracing
3) https://www.speedscope.app/
4) Qt Creator

I agree it would be slick to "embed" a viewer into napari itself. My guess is that might be a big undertaking though. And I don't think Qt Creator lets you do that, but not positive. Catapult is designed to be embedded though.
https://doc.qt.io/qtcreator/creator-ctf-visualizer.html
https://chromium.googlesource.com/catapult/+/HEAD/tracing/README.md

While it would be slick to have it embedded, using an external tool only takes a few seconds, so I think embedding is at best a nice-to-have. Some things that might be higher priority than embedding:

1) Have a way to create trace files from the GUI. Avoid setting the 2nd env var. Either have a button on the QtPerformance window, or have an always available menu option so you don't even need to set the 1st env var to make a trace.

2) Improve the QtPerformance window. It would be ideal if most of the time that window tells you all you need to know, so you don't even need to see a full trace. The window today is totally placeholder, I just wanted to prove the GUI can access the timers.

3) More timers and tracing features. Today we just record Qt Event durations. Adding napari-specific timers of things would yield richer traces. And there are many trace features we aren't using yet like threading, "instant events", "async events", "flow events", "object events". 

For all improvements though I think the key is what's actually useful to a developer trying to understand or improve performance. They don't necessary need full details of everything under the sun. I think the key is what we ourselves felt was needed/useful, and figure plugin developers will want the same things.

pwinston:
I added a new `perf_timer()` context object so you time arbitrary blocks of code. The block's timer will be placed into the same hierarchy as the Qt Event timers. So this:

```
            with perf_timer("transpose"):
                image = np.transpose(
                    np.asarray(self.data[level][tuple(indices)]), order
                )
```

Will look like this:

new-trace

You might ask is using a profiler easier than hand placing timers? I think both approaches have merits. There are trade-offs to each. Good to have both tools in the toolbox. I'm not sure why there are two `set_view_slice()` calls during one paint yet.

sofroniewn:
> I'm not sure why there are two set_view_slice() calls during one paint yet.

I fear at one point there may have been more :-( some of the refactoring in #1040 that @kne42 is working on will hopefully prevent some of the duplicate calls that we have

tlambert03:
I'm reminded of https://github.com/napari/napari/issues/625 and this image... maybe useful still?

Screen Shot 2019-10-25 at 4 59 20 PM

pwinston:
Events are great in a lot of ways, but you do have to watch out for spaghetti-like complexity. The irony is the code looks simple. You are not explicitly calling into other code willy nilly. But the emergent runtime behavior can become complicated and inefficient. 

I've seen systems that have "pause events" type context objects. Like "I'm going to make a bunch of changes, don't notify anyone as I go, notify only when I'm done". Not sure that applies here but just mentioning it.

If we can implement async loading then duplicate events are not all that bad, each one will happen very quickly. Still I can imagine future issues related to tracking down and streamlining event handling.

Not part of this PR certainly which is just the perf tools themselves.

jni:
> I've seen systems that have "pause events" type context objects. Like "I'm going to make a bunch of changes, don't notify anyone as I go, notify only when I'm done". Not sure that applies here but just mentioning it.

We have some of those peppered around the code base, but use is inconsistent.

> Still I can imagine future issues related to tracking down and streamlining event handling.

Yeah it's clearly the Wrong thing to do, even if harmless. I think #1040 should be able to fix it.

@pwinston this is super cool and it's so great that we're already finding all kinds of spots that need optimising. The only question I have so far is why it's necessary to replace QApplication with QtApplication in the GUI context manager? I didn't follow that logic to its source, it just makes me a bit nervous since starting with a QApplication is pretty standard.

tlambert03:
> why it's necessary to replace QApplication with QtApplication in the GUI context manager

His new `QtApplication` is just an alias for `QApplication` if the `NAPARI_PERFMON` environment variable is not set... But if it is set, he needs to intercept the [`QApplication.notify`](https://doc.qt.io/qt-5/qcoreapplication.html#notify) method to spy on (and time) events.  I like it!

pwinston:
@jni I found the idea to override QApplication in order to time Qt Events here:
https://stackoverflow.com/questions/15097861/qt-measure-event-handling-time

It seemed to work well so I didn't pursue any other options. As @tlambert03 says if the environment variable is not defined we then our `QtApplication` is the stock `QApplication`.

Side note: I first called the final class `NapariApplication` but I changed it to `QtApplication` just based on how we name a lot of the GUI classes. But `NapariApplication` might be clearer. So `NapariApplication` is either `QApplication` or `PerfmonApplication`.

(I just changed the `PerfmonApplication` name to `TimedApplication`, I like the name slightly better.)

Long term we might want performance monitoring to be more of a run time toggle. So people can flick it on at anytime. But for now I'm trying to make sure if the env var is off there's zero perfmon code running. Having your perf code accidentally slowing things down is a common pitfall!

Although `perf_counter_ns()` takes only around 100ns. So even 1000 calls per frame is not horrible. Although there's other code too. I'm not sure how many events per frame is typical but I think dozens only.

In multiple C++ projects I've been on we had 3 builds: debug, release with timers, and then the shipping build with no timers (and potentially no asserts, and other final optimizations).

jni:
> It seemed to work well so I didn't pursue any other options. As @tlambert03 says if the environment variable is not defined we then our QtApplication is the stock QApplication

I kinda would prefer it if we put the switch between QApplication and (say) QApplicationWithPerfmon within `napari.gui_qt` directly. What do you think @pwinston @tlambert03? My principle in coding is that I want to open as few files as possible to understand what's going on.

pwinston:
I kind of like that because it's more overt. At the higher level you pick one or other. I hid it so `import QtApplication` would choose under the hood. But I think maybe explicit is better here.

Although to me this is putting it into more files not less. As in `qt_application.py` stays around but probably gets renamed to `perfmon_application.py`? But there's now PERFMON logic `event_loop.py` where it wasn't before?

The comments say `gui_qt` is not used for IPython? Does putting the perfmon logic in `gui_qt` cover all possible ways to run napari or only some of them?

sofroniewn:
> The comments say gui_qt is not used for IPython? Does putting the perfmon logic gui_qt cover all possible ways to run napari or only some of them?

That is correct, we don't use `gui_qt` in IPython or jupyter notebooks, which maybe means we shouldn't put the perfmon logic there, or maybe it is ok if we say that we'll only do monitoring when launching from a script. Once launched I don't know how much is different after the different launch methods. Maybe @jni can comment

tlambert03:
@pwinston, unless you can monkeypatch the `QApplication.notify` method _after_ application instantiation, then it's probably going to be very hard to do anything with IPython or jupyter.  (there is always only one global QApplication, and if some other application has already started it, we don't own it).  Since performance monitoring with perfmon is a "sometimes treat"... I think it's plenty good to start out with the assumption that we are starting the app (i.e. excluding the IPython case).  (or, as mentioned, see if you can patch the already instantiated app?)

pwinston:
I think it's fine to only support `gui_qt` for now, maybe forever. I have a large update to this PR coming soon I'll include moving the perfmon decision point into `event_loop.py` where `gui_qt` is defined.

pwinston:
I re-wrote the main PR description to reflect these most recent updates. Mainly added the Debug menu with ability to record a trace file, and put all the perf files in their own `utils/perf` folder.

pwinston:
I rewrote the PR description again to make if fully up to date. This PR is ready for review. In theory performance code does not have to quite be at the level normal code, since it's opt-in and will not be run by most people. Still there should be standards even for opt-in code.

There are no tests yet. Not sure what's worth testing exactly. Certainly there's things we could do, open to suggestions.

Some of the PR description seems like it might need to be turned into permanent documention. Not sure if this would be an explanations-style markdown file or RST documentation.

sofroniewn (approved):
This is beautiful @pwinston - lots of fun to play with both with your widget and the tracing view in chrome!

perf_mon

I think I have a good understanding of how the code uses to QApplication notify method to do the timing, and everything is clearly documented. For me this is good to merge.

@tlambert03 do you want to take one final look before merge. Juan said this was good to go in while he was out, and it's something that as it is off by default and mainly for developers right now we can continue to iterate on and improve as we use it

napari/_qt/event_loop.py
```diff
@@ -6,6 +6,25 @@
...
+    """Create our QApplication.
+
+    Notes
+    -----
+
+    We substitute QApplicationWithTiming when using perfmon. Note that we also
+    call convert_app_for_timing() in Viewer but that one is for Jupyter/IPython
+    while this one is for running from the command line.
+    """
+    if perf.USE_PERFMON:
```
tlambert03: might we want to just directly use `os.getenv` here and put the `from .qt_event_timing` import below the conditional, to avoid importing everything in the perf module when we aren't using it?
(I looked through, and it seems to be small/fast imports... but even still?)

napari/_qt/qt_debug_menu.py
```diff
@@ -0,0 +1,107 @@
...
+        if filename:
+            filename = _ensure_extension(filename, '.json')
+
+            def start_trace():
+                perf.timers.start_trace_file(filename)
+                self._set_recording(True)
+
+            # Schedule this to avoid bogus "MetaCall" event for the entire
+            # time the file dialog was up.
+            QTimer.singleShot(0, start_trace)
```
tlambert03, line 101: this feels like a hack, how were you previously connecting the `start_trace` function and showing the dialog?   Could you use the `fileSelected` signal?

napari/viewer.py
```diff
@@ -5,11 +5,13 @@
 from qtpy.QtGui import QIcon
 from qtpy.QtWidgets import QApplication
 
+from ._qt.qt_event_timing import convert_app_for_timing
 from ._qt.qt_main_window import Window
 from ._qt.qt_viewer import QtViewer
 from ._qt.threading import wait_for_workers_to_quit, create_worker
 from .components import ViewerModel
 from . import __version__
+from .utils import perf
```
tlambert03: same questions about lazy imports in `event_loop.py` can we avoid importing any of that machinery if we don't need it?

napari/viewer.py
```diff
@@ -62,6 +64,13 @@ def __init__(
             )
             raise RuntimeError(message)
 
+        # If using perfmon we need a special QApplication. If we are run from
+        # the command line using our gui_qt context object then we already have
+        # the special one, and this will be a noop. When running inside IPython
+        # or Jupyter however this is where we switch to our QApplication.
+        if perf.USE_PERFMON:
+            app = convert_app_for_timing(app)
```
tlambert03, line 72: if we can do this here, do we need it in `event_loop.py` at all?  Or do you think it's still better to do it before instantiation if at all possible?

napari/utils/perf/_trace_file.py
```diff
@@ -0,0 +1,92 @@
...
+    """
+
+    def __init__(self, path: str):
+        """Open the tracing file on disk.
+
+        Parameters
+        ----------
+        path : str
+            Write the trace file to this path.
+        """
```
tlambert03: side note here:  I can't find the conversations where this has been discussed at the moment, but for better or worse, we follow the [numpy convention](https://numpydoc.readthedocs.io/en/latest/format.html#class-docstring) in the napari codebase, which is to put the full docstring for the init method in `Parameters` section in the class docstring.

napari/_qt/qt_viewer.py
```diff
@@ -189,6 +193,21 @@ def __init__(self, viewer):
...
+        """
+        if not perf.USE_PERFMON:
+            return None
+
+        return QtViewerDockWidget(
+            self,
+            QtPerformance(),
+            name='performance',
+            area='bottom',
+            allowed_areas=['bottom'],
```
tlambert03: while playing with this a bit ... I kinda _wanted_ to put it on the right?  Any reason to restrict this here?

napari/utils/perf/_stat.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+
+    def __init__(self, value: int):
+        """Create Stat with an initial value.
+
+        Parameters
+        ----------
+        value : int
+            Initial value.
+        """
```
tlambert03, line 32: see other comment about class docstrings

tlambert03 (commented):
The functionality is fantastic... love it @pwinston.

I do have one major concern at the moment, which is that I get this error everytime I quit the program.  Do you ever see this?  (If not, and you're on PySide2 ... that might be worth looking into?)

```pytb
WARNING: StopIteration

The above exception was the direct cause of the following exception:

SystemError: <class 'PyQt5.QtWidgets.QWidget'> returned a result with an error set
16:20:20 WARNING StopIteration

The above exception was the direct cause of the following exception:

SystemError: <class 'PyQt5.QtWidgets.QWidget'> returned a result with an error set
Abort trap: 6
/Users/talley/miniconda3/envs/napdev/lib/python3.8/multiprocessing/resource_tracker.py:216: 
UserWarning: resource_tracker: There appear to be 1 leaked semaphore objects to clean up at 
shutdown
  warnings.warn('resource_tracker: There appear to be %d '
```

I left some other comments throughout

napari/_qt/qt_performance.py
```diff
@@ -0,0 +1,174 @@
...
+        bar.setRange(0, 100)
+        bar.setValue(50)
+        bar.setFormat("%vms")
+        layout.addWidget(bar)
+        self.bar = bar
+
+        # We let the user set the "slow event" threshold.
+        self.thresh_ms = self.THRESH_DEFAULT
+        self.thresh_combo = QComboBox()
+        self.thresh_combo.addItems(self.THRESH_OPTIONS)
```
tlambert03, line 117: another trick worth knowing about... the [`QComboBox.addItem` method](https://doc.qt.io/qt-5/qcombobox.html#addItem) (but not `addItems`) accepts an optional _UserData_ value that lets you store arbitrary data with an item ... which you can then retrieve using [`currentData`](https://doc.qt.io/qt-5/qcombobox.html#currentData-prop).  This let's you avoid all the boilerplate of needing to convert back and forth between strings and floats

napari/utils/perf/_trace_file.py
pwinston: Didn't realize that was the convention. Fixed.

napari/utils/perf/_stat.py
pwinston: Fixed.

napari/_qt/qt_viewer.py
pwinston: I just picked bottom very early on. Seems to work fine on right so I changed it to allow either. We can change default to right if people want to?

tlambert03: I think bottom is a good default.  I'd just remove the `allowed_areas` restriction

napari/viewer.py
pwinston: The one in Viewer does not work for `gui_qt` because we've already created a QWidget (the splash screen) so we can't delete QApplication, it will crash. But with IPython/Jupyter we have not created any Qt objects at that point.

We could call `convert_app_for_timing` in `gui_qt`. That is call `convert_app_for_timing` in both places. But that means we'd create the `QApplication` then **immediately** delete it which seems kind of silly, since it's just as easy to create the right one.

So right now:
Command Line: create the right QApplication from the start.
IPython/Jupyter: delete the existing QApplication and replace it.

Where `convert_app_for_timing` does **not** delete the QApplication if it's already the perfmon one.

napari/viewer.py
pwinston: I made used env var and made the import lazy here too.

napari/viewer.py
tlambert03: oh fascinating!  So is that the distinction here?  For instance, if someone is using Jupyter and they create some other widget (outside of napari) before running napari with perfmon, will it crash there to?  (recognize that's basically never going to happen... just want to make sure I understand)

I'm definitely fine with the current approach, just wanted to understand

napari/_qt/qt_debug_menu.py
pwinston: Good idea but I just tried it, same problem you get a giant `MetaCall` event in the trace, covering the many seconds you had the dialog open. It must call that signal immediately in a nested way. This kind of ruins the trace file so we don't want it there.

This is a very specialized situation related to timing. When `_start_trace` is called `QApplicationWithTiming` has a timer running. If we turn on tracing immediately that super long timer goes into the trace file.

The very first thing I tried was build logic into PerfTimers to say "when you start a trace, don't really start tracing for a little bit" but that seemed messy. I think `singleShot` does exactly what we want, it's one line, it's a legal public method. Plus this is in the debug menu.

napari/_qt/event_loop.py
pwinston: Yes fixed this so perf is imported lazy. I suspect the import time is negligible but it seems reasonable to keep perf totally lazy.

napari/_qt/qt_performance.py
pwinston: That's good to know about. This doesn't seem bad with string/float but I could see `UserData` being cleaner in some cases.

pwinston:
Ugh, I don't see how fix the `StopIteration`. If I setup perfmon to do **nothing** except use this QApplication that does nothing, I still get the problem:

```
class QApplicationWithTiming(QApplication):
    def notify(self, receiver, event):
        return QApplication.notify(self, receiver, event)
```

A `try/catch` around the notify doesn't help. The `StopIteration` must be raised elsewhere during shutdown, not inside the `notify()` call.

So somehow just overriding notify at all in Python causes the problem. We know it's a C++ wrapped method because we couldn't monkey patch. So somehow Python is holding on to the references and then the C++ side deletes everything?

I tried `weakref` but not sure I did it right. If we knew we were shutting down could we do something to clean up different only for perfmon?

I tried an `atexit` handler to at least maybe print "perfmon was enable, shutdown might have errors" but it didn't get called.

Ugh, no ideas at the moment.

napari/_qt/qt_viewer.py
pwinston: Makes sense. I removed it.

tlambert03:
> Ugh, no ideas at the moment.

If you get thoroughly frustrated don't want to deal with it anymore, we can probably merge this PR without fixing that error.  But I do think it's something we need to look into eventually, and I don't think we should mention/advertise the tool to non core developers until we have it solved.

As noted in zulip, the minimal example in your previous comment _does_ work fine outside of napari, so it's definitely _something_ in our codebase that's making it problematic to reimplement `notify()`.

napari/viewer.py
pwinston: I think that's right. I think napari would delete the "shared" QApplication and it would crash. If it was not the first. Running multiple napari windows should be okay. Because the 2nd one will see it's already using QApplicationWithTiming.

Anyway this worked for timing #1320 Juan's notebook so that was the immediate goal. But this could bite someone if they just kept perfmon enabled, it would be a very mysterious crash for them, if they just opened some random notebook and it crash. Might take a long while to suspect perfmon.

Between this and the crash on exit it'd be **really** nice to find a completely new way of timing events. It'd be nice if perfmon were bulletproof. But I'm probably going to focus on #1320 itself.

Maybe there can be a V2 of perfmon at some point.

tlambert03: yeah that works for me.  as an internal tool, it's plenty useful as is

pwinston:
Yeah I'd like to merge it and create a new issue "crash on exit when using perfmon". I think that's clearly moving the ball forward.

I do wonder about the ~20 threads. In C++ I was resigned to that, mystery threads you had no idea about. In Python I would have assumed it would be more open/accessible what was going on. But in the VS Code debugger I couldn't get anything out of them, no stack, no code.

That might come up again in rendering, if we have mystery things running on us!

Maybe it's mining bitcoin.

sofroniewn:
> >If you get thoroughly frustrated don't want to deal with it anymore, we can probably merge this PR without fixing that error. But I do think it's something we need to look into eventually, and I don't think we should mention/advertise the tool to non core developers until we have it solved.
>
> Yeah I'd like to merge it and create a new issue "crash on exit when using perfmon". I think that's clearly moving the ball forward.

Based on the discussion here I'm good with this approach. I will merge now and then maybe @pwinston can make a dedicate issue for this. I'm excited to start using this more myself too!

pwinston:
I created #1349 for the errors on exit. I confirmed it's PyQt5 only, no errors with PySide2.

kevinyamauchi:
Thanks, @pwinston ! I am super excited to try this out!

Issue #1142, opened 2020-04-14
Collection of Labels Layer feature requests


hectormz:
## 🚀 Feature

Here are a collection of Labels Layer feature requests/ideas I've accumulated recently.

## Motivation

I've been using the Label Layers more, and have some ideas for new enhancements. For convenience or conversation, I've included them all here, but can break this up into many issues if that is useful.

### Fill label with double click on paint brush
There are many using labels that will likely create a closed contour with the paint brush, and then use the fill tool to flood fill. Switching the tool can be done by clicking the fill tool icon, or the `f` key.

After closing a contour, it would be nice to double-click within the contour (while still on the brush tool), which would then fill the area. This would only be done if the user is clicking within a closed contour. And it's up for debate if the color/label id must match (fill does not require this). 

All these capabilities exist, it's just automating tool changing etc

### Scroll through stack
When in fill, pick or brush mode, mouse scrolling does nothing. Switching to pan/zoom mode permanently (or temporarily with space bar) allows for zooming. I'd like to scroll through the dimensions of the stack somehow. This may relate to #1141 , and perhaps when in fill/pick/brush mode, `ctrl + scroll` or just `scroll` would scroll through image dimensions

### Convert Image layer to Labels layer
It would be nice to be able to load an image stack into `napari` and convert it to a labels layer. This can be done by a context menu when right clicking on eligible image layers. I think the only rule to determine eligibility would whether or not the `dtype` of the image is a type of `int` or perhaps if the image is only made of `0, 1`. When converting the image layer to a label layer, it could be upgraded to `int32` if not already the case.

### Select label instead of painting
I like the ability to select regions instead of painting (for visibility or accuracy concerns) like in [`ImageJ`'s `Segmentation Editor`](https://imagej.net/Segmentation_Editor):

![segmentation editor](https://imagej.net/_images/7/7c/Segmentation_Editor_3.png "Segmentation Editor")
We could add a tool in addition to the paintbrush, where with crosshairs, you could select perimeter around a region, and then add/remove it the label with another click/key press. If it's not clear, this would involve drawing the perimeter between pixels/voxels instead of painting the pixels directly.

### Contour view of labels
Visually similar to the selection method mentioned above. It could be useful to have a toggle that displays contours of labels instead of shaded regions. This might provided better/different clarity in some situations than just reducing opacity of the layer.

### Consider Eraser tool
[...]

PR #1264, opened 2020-05-19
add option to disable overwrite of labels during painting

liu-ziyang:
# Description
Add toggling during painting of labels, to turn off overwriting of existing label https://github.com/napari/napari/issues/1142

when overwrite is disabled, existing labels(other than background label - 0) won't be replaced
## Type of change

# References
https://github.com/napari/napari/issues/1142 

# How has this been tested?

- [X] created test to verify paint with overwrite toggled off correctly mark only empty areas

## Final checklist:

### DIFF
--- a/napari/_qt/layers/qt_labels_layer.py
+++ b/napari/_qt/layers/qt_labels_layer.py
@@ -107,7 +107,9 @@ class QtLabelsControls(QtLayerControls):
         self._on_n_dim_change()
 
         preserve_labels_cb = QCheckBox()
-        preserve_labels_cb.setToolTip('preserve existing labels while painting')
+        preserve_labels_cb.setToolTip(
+            'preserve existing labels while painting'
+        )
         preserve_labels_cb.stateChanged.connect(self.change_preserve_labels)
         self.preserveLabelsCheckBox = preserve_labels_cb
         self._on_preserve_labels_change()
--- a/napari/_qt/layers/qt_labels_layer.py
+++ b/napari/_qt/layers/qt_labels_layer.py
@@ -107,7 +107,7 @@ class QtLabelsControls(QtLayerControls):
         self._on_n_dim_change()
 
         preserve_labels_cb = QCheckBox()
-        preserve_labels_cb.setToolTip('preserve_labels editing')
+        preserve_labels_cb.setToolTip('preserve existing labels while painting')
         preserve_labels_cb.stateChanged.connect(self.change_preserve_labels)
         self.preserveLabelsCheckBox = preserve_labels_cb
         self._on_preserve_labels_change()
--- a/napari/_qt/layers/qt_labels_layer.py
+++ b/napari/_qt/layers/qt_labels_layer.py
@@ -66,6 +66,7 @@ class QtLabelsControls(QtLayerControls):
         self.layer.events.contiguous.connect(self._on_contig_change)
         self.layer.events.n_dimensional.connect(self._on_n_dim_change)
         self.layer.events.editable.connect(self._on_editable_change)
+        self.layer.events.overwrite.connect(self._on_overwrite_change)
 
         # shuffle colormap button
         self.colormapUpdate = QPushButton('shuffle colors')
--- a/napari/_qt/layers/qt_labels_layer.py
+++ b/napari/_qt/layers/qt_labels_layer.py
@@ -66,7 +66,9 @@ class QtLabelsControls(QtLayerControls):
         self.layer.events.contiguous.connect(self._on_contig_change)
         self.layer.events.n_dimensional.connect(self._on_n_dim_change)
         self.layer.events.editable.connect(self._on_editable_change)
-        self.layer.events.overwrite.connect(self._on_overwrite_change)
+        self.layer.events.preserve_labels.connect(
+            self._on_preserve_labels_change
+        )
 
         # shuffle colormap button
         self.colormapUpdate = QPushButton('shuffle colors')
@@ -104,11 +106,11 @@ class QtLabelsControls(QtLayerControls):
         self.ndimCheckBox = ndim_cb
         self._on_n_dim_change()
 
-        overwrite_cb = QCheckBox()
-        overwrite_cb.setToolTip('overwrite editing')
-        overwrite_cb.stateChanged.connect(self.change_overwrite)
-        self.overwriteCheckBox = overwrite_cb
-        self._on_overwrite_change()
+        preserve_labels_cb = QCheckBox()
+        preserve_labels_cb.setToolTip('preserve_labels editing')
+        preserve_labels_cb.stateChanged.connect(self.change_preserve_labels)
+        self.preserveLabelsCheckBox = preserve_labels_cb
+        self._on_preserve_labels_change()
 
         self.panzoom_button = QtModeRadioButton(
             layer, 'zoom', Mode.PAN_ZOOM, tooltip='Pan/zoom mode', checked=True
@@ -159,8 +161,8 @@ class QtLabelsControls(QtLayerControls):
         self.grid_layout.addWidget(self.contigCheckBox, 5, 1)
         self.grid_layout.addWidget(QLabel('n-dim:'), 6, 0)
         self.grid_layout.addWidget(self.ndimCheckBox, 6, 1)
-        self.grid_layout.addWidget(QLabel('overwrite:'), 7, 0)
-        self.grid_layout.addWidget(self.overwriteCheckBox, 7, 1)
+        self.grid_layout.addWidget(QLabel('preserve_labels:'), 7, 0)
+        self.grid_layout.addWidget(self.preserveLabelsCheckBox, 7, 1)
         self.grid_layout.setRowStretch(8, 1)
         self.grid_layout.setColumnStretch(1, 1)
         self.grid_layout.setSpacing(4)
@@ -254,8 +256,8 @@ class QtLabelsControls(QtLayerControls):
         else:
             self.layer.n_dimensional = False
 
-    def change_overwrite(self, state):
-        """Toggle overwrite state of label layer.
+    def change_preserve_labels(self, state):
+        """Toggle preserve_labels state of label layer.
 
         Parameters
         ----------
@@ -263,9 +265,9 @@ class QtLabelsControls(QtLayerControls):
             Checkbox indicating if overwriting label is enabled.
         """
         if state == Qt.Checked:
-            self.layer.overwrite = True
+            self.layer.preserve_labels = True
         else:
-            self.layer.overwrite = False
+            self.layer.preserve_labels = False
 
     def _on_selection_change(self, event=None):
         """Receive layer model label selection change event and update spinbox.
@@ -314,16 +316,16 @@ class QtLabelsControls(QtLayerControls):
         with self.layer.events.contiguous.blocker():
             self.contigCheckBox.setChecked(self.layer.contiguous)
 
-    def _on_overwrite_change(self, event=None):
-        """Receive layer model overwrite change event and update the checkbox.
+    def _on_preserve_labels_change(self, event=None):
+        """Receive layer model preserve_labels change event and update the checkbox.
 
         Parameters
         ----------
         event : qtpy.QtCore.QEvent, optional.
             Event from the Qt context.
         """
-        with self.layer.events.overwrite.blocker():
-            self.overwriteCheckBox.setChecked(self.layer.overwrite)
+        with self.layer.events.preserve_labels.blocker():
+            self.preserveLabelsCheckBox.setChecked(self.layer.preserve_labels)
 
     def _on_editable_change(self, event=None):
         """Receive layer model editable change event & enable/disable buttons.
--- a/napari/_qt/layers/qt_labels_layer.py
+++ b/napari/_qt/layers/qt_labels_layer.py
@@ -103,6 +103,12 @@ class QtLabelsControls(QtLayerControls):
         self.ndimCheckBox = ndim_cb
         self._on_n_dim_change()
 
+        overwrite_cb = QCheckBox()
+        overwrite_cb.setToolTip('overwrite editing')
+        overwrite_cb.stateChanged.connect(self.change_overwrite)
+        self.overwriteCheckBox = overwrite_cb
+        self._on_overwrite_change()
+
         self.panzoom_button = QtModeRadioButton(
             layer, 'zoom', Mode.PAN_ZOOM, tooltip='Pan/zoom mode', checked=True
         )
@@ -152,7 +158,9 @@ class QtLabelsControls(QtLayerControls):
         self.grid_layout.addWidget(self.contigCheckBox, 5, 1)
         self.grid_layout.addWidget(QLabel('n-dim:'), 6, 0)
         self.grid_layout.addWidget(self.ndimCheckBox, 6, 1)
-        self.grid_layout.setRowStretch(7, 1)
+        self.grid_layout.addWidget(QLabel('overwrite:'), 7, 0)
+        self.grid_layout.addWidget(self.overwriteCheckBox, 7, 1)
+        self.grid_layout.setRowStretch(8, 1)
         self.grid_layout.setColumnStretch(1, 1)
         self.grid_layout.setSpacing(4)
 
@@ -245,6 +253,19 @@ class QtLabelsControls(QtLayerControls):
         else:
             self.layer.n_dimensional = False
 
+    def change_overwrite(self, state):
+        """Toggle overwrite state of label layer.
+
+        Parameters
+        ----------
+        state : QCheckBox
+            Checkbox indicating if overwriting label is enabled.
+        """
+        if state == Qt.Checked:
+            self.layer.overwrite = True
+        else:
+            self.layer.overwrite = False
+
     def _on_selection_change(self, event=None):
         """Receive layer model label selection change event and update spinbox.
 
@@ -291,6 +312,17 @@ class QtLabelsControls(QtLayerControls):
         """
         with self.layer.events.contiguous.blocker():
             self.contigCheckBox.setChecked(self.layer.contiguous)
+
+    def _on_overwrite_change(self, event=None):
+        """Receive layer model overwrite change event and update the checkbox.
+
+        Parameters
+        ----------
+        event : qtpy.QtCore.QEvent, optional.
+            Event from the Qt context.
+        """
+        with self.layer.events.overwrite.blocker():
+            self.overwriteCheckBox.setChecked(self.layer.overwrite)
 
     def _on_editable_change(self, event=None):
         """Receive layer model editable change event & enable/disable buttons.
--- a/napari/_qt/layers/qt_labels_layer.py
+++ b/napari/_qt/layers/qt_labels_layer.py
@@ -161,7 +161,7 @@ class QtLabelsControls(QtLayerControls):
         self.grid_layout.addWidget(self.contigCheckBox, 5, 1)
         self.grid_layout.addWidget(QLabel('n-dim:'), 6, 0)
         self.grid_layout.addWidget(self.ndimCheckBox, 6, 1)
-        self.grid_layout.addWidget(QLabel('preserve_labels:'), 7, 0)
+        self.grid_layout.addWidget(QLabel('preserve labels:'), 7, 0)
         self.grid_layout.addWidget(self.preserveLabelsCheckBox, 7, 1)
         self.grid_layout.setRowStretch(8, 1)
         self.grid_layout.setColumnStretch(1, 1)
--- a/napari/layers/labels/_labels_key_bindings.py
+++ b/napari/layers/labels/_labels_key_bindings.py
@@ -78,7 +78,7 @@ def redo(layer):
 
 @Labels.bind_key('Shift')
 def preserve_labels(layer):
-    """toggle preserve label option when pressed"""
+    """Toggle preserve label option when pressed"""
     # on key press
     layer.preserve_labels = not layer.preserve_labels
 
--- a/napari/layers/labels/_labels_key_bindings.py
+++ b/napari/layers/labels/_labels_key_bindings.py
@@ -77,8 +77,8 @@ def redo(layer):
 
 
 @Labels.bind_key('Shift')
-def overwrite(layer):
-    """Disallow overwrite of labeling when pressed"""
+def preserve_labels(layer):
+    """toggle preserve label option when pressed"""
     # on key press
     layer.preserve_labels = not layer.preserve_labels
 
--- a/napari/layers/labels/_labels_key_bindings.py
+++ b/napari/layers/labels/_labels_key_bindings.py
@@ -80,9 +80,9 @@ def redo(layer):
 def overwrite(layer):
     """Disallow overwrite of labeling when pressed"""
     # on key press
-    layer.overwrite = False
+    layer.overwrite = not layer.overwrite
 
     yield
 
     # on key release
-    layer.overwrite = True
+    layer.overwrite = not layer.overwrite
--- a/napari/layers/labels/_labels_key_bindings.py
+++ b/napari/layers/labels/_labels_key_bindings.py
@@ -80,9 +80,9 @@ def redo(layer):
 def overwrite(layer):
     """Disallow overwrite of labeling when pressed"""
     # on key press
-    layer.overwrite = not layer.overwrite
+    layer.preserve_labels = not layer.preserve_labels
 
     yield
 
     # on key release
-    layer.overwrite = not layer.overwrite
+    layer.preserve_labels = not layer.preserve_labels
--- a/napari/layers/labels/_labels_key_bindings.py
+++ b/napari/layers/labels/_labels_key_bindings.py
@@ -74,3 +74,15 @@ def undo(layer):
 def redo(layer):
     """Redo any previously undone actions."""
     layer.redo()
+
+
+@Labels.bind_key('Shift')
+def overwrite(layer):
+    """Disallow overwrite of labeling when pressed"""
+    # on key press
+    layer.overwrite = False
+
+    yield
+
+    # on key release
+    layer.overwrite = True
--- a/napari/layers/labels/labels.py
+++ b/napari/layers/labels/labels.py
@@ -146,7 +146,7 @@ class Labels(Image):
 
         self.events.add(
             mode=Event,
-            overwrite=Event,
+            preserve_labels=Event,
             n_dimensional=Event,
             contiguous=Event,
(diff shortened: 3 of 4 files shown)


napari/layers/labels/_labels_key_bindings.py
```diff
@@ -74,3 +74,15 @@ def undo(layer):
...
+@Labels.bind_key('Shift')
+def overwrite(layer):
+    """Disallow overwrite of labeling when pressed"""
+    # on key press
+    layer.overwrite = False
+
+    yield
+
+    # on key release
+    layer.overwrite = True
```
tlambert03:
```suggestion
    
    current = layer.overwrite
    layer.overwrite = not current
    yield
    # on key release
    layer.overwrite = current
```
This would let the shift key toggle the current setting, rather than "do nothing" if overwrite is already False.  Note: this requires fixing the return type of the `@property`

napari/layers/labels/labels.py
```diff
@@ -341,6 +343,16 @@ def mode(self, mode: Union[str, Mode]):
         self.events.mode(mode=mode)
         self.refresh()
 
+    @property
+    def overwrite(self):
```
tlambert03: docstring

napari/layers/labels/labels.py
```diff
@@ -341,6 +343,16 @@ def mode(self, mode: Union[str, Mode]):
...
+        return str(self._overwrite)
+
+    @overwrite.setter
+    def overwrite(self, overwrite: Union[str, bool]):
```
tlambert03: what does setting with a `str` do vs setting with a `bool`?  It mostly seems like  `bool`  behavior... but the `@property` above always coerces to `str`?  (suggest limiting it to `bool`)

napari/layers/labels/labels.py
```diff
@@ -341,6 +343,16 @@ def mode(self, mode: Union[str, Mode]):
         self.events.mode(mode=mode)
         self.refresh()
 
+    @property
+    def overwrite(self):
+        return str(self._overwrite)
```
tlambert03:
```suggestion
        return self._overwrite
```

napari/layers/labels/labels.py
```diff
@@ -525,7 +537,37 @@ def paint(self, coord, new_label, refresh=True):
...
+        if self._overwrite:
+            self.data[slice_coord] = new_label
+        else:
+            self.paint_without_overwrite(self.data, slice_coord, new_label, 0)
 
         if refresh is True:
             self.refresh()
+
+    def paint_without_overwrite(
+        self, array, slice_coord, new_label, dimension
```
tlambert03: when `new_label == 0` ... painting with overwrite `False` never does anything right?  I can't think of anything more interesting at the moment... but is there anything we could use this for in that scenario?

napari/layers/labels/labels.py
```diff
@@ -525,7 +537,37 @@ def paint(self, coord, new_label, refresh=True):
...
+
+        if self._overwrite:
+            self.data[slice_coord] = new_label
+        else:
+            self.paint_without_overwrite(self.data, slice_coord, new_label, 0)
 
         if refresh is True:
             self.refresh()
+
+    def paint_without_overwrite(
```
tlambert03: docstring

napari/layers/labels/labels.py
```diff
@@ -525,7 +537,37 @@ def paint(self, coord, new_label, refresh=True):
...
+                if array[coord] == 0:
+                    array[coord] = new_label
+        else:
+            for coord in self.read_dimension(slice_coord, dimension):
+                self.paint_without_overwrite(
+                    array[coord], slice_coord, new_label, dimension + 1
+                )
+
+    @staticmethod
+    def read_dimension(slice_coord, dimension):
```
tlambert03: docstring

tlambert03 (commented):
welcome @ziyangczi!  this looks cool, and useful!
I left some suggestions about typing, and the behavior of the shift press.  Also, please add docstrings to all public methods.

napari/layers/labels/labels.py
```diff
@@ -525,7 +537,37 @@ def paint(self, coord, new_label, refresh=True):
             slice_coord = tuple(slice_coord)
 
         # update the labels image
-        self.data[slice_coord] = new_label
+
+        if self._overwrite:
+            self.data[slice_coord] = new_label
+        else:
+            self.paint_without_overwrite(self.data, slice_coord, new_label, 0)
```
sofroniewn: I wonder if instead of having a new method here which contains for loops if there is a faster method for grabbing the coordinates of the background values of the numpy array and then just writing into those directly - @jni or @kne42 might have an idea here.

napari/layers/labels/labels.py
```diff
@@ -525,7 +537,37 @@ def paint(self, coord, new_label, refresh=True):
...
 
         if refresh is True:
             self.refresh()
+
+    def paint_without_overwrite(
+        self, array, slice_coord, new_label, dimension
+    ):
+        if dimension == len(slice_coord) - 1:
+            for coord in self.read_dimension(slice_coord, dimension):
+                if array[coord] == 0:
```
sofroniewn: We should probably add a new property `self.background_label = 0` and use that here instead. Will be useful in case we expand to support colormaps with non-zero backgrounds at some point (some people use `-1` in the community). Will also make support features like an "eraser" more clear in the future as then it will pull the background label directly

sofroniewn:
Very exciting for your first PR @ziyangczi!!! Thanks for the initial review too @tlambert03, I've left a couple minor comments too. Great job :-)

napari/layers/labels/labels.py
sofroniewn: Maybe something like
```python
np.nonzero(self.data[slice_coord] == self.background_label)
```
though you'll have to make the coords to this new object line up with the coords of the full object. Generally though approaches using numpy commands like this tend to be faster I think.

liu-ziyang: I am a bit unsure about this as well, some expertise would be much appreciated. The challenge I am trying to work around is that slice hides each individual coordinate, where we would need an explicit mapped out matrix to track status for each point.

napari/layers/labels/labels.py
liu-ziyang: that is correct, it doesn't really matter what the target label is if the user want to avoid write over already labeled data. However, it is a good point since we can repurpose it to be a "eraser" function, which is raised in the same issue. We need more UXR input to map out if the behavior is desired

napari/layers/labels/_labels_key_bindings.py
liu-ziyang: I have been debating with myself whether we should make it a configurable in the panel, similar to contiguous/n-dim, or we change the painting behavior while user keeps the key pressed, if the later then without the key press, we would always default the value to true, so we would never see use of shift as a negate

tlambert03: Not everyone will be strictly interacting via the GUI.  Is there any specific reason to _disallow_ someone from programatically setting `layer.overwrite = False` (as you did in the test)?  If we allow that to be set programmatically (as it currently is), then at the very least i think that pressing shift shouldn't undo their setting... but I also think using shift to toggle (rather than assert) the behavior is a nice feature.

sofroniewn: I think `Shift` as a toggle is a nice approach, and someone can set the behavior they want to be unmodified. As a default we should overwrite.

sofroniewn: We can always revisit this with more advance UX testing, but for now I think the approach I just mentioned will be fine

napari/layers/labels/labels.py
```diff
@@ -341,6 +344,20 @@ def mode(self, mode: Union[str, Mode]):
         self.events.mode(mode=mode)
         self.refresh()
 
+    @property
+    def overwrite(self):
+        """
+        Defines if painting should overwrite existing labels, default to true to allow overwrite.
```
tlambert03: I'll do this for @jni :)
please make this (and other) docstrings [PEP-257](https://www.python.org/dev/peps/pep-0257/) compliant.  That is: one line, followed by period.  double space, then extra info:
```python
def function(arg: Type):
    """One-liner not exceeding line-length, ending in period.

    Any additional info you'd like to provide...
    """
```

napari/layers/labels/labels.py
```diff
@@ -525,7 +542,54 @@ def paint(self, coord, new_label, refresh=True):
...
+    ):
+        """
+        Paint the label without overwriting existing label, which is equivalent to only painting on background label.
+
+        Parameters
+        ----------
+        array : data array to paint on
+        slice_coord : coordinates to paint
+        new_label: new label to paint with
+        dimension: dimension of the current painting progress
```
tlambert03: follow [numpy docstrings](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html) for params:

```suggestion
        Parameters
        ----------
        array : np.ndarray
            Data array to paint on
        slice_coord : tuple
            Coordinates to paint
        new_label : int
            New label to paint with
        dimension : int
            Dimension of the current painting progress
```

napari/layers/labels/labels.py
sofroniewn: I just checked if you just do
```python
keep_coords = np.nonzero(self.data[slice_coord] == self._background_label)
self.data[slice_coord][keep_coords] = new_label
```
then everything works great - no need for any additional functions

napari/layers/labels/labels.py
```diff
@@ -525,7 +542,54 @@ def paint(self, coord, new_label, refresh=True):
             slice_coord = tuple(slice_coord)
 
         # update the labels image
-        self.data[slice_coord] = new_label
+
+        if self._overwrite:
```
sofroniewn: I wonder if we should also use this option when the brush is set to be the "background_label" as otherwise when painting with the background label nothing happens in non-overwrite mode

napari/layers/labels/labels.py
jni: Good stuff! We can do one simpler:

```python
keep_coords = (self.data[slice_coord] == self._background_label)
```

liu-ziyang: huuray!! thanks for the super cool one liner! 
on a related topic, I am bootstraping myself to python (I am able to write somewhat functional but lengthy code like this, but not able to produce elegant solutions like what you suggested) any recommended reading on the nice python tips/tricks?

sofroniewn: > I am bootstraping myself to python (I am able to write somewhat functional but lengthy code like this, but not able to produce elegant solutions like what you suggested) any recommended reading on the nice python tips/tricks?

I'm going to give a shout out to [Elegant SciPy: The Art Scientific Python](https://www.amazon.com/Elegant-SciPy-Art-Scientific-Python/dp/1491922877) written by our very own @jni! I should actually get myself a copy too now that I think about it ;-)

Also available as a repo here https://github.com/elegant-scipy/elegant-scipy

liu-ziyang:
Thanks everyone for your feedback! I give it more thought and added the eraser function as suggested in the issue, I think it is a better solution to address overwriting with background label

napari/layers/labels/_labels_key_bindings.py
liu-ziyang: I made it a switch in the panel in the end, during my testing I found it hard to tell if I am in overwrite mode or not if there is no explicit information. so I made it an explicit toggle

tlambert03 (commented):
don't have time for a full re-review yet... but a couple quick thoughts:
I like the direction this is going!  I think the explicit eraser is more intuitive than painting a label of zero.  It's introduced a couple new issues:

- icons are now too tightly packed on only the labels:
   Screen Shot 2020-05-20 at 5 22 45 PM
   (everything else):
   Screen Shot 2020-05-20 at 5 23 09 PM
we could potentially change the "shuffle colors" button to a push button with a standard "shuffle" logo...  might give us some space

- adding `self.selectionSpinBox.setMinimum(1)` in the Qt controller doesn't prevent someone from setting it manually with `layer.selected_label = 0` ... so it introduces the possibility that the GUI can become out of sync with the model.  For instance, set `layer.selected_label = 0` in the console and you'll get this:
        Screen Shot 2020-05-20 at 5 30 13 PM
    (the label in the spinbox does not agree with the color swatch _and_ the radio button is still on zoom, even though it should be on erase)
- certain key bindings are also now out of sync: pressing the "e" key for instance, should select the eraser tool and not change the label swatch... (it's the same as the previous point, but with a hotkey instead of going into the console).
- the "d" key (`decrease_label_id`) also lets the user get the GUI out of sync now, by decreasing the label down to 0...
- I actually really liked the shift key behavior!  personally, I would leave it in there as a toggle to the overwrite checkbox (i.e. the shift key swaps the overwrite behavior while pressed).
- I think it would also be nice to have eraser toggle key, perhaps the option key similar to photoshop and other "deselect" tools .... but maybe we don't need both that _and_ the `e` key, which permanently sets the eraser too?  thoughts?

liu-ziyang:
> * adding `self.selectionSpinBox.setMinimum(1)` in the Qt controller doesn't prevent someone from setting it manually with `layer.selected_label = 0` ... so it introduces the possibility that the GUI can become out of sync with the model.  For instance, set `layer.selected_label = 0` in the console and you'll get this:
>   Screen Shot 2020-05-20 at 5 30 13 PM
>   (the label in the spinbox does not agree with the color swatch _and_ the radio button is still on zoom, even though it should be on erase)
> * certain key bindings are also now out of sync: pressing the "e" key for instance, should select the eraser tool and not change the label swatch... (it's the same as the previous point, but with a hotkey instead of going into the console).
> * the "d" key (`decrease_label_id`) also lets the user get the GUI out of sync now, by decreasing the label down to 0...

Thanks for the feedback! I am personally thinking we can disallow using of label 0, other than an eraser use case, but I am very new to the project and can miss other's perspectives. Is there a valid case where someone would "paint" background than "erasing it"? I am thinking we should restrict every function from access the background element

liu-ziyang:
> * I actually really liked the shift key behavior!  personally, I would leave it in there as a toggle to the overwrite checkbox (i.e. the shift key swaps the overwrite behavior while pressed).

I like the shift behavior too! the problem I was having is that when there are 2 switch (panel and shift key), sometimes I got confused if I should keep the key pressed or not to do what I want, it is a little inexplicit, and I have to try paint somewhere first and then realize: oh I am in non-overwrite mode now. Especially when I am switching back and forth and forget where I am. That being said, I am not a long time graphical tool user, maybe it is less confusing to others? I am totally ok adding it back

liu-ziyang:
> * I think it would also be nice to have eraser toggle key, perhaps the option key similar to photoshop and other "deselect" tools .... but maybe we don't need both that _and_ the `e` key, which permanently sets the eraser too?  thoughts?

I was not aware there was an de-select tool >.< it makes sense we deprecate that and reassign hotkey e to eraser, which makes more sense to me. Happy to hear what others think too

liu-ziyang:
> * icons are now too tightly packed on only the labels:
>   Screen Shot 2020-05-20 at 5 22 45 PM
>   (everything else):
>   Screen Shot 2020-05-20 at 5 23 09 PM
>   we could potentially change the "shuffle colors" button to a push button with a standard "shuffle" logo...  might give us some space

> * I think it would also be nice to have eraser toggle key, perhaps the option key similar to photoshop and other "deselect" tools .... but maybe we don't need both that _and_ the `e` key, which permanently sets the eraser too?  thoughts?

I got this idea from "City Skylines" 😄 maybe other imaging tools have similar feature too, we can remove eraser as a button, but use right-click/mouse drag as "erasing" behavior, it might be a pain to erase on large area, maybe we can also implement double left click to fill and double right click to unfill/erasing? how's this idea sound?

napari/layers/labels/labels.py
```diff
@@ -341,6 +355,22 @@ def mode(self, mode: Union[str, Mode]):
...
+        with previous behavior, if set to false, painting would
+        only be performed on the background label.
+        """
+        return self._overwrite
+
+    @overwrite.setter
+    def overwrite(self, overwrite: bool):
+        self._overwrite = overwrite
+        self.events.overwrite(overwrite=overwrite)
+        self.refresh()
```
sofroniewn: I don't think the overwrite setter needs the `refresh` as changing the overwrite doesn't actually change how any of the data gets displayed - it just changes how data will get modified

napari/layers/labels/_labels_mouse_bindings.py
```diff
@@ -35,3 +35,27 @@ def fill(layer, event):
     """Fill in an area with the currently selected label."""
     # on press
     layer.fill(layer.coordinates, layer._value, layer.selected_label)
+
+
+def erase(layer, event):
```
sofroniewn: I worry this functions adds a lot of code duplication with the paint method for little gained - the only different is now we use `layer.selected_label` instead of `layer.background_label`. I think allowing the `selected_label` to be the `background_label` is fine, the background label is a label after all, just a special one, so I think I'd prefer this to be dropped and inside the ERASE mode setter just set the `selected_label` to be the `background_label`. I'll leave a comment there too

napari/layers/labels/labels.py
jni: > I'm going to give a shout out to Elegant SciPy: The Art Scientific Python

It's a good book. =P Another famous book in the space is the Python Data Science Handbook by Jake VanderPlas: https://jakevdp.github.io/PythonDataScienceHandbook/

Finally, not everything in @stefanv's and my Advanced NumPy tutorial made it onto the book, so that's a good place to get good ideas on what's possible with NumPy indexing:

https://github.com/ASPP/2019-camerino-advanced-numpy/blob/master/Advanced%20NumPy%20Patterns.ipynb

napari/layers/labels/labels.py
```diff
@@ -327,6 +335,12 @@ def mode(self, mode: Union[str, Mode]):
             self.interactive = False
             self.help = 'hold <space> to pan/zoom, drag to paint a label'
             self.mouse_drag_callbacks.append(paint)
+        elif mode == Mode.ERASE:
+            self.cursor_size = self.brush_size / self.scale_factor
+            self.cursor = 'square'
+            self.interactive = False
+            self.help = 'hold <space> to pan/zoom, drag to erase a label'
+            self.mouse_drag_callbacks.append(erase)
```
```suggestion
            self.mouse_drag_callbacks.append(paint)
            self.selected_label = self.background_label
```
I think it is fine to have the selected label be the background label too. It removes the need for a separate highly redundant function in the mouse key bindings, unless i am missing something - in which case please let me know

napari/layers/labels/labels.py
```diff
@@ -309,6 +315,8 @@ def mode(self, mode: Union[str, Mode]):
             self.mouse_drag_callbacks.remove(pick)
         elif self._mode == Mode.PAINT:
             self.mouse_drag_callbacks.remove(paint)
+        elif self._mode == Mode.ERASE:
+            self.mouse_drag_callbacks.remove(erase)
```
```suggestion
            self.mouse_drag_callbacks.remove(paint)
```

napari/layers/labels/labels.py
sofroniewn: Given these suggestions I think we should also add something to the `selected_label` setter that checks whether someone has selected the background color, and if so change the mode into `erase`.

There is an additional problem now though, which I just thought of, which is that now someone can't use the background label with the fill bucket ...

jni:
> I got this idea from "City Skylines" smile maybe other imaging tools have similar feature too, we can remove eraser as a button, but use right-click/mouse drag as "erasing" behavior, it might be a pain to erase on large area, maybe we can also implement double left click to fill and double right click to unfill/erasing? how's this idea sound?

Buttons are good, they are explicit for beginners. Mouse click combos are hard to discover, and right click should almost always be "open context menu". Middle click is flexible. =)

`e` could be "toggle eraser mode" rather than "set eraser mode". Indeed, maybe all mode shortcuts should work as toggles...?

@ziyangczi the other thing I would add is that scope creep in PRs is generally a bad idea — this PR was almost ready to merge but now we are discussing a totally unrelated UI issue! So I would recommend in the future to open a new branch + PR for each new feature.

sofroniewn:
Hi @ziyangczi, I just went through - overall looking good - but a couple comments.

It's great that you've begun to add erase mode, but I think this really belongs as two separate PRs. In general we try and keep our PRs as short and modular as possible - i.e. one piece of functionality at a time - as it makes review much easier and discussions more focused. This PR now adds both the overwrite toggle and the eraser. For the cases like this one when there are two related things and you want to work on both at the same time with functionality from the first one in the second one, then we make two PRs where the second one contains the changes from the first.

In this case if you could change this PR back to only having the overwrite and addition of the background label, and then make a new PR which had those changes + the addition of the erase button that would be great. I think we were very close to merging the overwrite toggle, but the eraser will require some more discussion as there are some quirks with who we think about erasing as a paintbrush, and erasing as a fill bucket (see some of my comments in the PR).

As to having a shift key vs the checkbox for the overwrite - I agree with @tlambert03 that having the modifier is much more natural, and it's what other tools in this space do. Maybe we can try having both, so the checkbox changes the default, and the modifier is a temporary toggle (which also updates the checkbox) but reverts on release. That way you have a visual representation of the state but you can very quickly change it using a hotkey. How does that sound?

napari/layers/labels/labels.py
sofroniewn: Think about that more we probably shouldn't auto change mode when `selected_label` is made equal to the background label, so people can still use fill as before

sofroniewn:
Ah as I was writing my comment @jni commented too, but I didn't see it till now!! My apologies!!

liu-ziyang:
reduced scope as suggested, going to open up another PR to discuss eraser

napari/_qt/layers/qt_labels_layer.py
```diff
@@ -103,6 +103,12 @@ def __init__(self, layer):
         self.ndimCheckBox = ndim_cb
```
sofroniewn, line 106: near the top of this file ~ line 68 you need to add
```python
        self.layer.events.overwrite.connect(self._on_overwrite_change)
```

so that changing the overwrite programmatically or with the hotkey triggers the change of the checkbox state

sofroniewn:
This is almost done @ziyangczi thanks for pulling out other code. There is one event that needs to be connected, then I think it is ready.

Maybe a final question for @tlambert03 and @jni though is do we like the word `overwrite`. I worry that overwrite sounds very intense, and this is the behavior that most people will expect by default. We could alternatively have `block_painting` which then get's set to `True` when you want the blocking behavior. I prefer this as it is more explicit and by default it is `False`.

napari/_qt/layers/qt_labels_layer.py
liu-ziyang: ha! nice catch! (admittedly I need to up my game in qt as well. To be completely honest I never worked with any graphical tools at all so far. Though I am familiar with MVC model but my only experience with UI is from an undergraduate course which was a nightmare.. any recommendation reading in qt as well?)

sofroniewn: For qt I've mainly just learnt through the qt docs / stack overflow. I'm not sure if there are any good books.

Another to note is that we havn't really got good GUI tests for the Qt side of things right now - i caught that manually by trying out the PR. With time we might want to start adding/ improving them

liu-ziyang:
> This is almost done @ziyangczi thanks for pulling out other code. There is one event that needs to be connected, then I think it is ready.
> 
> Maybe a final question for @tlambert03 and @jni though is do we like the word `overwrite`. I worry that overwrite sounds very intense, and this is the behavior that most people will expect by default. We could alternatively have `block_painting` which then get's set to `True` when you want the blocking behavior. I prefer this as it is more explicit and by default it is `False`.

Definitely welcome suggestion over naming, I felt "overwrite" is very broad, too, and can lead to misunderstanding. "block_painting" also sounds like we are disabling painting all together, how about "replace_existing_label" ?

sofroniewn:
> Definitely welcome suggestion over naming, I felt "overwrite" is very broad, too, and can lead to misunderstanding. "block_painting" also sounds like we are disabling painting all together, how about "replace_existing_label" ?

How about `preserve_existing_labels` and have it be `False` by default. I like having the default behavior as `False` because then people who don't want to use this functionality don't need to learn about it.

tlambert03:
> how about `preserve_existing_labels`

I was gonna say the same thing :)

liu-ziyang:
Thanks everyone for your thorough review and very useful feedback! I finalized the PR and please take another look and let me know if things look good. Sorry the PR is messier than it should have been. Hopefully, as I ramp up speed, I would be more thoughtful and productive contributing on other issues

napari/layers/labels/labels.py
```diff
@@ -525,7 +542,12 @@ def paint(self, coord, new_label, refresh=True):
             slice_coord = tuple(slice_coord)
 
         # update the labels image
-        self.data[slice_coord] = new_label
+
+        if not self._preserve_labels:
```
sofroniewn: very minor but i would just use `self.preserve_labels` here

napari/_qt/layers/qt_labels_layer.py
```diff
@@ -152,7 +161,9 @@ def __init__(self, layer):
         self.grid_layout.addWidget(self.contigCheckBox, 5, 1)
         self.grid_layout.addWidget(QLabel('n-dim:'), 6, 0)
         self.grid_layout.addWidget(self.ndimCheckBox, 6, 1)
-        self.grid_layout.setRowStretch(7, 1)
+        self.grid_layout.addWidget(QLabel('preserve_labels:'), 7, 0)
```
sofroniewn:
```suggestion
        self.grid_layout.addWidget(QLabel('preserve labels:'), 7, 0)
```

napari/_qt/layers/qt_labels_layer.py
```diff
@@ -103,6 +106,12 @@ def __init__(self, layer):
         self.ndimCheckBox = ndim_cb
         self._on_n_dim_change()
 
+        preserve_labels_cb = QCheckBox()
+        preserve_labels_cb.setToolTip('preserve_labels editing')
```
sofroniewn:
```suggestion
        preserve_labels_cb.setToolTip('preserve existing labels while painting')
```

sofroniewn (approved):
Looks great @ziyangczi very nice job, and congrats on your first napari PR!! Very exciting to get this in. I am approving now, I left a couple very minor comments. We we will then let sit for a little and see if anyone else has any additional feedback. Should be good to merge soon.

I really like the functionality - see here for a gif of using it!!

preserve_painting

jni (approved):

VolkerH:
Thanks for that ! Very useful.

kevinyamauchi:
Awesome work @ziyangczi ! Amazing first PR!

PR #1268, opened 2020-05-22
change readme screenshot

sofroniewn:
# Description
Changes readme screenshot to mesoscope image, same as in napari-tutorials


sofroniewn:
screenshot-add-image

sofroniewn:
screenshot-add-image

sofroniewn:
This PR would change the screenshot away from the astronaut to this more sciency image that we use on napari tutorials, but i just remembered one reason we have it like that is because we have to code that describes how to make the image right above ....

kevinyamauchi (commented):
I am in favor of adding a new readme screenshot and this one is pretty! As was discussed in an issue (I can't seem to find it at the moment), I think we should add a script in the examples or elsewhere showing the code to generate the image.

With regards to the current image being generated by the demo code just above it, I do think that is super nice. We may want to update it anyway though because I don't think it would work for a user that has installed `napari` in a fresh environment since we no longer depend on `skimage`.

Also, super minor, but it might be nice to name the file something more related to its usage here.

sofroniewn:
We want this to be autogenerated, so this particular choice of photo is out for now. Even if we revisit this issue it will not be using this approach, so closing now

PR #1269, opened 2020-05-22
Rich jupyter display for napari screenshots

### LINKED ISSUE #1267: Add `nbscreenshot` function or similar that will embed the current state of the viewer in screenshot
## 🚀 Feature

When using napari in notebooks, the recorded notebook becomes a sad shadow of all that went on in that notebook, because none of the napari windows or interactions are visible. It would be great to have some way of embedding the current viewer as a screenshot in the notebook. (see aside in #1237.) I proposed doing this implicitly in #756 but we dismissed that idea as too confusing. If made explicit though, it could be a powerful way to record interactions. For example, if nbscreenshot returns an object with a `_repr_png_` method, we could do:

```python
# cell 1
import napari

v = napari.Viewer()
v.add_image(my_data)
pts = v.add_points()
pts.mode = 'add'
# annotate
```

Then
```python
# cell 2
nbscreenshot(v)
```

would show the viewer with the annotated points. Then the notebook would continue with the analysis.

Taking it further, if you look at the IPython display documentation:

https://ipython.readthedocs.io/en/stable/api/generated/IPython.display.html

there's also the ability to display a video. So it would even be possible to do e.g. `napari.record(True)`, then `nbrecording(v)` would embed a video of the viewer's interactions.

I would make this a stretch goal though, the screenshot bit is easy-ish and important.

GenevieveBuckley:
# Description

This PR implements a new viewer method `nbscreenshot()`, which provides rich image display of the napari screenshots in jupyter notebooks.

[See example output of the new functionality in the notebook gist here](https://gist.github.com/GenevieveBuckley/97b49f9ac3e02422f98abe3c412ee2a0)

Based on Juan's approach in [#756](https://github.com/napari/napari/pull/756) (one difference to note is that this PR removes the `scikit-image` dependency in favour of `imageio`).

## Type of change

# References

Closes https://github.com/napari/napari/issues/1267

# How has this been tested?

- [x] test covering new functionality
- [x] all existing tests pass with my change

## Final checklist:

### DIFF
--- a/napari/utils/_tests/test_notebook_display.py
+++ b/napari/utils/_tests/test_notebook_display.py
@@ -1,0 +1,18 @@
+import numpy as np
+
+from napari.utils.notebook_display import nbscreenshot
+
+
+def test_nbscreenshot(viewer_factory):
+    """Test taking a screenshot."""
+    view, viewer = viewer_factory()
+
+    np.random.seed(0)
+    data = np.random.random((10, 15))
+    viewer.add_image(data)
+
+    rich_display_object = nbscreenshot(viewer)
+    assert hasattr(rich_display_object, '_repr_png_')
+    # Trigger method that would run in jupyter notebook cell automatically
+    rich_display_object._repr_png_()
+    assert rich_display_object.image is not None
--- a/napari/utils/_tests/test_notebook_display.py
+++ b/napari/utils/_tests/test_notebook_display.py
@@ -1,6 +1,6 @@
 import numpy as np
 
-from napari.utils.notebook_display import nbscreenshot
+from napari.utils import nbscreenshot
 
 
 def test_nbscreenshot(viewer_factory):
--- a/napari/utils/notebook_display.py
+++ b/napari/utils/notebook_display.py
@@ -3,11 +3,13 @@ from io import BytesIO
 __all__ = ['nbscreenshot']
 
 
-def nbscreenshot(viewer, canvas_only=False):
+def nbscreenshot(viewer, *, canvas_only=False):
     """Display napari screenshot in the jupyter notebook.
 
     Parameters
     ----------
+    viewer : napari.Viewer
+        The napari viewer.
     with_viewer : bool, optional
         If True includes the napari viewer frame in the screenshot,
         otherwise just includes the canvas. By default, True.
@@ -42,7 +44,7 @@ class NotebookScreenshot:
     ```
     """
 
-    def __init__(self, viewer, canvas_only=False):
+    def __init__(self, viewer, *, canvas_only=False):
         """Initalize screenshot object.
 
         Parameters
--- a/napari/utils/notebook_display.py
+++ b/napari/utils/notebook_display.py
@@ -1,7 +1,5 @@
 from io import BytesIO
 
-import imageio
-
 
 class NotebookScreenshot:
     """Display napari screenshot in the jupyter notebook.
@@ -36,6 +34,8 @@ class NotebookScreenshot:
             If True includes the napari viewer frame in the screenshot,
             otherwise just includes the canvas. By default, True.
         """
+        import imageio
+        
         self.viewer = viewer
         self.with_viewer = with_viewer
         self.image = None
--- a/napari/utils/notebook_display.py
+++ b/napari/utils/notebook_display.py
@@ -1,5 +1,24 @@
 from io import BytesIO
 
+__all__ = ['nbscreenshot']
+
+
+def nbscreenshot(viewer, canvas_only=False):
+    """Display napari screenshot in the jupyter notebook.
+
+    Parameters
+    ----------
+    with_viewer : bool, optional
+        If True includes the napari viewer frame in the screenshot,
+        otherwise just includes the canvas. By default, True.
+
+    Returns
+    -------
+    napari.utils.notebook_display.NotebookScreenshot
+        Napari screenshot rendered as rich display in the jupyter notebook.
+    """
+    return NotebookScreenshot(viewer, canvas_only=canvas_only)
+
 
 class NotebookScreenshot:
     """Display napari screenshot in the jupyter notebook.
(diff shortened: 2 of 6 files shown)


napari/utils/notebook_display.py
```diff
@@ -0,0 +1,54 @@
+from io import BytesIO
+
+import imageio
```
sofroniewn: I'm trying to remember now, but I think in the rest of the codebase we've made our imports of imageio lazy, to make launch faster. We could put it inside `_repr_png_`. I'd probably also go `from imageio import imsave` too. Thoughts @jni?

sofroniewn (approved):
This is amazing @GenevieveBuckley. The functionality is so nice!! I think I'll start using it in all the notebooks I have for teaching purposes. It's a game changer.
I've left one fairly minor comment on the imageio import, otherwise I think it is good to go :-)

image

kevinyamauchi (approved):
This is so awesome! Thanks @GenevieveBuckley !

tlambert03 (approved):
indeed!  lovely!

napari/utils/notebook_display.py
GenevieveBuckley: That's a good point, thanks. I've made those changes for you now.

jni:
I love this (thanks @GenevieveBuckley!) but fwiw I think it should be a function, not a method in the viewer. Do others have strong feelings about this? I think we should make an effort to keep the viewer namespace clean and tidy.

sofroniewn:
>  I think it should be a function, not a method in the viewer. Do others have strong feelings about this? I think we should make an effort to keep the viewer namespace clean and tidy.

That's fine with me. I see it as mostly something people will use when preparing tutorials/ sharing notebooks with other people etc see proposal here https://github.com/napari/tutorials/issues/70. But I'm sympathetic to other points of view here too if people like the convenience of a method.

GenevieveBuckley:
> I love this (thanks @GenevieveBuckley!) but fwiw I think it should be a function, not a method in the viewer. Do others have strong feelings about this? I think we should make an effort to keep the viewer namespace clean and tidy.

The only reason I didn't do this is because I didn't think this function signature was very readable:
```
nbscreenshot(viewer, with_viewer=...)
```
I don't like it because `viewer, with_viewer=` seems too confusing.

What I'd like to do is change the kwarg to `show_viewer_frame` (or something more descriptive), but then we also need to also change the regular `screenshot()` `with_viewer` kwarg  to `show_viewer_frame` as well. I feel very strongly that `screenshot()` and `nbscreenshot()` should have matching keyword arguments.

Are we happy for me to go ahead and change that?

sofroniewn:
How about `include_frame` or `include_viewer_frame`? I'd change in all places.

jni:
Well, the technical name is for the UI elements is "chrome", but Google has annoyingly made that term ambiguous. Here are a few options I prefer to `include_frame`:
- `include_ui=`
- `include_ui_elements=`
- `canvas_only=`
- `chrome=`

I think `canvas_only` is probably the clearest?

GenevieveBuckley:
> I think `canvas_only` is probably the clearest?

Agreed

napari/utils/_tests/test_notebook_display.py
```diff
@@ -0,0 +1,18 @@
+import numpy as np
+
+from napari.utils.notebook_display import nbscreenshot
```
sofroniewn: Do we want to expose `nbscreenshot` at the `napari.utils` level so one could just type
`from napari.utils import nbscreenshot`?

I think that would be nice

napari/viewer.py
```diff
@@ -103,27 +103,28 @@ def update_console(self, variables):
         else:
             self.window.qt_viewer.console.push(variables)
 
-    def screenshot(self, path=None, *, with_viewer=False):
+    def screenshot(self, path=None, *, canvas_only=False):
```
sofroniewn: I really like `canvas_only`. I'm ok with this now being false by default, but I know previously @jni had expressed desires for it to be `True` by default, and so I just want to make sure he's ok with this change.

For the nbscreenshot method I like `False` by default, so it might be then strange if our two screenshot functions had different defaults.

jni: Just btw this is a breaking API change. Maybe we want to keep `with_viewer` around for a couple of versions with a warning? Or with a ValueError pointing them to the new kwarg?

sofroniewn: I feel like most user won't be using `with_viewer` they will have nothing, and now they will need to add the `canvas_only` keyword to preserve the previous behavior. We might have to ask @guiwitz to update [naparimovie](https://github.com/guiwitz/naparimovie) to add `canvas_only=True` and pin to `napari>=0.3.2` until we get #780 merged otherwise `naparimovie` will break after this PR gets merged.

sofroniewn: and by break, i mean now include the napari viewer frame by default

guiwitz: No problem I'll update naparimovie for this. Thanks @GenevieveBuckley for this functionality which will be very helpful for demos/teaching! Do I correctly understand that we are going from having a default ```screenshot``` with no napari-viewer to a default with it? Isn't that ```screenshot``` function the only way for people to grab an output from napari that is usable e.g. to make a figure? In that case I don't think most people want the canvas by default. Except those using screenshots for demos/teaching, which now can use the ```nbscreenshot``` function which could have a default with napari-viewer.

GenevieveBuckley: It's @jni you need to thank, @guiwitz!

It was his idea, and this PR is virtually indistinguishable from his initial implementation. There's only one or two tiny changes here (`_repr_png_` lives in it's own object class now, not on the viewer class, and I removed the scikit-image dependency)

It's really nice, I like it a lot too. Thanks @jni

jni: Aw shucks. 😊 Thanks for modernising it @GenevieveBuckley!

I agree with @guiwitz that `.screenshot()` should default to `canvas_only=True` and `nbscreenshot()` to `canvas_only=False`. They are not the same function, but rather serve very different purposes, so different defaults make sense, imho. What do you think @sofroniewn @GenevieveBuckley?

jni (approved):
I'm going to approve this, happy to go with the group consensus on the default values issue, just wanted to make sure it's flagged and that everyone's opinion is on the table and recorded. =)

napari/viewer.py
sofroniewn: I am fine with those two functions having different defaults. Agreed they serve different purposes. @GenevieveBuckley if you agree do you want to change the `.screenshot()` default to `canvas_only=True`, and then I will merge this!! And I don't think @guiwitz will actually need to update `naparimovie` as a quick search of the code turns up no hits for `with_viewer`. Probably indicates we're making the right choice here

PR #1271, opened 2020-05-22
Have layer tooltip display the layer name: add self.layer.name to layer tooltip

cwood1967:
# Description
When hovering over a layer in the layer list, the tooltip just displays "Layer Name"

I changed the static string 'Layer Name' to self.layer.name in qt_layerlist.py, and also set the tool tip after a name change in QtLayerWidget.changeText().

## Type of change

# References
I couldn't find an open issue for this.

# How has this been tested?
Checked to see if the changes worked on another workstation after pushing and pulling.

## Final checklist:

### DIFF
--- a/napari/_qt/qt_layerlist.py
+++ b/napari/_qt/qt_layerlist.py
@@ -511,6 +511,7 @@ class QtLayerWidget(QFrame):
         textbox = QLineEdit(self)
         textbox.setText(layer.name)
         textbox.home(False)
+        # Chris Wood changed from 'Layer Name' 2020/05/22
         textbox.setToolTip(self.layer.name)
         textbox.setAcceptDrops(False)
         textbox.setEnabled(True)
@@ -557,6 +558,8 @@ class QtLayerWidget(QFrame):
     def changeText(self):
         """Update layer name attribute using layer name textbox contents."""
         self.layer.name = self.nameTextBox.text()
+        # set tool tip to layer name, Chris Wood 20202/05/22"""
+        self.nameTextBox.setToolTip(self.layer.name)
         self.nameTextBox.clearFocus()
         self.setFocus()
 
--- a/napari/_qt/qt_layerlist.py
+++ b/napari/_qt/qt_layerlist.py
@@ -511,7 +511,7 @@ class QtLayerWidget(QFrame):
         textbox = QLineEdit(self)
         textbox.setText(layer.name)
         textbox.home(False)
-        textbox.setToolTip('Layer name')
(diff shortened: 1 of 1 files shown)


tlambert03:
Thanks @cwood1967.  You know... now that you mention it, I'm not sure we need a tooltip there at all if it's just going to show exactly what is already in the layer box (just a little smaller 😊).  What do you think?  Do you think it adds something to have that tooltip?  (I saying "Layer Name" just tells you "this is the layer name" ... but if you found that redundant or too obvious enough to make a PR, maybe we remove it all together?)

cwood1967:
@tlambert03 I was using napari this morning; I had multiple layers with the same first part of the name. For example, an image is added with channel_axis=x, the layers are named Image, Image [1], etc. If a name is given in add_image that is descriptive it will likely be longer than the layer text can show, and the numbers will be hidden. So, I think it is helpful to hover and know exactly what layer you are going to select or delete. There might be a better way to display that information, however.

napari/_qt/qt_layerlist.py
```diff
@@ -511,7 +511,8 @@ def __init__(self, layer):
         textbox = QLineEdit(self)
         textbox.setText(layer.name)
         textbox.home(False)
-        textbox.setToolTip('Layer name')
+        # Chris Wood changed from 'Layer Name' 2020/05/22
```
tlambert03:
```suggestion
```
can remove... this is appropriately captured in the git commit history

napari/_qt/qt_layerlist.py
```diff
@@ -557,6 +558,8 @@ def changeVisible(self, state):
     def changeText(self):
         """Update layer name attribute using layer name textbox contents."""
         self.layer.name = self.nameTextBox.text()
+        # set tool tip to layer name, Chris Wood 20202/05/22"""
```
tlambert03:
```suggestion
```
same... can remove

tlambert03 (approved):
good point!  This works for me then!  (can just remove the comments)

napari/_qt/qt_layerlist.py
cwood1967: Ok, do I need to remove those and then re-commit?

tlambert03: you can also just click the "commit suggestion" button

tlambert03: (alternatively, you can delete them in your local branch, commit, then push... it will update here)

sofroniewn (approved):
Look's good to me, and reasoning makes sense. Thanks for the contribution

PR #1274, opened 2020-05-23
Fix adding points with new properties 

kevinyamauchi:
# Description
This PR fixes a big where adding points with a previously-unused property value when setting the face or edge color by a color cycle on a property caused an error. The error due to the color being added to the color cycle color map with the wrong shape.

## Type of change

# References

# How has this been tested?
- [x] I added a test for adding a point with a new property and using a color cycle for face/edge color

## Final checklist:


sofroniewn (approved):
LGTM! Nice catch, and thanks for adding a test. Will merge promptly.

PR #1275, opened 2020-05-23
Pin napari-svg to 0.1.2

kevinyamauchi:
I would like to temporarily pin napari-svg to the current version. As discussed in #1248 , we need to update `napari-svg` to reflect the changes we are making to shapes (namely dropping per-shape opacity and moving the color data from the individual shape models). It seems like the best way to do this is to pin the version of `napari-svg` for now so we can update it and then test it in #1248. See [this comment](https://github.com/napari/napari/pull/1248#issuecomment-631286707) in #1248.

### DIFF
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -16,7 +16,7 @@ wrapt>=1.11.1
 psutil>=5.0
 numpydoc>=0.9.2
 napari-plugin-engine>=0.1.5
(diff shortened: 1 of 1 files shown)


sofroniewn (approved):
Yup, agreed this is the best approach. I will merge once tests finish passing. Thanks

Issue #1272, opened 2020-05-23
Hookspecs for layer keybindings


sofroniewn:
## 🚀 Feature
Plugin developers are beginning to use our reader/ writer hookspecs with success, but are looking to add additional functionality like custom key bindings, mouse functions, or GUI elements (see https://github.com/hci-unihd/covid-if-annotations/issues/3).

I propose that the next hookspecs we add are for our key bindings for our layers. We can then grow to other hookspecs after.

## Motivation

I think we should continue with our functional approach, and our approach of not providing "napari" objects to plugins but instead providing things like layer data tuples, e.g. (data, meta, layer_type). The advantage of this approach is that it enforces a "loose coupling" with napari, our plugins can't just do anything like nuke the whole viewer or layer, and it gives us a lot of control over what a plugin does as napari calls the plugin code. The downsides are some reductions in flexibility, and potentially performance, but I think they are well worth it, if they are even really there.

## Pitch

It's unclear if hookspecs are the natural thing for what I have in mind, but I think each plugin should be able to provide one dictionary per layer type. The keys of that dictionary need to valid key press combinations for napari, and the values of that dictionary need to be functions with the following signature:

```python
image_key_bindings_func(data: Any, meta: dict) -> (Any, dict)
```

Note that this is different from the API of our current keybindings functions which take in a layer and can then do anything. We could keep that API around or eventually deprecate it for this one.

How these keybindings would work is that when the key was pressed we'd generate the (data, meta) tuple pass those to the function, and then get back a (new_data, new_meta) tuple. If new_data is not None we'd set `layer.data=new_data`, and then we'd go through every element of `new_meta` and update the layer property accordingly. This is a massive reduction in what keybindings can do, but I think it will actually make them much easier to understand and make this whole system more robust.

One question would be do we "copy" the data when we pass it to the keybindings function, I'm not sure if we'd want to do that as it might trigger some big delays, but it might make the whole system much more "safe" as no in place modifications would be possible.

Curious what do @jni @tlambert03 @kne42 think of this approach? and how to best integrate it with the napari_plugin_engine? We could add a special "hookspec" where people are able to provide dictionaries instead of functions.

An interesting exercise would be how many of our current layer keybindings could be rewritten to conform to this API? Unfortunately not all of them as we have keybindings do things like change `mode` (like `painting`/ `pan_zoom`) amongst other things, which aren't inside our `meta` dict, but maybe we could figure out how to fix that.

sofroniewn:
[...]

PR #1276, opened 2020-05-24
World Coordinates

sofroniewn:
# Description
This PR begins to add support for world coordinates, as described in #763 and other places.

It adds two properties to a layer `_data_range` and `_world_range` (both of which can be renamed), which provide the range of the layer in data coordinates and world coordinates respectively and leverages to `layer._transforms` to do this conversion.

The big question now is how these properties should interact with or live on the dims model? 

Basically now we're in a somewhat awkward position where some of the necessary info lives on the layer, then some lives on the layer dims model. We also have the dims model that lives on the viewer. The likely resolution of some of that strangeness might come with the layer groups in #1089, where we could just have a dims model for each layer/ layer group and clear ways to composite across them, but for now we'll need to decide where these properties should go.

One approach would be to put everything on the dims model, including the transforms. The other could be to try and move away from having the dims model on each layer all together. Alternatively we could really try and separate concerns between them.

I'll be working on this more tomorrow, but any early input from @jni or @tlambert03 would be welcome.

## Type of change

# References

# How has this been tested?

## Final checklist:

### DIFF
--- a/examples/custom_mouse_functions.py
+++ b/examples/custom_mouse_functions.py
@@ -19,7 +19,7 @@ with napari.gui_qt():
     @viewer.mouse_drag_callbacks.append
     def get_ndisplay(viewer, event):
         if 'Alt' in event.modifiers:
-            print('viewer display ', viewer.dims.ndisplay)
+            print('viewer display ', viewer.layers.dims.ndisplay)
 
     @labels_layer.mouse_drag_callbacks.append
     def get_connected_component_shape(layer, event):
@@ -39,7 +39,9 @@ with napari.gui_qt():
                 data[binary_new] = val
             size = np.sum(binary_new)
             layer.data = data
-            msg = f'clicked at {cords} on blob {val} which is now {size} pixels'
+            msg = (
+                f'clicked at {cords} on blob {val} which is now {size} pixels'
+            )
         else:
             msg = f'clicked at {cords} on background which is ignored'
         layer.status = msg
--- a/examples/nD_vectors.py
+++ b/examples/nD_vectors.py
@@ -47,4 +47,4 @@ with napari.gui_qt():
         vectors, edge_width=0.4, name='sliced vectors', edge_color='blue'
     )
 
-    viewer.dims.ndisplay = 3
+    viewer.layers.dims.ndisplay = 3
--- a/examples/swap_dims.py
+++ b/examples/swap_dims.py
@@ -33,4 +33,4 @@ with napari.gui_qt():
         points, size=[0, 6, 10, 10], face_color='blue', n_dimensional=True
     )
 
-    viewer.dims.order = (0, 2, 1, 3)
+    viewer.layers.dims.order = (0, 2, 1, 3)
--- a/napari/_qt/_tests/test_qt_dims.py
+++ b/napari/_qt/_tests/test_qt_dims.py
@@ -132,17 +132,17 @@ def test_slider_values(qtbot):
     # Check that values of the dimension slider matches the values of the
     # dims point at initialization
     first_slider = view.slider_widgets[0].slider
-    assert first_slider.value() == view.dims.point[0]
+    assert first_slider.value() == view.dims.step[0]
 
     # Check that values of the dimension slider matches the values of the
     # dims point after the point has been moved within the dims
-    view.dims.set_point(0, 2)
-    assert first_slider.value() == view.dims.point[0]
+    view.dims.set_step(0, 2)
+    assert first_slider.value() == view.dims.step[0]
 
     # Check that values of the dimension slider matches the values of the
     # dims point after the point has been moved within the slider
     first_slider.setValue(1)
-    assert first_slider.value() == view.dims.point[0]
+    assert first_slider.value() == view.dims.step[0]
 
 
 def test_slider_range(qtbot):
@@ -156,20 +156,16 @@ def test_slider_range(qtbot):
     # Check the range of slider matches the values of the range of the dims
     # at initialization
     first_slider = view.slider_widgets[0].slider
-    assert first_slider.minimum() == view.dims.range[0][0]
-    assert (
-        first_slider.maximum() == view.dims.range[0][1] - view.dims.range[0][2]
-    )
-    assert first_slider.singleStep() == view.dims.range[0][2]
+    assert first_slider.minimum() == 0
+    assert first_slider.maximum() == view.dims.nsteps[0]
+    assert first_slider.singleStep() == 1
 
     # Check the range of slider stays matched to the values of the range of
     # the dims
     view.dims.set_range(0, (1, 5, 2))
-    assert first_slider.minimum() == view.dims.range[0][0]
-    assert (
-        first_slider.maximum() == view.dims.range[0][1] - view.dims.range[0][2]
-    )
-    assert first_slider.singleStep() == view.dims.range[0][2]
+    assert first_slider.minimum() == 0
+    assert first_slider.maximum() == view.dims.nsteps[0]
+    assert first_slider.singleStep() == 1
 
 
 def test_singleton_dims(qtbot):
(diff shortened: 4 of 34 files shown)


d-v-b:
Great to see this moving forward! A couple of comments:

* I'm not sure what "data coordinates" are -- for image data, are these array indices? If so, since the size of the data array is already represented by it's shape property, I'm not sure what the utility is of representing the same thing twice.  Can you elaborate on this?

* Separation of concerns: I think it makes sense to define the mapping between the data and some coordinate space via a single object. Neuroglancer uses this pattern to construct a layer (very analogous to a layer in napari): neuroglancer [constructs a layer](https://github.com/google/neuroglancer/blob/master/python/examples/example.py#L42) with data (an array), and a [`CoordinateSpace`](https://github.com/google/neuroglancer/blob/3c574d2dc6e9ffedd3565cf138c3029f63c17de6/python/neuroglancer/viewer_state.py#L147) object, which fully describes how the the data is embedded into world coordinates. Similarly, an [xarray.DataArray](http://xarray.pydata.org/en/stable/generated/xarray.DataArray.html) is constructed with data (an array) and a `coords` object, which assigns a point to every array index in data, thereby mapping array indices to positions in some space.

Like you suggest, the natural locus for "object that maps array indices to positions in space" in napari is the `dims` object. But I think currently `dims` is trying to do two things at once -- it tries to account for properties of the data (the number of dimensions, the number of samples along each axis) while also tracking properties of the current rendering of the data (the current slice, the display mode, the axis display order) Via this second set of properties, `dims` is basically doing the work of a geometric sampling plane. It might be necessary to factor out these two distinct functions into separate objects. 

I'm interested to hear what other people think! This is exciting!

jni:
I like all your suggestions @d-v-b!

sofroniewn:
> I'm not sure what "data coordinates" are -- for image data, are these array indices? If so, since the size of the data array is already represented by it's shape property, I'm not sure what the utility is of representing the same thing twice. Can you elaborate on this?

For image data that is the mostly case, though if the data is `rgb` then it doesn't include the final dim, and if the data is multiscale then it just takes the highest resolution. Right now we need a concept like to achieve standardization with the other layers like points and shapes. In the case of points and shapes though one could argue that there should be no scale and translate properties on the layer and that data should already come in "world" coordinates. I'm open to revisiting this @jni

> But I think currently dims is trying to do two things at once -- it tries to account for properties of the data (the number of dimensions, the number of samples along each axis) while also tracking properties of the current rendering of the data (the current slice, the display mode, the axis display order) Via this second set of properties, dims is basically doing the work of a geometric sampling plane. It might be necessary to factor out these two distinct functions into separate objects.

Indeed well put - this is the source of a lot of the awkwardness right now. Right if there are three things - data, spatial information, and slicing, then we sort of awkwardly mix these things all together, but it is probably cleanest if data + spatial information live nicely together and then slicing can do its own thing on this object.

Note that @pwinston is thinking about some big improvements in #845 that could result in a new tile fetching system, but my sense @jni @royerloic is that we'd want that to live more on the "view" side of things rather than the "model" side of things, right? Maybe riffing off of some discussion in #1272 a good question to start asking ourselves is what would we like a plugin developer to have access to? Certainly they need the data + spatial information, but they might not need the slicing (or at least slicing could be exposed in a more advanced interface).

>  I think it makes sense to define the mapping between the data and some coordinate space via a single object. Neuroglancer uses this pattern to construct a layer (very analogous to a layer in napari): neuroglancer constructs a layer with data (an array), and a CoordinateSpace object, which fully describes how the the data is embedded into world coordinates. Similarly, an xarray.DataArray is constructed with data (an array) and a coords object, which assigns a point to every array index in data, thereby mapping array indices to positions in some space.

@jni do we want to bite the bullet, depend on xarray, and construct an xarray representation of data with coordinates internally? I'd imagine our public API remains unchanged - we still accept bare numpy arrays, translation, scale parameters, but then we construct an xarray internally where we have all these things nicely. We could either just put that xarray privately somewhere like `._coordinate_array` and `.data` returns whatever they gave us, or just make it so that `.data` always returns an xarray regardless of what was passed? If people pass us xarray's that conform to our pattern we just use it directly, if they pass us xarray's that have a different structure, i guess we'll need to adapt it to our own xarray conventions? As long as we don't copy data, force instantiation or clober peoples own metadata this could all be ok, but it might get a little crazy, thinking about zarr, dask etc too.

On option would be to always convert what is given to being an xarray, keep track of if an xarray was not given, then if an xarray was not given pull the array out of the xarray for the `.data` property, otherwise give the `xarray` back. In this model you'd have numpy in numpy out, xarray in xarray out, always xarray on the inside and only one thing on the inside.

Is this how you were thinking about it @d-v-b?

sofroniewn:
Following discussion with @jni i'm going to press on with what will hopefully be a fairly minimal feature addition, but more refactoring and improvements to this will hopefully becoming soon.

sofroniewn:
Progress with the minimal approach is coming along well, the coordinate transform inverse going back from world to data is now in place on the base layer.

I still have some stuff to do on the Qt slider side to get all the float stuff working, and we'll have to make some choices about what labels to appear next to the sliders, (i.e. right now you see the current slider position followed by the max slider position)

Screen Shot 2020-05-25 at 10 14 20 PM

Once these sliders are in a "world coordinates" I guess one might be interested in the minimum value, the maximum value, the step size, and the current value? Need to think about this more, and open to suggestions from all

tlambert03:
fwiw, I was definitely hoping that a move to world coordinates would _not_ have to change those labels there.  When I think about indexing a dataset (as we do with those sliders), I definitely think in terms of plane number pixel number, and I want the program to convert that to world coordinates/scales for me behind the scenes.  (I'd also like to be able to see distances/lengths _somewhere_ in world units, of course, but not on the sliders)

d-v-b:
I think eventually the sliders should show *both* array indices and world coordinates. As soon as you have data at multiple resolutions, you can't think exclusively in terms of plane number / pixel number any more. This was actually why I started using xarray in the first place: I had raw EM data at 4nm resolution, and ground-truth data that had been upscaled to 2nm resolution AND padded. Compared to manually tracking the offset + scaling conversion, it was much much easier to just assign coordinates to these datasets, make xarray.DataArrays, and query the data with arguments defined in the common coordinate system (nanometers). xarray took me from doing error-prone index arithmetic to `data.sel({'x': x_range_in_nm, 'y': y_range_in_nm, 'z' : z_range_in_nm})`, which is a completely uniform interface for both datasets.

> I definitely think in terms of plane number pixel number, and I want the program to convert that to world coordinates/scales for me behind the scenes.

I actually think the opposite -- I'd prefer that napari assume that all images are defined over some kind of space, and internally dispatch to different sampling methods depending on the source of the data (array indexing for arrays, something else for vectors, etc). Plane number / pixel number (i.e., array indices) is an implementation detail of data acquired in a certain way (gridded sampling) and represented internally as an array, but outside of light microscopy these images are actually a special case: e.g., array indices don't make a lot of sense for vector data, or data sampled with a spiral pattern.

> On option would be to always convert what is given to being an xarray, keep track of if an xarray was not given, then if an xarray was not given pull the array out of the xarray for the .data property, otherwise give the xarray back. In this model you'd have numpy in numpy out, xarray in xarray out, always xarray on the inside and only one thing on the inside.

Yeah this is roughly what I was thinking. The user does not need to know anything about xarray, and they can always get their data out the way they put it in, but xarray is used internally to track coordinates. An alternative approach would be to implement an xarray-compatible interface / protocol without having xarray as a dependency, but I'm not sure what would keep the APIs from drifting apart.

tlambert03:
> I think eventually the sliders should show both array indices and world coordinates

fine with me, as long as image indices are clearly available somewhere

sofroniewn:
@d-v-b @tlambert03 thanks both for weighing in. Sounds like people will want to see both representations, and to move back and forth between them, fairly easily. Will keep you posted!

d-v-b:
I have a really basic API question -- what is the difference between the `dims` attribute of a layer vs the `dims` attribute of the viewer?

sofroniewn:
> I have a really basic API question -- what is the difference between the dims attribute of a layer vs the dims attribute of the viewer?

The dims on the viewer has the full nD representation of all the layers and is linked to the sliders on the screen. The dims on the layers have the same dim as the layer itself and are then used for the slicing. 

There has been some discussion about removing `dims` from a layer (or at least making it private). This PR experimentally puts a dims on the layerlist rather than the viewer, but we'll see if that's a step in the right direction.

The alternative after #1040 is done is that we can maybe move the slicing logic off the layer itself and then drop the additional dims representations.

Czaki:
It will be nice if this PR fix exception on change spacing of layer like:

```
IndexError: index 44 is out of bounds for axis 1 with size 44
  File "D:\a\1\s\package\PartSeg\common_gui\napari_image_view.py", line 239, in update_spacing_info
  File "site-packages\napari\layers\base\base.py", line 336, in scale
  File "site-packages\napari\layers\base\base.py", line 403, in _update_dims
  File "site-packages\napari\layers\base\base.py", line 633, in refresh
  File "site-packages\napari\layers\base\base.py", line 586, in set_view_slice
  File "site-packages\napari\layers\image\image.py", line 527, in _set_view_slice
```

sofroniewn:
Superseded by #1539

PR #1279, opened 2020-05-24
support viewer annotation in magicgui

tlambert03:
# Description
This registers `napari.Viewer` as a type that `magicgui` can handle, and results from discussion in https://github.com/hci-unihd/covid-if-annotations/issues/3.  If an argument is annotated as a `napari.Viewer`, then if the magicgui is docked in a viewer, it will return the current viewer, otherwise it returns a list of all global viewer instances (probably won't be used much).  It's slightly weird to show the viewer instance as the sole item in a dropdown, but combined with https://github.com/napari/magicgui/pull/17, it will make it easier to have a viewer object as a parameter in a magicgui function.

## Type of change

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/utils/_magicgui.py
+++ b/napari/utils/_magicgui.py
@@ -9,8 +9,10 @@ end-user annotates one of their function arguments with a type hint using one
 of those custom classes, magicgui will know what to do with it.
 
 """
-from typing import Tuple, Type, Any
+from typing import Any, Tuple, Type
+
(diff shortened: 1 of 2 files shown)


sofroniewn (approved):
This seems fine with me for now, agreed it's a little odd in the sense that we only have one viewer most of the time, but it's consistent with how we treat the layers. We can always revisit in the future.

PR #1281, opened 2020-05-25
Adding properties attribute to Labels layers

### LINKED ISSUE #1214: Add properties attribute to Labels layers
## 🚀 Feature
As is the case with Points layers, it would be nice to have the option to add properties to Labels e.g. the class the label belongs to.

## Motivation

When viewing labels layers it's common for each label to have a classification and it would be great to see this information when hovering over a pixel without referring to an external source. You could also display associated information such as the confidence of the classifier for that point. This would make the editing labels process even smoother.

DragaDoncila:
# Description
Initial work on #1214 , adding properties dictionary attribute to Labels, and a label_index dictionary to map each label to the corresponding row in properties.

## Type of change

# References
Fixes #1214 

# How has this been tested?

## Final checklist:

### DIFF
--- a/examples/add_labels_with_properties.py
+++ b/examples/add_labels_with_properties.py
@@ -1,20 +1,49 @@
 """
-Display a labels layer with classes for each label, including the background
+Display a labels layer with various properties
 """
 
-import napari
-import numpy as np
-from skimage import data
-from scipy import ndimage as ndi
-from napari.layers import Labels
 
-blobs = data.binary_blobs(length=128, volume_fraction=0.1, n_dim=2)
-labelled, num_labels = ndi.label(blobs)
-class_dict = {
-    'class': ["Background"]
-    + ["Class " + str(i + 1) for i in range(num_labels)]
-}
+from skimage import data
+from skimage.filters import threshold_otsu
+from skimage.segmentation import clear_border
+from skimage.measure import label
+from skimage.morphology import closing, square, remove_small_objects
+import numpy as np
+import napari
+
 
 with napari.gui_qt():
-    viewer = napari.Viewer()
-    viewer.add_labels(labelled, properties=class_dict)
+    image = data.coins()[50:-50, 50:-50]
+
+    # apply threshold
+    thresh = threshold_otsu(image)
+    bw = closing(image > thresh, square(4))
+
+    # remove artifacts connected to image border
+    cleared = remove_small_objects(clear_border(bw), 20)
+
+    # label image regions
+    label_image = label(cleared)
+
+    # initialise viewer with coins image
+    viewer = napari.view_image(image, name='coins', rgb=False)
+
+    # get the size of each coin (first element is background area)
+    label_areas = np.bincount(label_image.ravel())[1:]
+
+    # split coins into small or large
+    size_range = max(label_areas) - min(label_areas)
+    small_threshold = min(label_areas) + (size_range / 2)
+    coin_sizes = np.where(label_areas > small_threshold, 'large', 'small')
+
+    label_properties = {
+        'row': ['none']
+        + ['top'] * 4
+        + ['bottom'] * 4,  # background is row: none
+        'size': ['none'] + list(coin_sizes),  # background is size: none
+    }
+
+    # add the labels
+    label_layer = viewer.add_labels(
+        label_image, name='segmentation', properties=label_properties
+    )
--- a/examples/add_labels_with_properties.py
+++ b/examples/add_labels_with_properties.py
@@ -1,0 +1,20 @@
+"""
+Display a labels layer with classes for each label, including the background
+"""
+
+import napari
+import numpy as np
+from skimage import data
+from scipy import ndimage as ndi
+from napari.layers import Labels
+
+blobs = data.binary_blobs(length=128, volume_fraction=0.1, n_dim=2)
+labelled, num_labels = ndi.label(blobs)
+class_dict = {
+    'class': ["Background"]
+    + ["Class " + str(i + 1) for i in range(num_labels)]
+}
+
+with napari.gui_qt():
+    viewer = napari.Viewer()
+    viewer.add_labels(labelled, properties=class_dict)
--- a/napari/components/add_layers_mixin.py
+++ b/napari/components/add_layers_mixin.py
@@ -406,7 +406,6 @@ class AddLayersMixin:
         *,
         num_colors=50,
         properties=None,
-        label_index=None,
         seed=0.5,
         name=None,
         metadata=None,
@@ -440,10 +439,8 @@ class AddLayersMixin:
             Number of unique colors to use in colormap.
         properties : dict {str: array (N,)}, DataFrame
             Properties for each label. Each property should be an array of length
-            N, where N is the number of labels.
-        label_index : dict {int: int}
-            Dictionary mapping labels (arbitrary integers) to row indices
-            (sequential integers in 0..N not inclusive).
+            N, where N is the number of labels, and the first property corresponds to
+            background.
         seed : float
             Seed for colormap random generator.
         name : str
@@ -478,7 +475,6 @@ class AddLayersMixin:
             data,
             num_colors=num_colors,
             properties=properties,
-            label_index=label_index,
             seed=seed,
             name=name,
             metadata=metadata,
--- a/napari/components/add_layers_mixin.py
+++ b/napari/components/add_layers_mixin.py
@@ -438,6 +438,12 @@ class AddLayersMixin:
             Labels data as an array or multiscale.
         num_colors : int
             Number of unique colors to use in colormap.
+        properties : dict {str: array (N,)}, DataFrame
+            Properties for each label. Each property should be an array of length
+            N, where N is the number of labels.
+        label_index : dict {int: int}
+            Dictionary mapping labels (arbitrary integers) to row indices
+            (sequential integers in 0..N not inclusive).
         seed : float
             Seed for colormap random generator.
         name : str
--- a/napari/components/add_layers_mixin.py
+++ b/napari/components/add_layers_mixin.py
@@ -405,6 +405,8 @@ class AddLayersMixin:
         data,
         *,
         num_colors=50,
+        properties=None,
+        label_index=None,
         seed=0.5,
         name=None,
         metadata=None,
@@ -469,6 +471,8 @@ class AddLayersMixin:
         layer = layers.Labels(
             data,
             num_colors=num_colors,
+            properties=properties,
+            label_index=label_index,
             seed=seed,
             name=name,
             metadata=metadata,
--- a/napari/layers/labels/labels.py
+++ b/napari/layers/labels/labels.py
@@ -154,7 +154,7 @@ class Labels(Image):
             if len(props) > 0:
                 arbitrary_key = list(props.keys())[0]
                 self._label_index = {
-                    i + 1: i for i in range(len(props[arbitrary_key]))
+                    i: i for i in range(len(props[arbitrary_key]))
                 }
             else:
                 self._label_index = {}
@@ -629,8 +629,11 @@ class Labels(Image):
         # if this labels layer has properties
         if self._label_index and self._properties:
             # if the cursor is not outside the image or on the background
-            if self._value is not None and self._value != 0:
-                idx = self._label_index[self._value]
-                for k, v in self._properties.items():
-                    msg += f' {k}: {v[idx]}'
+            if self._value is not None:
+                if self._label_index.get(self._value) is not None:
+                    idx = self._label_index[self._value]
+                    for k, v in self._properties.items():
+                        msg += f' {k}: {v[idx]}'
+                else:
+                    msg += f' [No Properties]'
         return msg
--- a/napari/layers/labels/labels.py
+++ b/napari/layers/labels/labels.py
@@ -630,7 +630,7 @@ class Labels(Image):
         if self._label_index and self._properties:
             # if the cursor is not outside the image or on the background
             if self._value is not None:
-                if self._label_index.get(self._value) is not None:
+                if self._value in self._label_index:
                     idx = self._label_index[self._value]
                     for k, v in self._properties.items():
                         msg += f' {k}: {v[idx]}'
--- a/napari/layers/labels/labels.py
+++ b/napari/layers/labels/labels.py
@@ -1,5 +1,5 @@
 from collections import deque
-from typing import Union
+from typing import Union, Dict
 
 import numpy as np
 from scipy import ndimage as ndi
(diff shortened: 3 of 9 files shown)


jni:
@napari/core-devs we have some design choices to make here. First up: right now, we have a hardcoded convention that 0 is background. I think we should keep that, but then the question is, do people pass in properties for the background label? yes, no, or optional? If no, does that mean that they need to pass in a list with an extra "dummy" element in front for the 0-label? Or that, in the absence of an index, we automatically shift the non-zero labels down by 1? example:

```python
labels = ...  # volume with integer labels 1, 2, 3, 0 is background
viewer = napari.add_labels(
    labels, properties={'class': ['membrane', 'cyto', 'mito']}
)  # requires mapping 1 to 0, 2 to 1, 3 to 2
# or:
viewer = napari.add_labels(
    labels, properties={'class': [None, 'membrane', 'cyto', 'mito']}
)  # no remapping, but user has to provide None
```

jni:
Having written it out, the remapping seems way better. :joy:

sofroniewn:
>  First up: right now, we have a hardcoded convention that 0 is background. I think we should keep that

I know @perlman a while ago advocated for being more flexible here, including letting `-1` or potentially any label be background, and I think he implied some EM datasets out there use `-1`. Note we added a `self._background_label ` in #1264 in preparation for this.

I sort of like "no remapping" but with the option to actually put a string there, not just have `None`, so I could have

```python
viewer = napari.add_labels(
    labels, properties={'class': ['extracellular space', 'membrane', 'cyto', 'mito']}
)  # no remapping
```

if I wanted where `extracellular space` corresponds to `0` and is transparent. I just worry that otherwise I'll have all sorts of `-1` everywhere "with remapping"

```python
layer.propterties['class'][self.selected_label - 1] # with remapping
layer.propterties['class'][self.selected_label] # no remapping
```

DragaDoncila:
> just worry that otherwise I'll have all sorts of -1 everywhere "with remapping"

The remapping works so that the `label_index` is offset, so you don't have to remember to subtract the 1.

I understand the main point though which is that someone might wish to provide background properties. It should be noted the remapping only occurs if `label_index` is not provided. So if you wanted the background to be labelled 'extracellular space', you could do it like so

```
viewer = napari.add_labels(
    labels, properties={'class': ['extracellular space', 'membrane', 'cyto', 'mito']},
    label_index={0:0, 1:1, 2:2, 3:3}
)  # no remapping
```
although I take the point that seems SUPER redundant...

I guess the question is can we rely on users to explicitly pass through a class for the background?

DragaDoncila:
The more I think of it the more I feel like it should be done without remapping. But that does also beg the question of what happens if the user changes the background_label before/after passing through a properties dict.

Do we assume the first property always belongs to the background label, whatever that label might be? If the background label value changes, do we update our label_index?

jni:
Ok, so if we do want to allow remapping of the background (and to be clear I hate this =P) (but acknowledge the utility), then we open a massive can of worms:

- @sofroniewn string is actually better than None in this case, in terms of being easier to handle, but
- do we then give users control over whether they want special treatment for the background?

Actually, how's this:
- by default, we don't do remapping. Users need to either provide properties for the background, or
- they need to provide the remapping themselves, and
- in `get_message`, any label that isn't mapped in `label_index` gets ignored (silently or we add e.g. `' [no properties]'` when there's no index?)

sofroniewn:
That sounds reasonable to me @jni @DragaDoncila. We shouldn't crash if no label property is mapped to an index. 

We can always encourage people to use a string like `unlabeled`, `transparent`, or `background` for the `_background_label`.

Maybe after using for a while we can revisit if we feel like we've gotten this wrong. I can certainly see it both ways. Also as we get more real world examples like from @constantinpape maybe the most natural behavior will become clear.

constantinpape:
> Maybe after using for a while we can revisit if we feel like we've gotten this wrong. I can certainly see it both ways. Also as we get more real world examples like from @constantinpape maybe the most natural behavior will become clear.

Just to give you some perspective on ideas I have about this:
The properties would be a very good fit to store information for semantic instance segmentation problems, i.e. problems where there is an instance cell (or other objects) segmentation + categorical labels associated with the cells.
These categorical labels could be associated with attributes and then one could  switch between viewing the instance segmentation and the semantic segmentation in the same layer, e.g. using a short-cut.

DragaDoncila:
@jni @sofroniewn The latest update has taken Juan's approach.

>Actually, how's this:
> - by default, we don't do remapping. Users need to either provide properties for the background, or
> - they need to provide the remapping themselves, and
> - in get_message, any label that isn't mapped in label_index gets ignored (silently or we add e.g. ' [no properties]' when there's no index?)

Rather than silently ignoring label indices with no properties, we display "[No Properties]" next to the label. I don't think silently ignoring is the best option because in the case where you meant to leave out some indices, it doesn't degrade from your experience, but if you thought you'd provided properties for everything then it's an explicit indicator that you haven't.

jni:
@tlambert03 @sofroniewn the timing tests are failing on PySide2/Linux, see:

https://cirrus-ci.com/task/4691261632282624?command=test_pyside2#L1114

this is the failing test:

https://github.com/napari/napari/blob/47c8a53d71096538e222908ef0330b5ae4eb3ea6/napari/layers/image/_tests/test_big_image_timing.py#L35-L39

and this one for the zarr array only, not dask:

https://github.com/napari/napari/blob/47c8a53d71096538e222908ef0330b5ae4eb3ea6/napari/layers/image/_tests/test_big_image_timing.py#L21-L25

Do either of you have any idea why this would happen?

sofroniewn:
> @tlambert03 @sofroniewn the timing tests are failing on PySide2/Linux, see:

Strange, i am rerunning, seeing if it is reproducible. nothing jumps out in this PR, and the tests don't seem to have failed elsewhere (they just ran on master after the last merge and passed ...)

napari/components/add_layers_mixin.py
```diff
@@ -436,6 +438,12 @@ def add_labels(
             Labels data as an array or multiscale.
         num_colors : int
             Number of unique colors to use in colormap.
+        properties : dict {str: array (N,)}, DataFrame
+            Properties for each label. Each property should be an array of length
+            N, where N is the number of labels.
+        label_index : dict {int: int}
+            Dictionary mapping labels (arbitrary integers) to row indices
+            (sequential integers in 0..N not inclusive).
```
jni: Add a disclaimer here like: "A label value not found in the index keys will be considered as having no properties associated with it."

napari/layers/labels/labels.py
```diff
@@ -551,3 +623,17 @@ def paint(self, coord, new_label, refresh=True):
...
         if refresh is True:
             self.refresh()
+
+    def get_message(self):
+        msg = super().get_message()
+        # if this labels layer has properties
+        if self._label_index and self._properties:
+            # if the cursor is not outside the image or on the background
+            if self._value is not None:
+                if self._label_index.get(self._value) is not None:
```
jni:
```suggestion
                if self._value in self._label_index:
```

napari/components/add_layers_mixin.py
sofroniewn: Does every row need to be mapped to? If so then maybe an idea is to have `label_index` be a length `N` list where each value corresponds to a label. We could even construct this dict from that list if we wanted faster indexing/ convenience. So you'd have

```python
[1, 10, 100] # equivalent to
{'1': 0, '10':1, '100':2}
```

Feel free to disregard

sofroniewn: Going further, could you actually then just have `label_index` as a special column within `properties`? Basically each row of properties would tell us which label it wanted to correspond to, and if no special column provided we just do default mapping.

jni:
Ok so this is now at the stage where it's not breaking anything! 🎉 ! 😝 

I think what's left is testing some of the new functionality:

- creating a labels layer with properties but no index shouldn't break
- creating a labels layer with a label index but no properties — dunno what that *should* do but it shouldn't be "explode in a fire"
- creating a labels layer with properties and an incomplete index should work gracefully (I think that's covered right now)
- creating a labels layer with properties and an index *and then adding a new label* shouldn't break things.

Another additional feature would be a quick way to create an off-by-one mapping as desired when the background doesn't need properties. However that *could* come in a separate PR. It's important that the base functionality is rock solid before expanding to new stuff.

Regarding adding new labels, *eventually*, we could add current properties etc. as in the points layer. However, I suggest leaving this for a future PR, as this is already useful as is (in many use cases there are never going to be additional labels).

Thoughts @sofroniewn @kevinyamauchi?

napari/components/add_layers_mixin.py
jni: @sofroniewn well, the thing is we need the dictionary for fast lookup. So we'll have to create the attribute anyway. However, your suggestion actually makes a lot of sense in terms of reducing the API. My preference would be for the column to be called `index`, as that is the default in pandas if you do `reset_index()` with an unnamed index.

We would also need to update `dataframe_to_properties` to include this convention: if there is no column named index *and* the index has type "int", use the pandas index.

sofroniewn: Yeah, I think we'd pay the price of creating that (private) attribute, but for the gain of less API. I like a convention using the word `index`, and think that's a reasonable update to `dataframe_to_properties`

We'd then have to decide for our defaults would the first row map to background? With this API I think not. We could have it map to `1` by default. But again, open to ideas/ suggestions

sofroniewn:
> Regarding adding new labels, eventually, we could add current properties etc. as in the points layer. However, I suggest leaving this for a future PR, as this is already useful as is (in many use cases there are never going to be additional labels).

combined with

> creating a labels layer with properties and an index and then adding a new label shouldn't break things.

Seems good!! I'll play with tomorrow

napari/components/add_layers_mixin.py
jni: Let's go with 0 default. Corresponds to NumPy indexing.

napari/layers/labels/labels.py
DragaDoncila: So the reason I went with the get method is I'm not sure whether `k in dict` searches effectively, whereas I'm pretty sure `dict.get(k)` is O(1). Happy to change it though, just thought these dicts might get quite big

napari/components/add_layers_mixin.py
DragaDoncila: OK so just double checking the upshot from this conversation:

- Make `label_index` private, rename it to `index`
- When being passed a `properties` dict, check if it has an `index` column
- If it does, pull it into our `index` dict
- If it doesn't, use the pandas index column

> We would also need to update dataframe_to_properties to include this convention: if there is no column named index and the index has type "int", use the pandas index.

@jni I'm a bit confused by this. if there is no column named index how are we checking it's type int? Did you mean to say if there IS an index and it's 0..N, we just use pandas index?

napari/layers/labels/labels.py
DragaDoncila: Python docs confirm `k in dict` is optimised.

DragaDoncila:
Tests are all passing locally, I believe the current failure is the same transient failure from above @sofroniewn @jni ?

Otherwise I believe everything has been addressed. Adding a new label (with a previously unseen index) just defaults to showing '[No Properties]' in the status message, but we will need to provide a way for the user to assign properties to this new label.

napari/layers/labels/labels.py
```diff
@@ -127,6 +138,22 @@ def __init__(
         self._num_colors = num_colors
         colormap = ('random', colormaps.label_colormap(self.num_colors))
 
+        if properties is None:
+            self._properties = {}
+            self._property_choices = {}
```
sofroniewn: Do we need to create `_property_choices` here?

I guess might use it in the future when we start thinking about adding properties as labels get added, but for now I don't think it is needed / doesn't appear in the other arm of the `else`

sofroniewn (approved):
This is great! Love the minimal PR, very easy to review. I left a fairly minor comment, below and am approving.

One thing that might be nice to add though is a new example `add_labels_with_properties.py` we have such an example for in `add_points_with_properties.py` and it will be nice to have that example to keep coming back to as we start adding more complex functionality around properties (like linking colormaps off them etc).

napari/layers/labels/labels.py
DragaDoncila: >Do we need to create _property_choices here?

No, we don't. Have removed it!

DragaDoncila:
> One thing that might be nice to add though is a new example add_labels_with_properties.py we have such an example for in add_points_with_properties.py and it will be nice to have that example to keep coming back to as we start adding more complex functionality around properties (like linking colormaps off them etc).

I've added a minimal example @sofroniewn , let me know if you'd like it changed at all

examples/add_labels_with_properties.py
```diff
@@ -0,0 +1,20 @@
...
+from skimage import data
+from scipy import ndimage as ndi
+from napari.layers import Labels
+
+blobs = data.binary_blobs(length=128, volume_fraction=0.1, n_dim=2)
+labelled, num_labels = ndi.label(blobs)
+class_dict = {
+    'class': ["Background"]
+    + ["Class " + str(i + 1) for i in range(num_labels)]
+}
```
jni: This example is sort of uninteresting, because each label is its own class. What I would do here is divide blobs into large and small and then map to just two classes, "small" and "large" or "simple" and "compound", for example

sofroniewn:
> I've added a minimal example @sofroniewn , let me know if you'd like it changed at all

Do you mind basing the example off of `add_labels.py` with the coins and the segmentation - maybe giving each of the coins a one/ two word `description`, and maybe also a property like `row` with values `top` and `bottom` depending on what row the coin is in!

Having a real world example like that will make it much easier to debug things going forward (say I want to color all the `top` `row` coins `red` and the `bottom` `row` coins `blue` etc.)

examples/add_labels_with_properties.py
DragaDoncila: Yeah I thought so too but I couldn't think of anything more interesting... lol

sofroniewn: We just commented at the same time, see my suggestion below about using the coins like we do in `add_labels.py` (I also just like looking at the coins) :-)

DragaDoncila:
@sofroniewn @jni  I did a little of both your suggestions to keep the peace.

napari/layers/labels/labels.py
```diff
@@ -551,3 +614,18 @@ def paint(self, coord, new_label, refresh=True):
...
+        msg = super().get_message()
+        # if this labels layer has properties
+        if self._label_index and self._properties:
+            # if the cursor is not outside the image or on the background
+            if self._value is not None:
+                if self._value in self._label_index:
+                    idx = self._label_index[self._value]
+                    for k, v in self._properties.items():
+                        if k != 'index':
+                            msg += f' {k}: {v[idx]}'
```
jni: @DragaDoncila playing with the example has convinced me that this should be:

```suggestion
                            msg += f', {k}: {v[idx]}'
```

DragaDoncila: Yeah I did wonder. I also wondered what would happen if you had a long list of properties, are we happy to just print it all?

jni (approved):

PR #1287, opened 2020-05-26
Move from versioneer to setuptools_scm

tlambert03:
# Description
This PR welcomes feedback before review.  I just figured it would be easier to discuss with code examples that you can checkout if desired.

I'd like to propose that we switch from `versioneer` to [`setuptools_scm` ](https://pypi.org/project/setuptools-scm/) for our git-tag-based version management.  I've been using it for all external projects over the last few months and like it a lot.  It doesn't require any code to be checked into our source (see 2,500 lines removed ... which also never gets updated).  In the mode I'm using here, it writes a `_version.py` file to the repo whenever it is "pip installed" (or when running `python setup.py sdist`, for example) based on git tags (just like versioneer). 
fwiw, it also claims to be the "blessed" setuptools version management (though I can't figure out where that "blessing" comes from).

curious to get thoughts.  @kne42?  @jni?

## Type of change

# How has this been tested?

tested in many side projects with github-actions-based release/deploy.  Also tested locally with `pip install -e .` and `python setup.py sdist`

## Final checklist:


pyproject.toml
```diff
@@ -1,3 +1,9 @@
+[build-system]
```
kne42, line 1: I'm not super familiar with pyproject.toml. Does this automatically get installed when you use pip or do you need another tool?

tlambert03: gets installed: https://pip.pypa.io/en/stable/reference/pip/#pep-517-and-518-support

kne42: good to know, thanks!

tlambert03: yeah, I'm learning as I go here too.  so don't necessarily assume I know what I'm doing! 😅 would be great if you get an opportunity to pull the PR, and try to break it locally.

jni: But note also #1246...

jni:
@tlambert03 my main question is why go through this? I don't have a strong sense about this but at least [some members of the SciPy community](https://twitter.com/asmeurer/status/1110241857122885644) aren't fans. So the question is, what does it bring us? Sure, we don't check in `_version.py`, but the fact that that code has never needed updating is an *advantage* rather than a disadvantage?

So my current feeling is "if it ain't broke..."

pyproject.toml
tlambert03: this PR includes [support for legacy behavior](https://github.com/napari/napari/pull/1287/files#diff-2eeaed663bd0d25b7e608891384b7298R95-R96) and doesn't require pyproject.toml

tlambert03:
I'll note that that tweet complains about using `import pkg_resources` in their `__init__.py` which is _not_ done here (There are multiple options)... as other people in the thread point out, there is no runtime dependency on pkg_resources. 

> So my current feeling is "if it ain't broke..."

but it _is_ causing problems for me in #1289.   I'll also note that the last release to versioneer was over 3 years ago. 

> what does it bring us?

if we don't use this approach it's fine, but then I _think_ I'll need a different way to write the version string to a file that gets added to the bundle and used for `napari.__version__` ... I'm struggling to get the `get_version` method to work there.  So this brings us an immediate solution to that problem

tlambert03:
perhaps worth adding: I proposed this because it solved a problem i had in #1287, which I'll explain below.  It's not that I think it probably _can't_ be solved using versioneer... rather that it was solved easily with setuptools_scm.  If we want to punt on this decision, maybe @kne42 can help me figure out how I'm supposed to use versioneer in #1287.  Here's the issue:

- due to the way code signing works in macOS gatekeeper, it's critical that the app bundle be exactly the same as it was during code signing when the user selects "yes, ok to open".
- that means we have to have the `_qt_resources.py` file bundled in with the app, otherwise it tried to add it at first run.
- in order for it not to try to build, the version string in the `_qt_resources*.py` file must match the current version of the software... and that's the problem I'm getting with versioneer.  It's not giving me the same version when we build in CI as it does when we download the bundled app.

jni:
> but it _is_ causing problems for me in #1289.

@tlambert03 maybe you should have led with that! I totally withdraw my objection. :sweat_smile:

tlambert03:
> @tlambert03 maybe you should have led with that!

yes, I suppose that _would_ have been reasonable 😂.  Still doesn't invalidate my previous point though.  I _do_ think it's probably possible to achieve what we need with bundling in versioneer... but it wasn't immediately obvious and I just might need some help with it.  (... I _do_ definitely prefer `setuptools_scm` in general, but if there's any objections I don't mind reconsidering)

kne42 (approved):
> It's not giving me the same version when we build in CI as it does when we download the bundled app.

This is because the CI creates a shallow clone of the repository, meaning that it does not pull down the tags that versioneer uses to determine its version. You can fix that with `CIRRUS_CLONE_DEPTH: 0` as in https://github.com/napari/napari/blob/f87e18641c529f6c592553b052805c9d75231b90/.cirrus.yml#L246.

With that being said, setuptools-scm sounds much better than versioneer, especially since it appears to be more actively developed, and I would be happy to switch to it :)

PR #1288, opened 2020-05-26
Add eraser button and functionality

liu-ziyang:
# Description
Add an eraser tool to replace use case of painting with background label:

- Added an eraser button in toolbox, resized the toolbox to make the UI consistent with before. 
- Add erase function using paint to use background label
- Disabled user from using background label 
- Experiment alternate keys for temporary mode switching

## Type of change

# References
Desired behavior is discussed here: https://github.com/napari/napari/issues/1142

# How has this been tested?
- The test suite for my feature covers cases of erasing labels with different brush size
- Manually run through using the new eraser tool, as well as changing between tools

## Final checklist:


napari/layers/labels/_labels_key_bindings.py
```diff
@@ -78,11 +78,39 @@ def redo(layer):
...
     layer.preserve_labels = not layer.preserve_labels
+
+
+@Labels.bind_key('Control')
+def switch_fill(layer):
+    """Switch to fill mode temporarily when pressed."""
+    previous_mode = layer.mode
+
+    # on key press
+    layer.mode = Mode.FILL
```
liu-ziyang, line 97: alternatively, I gave it a try using double click to switch to fill, but it is quite some work, because the first click always triggers a paint, the area would be painted with new label, and the second click tries to fill the area already using the new label, hence not filling the broader area of old label as intended.

napari/_qt/layers/qt_labels_layer.py
```diff
@@ -71,15 +73,15 @@ def __init__(self, layer):
         )
 
         # shuffle colormap button
-        self.colormapUpdate = QPushButton('shuffle colors')
+        self.colormapUpdate = QPushButton('shuffle\ncolors')
```
sofroniewn: I like making the label multiline, but the pushbutton is a little too tall now, see below - any chance you can make it same height as other icons 
image

napari/layers/labels/_labels_key_bindings.py
sofroniewn: yeah, let's stay away from double click for now

sofroniewn: This also doesn't seem to work for me. There maybe some problem binding to a modifier? Unclear, i can look in more. I think @jni may have run into problems here too in another context

napari/layers/labels/_labels_mouse_bindings.py
```diff
@@ -1,12 +1,24 @@
...
-    """Paint with the currently selected label."""
+    """Paint with the proper label.
+
+    Use current selected label when in paint mode,
+    and use background label when in erase mode (or right click).
+    """
+    if layer._mode == Mode.ERASE:
+        label = layer._background_label
+    # Experimental:
+    elif hasattr(event, 'button') and event.button == 2:
```
sofroniewn: This doesn't seem to be working for me as expected.

Actually what looks like is happening for me is this is getting triggered by `control` click, not `right` click on the mac, and that the `control` keybinding as above isn't working.

Let's hold off on the `right click` for now, and maybe we have to rethink the modifier keybinding above.

A designer will be joining the team in July and she can help us with some of these UI choices then too

napari/layers/labels/labels.py
```diff
@@ -265,8 +268,8 @@ def selected_label(self):
 
     @selected_label.setter
     def selected_label(self, selected_label):
-        if selected_label < 0:
-            raise ValueError('cannot reduce selected label below 0')
+        if selected_label < 1:
```
sofroniewn: I think for now we should leave this as before as otherwise someone can't use the `fill` bucket with `0`, which is quite nice. I think though that we should keep your change of making the default `self._selected_label = 1`, this way if someone adds a new layer and starts painting they will see something right away

sofroniewn (commented):
@ziyangczi this is looking good, I ran into some problems with the modifier keys as keybindings, and the mouse right click. I also think we should still let `selected_label` be `0` if someone wants so they can use the fill bucket, but then I think we'll be pretty close.

Nice work again!!

napari/_qt/layers/qt_labels_layer.py
sofroniewn: oh also, can you make the line width of the eraser one thicker, see how it is a little thin compared to the other icons

napari/layers/labels/_labels_key_bindings.py
liu-ziyang: I was able to use control to swap the mode as expected, wonder if our mac key binding is any different? I have been swapping cmd with ctrl as well

napari/_qt/layers/qt_labels_layer.py
sofroniewn: from team meeting - @tlambert03 is going to make shuffle color a shuffle icon and make the thickness of the eraser match the thickness of the others. thanks @tlambert03!!

sofroniewn:
Following discussion in team meeting, I think the recommendation was to switch to `option` click instead of right click for rapid switching to erase. It would be great if @tlambert03 and @jni can weigh in here on the intuitiveness of the keybindings/ mouse function modifiers and with any approvals/ recommendations.

Code looks good to go, so once we agree on keybindings/ mouse function modifiers and @tlambert03 updates the icons this will be ready for merge. Thanks @ziyangczi, great work here again!

tlambert03:
just update this with new buttons.  I also changed the fill button to a paint bucket, since that had been discussed in the past:

Screen Shot 2020-05-29 at 11 24 33 AM

couple remaining thoughts after playing with it:
- right click is still deleting, in addition to option click now
- I'm not sure what the "right" behavior is, but just want to point out that when "preserve labels" is checked, there's no way to use the eraser (without clicking shift to temporarily uncheck that button).  While that "preserve labels" makes a lot of sense for drawing new labels... i wonder whether that block on erasing might not be confusing?  If we want to keep that behavior, it might be nice to "notify" the user somehow that preserve labels is checked... maybe flash the label or something?

tlambert03:
maybe also worth revisiting the order of the buttons?

tlambert03:
> it might be nice to "notify" the user somehow that preserve labels is checked... maybe flash the label or something?

I made it so that if a user tries to erase something with preserve labels checked, the label for the checkbox goes temporarily red:

Untitled

sofroniewn:
> I made it so that if a user tries to erase something with preserve labels checked, the label for the checkbox goes temporarily red:

I love having a visual warning but I don't like turning the label red, makes it hard for me to read.

Could we maybe add a red X icon that would be the ~size of the checkbox and have show up to the right of the checkbox only when happens. Or maybe it would overlay on the checkbox if it has a transparent background?

sofroniewn:
> maybe also worth revisiting the order of the buttons?

Yeah, thats fine. I'd probably go - from RIGHT to LEFT, mag, paint, erase, pick, fill, shuffle

What are your thoughts @tlambert03?

sofroniewn:
> While that "preserve labels" makes a lot of sense for drawing new labels... i wonder whether that block on erasing might not be confusing? 

Actually maybe this is just too confusing and preserve labels should not be functional when the eraser tool is selected, then we wouldn't need any red warnings right?

tlambert03:
> makes it hard for me to read.

also... just curious did you test this locally?  or just looking at the GIF?  (the gif looks a lot worse)

sofroniewn:
I'm going to quickly fix the merge conflict and reorder the buttons, I think this will then be ready for a final review from @jni and could still go in to `0.3.3`, which I think will be nice for @constantinpape who had mentioned the having an erase button for his annotation tool

tlambert03:
@sofroniewn, as mentioned in https://github.com/napari/napari/pull/1288#issuecomment-636035691, right click and drag is still deleting ... I thought we decided against that behavior?

sofroniewn:
@tlambert03 I just updated this PR to fix merge conflicts, reoder the buttons, and drop the "right click" behavior. I've also changed it so that you can have the `selected_label` be `0` to enable fill with `0` again.

Playing more with it I think that the current behavior of `preserve labels` with the erase button is sub optimal, and I want to throw out a new option, especially now with the hotkey `opt` to rapidly enable it. 

Consider the use case that you have a label - say `7` selected and preserve labels turned on and are painting with it, I think if you switch to the erase key using the `opt` key it should ONLY erase label `7` while preserving all others, so you could rapidly change what you're working on - i.e. label 7, without fear of changing all the other labels.

How does that sound @tlambert03 @jni @ziyangczi?

tlambert03:
> if you switch to the erase key using the `opt` key it should ONLY erase label `7` while preserving all others, so you could rapidly change what you're working on

i like that

sofroniewn:
I've now given this a final update, making `preserve_labels` behave as noted above, and extending the preserve to the `fill` case too. I've also dropped the red warning if nothing changes. I'm happy to add the red warning back in if people think it is too confusing that nothing happens on a click somethings.

I think it is ready for final review and merge, would be great if @tlambert03 and @jni could give it a spin and an approve :-)

tlambert03 (approved):
i just pushed some additional changes, removing unnecessary code from the layout spanning, and undoing the event I added for the red text warning.  looks good and works well

sofroniewn:
Just found a bug in the nD labels case, will make a quick fix

jni:
@sofroniewn 

> Yeah, thats fine. I'd probably go - from RIGHT to LEFT, mag, paint, erase, pick, fill, shuffle

I'm curious about the rationale? I kinda think paint and fill should be together, as they are two forms of adding labels. New proposal: `mag, pick, paint, fill, erase, shuffle`.

jni:
btw all the new functionality here is pretty damn cool, thank you all. =)

sofroniewn:
> I'm curious about the rationale? I kinda think paint and fill should be together, as they are two forms of adding labels. New proposal: mag, pick, paint, fill, erase, shuffle.

Will change

Note nD bug slipped in with #1299, my fault!!, easy to fix I will add a test to catch now too :-) update coming right up

sofroniewn:
Ok this fixes bug from #1299 - are we good for a final approve and merge @jni? Then I will update release notes and we can make 0.3.3 release?

sofroniewn:
Would be good to have some thorough testing of all the labels functionality as @constantinpape will need that working. I've tried my hardest to break it now, nD different buttons etc and it seems good

napari/_qt/layers/qt_labels_layer.py
```diff
@@ -29,7 +29,7 @@ class QtLabelsControls(QtLayerControls):
     Attributes
     ----------
     button_group : qtpy.QtWidgets.QButtonGroup
-        Button group of labels layer modes: PAN_ZOOM, PICKER, PAINT, or FILL.
+        Button group of labels layer modes: PAN_ZOOM, PICKER, PAINT, ERASE or FILL.
```
jni: This line is too long... Why hasn't black caught it? :thinking: 

```suggestion
        Button group of labels layer modes: PAN_ZOOM, PICKER, PAINT, ERASE or
        FILL.
```

sofroniewn: not sure why black missed it/ what status of black testing doc strings is. i will fix

sofroniewn: just pushed change fixing doc string wrapping, making sure ERASE listed everywhere needed

jni (approved):
Just played with this. It's so slick! All the bindings are working great on Linux!

sofroniewn:
Great, thanks for the thorough testing @jni, i've updated doc string only in last commit, will wait for tests to pass and then merge!

PR #1290, opened 2020-05-26
0.3.2 release notes

sofroniewn:
# Description
This PR adds release notes for 0.3.2. Note it doesn't include #1210 yet, if we get #1210 in time for the release I will add it, otherwise I think it is good to go

### DIFF
--- a/docs/release/release_0_3_2.md
+++ b/docs/release/release_0_3_2.md
@@ -12,6 +12,7 @@ https://github.com/napari/napari
 
 
 ## New Features
+- General multithreading API and @thread_worker decorator (#1210)
 - Rich jupyter display for napari screenshots (#1269)
 
 ## Improvements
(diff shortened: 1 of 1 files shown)


tlambert03 (approved):
yep, that's what happened! :)

docs/release/release_0_3_2.md
```diff
@@ -0,0 +1,67 @@
...
+
+
+## New Features
+- General multithreading API and @thread_worker decorator (#1210)
+- Rich jupyter display for napari screenshots (#1269)
+
+## Improvements
+- make Qt component non private module (#1122)
+- add option to disable overwrite of labels during painting (#1264)
+- Refactor: move dask utils from misc.py to new dask_utils.py (#1265)
```
jni: Generally I would reserve this section for user-facing improvements. Should we move this to an "other improvements" section?

jni (approved):
@sofroniewn I've made a very minor and optional comment, otherwise this is ready to go, short and sweet! =)

docs/release/release_0_3_2.md
sofroniewn: Let's save for next time, i did fix the capitolizations

PR #1294, opened 2020-05-27
add threading tests

tlambert03:
# Description
Adds tests for #1210.

Note, turns out testing coverage of stuff running in another thread is tricky.  So in addition to actually running stuff in the other thread, I run the workers in the main thread here to test various methods.

## Type of change

- [x] tests

# How has this been tested?

- [x] the test suite for my feature covers the qt.threading module

## Final checklist:

### DIFF
--- a/napari/_qt/_tests/test_threading.py
+++ b/napari/_qt/_tests/test_threading.py
@@ -123,6 +123,9 @@ def test_thread_worker_in_main_thread():
 
     thread_func = threading.thread_worker(func)
     worker = thread_func(2)
+    # NOTE: you shouldn't normally call worker.work()!  If you do, it will NOT
+    # be run in a seperate thread (as it would for worker.start().
+    # This is for the sake of testing it in the main thread.
     assert worker.work() == 2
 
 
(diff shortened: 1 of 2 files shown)


napari/_qt/_tests/test_threading.py
```diff
@@ -0,0 +1,175 @@
...
+# this is just for the sake of coverage
+def test_thread_worker_in_main_thread():
+    """Test basic threadworker on a function"""
+
+    def func(x):
+        return x
+
+    thread_func = threading.thread_worker(func)
+    worker = thread_func(2)
+    assert worker.work() == 2
```
jni: What exactly in this code is causing it to be run on the main thread?

tlambert03: the fact that we call `worker.work()` instead of `worker.start()`... will add that to the docs

jni (approved):
@tlambert03 I asked a question but mostly for curiosity. This looks good to me. Thanks!

tlambert03:
I added a note, since we don't want people looking through tests and thinking calling `worker.work()` is a good practice.

napari/_qt/_tests/test_threading.py
jni: Lovely!

PR #1296, opened 2020-05-27
Allow add_dock_widget to accept a list of widgets

tlambert03:
# Description
as discussed in https://github.com/hci-unihd/covid-if-annotations/issues/3, it would be nice for `add_dock_widget` to accept a list/tuple of widgets, and combine them into a single dock widget, rather than having to add multiple individual widgets.  This PR does that, using a QHBoxLayout if it's on the bottom or top, and QVBoxLayout if it's on the left or right.  Also adds "stretch" at the bottom of the layout if it's a vertical layout, so the widgets aren't so spread aprt.

## Type of change

# How has this been tested?

- [x] new tests add, some of which cover previous dock_widget functionality

## Final checklist:

### DIFF
--- a/docs/release/release_0_3_2.md
+++ b/docs/release/release_0_3_2.md
@@ -14,6 +14,7 @@ https://github.com/napari/napari
 ## New Features
 - General multithreading API and @thread_worker decorator (#1210)
 - Rich jupyter display for napari screenshots (#1269)
+- Allow add_dock_widget to accept a list of widgets (#1296)
 
 ## Improvements
 - Make Qt component non private module (#1122)
--- a/napari/_qt/_tests/test_qt_dock_widget.py
+++ b/napari/_qt/_tests/test_qt_dock_widget.py
@@ -1,0 +1,64 @@
+import pytest
+from qtpy.QtWidgets import QDockWidget, QHBoxLayout, QPushButton, QVBoxLayout
+
+
+def test_add_dock_widget(viewer_factory):
+    """Test basic add_dock_widget functionality"""
+    view, viewer = viewer_factory()
+    widg = QPushButton('button')
+    dwidg = viewer.window.add_dock_widget(widg, name='test')
+    assert not dwidg.is_vertical
+    assert viewer.window._qt_window.findChild(QDockWidget, 'test')
+    assert dwidg.widget == widg
+    dwidg._on_visibility_changed(True)  # smoke test
+
+    widg2 = QPushButton('button')
+    dwidg2 = viewer.window.add_dock_widget(widg2, name='test2', area='right')
+    assert dwidg2.is_vertical
+    assert viewer.window._qt_window.findChild(QDockWidget, 'test2')
+    assert dwidg2.widget == widg2
+    dwidg2._on_visibility_changed(True)  # smoke test
+
+    with pytest.raises(ValueError):
+        # 'under' is not a valid area
+        viewer.window.add_dock_widget(widg2, name='test2', area='under')
+
+    with pytest.raises(ValueError):
+        # 'under' is not a valid area
+        viewer.window.add_dock_widget(
+            widg2, name='test2', allowed_areas=['under']
+        )
+
+    with pytest.raises(TypeError):
+        # allowed_areas must be a list
+        viewer.window.add_dock_widget(
+            widg2, name='test2', allowed_areas='under'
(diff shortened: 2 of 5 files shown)


tlambert03:
hope it wasn't too optimistic of me... but I also added this to the release notes in case this merges before 0.3.2 ... if not, I can remove that.

sofroniewn:
Love this, but I'm running into an unexpected behavior - this isn't a minimal bug report, i can work on making one, but if have this magicgui

```python
@magicgui(call_button="execute")
def segmentation(base_image: layers.Image) -> layers.Labels:
    """Segment an image based using stardist."""

    print("Segmenting...")

    # fit and predict
    if base_image.rgb:
        data = base_image.data.mean(axis=2)
    else:
        data = base_image.data

    # normalize image based on clims
    clims = base_image.contrast_limits
    data = (data - clims[0]) / (clims[1] - clims[0])

    labels, details = model.predict_instances(data)

    print("... Completed")  

    return labels
```

and then do

```python
    viewer = napari.Viewer()

    # instantiate the widget
    gui = segmentation.Gui()

    list_widget = QListWidget()
    for n in names:
        list_widget.addItem(n)

    list_widget.currentItemChanged.connect(open_name)

    viewer.window.add_dock_widget(list_widget, area='right')
    viewer.window.add_dock_widget(gui, area='right')

    # keep the dropdown menus in the gui in sync with the layer model
    viewer.layers.events.changed.connect(lambda x: gui.refresh_choices())

    gui.refresh_choices()

    list_widget.setCurrentRow(0)
```
then I see

Screen Shot 2020-05-27 at 8 41 01 AM

as expected. Now if i try and use the combine list functionality note two things happen

```python
viewer = napari.Viewer()

    # instantiate the widget
    gui = segmentation.Gui()

    list_widget = QListWidget()
    for n in names:
        list_widget.addItem(n)

    list_widget.currentItemChanged.connect(open_name)

    viewer.window.add_dock_widget([list_widget, gui], area='right')
    
    # keep the dropdown menus in the gui in sync with the layer model
    viewer.layers.events.changed.connect(lambda x: gui.refresh_choices())

    gui.refresh_choices()

    list_widget.setCurrentRow(0)
```

Screen Shot 2020-05-27 at 8 41 33 AM

Note that the string layer title in the dropdown menu in `gui` is no longer visible (this is the most concerning thing) and the stretch makes it look not so good.

If I drop the stretch then that looks better, but the string is still missing
Screen Shot 2020-05-27 at 8 46 09 AM

any ideas @tlambert03 on why the string disappeared?

Also should be add a `stretch` bool to `combine_widgets`? What about to `add_dock_widget`? Might be nice to let people explicitly control, might be API explosion? Is there a way to "guess" what the right thing to do is? I can't add a stretch factor to my `list_widget` can I? I can only do that on the layout, I think. Otherwise if it could set that to be a bigger factor it would all be ok?

Note that the answer of "you're using custom widgets already, you clearly know what's going on and so should just do you're own VLayout" is a totally fine answer

tlambert03:
ah... interesting.  yeah, that's because our [`_magicgui.get_layers` implementation](https://github.com/napari/napari/blob/master/napari/utils/_magicgui.py#L86) looks only at the first parent, so it's not finding the viewer, and therefore can't find the labels.

as for the stretch... let me think about how we might improve that, or pass the stretch arg through the chain

tlambert03:
ok, I think I have a nice solution.  It will now only add a vertical stretch if none of the contained widgets have a [sizepolicy](https://doc.qt.io/qt-5/qsizepolicy.html) of expanding (like your ListWidget does).  Otherwise, it lets the widgets expand themselves.  Also updated the magicgui support module to recurse the parent chain looking for the viewer

sofroniewn (approved):
Amazing @tlambert03 - looks beautiful!
Screen Shot 2020-05-27 at 9 56 45 AM

sofroniewn:
Does this PR contain #1295? As we are still getting the same segfault

```
napari/utils/_tests/test_notebook_display.py::test_nbscreenshot Fatal Python error: Segmentation fault
```

tlambert03:
it does unfortunately... trying to figure it out now

sofroniewn:
Merging now, test fails are unrelated

jni:
Just catching up, awesome work here @tlambert03 and @sofroniewn !

PR #1298, opened 2020-05-27
exit context before return on _repr_png

tlambert03:
# Description
lets try this fix for the notebook tests... exiting the context before returning.

I also think something funny is going on with cirrus today.  i tried to test this on my branch and after an hour, it still was hanging.

### DIFF
--- a/napari/utils/notebook_display.py
+++ b/napari/utils/notebook_display.py
@@ -73,4 +73,5 @@ class NotebookScreenshot:
         with BytesIO() as file_obj:
             imsave(file_obj, self.image, format='png')
             file_obj.seek(0)
-            return file_obj.read()
+            png = file_obj.read()
+        return png


sofroniewn:
> lets try this fix for the notebook tests... exiting the context before returning.

looks like it is fixing the linux fails!!

> I also think something funny is going on with cirrus today. i tried to test this on my branch and after an hour, it still was hanging.

looks like the mac fails are cirrus related ...

tlambert03:
yeah, encouraging... maybe if it passes we should run it again a couple times?

tlambert03:
ok I've run this a few times now... and except for the timeouts (where the test takes like 50+min due to some cirrus stuff), I haven't seen any errors.  so maybe it's good?

sofroniewn:
I think it is good - merging now!!

sofroniewn (approved):
LGTM!

sofroniewn (commented):
Actually @tlambert03 - should you initialize `png=None` or something else outside the context manager just in case it fails? Not sure if that is best practice?

jni:
I don’t think that’s necessary @sofroniewn. This isn’t a try/except so a failure above will raise an error. I could be mistaken but I don’t think you’ll ever hit an unbound local error here.

sofroniewn:
Ok sounds good @jni I will merge then

Issue #1235, opened 2020-05-10
Allow dragging in "fill" mode in labels


jni:
## 🚀 Feature

When painting many labels the same color (e.g. filling in background, or bigger objects from superpixels), if one already has reasonable superpixels, it's much easier to drag and roughly paint over the whole object than to click on each individual pixel.

sofroniewn:
Can you describe this a little more actually @jni I'm not sure i fully understand enough to know what is requested.

Say i have fill bucket tool highlighted, and then i click on a pixel, right now it fills all contiguous pixels of that label with the new label.

If I click and start dragging what do you want to happen? Do you want it to start moving that contiguous component around? Should it "paint" with that new label into new pixels? Should it fill in where it used to be with the background label or should it only grow? Should it actually only modifiy pixels when you release? That would be pretty cool, filling in background with where it used to be?

Screenshots might help too

jni:
dragging should be equivalent to clicking on each pixel in the drag path. In other words, it should be like moving and clicking fill many times over.

jni:
(which is what you need to do now to paint a superpixel image.)

jni:
Here's some code to play around with:

```python
import napari
from skimage import data, segmentation

astro = data.astronaut()
segmented = segmentation.slic(
    astro, compactness=20, sigma=1, start_label=1
)

with napari.gui_qt():
    v = napari.Viewer()
    imglayer = v.add_image(astro)
    seglayer = v.add_labels(segmented)
    seglayer.mode = 'fill'
```

Now we define the foreground as the astronaut's face and background as everything else. We want to fill the background superpixels with 0 and the foreground superpixels with 1. I guarantee you'll be tired of clicking before you're done with the background and you'll wish you could just drag across it. :joy:

jni:
(based on a true story =P)

VolkerH:
I like this example. From the docs, it appears that skimage can do slics only on colour images. With superpixels on greyvalue images, this would allow for a "magic wand"-like action which should greatly speed up annotations and accuracy around edges.

sofroniewn:
I get it now!! Thanks for the explanation @jni. Agreed clicking got quite tiring!! If I had been able to click and then fill while dragging that would be very nice. Here's a video of me clicking each superpixel:

![fill_drag](https://user-images.githubusercontent.com/6531703/82682427-17b8a300-9c04-11ea-8ff8-301facc3dbd2.gif)

liu-ziyang:
[...]

PR #1299, opened 2020-05-27
1235 allow drag to fill

liu-ziyang:
# Description
Address feature request in issue 1235 to allow drag to fill

## Type of change

# References
Feature requested in https://github.com/napari/napari/issues/1235

# How has this been tested?

- Manually tested using drag to fill and paint

## Final checklist:

### DIFF
--- a/napari/layers/labels/_labels_mouse_bindings.py
+++ b/napari/layers/labels/_labels_mouse_bindings.py
@@ -27,7 +27,7 @@ def draw(layer, event):
     if layer._mode == Mode.PAINT:
         layer.paint(layer.coordinates, layer.selected_label)
     elif layer._mode == Mode.FILL:
-        layer.fill(layer.coordinates, layer._value, layer.selected_label)
+        layer.fill(layer.coordinates, layer.selected_label)
     last_cursor_coord = layer.coordinates
     yield
 
@@ -41,11 +41,9 @@ def draw(layer, event):
                 layer.paint(c, layer.selected_label, refresh=False)
             elif (
                 layer._mode == Mode.FILL
-                and layer._value != layer.selected_label
+                and layer.get_value(c) != layer.selected_label
             ):
-                layer.fill(
-                    layer.coordinates, layer._value, layer.selected_label
-                )
+                layer.fill(c, layer.selected_label)
         layer.refresh()
         last_cursor_coord = layer.coordinates
         yield
--- a/napari/layers/labels/_labels_mouse_bindings.py
+++ b/napari/layers/labels/_labels_mouse_bindings.py
@@ -1,12 +1,16 @@
 from ._labels_utils import interpolate_coordinates
+from ._labels_constants import Mode
 
 
-def paint(layer, event):
+def draw(layer, event):
     """Paint with the currently selected label."""
     # on press
     layer._save_history()
     layer._block_saving = True
-    layer.paint(layer.coordinates, layer.selected_label)
+    if layer._mode == Mode.PAINT:
+        layer.paint(layer.coordinates, layer.selected_label)
+    elif layer._mode == Mode.FILL:
+        layer.fill(layer.coordinates, layer._value, layer.selected_label)
     last_cursor_coord = layer.coordinates
     yield
 
@@ -16,7 +20,15 @@ def paint(layer, event):
             last_cursor_coord, layer.coordinates, layer.brush_size
         )
         for c in interp_coord:
-            layer.paint(c, layer.selected_label, refresh=False)
+            if layer._mode == Mode.PAINT:
+                layer.paint(c, layer.selected_label, refresh=False)
+            elif (
+                layer._mode == Mode.FILL
+                and layer._value != layer.selected_label
+            ):
+                layer.fill(
+                    layer.coordinates, layer._value, layer.selected_label
+                )
         layer.refresh()
         last_cursor_coord = layer.coordinates
         yield
@@ -29,9 +41,3 @@ def pick(layer, event):
     """Change the selected label to the same as the region clicked."""
     # on press
     layer.selected_label = layer._value or 0
-
-
-def fill(layer, event):
-    """Fill in an area with the currently selected label."""
-    # on press
-    layer.fill(layer.coordinates, layer._value, layer.selected_label)
--- a/napari/layers/labels/_labels_mouse_bindings.py
+++ b/napari/layers/labels/_labels_mouse_bindings.py
@@ -3,7 +3,24 @@ from ._labels_constants import Mode
 
 
 def draw(layer, event):
-    """Paint with the currently selected label."""
+    """Draw with the currently selected label to a coordinate.
+
+    This method have different behavior when draw is called
+    with different labeling layer mode.
+
+    In PAINT mode the cursor functions like a paint brush changing any
+    pixels it brushes over to the current label. If the background label
+    `0` is selected than any pixels will be changed to background and this
+    tool functions like an eraser. The size and shape of the cursor can be
+    adjusted in the properties widget.
+
+    In FILL mode the cursor functions like a fill bucket replacing pixels
+    of the label clicked on with the current label. It can either replace
+    all pixels of that label or just those that are contiguous with the
+    clicked on pixel. If the background label `0` is selected than any
+    pixels will be changed to background and this tool functions like an
+    eraser
+    """
     # on press
     layer._save_history()
     layer._block_saving = True
--- a/napari/layers/labels/_labels_mouse_bindings.py
+++ b/napari/layers/labels/_labels_mouse_bindings.py
@@ -41,7 +41,7 @@ def draw(layer, event):
                 layer.paint(c, layer.selected_label, refresh=False)
             elif (
                 layer._mode == Mode.FILL
-                and layer.get_value(c) != layer.selected_label
+                and layer.get_value() != layer.selected_label
             ):
                 layer.fill(c, layer.selected_label)
         layer.refresh()
--- a/napari/layers/labels/_labels_mouse_bindings.py
+++ b/napari/layers/labels/_labels_mouse_bindings.py
@@ -39,11 +39,8 @@ def draw(layer, event):
         for c in interp_coord:
             if layer._mode == Mode.PAINT:
                 layer.paint(c, layer.selected_label, refresh=False)
-            elif (
-                layer._mode == Mode.FILL
-                and layer.get_value() != layer.selected_label
-            ):
-                layer.fill(c, layer.selected_label)
+            elif layer._mode == Mode.FILL:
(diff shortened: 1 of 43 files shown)


napari/layers/labels/_labels_mouse_bindings.py
```diff
@@ -16,7 +37,15 @@ def paint(layer, event):
...
         for c in interp_coord:
-            layer.paint(c, layer.selected_label, refresh=False)
+            if layer._mode == Mode.PAINT:
+                layer.paint(c, layer.selected_label, refresh=False)
+            elif (
+                layer._mode == Mode.FILL
+                and layer._value != layer.selected_label
+            ):
+                layer.fill(
+                    layer.coordinates, layer._value, layer.selected_label
```
sofroniewn: Inside the for loop this should be
```suggestion
                    c, layer._value, layer.selected_label
```
to use the "interpolated coordinates" instead of just the current clicked coordinate

sofroniewn (commented):
This looks like it is on the right track! Nice job.

It needs a couple changes around how the "interpolated coordinates" is handled in the draw so as to get intermediate mouse locations during a fast drag (see my comment in the main body).

We should also change the `layer.fill` function to drop `old_label` from the function signature, and instead compute the `old_label` based on the values of `int_coord` as now the label at the interpolated coordinate could be different then the label in `_value`. Does that make sense?

We also might need to do some clipping of the `int_coord` somewhere as I can currently get an error like 
```python-traceback
(base) czirwc1macos2701:image-demos nsofroniew$ python test-examples/dragging_fill.py
  File "/Users/nsofroniew/GitHub/vispy/vispy/app/backends/_qt.py", line 476, in mouseMoveEvent
  File "/Users/nsofroniew/GitHub/vispy/vispy/app/base.py", line 213, in _vispy_mouse_move
    self, cb_event=(cb, event))
    cb(event)
  File "/Users/nsofroniew/GitHub/napari/napari/_qt/qt_viewer.py", line 522, in on_mouse_move
    mouse_move_callbacks(layer, event)
  File "/Users/nsofroniew/GitHub/napari/napari/utils/interactions.py", line 105, in mouse_move_callbacks
    next(gen)
  File "/Users/nsofroniew/GitHub/napari/napari/layers/labels/_labels_mouse_bindings.py", line 47, in draw
    c, layer._value, layer.selected_label
  File "/Users/nsofroniew/GitHub/napari/napari/layers/labels/labels.py", line 473, in fill
    match_label = labeled_matches[slice_coord]
IndexError: index 513 is out of bounds for axis 1 with size 512
```

when moving the mouse very fast. Could do the clipping on `int_coord` inside `fill` as I think clipping is already handled inside painting.

Might be nice to expand our testing in `tests_labels_mouse_bindings.py` to make sure this case of cursor clicking, dragging out of range is covered.

napari/layers/labels/_labels_mouse_bindings.py
liu-ziyang: addressed the interpolation and clip on coordinates, nice catch!

liu-ziyang:
> This looks like it is on the right track! Nice job.
> 
> It needs a couple changes around how the "interpolated coordinates" is handled in the draw so as to get intermediate mouse locations during a fast drag (see my comment in the main body).
> 
> We should also change the `layer.fill` function to drop `old_label` from the function signature, and instead compute the `old_label` based on the values of `int_coord` as now the label at the interpolated coordinate could be different then the label in `_value`. Does that make sense?
> 
> We also might need to do some clipping of the `int_coord` somewhere as I can currently get an error like
> 
> ```
> (base) czirwc1macos2701:image-demos nsofroniew$ python test-examples/dragging_fill.py
> Traceback (most recent call last):
>   File "/Users/nsofroniew/GitHub/vispy/vispy/app/backends/_qt.py", line 476, in mouseMoveEvent
>     modifiers=self._modifiers(ev),
>   File "/Users/nsofroniew/GitHub/vispy/vispy/app/base.py", line 213, in _vispy_mouse_move
>     ev = self._vispy_canvas.events.mouse_move(**kwargs)
>   File "/Users/nsofroniew/GitHub/vispy/vispy/util/event.py", line 455, in __call__
>     self._invoke_callback(cb, event)
>   File "/Users/nsofroniew/GitHub/vispy/vispy/util/event.py", line 475, in _invoke_callback
>     self, cb_event=(cb, event))
>   File "/Users/nsofroniew/GitHub/vispy/vispy/util/event.py", line 471, in _invoke_callback
>     cb(event)
>   File "/Users/nsofroniew/GitHub/napari/napari/_qt/qt_viewer.py", line 522, in on_mouse_move
>     mouse_move_callbacks(layer, event)
>   File "/Users/nsofroniew/GitHub/napari/napari/utils/interactions.py", line 105, in mouse_move_callbacks
>     next(gen)
>   File "/Users/nsofroniew/GitHub/napari/napari/layers/labels/_labels_mouse_bindings.py", line 47, in draw
>     c, layer._value, layer.selected_label
>   File "/Users/nsofroniew/GitHub/napari/napari/layers/labels/labels.py", line 473, in fill
>     match_label = labeled_matches[slice_coord]
> IndexError: index 513 is out of bounds for axis 1 with size 512
> ```
> 
> when moving the mouse very fast. Could do the clipping on `int_coord` inside `fill` as I think clipping is already handled inside painting.
> 
> Might be nice to expand our testing in `tests_labels_mouse_bindings.py` to make sure this case of cursor clicking, dragging out of range is covered.

beyond scope of this PR, but I think we should handle the coordinates more consistently in a single place, it is a bit all over the place right now and developers have to trace down a few places to see how the transformation is done. Maybe there is some merit that I am missing? I am having the impression that all Mode handles mouse gestures similarly

napari/layers/labels/labels.py
```diff
@@ -447,25 +447,28 @@ def fill(self, coord, old_label, new_label):
...
             labels = self._data_raw
-            slice_coord = tuple(int_coord[d] for d in self.dims.displayed)
+            slice_coord = tuple(
+                np.round(np.clip(coord[d], 0, self.shape[d] - 1)).astype(int)
+                for d in self.dims.displayed
+            )
 
-        matches = labels == old_label
+        self.coordinates = slice_coord
+        matches = labels == self.get_value()
```
```suggestion
        old_label = self.get_value()
        if old_label == new_label:
            return
        matches = labels == old_label
```

See my note above, I think we should move the catch inside the fill function, lets us really make sure everything is the way we want it, and reduces the number of calls to `get_value`

napari/layers/labels/_labels_mouse_bindings.py
```diff
@@ -16,7 +37,13 @@ def paint(layer, event):
             last_cursor_coord, layer.coordinates, layer.brush_size
         )
         for c in interp_coord:
-            layer.paint(c, layer.selected_label, refresh=False)
+            if layer._mode == Mode.PAINT:
+                layer.paint(c, layer.selected_label, refresh=False)
+            elif (
+                layer._mode == Mode.FILL
+                and layer.get_value() != layer.selected_label
+            ):
```
sofroniewn:
```suggestion
            elif layer._mode == Mode.FILL:
```

I think we should drop the `and layer.get_value() != layer.selected_label` check here and put like that inside the `fill` function as I think the `get_value` could lag one interpolated coordinate behind the `c` coordinate, and we make two `get_value` calls instead of one. See comment below

napari/layers/labels/_labels_mouse_bindings.py
```diff
@@ -16,7 +37,13 @@ def paint(layer, event):
...
+                layer.fill(c, layer.selected_label)
```
sofroniewn: We should add the `refresh=True` kwarg to `fill` and set it to be `False` when called inside this for-loop.

We should then add
```python
        if refresh is True:
            self._save_history()
```
and

```python
        if refresh is True:
            self.refresh()
```

to the top and bottom of the `fill` function, in the same way they are used in the `paint` function, you can copy the doc string over too.

Having this be false during the drag should be a slight performance improvement as we we'll only trigger refresh once per mouse move, not for every interpolated point, and it will better match how `paint` works

sofroniewn (approved):
Looking good, left some final comments that should be addressed, but approving now. Should be good to merge tomorrow. Using the functionality feels great!

napari/layers/labels/_labels_mouse_bindings.py
jni: @sofroniewn does that mean though that we won't see the result of the filling as we're doing it?

sofroniewn: No you'll still see the result as you're doing it - every move of the mouse. This means you wont bombard with 10x the number of refreshes due to the "interpolation" function leading to many calls per mouse move due to the extra `for` loop

liu-ziyang: Added! nice catch

napari/layers/labels/_labels_mouse_bindings.py
liu-ziyang: good call, added

napari/layers/labels/labels.py
```diff
@@ -449,8 +449,16 @@ def fill(self, coord, new_label):
...
             Value of the new label to be filled in.
+        refresh : bool
+            Whether to refresh view slice or not. Set to False to batch paint
+            calls.
         """
-        self._save_history()
+        if refresh is True:
+            self._save_history()
+
+        old_label = self.get_value()
```
sofroniewn: The `get_value` call needs to come after the `self.coordinates = slice_coord` update otherwise it will use the old `coordinates`, we might want to refactor and improve this at some point as it is not intuitive/ could lead to some surprising bugs, but for now can you move the call down lower. The slicing code should be fast, so it's ok. The slow line is what used to be 471.

liu-ziyang: yep, the advanced test caught it! corrected it since

napari/layers/labels/labels.py
```diff
@@ -447,23 +447,35 @@ def fill(self, coord, old_label, new_label):
...
-            slice_coord = tuple(int_coord[d] for d in self.dims.displayed)
+            slice_coord = tuple(
+                np.round(np.clip(coord[d], 0, self.shape[d] - 1)).astype(int)
+                for d in self.dims.displayed
+            )
+
+        self.coordinates = slice_coord
+
+        old_label = self.get_value()
+        if old_label == self.selected_label:
```
```suggestion
        if old_label == new_label:
```
this check should be against `new_label` which is going to be filled in

sofroniewn:
I just updated with the latest changes from master, hopefully fixing any remaining test fails

sofroniewn:
I've just made a minor fix to this PR that prevents filling when the click location if out of the data range (my advice earlier to clip was wrong, we really should have just ignored).

I've played with this quite a lot now, the functionality and code both look great. I will merge after tests all pass.

PR #1306, opened 2020-05-28
make window public

tlambert03:
# Description
as per https://github.com/napari/napari/pull/1122#issuecomment-635058811, adds `qt_main_window.Window` to the public qt module, so it appears in the docs


jni (approved):

sofroniewn (approved):
Thanks @tlambert03!!!

Not sure why tests are being weird, will merge once I re-run

tlambert03:
I think pyside 5.15 _might_ be actually introducing some intermittent problems.  (been seeing strange stuff elsewhere too)

PR #1316, opened 2020-05-30
Pin PySide2 and PyQt5 at <5.15

### LINKED ISSUE #1312: Pyside 5.15.0 breaking buttons?
## 🐛 Bug

just found that all instances of `QtModePushButton` weren't working for me.  (like the "add layer" buttons)
Screen Shot 2020-05-29 at 11 13 35 AM

I had pyside 5.15.0 in my environment, and downgraded to 5.14.2 and they worked again...

Can someone else confirm this? 

If so... we should probably release a patch that pins pyside2 < 5.15

tlambert03:
# Description
We need to do some more testing on Qt 5.15 before we open it up in our requirements.  Definitely some things broken at the moment (though tests are passing)

## Type of change

# References
https://www.qt.io/blog/qt-for-python-5.15.0-is-out

# How has this been tested?

## Final checklist:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -79,9 +79,9 @@ with open(osp.join('requirements', 'default.txt')) as f:
             requirements.append(stripped)
 
 if pyqt:
-    requirements.append("PyQt5>=" + QT_MIN_VERSION)
+    requirements.append(f"PyQt5>={QT_MIN_VERSION},<5.15")
 if pyside or not (pyqt or pyside):
-    requirements.append("PySide2>=" + QT_MIN_VERSION)
+    requirements.append(f"PySide2>={QT_MIN_VERSION},<5.15")
 
 
 setup(


jni:
I guess we want to fix this on our end and then it doesn't matter if it's `<` or `!=`. BUT, if we do want to fix this on our end, then we should remove the "fixes #" above. =)

jni:
And if we would rather wait and see, then we should do `!=`

Either way we should probably add a test for this... Would that be hard? Sorry I am a noob at the Qt testing stuff...

tlambert03:
it's actually quite tricky to test!  The strange thing is that the button does not seem to be emitting a signal when clicked, but let me try to mock it.  might take some time, so I'll do it in another PR, and this one can go in

tlambert03:
yeah: `napari/_qt/_tests/test_qt_buttons.py::test_push_button` _should_ be catching this but it doesn't ... might need to actually mock a click at a specific part of the screen

tlambert03:
oh man... if you programmatically run `viewer.window.qt_viewer.layerButtons.newPointsButton.click()`, it adds a points layer... but actually clicking on it has no effect.

sofroniewn (approved):
LGTM! Will merge with approval in zulip from @jni too

Issue #1103, opened 2020-04-03
Functions to split/combine multiple layers along an axis


GenevieveBuckley:
## 🚀 Feature

Juan suggested on zulip that it would be a good idea to have:
1. A function to split an image axis into separate layers
2. A function to combine multiple layers into a single layer along a new axis.

## Motivation

This would allow users to more easily work with multichannel images, etc. since they may be saved as a single file and imported into napari as a single layer.

cwood1967:
Hi. Is anyone currently working on this? If not, I would like to see what I can to do contribute, and this seems like a good first issue. Splitting/combining layers is straight forward, but where to put the function isn't clear to me. 

Thanks for any suggestions,
Chris

GenevieveBuckley:
It's open, go and take a shot at it @cwood1967 

It's totally fine to get a working function together first, and then we can work out where's best to put it.

sofroniewn:
Closed by #1322!

PR #1322, opened 2020-06-01
Functions to split/combine multiple layers along an axis

cwood1967:
# Description
This pull request adds functions to split image layers along an axis and combine image layers into a new single layer along a new axis as described in issue #1103.

This feature will make it easy to split multichannel images that are opened into a single layer, and allow split images to be recombined. Also some print statements when input layers don't make sense; not sure how to handle that.

I put these functions in napari/utils/axis_functions.py; I wasn't sure the best place for it.

I noticed that creating and deleting layers of different dimensions will throw errors that seem related to viewer.dims not being up to date with viewer.active_layer.dim, but this happens just by adding and deleting without these functions.  I added issue #1323 for this.

## Type of change

# References
This is for feature issue #1103.

# How has this been tested?
Manually tested.

## Final checklist:


sofroniewn:
Hi @cwood1967 sorry for the delay in looking at this, I will try and get to it tomorrow!

GenevieveBuckley:
I took a quick look over this last night @cwood1967, thanks for working on this!

The major points I have to suggest are:
* We shouldn't pass around the whole viewer object, but instead only the actual image layers we want to operate on
* These functions should be kept separate from actually adding the output layers to the viewer
* I don't want us to try and guess which axis to split or combine along, I think that should be a required argument from the user
* I don't think there should be any special handling for RGB images

I think the functions should probably live in `napari/layers/util/layer_utils.py`. I'm happy to be overruled by @sofroniewn on this point though.

The function signatures could be like this:
```python
def stack_to_images(stack, axis):
    """Split an image stack into a list of images.

    Parameters
    ----------
    stack : napari.layers.Image
        Image stack to be split into a list of Image layers.
    axis : int
        The axis to split the image stack on.

    Returns
    -------
    list
        List of images
    """
    # ...do the thing
    images = [ImageLayer1, ImageLayer2, ..., ImageLayerN]
    return images


def images_to_stack(images, axis=-1):
    """Combine a list of images into a single image stack.

    Parameters
    ----------
    images : list
        List of napari Image layers to combine.
    axis : int
        Index to insert the new axis.
    
    Returns
    -------
    stack : napari.layers.Image
        Combined image stack
    """
    # ...do the thing
    return stack
```

And then you'd be able to write a demo script using your new functions (you don't have to put the demo script in the napari codebase, sharing in a comment here or as a gist works fine).

```python

import napari
from napari.layers import Image
from napari.layers.utils.layer_utils import stack_to_images, images_to_stack
import numpy as np

# Make some data
data = np.random.random((3, 512, 512))
original_stack = Image(data)
original_images = [Image(np.random.random(512, 512)) for _ in range(3)]

# Try the new functions
result_images = stack_to_images(original_stack, axis=0)
result_stack = images_to_stack(original_images, axis=0)

viewer = napari.Viewer()
# Display the original images
viewer.add_layer(original_stack, name='Original stack')
for i, image in enumerate(original_images):
    viewer.add_layer(image, name=f'Original image {i}')
# Display the results
viewer.add_layer(result_stack, name='Result stack')
for i, image in enumerate(result_images):
    viewer.add_layer(image, mame=f'Result image {i}')
```

cwood1967:
@GenevieveBuckley , thank you for your comments and suggestions. I made the changes you indicated. I tried to put the functions into layer_utils.py, but it seemed to be making some circular imports. I added a `from ..layers import Image` to layer_utils.py, but Image (or something else) imports this module. I put the functions into napari/layers/utils/stack_utils.py for now.

Thanks!

GenevieveBuckley:
You won't need to import `Image`, you can rely on an Image type always being passed as the input to your functions. Then circular imports shouldn't be a problem.

I'll take another look at your code in the next few days, although you might hear from Nick before then.

cwood1967:
I updated the functions without importing Image and put them into layer_utils.py.  The layer name is prepended with a "00_", "01_", etc. I think it makes selecting layers easier when the channel number (or however it is split) is at the start so it isn't hidden in the layer name label.

Here is a testing script that doesn't use the viewer:

```python
from napari.layers.image.image import Image
from napari.layers.utils.layer_utils import stack_to_images, images_to_stack
import numpy as np

def test_stack_to_images():
    
    data = np.random.randint(0,100, (10, 3, 128, 128))
    stack = Image(data)
    images = stack_to_images(stack, 1)

    assert isinstance(images, list)
    assert len(images) == 3
    
    for i in images:
        assert type(stack) == type(i)
        assert i.data.shape == (10, 128, 128)

def test_images_to_stack():

    images = [Image(np.random.randint(0, 255, (10, 128, 128))) for _ in range(3)]

    stack = images_to_stack(images, 1)

    assert isinstance(stack, Image)
    assert stack.data.shape == (10, 3, 128, 128)
```

napari/layers/utils/layer_utils.py
```diff
@@ -233,3 +234,101 @@ def compute_multiscale_level(
...
+    new_list = list()
+    for i in range(nc):
+        # if not isinstance(images[i], Image):
+        #     print("Selected layer {} is not an image".format(images[i].name))
+        #     return None
+        if i == 0:
+            shape = images[i].data.shape
+        else:
+            if shape != images[i].data.shape:
+                print("Layers not the same size")
```
GenevieveBuckley: I think it would be more efficient to not have this check every time we go through the loop, but instead rely on `np.stack` to throw an appropriate error. Then we can use a try/except around `np.stack` to handle that.

napari/layers/utils/layer_utils.py
```diff
@@ -233,3 +234,101 @@ def compute_multiscale_level(
...
+    """
+
+    nc = len(images)
+    if nc < 2:
+        print("less than two layers selected")
+        return None
+
+    new_list = list()
+    for i in range(nc):
+        # if not isinstance(images[i], Image):
```
GenevieveBuckley: Could you delete the commented lines?

napari/layers/utils/layer_utils.py
```diff
@@ -233,3 +234,101 @@ def compute_multiscale_level(
...
+    axis : int
+        The axis to split along.
+
+    Returns
+    -------
+    list
+        List of images
+    """
+
+    # if not isinstance(stack, Image):
```
GenevieveBuckley: Could you delete the commented lines? 

We should use type hinting in the function signature instead of these kinds of checks. Note that both `Image` and `Labels` layers should be accepted by these functions (label layers are more or less a special case of images in napari)

napari/layers/utils/layer_utils.py
```diff
@@ -2,6 +2,7 @@
 
 import numpy as np
 from vispy.color import Colormap
+from napari.utils.colormaps import simple_colormaps
```
GenevieveBuckley: I'm uneasy about the idea of automatically choosing colormaps for users. Ultimately I think this problem is too open-ended for us to reliably make a good and useful default choice. I think the default should be "you keep the colormap you had originally in the image stack"

I would be happy to see colormaps as an optional keyword argument, so users could pass in a list of colormaps in the order that they want them. (We could maybe do that with all possible keyword arguments too, if we wanted)

napari/layers/utils/layer_utils.py
```diff
@@ -233,3 +234,101 @@ def compute_multiscale_level(
...
+
+    Returns
+    -------
+    stack : napari.layers.Image
+        Combined image stack
+    """
+
+    nc = len(images)
+    if nc < 2:
+        print("less than two layers selected")
```
GenevieveBuckley: I don't think this check is needed, to be honest. Nothing bad happens if you try and stack a single image together with nothing else, so I don't think we need to waste energy trying to prevent people from trying to do it (accidentally or no)

napari/layers/utils/layer_utils.py
```diff
@@ -233,3 +234,101 @@ def compute_multiscale_level(
...
+        return None
+
+    new_list = list()
+    for i in range(nc):
```
GenevieveBuckley: This is simpler if we write it as 
```python
data = []
for image in images:
    data.append(image.data)
    ...
```

napari/layers/utils/layer_utils.py
```diff
@@ -233,3 +234,101 @@ def compute_multiscale_level(
...
+            shape = images[i].data.shape
+        else:
+            if shape != images[i].data.shape:
+                print("Layers not the same size")
+                return None
+        new_list.append(images[i].data)
+
+    new_data = np.stack(new_list, axis=axis)
+
+    stack = type(images[0])(new_data)
```
GenevieveBuckley: Oh I see what's happened here, this must be the spot where circular imports were making things tricky. @sofroniewn should we shift all of this someplace else?

napari/layers/utils/layer_utils.py
```diff
@@ -233,3 +234,101 @@ def compute_multiscale_level(
...
+    #     return None
+
+    data = stack.data
+    name = stack.name
+
+    cmaps = list(simple_colormaps.keys())
+    num_dim = len(data.shape)
+
+    if num_dim < 3:
+        print("not enough dimensions")
```
GenevieveBuckley: A python warning might be more appropriate than a print statement for this sort of thing

https://docs.python.org/3/library/warnings.html

GenevieveBuckley (commented):
I've left a few more comments.

You're right, circular imports are definitely a problem (sorry I didn't realize the point you were making earlier!). I'd like @sofroniewn to suggest a better place for these functions to live

napari/layers/utils/layer_utils.py
sofroniewn: I'd been thinking about this too. One approach would be to use `layer._get_state()` to get and propagate as many possible properties as possible. I do like letting people pass lists of properties too. I'd try and mirror the approach of `add_image` as much as possible. There we have sensible defaults if two, three, or more images are being stacked

napari/layers/utils/layer_utils.py
sofroniewn: Would it help if things were in our https://github.com/napari/napari/blob/master/napari/layers/image/_image_utils.py?

cwood1967: I tried using _image_utils.py, but it looks like _image_utils.py is imported when Image is imported:
```~/Code/napari/napari/layers/image/__init__.py in <module>
----> 1 from .image import Image

~/Code/napari/napari/layers/image/image.py in <module>
     11 from ..intensity_mixin import IntensityVisualizationMixin
     12 from ._image_constants import Interpolation, Interpolation3D, Rendering
---> 13 from ._image_utils import guess_rgb, guess_multiscale
     14
     15

~/Code/napari/napari/layers/image/_image_utils.py in <module>
      3
      4 import numpy as np
----> 5 from napari.layers.image.image import Image
      6 from napari.layers.labels.labels import Labels
      7 from vispy.color import Colormap
```

napari/layers/utils/layer_utils.py
cwood1967: Yes, that is much better. I issued a warning if a ValueError is raised.

napari/layers/utils/layer_utils.py
cwood1967: I added the type hinting to both functions. Is the hinting used at runtime anywhere?

napari/layers/utils/layer_utils.py
cwood1967: I agree about the colormaps. The keyword argument was added to work with strings or vispy Colormaps.

cwood1967:
I updated with the suggestions; I put the functions back into a separate file for now. I like the colormaps as a keyword argument much better. One of the first things I do when opening a multichannel image in some software is set the colormaps to something I like, so guessing doesn't make sense. Gray is used if colormaps aren't provided.

Using _image_utils.py for the functions also caused circular imports.

For the type hinting, are hints enforced at run time in any way? Or used as part of the CI checks?

GenevieveBuckley:
Type hints aren't enforced, but developers using [mypy](http://mypy-lang.org/) do see a benefit once enough code is annotated. (Talley is trying to get more of our layer methods annotated in https://github.com/napari/napari/pull/1097). There was some discussion about whether to include something like mypy in our CI, but this isn't currently the case.

For default colormaps in `stack_to_images`, do you think it's better to set the default to the colormap from the input stack, rather than stick with grey?

cwood1967:
After giving it some thought, I think it is better to use the colormap from the input stack. I would assume the input stack colormap was intentionally set, so that should be reflected in the output.

I modified stack_to_images to keep the original colormap if the colormap keyword argument isn't used.

napari/layers/utils/stack_utils.py
```diff
@@ -0,0 +1,103 @@
...
+            "Can't split along axis {}. The image has {} dimensions".format(
+                axis, num_dim
+            ),
+            UserWarning,
+        )
+        return None
+
+    cmap = stack.colormap
+    if colormaps is None:
+        colormaps = data.shape[axis] * [cmap]
```
jni: Sorry to restart this discussion, but I think when splitting a stack, we should follow the convention of `add_image(..., channel_axis=ax)`. This is handled here:

https://github.com/napari/napari/blob/b94ed99b8f60599aba61579f5711c8bf6e8896d2/napari/components/add_layers_mixin.py#L236-L247

We can definitely argue that choice and change it if we decide it was the wrong choice, but we haven't had any complaints, and either way the behaviours should match, preferably even through shared code.

napari/layers/utils/stack_utils.py
```diff
@@ -0,0 +1,103 @@
...
+
+    imagelist = list()
+
+    for i in range(data.shape[axis]):
+        layer_name = "{:02d}_{}".format(i, name)
```
jni: @cwood1967 although I understand the desire to keep the channel information out front, it really doesn't gel with convention. Ultimately, I think this is a usability fix that needs to happen on how we display layer names, but in my opinion the image number should appear at the end. I'd propose here:

```suggestion
        layer_name = f'{name} layer {i}'
```

or

```python
        layer_name = f'{name} ch{i}'
```

for example.

napari/layers/utils/stack_utils.py
```diff
@@ -0,0 +1,103 @@
...
+            name=layer_name,
+        )
+
+        imagelist.append(image)
+
+    return imagelist
+
+
+def images_to_stack(
+    images: List[Union[Image, Labels]], axis: int = -1
```
jni: It would be much more efficient to combine along `axis=0` by default... Not to mention, it would be guaranteed have the desired effect of adding a slider to the viewer, whereas the converse might result in rgb display (if there are three channels) or a super-thin slice being displayed from each channel (if there is any other number of channels).

napari/layers/utils/stack_utils.py
```diff
@@ -0,0 +1,103 @@
...
+
+    try:
+        new_data = np.stack(new_list, axis=axis)
+    except ValueError as err:
+        warnings.warn("{}".format(err))
+        return
+
+    stack = Image(new_data)
+
+    return stack
```
jni: I don't think we should necessarily be in the business of catching this error, vs letting it bubble up to the caller. If we do the latter, the function becomes super simple:

```python
new_data = np.stack([image.data for image in images], axis=axis)
stack = Image(new_data)
return stack
```

We might want to propagate things such as scale and translation into the new Image, though...

jni (commented):
@cwood1967 thanks for this! I've made some comments proposing a few changes for discussion. Additionally, some of the test scripts you've proposed in comments should be added as actual tests before merging.

The main thing that I think should be implemented within these functions is propagating other settings from the layers, as proposed by @sofroniewn. Several things such as scale and translation should be transferable with a bit of processing. (ie inserting a 1 in the position of the new axis, or removing the value in the position of the removed axis.)

cwood1967:
@jni Thanks for the feedback; I have implementing all of your suggestions. 

My thoughts on propagating settings:
When propagating settings from the input layer, scale and translate should be used from the input layer. I do not think contrast_limits and gamma should be propagated. Blending should be either additive or a keyword argument. I'm not sure about the other settings, but it would be easy enough to transfer from the input layer to the new layer.

Do those make sense, or am I missing anything?

GenevieveBuckley:
Why wouldn't we propagate contrast limits or gamma settings? I think the default should be to propagate everything wherever possible. (I do agree about additive blending through, that seems like a sensible exception to the above rule)

cwood1967:
Based on my use of napari, when a multichannel image is opened/added without channel_axis, the one contrast limit setting is used for all channels. Setting the slider for one channel could cause another channel to look over or under exposed. Finding the min and max for each channel independently would allow each channel to have decent initial contrast limit. The gamma setting is similar: what works for one channel might not work for another.

GenevieveBuckley:
> Based on my use of napari, when a multichannel image is opened/added without channel_axis, the one contrast limit setting is used for all channels. Setting the slider for one channel could cause another channel to look over or under exposed. Finding the min and max for each channel independently would allow each channel to have decent initial contrast limit. The gamma setting is similar: what works for one channel might not work for another.

That's a good argument, I'd say you've persuaded me

cwood1967:
I pushed changes to stack_to_images where I propagated all of settings from add_image. I added keyword arguments for contrast_limits, gamma, colormap, scale, translate, and blending and using the value of the input layer for the others. 

I also added test in napari/layers/utils/_tests.

How should the settings be handled going from multiple layers to one? Considering that each layer could have different scale, translate, contrast, etc, should the function def include all of the settings as keywords like is done with add_image and not try to propagate setting from the input?

cwood1967:
@GenevieveBuckley @jni I added keyword parameters to images_to_stack to it is quite similar to add_image.

napari/layers/utils/stack_utils.py
```diff
@@ -0,0 +1,171 @@
...
+            ),
+            UserWarning,
+        )
+        return None
+
+    if colormap is None:
+        if n_channels == 2:
+            colormap = iter(colormaps.MAGENTA_GREEN)
+        if n_channels > 2:
+            colormap = itertools.cycle(colormaps.CYMRGB)
```
jni: I think there should be an else clause here, for rare cases when an axis might be a singleton.

```suggestion
    if colormap is None:
        if n_channels == 2:
            colormap = iter(colormaps.MAGENTA_GREEN)
        elif n_channels > 2:
            colormap = itertools.cycle(colormaps.CYMRGB)
        else:
            colormap = itertools.repeat('gray')
```

napari/layers/utils/stack_utils.py
```diff
@@ -0,0 +1,171 @@
...
+    else:
+        colormap = iter(colormap)
```
jni: I also think we should cycle here for times when you have more channels than colours:

```suggestion
        colormap = itertools.cycle(colormap)
```

napari/layers/utils/stack_utils.py
```diff
@@ -0,0 +1,171 @@
...
+
+    if blending not in ['additive', 'translucent', 'opaque']:
+        blending = 'additive'
+
+    if scale is None:
+        scale = np.delete(stack.scale, axis)
```
jni: Whoa. Didn't know about `np.delete`. Nice use of it here!

napari/layers/utils/stack_utils.py
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    imagelist = list()
+
+    for i in range(n_channels):
+        layer_name = f'{name} layer {i}'
+        try:
+            color = next(colormap)
+        except IndexError:
+            color = 'gray'
```
jni: I think this try/except can be removed if you implement my suggestions from above.

jni: And actually I think this was incorrect, as the above will raise a StopIteration error. Whatever path we decide to adopt, this should be tested!

napari/layers/utils/stack_utils.py
```diff
@@ -0,0 +1,171 @@
...
+
+        kwargs['contrast_limits'] = contrast_limits[i]
+        kwargs['gamma'] = gamma[i]
+        kwargs['colormap'] = color
+
+        image = Image(np.take(data, i, axis=axis), name=layer_name, **kwargs)
+
+        imagelist.append(image)
```
jni: @cwood1967 you're right that this is now very similar to `add_image, axis=not_none`, in fact, maybe a little too similar? I feel like the two code chunks should match nearly exactly, and possibly even factored out into a single approach. Here's the code for reference:

https://github.com/napari/napari/blob/e43f730170aab1e011c42060c65ba44133f6f18f/napari/components/add_layers_mixin.py#L226-L275

napari/layers/utils/stack_utils.py
```diff
@@ -0,0 +1,171 @@
...
+        'metadata': metadata,
+        'scale': scale,
+        'translate': translate,
+        'opacity': opacity,
+        'visible': visible,
+        'multiscale': multiscale,
+    }
+    new_list = [image.data for image in images]
+    new_data = np.stack(new_list, axis=axis)
+    stack = Image(new_data, **kwargs)
```
jni: This one still needs some work. It's unclear what should be done with things like colormap and contrast limits, but scale and translate should certainly be pilfered from the top image and a scale of `1` inserted at the new axis.

jni (changes requested):
@cwood1967 sorry for the slow-burn reviewing here and thank you for your patience! The good thing is that the code is nice, compact, and clean, which made a full review very easy. My main issues are:

- the `stack_to_images` function needs to be reworked a little to prevent (different!) code duplication with `AddLayersMixin.add_image`, and
- the `images_to_stack` function should preserve scale and translate parameters, even if it discards the others.

The tests should also exercise a few more code paths.

It's very close though! Thanks!

cwood1967:
I noticed that a rgb z stack only has a scale or translate that is 3 elements long (for z, y, and x), so added some checks for rgb images to handle removing the right element from scale and translate. I also added more test cases.

I'm not sure how to handle the code duplication between this and add_layer_mixin. It looks the same, but it is for different cases: adding an image from scratch or using much of the information from an existing layer.

napari/layers/utils/_tests/test_stack_utils.py
```diff
@@ -0,0 +1,124 @@
+from napari.layers.image.image import Image
+from napari.layers.utils.stack_utils import stack_to_images, images_to_stack
+import numpy as np
+
+
+def test_stack_to_images_basic():
+
+    '''
+    Test 2 channels, no colormap
+    '''
```
tlambert03:
```suggestion
    """Test 2 channels, no colormap."""
```
([PEP 257](https://www.python.org/dev/peps/pep-0257/))

Also, if possible, I think it's nice to say what expected behavior you're testing.  like
```python
"""Test that a multichannel stack is split into single-channel layers."""
```

napari/layers/utils/_tests/test_stack_utils.py
```diff
@@ -0,0 +1,124 @@
...
+    for i in images:
+        assert type(stack) == type(i)
+        assert i.data.shape == (10, 128, 128)
+
+
+def test_stack_to_images_rgb():
+    '''
+    Test 3 channels, last index, no colormap - RGB
+    from rgb to single channels
+    '''
```
tlambert03:
```suggestion
    """Test 3 channels, negative index, no colormap - RGB to single channels."""
```

napari/layers/utils/_tests/test_stack_utils.py
```diff
@@ -0,0 +1,124 @@
...
+        assert type(stack) == type(i)
+        assert i.data.shape == (10, 128, 128)
+        assert i.scale.shape == (3,)
+        assert i.rgb is False
+
+
+def test_stack_to_images_4_channels():
+    '''
+    Test 4 channels, first index, colormap
+    '''
```
tlambert03:
```suggestion
    """Test 4 channels, first index, colormap."""
```

napari/layers/utils/stack_utils.py
```diff
@@ -0,0 +1,198 @@
...
+def stack_to_images(
+    stack: Union[Image, Labels],
+    axis: int,
+    colormap: List[Union[str, Colormap]] = None,
+    contrast_limits: List[List[int]] = None,
+    gamma: List[float] = None,
+    blending: str = None,
+    scale: Tuple[float] = None,
+    translate: Tuple[float] = None,
+) -> List[Image]:
```
tlambert03: thank you for the type annotations!! 😍

napari/layers/utils/_tests/test_stack_utils.py
```diff
@@ -0,0 +1,124 @@
+from napari.layers.image.image import Image
```
tlambert03:
```suggestion
from napari.layers import Image
```

napari/layers/utils/stack_utils.py
```diff
@@ -0,0 +1,198 @@
...
+import itertools
+
+import numpy as np
+from napari.layers import Image, Labels
+from vispy.color import Colormap
+from ...utils import colormaps
+
+
+def stack_to_images(
+    stack: Union[Image, Labels],
```
tlambert03:
```suggestion
    stack: Image,
```
(`Labels` _is_ an `Image`... so this is unnecessary)

napari/layers/utils/stack_utils.py
tlambert03: yeah, this is my main concern here too, this is pretty much the same thing as `channel_axis`.  So as @jni suggested, my feeling is that we should pull out the axis splitting logic from `viewer.add_image()` into a helper function that would also take the place of this `stack_to_images` function here, and `viewer.add_images` should call that function. 
 thoughts on that @jni @sofroniewn, @GenevieveBuckley?

The signature of that function could look like this
```python
def split_channels(data: ndarray, axis: int, **kwargs) -> List[LayerData]:
    ...
```

and then the `if channel_axis is not None` clause in `viewer.add_image` could look like:

```python
for datum in split_channels(data, **kwargs):
    self._add_layer_from_data(**datum)
```

and then splitting an image that's already been added to the viewer would be something like:

```python
def split_image(self, image: Image, axis, **kwargs):
    data, meta, layer_type = image.as_layer_data_tuple()
    if not meta["rgb"]:
        meta['scale'].pop(axis)
        meta['translate'].pop(axis)
    for key in ("contrast_limits", "colormap", "blending", "rgb"):
        del meta[key]
    meta.update(kwargs)
    self.layers.remove(image)
    self.add_image(data, channel_axis=axis, **meta)
    # or call the split_channels() directly
```

@cwood1967, regardless of which direction we go there... a good thing to point out in this function is the `data, meta, layer_type = image.as_layer_data_tuple()` which will let you remove a lot of code from your function where you gather up the metadata of the current image.

tlambert03 (commented):
left some thoughts.  I agree with @jni that this is too much like `viewer.add_image` with `channel_axis`... so we should probably extract that into a more reusable function.  I left some strategies for doing so

cwood1967:
I followed the suggestions and made a split_channel function and used it from stack_to_images. I put everything into one file for now, and created a class that inherits AddLayersMixin for the functions. I haven't changed AddLayersMixin.add_image with split_channels yet.

cwood1967:
@tlambert03 @jni @GenevieveBuckley I made the suggested updates that moves the duplicated code into a helper function called by both add_image and stack_to_images. Also added tests and updated doc strings.

jni (approved):
I don't have time for a deep review right now, but after a quick glance this looks great! Super nice structure now.

Some of the docstrings need a bit of fleshing out (e.g. missing description of "data") but I don't think it should hold up the PR.

Thanks @cwood1967!

napari/components/add_layers_mixin.py
```diff
@@ -229,55 +224,13 @@ def add_image(
...
-                    try:
-                        i_kwargs[key] = next(val)
-                    except StopIteration:
-                        raise IndexError(
-                            "Error adding multichannel image with data shape "
-                            f"{data.shape!r}.\nRequested channel_axis "
-                            f"({channel_axis}) had length {n_channels}, but "
-                            f"the '{key}' argument only provided {i} values. "
-                        )
+            layerdata_list = split_channels(data, channel_axis, **kwargs)
```
sofroniewn, line 227: Wow, love being able to use `split_channels` here, very nice!

napari/layers/utils/_tests/test_stack_utils.py
```diff
@@ -0,0 +1,174 @@
...
+from napari.layers import Image
+from napari.layers.utils.stack_utils import (
+    split_channels,
+    stack_to_images,
+    images_to_stack,
+)
+
+
+def test_stack_to_images_basic():
+
```
sofroniewn:
```suggestion
```

napari/layers/utils/stack_utils.py
```diff
@@ -0,0 +1,188 @@
...
+    -------
+    List of LayerData tuples: [(data: array, metadata: Dict, type: str )]
+
+    """
+
+    # Determine if data is a multiscale
+    multiscale = kwargs['multiscale']
+    if multiscale is None:
+        multiscale, data = guess_multiscale(data)
+    kwargs['multiscale'] = multiscale
```
sofroniewn:
```suggestion
    if 'multiscale' not in kwargs or kwargs['multiscale'] is None:
        multiscale, data = guess_multiscale(data)
        kwargs['multiscale'] = multiscale
```

sofroniewn: I'm suggesting this change so code will still run even if `multiscale` not in kwargs

napari/layers/utils/stack_utils.py
```diff
@@ -0,0 +1,188 @@
...
+    n_channels = (data[0] if multiscale else data).shape[channel_axis]
+    kwargs['blending'] = kwargs['blending'] or 'additive'
```
sofroniewn:
```suggestion
    if 'blending' not in kwargs or kwargs['blending'] is None:
        kwargs['blending'] = 'additive'
```
Suggesting this change so code will run even if 'blending' is not in kwargs

napari/layers/utils/stack_utils.py
```diff
@@ -0,0 +1,188 @@
...
+    ----------
+    data : array or list of array
+    channel_axis : int
+        Axis to split the image along.
+    kwargs: keyword arguments
+
+    Returns
+    -------
+    List of LayerData tuples: [(data: array, metadata: Dict, type: str )]
+
```
sofroniewn:
```suggestion
```

napari/layers/utils/stack_utils.py
```diff
@@ -0,0 +1,188 @@
...
+    """
+
+    data, meta, _ = stack.as_layer_data_tuple()
+
+    for key in ("contrast_limits", "colormap", "blending"):
+        meta[key] = None
+
+    name = stack.name
+    num_dim = len(data.shape)
+    # n_channels = data.shape[axis]
```
sofroniewn:
```suggestion
```

napari/layers/utils/stack_utils.py
sofroniewn: This change has been made I think

napari/layers/utils/stack_utils.py
```diff
@@ -0,0 +1,188 @@
...
+        imagelist.append(image)
+
+    return imagelist
+
+
+def images_to_stack(
+    images: List[Union[Image, Labels]], axis: int = 0, **kwargs: Dict,
+) -> Image:
+    """Function to combine selected image layers in one layer
+
```
sofroniewn, line 183:
```suggestion

    The new image layer will get the meta data properties of the first image layer in the stack.
```

napari/layers/utils/stack_utils.py
```diff
@@ -0,0 +1,188 @@
...
+        image = Image(idata, **imeta)
+
+        imagelist.append(image)
+
+    return imagelist
+
+
+def images_to_stack(
+    images: List[Union[Image, Labels]], axis: int = 0, **kwargs: Dict,
```
sofroniewn: I guess I see the value in allowing merging `Images` and `Labels` together, but we might need to be more careful inside this function then - for example right now if the first layer is in the list is a labels then our assumptions about propagating metadata will be off.

I think for this PR I'd be fine dropping the `Labels` from the function signature, keeping the assumption of only images and then revisiting this as part of #1442 where we want to add more functionality to convert between images and labels anyway.

napari/layers/utils/stack_utils.py
```diff
@@ -0,0 +1,188 @@
...
+    axis : int
+        Index to to insert the new axis
+
+    Returns
+    -------
+    stack : napari.layers.Image
+        Combined image stack
+    """
+
+    data, meta, _ = images[0].as_layer_data_tuple()
```
sofroniewn, line 203:
```suggestion
    if len(images) > 0:
        data, meta, _ = images[0].as_layer_data_tuple()
    else:
        return None
```
Curious what people think we should do if the list is empty, I suggest returning `None`, @tlambert03 @jni?

sofroniewn (approved):
This is looking great @cwood1967! I'm approving now too, I've gone through and left some suggestions, most are fairly minor. @tlambert03 might want to have a quick scan through my suggestions too, not that I'm pointing any of this in the wrong direction. I love how we're now using this functionality internally and have pulled all this logic out of `add_image` itself!!

napari/layers/utils/stack_utils.py
```diff
@@ -0,0 +1,188 @@
+import warnings
+from typing import Union, List, Tuple, Dict
+import itertools
+
+import numpy as np
+from napari.layers import Image, Labels
+from napari.utils.colormaps import colormaps
+from napari.layers.image._image_utils import guess_multiscale
+from napari.utils.misc import ensure_iterable, ensure_sequence_of_iterables
```
tlambert03: we've been generally sticking with relative imports, except in tests where we use absolute imports

napari/layers/utils/stack_utils.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+def split_channels(
+    data: np.ndarray, channel_axis: int, **kwargs,
+) -> List[Tuple]:
```
tlambert03:
```suggestion
) -> List[FullLayerData]:
```
and then add `from ...types import FullLayerData`

napari/layers/utils/stack_utils.py
```diff
@@ -0,0 +1,188 @@
...
+    kwargs: keyword arguments
+
+    Returns
+    -------
+    List of LayerData tuples: [(data: array, metadata: Dict, type: str )]
+
+    """
+
+    # Determine if data is a multiscale
+    multiscale = kwargs['multiscale']
```
tlambert03: even though the main "consumer" of this function is currently `viewer.add_image`, which _will_ have 'multiscale' in the `kwargs` dict... to make this more generally usable as a function that can split an array into multiple arrays, we shouldn't assume anything about the `kwargs` parameter.  In other words, the function should execute without error even if kwargs is an empty dict `{}`.  This line and other direct `kwarg[]` access lines will through a `KeyError` if the key doesn't exist.  Use `kwargs.get("key")` instead, which will return `None` if the key doesn't exist

napari/layers/utils/stack_utils.py
```diff
@@ -0,0 +1,188 @@
...
+    data: np.ndarray, channel_axis: int, **kwargs,
+) -> List[Tuple]:
+    """Function to split the data array into seperate arrays along an axis
+
+    Parameters
+    ----------
+    data : array or list of array
+    channel_axis : int
+        Axis to split the image along.
+    kwargs: keyword arguments
```
tlambert03: expand on the explanation of `kwargs`: something like "keyword arguments will override the keys in the ``meta`` dict of each returned layer data tuple".

napari/layers/utils/stack_utils.py
tlambert03: sorry, just saw @sofroniewn comments to the same effect.  I agree.  But instead of 
```python
if 'multiscale' not in kwargs or kwargs['multiscale'] is None:
```
use
```python
if not kwargs.get("multiscale"):
```

napari/layers/utils/stack_utils.py
```diff
@@ -0,0 +1,188 @@
...
+            except StopIteration:
+                raise IndexError(
+                    "Error adding multichannel image with data shape "
+                    f"{data.shape!r}.\nRequested channel_axis "
+                    f"({channel_axis}) had length {n_channels}, but "
+                    f"the '{key}' argument only provided {i} values. "
+                )
+
+        layerdata = (image, i_kwargs, 'image')
+        layerdata_list.append(layerdata)
```
tlambert03, line 97: @jni, @sofroniewn, should we make this function a generator that yields layer data?

napari/layers/utils/stack_utils.py
```diff
@@ -0,0 +1,188 @@
...
+from napari.utils.colormaps import colormaps
+from napari.layers.image._image_utils import guess_multiscale
+from napari.utils.misc import ensure_iterable, ensure_sequence_of_iterables
+
+
+def split_channels(
+    data: np.ndarray, channel_axis: int, **kwargs,
+) -> List[Tuple]:
+    """Function to split the data array into seperate arrays along an axis
+
```
tlambert03, line 28: This function now deserves some more explanation.  For instance, explain that this function will alter the blending mode, and the colormap, unless specifically overridden. etc...  (I know we didn't have that in the `add_image` docstring, but now that we have a dedicated function for it, we should explain what we're doing

napari/layers/utils/stack_utils.py
```diff
@@ -0,0 +1,188 @@
...
+    Returns
+    -------
+    imagelist: list
+        List of Image objects
+    """
+
+    data, meta, _ = stack.as_layer_data_tuple()
+
+    for key in ("contrast_limits", "colormap", "blending"):
+        meta[key] = None
```
tlambert03: maybe `del meta[key]` instead of `= None`? This way, if in the future we change the signature of the `Image()` to no longer allow `None`, this will still work.

napari/layers/utils/stack_utils.py
tlambert03: if you mean dropping the type annotation from the function signature, it A) doesn't change anything about the runtime (as you know) but B) it actually doesn't even change anything about the types! because Labels is an Image.  That is, as far as types are concerned `Union[Image, Labels]` is the same thing as `Image` (but as far as the reader of the signature is concerned ... it's true they might think that this works for labels, which we should at least test if we are going to say that)

napari/layers/utils/stack_utils.py
tlambert03: I think it should raise an exception.  If we return `None`, then the return type should be changed to `Optional[Layer]`, which reminds you that any consuming function needs to make sure it got a response.

napari/layers/utils/stack_utils.py
```diff
@@ -0,0 +1,188 @@
...
+                )
+
+        layerdata = (image, i_kwargs, 'image')
+        layerdata_list.append(layerdata)
+
+    return layerdata_list
+
+
+def stack_to_images(stack: Image, axis: int, **kwargs: Dict,) -> List[Image]:
+    """Function to split the active layer into separate layers along an axis
```
tlambert03: You can leave out the "Function to" part from all the docstrings

tlambert03: also, this function has no concept of an "active" or "inactive" layer.  So the docstring should be more specific.  "Splits a single `Image` layer into a list of Image layers along `axis`"

napari/layers/utils/stack_utils.py
```diff
@@ -0,0 +1,188 @@
...
+        return None
+
+    if axis >= num_dim:
+        warnings.warn(
+            "Can't split along axis {}. The image has {} dimensions".format(
+                axis, num_dim
+            ),
+            UserWarning,
+        )
+        return None
```
tlambert03: I think these should raise exceptions rather than warnings.  Otherwise the return type is incorrect and should be changed.  Alternatively, remove this altogether and let the `split_channels_function` below raise the exception (which it will if they are out of bounds)

tlambert03 (commented):
great progress.  I think this is generally the right organization now.  I have a handful of comments about the implementation still though.  Ping me when ready for another look

napari/layers/utils/stack_utils.py
sofroniewn: I defer to @jni

jni: I’m happy with a list here. We don’t expect this to be infinite or too big for RAM or anything like that, and lists are easier to work with.

cwood1967:
@tlambert03 I updated with the suggestion everyone made. I decided to raise exceptions in the dimension checks to give some indication about what happened.

cwood1967:
@tlambert03 I noticed that colormap is not set in split_channels unless it has a key (None or a value) in kwargs.  It came from add_image which sets the kwargs dictionary, so the key was always there. I just added a line to add the colormap key to kwargs, so colormap is always set.

napari/layers/utils/stack_utils.py
```diff
@@ -49,7 +49,7 @@ def split_channels(
     n_channels = (data[0] if multiscale else data).shape[channel_axis]
 
     kwargs['blending'] = kwargs.get('blending') or 'additive'
-
+    kwargs['colormap'] = kwargs.get('colormap')
```
tlambert03:
```suggestion
    kwargs.setdefault('colormap', None)
```
handy trick ;)

tlambert03 (approved):
This looks great @cwood1967!  Thanks for the hard work on this.  I'll approve and leave it for @jni if he wants to take a last look.
Note, I pushed a couple small changes to the tests and syntax here and there, so be sure to pull first if you make additional changes.

tlambert03:
sorry, just noticed @jni already looked yesterday.  So @sofroniewn, feel free to merge whenever

cwood1967:
I undid the kwargs.setdefault('blending', 'additive'); it was causing errors in the test.

napari/layers/utils/stack_utils.py
```diff
@@ -0,0 +1,212 @@
...
+
+    # Determine if data is a multiscale
+    multiscale = kwargs.get('multiscale')
+    if not multiscale:
+        multiscale, data = guess_multiscale(data)
+        kwargs['multiscale'] = multiscale
+
+    n_channels = (data[0] if multiscale else data).shape[channel_axis]
+
+    kwargs["blending"] = kwargs.get('blending') or 'additive'
```
sofroniewn:
```suggestion
    kwargs['blending'] = kwargs.get('blending') or 'additive'
```

napari/layers/utils/stack_utils.py
```diff
@@ -0,0 +1,212 @@
...
+    kwargs.setdefault("colormap", None)
```
sofroniewn:
```suggestion
    kwargs.setdefault('colormap', None)
```

sofroniewn (approved):
This looks great to me too - made a minor suggested changes on `'` vs `"` which i will commit now. Once tests pass again I will merge. Thanks for all the hard work @cwood1967!!

PR #1324, opened 2020-06-01
Packaging and setup.py refactor

tlambert03:
# Description
Expanding on #1287 (which is included in this PR), and following discussion in https://github.com/napari/napari/pull/1246#issuecomment-636476828, this PR is my proposal for how we might package napari:

- now requires `pip install napari[pyside2]` or `pip install napari[pyqt5]` ... see [this discussion](https://discuss.python.org/t/conditional-package-install-depending-on-other-packages-in-environment/4140) for background on why that's probably our best option at the moment
- moves to a [declarative config](https://setuptools.readthedocs.io/en/latest/setuptools.html#configuring-setup-using-setup-cfg-files) with all of the metadata going into `setup.cfg`.  Note, `setup.py` must unfortunately remain there as a shim because it is [necessary for editable installs](https://twitter.com/pganssle/status/1241161328137515008)... and it also contains legacy fallback behavior for setuptools_scm.
- I was persuaded by [this post](https://caremad.io/posts/2013/07/setup-vs-requirement/) to remove the `requirements` folder in favor of `pip install -e .[tests]`, `pip install -e .[dev]`, etc... defined in setup.cfg.  Makes it easier to single source everything (and `pip install -r requirements.txt` is now the same as `pip install -e .[dev]`) but if people are uncomfortable with all that it's easily reverted :)

discuss! :)

closes #1287

## Type of change

- [x] packaging refactor

# References
here are a few posts I found particularly helpful in trying to understand the ever-changing packaging ecosystem with PEP [517](https://www.python.org/dev/peps/pep-0517/) & [518](https://www.python.org/dev/peps/pep-0518/).
https://www.bernat.tech/pep-517-and-python-packaging/
https://www.bernat.tech/pep-517-518/
https://www.bernat.tech/growing-pain/
https://hynek.me/articles/sharing-your-labor-of-love-pypi-quick-and-dirty/
https://snarky.ca/what-the-heck-is-pyproject-toml/

# How has this been tested?

## Final checklist:


.gitignore
```diff
@@ -124,4 +124,6 @@ venv.bak/
 napari/resources/_qt_resources*.py
 napari/resources/res.qrc
 # ignore all generated themed svgs
-napari/resources/themes
\ No newline at end of file
+napari/resources/themes
+
+napari/_version.py
```
jni, line 129: Interesting...? 😬

README.md
```diff
@@ -24,7 +24,13 @@ We're working on [tutorials](https://napari.org/tutorials/), but you can also qu
 **napari** can be installed on most macOS, Linux, and Windows systems with Python 3.6, 3.7 and 3.8 by calling
 
 ```sh
-$ pip install napari
+$ pip install napari[pyside2]
```
jni: I think this needs a bit of extra narrative. Here's a proposed version:

"""
napari needs a library called Qt to run its user interface (UI). In Python, there are two alternative libraries to run this, called PyQt5 and PySide2. By default, we don't choose for you, and `pip install napari` will not install either. You might already have one of them installed thanks to other scientific packages such as Spyder and matplotlib. Once installed, running `napari` will result in an error message if neither is available.

To install napari *with* a UI framework, you can use `pip install napari[pyqt5]`, `pip install napari[pyside2]`, or `pip install napari[all]`. This last option will choose a framework for you — currently PyQt5, but this could change in the future.
"""

What do you think?

README.md
```diff
@@ -33,6 +39,8 @@ To clone the repository locally and install in editable mode use
 $ git clone https://github.com/napari/napari.git
 $ cd napari
 $ pip install -e .
+# or, to also grab all of the developer tools that we use
```
jni:
```suggestion
# or, to also grab all of the developer tools that we use
# (this is required if you want to contribute code back to napari)
```
?

docs/developers/CONTRIBUTING.md
```diff
@@ -110,7 +110,7 @@ $ git commit -m "my message"
 To run our test suite locally, install test requirements and run pytest as follows:
 
 ```sh
-pip install -r requirements/test.txt
+pip install -e .[tests]
```
jni: I love this hot new thing. =D

docs/source/index.rst
```diff
@@ -36,7 +36,13 @@ Python 3.6 or later by running:
...
 .. code-block:: console
 
-   $ pip install napari
+   $ pip install napari[pyside2]
+
+or, if you'd prefer to use PyQt5 for the GUI backend:
+
+.. code-block:: console
+
+   $ pip install napari[pyqt5]
```
jni: Same comment as above.

btw should we remove '$' from our console blocks for easier copy-paste?

napari/__init__.py
```diff
@@ -1,18 +1,12 @@
-from ._version import get_versions
-
-__version__ = get_versions()['version']
-del get_versions
+try:
+    from ._version import version as __version__
+except ImportError:
+    __version__ = "unknown"
```
jni: What exactly is happening here? I very much dislike 'unknown' as our version, which I don't think ever happened with versioneer?

setup.cfg
```diff
@@ -1,18 +1,116 @@
...
+name = napari
+url = https://napari.org
+download_url = https://github.com/napari/napari
+license = BSD 3-Clause
+license_file = LICENSE
+description = n-dimensional array viewer in Python
+long_description = file: README.md
+long_description_content_type = text/markdown
+author = napari team
+author_email = what_is_it@gmail.com  # FIXME
```
jni: Fix me indeed. LOL

Could be the steering council email? I think it's `napari-steering-council@googlegroups.com`. **Edit:** Yep, just verified.

jni: btw I don't think it would be too hard to make it `steering-council@napari.org`, which would be pretty badass. =P

setup.cfg
```diff
@@ -1,18 +1,116 @@
...
+    Sphinx==3.0.1
+    sphinx-rtd-theme==0.4.3
+    recommonmark==0.6.0
+    scikit-image
+dev =
+    pre-commit>=1.16.1
+    black==19.10b0
+    flake8==3.7.9
+    check-manifest>=0.42
+    %(pyside2)s
```
jni: As mentioned on Zulip, let's make this pyqt5, and let's add `all = %(pyqt5)s`

setup.cfg
```diff
@@ -1,18 +1,116 @@
...
 
 [flake8]
 # Ignores - https://lintlyci.github.io/Flake8Rules
 # E203  Whitespace before ':'  (sometimes conflicts with black)
 # E501 line too long (84 > 79 characters)  (sometimes too annoying)
+# E741 ambiguous variable name
 # W503	Line break occurred before a binary operator
 # C901 McCabe complexity test. Would be nice to reenable, but takes work
-ignore = E203,W503,E501,C901
+ignore = E203,W503,E501,C901,E741
```
jni: Hmm, why?

requirements.txt
```diff
@@ -1 +1,4 @@
--r requirements/default.txt
+# requirements are in setup.cfg as extras
+# https://caremad.io/posts/2013/07/setup-vs-requirement/
+
+-e ".[dev]"
```
jni: Unclear whether this should be `dev` or `all`... Can you comment on why you chose dev?

jni: It might also be worth leaving the other `requirements/*.txt` files with the appropriate `-e "..[]"`, *if that works*.

jni (commented):
@tlambert03 love this!

Just one q, is some of this feasible in the pyproject.toml rather than setup.cfg? To me that seems to be a more logical spot to list requirements. But if that's not where the standard is headed, that's fine too, just wanted to check since the blog post you link to is from 2013, when people were still debating whether there should be a Python 2.8. LOL

jni:
The find missing reqs test is failing. Since that was already pretty outdated we might have to strip it out...? Or do you see an easy fix @tlambert03?

tlambert03:
> Just one q, is some of this feasible in the pyproject.toml rather than setup.cfg? To me that seems to be a more logical spot to list requirements. But if that's not where the standard is headed, that's fine too, just wanted to check since the blog post you link to is from 2013, when people were still debating whether there should be a Python 2.8. LOL

I also had the initial impression that pyproject.toml was where we should be putting all of this stuff now, but that's actually not (yet) the case.  The main point of that file is to 1) (pep 518) tell the build front end what it needs to install for the build-backend to be able to build a wheel and 2) (pep 517) declare what the build backend will be.  It's still up to the build backend though (in our case, still setuptools) to decide what sort of configuration file they accept and where it should be.  For setuptools that is still setup.py and/or setup.cfg.  There is an [open issue](https://github.com/pypa/setuptools/issues/1688) at setuptools, where they are discussing support for putting these things in pyproject... so it may eventually become possible.  But not yet.

the only thing I posted from 2013 was the one discussing requirements.txt vs setup.py ... all the other ones from the "references" section though are very much up-to-date.  I definitely recommend them (the "Sharing Your Labor of Love" post looks like it's from 2013... but it's a living doc that is being updated by many of the main players in the packaging ecosystem).

tlambert03:
> The find missing reqs test is failing. Since that was already pretty outdated we might have to strip it out...? Or do you see an easy fix @tlambert03?

I'm in favor of stripping it.  doesn't look like an active enough project to be a primary tool in our pipeline yet.

requirements.txt
tlambert03: > Unclear whether this should be `dev` or `all`... Can you comment on why you chose dev?

I was kinda reserving `[all]` for the usage that @sofroniewn mentioned in https://github.com/napari/napari/pull/1246#issuecomment-636509692, which is more of a user-facing extra, that might be a place where we put all of the "batteries included" plugins (he mentioned `napari-svg` and stuff) ... whereas `[dev]` was meant to be the "kitchen sink" of tools for napari contributors.

tlambert03: > It might also be worth leaving the other requirements/*.txt files with the appropriate -e "..[]", if that works.

I can definitely add back do this if you wanted it... I think it would work.   my (slight) preference would to be leave it out.  seems unnecessarily repetitive to me (though i get that those used to `pip install -r` would look there first)

setup.cfg
tlambert03: that was actually because flake8 started failing locally for me with that error.  I realize now I was using a newer version of flake8 (3.8.2).  If we keep it pinned to `3.7.9`, it runs fine without ignoring that error.  I'll remove it for now and we can delay that one for later (I think avoiding "ambiguous variable names" is a good rule... but it would require a lot of corrections in our repo)

napari/__init__.py
tlambert03: I forget at the moment exactly what versioneer shows for __version__ when it can't find the tags, but it's also capable of failure. 

the `_version` file will _only_ be unavailable if someone has downloaded the source tree from github, and _not_ installed it with `pip install -e .` (in which case, they basically won't be able to import anything anyway, so it's a very rare case).  However, it gets added when we run `python setup.py sdist bdist_wheel`, so it will always be in our distributions, and it also gets run with `pip install -e .`, so it will be there with local editable installs.

.gitignore
tlambert03: yeah, we don't want to check that in.  basically, `napari/_version.py` will be written by `setuptools_scm` when you run `pip install -e .` or when you run `python setup.py sdist bdist_wheel`, and will create `napari/_version.py` at that moment.  (See also, the comment below)

here's an example of how it looks:
```python
# coding: utf-8
# file generated by setuptools_scm
# don't change, don't track in version control
version = '0.3.1.dev55+g676d8584.d20200602'
```

napari/__init__.py
jni: Hmm. Ok, so the next question is, do we want to make this an editable string that we maintain on each release, skimage style? So now it would read `0.3.4.dev0`. I'm gonna guess the answer is no, but I'd like to have the discussion anyway. 'unknown' is very unsatisfying to me.

Another option that doesn't require ongoing maintenance might be `'not-installed'`?

tlambert03: I'm not opposed to either!  I think `not-installed` is self-descriptive and requires no maintenance.  unless @sofroniewn wants to add "update that string" to the release checklist?

setup.cfg
jni: > it would require a lot of corrections

Wow, really! That'll be fun! =P

napari/__init__.py
tlambert03: will make it "not-installed" now, and we can update it later with `.dev0` if we go that way

setup.cfg
tlambert03: if you're curious:

<details>

```
napari/conftest.py:193:47: E741 ambiguous variable name 'l'
napari/conftest.py:195:33: E741 ambiguous variable name 'l'
napari/layers/labels/labels.py:664:28: F541 f-string is missing placeholders
napari/_tests/test_mouse_bindings.py:90:23: E741 ambiguous variable name 'l'
napari/_tests/test_mouse_bindings.py:106:23: E741 ambiguous variable name 'l'
napari/_tests/test_mouse_bindings.py:162:23: E741 ambiguous variable name 'l'
napari/_tests/test_mouse_bindings.py:178:23: E741 ambiguous variable name 'l'
napari/_tests/test_key_bindings.py:89:22: E741 ambiguous variable name 'l'
napari/_tests/test_key_bindings.py:98:28: E741 ambiguous variable name 'l'
napari/plugins/exceptions.py:143:59: F541 f-string is missing placeholders
napari/plugins/_tests/test_pypi.py:29:5: F541 f-string is missing placeholders
napari/utils/_magicgui.py:108:28: E741 ambiguous variable name 'l'
napari/components/layerlist.py:60:13: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:127:24: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:130:24: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:133:24: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:147:28: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:151:28: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:153:9: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:157:28: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:173:28: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:175:9: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:178:28: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:199:28: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:203:28: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:214:28: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:226:28: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:237:28: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:240:9: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:264:28: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:270:28: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:274:28: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:280:28: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:284:28: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:287:9: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:297:9: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:302:28: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:307:28: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:312:28: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:317:28: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:322:28: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:327:28: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:330:9: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:335:28: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:341:28: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:347:28: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:353:28: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:359:28: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:376:9: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:389:28: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:423:27: E741 ambiguous variable name 'l'
napari/components/_tests/test_layers_list.py:427:27: E741 ambiguous variable name 'l'
napari/components/_tests/test_add_layers.py:22:42: E741 ambiguous variable name 'l'
napari/_qt/_tests/test_qt_layerlist.py:155:9: E741 ambiguous variable name 'l'
napari/_qt/_tests/test_qt_layerlist.py:205:9: E741 ambiguous variable name 'l'
```

</details>

requirements.txt
jni: Yeah I would add the files back, so existing workflows aren't screwed up. It's not so much about future users as people who are already developing with napari.

jni:
> For setuptools that is still setup.py and/or setup.cfg. There is an open issue at setuptools, where they are discussing support for putting these things in pyproject... so it may eventually become possible. But not yet.

Great, thank you so much for that explanation, super clear. 💯 !

jni:
> I'm in favor of stripping it

Ok let's go with this. We *do* need to figure out a way to test whether a requirement belongs it [test] or default...

jni (approved):
Ok thanks @tlambert03! All of my remaining comments can be considered optional — thanks very much for the explanations so far! I'm approving now. 🎉

README.md
```diff
@@ -21,18 +21,40 @@ We're working on [tutorials](https://napari.org/tutorials/), but you can also qu
...
 ```
 
+napari needs a library called [Qt](https://www.qt.io/) to run its user interface
+(UI). In Python, there are two alternative libraries to run this, called
+[PyQt5](https://www.riverbankcomputing.com/software/pyqt/download5) and
+[PySide2](https://doc.qt.io/qtforpython/). By default, we don't choose for you,
+and simply running `pip install napari` will not install either. You *might*
+already have one of them installed in your environment, thanks to other
+scientific packages such as Spyder or matplotlib. If neither is available,
+running napari will result in an error message.
```
sofroniewn: "running napari will result in an error message asking you to install one of them".

docs/source/index.rst
```diff
@@ -32,11 +32,34 @@ installation
...
+
+
+napari needs a library called `Qt <https://www.qt.io/>`_ to run its user
+interface (UI). In Python, there are two alternative libraries to run this,
+called `PyQt5 <https://www.riverbankcomputing.com/software/pyqt/download5>`_
+and `PySide2 <https://doc.qt.io/qtforpython/>`_. By default, we don't choose
+for you, and simply running ``pip install napari`` will not install either. You
+*might* already have one of them installed in your environment, thanks to other
+scientific packages such as Spyder or matplotlib. If neither is available,
+running napari will result in an error message.
```
sofroniewn: "running napari will result in an error message asking you to install one of them".

napari/__init__.py
```diff
@@ -1,18 +1,24 @@
...
-__version__ = get_versions()['version']
-del get_versions
+try:
+    from qtpy import API_NAME
+except Exception as e:
+    if 'No Qt bindings could be found' in str(e):
+        raise type(e)(
+            "No Qt bindings could be found.\n\nnapari requires either PyQt5 or"
+            " PySide2 to be installed in the environment.\nTo install the "
+            'current default default backend (PyQt5), run "pip install '
```
sofroniewn: For "default" we should say `[all]` and then to pick one give the two choices below?

sofroniewn (approved):
Looks great to me @tlambert03 I've left some fairly minor comments about doc strings otherwise looks very nice.

It would be great if the rest of @napari/core-devs could weigh in on this, as it will be a breaking change to how installation works.

We should also probably have a companion PR ready to go to update the installation guide in our tutorials repo with the same language https://github.com/napari/tutorials/blob/master/fundamentals/installation.md

also as soon as this goes in we should release `0.3.4` to ensure the installation instructions are fully accurate

setup.cfg
kevinyamauchi: 😱

setup.cfg
```diff
@@ -1,10 +1,115 @@
...
+[options.extras_require]
+pyside2 = 
+    PySide2>=5.12.3,<5.15.0
+pyqt5 = 
+    PyQt5>=5.12.3,<5.15.0
+# all will be the full "batteries included" extra.
+# though for now, it's just the GUI backend
+all =
+    %(pyqt5)s
+tests =
```
kne42: I find including this in the distribution weird since the users won't ever really need these dependencies. If a user can see this popping up in pip's documentation it could be confusing.

tlambert03: I could get behind removing that.  Originally, this PR had removed all the the requirements folder, but that one specifically seems like a good candidate to leave in requirements/tests.txt if you like that better

README.md
```diff
@@ -21,18 +21,41 @@ We're working on [tutorials](https://napari.org/tutorials/), but you can also qu
...
+
+napari can be installed on most macOS, Linux, and Windows systems with
+Python 3.6, 3.7 and 3.8 using pip:
 
 ```sh
-$ pip install napari
+pip install napari[all]
 ```
 
+napari needs a library called [Qt](https://www.qt.io/) to run its user interface
```
kevinyamauchi: I think this is important discussion. I am also wondering if it is too detailed for the readme. In my mind, the install section readme should provide the minimal information to get going, targeted more at people who just want to use napari and less at "power users" and developers. My concern is that for a lot of users, I think the discussion about multiple qt libraries and the possibility to install without a qt library would raise more questions than it would answer.

Do you think it could make sense to just recommend `pip install napari[all]` and then link to the extended discussion somewhere else in the docs?

Also, have we gotten a lot of issues with people installing from pip and then having qt conflicts with other packages they have installed? Is it worth suggesting to install in a fresh environment?

kevinyamauchi (commented):
I think this looks good! I just left a comment below about how we communicate the install options to the "average" user. I am also happy if everybody feels the current text is the way to go.

README.md
tlambert03: I tend to agree on both accounts.  Maybe @jni wants to weigh in, I pasted his suggestion from https://github.com/napari/napari/pull/1324#pullrequestreview-422278361 into the readme... but I also feel like it's fine to hide some implementation details at first...

also agree on the suggestion for fresh env.

setup.cfg
tlambert03: ok, @kne42, I went back to keeping `dev`, `docs`, `release`, and `test` in the requirements folder... I agree those should not be included as extras in the distribution.  (and it also makes it easier to install them without installing the package in editable mode).

the one difference I left in is that `requirements.txt` is now:
```
-e ".[all]"
-r requirements/development.txt
```

so it's a one-stop-shop for setting up a dev environment.  that ok?

README.md
tlambert03: I updated the readme to move it lower down & put it in an "inset" block ... and added a blurb about environments.  look better?
https://github.com/napari/napari/blob/9557bc3b44eba5fd19898a0963463df87b075c85/README.md

tlambert03:
@sofroniewn, one more thing I changed is to put the codecov task in the linux tests.  Having it run at the end adds around 15-20 minutes to our CI, and that's just because it has to reinstall everything and run all the tests again.  So it feels like we might as well just test coverage the first time.  I did leave check-manifest as a separate task though, which doesn't need to install anything and runs super fast.

README.md
kevinyamauchi: Looks good to me. Thanks, @tlambert03 !

sofroniewn: these changes look good to me @tlambert03!

sofroniewn:
> @sofroniewn, one more thing I changed is to put the codecov task in the linux tests. Having it run at the end adds around 15-20 minutes to our CI, and that's just because it has to reinstall everything and run all the tests again. So it feels like we might as well just test coverage the first time.

Sounds good to me, I had noticed that it was quite slow at the end.

I think this is good to merge, thanks for the feedback all! Will let sit a tiny bit more and then merge and start release process.

PR #1326, opened 2020-06-02
Add QtViewerPushButton to qt public interface

Czaki:
# Description
Add QtViewerPushButton to qt public interface. 
## Type of change

# References

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/qt/__init__.py
+++ b/napari/qt/__init__.py
@@ -1,6 +1,10 @@
 from .._qt.qt_main_window import Window
 from .._qt.qt_viewer import QtViewer
-from .._qt.qt_viewer_buttons import QtNDisplayButton, QtViewerButtons
+from .._qt.qt_viewer_buttons import (
+    QtNDisplayButton,
+    QtViewerButtons,
+    QtViewerPushButton,
+)
 from .threading import create_worker, thread_worker
 
 __all__ = (
@@ -8,6 +12,7 @@ __all__ = (
     'QtNDisplayButton',
     'QtViewer',
     'QtViewerButtons',
+    'QtViewerPushButton',
     'thread_worker',
     'Window',
 )


Czaki:
Maybe all button should be importable from `napari.qt`

sofroniewn:
> Maybe all button should be importable from napari.qt

I'm really not sure about this. Things were made private intentionally because we didn't want to guarantee API stability around things that might be rapidly evolving. Rather than just make things public now, if we are serious about this we should go through and carefully evaluate what classes we want to expose and for what reasons and clean things up.

Thoughts @jni?

tlambert03:
@Czaki, can you comment on why you want them?  They're just _very_ thin wrappers around `QPushButton`... do you want to instantiate a button?  or are you wanting it for an `isinstance` check?

Czaki:
@tlambert03 To got same icons to control viewer as napari. Napari qss is connected with its name. Of course I can create class with same name, but it may case some stupid fail in future (connected with refactoring of napari code).

tlambert03:
so it's about styling?  and what are you hoping to gain by having it be public?  is it that you want your program to fail quickly if we change the name (so that you can fix it), or is it that you _don't_ want the name to be changed... and by making it public you hope it will be less likely?

Czaki:
> is it that you want your program to fail quickly if we change the name (so that you can fix it),

Exactly this. This can be found using automated test. 

So I can still import it from private module. But it is not best practices to use private part of third party packages.

tlambert03:
> But it is not best practices to use private part of third party packages.

yes this is of course true, but that is an implicit contract more than an unbreakable rule.  And in this case, we didn't make it public for exactly the reason that we don't _want_ to give the impression that it's stable.  If all you want is a quick error message for an automated test, then you should feel free to import the private class.  It will serve the same purpose for you whether it's private or not, but if we make it public, then _napari_ is making a statement about stability.

I think a better longterm plan is to move away from class-name based qss selectors, and decide on `#objectName` conventions instead... though that of course requires some agreement

Czaki:
Ok. So I close this PR.

tlambert03:
I do think this would make for a good issue though, if you'd like to open one.  We definitely should have a conversation about what the napari "style API" might look like, and it would be nice to keep this on the radar

Czaki:
@tlambert03 #1332

Issue #1320, opened 2020-05-31
Blocked UI while loading images in notebook


pwinston:
## 🐛 Bug

In Juan's "Intro to Neural Nets" napari notebook scrolling through the 10,000 tests cases is slow because some of the layers are computed on the fly. Unlike #845 (multi-scale) or #1300 (large images) in this case the images are all very small and there is minimal IO.

Instead here the slowness is a machine learning (torch) computation. Accessing the image data via `asarray` triggers a computation that takes some time to produce the data. While waiting for the computation to finish the UI is blocked.

## To Reproduce

1. Get the "napari" branch of https://github.com/jni/noise2self
2. Run `jupyter notebook` then run "Intro To Neural Nets" in the notebooks directory.
3. In cell 15 napari is launched showing a time series of 10,000 small images, 3 layers each.
4. Use the slider to scroll through the time series.

Each image takes maybe 500ms to load on a new macbook. This makes the slider extremely awkward to use, and the overall experience is poor.

## Expected behavior

1. The UI should not be blocked. You should be able to move the slider freely at all times.
2. The `clean_test_dask` layer is not computed on the fly. It should be display as soon as possible, before the other layers are ready. 

With both 1 and 2 working the end result would be almost more like a custom application than just an image viewer. You can quickly scroll through the input images to find one you like, then you pause and wait for the result to be computed and displayed.

Specifically the `clean_test_dask` layer could be shown quickly:

<img width="1085" alt="napari-clean" src="https://user-images.githubusercontent.com/4163446/83364008-ebbbc100-a36b-11ea-9f66-d72f1a63cb2d.png">

Then the other 2 layers added when available:

<img width="1091" alt="napari-all" src="https://user-images.githubusercontent.com/4163446/83364011-f6765600-a36b-11ea-8708-5cca48868f2a.png">

## Additional context

We are writing a document that will propose a solution for the three blocked UI issues: this one, #845 and #1300.

tlambert03:
Just curious: since that’s a delayed dask array, could the the delayed function just be a threadworker?

pwinston:
Not sure what you mean exactly @tlambert03. Do you mean could Juan use`@thread_worker` and make this work as desired with no napari changes?

tlambert03:
[...]

Issue #1300, opened 2020-05-27
Slider is slow with time series of large 2d images


pwinston:
## 🐛 Bug

For a time series of 8k x 8k x float64 images it takes around 200ms to switch slices. For 16k x 16k it takes around 1500ms. In both cases moving the slider through many slices feels slow and laggy.

## To Reproduce

```
# 8k images
napari.view_image(
            np.random.random((2, 8192, 8192)), name='two 8k 2d images'
        )
# 16k images
napari.view_image(
            np.random.random((2, 16384, 16384)), name='two 16k 2d images'
        )
```

**Note**: 16384 is the max texture size on a macbook w/ MD Radeon Pro 5300M 4 GB. Any image larger that will be downsampled to 16384. So we basically just cannot view images larger than that unless they are multiscale.

## Expected behavior

1. Switching images is much faster.
1. To the degree it's not faster the loading should be incremental and interruptible.

You want to be able to interactively "dial in" what slice you care about, moving the slider freely without delays. It's okay if the image takes a bit to fully load once you stop moving.

pwinston:
## Performance Monitoring Results

Using the performance monitoring stuff from #1262 here what it looks like to move between two 8k images using the "next" button:

<img width="1069" alt="napari-slow-8k" src="https://user-images.githubusercontent.com/4163446/83067499-c4e05080-a034-11ea-914e-ab7c27e277e0.png">

**Selected Times In Above Diagram**:
| step | 8k time (ms) | 16k time (ms) |
| --- | --- | --- |
| overall | 208 | 1551 |
| Dims.set_point | 85 | 714 |
| Paint | 114 | 818 | 
| 1 - data.astype | 64 | 625 |
| 2 - clim[0]| 43 | 578 |
| 3 - data /= | 22 | 105 |
| flush_commands | 39 | 125 | 

**Where**:
1. The `data.astype` conversion from `float64` for `float32` in  `napari._vispy.VispyImageLayer._on_data_change`
1. The line `data = (data - clim[0])` which triggers a copy in `vispy.visuals.ImageVisual._build_texture`
1. The line `data /= clim[1] - clim[0]` in `vispy.visuals.ImageVisual._build_texture`

**Notes**:
1. The last line `flush_commands` is waiting on the card to draw. This alone is maybe okay for 8k but way too slow for 16k. It's not clear how much of this is just the card drawing a texture that big, and how much is other overhead that vispy is adding. Would have to be investigated.
[...]

PR #1328, opened 2020-06-02
New document on rendering

pwinston:
# Description
Adds `explanations/rendering.md` based on all the recent rendering discussion and the plans for working on #845, #1300 and #1320.

## Type of change
- [x] New explanation document

## Final checklist:


docs/explanations/rendering.md
```diff
@@ -0,0 +1,344 @@
...
+
+To met our goal of never blocking the UI we need to satisfy two requirements:
+
+1. Always break data into chunks. The exact maximum chunk size is TBD.
+2. Never call `asarray` on user data from the GUI thread since we don't know
+   what it will do or how long it will take.
+
+# Render Algorithm
+
+The renderer will intersect the current view with the dataset to determine the
```
tlambert03: > The renderer will intersect the current view with the dataset to determine the working set.

I'm not sure I completely follow this sentence

docs/explanations/rendering.md
```diff
@@ -0,0 +1,344 @@
...
+makes the internal state of the application visible which is often leads to a
+better user experience.
+
+## Chunked File Formats
+
+The most common types of chunks are blocks of contiguous memory inside a chunked
+file format like **Zarr** and exposed by an API like **Dask**. If an
+image is stored without chunks then reading a 2D rectangle would require
+hundreds of small read operations from all over the file. With chunks you can
+read a rectangular region with a single read operation.
```
tlambert03: i feel like this sentence could be clearer or more precise (but I don't have a suggestion)

docs/explanations/rendering.md
```diff
@@ -0,0 +1,344 @@
...
+as chunks*.
+
+# Loading into RAM and VRAM
+
+There is a two step process to get data into VRAM where we can draw it. First it
+needs to be loaded into RAM and then transferred into VRAM.
+
+Loading into RAM must be done in a thread since we don't know how long it will
+take. For example loading data over the internet or doing a complex calculation
+to produce the data could both take really long time. We are going to use the
```
tlambert03:
```suggestion
to produce the data could both take a really long time. We are going to use the
```

docs/explanations/rendering.md
```diff
@@ -0,0 +1,344 @@
...
+Loading into VRAM is a different story because it must happen in the GUI thread,
+at least that is our assumption while we are using OpenGL. Therefore we need to
+amortize the load over some number of frames. We will set a budget, for
+example 5 milliseconds. Each frame can spend that much time loading data into
+VRAM, it will spend the rest of the frame drawing as normal. In this case it's
+important that no single chunk takes more than 5 milliseconds to transfer.
+
+![paging-chunks](images/paging-chunks.png)
+
+Viewed as a timeline the rendering thread render regularly spaced frames while
```
tlambert03:
```suggestion
Viewed as a timeline, the rendering thread renders regularly spaced frames, while
```
?

docs/explanations/rendering.md
```diff
@@ -0,0 +1,344 @@
...
+![example-1320](images/example-1320.png)
+
+# Example: #845
+
+In #845 we are drawing a multi-scale image which is chunked on disk.
+
+## Chunk Size
+
+It's confusing but there can be different chunk sizes in use at one time. If
+using **Dask** it's chunks are often larger than the file format's chunks. This
```
tlambert03:
```suggestion
using **Dask**, its chunks are often larger than the file format's chunks. This
```

docs/explanations/rendering.md
```diff
@@ -0,0 +1,344 @@
...
+if necessary.
+
+## Octree
+
+To solve #1320 our chunks will be layers. The ChunkManager can write the data
+into the `Image` object for those layers. However with #845 chunks are spatial
+so we need a new spatial datastructure that can keep track of which chunks are
+in memory and store the per-chunk data.
+
+We are doing to use an octree. See
```
tlambert03:
```suggestion
We are going to use an octree. See
```

docs/explanations/rendering.md
```diff
@@ -0,0 +1,344 @@
...
+no stand-in is available the placeholder can be blank or a grid potentially with
+a "loading" animation.
+
+## Beyond Images
+
+We are starting with 2D images but we are going to build the `ChunkManager` and
+octree in a generic way so that we can add in more layer types as we go,
+including 3D images, points, shapes and meshes. 2D images are the simplest case,
+but we believe most the infrastructure can be used by the they other layers
+times.
```
tlambert03: ?

tlambert03 (commented):
exciting read!  very clear thanks :)

docs/explanations/rendering.md
pwinston: Fixed now says just "Dask chunks are often larger"

docs/explanations/rendering.md
pwinston: fixed

docs/explanations/rendering.md
pwinston: Fixed to say `can be used by the other layer types`

docs/explanations/rendering.md
pwinston: Fixed to say "Viewed as a timeline, the rendering thread has regularly spaced frames, while the IO and compute threads load data into RAM in parallel.

docs/explanations/rendering.md
pwinston: Got it

docs/explanations/rendering.md
pwinston: I was thinking geometric interaction but really it doesn't have to be spatial so I changed it to "The renderer will compute the **working set** based on the current view. The working set is the set of chunks that we need to draw to fully render that
specific view."

docs/explanations/rendering.md
pwinston: I used more sentences it now says: "If an image is stored without chunks then reading any given small 2D region of the image requires many different read operations. The bytes for that region are spread throughout the file, intermingled with the bytes from other regions. With chunking you can read a rectangular region with a single read operation."

Chunking is basic stuff to many developers. But I think it's a kind of detail that a someone new to this stuff might not get at first.

sofroniewn (approved):
@pwinston this is document is great. Just read it through again and I really like it. It's very clear.

@toloudis I was wondering if you don't mind taking a look over this as well, I sent you an email with a little more context, but @pwinston is looking to make some improvements to make our rendering non-blocking. It could also be a good time to think about what integration with [agave rendering engine](https://www.allencell.org/pathtrace-rendering.html) might look like in the future if we wanted to provide a vulkan alternative to OpenGL.

toloudis:
This all looks great.  I will attempt to summarize the high level and comment at the same time.  
1. Don't do anything that will block the main UI thread - move work onto other threads.
2. Work that has to happen on the main thread should be minimized and spread out over many main loop cycles, to keep things responsive.

From a front end perspective, "done-ness" might be important to consider.  Users need to know when what they are seeing is the "fully loaded" data.

One thing that I didn't see explicitly mentioned was evicting data from gpu memory (and cpu memory for that matter).  In prior projects I have used fixed memory budgets to strategize removing chunks from memory in addition to loading them efficiently.  Chunks would be eligible for removal if they were occluded or offscreen (e.g. frustum culled) or there was a scheduled LOD swap.

Currently I am dealing with volume image data and the natural chunk format for file reads is still a 2D z slice.  I have not implemented such a strategy in my current volume rendering code but I would definitely do something similar that just reads smaller pieces of data and lets the rendering code just render what's available as it comes in.

Also, I don't have much experience with in-process Python multithreading, so the GIL would be a concern/technical risk for me too.  If the GIL allows concurrency during i/o operations, then that's at least a good start.

pwinston:
Thanks for the comments @toloudis. Your summary seems right.

Showing "doneness" is an interesting point. I've certainly used apps and wondered "is it still loading or this imagery itself blurry?" It might be nice to make that definitive.

Good point about eviction/deletion I didn't get into that at all. I have some ideas for RAM and VRAM based on what I've done in the past. I'll be sure to add those details once we have something working. I do want the document to morph into describing the system as-implemented, not just be a historical thing.

When I started on napari I was like gosh, they get amazingly decent "paging while drawing" behavior just using Dask, which can lazy load numpy arrays. So it was loading the data on the fly while drawing. Especially with a local SSD I think everyone was kind of impressed how that basically does work with *none* of this infrastructure. But when the data was remote, and other slow cases, everything fell apart and the app would freeze up. 

For a while I think people were hoping there was some "quick fix" but reluctantly we concluded we need basically everything in this document. With 2d images working though I'm hopeful we can use the same `ChunkLoader` and octree for the other types of data napari can view: 3d volumes, points, shapes and meshes.

toloudis:
> When I started on napari I was like gosh, they get amazingly decent "paging while drawing" behavior just using Dask, which can lazy load numpy arrays. So it was loading the data on the fly while drawing. Especially with a local SSD I think everyone was kind of impressed how that basically does work with _none_ of this infrastructure. But when the data was remote, and other slow cases, everything fell apart and the app would freeze up.
> 

I have a little bit of experience with dask for loading large files, from my involvement in the aicsimageio library which was recently refactored to use it.  We got excellent results, but still: Local compute + Remote data = sad face.

It is amazing, the hoops we have to jump through to take code from "it works!" to "it works and you don't have to wait!".

sofroniewn:
This is great @toloudis and @pwinston, thanks for weighing in @toloudis! I'm going to merge this now, but people should feel free to continue discussion on this PR or any of the linked issues. Thanks all

PR #1331, opened 2020-06-03
Use correct pyqt version in tests

tlambert03:
# Description
small fix to make sure we're testing the pinned version of pyqt5 on cirrus.
see fail in https://github.com/napari/napari/runs/734105957

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -79,8 +79,8 @@ linux_task:
     - codecov
 
   test_pyqt5_script:
-    - pip uninstall -y pyside2
-    - pip install pyqt5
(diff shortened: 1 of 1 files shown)


jni:
I have no understanding of that codecov-bot comment. Why does it show such carnage and then when you click on it (or in the checks list) it shows no changes???

jni (approved):

tlambert03:
> I have no understanding of that codecov-bot comment. Why does it show such carnage and then when you click on it (or in the checks list) it shows no changes???

yeah, I don't get it either... there are still many things I don't understand about codecov 😅

Issue #716, opened 2019-11-20
Let labels take a list of custom colors


sofroniewn:
## 🚀 Feature
Following up on discussion with @adamltyson on this [image.sc post](https://forum.image.sc/t/specify-labels-colors/31534) I think it would be nice if we allowed users to set their own list of colors for `add_labels` as an alternative to using random colors. If the user has more labels than colors we can just cycle through them.

sofroniewn:
There is another request for this here https://github.com/hci-unihd/covid-if-annotations/issues/5

constantinpape:
As mentioned in the other issue, I would have two use cases for this:
1. adding a mode where some labels in the segmentation layer are hidden, i.e. their alpha value is set to 0
2. Synchronize the colors between a label layer and a points layer.

I also looked at the image.sc post linked in the issue and I realized that for the second use-case the image layer with fixed color map makes more sense (I have just used the labels layer there to make zero transparent, otherwise I don't need any functionality provided by it).

However, for the first issue I think that passing a **dict** mapping label ids to colors would be the best solution. The color of a given label should then be determined like this:
```python
color = color_dict.get(label_id, random_color(label_id))
```
Also, I would need to turn this mode on and off, so it should be possible to set the `color_dict` dynamically:
```python
# set custom colors for the first labels
labels_layer.color_dict = {1: 'red', 2: 'green', 3: 'blue'}

# deactivate the custom colors
labels_layer.color_dict = {}
```

Let me know if this makes sense to you and if it's something that could be implemented in a rather short timeframe ;).

sofroniewn:
> Let me know if this makes sense to you and if it's something that could be implemented in a rather short timeframe ;).

Yes, we should be able to turn this around pretty quickly (I'm imagining an initial PR in a few days from now, reviewed and merged by the end of the week). The more crisply we can specify it the faster.

I'm thinking that we will want to reuse some of the API paradigm that we use in the [points layer](https://napari.org/docs/api/napari.layers.points.html#napari.layers.points.points.Points) and consider adding support for labels `properties` (see #1214) either in or immediately after this PR.

The points layer currently has the following for face and edge color
```python
layer.*_colormap # colormap name, vispy colormap tuple
layer.*_color_mode # str with different modes, DIRECT, CYCLE, COLORMAP
layer.*_color_cycle # list of colors for the cycle
layer.*_color # rgba array, that could also be set to the name of a property
layer.properties # properties dict
```

[...]

PR #1339, opened 2020-06-04
Issue 716: allow using of custom color dict in labelling layer

liu-ziyang:
# Description
Add support for custom color dictionary and switching between a random colormap and custom colormap

## Type of change

# References
https://github.com/napari/napari/issues/716

# How has this been tested?

- Added unit test to check get color behavior in both random and custom color map
- manually checked behavior when given no, one, and multiple color in dictionary

## Final checklist:

### DIFF
--- a/napari/_qt/layers/qt_labels_layer.py
+++ b/napari/_qt/layers/qt_labels_layer.py
@@ -371,7 +371,7 @@ class QtLabelsControls(QtLayerControls):
             self.preserveLabelsCheckBox.setChecked(self.layer.preserve_labels)
 
     def _on_custom_color_check_box_change(self, event=None):
-        """Receive layer model preserve_labels event and update the checkbox.
+        """Receive layer model enable custom color event and update the checkbox.
 
         Parameters
         ----------
--- a/napari/_qt/layers/qt_labels_layer.py
+++ b/napari/_qt/layers/qt_labels_layer.py
@@ -73,6 +73,9 @@ class QtLabelsControls(QtLayerControls):
         self.layer.events.preserve_labels.connect(
             self._on_preserve_labels_change
         )
+        self.layer.events.enable_custom_color_dict.connect(
+            self._on_custom_color_check_box_change
+        )
 
         # selection spinbox
         self.selectionSpinBox = QSpinBox()
@@ -158,7 +161,15 @@ class QtLabelsControls(QtLayerControls):
         button_row.setSpacing(4)
         button_row.setContentsMargins(0, 0, 0, 5)
 
+        self.customColorCheckBox = QCheckBox()
+        self.customColorCheckBox.setToolTip('enable custom color dict')
+        self.customColorCheckBox.stateChanged.connect(
+            self.change_enable_custom_color_dict
+        )
+        self._on_custom_color_check_box_change()
+
         color_layout = QHBoxLayout()
+        color_layout.addWidget(self.customColorCheckBox)
         color_layout.addWidget(QtColorBox(layer))
         color_layout.addWidget(self.selectionSpinBox)
 
@@ -288,6 +299,19 @@ class QtLabelsControls(QtLayerControls):
         else:
             self.layer.preserve_labels = False
 
+    def change_enable_custom_color_dict(self, state):
+        """Toggle enable custom color dict state of label layer.
+
+        Parameters
+        ----------
+        state : QCheckBox
+            Checkbox indicating if custom color dict is enabled in layer.
+        """
+        if state == Qt.Checked:
+            self.layer.enable_custom_color_dict = True
+        else:
+            self.layer.enable_custom_color_dict = False
+
     def _on_selection_change(self, event=None):
         """Receive layer model label selection change event and update spinbox.
 
@@ -345,6 +369,19 @@ class QtLabelsControls(QtLayerControls):
         """
         with self.layer.events.preserve_labels.blocker():
             self.preserveLabelsCheckBox.setChecked(self.layer.preserve_labels)
+
+    def _on_custom_color_check_box_change(self, event=None):
+        """Receive layer model preserve_labels event and update the checkbox.
+
+        Parameters
+        ----------
+        event : qtpy.QtCore.QEvent, optional.
+            Event from the Qt context.
+        """
+        with self.layer.events.enable_custom_color_dict.blocker():
+            self.customColorCheckBox.setChecked(
+                self.layer.enable_custom_color_dict
+            )
 
     def _on_editable_change(self, event=None):
         """Receive layer model editable change event & enable/disable buttons.
--- a/napari/_qt/layers/qt_labels_layer.py
+++ b/napari/_qt/layers/qt_labels_layer.py
@@ -14,7 +14,7 @@ from qtpy.QtCore import Qt
 
 import numpy as np
 from .qt_base_layer import QtLayerControls
-from ...layers.labels._labels_constants import Mode, ColorMode
+from ...layers.labels._labels_constants import Mode, LabelColorMode
 from ..qt_mode_buttons import QtModeRadioButton, QtModePushButton
 from ..utils import disable_with_opacity
 
@@ -161,7 +161,7 @@ class QtLabelsControls(QtLayerControls):
         button_row.setContentsMargins(0, 0, 0, 5)
 
         color_mode_comboBox = QComboBox(self)
-        color_mode_comboBox.addItems(ColorMode.keys())
+        color_mode_comboBox.addItems(LabelColorMode.keys())
         index = color_mode_comboBox.findText(
             self.layer.color_mode, Qt.MatchFixedString
         )
--- a/napari/_qt/layers/qt_labels_layer.py
+++ b/napari/_qt/layers/qt_labels_layer.py
@@ -8,12 +8,13 @@ from qtpy.QtWidgets import (
     QLabel,
     QSpinBox,
     QHBoxLayout,
+    QComboBox,
 )
 from qtpy.QtCore import Qt
 
 import numpy as np
 from .qt_base_layer import QtLayerControls
-from ...layers.labels._labels_constants import Mode
+from ...layers.labels._labels_constants import Mode, ColorMode
 from ..qt_mode_buttons import QtModeRadioButton, QtModePushButton
 from ..utils import disable_with_opacity
 
@@ -73,9 +74,7 @@ class QtLabelsControls(QtLayerControls):
         self.layer.events.preserve_labels.connect(
             self._on_preserve_labels_change
         )
-        self.layer.events.enable_custom_color_dict.connect(
-            self._on_custom_color_check_box_change
-        )
+        self.layer.events.color_mode.connect(self._on_color_mode_change)
 
         # selection spinbox
         self.selectionSpinBox = QSpinBox()
@@ -161,35 +160,39 @@ class QtLabelsControls(QtLayerControls):
         button_row.setSpacing(4)
         button_row.setContentsMargins(0, 0, 0, 5)
 
-        self.customColorCheckBox = QCheckBox()
-        self.customColorCheckBox.setToolTip('enable custom color dict')
-        self.customColorCheckBox.stateChanged.connect(
-            self.change_enable_custom_color_dict
+        color_mode_comboBox = QComboBox(self)
+        color_mode_comboBox.addItems(ColorMode.keys())
+        index = color_mode_comboBox.findText(
+            self.layer.color_mode, Qt.MatchFixedString
         )
-        self._on_custom_color_check_box_change()
+        color_mode_comboBox.setCurrentIndex(index)
+        color_mode_comboBox.activated[str].connect(self.change_color_mode)
+        self.colorModeComboBox = color_mode_comboBox
+        self._on_color_mode_change()
 
         color_layout = QHBoxLayout()
-        color_layout.addWidget(self.customColorCheckBox)
         color_layout.addWidget(QtColorBox(layer))
         color_layout.addWidget(self.selectionSpinBox)
 
         # grid_layout created in QtLayerControls
         # addWidget(widget, row, column, [row_span, column_span])
-        self.grid_layout.addLayout(button_row, 0, 0, 1, 2)
-        self.grid_layout.addWidget(QLabel('label:'), 1, 0)
-        self.grid_layout.addLayout(color_layout, 1, 1)
-        self.grid_layout.addWidget(QLabel('opacity:'), 2, 0)
-        self.grid_layout.addWidget(self.opacitySlider, 2, 1)
-        self.grid_layout.addWidget(QLabel('brush size:'), 3, 0)
-        self.grid_layout.addWidget(self.brushSizeSlider, 3, 1)
-        self.grid_layout.addWidget(QLabel('blending:'), 4, 0)
-        self.grid_layout.addWidget(self.blendComboBox, 4, 1)
-        self.grid_layout.addWidget(QLabel('contiguous:'), 5, 0)
-        self.grid_layout.addWidget(self.contigCheckBox, 5, 1)
-        self.grid_layout.addWidget(QLabel('n-dim:'), 6, 0)
-        self.grid_layout.addWidget(self.ndimCheckBox, 6, 1)
-        self.grid_layout.addWidget(QLabel('preserve labels:'), 7, 0)
-        self.grid_layout.addWidget(self.preserveLabelsCheckBox, 7, 1)
+        self.grid_layout.addLayout(button_row, 0, 0, 1, 4)
+        self.grid_layout.addWidget(QLabel('label:'), 1, 0, 1, 1)
+        self.grid_layout.addLayout(color_layout, 1, 1, 1, 3)
+        self.grid_layout.addWidget(QLabel('opacity:'), 2, 0, 1, 1)
+        self.grid_layout.addWidget(self.opacitySlider, 2, 1, 1, 3)
+        self.grid_layout.addWidget(QLabel('brush size:'), 3, 0, 1, 1)
+        self.grid_layout.addWidget(self.brushSizeSlider, 3, 1, 1, 3)
+        self.grid_layout.addWidget(QLabel('blending:'), 4, 0, 1, 1)
+        self.grid_layout.addWidget(self.blendComboBox, 4, 1, 1, 3)
+        self.grid_layout.addWidget(QLabel('color mode:'), 5, 0, 1, 1)
+        self.grid_layout.addWidget(self.colorModeComboBox, 5, 1, 1, 3)
+        self.grid_layout.addWidget(QLabel('contiguous:'), 6, 0, 1, 1)
+        self.grid_layout.addWidget(self.contigCheckBox, 6, 1, 1, 1)
+        self.grid_layout.addWidget(QLabel('n-dim:'), 6, 2, 1, 1)
+        self.grid_layout.addWidget(self.ndimCheckBox, 6, 3, 1, 1)
+        self.grid_layout.addWidget(QLabel('preserve labels:'), 7, 0, 1, 2)
+        self.grid_layout.addWidget(self.preserveLabelsCheckBox, 7, 1, 1, 1)
         self.grid_layout.setRowStretch(8, 1)
         self.grid_layout.setColumnStretch(1, 1)
         self.grid_layout.setSpacing(4)
@@ -299,18 +302,15 @@ class QtLabelsControls(QtLayerControls):
         else:
             self.layer.preserve_labels = False
 
-    def change_enable_custom_color_dict(self, state):
-        """Toggle enable custom color dict state of label layer.
+    def change_color_mode(self, text):
+        """Change color mode of label layer.
 
         Parameters
         ----------
-        state : QCheckBox
-            Checkbox indicating if custom color dict is enabled in layer.
+        state : QComboBox
+            ComboBox to change between color mode.
         """
-        if state == Qt.Checked:
-            self.layer.enable_custom_color_dict = True
-        else:
-            self.layer.enable_custom_color_dict = False
+        self.layer.color_mode = text
 
     def _on_selection_change(self, event=None):
         """Receive layer model label selection change event and update spinbox.
@@ -370,18 +370,19 @@ class QtLabelsControls(QtLayerControls):
         with self.layer.events.preserve_labels.blocker():
             self.preserveLabelsCheckBox.setChecked(self.layer.preserve_labels)
 
-    def _on_custom_color_check_box_change(self, event=None):
-        """Receive layer model enable custom color event and update the checkbox.
+    def _on_color_mode_change(self, event=None):
+        """Receive layer model color.
 
         Parameters
         ----------
         event : qtpy.QtCore.QEvent, optional.
             Event from the Qt context.
         """
-        with self.layer.events.enable_custom_color_dict.blocker():
-            self.customColorCheckBox.setChecked(
-                self.layer.enable_custom_color_dict
+        with self.layer.events.color_mode.blocker():
+            index = self.colorModeComboBox.findText(
+                self.layer.color_mode, Qt.MatchFixedString
             )
+            self.blendComboBox.setCurrentIndex(index)
 
     def _on_editable_change(self, event=None):
         """Receive layer model editable change event & enable/disable buttons.
--- a/napari/layers/labels/_labels_constants.py
+++ b/napari/layers/labels/_labels_constants.py
@@ -39,12 +39,12 @@ class LabelColorMode(StringEnum):
     """
     LabelColorMode: Labelling Color setting mode.
 
-    RANDOM (default) allows color to be set via a hash function with a seed.
+    AUTO (default) allows color to be set via a hash function with a seed.
 
     DIRECT allows color of each label to be set directly by a color dictionary.
     """
 
-    RANDOM = auto()
+    AUTO = auto()
     DIRECT = auto()
 
 
--- a/napari/layers/labels/_labels_constants.py
+++ b/napari/layers/labels/_labels_constants.py
@@ -35,4 +35,17 @@ class Mode(StringEnum):
     ERASE = auto()
 
 
+class ColorMode(StringEnum):
+    """
+    ColorMode: Color setting mode.
+
+    DIRECT (default mode) allows each color to be set arbitrarily
+
+    DICT allows color to be set via a color dictionary
+    """
+
+    DIRECT = auto()
+    DICT = auto()
+
+
 BACKSPACE = 'delete' if sys.platform == 'darwin' else 'backspace'
--- a/napari/layers/labels/_labels_constants.py
+++ b/napari/layers/labels/_labels_constants.py
@@ -35,17 +35,17 @@ class Mode(StringEnum):
     ERASE = auto()
 
 
-class ColorMode(StringEnum):
+class LabelColorMode(StringEnum):
     """
-    ColorMode: Color setting mode.
+    LabelColorMode: Labelling Color setting mode.
 
-    RANDOM (default mode) allows color to be set via a hash function with a seed.
+    RANDOM (default) allows color to be set via a hash function with a seed.
 
     DIRECT allows color of each label to be set directly by a color dictionary.
     """
 
+    RANDOM = auto()
     DIRECT = auto()
-    DICT = auto()
 
 
 BACKSPACE = 'delete' if sys.platform == 'darwin' else 'backspace'
--- a/napari/layers/labels/_labels_constants.py
+++ b/napari/layers/labels/_labels_constants.py
@@ -39,9 +39,9 @@ class ColorMode(StringEnum):
     """
     ColorMode: Color setting mode.
 
-    DIRECT (default mode) allows each color to be set arbitrarily
+    RANDOM (default mode) allows color to be set via a hash function with a seed.
 
-    DICT allows color to be set via a color dictionary
+    DIRECT allows color of each label to be set directly by a color dictionary.
     """
 
     DIRECT = auto()
--- a/napari/layers/labels/_tests/test_labels.py
+++ b/napari/layers/labels/_tests/test_labels.py
@@ -263,7 +263,8 @@ def test_custom_color_dict():
     assert (layer.get_color(1) == np.array([1.0, 1.0, 1.0, 1.0])).all()
 
     # test disable custom color dict
-    layer.color_mode = 'random'
+    # should not initialize as white since we are using random.seed
+    layer.color_mode = 'auto'
     assert not (layer.get_color(1) == np.array([1.0, 1.0, 1.0, 1.0])).all()
 
 
--- a/napari/layers/labels/_tests/test_labels.py
+++ b/napari/layers/labels/_tests/test_labels.py
@@ -263,7 +263,7 @@ def test_custom_color_dict():
     assert (layer.get_color(1) == np.array([1.0, 1.0, 1.0, 1.0])).all()
 
     # test disable custom color dict
-    layer.enable_custom_color_dict = False
+    layer.color_mode = 'direct'
     assert not (layer.get_color(1) == np.array([1.0, 1.0, 1.0, 1.0])).all()
 
 
--- a/napari/layers/labels/_tests/test_labels.py
+++ b/napari/layers/labels/_tests/test_labels.py
@@ -251,6 +251,22 @@ def test_colormap():
     assert type(layer.colormap[1]) == Colormap
 
 
+def test_custom_color_dict():
+    """Test custom color dict."""
+    np.random.seed(0)
+    data = np.random.randint(20, size=(10, 15))
+    layer = Labels(data, color_dict={1: 'white'})
+
+    # test with custom color dict
+    assert type(layer.get_color(2)) == np.ndarray
+    assert type(layer.get_color(1)) == np.ndarray
+    assert (layer.get_color(1) == np.array([1.0, 1.0, 1.0, 1.0])).all()
+
+    # test disable custom color dict
+    layer.enable_custom_color_dict = False
+    assert not (layer.get_color(1) == np.array([1.0, 1.0, 1.0, 1.0])).all()
+
+
 def test_metadata():
     """Test setting labels metadata."""
     np.random.seed(0)
@@ -301,11 +317,9 @@ def test_selecting_label():
     data = np.random.randint(20, size=(10, 15))
     layer = Labels(data)
     assert layer.selected_label == 1
-    assert (layer._selected_color == layer.get_color(1)).all
 
     layer.selected_label = 1
     assert layer.selected_label == 1
-    assert len(layer._selected_color) == 4
 
 
 def test_label_color():
--- a/napari/layers/labels/_tests/test_labels.py
+++ b/napari/layers/labels/_tests/test_labels.py
@@ -263,7 +263,7 @@ def test_custom_color_dict():
     assert (layer.get_color(1) == np.array([1.0, 1.0, 1.0, 1.0])).all()
 
     # test disable custom color dict
-    layer.color_mode = 'direct'
+    layer.color_mode = 'random'
     assert not (layer.get_color(1) == np.array([1.0, 1.0, 1.0, 1.0])).all()
 
 
--- a/napari/layers/labels/_tests/test_labels.py
+++ b/napari/layers/labels/_tests/test_labels.py
@@ -317,9 +317,11 @@ def test_selecting_label():
     data = np.random.randint(20, size=(10, 15))
     layer = Labels(data)
     assert layer.selected_label == 1
+    assert (layer._selected_color == layer.get_color(1)).all
 
     layer.selected_label = 1
     assert layer.selected_label == 1
+    assert len(layer._selected_color) == 4
 
 
 def test_label_color():
--- a/napari/layers/labels/labels.py
+++ b/napari/layers/labels/labels.py
@@ -7,7 +7,7 @@ from ..image import Image
 from ...utils.colormaps import colormaps
 from ...utils.event import Event
 from ...utils.status_messages import format_float
-from ._labels_constants import Mode
+from ._labels_constants import Mode, ColorMode
 from ._labels_mouse_bindings import draw, pick
 
 from ..utils.layer_utils import dataframe_to_properties
@@ -183,7 +183,7 @@ class Labels(Image):
             contiguous=Event,
             brush_size=Event,
             selected_label=Event,
-            enable_custom_color_dict=Event,
+            color_mode=Event,
         )
 
         self._data_raw = np.zeros((1,) * self.dims.ndisplay)
@@ -194,12 +194,7 @@ class Labels(Image):
         self._background_label = 0
         self._selected_label = 1
         self._selected_color = self.get_color(self._selected_label)
-        if color_dict is None:
-            self.color_dict = {}
-            self.enable_custom_color_dict = False
-        else:
-            self.color_dict = color_dict
-            self.enable_custom_color_dict = True
+        self.color_dict = color_dict
 
         self._mode = Mode.PAN_ZOOM
         self._mode_history = self._mode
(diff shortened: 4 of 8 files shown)


napari/_qt/layers/qt_labels_layer.py
```diff
@@ -158,7 +161,15 @@ def __init__(self, layer):
         button_row.setSpacing(4)
         button_row.setContentsMargins(0, 0, 0, 5)
 
+        self.customColorCheckBox = QCheckBox()
```
sofroniewn: I think here we will want a combo box with the different color modes. We don't expose those right now on the points layer, do we @kevinyamauchi - but do i remember that I saw an example of that on some demo of yours once?

napari/layers/labels/labels.py
```diff
@@ -135,11 +135,15 @@ def __init__(
         blending='translucent',
         visible=True,
         multiscale=None,
+        color_dict=None,
     ):
 
         self._seed = seed
         self._num_colors = num_colors
-        colormap = ('random', colormaps.label_colormap(self.num_colors))
+        self.random_colormap = (
```
sofroniewn: Can we make this private

napari/layers/labels/labels.py
```diff
@@ -340,6 +352,45 @@ def selected_label(self, selected_label):
...
+        )
+
+    @enable_custom_color_dict.setter
+    def enable_custom_color_dict(self, enable_custom_color_dict):
+
+        if not self.color_dict:
+            return
+
+        if enable_custom_color_dict:
+            self.custom_color = {0: 0.0}
```
sofroniewn: Can we move this to a function in https://github.com/napari/napari/blob/master/napari/utils/colormaps/colormaps.py, maybe `color_dict_to_colormap` that takes in the color dict and any other necessary args, like `custom_color` and then returns a vispy colormap. This file should not need to import either vispy `Colormap` or `Color` but should just import that function. We should also be able to test that function independently of the rest of the codebase.

@kevinyamauchi might also have tips on how the `Color` transformations are done, we might have some utilities that he knows of that will make that easier, maybe he can have a look here too.

napari/layers/labels/labels.py
```diff
@@ -159,7 +163,7 @@ def __init__(
         super().__init__(
             data,
             rgb=False,
-            colormap=colormap,
+            colormap=self.random_colormap,
```
sofroniewn: ideally here we already know what colormap we're using and can pass the actual one (i.e. based off of random, or the color_dict)

napari/layers/labels/labels.py
```diff
@@ -340,6 +352,45 @@ def selected_label(self, selected_label):
         self._selected_color = self.get_color(selected_label)
         self.events.selected_label()
 
+    @property
+    def enable_custom_color_dict(self):
```
sofroniewn: This should become `color_mode` which I think for now should either be
`RANDOM` or `DICT`, but in the future could be based off of an attribute in the properties dict. I do actually wonder @kevinyamauchi if since #1281 has merged we want to force this PR to use the properties dict rather than add an additional `color_dict` (in which case we might want to drop that from the API too)

napari/layers/labels/labels.py
```diff
@@ -465,9 +516,20 @@ def _raw_to_displayed(self, raw):
         image : array
             Image mapped between 0 and 1 to be displayed.
         """
-        image = np.where(
-            raw > 0, colormaps._low_discrepancy_image(raw, self._seed), 0
-        )
+        if self.enable_custom_color_dict:
+            u, inv = np.unique(raw, return_inverse=True)
```
sofroniewn, line 533: I think here we need to have some discussion around what we want the behavior to be for labels outside of the color_dict. Options are

1) Make them same as background (i.e. transparent)

2) Throw an error - painting outside of dict not supported

3) Have some other default color they get painted with

4) Use the color they would have had with the random color map

5) Do some mathematics to make the label "appear" inside the color_dict (i.e. treat like a cycle or something).

Unfortunately it's not clear to me what the right behavior should be with the dict approach, but I think overall we should aim for a high performance approach.

I think I like the idea of emitting a warning and using the background color right now, but I'm open to other ideas

sofroniewn (commented):
@ziyangczi this is coming along very nicely - definitely a tricky one, made slightly tricker by the fact that I don't think we've quite figured out yet exactly the behavior and API we want in all cases.

Reading through it I can't help but think that we want to leverage #1281 more and the `properties` dictionary which takes length N arrays with string keys.

Right now if that dictionary has a special key called index that is a list of integers it will map from those integers as labels into the position in their index in the array, if not it will just default as the identity map.

We could similarly add a special key `color` that would take a length N list of strings or an Nx3 or Nx4 array of colors. We could convert that length N thing into an Nx4 RGBA array using https://github.com/napari/napari/blob/e9df14f6c6f09641faa0e2b41f57a0d55632f7f0/napari/utils/colormaps/standardize_color.py#L31 (right @kevinyamauchi)

Given that Nx4 array we could then create a simple vispy colormap with those controls as keypoints.

Then inside `_raw_to_displayed` if we're using the custom method we need to transform the image with the `_label_index` if it exists (see #1281) and then normalize by `N` so that the result is between `[0, 1]`. Hopefully this won't be too slow, but it's probably our best bet for now.

Overall I'd also love to here from @kevinyamauchi if this approach makes sense, I'd really like the to be harmony between how we are thinking about colors, colormaps, and properties here and on `Points` and `Shapes`.

Let me know what you think of these ideas @ziyangczi, we're still figuring this one out as we go a little!!

sofroniewn:
Thinking a bit more, even if we don't go the `properties` route but stick with the `color_dict` route from an API perspective, I think internally we'd want to reuse the `_label_index` and `Nx4` array -> colormap approach anyway, which makes me think that we should really just own it and go the `properties` route and not allow a dedicated `color_dict`, but have a dedicated `color` property!!

kevinyamauchi:
@ziyangczi , thanks for working on this - it's a highly requested feature!

I agree that this is a bit of a tricky one. I like being able to set the label colormap based on the properties. I am not sure about storing the colors in the `Labels.properties`, as I think the properties should contain information about what labeled pixels are and not how they are visualized.

I could see 4 ways to set the mapping of labels to color:
- random on pixel values: colors are "randomly" assigned to each unique pixel value (how it currently works)
- custom on pixel values: the user provides a mapping of pixel value to color
- random on property value: colors are randomly assigned to the unique values of a property
- custom on property value: the user provides a mapping of property value to color

The two properties ones are a bit strange in that there could be a multiple label to one color mapping. However, this could also be interesting because it could be a way to merge labels together visually without altering the underlying data (maybe a step towards superpixels?).

Does that make any sense?

napari/layers/labels/labels.py
kevinyamauchi: I think we do need to somehow give the pixels with labels not in the custom colormap a non-transparent (usually background) color. This is because in the GUI, the user can select a label and start painting and they should see something. I suppose the alternative is the restrict the label picker to things in the color map.

napari/layers/labels/labels.py
kevinyamauchi: We have been trying to move to representing individual colors as RGBA arrays. @HagaiHargil made some nice utilities for converting common color representations to RGBA arrays. Some that you may be interested in:

- [transform_color()](https://github.com/napari/napari/blob/e9df14f6c6f09641faa0e2b41f57a0d55632f7f0/napari/utils/colormaps/standardize_color.py#L31): transform a single color (e.g., common string name, RGB) into an RGBA array
- [transform_color_with_defaults()](https://github.com/napari/napari/blob/e9df14f6c6f09641faa0e2b41f57a0d55632f7f0/napari/layers/utils/color_transformations.py#L20): calls `transform_color()` but you can provide fallback values
- [normalize_and_broadcast_colors()](https://github.com/napari/napari/blob/e9df14f6c6f09641faa0e2b41f57a0d55632f7f0/napari/layers/utils/color_transformations.py#L91): broadcast a color into an array of colors with the specified number of elements.

liu-ziyang: I was originally trying to go with RGBA array representation, but it seems beyond the scope, because I need to change how the label data layer is represented, which is in the colormap and float between [0,1], we would need to go very deep to change the entire flow to work with RGBA arrays, which felt like a major refactor

napari/layers/labels/labels.py
liu-ziyang: when the mode is updated a bit later in the initialization, the colormap would be updated anyway

liu-ziyang:
I would need quite some help in some of these possible reactors discussed, I would like to take a phased approach in converting the color map to a rgba array for example.

regarding the color_dict, I tried merging it in the properties, but it doesn't align well, it requires the color dict to have the same size as the properties, etc

napari/_qt/layers/qt_labels_layer.py
kevinyamauchi: We don't currently expose color mode on the Points layer GUI. However, we do so in the [Vectors layer controls](https://github.com/napari/napari/blob/e9df14f6c6f09641faa0e2b41f57a0d55632f7f0/napari/_qt/layers/qt_vectors_layer.py#L70-L75). Adding the functionality to the Points layer is on the road map though.

napari/layers/labels/labels.py
kevinyamauchi: Okay, that makes sense to me. I think there will be colormap overhaul in the not-so-distant future, so we can address this then.

kevinyamauchi:
> I would need quite some help in some of these possible reactors discussed, I would like to take a phased approach in converting the color map to a rgba array for example.

That makes sense to me! I am happy to help if I can be useful.

> regarding the color_dict, I tried merging it in the properties, but it doesn't align well, it requires the color dict to have the same size as the properties, etc

Yeah, that makes sense to me. For this reason and the reason I mentioned above (I think thematically, properties should describe what the label represents), I agree that we shouldn't put the color_dict in the properties.

napari/layers/labels/labels.py
sofroniewn: how about we map non found labels to black? Seems like a reasonable first choice for now - you see something, but it's maybe not what you want? Can improve in the future.

napari/layers/labels/labels.py
```diff
@@ -465,9 +506,20 @@ def _raw_to_displayed(self, raw):
...
+            image = np.array(
+                [
+                    self.custom_color[x]
+                    if x in self.custom_color
+                    else colormaps._low_discrepancy_image(x, self._seed)
+                    for x in u
+                ]
+            )[inv].reshape(raw.shape)
```
sofroniewn: This works quite nicely. I just did some timing checks using
```python
 image = np.vectorize(lambda x: self.custom_color.get(x, 1.0))(raw)
```
and 
```python
 image = np.vectorize(self.custom_color.__getitem__)(raw)
```
both of which would have been my first impulse (note for the second one you need to make sure that all the labels are in that dict), but in both cases they are slower than the np.unique/ inv approach. I found this helpful comment give some hints to why https://stackoverflow.com/a/16993364 but it makes sense, as we are in a regime here where I think the number of unique variables will be very low, so very nice implementation!

napari/layers/labels/labels.py
```diff
@@ -465,9 +506,20 @@ def _raw_to_displayed(self, raw):
...
-        image = np.where(
-            raw > 0, colormaps._low_discrepancy_image(raw, self._seed), 0
-        )
+        if self.enable_custom_color_dict:
+            u, inv = np.unique(raw, return_inverse=True)
+            image = np.array(
+                [
+                    self.custom_color[x]
+                    if x in self.custom_color
+                    else colormaps._low_discrepancy_image(x, self._seed)
```
```suggestion
                    else 1.0
```
I'd make this change and then adjust the colormap so that 1.0 always maps to black (or potentially some other color), rather than reuse the random color map here too

napari/utils/colormaps/colormaps.py
```diff
@@ -111,6 +117,31 @@ def _low_discrepancy_image(image, seed=0.5, margin=1 / 256):
...
+    Parameters
+    ----------
+    color_dict : dict of label and color string matches
+
+    Returns
+    -------
+    colormap : Colormap with provided control colors
+    custom_color : dict of label to color control point within colormap
+    """
+    custom_color = {0: 0.0}
```
sofroniewn: I think the addition of the background transparent color_dict should happen in the labels layer using something like
```python
if self._background_label is not in color_dict:
    color_dict[self._background_label] = '#000000' # transparent color?
```

sofroniewn: I think also in the labels layer we should have 
```python
if None not in color_dict:
    color_dict[None] = '#000000' #black color?
```
and then inside `custom_color` make `None` get mapped to `1.0` and inside the colormap make `1.0` get mapped to `color_dict[None]`

This way if someone want's to fully customize the coloring using
```python
color_dict = {0: 'white', 1: 'red', 2:'blue', None: 'green'}
```
then the background label would be white, 1, 2 would be red/ blue and all other labels would be green - pretty cool no?

napari/utils/colormaps/colormaps.py
```diff
@@ -111,6 +117,31 @@ def _low_discrepancy_image(image, seed=0.5, margin=1 / 256):
...
+    colors = ['#000000'] + [
```
sofroniewn: Naive question but why doesn't `'#000000'` here mean the background gets mapped to black instead of transparent? That is the hex code for black isn't it?

napari/utils/colormaps/colormaps.py
```diff
@@ -111,6 +117,31 @@ def _low_discrepancy_image(image, seed=0.5, margin=1 / 256):
...
+        color_str for label, color_str in color_dict.items()
+    ]
+    colormap = Colormap(colors)
+    for label, color_str in color_dict.items():
+        color = Color(color_str)
+        for i in range(len(colormap.colors)):
+            if colormap.colors[i] == color:
+                custom_color[label] = colormap._controls[i]
+    return colormap, custom_color
```
sofroniewn: Can we rename `custom_color` to `label_color_index` as it is sort of analogous to how we use `_label_index` in the labels layer

napari/layers/labels/labels.py
```diff
@@ -340,6 +351,36 @@ def selected_label(self, selected_label):
...
+
+        if not self.color_dict:
+            return
+
+        if enable_custom_color_dict:
+            custom_colormap, custom_color = colormaps.color_dict_to_colormap(
+                self.color_dict
+            )
+            self.colormap = custom_colormap
+            self.custom_color = custom_color
```
sofroniewn: Can we rename `custom_color` to the private `_label_color_index` here and elsewhere as it is sort of analogous to how we use `_label_index`

napari/utils/colormaps/colormaps.py
sofroniewn: For the sake of consistency and uniformity I think it would also be nice to use those color converter functions that @kevinyamauchi mentioned earlier to standardize all the values in the `color_dict` to be 1x4 arrays of RGBA values when the dictionary is passed or ever edited.

i.e.
```python
{1: 'red'} #becomes
{1: np.array([1, 0, 0, 1])}
```
I think having our internal representations of color always be the same will help us in the long run. Note here we don't need to actually make the `Nx4` array. Those color transformation utilities should make this conversion easy. We might want to make a dedicated one to standardize color_dicts that we can import into the labels layer and can be independently unit tested

napari/utils/colormaps/colormaps.py
```diff
@@ -111,6 +117,31 @@ def _low_discrepancy_image(image, seed=0.5, margin=1 / 256):
...
+    custom_color = {0: 0.0}
+    colors = ['#000000'] + [
+        color_str for label, color_str in color_dict.items()
+    ]
+    colormap = Colormap(colors)
+    for label, color_str in color_dict.items():
+        color = Color(color_str)
+        for i in range(len(colormap.colors)):
+            if colormap.colors[i] == color:
+                custom_color[label] = colormap._controls[i]
```
sofroniewn: Can you explain why this needs to be a double for loop? Is it because we can't guarantee the order we get items from the dictionary? I wouldn't have thought two passes were needed.

sofroniewn (commented):
Hi @ziyangczi I think this is coming along nicely, I've given some comments including a proposal on how to handle the background color and non-defined labels for the color dictionary (basically by adding a `None` key to the dictionary that gets mapped to the `1.0` or the colormap which all non-found labels then get mapped to. Seems to strike a nice balance between customizability and simplicity.

We don't need to use an `Nx4` colorarray representation here, but I would like us to convert the values in the `color_dict` to `1x4` arrays so that we always represent color as RGBA internally. We might want to make a special convenience function to this dict conversion. Thoughts @kevinyamauchi?

Can you also swap the `enabled/ diasabled` to a color mode as discussed earlier.

I think for now having `color_dict` separate from `properties` is ok, but I do think people will want to color according to properties, so maybe as this gets closer to finishing we think about that more (and that can be it's own PR).

Hope this helps! Nice work again!!

napari/layers/labels/labels.py
liu-ziyang: right now I am using a random value in the same colormap, the problem with black is that it does not exist in the given colormap, maybe we always add black in the custom color map?

napari/layers/labels/labels.py
liu-ziyang: (I didn't expect you caught this, this certainly is beyond my python knowledge, I found it in a stack overflow post :) )

napari/utils/colormaps/colormaps.py
liu-ziyang: the outer loop is looping through the dictionary, and the inner loop is looping through colormap, I don't think there is a lookup function built-in colormap, so I have to do a lookup as a loop

napari/layers/labels/labels.py
liu-ziyang: I tried the black approach, it looks fine to me, gonna push it soon

napari/utils/colormaps/colormaps.py
liu-ziyang: got it, I think I misunderstood some of the comments, the update will be pushed soon

napari/utils/colormaps/colormaps.py
liu-ziyang: i think for transparent it is the alpha value that matters, will push out fix soon

napari/layers/labels/labels.py
liu-ziyang: if we make it another drop down, the panel seems crowded, wonder we can move things around?

sofroniewn: I'd put it on a new line underneath blending. For now we'd just have those two options, but we know we want to add a third.

You might want to play with the `add_vectors_color_by_angle.py` example and see how the property `color mode` works there to get a sense of where this might go
Screen Shot 2020-06-08 at 8 04 50 PM

napari/layers/labels/labels.py
```diff
@@ -190,6 +194,12 @@ def __init__(
         self._background_label = 0
         self._selected_label = 1
         self._selected_color = self.get_color(self._selected_label)
+        if color_dict is None:
+            self.color_dict = {}
+            self.enable_custom_color_dict = False
+        else:
+            self.color_dict = color_dict
```
sofroniewn: I think `color_dict` should be a property with a setter, and when it is set it should trigger the necessary updates (like colormap and refresh) so that data then uses the new colormap.

sofroniewn: I think the two `color_dict` updates that you have inside `color_dict_to_colormap` right now should be inside the color_dict setter and should only occur if those keys aren't present, i.e. we shouldn't overwrite values someone gives us.

```python
    if self._background_label not in color_dict:
          color_dict[background_label] = 'transparent'
    if None not in color_dict:
          color_dict[None] = empty_color
```

napari/utils/colormaps/colormaps.py
```diff
@@ -111,6 +118,39 @@ def _low_discrepancy_image(image, seed=0.5, margin=1 / 256):
...
+    color_dict : dict of label and color string matches
+    background_label : int of background label, default to 0
+    empty_color : str of color used as placeholder when color not present
+
+    Returns
+    -------
+    colormap : Colormap with provided control colors
+    label_color_index : dict of label to color control point within colormap
+    """
+    color_dict[background_label] = 'transparent'
```
sofroniewn: see comment above but I think these should be inside main labels layer not in here

napari/utils/colormaps/colormaps.py
```diff
@@ -111,6 +118,39 @@ def _low_discrepancy_image(image, seed=0.5, margin=1 / 256):
...
 
 
+def color_dict_to_colormap(
+    color_dict, background_label=0, empty_color='black'
+):
+    """
+    Generate a color map based on the given color dictionary
+    Parameters
+    ----------
+    color_dict : dict of label and color string matches
```
sofroniewn: the indentations/ new lines on these doc strings don't match [PEP257](https://www.python.org/dev/peps/pep-0257/) and the [NumPy documentation guide](https://numpy.org/doc/stable/docs/howto_document.html).

kevinyamauchi:
Hey @ziyangczi , sorry for the delayed response! I will take a look first thing tomorrow my time (Basel).

napari/layers/labels/_labels_constants.py
```diff
@@ -35,4 +35,17 @@ class Mode(StringEnum):
     ERASE = auto()
 
 
+class ColorMode(StringEnum):
+    """
+    ColorMode: Color setting mode.
+
+    DIRECT (default mode) allows each color to be set arbitrarily
+
+    DICT allows color to be set via a color dictionary
```
sofroniewn:
```suggestion
    RANDOM (default mode) allows color to be set via a hash function with a seed.

    DIRECT allows color of each label to be set directly by a color dictionary.
```

I think I'd probably swap what we call DICT and DIRECT and use RANDOM, but curious what @kevinyamauchi thinks. Also what would be name the third mode where we allow settings based on a property?

napari/layers/labels/labels.py
```diff
@@ -282,6 +287,30 @@ def properties(self, properties: Dict[str, np.ndarray]):
...
+                None: 'black',
+            }
+            self.color_mode = ColorMode.DIRECT
+        else:
+            if self._background_label not in color_dict:
+                color_dict[self._background_label] = 'transparent'
+
+            if None not in color_dict:
+                color_dict[None] = 'black'
+
```
sofroniewn, line 308: I'd probably move the `transform_color(color_str)[0]` in here and out of `color_dict_to_colormap` too so that we know that each color is valid and in rgba form. I think this will make things like serialization and the `_get_state` call nicer because we'll know with more confidence exactly what goes in there. Thoughts @kevinyamauchi?

napari/layers/labels/labels.py
```diff
@@ -340,6 +370,37 @@ def selected_label(self, selected_label):
         self._selected_color = self.get_color(selected_label)
         self.events.selected_label()
 
+    @property
+    def color_mode(self):
+        """Color mode string"""
+        if hasattr(self, '_color_mode'):
```
sofroniewn: We generally have initialized the layer property with the default value, so in the init I'd have `self._color_mode = ColorMode('direct')` and then you wouldn't need the `hasattr` here, for example see how we treat Blending https://github.com/napari/napari/blob/master/napari/layers/base/base.py#L143.

napari/utils/colormaps/colormaps.py
```diff
@@ -111,6 +118,36 @@ def _low_discrepancy_image(image, seed=0.5, margin=1 / 256):
...
+        for label, color_str in color_dict.items()
+    ]
+
+    colormap = Colormap(colors)
+    label_color_index = {}
+    for label, color_str in color_dict.items():
+        color = Color(transform_color(color_str)[0])
+        for i in range(len(colormap.colors)):
+            if colormap.colors[i] == color:
+                label_color_index[label] = colormap._controls[i]
```
sofroniewn:
```suggestion
    for i, (label, color_str) in enumerate(color_dict.items()):
        label_color_index[label] = i / (len(colors) - 1)
```

The Colormap deterministically sets control points based on the order of the colors passed to it, evenly spacing them out between 0.0 and 1.0 so I don't think you actually need to do that second search through the colormap to find the color. I just tried the above code and it works fine (produces an identical dictionary up to floating point precision). Probably simpler just to have this single pass

napari/utils/colormaps/colormaps.py
```diff
@@ -111,6 +118,36 @@ def _low_discrepancy_image(image, seed=0.5, margin=1 / 256):
...
+
+    Returns
+    -------
+    colormap : Colormap
+        Colormap constructed with provided control colors
+    label_color_index : dict of int
+        Mapping of Label to color control point within colormap
+    """
+    colors = [
+        transform_color(color_str)[0]
```
sofroniewn: See comment about about moving `transform_color(color_str)[0]` into the dict setter

sofroniewn (commented):
@ziyangczi I think this is really close now, I left a naming suggestion that I'd love @kevinyamauchi to look at tomorrow (basically to use `RANDOM` and `DIRECT` for the modes we have now).

I also suggested moving the color transform into the dict setter and made a suggestion on simplifying the colormap creation code.

Very nice stuff, we should be able to get this in soon.

napari/layers/labels/_labels_constants.py
kevinyamauchi: I agree we should call the "allows color to be set via a color dictionary" mode `DIRECT`.

I am not sure what the best name would be for the "(default mode) allows color to be set via a hash function with a seed" mode. I think `RANDOM` is okay, but I don't think I would know how to interpret it without the docstring. Maybe `AUTO` and then include a slight variant of the description @sofroniewn suggested: "automatically assigns label colors via a hash function with a seed"

For setting with a property, I imagine we might want to allow the user to have an automated mapping between property value and color and also to set the mapping directly (see #1348). If that is the case, maybe `DIRECT_PROPERTY` and `AUTO_PROPERTY`?

kevinyamauchi: One more question: sorry for not catching this earlier, but should we name the Enum something different. We use `ColorMode` in the same way in the `Points`, `Vectors` and soon to be `Shapes` layers. Since this has a different usage, perhaps it would be best to give it a different name.

Maybe something like `LabelColorMode`? Too long? Sorry, I'm not good at naming...

edit: or `LabelsColorMode`?

napari/layers/labels/_tests/test_labels.py
```diff
@@ -251,6 +251,22 @@ def test_colormap():
...
+    layer = Labels(data, color_dict={1: 'white'})
+
+    # test with custom color dict
+    assert type(layer.get_color(2)) == np.ndarray
+    assert type(layer.get_color(1)) == np.ndarray
+    assert (layer.get_color(1) == np.array([1.0, 1.0, 1.0, 1.0])).all()
+
+    # test disable custom color dict
+    layer.color_mode = 'direct'
+    assert not (layer.get_color(1) == np.array([1.0, 1.0, 1.0, 1.0])).all()
```
kevinyamauchi, line 268: Is it possible that label 1 will randomly get assigned white causing this test to fail?

napari/layers/labels/labels.py
kevinyamauchi: I agree that `Labels.color_dict` colors should always have the same type. I think it would be nice to create a util function that lives in `napari.layers.utils.layer_utils` that ensures the values in the dictionary are RGBA numpy arrays with shape (4,) (you can use `transform_color(color_str)[0]` as @sofroniewn suggested.). We would make use of this in `Points` and `Shapes` when we address #1348 .

napari/layers/labels/labels.py
```diff
@@ -159,7 +162,7 @@ def __init__(
         super().__init__(
             data,
             rgb=False,
-            colormap=colormap,
+            colormap=self._random_colormap,
```
kevinyamauchi, line 168: Should we figure out what the correct colormap is going to be and pass it to the `Image` layer so we don't have to reset it below if a color_dict is passed? If it makes things too difficult, it's probably a minor optimization and can be skipped.

napari/layers/labels/labels.py
```diff
@@ -320,6 +349,7 @@ def _get_state(self):
                 'properties': self._properties,
                 'seed': self.seed,
                 'data': self.data,
+                'color_dict': self.color_dict,
```
kevinyamauchi, line 358: Is it important to save the color mode as well? A `Labels` layer from the state output here would always be in the color dict mode, even if the viewer was in the mode where the colors are automatically assigned. This would reproduce the displayed image, but the interactivity would be different (e.g., layer controls would be different on GUI).

kevinyamauchi (commented):
This is looking really nice, @ziyangczi ! I have made some comments below, but I think it's really close!

- I am wondering if when no `color_dict` is passed to the init method, `Labels.color_dict` should default to the `Labels._random_colormap` that is used rather than one with everything translucent as is currently done. My thought behind this is that I can imagine we will eventually have a color picker in the GUI and the user might start with the automatic mapping and then want to make changes. As it is now, switching to the color dict color mode would "reset" the `color_dict` to the translucent one. 
- it looks like you just haven't had time to make them yet, but just in case, don't forget to update the attributes and parameters docstrings (`Labels`, `Viewer.add_labels()`, `napari.view_labels()`)

napari/layers/labels/_labels_constants.py
liu-ziyang: I was hoping that we could unify the practices and later using a single enum. But for now I can rename it

napari/utils/colormaps/colormaps.py
liu-ziyang: I see, I was a bit scared to make this assumption, but as long as it works :)

napari/layers/labels/_tests/test_labels.py
liu-ziyang: though possible, it is really against the chance since the color is in float format and gets randomly assigned by hash

napari/layers/labels/labels.py
liu-ziyang: I think it would result in either duplicate code or recursive call, so I will defer on this one

napari/layers/labels/labels.py
liu-ziyang: I don't think I follow the rationale of moving it here, can you please explain a bit more about serialization and _get_state call? feel free to point me to docs that are relevant as well. I think I may not be making some design decisions right because I lack some bigger picture of how things are connected

napari/utils/colormaps/colormaps.py
```diff
@@ -2,10 +2,17 @@
 from typing import Tuple, List
 
 import numpy as np
-from vispy.color import BaseColormap, Colormap, get_colormap, get_colormaps
+from vispy.color import (
+    Color,
```
sofroniewn: `Color` import no longer used and so Flake8 is failing

napari/layers/labels/labels.py
sofroniewn: Yup agreed, that's a little funny. Right now we don't have color mode in the points layer serialized. One reason is we didn't add it to the main API. I'd be fine merging without for now and then revisiting.

napari/layers/labels/labels.py
sofroniewn: We should add some docs around color handling within napari @kevinyamauchi, right now there isn't any. The rationale was that by having everything converted to an RGBA representation there would then be one known way that other tools could interact with colors from napari. For example, one day if a plugin were to be passed colors via napari that plugin could assume that the color was an RGBA representation instead of a string like `lime-green` which that plugin might not know how to deal with. Putting the call inside the labels layer ensures that we'll achieve that uniformity here. Does that make sense?

napari/layers/labels/labels.py
sofroniewn: Good point @kevinyamauchi, but I agree with @ziyangczi i think it is fine to overwrite it later

napari/layers/labels/_tests/test_labels.py
sofroniewn: given that we have an `np.random.seed` call in this function we know that it won't happen for this example, so I think it is fine to leave as is, but you could @ziyangczi add a comment about this in case one day someone else looks at this and wonders the same thing

napari/layers/labels/_labels_constants.py
```diff
@@ -35,4 +35,17 @@ class Mode(StringEnum):
     ERASE = auto()
 
 
+class LabelColorMode(StringEnum):
+    """
+    LabelColorMode: Labelling Color setting mode.
+
+    RANDOM (default) allows color to be set via a hash function with a seed.
```
sofroniewn: I liked the suggestion from @kevinyamauchi to make this `AUTO` instead of `RANDOM` I think that is a bit more accurate.

napari/layers/labels/labels.py
```diff
@@ -135,11 +135,21 @@ def __init__(
...
         self._seed = seed
+        self._background_label = 0
         self._num_colors = num_colors
-        colormap = ('random', colormaps.label_colormap(self.num_colors))
+        self._random_colormap = (
+            'random',
+            colormaps.label_colormap(self.num_colors),
+        )
+        self._color_mode = LabelColorMode.AUTO
+        self._color_dict = {
```
sofroniewn: Do we actually need this initialization here if this get's overwritten in line `203`, `self.color_dict = color_dict`? If so can we just initialize as `{}` instead? I don't love having the `'transparent'` and `'black'` sneak in there because they use the string representation instead of the RBGA representation

napari/layers/labels/labels.py
sofroniewn: Tests are now failing because `color_mode` was added to serialization and not the main API. I think for now I'd keep the `color_mode` off of `_get_state` which is still private and then we can fix this later, when we fix it for points too. It's easier to add this later than to expand the API now and then have to remove it later.

kevinyamauchi: Yeah, that makes sense to me!

napari/layers/labels/labels.py
ziyangczi: I managed to decouple the color dict and color mode, and removed the init from here

napari/layers/labels/labels.py
liu-ziyang: I managed to decouple the color dict and color mode, and removed the init from here

napari/layers/labels/labels.py
liu-ziyang: it does, refactored

napari/utils/colormaps/colormaps.py
```diff
@@ -111,6 +116,29 @@ def _low_discrepancy_image(image, seed=0.5, margin=1 / 256):
     return image_out
 
 
+def color_dict_to_colormap(colors):
+    """
+    Generate a color map based on the given color dictionary
+    Parameters
```
sofroniewn, line 123: Missing new line needed here

sofroniewn:
```suggestion
    Generate a color map based on the given color dictionary

    Parameters
```

sofroniewn (commented):
Looks like our serialization test is failing right now
```python-traceback
napari/layers/_tests/test_serialize.py:41: ValueError
>               assert np.all(getattr(layer, prop) == getattr(new_layer, prop))
E               ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```
because of the rgba array truth comparison in the color dict.

We might just be able to do

```python
np.testing.assert_allclose(getattr(layer, prop), getattr(new_layer, prop))
```
and that will do everything we need

liu-ziyang:
> Looks like our serialization test is failing right now
> 
> ```
> napari/layers/_tests/test_serialize.py:41: ValueError
> >               assert np.all(getattr(layer, prop) == getattr(new_layer, prop))
> E               ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
> ```
> 
> because of the rgba array truth comparison in the color dict.
> 
> We might just be able to do
> 
> ```python
> np.testing.assert_allclose(getattr(layer, prop), getattr(new_layer, prop))
> ```
> 
> and that will do everything we need

unfortunately that didn't work, but I worked around by checking dict separately

sofroniewn (approved):
Great @ziyangczi all tests are passing this looks good to me! Will give one final spin tomorrow morning and then merge. Well done, and thanks for putting up with all the comments/ making the requested changes!!

kevinyamauchi (approved):
This looks great to me, @ziyangczi ! 🎉  Thank you so much for this nice PR and your patience with our comments!

napari/_qt/layers/qt_labels_layer.py
```diff
@@ -288,6 +302,16 @@ def change_preserve_labels(self, state):
...
             self.layer.preserve_labels = False
 
+    def change_color_mode(self, text):
+        """Change color mode of label layer.
+
+        Parameters
+        ----------
+        state : QComboBox
+            ComboBox to change between color mode.
+        """
```
jni, line 312: This docstring is outdated — should be fixed in a followup PR. The argument name 'text' isn't quite descriptive enough... `new_mode`?

napari/layers/labels/labels.py
```diff
@@ -340,6 +376,34 @@ def selected_label(self, selected_label):
...
+                custom_colormap,
+                label_color_index,
+            ) = colormaps.color_dict_to_colormap(self.color_dict)
+            self.colormap = custom_colormap
+            self._label_color_index = label_color_index
+        elif color_mode == LabelColorMode.AUTO:
+            self._label_color_index = {}
+            self.colormap = self._random_colormap
+        else:
+            raise ValueError("Unsupported Color Mode")
```
jni, line 398: This line will never be hit, because of `color_mode = LabelColorMode(color_mode)`, above. This will raise the following error if `color_mode` is not 'auto' or 'direct':

```pytb
~/conda/envs/pip7/lib/python3.7/enum.py in _missing_(cls, value)
    575     @classmethod
    576     def _missing_(cls, value):
--> 577         raise ValueError("%r is not a valid %s" % (value, cls.__name__))
    578
    579     def __repr__(self):

ValueError: 'hello' is not a valid LabelColorMode
```

So, the above two lines should be removed. In general, we should be paying more attention to our test coverage!

napari/layers/labels/labels.py
```diff
@@ -465,9 +529,22 @@ def _raw_to_displayed(self, raw):
...
+                    else self._label_color_index[None]
+                    for x in u
+                ]
+            )[inv].reshape(raw.shape)
+        elif self._color_mode == LabelColorMode.AUTO:
+            image = np.where(
+                raw > 0, colormaps._low_discrepancy_image(raw, self._seed), 0
+            )
+        else:
+            raise ValueError("Unsupported Color Mode")
```
jni, line 547: As above — these two lines should never be hit.

jni:
@ziyangczi 

> regarding the color_dict, I tried merging it in the properties, but it doesn't align well, it requires the color dict to have the same size as the properties, etc

This is actually fine, because the properties dict is required to have exactly one property per label.

I think this PR should be reworked to match more closely what @kevinyamauchi did with point colouring, but with the additional "random" mode to match existing behaviour. Specifically, there could be a "direct" mode that grabs the colour from a single property (it does not need to be named "color" @sofroniewn, it could have any name and be settable), "colormapped" that uses a colormap, or "auto"/"random" that uses the current behaviour.

What do people think?

sofroniewn:
I responded in #1356, felt like it was easier to more the convo there and keep it in one place

sofroniewn:
Hi @ziyangczi I think there are a few small things to put in a quick follow up PR for this one before we release. It should be very quick. My apologies for not catching them during review the first time.

I just spoke with @jni about the naming and we think renaming `color_dict` -> `color` both for the API kwarg and layer property is the way to go. It's more in line with how we use the word color in other parts of the API (points/ shapes) and more reusable then `color_dict` which was very specific.

Also I noticed now we didn't add `color` to the `view_labels` function which we should have, and we didn't add a doc string for `color` to any of `Labels`, `add_labels`, and `view_labels`. Note these doc strings all need to match.

Can you also drop the two `raise ValueError` that @jni flagged, and fix the `change_color_mode` doc string and variable name, then I think we'll be done.

Thanks so much!!!

liu-ziyang:
opened up subsequent PR to address this

napari/layers/labels/labels.py
liu-ziyang: This check also serves as future proof catchall trap. If we are to add another color mode and this would help capture unexpected behavior during debugging. I found the catch under "Mode" useful when I implemented a new mode, so I am inheriting the same paradigm here as well

jni: @ziyangczi Ah, I see that rationale, which I hadn't appreciated before. Thanks. I think though that this is the wrong way to go about it. We shouldn't have inaccessible parts of the code in case they become accessible later. If you want an "early warning system" for unimplemented colormodes, I would suggest making a test parametrised on an iteration of all the possible modes in the enum — which will fail if you just let things fall through here.

PR #1342, opened 2020-06-05
Add properties to shapes

### LINKED ISSUE #177: Annotations for shape layers
## 🚀 Feature
I think it would be useful to be able to add annotations to the shape layers. In short, I would like to be able to associate a short block of text with a given shape that is then displayed when the user mouses over the shape.

## Motivation
When collaborating on a data set, there is often a need to show a collaborator a particular observation (e.g., "this cell looks weird"). Adding an annotation to a shape that bounds a region of interest would be a great way to share such observations.

## Pitch
* This can build on top of the forthcoming shape layer (#100)
* Initially, the annotation can be displayed as a string in the tool bar (i.e., a tool tip). In the future, it might be nice to be able to display the text in a separate window or in the console to allow for more verbose messaging.
* Ideally, the annotations should be able to be saved and loaded with the complete viewer state (e.g., coordinates, clims settings, processing steps)

## Alternatives
I think this could be achieved purely in the notebook (e.g., the notations are a parameter stored in the data object representing a given shape that can be queried), but I think this is best handled in the viewer.

@sofroniewn, this is the feature we discussed earlier today. Please let me know if anything is unclear.

kevinyamauchi:
# Description
This PR adds properties to the `Shapes` layer and the ability to map display colors 
to properties.

## Type of change

# References
Closes #177 

# How has this been tested?

## Final checklist:


kevinyamauchi:
@sofroniewn , I have added properties to the shapes layer (mappable colors still in progress) and in doing so I realized that the `Shapes` layer `add()` method has shape display properties as kwargs. This is pretty different from the `Points` layer, where the new points just get the `current_*_color` value. Should do the same in `Shapes`?

One one hand, it's good to be consistent. On the other hand, it seems like we wouldn't want to get ride of the `shape_type` kwarg, so at the point, maybe we might as well leave the others just to avoid making a breaking change to the API?

napari/view_layers.py
```diff
@@ -544,18 +576,6 @@ def view_shapes(
         {'opaque', 'translucent', and 'additive'}.
     visible : bool
         Whether the layer visual is currently being displayed.
-    title : string
```
sofroniewn: Why are these doc strings being cut? I think they appear on every `view_*` method, and args are still there

sofroniewn:
>  I have added properties to the shapes layer (mappable colors still in progress) and in doing so I realized that the Shapes layer add() method has shape display properties as kwargs. This is pretty different from the Points layer, where the new points just get the current_*_color value. Should do the same in Shapes?

I think we do want to bring Shapes in line with Points, though I now recognize we have quite a bit of duplicate code. I think that's ok for now, but I think we want to start taking the principles of #1343 and making small internal classes that can be reused not through inheritance but through attributes and so moving away from properties like `edge_color_cycle` and `edge_colormap` to `edge_color.cycle` and `edge_color.map`. Not something I'd do in this PR, but maybe something to think about. I could even imaging a single Rendering class which had all the coloring, opacity, rendering mode, etc inside of it.

napari/view_layers.py
kevinyamauchi: Sorry, that was an accident! Fixed.

kevinyamauchi:
@sofroniewn , I am sorry for the late response - it's been a rough week.

I agree that we need to revisit how we're organizing the code related to different features, especially with these annotation layers (shapes, points, vectors) which have a lot of different possible states. I think a strategy using mixins as you've described makes a lot of sense.

In the meantime, I have finished adding properties and colors that are mappable to those properties to the `Shapes` layer using the approach we have been using thus far. I need to do a bit of cleanup of the docs and see if we can round out the test coverage, but it would be great if you could take a look when you have time.

sofroniewn:
> I think a strategy using mixins as you've described makes a lot of sense.

Here I wasn't thinking mixins, I was thinking more about have small classes and using composition instead of inheritance. See also comment here https://github.com/napari/napari/issues/1356#issuecomment-643786967.

I'll go through the rest of this promptly

sofroniewn (commented):
Overall this looks good to me, as you noted some doc strings still need clean up (i.e points -> shapes in a few places). I think it will be good to get this in and brings shapes in line with points and then maybe think about one last refactor where we can group all our color related properties onto a reusable object. Thoughts @jni?

napari/layers/shapes/_shapes_constants.py
```diff
@@ -35,6 +35,22 @@ class Mode(StringEnum):
     VERTEX_REMOVE = auto()
 
 
+class ColorMode(StringEnum):
```
jni, line 38: Not necessarily needed for this PR, but I think clearly there should only be one ColorMode enum shared by all the layers that need it!

napari/layers/shapes/shapes.py
```diff
@@ -420,6 +616,66 @@ def edge_color(self, edge_color):
         self._set_color(edge_color, 'edge')
         self.events.edge_color()
 
+    @property
+    def edge_color_cycle(self) -> np.ndarray:
+        """Union[list, np.ndarray] :  Color cycle for edge_color.
+        Can be a list of colors defined by name, RGB or RGBA
+
```
jni:
```suggestion

        Can be a list of colors defined by name, RGB or RGBA
```

napari/view_layers.py
```diff
@@ -544,7 +576,7 @@ def view_shapes(
...
+        title : string
```
jni: @kevinyamauchi not fixed, the indentation is wrong here!

jni (approved):
Yeah the code looks great, no great surprise since it's identical (?) to the points PR which had a very thorough review. =) I agree that the code duplication is a concern, but actually, it's nice to get this in *with the tests*, because then you can refactor with complete abandon and be sure that the refactor gives you identical functionality. Very excited about getting this all totally in line (ditto for labels!).

sofroniewn (approved):
Just looked through again, looks great to me! Will be nice to get this in for 0.3.5!!

sofroniewn:
Just noting windows failure is popping up a lot right now, independent of this PR, so will go ahead and merge anyway. Nice job here @kevinyamauchi!!

kevinyamauchi:
Thanks @sofroniewn for the review and @jni for getting this across the line!

Issue #1325, opened 2020-06-02
proposal: reducing complexity of models


kne42:
## 🚀 Feature
Reduce complexity of current model code by factoring out user-facing API as a view, business logic as a controller, and data structures as a model.

## Motivation

Model code is getting excessively complex. Files are typically over 1000 lines long and difficult to parse/keep track of what's going on. Currently, our models are dealing with data management, presentation logic (i.e. Python magic & data conversion to make it better for users), and business logic.

Right now I feel like we have to make tradeoffs between nice UX and maintainability for the API and separating these components makes it easier to maintain without sacrificing our nice API. Take for example how we handle private variables: either they go without any documentation and can be confusing to the developer or they wind up awkwardly in the public-facing documentation where it doesn't really belong and confuses the user.

Furthermore, I think that the current user-facing API functions more like a view since we do a lot of input/output conversions and use custom Python syntax for a nicer user experience. Perhaps the best example of this is the LayerList.

## Pitch

I want to split out the current model code into model, view, and controller classes.

### Model
The model would be a simple data class that notifies when its fields are changed.

```python
from dataclasses import dataclass

@dataclass
class PointsLayerModel(BaseLayerModel):
    data: Array = notifier()
    mode: Mode = notifier()
    ...
```

This makes it easy to keep track of all relevant information pertaining to the state. You can even easily convert this state into a dictionary for saving and sharing with `dataclasses.asdict`. Keeping this class simple may also make it easier to handle settings and create autogenerated code from the state.

### View
The view would be a simple shell that mimics what you can do with the GUI. It emits events when a user interacts with it and may clean the input before handing it off to the controller.

```python
class Points(Layer):
    @property
    def mode(self):
        return str(self._model.mode)

    @mode.setter
    def mode(self, mode):
        self.events.mode(value=Mode(mode))
```

### Controller
The controller handles the business logic, connects the model to the view, and modifies the model. In a lot of cases it will be a simple event handler that prevents setters from being called multiple times (this can be autogenerated).

```python
class PointsLayerController(BaseLayerController):
    def on_mode_change(self, event):
         with event.source.events.mode.blocker():
             self.model.mode = event.value

    def on_data_change(self, event):
         data = event.value
         data_view = ...
         # calculation of data view goes here
         self.model.data_view = data_view
```

## Alternatives

We could factor out the view without separating the model and the controller and it wouldn't be too big of an issue, but I believe that it would make it easiest to maintain in the future.

## Additional context

[...]

PR #1343, opened 2020-06-05
Factor out ImageSlice and ImageView from Image

pwinston:
# Description
Factor out two classes from `Image` in preparation for async loading for #1320.

## Type of change
- [x] Refactoring - no functional changes

# Explanation

`Image` is complicated and with async and async-multiscale it's going to get *way* more complicated. We need to start factoring out things that will need to change or grow.

Here we factor out two new classes `ImageSlice` and `ImageView`. Together they let us organize the 4 related images that `Image` stores, and makes it much harder to get into a broken state. Most importantly it gives us classes that can grow particularly `ImageSlice`.

The new classes might look like overkill to some people, but if you focus on the diff of `Image.py` it cleans up things a lot. I think we'll need to a lot more like this to prepare for growth.

| Before | After |
|---|---|
| _data | _data |
| _data_view | _slice.image.view |
| _data_raw | _slice.image.raw |
| _data_thumbnail | _slice.thumbnail.view |
| N/A | _slice.thumbnail.raw |

Problems before:
* `Image` class is messy, too much inline.
* The 4 _data values are confusingly named
    * `_data_raw` is not the raw version of `_data`!
    * Not clear which are the raw + view pairs
        * i.e. that `_data_thumbnail` is like `_data_view`, both are viewable and not raw.
    * The per-slice attributes are not grouped or distinguished enough from `_data`.
* Nothing enforces view will get updated when raw is updated.
* No where to put new Slice specific functionality.

Before set image and thumbnail raw:
```
    self._data_raw = np.clip(image, 0, 1)
    self._data_view = self._raw_to_displayed(self._data_raw)
    self._data_thumbnail = self._raw_to_displayed(	
        np.clip(thumbnail_source, 0, 1)	
    )	
```
After:
```
    self._slice.image.raw = np.clip(image, 0, 1)
    self._slice.thumbnail.raw = np.clip(thumbnail_source, 0, 1)
```

# Context and Comments

Pylint by default restricts you to at most 7 attributes per class. 7 is not random it's the [magical number](https://en.wikipedia.org/wiki/The_Magical_Number_Seven,_Plus_or_Minus_Two). More than 7 and it's hard to hold them in your head at the same time. The object becomes harder to understand, reason about, extend, test and debug. `Image` today has more than 70 attributes, public and private. You can see them in the debugger. A lot are through inheritance. 70 is way too many to understand or reason about. 

People think 7 is impossibly low but it's not that hard. When you add your 8th attribute you look at the those 8 and say is this really a list of 8 *totally unrelated* attributes? Is there really no connection between any of them? No pattern, no grouping? Or can you identify 2 or more than in fact do have some meaningful connection? If so combine those attributes into something new, a named tuple, dataclass, class, even a dict.

This sounds somewhat pointless at first, didn't you just move the attribute? How is that better? The benefit accrues over time because now there is a place to put functionality that deals with only those related attributes. Functionality that doesn't clutter the base object. Functionality that only has access to the small number of related attributes in its own class.

This new object then grows and then also splits, and so on recursively. The original object is now a tree of well factored objects with focused purposes. Each one taken alone is far easier understand, reason about, extend, test and debug than original kitchen-sink spaghetti object. Not to mention each one is potentially resable in ways the original class wasn't.

7 seems like a small number but consider if you only have 3 levels of hierarchy (a.b.c) then 7 gives you ~400 attributes (7 + 7^2 + 7^3). So your 3-level object could have much more state than `Image` yet no single object has more then 7 attributes. 

Some suggestions:
* Prefer small objects, a hard or soft limit of 7 attributes.
* Group attributes that are related: named similarly, updated together, derived from each other, etc.
* Prefer composition to inheritance, almost always.
* Prefer inheritance from small base classes, not big ones. Rather than inherit from `Image` have `Image` hold objects which have tiny inheritance hierarchies of their own.
* Prefer shallow inheritance hierarchies to deep ones.

Every rule and guideline can be broken in some cases if there's some strong reason to make an exception. However in general you need to constantly be pushing on these rules because code will constantly be growing in ways that violate them. This is almost entirely a problem related to **growth over time**. No one sits down and types up a class that has 50 attributes. They are added 1, 2 or 3 at a time by different people over a period of years. It's *almost always* easier to add attributes than to thoughtfully refactor things. Just like it's easier to throw something into the garage than to re-organize what's already there. And what can it hurt going from 17 to 18 attributes? What's one more?

# How has this been tested?

## Final checklist:


napari/layers/image/_image_slice.py
```diff
@@ -0,0 +1,43 @@
...
+    image : ImageView
+        The main image for this slice.
+
+    thumbnail : ImageView
+        The smaller thumbnail image for this slice.
+
+    Example
+    -------
+        # Create with some default image.
+        slice = ImageSlice(default_image)
```
d-v-b: this shadows the python built-in `slice` object, maybe change to `imslice`?

pwinston: I was like why didn't it warn me? But it was only in the comment. I changed it to `image_slice`, best to not shadow built-ins in examples.

sofroniewn:
I like the idea of pulling out the `ImageSlice` as it's own thing. I have a couple questions, just to help me see where this might be going in the future, but makes sense. It also only affects our private attributes and is fully backwards compatible for our public API.

1) Ultimately where do you see methods like `_set_view_slice` living? A method like that could take in the full nD image data and the "slicing" information and then generate an `ImageSlice`? In a world where we are imagining an Image Controller (as in #1325) is that something that a controller would do? Or is it an update method on `ImageSlice`?

2) Where do you see `_update_thumbnail` living? Now that could live in `ImageSlice` (I can probably figure out how to remove that `dims` call, all the info should be inside the `ImageSlice` (only ambiguity is if slice is a 3D image or an rgb image). It could also live in more "controller-like" parts of the code too after #1325.

Maybe more of a comment, the `raw` vs `view` is also a little bit unfortunate. We don't need it for the image layer, but we need it for labels right now, and even more so after #1339, but ideally we wouldn't need it at all if we fixed some things on the vispy end. It's really very much just a preprocessing of the data before sending to GPU, but unfortunately we also sort of need it for thumbnail generation, so it lives where it lives. I guess it makes sense now to add the extra class, but it would be great to be able to remove it in the future. 

Right now we have two failing tests, one in `test_swappable_dims` because we reference `_data_raw` which as a short term fix could just be renamed, but we also might want to consider other options. Not quite clear to me why the labels test fails, but I can poke around. Labels subclasses Image, so maybe not surprising things we affected there too.

sofroniewn:
Maybe one further comment though on

> Image is complicated and with async and async-multiscale it's going to get way more complicated. 

Agreed with `async-multiscale`, but for `async` in general I was imagining affecting more of the base layer and the `_set_view_slice` calls and the location of the `np.asarray` calls as we want async to work for all our layer types. i.e. I wouldn't have thought this changes would really be required for #1320 (not that they aren't good changes though!) and maybe it does make sense to think about async for Images first

pwinston:
> raw vs view is also a little bit unfortunate.

I didn't look very closely at labels, it's broken right now. It defines `_raw_to_displayed` but that doesn't exist in the`Image` anymore. I can put it back and yet still pass it down to `ImageView` to do the conversion. I'll have to fix that. I can think about the other questions once that's fixed.

pwinston:
> affecting more of the base layer 

I haven't designed it all, but I wasn't picturing any changes to base for a long time. #1320, #845 and #1300 are all just images. For a long time the `ChunkLoader` will only work for `Image`. That's how Talley's prototype had it also, and I'm roughly following that idea.

What did you picture was going to happen in the base layer? After all three of those issues are fixed, octree is totally done, then I think we start thinking about other layers.

In my mind the main thing for #1320 (async non-multiscale images) is something in `Image` which holds the slice image data but it can be in 3 states: Empty, Waiting, and Done. So it's **Empty** to start, default blank image, then you decide the data is needed you request it from the `ChunkLoader` and you are **Waiting**, then when the data is finally there it's **Done** and you can draw it.

So I didn't design it fully out, but this refactor seemed necessary because for this empty->waiting->done cycle to work all that stuff has to be clear and rock solid. The only way with 4 different `_data_*` variables I thought was really very not good.

But I very much don't have the whole system in my head. I'm really trying to just slot in sync cleanly with as few other changes, but it has to be solid and reliable. Making a fragile thing async will be a nightmare.

pwinston:
I added a "Context and Comments" section to the main PR description with some observations related to this reactoring. My experience with big complex OO codebases.

sofroniewn:
> What did you picture was going to happen in the base layer? After all three of those issues are fixed, octree is totally done, then I think we start thinking about other layers.

I guess I meant that something like the [thread_worker](https://github.com/napari/napari/compare/master...tlambert03:wip/threaded-set-view-slice#diff-18a0aa4953281ea93c8962f6ab57fbbdR530-R534) call in @tlambert03 example would actually happen in the base layer in some generic `refresh` / `set_view_slice` method and then each layer would be responsible for defining it's own `_set_data` and `_get_data` methods that would then get executed appropriately. 

Definitely best to just implement for one part of the Image layer first though, and then gradually expand the refactor to the other layers, so fine to continue on the trend, but ultimately I think we want to end in a place where we are reusing as much of the asyn machinery across all our layer types, no? 

> I added a "Context and Comments" section to the main PR description with some observations related to this refactoring. My experience with big complex OO codebases.

Looking now.

> Some suggestions:
>
> Prefer small objects, a hard or soft limit of 7 attributes.
> Group attributes that are related: named similarly, updated together, derived from each other, etc.
> Prefer composition to inheritance, almost always.
> Prefer inheritance from small base classes, not big ones.
> Rather than inherit from Image have Image hold objects which have tiny inheritance hierarchies of their own.
> Prefer shallow inheritance hierarchies to deep ones.

I find this very compelling, along with other points about how we got to where we are now with organic growth. I think there is a unique opportunity now to refactor napari to put us on a path towards a more maintainable codebase, we have quite a lot of stuff in there right now, a lot of eyeballs on the project, looking at many different parts of this (think event handler, MVC, world coordinates, layer groups), and it's not too late to clean things up. I think we should do this slowly, but I worry that if we don't do this now the project will suffer a year from now.

One concrete question is how do you think about "methods" on objects as opposed to "attributes", are they used interchangeably in this context? i.e. does the "rule of 7" also extend to all our functions, such as `_set_view_slice` or is it really just about properties like `data` etc?

Maybe @royerloic and @tlambert03 want to read through these comments and respond too. Ultimately if we do decide to codify these design principles, as always, I'd love to turn them into a stand alone doc we can point developers towards :-) But really great job laying these principles out @pwinston I'm learning so much from reading them.

I will make a separate comments as I review the actual PR.

sofroniewn (approved):
I really like this direction for the codebase. I find the chaining together of simple objects `layer._slice.image.view` much more natural and easy to remember that properties like `layer._data_view` or what might be `layer._slice_image_view`, and I really like how that leads to very transparent documentation with everything clearly having its own clear doc strings etc.

I've also pulled down the code and run through a battery of interactive demos and examples and everything worked fine. Our tests are also passing, and I love how this PR is fully backwards compatible!!! Great job there @pwinston.

I'd like @tlambert03 and @royerloic to review too before merge as it is a different way of organizing the codebase. If possible @pwinston and you're ok continuing to develop off of a branch of this PR for now we might also wait to merge until juan can take a look too.

I am also curious about where you see methods like `_set_view_slice` and `_update_thumbnail` living. That might help me get a better sense of how this continues to evolve (see https://github.com/napari/napari/pull/1343#issuecomment-639839349), but I get its also a work in progress.

pwinston:
Quick point, I will come back to the others later, but I've never heard anyone have an issue with too many methods in a class. I think `ndarray` has around 50 methods. That's not a problem, those are just all sorts of useful things you can do with the same bits of data.

pwinston:
Closed by mistake!

The problem with attributes is the more there are, the more possible states the object can be in. And this rises very quickly. With 5 boolean variables the object can be in one of 32 states, with 7 it's 128 states. 

Suppose you have an object and it has booleans `_is_visible`, `_is_frozen` and `_is_saved` or something. You can picture in some methods there will be if statements like "if visible and frozen but not saved" then do this. Even with 3 booleans that's getting complicated.

The FP people say "no mutable state". They are right, an object that doesn't change after construction is much easier to think about because its behavior never changes. With mutable state the way an object behaves is a potentially complex interaction between its current state and its code, you might have to fully understand both. 

In effect a large complicated object with lots of mutable state means you are no longer doing OO inside that object, *your one single object has itself become a program*. And not an OO program or an FP program but a 1970's structured program with a bunch of global variables (the attributes) and a bunch of very non-pure functions with side-effects (the methods that read and modify those attributes).

Objects and their relationships are the bones and the skeleton and the attributes are the meat. Generally someone at the start of the project creates the obvious domain objects, like in a game you create `Player` and `Enemy` and `Level`. But if you're not careful they will slowly grow into bloated kitchen-sinks. You need to keep adding bones as you go, all the time. But these are less obvious bones, they are not domain objects, they are made-up things related to the implementation rather than the problem.

napari/layers/image/_image_slice.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def __init__(
+        self,
+        view_image: ArrayLike,
+        image_converter: Optional[ImageConverter] = None,
+    ):
+        """
+        Create an ImageSlice with some default viewable image.
+
+        Parameters
```
kne42: normally you don't have a docstring for `__init__` and put it in the class docstring (simply move the parameters section to there)

pwinston: Just learned that in the other PR. Moved them for ImageSlice and ImageView here.

pwinston:
Going back to @sofroniewn's question "where do you see methods like _set_view_slice living?" Certainly `_set_view_slice` logic doesn't belong in `Image` itself, but I think you need to resolve #1325 to say here it should really go. #1325 is a huge question mark as to how things will ultimately look.

But if you factor `_set_view_slice` code into a helper that `Image` holds (whether `ImageSlice` or some new `Slicer` thing) you at least have it split out once you decide where you want to move it. Same with "Where do you see _update_thumbnail living", it should clearly be factored out of `Image` somehow, but then it depends on #1325 where it ultimately goes.

With this PR I'm trying to do the least I need to for #1320 and not just clean up `Image` for its own sake. But between #1320, #1300, and #845 I will probably need to keep making some changes to `Image`. Rendering just can't be bolted on, it's pretty invasive addition.

I don't have a good handle on #1325 but I think a lot is related to establishing the major layers in the system. I think what people are saying is they want napari's API to be as high as possible. I think this is kind of an API-first attitude which is unusual for a lot of GUI programs.

d-v-b:
re: where should `_set_view_slice` live, I always found this part of the API pretty confusing depending on too much mutable state (looking at you, `dims`...). 

Part of the problem is that historically `set_view_slice` is doing way too much work with the data -- is the data rgb?, is it a pyramid?, etc. Instead of this logic living in the layer code, it makes way more sense to push this logic into an object that manages access to the data. E.g.: 

```python
class DataSource:
  def __init__(self, data: array-like, space: object that describes the embedding of data in a physical space):
   self.data = data
   self.space = space

   # use self.space to figure out how to resample self.data at coords
   # this is different for different types of data -- pyramid vs rgb vs labels
   def sample(self, coords: some representation of positions in physical space): Future[array-like]
         ...
        return fut: Future[array-like]
          
# now in the image layer code: 
class Image:
...
# this can be the same for all layers!
    self.view_slice = await self.datasource.sample(coords: some representation of positions in physical space)
...
```

I feel like a lot of the bloat in the Image layer is representative of missing abstractions that hopefully gets solved by world coordinates. What is data (abstractly)? Abstractly, data is a collection of values in some space, _not_ a numpy array. What does it mean to scroll through data (abstractly)? Scrolling through data is resampling the data at sequence of planes (or volumes...) in space, _not_ array indexing.  

I think it would be best if the layer has absolutely no clue how spatial coordinates, or some sparse representation of coordinates (like a definition of the angle + offset of a plane) are turned into displayable intensities. The layer knows how to ask its data "what are your intensities $here?" and then it is responsible for preparing those intensities for display.

pwinston:
@d-v-b that's moving things in a useful direction. Although I think #1325 is saying that `Image` can't have a `self.view_sice` like you are showing. For multiple reasons but one is there could be N views of same data (e.g. XY, XZ, YZ). Really that's the open question, is `Image` the data being viewed or is it a view of the data? And not just `Image` it's all our layers. What are layers? Are they data or are they views of that data? I think #1325 is trying to resolve that?

d-v-b:
> Although I think #1325 is saying that Image can't have a self.view_sice like you are showing. For multiple reasons but one is there could be N views of same data (e.g. XY, XZ, YZ).

Right, that's a good point. So I'm not sure where `set_view_slice` should live, or what could take its place on the layer side...

> What are layers?

Yeah, this is a great question! I would love to see a high-level justification / explanation of what layers are, and how they relate to to the underlying image data.

pwinston:
I think a general pitfall when writing GUI apps is creating an object model that's based too closely on your "user interface objects". You can point to "a layer" on the screen, it's that little rectangle with eyeball on it, so surely there's a class which implements everything that rectangle does?

I think for any GUI program that's any easy path to get suckered down. But it's particularly problematic if you want a versatile and powerful API, because almost by definition the users of your API are *not* constructing a GUI that looks exactly like your GUI. They are not constructing a GUI at all, or they are making a very different GUI.

They don't want to construct their thing out of your *user interface objects*, they want to construct it out of some more general set of objects. Your GUI is the marquee example of what you can make with those objects, but it's like the LEGO model on the box, it's just one thing you can build.

It's a balancing act though because it's super easy to get bogged down making stuff too general. This is the idea of [YAGNI](https://martinfowler.com/bliki/Yagni.html) it says focus on what you need today, the rest can wait. But I think making an API is kind of a special case, you *do* need to look ahead to what people are going to build with that API, and you *do* need to acknowledge it won't be exactly what you yourself already built.

sofroniewn:
There was universal enthusiasm for this PR in team meeting today, @royerloic and @0x00b1 both said they love the direction.

As the PR is fully backwards compatible and only affects private objects I'd like to merge tomorrow. That will let @pwinston make https://github.com/pwinston/napari/pull/4 directly against the repo and will let me incorporate these changes into a new version of #1276 world coordinates which includes these design principles. If @jni has requested changes they can easily be included in a standalone PR or in the upcoming chunk loader PR.

napari/layers/image/_image_slice.py
```diff
@@ -0,0 +1,51 @@
...
+        # Create with some default image.
+        image_slice = ImageSlice(default_image)
+
+        # Set raw image or thumbnail, viewable is computed.
+        image_slice.image = raw_image
+        image_slice.thumbnail = raw_thumbnail
+
+        # Access the viewable images.
+        draw_image(image_slice.image.view)
+        draw_thumbnail(image_slice.thumbnail.view)
```
jni: Can this be reformatted into doctest-compatible syntax? See examples [here](https://numpy.org/doc/stable/docs/howto_document.html), search for "Examples".

napari/layers/image/_image_view.py
```diff
@@ -0,0 +1,87 @@
...
+
+    Example:
+    --------
+        # Set the viewable image and raw image to the same thing.
+        image = ImageView(view_image)
+        #  or
+        image.view = view_image
+
+        # Set the raw image, it will compute the new viewable one.
+        image.raw = raw_image
```
jni: Same here: use doctest format:

https://numpy.org/doc/stable/docs/howto_document.html

And the section name should be

```
Examples
--------
```

napari/layers/image/_image_slice.py
```diff
@@ -0,0 +1,51 @@
...
+    """The slice of the image that we are currently viewing.
+
+    Right now this just holds the image and its thumbnail, however future async
+    and multiscale-async changes will likely grow this class a lot.
+
+    Parameters
+    ----------
+    view_image : ArrayLike
+        The default image for the time and its thumbail.
+    image_converter : ImageConverter, optional
```
jni, line 17: I think in the docstring (rather than the actual type annotation) I would prefer this to be `Callable, mapping array->array`, or something in that vein. When I read this, I see a crazy new object. Nothing in this docstring actually tells me, as a reader and longtime user of the scipy ecosystem, that `lambda x: x % 256` is an ImageConverter!

jni: The description below should indeed explain what imageconverter should be, rather than how imageslice uses it, since readers typically read this docstring to figure out what they need to pass / can pass this constructor.

jni:
> In effect a large complicated object with lots of mutable state means you are no longer doing OO inside that object, your one single object has itself become a program. And not an OO program or an FP program but a 1970's structured program with a bunch of global variables (the attributes) and a bunch of very non-pure functions with side-effects (the methods that read and modify those attributes).

100%! 🙌  One of the reasons I hate OO in general. =P

Thanks for this PR @pwinston! I really like the discussion above. But @sofroniewn's final comment alarmed me enough to pull me from my idyllic slumber. =P

I agree with the general principles in the discussion, but in general I despise codebases with lots and lots of new objects. I think the 7 attribute rule is *excellent*, I'd love to see us stick to that. But having a three-level deep hierarchy of newly defined objects also alarms me, and in some sense is pushing the global state underground, rather than getting rid of it. What *does* get rid of global state is implementing as much as possible as *functions* with simple inputs (simple = standard, well-known objects) and simple outputs.

> I think ndarray has around 50 methods. That's not a problem, those are just all sorts of useful things you can do with the same bits of data.

I think you'll find that the NumPy devs disagree and think that they added too many methods that should have been functions. So I actually *would* like to see us aim to reduce the number of methods and implement as many of those as external functions that operate on attributes and return useful values with which to update those or other attributes.

The main reason I wanted to chime in here though, which shouldn't actually need to hold up this PR I don't think, is this paragraph:

> In my mind the main thing for #1320 (async non-multiscale images) is something in Image which holds the slice image data but it can be in 3 states: Empty, Waiting, and Done. So it's Empty to start, default blank image, then you decide the data is needed you request it from the ChunkLoader and you are Waiting, then when the data is finally there it's Done and you can draw it.

Sure, but (a) this something should eventually be something in *base*, not image, since all the other layers also get sliced, and (b) you're actually describing a Python concurrent.Future: https://docs.python.org/3/library/concurrent.futures.html (well, optionally a None, to represent Empty). So, regardless of whether we use an ImageSlice object in the end, I think we should probably use this standard API to represent async data, rather than build our own from scratch. But the caveat is that I don't know whether it's easy to make this gel with Qt. But I just wanted to make sure that you're aware of that API, which dask distributed also uses btw.

tlambert03:
Hearty +1 on using futures

napari/layers/image/_image_slice.py
pwinston: I agree in a way it would be nice to see the underlying type is `Callable[[ArrayLike], ArrayLike]` here. But I think that raises two questions:

1. Do we really want the type hint and doc string to purposely conflict? To me that looks like a mistake, and I see that mistake everywhere. Having the same type duplicated in two places where one is a comment is super error prone, against DRY/SPOT. They will inevitably get out of sync. But we are stuck with that format until they update things to suck out the type hints for docs.

2. If we keep them in sync and put `Callable[[ArrayLike], ArrayLike]` in both places are we just saying user defined types are not allowed? If not then when are they allowed? We have to spell that out I think, otherwise it's personal preference and thus hard to review.

In my experience using types without user defined types is the worst of all worlds and is a rookie mistake for people not used to typed languages. That said, I agree it can be overdone. Maybe some existing Python project's style guide discusses this?

Here's an example of where not using user defined types gets you, and a [basic discussion](https://realpython.com/python-type-checking/#type-aliases) of "type aliases" as Python calls user defined types:

beazley-types

tlambert03: > Do we really want the type hint and doc string to purposely conflict?

I've never seen the numpy docstrings as very useful for actually typing things.  There is no standard, and they're nearly impossible to parse.  So it's just there as a human readable string (and any type information is also just a human readable suggestion).  There _are_ ways to pull out [type hints for sphinx docs](https://pypi.org/project/sphinx-autodoc-typehints/)... so technically we don't need the type hints in the numpy docstrings at all.  But they are helpful as a suggestion if someone is looking at docstrings in an interactive session... I just don't see them as binding in any way, or something that can really be "synced" successfully with the (actual) type annotation.

I'm not sure what you mean by the second point.  If there's a standard type that we use in lots of places (like `ReaderFunction`, or `LayerData`), then we should definitely use type aliases like that.  But fwiw, i disagree with that twitter post.  I _like_ having those annotations like that.  It's immediately self-documenting, and I don't have to go searching around for what it means.  (And also, if you _really_ hate that... you can put the type hints into a separate `.pyi` file)

pwinston:
> this something should eventually be something in base, not image

It will eventually be in base, yes. Right now it's in `Image` because #1320, #845 and #1300 all only deal with `Image`, and I'm focusing on those issues.

But yeah that's why it's called `ChunkLoader` and not `ImageLoader`, a "chunk" is a deliberately vague term that eventually will include geometry, etc. A chunk is meant to be "the data backing an independently loadable and renderable subset of what you are trying to draw".

> you're actually describing a Python concurrent.Future

The WIP async PR has used concurrent futures from the start: #1354

However right now it keeps the futures internal to the `ChunkLoader` and does not pass futures back to the image layer. I think this is the right design but it would take some discussion to explain. Once I open a PR on that change we can get into it. That new PR is at least 10x maybe 100x bigger than this one in long term impact, so definitely something we'll need to scrutinize.

napari/layers/image/_image_slice.py
pwinston: Updated to doctest format.

napari/layers/image/_image_view.py
pwinston: Updated to doctest format.

napari/layers/image/_image_slice.py
sofroniewn: I can see arguments on both sides, and like the idea of finding a well respected style guide we can conform too (if such a thing exists!).

Given that this is something that we can very easily change and we'll be revisiting this part of the code shortly I'd like to merge now so as not hold up development (I'll be working on the world coordinates later today and so would like this in). This definitely doesn't mean though that I want to stop this convo or consider it settled.

sofroniewn:
As noted above I am going to merge now, all doc string comments were addressed apart from the use of custom types, but I think that warrants its own larger conversation and is something we can come back to

napari/layers/image/_image_slice.py
pwinston: I've heard this complaint many times about types, and about "too many" functions and methods. People feel like stuff is being "hidden from them" and they just want to see it all at once. Totally a guess but I feel like maybe 20% of programmers have this complaint. I could give you a list of names from over the years.

To me the twitter example above is beyond horrific. I wouldn't include code like that in any project I was working on unless my specific intent was to hinder the development process and sink that project. So that gives? How can reasonable people have 180 degree out of phase opinions on stuff like this? Over the years I've concluded two things:

1. People just think differently. Some people like to see all the parts spread all over the table and other people don't. There's no way or reason to try to change either person, they are just wired that way. FWIW I feel like the "smarter" people tend to like to see all the details at once, because they can make sense of it. And crucially they simply can't imagine how anyone would not like to see it that way.

2. It's possible to mentally or physically "undo" user defined types, or too many small functions or methods, with tools or by hand, and create the "all details at once view". But it's totally impossible to do the reverse. A good type system, or functional decomposition, reflects a **huge** amount of cognitive work, carefully breaking complex things down and coming up with meaningful names. So between the two approaches I think it's clear the one with more information has to win out, if any reasonable fraction of people like it that way.

Still this is one of those things where people will never agree, but there needs to be project style where people say "I personally would not write it like that, but it fits our style". I'm not asserting what the right style is for napari, but it needs to develop one or reviews will always go in circles as people push for their personal style.

Finally I can't remember where I first read this, but this is a really important anecdote. People like to say lines of code doesn't matter, but it matters a lot by order of magnitude. Consider a 200 line program vs. 2000, 20,000, 200,000, 2M, 20M, 200M. I've worked on 200 to 2M but not bigger.

What this person was saying, which I totally agree with, is if you are used to one size and you look at the next larger size you don't think "wow that's great code" you think "What the heck are they doing? That looks really clunky. Why are they making it so complicated? That's crazy I'd *never* write it like that. I could do it *much* more simply."

But if you expand your 2,000 line program into a 20,000 line program without adopting the styles of the next size up, everything grounds to halt and your codebase becomes an usable mess. Suddenly you are like "oh.... *now* I see why they did it that way". And the same with each "level jump" to the next larger size.

You see this "size discrepancy" all the time online because people argue about code without revealing what type of project they are working on. It's like you have 3 builders heatedly talking over lunch about building techniques, but one is building a dog house, one is building an big box store, and one is building a nuclear power plant. So they are all right but also all wrong.

pwinston: I'll attempt a TLDR:
* Programmers sometimes have *very* different ideas about what style of code is easier or harder to understand or work with. Sometimes the exact opposite instincts.
* Projects need a good clear style so reviews can focus on that and not personal preferences.
* Some styles that work well in smaller codebases fall apart in larger ones.
* Corollary: some styles that work well in large codebases can be clunky in smaller codebases.

Also I was thinking that for me the case for using a type aliase or user defined type gets stronger the more times that a type appears in the code. Maybe there is a crossover-point where the benefits of "factoring out the duplication" outweigh the burden of "one more level of indirection".

tlambert03: > Also I was thinking that for me the case for using a type aliase or user defined type gets stronger the more times that a type appears in the code.

that's what I meant by pointing out `LayerData` and `ReaderFunction`, where we do just that.

pwinston:
> But having a three-level deep hierarchy of newly defined objects also alarms me

My calculation that "3 levels of hierarchy gets you 400 attributes" was not at all meant to advocate we create object hierarchies with 400 attributes. I'm just saying when you raise 7 to a power, it gets big fast. I don't think even `Image`'s 70 attributes should turned into one big object hierarchy. I think `Image` should be split like in #1325 then those smaller objects decomposed. So I'd say:

1. Mega objects: bad
1. Mega object hierarchy: better, but still bad
2. Reasonable size objects, good.

> What does get rid of global state is implementing as much as possible as functions with simple inputs (simple = standard, well-known objects) and simple outputs.

This works well in some languages, but isn't it really hard with Python because there is no `const`? If you have a function `def calc_new_thing(old_thing):` what prevents the function from modifying `old_thing` with public methods or public attributes? What enforces immutability? A docstring?

That said, I'm all for simple/pure functions. They are great. I see no downside to adding good simple functions to a codebase. If done well their "conceptual weight" is zero, if they do their job and work they are all upside. As long as their implementations are reasonable: a nice pure function which inside is a huge messy imperative program is not good. 

Although I do think people that have done a lot of math/numerical code don't realize what a huge special case that type of code is. I think creating simple functions for math related code is like 100X easier doing it in some other places. A lot of math simply is a bunch of functions in the first place, so it's a natural fit.

However I definitely think a good OO codebase is a mix of simple functions and objects. OO does not at all mean "everything has to be an object" from my point of view.

napari/layers/image/_image_slice.py
```diff
@@ -0,0 +1,54 @@
...
+        ImageView uses this to convert from raw to viewable.
+
+    Attributes
+    ----------
+    image : ImageView
+        The main image for this slice.
+
+    thumbnail : ImageView
+        The smaller thumbnail image for this slice.
```
sofroniewn, line 26: @pwinston just in case you were wondering about what this "thumbnail" actually is, it's not actually the thumbnail itself, but the sliced source data for the thumbnail, which for a normal image is just the image itself, but for a multiscale is the `_thumbnail_level` level of the pyramid.

Apologies that it is a bit weird, but hope that helps in case it was unclear. Probably when all is said and done we can figure out a better way to do this

pwinston: Yeah I kind of suspected it was not really the thumbnail. One of the old variables was `thumbnail_source` and I probably shouldn't have shortened it. We should clear it up either just in docs or change the var names back. I'll look at in the next PR.

napari/layers/image/_image_slice.py
```diff
@@ -25,18 +25,21 @@ class ImageSlice:
...
+    Examples
+    --------
+    Create with some default image:
+
+    >> image_slice = ImageSlice(default_image)
+
+    Set raw image or thumbnail, viewable is computed.:
+
+    >> image_slice.image = raw_image
+    >> image_slice.thumbnail = raw_thumbnail
```
jni, line 37: fwiw these should be triple `>>>` as the standard Python prompt.

Additionally, ideally all of our doctests will be executable and thus tested, which helps prevent them falling out of sync. When run, the are executed in the namespace of the module in which they live. So you'd want to add `  # doctest: +SKIP` at the end of lines that are meant to be explanatory rather than executed, which I think applies to the lines below?

jni:
> This works well in some languages, but isn't it really hard with Python because there is no const? If you have a function def calc_new_thing(old_thing): what prevents the function from modifying old_thing with public methods or public attributes? What enforces immutability? A docstring?

Convention, which is the way most things work in Python. In scikit-image, we work with the implicit assumption that input variables are only modified if `in_place=True` is passed, or `out=arr` and `arr` is also the input array. I think this works totally fine.

> However right now it keeps the futures internal to the ChunkLoader and does not pass futures back to the image layer. I think this is the right design but it would take some discussion to explain. Once I open a PR on that change we can get into it. That new PR is at least 10x maybe 100x bigger than this one in long term impact, so definitely something we'll need to scrutinize.

Sure, we can postpone that discussion, but expect some gentle resistance from me. =P My preferred design is for what we now call `data_slice` to be not an np.array but a concurrent.Future whose result is a np.array. We can then have a ChunkLoader or whatever externally manage the computation of those futures.

But, as you say, this is probably something best left to discussion on the PR itself.

napari/layers/image/_image_slice.py
jni: > Also I was thinking that for me the case for using a type alias or user defined type gets stronger the more times that a type appears in the code.

That's exactly right, and I see a short future for ImageConverter, especially in a world in which NumPy is starting to develop its own annotation stubs, so an array -> array function will probably get its own name from there.

But, this is a bit of a side discussion on the main one, which was the NumPy style docstring type annotations. To answer your earlier questions, @pwiston,

- I don't think it's a huge problem if they diverge from the actual type annotation. One annotation is meant for humans, another for machines. Attempts to join the two modes have usually failed.
- the text description of the argument in the docstring certainly has *no* restrictions and that should clarify what we are talking about here.
- I actually completely disagree with Talley that these annotations are not useful — this is literally the first place I check when exploring a new API and if the object types are simple, the description is always sufficient for me to infer what I need to do. Contrast this with the standard library, which doesn't have this convention and I am often left wondering what exactly needs to go in that argument.

And of course, going back to the ImageConverter type alias — for this specific case, it is so simple that indeed I think `Callable[[np.ndarray], np.ndarray]` is preferable. It's not even that many more characters. And in fact, I think that type is too precise, since a callable mapping a dask array to a dask array would work here also. So it could just be `Callable`, or `Callable[[ArrayLike], ArrayLike]`.

> Maybe some existing Python project's style guide discusses this?

I think type hints are too new in Python for any style guides to have made pronouncements on this. But in Dave Beazley's example, I don't think type aliases would help. He would still hate the signature and it would be even less clear, unless, as discussed above, the type aliases are extremely common or totally self-explanatory. `ImageConverter` is neither of those, nor do I foresee it becoming so.

tlambert03: > I actually completely disagree with Talley that these annotations are not useful

? I feel like we're saying the exact same thing!  so I think you misunderstood my point.  I didn't say they are not useful "full stop"... I said they are not useful for actually typing things. (in a way that can be programmatically parsed), and I said they are there as human readable... exactly where a human might go first when exploring a new API, or in an interactive session.

jni: Oops! Sorry for misunderstanding your point! Agreement is good! 😅

napari/layers/image/_image_slice.py
pwinston: I think I'll remove these `Examples` sections from the new PR for a few reasons:

1. I invented the idea of an **Example** section without knowing there was a standard that included doctest.

2. There's nothing to show here, there are no return values. It's pretty pointless.

3. No one is going to use these classes except where we do in the real system. No reason to pretend people will. Better to just explain how they work in regular docs.

I think an `Examples` section makes sense for some classes but not these.

napari/layers/image/_image_slice.py
pwinston: This PR is merged, I see you opened #1355 which is good. If the answer to when to use type aliases is "sometimes" we should flesh out in more detail what the guidelines are, when to use which style. Long term IDE's should let you toggle showing expanded types. You can see *some* expanded types on hover in vscode today. That seems to be heading towards the best of both worlds.

ArrayLike

LayerData

sofroniewn: Those screenshots are pretty nice @pwinston. Maybe we continue this convo on #1355 and then if we do make changes have them be as part of a dedicated PR. I'd rather avoid mixing it in with #1354 where we'll have lots to discuss too.

jni: My personal position on IDEs is that they are nice for people that have them, but policies should not depend on people using an IDE. Code should just be readable. Probably 90% of my interaction with code these days is reading it on GitHub PRs, so until GitHub itself becomes an IDE, plaintext interpretation is the lowest common denominator that we should target.

PR #1344, opened 2020-06-05
Fix ranger slide position

### LINKED ISSUE #1337: Range slider does not render in proper place in multi monitor environment on linux
## 🐛 Bug

20200604_165422_HDR

On my system (Ubuntu 19.10 with gnome shell and nvidia drivers) when napari is not on the most left screen then range slider is not rendered in proper place. 

## To Reproduce

Steps to reproduce the behaviour:

1. Start napari on mulit screeen computer
2. Move window from left screen
3. Click right on range small slider to open big

## Expected behavior
Slider should be placed properly (inside napari window)

## Environment
napari: 0.3.5.dev3+ge9df14f6
Platform: Linux-5.3.0-55-generic-x86_64-with-debian-buster-sid
Python: 3.7.2 (default, Aug 1 2019, 23:29:28) [GCC 8.3.0]
Qt: 5.14.2
PySide2: 5.14.2.2
NumPy: 1.18.2
SciPy: 1.4.1
Dask: 2.12.0
VisPy: 0.6.4

GL version: 4.6.0 NVIDIA 390.132
MAX_TEXTURE_SIZE: 32768

Plugins:
- napari-plugin-engine: 0.1.5
- napari-plugins: 0.1.0
- svg: 0.1.3

Ubuntu 19.10, Gnome Shell with nvidia drivers on 3 screen configuration (FullHD, WQHD, FullHD)  

## Additional context

Opening this view for range slider block print screen.

Czaki:
# Description

This is bugfix for range slider outside main screen. 

Additional i add else clause to silence "variable may be referenced before assignment" and got better error message if wrong type argument is passed. 

Code for Qt below 5.10 was removed. 

## Type of change

# References

Fixes  #1337
# How has this been tested?

- [x] looks good outside main screen

## Final checklist:

### DIFF
--- a/napari/_qt/qt_modal.py
+++ b/napari/_qt/qt_modal.py
@@ -119,6 +119,8 @@ class QtPopup(QDialog):
         elif isinstance(position, (tuple, list)):
             assert len(position) == 4, '`position` argument must have length 4'
             left, top, width, height = position
+        else:
+            raise ValueError(f"Wrong type of position {position}")
 
         # necessary for transparent round corners
         self.resize(self.sizeHint())
--- a/napari/_qt/qt_modal.py
+++ b/napari/_qt/qt_modal.py
@@ -127,9 +127,20 @@ class QtPopup(QDialog):
         # make sure the popup is completely on the screen
         # In Qt ≥5.10 we can use screenAt to know which monitor the mouse is on
 
-        screen_geometry: QRect = QGuiApplication.screenAt(
-            QCursor.pos()
-        ).geometry()
+        if hasattr(QGuiApplication, "screenAt"):
+            screen_geometry: QRect = QGuiApplication.screenAt(
+                QCursor.pos()
+            ).geometry()
+        else:
+            from qtpy.QtWidgets import QDesktopWidget
+
+            window = self.parent().window() if self.parent() else None
+            screen_num = (
+                QDesktopWidget.screenNumber(window)
+                if window is not None
+                else 0
+            )
+            screen_geometry = QGuiApplication.screens()[screen_num].geometry()
 
         left = max(
             min(screen_geometry.right() - width, left), screen_geometry.left()
--- a/napari/_qt/qt_modal.py
+++ b/napari/_qt/qt_modal.py
@@ -135,12 +135,7 @@ class QtPopup(QDialog):
             # This widget is deprecated since Qt 5.11
             from qtpy.QtWidgets import QDesktopWidget
 
-            window = self.parent().window() if self.parent() else None
-            screen_num = (
-                QDesktopWidget.screenNumber(window)
-                if window is not None
-                else 0
-            )
+            screen_num = QDesktopWidget().screenNumber(QCursor.pos())
             screen_geometry = QGuiApplication.screens()[screen_num].geometry()
 
         left = max(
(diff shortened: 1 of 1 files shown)


tlambert03:
Thanks @Czaki!  Will take a look soon.  Can you add back the compatibility with <5.10 though?  while it’s not officially supported, I’m pretty confident this would be the only line preventing someone from using <5.10, and I see no need to exclude it just yet.  (I recently tested as far back as 5.6 and still pretty much works)

Czaki:
Base on documentation I wrote fallback code. It depends on deprecated class so I use local import.

tlambert03 (approved):
This is great @Czaki, thanks!  

I had to make a small change to get it working on Qt < 5.10:  `QDesktopWidget.screenNumber` didn't work as a static method (without instantiating a `QDesktopWidget`), so I instantiated it, and passed the `QCursor.pos()` for simplicity (tested locally on `Qt5.9.7`).  Have a look and let me know if everything looks ok to you.

Czaki:
In the documentation [screenNumber](https://doc.qt.io/qt-5/qdesktopwidget.html#screenNumber) accept only widget. Signature with QPoint is  defined in both PySide2 and PyQt documentation. But I'm not sure from which release. 

If you think that it is not a problem then I'm happy with this change.

tlambert03:
> In the documentation [screenNumber](https://doc.qt.io/qt-5/qdesktopwidget.html#screenNumber) accept only widget. Signature with QPoint is defined in both PySide2 and PyQt documentation. But I'm not sure from which release.
> 
> If you think that it is not a problem then I'm happy with this change.

yeah, I noticed that too...  it's true, this might be a little risky.  let's see if we can figure out which releases support it

Czaki:
Maybe this may be used for workaround https://doc.qt.io/qt-5/qapplication.html#widgetAt

tlambert03:
we're ultimately after the screen number right?  so we'd still need to feed that widget to screenNumber?  if so, might we just go back to the previous method? (but make sure to isntantiate the DesktopWidget first?)

it's frustrating that I can't seem to find signatures for older pyside2/pyqt5 versions

tlambert03:
went to `QDesktopWidget().screenNumber(self)` instead.  Since `self` is going to be the popup widget itself, and that's the screen we want, seems to be clean and fits the more reported signature.  I tested with `qt 5.9.7` from conda... if you want to test that with your dual monitor, that'd be great.  Otherwise I'm fine with it as is.  thanks again!

Czaki:
I found it https://doc.qt.io/qt-5/qdesktopwidget-obsolete.html
It is obsolete member of obsolete class. 

And last solution does not work. It placed everything on last screen.

tlambert03:
> And last solution does not work. It placed everything on last screen.

ack!  ok... well, since you have the dual monitor, can you create a 5.9 environment and test it?  I went down this path because I got an error when I tested your original PR in 5.9 ... so we need to make sure that works before merge

Czaki:
I should write this. I create test environment using pyenv and install pyqt5==5.9 from pip. 
When I test your second commit it fail because slider was placed on last screen. When I revert to your previous commit everything works.

tlambert03:
yeah, your call.  Play around with it, let me know when you're happy, then I'll merge

Czaki:
I test it and I'm happy.

tlambert03:
ok thanks!

Issue #1143, opened 2020-04-14
Temporary pan/zoom on Labels layer lasts <1-2 seconds


hectormz:
## 🐛 Bug

When in brush/fill/pick mode on a Labels layer, holding space bar temporarily enables pan/zoom mode, but reverts to the original mode after a second or two

## To Reproduce

1. Create Labels layer
2. Select pick/fill/brush mode
3. Hold space bar
    - Mode switches to pan/zoom for < 1-2 sec
    - Mode switches back to original mode despite space bar still held
    - No pan/zooming may have happened before reverting back to original mode

## Environment

```
napari: 0.2.12+51.g6ea95a9
Platform: Windows-10-10.0.19041-SP0
Python: 3.8.2 (tags/v3.8.2:7b3ab59, Feb 25 2020, 23:03:10) [MSC v.1916 64 bit (AMD64)]
Qt: 5.14.1
PySide2: 5.14.1
NumPy: 1.18.1
SciPy: 1.4.1
Dask: 2.12.0
VisPy: 0.6.4

GL version: 4.6.0 NVIDIA 441.45
MAX_TEXTURE_SIZE: 32768

Plugins:
- mat-images: 0.1.0
```

jni:
This happens to me on Linux also. I guess holding space only works on Mac. Possibly something to do with how key-hold events are sent or detected by Qt on different platforms... See #133 :joy:

hectormz:
@jni Ah, I didn't see that before. From doing some basic testing, it looks like the holding the space bar is only registering as a single event, while holding any of the arrow keys results in repeated events. I'm not sure if @kne42 has any insight into this, although it seemingly works fine on OS X?

sofroniewn:
This has been closed by #1350

Issue #133, opened 2019-03-01
Holding space to pan and zoom doesn't work correctly on Linux


jni:
## 🐛 Bug

In Ubuntu 18.04, "hold space to pan and zoom" while in annotation mode for a markers layer (or in object or vertex selection mode for shapes ­— this is where I first noticed it while reviewing #100) only works for the first drag. After mouse release, despite continuing to hold space, it switches back to annotation mode. I have to release space and press space again for it to work.

Zoom is even worse, as it appears to "deactivate" space e.g. when changing the scroll speed. So it only works for a fraction of a second before having to press space again.

## To Reproduce

1. Run `python examples/add_markers.py`
2. Select the markers layer and turn on annotation (A) mode
3. Hold down spacebar, click and drag, then attempt to click and drag again without releasing space. (the second click will insert a new marker)

## Expected behavior

The second click and drag should also pan. This matches the behaviour in macOS.

## Environment

 - napari-gui Version (e.g., 1.0): master / #100 
 - OS (e.g., Linux): Ubuntu 18.04
 - Python version: 3.6

sofroniewn:
Hmm ok this is strange (I only have a macOS right now where everything still works fine) but we can get a Linux machine at CZI for testing and figure out what events are getting triggered / when and how it relates to our detection and processing of keypress / click vs PyQt native vs vispy.

sofroniewn:
@jni do we know if this is still an issue? can you confirm it is still a problem with master and then we should look into fixing as this seems like important functionality to have working

jni:
Indeed it still is an issue for me, and very frustrating when using napari's annotation tools...

sofroniewn:
@jni do i remember you saying this has been resolved or is it still an issue?

jni:
No, still an issue, what's resolved is the tearing when moving quickly...

constantinpape:
I also just encountered this on 18.04.

sofroniewn:
One place to look in the code is where the `repeat` press is being handled https://github.com/napari/napari/blob/d52ddf87d90328a8a2c1757d43e3075bb672732f/napari/_qt/qt_viewer.py#L555

sofroniewn:
This has been closed by #1350

Issue #1282, opened 2020-05-25
Switching to pan mode via space works poorly


constantinpape:
## 🐛 Bug

Switching to the pan mode via holding `space` in the points layer works only poorly.
It switches to the pan mode when I hold space, but as soon as I do any action, e.g. zooming, it goes back to the previously selected mode.

**Edit:** The problem is not performing an action, but that the mode flips back to the selection mode  
after a second (while I still hold done space). I tried with two different keyboards, so I don't think it's a keyboard problem.

## To Reproduce

- open an image in napari
- add a points layer and some points
- go to the selection mode
- hold space and try to navigate the image

## Expected behavior

Holding space activates the pan mode and it's only deactivated when releasing the key.

## Environment

 - Please copy and paste the information at napari info option in help menubar here:
```
napari: 0.3.1
Platform: Linux-5.3.0-51-generic-x86_64-with-glibc2.10
Python: 3.8.2 | packaged by conda-forge | (default, Apr 24 2020, 08:20:52) [GCC 7.3.0]
Qt: 5.12.5
PyQt5: 5.12.3
NumPy: 1.18.4
SciPy: 1.4.1
Dask: 2.16.0
VisPy: 0.6.4

GL version: 3.0 Mesa 19.2.8
MAX_TEXTURE_SIZE: 16384

Plugins:
- covid-if-annotations: 0.1.dev2+gfeab50b.d20200522
- napari-plugin-engine: 0.1.5
- svg: 0.1.2
```

 - Any other relevant information:
- Napari was installed from conda-forge
- used inside of this tool: https://github.com/hci-unihd/covid-if-annotations

sofroniewn:
For better or for worse this is a duplicate of #133 I think. Unfortunately I don't have access to a linux machine and this does seem to be platform specific. I'm not sure where the bug really comes from see code for space press here https://github.com/napari/napari/blob/master/napari/layers/points/_points_key_bindings.py#L5

Any help debugging from @kne42 @jni @tlambert03 or others would be appreciated

constantinpape:
> For better or for worse this is a duplicate of #133 I think.

True, it's a duplicate.
I am going to close this.

PR #1350, opened 2020-06-10
#133 press down is treated as multiple press and release in linux

liu-ziyang:
# Description
Fix linux treating key held as multiple press and release
I need some advice for adding test for this, I am not sure how to simulate this yet

## Type of change

# References
https://github.com/napari/napari/issues/133

# How has this been tested?

- manually tested on linux

## Final checklist:
- [] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/napari/_qt/qt_viewer.py
+++ b/napari/_qt/qt_viewer.py
@@ -585,7 +585,7 @@ class QtViewer(QSplitter):
         event : qtpy.QtCore.QEvent
             Event from the Qt context.
         """
-        if event.key is None:
+        if event.key is None or event.native.isAutoRepeat():
             return
         combo = components_to_key_combo(event.key.name, event.modifiers)
         self.viewer.release_key(combo)
--- a/napari/_qt/qt_viewer.py
+++ b/napari/_qt/qt_viewer.py
@@ -585,7 +585,11 @@ class QtViewer(QSplitter):
         event : qtpy.QtCore.QEvent
             Event from the Qt context.
         """
-        if event.key is None or event.native.isAutoRepeat():
+        if event.key is None or (
+            # on linux press down is treated as multiple press and release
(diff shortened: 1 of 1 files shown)


sofroniewn (commented):
Given this line in https://github.com/napari/napari/blob/765349053f33f1537e345982cf804bc2a40199f6/napari/_qt/qt_viewer.py#L570 have you checked that things still work fine if someone presses and holds the `up` / `down` / `left` / `right` keys? i.e. things pressing and holding continues re-triggering events and say the slider keeps moving to the left?

sofroniewn (approved):
I'm approving this as everything still works great on Mac, but maybe @jni can test on linux, or @hectormz as this should also close #1143 or @constantinpape who reported this in #1282 

Would really be amazing if this fixed all those issues! Thanks for chasing this bug down @ziyangczi!!

hectormz:
@sofroniewn Tested on Windows, works great and closes #1143 for me!

hectormz:
Great work @ziyangczi

liu-ziyang:
> Given this line in
> 
> https://github.com/napari/napari/blob/765349053f33f1537e345982cf804bc2a40199f6/napari/_qt/qt_viewer.py#L570
> 
> have you checked that things still work fine if someone presses and holds the `up` / `down` / `left` / `right` keys? i.e. things pressing and holding continues re-triggering events and say the slider keeps moving to the left?

the change should only change key release behavior, when key is held or pressed there should be no effect

sofroniewn:
> the change should only change key release behavior, when key is held or pressed there should be no effect

Perfect, makes sense. Looks like this is the fix!!

jni:
NO WAY! What a holiday gift!!! 🎉 🎉 🎉  🙏 🙏 🙏 @ziyangczi!

Issue #1353, opened 2020-06-13
Layer models


sofroniewn:
## Discussion

There have been numerous discussions lately of our layer models, and various refactors at play in #1325, #1040, #1343 and more. There seems to be universal agreement that having the dims model which contains slicing information and having the sliced data on each layer is a problem and will prevent things like orthoviews and multicanvas features more broadly. With that in mind I'd like to make the following proposal for after the refactor is complete.

**After refactors layer model should have no knowledge of:**
- Where the sliders are
- Where the camera is
- Where the mouse cursor is
- What part of the data has been sliced
- What part of the data is being viewed
- What part of the data has been clicked on or hovered over
- The thumbnail

**It might also be nice if the layer model also had no knowledge of:**
- How the user currently wants to interact with the data (i.e. which “mode”)
- Any data that is currently selected or highlighted (i.e. in shapes/ points)
- If the layer is “selected” / “visible” / “editable”

Note “visible” is currently exposed in main add_layer API and can be used to make loading lazy, so we might want to keep that or at least deprecate it slowly.

**The layer model should know:**
- The full nD data
- How the data is placed in a physical/ world coordinate system
- Properties of the data (like class labels for points or shapes)
- How the data would look after being rendered (colormaps, colors, opacity etc)

The `add_layer` API and `layer_state` serialization method should allow users access to all these properties and be everything a users needs to know what data is, where in space it is, and what it would look like if rendered.

**Methods of the layer model should allow:**
- Slicing
- Interactivity
- Viewing

And should take slice information, cursor information, and camera information as needed to produce the desired outputs. The exact interactivity requested will depending on what "mode" the rest of the app is in.

Feedback welcome!

sofroniewn:
Here is a schematic of some of our current architecture in terms of properties and inheritance. Lots of these things are also coupled through out event system and not everything is here, but having this might aid some conversation 
![image](https://user-images.githubusercontent.com/6531703/84580038-2ff58b00-ad88-11ea-867c-e79ed0f58395.png)

sofroniewn:
Focusing just on the `ViewerModel` side of things and thinking only about all the `Model` code we need (including for a world where we support multiple canvases), one reorganization could be as follows:

![image](https://user-images.githubusercontent.com/6531703/84583637-64307200-adaf-11ea-856f-29583f85d97a.png)

[...]

PR #1354, opened 2020-06-13
[WIP] New ChunkLoader and ChunkCache

### LINKED ISSUE #1320: Blocked UI while loading images in notebook

pwinston:
# Description
Fix #1320 by adding asynchronous loading of single-scale `Image` layers.

# ChunkLoader

The new `ChunkLoader` can load asynchronously or synchronously. By "load asynchronously" we mean `asarray` is called in a worker thread. We call `asarray` in a worker thread so if it triggers IO or computation it does not block the GUI thread. When the worker thread finishes loading the chunk, it passes the chunk to the layer that requested it in the GUI thread. We pass via the GUI thread because we don't want worker threads directly calling into model/layer code.

#  Python's GIL

Python has a [Global Interpreter Lock](https://medium.com/python-features/pythons-gil-a-hurdle-to-multithreaded-program-d04ad9c1a63) that prevents two threads from running Python bytecode at the same time. This is not as bad as it sounds because there are two common cases where multiple threads *can* execute at the same time:

1. One is doing disk or network IO or blocking on almost any system call.
2. One is doing computation that releases the GIL while doing computation implemented in C/C++. Many CPU intensive `numpy` / `scipy`  packages are implemented this way.

If we do need total independence between threads we can switch to processes, we are using `concurrent.futures` and it supports both threads and processes. 

# Code Changes
 
| what | type | description |
| --- | --- | --- |
| `ChunkLoader` | new class | Calls `asarray` in worker threads, has a thread pool. Owns the `ChunkCache`. |
| `ChunkRequest` | new class | Contains the array we want to load and information about it. |
| `ChunkCache` | new class | Sizes itself as a fraction of RAM. Stores loaded chunks so re-loading is instant. |
| `QtChunkReceiver` | new class | Shuttles newly loaded chunks from worker threads to their layer in the GUI thread. |
| `synchronous_loading()` | new context object | Turn async loading on or off. |
| `ImageProperties` | new tuple | Information about the `Image` that's passed into `ImageSlice` |
| `ImageSlice` | modified class | Added new `load_chunk()` and `chunk_loaded()` methods.|
| `Image._load_single_scale()` | new method | Move the single-scale load to own method. |
| `Image._get_slice_shape()` | new method | Is this right? |

# Test Changes

## conftest.py changes

| what | type | description |
| --- | --- | --- |
| `configure_loading` | new fixture |  Toggle async vs. async loading. |
| `@pytest.mark.sync_only` | new mark | Mark tests that cannot be run async. |
| `skip_sync_only` | new fixture | Skip sync_only tests when appropriate. |
| `--async_only`| new pytest flag | Run only the async tests. |

## Two test passes

Ideally we'd use a parameterized fixture to run test as sync and async in one invocation of `pytest`. This is not possible today because we get a segfault as shown in #1390. The PR shows the segfault happens without any async code, it's just from running the pre-existing tests twice.

## Current Approach

To run sync tests: `pytest`
To run async tests: `pytest napari --async_only`
 In `.cirrus.yml` we run both of these separately.

## sync_only tests

14 tests have been marked `@pytest.mark.sync_only`. These tests will fail if run with async in most cases because they immediately read back data that was just set. With async that data might not be there yet.

# TODO for this PR
**New Debug menu items**:
  - [ ] Enable/disable async.
  - [ ] Configure a "delay" in the worker threads for testing purposes.
  - [ ] View cache usage and set the cache size?

**Other items**:
- [ ] Fix remaining async test failures
- [ ] Address `Image` calling `_create_image_slice` in __init__, seems wrong.
- [ ] Real world testing, beyond unit tests.
- [ ] Finish logging support, flag to turn it on?

# TODO for future PRs

- Use **Dask** cache instead of ours. One cache is better than two?
- Introduce UI that displays loading status, so user knows when data is done loading.
- Introduce UI that can lock layers together, so they display in an "all or nothing" manor.
- Async for multi-scale images (#845, #1300)
- Async for non-image layers.


pwinston:
Early demo showing 1 second long async loads with 1 worker thread. Since we can only clear pending requests and not in-progress ones, the delay once you rest on a slice is 1-2 seconds because you potentially have to wait for one stale request to finish, then you have to wait for your request.

The **loading: N** text screens are created on the fly as placeholders, something to show until the real image is loaded. It's TBD what we want to show long term, but this is useful for debug/development. The images with just the number on there is the actual dataset being viewed in this case.

async-3

pwinston:
## Status
Above roughly works as seen in the GIF.  However currently it break lots of tests and I'm working on cleaning it up and making it 100% correct, in progress.

## Futures

Right now the PR uses `concurrent.futures` but the futures are kept inside `ChunkLoader`. The `ChunkLoader` indirectly calls `Image.chunk_loaded()` in the GUI thread when the chunk has loaded. This 100% meets our needs for rendering. There was no reason to have the `Layer` hold `Future` objects for the purpose rendering so far, see [ChunkLoader source](https://github.com/pwinston/napari/blob/feature/chunk-loader/napari/utils/chunk_loader.py#L160-L162).

## Futures in Layers

I know of 2 tests the PR won’t pass even if I fix all issues. Both tests are where we read data (image or thumbnail) immediately after creating a Layer or assigning new data to a Layer. With async loading the data simply isn't there yet.

In another thread @jni suggest the Layer could hold the `Future`. We could do that and then start using `dask` or something similar we create. So if you ask for an array it’s really a `dask.delayed` whose body waits on a future. This raises a lot alarm bells for me though. We’re adding two layers of complexity, first one layer to making this async, then a second layer to make them synchronous again!

Below are 3 diagrams to try and explore our options.

## Sync Model

async-1

This **Sync Model** is what's in master today. The big question mark what's behind an `ArrayLike` interface: could be regular data or IO or a computation. To render a slice we resolve the data to create a `ndarray`, this can block which is what blocks the GUI today. Once the data is an `ndarray` people can access it with the API.

## Async Model 

async-2

This **Async Model** is if we just push the `Future` object into the Layers with no other changes. Now everything is `ArrayLike` and it potentially chains all the way back. So if a user grabs a thumbnail and tries to access the bytes, it’s really access this whole chain: resolve the `dask`, load asynchronously, process the thumbnail.

This is confusing and complicated because the model contains async objects in various states, so the state of the model is changing over time. Also renderer and API are sharing the same data structures. The renderer will often have to cancel loads as the user navigates, that will interact messily with the API.

## Async Slicer

async-3

This **Async Slicer** is an alternative idea with these properties:

1. A Slice is an object derived from the model, it's not part of the model.
2. The renderer creates slices, but these slices are not accessible to the API. In general the renderer's state will be private. The renderer is one client of the model, the API is another one, and they shouldn't be mixed. The renderer will constantly be cancelling async loads and these shouldn't affect the API.
3. Asking the API for a slice will block. The `ChunkCache`, not shown, is shared so asking for slice from the API that's already being rendered will be fast.
4. Asking the API for a slice will return a fully resolved slice with actual `ndarray` data.

## Multi-scale or otherwise chunked slices

Of the above 4 might eventually have to change. For non-multi-scale it should be fine because slices are small. For multi-scale we might want to allow the user to create giant slices but only access portions of the data, so the whole slice needs to lazy load. At that point we are somewhat heading back to the complicated **Async Model** diagram but with 2 important changes:
1) All the async objects are in the slice only, not in the model
2) The render has its own slice(s) not connected to the API slices

sofroniewn:
> This Async Slicer is an alternative idea with these properties:
>
> 1. A Slice is an object derived from the model, it's not part of the model.
> 2. The renderer creates slices, but these slices are not accessible to the API. In general the renderer's state will be private. The renderer is one client of the model, the API is another one, and they shouldn't be mixed. The renderer will constantly be cancelling async loads and these shouldn't affect the API.
> 3. Asking the API for a slice will block. The ChunkCache, not shown, is shared so asking for slice from the API that's already being rendered will be fast.
> 4. Asking the API for a slice will return a fully resolved slice with actual ndarray data.

I really like this approach and think it dovetails nicely with goals I laid out in #1353 in particular:

> After refactors layer model should have no knowledge of:
>
> - Where the sliders are
> - Where the camera is
> - Where the mouse cursor is
> - What part of the data has been sliced
> - What part of the data is being viewed
> - What part of the data has been clicked on or hovered over
> - The thumbnail

Looking forward to discussing more in the meeting today!! It definitely relates to the event handler / MVC refactors too

pwinston:
Napari meeting 2020-06-18:
1. Let's just disable async on those 2 problem tests.
2. Long term those tests will not exist, when model/view are decoupled.
3. For general reasons we will have a flag to toggle sync vs. async.

To summarize the goal is:

| setting | result |
| --- | --- |
| NAPARI_ASYNC_LOAD=0 | All tests pass in synchronous mode. |
| NAPARI_ASYNC_LOAD=1 | All tests pass async, except two are forcing sync mode on. |

Even with `NAPARI_ASYNC_LOAD=0` it's not running the original code, it has changes, it's just that the actual async part is bypassed.

Also we discussed some subtle things might be broken with async, like what is shown in status bar when the data is still loading. Might be some things we need to fix that are not part of any test.

sofroniewn:
Tests are all passing @pwinston, very nice!! Is this ready for some review now, should @jni @tlambert03, and I do a pass through the code?

pwinston:
Not ready yet. These are sync tests only. I’m still fixing some tests for Async and then need to tidy things up. I will ping everyone when ready for review.

pwinston:
There was this comment on `napari-omero`: https://github.com/tlambert03/napari-omero/pull/1#issuecomment-649629470

I'm pretty much restating above here, but it's worth spelling it out more than once I think. This is basically the thought process above:
1. Pre async everything was static and that was good.
2. Naively adding async **everything** can be in all different states all the time, that's bad.
3. Let's creates slices asynchronously, but the slices are concrete (ndarrays). That's good for non-multi-scale.
4. But for multi-scale we need a blend of 2 and 3. The slice won't be fully resolved with data, it will contain futures or promises of some kind, that only get resolved as needed.

As for what we are actually doing:
A. To solve #1320 I'm really do 3 above, but it's the slice is still wrongly inside the model.
B. We can improve 3 to factor out the view from the model, but isn't strictly required.
C. Solving #845 and #1300 will require doing 4. And not insignificantly adding an octree.

pwinston:
Done:

- New `ChunkLoader`, `ChunkRequest`, `ChunkCache` classes.
- New `synchronous_loading()` context object
- New `ChunkReceiver` in `layerlist.py` uses `weakref`
- `ImageSlice` has new `ImageProperties`
- `ImageSlice` has new `load_chunk` and `chunk_loaded` methods
- `Image` now creates an `ImageSlice` every time `_set_view_slice` is called
- New `Image._get_slice_shape()`, but is it right?
- Split out `Image._load_single_scale()` into its own method

TODO:
- [ ] Use `EmitterGroup` instead of `QObject` in `ChunkLoader`
- [ ] Add `cachetools` LRUCatch to `ChunkChache`
- [ ] Two test passes: async and sync working locally
- [ ] Two test passes: on CI
- [ ] Fix why does `Image` need to call `_create_image_slize` in __init__
- [ ] Full support for multiple layers (not yet even tested)

pwinston:
I removed the "status comment" above and updated the main PR description with a nicer accounting of what's done and what's remaining on this PR.

pwinston:
Below is showing 16 image layers each with 0.5 second sleeps, using 6 `ChunkLoader` threads.  Without async switching slices hangs for around 8 seconds. With async it takes around 400ms to switch slices and around 1.5 seconds for the slice to be fully loaded, and you can interrupt it during those 1.5 seconds.

This is a big improvement, but still slow. The 400ms is 2.5Hz. It looks like the time is spent calling `refresh()` 16 times and then vispy spends it setting up the 16 textures and drawing them. See below perf trace. Each layer here is one 1024x1024 image with one digit embedded in it. Napari itself no longer creates any images with text. 

16-layers-2

pwinston:
Above I added tracing of the ChunkLoader threads. So on the trace you see those running in parallel which is nice. The initial `MouseButtonPress` event is 438ms. It's probably because we are using full size but blank images as placeholders until the real data is loaded.

It's calling `refresh()` 16 times and then a 245ms `Paint` event where vispy probably loads those 16 textures onto the card and then draws them. This is a big problem though, because 400ms means the slider only refreshes at 2.5Hz, which means making things async did not give us interactive performance. 

Prior to async we inserted a 1 pixel/voxel size image first, before overwriting it with the real data. We maybe need to switch back to that. So then we incur no cost up front. The reason we used the full size blank images was because many tests checked that the layer data was the correct shape. If we change back to 1 pixel/voxel placeholders those tests will break, we'll have to set them to `async_only`. However I think we need to do something like that to get the interactive performance we want.

napari-load-16

pwinston:
This is now using 1x1 pixels placeholders but it's still slow! The `MouseButtonPress` switching to a new slice is 410ms: 92ms in `set_point` on the napari side and 310ms in `Paint` on the vispy side:

napari-load-16-single-pixel

Here it is drawing just one 1x1 image. It takes about 18ms, where 12ms is `flush_commands` which means just waiting on the card, which is hard to speed up:

napari-one-draw

My assumption is this is slow only because we are drawing it for the first time and so we have to setup resources including moving the data to VRAM. But the data is 1 pixel! Later we draw new 1024x1024 images and it takes about 20ms, so only a bit more, so it seems like there some large baseline cost for vispy or opengl to draw a new image regardless of the size.

Some ideas where to go from here:

1. Image still has a 1 pixel placeholder, but the vispy visual knows this is empty and so it doesn't actual create anything in the scene at all. That should cut out the 300ms.
2. Go further, the `Image` layer is in a special empty state while loading so it doesn't have data at all yet. That hopefully can cut down on the 90ms part as well as the 300ms part.
3. Go even further and the entire layer gets removed in some way when you switch to a new slice, and doesn't get added back until the data is loaded.
4. Without doing any of the above, or in addition, amortize the initiation of loading layers over multiple frames, instead of doing it all at once. 

The original hope with async was that while data is loading everything is the same as sync, but the image just happened to be "blank" until the data was loaded. But that doesn't seem to be possible for performance reasons since just setting up the blank images is expensive, even if the images are only 1x1 pixel. That's surprising because we moved **all** loading out of the GUI thread, but just putting in blank fully in-memory placeholder images is too expensive. Yet here we are.

The above 4 suggestions might work, but the downside is they are more invasive, they will make napari's "loading" state quite different from its "loaded" state. As discussed in the 3 diagrams at the start of the PR, having things change over time is complicated. We need to do something here though.

sofroniewn:
I'm just catching up here @pwinston and will need to take this in during a deeper dive tomorrow, but it does seem like we're hitting some limits of our architecture here with 16 layers present and how we use vispy. 

> The reason we used the full size blank images was because many tests checked that the layer data was the correct shape. If we change back to 1 pixel/voxel placeholders those tests will break, we'll have to set them to async_only

I think that's fine, it was an assumption that made sense at the time, we can think about better tests moving forward.

> so it seems like there some large baseline cost for vispy or opengl to draw a new image regardless of the size.

Indeed!

> Image still has a 1 pixel placeholder, but the vispy visual knows this is empty and so it doesn't actual create anything in the scene at all. That should cut out the 300ms.

I think this could be a good thing to try first. I wonder if you can use the `node.visible=False` parameter here to achieve this. I think with visible False there will be no draw and no overhead, you could turn visible on the node to `True` only when you needed/ wanted. https://github.com/napari/napari/blob/master/napari/_vispy/vispy_base_layer.py#L147

> Go further, the Image layer is in a special empty state while loading so it doesn't have data at all yet. That hopefully can cut down on the 90ms part as well as the 300ms part.

That could make sense too. We might be able to trim down each of those `set_point` calls too - seems like each one is ~5ms.

Looking forward to discussing more!

pwinston:
> I think this could be a good thing to try first. I wonder if you can use the node.visible=False parameter here to achieve this

This worked and was only a few lines. I created a new event `Layer.events.loaded`. Now `VispyBaseLayer` will not set the node to visible unless both`self.layer.visible` and `self.layer.loaded` are `True`. Non-async layers will just be always `loaded` so no change for them.

This cuts out the 300ms of rendering up front so the slider is more like 10-12Hz instead of 2.5Hz. This does not feel that bad for a slider with 16 layers. More layers will slow it back down though. To fix that I think we need to amortize. I think we need to amortize initiating the loads (what we've been talking about) but also when the loads finish. Right now if N chunks finish loading around the same time, we will draw all N the next frame which will be a slow frame. Really we should draw only M new chunks per frame.

This is the "page into VRAM over time" part of the rendering doc. Although it's really more general, it's "do everything over time" in order to prevent slow frames. Although we might want to let the framerate sink some, because then we can grind through the work faster. So it's a trade-off between framerate and load speed.

But I will try to push off amortizing for now. I'd rather focus on getting this usable by people for real-world work, so that we can prioritize improvements and bugs. Amortization may or may not be high on that list.

pwinston:
# Updates

## Tests

* CI is using the two separate invocations of `pytest` since there's no solution for #1390 yet. Not a big deal, it works. Tests that cannot possibly pass with async are marked `sync_only`.

## Multi-layer testing
1. I've been running 1 thread and 1 layer since day one. The goal now is get async into people's hands for testing, so I created a more real-world test with 16 layers and 6 threads.
2. Changing slices was 2.5Hz even with 100% of the work/sleep in worker threads, very disappointing.
3. Napari `set_point` initiating the loads was slow, and then drawing the blank textures was slow, even if they are sized down to 1x1 pixel. Fixed the draw part by marking the nodes invisible until loaded.
4. This gets us to 10Hz which seemed okay. Have ideas to go much faster but waiting.

## Testing Torch in Jupyter

Two problems with #1320:
1. Sometimes single-step will step ahead two slices.
2. Initiating the loads is **really** slow, like 800ms even with worker threads.

I spent a long time debugging 1. It only happens with async, with sync it's fine. With async the events get all messed up and the `MouseButtonRelease` gets pushed way out in time. Then Qt's timer for the slider does a second step. I hope to fix this by creating a queue of loaded chunks, and then have the worker signal `refresh` rather than `chunk_loaded` and `refresh` empties the queue.

For 2 I suspect the Torch stuff is holding the GIL and we might need to do it in a separate process, but need to do more investigation once 1 is fixed.

pwinston:
Closing now that we have #1495 with only the async portions. The perfmon portions were split out and merged as #1453.

PR #1360, opened 2020-06-15
Add world extents for layers

### LINKED ISSUE #986: change default zoom level
## 🐛 Bug

Rather than describing this with words I show a screen recording:

shapes_opaque

## To Reproduce

1. add a shape layer
2. set blending mode to additive or opaque
3. choose some colors other than the background color for face and edge 
4. draw a polygon

I wonder whether during drawing inside and outside of the polygon are not clearly defined.

## Expected behavior

Simple polygon with selected input and output color showing up.

## Environment
napari: 0.2.11+27.gbd53848.dirty
Python: 3.7.4 (default, Aug 9 2019, 18:34:13) [MSC v.1915 64 bit (AMD64)]
Qt: 5.13.1
PyQt5: 5.13.1
VisPy: 0.6.4
NumPy: 1.16.6
SciPy: 1.3.1
scikit-image: 0.16.1
Dask: 2.5.2

GL version: 4.5.0 - Build 25.20.100.6472
MAX_TEXTURE_SIZE: 16384

## Additional context

I have been working a bit with the shape layer today and found several things that I don't understand (possibly bugs). Trying to reproduce them and file some more issues.

### LINKED ISSUE #987: visibility and selection behaviour of multiple polygons in a shape layer is bizarre (broken?)
## 🐛 Bug

When defining multiple polygons in a shape layer, they are not all visible in translucent mode although they are non-overlapping. Visibility sometimes stays on a newly created polygon, sometimes it immediately jumps back to a previously defined polygon.
When a polygon is selected and the mouse cursor is moved, visibility and selectability of polygons changes in a way that is inconsistent with the location of the mouse pointer.

This is possibly related to https://github.com/napari/napari/issues/986 (napari version etc is the same).

visibility_and_selection_of_polygons

sofroniewn:
# Description
This PR is another step in the direction of adding support for full world coordinates, but it does not actually do so yet.

It hopefully clears up some of the confusion in #922 (though it doesn't actually change the slicing).

It prepares the `layer.shape` property for deprecation by removing all of its use from the codebase (it is still used in tests). The property doesn't really make sense for `Points` / `Shapes` layers, and instead adding the `_extent_data` and `_extent_world` properties which give the extent (i.e. min and then max) values of the data in data coordinates and world coordinates (think of it as what you need to define and nD bounding box around the data, which is useful for many things). The shape is now the size of that box in world coordinates rounded to the nearest integers.

It also adds a `_extent_world` property to the layers list which takes the extrema across all the layers, and adds an `ndim` property to the layers list which takes the maximum across all the layers. It also adds a `_step_size` property to the layers list which controls how fine the steps will be in the dimensions sliders (but doesn't go all the way to put those sliders in a full world coordinate system).

With no data present the world size defaults to 512 x 512, which will close #986 and close #987

In the viewer_model it adds a `_sliced_extent_world` which is the extrema only for the sliced dimensions, which we need to setting cameras etc.

Hopefully this PR can go in with functionality relatively as is, and then there will be another PR that does all the slicing/ dims stuff. I'm very open to feedback on naming etc, and should probably add some dedicated tests before merge, but it is ready to be looked at @GenevieveBuckley @jni 

## Type of change


GenevieveBuckley:
> The shape is now the size of that box in world coordinates rounded to the nearest integers.

I'm not sure it makes sense to round to the nearest integers? My understanding is that "world coordinates" are the universe of space our data lives in, and therefore has some real space units attached to it. Changing the unit would dramatically alter what the rounding results in and you don't want things to alter when you change the spatial units on your dataset from millimeters to microns, for example.

> With no data present the world size defaults to 512 x 512

This is an improvement over what we have now, seems reasonable to me.

sofroniewn:
> I'm not sure it makes sense to round to the nearest integers?

Yeah completely. We're going to add a deprecation warning here and eventually remove this. The current definition is left over from the past

sofroniewn:
This PR is now ready for review.

I adds a unified way to calculate layer extent in both data and world coordinates for each layer type, including tests for each layer type.

It then adds a way to calculate the world extent of all the layers by taking extrema of the individual layers. This is tested too.

With no data present the world defaults to 512x512.

This PR now closes #1407 by making use of the scale and translate parameters to position the labels array correctly. If you have a single image that is scaled and translated you'll end up with a new labels of the same shape as the image with the same scale and translation. If you have something more complicated, the "right" thing will hopefully still happen too.

I had to make one vispy test ignore `multiscale` layers now, because the camera position is now taking some translation into account. The test was always a bit of a hack for the multiscale case as you had to do some strange transforms to make it ever work and I think it's best just to drop.

sofroniewn:
@tlambert03 curious from what you're thinking about for layer groups if you see this PR as stepping in the right direction or not.

Note the plan after this merges is to have follow-up PRs that change our sliders to use the world coordinates!

sofroniewn:
@Czaki can you confirm this fixes #1407 and gives you the desired behaviour when you create a new labels layer from the GUI?

napari/components/layerlist.py
```diff
@@ -186,6 +188,60 @@ def toggle_selected_visibility(self):
...
+            return min_scales
+
+    @property
+    def ndim(self):
+        """int: Maximum dimensionality of layers.
+
+        Defaults to 2 if no data is present.
+        """
+        ndims = [l.ndim for l in self]
+        return max(ndims, default=2)
```
tlambert03:
```suggestion
        return max((l.ndim for l in self), default=2)
```
not that we need speed here, but just cause it's interesting: that's about 3x faster on my computer than first making a list.

napari/components/layerlist.py
```diff
@@ -186,6 +188,60 @@ def toggle_selected_visibility(self):
             if layer.selected:
                 layer.visible = not layer.visible
 
+    @property
+    def _extent_world(self):
+        """(2, D) array: Extent of layers in world coordinates.
```
tlambert03: I'd really like to see us get away from this sort of type annotation (which can't be parsed by anything) if possible.  either a return type annotation (` -> np.ndarray`), and/or a numpy docstring `Returns` field since it's trickier to encode dimensionality at the moment in type annotations.

tlambert03 (commented):
I mostly just looked through the layerlist stuff and how it's being used, but I don't see anything in here that shouldn't work just fine with changes needed for layergroup 👍

_edit:_ I guess that main thing I was looking for is that these are all methods that are dynamically calculated from the children, rather than being stored on the list itself

Czaki:
@sofroniewn For my use case this fix works.

But GCD is not so good solution for future, because scale could be fraction.

napari/components/viewer_model.py
```diff
@@ -246,34 +244,30 @@ def active_layer(self, active_layer):
...
+    def _sliced_extent_world(self):
+        """(2, D) array: Extent of layers in world coordinates after slicing.
 
-        return size, corner
+        D is either 2 or 3 depending on if the displayed data is 2D or 3D.
+        """
+        if len(self.layers) == 0 and self.dims.ndim != 2:
+            # If no data is present and dims model has not been reset to 0
+            # It means someone has passed more than two axis labels which are
+            # being saved.
```
kevinyamauchi:
```python
 # It means someone has passed more than two axis labels which are
 # being saved.
```

I am not sure what this means. Can this be clarified?

napari/layers/base/base.py
```diff
@@ -414,10 +416,34 @@ def data(self):
...
+        return self._transforms['data2world'](self._extent_data)
+
+    @property
+    def shape(self):
+        """tuple: Size of layer in world coordinates (compatibility)."""
+
+        warnings.warn(
+            (
+                f'layer.shape is deprecated and will be removed '
+                f'in napari > 0.3.7'
```
kevinyamauchi:
```suggestion
                f'layer.shape is deprecated and will be removed '
                f'please use layer._extent_data[1] to get the shape'
                f'in napari > 0.3.7'
```

Should we provide a suggestion on what to do when `layer.shape` is deprecated?

kevinyamauchi: Ah, I now see that the suggestion would likely be different for some layers. I'm not sure the best thing to do there. I just know that as a user, I am usually a little miffed when I get told something is being deprecated without being suggested a path forward.

kevinyamauchi (approved):
This looks good to me!

I am curious if we can reduce code duplication by adding a util function for the tests that takes care of a majority of the testing. Perhaps it would take in the layer, the untransformed extents and if we wanted to be general the scale/translation. In this case, we would just have to create the layer in each test and call the util function.

I agree with @tlambert03 on moving the property docstrings to something more standard. I am fine if we hold off for our forthcoming docathon.

napari/_tests/utils.py
```diff
@@ -118,6 +118,8 @@ def check_viewer_functioning(viewer, view=None, data=None, ndim=2):
 def check_view_transform_consistency(layer, viewer, transf_dict):
     """Check layer transforms have been applied to the view.
 
+    Note this check only works for non-mulitscale data.
```
jni:
```suggestion
    Note this check only works for non-multiscale data.
```

napari/_tests/utils.py
```diff
@@ -129,25 +131,15 @@ def check_view_transform_consistency(layer, viewer, transf_dict):
         the transform property (i.e. `scale`, `translate`) and the value
         corresponding to the array of property values
     """
+    if layer.multiscale:
+        return
```
jni: I feel like this check should be changed to return True/False/None, or something like that, so that multiscale doesn't just look like it's passed. That seems to be courting trouble to me...

napari/components/_tests/test_viewer_model.py
```diff
@@ -193,16 +193,19 @@ def test_mix_dims():
...
-    # Add labels to empty viewer
+def test_new_labels_empyty():
+    """Test adding new labels layer to empty viewer."""
     viewer = ViewerModel()
     viewer._new_labels()
     assert len(viewer.layers) == 1
     assert np.max(viewer.layers[0].data) == 0
     assert viewer.dims.ndim == 2
+    # Default shape when no data is present is 512x512
+    np.testing.assert_almost_equal(viewer.layers[0].data.shape, (512, 512))
```
jni: Almost?

jni: More specifically, 
```suggestion
    np.testing.assert_equal(viewer.layers[0].data.shape, (512, 512))
```

napari/components/_tests/test_viewer_model.py
```diff
@@ -211,6 +214,39 @@ def test_new_labels():
     assert len(viewer.layers) == 2
     assert np.max(viewer.layers[1].data) == 0
     assert viewer.dims.ndim == 2
+    np.testing.assert_almost_equal(viewer.layers[1].data.shape, (10, 15))
+    np.testing.assert_almost_equal(viewer.layers[1].scale, (1, 1))
+    np.testing.assert_almost_equal(viewer.layers[1].translate, (0, 0))
```
jni:
```suggestion
    np.testing.assert_equal(viewer.layers[1].data.shape, (10, 15))
    np.testing.assert_equal(viewer.layers[1].scale, (1, 1))
    np.testing.assert_equal(viewer.layers[1].translate, (0, 0))
```

napari/components/_tests/test_viewer_model.py
```diff
@@ -211,6 +214,39 @@ def test_new_labels():
...
+    np.random.seed(0)
+    data = np.random.random((10, 15))
+    viewer.add_image(data, scale=(3, 3))
+    viewer._new_labels()
+    assert len(viewer.layers) == 2
+    assert np.max(viewer.layers[1].data) == 0
+    assert viewer.dims.ndim == 2
+    np.testing.assert_almost_equal(viewer.layers[1].data.shape, (10, 15))
+    np.testing.assert_almost_equal(viewer.layers[1].scale, (3, 3))
+    np.testing.assert_almost_equal(viewer.layers[1].translate, (0, 0))
```
jni:
```suggestion
    np.testing.assert_equal(viewer.layers[1].data.shape, (10, 15))
    np.testing.assert_equal(viewer.layers[1].scale, (3, 3))
    np.testing.assert_equal(viewer.layers[1].translate, (0, 0))
```

napari/components/_tests/test_viewer_model.py
jni: almost equal is for floating point operations that may be subject to rounding/approximation error.

napari/components/viewer_model.py
jni: I'd replace "it means" with "this situation can arise when". Is it also possible to end up here by adding and then deleting layers, or would that result in the dims model being reset?

napari/components/viewer_model.py
```diff
@@ -298,15 +292,17 @@ def reset_view(self, event=None):
...
-        else:
-            dims = self._calc_bbox()[1]
-            dims = [np.ceil(d).astype('int') if d > 0 else 1 for d in dims]
-            if len(dims) < 1:
-                dims = (512, 512)
-        empty_labels = np.zeros(dims, dtype=int)
-        self.add_labels(empty_labels)
+        """Create new labels layer filling full world coordinates space."""
+        extent = self.layers._extent_world
+        scale = self.layers._step_size
```
jni, line 305: Nice!

napari/layers/base/base.py
```diff
@@ -396,9 +396,11 @@ def _update_dims(self, event=None):
 
         self.dims.ndim = ndim
 
-        curr_range = self._get_range()
-        for i, r in enumerate(curr_range):
-            self.dims.set_range(i, r)
+        step_size = self.scale
+        # For now dims don't use world coordiates, but scaled data coordinates
```
```suggestion
        # For now dims don't use world coordinates, but scaled data coordinates
```

napari/layers/base/base.py
jni: @sofroniewn I see your thumbs up but not any suggestion of how you intend to fix this? I'll point out that we shouldn't suggest attributes beginning in `_`.

jni (commented):
Nice! I'm excited about this. Some comments:

- I agree with @kevinyamauchi that ideally the tests here should be parametrised rather than copy-pasted all over the place
- I'm actually not sure that it makes much sense to have scale/translate on points and shapes, since they are vector entities... But, it might make sense on input (e.g. annotating an image in pixel space rather than world space). I just don't have a clear idea of what would/should happen in the case of an image with scale/translate attributes and a new point layer getting added... What happens in this PR? Will the point data be in world coordinates or image coordinates, if I save it as csv?

napari/layers/base/base.py
sofroniewn: Will put in something else, will have to think about what makes most sense

sofroniewn:
> I'm actually not sure that it makes much sense to have scale/translate on points and shapes, since they are vector entities... But, it might make sense on input (e.g. annotating an image in pixel space rather than world space). I just don't have a clear idea of what would/should happen in the case of an image with scale/translate attributes and a new point layer getting added... What happens in this PR? Will the point data be in world coordinates or image coordinates, if I save it as csv?

After this PR points in the csv would be saved in "data coordinates" and not "world coordinates" (which include scale/ translate). But if we drop scale and translate then points only has one "coordinates" which is identical to "data coordinates" with scale of 1 translate of 0, so it's strictly more general to allow the scale and translate and seemed like a nice add to me, i can imagine times i'd use it to keep symmetry between points coordinates and image coordinates when an image also have a non-uniform scaling applied to it. given that it is strictly more general i suggest we keep it as is

i'll address other comments tomorrow

napari/_tests/utils.py
sofroniewn: I can make `return None` agreed though this feels a little dicey

napari/components/layerlist.py
sofroniewn: @tlambert03 going to go with `def _extent_world(self) -> np.ndarray` is that what you had in mind? Agreed it would be great to standardize this stuff, but I can do it now for things I am adding new

sofroniewn: Actually, I can do both

napari/components/viewer_model.py
sofroniewn: have tried to clarify, i think dims model will be reset in that case

napari/layers/base/base.py
sofroniewn: we can wait on deprecation. i've added a comment in code saying to deprecate after full world coords refactor is done

napari/components/_tests/test_viewer_model.py
```diff
@@ -193,16 +193,19 @@ def test_mix_dims():
     assert viewer.dims.ndim == 3
 
 
-def test_new_labels():
-    """Test adding new labels layer."""
-    # Add labels to empty viewer
+def test_new_labels_empyty():
```
jni: empyty LOL

napari/components/layerlist.py
```diff
@@ -186,6 +188,70 @@ def toggle_selected_visibility(self):
...
+        else:
+            extrema = [l._extent_world for l in self]
+            mins = [e[0] for e in extrema]
+            maxs = [e[1] for e in extrema]
+
+            min_v = np.nanmin(
+                list(itertools.zip_longest(*mins, fillvalue=np.nan)), axis=1
+            )
+            max_v = np.nanmax(
+                list(itertools.zip_longest(*maxs, fillvalue=np.nan)), axis=1
```
jni, line 213: Is this right? We right-align on dimensions, but this seems to be left-aligning?

napari/components/layerlist.py
```diff
@@ -186,6 +188,70 @@ def toggle_selected_visibility(self):
...
+        Returns
+        -------
+        step_size : array, shape (D,)
+        """
+        if len(self) == 0:
+            return np.ones(self.ndim)
+        else:
+            scales = [l.scale for l in self]
+            full_scales = np.array(
+                list(itertools.zip_longest(*scales, fillvalue=np.nan))
```
jni: As above — dimensions should be right-aligned. This would involve using `reversed()` on the above and then `reversed()` after zipping.

napari/components/viewer_model.py
jni: Looking at it again... 😂 I wonder whether this is the right approach. If I have a points layer, scale will be 1 and it'll throw a spanner in the works here, right? I wonder if we should only look at *image* layers when determining scale...?

napari/components/_tests/test_viewer_model.py
sofroniewn: fixed

napari/components/viewer_model.py
jni: @sofroniewn and I decided to punt on this one, see #1513 for tracking.

napari/components/layerlist.py
sofroniewn: correct, this is now fixed

napari/components/layerlist.py
sofroniewn: correct, this is now fixed

napari/components/viewer_model.py
sofroniewn: Yup, we can revisit in next PR when we hook up `_step_size` to sliders, will be more natural to address then, and won't actually be too hard to implement :-)

sofroniewn:
I have now fixed merge conflicts, this is now ready for final approval and merge @jni

jni (approved):
Ok, just exercised it a bit more with a points layer. Lookin' good. =D

Issue #1356, opened 2020-06-14
Color handling refactor


jni:
## 📚 Documentation

#1339 fixed a longstanding issue of directly mapping labels to colors. However, the parameter name "color_dict" is too restrictive in my opinion, since many color schemes can be conceived for labels. My suggestion is `colormap` which can be a callable mapping labels to colors, or a dict, or floats to colors, and could be in the future expanded to include e.g. an object that includes a GL shader to pass to vispy so that the full colormapping (labels -> graphics) can be done on the GPU, addressing #204 . 

This issue should be addressed before the next release so that we don't have to deprecate an API keyword shortly after introducing it.

sofroniewn:
Ooo, sounds very intriguing @jni. Picking up on https://github.com/napari/napari/pull/1339#issuecomment-643723379

> I think this PR should be reworked to match more closely what @kevinyamauchi did with point colouring, but with the additional "random" mode to match existing behaviour. Specifically, there could be a "direct" mode that grabs the colour from a single property (it does not need to be named "color" @sofroniewn, it could have any name and be settable), "colormapped" that uses a colormap, or "auto"/"random" that uses the current behaviour.

The next plan for this chunk of work was going to be to support coloring based on properties, see some comments from @kevinyamauchi here https://github.com/napari/napari/pull/1339#issuecomment-639650482. If we could think from an API perspective what we'd like that to look like and then rename this kwarg appropriately so that we don't have to rename it again, I think that would be best.

I'm not so sure about `colormap` as in the `Image` layer which this subclasses `colormap` means one very specific thing and now we'd have it mean something more general. I will think on it more too.

> This issue should be addressed before the next release so that we don't have to deprecate an API keyword shortly after introducing it.

Agreed

sofroniewn:
Looking at `Points` https://github.com/napari/napari/blob/master/napari/layers/points/_points_constants.py#L6-L15

```python
class ColorMode(StringEnum):
    """
    ColorMode: Color setting mode.
    DIRECT (default mode) allows each point to be set arbitrarily
    CYCLE allows the color to be set via a color cycle over an attribute
    COLORMAP allows color to be set via a color map over an attribute
    """
```

I read `DIRECT` mode as setting arbitrarily with the rgba array - which is why it is similar to the `color_dict` in `Labels` and that `COLORMAP` would be more akin to grabbing the color from a single property.

I guess stepping back what do I think we need:
- A place to store what color mode we're in
- Our autogenerated colormap
[...]

PR #1362, opened 2020-06-15
address comments on color_dict

liu-ziyang:
# Description
Address comments on renaming color_dict in label layer

## Type of change

# References
https://github.com/napari/napari/issues/1356

# How has this been tested?

- manually tested the behavior 

## Final checklist:

### DIFF
--- a/napari/components/add_layers_mixin.py
+++ b/napari/components/add_layers_mixin.py
@@ -466,8 +466,8 @@ class AddLayersMixin:
             the user and if the data is a list of arrays that decrease in shape
             then it will be taken to be multiscale. The first image in the list
             should be the largest.
-        color : dict of int to str
-            Custom label to color mapping
+    color : dict of int to str or array
+        Custom label to color mapping. Values must be valid color names or RGBA arrays.
 
         Returns
         -------
--- a/napari/components/add_layers_mixin.py
+++ b/napari/components/add_layers_mixin.py
@@ -415,7 +415,7 @@ class AddLayersMixin:
         blending='translucent',
         visible=True,
         multiscale=None,
-        color_dict=None,
+        color=None,
     ) -> layers.Labels:
         """Add a labels (or segmentation) layer to the layers list.
 
@@ -466,6 +466,8 @@ class AddLayersMixin:
             the user and if the data is a list of arrays that decrease in shape
             then it will be taken to be multiscale. The first image in the list
             should be the largest.
+        color : dict of int to str
+            Custom label to color mapping
 
         Returns
         -------
@@ -485,7 +487,7 @@ class AddLayersMixin:
             blending=blending,
             visible=visible,
             multiscale=multiscale,
-            color_dict=color_dict,
+            color=color,
         )
         self.add_layer(layer)
         return layer
--- a/napari/components/add_layers_mixin.py
+++ b/napari/components/add_layers_mixin.py
@@ -406,6 +406,7 @@ class AddLayersMixin:
         *,
         num_colors=50,
         properties=None,
+        color=None,
         seed=0.5,
         name=None,
         metadata=None,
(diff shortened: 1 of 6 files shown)


napari/view_layers.py
```diff
@@ -440,6 +441,8 @@ def view_labels(
         Dimension names.
     show : bool, optional
         Whether to show the viewer after instantiation. by default True.
+    color : dict of int to str
+        Custom label to color mapping
```
sofroniewn:
```suggestion
    color : dict of int to str or array
        Custom label to color mapping. Values must be valid color names or RGBA arrays.
```

napari/layers/labels/labels.py
```diff
@@ -108,6 +108,8 @@ class Labels(Image):
 
         In ERASE mode the cursor functions similarly to PAINT mode, but to
         paint with background label, which effectively removes the label.
+    color : dict of int to str
+            Custom label to color mapping
```
```suggestion
    color : dict of int to str or array
        Custom label to color mapping. Values must be valid color names or RGBA arrays.
```

napari/components/add_layers_mixin.py
```diff
@@ -466,6 +466,8 @@ def add_labels(
             the user and if the data is a list of arrays that decrease in shape
             then it will be taken to be multiscale. The first image in the list
             should be the largest.
+        color : dict of int to str
+            Custom label to color mapping
```
sofroniewn:
```suggestion
    color : dict of int to str or array
        Custom label to color mapping. Values must be valid color names or RGBA arrays.
```

sofroniewn:
Tests are failing as doc strings aren't synched. You've added `color` to the Attributes section in `Labels`, but also need to add it to the `Parameters` section (which is used for the doc string comparison). Can you also move the `color` parameter to come after `properties` in both the function signature and the doc strings for all cases where it appears (Labels, add_labels, view_labels) as we've tried to group the layer specific kwargs together. Hopefully that should fix things!!

Also can you address the comment here https://github.com/napari/napari/pull/1339/files#r439786537 and fix that doc string/ parameter name.

If you could also comment on https://github.com/napari/napari/pull/1339/files#r439794714 that would be great too

Thanks!!

jni:
@ziyangczi @sofroniewn I've pushed some minor docstring updates to this branch, which should be good to go after tests pass!

jni (approved):

sofroniewn (approved):
LGTM! Will merge now, windows fail is not related to this PR and popping up on all PRs right now

liu-ziyang:
> @ziyangczi @sofroniewn I've pushed some minor docstring updates to this branch, which should be good to go after tests pass!

hey, is there a plugin that helps you catch docstring format issues? Being still new to python docstrings, I wish there is some way auto-detecting my mistakes

sofroniewn:
> hey, is there a plugin that helps you catch docstring format issues? Being still new to python docstrings, I wish there is some way auto-detecting my mistakes

It' benefit a lot from one too! I think I've heard some discussion about this in places, but I'm not sure. Maybe @napari/core-devs we think about adding this to our linting?

tlambert03:
There’s https://pypi.org/project/pydocstyle/
(And a flake8 extension for it)

I have it enabled locally in my IDE, which is an “opt-in” option for devs.  Or we can add it and enforce it in CI too, (it’s enabled at magicgui) but it can definitely get a little “chatty”.

kevinyamauchi:
> It' benefit a lot from one too! I think I've heard some discussion about this in places, but I'm not sure. Maybe @napari/core-devs we think about adding this to our linting?

I think @jni would benefit too because he wouldn't have to catch so many of my improperly-indented docstrings ;)

> There’s https://pypi.org/project/pydocstyle/
(And a flake8 extension for it)

Thanks! I will try it out.

tlambert03:
If you’re using VSCode, there is a handy extension that will give you a docstring snippet based on your signature:

https://marketplace.visualstudio.com/items?itemName=njpwerner.autodocstring

Supports numpy format.  There’s probably similar extensions for other IDEs

PR #1367, opened 2020-06-17
WIP: Add save method to the viewer

### LINKED ISSUE #1286: Add viewer.save method
## 🚀 Feature

Currently we have `viewer.open()` to open a dataset, but `viewer.layer[].save` to save data. It would be neatly symmetric to provide `viewer.save(layer, path)`.

I thought we agreed that we would do this but clearly I'm misremembering that discussion...

jni:
# Description

As mentioned at meeting, it's been bugging me that we have `viewer.open` but not `viewer.save`. This is a proposal to add `viewer.save`. The signature is a bit different from `viewer.layers.save`, but I think it makes sense. Looking for feedback though before I add some tests, in case people don't like this API.

Closes #1286 

## Type of change

# References

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/components/viewer_model.py
+++ b/napari/components/viewer_model.py
@@ -5,7 +5,7 @@ import numpy as np
 from .add_layers_mixin import AddLayersMixin
 from .dims import Dims
 from .layerlist import LayerList
-from .layers import Layer
+from ..layers import Layer
 from ..utils.event import EmitterGroup, Event
 from ..utils.key_bindings import KeymapHandler, KeymapProvider
 from ..utils.theme import palettes
--- a/napari/components/viewer_model.py
+++ b/napari/components/viewer_model.py
@@ -5,6 +5,7 @@ import numpy as np
 from .add_layers_mixin import AddLayersMixin
 from .dims import Dims
 from .layerlist import LayerList
+from .layers import Layer
 from ..utils.event import EmitterGroup, Event
 from ..utils.key_bindings import KeymapHandler, KeymapProvider
 from ..utils.theme import palettes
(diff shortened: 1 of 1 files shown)


napari/components/viewer_model.py
```diff
@@ -504,6 +505,75 @@ def stack_view(self):
...
+        list of str
+            File paths of any files that were written.
+        """
+        from ..plugins.io import save_layers
+
+        if isinstance(layer, Layer):
+            layers = [layer]
+        elif isinstance(layer, [int, str]):
+            layers = [self.layers[layer]]
+        else:  # sequence
```
tlambert03, line 569: any thoughts on using `layers = [layers] if not is_sequence(layers) else layers` and then doing the conversion on `for layer in layers` instead of this?

(`from napari.utils.misc import is_sequence`)

napari/components/viewer_model.py
```diff
@@ -504,6 +505,75 @@ def stack_view(self):
         """
         self.grid_view(n_row=1, n_column=1, stride=1)
 
+    def save(self, path: str, *, layer=None, plugin):
```
tlambert03, line 508: type hint?

napari/components/viewer_model.py
```diff
@@ -504,6 +505,75 @@ def stack_view(self):
...
+        """Save given or all layers to a path using writer plugins.
+
+        If ``plugin`` is not provided and only one layer is given, then we
```
tlambert03, line 511: i know docs are always good... but it slightly bothers me that we now have this very long explanation mostly repeated in three places. (io.py, layerlist.py, and here).  I know that referencing docstrings of other methods doesn't work great for interactive usage... but it works fine for sphinxdocs.  Thoughts?

tlambert03 (commented):
this is pretty close, shall we push to get it in?  I left relatively minor comments

sofroniewn:
Closed as we need to have further discussion in #1286. If we decide to go with this approach we can reopen

Issue #1049, opened 2020-03-13
Docs locations and format


tlambert03:
I think we could use some more discussion on exactly where our docs go, and what format they are in.

I agree with @sofroniew's https://github.com/napari/napari/issues/1026#issuecomment-598867558 that it would be nice to have a place to put some "troubleshooting" notes.  At the moment, the pattern is to (probably) make another markdown file in `docs/developers`...  
That worked pretty well when the docs were mostly viewed on github, and the sphinxdocs were strictly the autogenerated API docs, but it's really not ideal for lots of documentation (the readme has to basically become the table of contents for the dev docs, and integration with sphinxdocs becomes clunky)

It would be pretty easy make a "troubleshooting" section on our main [sphinx docs](https://napari.org/docs/) page, but the best place for that is really in `docs/source`, and _probably_ in `.rst` format as long as we're going with sphinx.  I personally prefer markdown in general, but resign myself to using rst when working with sphinx.

Anyway, therein lies the conflict... we have a handful of ever-growing markdown files that _can_ be integrated into the sphinx docs as is being done in #967, but I kinda saw that more as a "stop-gap" way to put our existing markdown docs somewhere visible in sphinx... rather than a sustainable long-term plan for expansion of docs.

thoughts?

will-moore:
I only found https://github.com/napari/napari/blob/master/docs/source/plugins/for_plugin_developers.rst by searching for "plugin" in the source code. Don't know if I missed an easier way to find it?
But the page itself was great. Ended up creating https://github.com/will-moore/napari-idr (very work-in-progress).

tlambert03:
Hi @will-moore, our docs are served at https://napari.org/docs/
(and that specific page is [here](https://napari.org/docs/plugins/for_plugin_developers.html))

glad you found it helpful!  Definitely will be watching napari-idr with interest!

tlambert03:
by the way ... I _just about an hour ago_ updated the [cookiecutter](https://github.com/napari/cookiecutter-napari-plugin) significantly (and added a lot of info to the readme there) ... now with auto-deployment and testing using github actions, and version control based on git tags.  If you want to try the cookiecutter again, hopefully it won't be too much to move the code you did write into it, to benefit from all the new stuff.

will-moore:
Thanks for the update.
We've actually moved the napari plugin code to https://github.com/ome/ome-zarr-py which will be more general-purpose.

jni:
[...]

PR #1368, opened 2020-06-17
Fix docs version, reformat and add explanations

### LINKED ISSUE #1365: Docs version is incorrect, and is missing pages
## 📚 Documentation

The current documentation shows napari version 0.1.dev742+g0adb11c:

Screen Shot 2020-06-16 at 1 50 53 pm

I suspect this might be an issue with #1324 interacting badly with how we build docs.

Additionally, the pages under [`docs/explanations`](https://github.com/napari/napari/tree/0adb11c53dbd29ae82fbf46b68c61bda63128775/docs/explanations) don't appear in the docs. I think the fix is to add `explanations` to the toctree here:

https://github.com/napari/napari/blob/0adb11c53dbd29ae82fbf46b68c61bda63128775/docs/source/index.rst

tlambert03:
# Description
fixes #1365 by fetching tags prior to `pip install .` when building docs (as @jni suspected).
Adds the "explanations" folder to the sphinx docs, and moves the "developers" and "explanations" folders into `docs/source` with the rest of the docs.

I had to reformat number of things in the markdown, so a couple reminders when writing markdown that we want to use in docs:

- There should only be a single top level heading (`# Title`) in a document
- Don't use `**bold**` text as a heading, use `#, ##, ###` instead 

Also, it's worth pointing out that the linking strategy used in our roadmap (e.g. `[mission](MISSION_AND_VALUES.md#our-mission)`) works fine when viewing in github... but most of the links are broken in the [sphinx docs](https://napari.org/docs/developers/ROADMAP_0_3.html).  

## Type of change


jni (approved):
Wow! Thanks for the massive number of  fixes @tlambert03!

sofroniewn (approved):
Looks great @tlambert03 thanks for this, and sorry you had to do all this clean up! We'll get #1049 sorted out soon and then hopefully this will become more robust

sofroniewn:
Docs render great for me - see https://napari.org/docs/explanations/rendering.html

PR #1371, opened 2020-06-18
Revert "remove scipy.stats import (#1250)"

jni:
This reverts commit 96e9722419a5b0adddbc14fe12c69a7d769f8758.

Fixes #1347

See https://github.com/napari/napari/issues/1347 for more details.

# Description

## Type of change

- [ ] Bug-fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# References

# How has this been tested?

- [ ] example: the test suite for my feature covers cases x, y, and z
- [ ] example: all tests pass with my change

## Final checklist:
- [ ] My PR is the minimum possible work for the desired functionality
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #1372, opened 2020-06-18
Refactor all tests to hide GUI

### LINKED ISSUE #1363: Splitting gui vs non-gui tests
## 🚀 Feature
Make it possible clear how to run gui/ non-gui tests independently of each other

## Motivation

Right now you can run all the tests together which can take multiple minutes and which sometimes causes Widgets to pop up on the screen, which can give the impression your machine has been "taken over", and which can delay development.

## Pitch

It should be possible to run all our non gui tests together

## Additional context

Note that one piece of complexity here is that not all Qt tests actually launch the viewer. I think there are some ways to run Qt tests which prevent the viewer from launching that @tlambert03 implemented, but that the viewer needs to launch for some of them.

One option would be just to have all the qt tests and non-qt tests grouped.

Right now the qt tests are found in
- https://github.com/napari/napari/tree/master/napari/_tests
- https://github.com/napari/napari/tree/master/napari/_qt/_tests
- https://github.com/napari/napari/tree/master/napari/_vispy

Non-qt tests are found in
- https://github.com/napari/napari/tree/master/napari/components/_tests
- https://github.com/napari/napari/tree/master/napari/layers/_tests (though it looks like the test_dask_layers.py might violate this and need to be moved to `napari/_tests`)
- https://github.com/napari/napari/tree/master/napari/utils/_tests
- https://github.com/napari/napari/tree/master/napari/plugins/_tests

If any qt tests have leaked into those locations they should probably be moved.

One option would be to use markers in pytest https://docs.pytest.org/en/latest/example/markers.html which can then be included/ excluded (h/t @pwinston) Another is to see if there are ways to make aliases to just run tests in the non-qt folders and make sure our organization is obeyed.

Thoughts @jni @tlambert03 @ziyangczi

tlambert03:
# Description
this is how I would propose we fix #1363
most of the tests that are showing gui elements don't need to be.  So rather than move tests or introduce new keywords and conventions into our tests to run or not run those tests, I propose we try to write tests that don't need to show anything.  In all but 2 cases (screenshot tests, and the full-screen keybinding test), I was able to remove the GUI without changing what the test is covering. 
couple strategies:
- use [`isVisibleTo(parent_widget)`](https://doc.qt.io/qt-5/qwidget.html#isVisibleTo) instead of `isVisible` which prevents the need to show the parent widget to test visibility.
- for `viewer_factory` always start without `show=True` ... if `show=True` is required, take a closer look at why.
- if we're just testing to see whether `show()` has been called.  mock the show method and `assert_called_once()` ... (we can trust Qt that `show()` actually works).
- if all else fails (as in the screenshot tests), we can add `if os.getenv('CI')` to restrict those tests to CI... or use `export CI` to "opt in" to testing them locally.

(sidenote: I think if we could figure out #1109, we could also hide the screenshot tests). and the  "full-screen keybinding" test is basically just calling a Qt method, so I feel fine relegating that one to CI as well since we don't need to test an external library.

 
## Type of change

- [x] Refactor tests

# How has this been tested?

- [x] all tests pass (locally at least) with my change

## Final checklist:

### DIFF
--- a/docs/source/developers/CONTRIBUTING.md
+++ b/docs/source/developers/CONTRIBUTING.md
@@ -109,6 +109,14 @@ pip install -r requirements/test.txt
 pytest
 ```
 
+There are a very small number of tests (<5) that require showing GUI elements, (such
+as testing screenshots). By default, these are only run during continuous integration.
+If you'd like to include them in local tests, set the environment variable "CI":
+
+```sh
+export CI=1
+```
+
 ## Writing Tests
 
 Writing tests for new code is a critical part of keeping napari maintainable as
--- a/napari/_qt/_tests/test_qt_dims.py
+++ b/napari/_qt/_tests/test_qt_dims.py
@@ -3,10 +3,11 @@ from sys import platform
 
 import numpy as np
 import pytest
-from qtpy.QtCore import Qt
-
 from napari._qt.qt_dims import QtDims
 from napari.components import Dims
+from qtpy.QtCore import Qt
+
+from unittest.mock import patch
 
 
 def test_creating_view(qtbot):
@@ -18,7 +19,6 @@ def test_creating_view(qtbot):
     view = QtDims(dims)
 
     qtbot.addWidget(view)
-    view.show()
 
     # Check that the dims model has been appended to the dims view
     assert view.dims == dims
@@ -29,7 +29,7 @@ def test_creating_view(qtbot):
     assert np.sum(view._displayed_sliders) == view.dims.ndim - 2
     assert np.all(
         [
-            s.isVisible() == d
+            s.isVisibleTo(view) == d
             for s, d in zip(view.slider_widgets, view._displayed_sliders)
         ]
     )
@@ -43,7 +43,6 @@ def test_changing_ndim(qtbot):
     view = QtDims(Dims(ndim))
 
     qtbot.addWidget(view)
-    view.show()
 
     # Check that adding dimensions adds sliders
     view.dims.ndim = 5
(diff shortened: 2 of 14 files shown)


tlambert03:
(windows fail is the same as before... maybe fixed by #1371 ?)

sofroniewn (commented):
I'm fine with this approach. Definitely good to minimize the tests that force the viewer popping up, and having the environment variable to control it to be off by default seems reasonable.

I guess is `CI` something that is getting set automatically already? Do we do that, or is that done by our CI. I'd be tempted to rename `CI` to `NAPARI_CI_ONLY` and have us set it to be more explicit, but otherwise this looks good to me. What do you think @tlambert03? 

Thoughts @ziyangczi? Can you trying this out and give this a review too @ziyangczi?

tlambert03:
CI is an environment variable convention used by pretty much every CI provider, and cirrus sets it.  For that reason I’m -1 on using a napari specific variable name

sofroniewn (approved):
This looks good to me. Will see if anyone else wants to weigh in and then merge tomorrow

sofroniewn:
Maybe I was a little quick to merge this. I just tested it out more locally and my screen still gets taken over. See

Screen Shot 2020-06-20 at 12 19 32 PM

If I run the `napari/_tests` folder and try and leave the terminal to another window. Does this not happen to you @tlambert03?

I don't see the actual GUI pop up like I did before, just that gray screen so this PR is definitely an improvement on before

tlambert03:
that's the phenotype of the `full-screen` keybinding leaking through.  It hadn't been happening for me... but yeah, it is right now :/  Lemme see what happened

tlambert03:
interesting.  it's actually rather context dependent.  When I let the tests run without going to another window, nothing pops up.  but if I go elsewhere and type and stuff (the whole point), it does pop up.  furthermore, if I just run `napari/_tests/test_napari.py`, it pops up every time.  so I must have missed a test or two in there.  will investigate

sofroniewn:
> interesting. it's actually rather context dependent. When I let the tests run without going to another window, nothing pops up. but if I go elsewhere and type and stuff (the whole point), it does pop up

yes, I had noticed something to that effect! Thanks for looking in!!

PR #1373, opened 2020-06-19
Add ability to run napari script with/without gui_qt from CLI

### LINKED ISSUE #1333: napari CLI script-runner
## 🚀 Feature

In the last meeting and in https://github.com/napari/tutorials/issues/70, we discussed the idea of being able to run a napari script from the command line *regardless* of whether it contained `napari.gui_qt()`.  Something like:
```sh
napari path/to/script.py
```

The question here is whether we want to just hijack the positional "images" (paths) argument in the existing CLI, look for something ending with `.py`, and run it as a script if so (as done in https://github.com/napari/tutorials/issues/70#issuecomment-635691269) ... Or if we should consider moving to subparsers like `napari open image.tif` and `napari run script.py`.
While it _is_ possible to arrange it such that `napari image.tif` _still_ opens a file... it gets more complicated to do it in a "clean" way such that we see the same help menu for both `napari open --help` and `napari --help`.  

## Options

1. **Do** use subparsers (`open` and `run`), where `open` is the default... and live with the fact that `napari --help` will look like this:
```sh
$ napari --help
napari command line viewer.

positional arguments:
  {open,run}

optional arguments:
  -h, --help  show this help message and exit
  --version   show program's version number and exit
  --info      show system information and exit
  --citation  show citation information and exit
```
and `napari open --help` has all the useful options, _even though_ they also apply even if you didn't use `napari open`

```sh
$ napari open --help
positional arguments:
  images             image(s) to view.

optional arguments:
  -h, --help         show this help message and exit
  -v, --verbose      increase output verbosity
  --stack            concatenate multiple input files into a single stack.
  --plugin PLUGIN    specify plugin name when opening a file
  --layer-type TYPE  force file to be interpreted as a specific layer type.
                     one of {'shapes', 'points', 'vectors', 'labels',
                     'image', 'surface'}
```
2. **Don't** use subparsers, and simply inspect the list of paths passed, and if any (?) of them is a `.py` file, run it as a script (and ignore any other position args?)... 

3. **Don't** use subparsers, but require that the first positional argument be the string "run". 
 This will work for `napari run path/to/script.py` but not for `napari path/to/script.py` ... and "`run`" merely "looks" like an actual subcommand... but it's not, and doesn't come with any of its own help.

4. don't give up on subparsers (option #1) ... but allow for the fact that `napari.__main__` may need to get more complicated to achieve the desired subcommand & help behavior we're after.

@GenevieveBuckley, @jni?

tlambert03:
# Description
as discussed in https://github.com/napari/tutorials/issues/70 and #1333, this PR adds the ability to run a python script inside of a napari/qt event loop (regardless of whether the script contains `napari.gui_qt()`), by using `napari path/to/script.py` on the command line.

If any of the positional arguments end in .py and more than a single positional argument is used, the program exits and prints an error.

closes #1333 

## Type of change

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/__main__.py
+++ b/napari/__main__.py
@@ -3,6 +3,7 @@ napari command line viewer.
 """
 import argparse
 import logging
+import runpy
 import sys
 from ast import literal_eval
 from textwrap import wrap
@@ -92,7 +93,7 @@ def main():
         epilog="optional layer-type-specific arguments (precede with '--'):\n"
         + "\n".join(kwarg_options),
     )
-    parser.add_argument('images', nargs='*', help='image(s) to view.')
+    parser.add_argument('paths', nargs='*', help='path(s) to view.')
     parser.add_argument(
         '-v',
         '--verbose',
@@ -135,10 +136,10 @@ def main():
 
     args, unknown = parser.parse_known_args()
     # this is a hack to allow using "=" as a key=value separator while also
-    # allowing nargs='*' on the "images" argument...
-    for idx, item in enumerate(reversed(args.images)):
+    # allowing nargs='*' on the "paths" argument...
+    for idx, item in enumerate(reversed(args.paths)):
         if item.startswith("--"):
-            unknown.append(args.images.pop(len(args.images) - idx - 1))
+            unknown.append(args.paths.pop(len(args.paths) - idx - 1))
     kwargs = validate_unknown_args(unknown) if unknown else {}
 
     # parse -v flags and set the appropriate logging level
@@ -152,7 +153,7 @@ def main():
 
     if args.plugin:
         # make sure plugin is only used when files are specified
-        if not args.images:
+        if not args.paths:
             sys.exit(
                 "error: The '--plugin' argument is only valid "
                 "when providing a file name"
@@ -162,14 +163,26 @@ def main():
         # so remove --plugin from sys.argv to prevent that warningz
         sys.argv.remove('--plugin')
 
-    with gui_qt(startup_logo=True):
-        view_path(
-            args.images,
-            stack=args.stack,
-            plugin=args.plugin,
-            layer_type=args.layer_type,
-            **kwargs,
-        )
+    if any(p.endswith('.py') for p in args.paths):
+        if len(args.paths) > 1:
+            sys.exit(
+                'When providing a python script, only a '
+                'single positional argument may be provided'
+            )
+
+        with gui_qt(startup_logo=False) as app:
+            runpy.run_path(args.paths[0])
+            if getattr(app, 'existed', False):
+                sys.exit()
+    else:
+        with gui_qt(startup_logo=True):
+            view_path(
+                args.paths,
+                stack=args.stack,
+                plugin=args.plugin,
+                layer_type=args.layer_type,
+                **kwargs,
+            )
 
 
 if __name__ == '__main__':
--- a/napari/__main__.py
+++ b/napari/__main__.py
@@ -170,7 +170,9 @@ def main():
                 'single positional argument may be provided'
             )
 
-        with gui_qt(startup_logo=False) as app:
+        with gui_qt(startup_logo=True) as app:
+            if hasattr(app, '_splash_widget'):
+                app._splash_widget.close()
             runpy.run_path(args.paths[0])
             if getattr(app, '_existed', False):
                 sys.exit()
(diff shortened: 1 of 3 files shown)


napari/__main__.py
```diff
@@ -162,14 +163,26 @@ def main():
...
-            **kwargs,
-        )
+    if any(p.endswith('.py') for p in args.paths):
+        if len(args.paths) > 1:
+            sys.exit(
+                'When providing a python script, only a '
+                'single positional argument may be provided'
+            )
+
+        with gui_qt(startup_logo=False) as app:
```
sofroniewn: I'd go `startup_logo=True` for launching here, especially as file may do some complex loading, will be nice to give feedback right away that something is happening

sofroniewn (approved):
This is great, just gave it a try.

Do we want to change all our examples to drop with context manger? Should we do that in this PR too? If so we'll need to update this message too to drop the reference to them. https://github.com/napari/napari/blob/master/napari/viewer.py#L55-L56

We might also want to add the `launch with napari` option in that error message too even in this PR. We'll also want to probably update the relevant section of our launching napari tutorial to include this option too https://github.com/napari/tutorials/blob/master/fundamentals/getting_started.md#python-script-usage (can happen later, but just want to flag now)

napari/__main__.py
tlambert03: problem is that the start logo then doesn't go away.  If you want that behavior, I can add the `splash_widget` as an attribute of the application in `event_loop.py` and then manually close it in the CLI?

tlambert03: just applied that hack.  try again

tlambert03:
> Do we want to change all our examples to drop with context manger?

I don't have a strong opinion on this...  can do if that's the general consensus.

napari/__main__.py
sofroniewn: Hmm it seemed to work for me already - see this gif, where I did `startup_logo=True` and ran `add_image.py` without the context manager there

launch

Interestingly though if the `.py` file also contains a context manager then it just hangs. Is that expected?

I will try your new approach now

sofroniewn: Now it with your new approach it works regardless, and it's actually nicer as the splash screen closes faster and doesn't overlap with the viewer. I like this behaviour quite a lot:

launch_2

sofroniewn:
> > Do we want to change all our examples to drop with context manger?
>
> I don't have a strong opinion on this... can do if that's the general consensus.

I vote yes, but let's see what @jni and @GenevieveBuckley think. I think one of the motivations for this PR according to https://github.com/napari/tutorials/issues/70 is that we can now have sphinx-gallery and notebook compatible examples with the same syntax if we drop the context manager from our examples :-)

napari/__main__.py
tlambert03: yeah, the previous one didn't work, but only if you had a context manager in the script.  that was the problem

GenevieveBuckley:
I think we probably want to change the examples (at least eventually) but that should maybe happen in a separate PR. We'd still leave at least one as an example of using the with gui_qt context manager.

I also feel like it's a bit weird to have some examples in the tutorials repository, and some in the examples folder here (and this is only if you know to look in the source code, I don't think they're linked to from our docs right now?) That's likely a discussion for another time though.

jni:
I'd be very cautious about removing `with gui_qt()` from examples at first. Ideally, we would want to have a js-based button in the examples gallery to include or strip the `gui_qt` blocks, with a hover button that said something like "Usage with python" vs "Usage with IPython/Jupyter/napari command". As an example of what's possible, see the prompt (`>>>`) button in the code blocks in this page:

https://scikit-image.org/docs/stable/user_guide/numpy_images.html

Creating this button would be a significant undertaking, but I think it is worth waiting until then. Running examples with `napari path/to/file.py` is super cute but also super non-standard, so I would like to be gentle with this transition.

At any rate, let's leave the examples to a later PR.

By the way, it also occurs to me that this doesn't work with examples that require more than one viewer, does it?

jni:
**Edit:** that require a viewer, waiting for user input, then a viewer on the output.

jni (approved):

sofroniewn:
Yup, fine with me to wait on changing the examples until we are actually addressing them. Great to get the basic functionality in! Will be fun to use :-)

PR #1374, opened 2020-06-20
Add text to shapes and points via TextManager

kevinyamauchi:
# Description
This PR adds the ability to add text to existing layer types. In particular, it introduces a new `TextManager` class that groups the properties and methods associated with setting and interacting with the text.

Similar to colors, text can be set as a function of the layer properties.
```python
>>> data = [[0, 0], [10, 10]]
>>> properties = {'confidence': [0.5, 1], 'class': ['A', 'B']}
>>> viewer.add_points(data, text='confidence')
>>> viewer.text
['0.5', '1']
```
We can also set the text using formatting similar to f-strings.
```python
>>> data = [[0, 0], [10, 10]]
>>> properties = {'confidence': [0.5, 1], 'class': ['A', 'B']}
>>> viewer.add_points(data, text='class={class}, confidence={confidence:.2f}')
>>> viewer.text
['class=A, confidence=0.50', 'class=B, confidence=1.00']
```

Finally, we can pass in a dictionary of the `TextManager` kwargs to set all of text properties. Excerpt from the `add_points_with_text.py` example below:
```python
text = {
    'text': 'Confidence is {confidence:.2f}',
    'size': 20,
    'color': 'green',
    'translation': np.array([-30, 0]),
}

    points,
    properties=properties,
    text=text,
    ...
)
```
Screenshot 2020-06-20 at 15 51 32

## Type of change

# References
Closes #600 and #593 

# How has this been tested?
- [ ] added tests for TextManager and utils

## Final checklist:


pablooriol2:
This is awesome! Thanks for picking it up again @kevinyamauchi 

You show examples with a points layer, and mention that this feature would work with any kind of layer type, right? Namely, labels, shapes, image, surface & vectors? If so, would the feature be used in the same way for all layer types?

Also - may I suggest to have the option of text placement. Case in point would be, for example, the addition of text to labels of a mask, whereby text would be displayed at the centroid position of each label if that makes sense? This property might already be thought of as it might be crucial in some layer types like images, as I don't see how text would be placed otherwise

kevinyamauchi:
Thanks for the feedback @pablooriol2 !

> You show examples with a points layer, and mention that this feature would work with any kind of layer type, right? Namely, labels, shapes, image, surface & vectors? If so, would the feature be used in the same way for all layer types?

This initial PR will add text to just the points layer. We definitely want to add text to the shapes layer as well. The goal is to have the same way of interfacing with text wherever we make it available.

I plan to make it possible to hide the points, so for layers that do not have text integrated, you can add a text layer on top. This isn't as nice as full integration, but I think it could be a decent intermediate workaround as we figure out the right way to implement text. What do you think?

> Also - may I suggest to have the option of text placement. Case in point would be, for example, the addition of text to labels of a mask, whereby text would be displayed at the centroid position of each label if that makes sense? This property might already be thought of as it might be crucial in some layer types like images, as I don't see how text would be placed otherwise

Feature suggestions are always welcome! For now, I have added a translation and a rotation argument for offsetting the text from the point coordinate. Currently, it applies the same translation and rotation for all text elements in the layer. Would this meet your needs? It looks like vispy currently doesn't support specifying a unique offset/rotation for each text element (see [here](http://vispy.org/visuals.html#vispy.visuals.TextVisual)). It would be possible to see about contributing that feature upstream to vispy in the future, but I wouldn't be able to do it in this PR.

sofroniewn:
@kevinyamauchi there was more discussion of this PR this afternoon with @jni and he likes the approach you've taken here with `TextManager` as an object that can put onto Points or Shapes etc. and not the approach of making it its own layer with linked data which he was worried would be very complex. In the language of @pwinston Points now "has a" TextManager which is good and we can resuse that object again. @jni did agree thought that having it be it's own `TextVisual` makes sense so that we can reuse that easily too if that is easier than the compound visual.

As for UI elements we'd then need to add more to the top left controls panel because we now wouldn't have two layers, but maybe that's ok, and we'll figure out stylings as we go. I know this is a little different then we left it Wednesday, but it is very in line with how this PR is right now. How does that sound?

kevinyamauchi:
All of that sounds great! I should have some time to push on it this weekend.

kevinyamauchi:
I have completed most of the integration of `TextManager` into the points model (see the `add_points_with_text.py` example). I made it so that text can only be set as mapped to properties (either as the value of a property or an f-string utilizing one or more properties). My rationale is that the text shouldn't add new data to the layer, but should be a view of annotations of the layer data (i.e., properties in napari parlance). How does that sound, @sofroniewn and @jni?

I have made a hacky change to the vispy file just for testing the points layer out with text. I agree that it would be nice to have a separate vispy file for text for easier integration with other layers, so I will work on that next!

sofroniewn:
> I made it so that text can only be set as mapped to properties (either as the value of a property or an f-string utilizing one or more properties). My rationale is that the text should add new data to the layer and should be a view of annotations of the layer data (i.e., properties in napari parlance).

ooo I love this, yes makes lots of sense - anything which is a property "one per point" should be settable via `properties` like that. I do wonder if we'd also support a list of strings as a very simple API, but internally then we'd probably just want one representation.

napari/layers/points/points.py
```diff
@@ -49,6 +51,8 @@ class Points(Layer):
     properties : dict {str: array (N,)}, DataFrame
         Properties for each point. Each property should be an array of length N,
         where N is the number of points.
+    text : str, array, dict
+        Text to be displayed with the points.
```
sofroniewn: Can you expand here what the different options do. Would a single str get applied to all the points?

sofroniewn:
How are you thinking about exposing GUI controls to the text? Like color/ values etc. Maybe for this PR it's sufficient just to have a `text_visible` property that can be controlled by a checkbox, or you could add more

napari/layers/points/points.py
kevinyamauchi: Will do!

kevinyamauchi:
> I do wonder if we'd also support a list of strings as a very simple API, but internally then we'd probably just want one representation.

I think we should consider it. Internally, we could add them to the properties. We would have to do some magic on the key naming (e.g., try `text`, and then increment `text_1`, `text_2`,..., `text_n` until we find an unused key).

I think the think a counterargument is that we should encourage users to use the `properties` and it really isn't much more code to do so. I also think allowing users to set text directly via an array of strings is a bit less transparent as to the values actually get stored.

I don't have strong feelings either way and I'm happy to hear everyone's thoughts!

> How are you thinking about exposing GUI controls to the text? Like color/ values etc. Maybe for this PR it's sufficient just to have a text_visible property that can be controlled by a checkbox, or you could add more

My thought is we would have a "text visibility" checkbox and if the text is visible, show GUI elements for text size, color, and maybe rotation (we had those in #600). I am considering also including a dropdown menu populated with the available property keys. Perhaps we can in the future add something for constructing (and previewing?) the f-strings.

pablooriol2:
Apologies for the late reply @kevinyamauchi 

> I plan to make it possible to hide the points, so for layers that do not have text integrated, you can add a text layer on top. This isn't as nice as full integration, but I think it could be a decent intermediate workaround as we figure out the right way to implement text. What do you think?

Sounds good to me. Better than not having the option of adding text. So although it may not be 'clean' and still works somehow, I'm fully supportive of that.

> Feature suggestions are always welcome! For now, I have added a translation and a rotation argument for offsetting the text from the point coordinate. Currently, it applies the same translation and rotation for all text elements in the layer. Would this meet your needs? It looks like vispy currently doesn't support specifying a unique offset/rotation for each text element (see [here](http://vispy.org/visuals.html#vispy.visuals.TextVisual)). It would be possible to see about contributing that feature upstream to vispy in the future, but I wouldn't be able to do it in this PR.

Not sure what is meant by 'translation' but I assume/guess that is the position of text in relation to the point coordinate? Ah and yes, applying translation/position & rotation to all text elements is what I meant (and not specific text elements) - suits my needs perfectly. What will the default position offset be, if any?

kevinyamauchi:
No worries, @pablooriol2 !

> Not sure what is meant by 'translation' but I assume/guess that is the position of text in relation to the point coordinate? Ah and yes, applying translation/position & rotation to all text elements is what I meant (and not specific text elements) - suits my needs perfectly. What will the default position offset be, if any?

You guessed right, the translation is the offset from the point coordinate. Currently the default offset is 0 is all dimensions. I could imagine we might handle the default position differently when we integrate text into `Shapes`. For example, we might make the "origin" (i.e., the point the text go when the offset is 0 in all dimensions) settable to a few different options (e.g., upper left corner of the shape, centroid of the shape), as we want to support annotation use cases like this:

https://en.wikipedia.org/wiki/Object_detection#/media/File:Detected-with-YOLO--Schreibtisch-mit-Objekten.jpg

kevinyamauchi:
Thanks to @floryst, we now have text in `Points` and `Shapes`! There are still some tests to clean up, some improvements to default text anchoring relative to the shapes, and some GUI improvements to make. @sofroniewn, the core of it is there and it would be great to get some feedback on the direction we have taken.

Screenshot 2020-07-12 at 21 37 22
Screenshot 2020-07-12 at 21 50 11

napari/_vispy/vispy_points_layer.py
```diff
@@ -1,10 +1,11 @@
 import numpy as np
-from vispy.scene.visuals import Line, Compound
+from vispy.scene.visuals import Line, Compound, Text
 from .markers import Markers
 from vispy.visuals.transforms import ChainTransform
 
 from .vispy_base_layer import VispyBaseLayer
 from ..utils.colormaps.standardize_color import transform_color
+from .text_utils import update_text
```
sofroniewn: I love this approach and how clean `vispy_points_layer.py` and `vispy_shapes_layer.py` is!!!

napari/_vispy/text_utils.py
```diff
@@ -0,0 +1,31 @@
+import numpy as np
+
+
+def update_text(layer, text_node):
```
sofroniewn: I'd probably prefer this function to just take the `layer._text`, `layer._indices_view` , and `layer.dims.ndisplay` variables, it just feels a little safer, more modular, so function signature would be 

```python
def update_text(text, coordinates, ndisplay, text_node):
```
Does that sound reasonable?

napari/layers/utils/text.py
```diff
@@ -0,0 +1,192 @@
...
+
+import numpy as np
+
+from ._text_constants import TextMode
+from .text_utils import format_text_direct, format_text_properties
+from ...utils.colormaps.standardize_color import transform_color
+from ...utils.event import EmitterGroup, Event
+
+
+class TextManager:
```
sofroniewn, line 16: This is looking good - I'm thinking now about how much simpler this class will become if we go with the approach in #1436!

napari/layers/points/points.py
```diff
@@ -1288,6 +1321,30 @@ def _view_data(self) -> np.ndarray:
...
+
+    @property
+    def _view_text_coords(self) -> np.ndarray:
+        if self._text._mode in [TextMode.FORMATTED, TextMode.PROPERTY]:
+            text_coords = (
+                get_text_anchors(self._view_data) + self._text.translation
+            )
+        else:
+            text_coords = np.zeros((0, self.dims.ndisplay))
+        return text_coords
```
sofroniewn: Why don't these attributes live inside the text manager? Could they instead be methods on the text manager, like `slice_text()` with the necessary args?

sofroniewn (commented):
This looks like it is moving in the right direction - I really like the `TextManager` approach and I really like the `update_text` method for doing the vispy update. Left some minor questions in a couple places. Great work!! The component design + update function is much easier to understand

napari/layers/utils/text.py
kevinyamauchi: > I'm thinking now about how much simpler this class will become if we go with the approach in #1436!

For sure. I think the majority of the typing I did for `TextManager` was adding the property getters/setters.

napari/layers/points/points.py
kevinyamauchi: We can move all of this into a function in `TextManager`. We will still have to pass in `layer._view_data`, but I think that is inline with the current implementation.

napari/_vispy/text_utils.py
kevinyamauchi: Yeah, I think that's totally reasonable!

kevinyamauchi: I definitely agree that we should limit passing the full layer model around as much as possible.

napari/layers/points/points.py
kevinyamauchi: Done!

napari/layers/utils/text.py
kevinyamauchi: Done!

napari/_vispy/text_utils.py
kevinyamauchi: Done!

kevinyamauchi:
Thanks for the comments, @sofroniewn . I have addressed them and also added some more options for the text anchors (center and the 4 corners of the bounding box). I have updated the [`add_shapes_with_text.py`](https://github.com/napari/napari/blob/c6713fcba7c8805e58658b4b5459f42ddeb2ccc6/examples/add_shapes_with_text.py) example to give an idea of how we can use the text and colors APIs together (screenshot below).

Screenshot 2020-07-14 at 17 06 44

In terms of wrapping this up, I just need to add a bit of code to make sure we're handling nD correctly and add few more tests/docstrings.

kevinyamauchi:
I think this is ready for review! I have made a companion PR (https://github.com/napari/tutorials/pull/80) with an example demonstrating the usage of the text with the Shapes layer (screenshot of the result below). I have added the code for the tutorial in this PR (`examples/annotate_segmentation_with_text.py`)

annotated_bbox

In particular, I am curious about what @napari/core-devs think about the following:

- should we expose all of the text display properties (e.g., size, color) on the layer model as pass through properties?
- how do we want to treat the text anchoring in 3D? In 2D we have nice options for anchoring to the corners of a shapes bounding box. in 3D, it is less clear to me how we should make that work (e.g., where is "upper_left"?). I think the simplest option is to fall back to centering in the bounding box in 3D. What do you all think?
- where should the text-related files live?
    - currently, the text model and related utils are in `/napari/layers/utils/`
    - the vispy-related text utils are in `/napari/_vispy/text_utils.py`. They can't go in `/napari/_vispy/utils.py` because it creates a circular import due to the `layer_to_visual` dictionary.

sofroniewn:
### Usability testing observations 

- First off this is so incredible and I love it!!

- We could add a None for Points shape that would allow you to see text with no marker (could be a good first issue after this PR, just want to note the idea).

- [ ] Something funny happening with toggle ndisplay 2 to 3 back to 2 in `add_points_with_text.py` probably around z-order.
Also see if you make image data `np.ones` and not `np.zeros` the points text is "behind" the image. See
Screen Shot 2020-07-15 at 2 33 14 PM

- Funny things happen to the text position if you rotate the bounding box. I think it's ok, fine to merge now, but in follow-up could make text rotate too. Not a big deal, but just want to note.
Screen Shot 2020-07-15 at 2 35 42 PM

- Side comment, it looks like when selecting shapes the "edge color" visible in the swatch doesn't change, I though it was supposed to - can fix in followup, it seems unrelated to this PR. 

Probably only bug that needs to be fixed is the z-ordering one, others are just observations.

### Addressing your comments
 
> should we expose all of the text display properties (e.g., size, color) on the layer model as pass through properties?

If I understand correctly, then no I don't think so, I need to review code now, but I think we want `layer.text.color = 'red'` rather than adding something like `layer.text_color = 'red'`. We need to think about this stuff more, but to me that API change is a huge reason for the `TextManager` class approach. I think our next job after this will be a `ColorManager` component too or a `RenderingManager` component, but we can discuss on another issue.

>  I think the simplest option is to fall back to centering in the bounding box in 3D. What do you all think?

I think that's fine for now, easy to improve with concrete use cases.

> where should the text-related files live?

can we make it `_text_utils.py` so private. Otherwise choices look go.

I will go through code now.

napari/_vispy/text_utils.py
```diff
@@ -0,0 +1,35 @@
+import numpy as np
+from vispy.scene.visuals import Text
+
+
+def update_text(
+    text_values: np.ndarray,
+    text_coords: np.ndarray,
+    anchor_x: str,
+    anchor_y: str,
```
sofroniewn: Can this be `anchor` a 2-tuple of str?

napari/_vispy/text_utils.py
```diff
@@ -0,0 +1,35 @@
+import numpy as np
+from vispy.scene.visuals import Text
+
+
+def update_text(
```
sofroniewn: I'd probably drop the `text_` from the kwards like `text_color` and just make them `color` as function is called `update_text` it's probably unambigious what's going on.

Also needs doc string.

napari/layers/utils/text.py
```diff
@@ -0,0 +1,267 @@
...
+
+    @property
+    def text(self):
+        """np.ndarray: the text values to be displayed"""
+        return self._text
+
+    # @text.setter
+    # def text(self, text):
+    #
+    #     self._text = text
```
sofroniewn: Why is this commented?

sofroniewn (approved):
This is really incredible @kevinyamauchi it'll be so great to get this in. I've just gone through the code and think it's basically there. Maybe @tlambert03 and @jni can go through it too. There's one z-ordering bug I found, I think everything else can be improved on after we get people using it. I'd like to get this into the 0.3.6 release that's coming right up!!

kevinyamauchi:
Thanks for the review and comments, @sofroniewn !

> Funny things happen to the text position if you rotate the bounding box. I think it's ok, fine to merge now, but in follow-up could make text rotate too. Not a big deal, but just want to note.

For better or for worse, that's the intended behavior. For the corner anchoring options (e.g., `Anchor.UPPER_LEFT`), we're just calculating the bounding box around all points in the polygon and anchoring the text to the appropriate corner of that bounding box (upper left in the case of `Anchor.UPPER_LEFT`). My rationale is that would be the most predictable thing for polygons that don't have an obvious "upper left" corner. The side effect is that when shapes like rectangles have weird placement when they are rotated such that the major/minor axes don't align with the vertical/horizontal axes of the canvas. I tried to design the anchoring calculation in a way that's reasonably expandable, so I think we can add different flavors fairly easily in future PRs. In short, one has to: (1) create a new [`Anchor` enum](https://github.com/napari/napari/blob/501fa2f1a503cf8ee105eecd86c54cc8f909b919/napari/layers/utils/_text_constants.py#L22) for their new anchoring mode, (2) create a function that transforms the `view_data` from the model into an array of anchor points, and (3) register the function in the [`TEXT_ANCHOR_CALCULATION`](https://github.com/napari/napari/blob/501fa2f1a503cf8ee105eecd86c54cc8f909b919/napari/layers/utils/text_utils.py#L116) dict.

This discussion does raise the question of the `TextManager` should know what type of annotation the text is attached to. I leaned against this because I think it would lead to a lot of possible states for the `TextManager` given how many different types of annotations can now have properties. My vote is we start simple and wait to see how many people need fancy anchoring options. It is also worth noting that the `TextManager.translation` value can also be an NxD array so the user can specify their own offset for each text element. That could be a potential handle for users that have special text placement needs.

> Side comment, it looks like when selecting shapes the "edge color" visible in the swatch doesn't change, I though it was supposed to - can fix in followup, it seems unrelated to this PR.

I am guessing an event connection is missing. I think a separate PR makes sense. Thanks for noting that. I will make an issue.

> Something funny happening with toggle ndisplay 2 to 3 back to 2 in add_points_with_text.py probably around z-order.
Also see if you make image data np.ones and not np.zeros the points text is "behind" the image. 

Hmm strange. I will look into this.

> We could add a None for Points shape that would allow you to see text with no marker (could be a good first issue after this PR, just want to note the idea).

I agree that we need an API for text only. I am not sure about the `None` argument for `Points` as that doesn't feel very discoverable/intuitive. I keep going back to thinking we should have a standalone `Text` layer. What if we exposed an option to make the points markers visible separately from the layer? Then it would be something like 

```python
viewer.add_points(data, properties=properties, text=text, markers_visible=False)
```

kevinyamauchi:
> If I understand correctly, then no I don't think so, I need to review code now, but I think we want layer.text.color = 'red' rather than adding something like layer.text_color = 'red'. We need to think about this stuff more, but to me that API change is a huge reason for the TextManager class approach. I think our next job after this will be a ColorManager component too or a RenderingManager component, but we can discuss on another issue.

I agree with this. My follow up question is, should `Points.text` return the text values (as it is now) or the `TextManager` object. It needs to return the `TextManager` object if we want to be able to do `Points.text.color='green'`. Currently, since we are returning the array of text values, one needs to do `Points._text.color='green'`.

>> I think the simplest option is to fall back to centering in the bounding box in 3D. What do you all think?

> I think that's fine for now, easy to improve with concrete use cases.

I agree. I will implement the 3D centered anchor point fallback.

> can we make it _text_utils.py so private. Otherwise choices look go.

Do you mean for both the `text_utils.py` in the `_vispy` and `layers.utils` modules? I think it makes sense for both.

napari/layers/utils/text.py
kevinyamauchi: Ah I was debating if we want to have a `TextManager.text` setter method. It feels a bit weird because we don't just directly pass text values. If we had one, I think we'd have to accept a few different types:

```python
# keep the same text pattern, just update values based on new properties
points.layer.text.text = new_properties

# same properties, new text pattern
new_text_pattern = 'new message, same {property_name}'
point.layer.text.text = new_text_pattern

# new properties, new text pattern
points.layer.text.text = (new_properties, new_text_pattern)
```

Currently, the text setting is taken care of on the layer model using the existing method for setting text values ([`TextManager._set_text()`](https://github.com/napari/napari/blob/501fa2f1a503cf8ee105eecd86c54cc8f909b919/napari/layers/utils/text.py#L101)).

All of that is to say, I would like to delete the `TextManager.text` setter. 😛

napari/_vispy/text_utils.py
kevinyamauchi: I updated the signature and added a docstring.

napari/_vispy/text_utils.py
kevinyamauchi: I tend to prefer explicit kwargs that take a single value since I think it is simpler to remember (e.g., one doesn't have to remember what each element in the tuple is). In this particular case, it's only used internally and we would always provide both values, so I don't have a strong preference. I went ahead and made the change!

kevinyamauchi:
>  Something funny happening with toggle ndisplay 2 to 3 back to 2 in add_points_with_text.py probably around z-order.
Also see if you make image data np.ones and not np.zeros the points text is "behind" the image. See

Hey @sofroniewn , I was just looking into the 2d->3d->2d z ordering bug. I was able to reproduce it with `add_points.py` on master (the points end up behind the image). Do you have the same issue? This might be orthogonal to this PR.

sofroniewn:
> Hey @sofroniewn , I was just looking into the 2d->3d->2d z ordering bug. I was able to reproduce it with add_points.py on master (the points end up behind the image). Do you have the same issue? This might be orthogonal to this PR.

Yup, just reproduced on master. I can make a new issue

[Edit: Done in #1462]

sofroniewn:
> Do you mean for both the text_utils.py in the _vispy and layers.utils modules? I think it makes sense for both.

Private for both sounds good

sofroniewn:
> > Funny things happen to the text position if you rotate the bounding box. I think it's ok, fine to merge now, but in follow-up could make text rotate too. Not a big deal, but just want to note.
>
> For better or for worse, that's the intended behavior. 

Thats fine, thanks for the detailed explanation - seems like a good choice for now

> This discussion does raise the question of the TextManager should know what type of annotation the text is attached to.

I don't want it too - ideally not.


> > Side comment, it looks like when selecting shapes the "edge color" visible in the swatch doesn't change, I though it was supposed to - can fix in followup, it seems unrelated to this PR.
>
> I am guessing an event connection is missing. I think a separate PR makes sense. Thanks for noting that. I will make an issue.

Thanks!!

sofroniewn:
> My follow up question is, should Points.text return the text values (as it is now) or the TextManager object. It needs to return the TextManager object if we want to be able to do Points.text.color='green'. Currently, since we are returning the array of text values, one needs to do Points._text.color='green'.

Ah tricky, I get it now - I like it `Points.text` returning the `TextManager` for the API convenience reasons I mentioned above. I'd then do something like `Points.text.coords` or `Points.text.values` to get the coordinates/ strings respectively. I don't like the `Points._text.color='green'` as much as that's not something we'd then be encouraging a user to do, whereas `Points.text.color='green'` feels very natural.

I think if you do `_get_state()` then inside that dictionary you put `text: {....}` where that inner dictionary has all the data needed, so the serialization test still passes, but `Points.text` returns the `TextManager`.

I'd make that change still in this PR, but curious what @jni and @tlambert03 think too?

kevinyamauchi:
> Ah tricky, I get it now - I like it Points.text returning the TextManager for the API convenience reasons I mentioned above. I'd then do something like Points.text.coords or Points.text.values to get the coordinates/ strings respectively. I don't like the Points._text.color='green' as much as that's not something we'd then be encouraging a user to do, whereas Points.text.color='green' feels very natural.

This makes sense to me. I had initially used the current pattern (`layer.text` getter returns the array of text rather than `TextManager`) because I wasn't sure if we were ready to commit to the `TextManager` pattern as part of the public API. However, I think it seems like the way to go, so I am happy to make `layer.text` return the `TextManager` object. A quick question: the text values are currently accessed via `TextManager.text`. Do you prefer `TextManager.values` or was that just a quick example? I don't have a strong preference either way.

Also, in the `TextManager` doesn't store the point coordinates. When the coordinates are required (e.g., rendering), the layer passes the `view_data` to `TextManager.compute_coords()`, which returns the text coordinates (i.e., it applies the translation and determines the anchor point). I went with this approach so we don't have to duplicate the data. I just mention this because `layer.text.coords` wouldn't work unless the model data is passed in (`TextManager` can't directly access layer data, which I think is desirable).

Sorry if I'm reading too much into your comment - I just want to make sure we're on the same page.

> I think if you do _get_state() then inside that dictionary you put text: {....} where that inner dictionary has all the data needed, so the serialization test still passes, but Points.text returns the TextManager.

`TextManager` has it's own `_get_state()` method, so we just call that when serializing the layer. See an example from the Points layer below
https://github.com/napari/napari/blob/2e992bf633c104d1fd6b8003af7c9214aab5e774/napari/layers/points/points.py#L1153-L1172

tlambert03:
> > I like it Points.text returning the TextManager for the API convenience reasons I mentioned above.

> This makes sense to me. I had initially used the current pattern (layer.text getter returns the array of text rather than TextManager) because I wasn't sure if we were ready to commit to the TextManager pattern as part of the public API. However, I think it seems like the way to go, so I am happy to make layer.text return the TextManager object.

fwiw, I agree with the conclusion here!

examples/add_points_with_text.py
```diff
@@ -0,0 +1,49 @@
...
+    properties = {
+        'confidence': np.array([1, 0.5, 0]),
+        'good_point': np.array([True, False, False]),
+    }
+
+    # define the color cycle for the face_color annotation
+    face_color_cycle = ['blue', 'green']
+
+    text = {
+        'text': 'Confidence is {confidence:.2f}',
```
tlambert03, line 28: i love this

tlambert03 (approved):
mostly looked through the example and the general API, but this is really awesome!  will make for some very rich data presentations and I look forward to seeing what people do with it :)

sofroniewn:
> Sorry if I'm reading too much into your comment - I just want to make sure we're on the same page.

No this is all great to discuss, helps me think about it too.

> A quick question: the text values are currently accessed via TextManager.text. Do you prefer TextManager.values or was that just a quick example? I don't have a strong preference either way.

Hmm I'm not sure here `layer.text.text` feels a little funny to me so I guess I prefer `layer.text.values` - I worry `layer.text.strings` is weird too.

> Also, in the TextManager doesn't store the point coordinates. When the coordinates are required (e.g., rendering), the layer passes the view_data to TextManager.compute_coords(), which returns the text coordinates (i.e., it applies the translation and determines the anchor point). I went with this approach so we don't have to duplicate the data. I just mention this because layer.text.coords wouldn't work unless the model data is passed in (TextManager can't directly access layer data, which I think is desirable).

Yup that's fine, that makes sense - good not to have duplication of the `coords` - there should only be one source of ground truth, and we've decided that lives on the `layer` itself!

kevinyamauchi:
I have implemented the change to `layer.text` to return the `TextManager` object. Everything seems to be working, but I am running into the following test issue. I think the issue is we can't test the equality of the `TextManager` objects. Do you think we should add an `__eq__` method to `TextManager`? We could compare the outputs of `TextManager._get_state()`.


```python
====================================================================================== short test summary info ======================================================================================
FAILED napari/layers/_tests/test_serialize.py::test_attrs_arrays[Points-data5-2] - AssertionError: assert False
FAILED napari/layers/_tests/test_serialize.py::test_attrs_arrays[Points-data6-3] - AssertionError: assert False
FAILED napari/layers/_tests/test_serialize.py::test_attrs_arrays[Shapes-data8-2] - AssertionError: assert False
FAILED napari/layers/_tests/test_serialize.py::test_attrs_arrays[Shapes-data9-2] - AssertionError: assert False

```
```python
    @pytest.mark.parametrize('Layer, data, ndim', layer_test_data)
    def test_attrs_arrays(Layer, data, ndim):
        """Test layer attributes and arrays."""
        np.random.seed(0)
        layer = Layer(data)
        # Check layer has been correctly created
        assert layer.ndim == ndim
    
        properties = layer._get_state()
    
        # Check every property is in call signature
        signature = callsignature(Layer)
        for prop in properties.keys():
            assert prop in signature.parameters
    
        # Check number of properties is same as number in signature
        assert len(properties) == len(signature.parameters)
    
        # Check new layer can be created
        new_layer = Layer(**properties)
    
        # Check that new layer matches old on all properties:
        for prop in properties.keys():
            # If lists check equality of all elements with np.all
            if isinstance(getattr(layer, prop), list):
                assert np.all(
                    [
                        np.all(ol == nl)
                        for ol, nl in zip(
                            getattr(layer, prop), getattr(new_layer, prop)
                        )
                    ]
                )
            elif isinstance(getattr(layer, prop), dict):
                assert np.all(
                    [
                        np.all(value == getattr(new_layer, prop)[key])
                        for key, value in getattr(layer, prop).items()
                    ]
                )
            else:
>               assert np.all(getattr(layer, prop) == getattr(new_layer, prop))
E               AssertionError: assert False
E                +  where False = <function all at 0x112d8dd40>(<napari.layers.utils.text.TextManager object at 0x17899db10> == <napari.layers.utils.text.TextManager object at 0x1788ee2d0>)
E                +    where <function all at 0x112d8dd40> = np.all
E                +    and   <napari.layers.utils.text.TextManager object at 0x17899db10> = getattr(<Shapes layer 'Shapes' at 0x1432f1fd0>, 'text')
E                +    and   <napari.layers.utils.text.TextManager object at 0x1788ee2d0> = getattr(<Shapes layer 'Shapes' at 0x17899df90>, 'text')

napari/layers/_tests/test_serialize.py:48: AssertionError
```

sofroniewn:
>  but I am running into the following test issue. I think the issue is we can't test the equality of the TextManager objects. Do you think we should add an __eq__ method to TextManager? We could compare the outputs of TextManager._get_state().

I think that would be nice, I think @tlambert03 has considered things like that in various places in the codebase, see #850 which we might want to finish up at some point.

kevinyamauchi:
> I think that would be nice, I think @tlambert03 has considered things like that in various places in the codebase, see #850 which we might want to finish up at some point.

I implemented a less fancy version that I basically copied from the serialization test (I had finished it by the time I saw your comment). I am happy to implement @tlambert03 's if we think hashing first has advantages. I suppose it can also be a follow-up PR. Here's what I did:
https://github.com/napari/napari/blob/37e8df9f68cc840b388794e55de202d8b40c89e9/napari/layers/utils/text.py#L252-L266

kevinyamauchi:
I have addressed @sofroniewn 's comments by exposing the `TextManager` object via the `layer.text` property. Also, I added a refresh method to text (exposed through `layer.refresh_text()`) so users can update the text if they modify the properties without using the properties setter (it is automatically called via the `layer.properties` setter).

I think this is feature complete for this first round. I am signing off for the night, but I can address any lingering comments in the morning. It would be great if we could also finish up the companion tutorial (https://github.com/napari/tutorials/pull/80) by when we cut the final release.

If you do end up merging it, can you please add a co-author tag on the squash+commit for @floryst ? We worked together on this a fair amount during the scipy sprint and somehow our commits during that time weren't tagged properly by vscode. https://github.blog/2018-01-29-commit-together-with-co-authors/

examples/add_shapes_with_text.py
```diff
@@ -0,0 +1,55 @@
...
+    }
+    edge_color_cycle = ['blue', 'magenta', 'green']
+
+    text_properties = {
+        'text': '{class}: {likelihood:0.1f}%',
+        'anchor': 'upper_left',
+        'translation': [-5, 0],
+        'size': 8,
+        'color': 'green',
+    }
```
jni, line 36: Playing around with this example, I was disappointed that I couldn't delete everything except for the text and have it still be functional. I'd like it if there were reasonable defaults for everything and `text='text_property_name'`, `text='{arbitrary_property_name}'`, or `text={'text': '{arbitrary_property_name}'` would work. Currently, no errors or warnings are thrown, but I see no text.

jni: It looks like 'color' is the missing default argument.

examples/annotate_segmentation_with_text.py
```diff
@@ -0,0 +1,130 @@
...
+labels = [reg.label for reg in regions]
+circularity = [
+    calculate_circularity(reg.perimeter, reg.area) for reg in regions
+]
+
+# create the properties dictionary
+properties = {
+    'label': labels,
+    'circularity': circularity,
+}
```
jni: See [this comment](https://github.com/napari/tutorials/pull/80/files#r456738023) from the tutorials PR

examples/nD_shapes_with_text.py
```diff
@@ -0,0 +1,25 @@
...
+from skimage import data
+import napari
+
+
+with napari.gui_qt():
+    blobs = data.binary_blobs(
+        length=100, blob_size_fraction=0.05, n_dim=3, volume_fraction=0.03
+    ).astype(float)
+
+    viewer = napari.view_image(blobs.astype(float))
```
jni:
```suggestion
    viewer = napari.view_image(blobs.astype(float), ndisplay=3)
```

napari/_vispy/_text_utils.py
```diff
@@ -0,0 +1,59 @@
...
+    text_values: np.ndarray,
+    coords: np.ndarray,
+    anchor: Tuple[str, str],
+    rotation: float,
+    color: np.ndarray,
+    size: float,
+    ndisplay: int,
+    text_node: Text,
+):
+    """ Update the vispy text node with the current text and display parameters
```
jni:
```suggestion
    """Update the vispy text node with the current text and display parameters.
```

@Carreau, help us! 😂

napari/_vispy/_text_utils.py
```diff
@@ -0,0 +1,59 @@
...
+    text_values : np.ndarray
+        The array of text strings to display.
+    coords : np.ndarray
+        The coordinates for each text element.
+    anchor : Tuple[str, str]
+        The name of the vispy anchor positions provided as (anchor_x, anchor_y).
+        anchor_x should be one of: 'left', 'center', 'right'.
+        anchor_y should be one of: 'top', 'center', 'middle', 'baseline', 'bottom'.
+    rotation : float
+        The rotation (degrees) of the text element around it's coordinate.
```
jni:
```suggestion
        The rotation (degrees) of the text element around its anchor.
```
?

napari/layers/points/points.py
```diff
@@ -297,6 +311,17 @@ def __init__(
             }
             self._properties = empty_properties
 
+        # make the text
+        if type(text) in (list, np.ndarray, str) or text is None:
```
```suggestion
        if text is None or isinstance(text, [list, np.ndarray, str]):
```

(I've changed the order because `is None` will be by far the most common branch point and we might as well shortcut it!)

napari/layers/points/points.py
```diff
@@ -597,6 +627,23 @@ def _validate_properties(
...
+
+    @text.setter
+    def text(self, text):
+        self._text._set_text(
+            text, n_text=len(self.data), properties=self.properties
+        )
+
+    def refresh_text(self):
+        """Refresh the text values.
+        This is generally used if the properties were updated without changing the data
```
jni, line 644:
```suggestion

        This is generally used if the properties were updated without changing the data
```

napari/layers/points/points.py
```diff
@@ -1287,6 +1335,30 @@ def _view_data(self) -> np.ndarray:
...
         return data
 
+    @property
+    def _view_text(self) -> np.ndarray:
+        """Get the values of the text elements in view
+
+        Returns
+       -------
+       text : (N x 1) np.ndarray
+           Array of text strings for the N text elements in view
```
```suggestion
        -------
        text : (N x 1) np.ndarray
            Array of text strings for the N text elements in view
```

napari/layers/points/points.py
```diff
@@ -1287,6 +1335,30 @@ def _view_data(self) -> np.ndarray:
...
+        return self.text.view_text(self._indices_view)
+
+    @property
+    def _view_text_coords(self) -> np.ndarray:
+        """Get the coordinates of the text elements in view
+
+        Returns
+       -------
+       text_coords : (N x D) np.ndarray
+           Array of coordindates for the N text elements in view
```
```suggestion
        -------
        text_coords : (N x D) np.ndarray
            Array of coordindates for the N text elements in view
```

napari/layers/shapes/shapes.py
```diff
@@ -1195,6 +1231,34 @@ def _get_state(self):
...
+        return np.where(self._data_view._displayed)[0]
+
+    @property
+    def _view_text(self) -> np.ndarray:
+        """Get the values of the text elements in view
+
+        Returns
+       -------
+       text : (N x 1) np.ndarray
+           Array of text strings for the N text elements in view
```
jni:
```suggestion
        -------
        text : (N x 1) np.ndarray
            Array of text strings for the N text elements in view
```

How did these even happen so consistently??? 😂

napari/layers/shapes/shapes.py
```diff
@@ -1195,6 +1231,34 @@ def _get_state(self):
...
+        return self.text.view_text(self._indices_view)
+
+    @property
+    def _view_text_coords(self) -> np.ndarray:
+        """Get the coordinates of the text elements in view
+
+        Returns
+       -------
+       text_coords : (N x D) np.ndarray
+           Array of coordindates for the N text elements in view
```
jni:
```suggestion
        -------
        text_coords : (N x D) np.ndarray
            Array of coordindates for the N text elements in view
```

napari/layers/shapes/shapes.py
```diff
@@ -1622,6 +1688,23 @@ def _validate_properties(
...
+
+    @text.setter
+    def text(self, text):
+        self._text._set_text(
+            text, n_text=len(self.data), properties=self.properties
+        )
+
+    def refresh_text(self):
+        """Refresh the text values.
+        This is generally used if the properties were updated without changing the data
```
jni, line 1705:
```suggestion

        This is generally used if the properties were updated without changing the data
```

This doesn't feel like it should be a public method, should it?

napari/layers/utils/_text_constants.py
```diff
@@ -0,0 +1,38 @@
...
+
+class TextMode(StringEnum):
+    """
+    TextMode: Text setting mode.
+
+    NONE (default mode) no text is displayed
+
+    PROPERTY the text value is a property value
+
+    FORMATTED allows text to be set with an f string like syntax
```
jni:
```suggestion
    FORMATTED allows text to be set with an f-string like syntax
```

napari/layers/utils/_text_utils.py
```diff
@@ -0,0 +1,225 @@
...
+) -> Tuple[np.ndarray, str, str]:
+
+    if ndisplay == 2:
+        bbox_min, bbox_max = _calculate_bbox_extents(view_data)
+        text_anchors = np.array([bbox_min[:, 0], bbox_min[:, 1]]).T
+
+        anchor_x = 'left'
+        anchor_y = 'top'
+    else:
+        # in 3D, use centered anchor
```
jni, line 59: Is this necessarily what we want?

jni: To rephrase... I think if I had 3D points, and I asked to bottom-left-anchor my text, I'd expect it to also work in 3D... Is there a fundamental reason that we don't (e.g. VisPy makes it hard)?

napari/layers/utils/text.py
```diff
@@ -0,0 +1,328 @@
...
+from ._text_utils import (
+    format_text_properties,
+    get_text_anchors,
+)
+from ...utils.colormaps.standardize_color import transform_color
+from ...utils.event import EmitterGroup, Event
+
+
+class TextManager:
+    """ Manages properties related to text displayed in conjunction with the layer
```
jni:
```suggestion
    """Manages properties related to text displayed in conjunction with the layer.
```

napari/layers/utils/text.py
```diff
@@ -0,0 +1,328 @@
...
+
+        Parameters
+        ----------
+        indices_view : (N x 1) np.ndarray
+            Indices of the text elements in view
+
+        Returns
+       -------
+       text : (N x 1) np.ndarray
+           Array of text strings for the N text elements in view
```
jni:
```suggestion
        -------
        text : (N x 1) np.ndarray
            Array of text strings for the N text elements in view
```

napari/layers/utils/text.py
```diff
@@ -0,0 +1,328 @@
...
+
+    def __init__(
+        self,
+        text,
+        n_text,
+        properties={},
+        rotation=0,
+        translation=0,
+        anchor='center',
+        color='black',
```
jni: This will be controversial, but I'm thinking that maybe we should consider using cyan as the default text colour. I'm finding that cyan on white looks reasonable, but black on black (most fluorescent images) is unusable.

napari/layers/utils/text.py
```diff
@@ -0,0 +1,328 @@
...
+            'translation': self.translation,
+            'size': self.size,
+            'visible': self.visible,
+        }
+
+        return state
+
+    def _connect_update_events(self, update_function):
+        """Function to connect all property update events to the update callback.
+        This is typically used in the vispy view file.
```
jni, line 347:
```suggestion

        This is typically used in the vispy view file.
```

napari/layers/utils/text.py
```diff
@@ -0,0 +1,328 @@
...
+                    for key, value in my_state.items()
+                ]
+            )
+
+        else:
+            equal = False
+
+        return equal
+
+    def ne(self, other):
```
jni: Clearly this should not be public?

jni: Should this be `__ne__`?

jni (commented):
@kevinyamauchi @floryst what a tour de force!

I love this thing. I really like the API, how it's super flexible, and in particular, how everything but the text is intended to be optional, which means you have a simple API but hidden power if you want to do more sophisticate things — like you do in the examples, which are excellent.

I've made mostly minor comments in the code. However, there's a major issue that I think needs to be fixed before merge.

In the camera example at least, if I remove all attributes from the dictionary except the text, the text is not shown. At first I thought it was to do with not having a default colour, but in fact we do have one, black. Weirdly, if I change the default to cyan, it works fine. But, then I was testing whether cyan on white looks ok by adding `viewer.add_image(np.ones((512, 512), dtype=float), name='blank')` immediately after adding camera. This had the effect shown below, which makes me think that what's happening is that the blending *on the text specifically* is additive rather than translucent. That would explain why black doesn't show up at all, and why the cyan text goes white when it's over the pink shape.

I haven't had time to track down the blending issue, but once that's fixed we can move forward! My suggestion of cyan as a default is a weak one, but I do think it'll be more useful than black.

Screen Shot 2020-07-18 at 3 03 43 pm

napari/_vispy/vispy_points_layer.py
```diff
@@ -123,3 +128,54 @@ def _on_highlight_change(self, event=None):
...
+        return text_node
+
+    def _on_blending_change(self, event=None):
+        """Function to set the blending mode"""
+        self.node.set_gl_state(self.layer.blending)
+
+        # the text blending mode should always be additive
+        # see: https://github.com/napari/napari/pull/600#issuecomment-554142225
+        text_node = self._get_text_node()
+        text_node.set_gl_state('additive')
```
jni: This bit!!!!

... Aaaand I just saw that the comment linked above was mine! 😂 BUT, I'll emphasise one thing I said in the linked comment:

> I'm not entirely sure of all the consequences, though...

Now we know!!! What do people think? I mean, certainly default black text with additive blending is the worst of all worlds. 😂

jni:
Ok, the world is sane, by changing the blending to the main node blending, it works fine (though it does hurt my eyes a bit):

 
Screen Shot 2020-07-18 at 3 29 52 pm

Not sure what to do about this... `layer.text.blending=`? With a default of... translucent? matching the layer?

napari/_vispy/vispy_points_layer.py
kevinyamauchi: Ah, so you have met past Juan ;) I agree the default of black was a bad choice on my part. I think we have a couple of options:

1. we can add an blending parameter (with default) to `TextManager` that is set independently of the layer blending mode.

2. we can keep the blending mode as always additive, but change the default to something more sensible (e.g., cyan as you suggested).

I am leaning towards option (1) since it will be functionally the same as option (2) for the vast majority of users, but give some flexibility for "power users".

kevinyamauchi:
> Not sure what to do about this... layer.text.blending=? With a default of... translucent? matching the layer?

Haha, [I think we converged to the same solution](https://github.com/napari/napari/pull/1374#discussion_r456752694). I will add a `TextManager.blending` property with a translucent default.

napari/_vispy/_text_utils.py
kevinyamauchi: I am really looking forward to your docstring formatter, @Carreau! Until then, sorry @jni - it's not intentional!

napari/layers/shapes/shapes.py
kevinyamauchi: I have no idea! I just press tab!

napari/layers/utils/text.py
kevinyamauchi: Yikes - yes! Fixed.

napari/_vispy/vispy_points_layer.py
kevinyamauchi: To address this I have:

- added `TextManager.blending`
- `TextManager.blending` cannot be set to opaque. If the user attempts to set it to opaque, napari emits a warning and instead falls back to translucent.

napari/layers/utils/text.py
kevinyamauchi: I agree that black is definitely not the right choice. It's cyan now. We can revise it later if there is an uprising 😛

examples/add_shapes_with_text.py
kevinyamauchi: I believe the underlying issue is that the default color was black and thus the text couldn't be seen. The default color is now cyan and the default blending mode is now translucent. Please let me know if I have misunderstood.

jni (approved):
Awesome! 🎉

jni:
Dang, sorry, my bad, the second argument to isinstance must be a tuple, not a list! I think that should fix the CI @kevinyamauchi!

kevinyamauchi:
> Dang, sorry, my bad, the second argument to isinstance must be a tuple, not a list! I think that should fix the CI @kevinyamauchi!

All good! Fixed!

examples/annotate_segmentation_with_text.py
kevinyamauchi: @jni I have updated the example here. I will mirror the change in the tutorials PR soon.

I have to say, I love that we can pass the regionprops_table directly in as layer properties.

synergy

kevinyamauchi:
I have updated the example to use the skimage `regionprops_table`, as suggested by @jni. We can let this sit for a bit longer. If there are no further comments, I think this is good to go!

Let's not forget to add the co-author tag for @floryst in the final squash+merge!

jni (approved):

jni:
Woo! It worked! https://github.com/napari/napari/commit/4b133241babe179a1384d1289fa147bc6a7e2373 =)

napari/layers/utils/text.py
sofroniewn: > I agree that black is definitely not the right choice. It's cyan now. We can revise it later if there is an uprising 😛

I would consider our hightlight color - which I find a bit easier on the eyes then cyan. One downside though is it is the highlight color and so that might be confusing too .... cyan is fine for now

sofroniewn:
Amazing work @kevinyamauchi @floryst @jni so great to see this in!! Very excited for the `0.3.6` release which I will start preparing now.

napari/_vispy/_text_utils.py
Carreau: you can run it here: https://github.com/carreau/minirst still highly in progress.

kevinyamauchi:
Thanks for all of the hard work, everyone! It feels pretty good to close #600 and #593!

FYI, @pablooriol2 , text will be added in our next (0.3.6) release. I am looking forward to any feedback you may have!

kevinyamauchi:
> Woo! It worked! 4b13324 =)

Awesome, thanks @jni !

tlambert03:
too late for this PR... and sorry I didn't check this sooner. but I'm playing with implementing the `TextManager` as an [evented dataclass](https://github.com/napari/napari/pull/1475) and noticed a few type inconsistencies in the docstring and the signature:
- `n_text` is a required positional argument, but not mentioned in the docstring.
- `properties` isn't mentioned in the docstring, so the reader has to search the code for an example instantiation to see an example of what it might hold 
- docstrings say `translation` should be an `np.ndarray` but the default value is `0`.
- sometimes it says `array`, other times it says `np.ndarray`

side note: it would be a lot harder to have these inconsistencies if we had #1097 ;) so just wanna bump that up on people's radar's again

pablooriol2:
> Thanks for all of the hard work, everyone! It feels pretty good to close #600 and #593!
> 
> FYI, @pablooriol2 , text will be added in our next (0.3.6) release. I am looking forward to any feedback you may have!

That's awesome @kevinyamauchi , it's looking great! Will try it as soon as I update :-D

PR #1375, opened 2020-06-20
Fix two remaining tests that try to show the viewer.

tlambert03:
# Description
this hides the GUI in two tests (testing the `napari.view_*` methods) that weren't caught in #1372.  

There is one remaining thing that is slightly annoying that I don't think we can do anything about:  If you tab away from the terminal right after starting `pytest`, then right at the moment when the first `qtbot` gets instantiated your current app will switch to the `python 3.x` app running the qt application.  (This is the same thing that happens when running the `%gui qt` magic in ipython ... so it's a general Qt thing, not testing specific).  You can just tab back and keep working, but just fyi...

## Type of change

- [x] refactor tests

### DIFF
--- a/napari/_tests/test_napari.py
+++ b/napari/_tests/test_napari.py
@@ -23,7 +23,7 @@ def test_view_multichannel(qtbot):
     """Test adding image."""
     np.random.seed(0)
     data = np.random.random((15, 10, 5))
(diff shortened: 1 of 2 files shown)


sofroniewn (approved):
This works great, much nicer experience when running the tests now! Will merge tomorrow morning

Issue #1347, opened 2020-06-09
Napari 0.3.4 release crashes when switching to volume rendering


VolkerH:
## 🐛 Bug

Napari 0.3.4 (release version) crashes when trying to render volume in 3D.

## To Reproduce

1. Load a volume
2. Display and navigate through slices in 2D
3. Switch to 3D -> Crash

![napari_vol_crash](https://user-images.githubusercontent.com/2210044/84094896-3232af00-aa41-11ea-8477-87c4d5257e21.gif)

```
(stardist) λ napari
WARNING: Error drawing visual <Volume at 0x1b6364c0d48>
10:17:55 WARNING Error drawing visual <Volume at 0x1b6364c0d48>
WARNING: Traceback (most recent call last):
  File "c:\users\volker\anaconda3\envs\stardist\lib\site-packages\OpenGL\latebind.py", line 43, in __call__
    return self._finalCall( *args, **named )
TypeError: 'NoneType' object is not callable

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "c:\users\volker\anaconda3\envs\stardist\lib\site-packages\vispy\app\backends\_qt.py", line 825, in paintGL
    self._vispy_canvas.events.draw(region=None)
  File "c:\users\volker\anaconda3\envs\stardist\lib\site-packages\vispy\util\event.py", line 455, in __call__
    self._invoke_callback(cb, event)
  File "c:\users\volker\anaconda3\envs\stardist\lib\site-packages\vispy\util\event.py", line 475, in _invoke_callback
    self, cb_event=(cb, event))
  File "c:\users\volker\anaconda3\envs\stardist\lib\site-packages\vispy\util\event.py", line 471, in _invoke_callback
    cb(event)
  File "c:\users\volker\anaconda3\envs\stardist\lib\site-packages\vispy\scene\canvas.py", line 217, in on_draw

    self._draw_scene()
  File "c:\users\volker\anaconda3\envs\stardist\lib\site-packages\vispy\scene\canvas.py", line 266, in _draw_scene
    self.draw_visual(self.scene)
  File "c:\users\volker\anaconda3\envs\stardist\lib\site-packages\vispy\scene\canvas.py", line 304, in draw_visual
    node.draw()
  File "c:\users\volker\anaconda3\envs\stardist\lib\site-packages\vispy\scene\visuals.py", line 99, in draw
    self._visual_superclass.draw(self)
[...]

Issue #1091, opened 2020-03-27
Discuss policy on camelCase in QtSubclasses, then add to docs


tlambert03:
## 📚 Documentation
@jni asked in  https://github.com/napari/napari/pull/1023#discussion_r399197252
>Do we have have a policy on how we deal with camelCase vs under_scores when adding functions to Qt subclasses? Whatever it is, we should put it in the contributing guide.

I don't think there is a formal policy... but we can discuss what it might be here and then add it to contributing.md

sofroniewn:
See also #6 our oldest open issue from @kne42!!! I'd love to resolve this, standardize the repo and update the contributing guidelines. I don't come into this with strong opinions, so if anyone wants to put out a proposal that would be great!!!

tlambert03:
interesting!  I also don't really have strong opinions either way, though my habit had been to follow #6.  It was my impression though that @jni had the opposite preference to what @kne42  was saying there in #6 (i.e. our stuff should always be pythonic) ... but I could have misinterpreted that.

tlambert03:
Would be nice to get something like a show of hands on this @napari/core-devs:

1. Stick with camelCase for method names in Qt Subclasses. (see [`QRangeSlider`](https://github.com/napari/napari/blob/master/napari/_qt/qt_range_slider.py#L13) for an example)
2. Always prefer snake_case for our own method names (see [`QtLayerList`](https://github.com/napari/napari/blob/master/napari/_qt/qt_layerlist.py#L19) for an example)
   - downside: because some of these subclasses will be reimplementing Qt methods, this will inevitably lead to the case where we have a mix of camelCase (for reimplemented methods) and snake_case (for our added methods).  But we could probably create a test that says "if the superclass doesn't have this method, make sure it's snake case)
   - upside: it's more pythonic... and in #1376 we are about to rely on the assumption that event handler methods are named `_on_something_change`... so we're more or less demanding at least partial snake_case names.

If people want to just reply "`#1`" or "`#2`", I can maybe start on writing a test and standardizing

sofroniewn:
Yeah, the #1376 case is a little weird, otherwise I would have wanted to go with option 1.

@ziyangczi recently suggested we go with a function annotation approach instead of a naming convention in #1376, which could be quite nice and mean that we can have camelCase is all Qt Subclasses. If anyone has a concrete proposal for the function annotation case for #1376 I could do it over there now before merge.

[...]

PR #1376, opened 2020-06-20
Event handler refactor for image layer

sofroniewn:
# Description
This PR is from the same branch as is PRed here https://github.com/shanaxel42/napari/pull/2 to #1040 directly. Due to my badness at git though that PR has all changes from master in it making it hard to review so I'm just going to make this one directly. It will also let me check if all tests pass on our CI.

Copying over my last comment from there:

The PR should be >99% backwards compatible in that no tests changed, I only added tests. Some private methods/ attributes changed. I don't think any public ones did, certainly not public ones that are part of our main API.

I will write more about this elsewhere, but the PR is a bit of a hodgepodge in places as we havn't fully moved over to the event handler for all attributes, all layers etc, or the dims models, but at least all functionality is working and it will be more clear how to extend things in the future.

There are a couple major points to discuss

1. Do we like this `_transform_*_change` method that I added which allows us to preprocess data on events inside the event handler before propagating to all listeners. It was basically essential for the name case I described above, but I can see it being useful elsewhere, see [comment in #1040](https://github.com/napari/napari/pull/1040/files#r443139668).

2. Do we want all our view like objects now only to make updates based on the value they receive in the `_on_*_change` method or can they continue to have access to the whole data model. As has been pointed out in a number of comments now this hybrid is a little odd, see our colormap handling and [comment in #1040](https://github.com/napari/napari/pull/1040/files#r391268444).

3. Do we need the Interface classes, and if so how should they look? I went the the checking if attr exists approach instead as sometimes I wanted to register a listener that only implemented an update method for one or two of the methods, see how I added listeners for the layerlist which has the _transform_name_method and the viewer which only has a few like _on_help_change. If each one of these had to have its own interface or if we allowed more partial interfaces then I'm not sure what's gained by having them at all, see [comment in #1040](https://github.com/napari/napari/pull/1040/files#r393260527).

4. We should revisit the changes to the Event class we pulled from vispy and make our changes backwards compatible or clearly document why we needed to make breaking changes there (for example the autoconnect stuff, which although we were setting before I don't think we were really making much use of). See [comment in #1040](https://github.com/napari/napari/pull/1040/files#r391075515) and [comment in #1040](https://github.com/napari/napari/pull/1040/files#r393257546).

5. We should think about how many event handlers we're going to need and where they should live and how they might talk with each other. I had to register the viewer and the layerlist to the layer event handler to get some of the stuff to work (name and active_layer) is this bad, is this ok? Will viewer get one and dims get one? Then maybe we are good?

6. We should think a bit about where more controller like stuff will live (thinking now about things like napari#1353) and the slicing of data. At least working that example out a bit before merge will probably be good.

There are also a couple minor points

- A. We should add tests specifically for all the event handler stuff.

- B. We should clean up all the doc strings, many are out of date now or wrong or missing.

- C. We should think if we like having value everywhere in function signatures or if we want the name of the property etc.

We can always close this PR and work with the other ones, there are some great discussions happening there, or we can just press on with this one. I don't mind.

Probably next steps are in @napari/core-devs and @0x00b1 can weigh in on some of the big picture questions I mentioned above, and maybe @kne42 can go through and clean up some of the doc strings/ standardize function signatures/ add more comprehensive testing dedicated to the new functionality.

## Type of change

## Final checklist:


sofroniewn:
Interestingly this now has the failing windows test, which may previously have been fixed by #1347 and our mysterious scipy stats import. Not sure why this has reappeared here as that code should have been merged in ... The mystery there continues @tlambert03, but I wonder if this PR gives any clues ....

sofroniewn:
Windows tests were fixed by #1377, all tests are now passing!!

jni:
1. Do we like this `_transform_*_change` method that I added which allows us to preprocess data on events inside the event handler before propagating to all listeners. It was basically essential for the name case I described above, but I can see it being useful elsewhere, see [comment in #1040](https://github.com/napari/napari/pull/1040/files#r443139668).

I don't really. I think it would be nicer to write transform functions that get imported and used as needed.

2. Do we want all our view like objects now only to make updates based on the value they receive in the `_on_*_change` method or can they continue to have access to the whole data model. As has been pointed out in a number of comments now this hybrid is a little odd, see our colormap handling and [comment in #1040](https://github.com/napari/napari/pull/1040/files#r391268444).

I prefer things being based on the received value. I think this will make for more modular code and will make things like shrinking the number of object attributes much easier.

3. Do we need the Interface classes, and if so how should they look? I went the the checking if attr exists approach instead as sometimes I wanted to register a listener that only implemented an update method for one or two of the methods, see how I added listeners for the layerlist which has the _transform_name_method and the viewer which only has a few like _on_help_change. If each one of these had to have its own interface or if we allowed more partial interfaces then I'm not sure what's gained by having them at all, see [comment in #1040](https://github.com/napari/napari/pull/1040/files#r393260527).

Partial views are a very interesting use case, you're right. I still think it would be nice to document our models somehow with my [dataclass "spec"](https://github.com/napari/napari/pull/1040/files#r428545893), so that people know what a complete view looks like.

I think there is a middle ground we need to find here. For example, one use case I was envisioning was a view that ignores interpolation changes, because they only support nearest neighbour interpolation. (For example, the thumbnails could be views of the layer!) However, it's good to know that the view is doing this ignoring, and have the ignoring be explicit somehow, rather than have it be implicit.

There also needs to be a check that all public attributes have a corresponding event.

How do you like this test idea: construct a "view" of the Image layer that is in fact another image layer. Randomly mutate the values of the original layer, serialize them both, and check that the serializations are identical.

4. We should revisit the changes to the Event class we pulled from vispy and make our changes backwards compatible or clearly document why we needed to make breaking changes there (for example the autoconnect stuff, which although we were setting before I don't think we were really making much use of). See [comment in #1040](https://github.com/napari/napari/pull/1040/files#r391075515) and [comment in #1040](https://github.com/napari/napari/pull/1040/files#r393257546).

Can you elaborate on this? I wasn't aware that we were diverging from VisPy, and I'd like any changes here to be somehow mirrored in a VisPy PR. Maybe that looks like separating VisPy events into BasicEvent and FancyEvent (aka Event), and we only use Basic. But I'd prefer that we don't diverge.

5. We should think about how many event handlers we're going to need and where they should live and how they might talk with each other. I had to register the viewer and the layerlist to the layer event handler to get some of the stuff to work (name and active_layer) is this bad, is this ok? Will viewer get one and dims get one? Then maybe we are good?

Why would it be bad to have multiple handlers? Isn't the idea to have one emitter (per model instance), many listeners?

6. We should think a bit about where more controller like stuff will live (thinking now about things like napari#1353) and the slicing of data. At least working that example out a bit before merge will probably be good.

I actually was hoping to keep that discussion separate and happening *after* this PR. They interact in lots of ways and trying to consider both changes simultaneously feels like too much to chew on. I think we can probably handle this first and then simplify the models — and having well defined interfaces as discussed in point 2 will only help with that. For example, in my dataclass model, that's not actually the full interface but rather `ImageRenderingParametersInterface`, to be fully verbose. So all those parameters might form a nice bundle of a class. The Image itself also has Data, for example.

sofroniewn:
Had good chats with @shanaxel42 and @jni about this PR today, and there is good alignment. A couple key takeaways (numbering aligns with numbers above).

1. People agree the `_transform_*_change` methods aren't great, and it would be nice to see if we can figure out if we don't need them. It might necessitate adding a `layerlist` event handler to this PR, but that was going to happen eventually, so I'll investigate and post more.

2. People agree that getting to a world where we can drop the `layer` property from some of our view objects and just rely on the `value` and the function to the update is what we want. Some ways to approach this that were discussed include passing a dictionary for `value` with all the important things, but it was also raised that for many of the places that it isn't working right now we might actually be able to clean up the code to make it work, and this might make things much more modular and robust. What's nice about this change is that we can approach the refactor gradually doing it piece by piece.

3. We didn't resolve the Interfaces, right now I'm not using them everywhere, but they do have nice documentation properties. The main issue right now is that some of our "listeners" might not want to implement `_on_*_change` methods for all the events, so unclear if we just want to have methods that `pass` or use the `getattr` approach I have now. What's nice here is that we can take a gradual approach to changing this too.

I will try and add a `layerlist` handler remove the `_transform_*_change` method from this PR as a next step.

sofroniewn:
So after more investigation the case of forcing the `name` to be unique in the layerlist remains tricky. It is a bit of a special case/ one off and so @jni and I are wondering @kne42 if maybe that can be solved with a mechanism entirely contained to the layerlist base class - i.e. the concept of a uniquely named list. The hard thing though is it will have to maintain unique names even as the user tries to update them. Need to think about this more, but it seems a shame to radically change the event handler system to accommodate this one special case where we need to inject some global state (the full list of layer names) into what is otherwise a pretty local operation (updating the layer name). Would love if @shanaxel42 has any more ideas on how to do this one too ...

jni:
Yeah the order of events that makes sense to me is:

- the layer attempts to change its name, and as far as it's concerned, it's succeeded
- then the layer list, which is subscribed to layers changing their names, gets notified, checks against the global list, and *resolves any conflicts*.
- it then goes back and changes the layer's name.
- this triggers another event, but at this point, the layer list is happy and the loop stops.

For step 2, you need to know which layer changed its name. This can only work by knowing the event source *or* by emitting an (old_name, new_name) tuple rather than just the new name.

The only other alternative I see is that layers *don't in themselves have names*. Instead, only the layer list gives them names. Then we are sidestepping this whole dance and this will always be handled by the layer list.

tlambert03:
> This can only work by knowing the event source or by emitting an (old_name, new_name) tuple rather than just the new name.

maybe this is already obvious (can't tell by the phrasing here), but just wanted to point out that the event object does have a `source` attribute that provides the object that emitted the event.

sofroniewn:
To make that possible @jni we need to add a concept to the event handler of "priority" or "ordering" of the callbacks, as we need to make the layerlist a listener to the layer but make sure that it's callback executes last (otherwise even if it renames the layer during it's update there will be other updates that have yet to happen and when they do they will use the old name and things will get out of sync).

Also if the layerlist update happens last at that point we'll have two layers with the same name and know knowledge of which was which, and which had the old name as the update has already happened. Passing a tuple of `(index, new_name)` will solve that part. [EDIT: that won't work as the layer doesn't know it's index ...]

> maybe this is already obvious (can't tell by the phrasing here), but just wanted to point out that the event object does have a source attribute that provides the object that emitted the event.

Indeed, but we've lost that by the time the `_on_name_change(value)` gets called inside the layerlist


[EDIT: I'm really struggling to see how we can make this use case possible without some major changes]

sofroniewn:
I'm going to try a new approach which will use two events instead of one ....

jni:
> To make that possible @jni we need to add a concept to the event handler of "priority" or "ordering" of the callbacks, as we need to make the layerlist a listener to the layer but make sure that it's callback executes last (otherwise even if it renames the layer during it's update there will be other updates that have yet to happen and when they do they will use the old name and things will get out of sync).

I don’t think this matters. In step 3, the layer list updates the layer name. At that point, the layer eventhandler emits events with the new name, and of I’m not mistaken these will *necessarily* be processed by listeners *after* the original name change — so things will get in sync after just a few “clock cycles”. Am I missing something?

sofroniewn:
> I don’t think this matters. In step 3, the layer list updates the layer name. At that point, the layer eventhandler emits events with the new name, and of I’m not mistaken these will necessarily be processed by listeners after the original name change — so things will get in sync after just a few “clock cycles”. Am I missing something?

This is really about guaranteeing that step 1 is complete before this change happens, i.e. if the layer list event handler can make this change while the other event is still processing then it won't result in serial updates, it will result in a nested update that then gets overwritten. Certainly the implementation I tried suffered from this in practice.

If the layer widget emits name A then the layerlist hears it first (which may happen with no "priority"/ "ordering") and then changes it to A_1 then emits again then the layer hears it takes on A_1 and that event finishes, but then the original A from the layer widget come in and sets the layer name back to A!!!

The "two event approach" might work, though

sofroniewn:
> The "two event approach" might work, though

Whoo it worked, thanks @shanaxel42 who suggested it!! I dropped the `_transform_*_change` approach and I didn't have to change the event handler at all, I just need to emit two events now. First one to check uniqueness and then one for the update. If there is no layerlist the uniqueness check does nothing and things proceed as normal, if there is then it does the check and things work.

This is ready for @kne42 to go through again :-)

napari/_qt/layers/qt_image_layer.py
```diff
@@ -7,51 +8,54 @@
...
 
-class QtImageControls(QtBaseImageControls):
+class QtImageControls(QtBaseImageControls, ImageLayerInterface):
     """Qt view and controls for the napari Image layer.
 
     Parameters
     ----------
     layer : napari.layers.Image
-        An instance of a napari Image layer.
+       An instance of a napari Image layer.
```
kne42: this is a relic from #1040 but I believe this is the wrong indentation as we should use 4 spaces to indent instead of 3

napari/utils/naming.py
```diff
@@ -75,3 +75,10 @@ def magic_name(value, *, path_prefix):
     for name in names:
         if name.isidentifier() and name in varmap and varmap[name] is value:
             return name
+
+
+def force_name_unique(name, names):
+    if names.count(name) == 0:
```
kne42: maybe this is nitpicky but it would be slightly more performant if you used

```suggestion
    if name not in names:
```

since in the case where it's `False` it will stop at the first duplicate name instead of continuing to search the rest of the list. i don't think it'll be noticeable unless you have like 1000 layers tho

kne42 (approved):
Great work @sofroniewn, thanks for taking over! :)

3. I think they're useful as a form of documentation but a bit overly verbose and I don't think we should inherit from them (e.g. used as a traditional interface) as their methods are not required to be implemented. We could use what @jni suggested in #1040. I think they could be used in a test on the layer models as I believe that they must all be implemented on them. We could also use them to autogenerate documentation for the types of callbacks we can have.

4. I don't mind diverging from VisPy and I might even suggest making a custom event system for us anyways. The only reason we copied their code was because they did a lot of work for us.

> So after more investigation the case of forcing the name to be unique in the layerlist remains tricky. It is a bit of a special case/ one off and so @jni and I are wondering @kne42 if maybe that can be solved with a mechanism entirely contained to the layerlist base class - i.e. the concept of a uniquely named list.

Sounds fine to me :)

sofroniewn:
Thanks for the review @kne42, I'll make changes promptly. I also heard from @shanaxel42 on how to make the event stuff backwards compatible with vispy and everyone's on the same page with trying to drop the models from our views as much as possible and just rely on the values passed by the event handler, @pwinston had some good comments yesterday about how that pattern makes for simpler more modular code.

The only remaining big thing to settle on this PR is what to do with the `Interfaces` especially now that I have the getattr approach. I think at this point they really strictly are for "documentation" like purposes, but then we should think about if they are the most effective approach or not, particularly as I'm now often in a situation where a handler is only listening to a small number of possible events that an emitter might send. Any further thoughts from @jni and @tlambert03 here would be great, then we can wrap this PR up!!

napari/utils/naming.py
sofroniewn: Will change, I actually had a version where i needed count > 1 which is why it was there. Good catch

napari/_qt/layers/qt_image_layer.py
sofroniewn: fixed!

sofroniewn:
I have dropped the interface. This PR now just needs some love and attention around doc strings, some more tests around the new event handler stuff specifically, and for me to make the event/ event emitter class changes themselves be backwards compatible, then it will be done!

sofroniewn:
Ok! All the non-backwards compatible changes to the Event Class/ EventEmitter Class have now been reverted. I kept one really nice convenience which is to default a single arg passed to have the same effect as the `value` kwarg which means we can do
```python
layer.events.name('new_name')
# and NOT
layer.events.name(value='new_name')
```
which is quite nice.

This still needs doc strings and more tests, but will then be done!!

napari/utils/event_handler.py
```diff
@@ -0,0 +1,13 @@
+from abc import abstractmethod
```
tlambert03:
```suggestion
from abc import ABC, abstractmethod
```

napari/utils/event_handler.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class EventHandler:
```
tlambert03, line 7:
```suggestion
class EventHandler(ABC):
```
unlike the interface which had subclassing conflicts (right?) this one actually _can_ be a "real" abstract base class if we just just subclass ABC.  

EDIT:  but see my other comment wondering  why we need this _and_ LayerEventHandler

napari/layers/layer_event_handler.py
```diff
@@ -0,0 +1,40 @@
...
+    """
+    Event handler for layer specific events. Receives change events made from
+    the data, controls, or visual interface and updates all associated
+    components.
+
+    Ex. ImageLayer components = [qt_image.py, vispy_image.py, image.py]
+    """
+
+    def __init__(self, component=None):
+        super().__init__(component)
```
tlambert03: you can remove this whole method if you're just calling super()

sofroniewn:
Ok @tlambert03 @jni @ziyangczi this is ready for your review! @shanaxel42 feel free to take a look too, thanks for all your hard work getting this started, and thanks @kne42 for help bringing this PR to the finish line!!

One of the best things about this PR is that it is fully backwards compatible, tests were only added, not changed which I really love.

We definitely have more to do to remove the whole model from the view which everyone things will result in an even cleaner more modular codebase but that can come later (I might look into it a little more after this PR merges too, but wanted to keep this one somewhat bounded).

After this merges @ziyangczi is going to take over some of the effort to propagate the event handler refactor to our other layers - Labels, Surface, Vectors, Points, Shapes (probably with individual PRs in that order)!

napari/layers/layer_event_handler.py
```diff
@@ -0,0 +1,42 @@
+import logging
+
+
+from ..utils.event_handler import EventHandler
+
+
+logger = logging.getLogger(__name__)
+
+
+class LayerEventHandler(EventHandler):
```
tlambert03: curious, what makes anything about this layer-specific.  Isn't this pretty much what a generic EventHandler is going to look like?  why can't this just be "`EventHandler`" itself

napari/utils/event_handler.py
```diff
@@ -0,0 +1,13 @@
...
+    def __init__(self, component=None):
+        self.components_to_update = [component] if component else []
+
+    def register_component_to_update(self, component):
+        self.components_to_update.append(component)
+
+    @abstractmethod
+    def on_change(self, event=None):
+        ...
```
tlambert03: maybe a short docstring  instead of an ellipses?

napari/layers/layer_event_handler.py
```diff
@@ -0,0 +1,42 @@
...
+    Ex. ImageLayer components = [qt_image.py, vispy_image.py, image.py]
+    """
+
+    def __init__(self, component=None):
+        super().__init__(component)
+
+    def on_change(self, event=None):
+        """
+        Process changes made from any interface
+        """
```
tlambert03:
```suggestion
        """Process changes made from any interface."""
```

([PEP257](https://www.python.org/dev/peps/pep-0257/#one-line-docstrings))

napari/layers/base/base.py
```diff
@@ -318,57 +337,65 @@ def editable(self):
...
-        self._set_editable(editable=editable)
-        self.events.editable()
+    def editable(self, value):
+        self.events.editable(value)
+
+    def _on_editable_change(self, value):
+        self._editable = value
+
+    def _update_editable(self):
+        self.editable = self._is_editable
```
tlambert03, line 349: can you help me understand the flow here?  why do we need `editable`, `is_editable`,  and the indirection?

napari/layers/intensity_mixin.py
```diff
@@ -121,7 +125,9 @@ def gamma(self):
 
     @gamma.setter
     def gamma(self, value):
+        self.events.gamma(value=value)
```
tlambert03: this is where we could just use `self.events.gamma(value)` right (and other places above)?   are we going to try to use that form?

napari/components/add_layers_mixin.py
```diff
@@ -48,17 +48,12 @@ def add_layer(self, layer: layers.Layer) -> layers.Layer:
...
-        layer.events.deselect.connect(self._update_active_layer)
-        layer.events.status.connect(self._update_status)
-        layer.events.help.connect(self._update_help)
-        layer.events.interactive.connect(self._update_interactive)
-        layer.events.cursor.connect(self._update_cursor)
-        layer.events.cursor_size.connect(self._update_cursor_size)
-        layer.events.data.connect(self._on_layers_change)
+        layer.event_handler.register_component_to_update(self)
+
         layer.dims.events.ndisplay.connect(self._on_layers_change)
```
tlambert03, line 53: satisfying :)

tlambert03 (approved):
I'm approve! I'm happy with the direction this is taking us...
I left a few comments inline, might have some more thoughts as we start to proliferate this through the rest of the events code.  I admit I haven't carefully looked at every single line (mostly the patterns), but it's nice that none of the tests had to change and they pass :).

napari/utils/event_handler.py
sofroniewn: To prepare us for having `ViewerEventHandler` and `DimsEventHandler`. Agreed, might be a bit thin right now though

napari/layers/layer_event_handler.py
sofroniewn: Ahhh ok misunderstood point above. I agree actually. I think the `EventHandler` as it is right now can be fully generic. You're right there is nothing Layer specific. Maybe in the future we have proper controllers etc that have logic in them specific to our different objects, but maybe we simplify things for now and just simply have a single event handler.

napari/utils/event_handler.py
sofroniewn: Ah, seeing point below, you're right - we could just make this a fully reusable `EventHandler`. Probably a good idea for now, we can easily make specific ones later

napari/layers/layer_event_handler.py
tlambert03: yeah. I can't immediately think of stuff that _has_ to be specific to any of those things (with regards to events).  and it's always nicer not to have to remember the "5 different event handlers".  I think it'd be nice to try not to have all the variants in the future if at all possible.

napari/layers/base/base.py
sofroniewn: yeah, this is a bit clunky, it really is because different layers can be editable in different configurations and so this allows layers to have some specific control over this. I don't love the pattern though ....

napari/layers/intensity_mixin.py
sofroniewn: yes, i'll change that - not sure how that one slipped by!! not needed anymore

napari/utils/event_handler.py
sofroniewn: I can add a doc string etc, but if I think dropping the layer specific one for now makes sense and then this will go away

napari/utils/event_handler.py
sofroniewn: dropping LayerEventHandler, so no longer needed

napari/utils/event_handler.py
sofroniewn: dropping LayerEventHandler, so no longer needed

napari/layers/layer_event_handler.py
sofroniewn: dropping LayerEventHandler, so no longer needed

napari/utils/event_handler.py
sofroniewn: dropping LayerEventHandler, so no longer needed

napari/layers/layer_event_handler.py
sofroniewn: dropping LayerEventHandler, so no longer needed

napari/layers/layer_event_handler.py
sofroniewn: done

napari/layers/base/base.py
sofroniewn: This stuff also really isn't used by the image layer, but I had to change it a bit, as long as it's not broken (which I don't think it is) maybe we live with it for a little longer. Ultimately I don't think we want this stuff living on the layer model anyway (see #1353 etc for example)

sofroniewn:
On the suggestion of @tlambert03 I've removed the concept of component specific event handlers for now, which we didn't need, and it wasn't clear we'd need in the future.

One thing that it might be nice to do is implement a suggestion of @ziyangczi which is to use some form of function annotation rather than the naming convention `_on_*_change` to indicate that this function is a callback for an event of a particular type. This would allow us to do things like uniformly use camelCase in the names of our Qt functions - see discussion https://github.com/napari/napari/issues/1091#issuecomment-650563169 and just be more explicit with more control.

I'm not sure if this is exactly what @ziyangczi had in mind but I could imaging a syntax like

```python
@event_callback('colormap')
def my_func_name(self, value):
    ....

# instead of

def _on_colormap_change(self, value)
    ....
```

I would need help implementing that / I'd need someone else to implement it. If we know we like that approach probably makes sense to do it in this PR still.

Thoughts @jni @tlambert03 @ziyangczi? This is ready for a final review from @jni

tlambert03:
`@event_callback('colormap')` ... definitely intriguing!  will think about how it might look to implement

tlambert03:
actually, one more thing: I'm not crazy about the method name `register_component_to_update`.  It's more of a notification than an update (i.e. the listener on the other end needn't actually be updated or do anything)...  what do you think of `event_handler.register_listener(object)`?

sofroniewn:
I like `register_listener`, will change

tlambert03:
one more question: The pattern implemented here, where we:
```python
# create EventHandler
self.event_handler = EventHandler(component=self)
#create EmitterGroup
self.events = EmitterGroup(source=self, ...)
# immediately connect all of the events from the emitter group to the EventHandler
self.events.connect(self.event_handler.on_change)
# ... where the event_handler is expecting stereotyped callback names: `_on_*_change`
```
Is almost exactly what the `auto_connect` argument of the `EmitterGroup` is there to do.  So in most cases, we set `auto_connect=False`, and then end up accomplishing the same end result (except that our auto connects are named `_on_*_change` and theirs are just called `_on_*`).    I briefly mentioned this back at the end of my comment in  https://github.com/napari/napari/pull/1020#pullrequestreview-369990071 and will copy here:

> the last comment I would make is that I don't know that we're using the `EmitterGroups` and `EventEmitters` to their fullest. They have lots of nice tricks on them (autoconnecting of methods to events, detection of the source of the event, etc...) that the controller here is kinda duplicating. I'm not saying the controller doesn't belong, but I do think we shouldn't duplicate functionality that already exists in those Emitters.

If this similarity was noticed, got discussed and subsequently rejected, that's fine... just wanted to point it out again in case that got missed before.

**EDIT:** I guess the main issue/difference is that `auto_connect` is using the `source` of the `EmitterGroup` as the owner of the `_on_change` method... and the current PR provides a way for any other arbitrary object to also register themselves as listeners.  so they're definitely not the same.  but close!

sofroniewn:
> I'm not saying the controller doesn't belong, but I do think we shouldn't duplicate functionality that already exists in those Emitters.

Indeed, we have reached the point where I think the whole PR could probably be done using functionality within the `EventEmitterGroup` using the `auto_connect=True`, their `on_*` naming convention without us adding our own event handler class, and passing explicit `source` variables (i.e. you don't have to have the default source, then you can achieve the arbitrary object listeners I think, but it is a bit weird). If we edited the those classes more and diverged them from vispy we could even add our own naming conventions etc. and do more. They are very powerful, and right now we're basically using very little of them.

I guess having our own "handler" here gives us more flexibility for the future, and is a single bottleneck events must flow through and it is a massive reduction in the complexity of what events are allowed to do. Basically instead of passing events to all our views we just call simple update methods that take values (but note that isn't really fully realized here as our views still have the models on them).

If we are going to lock things down with this event handler we could replace the vispy event system with something far far simpler (no blockers, simple flow etc.). If it ain't broke why fix it, but on the other hand this refactor has always been about trying to make things simpler for people to understand and that portion of the code was always something we just copied over which is more complex then we use, so I'd be in favor of that by the time this is all done!!

tlambert03:
> I guess having our own "handler" here gives us more flexibility for the future, and is a single bottleneck events must flow through

Definitely, I wouldn't want to lose the handler bottleneck.  that's great.  I haven't thought this all the way through.  if the current pattern is necessary to achieve the bottleneck, then it should stand as is.

> If we are going to lock things down with this event handler we could replace the vispy event system with something far far simpler (no blockers, simple flow etc.).

yeah.  I'm was a bit curious about the opposition to touching the `EventEmitter` object (and the recent reversion to make sure it matched the vispy code).  why are we vendoring that code if we don't want to modify it?

sofroniewn:
> yeah. I'm was a bit curious about the opposition to touching the EventEmitter object (and the recent reversion to make sure it matched the vispy code). why are we vendoring that code if we don't want to modify it?

I think it was more the concept of non-backwards compatible changes/ understanding if the changes were really necessary.

sofroniewn:
Just made some more minor changes, trying to reduce the usage of the model within the qt views. It's going pretty well there, should be very doable. We'll have to change the popup sliders a bit too, but I'd still like us to make every slider popup (see #879, but we can wait on that)

tlambert03:
regarding the `@callback` decorator.  one the tricky thing is the fact that the decorator would be decorating the unbound class method.  and what we need is to connect is the _bound_ (instantiated) method to the yet-to-be-created `self.layer.event_handler`.  but it's not that hard... one can "search for decorated methods" in the `__init__` method, right where `register_listener(self)`, is currently being called.

I took a crack at implementing it, if you want to see how one version of that might look:
https://github.com/sofroniewn/napari/compare/event_handler_njs...tlambert03:event-decorator
... with the meat of it being [this `@call_on.<event_name>` decorator](https://github.com/tlambert03/napari/blob/bf762187324c9d752411d1409c50485391e3b2eb/napari/utils/event_handler.py#L12-L21), and [this method of searching an object for decorated methods](https://github.com/tlambert03/napari/blob/bf762187324c9d752411d1409c50485391e3b2eb/napari/utils/event_handler.py#L40-L48).

not all tests are passing yet, but I'm actually not sure I love it after working on it.  While the decorator _does_ get us away from an implicit method name-based connection (and allows connecting arbitrarily named functions like `my_func_name`), in most cases we'll probably still just leave the function name as is, and it leaves us with this pattern everywhere (and adds a lot of lines of code):
```python
    @call_on.interpolation
    def _on_interpolation_change(self, text):
```

however, going through that process did reveal a limitation of the current method (that might be easier to fix with the explicit connections via decorator).  Specifically: if we assume that a given object listens to multiple event_handler instances, and two event handlers have an event with the same name (for instance "`_on_name_change`"). Currently, I think we're stuck calling the same callback function ("_on_name_change") regardless of which event handler is emitting the event.  We only have a single event_handler on the base `Layer` at the moment... but that will likely change in the future.

sofroniewn:
Amazing @tlambert03, let's have @jni and @ziyangczi weigh in before deciding next steps here, but I think your point about multiple listeners for the same object is an interesting one. We'll have to see how this all evolves!

jni:
Hi everyone, just fyi I'm not available this weekend but I'll review this tomorrow (Monday my time). The discussion looks great at first glance. =)

napari/_qt/layers/qt_base_layer.py
```diff
@@ -47,62 +62,38 @@ def __init__(self, layer):
...
-        value : float
-            Opacity value for shapes.
-            Input range 0 - 100 (transparent to fully opaque).
-        """
-        with self.layer.events.blocker(self._on_opacity_change):
-            self.layer.opacity = value / 100
+        # Once EVH refactor is done, these can be moved to an initialization
+        # outside of this object
+        self._on_opacity_change(self.layer.opacity)
+        self._on_blending_change(self.layer.blending)
```
jni, line 76: Can you explain this for me? Currently I found it confusing that this wasn't happening in *this* refactor. At any rate, I think at least the `_on_*_change` calls should get bundled with their respective box initialisations... (ie move `_on_opacity_change` to line 67 and `on_blending_change` to line 72)

napari/_qt/layers/qt_base_layer.py
```diff
@@ -47,62 +62,38 @@ def __init__(self, layer):
...
         """Receive layer model blending mode change event and update slider.
 
         Parameters
         ----------
-        event : qtpy.QtCore.QEvent, optional.
-            Event from the Qt context, by default None.
+        text : str
+           Blending mode used by VisPy. Must be one of our supported
+           modes:
+           'transluenct', 'additive', 'opaque'
```
jni:
```suggestion
           'translucent', 'additive', or 'opaque'
```

napari/_qt/layers/qt_image_base_layer.py
```diff
@@ -73,25 +74,35 @@ def __init__(self, layer):
...
         self.gammaSlider = sld
-        self.gamma_slider_update()
 
         self.colorbarLabel = QLabel(parent=self)
         self.colorbarLabel.setObjectName('colorbar')
         self.colorbarLabel.setToolTip('Colorbar')
 
-        self._on_colormap_change()
+        # Once EVH refactor is done, these can be moved to an initialization
+        # outside of this object
```
jni, line 86: As above... I'd like a bit of clarification of *when* this is supposed to happen (ie what exactly has to be done), and maybe move the items to their respective sliders, although now that it's more things to change than two I see the value of grouping near-obsolete things together...

napari/_qt/layers/qt_image_base_layer.py
```diff
@@ -120,39 +132,56 @@ def _clim_mousepress(self, event):
...
         ----------
-        event : qtpy.QtCore.QEvent, optional.
-            Event from the Qt context, by default None.
+        value : text
+            Colormap name.
         """
-        name = self.layer.colormap[0]
+        # To complete the EVH we need to standardize passing of colormaps.
+        # including generation of the colorbar.
+        name = value if value else self.layer.colormap[0]
```
jni: what is `if value`? Does this mean `if value is not None`? Or `if value != ''`? When would either of those things happen?

jni: (In short: maybe document this behaviour in the docstring!)

napari/_qt/layers/qt_image_layer.py
```diff
@@ -243,27 +198,31 @@ def _toggle_rendering_parameter_visbility(self):
             self.attenuationSlider.hide()
             self.attenuationLabel.hide()
 
-    def _update_interpolation_combo(self):
+    def _update_interpolation_combo(self, ndisplay):
+        """Set allowed interploation modes for dimensionality of display.
```
jni:
```suggestion
        """Set allowed interpolation modes for dimensionality of display.
```

napari/_qt/layers/qt_image_layer.py
```diff
@@ -114,28 +112,56 @@ def __init__(self, layer):
...
         """Change interpolation mode for image display.
 
+       Parameters
+       ----------
+       text : str
+           Interpolation mode used by VisPy. Must be one of our supported
+           modes:
+           'bessel', 'bicubic', 'bilinear', 'blackman', 'catrom', 'gaussian',
+           'hamming', 'hanning', 'hermite', 'kaiser', 'lanczos', 'mitchell',
+           'nearest', 'spline16', 'spline36'
```
jni, line 132: Are all of these available for 3D? I'm also wondering whether we should dramatically reduce how many of these we actually expose...

napari/_qt/qt_layerlist.py
```diff
@@ -496,70 +508,78 @@ def __init__(self, layer):
...
-        self._on_thumbnail_change()
         self.layout.addWidget(tb)
 
         cb = QCheckBox(self)
         cb.setObjectName('visibility')
         cb.setToolTip('Layer visibility')
-        cb.setChecked(self.layer.visible)
         cb.setProperty('mode', 'visibility')
-        cb.stateChanged.connect(self.changeVisible)
+        cb.stateChanged[int].connect(self.events.visible)
```
jni, line 518: Does this have to be [int] instead of [bool] because of C++ quirks?

napari/_qt/qt_layerlist.py
```diff
@@ -496,70 +508,78 @@ def __init__(self, layer):
...
+        if old_name == name:
+            return
+        self.events.name_unique((old_name, name))
+        self.nameTextBox.oldText = self.nameTextBox.text()
+
+        # Prevent retriggering during clearing of focus
+        self.nameTextBox.blockSignals(True)
         self.nameTextBox.clearFocus()
         self.setFocus()
+        self.nameTextBox.blockSignals(False)
```
jni: Is there really not a context manager for this?

napari/_vispy/_tests/test_vispy_calls.py
```diff
@@ -11,7 +11,9 @@ def test_data_change_ndisplay_image(viewer_factory):
     layer = viewer.add_image(data)
     visual = view.layer_to_visual[layer]
 
-    @patch.object(visual, '_on_data_change', wraps=visual._on_data_change)
+    @patch.object(
+        visual, '_on_slice_data_change', wraps=visual._on_slice_data_change
+    )
```
jni, line 16: I know this isn't new, but can you explain to me for my own education what `@patch.object` does?

napari/_vispy/vispy_base_layer.py
```diff
@@ -134,17 +133,40 @@ def scale_factor(self):
...
+        """Receive layer model blending mode and update the visual.
 
-    def _on_blending_change(self, event=None):
-        self.node.set_gl_state(self.layer.blending)
+        Parameters
+        ----------
+        text : str
+           Blending mode used by VisPy. Must be one of our supported
+           modes:
+           'transluenct', 'additive', 'opaque'
```
jni:
```suggestion
           'translucent', 'additive', 'opaque'
```

napari/components/layerlist.py
```diff
@@ -42,33 +45,25 @@ def __init__(self, iterable=()):
...
-        return name
-
-    def _update_name(self, event):
-        """Coerce name of the layer in `event.layer`."""
-        layer = event.source
-        layer.name = self._coerce_name(layer.name, layer)
+        old_name, new_name = value
+        unique_name = force_unique_name(new_name, [l.name for l in self])
+        # Re-emit unique name
+        self[old_name].events.name(unique_name)
```
jni, line 66: Love this. Great job. =D

napari/layers/base/base.py
```diff
@@ -247,13 +254,21 @@ def name(self):
...
             return
-        if not name:
-            name = self._basename()
-        self._name = name
-        self.events.name()
+        # When the layer is used in the layer list the name must be unique.
+        # Emitting the name_unique event first allows the layer list to check
+        # if the name is unique and change it if necessary.
+        old_name = copy(self.name)
+        self.events.name_unique((old_name, value))
```
jni: Again. Just love this solution. =)

napari/layers/base/base.py
```diff
@@ -262,16 +277,18 @@ def opacity(self):
...
-    def opacity(self, opacity):
-        if not 0.0 <= opacity <= 1.0:
+    def opacity(self, value):
+        self.events.opacity(value)
+
+    def _on_opacity_change(self, value):
+        if not 0.0 <= value <= 1.0:
             raise ValueError(
-                'opacity must be between 0.0 and 1.0; ' f'got {opacity}'
+                'Opacity must be between 0.0 and 1.0; ' f'got {value}.'
```
```suggestion
                f'Opacity must be between 0.0 and 1.0; got {value}.'
```

napari/layers/base/base.py
```diff
@@ -470,28 +500,30 @@ def thumbnail(self):
         return self._thumbnail
 
     @thumbnail.setter
-    def thumbnail(self, thumbnail):
-        if 0 in thumbnail.shape:
-            thumbnail = np.zeros(self._thumbnail_shape, dtype=np.uint8)
-        if thumbnail.dtype != np.uint8:
+    def thumbnail(self, value):
```
jni: To be honest I would revert these. I think it should be `image` or `thumbnail`. If you *really* care about making it value for consistency with events, then the first thing I would do is `image = value` and proceed from that. `value is just so generic, and the rest of the code makes no sense without context.

napari/layers/intensity_mixin.py
```diff
@@ -70,36 +77,52 @@ def contrast_limits(self):
...
-        self._contrast_limits = contrast_limits
+        self.events.contrast_limits(contrast_limits)
+
+    def _on_contrast_limits_change(self, value):
+        """Set the contrast limits.
+
+        Parameters
+        ----------
+        value : tuple
+            Contrast limits, (min, max).
```
jni, line 87: Yeah, the more I read of this PR, the more I'm disliking "unifying on 'value'", which is sounding more and more like "unifying on 'x'". I don't necessarily want to hold it up, but I want us to think about reverting and calling the event values by what they actually hold for each event.

napari/layers/points/points.py
```diff
@@ -451,7 +450,7 @@ def data(self, data: np.ndarray):
         elif len(data) > cur_npoints:
             # If there are now more points, add the size and colors of the
             # new ones
-            with self.events.set_data.blocker():
+            with self.events.slice_data.blocker():
```
jni, line 453: Can you comment on why we need blockers here? I thought the whole point of the refactor was to not need blockers? Or is it because the refactor isn't actually done with the points layer?

napari/layers/vectors/vectors.py
```diff
@@ -243,7 +243,8 @@ def data(self, vectors: np.ndarray):
         self._displayed_stored = copy(self.dims.displayed)
 
         self._update_dims()
-        self.events.data()
+        self._update_editable()
```
jni, line 246: What's editable exactly? I noticed @tlambert03 asking about it in another comment but to be honest the answer didn't really clarify anything for me...?

napari/utils/event.py
```diff
@@ -459,13 +459,15 @@ def _normalize_cb(self, callback):
         return callback
 
     def __call__(self, *args, **kwargs):
-        """ __call__(**kwargs)
+        """ __call__(*args, **kwargs)
         Invoke all callbacks for this emitter.
```
jni:
```suggestion
        """Invoke all callbacks for this emitter.
```

napari/utils/event.py
```diff
@@ -537,16 +539,23 @@ def _prepare_event(self, *args, **kwargs):
...
+                kwargs['value'] = args[0]
+            else:
+                raise ValueError(
+                    "Event emitters may be called with an Event instance, "
+                    "a single argument, or with keyword arguments only."
+                    f"\ngot: args={args}, kwargs={kwargs} "
+                )
+        args = self.default_args.copy()
+        args.update(kwargs)
+        return self.event_class(**args)
```
jni, line 557: Oof. This whole change makes me dislike the `value` kwarg even more. Perhaps the keyword can be the event name?

napari/utils/event_handler.py
```diff
@@ -0,0 +1,55 @@
...
+        try:
+            value = event.value
+            logger.debug(f" value: {value}")
+        except AttributeError:
+            logger.debug(f" did not handle event {type}")
+            return
+
+        # Update based on event value
+        for component in self.components_to_update:
+            update_method_name = f"_on_{type}_change"
```
jni: I don't like shadowing Python built-in names, and in this case I think the value in doing so is minimal, so I would suggest:

```suggestion
        logger.debug(f"event: {event.type}")
        # until refactor on all layers is complete, not all events will have a
        # value property
        try:
            value = event.value
            logger.debug(f" value: {value}")
        except AttributeError:
            logger.debug(f" did not handle event {event.type}")
            return

        # Update based on event value
        for component in self.components_to_update:
            update_method_name = f"_on_{event.type}_change"
```

jni (approved):
@sofroniewn very good! I've left mostly very minor comments, or requests for clarification. I would say the biggest concern is that I'm unsure again about unifying on `value`, as it makes the actual code more difficult to read when it's not a one-liner.

I also think a high priority for this refactor is to enshrine @shanaxel42's information flow diagram to the docs somewhere so that new contributors are comfortable with the napari eventhandler model.

I'll approve because I don't want to hold this up any longer, but stuff to think about. Everything else is just typos. =) Thanks to all involved!!!

jni:
> yeah. I'm was a bit curious about the opposition to touching the EventEmitter object (and the recent reversion to make sure it matched the vispy code). why are we vendoring that code if we don't want to modify it?

I don't know about this specific instance, but in general the idea we've had with vendoring is to allow us to incorporate a quick bug or behaviour fix but always with the idea of contributing upstream and eventually removing the vendored copy. Whereas modifying entails a full divergence and I hate that — I want to play nicely with the ecosystem and reduce redundancy.

jni:
(which would argue against the modifications to the event classes in this PR...)

jni:
@sofroniewn please rename this PR to something less self-referential before merging! =)

napari/_qt/layers/qt_base_layer.py
sofroniewn: Yeah, this is a good question. It's hard right now as we don't have an explicit concept of a controller which is I guess where these imports, initialization and wiring up would be done (as it would be the one place which knows about both model and view). We can't do it in the event handler itself as we import the event handler into the layer model directly so that we can edit the layer model without the views, so we need an additional concept here. I'll work on some architecture docs in a new PR and we can revisit this there.

As to the grouping, I was grouping the soon to be obsolete properties together to make easier to remove them later.

napari/_qt/qt_layerlist.py
sofroniewn: yes, i got an error with bool. maybe it's because in general checkboxes can be tristate?

sofroniewn: does @tlambert03 know more?

napari/_qt/qt_layerlist.py
sofroniewn: there might be, does @tlambert03 know more?

napari/_vispy/_tests/test_vispy_calls.py
sofroniewn: From https://docs.python.org/3/library/unittest.mock.html#patch-object here you can "patch the named member (attribute) on an object (target) with a mock object." i.e. patching on `_on_slice_data_change` the `visual` object so that we can do things like count the number of times it gets called. Because we're using `wraps` too, we'll actually make the function call at the same time. This allows us to count the calls and also have the needed code execute

napari/_qt/layers/qt_image_layer.py
sofroniewn: They aren't all 3D, only two are. Agreed. Things would actually be much simpler is we only supported `nearest` and `linear` (called `bilinear` here) as I think these are the only two supported in 3D. Do you think we need to support more? Can we make a dedicated issue for this/ do this in a follow-up PR?

napari/_qt/layers/qt_base_layer.py
sofroniewn: Yeah, this is a good question. It's hard right now as we don't have an explicit concept of a controller which is I guess where these imports, initialization and wiring up would be done (as it would be the one place which knows about both model and view). We can't do it in the event handler itself as we import the event handler into the layer model directly so that we can edit the layer model without the views, so we need an additional concept here. I'll work on some architecture docs in a new PR and we can revisit this there.

As to the grouping, I was grouping the soon to be obsolete properties together to make easier to remove them later.

napari/_qt/layers/qt_image_base_layer.py
sofroniewn: See discussion above

napari/layers/intensity_mixin.py
sofroniewn: I'm happy to revert. I agree for long functions it's confusing, particularly now when we don't really use `value`. For short ones I think it's ok, but I'd change them too for consistency. Thoughts @tlambert03? If you have no objections I'll change them all now

napari/layers/points/points.py
sofroniewn: `Points` refactor not done yet - hopefully this will go when it is!!!!

napari/layers/vectors/vectors.py
sofroniewn: Editable is both a user setting and an attribute automatically determined that sets where things like adding points/ drawing with the paintbrush etc are allowed or not. We added both because someone might want to disable such functionality themselves AND because there are certain configurations like multiscale or 3D viewing where we don't support such interactivity. Definitely room for improvement here, but that's what the parameter is

napari/utils/event.py
sofroniewn: I actually don't mind `value` on the event as the one place where the payload is being carried. To me this makes a lot of sense and is fairly readable, but I see your comment about `value` inside the handlers and think we should change it there. @tlambert03 thoughts?

sofroniewn:
@jni thanks for the review! Did you have any feedback on the decorator approach instead of `_on_*change` approach that @tlambert03 tried out here https://github.com/napari/napari/pull/1376#issuecomment-650631537 for identifying callbacks. It does strike me that if we need to have an object listening to multiple event handlers something like this might necessary otherwise we can't reuse the names of events.

There's some discussion happening in the comments about where we will initialization of our views. We need to think about that more, I feel like we might need a controller concept, see here https://github.com/napari/napari/pull/1376#discussion_r447085459 as it would be the one place that knows about both models and viewers, but it is nuanced.

As to using `value` vs actual name like `rendering` in function signature, I'm fine going with actual name, but want to hear from @tlambert03 before making the change.

I'll work on an architecture doc in a new docs only PR

napari/utils/event.py
tlambert03: Sorry I’m on mobile at the moment and can’t see the full conversation.  I don’t care what you call the argument in the handler... but I agree with @sofroniewn that the event payload should be standardized.  I hated having to go find the event emitter to figure out the name of the event object attribute I’m supposed to be querying to retrieve the value.  I also don’t mind this code block either if it lets us make connections easier. 
@jni  is your main objection the argument name in the handler?  Or do you want the actual event object to have variable attribute names

tlambert03:
FWIW, I’m extremely Luke warm on the decorator as it stands.  I would support merging this and revisiting if needed (it’s mostly additions to this rather than replacements)

napari/_qt/layers/qt_image_base_layer.py
sofroniewn: this has been fixed :-) not needed anymore :-)

napari/_qt/layers/qt_base_layer.py
```diff
@@ -29,9 +30,23 @@ class QtLayerControls(QFrame):
...
 
+        self.events = EmitterGroup(
+            source=self,
+            auto_connect=False,
+            blending=Event,
+            opacity=Event,
+            status=Event,
+        )
+
+        # When the EVH refactor #1376 is done we might not even need the layer
```
liu-ziyang, line 41: shall we use a TODO signature? haven't seen much in napari codebase, is it discouraged?

napari/_qt/qt_layerlist.py
```diff
@@ -479,12 +480,23 @@ class QtLayerWidget(QFrame):
...
+        self.events = EmitterGroup(
+            source=self,
+            auto_connect=False,
+            selected=Event,
+            name=Event,
+            name_unique=Event,
+            visible=Event,
+        )
+
+        # When the EVH refactor #1376 is done we might not even need the layer
```
liu-ziyang, line 493: see above for TODO

napari/utils/event_handler.py
```diff
@@ -0,0 +1,54 @@
...
+        # value property
+        try:
+            value = event.value
+            logger.debug(f" value: {value}")
+        except AttributeError:
+            logger.debug(f" did not handle event {event.type}")
+            return
+
+        # Update based on event value
+        for component in self.components_to_update:
```
liu-ziyang, line 50: we talked about this regarding having annotation-based detection vs name-based detection, wonder what others' opinion on this too

napari/utils/naming.py
```diff
@@ -75,3 +75,27 @@ def magic_name(value, *, path_prefix):
...
+
+    Returns
+    -------
+    name : str
+        Name that is not present in the list of names.
+    """
+    if name not in names:
+        return name
+    else:
+        return force_unique_name(inc_name_count(name), names)
```
liu-ziyang, line 101: what if the forced name already exists? shall this method loop and check?

napari/utils/event_handler.py
sofroniewn: See discussion above and bellow this comment, and links in this comment for potential implementation https://github.com/napari/napari/pull/1376#issuecomment-650631537

napari/_qt/layers/qt_base_layer.py
sofroniewn: I'm open - thoughts @jni? @tlambert03?

napari/utils/naming.py
sofroniewn: The looping and checking is happening recursively as the original function is called `force_unique_name` and then calling itself inside itself.

liu-ziyang:
agreed with talley, let's merge this one and think more on the decorator. I still love to have a decorator based detection, but like talley said it is more complicated and introduced a lot of lines. (maybe a composite method where we name search and use a decorator for special cases?)

napari/_qt/qt_layerlist.py
sofroniewn: Yes we have one `qt_signals_blocked` that @tlambert03 already made. Am switching to use it now

napari/_qt/layers/qt_base_layer.py
sofroniewn: Ooo - ok, I have one idea - we could add it to the `register_listener` method of the event handler. It would be nice if we had a standardized way of figuring out what the event values should be and then re-emitting them.  i.e. when you register the `qt_image_layer` we see that it has the `_on_visible_change` method and we know that the `Image` layer emits the `visible` event, and we know the current `visible` value lives at `layer.visible` so we call `_on_visible_change(layer.visible)` for the `qt_image_layer`?

We might be baking in a lot of assumptions there, but it would remove the need to a lot of initialization code throughout the repo. Curious @tlambert03 @jni @ziyangczi what you think of this idea?

PR #1377, opened 2020-06-21
Fix windows tests

tlambert03:
# Description
looks like some of the mocks used for reducing GUI events didn't work in the windows CI environment.  They work locally on windows, and they work on every other platform, so this PR just reshows the viewer for a couple tests on CI, and skips a mock assertion on windows CI.

the `test_nbscreenshot` test is still failing occasionally... still not sure why.

## Type of change

- [x] tests

# How has this been tested?

## Final checklist:


sofroniewn (approved):
Hmm, ok good to know. Thanks for chasing down. If we do pull out of our `tests.md` file this will be good info to put in there!

PR #1378, opened 2020-06-21
Dedicated testing doc

sofroniewn:
# Description
This PR adds a dedicated testing doc and moves a couple GUI tests into the right places in the repo, inspired by conovs coming out of #1363.

I added some language about unit tests, integration tests, functional tests from https://www.softwaretestinghelp.com/the-difference-between-unit-integration-and-functional-testing/. Some of this is more "aspirational" then where we are in practice right now as @tlambert03 recently pointed out, but documenting this stuff might help us get there.

I also added a `known issues` section which is a little thin right now, but maybe @tlambert03 if you wanted to take this PR over and flesh it out that would be great!! (Also feel free to rework the intro).

I added some language in the contributing guide to now point to this new doc.

Let me know what you think!

## Type of change

- [x] Documentation, refactor


tlambert03:
[added](https://github.com/napari/napari/pull/1378/commits/86e87cc9e14482c53779882c4a22ac8a1af781c6) a few of the additions discussed in #1363

tlambert03 (approved):

sofroniewn:
Love the additions @tlambert03 very nice! Maybe @jni can give this a read over before merge?

docs/source/developers/TESTING.md
```diff
@@ -0,0 +1,168 @@
...
+Unit tests are at the base of the pyramid because they are the easiest to write and
+the quickest to run. The time and effort to implement and maintain tests increases
+from unit tests to integration and functional tests.
+
+## Test organization
+
+All of `napari` tests are located in folders named `_tests`. We keep our unit
+tests located in the individual folders with the modules or functions they are
+testing (e.g. the tests for the `Image` layer are located in a `_tests` along
+side the `image.py` file). Our integration and functional tests are located in
```
jni:
```suggestion
testing (e.g. the tests for the `Image` layer are located in a `_tests` folder
alongside the `image.py` file). Our integration and functional tests are located in
```

docs/source/developers/TESTING.md
```diff
@@ -0,0 +1,168 @@
...
+pip install -r requirements/test.txt
+pytest
+```
+
+There are a very small number of tests (<5) that require showing GUI elements, (such
+as testing screenshots). By default, these are only run during continuous integration.
+If you'd like to include them in local tests, set the environment variable "CI":
+
+```sh
+export CI=1
```
jni: This could also be:

```suggestion
CI=1 pytest
```

docs/source/developers/TESTING.md
```diff
@@ -0,0 +1,168 @@
...
+   any additional cleanup (such as using `qtbot.addWidget` or calling
+   `viewer.close()`) at the end of the test.  Duplicate cleanup may cause an
+   error.  Use the factory as follows:
+
+    ```python
+    # the viewer_factory fixture is defined in napari/conftest.py
+    def test_something_with_a_viewer(viewer_factory):
+        # viewer factory takes any keyword arguments that napari.Viewer() takes
+        view, viewer = viewer_factory()
+        # note, `view` here is just a pointer to viewer.window.qt_viewer
```
jni, line 143: Incidentally, how necessary is that pointer? Maybe we should just return `viewer` and use the pattern `view = viewer.window.qt_viewer` where needed?

docs/source/developers/TESTING.md
```diff
@@ -0,0 +1,168 @@
...
+
+        # do stuff with the viewer, no qtbot or viewer.close() methods needed.
+        ...
+    ```
+
+> If you're curious to see the actual `viewer_factory` fixture definition, it's in `napari/conftest.py`
```
jni, line 149: Also incidentally, does conftest.py need to be public as opposed to `_conftest.py`? I was annoyed at my terminal trying to suggest it as an import path the other day...

docs/source/developers/TESTING.md
```diff
@@ -0,0 +1,168 @@
...
+
+### Mocking: "Fake it until you make it"
```
jni:
```suggestion
### Mocking: "Fake it till you make it"
```

Might as well go all-in with colloquialisms. =P

jni (approved):
Looooove these! I made some very minor suggestions.

docs/source/developers/TESTING.md
tlambert03: good call.  that's probably better

docs/source/developers/TESTING.md
tlambert03: very unnecessary!  i think it's vestigial from some early test patterns.  I like the idea of just returning the viewer.  I'll open a PR for that and we can change these docs

docs/source/developers/TESTING.md
tlambert03: just tested.  `_conftest` does not work.  I think we could however use a `pytest11` entry_point in our setup.cfg instead... should we choose to go in that direction.

docs/source/developers/TESTING.md
sofroniewn: That PR #1386 looks great to me. How about I merge this PR first and then you can update docs over there?

docs/source/developers/TESTING.md
sofroniewn: I'm going merge this PR as is now and we can revisit this in follow up PRs

docs/source/developers/TESTING.md
tlambert03: sure

sofroniewn:
Actually @tlambert03 should we add more info to this doc about our CI, including what we test: different python versions, different OS versions, different Qt versions, showing the full grid for merge into master, and the partial grid on our PRs?

Also what we've disabled for the various platforms - for example I always keep forgetting what fraction of vispy/ Qt stuff is enabled on windows CI right now (I seem to recall not all) and this would be a great place to document that.

We could rename whole doc `Testing and CI`. Thoughts?

tlambert03:
I don't have strong feelings about this.  I haven't too frequently come across documentation on CI strategies in other projects, and they do change/evolve over time, so it would make another place we have to check each time we make changes (lest it go out of sync).  But if you want to have a place to go to look those values up (and want to make sure that new PRs always update it 😂) then I don't have any objections.

Issue #1257, opened 2020-05-16
Add an easy way to create empty layers of arbitrary dimensionality


jni:
## 🚀 Feature

A user on image.sc [found it hard](https://forum.image.sc/t/controlling-dimensionality-of-napari-point-shape-layers/37896) to create shapes layers of arbitrary dimensionality. Instead, `add_shapes` without data behaves differently depending on the state of the viewer, which is confusing. One may want, for example, to create a 2D shapes layer that broadcasts on a 3D movie, say, for calcium imaging data. It's entirely not obvious that you do this by passing an empty array of shape `(0, 0, ndim)`. A clearer way would be to add an `ndim` keyword argument that controlled dimensionality when no data is given.

sofroniewn:
Indeed, we should improve this. Note this is basically a duplicate of #835 which still needs to be resolved

jni:
Heh, always fun to disagree with my past self. 😂😅

PR #1379, opened 2020-06-21
Add ndim as a keyword argument to shapes to support creating empty layers

DragaDoncila:
# Description
- Add ndim keyword arg
- Move inferring ndim from data (from constructor to shapes utils)

Fixes issue #1257.

Currently it fails the serialization test. Not entirely sure why but I think it's expecting ndim to be a property, but it's simply a parameter.

## Type of change

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/components/add_layers_mixin.py
+++ b/napari/components/add_layers_mixin.py
@@ -496,6 +496,7 @@ class AddLayersMixin:
     def add_shapes(
         self,
         data=None,
+        ndim=None,
         *,
         properties=None,
         shape_type='rectangle',
@@ -525,6 +526,9 @@ class AddLayersMixin:
             List of shape data, where each element is an (N, D) array of the
             N vertices of a shape in D dimensions. Can be an 3-dimensional
             array if each shape has the same number of vertices.
+        ndim : int
+            Number of dimensions for shapes. When data is not None, ndim must be D.
+            An empty shapes layer can be instantiated with arbitrary ndim.
         properties : dict {str: array (N,)}, DataFrame
             Properties for each shape. Each property should be an array of
             length N, where N is the number of shapes.
@@ -602,7 +606,8 @@ class AddLayersMixin:
             The newly-created shapes layer.
         """
         if data is None:
-            ndim = max(self.dims.ndim, 2)
+            if ndim is None:
+                ndim = max(self.dims.ndim, 2)
             data = np.empty((0, 0, ndim))
 
         layer = layers.Shapes(
--- a/napari/components/add_layers_mixin.py
+++ b/napari/components/add_layers_mixin.py
@@ -496,8 +496,8 @@ class AddLayersMixin:
     def add_shapes(
         self,
         data=None,
-        ndim=None,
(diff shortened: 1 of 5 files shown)


jni:
@sofroniewn would you mind checking the serialization test failures here? It looks like it's checking that the number of attributes is the same as the number of keyword arguments, which seems to me like it's too specific a test? But we didn't want to change it without fully understanding it.

sofroniewn:
> @sofroniewn would you mind checking the serialization test failures here? It looks like it's checking that the number of attributes is the same as the number of keyword arguments, which seems to me like it's too specific a test? 

Right now the serialization is checking an exact match with the `__init__` and that a full round trip with the layer data reproduces the same layer. I've always thought this was quite nice, so every param in the `_get_state` dict appears in the main API and visa-versa.

I'd just go ahead and include `ndim` in the `_get_state` dict here https://github.com/napari/napari/blob/master/napari/layers/shapes/shapes.py#L1189 and then everything should pass fine without having to change the tests :-)

jni:
@sofroniewn 

> the serialization is checking an exact match with the `__init__` and that a full round trip with the layer data reproduces the same layer.

I feel like the second test is the only needed one. This will be especially true once we start implementing @pwinston's refactor proposal of at-most 7 attributes per object. Having said that

> I'd just go ahead and include ndim in the _get_state dict here

this is the fastest fix right now so I'm in favour of revisiting this discussion if it seems appropriate when refactoring. @DragaDoncila want to make this change and push? At which point @sofroniewn this will be ready for review!

DragaDoncila:
That should be ready now @sofroniewn.

napari/components/add_layers_mixin.py
```diff
@@ -602,11 +606,13 @@ def add_shapes(
             The newly-created shapes layer.
         """
         if data is None:
-            ndim = max(self.dims.ndim, 2)
+            if ndim is None:
```
sofroniewn, line 609: This default was added to make https://github.com/napari/napari/blob/master/napari/_qt/qt_viewer_buttons.py#L43 clean, but we could now make that pass `ndim=viewer.dims.ndim` and then rely on the default behavior inside the layer itself if all None is passed. I think I might prefer that. Thoughts @jni?

sofroniewn (commented):
Looks great. Agreed @jni we can revisit the test as we do more refactoring.

I have one suggestion about using the new functionality internally here https://github.com/napari/napari/blob/master/napari/_qt/qt_viewer_buttons.py#L43 and then dropping the additional code I flagged inside add_shapes.

I think a great follow-up PR will be to do the same for `Points` :-)

napari/components/add_layers_mixin.py
jni: I suggest making this a follow-up PR

sofroniewn: That's fine with me! I will approve this now then :-)

sofroniewn (approved):
LGTM! Great work!!

PR #1390, opened 2020-06-28
Segfault with two test passes

pwinston:
# Description
This PR is for debugging. Do not merge. It shows something we want to do for #1354 that is causing a segfault there, and it causes a segfault here. So this PR is master plus just these 3 lines:

```
@pytest.fixture(params=[0, 1], scope="session", autouse=True)
def two_passes(request):
    pass
```

That means run all the tests twice and it causes a segfault.

# segfault

It segfaults in various places, one is in `pytestqt` where it calls `app.processEvents()`:
https://github.com/pytest-dev/pytest-qt/blob/master/src/pytestqt/plugin.py#L194-L200

Where _process_events() is called by pytest_runtest_call:
https://github.com/pytest-dev/pytest-qt/blob/master/src/pytestqt/plugin.py#L194-L200

Stack is:
```
  File "/Users/pbw/miniconda3/envs/new-base/lib/python3.8/site-packages/pytestqt/plugin.py", line 197 in _process_events
  File "/Users/pbw/miniconda3/envs/new-base/lib/python3.8/site-packages/pytestqt/plugin.py", line 166 in pytest_runtest_call
  File "/Users/pbw/miniconda3/envs/new-base/lib/python3.8/site-packages/pluggy/callers.py", line 203 in _multicall
  File "/Users/pbw/miniconda3/envs/new-base/lib/python3.8/site-packages/pluggy/manager.py", line 84 in <lambda>
  File "/Users/pbw/miniconda3/envs/new-base/lib/python3.8/site-packages/pluggy/manager.py", line 93 in _hookexec
  File "/Users/pbw/miniconda3/envs/new-base/lib/python3.8/site-packages/pluggy/hooks.py", line 286 in __call__
  File "/Users/pbw/miniconda3/envs/new-base/lib/python3.8/site-packages/_pytest/runner.py", line 217 in <lambda>
```

### DIFF
--- a/napari/components/_tests/test_multichannel.py
+++ b/napari/components/_tests/test_multichannel.py
@@ -1,9 +1,12 @@
-import numpy as np
+import copy
+
 import dask.array as da
+import numpy as np
+import pytest
+
 from napari.components import ViewerModel
 from napari.utils.colormaps import colormaps, ensure_colormap_tuple
 from napari.utils.misc import ensure_sequence_of_iterables, ensure_iterable
-import pytest
 
 base_colormaps = colormaps.CYMRGB
 two_colormaps = colormaps.MAGENTA_GREEN
@@ -87,6 +90,8 @@ def test_multichannel(shape, kwargs):
     viewer = ViewerModel()
     np.random.seed(0)
     data = np.random.random(shape or (15, 10, 5))
+    # Copy because we pop, in case multiple test passes
+    kwargs = copy.deepcopy(kwargs)
     channel_axis = kwargs.pop('channel_axis', -1)
     viewer.add_image(data, channel_axis=channel_axis, **kwargs)
 
--- a/napari/conftest.py
+++ b/napari/conftest.py
@@ -291,3 +291,8 @@ def irregular_images():
 @pytest.fixture
 def single_tiff():
     return [image_fetcher.fetch('data/multipage.tif')]
+
+
+@pytest.fixture(params=[0, 1], scope="session", autouse=True)
+def two_passes(request):
+    pass
--- a/napari/utils/_tests/colors_data.py
+++ b/napari/utils/_tests/colors_data.py
@@ -72,18 +72,23 @@ single_colors_as_array = [
     np.zeros((1, 4), dtype=np.float32),
 ]
 
+
+# Use lambda as factory to produce a generator, otherwise the first pass of
+# the test would consume it.
 two_color_options = [
     ['red', 'red'],
     ('green', 'red'),
     ['green', '#ff0000'],
     ['green', 'g'],
-    ('r' for r in range(2)),
+    lambda: ('r' for r in range(2)),
     ['r', 'r'],
     np.array(['r', 'r']),
     np.array([[1, 1, 1, 1], [0, GREENV, 0, 1]]),
     (None, 'green'),
     [GREENARR[0, :3], REDARR[0, :3]],
 ]
+
+
 # Some of the options below are commented out. When the bugs with
 # vispy described above are resolved, we can uncomment the lines
 # below as well.
--- a/napari/utils/_tests/test_color_to_array.py
+++ b/napari/utils/_tests/test_color_to_array.py
@@ -41,7 +41,9 @@ def test_warns_but_parses():
     "colors, true_colors", zip(two_color_options, two_colors_as_array)
 )
 def test_twod_points(colors, true_colors):
-    np.testing.assert_array_equal(true_colors, transform_color(colors))
+    # If callable it's a factory for our value.
+    expanded = colors() if callable(colors) else colors
+    np.testing.assert_array_equal(true_colors, transform_color(expanded))
 
 
 @pytest.mark.parametrize("color", invalid_colors)


tlambert03:
Fun!  :)
I’ll pull and try it tomorrow 
Does it segfault immediately or in the same test each time?

tlambert03:
Looks like not the same test,  but always right around the same point.  I’d look just before it for clues.  Some of the “play” tests have threading ... Id try skipping some of those tests (and then bigger chunks of tests) to see if you can narrow it down

pwinston:
Yeah it moves around. But I've never seen it not segfault, and it's on all platforms. I'll be travelling tomorrow but I can look at it Monday. But yeah if you have any ideas! I wonder if there's some sort of sanitizing or checking thing we can do as a debugging approach. Like run through all the Qt objects or something. If we are stumped I mean.

tlambert03:
after running a number of combinations of pytest, selecting folders and/or ignoring folders with `--ignore`, here's a couple observations:

- when it faults, it _usually_ (but not always, depending on the exact tests) faults right around the 50% point, when the tests are just starting to run for the second time.  This suggests that one of our commonly used things is not entirely pure, and leaves something in a state that can't be run again.  it's pretty hard to track down though, since it's not just one tests that's failing
- similarly, some of the tests fail (but don't fault) when run a second time.  that's a separate issue, but will need to be cleaned up. as well

oddly enough, I _think_ I can't get this test to fault... even though it's all (?) of our tests:
`pytest napari/_tests napari/layers/ napari/components/ napari/_qt/ napari/utils/ napari/_vispy/ napari/resources/ -v`

pwinston:
I tracked down one of the failing tests, not the crash. So we'd have to fix this eventually. The test is `napari/utils/_tests/test_color_to_array.py::test_twod_points`. The test uses this data:
```
two_color_options = [
    ['red', 'red'],
    ('green', 'red'),
    ['green', '#ff0000'],
    ['green', 'g'],
    ('r' for r in range(2)),
    ['r', 'r'],
    np.array(['r', 'r']),
    np.array([[1, 1, 1, 1], [0, GREENV, 0, 1]]),
    (None, 'green'),
    [GREENARR[0, :3], REDARR[0, :3]],
]
```
The problem is `('r' for r in range(2))` is a generator. It expands to `['r', 'r']` which is the very test value, so I assume to the goal was specifically to test a generator. But as a generator it gets consumed the first time the test is run, so it's empty the second time.

I tried making `two_color_options` a function that returns a list, instead of a list. So it would return a generator each time. But pytest only calls it once during collection. So not really sure how to fix except just eliminate the generator test case. Anyway it's a concrete example of a how a test could fail when called twice.

tlambert03:
Good catch.  Did you try running the line I pasted above?  Does it crash for you?  I ran it about 5 times without a crash, then ran “pytest” and it crashed the first time. Don’t understand it, unless it was a strange race condition coincidence

pwinston:
I didn't try that yet but I wrote script that ran every test 10 times. That is, I called pytest 630 times. I used a fixture to run the test 10 times in a row during that one invocation of pytest. These were the results:
```
-6 napari/_qt/_tests/test_qt_plugin_list.py::test_qt_plugin_list
1 napari/_qt/layers/_tests/test_qt_image_base_layer_.py::test_changing_model_updates_view
1 napari/_qt/layers/_tests/test_qt_image_base_layer_.py::test_range_popup_clim_buttons
1 napari/components/_tests/test_multichannel.py::test_multichannel
1 napari/utils/_tests/test_color_to_array.py::test_twod_points
1 napari/utils/_tests/test_key_bindings.py::test_handle_multiple_keymap_providers
```
-6 means SIGABORT and 1 means the test failed. 

I used `pytest --collect-only -q > all_tests.txt` to get the list of tests and then `awk 'BEGIN {FS="["} ; {print $1}' < all_tests.txt | uniq > unique_tests.txt` to strip out the brackets (parameterized tests) and get just the unique tests. There were 630 like:
```
$ head unique_tests.txt
napari/_tests/test_viewer.py::test_no_qt_loop
napari/_qt/_tests/test_qt_buttons.py::test_radio_button
napari/_qt/_tests/test_qt_buttons.py::test_push_button
napari/_qt/_tests/test_qt_console.py::test_console
```
Then I used a fixture like in this PR but had it do 10 passes. 
```
@pytest.fixture(params=list(range(10)), scope="session", autouse=True)
def two_passes(request):
    pass
```
So these may or may not lead to the segfault but these 6 tests need to be fixed. I can look at them Monday.

pwinston:
The ABORT in `test_qt_plugin_list` is not from running the test multiple times, it happens even if I just run the test once. So I must have something wrong locally. Not sure what. But this probably unrelated.
```
Current thread 0x00000001043b7dc0 (most recent call first):
  File "/Users/pbw/dev/napari/napari/_qt/qt_main_window.py", line 446 in _update_palette
  File "/Users/pbw/dev/napari/napari/_qt/qt_main_window.py", line 98 in __init__
  File "/Users/pbw/dev/napari/napari/viewer.py", line 99 in __init__
  File "/Users/pbw/dev/napari/napari/conftest.py", line 80 in actual_factory
  File "/Users/pbw/dev/napari/napari/_qt/_tests/test_qt_viewer.py", line 58 in test_add_layer
  File "/Users/pbw/miniconda3/envs/new-base/lib/python3.8/site-packages/_pytest/python.py", line 182 in pytest_pyfunc_call
```

tlambert03:
Nice!  Does this approach of yours run a single test 10 times in isolation though?
I fear that won’t catch combinatorial problems... and i suspect that’s what we have here (something in Test A that causes a fault in test B when run the second time)

The line I was asking you to test is particularly interesting because (if you can repeat it) it shows that we can run pretty much all of our tests twice if we specify them directly. And it’d be interesting to figure out what the pytest difference is

pwinston:
Set back down to 2 passes your line segfaulted for me:

```
napari/_tests/test_dask_layers.py::test_dask_array_creates_cache[1] FAILED              [ 50%]
napari/_tests/test_dask_layers.py::test_list_of_dask_arrays_creates_cache[1] PASSED     [ 50%]
napari/_tests/test_dask_layers.py::test_dask_optimized_slicing[1] PASSED                [ 50%]
napari/_tests/test_dask_layers.py::test_dask_unoptimized_slicing[1] FAILED              [ 50%]
napari/_tests/test_dask_layers.py::test_dask_cache_resizing[1] PASSED                   [ 51%]
napari/_tests/test_dask_layers.py::test_prevent_dask_cache[1] PASSED                    [ 51%]
napari/_tests/test_draw.py::test_canvas_drawing[1] Fatal Python error: Segmentation fault
```

Yeah I planned to do some type "search" with running different subsets of test. But running each in isolation was simpler and seemed worth doing. I thought maybe the 10x might find something. Really we should fix all these because any failure means something unexpected is happening. But yeah it might be the segfault requires an interaction.

Not sure about my `test_qt_plugin_list` aborting. My `napari-plugin-engine` needed upgrading but it didn't help. Must be something local to me though because it fails the first time for me. So unrelated to all this I think.

tlambert03:
ok, thanks for testing.  strange that I got about 5-6 successful tests with that, and then immediately segfaulted when I went back to bare `pytest`

pwinston:
First I tried binary search to find a smaller range of tests that segfaults but it didn't get very far:
```
{"return_code": -11, "start": 0, "end": 630, "num_tests": 631}
{"return_code": 1, "start": 0, "end": 315, "num_tests": 316}
{"return_code": -11, "start": 315, "end": 630, "num_tests": 316}
{"return_code": 0, "start": 315, "end": 472, "num_tests": 158}
{"return_code": 1, "start": 472, "end": 630, "num_tests": 159}
```
I repeated it and it was identical. So I tried thirds:
```
{"return_code": -11, "start": 0, "end": 630, "num_tests": 631}
{"return_code": 1, "start": 0, "end": 210, "num_tests": 211}
{"return_code": 0, "start": 210, "end": 420, "num_tests": 211}
{"return_code": 1, "start": 420, "end": 630, "num_tests": 211}
```
Where -11 means segfault, 1 means failure, 0 means success. 

Playing around with other ranges I found things were intermittant:
```
{"return_code": -11, "start": 525, "end": 630, "num_tests": 106}
{"return_code": -11, "start": 525, "end": 630, "num_tests": 106}
{"return_code": 1, "start": 525, "end": 630, "num_tests": 106}
{"return_code": 1, "start": 525, "end": 630, "num_tests": 106}
{"return_code": -11, "start": 525, "end": 630, "num_tests": 106}
```
So that's showing [525..630] segfaulting 3/5 times.

Searching by hand smallest range I sound that sometimes segfaults is:

```
{"return_code": -11, "start": 550, "end": 570, "num_tests": 21}
{"return_code": -11, "start": 550, "end": 570, "num_tests": 21}
{"return_code": -11, "start": 552, "end": 570, "num_tests": 19}
{"return_code": -11, "start": 552, "end": 570, "num_tests": 19}
{"return_code": -11, "start": 552, "end": 570, "num_tests": 19}
{"return_code": -11, "start": 552, "end": 570, "num_tests": 19}
{"return_code": -11, "start": 552, "end": 570, "num_tests": 19}
{"return_code": -11, "start": 552, "end": 570, "num_tests": 19}
```
where [552..570] are:
```
napari/plugins/_tests/test_builtin_write_layers.py::test_no_write_layer_bad_extension
napari/plugins/_tests/test_builtin_write_layers.py::test_write_layer_no_metadata
napari/plugins/_tests/test_exceptions.py::test_format_exceptions
napari/plugins/_tests/test_hook_specifications.py::test_hook_specification_naming
napari/plugins/_tests/test_hook_specifications.py::test_docstring_on_hook_specification
napari/plugins/_tests/test_hook_specifications.py::test_annotation_on_hook_specification
napari/plugins/_tests/test_hook_specifications.py::test_docs_match_signature
napari/plugins/_tests/test_pypi.py::test_get_packages_by_prefix
napari/plugins/_tests/test_pypi.py::test_get_package_versions
napari/plugins/_tests/test_reader_plugins.py::test_builtin_reader_plugin
napari/plugins/_tests/test_reader_plugins.py::test_builtin_reader_plugin_csv
napari/plugins/_tests/test_reader_plugins.py::test_builtin_reader_plugin_stacks
napari/plugins/_tests/test_save_layers.py::test_save_layer_single_named_plugin
napari/plugins/_tests/test_save_layers.py::test_save_layer_no_results
napari/plugins/_tests/test_save_layers.py::test_save_layer_single_no_named_plugin
napari/plugins/_tests/test_save_layers.py::test_save_layer_multiple_named_plugin
napari/plugins/_tests/test_save_layers.py::test_save_layer_multiple_no_named_plugin
napari/plugins/_tests/test_save_layers.py::test_default_save_order
napari/resources/test_resources.py::test_resources
```
Just by inspection it _seems_ like if `napari/plugins/_tests/test_reader_plugins.py::test_builtin_reader_plugin[0]` pauses while it's running then often it will later crash. And sometimes when it does _not_ crash it prints:
```
napari/plugins/_tests/test_reader_plugins.py::test_builtin_reader_plugin[0]
  /Users/pbw/dev/napari/napari/conftest.py:69: UserWarning: Widgets leaked!: {<vispy.app.backends._qt.CanvasBackendDesktop object at 0x7fee08539a60>}
    warnings.warn(f'Widgets leaked!: {leaks}')
```
So I wonder if `test_builtin_reader_plugin[0]` is something leaking this widget but then only sometimes it leads a crash.

pwinston:
What exactly does the **Widgets leaked** warning mean? Does it indicate a serious error for sure or does it just happen sometimes?

pwinston:
Okay giving up doing a contiguous range of test and just trying different combinations these 3 tests segfault maybe 1/3rd of the time when run together like this:
```
pytest napari/plugins/_tests/test_reader_plugins.py::test_builtin_reader_plugin 
       napari/plugins/_tests/test_reader_plugins.py::test_builtin_reader_plugin_csv 
       napari/resources/test_resources.py::test_resources
```
That's a small enough test case I think need to turn the corner and dive into what they are actually doing. This is all with 2 passes of course.

tlambert03:
> What exactly does the Widgets leaked warning mean? Does it indicate a serious error for sure or does it just happen sometimes?

If it happens in addition to some other error in the same test, you can ignore it. Usually means failed cleanup.

If it’s the only error, it usually means that cleanup is incomplete (like a widgets wasn’t given a parent, or something like that).

But in either case, it’s not a serious problem.  And we didn’t use to check it at all.  Just means some “top level widgets” remain in the Qt app after everything has been cleaned up.  Good idea to find and solve the problem, but probably wouldn’t affect the average app experience

pwinston:
Printing out `id(app)` in `Viewer.__init__` I see all the unit tests are sharing the same `QApplication`. However I read that `qtbot` has a `pytest` scope of `function`. So you'd hope everything was started fresh for each test. But maybe `QApplication` is special and it basically cannot exit?

If `QApplication` has to live on I think that means the exit/cleanup code has to absolutely perfect, or the `QApplication` could easily have state that lives from one test to the next? 

I don't see any other threads running. Threads calling into GUI/Qt code seems like a common source of crashes, but I don't think that's a suspect here, unless it's under the hood somewhere. It crashes in different places but one I've seen a few times is on the line `self.setStyleSheet(themed_stylesheet)` in `QtViewer._update_palette`. That seems really harmless, there's nothing to that call that seems like it could be wrong. So it seems like things must be in very screwed up state for that to crash.

I seems like somehow we need to have a fresh `QApplication` or exit/clean needs to be perfect. And PyQt/Qt is not very helpful with ensuring cleanup is done right, in that at least with our situation it just segfaults.

We could punt and not run two passes within one call to pytest, instead invoke pytest twice like `pytest napari --sync_only` and `pytest napari --async_only`. That's not so horrible really. Although it might imply we are papering over something we are doing wrong that will surface in other places, possibly for a user who creates multiple viewers sequentially.

I did see one person who mentioned segfaults who had a very complicated way to exit that involved posting messages to itself. Kind of like turn things off, post message to shutdown, presumably handle events "during shutdown", then shutdown. Kind of like flushing out events. Although you'd think PyQT/Qt would already do that.

pwinston:
It seems notable we have no problems during the 1st pass even though we create and close many viewers. If closing a viewer leaves things in a bad state, none of the tests during the 1st pass seem to notice this bad state, no matter many how times you run it.

So I think one question is does running the 2nd pass just give this "more chances to fail"?  Or is there something very specific about re-running a test that was run before which is problematic. As in, the first pass does something that you simply cannot safely redo. Because maybe the first time it did something to the `QApplication`?

pwinston:
I found out `QtDims.closeEvent()` was not being called:
https://github.com/napari/napari/blob/master/napari/_qt/qt_dims.py#L405-L408

I put `self.dims.close()` is QtViewer.closeEvent()` then it does get called. Doesn't fix anything, but I wonder If there are other things like that. With C++ and memory there are tools to look for memory leaks. With PyQt I'm not sure is there anyway to know things are being cleaned up properly besides crashes?

pwinston:
I'm tempted to put this on hold and return to #1354 and in that PR create a script that runs `pytest napari --sync_only` and then `pytest napari --async_only`. A user that wants to run **all** tests has to run that tiny script instead of calling bare pytest. And the CI will have to run that script or otherwise call pytest twice.

Obviously we want to fix this. Having bare `pytest` run all tests should be the goal. But this is some sort of pre-existing condition that potentially goes back a very long time. It's totally unrelated to async, since it happens in this PR which contains no async code. I don't feel like we have any good theories or debugging techniques right now, and have been looking at this for 4 days. The async feature itself doesn't care if we have to run pytest once or twice. it will be just as useful either way.

pwinston:
As for this PR maybe we need to document exactly what a well-behaved PyQt/Qt application needs to do as far as memory management and cleanup. If we don't have that written out already. And then document how do we ensure napari is doing those things, whether by inspection or with tools. Including how we make sure new changes don't regress on these things.

Assuming what's causing this segfault is some type of mis-behavior on our part.

Issue #790, opened 2019-12-13
How do I update the x/y axis in a VisPy plot using set_data()


cudmore:
I am creating a Napari plugin that has an external Qt window which includes a VisPy plot in a QHBoxLayout. See [my plugin](http://blog.cudmore.io/ShapeAnalysisPlugin/)

The initial VisPy plots are working fine but when I try to update the plot with set_data(), the axis are not scaling to the new data values? Not sure if this is Napari or VisPy but your help would be welcomed.

Here is some working code where I bring up the interface and then set some new data with set_data() and the axis do not update. If I mouse scroll out I can eventually see the new plotted points...

```
import numpy as np
import napari
from PyQt5 import QtWidgets
import vispy.plot as vp

class TryToUpdatePlot(QtWidgets.QWidget):
	def __init__(self):
		super(TryToUpdatePlot, self).__init__()
		viewer = napari.Viewer()
		self.initUI()

	def initUI(self):
		fig = vp.Fig(size=(800, 600), show=False)
		# random data
		data = np.random.rand(100, 2)

		line1 = fig[0, 0].plot(data=data)
		line2 = fig[1, 0].plot(data=data)

		self.myHBoxLayout = QtWidgets.QHBoxLayout(self)

		self.myHBoxLayout.addWidget(QtWidgets.QPushButton('A Button'))
		self.myHBoxLayout.addWidget(fig.native)

		self.show()

		# update line1 plot with new data offset by 100
		data = np.random.rand(100, 2) + 100
		# this call to set_data does not update the range of x/y axis
		# it remains at the original scale ???
		line1.set_data(data)

		# Is there something I can add here ???
		#line1.update()
		#line1.view.camera.set_range()

if __name__ == '__main__':
	with napari.gui_qt():
		test = TryToUpdatePlot()

```

jni:
@cudmore Very cool!!

It definitely seems to be a vispy issue more than a napari one, but I wonder: have you tried making *just* a vispy plot and using `set_data()`? Does that update?

In short, my advice would be to work on the *minimal* example that demonstrates your issue, because that might help us get some vispy developers on board with helping you. Personally I don't know immediately what would help, or whether this is even broken or expected behavior.

tlambert03:
FWIW, this is exactly the kind of plot (with axis linking) that #675 is working on

tlambert03:
Hi @cudmore, in the meantime, you need to add something like this to your code to get it working (right where you have `# Is there something I can add here ???`)
```
[...]

Issue #1380, opened 2020-06-22
layerlist doesn't always behave like a list, not always synced with viewer


tlambert03:
## 🐛 Bug
our `LayerList` inherits from python List, and has various methods (append, pop, etc...) that are list like, but not all list methods work in synchrony with the viewer:

```python
In [3]:viewer.layers.clear()

In [4]:viewer.layers
Out[4]: []
# but all the layers are still in the viewer
```

I feel like I've seen others in the past, but can't remember at the moment.

## Expected behavior

LayerList methods should propagate to the current layers in the viewer.  and viewer.layers should always match what's in the viewer.

sofroniewn:
I think this comes down to the events not emitting as expected. We use the `added`, `removed`, `reordered` events in the `qt_layerlist` here https://github.com/napari/napari/blob/master/napari/_qt/qt_layerlist.py#L64-L66 and the `qt_viewer` here https://github.com/napari/napari/blob/master/napari/_qt/qt_viewer.py#L185-L187 and the  `changed` event in the `viewer_model` here https://github.com/napari/napari/blob/master/napari/components/viewer_model.py#L87-L89.

My guess is that the event emission isn't happening as expected for all methods and that is what is causing problems. (i.e. if we test all the events always emit right then we should be ok).

There was another layer list bug and to do list while back in #312 which still needs to be addressed.

tlambert03:
thanks @sofroniewn, makes sense.  Since the main thing I was trying to do is clear the viewer, I'll add a couple more potential "gotchas" in case anyone finds this.  These are related to problems with iterating through a list that you are altering:

doesn't work:
```python
for layer in viewer.layers:
    viewer.layers.remove(layer)
```
```pytb
Traceback (most recent call last):
  File "test.py", line 10, in <module>
    for layer in viewer.layers:
  File "/napari/napari/utils/list/_base.py", line 25, in __iter__
    yield self[i]
  File "/napari/napari/utils/list/_base.py", line 28, in __getitem__
    indices = self.__prsitem__(key)
  File "/napari/napari/utils/list/_multi.py", line 11, in __prsitem__
    return super().__prsitem__(keys)
  File "/napari/napari/utils/list/_base.py", line 80, in __prsitem__
    return self.__locitem__(key)
  File "/napari/napari/utils/list/_typed.py", line 37, in __locitem__
    return super().__locitem__(key)
  File "/napari/napari/utils/list/_base.py", line 111, in __locitem__
    raise IndexError(
[...]

Issue #312, opened 2019-05-30
fix TypedList error throwing and add comprehensive test suite


kne42:
- `List.__contains__` should handle all exception types
- `TypedList.index` should throw a `ValueError` instead of a `KeyError`
- comprehensive test suites should be added for all custom lists

as brought up in https://github.com/napari/napari/pull/306#issuecomment-497146532

sofroniewn:
Note this no longer be needed after #1444 and followup PRs using the new evented list

PR #1391, opened 2020-06-29
Use qlistwidget for QtLayerList

sofroniewn:
# Description
It's been pointed out in numerous places that our QtLayerList implementation is very complex right now as it implements a lot of custom drag/drop and scroll functionality ourself. This could make it very hard when we come and try and represent a TreeView for LayerGroups as proposed in #790. This PR aims to use a QListWidget instead and reduce the amount of custom code.

It is very WIP right now, lots of functionality is still missing, but I am able to display the widgets and drag and drop works on the view side. I need to do some linking with our model code, but that can come later. I might even wait until after #1376 merges and do this in the new EVH style :-)

Also styles will need to be tweaked again, but here is what it looks like right now:

Screen Shot 2020-06-28 at 4 59 47 PM

## Type of change

## Final checklist:


tlambert03:
Do you imagine this will be usable after layer groups?  Or just feel like the need is immediate enough to do it anyway?

sofroniewn:
> Do you imagine this will be usable after layer groups? Or just feel like the need is immediate enough to do it anyway?

Combination of immediate need (more maintable code, will make EVH refactor easier), and thinking the transition to layer groups will be much easier from the `QListWidget` to the `QTreeWidget` rather than from our custom implementation.

What's your take on this?

tlambert03:
I think things will be learned, and it can’t hurt to start working on it if you’re motivated.  But I think a fair amount will change with groups.  I suspect we’ll use qtreeview and qabstractitemmodel instead of qtreewidget.  (That doesn’t mean we won’t be able to borrow some code from a new layerlist widget though)

jni:
Very nice @sofroniewn! In that screenshot it's quite a bit uglier than what we have now imho, but I suspect as you say that's just some style tweaking needed. I'm totally agnostic as to whether this is better than a direct path from custom to trees.

tlambert03:
note from meeting: don't forget about `napari._qt.utils.drag_with_pixmap`.  example usage in `_qt.qt_plugin_sorter.QtHookImplementationListWidget.startDrag`

sofroniewn:
The reordering is proving pretty tricky - see discussion here https://github.com/napari/napari/pull/1394#issuecomment-653623618, I might need to enlist some help from @tlambert03 and @jni.

We should probably also think about fixing as much of #1380 as we can. The layerlist is a pretty key object, and it's not intuitive how to do basic things with it like reorder/ remove, see people asking questions like https://twitter.com/kD3AN/status/1279086938650218497

[Edit] I think step 1 would be to make sure `ListModel` is rock solid and fully tested (see #312 too) and then worry about the rest

tlambert03:
I don't want to sound like a broken record, but I really do think a lot of this could be improved with LayerGroups... not just because of the layergroup itself, but because we would be using Qt's model/view architecture properly ... (wherein we provide the model, and use a Qt View, rather than using QWidgets, which come with their own models included.)

in case you're curious, here's a very rough implementation that I've been playing with for a QTreeModel that wraps our layergroup model:
https://github.com/tlambert03/napari/blob/layergroups-tjl/napari/_qt/qt_layertree.py#L8

the main idea here is that you provide concrete a implementation (called `QtLayerTreeModel` in that example) for an _interface_ that Qt declares.  (Their view classes are designed to work with their model interfaces... but you don't need the model to actually be an internal Qt object).  Then when our internal python model changes, you just need to tell the view to refresh itself, and it will pull data directly from the python object (via the concrete `QtLayerTreeModel` interface object).  And vice versa: if the user changes something in the view, the model interface directly updates the layergroup object (whereas the `QLayerList` first updates its own internal model... and then we have to sync our external layerlist model)

sofroniewn:
We could do take this approach right now with the `QListView` no? Instead of waiting for the layer groups and the `QTreeView`. I really don't think we want to try and make this transition all in one go and think it will still be a while before we have layer groups figured out.

I can see how using the Qt View/Model instead of the Widget might make things easier in this case, but the problem effects every single QWidget we have that can both update and be updated.

I'm also struggling with the reordering of just out custom `ListModel` - i.e. no Qt stuff at all. I've started adding some `ListModel` tests (there were none before!). Hopefully we can get that figured out first

tlambert03:
> We could do take this approach right now with the `QListView` no?

yep!

sofroniewn:
I've made progress with emitting events from the `ListModel`, but I am back to problems with the `QListWidget`, particularly because we're trying to style it with a custom widget - `setItemWidget` which comes with this rather ominous warning about only being for static lists - see here https://doc.qt.io/qt-5/qlistwidget.html#setItemWidget.

While I'm able to get drag and drop working if I try and programmatically reorder the widget nothing works as expected, and using `setItemWidget` again in the way I would have liked results in segfaults. Their recommended approach is the `QListView` approach, so we'll see if I start going down that path tomorrow!!

sofroniewn:
Ok @tlambert03 - looks like our hand will really be forced here by limitations of `setItemWidget` and that to get any reordering programatically we need to go down the `QListView` road as mentioned in the docs link above (see this [stackoverflow post](https://stackoverflow.com/questions/27332310/items-in-a-qlistwidget-turn-invisible-after-reordering) and this thread on the [Qt bug list](https://bugreports.qt.io/browse/QTBUG-16386?focusedCommentId=136782&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel) to indicate that I really don't think we can achieve what we want with just a `QListWidget`

I'm a little worried about the added complexity of the `QListView` + `QAbstractItemModel` approach as we'll also have our own `LayerList` model that I really do want to keep off the `QAbstractItemModel`. We'll still be maintaining two separated models, but I guess we will have more flexibility now that the actual Qt one is decoupled from its view.

I might start down this road tomorrow, we'll see. Let me know if you have further thoughts. I'd love for us to get this wrapped up soon!!!

sofroniewn:
I'm curious @tlambert03 if you think we can use https://doc.qt.io/QT-5/qabstractitemmodel.html#moveRow and stick with the `QListWidget`? I'm going to look into it in the morning, might make it easier if we could!!

tlambert03:
> I'm a little worried about the added complexity of the QListView + QAbstractItemModel approach as we'll also have our own LayerList model that I really do want to keep off the QAbstractItemModel.

not sure I follow here...  Are you saying that you don't want the `QAbstractItemModel` to have awareness of the the `LayerList` object?

tlambert03:
> We'll still be maintaining two separated models

that's just it... you _won't_ be.  don't think of `QAbstractItemModel` as an independent model from `LayerList` (and by model here, I mean source of truth).  Think of it as a proxy object for the `LayerList` that provides the appropriate interface for the Qt View classes.  It doesn't need to actually hold data.  (it _can_ if you want to, but it doesn't need to).  For this to work of course, the `QAbstractItemModel` does need to know about the `LayerList` object.  But something needs to know about something, and it's better that way (IMO) than having two separate models (sources of truth) that need be syncronized.

I do agree it seems more complex than the widget classes provided by Qt.  But from my reading on this, so far I _think_ this is the "proper" way to do create a Qt View onto a complex python object without duplicating models.

sofroniewn:
> Are you saying that you don't want the QAbstractItemModel to have awareness of the the LayerList object?

I think that would be nice, and that we should try and keep as strong a separation as possible between the Qt code and our own models and have them communicate only via our event handler, with the goal of improving the modularity and separation of concerns in the codebase.

I'm still struggling to get the reordering stuff working, but that's because of Qt limitations, nothing to do with the EVH stuff ... but we'll see how it all goes ...

sofroniewn:
> For this to work of course, the QAbstractItemModel does need to know about the LayerList object. But something needs to know about something, and it's better that way (IMO) than having two separate models (sources of truth) that need be syncronized.

Yeah I guess this is the tradeoff. I'd still rather try and keep our Qt stuff separated, especially as the list views only need the name, thumbnail, visibility and selected status - i'd rather avoid giving them access to everything, but let's see how it goes!!

tlambert03:
> I think that would be nice, and that we should try and keep as strong a separation as possible between the Qt code and our own models and have them communicate only via our event handler,

for me, keeping Qt separated means that our models shouldn't know about Qt... But it doesn't mean that Qt stuff shouldn't know about our models.  In most MVC designs, the view has to see the model.  The `QtLayerList` here is a view into the `LayerList` model.  I don't think the event handler is anywhere close to being "rich" enough to provide everything necessary to the Qt objects.  And even if it were, it necessitates duplicated sources of truth.

sofroniewn:
Ok, I've made big progress here, all functionality is now working and all tests passing with minimal changes (no functionality dropped, some name changes, and some extra tests added!!). I was able to get it all working fairly elegantly with the `QListWidget` the only things I had to do that might appear slightly unnatural were leverage the `item.text` property for sorting and overwrite the `dropEvent` and manually determine the parameters of the reorder call, but neither of those are too bad.

One thing I can't figure out is how though is the `test_qt_layerlist.py` is still leaking widgets. Note this test doesn't create the full viewer, it only creates the `QtLayerList` for testing in isolation, so I thought I'd be able to just call `close` and it would be fine. Maybe @tlambert03 can take a look at that one.

I still have to work on stylings, I will give it a first pass in a couple hours and then maybe turn this over to @tlambert03 for further improvements. 

I think it will be good to try and get this is in relatively as is, and then we can make further improvements when we go to tree views etc.

I'll post when this is ready for through review, for now though just help fixing the leaking widget would be great

sofroniewn:
This PR is now ready for full review. The stylings have been fixed, though feedback, adjustments welcome. Play around with the drag/ drop too and let me know if you like it.

Screen Shot 2020-07-05 at 11 26 10 AM

and 

Screen Shot 2020-07-05 at 11 28 10 AM

with lots of layers, which I think is quite nice.

A couple key points are:
- all the custom qt layer list stuff is now gone, we use the qlistwidget instead, which is much cleaner.

- we now use the event handler on our `ListModel` itself (and so the `LayerList` too).

 - grid mode is still broken, see #1400, but I'd like to fix that in its own PR after this merges.

- I made some improvements to `active_layer` but that is still not user settable. I'm thinking it belongs on the layer list in the future too, like how [QListWidget has a current row property](https://doc.qt.io/qt-5/qlistwidget.html#currentRow-prop) in addition to the selected properties, but that fix can come later. These changes don't make it worse.

- I would like to drop the `changed` event from the `ListModel` in the upcoming world coords stuff, but as it was there before no need to drop it now.

- I'm still leaking widgets in the `test_qt_layerlist.py` test and need help with that.

Looking forward to feedback from @jni @tlambert03 @GenevieveBuckley

tlambert03:
> One thing I can't figure out is how though is the test_qt_layerlist.py is still leaking widgets. Note this test doesn't create the full viewer, it only creates the QtLayerList for testing in isolation, so I thought I'd be able to just call close and it would be fine. Maybe @tlambert03 can take a look at that one.

looks like the reference to the Widget in the EventHandler is causing this.  comment out the `register_listener` bit and it's fine.  One potential fix is to add `weakref`s instead of direct references... for instance, this also works in the `EventHandler`
```python
import weakref

...

    def register_listener(self, component):
        self.components.append(weakref.ref(component))

    ...

    def on_change(self, event=None):
        ...
        # Update based on event value
        for componentref in self.components:
            component = componentref()
            if component is not None:

```

**EDIT**: this of course has a bit of a bad smell... but I'm not immediately sure how to handle it.  as is, the event handler is gradually going to maintain pointers to most objects in the program, so this will certainly pop up again

jni:
@sofroniewn before I review the code, here are some notes from our Zulip chat about potential usability improvements/issues:

- [x] dragging a layer into the trashcan is allowed but crashes napari [1]
- [ ] dragging a layer to the top or bottom of the list is finicky, as it's a small drop space and there's little visual indication of whether it's going to work. All of the bottom space should be available to drop. My guess is that the *most* used drag/drop will be to drag to the top, so that should be easy to do.
- [ ] dropping in between two layers is tricky. There's a few issues, and it's hard to tell a priori which combination of the following suggested fixes will give the best user experience:
  * the insertion marker is hard to see under the layer being dragged. It would be excellent to have it be more prominent, and extend all the way out under the thumbnail, maybe even a bit past it if there's room for that.
  * the layer being dragged obstructs the insertion marker most of the time, so the layer marker opacity should be decreased while that layer is being dragged.
  * the insertion location is determined by the location of the cursor rather than the location of the dragged glyph. This one I'm not sure about but it's worth experimenting.
- [x] it's finicky to find a grab point, because click and drag over the text field results in click-through and text selection. Click and drag over the text should drag the layer unless the text box has been activated with a click (similar to the macOS Finder).

Ok time for code review now. =)

[1]:

<details>

<summary>
napari crash when adding a points layer then dragging it to the trash icon
</summary>

```pytb
 $ python examples/add_image.py 
  File "/home/jni/projects/napari/napari/_qt/qt_viewer_buttons.py", line 188, in dropEvent
    layer = self.viewer.layers[layer_name]
  File "/home/jni/projects/napari/napari/utils/list/_base.py", line 28, in __getitem__
    indices = self.__prsitem__(key)
  File "/home/jni/projects/napari/napari/utils/list/_multi.py", line 11, in __prsitem__
    return super().__prsitem__(keys)
  File "/home/jni/projects/napari/napari/utils/list/_base.py", line 80, in __prsitem__
    return self.__locitem__(key)
  File "/home/jni/projects/napari/napari/utils/list/_typed.py", line 36, in __locitem__
    key = self.index(key)
  File "/home/jni/projects/napari/napari/utils/list/_typed.py", line 82, in index
KeyError: 'could not find element  was referencing'
Aborted (core dumped)
```

</details>

napari/_tests/utils.py
```diff
@@ -151,3 +151,26 @@ def check_view_transform_consistency(layer, viewer, transf_dict):
             # expected translate values
             correct_vals = np.add(transf[disp_dims], tile_transf)
         assert (vis_vals == correct_vals).all()
+
+
+def assert_layout_layers(widget, layers):
```
jni: I found this name really non-descriptive of what it does. How about `assert_layer_model_synced`?

napari/_qt/qt_layer_widget.py
```diff
@@ -0,0 +1,215 @@
...
+        self.visibleCheckBox = cb
+        self.layout.addWidget(cb)
+
+        textbox = QLineEdit(self)
+        textbox.setAcceptDrops(False)
+        textbox.setEnabled(True)
+        textbox.editingFinished.connect(self.changeText)
+        self.nameTextBox = textbox
+        self.nameTextBox.old_name = ''
+        self.layout.addWidget(textbox)
```
jni: I guess this is where you'd need to alter the "click-through" behaviour...

napari/_qt/qt_layer_widget.py
```diff
@@ -0,0 +1,215 @@
...
+        ltb.setProperty('layer_type_label', True)
+        ltb.setToolTip('Layer type')
+        self.typeLabel = ltb
+        self.layout.addWidget(ltb)
+
+        msg = 'Click to select\nDrag to rearrange'
+        self.setToolTip(msg)
+
+        # Once EVH refactor is done, these can be moved to an initialization
+        # outside of this object
```
jni: In these comments, can you clarify what exactly needs to be done? Does this mean when the EVH refactor is done *for all layers*?

napari/_viewer_key_bindings.py
```diff
@@ -95,30 +95,6 @@ def remove_all_layers(viewer):
...
-@Viewer.bind_key('Shift-Up')
-def also_select_layer_above(viewer):
-    """Also select layer above."""
-    viewer.layers.select_next(shift=True)
-
-
-@Viewer.bind_key('Shift-Down')
-def also_select_layer_below(viewer):
-    """Also select layer below."""
-    viewer.layers.select_previous(shift=True)
```
jni, line 119: Nice! Just tested that these still work! :tada: 

My only concern is that I found it a bit tricky to "activate" them. For example, at one point I found myself in the text editing box instead of in the layer list. I'm not sure what to do about it though, just wanted to note it down... Let me know if you encounter the same issue.

napari/components/_tests/test_layers_list.py
```diff
@@ -29,9 +29,9 @@ def test_adding_layer():
     layer = Image(np.random.random((10, 10)))
     layers.append(layer)
 
-    # LayerList should err if you add anything other than a layer
-    with pytest.raises(TypeError):
-        layers.append('something')
+    # # LayerList should err if you add anything other than a layer
+    # with pytest.raises(TypeError):
+    #     layers.append('something')
```
jni: Should these be deleted? Or?

napari/components/_tests/test_layers_list.py
```diff
@@ -243,159 +243,6 @@ def test_remove_selected():
     assert len(layers) == 0
 
 
-def test_move_selected():
-    """
-    Test removing selected layers
-    """
```
jni, line 249: Why are these tests deleted? All these things should still work, right?

napari/components/add_layers_mixin.py
```diff
@@ -50,9 +50,10 @@ def add_layer(self, layer: layers.Layer) -> layers.Layer:
         """
         layer.event_handler.register_listener(self)
 
-        layer.dims.events.ndisplay.connect(self._on_layers_change)
-        layer.dims.events.order.connect(self._on_layers_change)
-        layer.dims.events.range.connect(self._on_layers_change)
+        # These connections will be removed once the EVH reaches the Dims
```
jni, line 53: Yeah, this is a good comment. =)

napari/components/layerlist.py
```diff
@@ -67,43 +73,13 @@ def _on_name_unique_change(self, names):
...
-            indices.insert(insert_idx, elem_idx)
-        # reorder list
-        self[:] = self[tuple(indices)]
+    @selected.setter
+    def selected(self, selected):
+        """List: Indices of layers to select layers."""
+        self.events.selected_layers(selected)
 
     def unselect_all(self, ignore=None):
         """Unselects all layers expect any specified in ignore.
```
jni: except

napari/components/viewer_model.py
```diff
@@ -80,13 +80,13 @@ def __init__(
...
-        self.dims.events.order.connect(self._update_layers)
-        self.dims.events.axis.connect(self._update_layers)
-        self.layers.events.changed.connect(self._update_active_layer)
-        self.layers.events.changed.connect(self._update_grid)
-        self.layers.events.changed.connect(self._on_layers_change)
+        self.dims.events.ndisplay.connect(lambda e: self._update_layers())
+        self.dims.events.order.connect(lambda e: self._update_layers())
+        self.dims.events.axis.connect(lambda e: self._update_layers())
+
+        self.layers.event_handler.register_listener(self)
```
jni, line 89: This change might be relevant to the grid mode failures?

napari/components/viewer_model.py
```diff
@@ -347,40 +363,22 @@ def _toggle_theme(self):
...
+        Only one layer can be selected for that layer to be the  active layer.
 
-        if active_layer is None:
-            self.status = 'Ready'
-            self.help = ''
-            self.cursor = 'standard'
-            self.interactive = True
-            self.active_layer = None
+        Parmeters
+        ---------
```
jni:
```suggestion
        Parameters
        ----------
```

napari/components/viewer_model.py
```diff
@@ -436,11 +434,19 @@ def _calc_layers_num_dims(self):
...
         return max_dims
 
-    def _on_selected_change(self, selected):
-        self._update_active_layer(None)
-
     def _on_data_change(self, data):
-        self._on_layers_change(None)
+        self._update_layer_dims()
+
+    def _on_changed_change(self, value):
```
jni, line 440: Any other suggestions for this? How about `on_any_change`?

napari/utils/event_handler.py
```diff
@@ -23,7 +26,7 @@ def register_listener(self, component):
             Object that contains callbacks for specific events. These are
             methods named according to an '_on_*_change' convention.
         """
-        self.components_to_update.append(component)
+        self.components.append(weakref.ref(component))
```
jni, line 32: Maybe add a comment here linking to Talley's [comment](https://github.com/napari/napari/pull/1391#issuecomment-653939143), so that future generations are not puzzled and/or tripped up by this?

napari/utils/event_handler.py
```diff
@@ -40,15 +43,18 @@ def on_change(self, event=None):
         # until refactor on all layers is complete, not all events will have a
         # value property
         try:
+            # print(event.type, event.value, self.components)
```
jni: Remove before merging. =)

napari/utils/event_handler.py
```diff
@@ -40,15 +43,18 @@ def on_change(self, event=None):
...
             return
 
         # Update based on event value
-        for component in self.components_to_update:
-            update_method_name = f"_on_{event.type}_change"
-            update_method = getattr(component, update_method_name, None)
-            if update_method:
-                update_method(value)
+        for componentref in self.components:
+            component = componentref()
```
jni: As above — add a comment here reminding readers that `self.components` contains weakrefs rather than refs.

I agree with Talley that this doesn't smell right, but it's not too terrible either for now...

napari/utils/list/_model.py
```diff
@@ -28,38 +29,66 @@ def __init__(self, basetype, iterable=(), lookup=None):
...
+        obj = self[key]
+        self.events.removed((obj, key))
+        self.events.changed(None)
         return obj
+
+    def clear(self):
+        while len(self) > 0:
+            obj = self[0]
+            self.events.removed((obj, 0))
+        self.events.changed(None)
```
jni, line 65: The `changed` seems hacky... again, (a) can we rename to `any`, so that responding is `_on_any_change`, and (b), is there some way to somehow make all other events raise this automatically, rather than have to sprinkle `self.events.changed(None)`/`self.events.any(None)` everywhere?

jni (approved):
@sofroniewn I've left some comments. I'll approve in case you need to move on, but I'd love to see a  solution to the `changed` events!

jni:
(Also, can you make sure that all new code lines are covered? I did not check that.)

tlambert03:
some usage observations before looking at code:
- styles:
    - [x] corners of the widgets have lost their rounded, border radius
    - [x] I like that the widgets now adjust their size when a scrollbar appears, but it's rather tight.  would be nice to have some padding as before:
    Montage


- [x] selection behavior:  it used to be the case that you could drag and drop a _non-selected_ widget to reorder.  now what happens is that it highlights the text, but does not move the widget.  Furthermore that highlight persists until you select and edit the label to remove the highlight.
    **old behavior:**
    Untitled 2

    **new behavior:**
    Untitled

napari/_qt/qt_layer_widget.py
sofroniewn: Ended up having to subclass QLineEdit, see bottom of this file

sofroniewn:
Fixed selection behaviour (i.e. problem with text) and the rounded corners. Still need to fix the space between list items and scrollbar and the three other drag/ drop related ones.

sofroniewn:
Scroll bar padding now fixed too 
image

napari/_qt/qt_layer_widget.py
sofroniewn: Added a line, but details of implementation are yet to be decided, something that we should discuss on a dedicated thread (i.e. where does initialization happen)

napari/_tests/utils.py
sofroniewn: renamed `assert_layerlist_model_view_synced`

napari/_viewer_key_bindings.py
sofroniewn: i've made some improvements here. You have to click once into the layerlist and then they work great.

napari/components/_tests/test_layers_list.py
sofroniewn: This can be added back, but the error in the event handling was off. I was getting an additional error whenever there was an error that could be fixed by

```python
    @property
    def ignore_callback_errors(self):
        return super(EventEmitter, self).ignore_callback_errors
```

changing to 
```python
    @property
    def ignore_callback_errors(self):
        return super().ignore_callback_errors
```

here https://github.com/napari/napari/blob/master/napari/utils/event.py#L826-L828

Does that change make sense to you?

napari/components/_tests/test_layers_list.py
sofroniewn: we dropped the layers.move_selected method, it should have been private in the first place and was really more of an internal "view" call necessary when we implemented our own drag and drop. We don't need this whole thing anymore, and the equivalent functionality is tested on the view side.

napari/components/viewer_model.py
sofroniewn: no grid mode is failing because we have a callback for `_on_translate_grid_change` but we never emit a `translate_grid` event!! Hard for it to work without that!!!

napari/components/viewer_model.py

napari/utils/event_handler.py
sofroniewn: commend added

napari/utils/event_handler.py
sofroniewn: done

napari/utils/event_handler.py
sofroniewn: done

napari/utils/list/_model.py
sofroniewn: I'd like to drop the whole event during my next PR, but can we live with `changed` for one more, as it already existed before this? See how it got connected above.

napari/components/viewer_model.py
sofroniewn: See comment below

> I'd like to drop the whole event during my next PR, but can we live with `changed` for one more, as it already existed before this?

sofroniewn:
Ok I've gone through and addressed changes as best I can. I dropped the ability to drag and drop onto the trashcan (we can always add back in later if people feel strongly, but hover stylings weren't really working for it anyway).

I also couldn't figure out how to improve the stylings for the drop indicator or improve the drag/ drop experience at the top/ bottom of the list widget. Maybe @tlambert03 can give that a try? I did improve the text highlighting problem which I think was far worse. We can also continue to improve the stylings after merge too.

I've also left in the `changed` event and `_on_changed_change` even though it's a bit weird. I want to remove them in the next PR and we already had the `changed` event, so I feel like it's much simpler to just leave it rather than rename it only to take it out next PR.

Would be great @tlambert03 if you could go through the code, and also let me know if you can figure out how to improve the drop. A couple things I looked into were painting our own drop indicator (I couldn't figure out how to change it's thickness), and improving this line https://github.com/napari/napari/pull/1391/files#diff-6408ff7cd9e06a487f66df8d17583e6eR155-R156 which ignores the drop if the cursor isn't over one of the list items.

Feel free to take another pass too @jni!!

sofroniewn:
I just added additional opacity scaling at a level of `0.7` for the drag pixmap. Easy to add more

sofroniewn:
Upon suggestion from @jni I've lowered the opacity to `0.5` and I've further improved the interaction between the QLineEdit box and the dragging functionality - you only enter the edit mode (focus) of the QLineEdit after you've clicked on it once and then released a click. This change allows you to drag a `QtLayerWidget` around quite nicely.

The only things not implemented at this point on the stylings are better drag/ drop at the top/ bottom and a nicer looking `dropIndicator`.

sofroniewn:
Ok, looks like there is another selection bug now where sometimes clicking/ dragging selects multiple items. Not sure how to deal with this one, but if @tlambert03 could look into that too that would be great! I think I've done all the Qt styling I can for one night

sofroniewn:
> Ok, looks like there is another selection bug now where sometimes clicking/ dragging selects multiple items.

I've determined the cause of this problem, it comes when clicking and dragging "between" the list elements, it's very easy to see if you increase the list spacing size. I havn't quite figured out the best solution yet, but at least I understand the problem!!

List with increased size, where click dragging in between elements lead to selection!
image

sofroniewn:
Ok this is better, but I still can't quite get it, I still have problems with selections sometimes happening when they shouldn't, but I really don't know why. For me this works as expected though > 95% of the time. Curious what it is like for others!!

sofroniewn:
Ok, I've fiddled with the stylings and interactivity one more time. I reverted the very last change around improving selectability - but I now think interactivity is close to identical with that on master right now (and improved since the last review from @tlambert03 which had deficits from master).

Neither the top/ bottom drag/ drop nor the highlighting of text after the layer has been selected are different from master, but I see this PR really as a refactor - NOT - a time to add new functionality.

I'd like to merge this tonight @jni @tlambert03 so I can press on with world coordinates. Can you give it one last try. I'm happy to make more changes to stylings if you find performance lagging master, but I don't want to spend more time trying to add functionality not present in master.

tlambert03:
About to take a look. Is there anything in here that world coordinates depends on?  Or is it just the mental burden of having two things going on at once?  I definitely don't think you should spend much more time on this, since we don't know how much is going to survive the change to layer groups.  At the same time, this is a pretty key part of the interface, so I don't necessarily want to rush a merge... (but I agree, no reason to go above and beyond the current master, just don't want to go backwards in the UX in any way).

sofroniewn:
> Is there anything in here that world coordinates depends on? Or is it just the mental burden of having two things going on at once?

More so the mental burden/ refactor burden. Will be easier to not be making two big changes to layerlist at same time. I want to revisit and get #1360 merged after this.

> I definitely don't think you should spend much more time on this, since we don't know how much is going to survive the change to layer groups.

Completely agree here.

> At the same time, this is a pretty key part of the interface, so I don't necessarily want to rush a merge... (but I agree, no reason to go above and beyond the current master, just don't want to go backwards in the UX in any way).

100% yeah, I don't want to make things worse. Let me know how it feels now, I think it's back to parity, but can be hard to catch everything!!

tlambert03:
noticing one change from master WRT sorting: if you select multiple layers and move them, their order is reversed.  I think it's fine to let this pass for now, but just wanted to note it somewhere, and when we refactor for Tree view, we should make sure that moving layers preserves relative order.

Untitled

sofroniewn:
> noticing one change from master WRT sorting: if you select multiple layers and move them, their order is reversed.

Nice catch! That was an easy fix, just had to reverse one list!! Let me know if that now works for you and if you have any other requests, otherwise I'll consider this good to merge.

tlambert03 (approved):
I'm sorry I don't have time to look super carefully through the whole thing, but I definitely love the amount of code you were able to remove, and the new visual during drag is lovely!  The styles look good and the UX behavior looks great.  Good by me if you want to move on to other things

sofroniewn:
After more discussion with @jni I've gotten the green light for merge. We'll need to add some qt tests to cover our custom drag/ drop (see coverage report here https://codecov.io/gh/napari/napari/compare/e43f730170aab1e011c42060c65ba44133f6f18f...87910700153881903401c4f1d3f58285c94a674b/src/napari/_qt/qt_layerlist.py?before=napari/_qt/qt_layerlist.py) but i will make a dedicated issue for that as we havn't had test like that before (the code before the refactor in this place was also un-tested)

PR #1394, opened 2020-06-30
Add EventHandler docs

sofroniewn:
# Description
This PR adds explanation docs for the EventHandler refactor #1376. It includes the amazing diagram from #1040 that @shanaxel42 made, and a link to this wikipedia article on the [model-view-presenter](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93presenter) which I thought might be helpful.

The doc is definitely aspirational and describes where we want to be, but I hope it will be helpful to people trying to understand our architecture. It will certainly need updating and refining as we go, but I think it would be nice to get in fairly soon after #1376 and then update it so we at least have something there.

Thoughts @jni @tlambert03 @ziyangczi @kevinyamauchi @shanaxel42? This is ready for review!


docs/source/explanations/event_handling.md
```diff
@@ -0,0 +1,84 @@
...
+
+```python
+layer.events.colormap('my_colormap')
+```
+
+If this ```EventEmitterGroup``` was connected to an ```EventHandler``` then
+the ```EventHandler``` will receive this event and its payload and then will
+call any corresponding update methods on any object that has been registered
+as a listener. Update methods are linked to particular emitters via a naming
+convention ```_on_*change``` where the ```*``` corresponds to the name of the
```
liu-ziyang:
```suggestion
convention ```_on_*_change``` where the ```*``` corresponds to the name of the
```

liu-ziyang (approved):

sofroniewn:
So there's something a little undesirable happening now on the Qt/ view side of things, which we maybe should have thought more about earlier. Every time you do something like select a new option from a dropdown menu we emit an event saying that has changed, the event handler receives it and then sends to the model the new value, but ALSO sends back to the Qt/ view the new value which updates it again! This unnecessary call back to the view isn't so bad most of the time as most of the time you're just setting a value to the thing it already has, for example the value in the dropdown menu, but its a bit weird if the value and response corresponds to a more incremental operation - like "swap these two things", you basically then swap them twice and it's super weird.

Part of the goals of the event handler were to avoid these duplicate calls, loops etc. and we did that on the model side by having our `setters` and our `_on_*_change` methods, but on the Qt side it's harder because we can't just make the setter not set. Curious @tlambert03 how you think about this? If this makes sense?

For a concrete example see #1391 and the `_on_reordered_change` method that gets unnecessarily called by the event handler after the drag and drop emits a `reordered` event (via the custom `orderChanged` slot). https://github.com/napari/napari/blob/034c258878838c774be5ed2436d6d0a4b25153d2/napari/_qt/qt_layerlist.py#L112-L142

tlambert03:
i'm partially following, but not completely, so let me know if I'm missing it

With the concrete `orderChanged ` example here, I feel like that `orderChanged` is itself a little funny.  that's new in that PR right?  If we are already adding `EmitterGroups` to our Qt classes (which is fine), then I think we should try to use it.  So  I don't totally follow why a new `Signal` is being added, that is only being connected within the `QtLayerList` class itself.  Why not just directly emit `self.events.reordered` [here](https://github.com/napari/napari/blob/034c258878838c774be5ed2436d6d0a4b25153d2/napari/_qt/qt_layerlist.py#L142)?

> on the Qt side it's harder because we can't just make the setter not set. 

what is the specific setter method here that we can't avoid?

> Part of the goals of the event handler were to avoid these duplicate calls, loops etc. and we did that on the model side by having our setters and our _on_*_change methods... 

I'm sort of waiting to see how the whole thing settles :) I'm _slightly_ uncomfortable with the new "setters don't actually set" pattern, but willing to see where it goes.  While the old blockers were always a bit ugly, I've more or less accepted that as a pattern that arises in event-driven programming, and there weren't that many of them.

sofroniewn:
> With the concrete orderChanged example here, I feel like that orderChanged is itself a little funny. that's new in that PR right? If we are already adding EmitterGroups to our Qt classes (which is fine), then I think we should try to use it. So I don't totally follow why a new Signal is being added, that is only being connected within the QtLayerList class itself. Why not just directly emit self.events.reordered here?

That's easy to change, probably was overkill on my part to use a Qt signal, but it won't fix the underlying issue.

The issue here is that when the drop happens Qt does the reordering when I call `super().dropEvent(event)` and then emits the `reordered` event. To be more inline with our EVH pattern I'd have to skip the `super().dropEvent` call, determine the old and new indices myself and then emit them in the `reordered` event which would flow through the EVH and then come back to the `_on_reordered_change` method to do the actual reordering.

This flow back happens every time now on the Qt side of things, and I guess that is just a little weird. In this particular case we were probably going to overwrite a Qt call anyway, but to really fix this we'd have to overwrite every qt call which makes no sense.

> I'm sort of waiting to see how the whole thing settles :) I'm slightly uncomfortable with the new "setters don't actually set" pattern, but willing to see where it goes. While the old blockers were always a bit ugly, I've more or less accepted that as a pattern that arises in event-driven programming, and there weren't that many of them.

Yeah, I do feel what I'm doing inside the `ListModel` right now is pretty intense, and to me far less intuitive than before (i.e. emitting an event before the item gets added rather than after ....). I do like that we'd need to pass around our layer model less now than we did before, but I also worry that we're reducing complexity in some parts of the napari codebase at the expense of depending on hidden complexity in Qt and vispy (thinking about everything that I now pull from the `.node` property on the vispy object or re-request from Qt in our controls) and pretty fragile before in others - especially when thinking about coordinating between different levels of the app (viewer, layerlist, layers)

tlambert03:
> The issue here is that when the drop happens Qt does the reordering when I call `super().dropEvent(event)` and then emits the `reordered` event. To be more inline with our EVH pattern I'd have to skip the `super().dropEvent` call, determine the old and new indices myself and then emit them in the `reordered` event which would flow through the EVH and then come back to the `_on_reordered_change` method to do the actual reordering.

oh I see, so you're calling `dropEvent()` here the "setter" in that it executes the QListWidget model reordering?  At a first approximation, I would not try to send the actual "drop event" through the event handler.  I'd say let's leave things that are strictly internal GUI events (i.e. a slider is moving, or a widget is dragging) as internal.  The real event here that needs to be emitted (so that the LayerList model can update itself) is the reordering of the internal model in the QListWidget.  In other words, I think your pattern is ok... and if I just get rid of your `_on_reordered_change` method it seems to work fine (though I assume that breaks things when the layerlist emits reordered?).  So I would focus on implementing `_on_reordered_change` such that it receives layerlist events, but ignores internal reorder events.  if you must, I would add back a blocker... this seems to be exactly why they were created.

sidenote: a major difference between QTreeView and QListWidget is that QTreeView _has no model_ (we'll make a simple wrapper QAbstractItemModel that basically just proxies the real LayerGroup model), whereas QListWidget is both a view and a model.  That's generally true: QWidgets have both a view and model, Views are just "views" (and you provide the model).  So all of these things you're having to do to keep the ListWidget model and the LayerList model in sync will go away.

sofroniewn:
>  I would not try to send the actual "drop event" through the event handler. I'd say let's leave things that are strictly internal GUI events (i.e. a slider is moving, or a widget is dragging) as internal. The real event here that needs to be emitted (so that the LayerList model can update itself) is the reordering of the internal model in the QListWidget. 

I'm not sending the drop event through our event system, I was just using it to trigger the reordering of the model, instead of things like the `rowMoved` signal or alternatives, but the same problem remains regardless which as you identified is that the `_on_reordered_change` change gets triggered via our event handler and as that's fully part of our event system (we need it for when that layer model reorders the layer list) there's no way to stop this without adding a blocker to our event system, which is precisely one of the things we were trying to avoid.

I see how using the `QListView` here might help, but I still think there are larger issues.

I recall us discussing at length how to resolve this on our model side, but we I don't think we really discussed as fully on the view side. Take one of the simplest cases we now have which is the gamma slider. Every time the gamma slider moves it emits an event to the EVH which then calls all the `_on_gamma_change` methods, including the one on the Qt object which contains the `self.gammaSlider.setValue` call which is fully redundant because the slider already moved. As you said in https://github.com/napari/napari/pull/1391#issuecomment-653652725 this is because we're using QWidgets which have both models/ views themselves, and it's not clear to me how to resolve this situation, even in this simple case of a single slider. We could change the event handler so that it never called the `_on_*_change` method for the thing which did the emitting and let setters work on themselves directly everywhere, but that seems a bit strange yet again.

TBH I'm at a bit of a loss here, but this stuff is probably worth discussing more before we go too far with the EVH refactor ...

tlambert03:
All super good points.  It's tough!   I definitely don't have a clear view to the best solution myself, and I think we're still doing the right thing by testing how the pattern spreads to various other components... finding stuff along the way.
I agree that removing the blockers looked like a promising feature, and it may still be possible of course.  But for me that was more of a "bonus" to what feels like the real feature of the new EVH, which is the bottleneck, as you described it elsewhere.  That part still feels solid, achievable, and useful, even if some of the mechanics need to be modified slightly.

I'd say keep pushing in the directions you're going in, and keep pointing out these hiccups.  I'll try to play with this over the weekend as well (it's so hard to be all that confident about anything without really digging into it)

docs/source/explanations/event_handling.md
```diff
@@ -0,0 +1,84 @@
...
+to reflect the latest progress. Some of the document is written from an
+aspirational perspective of where we'd like napari to be after this refactor
+is complete.
+
+## Overview of Models, Views, Events, and Event Handlers
+
+napari is architected to have a strict separation of its model code and its 
+view code.
+
+- **models** directly manages data, logic, and rules of within napari,
```
willingc:
```suggestion
- **models** directly manage data, logic, and rules of within napari,
```

docs/source/explanations/event_handling.md
```diff
@@ -0,0 +1,84 @@
...
+independent of any user interface and emit updates to an event handler via
+events.
+
+- **views** contain a representation of information from a model and route
+user commands to an event handler via events.
+
+- **events** are emitted from both views and models in napari and all flow
```
willingc:
```suggestion
- **events** are emitted from both views and models in napari and all events flow
```

docs/source/explanations/event_handling.md
```diff
@@ -0,0 +1,84 @@
...
+acting upon them to update the models and views.
+
+![event-handler](images/event-handler.png)
+
+For example the above diagram shows the flow of information for the ```Image```
+layer model, and two views, one ```VisPy``` based and one ```Qt``` based that
+are all being connected via an event handler. 
+
+This architecture is probably most similar to the [model-view-presenter](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93presenter) architecture. A unique aspect of napari
+that distinguishes it from other GUI applications is that advanced users expect
```
willingc:
```suggestion
that distinguishes it from other GUI applications is that advanced users expect: 
```

docs/source/explanations/event_handling.md
```diff
@@ -0,0 +1,84 @@
...
+
+![event-handler](images/event-handler.png)
+
+For example the above diagram shows the flow of information for the ```Image```
+layer model, and two views, one ```VisPy``` based and one ```Qt``` based that
+are all being connected via an event handler. 
+
+This architecture is probably most similar to the [model-view-presenter](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93presenter) architecture. A unique aspect of napari
+that distinguishes it from other GUI applications is that advanced users expect 
+to be able to interact directly with the napari model from interactive
```
willingc:
```suggestion
- to be able to interact directly with the napari model from **interactive**
```

docs/source/explanations/event_handling.md
```diff
@@ -0,0 +1,84 @@
...
+environments like ```IPython``` and ```Jupyter``` and to use napari models
```
willingc:
```suggestion
environments like ```IPython``` and ```Jupyter``` 
- to use napari models
```

docs/source/explanations/event_handling.md
```diff
@@ -0,0 +1,84 @@
...
+headless, without any views when performing batch analysis. This feature forces
```
willingc:
```suggestion
headless, without any views when performing **batch** analysis.

These uses motivate
```

docs/source/explanations/event_handling.md
```diff
@@ -0,0 +1,84 @@
...
+us to put more logic into our model files so that they can be used stand-alone.
+
+Right now napari just supports one front-end, built using [Qt](https://www.qt.io/),
+and one rendering library using [VisPy](http://vispy.org/), but in the future
```
willingc:
```suggestion
and one rendering library using [VisPy](http://vispy.org/). In the future
```

docs/source/explanations/event_handling.md
```diff
@@ -0,0 +1,84 @@
...
+
+napari has vendored the event system from ```VisPy``` which is documented
+[here](http://vispy.org/util.html#module-vispy.util.event) and is largely
+unchanged. 
+
+Each **model** or **view** is capable of having an ```EventEmitterGroup```
+connected to an ```EventHandler``` that contains various named ```EventEmitter```
+objects, each of which is capable of emitting events that can carry a single
+value as a payload. For example you can emit a custom colormap string from
+the ```colormap``` emitter as follows:
```
willingc:
```suggestion
the ```colormap``` event emitter as follows:
```

docs/source/explanations/event_handling.md
```diff
@@ -0,0 +1,84 @@
...
+
+```python
+layer.events.colormap('my_colormap')
+```
+
+If this ```EventEmitterGroup``` was connected to an ```EventHandler``` then
```
willingc:
```suggestion
By connecting an ```EventEmitterGroup``` to an ```EventHandler```, events can trigger an event handler to perform actions. When the event is emitted,
```

docs/source/explanations/event_handling.md
```diff
@@ -0,0 +1,84 @@
...
+the ```EventHandler``` will receive this event and its payload and then will
```
willingc:
```suggestion
the ```EventHandler``` will receive this event and its payload and act on the event by
```

docs/source/explanations/event_handling.md
```diff
@@ -0,0 +1,84 @@
...
+call any corresponding update methods on any object that has been registered
```
willingc:
```suggestion
calling any corresponding update methods on any object that has been registered
```

docs/source/explanations/event_handling.md
```diff
@@ -0,0 +1,84 @@
...
+emitter, ```colormap``` in the example above. These update methods take one
+variable as an input argument, which is always the payload carried by its
+corresponding event.
+
+This event system allows us to have a loose coupling between our models and
+views but ensures that information never gets out of sync or gets into circular
+updating patterns. 
+
+As we expect napari users to be able to interact with our model object directly
+we use ```setters``` on our attributes to emit events and ```_on_*_change```
```
willingc:
```suggestion
we use ```setters``` on model attributes to emit events and ```_on_*_change```
```

docs/source/explanations/event_handling.md
```diff
@@ -0,0 +1,84 @@
...
+methods to actually perform updates. When instantiating our models the
+necessary event handlers are created and attached to them so that they can
+be used without additional imports. The separation of our setters, event
```
willingc:
```suggestion
be used without additional imports. The clear separation of our setters, event
```

willingc (approved):
@sofroniewn This looks great. I've made some suggestions to improve clarity. Event handling is something that new contributors and users find conceptually challenging. In my experience a bit more precision (re: pronoun references) aids understanding.

willingc:
Looks like the CI error on Windows is unrelated to this PR.

sofroniewn:
Closing now due to reversion of EVH in #1416. We'll likely preserve much of this content in a new PR after we sort our things like the approach in #1436

PR #1396, opened 2020-06-30
Event handler surface layer

liu-ziyang:
# Description
Add event handler for surface layer, with same pattern to the basic layer

## Type of change
- Refactor of an existing feature

# References
https://github.com/napari/napari/pull/1394
https://github.com/napari/napari/pull/1376
https://github.com/napari/napari/pull/1040

# How has this been tested?
- [X] all tests pass with my change

## Final checklist:


napari/layers/surface/surface.py
```diff
@@ -123,7 +123,7 @@ def __init__(
             visible=visible,
         )
 
-        self.events.add(interpolation=Event, rendering=Event)
```
liu-ziyang, line 126: These events seems deprecated, I didn't find reference of it anywhere, if needed I can reinstate them

napari/layers/surface/surface.py
```diff
@@ -194,9 +200,12 @@ def faces(self) -> np.ndarray:
 
     @faces.setter
     def faces(self, faces: np.ndarray):
+        self.events.faces(faces)
+
+    def _on_faces_change(self, faces: np.ndarray):
         """Array of indices of mesh triangles.."""
 
-        self.faces = faces
```
liu-ziyang, line 199: this seems like a recursion bug

napari/layers/surface/surface.py
```diff
@@ -165,6 +165,9 @@ def vertices(self):
 
     @vertices.setter
     def vertices(self, vertices):
+        self.events.vertices(vertices)
+
+    def on_vertices_change(self, vertices):
```
sofroniewn:
```suggestion
    def _on_vertices_change(self, vertices):
```

napari/layers/surface/surface.py
sofroniewn: hmm looks like you're right, I don't see them being called anywhere so maybe we can drop for now. We don't support multiple interpolation modes or rendering modes for surface right now anyway. maybe was a mistake that they are in there

napari/layers/surface/surface.py
sofroniewn: yes, this looks like a bug, good catch

sofroniewn (commented):
This looks great on the model side @ziyangczi.

On the view side as you noted the [QtSurfaceControls](https://github.com/napari/napari/blob/master/napari/_qt/layers/qt_surface_layer.py) don't need any additional refactoring but the [VispySurfaceLayer](https://github.com/napari/napari/blob/master/napari/_vispy/vispy_surface_layer.py) needs some work. 

Look at what I did for the VispyBaseLayer and VispyImageLayer in #1376 for guidance. See how far you get with the simple stuff, it might not be possible to fix `_on_slice_data_change` until all the layers are fixed but at the very very end of this process we'll want to be able to remove the layer entirely from those vispy views!!

liu-ziyang:
> This looks great on the model side @ziyangczi.
> 
> On the view side as you noted the [QtSurfaceControls](https://github.com/napari/napari/blob/master/napari/_qt/layers/qt_surface_layer.py) don't need any additional refactoring but the [VispySurfaceLayer](https://github.com/napari/napari/blob/master/napari/_vispy/vispy_surface_layer.py) needs some work.
> 
> Look at what I did for the VispyBaseLayer and VispyImageLayer in #1376 for guidance. See how far you get with the simple stuff, it might not be possible to fix `_on_slice_data_change` until all the layers are fixed but at the very very end of this process we'll want to be able to remove the layer entirely from those vispy views!!

updated vispy surface code, let me know if this looks right

napari/_vispy/vispy_surface_layer.py
```diff
@@ -51,8 +45,8 @@ def _on_slice_data_change(self, event=None):
         self._on_scale_change()
         self._on_translate_change()
 
-    def _on_colormap_change(self, event=None):
-        cmap = self.layer.colormap[1]
+    def _on_colormap_change(self, colormap):
```
sofroniewn, line 50: you'll actually need to copy the entire function approach here https://github.com/napari/napari/blob/master/napari/_vispy/vispy_image_layer.py#L144-L162 - it does two additional things, one is the `ensure_colormap_tuple` and the other is the `self._gamma` so we don't need the `layer.gamma` property anymore (which we will eventually want to remove).

You'll also need to do https://github.com/napari/napari/blob/master/napari/_vispy/vispy_image_layer.py#L178-L195 and add https://github.com/napari/napari/blob/master/napari/_vispy/vispy_image_layer.py#L25-L28.

I guess we could have an additional baseclass `VispyImageBaseLayer` that both inherit from if you want to reduce code duplication.

sofroniewn: Can you also add a doc string

napari/_vispy/vispy_surface_layer.py
```diff
@@ -63,13 +57,13 @@ def _on_colormap_change(self, event=None):
...
         self.node.cmap = cmap
 
-    def _on_contrast_limits_change(self, event=None):
-        self.node.clim = self.layer.contrast_limits
+    def _on_contrast_limits_change(self, contrast_limits):
+        self.node.clim = contrast_limits
 
-    def _on_gamma_change(self, event=None):
-        self._on_colormap_change()
+    def _on_gamma_change(self, gamma):
```
sofroniewn, line 82: Doc string and change method, as described above

napari/_vispy/vispy_surface_layer.py
```diff
@@ -63,13 +57,13 @@ def _on_colormap_change(self, event=None):
             )
         self.node.cmap = cmap
 
-    def _on_contrast_limits_change(self, event=None):
-        self.node.clim = self.layer.contrast_limits
+    def _on_contrast_limits_change(self, contrast_limits):
```
sofroniewn, line 71: Can you add a doc string - can be identical to this https://github.com/napari/napari/blob/master/napari/_vispy/vispy_image_layer.py#L165-L171

napari/layers/surface/surface.py
```diff
@@ -165,6 +165,9 @@ def vertices(self):
 
     @vertices.setter
     def vertices(self, vertices):
+        self.events.vertices(vertices)
+
+    def _on_vertices_change(self, vertices):
```
sofroniewn, line 170: Can you make doc string here and for other `_on*_change` methods look more like 
```
"""Receive vertices change event and update the model.

Parameters
----------
    vertices : array
         Array of vertices of mesh triangles.
"""
```

sofroniewn (commented):
Looking good.

I think for the vispy object refactoring as you go through the other layers it would be really great to remove all of the mentions of `self.layer` that aren't in the `reset` function call or in `_on_slice_data_change` or `_on_display_change` - we can try removing those from all layers at the same time at the very end, and then we can be gone with `self.layer` entirely!!!

I've left some tips on how to copy the `VispyImageLayer` to remove `self.layer` from `_on_colormap_change` and `_on_gamma_change`. You might actually want to create an intermediate class `VispyImageBaseLayer` that has the duplicated methods and inherit that one in both, it will reduce some otherwise duplicate code.

I've also requested more doc strings/ doc strings more in the desired format, but should be close to being done!!

napari/_vispy/vispy_surface_layer.py
liu-ziyang: this is what confused me, the gamma is inherited from intensity_mixin.py, which seems different from what other layers do, is it a legacy component not updated or is there anything special for surface layer?

liu-ziyang: I am updating the PR so I can move on, but I do want to know some history behind surface layere :)

sofroniewn: The properties in the `intensity_mixin` are shared between Image and Surface, and so we used an inheritance pattern to reduce code duplication. Looking forward I think we'll want to refactor to use a component pattern instead, which will make this much much nicer, for example both Image and Surface could then use a `ColorManger` or `RenderingManager` component that would control properties now in the `intensity_mixin`. Does that make more sense?

napari/_vispy/vispy_surface_layer.py
```diff
@@ -51,25 +46,50 @@ def _on_slice_data_change(self, event=None):
...
-        self.node.clim = self.layer.contrast_limits
+        Parameters
+        ----------
+        gamma : float
+            New gamma value.
+        """
 
-    def _on_gamma_change(self, event=None):
-        self._on_colormap_change()
+        self._on_colormap_change(self.layer.colormap)
```
sofroniewn: This code block needs to be duplicated too https://github.com/napari/napari/blob/master/napari/_vispy/vispy_image_layer.py#L186-L195

napari/_vispy/vispy_surface_layer.py
```diff
@@ -17,12 +18,6 @@ def __init__(self, layer):
 
         super().__init__(layer, node)
```
sofroniewn, line 20: You still need to add this initialization

https://github.com/napari/napari/blob/master/napari/_vispy/vispy_image_layer.py#L25-L28

tlambert03: somehow I missed these attribute changes/additions to the Image Layer in that previous PR.  Tell me again why we need to duplicate what seem to be model properties on the vispy classes?  Is this because the event.value itself doesn't contain all of the necessary info?

sofroniewn: in some ways yes, but we'll be able to drop them again after the gamma PRs go through at vispy https://github.com/vispy/vispy/pull/1842 and https://github.com/vispy/vispy/pull/1844.

The issue now is that the vispy node only tracks the actual `cmap` which is a combination of the `_raw_cmap` and the `_gamma` value according to the CPU computation from before. When either the colormap or the gamma alone changes you need to access the value of the other variable now to get the correct final colormap. Once all this is within vispy that won't be necessary anymore and we'll just be able to update `node.gamma` and `node.cmap` and everything will work, I think!!

Does that checkout? It will be a good thing to verify before the next vispy release as you finalize that last PR

tlambert03: that checkouts for gamma and cmap, but what about iso_threshold? and `_attenuation`?  I guess in general I missed the fact that changing the event handler required us to duplicate these model values at all.  in other words, it shouldn't require a PR to an external library to avoid that pattern, right?  like, maybe for complex cases like this, the event value itself needs to get more complex?

will open a different issue for this, but a related thing I'm sort of mourning is the loss of the `event.source` attribute after the event handler passes it on (I often used that for my own connections).  I can certainly just make my own connections without going through the event_handler ... we can discuss elsewhere.

sofroniewn: For `_attenuation` and `_iso_threshold` I actually think it was my bad design when I made my PR to vispy without adding a separate `attenuation` parameter, and instead overloading their `threshold` parameter that got us here. I might want to try and fix that before their next release, but if we don't fix in vispy, I think maintaining some state on the view like this is still ok. Much better to just maintain these two floats for a complex case than maintain the whole layer model, right?

As to the `event.source` stuff, we can discuss at team meeting. There's also some interesting stuff coming to light during in #1391 as I add the EVH to the layerlist that will be worth discussing too.

sofroniewn (approved):
Works great, I just pulled down and everything looks good. Might be nice for @jni to take a look too before merge as it makes clear some of the changes required to bring a layer in line with the new event handler style!!

sofroniewn:
Note this PR does not require the changes added in #1402 as it uses a `Mesh` visual and not a `Volume` visual

PR #1402, opened 2020-07-05
Fix vispy volume colormap changing

### LINKED ISSUE #1401: Colormap in 3D broken by Image Layer Event Handler
## 🐛 Bug

Changing the colormap in 3D doesn't cause the colormap for the actual data to update. The thumbnail does update. This was likely introduced in #1376. Changing colormap in 2D still works fine.

sofroniewn:
# Description
This PR closes #1401 by reverting some code that was removed as part of #1376. I had thought the code was no longer needed due to the merge of https://github.com/vispy/vispy/pull/1712 in vispy, but in fact it is still needed, and the correct fix in vispy would be to insert this change into the cmap setter here https://github.com/liuyenting/vispy/blob/e9551f4880a7a6d8839be5f63b3a3552341ebfaf/vispy/visuals/volume.py#L513

### DIFF
--- a/napari/_tests/test_with_screenshots.py
+++ b/napari/_tests/test_with_screenshots.py
@@ -1,0 +1,68 @@
+import numpy as np
+import os
+import sys
+import pytest
+
+
+@pytest.mark.skipif(
+    sys.platform.startswith('win') or not os.getenv("CI"),
+    reason='Screenshot tests are not supported on napari windows CI.',
+)
+def test_changing_image_colormap(make_test_viewer):
+    """Test changing colormap changes rendering."""
+    viewer = make_test_viewer(show=True)
+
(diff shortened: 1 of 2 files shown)


jni (approved):

jni:
@sofroniewn would you like to add a test for this?

sofroniewn:
> @sofroniewn would you like to add a test for this?

I've now added two screenshot tests one for colormap and one for gamma that cover setting both 2D and 3D and would have caught this bug. These tests will be robust to changes in underlying rendering approach (i.e. #1404) as they use screenshots.

In general I'd like to add more screenshot tests like this. We now have a good way of flagging them for CI only and I think they will be very helpful

jni (approved):

Issue #1392, opened 2020-06-29
Default value for attenuation in attenuated-MIP rendering should be lower


jni:
## 🐛 Bug

I'm not sure if it's my data or a general feature, but I have always found that the attenuation value for 3D rendering is way too high for most volumes I look at. I invariably end up reducing it to just above the minimum value of the slider. So we should probably revisit that default value.

sofroniewn:
Interesting, might be good to think why that is/ if more normalization is needed or if the value is just too high!

AdvancedImagingUTSW:
The fix worked beautifully!  As you know, I used an attenuation value of .04 for this dataset.  However, I'm going to test a couple of different datasets out with the attenuated MIP setting and I'll get back to you as soon as possible on the range of values I find useful.  Thanks again, I really appreciate your work and I think you are building the foundation of something potentially transformative.  

Best,
Kevin

<img width="1002" alt="image" src="https://user-images.githubusercontent.com/42547789/87490975-98948980-c60c-11ea-98b8-425d4a7a4a25.png">

tlambert03:
a while back we had (on slack) talked about, and rejected using a nonlinear $cmap function in the shader.... pasting that code here again in case we want to play with it:

```python
ATTENUATED_MIP_SNIPPETS = dict(
    before_loop="""
        float maxval = -99999.0; // The maximum encountered value
        float minval = 99999.0; // The minimum encountered value
        float sumval = 0.0; // The sum of the encountered values
        float scaled = 0.0; // The scaled value
        int maxi = 0;  // Where the maximum value was encountered
        vec3 maxloc = vec3(0.0);  // Location where the maximum value was encountered
        """,
    in_loop="""
        sumval = sumval + val;
        scaled = val * exp(-u_threshold * sumval / u_relative_step_size);
        if( scaled > maxval ) {
            maxval = scaled;
            maxi = iter;
            maxloc = loc;
        }
        if( val < minval ) {
            minval = val;
        }
        """,
    after_loop="""
        gl_FragColor = $cmap((maxval - minval) * (1 + pow(u_threshold, .5)));
        """,
)
```

sofroniewn:
Linking to this comment with great images, examples https://github.com/napari/napari/issues/1399#issuecomment-658866839

Seems like values there are 0.02, 0.06, 0.04.

How about we bring the slider range down to `(0, 0.2)` with a default of `0.1`. I can do that now

PR #1403, opened 2020-07-05
fix attenuation setting

### LINKED ISSUE #1399: Attenuated MIP Initialization Error
## 🐛 Bug

The viewer is not programmatically updating the attenuation value for an 'attenuated_mip' rendering.  
Full discussion can be found here: https://forum.image.sc/t/attenuated-mip-refresh/39884/9

## To Reproduce

1.  Launch a simple script with a 3D image.
2. Try various attenuation values.
3. In general, it seems as if it is automatically showing an attenuation value of 0 regardless of what I provide in the viewer.add_image function.

`
import napari
import numpy as np
from tifffile import imread
import os

data_location = '/archive/bioinformatics/Danuser_lab/Fiolka/MicroscopeDevelopment/OPM/Drosophila/GAP43-mcherry/200703/Cell1'
membrane = imread(os.path.join(data_location, f'1_CH00_{1:06}.tif'))
membrane = np.array(membrane)

with napari.gui_qt():
    layer = viewer.add_image(data=membrane, rendering='attenuated_mip', attenuation=1)
`

## Expected behavior
I would expect the attenuation value, which is accurately reflected in the slider position to also be shown in the image.  

## Environment

 napari: 0.3.6.dev11+ge43f730
Platform: Linux-3.10.0-957.el7.x86_64-x86_64-with-redhat-7.6-Maipo
Python: 3.6.10 | packaged by conda-forge | (default, Apr 24 2020, 16:44:11) [GCC 7.3.0]
Qt: 5.14.2
PyQt5: 5.14.2
NumPy: 1.18.4
SciPy: 1.4.1
Dask: 2.17.2
VisPy: 0.6.4

GL version: 2.1 Mesa 18.3.4
MAX_TEXTURE_SIZE: 8192

Plugins:

napari-plugin-engine: 0.1.5
napari_volume
svg: 0.1.3

 - Linux OS

## Additional context
Nicholas Sofroniew identified the problem: 
...it comes from this line in vispy https://github.com/vispy/vispy/blob/caed4f0326dc9c0804b254339c4d89f66e6a01b6/vispy/visuals/volume.py#L598 and the fact that on startup the u_threshold is not in the shared_program. It seems to get added once things have appeared on screen, hence why things only start working when you move a slider.

I need to think about the best fix. It would be nice to fix on the vispy side, though presumably they had that line there for a reason. If we bypass that line by doing

self.node._threshold = float(attenuation)
self.node.shared_program['u_threshold'] = self.node._threshold
self.node.update()
ourselves inside here https://github.com/napari/napari/blob/e43f730170aab1e011c42060c65ba44133f6f18f/napari/_vispy/vispy_image_layer.py#L225 then it works fine. Curious what @talley thinks we should do. We can probably move this convo to github as it’s a pretty clear bug at this point.

sofroniewn:
# Description
This PR closes #1399 the attenuation initialization in vispy, and also fixes an `iso_threshold` / `attenuation` initialization in napari. The vispy fixes are probably best moved to vispy in the long run, but this works for now.

### DIFF
--- a/napari/_vispy/vispy_image_layer.py
+++ b/napari/_vispy/vispy_image_layer.py
@@ -29,7 +29,7 @@ class VispyImageLayer(VispyBaseLayer):
 
         # Until we add a specific attenuation parameter to vispy we have to
         # track both iso_threshold and attenuation ourselves.
-        self._iso_threshold = 1
+        self._iso_threshold = 0.5
         self._attenuation = 1
 
         self._on_display_change()
@@ -141,6 +141,11 @@ class VispyImageLayer(VispyBaseLayer):
             elif Rendering(rendering) == Rendering.ATTENUATED_MIP:
                 self.node.threshold = float(self._attenuation)
 
+            # Fix for #1399, should be fixed in the VisPy threshold setter
+            if 'u_threshold' not in self.node.shared_program:
+                self.node.shared_program['u_threshold'] = self.node._threshold
+                self.node.update()
+
     def _on_colormap_change(self, colormap):
         """Receive layer model colormap change event and update the visual.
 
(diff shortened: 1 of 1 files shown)


jni (approved):
Changes look good, but there should be a test! Perhaps one based on screenshots, as attenuation=0 should look identical to MIP?

This might also be a good time to change the default value, see #1392 ?

sofroniewn:
This has been superseded by #1454

PR #1412, opened 2020-07-09
fix grid mode

### LINKED ISSUE #1400: Grid mode broken by Image Layer Event Handler Refactor
## 🐛 Bug

Grid mode has been broken by the Image Layer Event Handler Refactor. Run `examples/layers.py` and see how grid mode no longer works properly. If you click the grid button you'd expect the grid mode to return to normal. This is because we don't have an explicit `translate_grid` event. I'd actually like to remove the translate grid from the layers themselves, it never quite made sense there, but we'd need to think about where to put it. It's more a property of the viewer rather than the layer itself.

I'll give fixing some thought, we should also add tests too. This would have been easy to catch/ fix if it had been tested!!

sofroniewn:
# Description
This PR closes #1400 by fixing grid mode which was broke in #1376

It still needs tests/ doc strings

### DIFF
--- a/napari/_qt/qt_layerlist.py
+++ b/napari/_qt/qt_layerlist.py
@@ -33,6 +33,7 @@ class QtLayerList(QListWidget):
             auto_connect=False,
             selected_layers=Event,
             reordered=Event,
+            changed=Event,
         )
 
         # When the EVH refactor is fully done we can do the initialization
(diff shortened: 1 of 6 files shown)


sofroniewn:
This now has comprehensive tests model of model and vispy with screenshot and is ready for review @jni

sofroniewn:
tests failing, i will look tomorrow

sofroniewn:
Closing this PR due to revision of EVH in #1416, will preserve tests in new PR

PR #1415, opened 2020-07-09
Fix vertical aligment of QLabel in QtDimSliderWidget

Czaki:
# Description

QtDim slider label which inform about number of layers is shifted down

Zrzut ekranu z 2020-07-10 00-14-55

This PR fix this

Zrzut ekranu z 2020-07-10 00-15-18

Need to be tested on other systems. 

## Type of change

# References

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/_qt/qt_dims_slider.py
+++ b/napari/_qt/qt_dims_slider.py
@@ -84,6 +84,7 @@ class QtDimSliderWidget(QWidget):
         layout.addWidget(self.totslice_label)
         layout.setContentsMargins(0, 0, 0, 0)
         layout.setSpacing(2)
+        layout.setAlignment(Qt.AlignVCenter)
         self.setLayout(layout)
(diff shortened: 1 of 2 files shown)


tlambert03:
oh hell yes!  I really hope this persists on all platforms, because that has been plaguing me for a long time.  (though it was always the cross-platform nature of it that was the problem).  It would be amazing if you can fix that for across the board!

Czaki:
Aliment work on all system. But there is problem with render on windows:
obraz

@tlambert03 Try to solve this in this PR or open separate when I found solution?

tlambert03:
thanks @Czaki, if that clipping problem was there before on windows (and I think it was right?) then I think we can leave that for a follow up PR at your convenience.

Czaki:
@tlambert03 Yes. It also happens before this patch.

tlambert03 (approved):
thanks for doing this!

PR #1428, opened 2020-07-11
Docreformat

Carreau:
This does two things: 

 -  Minimal manual changes to docs for autoformatting.

    underlines too short, missing s in a couple of spots.

 -  Apply automatic docstring reformatting

    This is based on one of the projects I'm working during those sprints.
    It does some things like sections reordering in the order numpy suggests
    in the numpydoc format, and other things that I understand would be
    controversial

This is of course also in particular an occasion to give feedback and the automatic reformatting tool which is imcomplete.


tlambert03 (approved):
thank you @Carreau!  Yes we've all been buzzing about this sprint project of yours :)
These changes are great.  I have no issues at all with the few ordering changes.  where is the reformatting code at this point?  is there a repo yet?

Carreau:
Not final yet, need t find a name and extract the code but https://github.com/Carreau/minirst/

sofroniewn (approved):
Love it! All the changes are great - section reordering, spacing fixes, section name fixes. Everything is accurate and this can be merged now as far as I am concerned.

I can come find you in the sprints @Carreau - I'm excited to hear about what more functionality you're planning to add - for example as I reviewed this PR I saw inconsistencies like whether we ended descriptions with a `.` or not - would that be in scope or not (I guess I should look at the spec to find out). Very exciting all around though!!

Carreau:
> I can come find you in the sprints @Carreau - I'm excited to hear about what more functionality you're planning to add - for example as I reviewed this PR I saw inconsistencies like whether we ended descriptions with a `.` or not - would that be in scope or not (I guess I should look at the spec to find out). Very exciting all around though!!

I don't see any problem with having automatic `.` or other. right now it's mostly line based, and rely on numpydoc to parse, so we might want to have changes upstream in numpydoc. 

I also try to be a bit conservative on the whitespace when possible to minimise changes.

```
""" doc
```
vs
```
"""doc
```
vs
```
"""
doc
```

It tries to keep the start of the docstring the same as it was.

I'm still going slowly as there are still subtle bugs where some token get eaten. 

Typically

```
dtype: whatever (default: float)
```

What is after the second colon is eaten and get reformatted as

```
dtype : whatever (default:
```

willingc:
@Carreau This is fabulous 👍

jni (approved):

jni:
@Carreau 

> I also try to be a bit conservative on the whitespace when possible to minimise changes.

I actually am very strongly partial to `"""Document this function.` as specified by PEP257, so I'd love to have that autoformat in there as an option.

But anyway, this is just crazy awesome, the diff is incredible, and ready to merge as far as I'm concerned. I'll leave it to @sofroniewn or @tlambert03 to do the honours.

Carreau:
> I actually am very strongly partial to `"""Document this function.` as specified by PEP257, so I'd love to have that autoformat in there as an option.

AFAICT PEP257 suggest same line for single line docstring, so `"""\ndoc` appear to be fine for multiline docstring.

I'm _trying_ to avoid having too many config options as otherwise everything is up for debate, but there's probably going to be a couple.


I've now implemented recursing in the AST, which touches 33 more files, should I update this PR or wait and push a new one later ?

sofroniewn:
> I've now implemented recursing in the AST, which touches 33 more files, should I update this PR or wait and push a new one later ?

I will merge this now, and then you can make a new PR for review.

PR #1429, opened 2020-07-11
Add circle/spherical brush to Labels paint brush

### LINKED ISSUE #1418: Make labels painting brush a circle
## 🚀 Feature
As proposed in https://github.com/napari/napari/issues/1142#issuecomment-649548968: change the labels brush shape to a circle (as requested on image.sc): https://forum.image.sc/t/pixel-is-not-a-square/39455

## Motivation

As described in  https://forum.image.sc/t/pixel-is-not-a-square/39455, labeling round structures is challenging with a square brush.

## Pitch

From @sofroniewn in https://github.com/napari/napari/issues/1142#issuecomment-649645267

Should be very doable. Might take a tiny hit on performance, but shouldn't be too bad. Basically these two slice calls https://github.com/napari/napari/blob/master/napari/layers/labels/labels.py#L689-L697 and https://github.com/napari/napari/blob/master/napari/layers/labels/labels.py#L704 will now need to be optionally done based on a center radius instead of center bounding box (we should be able to find away to just have one call to this new method too probably) and we'll need to make a new `circle_pixmap` for the cursor, like our `square_pixmap` here https://github.com/napari/napari/blob/f706a4157cf80e6e0b0d3f21adc0a1c3f7c3b410/napari/_qt/utils.py#L75-L85 which then get's used here https://github.com/napari/napari/blob/master/napari/_qt/qt_viewer.py#L430

hectormz:
# Description

This pull request adds the ability to use a circle/spherical paint brush on labels layers

## Type of change

# References

This closes #1418 

# How has this been tested?

- [X] Updated & verified existing square brush test runs
- [X] Added 2D circle paint brush test
- [X] Added 3D circle/sphere paint brush test

## Final checklist:


hectormz:
It's not as pretty as I'd like, but I have basic circle/sphere painting in Layers. I pushed the changes to the PR, which contains hardcoded circle setting.
To Do:
- Store brush_shape attribute in Labels class
- Add right click popup to paintbrush to select brush shape
- Verify math of determining coordinates within circle/sphere
    - Look into potentially faster ways to determine circle membership of pixels
- Add and edit paintbrush tests
- Clean up code when dust settles

napari/layers/labels/labels.py
```diff
@@ -414,6 +416,17 @@ def color_mode(self, color_mode: Union[str, LabelColorMode]):
         self.events.selected_label()
         self.refresh()
 
+    @property
+    def brush_shape(self):
+        """float: Paintbrush shape"""
```
sofroniewn: not float - should be str

hectormz:
Note: @jni suggested at least caching indices for sphere/circle for given radius

napari/_qt/layers/qt_labels_layer.py
```diff
@@ -315,6 +331,17 @@ def change_color_mode(self, new_mode):
...
         self.layer.color_mode = new_mode
 
+    def change_brush_shape(self, brush_shape):
+        """Change paintbrush shape of label layer.
+
+        Parameters
+        ----------
+        brush_shape : str
+            SQUARE (default) uses square paintbrush.
+            CIRCLE uses circle paintbrush.
```
jni: Can you specify these as strings (CIRCLE -> 'circle') and also specify whether these are case sensitive?

napari/layers/labels/_labels_constants.py
```diff
@@ -48,4 +48,17 @@ class LabelColorMode(StringEnum):
     DIRECT = auto()
 
 
+class LabelBrushShape(StringEnum):
+    """
+    LabelBrushShape: Labelling brush shape.
+
+    SQUARE (default) uses square paintbrush.
+
+    CIRCLE uses circle paintbrush.
```
jni: Having played around with this, I suggest using circle as the default, honestly! It's good stuff!

napari/layers/labels/labels.py
```diff
@@ -665,6 +679,44 @@ def fill(self, coord, new_label, refresh=True):
...
+            slice(-vol_radius, vol_radius + 1) for __ in range(sphere_dims)
+        ]
+
+        indices = np.mgrid[sliced_index].T.reshape(-1, sphere_dims)
+        distances = np.mgrid[sliced_dist].T.reshape(-1, sphere_dims)
+
+        distances = distances ** 2
+        distances = np.sqrt(np.sum(distances, axis=1))
+        # Use distances within desired radius to mask indices in grid
+        mask_indices = indices[distances <= radius].astype(int)
```
jni: This can be simplified quite a bit, and sped up by avoiding a square root:

```suggestion
        indices_slice = [slice(-vol_radius, vol_radius + 1),] * sphere_dims
        indices = np.mgrid[indices_slice].T.reshape(-1, sphere_dims)
        distances_sq = np.sum(indices ** 2, axis=1)
        # Use distances within desired radius to mask indices in grid
        mask_indices = indices[distances_sq <= radius ** 2].astype(int)
```

napari/layers/labels/labels.py
```diff
@@ -682,46 +734,64 @@ def paint(self, coord, new_label, refresh=True):
...
+            # on edge by keeping radius as ##.5
+            radius = np.floor(self.brush_size / 2) + 0.5
+            mask_indices = self.sphere_indices(radius, sphere_dims)
+
+            mask_indices = mask_indices + np.round(np.array(coord)).astype(int)
+
+            # discard candidate coordinates that are out of bounds
+            discard_coords = np.logical_and(
+                ~np.any(mask_indices < 0, axis=1),
+                ~np.any(mask_indices > np.array(shape), axis=1),
```
jni: I was getting a bug on the right/bottom edge of the array when painting in `examples/add_labels.py`. That's because this should be:

```python
np.all(mask_indices < np.array(shape), axis=1),
```

(or, less intrusive:

```python
~np.any(mask_indices >= np.array(shape), axis=1),
```
, note the `>=` rather than `>`)

Additionally, these are not the discard coords, but rather the kept coords!

jni: Incidentally, worth writing a test for this, then fixing!

napari/layers/labels/labels.py
```diff
@@ -682,46 +734,64 @@ def paint(self, coord, new_label, refresh=True):
...
+            # Transfer valid coordinates to slice_coord,
+            # or expand coordinate if 3rd dim in 2D image
+            slice_coord_temp = [m for m in mask_indices.T]
+            if not self.n_dimensional and self.ndim > 2:
+                for j, i in enumerate(self.dims.displayed):
+                    slice_coord[i] = slice_coord_temp[j]
+                for i in self.dims.not_displayed:
+                    slice_coord[i] = slice_coord[i] * np.ones(
+                        mask_indices.shape[0], dtype=int
+                    )
```
jni, line 748:
```suggestion
```

indices in fancy indexing are broadcast together, so this step is unnecessary — integers will get broadcast normally:

```python
In [8]: arr = np.random.random((5, 12, 12))

In [9]: indices = (np.array([5, 5, 6]), np.array([8, 9, 9]))

In [10]: arr[(1,) + indices]
Out[10]: array([0.46793242, 0.57033286, 0.64759796])

In [11]: arr[(1 * np.ones(indices[0].shape, dtype=int),) + indices]
Out[11]: array([0.46793242, 0.57033286, 0.64759796])
```

napari/layers/labels/labels.py
```diff
@@ -731,7 +801,11 @@ def paint(self, coord, new_label, refresh=True):
                 keep_coords = self.data[slice_coord] == self.selected_label
             else:
                 keep_coords = self.data[slice_coord] == self._background_label
-            self.data[slice_coord][keep_coords] = new_label
+            if self.brush_shape == "circle":
+                slice_coord = tuple(sc[keep_coords] for sc in slice_coord)
+                self.data[slice_coord] = new_label
+            else:
+                self.data[slice_coord][keep_coords] = new_label
```
jni, line 770: This new if/else is bugging me but after 30min of playing around with it in my head, I can't come up with something better, so I guess it stays for now! 😂

jni (changes requested):
@hectormz there's one remaining bug, and I think circle should be the default now, and finally I've suggested a few code simplifications, but this is (a) very very close, and (c) SO COOL, feels so good painting `examples/add_labels.py` with a circular brush!

sofroniewn:
@hectormz can you also add a benchmark `time_paint_circle` and maybe modify this one to be `time_paint_square` https://github.com/napari/napari/blob/master/napari/benchmarks/benchmark_labels_layer.py#L47-L49 and this one https://github.com/napari/napari/blob/master/napari/benchmarks/benchmark_labels_layer.py#L106-L108 and then re-run the benchmarks, see info https://github.com/napari/napari/blob/master/docs/source/developers/BENCHMARKS.md. (I'll note last time I tried to run the benchmarks I needed to add an `__init__.py` to the `napari/benchmarks` folder. Not sure if @DragaDoncila had to do that too when running benchmarks recently.

If you can then paste the paint circle vs square results for both the 2D and 3D data case into this issue (across the whole parameter sweep we have setup) that would be great! I think we should see if the circle comes with a performance hit, especially before deciding whether to make it the default or not.

napari/layers/labels/_labels_constants.py
sofroniewn: That's great! See my comment below about doing some benchmarks first, but if the performance from circle is on par with the performance of square and if the behavior is much nicer (which I could imagine) then we should go for it as default!

napari/layers/labels/labels.py
hectormz: Ah true! I never fixed that up after refactoring to the function

napari/_qt/layers/qt_labels_layer.py
hectormz: I've never written this before, just followed the pattern. Are you looking for something like:
```
        ----------
        brush_shape : str
            'square' (default) uses square paintbrush (case sensitive).
            'circle' uses circle paintbrush (case sensitive).
```

napari/layers/labels/labels.py
hectormz: Great, will account for this in a test!

napari/layers/labels/labels.py
hectormz: Right, originally I tried `self.data[slice_coord][keep_coords] = new_label` for both, but when `slice_coord` is a tuple of arrays, I guess we are just getting a view of the numpy array, and the values of `.data` aren't actually being updated

napari/layers/labels/labels.py
hectormz: @jni I added this part because if a dimension is in `dims.not_displayed`, it's position in `slice_coord` tuple is an integer and not an array.

For example, if we had a 3D stack, but we're only looking at two dimensions, `slice_coords` would be something like:

```python
slice_coords = (10, np.array([1, 2, 3]), np.array([1, 2, 3]))
```

which fails for me, so I used this section of code to get:
```python
slice_coords = (np.array([10, 10, 10]), np.array([1, 2, 3]), np.array([1, 2, 3]))
```

hectormz (commented):
@sofroniewn @jni here is my benchmark results (I just pushed the updated benchmarks too). In this case, is the parameter sweep for brush size, or number of brush strokes?

## 2D
```
(napari-venv) > asv run -E existing -b Labels2DSuite 
· No executable found for python 3.7
· Discovering benchmarks
· Running 12 total benchmarks (1 commits * 1 environments * 12 benchmarks)
[  0.00%] ·· Benchmarking existing-pyc__users_quique_.virtualenvs_napari-venv_scripts_python.exe
[ 12.50%] ··· Running (benchmark_labels_layer.Labels2DSuite.time_create_layer--).
[ 16.67%] ··· Running (benchmark_labels_layer.Labels2DSuite.time_fill--).
[ 20.83%] ··· Running (benchmark_labels_layer.Labels2DSuite.time_get_value--).
[ 25.00%] ··· Running (benchmark_labels_layer.Labels2DSuite.time_paint_circle--).
[ 29.17%] ··· Running (benchmark_labels_layer.Labels2DSuite.time_paint_square--).
[ 33.33%] ··· Running (benchmark_labels_layer.Labels2DSuite.time_raw_to_displayed--).
[ 37.50%] ··· Running (benchmark_labels_layer.Labels2DSuite.time_refresh--).
[ 41.67%] ··· Running (benchmark_labels_layer.Labels2DSuite.time_save_history--).
[ 45.83%] ··· Running (benchmark_labels_layer.Labels2DSuite.time_set_view_slice--).
[ 50.00%] ··· Running (benchmark_labels_layer.Labels2DSuite.time_update_thumbnail--).
[ 54.17%] ··· benchmark_labels_layer.Labels2DSuite.mem_data                                                                                                                                                  ok
[ 54.17%] ··· ======== =======
               param1
              -------- -------
                 16     1.14k
                 32     4.21k
                 64     16.5k
                128     65.6k
                256      262k
                512     1.05M
                1024    4.19M
                2048    16.8M
                4096    67.1M
              ======== =======

[ 58.33%] ··· benchmark_labels_layer.Labels2DSuite.mem_layer                                                                                                                                                 ok
[ 58.33%] ··· ======== =====
               param1
              -------- -----
                 16     792
                 32     792
                 64     792
                128     792
                256     792
                512     792
                1024    792
                2048    792
                4096    792
              ======== =====

[ 62.50%] ··· benchmark_labels_layer.Labels2DSuite.time_create_layer                                                                                                                                         ok
[ 62.50%] ··· ======== =============
               param1
              -------- -------------
                 16      7.59±0.1ms 
                 32     7.65±0.06ms 
                 64      8.05±0.1ms 
                128      8.75±0.2ms 
                256      12.6±0.7ms 
                512      36.8±0.8ms 
                1024      137±4ms   
                2048      575±50ms
                4096     2.20±0.03s
              ======== =============

[ 66.67%] ··· benchmark_labels_layer.Labels2DSuite.time_fill                                                                                                                                                 ok
[ 66.67%] ··· ======== ==========
               param1
              -------- ----------
                 16     29.9±1μs
                 32     30.7±1μs
                 64     31.8±2μs
                128     32.7±1μs
                256     32.9±2μs
                512     33.7±4μs
                1024    32.7±8μs
                2048    30.6±1μs
                4096    32.1±2μs 
              ======== ==========

[ 70.83%] ··· benchmark_labels_layer.Labels2DSuite.time_get_value                                                                                                                                            ok
[ 70.83%] ··· ======== ============
               param1
              -------- ------------
                 16      20.7±2μs
                 32     17.9±0.3μs
                 64     18.7±0.5μs
                128      19.4±2μs
                256      18.1±2μs
                512      18.8±2μs
                1024    17.9±0.5μs
                2048    17.8±0.8μs
                4096     17.6±2μs
              ======== ============

[ 75.00%] ··· benchmark_labels_layer.Labels2DSuite.time_paint_circle                                                                                                                                         ok
[ 75.00%] ··· ======== =============
               param1
              -------- -------------
                 16     1.25±0.05ms
                 32      1.32±0.1ms
                 64     1.43±0.08ms
                128     1.57±0.05ms
                256       3.16±1ms
                512       12.5±2ms
                1024      45.0±5ms  
                2048      187±3ms
                4096     852±200ms
              ======== =============

[ 79.17%] ··· benchmark_labels_layer.Labels2DSuite.time_paint_square                                                                                                                                         ok
[ 79.17%] ··· ======== ============
               param1
              -------- ------------
                 16     1.65±0.6ms
                 32     1.80±0.5ms 
                 64     1.83±0.5ms
                128      4.02±2ms
                256      4.97±3ms
                512      19.7±3ms
                1024     57.5±9ms
                2048     200±60ms
                4096    779±400ms
              ======== ============

[ 83.33%] ··· benchmark_labels_layer.Labels2DSuite.time_raw_to_displayed                                                                                                                                     ok
[ 83.33%] ··· ======== =============
               param1
              -------- -------------
                 16      12.8±0.9μs
                 32       19.2±3μs
                 64      43.0±0.5μs
                128       159±8μs
                256      536±200μs
                512     4.48±0.08ms
                1024      20.7±1ms  
                2048      88.8±7ms
                4096      367±60ms
              ======== =============

[ 87.50%] ··· benchmark_labels_layer.Labels2DSuite.time_refresh                                                                                                                                              ok
[ 87.50%] ··· ======== =============
               param1
              -------- -------------
                 16     1.09±0.02ms
                 32     1.20±0.06ms 
                 64     1.21±0.08ms
                128     1.41±0.05ms
                256       3.05±2ms
                512       10.8±2ms
                1024      43.4±4ms
                2048      195±30ms
                4096     743±200ms
              ======== =============

[ 91.67%] ··· benchmark_labels_layer.Labels2DSuite.time_save_history                                                                                                                                         ok
[ 91.67%] ··· ======== ============
               param1
              -------- ------------
                 16     5.04±0.9μs
                 32     4.42±0.3μs
                 64     5.07±0.8μs
                128     9.24±0.9μs
                256      45.0±2μs
                512      210±9μs
                1024    823±100μs
                2048    3.45±0.2ms
                4096     14.8±2ms
              ======== ============

[ 95.83%] ··· benchmark_labels_layer.Labels2DSuite.time_set_view_slice                                                                                                                                       ok
[ 95.83%] ··· ======== ============
               param1
              -------- ------------
                 16      111±10μs
                 32      132±20μs
                 64      244±60μs
                128      450±70μs
                256     1.49±0.1ms 
                512     9.82±0.5ms
                1024     44.5±5ms
                2048     181±10ms
                4096    926±100ms
              ======== ============

[100.00%] ··· benchmark_labels_layer.Labels2DSuite.time_update_thumbnail                                                                                                                                     ok
               param1
              -------- ------------
                 16      774±70μs
                 32      758±30μs
                 64      809±60μs
                128     855±100μs
                256     982±400μs
                512     1.07±0.4ms
                1024    798±200μs
                2048     730±7μs
                4096     800±30μs
              ======== ============
```
and
## 3D
```
(napari-venv) >asv run -E existing -b Labels3DSuite 
· No executable found for python 3.7
· Discovering benchmarks
· Running 12 total benchmarks (1 commits * 1 environments * 12 benchmarks)
[  0.00%] ·· Benchmarking existing-pyc__users_quique_.virtualenvs_napari-venv_scripts_python.exe
[ 12.50%] ··· Running (benchmark_labels_layer.Labels3DSuite.time_create_layer--).
[ 16.67%] ··· Running (benchmark_labels_layer.Labels3DSuite.time_fill--).
[ 20.83%] ··· Running (benchmark_labels_layer.Labels3DSuite.time_get_value--).
[ 25.00%] ··· Running (benchmark_labels_layer.Labels3DSuite.time_paint_circle--).
[ 29.17%] ··· Running (benchmark_labels_layer.Labels3DSuite.time_paint_square--).
[ 33.33%] ··· Running (benchmark_labels_layer.Labels3DSuite.time_raw_to_displayed--).
[ 37.50%] ··· Running (benchmark_labels_layer.Labels3DSuite.time_refresh--).
[ 41.67%] ··· Running (benchmark_labels_layer.Labels3DSuite.time_save_history--).
[ 45.83%] ··· Running (benchmark_labels_layer.Labels3DSuite.time_set_view_slice--).
[ 50.00%] ··· Running (benchmark_labels_layer.Labels3DSuite.time_update_thumbnail--).
[ 54.17%] ··· benchmark_labels_layer.Labels3DSuite.mem_data                                                                                                                                                  ok
[ 54.17%] ··· ======== =======
               param1
              -------- -------
                 16     16.5k 
                 32      131k
                 64     1.05M
                128     8.39M
                256     67.1M
                512      537M
                1024    4.29G
              ======== =======

[ 58.33%] ··· benchmark_labels_layer.Labels3DSuite.mem_layer                                                                                                                                                 ok
[ 58.33%] ··· ======== =====
               param1
              -------- -----
                 16     792
                 32     792
                 64     792
                128     792
                256     792
                512     792
                1024    792
              ======== =====

[ 62.50%] ··· benchmark_labels_layer.Labels3DSuite.time_create_layer                                                                                                                                         ok
[ 62.50%] ··· ======== ============
               param1
              -------- ------------
                 16     9.59±0.7ms
                 32     9.10±0.5ms 
                 64     8.63±0.3ms
                128     9.45±0.2ms
                256      12.4±1ms
                512      38.6±1ms
                1024    443±100ms
              ======== ============

[ 66.67%] ··· benchmark_labels_layer.Labels3DSuite.time_fill                                                                                                                                                 ok
[ 66.67%] ··· ======== ==========
               param1
              -------- ----------
                 16     32.6±3μs
                 32     36.7±3μs
                 64     36.1±4μs
                128     41.9±7μs
                256     37.1±4μs
                512     41.9±7μs
                1024    52.6±9μs
              ======== ==========

[ 70.83%] ··· benchmark_labels_layer.Labels3DSuite.time_get_value                                                                                                                                            ok
[ 70.83%] ··· ======== ============
               param1
              -------- ------------
                 16      20.2±2μs
                 32      19.8±1μs
                 64     19.2±0.6μs
                128      18.9±1μs
                256     19.0±0.7μs
                512      21.3±3μs
                1024    20.1±0.9μs
              ======== ============

[ 75.00%] ··· benchmark_labels_layer.Labels3DSuite.time_paint_circle                                                                                                                                         ok
[ 75.00%] ··· ======== =============
               param1
              -------- -------------
                 16     1.46±0.09ms
                 32     1.50±0.07ms
                 64      1.56±0.1ms
                128     1.76±0.05ms
                256      3.21±0.5ms
                512      11.2±0.3ms
                1024      114±30ms
              ======== =============

[ 79.17%] ··· benchmark_labels_layer.Labels3DSuite.time_paint_square                                                                                                                                         ok
[ 79.17%] ··· ======== ============
               param1
              -------- ------------
                 16     1.96±0.9ms
                 32     1.67±0.3ms
                 64     1.53±0.1ms
                128     1.81±0.1ms
                256     3.11±0.2ms
                512      12.0±1ms
                1024     141±40ms
              ======== ============

[ 83.33%] ··· benchmark_labels_layer.Labels3DSuite.time_raw_to_displayed                                                                                                                                     ok
[ 83.33%] ··· ======== ============
               param1
              -------- ------------
                 16      12.4±1μs
                 32      20.4±2μs
                 64      44.0±1μs
                128      144±4μs
                256     653±200μs
                512     4.69±0.1ms
                1024     123±50ms
              ======== ============

[ 87.50%] ··· benchmark_labels_layer.Labels3DSuite.time_refresh                                                                                                                                              ok
[ 87.50%] ··· ======== ============
               param1
              -------- ------------
                 16     1.20±0.1ms
                 32     1.34±0.2ms
                 64     1.73±0.3ms
                128     1.55±0.1ms
                256     2.74±0.3ms
                512     11.6±0.5ms
                1024     166±60ms
              ======== ============

[ 91.67%] ··· benchmark_labels_layer.Labels3DSuite.time_save_history                                                                                                                                         ok
[ 91.67%] ··· ======== ============
               param1
              -------- ------------
                 16      34.4±1μs
                 32     34.6±0.9μs
                 64      37.7±1μs
                128     59.3±10μs
                256      136±20μs
                512      241±10μs
                1024    1.56±0.4ms
              ======== ============

[ 95.83%] ··· benchmark_labels_layer.Labels3DSuite.time_set_view_slice                                                                                                                                       ok
[ 95.83%] ··· ======== ============
               param1
              -------- ------------
                 16      153±4μs
                 32      191±20μs
                 64      333±80μs
                128      502±40μs
                256     1.47±0.1ms
                512      12.5±1ms
                1024     147±3ms
              ======== ============

[100.00%] ··· benchmark_labels_layer.Labels3DSuite.time_update_thumbnail                                                                                                                                     ok
[100.00%] ··· ======== ==========
               param1
              -------- ----------
                 16     715±20μs
                 32     778±60μs
                 64     727±20μs
                128     763±60μs
                256     728±40μs
                512     748±10μs
                1024    836±30μs
              ======== ==========
```

sofroniewn:
The parameter is the height of the image - the `circle` brush is faster!!  `12.5±2ms` vs `19.7±3ms` to paint a single spot in a `512x512` image - we should definitely go with the circle then as default!! So great to see the result of the benchmark, it really helps with decisions like this!!

[Edit: I don't see the benchmark code committed, but these times might be longer for whichever one is not default if time spent changing the brush is included - regardless benchmarks look comparable and good overall]

hectormz:
@sofroniewn sorry I just pushed! I also realized that for 3D I ran 

```
asv dev -b Labels3DSuite
```

```
asv run -E existing -b Labels3DSuite
```
and not sure how exactly they compare, so I was rerunning now.

hectormz:
@sofroniewn corrected the 3D run, but yes, assuming I updated the benchmarks correctly, the circle brush does look faster.

sofroniewn:
> assuming I updated the benchmarks correctly, the circle brush does look faster.

Yep, benchmarks look good - so much for my intuition though 🤣

hectormz:
I'm curious if the apparent speed difference is due to brush creation, or painting fewer pixels. The square brush width should match the circle diameter.

sofroniewn:
> I'm curious if the apparent speed difference is due to brush creation, or painting fewer pixels. The square brush width should match the circle diameter.

Indeed I think now that could be a major driver - I'll note that `12.2/19.7 => 61%`  and `3.141/4 => 78.5%` so it seems even faster than would be purely explained by that (assuming by approximations for number of pixels edited is valid, which is only true in limit).

napari/_qt/layers/qt_labels_layer.py
jni: Yup!

But *is it* case sensitive?

napari/layers/labels/labels.py
jni: > which fails for me,

Can you demonstrate this failure? If you look at the code I showed above, it works totally fine!

napari/layers/labels/labels.py
```diff
@@ -152,6 +153,7 @@ def __init__(
             colormaps.label_colormap(self.num_colors),
         )
         self._color_mode = LabelColorMode.AUTO
+        self._brush_shape = LabelBrushShape.SQUARE
```
sofroniewn: Let's make default `circle`.

napari/layers/labels/labels.py
```diff
@@ -665,6 +679,35 @@ def fill(self, coord, new_label, refresh=True):
         if refresh is True:
             self.refresh()
 
+    @lru_cache(maxsize=64)
+    def sphere_indices(self, radius, sphere_dims):
```
sofroniewn: This should be function in `_labels_utils.py` that then gets its own tests in `_testst/test_labels_utils.py` and then be imported in here and used rather than being on the object itself.

Ideally we try and keep methods unlikely to be directly used by the user off the main object, and have as simple functions as we can in our utils so they can be tested in isolation

sofroniewn (commented):
I love it!! Just gave it a spin, all functionality is working for me, but let's make circle the default, it's so much nicer!!!

I've left a couple additional minor comments (like moving a function to our utils and testing it independently)

napari/layers/labels/labels.py
hectormz: Ah yes, that does work fine, and I see that now!

So, the problem arises however when `preserve_labels = True`, and below we have:

```python
slice_coord = tuple(sc[keep_coords] for sc in slice_coord)
```
and when one of the items in the tuple is an `int` instead of an array, it throws an exception.

that line would have to become

```python
slice_coord = tuple(sc[keep_coords] if isinstance(sc, np.ndarray) else sc for sc in slice_coord)
```
Or break it out of list comprehension really.


All of these problems melt away if we could just do 

```python
self.data[slice_coord][keep_coords] = new_label
```

downstream!!! ❗

napari/layers/labels/labels.py
hectormz: Got it!

I'm also realizing I mucked up the tests, since I wasn't really testing the mouse interactions.

I should have put all those new tests in `test_labels.py`, right?

Or at least only add one new circle test to `test_labels_mouse_bindings.py`.

sofroniewn: for tests it would be nice to use a parameterized test where you move over the brush - i.e. from `circle` to `square`, see this example here https://github.com/napari/napari/blob/master/napari/_tests/test_dtypes.py#L4-L20 that will reduce some of the code duplication.

sofroniewn: actually that might be hard as you then need to pass the shape of what things need to look like too, maybe @jni can advise here, i'm signing off for the night

napari/layers/labels/labels.py
jni: Ooooooooh! =) Ok, all good, let's keep the current scheme! 🎉

napari/layers/labels/labels.py
hectormz: I _think_ that should be doable if the test is designed for it. 

I've been lazy with the circle tests and just checked if there are the expected number of painted voxels/pixels for each label. It doesn't actually check the location. 

But we can parametrize the square and circle paintbrushes, and then verify that the number of labeled pixels in specific volumes are appropriate for a square or circle brush.

So our expected values for the param will just be two lists of values

napari/layers/labels/labels.py
hectormz: Sorry 😔. 

Now that the circle brushes have tests, I welcome someone to come in and hack it up with a more elegant solution

jni (approved):

napari/_qt/layers/qt_labels_layer.py
hectormz: Apparently it's not!

hectormz:
To Do:
- ~~Update `brush_shape` case insensitivity (it's not case sensitivite)~~
- Update/Move/Parameterize tests
- ~~Make circle default brush shape~~

jni:
@sofroniewn @hectormz as far as I understand it, this is ready to go but the tests could be a bit better? But they already cover all the functionality, don't they? If so, then maybe we should just merge and leave test refactors for a later PR?

sofroniewn:
> @sofroniewn @hectormz as far as I understand it, this is ready to go but the tests could be a bit better? But they already cover all the functionality, don't they? If so, then maybe we should just merge and leave test refactors for a later PR?

Yes, I agree - tests cover all functionality and it works great. I actually don't think we need to parameterize them at this point, they are very readable as is. We can always update them later if we want.

I'm going to merge then in the next couple hours unless i hear anything from @hectormz - great job with this!!

sofroniewn (approved):

hectormz:
@sofroniewn I'd like to get this done today, but if I haven't pushed changes and you're otherwise ready, I can make a separate PR. 

As I see it, parameterizing could hopefully simplify things. Maybe the most pressing thing in my mind is that I put all my circle brush tests in the mouse click section, but a lot of them can really be in the normal labels tests.

sofroniewn:
> @sofroniewn I'd like to get this done today, but if I haven't pushed changes and you're otherwise ready, I can make a separate PR.
>
> As I see it, parameterizing could hopefully simplify things. Maybe the most pressing thing in my mind is that I put all my circle brush tests in the mouse click section, but a lot of them can really be in the normal labels tests.

Sounds great!

hectormz:
@jni  @sofroniewn I've moved the tests and parametrized some of the tests. Currently, my tests verify that within a square (2d) or cube(3d), there is an expected number of painted pixels for each label. This handles diff number of expected pixels for square or circle brush. In the 3D test, it also checks for preserve_labels behaving correctly.

These tests could in theory replace `test_paint_with_preserve_labels`, `test_paint` in `test_labels.py` as-is, or with more work.

One difference in `test_paint` is that it tests painting on top of randomized, existing labels. That works for a square brush, but is hard to test with a circle brush because within that bounding box, there may be matching labels that already existed.

hectormz:
Should circle/brush shape switching be extended to the eraser too? Seemingly it is as simple as changing line 501 in `labels.py` from:

```python
self.cursor = "square"
```

```python
self.cursor = self.brush_shape
```

sofroniewn:
> Should circle/brush shape switching be extended to the eraser too? Seemingly it is as simple as changing line 501 in labels.py from:

yes - i'd change that before we merge

sofroniewn:
Tests look great, much cleaner - thanks @hectormz 

If you can add the `self.cursor = self.brush_shape` for eraser and maybe a add test for that - just to check that eraser always has correct brush shape I will then merge

hectormz:
Great! Change and tests coming in within minutes!

hectormz:
@sofroniewn the updated test for the eraser now expects different pixel labeling results for circle vs square brushes.

PS, running/debugging parametrized tests in VSCode is 🎉

sofroniewn:
Beautiful - merging now - thanks for all the hard work here!!

tlambert03:
Congrats on a lovely PR @hectormz!

jni:
😍 ❤️

hectormz:
Eventually I hope to have closed more issues than I've created! 
Thanks for all the input and guidance, everyone!

PR #1431, opened 2020-07-11
Reorganize _qt module

tlambert03:
# Description
The _qt folder is starting to get very large as a flat directory, so this reorganizes things that are widgets (anything that inherits from `QWidget`, like one-off dialogs, etc...) into `_qt/widgets`, and renames the qt controls files to say "controls" in the filename rather than "layer" (like the classes do).  eg `qt_image_layer.py` -> `qt_image_controls.py` and the stacked widget is now "`QtLayerControlsContainer`" instead of "`QtControls`" ... which I think was hard to tell apart from `QtLayerControls` (the thing it actually contains).
Many of the imports were also starting to get ragged, so I used iSort on everything I could.

**don't be scared** :) It looks like a lot of changes.  but they are strictly file moves/renames and import sorting.  (With one exception, moving `create_qt_layer_controls` into the only file that uses it, and removing a file.)

## Type of change

- [x] refactor 

# How has this been tested?

## Final checklist:


jni:
@tlambert03 I guess this should wait until after the sprint, right?

tlambert03:
Yes for sure :). No hurry

sofroniewn:
Overall looking pretty good - I might rename `_qt/layers` to `_qt/layer_controls` now?

We can also gradually run iSort on more of the codebase too

tlambert03:
@sofroniewn,  I changed `layers` to `layer_controls`

sofroniewn:
See comment here

> I have found giving a file in your project the same name as a built-in module leads to surprisingly confusing errors. It's a very easy mistake to make. It's not chance you name it X.py precisely because it wraps or deals with the built-in module X.

and [discussion in zulip](https://napari.zulipchat.com/#narrow/stream/212875-general/topic/Qt.20error.20in.20theme_sample/near/204271387) - we might want to rename `threading.py` inside our qt folder too, could be done here

tlambert03:
>  we might want to rename threading.py inside our qt folder too, could be done here

I changed `_qt.threading` to `_qt.qthreading` but I actually don't agree with / want to change the public facing `napari.qt.threading` module.  I tend to agree more with @Czaki's comment:

> It is bad idea to call file from module as python script since python 3 force from . import threading instead of import threading when import from module there is no option for collision.

i.e. people should not be running `python napari/qt/module.py` to execute some module deep in the napari package as a script.  (If they must execute it, it should be `python -m napari.qt.module`)  so yes, it can be confusing the first time you attempt it, and I agree that it would be nice if python were able to figure that out and warn you... but I don't think everyone should go avoiding every possible builtin module name deep in their own packages.

sofroniewn:
I think now that #1453 has merged I think this would be a good time to get this merged if you want to revisit @tlambert03

tlambert03:
Does anyone else want this? Or just don't care either way

sofroniewn:
> Does anyone else want this? Or just don't care either way

I like reorganizing the `_qt` folder, I'm pretty open to things, but looking at PR now I like the new `perfmon` folder - I might call it `debug` or `tracing`.

I like `layer_controls` too. I might consider putting `qt_color_dialog.py` in with `layer_controls` as it only gets used there I think.

I might want to separate out `dialogs` like `qt_about.py` and `qt_about_key_bindings.py` into their own folder

Maybe have one for `plugins` where we put `qt_plugin_report.py`, `qt_plugin_sorter.py`, `qt_plugin_table.py`.

We could have a folder `components` where the `qt_layerlist.py`, `qt_dims.py`, `qt_dims_slider.py`, `qt_console.py`, the qtviewer live.

Then maybe in widgets I'd just keep the really napari agnostic ones like the range slider, popup, scrollbar etc.

Basically I'm saying I like the direction and think we should organize more :-) how does that sound to you ?

tlambert03:
ok, I'm back to wanting this again... I pulled dialogs into a new folder and renamed perfmon to tracing.  I think I'd rather not continue to subdivide the widgets into things that mirror organization elsewhere in the repo (like components) for now.

jni:
I agree with everything that's been said here, in order. =) I don't like mirroring trees, especially when the mapping is not strictly 1-1. I haven't looked at the diff but from @tlambert03's description I like the reorg.

sofroniewn (approved):
Ok this looks good to me @tlambert03! Thanks for keeping it updated, let's get it merged, the new folder structure is much easier to parse.

PR #1433, opened 2020-07-12
Fix viewing properties for mutliscale labels layers

### LINKED ISSUE #1397: Label properties status message doesn't work for multiscale labels
## 🐛 Bug

In #1281, we added the ability of labels layers to receive a properties dictionary, which would then be displayed per-label in the status bar. This is working great but is broken when using a multiscale labels pyramid.

## To Reproduce

First, try it out with single-scale labels:

```python
import numpy as np
import napari

labels = np.zeros((10_000, 10_000))
labels[5000:] += 1
labels[:, 5000:] += 2

properties = {'class': ['background', 'pears', 'apples', 'oranges']}

v = napari.view_labels(labels, properties=properties)
```

This gives the following behaviour:

Screen Shot 2020-07-02 at 3 11 53 pm

Now, try:

```python
labels_pyramid = [labels[::i, ::i] for i in [1, 2, 4, 8]]

v = napari.view_labels(labels_pyramid, properties=properties)
```

Now, it appears that the properties for each label does not exist:

Screen Shot 2020-07-02 at 3 15 51 pm

This happens regardless of the zoom level.

## Expected behavior

The property corresponding to the label value under the cursor should be displayed, regardless of zoom level.

## Environment

```
napari: 0.3.6.dev11+ge43f730
Qt: 5.14.2
PyQt5: 5.14.2
NumPy: 1.17.2
SciPy: 1.3.1
Dask: 2.6.0
VisPy: 0.6.4

GL version:  2.1 NVIDIA-12.0.24 355.11.10.50.10.103
MAX_TEXTURE_SIZE: 16384

Plugins:
  - napari-plugin-engine: 0.1.6
  - svg: 0.1.3
```

DragaDoncila:
# Description
Change status message to check whether image is multiscale before grabbing the label value. Fixes #1397 .

## Type of change

# How has this been tested?
Added test specifically for multiscale labels layer.

## Final checklist:

### DIFF
--- a/napari/layers/labels/_tests/test_labels.py
+++ b/napari/layers/labels/_tests/test_labels.py
@@ -236,6 +236,41 @@ def test_properties():
     assert layer_message.endswith('Class 12')
 
 
+def test_multiscale_properties():
+    """Test adding labels with multiscale properties."""
+    np.random.seed(0)
+    data0 = np.random.randint(20, size=(10, 15))
+    data1 = data0[::2, ::2]
+    data = [data0, data1]
+
+    layer = Labels(data)
+    assert isinstance(layer.properties, dict)
+    assert len(layer.properties) == 0
+
+    properties = {'class': ['Background'] + [f'Class {i}' for i in range(20)]}
+    label_index = {i: i for i in range(len(properties['class']))}
+    layer = Labels(data, properties=properties)
+    assert isinstance(layer.properties, dict)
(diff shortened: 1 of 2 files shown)


tlambert03 (approved):
looks great!  thanks for the test too

sofroniewn (approved):
LGTM!

Issue #1141, opened 2020-04-14
Increment dimensions slider with ctrl + mouse scroll


hectormz:
## 🚀 Feature

I'd like to enable scrolling through an image stack using `ctrl + scroll`

## Motivation

My standard image is usually a stack of several thousand slices that I want to look through quickly. To navigate through the stack, I either manipulate the scroller on the bottom, or use the left/right arrow keys. Using the scroll would be more convenient for me.

## Pitch

In `napari` the mouse scroll is used for zoom, but `ctrl + scroll` could be used to scroll the image dimensions in addition to the arrow keys.

As an aside (or future issue if it doesn't exist), sometimes I scroll through stacks way more often than zooming into an image, and it would be great to rearrange or change key bindings (like in `Image J`) so I could choose to make mouse scrolling for dimension scrolling instead of zooming. These settings could be saved in `~/.naparirc` or similar.

sofroniewn:
Yes, that makes a lot of sense - this functionality would be enabled by the proposed changes in #1048 (cc @kne42). Thanks for the feature request!!

hectormz:
@sofroniewn awesome, I'll loop back to this after #1048 then

jni:
As I was reading through the issues, it occurred to me that horizontal scrolling does nothing right now and could also be used to move through a stack. Horizontal scrolling is very easy and natural on Mac trackpads.

hectormz:
Is horizontal scrolling on Mac trackpad simply swiping fingers horizontally? Is that achieved on a Mac with a  mouse by scrolling and holding a key?

jni:
This was closed by #1434! 🎉

hectormz:
Woohoo! Thanks @DragaDoncila !!

PR #1434, opened 2020-07-12
Add CTRL+ Mouse Scroll to scroll through last active stack

DragaDoncila:
# Description
Add event and binding for scrolling through stack using CTRL+Mouse Wheel, as per #1141 

## Type of change

# How has this been tested?

## Final checklist:

### DIFF
--- a/examples/custom_mouse_functions.py
+++ b/examples/custom_mouse_functions.py
@@ -16,6 +16,10 @@ with napari.gui_qt():
     labeled = ndi.label(blobs)[0]
     labels_layer = viewer.add_labels(labeled, name='blob ID')
 
+    @viewer.mouse_press_callbacks.append
+    def get_event(viewer, event):
+        print(event)
+
     @viewer.mouse_drag_callbacks.append
     def get_ndisplay(viewer, event):
         if 'Alt' in event.modifiers:
@@ -39,7 +43,9 @@ with napari.gui_qt():
                 data[binary_new] = val
             size = np.sum(binary_new)
             layer.data = data
-            msg = f'clicked at {cords} on blob {val} which is now {size} pixels'
+            msg = (
+                f'clicked at {cords} on blob {val} which is now {size} pixels'
+            )
         else:
             msg = f'clicked at {cords} on background which is ignored'
         layer.status = msg
--- a/napari/_qt/qt_viewer.py
+++ b/napari/_qt/qt_viewer.py
@@ -36,7 +36,11 @@ from .qt_performance import QtPerformance
 from .._vispy import create_vispy_visual
 
 
-class OurSceneCanvas(SceneCanvas):
+class KeyModifierFilterSceneCanvas(SceneCanvas):
+    """SceneCanvas which overrides default vispy behaviour for events with key modifiers.F
+
+    """
+
     def _process_mouse_event(self, event):
         if event.type == 'mouse_wheel' and len(event.modifiers) > 0:
             return
@@ -149,7 +153,9 @@ class QtViewer(QSplitter):
             self.toggle_console_visibility
         )
 
-        self.canvas = OurSceneCanvas(keys=None, vsync=True, parent=self)
+        self.canvas = KeyModifierFilterSceneCanvas(
+            keys=None, vsync=True, parent=self
+        )
         self.canvas.events.ignore_callback_errors = False
         self.canvas.events.draw.connect(self.dims.enable_play)
         self.canvas.native.setMinimumSize(QSize(200, 200))
--- a/napari/_qt/qt_viewer.py
+++ b/napari/_qt/qt_viewer.py
@@ -22,6 +22,7 @@ from ..utils.interactions import (
     mouse_press_callbacks,
     mouse_move_callbacks,
     mouse_release_callbacks,
+    mouse_wheel_callbacks,
 )
 from ..utils.key_bindings import components_to_key_combo
 from ..utils import perf
@@ -152,6 +153,7 @@ class QtViewer(QSplitter):
         self.canvas.connect(self.on_mouse_release)
         self.canvas.connect(self.on_key_press)
         self.canvas.connect(self.on_key_release)
+        self.canvas.connect(self.on_mouse_wheel)
 
         self.view = self.canvas.central_widget.add_view()
         self._update_camera()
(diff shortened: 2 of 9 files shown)


tlambert03:
@DragaDoncila, what still needs to be done here? (haven't looked at the code)

DragaDoncila:
@tlambert03 the functionality is all there, but we need to refactor a few things and add tests. I'll go through and comment where we need to refactor. I can take a look at this tomorrow, but if you'd like to push it through then of course, go for it.

napari/_qt/qt_viewer.py
```diff
@@ -35,6 +36,13 @@
 from .._vispy import create_vispy_visual
 
 
+class OurSceneCanvas(SceneCanvas):
+    def _process_mouse_event(self, event):
+        if event.type == 'mouse_wheel' and len(event.modifiers) > 0:
+            return
+        super()._process_mouse_event(event)
+
```
DragaDoncila, line 46: We need to move this to a more appropriate spot - I'm not sure where would be best - qt_utils? It's own class? A new dedicated spot for monkey patching vispy?

napari/components/viewer_model.py
```diff
@@ -94,8 +94,28 @@ def __init__(
         self.mouse_move_callbacks = []
         # Hold callbacks for when mouse is pressed, dragged, and released
         self.mouse_drag_callbacks = []
+        # Hold callbacks for when mouse wheel is scrolled
+
+        def _scroll(viewer, event):
+            from .._viewer_key_bindings import (
```
DragaDoncila: This import breaks at top level as a circular import, so we should pull out increment_dims_left and increment_dims_right. I think it would work well as a dims method but I think @jni had some bones to pick with that?

jni: I didn't have specific bones, I just wasn't sure. =) I think I was worried about the fact that dims will get refactored soon, but this will be affected either way. Thinking about it now, I agree, should be a Dims method with an axis kwarg and a None default, in which case last axis is used.

napari/_qt/qt_viewer.py
jni: I think this is an ok spot for it since this is where we import SceneCanvas, but maybe we should rename it KeyFilterSceneCanvas or something descriptive like that? And the class docstring should describe the reason for its existence.

@DragaDoncila note that this is no longer monkey patching, it's subclassing, which is the Right and Good way to do this =), since we won't affect other running libraries using VisPy.

napari/components/viewer_model.py
```diff
@@ -94,8 +94,28 @@ def __init__(
...
+
+            if 'Control' not in event.modifiers:
+                return
+            delta = int(event.delta[1])
+            if abs(delta) >= 1:
+                for i in range(abs(delta)):
+                    if delta < 0:
+                        increment_dims_left(self)
+                    else:
+                        increment_dims_right(self)
```
jni: The last thing that needs to be done is that we need an accumulator for trackpad scrolling, which is fractional. So we would do something like:

```python
vertical_delta = event.delta[1]
int_delta = int(vertical_delta)  # vertical scrolling
fractional_delta = 
self._cumulative_delta += vertical_delta - int_delta
if self._cumulative_delta > 1:
    int_delta += 1
    self._cumulative_delta -= 1
```

... except there's two complicating factors:

- need to handle positive and negative deltas, which the above doesn't, and
- we need to reset cumulative delta to 0 after a certain time, or after any other event, since you don't really want the fractional scrolling from 3 minutes ago sitting around to modify your next scroll.

Probably I would say merge this without fractional deltas and then raise an issue for adding fractional deltas...

napari/components/viewer_model.py
DragaDoncila: So I tried pulling this out but I don't think it's that simple? If I put `increment_dims` on Dims, then we don't have access to `last_used`, because it belongs to `qt_dims`. If we put it in `qt_dims`, then we can't access it in `viewer_model` for the `_scroll` method. 

Also, regardless of which we put it in, we still need the `increment_dims_left/right` in the key bindings, in order to bind left and right arrows, so it would just call either `qt_dims`' or `dims`' method since it has access to both.

So now I'm not too sure where to put it. Left and right arrows works with it pretty much anywhere, but the scroll not so much.

sofroniewn: > So now I'm not too sure where to put it. Left and right arrows works with it pretty much anywhere, but the scroll not so much.

For now we should probably put `last_used` on dims and remove it from QtDims, but we might want to think about another place longer term!

DragaDoncila:
@ziyangczi @jni @sofroniewn I've added last_used as an attribute to dims in this last commit, and changed references from `qt_dims.last_used` to `dims.last_used`. The code to set and update `last_used` remains in `qt_dims`. All tests pass and scrolling (with and without modifier) works as expected from what I can tell. 

Still need to add a test, but if you'd like to take a look and make sure we're all on the same page that'd be great.

napari/components/viewer_model.py
```diff
@@ -396,7 +396,7 @@ def _on_layers_change(self, event):
                 self.dims.set_range(i, r)
         self.events.layers_change()
 
-    def _scroll(self, event):
+    def _scroll(self, viewer, event):
```
jni, line 399: Why is this change needed?

jni:
@DragaDoncila very cool! Do you also want to tackle the fractional scrolling [issue](https://github.com/napari/napari/pull/1434#discussion_r456730417), or do you want to leave that for later?

napari/components/viewer_model.py
DragaDoncila: Without the viewer I got errors about the mouse callback expecting 2 arguments but receiving 3. When I looked at an older commit, we were passing the viewer so I added it back in

DragaDoncila:
@jni totally happy to do the fractional scrolling, the only part I'm not sure about is resetting after a certain time/event. It needs to be done obviously but not sure what best approach would be. I'll leave it to you to decide if it should go in this PR or a follow up

examples/custom_mouse_functions.py
```diff
@@ -16,6 +16,10 @@
     labeled = ndi.label(blobs)[0]
     labels_layer = viewer.add_labels(labeled, name='blob ID')
 
+    @viewer.mouse_press_callbacks.append
+    def get_event(viewer, event):
+        print(event)
+
```
jni, line 22: This needs to go! 😂

jni: No I take it back, I just saw it's in the examples 🤦 Clearly I need an afternoon ☕

napari/_qt/qt_viewer.py
```diff
@@ -35,6 +36,17 @@
 from .._vispy import create_vispy_visual
 
 
+class KeyModifierFilterSceneCanvas(SceneCanvas):
+    """SceneCanvas which overrides default vispy behaviour for events with key modifiers.F
+
+    """
```
jni: PEP8/PEP257: one-liners should fit in one 79c line. 
    
```suggestion
    """SceneCanvas overriding VisPy when mouse wheel events have modifiers."""
```

jni (approved):

napari/components/viewer_model.py
sofroniewn: hmm this feels a little funny, i think i see why it was needed, but we might want to improve in a followup - this scroll method shouldn't need to live on the viewer model like this - the mixing of `self` and `viewer` seems like trouble and not something we've needed to do elsewhere.

Issue #1064, opened 2020-03-19
Use scipy.spatial.cKDTree for point subsetting and selection


jni:
## 🚀 Feature
Currently, we are brute-force linear searching through points for slicing and selection. Using a cKDTree could speed things up dramatically. However, for now we would have to use some hacks because:

- slicing would require searching only along two dimensions out of many. This problem could be alleviated with the use of spacing/scales in the cKDTree, or equivalently we could modify the points going into the tree by some factor along the slicing dimension.
- currently, rectangular range searches are not supported, but there is an [issue](https://github.com/scipy/scipy/issues/7822) for it. This would be a useful contribution we could make upstream — it seems that the algorithm is not too complicated.

PR #1435, opened 2020-07-12
Points slicing benchmark

DragaDoncila:
# Description
Add benchmark to time slicing large points volume. This will also be useful when working on #1064 

## Type of change

## Final checklist:


sofroniewn (approved):
LGTM!

sofroniewn:
We should probably exclude our codecov from looking in the benchmarks folder, but that can be for a future PR. I also needed to add an `__init__.py` to the benchmarks folder to be able to run them, so that should stay too.

PR #1437, opened 2020-07-12
DOC: autoreformat of more docstrings.

Carreau:
Unlike previous PR this is

1) more conservative on whitespace at start and end of docstrings,
2) does visit the full ast, and hence also fix methods.

There are a couple of manual changes where vélin had issues, and used a
custom fork of numpydoc to handle the typoes instead of failing.

### DIFF
--- a/napari/_qt/layers/qt_labels_layer.py
+++ b/napari/_qt/layers/qt_labels_layer.py
@@ -308,10 +308,8 @@ class QtLabelsControls(QtLayerControls):
         Parameters
         ----------
         new_mode : str
-
-        AUTO (default) allows color to be set via a hash function with a seed.
-
-        DIRECT allows color of each label to be set directly by a color dictionary.
+            AUTO (default) allows color to be set via a hash function with a seed.
+            DIRECT allows color of each label to be set directly by a color dictionary.
         """
         self.layer.color_mode = new_mode
 
--- a/napari/resources/__init__.py
+++ b/napari/resources/__init__.py
@@ -65,6 +65,7 @@ def import_resources(version: str = '', overwrite: bool = False) -> str:
         Version string, by default the resources will be written to a file that
         encodes the current napari version, as well as Qt backend and version:
         ``_qt_resources_{napari.__version__}_{API}_{QT_VERSION}.py``
+
     overwrite : bool, optional
         Whether to recompile and overwrite the resources.
         Resources will be rebuilt if any of the following are True:
--- a/napari/utils/event.py
+++ b/napari/utils/event.py
@@ -585,8 +585,8 @@ class EventEmitter(object):
     def blocker(self, callback=None):
         """Return an EventBlocker to be used in 'with' statements
 
-           Notes
-           -----
+        Notes
+        -----
            For example, one could do::
 
                with emitter.blocker():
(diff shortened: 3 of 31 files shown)


napari/_qt/layers/qt_labels_layer.py
```diff
@@ -308,10 +308,8 @@ def change_color_mode(self, new_mode):
         Parameters
         ----------
         new_mode : str
-
-        AUTO (default) allows color to be set via a hash function with a seed.
-
-        DIRECT allows color of each label to be set directly by a color dictionary.
+            AUTO (default) allows color to be set via a hash function with a seed.
+            DIRECT allows color of each label to be set directly by a color dictionary.
```
Carreau, line 312: This was done manully.

napari/resources/__init__.py
```diff
@@ -65,6 +65,7 @@ def import_resources(version: str = '', overwrite: bool = False) -> str:
         Version string, by default the resources will be written to a file that
         encodes the current napari version, as well as Qt backend and version:
         ``_qt_resources_{napari.__version__}_{API}_{QT_VERSION}.py``
+
```
Carreau, line 68: The heuristic is : if one section have one blankline somewhere we'll separate items with blanklines.

napari/utils/colormaps/vendored/cm.py
```diff
@@ -242,7 +241,6 @@ def to_rgba(self, x, alpha=None, bytes=False, norm=True):
 
         If norm is False, no normalization of the input data is
         performed, and it is assumed to be in the range (0-1).
-
```
Carreau: this is likely due to the fact that this docstring has no sections with headers otherwise it would leave a blankline.

napari/utils/colormaps/vendored/cm.py
```diff
@@ -189,7 +189,6 @@ class ScalarMappable(object):
     """
```
sofroniewn: Just flagging that this change happens inside our `vendored` folder which we might want to ignore. Thoughts on this @jni? We could easily revert the few changes inside vendored files before merging, but your call

sofroniewn (approved):
LGTM! All changes are great - only question for @jni is if we want to revert changes in the ~2 files inside one of our `vendored` folders

napari/utils/colormaps/vendored/cm.py
Carreau: oh, yeah, I can revert that or redo just the part not in vendor.

Carreau: done.

napari/utils/event.py
```diff
@@ -585,8 +585,8 @@ def unblock(self, callback=None):
     def blocker(self, callback=None):
         """Return an EventBlocker to be used in 'with' statements
 
-           Notes
-           -----
+        Notes
+        -----
            For example, one could do::
```
jni, line 590: @Carreau I'm going to merge the PR but this looks like it missed this indentation, maybe because it was "dependent" on another fix?

jni: Or maybe this is the "more conservative with whitespace" version...? Incidentally, I've very much liked all of your whitespace fixes so far...

jni (approved):
Awesome, as before! I love this tool! ❤️

PR #1441, opened 2020-07-13
Handle nested zarr path

### LINKED ISSUE #1406: Support for Zarr files with multiple datasets (groups)
## 🚀 Feature

`napari` 0.3.5 does not open a Zarr folder if it contains several datasets (for example, saved with `dask.array.to_zarr(data, 'data.zarr', component='myimage')`, where 'myimage' is the name of the group). `napari` says

> 'Could not find a format to read the specified file in single-image mode'*.

If the argument `component` of `to_zarr` is omitted, the resulting image can be opened by `napari` with menu 'file->open folder'.

It is easy to specify the group to be opened as it is represented by a folder in the file system, inside of the .zarr directory.

## Motivation

During data analysis I often want to have several related datasets saved together in a single Zarr container. For example, I may want to do dark frame and flat field correction of an image stack, and then store the dark frame and the flat frame alongside with the image stack. This way I can have a pre-processed dataset that contains data used for pre-processing, so I can restore the uncorrected raw dataset, if necessary.

## Pitch

I want to open a group (e.g. named 'myimage') from a Zarr folder using command line
```
$ napari data.zarr/path/to/myimage
```
or using menu option 'file->open folder'.

## Alternatives

Currently I use this as a workaround:

```python
import zarr as z
import napari

g = z.open_group('data.zarr')
with napari.gui_qt(): 
    napari.view_image(g['myimage']) 
```

manzt:
# Description
Talking to @tlambert03 today, I thought this change would be pretty minimal to support more generic zarr IO in napari. It is a convention that the root directory of a zarr store ends with `.zarr`. However, arrays and groups can be nested at many levels deeper than the root. The previous `magic_imread` would only handle opening a root:

```bash
$ napari dataset.zarr # group or array
```

This PR keeps the previous behavior, but since all directories within a store are arrays or groups, the `magic_imread` now uses the same zarr logic on nested directory (group or array) which is requested in #1406. 

```bash
$ napari dataset.zarr/path/to/nested # group or array
```

## Type of change

# References
This addresses some aspects of #1406, but it does not change _how_ napari loads zarr arrays and groups.

# How has this been tested?
I've added a test for a nested array.

## Final checklist:

### DIFF
--- a/napari/utils/_tests/test_io.py
+++ b/napari/utils/_tests/test_io.py
@@ -111,6 +111,13 @@ def test_single_filename(single_tiff):
     assert images.shape == (2, 15, 10)
 
 
+def test_guess_zarr_path():
+    assert io.guess_zarr_path('dataset.zarr')
+    assert io.guess_zarr_path('dataset.zarr/some/long/path')
+    assert not io.guess_zarr_path('data.tif')
+    assert not io.guess_zarr_path('no_zarr_suffix/data.png')
+
+
 @pytest.mark.skipif(not zarr_available, reason='zarr not installed')
 def test_zarr():
     image = np.random.random((10, 20, 20))
--- a/napari/utils/_tests/test_io.py
+++ b/napari/utils/_tests/test_io.py
@@ -124,6 +124,19 @@ def test_zarr():
 
 
 @pytest.mark.skipif(not zarr_available, reason='zarr not installed')
+def test_zarr_nested(tmp_path):
+    image = np.random.random((10, 20, 20))
+    image_name = 'my_image'
+    root_path = tmp_path / 'dataset.zarr'
+    grp = zarr.open(str(root_path), mode='a')
+    grp.create_dataset(image_name, data=image)
+    image_in = io.magic_imread([str(root_path / image_name)])
+    # Note: due to lazy loading, the next line needs to happen within
+    # the context manager. Alternatively, we could convert to NumPy here.
+    np.testing.assert_array_equal(image, image_in)
+
+
+@pytest.mark.skipif(not zarr_available, reason='zarr not installed')
 def test_zarr_multiscale():
     multiscale = [
         np.random.random((20, 20)),
--- a/napari/utils/_tests/test_io.py
+++ b/napari/utils/_tests/test_io.py
@@ -137,9 +137,8 @@ def test_zarr_nested(tmp_path):
     root_path = tmp_path / 'dataset.zarr'
     grp = zarr.open(str(root_path), mode='a')
     grp.create_dataset(image_name, data=image)
+
     image_in = io.magic_imread([str(root_path / image_name)])
-    # Note: due to lazy loading, the next line needs to happen within
-    # the context manager. Alternatively, we could convert to NumPy here.
     np.testing.assert_array_equal(image, image_in)
 
 
--- a/napari/utils/io.py
+++ b/napari/utils/io.py
@@ -174,11 +174,8 @@ def magic_imread(filenames, *, use_dask=None, stack=True):
     # replace folders with their contents
     filenames_expanded = []
     for filename in filenames:
-        is_zarr_path = any(
-            part.endswith(".zarr") for part in Path(filename).parts
-        )
         # zarr files are folders, but should be read as 1 file
-        if os.path.isdir(filename) and not is_zarr_path:
+        if os.path.isdir(filename) and not guess_zarr_path(filename):
             dir_contents = sorted(
                 glob(os.path.join(filename, '*.*')), key=_alphanumeric_key
             )
@@ -202,10 +199,7 @@ def magic_imread(filenames, *, use_dask=None, stack=True):
     images = []
     shape = None
     for filename in filenames_expanded:
-        is_zarr_path = any(
-            part.endswith(".zarr") for part in Path(filename).parts
-        )
-        if is_zarr_path:
+        if guess_zarr_path(filename):
             image, zarr_shape = read_zarr_dataset(filename)
             if shape is None:
                 shape = zarr_shape
@@ -244,6 +238,22 @@ def magic_imread(filenames, *, use_dask=None, stack=True):
         else:
             image = images  # return a list
     return image
+
+
+def guess_zarr_path(path):
+    """Guess whether string path is for zarr hierarchy.
+
+    Parameters
+    ----------
+    path: str
+        Path to a file or directory.
+
+    Returns
+    -------
+    bool
+        Whether path is for zarr.
+    """
+    return any(part.endswith(".zarr") for part in Path(path).parts)
 
 
 def read_zarr_dataset(path):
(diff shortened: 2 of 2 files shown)


napari/utils/_tests/test_io.py
```diff
@@ -123,6 +123,19 @@ def test_zarr():
         np.testing.assert_array_equal(image, image_in)
 
 
+@pytest.mark.skipif(not zarr_available, reason='zarr not installed')
+def test_zarr_nested(tmp_path):
+    image = np.random.random((10, 20, 20))
+    image_name = 'my_image'
+    root_path = tmp_path / 'dataset.zarr'
```
sofroniewn, line 137: Does this syntax `tmp_path / 'dataset.zarr'` automatically do a string concatenation with `/` as a file separator? Is there a problem with `/` vs `\` if it is a file separator across platforms? (evidently not as tests run, but just asking to best understand - apologies for my ignorance here!) I'm guessing it's an "inside zarr thing" and so ok?

napari/utils/io.py
```diff
@@ -174,9 +174,11 @@ def magic_imread(filenames, *, use_dask=None, stack=True):
     # replace folders with their contents
     filenames_expanded = []
     for filename in filenames:
-        ext = os.path.splitext(filename)[-1]
+        is_zarr_path = any(
+            part.endswith(".zarr") for part in Path(filename).parts
+        )
```
sofroniewn: Can this be its own function called `guess_zarr_path` that takes a `filename` and returns a `bool`. This function can then get it's own tests too.

napari/utils/io.py
```diff
@@ -200,8 +202,10 @@ def magic_imread(filenames, *, use_dask=None, stack=True):
     images = []
     shape = None
     for filename in filenames_expanded:
-        ext = os.path.splitext(filename)[-1]
-        if ext == '.zarr':
+        is_zarr_path = any(
+            part.endswith(".zarr") for part in Path(filename).parts
+        )
+        if is_zarr_path:
```
sofroniewn: Here you could re-use `guess_zarr_path`

sofroniewn (commented):
Looking good, left a question and a comment about putting some functionality inside its own function that can then be tested

napari/utils/_tests/test_io.py
tlambert03: That’s a (very cool) feature of pathlib.Path objects (which the tmp_path is): https://stackoverflow.com/questions/53083963/python-pathlib-operator-how-does-it-do-it
It’s not a string, but rather an operation, and it works across platforms like os.path.join

manzt: Yes my understanding is that [`pathlib.Path`](https://docs.python.org/3/library/pathlib.html) implements `__truediv__` and deals with both separators. Zarr (if I remember correctly) cannot take a `pathlib.Path` object and requires a string, which is why I cast down when creating/reading the array.

Somewhat of a side note, but related, I noticed the `magic_read` docstring says `filenames` is `pathlib.Path | [pathlib.Path]` but `filenames` is immediately cast to a string. The function then operates on these strings with `os.path`. I was tempted to remove the top-level casting to strings and use the `pathlib.Path`s (as I assume may have been intended?) to to glob files and build paths, only converting to strings when calling thirdparty readers. However, that was beyond the scope of this PR.

For instance, the current implementation for `guess_zarr_path` is somewhat of an antipattern (if I'm using that word correctly). It requires the recreation a `pathlib.Path` to check the parts, but this wouldn't be necessary if the `filename` was still a `pathlib.Path`.

```python
def guess_zarr_path(filename):
    return any(
        part.endswith(".zarr") for part in Path(filename).parts
    )
```

jni: @manzt the reason for that is that there's various parts of the ecosystem that historically haven't played nice with pathlib Paths, see e.g. https://github.com/scikit-image/scikit-image/issues/4138 or https://github.com/scikit-image/scikit-image/issues/3965. I haven't been keeping track of progress community-wide, but basically it felt safer to "contain" the pathlib elements and use strings preferentially except where pathlib is handy.

Perhaps the situation has changed now and we should take the leap and embrace them, but to be honest I've regretted it each time I've taken that leap! 😂

napari/utils/_tests/test_io.py
```diff
@@ -123,6 +130,19 @@ def test_zarr():
...
+    grp = zarr.open(str(root_path), mode='a')
+    grp.create_dataset(image_name, data=image)
+    image_in = io.magic_imread([str(root_path / image_name)])
+    # Note: due to lazy loading, the next line needs to happen within
+    # the context manager. Alternatively, we could convert to NumPy here.
```
jni: "within the context manager"?

napari/utils/io.py
```diff
@@ -242,6 +240,22 @@ def magic_imread(filenames, *, use_dask=None, stack=True):
     return image
 
 
+def guess_zarr_path(path):
+    """Guess whether string path is for zarr hierarchy.
```
jni: for -> part of a?

jni (approved):
@manzt Awesome, thanks! I've left a couple of minor comments but only in comments/docstrings, code-wise this is ready to go in imho!

napari/utils/_tests/test_io.py
```diff
@@ -111,6 +111,13 @@ def test_single_filename(single_tiff):
     assert images.shape == (2, 15, 10)
 
 
+def test_guess_zarr_path():
+    assert io.guess_zarr_path('dataset.zarr')
+    assert io.guess_zarr_path('dataset.zarr/some/long/path')
+    assert not io.guess_zarr_path('data.tif')
+    assert not io.guess_zarr_path('no_zarr_suffix/data.png')
```
jni, line 118: post review comment: it would be good to test the behaviour on `'path/to/image.zarr/component.png'`. That will return True currently and I think that's the correct behaviour, but one would have to think about it, it would be great to have it be "documented". Maybe this test belongs in a doctest in the docstring for the function.

napari/utils/_tests/test_io.py
manzt: Makes sense, thanks for the clarification! To be clear, I wasn't suggesting trying to interact with other parts of ecosystem with pathlib objects, but instead, rather than casting to strings immediately and then trying to glob files or build paths (with `os.path` + strings), just work with pathlib _until_ all desired paths are created.

Essentially wrangle all files that need to be opened with pathlib, and _then_ cast those paths to strings when calling `skimage.imread` etc.

napari/utils/_tests/test_io.py
manzt: Ahh sorry, copied from the test above.. removing..

napari/utils/_tests/test_io.py
manzt: I added a doctest, but left the unit test with more edge cases that hopefully aren't up for debate over what `guess_zarr_path` should return. For the case you mentioned, I think it's the correct behavior as well, since having `.zarr` _somewhere_ in the path should indicate directories below are groups or arrays (regardless of the component name).

sofroniewn (approved):
LGTM! Thanks @manzt - great work!!

PR #1444, opened 2020-07-13
Evented list

### LINKED ISSUE #312: fix TypedList error throwing and add comprehensive test suite

tlambert03:
# Description
This PR introduces a new evented list model.  This addresses some of the issues in https://github.com/napari/napari/issues/1380, by making sure that all mutation operations on the list go through a few key methods (`__setitem__`, `__delitem__`, `insert`).  All other operations (like `remove`, `append`, `clear` go through these methods.  It's also a proper Generic type so you can use it in type annotations: 

```python
a: EventedList[int] = EventedList(['a'])  # type error
```
This also provides a `NestableEventedList` that emits recursive events when any of its children are altered (this is going to be key for connecting up events in the `LayerGroup`).  It also defines a Protocol `SupportsEvents` (any object that has an attribute called `events` that is an instance of an `EmitterGroup`).  All events emitted by nested children will bubble up to the root event, with an "index" attribute on the event that holds the index of the original emitter.  For instance:

```python
In [1]: from napari.utils.events import NestableEventedList

In [2]: from napari.layers import Points

In [3]: root = NestableEventedList([0, [[], 1]])  # all nested mutable lists will become "eventful"

In [4]: root.events.connect(lambda e: print(e.index, e.type))  # let's watch the events

In [5]: p = Points()

In [6]: root[1, 0].append(p)
(1, 0, 0) inserting
(1, 0, 0) inserted

In [7]: p.opacity = 0.6 # any `SupportsEvents` interface will bubble up the event
(1, 0, 0) thumbnail
(1, 0, 0) status
(1, 0, 0) opacity

In [8]: root.pop()
1 removing
1 removed
Out[8]: [[<Points layer 'Points' at 0x7fbe081f9310>], 1]

In [9]: root.clear()
0 removing
0 removed
```

there's still one broken test (a new test that should work, but needs tweaking)

### in future PRs

this isn't actually hooked up to anything at the moment.  but it forms a solid basis for eventful containers.  (I have an evented dict ready to go as well.  The idea would be to replace the current list model with this, but that will require porting a couple other niceties from that class first.

## Type of change

# How has this been tested?

- [ ] new tests make sure the list complies with the standard `list` API, and confirms events.  Still one broken test.

## Final checklist:

closes #312


napari/utils/events/_tests/test_evented_list.py
```diff
@@ -0,0 +1,249 @@
...
+
+@pytest.fixture(params=[EventedList, NestableEventedList])
+def test_list(request, regular_list):
+    test_list = request.param(regular_list)
+    test_list._events = []
+    test_list.events.connect(test_list._events.append)
+    return test_list
+
+
+@pytest.mark.parametrize(
```
sofroniewn, line 21: When I did the list model refactoring stuff in #1391 I added a test for event emission using `Mock` - see here https://github.com/sofroniewn/napari/blob/87910700153881903401c4f1d3f58285c94a674b/napari/utils/list/_tests/test_list_model.py

It seems like you've taken a different approach (which is totally fine), but can you explain if it tests the same stuff/ why you prefer this approach over `mock` - it would be a good way to standardize testing event emission etc in the repo too!!

sofroniewn (commented):
Very exciting! Looks great on a quick glance. I asked one question about the tests. Let me know when it's ready for review to be played with. I can't wait to refactor dims!!

napari/utils/events/_tests/test_evented_list.py
tlambert03: oh actually, I think I love what you did there!  but maybe we can do it slightly more generically such that we don't have to know names of the methods.  let me see if I can refactor tests here to use something like that

sofroniewn: Ok great, refactor and improve away!

tlambert03: ok, used Mocks where possible.  I decided against creating a full fixture as you did there, instead using the [`wraps`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock) argument to Mock.  `test_list.events = Mock(wraps=test_list.events)`... That turns `test_list.events` into a pass-through object that you can spy on, and make assertions about what events were emitted, without explicitly saying what the event names should be.

(If you're wondering why I left that last test without a Mock, it's an idiosyncrasy/difference in the way events are emitted from nested children that makes it easier to test that way)

tlambert03: (still one broken test for moving multiple items simultaneously)

tlambert03:
ok, everything is working here now, and this is ready for review.

I added `typing_extensions`  to our dependencies, to support `Protocol`... which we've discussed using elsewhere anyway.  I believe that needs to be there for anything <3.8.

also just want to make a quick note that I was able to dramatically simplify the logic for moving multiple items, but for the moment the `NestableEventedList.move_multiple` method is _not_ emitting any move events.  Ultimately, that will probably only be used for drag/drop events with the layer Tree, and I'll update it when that PR is ready.

sofroniewn (approved):
I've read through and have nothing to add beyond that it looks great. Given that this isn't currently used in the code base I think it's ok to let it merge and then revisit as we begin to use it.

I will note that I'm a little apprehensive about the additional complexity of the nested list vs just the normal list, but I see why that will be important for the layer groups. I'm curious though if you can anticipate us needing it for anything else? 

Given that nested list is mostly done and tested here I think it's fine to merge as is, but I'd prefer if we started using the new evented list in the current layer list model and salvage the QListWidget approach from #1391 before adding layer groups - which is a major new feature. I think it will be easier to do this piecemeal. I'd also like to start trying to use the evented list in the dims model before we go too far.

Regardless, @jni should give this a review before merge!!

tlambert03:
I would encourage people not to use the nested list unless they actually need a nested list that bubbles events up to the root (Which, I really do see as the key to getting layer groups to work).  It is a complex class (recursion always is), but it makes *using* the class much easier (since you only have to connect/listen to the top

I’d say use it (EventedList) in the dims model.  But, as you know, my opinion is that not enough of the layer widget will be reused for LayerGroups and TreeViews.  Even the item widget itself will have to turn into a QStyledItemDelegate.

sofroniewn (commented):
Playing with this in Dims right now and running into no ability to do `+` with the lists

```python-traceback
self._range = EventedList([(0, 2, 1)] * (ndim - cur_ndim)) + self._range
TypeError: unsupported operand type(s) for +: 'EventedList' and 'EventedList'
```

or 

```python-traceback
self._range = [(0, 2, 1)] * (ndim - cur_ndim) + self._range
TypeError: can only concatenate list (not "EventedList") to list
```

Definitely possible to get around this, but the `+` syntax is nice. Curious how you might think about it

tlambert03:
sure, I like that syntax too.  I can add  `__add__` /`__iadd__` methods

tlambert03:
> running into no ability to do + with the lists

I added this (and merged master) go ahead and merge this into your branch and give it a try.
Also, an important note.  These two forms will return an `EventedList`
```python
a = EventedList([1])
b = [2]
a += b  # a is modified in place and becomes [1, 2]
c = a + b  # a is not modified, c becomes an evented list [1, 2, 2]

# regular list is on the left, this does NOT return an evented list 
c = b + a  #  c becomes [2, 1 ,2] but is NOT an EventedList
```

we can probably force all "sums" to be evented by changing  the `__radd__` method... lemme know if you want that too

sofroniewn:
Thanks!! I've given using this ago in #1486 - looks like it's best to probably not create/ destroy the evented list though and rather modify in place (otherwise I'm worried I might break connections?), in which case I probably wont use this too much, though the `a += b` syntax is quite nice for in place modifications. Unfortunately in #1486 I want to pre-pend to the list, but maybe there is a better way of making the change I want (see the `order` and `axis_labels` properties in the `ndim` setter).

sofroniewn:
I'm starting to explore using the EventedList for the layer list. One thing that's not clear to me is how to combine the EventedList with the type checking/ custom lookup functionality provided by our current list model. Any pointers in the right direction here would be appreciated @tlambert03!!

tlambert03:
> I'm starting to explore using the `EventedList` for the layer list. One thing that's not clear to me is how to combine the `EventedList` with the type checking/ custom lookup functionality provided by our current list model. Any pointers in the right direction here would be appreciated @tlambert03!!

Just updated this PR to provide that functionality... you can import it `from napari.utils.events import TypedEventedList`.  It works pretty much the same as before, except the lookup function signature has one less argument: e.g. `lookup = {str: lambda x: x.name}` instead of `lookup={str: lambda q, e: q == e.name}`

tlambert03:
@jni will like this one:  I added IPython tab autocompletion when using custom "name lookups" with a TypedList:

```python
from napari.layers import Points, Layer
from napari.utils.events.containers import TypedList
layer_list = TypedList(
    [Points(name="supercalifragilisticexpialidocious")],
    basetype=Layer,
    lookup={str: lambda x: x.name}
)

layer = layer_list['sup  # then hit tab!

**Comment by sofroniewn** [comment]
Thanks @tlambert03 this is looking great - it was very easy to drop the new `TypedList` and `TypedEventedList` in for `LayerList` and `TransformChain` in #1504, but I did encounter one problem with the new stuff

```python
# Rearrange layers by tuple
layers[:] = layers[(1, 0, 2)]
```

```python-traceback
napari/components/_tests/test_layers_list.py:105:
napari/utils/events/containers/_typed_list.py:68: in __getitem__
    return super().__getitem__(key)
    def __getitem__(self, key):  # noqa: F811 (redefinition)
>       result = self._list[key]
E       TypeError: list indices must be integers or slices, not tuple
```

tlambert03:
awesome,  I asked over in https://github.com/napari/napari/pull/1504#issuecomment-663876341, but the discussion is probably better off here.  How do you feel about that `layers[:] = layers[(1, 0, 2)]` syntax vs `layers.sort((1,0,2))`.  I favor something like the latter, since it's a more numpy like syntax, makes clearer that we are not mutating the original object, and because we'll also need tuple-based indexing for nested lists more than we need it for sorting.

sofroniewn:
I guess neither `layers[(1, 0, 2)]`  nor `layers.sort((1,0,2))` are standard syntax provided on normal `list` objects are they? In which case I wouldn't want to add them to our list object just for convenience unless we really need to as it starts to become confusing.

I agree `layers[(1, 0, 2)]` could be really confusing if we start wanting to use syntax like that for nested lists!!

What is the way to accomplish a reordering like this with a normal list? I'm pretty open to changing syntax here. Curious what @jni thinks too.

tlambert03:
>  I guess neither layers[(1, 0, 2)] nor layers.sort((1,0,2)) are standard syntax provided on normal list objects are they?

well, actually... `sort` _is_ a method on builtin list, but it doesn't do what this does.  I guess we want something a bit more like `reorder`.

> What is the way to accomplish a reordering like this with a normal list?

as far as I know, there is no builtin method to reorder a list in-place.  So what I'm mostly seeing people do is `mylist[:] = [mylist[i] for i in myorder]`. That's already pretty straightforward, so perhaps we just:

```python
# drop this syntax
layers[:] = layers[(1, 0, 2)]
# in favor of 
layers[:] = [layers[i] for i in (1, 0, 2)]
```

_edit_: that should also work with `TypedList`here if you did
```python
layers[:] = [layers[name] for name in ('Image', 'Points', 'Points [1]')]
```

tlambert03:
ok, I made a bunch of changes that generally improve handling of slice indices ... but the last thing I'm running into when using this for the layerlist is issues around layer selection.  That `_add` method that listens to the layerlist's own added event should probably not have been connected as an event, but rather as a direct method.  and the way it unselects everything but the last added layer is giving me problems when using `move_selected` which also has a lot of conditional logic to manipulate the selection status of layers ... this is how these PRs end up getting big 😂

napari/utils/events/containers/_list.py
```diff
@@ -205,17 +206,69 @@ def move(self, cur_index: int, new_index: int) -> bool:
...
+                to_move.extend(list(range(*idx.indices(len(self)))))
+            elif isinstance(idx, int):
+                to_move.append(idx)
+            else:
+                raise TypeError("Can only move integer or slice indices")
+
+        dest_index -= len([i for i in to_move if i < dest_index])
+
+        self.events.moving(index=to_move, new_index=dest_index)
+        with self.events.blocker():
```
sofroniewn: I'm still getting problems here with the blocker not blocking as expected - see this comment https://github.com/sofroniewn/napari/pull/6#discussion_r460472336 and one immediately above it. Does that reproduce for you? I think we might need something more like the `block_all` call https://github.com/napari/napari/blob/master/napari/utils/event.py#L766-L771, or modify this context.

tlambert03: will look into it.  it's possible I'm misunderstanding the `self.events.blocker()` and that we may need 
```python
with self.events.removed.blocker():
    with self.events.inserted.blocker():
        ...
```
but I want to look through the event code again

sofroniewn: I found I needed to add something like https://github.com/napari/napari/pull/1504/files#diff-b3f8eaee1d3d919bf96270862a2cebefR840-R883, but I encourage you to take a look too as I maybe missing something also.

tlambert03: yes, you're absolutely right.  I was misunderstanding `events.blocker()`,  `block_all` is definitely what we needed instead, and your `blocker_all` context manager is great.  (or blocking the specific events as above)

tlambert03: pulled in the `blocker_all()` context manager and added a test.

tlambert03:
ok, I've tidied this up, and it's ready for review.  we've been using it in a number of other experimental branches and it's doing well.  and the test coverage is also pretty good.  This will ultimately replace the current list model, but that will be in a later PR.  So at the moment, merging this doesn't actually modify any actively used code in napari, it just makes it easier to work with in other PRs.

kne42 (approved):
GJ and very impressive work! LGTM!

sofroniewn (approved):
This looks good to me @tlambert03, I had another look through the code now and have no additional comments (I already gave the code some review during the development of this PR and everything has been addressed).

It might be nice if @jni could have a look too, but we have quite a back log of PRs to merge so we can also merge this one as is and make any changes needed in followups as we start to integrate more with it. I'll let it sit a little longer though

napari/utils/events/_tests/test_evented_list.py
```diff
@@ -0,0 +1,342 @@
...
+
+from napari.utils.events import EmitterGroup, EventedList, NestableEventedList
+
+
+@pytest.fixture
+def regular_list():
+    return list(range(5))
+
+
+def flatten(_lst):
```
jni: I don't quite understand what this does, can you add a docstring @tlambert03? Is it like `itertools.chain` but with infinite nesting? And why does the argument name have a leading underscore?

napari/utils/events/_tests/test_evented_list.py
```diff
@@ -0,0 +1,342 @@
...
+        ('pop', (-2,), ('removing', 'removed')),
+        ('remove', (3,), ('removing', 'removed')),
+        ('reverse', (), ('reordered',)),
+        ('__add__', ([7, 8, 9],), ()),
+        ('__iadd__', ([7, 9],), ('inserting', 'inserted') * 2),
+        ('__radd__', ([7, 9],), ('inserting', 'inserted') * 2),
+        # sort?
+    ],
+    ids=lambda x: x[0],
+)
```
jni, line 51: this is super nice. =)

napari/utils/events/_tests/test_evented_list.py
```diff
@@ -0,0 +1,342 @@
...
+    assert isinstance(test_slice, test_list.__class__)
+
+
+NEST = [0, [10, [110, [1110, 1111, 1112], 112], 12], 2]
+
+
+def test_nested_indexing():
+    """test that we can index a nested list with nl[1, 2, 3] syntax."""
+    ne_list = NestableEventedList(NEST)
+    indices = [tuple(int(x) for x in str(n)) for n in flatten(NEST)]
```
jni, line 225:
```suggestion
    # 110 -> '110' -> (1, 1, 0)
    indices = [tuple(int(x) for x in str(n)) for n in flatten(NEST)]
```

it took me a bit to grok this expression!

napari/utils/events/_tests/test_evented_list.py
```diff
@@ -0,0 +1,342 @@
...
+def test_nested_move_multiple(source, dest, expectation):
+    """Test that moving multiple indices works and emits right events."""
+    ne_list = NestableEventedList([0, 1, [20, [210, 211], 22], 3, 4])
+    ne_list.events = Mock(wraps=ne_list.events)
+    ne_list.move_multiple(source, dest)
+    ne_list.events.reordered.assert_called_with(value=expectation)
+
+
+def test_arbitrary_child_events():
+    """Test that any object that supports the events protocol bubbles events.
```
jni, line 311: Could you explain this a bit more in the docstring or in comments? I don't actually understand what is supposed to happen or what is being tested. Is it that if an object `e_obj` is in the list, and something happens to it, its event will be emitted by the root of the list?

napari/utils/events/_tests/test_typed_list.py
```diff
@@ -0,0 +1,158 @@
...
+    NestableEventedList,
+    TypedMutableSequence,
+)
+
+
+@pytest.fixture(
+    params=[TypedMutableSequence, EventedList, NestableEventedList]
+)
+def list_type(request):
+    return request.param
```
jni, line 17: I actually haven't seen this type of fixture before. Does this mean that a test taking this fixture will be run as many times as the fixture has parameters, one for each param?

jni: And if so can you write as much in a comment? =)

napari/utils/events/_tests/test_typed_list.py
```diff
@@ -0,0 +1,158 @@
...
+    assert a[1].name == 'hi'
+    assert a.index("hi") == 1
+
+    # index with string also works
+    assert a['hi'] == hi
+
+    # index with tuple will use the `tuple` type lookup
+    assert a[{'some': 'data'}].data == {'some': 'data'}
+    assert a.index({'some': 'data'}) == 3
+    assert a[{'some': 'data'}] == dct
```
jni: The comment doesn't seem to match the code?

Also, does this API mean:

- a[int] is regular indexing
- a[tuple[int]] is nested indexing
- a.index(int) is custom lookup based on int/tuple, if provided?

napari/utils/events/_tests/test_typed_list.py
```diff
@@ -0,0 +1,158 @@
...
+
+    # also works during instantiation
+    with pytest.raises(TypeError):
+        _ = NestableEventedList([1, 1, ['string']], basetype=int)
+
+    with pytest.raises(TypeError):
+        _ = NestableEventedList([1, 2, [3, ['string']]], basetype=int)
+
+
+@pytest.mark.xfail(reason="Need to enable custom indexing on nestable lists")
```
jni, line 139: Meaning, not implemented yet?

napari/utils/events/containers/_evented_list.py
```diff
@@ -0,0 +1,245 @@
...
+
+- ``insert`` = "create" : add a new item/index to the list
+- ``__getitem__`` = "read" : get the value of an existing index
+- ``__setitem__`` = "update" : update the value of an existing index
+- ``__delitem__`` = "delete" : remove an existing index from the list
+
+All of the additional list-like methods are provided by the MutableSequence
+interface, and call one of those 4 methods.  So if you override a method, you
+MUST make sure that all the appropriate events are emitted.  (Tests should
+cover this in test_evented_list.py)
```
jni, line 22: ❤️ this comment

napari/utils/events/containers/_evented_list.py
```diff
@@ -0,0 +1,245 @@
...
+
+        # For inheritance: If the mro already provides an EmitterGroup, add...
+        if hasattr(self, 'events') and isinstance(self.events, EmitterGroup):
+            self.events.add(**_events)
+        else:
+            # otherwise create a new one
+            self.events = EmitterGroup(source=self, **_events)
+        super().__init__(data, basetype=basetype, lookup=lookup)
+
+    # WAIT!! ... Read the module docstring before reimplement these classes
```
jni:
```suggestion
    # WAIT!! ... Read the module docstring before reimplement these methods
```

napari/utils/events/containers/_nested_list.py
```diff
@@ -0,0 +1,398 @@
...
+        return (index,)  # single integer inserts to self
+    elif isinstance(index, tuple):
+        return index
+    raise TypeError(f"Invalid nested index: {index}. Must be an int or tuple")
+
+
+def split_nested_index(
+    index: Union[NestedIndex, Index]
+) -> Tuple[ParentIndex, Index]:
+    """ Given a nested index, return (nested_parent_index, row).
```
jni:
```suggestion
    """Given a nested index, return (nested_parent_index, row).
```

😜

napari/utils/events/containers/_nested_list.py
```diff
@@ -0,0 +1,398 @@
...
+    # WAIT!! ... Read the ._list module docs before reimplement these classes
+    # def append(self, item): ...
+    # def clear(self): ...
+    # def pop(self, index=-1): ...
+    # def extend(self, value: Iterable[_T]): ...
+    # def remove(self, value: T): ...
+
+    @overload
+    def __getitem__(self, key: int) -> _T:
+        ...  # pragma: no cover
```
jni: Doesn't an int return *either* a `_T` or a `NestableEventedList[_T]`?

This is all super cool btw — I'd seen `overload` but never IRL. Since the implementation is unchanged, I guess this is all only to map input types to output types? Probably worth mentioning in a comment right above. Actually I don't really understand why the implementation falls back to the default one...

napari/utils/events/containers/_nested_list.py
```diff
@@ -0,0 +1,398 @@
...
+            return [(cast(NestableEventedList[_T], self[parent_i]), index)]
+        return super()._delitem_indices(key)
+
+    def __delitem__(self, key):
+        # delete from the end
+        for parent, index in self._delitem_indices(key):
+            self._disconnect_child_emitters(parent[index])
+        super().__delitem__(key)
+
+    # TODO: implement __eq__
```
jni:
```suggestion
```

Can we replace this comment by an issue? =)

napari/utils/events/event.py
```diff
@@ -828,6 +837,18 @@ def ignore_callback_errors(self, ignore):
...
+        """Return an EventBlockerAll to be used in 'with' statements
+
+        Notes
+        -----
+        For example, one could do::
+
+            with emitter.blocker_all():
+                pass  # ..do stuff; no events will be emitted..
+        """
+        return EventBlockerAll(self)
```
jni, line 850:
```suggestion
    @contextmanager
    def blocker_all(self):
        """Return an EventBlockerAll to be used in 'with' statements

        Notes
        -----
        For example, one could do::

            with emitter.blocker_all():
                pass  # ..do stuff; no events will be emitted..
        """
        self.block_all()
        yield
        self.unblock_all()
```
?

jni (approved):
@tlambert03 A tour de force! 😊 I've made some minor requests for clarification, and a couple of tiny implementation suggestions. Otherwise this should be good to go as far as I'm concerned!

napari/utils/events/_tests/test_evented_list.py
tlambert03: sorry that was opaque. yes it flattens arbitrarily nested lists:
```python
a = [1, [2, [3], 4], 5]
list(flatten(a)) == [1, 2, 3, 4, 5]
```
I changed it to a much clearer generator, and added docstring

napari/utils/events/_tests/test_evented_list.py
tlambert03: that explanation is in the [docstring of `NestableEventedList` ](https://github.com/napari/napari/pull/1444/files#diff-e68dd71c52551aecc697df17b73e30cfR87-R95) itself, but I will add a note and inline comments to the test as well.  The point is: if you add an object that implements the `SupportsEvents` Protocol (i.e. has an attribute `events` that is an `EmitterGroup`), then the parent container will re-emit those events (and this works recursively to the root container), and the index/indices of each child(ren) that bubbled the event will be added to the event.  This is going to be super-important for events in a future `LayerTree`.

napari/utils/events/_tests/test_typed_list.py
tlambert03: yep, that's what it does.  It's called a "[parametrized fixture](https://docs.pytest.org/en/stable/fixture.html#parametrizing-fixtures)", and it's super useful for growing multi-parameter test matrices.  Will add comment and links to docs

napari/utils/events/_tests/test_typed_list.py
tlambert03: ah damn it... good catch.  I used a tuple at first, but yet... then I switched to a `dict` as an example (because `tuple` is going to be used for nested indexing... so it was confusing).  changed the comment to `index with a dict will use the `dict` type lookup`

> Also, does this API mean...

more generally: `int`/`slice` work as usual, `tuple` will be used for nested indexing... but the main point here is that one can change the behavior of `__getitem__` and `index` by providing a `lookup` (mapping from type to callable that returns a value) when instantiating the list, like `lookup={str: lambda x: x.name, dict: lambda x: x.data},`.  Then whenever you get an item from the list, if `type(key) in lookup`, it uses the index of the returned value instead.

napari/utils/events/_tests/test_typed_list.py
tlambert03: yeah.  `xfail` means this test should ideally work but is "known to fail" (test-driven development before implementation).  I'd like to leave that for a later PR, since it's it's not immediately obvious to me that we want to recurse the full tree in that case.   will open an issue.

napari/utils/events/containers/_nested_list.py
tlambert03: you're absolutely right.  I think I left it out because mypy yelled at me for it.  here's the error if you're curious:

Screen Shot 2020-08-15 at 6 35 41 PM

however, I think in this case, it's probably better to give the proper return type and ignore the mypy error.

> I guess this is all only to map input types to output types?

Yes exactly. `@overload` is amazing and it really works wonders for catching errors.

> Actually I don't really understand why the implementation falls back to the default one...

you mean at runtime?  I think it's just because it's the last one declared right?

napari/utils/events/containers/_nested_list.py
tlambert03: will delete for now.  I'm actually not sure we _need_ to implement that.  It will become obvious if we do :)

napari/utils/events/event.py
tlambert03: yeah, I prefer that too.   @sofroniewn added that and I think he was just trying to stick with the patterns elsewhere in that module for consistency sake.  Do you want to stay with the way vispy does it?  or go our own way with it?

tlambert03 (commented):
thanks so much for the review @jni.  I think everything was pretty easily addressed here.

napari/utils/events/event.py
sofroniewn: i was just copying vispy for consistency, i don't have a strong feeling on this one. i'd say we can revisit in the future if we want to change as I don't want to hold up the PR for this, but absolutely fine to revisit

tlambert03: i think it's better to be consistent, and I don't think I want to go chopping up their eventblocker just yet.  so let's leave it

sofroniewn:
Thanks for the review @jni and the quick turn around on the changes @tlambert03, I think this is good to merge now!

As we start using in other PRs we can also make updates (including to the vispy event stuff), but it will be great to have this in and be able to press on with the evented data refactor!!

tlambert03:
> As we start using in other PRs we can also make updates (including to the vispy event stuff), but it will be great to have this in and be able to press on with the evented data refactor!!

agreed!  will work on polishing up the dataclass next.  (after 0.3.7 is of course just fine)

PR #1445, opened 2020-07-13
Toggle image/ volume nodes on ndisplay change

sofroniewn:
# Description
This PR supersedes #1404 to hopefully fix the same windows bugs it was fixing by no longer creating/ destroying vispy visuals just toggling between them when ndisplay changes.

## Type of change

# References
- closes #1404 

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/_vispy/vispy_image_layer.py
+++ b/napari/_vispy/vispy_image_layer.py
@@ -18,8 +18,9 @@ texture_dtypes = [
 
 class VispyImageLayer(VispyBaseLayer):
     def __init__(self, layer):
-        node = ImageNode(None, method='auto')
-        super().__init__(layer, node)
+        self._image_node = ImageNode(None, method='auto')
+        self._volume_node = VolumeNode(np.zeros((1, 1, 1)), clim=[0, 1])
+        super().__init__(layer, self._image_node)
 
         self.layer.events.rendering.connect(self._on_rendering_change)
         self.layer.events.interpolation.connect(self._on_interpolation_change)
@@ -39,11 +40,13 @@ class VispyImageLayer(VispyBaseLayer):
         self.node.parent = None
 
         if self.layer.dims.ndisplay == 2:
-            self.node = ImageNode(data, method='auto')
+            self.node = self._image_node
+            self.node.set_data(data)
         else:
             if data is None:
                 data = np.zeros((1, 1, 1))
-            self.node = VolumeNode(data, clim=self.layer.contrast_limits)
+            self.node = self._volume_node
+            self.node.set_data(data, clim=self.layer.contrast_limits)
 
         self.node.parent = parent
         self.reset()
--- a/napari/_vispy/vispy_image_layer.py
+++ b/napari/_vispy/vispy_image_layer.py
@@ -40,13 +40,13 @@ class VispyImageLayer(VispyBaseLayer):
         self.node.parent = None
 
         if self.layer.dims.ndisplay == 2:
+            self._image_node.set_data(data)
             self.node = self._image_node
-            self.node.set_data(data)
         else:
             if data is None:
                 data = np.zeros((1, 1, 1))
+            self._volume_node.set_data(data, clim=self.layer.contrast_limits)
             self.node = self._volume_node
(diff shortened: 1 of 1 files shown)


sofroniewn:
Ok, so disappointingly @tlambert03 @jni @ziyangczi this much more minor approach to #1404 has not succeeded (windows only test still fails) - it seems like attaching/ detaching the parent is causing the problem. Note that here I havn't taken the full on compound visual approach. I'll see if there's more simple things I can do. Very open to suggestion though.

liu-ziyang:
I am trying something along the line, have not succeeded myself either, (so it is not just my problem!)
I think we should go with the review of the bigger change

liu-ziyang:
This code actually works for me, glad I did try it. I think I overly complicated the issue, and went too fancy. Maybe we should just rerun the windows test, I hope that is a different problem

sofroniewn:
> This code actually works for me, glad I did try it. I think I overly complicated the issue, and went too fancy. Maybe we should just rerun the windows test, I hope that is a different problem

Crazy, ok I'm rerunning the windows test - but it could easily be something else.

Maybe @VolkerH can jump in to test if he can now render the volume from #1347 and @Czaki can give this a try on his data too (not with the change the scaling should working too).

Would be good to know if this is at least stepping us in the right direction. This is a pretty minor change that seems strictly better, so I think we might as well merge it regardless.

sofroniewn:
@ziyangczi what happens when you run windows CI locally before/ after this PR? Does it always pass locally? If so it would be great if you could adapt your current example into a test that fails windows CI locally before this PR but that passes afterwards.

liu-ziyang:
sure I can try adding some test for this issue, I was using nd_label example to manually test it. 
regarding the CI issue, I have not succeeded in running `pytest` locally either with or without these changes

Czaki:
@sofroniewn 
It works (spacing also).

VolkerH:
Hi @sofroniewn 
I installed the branch into the problematic environment and could successfully render the volume that previously crashed.

napari/_vispy/vispy_image_layer.py
```diff
@@ -39,11 +40,13 @@ def _on_display_change(self, data=None):
...
         if self.layer.dims.ndisplay == 2:
-            self.node = ImageNode(data, method='auto')
+            self.node = self._image_node
+            self.node.set_data(data)
         else:
             if data is None:
                 data = np.zeros((1, 1, 1))
-            self.node = VolumeNode(data, clim=self.layer.contrast_limits)
+            self.node = self._volume_node
+            self.node.set_data(data, clim=self.layer.contrast_limits)
```
jni: @sofroniewn the only thing I would change here is to *first* set the data on the appropriate node, and *then* swap out the node. ie:

```python
self._image_node.set_data(data)
self.node = self._image_node
...
self._volume_node.set_data(data)
self.node = self._volume_node
```

To me that seems cleaner in that you never enter an inconsistent state, in some sense...

Other than that, since this seems to fix the issue for everyone, I suggest merging now and continue investigating in future PRs!

sofroniewn: done

sofroniewn:
@jni I made your fix. Can you approve - then we can merge without test, and @ziyangczi can continue to work on a test that would at least fail locally.

We thought we might be getting close with something like the following, but it still passed, curious if anyone else has any ideas!!
```python
def test1(make_test_viewer, qtbot):
    viewer = make_test_viewer(show=True)
    blobs = data.binary_blobs(length=128, volume_fraction=0.1, n_dim=3)
    viewer.add_image(blobs.astype(float), name='blobs')
    viewer.add_labels(blobs, name='blob ID')
    viewer.window.qt_viewer.canvas.on_draw(None)
    viewer.dims.ndisplay = 3
    viewer.window.qt_viewer.view.camera.scale_factor = 100
    viewer.window.qt_viewer.canvas.on_draw(None)
    qtbot.mouseClick(viewer.window.qt_viewer.canvas.native, QtCore.Qt.LeftButton, pos=QtCore.QPoint(100, 100))
```

sofroniewn:
Closes #1347

jni (approved):

PR #1451, opened 2020-07-14
DOC: try to apply black to example inside docstrings.

Carreau:
This might be a bit controversial to apply on all docstrings' examples,
so, so far I ran it only a chosen file, and have to think about how to
make docstring on which to not apply black; this is mostly problematic
on example for two things:

 - One the code width black is ~100 ish, but is not aware those
   docstring are themselves indented by 4 to 8 spaces, +4 because of `>>>`
   prompts.
 - It does reformat carefully bayed out arrays.

Note that I also manually fixed one example to prepend continuations
prompt with `... ` instead of 4 spaces.

---

Your feedback on whether black in docstring is desirable (or not) and whether it should be configurable on a per file/function basis and how is appreciated.

### DIFF
--- a/napari/components/add_layers_mixin.py
+++ b/napari/components/add_layers_mixin.py
@@ -757,7 +757,7 @@ class AddLayersMixin:
         edge_width : float
             Width for all vectors in pixels.
         length : float
-             Multiplicative factor on projections for length of all vectors.
+            Multiplicative factor on projections for length of all vectors.
         edge_color : str
             Color of all of the vectors.
         edge_color_cycle : np.ndarray, list
@@ -982,8 +982,8 @@ class AddLayersMixin:
         >>> viewer = napari.Viewer()
         >>> data = (
         ...     np.random.random((10, 2)) * 20,
-        ...     {'face_color': 'blue'},
-        ...     'points',
+        ...     {"face_color": "blue"},
+        ...     "points",
         ... )
         >>> viewer._add_layer_from_data(*data)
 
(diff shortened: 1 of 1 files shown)


jni:
@Carreau Honestly I hate this practice and the first time this happened to me I wanted to murder someone:

black

Thankfully I don't use Jupyter that often and no one was harmed. 😜 Now I am wondering if this is all your fault, though! 😂 

For reference, the unadulterated signature would have fit perfectly in the popup:

Screen Shot 2020-07-14 at 4 40 14 pm

Finally, we turned off string normalisation in our black config. Even at my most generous, I would at least want the plugin to respect our black preferences. =)

So, my preference is to close this PR, although the indentation and `...` fixes I would of course keep!

sofroniewn:
> Your feedback on whether black in docstring is desirable (or not) and whether it should be configurable on a per file/function basis and how is appreciated.

I am maybe a little more open minded here then @jni and am excited about this with the following caveat.

> Even at my most generous, I would at least want the plugin to respect our black preferences. =)

I certainly agree with this, we want to use our black preferences, including the single vs double quote and the line length. This seems like a very reasonable request and I don't think I'd need additional configuration on a per file/ function basis.

As to the function signature expansion that @jni mentioned that does seem undesirable, but havn't we already crossed that bridge? I don't see how that screenshot is connected to the changes in this PR.

Carreau:
> For reference, the unadulterated signature would have fit perfectly in the popup:

Oh, that's bad. but we can likely reformat tha in IPython itself to put it in a singe line.

> I certainly agree with this, we want to use our black preferences, including the single vs double quote and the line length. This seems like a very reasonable request and I don't think I'd need additional configuration on a per file/ function basis.

That is reasonable, and the per-function basis would be for case like:

```
     Examples
     --------
-    >>> colors = np.array([[  0. , 1.,  1.  ],
-    ...                    [  1.1, 0., -0.03],
-    ...                    [  1.2, 1.,  0.5 ]])
+    >>> colors = np.array([[0.0, 1.0, 1.0], [1.1, 0.0, -0.03], [1.2, 1.0, 0.5]])
```

I'll see how I can find out how to load the black configuration. Toughest will likely be line length.

jni:
> As to the function signature expansion that @jni mentioned that does seem undesirable, but haven't we already crossed that bridge?

No, what I'm saying is, we will get in similar situations where black will mess up formatting where it was fine — such as in the array example that @Carreau provided above. In short, I think *judicious* use of black is useful, but I don't like the trend of using black everywhere — of which my example was just that, an example, to illustrate why "black everywhere" is bad. Sometimes black produces really terrible formatting, and in the case of examples that are *primarily* designed to be read, more than executed, that is a very bad thing.

As a side note, @NelleV and I have grand plans to fork black into gray, a formatter that won't mess with PEP8-compliant code. It occurs to me @Carreau that this might be the sort of thing you love to build. 😂 https://github.com/NelleV/gray

Note that we have literally done no work on this yet. 😂

Carreau:
I hear you, though I think that something like `gr[a|e]y`, is way more complicated to write than black or pep8, in the first you only have to complain, in the other you can work with an AST and throw away a lot of details. 

You may want to look into [redbarron](https://redbaron.readthedocs.io/en/latest/) which should allow you to get a Concrete Syntax Tree that you can manipulate.

And thanks, the feedback on those pull request will help me tune the features.

sofroniewn:
@Carreau is this something you think is still worth pursuing? Doing some housekeeping right now and came across this one

PR #1416, opened 2020-07-10
Revert event handler for now

tlambert03:
# Description
as per discussion in meeting today, this PR reverts the event handler stuff for now... for the upcoming sprint, and until we can discuss further.

I'm trying to retain the layerlist refactor from #1391 ... though it's currently broken due to some residual event handler logic.

## Type of change
reversion

# References

# How has this been tested?

- [ ] example: the test suite for my feature covers cases x, y, and z
- [ ] example: all tests pass with my change

## Final checklist:
- [ ] My PR is the minimum possible work for the desired functionality
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #1452, opened 2020-07-14
Preserve tests from EVH revert

sofroniewn:
# Description
This PR preserves tests added during the EVH refactor #1376 and follow-ups. Each of these tests was added when a bug from the EVH refactor was being fixed but also pass on master now. They will prevent further bugs slipping in during refactoring in the future.

See https://github.com/napari/napari/pull/1416#pullrequestreview-446098316 for more context.

I've actually moved in here now a few tests from #1404 which is unlikely to merge and from #1412 which we closed


sofroniewn:
Had to clean up some things - this is now ready for review!

tlambert03 (approved):

jni (approved):
🎉 good stuff

PR #1453, opened 2020-07-14
Perfmon System version 2

pwinston:
# Description
Updates to the **perfmon system** which were made while investigating and solving #1320:
1. Use monkey patching from a config file instead of decorators.
2. Config file has "sets" of callables you can toggle on and off.
3. Config file has new option to turn Qt event tracing on or off.
4. Added `kwargs` to `perf_timer` to display per-event key/values in Chrome tracing.
5. Added `add_instant_event()`per API to add vertical lines in Chrome tracing.
6. New `napari.utils.patcher.py` does the monkey patching, it's generic not perf specific.

# Type of change
- [x] Update existing feature (non-breaking change which adds functionality)

# Example 1

87239376-7caaa100-c3dc-11ea-9846-f0258f82b6ce

# Example 2

napari-slow-8k

# Notes

### New NAPARI_TRACE_FILE env var

Set this to a file path and Napari will trace to file on startup. I basically always use this and never start a trace from the debug menu anymore, although we should keep that option because it's more discoverable and useful for one-time tracing. This is how I tend to use perfmon today:

1. Set `NAPARI_TRACE_FILE` to some local `latest.json` file
2. Run Napari, do what I want, exit with Command-Q (hitting red circle to exit does not work?)
3. Drop `latest.json` into `chrome://tracing`.

So every time I run napari I get a trace file, and if I care what's in it I just throw it into Chrome.

### Checking NAPARI_PERFMON

Before we took care to check the `NAPARI_PERFMON` env var without importing perf in a few place. Now with a config file we have to check the settings in that file to know whether to trace Qt events, so it's more natural to import perf. We check whether to trace Qt events like this:

```
if perf_config and perf_config.trace_qt_events:
    # setup tracing
```

### Perfmon Config File

If `NAPARI_PERFMON=1` we turn on Qt Event tracing like before, and we do nothing else. If `NAPARI_PERFMON` is a file path however, we load that as our perfmon config file. The format of the config file is:

```
{
    "trace_qt_events": true,
    "trace_callables": [
        "chunk_loader",
        "vispy_rendering"
    ],
    "callable_lists": {
        "chunk_loader": [
            "napari.components.dims.Dims.set_point",
            "napari.components.viewer_model.ViewerModel._update_layers",
            . . . 
        ],
        "vispy_rendering": [
            "vispy.app.canvas.Canvas.set_current",
            "vispy.gloo.program.Program.draw",
            "vispy.gloo.context.GLContext.flush_commands",            
            . . . 
        ]
    }
}
```

The `PerfmonConfig` class monkey patches our `perf_timer` context object into each function or method above. This means we don't have to modify the code at all, not even to add a decorator. And it also means we can have different trace sets, even overlapping ones, that we can toggle on or off as needed. You want enough tracing to see what you want, but not so much that the overhead is annoying.

### Add kwargs to perf_timer

Now `PerfTimer` has a `kwargs` field that lets you set arbitrary key=value pairs that will be displayed into the `chrome://tracing` GUI when you click on an event. With monkey patching you can't use these, but with a manual `perf_timer` you can.

### New add_instant_event()

New `add_instant_event()` is shown in `chrome://tracing` as a vertical line. This is useful because very short events are invisible unless you zoom way in. But instant events are always visible as lines. For instance when debugging I created instant events for `MousePressed` and `MouseReleased`. There are 3-4 other graphical things in the [chrome tracing spec](https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/preview) that we aren't using yet, need to look into those.

### "Tracing" vs. "Timing"

I've been switching to the term "tracing" based on my experience using the perf stuff. The Chrome feature is called `chrome://tracing` and tracing really is the right term. There is timing information, but just visualizing how the code executes is a big part of things, beyond the timing information. I might rename the whole module from "perf" to "tracing" at some point.

### Call Stack

The `perf_timer` wrappers clutter up the stack. If you want to debug might want to turn off perfmon.

### Patcher

The new `napari.utils.patcher.py` is actually totally generic for patching functions/methods. The perf code uses `patcher` to patch in `perf_timer` but you could use it to patch anything. Not sure if there is some library (event `wrapt` itself?) that can patch lists of callables like we are doing, but it was not hard to implement. It's only 150 lines with comments.

# How has this been tested?
- [x] Manual testing while investigation of #1320 

## Final checklist:


pwinston:
I still need to polish some comments and add types, but ready for high level review.

sofroniewn (approved):
Wow, this is amazing! Just gave it a spin and everything works great. I love the config file, so great to be able to specify the callables that I want to trace, I can imagine this being incredibly useful when developing and debuging. Right now I need to use benchmarks and airspeed velocity for all profiling, which can be nice for seeing everything, but I see this as being much faster and easier to use when figuring things out the first time.

Curious how @tlambert03 thinks about this, but I guess after we merge #1183 we'll probably want to combine that config file into that approach, but that can come later. Definitely nothing that needs to hold this PR up.

> Now PerfTimer has a kwargs field that lets you set arbitrary key=value pairs that will be displayed into the GUI when you click on an event. With monkey patching you can't use these, but with a manual perf_timer you can.

When you say GUI here you mean the chrome tracing GUI right? not napari? this is how you injected your additional metadata that you wanted to view in the tracing tool right? just want to make sure I understand.

> but you could use it to patch anything

Just curious what else could you imagine we might want to patch in like this?

### final comment

I also read through the whole code and everything seemed very simple and clear - I won't claim to understand every line after this read, but it all looks very good to me and well documented. I think this is good to merge, but will let @jni and @tlambert03 review before then

pwinston:
> When you say GUI here you mean the chrome tracing GUI right? not napari? 

Good point. I corrected it above to say "chrome://tracing GUI". Before I actually just tacked the key/value pairs onto the string name like `with perf_timer(f"calculate count="{count}")`. In some ways that's even better because the values are visible in the timeline rectangles themselves. However with this `kwargs` feature you can do this:

```
        with perf_timer("clear_pending", data_id=request.data_id, my_value="pizza"):
```

When you click on that event in Chrome you can see your values under **Args**:

Screen Shot 2020-07-16 at 7 45 26 AM

So this might be useful if you had a lot of values or big values. In fact looking at those **Args** it looks like the variable view in the debugger. And the nested rectangles depict the stack. So tracing is sort of a marriage between "debug prints" and "debugger technology". It's after the fact, but a fairly rich display of what happened.

I'm bullish on tracing being useful well beyond timing. From my experience debugging with it, the code could statically look good but the trace looks wonky. You might see A calls B and B takes unexpectedly long. Or why do we call C from 3 different places? Or doing D triggers a ton of stuff. And it's all through the lens of time, because really fast stuff basically disappear unless you zoom way in.

> Just curious what else could you imagine we might want to patch in like this?

I'm not really sure. At first the patcher was perf-specific but then I realized perf could just pass `perf_timer` as a `patch_func` to the patcher. That made the patcher is generic, so I pulled it out of the `utils.perf` directory.

I guess you could have a `patch_func` that created some totally different profiling or tracing format? You could maybe disable or replace/mock some functions for testing or other reasons? Really anything you could do with a decorator, but without actually modifying the code to add the decorator. And being able to do it in other people's modules/packages. Although as we learned with Qt you can't monkey patch some C++/sip functions.

The patcher format is so simple and useful I'm guessing there are packages that do the same thing? Maybe even `wrapt` itself? I didn't hunt around since it was easy enough to just do it, the code is like 150 lines. Taking a list like this and patching all the functions and method is pretty nice. And we could add wildcards to make it even more powerful.

```
            "napari.components.dims.Dims.set_point",
            "napari.components.viewer_model.ViewerModel._update_layers",
            "napari.layers.image._image_slice.ImageSlice.chunk_loaded",
            "napari.layers.image._image_slice.ImageSlice.load_chunk",
```

tlambert03 (approved):
I'm with @sofroniewn... this looks amazing & useful and the code is very clean and readable.  I also can't claim to fully follow everything that's happening here without spending more time than I currently have... but since this is still an opt-in feature, I have no issues merging this to master.

I guess one request for a follow-up PR would be a bit of a "beginners guide" to using it.  I'm new to this sort of performance tracing, so it's not totally clear to me whether the only output here is the "`latest.json`" file which should be evaluated in `chrome://tracing`, or whether there are other ways that you're using this.  I wouldn't want you to regurgitate some chrome tracing tutorial that is already available elsewhere, but some docs that point to useful resources, with a quick example workflow (e.g. create something that doesn't perform that well, monitor it, fix it, and show how you know it got fixed) would be great.

pwinston:
Yes I think it very much needs kind of walkthrough of usage. I can create a folder `docs/source/perfmon` maybe? With the title "Performance Monitoring"?

Although as an aside, I've grown to somewhat feel that "tracing" is actually a bigger thing than just "performance monitoring". I think "tracing" is kind of purpose-neutral in the same way that "logging" is purpose-neutral. And in a way tracing is just logging, but it's logging in a very specific format.

Because you could use tracing or logging to investigate performance issues, but you could use them for other reasons. A lot of the traces I made working on async where not really for "performance" they were getting things working correctly, at whatever speed.

That's a pretty subtle point though. All the code is in `utils.perf` and it was written with performance monitoring in mind. So I think using the umbrella term is fine. If I were to start over from scratch I might brand it all "tracing" instead, but I don't think it's worth changing.

sofroniewn:
> Yes I think it very much needs kind of walkthrough of usage. I can create a folder docs/source/perfmon maybe? With the title "Performance Monitoring"?

Sounds great @pwinston thanks. I think I'll see if @jni has any additional comments on this PR in the next few hours and if not will merge before I go to sleep tonight so you can make the updated async PR tomorrow morning. That docs change can then come in its own PR.

> Although as an aside, I've grown to somewhat feel that "tracing" is actually a bigger thing than just "performance monitoring". I think "tracing" is kind of purpose-neutral in the same way that "logging" is purpose-neutral.

Good points! Exciting to get this in!!

jni:
> I can create a folder docs/source/perfmon maybe

I'd stick it in docs/source/developers, where benchmarks live. (For now, anyway.) In the future, maybe we can collect this together with your previous performance doc in a "speed" folder?

btw why are all the filenames yelling at me in that folder? :joy: All in favour of changing the filenames to lowercase? 🙋

> I'll see if @jni has any additional comments on this PR in the next few hours

Gonna have a look in a bit but if you don't hear from me, go ahead. =)

napari/utils/patcher.py
```diff
@@ -0,0 +1,212 @@
...
+        patch_func(parent, callable_str, label)
+
+
+class Importer:
+    def __init__(self):
+        # Keep track of modules so we don't re-import them over and over.
+        # This is probably not needed because importing a module the second
+        # time seems instant. But it's easy to cache these and possibly
+        # it's better in some way.
+        self.modules: Dict[str, Module] = {}
```
jni: This seems like a reimplementation of [`sys.modules`](https://docs.python.org/3/library/sys.html#sys.modules)?

napari/utils/patcher.py
```diff
@@ -0,0 +1,212 @@
+"""Patch functions and methods. Our perfmon system using this to patch in
+perf_timers, but this can be used for any type of patching.
+
+See patch_callables() below as the main entrypoint.
+"""
```
jni: Overall, this module seems like a lot of OO overkill (OOverkill? I like it!). It defines a Module object, but Python modules are already objects, which have the required attributes (`module.__path__`, `module.__name__`, etc.). The method `patch_attribute` could easily (I think!) be implemented as a function that takes a Python module as input. It defines an Importer object, ostensibly for caching but importlib already does importing and modules are already cached in sys.modules. Finally, the Patcher class seems to be there to keep track of the Importer, but again, that's not really necessary — if using built-ins, this could just be defined as functions.

As others have pointed out, this is all internal to napari (might be good to note that by renaming this file to `_patcher.py`), so a refactor can come later, but I think it should come for sure.

napari/utils/perf/__init__.py
```diff
@@ -12,7 +12,15 @@
...
+
+from ._compat import perf_counter_ns
+from ._config import perf_config
 from ._event import PerfEvent
-from ._timers import timers
+from ._timers import timers, add_instant_event
 from ._utils import perf_timer, perf_func
+
+
+# If not using perfmon timers will be 100% disable with hopefully zero
```
jni:
```suggestion
# If not using perfmon timers will be 100% disabled with hopefully zero
```

napari/utils/perf/_compat.py
```diff
@@ -1,11 +1,11 @@
...
 """
+import sys
 import time
 
-from ._config import PYTHON_3_7
+PYTHON_3_7_OR_NEWER = sys.version_info[:2] >= (3, 7)
 
-
-if PYTHON_3_7:
+if PYTHON_3_7_OR_NEWER:
```
jni: I feel like this doesn't need to be defined if it only gets used in this one spot? `sys.version_info` is pretty standard.

napari/utils/perf/_config.py
```diff
@@ -1,11 +1,150 @@
...
+        self.patched = False
+
+        self.config_path = config_path
+        if config_path is None:
+            return  # Legacy mode, trace Qt events only.
+
+        path = Path(config_path)
+        if not path.exists():
+            raise FileNotFoundError(
+                errno.ENOENT, f"Config file {PERFMON_ENV_VAR} not found", path,
```
jni: I feel like this should be the filename, ie `{os.getenv(PERFMON_ENV_VAR)}`?

napari/utils/perf/_config.py
```diff
@@ -1,11 +1,150 @@
...
+        This function potentially imports a lot of your modules.
+        """
+        if self.config_path is None:
+            return  # disabled
+
+        assert self.patched is False
+        self._patch_callables()
+        self.patched = True
+
+    def _get_callables(self, callable_list) -> List[str]:
```
jni: What is 'callable_list'? It should be type annotated, or annotated in the docstring, or both.

napari/utils/perf/_event.py
```diff
@@ -1,32 +1,49 @@
...
         The name of this event like "draw".
     start_ns : int
         Start time in nanoseconds.
     end_ns : int
         End time in nanoseconds.
-
+    category :str
+        Comma separated categories such has "render,update".
+    **kwargs : dict
+        Additional keyword arguments for the "args" field of the event.
```
jni, line 26: Can you add an Attributes section with descriptions of the attributes? e.g. I was surprised to read `self.type = "X"` :joy: but it might make sense with a description.

napari/utils/perf/_event.py
```diff
@@ -1,32 +1,49 @@
...
+        pid=os.getpid(),
+        **kwargs,
+    ):
         self.name = name
         self.start_ns = start_ns
         self.end_ns = end_ns
+        self.category = category
+        self.args = kwargs
+        self.pid = pid
+        self.tid = threading.get_ident()
```
jni: `pid` is pretty common, but I've never heard of tid... I wonder if these could/should be renamed to `process_id` and `thread_id`?

napari/utils/perf/_event.py
```diff
@@ -47,3 +64,9 @@ def duration_us(self):
     @property
     def duration_ms(self):
         return self.duration_ns / 1e6
+
+
+class InstantEvent(PerfEvent):
+    def __init__(self, name: str, time_ns: int, **kwargs):
+        super().__init__(name, time_ns, time_ns, **kwargs)
+        self.type = "I"  # instant event
```
jni: What's the difference between these two event types, and if there is a class for each, why not use `type(event)`?

napari/utils/perf/_timers.py
```diff
@@ -52,24 +56,35 @@ def __init__(self):
...
+            self.trace_file.add_event(event)
 
-        # Update our self.timers (in milliseconds).
-        name = event.name
-        duration_ms = event.duration_ms
-        if name in self.timers:
-            self.timers[name].add(duration_ms)
-        else:
-            self.timers[name] = Stat(duration_ms)
+        if event.type == "X":
```
jni: ie here you would write `if type(event) == PerfEvent`

napari/utils/perf/_trace_file.py
```diff
@@ -71,22 +84,33 @@ def write_event(self, event: PerfEvent) -> None:
...
         # know the duration. The format wants times in micro-seconds.
         data = {
-            "pid": self.pid,
+            "pid": event.pid,
+            "tid": event.tid,
             "name": event.name,
-            "cat": event.category,
-            "ph": "X",
+            "cat": category,
+            "ph": event.type,
```
jni: Ah, this answers my questions above. What is `ph` short for and is there a better description we could use for our attribute?

napari/utils/perf/_trace_file.py
```diff
@@ -71,22 +84,33 @@ def write_event(self, event: PerfEvent) -> None:
...
+
+        if event.type == "X":
+            data["dur"] = event.duration_us
+        else:
+            assert event.type == "I"
+            # For instant events the scope "s" is one of:
+            #     "g" - global
+            #     "p" - process
+            #     "t" - thread
+            data["s"] = "p"
```
jni, line 113: It's unclear to me what scope means in this context or why it would always be p here?

napari/utils/perf/_utils.py
```diff
@@ -1,30 +1,38 @@
-"""Utilities to support performance monitoring:
+"""Performance monitoring utilities.
```
jni: meta-utils! :joy:

napari/utils/perf/_utils.py
```diff
@@ -81,29 +96,26 @@ def draw(self):
 
         def decorator(func):
             @functools.wraps(func)
-            def time_function(*args, **kwargs):
+            def wrapper(*args, **kwargs):
```
jni: I like the previous name better, as it's more descriptive... But maybe `timed_function`?

jni: Also, the multiple nesting can be avoided with `toolz.curry`, which we already depend on and use in the `thread_worker` decorator:

```python
@tz.curry
def perf_func_named(timer_name : str, func : Callable) -> Callable:
    @functools.wraps(func)
    def timed_func(*args, **kwargs):
        with perf_timer(timer_name, "decorator"):
            return func(*args, **kwargs)
    return timed_func
```

By the way, what is the second argument, 'decorator', to perf_timer?

jni: ... And you could even combine things with the above non-named decorator:

```python
@tz.curry
def perf_func(func : Callable, timer_name : str = None) -> Callable:
    if timer_name is None:
        timer_name = f"{func.__qualname__}"

    @functools.wraps(func)
    def timed_func(*args, **kwargs):
        with perf_timer(timer_name, "decorator"):
            return func(*args, **kwargs)
    return timed_func
```

Then you would use

```python
func = perf_func(func, timer_name=my_name)
```

(using keyword arguments for `timer_name`) when wanting to overwrite the name, but just `func = perf_func(func)` when not.

I'm not sure what `full_name` is above but the same principle applies.

napari/utils/perf/_utils.py
```diff
@@ -81,29 +96,26 @@ def draw(self):
...
-        perf_timer = contextlib.nullcontext()
-    else:
-
-        @contextlib.contextmanager
-        def perf_timer(name: str):
-            yield
+    # contextlib.nullcontext does not work with kwargs?
+    @contextlib.contextmanager
+    def perf_timer(name: str, category: Optional[str] = None, **kwargs):
+        yield
```
jni: Does this context ever get used? I thought if perfmon was off there were zero runtime modifications?

napari/viewer.py
```diff
@@ -161,3 +161,15 @@ def close(self):
...
+def _convert_app_for_tracing(app):
+    """Return a new QApplicationWithTracing if we don't have it already.
+    """
+    from ._qt.qt_event_tracing import convert_app_for_tracing
+
+    # For tracing Qt events we need a special QApplication. If using
+    # `gui_qt` we already have the special one, and no conversion is done
+    # here. However when running inside IPython or Jupyter this is where we
+    # switch out the QApplication.
+    return convert_app_for_tracing(app)
```
jni: Why did you pull this out into a pass-through function...? The deleted code above seems clearer than this, which adds another layer of indirection...

jni (commented):
@pwinston I've made quite a few comments. They are all about code style and not functionality (although I did ask a question about the "empty" context manager. Also, is there any testing of all this? Given the changes in this PR I would expect some tests to change/be added, but there's no tests here.

Happy to postpone the suggested changes to later PRs, although happier still if they could happen here, since my experience is that "later PRs" often never come...!

Also, there may well be complexities I haven't appreciated in some of my comments — if so, please feel free to point them out!

Thanks!!!

napari/utils/patcher.py
pwinston: It was more as functions at first but it grew in this direction. I didn't think of using `sys.modules` that would simplify things.  And then I didn't realize `module.__name__` is the "path" in my terms, it's `napari.utils.perf` or whatever.

Any class can be mechanically turned into a module where the attributes become module globals and the methods become functions. I don't think that's automatically a good idea just because it's possible. But using `sys.modules` and `module` here does make sense. And if after making those changes there just was no point to some of the classes it might make sense to get rid of them. I'll try a version 2 of it.

napari/utils/perf/_compat.py
pwinston: Yeah I will just use `sys.version_info` directly. It used to be used in multiple places but no more.

napari/utils/perf/_trace_file.py
pwinston: The "scope" here is strictly a chrome://tracing term. It's saying what type of "instant event" is this? Is it an event that applies only to one thread, to one process, or to the whole system. It basically just comes down to when `chrome://tracing` draws the "instant event" where does it draw on the line.

I just hardcoded it to process based on how I was using it. Probably we should expose all 3 options since it's not hard. I just didn't even experiment with them to see what they looked like, so I didn't want to expose them without at least trying them.

napari/viewer.py
pwinston: I agree the one line `_convert_app_for_tracing` is not needed. I think I had it doing more at first and it whittled down to just the one line. I'll get rid of it.

napari/utils/patcher.py
pwinston: Now relies on sys.modules in instead of this.

napari/utils/patcher.py
pwinston: New version is all functions. Using sys.modules and module.__name__ and then unwinding some things, no more needs for the classes.

napari/utils/perf/_config.py
pwinston: That's on purpose, the `path` variable is the actual path. But it was kind of silly I deleted the existence check and this raise entirely. Now it fails like:
```
FileNotFoundError: [Errno 2] No such file or directory: '/Users/pbw/.napari-does-not-exist'
```
Which is probably fine. I was just trying to be nice and remind you what env var was involved, but probably people can figure that out.

napari/utils/perf/_trace_file.py
pwinston: I added a better comment.

napari/viewer.py
pwinston: I removed it.

napari/utils/perf/_utils.py
pwinston: The "decorator" string is the second argument of `perf_timer` which is "category", which is like a comma separated list of strings that supposed let you toggle on and off events by "category". Like "rendering" vs. "IO" maybe? I've never actually seen it in the UI. I should probably just have no category until/unless we determine it's useful for something.

I've never used `toolz.curry` so that would take some investigation, but sounds potentially useful. Same with collapsing it down to less decorators.

napari/utils/perf/_utils.py
pwinston: I used to assign it to `contextlib.contextmanager` when disabled but I had to switch to this then I added the `kwargs` argument. I would hope Python could optimize away a generator that does absolutely nothing, but I'm not super optimistic about that, I'll do some timings.

With the patcher now I use the patcher for 99% of things. So it simply doesn't patch in the `perf_timer` at all when disabled. But yeah if you leave a raw `perf_timer` in the code at all times, this null-generator will run. So I'll see if I can measure it.

napari/utils/perf/_compat.py
pwinston: Used it directly now

napari/utils/perf/_utils.py
jni: > But yeah if you leave a raw perf_timer in the code at all times

But we don't do this, right? I'd prefer to not enable this kind of pattern...

napari/utils/perf/_config.py
jni: @pwinston did you miss this comment?

napari/utils/perf/_patcher.py
```diff
@@ -0,0 +1,155 @@
...
+
+class PatchError(Exception):
+    """Failed to patch target, config file error?"""
+
+    def __init__(self, message):
+        self.message = message
+
+
+def _patch_attribute(module, attribute_str, patch_func):
+    """Patch the module's callable pointed to by the attribute string.
```
jni, line 29: This needs a full docstring.

napari/utils/perf/_patcher.py
```diff
@@ -0,0 +1,155 @@
...
+    """Import the module portion of this target string.
+
+    Try importing successively longer segments of the target_str. For example
+    with "napari.utils.chunk_loader.ChunkLoader.load_chunk" we will import:
+        napari (success)
+        napari.utils (success)
+        napari.utils.chunk_loader (success)
+        napari.utils.chunk_loader.ChunkLoader (failure)
+
+    The last one fails because ChunkLoader is a class not a module.
```
jni, line 89: Needs a Parameters section

napari/utils/perf/_patcher.py
```diff
@@ -0,0 +1,155 @@
...
+
+    Returns
+    -------
+    Tuple[types.ModuleType, str]
+        Where the module is the inner most imported module, and the string
+    is the rest of target_str that was not modules.
```
jni:
```suggestion
        is the rest of target_str that was not modules.
```

napari/utils/perf/_patcher.py
```diff
@@ -0,0 +1,155 @@
...
+    Nested classes and methods not allowed, but support could be added.
+
+    An example patch_func is:
+
+    import wrapt
+    def _my_patcher(parent: CallableParent, callable: str, label: str) -> None:
+        @wrapt.patch_function_wrapper(parent, callable)
+        def my_announcer(wrapped, instance, args, kwargs):
+            print(f"Announce {label}")
+            return wrapped(*args, **kwargs)
```
jni: This won't render properly as is. For reference, here's a cheat sheet for ReStructuredText syntax, which is what's being used in these docstrings: https://thomas-cokelaer.info/tutorials/sphinx/rest_syntax.html#inserting-code-and-literal-blocks

```suggestion
    An example patch_func is::

        import wrapt
        def _my_patcher(parent: CallableParent, callable: str, label: str):
            @wrapt.patch_function_wrapper(parent, callable)
            def my_announcer(wrapped, instance, args, kwargs):
                print(f"Announce {label}")
                return wrapped(*args, **kwargs)
```

jni (commented):
Thanks @pwinston, much simpler now from my perspective! I've made a couple of extra notes around docstrings.

I'm unclear from your comment whether we should wait to merge for you to use my toolz suggestion to simplify the timer, or whether you'd rather move on to other things! Either is fine. (But toolz is awesome ;)

napari/utils/perf/_config.py
pwinston: I think I saw it but didn't come back to it. I changed the var name to `list_name` and added the type `str`.

napari/utils/perf/_event.py
pwinston: I added attributes just for `self.type` and `self.args` and a comment for `self.type`. The "X" is just part of the `chrome://tracing` format

napari/utils/perf/_event.py
pwinston: Changed to `process_id` and `thread_id`

napari/utils/perf/_event.py
pwinston: I added some comments about the different event types, but if you Google for "Trace Event Format" there's a google doc with the full spec, it's about 19 pages.

I'm not sure what you are proposing exactly but in general with OO it's considered much better to ask objects for information like such as `write event.type`, where the object "knows" the right answer, than to externally interrogate their type like `write "I" if type(event) == InstantEvent" else "X"` which is considered more fragile.

That said confining the details of their format to one place in the code would be nice, so maybe it is worth using `type(event)` or an enum to have a translation spot from our code to their format. I think with just "X" and "I" it's not bad as-is, but maybe if we add more even types we should consider that.

napari/utils/perf/_timers.py
pwinston: Addressed above.

napari/utils/perf/_trace_file.py
pwinston: `ph` is `phase type` but in the spec they define it like this `ph: The event type.` so that's why we call it `event.type`, we could call it `event.phase` I guess "type" is kind of overloaded in Python. The whole definition from their spec:

>  ph: The event type. This is a single character which changes depending on the type of event being output. The valid values are listed in the table below. We will discuss each phase type below.

pwinston: I changed it to `event.phase` I think that's much clearer. Both `event` and `type` are very overloaded so `event.type` was confusing. They do call it "event type" in places, but phase is the official name.

napari/utils/perf/_patcher.py
pwinston: I was sort of a assuming "private" functions/methods did not necessarily need docstrings. Although sometimes I have added them anyway, and I can here. But did I just invent that guideline, does everything need docstrings equally whether public or "private"?

napari/utils/perf/_patcher.py
pwinston: Added parameters section

napari/utils/perf/_utils.py
pwinston: There's no `perf_timers` checked in except one in the class `QApplicationWithTracing` but that is separately disabled. So there's no "active" `perf_timers` checked in as of this PR. The developer is expected to add them using the patcher. 

However there's two reasons you might want to temporarily add a `perf_timer` by hand while doing an intensive investigation of something:

1. You want to time a line of code or a block of code within a bigger function. And you don't (yet) want to refactor that code to extract that block into its own function that you can patch.

2. You want to add arguments to the timer. This is really for "tracing" not such much "timing". For example you could put the `shape` of an array as an argument to a `perf_timer` or the filename of what chunk you are loading, then in the chrome GUI you can see those values and that might help explain the timeline.

So I'd consider `perf_timers` like anything you'd add temporarily while debugging, like noisy print statement. Something you should remove before you commit or at least **before** you merge. This should be covered in the documentation I haven't written. Basically explain patching is "free" but manually added `perf_timers` are not completely free. I timed a "do nothing" generator as taking like a microsecond.

If there was a problem with people commiting and merging `perf_timers` we could write a CI step that checked for them, but I wouldn't bother unless it becomes a problem.

napari/utils/perf/_patcher.py
pwinston: I added a Parameters section and type annotations.

napari/utils/perf/_utils.py
pwinston: I just deleted `perf_func` and `perf_func_named`. Those came long before we had the patcher. With the patcher there's really no reason to have these. The patcher patches in `perf_timers` it does not use these decorators. And patching using the patcher is really better in every way that using these decorators except you lose the "name" part. But we could actually add that to the patcher if we wanted.

That is the patcher config file could support something like:

    ("module1.Class2.method2", "redraw everything")

meaning patch `method2` to give the timer the name "redraw everything".

So if we ever need names like that I'd add them with the patcher.

pwinston:
I think everything is resolved. All feedback/changes were beneficial I think, improved things.

Only totally missing thing is external documentation, like a "how to" guide. I don't think that needs to be part of this PR relative to getting async further along

sofroniewn:
> I think everything is resolved. All feedback/changes were beneficial I think, improved things.

Great! Thanks for the review @jni and making those changes @pwinston!

> I don't think that needs to be part of this PR relative to getting async further along

Agreed. I will wait till tests pass here and then merge shortly. I'm looking forward to checking out the async PR!!

pwinston:
I snuck in the previous commit to get rid of `NAPARI_TRACE_FILE` env var which was now legacy. I switched it to use the config file instead, since we're doing everything else with the config file. So my config file starts:

```
{
    "trace_qt_events": true,
    "trace_file_on_start": "/Users/pbw/Desktop/perf/latest.json",
```

Which means if perfmon is enabled it starts writing to latest.json as soon as it starts. I find this really useful because when investigating something I just want to:

1. Start napari.
2. Do the thing I'm testing.
3. Hit command-Q to exit
4. Drag the `latest.json` file into Chrome to view it.

No need to use our Debug->Start Trace File menu option. Although I think it makes sense to keep that option.

sofroniewn:
> No need to use our Debug->Start Trace File menu option. Although I think it makes sense to keep that option.

Yes, love this! Very nice

jni (approved):

PR #1456, opened 2020-07-15
Move contrast limits and gamma to the shader by vendoring vispy code

sofroniewn:
# Description
This PR supersedes #1056 by porting @tlambert03 newly contributed vispy code from https://github.com/vispy/vispy/pull/1844 and https://github.com/vispy/vispy/pull/1842 for Image and Volume which have merged or are close to merge respectively at vispy but are not yet released.

The PR tries to keep the vendored code well isolated though an additional line  `uniform float u_attenuation;` is present in line 88 of `napari/_vispy/vendored/volume.py` as we have yet to contribute the attenuated max intensity projection shader back to vispy.

The most exciting thing about the PR are the performance improvements, and these are particularly impressive for the 3D case

### before this PR
```
(base) czirwc1macos2701:napari nsofroniew$ asv run -b QtImageRenderingSuite --python=same
· Discovering benchmarks
· Running 2 total benchmarks (1 commits * 1 environments * 2 benchmarks)
[  0.00%] ·· Benchmarking existing-py_Users_nsofroniew_opt_anaconda3_bin_python
[ 25.00%] ··· Running (benchmark_qt_viewer_image.QtImageRenderingSuite.time_change_contrast--).
[ 50.00%] ··· Running (benchmark_qt_viewer_image.QtImageRenderingSuite.time_change_gamma--).
[ 75.00%] ··· benchmark_qt_viewer_image.QtImageRenderingSuite.time_change_contrast                                                                                                                        ok
[ 75.00%] ··· ======== =============
               param1
              -------- -------------
                 16      2.64±0.2ms
                 32      2.73±0.3ms
                 64      2.59±0.2ms
                128      2.64±0.2ms
                256     2.59±0.08ms
                512     2.40±0.07ms
                1024    2.46±0.06ms
                2048    2.42±0.07ms
                4096     2.64±0.4ms
              ======== =============

[100.00%] ··· benchmark_qt_viewer_image.QtImageRenderingSuite.time_change_gamma                                                                                                                           ok
               param1
              -------- ------------
                 16     5.47±0.7ms
                 32     5.58±0.4ms
                 64     5.22±0.3ms
                128     5.21±0.3ms
                256     5.18±0.3ms
                512     5.37±0.4ms
                1024    5.35±0.4ms
                2048    5.36±0.4ms
                4096    5.22±0.3ms
              ======== ============

(base) czirwc1macos2701:napari nsofroniew$ asv run -b QtVolumeRenderingSuite --python=same
· Discovering benchmarks
· Running 2 total benchmarks (1 commits * 1 environments * 2 benchmarks)
[  0.00%] ·· Benchmarking existing-py_Users_nsofroniew_opt_anaconda3_bin_python
[ 25.00%] ··· Running (benchmark_qt_viewer_image.QtVolumeRenderingSuite.time_change_contrast--).
[ 50.00%] ··· Running (benchmark_qt_viewer_image.QtVolumeRenderingSuite.time_change_gamma--).
[ 75.00%] ··· benchmark_qt_viewer_image.QtVolumeRenderingSuite.time_change_contrast                                                                                                                       ok
[ 75.00%] ··· ======== =============
               param1
              -------- -------------
                 16      7.72±0.1ms
                 32      7.99±0.1ms
                 64     11.4±0.08ms
                128      37.7±0.7ms
                256       251±4ms
                512      3.13±0.6s
              ======== =============

[100.00%] ··· benchmark_qt_viewer_image.QtVolumeRenderingSuite.time_change_gamma                                                                                                                          ok
[100.00%] ··· ======== =============
               param1
              -------- -------------
                 16     5.22±0.06ms
                 32      5.37±0.2ms
                 64     6.45±0.07ms
                128     17.0±0.08ms
                256       81.6±2ms
                512       551±4ms
              ======== =============
```

### after this PR
```
(base) czirwc1macos2701:napari nsofroniew$ asv run -b QtImageRenderingSuite --python=same
· Discovering benchmarks
· Running 2 total benchmarks (1 commits * 1 environments * 2 benchmarks)
[  0.00%] ·· Benchmarking existing-py_Users_nsofroniew_opt_anaconda3_bin_python
[ 25.00%] ··· Running (benchmark_qt_viewer_image.QtImageRenderingSuite.time_change_contrast--).
[ 50.00%] ··· Running (benchmark_qt_viewer_image.QtImageRenderingSuite.time_change_gamma--).
[ 75.00%] ··· benchmark_qt_viewer_image.QtImageRenderingSuite.time_change_contrast                                                                                                                        ok
[ 75.00%] ··· ======== =============
               param1
              -------- -------------
                 16     2.32±0.02ms
                 32     2.37±0.05ms
                 64     2.37±0.03ms
                128     2.34±0.03ms
                256     2.41±0.04ms
                512     2.51±0.06ms
                1024    2.45±0.08ms
                2048    2.48±0.03ms
                4096     2.37±0.1ms
              ======== =============

[100.00%] ··· benchmark_qt_viewer_image.QtImageRenderingSuite.time_change_gamma                                                                                                                           ok
[100.00%] ··· ======== =============
               param1
              -------- -------------
                 16     2.06±0.02ms
                 32     2.08±0.05ms
                 64     2.06±0.05ms
                128     2.06±0.01ms
                256     2.06±0.02ms
                512     2.10±0.03ms
                1024    2.11±0.02ms
                2048    2.07±0.03ms
                4096    2.10±0.02ms
              ======== =============

(base) czirwc1macos2701:napari nsofroniew$ asv run -b QtVolumeRenderingSuite --python=same
· Discovering benchmarks
· Running 2 total benchmarks (1 commits * 1 environments * 2 benchmarks)
[  0.00%] ·· Benchmarking existing-py_Users_nsofroniew_opt_anaconda3_bin_python
[ 25.00%] ··· Running (benchmark_qt_viewer_image.QtVolumeRenderingSuite.time_change_contrast--).
[ 50.00%] ··· Running (benchmark_qt_viewer_image.QtVolumeRenderingSuite.time_change_gamma--).
[ 75.00%] ··· benchmark_qt_viewer_image.QtVolumeRenderingSuite.time_change_contrast                                                                                                                       ok
[ 75.00%] ··· ======== =============
               param1
              -------- -------------
                 16     2.45±0.07ms
                 32     2.62±0.04ms
                 64     3.75±0.09ms
                128      13.9±0.6ms
                256      77.9±0.7ms
                512       582±9ms
              ======== =============

[100.00%] ··· benchmark_qt_viewer_image.QtVolumeRenderingSuite.time_change_gamma                                                                                                                          ok
[100.00%] ··· ======== =============
               param1
              -------- -------------
                 16     2.24±0.04ms
                 32     2.28±0.08ms
                 64      3.34±0.1ms
                128      13.3±0.6ms
                256       77.3±1ms
                512       571±4ms
              ======== =============
```

One question I have for @tlambert03 is if he wants to make better use of the `contrast_limits_range` parameter in this new system - I recall those values potentially being useful, and I'm not making use of them here. Feel free to update this PR directly if that is the case.

Otherwise this is ready for review, with all tests passing locally (and all tests unchanged). 

## Type of change

# References

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/_vispy/image.py
+++ b/napari/_vispy/image.py
@@ -1,0 +1,5 @@
+from .vendored import ImageVisual
+from vispy.scene.visuals import create_visual_node
+
+
+Image = create_visual_node(ImageVisual)
--- a/napari/_vispy/vendored/__init__.py
+++ b/napari/_vispy/vendored/__init__.py
@@ -1,0 +1,2 @@
+from .volume import VolumeVisual
+from .image import ImageVisual
--- a/napari/_vispy/vendored/image.py
+++ b/napari/_vispy/vendored/image.py
@@ -1,0 +1,564 @@
+# -*- coding: utf-8 -*-
+# Copyright (c) Vispy Development Team. All Rights Reserved.
+# Distributed under the (new) BSD License. See LICENSE.txt for more info.
+
+from __future__ import division
+
+import numpy as np
+
+from vispy.gloo import Texture2D, VertexBuffer
+from vispy.color import get_colormap
+from vispy.visuals.shaders import Function, FunctionChain
+from vispy.visuals.transforms import NullTransform
+from vispy.visuals.visual import Visual
+from vispy.ext.six import string_types
+from vispy.io import load_spatial_filters
+
+VERT_SHADER = """
+uniform int method;  // 0=subdivide, 1=impostor
+attribute vec2 a_position;
+attribute vec2 a_texcoord;
+varying vec2 v_texcoord;
+
+void main() {
+    v_texcoord = a_texcoord;
+    gl_Position = $transform(vec4(a_position, 0., 1.));
+}
+"""
+
+FRAG_SHADER = """
+uniform vec2 image_size;
+uniform int method;  // 0=subdivide, 1=impostor
+uniform sampler2D u_texture;
+varying vec2 v_texcoord;
+
+vec4 map_local_to_tex(vec4 x) {
+    // Cast ray from 3D viewport to surface of image
+    // (if $transform does not affect z values, then this
+    // can be optimized as simply $transform.map(x) )
+    vec4 p1 = $transform(x);
+    vec4 p2 = $transform(x + vec4(0, 0, 0.5, 0));
+    p1 /= p1.w;
+    p2 /= p2.w;
+    vec4 d = p2 - p1;
+    float f = p2.z / d.z;
+    vec4 p3 = p2 - d * f;
+
+    // finally map local to texture coords
+    return vec4(p3.xy / image_size, 0, 1);
+}
+
+
+void main()
+{
+    vec2 texcoord;
+    if( method == 0 ) {
+        texcoord = v_texcoord;
+    }
+    else {
+        // vertex shader outputs clip coordinates;
+        // fragment shader maps to texture coordinates
+        texcoord = map_local_to_tex(vec4(v_texcoord, 0, 1)).xy;
+    }
+
+    gl_FragColor = $color_transform($get_data(texcoord));
+}
+"""  # noqa
+
+_interpolation_template = """
+    #include "misc/spatial-filters.frag"
+    vec4 texture_lookup_filtered(vec2 texcoord) {
+        if(texcoord.x < 0.0 || texcoord.x > 1.0 ||
+        texcoord.y < 0.0 || texcoord.y > 1.0) {
+            discard;
+        }
+        return %s($texture, $shape, texcoord);
+    }"""
+
+_texture_lookup = """
+    vec4 texture_lookup(vec2 texcoord) {
+        if(texcoord.x < 0.0 || texcoord.x > 1.0 ||
+        texcoord.y < 0.0 || texcoord.y > 1.0) {
+            discard;
+        }
+        return texture2D($texture, texcoord);
+    }"""
+
+_apply_clim_float = """
+    float apply_clim(float data) {
+        data = data - $clim.x;
+        data = data / ($clim.y - $clim.x);
+        return max(data, 0);
+    }"""
+_apply_clim = """
+    vec4 apply_clim(vec4 color) {
+        color.rgb = color.rgb - $clim.x;
+        color.rgb = color.rgb / ($clim.y - $clim.x);
+        return max(color, 0);
+    }
+"""
+
+_apply_gamma_float = """
+    float apply_gamma(float data) {
+        return pow(data, $gamma);
+    }"""
+_apply_gamma = """
+    vec4 apply_gamma(vec4 color) {
+        color.rgb = pow(color.rgb, vec3($gamma));
+        return color;
+    }
+"""
+
+_null_color_transform = 'vec4 pass(vec4 color) { return color; }'
+_c2l = 'float cmap(vec4 color) { return (color.r + color.g + color.b) / 3.; }'
+
+
+def _build_color_transform(data, clim, gamma, cmap):
+    if data.ndim == 2 or data.shape[2] == 1:
+        fclim = Function(_apply_clim_float)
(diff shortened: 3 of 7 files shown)


tlambert03 (approved):
yeah, great!  I like how you changed `_vispy/volume.py` when we delete the vendoring folder it will leave us even trimmer.  nice.  this all seems to work for me and I can't find anything wrong in the code.

Just want to make it very clear what's happening here @jni, @sofroniewn:
every time `volume.set_data` is called, a new 8bit texture is calculated and sent to the GPU.  changing contrast/gamma then just changes the lookup table that the GPU uses to convert values to intensity/color.  That means there is the potential for visual posterization if you change the contrast limits to a small fraction of what the texture was calculated for.  By default, this will _recalculate_ the texture (i.e. call `set_data`) whenever that range drops below 20% of what the original texture was calculated with (in other words, when there are only about 50 grayscale values being used for display.  At that point, if start you _increasing_ the contrast limits range you will see much more lag (as it used to be).  We can adjust that threshold with the `Volume.clim_range_threshold` parameter.

sofroniewn:
Thanks for the review @tlambert03! 

> At that point, if start you increasing the contrast limits range you will see much more lag (as it used to be). We can adjust that threshold with the Volume.clim_range_threshold parameter.

Yeah, I wonder @tlambert03 if we shouldn't just set this according to be `contrast_limits_range` and then update it whenever `contrast_limits_range` changes which won't be so often.

Does that make sense? If so do you want to quickly add that to this PR? I guess we don't need to - but might as well, no? I think that's how we can guarantee good performance when the clim is appropriate for the clim range.

jni:
@sofroniewn I dunno, I don't like that the image can end up looking really terrible. I think I prefer @tlambert03's heuristics.

@tlambert03 medium/long term, how hard would it be to allow 16bit textures or even 32-bit float?

jni:
@sofroniewn can you add a comment in the vendored code to indicate your modification? We should probably keep some kind of registry somehow that tells us what part of the code is not yet contributed upstream. In general, I prefer vendoring to be very temporary, rather than temporary forever. =P

sofroniewn:
> @sofroniewn can you add a comment in the vendored code to indicate your modification?

Done

> We should probably keep some kind of registry somehow that tells us what part of the code is not yet contributed upstream.

I think this file organization along with that comment makes it more clear than before.

> In general, I prefer vendoring to be very temporary, rather than temporary forever. =P

Agreed, we can try and work on this once the big PR from vispy gets merged. We'll get there.

> @sofroniewn I dunno, I don't like that the image can end up looking really terrible. I think I prefer @tlambert03's heuristics.

That's fine. I will leave as is.

> @tlambert03 medium/long term, how hard would it be to allow 16bit textures or even 32-bit float?

I think this is a good goal, but my big concern is that it might lead to more install/ hardware incompatibility problems. Maybe that's not true in 2020 anymore, but in 2016 I think that was a concern.

I will merge once tests pass unless @tlambert03 or others have any additional comments.

tlambert03:
> @tlambert03 medium/long term, how hard would it be to allow 16bit textures or even 32-bit float?

it's doable.  we'd need to fix [hard-coded 8bit texture sizes](https://github.com/vispy/vispy/blob/bb2cd96eb16dde1ecac3c70f9e572791bb5699e9/vispy/gloo/glir.py#L1635) in `vispy/gloo/glir.py`.  It's really just an issue of memory, so yes it will increase and we could probably catch errors with appropriate messages to "turn it back down to 8bit".

> if we shouldn't just set this according to be contrast_limits_range and then update it whenever contrast_limits_range changes which won't be so often.

yeah, if we pinned it to `contrast_limits_range`, then you'd end up with very posterized looking images every time someone has set their range too large, and then adjusted the slider to try to make the image look better.  (i.e. people will be surprised if the range of the slider is affecting their image display)

Issue #938, opened 2020-02-03
Log-scale slider


jni:
## 🚀 Feature

In tutorials about e.g. image segmentation with scikit-image, I often advise students to first tune parameters (such as compactness for watershed or SLIC, or alpha and beta for active contours) on a log scale, before refining further if at all needed. For many plugins, this would be the right type of slider for floating point arguments by default.

Therefore, it would probably be good if we provided such a slider for developers to use.

tlambert03:
In case it's useful to start with, I implemented something of a cheater log slider in the histogram widget here:
https://github.com/napari/napari/blob/0cbb95dedf71232a9841b12f125920496b0d56a7/napari/_qt/qt_histogram.py#L64-L75

would be great to have our own class though.

GenevieveBuckley:
Would something like this work? Based on the DoubleSlider class in magicgui:

`napari/qt/qt_log_slider.py`
```python
"""A Slider with a natural logarithmic scale."""
from qtpy.QtCore import Qt
from qtpy.QtWidgets import QSlider


class QLogSlider(QSlider):
    """A Slider Widget with a natural logarithmic scale."""

    PRECISION = 1000

    def __init__(self, parent=None):
        super().__init__(Qt.Horizontal, parent=parent)
        self.setMaximum(10)

    def value(self):
        """Get the natural log slider value as a float."""
        return np.log(super().value() / self.PRECISION)

    def setValue(self, value):
        """Set integer slier position from float ``value``."""
        super().setValue(np.exp(value) * self.PRECISION)

    def setMaximum(self, value):
        """Set maximum position of slider in float units."""
        super().setMaximum(np.exp(value) * self.PRECISION)
```

GenevieveBuckley:
We've made a few tweaks to the sketch above, and merged a QLogSlider class into the magicgui codebase. If we want this in napari as well, and figure out where it should live (where it should live is not clear to me), then it should be possible to reuse the code from here: https://github.com/napari/magicgui/pull/38/files

dstansby:
Can this be closed now there's a log slider available in magicgui? As far as I can tell, this isn't a proposal to *use* it anywhere in napari, just to have it available for plugin developers.

PR #1460, opened 2020-07-16
Reduce attenuation default and range

### LINKED ISSUE #1392: Default value for attenuation in attenuated-MIP rendering should be lower

sofroniewn:
# Description
Closes #1392 by reducing attenuation range from `(0, 2)` to `(0, 0.5)`. This is still bigger than a lot of values reported in #1392, which were `(0.02, 0.06)`, but one of the datasets I have still looks good out to `0.5`. @tlambert03 suggested something non-linear which we might want to try in the future, but for now I think this is an improvement on what we have, without being a major change.

## Type of change

### DIFF
--- a/napari/_qt/layers/qt_image_layer.py
+++ b/napari/_qt/layers/qt_image_layer.py
@@ -77,9 +77,9 @@ class QtImageControls(QtBaseImageControls):
         sld = QSlider(Qt.Horizontal, parent=self)
         sld.setFocusPolicy(Qt.NoFocus)
         sld.setMinimum(0)
-        sld.setMaximum(200)
+        sld.setMaximum(100)
         sld.setSingleStep(1)
-        sld.setValue(int(self.layer.attenuation * 100))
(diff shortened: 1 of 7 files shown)


jni:
@sofroniewn the tests are failing. Also, I would suggest making the default 0.05 rather than the middle of the range...?

sofroniewn:
> @sofroniewn the tests are failing. Also, I would suggest making the default 0.05 rather than the middle of the range...?

Fixed test fail, my bad - forgot to update something, and made default `0.05`, but kept range `(0, 0.5)`.

[Edit: I guess when we have log sliders - see #938 - we should make attenuation on a log scale]

jni (approved):

PR #1463, opened 2020-07-16
fix toggle ndisplay z-order points

### LINKED ISSUE #1462: 2d->3d->2d points z-ordering bug
## 🐛 Bug

2d->3d->2d z-ordering bug using `examples/add_points.py`

## To Reproduce

1. Run `examples/add_points.py`
2. Click 3D rendering button
3. Click 2D rendering button
4. Observe points that were in front of image are now behind image.

## Expected behavior

Points should retain correct z-order

## Environment

Last commit 0179c0dad1d93326388d3ab0f768e4ef028b3f09

## Additional context

Identified in https://github.com/napari/napari/pull/1374#issuecomment-659233743

sofroniewn:
# Description
Closes #1462 by fixing z-ordering change when toggling ndisplay for images (bug introduced in #1445) and for points where we were previously using the node create destroy approach we wanted to move away from in #1445.

I will add two screenshots test for this - but I'm hoping we might have fixed our windows CI bug too! We'll see!!

## Type of change

# References

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/_qt/_tests/test_qt_viewer.py
+++ b/napari/_qt/_tests/test_qt_viewer.py
@@ -158,6 +158,39 @@ def test_new_shapes_empty_viewer(make_test_viewer):
     assert np.sum(view.dims._displayed_sliders) == 0
 
 
+def test_z_order_adding_removing_images(make_test_viewer):
+    """Test z order is correct after adding/ removing images."""
+    data = np.ones((10, 10))
+
+    viewer = make_test_viewer()
+    vis = viewer.window.qt_viewer.layer_to_visual
+    viewer.add_image(data, colormap='red', name='red')
+    viewer.add_image(data, colormap='green', name='green')
+    viewer.add_image(data, colormap='blue', name='blue')
+    order = [vis[x].order for x in viewer.layers]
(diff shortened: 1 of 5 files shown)


sofroniewn:
🥳 I think out windows CI bug is finally fixed @napari/core-devs!!! Looks like it was the same vispy create/ destroy but it was happening on Points layer too!! This needs a couple screenshot tests, but then we might be golden!!

sofroniewn:
Fix for #1467 now included with screenshot test. I might also want to add a non-screenshot lest looking at the order parameter on the node as is mentioned in #1467 by @tlambert03, but that will have to come tomorrow!

jni (approved):
@sofroniewn so awesome!!! 🎉 

I'll approve but I'll let you self merge in case you want to add to this PR before merging.

sofroniewn:
I've now added one more test and fixed a merge conflict. Will merge once tests pass

sofroniewn:
Windows test still failed on CI :-( I'm merging anyway, as while I had hoped this PR would fix that, it looks like we still might have windows CI/ vispy issues .... i guess we'll have to keep looking into them.

PR #1468, opened 2020-07-16
Fix typos

camlloyd:
A small PR: The ones that got away from codespell.


sofroniewn (approved):
Love it, thanks so much for these fixes!! Really appreciated

PR #1469, opened 2020-07-17
DOC: minor doc reformatting.

Carreau:
Too short underline make numpydoc not see that as a header.

... as continuation prompt are necessay to non ambiguously detect conde
continuation.


sofroniewn (approved):
LGTM!

Issue #1436, opened 2020-07-12
Base EventedModel proposal


tlambert03:
## 🚀 Feature
I've been playing around with an idea for a generic model base class that would ultimately reduce a lot of our model code to something like this:

```python
class Layer(EventedModel):
    name: str
    x: int = 3
```
Where `EventedModel` provides the following features:

- type validation/coersion: on class instantiation _and_ on attribute assignment, the type will be validated and coerced if possible (this would also work for Enums and stuff).  ensuring that the model state is always valid, or an exception gets raised. 
- event emission after successful attribute assignment.
- Those two things cover the majority of our boilerplate model code, but it would still possible to override complex setter behavior, by adding an explicit setter method (called `_on_<attr_name>_set` in the example below)

Each characteristic is optional... the type checking and coercion could be provided by [pydantic](https://pydantic-docs.helpmanual.io/) (a very cool package I have used in other projects like [ome-types](https://github.com/tlambert03/ome-types)).  Custom type validation is possible (for instance, make sure it's a M x N array, etc...) and [this blog post](https://www.madelyneriksen.com/blog/validated-container-types-python-pydantic) provides a good overview on that.  However, the whole type checking is optional, and does add slight overhead.  The auto event emission and `dataclass` style syntax would be possible independently.

Here's a working prototype (need to `pip install pydantic`):

```python
from pydantic import BaseModel
from napari.utils.event import EmitterGroup


# BaseModel is a dataclass-like class from pydantic
# they also provide @dataclass, but don't yet know how to use it with additional methods
class EventedModel(BaseModel):
    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        self.events = EmitterGroup(
            source=self,
            auto_connect=False,
            **{name: None for name in self.__fields__},
        )

    class Config:
        extra = 'allow'
        validate_assignment = True  # validate types on assignment as well

    def __setattr__(self, name, value):
        super().__setattr__(name, value)  # set with type coercion
        value = getattr(self, name)  # get the validated value
        # if custom set method `_on_<name>_set` exists, call it
        setter_method = getattr(self, f'_on_{name}_set', None)  # 
        if callable(setter_method):
            # the method can return True, if it wants to handle its own events
            if setter_method(value):
                return
        # otherwise, we emit this event
        if name in self.events:
            getattr(self.events, name)(value=getattr(self, name))



class Layer(EventedModel):
    name: str
    x: int = 3


layer = Layer(name='hi')
layer.events.connect(
[...]

PR #1475, opened 2020-07-17
@dataclass decorator with events and properties

tlambert03:
# Description
continuing from #1436, This is a working example of an `@dataclass` decorator that extends the builtin dataclass decorator with two additional parameters: `events`, `properties`
- `@dataclass(events=True)`, will:
   - add an `EmitterGroup` named `events` to the decorated class that emits an event whenever one of the dataclass fields is changed.  (note, if an attribute is added after instantiation, it's not technically a "field" as per the [dataclass definition](https://docs.python.org/3/library/dataclasses.html#dataclasses.dataclass) and will not emit an event).
   - look for and execute any methods named `_on_<field_name>_set` following the pattern described in https://github.com/napari/napari/issues/1436#issuecomment-657272314
- `@dataclass(properties=True)` will:
   - turn all of the dataclass fields into property descriptors (as if you had used `@property`)
   - populate the `__doc__` of that property with the docstring for the corresponding parameter in the `Parameters` section of the class docstring (assuming it is a properly formatted numpydoc).  This solves the IPython `?` problem mentioned [here](https://github.com/napari/napari/issues/1436#issuecomment-659077954), (and also makes docs show on hover in a dynamic IDE like VScode).

- `properties` and `events` can be used together or individually... and if neither is used, then `@dataclass` is exactly like the [builtin dataclass](https://docs.python.org/3/library/dataclasses.html#dataclasses.dataclass) (which also has its own optional parameters).

see [`test_dataclass.py`](https://github.com/napari/napari/compare/master...tlambert03:feat/evented-dataclass?expand=1#diff-4edf07c4341b90a05e2c644668752606) for example.

### a note about python dataclasses

It's a bit hard to subclass python dataclasses, and this may make us want to return to the pure class inheritance model in https://github.com/napari/napari/issues/1436#issue-655404990 that doesn't actually use a python dataclass.  Namely: subclasses cannot have any non-default arguments if the superclass has a default argument.

```python
from dataclasses import dataclass

@dataclass
class A:
    x: int
    y: int = 3

# this is illegal because A has default argument (y)
@dataclass
class B(A):
    z: int

# this is ok, because `z` has a default value
@dataclass
class C(A):
    z: int = 4
```

This isn't hooked up to anything, but it'd be nice to get some feedback on the general pattern, and whether we actually want to use dataclasses, or just kinda steal the "syntax" of a dataclass declaration without actually using dataclasses.

## Type of change

# References

# How has this been tested?

## Final checklist:


tlambert03:
oh right... 3.6 is going to fail until I add the `dataclasses` backport (or we decide to drop 3.6).

tlambert03:
an example was requested:

```python
from napari.utils.dataclass import dataclass

@dataclass(events=True, properties=True)
class Model:
    """Just a test.

    Parameters
    ----------
    a : int
        Description of parameter `a`.
    b : str, optional
        Description of parameter `b`. by default 'some value'
    """

    a: int
    b: str = 'some value'
```
in usage:
```python
In [2]: model = Model(a=3)                                                                           

In [3]: model                                                                                        
Out[3]: Model(a=3, b='some value')

In [4]: model.a?                                                                                     
Type:        property
String form: <property object at 0x10ae6d9f0>
Docstring:   Description of parameter `a`.

In [5]: model.events.a.connect(lambda e: print(f"you changed a to {e.value}"))                       

In [6]: model.a = 7                                                                                  
you changed a to 7

In [7]: model                                                                                        
Out[7]: Model(a=7, b='some value')
```

sofroniewn:
I love how short the one file is!! I will look into the code in more detail but I love the direction and the `events` and `properties` options

> 3.6 is going to fail until I add the dataclasses backport

I'd add the backport at least until we are ready for the `0.4.0` napari release (which could be very soon, but would nice to be able to group things all together to make one breaking change).

> (note, if an attribute is added after instantiation, it's not technically a "field" as per the dataclass definition and will not emit an event)

> It's a bit hard to subclass python dataclasses, and this may make us want to return to the pure class inheritance model

I'm actually excited about these changes - I think this will force us away from inheritance towards simple composition. It will be a bit of a shock to the codebase to move away from `BaseLayer` and `ImageLayer` pattern, but I think we can get there gradually.

I'd probably replace our `LayerList`, our `Dims`, and our `ViewerModel` first as these don't use composition, then I'd replace `TextManager`, then I'd try adding a `ColorManager` and make things simpler and simpler before tackling the layers. Thoughts on that role out?

I will look into code now

napari/utils/dataclass.py
```diff
@@ -0,0 +1,156 @@
+import dataclasses as _dataclasses
+from typing import Any, Callable, Type, TypeVar
+
+import toolz as tz
+
+from .event import EmitterGroup
+
+WHEN_SET = "_on_{name}_set"
```
sofroniewn: I like this!

napari/utils/dataclass.py
```diff
@@ -0,0 +1,156 @@
...
+        If true, rich comparison dunder methods are added, by default False
+    unsafe_hash : bool, optional
+        If true, a __hash__() method function is added, by default False
+    frozen : bool, optional
+        If true, fields may not be assigned to after instance creation., by
+        default False
+    events : bool, optional
+        If true, an ``EmmitterGroup`` instance is added as attribute "events".
+        Events will be emitted each time one of the dataclass fields are
+        altered. by default False
```
sofroniewn:
```suggestion
        altered, by default False
```

napari/utils/dataclass.py
```diff
@@ -0,0 +1,156 @@
...
+    properties : bool, optional
+        If true, field attributes will be converted to property descriptors.
+        If the class has a class docstring in numpydocs format, docs for each
+        property will be taken from the ``Parameters`` section for the
+        corresponding parameter. by default False
```
sofroniewn:
```suggestion
        corresponding parameter, by default False
```

napari/utils/dataclass.py
```diff
@@ -0,0 +1,156 @@
...
+    repr : bool, optional
+        If true, a __repr__() method is added, by default True
+    eq : bool, optional
+        [description], by default True
+    order : bool, optional
+        If true, rich comparison dunder methods are added, by default False
+    unsafe_hash : bool, optional
+        If true, a __hash__() method function is added, by default False
+    frozen : bool, optional
+        If true, fields may not be assigned to after instance creation., by
```
sofroniewn:
```suggestion
        If true, fields may not be assigned to after instance creation, by
```

sofroniewn (approved):
LGTM! Will be very exciting to see this and #1444 interact.

Am I right to assume that now if I have an evented dataclass with an evented list then the events from the list will propogate up to the dataclass and so I only need to listen to the dataclass to hear events from the list?

tlambert03:
> Am I right to assume that now if I have an evented dataclass with an evented list then the events from the list will propogate up to the dataclass and so I only need to listen to the dataclass to hear events from the list?

no ... just the other way around (if an evented dataclass were a deeply nested member of a `NestableEventedList`, its events would bubble up to the root).  Since that concept makes the most sense with an explicit tree model (parents and children), and since we only have a single level "container" here, I think we should probably stick with explicit connections inside the dataclass for the short term.

tlambert03:
> LGTM! Will be very exciting to see this and #1444 interact.

cool, will flesh out the docstrings then

sofroniewn:
> no ... just the other way around (if an evented dataclass were a deeply nested member of a NestableEventedList, it's events would bubble up to the root). Since that concept makes the most sense with an explicit tree model (parents and children), and since we only have a single level "container" here, I think we should probably stick with explicit connections for the short term.

Intriguing, but makes sense. Agreed lets stick with explicit things for now. Maybe it's something that as we go through rolling this out on `Dims` we'll get more clarity on if it is even desirable etc.

tlambert03:
bringing the conversation started in https://github.com/napari/napari/pull/1485#discussion_r456944893 back to this thread, as it relates more to this PR.

I've now added support for type-coercion for `@dataclass(properties=True)` using this syntax:
```python
from typing_extensions import Annotated

@dataclass(properties=True)
class MyModel:
    attr: Annotated[Type, getter_converter, setter_converter] = default_value
    # where *_converter are functions that accept a value and return the coerced value

    # each is optional, but positional, so if you only want the getter you can do this:
    attr: Annotated[Type, getter_converter] = default_value
    # but if you only want the setter you have to do this:
    attr: Annotated[Type, None, setter_converter] = default_value
```

note... for now, this means that there might be some redundant syntax, for instance to coerce all strings to an enum on setting, you must supply all three arguments:
```python
@dataclass(properties=True)
class MyModel:
    anchor: Annotated[Anchor, None, Anchor] = Anchor.CENTER

    # optionally... to make sure that the user always gets it back in string form:
    anchor: Annotated[Anchor, str, Anchor] = Anchor.CENTER
```

tlambert03:
> ```python
>anchor: Annotated[Anchor, None, Anchor] = Anchor.CENTER
> ```

I added a Type alias `Property` that makes this a little clearer, and performs basic checking to make sure the getter/setter are actually callable or None.  This becomes:
```python
from napari.utils.dataclass import dataclass, Property

@dataclass(properties=True)
class MyModel:
    attr_name: Property[Type, getter, setter] = default_value
```

tlambert03:
I'm hitting a couple snags in the `Property[]` pattern when trying to hook up `Image`.  We have a few properties there (and probably elsewhere in our layers) that are really just kinda "facades"  that present a nice exterior API for a more complicated underlying property.  For instance, `colormap` actually sets both `_colormap_name` and `_cmap`, and  `interpolation` is actually a dict whose current key depends on `ndisplay`.  Both of these need `self` in the setter though, and are no longer simple type conversions/assertions.   I see a couple options:

1.  Do away with the `Property[]` type.  As nice as that felt when it was clean, we were starting to see a few cases where the effect was just to move the setter function into a new function above the class.  And also it doesn't get `self` ... an alternative pattern to what I proposed [here](https://github.com/napari/napari/pull/1475#issuecomment-661318172) could be a couple new (optional) method decorators that take the name of the field:

```python
@dataclass(properties=True)
class MyModel:
    attr_name: Type = default_value

    # both of these are optional
    @setter('attr_name')
    def _some_method(self, value):
        ...
    
    @getter('attr_name')
    def _some_other_method(self):
        ...
```

2. Revisit attributes like `colormap` and `interpolation` and change their internal representations.  Definitely seems like `colormap` could be changed, and `interpolation` could be made `interpolation_2d` and `interpolation_3d`.  Adds more work to the integration of the new dataclass pattern, but might leave us with leaner models on the other side.

@sofroniewn, thoughts?

sofroniewn:
> Revisit attributes like colormap and interpolation and change their internal representations. Definitely seems like colormap could be changed, and interpolation could be made interpolation_2d and interpolation_3d. Adds more work to the integration of the new dataclass pattern, but might leave us with leaner models on the other side.

I lean towards this option. `Colormap` and `Interpolation` were big challenges for the original EVH too and required a bunch of fiddling. I really see them as examples of suboptimal design and that the extra constraints of `Property[]` will actually make our code easier to understand and more maintainable rather than trying to allow these special cases. `name` is another one that may prove challenging due to its uniqueness constraint, but I think it will be worth persevering through too.

As to how to do address `interpolation` and `Colormap`. One option for interpolation is to harmonize the 3D modes and 2D modes - 3D is only `linear` and `nearest`, 2D has `nearest` and `bilinear` (which is same as `linear` in this context). We could just make those the only valid interpolation modes and when changing 2D/ 3D you just stay in which ever mode you were in, or we could continue to support more of the 2D modes, but when you change to 3D if in an unrecognized mode drop into `linear` by default (this may still be a little complex, but might be simpler than remembering a whole other value). Finally we could only support `linear` and `nearest` for now but look into adding 3D support for other key 2D modes.

As for colormap, I guess that would argue for having a dedicated colormap object with its own name option or always forcing it to be a tuple, without those additional `_colormap_name`, `_cmap` options.

How does that sound @tlambert03?

jni:
I agree with @sofroniewn's comment. I've been wanting to drop most of our 2D interpolation modes anyway. I think cubic should probably stay, as it's very common, but then yes, we should interpolate on nearest, linear, cubic, with cubic dropping to linear in 3D. And, as he says, this is all just a symptom that our models are too hairy. This is just blatantly obvious with colormaps and names, and again, is something that we've been wanting to resolve in many contexts.

tlambert03:
great, that's what I was hoping you would both say :)

sofroniewn:
> we should interpolate on nearest, linear, cubic, with cubic dropping to linear in 3D

I like those 3 options, we can even look into what it would take to do cubic for 3D, shouldn't be too hard.

> great, that's what I was hoping you would both say :)

:-)

napari/utils/events/dataclass.py
```diff
@@ -0,0 +1,421 @@
...
+    setattr(cls, '__setattr__', set_with_events)
+    setattr(cls, '__post_init__', evented_post_init)
+    return cls
+
+
+def getattr_with_conversion(self: C, name: str) -> Any:
+    """Modified __getattr__ method that allows class override.
+    Parameters
+    ----------
+    self : T
```
kne42: did you mean
```suggestion
    self: C
```
?

also is the function body meant to be empty?

napari/utils/events/dataclass.py
```diff
@@ -0,0 +1,421 @@
...
+class Property:
+    """Declare a dataclass field as a property with getter/setter functions"""
+
+    def __new__(cls, *args, **kwargs):
+        raise TypeError("Type Property cannot be instantiated")
+
+    def __init_subclass__(cls, *args, **kwargs):
+        raise TypeError(f"Cannot subclass {cls.__module__}.Property")
+
+    import typing_extensions
```
kne42: is there a reason for importing this despite there being `_te`? if so, could you comment why please?

kne42 (commented):
Very cool stuff @tlambert03!

I'm not sure if I understand 100% what is going on here, but there are a few things I'm concerned about: it may break the PyCharm IDE because they statically inspect code, but I suppose that is their problem and not ours. I also think it's kind of risky to be using so many private functions from other libraries as they may be subject to breaking changes.

sofroniewn:
hi @tlambert03, with #1652 we've officially dropped python3.6!! So this PR can move full steam ahead using dataclasses without needing to import the backport. Is there anything else you think this PR needs before merge @tlambert03? Will be so exciting to get this in!!

jni (approved):
Ok, I've given this another pass. There's still concerns about using private APIs from the standard library, and probably `dataclass` should be renamed so that it doesn't shadow the standard library one (even if it's a drop-in replacement), but these things can come in a later iteration. I'll make an issue.

I also addressed @kne42's well-noted comments. 😊

sofroniewn:
Ok all tests are passing and we have some clear post-merge plans, so I'm going to merge! Thank you @tlambert03 for this beautiful work, I can see it being used even far beyond napari!

PR #1476, opened 2020-07-17
Add global exception catching in Qt

liu-ziyang:
# Description
Add global exception catching, building on top work from https://github.com/napari/napari/pull/794

## Type of change
New feature (non-breaking change which adds functionality)

# References
closes #794

# How has this been tested?

## Final checklist:


sofroniewn (approved):
LGTM! But I'm wondering @ziyangczi @tlambert03 if this needs a test?

I was able to find this difference in tracebacks before/ after this PR if I introduced a bug into napari, so it does look like our handler is working, but still might be nice to have a small explicit test.

### before this PR

```python-traceback
  File "/Users/nsofroniew/GitHub/napari/napari/_qt/layers/qt_image_layer.py", line 166, in changeIsoThreshold
    self.layer.iso_threshold = value / 100
  File "/Users/nsofroniew/GitHub/napari/napari/layers/image/image.py", line 349, in iso_threshold
    w
NameError: name 'w' is not defined
```

### after this PR

```python-traceback
ERROR:root:Unhandled exception: Traceback (most recent call last):
  File "/Users/nsofroniew/GitHub/napari/napari/_qt/layers/qt_image_layer.py", line 166, in changeIsoThreshold
    self.layer.iso_threshold = value / 100
  File "/Users/nsofroniew/GitHub/napari/napari/layers/image/image.py", line 349, in iso_threshold
    w
NameError: name 'w' is not defined
```

tlambert03 (approved):
this looks good to me.  nice and minimal, and will immediately stop stuff from crashing to the command line.

> But I'm wondering @ziyangczi @tlambert03 if this needs a test?

everything needs a test! 😂
I added one.  In this case, it's nearly impossible to do something like "integration testing" by raising and error and seeing that it gets to our handler, because pytest itself has to hijack `sys.excepthook` to do its thing.  So we just have to trust python that uncaught exceptions will be given to `sys.excepthook` (it's a safe bet 😉 )... and then (for now) the test just needs to assert that a signal is emitted and the traceback is logged.  have a look [at the test](https://github.com/napari/napari/pull/1476/files#diff-2437b5e7f9e036b0328abcecaeff64f9R5-R15) if you're curious

tlambert03:
> @sofroniewn: I introduced a bug into napari

now why would you do that.

sofroniewn:
Test looks great @tlambert03, thanks for adding! Here's a like to the `caplog` pytest fixture docs in case people aren't familiar with it (I had to look it up) https://docs.pytest.org/en/stable/logging.html#caplog-fixture. I'll aim to merge this ~Monday after we have `0.3.6` out.

tlambert03:
closes #794

napari/_qt/event_loop.py
```diff
@@ -64,6 +66,11 @@ def gui_qt(*, startup_logo=False):
             app._splash_widget = splash_widget
     else:
         app._existed = True
+
+    # instantiate the execption handler
```
jni:
```suggestion
    # instantiate the exception handler
```

jni (approved):
LGTM, just left a minor typo fix!

jni:
@sofroniewn before merging, could you change the title to something a bit more descriptive? e.g. "Add global exception catching in Qt"

sofroniewn:
Typo fixed and title changed. I will merge once tests finish re-running

PR #1479, opened 2020-07-17
DOC: update param names to match function signature.

Carreau:
This attempt to match – in some non ambiguous case parameter names in
the docstring of functions to actually match the name in the signature.

This so far only works when there is a single misspelling and all the
other parameters are documented.

This also try respect a couple of other things with heuristics, to match
numpydocs:
  - Parameters with no type given should not have a ` : `
  - if there is a `:` without spaces assume it is still the type
  separator.

### DIFF
--- a/napari/_qt/qt_viewer_buttons.py
+++ b/napari/_qt/qt_viewer_buttons.py
@@ -244,7 +244,7 @@ class QtGridViewButton(QCheckBox):
 
         Parameters
         ----------
-        event : qtpy.QtCore.QEvent
+        state : qtpy.QtCore.QEvent
             Event from the Qt context.
         """
         if state == Qt.Checked:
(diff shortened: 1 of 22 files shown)


tlambert03:
i love this one.  thanks!

tlambert03 (approved):

napari/_qt/qt_viewer_buttons.py
```diff
@@ -244,7 +244,7 @@ def change_grid(self, state):
 
         Parameters
         ----------
-        event : qtpy.QtCore.QEvent
+        state : qtpy.QtCore.QEvent
             Event from the Qt context.
```
sofroniewn:
```suggestion
        state : qtpy.QtCore.Qt.CheckState
            State of the checkbox.
```

sofroniewn (approved):
LGTM!

I fixed the actual doc string for one where we had the wrong thing there

PR #1481, opened 2020-07-18
Fix events docstring type

tlambert03:
# Description
small docs change to correct the event type in many of our callbacks that were incorrectly labeled as QEvents instead of our native/vispy `event.Event`.  As a quick heuristic, camel case methods in `QWidget` subclasses that have the word "event" in them (like `mouseMoveEvent()`) are likely reimplemented methods that _do_ take a `QEvent`... everything else is likely our own event object.

## Type of change

- [x] docs update


sofroniewn (approved):
LGTM! Good catch, that is correct!!

PR #1485, opened 2020-07-19
Use evented dataclass for TextManager

tlambert03:
# Description
@sofroniewn [suggested](https://github.com/napari/napari/pull/1475#issuecomment-660506800) seeing how the evented dataclass from #1475  looks on the `TextManager` from #1374, and it was actually very encouraging!  I was nearly able to delete the entire class, without touching the tests! :) including all the `__eq__` and `_get_state`  methods which are nicely provided by [`dataclass.asdict`](https://docs.python.org/3/library/dataclasses.html#dataclasses.asdict). (most of the added stuff in this PR is just the unmerged dataclass file from #1475)

The main methods left over were cases where I needed to coerce the value type to something different on access (like converting an Enum back to a string), or on set (like making sure something is an ndarray).  It's nice having the `_on_name_get/set` fallback, but for setting things like Enums, I was even able to do the coercion in the dataclass logic using the type annotation, and that pattern could probably be reused for other simple types.  For declaring the "type on get", I can imagine using `Annotated` types, like:

```python
# coerces to Anchor when set, and coerces to str on get
anchor: Annotated[Anchor, str] = Anchor.CENTER
```

## Type of change

# References

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/layers/utils/text.py
+++ b/napari/layers/utils/text.py
@@ -1,6 +1,4 @@
 from typing import Tuple, Union, Dict, ClassVar, Optional
-import warnings
-
 import numpy as np
 
 from ..base._base_constants import Blending
@@ -10,9 +8,28 @@ from ._text_utils import (
     get_text_anchors,
 )
 from ...utils.colormaps.standardize_color import transform_color
+from ...utils.colormaps.vendored.colors import to_hex
 from ...utils.dataclass import dataclass
-from dataclasses import field
+from dataclasses import InitVar
 from typing_extensions import Annotated
+
+
+# https://github.com/napari/napari/issues/1491
+def transform_1_color(color):
+    return transform_color(color)[0]
+
+
+def no_opaque(blending):
+    val = Blending(blending)
+    if val == Blending.OPAQUE:
+        import warnings
+
+        warnings.warn(
+            'opaque blending mode is not allowed for text. '
+            'setting to translucent.'
+        )
+        return Blending.TRANSLUCENT
+    return val
 
 
 @dataclass(events=True, properties=True)
@@ -46,71 +63,45 @@ class TextManager:
         Set to true of the text should be displayed.
     """
 
-    text: Union[np.ndarray, str]
-    n_text: int
-    properties: Dict[str, np.ndarray] = field(default_factory=dict)
+    text: InitVar[Union[np.ndarray, str]]
+    n_text: InitVar[int]
+    properties: InitVar[Dict[str, np.ndarray]] = None
+
     rotation: float = 0.0
-    translation: np.ndarray = np.asarray(0)
+    translation: Annotated[np.ndarray, None, np.asarray] = np.asarray(0)
     anchor: Annotated[Anchor, str, Anchor] = Anchor.CENTER
-    color: Union[np.ndarray, str] = 'cyan'
+    color: Annotated[
+        Union[np.ndarray, str], to_hex, transform_1_color
+    ] = 'cyan'
     size: float = 12.0
-    blending: Annotated[Blending, str, Blending] = Blending.TRANSLUCENT
+    blending: Annotated[Blending, str, no_opaque] = Blending.TRANSLUCENT
     visible: bool = True
+    values: Optional[np.ndarray] = None
+    mode: TextMode = TextMode.NONE
 
     # private attributes not in dataclass
     _text_format_string: ClassVar[str] = ''
-    _values: ClassVar[Optional[np.ndarray]] = None
-    _mode: ClassVar[TextMode] = TextMode.NONE
 
-    def __post_init__(self):
+    def __post_init__(self, text, n_text, properties):
+        self.events.add(text=None)
         # called after __init__
-        self._set_text(self.text, self.n_text, self.properties)
-
-    @property
-    def values(self):
-        """np.ndarray: the text values to be displayed"""
-        return self._values
-
-    @property
-    def mode(self) -> str:
-        """str: The current text setting mode."""
-        return str(self._mode)
-
-    # Things that need to modify the value on set
-
-    def _on_translation_set(self, translation):
-        if not isinstance(translation, np.ndarray):
-            self.translation = np.asarray(translation)
-            return True
-
-    def _on_color_set(self, color):
-        if color != self.color:
-            self.color = transform_color(color)[0]
-        return True
-
-    def _on_blending_set(self, blending):
-        if Blending(blending) == Blending.OPAQUE:
-            warnings.warn(
-                'opaque blending mode is not allowed for text. '
-                'setting to translucent.'
-            )
-            self.blending = Blending.TRANSLUCENT
-            return True
+        self._set_text(text, n_text, properties or {})
 
     def _set_text(
         self, text: Union[None, str], n_text: int, properties: dict = {}
     ):
         if len(properties) == 0 or n_text == 0 or text is None:
-            self._mode = TextMode.NONE
+            self.mode = TextMode.NONE
             self._text_format_string = ''
-            self._values = None
+            self.values = None
         else:
-            formatted_text, text_mode = format_text_properties(
+            self.values, self.mode = format_text_properties(
                 text, n_text, properties
             )
             self._text_format_string = text
-            self._values = formatted_text
-            self._mode = text_mode
+        # `values` and `text` is slightly duplicated.  If this were changed to
+        # "self.events.values", then we wouldn't have to add a dedicated event
+        # (since self.text isn't actually used except for initialization)
         self.events.text()
 
     def refresh_text(self, properties: dict):
--- a/napari/layers/utils/text.py
+++ b/napari/layers/utils/text.py
@@ -1,30 +1,32 @@
-from typing import Tuple, Union
 import warnings
+from dataclasses import field
+from typing import Dict, Tuple, Union
 
 import numpy as np
 
-from ..base._base_constants import Blending
-from ._text_constants import TextMode, Anchor
-from ._text_utils import (
-    format_text_properties,
-    get_text_anchors,
-)
 from ...utils.colormaps.standardize_color import transform_color
-from ...utils.event import EmitterGroup, Event
+from ...utils.dataclass import dataclass
+from ..base._base_constants import Blending
+from ._text_constants import Anchor, TextMode
+from ._text_utils import format_text_properties, get_text_anchors
 
 
+@dataclass(events=True, properties=True)
 class TextManager:
     """Manages properties related to text displayed in conjunction with the layer.
 
     Parameters
     ----------
     text : array or str
-        the strings to be displayed
+        The strings to be displayed
+    n_text : int
+        The length of data being annotated.
     rotation : float
         Angle of the text elements around the anchor point. Default value is 0.
     anchor : str
         The location of the text origin relative to the bounding box.
-        Should be 'center', 'upper_left', 'upper_right', 'lower_left', or 'lower_right'
+        Should be 'center', 'upper_left', 'upper_right', 'lower_left', or
+        'lower_right'
     translation : np.ndarray
         Offset from the anchor point.
     color : array or str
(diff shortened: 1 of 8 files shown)


tlambert03:
I was able to get this pattern to work... but I won't push it yet so you can have a look at the PR without it.

```python
# coerces to Anchor when set (internal representation), and coerces to str on get
anchor: Annotated[Anchor, str] = Anchor.CENTER
```

napari/layers/utils/text.py
```diff
@@ -39,74 +43,68 @@ class TextManager:
...
+
+    # Things that need to modify the value on set
+
+    def _on_translation_set(self, translation):
+        if not isinstance(translation, np.ndarray):
+            self.translation = np.asarray(translation)
+            return True
+
+    def _on_color_set(self, color):
+        if color != self.color:
```
sofroniewn: why have the `!=` catch here? seems a bit of "one off" to have it and unlike the others

napari/layers/utils/text.py
```diff
@@ -39,74 +43,68 @@ class TextManager:
...
+            self.color = transform_color(color)[0]
+        return True
+
+    def _on_blending_set(self, blending):
+        if Blending(blending) == Blending.OPAQUE:
+            warnings.warn(
+                'opaque blending mode is not allowed for text. '
+                'setting to translucent.'
+            )
+            self.blending = Blending.TRANSLUCENT
```
sofroniewn: here we could just ignore this and leave value unchanged rather that changing to translucent (i know that this matches original behviour, but just noting)

sofroniewn (commented):
Looking very nice @tlambert03 - so cool to see so much code disappear.

i do wonder though if we can do more than the custom `_on_*_set` approach. It looks like in most cases we'll want to do one of two things

1. transform the input value - i.e. `self.color = transform_color(color)[0]` or `self.translation = np.asarray(translation)` If we knew that was all we wanted to do then maybe you could just register that transformation somehow?

2. Block certain values see `if Blending(blending) == Blending.OPAQUE:` there could also maybe be a simpler syntax for this.

I think it's fine to still leave in the `_on_*_set` for more complex cases, but might be nice to accomplish the first one with simpler - less error prone syntax. (similarly for the `_on_*_get` method, where it looks like we just need a simple transformation, i.e. `str(anchor)`)

Apologies if this has been noted elsewhere in the comments and I havn't seen it yet

kevinyamauchi:
Yo! I haven't had a chance to look at it too closely, but what I see is awesome! I will take a closer look tomorrow.

napari/layers/utils/text.py
tlambert03: That's to avoid a `RecursionError` (comments [here](https://github.com/napari/napari/blob/0387244cea007d83aa1aa733dafe6f72f25e5fba/napari/utils/dataclass.py#L58-L61)).  There are other patterns we could employ, but since we have a setter calling another setter, one has to be careful.

sofroniewn: hmm yeah, all the more reason I'd prefer a different way to pass a single transformation function that get's called on setting for these simple cases.

Originally I thought that would block all event emission, but i now realize with the `super` call it wouldn't, it just prevents the recursion, so it makes sense, but requires some thought

tlambert03: Thats what https://github.com/napari/napari/pull/1485#issuecomment-660565175  is proposing... (note, those types _are_ functions). Using color as an example, here are two I could support, let me know what you like:

1.  `Annotated` type ([PEP 593](https://www.python.org/dev/peps/pep-0593/)), that takes (up to) 3 arguments, with the first being what type(s) are accepted from the user, the second being a function that is called on the value when "getting" the value (can also be a callable `type`), and the third is the function that is called when "setting" the value.

```python
from typing_extensions import Annotated

# syntax = Annotated[type, get_func, set_func (if different than type)]

# color that may be either ndarray or string, and calls transform_color on set
color: Annotated[Union[np.ndarray, str], None, transform_color] = 'cyan'

# internal enum that "gets" as a string
anchor: Annotated[Anchor, str] = 'cyan'
```

2.  using `metadata` on [dataclass `fields`](https://docs.python.org/3/library/dataclasses.html#dataclasses.field)
```python
from dataclasses import field

# where both get_with & set_with are optional. 
color: Union[np.ndarray, str] = field(default='cyan, metadata={'set_with': transform_color})

# example for an internal enum that "gets" as a string
anchor: Anchor = field(default=Anchor.CENTER, metadata={'get_with': str})
```

tlambert03: @jni, if you're holding off on looking at this whole thing, i'd be curious to get your opinion on this one comment... (how to indicate some simple getter/setter transformation that used to live in @properties/setters)

tlambert03: this is also all basically a "light" version of what we'd get with [pydantic](https://pydantic-docs.helpmanual.io/), (see https://github.com/napari/napari/issues/1436) which has a whole framework for expressing validators and type converters... so, with pydantic I _think_ it would just be:
```python
color: ColorType = 'cyan'
# where somewhere else you've told pydantic how to validate & coerce that type 
```
and pydantic guarantees that your class always has a valid `ColorType` (or an exception if it cannot convert)

sofroniewn: Yes, I was wondering if https://github.com/napari/napari/pull/1485#issuecomment-660565175 might be the answer!!

I really like the Annotated Type approach - great that that is a standard now, it's really I think what we want, and I imagine we could get most functionality we need with that triple.

I like the
```python
color: Annotated[Union[np.ndarray, str], None, transform_color] = 'cyan'
anchor: Annotated[Anchor, str] = 'cyan'
```

Syntax, I find that easy to read and understand. I prefer it to the `field`  syntax.

I guess the `ColorType` syntax is quite slick and is nice because we could reuse that in many places, but I still like the idea of growing into that slowly.

Not sure what your plan for this PR next is, but I'd be down for a version using the `Annotated` syntax, and then maybe all our custom get/ set would go away (I think we could even have a coerce translucent function for the blending case)!!!

tlambert03: > Syntax, I find that easy to read and understand. I prefer it to the field syntax.

yeah, I also prefer it to the field syntax, except perhaps that i liked the `get_with`/`set_with` in the `field()` option... but we could do that with annotated too.  Which do you prefer?

```python
color: Annotated[ColorType, None, transform_color] = 'cyan'
# or 
color: Annotated[ColorType, {'set_with': transform_color}] = 'cyan'
```

> I guess the ColorType syntax is quite slick

we could probably do something like that internally without `pydantic` with a little type->converter `dict`.  I'll play with that.

> Not sure what your plan for this PR next is, but I'd be down for a version using the `Annotated` syntax

I'll go with that (but easy to switch if @jni hates it).  I think we just leave this PR open for a while, and I'll also make one for `ViewerModel` and stuff... until we get something that feels right for them all

sofroniewn: I guess I still lean to `color: Annotated[ColorType, None, transform_color] = 'cyan'`. It's slightly less explicit, but I think it will be such a major pattern we use as long as it is clearly documented somewhere then I think it's ok and it is cleaner to read.

> we could probably do something like that internally without pydantic with a little type->converter dict. I'll play with that.

yeah, but i'm also ok with us eventually going down the pydantic road too, so no need to spend too long.

>  I think we just leave this PR open for a while, and I'll also make one for ViewerModel and stuff... until we get something that feels right for them all

Ok sounds great. I'd like to do `Dims` just to get the experience working with the pattern too. Making the layer list version of the EVH really helped me understand it. I could do now, but might take both #1444 and #1475 to do it.

tlambert03: > I'd like to do Dims just to get the experience working with the pattern too.

That would be awesome!  I'm keeping #1475 up to date as make changes here... and I think #1444 is mostly done/static at this point.  So I have no problem just making a new PR for each new object we're porting over, merging in #1475 and/or #1444, and then just letting those PRs sit there for a bit.  We can play with the pattern without worrying about having to revert.  If you hit a roadblock in `Dims`, just let me know and we'll decide if the dataclass needs to change, or if we just workaround it in `Dims`

napari/layers/utils/text.py
kevinyamauchi: I'd be okay with this. I think it would be nice to keep the warning so the user knows that the opaque blending is working (it's just ugly 😛 )

napari/layers/utils/text.py
```diff
@@ -123,109 +113,6 @@ def _set_text(
             self._mode = text_mode
         self.events.text()
 
-    @property
```
kevinyamauchi, line 126: I love seeing this all red!

napari/layers/utils/text.py
```diff
@@ -39,74 +44,59 @@ class TextManager:
...
-        self._anchor = Anchor(anchor)
-        self._translation = translation
-        self._color = transform_color(color)[0]
-        self._size = size
-        self._blending = self._check_blending_mode(blending)
-        self._visible = visible
-
-        self._set_text(text, n_text, properties)
-        self.events.unblock_all()
+    text: Union[np.ndarray, str]
```
kevinyamauchi: I am sorry if I missed this in some of the previous threads, but are the events associated with these properties emitted when they are set in the `__init__()` method?

tlambert03: no apology necessary... all of this needs some nice documentation to go with it to introduce the new patterns without needing to go through all of the threads!
Nope, events will not be emitted during init. (but, now that you mention it, I don't think I have an explicit test for that... so will add)

tlambert03: actually... is that the behavior you were hoping for?  I noticed that you had that event blocker in the `__init__` before ...

napari/layers/utils/text.py
```diff
@@ -39,193 +61,49 @@ class TextManager:
...
-        return self._values
+    text: InitVar[Union[np.ndarray, str]]
+    n_text: InitVar[int]
+    properties: InitVar[Dict[str, np.ndarray]] = None
+
+    rotation: float = 0.0
+    translation: Annotated[np.ndarray, None, np.asarray] = np.asarray(0)
+    anchor: Annotated[Anchor, str, Anchor] = Anchor.CENTER
+    color: Annotated[
+        Union[np.ndarray, str], to_hex, transform_1_color
```
tlambert03: see #1491... this is a good example of needing a better general representation of colors.  The difficulty in evaluating the equivalence of `np.ndarray` requires us to change the `__eq__` logic of the dataclass, (or coerce the value to something like `to_hex` when getting it, as done here).  It should be much easier to check for color equivalence.

napari/layers/utils/text.py
tlambert03: I made it the [property setter](https://github.com/napari/napari/blob/32f53357c5d5a9ffce2d57545da9e0f62f680e2c/napari/layers/utils/text.py#L21-L31)

napari/layers/utils/text.py
kevinyamauchi: > actually... is that the behavior you were hoping for? I noticed that you had that event blocker in the __init__ before ...

I think that most of the time, we don't want to emit the property change events during the `__init__`since those are often connected to some sort of refresh. I think if there are ever cases where that's required, we can add it explicitly in the `__post_init__` right?

kevinyamauchi:
@tlambert03 , I have some time to work on converting `TextManager` into a dataclass this week. Do you mind if I take a go at finishing it up? I would start a new PR, since this contains changes that I think are no longer necessary.

tlambert03:
Absolutely, go for it.  New PR is great.  Lemme know if I can help

kevinyamauchi:
> Absolutely, go for it.  New PR is great.  Lemme know if I can help

Sounds good. Will do!

tlambert03:
can we close this PR @sofroniewn, @kevinyamauchi ?

kevinyamauchi:
Yes, go for it! I'm (slowly) working on it in #2047 .

PR #1486, opened 2020-07-20
Use Evented List in Dims

sofroniewn:
# Description
This PR builds on #1444 to use an evented list for a couple properties in the dims model and allows us to deprecate the `set_point` and `set_range` methods.

I also drop the `dims.mode` and `dims.interval` which are not hooked up and used - they were never implemented but merely placeholders.

We could probably use the evented list for more properties on dims like `axis_labels` and `order`, but curious how this looks to @tlambert03 before going further. Note I havn't tried any of the dataclass stuff like in #1475.

Note some of the changes in the diff are simply from #1444 which hasn't merged yet

## Type of change

- [x] Refactor

## Final checklist:


napari/components/dims.py
```diff
@@ -201,14 +166,12 @@ def ndim(self, ndim):
...
-            # if in interval mode
-            self._interval = [(0, 1)] * (ndim - cur_ndim) + self._interval
-            self._mode = [DimsMode.POINT] * (ndim - cur_ndim) + self._mode
+            for i in range(ndim - cur_ndim):
+                # Range value is (min, max, step) for the entire slider
+                self.range.insert(0, (0, 2, 1))
+                # Point is the slider value
+                self.point.insert(0, 0)
+
             self._order = list(range(ndim - cur_ndim)) + [
```
sofroniewn: Have to think about how to do things like this with the evented list - i probably want to modify in place rather than create/destroy like with the `+` syntax, but thanks for adding to #1444

napari/components/dims.py
```diff
@@ -219,25 +182,18 @@ def ndim(self, ndim):
                 self._axis_labels = (
```
sofroniewn: Have to think about how to do things like this with the evented list - i probably want to modify in place rather than create/destroy

napari/components/dims.py
tlambert03: do what exactly? prepend a list to a list?  There needn't be anything different about the evented list from a "regular" list (though we may need to update a couple magic methods).  What syntax do you want to use?  I think you have three options:
1. use list slice insertions syntax: `ev_list[:0] = [x,x,x,...]`  (currently, this will emit one changed even but we could make it emit as many as there are insertions)
2. use for loop with `ev_list.insert(0, ...)` (this will emit one `changed` event for each insertion)
3. we _could_ modify the `__radd__` method on the evented list to prepend when you do `x = [] + ev_list`, but that would be breaking standard python practice (i.e. that always returns a new list object)

napari/components/dims.py
tlambert03: yeah, same as above right? you're right that creating a new object will break your connections, so either use slice insertion, or for-loop insertion

napari/components/dims.py
sofroniewn: ooo `use list slice insertions syntax` this is quite nice, otherwise I was just going to do a for loop. I probably would have expected the slice insertion to emit as many events as insertions, but I guess I expect the `clear` to probably just emit one, so maybe it's ok.

main point was to note not to create/ destroy as I had been doing. I would not modify `__radd__`

tlambert03: if you find that you _need_ it to emit that many events, it would be easy... we'd just need to check whether `value` was iterable and iterate through it instead of passing on the slice syntax to the private list.
https://github.com/napari/napari/blob/7bac12ddb0e26d54006d686c599152162de242e5/napari/utils/events/containers/_list.py#L131-L135

> but I guess I expect the clear to probably just emit one, so maybe it's ok.

for synchronization with `QAbstracItemModels`, it ends up being easier to have there be a 1:1 correspondence between changing items and events... but it's not strictly necessary.  so if you run into any issue with clear emitting an event for each item being removed from the list, let me know.

sofroniewn:
Just refactored `axis_labels` to be an `EventedList`, see changes here https://github.com/napari/napari/pull/1486/commits/11ff3096dbfc2988e9bc89f4e7b555e5db76166c

I'd probably still like to refactor `order` to be an evented list, though order is an interesting one as it's always a permutation, not just any type of list. Curious if @jni or @tlambert03 have any ideas there.

Note the `set_*` methods that are still left but now have deprecation warnings can be removed in the future. I'll improve the warnings before this is ready for final review.

sofroniewn:
Ok, i've cleaned this up again and am quite happy with it now. It deprecates the `set_point`, `set_range`, and `set_axis_labels` methods as they are no longer needed. I still have the `setter` for the `axis_labels`, but after talking with @tlambert03 I can maybe drop as `dims.axis_labels[:] = my_list` would now achieve the same thing as `dims.axis_labels = my_list`.

I made `dims.order` into a tuple and not a list, as that is better thought of as a tuple, and so it did not become an eventedlist.

The next thing for this work is to try and use #1475 and the dataclass, but I am going to start that in a fresh PR.

I havn't cleaned up doc strings or added any new tests, old ones pass fine. @tlambert03 I'd say this is now ready for a "soft review". Curious if you identify any antipatterns / missed opportunities

sofroniewn:
Although #1919 doesn't use an `EventedList` for the Dims properties yet, that PR will supersede this one. We can alway pull out elements of this PR if we decide to add that in the future

PR #1487, opened 2020-07-20
Prevent napari crashing after layer is dragged into trash

### LINKED ISSUE #1484: When using PyQt5 backend on Windows, dragging layer to trash crashes napari
## 🐛 Bug

When dragging a layer to the trash (to delete it), napari crashes. This happens for me on Windows when using `PyQt5`, but does not crash when using `PySide2`. 

I've been meaning to report this for a bit, but thought it was a general Windows trouble until I downloaded the `briefcase` install which uses `PySide2`. I've tested this behavior by installing in a new virtual environment. This used to behave properly for me in the past, but I'm not sure if `napari` changed, or I reinstalled and ended up with `PyQt5` instead. 

In both cases, I get the same traceback (below), but `napari` does not crash if using `PySide2`.

## To Reproduce

1. Launch `napari` from command line
2. Open image (dragged image file onto `napari`)
3. Drag image to trash
4. Traceback printed
5. `napari` crashes

This also occurs if there any number of layers. I'm not just deleting the sole layer.

## Expected behavior

Layer is deleted, and `napari` doesn't crash. As if the layer were selected, and the trash can icon is clicked.

### No Crash:

```
napari: 0.3.6.dev35+ga119296
Python: 3.8.3 (tags/v3.8.3:6f8c832, May 13 2020, 22:37:02) [MSC v.1924 64 bit (AMD64)]
Qt: 5.14.2
PySide2: 5.14.2.3
NumPy: 1.19.0
SciPy: 1.5.1
Dask: 2.21.0
VisPy: 0.6.4

GL version: 4.6.0 NVIDIA 442.70
MAX_TEXTURE_SIZE: 32768

Plugins:
- napari-plugin-engine: 0.1.6
- svg: 0.1.3
```

### Crash

```
napari: 0.3.6.dev35+ga119296
Python: 3.8.3 (tags/v3.8.3:6f8c832, May 13 2020, 22:37:02) [MSC v.1924 64 bit (AMD64)]
Qt: 5.14.2
PyQt5: 5.14.2
NumPy: 1.19.0
SciPy: 1.5.1
Dask: 2.20.0
VisPy: 0.6.4

GL version: 4.6.0 NVIDIA 442.70
MAX_TEXTURE_SIZE: 32768

Plugins:
- napari-plugin-engine: 0.1.6
- svg: 0.1.3
```

## Additional context

Traceback in both cases:
```python
  File "c:\users\hectormz\code\contribute\napari\napari\_qt\qt_layerlist.py", line 307, in mouseMoveEvent
    index = self.layers.index(self._drag_name)
  File "c:\users\hectormz\code\contribute\napari\napari\utils\list\_typed.py", line 82, in index
KeyError: 'could not find element stack was referencing'
```

hectormz:
# Description

This PR fixes behavior where `napari` crashes when a layer is dragged to the trash, while using `PyQt5` backend.
Now we check if the selected layer has been deleted after being dragged, before trying to index it.

## Type of change

- [X] Bug-fix (non-breaking change which fixes an issue)

# References

This closes #1484 

# How has this been tested?

This bug fix does not have a test, in part because this file will not exist for much longer.

## Final checklist:


jni (approved):
LGTM! It would be awesome if this could be tested, but I don't know how to tests drags... @sofroniewn @tlambert03 do either of you know?

tlambert03:
normally one would probably unit-test this by mocking the [`QMimeData`](https://doc.qt.io/qt-5/qmimedata.html) object, which _should_ hold all of the information about the object being dropped.  But the current class here stores some of the drag information on the class itself as well (`self._drag_name`) so probably not going to work here.  I would try to test this as  directly as possible by creating a `QtLayerlist`, and calling its `mouseMoveEvent` method with a madeup [`QMouseEvent` ](https://doc.qt.io/qt-5/qmouseevent.html), that has an appropriate `.pos()`.

but again, don't worry about it too much if it's taking a lot of time.  The LayerGroup branch will have a new `QMimeData` object that will change what's happening here

sofroniewn:
You could also just set the `self._drag_name` and then call `mouseMoveEvent` and not worry about the `pos`, but I agree with @tlambert03 this is going to change soon anyway, even my `QLayerList` PR changed this logic, and we don't have any full drag/ drop tests with Qt like this either - not sure if we could use the `Qtbot` to simulate clicking/ dragging either.

As @tlambert03 said, let us know how it goes @hectormz as this could be a huge time sync that isn't worth it, but would be great if you got something we could keep using during the refactors.

hectormz:
Okay, got it. I'll look into this in the next day or two and report back if it seems like it would be too much time to get working

hectormz:
I haven't had a lot of time to look into this and figure the test out, but I did confirm that this is a PyQT + Windows behavior.

sofroniewn:
I have just updated from main, will merge once all tests pass. I think it is fine that this PR doesn't have a test for drag and drop given what @tlambert03 notes about the upcoming refactors. Thanks @hectormz for fixing this!!

sofroniewn (approved):
LGTM!

Issue #1369, opened 2020-06-17
Dimension slider wrong round rule.


Czaki:
## 🐛 Bug

Napari wrong rounds on dimension slider. When image layer has scale above 1 on dimesion with toggle slider.  

## To Reproduce

1. Run napari with this code:
```python
import napari
import numpy as np

with napari.gui_qt():
    scale = (1, 209.9999999999999, 77.52248561765859, 77.52248561765859)
    viewer = napari.Viewer()
    image = np.random.random((1, 2, 100, 100))
    viewer.add_image(image, scale=(1, 100, 1, 1))
```
2. Click "New points layer"
3. Move dimension slider right (notice thar image change on 1/4 of slider, not 1/2). 

```
WARNING: Traceback (most recent call last):
  File "/home/czaki/Projekty/napari/napari/_qt/qt_dims_slider.py", line 135, in <lambda>
    lambda value: self.dims.set_point(self.axis, value)
  File "/home/czaki/Projekty/napari/napari/components/dims.py", line 359, in set_point
    self.events.axis(axis=axis, value=value)
  File "/home/czaki/Projekty/napari/napari/utils/event.py", line 508, in __call__
    self._invoke_callback(cb, event)
  File "/home/czaki/Projekty/napari/napari/utils/event.py", line 529, in _invoke_callback
    cb_event=(cb, event),
  File "/home/czaki/Projekty/napari/napari/utils/event.py", line 523, in _invoke_callback
    cb(event)
  File "/home/czaki/Projekty/napari/napari/components/viewer_model.py", line 341, in _update_layers
    layer.dims.set_point(axis, point)
  File "/home/czaki/Projekty/napari/napari/components/dims.py", line 359, in set_point
    self.events.axis(axis=axis, value=value)
  File "/home/czaki/Projekty/napari/napari/utils/event.py", line 508, in __call__
    self._invoke_callback(cb, event)
  File "/home/czaki/Projekty/napari/napari/utils/event.py", line 529, in _invoke_callback
    cb_event=(cb, event),
  File "/home/czaki/Projekty/napari/napari/utils/event.py", line 523, in _invoke_callback
    cb(event)
  File "/home/czaki/Projekty/napari/napari/layers/base/base.py", line 633, in refresh
    self.set_view_slice()
  File "/home/czaki/Projekty/napari/napari/layers/base/base.py", line 586, in set_view_slice
    self._set_view_slice()
[...]

Issue #1384, opened 2020-06-25
Create a napari-tifffile plugin


jni:
## 🚀 Feature

For a brief, glorious moment, there was a napari IO plugin baked into tifffile. It has since been pulled out. It would be nice to make an independent plugin that would read tiff metadata into napari.

The code for the pulled out implementation is available here:

https://github.com/cgohlke/tifffile/commit/3998ad576cfbb05c1d5a2a4c1cd1d77536669c74?diff=unified

Hooray for git!

GenevieveBuckley:
Wonder why that glorious moment was so brief? I've made a start [here](https://github.com/GenevieveBuckley/napari-tifffile), still need to add tests and a few other things...

jni:
> Wonder why that glorious moment was so brief?

Well, it turned out to break our tests, which was fun, and I expect Christoph did not want to take on the maintenance burden. Very exciting that you've started it!

GenevieveBuckley:
Wait, how did it break *your* tests?

jni:
Because we autodiscover plugins and we depend on tifffile, so now instead of "plainly" opening tiffs, metadata was getting added and changing how the viewer ended up.

GenevieveBuckley:
I'm running into problems trying to upload a release candidate to PyPI - it's giving me the same error as if the `napari-tifffile` project name has already been claimed (even though searching for `napari-tifffile` on PyPI turns up zero exact matches). See [the github actions log for the failed deployment here](https://github.com/GenevieveBuckley/napari-tifffile/runs/810686481?check_suite_focus=true).

@kne42 is it possibly related to the early start you made on a napari-tifffile plugin here: https://github.com/kne42/napari-tifffile
I'm very happy to transfer ownership of the repo across to you if we need, that's fine by me.

Czaki:
some handling metadata also for oif and czi file (using other cgholke packages) can be found here https://github.com/4DNucleome/PartSeg/blob/develop/package/PartSegImage/image_reader.py (when writing this I mainly test it on confocal microscopy images) 

This what I see at first is that current code does not check units in scale. And napari does not work well with samm values (eg 210 nm as 210*10**-9). Also imgeJ may use different units in z and XY (different tags) 

Second thing, how napari will handle colormaps passed as arrays not name?

And I'm not sure if it reads TIFF-OME metadata (my code read it partially)

There are still plenty of TIFF metadata format for which I do not have examples so I do not know how to handle its.

jni:
[...]

PR #1488, opened 2020-07-20
Fix rounding rules for indices

Czaki:
# Description

When having images with spacing greater than on and different size along slider axis then it fails with with `IndexError`. when get slider value greater than smaller image size. 

Also original test is wrong:
https://github.com/napari/napari/blob/758a73002460a9da13b3e310a319c29d2521883f/napari/components/_tests/test_dims.py#L130
Because indices 2 is out of range for axis with size 2.

I check and #1360 does not fix this problem.  

sample stack trace:
```
Exception(most recent call first)

IndexError: index 6 is out of bounds for axis 1 with size 6
  File "napari/_qt/qt_dims_slider.py", line 135, in <lambda>
    lambda value: self.dims.set_point(self.axis, value)
  File "napari/components/dims.py", line 359, in set_point
    self.events.axis(axis=axis, value=value)
  File "napari/utils/event.py", line 508, in __call__
  File "napari/utils/event.py", line 529, in _invoke_callback
    cb_event=(cb, event),
  File "napari/utils/event.py", line 523, in _invoke_callback
    cb(event)
  File "napari/components/viewer_model.py", line 341, in _update_layers
    layer.dims.set_point(axis, point)
  File "napari/components/dims.py", line 359, in set_point
    self.events.axis(axis=axis, value=value)
  File "napari/utils/event.py", line 508, in __call__
  File "napari/utils/event.py", line 529, in _invoke_callback
    cb_event=(cb, event),
  File "napari/utils/event.py", line 523, in _invoke_callback
    cb(event)
  File "napari/layers/base/base.py", line 633, in refresh
    self.set_view_slice()
  File "napari/layers/base/base.py", line 586, in set_view_slice
    self._set_view_slice()
  File "napari/layers/image/image.py", line 545, in _set_view_slice
    image = np.asarray(self.data[self.dims.indices]).transpose(order)
```
## Type of change

# References

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/components/dims.py
+++ b/napari/components/dims.py
@@ -269,7 +269,7 @@ class Dims:
                     p = np.clip(
                         self.point[axis],
                         np.round(self.range[axis][0]),
-                        np.round(self.range[axis][1]) - 1,
+                        np.round(self.range[axis][1]) - self.range[axis][2],
                     )
                 else:
                     p = self.point[axis]
--- a/napari/components/dims.py
+++ b/napari/components/dims.py
@@ -268,8 +268,8 @@ class Dims:
                 if self.clip:
                     p = np.clip(
                         self.point[axis],
-                        np.round(self.range[axis][0]),
-                        np.round(self.range[axis][1]) - self.range[axis][2],
+                        self.range[axis][0],
+                        self.range[axis][1] - self.range[axis][2],
                     )
                 else:
                     p = self.point[axis]
--- a/napari/components/dims.py
+++ b/napari/components/dims.py
@@ -266,10 +266,13 @@ class Dims:
                 slice_list.append(slice(None))
             else:
                 if self.clip:
-                    p = np.clip(
-                        self.point[axis],
-                        self.range[axis][0],
-                        self.range[axis][1] - self.range[axis][2],
+                    p = (
+                        np.clip(
+                            self.point[axis],
+                            self.range[axis][0],
+                            self.range[axis][1] - self.range[axis][2],
+                        )
+                        - self.range[axis][0]
                     )
                 else:
                     p = self.point[axis]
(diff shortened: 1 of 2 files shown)


jni (approved):
@Czaki This makes sense to me. It's a bit weird because it's unclear whether the index is supposed to be multiplied by the step size in the clipping, and I guess that's what world coordinates is supposed to fix. I agree that the clipping should be to -step, anyway.

I'll let @sofroniewn review.

sofroniewn (approved):
This looks good to me @Czaki - #1360 hasn't actually changed any slicing so it won't have fixed this yet. Did you have an old issue mentioning this bug that this PR will close or had you just mentioned it on zulip?

If there is an old issue can you mention it here so we can be sure to close it. Thanks!!

Czaki:
fix #1369

napari/components/dims.py
```diff
@@ -269,7 +269,7 @@ def indices(self):
                     p = np.clip(
                         self.point[axis],
                         np.round(self.range[axis][0]),
-                        np.round(self.range[axis][1]) - 1,
+                        np.round(self.range[axis][1]) - self.range[axis][2],
```
sofroniewn:
```suggestion
                        np.round(self.range[axis][1] - self.range[axis][2]),
```
I think the rounding should happen after the step size is subtracted no? As @jni mentions this will be improved, but happy to fix a bug with what looks like backwards compatible behaviour for the majority of cases (where the range value is 1)

Czaki: hm. When I rethink this I do not understand why there is round here. I remove it and add tests which point what I expect. 

It is really important in context of reading metadata form files and set scaling in physical units like in #1384  

I finally understand why I need to multiply image voxels size by 10**9 To reduce IndexErrors. in napari. 

I vote to remove round. There is conversion to int bellow.

sofroniewn:
@Czaki I have verified that #1539 fixes #1369 too with a number of additional improvements. I am going to close this PR as we get close to merging that one, but thanks for getting this going and reporting this bug. We can keep the discussion going on #1539 if there are still issues there that havn't been fixed.

PR #1489, opened 2020-07-20
Use evented dataclass for ViewerModel

tlambert03:
# Description
This ports the `ViewerModel` to an evented dataclass with properties (#1475).  Tests are passing, but it's still a work in progress as I had to make a few hacks to get it to initialize.

## Type of change

# References

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/components/viewer_model.py
+++ b/napari/components/viewer_model.py
@@ -5,11 +5,16 @@ import numpy as np
 from .add_layers_mixin import AddLayersMixin
 from .dims import Dims
 from .layerlist import LayerList
-from ..utils.event import EmitterGroup, Event
 from ..utils.key_bindings import KeymapHandler, KeymapProvider
 from ..utils.theme import palettes
+from ..utils.dataclass import dataclass
+from dataclasses import InitVar
+from typing import Tuple, Optional, Sequence, ClassVar, Dict
+
+DEFAULT_THEME = 'dark'
 
 
+@dataclass(events=True, properties=True)
 class ViewerModel(AddLayersMixin, KeymapHandler, KeymapProvider):
     """Viewer containing the rendered scene, layers, and controlling elements
     including dimension sliders, and control bars for color limits.
@@ -39,46 +44,31 @@ class ViewerModel(AddLayersMixin, KeymapHandler, KeymapProvider):
         Preset color palettes.
     """
 
-    themes = palettes
+    title: str = 'napari'
+    status: str = 'Ready'
+    help: str = ''
+    cursor: str = 'standard'
+    cursor_size: Optional[int] = None
+    interactive: bool = True
+    active_layer: Optional[int] = None
+    grid_size: Tuple[int, int] = (1, 1)
+    theme: str = DEFAULT_THEME
 
-    def __init__(
-        self, title='napari', ndisplay=2, order=None, axis_labels=None
-    ):
-        super().__init__()
+    grid_stride: ClassVar[int] = 1
+    palette: ClassVar[Dict[str, str]] = palettes[DEFAULT_THEME]
+    layers: ClassVar[LayerList] = []
 
-        self.events = EmitterGroup(
-            source=self,
-            auto_connect=True,
-            status=Event,
-            help=Event,
-            title=Event,
-            interactive=Event,
-            cursor=Event,
-            reset_view=Event,
-            active_layer=Event,
-            palette=Event,
-            grid=Event,
-            layers_change=Event,
-        )
+    ndisplay: InitVar[int] = 2  # type: ignore
+    order: InitVar[Optional[Tuple[int, ...]]] = None  # type: ignore
+    axis_labels: InitVar[Optional[Sequence[str]]] = None  # type: ignore
 
+    def __post_init__(self, ndisplay, order, axis_labels):
+        super().__init__()  # will this work?
+        self.events.add(reset_view=None, grid=None, layers_change=None)
+        self.layers = LayerList()
         self.dims = Dims(
             ndim=None, ndisplay=ndisplay, order=order, axis_labels=axis_labels
         )
-
-        self.layers = LayerList()
-
-        self._status = 'Ready'
-        self._help = ''
-        self._title = title
-        self._cursor = 'standard'
-        self._cursor_size = None
-        self._interactive = True
-        self._active_layer = None
-        self._grid_size = (1, 1)
-        self.grid_stride = 1
-
-        self._palette = None
-        self.theme = 'dark'
 
         self.dims.events.camera.connect(self.reset_view)
         self.dims.events.ndisplay.connect(self._update_layers)
@@ -89,7 +79,6 @@ class ViewerModel(AddLayersMixin, KeymapHandler, KeymapProvider):
         self.layers.events.changed.connect(self._on_layers_change)
 
         self.keymap_providers = [self]
-
         # Hold callbacks for when mouse moves with nothing pressed
         self.mouse_move_callbacks = []
         # Hold callbacks for when mouse is pressed, dragged, and released
@@ -97,154 +86,27 @@ class ViewerModel(AddLayersMixin, KeymapHandler, KeymapProvider):
         self._persisted_mouse_event = {}
         self._mouse_drag_gen = {}
 
-    @property
-    def palette(self):
-        """dict of str: str : Color palette with which to style the viewer.
-        """
-        return self._palette
-
-    @palette.setter
-    def palette(self, palette):
-        if palette == self.palette:
-            return
-
-        self._palette = palette
-        self.events.palette()
-
-    @property
-    def theme(self):
-        """string or None : Preset color palette.
-        """
-        for theme, palette in self.themes.items():
-            if palette == self.palette:
-                return theme
-
-    @theme.setter
-    def theme(self, theme):
-        if theme == self.theme:
-            return
-
+    def _on_theme_set(self, theme):
+        # FIXME: in dataclass, need to fallback if an exception is raised in _on_name_set
         try:
-            self.palette = self.themes[theme]
+            self.palette = palettes[theme]
         except KeyError:
             raise ValueError(
-                f"Theme '{theme}' not found; "
-                f"options are {list(self.themes)}."
+                f"Theme '{theme}' not found; " f"options are {list(palettes)}."
             )
 
-    @property
-    def grid_size(self):
-        """tuple: Size of grid
-        """
-        return self._grid_size
-
-    @grid_size.setter
-    def grid_size(self, grid_size):
-        if np.all(self.grid_size == grid_size):
-            return
-        self._grid_size = grid_size
+    def _on_grid_size_set(self, grid_size):
         self.reset_view()
-        self.events.grid()
 
-    @property
-    def status(self):
-        """string: Status string
-        """
-        return self._status
-
-    @status.setter
-    def status(self, status):
-        if status == self.status:
-            return
-        self._status = status
-        self.events.status(text=self._status)
-
-    @property
-    def help(self):
-        """string: String that can be displayed to the
-        user in the status bar with helpful usage tips.
-        """
-        return self._help
-
-    @help.setter
-    def help(self, help):
-        if help == self.help:
-            return
-        self._help = help
-        self.events.help(text=self._help)
-
-    @property
-    def title(self):
-        """string: String that is displayed in window title.
-        """
-        return self._title
-
-    @title.setter
-    def title(self, title):
-        if title == self.title:
-            return
-        self._title = title
-        self.events.title(text=self._title)
-
-    @property
-    def interactive(self):
-        """bool: Determines if canvas pan/zoom interactivity is enabled or not.
-        """
-        return self._interactive
-
-    @interactive.setter
-    def interactive(self, interactive):
-        if interactive == self.interactive:
-            return
-        self._interactive = interactive
-        self.events.interactive()
-
-    @property
-    def cursor(self):
-        """string: String identifying cursor displayed over canvas.
-        """
-        return self._cursor
-
-    @cursor.setter
-    def cursor(self, cursor):
-        if cursor == self.cursor:
-            return
-        self._cursor = cursor
-        self.events.cursor()
-
-    @property
-    def cursor_size(self):
-        """int | None: Size of cursor if custom. None is yields default size
-        """
-        return self._cursor_size
-
-    @cursor_size.setter
-    def cursor_size(self, cursor_size):
-        if cursor_size == self.cursor_size:
-            return
-        self._cursor_size = cursor_size
-        self.events.cursor()
-
-    @property
-    def active_layer(self):
-        """int: index of active_layer
-        """
-        return self._active_layer
-
-    @active_layer.setter
-    def active_layer(self, active_layer):
-        if active_layer == self.active_layer:
-            return
-
-        if self._active_layer is not None:
-            self.keymap_providers.remove(self._active_layer)
+    def _on_active_layer_set(self, active_layer):
+        # FIXME: we need access to the OLD value
+        # if self._active_layer is not None:
+        #     self.keymap_providers.remove(self._active_layer)
 
         self._active_layer = active_layer
 
         if active_layer is not None:
             self.keymap_providers.insert(0, active_layer)
-
-        self.events.active_layer(item=self._active_layer)
 
     def _scene_shape(self):
         """Get shape of currently viewed dimensions.
@@ -263,8 +125,13 @@ class ViewerModel(AddLayersMixin, KeymapHandler, KeymapProvider):
         # Scale the camera to the contents in the scene
         min_shape, max_shape = self._calc_bbox()
         size = np.subtract(max_shape, min_shape)
-        size = [size[i] for i in self.dims.displayed]
-        corner = [min_shape[i] for i in self.dims.displayed]
+
+        if not hasattr(self, 'dims'):
+            dd = [0, 1]
+        else:
+            dd = self.dims.displayed
+        size = [size[i] for i in dd]
+        corner = [min_shape[i] for i in dd]
 
         return size, corner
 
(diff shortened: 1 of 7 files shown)


napari/components/viewer_model.py
```diff
@@ -39,47 +44,32 @@ class ViewerModel(AddLayersMixin, KeymapHandler, KeymapProvider):
...
+    grid_size: Tuple[int, int] = (1, 1)
+    theme: str = DEFAULT_THEME
+
+    grid_stride: ClassVar[int] = 1
+    palette: ClassVar[Dict[str, str]] = palettes[DEFAULT_THEME]
+    layers: ClassVar[LayerList] = []
+
+    ndisplay: InitVar[int] = 2  # type: ignore
+    order: InitVar[Optional[Tuple[int, ...]]] = None  # type: ignore
+    axis_labels: InitVar[Optional[Sequence[str]]] = None  # type: ignore
```
sofroniewn, line 70: Interesting with the `InitVar` (see https://docs.python.org/3/library/dataclasses.html#post-init-processing) for those unfamiliar with it.

Will `InitVar` not get their own event emission? Can they be modified later? I guess not - see
`They are not otherwise used by dataclasses.` in the docs. I think that's ok though, just want to check.

napari/components/viewer_model.py
```diff
@@ -39,47 +44,32 @@ class ViewerModel(AddLayersMixin, KeymapHandler, KeymapProvider):
...
+    cursor: str = 'standard'
+    cursor_size: Optional[int] = None
+    interactive: bool = True
+    active_layer: Optional[int] = None
+    grid_size: Tuple[int, int] = (1, 1)
+    theme: str = DEFAULT_THEME
+
+    grid_stride: ClassVar[int] = 1
+    palette: ClassVar[Dict[str, str]] = palettes[DEFAULT_THEME]
+    layers: ClassVar[LayerList] = []
```
sofroniewn: `ClassVar` are just treated as normal variables by the dataclass, and won't get their own event emission, which is quite nice https://docs.python.org/3/library/dataclasses.html#class-variables

napari/components/viewer_model.py
```diff
@@ -39,47 +44,32 @@ class ViewerModel(AddLayersMixin, KeymapHandler, KeymapProvider):
...
+
+    ndisplay: InitVar[int] = 2  # type: ignore
+    order: InitVar[Optional[Tuple[int, ...]]] = None  # type: ignore
+    axis_labels: InitVar[Optional[Sequence[str]]] = None  # type: ignore
+
+    def __post_init__(self, ndisplay, order, axis_labels):
+        super().__init__()  # will this work?
```
sofroniewn: I guess we needed the `super().__init__()` call because of our inheritance pattern above!! Pretty crazy if all that inheritance worked just fine with the dataclass approach and didn't need to be changed.

We're we adding attributes in those base classes? I could see how that would cause problems with the dataclass syntax, but I guess if we were just adding methods it might be ok

napari/components/viewer_model.py
```diff
@@ -39,47 +44,32 @@ class ViewerModel(AddLayersMixin, KeymapHandler, KeymapProvider):
...
+        self.events.add(reset_view=None, grid=None, layers_change=None)
+        self.layers = LayerList()
```
sofroniewn: Is it not possible to do the `LayerList()` call in the actual init? i.e. `layers: ClassVar[LayerList] = LayerList()`?

napari/components/viewer_model.py
```diff
@@ -263,8 +125,13 @@ def _scene_shape(self):
         # Scale the camera to the contents in the scene
         min_shape, max_shape = self._calc_bbox()
         size = np.subtract(max_shape, min_shape)
-        size = [size[i] for i in self.dims.displayed]
-        corner = [min_shape[i] for i in self.dims.displayed]
+
+        if not hasattr(self, 'dims'):
```
sofroniewn: why do we need this new `hasattr(self, 'dims')`, isn't this added in the `post_init` which will happen before it is ever needed?

sofroniewn (commented):
Looking great, I love how much code is getting removed. I left a few comments, most are just helping me understand dataclasses and their syntax, if it looks like i'm getting something wrong feel free to correct me. This will be a big help for when I try and do the Dims model - particularly the `InitVar` syntax.

The one thing I'm not sure why you need now is the `hasattr` checks around the dims model. It's not clear to me why those methods would be called without the dims model being there.

napari/components/viewer_model.py
tlambert03: Yeah, that's by design.  These are variables that are required to initialize the dataclass state, but have no independent meaning after that.  So there are no events (because they don't exist after initialization).
Think of them as parameters that get used only during init (which is how those three are used here)
```python
class Test:
    def __init__(self, a, b):
        self.c = a * b
    # a and b are gone now...
```
if they are actually important attributes for the dataclass, just remove the InitVar class and they will have events.

and, as you probably saw in the docs, `InitVars` will be passed to the `__post_init__` method.

napari/components/viewer_model.py
tlambert03: yeah, I'm not totally in love with that pattern... 
Technically, according to [PEP526](https://www.python.org/dev/peps/pep-0526/#class-and-instance-variable-annotations):

> `ClassVar` is a special class defined by the typing module that indicates to the static type checker that this variable should not be set on instances.

so while that has no runtime effect, and it _does_ have the desired effect of making these "private" attributes, it's not exactly _semantically_ correct.  So in the commit I just pushed, I added another syntax for excluding events, that is probably better:
```python
layers: LayerList = field(default_factory=LayerList, metadata={'events': False})

# note also that using mutables as default values like I did there is a NO-NO
layers: ClassVar[LayerList] = []
```

napari/components/viewer_model.py
tlambert03: yeah, it's working!  I _think_ we're just adding methods... but I need to check.  If we were also adding attributes, it can sometimes still be ok, but it's a little more tricky.

napari/components/viewer_model.py
tlambert03: updated, 
```python
layers: LayerList = field(default_factory=LayerList, metadata={'events': False})
```
and see comment above about mutable defaults:

napari/components/viewer_model.py
tlambert03: those are gone now too.  I needed to fix some things about the evented dataclass that was causing events to be emitted during the init.

tlambert03:
Almost everything is working ... with one very strange bug that probably hints at some other deeper issue.  If I launch the program (`python -m napari`), and then click once on the blank canvas... it crashes back to the command line with:
```pytb
WARNING: Error drawing visual <vispy.visuals.mesh.MeshVisual object at 0x13032c880>
11:50:22 WARNING Error drawing visual <vispy.visuals.mesh.MeshVisual object at 0x13032c880>
```
(but no traceback)

strange huh??  I can't figure out what `MeshVisual` it would possibly be trying to draw.  And, if I add a layer and delete it, then it doesn't happen again.  In fact, if I simply open the console window, it doesn't happen.  So there's some sort of vispy initialization that's wonky, and this seems to be exposing it.

napari/components/viewer_model.py
sofroniewn: hmm yeah, i guess in a couple places we could use the `field` syntax or not, and ultimately we'll probably want to make one decision for things to be as consistent as possible ....

napari/components/viewer_model.py
sofroniewn: nice, yeah great not to need those!!

>  I needed to fix some things about the evented dataclass that was causing events to be emitted during the init.

It really is huge to be doing these PRs now to try this stuff out before we've merged everything, it's great to see all these little quirks that come up during real world usage!

napari/components/viewer_model.py
tlambert03: we're not going to be able to avoid using `field` entirely since it's the only way to provide a mutable default.  For instance, if you want something to start as an empty list you have to use `default_factory=list` (That's a good thing since doing `__init__(self, x = [])` is a [bad](https://docs.python-guide.org/writing/gotchas/) [thing](https://florimond.dev/blog/articles/2018/08/python-mutable-defaults-are-the-source-of-all-evil/)®).  And [`field`](https://docs.python.org/3/library/dataclasses.html#dataclasses.field) is a central player in the dataclass API... so I think also using them to sprinkle metadata here or there in unique cases is fine.

tlambert03:
> If I launch the program (`python -m napari`), and then click once on the blank canvas... it crashes back to the command line with:

turns out this isn't something from the new dataclass, I was actually getting this in master as well, starting after #1374 ... however if I create a new environment, I don't get it anymore.  So while something funny did happen there, it seems to be specific to my environment.  I'm wiping it but just so I have the details somewhere:

<details>

<summary>conda list</summary>

```
aiobotocore               0.12.0                   pypi_0    pypi
aiohttp                   3.6.2                    pypi_0    pypi
aioitertools              0.7.0                    pypi_0    pypi
alabaster                 0.7.12                   pypi_0    pypi
appdirs                   1.4.4                    pypi_0    pypi
appnope                   0.1.0                    pypi_0    pypi
asciitree                 0.3.3                    pypi_0    pypi
async-timeout             3.0.1                    pypi_0    pypi
attrs                     19.3.0                   pypi_0    pypi
babel                     2.8.0                    pypi_0    pypi
backcall                  0.2.0                    pypi_0    pypi
black                     19.10b0                  pypi_0    pypi
bleach                    3.1.5                    pypi_0    pypi
botocore                  1.15.15                  pypi_0    pypi
ca-certificates           2020.6.20            hecda079_0    conda-forge
cachey                    0.2.1                    pypi_0    pypi
certifi                   2020.6.20        py38h32f6830_0    conda-forge
cfgv                      3.1.0                    pypi_0    pypi
chardet                   3.0.4                    pypi_0    pypi
check-manifest            0.42                     pypi_0    pypi
click                     7.1.2                    pypi_0    pypi
cloudpickle               1.5.0                    pypi_0    pypi
coiled                    0.0.10                   pypi_0    pypi
commonmark                0.9.1                    pypi_0    pypi
coverage                  5.2                      pypi_0    pypi
cycler                    0.10.0                   pypi_0    pypi
dask                      2.20.0                   pypi_0    pypi
decorator                 4.4.2                    pypi_0    pypi
defusedxml                0.6.0                    pypi_0    pypi
distlib                   0.3.1                    pypi_0    pypi
distributed               2.20.0                   pypi_0    pypi
docutils                  0.15.2                   pypi_0    pypi
entrypoints               0.3                      pypi_0    pypi
fasteners                 0.15                     pypi_0    pypi
filelock                  3.0.12                   pypi_0    pypi
flake8                    3.7.9                    pypi_0    pypi
freetype-py               2.1.0.post1              pypi_0    pypi
fsspec                    0.7.4                    pypi_0    pypi
heapdict                  1.0.1                    pypi_0    pypi
identify                  1.4.21                   pypi_0    pypi
idna                      2.10                     pypi_0    pypi
imageio                   2.9.0                    pypi_0    pypi
imagesize                 1.2.0                    pypi_0    pypi
ipykernel                 5.3.1                    pypi_0    pypi
ipython                   7.16.1                   pypi_0    pypi
ipython-genutils          0.2.0                    pypi_0    pypi
ipywidgets                7.5.1                    pypi_0    pypi
jedi                      0.17.1                   pypi_0    pypi
jinja2                    2.11.2                   pypi_0    pypi
jmespath                  0.10.0                   pypi_0    pypi
jsonschema                3.2.0                    pypi_0    pypi
jupyter-client            6.1.5                    pypi_0    pypi
jupyter-core              4.6.3                    pypi_0    pypi
kiwisolver                1.2.0                    pypi_0    pypi
libcxx                    10.0.0               h1af66ff_2    conda-forge
libffi                    3.2.1             h4a8c4bd_1007    conda-forge
magicgui                  0.1.5                    pypi_0    pypi
markupsafe                1.1.1                    pypi_0    pypi
matplotlib                3.2.2                    pypi_0    pypi
mccabe                    0.6.1                    pypi_0    pypi
mistune                   0.8.4                    pypi_0    pypi
monotonic                 1.5                      pypi_0    pypi
more-itertools            8.4.0                    pypi_0    pypi
msgpack                   1.0.0                    pypi_0    pypi
multidict                 4.7.6                    pypi_0    pypi
mypy                      0.782                    pypi_0    pypi
mypy-extensions           0.4.3                    pypi_0    pypi
napari                    0.3.6.dev37+g4b133241           dev_0    <develop>
napari-plugin-engine      0.1.6                    pypi_0    pypi
napari-svg                0.1.3                    pypi_0    pypi
nbconvert                 5.6.1                    pypi_0    pypi
nbformat                  5.0.7                    pypi_0    pypi
ncurses                   6.1               h0a44026_1002    conda-forge
networkx                  2.4                      pypi_0    pypi
nodeenv                   1.4.0                    pypi_0    pypi
notebook                  6.0.3                    pypi_0    pypi
numcodecs                 0.6.4                    pypi_0    pypi
numpy                     1.19.0                   pypi_0    pypi
numpydoc                  1.1.0                    pypi_0    pypi
open3d                    0.10.0.0                 pypi_0    pypi
openssl                   1.1.1g               h0b31af3_0    conda-forge
packaging                 20.4                     pypi_0    pypi
pandas                    1.0.5                    pypi_0    pypi
pandocfilters             1.4.2                    pypi_0    pypi
parso                     0.7.0                    pypi_0    pypi
pathspec                  0.8.0                    pypi_0    pypi
pep517                    0.8.2                    pypi_0    pypi
pexpect                   4.8.0                    pypi_0    pypi
pickleshare               0.7.5                    pypi_0    pypi
pillow                    7.2.0                    pypi_0    pypi
pip                       20.1.1                     py_1    conda-forge
pluggy                    0.13.1                   pypi_0    pypi
pooch                     1.1.1                    pypi_0    pypi
pre-commit                2.6.0                    pypi_0    pypi
prometheus-client         0.8.0                    pypi_0    pypi
prompt-toolkit            3.0.5                    pypi_0    pypi
psutil                    5.7.0                    pypi_0    pypi
ptyprocess                0.6.0                    pypi_0    pypi
py                        1.9.0                    pypi_0    pypi
pycodestyle               2.5.0                    pypi_0    pypi
pyflakes                  2.1.1                    pypi_0    pypi
pygments                  2.6.1                    pypi_0    pypi
pyopengl                  3.1.5                    pypi_0    pypi
pyparsing                 2.4.7                    pypi_0    pypi
pypcd                     0.1.1                    pypi_0    pypi
pyqt5                     5.14.2                   pypi_0    pypi
pyqt5-sip                 12.8.0                   pypi_0    pypi
pyrsistent                0.16.0                   pypi_0    pypi
pytest                    5.4.3                    pypi_0    pypi
pytest-faulthandler       2.0.1                    pypi_0    pypi
pytest-ordering           0.6                      pypi_0    pypi
pytest-qt                 3.3.0                    pypi_0    pypi
pytest-timeout            1.4.1                    pypi_0    pypi
python                    3.8.3           cpython_hd5f0129_0    conda-forge
python-dateutil           2.8.1                    pypi_0    pypi
python-lzf                0.2.4                    pypi_0    pypi
python_abi                3.8                      1_cp38    conda-forge
pytz                      2020.1                   pypi_0    pypi
pywavelets                1.1.1                    pypi_0    pypi
pyyaml                    5.3.1                    pypi_0    pypi
pyzmq                     19.0.1                   pypi_0    pypi
qtconsole                 4.7.5                    pypi_0    pypi
qtpy                      1.9.0                    pypi_0    pypi
readline                  8.0                  hcfe32e1_0    conda-forge
recommonmark              0.6.0                    pypi_0    pypi
regex                     2020.6.8                 pypi_0    pypi
requests                  2.24.0                   pypi_0    pypi
rope                      0.16.0                   pypi_0    pypi
scikit-image              0.17.2                   pypi_0    pypi
scipy                     1.5.1                    pypi_0    pypi
send2trash                1.5.0                    pypi_0    pypi
setuptools                49.1.0           py38h32f6830_0    conda-forge
six                       1.15.0                   pypi_0    pypi
snowballstemmer           2.0.0                    pypi_0    pypi
sortedcontainers          2.2.2                    pypi_0    pypi
sphinx                    3.0.1                    pypi_0    pypi
sphinx-rtd-theme          0.4.3                    pypi_0    pypi
sphinxcontrib-applehelp   1.0.2                    pypi_0    pypi
sphinxcontrib-devhelp     1.0.2                    pypi_0    pypi
sphinxcontrib-htmlhelp    1.0.3                    pypi_0    pypi
sphinxcontrib-jsmath      1.0.1                    pypi_0    pypi
sphinxcontrib-qthelp      1.0.3                    pypi_0    pypi
sphinxcontrib-serializinghtml 1.1.4                    pypi_0    pypi
sqlite                    3.32.3               h93121df_0    conda-forge
tblib                     1.6.0                    pypi_0    pypi
terminado                 0.8.3                    pypi_0    pypi
testpath                  0.4.4                    pypi_0    pypi
tifffile                  2020.7.4                 pypi_0    pypi
tk                        8.6.10               hbbe82c9_0    conda-forge
toml                      0.10.1                   pypi_0    pypi
toolz                     0.10.0                   pypi_0    pypi
tornado                   6.0.4                    pypi_0    pypi
traitlets                 4.3.3                    pypi_0    pypi
typed-ast                 1.4.1                    pypi_0    pypi
typing-extensions         3.7.4.2                  pypi_0    pypi
typing-inspect            0.6.0                    pypi_0    pypi
urllib3                   1.25.9                   pypi_0    pypi
virtualenv                20.0.25                  pypi_0    pypi
vispy                     0.6.4                    pypi_0    pypi
wcwidth                   0.2.5                    pypi_0    pypi
webencodings              0.5.1                    pypi_0    pypi
wheel                     0.34.2                     py_1    conda-forge
widgetsnbextension        3.5.1                    pypi_0    pypi
wrapt                     1.12.1                   pypi_0    pypi
xarray                    0.15.1                   pypi_0    pypi
xz                        5.2.5                h0b31af3_0    conda-forge
yarl                      1.4.2                    pypi_0    pypi
zarr                      2.4.0                    pypi_0    pypi
zict                      2.0.0                    pypi_0    pypi
zlib                      1.2.11            h0b31af3_1006    conda-forge
```

</details>

napari/components/viewer_model.py
```diff
@@ -97,155 +93,26 @@ def __init__(
...
-
-    @active_layer.setter
-    def active_layer(self, active_layer):
-        if active_layer == self.active_layer:
-            return
-
-        if self._active_layer is not None:
-            self.keymap_providers.remove(self._active_layer)
+    def _on_active_layer_set(self, active_layer):
+        # FIXME: we need access to the OLD value
```
sofroniewn, line 107: Why isn't `self._active_layer` still the old value? i.e. if the arg `active_layer` is the new value and this method is getting called as a `super` first thing in the setter then when it gets called the old value is still available right and the new value hasn't been set yet?

tlambert03: > this method is getting called as a super first thing in the setter

That's not what this method is doing. It's not meant as a setter, (that's what the `Property[..., ..., setter]` is for.  This method was there in case you needed to do complex things like emit different events, _prevent_ the event in certain cases, and/or call additional internal methods after setting.  If you just want to modify the setting process, I think we should be using a `Property`

sofroniewn:
This will be superseded by #2005

PR #1492, opened 2020-07-21
Consolidate constants

tlambert03:
# Description
I often find myself wondering where a particular enum was defined.  I've also noticed that we have a lot of different enums all called `Mode` whose definition is context specific (and so could not easily be used in the same module).  Also we have the same `ColorMode` duplicated in three different places.

Curious how everyone feels about this PR, that consolidates all of our constants into a single `napari.utils.constants` module ... so you always know where to look for them?

## Type of change

- [x] refactor (mostly just moving files)

# References

# How has this been tested?

## Final checklist:


napari/utils/misc.py
```diff
@@ -121,66 +120,6 @@ def formatdoc(obj):
         del frame
 
 
-class StringEnumMeta(EnumMeta):
```
sofroniewn, line 124: I'd be tempted to move these into a new file that is not the actual file with the constants either, might keep things most clear/ readable

tlambert03: that's fine, they can also go back to utils.misc where they came from

sofroniewn:
Curious @kevinyamauchi how you think about this, you've spent a lot of time with our constants. Are there many things shared across Points, Shapes, Labels now such that having things centralized would make it easier?

Curious how @jni feels too?

pwinston:
I like the idea of "constants", "defines", "types" files that pull together or flatten definitions that are scattered about. I think it's pragmatic and convenient even if it means less of a hierarchy.

One quibble though to me "util/utilities" are generally things that could be easily pulled out and thrown into another codebase, not just things that don't physically depend on the rest of the codebase. Like they need to thematically be separable, not just physically separable. 

I wonder if we could make a new directory like "common" or "system" or something? Maybe there is other stuff we could move to there. To me the hallmark of a utility is something you could throw into a different application and it would be useful there. But that may just the be the code bases I've seen, maybe there is a different convention with "utils" and python.

kevinyamauchi (commented):
I am definitely in favor of the renaming (e.g., `Mode` to `<layer>Mode`) and consolidation of duplicated constants. I think both of those things improve the readability of the code and make it easier to determine where things come from.

I could go either way on moving all of the constants to a single file. In some ways, I think having the constants close to where they are used makes it easier to find them, especially if we rename the constants such that they have unique names. The exception is when we have common constants that are shared across multiple files and are thus in some other constants file, likely at a different level in the hierarchy. That being said, I could see that becoming much less common as we move to having more data classes (e.g., `ColorMode` would then live near `ColorManager`).

tlambert03:
> In some ways, I think having the constants close to where they are used makes it easier to find them, especially if we rename the constants such that they have unique names.

I definitely agree with this... but this only really works if they are only used in one place.  And they kinda _all_ have to be used in one "close" place, otherwise some of them are used locally, and others are all across multiple modules, which is the current situation.  So then you kinda have to know "this is one of those constants that isn't just used in one place"

sofroniewn:
ok, seems like people generally like or are at least ok with consolidation and see some other benefits around naming, reduced duplication. I'm on board too, but I probably wouldn't this new file in utils. We could create a new folder called types, put `types.py` and `constants.py` inside there? I'd leave things like `class StringEnumMeta` inside `utils`.

How does that sound?

PR #1494, opened 2020-07-21
Close napari window when KeyboardInterupt

### LINKED ISSUE #1493: CTRL-C should exit napari gracefully
## 🐛 Bug

After #1476 napari just prints `KeyboardInterrupt` when CTRL-C is pressed in the system terminal window that used to launch napari.  Prior to 1476 it exited with a crash, which got the job done but was not great. 

Ideally napari would exit gracefully when you hit CTRL-C in the system terminal window.

## To Reproduce

1. From a system terminal (e.g. Terminal program on mac)
2. Run "napari" or a script that uses `napari.gui_qt()`
3. Switch back to the terminal window and type CTRL-C

## Expected behavior

Napari exits gracefully.

## Environment

```
napari: not-installed
Platform: macOS-10.15.3-x86_64-i386-64bit
Python: 3.8.1 (default, Jan 8 2020, 16:15:59) [Clang 4.0.1 (tags/RELEASE_401/final)]
Qt: 5.14.2
PyQt5: 5.14.2
NumPy: 1.18.4
SciPy: 1.4.1
Dask: 2.17.2
VisPy: 0.6.5.dev111+g8387ea1a.d20200424

GL version: 2.1 ATI-3.5.5
MAX_TEXTURE_SIZE: 16384

Plugins:
- napari-plugin-engine: 0.1.6
- svg: 0.1.3
```

## Additional context

This is low priority since you can exit with the Quit command, or from the system terminal hit CTRL-Z and `kill %1` the app if necessary. However it seems like exiting gracefully is the right behavior long term.

I tried adding this to our new `ExceptionHandler` class:
```
        # Interpret CTRL-C as a request to quit.
        if isinstance(value, KeyboardInterrupt):
            QApplication.instance().quit()
            return
```
but while it exits cleanly sometimes, sometimes it bus errors or seg faults.

Czaki:
# Description

This is related to discussion on zulip chat. This allows to close napari with `Ctrl+C` from terminal. 
fix #1493

## Type of change

# References

# How has this been tested?

## Final checklist:


pwinston:
This is not "graceful" in that `sys.exit(1)` does not give anything a chance to cleanup or shutdown. However it seems to successfully exit without any errors, in some quick tests, which is good.

So it seems certainly better than not exiting at all. 

I think any remaining bug would be of the form "if I hit Quit then such and such happens but if I hit CTRL-C it doesn't happen" and I'm not sure we'd care too much about that?

Czaki:
> This is not "graceful" in that `sys.exit(1)` does not give anything a chance to cleanup or shutdown. However it seems to successfully exit without any errors, in some quick tests, which is good.

I think that if you need to clean application with Ctrl+C is not a problem that you cannot clean, because something wrong happen.

Even sys.exit could be handled because it raise `SystemExit` exception. It could be changed, and first call some function for clean. Then on second `Ctrl+C` call sys.exit.

pwinston:
It seems reasonable to me. I would ask some others to review it though.

Even if someday we do something fancier, this seems better than not exiting at all.

tlambert03 (approved):
I'm fine with this as a short-term solution.  I think this should probably be moved in the future as @ziyangczi builds out the "delegation" of different exception types.  I would like to know why `app.quit` doesn't work.  but that too can wait if people want to get ctrl-c back

sofroniewn (approved):
I'm also good with this as a short-term fix. I'll note though that for me on my mac I have to focus back from the terminal to the GUI before `ctrl-C` actually has the effect of quitting. Not sure if that is the same for everyone? It is how things used to be for me before the other PR merged, so this PR returns the behaviour to what it used to be, but it would be nice if `ctrl-C` from the terminal quit the GUI without having to toggle focus.

See this gif for what I mean. Note the `ctrl-C` get detected in the console, but the quit not happen until the focus change
ctrl_c

napari/_qt/exceptions.py
```diff
@@ -42,4 +43,7 @@ def handle(
         # can add custom exception handlers here ...
         text = "".join(traceback.format_exception(etype, value, tb))
         logging.error("Unhandled exception:\n%s", text)
+        if isinstance(value, KeyboardInterrupt):
+            print("Close by KeyboardInterrupt", file=sys.stderr)
```
jni:
```suggestion
            print("Closed by KeyboardInterrupt", file=sys.stderr)
```

napari/_qt/_tests/test_exception_handler.py
```diff
@@ -15,3 +18,6 @@ def test_exception_handler(qtbot, caplog):
     assert record.levelname == 'ERROR'
     assert 'Unhandled exception' in record.message
     assert 'ValueError: whoops' in record.message
+    with pytest.raises(SystemExit):
```
jni, line 26: This is great! But generally I'd prefer a separate test, rather than a single test testing every single behaviour.

jni (commented):
The behaviour on Mac (haven't yet tested on Linux) isn't quite right on my machine... If I press Ctrl-C with the Terminal in focus, nothing happens. Only once I click back to napari does it actually close. Has anyone else encountered this?

```python
^C^C14:17:11 ERROR Unhandled exception:
  File "/Users/jni/conda/envs/all/lib/python3.8/site-packages/vispy/app/backends/_qt.py", line 507, in event
    def event(self, ev):
KeyboardInterrupt

Close by KeyboardInterrupt
```

It would also be great if we could remove the "unhandled exception" message for this specific use case... Dunno how hard that would be. @ziyangczi?

jni:
Oh @sofroniewn beat me to it! Stupid GH non-auto-refresh...

Personally I'd love it if we spent a bit more time to investigate how we can do this properly (Ctrl-C actually exits). I don't think there's a massive rush here...

Czaki:
> The behaviour on Mac (haven't yet tested on Linux) isn't quite right on my machine... If I press Ctrl-C with the Terminal in focus, nothing happens. Only once I click back to napari does it actually close. Has anyone else encountered this?

I see same behaviour. It looks like signals are ignored till there is no python  code execution. When focus comes to window then refresh function (ex from vispy) is called.

Write signal handler does not help. 

It looks like possible solution are described here
https://stackoverflow.com/questions/4938723/what-is-the-correct-way-to-make-my-pyqt-application-quit-when-killed-from-the-co

liu-ziyang:
ah, the evil keyboard interrupt 🤣 
my take: 
1. yeah we need to handle it specially, a minor change I would make to the PR is to let the exception handler have the app instance, and call the shutdown before sys.exit, which is a more elegant/proper exit. (you also need to update the init method to pass the app instance) I tested it and it works out fine. (regarding the exit code, I argue it should be 0 since this is an intended way to shut down app)
```        
       if isinstance(value, KeyboardInterrupt):
            self.app.shutdown()
            sys.exit(0)
```
2. regarding the focus glitch, as pointed out, this is a qt event loop thing, unless we force it to run frequently to process the event, (which I don't think we should just for this one glitch), we pretty much live with this behavior. (one more reason why I don't like the qt event loop 🤣 )

I also don't know how to capture gifs so here's my test result: 

(venv) zliu@czirwc1macos743 napari % python examples/add_image.py
^C%                                                                                                                                                                                                       (venv) zliu@czirwc1macos743 napari % echo $?
0

tlambert03:
> you also need to update the init method to pass the app instance

don't pass the app instance, it's a global/singleton:

```python

app = QApplication.instance()
app.quit()
```

liu-ziyang:
> > you also need to update the init method to pass the app instance
> 
> don't pass the app instance, it's a global/singleton:
> 
> ```python
> from qtpy.QtWidgets import QApplication
> 
> app = QApplication.instance()
> app.quit()
> ```

Ah! I didn't know that, even better then!

Czaki:
hm...

For me Ctrl+Z is not convenient way to close program, but method to break calculation or kill program if it frezee. 
Then I personally expect other return code than 0. 
Of course there is still SYGKILL on UNIX like systems. But it consume more time and less user know it.  (Ctrl+Z then `ps -a` check process pid  and `kill -9 process_pi`)

So @sofroniewn if You prefer to call `app.quit()` before this change I will add it to this PR but for me it could call part of code which could stuck somehow which is not expected during emergency quit.

sofroniewn:
@ziyangczi @Czaki cross-posting link here from @tacaswell that matplotlib is looking into solving this problem with this approach - could be something we might be interested in adapting too. https://github.com/matplotlib/matplotlib/pull/13306 I havn't looked into details, but might be worth investigating

Czaki:
@sofroniewn It looks interesting. I review code and it looks interesting, but my whole experience with sockets is on linux. Whole socket trick is to not wait on python code from interaction with interface. 

I could add something similar, but I'm not sure if this should not be moved to separated PR. There are used some system specific solutions, and may need longer discussion/check.

And still  matplotlib/matplotlib#13306 uses Ctrl+C as alternative of quit button, not as emergency quit. 

I prefer to wait unitll matplotlib/matplotlib#13306 is merged (more people will check this approach)

tacaswell:
From a slack conversation, pyqtgraph does not do anything special with sigint.

Czaki:
> From a slack conversation, pyqtgraph does not do anything special with sigint.

@tacaswell But what happen when sigint is send? It is ignored, or its close application?  And if pqqtgraph provide own excepthook? . If `sys.excepthook` is not overwritten then application is closed by sigint.

Because napari use custom exception handler to print debug information in dialog.

sofroniewn:
I'd like us to revisit merging this. I (and judging from the comments others) find this behavior strictly better than what we have now, and I don't think we should let the perfect be the enemy of the good. Note that the focus issue remains, but that was there before this and will be there after. It's also much nicer to have this functionality than not having at all.

@Czaki if you can fix the merge conflict, then I think we should merge.

tlambert03 (approved):

Czaki:
Rebased

napari/_qt/_tests/test_exception_handler.py
```diff
@@ -29,4 +40,4 @@ def test_exception_handler_gui(qtbot, make_test_viewer):
     assert handler.message.property('expanded') is True
     handler.message.toggle_expansion()
     assert handler.message.property('expanded') is False
-    handler.message.close()
+    handler.message.close()
```
sofroniewn:
```suggestion
    handler.message.close()

```

napari/_qt/_tests/test_exception_handler.py
sofroniewn: this has now been done

sofroniewn:
I have just fixed another round of merge conflicts, and will merge now. We can always continue to improve. Thanks for the work on this @Czaki!!

PR #1329, opened 2020-06-02
0.3.4 release notes

sofroniewn:
# Description

0.3.4 release notes. Note sure how much detail we want to put in the release notes around the installation changes @tlambert03 @jni?

PR #1495, opened 2020-07-22
Async loading of Image layer

### LINKED ISSUE #1320: Blocked UI while loading images in notebook

pwinston:
# Description
Attempt to fix #1320 by adding asynchronous loading of `Image` layers. This includes single and multi-scale images but we have not yet introduced an octree or grid or anyway to load tiles/portions of larger images. That will be added when we address #845 and #1300.

# Status

**Mostly implemented, needs more real world testing and bug finding!!**

The Torch example in #1320 performs well outside of Jupyter using processes such as with [this script](https://github.com/pwinston/napari-pwinston/blob/master/noise2self/noise2self_script.py). You can run in Jupyter with threads, but it's no faster than the original issue because the Torch thread slows down the GUI thread (GIL?).  Or you can run in Jupyter with processes but you will get this crash: https://github.com/napari/napari/issues/1320#issuecomment-662043521.

So solving #1320 was mixed success. However the PR brings async to many cases beyond #1320. And threads are totally adequate for IO or "light" computations.

# How To Use

There are two ways to enable async loading:

1. Define env var `NAPARI_ASYNC=1` to get the defaults which are currently:

```
DEFAULT_ASYNC_CONFIG = {
    "synchronous": False,
    "num_workers": 6,
    "log_path": None,
    "use_procesess": False,
    "delay_seconds": 0.1,
    "load_seconds": 0,
}
```

2. Define env var `NAPARI_ASYNC=/some/path/async-config.json` to configure settings above.

The config file can:

* Set the number workers. 
* Set worker type to be threads or processes.
    * Processes are necessary only if threads are not performing well.
    * Processes are experimental and tricky and may or may not work.
* Set `delay_seconds` to avoid spamming the thread pool.
* Set `load_seconds` to simulate load latency.
* Record a log file, format is not useful for most people.

# How It Works

If async is not enabled we use the same `Image` class that's in master, so the user it taking on zero risk from any new code. If async is enabled then `Image` is defined to be `AsyncImage` which is new, which interacts with the new `ChunkLoader`.

The new`ChunkLoader` class creates a `concurrent.futures` pool of threads (or processes). When the `Image` class needs raw data to render it asks the `ChunkLoader` to load the data. There are two cases:

1. If the data is an `ndarray` we call `asarray` right there in the GUI thread. This seems to take basically no time, so there is no point in involving workers.
2. The data is not `ndarray` in which case we submit it to the worker pool, which returns a `conurrent.Future` that the `ChunkLoader` holds on to. The worker does basically nothing except call `asarray` on the data. With Dask or other means this might lead to IO or computation.

When the worker finishes loading it calls `ChunkLoader._done`. The call to `_done` can technically be in any thread but so far I've only seen it happen in the worker thread. Because we don't want to call into the `Layer` code in a worker thread, we fire a `chunk_loaded` event which `QtChunkReceiver` listens to. `QtChunkReceiver` is funky little class that does nothing but use a slot/signal to call itself in the GUI thread, and then call into `Image.chunk_loaded` in the GUI thread.

# Threads or Processes?

### Advantages of threads

* Fast to startup, nothing is start except the one function.
* Threads can freely access all memory in the process, no copying data.
* Threads can fire our normal model events.
* Python threads are first class operating system threads.
* Threads can run in parallel inside system calls or in C/C++ modules (numpy, etc).

### Disadvantages of threads

* Only one Python thread is allowed to run Python bytecode at a time.
    * See this [short article](https://medium.com/python-features/pythons-gil-a-hurdle-to-multithreaded-program-d04ad9c1a63) or this [long presentation](http://www.dabeaz.com/GIL/)  on the Global Interpreter Lock (GIL).

### Advantages of processes

* Processes can run totally independently, each has its own GIL.
* A worker process cannot directly crash the main process.
* A worker process can be throttled or killed separately from the main process.

### Disadvantages of processes

* Starting up a worker process is slow.
* Worker processes can crash on startup, they might do things that shouldn't be done a 2nd time.
* By default no memory is shared so you must serialize and copy data back and forth.
    * Even with shared memory you must serialize
    * Pointers/references in one process are not meaningful in the other process.
* Napari will simply crash if it can't serialize the data.

### Conclusions

* For IO bound loading (disk or network) threads should work great.
* For computation threads should work if the computation is "thread friendly".
* For heavy computations that do have contention with the GUI thread, processes might be necessary.
* It might be nice to say napari only does threads, to keep napari simple. If you need processes use Dask to create an external process that napari somehow "connects to"? This would require some research and Dask-expertise to setup.

# Async Tests

* We want to test napari with both sync and async, since both are valid modes.
* We can't run sync and async as two passes, see crash in #1329.
* Instead in our [.cirrus.yml](https://github.com/pwinston/napari/blob/feature/chunk-loader-2/.cirrus.yml) we run `pytest` twice once regular and once with `--async_only` flag.
* Any tests marked `@pytest.mark.sync_only` are skipped during the async pass
    * These tests will fail in async mode because they need the data to be instantly available.

# Related:
- Open issue this PR is fixing: #1320
- Merged PR with perfmon changes only: #1453 
- Closed PR that combined perfmon and async changes: #1354 
- Closed PR with async changes only: https://github.com/pwinston/napari/pull/5
- Future async issues: #1300 and #845 

# How has this been tested?

## Final checklist:


pwinston:
multi-scale-async

We could call this "single-tile multi-scale" but it's working. Before it was 0.5Hz to move the slider now it's around 30Hz. Using this PR running this command:
```
NAPARI_ASYNC=1 napari https://s3.embassy.ebi.ac.uk/idr/zarr/v0.1/6001240.zarr
```

pwinston:
Before the "delay queue" fix above we were starting lots of loads while scrolling rapidly through slices. As long as we kept scrolling their results were ignored:

multi-scroll-1

With the delay queue we delay all submits by 100ms. So we can easily cancel loads during the delay period. Even though there's a delay, this can actually be *faster* than without the delay because it's more likely a worker will be available to immediately start the load when we finally pause long enough, they won't be busy loading stale data we no longer need.

multi-scroll-2
Above there are two layers, so two worker threads. Each thread loads the `image` array and the `thumbnail_source` array, shown in closeup below:

pwinston:
The new `ChunkRequest` created for multi-scale can take a dict of arrays to load. For multi-scale we request two arrays one with the key `image` and one `thumbnail_source`. They are separately timed:

multi-load
For single scale we just request `image` and use it also as the `thumbnail_source`. Being able to load N chunks in one request might come in handy with tuning things. If chunks are small or access is cheap it might make sense to bundle them together in one request.

pwinston:
Loading this small remote zarr works great:
```
https://s3.embassy.ebi.ac.uk/idr/zarr/v0.1/6001240.zarr
```
Loading this large one works horribly. In master you can pan it freely from where it loads but with async branch you can't even pan smoothly. I know why are creating a new slice on every refresh but with multi-scale we refresh all the time. We just need to only re-create the slice when needed. Should not be too hard.
```
https://s3.embassy.ebi.ac.uk/idr/zarr/v0.1/4495402.zarr
```

pwinston:
The number of changes to `Image` even if set to synchronous mode were piling up. For the moment I've reverted `Image.py` basically back to the version that's in master. Only if `NAPARI_ASYNC` is set do you get the new class which is now called `AsyncImage`. This should make merging this PR much less risky, since only those opting into `NAPARI_ASYNC` will get new code.

Later we can switch over to the new `ImageAsync.py` class, rename it to `Image`. It can operate in synchronous and asynchronous mode. But for the now let's not force people onto the new class unless they opt-in with `NAPARI_ASYNC`.

The one downside is if someone modifies `Image` say to handle a new event, we need to modify `AsyncImage` to match.

.cirrus.yml
```diff
@@ -25,12 +25,21 @@ lint_task:
...
+    env:
       PY_VER: "3.7"
+      PYTEST_FLAGS: ""
+    env:
       PY_VER: "3.8"
-
+      PYTEST_FLAGS: ""
+    env:
+      PY_VER: "3.8"
+      PYTEST_FLAGS: --async_only
```
pwinston, line 41: Maybe there is a nicer way to do this? But this basically adds a new pass which is `async_only` which means run all the tests with async on, except skip any test marked `sync_only` since we know they would fail in async mode.

napari/__init__.py
```diff
@@ -75,6 +75,8 @@
 # register napari object types with magicgui if it is installed
 _magicgui.register_types_with_magicgui()
 
+# Context object to disable asynchronous loads.
+from .utils.chunk import synchronous_loading
```
pwinston, line 79: Only our `conftest.py` uses this. But could be useful to API users?

napari/_qt/qt_chunk_receiver.py
```diff
@@ -0,0 +1,48 @@
+from qtpy.QtCore import QObject, Signal
+
+from ..layers.base import Layer
+from ..utils.chunk import ChunkRequest, chunk_loader
+from ..utils.perf import add_instant_event
+
+
```
pwinston, line 7: This `ChunkLoader` specific class seems like it could be turned into something totally generic. Some sort of thing that maps our events to Qt slots. Maps workers thread events to GUI events. Probably not worth working make a generic thing. But it is pointing towards some type of "call this event but in the GUI thread" type of generic thing.

jni:
I'm testing this before diving into the code. I'll write notes here as I test things.

Note 1: in the small remote zarr example, switching to 3D view is broken. Here's the traceback:

```pytb
14:24:48 ERROR Unhandled exception:
  File "/Users/jni/projects/napari/napari/_qt/qt_viewer_buttons.py", line 295, in change_ndisplay
    self.viewer.dims.ndisplay = 3
  File "/Users/jni/projects/napari/napari/components/dims.py", line 296, in ndisplay
    self.events.ndisplay()
  File "/Users/jni/projects/napari/napari/utils/event.py", line 508, in __call__
  File "/Users/jni/projects/napari/napari/utils/event.py", line 525, in _invoke_callback
    _handle_exception(
  File "/Users/jni/projects/napari/napari/utils/event.py", line 523, in _invoke_callback
    cb(event)
  File "/Users/jni/projects/napari/napari/components/viewer_model.py", line 335, in _update_layers
    layer.dims.ndisplay = self.dims.ndisplay
  File "/Users/jni/projects/napari/napari/components/dims.py", line 296, in ndisplay
    self.events.ndisplay()
  File "/Users/jni/projects/napari/napari/utils/event.py", line 508, in __call__
  File "/Users/jni/projects/napari/napari/utils/event.py", line 525, in _invoke_callback
    _handle_exception(
  File "/Users/jni/projects/napari/napari/utils/event.py", line 523, in _invoke_callback
    cb(event)
  File "/Users/jni/projects/napari/napari/layers/base/base.py", line 413, in _update_dims
    self.refresh()
  File "/Users/jni/projects/napari/napari/layers/base/base.py", line 643, in refresh
    self.set_view_slice()
  File "/Users/jni/projects/napari/napari/layers/base/base.py", line 596, in set_view_slice
    self._set_view_slice()
  File "/Users/jni/projects/napari/napari/layers/image/_async_image.py", line 535, in _set_view_slice
    self.data_level = len(self.data) - 1
  File "/Users/jni/projects/napari/napari/layers/image/_async_image.py", line 351, in data_level
    self.refresh()
  File "/Users/jni/projects/napari/napari/layers/base/base.py", line 644, in refresh
    self.events.set_data()
  File "/Users/jni/projects/napari/napari/utils/event.py", line 508, in __call__
  File "/Users/jni/projects/napari/napari/utils/event.py", line 525, in _invoke_callback
    _handle_exception(
  File "/Users/jni/projects/napari/napari/utils/event.py", line 523, in _invoke_callback
    cb(event)
  File "/Users/jni/projects/napari/napari/_vispy/vispy_image_layer.py", line 94, in _on_data_change
    self._on_display_change(data)
  File "/Users/jni/projects/napari/napari/_vispy/vispy_image_layer.py", line 48, in _on_display_change
    self.node.set_data(data)
  File "/Users/jni/projects/napari/napari/_vispy/vendored/volume.py", line 520, in set_data
    raise ValueError('Volume visual needs a 3D image.')
ValueError: Volume visual needs a 3D image.
```

The viewer still "works" but is stuck in a strange state, with a 2D canvas but no slider. After some time (I think the time it takes to actually fetch the full dataset), I get the next traceback:

```pytb
14:25:06 ERROR Unhandled exception:
  File "/Users/jni/projects/napari/napari/_qt/qt_chunk_receiver.py", line 42, in _chunk_loaded_gui
    layer.chunk_loaded(request)
  File "/Users/jni/projects/napari/napari/layers/image/_async_image.py", line 664, in chunk_loaded
    if not self._slice.chunk_loaded(request):
  File "/Users/jni/projects/napari/napari/layers/image/_image_slice.py", line 135, in chunk_loaded
    image = chunks['image'].transpose(order)
ValueError: axes don't match array
```

jni:
On the other hand, 3D async works fine normally, so there's something strange going on with the zarr data. The following works great (!) in both 2D and 3D:

```python
import dask.array as da
import napari

arr = da.random.random((200, 512, 512, 512), chunks=(1, 512, 512, 512))

with napari.gui_qt():
    _ = napari.view_image(arr)
```

It's damn cool actually. =D

jni:
If I run @pwinston scriptification of the N2S notebook [here](https://github.com/pwinston/napari-pwinston/blob/master/noise2self/noise2self_script.py), using this json config:

```json
{
    "synchronous": false,
    "num_workers": 6,
    "log_path": null,
    "use_procesess": true,
    "delay_seconds": 0.1,
    "load_seconds": 0
}
```

everything works super smooth (❤️!) ~~*unless* I hit the edges of the volume... I think?~~ actually, unless I pause at a slice long enough to trigger a load, and then moving away while it's loading is sluggish again. I get the following log lines when this happens:

```
ImageSlice.chunk_loaded: IGNORE CHUNK (140464670693392, (2984, (None, None, None), (None, None, None)))
```

I *guess* this is caused by the same issues that causes this example to fail on threads only. But then that again points to something *other* than PyTorch + GIL, since in this case that is happening on a different process... Right?

jni:
When running the CREMI example with labels, I get weird behaviour, neither of which is desirable:

- if I download [CREMI sampleA](https://cremi.org/static/data/sample_A_20160501.hdf), and run [this script](https://gist.github.com/jni/e41b101d12717a25a2156fa6c1004f37), `NAPARI_ASYNC=1` doesn't make any difference, nor does using processes, it seems equally sluggish to move between layers. [**edited to provide correct data link**]
- If I convert it to zarr with [this script](https://gist.github.com/jni/65d0e98f4cc2007870289250378c9354), and then load the *zarr* data with [this script](https://gist.github.com/jni/01be74639d6b975596d1ce4437d333fd), then with ASYNC the slider is more responsive, **but**, the experience is actually worse, because the blacking out of all layers is so jarring. In this specific case, the raw images are actually super fast, but the labels slow things down. (Turn off visibility for the labels layer to see this.) (This is related to #204.) In some sense I think what we want is to make layers sync or async based on median latency in slice switching or something like that — anything that is faster than, say, 20Hz, should be synchronous. Maybe? I dunno, looking forward to discussing this use case, it's tricky and I don't know what the right solution is in my head.

napari/layers/image/_async_image.py
```diff
@@ -0,0 +1,765 @@
...
+
+        self._slice = None
+
+        # TODO_ASYNC: creating self._slice here seems totally wrong. This
+        # echoes how we did things before, and it's currently needed because
+        # stuff in _init_ requires it.
+        #
+        # We should get rid of this somehow and not create the slice
+        # until _update_dims() is called at the end of __init__ which
+        # will call refresh() and _set_view_slice().
```
jni: 👍

pwinston: This is totally gone with recent changes. I basically redid image.py from scratch and now there are much less changes. We now create the ImageSlice identically to how we already created it. Which is great, less change.

pwinston:
I made a huge simplification to `image.py`. Basically I started over on the file from scratch given lessons learned during this whole PR. I now rely much more heavily on visibility which is something we didn't come up with until pretty far into the PR.  By relying on visibility more we have to make much fewer changes to the `Image` class to introduce async.

We check visibility in a few new places:

1. In `_vispy.vendored.image.py` in `ImageVisual.set_data` we do nothing if not visible.
2. In `_vispy.vispy_base_layer.py` in `VispyBaseLayer._on_visible_change` we check `layer.loaded`
3. In `_vispy.vispy_image_layer.py` in `VispyImageLayer._on_data_change` we check `self.node.visible`

Now whether this is right or causes bugs we'll have to see. But this **vastly** simplifies the changes we have to make to `Image.py`. Such that I got rid of the `Image` vs. `AsyncImage` class. We are back to just one `Image` class again, because I feel like it's very clear that the **synchronous** path through that class should be equivalent to the pre-async version.

pwinston:
1. The small zarr file seems to work now switching to 3D and back to 2D give it a try. The code changed a lot since you tried it:
```
https://s3.embassy.ebi.ac.uk/idr/zarr/v0.1/6001240.zarr
```
2.  I'm not seeing your `da.random.random((200, 512, 512, 512))` work even though you say it's great! It hangs on load for me. How long did it take to load? I tried the older code I think you were using and latest.

3. The noise2self script with processes is always fast for me. Zero hangs that I saw. I didn't see any issues with the code I think you were using or the new code that's in there now. If it is slow for you we might need better ways to do repros. Maybe record a perfmon JSON? Or a GIF or both? I uploaded my version working fine to YouTube to get around GIF limits: https://youtu.be/mdy0RofUiLw

4. Your cremi links was to the 20160601 file, it doesn't have labels. I had to download the [20160501](https://cremi.org/static/data/sample_C_20160501.hdf) one. The layers are both `ndarray` so there is no async. This might be the motivation we need for a Layer Properties dialog that shows what type layers really are and their latency or bandwidth. Slices of h5py Datasets are ndarray they aren't anything deferred:
```
>>> d = data['volumes/raw']
>>> type(d)
<class 'h5py._hl.dataset.Dataset'>
>>> s = d[0, :, :]
>>> type(s)
<class 'numpy.ndarray'>
```
If I use dask like `raw = da.from_array(data['volumes/raw'])` then the image layer is async, but labels is sync. If I do it for labels too then the slider flies. It renders black but we have to address that globally it's not specific to this example.

5. Converting to zarr it flies the same as in 4 so it's all async and it works. In perfmon it looks like both image and labels are fast to load, so both could be synchronous. I think we could auto-detect pretty easily, but I could also imagine that making mistakes sometimes. Even more motivation to have a Layer Properties dialog that shows layers and lets you toggle sync or async, at least for developers while we are still learning.

Separately we need to just never go to the black, we have to figure out what to show. For example in multi-scale if you pan right now it goes black. I think in both single-scale and multi-scale we want to show the old visual but somehow indicate it's stale. But it is still useful as a reference, more useful than black. Especially in multi-scale if you pan or zoom --- going black is totally disorienting. 

I don't have any good design ideas to show a "stale" slice. Can't really "gray it out" since a lot of our data is gray to begin with. Maybe just a loading icon/animation that gets drawn on top. We need something simple to prove we can do it, then we can get some ideas from @liaprins-czi.

Note: I can see repro's being an issue with async because one person might be doing subtle things with timing or clicking that is hard to convey to the other person.

napari/layers/image/_async_image.py
```diff
@@ -586,101 +559,107 @@ def _set_view_slice(self):
...
+
+    def _load_async(self, image_indices, image, thumbnail_source=None):
+        """Request the ChunkLoader to load this image and maybe thumbnail source.
 
-    def _load_single_scale(self) -> None:
-        """Load non-multiscale image.
+        The load will happen synchronously if:
+        1) The array is already an ndarray.
+        2) We are in synchronous mode.
+        3) The array was found in the ChunkCache.
```
jni: It should be made clear whether this is an "and" or an "or". From context I guess "or". =)

```suggestion
        1) The array is already an ndarray;
        2) We are in synchronous mode; or,
        3) The array was found in the ChunkCache.
```

(alternatively the top line could say "if any of the following conditions holds:") (or you could make both changes)

napari/layers/image/_async_image.py
```diff
@@ -586,101 +559,107 @@ def _set_view_slice(self):
...
+
         image = request.chunks['image']
 
-        # ASYNC_TODO: Complain if the loaded data's shape's rgb status was
-        # different from what we inferred it to be on load. Basically if
-        # the user lied about the shape of their array. Not sure if we want
-        # to handle this better somehow?
+        # ASYNC_TODO: If the shape is wrong relative to rgb it means the user
+        # liked to Dask about the shape of their array. When you create a
+        # Dask array you tell it the same and you could be wrong.
```
jni:
```suggestion
        # lied to Dask about the shape of their array. When you create a
        # Dask array you tell it the shape and you could be wrong.
```

napari/layers/image/_async_image.py
```diff
@@ -586,101 +559,107 @@ def _set_view_slice(self):
...
         self.events.loaded()
 
         # Update vispy, draw the new slice
         self.events.set_data()
 
     def _update_thumbnail(self):
         """Update thumbnail with current image data and colormap."""
-        if self._slice is None or not self._slice.loaded:
+        # ASYNC_TODO: no point in computing the thumbnail until we've
+        # loaded, but do we need to have teh caller not call this until
```
jni:
```suggestion
        # loaded, but do we need to have the caller not call this until
```

jni:
> Slices of h5py Datasets are ndarray they aren't anything deferred

Whoa! I *think* that's new-ish behaviour. (Haven't played with h5py properly since Janelia!) Good to know, and that explains everything. 👍 

> Even more motivation to have a Layer Properties dialog that shows layers and lets you toggle sync or async, at least for developers while we are still learning.

👍 👍 👍 

Thanks @pwinston, great comments, and indeed I thought I probably should have produced a screen recording for the sluggish issues. Will do so shortly on the latest code.

jni:
(Apologies for my outdated code comments above: github notifications dumped me into a view of a single old commit — normally it takes me to a collection of recent commits, dunno why that happened.)

napari/layers/image/_async_image.py
pwinston: Good point, I updated it see [new version here](https://github.com/pwinston/napari/blob/feature/chunk-loader-2/napari/layers/image/image.py#L580-L583). Now back in `Image.py`.

napari/layers/image/_async_image.py
pwinston: Fixed [see commit](https://github.com/napari/napari/pull/1495/commits/3c2b7eec2b817cc9961cbc5013c5cbd890886a49).

napari/layers/image/_async_image.py
pwinston: That comment/typo when away in latest `image.py`.

pwinston:
Tried replacing this but didn't work, leaving this open for now.

pwinston:
Opened new #1510 with these same changes but squashed down to just 5 commits that are logically the 5 major steps of the PR, closing this one.

Issue #903, opened 2020-01-23
Running napari from a docker container


AhmetCanSolak:
## 🚀 Feature

Running napari from a docker container. Simple idea but not that simple to implement.

## Motivation

Could be nice to be able to run napari from a docker container at least on same machine. There are certain software pieces delivered as a docker image only, we can help users of such software pieces by letting them starting napari from docker container. So people will not try to map files/folders and keep switching between terminals.

## Pitch

This is not easy to implement. There are some previous work for limited experiences with X11 forwarding: https://github.com/jozo/docker-pyqt5 . Here another related discussion: https://stackoverflow.com/questions/24095968/docker-for-gui-based-environments. Possibly future-napari-webviewer can address this issue as well.

## Additional context

I just wanted to put this idea out there and see if there are any creative solutions.

tlambert03:
Definitely an idea worth pursuing!  Just cause I've been reading about it the last couple days, I would throw in the idea of building a container with [virtualGl](https://virtualgl.org/About/Background) baked in.  From what (little) i've learned so far, i think it might well be a good first-pass solution for many of our remote problems, but it comes with a somewhat difficult installation process (for most users).  perhaps a docker container could ease that?

DrLachie:
Just came across this when running a gui interface from docker, as in the links above. Would love to be able to fold napari into the gui we're working on (works locally, just not on the docker image).

tlambert03:
tagging this with "sprint".  For me, this would be a challenging task, but given the broad range of backgrounds at SciPy, there may will be someone who's got a lot of docker/container experience.  The main challenge, as I see it, would be to get napari running in a docker container that _also_ has some kind of [virtualGl](https://www.virtualgl.org/) or other virtual frame buffer server running.  See brief conversation starting around https://github.com/napari/napari/issues/896#issuecomment-577197014

When getting it running locally, I found this gist
https://gist.github.com/cyberang3l/422a77a47bdc15a0824d5cca47e64ba2
and these docs
https://cdn.rawgit.com/VirtualGL/virtualgl/2.6.3/doc/index.html
were useful

sofroniewn:
I wonder here if we tried using vispy's webgl backend - see here http://vispy.org/installation.html#jupyter-notebook-extension if that might open additional avenues for us. Note that they describe webgl support as experimental, and possibly not fully featured, but napari uses only a subset of vispy anyway. Could be one avenue worth pursuing

liu-ziyang:
[...]

PR #1496, opened 2020-07-22
Docker Integration by adding dockerfile and docker.md

liu-ziyang:
# Description
Add docker file for docker builds

## Type of change
- This change requires a documentation update

# References
https://github.com/napari/napari/issues/903

# How has this been tested?

## Final checklist:


Czaki:
Which libraries are missed? 
Maybe this list will help https://pytest-qt.readthedocs.io/en/latest/troubleshooting.html#github-actions

or method from this comment allow to find missed one https://github.com/pytest-dev/pytest-qt/issues/266#issuecomment-510035926

sofroniewn:
> Which libraries are missed?

Sorry @Czaki I'm not quite sure what you mean here, can you elaborate?

@ziyangczi is there any follow up here, if not I will merge asap, sorry for the delay!!

dockerfile
```diff
@@ -0,0 +1,21 @@
...
+ENV DEBIAN_FRONTEND noninteractive
+
+# install graphical libraries used by qt and vispy
+RUN apt-get install -qqy libxi6 libglib2.0-0 fontconfig libgl1-mesa-glx libfontconfig1 libxrender1 libdbus-1-3
+
+# install napari release version
+RUN pip3 install napari[all]
+
+# library missing when using provided pyqt version, reversed to an earlier version for now
+RUN pip3 install PyQt5==5.11.3
```
Czaki: @sofroniewn I think about this comment in Dockerfile

liu-ziyang: I think something was fixed along with 0.3.8, I was able to build with this line removed, updating now

sofroniewn:
We're getting an error on checking of the manifest

```
check-manifest
lists of files in version control and sdist do not match!
missing from sdist:
  dockerfile
suggested MANIFEST.in rules:
  include dockerfile
```

we probably don't want the dockerfile in there (right @jni?) so could just do
```
exclude dockerfile
```
and then the test will be happy I think

jni:
> we probably don't want the dockerfile in there (right @jni?)

Right! It would be pretty weird to ship that in the sdist. 😂

sofroniewn (approved):
Ok, this looks good to me! Let's get it in there, we can continue to update if we need / if people make requests!! I've also fixed the manifest.

johschmidt42:
Any progress on a working dockerfile for linux OS? I'd be satisfied with a nvidia-docker image.

sofroniewn:
> Any progress on a working dockerfile for linux OS? I'd be satisfied with a nvidia-docker image.

I believe the latest was merged here and the most up-to-date resource we have on napari and docker is here https://napari.org/guides/stable/docker.html. @johschmidt42 if you'd like to contribute anything to improve the situation it would be more than welcome, or if anyone else reading has any ideas too that would be great. I'm afraid I don't know any more than this

johschmidt42:
I've tried the most recent and past docker images but none of them worked for me with X11 forwarding on a Linux machine. I've spent a day to figure out why, built my own docker images to make it run, but no success. Surprisingly, running a X Window server on a windows machine as suggested in the docs actually works but is far from desireable because it is painfully slow..
Running a simple qt app with python works in a docker container with x window forwarding but fails with napari due to driver issues and opengl. There are nvidia-OpenGL docker images that should theoretically make it work if you have a nvidia card but that also wasn't successful. So I'd really appreciate it if someone figures out how to make it run..

PR #1497, opened 2020-07-23
Evented data class for dims

sofroniewn:
# Description
This PR continues on from #1486 but now adding in the dataclass from #1475 too. It's still WIP but the basics are working.

I like that a lot of code has disappeared. The hardest thing to figure out is the initialization as @tlambert03 mentioned to me in passing recently.

I'm still not sure how to pass in the `order` and `axis_labels` for initialization. We previously used `None`, but that value couldn't be coerced into the `EventedList`. I'm also not quite sure I'm doing the right thing with the `field` syntax for the eventedlist initialization, and there might want a property that cannot be set or destroyed - that list should only be modified in place.

Feedback welcome!

## Type of change


sofroniewn:
I've fixed the initialization I think, but I had to move away from the `field` syntax to the `Property` syntax. It would still be nice to have a way to make the evented lists unsettable as doing `viewer.dims.point = [10, 0, 0]` will now create a new EventedList that isn't hooked up to anything and so the slider will no longer function.

Also the `viewer.dims.point[:] = [10, 0, 0]` syntax doesn't emit the right events, nor does the `viewer.dims.axis_labels[:] = ['a', 'b', 'c']`

also it doesn't look like the `_on_order_set` method gets called during the init, so the "error checking" i'm doing there doesn't get run - which might actually be good in this case as if it did it might error, but i just want to point that out as someone might expect different.

Basically i think this is pretty done, the two failing tests in `test_dims.py` both relate to initialization errors not being caught anymore and failing test in `test_qt_dims.py` relates to the `viewer.dims.axis_labels[:]` syntax.

Overall it's been a pretty smooth experience using both #1444 and #1475!!

tlambert03:
> It would still be nice to have a way to make the evented lists unsettable as doing `viewer.dims.point = [10, 0, 0]` will now create a new EventedList that isn't hooked up to anything and so the slider will no longer function.

the `dataclass` decorator provides a `frozen` parameter that prevents mutations, but it applies all-or-nothing to the whole class.  But I was thinking we could look for `metadata={'frozen':True}` in the `field` param to accomplish this.  Alternatively, i could make it `field(frozen=True)` but that would make it so that you had to use the `field` from napari instead of `dataclasses`.


> Also the `viewer.dims.point[:] = [10, 0, 0]` syntax doesn't emit the right events, nor does the `viewer.dims.axis_labels[:] = ['a', 'b', 'c']`

what does it currently emit (if anything) and what would you like it to emit?

> also it doesn't look like the `_on_order_set` method gets called during the init, so the "error checking" i'm doing there doesn't get run - which might actually be good in this case as if it did it might error, but i just want to point that out as someone might expect different.

thats true.  those are not hooked up until after `__init__` (but I think before `__post__init__`). I _think_  that's generally what we'll want, to avoid excessive events during init, but you could try `self.x = self.x` in post init to see if it works?
 
> Overall it's been a pretty smooth experience using both #1444 and #1475!!

🎉

sofroniewn:
> > Also the viewer.dims.point[:] = [10, 0, 0] syntax doesn't emit the right events, nor does the viewer.dims.axis_labels[:] = ['a', 'b', 'c']
>
> what does it currently emit (if anything) and what would you like it to emit?

I get a single `changed` call with `event.index = slice(None, None, None)`. I think that's actually good, I just need to adapt my listener to deal with that. Really this belongs in #1486, I am just discovering it here. I can work with this.

> those are not hooked up until after __init__ (but I think before __post__init__). I think that's generally what we'll want, to avoid excessive events during init

Sounds reasonable to me.

> but you could try self.x = self.x in post init to see if it works?

makes sense, i can give that a go tomorrow if i think it will help

sofroniewn:
This is superseded by #1917

PR #1502, opened 2020-07-24
[WIP] Layergroups (TJL)

tlambert03:
# Description
This is a branch from #1089 that has diverged quite a bit, and is still a long way from ready (and it will be broken into smaller PRs as well) ... but I want to put it up here so I can link to and refer to a couple things.  Kinda focused on the evented models right now, but i feel like all the key parts are coming together nicely in this PR.  I'll write up more of an explanation later, but here are the key classes:

- `napari.utils.tree.node.Node` all of the logic required to be a member of a generic tree (needn't be a layer tree)... so this is reusable
- `napari.utils.tree.node.Group` a composite node that can have children.  This is also a subclass of  `NestableEventedList` from #1444.
-  `napari.layers.base.base.Layer(..., Node, ...)` - `Layer` now subclasses Node and becomes a specific type of tree Node.
-  `napari.layers.layergroup.LayerGroup(Layer, Group)` - subclasses both Layer and Group.
- `napari._qt.tree._tree_model.QtNodeTreeModel & QtLayerTreeModel` - this is where a lot of good stuff is happening.  This is a concrete implementation of [`QAbstractItemModel`](https://doc.qt.io/qt-5/qabstractitemmodel.html) that allows the aforementioned `Group` class to provide the interface needed by any of the `QtView` types (including `QTreeView`), without actually needing to store any information directly (unlike the widgets we typically use).  This means no synchronization is necessary between "our" model, and the "view model".
- `napari._qt.tree._tree_model.NodeMimeData` - still very minimal, this will be a better container for layer information as it's being dragged/dropped (for instance, between viewers, or even into other applications!)
- `napari._qt.tree._tree_model.QtLayerTreeView` - the view into the layergroup model... This will become the QtLayerList
- `napari._qt.tree._tree_model.LayerDelegate` - totally unstyled for now... this is what will provide the look of the layer widgets in the tree, by extending [`QStyledItemDelegate`](https://doc.qt.io/qt-5/qstyleditemdelegate.html)


sofroniewn:
@tlambert03 are you still working on this one, or is it all now in other PRs? Just doing some housekeeping

tlambert03:
I'm not 100% sure... i'll close this one when i'm confident

tlambert03:
see #2261

PR #1504, opened 2020-07-25
Use TypedEventedList and TypedList for LayerList and TransformChain

sofroniewn:
# Description
This PR builds on #1444 by using the `TypedEventedList` for our `LayerList` and the `TypedList` for our `TransformChain`.

It was pretty much a drop in replacement and a smooth experience.

Right now tests are failing though as the 

```python
# Rearrange layers by tuple
layers[:] = layers[(1, 0, 2)]
```

syntax no longer works. This can likely be fixed in #1444 and then all tests will pass.

```python-traceback
napari/components/_tests/test_layers_list.py:105:
napari/utils/events/containers/_typed_list.py:68: in __getitem__
    return super().__getitem__(key)
    def __getitem__(self, key):  # noqa: F811 (redefinition)
>       result = self._list[key]
E       TypeError: list indices must be integers or slices, not tuple
```

## Type of change

- [x] Refactor


tlambert03:
> layers[:] = layers[(1, 0, 2)] no longer works

ahhhh,  _there's_ that test!! 😄  I was looking all over for how the `__prsitem__`, `__locitem__` methods were actually used/tested (another unit-test before integration-test argument I guess).  OK let me think about that.
What are your thoughts here?  I kinda want it to be a method like `sort` or something, i find that tuple-indexing syntax a little mysterious at first glance (not to mention that we'll need to use tuple indexing differently for a nested layer list)

sofroniewn:
> the last thing I'm running into when using this for the layerlist is issues around layer selection. That _add method that listens to the layerlist's own added event should probably not have been connected as an event, but rather as a direct method. and the way it unselects everything but the last added layer is giving me problems when using move_selected which also has a lot of conditional logic to manipulate the selection status of layers

Following up on this https://github.com/napari/napari/pull/1444#issuecomment-663903783, I'm running into the same issues. It's not clear to me why the `_add` method is causing the problems related to `move_selected` but regardless I agree the `_add` seems a little undesirable. Could be instead overwrite and call `super` inside `__setitem__`?

tlambert03:
Yeah, see my PR to your branch for that.  It's not the add method itself per se... but I think the connection method should change anyway.

I think it has more to do with the unselect_all/ignore step. And now the order of operations matters.  So it's be nice to decouple selection from the insertion/setting process

sofroniewn:
> It's not clear to me why the _add method is causing the problems related to move_selected but regardless I agree the _add seems a little undesirable. Could be instead overwrite and call super inside __setitem__?

Ok I think I see what's happening now - the new `layers[:] = [layers[i] for i in (1, 0, 2)]` syntax is triggering `inserted` events which is causing problems as previously the reordering triggered a `reordered` event and not an `added` event (which has now been replaced by inserted).

I've been able to drop the `_add` method by overwriting and then calling `super` on the `inserted` method, which works great, but we should figure out a way to make the reordering not trigger `inserted` events if we can

sofroniewn:
> Yeah, see my PR to your branch for that. It's not the add method itself per se... but I think the connection method should change anyway.

ah sorry, missed this, will look now!!

> I think it has more to do with the unselect_all/ignore step. And now the order of operations matters. So it's be nice to decouple selection from the insertion/setting process

I guess, but on the other hand if we want adding a layer to deselect all the other layers then we sort of need this coupling somewhere? Maybe there is a cleaner way to do it, but I think we like and want to preserve the overall functionality, no?

sofroniewn:
Playing around with this right now, just want to note that

```python
viewer.layers[:] = [viewer.layers[0], viewer.layers[0]]
```

gives some pretty pathological behaviour as we now have two references (not copies) to the same layer in our layerlist each with the same name (because they are the same layer). Not sure what to do in this case, but maybe we want to prevent it.

sofroniewn:
@tlambert03 I've gone through and fixed all tests now and fixed some untested behaviour that I couldn't figure out how to test, but that was causing qt bugs if not addressed.

I've had to add a `blocker_all` context to the `EventEmitter` and went back to having some of the selection behaviour in the layer list triggered by events rather than overwriting builtins as it can be disabled by blockers, whereas overwriting the builtins cannot.

I had to add all this logic for disconnecting the events when removing an item, which is a little horrific, but otherwise I can trigger all sorts of nasty errors by doing

```python
viewer.layers[:] = [viewer.layers[i] for i in (1, 0)]
viewer.layers[:] = [viewer.layers[i] for i in (1, 0)]
```
as during the second call I'm triggering events on the deleted qt object that look like

```python-traceback
ERROR:root:Unhandled exception:
  File "/Users/nsofroniew/GitHub/napari/napari/_qt/qt_layerlist.py", line 283, in mouseReleaseEvent
    layer.selected = True
  File "/Users/nsofroniew/GitHub/napari/napari/layers/base/base.py", line 520, in selected
    self.events.select()
  File "/Users/nsofroniew/GitHub/napari/napari/utils/events/event.py", line 513, in __call__
  File "/Users/nsofroniew/GitHub/napari/napari/utils/events/event.py", line 534, in _invoke_callback
    cb_event=(cb, event),
  File "/Users/nsofroniew/GitHub/napari/napari/utils/events/event.py", line 528, in _invoke_callback
    cb(event)
  File "/Users/nsofroniew/GitHub/napari/napari/_qt/qt_layerlist.py", line 483, in <lambda>
    layer.events.select.connect(lambda v: self.setSelected(True))
  File "/Users/nsofroniew/GitHub/napari/napari/_qt/qt_layerlist.py", line 540, in setSelected
    self.setProperty('selected', state)
RuntimeError: Internal C++ object (QtLayerWidget) already deleted.
```

Definitely some stuff still to clean up, but this should now be 100% working and bug free!!

sofroniewn:
Ah ok, last fails are current screenshot tests which have syntax
```python
viewer.layers[0, 5] = viewer.layers[5, 0]
# and 
viewer.layers[0, 1] = viewer.layers[1, 0]
```
what is the new recommended syntax for this operation?

Do we need a `swap` method too?
```python
viewer.layers.swap(0, 5)
# and 
viewer.layers.swap(0, 1)
```

tlambert03:
>  we now have two references (not copies) to the same layer in our layerlist each with the same name (because they are the same layer). Not sure what to do in this case, but maybe we want to prevent it.

funny, I was just going to say that it's good that we now have a "`prepare_layer`" method as it can raise an error if the layer is already in the list... but

> went back to having some of the selection behaviour in the layer list triggered by events rather than overwriting builtins as it can be disabled by blockers, whereas overwriting the builtins cannot.

I really think this is the wrong direction.  Having the layerlist's own methods triggered by its own events is, I believe, an anti-pattern we generally want to avoid if at all possible.  I believe this is also the sort of "event abuse" that @pwinston was talking about in meeting a while back.  When we _respond_ to events, the "insertion" has already happened and then the layerlist can only react and try to "undo it" if something went wrong, whereas if we reimplement the method we can prevent "weird stuff" from ever happening in the first place.  While that kind of thing might sometimes be necessary for two seperate objects trying to communicate, it should never be necessary for a single object "communicating with itself".   The layer shouldn't be added in the first place, the name should be changed before addition, etc...

> `viewer.layers[0, 5] = viewer.layers[5, 0]`
> what is the new recommended syntax for this operation?

I need to look at where that was being used.  Using tuple indexing in `__setitem__` is super non-standard...  Other than being a "neat" one-step trick... is there any case where we critical need to swap the position of two random layers in a single operation rather than two?

sofroniewn:
> I need to look at where that was being used. Using tuple indexing in __setitem__ is super non-standard, and I need to look into where we were using that. Other than being a "neat" one-step trick... is there any case where we critical need to swap the position of two random layers in a single operation rather than two?

Not particularly, but reordering is a super common thing I think people will want to do and we should just make easy. I find things like a `layers.reorder(my_tuple)` where `list(my_tuple).sort()==list(range((layers))` (i.e. must be a permutation) and `layer.swap(index_a, index_b)` where `index_a`, `index_b` are keys (could be `int` or `layer.name` for a typed list) the most natural to reason about and probably most intuitive for a user. I might be inclined to even treat `move_multiple` as private as that really is just simulating drag/ drop and just expose a `reorder` and a `swap`. I'm completely fine dropping "using tuple indexing in __setitem__" and always found that confusing anyway.

> funny, I was just going to say that it's good that we now have a "prepare_layer" method as it can raise an error if the layer is already in the list... but
> 
> > went back to having some of the selection behaviour in the layer list triggered by events rather than overwriting builtins as it can be disabled by blockers, whereas overwriting the builtins cannot.
>
> I really think this is the wrong direction. Having the layerlist's own methods triggered by its own events is, I believe, an anti-pattern we generally want to avoid if at all possible. 

🤣  you're probably right here, and I'm all down for reducing event emission, but I this will then require more significant changes (which might ultimately bring us in a better direction anyway).

I think the conflict right now is between using blockers during the "move like operation" and overwriting internal methods as the blockers prevent the listeners from responding but the internal method still gets called as before. To the outside it looks like it was just "a move" which is what we want, but to the inside it looks like it was a bunch of removes and a bunch of adds, and now things have "gotten out of sync"

Take for example the disconnecting I'm now doing when a layer gets deleted. If the layer really has been deleted then I need to disconnect all those methods otherwise bad things will happen. If the layer is only being moved then I don't. If I'm listening to the removed event then I know that's a real deletion so I can just wait for that to happen. If I overwrite the builtin delete item I don't know if it's a remove or just a move. If I treat them the same and go ahead and do all the disconnecting regardless then when I add the layer back in I better be sure that I reconnect everything, but now because the inserted event is blocked and the connections are being made by the listeners they don't get added back in. If we removed all the blockers then everything should actually work ok, but then you never actually want to listen to any move/ reorder events as every move/ reorder is really just a sequences of removes and adds. Listening to those removes/ adds alone will get you to the right place. Listening to that extra reorder will just cause an extra unneeded step. I do think though it feels very heavy to have every reordering of the list be treated like a full burn down and rebuild.

Anyway, I'm definitely excited to do less weird things with events, we just need to figure out how to do it in a way that preserves all the functionality we want, and this was the first thing I was able to do that got all the `layerlist.py` tests passing

tlambert03:
> I think the conflict right now is between using blockers during the "move like operation" and overwriting internal methods as the blockers prevent the listeners from responding but the internal method still gets called as before. To the outside it looks like it was just "a move" which is what we want, but to the inside it looks like it was a bunch of removes and a bunch of adds, and now things have "gotten out of sync"

I'm able to reproduce the aberrant signal emission and method calling in qt_layerlist that you saw in https://github.com/sofroniewn/napari/pull/6#discussion_r460460096. Let's figure out what's going on there before worrying about the upstream stuff.

I think maybe you should hold off on the layerlist for now.  I have pretty much all of this working well in #1089, so I feel like there's a good chance that most of the issues here are on the Qt implementation side, and how it's using the model API (which may now be emitting events that the last one was skipping due to the aforementioned "leaks") ... but I need some time to sit down and figure out what's going on.  I feel like the list model itself is very robust and clean at this point, and I don't want to start making edits to the EventedList model without knowing more about how our other objects are trying to use the model... and whether perhaps we need to make improvements there (particularly because it's working pretty well in #1089 ... which I see as the more important long term thing anyway). 

I agree that the reordered event is probably unnecessary... that was mostly there as a placeholder to try to match one of the previous events.  I _don't_ however think the moved events are unnecessary.  If you look at the signals offered by [`QAbstractItemModel`](https://doc.qt.io/qt-5/qabstractitemmodel.html#protected-functions), you'll see "inserted", "removed" and "moved"... I've actually been digging through the Qt source code as well and the purpose for each signal is quite clear.  The trick on the view side is to make sure that a true move event is not incurring any removed or inserted events.  This is what I have working nicely in #1089, and I'm still confident that it's possible.

tlambert03:
ok @sofroniewn, thanks for your patience, and thanks for working through the code with me.  I was able to fix the bug in #1444 that caused `del x[0:0]` to delete the full array, and I also pulled in your `blocker_all` context into #1444 ... I added tests for both things, so it should behave better for you here.

the `move_selected` block we came to earlier was:
```python
        if not self[index].selected:
            self.unselect_all()
            self[index].selected = True
            moving = (index,)
        else:
            moving = [i for i, item in enumerate(self) if item.selected]
        with self.remember_layers():
            self.move_multiple(moving, insert)
```

looks like I didn't save the `remember_layers` context manager I made... but it was something like this:
```python
def __init__(self):
    self._remembered_layers = set()
    super().__init() 
    ...

@contextmanager
def remember_layers(self):
    self._remembered_layers = set(self)
    yield
    self._remembered_layers.clear()

def insert(self, index, layer):
    if layer not in self._remembered_layers:
        self._prepare_layer_for_add(layer)
    super()
    ...
    if layer not in self._remembered_layers:
        # do the selection update
```

I'm also definitely not sure that's the best way we can do that... but it might get things working for now without having to connect events in `__init__`

sofroniewn:
Thanks @tlambert03 I've updated the evented list and am using the new `move_selected` block from above. I did have to change one or two of the `move_selected`  tests which we dealing with a previously ambiguous case about moving multiple things right next to where one already is, but I'm pretty confident everything performs now as it should in all cases.

I still havn't adopted the `remember_layers` syntax or removed the events connected in the layerlist itself. Will likely do that tomorrow.

napari/components/layerlist.py
```diff
@@ -98,23 +96,14 @@ def move_selected(self, index, insert):
...
-        offset = sum([i < insert and i != index for i in selected])
-        # insert indices to be moved at correct start
-        for insert_idx, elem_idx in enumerate(selected, start=insert - offset):
-            indices.insert(insert_idx, elem_idx)
-        # reorder list
-        self.move_multiple(indices, 0)
+            moving = (index,)
+        else:
+            moving = [i for i, item in enumerate(self) if item.selected]
+        offset = insert >= index
```
sofroniewn, line 88: I had to add this offset in here, but I think it might actually belong in `move_multiple` @tlambert03 - look at `viewer.layers.move_multiple((0, 1), 2)` I think it doesn't do the right thing, it's off by one, but it's late now so I might be wrong about this!!! Will look more tomorrow

tlambert03: we can certainly change this, but currently `move_multiple((0, 1), 2)` should be a no-op... it's saying "insert the items at index 0 and 1 right before the item at index 2" (And they already are just before index 2)... just like `a.insert(2, 'something')` is saying "insert 'something' right before the item at index 2".  
```python
a = EventedList(range(10))
a.move_multiple((0, 1), 0)
a.move_multiple((0, 1), 1)
a.move_multiple((0, 1), 2)
a == list(range(10)  # True ... hasn't changed
```
thoughts?

sofroniewn: > it's saying "insert the items at index 0 and 1 right before the item at index 2" (And they already are just before index 2)... just like a.insert(2, 'something') is saying "insert 'something' right before the item at index 2".

Interesting, if we just think about the `a.move` case I would have expected `a.move(old_index, new_index)` to have meant that `a[new_index]` after the move has the same value as `a[old_index]` before the move, but this is not generally the case. Note that for the `a.insert(index, 'something')` example it is always the case that `a[index]='something'` after the insertion. I would not have expected `a.move(4, 5)` to be a no-op, but to move the item at 4 to be at 5.


Similarly, for example if
```python
a = TypedEventedList(['a', 'b', 'c', 'd', 'e', 'f', 'g'])
a[1] # `b`
a.move(1, 3)
a[3] # `d` (I would have expected `b`, but that is at a[2]) 
```
but note though the following
```python
a = TypedEventedList(['a', 'b', 'c', 'd', 'e', 'f', 'g'])
a[5] # `f`
a.move(5, 3)
a[5] # `f` (which is what I would expected)
```
The difference here is something is moving to a lower index. I think for the `a.move` case you do need an `offset = old_index < index`.

If the `a.move` behaves this way, then it's natural for `a.move_multiple` to behave similarly, I guess the `a.move_multiple((old_index,), new_index)` should always behave the same as `a.move(old_index, new_index)` and then the question is for `a.move_multiple((old_index1, old_index2), new_index)` with `old_index1 < old_index2` I guess one has to make a choice and I think it's fine to say after the move `a[new_index]` should have the value of `a[old_index2]` before the move.

In your example about then
```python
a = EventedList(range(10))
a.move_multiple((0, 1), 0)
a.move_multiple((0, 1), 1)
a == list(range(10)  # True ... hasn't changed
a.move_multiple((0, 1), 2)
a == [2, 0, 1] + range(3, 10) # Now the list has changed and a[2] == 1
```

I'll note that all of this was just how I would have guessed it was working, so if there is compelling prior art or other reasons to do it with the "insert right before the item at index 2" that's an option too, but curious how you find my reasoning.

tlambert03: I agree, it's weird at first... and I also wanted those simple cases to have the `dest_index` be in post-move space, and I understand it flips if you're moving up vs moving back ... but I went with having the user provide "pre-move" coordinates instead of "post-move" coordinates following patterns I saw elsewhere (such as Qt's [MoveRows/BeginMoveRows](https://doc.qt.io/qt-5/qabstractitemmodel.html#beginMoveRows)), and here's why I was ultimately swayed:

first, I totally agree that `move_multiple((old,), new)` has to equal `move(old, new)`... and since move_multiple is much harder, let's take that case.  It's simple if you think about adjacent objects, like `move_multiple((0,1), 2)` ... but it's much harder if you consider that the moving objects needn't be adjacent. 
```python
a = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
a.move_multiple((2, 3, 7, 9), 5)
```
If we ask the user to provide pre-move coordinates, they look at the current list and say "i want the 2,3,7 & 9 to move together to right before the 5".  That's very much like what a drag and drop would do, where you have a handful of collected/dragged pre-move indices, and then a specific position (in pre move space) where they release the mouse and drop things.  so this hooks up to that use case nicely.
If we ask the user to provide post-move coordinates... then _they_ have to do all that math and say "ok... I'm moving the 2 and the 3 back, but the 7 and the 9 up ... so that means that after the move I guess I want them to be at ...", and then they're going to have to change that logic depending on whether part of the selection is before the destination or not.

So providing pre-move coordinates sometimes yields the undesirable effect that `dest_index` is not always where the objects land after the move, but using post-move coordinates requires being careful whenever you call the function to make sure that you call it just right depending on exactly what you're moving, making the API a bit more challenging to use.

tlambert03: I'll also point out that if you want to know where they land after the move, the `moved` event has an `event.new_index` value

tlambert03: and actually, currently the `move_multiple` function returns the number of successful moves (which is probably always just going to be the number of objects).  If we stick with the current pre-move space input, maybe we should have the function return the destination index in post-move space?

sofroniewn: > If we ask the user to provide post-move coordinates... then they have to do all that math and say "ok... I'm moving the 2 and the 3 back, but the 7 and the 9 up ... so that means that after the move I guess I want them to be at ...", and then they're going to have to change that logic depending on whether part of the selection is before the destination or not.

I find this compelling

> and actually, currently the move_multiple function returns the number of successful moves (which is probably always just going to be the number of objects). If we stick with the current pre-move space input, maybe we should have the function return the destination index in post-move space?

yes this might be nicer, but i don't really have a strong feeling on this one.

ok - this all checks out, I see the reasoning for it. As long as it's all clearly documented I think users will figure it out pretty quickly. Definitely @jni should give this stuff a play before it all merges, and I'm curious how he'll view this, but I think this is as fine a choice as any.

tlambert03: cool.  I'm also realizing that we really don't need both `move()` and `move_multiple()` ... rather we can just check if the first argument is iterable or not and use the move_multiple logic if so... yeah?

sofroniewn: > I'm also realizing that we really don't need both move() and move_multiple() ... rather we can just check if the first argument is iterable or not and use the move_multiple logic if so... yeah?

Yes! wow - definitely just having move would be amazing!! I also think if second arg not provided then it defaults to `0`!! So clean!!!

sofroniewn:
I'd like to pick this up now as it will fix #1829 and #1380 I think.

I merged in master the other day to fix merge conflicts, but I'm wondering if I messed up somehow because I've got a couple additional file in the `utils/events/containers` folder and some of my imports aren't found anymore. Do you remember @tlambert03 if you added things to utils events specifically in this PR that weren't in #1444? Or if can just make this PR have the same `utils/events/` folder than do I still want to use `from ..utils.events import TypedEventedList`? I'm not sure that exists anymore. Any advice/ help much appreciated

napari/components/layerlist.py
sofroniewn: just rereading this thread now @tlambert03 - it sounds like this comment https://github.com/napari/napari/pull/1504/files#r460995503 was a good idea, but hasn't been implemented yet. Nothing to hold up this PR for, but wanted to flag so we don't forget the good idea. I can make an issue too

sofroniewn: issue made https://github.com/napari/napari/issues/1845

sofroniewn:
This PR will close #1380 and is now ready for review @jni @tlambert03. I've made a new issue in #1845 to discuss dropping `move_multiple`. It's been quite some time since this was originally worked on and so I want to make sure nothing silly slipped through.

We're no longer using the anything in `utils/list` and so could make a followup PR to drop that code after this merges. https://github.com/napari/napari/tree/master/napari/utils/list

napari/components/layerlist.py
```diff
@@ -6,49 +6,28 @@
...
-            iterable=iterable,
-            lookup={str: lambda q, e: q == e.name},
+            data=data, basetype=Layer, lookup={str: lambda e: e.name},
         )
 
-        self.events.added.connect(_add)
-
-    def __newlike__(self, iterable):
-        return ListModel(self._basetype, iterable, self._lookup)
+    def __newlike__(self, data):
```
tlambert03, line 29: I don't think this is necessary anymore ... the inherited method from `TypedMutableSequence` is the same

napari/components/viewer_model.py
```diff
@@ -95,9 +95,12 @@ def __init__(
...
         self.cursor.events.position.connect(self._on_cursor_position_change)
-        self.layers.events.changed.connect(self._update_active_layer)
-        self.layers.events.changed.connect(self._on_grid_change)
-        self.layers.events.changed.connect(self._on_layers_change)
+        self.layers.events.inserted.connect(self._on_grid_change)
+        self.layers.events.removed.connect(self._on_grid_change)
+        self.layers.events.reordered.connect(self._on_grid_change)
+        self.layers.events.inserted.connect(self._on_layers_change)
+        self.layers.events.removed.connect(self._on_layers_change)
+        self.layers.events.reordered.connect(self._on_layers_change)
```
tlambert03, line 103: this is a bit unfortunate huh?  I remember `changed` sometimes being problematic, but I can't remember why.  do you know why this is necessary?

napari/components/viewer_model.py
```diff
@@ -479,7 +483,7 @@ def grid_view(self, n_row=None, n_column=None, stride=1):
         self.grid.enabled = True
 
     def stack_view(self):
-        """Arrange the current layers in a stack.
+        """Arrange the current layers is a stack.
```
tlambert03: ?

napari/utils/transforms/transforms.py
```diff
@@ -81,12 +81,12 @@ def expand_dims(self, axes: Sequence[int]) -> 'Transform':
...
 
 
-class TransformChain(ListModel, Transform):
+class TransformChain(EventedList, Transform):
     def __init__(self, transforms=[]):
         super().__init__(
             basetype=Transform,
-            iterable=transforms,
-            lookup={str: lambda q, e: q == e.name},
+            data=transforms,
```
tlambert03, line 87: just to stick with the (new) convention, maybe put `data` before `basetype`?

napari/components/_tests/test_layers_list.py
```diff
@@ -50,6 +50,16 @@ def test_removing_layer():
     assert len(layers) == 0
 
 
+def test_popping_layer():
```
tlambert03, line 53: this test seems unnecessary since you're not overriding the `pop` method and we already test all of the standard list methods on `EventedList` [here](https://github.com/napari/napari/blob/f0cec79f9de5c5fbd2153a1614b0df2e030e93ae/napari/utils/events/_tests/test_evented_list.py#L42)

napari/_qt/widgets/_tests/test_qt_layerlist.py
```diff
@@ -160,6 +160,33 @@ def test_removing_layers(qtbot):
...
+    qtbot.addWidget(view)
+
+    layer_a = Image(np.random.random((10, 10)))
+    layer_b = Image(np.random.random((15, 15)))
+    layer_c = Image(np.random.random((15, 15)))
+    layer_d = Image(np.random.random((15, 15)))
+    layers.append(layer_a)
+    layers.append(layer_b)
+    layers.append(layer_c)
+    layers.append(layer_d)
```
tlambert03:
```suggestion
    layers.extend([Image(np.random.random((15, 15))) for _ in range(4)])
```
... or is it important that the first layer be `(10, 10)`?

tlambert03 (approved):
looks good.  only a couple small comments

napari/components/layerlist.py
sofroniewn: If I delete this I get the following error message, but agreed would be nice to be able to drop

```python-traceback
(base) czirwc1macos2701:napari nsofroniew$ python examples/add_image.py
ERROR:root:Unhandled exception:
  File "/Users/nsofroniew/GitHub/napari/napari/_qt/event_loop.py", line 79, in gui_qt
    yield app
  File "examples/add_image.py", line 11, in <module>
    viewer = napari.view_image(data.astronaut(), rgb=True)
  File "/Users/nsofroniew/GitHub/napari/napari/view_layers.py", line 9, in view_image
    (where <layer_type> is replaced with one of the layer types):
  File "/Users/nsofroniew/GitHub/napari/napari/components/add_layers_mixin.py", line 267, in add_image
    return self.add_layer(image_class(data, **kwargs))
  File "/Users/nsofroniew/GitHub/napari/napari/components/add_layers_mixin.py", line 64, in add_layer
    self.layers.append(layer)
  File "/Users/nsofroniew/opt/anaconda3/lib/python3.7/_collections_abc.py", line 971, in append
    self.insert(len(self), value)
  File "/Users/nsofroniew/GitHub/napari/napari/utils/events/containers/_evented_list.py", line 159, in insert
    self.events.inserted(index=index, value=value)
  File "/Users/nsofroniew/GitHub/napari/napari/utils/events/event.py", line 514, in __call__
  File "/Users/nsofroniew/GitHub/napari/napari/utils/events/event.py", line 535, in _invoke_callback
    cb_event=(cb, event),
  File "/Users/nsofroniew/GitHub/napari/napari/utils/events/event.py", line 529, in _invoke_callback
    cb(event)
  File "/Users/nsofroniew/GitHub/napari/napari/components/viewer_model.py", line 445, in _on_grid_change
    for i, layer in enumerate(self.layers[::-1]):
  File "/Users/nsofroniew/GitHub/napari/napari/utils/events/containers/_typed.py", line 124, in __getitem__
    return self.__newlike__(result) if isinstance(result, list) else result
  File "/Users/nsofroniew/GitHub/napari/napari/utils/events/containers/_typed.py", line 142, in __newlike__
    iterable, basetype=self._basetypes, lookup=self._lookup
TypeError: __init__() got an unexpected keyword argument 'basetype'
```

napari/components/viewer_model.py
sofroniewn: i think made `changed` emit when a value gets changed, i.e. `layer[0] = new_value` and it doesn't get emitted on `inserted` or `removed` etc. I just tried switching and it doesn't work. It's not ideal maybe, but I can live with it

napari/components/viewer_model.py
sofroniewn: lol, not worth explaining why that even appeared!!! (me dealing with merge conflicts)

napari/utils/transforms/transforms.py
sofroniewn: done

napari/components/_tests/test_layers_list.py
sofroniewn: That's fair, and i can see why we don't want our tests to proliferate, but I added because we get asked a lot about how to delete layers from the layer list and so wanted to make doubly sure that works as expected

napari/components/add_layers_mixin.py
```diff
@@ -58,7 +58,11 @@ def add_layer(self, layer: layers.Layer) -> layers.Layer:
         layer.events.cursor.connect(self._update_cursor)
         layer.events.cursor_size.connect(self._update_cursor_size)
         layer.events.data.connect(self._on_layers_change)
+        layer.name = self.layers._coerce_name(layer.name, layer)
```
sofroniewn, line 61: I just want to note one design flaw we have right now which this PR isn't really making worse but just exposing further - if you call `viewer.layers.append(layer)` instead of `viewer.add_layer(layer)` than none of this wiring up happens correctly, and now the name coercing won't happen either. Not sure if we should try and address this now @tlambert03 @jni?

The ways I can see it are we pull out all the event connection into its own method that gets triggered via events either when calling `add_layer` or `layers.append`, but this logic of events triggering the connection of events is really ugly to maintain, especially in this case where we have to deal with `layers.inserted` which gets emitted during a layer move too, or we work to eliminate the need for all these events, but that will be hard

sofroniewn:
Thanks for the review @tlambert03 - I've mostly made the requested changes and left a couple comments. I've also posed a new question here https://github.com/napari/napari/pull/1504#discussion_r519191685 about how our `layers.append` method can't really be used with the viewer right now as all our events don't get wired up properly.

napari/components/add_layers_mixin.py
tlambert03: yeah, that `AddLayersMixin.add_layer` method has been something that I've wanted to see disappear for a while.  I would like to see a situation where the viewer only listens to a `LayerList`, and simply inserting/removing layers from that list is sufficient. (that's going to be important if the `LayerList` is to eventually be our "unit" of a group of layers... layers shouldn't be connected both to the viewer and the LayerList, but rather layer to List and List to viewer)

If you've been able to get everything working here as it was before, then I think all that should be addressed in a later PR (since it's not a "new" problem)

napari/components/viewer_model.py
tlambert03: ah right.  yeah, I guess we could reconsider the "meaning" of changed, and emit it all the time.  or just leave it as is.  fine by me either way

napari/components/add_layers_mixin.py
sofroniewn: Ok nice, we see this the same way. The one change this PR would make right now is the coercing of uniqueness in the name would only happen on `viewer.layers` and making only the last layer selected would only happen on `viewer.layers`, but otherwise as you say this is not a new problem. If we can live with that then I'll make a new issue right now describing this and we can merge this PR without anything more here

napari/components/layerlist.py
tlambert03: oh I'm sorry, of course.  the `__init__` has been simplified.  my bad.  needs to stay

napari/components/add_layers_mixin.py
sofroniewn: issue made here https://github.com/napari/napari/issues/1846

jni (approved):
@sofroniewn I have only given this a cursory look but I agree with everything @tlambert03 said, this does not make things worse so let's get it in!

PR #1505, opened 2020-07-25
format imports with isort

tlambert03:
# Description
this PR would add [iSort](https://github.com/timothycrosley/isort) to our pre-commit hooks:

**arguments for:**
- automatically organized/standardized import orders
- reduce formatting variability across developers (like with black)
- easier diffs in PRs

**arguments against:**
- every so often, you _might_ run into a circular import if it changes the order and exposes an underlying issue.  At the moment, these are all fixed.  But you can also use `# isort: skip` to tell isort not to move a line
- like #322,  this PR touches **lots** of files.
- yet another tool.

at the moment, this isn't active in our CI (but it can be and probably should be so we don't have to rely on someone having installed pre-commit).  But wanted to get feedback first.

## Type of change

# References

# How has this been tested?

## Final checklist:


sofroniewn:
Broadly speaking I like PRs like this that reduce variability and would like us to adding things like this to CI if we can, to ensure standardization without increasing developer burden.

I see how some clear things get better - like 

```python
from qtpy.QtWidgets import QComboBox, QLabel, QSlider, QPushButton	
# Goes to
from qtpy.QtWidgets import QComboBox, QLabel, QPushButton, QSlider
```
which improves alphabetical ordering.

But if I look at files like https://github.com/napari/napari/blob/faea39c6cfbfb85a8965d3cc62971f6366fe01dd/napari/_qt/_tests/test_qt_dims.py which got somewhat rearranged the final order of things still confuses me

```python
import os
from sys import platform
from unittest.mock import patch

import numpy as np
import pytest
from qtpy.QtCore import Qt

from napari._qt.qt_dims import QtDims
from napari.components import Dims
```
It mixes `import` and `from` blocks and doesn't appear to have put everything in a global alphabetical order. Am I missing something in this example or was this to be expected? I guess the `napari` block comes at the end because it's our library? Is the top block imports from python builtins? Then second block from our dependencies? 

Seems like we have some control over things here https://github.com/timothycrosley/isort#custom-sections-and-ordering and https://github.com/timothycrosley/isort#controlling-how-isort-sections-from-imports

I'm curious how others feel about the `import` vs `from` and the builtin section/ dependency section/ our library section?

I'm actually ok with the settings we have now, but it did take me some time to wrap my head around them, and curious how others feel.

tlambert03:
> it doesn't appear to have put everything in a global alphabetical order.

yep that's the default.  builtins follow by third party, followed by internal imports.  That's keeping with the recommendations in the [PEP 8 style guide](https://www.python.org/dev/peps/pep-0008/#imports).  

> Imports should be grouped in the following order:
>
>1. Standard library imports.
>2. Related third party imports.
>3. Local application/library specific imports.
>
>You should put a blank line between each group of imports.

I'm personally fine with it/used to it... but as you point out, it can be overridden if you don't like it.

sofroniewn:
> yep that's the default. builtins follow by third party, followed by internal imports. That's keeping with the recommendations in the PEP 8 style guide.

Thanks for the clarification!!

jni:
I'm 100% in favour of the three blocks. And import/from is a style preference, I think different things make sense in different contexts, so I wouldn't mess with it or try to normalise it. Some libraries such as `dask_image` always do fully qualified imports (`import dask.array` ... then `dask.array.sum()`), and I really hate it.

I'm less sure about isort in general but I haven't looked at the diff yet. @tlambert03 you've done a good job listing the pros and cons, and since I am an old-school dev I lean more in favour of the cons, but at the same time, even though I dislike black, it *has* been nice to never have a code style discussion. LOL. I'll try to have a look at the diff soon.

jni (commented):
@tlambert03 sorry for the slowness! I've gone through the diff and I'm ok with this. One Q I have is the config, "known third party" — does this mean we need to update that any time we add a dependency?

Also, since this PR will basically cause merge conflicts with most other PRs, it'd be good to coordinate whether there are other things we want to merge before this one.

Finally, some of the repos listed in the pre-commit config are in personal github accounts... Should we instead depend on versions published to PyPI?

jni (approved):
LOL meant to hit approve. 😅

tlambert03:
> One Q I have is the config, "known third party" — does this mean we need to update that any time we add a dependency?

that setting _was_ automatically provided/updated by [`seed-isort-config`](https://github.com/asottile-archive/seed-isort-config) ... but I just realized it's no longer needed with the most recent version of isort... so it's gone now... and no we don't need to update it.

> Also, since this PR will basically cause merge conflicts with most other PRs, it'd be good to coordinate whether there are other things we want to merge before this one.

yeah... def.  I'm more than happy to handle all the merge conflicts for people too.  In most cases, it will just mean accepting "current" in case anything was added/removed, and then re-rerunning isort

> Finally, some of the repos listed in the pre-commit config are in personal github accounts... Should we instead depend on versions published to PyPI?

good question... That's how I've always seen it with pre-commit! ([even in the docs](https://pre-commit.com/#2-add-a-pre-commit-configuration)).  I can't even (quickly) find examples of using directly from PyPI... so I guess I'm gonna say if it ain't broke? 🤷‍♂️

sofroniewn (approved):
This LGTM!

Not sure if it's in this PR (I didn't see it, but it can also come in a followup) but can we add some info on ISort to our contributing guide here https://napari.org/docs/developers/CONTRIBUTING.html#setting-up-a-development-installation if it is now in our pre-commit (we might want to just make a section on pre-commit hooks) and to our further resources here https://napari.org/docs/developers/CORE_DEV_GUIDE.html#further-resources

I'm all for this change, and think it will be great, but I think we should make sure we're not adding barriers to contributing by depending on too many dev tools, and having clear dev documentation is a way to avoid that.

sofroniewn:
Merging now, thanks for updating the docs!

sofroniewn:
@tlambert03 should `iSort` be listed in our dev requirements now?

tlambert03:
Sure I can add that... but it's not technically necessary for pre-commit to work... it does that install into its own environment based on the urls in the pre-commit yaml

sofroniewn:
> Sure I can add that... but it's not technically necessary for pre-commit to work... it does that install into its own environment based on the urls in the pre-commit yaml

Ahhh thanks for the clarification, no need to then. Is there a reason why we have black/ flake8 in there though? Apologies for the basic questions, clearly I've never really gotten this stuff, I was just looking at consistency

jni:
> Apologies for the basic questions,

I have the same questions! :sweat_smile:

tlambert03:
>  Is there a reason why we have black/ flake8 in there though?

all I can tell you is that they don't need to be there for pre-commit :)
I just double checked to make sure I'm giving you the right info... when you add a "repo" to the pre-commit config, it will download that repo and store it for you in `~/.config/pre-commit/`.  (So this is independent of a project, and can be used locally across projects).  When you use `pre-commit run`, it makes sure it has the necessary repos, and if not, it downloads and you'll see something like this:

```
❯ pre-commit run --all-files
[INFO] Initializing environment for https://github.com/timothycrosley/isort.
[INFO] Initializing environment for https://github.com/python/black.
[INFO] Initializing environment for https://gitlab.com/pycqa/flake8.
[INFO] Installing environment for https://github.com/timothycrosley/isort.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
[INFO] Installing environment for https://github.com/python/black.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
[INFO] Installing environment for https://gitlab.com/pycqa/flake8.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
```

so it doesn't matter if you have them in your requirements.txt ... and it only cares about the versions declared in pre-commit-config.  If you change the version, it will download and create a new cache.  To force a rebuild of repos, you can use `pre-commit clean` and it will delete the files in `~/.cache/pre-commit`

PR #1509, opened 2020-07-29
Async loading of Image layer

### LINKED ISSUE #1320: Blocked UI while loading images in notebook

pwinston:
# Description

Experimental PR, closed.


pwinston:
Close this didn't work. Leaving #1495 open for now, same changes.

PR #1510, opened 2020-07-29
Async loading of Image layer

pwinston:
# Description
The PR introduces **asynchronous loading** of images to address #1320 and other situations where the GUI thread was being blocked. Some situations that were 1-2Hz are now 20Hz.

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

**New**: 1/2 hour [walkthrough video](https://youtu.be/Yiw3oXTI6UA) of this PR.
**See also**: the older 11 minute [demo video](https://youtu.be/Jlm_jGRwH2Y) of async with perfmon traces.

# Single-scale, multi-scale, and chunks

The PR includes support for async loading of **single** and **multi-scale** images, but there is no support for chunked images or breaking up big images into chunks. 

The PR does introduce a `ChunkLoader`, but for this PR an`ImageSlice` is always a single chunk. Although realize with multi-scale one `ImageSlice` might only be small portion of the image, if you are zoomed in, but that portion will load in as one chunk.

To solve #845 and #1300 in a future PR we plan to introduce an octree. Then a single `ImageSlice` might consist of many chunks. Then you will get the standard "tiled image viewer" effect where tiles (2D) or sub-volumes (3D) of an image will load in one at a time, like a patchwork.

# TODO

- [ ] More real-world testing on various datasets
- [ ] Special async tests that test for stability with random load times

# How To Use

There are two ways to enable async loading:

1. Define env var `NAPARI_ASYNC=1` to get the defaults which are currently:

```
DEFAULT_ASYNC_CONFIG = {
    "synchronous": False,
    "num_workers": 6,
    "log_path": None,
    "use_procesess": False,
    "delay_seconds": 0.1,
    "load_seconds": 0,
}
```

2. Define env var `NAPARI_ASYNC=/some/path/async-config.json` to configure settings above.

# Results 

* Short 30 second [YouTube Video](https://youtu.be/mdy0RofUiLw) of this [noise2self script](https://github.com/pwinston/napari-pwinston/blob/master/noise2self/noise2self_script.py) running with async processes.

* Longer 11 minute [YouTube Video](https://youtu.be/Jlm_jGRwH2Y) showing the steps we took including perfmon traces.

Below is a `chrome://tracing` view showing the background process doing the Torch (Machine Learning) computation while rendering proceeds unblocked:

87239376-7caaa100-c3dc-11ea-9846-f0258f82b6ce

# How It Works

[...]

Issue #1395, opened 2020-06-30
Error Handling Design Proposal


liu-ziyang:
## Motivation
There are a lot of reported issues on the napari repo related to error handling. We need to have a unified solution for error reporting and logging, including the plugin error handling/logging.

## Meta Analysis
Expanding on the work and analysis of the [event handler](https://github.com/napari/napari/pull/1040), I found that the current flow of error handling is not intuitive for a few reasons:
- The error handling framework is now largely built on top of qt event loop, we leverage qt to deal with error and make sure exceptions are caught and blocked on malfunctioning code path/user action. Resulting in the lack of a middle layer to capture errors, and many internal errors propagate up the stack, and hence hard to enforce a MVC methodology since the flow itself is not controller centric.
- Developers would not have a sense of safety without comprehensive understanding of the existing codebase, they would not be able to declare and raise exceptions as freely in their code. It can be overwhelming to write code that may raise an exception when the consequences are unknown (would it crash the application if an exception is raised here?), even if throwing an exception is the right thing to do. 
- There is no consensus of how errors should be handled, resulting in possible dispersion of code style and different practices when coming down to logging, notifications, etc.

## Case Analysis
- [Bug report](https://github.com/napari/napari/issues/1349): SystemError on exit with perfmon and PyQt5
- [Bug report](https://github.com/napari/napari/issues/1336): Napari crash on Windows 10 - QT issue?
- [Bug report](https://github.com/napari/napari/issues/1347): Napari 0.3.4 release crashes when switching to volume rendering

## Design principles:
- Isolating error handling from the view layer, continuing decoupling of view class to avoid crashing on errors. The end goal is to have the UI(view) to be bullet-proof, regardless of errors going on behind the scene (model layers) on the controller side, and unify handling of such errors to the controller end, which gives peace of mind to both view and model implementations. Rely on the Qt error handling only for view related issues
- Enhance error message propagation, and make them visible/interactive to the user. The end goal is to make sure the user understands what went wrong, and how to proceed if possible. Such error messages should be displayed consistently across different types of errors, too.
- Provide unified error handling flow to be used consistently throughout the codebase, defines and extend logging functions to be used across codebase to track errors, while using standard log level to control robustness of such logging
 
## Design considerations
- Scope of the work: These changes are going to be long term changes unlikely addressed by a single PR. We need to carefully scope the steps to make sure we are heading in the right direction without stepping on our toes
[...]

PR #1511, opened 2020-07-29
add error dialog for global error handler

liu-ziyang:
# Description
Add error dialog for the global error handler

## Type of change

# References
https://github.com/napari/napari/pull/794/files

# How has this been tested?

## Final checklist:


liu-ziyang:
new look 
Screen Shot 2020-08-04 at 10 46 15 AM

liu-ziyang:
Courtesy to Lia: resources useful for updating visual: 
warning-triangle-#FF8761-100px
close-x-#D1D2D4-100px
warning-circle-#FFB661-100px

tlambert03:
I pushed some changes... still plenty left to do, but this fixes the relative positioning and adds a fade/slide in animation, and a fade out that happens automatically after 5 seconds.  This gif makes it look choppier than it is... but here's a preview:

Untitled

still to be done:
- prevent the automatic closing if the user interacts with the error "somehow" (maybe if they are hovering over it?)
- update the position of the error message if the viewer is resized while the message is showing.
- improve styling (fonts, font-color, close button hover effect, possibly icons...)

I'm also kind liking the way vscode has a short message with a little up arrow for "show more", and a close button on the right, instead of the close button on top left.
Screen Shot 2020-08-06 at 4 37 26 PM

@liaprins-czi, @sofroniewn ?

liu-ziyang:
Thanks for the update @tlambert03 
I am running into some warnings locally when I click on the X to close the message, and the message won't go away: 
```
WARNING: QPropertyAnimation::updateState (opacity): Changing state of an animation without target
WARNING:vispy:QPropertyAnimation::updateState (opacity): Changing state of an animation without target
WARNING: QPropertyAnimation::updateState (opacity): Changing state of an animation without target
WARNING:vispy:QPropertyAnimation::updateState (opacity): Changing state of an animation without target
```

tlambert03:
might need to keep a pointer to the graphics effect... try again?

liu-ziyang:
> might need to keep a pointer to the graphics effect... try again?

works like a charm, thanks and good job!

tlambert03:
made some progress on this, here's where it's at now:

Untitled

The "do something" button is optional, and corresponds to the `actions` parameter of the new `NapariNotification` object (which takes a sequence of tuples, with the name for the button, and a callback when clicked).

```python
class NapariNotification(QDialog):
    def __init__(
        self,
        message: str,
        severity: NotificationSeverity = NotificationSeverity.INFO,
        source: str = None,
        actions: Sequence[Tuple[str, Callable[[], None]]] = (),
    ):
```
the notification also takes a severity that will correspond to the various icons shown.  
I also addressed the issues mentioned before related to styles and resizing.  Currently the message will auto-hide after 5 seconds, unless the user hovers over it, in which case they have to close by pressing the "X" button.

tlambert03:
this is probably ready for some people to kick the tires a bit.  Styles and behavior are decent.  Still to be done (probably in another PR):

- handle multiple viewers: currently, this just puts the notification in the first top level QMainWindow it finds
- handle multiple notifications: currently if a second notification appears before the first has disappeared, they will just appear on top of each other.  They should push each other up, and be aware of their height/expanded state.  Probably with a QListWidget container.
- minor issue resulting from using the canvas as the parent: if the canvas is smaller than the widget of the notification (~400 pixels). the left edge of the notification will be occluded by the controls on the left.

sofroniewn:
This is coming along so nicely!! Great work @tlambert03! I just gave it a spin. A couple observations and thoughts:

First - I'd reduce the time to fade out. I think it is ~5 seconds now, I might reduce to ~3 (one day that can be a user setting) and one day we can maybe have a little error icon in the bottom right that if you click on shows all the previous errors somehow.

I found it a little surprising that hovering over the widget then meant I had to dismiss it manually. Maybe one option would be when hovered the boundary can change (maybe becomes a different color) to indicate that it now needs to be dismissed manually (i.e. it somehow looks highlighted). It could also change on click not just hover.

I was able to expand the widget by double clicking, but it took some time to discover. I wonder if a single click should expand. A single click could also trigger the highlight mentioned above and the need to explicitly dismiss.

Looking at the screenshot below I found the icon on the left hand side with the circle and the x look too much like a "close button" rather than a warning or error symbol. I think the exclamation mark in either a triangle or circle would be nicer. We could have yellow triangle exclamation for warning/ minor error, red circle exclamation for severe error which requires interactivity.

Screen Shot 2020-08-08 at 12 55 18 PM

tlambert03:
> First - I'd reduce the time to fade out. I think it is ~5 seconds now, I might reduce to ~3

sure, can do.

> one day we can maybe have a little error icon in the bottom right that if you click on shows all the previous errors somehow.

definitely... like the little "bell" in VScode :)

Screen Shot 2020-08-08 at 4 06 21 PM

> I found it a little surprising that hovering over the widget then meant I had to dismiss it manually...It could also change on click not just hover.

I can see this going either way.  I was reasoning that, given the relatively small area of the notification, what are the chances that they are going to have their mouse over it if they're not attending to it?  And if they _are_ attending to it, it's probably more frustrating to have it disappear unexpectedly than it is to have to click the "X" icon. (at least, until we have some sort of "show previous errors" function).  @liaprins-czi, thoughts on this?  There already is a hover effect on there, so at least it's clear that their mouse hover "has been noted".

> I wonder if a single click should expand (instead of a double click).

could do... though we might want to reserve a single click for "selection" (if we eventually allow for notifications to stack up in a list as in VSCode).  I took that double click behavior from VScode, where single click is used for selection.

> Looking at the screenshot below I found the icon on the left hand side with the circle and the x look too much like a "close button" rather than a warning or error symbol. I think the exclamation mark in either a triangle or circle would be nicer.

sure, can change that up.  Note: that's the "error" icon (also borrowed from VScode), the exclamation mark in a triangle is in there too... but it's for the "warning" level.  Giving these things colors is also possible, but does require us to add some features to the way we build/theme icons.  Currently, _all_ icons are themed light or dark ahead of time (i.e. we can't dynamically give our SVGs color).  So we'd need to add some new machinery to the resource building process.  Definitely doable, but probably better for another PR.

tlambert03:
also @sofroniewn, there should be a little chevron icon to expand the panel... but I don't see it in your screenshot.  did you ever see it?  If not, you might need to throw away your `qt_resources.py` to force it to rebuild

sofroniewn:
> There already is a hover effect on there, so at least it's clear that their mouse hover "has been noted".

Ah I missed it! It is quite subtle, the whole thing changes color slightly. We definitely don't have to get stuck here, but maybe just changing boundary would make it more noticeable? Or maybe a more dramatic color change?

> could do... though we might want to reserve a single click for "selection" (if we eventually allow for notifications to stack up in a list as in VSCode). I took that double click behavior from VScode, where single click is used for selection.

Ok that's fair

> sure, can change that up. Note: that's the "error" icon (also borrowed from VScode), the exclamation mark in a triangle is in there too... but it's for the "warning" level. Giving these things colors is also possible, but does require us to add some features to the way we build/theme icons. Currently, all icons are themed light or dark ahead of time (i.e. we can't dynamically give our SVGs color). So we'd need to add some new machinery to the resource building process. Definitely doable, but probably better for another PR.

fair too. let's save the custom coloring for another PR, but I still might prefer the `!` at least for now. that's easy to improve as we go

> also @sofroniewn, there should be a little chevron icon to expand the panel... but I don't see it in your screenshot. did you ever see it? If not, you might need to throw away your qt_resources.py to force it to rebuild

Nope I didn't - but I did get these
```python-traceback
WARNING: Cannot open file ':/themes/dark/chevron_up.svg', because: No such file or directory
WARNING:vispy:Cannot open file ':/themes/dark/chevron_up.svg', because: No such file or directory
WARNING: Cannot open file ':/themes/dark/chevron_up.svg', because: No such file or directory
WARNING:vispy:Cannot open file ':/themes/dark/chevron_up.svg', because: No such file or directory
WARNING: Cannot open file ':/themes/dark/chevron_up.svg', because: No such file or directory
WARNING:vispy:Cannot open file ':/themes/dark/chevron_up.svg', because: No such file or directory
WARNING: Cannot open file ':/themes/dark/chevron_up.svg', because: No such file or directory
WARNING:vispy:Cannot open file ':/themes/dark/chevron_up.svg', because: No such file or directory
```

tlambert03:
> WARNING: Cannot open file ':/themes/dark/chevron_up.svg', because: No such file or directory

yeah... `chevron_up.svg` is in this PR.  run `rm napari/resources/_qt_resources*`

sofroniewn:
> yeah... chevron_up.svg is in this PR. run rm napari/resources/_qt_resources*

Fixed it, chevron now appears!

sofroniewn:
Hover is more noticeable but I think I'd still like some "permanent" styling change that happens after the widget transitions from one that is going to fade away to one that will need to be dismissed. If i move the mouse away the hover disappears and I assume the widget will then too, but now it just stays there.

I like the chevron and i like the warning ! triangle

Screen Shot 2020-08-08 at 1 35 40 PM

This really is coming together so nicely. It's going to be so slick :-)

tlambert03:
ok, I think this is at a decent point and ready for review.
Things for future PRs (can create a new issue assuming this merges as is):

- consider subtle style modifications to indicate that the notification must be manually dismissed (after clicking/hovering)
- this is probably useful enough that we will want to use it even without having to raise an exception and killing the flow... so we could rename `ExceptionHandler` to `NotificationHandler`, and give it a `notify()` method that could be called anywhere to generate a notification.
- fix behavior with more than one viewer.  Currently it just shows in the first TopLevel QMainWindow... we need a better way to be aware of global viewer windows, with perhaps a "current" attribute that can be queried.

napari/_qt/exceptions.py
```diff
@@ -37,9 +40,25 @@ def handle(
...
+            logging.error("Unhandled exception:\n%s", text)
+        else:
+            self._show_error_dialog(value)
+
+    def _show_error_dialog(self, exception: BaseException):
+        self.message = NapariNotification.from_exception(exception)
+        self.message.show()
+
+
+def session_is_interactive() -> bool:
```
sofroniewn: I feel like this should go in our main utils no? We might want to reuse this to fix #380 where we only call raise if session is interactive?

napari/_qt/qt_error_notification.py
```diff
@@ -0,0 +1,317 @@
...
+    def as_icon(self):
+        return {
+            self.ERROR: "ⓧ",
+            self.WARNING: "⚠️",
+            self.INFO: "ⓘ",
+            self.NONE: "",
+        }[self]
+
+
+class NapariNotification(QDialog):
```
sofroniewn, line 50: Very nice work here!!

napari/plugins/io.py
```diff
@@ -106,6 +106,7 @@ def read_data_with_plugins(
             path_repr = repr(path)
         msg = f'No plugin found capable of reading {path_repr}.'
         logger.warn(msg)
+        raise ValueError(f'No plugin found capable of reading {path_repr}.')
```
sofroniewn, line 108:
```suggestion
        raise ValueError(msg)
```

sofroniewn: Right?

sofroniewn (approved):
This looks great @tlambert03 - I'm good for this to go in as is. I just gave functionality another try and it works nicely. I've left one or two minor code comments too.

Also after merge can you add that list of follow up bullet points to #1395 so we don't loose track of them.

napari/plugins/io.py
tlambert03: ahh right.  yeah, that was one of the things that made me start writing that list "follow up" bullet points and I never got back to it.  The issue here is that by using ValueError, we prevent the logging of other errors below.  I don't think it's a big deal at this point, since ultimately all anyone cares is that nothing worked.  (and the errors will still be in the plugin error list).  But it's an issue, and it's why we need a simple non-`raise` based notification mechanism ... perhaps we could use python's own [`warnings` system](https://docs.python.org/3.8/library/logging.html#logging.captureWarnings) like we've already done for `sys.exechook`.

anyway, here we should just delete the logger above, since it gets logged in the error handler

napari/_qt/exceptions.py
tlambert03: sure

sofroniewn:
Thanks for those updates @tlambert03 - I think this is now good to go. I'll let sit for a little in case @jni or anyone else wants to weigh in, and then will merge!! Will be fun to have in

tlambert03:
just realized that the whole `session_is_interactive` thing is probably not what we want here.  Since this ErrorHandler only ever gets created if we own the event loop (and _that's_ really what we want to know), doing that check is repetitive.  Furthermore, it's problematic because if someone opens the ipython console just once, all notifications now go back to the terminal window that launched the program in the first place (not the integrated console). So I think we need to remove that...

tlambert03:
so... the real question for merging this PR becomes, are we ready for _all_ uncaught exceptions to be shown as notifications like that for now (provided one is not using IPython)?  Until we have logic that decides where to put exceptions of various types (for instance, if it derives from `NapariException` show it here, otherwise go back to the console with it), it kinda has to be all or None.

tlambert03:
I just removed the check for `session_interactive` for now... behavior is better.  the main "issue" now is that it's sorta hard for the GUI-only user to report a full traceback to us, since it won't be printed to the console anymore (which was kinda the whole point of this in the first place).  so we need to follow up with those "copy traceback" and/or "report bug" hooks that we've previously discussed pretty soon

sofroniewn:
@tlambert03 should we maybe add an environment variable to explicitly disable the GUI error handling and have all traces print as before?

The reason is that I love the functionality when I'm using napari, but when I'm developing and there's a bug I want to see the full stack trace in the terminal I launched napari from. I often do thinks like `python examples/add_image.py` and now if there is a bug it disappears, having the environment variable would give me the flexibility to have both depending on what I'm doing.

Curious what @jni thinks and if he has had a chance to try the IPython usage yet

jni:
> Curious what @jni thinks and if he has had a chance to try the IPython usage yet

Haven't tried it yet but this to me points that we want to use the error handling under exactly the same circumstances as the splash screen, so I think we should use that same logic here, and have n possible run states rather than n * m. What do you think @sofroniewn?

tlambert03:
want to point out that I added a small but very important try/except to `event_loop.py` in this PR.  This makes sure that napari doesn't crash to console if there's (even a small) exception encountered during the `yield` step in our context manager (see #1549).  If we don't merge this PR soon, I think we should pull that feature out and merge it

sofroniewn:
> Haven't tried it yet but this to me points that we want to use the error handling under exactly the same circumstances as the splash screen, so I think we should use that same logic here, and have n possible run states rather than n * m. What do you think @sofroniewn?

I'm not so sure, we made the splash screen off by default, but I'd like this on by default eventually, but a way to turn off for developers. The could just be one developer mode, but i actually think having the 2^n states is much nicer for developers. Hopefully with time this functionality improves so that all messages get routed to the correct place always, and maybe there are other ways to turn on/ off, but for now without the ability to turn this off I'd find developing on napari incredibly challenging without being able to see stack traces.

I think @pwinston has made some really great points about what it means to be able to merge code safely and try and new features. Here having an on by default temporary environment variable seems like it would strike some balance (it could also be off by default if we're worried about our users not having stack traces either).

tlambert03:
ok you can set `NAPARI_SHOW_CONSOLE_ERRORS` and it will show full tracebacks in the console.

napari/_qt/exceptions.py
```diff
@@ -37,9 +42,13 @@ def handle(
         tb : TracebackType
             The traceback object associated with the error.
         """
-        # etype.__module__ contains the module raising the error
-        # Custom exception classes can have different behavior
-        # can add custom exception handlers here ...
-        text = "".join(traceback.format_exception(etype, value, tb))
-        logging.error("Unhandled exception:\n%s", text)
         self.error.emit((etype, value, tb))
+        if session_is_interactive() or os.getenv("NAPARI_SHOW_CONSOLE_ERRORS"):
```
sofroniewn: Thanks for adding this. I'm sorry for missing something really basic here, but I'm having a hard time turning this off!!

If I run
```
export NAPARI_SHOW_CONSOLE_ERRORS=0
```
in the command line, and then run
```python
        print(os.getenv("NAPARI_SHOW_CONSOLE_ERRORS"), bool(os.getenv("NAPARI_SHOW_CONSOLE_ERRORS")))
# 0, True
```
Just before this call, then I get a value of `0`, but calling bool gives `True`. Is that maybe because it is treating that `0` like a string instead of an int?

Similarly if I do
```
export NAPARI_SHOW_CONSOLE_ERRORS=False
```
I then get
```python
print(os.getenv("NAPARI_SHOW_CONSOLE_ERRORS"), bool(os.getenv("NAPARI_SHOW_CONSOLE_ERRORS")))
# False, True
```
where my `False` is getting interpreted as True too!

Looks like in #1510 @pwinston did an explicit check `=="0"` and `=="1"` and I think that worked for me https://github.com/napari/napari/pull/1510/files#diff-33a998e174214127b7c29099bc60f479R126-R131. We might want to consider that approach here too.

tlambert03: hah.  sorry... sure I can do that.   But you can also use `unset NAPARI_SHOW_CONSOLE_ERRORS`

sofroniewn: > But you can also use unset NAPARI_SHOW_CONSOLE_ERRORS

Ah nice, thanks for the tip!! I didn't know this. I think having both `=='0'` and the unset approach will be good, thanks!!

sofroniewn: I'll add that the behaviour is now perfect and the toggling is very nice!!

jni:
@sofroniewn @tlambert03 personally I think the default should be on *unless* we launch napari with `main()`. I really think people running `python my-script.py` have a strong expectation that stack traces will go in the terminal. People launching a command, though, generally have the opposite expectation of a rather quiet console.

So my proposal if we go with an env var (which I'm ok with, it's a good point @sofroniewn) is:

| env | launch method | error catching |
|-----|----------------|----|
| NAPARI_CATCH_ERRORS=1 | any | on |
| NAPARI_CATCH_ERRORS=0 | any | off |
| none | IPython | off |
| none | script | off |
| none | `napari` | on |


@tlambert03 good job on quickly adding that error catching in the context manager. =)

**EDIT:** updated to allow an "off" switch

tlambert03:
I like that.  Do you have a suggestion for how to differentiate between script using our event loop and 'napari'?  sys.argv and endswith py?

sofroniewn:
Proposal is fine with me - I assume after the merge of #1289 the bundled app users will be on by default? I think that is important

jni:
@sofroniewn yes, I debated putting the bundled app in there but thought, nah, that is self-evident! 😂 "explicit is better than implicit" 😂 

@tlambert03 you could check/set the env variable inside `napari.__main__.main`?

tlambert03:
OK, I made some changes that meet almost all of @jni's conditions [here](https://github.com/napari/napari/pull/1511#issuecomment-674333643).  I added a `gui_exceptions=False` argument to `gui_qt()` that defaults to false, but is set to `True` when running from `napari.__main__`.  The env variable is now `NAPARI_CATCH_ERRORS`, and the main usage would be to set it to `"0"` or `"False"` when running with `python -m napari`

Setting `NAPARI_CATCH_ERRORS=1` (and even using `gui_qt(gui_exceptions=True)` to force GUI error messages regardless of the launch method is actually a bit difficult, since, for instance, we would have to do something different for IPython users vs other users (due to IPython stealing `sys.excepthook` each time).  Let's leave that for later.

> I assume after the merge of #1289 the bundled app users will be on by default? I think that is important

yep, the bundled app launches by calling `napari.__main__`

sofroniewn:
Ok I just gave this one more run through, performance is very nice, matches my expectations. We can continue to iterate as we do, thanks for pushing hard on this @tlambert03! Will merge now

PR #1517, opened 2020-08-03
Add evented colormap

### LINKED ISSUE #862: Generic colormap representation

sofroniewn:
# Description
This is a super preliminary attempt at adding our own named colormap object that is pure python (no shader code) and vispy independent. Will close #862, can up in context of https://github.com/napari/napari/pull/1475#issuecomment-665295798.

I've got a bunch of stuff still to do, so not ready for review yet, but just putting out there in case @tlambert03 wants to see. Note this includes #1475, but definitely didn't need to and all sorts of stuff isn't hooked up yet!!

## Type of change

# References

# How has this been tested?

## Final checklist:


tlambert03:
I can’t seem to find the main color map object... is this missing /utils/colormaps/colormap.py?

sofroniewn:
> I can’t seem to find the main color map object... is this missing /utils/colormaps/colormap.py?

I forgot to do git add .... done now

napari/utils/colormaps/colormap.py
```diff
@@ -0,0 +1,46 @@
...
+        Colormap interpolation mode, either `linear` or
+        `zero`.
+    """
+
+    name: str = 'undefined'
+    colors: Property[np.ndarray, None, transform_color] = np.zeros((2, 4))
+    controls: Property[np.ndarray, None, np.asarray] = np.asarray(
+        [0, 1]
+    )  # Not yet implemented
+    interpolation: str = 'linear'  # Only linear supported
```
tlambert03: huh... that is quite nice :joy:

napari/utils/colormaps/colormaps.py
```diff
@@ -438,4 +422,35 @@ def ensure_colormap_tuple(colormap: ValidColormapArg) -> Tuple[str, Colormap]:
             'Must be a {str, tuple, dict, vispy.colormap.Colormap}'
         )
 
-    return name, AVAILABLE_COLORMAPS[name]
+    return AVAILABLE_COLORMAPS[name]
+
+
+def ensure_colormap_tuple(colormap: ValidColormapArg,) -> Tuple[str, Colormap]:
```
tlambert03, line 428: will we need this anymore once we have a class that has a `name` attribute?

One thing I was also playing with the other day was mucking with the `__new__` method of the `Colormap` class, such that calling `Colormap(ValidColormapArg)` always does the right thing.  The goal there would be to be able to use `Property[ValidColormapArg, None, Colormap]` as easily as we use our `StringEnums`... thoughts?

napari/utils/colormaps/colormap.py
sofroniewn: we can reproduce all the functionality of all the colormaps we currently have (apart from the two 3D ones that I've never been such a fan of) without `controls` or `interpolation`.

I could see how they might be nice to have, and will be easy to implement, but we could also not

napari/utils/colormaps/colormaps.py
sofroniewn: > will we need this anymore once we have a class that has a name attribute?

Nope, just for backwards compatibility right now, i've only converted `Image` but will probably drop this when I'd all done

sofroniewn: > mucking with the __new__ method of the Colormap class, such that calling Colormap(ValidColormapArg) always does the right thing. The goal there would be to be able to use Property[ValidColormapArg, None, Colormap] as easily as we use our StringEnums... thoughts

This sounds quite nice!

sofroniewn: But I will add I think it is also nice to have all the logic for going from string to a colormap outside the colormap object. I.e. `ensure_colormap` is actually quite intense, but `Colormap` right now is very simple. I like them being well separated.

napari/utils/colormaps/colormap.py
sofroniewn: For `controls` the thing I'm not so sure how to do with the dataclass is handle the init where None is provided and then it auto-calculates based on length of `colors` and `interpolation`. In general the dataclasses I find a little more tricky when there is coupling between the attributes.

napari/utils/colormaps/colormaps.py
tlambert03: it could mostly remain that way probably... also, there will probably be some pushback in general to touching `__new__`, since I think people tend not to like that :joy:

napari/utils/colormaps/colormap.py
sofroniewn: I've now added support for controls, interpolation, though i need to test `interpolation='zero'`. Also there are no custom setters now so things could fall out of place once made, but we might want to make this whole thing frozen ...

napari/utils/colormaps/colormap.py
```diff
@@ -0,0 +1,58 @@
...
+        Control points of the colormap.
+    interpolation : str
+        Colormap interpolation mode, either 'linear' or
+        'zero'. If 'linear', ncontrols = ncolors (one
+        color per control point). If 'zero', ncontrols
+        = ncolors+1 (one color per bin).
+    """
+
+    name: str = 'undefined'
+    colors: Property[np.ndarray, None, transform_color] = np.zeros((2, 4))
```
tlambert03, line 38: suggest either removing colors and controls from the repr with [`field(repr=False)`](https://docs.python.org/3/library/dataclasses.html#dataclasses.field), or declare your own `__repr__`.

napari/utils/colormaps/colormap.py
```diff
@@ -0,0 +1,58 @@
...
+
+class ColormapInterpolationMode(StringEnum):
+    """INTERPOLATION: Interpolation mode for colormaps."""
+
+    LINEAR = auto()
+    ZERO = auto()
+
+
+@dataclass(events=True, properties=True)
+class Colormap:
```
tlambert03, line 19: could be nice to add an `__iter__` method: 
```python
def __iter__(self):
    yield from (self.colors, self.controls, self.interpolation)
```
so that you can easily convert to a vispy colormap with:
```python
cm = napari.utils.Colormap(...)
vispy.colors.Colormap(*cm)
```

tlambert03: btw, it takes 153 µs ± 1.25 µs on my computer to convert to a vispy colormap like that.  no biggie

napari/utils/colormaps/colormaps.py
```diff
@@ -4,13 +4,13 @@
...
     get_colormap,
     get_colormaps,
 )
 
 from ...types import ValidColormapArg
 from .vendored import cm, colorconv
+from .colormap import Colormap
 
 _matplotlib_list_file = os.path.join(
     os.path.dirname(__file__), 'matplotlib_cmaps.txt'
```
tlambert03, line 16: in this PR or another one... can we just get rid of `matplotlibs_cmaps.txt`... I get the original inspiration, but it's the only place where we are reading a text file to create a list of strings.

napari/utils/colormaps/colormaps.py
```diff
@@ -273,57 +279,13 @@ def vispy_or_mpl_colormap(name):
                 'or matplotlib.'
             )
         mpl_colors = mpl_cmap(np.linspace(0, 1, 256))
-        cmap = Colormap(mpl_colors)
+        cmap = Colormap(name=name, colors=mpl_colors)
```
tlambert03, line 282: perhaps have some `@classmethods` declared on our Colormap (`from_mpl` & `from_vispy`) ... so you can just do, e.g. `cmap = Colormap.from_mpl(mpl_cmap)`

napari/utils/colormaps/colormaps.py
```diff
@@ -393,16 +354,23 @@ def ensure_colormap_tuple(colormap: ValidColormapArg) -> Tuple[str, Colormap]:
...
         name = None
         for key, val in AVAILABLE_COLORMAPS.items():
             if colormap == val:
                 name = key
                 break
         if not name:
             name = increment_unnamed_colormap(AVAILABLE_COLORMAPS)
-        AVAILABLE_COLORMAPS[name] = colormap
+        # Convert from vispy colormap
+        cmap = Colormap(
```
tlambert03, line 367: see comment elsewhere about the `Colormap.from_vispy` classmethod idea.

tlambert03 (commented):
I'm feeling good about this @sofroniewn, particularly if you like it.  There's only a _tiny_ bit of logic duplication from other libraries, and since they don't all do it the same way, it provides us with a place to unify their APIs internally if need be (see the `@classmethod` comment in review).

It'd be nice to evaluate this without the evented dataclass stuff for now (since we're not really using the events yet).   Since it's a bit hard to use `@property` on a regular python dataclass, maybe just make this a "good old fashioned" class for now (😄 ).  We can pull in the dataclass stuff later.

I'd also like to see the `ensure_colormap_tuple` usage removed.

napari/utils/colormaps/colormaps.py
sofroniewn: interesting idea, i'll look into that

sofroniewn:
Thanks for the review @tlambert03 glad you're ok with the general direction.

I think I'll make a new PR with a pure class based approach and your additional suggestions, we can then think about the dataclass one later

sofroniewn:
Superceeded by #1523

PR #1521, opened 2020-08-04
Fix potentially-confusing typo in comment.

danielballan:
# Description
This fixes what I _believe_ is a typo in a comment. It's not so hard to figure out what is meant, but `__getattr__` magic can be confusing so I figure it might be worth patching.

## Type of change

- [x] Documentation fix


tlambert03 (approved):
indeed!  thanks for catching, and thanks for the fix!

PR #1522, opened 2020-08-04
Fix magic_name and kwargs propagation with view_path

### LINKED ISSUE #1518: magic name guessing no longer works with napari.view_*
## 🐛 Bug

Tragically, it looks like 0e53da7965c96315e74716f25a477c918d1c18ff / #978 broke @kne42's wonderful magic name guessing when using `v = napari.view_image(my_named_data)`. This makes sense, since the generated code is no longer easily detectable as being part of our stack.

One option would be to generate the code at *install time* rather than *run time*. Thoughts @tlambert03?

## To Reproduce

```
import numpy as np
import napari

image = np.random.random((5, 48, 48))

with napari.gui_qt():
    v = napari.Viewer()
    v.add_image(image)
    v2 = napari.view_image(image)
```

Gives this for `v`:

Screen Shot 2020-08-03 at 11 49 33 pm

and this for `v2`:

Screen Shot 2020-08-03 at 11 49 28 pm

tlambert03:
# Description

this fixes #1518 and also supersedes #1520 (which I just closed).  This was quite interesting!
The fix for #1518 was actually quite simple... I just needed to change this line:

https://github.com/napari/napari/blob/bba25417a41f724257fab49d910305b542a4e0f0/napari/view_layers.py#L136

to `compile(fakefunc, __file__, "exec")`.  That second argument is the filepath that python thinks it's compiling this code block.  So to fix magic naming, it just needed to appear to be within the codebase (instead of some random file named "fakesource")

I went a step further though and cleaned up the stack trace.  Before this PR, the stack trace for a `view_image` call (from the perspective of the `magic_name` function) looked like this:
```python
# file::method
napari/napari/layers/base/base.py::__init__
napari/napari/layers/intensity_mixin.py::__init__
napari/napari/layers/image/image.py::__init__
napari/napari/components/add_layers_mixin.py::add_image
napari/napari/view_layers.py::real_func  # 👀
fakesource::view_image  # this was the problem with magic_name
```
after this PR it looks like this
```
# file::method
napari/napari/layers/base/base.py::__init__
napari/napari/layers/intensity_mixin.py::__init__
napari/napari/layers/image/image.py::__init__
napari/napari/components/add_layers_mixin.py::add_image
napari/napari/view_layers.py::view_image
```
Note that the wrapper `real_func` is gone and it looks just like the stack traces would have looked before the autogeneration.

from #1520, it also fix another issue with the autogenerated `view_path` function wherein `**kwargs` were not getting re-expanded and so were being passed to the inner function as `kwargs={'name': 'something'}` instead of `name='something'`.  So the CLI kwargs passing stuff in #1220 was not working

Tests are added. This PR also deletes the ambiguously `test_napari.py` file and puts its contents into `test_view_layers` where all of the other similar tests were.

## Type of change

# How has this been tested?

- [x] tests added for both CLI and `view_path` behavior

## Final checklist:


sofroniewn (approved):
This LGTM! Thanks for the fixed and detailed explanation of what was going on!!

jni (approved):
Wonderful.

PR #1361, opened 2020-06-15
[WIP] Tracks layer

quantumjot:
# Description
This PR provides a prototype `Tracks` layer for napari, as discussed in https://github.com/napari/napari/issues/539.

napari

The layer provides several functionalities:
1. Visualization of nD+t tracks (sequences of points in space, connected over time) as projections in the viewer.
2. Visualization of an associated graph - or meta-links between tracks. For example, a cell track may branch. The graph stores the relationship between the end of the parent track, and the beginning of the child tracks. This functionality should work for arbitrary number of input/output edges, and merges too.
3. Coloring of track vertices by associated scalar or vector properties
4. Text identifiers for the track ID

#### Implementation

I've implemented the tracks layer like this:
```python
viewer.add_tracks(data, properties=[], colormaps_dict={})
```

Data (`data`) are provided as a list of NxD arrays. Each array represents a track, with the first axis representing time (e.g. time, x, y, z ...). Rather than slice these arrays, which may not neatly fall on integer boundaries, a vertex/fragment shader (`_track_shader.py`) modulates the alpha value according to distance from the dimensional indices. This has good performance and means that the time dimension can be a float.

`properties` is a list (of equal length to data) such that there is a one-to-one mapping between the track and the properties. Each item of the list is a dictionary, which must be either a scalar property (such as track ID) or a vector property which describes some property of each vertex. This enables the tracks to be colored appropriately by the user. These properties are added to a `color_by` combo box in the layer. If no properties are provided, the layer automatically generates a track `ID` property for each input. This can be used by the text visual to identify each track.

Properties also provide a way to implement the graph functionality. Since each track always has an `ID`, it can also be provided with a `parent` in the form of a scalar or list. The parent property maps any branches in the graph, such that splitting and merging events can be visualized.

Custom colormaps can be provided via `colormaps_dict` to the layer, where the key is matched to a property key. If nothing is provided, the layer autoscales the colormap to the data. In reality this can be anything that returns an RGBA value given an input value.

If you want to try this, I'm providing some simple 2D+t and 3D+t tests here:
https://gist.github.com/quantumjot/f400e28ff90d02b6f843b0ca6974d27b

#### Extra thoughts

+ Using the vertex shader to slice the data provides a neat way to interpolate and project the data. It's really fast.
+ Currently, when rolling the displayed dimensions, there are no slicing operations (except for the time index). This is because the data may need to be interpolated and/or binned and I'm not sure how you want to deal with this in general. I'm guessing that this will also cause a performance drop.
[...]

Issue #899, opened 2020-01-23
Matplotlib colormaps for images aren't recognized


HagaiHargil:
## 🐛 Bug

When trying to use `matplotlib`'s colormaps with napari many of them return a `KeyError` even though the docs says that napari should recognize these names.

## To Reproduce

```python
img = np.array([[1, 0], [0, 1]])
with napari.gui_qt():
    napari.view_image(img, colormap='BuPu')  # KeyError
```

## Expected behavior

napari should support all matplotlib colormaps (as stated), or at least change the docs accordingly.

## Environment

Python 3.7, napari 0.2.10

Full traceback:
```
Traceback (most recent call last):
  File "/data/neural_collision_detection/results/for_article/fig1/code/find_collisions_distribution.py", line 58, in <module>
    napari.view_image(img, colormap='BuPu')  # KeyError
  File "/home/hagai/.conda/envs/ncd/lib/python3.7/site-packages/napari/view_layers.py", line 137, in view_image
    path=path,
  File "/home/hagai/.conda/envs/ncd/lib/python3.7/site-packages/napari/components/viewer_model.py", line 519, in add_image
    visible=visible,
  File "/home/hagai/.conda/envs/ncd/lib/python3.7/site-packages/napari/layers/image/image.py", line 214, in __init__
    self.colormap = colormap
  File "/home/hagai/.conda/envs/ncd/lib/python3.7/site-packages/napari/layers/intensity_mixin.py", line 77, in colormap
    self._cmap = self._colormaps[name]
KeyError: 'BuPu'
```

jni:
We ultimately decided to trim the supported colormaps to a curated list. I guess the easiest fix is to update the docs! And maybe provide a simple code snippet to add matplotlib colormaps. It should be quite easy.

jni:
Specifically, here is how to add a colormap:

https://github.com/napari/napari/blob/master/examples/set_colormaps.py

So you could use

```python
layer.colormap = 'BuPu', vispy.color.Colormap(plt.cm.BuPu(np.linspace(0, 1, 256)))
```

HagaiHargil:
Fair enough, I guess a reference to the available colormaps and to the linked example should be pretty clear.

GenevieveBuckley:
> We ultimately decided to trim the supported colormaps to a curated list.

[...]

PR #1523, opened 2020-08-05
add simple colormap object

### LINKED ISSUE #862: Generic colormap representation

sofroniewn:
# Description
This PR adds a simple numpy based colormap object that can easily be converted to a vispy or other colormap. It superceeds #1517. Will close #862 

It is still a WIP, I will update as I go. So far I've only looked at image/ surface properly. Labels, points, vectors, shapes still need work.

## Type of change

# References

# How has this been tested?

## Final checklist:


sofroniewn:
Just did some more work on this, it's getting close. I now have to colormap conversion working for all layers and a test that ensure parity between our colormaps and a vispy colormap object to 6 decimals on all our available colormaps.

I need to fix some tests which are now basic syntax errors and update some doc strings. I might also work on some additional ways to accept colormaps, and I need to fix the serialization tests.

sofroniewn:
@jni @kevinyamauchi @tlambert03 this PR is now ready for review.

It adds a simple Colormap object capable of mapping colors with simple numpy operations. I've preserved the API so that all vispy Colormap objects can still be passed. We might have problems with some vispy colormaps that are purely shader defined (like `grays` and `fire`, but they are not common I think).

I'm also still just putting `colormap.name` inside the layer data tuple so the serialization tests pass, but fixing that would be easy to address in a new PR.

I noticed that @quantumjot was doing things with custom colormaps here https://github.com/quantumjot/arboretum/blob/master/arboretum/_colormaps.py for the Tracks Layer work and maybe in #1361. I'm curious if this work would still allow your use cases?

quantumjot:
Hi @sofroniewn - this looks great! From a cursory look, I don't see why this will cause any problems for the track layer going forward, and I think having a unified colormap object makes things easier in general. Only thing I didn't see is a color cycle method for discrete colormaps, but maybe I missed that.

sofroniewn:
> Hi @sofroniewn - this looks great! From a cursory look, I don't see why this will cause any problems for the track layer going forward, and I think having a unified colormap object makes things easier in general

That's great, thanks for the quick look @quantumjot!

> Only thing I didn't see is a color cycle method for discrete colormaps, but maybe I missed that.

There is a way to make a discrete colormap with n colors in n+1 bins using the `interpolation='zero'` option. Right now the API there matches the vispy one identically, but we could add more convenience features here if desired in the future

kevinyamauchi:
Sorry, things have been crazy at work! I can review on Sunday.

napari/layers/shapes/_tests/test_shapes.py
```diff
@@ -1121,9 +1120,9 @@ def test_add_colormap(attribute):
     args = {color_kwarg: 'shape_type', colormap_kwarg: 'viridis'}
     layer = Shapes(data, properties=annotations, **args)
 
-    setattr(layer, f'{attribute}_colormap', get_colormap('gray'))
+    setattr(layer, f'{attribute}_colormap', 'gray')
     layer_colormap = getattr(layer, f'{attribute}_colormap')
-    assert 'unnamed colormap' in layer_colormap[0]
+    assert layer_colormap.name == 'gray'
```
kevinyamauchi, line 1125: I think it's neat that this works now!

napari/utils/colormaps/colormaps.py
```diff
@@ -360,16 +353,15 @@ def increment_unnamed_colormap(
     return name
 
 
-def ensure_colormap_tuple(colormap: ValidColormapArg) -> Tuple[str, Colormap]:
+def ensure_colormap(colormap: ValidColormapArg) -> Tuple[str, Colormap]:
```
kevinyamauchi: Since this function has a lot of branches, I think it would be nice to add some tests.

napari/layers/labels/_tests/test_labels.py
```diff
@@ -279,14 +279,12 @@ def test_colormap():
     np.random.seed(0)
     data = np.random.randint(20, size=(10, 15))
     layer = Labels(data)
-    assert type(layer.colormap) == tuple
-    assert layer.colormap[0] == 'random'
-    assert type(layer.colormap[1]) == Colormap
+    assert type(layer.colormap) == Colormap
```
kevinyamauchi: I'm not sure if we have plans to subclass `Colormap`, but should we use `isinstance` instead?

```suggestion
    assert isinstance(layer.colormap, Colormap)
```

napari/_vispy/vispy_surface_layer.py
```diff
@@ -53,11 +53,15 @@ def _on_data_change(self, event=None):
...
         if self.layer.gamma != 1:
             # when gamma!=1, we instantiate a new colormap with 256 control
             # points from 0-1
-            cmap = Colormap(cmap[np.linspace(0, 1, 256) ** self.layer.gamma])
+            colors = self.layer.colormap.map(
+                np.linspace(0, 1, 256) ** self.layer.gamma
+            )
+            cmap = Colormap(colors)
+        else:
+            cmap = Colormap(*self.layer.colormap)
```
kevinyamauchi: I think it's a bit confusing that we now have 2 types of `Colormap` objects now (vispy and napari). Perhaps we can import the vispy colormap as `VispyColormap`?

napari/utils/colormaps/colormaps.py
kevinyamauchi:
```suggestion
def ensure_colormap(colormap: ValidColormapArg) -> Colormap:
```

It looks like just the `Colormap` object is returned now. If so, I think the `Returns` section of the docstring needs to be updated as well.

napari/utils/colormaps/colormap.py
```diff
@@ -0,0 +1,73 @@
...
+        Colormap interpolation mode, either 'linear' or
+        'zero'. If 'linear', ncontrols = ncolors (one
+        color per control point). If 'zero', ncontrols
+        = ncolors+1 (one color per bin).
+    name : str
+        Name of the colormap.
+    """
+
+    def __init__(
+        self, colors, controls=None, interpolation='linear', name='undefined'
```
kevinyamauchi: This is a bit of a nitpick, but I think `undefined` is a bit of a confusing name since the color map is defined. Perhaps something like `custom_colormap` instead?

napari/utils/colormaps/colormaps.py
```diff
@@ -265,6 +301,7 @@ def vispy_or_mpl_colormap(name):
     vispy_cmaps = get_colormaps()
     if name in vispy_cmaps:
         cmap = get_colormap(name)
+        cmap = convert_vispy_colormap(cmap, name=name)
```
kevinyamauchi: I think the docstring needs to be updated now that the returned colormap is a napari `Colormap`.

napari/layers/image/image.py
```diff
@@ -450,7 +450,7 @@ def _get_state(self):
             {
                 'rgb': self.rgb,
                 'multiscale': self.multiscale,
-                'colormap': self.colormap[0],
+                'colormap': self.colormap.name,
```
kevinyamauchi, line 465: What do you think about storing a dictionary with the kwargs required to make the `Colormap` object? That way even custom colormaps could be reloaded from the `layer._get_state()` output.

napari/utils/colormaps/colormap.py
```diff
@@ -0,0 +1,73 @@
...
+
+import numpy as np
+
+from ...utils.misc import StringEnum
+from .colorbars import make_colorbar
+from .standardize_color import transform_color
+
+
+class ColormapInterpolationMode(StringEnum):
+    """INTERPOLATION: Interpolation mode for colormaps."""
```
kevinyamauchi: I believe we include a definition of the enum values in our other mode enums. I think it would be nice to do so here as well.

kevinyamauchi (commented):
I think this is looking good, @sofroniewn ! I have left some comments below.

Also, I was wondering if we should rename `/napari/utils/colormaps/colormaps.py`. It's a bit confusing because there's also a `colormaps.py` in the same folder. Maybe we can name it `colormap_utils.py` instead?

napari/_vispy/vispy_surface_layer.py
tlambert03: was just about to make the same comment on the other vispy file :)

napari/_vispy/vispy_surface_layer.py
```diff
@@ -53,11 +53,15 @@ def _on_data_change(self, event=None):
         self._on_translate_change()
 
     def _on_colormap_change(self, event=None):
-        cmap = self.layer.colormap[1]
         if self.layer.gamma != 1:
```
tlambert03, line 56: I think this check is no longer necessary after #1456, now that gamma is performed by the shader?  Can you check to see if it works to just assign `cmap = Colormap(*self.layer.colormap)` without the gamma check?

napari/layers/image/image.py
tlambert03: as briefly discussed before... I was kinda hoping this would actually just be the colormap object.  While that may not be immediately serializable (which I know was the original intended use for `_get_state`), truth is we've never actually used it that way, and now it forms a main part of the layer_data_tuple that plugins might get.  If we give them the object, they immediately have something useful.  If we just give them the name (or stuff to construct it), then they have to either depend on some of our convenience functions, or recreate them.  Serialization can be solved if/when we've ever truly need it, but passing the actual object would be immediately useful in (e.g.) napari-svg right?

napari/layers/labels/labels.py
```diff
@@ -580,7 +574,7 @@ def get_color(self, label):
             col = None
         else:
             val = self._raw_to_displayed(np.array([label]))
-            col = self.colormap[1][val].rgba[0]
+            col = self.colormap.map(val)[0]
```
tlambert03, line 580: ❤️

napari/layers/surface/surface.py
```diff
@@ -237,7 +237,7 @@ def _get_state(self):
         state = self._get_base_state()
         state.update(
             {
-                'colormap': self.colormap[0],
+                'colormap': self.colormap.name,
```
tlambert03, line 240: same here... would love to see this just be `self.colormap`

napari/utils/colormaps/colormaps.py
```diff
@@ -45,6 +46,38 @@ def _all_rgb():
 LABRNG = LABMAX - LABMIN
 
 
+def convert_vispy_colormap(colormap, name=None):
```
tlambert03: just personal preference... but since this is basically just a modified `Colormap` constructor, I think it would be lovely for it to live on the `Colormap` (or at least be a private function if you really want it outside the class):

```python
class Colormap:
    @classmethod
    def from_vispy(cls, colormap, name=None):
        ...
        return cls(
            name=name,
            colors=colormap.colors.rgba,
            controls=controls,
            interpolation=interpolation,
        )
```

that cleans up imports elsewhere, so it's always:
```python
from napari.utils import Colormap

cmap = Colormap.from_vispy(vispy_colormap)
```
rather than sometimes importing Colormap and other times importing `convert_vispy_colormap` depending on the starting material.

also, having `name=None` here is going to break the typing on our Colormap object.  for example, if a name is not provided, and someone tries to do something like `cmap.name.lower()`, it would now be an attribute error. Either add `name=vispy_colormap` or something here... or make the default of `Colormap` be `None` and then set the default to `custom_colormap` in the `Colormap.__init__` function

napari/utils/colormaps/colormaps.py
```diff
@@ -45,6 +46,38 @@ def _all_rgb():
...
+    colormap : vispy.color.Colormap
+        Vispy colormap object that should be converted.
+    name : str
+        Name of colormap, optional.
+
+    Returns
+    -------
+    napari.utils.Colormap
+    """
+    if hasattr(colormap, '_controls'):
```
tlambert03, line 82: I think it would be reasonable to just check `isinstance(colormap, VispyColormap)` and throw a `TypeError` otherwise, rather than checking hasattr.  This function is, afterall, dedicated to vispy objects.

napari/utils/colormaps/colormap.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def __init__(
+        self, colors, controls=None, interpolation='linear', name='undefined'
+    ):
+
+        self.name = name
+        self.colors = transform_color(colors)
+        self.interpolation = ColormapInterpolationMode(interpolation)
+        if controls is None:
+            N = len(self.colors) + int(self.interpolation == 'zero')
```
tlambert03: lowercase?

tlambert03 (approved):
really great work @sofroniewn!  love what this does to a lot of our code.
I'm marking as "approve" but I'd definitely like to hear your thoughts on some of my comments before merge.

sofroniewn:
Thanks for review @kevinyamauchi @tlambert03, I'll look into addressing concerns promptly

napari/utils/colormaps/colormaps.py
kevinyamauchi: I think making a `Colormap.from_vispy()` constructor is a good idea!

napari/_vispy/vispy_surface_layer.py
sofroniewn: The surface layer uses the `Mesh` visual from vispy, which we didn't vendor in #1456. Not sure if gamma has been added to Mesh upstream in vispy, but I just tried removing the gamma check and it didn't work.

napari/_vispy/vispy_surface_layer.py
sofroniewn: done in both files

napari/layers/labels/_tests/test_labels.py
sofroniewn: fixed here and everywhere

napari/utils/colormaps/colormap.py
sofroniewn: i've gone with `custom`

napari/utils/colormaps/colormap.py
sofroniewn: gone with `n_controls`

napari/utils/colormaps/colormaps.py
sofroniewn: I've added in the type check, unfortunately I think I need to add the `hasattr` check as I have encountered vispy colormaps that don't have those attrs! They may have been strangely constructed ones, like using mostly custom shader code, which we might not be able to translate properly anyway, but the hasattr check seemed safest ....

napari/utils/colormaps/colormaps.py
sofroniewn: I've changed the default `name='vispy'` in `convert_vispy_colormap`, but I think I'd rather leave the `from_vispy` method off of the `Colormap` object. We only use `convert_vispy_colormap` in one file in the code base, the one it is currently defined in, and I wouldn't really want to encourage users to me making vispy colormaps and then converting them to napari ones (note we can accept vispy colormaps fine through the API).

napari/utils/colormaps/colormaps.py
sofroniewn: done

napari/utils/colormaps/colormaps.py
sofroniewn: Done

napari/layers/image/image.py
sofroniewn: I'd like to wait on this until after this PR and make a dedicated issue to follow up. Making that change will require additional PRs to napari-svg to be able to use the new return value and it could get tricky to coordinate all the PRs across repos. As the current change is doesn't actually change any functionality, I'd like to get this in first and then fix, but agreed it would be nice to resolve

napari/layers/surface/surface.py
sofroniewn: Agreed, see comment above about staging this into multiple PRs so as not to break napari-svg

napari/utils/colormaps/colormaps.py
sofroniewn: Comprehensive tests added, and existing tests consolidate into `napari.utils.colormaps` folder

sofroniewn (commented):
Ok, I have gone through and addressed all the reviewer comments apart from two.

I have not added the `from_vispy` method to the Colormap object we only need it in one file, see comments above.

I have not made the `_get_state` dict get the whole colormap, just the `colormap.name` as giving the colormap now would be a breaking change and require a series of PRs with napari-svg to ensure everything works. In the interest of keeping our PRs small and manageable and the project moving along I'd like us to merge as is now which is fully backwards compatible, and then make breaking changes in a new PR. I will make a dedicated issue for that if this gets merged as is so we don't forget.

@jni if you could give this a quick pass that would be great, it has already received two detailed reviews and an approval, but maybe you can focus on changes since the last reviews in case anything funny crept in. Thanks!!

napari/_qt/layer_controls/qt_labels_controls.py
```diff
@@ -494,8 +495,8 @@ def paintEvent(self, event):
                         painter.setBrush(QColor(25, 25, 25))
                     painter.drawRect(i * 4, j * 4, 5, 5)
         else:
-            color = 255 * self.layer._selected_color
-            color = color.astype(int)
+            color = np.multiply(self.layer._selected_color, self.layer.opacity)
```
jni, line 498: Can you explain this change? Why mix in opacity now?

napari/components/_tests/test_multichannel.py
```diff
@@ -3,16 +3,23 @@
...
-green_cmap = colormaps.simple_colormaps['green']
-red_cmap = colormaps.simple_colormaps['red']
-fire = colormaps.AVAILABLE_COLORMAPS['fire']
-cmap_tuple = ("my_colormap", colormaps.Colormap(['g', 'm', 'y']))
-cmap_dict = {"your_colormap": colormaps.Colormap(['g', 'r', 'y'])}
+base_colormaps = CYMRGB
+two_colormaps = MAGENTA_GREEN
+green_cmap = simple_colormaps['green']
+red_cmap = simple_colormaps['red']
+blue_cmap = AVAILABLE_COLORMAPS['blue']
```
jni, line 20: Sorry, this is nitpicky, but why isn't this simple_colormaps like the first two?

napari/layers/intensity_mixin.py
```diff
@@ -45,16 +45,13 @@ def reset_contrast_limits_range(self):
 
     @property
     def colormap(self):
-        """2-tuple of str, vispy.color.Colormap: colormap for luminance images.
+        """napar.utils.Colormap: colormap for luminance images.
```
jni: napari

napari/utils/colormaps/_tests/test_colormap.py
```diff
@@ -0,0 +1,53 @@
...
+    assert camp.interpolation == 'linear'
+    assert len(camp.controls) == len(colors)
+    np.testing.assert_almost_equal(camp.colors, colors)
+    np.testing.assert_almost_equal(camp.map([0.75]), [[0, 0.5, 0.5, 1]])
+
+
+def test_linear_colormap_with_control_points():
+    """Test a linear colormap with control points."""
+    colors = np.array([[0, 0, 0, 1], [0, 1, 0, 1], [0, 0, 1, 1]])
+    camp = Colormap(colors, name='testing', controls=[0, 0.75, 1])
```
jni: cmap?

napari/utils/colormaps/_tests/test_colormaps.py
```diff
@@ -0,0 +1,123 @@
...
+    # colormap
+    vispy_cmap = VispyColormap(*cmap)
+    vispy_colors = vispy_cmap.map(values)
+    np.testing.assert_almost_equal(colors, vispy_colors, decimal=6)
+
+
+def test_increment_unnamed_colormap():
+    # test that unnamed colormaps are incremented
+    names = [
+        '[unnamed colormap 0',
```
jni: Should there be a ]?

napari/utils/colormaps/colormap_utils.py
```diff
@@ -45,6 +47,47 @@ def _all_rgb():
...
+        )
+
+    # Not all vispy colormaps have an `_controls`
+    # but if they do, we want to use it
+    if hasattr(colormap, '_controls'):
+        controls = colormap._controls
+    else:
+        controls = None
+
+    # Not all vispy colormaps have an `_controls`
```
jni: Interpolation

napari/utils/colormaps/colormap_utils.py
```diff
@@ -360,21 +362,20 @@ def increment_unnamed_colormap(
...
     """Accept any valid colormap arg, and return (name, Colormap), or raise.
 
     Adds any new colormaps to AVAILABLE_COLORMAPS in the process.
 
     Parameters
     ----------
     colormap : ValidColormapArg
-        colormap as str, Colormap, {name: Colormap} ``dict``, or
-        (name, Colormap) ``tuple``.
+        colormap as str, napari.utils.Colormap.
```
jni, line 384: @sofroniewn it looks to me like you need to update the definition of ValidColormapArg but haven't done so?

jni (approved):
@sofroniewn most of my comments are minor. Things I'm worried about:

- the type logic in ensure_colormap is pretty messy right now. Which is not a good prelude to:
- on reading this, my overarching feeling is that it's really sad to be adding yet another Colormap object to the ecosystem. I just saw that matplotlib colormaps do have a name attribute. But, we don't want to depend on matplotib. Fine. But it would be great if we could *accept* matplotlib colormaps as input. It would also be great if our colormaps could mimic the matplotlib colormap API, so that our colormaps could (conceivably) be used by matplotlib.
- of course, all this can come later and shouldn't block this PR. Would you mind creating an issue about this so we can follow up?

napari/_qt/layer_controls/qt_labels_controls.py
sofroniewn: This change is definitely independent of this PR, and could be done separately, but it makes the color of the swatch match the color of the label in the canvas and the color in the thumbnail, which I think will be nicer for people working with labels at lower opacity values

napari/components/_tests/test_multichannel.py
sofroniewn: I have made `SIMPLE_COLORMAPS` for consistency, I think reasons were purely historical, good catch

napari/layers/intensity_mixin.py
sofroniewn: done

napari/utils/colormaps/_tests/test_colormap.py
sofroniewn: done

napari/utils/colormaps/_tests/test_colormaps.py
sofroniewn: done

napari/utils/colormaps/colormap_utils.py
sofroniewn: done

napari/utils/colormaps/colormap_utils.py
sofroniewn: I have added, but had to move type definition into colormaps folder to avoid circular imports. This is probably not a long term ideal solution but seems like a reasonable step for now

sofroniewn:
> - the type logic in ensure_colormap is pretty messy right now. Which is not a good prelude to:
> - on reading this, my overarching feeling is that it's really sad to be adding yet another Colormap object to the ecosystem. I just saw that matplotlib colormaps do have a name attribute. But, we don't want to depend on matplotib. Fine. But it would be great if we could accept matplotlib colormaps as input. It would also be great if our colormaps could mimic the matplotlib colormap API, so that our colormaps could (conceivably) be used by matplotlib.
> - of course, all this can come later and shouldn't block this PR. Would you mind creating an issue about this so we can follow up?

I completely agree the `ensure_colormap` is quite messy right now, and that is mainly for a desire to be backwards compatible right now. We could deprecate some of these more esoteric inputs like a dictionary of named vispy colormaps, which I don't think is used anywhere in the codebase, and fairly redundant with named tuples. We could then keep only the simplest options and add support for all matplotlib colormaps, which has been requested in #899, and make it clear how to convert our colormaps to matplotlib colormaps (which is a one-linear for each of the `linear` and `zero` options)

```python
# for linear interpolated napari colormap
from matplotlib.colors import LinearSegmentedColormap
mp_cmap = LinearSegmentedColormap.from_list(cmap.name, cmap.map(np.linspace(0, 1, 256)))
```

```python
# for zero interpolated napari colormap
from matplotlib.colors import ListedColormap
mp_cmap = ListedColormap(cmap.map(np.linspace(0, 1, 256)), name=cmap.name)
```

See here using converting the napari `turbo` colormap to matplotlib and using inside the console:

Screen Shot 2020-08-23 at 1 17 15 PM

I will create a new issue referencing this and #899 to continue discussion, I really think this barebones colormap object will help us unify matplotlib and vispy colormaps under-the-hood and support both.

sofroniewn:
Issue created here https://github.com/napari/napari/issues/1575. All tests are passing, we can make fixes in followup PRs and continue discussion there!!

Issue #1514, opened 2020-07-31
Support for fractional ctrl-scrolling through stack


jni:
## 🚀 Feature

In #1434 we added support for using the scroll-wheel while holding Ctrl to change slices. However, this does not work well with a trackpad rather than a mouse, because the trackpad allows deltas <1, so holding Ctrl and slowly scrolling (ie two finger drag on Mac) actually does nothing but accumulate a whole bunch of events with delta <1. Ideally, we would add an accumulator value that "ticks over" and triggers a scroll when it adds up to more than 1 (or less than -1). The accumulator should be cleared when Ctrl is released, which on the plus side is something that is quite easy to detect! (In earlier discussions we mentioned timers etc which is hairy.)

Here is the relevant bit of code:

https://github.com/napari/napari/blob/936d2890410ac2dc92a42e397bfd7daec37657ea/napari/components/viewer_model.py#L399-L408

Here is an untested start:

https://github.com/napari/napari/pull/1434#discussion_r456730417

sofroniewn:
I just had an idea - given that two finder scrolling when the slider is in focus works nicely to scroll through the stack do we just want holding the `ctrl` key to temporarily toggle focus to the last used slider dims? Curious what @jni @tlambert03 think of this idea?

tlambert03:
yeah, that seems nice... what currently happens if you try two-finger scrolling with the `ctrl` down if no slider has focus?  And... isn't there always a "current" (last used) slider?

sofroniewn:
>  what currently happens if you try two-finger scrolling with the ctrl down if no slider has focus?

Pre-merge of #1434 holding ctrl and hovering over canvas 2 finger scrolling causes pan-zoom just like without ctrl. After #1434 it doesn't, so we'd need to use the #1434 approach of intercepting the scroll wheel call too, but instead of calling the increment dims methods on the model, I guess we'd just reroute the call to the Qt Slider.

> isn't there always a "current" (last used) slider?

I think if there are no sliders present it is `None` and I'm not sure what the default is after launch before any sliders are touched, but those are both easy problems to fix.

tlambert03:
I think I'm missing the point :) what's the actual behavior this would change/fix?  and is the point here specific for 2-finger-scrolling vs mouse-wheel scrolling?

sofroniewn:
[...]

PR #1525, opened 2020-08-05
Address trackpad support for dim scrolling

liu-ziyang:
# Description
addressed comments from https://github.com/napari/napari/issues/1514

## Type of change

# References
https://github.com/napari/napari/issues/1514

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/components/_tests/test_viewer_mouse_bindings.py
+++ b/napari/components/_tests/test_viewer_mouse_bindings.py
@@ -38,12 +38,12 @@ class WheelEvent:
         (
             ["Control"],
             WheelEvent(False),
-            [[5, 5, 5], [5, 5, 4], [5, 5, 4], [5, 5, 3]],
+            [[5, 5, 5], [5, 5, 4], [5, 5, 3], [5, 5, 0]],
         ),
         (
             ["Control"],
             WheelEvent(True),
-            [[5, 5, 5], [5, 5, 6], [5, 5, 6], [5, 5, 7]],
+            [[5, 5, 5], [5, 5, 6], [5, 5, 7], [5, 5, 9]],
         ),
     ],
 )
@@ -80,7 +80,7 @@ def test_paint(mouse_event, modifiers, native, expected_dim):
 
     # Simulate large scroll
     event = ReadOnlyWrapper(
-        mouse_event(delta=[0, 0.6], modifiers=modifiers, native=native)
+        mouse_event(delta=[0, 3], modifiers=modifiers, native=native)
     )
     mouse_wheel_callbacks(viewer, event)
     assert np.equal(viewer.dims.point, expected_dim[3]).all()
--- a/napari/components/_tests/test_viewer_mouse_bindings.py
+++ b/napari/components/_tests/test_viewer_mouse_bindings.py
@@ -1,0 +1,90 @@
+import pytest
+import numpy as np
+import collections
+from napari.components import ViewerModel
+from napari.utils.interactions import (
+    ReadOnlyWrapper,
+    mouse_wheel_callbacks,
+)
+
+
+@pytest.fixture
+def mouse_event():
+    """Create a subclass for simulating vispy mouse events.
+
+    Returns
+    -------
+    Event : Type
+        A new tuple subclass named Event that can be used to create a
+        NamedTuple object with fields "delta", "modifiers" and "inverted".
+    """
+    return collections.namedtuple(
+        'Event', field_names=['delta', 'modifiers', 'native']
+    )
+
+
+class WheelEvent:
+    def __init__(self, inverted):
+        self._inverted = inverted
+
+    def inverted(self):
+        return self._inverted
+
+
+@pytest.mark.parametrize(
+    "modifiers, native, expected_dim",
+    [
+        ([], WheelEvent(True), [[5, 5, 5], [5, 5, 5], [5, 5, 5], [5, 5, 5]]),
+        (
+            ["Control"],
+            WheelEvent(False),
+            [[5, 5, 5], [5, 5, 4], [5, 5, 4], [5, 5, 3]],
+        ),
+        (
+            ["Control"],
+            WheelEvent(True),
+            [[5, 5, 5], [5, 5, 6], [5, 5, 6], [5, 5, 7]],
+        ),
+    ],
+)
+def test_paint(mouse_event, modifiers, native, expected_dim):
+    """Test painting labels with circle/square brush."""
+    viewer = ViewerModel()
+    data = np.random.random((10, 10, 10))
+    viewer.add_image(data)
+    viewer.dims.last_used = 2
+    viewer.dims.set_point(axis=0, value=5)
+    viewer.dims.set_point(axis=1, value=5)
+    viewer.dims.set_point(axis=2, value=5)
+
+    # Simulate tiny scroll
+    event = ReadOnlyWrapper(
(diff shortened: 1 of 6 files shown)


sofroniewn:
@ziyangczi this is working very nice, functionality works great. I pullout out the `_scroll` method into it's own file `_viewer_mouse_bindings.py` to match how we do things for our layers (see `_points_mouse_bindings.py`) which means that we didn't need both the `self` and `viewer` parameter. This probably should have been done in #1434, but it's fine to do now too.

I also moved `_scroll_progress` onto `dims` directly as it is only needed there, trying to keep `ViewerModel` a little more clean.

Can you add a test for the new `dims_scroll` function? See`test_points_mouse_bindings.py` for some examples of how we have tested mouse functions in the past. This test can go in a new file `napari/components/_tests/test_viewer_mouse_bindings.py`. Note that the mocking for this might be slightly different as the mouse wheel event is slightly different, but the overall principle holds https://github.com/napari/napari/blob/master/napari/layers/points/_tests/test_points_mouse_bindings.py

Thanks for the great work here!!

liu-ziyang:
added test to check for different directions scrolling

napari/components/_tests/test_viewer_mouse_bindings.py
```diff
@@ -0,0 +1,90 @@
...
+    viewer.dims.set_point(axis=0, value=5)
+    viewer.dims.set_point(axis=1, value=5)
+    viewer.dims.set_point(axis=2, value=5)
+
+    # Simulate tiny scroll
+    event = ReadOnlyWrapper(
+        mouse_event(delta=[0, 0.6], modifiers=modifiers, native=native)
+    )
+    mouse_wheel_callbacks(viewer, event)
+    print(viewer.dims.point)
```
sofroniewn:
```suggestion
```

napari/components/_tests/test_viewer_mouse_bindings.py
```diff
@@ -0,0 +1,90 @@
...
+    assert np.equal(viewer.dims.point, expected_dim[0]).all()
+
+    # Simulate tiny scroll
+    event = ReadOnlyWrapper(
+        mouse_event(delta=[0, 0.6], modifiers=modifiers, native=native)
+    )
+    mouse_wheel_callbacks(viewer, event)
+    print(viewer.dims.point)
```
sofroniewn:
```suggestion
```

napari/components/_tests/test_viewer_mouse_bindings.py
```diff
@@ -0,0 +1,90 @@
...
+    assert np.equal(viewer.dims.point, expected_dim[1]).all()
+
+    # Simulate tiny scroll
+    event = ReadOnlyWrapper(
+        mouse_event(delta=[0, 0.9], modifiers=modifiers, native=native)
+    )
+    mouse_wheel_callbacks(viewer, event)
+    print(viewer.dims.point)
```
sofroniewn:
```suggestion
```

napari/components/_tests/test_viewer_mouse_bindings.py
```diff
@@ -0,0 +1,90 @@
...
+    assert np.equal(viewer.dims.point, expected_dim[2]).all()
+
+    # Simulate large scroll
+    event = ReadOnlyWrapper(
+        mouse_event(delta=[0, 0.6], modifiers=modifiers, native=native)
+    )
+    mouse_wheel_callbacks(viewer, event)
+    print(viewer.dims.point)
```
sofroniewn:
```suggestion
```

sofroniewn (approved):
Amazing, thanks! @ziyangczi this looks great. I'd probably drop the print statements from test, see suggested edit, but otherwise this looks good to go!!

Maybe @jni and @DragaDoncila want to give it a go too. Very nice

napari/components/_tests/test_viewer_mouse_bindings.py
liu-ziyang: > Amazing, thanks! @ziyangczi this looks great. I'd probably drop the print statements from test, see suggested edit, but otherwise this looks good to go!!
> 
> Maybe @jni and @DragaDoncila want to give it a go too. Very nice

ah sneaky dev code made to commit, removed! nice catch!

DragaDoncila:
@ziyangczi this is very nice. So much cleaner not to have the scroll method in the viewer class. Works perfectly for me.

Only thing I remember being discussed that I don't see here is the potential for an old fractional delta to be stored in `self._scroll_progress` and cause the user to scroll further than they meant to. However I tend to think that's a slight non-issue, as I couldn't really notice it happening, and scrolling with a touchpad will always be slightly less precise than with a wheel. Perhaps @jni  feels differently.

napari/components/_viewer_mouse_bindings.py
```diff
@@ -0,0 +1,15 @@
...
+        viewer.dims._scroll_progress += event.delta[1]
+    else:
+        viewer.dims._scroll_progress -= event.delta[1]
+    if abs(viewer.dims._scroll_progress) >= 1:
+        for i in range(int(abs(viewer.dims._scroll_progress))):
+            if viewer.dims._scroll_progress < 0:
+                viewer.dims._increment_dims_left()
+            else:
+                viewer.dims._increment_dims_right()
+        viewer.dims._scroll_progress = 0
```
jni: Rather than resetting to 0, this should be

```suggestion
        viewer.dims._scroll_progress %= 1
```

jni (commented):
@ziyangczi I've made a suggestion about reducing rather than resetting the value of the scroll accumulator after a scroll event. Additionally, would you be able to address the comment noted by Draga? ie: after I release Ctrl, the accumulator should be reset. I *hope* that this is a straightforward adding of an on-key-release shortcut on the viewer model.

Thanks! The code is very nice, I have no other comments. =)

napari/components/_viewer_mouse_bindings.py
sofroniewn: This won't get the `-` case correct, if < 0 need to also subtract `1` after modulus operation

liu-ziyang: yep, I am going to update the loop to do so

liu-ziyang: I updated the calculation loop, hopefully addressed this comment and the other concern about residual value

sofroniewn: Update to calculation looks good @ziyangczi, but I think i'd still put the reset in the control button release and not in the mouse wheel event. The use case I'm imagining is someone presses control they scroll, an increment is left, they release, then the press control and start scrolling again and that increment is used. The approach of putting the reset in the the mouse wheel event with no control press won't reset the increment in this case, but if we put it on the control release then it will reset. Does that make sense?

liu-ziyang: added in the keybinding section. I am slightly concerned since this method touches Viewer, instead of ViewerModel, but this should reset things cleanly

sofroniewn: > I am slightly concerned since this method touches Viewer, instead of ViewerModel

Yes you are right to notice that this is slightly weird. We might actually want to pull most of those keybindings out into the ViewerModel and just leave https://github.com/napari/napari/blob/master/napari/_viewer_key_bindings.py#L5 there, which I think is the only one that really needs to be there. We can do that in a new PR though

Issue #1176, opened 2020-04-28
layer_to_visual prevent custom Layer subclasses


tlambert03:
### :bug: Bug
Was just playing around with making some layer subclasses with some method overrides as an example for someone, and found that it was _very_ difficult to add a subclassed `Layer` to the viewer with our `layer_to_visual` pattern.

when you try `viewer.add_layer(MyLayerSubclass(data))` ... it will throw a `KeyError` in both the `_vispy`  `visual = layer_to_visual[type(layer)](layer)` function (same with `_qt` `layer_to_controls`), because those dicts only know about our builtin layer types.

Just wanted to mention that somewhere so it's on the radar.

jni:
I definitely think that's a bug, and we do want to allow custom layers. Thanks for flagging!

sofroniewn:
Agreed - getting rid of the `layer_to_visual` approach would be good

sofroniewn:
This was closed by #1528

PR #1528, opened 2020-08-06
Changed class evaluation from `type` to `isinstance`

JoOkuma:
# Description
This PR closes issue #1176.
It changes the `create_qt_controls` and `create_vispy_visual` to evaluate the layer's class with `isinstance` and not `type`.

## Type of change
- [X] Bug-fix (non-breaking change which fixes an issue)

# References
Issue #1176 and [image.sc forum post](https://forum.image.sc/t/is-it-possible-to-constrain-regions-where-it-is-possible-to-insert-points/41286/4).

# How has this been tested?
- [X] Using the standard `pytest` command as described [here](https://github.com/napari/napari/blob/master/docs/source/developers/TESTING.md).

## Final checklist:

## Comments

The `create_qt_controls` need to respect a certain ordering because `Labels` inherits `Image` and they have different `QtControls`. The sequence is of if and else is not very elegant but I think it is less error-prone than using a list in place of the dictionary.

### DIFF
--- a/napari/_qt/layers/utils.py
+++ b/napari/_qt/layers/utils.py
@@ -36,4 +36,4 @@ def create_qt_controls(layer):
         if isinstance(layer, layer_type):
             return controls(layer)
 
-    raise ValueError
+    raise TypeError(f'Could not find QtControls for layer of type {type(layer)}')
--- a/napari/_qt/layers/utils.py
+++ b/napari/_qt/layers/utils.py
@@ -7,16 +7,6 @@ from .qt_surface_layer import QtSurfaceControls
 from .qt_vectors_layer import QtVectorsControls
 
 
-layer_to_controls = {
-    Image: QtImageControls,
-    Labels: QtLabelsControls,
-    Points: QtPointsControls,
-    Shapes: QtShapesControls,
-    Surface: QtSurfaceControls,
-    Vectors: QtVectorsControls,
-}
-
-
 def create_qt_controls(layer):
     """
     Create a qt controls widget for a layer based on its layer type.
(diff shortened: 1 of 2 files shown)


sofroniewn (approved):
LGTM! Thanks for the quick PR, I hope this helps your use case.

@tlambert03 @jni is there anything else you'd like to see from this before merge/ any alternative approaches to consider?

tlambert03:
I like what you did with the vispy layer:
```python
    for layer_type, visual in layer_to_visual.items():
        if isinstance(layer, layer_type):
            return visual(layer)
```

I would do it with the qt controls as well... but use an `OrderedDict` to make sure that `Labels` comes before image.
also... You can get rid of `Labels` in the vispy `layer_to_visual` dict now.

JoOkuma:
> I like what you did with the vispy layer:
> 
> ```python
>     for layer_type, visual in layer_to_visual.items():
>         if isinstance(layer, layer_type):
>             return visual(layer)
> ```
> 
> I would do it with the qt controls as well... but use an `OrderedDict` to make sure that `Labels` comes before image.
> also... You can get rid of `Labels` in the vispy `layer_to_visual` dict now.

I appreciate your feedback.

It seems that from Python 3.6 the standard dictionary is sorted and from 3.7+ this behavior is part of the Python specs.

Sources:
- [Python 3.6](https://docs.python.org/3/whatsnew/3.6.html)
- [Python 3.7+](https://docs.python.org/3/whatsnew/3.7.html)

tlambert03:
yeah, you can use the standard dict too.  we'll also be dropping 3.6 in the near future.

napari/_qt/layers/utils.py
```diff
@@ -31,6 +31,9 @@ def create_qt_controls(layer):
...
         Qt controls widget
     """
-    controls = layer_to_controls[type(layer)](layer)
 
-    return controls
+    for layer_type, controls in layer_to_controls.items():
+        if isinstance(layer, layer_type):
+            return controls(layer)
+
+    raise ValueError
```
tlambert03:
```suggestion
    raise TypeError(f'Could not find QtControls for layer of type {type(layer)}')
```

napari/_vispy/utils.py
```diff
@@ -29,6 +28,8 @@ def create_vispy_visual(layer):
     visual : vispy.scene.visuals.VisualNode
         Vispy visual node
     """
-    visual = layer_to_visual[type(layer)](layer)
+    for layer_type, visual in layer_to_visual.items():
+        if isinstance(layer, layer_type):
+            return visual(layer)
 
-    return visual
+    raise ValueError
```
tlambert03:
```suggestion
    raise TypeError(f'Could not find VispyLayer for layer of type {type(layer)}')
```

tlambert03 (approved):

tlambert03 (approved):
Thanks @JoOkuma.  not a lot of lines, but an important change we very much needed! :)

jni:
@tlambert03 do we want to add a test for the typeerror?

PR #1529, opened 2020-08-06
Add new viewer.cmd commands

pwinston:
# Description
This PR adds 3 "command line" commands meant to be typed into the IPython console:
* `viewer.cmd.help`
* `viewer.cmd.layers`
* `viewer.cmd.levels(layer_id)`

We created these while working on #1510. We have extracted them into this PR so we can get them merged separately, to keep the other PR smaller and in case they are useful on their own.

There's quite a lot of code related to formatting the tables nicely. I looked up some packages for command line tables but didn't evaluate any of them.

# Type of change

# New Dependencies

* [humanize](https://pypi.org/project/humanize/)

This is to print sizes in KB/MB/GB. I started doing this myself but realized might as well use this dependency which I've used before. They have other "human readable" thing like time ("two days ago" ) and big numbers ("12.3 billion").

# Examples

viewer-cmd-master

viewer-cmd-2

# References

The "debug menu" is somewhat similar in spirit to these commands. Both are meant to be things that developers can use, which might not be polished enough or useful to regular viewers. For #1510 it seems like command line was the better route, but maybe debug menu has a role there too.

# How has this been tested?
- [x] Manual testing

## Final checklist:


tlambert03 (approved):
This is a very nice feature @pwinston!

the `check_manifest` test is failing because there is no `__init__.py` file in `napari/commands/` ... so setuptools doesn't see it as a package and it would not be included in the distribution.

And though humanize certainly looks like a useful library, I also lean away from taking on a new dependency just for that one (small) function until we use more of it.

tlambert03 (changes requested):
actually... can you hook up at least one or two basic smoke tests?  It seems likely that this might break soon when we move to layergroups... so it will be good to have some early warning.

You can use the `capsys` pytest fixture to assert things like print statements

pwinston:
And so do what, copy the function from them and insert it, where? That is in the past if there's a dependency we just want to take a function from, where do we put it?

tlambert03:
> And so do what, copy the function from them and insert it, where? That is in the past if there's a dependency we just want to take a function from, where do we put it?


with what would amount to be like a 8-line function here, I'd probably just put it in the module that's using it. `napari/commands/_layers.py `

pwinston:
Some notes from napari sync meeting:
* The table formatting is nice. Instead of `viewer.cmd.layers` could this just be a `__str__` method of `LayerList` so you just type `viewer.layers` and it shows the same table?
* If so does `viewer.cmd.levels()` still exist or is there another way to get the levels?
* In #1510 the table was supposed to have a mix of layer and `ChunkLoader` information. Doesn't make sense for `LayerList` to know the async stuff. Have two tables?
* Make all this command stuff private since experimental?

In general there might be two paths here.
1. Short term just getting in place what #1510 needs so we can debug async loading.
2. Considering this a prod that tables or nicely formatted representations of things might be useful.

pwinston:
I made the review changes:
* It longer uses `humanize` dependency. I created its own `_humanize.py` file, I thought that was showing respect for their package. And I think other cmd's will want to use it.
* I moved it from `napari.commands` to `napari._commands` since it's experimental for sure. I added an `__init__.py`.
* I added some simple tests.

I think this is good enough for #1510?

I agree putting nicely formatted tables as `__str__` of objects sounds intriguing. But I'm wondering if you still need some namespace like `cmd.*` for "commands" meant to be used in the console. Some "flatter" set of commands that are easier to use interactively without really knowing the full API? Or is the API so simple it's just as good? Certainly `viewer.layers` is simple, but is everything that simple?

For instance what's the API equivalent of `cmd.levels`? Because `cmd.levels` takes the indexes printed by `cmd.layers` (like 0, 1, 2...). So they kind of work together?

I do like the idea of different levels/styles of access for different audiences or purposes:
1) The real first-class GUI
2) Debug menu, or any developer tools. Real GUI but off by default.
3) Simple flat set of IPython commands (this PR?)
4) The real API

I think the question is should 3 really exist? Or it just a question of designing 4 nicely for interactive use?

napari/_commands/_layers.py
```diff
@@ -0,0 +1,248 @@
...
+        ----------
+        data
+            Layer data.
+
+        Returns
+        -------
+        int
+            The number of levels of the data.
+        """
+        if isinstance(data, list):
```
sofroniewn, line 122: For shape layers this causes the number of shapes to be returned, which I'm not sure is the desired behavior. It's hard to think how to get the `levels` in a clear way from `data` all our layer types in the same way. You could look at the `layer.multiscale` property and if `False` say `1` and if `True` look at the length of `layer.level_shapes`. I will point out that attribute only currently exists on Image and Labels layers which are the only ones that can be multiscale.

napari/_commands/_layers.py
```diff
@@ -0,0 +1,248 @@
...
+        )
+        for i, layer in enumerate(self.layers):
+            self._add_row(i, layer)
+
+    def _get_shape_str(self, data):
+        """Get shape string for the data.
+
+        Either "NONE" or a tuple like "(10, 100, 100)".
+        """
+        if isinstance(data, list):
```
sofroniewn, line 100: you need a catch here for a tuple which is currently the layer type of our `surface` layer (run `examples/add_surface_2D.py` to see a very simple one. The data is a tuple of (vertices, faces, values), and so you could do `data[0].shape` for the shape of the vertex array

napari/_commands/_layers.py
```diff
@@ -0,0 +1,248 @@
...
+            if len(data) == 0:
+                return "NONE"  # Shape layer is empty list?
+            else:
+                return f"{data[0].shape}"  # Multi-scale
```
sofroniewn, line 104: This will also return the shape of the first `Shape` in a shapes layer. I guess I'd rather return the number of shapes (i.e. length of the list in this case).

We actually have a `layer.shape` property, but it is being deprecated/ not what you want to use. We could change it now though so it was what you wanted to use

napari/_commands/_layers.py
```diff
@@ -0,0 +1,248 @@
...
+        layer_id : int
+            The layer id (from view.cmd.layers).
+        layer : Layer
+            The layer itself.
+        """
+        layer_type = type(layer).__name__
+        num_levels = self._get_num_levels(layer.data)
+        shape_str = self._get_shape_str(layer.data)
+        data_type = _get_type_str(layer.data)
+        size_str = _get_size_str(layer.data)
```
sofroniewn, line 140: Given my other comments above I think these functions/ calculations should live on the layers. It would also give you a bit more flexibility deal with all the corner cases of the different layers. i.e. I don't think file should be worried about differences between Surface/ Shapes/ and Image.

Instead you could maybe put the layer specific stuff in something `layer.info` and get back a tuple of `(layer.name, layer.layer_type, layer.data_type, layer.num_levels, layer.shape, layer.size)` where those things can be defined in the base layer and then modified as needed in layers like `Surface` or `Shapes`.

Alternatively you could use the `layer.__str__` to make the layer row be nicely formatted and then just compose those rows into a table here, adding the `layer_id` as needed. At that point though you might want `viewer.layers.__str__` to do that composition for you and just print the nice table too.

We could still keep the `cmd` though if you need to have more complex functionality. Or you could maybe one day we might have a `layer.levels` object and `layer.levels.__str__` that then gives you everything you need.

How does all this sound?

napari/_commands/_tests/test_commands.py
```diff
@@ -0,0 +1,60 @@
...
+def test_one_layer(make_test_viewer, capsys):
+    """Test cmd.layer with one layer."""
+    viewer = make_test_viewer()
+    data = np.random.random((10, 15))
+    viewer.add_image(data, name="pizza")
+    viewer.cmd.layers
+    out, _ = capsys.readouterr()
+    assert "pizza" in out
+    assert "(10, 15)" in out
+
```
sofroniewn, line 32: We might also want to add a test that cycles through all our standard `layer_data` examples, see here https://github.com/napari/napari/blob/master/napari/_tests/utils.py#L8-L28 and https://github.com/napari/napari/blob/c2c6b8e32f79397225c8883ce03a8900b8708a94/napari/layers/_tests/test_serialize.py#L4-L13 for usage. This would catch the error with the surface layer that happens now (see comments above).

sofroniewn (commented):
I just went through @pwinston. I'm generally on board. I really like how it looks when printed, and I like the vendoring of `_humanize`. I did find some edge cases where things didn't work or work as expected (for Surface layers or Shapes layers, see comments above) and worried that too much layer specific logic might need to creep through into the `_commands` folder to fix them all.

We might want to consider going for a middle ground/ hybrid approach between "nicely formatted tables as __str__ of objects" and where we are now with the `cmd` function but some layer specific logic inside it by making the layer produce all the info it needs for the `cmd` function in a very simple way (possible using the `__str__` of an `info` property, see my detailed comments above).

 Hope these comments and perspective help!!

pwinston:
I think having the layer types produce their own formatted strings might make sense. And it might make sense to have the `__str__` method of `LayerList` combine them all. But neither of those changes are needed for #1510. Really these were the requirements for #1510:

1. List async status of image layers 
2. Toggle async settings of image layers.

Creating this PR with no async code was meant to slim down #1510. But it's clear this PR is not fully baked as a general text interface to all layers. I would suggest:

1. We put the text code back into #1510 but we only support `viewer.cmd.async` and `viewer.cmd.set_async` so it's totally async specific.
2. We either close this PR or leave it open "for later". Where the future task is "take the text stuff created for async, and use in wider ways for all layer types."

This will also let us get experience with actually using the async text interface. While people really like the idea are saying let's do it widely, if we actually had experience using it for #1510 that would more definitively motivate the wider work. If it worked well there.

Also this wider text interface is something anyone could take on, there's no graphics/async part to it at all. Although I'm happy to circle back and do it.

sofroniewn:
> 1. We put the text code back into #1510 but we only support viewer.cmd.async and viewer.cmd.set_async so it's totally async specific.
> 2. We either close this PR or leave it open "for later". Where the future task is "take the text stuff created for async, and use in wider ways for all layer types."

I'm ok with this approach.

> Also this wider text interface is something anyone could take on, there's no graphics/async part to it at all. Although I'm happy to circle back and do it.

This sounds good to me too.

jni:
Cool, I like where this has settled.

fwiw re "levels", I think that points to us creating a multiscale object, ie rather than just having a list of array-likes, and making the printing of that more sophisticated. In your summary of the different API levels, I do think that we should aim for our Python API to be the interface. Layers like cmd, to me, point to something missing in the API.

just as an aside, `__str__` is called on `print(obj)`, otherwise just typing `obj` gives you the `__repr__`. There is a question of whether there is an IPython equivalent of `_repr_html_`, `_repr_png_` etc that Jupyter notebooks have. Turns out there is, see: https://ipython.readthedocs.io/en/stable/api/generated/IPython.display.html#IPython.display.display, and `_repr_pretty_`. Though it is used in the doctests, but not in the "complete" list later on in the docstring. @Carreau is this an oversight?

Carreau:
> @Carreau is this an oversight?

Yes, likely, it should be in the docstring.

There is also [`_ipython_display_`](https://ipython.readthedocs.io/en/stable/config/integrating.html#MyObject._ipython_display_) which is not that well documented for even more flexibility.

`display` is injected in builtins, so no need to import it. 

And _if_  you want to extends objects you don't control , or want a `_repr_napari_`, you can also teach do that with [custom formatter](https://ipython.readthedocs.io/en/stable/config/integrating.html#formatters-for-third-party-types).
(or you can use that if you want to switch the types of rerp that an object has)

I haven't read the whole PR yet, but will see if there is anything to add.

napari/_commands/_processor.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def __init__(self, layerlist):
+        self.layer_commands = LayerCommands(layerlist)
+
+    def __repr__(self):
+        return HELP_STR
+
+    @property
+    def help(self):
+        print(HELP_STR)
```
Carreau, line 28: You do want to be careful with side effect on property it may be triggerd on tab completions.

Carreau:
for terminal pretty printing, tables and stuff, look at https://github.com/willmcgugan/rich

Carreau:
> for terminal pretty printing, tables and stuff, look at [willmcgugan/rich](https://github.com/willmcgugan/rich)

Note: They do have , like IPython a `__rich__` repr method, which I'm thinking of integrating as well into IPython, but haven't have enough time to look into it yet.

Carreau:
finally it's late and my brain is mush, but aren't those viewer command like tryin to do [custom magics](https://ipython.readthedocs.io/en/stable/config/custommagics.html) ?

jni:
Thanks for all that @Carreau! :pray:

pwinston:
Custom magics sounds very promising @Carreau. It sounds it gives you CLI-style commands with "CLI syntax" for arguments and flags, but within the IPython console.

If I'm reading it right, we could have a `%napari` magic function and then commands like:
```
%napari layers --visible --details
```
I think that CLI-style is better for interactive use than Python function syntax:
```
viewer.cmd.layers(visible=True, details=True)
```

Although this is totally opposite to @jni's feeling that there should be no "commands" at all, that the API should be simple enough and people just always interactive with napari through the API.

I think this is a product design question. The anti-CLI argument is there are only two types of users: GUI-users and "full programmers". The pro-CLI argument is there are three types of users, GUI-users, power users who will use the command line but not the API, and then full programmers. And perhaps that even for full programmers these CLI commands could be simpler/flatter/easier than the API for interactive use, like inspecting tables of data and simple configuration.

If napari has tons of users, I think it's more likely that middle category of power users will exist, who don't want to learn the full API. Maybe it requires experimenting with both approaches to see which feels most natural.

Carreau:
You don't **have to** have cli syntax, but you could I think having the API is nice, but you might want to wrap those API in magics for convenient access, one things magics do have access that API do not is a reference to the IPython shell, so you can do things you can't to the environment (loopup in globals and stuff).

The other thing is those magics have access to the full expressions the user typed, so you can have macro-like things. You could do for examples

```
%viewer add x + y
```

and translate it to

```
viewer.layer.add(x+y, title='x+y')
```

which can be way more verbose in Python.

magics also vive in a different namespace; so user can reassign the name. the % at the beginning is necessary only when the magic is shadowed; typical example is `%matplotlib inline`; you can drop the % until matplotlib is not imported in current namespace.

tlambert03:
> magics also vive in a different namespace; so user can reassign the name. the % at the beginning is necessary only when the magic is shadowed; typical example is %matplotlib inline; you can drop the % until matplotlib is not imported in current namespace.

Oh I did not realize _that_.  That's super cool

pwinston:
No `%` required that's great. That was my only reservation, didn't love that `%`. It seems like this was very much designed to fashion a CLI within IPython. And why would people want to do that? Given you can do anything with real Python?

I think the reason is the "CLI syntax"  is simpler with less typing than calling real Python functions. But beyond syntax I think the CLI has to be flat. @jni suggested above we could return a "multiscale object" from each multi-scale layer, then you interact with that object. That's very programmery, with the CLI mindset you might have a command like `napari levels 4` to show levels for layer 4.

Interesting to think about how/why CLI is different from "real" programming.

pwinston:
There is some useful discussion in this PR but there is zero chance we are merging this code, so we should open a new issue or PR if there is a viable new propsal.

This PR has an async-only version of the commands: https://github.com/pwinston/napari/pull/22

Once nice idea from this PR's comment thread is using IPython magic commands.

PR #1534, opened 2020-08-07
Add test for theme changing

### LINKED ISSUE #669: switching themes causes incomplete updates

tlambert03:
# Description
closes #669 and adds a test to make sure that >99.5% of the pixels in a screenshot have changed when switching between dark and light themes. 

## Type of change

- [x] Test

# How has this been tested?

- [x] takes two screenshots and looks for the ratio of non-equal pixels between dark/light themes

## Final checklist:

### DIFF
--- a/napari/_tests/test_viewer.py
+++ b/napari/_tests/test_viewer.py
@@ -152,12 +152,21 @@ def test_update(make_test_viewer):
 
 
 def test_changing_theme(make_test_viewer):
-    """Test instantiating viewer."""
+    """Test changing the theme updates the full window."""
     viewer = make_test_viewer()
     assert viewer.palette['folder'] == 'dark'
 
+    screenshot_dark = viewer.screenshot(canvas_only=False)
+
     viewer.theme = 'light'
     assert viewer.palette['folder'] == 'light'
+
+    screenshot_light = viewer.screenshot(canvas_only=False)
+    equal = (screenshot_dark == screenshot_light)[..., :3]
+    np.count_nonzero(equal)
+
+    # (less than 1 in 100,000 pixels are the same)
+    assert (np.count_nonzero(equal) / equal.size) < 1e-5, "Themes are similar!"
 
     with pytest.raises(ValueError):
         viewer.theme = 'nonexistent_theme'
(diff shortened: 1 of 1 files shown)


sofroniewn:
This is great, but test is currently failing. I think we need to add the `show=True` to the making of the viewer and then with windows and CI skip like we do here https://github.com/napari/napari/blob/master/napari/_tests/test_with_screenshot.py#L7-L10. I might move this test to the `test_with_screenshot.py` folder too, just to keep all those tests together. Thoughts?

tlambert03:
> This is great, but test is currently failing. I think we need to add the show=True to the making of the viewer and then with windows and CI skip like we do here 

I think it works ok without showing... it was failing because the [ratio of non-different pixels](https://cirrus-ci.com/task/6231021521731584?command=test#L1136-L1137) was _slightly_ higher on CI than it was locally. I just tweaked the algorithm, and it's working now.

> I might move this test to the test_with_screenshot.py folder too, just to keep all those tests together. Thoughts?

I kinda like keeping the theme test where it already was in `test_viewer`...  it _uses_ the screenshot as a tool, but it is not testing the screenshot in any way... so it feels a bit out of place in the screenshot file.

sofroniewn:
> I think it works ok without showing... it was failing because the ratio of non-different pixels was slightly higher on CI than it was locally. I just tweaked the algorithm, and it's working now.

ah nice, my bad. i thought we needed to show for all screenshots (see below)

> I kinda like keeping the theme test where it already was in test_viewer... it uses the screenshot as a tool, but it is not testing the screenshot in any way... so it feels a bit out of place in the screenshot file.

the tests in screenshot file also just use screenshot as a tool (mainly for vispy rendered things). i maybe misunderstood what we needed to `show` and there could be some things in that file that don't actually need to `show`.

sofroniewn (approved):
LGTM! thanks for adding

Issue #1519, opened 2020-08-04
One color per multiple labels results in gradients


vzinche:
## 🐛 (Probably a) Bug

Related to #716 

## Problem 
I have a segmentation image, which I want to color by some metadata. Let's say, each segment id has a property A or B assigned. To color segments by properties, I supply a dictionary to the labels layer that looks like:
`color_dict =  {0: 'black', 1: 'red', 2: 'red', 3: 'blue', 4: 'blue', etc}`
`viewer.add_labels(segmentation, color=color_dict)`
Note that in comparison to #716, I have one color per multiple labels. 

That is a sample resulting image I get:
![napari_color](https://user-images.githubusercontent.com/32451366/89294702-2c7fd100-d660-11ea-95a4-1aa350079a2d.png)

As you can notice, there are gradients of red and blue present (the purple cells). I've tried the very same with Image layer and vispy Colormap as described [here](https://forum.image.sc/t/specify-labels-colors/31534) and I still get the gradients. Where do they come from? It is not such a big issue for two colors, but once I have more it gets worse. 

## Expected behavior
Only blue, red and black colors on the image. No gradients

## Minimal reproducible example

```
def get_weird_gradients():
    sample_array = np.arange(100)
    np.random.shuffle(sample_array)
    labels_image = sample_array.reshape(10,10)
    colors = ['black', 'red', 'blue']

    color_map = {i: colors[np.random.randint(3)]
                 for i in range(100)}

    with napari.gui_qt():
        viewer = napari.Viewer()
        viewer.add_labels(labels_image, color=color_map) 
```

## Environment

napari: 0.3.6
Platform: Linux-4.17.9-041709-generic-x86_64-with-debian-buster-sid
Python: 3.7.6 | packaged by conda-forge | (default, Jun 1 2020, 18:57:50) [GCC 7.5.0]
Qt: 5.12.5
PyQt5: 5.12.3
NumPy: 1.18.4
SciPy: 1.4.1
Dask: 2.17.2
VisPy: 0.6.4

GL version: 4.6.0 NVIDIA 440.100
MAX_TEXTURE_SIZE: 32768

Plugins:
- napari-plugin-engine: 0.1.5
- svg: 0.1.3

## Additional context
Alternatively, is there any better way to color by metadata? 

cc @constantinpape

constantinpape:
> Alternatively, is there any better way to color by metadata?

For some context here, @vzinche and me have been looking into similar problems of coloring labels by data associated with the labels.

[...]

PR #1537, opened 2020-08-08
addressed colormap control color with same label color

liu-ziyang:
# Description
addressed color difference when multiple labels using the same color by reducing control colors

## Type of change

# References
https://github.com/napari/napari/issues/1519

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/layers/labels/_tests/test_labels.py
+++ b/napari/layers/labels/_tests/test_labels.py
@@ -294,15 +294,15 @@ def test_custom_color_dict():
     np.random.seed(0)
     data = np.random.randint(20, size=(10, 15))
     layer = Labels(
-        data, color={1: 'white', 2: 'red', 3: 'red', 4: 'blue', 5: 'blue'}
+        data, color={1: 'white', 2: 'red', 3: 'blue', 4: 'red', 5: 'blue'}
     )
 
     # test with custom color dict
     assert type(layer.get_color(2)) == np.ndarray
     assert type(layer.get_color(1)) == np.ndarray
     assert (layer.get_color(1) == np.array([1.0, 1.0, 1.0, 1.0])).all()
-    assert (layer.get_color(2) == layer.get_color(3)).all()
-    assert (layer.get_color(4) == layer.get_color(5)).all()
+    assert (layer.get_color(2) == layer.get_color(4)).all()
+    assert (layer.get_color(3) == layer.get_color(5)).all()
 
     # test disable custom color dict
     # should not initialize as white since we are using random.seed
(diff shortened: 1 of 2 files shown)


napari/utils/colormaps/colormaps.py
```diff
@@ -133,10 +133,17 @@ def color_dict_to_colormap(colors):
         Mapping of Label to color control point within colormap
     """
 
-    colormap = Colormap([color for label, color in colors.items()])
+    control_colors = np.unique(
+        [color for label, color in colors.items()], axis=0
+    )
```
jni:
```suggestion
    control_colors = np.unique(list(colors.values()), axis=0)
```

napari/utils/colormaps/colormaps.py
```diff
@@ -133,10 +133,17 @@ def color_dict_to_colormap(colors):
...
+    colormap = Colormap(control_colors)
     label_color_index = {}
     for i, (label, color) in enumerate(colors.items()):
-        label_color_index[label] = i / (len(colors) - 1)
+        for j, control_color in enumerate(control_colors):
+            if (control_color == color).all():
+                label_color_index[label] = j / (len(control_colors) - 1)
+                break
+
```
jni: This seems sub-optimal. I'd suggest the following:

```python
control2index = {
    tuple(ctrl): i / (len(control_colors) - 1)
    for i, ctrl in enumerate(control_colors)
}
label_color_index = {
    label: control2index[tuple(color)]
    for label, color in colors.items()
}
```

Normally there'd be some concerns about equality of floating point numbers, but since we are using `np.unique` I don't think they apply here.

napari/layers/labels/_tests/test_labels.py
```diff
@@ -293,12 +293,16 @@ def test_custom_color_dict():
     """Test custom color dict."""
     np.random.seed(0)
     data = np.random.randint(20, size=(10, 15))
-    layer = Labels(data, color={1: 'white'})
+    layer = Labels(
```
jni, line 296: I feel like for a really good test we need a much bigger dataset, and non-consecutive labels should map to the same color. ie

```
{1: 'white', 2: 'red', ..., 18: 'red', 19: 'white'}
```

You'd probably want to construct this programmatically! :grimacing: Although it seems we're ok with missing entries so I guess it doesn't matter.

jni (dismissed):
@ziyangczi thanks for tackling this! :tada: I've made some suggestions for tidying up the code a bit, and making the test a bit more stringent, but the approach is great!

napari/layers/labels/_tests/test_labels.py
liu-ziyang: switched to a binary numbering system :)

sofroniewn (approved):
LGTM! Thanks @ziyangczi

Issue #1536, opened 2020-08-08
RuntimeWarning when creating empty points


tlambert03:
## 🐛 Bug
not a bug exactly, but a runtime warning that users will now see when creating empty points after #1360 

```python
In [1]: viewer = napari.Viewer()     

In [2]: viewer.add_points()    
napari/napari/components/layerlist.py:209: RuntimeWarning: All-NaN axis encountered
  min_v = np.nanmin(
napari/napari/components/layerlist.py:212: RuntimeWarning: All-NaN axis encountered
  max_v = np.nanmax(
```

@sofroniewn, would it be possible to catch the all-NaN cases before running np.nanmax/min here?

https://github.com/napari/napari/blob/2315443d9bd4fb6b8cc30967cf433d017f7c8ca2/napari/components/layerlist.py#L209-L214

sofroniewn:
Yup, that shouldn't have happened. I'll look into fixing. Thanks for the catch.

PR #1532, opened 2020-08-07
World coordinates take 3

sofroniewn:
# Description
This PR will supersede #1276 and add support for a world coordinate system.

It may also drop the `layer.dims` property which I think would be a dramatic simplification of our codebase. I'm actually probably not too far away from being able to achieve that cleanly in this PR now. 

There's much still to do, but wanted to get the PR started

## Type of change

- [ ] Bug-fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# References

# How has this been tested?

- [ ] example: the test suite for my feature covers cases x, y, and z
- [ ] example: all tests pass with my change

## Final checklist:
- [ ] My PR is the minimum possible work for the desired functionality
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #1538, opened 2020-08-08
Clean up all warnings in tests

tlambert03:
# Description
It would be nice if the warnings section of our tests were more informative, but it's filled with so much stuff at the moment that i suspect no one really looks at it.  This PR cleans up all of the warnings that are due to our code, (with the exception of #1536 ... which I'd like @sofroniewn to weigh in on), and adds a few explicit `filterwarnings` decorators where the warnings are due to external libraries that are mostly just adding noise to our reports.

## Type of change

- [x] tests: quality of life


napari/layers/vectors/_tests/test_vectors.py
```diff
@@ -121,6 +121,7 @@ def test_random_3D_vectors_image():
     assert layer._view_data.shape[2] == 2
 
 
+@pytest.mark.filterwarnings("ignore:Passing `np.nan`:DeprecationWarning:numpy")
```
sofroniewn, line 124: Is this not fixable?

napari/layers/vectors/_tests/test_vectors.py
```diff
@@ -409,6 +410,7 @@ def test_edge_color_map_non_numeric_property():
         layer.edge_color_mode = 'colormap'
 
 
+@pytest.mark.filterwarnings("ignore:elementwise comparis:FutureWarning:numpy")
```
sofroniewn, line 413: I this not fixable?

sofroniewn (commented):
I definitely love the clean up idea, so thanks for starting this @tlambert03!! And it's true there are so many warnings right now I don't even look at them regularly :-(

I think its fine to ignore our user generated warnings, but I'm a little nervous about some of the deprecation ignores. In particular it seems like some of the numpy related ones we should still be able to fix, also maybe the skimage one?

I'd probably also leave the console ones in and make a dedicated issue for console cleanup filed under "bug", the console does still cause compatibility issues and fixing those warnings could make them go away (might be a long shot?)

Thoughts?

tlambert03:
will address your two comments, but more generally: all of these could probably be fixed somewhere... most would be upstream PRs.  I don't see any reason to be nervous about them though, no one is actually looking at these warnings anyway right? :joy: ... now at least we'll notice when we introduce a new one, while we address the (already ignored) ones.
tracking these down would make for a good first issue.

napari/layers/vectors/_tests/test_vectors.py
tlambert03: probably could/should be addressed in #1536 
the issue is that with empty vectors, `dims.range` gives `[(nan, nan, 1.0), (nan, nan, 1.0), (nan, nan, 1.0)]`... so these lines in dims:
https://github.com/napari/napari/blob/2315443d9bd4fb6b8cc30967cf433d017f7c8ca2/napari/components/dims.py#L270-L275

cause this warning
```
DeprecationWarning: Passing `np.nan` to mean no clipping in np.clip has always been unreliable, and is now deprecated. In future, this will always return nan, like it already does when min or max are arrays that contain nan. To skip a bound, pass either None or an np.inf of an appropriate sign.
```

napari/layers/vectors/_tests/test_vectors.py
tlambert03: caused at line `layer.edge_color = 'vector_type'`
```pytb
FutureWarning                             Traceback (most recent call last)
<ipython-input-8-702e23f49dcf> in <module>
      1 # switch to color cycle
      2 layer.edge_color_mode = 'cycle'
----> 3 layer.edge_color = 'vector_type'
      4 edge_color_array = transform_color(color_cycle * int((shape[0] / 2)))

~/Dropbox (HMS)/Python/forks/napari/napari/layers/vectors/vectors.py in edge_color(self, edge_color)
    384                 self._edge_color_mode = new_mode
    385             self._edge_color_property = edge_color
--> 386             self.refresh_colors()
    387
    388         else:

~/Dropbox (HMS)/Python/forks/napari/napari/layers/vectors/vectors.py in refresh_colors(self, update_color_mapping)
    438                     # and update_color_mapping==False
    439                     edge_color_cycle_keys = [*self.edge_color_cycle_map]
--> 440                     props_in_map = np.in1d(
    441                         edge_color_properties, edge_color_cycle_keys
    442                     )

<__array_function__ internals> in in1d(*args, **kwargs)

~/miniconda3/envs/napdev/lib/python3.8/site-packages/numpy/lib/arraysetops.py in in1d(ar1, ar2, assume_unique, invert)
    578             mask = np.zeros(len(ar1), dtype=bool)
    579             for a in ar2:
--> 580                 mask |= (ar1 == a)
    581         return mask
    582

FutureWarning: elementwise comparison failed; returning scalar instead, but in the future will perform elementwise comparison
```

would prefer that be addressed in a different PR

tlambert03:
> I'd probably also leave the console ones in and make a dedicated issue for console cleanup filed under "bug"

just want to reiterate that a lot of these warnings, particularly the console ones are completely independent of our code.  These are upstream warnings that require upstream PRs... personally, I'd prefer to quit them down so we actually notice the warnings that are related to things we're doing.

napari/layers/vectors/_tests/test_vectors.py
sofroniewn: this is changing in #1532 too

sofroniewn (approved):
> now at least we'll notice when we introduce a new one, while we address the (already ignored) ones.
tracking these down would make for a good first issue.

yeah I'm fine with that, i just don't want us to forget about these that's all.

i'm approving, but curious what @jni thinks before merge

tlambert03:
> i just don't want us to forget about these that's all.

agreed.  I don't see this as sweeping these things under the rug.  quite the opposite:  It's _already_ swept under the rug by nature of there being so many of them.  now at least we can search "filterwarnings" and see a list of exactly which tests have warnings, what type of warnings, and what upstream packages are responsible.    And if new ones arise (which are arguably more important), then they are visible.  And if someone wants to get deeper into warnings, all of this is very [configurable with pytest](https://docs.pytest.org/en/stable/warnings.html).

sofroniewn:
Got a zulip green light from @jni will merge this now

PR #1539, opened 2020-08-08
Better support for multiple translated arrays

### LINKED ISSUE #659: Axis sliders with multiple translated arrays
## 🐛 Bug

The 'translate` kwarg to `view_image` does weird stuff.
```python
offset = 100
t1 = (0,0,0)
# shifted in "Z"
t2 = (offset,0,0)
data = np.random.randn(offset,offset,offset)
# view the unshifted array
v = napari.view_image(data, translate=t1, name='fixed', blending='additive')
# view the same data with a shift
v.add_image(data, translate=t2, name='moving', blending='additive') 
```

This renders both arrays overlapping in the same region in viewer space, which seems weird, given that I translated the second array in Z expressly so that they would not overlap.

I would expect the "roll dimensions" button to offer me a "side view" of the arrays that represents the Z offset, but this is not the case -- all the views afforded by the "roll dimensions" button show the arrays overlapping in space.

The volume rendering button does show the arrays in the expected shifted positions in space. Interestingly, if I roll the displayed dimensions once, then engage volume rendering, then disable volume rendering, I see the two arrays shifted as expected. 

Presumably this can all by avoided by padding the shifted array with 0s, but that's far from ideal....

## Expected behavior
When a single array is being rendered, the axis sliders are both incrementing the slice of the data to be displayed (i.e., changing coordinates in the reference frame of the array) AND (implicitly) moving the clipping plane of the data (i.e., changing coordinates in the reference frame of the viewer area). But these are actually two very distinct operations, as I think the two-shifted-arrays example demonstrates. So maybe the job of the axis slider should be formally changed from "change the index of the arrays" to "move the clipping plane through viewer space, which happens to contain arrays". Unfortunately, this would break a lot of stuff! So I'm not really sure what's right here.  

## Environment

napari, 0.2.4
Qt, 5.9.6
NumPy, 1.17.2
SciPy, 1.3.1
VisPy, 0.6.1
scikit-image, 0.16.2

Platform: Linux-4.15.0-54-generic-x86_64-with-debian-buster-sid
Python:   3.7.4 (default, Aug 13 2019, 20:35:49)  [GCC 7.3.0]
Backend:  PyQt5
GL version:  '4.6.0 NVIDIA 430.26'
MAX_TEXTURE_SIZE: 32768 (is this a lot?)

sofroniewn:
# Description
This PR supersedes #1532 and #1276 and now implements a global world coordinates.

Most notably the `points` and `range` properties of the `dims` model are now in world coordinate. These can be positive, negative, floats etc. The `step` and `nsteps` properties are in the slider coordinates and are non-negative integers.

The dims model understands how to convert a `point` in world coordinates to a `step` in the slider and vice-versa.

Given a point in world coordinates the `layer` understands how to convert that back to data coordinates for slicing. This is done in the new `_slice_indices` property.

Given slice indices the layer can recognize them as out of bounds for a dense array like layer (image/ labels) or ignore them for points/ shapes.

I've only tried this PR with image so far, so this is still a WIP.

This PR will actually make it pretty easy to pull off the dims model from the individual layers, but I might wait to do that in a refactor only PR.

I need to add comprehensive tests. There are still maybe so "rounding" problems at the very top/ bottom of stacks when scaled. I also need to check the roll dims works fine and the camera setting appears like it needs a fix or two. Once this is in I am going to return to #854 the camera model, which I can't wait for.

To see the effect of the new behavior you can try

```python
blobs_raw = data.binary_blobs(length=64, n_dim=3, volume_fraction=0.1)
# two arrays should be "stacked on top" of each other in Z
viewer.add_image(blobs_raw, colormap='blue')
viewer.add_image(blobs_raw, translate=[64, 0, 0], blending='additive', colormap='red')
```

which gives

world_coords4_1

note now how you scroll first through the blue then through the red in z while in 2d rendering mode. This most concretely fixes #659 from @d-v-b. I still need to check that things like #1369 from @Czaki are fixed too.

This might still need more work before it is ready to play with, but an end here to finally adding world coordinates should be in sight!! 

## Type of change

# References

# How has this been tested?

## Final checklist:


Czaki:
You could take tests from #1488 for check round rules for Dims. After finish this PR it may supress #1488 also.

sofroniewn:
> You could take tests from #1488 for check round rules for Dims. After finish this PR it may supress #1488 also.

Yes @Czaki I was just looking at that. This PR drops the `dims.indices` property, replacing it with the `layer._slice_indices`, but I still need to do a rounding operation inside there. I'm actually taking a floor operation as it gives the best results with data like this

```python
blobs_raw = data.binary_blobs(length=64, n_dim=3, volume_fraction=0.1)
viewer.add_image(blobs_raw, colormap='blue')
viewer.add_image(blobs_raw[::2], scale=[2, 1, 1], blending='additive', colormap='red')
```

causing them to nicely overlap for every value of the slider. See https://github.com/napari/napari/pull/1539/files#diff-9ef2d7d09674e6ab30a1456a11fd74a8R445-R446

It also looks to me like #1369 is now fixed - can you confirm that too?

I'm going to start working on tests now (fixing broken/ adding new) and looking at layers aside from image, but it seems to me like most basic behaviour is now working, so trying it out in the real world @Czaki @d-v-b @GenevieveBuckley @jni would be appreciated.

After this merges I think there will be more clean up PRs where we might simplify some things organizationally, maybe drop `layer.dims`, but this PR should now have the desired functionality with all the coordinate system bugs fixed at the end of it!!

Czaki:
I'm during vacation till next saturday but i will try to find time and check it.

sofroniewn:
I have gone through and fixed all the tests. I've also taken the plunge to make `layer.dims` private `layer._dims`. A user should no longer interact with `layer.dims` and instead call `layer._slice_dims` (which we might want to rename, and which will ultimately be taken off the layer).

I will probably add more dedicated world coordinates tests, but this is ready for some initial review @kevinyamauchi @tlambert03 @jni

jni:
@sofroniewn here's a little test that I haven't formalised, perhaps you have ideas on how to do so:

```python
from skimage import data, color
import napari


with napari.gui_qt():
    viewer = napari.view_image(astro := data.astronaut(), rgb=True)
    viewer.add_image(
        gray_astro := color.rgb2gray(astro[::2, ::2]),
        scale=[2, 2],
        opacity=0.7,
    )
```

Now, zoom in on the top left corner, and you'll notice that the coordinates don't quite line up. Move your cursor to a position near a pixel transition, and then switch layers: the coordinate will be different for the different layers:

Screen Shot 2020-08-10 at 1 19 27 pm

Screen Shot 2020-08-10 at 1 19 37 pm

This is a sad consequence of our choice (which I still defend, but complicates things in this use case) of using 0, 0 as the pixel centre coordinates. Then, when layers have different scales, if you want to make the squares line up, you have to have a translation on the lower-resolution layers.

Right now, the squares line up, probably because the -0.5 correction to go from our coordinates to VisPy coordinates is done *before* scaling.

But, to have "real" world coordinates, I think the coordinates under the cursor should be constant regardless of which layer is selected.

sofroniewn:
> This is a sad consequence of our choice (which I still defend, but complicates things in this use case) of using 0, 0 as the pixel centre coordinates. Then, when layers have different scales, if you want to make the squares line up, you have to have a translation on the lower-resolution layers.
>
> Right now, the squares line up, probably because the -0.5 correction to go from our coordinates to VisPy coordinates is done before scaling.
>
> But, to have "real" world coordinates, I think the coordinates under the cursor should be constant regardless of which layer is selected.

Ah yes - ok I have not attempted to address any of that in this PR. I think we should view this PR then as scoped to fix #659 and then press on with more world coords PRs afterwards. I'll rename this PR as such

d-v-b:
This PR definitely addresses the case where arrays are translated by an integer quantity, but stuff gets a little wonky if you shift an array by a floating point quantity like 0.1 -- when I modify nick's example with floats instead of integer shifts, I get non-deterministic behavior from the axis slider, and inaccurate position information from the slider status area. I'm not sure if fixing this behavior is in-scope for this PR though.

sofroniewn:
> This PR definitely addresses the case where arrays are translated by an integer quantity, but stuff gets a little wonky if you shift an array by a floating point quantity like 0.1 -- when I modify nick's example with floats instead of integer shifts, I get non-deterministic behavior from the axis slider, and inaccurate position information from the slider status area. I'm not sure if fixing this behavior is in-scope for this PR though.

That's useful feedback, thanks. I'll try and reproduce. I can try and make some improvements in this PR, I'll report back. If you had a use case where it goes wrong and is obvious what the correct behaviour should be feel free to share!!

d-v-b:
>If you had a use case where it goes wrong and is obvious what the correct behaviour should be feel free to share!!

Here's the example where stuff got wonky:

```python
import numpy as np
im = np.ones((3,3,3))
im[1,1,1] = 0
viewer.add_image(im, colormap='blue', rgb=False)
viewer.add_image(im, translate=[0.1, 0, 0], blending='additive', colormap='red', rgb=False)
```

The slider breaks here in a funny way -- the slider status indicates that there are 3 total planes (0 - 2), i.e. starting at slice 0 and incrementing the slider twice should take you to the end of the range. However, I can use my arrow keys to increment the slider _three_ times, but the last increment doesn't update the slider status (it says 2|2 before the last increment, and after it). But this extra incrementing of the slider has somehow changed the inner state of the slider such that increments / decrements show me different slices of data. It's kind of hard to describe, it might be best just to play with my example.

As for what the "correct" behavior would be, for two arrays with a relative offset of .1, it seems like the slider for that axis should increment with a step size of at least 0.1. This requires a bit of a re-implementation of sliders -- if sliders operate in world coordinates (which they should), then the effect of changing a slider cannot just be array indexing at some integer index. Instead, the effect of changing a slider should be data _resampling_ at some floating point _position_. But I can understand if that doesn't happen in this PR.

sofroniewn:
> It's kind of hard to describe, it might be best just to play with my example.

Yup I can reproduce the "funky" behavior, agreed something needs to be fixed.

> As for what the "correct" behavior would be, for two arrays with a relative offset of .1, it seems like the slider for that axis should increment with a step size of at least 0.1.

There is now data at Z= [0, 0.1, 1, 1.1, 2, 2.1]. One option is that the slider now has increments of `0.1` and every option between `[0, 2.1]` is available, but most slider positions show the "same" data, i.e. a slider of `0.3` and `0.4` show the same as a `0` and `0.1`, when you hit `0.5` one transitions and the other doesn't.

Another option is to just have slider increments of `1.0` and only 3 states `[0, 1, 2]` and always round.

I'll give that first one a go. It will be important to think about images of different scales/ translations too, and also make sure we don't get in some crazy rounding error problems with translations of `0.9999` giving us increments of `0.0001`

sofroniewn:
Ok, I've cleaned up the "funky" behaviour with the extra slider positions, that was just a bug.

As to now whether the increment is `0.1` vs `1.0` for the above example and how to do the rounding, right now the increment is `1.0` and rounding is based on to the nearest int (note `0.5` gets rounded down to `0` by default by numpy).

If you want to see what a finer increment looks like in the above example you can do `viewer.dims.set_range(0, (0, 3.1, 0.1))`

It's not clear to me which is preferable, feedback welcome

> This requires a bit of a re-implementation of sliders -- if sliders operate in world coordinates (which they should), then the effect of changing a slider cannot just be array indexing at some integer index. Instead, the effect of changing a slider should be data resampling at some floating point position.

This has now been done. Sliders themselves are still discrete the slider (as all sliders in Qt are under the hood) and for convenience they are taken to have a min value of 0 a max value of N and a step size of 1. A given position of the slider is then mapped into a world coordinate space according the the relevant offsets and scalings, this value is now a float. That float is then mapped into the appropriate integer slicing indices for each layer according to the the inverse of its data to world transform. I think this now matches the expectations of @d-v-b and others right?

In the context of the above, the only question left is how fine to make the slider, and whether the fineness of the slider should take the translation into account.

Note that the numbers to the right of the slider are now in slider coordinates, and the full mapping goes slider -> world -> data.

More input welcome, i'd really like us to wrap at least this part up!!

sofroniewn:
Just adding here that to change the step size of a particular `axis` while keeping the world extent constant you can call

```python
viewer.dims.set_range(axis, (viewer.dims.range[axis][0], viewer.dims.range[axis][1], new_step))
```

sofroniewn:
This PR now has dedicated testing for the translation/ scaling in `napari/components/_tests/test_world_coordinates.py` including full slicing through every available z step. This was the last thing I wanted to add. It is now ready for more real world testing and final review.

kevinyamauchi:
@sofroniewn , this is awesome! I spent some time playing around with translating and scaling images and things seem to largely work as I would expect. I listed a couple of comments below and I will take a closer look tomorrow.

Do you think we can extend the translation to include the out of plane dimensions? For example if I have a 2D image, it would be great if I could translate it within a 3D volume. I suppose the trick is to specify which dimensions are the out of plane dimensions. Perhaps maybe the `scale` argument could be used? I think it's a bit of an odd usage of `scale`, but maybe  something like:

```python
# add a 2D image with the data parallel with axes 1 and 2.
# translate the image 10 units along axis 0
viewer.add_image(image, translate = [10, 0, 0], scale = [0, 1, 1])
```

Also, in the future, I think it would be nice if the displayed index for the current slider "step" could be float. For example, if I am scrolling through a movie acquired at 10 Hz, ideally each step could be 0.1 s or 100 ms.

Czaki:
It looks like this PR works in my case. 

And I need to prepare to upgrade to napari 0.3.7

sofroniewn:
> @sofroniewn , this is awesome! I spent some time playing around with translating and scaling images and things seem to largely work as I would expect. I listed a couple of comments below and I will take a closer look tomorrow.

Thanks @kevinyamauchi sounds good!!

> It looks like this PR works in my case.

Great, thanks fro checking @Czaki! 

> And I need to prepare to upgrade to napari 0.3.7

Yes, @Czaki i'm hoping we release fairly soon. At this point I want to get the current PRs labeled as 0.3.7 in and no more.

Czaki:
> > And I need to prepare to upgrade to napari 0.3.7
> 
> Yes, @Czaki i'm hoping we release fairly soon. At this point I want to get the current PRs labeled as 0.3.7 in and no more.

What I such change in this line?:
```
self.viewer.dims.events.axis.connect(self._view_changed, position="last")
```

Because this PR remove `axis` from EventGroup. It is renamed somehow?

sofroniewn:
> Because this PR remove axis from EventGroup. It is renamed somehow?

Try
```python
self.viewer.dims.events.step.connect(...)
```
There isn't a 1-1 emit correspondence, so if it doesn't work I can help more

Czaki:
> Try
> 
> ```python
> self.viewer.dims.events.step.connect(...)
> ```
> 
> There isn't a 1-1 emit correspondence, so if it doesn't work I can help more

I will try. Could you add 0.3.6 tag in your repository? this will simplify testing. now I have:
```
ERROR: partseg 0.11.2.dev5+g795e018.d20200817 has requirement napari>=0.3.6, but you'll have napari 0.1.dev821+gd8adf16 which is incompatible.
```

sofroniewn:
> Could you add 0.3.6 tag in your repository? this will simplify testing

This sounds like a very reasonable request. I actually don't know how to implement this. Maybe you can make a PR and then @jni can review. Sounds like it will be helpful to many people!

jni:
I'm confused — the 0.3.6 tag is present here, one just needs to fetch it? It's annoying that tags aren't fetched automatically, but there's little we can do on our end afaik? Other than maybe documenting it in our developer notes...

Czaki:
`pip install git+https://github.com/napari/napari.git` works well and provide proper version `Successfully installed napari-0.3.7.dev27+g80ac620` but when install from @sofroniewn fork `pip install git+https://github.com/sofroniewn/napari.git@world-coordinates4`  its result is `Successfully installed napari-0.1.dev821+gd8adf16` 

Today I found that  `pip install git+https://github.com/napari/napari.git@refs/pull/1539/merge` works but I'm not shure if everything works properly where there are reported merge conflicts  for this PR. 

My only request is to add tag in @sofroniewn orked repository.

d-v-b:
These changes introduce some strange interaction between translation and opacity settings: when using opacity < 1 for a layer that is translated along the slider axis, a mysterious gray pixel appears:

```python
import numpy as np
import napari

translate = (10,) * 3
with napari.gui_qt():
    data = np.random.randint(0,255, (10,10,10), dtype='uint8')
    viewer = napari.Viewer()
    viewer.add_image(data, name='raw', opacity=.5)
    viewer.add_image(data, translate=translate, name=f'translated_{translate}', colormap='blue', opacity=.5)

```

image

sofroniewn:
Following up to this https://github.com/napari/napari/pull/1539#issuecomment-675112449 from @kevinyamauchi   

> Do you think we can extend the translation to include the out of plane dimensions? For example if I have a 2D image, it would be great if I could translate it within a 3D volume. I suppose the trick is to specify which dimensions are the out of plane dimensions. 

```python
viewer.add_image(stent, name='stent')
viewer.add_image(np.expand_dims(stent[55], axis=0), name='2D_slice', translate=[55, 0, 0], colormap='blue')
```

The following sort of words to visualize a single 2D slice in 3D data. The blending/ colormaps/ z-ordering don't quite look right, which I think is something that @jni has commented on in the past, but I think we're off to an ok start with room for improvement!

Screen Shot 2020-08-23 at 7 53 15 PM

sofroniewn:
> These changes introduce some strange interaction between translation and opacity settings: when using opacity < 1 for a layer that is translated along the slider axis, a mysterious gray pixel appears:

Thanks for catching this @d-v-b, it is because when the slice indices were fully out of range of the data we used a 1^n zero cube, but we should do something else instead.

I have some merge conflicts to fix, and will look into fixing this then

sofroniewn:
@d-v-b I have fixed the single square transparency bug by adding a `layer._empty` property for when the slicing is out of the range of the layer data, and things now work as expected:

fix_hidden_square

I have also fixed merge conflicts.@jni this is now ready for review.

sofroniewn:
@Czaki I have just verified that this fixes #1369 and so can supersede #1488. 

@d-v-b I have also verified that this fixes #1572, see movie of playing below (note I was able to reproduce error in master)

world_ex_4

DragaDoncila:
Have tested this with our Sentinel satellite images and setting the scale and step_size works as expected, provided the visible scale parameters are 1 in the case of multiscale data. It's actually very nice! Love it.

napari/components/dims.py
```diff
@@ -353,28 +300,30 @@ def set_point(self, axis: int, value: Union[int, float]):
...
-        if self.point[axis] != value:
-            self._point[axis] = value
-            self.events.axis(axis=axis, value=value)
+        (min_val, max_val, step_size) = self._range[axis]
+        raw_step = (value - min_val) / step_size
+        self.set_step(axis, raw_step)
 
-    def set_interval(self, axis: int, interval: Sequence[Union[int, float]]):
-        """Sets the interval used for cropping and projecting this dimension.
+    def set_step(self, axis: int, value: int):
```
jni: I don't find this self-explanatory at all... From the name, I would expect this to be equivalent to `self.set_range(axis, *self.range[:2], step)`, ie setting the step size. Maybe `self.set_step_number`? Or `self.set_point(axis, pt, coordinates='data')`? (as opposed to `'world'`)

napari/components/dims.py
```diff
@@ -207,11 +184,7 @@ def ndim(self, ndim):
             # Range value is (min, max, step) for the entire slider
             self._range = [(0, 2, 1)] * (ndim - cur_ndim) + self._range
             # Point is the slider value if in point mode
-            self._point = [0] * (ndim - cur_ndim) + self._point
-            # Interval value is the (min, max) of the slider selection
-            # if in interval mode
-            self._interval = [(0, 1)] * (ndim - cur_ndim) + self._interval
-            self._mode = [DimsMode.POINT] * (ndim - cur_ndim) + self._mode
```
jni, line 214: I know we don't have other modes right now, but removing these lines seems a bit dangerous? Shouldn't the point be set as well?

napari/components/dims.py
```diff
@@ -353,28 +300,30 @@ def set_point(self, axis: int, value: Union[int, float]):
         ----------
         axis : int
             Dimension index.
-        value : int or float
+        point : int or float
```
jni: You've changed the name in the docstring but not in the signature itself.

napari/components/dims.py
jni: `set_current_step`?

jni: I also don't like that set_point now depends on this. The vision for set_point is that it should be a continuous value, and future applications might include for example setting the point and then using that point to pivot oblique cut planes. For that, you need the point *before* rounding. So it seems to me that this should be changed so that:

- `set_point` is the "canonical" API and keeps all the info (including fractional coordinates)
- the `step` property (maybe `current_step`?) grabs the point attribute and rounds it appropriately using `range`.
- `set_step` uses `range` to call `set_point`.

What do you think @sofroniewn? Is this something you tried before and it failed?

napari/layers/base/base.py
```diff
@@ -448,6 +440,26 @@ def _extent_world(self) -> np.ndarray:
         """
         return self._transforms['data2world'](self._extent_data)
 
+    @property
+    def _slice_indices(self):
+        """(D, ) array: Slice indices in data coordinates."""
+        world_pts = [self._dims_point[ax] for ax in self._dims.not_displayed]
+        inv_transform = self._transforms['data2world'].inverse
+        data_pts = inv_transform.set_slice(self._dims.not_displayed)(world_pts)
```
jni: This is pretty. =)

napari/layers/base/base.py
```diff
@@ -621,6 +633,64 @@ def set_view_slice(self):
...
+
+        if order is None:
+            order = list(range(ndim))
+
+        # adjust the order of the global dims based on the number of
+        # dimensions that a layer has - for example a global order of
+        # [2, 1, 0, 3] -> [0, 1] for a layer that only has two dimensions
+        # or -> [1, 0, 2] for a layer with three as that corresponds to
+        # the relative order of the last two and three dimensions
+        # respectively
```
jni, line 664: Thank you for this comment!!! 🙏

napari/layers/surface/_tests/test_surface.py
```diff
@@ -35,7 +34,7 @@ def test_random_3D_surface():
     assert layer._data_view.shape[1] == 2
     assert layer._view_vertex_values.ndim == 1
 
-    layer.dims.ndisplay = 3
+    layer._slice_dims(ndisplay=3)
```
jni, line 37: This kinda makes me question the choice of making dims private for now. It's very clunky. What's the long term vision here? Don't we want individual layers to be able to have different ndisplays?

jni (commented):
@sofroniewn Finally went through this. 😬 Main bits of feedback:

- I'm not sure about the changes to layer.dims. I kinda think they should be reverted. It'll make the PR smaller and afaict won't affect *any* of the functionality added, right?
- I don't think set_point should depend on set_step, but rather the other way around. There should always be a point with fractional values in world space.
- I'd like set_step to be renamed to something like set_step_number to differentiate it from e.g. set_step_size, which is what I implicitly thought it was for when I first saw it.

Other than the above, it's minor fixes noted. And thanks for the very extensive comments, I super appreciated them during review. =)

napari/components/dims.py
sofroniewn: > What do you think @sofroniewn? Is this something you tried before and it failed?

you are correct, this is where i started, for the reasoning you provided, but it created all sorts of weird behavior due to the round trip nature of setting the slider value and the slider value then corresponding to a point value. Some of that may be symptomatic of deeper issues that could get resolved, but it was much more awkward from an implementation stand point to do it that way.

napari/components/dims.py
sofroniewn: If I drop the concept of modes and interval (which isn't implemented anywhere) then those two lines do nothing and aren't needed. Point doesn't need to be set, as point is never stored, it is calculated from step

napari/layers/surface/_tests/test_surface.py
sofroniewn: long term vision is that layers shouldn't know about ndisplay as it relates to slicing, and we'll probably have some public slice method that is very clean, but ndisplay never gets stored on the layer itself and the sliced data never gets stored on the layer itself

sofroniewn:
> I'm not sure about the changes to layer.dims. I kinda think they should be reverted. It'll make the PR smaller and afaict won't affect any of the functionality added, right?

Do you mean swapping back to `layer.dims` instead of `layer._dims`? I could do that, but now a user shouldn't interact with the `layer` `dims` model at all and if they did unexpected or bad things might happen. We also definitely want to pull the dims model off the layer itself as part of our refactors in #1353, so that layers don't have any concept of slicing. Given then that making private seemed like a good idea.

jni:
> Given then that making private seemed like a good idea.

To me the diff "clutter" doesn't seem worth it. Since all these commits are going to get squashed, for future reference to the PR/commits, it is really useful to keep the changes minimal. I find myself going back to PRs very often to remember how something was implemented.

napari/components/dims.py
jni: Hmmm. Ok. I guess we can punt on this, but it makes me uncomfortable! All this might fall out nicely in #1353 anyway. Perhaps add a comment in both docstrings (set_step/set_point) that make it clear how each is implemented in terms of the other, vs the ideal case?

jni (approved):
@sofroniewn in light of your responses and to avoid blocking progress here, I'll approve, but as mentioned in my comments, yes, I do think it's worth reverting the `dims`->`_dims` changes before merging. They can and should come as part of #1353 . Thanks!

napari/components/dims.py
sofroniewn: reverted

napari/components/dims.py
sofroniewn: I've gone with `current_step` and added comments. I am leaving the `point` vs `current_step` as they are now, but agreed we might want to revisit. As noted above I wanted to do it that way originally

sofroniewn:
> @sofroniewn in light of your responses and to avoid blocking progress here, I'll approve, but as mentioned in my comments, yes, I do think it's worth reverting the dims->_dims changes before merging. They can and should come as part of #1353 . Thanks!

I have made `layer.dims` public again, agreed that change was not necessary as part of this PR and the diff is much cleaner without it (though still large!). I'd like to make an immediate followup PR that only makes the layer dims private as I think now is the right time to do it, but can continue discussion on that there!

I've also renamed `step` to `current_step` as that was confusing, but havn't changed the `step` vs `point` transformation, we can revisit that if the behaviour is counterintuitive / breaks.

I will merge after tests pass and then make the follow-up PR.

PR #1540, opened 2020-08-09
Filter all nan axis warning on layer extent

### LINKED ISSUE #1536: RuntimeWarning when creating empty points

sofroniewn:
# Description
Closes #1536 by filtering the warning - see comment in code about why this is ok. Thoughts @tlambert03 if you like this approach or would prefer another?

## Type of change

### DIFF
--- a/napari/components/layerlist.py
+++ b/napari/components/layerlist.py
@@ -1,6 +1,7 @@
 import numpy as np
 import itertools
 from typing import Optional, List
+import warnings
 from ..layers import Layer
 from ..utils.naming import inc_name_count
 from ..utils.list import ListModel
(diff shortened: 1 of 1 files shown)


tlambert03 (approved):
Sure, works for me!

sofroniewn:
Thanks for the quick review @tlambert03.

Looks like test fails on linux are conda related and not connected to this PR https://github.com/napari/napari/pull/1540/checks?check_run_id=964211252.

I'll let sit for a little in case anyone else wants to comment and will then merge

sofroniewn:
Got a zulip green light from @jni on this one. Re-running the failing linux tests to see if CI has fixed them, but they are independent of this PR. Will merge soon

sofroniewn:
CI still failing here and now on master, but I will merge this regardless, as it is independent

PR #1541, opened 2020-08-10
pin conda in CI tests to 4.8.3

tlambert03:
# Description
let's see if this lets us test again... there's very little need to be on the bleeding edge conda for our CI, and we can unpin later


tlambert03:
looks like it it's working ... what do you say @sofroniewn, shall we merge this so we don't have to worry about 3.6/3.7 for now?  can unpin in a week or two after the new conda version is settled

sofroniewn (approved):
> shall we merge this so we don't have to worry about 3.6/3.7 for now? can unpin in a week or two after the new conda version is settled

LGTM! plan sounds good @tlambert03

Issue #1527, opened 2020-08-06
Mac OSX 3.7 CI fail


sofroniewn:
## 🐛 Bug

We are getting a consistent Mac OSX 3.7 CI fail

See https://github.com/napari/napari/runs/951608936

looks like it might be pooch/ scikit-image related

```
Collecting monotonic>=0.1
  Using cached monotonic-1.5-py2.py3-none-any.whl (5.3 kB)
Requirement already satisfied: urllib3!=1.25.0,!=1.25.1,<1.26,>=1.21.1 in /Users/anka/conda/lib/python3.7/site-packages (from requests->pooch->-r requirements/test.txt (line 11)) (1.25.9)
Requirement already satisfied: idna<3,>=2.5 in /Users/anka/conda/lib/python3.7/site-packages (from requests->pooch->-r requirements/test.txt (line 11)) (2.10)
Requirement already satisfied: certifi>=2017.4.17 in /Users/anka/conda/lib/python3.7/site-packages (from requests->pooch->-r requirements/test.txt (line 11)) (2020.6.20)
Requirement already satisfied: chardet<4,>=3.0.2 in /Users/anka/conda/lib/python3.7/site-packages (from requests->pooch->-r requirements/test.txt (line 11)) (3.0.4)
ERROR: THESE PACKAGES DO NOT MATCH THE HASHES FROM THE REQUIREMENTS FILE. If you have updated the package versions, please update the hashes. Otherwise, examine the package contents carefully; someone may have tampered with them.
    scikit-image from https://files.pythonhosted.org/packages/9f/21/62a5e3b6b5a84d182c181d9787f9b0210d44ff2e2d6305a08a6664253266/scikit_image-0.17.2-cp37-cp37m-macosx_10_13_x86_64.whl#sha256=87ca5168c6fc36b7a298a1db2d185a8298f549854342020f282f747a4e4ddce9 (from -r requirements/test.txt (line 10)):
        Expected sha256 87ca5168c6fc36b7a298a1db2d185a8298f549854342020f282f747a4e4ddce9
             Got        2a227617e44a184943f657c0175ec7c279f596d4debc614f229b91fc310dfb7c
```

sofroniewn:
Looks like CI just passed, so it might not happen every time, or it might no longer be an issue .... https://github.com/napari/napari/runs/956001037

sofroniewn:
[...]

PR #1544, opened 2020-08-11
attempt to invalidate cache

liu-ziyang:
# Description
tries to invalidate cache without being merged

## Type of change

# References

# How has this been tested?
- example: all tests pass with my change

## Final checklist:

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -118,11 +118,11 @@ mac_task:
     populate_script:
       - export MACOSX_DEPLOYMENT_TARGET=10.10
       - python -m pip install --retries 3 --upgrade pip
-      - python -m pip install --retries 3 -r requirements/test.txt
+      - python -m pip install --retries 3 -r requirements/test.txt --no-cache-dir
 
   install_script:
     - python -m pip install --retries 3 --upgrade pip
-    - python -m pip install --retries 3 -r requirements/test.txt
+    - python -m pip install --retries 3 -r requirements/test.txt --no-cache-dir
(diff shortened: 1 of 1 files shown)


.cirrus.yml
```diff
@@ -118,11 +118,11 @@ mac_task:
     populate_script:
       - export MACOSX_DEPLOYMENT_TARGET=10.10
       - python -m pip install --retries 3 --upgrade pip
-      - python -m pip install --retries 3 -r requirements/test.txt
+      - python -m pip install --retries 3 -r requirements/test.txt --no-cache-dir
 
   install_script:
     - python -m pip install --retries 3 --upgrade pip
-    - python -m pip install --retries 3 -r requirements/test.txt
+    - python -m pip install --retries 3 -r requirements/test.txt --no-cache-dir
```
tlambert03, line 125: why use the `--no-cache-dir` option instead of just getting rid of the `pip_cache` step in the first place?  With this change, we're basically saying "cache it, but don't ever use the cache!" right? :)

also... is this a one-time and then switch-it-back sort of thing?  Or have you learned that we will no longer be able to take advantage of caching at all to speed up builds?

liu-ziyang: I was hoping this would clear the cache so we don't need to merge anything, but I am not very confident

liu-ziyang:
the default ci doesn't run on mac with pytbon 3.7! I opened up another empty PR to reset the cache, we should merge this empty PR and have the ci reset the cache: https://github.com/napari/napari/pull/1548

PR #1546, opened 2020-08-13
viewer.watch_path method to auto-concatenate data from a monitored folder

tlambert03:
# Description
Got inspired thinking about [this zulip thread](https://napari.zulipchat.com/#narrow/stream/212875-general/topic/live.20data), and wrote up a proposal for a `viewer.watch_path` method.  This PR would allow the following:

```python
import napari
from napari.utils.watch_path import regex_parser

with napari.gui_qt():
    viewer = napari.Viewer()
    channel_parser=regex_parser(r'.*ch(\d{1}).*'),
    viewer.watch_path('some/path', plugin=None, channel_parser=channel_parser)
```

and then everytime a new image appears in `'some/path'`, it will be appended to the corresponding channel (where channels are gleaned from the pathname using the `channel_parser`, a function that takes a string and returns a list of channel names).  The slider will also keep up, showing the most recent dataset. 

This works fine with multiple channels across multiple files, but it does assume that a given channel will have the same shape across all timepoints.  Probably a lot of edge cases to consider, but curious what you think of the API.

## Type of change

# References

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/components/add_layers_mixin.py
+++ b/napari/components/add_layers_mixin.py
@@ -3,7 +3,17 @@ import itertools
 import os
 from functools import lru_cache
 from logging import getLogger
-from typing import Any, Dict, List, Optional, Sequence, Set, Union
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    List,
+    Optional,
+    Sequence,
+    Set,
+    Tuple,
+    Union,
+)
 
 import numpy as np
 
@@ -14,6 +24,7 @@ from ..plugins.io import read_data_with_plugins
 from ..types import FullLayerData, LayerData
 from ..utils.colormaps import ensure_colormap_tuple
 from ..utils.misc import is_sequence
+from ..utils.watch_path import watch_path
 
 logger = getLogger(__name__)
 
@@ -792,6 +803,8 @@ class AddLayersMixin:
         stack: bool = False,
         plugin: Optional[str] = None,
         layer_type: Optional[str] = None,
+        watch: bool = False,
+        channel_parser: Callable[[str], Tuple[str, ...]] = None,
         **kwargs,
     ) -> List[layers.Layer]:
         """Open a path or list of paths with plugins, and add layers to viewer.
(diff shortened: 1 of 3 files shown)


jni:
Some additional thoughts on this:

- at the group meeting, we settled on `viewer.open(..., watch=True)`. Just noting that down here. ;)
- We could potentially supplement this with a "File > Open Live..." menu. Or it would go in the yet-to-be-implemented Open Advanced... dialog.
- the pinning to the last timepoint should only happen *if the current timepoint is the last timepoint*. ie I want to be able to browse around without napari constantly dragging me back to the end. This is how e.g. most terminals work — if something is printing to the terminal, if I scroll up my current position is preserved, but if I scroll all the way to the bottom the printing updates my screen.
- Finally, I was thinking about how this could be an IO plugin, and here's a half-baked proposal:

```python
def napari_get_live_reader(path, **kwargs) -> Generator[LayerData]:
```

ie the plugins would only need to define something that (indefinitely — though perhaps we could add a timeout kwarg) watches a path and yields new layerdata tuples every time the data at that path is updated, and we would take care to (a) create the layer on the first yield, and (b) update the layer on subsequent yields.

Thoughts?

tlambert03:
> * at the group meeting, we settled on `viewer.open(..., watch=True)`. Just noting that down here. ;)

already changed in this PR :)

> * We could potentially supplement this with a "File > Open Live..." menu. Or it would go in the yet-to-be-implemented Open Advanced... dialog.

👍

> * the pinning to the last timepoint should only happen _if the current timepoint is the last timepoint_. 

💯

> * Finally, I was thinking about how this could be an IO plugin, and here's a half-baked proposal:
> 
> ```python
> def napari_get_live_reader(path, **kwargs) -> Generator[LayerData]:
> ```
> 
> ie the plugins would only need to define something that (indefinitely — though perhaps we could add a timeout kwarg) watches a path and yields new layerdata tuples every time the data at that path is updated, and we would take care to (a) create the layer on the first yield, and (b) update the layer on subsequent yields.

jury is still out for me on this one.  I mostly like the hookspec, it's pretty much like the [`path_reader`](https://github.com/napari/napari/blob/9c40c0f4b2e81d9a9a864fe4b5e34a4306fb00f9/napari/utils/watch_path.py#L31) in this PR... with the main difference that the function itself doesn't take arguments, but it yields a generator that expects string inputs using `.send`, which is hinted like this:
```python
# accepts an optional string, and yields lists of layerdata
def path_reader() -> Generator[List[FullLayerData], Optional[str], None]: ...
```
we can then wrap that for them as a threadworker and let 'er rip, passing the generator new files as they appear.  However, if you  look at [my implementation of that generator](https://github.com/napari/napari/blob/9c40c0f4b2e81d9a9a864fe4b5e34a4306fb00f9/napari/utils/watch_path.py#L31), it's _already_ calling plugins behind the scenes to do the actual reading ... so I'm kinda wondering whether this can be achieved more as a service to plugins, and less as a hook-spec itself.

The main challenge here is metadata (and perhaps that's what you're already intuiting).  We need the user to be able to indicate what additional dimension(s) are contained in the "file axis".

I have some more (local) code where I implemented this is a zarr store, but remaining challenges are still how to let the user indicate axes (e.g. metadata), and how best to update the shape of a zarr-backed dask array.

Issue #1307, opened 2020-05-28
Test on github actions?


tlambert03:
## 🐙😸 Test on github actions?
This definitely falls under the category of "low priority" and "if it ain't broke..."
but I just wanted to show an example of our full test suite [running on github actions](https://github.com/tlambert03/napari/actions/runs/118097110) (branch diff [here](https://github.com/napari/napari/compare/master...tlambert03:feature/github-tests))

this is testing linux36,37,38 with pyqt/pyside, mac38 with pyqt/pyside, and windows38 with pyside only (vm failed with pyqt as it does on cirrus)
 
I'm not making a PR at the moment, because I also took the opportunity to learn/try other things there, including:
- using `tox.ini` to define environments
   - i like! ... note, you can `pip install tox-conda` to use with conda envs locally... and it still doesn't prevent you from running `pytest` in your current environment... but does make it easier to define broad dependency matrices (like python version, qt, etc...)
- move the vast majority of our `setup.py` stuff into `setup.cfg` (like #1228)

but it's here for people to evaluate in case we ever want to move in that direction...

## Motivation

just always reconsidering best practices. sometimes cirrus throws strange errors (but is mostly reliable!).  And there's a growing community of examples and actions for github

## Alternatives

just stick with what we've got!

fkorotkov:
@tlambert03 what kind of strange errors do you see on Cirrus?

tlambert03:
eek.  you mean I have to defend my bold & baseless claims 😅!??
well, as mentioned, it's mostly been very reliable (and the support is good!), but here are a few of the runs we've had recently where I don't understand what's going wrong, (also absolutely acknowledging that this may be our own issues!)

recently been seeing an uptick in connectivity problems like this one:
https://cirrus-ci.com/task/5756071266484224

and an uptick in tests that seem to hang which we force restart after ~30-40 mins (and then they pass) 
https://cirrus-ci.com/task/5080628334952448
https://cirrus-ci.com/task/6072148848541696

anyway,  definitely not trying to throw anyone under the bus... mostly just wanting to learn all of the available CI options out there.

fkorotkov:
Hm. Interesting. Not suer about the SSL errors but will take a look into the hanging cache uploads. 👌

tlambert03:
thanks as always!

sofroniewn:
[...]

PR #1548, opened 2020-08-14
Ci cache reset

liu-ziyang:
# Description
Reset cache on mac 3.7

## Type of change
Nothing changed

# References

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -118,7 +118,7 @@ mac_task:
     populate_script:
       - export MACOSX_DEPLOYMENT_TARGET=10.10
       - python -m pip install --retries 3 --upgrade pip
-      - python -m pip install --retries 3 -r requirements/test.txt
+      - python -m pip install --retries 4 -r requirements/test.txt
 
   install_script:
     - python -m pip install --retries 3 --upgrade pip


liu-ziyang:
I tested this on this branch to run on mac with python 3.7 (this change has been rolled back reset --hard) 
When I tested it, CI was indeed rerunning pip install instead of using a existing matching, however I cannot guarantee this happens too on master, because branches was not running 3.7 before
Worst case scenario we would have to merge the 2 commits in this PR separately, which themselves individually are pretty harmless and cancels out each other anyway

liu-ziyang:
supposedly, we just need ci to think the build process is different, it looks like it uses a way to identify (during the PIP cache step it says it matches a hash) so ones we modify anything, the cache would be invalidated, so we are going to attempt change a harmless config flag (retry number) and trick ci to run the new flow and invalidate the existing cache, and maybe revert it if desired later

sofroniewn (approved):
> supposedly, we just need ci to think the build process is different, it looks like it uses a way to identify (during the PIP cache step it says it matches a hash) so ones we modify anything, the cache would be invalidated, so we are going to attempt change a harmless config flag (retry number) and trick ci to run the new flow and invalidate the existing cache, and maybe revert it if desired later

Thanks for that explanation @ziyangczi - this does indeed seem harmless and if this does the trick that's great!! We can then revert in a follow up PR.

I'll merge shortly unless @tlambert03 has additional ideas. I did look into if the cache could be reverted in the CI UI as @jni asked, but I couldn't see a way.

I wonder if @ziyangczi should look into the github actions proposal from @tlambert03 in #1307 as a long term approach.

tlambert03:
> I wonder if @ziyangczi should look into the github actions proposal from @tlambert03 in #1307 as a long term approach.

fwiw, [that branch](https://github.com/napari/napari/compare/master...tlambert03:feature/github-tests) is still mostly in good shape.  Just updated and tests are mostly passing: https://github.com/tlambert03/napari/runs/986291638?check_suite_focus=true

needs a bit of love on windows, and needs isort added to CI.  (and a discussion as to whether people are ok with using tox as I've done there, or prefer not to)

sofroniewn:
Unfortunately @ziyangczi this approach has not succeeded - see 
image

What do you recommend @ziyangczi we try next?

liu-ziyang:
I took a look at 1307, I think we should move forward with the migration, I can take over the branch if desired. I can help with the windows test part.

regarding tox, We can talk about it in the next meeting, I don't think it is a deal-breaker, I personally would hope we do not use a tool to facilitate testing on different environments because it introduced some complexity/dependency in ci, and can be tricky later if breaks, etc

liu-ziyang:
regarding the cirrus ci fix, I think the most viable fix would be updating some dependency in our pip file to force a new dependency installation process. I can't find any doc from cirrus explaining how they match the pip cache so we can find a way to work around it or any tutorial to do so manually.

tlambert03:
> because it introduced some complexity/dependency in ci

would be curious to know what difficulties you ran into... I actually added it there because, combined with [tox-gh-actions](https://github.com/ymyzk/tox-gh-actions) I found that it actually simplified testing on CI with a variety of environments (and as a bonus, makes it much easier to test a matrix of environments locally, such as testing both pyside/pyqt on py3.6-3.x just by running `tox`)

liu-ziyang:
> > because it introduced some complexity/dependency in ci
> 
> would be curious to know what difficulties you ran into... I actually added it there because, combined with [tox-gh-actions](https://github.com/ymyzk/tox-gh-actions) I found that it actually simplified testing on CI with a variety of environments (and as a bonus, makes it much easier to test a matrix of environments locally, such as testing both pyside/pyqt on py3.6-3.x just by running `tox`)

not any difficulties immediately, but in my opinion, that is too much magic for the CI setup. I would argue CI should be the pure dumb code setup. If we need to test out on a matrix of setups, I would rather just flush out the code completely instead of trying to be smart and optimize it. I have had enough nightmares dealing with "smart" CIs, the pip cache is certainly the most recent one :) 
That being said, though I would not be setting up such a thing in my own project, I don't think this is a deal-breaker that we need to bang out head in arguing whether to have it or not, we can improve our CI setup later anyway since this is nothing relevant to our "business model", I am more concerned about what is the difference in cost than the tox. Hoping github is nice enough to allow us using it for free?

PR #1550, opened 2020-08-14
Fix perfmon for qt changes, add counters, cleanup

pwinston:
# Description
* Fix perfmon from Qt re-org,.
* Added new perfmon `add_counter_event()` API.
* Lots of docstring cleanup and type fixes.
* Fix `PerfTraceFile`to only write at the end:
    * We used to write _as we went_, and some stuff was legacy from that era.
* Add one unit test: `test_perf.py`
   * Add pytest flag `--perfmon-only`
   * Add pytest mark `@pytest.mark.perfmon`
   * Add new Cirrus.yml test pass for this one perfmon test

We need a seperate test pass so we can define `NAPARI_PERFTEST` which is only checked once on module import.

## Type of change
- [x] Bug file (Qt issue)
- [x] New Features: `add_counter_event()`
- [x] New unit test

# How has this been tested?
- [x] Manual testing
- [x] The one new unit test

## Final checklist:

# Example of add_counter_event()

A counter displays as a bar graph or stacked bar graph (if multiple values). For instance here it's showing the length of the delay queue. Basically it's visibility into what scalar values are doing over time.

perfmon-counters


napari/utils/perf/_event.py
```diff
@@ -2,7 +2,14 @@
 """
 import os
 import threading
-from typing import Optional
+from collections import namedtuple
+from typing import Dict, Optional
+
+# A span of time for an event.
```
sofroniewn: I really like this use of namedtuple, blog your post here https://tobeva.com/articles/brain-compatible-code/ really resonated with me!!

sofroniewn (approved):
This looks good to me, thanks for the small digestible PR. Really nice to have this good tooling in for the perfmon!

jni (commented):
@pwinston thanks for the fix! My main Q is: could we add a test that checks that this doesn't break? Otherwise it's bound to happen again and again!

pwinston:
I'm not sure about this CI error @jni. When calculating *coverage* it says that `make_test_viewer()` leads to `import sip` failing. As if pyside2 is not installed? I don't really following what our cirrus.yml is doing with the switching from pyside2 to pyqt. Are the steps robust to adding a 2nd permutation to the matrix?

Calculating coverage with `--perfmon-only` is pretty pointless. So we could try to bypass that. But I don't see why it should fail.

pwinston:
I disabled the code coverage for the perfmon pass. There is no point, it's just one test. Although I'm not sure why it didn't successfully run, something with pyside/pyqt maybe.

So the **new perfmon integration test** is in `napari/utils/perf/_test/test_perf.py`. Writes a trace JSON and makes sure it’s not empty and the entries have the right fields. Unit tests would still be nice but this is a big improvement.

See `test_perf.py`, `.cirrus.yml`, `conftest.py`. Also see the fix to `qt_main_window.close()`:
https://github.com/napari/napari/pull/1550/commits/2a35e2bdb600cebd59efc4e58c402b629787b22d

I think it's ready to merge, but maybe @jni and @kevinyamauchi can take a look.

napari/conftest.py
```diff
@@ -37,15 +37,22 @@ def fetch(data_name):
...
 def pytest_addoption(parser):
-    """An option to show viewers during tests. (Hidden by default).
+    """Add napari specific command line options.
 
-    Showing viewers decreases test speed by about %18.  Note, due to the
-    placement of this conftest.py file, you must specify the napari folder (in
-    the pytest command) to use this flag.
+    --show-viewer
+        Show viewers during tests, they are hidden by default. Showing viewers
+        decreases test speed by about %18.
```
jni:
```suggestion
        decreases test speed by about 18%.
```
? =) Also, maybe we can just round it to 20%? "decreases test speed by approximately 18.42927%" 😂

napari/conftest.py
```diff
@@ -37,15 +37,22 @@ def fetch(data_name):
...
-    Examples
-    --------
-    $ pytest napari --show-viewer
+    --perfmon-only
+        Run only perfmon test.
+
+    Notes
+    -----
+    Due to the placement of this conftest.py file, you must specifically name
+    the napair folder such as "pytest napari --show-viewer"
```
jni:
```suggestion
    the napari folder such as "pytest napari --show-viewer"
```

napari/utils/perf/_event.py
```diff
@@ -16,34 +23,41 @@ class PerfEvent:
...
+    category :str
+        Comma separated categories such has "render,update".
     process_id : int
         The process id that produced the event.
     thread_id : int
         The thread id that produced the event.
-    category :str
-        Comma separated categories such has "render,update".
+    phase : str
+        The Chrome Tracing "phase" such as "X", "I", "C".
```
jni, line 33: It seems you've deleted the little table of meanings for these — worth bringing it back! Specifically this description was very useful:

     phase : str
         The chrome://tracing "phase" (event type). The spec defines
         around 20 phases we only support two right now:
              "X" - Complete Events
              "I" - Instant Events

napari/utils/perf/_event.py
```diff
@@ -54,39 +68,76 @@ def __init__(
...
+    time_ns : int
+        The time in nanoseconds when the event happened.
+    kwargs : dict
+        Arguments to display in the Args section of the GUI.
+    """
+
     def __init__(self, name: str, time_ns: int, **kwargs):
-        super().__init__(name, time_ns, time_ns, **kwargs)
-        self.phase = "I"  # instant event
+        # Instant events are phase "I" in their trace space.
```
jni: Is this supposed to say "their trace space" or "the trace spec"?

napari/utils/perf/_timers.py
```diff
@@ -83,16 +74,34 @@ def add_event(self, event: PerfEvent):
...
+        name : str
+            The name of this event like "draw".
+        kwargs : Dict[str, float]
+            The individual counters for this event.
+
+        Notes
+        -----
+        For example add_counter_event("draw", triangles=5, squares=10).
+        """
+        self.add_event(CounterEvent(name, perf_counter_ns(), **kwargs))
```
jni: It seems to me that these classes are a rather heavyweight way to express this code — would this not be equally simple?

```suggestion
        self.add_event(PerfEvent(name, perf_counter_ns(), phase='C', **kwargs))
```

and so on for each supported phase?

napari/utils/perf/_timers.py
```diff
@@ -159,17 +197,18 @@ def perf_timer(
...
+    def add_counter_event(name: str, **kwargs: Dict[str, float]) -> None:
         pass
 
     # contextlib.nullcontext does not work with kwargs, so we just
-    # create a do-nothing context object. This is not zero overhead
-    # but it's very low, about 1 microsecond? But because it's not
-    # zero it's best practice not to commit perf_timers, think of
-    # them like debug prints.
+    # create a do-nothing context object. Not zero overhead but
+    # very small.
```
jni, line 213: I think the previous message was more informative and useful, is there a reason you removed it?

jni (approved):
Thanks @pwinston! I've left some minor comments — mostly around comments! Please fix the documentation bits before merging, where I think there were a couple of regressions. I'd also love the extra class definitions removed but that can happen later.

napari/utils/perf/__init__.py
```diff
@@ -1,41 +1,42 @@
-"""Performance Monitoring init.
+"""Performance Monitoring.
 
-USE_PERFMON is true only if NAPARI_PERFMON environment variable is set and not
-zero.
+To enable perfmon define the env var NAPARI_PERFMON to be non-zero.
```
kevinyamauchi: I think it would be nice to add a bit more detail on what is meant by "non-zero". Perhaps something like:

```suggestion
To enable perfmon define the env var NAPARI_PERFMON to be non-zero. For example:

- set to 1 to activate perfmon and use the menu Debug -> Performance Trace in the napari GUI
- set to the path to your perfmon config file to configure napari to start up with perfmon set up as specified in the config file. See below for more information of the perfmon config file.
```

kevinyamauchi (approved):
This looks good to me! Stoked to have the qt tracing back. I left just have a minor comment below. I assume we will eventually have a docs page for this, but it would be nice to have a bit more detail on how to activate/configure perfmon in the meantime.

napari/conftest.py
pwinston: That was not my comment, Talley's I think, but I changed it to say "by about 20%".

napari/conftest.py
pwinston: Fixed

napari/utils/perf/_event.py
pwinston: Added back the table.

napari/utils/perf/_event.py
pwinston: "Spec" but I just removed that line. I think it's redundant with the above table added back.

napari/utils/perf/_timers.py
pwinston: I was going to say I like `CounterEvent` on that line and not just `C`. But yes the sub-classes were not at all pulling their weight, so I got rid of them. One thing they were doing is Instant and Counter events had only one time, the time it happened, and not both a start and an end time.. But not really worth it just for that. I just pass "now" in as the start and end now.

napari/utils/perf/_timers.py
pwinston: I think "think of them as debug prints" is a higher level piece of advice, that's not part of describing what the context object actually does. This higher level advice is now part of the overall perfmon doc string in `utils.perf.__init__`.

napari/utils/perf/__init__.py
pwinston: I thought that was covered somewhere, buy couldn't find where. I redid that whole comment block just now and included your ideas. Seems better now.

pwinston:
Okay everything has been addressed I think.

sofroniewn:
Thanks @pwinston comments look like they have all been addressed, and I just looked over it all once again, looks good to merge, I will do so now!!

napari/utils/perf/__init__.py
kevinyamauchi: Awesome! Thanks!

PR #1551, opened 2020-08-14
Revert "Ci cache reset"

liu-ziyang:
Reverts napari/napari#1548


sofroniewn (approved):
Thanks @ziyangczi - given that the attempted fix in #1548 to trigger clearing of the cache didn't work I will merge this PR to revert. (Note that we would have reverted the PR regardless of if it worked or not). It was definitely worth the try though!!

PR #1553, opened 2020-08-16
add fingerprint script to pip caching

liu-ziyang:
# Description
Enforce fingerprint of pip caching, this is pretty much last resort to force cache update, I was reluctant to go this path because it could potentially lead to a refresh of all caches, but it is a more appropriate way of setting up ci to recognize caching opportunity. (According to cirrus ci doc: ```By default the task name is used as a fingerprint value.```)

## Type of change

# References
https://cirrus-ci.org/guide/writing-tasks/#cache-instruction

# How has this been tested?

## Final checklist:

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -62,7 +62,7 @@ linux_task:
     folder: ~/.cache/pip
     fingerprint_script:
       - echo $PYTHON_VERSION
-      - type requirements/test.txt
+      - cat requirements/test.txt
     populate_script: 
       - python -m pip install --retries 3 --upgrade pip
       - python -m pip install --retries 3 -r requirements/test.txt
@@ -120,7 +120,7 @@ mac_task:
     folder: $HOME/Library/Caches/pip
     fingerprint_script:
       - echo $PYTHON_VERSION
-      - type requirements/test.txt
+      - cat requirements/test.txt
     populate_script:
       - export MACOSX_DEPLOYMENT_TARGET=10.10
       - python -m pip install --retries 3 --upgrade pip
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -187,7 +187,7 @@ win_task:
     folder: $LOCALAPPDATA\pip\Cache
     fingerprint_script:
       - echo $PYTHON_VERSION
-      - type requirements/test.txt
+      - type requirements\test.txt
     populate_script: 
       - python -m pip install --retries 3 --upgrade pip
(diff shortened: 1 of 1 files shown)


tlambert03:
cc @fkorotkov

@ziyangczi, @fkorotkov has been very helpful in the past with cirrus stuff...  if you have any specific how-to questions, I'm sure he can answer them

fkorotkov (approved):

fkorotkov:
On the second glance it seems `requirements/test.txt` doesn't contain specific version of the dependencies which means you'll only invalidate the cache if a dependency will be added or removed. 🤔

tlambert03:
Thanks @fkorotkov!  The issue we're trying to solve is here: https://cirrus-ci.com/task/5094878520541184?command=install#L65

we can certainly pin some of our test dependencies to force cache expiration... but looking at that error, do you think the issue is that the cirrus cache is problematic and just needs to be refreshed?  or that we actually have some issues in our requirements files that need fixing?

fkorotkov:
@tlambert03 this might be because you do `pip install` once again even for a cache hit. 🤔 I wonder if you need the extra `pip install` at all in case you already have it in `populate_script`. 🤷‍♂️

tlambert03:
That seems reasonable.  I think there was confusion as to whether the pip_cache was caching downloaded (but uninstalled) pip wheels (in which case we still needed to install after a cache hit), or whether it was actually caching the fully installed environment (in which case we just move on to testing). The assumption was the former

liu-ziyang:
@sofroniewn shall we merge this one? I hope this will address it once and for all
we may want to update test dependency file to include version later if that becomes an issue too

sofroniewn (approved):
> @sofroniewn shall we merge this one? I hope this will address it once and for all
we may want to update test dependency file to include version later if that becomes an issue too

Yes, happy to approve and then merge. @ziyangczi will these changes be reverted after the problem is fixed, like the last approach, or are they meant to be permanent?

liu-ziyang:
> > @sofroniewn shall we merge this one? I hope this will address it once and for all
> > we may want to update test dependency file to include version later if that becomes an issue too
> 
> Yes, happy to approve and then merge. @ziyangczi will these changes be reverted after the problem is fixed, like the last approach, or are they meant to be permanent?

these ones meant to be permanent if it fixed the problem, fingers crossed

sofroniewn:
It passed!! 
image

thanks for preserving here @ziyangczi!!

PR #1554, opened 2020-08-16
fix menubar raise on mac by running with framework python

sofroniewn:
# Description
Following the incredibly helpful and detailed example of @cbrnr https://github.com/napari/napari/issues/380#issuecomment-659656775 I have now fixed the menubar issue on mac when launching napari from the CLI with `napari`, see #380.

Note that the issue is not fixed when running `python examples/add_image.py` as the needed code doesn't get called.

I'm not an expert in this stuff and we might want to make some adjustments to the code that @cbrnr to meet all of napari's use cases, but it's pretty huge to get this working!!

Feedback, review, and testing from @ziyangczi @tlambert03 @jni welcome and thanks again to @cbrnr - truly incredible to have this bug fixed.

## Type of change

### DIFF
--- a/napari/__main__.py
+++ b/napari/__main__.py
@@ -195,8 +195,8 @@ def _run():
 def _run_pythonw(python_path):
     """Execute this script again through pythonw.
 
-    This can be used to ensures we're using a framework
-    build of Python on macOS which fixes menubar issues.
+    This can be used to ensure we're using a framework
+    build of Python on macOS, which fixes frozen menubar issues.
 
     Parameters
     ----------
--- a/napari/__main__.py
+++ b/napari/__main__.py
@@ -10,6 +10,7 @@ import sys
 import warnings
 from ast import literal_eval
 from distutils.version import StrictVersion
+from pathlib import Path
 from textwrap import wrap
 from typing import Any, Dict, List
 
@@ -191,31 +192,20 @@ def _run():
             )
 
 
-def _run_pythonw():
+def _run_pythonw(python_path):
     """Execute this script again through pythonw.
 
-    This ensures we're using a framework build of Python on macOS.
+    This can be used to ensures we're using a framework
+    build of Python on macOS which fixes menubar issues.
+
+    Parameters
+    ----------
+    python_path : pathlib.Path
+        Path to python framework build.
     """
-    import pathlib
     import subprocess
 
-    cwd = pathlib.Path.cwd()
-    python_path = pathlib.Path(sys.exec_prefix) / 'bin' / 'pythonw'
-
-    if not python_path.exists():
-        msg = (
-            'pythonw executable not found. '
-            'The menubar might only become '
-            'functional on the macOS after '
-            'focus is toggled in and out of '
-            'napari. To fix this problem, '
-            'please install python.app via conda, '
-            'which can be done with '
-            '`conda install -c conda-forge python.app`.'
-        )
-        raise warnings.warn(msg)
-        _run()
-
+    cwd = Path.cwd()
     cmd = [python_path, '-m', 'napari']
     env = os.environ.copy()
 
@@ -241,9 +231,25 @@ def main():
     _RUNNING_PYTHONW = "PYTHONEXECUTABLE" in os.environ
 
     if _MACOS_LATEST and _RUNNING_CONDA and not _RUNNING_PYTHONW:
-        _run_pythonw()
-    else:
-        _run()
+        python_path = Path(sys.exec_prefix) / 'bin' / 'pythonw'
+
+        if python_path.exists():
+            # Running again with pythonw will exit this script
+            # and use the framework build of python.
+            _run_pythonw(python_path)
+        else:
+            msg = (
+                'pythonw executable not found. '
+                'The menubar might only become '
+                'functional on the macOS after '
+                'focus is toggled in and out of '
+                'napari. To fix this problem, '
+                'please install python.app via conda, '
+                'which can be done with '
+                '`conda install -c conda-forge python.app`.'
+            )
+            raise warnings.warn(msg)
+    _run()
 
 
 if __name__ == '__main__':
--- a/napari/__main__.py
+++ b/napari/__main__.py
@@ -4,9 +4,12 @@ napari command line viewer.
 import argparse
 import logging
 import os
+import platform
 import runpy
 import sys
+import warnings
 from ast import literal_eval
+from distutils.version import StrictVersion
 from textwrap import wrap
 from typing import Any, Dict, List
 
@@ -190,6 +193,7 @@ def _run():
 
 def _run_pythonw():
     """Execute this script again through pythonw.
+
     This ensures we're using a framework build of Python on macOS.
     """
     import pathlib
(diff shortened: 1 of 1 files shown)


tlambert03 (approved):
this is awesome @sofroniewn!  I happy to merge this as is, but here are few thoughts:

1) while the majority of our users will probably be getting their framework builds from conda, it's not a strict requirement here, so it'd be nice if we had a way to detect that naturally without assuming/forcing conda
2) if someone used `pythonw -m napari`, it would work even without this PR, so we should probably let that case pass with intervening?  Assuming they are running `pythonw` installed by `python.app`, then the environment variable `PYTHONEXECUTABLE` will be set.  So we might want to look for that as well (commented inline)
3) as far as I know, this was only required for catalina. we _could_ consider only redirecting if if someone is using 10.15 or later (parsed from `platform.platform()`), that would at least mean that some people don't have to install `python.app` if they don't need or want to.
4) people _not_ running in a conda environment (like the bundled app), will not benefit from this right?  for the bundle, we'll need to look into how briefcase actually packages python, and whether they provide a framework build...  I don't think we need to do anything about that for now, but perhaps just remember in case anyone still reports this issue after this merges

napari/__main__.py
```diff
@@ -187,5 +188,46 @@ def main():
...
+    env["NAPARI_RUNNING_PYTHONW"] = "True"
+
+    subprocess.run(cmd, env=env, cwd=cwd)
+    sys.exit()
+
+
+def main():
+    # Ensure we're always using a "framework build" on macOS.
+    _MACOS_CONDA = sys.platform == "darwin" and "CONDA_PREFIX" in os.environ
+    _RUNNING_PYTHONW = "NAPARI_RUNNING_PYTHONW" in os.environ
```
tlambert03: or `PYTHONEXECUTABLE in os.environ` ... in fact, since we're basically requiring this to be `python.app` from conda, we could use that instead of `NAPARI_RUNNING_PYTHONW` if wanted

napari/__main__.py
```diff
@@ -187,5 +188,46 @@ def main():
...
+    import pathlib
+    import subprocess
+
+    cwd = pathlib.Path.cwd()
+    python_path = pathlib.Path(sys.exec_prefix) / 'bin' / 'pythonw'
+
+    if not python_path.exists():
+        msg = (
+            'pythonw executable not found. '
+            'Please install python.app via conda.'
```
tlambert03: i wonder if we want to give the exact instructions: `conda install -c conda-forge python.app`
some people might not notice the ".app" and think that they already have it installed

tlambert03: also... do we actually want to raise an exception and completely prevent napari from running here just because the menubar won't be active unless they refocus?  Maybe a warning instead?

tlambert03:
lastly... maybe we should add `python.app` to our conda recipe so this just works for most people who `conda install napari`

sofroniewn:
These are great comments @tlambert03, I'm open to all the suggestions (either in this PR or future ones). I'll let @jni respond first and then we can take it from there!

liu-ziyang:
similar to other comments, I can't say I am a fan of these env checks, they are too specific, and hardly future proof in any way. I am fine with this PR in general since it would address this particular issue, but I would think this is yet another straw on the camel of cross-platform compatibility from qt side

napari/__main__.py
```diff
@@ -187,5 +188,46 @@ def main():
...
+
+    # Append command line arguments.
+    if len(sys.argv) > 1:
+        cmd.append(*sys.argv[1:])
+
+    env = os.environ.copy()
+    env["NAPARI_RUNNING_PYTHONW"] = "True"
+
+    subprocess.run(cmd, env=env, cwd=cwd)
+    sys.exit()
```
liu-ziyang: does this exit hide any exit code? the subprocess is run from the previous line, should we inherit the exit code from there?

tlambert03:
> yet another straw on the camel of cross-platform compatibility from qt side

:joy: boy you really hate qt huh?  Do you have suggestions for alternatives?  just want to point out here that the major alternative to python gui (wxpython) won't even _launch_ if you don't use the framework build (`pythonw`) ... so it's really hard to blame this one on Qt.  This is more about accessing the display from python on a mac.

napari/__main__.py
tlambert03: good point.  I think you can do:
```python
    result = subprocess.run(cmd, env=env, cwd=cwd)
    sys.exit(result.returncode)
```
https://docs.python.org/3/library/subprocess.html#subprocess.CompletedProcess

liu-ziyang:
> > yet another straw on the camel of cross-platform compatibility from qt side
> 
> 😂 boy you really hate qt huh? Do you have suggestions for alternatives? just want to point out here that the major alternative to python gui (wxpython) won't even _launch_ if you don't use the framework build (`pythonw`) ... so it's really hard to blame this one on Qt. This is more about accessing the display from python on a mac.

(Not going to hide I am anti-qt). well, taking this problem from another angle is that qt didn't do the promised job of having the behavior being consistent across platforms. Yes, by all means, macos has been hard to work with in terms of GUI, pretty much any GUI from what I heard, that you have to build very specifically for macos otherwise it is just not happy. Even then I don't think it is justified to leave these heavy liftings to software developers. I would then argue throwing front end framework to javascript based election etc is a better choice in 2020 if we are truly committed to cross-platform compatibility, or build a specific frontend for each platform.

jni:
@ziyangczi to reiterate @tlambert03's point, this is not a Qt thing but rather a conda/Anaconda thing, see e.g. https://github.com/ContinuumIO/anaconda-issues/issues/199. Electron not being in Python means we would swap these problems for a whole other set of problems in coordinating js and py. Yes it's 2020 and Apple hates open source and nvidia and now Intel, and all these things make our lives hard, but I don't think this is a productive line of enquiry — and I say this as someone who also hates Qt, though probably for different reasons! (And in many cases loves Apple, though they are becoming harder and harder to love.)

@sofroniewn upon a quick glance, I agree with everything said so far (fixing sys.exit, trying to work with existing env vars, warn rather than fail). I have a sneaking suspicion that we could do a bit better by having a similar pattern within `napari.gui_qt` rather than `main` (or maybe in addition to `main`). Sorry I don't have time to think about it more deeply right now, will try to prioritise soon, agree it's a super important fix!

One thing I would say is could you add links in comments to all the relevant issues (incl Anaconda above), demo PRs as linked by @cbrnr, etc? I can definitely imagine coming over to this code in a year and thinking, wth are we doing here. 😂

sofroniewn:
Thanks for feedback all, I'll look into addressing concerns promptly

napari/__main__.py
sofroniewn: done

napari/__main__.py
sofroniewn: done

napari/__main__.py
sofroniewn: done

sofroniewn:
>  I agree with everything said so far (fixing sys.exit, trying to work with existing env vars, warn rather than fail).

I have done all of these things as requested and the other requests.

> I have a sneaking suspicion that we could do a bit better by having a similar pattern within napari.gui_qt rather than main (or maybe in addition to main).

I have not looked into this, but I think it is ok - you can run scripts with `pythonw exampes/add_image.py` and it works. Probably more important right now to put that in the documentation somewhere clearly rather than try and fix in `gui_qt`.

> lastly... maybe we should add python.app to our conda recipe so this just works for most people who conda install napari

@jni can you do this? this isn't done in this repo is it?

@jni I have made requested changes, everything works as expected for me, can you give one final review. Agreed this isn't ideal, but it's definitely a step forward

napari/__main__.py
```diff
@@ -187,5 +192,65 @@ def main():
             )
 
 
+def _run_pythonw(python_path):
+    """Execute this script again through pythonw.
+
+    This can be used to ensures we're using a framework
+    build of Python on macOS which fixes menubar issues.
```
jni:
```suggestion
    This can be used to ensure we're using a framework
    build of Python on macOS, which fixes frozen menubar issues.
```

napari/__main__.py
```diff
@@ -187,5 +192,65 @@ def main():
...
+                'pythonw executable not found. '
+                'The menubar might only become '
+                'functional on the macOS after '
+                'focus is toggled in and out of '
+                'napari. To fix this problem, '
+                'please install python.app via conda, '
+                'which can be done with '
+                '`conda install -c conda-forge python.app`.'
+            )
+            raise warnings.warn(msg)
```
jni:
```suggestion
            warnings.warn(msg)
```

Oops! This should not be a raise! Here's what happens when I run `python -m napari` without python.app installed:

```
 ~/projects/napari (fix_mac_menubar)
 $ python -m napari
/Users/jni/projects/napari/napari/__main__.py:251: UserWarning: pythonw executable not found. The menubar might only become functional on the macOS after focus is toggled in and out of napari. To fix this problem, please install python.app via conda, which can be done with `conda install -c conda-forge python.app`.
  raise warnings.warn(msg)
  File "/Users/jni/conda/envs/all/lib/python3.8/runpy.py", line 193, in _run_module_as_main
  File "/Users/jni/conda/envs/all/lib/python3.8/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/Users/jni/projects/napari/napari/__main__.py", line 256, in <module>
    sys.exit(main())
  File "/Users/jni/projects/napari/napari/__main__.py", line 251, in main
    raise warnings.warn(msg)
TypeError: exceptions must derive from BaseException
```

napari/__main__.py
```diff
@@ -187,5 +192,65 @@ def main():
...
+        else:
+            msg = (
+                'pythonw executable not found. '
+                'The menubar might only become '
+                'functional on the macOS after '
+                'focus is toggled in and out of '
+                'napari. To fix this problem, '
+                'please install python.app via conda, '
+                'which can be done with '
+                '`conda install -c conda-forge python.app`.'
```
jni:
```suggestion
                'pythonw executable not found.\n'
                'To unfreeze the menubar on macOS, '
                'click away from napari to another app, '
                'then reactivate napari. To avoid this problem, '
                'please install python.app in conda using:\n'
                'conda install -c conda-forge python.app'
```

jni (approved):
@sofroniewn I've suggested one critical fix and some wording changes. Please fix the `raise` call before merging!

jni:
Yes I do need to update the conda-forge recipe, for reference the repo is at: https://github.com/conda-forge/napari-feedstock

Issue #1516, opened 2020-08-02
Enable per-label visibility control for labels layer


qizhuli:
## 🚀 Feature
The requested feature is for the labels layer to have a fine-grained visibility control for individual label values, which enables users to focus on certain label values by turning the rest invisible.

## Motivation
Often labels image can have a large number of labels. To make things worse, one label can have bits and pieces scattered in different parts of an image. These two factors can make inspecting and understanding a label image quite difficult at times. 

An example of a complex label image:
![image](https://user-images.githubusercontent.com/15232694/89121625-4d0d2700-d4b8-11ea-93d8-b31919d4127f.png)

My work involves inspecting and analysing such label images. As a workaround, I have been parsing my label images, generating further `N` label images corresponding to `N` labels in the image, and adding all these `N+1` label images to the viewer. But it is inefficient, and creates a lot of clutter in the user interface -- especially when `N` is large.

## Pitch
Therefore, it would be desirable to build in a control mechanism for labels layer that allows the users to turn on/off the visibility of certain label values in a labels image, while keeping the label colours unaffected.

sofroniewn:
This is a great feature request! I wonder in the short term if a better workaround might involve the new `properties` dictionary added to the labels layer and the ability to use custom colormaps. See this example here https://github.com/napari/napari/blob/master/examples/add_labels_with_properties.py Maybe @DragaDoncila can also give some suggestions.

I'm curious how that goes, we can then think if there are things we could do to make the processes simpler. @hectormz has also been working on labels lately and might have some ideas too!

DragaDoncila:
This seems like a fairly key feature for working with labels. I would say the `properties` dictionary could be leveraged to enable/disable visibility based on the values of the properties or the value of the label itself e.g. show all labels with a particular classification.

In terms of working with existing features, one option I can think of (which is certainly suboptimal), is to use the custom colormap to assign colours only to labels which exhibit a particular property. Using the [coins example](https://github.com/napari/napari/blob/master/examples/add_labels_with_properties.py), line 46 could be replaced by 

```
small_coin_indices = list(np.where(coin_sizes == 'small')[0])
color = dict(zip(
        small_coin_indices,
        ['white' for i in range(len(small_coin_indices))]
    ))
```
to colour all small coins white.

In terms of toggling individual labels though, especially through the viewer which would be ideal, I think that needs to be implemented, perhaps by adding a `visible` property which could be manipulated both based on the value of other properties, and individually.

[...]

PR #1555, opened 2020-08-17
Add SELECTED color-mode for individual label visibility

DragaDoncila:
# Description
Following on from discussions in #1516 , this PR adds a Selected color-mode for labels layers, which allows for only selected labels to be visible. The user can select 'Selected' from the combo box. Deselected labels are coloured black, and the user can change the selected label using the existing spin box.

If the color_mode is changed to auto and then back to selected, the colormap is preserved.

selected_labels

So far we don't have support for multiple selections, but this can come in a future PR.

We still need to add some dedicated tests, and of course, all suggestions are welcome!

## Type of change

# How has this been tested?

## Final checklist:


jni:
This PR ended up so tiny! I love it! LOL

DragaDoncila:
@jni @sofroniewn I had a look at adding some tests today, but I can't find anywhere where we explicitly test setting the `color_mode`. Have I missed it?

If I haven't missed it, should setting `color_mode` be a test in its own right in `test_labels.py` or `test_labels_mouse_bindings.py` (or both), or is it being covered implicitly through other tests?

jni:
@DragaDoncila you might recall that you had to fix the color mode setting on the UI, so that would indicate that it's probably not a well-tested part of the code. I would make a test in `test_labels.py`. It would be great to also add a test to test the UI. @sofroniewn could we have some guidance about what that would look like? 😬

jni:
@DragaDoncila @sofroniewn this folder is probably the most useful place to add a test (and refer to) for the color mode UI updating correctly:

https://github.com/napari/napari/tree/67eb6908f77d93d064a69a6b4f6b48034ac60a90/napari/_qt/layer_controls/_tests

sofroniewn:
> @DragaDoncila @sofroniewn this folder is probably the most useful place to add a test (and refer to) for the color mode UI updating correctly:
>
> https://github.com/napari/napari/tree/67eb6908f77d93d064a69a6b4f6b48034ac60a90/napari/_qt/layer_controls/_tests

Yes, that seems like a good place to me!!

DragaDoncila:
I've added a couple of tests, making sure the colors of each label are correct when changing to selected mode. I've also added two tests for the GUI, one going from model change to GUI change, and one from GUI change to model change. I've left a comment on that latter one, because we may wish to change the event we connect to for selecting color mode.

napari/_qt/layer_controls/_tests/test_qt_labels_layer.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def test_changing_combo_box_updates_layer_color_mode(qtbot):
+    """Selecting a new color mode via combo box updates layer's color mode"""
+    layer = Labels(_LABELS)
+    qtctrl = QtLabelsControls(layer)
+    qtbot.addWidget(qtctrl)
+    color_mode_comboBox = qtctrl.colorModeComboBox
+
+    index = color_mode_comboBox.findText('selected', Qt.MatchFixedString)
```
DragaDoncila: This fails because we connect the `color_mode_comboBox` to `activated` rather than `current_index_changed`. `activated` only fires through user interaction, not programmatic changes. I tried to simulate a button click instead, similar to this image layer test [here](https://github.com/napari/napari/blob/62a046de2e5009f0c9137e3fd3999fbf902928ca/napari/_qt/layer_controls/_tests/test_qt_image_base_layer_.py#L42), but I cannot for the life of me figure out how to select the item for clicking, as it's not a button in the eyes of qt.

@sofroniewn perhaps you know how to simulate the click on a `QComboBox` item?

DragaDoncila: On discussion with @jni have removed this test and only kept the test checking api changes reflect in gui.

jni (approved):

sofroniewn (approved):
LGTM! Very nice how small a PR this ended up!

PR #1560, opened 2020-08-18
Fix typo in perfmon patcher

kevinyamauchi:
# Description
This PR fixes a small typo in the perfmon patcher (`module.__name___` -> `module.__name__`).

## Type of change

# References

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/utils/perf/_patcher.py
+++ b/napari/utils/perf/_patcher.py
@@ -57,7 +57,7 @@ def _patch_attribute(
         # Assume attribute_str is <function>.
         class_str = None
         parent = module
-        parent_str = module.__name___
+        parent_str = module.__name__
         callable_str = attribute_str
 
     try:


sofroniewn (approved):
Thanks for that fix @kevinyamauchi, it looks like it wasn't caught in #1550 either, cc @pwinston.

pwinston:
@kevinyamauchi good find. That's weird though because it ran okay, wouldn't the old version have been an error? Doesn't matter I guess, good to have it fixed. Thanks.

kevinyamauchi:
> That's weird though because it ran okay, wouldn't the old version have been an error?

For me, it was on a code path that was only executed when the thing being profiled was a function. Patching class methods worked fine!

pwinston:
Thanks, that makes sense. I probably never traced functions. I'm going to create a very simple perfmon test for #1550. Not sure if I'll test the Patcher, but at least make sure it can create a trace JSON at all. We can add more tests over time. Tests are great, but if more people use it that will also create a good feedback loop.

sofroniewn:
> I'm going to create a very simple perfmon test for #1550. Not sure if I'll test the Patcher, but at least make sure it can create a trace JSON at all. We can add more tests over time. 

Thanks @pwinston that sounds great! I'm going to merge this now as the fix seems uncontroversial, but absolutely agree with the following too

> Tests are great, but if more people use it that will also create a good feedback loop.

and thanks again for the catch @kevinyamauchi!

PR #1561, opened 2020-08-18
Improve Shapes layer performance

kevinyamauchi:
# Description
This PR implements some small changes to improve the responsiveness of the Shapes layer with a large number of shapes. In particular, based on the profiling we did in #1562 , I aimed to do the following:

- Reduce startup time
    - Delay re-indexing the mesh by z order until all shapes are added
    - Eliminate a redundant `Shapes._update_thumbnail()` call
- Decrease latency when switching interaction modes and selecting shapes
    - Eliminate a redundant `Shapes._update_thumbnail()` call
    - Do not call `Shapes._update_thumbnail()` when selecting shapes and there are more than 100 shapes
- Decrease latency during dragging
    - Delay updating the thumbnail until the drag is complete

### Profiling
I used the airspeed velocity benchmarks listed below (`/napari/benchmarks/benchmarks_shapes_layer.py`).

| Benchmark                                | n_shapes | Master [sec.] | PR [sec.]  | Change |
|------------------------------------------|----------|--------|------|--------|
| Shapes2DSuite.time_create_layer          | 256      | 1.93   | 1.25 | -35%   |
| ShapesInteractionSuite.time_select_shape | 256      | 0.79   | 0.01 | -99%   |
| ShapesInteractionSuite.time_drag_shape   | 256      | 1.23   | 0.18 | -85%   |

A couple of notes:
- `ShapesInteractionSuite.time_drag_shape` benchmark performs 5 drag events, the time per drag event is ~30 ms
- These only include the time for calculations performed on the model (i.e., they don't include the vispy draw time)
-  The layer creation is likely limited by the creation of each `Shape` object and the thumbnail generation.

### Demo
If you'd like to try it out, you can use the script below to generate a Shapes layer with 1000 shapes and compare between the PR and master. At least on my computer, this seems to be an improvement, but there are definitely more things we can do (e.g., speed up the calculation for selecting shapes with a bounding box)).

<details><summary>Demo script</summary>
<p>

```python
import numpy as np
import napari
from skimage import data

# make 3D image
blobs = data.binary_blobs(
    length=128, blob_size_fraction=0.05, n_dim=3, volume_fraction=0.1
).astype(float)


def make_square(lower_left_corner, width):
    upper_left_corner = lower_left_corner + [0, width]
    upper_right_corner = lower_left_corner + [width, width]
    lower_right_corner = lower_left_corner + [width, 0]

    return np.vstack((lower_left_corner, upper_left_corner, upper_right_corner, lower_right_corner))

# create random rectangles
n_shapes = 1000
np.random.seed(0)
corners = 100 * np.random.random((n_shapes, 2))
widths = 15 * np.random.random((n_shapes,))
squares = [make_square(llc, w) for llc, w in zip(corners, widths)]

base_cols = ['red', 'green', 'blue', 'white', 'yellow', 'magenta', 'cyan']
colors = np.random.choice(base_cols, size=n_shapes)

with napari.gui_qt():

    viewer = napari.view_image(blobs)

    shapes_layer = viewer.add_shapes(
        squares,
        shape_type='rectangle',
        face_color=colors,
        name='sliced',
    )
```

</p>
</details>

## Type of change

# References

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/layers/base/base.py
+++ b/napari/layers/base/base.py
@@ -192,6 +192,7 @@ class Layer(KeymapProvider, ABC):
         self._thumbnail_shape = (32, 32, 4)
         self._thumbnail = np.zeros(self._thumbnail_shape, dtype=np.uint8)
         self._update_properties = True
+        self._allow_thumbnail_update = True
         self._name = ''
         self.events = EmitterGroup(
             source=self,
@@ -637,6 +638,12 @@ class Layer(KeymapProvider, ABC):
         self._update_properties = False
         yield
         self._update_properties = True
+
+    @contextmanager
+    def block_thumbnail_update(self):
+        self._allow_thumbnail_update = False
+        yield
+        self._allow_thumbnail_update = True
 
     def _set_highlight(self, force=False):
         """Render layer highlights when appropriate.
--- a/napari/layers/base/base.py
+++ b/napari/layers/base/base.py
@@ -192,7 +192,6 @@ class Layer(KeymapProvider, ABC):
         self._thumbnail_shape = (32, 32, 4)
         self._thumbnail = np.zeros(self._thumbnail_shape, dtype=np.uint8)
         self._update_properties = True
-        self._allow_thumbnail_update = True
         self._name = ''
         self.events = EmitterGroup(
             source=self,
@@ -638,13 +637,6 @@ class Layer(KeymapProvider, ABC):
         self._update_properties = False
         yield
         self._update_properties = True
-
-    @contextmanager
-    def block_thumbnail_update(self):
-        """Use this context manager to block thumbnail updates"""
-        self._allow_thumbnail_update = False
-        yield
-        self._allow_thumbnail_update = True
 
     def _set_highlight(self, force=False):
         """Render layer highlights when appropriate.
--- a/napari/layers/base/base.py
+++ b/napari/layers/base/base.py
@@ -641,6 +641,7 @@ class Layer(KeymapProvider, ABC):
 
     @contextmanager
     def block_thumbnail_update(self):
+        """Use this context manager to block thumbnail updates"""
         self._allow_thumbnail_update = False
         yield
         self._allow_thumbnail_update = True
--- a/napari/layers/shapes/_shape_list.py
+++ b/napari/layers/shapes/_shape_list.py
@@ -217,7 +217,7 @@ class ShapeList:
         face_color=None,
         edge_color=None,
         shape_index=None,
-        reindex_z=True,
+        z_refresh=True,
     ):
         """Adds a single Shape object
 
@@ -230,8 +230,10 @@ class ShapeList:
             If int then edits the shape date at current index. To be used in
             conjunction with `remove` when renumber is `False`. If None, then
             appends a new shape to end of shapes list
-        reindex_z : bool
+        z_refresh : bool
             If set to true, the mesh elements are reindexed with the new z order.
+            When shape_index is provided, z_refresh will be overwritten to false,
+            as the z indices will not change.
             When adding a batch of shapes, set to false  and then call
             ShapesList._update_z_order() once at the end.
         """
@@ -239,10 +241,6 @@ class ShapeList:
             raise ValueError('shape must be subclass of Shape')
 
         if shape_index is None:
-            if reindex_z is True:
-                z_refresh = True
-            else:
-                z_refresh = False
             shape_index = len(self.shapes)
             self.shapes.append(shape)
             self._z_index = np.append(self._z_index, shape.z_index)
--- a/napari/layers/shapes/_shape_list.py
+++ b/napari/layers/shapes/_shape_list.py
@@ -841,7 +841,9 @@ class ShapeList:
 
         return labels
 
-    def to_colors(self, colors_shape=None, zoom_factor=1, offset=[0, 0]):
+    def to_colors(
+        self, colors_shape=None, zoom_factor=1, offset=[0, 0], max_shapes=None
+    ):
         """Rasterize shapes to an RGBA image array.
 
         Each shape is embedded in an array of shape `colors_shape` with the
@@ -859,6 +861,11 @@ class ShapeList:
         offset : 2-tuple
             Offset subtracted from coordinates before multiplying by the
             zoom_factor. Used for putting negative coordinates into the mask.
+        max_shapes : None | int
+            If provided, this is the maximum number of shapes that will be rasterized.
+            If the number of shapes in view exceeds max_shapes, max_shapes shapes
+            will be randomly selected from the in view shapes. If set to None, no
+            maximum is applied. The default value is None.
 
         Returns
         -------
@@ -872,15 +879,25 @@ class ShapeList:
         colors = np.zeros(tuple(colors_shape) + (4,), dtype=float)
         colors[..., 3] = 1
 
-        for ind in self._z_order[::-1]:
-            if self._displayed[ind]:
-                mask = self.shapes[ind].to_mask(
-                    colors_shape, zoom_factor=zoom_factor, offset=offset
-                )
-                if type(self.shapes[ind]) in [Path, Line]:
-                    col = self._edge_color[ind]
-                else:
-                    col = self._face_color[ind]
-                colors[mask, :] = col
+        z_order = self._z_order[::-1]
+        shapes_in_view = np.argwhere(self._displayed)
+        z_order_in_view_mask = np.isin(z_order, shapes_in_view)
+        z_order_in_view = z_order[z_order_in_view_mask]
+
+        if max_shapes is not None and len(z_order_in_view) > max_shapes:
+            indices_to_use = np.random.randint(
+                0, len(z_order_in_view), max_shapes
+            )
+            z_order_in_view = z_order_in_view[indices_to_use]
+
+        for ind in z_order_in_view:
+            mask = self.shapes[ind].to_mask(
+                colors_shape, zoom_factor=zoom_factor, offset=offset
+            )
+            if type(self.shapes[ind]) in [Path, Line]:
+                col = self._edge_color[ind]
+            else:
+                col = self._face_color[ind]
+            colors[mask, :] = col
 
         return colors
--- a/napari/layers/shapes/_shape_list.py
+++ b/napari/layers/shapes/_shape_list.py
@@ -885,10 +885,7 @@ class ShapeList:
         z_order_in_view = z_order[z_order_in_view_mask]
 
         if max_shapes is not None and len(z_order_in_view) > max_shapes:
-            indices_to_use = np.random.randint(
-                0, len(z_order_in_view), max_shapes
-            )
-            z_order_in_view = z_order_in_view[indices_to_use]
+            z_order_in_view = z_order_in_view[0:max_shapes]
 
         for ind in z_order_in_view:
             mask = self.shapes[ind].to_mask(
--- a/napari/layers/shapes/_shape_list.py
+++ b/napari/layers/shapes/_shape_list.py
@@ -211,7 +211,14 @@ class ShapeList:
         self.displayed_vertices = self._vertices[disp_vert]
         self.displayed_index = self._index[disp_vert]
 
-    def add(self, shape, face_color=None, edge_color=None, shape_index=None):
+    def add(
+        self,
+        shape,
+        face_color=None,
+        edge_color=None,
+        shape_index=None,
+        reindex_z=True,
+    ):
         """Adds a single Shape object
 
         Parameters
@@ -223,12 +230,19 @@ class ShapeList:
             If int then edits the shape date at current index. To be used in
             conjunction with `remove` when renumber is `False`. If None, then
             appends a new shape to end of shapes list
+        reindex_z : bool
+            If set to true, the mesh elements are reindexed with the new z order.
+            When adding a batch of shapes, set to false  and then call
+            ShapesList._update_z_order() once at the end.
         """
         if not issubclass(type(shape), Shape):
             raise ValueError('shape must be subclass of Shape')
 
         if shape_index is None:
-            z_refresh = True
+            if reindex_z is True:
+                z_refresh = True
+            else:
+                z_refresh = False
             shape_index = len(self.shapes)
             self.shapes.append(shape)
             self._z_index = np.append(self._z_index, shape.z_index)
--- a/napari/layers/shapes/_shape_list.py
+++ b/napari/layers/shapes/_shape_list.py
@@ -884,6 +884,8 @@ class ShapeList:
         z_order_in_view_mask = np.isin(z_order, shapes_in_view)
         z_order_in_view = z_order[z_order_in_view_mask]
 
+        # If there are too many shapes to render responsively, just render
+        # the top max_shapes shapes
         if max_shapes is not None and len(z_order_in_view) > max_shapes:
             z_order_in_view = z_order_in_view[0:max_shapes]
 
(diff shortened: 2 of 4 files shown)


kevinyamauchi:
Okay, I think this is ready for an initial round of review. I have added some details on what was changed and the profiling results above. I think there are still many things we can do, but I think these are the low hanging fruit and it is a step in the right direction.

napari/layers/base/base.py
```diff
@@ -638,6 +639,13 @@ def block_update_properties(self):
         yield
         self._update_properties = True
 
+    @contextmanager
```
kevinyamauchi: I put this here because I thought we might want to use it on other layers. However, I didn't implement it on the other layers. In hindsight, maybe it's better to put it just on the Shapes layer for now and then if it becomes a more general strategy, we can move it back to the Base layer. Thoughts?

napari/layers/shapes/_shape_list.py
```diff
@@ -223,12 +230,19 @@ def add(self, shape, face_color=None, edge_color=None, shape_index=None):
...
+            If set to true, the mesh elements are reindexed with the new z order.
+            When adding a batch of shapes, set to false  and then call
+            ShapesList._update_z_order() once at the end.
         """
         if not issubclass(type(shape), Shape):
             raise ValueError('shape must be subclass of Shape')
 
         if shape_index is None:
-            z_refresh = True
+            if reindex_z is True:
```
sofroniewn: Can you just drop `z_refresh` and use `reindex_z`?

Or just do `z_refresh = bool(reindex_z)`? And drop the if else?

napari/layers/base/base.py
sofroniewn: if it goes on base it should be implemented everywhere. maybe for now we just put on shapes as only implemented there, and then we can always move to base and implement everywhere later

sofroniewn (approved):
Wow, love this PR!!! Really noticeable on rearranging polygons on a ~300 segmented neuron calcium imaging dataset I have to play with :-)

Very nice! I left some fairly minor comments which should be easy to address.

Could you run the most relevant benchmarks before/ after this PR and paster results in as a comment, could just help us keep up with performance improvements/ and future regressions by having some numbers in here.

Thanks!!

napari/layers/base/base.py
kevinyamauchi: That makes sense to me!

napari/layers/shapes/_shape_list.py
kevinyamauchi: Good idea. I will reuse `z_refresh`

kevinyamauchi: Done!

napari/layers/base/base.py
kevinyamauchi: Done!

kevinyamauchi:
edit: the benchmarks have been updated to reflect the changes in https://github.com/napari/napari/pull/1561#issuecomment-679348352

# Benchmarks

### Key results
Here is a table of the most relevent benchmarks to this PR. I have pasted the full benchmarks below.

| Benchmark                                | n_shapes | Master [sec.] | PR [sec.]  | Change |
|------------------------------------------|----------|--------|------|--------|
| Shapes2DSuite.time_create_layer          | 256      | 2.17   | 1.34 | -38%   |
| ShapesInteractionSuite.time_update_thumbnail | 256      | 0.278   | 0.114 | -59%   |
| ShapesInteractionSuite.time_select_shape | 256      | 0.79   | 0.008 | -99%   |
| ShapesInteractionSuite.time_drag_shape   | 256      | 1.54   | 0.065 | -95%   |


### Shapes2DSuite

<details><summary> Master Results</summary>
<p>

```
[  0.00%] ·· Benchmarking existing-py
[  7.14%] ··· benchmark_shapes_layer.Shapes2DSuite.mem_data                   ok
[  7.14%] ··· ======== =======
               param1
              -------- -------
                 16      200
                 32      328
                 64      584
                128      1.1k
                256     2.12k
              ======== =======

[ 14.29%] ··· benchmark_shapes_layer.Shapes2DSuite.mem_layer                  ok
[ 14.29%] ··· ======== =====
               param1
              -------- -----
                 16     824
                 32     824
                 64     824
                128     824
                256     824
              ======== =====

[ 21.43%] ··· ...shapes_layer.Shapes2DSuite.time_create_layer                 ok
[ 21.43%] ··· ======== =========
               param1
              -------- ---------
                 16     140±0ms
                 32     240±0ms
                 64     541±0ms
                128     1.09±0s
                256     2.17±0s
              ======== =========

[ 28.57%] ··· ...rk_shapes_layer.Shapes2DSuite.time_get_value                 ok
[ 28.57%] ··· ======== =========
               param1
              -------- ---------
                 16     123±0μs
                 32     207±0μs
                 64     262±0μs
                128     595±0μs
                256     876±0μs
              ======== =========

[ 35.71%] ··· ...mark_shapes_layer.Shapes2DSuite.time_refresh                 ok
[ 35.71%] ··· ======== ==========
               param1
              -------- ----------
                 16     17.4±0ms
                 32     32.3±0ms
                 64     75.2±0ms
                128     145±0ms
                256     272±0ms
              ======== ==========

[ 42.86%] ··· ...apes_layer.Shapes2DSuite.time_set_view_slice                 ok
[ 42.86%] ··· ======== ==========
               param1
              -------- ----------
                 16     58.9±0μs
                 32     64.9±0μs
                 64     73.9±0μs
                128     111±0μs
                256     136±0μs
              ======== ==========

[ 50.00%] ··· ...es_layer.Shapes2DSuite.time_update_thumbnail                 ok
[ 50.00%] ··· ======== ==========
               param1
              -------- ----------
                 16     14.8±0ms
                 32     34.6±0ms
                 64     102±0ms
                128     138±0ms
                256     278±0ms
              ======== ==========

```
</p>
</details>

<details><summary>PR Results</summary>
<p>

```
[  0.00%] ·· Benchmarking existing-py_Users_yamauc0000_miniconda3_envs_napari-dev_bin_python
[  7.14%] ··· benchmark_shapes_layer.Shapes2DSuite.mem_data                                                                                                                                        ok
[  7.14%] ··· ======== =======
               param1         
              -------- -------
                 16      200  
                 32      328  
                 64      584  
                128      1.1k 
                256     2.12k 
              ======== =======

[ 14.29%] ··· benchmark_shapes_layer.Shapes2DSuite.mem_layer                                                                                                                                       ok
[ 14.29%] ··· ======== =====
               param1       
              -------- -----
                 16     824 
                 32     824 
                 64     824 
                128     824 
                256     824 
              ======== =====

[ 21.43%] ··· benchmark_shapes_layer.Shapes2DSuite.time_create_layer                                                                                                                               ok
[ 21.43%] ··· ======== =========
               param1           
              -------- ---------
                 16     102±0ms 
                 32     181±0ms 
                 64     353±0ms 
                128     707±0ms 
                256     1.34±0s 
              ======== =========

[ 28.57%] ··· benchmark_shapes_layer.Shapes2DSuite.time_get_value                                                                                                                                  ok
[ 28.57%] ··· ======== =========
               param1           
              -------- ---------
                 16     114±0μs 
                 32     181±0μs 
                 64     286±0μs 
                128     373±0μs 
                256     681±0μs 
              ======== =========

[ 35.71%] ··· benchmark_shapes_layer.Shapes2DSuite.time_refresh                                                                                                                                    ok
[ 35.71%] ··· ======== ==========
               param1            
              -------- ----------
                 16     16.8±0ms 
                 32     32.9±0ms 
                 64     63.6±0ms 
                128     103±0ms  
                256     103±0ms  
              ======== ==========

[ 42.86%] ··· benchmark_shapes_layer.Shapes2DSuite.time_set_view_slice                                                                                                                             ok
[ 42.86%] ··· ======== ==========
               param1            
              -------- ----------
                 16     61.3±0μs 
                 32     71.5±0μs 
                 64     85.9±0μs 
                128     60.5±0μs 
                256     72.9±0μs 
              ======== ==========

[ 50.00%] ··· benchmark_shapes_layer.Shapes2DSuite.time_update_thumbnail                                                                                                                           ok
[ 50.00%] ··· ======== ==========
               param1            
              -------- ----------
                 16     17.4±0ms 
                 32     32.1±0ms 
                 64     73.1±0ms 
                128     104±0ms  
                256     114±0ms  
              ======== ==========

```
</p>
</details>

### ShapesInteractionSuite

<details><summary>Master Results</summary>
<p>

```
[  0.00%] ·· Benchmarking existing-py
[ 25.00%] ··· ...layer.ShapesInteractionSuite.time_drag_shape                 ok
[ 25.00%] ··· ========== ==========
               n_shapes
              ---------- ----------
                  16      76.0±0ms
                  32      265±0ms
                  64      368±0ms
                 128      797±0ms
                 256      1.54±0s
              ========== ==========

[ 50.00%] ··· ...yer.ShapesInteractionSuite.time_select_shape                 ok
[ 50.00%] ··· ========== ==========
               n_shapes
              ---------- ----------
                  16      58.0±0ms
                  32      133±0ms
                  64      387±0ms
                 128      850±0ms
                 256      913±0ms
              ========== ==========
```

</p>
</details>

<details><summary>PR Results</summary>
<p>

```
[  0.00%] ·· Benchmarking existing
[ 25.00%] ··· benchmark_shapes_layer.ShapesInteractionSuite.time_drag_shape                                                                                                                        ok
[ 25.00%] ··· ========== ==========
               n_shapes            
              ---------- ----------
                  16      12.9±0ms 
                  32      21.2±0ms 
                  64      35.6±0ms 
                 128      63.5±0ms 
                 256      65.3±0ms 
              ========== ==========

[ 50.00%] ··· benchmark_shapes_layer.ShapesInteractionSuite.time_select_shape                                                                                                                      ok
[ 50.00%] ··· ========== ==========
               n_shapes            
              ---------- ----------
                  16      977±0μs  
                  32      815±0μs  
                  64      3.96±0ms 
                 128      2.16±0ms 
                 256      770±0μs  
              ========== ==========

```

<\p>
<\details>

kevinyamauchi:
@sofroniewn , thanks for the review! I have addressed your comments.

napari/layers/shapes/shapes.py
```diff
@@ -55,6 +56,10 @@
 
 DEFAULT_COLOR_CYCLE = np.array([[1, 0, 1, 1], [0, 1, 0, 1]])
 
+# If there are more than this number of shapes, the thumnail won't update during
+# interactive events
+INTERACTIVE_THUMBNAIL_UPDATE_THRESHOLD = 100
```
sofroniewn: Something I wonder about is instead of having this threshold for interactive updating is to make all interactive updating off, AND have us use at most this number of randomly chosen shapes for a thumbnail (which is only ever 32x32 pixels), see what we do here in vectors https://github.com/napari/napari/blob/master/napari/layers/vectors/vectors.py#L676-L681

napari/layers/shapes/shapes.py
```diff
@@ -1369,9 +1387,20 @@ def mode(self, mode):
         ]
 
         self.events.mode(mode=mode)
-        if not (mode in draw_modes and old_mode in draw_modes):
-            self._finish_drawing()
-        self.refresh()
+        # only update the thumbnail if there are < 100 shapes
+        if len(self.data) > self._thumbnail_update_thresh:
```
sofroniewn: See comment above, I think it's cleaner if we just always do the block thumbnail path here AND additionally have a catch if we have > 100 shapes to pick those 100 randomly like we do in vectors

sofroniewn: Thinking about this one more, changing mode should also never need a thumbnail update, as changing mode shouldn't change the data, so it's probably always worth blocking thumbnail generation here!

sofroniewn:
Amazing @kevinyamauchi 99% timesavings for dragging - this PR is huge. Thanks for including those benchmark stats!!!

I've left one final comment that we might want to additionally actually cap all thumbnail updates to at most use 100 random shapes if there are more than 100 shapes on that slice (note not 100 overall, but on that slice). For a 32 x 32 pixel thumbnail, I wonder if more than 100 are ever needed. You could also take the top 100 shapes according to z-index. You could also then remove the additional split on if to block thumbnail during the changing of mode, which should probably never need a thumbnail calculation.

Let me know if you want to make those two changes, I think it is very close to merge.

napari/layers/shapes/shapes.py
kevinyamauchi: > Thinking about this one more, changing mode should also never need a thumbnail update, as changing mode shouldn't change the data, so it's probably always worth blocking thumbnail generation here!

That make sense to me. I will set it so that we always block thumbnail regeneration during mode changes.

napari/layers/shapes/shapes.py
kevinyamauchi: Oh yeah, I forgot that we do that on some layers! That's a good idea. I agree that we should do the same here.

napari/layers/shapes/shapes.py
kevinyamauchi: Done!

kevinyamauchi:
@sofroniewn, I have done the following, which I think addresses your comments.

- Made it such that the thumbnail is not updated for any layer mode change
- Made it so that the selection events never update the thumbnail. The thumbnail still updates if a shape is moved, modified, or created.
- Made it so when the number of shapes _in view_ exceeds a threshold (`Shapes._max_shapes_thumbnail`), a number of in view shapes are randomly selected to be rasterized. I set the threshold to 100 for now, but we can relax this as we further improve the rasterization code (this yields update times of ~100 ms on my Macbook Air).

I have updated the benchmarks in this comment: https://github.com/napari/napari/pull/1561#issuecomment-678771901

napari/layers/shapes/shapes.py
kevinyamauchi: Done!

napari/layers/shapes/_shape_list.py
```diff
@@ -860,15 +879,25 @@ def to_colors(self, colors_shape=None, zoom_factor=1, offset=[0, 0]):
...
+        z_order = self._z_order[::-1]
+        shapes_in_view = np.argwhere(self._displayed)
+        z_order_in_view_mask = np.isin(z_order, shapes_in_view)
+        z_order_in_view = z_order[z_order_in_view_mask]
+
+        if max_shapes is not None and len(z_order_in_view) > max_shapes:
+            indices_to_use = np.random.randint(
+                0, len(z_order_in_view), max_shapes
+            )
+            z_order_in_view = z_order_in_view[indices_to_use]
```
sofroniewn:
```suggestion
            z_order_in_view = z_order_in_view[:max_shapes]
```

I think we should actually take to top shapes here (syntax could be with `z_order_in_view[-max_shapes:]`) instead, you have to check.

Reason is that right now every time i drag a shape around the thumbnail picks a random 100, which looks really funny - see

shapes_thumbnails_changing

sofroniewn:
Hi @kevinyamauchi thanks for all those updates. Playing with the new behavior I have one final request which is to not make the selection randomly, but take the top `Shapes._max_shapes_thumbnail` shapes for the thumbnail, should be a simple one line change, but you should make sure to take the top 100 not the bottom 100.

Reason for this is shown in the gif placed in the inline comment, but right now a small change to the shapes triggers a big change to the thumbnail as the random 100 chosen are different every time which is quite visually distracting if you're moving shapes around.

Otherwise everything looks great, and we can merge after this!

napari/layers/shapes/_shape_list.py
kevinyamauchi: Hey @sofroniewn , that's a great point. I've made the suggested change.

kevinyamauchi:
Thanks for the careful review, @sofroniewn ! I agree that taking the "top" (by z order) `max_shapes` is the way to go so the thumbnail is stable for dense layers. I have made the suggested change.

sofroniewn:
Looks great to me @kevinyamauchi, I am merging now!!

Issue #1355, opened 2020-06-13
ImageSlice docstring needs updating


jni:
## 📚 Documentation

See discussion here:

https://github.com/napari/napari/pull/1343#discussion_r439445553

Specifically, I object to using the type alias ImageConverter, and I strongly object to using it in the docstring. Additionally, the description of the argument should explain that any array -> array callable will work here.

pwinston:
I think there are two separate questions here:
1. When should a type alias be used vs. just spelling out the full type everytime.
2. When using a type alias, what should the docstring say: the alias, the full type, or both.

For 1 my feeling is the codebase is nothing but a huge pile of symbols: variables, arguments, parameters, methods, functions, classes, modules, packages. To me introducing nicely-named symbols for types is just part of programming. And not doing so leads to repetitive bloated code that's hard to read and maintain. That said, I know some reasonable people feel the exact opposite, they feel that type aliases obscure things.

For 2 the idea that the type should be different in both places seems crazy to me, as in you are telling the developer two different answers the same question. But the idea that it should be repeated identically in both places seems crazy too, as in it's pure duplication. The whole idea of tediously entering types into comment blocks which might get totally out of sync with the code seems like a bad idea. It must be vestigial from before type annotations existed? I think the numpydoc world is heading towards using type annotations as the single point of truth, but they have not made the transition yet:
* https://github.com/numpy/numpydoc/issues/196
* [sphynx-autdoc-typehints](https://github.com/agronholm/sphinx-autodoc-typehints)

In both cases I'm happy to do whatever the project wants here. I think projects know their own unique needs and goals. However ideally we can set clear guidelines so in the future people can get the PR right the first time without as much discussion.

pwinston:
I think the concrete options here are:

pwinston:
I think there's just two concrete decisions to make for this PR:

1. Leave `ImageConverter` in the code or replace it with `Callable[[ArrayLike], ArrayLike]` in the two places it appears?
2. Leave `ImageConverter` in the docstring or put `Callable[[ArrayLike], ArrayLike]` there instead?

And then decide in general when do we use type aliases and how do we document them?

jni:
Finally revisiting this during my review of #1565. I would like to replace it with `Callable[[ArrayLike], ArrayLike]` wherever it appears. It is longer but instantly readable to anyone with a passing familiarity with NumPy and type annotations.

pwinston:
[...]

PR #1565, opened 2020-08-20
async-1: Stub ChunkLoader

pwinston:
# Description
* Image changes and stub `ChunkLoader`

# Details
* Adds `napari.components.chunk` module with stub `ChunkLoader`, `ChunkKey` and `
ChunkRequest`
* `ChunkLoader` is a stub that always loads **synchronously**.
* Add `Layer.loaded` property and `loaded` event
    * By default layers are permanently in the loaded state
* Adds `Image._load_images()` and `Image.chunk_loaded()` methods.
* Small `ImageView` change.
* New `ImageSlice.load_chunk()` and `ImageSlice.chunk_loaded()` methods.
* It loads **synchronously** 100% of the time.

### DIFF
--- a/napari/components/chunk/_request.py
+++ b/napari/components/chunk/_request.py
@@ -1,7 +1,7 @@
 """ChunkRequest is passed to ChunkLoader.load_chunks().
 """
 import logging
-from typing import Tuple, Union
+from typing import Optional, Tuple, Union
 
 import numpy as np
 
@@ -30,13 +30,13 @@ class ChunkKey:
         The id of the layer making the request.
     data_level : int
         The level in the data (for multi-scale).
-    indices : Indices
+    indices : Tuple[Optional[slice], ...]
         The indices of the slice.
     key : Tuple
         The combined key, all the identifies together.
     """
 
-    def __init__(self, layer: Layer, indices):
+    def __init__(self, layer: Layer, indices: Tuple[Optional[slice], ...]):
         self.layer_id = id(layer)
         self.data_level = layer._data_level
 
--- a/napari/components/chunk/_request.py
+++ b/napari/components/chunk/_request.py
@@ -11,7 +11,7 @@ from ...types import ArrayLike, Dict
 LOGGER = logging.getLogger("napari.async")
 
 # We convert slices to tuple for hashing.
-SliceTuple = Tuple[int, int, int]
+SliceTuple = Tuple[Optional[int], Optional[int], Optional[int]]
 
 
 class ChunkKey:
@@ -33,7 +33,7 @@ class ChunkKey:
     indices : Tuple[Optional[slice], ...]
         The indices of the slice.
     key : Tuple
-        The combined key, all the identifies together.
+        The combined key, all the identifiers together.
     """
 
     def __init__(self, layer: Layer, indices: Tuple[Optional[slice], ...]):
@@ -119,16 +119,6 @@ class ChunkRequest:
             # No thumbnail_source so return the image instead. For single-scale
             # we use the image as the thumbnail_source.
             return self.chunks.get('image')
-
-    def is_compatible(self, layer: Layer) -> bool:
-        """Return True if the given data is compatible with this request.
-
-        Parameters
-        ----------
-        layer : Layer
-            Compare this data to the data_id in the request.
-        """
-        return True  # Stub for now, will grow in the next version.
 
 
 def _index_to_tuple(index: Union[int, slice]) -> Union[int, SliceTuple]:
--- a/napari/components/chunk/_request.py
+++ b/napari/components/chunk/_request.py
@@ -1,0 +1,164 @@
+"""ChunkRequest is used to ask the ChunkLoader to load chunks.
+"""
+import logging
+from typing import List, Tuple, Union
+
+import numpy as np
+
+from ...types import ArrayLike, Dict
+
+LOGGER = logging.getLogger("ChunkLoader")
+
+# We convert slices to tuple for hashing.
+SliceTuple = Tuple[int, int, int]
+
+LayerData = Union[ArrayLike, List[ArrayLike]]
+
+
+class ChunkKey:
+    """The key which a ChunkRequest will load.
+
+    Parameters
+    ----------
+    layer : Layer
+        The layer to load data for.
+    indices : ?
+        The indices to load from the layer
+    """
+
+    def __init__(self, layer, indices):
+        self.layer_id = id(layer)
+        self.data_level = layer._data_level
+
+        # Slice objects are not hashable, so turn them into tuples.
+        self.indices = tuple(_index_to_tuple(x) for x in indices)
+
+        # All together as one tuple for easy comparison.
+        self.key = (self.layer_id, self.data_level, self.indices)
+
+    def __str__(self):
+        return (
+            f"layer_id={self.layer_id} data_id={self.data_id} "
+            f"data_level={self.data_level} indices={self.indices}"
+        )
+
+    def __eq__(self, other):
+        return self.key == other.key
+
+
+class ChunkRequest:
+    """A request asking the ChunkLoader to load one or more arrays.
+
+    Parameters
+    ----------
+    layer_id : int
+        Python id() for the Layer requesting the chunk.
+    data_id : int
+        Python id() for the Layer._data requesting the chunk.
+    indices
+        The tuple of slices index into the data.
+    array : ArrayLike
+        Load the data from this array.
+
+    Attributes
+    ----------
+    layer_ref : weakref
+        Reference to the layer that submitted the request.
+    data_id : int
+        Python id() of the data in the layer.
+    load_seconds : float
+        Delay for this long during the load portion.
+    """
+
+    def __init__(self, key: ChunkKey, chunks: Dict[str, ArrayLike]):
+        # Make sure chunks is str->array as expected.
+        for chunk_key, array in chunks.items():
+            assert isinstance(chunk_key, str)
+            assert array is not None
+
+        self.key = key
+        self.chunks = chunks
+
+        # No delay by default, ChunkLoader.load_chunk() will set this if desired.
+        self.load_seconds = 0
+
+    @property
+    def num_chunks(self) -> int:
+        """Return the number of chunks in this request."""
+        return len(self.chunks)
+
+    @property
+    def num_bytes(self) -> int:
+        """Return the number of bytes that were loaded."""
+        return sum(array.nbytes for array in self.chunks.values())
+
+    @property
+    def in_memory(self) -> bool:
+        """True if all chunks are ndarrays."""
+        for array in self.chunks.values():
+            if not isinstance(array, np.ndarray):
+                return False
+        return True
+
+    def load_chunks(self):
+        """Load all of our chunks now in this thread."""
+        for key, array in self.chunks.items():
+            loaded_array = np.asarray(array)
+            self.chunks[key] = loaded_array
+
+    def transpose_chunks(self, order):
+        """Transpose all our chunks.
+
+        Parameters
+        ----------
+        order
+            Transpose the chunks with this order.
+        """
+        for key, array in self.chunks.items():
+            self.chunks[key] = array.transpose(order)
+
+    @property
+    def image(self):
+        """The image chunk if we have one or None.
+        """
+        return self.chunks.get('image')
+
+    @property
+    def thumbnail_source(self):
+        """The chunk to use as the thumbnail_source or None.
+        """
+        try:
+            return self.chunks['thumbnail_source']
+        except KeyError:
+            # For single-scale we use the image as the thumbnail_source.
+            return self.chunks.get('image')
+
+    def is_compatible(self, layer) -> bool:
+        """Return True if the given data is compatible with this request.
+
+        Parameters
+        ----------
+        data : LayerData
+            Compare this data to the data_id in the request.
+        """
+        return True  # stub for now
+
+
+def _index_to_tuple(index: Union[int, slice]) -> Union[int, SliceTuple]:
+    """Get hashable object for the given index.
+
+    Slice is not hashable so we convert slices to tuples.
+
+    Parameters
+    ----------
+    index
+        Integer index or a slice.
+
+    Returns
+    -------
+    Union[int, SliceTuple]
+        Hashable object that can be used for the index.
+    """
+    if isinstance(index, slice):
+        return (index.start, index.stop, index.step)
+    return index
--- a/napari/components/chunk/_request.py
+++ b/napari/components/chunk/_request.py
@@ -1,7 +1,7 @@
 """ChunkRequest is passed to ChunkLoader.load_chunks().
 """
 import logging
-from typing import List, Tuple, Union
+from typing import Tuple, Union
 
 import numpy as np
 
@@ -12,9 +12,6 @@ LOGGER = logging.getLogger("napari.async")
 
 # We convert slices to tuple for hashing.
 SliceTuple = Tuple[int, int, int]
-
-# The type of Layer.data
-LayerData = Union[ArrayLike, List[ArrayLike]]
 
 
 class ChunkKey:
@@ -123,12 +120,12 @@ class ChunkRequest:
             # we use the image as the thumbnail_source.
             return self.chunks.get('image')
 
-    def is_compatible(self, layer: LayerData) -> bool:
+    def is_compatible(self, layer: Layer) -> bool:
         """Return True if the given data is compatible with this request.
 
         Parameters
         ----------
-        data : LayerData
+        layer : Layer
             Compare this data to the data_id in the request.
         """
         return True  # Stub for now, will grow in the next version.
--- a/napari/components/chunk/_request.py
+++ b/napari/components/chunk/_request.py
@@ -7,11 +7,12 @@ import numpy as np
 
 from ...types import ArrayLike, Dict
 
-LOGGER = logging.getLogger("ChunkLoader")
+LOGGER = logging.getLogger("napari.async")
 
 # We convert slices to tuple for hashing.
 SliceTuple = Tuple[int, int, int]
 
+# The type of Layer.data
 LayerData = Union[ArrayLike, List[ArrayLike]]
 
 
@@ -51,23 +52,19 @@ class ChunkRequest:
 
     Parameters
     ----------
-    layer_id : int
-        Python id() for the Layer requesting the chunk.
-    data_id : int
-        Python id() for the Layer._data requesting the chunk.
-    indices
-        The tuple of slices index into the data.
-    array : ArrayLike
-        Load the data from this array.
+    key : ChunkKey
+        The key of the request.
+    chunks : Dict[str, ArrayLike]
+        The chunk arrays we need to load.
 
     Attributes
     ----------
-    layer_ref : weakref
-        Reference to the layer that submitted the request.
-    data_id : int
-        Python id() of the data in the layer.
+    key : ChunkKey
+        The key of the request.
+    chunks : Dict[str, ArrayLike]
+        The chunk arrays we need to load.
     load_seconds : float
-        Delay for this long during the load portion.
+        Delay for this long during the load portion, for testing.
     """
 
     def __init__(self, key: ChunkKey, chunks: Dict[str, ArrayLike]):
@@ -115,7 +112,7 @@ class ChunkRequest:
             # For single-scale we use the image as the thumbnail_source.
             return self.chunks.get('image')
 
-    def is_compatible(self, layer) -> bool:
+    def is_compatible(self, layer: LayerData) -> bool:
         """Return True if the given data is compatible with this request.
 
         Parameters
--- a/napari/components/chunk/_request.py
+++ b/napari/components/chunk/_request.py
@@ -1,4 +1,4 @@
-"""ChunkRequest is used to ask the ChunkLoader to load chunks.
+"""ChunkRequest is passed to ChunkLoader.load_chunks().
 """
 import logging
 from typing import List, Tuple, Union
@@ -18,14 +18,25 @@ LayerData = Union[ArrayLike, List[ArrayLike]]
 
 
 class ChunkKey:
-    """The key which a ChunkRequest will load.
+    """The key for one single ChunkRequest.
 
     Parameters
     ----------
     layer : Layer
         The layer to load data for.
-    indices : ?
-        The indices to load from the layer
+    indices : Indices
+        The indices to load from the layer.
+
+    Attributes
+    ----------
+    layer_id : int
+        The id of the layer making the request.
+    data_level : int
+        The level in the data (for multi-scale).
+    indices : Indices
+        The indices of the slice.
+    key : Tuple
+        The combined key, all the identifies together.
     """
 
     def __init__(self, layer: Layer, indices):
@@ -69,7 +80,7 @@ class ChunkRequest:
     """
 
     def __init__(self, key: ChunkKey, chunks: Dict[str, ArrayLike]):
-        # Make sure chunks is str->array as expected.
+        # Make sure chunks dict is what we expect.
         for chunk_key, array in chunks.items():
             assert isinstance(chunk_key, str)
             assert array is not None
@@ -77,7 +88,7 @@ class ChunkRequest:
         self.key = key
         self.chunks = chunks
 
-        # No delay by default, ChunkLoader.load_chunk() will set this if desired.
+        # No delay by default.
         self.load_seconds = 0
 
     def load_chunks(self):
@@ -86,31 +97,30 @@ class ChunkRequest:
             loaded_array = np.asarray(array)
             self.chunks[key] = loaded_array
 
-    def transpose_chunks(self, order):
+    def transpose_chunks(self, order: tuple) -> None:
         """Transpose all our chunks.
 
         Parameters
         ----------
         order
-            Transpose the chunks with this order.
+            Transpose the chunks into this order.
         """
         for key, array in self.chunks.items():
             self.chunks[key] = array.transpose(order)
 
     @property
     def image(self):
-        """The image chunk if we have one or None.
-        """
+        """The image chunk if we have one or None."""
         return self.chunks.get('image')
 
     @property
     def thumbnail_source(self):
-        """The chunk to use as the thumbnail_source or None.
-        """
+        """The chunk to use as the thumbnail_source or None."""
         try:
             return self.chunks['thumbnail_source']
         except KeyError:
-            # For single-scale we use the image as the thumbnail_source.
+            # No thumbnail_source so return the image instead. For single-scale
+            # we use the image as the thumbnail_source.
             return self.chunks.get('image')
 
     def is_compatible(self, layer: LayerData) -> bool:
@@ -121,7 +131,7 @@ class ChunkRequest:
         data : LayerData
             Compare this data to the data_id in the request.
         """
-        return True  # stub for now
+        return True  # Stub for now, will grow in the next version.
 
 
 def _index_to_tuple(index: Union[int, slice]) -> Union[int, SliceTuple]:
--- a/napari/components/chunk/_request.py
+++ b/napari/components/chunk/_request.py
@@ -82,24 +82,6 @@ class ChunkRequest:
         # No delay by default, ChunkLoader.load_chunk() will set this if desired.
         self.load_seconds = 0
 
-    @property
-    def num_chunks(self) -> int:
-        """Return the number of chunks in this request."""
-        return len(self.chunks)
-
-    @property
-    def num_bytes(self) -> int:
-        """Return the number of bytes that were loaded."""
-        return sum(array.nbytes for array in self.chunks.values())
-
-    @property
-    def in_memory(self) -> bool:
-        """True if all chunks are ndarrays."""
-        for array in self.chunks.values():
-            if not isinstance(array, np.ndarray):
-                return False
-        return True
-
     def load_chunks(self):
         """Load all of our chunks now in this thread."""
         for key, array in self.chunks.items():
--- a/napari/components/chunk/_request.py
+++ b/napari/components/chunk/_request.py
@@ -5,6 +5,7 @@ from typing import List, Tuple, Union
 
 import numpy as np
 
+from ...layers.base.base import Layer
 from ...types import ArrayLike, Dict
 
 LOGGER = logging.getLogger("napari.async")
@@ -27,7 +28,7 @@ class ChunkKey:
         The indices to load from the layer
     """
 
-    def __init__(self, layer, indices):
+    def __init__(self, layer: Layer, indices):
         self.layer_id = id(layer)
         self.data_level = layer._data_level
 
--- a/napari/components/chunk/_tests/test_chunk.py
+++ b/napari/components/chunk/_tests/test_chunk.py
@@ -59,5 +59,8 @@ def test_loader():
 
     assert np.all(request.image.data == request.chunks['image'].data)
 
+    # Since we didn't ask for a thumbnail_source it should be the image.
+    assert np.all(request.thumbnail_source.data[:] == request.image.data[:])
+
     with pytest.raises(KeyError):
-        request.chunks['this_array_does_not_exist']
+        request.chunks['missing_chunk_name']
--- a/napari/components/chunk/_tests/test_chunk.py
+++ b/napari/components/chunk/_tests/test_chunk.py
@@ -9,7 +9,7 @@ from .. import ChunkKey, chunk_loader
 
 def _create_layer() -> Image:
     """Return a small random Image layer."""
-    data = np.random.random((16, 16))
+    data = np.random.random((32, 16))
     return Image(data)
 
 
@@ -53,8 +53,11 @@ def test_loader():
     layer = _create_layer()
     key = ChunkKey(layer, (0, 0))
 
+    shape = (64, 32)
+    transpose_shape = (32, 64)
+
     # Just load one array.
-    data = np.random.random((64, 64))
+    data = np.random.random(shape)
     chunks = {'image': data}
 
     # Give data2 different data.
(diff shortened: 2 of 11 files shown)


pwinston:
I added a few tests just to get the balling rolling a bit. Much more more are needed. Not sure how to turn code coverage on for this branch?

pwinston:
I think this is ready for review @jni.

napari/components/chunk/_tests/test_chunk.py
```diff
@@ -0,0 +1,91 @@
+"""Tests for components.chunk."""
+
+import numpy as np
+import pytest
+
+from ....layers.image import Image
+from .. import ChunkKey, chunk_loader
```
jni: We use absolute imports in tests.

```suggestion
from napari.layers.image import Image
from napari.components.chunk import ChunkKey, chunk_loader
```

napari/components/chunk/_request.py
```diff
@@ -0,0 +1,151 @@
...
+    Attributes
+    ----------
+    layer_id : int
+        The id of the layer making the request.
+    data_level : int
+        The level in the data (for multi-scale).
+    indices : Tuple[Optional[slice], ...]
+        The indices of the slice.
+    key : Tuple
+        The combined key, all the identifies together.
```
jni:
```suggestion
        The combined key, all the identifiers together.
```

napari/components/chunk/_request.py
```diff
@@ -0,0 +1,151 @@
...
+
+import numpy as np
+
+from ...layers.base.base import Layer
+from ...types import ArrayLike, Dict
+
+LOGGER = logging.getLogger("napari.async")
+
+# We convert slices to tuple for hashing.
+SliceTuple = Tuple[int, int, int]
```
jni: These should be `Optional[int]`s, since any of them can be None:

```python
In [1]: x = slice(4, 8)

In [2]: x.step is None
Out[2]: True
```

napari/layers/image/_image_slice.py
```diff
@@ -1,54 +1,125 @@
...
-    Create with some default image:
+    def __init__(
+        self,
+        image: ArrayLike,
+        image_converter: ImageConverter,
+        rgb: bool = False,
+    ):
+        LOGGER.debug("ImageSlice.__init__")
+        self.image: ImageView = ImageView(image, image_converter)
+        self.thumbnail: ImageView = ImageView(image, image_converter)
```
jni, line 49: Reading this, I am wondering what the difference is between an ImageSlice and an ImageView...

napari/layers/image/_image_slice.py
```diff
@@ -1,54 +1,125 @@
...
         """
-        Create an ImageSlice with some default viewable image.
+        LOGGER.debug("ImageSlice.load_chunk: %s", request.key)
+
+        # Now "showing" this slice, even if it hasn't loaded yet.
+        self.current_key = request.key
+        self.loaded = False
+
+        # This will return a satisfied request in ChunkLoader is doing
+        # syncrhonous loading or the chunk was in the cache. If it returns
```
jni:
```suggestion
        # This will return a satisfied request if ChunkLoader is doing
        # synchronous loading or the chunk was in the cache. If it returns
```

napari/layers/image/_image_slice.py
jni: ... which upon further investigation reminds me of how I objected to having too many different objects back in #1343 but my objections were ignored, and now I am paying the price. =P

napari/layers/image/_image_slice.py
```diff
@@ -1,54 +1,125 @@
...
+        Has the data for this slice been loaded yet.
+    """
 
-    Examples
-    --------
-    Create with some default image:
+    def __init__(
+        self,
+        image: ArrayLike,
+        image_converter: ImageConverter,
```
jni: See #1355. Now is an excellent time to implement it.

napari/components/chunk/_request.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+class ChunkKey:
```
jni, line 17: I'm a little hesitant to have this class. After watching the video for the whole architecture, I thought, ok, maybe my dislike of new objects is a little too dogmatic, and this isn't so bad. But now upon reviewing this, I am finding myself confused by ImageSlice vs ImageView and ImageConverter. To me, the takeaway is that we should strive to use simple Python types as much as possible.

Now, in this case, the conversion of a tuple to a key is not *quite* trivial, so I'm willing to get some pushback on this, but I'd certainly like to moot the idea of using a single function for this, `chunk_to_key()`. It seems pretty similar ultimately, but I am newly entranced by the idea of losing an extra class definition.

napari/components/chunk/_request.py
```diff
@@ -0,0 +1,151 @@
...
+    def transpose_chunks(self, order: tuple) -> None:
+        """Transpose all our chunks.
+
+        Parameters
+        ----------
+        order
+            Transpose the chunks into this order.
+        """
+        for key, array in self.chunks.items():
+            self.chunks[key] = array.transpose(order)
```
jni, line 106: This is interesting — is it correct? I'm wondering whether the indices in key itself need transposing...?

napari/components/chunk/_request.py
```diff
@@ -0,0 +1,151 @@
...
+
+    def is_compatible(self, layer: Layer) -> bool:
+        """Return True if the given data is compatible with this request.
+
+        Parameters
+        ----------
+        layer : Layer
+            Compare this data to the data_id in the request.
+        """
+        return True  # Stub for now, will grow in the next version.
```
jni: I think this can be removed in this PR. I know it seems pointless to remove to only add later, and I'm sorry, but I think it's part of the idea of keeping this PR small and simple. That will make it clearer in the later PR why it needs adding.

napari/components/chunk/_request.py
```diff
@@ -0,0 +1,151 @@
...
+    key : ChunkKey
+        The key of the request.
+    chunks : Dict[str, ArrayLike]
+        The chunk arrays we need to load.
+
+    Attributes
+    ----------
+    key : ChunkKey
+        The key of the request.
+    chunks : Dict[str, ArrayLike]
```
jni, line 73: I'm *really* confused about why this is a dictionary?

napari/layers/image/image.py
```diff
@@ -546,27 +556,112 @@ def _set_view_slice(self):
...
+            # from the ChunkLoader because in ImageSlice.chunk_loaded we
+            # call request.thumbnail_source() and it knows to just use the
+            # image itself is there is no explicit thumbnail_source.
+            thumbnail_source = None
+
+        # Load our images, might be sync or async.
+        key = ChunkKey(self, image_indices)
+        self._load_images(key, image, thumbnail_source)
+
+    def _load_images(self, key: ChunkKey, image, thumbnail_source=None):
```
jni, line 575: What is a thumbnail source?

napari/layers/image/image.py
```diff
@@ -546,27 +556,112 @@ def _set_view_slice(self):
...
+
+            # For single-scale we don't request a separate thumbnail_source
+            # from the ChunkLoader because in ImageSlice.chunk_loaded we
+            # call request.thumbnail_source() and it knows to just use the
+            # image itself is there is no explicit thumbnail_source.
+            thumbnail_source = None
+
+        # Load our images, might be sync or async.
+        key = ChunkKey(self, image_indices)
+        self._load_images(key, image, thumbnail_source)
```
jni, line 573: Given my proposal above to drop ChunkKey, this would change to:

```python
self._load_images(chunk_to_key(self, image_indices), image, thumbnail_source)
```

I'm biased but I like it. ;)

napari/layers/image/image.py
```diff
@@ -546,27 +556,112 @@ def _set_view_slice(self):
             )
             indices[not_disp] = downsampled_indices
 
-            thumbnail_source = np.asarray(
-                self.data[self._thumbnail_level][tuple(indices)]
-            ).transpose(order)
+            thumbnail_source = self.data[self._thumbnail_level][tuple(indices)]
```
jni, line 559: Ah, so is the idea that ImageSlice is the slice, and ImageView has a possible transposition? Is that the only difference?

napari/layers/image/image.py
```diff
@@ -546,27 +556,112 @@ def _set_view_slice(self):
...
+
+        # Chunks get transposed after load.
+        request.transpose_chunks(self._get_order())
+
+        # The shape of the array after we loaded it was "wrong" in that it
+        # didn't match the rgb flag that the user specified (or we inferred?).
+        if self.rgb != guess_rgb(request.image.shape):
+            raise ValueError(
+                "Loaded chunk was the wrong shape, was rgb set correctly?"
+            )
```
jni, line 633: Why is this here? Seems unlikely to ever be needed...?

napari/layers/image/image.py
```diff
@@ -546,27 +556,112 @@ def _set_view_slice(self):
...
+            # The load is going to be async, announce that a load is
+            # underway, that we are no longer in the loaded state.
+            self.events.loaded()
         else:
-            self._slice.image.raw = image
-            self._slice.thumbnail.raw = thumbnail_source
+            # The load was sync, it's done, so use the chunk now.
+            self.chunk_loaded(satisfied_request, sync=True)
+
+    def chunk_loaded(self, request: ChunkRequest, sync: bool = False) -> None:
```
jni: Should this be maybe renamed `_on_chunk_loaded`? It looks like this is only ever called by `self`?

napari/layers/image/image.py
```diff
@@ -546,27 +556,112 @@ def _set_view_slice(self):
...
+        ----------
+        request : ChunkRequest
+            The request that was satisfied/loaded.
+        sync : bool
+            If True the chunk was loaded synchronously.
+        """
+        # Confirm this chunk is for us. We shouldn't be receiving any chunks
+        # that are not for us, but this takes no time so lets be sure.
+        if not request.is_compatible(self):
+            return  # Request was not for us, drop it.
```
jni: As mentioned above, I'd prefer to drop this from this PR.

napari/components/chunk/_request.py
jni: It looks like this contains just two keys, `image` and `thumbnail`. Is there a vision for more? It's funny because I'm always advocating for simple Python objects, and now here it really feels like these should be attributes rather than a dictionary.

napari/layers/image/image.py
```diff
@@ -546,27 +556,112 @@ def _set_view_slice(self):
...
+
+        if satisfied_request is None:
+            # The load is going to be async, announce that a load is
+            # underway, that we are no longer in the loaded state.
+            self.events.loaded()
         else:
-            self._slice.image.raw = image
-            self._slice.thumbnail.raw = thumbnail_source
+            # The load was sync, it's done, so use the chunk now.
+            self.chunk_loaded(satisfied_request, sync=True)
```
jni: Ok so reading this I'm wondering:

- can we replace `chunk_loader` with a simple `concurrent.futures.ThreadPoolExecutor()`
- instead of `self._slice.load_chunk(request)` we do `future = executor.submit(self._slice.load)`
- since we are doing sync for now, we immediately do `self._slice.set_data(future.result())`

Would this work, or am I missing some steps?

I'm not necessarily trying to get rid of the ChunkLoader, but rather, I want to defer it until it is actually needed. My suspicion is that we might end up with a simpler design if we simplify maximally in this PR. In particular, ChunkRequest is quite lightweight and I'm wondering whether we can do without it (same as ChunkKey) and have it be replaced by Futures having (key, Array) inputs and outputs.

jni (commented):
Thanks @pwinston! This is still a major PR but was reasonable to review. I'm wondering whether it can be minimised further, specifically:

- ChunkKey -> chunkid_to_key()
- ChunkRequest -> concurrent.futures.Future
- ChunkLoader -> Executor

There's more minor comments throughout but that's the gist.

jni:
I forgot to say: thanks for the hard work ripping all this out for review! I know it's extra work but I think everyone on the team will end up understanding the finished product much more clearly!

napari/layers/image/_image_slice.py
pwinston: From the comments an `ImageSlice` is "The slice of the image that we are currently viewing" while an `ImageView` is "A raw image and a viewable version of it.". Those are literally the first words in the doc strings of the two classes.

In napari's `Image` layer there are images and there are thumbnails of those images. So when you are viewing slice of an image, the `ImageSlice` is the image slice that you are viewing, and it has two attributes: `self.image` and `self.thumbnail`, both of which are of the type `ImageView`.

Each `ImageView` meanwhile has two properties `self.view` and `self.raw` both of which are `ArrayLike`. The raw one is the raw image and the view one is the viewable able, which was converted from raw to viewable using the `ImageConverter` (now called the `Callable[[ArrayLike], ArrayLike]` converter). 

So an `ImageSlice` contains two `ImageViews`, and each `ImageView` contains two images. This exact behavior and data existed in `Image` long before I made any changes, but it was all in the base class `Image`. Does that make sense?

napari/layers/image/_image_slice.py
pwinston: I removed `ImageConverter` in this PR, and commented at #1355.

napari/components/chunk/_request.py
pwinston: I think `ChunkKey` gains more functionality in the following PR's. This PR was extracted from #1510 and the stripped down. Although actually it does not gain much. Still I'd like to keep it as-is while we focus on splitting and fixing issues that are preventing merging.

If we change it now all the following PR's change. I don't totally disagree with this suggestion, but it seems more style and could be easily changed some other time once all the parts are in, if it in fact we still see it as a problem.

napari/components/chunk/_request.py
pwinston: It is required to transpose the images. This was always done. We can move the transposing to someone where, but it has to be done somewhere. Are you saying you want to move it to somewhere else? If you don't think it's required try commenting it out!

napari/components/chunk/_request.py
pwinston: I removed it but yes it will be immediately added back in the in the next PR.

napari/components/chunk/_request.py
pwinston: The `ChunkLoader` knows nothing about images. The `Image` layer currently asks the `ChunkLoader` to load 1 or 2 chunks. However other layers might load different chunks with different names, and other types of images might load more chunks with different names in one request.

napari/layers/image/image.py
pwinston: A thumbnail source is the image that's used to generate the thumbnail. This was in `Image` class long before I started, so nothing was changed.

napari/layers/image/image.py
pwinston: No, explained in comment above.

napari/layers/image/image.py
pwinston: I hit this in practice so I added this check. Without this check you get a very cryptic error.

napari/layers/image/image.py
pwinston: Maybe, but I wasn't user if the `_on_` prefix implied something to do with events? Also this is called externally by `QtChunkReceiver` in #1510 so it can't start with a leading underscore. But it could be `on_chunk_loaded` if `on` does not mean event related?

napari/layers/image/image.py
pwinston: I removed it.

napari/layers/image/image.py
pwinston: See #1510 for how the `ChunkLoader` grows. Your suggestion at the bottom of #1510 was "add the changes to Image/ImageSlice, calling out to a dummy `ChunkLoader` object/module that simply loads things synchronously" so that's what I've done here. I say we stick with the plan, since no new information has come to light since the plan was made.

napari/layers/image/image.py
pwinston: I think that's a valid way to do it, but that it's a question but it's a matter of style. This PR was split out from #1510. We could re-write this in a different style but then we'd have to redo every other PR that's part of #1510. I'd like to focus on splitting up #1510 and fixing the issues that are preventing #1510 from being merged first.

I think a style pass to remove OO would not be crazy at all. Especially if the Napari Style Guide discourages OO. But I think combining that style pass with simply splitting up the PRs would be mixing the two efforts, and would complicated the splitting.

napari/layers/image/_image_slice.py
pwinston: Fixed!

napari/components/chunk/_request.py
pwinston: Fixed, thanks.

napari/components/chunk/_request.py
pwinston: Fixed, thanks.

napari/components/chunk/_tests/test_chunk.py
pwinston: Fixed, is there a reason tests do it differently? Is this in the Napari Style Guide?

pwinston:
Addressed all @jni comments, fixed or replied to them. Thanks!

napari/layers/image/image.py
jni: @pwinston True! But the point of the plan is to be able to more clearly catch such simplifying steps in review, not to just end up with the exact same code at the end...

Right now, it looks to me like we might not need a lot of the machinery that you are setting up hook points for. Once the hooks are there, it will be tricky in the next PR to see what was necessary versus what was there for pre-review historical reasons.

On the other hand, if we go the simple route now, in the next PR you can go, "well, now we need a place to simultaneously hold the threadpoolexecutor and the queue, so that's the ChunkLoader". And I'll be like, well, fair enough. Or, maybe another simplification will occur to me in that context...!

So, I'd like to discuss whether this simplification makes sense, rather than whether it matches the exact plan we laid out.

At any rate, I'll refer to the "future version" of the chunkloader in #1510, good point, but that'll have to wait until tomorrow.

napari/layers/image/_image_slice.py
jni: Yeah, makes sense, thanks for the explainer. I'll need to think about things more — I feel like the class names are not descriptive enough. I'm old school and I think code should be readable without jumping back and forth between class definitions. =) That's a separate issue — I'll raise it.

napari/components/chunk/_request.py
jni: OHHHHHH Sorry — I was confusing `key` with ChunkKey, as opposed to image and thumbnail. 👍

napari/layers/image/image.py
pwinston: To put it another way, the reason it looks like the `ChunkLoader` is not doing much is your suggestion at the bottom of #1510 was explicitly "create a preliminary version of the `ChunkLoader` that's a stub, that does not do very much". So it's a stub (that could be replaced) only because the goal of the PR was to make a stub.

napari/components/chunk/_request.py
jni: As mentioned elsewhere, this PR should have a minimum of extraneous stuff. As it stands it's very close! Thanks for removing it.

napari/components/chunk/_request.py
jni: Could we roll this into the ChunkKey? (Whether or not we turn that into a function.) The extra level of indirection confused me repeatedly while reviewing this PR...

napari/layers/image/image.py
jni: @pwinston this method is new I believe? Either way it should have a proper NumPy docstring that describes what the arguments are.

napari/layers/image/image.py
jni: > Especially if the Napari Style Guide discourages OO.

Oooh I like this idea! The Juan Style Guide certainly does. Let's codify it. =P

@pwinston how are you envisioning the coming steps? If you are cherry-picking earlier commits, I think merge conflicts are going to start arising just from all the other suggested changes. If not, then I think the removal of ChunkKey at least would not be that additionally onerous to maintain, since it's a simple text replacement.

In general, though, I know a simple copy-paste for the breakdown is appealing, but the point of the breakdown into smaller PRs is to see along the way how we can simplify, and *only add complexity when and where we need it*. So although I acknowledge it might be more effort, I think we'll end up in a better place.

One idea: can you open the second PR against this PR (ie on your repo), so that I can judge for myself what I'm ok with keeping? =)

napari/layers/image/image.py
jni: Does this not mean that it should get checked in the image init? I agree that things should be caught early, but this is actually really not that early! 😂

napari/layers/image/image.py
jni: It generally *is* what's done on events, but it's also kind of what makes sense in English... This function is called when a chunk has been loaded... Right? Maybe `when_chunk_loaded`? `after_chunk_loaded`? `finalize_loaded_chunk`?

napari/layers/image/image.py
jni: Yup yup, I'm just saying, the stub might be even more minimal, to the point that it could be removed. Sometimes these things aren't clear at the outset!

jni:
@pwinston I think the best approach to resolve some of the discussions above is to open async-2 against async-1 in your own repo and ping me here. =) 

Also, please let reviewers mark conversations as resolved. Otherwise it is very hard for me to see your responses — and these are useful for other reviewers as well.

napari/components/chunk/_tests/test_chunk.py
jni: This will test the installed version, while in modules/packages you want to have a *consistent* version and that means relative imports. You could argue that you want to test a consistent version also, but generally tests are not necessarily in packages and relative imports are forbidden in non-packages.

pwinston:
Sorry, I thought the author was supposed to mark them resolved. So if the reviewer asks for a trivial change, and you verbatim make their change, you still let the reviewer mark it resolved? That seems like pushing work on to them? Is that a universal GitHub thing or a Napari thing? 

I think "should we remove this class" when it's not against any written guideline is best handled with a top-level issue for several reasons:

1. This issue is more visible.
2. These are exactly the types of issues that people need to see and learn from, since they are judgment calls.
3. The PR from resolving _that small issue_ will be incredibly bite-sized and targeted, a better more focused PR than even this PR (which is in fact big).
4. The change is visible in the record as a PR, as a focused change, so people can see that, it's not buried inside this PR.
3. Anyone can resolve that issue and by definition it's absolutely perfect for first timer, since it's small and focused and into existing functionality that works.

The idea that "this is our last opportunity to make changes to the code" I think is a bad idea to foster about reviews. And it's bad to promote the idea that only the author can make the changes, even if those changes are totally mechanical and easy, like a name change. I think the best time to address style questions is after some use of the new code. Like "I was adding async support to Surface, while the `ChunkLoader` work I think it this part was awkward, here's a PR to improve it."

I can create create `async-2` next week and maybe ping you next friday, so you can continue your review of `async-1` the following monday, which is a US Holiday, so the first we can merge `async-1` is September 2nd, which is a few days after my current contract is over. 

Or we can merge it now and you file a nice cogent top-level issue, or 2 or 3, with your suggestions, I start on async-2 this weekend and maybe get async-3 done next week as well. My preference is make the concrete changes now, and I made all the concrete changes you requested, and file _issues_ on the fuzzier _issues_, and move on to the next 4 PR's which are backed up.

Napari should, and does to my knowledge, have systems in place so that well-written issues will be addressed _just as reliably_ as buried review comments, and top-level issues are a better platform for visibility, learning, and enticing contributors into helping on juicy easy changes.

I would call focusing on the most important problems in the PR and filing issues on style/judgement concerns async reviewing. Which in the end will get you the same result much faster than synchronous reviewing, where people are blocked for weeks at a time waiting. I think async reviewing is particularly the right choice for an opt-in feature, soon to be called an `experimental` feature.

napari/layers/image/image.py
pwinston: This problem is new with async loading. With async loading we don't have the image data until it has been loaded. At that time it's possible to discover the user set the "rgb" flag incorrectly. If we don't explicitly check here, it leads to a nasty confusing error. So I added the check you see here.

That is I went through this experience:
1) I had a test case that produced a nasty confusing error.
2) I added this check and re-ran the test case.
3) I saw the error was now reasonable.
4) I fixed the test case.

I felt that was a good approach.

napari/layers/image/image.py
pwinston: I changed it to `on_chunk_loaded` which is definitely better than `chunk_loaded`. But maybe it could be confusing, I'm not sure.

napari/layers/image/image.py
pwinston: The stub could be removed. In fact that was my very first suggestion to you on the call. I said Juan, we don't even need the `ChunkLoader`, we can just modify the `Image` class to not call into anything, to do the sync load right there.

And you said no, Phil, we do need the `ChunkLoader` in the initial PR but make it a stub, it's just a stub in the first version that does sync loads. So in the initial PR that I split out of #1510 I created a stub `ChunkLoader` as we discussed as you asked for. I think it's fine, because it's all about to be stomped with the next PR, if we can finish thing PR.

Doing a 180 degree reversal now, just so that we can do a 2nd 180 degree reverse in three days, would be a colossal waste of time. That's the issue. As you can see in #1510 the ChunkLoader is there in every step. It's like 500 lines of code.

My strong advisement with splitting up #1510 is:
1) In each split PR find every serious problem and fix them **before merging**.
2) Style and design issues wait until the parts are merged, because they are parts of whole.
3) Everything is merged. It's opt-in, it's experimental, no one is using it.
4) File issues on every remaining concern, no matter how minor, and they will be addressed.

I feel you completely disagree with the above steps? Yet I don't understand what steps you are attempting to follow. It seems you simply don't believe that the code can ever be changed once it's in master? Is that the case? Because if that's true then it does change things.

sofroniewn:
> So if the reviewer asks for a trivial change, and you verbatim make their change, you still let the reviewer mark it resolved? That seems like pushing work on to them? Is that a universal GitHub thing or a Napari thing?

I resolve trivial, uncontroversial changes where I have very clearly done what was requested. If I have made a comment not wanting to do something, or done an alternative I leave those for the reviewer to resolve if they want. Sometimes we merge with those comments left open if, we decided to do something else, but want to leave that discussion as very easily discoverable.

> Or we can merge it now and you file a nice cogent top-level issue, or 2 or 3, with your suggestions, I start on async-2 this weekend and maybe get async-3 done next week as well. My preference is make the concrete changes now, and I made all the concrete changes you requested, and file issues on the fuzzier issues, and move on to the next 4 PR's which are backed up.

This seems like the right approach to me. I will do a quick pass through the code right now. @jni will be up soon too and can hopefully give one last green light.

Removing a class inside the chunk loader part of the code can easily be done later.

sofroniewn (approved):
I have just read through, focusing on the image layer code - which is the most connected to the rest of the code base, and so where changes will have the most impact on other PRs and efforts.

I also wonder long-term about `ChunkKey` as an object, and how and where it gets used, but I think its absolutely fine to get this in now and then continue to iterate and improve in subsequent PRs.

I'd like @jni to be able to have one final look at this in the morning when he wakes up, but then I'd like to merge so as to not leave @pwinston blocked.

jni:
Just to clarify: I'm not blocking anything right now, async-2 can and should be open right now against this branch. Then if we immediately merge this branch it is a simple matter of opening it against master rather than this branch.

pwinston:
> I'm not blocking anything right now.

Above you said you are blocking the merge of `async-1` until I make `async-2`. That's very definition of blocking. If you could explain what you meant by "I'm not blocking anything" and what you mean by "anything" in that sentence? It makes me wonder how much else is getting lost in translation.

The problem with your approach is then `async-1` is in limbo, you are asserting it's in limbo until I create `async-2`, and I have to create `async-2` not knowing whether `async-1` will require more changes. Then when you see `async-2` you want to go back to `async-1` and possibly make changes, which will then result in changes in `async-2`.  To me this totally subverts all of our motivations for splitting up the PR, and I really could not dream up a worse, slower or more unproductive way to review a big change.

The best approach in my eyes is merge the split PR's in series, in order. Review each one, fix all major issues, merge it. Review the next one, fix all major issues, merge it. Then with the experimental opt-in code in master I'm happy to entertain 100 issues and iterate on it endlessly.

It's like we have a 7 car pile up here. This is not a normal situation. We are trying to untangle the pile up. We came up with a plan of splitting up the PR. Let's follow through with that plan. Fix major issues while merging everything. Then come up for air and figure out what's next, whether re-work and re-design or fix the gross visuals problem or what.

I'm happy to keep iterating on async code, even just what's in #1510, as long as CZI wants to fund me. But at this point the most efficient way to use those funds is get this stuff merged then iterate. I think there's some huge disconnect here and I'm not sure what they right way is to get to the bottom of it.

jni:
@pwinston yeah, sorry about this, I totally understand how you would be reluctant to make async-2 until this is final. I was under the impression that you have async-2 in your back pocket, so you could grab it and use it to convince me that most of my concerns are unfounded.

Also and btw, sorry about the somewhat curt message — I was on my phone and merely wanted to get async-2 surfaced, rather than wait for a weekend cycle.

> Review each one, fix all major issues, merge it. 

For me, the object proliferation *is* a major issue. However, I can see that the whole sequence is going to be too hard to re-engineer. At least having the PRs separated will make review easier, and it will also make it easier to revisit these design choices after the whole train is merged. It's like a written record of your video. So 👍 on that for sure.

> > So if the reviewer asks for a trivial change, and you verbatim make their change, you still let the reviewer mark it resolved? That seems like pushing work on to them? Is that a universal GitHub thing or a Napari thing?
> 
> I resolve trivial, uncontroversial changes where I have very clearly done what was requested. If I have made a comment not wanting to do something, or done an alternative I leave those for the reviewer to resolve if they want. Sometimes we merge with those comments left open if, we decided to do something else, but want to leave that discussion as very easily discoverable.

👆 like that. We should probably codify this in our contributing guide. The reason it's not is historical — the bulk of it (based on scikit-image's) was written before the Resolve button existed!

> Napari should, and does to my knowledge, have systems in place so that well-written issues will be addressed just as reliably as buried review comments

Well, this is called technical debt as you know. Like real debt, a little is good and necessary, a lot means you can go bankrupt. You might be right that the balance is a little too conservative right now, I don't know. But if you look at other packages in the ecosystem, the intuition develops that the *likely* fate of an issue is that it won't get resolved, especially if it is not blocking critical functionality.

I *also* think that the PR is indeed the best time to fix code issues, because that is when the code is fresh in your mind. I'm pretty confident speculating that everyone reading this has had the experience of revisiting old code they wrote and thinking, now what the hell was I doing here...? 

Anyway, having said all of the above, I'm happy to move forward — let's take on that debt! I'll write up the ~~loan contracts~~ issues for the post-async world. 😊

napari/layers/image/image.py
jni: > I felt that was a good approach.

It is! I am just wondering whether there is an even better one lurking. For example, having the first request go through a slightly different code path than subsequent ones. Does that sound reasonable to you? It just feels super weird to me to have this check every time we get a request back. Is it worth me raising an issue? (Thoughts @sofroniewn?)

napari/layers/image/image.py
jni: Just to clarify: yes, I said something on the call based on partial understanding. Thanks to the actions we agreed on in the call, my understanding has improved (I think), which means I was now advocating a slightly different approach. There is no paradox.

> It seems you simply don't believe that the code can ever be changed once it's in master? Is that the case? Because if that's true then it does change things.

Well, code does have pretty massive inertia. It's not that it can never be changed, it's that it *might* never be changed. Resources are finite. Anyway, we'll see how we go! 🤞

pwinston:
I think if people feel the PR review process is the only opportunity to add value to the codebase, and if people feel that any issues they file are going to be ignored, no matter how valid the problem, then those are both pretty big problems. I don't have any quick answers. But it'd be interesting to compare notes and ideas.

sofroniewn:
Glad we got this merged @pwinston @jni, thanks for all the effort to get this in! I think it's onwards and upwards from here, with lots of lessons learnt and things to continue to think over and improve!!!

napari/layers/image/image.py
sofroniewn: >  It just feels super weird to me to have this check every time we get a request back. Is it worth me raising an issue? (Thoughts @sofroniewn?)

I agree, something feels funny here, but it could represent something deeper than rgb/ async that we want to consider even more now. We do have it so that you can change the dimensionality of the image data by modifying the data property, similarly the multiscale/ rgb too, but I'm not sure if we should allow that - i.e. once you create a layer should the dimensionality/ rgb/ multiscale be fixed, this could dramatically simplify lots of things, while being a reasonable restriction in functionality. In the absence of this restriction, I think the checks that @pwinston has here might be critical and we might need more for "dimensionality changes" too.

A dedicated issue to continue discussion seems the right approach @jni

d-v-b: What was the nasty confusing error? Maybe that could be made less confusing.

pwinston: I don't remember the error, it was a few months ago. But it was not ours. Just something from numpy about brocasting dimensions or mis-aligned shapes. Something that gave zero indication it had anything to do the with the fact you lied about whether your image was rgb or not. Such as you said the image was not rgb, but it was, or vice verse. Without async that's found out immediately, with async it's not. I agree a top-level issue might be the best way to sort it out. Perhaps someone else can take an interest and come up with a better solution.

pwinston: I suspect there are a lot more issues like this that we won't find until we get heavy testing, like people using it all day for routine work. Something we are aiming towards. I feel like my usage is VERY gingerly, I'm just focusing on the happy path, seeing the async logic work, and not on all the many edge cases.

jni: Raised in #1576. I won't resolve this so that the reference to this discussion loads correctly.

PR #1567, opened 2020-08-21
Remove silence overwritte during screenshot

Czaki:
# Description

During take screenshot, when extension is missed then existing file could be silent overwritte. 

## Type of change

# References

# How has this been tested?

- [x] manually 
- [x] all test passed
## Final checklist:

### DIFF
--- a/napari/_qt/dialogs/screenshot_dialog.py
+++ b/napari/_qt/dialogs/screenshot_dialog.py
@@ -1,0 +1,46 @@
+import os
+from pathlib import Path
+from typing import Any, Callable
+
+from qtpy.QtWidgets import QFileDialog, QMessageBox
+
+
+class ScreenshotDialog(QFileDialog):
+    def __init__(
+        self,
+        save_function: Callable[[str], Any],
+        parent=None,
+        directory=str(Path.home()),
+        qt_dialog=False,
+    ):
+        super().__init__(parent, "Save screenshot")
+        self.setAcceptMode(QFileDialog.AcceptSave)
+        self.setFileMode(QFileDialog.AnyFile)
+        self.setNameFilter("Image files (*.png *.bmp *.gif *.tif *.tiff)")
+        if qt_dialog:
+            self.setDefaultSuffix(".png")
+        self.setOption(QFileDialog.DontUseNativeDialog, qt_dialog)
+        self.setDirectory(directory)
+
+        self.save_function = save_function
+
+    def accept(self):
+        print("aaA", self.selectedFiles())
+        print(self.selectedUrls())
+        save_path = self.selectedFiles()[0]
+        if not (self.options() & QFileDialog.DontUseNativeDialog):
+            if os.path.splitext(save_path)[1] == "":
+                save_path = save_path + ".png"
+                if os.path.exists(save_path):
+                    res = QMessageBox().warning(
+                        self,
+                        "Confirm overwrite",
+                        f"{save_path} already exists. Do you want to replace it?",
+                        QMessageBox.Yes | QMessageBox.No,
+                        QMessageBox.No,
+                    )
+                    if res != QMessageBox.Yes:
+                        # standard accept return 1, reject 0. This inform that dialog should be reopened
+                        return self.done(2)
+        self.save_function(save_path)
+        return super().accept()
--- a/napari/_qt/dialogs/screenshot_dialog.py
+++ b/napari/_qt/dialogs/screenshot_dialog.py
@@ -6,42 +6,49 @@ from qtpy.QtWidgets import QFileDialog, QMessageBox
 
 
 class ScreenshotDialog(QFileDialog):
+    """
+    Dialog to chose save location of screenshot.
+
+    Parameters
+    ----------
+    save_function : Callable[[str], Any],
+        Function to be called on success of selecting save location
+     parent : QWidget, optional
+        Optional parent widget for this widget..
+    directory : str, optional
+        Starting directory to be set to File Dialog
+
+    """
+
     def __init__(
         self,
         save_function: Callable[[str], Any],
         parent=None,
         directory=str(Path.home()),
-        qt_dialog=False,
     ):
         super().__init__(parent, "Save screenshot")
         self.setAcceptMode(QFileDialog.AcceptSave)
         self.setFileMode(QFileDialog.AnyFile)
         self.setNameFilter("Image files (*.png *.bmp *.gif *.tif *.tiff)")
-        if qt_dialog:
-            self.setDefaultSuffix(".png")
-        self.setOption(QFileDialog.DontUseNativeDialog, qt_dialog)
         self.setDirectory(directory)
 
         self.save_function = save_function
 
     def accept(self):
-        print("aaA", self.selectedFiles())
-        print(self.selectedUrls())
         save_path = self.selectedFiles()[0]
-        if not (self.options() & QFileDialog.DontUseNativeDialog):
-            if os.path.splitext(save_path)[1] == "":
-                save_path = save_path + ".png"
-                if os.path.exists(save_path):
-                    res = QMessageBox().warning(
-                        self,
-                        "Confirm overwrite",
-                        f"{save_path} already exists. Do you want to replace it?",
-                        QMessageBox.Yes | QMessageBox.No,
-                        QMessageBox.No,
-                    )
-                    if res != QMessageBox.Yes:
-                        # standard accept return 1, reject 0. This inform that dialog should be reopened
-                        super().accept()
-                        self.exec_()
+        if os.path.splitext(save_path)[1] == "":
+            save_path = save_path + ".png"
+            if os.path.exists(save_path):
+                res = QMessageBox().warning(
+                    self,
+                    "Confirm overwrite",
+                    f"{save_path} already exists. Do you want to replace it?",
+                    QMessageBox.Yes | QMessageBox.No,
+                    QMessageBox.No,
+                )
+                if res != QMessageBox.Yes:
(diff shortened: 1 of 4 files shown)


sofroniewn:
Thanks for this @Czaki - this is looking good I wonder though if we can take this opportunity to simplify both of these `_screenshot_dialog` methods by creating out own `QtSaveScreenshotDialog` and put it in here, https://github.com/napari/napari/tree/master/napari/_qt/dialogs, and then reduce the amount of duplicated code and logic? What are your thoughts on that?

Czaki:
Hm. There is Question if it should be class for save screenshot, or more general case?

Switch to using instance of `QFileDialog` is possibility to use `setDefaultSuffix` method which allow remove whole logic of adding suffix. 

And what You think about store history of last visited locations instead of only last one?

sofroniewn:
> Hm. There is Question if it should be class for save screenshot, or more general case?

I think for now I'd just focus on reducing the duplicate code between `QtMainWindow._screenshot_dialog` and `QtViewer._screenshot_dialog`, but there could be other places where we could reduce duplicate code in the future.

> Switch to using instance of QFileDialog is possibility to use setDefaultSuffix method which allow remove whole logic of adding suffix.

I remember us trying a bunch of stuff with that, but we could never get it working the way we wanted, see a long discussion with @GenevieveBuckley and I trying to do that here https://github.com/napari/napari/pull/944#discussion_r375638126, so I think I'd rather stick with the logic we have and just fix the "overwrite" bug you observed rather than try that again.

> And what You think about store history of last visited locations instead of only last one?

I'm open to that, I think that is fairly standard these days.

Czaki:
In last commit there is my proposition for hide everything in dialog. In PartSeg I give up from using native dialog. Th Qt one is much simpler to control.  

If it looks ok I will think about tests.

napari/_qt/qt_main_window.py
```diff
@@ -474,23 +475,20 @@ def _help_changed(self, event):
...
-        if (filename != '') and (filename is not None):
-            # double check that an appropriate extension has been added as the
-            # filter option does not always add an extension on linux and windows
-            # see https://bugreports.qt.io/browse/QTBUG-27186
-            image_extensions = ('.bmp', '.gif', '.png', '.tif', '.tiff')
-            if not filename.endswith(image_extensions):
-                filename = filename + '.png'
-            self.screenshot(path=filename)
+        while True:
+            res = dial.exec_()
```
sofroniewn: This may be personal preference, but do we have to use the `while True` and `res = dial.exec_()` syntax? Can we make the dialog modal, (blocking) and then show it, and catch the return value only when it is closed?

napari/_qt/qt_main_window.py
```diff
@@ -474,23 +475,20 @@ def _help_changed(self, event):
         """
         self._help.setText(event.text)
 
-    def _screenshot_dialog(self):
+    def _screenshot_dialog(self, *, default_name=""):
```
sofroniewn: I think you can drop the `default_name` now

napari/_qt/dialogs/screenshot_dialog.py
```diff
@@ -0,0 +1,46 @@
...
+                        "Confirm overwrite",
+                        f"{save_path} already exists. Do you want to replace it?",
+                        QMessageBox.Yes | QMessageBox.No,
+                        QMessageBox.No,
+                    )
+                    if res != QMessageBox.Yes:
+                        # standard accept return 1, reject 0. This inform that dialog should be reopened
+                        return self.done(2)
+        self.save_function(save_path)
+        return super().accept()
```
sofroniewn, line 54: How about returning `os.path.dirname(self.selectedFiles()[0])` if successful and `None` otherwise, so that the return can always just `self._last_visited_dir = result`?

napari/_qt/qt_main_window.py
```diff
@@ -474,23 +475,20 @@ def _help_changed(self, event):
...
-            self.screenshot(path=filename)
+        while True:
+            res = dial.exec_()
+            print(res, ScreenshotDialog.Accepted, ScreenshotDialog.Rejected)
+            if res == ScreenshotDialog.Accepted:
+                self._last_visited_dir = os.path.dirname(
+                    dial.selectedFiles()[0]
+                )
+            if res in (ScreenshotDialog.Accepted, ScreenshotDialog.Rejected):
+                break
```
sofroniewn: See my comment above about the dialog just returning the right thing so we don't need to duplicate this either. So instead of returning Accepted or Rejected, you just get the directory or None and update `self._last_visited_dir` regardless

sofroniewn (commented):
I like the direction this is going @Czaki. I think having one dialog is simpler, but I'm wondering if we can move a way from the `while` syntax and make the dialog return type be the exact thing we want, rather than the Accepted/ Rejected, which will simplify things even more.

You also have a couple of print statements still in there :-)

Czaki:
The whole while is obsolete if ther wil be decision to use qt dialog instead of native one. 
I do not found way to prevent close native dialog when select accept. I could hide while in additional functionbut it not looks like thing you expect. You could simple compare both dialog with change value of `qt_dialog` in constructor.

sofroniewn:
> The whole while is obsolete if ther wil be decision to use qt dialog instead of native one.
I do not found way to prevent close native dialog when select accept.

Hmm ok, I think we still have a strong preference towards using the native. I can look more into this later tonight

Czaki:
Recursion may be an answer. 
I have hope that it does not produce multiple save dialog on some system.

napari/_qt/dialogs/screenshot_dialog.py
```diff
@@ -0,0 +1,47 @@
...
+        if qt_dialog:
+            self.setDefaultSuffix(".png")
+        self.setOption(QFileDialog.DontUseNativeDialog, qt_dialog)
+        self.setDirectory(directory)
+
+        self.save_function = save_function
+
+    def accept(self):
+        print("aaA", self.selectedFiles())
+        print(self.selectedUrls())
```
sofroniewn:
```suggestion
```
these can be cut

napari/_qt/qt_main_window.py
sofroniewn: @Czaki thoughts?

napari/_qt/dialogs/screenshot_dialog.py
sofroniewn: @Czaki thoughts? fine to not do this if it breaks a pattern for dialogs/ causes other problems

napari/_qt/qt_viewer.py
```diff
@@ -367,23 +369,11 @@ def screenshot(self, path=None):
             imsave(path, QImg2array(img))  # scikit-image imsave method
         return QImg2array(img)
 
-    def _screenshot_dialog(self):
+    def _screenshot_dialog(self, *, default_name=""):
```
sofroniewn:
```suggestion
    def _screenshot_dialog(self):
```
`deafult_name` can be cut

sofroniewn (commented):
@Czaki i've added a couple comments, maybe we can sync on this tomorrow and get this merged, will be nice to have it in for 0.3.8 too.

Thanks for being patient, I think it is close!!

napari/_qt/dialogs/screenshot_dialog.py
```diff
@@ -0,0 +1,47 @@
...
+        directory=str(Path.home()),
+        qt_dialog=False,
+    ):
+        super().__init__(parent, "Save screenshot")
+        self.setAcceptMode(QFileDialog.AcceptSave)
+        self.setFileMode(QFileDialog.AnyFile)
+        self.setNameFilter("Image files (*.png *.bmp *.gif *.tif *.tiff)")
+        if qt_dialog:
+            self.setDefaultSuffix(".png")
+        self.setOption(QFileDialog.DontUseNativeDialog, qt_dialog)
```
sofroniewn: @Czaki do you actively plan to use the non-native dialogs? If not I don't think we need to provide this option

napari/_qt/dialogs/screenshot_dialog.py
```diff
@@ -0,0 +1,47 @@
+import os
+from pathlib import Path
+from typing import Any, Callable
+
+from qtpy.QtWidgets import QFileDialog, QMessageBox
+
+
+class ScreenshotDialog(QFileDialog):
```
sofroniewn, line 8: Would be great to get a doc string on this class and the input parameters

napari/_qt/dialogs/screenshot_dialog.py
sofroniewn: i.e. we could drop the `qt_dialog` arg

napari/_qt/dialogs/screenshot_dialog.py
Czaki: added

napari/_qt/dialogs/screenshot_dialog.py
Czaki: https://doc.qt.io/qt-5/qdialog.html#exec
`exec_` should return int. 

But setting `_last_visited_dir` could be done outside if clause

napari/_qt/qt_main_window.py
Czaki: removed

napari/_qt/qt_main_window.py
Czaki: recursion done

sofroniewn (approved):
Thanks for those fixes @Czaki I have given this a try and everything works fine and code looks good to go. Thanks for having the test too!!

PR #1577, opened 2020-08-24
Status bar value bug for automatically down-sampled images

### LINKED ISSUE #1557: Wrong intensity values in status bar for automatically down-sampled images.
## 🐛 Bug

When displaying a 2D image larger than 'GL_MAX_TEXTURE_SIZE', it is automatically downsampled. While the coordinate values displayed in the status bar are correct, the intensity values that are shown don't take the downsampling into account.

## To Reproduce

1. Create a 2D image larger than 'GL_MAX_TEXTURE_SIZE'.
2. Hover over the image and notice that the intensity values displayed in the status bar correspond to a 'larger' image than the one displayed (i. e. are not taking the downsampling into account).

```Python
gl_max_texture_size = 16384 # insert hardware specific gl_max_texture_size
img = np.zeros((gl_max_texture_size*2, gl_max_texture_size*2), dtype=np.uint16) # create an empty image twice as large
img[img.shape[0]//2:, img.shape[1]//2:] = 1 # set the bottom right corner to 1

with napari.gui_qt():
    viewer = napari.Viewer()
    viewer.add_image(img, colormap='viridis')
```
## Expected behavior

Intensity values displayed in the status bar should correspond to the image on display.

## Environment

napari: 0.3.6
Qt: 5.14.2
PyQt5: 5.14.2
NumPy: 1.18.3
SciPy: 1.4.1
Dask: 2.15.0
VisPy: 0.6.4

GL version: 4.6.0 - Build 26.20.100.7812
MAX_TEXTURE_SIZE: 16384

Plugins:
- napari-plugin-engine: 0.1.5
- napari_bug
- svg: 0.1.3

## Additional context

While creating a reproducible example I noticed that if I use `dtype=np.uint8` with that same example there seems to be a display bug _in addition_ to the bug in the status bar (i. e. the whole image is displayed as 0).

MaksHess:
# Description
Fix the status bar bug described in #1557 . The initial fix lead to a test fail in `_vispy/_tests/test_vispy_multiscale` which led me to the current fix that falls back to the old behavior in case of a multiscale layer. It might be worthwhile to overthink the whole coordinate system for downsampled/multiscale images to have a unified concept that works in both cases.

## Type of change

# References
Closes #1557 

# How has this been tested?

- [ ] `test_statusbar.py` has been added to `_vispy/_tests` covering test cases with small and large images (ie images that need to be downsampled)
- [ ] running the full test suite via `pytest napari` without adding `test_statusbar.py` leads to 0 failed tests
- [ ] adding `test_statusbar.py` leads to 2 failed tests:
```
FAILED napari/_tests/test_viewer.py::test_no_qt_loop - Failed: DID NOT RAISE <class 'RuntimeError'>
FAILED napari/_qt/_tests/test_exception_handler.py::test_exception_handler_gui - assert <napari._qt.qt_error_notification.NapariNotification object at 0x000001F4CA479EE8> in []
```

## Final checklist:


sofroniewn:
Hi @MaksHess thanks for submitting this bug fix! It looks like something might be off with the Flake8 and Black formatting on the PR, were you able to install our pre-commit hooks? See detail in our [contributing guide](https://napari.org/docs/developers/CONTRIBUTING.html#setting-up-a-development-installation)

Also it looks like the creation of the canvas is causing problems for on one of our other tests, probably because the canvas isn't being closed. Can you instead use out [get_max_texture_sizes ](https://github.com/napari/napari/blob/5e8be0987b3ab691da5755380ce1dd798978b1c6/napari/_vispy/vispy_base_layer.py#L252) method which should handle all of that for you.

Looking at this more closely, I think you actually only want to apply the `layer._transforms['tile2data']` transform, not the `.simplified` transform (and possibly the inverse), and you'll only want to apply it in the non-multiscale case as the downsampling only happens in the non-multiscale case, see here for the setting of the additional transform that you want to undo https://github.com/napari/napari/blob/5e8be0987b3ab691da5755380ce1dd798978b1c6/napari/_vispy/vispy_image_layer.py#L170.

Stepping back all of this will get revisited and fixed as we improve both the global coordinate system work and the octree work for multiscale rendering, but this should be a minimal fix for now.

MaksHess:
Hi @sofroniewn,

sorry for the delay! Installing the pre-commit hooks seems to have worked, also I'm using the `self.layer._transforms['tile2data']` transforms in the code and test.

The usage of `get_max_texture_sizes` does not seem to solve the problem with the two test failing when I run the tests locally though. Any ideas what might be the problem?

sofroniewn:
Hi @MaksHess, sorry for not getting back to you on this one earlier. It ended up requiring a bit more of an overhaul to our coordinate system work that came in #1673, which I think fixes this issue. There was an additional multiscale coordinate value bug which I flagged in #1691, but I think it will be easier to work off of this PR.

I just updated this PR with the some more code, but I might need to merge master in too to get all the latest here

sofroniewn:
Ok, I've update this PR now to include the latest changes and fixed the tests. Note some of the fix here is also for a multiscale bug that was going to be fixed in #1691, but it was actually easier just to fix here.

You can see the current behaviour now for a downsampled image:

downsample_coords

@MaksHess please let us know if this doesn't fix your use case, and thanks for your contribution that got this started! I hope you don't mind me finishing this up!! I will merge once tests pass.

sofroniewn (approved):
This looks goods to me, all tests are now passing. Thanks for adding the special test for this @MaksHess!

MaksHess:
Hi @sofroniewn 
No worries, thanks for taking care of this!
I tested it for my use case and it works like a charm :)

PR #1580, opened 2020-08-26
Longer spinbox

Czaki:
# Description

With current width, on ubuntu it looks 
image
instead of 
image
So user cannot read value inside it. 

In Qt stylesheet is more important than object property so `setMinimumWidth` does not work 

## Type of change

# References

# How has this been tested?

- [x] manually

## Final checklist:

### DIFF
--- a/napari/resources/styles/00_base.qss
+++ b/napari/resources/styles/00_base.qss
@@ -148,7 +148,7 @@ QAbstractSpinBox {
   background-color: {{ foreground }};
   border: none;
   padding: 1px 10px;
-  min-width: 40px;
+  min-width: 70px;
   min-height: 18px;
   border-radius: 2px;
 }


sofroniewn:
@Czaki I know we use spin-boxes in the current vectors layer and the play button, can you see if this change effects how those two items are rendered?

sofroniewn:
I just tried this out with `python examples/add_vectors_color_by_angle.py` and it looks good
Screen Shot 2020-08-26 at 8 53 42 AM

I also tried it with `python examples/nD_image.py` and clicking the play button and it looks good too
Screen Shot 2020-08-26 at 8 54 21 AM

Those spin-boxes look longer/ might have custom lengths and so not be affected by this change

sofroniewn (approved):
This is fine with me, I have checked on a couple of our internal use cases (see above screenshots) and behavior there is unchanged.

@tlambert03 do you have any concern about this for widgets provided by magicgui?

I will let sit for a little, but if no-one objects I will merge to get this in for our 0.3.7 release

tlambert03:
> @tlambert03 do you have any concern about this for widgets provided by magicgui?

Nope, fine by me

PR #1581, opened 2020-08-26
Make layer dims private

sofroniewn:
# Description
This is a follow on from #1539, which makes the layer dims private i.e. `layer.dims` -> `layer._dims`. We pulled this out of #1539 to make the PR easier to parse, but it is an important change as users should no longer modify the `layer.dims` properties, see comment here https://github.com/napari/napari/pull/1539#issuecomment-680581123, and this moves us towards our goal of removing all the slice information off the layer, as discussed in #1353. Merging this now will make that transition smoother in the future.

## Type of change


napari/_vispy/vispy_points_layer.py
```diff
@@ -42,7 +42,7 @@ def _on_data_change(self, event=None):
         # reversed to make the most recently added point appear on top
         # and the rows / columns need to be switch for vispys x / y ordering
         if len(self.layer._indices_view) == 0:
-            data = np.zeros((1, self.layer.dims.ndisplay))
+            data = np.zeros((1, self.layer._dims.ndisplay))
```
jni, line 45: @sofroniewn for me, it's a bit of a bad smell that the vispy class needs to access a private attribute of the layer... Is this something that is supposed to be fixed in future PRs?

jni (commented):
@sofroniewn this is good. I see that almost all changes are from `self.dims` to `self._dims`, which is how it should be. However, there's a few `layer.dims` -> `layer._dims` (outside of tests), which I don't think is ideal. Is there a plan for how this will work in the future, so that object consumers are not having to access private attributes?

napari/_vispy/vispy_points_layer.py
sofroniewn: yeah, i could actually work on eliminating the references in vispy in this PR. i'll give that a go tomorrow

napari/layers/base/base.py
```diff
@@ -547,7 +547,7 @@ def thumbnail(self, thumbnail):
     @property
     def ndim(self):
         """int: Number of dimensions in the data."""
-        return self.dims.ndim
+        return self._dims.ndim
```
Czaki, line 687:
```suggestion

@property
def dims()
        warnings.warn(
            "dims parameter is deprecated and will be removed in version 0.4.1",
            DeprecationWarning
        )
        return self._dims

```

This suggestion does not contain needed imports.

sofroniewn: Ooo, I like this idea. Thanks for the suggestion @Czaki. Thoughts @jni? Maybe we get this in before `0.4.0`

jni: This would be great, thanks @Czaki for the suggestion.

sofroniewn: done

sofroniewn:
Ok, I've update this PR to now include the deprecation warning as suggested by @Czaki. I think that is quite nice, and given that we know that in the long term layers won't contain slice information it makes sense.

I guess I could update the deprecation warning to explain that and tell people to use `viewer.dims`. Let me know @jni

Otherwise this is ready for another round of review @jni @Czaki

jni:
@sofroniewn "fix tracks"??? 🤔 

Anyway 😂 I do think the deprecation warning should point users to an alternative. Could you add that?

sofroniewn:
> @sofroniewn "fix tracks"??? 🤔

yes, i was surprised i need to do that, i would have thought my dims property would have let the tests pass ... any ideas?

I'll add comment now

sofroniewn:
> Anyway 😂 I do think the deprecation warning should point users to an alternative. Could you add that?

done

jni:
Oh! What was the failure? I got confused because you added the deprecated attribute and the parameter rename in the same commit. I agree with you that it's weird that it fails. Perhaps our pytest config turns (internal) warnings into failures, which is fair enough, it would be annoying to get this warning just for using a tracks layer!

sofroniewn:
> it would be annoying to get this warning just for using a tracks layer!

Agreed! It's fixed now :-)

sofroniewn:
> Oh! What was the failure?

failure was actually an access one ...

napari/_vispy/vispy_points_layer.py
jni: @sofroniewn could you have another look at this before merging?

jni (commented):
@sofroniewn this is ready from my perspective except for the fact that the vispy layers now have to access a private attribute... What is your current thinking on this?

sofroniewn:
> @sofroniewn this is ready from my perspective except for the fact that the vispy layers now have to access a private attribute... What is your current thinking on this?

So we've actually been rather terrible about this already, and while in a sense this makes it worse, in this case it's very much as a temporary thing (see other places in the file where we access `_` methods already like https://github.com/napari/napari/blob/v0.3.8/napari/_vispy/vispy_image_layer.py#L75).

There's a longer post 0.4.0 convo I want to have about how we can fix, possibly by having slicing generate a new Image layer with the slice of the data and all the needed public properties, but that's for later!

jni (approved):

PR #1582, opened 2020-08-26
Add support for deprecate signal in EmitterGroup

Czaki:
# Description

This PR  allow to provide alias for changed/removed signal in emitter group to not break external scripts without warnings. 

## Type of change

# References

# How has this been tested?

- [X] example: the test suite for my feature covers cases x, y, and z
- [X] example: all tests pass with my change

## Final checklist:

### DIFF
--- a/napari/utils/events/_tests/test_evented_list.py
+++ b/napari/utils/events/_tests/test_evented_list.py
@@ -366,3 +366,17 @@ def test_evented_list_subclass():
     assert hasattr(lst, 'events')
     assert 'boom' in lst.events.emitters
     assert lst == [1, 2]
+
+
+def test_event_group_depr(capsys):
+    events = EmitterGroup(b=None, deprecated={"a": "b"})
+    assert events.b == events.a
+    captured = capsys.readouterr()
+    assert captured.err == "emitter a is deprecated, b provided instead\n"
+    with pytest.raises(AttributeError):
+        events.c.connect()
+    assert events["b"] == events["a"]
+    captured = capsys.readouterr()
+    assert captured.err == "emitter a is deprecated, b provided instead\n"
+    with pytest.raises(KeyError):
+        events["c"].connect()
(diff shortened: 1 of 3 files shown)


sofroniewn:
Thanks for the PR @Czaki, I will take a look promptly. See [discussion in zulip](https://napari.zulipchat.com/#narrow/stream/212875-general/topic/deprecate_signal_support/near/208114097) for context. cc @jni

napari/utils/events/event.py
```diff
@@ -690,6 +702,12 @@ def __getitem__(self, name):
         Note that emitters may also be retrieved as an attribute of the
         EmitterGroup.
         """
+        if name in self._deprecated:
+            print(
+                f"emitter {name} is deprecated, {self._deprecated[name]} provided instead",
+                file=sys.stderr,
+            )
```
jni: These should use `warnings.warn(message, category=FutureWarning)`, rather than print.

napari/utils/events/_tests/test_evented_list.py
```diff
@@ -366,3 +366,17 @@ class B(A, EventedList):
     assert hasattr(lst, 'events')
     assert 'boom' in lst.events.emitters
     assert lst == [1, 2]
+
+
+def test_event_group_depr(capsys):
+    events = EmitterGroup(b=None, deprecated={"a": "b"})
+    assert events.b == events.a
+    captured = capsys.readouterr()
+    assert captured.err == "emitter a is deprecated, b provided instead\n"
```
jni: As mentioned below, rather than printing to stderr, we should use warnings, and use `with pytest.warns(FutureWarning):`, see here: https://docs.pytest.org/en/stable/warnings.html#warns

jni (changes requested):
@Czaki awesome, very neat idea and cleanly implemented. My only request is that we use `warnings.warn` rather than `print`, as this comes with much more control for library users. I've linked below to the pytest documentation for asserting warnings.

jni (approved):
🎉 👏

PR #1583, opened 2020-08-26
async-2: ChunkLoader Thread Pool

pwinston:
# Description
* New `NAPARI_ASYNC` env var to toggle on async loading via a thread pool.

### DIFF
--- a/napari/_qt/qt_chunk_receiver.py
+++ b/napari/_qt/qt_chunk_receiver.py
@@ -14,9 +14,9 @@ class QtGuiEvent(QObject):
     """Fires an event in the GUI thread.
 
     Listens to an event in any thread. When that event fires, it uses a Qt
-    Signal/Slot to fire a matching even in the GUI thread. If the original
-    event is already in the GUI thread that's fine, the matching event will
-    be immediately fired in the GUI threads.
+    Signal/Slot to fire a gui_event in the GUI thread. If the original
+    event is already in the GUI thread that's fine, the gui_event will
+    be immediately fired the GUI thread.
     """
 
     signal = Signal(Event)
--- a/napari/_qt/qt_chunk_receiver.py
+++ b/napari/_qt/qt_chunk_receiver.py
@@ -31,6 +31,16 @@ class QtGuiEvent(QObject):
         The event we are listening to.
     events : EmitterGroup
         The only event we report is events.gui_event.
+
+    Notes
+    -----
+    Qt's signal/slot is the only way we know of to "call" from a worker
+    thread to the GUI thread. When Qt signals from a worker thread it posts
+    an event. When the GUI thread is next processing messages it will call
+    into the Slot to deliver that event.
+
+    If the original event was already in the GUI thread that's fine,
+    the resulting event will just be triggered right away.
     """
 
     signal = Signal(Event)
@@ -73,6 +83,29 @@ class QtChunkReceiver:
     ----------
     gui_event : QtGuiEvent
         We use this to call _on_chunk_loaded_gui() in the GUI thread.
+
+    Notes
+    -----
+    ChunkLoader._done "may" be called in a worker thread. The
+    concurrent.futures documentation only guarantees that the future's done
+    handler will be called in a thread in the correct process, it does not
+    say which thread.
+
+    We need to call Layer.on_chunk_loaded() to deliver the loaded chunk to the
+    Layer. We do not want to make this call from a worker thread, because our
+    model code is not thread safe. We don't want the GUI thread and the worker
+    thread changing things at the same time, both triggering events, potentially
+    calling into vispy or other things that also aren't thread safe.
+
+    We could add locks, but it's simpler and better if we just call
+    Layer.on_chunk_loaded() from the GUI thread. This class QtChunkReceiver
+    listens to the ChunkLoader's chunk_loaded event. It then uses QtUiEvent
+    to call its own _on_chunk_loaded_gui() in the GUI thread. From that
+    method it can safely call Layer.on_chunk_loaded.
+
+    If ChunkLoader's chunk_loaded event is already in the GUI thread for
+    some reason, this class will still work fine, it will just run
+    100% in the GUI thread.
     """
 
     def __init__(self, parent: QObject):
--- a/napari/_qt/qt_chunk_receiver.py
+++ b/napari/_qt/qt_chunk_receiver.py
@@ -54,6 +54,11 @@ class QtChunkReceiver:
     ----------
     parent : QObject
         Parent for QtGuiEvent.
+
+    Attributes
+    ----------
+    gui_event : QtGuiEvent
+        Listens for an event and signals us in the GUI thread.
     """
 
     def __init__(self, parent):
@@ -62,15 +67,20 @@ class QtChunkReceiver:
         self.gui_event.events.gui_event.connect(self._on_chunk_loaded_gui)
 
     def _on_chunk_loaded_gui(self, event) -> None:
-        """A chunk was loaded. This method called in the GUI thread
+        """A chunk was loaded. This method is called in the GUI thread.
 
         Parameters
         ----------
         event : Event
-            The event object passed with the EmitterGroup event.
+            The event object from the original event.
         """
         layer = event.event.layer
         request = event.event.request
+
+        LOGGER.info(
+            "QtChunkReceiver._on_chunk_loaded_gui: data_id=%d",
+            request.key.data_id,
+        )
 
         layer.on_chunk_loaded(request)  # Pass the chunk to its layer.
 
--- a/napari/_qt/qt_chunk_receiver.py
+++ b/napari/_qt/qt_chunk_receiver.py
@@ -1,62 +1,67 @@
+"""QtChunkReceiver and QtGuiEvent classes.
+"""
 import logging
 
 from qtpy.QtCore import QObject, Signal
 
-from ..components.chunk import ChunkRequest, chunk_loader
-from ..layers.base import Layer
+from ..components.chunk import chunk_loader
+from ..utils.events import EmitterGroup, Event
 
 LOGGER = logging.getLogger('napari.async')
 
 
-class QtChunkReceiver(QObject):
-    """Shuttles newly loaded chunks from the ChunkLoader to their layers.
+class QtGuiEvent(QObject):
+    """Fires an event in the GUI thread.
 
-    Notes
-    -----
-    The ChunkLoader's chunk_loaded event might be signaled in a worker
-    thread. The documentation only guarantees it will be signaled in thread
-    in this process, in theory could be any thread.
-
-    We do not want to call into model/layer code in a worker thread so we
-    signal ourselves in the GUI thread. If we are already in the GUI thread
-    this does no harm and happens instantly.
+    Listens to an event in any thread. When that event fires, it uses a Qt
+    Signal/Slot to fire a matching even in the GUI thread. If the original
+    event is already in the GUI thread that's fine, the matching event will
+    be immediately fired in the GUI threads.
     """
 
-    chunk_loaded_gui = Signal(Layer, ChunkRequest)
+    signal = Signal(Event)
 
-    def __init__(self, parent=None):
+    def __init__(self, parent, listen_event):
         super().__init__(parent)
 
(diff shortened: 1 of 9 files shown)


pwinston:
This is ready to review. This is a smaller step that outlined in #1510 but I think it's needed. I left out the Cache and all the statistics which seemed like they could go in as separate steps.

pwinston:
Still ready for review. I just did some cleanup and tried to anticipate some objections.

sofroniewn (approved):
This looks good to me @pwinston I have given the functionality a spin and can confirm the non-asyn behaviour is still as expected.

For async the basics are ok, but I did notice a couple of things are off like there are no thumbnails, and the image doesn't go to black/ when loading and the multiscale doesn't work (camera jumps around a lot). See this gif of the async working where the black and white image comes in later as it is a processed version of the magma image.

async-2_a

I think this behaviour is fine, whole point is to role out the full functionality slowly, but I just wanted to document this in case others try this out.

I've read through the code and it looks good to me, I don't have additional comments, but maybe @jni can do a deep review before merge.

napari/_qt/qt_chunk_receiver.py
```diff
@@ -0,0 +1,103 @@
...
+    def _on_chunk_loaded_gui(self, event) -> None:
+        """A chunk was loaded. This method is called in the GUI thread.
+
+        Parameters
+        ----------
+        event : Event
+            The event object from the original event.
+        """
+        layer = event.event.layer
+        request = event.event.request
```
jni: Sorry can you explain this? I'm lost now with the multiple event nesting...

napari/_qt/widgets/qt_layerlist.py
```diff
@@ -67,6 +69,8 @@ def __init__(self, layers):
         self._drag_start_position = np.zeros(2)
         self._drag_name = None
 
+        self.chunk_receiver = QtChunkReceiver(self)
```
jni, line 72: Is this doing anything? Or is it just that it needs to live *somewhere*?

napari/components/chunk/_info.py
```diff
@@ -0,0 +1,50 @@
...
+"""
+import logging
+import weakref
+
+from ...layers.base import Layer
+
+LOGGER = logging.getLogger("napari.async")
+
+
+class LayerInfo:
```
jni, line 11: I suspect that you'll want to keep this until the end of the async train but I feel like we can put this class on the chopping block also!

napari/components/chunk/_loader.py
```diff
@@ -1,52 +1,278 @@
...
+
+        # Log what we did.
+        if num_before == 0:
+            LOGGER.debug("ChunkLoader.clear_pending: empty")
+        else:
+            LOGGER.debug(
+                "ChunkLoader.clear_pending: %d of %d cleared -> %d remain",
+                num_cleared,
+                num_before,
+                num_after,
```
jni, line 189:
```suggestion
                f'ChunkLoader.clear_pending: {num_cleared} of {num_before} '
                f'cleared -> {num_after} remain'
```

Not positive this is better, but it feels more like modern Python. The C-style % formatting is increasingly rare.

napari/components/chunk/_loader.py
```diff
@@ -1,52 +1,278 @@
...
+
+    def _submit_async(self, request: ChunkRequest) -> None:
+        """Initiate an asynchronous load of the given request.
+
+        Parameters
+        ----------
+        request : ChunkRequest
+            Contains the arrays to load.
         """
+        LOGGER.debug("ChunkLoader._load_async: %s", request.key)
```
jni, line 148:
```suggestion
        LOGGER.debug(f'ChunkLoader._load_async: {request.key}')
```

napari/components/chunk/_loader.py
```diff
@@ -1,52 +1,278 @@
...
+        """
+        try:
+            request = self._get_request(future)
+        except ValueError:
+            return  # Pool not running, app exit in progress.
+
+        if request is None:
+            return  # Future was cancelled, nothing to do.
+
+        LOGGER.debug("ChunkLoader._done: %s", request.key)
```
jni, line 238:
```suggestion
        LOGGER.debug(f'ChunkLoader._done: {request.key}')
```

napari/components/chunk/_loader.py
```diff
@@ -1,52 +1,278 @@
-"""Stub ChunkLoader: sync only so far.
+"""ChunkLoader: synchronous or asynchronous chunk loading
 """
-from typing import Dict
+import logging
+import os
+from concurrent import futures
```
jni: Generally I like importing the module and accessing attributes with `.`, but in this case I think it clutters a lot of the code (`self.futures` (the list of futures) vs `futures` (the module), `futures.Future` in type signatures). So, a weak suggestion (totally optional) to change this to

```python
from concurrent.futures import Future, ThreadPoolExecutor, CancelledError
```

and make the corresponding changes in the code.

napari/components/chunk/_loader.py
```diff
@@ -1,52 +1,278 @@
...
+        else:
+            LOGGER.debug(
+                "ChunkLoader.clear_pending: %d of %d cleared -> %d remain",
+                num_cleared,
+                num_before,
+                num_after,
+            )
+
+    @staticmethod
+    def _get_request(future: futures.Future) -> Optional[ChunkRequest]:
```
jni: Just pure curiosity, why is this a staticmethod?

napari/components/chunk/_request.py
```diff
@@ -85,8 +85,10 @@ def __init__(self, key: ChunkKey, chunks: Dict[str, ArrayLike]):
         self.key = key
         self.chunks = chunks
 
-        # No delay by default.
-        self.load_seconds = 0
+    @property
+    def in_memory(self) -> bool:
+        """True if all chunks are ndarrays."""
+        return all(isinstance(x, np.ndarray) for x in self.chunks.values())
```
jni, line 91: This is a great little hack. =)

napari/components/chunk/_utils.py
```diff
@@ -0,0 +1,24 @@
...
+    We use data_id rather than just the layer_id, because if someone
+    changes the data out from under a layer, we do not want to use the
+    wrong chunks.
+    """
+    data = layer.data
+    if isinstance(data, list):
+        assert data  # data should not be empty for image layers.
+        return id(data[0])  # Just use the ID from the 0'th layer.
+
+    return id(data)  # Not a list, just us it.
```
jni:
```suggestion
    return id(data)  # Not a list, just use it.
```

napari/_qt/qt_chunk_receiver.py
```diff
@@ -0,0 +1,103 @@
...
+
+        LOGGER.info(
+            "QtChunkReceiver._on_chunk_loaded_gui: data_id=%d",
+            request.key.data_id,
```
jni, line 130:
```suggestion
            'QtChunkReceiver._on_chunk_loaded_gui: '
            f'data_id={request.key.data_id}'
```

jni (commented):
Thanks @pwinston!

I've made mostly small style changes/typo fixes. The LayerInfo class seems lightweight enough to remove but you might want to leave that to later PRs. Finally, the biggest things for clarification for me are:

- I'm unclear about the whole QtChunkReceiver/QtGuiEvent machinery... Can you explain that? I didn't follow the earlier discussion in the Zulip chat, and this part is really testing my understanding of threads and Qt.
- I'm also unclear about the event nesting, `event.event.layer`... Perhaps there is a way to rename some attributes that will make this code clearer?

Other than that, onwards and upwards! Great that this already adds some async functionality — didn't test it locally but @sofroniewn's little demo speaks volumes. =D

btw @sofroniewn your comments about multiscale not working properly are *only* for NAPARI_ASYNC=1, right? It might actually be worth exploring why that's happening, since I don't see anything in the code here that would look suspicious in that respect...!

jni:
> tried to anticipate some objections.

Nobody expects the Spanish Inquisition!!!

(Couldn't resist, I mean, it's friggin Python code...)

napari/_qt/widgets/qt_layerlist.py
pwinston: It needs to live somewhere and it's doing something, it's moving chunks from the worker thread to the GUI thread.

napari/_qt/qt_chunk_receiver.py
pwinston: I added two big comments to that file and renamed `event` to `original_event`.

napari/components/chunk/_info.py
pwinston: LayerInfo gets really big as you can see in #1510. It ends up storing all the statistics that we use to do the auto-async stuff.

napari/components/chunk/_loader.py
pwinston: With f-strings it will format all the strings, then call into the logging module, then perhaps 99.999% of the time it will discover the log-level is such that the logging is disabled, so the formatting was wasted effort.

This is why the logging module uses the older style of string formatting. Pylint wil flag you if you try to use other types of string formatting, which is the only reason I know about this guideline, from projects in the past that used pylint. Logging performance when logging is disabled is important, since that's the most common case, it will mostly be disabled.

See also:
https://docs.python.org/3/howto/logging.html#optimization
https://blog.pilosus.org/posts/2020/01/24/python-f-strings-in-logging/

There is potentially a way to do it in the "Use of alternative formatting styles" section of the logging cookbook, but it involves creating a new custom class for every log message:
https://docs.python.org/3/howto/logging-cookbook.html

I'd like to switch the async logging to JSON format which would wipe away this issue as well, although it might create other performance issues. Ultimately once in a while it would be good to time things with all the log messages totally commented out, and see if it's slowing things down even when disabled.

napari/components/chunk/_loader.py
pwinston: Same comment as above about logging performance when disabled.

napari/components/chunk/_loader.py
pwinston: Same comment as above about logging performance when disabled.

napari/components/chunk/_loader.py
pwinston: Okay changed this.

napari/components/chunk/_loader.py
pwinston: It's a staticmethod because it doesn't use self. Pylint will warn if a method doesn't use self. You can either make it a staticmethod or make it a stand-alone function. I usually make it a stand-alone function if it seems like something generic that we might use more than once. And I tend to make it a staticmethod if it seems very tied to class, something we aren't going to use again. Here I really could go either way, it seems like a toss-up.

But napari style could offer guidance, when to make a staticmethod (never, always, sometimes?) vs. a stand-alone function. If we wanted to try to be consistent. Either way it will execute identically, just has to live somewhere.

napari/components/chunk/_request.py
pwinston: I only changed that after the splitting. It was a regular loop before.

napari/_qt/qt_chunk_receiver.py
pwinston: Same comment as above about logging performance when disabled.

napari/components/chunk/_utils.py
pwinston: Fixed.

napari/components/chunk/_loader.py
jni: Fascinating! Thank you for the explanation. 👍 Perhaps put those two links in a comment above the first usage? That way I won't be tempted to "fix" it in 6 months when I come across this code and have forgotten this conversation. 😅

jni (approved):

napari/components/chunk/_loader.py
jni: In the future I generally would prefer functions, but definitely not a big deal here. As you say — one for the style guide.

napari/components/chunk/_loader.py
pwinston: I added a comment about the string formatting.

pwinston:
The visibility and thumbnails gets added in `async-3`, see table now at the description at the top of this PR.

Issue #1542, opened 2020-08-11
Colormap dropdown should be disabled for rgb images


sofroniewn:
## 🚀 Feature
The colormap dropdown in the top left corner should be disabled/ hidden for rgb images. Now it just shows `gray` which is not accurate

![image](https://user-images.githubusercontent.com/6531703/89918784-52046180-dbaf-11ea-8455-ec2a843d8b30.png)

## Additional context

napari 0.3.6

ekdnam:
Can I work on this?

sofroniewn:
> Can I work on this?

Yes, that would be great, a PR would be very welcome. Feel free to checkout out our [contributing guide](https://napari.org/docs/developers/CONTRIBUTING.html) if you havn't already.

ekdnam:
Okay sure

ekdnam:
Okay sure!

ekdnam:
how can the parameters of the `view_image` function be modified? and also the colormap dropdown menu? in the `view_layers.py` script, it is said that the `view_image` function is auto-generated

sofroniewn:
Hi @ekdnam you shouldn't need to modify view_image. Instead you could just add an if statement looking at the `layer.rgb` and if `True` not add the colormap widget to the qt controls for the image layer here https://github.com/napari/napari/blob/master/napari/_qt/layer_controls/qt_image_controls.py#L101-L102.

You could then add a test in a new file called `test_qt_image_layer.py`, similar to this file that tests that when an image is rgb the colormap dropdown has not been added. https://github.com/napari/napari/blob/master/napari/_qt/layer_controls/_tests/test_qt_image_base_layer_.py

I think having the dropdown not even get added is probably better than having it disabled, but you might want to try both to see

ekdnam:
Okay. Will do so. Thanks

ekdnam:
Hi @sofroniewn . I have removed the dropdown as well as the 'Colormap' label. But can't remove this ->
![napari-colormap-dropdown](https://user-images.githubusercontent.com/40426312/91309155-e258be00-e7cd-11ea-8c6f-410658a48bcc.jpg)

Which function / statement controls this icon? I declared the `colormapLayout` object such that it gets created only when `layer.rgb == False`, but then the icon gets shifted to the top layout corner. Sorry if I am needing a lot of help, I am new to GUI development and testing :)

sofroniewn:
[...]

PR #1586, opened 2020-08-27
don't show colormap for rgb images

ekdnam:
# Description

Solves issue: https://github.com/napari/napari/issues/1542

## Type of change

# References

# How has this been tested?

## Final checklist:


ekdnam:
@sofroniewn Please take a look

sofroniewn:
Hi @ekdnam I just gave this a try and i could reproduce the same problem you were mentioning about the colorbar appearing in the top left, which is likely due to the parent not being set properly. See

Screen Shot 2020-08-27 at 9 25 43 PM

Let's try a different approach. Instead of not showing the widgets, let's just hide them. I think that will be something like

```python
        if layer.rgb:
            self.grid_layout.itemAtPosition(3, 0).widget().hide()
            self.colorbarLabel.hide()
            self.colormapComboBox.hide()
```

Where we need to hide the label, the colorbar, and the combobox.

For ease of reading do you want to try putting this at the end of the init method (after everything gets added). You can also remove the extra commented lines you added above.

I think this should work. Let me know if you run into more problems!!

ekdnam:
@sofroniewn I have made the changes, please take a look. 

the line `self.grid_layout.itemAtPosition(3, 0).widget().hide()` led to an error, so I removed it, and the code works just fine.

napari/_qt/layer_controls/qt_image_controls.py
```diff
@@ -98,8 +98,12 @@ def __init__(self, layer):
         self.grid_layout.addWidget(self.contrastLimitsSlider, 1, 1)
         self.grid_layout.addWidget(QLabel('gamma:'), 2, 0)
         self.grid_layout.addWidget(self.gammaSlider, 2, 1)
-        self.grid_layout.addWidget(QLabel('colormap:'), 3, 0)
-        self.grid_layout.addLayout(colormap_layout, 3, 1)
+
+        # The colormap widget won't be created when image has a rgb layer
+        if layer.rgb is False:
+            self.grid_layout.addWidget(QLabel('colormap:'), 3, 0)
+            self.grid_layout.addLayout(colormap_layout, 3, 1)
```
sofroniewn: I'd undo these changes, I think if we are going with the hide approach it is fine to add them and then hide them.

napari/_qt/layer_controls/qt_image_controls.py
```diff
@@ -98,8 +98,12 @@ def __init__(self, layer):
         self.grid_layout.addWidget(self.contrastLimitsSlider, 1, 1)
         self.grid_layout.addWidget(QLabel('gamma:'), 2, 0)
         self.grid_layout.addWidget(self.gammaSlider, 2, 1)
-        self.grid_layout.addWidget(QLabel('colormap:'), 3, 0)
-        self.grid_layout.addLayout(colormap_layout, 3, 1)
+
+        # The colormap widget won't be created when image has a rgb layer
+        if layer.rgb is False:
+            self.grid_layout.addWidget(QLabel('colormap:'), 3, 0)
```
sofroniewn: let's actually store the reference to the colormap text label explicitly by making `self.colormapLabel = QLabel('colormap:')`. Can you do that here https://github.com/napari/napari/blob/master/napari/_qt/layer_controls/qt_image_controls_base.py#L56

And then here can call `self.grid_layout.addWidget(self.colormapLabel, 3, 0)`

You can also then change this line accordingly too https://github.com/napari/napari/blob/master/napari/_qt/layer_controls/qt_surface_controls.py#L39

Having this reference to the QLabel will make hiding the label much cleaner

napari/_qt/layer_controls/qt_image_controls.py
```diff
@@ -114,6 +118,11 @@ def __init__(self, layer):
         self.grid_layout.setColumnStretch(1, 1)
         self.grid_layout.setSpacing(4)
 
+        # Similarly, the colormap combo box will be hidden if the layer is rgb
+        if layer.rgb:
+            self.colorbarLabel.hide()
```
sofroniewn: If you see my other comment about leaving the adding of the widgets and creating `self.colormapLabel`, then here we just need to add `self.colormapLabel.hide()` and everything should work fine

sofroniewn (changes requested):
Hi @ekdnam this is coming along very nicely I think. I've requested a couple changes to make us only use the hide functionality.

We'll also want to add a dedicated test for this functionality. Basically you'll want to create one test with a non-rgb layer and check that these widgets are visible, and then you'll want to create another test with an rgb layer and check they are hidden.

You'll want to put the two new tests in a new file called `test_qt_image_layer.py` which will be very similar to this file, but which will for now only contain those two tests. https://github.com/napari/napari/blob/master/napari/_qt/layer_controls/_tests/test_qt_image_base_layer_.py

d-v-b:
In the interest of maintaining UX uniformity, why not keep the colormap dropdown but add "RGB" to it? Then someone can easily switch between viewing an image with shape (..., [3,4])-sized image as RGB or separate slices. When viewing an image where the dimensions don't permit a conversion to RGB, the "RGB" option is grayed out but still visible in the dropdown.

ekdnam:
@d-v-b I tried something different, now for RGB images the widgets would be turned off, by,

```
if layer.rgb is True:
            self.colorbarLabel.setDisabled(True)
            self.colormapComboBox.setDisabled(True)
```

The functionality of the colorbar label and the combo box gets disabled (they cannot be used now), but they don't become _grey_ per se.

Screenshot attached below: 

napari-disabled

ekdnam:
@sofroniewn @jni @tlambert03 @kne42 is disabling the widgets also an option that can be implemented here?

sofroniewn:
> In the interest of maintaining UX uniformity, why not keep the colormap dropdown but add "RGB" to it? Then someone can easily switch between viewing an image with shape (..., [3,4])-sized image as RGB or separate slices. When viewing an image where the dimensions don't permit a conversion to RGB, the "RGB" option is grayed out but still visible in the dropdown.

@d-v-b We were planning to expose those sorts of conversions in the layer list, rather than the colormap dropdown as they will change the layer representation, but it is an interesting idea, and maybe one we can follow up on. It is unclear to me how to undo that operation cleanly from that dropdown menu.

An alternative might just be to show `RGB` as a string in the combo box and disable it entirely as was done by @ekdnam here

> The functionality of the colorbar label and the combo box gets disabled (they cannot be used now), but they don't become grey per se.

We could also add hide the colorbar icon.

The following seems like a reasonable improvement on where we are now and will be easy to implement

- hide the colorbar
- set the combobox to read RGB
- disable the combobox so it can't be changed

I think we can think about more complex functionality in the future

ekdnam:
@sofroniewn should I move ahead with your suggestions?

Namely,

> hide the colorbar
> set the combobox to read RGB
> disable the combobox so it can't be changed

ekdnam:
@d-v-b what do you think?

d-v-b:
I agree with @sofroniewn

sofroniewn:
> @sofroniewn should I move ahead with your suggestions?

Go for it @ekdnam

ekdnam:
The ui now looks like this

combobox

ekdnam:
Quite a few tests are failing. What should be done to pass 'em?

tlambert03:
> Quite a few tests are failing. What should be done to pass 'em?

two issues:
- you need to give you QLabel a parent, so that it is properly cleaned up when the widget is destroyed:
`self.colormapLabel = QLabel('colormap:', parent=self)`
- surface `add_layer` tests are failing because `QtSurfaceControls` inherits `QtBaseImageControls` ... but surface layers have no attribute `rgb`, so this code fails for surface layers:
https://github.com/napari/napari/blob/0c9022ff73a449f9fe366e8cef879cd9909e2431/napari/_qt/layer_controls/qt_image_controls_base.py#L54-L58

you could check for `if hasattr(layer, 'rgb')` ... but I wonder whether this isn't indicative of a deeper design issue.  @sofroniewn ?

sofroniewn:
> you could check for if hasattr(layer, 'rgb') ... but I wonder whether this isn't indicative of a deeper design issue. @sofroniewn ?

Agreed indicative of a large issue that we might want to resolve when do things like the `ColorManger` clean up see #1356, so I think it's probably ok to use that quick fix for now.

Thinking about this more @ekdnam I wonder if instead of adding `RGB` to the combobox and disabling it, maybe we should just hide the combobox and replace it with a `QLabel` with RGB instead, then it will be clear that would can't click on the combobox at all.

I worry now that people might try and click the combobox and be confused about why it is disabled

sofroniewn:
Also in conversations @jni made the suggestion to make the label `(RBG)` instead as it isn't a colormap in quite the same way that the others are

ekdnam:
Okay, I will add a QLabel for RGB images, and also `parent=self`.

liu-ziyang:
@ekdnam Hi there, I created another PR to update your proposed change, you may need to pull master to your branch to see the clean change list, hopefully this unblocks your PR: https://github.com/ekdnam/napari/pull/1

ekdnam:
@ziyangczi  Okay, will rebase my branch with master, and also look into your changes. Thanks!

ekdnam:
What should be done to solve these errors? cc @sofroniewn @ziyangczi

napari/_qt/layer_controls/qt_surface_controls.py
```diff
@@ -36,7 +36,7 @@ def __init__(self, layer):
         self.grid_layout.addWidget(self.contrastLimitsSlider, 1, 1)
         self.grid_layout.addWidget(QLabel('gamma:'), 2, 0)
         self.grid_layout.addWidget(self.gammaSlider, 2, 1)
-        self.grid_layout.addWidget(QLabel('colormap:'), 3, 0)
+        self.grid_layout.addWidget(self.colormapLabel, 3, 0)
```
liu-ziyang: I believe this is the line causing issue, there is no colormapLabel in the object, you may want to revert it

ekdnam:
@ziyangczi thanks! all checks have been passed.
@sofroniewn can this PR be merged now?

sofroniewn (approved):
Yes this looks good to me. Here's what the final product looks like, very nice.

Thanks for the work here @ekdnam and thanks @ziyangczi for help finishing this up! I will merge now

Screen Shot 2020-10-20 at 8 28 52 AM

PR #1590, opened 2020-08-28
async-3: new ChunkCache

pwinston:
# Description
* New `ChunkCache`
* See more async PRs [Napari Rendering GitHub Project](https://github.com/users/pwinston/projects/1)


pwinston:
Not sure about test failures. I will take a look sometime today.

sofroniewn:
> Not sure about test failures. I will take a look sometime today.

That `check_view_transform_consistency` has given me some problems in the past. We'll probably want to simplify it at some point as there is quite a lot going on in there.

The other fail looks like a simple requirements one now that we have the bundled app merge, the requirements may need to be updated in two places, though I havn't checked

pwinston:
To fix `test_viewer.py::test_add_layer` here I had to add a vispy `Visual.loaded` boolean and use that instead of the `Visual.visible`. Not sure how this didn't bite us in #1510. But basically `test_add_layer` purposely adds invisible layers and expects them to be updated normally. But we were not updating invisible layers.

With `Visual.loaded` that problem goes away. Because only we set that during async load. So adding invisible layers works normally now. But async loading still works.

The octree-style visual `Visual.loaded` can probably can go away. Since that will always be visible and will manage its children explicitly. Which is good because added `loaded` flag to `Visual` is probably not something anyone would want upstream.

sofroniewn:
hi @pwinston, looks like tests are failing on a `AttributeError: 'Image' object has no attribute 'loaded'` which is probably a pretty simple fix at this point. If you update I can do a functionality test after all the tests are passing. Thanks!!

pwinston:
The tests are fixed now. Caught a real problem even when not using async. So that's good. Now should be zero change if not using async.

napari/_vispy/vispy_image_layer.py
```diff
@@ -57,7 +57,18 @@ def _on_display_change(self, data=None):
...
+    def _data_astype(self, data, dtype):
+        """Broken out as a separate function so we can time with perfmon."""
+        return data.astype(dtype)
+
     def _on_data_change(self, event=None):
+        """Our self.layer._data_view has been updated, update our node.
+        """
+        if not self.node.loaded:
+            # Do nothing if we are not yet loaded. Calling astype below could
+            # be very expensive. Lets no do it until our data has been loaded.
```
jni:
```suggestion
            # be very expensive. Lets not do it until our data has been loaded.
```

napari/components/chunk/_cache.py
```diff
@@ -0,0 +1,117 @@
...
+        Parameters
+        ----------
+        request : ChunkRequest
+            Add the data in this request to the cache.
+        """
+        if not self.enabled:
+            LOGGER.info("ChunkCache.add_chunk: disabled")
+            return
+        LOGGER.info("ChunkCache.add_chunk: %s", request.key)
+        self.chunks[request.key.key] = request.chunks
```
jni, line 98: I think the `.key.key` idiom can be avoided by giving ChunkKey a `__hash__` method that returns `hash(self.key)`...? See https://docs.python.org/3/reference/datamodel.html#object.__hash__

jni: In fact, maybe we should do this for the *request* directly?

pyproject.toml
```diff
@@ -25,6 +25,7 @@ icon = "resources/icon"
 requires = [
   "appdirs>=1.4.4",
   "cachey>=0.2.1",
+  "cachetools>=4.1.0",
```
jni: How much are we using/envisioning using the features provided by cachetools? I get the impression that we could create our own thin wrapper around `functools.lru_cache`. In fact, having just checked the repo for cachetools, I do think it's too small for us to depend on it and I'd rather vendor the code in our own repo...

jni (dismissed):
@pwinston does it break the sequence dramatically if we remove the cache altogether for now? As you note in your todo, the motivating examples for single chunk async is dynamic computation, where I do think the cache should be off by default given the napariboard example, *and* for other dask-backed compute this buys us very little (since we already have the dask cache). I *think* this helps with `zarr.open`ed arrays — but that also [has its own cache](https://zarr.readthedocs.io/en/stable/api/creation.html#zarr.creation.open_array)!

So, this PR would then become just making the node invisible until loaded, which I think stands alone anyway. From my reading of this PR, it's *not* turning off the not loaded/loaded flickering if there is a cache hit (?), which might indeed be a good reason to manage our own cache.

At any rate, whatever we decide, I think in the best case we should vendor the LRUCache code from cachetools rather than depend on them.

Finally, I've made a rather small suggestion for improvement to the whole chunkkey/chunkrequest infrastructure which is probably worth making with or without the cache.

I wonder what @sofroniewn thinks about all this...

napari/_vispy/vispy_image_layer.py
```diff
@@ -57,7 +57,18 @@ def _on_display_change(self, data=None):
...
         self.reset()
 
+    def _data_astype(self, data, dtype):
+        """Broken out as a separate function so we can time with perfmon."""
+        return data.astype(dtype)
+
     def _on_data_change(self, event=None):
+        """Our self.layer._data_view has been updated, update our node.
+        """
+        if not self.node.loaded:
```
sofroniewn: Is there a reason you can't just use `self.layer.loaded` here too? I don't really like having `loaded` on the `node` as we're going to want to remove that vendored copy of the vispy Image Visual once vispy makes their next release, and I don't think we are likely to push the `loaded` variable upstream to them any time soon. We could just have it on this object instead if we need.

sofroniewn:
I'm fine with either leaving the cache stuff and the visible stuff in here together or splitting it out.

I'm also fine with vendoring the necessary cachetools code

I did leave one comment about maybe not putting the `loaded` attribute on the vendored vispy code. Maybe we should make it more clear that the vispy visual code we only imagine as temporarily vendored until the next vispy release, which is different from all of our other vendored code ...

I was also able to confirm that all the functionality behaved as expected, including the waiting for loaded with async on, see below:

async3

pyproject.toml
pwinston: I personally strongly dislike the idea of knee-jerk vendoring things out of fear when there are no actual known issues with the dependency. I feel like it's disrespectful to the other project and severs what could be a useful channel for both bug fixes and improvements. Who is ever going to manually  check the upstream project and diff it against our vendored version? In the past I've only seen people vendor things when they need to modify them. I'm new to this idea of just copy-n-pasting other people's projects into your own. But if that's the napari-way I can do that?

pwinston: As for envisioning I have not envisioned anything beyond #1510 so I can't predict if/when/how we will use cachetools in the future. As for using `functools.lru_cache` the `cachetools` API is exactly what we wanted and we've been using it for months without any problems. I cannot estimate how long it would take to make functools do the same thing, without more or less doing it, but I'm confident it's a non-zero amount of time.

napari/_vispy/vispy_image_layer.py
pwinston: Fixed

napari/_vispy/vispy_image_layer.py
pwinston: I changed it to self.layer. This used to be "visible" not "loaded" and when I switched it to loaded I didn't make this change. However there is still a "loaded" in the vendored image/volume because it's checked in `VispyImageLayer._on_data_change`. There might be some other way to do that, but that's how it's always worked in #1510 (although it changed from visible to loaded recently).

So I think the plan is finished splitting #1510, then add the octree visual version 0 which will get rid of this "loaded" flag. Instead we'll have an always visible container of some kind that manages its subchildren. That's where we need to be for many reasons.

napari/components/chunk/_cache.py
pwinston: I tried putting `__hash__` on `ChunkRequest` and it didn't work, just got blank everything. I didn't try to debug it. In theory cachetools can do something with the key that doesn't go through `hash()`. Although I didn't see it doing that at first glance. But externally you can't know it only uses the key for dict lookups.

I agree `key.key` is ugly, but I think we have major cache changes in store. In general I think the cache stuff is like 10% done. I just put in the barest minimum. We have to deal with single vs. multi-scale, dask, and then real chunks with the octree, etc.  So once we have things actually working across the board, I think we can worry about tidying up the exact syntax.

pwinston:
> can we remove the cache altogether for now? 

I tried running it without the cache when I created this PR. It did not function at all without the cache. This was surprising to me but I didn't investigate. I think currently we redundantly ask for the same slice, and if we don't have a cache it stacks the requests? I just know the cache has been in there for months in #1510 and we were able to get threads and processes working, single and multi-scale, with this same code.

In `async-4` we will add the config file and we can introduce settings to turn it off (assuming that works) and other things. Whether it's on or off "by default" I think is a minor detail compared to being able to configure it explicitly. And the config file in `async-4` is how to configure it.

As I said above I think caching in general is like 10% done. We put in something very basic and with it we were able to show a lot working. But certainly we need to do the other 90% of the work and figure out what cache we are doing to use under what conditions, etc.

It will be good to get async stable and used enough that people start to care about the cache behavior. I have not seen anyone use it enough to care yet. Once people are like "async is rock solid but when I use it for hours I don't like how the caching works" that will be a good day. Hopefully we can get there soon.

napari/_vispy/vispy_image_layer.py
sofroniewn: > However there is still a "loaded" in the vendored image/volume because it's checked in VispyImageLayer._on_data_change. There might be some other way to do that, but that's how it's always worked in #1510 (although it changed from visible to loaded recently).

Looking at the code in #1510 and the code now in this PR I really do think you could drop the entire `loaded` property from the visual right now if you wanted. I see how it gets used inside `set_data` on the visual to trigger a return, but we control when that the `set_data` call happens in our `_on_data_change` and `_on_display_change` methods and so can check `self.layer.loaded` before making the call from both those functions (right now you already check from `_on_data_change`). It doesn't look like there are additional calls to `set_data` that can get triggered internally by vispy, and I'm not quite sure how they would use the method if they wanted, but I guess that might be one thing to verify. If not, then having the checks in our code would mean you wouldn't need the check in `set_data` anymore and could drop the `loaded` property entirely from the visual.

I did just give it a try to delete `loaded` from the node and comment out the `self.node.loaded` references and async still seemed to worked as before, so it is probably pretty harmless to make this change and would mean that we'd have a much easier upgrade path after the next vispy release.

> So I think the plan is finished splitting #1510, then add the octree visual version 0 which will get rid of this "loaded" flag. Instead we'll have an always visible container of some kind that manages its subchildren. That's where we need to be for many reasons.

That sounds like a good plan, and I won't hold up the PR for this change, but I do think dropping `loaded` from our vispy image layer object will be easier than dropping from the vendored vispy visual, which really shouldn't be in the napari code base and is only there due to the lack of a recent vispy release.

pwinston: Okay yes it works your way, I changed it. It's better. I did it the other way basically just because checking inside "set_data" I knew we were 100% covered and it's pretty hard to search for "all places we trigger an event". But it seems to work how you say and yes it's much better to not change the vendored files.

But even the checks should go away with the octree visuals version 0, hopefully.

sofroniewn: Great, thanks @pwinston much appreciated

sofroniewn (approved):
Ok, this looks good to me now, thanks for pulling the `loaded` property off the vispy visual!

Given the comments around the cache I think this is good to go in as in, but maybe @jni can give it one final look.

napari/components/chunk/_cache.py
jni: @pwinston Ok, fair that it doesn't work, what about putting it in ChunkKey though?

pyproject.toml
jni: @pwinston I'm totally on board with pushing to upstream projects — you will find scattered across the repo many comments from me about how I want us to contribute more directly to VisPy to the point that we never need to vendor their code again. But one difference is VisPy has 116 contributors (!) while cachetools has 3. That's a low number of eyeballs to prevent things like left-pad or even worse, event-stream from happening. I wouldn't say it's "the napari way" so much as general wisdom gathered in the FOSS community over the past couple of decades, see e.g. https://twitter.com/gaelvaroquaux/status/1289247135293005824. scikit-learn, for example, requires going through an enhancement proposal process to acquire new runtime dependencies.

Ultimately we want napari to sit right alongside matplotlib in the scientific Python stack. That means lots of downstream dependents which means we have to be super careful about what *we* depend on.

So, I definitely would like to vendor the LRUCache. Vendoring absolutely involves attribution to the source *and*, if we make improvements, contribution of those improvements upstream. That's on us. In terms of checking upstream for improvements, well, we can check *if* we run into any issues, which as you say, we haven't. =) There *is* a big advantage to small code!

jni:
> I tried running it without the cache when I created this PR. It did not function at all without the cache. This was surprising to me

It's surprising to me also! I've started investigating. I'm not sure whether I'll get to the bottom of it tonight so I'm going to start making some notes. My test case is `NAPARI_ASYNC=1 napari https://s3.embassy.ebi.ac.uk/idr/zarr/v0.1/6001240.zarr`.

- setting the cache to 0 or a small value fails, because if a single cache item is bigger than the cache size, cachetools [raises a ValueError](https://github.com/tkem/cachetools/blob/31a78b5238a686d12eb11ba2850a96d9d68f49cd/cachetools/cache.py#L46-L47). As far as I can tell from their docs, disabling the cache is something we need to do locally — there's no way to turn it off on the cachetools size.
- setting the cache size to be exactly just above a single chunk works fine, even when loading multiple chunks (as in the case of multiple layers, as with this test case. In my case the cache size is 187,260 bytes (setting CACHE_MEM_FRACTION to 1.09e-5 on my 16GB Mac).
- Unrelated to this PR, but 3D viewing in async is broken. The error is [here](https://gist.github.com/jni/a49c8d285a07f52ba68333b4888900f8). This should be fixed. (You can try the volume above, then turn on 3D view.)
- If I try to rip out the cache ([this commit](https://github.com/jni/napari/commit/56a65ab0d2e9cf70cc9834b452cb01533a0bf17c)), indeed the volume is always black. Checking the layer status with e.g. `lamin = viewer.layers[0]; lamin.loaded`, I always get `False`. So to me it looks like the chuckloader/catcher aren't working correctly in this PR — and also in master, I just discovered.

That's what I've got for now. I'll try to keep hunting in a bit, for now it's dinner time though. But I think

- (a) we should figure out what's going wrong here — somewhere, the cache path gets the layer's "loaded" attribute set to True, and that doesn't happen without the cache. That means there's something wrong with the architecture somewhere, because a layer's load status should not depend on the cache.
- (b) We should not let cachetools raise when a chunk is bigger than the cache. So that's a decent extra incentive for vendoring.
- As a bonus, it would probably be good to fix async 3D. That may or may not be something to do in this PR, though. But noting it in case the cause jumps out at you @pwinston.

Side note, in investigating this I came across a big warning in the cachetools docs that the caches are not thread safe and that if multiple threads are accessing the cache, they should use locks. In this particular scenario multiple threads *do* write to the cache (one per layer I think?), so this might be something to fix sooner rather than later. (The comment in the code asserts that it's safe to do it in the worker, but that seems at odds with the docs? Please correct me if I'm mistaken here.)

pyproject.toml
pwinston: Okay I can vendor it. Where does the code live it's not one file like humanize was?
https://github.com/tkem/cachetools/tree/master/cachetools
or are we going to rip apart the dependency and take only the files we are currently using today? Only the lines? Also do we exclude vendored code from Black and isort, or do we reformat them so we can't as easily diff them against future versions?

I don't think you addressed my question that vendoring "severs what could be a useful channel for both bug fixes and improvements.". How would we ever know if a dependency fixed a bug or added a feature if we have our own copy of the code? Do we have a list of code we've copied? When was each last checked for fixes or improvements?

As for "left pad" there are 85 forks of cachetools. If someone yanked cachetools when we vendor it from a fork. I feel like you are glossing over the disadvantages of vendoring and citing a single fear that does not happen often. We used many packages at my last job for 3 years and did not see one problem. Yes it's a annoying if a dependency disappears, but it feels like it's very easy to vendor-on-demand instead of pre-emptively. Vendor that one in a hundred or one in a thousand package that has a problem, instead of vendoring everything.

Copying and pasting in code whenever you want to use someone's code feels prima facie wrong to me. It feels like a pre-GitHub solution. It feels very asymmetric, we are totally behind people using our package, in fact highly desirous of that, but instead of using other people's package's we'll copy and paste them.

tlambert03: I agree with @jni on this one

pwinston:
This is the 90% of the work I said we have left to do. The cache has worked for months in #1510 and we are trying to get #1510 split up and into master as opt-in so that we can iterate on it. #1510 works for single scale, multi-scale, threads and processes. I think getting to the bottom of caching and improving caching is on the development path for async. But I don't see why it's necessary to interrupt the splitting process, and delay it, and delay getting known-working code into wider use, wider testing and more people working on improvements,  in order to iterate on the cache right now at this moment?

pyproject.toml
jni: > Okay I can vendor it. Where does the code live it's not one file like humanize was?
> https://github.com/tkem/cachetools/tree/master/cachetools

Use your best judgment. `napari/components/chunks/vendored/cachetools` would be fine.

> or are we going to rip apart the dependency and take only the files we are currently using today? Only the lines?

I would only keep LRUCache and what's needed for that (the base Cache class). Can probably go in one file. Include the original license and link to the original repo.

> Also do we exclude vendored code from Black and isort, or do we reformat them so we can't as easily diff them against future versions?

I'd exclude but I am agnostic on this — if you don't like the look of the files feel free to reformat. They are not a lot of code.

> I don't think you addressed my question that vendoring "severs what could be a useful channel for both bug fixes and improvements.". 

I addressed one direction of it — our contributions to the package. We are in full control there, if we make improvements or bug fixes we should contribute them upstream.

As to bug fixes in the source, it's a problem, granted, but this is not huge code. If we are affected by the bug I'm sure we'll find out sooner than later, and fix it. If we are not affected, well, we can check periodically — development on that repo is hardly blistering.

> We used many packages at my last job for 3 years and did not see one problem.

I presume you used a fair bit of version pinning, static linking, or bundling together with your dependencies. In the Scientific Python ecosystem, pinning is rare and not recommended (for reasons like the ones you mentioned — you don't benefit from updates), and updates that break stuff happen **all the time**, not rarely. I can't count the number of times we had to fix stuff in skimage because a dependency changed something. (Downstream packages suffer the same fate, and napari had to scramble to fix our CI with skimage 0.17.) Dependencies are very much a liability, at least in this ecosystem

> Copying and pasting in code whenever you want to use someone's code feels prima facie wrong to me. It feels like a pre-GitHub solution. It feels very asymmetric, we are totally behind people using our package, in fact highly desirous of that, but instead of using other people's package's we'll copy and paste them.

There's a balance. For something the size of NumPy, you'd depend on it. For one function, you probably would just copy the function (most often from Stack Overflow). The in-between is gray. I'm arguing that this falls more on the "one function" side.

I absolutely don't mind that people vendor napari code if they only need a couple of functions from it, btw, as long as they attribute the source. If they vendor the whole thing, I also don't mind that — it's more work for them. This is not about courtesy (though courtesy is great and again, we should always attribute code to the author), but about pragmatism.

tlambert03: > In the Scientific Python ecosystem, pinning is rare and not recommended (for reasons like the ones you mentioned — you don't benefit from updates), and updates that break stuff happen **all the time**, not rarely. I can't count the number of times we had to fix stuff in skimage because a dependency changed something. (Downstream packages suffer the same fate, and napari had to scramble to fix our CI with skimage 0.17.) Dependencies are very much a liability, at least in this ecosystem

couldn't agree more.  I have been burned so many times by this... and the pinning "solution" makes solving (e.g.) conda environments extremely difficult over the long term.   this is probably the main thing I "wish I had known" when preparing earlier packages for distribution.

pwinston: > I would only keep LRUCache and what's needed for that (the base Cache class). Can probably go in one file. Include the original license and link to the original repo.

Reformatting and file-refactoring means you can no longer diff our copied version with their version. Since you've copied-n-pasted instead of using git, diffing is the only way to compare our version with their version.

Do you still maintain it's a good idea to do this type of reformatting given it will make it harder to diff? What is the plus side of combining into less files? Reformatting and file-refactoring seems to problems with zero gain. What is the gain? Seems like this is not a well thought out convention when copying other people's code.

> I'd exclude but I am agnostic on this

Now do you exclude vendored files from reformatting, what is the procedure?

Where is the list of vendored/copied packages we are using, so I can add it to the list? Do we keep track of when we've checked those packages for bug fixes and improvements? I feel like I asked these question already.

tlambert03: > Do you still maintain it's a good idea to do this type of reformatting given it will make it harder to diff?

doesn't look like @jni said this at all.  in fact he said:

> Use your best judgement. I'd exclude but I am agnostic on this — if you don't like the look of the files feel free to reformat. They are not a lot of code.

historically here we've taken this on a case-by-case basis.  See, for instance, all of the code (including the "vendored" folder) that we do not reformat in the `[tool.black]['exclude']` section of `pyproject.toml`.  On the other hand, if it's a small amount of code that we probably will never diff or push upstream in the future, we simply absorb the code and reformat.  You can do what feels best to you here.

> Seems like this is not a well thought out convention when copying other people's code.

please refrain from this sort of statement.  If you don't understand something,  first make sure that you've actually read and internalized what others are saying.  Then ask questions or provide counter arguments and leave out your general evaluation of our thought process.

pwinston: > doesn't look like @jni said this at all

He did say it. He said "if you don't like the look of the files feel free to reformat. They are not a lot of code." and he said "can probably go in one file" so my response was "Do you still maintain it's a good idea to do this type of reformatting given it will make it harder to diff?"

I said that because given my concerns, reformatting and re-file factoring sounds like all downside with zero upside. What is the purported upside? Putting our stamp on their code so it looks like our code? And is harder to maintain? To "save space"?

> please refrain from this sort of statement.

I said "seems like this is not a well thought out convention when copying other people's code" because I'm being paid to give my recommendations and I feel very strongly these conventions around copying instead of re-using are not well thought out. That's my professional determination, I'm ethically bound to give that advice since I believe the current approach is harmful, will waste time and money in the long run, and will incur unnecessary risk.

If napari is going to standardize around copying I'd make the following recommendations:

1. Put all copied code under one root so it's not hidden throughout the code base. This makes it easier to keep of how much code was copied and from where.

2. Vendor the full source code unless it's gigantic. This means people using the code know what features are available. Subsetting creates a new artifact that is not the original.

3. Take great care not to reformat the code in anyway, including via CI, so that's it's easier to diff, easier to compare with the original for bug fixes and improvements.

4. Have a very clear legal review process. Certainly do not count on the average developer to make a legal determination about copy-n-pasting and modifying the code.

5. Have some type of review process so dependencies are checked for changes periodically, once a quarter or once a year, some interval more often than "never".

I'm not going to insist napari follow these practices by any means. But these seem like no-brainers to me. 

 The opposite approach is copy in other's people's code willy-nilly and hide it throughout our source tree, subset it and reformat so it's unclear what the original provided and it's harder to maintain, and count on developers to make legal determinations about copying in an ad hoc disorganized way. Those seems like bad practices to me.

To me it sounds like the recommended procedures are built around a "one time grab" and are not appreciating the full lifecycle where ongoing maintenance can be a huge factor and cost.

I think a good thought experiment is if someone wants to conduct an audit on what code napari has copied vs. written, for legal or technical reasons, do we have a nice simple answer for them?

pwinston: I would summarize by saying that including dependencies with pip or similar is the normal way to leverage someone else's work. And that using this standard mechanism, flaws and all, would likely not fail any due diligence test.

If you are going to vendor instead, even when no changes are needed, then you are taking on additional work and responsibility. If individual developers are going to paste in other people's code as their own, that's hard to police, but I'd make the project's approved method be for a very clear separation between napari's code and 3rd party code, for both technical and other reasons.

I acknowledge that maybe in some open source circles things are done differently. I can imagine some people just glance at the license then paste it in and never look back. This is not an approach I can recommend to an employer. But ultimately it's up to the project.

Here's the vendoring policy of pip itself, including a single parent `_vendor` directory and a rule to not modify them unless needed, modifying would include formatting, combining files, or subsetting them:
https://pip.pypa.io/en/stable/development/vendoring-policy/

pwinston:
I can try to make an `async-2.5` without the cache. It will require some debugging. And then we will immediately come back to wanting to merge this `async-3` PR. But I can give it a shot.

For the vendoring I can look at properly vendoring the full cachetools dependency. Even the whole thing is tiny. I don't think it's appropriate to subset or combine files, for the technical, legal and common sense reasons outlined above. I think the PIP vendoring policies are all sane, they appear to have been down this road and left a well lighted path for us to follow, unless we feel like we can innovate here and improve on their approach, I can't see a good reason to do otherwise. And their method is arguably easier, not harder, since you just drop it in unchanged instead of creating a custom thing.

As always having any policy, even if it's not perfect, it's the big step. What that policy entails is more of a tweak.

pwinston:
I created async-2.5 which has the vispy changes from this PR without the cache: https://github.com/napari/napari/pull/1607

To fix it I had to add a check in `ImageSlice` so that if we re-ask for a slice we are loading, it will not re-submit the request. Why we re-ask I'm not sure, it's related to `view_view_slice` and `refresh`. However not going to into a loop is clearly an improvement, and it fixes the problem. The cache avoided this problem a different way.

The index at the top updated to have links to `async-1` through `async-6` plus `async-2.5`.

sofroniewn:
> To fix it I had to add a check in ImageSlice so that if we re-ask for a slice we are loading, it will not re-submit the request. Why we re-ask I'm not sure, it's related to view_view_slice and refresh. However not going to into a loop is clearly an improvement, and it fixes the problem.

Yeah, we've had some problems with multiple calls to `refresh` in other parts of the code base. I think we're hoping that things like #1475 and followups will make the event logic and flow a little simpler, but good to fix this issue now. Thanks!!

pwinston:
I put `cachetools` in a new top-level `_vendor` directory which is the convention that PIP uses:
https://pip.pypa.io/en/stable/development/vendoring-policy/

Note that PIP has special motivation towards vendoring because they don't want depend on very much, because they are pip. They also have special tools I didn't bring over at all.

Basically I took these practices for their guidlines:
1. Put the entire dependency in `_vendor`. Yes this is different from `vendored` I think that's okay in that that's the existing convention and this is a new convention. But we could change the name obviously.
2. I actually did delete a few files that were purely about packaging or CI, but the entire source is there.
3. I put version in `_vendor/vendor.txt`
4. I not use their fancy stuff from their `_vendor/__init__.py`. Our `__init__.py` just imports cachetools.
5. So in our code we do `from ..._vendor.cachetools.cachetools import LRUCache`

Seems like the double `cachetools.cachetools` should not be require, just `from ..._vendor.cachetools import LRUCache`, but that didn't immediately work. Doesn't matter too much.

pwinston:
Okay the `_vendor/cachetools` is working now.

The `all() or any()` thing I actually fixed a very different way. I now `hash()` everything up front here:
https://github.com/pwinston/napari/blob/feature/async-3/napari/components/chunk/_request.py#L56-L62

It was messy to have tuples with slices, and comparisons. Hash seems like it does the job. The only point is to tell one request from another.

The `ChunkKey` class is a good example of something we can re-write and tune whenever. As long as it does the job. There's nothing I like about the class except that it works. Which is an important attribute.

pwinston:
Okay I think the `_vendor/cachetools` is working now.

I had to put this `graft` line in the `MANIFEST.in` which seems like it should not be needed?
* https://github.com/pwinston/napari/blob/feature/async-3/MANIFEST.in#L2

pwinston:
I converted this to DRAFT because we don't want to merge any more async PR's until we get stuff into `experimental` namespaces. That's in progress.

pwinston:
Closing this since I opened #1659 which is async-2.6 to move async stuff into "experimental".

PR #1592, opened 2020-08-29
Minor updates to custom mouse functionality examples

jni:
# Description

Our examples were using outdated mouse callback machinery. Additionally, one gets deprecation warnings from scikit-image.

See https://forum.image.sc/t/how-to-get-the-coordinates-of-the-mouse-when-it-is-clicked-over-an-image-labels-layer/42175/2 for more.

Incidentally, I wonder whether we should just provide `mouse_press_callbacks` as a synonym for `mouse_drag_callbacks`. I don't find it very intuitive to write a drag callback for a button press...

### DIFF
--- a/examples/custom_mouse_functions.py
+++ b/examples/custom_mouse_functions.py
@@ -16,7 +16,7 @@ with napari.gui_qt():
     labeled = ndi.label(blobs)[0]
     labels_layer = viewer.add_labels(labeled, name='blob ID')
 
-    @viewer.mouse_press_callbacks.append
+    @viewer.mouse_drag_callbacks.append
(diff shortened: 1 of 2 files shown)


sofroniewn:
> Incidentally, I wonder whether we should just provide mouse_press_callbacks as a synonym for mouse_drag_callbacks. I don't find it very intuitive to write a drag callback for a button press...

Maybe we should. I like the proposal in #1048, but this could be faster in the mean time.

sofroniewn (approved):
LGTM!

sofroniewn:
We also might want to run our examples again as part of our CI

PR #1595, opened 2020-08-31
do not raise viewer in IPython with pre-existing event loop

### LINKED ISSUE #1594: Consider not raising when launching from IPython
## 🐛 Bug

When launching from Jupyter, it's common to put the entire preparation code (create the viewer, add an image, add labels, add points, ...) in the same cell, so raising by default is the natural choice. On the other hand, when I launch napari from IPython, typically I launch a Viewer, and still have quite a few statements to type before I want to actually use the viewer. In this situation, the new-ish "always raise" behaviour is very annoying. I'd like us to detect when we are launching from IPython and not raise in that situation.

tlambert03:
# Description
fixes #1594
The viewer will no longer be automatically brought to the front when creating a viewer in an IPython session that already has a `%gui qt` event loop running.  It will, however, always be brought to the front in a jupyter lab/notebook environment... and it will also be brought to front in an IPython environment that is _not_ already running an event loop, when using `gui_qt` (since there's not much you can do in that blocked context manager in IPython until you close the viewer anyway).
In all cases, the window can be brought to front by using the newly added `viewer.window.activate()`

@jni let me know if this works for you

## Type of change

# References
#721, #732, #735, #795, #1594 

# How has this been tested?

- [ ] manually in jupyter lab, ipython, regular scripts.  Not sure how to test for running in, for instance, a jupyter environment besides simply mocking the test that I use, which feels pretty useless.

## Final checklist:

### DIFF
--- a/napari/_qt/qt_main_window.py
+++ b/napari/_qt/qt_main_window.py
@@ -22,6 +22,7 @@ from qtpy.QtWidgets import (
 from ..resources import get_stylesheet
 from ..utils import perf
 from ..utils.io import imsave
+from ..utils.misc import in_jupyter
 from ..utils.theme import template
 from .dialogs.qt_about import QtAbout
 from .dialogs.qt_plugin_report import QtPluginErrReporter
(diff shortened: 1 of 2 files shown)


jni (approved):
Works a treat! 🎉🎉🎉 thank you!

At first I was unsure about `in_ipython` since you're not using it, but tbh I like having it there as a reference implementation.

I'll let @sofroniewn do the honours.

sofroniewn (approved):
LGTM!

PR #1597, opened 2020-08-31
Restore multiple command-line arguments

joshmoore:
`napari -vvv my.zarr` fails with

```
  File "/opt/napari/napari/__main__.py", line 214, in _run_pythonw
    cmd.append(*sys.argv[1:])
TypeError: append() takes exactly one argument (2 given)
```

see: #1554

### DIFF
--- a/napari/__main__.py
+++ b/napari/__main__.py
@@ -211,7 +211,7 @@ def _run_pythonw(python_path):
 
     # Append command line arguments.
     if len(sys.argv) > 1:
-        cmd.append(*sys.argv[1:])
+        cmd.extend(sys.argv[1:])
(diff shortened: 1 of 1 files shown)


tlambert03 (approved):
good catch thanks

joshmoore:
Purely self-motivated :wink: (Definitely rely on debug output for plugin development) I was wondering why this didn't trigger a failure. Digging in a bit, it looks like the GHA matrix is going to need a version with and without `python.app` installed on OS X to trigger the testing of _run_pythonw and _run respectively.

cc: @jni

jni (approved):

PR #1289, opened 2020-05-26
Add briefcase app bundles to release assets

tlambert03:
# Description

closes #1074 

This PR adds non-frozen apps bundled with [briefcase](https://github.com/beeware/briefcase), for all three platforms, to our github release action.  It was a joint effort with @GenevieveBuckley, with lots of fantastic discussion from everyone in in #1074.

In order to demonstrate the PR, I am temporarily commenting a few lines in our `make_release.yml` that must be uncommented before merging.  I'll also be tagging releases in my personal fork, so you can see the actual build artifacts and how they get added to the release notes. 

Note: this PR also includes changes from #1287, since they were required to get mac working... but if that proposal is rejected, I can undo it here and figure out another way.

##  Please test!
At this point the thing we need the most is for people to test these bundles.  The artifacts will show up in [releases in my fork](https://github.com/tlambert03/napari/releases) for now.  Since these commits will not be tagged in the napari repo (and the github action requires a tagged commit), they will _not_ build in this repository yet.

**Download links:** (updated 6/20)
[Mac Bundle](https://github.com/tlambert03/napari/suites/821133320/artifacts/9061283)
[Windows Bundle](https://github.com/tlambert03/napari/suites/821133320/artifacts/9061284)
[Linux Bundle](https://github.com/tlambert03/napari/suites/821133320/artifacts/9061282)

###  Current status:

**Mac**: should be working without the `open ...` trick in the terminal anymore (though you will, as usual, need to go into security and say "open anyway").  The critical thing there was making sure that the `_qt_resources.py` is built and bundled with the app _before_ code signing... and a check for that has been added to the github action.  Also, as @GenevieveBuckley pointed out in https://github.com/GenevieveBuckley, we will likely also need to notarize the app for it to run on Catalina.  (But I haven't tested anything on with catalina yet...).

**Windows**: should be fully working now, and I've patched the app to open with the non-windowed python.exe: which means a somewhat ugly cmd window will open up in addition to the napari app, but that is the compromise at the moment to have the built in ipython console also work.  Note it's very slow to start up the first time.  (actually, having that blank command window there is kinda nice in light of that because it shows you at least something is happening 😂)

**Linux**: Building fine (note, requires a ubuntu-16.04 VM on github to build), but the AppImage is not yet running, due to an upstream bug in the ctypes implementation in the briefcase Linux support package (so currently numpy will not work).  I _think_ I have a fix for that, and I submitted a PR at https://github.com/beeware/Python-Linux-support/pull/1.

[...]

PR #1599, opened 2020-09-01
lock briefcase version

liu-ziyang:
# Description
Fix ubuntu build failure by locking the briefcase version, this works for now but is certainly a bigger discussion of whether briefcase is the right call in longer-term if we can't use their up-to-date version release.

I tried to resolve the underlying issue, but no luck after quite a few days, the issue is quite deep in dependency linkings, and hard to debug in itself, and given that the Linux build doesn't work within docker mode for us, it is even harder to lock down the build environment.

## Type of change

# References
https://github.com/napari/napari/pull/1289

# How has this been tested?
- all tests pass with my change: https://github.com/ziyangczi/napari/runs/1054182098?check_suite_focus=true

## Final checklist:

### DIFF
--- a/.github/workflows/make_bundle.yml
+++ b/.github/workflows/make_bundle.yml
@@ -38,7 +38,7 @@ jobs:
       - name: Install Dependencies
         run: |
           python -m pip install --upgrade pip 
-          python -m pip install briefcase tomlkit wheel
+          python -m pip install briefcase==0.3.1 tomlkit wheel
           python -m pip install -e .[pyside2]
       - name: get tag
         shell: bash


sofroniewn:
Hi @ziyangczi, thanks for making this fix. Are there any issues on the briefcase repo/ briefcase release notes indicating what changed that you can link to? I'd imagine the briefcase folks would be pretty interested to know that something broke on their latest release, but is fixed on an earlier one. It could also help anyone in the community who comes across this understand why we pinned. Thanks for digging in!!

liu-ziyang:
> Hi @ziyangczi, thanks for making this fix. Are there any issues on the briefcase repo/ briefcase release notes indicating what changed that you can link to? I'd imagine the briefcase folks would be pretty interested to know that something broke on their latest release, but is fixed on an earlier one. It could also help anyone in the community who comes across this understand why we pinned. Thanks for digging in!!

I didn't find anything explicitly catchy, that being said my experience with c library linking is fairly limited, but we can certainly let them know about this

tlambert03 (approved):
I don't mind pinning at all, particularly since this is just for the process of bundling, and no downstream users need to worry about conflicts.

sofroniewn (approved):
Ok great, this sounds good to me, thanks @ziyangczi @tlambert03 I will merge now! We can always follow up with briefcase later

GenevieveBuckley:
@ziyangczi would you be able to make an issue over at the briefcase repo in the next day or two? PyConAU is this Fri-Sat-Sun (we're a day ahead of your timezone) and I might be able to have a conversation with some of the briefcase devs about it.

GenevieveBuckley:
Is the problem we're having similar to https://github.com/beeware/briefcase/issues/470

tlambert03:
> Is the problem we're having similar to [beeware/briefcase#470](https://github.com/beeware/briefcase/issues/470)

Yes! Great find!
Here’s our error
https://github.com/napari/napari/runs/1042199372?check_suite_focus=true

GenevieveBuckley:
I have a "hello world" example of a PySide2 project packaged with briefcase here: https://github.com/GenevieveBuckley/hellopyside 

Maybe we could use that to make a minimal & reproducible example for the briefcase devs - they've said that would help them to investigate. @tlambert03 do you think there's an easy way to combine it with the yaml for GitHub actions on CI?

tlambert03:
> @tlambert03 do you think there's an easy way to combine it with the yaml for GitHub actions on CI?

will submit a small PR to your repo that adds an build action

GenevieveBuckley:
FYI, I started work adding tests that briefcase can bundle some helloworld type apps on a range of different platforms and GUI framesworks here: https://github.com/beeware/briefcase/pull/484

This should give us more confidence in the future that changes to briefcase won't cause regressions in the functionality we care about.

PR #1604, opened 2020-09-02
exclude test_bundle from distribution

tlambert03:
# Description
`pytest --pyargs napari` was failing because `test_bundle.py` needs files that we do not distribute.  This PR just removes `test_bundl.py` from the distribution, since it doesn't make sense to include it.


sofroniewn (approved):
LGTM! Thanks for the quick fix, and thanks for updating the release notes. I will make a new RC after this!!

PR #1608, opened 2020-09-03
Skip bundle test when setup.cfg is missing

jni:
# Description

Bundle tests were failing in `pytest --pyargs napari` with wheel builds. This skips the test is the necessary files are not present. See the [Zulip extended discussion](https://napari.zulipchat.com/#narrow/stream/215289-release/topic/0.2E3.2E7/near/208840108) for reference.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,6 +1,5 @@
 include LICENSE
 recursive-include napari _tests/*.py
-exclude napari/_tests/test_bundle.py
 recursive-include napari/resources *.png *.jpeg *.jpg *.svg *.qss
 exclude napari/resources/_qt_resources*.py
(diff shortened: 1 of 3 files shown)


jni:
Confirmed that this works locally when using `pytest --pyargs napari` when installing both from sdist and bdist artifacts! 🎉

tlambert03 (approved):
Works for me!  Should we remove those first 6 comment lines in test_bundle now that the file technically is included in the dist?

jni:
Oops, thanks for noticing that @tlambert03! Fixed.

sofroniewn:
Thanks both, I will merge after tests pass, then make a final RC and then hopefully release!

PR #1610, opened 2020-09-03
Live tiff loader example

zeroth:
# Description
This PR adds an example to simulate live loading of data as it gets acquired by microscope. 
There are two programs:
-  live_tiffs_generator.py 
    This program generates tiffs in a given directory with specific time delay. 
    Simulation of microscope data acquisition.
- live_tiffs.py
   This program monitors the directory for new tiff files and loads them to Napari viewer as they get created in the directory.

## Type of change
- [ ] example

## Final checklist:


examples/live_tiffs.py
```diff
@@ -0,0 +1,79 @@
...
+
+            elif len(current_files):
+                last_file = sorted(current_files, key=alphanumeric_key)[-1]
+                current_files.remove(last_file)
+                files_to_process = current_files - processed_files
+            
+            for p in sorted(files_to_process, key=alphanumeric_key):
+                yield delayed(imread)(os.path.join(path, p))
+            else:
+                yield
```
jni, line 115: Can you add a comment explaining this for/else?

examples/live_tiffs.py
```diff
@@ -0,0 +1,79 @@
...
+            return
+
+        if viewer.layers:
+            layer = viewer.layers[0]
+            image_shape = layer.data.shape[1:]
+            image_dtype = layer.data.dtype
+            dask_image = da.from_delayed(
+                delayed_image,
+                shape=image_shape,
+                dtype=image_dtype,
```
jni: Can you run black on this file? I think this should fit...

```suggestion
                delayed_image, shape=image_shape, dtype=image_dtype
```

examples/live_tiffs.py
```diff
@@ -0,0 +1,79 @@
+"""
+Loads and Displays tiffs as they get generated in the specific directory.
+Trying to simulate the live display of data as it gets acquired by microscope.
```
jni: Add comment that this should be run together with `live_tiffs_generator.py`

examples/live_tiffs.py
```diff
@@ -0,0 +1,79 @@
...
+    # pass a directory to monitor or it will monitor current directory.
+    path = sys.argv[1] if len(sys.argv) > 1 else '.' 
+    path = os.path.abspath(path)
+    end_of_experiment = 'final.log'
+
+    def append(delayed_image):
+        if delayed_image is None:
+            return
+
+        if viewer.layers:
```
jni, line 43: Add a bit of narrative? Actually maybe here we should do `if not viewer.layers:  # first run, no layer added yet` ... `else:  # layer is present, append to its data`

examples/live_tiffs.py
```diff
@@ -0,0 +1,79 @@
...
+    @thread_worker(connect={'yielded': append})
+    def watch_path(path):
+        current_files = set()
+        processed_files = set()
+        while True:
+            files_to_process = set()
+            current_files = set(os.listdir(path))
+            if end_of_experiment in current_files:
+                files_to_process = current_files - processed_files
+                files_to_process.remove(end_of_experiment)
```
jni, line 102: This seems clunky because at this point, we should close the worker, rather than just ignore it, no?

examples/live_tiffs_generator.py
```diff
@@ -0,0 +1,39 @@
...
+
+def main(argv=sys.argv[1:]):
+    args = parser.parse_args(argv)
+    outdir = args.outdir
+    sleep_time = args.sleep_time
+    n = args.n
+    fractions = np.linspace(0.05, 0.5, n)
+    os.makedirs(outdir, exist_ok=True)
+    for i, f in enumerate(fractions):
+        curr_vol = 255 * data.binary_blobs(
```
jni, line 39: Add a comment explaining what this is generating

jni (changes requested):
@zeroth I've made some comments where I think the code needs some clarification. The examples are meant to be instructive to new users, so there should be a fair bit of comments explaining (a) the overarching idea, and (b) what each bit of code is doing.

Other than that, as you know I ❤️ this example. =)

jni:
@zeroth kind ping

examples/live_tiffs.py
```diff
@@ -0,0 +1,79 @@
...
+            image = delayed_image.compute()
+            dask_image = da.from_delayed(
+                delayed_image,
+                shape=image.shape,
+                dtype=image.dtype,
+            ).reshape((1,) + image.shape)
+            layer = viewer.add_image(dask_image, rendering='attenuated_mip')
+
+        if viewer.dims.point[0] >= layer.data.shape[0] - 2:
+            viewer.dims.set_point(0, layer.data.shape[0] - 1)
```
GenevieveBuckley, line 66: I like moving the slider to always display the most recent image, but why do you need an if statement here?

examples/live_tiffs.py
```diff
@@ -0,0 +1,79 @@
...
+import napari
+from napari.qt import thread_worker
+
+
+with napari.gui_qt():
+    viewer = napari.Viewer(ndisplay=3)
+    # pass a directory to monitor or it will monitor current directory.
+    path = sys.argv[1] if len(sys.argv) > 1 else '.' 
+    path = os.path.abspath(path)
+    end_of_experiment = 'final.log'
```
GenevieveBuckley, line 31: I think most people won't have a handy "final" file that is generated (eg: that's not typical for data generated on any of the instruments in our lab at all). When I wrote a directory watcher for napari I made sure that the user could cancel watching with Ctrl+C in the terminal. That might be a nice thing to add to this example as well.

examples/live_tiffs.py
```diff
@@ -0,0 +1,79 @@
...
+                delayed_image,
+                shape=image_shape,
+                dtype=image_dtype,
+            ).reshape((1,) + image_shape)
+            layer.data = da.concatenate(
+                (layer.data, dask_image), axis=0
+            )
+        else:
+            image = delayed_image.compute()
+            dask_image = da.from_delayed(
```
GenevieveBuckley: This variable name is potentially confusing, since dask_image is also the name of a package (eg: `import dask_image`). People new to dask in particular are likely to get tripped up by that.

examples/live_tiffs.py
```diff
@@ -0,0 +1,79 @@
...
+                delayed_image,
+                shape=image_shape,
+                dtype=image_dtype,
+            ).reshape((1,) + image_shape)
+            layer.data = da.concatenate(
+                (layer.data, dask_image), axis=0
+            )
+        else:
+            image = delayed_image.compute()
```
GenevieveBuckley, line 54: Isn't it counterproductive to be calling compute here?

examples/live_tiffs.py
```diff
@@ -0,0 +1,79 @@
...
+            if end_of_experiment in current_files:
+                files_to_process = current_files - processed_files
+                files_to_process.remove(end_of_experiment)
+
+            elif len(current_files):
+                last_file = sorted(current_files, key=alphanumeric_key)[-1]
+                current_files.remove(last_file)
+                files_to_process = current_files - processed_files
+            
+            for p in sorted(files_to_process, key=alphanumeric_key):
```
GenevieveBuckley, line 112: It looks like a key assumption here is that files will be written to disk with filenames that are ordered alphanumerically? I think that assumption is fine, but should be made clear in the docstring.

examples/live_tiffs.py
zeroth: This is just an example we can have multiple way of how we treminate the program potentially want the way to stop the `while: True` I am adding that in my next commit.  The `final.log` file is generated by the `live_tiffs_generator.py` at the moment. But, yes you are write that we can have user interupted way to do that as well. Ctrl + C sound terminating the programs and I dont think I as a user want to terminate the programs after the aqusition I rather look at the data and once all done I will close the program.

examples/live_tiffs.py
zeroth: Thanks for pointing that I will change it.

examples/live_tiffs.py
zeroth: well the input is dask delayed function. As far as I know to get the image shape we have to call compute(). Please correct me if I am wrong.

examples/live_tiffs.py
zeroth: Yes, lots of doc is missing writting that, thanks for the pointers. :)

examples/live_tiffs.py
```diff
@@ -0,0 +1,79 @@
...
+                dtype=image.dtype,
+            ).reshape((1,) + image.shape)
+            layer = viewer.add_image(dask_image, rendering='attenuated_mip')
+
+        if viewer.dims.point[0] >= layer.data.shape[0] - 2:
+            viewer.dims.set_point(0, layer.data.shape[0] - 1)
+   
+  
+    @thread_worker(connect={'yielded': append})
+    def watch_path(path):
```
GenevieveBuckley, line 69: Are you trying to avoid using a directory watcher like [watchdog](https://github.com/gorakhargosh/watchdog) because you don't want another dependency for this example?

zeroth: Yes, thats the main reason.  The other reason is I tried watchdog it has its own thread pool going on that just doesn't want to play along with Qt thereads (Napari Main Thread). I had QFileSystemWatcher based example as well but @jni suggested to go with as minila as possible.

GenevieveBuckley: Oh really? That's potentially worrying, I don't remember having that particular problem.

examples/live_tiffs.py
GenevieveBuckley: I think your code would be a lot simpler if you used the [dask-image imread function](http://image.dask.org/en/latest/dask_image.imread.html) for each filename (which would figure out the shape for you), and then [concatenated or stacked](https://docs.dask.org/en/latest/array-stack.html) those dask arrays together.

examples/live_tiffs.py
jni: @GenevieveBuckley one reason for this is that (a) we couldn't open the last file because it might still be getting data written to it (so the program was crashing), so (b) we would never open the final file in a series this way. It's not a huge deal, but this was the simplest thing to check for. Another option is to check the modification time for each file and only add it to the list if it hasn't been touched for, say, 1s.

examples/live_tiffs.py
jni: @GenevieveBuckley it's super annoying to have the slider move to the end if you are scrolling through the middle of the stack. This is akin to terminal output scrolling only when you are at the bottom, but not if you scroll back to check something in the middle of the output.

@zeroth probably worth adding a comment above the if statement explaining this.

examples/live_tiffs.py
jni: I'm ok with adding that logic in a future PR or adding it in this one, @zeroth

examples/live_tiffs.py
jni: @GenevieveBuckley I think we actually want to move to using `da.map_blocks` given my demo at the Friday meeting (specifically re https://github.com/dask/dask/issues/5913). I'm ok with doing this in a later PR, or using `dask_image`. (Possibly `dask_image` itself could use `map_blocks`.) Renaming the `dask_image` variable is probably a good idea either way though. =)

examples/live_tiffs.py
zeroth: @GenevieveBuckley Yes, tbh I didnt give much of time on watchdog as it wasn't very simple to get it working with Napri. I am sure we can get it working just that will be lots of code. In current version we can see the important part of loading the files as the get generated. For sure if we want to make it a more robust tool we can do few more things.

examples/live_tiffs.py
zeroth: @jni well I think this example makes it simple to understand. so maybe we can make it robust once we have some better way of checking the file state.

examples/live_tiffs.py
```diff
@@ -0,0 +1,129 @@
...
+        Parameters
+        ----------
+        delayed_image : dask.delayed function object
+        """
+        if delayed_image is None:
+            return
+
+        if viewer.layers:
+             # first run, no layer added yet
+            layer = viewer.layers[0]
```
jni, line 45: @zeroth you've got the comments backwards now (see my suggestion "if **not** viewer.layers"). Now it's like "no layers added — grab layers[0]" 😂

jni (approved):

PR #1612, opened 2020-09-03
Bump numpy version

Czaki:
# Description

Bump numpy version 1.17 because nan parameters of nan_to_num is introduced in this version. 
Bug introduced in #1360

## Type of change

# References

https://numpy.org/doc/stable/reference/generated/numpy.nan_to_num.html

https://github.com/napari/napari/blob/ea0fc9ca29d730808b1a66a8710679ee0fbe7ab5/napari/components/layerlist.py#L230-L231

https://travis-ci.com/github/4DNucleome/PartSeg/jobs/380612109 line 663
# How has this been tested?

## Final checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,7 @@ requires = [
   "IPython>=7.7.0",
   "napari-plugin-engine>=0.1.5",
   "napari-svg>=0.1.3",
-  "numpy>=1.10.0",
+  "numpy>=1.17.0",
   "numpydoc>=0.9.2",
   "Pillow!=7.1.0,!=7.1.1",
   "psutil>=5.0",
--- a/setup.cfg
+++ b/setup.cfg
@@ -47,7 +47,7 @@ install_requires =
     IPython>=7.7.0
     napari-plugin-engine>=0.1.5
     napari-svg>=0.1.3
-    numpy>=1.10.0
+    numpy>=1.17.0
     numpydoc>=0.9.2
     Pillow!=7.1.0,!=7.1.1  # not a direct dependency, but 7.1.0 and 7.1.1 broke imageio
     psutil>=5.0


jni:
Gah. Thank you for spotting this @Czaki.

According to [NEP-29](https://numpy.org/neps/nep-0029-deprecation_policy.html), we should continue to support 1.16 until January. So I think the correct fix is to bump to 1.16 and write our own nan_to_num.

We should also add a "minimum requirements" build to our test matrix to avoid these issues in the future. We do this with an env var in scikit-image, see here:

https://github.com/scikit-image/scikit-image/blob/73a8dbf0128e8832b15d86c88c9304bc1280aa71/tools/travis/before_install.sh#L60-L64

Czaki:
Ok. I created #1613 PR which should work with 1.16.

I do not see sh script in napari and do not familiar with cirrus yaml so cannot fast prepare such task. But simple If in yaml should be enough.

Czaki:
```python 
import os
from configparser import ConfigParser

config = ConfigParser()
config.read(os.path.join(os.path.dirname(__file__), "..", "setup.cfg"))

minimum_requirements = config["options"]["install_requires"].replace(">=", "==")

with open(os.path.join(os.path.dirname(__file__), "minimal-req.txt"), "w") as f_p:
    f_p.write(minimum_requirements)
    f_p.write("\n")
```

I tes nowt such code in my project to generate minimum requirements during test base on setup.cfg

sofroniewn:
Ok, thanks for the fix in #1613, agreed testing minimum versions sounds good, and I'm very open to different ways to do it in our CI, but that script from @Czaki looks like a good place to start

Czaki:
new version of script (with python dependecies)

```python
import os
from configparser import ConfigParser

config = ConfigParser()
config.read(os.path.join(os.path.dirname(__file__), "..", "setup.cfg"))

minimum_requirements = config["options"]["install_requires"].replace(">=", "==")
minimum_qt = config["options.extras_require"]["pyqt"].replace(">=", "==")

with open(os.path.join(os.path.dirname(__file__), "minimal-req.txt"), "w") as f_p:
    f_p.write(minimum_requirements)
    f_p.write(minimum_qt)
    f_p.write("\n")
```

sofroniewn:
This can now be closed as #1613 and #1617 will supersede it

PR #1613, opened 2020-09-03
fix usage nan_to_num to work with numpy 1.16

Czaki:
# Description

alternative solution to #1612 

## Type of change

# References

# How has this been tested?

## Final checklist:

@jni this should work with numpy 1.16.


jni (approved):
Awesome. 🎉 Thank you @Czaki!

sofroniewn (approved):
LGTM! And thanks for that catch

PR #1615, opened 2020-09-03
increase screenshot performance

kne42:
Remove unnecessary double call of `QImg2array`.

### DIFF
--- a/napari/_qt/qt_viewer.py
+++ b/napari/_qt/qt_viewer.py
@@ -26,7 +26,7 @@ from ..utils.key_bindings import components_to_key_combo
 from ..utils.theme import template
 from .dialogs.qt_about_key_bindings import QtAboutKeyBindings
 from .tracing.qt_performance import QtPerformance
-from .utils import circle_pixmap, square_pixmap
+from .utils import QImg2array, circle_pixmap, square_pixmap
 from .widgets.qt_dims import QtDims
 from .widgets.qt_layerlist import QtLayerList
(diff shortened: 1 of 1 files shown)


tlambert03:
This is a duplicate of #1109... @sofroniewn mentioned that he was a bit wary of this approach at the time, but it's been a while I can't remember the details.

sofroniewn:
> @sofroniewn mentioned that he was a bit wary of this approach at the time, but it's been a while I can't remember the details.

I tried pretty hard to make this work a long time ago (maybe around the vispy 0.5->0.6 transition), but it was always buggy - i.e. didn't handle things like opacity, z-ordering, and blending correctly. I think the offscreen rendering is a pretty different code path from what we normally use, so taking the pixels from the canvas frame seemed easiest at the time. I know that that vispy was working on this too, and so maybe everything is fixed now, in which case it's ok - see my comments here https://github.com/napari/napari/pull/1109#issuecomment-608506085. So I definitely don't have any objections, but I just want us to check (and maybe add tests) for all the necessary usecases - for example does `add_shapes.py` work as expected? If that works well, that's already a great step forward!!!

kne42:
Huh, that's weird, it turns out that `canvas.render()` is different from `QImg2array(canvas.native.grabFramebuffer()`, and I have no idea why. It's under the same circumstances as all the other bugs, e.g. when you have a transparent layer. There's still another way to slightly increase performance so I'll change this PR to that.

sofroniewn (approved):
LGTM! Thanks!!

PR #1617, opened 2020-09-04
Add test matrix entry to test minimum requirements

jni:
# Description

This was 10x more painful than it needed to be. 😂 But I'm there. 🤞 

## Type of change

# References

# How has this been tested?

## Final checklist:


sofroniewn (approved):
This looks good to me! Still failing, but approach looks good so I am approving, so @jni can merge when ready

jni:
So, I tried bumping the Qt console slowly to the current version but the tests still fail, so it's not that. I'm going to take a break, will revisit later. Next thing to try is probably ipykernel.

sofroniewn:
Some of the errors here are the same 
```python-traceback
/root/conda/lib/python3.7/site-packages/napari_svg/hook_implementations.py:99: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
xml_list = [<Element 'image' at 0x7fd4a5e7e470>, <Element 'image' at 0x7fd4a5e7e650>, <Element 'circle' at 0x7fd4a5e7e770>, <Element 'circle' at 0x7fd4a5e7e0b0>, <Element 'circle' at 0x7fd4a5e7e2f0>, <Element 'circle' at 0x7fd4a5e7e4d0>, ...]
extrema = array([[ 0.,  0.],
       [20., 20.]])
    def xml_to_svg(xml_list, extrema):
        """Convert a list of xml into an SVG string.
        Parameters
        ----------
        xml : list of xml.etree.ElementTree.Element
            List of a xml elements in the svg specification.
        extrema : array (2, 2)
            Extrema of data, specified as a minumum then maximum of the (x, y)
            coordinates. Used to specify the view box of SVG canvas.
        Returns
        ----------
        svg : string
            SVG representation of the layer.
        """
>       extrema = np.nan_to_num(extrema, nan=0)
E       TypeError: nan_to_num() got an unexpected keyword argument 'nan'
```

But because of code in `napari-svg`. I can fix and upgrade napari-svg!

sofroniewn:
The other ones are 

```python-traceback
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
napari/_qt/qt_viewer.py:229: in console
    self.console = QtConsole({'viewer': self.viewer})
napari/_qt/widgets/qt_console.py:96: in __init__
    self.kernel_client = kernel_client
/root/conda/lib/python3.7/site-packages/qtconsole/base_frontend_mixin.py:63: in kernel_client
    if kernel_client.channels_running:
/root/conda/lib/python3.7/site-packages/jupyter_client/client.py:141: in channels_running
    self.control_channel.is_alive())
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <qtconsole.inprocess.QtInProcessKernelClient object at 0x7fd4c868caf0>
    @property
    def control_channel(self):
        """Get the control channel object for this kernel."""
        if self._control_channel is None:
            url = self._make_url('control')
            self.log.debug("connecting control channel to %s", url)
            socket = self.connect_control(identity=self.session.bsession)
            self._control_channel = self.control_channel_class(
>               socket, self.session, self.ioloop
            )
E           TypeError: ChannelABC() takes no arguments
/root/conda/lib/python3.7/site-packages/jupyter_client/client.py:200: TypeError
```

Which I've seen before and I think is in `ipykernel`

sofroniewn:
The `Check Manifest` is also failing

```
check-manifest
lists of files in version control and sdist do not match!
missing from sdist:
  tools/minreq.py
suggested MANIFEST.in rules:
  recursive-include tools *.py
```

tlambert03:
cool test!

> The `Check Manifest` is also failing
>
> suggested MANIFEST.in rules:
>   recursive-include tools *.py


if you don't want to include this file in the distribution, to make `check-manifest` happy you can either add an explicit exclude to MANIFEST.in (like the others at the bottom of that file), or add it to the `[tool.check-manifest]` ignore section in `pyproject.toml`

Czaki:
Nice job. I like idea of update `setup.cfg`. This should be ported to other napari package (like napari-svg), to observe problem as fast as possible.

sofroniewn:
> add it to the [tool.check-manifest] ignore section in pyproject.toml

Thanks for that suggestion @tlambert03 I have gone with this option. If @jni prefers other way no problem if he changes it!

I have also released a new version of `napari-svg` `0.1.4` that should now work with our minimal numpy version.

I still have to fix `ipykernel`, but will wait to see if everything else works first

sofroniewn:
We also have an issue with `test_multichannel_dask_array`

```python-traceback
_________________________ test_multichannel_dask_array _________________________
    def test_multichannel_dask_array():
        """Test adding multichannel dask array."""
        viewer = ViewerModel()
        np.random.seed(0)
        data = da.random.random((2, 10, 10, 5))
        viewer.add_image(data, channel_axis=0)
        assert len(viewer.layers) == data.shape[0]
        for i in range(data.shape[0]):
            assert viewer.layers[i].data.shape == data.shape[1:]
>           assert isinstance(viewer.layers[i].data, da.Array)
E           AssertionError: assert False
E            +  where False = isinstance(array([[[0.18722244, 0.3598777 , 0.62570796, 0.56056966, 0.45481825],\n        [0.13954979, 0.14048273, 0.11438004, 0.9....98847445, 0.22780785, 0.54694169, 0.7296056 ],\n        [0.5356861 , 0.66645397, 0.5906914 , 0.59228675, 0.69606446]]]), <class 'dask.array.core.Array'>)
E            +    where array([[[0.18722244, 0.3598777 , 0.62570796, 0.56056966, 0.45481825],\n        [0.13954979, 0.14048273, 0.11438004, 0.9....98847445, 0.22780785, 0.54694169, 0.7296056 ],\n        [0.5356861 , 0.66645397, 0.5906914 , 0.59228675, 0.69606446]]]) = <Image layer 'Image' at 0x7fba567d8350>.data
E            +    and   <class 'dask.array.core.Array'> = da.Array
napari/components/_tests/test_multichannel.py:194: AssertionError
```

Looks like we need to change that assert/ update the dask dependence. Thoughts on this @tlambert03?

sofroniewn:
For the console errors, sounds like the recommendation here https://github.com/jupyter/jupyter_client/issues/531 is to use `ipykernel>5.2.0` https://github.com/jupyter/jupyter_client/issues/531#issuecomment-631763829. I will try bumping that to be the minimum version

sofroniewn:
It also looks like we get this test_bundle error ONLY on the `MIN_REQ:1` pass - see

```python-traceback
___________________________ test_bundle_requirements ___________________________
    @pytest.mark.skipif(
        not setup_file.is_file(),
        reason='Bundle not tested in source or wheel distributions',
    )
    def test_bundle_requirements():
        """Test that briefcase requirements are superset of setup.cfg requirements.
        """
        parser = configparser.ConfigParser()
        parser.read(setup_file)
        requirements = parser.get("options", "install_requires").splitlines()
        requirements = {r.split('#')[0].strip() for r in requirements if r}
        with open(root_dir / 'pyproject.toml', 'r') as f:
            pyproject = f.read()
        toml = tomlkit.parse(pyproject)
        toml_recs = set(toml['tool']['briefcase']['app']['napari']['requires'])
>       assert requirements.issubset(toml_recs)
E       AssertionError: assert False
E        +  where False = <built-in method issubset of set object at 0x7fba9423f7d0>({'IPython>=7.7.0', 'Pillow!=7.1.0,!=7.1.1', 'PyOpenGL>=3.1.0', 'PySide2==5.14.2.2', 'PyYAML>=5.1', 'appdirs>=1.4.4', ...})
E        +    where <built-in method issubset of set object at 0x7fba9423f7d0> = {'IPython==7.7.0', 'Pillow!=7.1.0,!=7.1.1', 'PyOpenGL==3.1.0', 'PyYAML==5.1', 'appdirs==1.4.4', 'cachey==0.2.1', ...}.issubset
napari/_tests/test_bundle.py:30: AssertionError
```

Does that make sense to you @tlambert03? Is something going on that would cause only a fail then?

tlambert03:
> Does that make sense to you @tlambert03? Is something going on that would cause only a fail then?

Ahh... yep that makes sense, that test is checking to make sure if the requirements _strings_ in `setup.cfg` are a literal subset of those in `pyproject.toml`.  If it's ok with everyone, I'd like to undo the change [requested in #1289](https://github.com/napari/napari/pull/1289#pullrequestreview-474717867), wherein we avoided dynamically populating the briefcase dependencies from `setup.cfg` ... which necessitated the new `test_bundle.py` to make sure that the two are in sync.  I think the annoyances of data duplication and `test_bundle` seem to be outweighing the benefits at this point, what do you think?
(though: if we just wanted to make this test pass here ... we could edit `test_bundle` to strip those strings of any `==`, `>=` requirement specifiers before comparison)

sofroniewn:
> If it's ok with everyone, I'd like to undo the change requested in #1289, wherein we avoided dynamically populating the briefcase dependencies from setup.cfg ... which necessitated the new test_bundle.py to make sure that the two are in sync. I think the annoyances of data duplication and test_bundle seem to be outweighing the benefits at this point, what do you think?

I'm ok with that, would there maybe still be an easy way to say overwrite the default ones? So I could make the bundle use this version of vispy or make the bundle without napari-svg, but if I don't do anything then I just get the same ones. How does that sound? cc @jni @ziyangczi

sofroniewn:
Good news is we're down to only two failing tests on the MIN_REQ path now! The dask assert one which looks easy to fix, and the bundle one which @tlambert03 is looking into addressing. So we should be good here soon

tlambert03:
> I'm ok with that, would there maybe still be an easy way to say overwrite the default ones? So I could make the bundle use this version of vispy or make the bundle without napari-svg, but if I don't do anything then I just get the same ones.

simplest way would just be for you to modify bundle.py directly while experimenting, but if you think that's something you're going to want to do a lot of, we could add a command line argument to `bundle.py` that takes a `pip install`-like string of requirements to override/include

sofroniewn:
> > I'm ok with that, would there maybe still be an easy way to say overwrite the default ones? So I could make the bundle use this version of vispy or make the bundle without napari-svg, but if I don't do anything then I just get the same ones.
>
> simplest way would just be for you to modify bundle.py directly while experimenting, but if you think that's something you're going to want to do a lot of, we could add a command line argument to bundle.py that takes a pip install-like string of requirements to override/include

My comment is more for the future/ hypothetical, I think @jni had a comment indicating something like that might be useful at some point.

My recommendation now would be to make the revert in a fresh PR as maintaining both lists is causing these difficulties right now. If @jni @ziyangczi have additional thoughts they can weigh in on that PR. This is certainly something we can iterate and improve on in the future, and doesn't have to be the last thing we try!

sofroniewn:
@jni can you fix the failing dask check, i tried but failed. then i think this is ready to merge!!

sofroniewn:
> @jni Use the type of the original array to check layer type

Dask check still failing :-(

Czaki:
I check on computer. There is problem with numpy. `numpy.take` from version 1.16 on python 3.6 return `numpy.array` instead of `dask.array`.
 
(line 368  in `add_layer_minix`)

jni:
Dang, LOL, I knew something was fishy about that failure as I didn't think dask would actually move that!

Unless anyone objects, I'm just going to skip that test in py3.6, since I don't want to complicate the logic in add_layer_mixin for a use case that will go away soon.

Czaki:
> Unless anyone objects, I'm just going to skip that test in py3.6, since I don't want to complicate the logic in add_layer_mixin for a use case that will go away soon.

this test or only this assert?

sofroniewn:
> Unless anyone objects, I'm just going to skip that test in py3.6, since I don't want to complicate the logic in add_layer_mixin for a use case that will go away soon.

I think it fails on py3.7 too - see the test in the matrix `
linux 3.7 MIN_REQ:1`, so it won't actually go away if we bump to py3.7, and this will be with us until we move on from numpy 1.16 no?

jni:
Oh dang, of course. 🤦 Hmmmmmmmm. Thoughts? Do we complicate the logic? Or do we skip for NumPy <= 1.16?

jni:
Right, it's coming back to me, this is NEP-18 implementation. iirc, you could opt-in to NEP-18 with an environment variable in 1.16! Should we do that? It does mean users will miss out if they are on 1.16 and don't have the right env set. But that doesn't seem to have affected anyone till now!

jni:
https://numpy.org/neps/nep-0018-array-function-protocol.html#implementation : 

> In NumPy 1.16, you need to set the environment variable `NUMPY_EXPERIMENTAL_ARRAY_FUNCTION=1` before importing NumPy to test NumPy function overrides.

Czaki:
so https://numpy.org/doc/1.16/reference/generated/numpy.take.html has incorrect information about return type.

jni:
@Czaki I guess none of the NumPy individual function docs include language about NEP-18 and dispatching, e.g. https://numpy.org/doc/stable/reference/generated/numpy.amin.html. It would be a major undertaking to fix them all, especially to do so in a way that is not confusing to pure NumPy users.

PR #1619, opened 2020-09-04
change bundle deps approach

tlambert03:
# Description
as discussed in https://github.com/napari/napari/pull/1617#issuecomment-687369172, this removes the duplicated requirements listed in the briefcase section of `pyproject.toml` (along with `test_bundle.py`) and grabs them from `setup.cfg`, packages can be added and/or overridden during bundling as follows:

```bash
python bundle.py --add ome-zarr 'PyQt5<5.15' 'napari-svg>=0.1.4'
```
this would _add_ the `ome-zarr` package, _change_ the version of `napari-svg` and (a special case for PyQt), _remove_ PySide2 and use PyQt5 instead.

## Type of change

- [x] change in bundling strategy

# How has this been tested?

- [x] test_bundle.py removed

## Final checklist:


pyproject.toml
```diff
@@ -16,42 +16,15 @@ author = "napari"
 url = "https://napari.org/"
 license = "BSD license"
 # version populated in bundle.py
+version = "0.3.7rc11.dev7"
```
sofroniewn:
```suggestion
version = "0.0.1"
```
If I understand correctly and this gets overwritten in the bundle can we go for a more neutral option here? I can just imagine someone being confused and wondering what was special about those numbers

sofroniewn (approved):
This looks good to me. I left one minor comment, but will let @jni weigh in before merge.

pyproject.toml
sofroniewn: It will also maybe make things more clear if the overwrite fails to happen correctly

tlambert03: Yep, sorry meant to change that.  But let’s make it 0.0.1 ... since briefcase on windows will fail with 0.0.0 (if someone tried to build directly without using bundle.py)

jni:
@tlambert03 I'm totally not sure whether this suggestion makes sense, but my impression is that packaging is moving to pyproject.toml as the One Place for all the build settings, including dependencies. So I'm wondering whether it would be possible/desirable to have *that* contain the master requirements list? Or is that still too annoying with the current tooling?

jni:
(I'm otherwise totally ok with the dynamic patching, makes a lot of sense.)

jni (approved):

jni:
(@tlambert03 if you think my proposal above is more suited to a future PR, feel free to merge as-is. I've approved the PR for that purpose.)

tlambert03:
> @tlambert03 I'm totally not sure whether this suggestion makes sense, but my impression is that packaging is moving to pyproject.toml as the One Place for all the build settings, including dependencies. So I'm wondering whether it would be possible/desirable to have _that_ contain the master requirements list? Or is that still too annoying with the current tooling?

"moving there" yes... "there" no.  the primary purpose of `pyproject.toml` is to specify the build _tools_ required, but not necessarily the config for the build tool.  I think some build tools like flit or poetry also let you put the full build config and runtime dependencies in `pyproject` but setuptools still requires either setup.py or setup.cfg.  But even if setuptools _did_ allow them in pyproject.toml, I think we would probably still need two different sections here, because briefcase is technically a completely different pep518-compliant build tool, and it has its own `[tool.briefcase.app.napari]` section of `pyproject.toml`... so I think we're stuck with the duplication problem regardless of where we place the main list.

jni:
> the primary purpose of pyproject.toml is to specify the build tools required, but not necessarily the config for the build tool.

Thanks!!! I don't think I'd made that distinction in my head yet!

tlambert03:
> > the primary purpose of pyproject.toml is to specify the build tools required, but not necessarily the config for the build tool.
> 
> Thanks!!! I don't think I'd made that distinction in my head yet!

👍  ... I was also quite confused about that for a while.  I found this blog article helpful: https://snarky.ca/what-the-heck-is-pyproject-toml/

PR #1622, opened 2020-09-06
Update colortransform when building texture

### LINKED ISSUE #1585: interaction between visibility and contrast limits
## 🐛 Bug

I've just found a funny bug where there appears to be an interaction between the visibility and contrast limits

## To Reproduce

1.  napari.view_image(np.random.random((200, 200)), contrast_limits=[0, 10])
2. move clim slider to be [0, 2]
3. toggle visibility off/ on
4. notice that after toggle image is very saturated. adjusting slider a tiny bit causes image to look good again

clim_bug

## Environment

last commit https://github.com/napari/napari/commit/62a046de2e5009f0c9137e3fd3999fbf902928ca

tlambert03:
# Description
fixes #1585
fixed upstream in https://github.com/vispy/vispy/pull/1911

## Type of change

# References

# How has this been tested?

## Final checklist:


jni (approved):
Thanks @tlambert03! Is there an easy way to test for this? Maybe with a screenshot?

tlambert03:
> Thanks @tlambert03! Is there an easy way to test for this? Maybe with a screenshot?

lemme give it a shot

tlambert03:
oh man... I don't even want to know why this doesn't work.  buuut, there seems to be some sort of race condition here making this very hard to test.  if I run the entire following code block at once, it passes, but if I run it in two chunks (where marked), or if I manually change the contrast limits, it fails

```python
    import numpy as np, napari
    viewer = napari.Viewer()
    layer = viewer.add_image(np.random.random((200, 200)), contrast_limits=[0, 10])
    visual = viewer.window.qt_viewer.layer_to_visual[layer]

    # if pause right here, then this fails, showing the weird contrast phenotype
    layer.contrast_limits = (0, 2)
    screen1 = viewer.screenshot()
    visual._on_data_change()  # this is the the problematic method when toggling visibility 
    screen2 = viewer.screenshot()
    assert np.allclose(screen1, screen2)
```

I also tried putting in a `qtbot.wait()` or a `time.sleep` there, but it still didn't fail the test on the current master.  So I'm actually not immediately sure how to get the failing test at the moment!

tlambert03:
ok, I found a test that works (fails on master, passes here).  It's not pretty – testing for screenshot equality ended up being weirder/harder than expected – but it gets the job done

sofroniewn (approved):
LGTM! Thanks!!

PR #1632, opened 2020-09-16
Big update of rendering explanation doc

pwinston:
# Description
First version of this doc was written before we had done any implementation work. This is a big update based on the first 4 months of work on the renderer.

## Type of change
- [x] Doc update

# Notes
At first I was going to include more system/code documentation in this doc. But I decided this works best as a generic "explanation" document. I think it can take someone who knows very little about rendering to understanding basically what we are doing and why.

We should do system documentation in a different file/format. More like the events and threading documentation.

### DIFF
--- a/docs/source/explanations/rendering.md
+++ b/docs/source/explanations/rendering.md
@@ -1,38 +1,23 @@
 # Rendering in napari
 
-This document outlines our plans for making napari's rendering non-blocking. We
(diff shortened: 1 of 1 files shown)


sofroniewn (approved):
This is great @pwinston - thanks for the update!! This doc really helps me understand the big picture of where the rendering is going and it is really exciting :-)

PR #1633, opened 2020-09-17
Do not ship bundle.py in source distribution

jni:
I noticed this when conda-forge started asking me to include tomlkit in our dependencies!

# Description
We don't need to include bundle.py in our source distribution, since it is an internal developer utility.


sofroniewn (approved):
Fine by me, will merge now!

PR #1634, opened 2020-09-17
New perfmon doc

pwinston:
# Description
* Added a new guide for the `napari.utils.perf` module.
* Renamed heading from **Events and Threading** to **Guides** since there are now 3 docs in that folder.
* Images are hyperlinks to the GitHub hosted images in the PR. A trick to avoid putting binaries into the repo. Maybe we will come up with a better image hosting approach later. GitHub could disable this out from under us someday.

Guide is just an invented name, I'm open to other suggestions. New website might change this.

## Type of change
- [x] New documentation

### DIFF
--- a/docs/source/events/index.rst
+++ b/docs/source/events/index.rst
@@ -6,6 +6,7 @@ Events and Threading
 
    event_loop
    threading
+   perfmon
 
 If you'd like to start customizing the behavior of napari, it pays to
 familiarize yourself with the concept of an Event Loop. For an introduction to
@@ -16,3 +17,7 @@ If you use napari to view and interact with the results of long-running
 computations, and would like to avoid having the viewer become unresponsive
 while you wait for a computation to finish, you may benefit from reading about
 :ref:`multithreading-in-napari`.
+
+Performance is a core feature of napari, and napari contains tools to help
+you monitor performance and diagnose and fix performance problems, see
+:ref:`perfmon`.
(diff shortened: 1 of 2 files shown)


pwinston:
This PR puts the new **Performance Monitoring** under the heading which right now is **Events and Threading**. Since we probably don't want to change the heading to **Events and Threading and Perfmon** I've changed it to say **Guides**. That is a strawman suggestion. I have no good ideas as to what it should really be called.

napari-guides

I would also suggest maybe rename "An Introduction to the Event Loop in napari" to be just "Event Loop" and "Multithreading in napari" to be just "Multithreading". But that's separate and totally optional.

pwinston:
Some images that we will use in the docs, hosted here for now:

The performance widget:

napari-perf-widget

A trace showing only Qt Events:

napari-perf-trace-1

A trace after adding `Labels.paint` method:

napari-perf-trace-2

pwinston:
I'm marking this as ready for review. It's a pretty good version 1. We could obviously add even more in the future. Maybe @kevinyamauchi can take a look. I'm not sure how I can assign him as a reviewer.

To view the formatted document I've been typing `make html` in the `napari docs` directory, then opening `napari/docs/build/html/index.html` then clicking into the new perfmon page.

sofroniewn (approved):
This looks great @pwinston I've given it a read and I think it's good to go. As the change to the `Guide` name won't effect the URL I think we're fine to merge this now as is. We can continue to refine these docs as more people use the perfmon!

PR #1635, opened 2020-09-19
Magic name handle attribute exception

### LINKED ISSUE #1630: napari.utils.naming.magic_name should return a default name on exception
## To Reproduce

1. Execute https://gist.github.com/mkitti/2f7c5fc3d3f8b0d15dd13f6d67b0e73d in Julia
2. On executing `napari.view_image(astronaut)`  the following error stack trace occurs

```python
AttributeError("'NoneType' object has no attribute 'f_code'")
  File "C:\Users\asdf\.julia\conda\3\lib\site-packages\napari\components\add_layers_mixin.py", line 225, in add_image
    return self.add_layer(layers.Image(data, **kwargs))
  File "C:\Users\asdf\.julia\conda\3\lib\site-packages\napari\layers\image\image.py", line 177, in __init__
    super().__init__(
  File "C:\Users\asdf\.julia\conda\3\lib\site-packages\napari\layers\intensity_mixin.py", line 22, in __init__
    super().__init__(*args, **kwargs)
  File "C:\Users\asdf\.julia\conda\3\lib\site-packages\napari\layers\base\base.py", line 138, in __init__
    name = magic_name(data, path_prefix=ROOT_DIR)
  File "C:\Users\asdf\.julia\conda\3\lib\site-packages\napari\utils\naming.py", line 70, in magic_name
    code = frame.f_code
```

## Expected behavior

The astronaut image should appear in the Napari viewer.

## Environment

napari: 0.3.6
Python: 3.8.3 (default, May 19 2020, 06:50:17) [MSC v.1916 64 bit (AMD64)]
Qt: 5.12.5
PyQt5: 5.12.3
NumPy: 1.19.1
SciPy: 1.5.0
Dask: 2.26.0
VisPy: 0.6.4

GL version: 4.6.0 - Build 26.20.100.7985
MAX_TEXTURE_SIZE: 16384

Plugins:
- napari_plugin_engine: 0.1.7
- svg: 0.1.3

I'm calling Napari from Julia. I installed Napari via conda-forge.

## Additional context

https://github.com/napari/napari/blob/ce6f226720e782e224408ad1f5471b364662591e/napari/utils/naming.py#L64-L69

## Workaround

`napari.view_image(astronaut,name="astronaut")` works

## Problem

The function `magic_name` uses the `inspect` package to navigate stack frames. However, it is possible that one of the stack frames could be `None`. This is not checked for nor is a possible exception caught. Is this case, `frame.f_back` may eventually be `None` and thus not have a `f_code` field.

https://github.com/napari/napari/blob/master/napari/utils/naming.py#L64-L69

https://docs.python.org/3/library/inspect.html#inspect.currentframe

<blockquote>
inspect.currentframe()
Return the frame object for the caller’s stack frame.

CPython implementation detail: This function relies on Python stack frame support in the interpreter, which isn’t guaranteed to exist in all implementations of Python. If running in an implementation without Python stack frame support this function returns None.
</blockquote>

## Possible solutions

1. Add exception handling and return a default name if one could not be found.
2. When iterating frames, check if each frame is `None`
3. Catch exception at a higher level and recommend the use of the `name` keyword argument if `magic_name` fails.

### LINKED ISSUE #1709: magic_name should work for all layer types
## 🐛 Bug

The `magic_name` function should work for all layer types, imo.

https://github.com/napari/napari/blob/5e8be0987b3ab691da5755380ce1dd798978b1c6/napari/utils/naming.py#L49

## To Reproduce

```python
viewer = napari.Viewer()

my_points = np.random.random((2, 10))
points_layer = viewer.add_points(my_points)

print(points_layer.name)
# expected output "my_points", got "data" instead
```
napari-magicname-screenshot1

```python
my_image = np.random.random((100, 100))
image_layer = viewer.add_image(my_image)

print(image_layer.name)
# expected output "my_image", got expected result
```
napari-magicname-screenshot2

## Expected behavior

I expect if I add a points layer where the name of the point data variable is "my_points", then the name of the napari layer should also be "my_points".

## Environment

```
# Name                    Version                   Build  Channel
alabaster                 0.7.12                   pypi_0    pypi
appdirs                   1.4.4                    pypi_0    pypi
argon2-cffi               20.1.0                   pypi_0    pypi
asciitree                 0.3.3                    pypi_0    pypi
async-generator           1.10                     pypi_0    pypi
atomicwrites              1.4.0                    pypi_0    pypi
attrs                     20.2.0                   pypi_0    pypi
babel                     2.8.0                    pypi_0    pypi
backcall                  0.2.0                    pypi_0    pypi
black                     19.10b0                  pypi_0    pypi
bleach                    3.2.1                    pypi_0    pypi
ca-certificates           2020.7.22                     0  
cachey                    0.2.1                    pypi_0    pypi
certifi                   2020.6.20                py37_0  
cffi                      1.14.3                   pypi_0    pypi
cfgv                      3.2.0                    pypi_0    pypi
chardet                   3.0.4                    pypi_0    pypi
check-manifest            0.44                     pypi_0    pypi
click                     7.1.2                    pypi_0    pypi
colorama                  0.4.3                    pypi_0    pypi
commonmark                0.9.1                    pypi_0    pypi
cycler                    0.10.0                   pypi_0    pypi
dask                      2.30.0                   pypi_0    pypi
decorator                 4.4.2                    pypi_0    pypi
defusedxml                0.6.0                    pypi_0    pypi
distlib                   0.3.1                    pypi_0    pypi
docutils                  0.16                     pypi_0    pypi
entrypoints               0.3                      pypi_0    pypi
fasteners                 0.15                     pypi_0    pypi
filelock                  3.0.12                   pypi_0    pypi
flake8                    3.8.3                    pypi_0    pypi
freetype-py               2.2.0                    pypi_0    pypi
fsspec                    0.8.3                    pypi_0    pypi
heapdict                  1.0.1                    pypi_0    pypi
identify                  1.5.5                    pypi_0    pypi
idna                      2.10                     pypi_0    pypi
imageio                   2.9.0                    pypi_0    pypi
imagesize                 1.2.0                    pypi_0    pypi
importlib-metadata        2.0.0                    pypi_0    pypi
iniconfig                 1.0.1                    pypi_0    pypi
ipykernel                 5.3.4                    pypi_0    pypi
ipython                   7.18.1                   pypi_0    pypi
ipython-genutils          0.2.0                    pypi_0    pypi
ipywidgets                7.5.1                    pypi_0    pypi
jedi                      0.17.2                   pypi_0    pypi
jinja2                    2.11.2                   pypi_0    pypi
jsonschema                3.2.0                    pypi_0    pypi
jupyter                   1.0.0                    pypi_0    pypi
jupyter-client            6.1.7                    pypi_0    pypi
jupyter-console           6.2.0                    pypi_0    pypi
jupyter-core              4.6.3                    pypi_0    pypi
jupyterlab-pygments       0.1.2                    pypi_0    pypi
kiwisolver                1.2.0                    pypi_0    pypi
markupsafe                1.1.1                    pypi_0    pypi
matplotlib                3.3.2                    pypi_0    pypi
mccabe                    0.6.1                    pypi_0    pypi
mistune                   0.8.4                    pypi_0    pypi
monotonic                 1.5                      pypi_0    pypi
napari                    0.3.9.dev53+gb230dc22           dev_0    <develop>
napari-plugin-engine      0.1.8                    pypi_0    pypi
napari-svg                0.1.4                    pypi_0    pypi
nbclient                  0.5.0                    pypi_0    pypi
nbconvert                 6.0.7                    pypi_0    pypi
nbformat                  5.0.7                    pypi_0    pypi
nest-asyncio              1.4.1                    pypi_0    pypi
networkx                  2.5                      pypi_0    pypi
nodeenv                   1.5.0                    pypi_0    pypi
notebook                  6.1.4                    pypi_0    pypi
numcodecs                 0.7.2                    pypi_0    pypi
numpy                     1.19.2                   pypi_0    pypi
numpydoc                  1.1.0                    pypi_0    pypi
openssl                   1.1.1h               he774522_0  
packaging                 20.4                     pypi_0    pypi
pandas                    1.1.3                    pypi_0    pypi
pandocfilters             1.4.2                    pypi_0    pypi
parso                     0.7.1                    pypi_0    pypi
pathspec                  0.8.0                    pypi_0    pypi
pep517                    0.8.2                    pypi_0    pypi
pickleshare               0.7.5                    pypi_0    pypi
pillow                    7.2.0                    pypi_0    pypi
pip                       20.2.3                   py37_0  
pluggy                    0.13.1                   pypi_0    pypi
pooch                     1.2.0                    pypi_0    pypi
pre-commit                2.7.1                    pypi_0    pypi
prometheus-client         0.8.0                    pypi_0    pypi
prompt-toolkit            3.0.7                    pypi_0    pypi
psutil                    5.7.2                    pypi_0    pypi
py                        1.9.0                    pypi_0    pypi
pycodestyle               2.6.0                    pypi_0    pypi
pycparser                 2.20                     pypi_0    pypi
pyflakes                  2.2.0                    pypi_0    pypi
pygments                  2.7.1                    pypi_0    pypi
pyopengl                  3.1.5                    pypi_0    pypi
pyparsing                 2.4.7                    pypi_0    pypi
pyqt5                     5.14.2                   pypi_0    pypi
pyqt5-sip                 12.8.1                   pypi_0    pypi
pyrsistent                0.17.3                   pypi_0    pypi
pytest                    6.1.1                    pypi_0    pypi
pytest-faulthandler       2.0.1                    pypi_0    pypi
pytest-ordering           0.6                      pypi_0    pypi
pytest-qt                 3.3.0                    pypi_0    pypi
pytest-timeout            1.4.2                    pypi_0    pypi
python                    3.7.9                h60c2a47_0  
python-dateutil           2.8.1                    pypi_0    pypi
pytz                      2020.1                   pypi_0    pypi
pywavelets                1.1.1                    pypi_0    pypi
pywin32                   228                      pypi_0    pypi
pywinpty                  0.5.7                    pypi_0    pypi
pyyaml                    5.3.1                    pypi_0    pypi
pyzmq                     19.0.2                   pypi_0    pypi
qtconsole                 4.7.7                    pypi_0    pypi
qtpy                      1.9.0                    pypi_0    pypi
recommonmark              0.6.0                    pypi_0    pypi
regex                     2020.9.27                pypi_0    pypi
requests                  2.24.0                   pypi_0    pypi
scikit-image              0.17.2                   pypi_0    pypi
scipy                     1.5.2                    pypi_0    pypi
send2trash                1.5.0                    pypi_0    pypi
setuptools                50.3.0           py37h9490d1a_1  
six                       1.15.0                   pypi_0    pypi
snowballstemmer           2.0.0                    pypi_0    pypi
sphinx                    3.0.1                    pypi_0    pypi
sphinx-rtd-theme          0.4.3                    pypi_0    pypi
sphinxcontrib-applehelp   1.0.2                    pypi_0    pypi
sphinxcontrib-devhelp     1.0.2                    pypi_0    pypi
sphinxcontrib-htmlhelp    1.0.3                    pypi_0    pypi
sphinxcontrib-jsmath      1.0.1                    pypi_0    pypi
sphinxcontrib-qthelp      1.0.3                    pypi_0    pypi
sphinxcontrib-serializinghtml 1.1.4                    pypi_0    pypi
sqlite                    3.33.0               h2a8f88b_0  
terminado                 0.9.1                    pypi_0    pypi
testpath                  0.4.4                    pypi_0    pypi
tifffile                  2020.10.1                pypi_0    pypi
toml                      0.10.1                   pypi_0    pypi
toolz                     0.11.1                   pypi_0    pypi
tornado                   6.0.4                    pypi_0    pypi
traitlets                 5.0.4                    pypi_0    pypi
typed-ast                 1.4.1                    pypi_0    pypi
typing-extensions         3.7.4.3                  pypi_0    pypi
urllib3                   1.25.10                  pypi_0    pypi
vc                        14.1                 h0510ff6_4  
virtualenv                20.0.33                  pypi_0    pypi
vispy                     0.6.5                    pypi_0    pypi
vs2015_runtime            14.16.27012          hf0eaf9b_3  
wcwidth                   0.2.5                    pypi_0    pypi
webencodings              0.5.1                    pypi_0    pypi
wheel                     0.35.1                     py_0  
widgetsnbextension        3.5.1                    pypi_0    pypi
wincertstore              0.2                      py37_0  
wrapt                     1.12.1                   pypi_0    pypi
xarray                    0.16.1                   pypi_0    pypi
zarr                      2.5.0                    pypi_0    pypi
zipp                      3.3.0                    pypi_0    pypi
zlib                      1.2.11               h62dcd97_4  
```

mkitti:
# Description
Addresses #1630 by catching an `AttributeException` in magic_name

## Type of change

- [x ] Bug-fix (non-breaking change which fixes an issue)

# References

Closes #1630
Closes #1709 
Includes #1714 

# How has this been tested?

- [x] Test results are the same before and after my changes
- [x] Basic viewing succeeds in IPython and Julia REPL

## Final checklist:


mkitti:
@jni, I found time for a PR.

napari/utils/naming.py
```diff
@@ -61,16 +64,25 @@ def magic_name(value, *, path_prefix):
     name : str or None
         Name of the variable, if found.
     """
-    frame = inspect.currentframe().f_back
-    code = frame.f_code
-
-    while code.co_filename.startswith(path_prefix):
-        frame = frame.f_back
+    try:
+        frame = inspect.currentframe().f_back
```
jni: @mkitti thanks for this!

One thing I would say is that generally you want to wrap the smallest amount of code possible in a try/except, so you are not accidentally catching things you didn't mean to catch.

In this case, the first `currentframe().f_back` is always guaranteed to be within napari and therefore exist (?) (my lack of knowledge of the Julia/Python interface is probably showing :sweat_smile:). In that case, you could wrap just one line in try/except: `frame = frame.f_back`.

Even if not, I'd suggest wrapping both `f_back` lines in try/except and leaving the other lines untouched.

How does this sound?

Also, can you think of a simple way that this can be tested? It's fine if not, just thought it would be worth thinking about.

mkitti: Let's itemize where the potential issues are:
1. `inspect.currentframe()` could be `None` if the underlying Python implementation lacks stack frame support causing an `AttributeError` on master ( 5e8be09 ) Line 64.
https://docs.python.org/3/library/inspect.html#inspect.currentframe
2. `frame = currentframe().f_back` could be `None` depending on the caller within Python on Line 64 causing an error on Line 65 when trying to obtain `frame.f_code`
3. `frame.f_back` could be `None` on Line 72 depending on the caller within Python resulting in `AttributeError` on line Line 73.

Additonally, it might make sense to keep current structure and add a `finally` to delete `frame` explicitly to avoid a stack frame leak:
https://docs.python.org/3/library/inspect.html#the-interpreter-stack
https://bugs.python.org/issue543148

In terms of control flow, the current situation makes sense to me in a structured programming sense. If an exception occurs anywhere in the try block, we should stop executing the rest of the try block. While returning from the `except` would have the same effect, I find it a bit harder to follow.

In light of this, I have a more explicit refactoring in mind if we want to check each line.

mkitti:
I have modified the code to use `inspect.isframe` and `inspect.iscode` before retrieving properties which should resolve #1630 by screening out `None`. Is this heading in the right direction?

mkitti:
I am unclear what is causing the test failures on windows and macos. I believe they are unrelated the changes here.

Now that we have a test and what I believe to be a robust solution, this PR is ready. I have requested re-review by @jni .

napari/utils/naming.py
```diff
@@ -61,16 +61,31 @@ def magic_name(value, *, path_prefix):
...
+            for name in names:
+                if (
+                    name.isidentifier()
+                    and name in varmap
+                    and varmap[name] is value
+                ):
+                    return name
+    finally:
+        # Prevent stack frame leak
+        del frame
```
Czaki, line 104: Why this is need? Why GC is not enough? After go out from function reference counter will go down. 
I suggest rather `frame=None` to not break debugger.

mkitti: See here:

https://docs.python.org/3/library/inspect.html#the-interpreter-stack
https://bugs.python.org/issue543148

Czaki: If I good understand You should move bellow `return None` in try clause. Otherwise you could meet some problem with execution code when frame was deleted. 

I also suggest use `frame.clear()` protected by `None` check. Reason is mentioned in documentation:

> If you want to keep the frame around (for example to print a traceback later), you can also break reference cycles by using the frame.clear() method.

As you see issue which you point is for Python in version 2.1. Later one gc clear this, but with some delay.

mkitti: I moved the `return None` into the try block. `frame.clear()` causes a RuntimeError though.

Czaki: I see. It looks like there should be created bug report to python documentation.

mkitti: The problem is that frames being referenced are still executing, so we cannot clear them.

Czaki: I run test with simple code 
```python 
import inspect
import os

print(os.getpid())

def leak():
    frame = inspect.currentframe()

while 1:
    leak()
```
and record memory usage using `procpath`  on python 3.8. 

It looks like there is no memory leak: 
Zrzut ekranu z 2020-10-08 12-25-10

Its also look like `del` only remove frame reference. I could access to variable values with debugger even when frame was deleted in `finally` step. 

Fro my test it looks like `del frame` is obsolete but do not have negative impact.

mkitti: Maybe try a deeper call stack? Are there any downsides to having `try` / `finally` / `del frame` ?

Czaki: Only which I see is code complexity. I will try later

mkitti: I see that you're using RSS to check, but I think we also might want to use something that can probe the Python interpreter and garbage collection more directly.

I'm not sure if Napari will run on anything other than CPython. However, the presence of a cycle detector is an interpreter specific matter.

Given that this pattern is currently recommended in the docs for inspect, I would lean towards keeping it. It would be worthwhile to inquire upstream whether that note still applies though.

sofroniewn: @jni I'm curious what your thoughts are here?

sofroniewn:
@mkitti now that i've merge #1716 too can you update from master one more time.

@jni can you give this a final review, I'm curious your take on the `try` statement as discussed above.

Also @Czaki curious if you have any follow up thoughts on that discussion too?

I'd like to get this merge in for `0.4.0`. Thanks all!!

napari/utils/naming.py
jni: I'm afraid I'm out of my stack depth. 😂 It would be good at least to add comments above the `del frame` line, linking to this discussion as well as the references therein so that future code archeologists can understand the current design. Based on the CPython docs I agree with @mkitti that the cycle detector is an implementation detail. We probably depend on CPython vs other interpreters like, a lot 😂 but we should try to avoid hardcoding that dependency as much as possible.

I'll note one difference between the current code and the code in the linked CPython docs: `frame = inspect.currentframe()` happens outside the `try:` in the docs. Should we mimic that?

jni (approved):
I've suggested adding some comments and links above the `del frame` line, as well as moving `frame = inspect.currentframe()` out of the try: block, but other than that I'm happy for this to go in. Thank you @mkitti for all your work here! (And thank you @Czaki for your review, it's been immensely valuable for my understanding, but ultimately I agree with @mkitti that we should follow the pattern in the CPython docs, as long as the code is working for everyone here.)

napari/utils/naming.py
mkitti: I have added comments with references and moved the initial line out of try block. Thanks for taking a look at this again, @jni.

mkitti:
Thanks @jni ,@czaki, and @sofroniewn . I've made the requested changes and merged master again. I'm glad I was able to able to help here and with the other issue. I think the function should be pretty robust going forward. Let me know if there is anything else to do here.

jni:
Awesome, thanks so much for this work @mkitti!

@Czaki if this breaks some of your debugging workflows, please raise an issue with a reproducible example and we'll see how we can modify!

PR #1638, opened 2020-09-21
Build docs action

kne42:
add an action that automatically places docs into different folders depending on their version, e.g. `dev`, `0.3.4`, etc.

I also have a bad-looking index page to add to the docs repo once this gets merged. I might look into this more and see if I can do it with Github Pages instead so it looks okay.

Screen Shot 2020-09-18 at 12 59 24 PM


sofroniewn:
Thanks @kne42! I see this is still marked as draft, did you want us to review/ merge now or did you want us to wait until you've looked into Github Pages more?

kne42:
I want to look into GH Pages more so the index page doesn't look so terrible

sofroniewn:
> I want to look into GH Pages more so the index page doesn't look so terrible

Sounds good! Let us know when this is ready for review :-)

sofroniewn (approved):
This looks good to me @kne42, thanks so much for adding this. As discussed in meeting there are a number of further things we might want to do like manually add some older releases, but it's great to get this going like this. I will merge now.

sofroniewn:
Looks like there has been an error in the build because of a missing token
```
github-actions
/ Build & Publish Docs

.github#L1
Input required and not supplied: token
```
see here https://github.com/napari/napari/actions/runs/283517324

Any ideas for a fix @kne42?

Issue #1621, opened 2020-09-06
New point color off


sofroniewn:
## 🐛 Bug

Changing color before adding the first point to a layer doesn't cause new point to have new color 

## To Reproduce

1. Run `python examples/add_image.py`
2. Click the new points layer
3. Change the face color from white to red - see color icon change
4. Click in Add mode and add point. Point is white and color icon is now white too

## Expected behavior

Point would be red. Note that changing color to red and then adding more points give red points, this seems to affect only the first point added.

## Environment

napari 0.3.7

sofroniewn:
@kevinyamauchi @ziyangczi @jni just flagging, this one needs to be reproduced but looks probably pretty simiple. Would be great to get it in for the 0.3.8 release which I think we'd like to do soon (will be a small bug fix release and likely our last python 3.6 release)

kevinyamauchi:
I can take a look tomorrow!

sofroniewn:
> I can take a look tomorrow!

Thanks @kevinyamauchi!! I think @ziyangczi might be looking into this afternoon, so look out if there is a PR first, but that would be great. PR or tips in the right direction would be appreciated. Also @ziyangczi if you run into questions looking into today it you could ask @kevinyamauchi

liu-ziyang:
I am planning to dig into this tomorrow as well, but happy to see more people able to reproduce it!

jni:
To clarify the symptoms, when you start adding points, they *alternate* between white and red, and the little face_color box alternates as well. It's really quite fun and hilarious! 😂 

<img width="1200" alt="Screen-Shot-2020-09-22-at-12 08 19-pm" src="https://user-images.githubusercontent.com/492549/93838625-ae3dc380-fccd-11ea-8f34-6650b8bd5323.png">

#1623 does **not** fix the issue, in case anyone's wondering.

kevinyamauchi:
> Thanks @kevinyamauchi!! I think @ziyangczi might be looking into this afternoon, so look out if there is a PR first, but that would be great. PR or tips in the right direction would be appreciated. Also @ziyangczi if you run into questions looking into today it you could ask @kevinyamauchi

> I am planning to dig into this tomorrow as well, but happy to see more people able to reproduce it!

Okay, sounds good! Thanks, @ziyangczi !

sofroniewn:
Closed by #1643

PR #1643, opened 2020-09-22
Fix new color point off

liu-ziyang:
# Description
Fix new color point off issue https://github.com/napari/napari/issues/1621 

## Type of change

# References
https://github.com/napari/napari/issues/1621

# How has this been tested?
- Manually verified reported case is behaving as expected
- example: all tests pass with my change

## Final checklist:


napari/layers/points/points.py
```diff
@@ -390,6 +390,8 @@ def __init__(
         else:
             self._current_edge_color = self.edge_color[-1]
             self._current_face_color = self.face_color[-1]
+            self._face_color = []
+            self._edge_color = []
```
kevinyamauchi: Even though it's empty, I think it might be better to initialize as a numpy since that is the assumed type for the colors.

```suggestion
            self._face_color = np.empty((0, 4))
            self._edge_color = np.empty((0, 4))
```

kevinyamauchi (approved):
Thanks, @ziyangczi ! I was able to reproduce the issue on master and confirmed this fixes it.

In addition to the minor comment below, I think it would be nice to add a test that covers this bug. Other than that, I think it's good to go!

sofroniewn (approved):
LGTM! Thanks for the fix and test @ziyangczi and the review @kevinyamauchi!!

PR #1652, opened 2020-09-25
Drop Python 3.6

sofroniewn:
# Description
This PR drops Python 3.6 by removing it from all our CI frameworks. It will pave the way for merging #1475.

I imagine @jni that we might be able to change some imports / remove some dependencies that were providing back-ported features, is that right? If so can you take over this PR and make those changes in them too, thanks!!

## Type of change

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -28,9 +28,6 @@ linux_task:
   
   matrix:
     env:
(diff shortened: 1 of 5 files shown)


jni (approved):
🚀

PR #1654, opened 2020-09-25
Fix pan/zoom jump

### LINKED ISSUE #1579: pan/zoom broken for multiscale images with scale= parameter
## 🐛 Bug

When viewing a multiscale image, setting the scale on the viewed dimensions to anything other than 1, 1 causes the pan/zoom to break.

```python
import napari
import dask.array as da

with napari.gui_qt():
    v = napari.view_image(
        [da.random.random((10000, 10000)), da.random.random((5000, 5000))],
        scale=[10, 10],
    )
```

See the screengrab video. The jumps are just me scrolling.

https://imgur.com/a/UpzyOC6

When you click and drag, the coordinates displayed in the status bar change while dragging. 🤷

liu-ziyang:
# Description
Address pan/zoom jump

## Type of change

# References
closes #1579 

# How has this been tested?

- manually verified that the behavior is fixed

## Final checklist:


liu-ziyang:
@sofroniewn Ok, this is crazy, I definitely need you to take another look on this code change, I dig a few days, very confused about how this jump is coming from, after I dig the transformation inside out, and ruling out other possibilities, I found this nonsense code change actually would fix the problem 🤯 . We would have to update the unit tests etc to reflect this later if we think this is right. This seems to indicate that vispy scaling and translation doesn't seem to be composite the way we think?

Note that unit test: test_transform_chain_simplified screams very hard against this change

liu-ziyang:
I was originally thinking there was a missing clips somewhere which didn't constrain the coordinate when we pan/zoom out of bound, but after the digging, I am more convinced that the behavior is with transformation than clipping

sofroniewn:
Yikes! Great job chasing that down, but I don't think that will be the final solution we want as I'm pretty sure about the mathematics of that composition rule. I just did some more looking around and found this line which looks off https://github.com/napari/napari/blob/master/napari/layers/image/image.py#L559. I think it should be 

```python
                self._transforms['tile2data'].translate = (
                    corner_pixels[0]
                    * self._transforms['tile2data'].scale
                )
```

and we havn't run into this problem before because `self._transforms['data2world'].scale` has been `1.0`. Removing that extra multiplication should have the same effect as removing the multiplication in the composition that you did, and I suspect that we used to need it, but when we added the transforms it now gets applied twice!!

Let me know if this works too, and if so then I think this will be the fix we want

liu-ziyang:
> Yikes! Great job chasing that down, but I don't think that will be the final solution we want as I'm pretty sure about the mathematics of that composition rule. I just did some more looking around and found this line which looks off https://github.com/napari/napari/blob/master/napari/layers/image/image.py#L559. I think it should be
> 
> ```python
>                 self._transforms['tile2data'].translate = (
>                     corner_pixels[0]
>                     * self._transforms['tile2data'].scale
>                 )
> ```
> 
> and we havn't run into this problem before because `self._transforms['data2world'].scale` has been `1.0`. Removing that extra multiplication should have the same effect as removing the multiplication in the composition that you did, and I suspect that we used to need it, but when we added the transforms it now gets applied twice!!
> 
> Let me know if this works too, and if so then I think this will be the fix we want

makes more sense, let me give it a try, I am also pretty confident that the transform is correct, so I was very confused where else the bug come from

PR #1655, opened 2020-09-27
magic_name: Add test for an empty path_prefix that matches all paths

mkitti:
This also tests when magic_name reaches the end of the stackframes

# Description
This is a noop PR to see if this test fails in CI on the master branch

## Type of change
- [x] Adds a test for `magic_name`

# References
This tests for the issue in #1630 . The test should succeed when merged with #1635

# How has this been tested?
Running pytest locally for Python 3.8 indicates that the new fails on master

## Final checklist:

### DIFF
--- a/napari/utils/_tests/test_naming.py
+++ b/napari/utils/_tests/test_naming.py
@@ -116,3 +116,25 @@ def test_path_prefix():
 
     if walrus:
         assert eval_with_filename('foo(i:=33)', 'rye.py') == 'i'
+
+
+def test_empty_path_prefix():
+    """Test an empty path prefix that matches the entire stack"""
+    # Repeat tests with an empty path_prefix
+    mname = functools.partial(magic_name, path_prefix="")
+
+    def foo(x):
+        def bar(y):
+            return mname(y)
+
+        return bar(x)
+
+    # Test are all None because the path_prefix matches everything
+    # magic_name reads through until the end of the stack
+    assert eval_with_filename('foo(42)', 'hi.py') is None
+
+    r = 8  # noqa
+    assert eval_with_filename('foo(r)', 'bye.py') is None
+
+    if walrus:
+        assert eval_with_filename('foo(i:=33)', 'rye.py') is None


mkitti:
As expected, [`test_empty_path_prefix`](https://github.com/napari/napari/pull/1655/files#diff-fbf726ba410bb74e66add830aff03710R121-R140) causes an `AttributeError` similar to the one seen in #1630 

```
____________________________ test_empty_path_prefix ____________________________
    def test_empty_path_prefix():
        """Test an empty path prefix that matches the entire stack"""
        # Repeat tests with an empty path_prefix
        mname = functools.partial(magic_name, path_prefix="")
        def foo(x):
            def bar(y):
                return mname(y)
            return bar(x)
        # Test are all None because the path_prefix matches everything
        # magic_name reads through until the end of the stack
>       assert eval_with_filename('foo(42)', 'hi.py') is None
napari/utils/_tests/test_naming.py:134: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
napari/utils/_tests/test_naming.py:51: in eval_with_filename
    return eval(code, frame.f_globals, frame.f_locals)
hi.py:1: in <module>
    ???
napari/utils/_tests/test_naming.py:130: in foo
    return bar(x)
napari/utils/_tests/test_naming.py:128: in bar
    return mname(y)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
value = 42
    def magic_name(value, *, path_prefix):
        """Fetch the name of the variable with the given value passed to the calling function.
        Parameters
        ----------
        value : any
            The value of the desired variable.
        path_prefix : absolute path-like, kwonly
            The path prefixes to ignore.
        Returns
        -------
        name : str or None
            Name of the variable, if found.
        """
        frame = inspect.currentframe().f_back
        code = frame.f_code
        while code.co_filename.startswith(path_prefix):
            frame = frame.f_back
>           code = frame.f_code
E           AttributeError: 'NoneType' object has no attribute 'f_code'
napari/utils/naming.py:69: AttributeError
=============================== warnings summary ===============================
napari/layers/surface/_tests/test_surface.py::test_random_3D_timeseries_surface
  /private/var/folders/3y/l0z1x3693dl_8n0qybp4dqwh0000gn/T/cirrus-ci-build/napari/layers/surface/surface.py:260: UserWarning: Assigning multiple values per vertex after slicing is
                      not allowed. All dimensions corresponding to vertex_values
                      must be non-displayed dimensions. Data will not be
                      visible.
    warnings.warn(
-- Docs: https://docs.pytest.org/en/stable/warnings.html
=========================== short test summary info ============================
FAILED napari/layers/image/_tests/test_big_image_timing.py::test_timing_fast_big_dask_nothing_specified[data0]
FAILED napari/layers/utils/_tests/test_layer_utils.py::test_calc_data_range_fast_big
FAILED napari/utils/_tests/test_naming.py::test_empty_path_prefix - Attribute...
= 3 failed, 1279 passed, 3 skipped, 1 xfailed, 1 warning in 159.15s (0:02:39) ==
```

sofroniewn:
hi @mkitti thanks for writing a test! Am I correct that you've added in to #1635 too, so that after we merge that PR we won't have to actually merge this one as the code will already be in?

mkitti:
> Am I correct that you've added in to #1635 too, so that after we merge that PR we won't have to actually merge this one as the code will already be in?

That is correct. I cherry picked off the commit adding the new test and added it to #1635. I think we can close this PR immediately after merging #1635 if that is what you intend to do. Otherwise, an alternative fix than the one I proposed in #1635 should at least resolve the issue identified in the test here.

sofroniewn:
I'm hopeful that we're getting close to merging #1635 which contains a copy of this test and so I am going to close this PR as no longer needed and so as not to confuse people looking for which one to review etc. If we end up deciding to do something different we can always reopen this PR or copy this test to a new PR. Thanks!!

Issue #1656, opened 2020-09-28
failing dask test


sofroniewn:
## 🐛 Bug

Looks like some of our tests are now failing on master possibly due to a new dask release. I will look into and post more detail here when available.

sofroniewn:
See the CI report here, looks like it is only python3.8 macos https://cirrus-ci.com/task/5702782638358528 right now ...

sofroniewn:
Looks like the failings are timeout related. I notice that `pytest napari/layers/image/_tests/test_big_image_timing.py` is much slower with dask `2.28.0` compared to `2.27.0` and I think we're now passing our 2sec timeouts.

This could relate to https://github.com/dask/dask/pull/6665 and discussions around [efficiency](https://docs.dask.org/en/latest/array-slicing.html#array-slicing-efficiency) here. I think we might now be hitting some of the "interactive overhead of a second or two".

It could also relate to our dask config which might now need updating see, https://github.com/napari/napari/blob/master/napari/utils/dask_utils.py#L121.

We might want to now adjust `dask.config.set({"array.slicing.split_large_chunks": True})`.

Thanks @jni for some tips, I'll play around some more.

sofroniewn:
I've tried
```python
        def dask_optimized_slicing():
            with dask.config.set{"optimization.fuse.active": False, "array.slicing.split-large-chunks": False}) as cfg:
                yield cfg
```
which I think should have reverted to the behavior before `2.28.0` is released, but it doesn't seem to have. The config might not be getting set as i expect though ...

sofroniewn:
It seems like multiscale performance with dask is actually much worse on `2.28.0` too, so this isn't just making our tests worse, it's making the actual performance worse too (panning and zooming with one of my pathology slide examples is slower). Maybe one of our dask experts @jakirkham @GenevieveBuckley has ideas on this one? We might want to pin `<2.28.0` until we figure it out. Thoughts @jni?

GenevieveBuckley:
I support pinning until we work it out.

I don't have an idea about what's happened, though. It might be worth mentioning that in a dask issue, or maybe over on the dask gitter. Other people might have a better idea of what the most recent changes were.

[...]

PR #1657, opened 2020-09-28
Fix pan zoom scaling

### LINKED ISSUE #1579: pan/zoom broken for multiscale images with scale= parameter

liu-ziyang:
# Description
Address pan/zoom jump

## Type of change

# References
closes #1579 
replace #1654

# How has this been tested?

- manually verified that the behavior is fixed

## Final checklist:


sofroniewn:
Amazing @ziyangczi That's great. Adding a test maybe tricky. You could try one based on screenshots like are in here which take screenshots and do zooms https://github.com/napari/napari/blob/master/napari/_vispy/_tests/test_vispy_multiscale.py

I'll give this a test later today. Maybe @jni can test too as he originally raised #1579

Also looks like the test fails are dask related which I've made an issue for in #1656, if you had any ideas around that it would they would be much appreciated too!!

liu-ziyang:
> Amazing @ziyangczi That's great. Adding a test maybe tricky. You could try one based on screenshots like are in here which take screenshots and do zooms https://github.com/napari/napari/blob/master/napari/_vispy/_tests/test_vispy_multiscale.py
> 
> I'll give this a test later today. Maybe @jni can test too as he originally raised #1579
> 
> Also looks like the test fails are dask related which I've made an issue for in #1656, if you had any ideas around that it would they would be much appreciated too!!

I may need some help understanding the rationale behind the coordinate transform, as well as possible impact on other pieces of the code to justify this code change, I still don't understand exactly what this does, other than it fixed the reported issue 😃

DragaDoncila:
Just tested this locally and it works great! Thanks @ziyangczi 🙌

liu-ziyang:
@sofroniewn yeah adding test for this one is tricky. maybe we can skip for this one liner.

But I do have a question though, if we commit this code change, how is the data translated to the world coordinate without the rescaling? do we need to add the scaling somewhere else to avoid another bug?

sofroniewn:
> But I do have a question though, if we commit this code change, how is the data translated to the world coordinate without the rescaling? do we need to add the scaling somewhere else to avoid another bug?

The `tile2data` and `data2world` transforms now get combined here https://github.com/napari/napari/blob/master/napari/_vispy/vispy_base_layer.py#L162-L164, and I think what's happening is that scaling is getting applied twice, whereas before we needed to apply it in this part of the code.

This code only gets used in the multiscale case, so I think it is fine to merge if it has been checked there.

I did look and it seems like using translate parameter also causes a bug that is neither fixed nor made worse by this PR, see

```python
with napari.gui_qt():
    v = napari.view_image(
        [da.random.random((10000, 10000)), da.random.random((5000, 5000))],
        scale=[10, 10], translate=[1000, 100], 
    )
```

maybe @ziyangczi we can look into trying to fix that too before merging this one? i'll have a look in the code now and see if anything looks off. note in this case black starts appearing at the top of the screen as you zoom in

sofroniewn:
Looking at it more I'm now worried that with scale/ translate parameters present the `corner_pixels` calculation happening here might not be accurate https://github.com/napari/napari/blob/master/napari/_vispy/vispy_base_layer.py#L202 (i.e. it assumed no scale/ translates). If that is wrong then these other calculations will be off too.

Ultimately all this code will go away with the finishing touches on our world coordinates and the octree work, but still would be good to fix now if we can!!

liu-ziyang:
> > But I do have a question though, if we commit this code change, how is the data translated to the world coordinate without the rescaling? do we need to add the scaling somewhere else to avoid another bug?
> 
> The `tile2data` and `data2world` transforms now get combined here https://github.com/napari/napari/blob/master/napari/_vispy/vispy_base_layer.py#L162-L164, and I think what's happening is that scaling is getting applied twice, whereas before we needed to apply it in this part of the code.
> 
> This code only gets used in the multiscale case, so I think it is fine to merge if it has been checked there.
> 
> I did look and it seems like using translate parameter also causes a bug that is neither fixed nor made worse by this PR, see
> 
> ```python
> with napari.gui_qt():
>     v = napari.view_image(
>         [da.random.random((10000, 10000)), da.random.random((5000, 5000))],
>         scale=[10, 10], translate=[1000, 100], 
>     )
> ```
> 
> maybe @ziyangczi we can look into trying to fix that too before merging this one? i'll have a look in the code now and see if anything looks off. note in this case black starts appearing at the top of the screen as you zoom in

I think I identified the translate issue, it is a different bug as you found on the vispy base layer where it shouldn't include the offset in transformation. However there is a phantom update on the transform node value and caused the scaling to jump on micro scale, I am chasing it down now

liu-ziyang:
I think I figured the phantom update out, it is actually related to async loading and how we compute translation:
we use the transformation to figure out our canvas corner: 
https://github.com/napari/napari/blob/master/napari/_vispy/vispy_base_layer.py#L162-L164, and we use the canvas corner to figure out the translation: https://github.com/napari/napari/blob/master/napari/layers/image/image.py#L557-L561.
However because there is a lag between when events are fired off and processed, you will see a phantom update of the coordinates because one part will be behind while visual is updating...
Given our current setup I don't think we can easily fix this, one workaround that I can think of is to always loads slightly more data in the slice than we have to, this way we avoid having a visual flash in users' eyes even though the corner coords and slice would still bounce around..

sofroniewn:
> I think I figured the phantom update out, it is actually related to async loading and how we compute translation:
we use the transformation to figure out our canvas corner:

Hmm, ok interesting. Can you role back to an old commit before any of the async stuff merged, maybe the 0.3.6 release tag and check if the problem is the same then. I wonder if the problem is strictly with the "async" PR code vs just how/ when we are doing that update, which maybe asynchronous in a slightly looser sense

sofroniewn:
I wonder if the `offset` was needed to account for the corner coords bouncing around.

I think that we might want to merge with the offset left in and the `scale` fixed, that is at least real progress and then after #1660 merges we can redo the corner calculation to avoid this code path entirely. This will be much easier as we'll finally have a real world coordinate system that is the same for every layer

liu-ziyang:
> I wonder if the `offset` was needed to account for the corner coords bouncing around.
> 
> I think that we might want to merge with the offset left in and the `scale` fixed, that is at least real progress and then after #1660 merges we can redo the corner calculation to avoid this code path entirely. This will be much easier as we'll finally have a real world coordinate system that is the same for every layer

that sounds fine to me, I am pretty sure we need to remove the offset to fix the translate bug, but it technically is a different bug. so we can file and address later

sofroniewn (approved):
Ok, this looks good now - at least we have fixed the scale issue. We'll address the translate issue in followup PRs after #1660 has merged cc @jni

PR #1659, opened 2020-09-29
async-2.6: Move async into experimental 

pwinston:
# Description

This is a **wedge PR** for async rendering. It wedges in the **async rendering** functionality by making a few strategic changes mostly related to the `Image` class, while moving most async related functionality into a new `experimental` namespace. The `experimental` code is not even imported unless `NAPARI_ASYNC` is defined.

The intent is to protect regular users from any problems with the async code. Unless you opt-in to async you are not running that code at all. After this wedge PR, the hope is the PR's `async-3` through `async-12` are easier to review since it will be clear which changes apply only to experimental code.

## Changes:

1. Move `chunk` module into `napari.components.experimental`.
2. Move `QtChunkReceiver` class into `napari._qt.experimental`.
3. New async `ChunkedImageLoader` and `ChunkedSliceData` classes in `napari.layers.image.experimental`.
4. New legacy `ImageLoader` and `ImageSliceData` classes in `napari.layers.image` (for non-async).

## When NAPARI_ASYNC is defined

1. `viewer.Viewer.close()` calls into `chunk_loader` to cleanup.
2. `QtLayerList.__init__()` creates a `QtChunkReceiver`.
3. `Image` gets one extra method: `Image.on_chunk_loaded()`.
4. `Image` uses `ChunkedSliceData` instead of `ImageSliceData`.
5. `ImageSlice` uses `ChunkedImageLoader` instead of `ImageLoader`.

**Notes:**
1. Only `QtChunkReceiver` calls into the new `Image.on_chunk_loaded()` method, the only purpose of `on_chunk_loaded()` is to receive chunks from the `ChunkLoader` via the `QtChunkReceiver`.
2. `ChunkedSliceData` derives from `ImageSliceData` to add two async related methods: `load_chunks` and `from_request`. Both are only used when a `ChunkRequest` was asynchronously loaded.
3. `ChunkedImageLoader` derives from `Imageloader` and overrides the `load()` and `match()` methods for async.

We use `Chunked` prefix here instead of `Async` because `NAPARI_ASYNC` really means "load with the `ChunkLoader`" but those loads might be synchronous or asynchronous. We do synchronous loads for `ndarray` and for fast loading arrays of any type. So `ChunkedImageLoader` is a better than name `AsyncImageLoader` since the loads will not always be async.

## Type of change
- [x] Refactoring, moving code into experimental, no user visible changes.


pwinston:
Marking ready for review. I've looked it over a bunch of times.
I want to fix one bug with multi-scale (async only bug).
See top description for details.

napari/layers/image/image.py
```diff
@@ -645,18 +635,19 @@ def on_chunk_loaded(
...
-        # Chunks get transposed after load.
-        request.transpose_chunks(self._get_order())
+        # Transpose after the load.
+        data.transpose(self._get_order())
 
         # The shape of the array after we loaded it was "wrong" in that it
         # didn't match the rgb flag that the user specified (or we inferred?).
-        if self.rgb != guess_rgb(request.image.shape):
+        # TODO_ASYNC: do something sooner so this check is not required?
+        if self.rgb != data.rgb:
```
sofroniewn: I know this isn't a change in this PR but, it was already added in an earlier async PR but I want to flag that this still causes a bug in the synchronous code that didn't exist before the async work started (i.e. roll back to 0.3.6 and the following works fine). To reproduce the bug just run

```python
    viewer = napari.view_image(data.astronaut(), rgb=False, ndisplay=3)
```

When it was added to async I think the justification was something like it may have prevented an error, but given that this introduces a bug where there wasn't one before I think we should probably just drop it.

napari/layers/image/_image_slice_data.py
```diff
@@ -0,0 +1,64 @@
...
+            Transpose the image into this order.
+        """
+        self.image = self.image.transpose(order)
+
+        if self.thumbnail_source is not None:
+            self.thumbnail_source = self.thumbnail_source.transpose(order)
+
+    @property
+    def rgb(self) -> bool:
+        return guess_rgb(self.image.shape)
```
sofroniewn: See comment below about dropping

sofroniewn (approved):
Ok, this looks good to me and I'm approving. I've flagged one more bug around the `rgb` guessing, that I think we should fix before merge, but that is just dropping a line of code. I've run this PR through my battery of interactive examples, including multiscale and painting and not found any other problems.

Overall I also like the better separation between experimental and non-experimental code, thanks for that.

I will flag though the addition of two objects `ImageSliceData` and `ImageLoader` which now adds some indirection to the sync code to make the parallelism between async and sync easier. I think this is ok as the objects are very small and easy to understand, and the data flow is clear to me, but at some point when we have everything working we might want to look into reducing the number of objects if it makes things simpler again.

Along that vein I will also flag that I think a lot of the additional complexity comes from the "thumbnail" and the need to carry around the additional "thumbnail source" which is only relevant for multiscale data (where we use one of the low resolution levels for the thumbnail). Maybe when we get to the octree we can think if there are ways to reduce some of the complexity. In general we probably want to revisit thumbnail generation for all the layers too.

I see @pwinston you mentioned you might want to look into one or two other things tomorrow before I merge, so I'll wait for you to give the green light, and then it will be good to go!

pwinston:
> I will flag though the addition of two objects `ImageSliceData` and `ImageLoader` 

Yes `ImageSliceData` and `ImageLoader` were only added to meet the "push as much async code into experimental" goal. The way it worked in #1510 is that everything went through `ChunkLoader`, but `ChunkLoader` could be set to do synchronous loading 100% of the time if not using `NAPARI_ASYNC`. But all loads went through the `ChunkLoader` regardless.

In this PR since `ChunkLoader` goes into experimental, we don't want anything using `ChunkLoader` if we are not using `NAPARI_ASYNC`. And we wanted to get the Chunk-related code out of `Image`, since that too needs to be only in experimental. When `ChunkLoader` is no longer experimental it can go back to how #1510 worked. The extra classes then can optionally be removed, since there will no longer be two flavors of them.

I think this is typical for a complicated feature-toggle, that you write some scaffolding code that you hope to eventually remove. That code is basically the price for having a run-time toggle instead of a git branch. And that might be a reason why sometimes a git branch is better or easier, if you really don't need to toggle it at run time.

napari/layers/image/image.py
pwinston: Okay I removed the rgb test entirely. When that error comes up again we can try to find a better way to flag it and help the user figure out what went wrong.

napari/layers/image/_image_slice_data.py
pwinston: Removed

pwinston:
>  you mentioned you might want to look into one or two other things

Yes I want to look into the async multi-scale issue on load, not getting the view bounds correct until the 3rd try.

I'll reply here if I can resolve it or if I punt for now, since it's an async-only issue.

napari/layers/image/image.py
sofroniewn: sounds good, thanks!

sofroniewn:
Ok, I have tested this again, it looks good to me! As noted above we've still got an issue with async / multiscale on load, which resolves after panning, due to the `corner_pixels` update but let's get this in now and I can look into that `corner_pixels` update over the weekend. The issue is not present for sync so it's ok to merge as is.

PR #1660, opened 2020-09-30
fix world origin

sofroniewn:
# Description
This PR now adds a correct world based coordinate system that follows image-like axes conventions. The center of the [0, 0] pixel is not truely at [0, 0]. I think this is the way that @jni has been asking for a long time.

You can see this when looking at layers of different scales, for example

```python
with napari.gui_qt():
    viewer = napari.Viewer()
    viewer.add_image(np.random.random((11, 11)), colormap='red', opacity=.5, scale=(2, 2), name='half')
    viewer.add_image(np.random.random((21, 21)), colormap='blue', opacity=.5, name='full')
    viewer.add_points([[0, 0], [10, 10]], size=0.5, opacity=.5)
```

Currently on master yields:

Screen Shot 2020-09-29 at 4 59 23 PM

where you can see the top left corners of the arrays are at the same location, but the points have been offset 0.5

Now on this PR:

Screen Shot 2020-09-29 at 4 59 46 PM

Where now the center of the pixels of both arrays are overlapping and the point is in the correct place, which is what we want.

This also drops some of the strange `0.5` pixel offsets we needed before, instead introducing the concept of `_array_like` for a vispy visual. Depending on what merges first we'll want to update #1361 to reflect this too.

I have checked a bunch of the interactivity, cursor selection, painting etc. I want to add a cursor model in a PR after this and simplify some stuff, but that won't actually change the functionality, just make things simpler.

I can also add some tests for this before merge, but maybe want to hand it over to @jni @DragaDoncila @GenevieveBuckley @kevinyamauchi @d-v-b for playing with. Hopefully people like the new behavior!

Also note that this can merge before #1616 and then I can update #1616 so that it preserves this new world coordinate system.

## Type of change

### DIFF
--- a/napari/_tests/test_viewer.py
+++ b/napari/_tests/test_viewer.py
@@ -4,7 +4,7 @@ import time
 import numpy as np
 import pytest
 
-from napari import Viewer
+from napari import Viewer, layers
 from napari._tests.utils import (
     add_layer_by_type,
     check_view_transform_consistency,
@@ -180,6 +180,9 @@ def test_roll_traspose_update(make_test_viewer, layer_class, data, ndim):
     }
     for k, val in transf_dict.items():
         setattr(layer, k, val)
+
+    if layer_class in [layers.Image, layers.Labels]:
+        transf_dict['translate'] -= transf_dict['scale'] / 2
 
     # Check consistency:
     check_view_transform_consistency(layer, viewer, transf_dict)
--- a/napari/_vispy/vispy_base_layer.py
+++ b/napari/_vispy/vispy_base_layer.py
@@ -47,6 +47,7 @@ class VispyBaseLayer(ABC):
         super().__init__()
 
         self.layer = layer
+        self._array_like = False
         self.node = node
 
         MAX_TEXTURE_SIZE_2D, MAX_TEXTURE_SIZE_3D = get_max_texture_sizes()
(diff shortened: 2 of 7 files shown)


jni:
My mind is 🤯!

This is awesome @sofroniewn! One thing I would say is in the docs/examples, we should document that users can set `translate=0.5 * scale` to move to corner-centric coordinates, which some users want. (Specifically, our satellite data is corner-aligned.)

sofroniewn:
> This is awesome @sofroniewn! One thing I would say is in the docs/examples, we should document that users can set translate=0.5 * scale to move to corner-centric coordinates, which some users want. (Specifically, our satellite data is corner-aligned.)

Thanks! Yes, great idea, we should do that. Probably will be worth have a whole docs page on the coordinate system.

I've fixed the tests so they all pass locally. This is now ready for review!!

After this goes in we can make some refactor changes to clean up some of our transforms, prepare for the new cursor handling which will be much nicer. It should help chase down some of the nasty bugs like are popping up in #1657 as all the transform stuff will now be unified and simpler!

sofroniewn:
Note that the two CI fails are unrelated to this PR. The windows isn't installing, and the mac is the dask issue in #1656

jni (approved):
Wow, so simple! I haven't tested locally but I'm very happy with the code changes. Happy to test more later in the week but feel free to push forward if you're happy with your local tests.

Issue #1664, opened 2020-09-30
Skip roll over 1 size axis 


Czaki:
## 🚀 Feature

Ignore empty dimension on "Roll Dimension"
## Motivation

To have consistency of coordinates between different data I normalize axis order and input single dimension axes.
napari does not show slider for such dimension, but when using "Roll dimensions order" this axes could be set as visible, which produce black screen. 

## Pitch

When do roll over dimension skip this combination which contains axes which size is 1. 

## Alternatives
Add separate button/combo box for side view (`zx` and `zy`)
or  even drag'n' drop mechanism for manually modify order of axes (with mnemonic names)

sofroniewn:
Nice catch, fixing this makes sense. It should be a relatively simple change to the logic here https://github.com/napari/napari/blob/master/napari/components/dims.py#L399-L401

If you'd like to make a PR @Czaki I'd happily review

Czaki:
Created, but I stil think that there should be separated option for side view (this will not provide side view for 3d time data for example). Maybe using affinity from #1616,

sofroniewn:
> Created, but I stil think that there should be separated option for side view (this will not provide side view for 3d time data for example). Maybe using affinity from #1616,

Can you explain this a bit more? I'm not sure I understand

Czaki:
> Can you explain this a bit more? I'm not sure I understand

It is common to to view data not only in xy but also in zy and zx. 
But if there is and image with `tzyx` axis order (all axis with more than 1 step) then using roll operation user could see: `yx`, `zy`, `tz`, `xt` planes. So no view on `zx` (and for 3d data this operation will produce view in `xz` not `zx`)

It may be done with changing order, or in future using afinity. I'm not sure which option is better. (mixing such option with roll may produce unexpected order)

sofroniewn:
Ah ok yes, I see -  we should probably fix that in a new PR, and maybe make a new issue specifically for that as it is unrelated to this issue. We do have the option to do a transpose, which could be combined with a roll to achieve what you want, but not in a simple way!!

sofroniewn:
This has been closed by #1665.

PR #1665, opened 2020-09-30
Skip axes with size 1 in roll operation

Czaki:
# Description

Described in #1664. Prevent from black screen when image has dimension of size 1. 

## Type of change

# References

# How has this been tested?

## Final checklist:

@sofroniewn I'm not sure if my test are enough. Maybe some additional case should be added?


sofroniewn:
> @sofroniewn I'm not sure if my test are enough. Maybe some additional case should be added?

It looks right now like we don't have any tests for roll, so can you maybe split this one into two. The first one `nsteps` are all large and you can just check that we iterate through all axes as expected.

The second test should check that we skip an axis if nsteps is `1` for that axis and should have a name/ doc string that makes it clear that is what we are testing.

How does that sound?

Czaki:
> How does that sound?

Ok, I will add something like this tomorrow.


Did you understand macos test fail?

sofroniewn:
> Did you understand macos test fail?

Mac OS fail is unrelated to this PR, we have an issue discussing it in #1656 - it relates to the latest dask release, which we need to fix asap.

Windows test is also unrelated to this PR and CI related

Czaki:
> > Did you understand macos test fail?
> 
> Mac OS fail is unrelated to this PR, we have an issue discussing it in #1656 - it relates to the latest dask release, which we need to fix asap.
> 
> Windows test is also unrelated to this PR and CI related

So why not pin working version until solution was found?

sofroniewn:
> So why not pin working version until solution was found?

The issue seems to be more about a timeout on our test failing rather than napari not working with the latest version of dask, so i'm not sure we need to pin, but if we go much longer without finding a solution we might need to. I can try looking at this more tonight/ tomorrow. Maybe @jni can too

sofroniewn:
no rush on this @Czaki, but in case it got lost in chat about out CI/ tests - which should now be resolved, i'm still waiting on splitting the test into two before merge, if that's still ok!!

Czaki:
@sofroniewn Done. I expand a little test.

sofroniewn (approved):
LGTM! thanks for the additional tests, I will merge once tests pass!!

PR #1666, opened 2020-10-01
Reduce the default dask cache size

jni:
# Description

It's not uncommon for me to have several long-running IPython sessions in which I test napari in a variety of scenarios, including with large dask arrays. This means I end up with multiple Python processes taking up 5GB+ RAM for no good reason. In general, I think it would be preferable to default to a modest cache size — I might even say 0, happy to discuss that option if people are happy — and let those users that need more, which I expect are in a minority, configure it for their purposes.

## Type of change

# References

# How has this been tested?

## Final checklist:
- [ ] ~~I have commented my code, particularly in hard-to-understand areas~~
- [ ] ~~I have made corresponding changes to the documentation~~
- [ ] ~~I have added tests that prove my fix is effective or that my feature works~~

### DIFF
--- a/napari/utils/dask_utils.py
+++ b/napari/utils/dask_utils.py
@@ -12,7 +12,7 @@ from .. import utils
 
 
 def create_dask_cache(
-    nbytes: Optional[int] = None, mem_fraction: float = 0.5
+    nbytes: Optional[int] = None, mem_fraction: float = 0.1
 ) -> Cache:
     """Create a dask cache at utils.dask_cache if one doesn't already exist.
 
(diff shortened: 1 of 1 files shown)


sofroniewn (approved):
I'm fine with this. I know we have some upcoming async cache PRs, so we might end up revisiting some of our caching in the future.

@jni can you add a code snippet here on how someone would recover the 50% behaviour.

Does the following work?
```python
import napari
from napari.utils.dask_utils import resize_dask_cache

with napari.gui_qt():
    resize_dask_cache(mem_fraction=0.5)
    ....
```
Have I gotten the call in the right place?

I do wonder too if the cache should be off/ zero by default/ Presumably this will be controllable via our settings too ...

jni:
Sorry @sofroniewn forgot to respond here — you did have it right. It can happen outside of the `gui_qt` call but it doesn't matter.

PR #1668, opened 2020-10-01
Remove stray print

pwinston:
# Description
Removes stray print added by `async-2.6` #1659 .

## Type of change
- [x] Bug-fix - remove print


sofroniewn (approved):
LGTM! Thanks

PR #1671, opened 2020-10-01
async-2.7: Vendor Cachetools

pwinston:
# Description
Add `napari._vendor.experimental.cachetools`
Used by `async-3`.

## Type of change
- [x] Add Vendor Package


sofroniewn (approved):
LGTM, we agreed on this vendoring approach during some of the earlier discussions. If we decide in the future to depend on this package or take an alternative approach we can always readdress this

Issue #1388, opened 2020-06-26
Pyramid view detection doesn't take grid mode into account


jni:
## 🐛 Bug

You're gonna love this one, @sofroniewn! 😂 Co-discovered by @DragaDoncila. 

## To Reproduce

```python
import numpy as np
import napari
pyramid0 = [np.random.random((4096 // 2**i, 4096 // 2**i)) for i in range(5)]
pyramid1 = [np.random.random((4096 // 2**i, 4096 // 2**i)) for i in range(5)]
v = napari.Viewer()
v.add_image(pyramid0, cmap='green')
v.add_image(pyramid0, colormap='green')
v.add_image(pyramid1, colormap='magenta', blending='additive')
```

Then enter grid mode and pan around.

## Actual behavior

![napari-pyramid-bug](https://user-images.githubusercontent.com/492549/85840775-639cd000-b7e0-11ea-88ff-7594858dd73e.gif)

## Expected behavior

Pyramid data should work fine in grid mode.

sofroniewn:
Sigh, indeed, this doesn't surprise me. Will add to my To Do's :-) 

I want to pull out the grid stuff from the layer itself, but we don't have a nice place to put it now. In the sort term the fix is probably just needing to make sure some coordinate info passes through the inverse of the grid transform, but I'll need to check to be sure

sofroniewn:
This has been closed by #1673

PR #1672, opened 2020-10-01
Fix translate boarder

liu-ziyang:
# Description
fix clipping when translate is set

## Type of change
- Bug-fix (non-breaking change which fixes an issue)

# References
closes #1606

# How has this been tested?
- [X] manually tested the change

## Final checklist:
- [X] My PR is the minimum possible work for the desired functionality
- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] I have added tests that prove my fix is effective or that my feature works

Issue #1207, opened 2020-05-06
Error while browsing pyramid dataset


jni:
## 🐛 Bug

I was panning and zooming around the [IDR COVID gut dataset](https://forum.image.sc/t/publication-of-idr0083-sars-cov-2-up-close-in-a-human-organoid/37163), minding my own business, when BAM! This happened:

```pytb
  File "/home/jni/miniconda3/envs/f7/lib/python3.7/site-packages/vispy/app/backends/_qt.py", line 825, in paintGL
    self._vispy_canvas.events.draw(region=None)
  File "/home/jni/miniconda3/envs/f7/lib/python3.7/site-packages/vispy/util/event.py", line 455, in __call__
    self._invoke_callback(cb, event)
  File "/home/jni/miniconda3/envs/f7/lib/python3.7/site-packages/vispy/util/event.py", line 475, in _invoke_callback
    self, cb_event=(cb, event))
  File "/home/jni/miniconda3/envs/f7/lib/python3.7/site-packages/vispy/util/event.py", line 471, in _invoke_callback
    cb(event)
  File "/home/jni/projects/napari/napari/_vispy/vispy_base_layer.py", line 246, in on_draw
    shape_threshold=self.node.canvas.size,
  File "/home/jni/projects/napari/napari/layers/base/base.py", line 692, in _update_multiscale
    self.data_level = data_level
  File "/home/jni/projects/napari/napari/layers/image/image.py", line 291, in data_level
    self.refresh()
  File "/home/jni/projects/napari/napari/layers/base/base.py", line 633, in refresh
    self.events.set_data()
  File "/home/jni/projects/napari/napari/utils/event.py", line 508, in __call__
    self._invoke_callback(cb, event)
  File "/home/jni/projects/napari/napari/utils/event.py", line 529, in _invoke_callback
    cb_event=(cb, event),
  File "/home/jni/projects/napari/napari/utils/event.py", line 523, in _invoke_callback
    cb(event)
  File "/home/jni/projects/napari/napari/_vispy/vispy_image_layer.py", line 73, in _on_data_change
    data = self.downsample_texture(data, self.MAX_TEXTURE_SIZE_2D)
  File "/home/jni/projects/napari/napari/_vispy/vispy_image_layer.py", line 167, in downsample_texture
    f"Shape of individual tiles in multiscale {data.shape} "
[...]

PR #1673, opened 2020-10-02
Pull out layer visual transform code

### LINKED ISSUE #1667: Add ability to query highest resolution coordinates for multiscale images
## 🚀 Feature
Add function or property which returns coordinates of layer from highest resolution and/or current level of multiscale pyramid.

## Motivation
Currently, adding a click callback to the layer which prints coordinates displays different coordinates when the user clicks on the same spot of the image, depending on the pyramid level currently displayed and the zoom level.

For example,

```
import napari
import dask.array as da
import numpy as np

with napari.gui_qt():
    v = napari.view_image(
        [da.random.random((10000, 10000)), da.random.random((5000, 5000))]
    )
    layer = v.layers[0]

    @layer.mouse_drag_callbacks.append
    def update_intensity(layer, event):
        print(layer.coordinates)
```
If you click on the bottom right hand corner when the image is fully zoomed out, you get roughly (5000, 5000). If you zoom in to the bottom right hand corner and click in the same spot, the coordinates get smaller as you zoom in, and become relative to the top left of the canvas, rather than the image itself.

This might be the desired behaviour for some applications, but some might require the "true coordinates" i.e. the corresponding coordinates in the full resolution level of the pyramid.

## Pitch

A function or property should be added to layers which returns coordinates relative to the top left corner of the full resolution image e.g. `layer.global_coordinates` although global is probably not the right word. 

This maybe should be the default? It's not clear that clicking on the same 'spot' should return different values depending on zoom level.

## Alternatives

Keep current behaviour and provide public interface to transforms which can give back the full resolution coordinates.

sofroniewn:
# Description
This PR will dramatically simply our visual transform code by no longer using transform on the individual layer visuals but instead using a single camera transform to get from canvas coordinates to world slice coordinates and then doing everything else with our new transform system.

This PR is still a WIP - multiscale isn't working yet - and it will need extensive testing before merge. Once this PR is in it will make things like adding a cursor model or camera model even easier too.

I might take this PR as an opportunity to fix #1667 too.

## Type of change

### DIFF
--- a/napari/layers/base/base.py
+++ b/napari/layers/base/base.py
@@ -13,7 +13,10 @@ from ...utils.misc import ROOT_DIR
 from ...utils.naming import magic_name
 from ...utils.status_messages import format_float, status_format
 from ..transforms import ScaleTranslate, TransformChain
-from ..utils.layer_utils import compute_multiscale_level, convert_to_uint8
+from ..utils.layer_utils import (
+    compute_multiscale_level_and_corners,
+    convert_to_uint8,
+)
 from ._base_constants import Blending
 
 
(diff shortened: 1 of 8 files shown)


sofroniewn:
I have now gotten multiscale working including for scaled and translated layers. This PR will supersede #1672. It aborbs that change, (thanks @ziyangczi for finding it!) into the new code structure.

What's nice about this PR is now there is a single transform going from canvas to worldslice through the camera and then each layer can do worldslice to world and world to data transforms, where data is now the highest resolution for multiscale data. A multiscale layer itself can then figure out the tile2data transform needed if it want's to show a lower resolution.

A nice improvement of this PR is now with async we get the correct zoomed out view on load, so @pwinston I think this fixes that `corner_pixels` issue you found the other day.

This will also close #1388! Though I did see the weird vispy error reported in #1207 again once where vispy looks like it's trying to show something with shape 1. It was a big big example with lots of data, and came after lots of panning so something else may have gone wrong.

This will also close #1667, `layer.coordinates` will now be in the highest resolution @DragaDoncila.

I've just noticed some things aren't quite right with coordinates and painting for scaled/ translated data. Will fix before marking ready for review.

sofroniewn:
Ok, I've fixed the painting labels at different scales. I'm pretty excited about this PR I think it simplifies a lot of things (see some comments above), fixes a couple of bugs like #1207 #1388 #1667 and an async one which doesn't have an issue yet.

It will probably need some extensive testing for things like multiscale, painting with different scales etc. Maybe @ziyangczi @jni @DragaDoncila can give it a go and see if they find any problems. This is also ready for review!!

napari/layers/base/base.py
```diff
@@ -76,14 +79,14 @@ class Layer(KeymapProvider, ABC):
...
         Coordinates of the top-left and bottom-right canvas pixels in the data
-        space of each layer. The length of the tuple is equal to the number of
-        dimensions of the layer.
-    position : 2-tuple of int
-        Cursor position in the image space of only the displayed dimensions.
+        coordinates of each layer. For multiscale data the coordinates are in
+        the space of the currently viewed data level, not the highest resolution
+        level.
+    position : tuple
+        Cursor position in world coordinates.
```
jni, line 89: @sofroniewn can come in a later PR (this is amazing) but what do you think about coordinates -> data_coordinates, and position -> world_position/real_position/similar? Currently "coordinates" and "position" is confusing to me without coming over to read this.

napari/layers/base/base.py
```diff
@@ -367,46 +369,36 @@ def translate_grid(self, translate_grid):
...
             self._transforms = self._transforms.expand_dims(new_axes)
-            self.coordinates = (0,) * (ndim - old_ndim) + self.coordinates
             self._dims_point = [0] * (ndim - old_ndim) + self._dims_point
+            self._position = (0,) * (ndim - old_ndim) + self._position
 
         self.dims.ndim = ndim
 
         self.refresh()
-        self._update_coordinates()
+        self._update_value_and_status()
```
jni, line 402: Ha, I love this.

napari/layers/base/base.py
```diff
@@ -735,67 +722,62 @@ def refresh(self, event=None):
...
-        tuple of indices and update the cursor coordinates.
-        """
-        coords = list(self.coordinates)
-        for d, p in zip(self.dims.displayed, self.position):
-            coords[d] = p
-        self.coordinates = tuple(coords)
+    @property
+    def coordinates(self):
+        """Cursor position in data coordinates."""
+        return tuple(self._transforms[1:].simplified.inverse(self.position))
```
jni, line 733: What is 1: for here? Maybe add a comment to explain it?

Also, is simplified cached? If not, it's probably not worth accessing it over and over.

jni (approved):
@sofroniewn this is just awesome! I'm going to let this sit for now in case anyone else wants to play with it before it merges, but all good from me if you want to merge when you get up! 🎉

jni:
@sofroniewn what exactly do you mean you've fixed painting at different scales — are you saying we can paint into a multiscale dataset now???

napari/layers/base/base.py
sofroniewn: yes! I want that, probably in the next PR or so, but before `0.4.0`. That will be a stronger backwards compatibility break, so I want to put it in its own thing. Ultimately I think we want to pull this off the layer entirely!

napari/layers/base/base.py
sofroniewn: Simplified is not cached, but that would be nice to do. Maybe something i can get help with in a follow-up PR!!

sofroniewn: comment added

sofroniewn:
> @sofroniewn what exactly do you mean you've fixed painting at different scales — are you saying we can paint into a multiscale dataset now???

Ah no, not yet, but it's getting closer. The coordinates should work out now, the thing we need to do is update all levels of resolution on the fly. We're not doing that yet, but at this point it should actually just be some fairly straight forward maths. If you thought this was high priority I definitely see how we could get it done!!

Thanks for the quick review @jni. I'm going to merge now and press on with some of the camera stuff, possibly more cursor stuff still to come too. If this does end up introducing some corner case bugs, will be good to fix with tests before the `0.4.0` release

PR #1676, opened 2020-10-02
async-4: Config File

pwinston:
# Description
Now `NAPARI_ASYNC` can point to the path of a config file, although it can still be "1" for defaults.

Named tuple `AsyncConfig` stores the values read the from the config file, so anyone can access them.

Initial config file format (more to come later):

```
{
    "synchronous": false,
    "num_workers": 6,
    "log_path": "/Users/pbw/dev/napari/chunk.log"
}
```

## Type of change
- [x] New feature experimental async code

## Files

async-4-files

# How has this been tested?
- [x] Manual testing

## Final checklist:


sofroniewn (approved):
LGTM! In the long run we might want to unify this config approach with something like #1183, but I think this is fine for now. I will merge

PR #1678, opened 2020-10-02
async-5: Timing Infrastructure

pwinston:
# Description

Mostly `perf` changes to support always-on timing in `ChunkRequest`. The `perf` changes are backward compatible, no user visible change unless you use the new `block_timer` context object. 

We want `ChunkRequest` to always time the loads, so that in `async-7` we can do auto-async. Rather than do custom timing code, we make some clever `perf` changes that gives use `block_timer`, like a `perf_timer` but always enabled.

# Perf Changes

There is new context manager called `block_timer` which is basically the same as `perf_timer` was except two key changes:
1. `block_timer` is always on, whether PERFMON is enabled or not.
2. `block_timer` yields a `PerfEvent`, so you can get at the timing information.

The pay off is `ChunkRequest` can use `block_timer`, thus getting all that nice infrastructure for free.

Meanwhile `perf_timer` is just a `block_timer` if PERFMON is enabled, otherwise it's null.

Still not a great idea to leave `perf_timer`'s in the code, but the overhead would be very close to zero if you did.

## Type of change
- [x] New features in `perf` and in experimental async code

## Files

async-5-files

# How has this been tested?
- [x] Manual testing with and without PERFMON and with and without ASYNC

## Final checklist:


pwinston:
This is ready for review. Pending CI.

sofroniewn (approved):
This looks good to me too. I think it's nice to reuse some of the timing infrastructure from the performance monitoring in async, and the new generic `block_timer` looks good. Might be interesting to think about other places it could be used in the long run.

PR #1679, opened 2020-10-02
async-6: DelayQueue

pwinston:
# Description

* Add new `DelayQueue` which prevents spamming the worker thread with requests for chunks we don't need.
* This makes the noise2self notebook work in Jupyter with threads only, processes not needed.
    * Because it avoids doing pointless computation in the background while scrolling.
    * When you pause for 100ms the computation will start.

## Type of change

## Files
async-6-files

# How has this been tested?
- [x] Manual testing

## Final checklist:


napari/components/experimental/chunk/_loader.py
```diff
@@ -46,16 +47,20 @@ class ChunkLoader:
     ----------
     synchronous : bool
         If True all requests are loaded synchronously.
+    num_workers : int
+        The number of worker threads or processes.
```
sofroniewn:
```suggestion
        The number of worker threads.
```
We just support processes at this point right? Or can it be either?

sofroniewn (approved):
LGTM! I left one minor suggestion on doc string, but otherwise looks good to go.

I think when we have this full pass at async merge we might want to look again at the DelayQueue and if we need it in all cases, for example as you mentioned when we just click to a new location on the slider.

pwinston:
`async-12` adds processes, so yeah I committed the suggestion to remove that for now.

sofroniewn:
I will merge once CI re-runs

PR #1680, opened 2020-10-02
fix is_running, fix tests

Czaki:
# Description

Bugfix for `is_running` property. Set `_is_running` variable to `False` after job finish. 

Change test for  `thread_worker` to be sure that callback functions was executed

EDIT
Raise AttributeError for non existing parameter of WorkerBase

## Type of change

# References

# How has this been tested?

## Final checklist:

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -208,7 +208,7 @@ win_task:
   install_script:
     - python -m pip install --retries 3 --upgrade pip
     - python -m pip install --retries 3 -r requirements/test.txt
-    - python -m pip install -e .[pyqt5]
+    - python -m pip install -e .[pyside2]
 
   check_script:
     - conda list
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -208,7 +208,7 @@ win_task:
   install_script:
     - python -m pip install --retries 3 --upgrade pip
     - python -m pip install --retries 3 -r requirements/test.txt
-    - python -m pip install -e .[pyside2]
+    - python -m pip install -e .[pyqt5]
 
   check_script:
     - conda list
--- a/napari/_qt/_tests/test_threading.py
+++ b/napari/_qt/_tests/test_threading.py
@@ -32,12 +32,14 @@ def test_thread_worker(qtbot):
     def func():
         print("[func]", file=sys.stderr)
         func_val[0] = 1
+        print("[func] end", file=sys.stderr)
         return 1
 
     def test(v):
         print("[test]", file=sys.stderr)
         test_val[0] = 1
         assert v == 1
+        print("[test] end", file=sys.stderr)
 
     thread_func = qthreading.thread_worker(
         func, connect={'returned': test}, start_thread=False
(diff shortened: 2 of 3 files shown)


sofroniewn (approved):
This looks good to me, thanks for the fix @Czaki! I'm just rerunning some tests which should pass now and then merge

sofroniewn:
@Czaki looks like this now has one failing windows test

```python-traceback
================================== FAILURES ===================================
_____________________________ test_thread_worker ______________________________
qtbot = <pytestqt.qtbot.QtBot object at 0x000001F6CD2D6BE0>
    def test_thread_worker(qtbot):
        """Test basic threadworker on a function"""
        func_val = [0]
        test_val = [0]
        def func():
            func_val[0] = 1
            return 1
        def test(v):
            test_val[0] = 1
            assert v == 1
        thread_func = qthreading.thread_worker(
            func, connect={'returned': test}, start_thread=False
        )
        worker = thread_func()
        assert isinstance(worker, qthreading.FunctionWorker)
        assert func_val[0] == 0
        with qtbot.waitSignal(worker.finished):
>           worker.start()
E           pytestqt.exceptions.TimeoutError: Signal  not emitted after 1000 ms
napari\_qt\_tests\test_threading.py:45: TimeoutError
```

not clear to be if this is just a one off or sporadic, or maybe I made a mistake when fixing the merge conflict with #1681. Can you look at this one more time before I merge? Thanks!!

Czaki:
@sofroniewn I see no logical reason why this test fail and other pass. I need to create windows environment and check this.

Czaki:
@sofroniewn I add print and it pass. I do not understand.

sofroniewn:
> @sofroniewn I add print and it pass. I do not understand.

That's very strange - why don't you try taking them away again. If that fails then we might be on to something, otherwise maybe the failures are just at random times

sofroniewn:
Test fails again @Czaki! Now we're onto something!! It looks like you added a `sys` import too in addition to the print statements. Do you want to try adding that? Maybe that fixes things. If you (or anyone else) could reproduce locally that would make this much easier to figure out :-)

Czaki:
@sofroniewn 
I finally reproduce error on my computer. It happens only when running whole data set (and I need to reduce time limits) 
It looks like some previous executed test create events which are proceed before signal. increase timeout solve this. 

I already meet something similar in the past. I do not know how to debug this. As you decide to use PySide2 I cannot ask on PyQt mailing list if anyone has this experience.

sofroniewn:
> I finally reproduce error on my computer. It happens only when running whole data set (and I need to reduce time limits)
It looks like some previous executed test create events which are proceed before signal. increase timeout solve this.

Looks like the timeout is now 20 seconds, is that right? I guess that's ok, it was only an issue on the windows test. Really nice job chasing this down. If we run into this problem again at least we have this investigation to guide us. I'm ok merging now, unless there's additional stuff you want to look into @Czaki? (cc @jni)

jni (approved):
LGTM, thank you @Czaki for the fix and @sofroniewn for the earlier review!

Czaki:
Merge

PR #1681, opened 2020-10-02
Make signals public attribute of WorkerBase

Czaki:
# Description

Make `signal` parameter of `WorkerBase` public to allow identify worker object with `self.sender()` without using private attribute.
It is important when keeping list of workers to be finished before close application. 

## Type of change

# References

# How has this been tested?

## Final checklist:


sofroniewn (approved):
I'm fine with this, will merge now

PR #1683, opened 2020-10-02
async-7: Auto-async

pwinston:
# Description

* Add `LoadType` and `LoadStats` and `ChunkLoader._should_load_async`
* We will now switch from sync to async and back based on the load speed of the chunks.

## Type of change
- [x] Bug fix related to caching in `_chunked_slice_data.py`.

## Files

async-7-files

# How has this been tested?
- [x] Manual testing

## Final checklist:


sofroniewn (approved):
Auto methods and stats look good! I see now we have a threshold of 30ms for the transition, but plans to make that a config setting in the future. I'll need to play with things more to get a sense for the performance with different values, but the approach looks good.

All tests are passing, so I will merge now.

PR #1684, opened 2020-10-03
async-8: config cleanup

pwinston:
# Description

* Add `auto_sync_ms` so it's no longer hard-coded.
* Rename `delay_seconds` to `delay_queue_ms`.
* Doc strings and other comments.

## Type of change

## Files
async-8-files

# How has this been tested?
- [x] Manual testing

## Final checklist:


sofroniewn (approved):
Looks good to me! Making the automatically synchronous threshold a config setting is nice, and switching to units of ms in this context makes sense too. Nice to have it labeled to so that it is clear.

Tests are still finishing up, I will merge once they all pass!

PR #1685, opened 2020-10-03
async-9: async unit testing infra

pwinston:
# Description
* **cirrus.yml**
    * New test pass: Python 3.8 with new flag `--async_only`
* **conftest.py**
    * New flag `--async_only`
    * New `configure_loading` fixture:  enable async/async depending on `--async_only` flag
    * New `skip_sync_only` fixture: in `--async-only` mode skip tests marked `sync_only`.
* **test_dask_layers.py**
    * Mark some tests `sync_only` that would fail with async.
* **experimental.chunk**
    * New `wait_for_async` not used yet, but for better async tests

## Type of change
- [x] New feature **in experimental async code** plus in unit test code

## Files

async-9-files

# How has this been tested?
- [x] Manual testing

## Final checklist:


pwinston:
Marking draft so I can experiment with this a bit first. Today this really doesn't test async all that much, because the "async" tests are mostly loading synchronously due to the arrays being `ndarray`. By default we load `ndarray`'s synchronously even with `NAPARI_ASYNC`. So it's TBD if we can do better than that, maybe force `ASYNC`, but then we somehow need to wait for the loads to complete.

napari/components/experimental/chunk/_config.py
```diff
@@ -18,17 +18,25 @@
 
 # NAPARI_ASYNC=1 will use these default settings:
 DEFAULT_ASYNC_CONFIG = {
```
sofroniewn: Weren't these changes already in async-8? See https://github.com/napari/napari/pull/1684/files#diff-19a4eb1e1e6ca1d1e9a5d5905af8e06bR20-R21. Any idea why they appear again here in this diff?

pwinston: I guess I hadn't merged in master with `async-8` yet. I merged it now. The lines went away.

pwinston:
I'm marking this ready for review. This is a drop in the bucket on testing async, but it's useful. This makes CI run a separate past that runs every unit test except 3 which are marked `sync_only`.

The limitation though is every test that uses `ndarray`, most of them, will be run synchronously. It's still testing new code, it's testing the `ChunkLoader`, but the `ChunkLoader` is loading synchronously.

So the TODO here is figuring how how to test true asynchronous loading, testing the thread pool and the `QtChunkReceiver` and `Image.load_chunk` and all that. Either making all the tests run true-async, or created in async-only tests,  or both.

But that's a future task I think. Merging this moves the ball forward and we can finish merging through `async-12` then we are prioritize automated async testing vs. visuals, etc.

sofroniewn (approved):
This looks good to me. I think it's ok now that our async tests runs with "sync" data. We can add more "async" async tests as we go! I will merge

PR #1692, opened 2020-10-05
Allow dask development version to work with napari

### LINKED ISSUE #1670: Support development versions of Dask
Dask uses versioneer for development versions

```python
import dask
dask.__version__
```
```
2.28.0+8.g51b76e5d
```

Unfortunately it seems that Napari does some direct checking of version numbers here, which assumes that the user is using a released version.  See the traceback below.

```python-traceback
~/miniconda/lib/python3.8/site-packages/napari/utils/dask_utils.py in configure_dask(data)
    168     if _is_dask_data(data):
    169         create_dask_cache()  # creates one if it doesn't exist
--> 170         dask_version = tuple(map(int, dask.__version__.split(".")))
    171         if dask_version < (2, 15, 0):
    172             warnings.warn(

ValueError: invalid literal for int() with base 10: '0+8'
```

I recommend using either `distutils.LooseVersion` (this is what Dask uses) or perhaps instead just straight string comparison.  I understand the drawbacks of comparing version strings, but I'm personally willing to guarantee to you that a version "2.100" will never exist :)

GenevieveBuckley:
# Description

This PR allows the development version of dask to work with napari. We do this by using `LooseVersion` from `distutils` to check the dask version numbers, instead of the string parsing we had earlier.

## Type of change

# References

Closes https://github.com/napari/napari/issues/1670

# How has this been tested?

- [x] example: all existing tests pass with my change

## Final checklist:
- [x] I have added tests that prove my fix is effective or that my feature works (this change is too small to really require an extra test of it's own, the existing tests should be fine for this)

### DIFF
--- a/napari/_tests/test_dask_layers.py
+++ b/napari/_tests/test_dask_layers.py
@@ -1,4 +1,5 @@
 from contextlib import contextmanager
+from distutils.version import LooseVersion
 
 import dask
 import dask.array as da
@@ -6,8 +7,6 @@ import numpy as np
 import pytest
 
 from napari import layers, utils, viewer
-
-dask_version = tuple(map(int, dask.__version__.split(".")))
 
 
 def test_dask_array_creates_cache():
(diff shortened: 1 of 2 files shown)


sofroniewn (approved):
Love this! Thanks @GenevieveBuckley this looks great to me. I will merge!!

PR #1694, opened 2020-10-05
autogen add_* methods

### LINKED ISSUE #1677: Autogenerate add_* methods (revert #303)
## Refactor

In #303 we decided to explicitly put our `add_*` methods onto the viewer and no longer use the very nice `add_to_viewer` decorator that @kne42 had made.

As @kne42 noted at the time

> Ideally we shouldn't have to do anything; this is PyCharm's fault for statically providing completions and documentation in a dynamic language 🙄.
>
> I feel like every situation here is a loss, so I'll just go with whatever is best for the user and I think that's this PR, but with the signatures/docstrings properly copied from their respective classes, not the whole "please see X documentation."

I think now experiences like #1616 where i add a keyword arg to the base layer and have to add 14 copies of the doc string, make me thing we might want to revert this change.

For all our layers apart from `Image` it should be very simple. For `Image` we have some additional complexity around the splitting of axes channels, but that can maybe be addressed separately. 

@jni @kne42 do either of you have additional thoughts on this or a recommendation?

sofroniewn:
# Description
This PR closes #1677 by reverting #303 and returning to autogenerating our `add_*` using the great `create_func` written by @kne42. This registration is now done on our `AddLayersMixin` so that layers themselves don't need to have any concept of the `ViewerModel`. We might want to drop `AddLayersMixin` object entirely at some point.

Note that `add_image` is not autogenerated due to the additional complexity of the splitting of channels. We might want to revisit this at some point, but this PR already represents a dramatic simplification.

Note I had to change a line or two in our new layer button code to maintain backwards compatibility.

## Type of change


sofroniewn:
We're running into some fails now with our [test_view_layers](https://github.com/napari/napari/blob/master/napari/_tests/test_view_layers.py) because we can't read doc strings from the `add_*` methods. Not sure how we want to think about adjusting those tests now that we autogenerate the `add_*` methods. It looks like the `view_*` methods are still getting generated properly.

sofroniewn:
This is now ready for review @jni @kevinyamauchi, i've fixed the tests dropping part that inspected source code for the `view_*` methods that we don't need anymore.

I'd like to get this merge before continuing on #1616 as it will dramatically reduce the number of doc strings I need to edit!!

jni (approved):

PR #1696, opened 2020-10-05
fix image extent

### LINKED ISSUE #1686: layer._extent_data inconsistent between image and points/shapes/tracks
## 🐛 Bug
The values for `_extent_data` are calculated differently for the image layer and points/shapes/tracks. In the image layer, the maximum value for the extent is the `maximum index + 1` (i.e., the result of `data.shape`) and in the points/shapes/tracks layer it is the maximum index (i.e., the results of `np.max(data)`. It appears that our Dims model assumes the maximum value of the extent is `maximum_index + 1` so the range of the slider is incorrectly calculated for the points/shapes/tracks layers (it is 1 step less).

Is `layer._extent_data` defined as `(min_index, max_index + 1 )`, where `min_index `and `max_index` are the layer data array indices?

## To Reproduce
1. create 3D data with indices from 0 through 4 for axis 0 for image, points, shapes, and tracks layers
2. view those data in napari in 2D mode (axis slider is the 0th axis). Note that the axis slider correctly goes from 0-4
Screenshot from 2020-10-03 13-24-08

3. view the `layer._extent_data` property for each layer

```python
viewer.layers['image']._extent_data
Out[1]: 
array([[   0.,    0.,    0.],
       [   5., 1000., 1000.]])

viewer.layers['points']._extent_data
Out[2]: 
array([[   0,  236,    0],
       [   4,  636, 1000]])

viewer.layers['tracks']._extent_data
Out[3]: 
array([[   0,  236,    0],
       [   4,  636, 1000]])
viewer.layers['shapes']._extent_data
Out[4]: 
array([[ 0., 10., 10.],
       [ 4., 20., 20.]])
```
4. If we  delete the image layer, the axis slider now goes from 0 - 3.
Screenshot from 2020-10-03 13-14-07

<details>
<summary>code</summary>

```python

test_im = np.random.random((5, 1000, 1000))

tracks_data = np.asarray([
    [1, 0, 236, 0],
    [1, 1, 236, 100],
    [1, 2, 236, 200],
    [1, 3, 236, 500],
    [1, 4, 236, 1000],
    [2, 0, 436, 0],
    [2, 1, 436, 100],
    [2, 2, 436, 200],
    [2, 3, 436, 500],
    [2, 4, 436, 1000],
    [3, 0, 636, 0],
    [3, 1, 636, 100],
    [3, 2, 636, 200],
    [3, 3, 636, 500],
    [3, 4, 636, 1000]
])

shapes_data = [
    [
        [0, 10, 10],
        [0, 10, 20],
        [0, 20, 20],
        [0, 20, 10]
    ],
    [
        [4, 10, 10],
        [4, 10, 20],
        [4, 20, 20],
        [4, 20, 10]
    ]
]

with napari.gui_qt():
    viewer = napari.view_image(test_im, name='image')
    viewer.add_tracks(tracks_data, name='tracks')
    viewer.add_points(tracks_data[:, 1:], name='points')
    viewer.add_shapes(shapes_data, name='shapes')
```
</details>

## Expected behavior

The values for the 0th dimension in `layer._extent_data` should be the same for the image

## Environment
```
napari: 0.3.8.dev14+g504ef2f
Platform: Linux-5.4.0-48-generic-x86_64-with-debian-bullseye-sid
Qt: 5.14.2
PyQt5: 5.14.2
NumPy: 1.19.1
SciPy: 1.5.2
Dask: 2.24.0
VisPy: 0.6.4

MAX_TEXTURE_SIZE: 32768

Plugins:
- napari_plugin_engine: 0.1.7
- ome_zarr: 0.0.14
- svg: 0.1.4
```
## Additional context

This is related to https://github.com/napari/napari/issues/922

sofroniewn:
# Description
This PR will close #1686. I had to copy and redefine the `layer.shape` property on the image layer, but I think we are going to remove anyway for `0.4.0`, but let's wait to do that in a new PR.

It does change the `_extend_data` property, so I have to fix a bunch of tests to accommodate the new behavior.

## Type of change

### DIFF
--- a/napari/components/dims.py
+++ b/napari/components/dims.py
@@ -104,7 +104,7 @@ class Dims:
         """Number of slider steps for each dimension.
         """
         return [
-            int((max_val - min_val - step_size) // step_size) + 1
+            int((max_val - min_val) // step_size) + 1
             for min_val, max_val, step_size in self._range
         ]
 
@@ -327,7 +327,7 @@ class Dims:
             Value of the point.
         """
         axis = self._assert_axis_in_bounds(axis)
-        step = np.round(np.clip(value, 0, self.nsteps[axis])).astype(int)
+        step = np.round(np.clip(value, 0, self.nsteps[axis] - 1)).astype(int)
 
         if self._current_step[axis] != step:
             self._current_step[axis] = step
--- a/napari/components/dims.py
+++ b/napari/components/dims.py
@@ -327,7 +327,7 @@ class Dims:
             Value of the point.
         """
         axis = self._assert_axis_in_bounds(axis)
-        step = np.round(np.clip(value, 0, self.nsteps[axis] - 1)).astype(int)
+        step = np.round(np.clip(value, 0, self.nsteps[axis])).astype(int)
 
         if self._current_step[axis] != step:
             self._current_step[axis] = step
--- a/napari/components/viewer_model.py
+++ b/napari/components/viewer_model.py
@@ -470,8 +470,8 @@ class ViewerModel(AddLayersMixin, KeymapHandler, KeymapProvider):
             Size of the grid that is being used.
         """
         extent = self._sliced_extent_world
-        scene_size = extent[1] - extent[0]
-        translate_2d = np.multiply(scene_size[-2:], position)
+        scene_shift = extent[1] - extent[0] + 1
+        translate_2d = np.multiply(scene_shift[-2:], position)
         translate = [0] * layer.ndim
         translate[-2:] = translate_2d
         layer.translate_grid = translate
--- a/napari/components/viewer_model.py
+++ b/napari/components/viewer_model.py
@@ -307,7 +307,7 @@ class ViewerModel(AddLayersMixin, KeymapHandler, KeymapProvider):
         scene_size = extent[1] - extent[0]
         corner = extent[0]
         shape = [
-            np.round(s / sc).astype('int') if s > 0 else 1
+            np.round(s / sc).astype('int') + 1 if s > 0 else 1
             for s, sc in zip(scene_size, scale)
         ]
         empty_labels = np.zeros(shape, dtype=int)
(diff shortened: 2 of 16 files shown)


sofroniewn:
I have created #1697 to discuss the dropping of `layer.shape` and the renaming of `layer._extent_data`

sofroniewn:
I've fixed a couple tests, but @kevinyamauchi it would be amazing if you could take this PR over and fix/ improve the remaining tests (there are a lot of "special" numbers) present, and check that everything now works as expected. We might want a couple new tests as well that really make sure this is the way we want it to be!

kevinyamauchi:
On it!

napari/components/dims.py
```diff
@@ -104,7 +104,7 @@ def nsteps(self):
         """Number of slider steps for each dimension.
         """
         return [
-            int((max_val - min_val - step_size) // step_size) + 1
+            int((max_val - min_val) // step_size) + 1
```
kevinyamauchi, line 107: Should this be

```suggestion
            int((max_val - min_val) // step_size)
```

kevinyamauchi: Ah nevermind!

napari/components/viewer_model.py
```diff
@@ -470,8 +470,8 @@ def _subplot(self, layer, position):
             Size of the grid that is being used.
         """
         extent = self._sliced_extent_world
-        scene_size = extent[1] - extent[0]
-        translate_2d = np.multiply(scene_size[-2:], position)
+        scene_shift = extent[1] - extent[0] + 1
```
kevinyamauchi, line 473: I changed this to make it so next image starts one pixel after the previous image ends. It looks like the same behavior on master and the test passes. Is this right, @sofroniewn ?

kevinyamauchi (approved):
The tests are now passing. Perhaps we can let this sit for a bit in case anybody wants to weight in or we want to add any tests. Otherwise, I think this is good to go!

napari/components/viewer_model.py
sofroniewn: Yes, I think this makes sense, thanks for that fix!!

sofroniewn:
This looks great @kevinyamauchi!! Happy to let sit for a little, and then will merge. Thanks so much for your help here!!

kevinyamauchi:
No problem. Sounds good!

PR #1698, opened 2020-10-06
async-10: add experimental vendor package humanize

pwinston:
# Description
* Add _vendor/experimental/humanize, used by `async-11` IPython commands
* add `_PYTEST_RAISE` flag to `conftest.py`. This is very useful for debugging with VSCode:

```
# _PYTEST_RAISE=1 will prevent pytest from handling exceptions.
# Use with a debugger that's set to break on "unhandled exceptions".
# https://github.com/pytest-dev/pytest/issues/7409
```

## Type of change
- [x] Experimental vendor'd library
- [x] Small opt-in conftest change

# Files

async-10-files


sofroniewn (approved):
Looks good to me, vendoring of humanize is very clean and opt in way to prevent pytest from handling exceptions looks good too. I will merge now.

PR #1703, opened 2020-10-06
Use evented list in dims take 2

sofroniewn:
# Description
This PR supersedes #1486 by using the evented list in the dims.

This is a WIP right now, tests are broken, but it also seems to be quite a bit slower so I will do some performance monitoring to see what's going on.

## Type of change

- [x] Refactor

### DIFF
--- a/napari/_qt/qt_viewer.py
+++ b/napari/_qt/qt_viewer.py
@@ -1,5 +1,4 @@
 import os.path
-from copy import copy
 from pathlib import Path
 
 import numpy as np
@@ -527,7 +526,7 @@ class QtViewer(QSplitter):
         mapped_position = transform.map(list(position))[:nd]
         position_world_slice = mapped_position[::-1]
 
-        position_world = copy(self.viewer.dims.point)
+        position_world = list(self.viewer.dims.point)
         for i, d in enumerate(self.viewer.dims.displayed):
             position_world[d] = position_world_slice[i]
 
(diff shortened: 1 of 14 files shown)


sofroniewn:
This PR is getting close, but I have some doubts about it. There are some nice things it does, for example you can now do

```python
viewer.dims.point[0] = 22
viewer.dims.point[:] = [22, 0, 0]
```
to change the location of the sliders.

You can't do 
```python
viewer.dims.point = [22, 0, 0]
```
anymore, but @jni if you'd like me to I can support that (by making the setter just call `viewer.dims.point[:]`)

You could do all sorts of weird things like try and edit the length of the list, insert, or remove elements and things would break. We could get more intense and add something like a keyword arg to our `EventedList` called `fixed_length` or something that would prevent that, though there are times when we do want to do that ourselves.

I also had to change the default behavior of the `EventedList` when editing like `my_list[:]` and how events get emitted, and now a number of tests are failing. I could revert but then the `viewer.dims.point[:] = ` behavior might not be supported.

Guidance here from @jni is needed

jni:
> but @jni if you'd like me to I can support that (by making the setter just call viewer.dims.point[:])

I think `viewer.dims.point = [...]` should be supported, it's super convenient, probably my favourite way to interact with `point`!

> I also had to change the default behavior of the EventedList when editing like my_list[:] and how events get emitted, and now a number of tests are failing. I could revert but then the viewer.dims.point[:] = behavior might not be supported.

I'm not sure about what the changes are or whether they make sense, just want to give you some quick feedback before I'm back at work next week, but imho the key is whether you can *detect* `viewer.dims.point[:] = ...` and raise a meaningful error, or whether it'll just succeed but not emit any events. The latter would be bad.

What is the current behaviour, and what would be the new behaviour? If it makes sense generally then the tests should be fixed. To me it seems that this should work and emit events and the tests should be modified to reflect that.

sofroniewn:
This is superseded by #1919 which doesn't use the `EventedList` but we could add one day

PR #1704, opened 2020-10-06
async-12: Processes

pwinston:
# Description
* Adds `use_processes` flag to config, so now we optionally can use processes instead of threads.
* Turns out we don't need this noise2self
* So there are really no known needs for processes but the change is pretty minor or might as well allow it for now.
* We could cut it if there just ever is a good use for it.

# Files

async-12-files

## Type of change
- [x] New feature in async experimental code

# How has this been tested?
- [x] Manual testing


sofroniewn (approved):
Looks good to me! I will merge once tests pass. It's pretty cool that enabling processes is such a minor addition to the code

PR #1707, opened 2020-10-07
async-14: Prototyping Octree Visuals

pwinston:
# Description

* Wedge PR to get experimental octree-based image rendering in place.
* The experimental part is very WIP and will crash except for some specific test cases.
* For octree visuals set `NAPARI_ASYNC` to the path of a JSON config file:
    * If `octree_visuals` setting is missing or false: standard async.
    * If `octree_visuals` setting is true: octree visuals

## Type of change
- [x] Mainstream files changed (most guarded by `NAPARI_ASYNC`)
- [x] New experimental files and changes.

# Non-Experimental Changes

Safe/minimal change to `vispy_image_layer.py`:

The old `_on_data_change()` method took the data from `self.layer._data_view`, processed it, and passed the data to `self.node`. Now instead `_on_data_change()` calls a new method `_set_node_data(self, node, data):` which does the real work. The benefit is the derived class `VispyTiledImageLayer` can call `VispyImageLayer._set_node_data` over and over, to set the data for every node/tile.

|File|Comments|
|---|---|
|napari/_vispy/vispy_image_layer.py|Extract `_set_node_data()` from `_on_data_change()`|

# Non-Experimental Changes (Guarded)

These 5 changes are guarded by `NAPARI_ASYNC`:

|File|Comments|
|---|---|
|napari/_qt/qt_viewer.py|Create `dockerRender` = `QtRenderContainer`|
|napari/_qt/qt_main_window.py|Add `dockRender`|
|napari/_qt/layer_controls/qt_layer_controls_container.py|Use regular image controls for `OctreeImage`|
|napari/components/add_layers_mixin.py|Create `Image` or `OctreeImage`|
|napari/_vispy/utils.py|Map `OctreeImage` -> `VispyTiledImageLayer`|

# Experimental: Changes
|File|Comments|
|---|---|
|napari/components/experimental/chunk/_config.py|New Octree config setting|

# Experimental: New
|File|Comments|
|---|---|
|napari/_qt/experimental/qt_render.py|`QtRender` widget|
|napari/_qt/experimental/qt_render_container.py|`QtRenderContainer`|
|napari/_vispy/experimental/vispy_tiled_image_layer.py|Tiled rendering with multiple `ImageVisuals`|
|napari/layers/image/experimental/octree_image.py|Will replace `Image` class|
|napari/layers/image/experimental/_octree_image_slice.py| Will replace `ImageSlice` class|
|napari/layers/image/experimental/_octree.py|Octree Datastructure|

# Files

pr_1707_files

# TiledImageVisual?

Above we are using `VispyTiledImageLayer` which creates one `ImageNode` per tile. On Vispy's Gitter David Hoese said that will not perform well. He said we should create a new single visual called `TiledImageVisual` which manages tiles with a texture atlas. So it can draw all the tiles in one draw command.

However, we are starting with `VispyTiledImageLayer` because it was simple. And we can upgrade it when we need the performance.

# References

* **Articles**
    * [Wikipedia: Octree](https://en.wikipedia.org/wiki/Octree)
    * [Loose Octrees](https://anteru.net/blog/2008/loose-octrees/)
    * [Introduction to Octree](https://www.gamedev.net/tutorials/programming/general-and-gameplay-programming/introduction-to-octrees-r3529/) (gamedev)
     * [Slippy Map](https://wiki.openstreetmap.org/wiki/Slippy_Map)
* **Python**
    * [ActiveState](https://code.activestate.com/recipes/498121-python-octree-implementation/)
    * [pyoctree](https://github.com/jcummings2/pyoctree)
* **C++**
    * [SimpleOctree](https://github.com/brandonpelfrey/SimpleOctree) (2013)
    * [PointCloudLibrary](https://github.com/PointCloudLibrary/pcl/tree/master/octree) (5000 lines, gold standard for correctness?)

# How has this been tested?
- [x] manual testing


napari/layers/image/experimental/octree.py
```diff
@@ -0,0 +1,197 @@
...
+    tiles might be smaller if the array did not divide evenly.
+    """
+    if array.ndim != 3:
+        raise ValueError(f"Unexpected array dimension {array.ndim}")
+    (rows, cols, _) = array.shape
+
+    tiles = []
+
+    row = 0
+    while row < rows:
```
d-v-b, line 57: A lot of the tooling for dividing arrays into chunks exists in dask -- I'm pretty sure the thing you are doing here is done in n dimensions without while loops in the `slices_from_chunks` function in `dask.array.core` https://github.com/dask/dask/blob/45fd9092049a61c1d43067a0132574b3fe715153/dask/array/core.py#L206

pwinston: Yeah. We haven't really had the discussion of using dask inside the renderer itself, as in everything turns into dask whether it started as dask or not. Ask of today I'm assuming we don't want to do that. But that's a valid conversation to have, even soon-ish.

d-v-b: That function I referenced doesn't use dask arrays at all. It's just a convenience function to generate slices for dividing an array into chunks.

pwinston: This specific code of building a tree from a single image is just for testing. Almost all multi-scale viewing will not go through this code, because it will be a pre-chunked format.

To fix #1300 though we might need to build a tree on the fly. If we did that I'd consider a better way to build the tree. Although the real solution there might be more about increment/amortized building. Because no matter how efficiently we did it, it might take too long if we did it in one gulp. At any rate not worrying about #1300 or on-the-fly yet, except just for this quick-n-dirty testing.

d-v-b: That's all fine, I'm just letting you know that the _specific thing you are doing in this loop_ (slice an array into chunks) is already implemented via the dask core utils, so you could remove some of your code and just use / copy the dask function, if you want. For myself, whenever I'm breaking an array into little chunks I often find clean functions for that in the dask array core utils, even when I'm not working with dask arrays.

pwinston: Yes that's good to know about, thanks.

pwinston:
Some very early tiled rendering work. This is an octree built from a single 1024x1024 image broken into 64x64 pixel tiles. Below the tree is shown at  levels 4, 3 and 0. There is no "octree rendering" yet, all it can do is draw all of the tiles at a single level.

The `OctreeImage` class is meant to eventually become the new `Image` class. Its `OctreeImageSlice` is like the old `ImageSlice` except it creates an `Octree` on the fly when the slice is loaded.

`VispyTiledImageLayer` can draw tiles as regular `ImageNode`'s. It doesn't know about octrees but it can draw any set of tiles with given X/Y/SIZE. It's derived from `VispyImageLayer2` which is meant to replace `VispyImageLayer`, in that it's backward compat.

The expectation is creating individual `ImageNode`'s will be too slow and we'll need to create a new `TiledImageVisual` that uses a texture atlas as David Hoese suggested on Gitter, but we are deferring that a bit until things are further along, since it's strictly an optimization.

octree_levels

pwinston:
I heavily updated the PR description to fully explain the changes being made here. And marked it ready for review. The actual octree functionality is not functional except for specific test cases I have. You cannot render normal images with it. But the wedge/feature-toggle aspect is solid and gives us quite a few new experimental files we can evolve.

napari/_qt/qt_viewer.py
```diff
@@ -152,8 +153,12 @@ def __init__(self, viewer):
         self.dockLayerList.setMaximumWidth(258)
         self.dockLayerList.setMinimumWidth(258)
 
+        # Only created if using perfmon.
         self.dockPerformance = self._create_performance_dock_widget()
 
+        # Only created if using async rendering.
+        self.dockRender = self._create_render_dock_widget()
+
```
sofroniewn, line 161: Can we make this off by default, something like

```suggestion
    if self.dockRender is not None:
        self.dockRender.setVisible(False)
```

Right now when I open my pathology example I just get an empty widget at the right. It's pretty easy to turn on with the shortcut

napari/_vispy/experimental/vispy_tiled_image_layer.py
```diff
@@ -0,0 +1,105 @@
...
+    rebuilding the shader alone.
+
+    Since we might need to draw 10+ new tiles at one time, the total delay
+    could be up to 200ms total, which is too slow.
+
+    However, this initial approach with separate ImageVisuals should be a
+    good starting point, since it will let us iron out the octree
+    implementation and the tile placement math. Then we can upgrade
+    to the new TiledImageVisual when its ready, and everything should
+    still look the same, it will just be faster.
```
sofroniewn, line 47: Good explanation, thanks!!

napari/_vispy/vispy_image_layer.py
```diff
@@ -66,14 +66,17 @@ def _data_astype(self, data, dtype):
...
             # Do nothing if we are not yet loaded. Calling astype below could
             # be very expensive. Lets not do it until our data has been loaded.
             return
 
-        data = self.layer._data_view
+        self._set_node_data(self.node, self.layer._data_view)
+
+    def _set_node_data(self, node, data):
+        """Our self.layer._data_view has been updated, update our node.
+        """
```
sofroniewn, line 78: I'm fine with this, is very clever for usage in tiled version

napari/layers/image/experimental/octree.py
```diff
@@ -0,0 +1,307 @@
...
+    num_rows = len(tiles)
+
+    num_cols = None
+    for row in tiles:
+        if num_cols is None:
+            num_cols = len(row)
+        else:
+            assert num_cols == len(row)
+
+    print(f"{num_rows} x {num_cols} = {num_rows * num_cols}")
```
sofroniewn, line 132: is there a reason in this file you're using all the print statements instead of the logging ones? Doesn't need to be fixed now, but maybe something for the future

sofroniewn (approved):
Ok, this looks good to me. It's still heavily in-progress in certain areas deep in the experimental portion of the code, but it does make a very nice wedge into the main code base in a clean way. The experimental portion like the octree and tiled visual can now be refined in a much easier way.

As already noted in the long run there maybe opportunities to reuse some dask functionality within the octree code, but we can address that later too.

I've left a couple of minor comments, but would only like to see one thing addressed before merge, which is making the render dock widget not visible by default when using `NAPARI_ASYNC`. You could @pwinston add an async config variable that would make it on by default if you want, but otherwise it's just a shortcut key combo away!

napari/layers/image/experimental/octree.py
pwinston: No real reason. For super temporary stuff I still sometimes use prints. But I migrate things to logging which I feel like we'll always want available. Always want to be able to toggle back on.

An example is like "I just loaded this data_id async" is probably something you always want to know, forever. But like while building the octree we might print a bunch of junk. But in the end maybe we just print the build time and the number of tiles or something minimal.

pwinston: As we discussed once, there should be no prints in mainstream code. That could be CI enforced someday. Although every project I've worked on people just shout "your stuff is printing junk!" and someone gets rid of it. Has not been a huge issue. Yet stray prints pop up fairly regularly.

napari/_vispy/vispy_image_layer.py
pwinston: This works well since we are letting`VispyImageLayer` do all the pre-processing it was always doing. The derived `VispyTiledImageLayer` is just calling into `VispyImageLayer._set_node_data()` once per tile. So yeah this was nice.

If we do switch to a `TiledImageVisual` as David suggested, I'm not as positive who would do this preprocessing. Maybe `VispyImageLayer` can continue to do it, and it just sends the processed data to the `TiledImageVisual` rather than making an `ImageNode` out of it, like we do here. That would also be nice.

napari/_qt/qt_viewer.py
pwinston: See latest commit, the render widget now only appears with octree visuals. It does nothing with non-octree visuals anyway so this is better.

napari/_vispy/vispy_image_layer.py
sofroniewn: > If we do switch to a TiledImageVisual as David suggested, I'm not as positive who would do this preprocessing. Maybe VispyImageLayer can continue to do it, and it just sends the processed data to the TiledImageVisual rather than making an ImageNode out of it, like we do here. That would also be nice.

Indeed I was wondering this too, but we can address later

napari/_qt/qt_viewer.py
sofroniewn: nice, thanks!

napari/layers/image/experimental/octree.py
sofroniewn: sounds good, we can address later. i like the idea of a CI test at some point

sofroniewn:
The requested change to dock widget visibility has been made. I'm merging this now

PR #1715, opened 2020-10-09
async-13: Fix "render to black" problem

pwinston:
# Description

No longer render black while loading the next slice. Continue to render the "last" slice until the new one has loaded. This fixes the "render to black" problem that has been bugging us for a long time. Fixes it for time-series and multi-scale.

Making the node invisible must have been necessary back when we had an "update loop" or for some reason a long time ago. It's no longer need. This works fine.

# Experimental?

The change is not in "experimental," but `layer.loaded` is always true for non-async layers, so the change should have no effect.

## Type of change
- [x] Fix visuals with async

# How has this been tested?
- [x] Manual testing

## Final checklist:

### DIFF
--- a/napari/_vispy/vispy_base_layer.py
+++ b/napari/_vispy/vispy_base_layer.py
@@ -58,7 +58,6 @@ class VispyBaseLayer(ABC):
         self.layer.events.blending.connect(self._on_blending_change)
         self.layer.events.scale.connect(self._on_scale_change)
         self.layer.events.translate.connect(self._on_translate_change)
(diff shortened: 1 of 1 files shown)


sofroniewn:
Nice fix! I've given this a play with and it definitely works better the flickering to black for the multiscale.

multiscale_no_flicker

But there are still issues when fetching large data. That should be addressable with the octree work though.

We'll have to think longer term about the behaviour when dealing with timeseries data or data that updates like in the napariboard example, but I think this is much better for now

sofroniewn (approved):

PR #1718, opened 2020-10-10
fix delete all shapes

### LINKED ISSUE #1674: Simpler way to erase all data in shapes layer
## 🚀 Feature
An [image.sc post](https://forum.image.sc/t/proper-way-to-erase-all-data-in-a-layer/40702) requested a simpler way to erase all data in a shapes layer, possibly via

```python
layer.data = []
```
or
```python
layer.data = np.empty((0, 2))
```

See that post for additional discussion and suggestions, which we can continue here.

sofroniewn:
# Description
This PR closes #1674, originally flagged on https://forum.image.sc/t/proper-way-to-erase-all-data-in-a-layer/40702
by making 

```python
viewer.layers['name'].data = []
# or
viewer.layers['name'].data = np.empty((0, 2))
```
Delete all shapes in a shapes layer. Two tests have been added to verify this change. Maybe @kevinyamauchi you can give a quick review 

## Type of change

### DIFF
--- a/napari/layers/shapes/_tests/test_shapes.py
+++ b/napari/layers/shapes/_tests/test_shapes.py
@@ -600,6 +600,28 @@ def test_selecting_shapes():
     assert layer.selected_data == set()
 
 
+def test_removing_all_shapes_empty_list():
+    """Test removing all shapes with an empty list."""
+    data = 20 * np.random.random((10, 4, 2))
+    np.random.seed(0)
+    layer = Shapes(data)
+    assert layer.nshapes == 10
+
+    layer.data = []
+    assert layer.nshapes == 0
+
+
+def test_removing_all_shapes_empty_array():
+    """Test removing all shapes with an empty list."""
+    data = 20 * np.random.random((10, 4, 2))
+    np.random.seed(0)
+    layer = Shapes(data)
+    assert layer.nshapes == 10
+
+    layer.data = np.empty((0, 2))
+    assert layer.nshapes == 0
+
+
 def test_removing_selected_shapes():
     """Test removing selected shapes."""
     np.random.seed(0)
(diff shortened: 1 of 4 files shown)


napari/layers/shapes/_tests/test_shapes.py
```diff
@@ -600,6 +600,28 @@ def test_selecting_shapes():
     assert layer.selected_data == set()
 
 
+def test_removing_all_shapes_empty_list():
+    """Test removing all shapes with an empty list."""
```
kevinyamauchi, line 604: @sofroniewn , I didn't have a chance to try this out yet, but I took a quick look. I think it would be good to also check that the colors, properties, etc. are also reset in these tests.

sofroniewn: good idea and thanks for the quick look!! If you'll have a chance to look in more detail in the future, maybe you can take over the PR and add that in too - there's no rush on this one. If not I can try too. thanks!!

kevinyamauchi: No problem. I can work on that this week.

sofroniewn:
I've updated this to include a fix for #1730. Please let us know @NHPatterson if this now works for you. I've added a test too to check for all the color setting too that failed on master and the old version of this PR.

@kevinyamauchi this is ready for review!! Thanks!!

napari/layers/shapes/_tests/test_shapes.py
sofroniewn: I've done this now and updated the PR to include the fix for #1730

NHPatterson:
> I've updated this to include a fix for #1730. Please let us know @NHPatterson if this now works for you. 

This fixed the issue for me. Thanks!

jni (approved):
@sofroniewn to me, fixing #1730 feels a bit magical, and not something we necessarily want to support (ie now `add_shapes` behaves like `add_shape_or_shapes`). Having said that I can't come up with an ambiguous case here, so I think we can pull the trigger on this for now and revert if it starts causing problems.

sofroniewn:
> @sofroniewn to me, fixing #1730 feels a bit magical, and not something we necessarily want to support (ie now add_shapes behaves like add_shape_or_shapes).

We've always been that way from a data standpoint, but agreed, maybe we should change that. Can revisit in the future though

Issue #1688, opened 2020-10-03
Show scale bar


sofroniewn:
## 🚀 Feature
Now that we're making good progress on our world coordinate system it is finally possible for us to add a scale bar in world coordinates. Rather like the proposed axes visual #1687, I'm imagining an additional visual and a way to toggle the scale bar on and off from the viewer.

PR #1719, opened 2020-10-11
add axes visual

### LINKED ISSUE #952: Add 3D vector coordinates object to the viewer
## 🚀 Feature
It would be nice if the viewer could also show the current direction of each axis via a simple 3D vector triplet which can be found in many 3D viewing applications, like Blender ([example](https://en.wikibooks.org/wiki/Blender_3D:_Noob_to_Pro/Coordinate_Spaces_in_Blender)).

## Motivation

Having the current axis directions spelled out in the viewer may help with the general orientation of the object and the viewer. In addition, if the axes are labeled, the name of each dimension will also be shown on the tips of these arrows, which can be quite helpful. Additionally, if and when napari will have support for global and relative coordinates, these arrows will also help determining the viewer's perspective as well.

## Alternatives

napari's awesome regardless of this little pointing rotating object :)

### LINKED ISSUE #1687: Show axes in 3D rendering mode.
## 🚀 Feature
In a couple of places it's come up that we want to show reference axes in 3D rendering mode. vispy has a little visual for this that we can use [XYZvisual](https://github.com/vispy/vispy/blob/master/vispy/visuals/xyz_axis.py#L7-L33) or we could make our own. You should be able to turn it on/ off from the viewer.

See https://forum.image.sc/t/show-axes-reference-in-the-3d-canvas/43578 and https://github.com/napari/napari/pull/1616#discussion_r495662377

sofroniewn:
# Description
This PR will close #1687 close #952 and address https://forum.image.sc/t/show-axes-reference-in-the-3d-canvas/43578 by adding an axes visual that shows the position of the axes in 2d and 3d

It's still a WIP, (toggling between 2d and 3d currently has a bug) but I'd love to get some input on what people want to see in more detail.

Right now the axes are tied to the world origin and have a fixed length in world space (hard coded to 100), this means that as you zoom in and out they change size instead of remaining fixed. It's probably worth making them remain fixed in length, say always 100 canvas pixels.

Another question is where the origin should be, should it be tied to the world coordinate 0, which it is now, or should it be tied to the center of the camera (which can move) or should it always just be in the bottom left corner say fixed in canvas space?

I'll be working on a fix for #1688 and a scale bar soon, which I see as somewhat related.

Here's a gif of where we are at now

axes_visible

Input welcome!! cc @jni @GenevieveBuckley @kevinyamauchi @bkcooper

## Type of change


sofroniewn:
> It's still a WIP, (toggling between 2d and 3d currently has a bug)

Just fixed this

sofroniewn:
Options for axes location are
- Fixed at world origin
- Fixed at camera origin
- Fixed in a corner of the canvas (say bottom left/ top right)

Options for axes lengths are
- Fixed in world coordinates
- Fixed in canvas coordinates

In theory all of these could be toggle-able / settable, and one day we could support all of them. I think I've seen examples of all of these in different 3D imaging software, so clearly they all can have pros/ cons, but it would be great to know how people are leaning at this stage!!

sofroniewn:
@Czaki not sure if this would be helpful to you too, but very curious what you think also

jni:
This is great! I like fixed at world origin and fixed as a small proportion (~5-10%?) of the world extent.

I'd also suggest using CYM as the colours, as blue is really low contrast on black.

Czaki:
> @Czaki not sure if this would be helpful to you too, but very curious what you think also

I like this idea. 

> Another question is where the origin should be, should it be tied to the world coordinate 0, which it is now, or should it be tied to the center of the camera (which can move) or should it always just be in the bottom left corner say fixed in canvas space?

I prefer tie to world coordinate 0 or corner. Axes going through data not looks like something witch produce clear image. 

> I'd also suggest using CYM as the colours, as blue is really low contrast on black.

I suggest to think about colorblind persons. There should be some way to control at least one of colors, thickness, length to allow person with problem with color distinguish to use this feature (and produce publication figure which satisfy coloring requirements in publication).

sofroniewn:
> This is great! I like fixed at world origin and fixed as a small proportion (~5-10%?) of the world extent.

Ok nice, when you say small proportion of world extent do you mean
 - like in the scale bar PR #1720 where they do change size with zoom, but it always stays ~100 canvas pixels
- or should it always just be 100 canvas pixels independent of the zoom level of the camera?

> I prefer tie to world coordinate 0 or corner. Axes going through data not looks like something witch produce clear image.

Ok seems like people like world coordinate 0 which is where it is now.

> > I'd also suggest using CYM as the colours, as blue is really low contrast on black.
>
> I suggest to think about colorblind persons. There should be some way to control at least one of colors, thickness, length to allow person with problem with color distinguish to use this feature (and produce publication figure which satisfy coloring requirements in publication).

I like both these ideas, we could also add text labels, or little arrows / dashes too?

Czaki:
> I like both these ideas, we could also add text labels, or little arrows / dashes too?

I think yes. Maybe instead of single position add submenu in view with options like:

* None
* Colored
* Dashed (one solid, one dotted, one dased)
* Text labels

sofroniewn:
Ok I've now update the scaling behavior so that the axes are always 100 canvas pixels long, irrespective of zoom and always located at the world coordinates origin using some code also present in #1720. I've also changed the coloring to CYM.

Do people like the zoom / positioning behaviour now?

axes_visible2

> I think yes. Maybe instead of single position add submenu in view with options like:
> 
> - None
> - Colored
> - Dashed (one solid, one dotted, one dashed)
> - Text labels

I like this idea, can give this a go next!! Thanks @Czaki!

sofroniewn:
Ok, I've now added support for dashed axes. I currently default them to be white, but I wonder if I should keep them CYM or make them black, or something more complicated, like the opposite of the canvas background (I can do that, it's not too hard).

Here's what it looks like.

axes_visible_dashed

I've also realized that the "magic scale factor" I use for 3D (both here and in #1720) actually depends on canvas size, so I need to do some more looking into that before we're ready to merge.

I'll also add to the menubar an option to turn this stuff on. I like the `None`, `Colored`, and `Dashed` choices right now and can add those. I'll probably wait on text for followups unless people really want

sofroniewn:
Ok, I've fixed things so color of dashed will update to be opposite of color of background canvas

I've also added a menubar with the `None`, `Colored`, and `Dashed` options, see here

axes_visible_dashed

The only thing not quite right is the scaling in 3D, but I think that is less important to fix in this PR, but very important to fix in #1720. I suggest we review and merge this one without worrying about that, and then get right in #1720. It will be much easier there.

This is now ready for review!! cc @jni @GenevieveBuckley @Czaki

napari/_qt/qt_main_window.py
```diff
@@ -336,6 +350,19 @@ def _add_help_menu(self):
...
         self.help_menu.addAction(about_key_bindings)
 
+    def _toggle_axes(self, action):
+        text = action.text()
+        if text == 'None':
+            self.qt_viewer.viewer.axes_visible = False
+        elif text == 'Colored':
+            self.qt_viewer.viewer.axes_visible = True
+            self.qt_viewer.viewer.axes_style = 'Colored'
+        elif text == 'Dashed':
```
GenevieveBuckley: Is dashed only in black and white?

* If possible, I'd prefer to keep the colours (unless that looks super horrible when you convert it to grayscale, like you might for a paper figure). Random thought: is it worth adding a grayscale keyword argument to our `nbscreenshot` method? If set to true, it could just run the scikit-image `rgb2grey` function on the output screenshot.
* I think one of the dashed lines (maybe the shorter dashes) should be made a little bit bold, if possible. As you rotate an object in 3D, the perspective on the dashed lines changes so it's sometimes hard to tell which dashed line is which.

napari/_qt/qt_main_window.py
```diff
@@ -336,6 +350,19 @@ def _add_help_menu(self):
         )
         self.help_menu.addAction(about_key_bindings)
 
+    def _toggle_axes(self, action):
```
GenevieveBuckley: I don't know where, but it should be documented *somewhere* what the default values are, for both colours and dashes. Eg: x-axis is green, etc. or whatever it really is.

napari/_qt/qt_main_window.py
```diff
@@ -336,6 +350,19 @@ def _add_help_menu(self):
...
+        text = action.text()
+        if text == 'None':
+            self.qt_viewer.viewer.axes_visible = False
+        elif text == 'Colored':
+            self.qt_viewer.viewer.axes_visible = True
+            self.qt_viewer.viewer.axes_style = 'Colored'
```
GenevieveBuckley: Is there a common convention about which color means which axis? To be honest, even if there's no common convention, we should probably consider matching what Imaris does. I imagine a reasonable chunk of our users might have previous experience with Imaris.

napari/_vispy/vispy_axes_visual.py
```diff
@@ -0,0 +1,116 @@
...
+    dashed_line = np.zeros((2 * num_dashes, 3))
+    dashed_line[:, axis] = np.array(dashed_line_ends)
+    return dashed_line
+
+
+class VispyAxesVisual:
+    """Axes indicating world coordinate origin and orientation.
+
+    Axes are colored x=cyan, y=yellow, z=magenta or dashed with
+    x=solid, y=dotted, z=dashed, depending on styling.
```
GenevieveBuckley: Ah, here's the docstring.

I think "depending on styling" is very vague, and doesn't give much of a clue about how to use it.

napari/_qt/qt_main_window.py
sofroniewn: I could make all of the options independent - so you could have 
 - visible or not visible
- colored or not colored
- dashed or not dashed

You'd be able to set them individually in the dropdown menu and get a checkmark as appropriate. Default would probably be `not visible`, `colored`, and `not dashed`. I kinda like this idea. Thoughts @GenevieveBuckley @Czaki ?

> Random thought: is it worth adding a grayscale keyword argument to our nbscreenshot method? If set to true, it could just run the scikit-image rgb2grey function on the output screenshot.

I like that idea, but we don't depend on skimage so would have to do the grayscale conversion ourselves, which isn't too hard.

> I think one of the dashed lines (maybe the shorter dashes) should be made a little bit bold, if possible. As you rotate an object in 3D, the perspective on the dashed lines changes so it's sometimes hard to tell which dashed line is which.

Yeah, i noticed that too, it's not idea, good idea though - I can try that!!

sofroniewn:
Following discussion here https://github.com/napari/napari/pull/1719#discussion_r503021234 and suggestion from @GenevieveBuckley I've now made it so that the axes can be both dashed and colored. I ended up adding a little `Axes` class that will be super simple once #1475 merges.

I didn't make the dashed lines different widths yet, let me know if that is still important.

I dropped that enum and have hopefully improved / centralized some of the doc strings.

Here's a gif of the drop down menu as it is now

axes_visible_toggle2

napari/_vispy/vispy_axes_visual.py
sofroniewn: this language has now been changed, agreed was vague

napari/_qt/qt_main_window.py
sofroniewn: I'm not aware, but if there is I'd happily update. I think I've seen RBG in autodesk products

napari/_qt/qt_main_window.py
sofroniewn: this has now been improved that the enum has been dropped

napari/_qt/qt_main_window.py
sofroniewn: ok, i've given it a go with making `visible`, `colored`, and `dashed` all independent bools, I think it's much nicer now, but curious what others think!!

Czaki:
Amazing.

How hard it will be to add arrows to inform about orientation?

sfcaracciolo:
Hi people, I hope you are ok!

Nice axes! I agree with everything I read in the comments, I think with toggle-able options most application cases will be covered. I have some thoughts/comments based on my normal usage (big bias, sorry!):

1. Length of axes: I think both fixed or variable size would be fine
2. I think, If axes are non-colored, then text labels are very important
3. Editable text label is another good feature, in medical imaging typical case is Right, Superior, Anterior, etc.
4. Corner axes location is a great idea, several times the only requirement is known the view rather than origin of the object. In this sense, a cube instead axes is good too.
5. Sorry for my lack of knowledge, but what the view would look like with axes at camera origin?

Cheers!

sofroniewn:
Thanks for weighing in @sfcaracciolo!

> - I think, If axes are non-colored, then text labels are very important
> - Editable text label is another good feature, in medical imaging typical case is Right, Superior, Anterior, etc.

We could do something here leveraging the axes labels stored in `dims.axes_labels`. I need to think about the best way to do this though, and maybe we end up making an issue and saving for a new PR, but it's a good idea.

> - Corner axes location is a great idea, several times the only requirement is known the view rather than origin of the object. In this sense, a cube instead axes is good too.
> - Sorry for my lack of knowledge, but what the view would look like with axes at camera origin?

I've actually gone ahead now and implemented a `centered` option which keeps the axes at the camera origin. It was a bit easier to implement than the corner and I don't think looks to bad. Maybe in the future if people prefer we can put it in the corner.

axes_visible_center

> How hard it will be to add arrows to inform about orientation?

I think I can do that. I'll give it a try.

Czaki:
> I've actually gone ahead now and implemented a `centered` option which keeps the axes at the camera origin. It was a bit easier to implement than the corner and I don't think looks to bad. Maybe in the future if people prefer we can put it in the corner.

I strongly prefer to have corner options. At least for publication screenshots. 
I'm worry about multiple images with coloring (RGB for example). I work with confocal microscopy data which are multichannel and need coloring to distinguish objects. I such data axes on camera center may looks bad. 

And if You use dashed without coloring on given example there could be problem with readability.

sofroniewn:
> I strongly prefer to have corner options. At least for publication screenshots.
I'm worry about multiple images with coloring (RGB for example). I work with confocal microscopy data which are multichannel and need coloring to distinguish objects. I such data axes on camera center may looks bad.

Ok I hear you, will give that a go! Should be doable.

> How hard it will be to add arrows to inform about orientation?

I've given this a go, see here now. Thoughts @Czaki?

axes_arrows

Seems to me like last thing to do is make the `centered` option move to the bottom corner of the screen instead and then maybe we can finalize this and make additional improvements like `text` in followup PRs.

GenevieveBuckley:
Wow, this is looking really nice!

I also agree that having the option to put the axes orientation marker in the bottom corner would be desirable for screenshots intended for publication.

Last question: I'm sure you've checked already, but the behavior when you "roll" through the 2D axes orientations (the button next to the 2D -> 3D one on the GUI) is what we'd expect to see, right?

sofroniewn:
> Last question: I'm sure you've checked already, but the behavior when you "roll" through the 2D axes orientations (the button next to the 2D -> 3D one on the GUI) is what we'd expect to see, right?

It did not, but it does now! Thanks for the suggestion. If you have a 4D dataset it will cycle through the same 3 colors for the all axes, we could add a 4th color in the future, but I'd rather not for this PR, it's already grown quite a bit.

axes_order

> I also agree that having the option to put the axes orientation marker in the bottom corner would be desirable for screenshots intended for publication.

Absolutely agreed, this has actually proven a little more challenging, but will be much easier to do once we have a proper camera model with #854. I've removed the `centered` mode which no one really liked and we can then add this later.

I think functionality wise this is now good enough for a first pass and in the interest of keeping this PR small and reviewable we should probably save further enhancements like text, bottom right corner, and more colors for more axes for follow up PRs.

I still need to add some tests for this PR, but I think it is now ready for another round of code review.

sofroniewn:
I've now added a couple of minimal tests - just instantiating the axes object, making the axes visible/ not visible, and a screenshot test to determine that you can indeed see something when axes turn on. I think this is sufficient for now, will catch major regressions in functionality, we can then work on more comprehensive tests at a future date I think, unless the reviewers have additional suggestions

napari/_vispy/vispy_axes_visual.py
```diff
@@ -0,0 +1,183 @@
...
+    def _on_data_change(self, event):
+        """Change style of axes."""
+        if self.axes.dashed:
+            data = self._dashed_data
+        else:
+            data = self._default_data
+
+        if not self.axes.colored:
+            color = np.subtract(1, self.axes.background_color)[:3]
+            arrow_color = [color] * 300
```
jni: What is "300" here? I guess a bunch of constants in this module could use either `DEFINING =` or `# explaining`.

napari/_vispy/vispy_axes_visual.py
```diff
@@ -0,0 +1,183 @@
...
+        self.axes.events.colored.connect(self._on_data_change)
+        self.axes.events.dashed.connect(self._on_data_change)
+        self.axes.events.arrows.connect(self._on_data_change)
+        self.dims.events.order.connect(self._on_data_change)
+
+        self._on_visible_change(None)
+        self._on_data_change(None)
+
+        self._scale = 0.1
+        self.update_scale(1)
```
jni: Another set of constants that should probably be documented.

napari/_vispy/vispy_axes_visual.py
```diff
@@ -0,0 +1,183 @@
...
+            v, f = make_arrow_head(100, axis)
+            faces = np.concatenate([faces, f + len(vertices)], axis=0)
+            vertices = np.concatenate([vertices, v], axis=0)
+        self._default_arrow_vertices = vertices
+        self._default_arrow_faces = faces.astype(int)
+        self._default_arrow_color = np.concatenate(
+            [
+                [[1, 0, 1, 1]] * 100,
+                [[1, 1, 0, 1]] * 100,
+                [[0, 1, 1, 1]] * 100,
```
jni: Is there a reason why the order is MYC? Given that the axis order is currently not well documented (no text labels), I wonder whether CMY (as in CMYK) is the more intuitive ordering.

napari/_vispy/vispy_axes_visual.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+def make_dashed_line(num_dashes, axis):
+    """Make a dashed line.
+
+    Parameters
+    ----------
+    num_dashes : int
+        Number of dashes in the line.
+    axis :
```
jni:
```suggestion
    axis : int
```

napari/_vispy/vispy_axes_visual.py
```diff
@@ -0,0 +1,183 @@
...
+        Axis which is dashed.
+
+    Returns
+    -------
+    np.ndarray
+        Dashed line.
```
jni: What is the dimensionality of the array? I don't know how a dashed line is represented. Is it a bunch of line segments? Is it a 3D array of start/stop coordinates? Along how many dimensions?

napari/_vispy/vispy_axes_visual.py
```diff
@@ -0,0 +1,183 @@
...
+            arrow_faces = self._default_arrow_faces
+        else:
+            arrow_vertices = np.zeros((3, 3))
+            arrow_faces = np.array([[0, 1, 2]])
+            arrow_color = [[0, 0, 0, 0]]
+
+        order = tuple(self.dims.order[-3:])
+        if len(order) == 2:
+            order = (0,) + tuple(np.add(order, 1))
+        order = tuple([i % 3 for i in order[::-1]])
```
jni, line 168: I'm really confused by this expression, could you add a comment explaining what it's doing?

jni (approved):
@sofroniewn very nice, I've just left some minor comments where I think the code could use clarification. No big deal, would be good to address some of those before merging.

One thing I can finally notice now that you've added this feature is that we are displaying 3D with a left-handed reference frame, rather than right-handed. (I think! This always does my head in. 😂) We should now very carefully check whether this is a problem by e.g. loading 3D medical data, e.g. a dataset of a brain tumour on the left side of the brain, and check that it does indeed appear on the left side of the brain. Or a 3D point model of DNA to see that it is right-handed. If not, then we should invert the direction of our z/planes axis.

napari/_vispy/vispy_axes_visual.py
sofroniewn: constant defined

napari/_vispy/vispy_axes_visual.py
sofroniewn: fixed

napari/_vispy/vispy_axes_visual.py
sofroniewn: reordered

napari/_vispy/vispy_axes_visual.py
sofroniewn: comments added

napari/_vispy/vispy_axes_visual.py
sofroniewn: explanation added

sofroniewn:
I've made the additional clarifications and minor updates requested by @jni. I've also made a new issue #1726 to capture some of the additional enhancements that people have requested. Thanks all for the input and suggestions here. It will be great to get this first pass in, but I'm sure we'll be making improvements to it quickly. I will merge once all tests pass

Issue #1701, opened 2020-10-06
Allow scale and translation parameters to have units


jni:
## 🚀 Feature

Some popular unit frameworks include [pint](https://pint.readthedocs.io/en/stable/), [astropy.units](https://docs.astropy.org/en/stable/units/), and [unyt](https://unyt.readthedocs.io/en/stable/). For scientists I think it would be great to write something like `scale=np.array([1, 0.2, 0.2]) * u['µm']` (pint syntax) and have the rest of the napari UI take the units into account, e.g. when displaying coordinates, measuring distances, or displaying scale bars.

kevinyamauchi:
May be of interest:[ `pint` integration with `xarray`](https://github.com/xarray-contrib/pint-xarray)

sofroniewn:
Picking up in the discussion of units happening in https://github.com/napari/napari/pull/1720#issuecomment-706778646 and https://github.com/napari/napari/pull/1720#issuecomment-706779755

Do you @GenevieveBuckley and @Czaki have a preferred library or way of dealing with units in Python or experience with the packages that @jni suggests?

If we get a PR adding unit support in first we can then extend that to #1720 to make sure the scale bar works properly

Czaki:
I use custom implementation of units. But presented libraries looks great and I think that I should switch to one of them.

GenevieveBuckley:
I don't have a lot of direct experience with this, but if pint can integrate with xarray I think that's a point in it's favour. 

I would like to see stronger xarray support here too, we might find that we get some of what we need here for free, or in a much easier way.

sofroniewn:
> I don't have a lot of direct experience with this, but if pint can integrate with xarray I think that's a point in it's favour.

Agreed! Maybe @Czaki you can investigate switching over your work to pint and report back on if that looks like a good idea for us to adopt here.

@GenevieveBuckley let us know if you look in pint more too.

It seems relatively widely used at this point > 2000 projects, which is nice!!

jni:
I've used pint before in small projects and it's great, others were included for completeness. Unyt is the new hotness and claims to compute things up to 2x as fast as pint (if I remember correctly), but I think for the uses we are likely to have in napari this is not a huge concern. It's also a much heavier dependency as it pulls down SymPy. So I'm pro-pint.

Czaki:
I start to do a review of `unyt` and `pint`. 
The big disadvantage of `pint` is a lack of the global register of units:

> Warning

[...]

PR #1720, opened 2020-10-11
Add scale bar

### LINKED ISSUE #1688: Show scale bar

sofroniewn:
# Description
This PR will close #1688 by adding a scale bar that works in both 2D and 3D, and with multiscale. It rounds to the nearest 1, 2, 5 or 10 on a log scale which is the same behavior as google maps and I think it is quite nice.

It is off by default, but you can turn on with `viewer.scale_bar_visible = True`. If is white by default, but the color can be changed with `viewer.window.qt_viewer.scale_bar.color= 'black'` We can decide how to better expose that, or to do something fancy like detect the background luminance and toggle based on that in future PRs. It will be good just to get the basics in.

Once the camera model PR #854 or it's equivalent merges this will clean up a bit too, but then it will be very clear what the camera model is doing, especially after we get #1719 in too!!

Here's a gif in action:

scale_bar_visible2

Feedback very welcome!! cc @jni @kevinyamauchi @GenevieveBuckley @DragaDoncila 

## Type of change


sofroniewn:
@Czaki I know you do a lot with "physical scales" too would be great to get your feedback on this one also!!

jni:
Small suggestions:

- remove the ticks at the ends of the bar
- move it to the bottom right (it's where I see most scale bars?)
- (goes for #1719 also) add a View menu and put visibility toggles there

The functionality looks so nice and smooth. Will be amazing once we have units support!

sofroniewn:
> Small suggestions:
>
> - remove the ticks at the ends of the bar

I really like the ticks! I can remove them if consensus is against them, I agree it's a little less common to find them in scientific visualization, but they are there in things like google maps, and I wonder if that is just because it's harder to add them. I found them very useful when trying to measure the actual number of pixels inside the line - but maybe people won't really be doing that. As I said though, happy to remove / make toggle-able but want a third opinion.

> - move it to the bottom right (it's where I see most scale bars?)

Yeah, it's a tiny bit trickier to put there because I need to know the size of the canvas to do that, which can change when the viewer resizes etc. It can certainly be done, and am also happy to put there now or later if people want, but that was why I didn't at first.

> - (goes for #1719 also) add a View menu and put visibility toggles there

Sounds like a good idea, I'll do that. We can add keyboard short cut either now or later, depending on how certain people feel about that!!

GenevieveBuckley:
Nice!

There should be units attached to that scale bar. I know that we don't require users to pass in what the units are (you can just say `scale=[1,1,5]` and not whether those are pixels, microns, or parsecs, etc.). Maybe if the user has provided unit information we also display that, otherwise leave it out. We should avoid having a plausible-ish default that wasn't actually specified by the file metadata or user (from memory ImageJ uses 1 micron as the default scale, which is just plausible enough to be misleading if you're not paying enough attention).

Other questions:
* Does the scale bar always sit on top of all of the other images/layers? I don't want it to ever get accidentally hidden underneath something else.
* I seem to remember there's a way to change the napari theme to "light mode". If that's true, will the scale bar automatically change itself to black then?

sofroniewn:
> There should be units attached to that scale bar. I know that we don't require users to pass in what the units are (you can just say scale=[1,1,5] and not whether those are pixels, microns, or parsecs, etc.). Maybe if the user has provided unit information we also display that, otherwise leave it out. We should avoid having a plausible-ish default that wasn't actually specified by the file metadata or user (from memory ImageJ uses 1 micron as the default scale, which is just plausible enough to be misleading if you're not paying enough attention).

Agreed!! There's some discussion of this in #1701. Maybe we figure out that first and then we can add the units in a follow-up PR, for now it's just in the unit of pixels in a world coordinate - which means it only cares about camera zoom. It doesn't got all the way back to the data coordinates, and so anything in the `data2world` transform has no effect on it. i.e. doing `viewer.layers[0].scale = [2, 2]` doesn't change the scalebar right now!!! Maybe this is counterintuitive and needs to be changed, but that means it's hard to think about a global world that applies to all layers equally. Very curious what people want to see here, but we might need to add an additional `scale` factor to the Viewer that is global and handles this sort of units conversion @GenevieveBuckley @jni 

> Does the scale bar always sit on top of all of the other images/layers? I don't want it to ever get accidentally hidden underneath something else.

Yes it should, i hard code an "order" value of `1e6` which should be high enough

> I seem to remember there's a way to change the napari theme to "light mode". If that's true, will the scale bar automatically change itself to black then?

Not yet, but agreed we should add that!

Czaki:
Really nice. 

As there is no units information in current layers model there should be at least option to set units string like `viewer.scale_bar_units = "nm"` 

The best on there should be units information on layer and select of scale bar units. Something like: 

```
viewer.set_image(data, scale=[1, 210, 70, 70], units=Units.nm)
viewer.scaler.units = Units.um
```

(with units there is problem with time axes)

I like ticks. Someone could remove it using some graphics editor.

Czaki:
> > I seem to remember there's a way to change the napari theme to "light mode". If that's true, will the scale bar automatically change itself to black then?
> 
> Not yet, but agreed we should add that!

Maybe use opposite color to canvas background? I overwrite background color for light mode (because it works better with confocal data). 

But when think about color we should also remember about electron microscopy data.

GenevieveBuckley:
I also agree that the positioning should be at the bottom right, for consistency with user expectations.

GenevieveBuckley:
Also, I like the ticks too. I think @jni is outnumbered!

kevinyamauchi:
Super cool! 

Do you think it's worth making a small dataclass for the scalebar? From the discussion here and thinking about future usage, it seems like there are going to be options to customize the scale bar (e.g, position, units, visibility). It would be great to have those represented in a model. Also, I think it would clean up the API a bit. For example:

```python

# as is
viewer.scale_bar_visible = True
viewer.scale_bar_units = 'nm'
viewer.scale_bar_position = 'lower_right'

# with data class
viewer.scale_bar.visible = True
viewer.scale_bar.units = 'nm'
viewer.scale_bar.position = 'lower_right'
viewer.scale_bar.tick_mode = 'juan'
```
A `ScaleBar` data class would also make it easy to add some flexibility to setting up the scale bar in the `napari.view_*` methods

```python
# simple scale bar with defaults
viewer.add_image(im, scalebar=True)

# more complex setup
scale_bar_settings = {
    'visible': True,
    'units': 'nm',
    'position': 'lower_right'
}
viewer = napari.add_image(im, scalebar=scale_bar_settings)

```

kevinyamauchi:
>> Does the scale bar always sit on top of all of the other images/layers? I don't want it to ever get accidentally hidden underneath something else.

> Yes it should, i hard code an "order" value of 1e6 which should be high enough

Famous last words, @sofroniewn  :stuck_out_tongue:

jni:
> `viewer.scale_bar.tick_mode = 'juan'`

I LOLed and LOLed but now I totally want this as an easter-egg. :joy:

> `viewer = napari.add_image(im, scalebar=scale_bar_settings)`

Note that scale bar is a viewer/canvas property, not a layer property, so it should go in the Viewer constructor or `view_` functions, not in the `add_` methods.

sofroniewn:
> Do you think it's worth making a small dataclass for the scalebar? From the discussion here and thinking about future usage, it seems like there are going to be options to customize the scale bar (e.g, position, units, visibility). It would be great to have those represented in a model. Also, I think it would clean up the API a bit. 

Yes, I love this idea! I actually started moving #1719 over in that direction too and it's much nicer. Will do that in the next update of this PR

sofroniewn:
Quick update. This is coming along nicely, I've made a new `ScaleBar` class, with options like `ticks`, default `True` ;-) that are editable from a menubar, like in #1719

I'm going to add a `position` which could be one of `top_left`, `top_right`, `bottom_left`, `bottom_right` and add some tests and then this will be ready for review. I think we're going to save the addition of units until after progress has been made on the discussion of units in #1701

sofroniewn:
This is getting very close now. I've added the ability to set the position of the scale bar using the `scale_bar.position` property to all four corners, see gif below. One question though is should that be settable through another nested menubar dialog? We can always wait and merge without or I can just try and quickly add now and then be done with it. I'm just worried that might be too much nesting! Alternatively we could have a bool toggle between two spots - i.e. `Left` / `Right` and always keep it on the bottom? That might be nice. Thoughts @jni @GenevieveBuckley @Czaki ?

I think that is the last thing to decide. I will add a couple test now and then this will be ready for review

scale_bar_position

jni (approved):
@sofroniewn I think let's punt on the menubar option for positioning. I think `Position > (4 options)` is much more intuitive than `Left` with a tick mark. But bottom right is just sensible as a default and we can add the menubar option later, get this in now! Still lots to fiddle with e.g. units.

sofroniewn:
> @sofroniewn I think let's punt on the menubar option for positioning. I think Position > (4 options) is much more intuitive than Left with a tick mark. But bottom right is just sensible as a default and we can add the menubar option later, get this in now! Still lots to fiddle with e.g. units.

Ok, sounds great @jni thanks!! I've added a couple tests, this is now ready for review @jni @GenevieveBuckley @Czaki

GenevieveBuckley:
> 
> 
> @sofroniewn I think let's punt on the menubar option for positioning. I think `Position > (4 options)` is much more intuitive than `Left` with a tick mark. But bottom right is just sensible as a default and we can add the menubar option later, get this in now! Still lots to fiddle with e.g. units.

I agree with @jni here too, bottom right by default is fine.

Some of the discussions here and in the axes PR make me a little worried we might be starting to head down the path of just adding options upon options upon options whenever we're not sure about making a good choice for the defaults. I'd like us to try and be mindful of not falling into this particular trap.

GenevieveBuckley:
Also, I just want to say that I really appreciate all the gifs in the conversation thread @sofroniewn 
It's really helpful to see them.

sofroniewn:
> Some of the discussions here and in the axes PR make me a little worried we might be starting to head down the path of just adding options upon options upon options whenever we're not sure about making a good choice for the defaults. I'd like us to try and be mindful of not falling into this particular trap.

Absolutely yeah, good point. We can roll back the tick option on the menubar too 🤣 I definitely want to be mindful of too many options that could overwhelm a user!

GenevieveBuckley:
I don't have any further comments, this looks good to me @sofroniewn

napari/components/scale_bar.py
```diff
@@ -0,0 +1,108 @@
...
+        """bool: If scale bar is colored or not."""
+        return self._colored
+
+    @colored.setter
+    def colored(self, colored):
+        self._colored = colored
+        self.events.colored()
+
+    @property
+    def background_color(self):
```
kevinyamauchi, line 71: I think that, ideally, we shouldn't store the background color of the canvas in `ScaleBar`.  That is, to avoid two sources of "truth", the background color shouldn't be duplicated here (the background color also lives in `ViewerModel.palette`)". Perhaps it would be better to connect the color update function to the ViewerModel `palette` event.

That being said, I see that we are using this pattern elsewhere (e.g., `Axes`), so I think we should proceed this way for now.

kevinyamauchi (approved):
Thanks, @sofroniewn ! This looks good to me. I have just left a comment below that I think we can address in the future.

napari/components/scale_bar.py
Czaki: But it is updated in palette event of viewer.

Czaki:
Looks great.

napari/components/scale_bar.py
kevinyamauchi: > But it is updated in palette event of viewer.

Hey @Czaki ! It is true that the value of `ScaleBar.background_color` is updated by the `ViewerModel`. My concern here is that we are storing the value of the viewer background color in multiple places, which increases the amount of bidirectional synchronization we need to/should do. I say bidirectional synchronization because I think it would be reasonable for a user who reads the docs for `ScaleBar.background_color` (a public property) to expect to be able to change the viewer canvas background color via `ScaleBar.background_color` (thus the need for bidirectional sync). For just a couple of properties, this isn't such a big deal, but as napari grows in complexity, managing all of these events and updates becomes burdensome.

A possible alternative would be to use a function that takes the new canvas background color as an input and sets the scale bar color accordingly without storing the new background color.

All of that being said, as I mentioned above, I think we should move forward as is, since it both works and we have used that pattern elsewhere (e.g., `Axes`). I just wanted to make a note here in case we want to revisit in the future.

sofroniewn: Yeah, agreed @kevinyamauchi I didn't love that, something we can think about fixing for sure

PR #1721, opened 2020-10-12
Add a welcome visual

### LINKED ISSUE #1132: Minor improvement: Guide the user on how to open images
## 🚀 Feature
Improve the user experience for new users when there's an empty canvas. Provide instructions on how to open images or even a quick message for how to get started.

## Motivation

It's not totally clear that I can drag images to the canvas to open them. A clear message would help the user to get started quicker.

## Pitch

I recommend having a message on the canvas when I first open napari with an empty canvas that says something like “Drag image(s) here to open or use the menu: File > Open image(s)”
empty_canvas_message

sofroniewn:
# Description
This PR closes #1132 originally raised by @justinelarsen. It most closely follows the design from https://github.com/napari/napari/issues/1132#issuecomment-613795478 but without the keyboard shortcuts. We could definitely add those in later.

I decided to use a vispy visual for this as it was relatively easy to do so, but we could also use a qt visual if we preferred.

Here is what it looks like right now

Screen Shot 2020-10-11 at 5 17 34 PM

Comments and feedback on visuals welcome!

## Type of change


jni:
Good stuff! :tada: I think drag-and-drop is less common when launching napari from a script, so I would propose one of the following tweaks:

- add another option, "or call `viewer.add_*`", or
- don't show the welcome visual unless launching from command-line or bundled app, or
- (fanciest) detect the launch origin and show *either* `viewer.add_*` or the current visual.

This is all somewhat optional — it's not massively disruptive to have this visual there when launching from Python... But it just feels a bit weird.

sofroniewn:
hmm, yeah good points @jni. we can think this one over for a little. agreed bundled app/ cli is probably where the most critical need is. right now the visual will also reappear when the layer list returns to having no layers, which I sort of like, but I can understand if people found strange. curious what others think

jni:
Reappearing is good I think. =)

justinelarsen:
@sofroniewn Looks good! I like that it appears when the layer list doesn't have any layers.

napari/_vispy/vispy_welcome_visual.py
```diff
@@ -0,0 +1,68 @@
...
+            4 / 3 * center[1],
+            0,
+            0,
+        ]
+        self.text_node.font_size = 25
+        self.text_node.anchors = ('center', 'center')
+        self.text_node.text = (
+            'Drag file(s) here to open \n or \n use File -> Open File(s)'
+        )
+        self.text_node.color = [0.7, 0.7, 0.7, 1]
```
justinelarsen: Slight transparency of the text and logo may look good.

napari/_vispy/vispy_welcome_visual.py
```diff
@@ -0,0 +1,68 @@
...
+        self.text_node.transform.translate = [
+            center[0],
+            4 / 3 * center[1],
+            0,
+            0,
+        ]
+        self.text_node.font_size = 25
+        self.text_node.anchors = ('center', 'center')
+        self.text_node.text = (
+            'Drag file(s) here to open \n or \n use File -> Open File(s)'
```
alisterburt: Silly details but I really like the unicode arrows in these situations
'Drag file(s) here to open \n or \n use File → Open File(s)'

sofroniewn: Unfortunately the unicode character written as `'use File \u2192 Open File(s)'` doesn't render right in vispy, but I love the suggestion. 
image

napari/_vispy/vispy_welcome_visual.py
sofroniewn: Yes, I quite like it and in general I've found out recently that one should use opacity instead of grayscale on text in general for accessibility reasons - see https://material.io/design/color/text-legibility.html#text-backgrounds for an interesting read

napari/_vispy/vispy_welcome_visual.py
sofroniewn: ok for baffling reasons it looks like I can create a left arrow, but not right arrow 
image
something seems off with that one too though, I'll keep investigating

sofroniewn:
Ok, this welcome visual is now ready for review. It's fairly crude - I couldn't get the unicode arrow work for the menu option (see discussion above).

Right now this message will appear whenever the layer list goes to 0, regardless of where the napari was launched from, and there is no way to enable/ disable. I could add a `disabled` flag if developers like @Czaki might not want this message (curious if you'd prefer that) or try and make the message a little more customizable, but maybe that's for the future.

I think it would be nice to get this in for the 0.4.0 release, so I consider this ready for re-review @jni. Curious if @liaprins-czi or @LCObus have any recommendation too before merge!!

Here's the current state (note the sizings will scale as the canvas resizes)

Screen Shot 2020-10-23 at 8 10 13 PM

sofroniewn:
Note also that we can and should continue to iterate and improve this dialog as things go even after merge!! I consider this only a "better than nothing" start

napari/_vispy/vispy_welcome_visual.py
sofroniewn: unfortunately I couldn't figure this out easily, looks like something is off with some clipping, so I've just used a `:`, see screenshot here https://github.com/napari/napari/pull/1721#issuecomment-715664646

jni (approved):
@sofroniewn I don't want to hold up release for this, but did you have issues detecting the launch type? Personally I just don't love the visual for IPython work, it feels weird and somewhat condescending. Like Clippy. 😂

On the other hand this is very minor and doesn't affect any actual use cases in IPython, so happy to just merge this for now and see how people like it, maybe I'm way off here!

jni:
Ha, @sofroniewn, looks like your scale bar and axes tests are failing now because the screenshot is no longer black. 🤦 😂 

https://cirrus-ci.com/task/4946334576803840?command=test_pyside2_with_coverage#L1443

Czaki:
Could You rethink and place `VispyWelcomeVisual` initialization outside `QtViewer` init? Application embedding napari as the viewer may export another interface to load files. 

(My application is not affected because I have a placeholder image on startup)

sofroniewn:
> Ha, @sofroniewn, looks like your scale bar and axes tests are failing now because the screenshot is no longer black. 🤦 😂

Lol, ok something will need to be reconsidered!

> Could You rethink and place VispyWelcomeVisual initialization outside QtViewer init? Application embedding napari as the viewer may export another interface to load files.

I'll look into this or at least providing an arg to QtViewer to disable if people prefer

> Personally I just don't love the visual for IPython work, it feels weird and somewhat condescending. Like Clippy. 😂

Ok yeah, don't want to give off that vibe!

If I look at what VSCode does it's more minimal, doesn't have the `!` which was maybe a little much. We could emulate this a bit more

Screen Shot 2020-10-24 at 3 25 42 PM

We can release `0.4.0` without this too, it's not blocking either way, so we can just see where we land as we get close

alisterburt:
@sofroniewn I love the minimal vscode splash - what about something along these lines for napari? personally I like the one on the far right the most

splash

Github won't inlin -upload the svg but I'm happy to send to anyone who wants to play with colours etc

sofroniewn:
ooo @alisterburt I like what you've done with the last one on the far right - very minimal, will hopefully be less obtrusive. Sending over the svg would be great. If GitHub doesn't allow that maybe you can made a new thread in the Zulip and paste it there, thanks!!

alisterburt:
@sofroniewn started the thread on Zulip!

sofroniewn:
Ok, I've update this to do a little image manipulation on the logo to make it look more minimal, inspired by ideas from @alisterburt. The approach now also adjusts the coloring based on our theme, so for dark and light respectively it appears as 

Screen Shot 2020-10-25 at 1 25 08 PM

and 

Screen Shot 2020-10-25 at 1 25 14 PM

I've also toned down the language a little, so I think it is more subtle now @jni. My preference would be to get this to a point where it is so minimal and unobtrusive that you don't really care about it if you're an everyday napari user from the terminal. I think this is moving in that direction, but as a last resort we can still make it display depending on launch method.

I still need to fix the tests - will be easy I'll look for diff instead of max change.

I'll also add a way to disable this for devs who don't want it there.

Further thoughts?

sofroniewn:
Here's a new version from @alisterburt 

Screen Shot 2020-10-25 at 3 30 59 PM

I like no capitalization, but might go back to the old colors above. curious what others think

sofroniewn:
Here's one more hybrid version:

Screen Shot 2020-10-25 at 3 36 38 PM

Definitely something very easy to keep on changing after merge. I'm pretty happy with any of these to be honest, and am curious what @jni and others like best!!

jni:
Love where this has gone! Now that it's so much more subtle, I'm happy having this as default. @sofroniewn have you had a chance to address @Czaki's objection? (ie make this a parameter of the QtViewer?)

A slight tweak might be to use a thinner font. But no big deal. I'm happy to have this merged as it stands.

sofroniewn:
>  @sofroniewn have you had a chance to address @Czaki's objection? (ie make this a parameter of the QtViewer?)

yes - you can do `QtViewer(viewer, welcome=False)` and it will never be there

sofroniewn:
Ok, I'm going to merge now. We can continue to update in the future, but we're now only a couple PRs away from `0.4.0`, and there's a good chance if this merges now I can make the first release candidate tonight. Continued feedback welcome though!! I really see this as our first attempt here and making space from an implementation point of view rather than from a design / content point of view, which can come later!!

LCObus:
Love seeing the iterations over the course of the weekend and think this is a great first implementation! Definitely second the more subtle color palette that matches the panels overall, and love the activation energy here for first time users. Not a blocker to release, but I think a more direct CTA would be helpful, something like
"To add image(s) 
1. drag and drop file(s) here
2. click the file menu dropdown _or_
3. call a viewer.add_* method"

Playing with a way to let folks know the text in context (multiple methods to achieve the same thing) and ideally, the copy should line up with how it is phrased in getting started tutorials.

liaprins-czi:
I like the subtle low contrast but I think it's too low contrast to be accessible, so we can work on figuring out a good balance!

Also agree with Lucy about separating the three methods as bullets, and matching the instructions (or at least order) with how we'll want the docs to list these methods (its okay if we are much more succinct in phrasing here. I would not number the bullets because then it seems more like a step-by-step instruction list, but instead left-align the text and use regular dot-bullets • Also the text "File : Open" in the original was not too clear, but could say something like Lucy suggests or "Select 'File' from the menubar, then 'Open'".

PR #1725, opened 2020-10-14
fix initialization of tracks color_by

### LINKED ISSUE #1722: Tracks color_by not handled correctly
## 🐛 Bug

Two issues with `color_by`:

1. When creating a tracks layer with multiple properties and specifying a specific property name using `color_by`, the property is not selected for coloring (it always seems to default to the same property). Color_by always seems to default to the first dictionary key. image
Looking at the Track `__init__` method I don't really see why that happens, as the property seems to be initialized [here](https://github.com/napari/napari/blob/5fff516d237a8786572474a36b41024dfb3267cc/napari/layers/tracks/tracks.py#L136) and [here](https://github.com/napari/napari/blob/5fff516d237a8786572474a36b41024dfb3267cc/napari/layers/tracks/tracks.py#L154).

1. The second issue is a silent failure when specifying a property name that does not exist as a dictionary key: https://github.com/napari/napari/blob/5fff516d237a8786572474a36b41024dfb3267cc/napari/layers/tracks/tracks.py#L426  I think the user should be notified. Not sure about the best method to do this but maybe just throw a `ValueError` ?
   

## To Reproduce

1. Create a track layer with a properties dictionary that contains multiple keys. 
2. Try different keys for `color_by`.

Apologies that I'm only showing a screenshot above. Need to generate some fake tracking data to make
a share-able code snippet.

## Expected behavior

The track should be colored by the property corresponding to the specified `color_by` key.

## Environment

```
napari: 0.3.9.dev40+gd3511ee
Platform: Linux-5.4.0-48-generic-x86_64-with-debian-bullseye-sid
Qt: 5.14.2
PyQt5: 5.14.2
NumPy: 1.19.1
SciPy: 1.5.2
Dask: 2.25.0
VisPy: 0.6.4

MAX_TEXTURE_SIZE: 32768

Plugins:
- napari_plugin_engine: 0.1.7
- svg: 0.1.4
```

 

## Additional context

@quantumjot

kevinyamauchi:
# Description
This PR fixes the handling of the `color_by` argument when initializing a `Tracks` layer (see #1722)

## Type of change

# References
Closes #1722 

# How has this been tested?
- [x] added a test for the qt controls that fails on master and passes with this patch

## Final checklist:

### DIFF
--- a/napari/_qt/layer_controls/qt_tracks_controls.py
+++ b/napari/_qt/layer_controls/qt_tracks_controls.py
@@ -42,7 +42,7 @@ class QtTracksControls(QtLayerControls):
         self.color_by_combobox.addItems(self.layer.properties_to_color_by)
 
         self.colormap_combobox = QComboBox()
-        self.colormap_combobox.addItems(AVAILABLE_COLORMAPS.keys())
+        self.colormap_combobox.addItems(list(AVAILABLE_COLORMAPS.keys()))
 
         # slider for track tail length
         self.tail_length_slider = QSlider(Qt.Horizontal)
--- a/napari/_qt/layer_controls/qt_tracks_controls.py
+++ b/napari/_qt/layer_controls/qt_tracks_controls.py
@@ -40,6 +40,7 @@ class QtTracksControls(QtLayerControls):
         # keys
         self.color_by_combobox = QComboBox()
         self.color_by_combobox.addItems(self.layer.properties_to_color_by)
+
         self.colormap_combobox = QComboBox()
         self.colormap_combobox.addItems(AVAILABLE_COLORMAPS.keys())
 
@@ -99,7 +100,6 @@ class QtTracksControls(QtLayerControls):
 
         self._on_tail_length_change()
         self._on_tail_width_change()
-        self._on_properties_change()
         self._on_colormap_change()
         self._on_color_by_change()
 
--- a/napari/_qt/layer_controls/_tests/test_qt_tracks_layer.py
+++ b/napari/_qt/layer_controls/_tests/test_qt_tracks_layer.py
@@ -10,23 +10,26 @@ _PROPERTIES = {'speed': [50, 30], 'time': [0, 1]}
 
 def test_tracks_controls_color_by(qtbot):
     """Check updating of the color_by combobox."""
-    layer = Tracks(_TRACKS, properties=_PROPERTIES, color_by='speed')
+    inital_color_by = 'time'
+    layer = Tracks(_TRACKS, properties=_PROPERTIES, color_by=inital_color_by)
     qtctrl = QtTracksControls(layer)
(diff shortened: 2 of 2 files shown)


napari/_qt/layer_controls/qt_tracks_controls.py
```diff
@@ -99,7 +100,6 @@ def __init__(self, layer):
 
         self._on_tail_length_change()
         self._on_tail_width_change()
-        self._on_properties_change()
```
kevinyamauchi, line 102: @quantumjot , I removed this call to the properties change callback. The `color_by` dropdown is populated earlier in the `__init__` method, so it seems safe to remove. It was causing the `color_by` value set in the layer init to be overwritten due to line 135 below. When the combox is cleared, the `textChanged` callback is triggered, which tries to update `layer.color_by`.

https://github.com/napari/napari/blob/a8f8eaa928445d5080683188ae9943f16d39f481/napari/_qt/layer_controls/qt_tracks_controls.py#L132-L136

Do you think this is safe to remove?

quantumjot: I think it makes a lot of sense to remove it! The properties change needs to reset the `color_by` combobox since the user *could* change the properties dict once the layer is instantiated.

kevinyamauchi: Great! Thanks for the quick feedback, @quantumjot ! I totally agree we need to keep the `_on_properties_change()` method for when the properties are altered on the model after initialization.

On a related note, a comment about the `_on_properties_change()` method: I think we may want to block the`textChanged` event from being emitted following the `color_by_combobox.clear()`. I don't think it's causing any issues now, but I think it's generally good to avoid unnecessary events/callbacks. Both for performance reasons and I find following the code through events/callbacks when debugging challenging (thus minimizing them reduces places to check). I need to check the qt docs, but I am pretty sure it's possible to temporarily block specific signals.

quantumjot: That makes a lot of sense. Glad I'm not the only one that finds the events and callbacks challenging to debug!

kevinyamauchi:
All tests are passing, so this is ready for review, @napari/core-devs !

sofroniewn (approved):
This looks good to me, thanks for adding that test @kevinyamauchi and thanks for the input @quantumjot!!

sofroniewn:
I'll leave this open for a little bit more so @VolkerH can try this out and give feedback if we wants. Then I will merge. Thanks again all

VolkerH:
As mentioned in the issue, I tested this on my code sample and found it solves the color_by initialization problem I encountered.
Thanks again.

jni:
So awesome, thanks everyone for the fast iteration here!

Issue #1423, opened 2020-07-11
Switching between monitors shrinks napari viewing window


hhelmbre:
## 🐛 Bug

When I open the napari GUI on my second monitor (Dell P2719H) and switch to my mac book pro, the viewing window within the napari GuI for my image shrinks to the bottom left corner - still works as normal with a much smaller area.

## To Reproduce

1. Open a GuI on second monitor
2. Insert Image with view_image in jupyter notebook
3. Move the viewer window from second monitor to mac book

## Expected behavior

Expected the viewer to remain the same between monitors

tlambert03:
Thanks Hawley.  I think what might be happening here is that the Dell P2719H has 1920 x 1080 native resolution, whereas a [macbook pro](https://support.apple.com/en-us/HT202471#:~:text=13%2Dinch%20MacBook%20Pro%20models,Support%20for%20millions%20of%20colors.) 15" has 2880 x 1800 native resolution (and the 16" has 3072 x 1920).  So Qt must be scaling for whatever the main monitor is when starting the program (and unable to switch scaling after launching the viewer?).  is it possible for you to test this in the other direction?  Starting on macbook pro, then moving to the monitor?

hhelmbre:
Yep! No problem the other way around.

tlambert03:
oh now that's fascinating

tlambert03:
will try to recreate

sofroniewn:
This may be related to #566, though behavior described is slightly different.

sofroniewn:
#1729 has now been merged and so this may likely be fixed! Let us know @hhelmbre if you're able to verify, otherwise I'll leave this open for a little bit longer and then close.

sofroniewn:
All indications are that this was fixed by #1729 which has merged, and so I will close this. If it turns out that this issue remains we can always reopen. Thanks all!

PR #1729, opened 2020-10-14
fix screen change glitch on pixel scale

### LINKED ISSUE #566: Connecting to external monitors breaks pan/ zoom
## 🐛 Bug

I've noticed a couple times now that when napari is already open and you connect to an external monitor the vispy canvas get's all confused with its zoom / camera / panning.

I'm not sure if you still get the same effects if you open napari after you've already connected to an external monitor. I will investigate.

Vispy has already tried to fix some of these issues - see this here https://github.com/vispy/vispy/pull/1598/files
but we might need to look into them too on napari, as it's not so great for giving live demos during talks ;-)

@jni you might remember this happened during might lightning talk at SciPy

liu-ziyang:
# Description
fixed screen change event connection, which was broken because we initialize viewer and canvas before the window
closes #566 

## Type of change

# References
https://github.com/napari/napari/issues/566

# How has this been tested?
- manually tested screen change and verified the glitch is gone

## Final checklist:

### DIFF
--- a/napari/_qt/qt_main_window.py
+++ b/napari/_qt/qt_main_window.py
@@ -69,7 +69,11 @@ class Window:
         self._qt_window = QMainWindow()
         self._qt_window.setAttribute(Qt.WA_DeleteOnClose)
         self._qt_window.setUnifiedTitleAndToolBarOnMac(True)
-        self._qt_window.windowHandle().screenChanged.connect(self.qt_viewer.canvas._backend.screen_changed)
+
+        # since we initialize canvas before window, we need to manually connect them again
+        self._qt_window.windowHandle().screenChanged.connect(
+            self.qt_viewer.canvas._backend.screen_changed
+        )
         self._qt_center = QWidget(self._qt_window)
 
         self._qt_window.setCentralWidget(self._qt_center)
--- a/napari/_qt/qt_main_window.py
+++ b/napari/_qt/qt_main_window.py
@@ -71,7 +71,7 @@ class Window:
         self._qt_window.setUnifiedTitleAndToolBarOnMac(True)
 
         # since we initialize canvas before window, we need to manually connect them again.
-        if hasattr(self.qt_viewer.canvas, "_backend"):
+        if hasattr(self._qt_window, "windowHandle"):
             self._qt_window.windowHandle().screenChanged.connect(
                 self.qt_viewer.canvas._backend.screen_changed
             )
--- a/napari/_qt/qt_main_window.py
+++ b/napari/_qt/qt_main_window.py
@@ -70,7 +70,11 @@ class Window:
         self._qt_window.setAttribute(Qt.WA_DeleteOnClose)
         self._qt_window.setUnifiedTitleAndToolBarOnMac(True)
 
-        self._on_screen_change()
+        # since we initialize canvas before window, we need to manually connect them again.
+        if hasattr(self.qt_viewer.canvas, "_backend"):
+            self._qt_window.windowHandle().screenChanged.connect(
+                self.qt_viewer.canvas._backend.screen_changed
+            )
         self._qt_center = QWidget(self._qt_window)
 
         self._qt_window.setCentralWidget(self._qt_center)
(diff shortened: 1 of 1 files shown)


liu-ziyang:
oops I pushed the branch to main repo, dang I will delete the branch after merged

sofroniewn (commented):
Amazing job identifying that the problem can be solved by adding this missing connection!! I can get access to another monitor later this evening and test this, but in the mean time I'm curious why you put this inside a function and the try catch. Are we sometimes unable to find the window? You also call out PyQt5 in the comment, but does the same apply to PySide2 or is this a PyQt5 specific problem?

Could we just get the window handle and do the connection in the init in one line?

Again really great detective work chasing all this down, I know that even though fix in the end is very short it was a lot of effort to identify this as the root cause.

sofroniewn:
Maybe @jni can test this out too. Also I'm curious if this fixes #1423 too, maybe @hhelmbre might be able to try this fix out and let us know!!

liu-ziyang:
> Amazing job identifying that the problem can be solved by adding this missing connection!! I can get access to another monitor later this evening and test this, but in the mean time I'm curious why you put this inside a function and the try catch. Are we sometimes unable to find the window? You also call out PyQt5 in the comment, but does the same apply to PySide2 or is this a PyQt5 specific problem?
> 
> Could we just get the window handle and do the connection in the init in one line?
> 
> Again really great detective work chasing all this down, I know that even though fix in the end is very short it was a lot of effort to identify this as the root cause.

The unit test for pyside version cannot find the window. However this piece is only in vispy qt backend, so I am not sure if pyside would see this issue at all

sofroniewn:
> The unit test for pyside version cannot find the window. However this piece is only in vispy qt backend, so I am not sure if pyside would see this issue at all

I think both pyside2 and pyqt5 are looking for that qt window so it is probably the same for both, but I'm not sure either.

Do I understand though that we need the try catch to make some of our unit tests pass, but we don't need it in normal operation? If so maybe we can find an alternative approach where we adjust our unit tests in some way. We've generally tried to avoid try catch statements like this in the codebase, but curious what @jni thinks in this case.

liu-ziyang:
> > The unit test for pyside version cannot find the window. However this piece is only in vispy qt backend, so I am not sure if pyside would see this issue at all
> 
> I think both pyside2 and pyqt5 are looking for that qt window so it is probably the same for both, but I'm not sure either.
> 
> Do I understand though that we need the try catch to make some of our unit tests pass, but we don't need it in normal operation? If so maybe we can find an alternative approach where we adjust our unit tests in some way. We've generally tried to avoid try catch statements like this in the codebase, but curious what @jni thinks in this case.

fwiw let me see if the _backend check does the trick, maybe we can get rid of the try catch, I was just being lazy and copied from vispy codebase

sofroniewn:
It works!!!! This is amazing!! I just tested it with another monitor. It literally would have saved my embarrassment in front of an audience of ~1000 people at SciPy two years ago 🤣 . We can sort out the testing/ try catch stuff tomorrow, but so great to have this working, it will make a big difference when people give presentations using projectors (if we ever stop using zoom that is) cc @jni

jni (approved):
Haha the sheer simplicity is beautiful, thank you @ziyangczi!

jni:
@sofroniewn just caught up on the discussion about previous versions of the code with the try/except. I agree that it would be nice if it could be avoided, but I wouldn't hold back the PR for very long if figuring out a workaround for CI is a pain. Thanks @ziyangczi for trying out various options!

sofroniewn (approved):
Looks you found the fix @ziyangczi!! All tests are passing and the functionality works as expected, and the code change is beautifully simple.

I think this is good to merge. Was there anything else you wanted to look at before doing so?

liu-ziyang:
> Looks you found the fix @ziyangczi!! All tests are passing and the functionality works as expected, and the code change is beautifully simple.
> 
> I think this is good to merge. Was there anything else you wanted to look at before doing so?

I was just waiting for people to test it out, maybe particularly on pyside2, but we merge now can reopen the issue if more to be found

PR #1738, opened 2020-10-18
Fix scale bar length in 3D

### LINKED ISSUE #1735: 3D scale bar length when adjusting window size
## 🐛 Bug

The scale bar no longer has the right length when adjusting the window size in 3D rendering mode, due to a hard coded value in #1720 that is wrong. This should be fixed in #854 by an accurate camera model.

sofroniewn:
# Description
This PR fixes the scale bar in 3D by removing a magic number and getting the correct scaling from the vispy camera matrix.
This closes #1735.

I also added a small example `examples/scale_bar.py` that can be used to verify the accuracy of the scale bar length in both 2D and 3D, see the gif below. This is ready for review.

scale_bar_position2

## Type of change

# References

Closes #1735
# How has this been tested?

- [x] added `examples/scale_bar.py`

## Final checklist:

### DIFF
--- a/examples/scale_bar.py
+++ b/examples/scale_bar.py
@@ -1,0 +1,12 @@
+"""
+Display a 3D volume and the scale bar
+"""
+import numpy as np
+import napari
+
+
+with napari.gui_qt():
+    np.random.seed(0)
+    viewer = napari.Viewer()
+    viewer.add_image(np.random.random((5, 5, 5)), colormap='red', opacity=0.8)
+    viewer.scale_bar.visible = True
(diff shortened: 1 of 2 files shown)


sofroniewn:
Note the scale bar also remains correct when the canvas resizes - which was the bug in #1735, that was hard coded to work for one value of the canvas size.

jni (approved):
Ha! Gorgeous fix. I can't believe I missed that 598 in the first review — definitely the sort of thing that would have my spidey-senses tingling during a review! :joy:

PR #1740, opened 2020-10-19
Fix world coordinates origin in 3D

### LINKED ISSUE #1739: World coordinates origin off in 3D
## 🐛 Bug

The world coordinates origin is off when toggling between 2D and 3D. This can be really clearly seen when running 
`python examples/affine_transforms.py` and toggling the number of displayed dimensions. See how the images move relative to the squares.

world_coordinates_volume

sofroniewn:
# Description
This closes #1739 by only applying the pixel scale offset when doing 2D rendering. Evidently vispy uses a top left corner pixel origin for 2D data and a center pixel origin for 3D data. This change now takes that into account, as seen in the gif below (see issue for example of broken gif):

world_coordinates_volume_fix

## Type of change

### DIFF
--- a/napari/_vispy/vispy_base_layer.py
+++ b/napari/_vispy/vispy_base_layer.py
@@ -127,9 +127,11 @@ class VispyBaseLayer(ABC):
         affine_matrix[: matrix.shape[0], : matrix.shape[1]] = matrix
         affine_matrix[-1, : len(translate)] = translate
 
-        if self._array_like:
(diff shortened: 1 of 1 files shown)


jni (approved):

sofroniewn:
Note the the zoom change will hopefully be fixed by #854

sofroniewn:
Notice this is even more clear if the volume is rendered with nearest interpolation instead of linear, so you can see the pixels in both (thanks @jni for the suggestion):

world_coordinates_volume_fix2

Issue #1636, opened 2020-09-21
Layer._get_state(), or something like it, should be public


jni:
## 🐛 Bug

The `napari_get_writer` [hookspec docs](https://napari.org/docs/plugins/hook_specifications.html#napari.plugins.hook_specifications.napari_get_writer) say:

> If `path` is a recognized format, this function should return a function that accepts the same path, and a list of tuples containing the data for each layer being saved in the form of `(Layer.data, Layer._get_state(), Layer._type_string)`

Unfortunately, `Layer._get_state()` is a private method, which should raise alarm bells, and as a consequence it is not even documented in our API docs (see e.g. for [`Image`](https://napari.org/docs/api/napari.layers.image.html)). So, to write a writer plugin, one needs to go fish around in the image layer source code!

Note: we might need to version the `get_state` API, but this should not be too hard.

sofroniewn:
I agree with this proposal, I think there was some discussion in #1584 about API around this method - we should have one function call to return something serializable, we should have another function call that returns our simple objects that plugins could use. The only place where there would be a difference right now is the `colormap` object.

adamltyson:
Resurrecting this. I'm developing a writer plugin that uses `Layer._get_state()` to work out how to write the data. Making some version of this public would be great, it feels fragile ATM.

sofroniewn:
@alisterburt and I were just talking about this on the napari-animation repo. The future direction for this will be to get the layers to be our pydantic evented models, and then we'll get `layer.dict()` and `layer.json()` methods on them for free. We've done that now for the `viewer`, but havn't make it for the layers yet. We could push on this soon. Maybe as soon as we get the `LayerGroup` stuff in from @tlambert03 we can do this? Or maybe we do it ahead of that to make that transition easier. Curious what @tlambert03 recommends the order should be. I don't think we need to wait to get slicing off the layer, that is separate.

tlambert03:
I'm not 100% sure what the order should be... That's going to be a very important and useful public API once it "settles", so I'm not sure I want to be in a hurry to "settle it"... or decide right now what order is best.  That _may_ mean that you should just use `_get_state()`  for now.  It's true it's a bit brittle... but rest assured we agree it's an important thing to make public soonish?

adamltyson:
Sounds good. I'm happy using `_get_state()` for now. At least now I'll get a notification when this changes :)

andy-sweet:
[...]

PR #1746, opened 2020-10-20
Fix for #1745, introduces `on_matrix_change` in VispyTrackLayer

### LINKED ISSUE #1745: Recently merged  Affine PR causes AttributeError: 'VispyTracksLayer' object has no attribute '_on_scale_change'
## 🐛 Bug

With the  merging of #1616 into master (@jni, @sofroniewn), a bug appeared when adding a tracks layer.
The bug does not appear with the last master branch prior to the merge.

## To Reproduce

Haven't made a generic example yet, but I submitted the following in a notebook

```
v = napari.Viewer()
v.add_tracks(tracks, name='ilastik tracks')
```

and ended up with this error:

```
AttributeError                            Traceback (most recent call last)
<ipython-input-10-05f52b983105> in <module>
      1 v = napari.Viewer()
----> 2 v.add_tracks(tracks, name='ilastik tracks')

<string> in add_tracks(self, data, properties, graph, tail_width, tail_length, name, metadata, scale, translate, opacity, blending, visible, colormap, color_by, colormaps_dict)

~/miniconda3/envs/fastai/lib/python3.7/site-packages/napari/components/add_layers_mixin.py in add_layer(self, layer)
     48         layer.events.cursor_size.connect(self._update_cursor_size)
     49         layer.events.data.connect(self._on_layers_change)
---> 50         self.layers.append(layer)
     51         self._update_layers(layers=[layer])
     52 

~/miniconda3/envs/fastai/lib/python3.7/site-packages/napari/utils/list/_model.py in append(self, obj)
     57     def append(self, obj):
     58         TypedList.append(self, obj)
---> 59         self.events.added(item=obj, index=len(self) - 1)
     60 
     61     def pop(self, key):

~/miniconda3/envs/fastai/lib/python3.7/site-packages/napari/utils/events/event.py in __call__(self, *args, **kwargs)
    512                     continue
    513 
--> 514                 self._invoke_callback(cb, event)
    515                 if event.blocked:
    516                     break

~/miniconda3/envs/fastai/lib/python3.7/site-packages/napari/utils/events/event.py in _invoke_callback(self, cb, event)
    533                 self.print_callback_errors,
    534                 self,
--> 535                 cb_event=(cb, event),
    536             )
    537 

~/miniconda3/envs/fastai/lib/python3.7/site-packages/napari/utils/events/event.py in _invoke_callback(self, cb, event)
    527     def _invoke_callback(self, cb, event):
    528         try:
--> 529             cb(event)
    530         except Exception:
    531             _handle_exception(

~/miniconda3/envs/fastai/lib/python3.7/site-packages/napari/_qt/qt_viewer.py in _add_layer(self, event)
    281         layers = event.source
    282         layer = event.item
--> 283         vispy_layer = create_vispy_visual(layer)
    284         vispy_layer.node.parent = self.view.scene
    285         vispy_layer.order = len(layers) - 1

~/miniconda3/envs/fastai/lib/python3.7/site-packages/napari/_vispy/utils.py in create_vispy_visual(layer)
     32     for layer_type, visual in layer_to_visual.items():
     33         if isinstance(layer, layer_type):
---> 34             return visual(layer)
     35 
     36     raise TypeError(

~/miniconda3/envs/fastai/lib/python3.7/site-packages/napari/_vispy/vispy_tracks_layer.py in __init__(self, layer)
     48         self._reset_base()
     49 
---> 50         self._on_data_change()
     51         self._on_appearance_change()
     52 

~/miniconda3/envs/fastai/lib/python3.7/site-packages/napari/_vispy/vispy_tracks_layer.py in _on_data_change(self, event)
     66         self.node.update()
     67         # Call to update order of translation values with new dims:
---> 68         self._on_scale_change()
     69         self._on_translate_change()
     70 

AttributeError: 'VispyTracksLayer' object has no attribute '_on_scale_change'
```

## Environment

```
napari: 0.3.9.dev50+ge99cb1e
Platform: Linux-5.4.0-51-generic-x86_64-with-debian-bullseye-sid
Qt: 5.14.2
PyQt5: 5.14.2
NumPy: 1.19.1
SciPy: 1.5.2
Dask: 2.25.0
VisPy: 0.6.4

MAX_TEXTURE_SIZE: 32768

Plugins:
- napari_plugin_engine: 0.1.7
- svg: 0.1.4
```

VolkerH:
# Description
Fixes  #1745

The introduction of affine transforms seems to have
replaced the methods `_on_scale_change` and `on_translate` with
`on_matrix_change` in `VispyBaseLayer`. 
However, the `VispyTracksLayer` was still using `on_scale_change`.
This is fixed now.

## Type of change

# References

# How has this been tested?
manually tested on a notebook that threw an exception before the fix.

## Final checklist:

I also ran black which did some minor reformatting in some otherwise bug-free lines.


sofroniewn:
@VolkerH I've added the `Tracks` to our `layer_test_data` which means it will get run on a whole bunch of tests "for free" which is really nice. It catches the bug that this PR is fixing, unfortunately we've got a couple more that will need fixing before merge. Maybe you or @quantumjot or @kevinyamauchi can check this out later today/ tomorrow and we can fix those still in this PR before merge. They could relate to my choice of "test data" which is maybe a little random. Would be great to have all those tests passing though, as it will definitely make it easier to spot regressions!!!

jni:
@sofroniewn I've made a fix to your tracks data definition. The ndim is not the "raw" dimensionality of the data but rather what gets passed to ndisplay.

As an aside we should probably add some 4D+ datasets in there!

jni:
> The ndim is not the "raw" dimensionality of the data but rather what gets passed to ndisplay.

> As an aside we should probably add some 4D+ datasets in there!

Oh wow, it's so much worse. Here:

https://github.com/napari/napari/blob/0055113d7ffc68e06c032f9cdf22ed2fbf9db0c6/napari/_tests/utils.py#L79

we check that the data ndim is the same as the ndim associated with the `layer_test_data` item. But here:

https://github.com/napari/napari/blob/0055113d7ffc68e06c032f9cdf22ed2fbf9db0c6/napari/_qt/_tests/test_qt_viewer.py#L57-L63

we use ndim to set ndisplay for the test viewer! So we actually have *no way* to test tracks right now. :joy: nor 4D image data.

quantumjot:
Looks like `rotate` also needs to be added to `Tracks`:
https://github.com/napari/napari/blob/192facf0b8bc96cb65dffe9f841270b93929fc11/napari/layers/tracks/tracks.py#L83-L86

sofroniewn:
> Looks like rotate also needs to be added to Tracks:

I've added this.

napari/layers/tracks/tracks.py
```diff
@@ -194,11 +209,9 @@ def _get_state(self):
                 'data': self.data,
                 'properties': self.properties,
                 'graph': self.graph,
-                'display_id': self.display_id,
-                'display_tail': self.display_tail,
-                'display_graph': self.display_graph,
```
sofroniewn, line 199: @quantumjot For better or for worse the keys of the `_get_state` dictionary are checked against input parameters to the function. These there are missing, as they are the display ones - is it ok if we drop them from this function? We want to improve the `_get_state` method soon, see discussion in #1636

quantumjot: I don't see why removing them would be a problem.

sofroniewn:
> @sofroniewn I've made a fix to your tracks data definition. The ndim is not the "raw" dimensionality of the data but rather what gets passed to ndisplay.
> we use ndim to set ndisplay for the test viewer! So we actually have no way to test tracks right now. 😂 nor 4D image data.

Yikes! I don't love the overload of ndim and ndisplay, but it looks like your solution of clipping ndim to get display is working ok. Something we might want to return too!

sofroniewn:
Only four more failing tests to go - all around `assert np.sum(view.dims._displayed_sliders) == ndim - 2` I think, which could be because the "range" of the data is such that sliders aren't getting shown, or something funky like that. I am about to hop on meetings for a while, but we should be able to chase this down!

jni:
Whoo!!! Thanks @sofroniewn!!!

jni (approved):

sofroniewn:
> Only four more failing tests to go - all around assert np.sum(view.dims._displayed_sliders) == ndim - 2 I think, which could be because the "range" of the data is such that sliders aren't getting shown, or something funky like that. 

Yup, as suspected this was the problem. All is fixed now and test are all passing. Thanks for getting this started @VolkerH and the contribution and review @jni. I will merge!

VolkerH:
Wow, just revisited that now. 
That blew up quickly from a small fix into a much larger change.  Thanks everyone for taking over, I admittedly would have had some difficulties tracking some of the other things down as I am not that familiar with the code base.

PR #1751, opened 2020-10-21
async-15: Octree Rendering

pwinston:
# Mainstream Changes
* `napari/_qt/qt_viewer.py`
    * Fix import to experimental code.

# Experimental Changes
* `Octree` class
    * Now an implicit tree instead of explicit nodes with children.
         * We can do this b/c for now it's a full/complete region octree.
    * Implement `Octree.get_chunks`
          * All the math to iterate over just the visible tiles.
* New `QtRender` dockable widget
    * Set the octree level you are viewing
    * Create new test images of any dimensions and any `tile_size`.
* `VispyTiledImageLayer`
    * Now draws a grid using lines, no longer modifies the tile data.

# Files
async-15-files

# Debug GUI
* Grid now drawn with lines (instead of modifying tile data)
* Upper frame shows image info and lets you set octree level
* Lower frame lets you create new image layers with any width/height/tile_size
async-15-ui


pwinston:
Marked ready for review. PR description updated to describe changes in detail. All experimental except 1 line change to fix path when importing experimental code. I will start `async-16` now.

napari/_qt/experimental/render/qt_image_info.py
```diff
@@ -0,0 +1,65 @@
...
+        self.octree_level = QtLabeledSpinBox(
+            "Octree Level",
+            max_level,
+            range(0, max_level, 1),
+            connect=on_new_octree_level,
+        )
+        self.addLayout(self.octree_level)
+
+        shape = layer.data.shape
+        height, width = shape[1:3]  # Which dims are really width/height?
```
sofroniewn, line 35: It depends on if RGB or not (and if multiscale or not) and which dimensions are the "displayed" ones. The full answer is therefor a little complex. We can ignore for now though and I can help you fix later

napari/_qt/experimental/render/test_image.py
```diff
@@ -0,0 +1,73 @@
+"""create_text_image: create a PIL image with centered text.
+"""
+import numpy as np
+from PIL import Image, ImageDraw, ImageFont
```
sofroniewn, line 4: Do you think these PIL imports will be needed in the long run?

We only depend on it inadvertently right now because ImageIO depends on it. I guess if it will be there long run we should change this comment then https://github.com/napari/napari/blob/v0.3.8/setup.cfg#L52 so people aren't confused.

Fine to leave for now though

napari/layers/image/experimental/octree.py
```diff
@@ -67,240 +74,285 @@ def _create_tiles(array: np.ndarray, tile_size: int) -> np.ndarray:
...
+    return all(x is None for x in items)
+
+
+def _combine_tiles(*tiles: np.ndarray) -> np.ndarray:
+    """Combine 1-4 tiles into a single tile.
+
+    Parameters
+    ----------
+    tiles
+        The 4 child tiles, some might be None.
```
sofroniewn, line 94: I take it right now this is `Quadtree`, but will then become `Octree` soon. That's fine, but I just want to check I am not missing something

napari/layers/image/experimental/octree.py
```diff
@@ -67,240 +74,285 @@ def _create_tiles(array: np.ndarray, tile_size: int) -> np.ndarray:
...
+
+    Future Work: Geometry
+    ---------------------
+    Eventually we want our octree to hold geometry, not just images.
+    Geometry such as points and meshes. For geometry a sparse octree might
+    make more sense than this full/complete region octree.
+
+    With geometry there might be lots of empty space in between small dense
+    pockets of geometry. Some parts of tree might need to be very deep, but
+    it would be a waste for the tree to be that deep everywhere.
```
sofroniewn, line 312: Thanks for including this note

napari/layers/image/experimental/octree.py
```diff
@@ -67,240 +74,285 @@ def _create_tiles(array: np.ndarray, tile_size: int) -> np.ndarray:
...
-            f"{indent}level={level} row={self.row:>3}, col={self.col:>3} "
-            f"shape={self.tile.shape}"
-        )
+def _one_tile(tiles: TileArray) -> bool:
+    return len(tiles) == 1 and len(tiles[0]) == 1
 
 
 class Octree:
-    """An octree.
+    """A region octree that holds hold 2D or 3D images.
```
sofroniewn, line 292: Can be a discussion for another time, but I'm curious how you see this working eventually with 4D or 5D data, say mulitcolor volumetric timeseries? Do the first two, non-spatial dimensions (color, time) just get sliced/ index in a non multiresolution way with a chunk size of (1, 1, 64, 64, 64)? Don't need to worry about now, but we know that we'll want to support data like that eventually

sofroniewn (approved):
I'm approving and will merge now. I've left a bunch of comments, but non of them need to hold up merge. Maybe the most significant is the note about PIL, but that can be addressed in the future too. Will be exciting to move forward this!!

napari/_qt/experimental/render/test_image.py
pwinston: It's nice to be able to create "test images" with numbers/letters and lines and stuff. Probably there are other ways to do that? I used PIL because it's common not because I need anything too special.

napari/layers/image/experimental/octree.py
pwinston: It started as a real Octree with 8 children per node, but we just happened to only use 4 children because we are only doing 2D. But then I realized that a "region quadtree" or "region octree" that holds images doesn't really need nodes and children. The Raveler one didn't have them.

So now it is really a region quadtree. Where you have just the N levels and each level is just a 2D array. And there are no nodes and no child pointers/references. But you can go from parent to child and back just be halving or doubling the indexes.

But everything should be the same with 3D/octree but just extended to 3D. At that point we'll probably just have one set of octree/3D code and just use it in 2D for 2D.

But yes right now it's really a region quadtree that aspires to be an octree.

napari/layers/image/experimental/octree.py
pwinston: The PCL octree is the best example of a fully developed octree for geometry I've seen. I didn't search far and wide but that's the most real-looking one I've seen:
https://pointclouds.org/documentation/group__octree.html

And yeah it's a lot more involved than a region/image octree because there's stuff in there, not just images. Basically a region/image quadtree/octree contains only 1 thing everywhere in every node, and it's always the same thing everywhere. So it's much simpler than geometry and therefore the search/algorithms are a lot more involved.

napari/layers/image/experimental/octree.py
pwinston: Even for 2D timeseries there's a choice. It's either an array of quadtrees, or it's a quadtree where every node is a timeseries array of just that tile. This is the classic "array of structures or a structure of arrays" type of thing.

I think in many cases "array of structures" is your first guess, but "structure of arrays" is very often more efficient, and I think it's more efficient here. And that's what we did for Raveller. 

The only tweak is the nodes don't contain arrays they contained dicts, so like sparse arrays. So a node might only have data for 1 slice in it. Or it might have data for all the slices if you loaded them all. However, that's basically a type of caching so it would have to work with the cache, you'd have to consider where you want to cache stuff.

For example if you have a giant stack of images and you zoom way in and then go from slice-to-slice following something, say you follow it through 100 slices. Maybe you are viewing just 12 tiles, so now those 12 nodes in the quadtree each contain a dict of 100 slices.

For 3D time series I think it's the same. So now it's a true octree, but each node is a dict/timeseries for what's loaded at that level of detail in that location. So it's a dict of sub-volumes.

Color or other channels I'd also put down into the nodes also. We did have 2 channels for Raveller the grayscale image and an RGB-style superpixel index image. So we did that and it worked.

It requires more thought though to see of napari really should do it the same way or not. I have not written anything in either direction yet. Just have not gotten there yet, but might in async-16 or 17.

napari/_qt/experimental/render/qt_image_info.py
pwinston: Yeah I mean I see us dealing with stuff like this all over. Just wasn't sure the right way to do it and the right place to put it, etc. Ideally, this GUI thing doesn't have to do anything fancy. It basically just wants to ask the layer for `(width, height)` or `(width, height, depth)` for 3D. It doesn't want to slice and dice things. Maybe a for-dummies method on `Image`, or the slice/viewer.

pwinston: But yeah this works for my test cases.

napari/layers/image/experimental/octree.py
pwinston: And maybe there is where we consider using dask deep in the internals, such that everything is dask regardless of what it started as. That may or may not be a good idea. I don't think it's obvious either way.

PR #1752, opened 2020-10-21
Fix grid size calculation

### LINKED ISSUE #1689: Grid mode auto-zoom should be improved
## 🚀 Feature

When using grid view with just two layers, the zoom is set as if I was looking at a 4x4 grid. imho it should be a bit smarter and zoom in further. Here's a small example to reproduce:

```
import numpy as np
import napari


n = 28
image = np.random.random((2, n, n))

with napari.gui_qt():
    viewer = napari.view_image(image, channel_axis=0)
    viewer.grid_view()
```

This looks like this:

Screen Shot 2020-10-04 at 1 21 59 pm

Whereas I think it should look like this:

Screen Shot 2020-10-04 at 1 22 16 pm

sofroniewn:
# Description
This PR closes #1689 by updating the calculation for the grid size. See gif below of `example/nD_labels.py`

grid_size

I'm actually hoping that multicanvas support will be one of the very next things we work on after the `0.4.0` release and that we might actually drop grid mode after that is done, so I'd rather not do too much more here. At least the behavior @jni noted is improved.

## Type of change

### DIFF
--- a/napari/_tests/test_with_screenshot.py
+++ b/napari/_tests/test_with_screenshot.py
@@ -256,7 +256,7 @@ def test_grid_mode(make_test_viewer):
 
     # enter grid view
     viewer.grid_view()
-    assert np.all(viewer.grid_size == (3, 3))
+    assert np.all(viewer.grid_size == (2, 3))
     assert viewer.grid_stride == 1
     translations = [layer.translate_grid for layer in viewer.layers]
     expected_translations = [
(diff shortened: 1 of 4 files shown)


jni (approved):
@sofroniewn for me the code changes look good, but the gif makes me wonder why it is relatively zoomed out compared to what I think would be a "fit to viewer" fit? Any thoughts? Anyway I'm approving since this is already an improvement.

sofroniewn:
> @sofroniewn for me the code changes look good, but the gif makes me wonder why it is relatively zoomed out compared to what I think would be a "fit to viewer" fit? Any thoughts? Anyway I'm approving since this is already an improvement.

Ah so this is exactly related to the comment here https://github.com/napari/napari/pull/854#issuecomment-713843658, it's basically not possible to have "fit to viewer" take aspect ratio into account and then be guaranteed not to have clipping when you resize the canvas without always calling "fit to viewer" when you resize. Ignoring aspect ratio and avoiding clipping seemed like the lesser of two evils.

jni:
Thanks @sofroniewn! It's a small thing, we can tweak the canvas behaviour any time.

PR #1753, opened 2020-10-21
Add affine doc strings

### LINKED ISSUE #1747: Affine kwarg missing from some doc strings
## 🐛 Bug

The affine kwarg is missing from some doc strings (the individual layers themselves) and needs to be added

sofroniewn:
# Description
Closes #1747 by adding missing affine doc strings for a couple of layer classes. Doc string was already present on others and text matches the others.

## Type of change

### DIFF
--- a/napari/layers/image/image.py
+++ b/napari/layers/image/image.py
@@ -85,6 +85,12 @@ class Image(IntensityVisualizationMixin, Layer):
     shear : 1-D array or n-D array
         Either a vector of upper triangular values, or an nD shear matrix with
         ones along the main diagonal.
+    affine: n-D array or napari.utils.transforms.Affine
+        (N+1, N+1) affine transformation matrix in homogeneous coordinates.
+        The first (N, N) entries correspond to a linear transform and
+        the final column is a lenght N translation vector and a 1 or a napari
+        AffineTransform object. If provided then, scale, rotate, and shear
+        values are ignored.
     opacity : float
         Opacity of the layer visual, between 0.0 and 1.0.
     blending : str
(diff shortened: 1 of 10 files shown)


napari/layers/image/image.py
```diff
@@ -85,6 +85,12 @@ class Image(IntensityVisualizationMixin, Layer):
     shear : 1-D array or n-D array
         Either a vector of upper triangular values, or an nD shear matrix with
         ones along the main diagonal.
+    affine: n-D array or napari.utils.transforms.Affine
+        (N+1, N+1) affine transformation matrix in homogeneous coordinates.
+        The first (N, N) entries correspond to a linear transform and
+        the final column is a lenght N translation vector and a 1 or a napari
+        AffineTransform object. If provided then, scale, rotate, and shear
```
GenevieveBuckley, line 92:
```suggestion
        AffineTransform object. If provided then 'scale', 'rotate', and 'shear'
```

GenevieveBuckley (approved):
One nice to have suggestion might be adding a line about compatibility with other libraries like scikit-image (and others?). 

People who look at the docstrings are really trying to answer the question: how do I use this thing? So it might be good to say you can compute your transform with another library, and then just pass that transform matrix straight back in to visualize with napari.

sofroniewn:
> One nice to have suggestion might be adding a line about compatibility with other libraries like scikit-image (and others?).
>
> People who look at the docstrings are really trying to answer the question: how do I use this thing? So it might be good to say you can compute your transform with another library, and then just pass that transform matrix straight back in to visualize with napari.

Good idea. I've added something to the doc string on the main `Affine` class, which can probably go into more detail than the kwargs on the layers. We'll certainly want to make some nice tutorials, and likely continue to improve our doc strings in the process.

Thanks for the quick review! I'll let sit a little, but let me know if you see opportunities for further improvements. I also fixed a `,` typo - but hopefully I didn't introduce any more!!

PR #1763, opened 2020-10-23
Add cursor model

### LINKED ISSUE #1614: Retrieve mouse global coordinates
## 🚀 Feature
It would be great if mouse coordinates could be returned independent of the currently active layer, i.e. "globally" in respect to the displayed data. 

## Motivation

Currently mouse coordinates can be retrieved from individual layers (`layer.coordinates`) when they are active, but it is sometimes desirable to return coordinates no matter what layer is active and tie for example a `mouse_move_callback` to the whole viewer instead of individual layers. 

## Pitch

In parallel to `layer.coordinates` there could be `viewer.coordinates` or something nested within for the central canvas that returns mouse position in respect to the displayed data. 

## Alternatives

In principle right now, coordinates can be extracted from the `viewer.status` string, which contains a list `[x,y,z]` of mouse position (I am using a regex for that currently). I didn't find a more elegant alternative for that.

sofroniewn:
# Description
This PR will close #1614 by adding a model for the cursor. Right now it only does position of the cursor in world coordinates. It's also not bidirectionally synced, so `viewer.cursor.position = ...` won't move the cursor, but is actually settable. I guess I should probably not make it settable or do the syncing, but curious what others think.

We might want to remove the cursor from the layer model, but that can come later.

## Type of change


jni (approved):
@sofroniewn this is very nice! But I am confused about how to get access to the cursor in *data* coordinates? Or maybe that's an antipattern, since the cursor is part of the viewer, not a layer. Perhaps updating some of the examples would be useful!

For reference, here's my MS imaging example in which I create a click callback based on layer coordinates (last chunk of code):

https://gist.github.com/jni/a0ae9793a0ca43868dd7dce7ea21df79

There I am using `layer.coordinates`. I guess this PR doesn't change that, but I wonder whether it might be nice to have `event.cursor.data` or similar...

sofroniewn:
> But I am confused about how to get access to the cursor in data coordinates? Or maybe that's an antipattern, since the cursor is part of the viewer, not a layer. Perhaps updating some of the examples would be useful!

Indeed, I can work on this. Ultimately I think it will be provided by a method, and the cursor position won't actually live on the layer itself - i.e. layers shouldn't know about how they are sliced or where the cursor is etc.

I'll work on this a bit more

sofroniewn:
Ok @jni I've update all I want to update. Here's the API as it is now

```python
viewer.cursor.position #position of cursor in world coordinates
layer.position #position of cursor in world coordinates, identical to viewer.cursor.position, gets updated whenever that changes
layer.coordinates #position of cursor in data coordinates, computed from layer.position based on transforms
```

I think we can continue to refine this during the `0.4` series of release. I imagine that we will drop `layer.position` and `layer.coordinates` as explicit properties ultimately under the logic of #1353 and that layers shouldn't know about where the cursor is, but that can come for another day.

Want to give this a final look before merge?

napari/layers/labels/_tests/test_labels_mouse_bindings.py
```diff
@@ -63,14 +63,13 @@ def test_paint(Event, brush_shape, expected_sum):
...
-    "brush_shape, expected_sum", [("circle", 244), ("square", 274)]
+    "brush_shape, expected_sum", [("circle", 240), ("square", 250)]
 )
 def test_paint_scale(Event, brush_shape, expected_sum):
     """Test painting labels with circle/square brush when scaled."""
     data = np.ones((20, 20))
     layer = Labels(data, scale=(2, 2))
-    layer.brush_size = 10
-    assert layer.cursor_size == 20
+    layer.brush_size = 20
```
jni, line 72: @sofroniewn can you explain why this test changed?

napari/layers/labels/labels.py
```diff
@@ -784,7 +775,13 @@ def paint(self, coord, new_label, refresh=True):
...
             # Ensure circle doesn't have spurious point
             # on edge by keeping radius as ##.5
-            radius = np.floor(self.brush_size / 2) + 0.5
+            radius = (
+                np.floor(
+                    np.mean([brush_size_dims[d] for d in self._dims.displayed])
+                    / 2
+                )
+                + 0.5
+            )
```
jni, line 788: Can you maybe expand the comment above to explain what is happening here?

napari/layers/labels/_tests/test_labels_mouse_bindings.py
sofroniewn: brush size is now in world coordinates, not in data coordinates, and cursor_size is now always identical to brush size. We don't actually need both anymore - one can be dropped from the layer. Probably cursor_size which is on the base and basically never used anymore

napari/layers/labels/labels.py
sofroniewn: comment added

sofroniewn:
Ok, I've added comments will merge once tests pass!!

napari/layers/labels/labels.py
jni: @sofroniewn I probably should have expanded on my question: I don't understand why we've gone from a single size to the mean of several sizes — which seems very brittle. Indeed, following the code above I would say that this is incorrect — in this case in data coordinates we would want an elliptical brush, right? We can merge as-is (our out-of-plane annotation has never been great anyway), **but**, the comment should reflect that this is a huge hack. 😂

sofroniewn: > in this case in data coordinates we would want an elliptical brush, right? We can merge as-is (our out-of-plane annotation has never been great anyway), but, the comment should reflect that this is a huge hack. 😂

Correct, we don't support that yet, and never have. Problem is definitely not worse now, it's always been an issue, but I will update comment

napari/_qt/qt_viewer.py
```diff
@@ -445,9 +446,14 @@ def _on_cursor(self, event):
         event : napari.utils.event.Event
             The napari event that triggered this method.
         """
-        cursor = self.viewer.cursor
+        cursor = self.viewer.cursor.style
+        # Scale size by zoom if needed
+        if self.viewer.cursor.scaled:
+            size = self.viewer.cursor.size * self.viewer.camera.zoom
+        else:
+            size = self.viewer.cursor.size
```
Czaki, line 454: whyyyy

sofroniewn: this is for painting in the labels layer to adjust the size of the paintbrush with zoom

Czaki: Ok. Problem is harder. Zoom is wrong...

PR #1765, opened 2020-10-23
Fix warning that filenames might not be set

pwinston:
# Description
I see this warning in VSCode not sure why CI does not flag it.

It's saying it cannot guarantee that filenames will be assigned to anything. Even though presumably in our usage it is always set. The one-line fix here is totally harmless and fixes the warning.

## Type of change
- [x] Fix warning

## Final checklist:


sofroniewn (approved):
LGTM! Thanks. I'll merge after CI passes

PR #1769, opened 2020-10-23
async-16: New utils.config.py

pwinston:
# Description
* Config
    * New non-experimental `utils.config.py` which controls access to experimental features
    * Replaces `os.getenv("NAPARI_ASYNC", "0") != "0"` with `config.async_loading`
    * New `NAPARI_OCTREE` variable which implies async
        * Can be tested with `config.async_octree`.
* GUI:
    * `QtRender` now shows up even with no layer, on clean startup
    * Can use it to create new layers.

# How To Use

You can test octree code using an internal test image:
1. Run `NAPARI_OCTREE=1 napari` to run with octree visuals.
2. Click **Create Test Image** button.
3. Set the octree level.

True octree rendering not hooked up yet though. So not that useful yet.

napari-octree
napari-octree-2

## Type of change
- [x] New non-experimental `utils.config.py` and related simple changes.
- [x] New features in experimental async code

# How has this been tested?
- [x] Manual testing

## Final checklist:


pwinston:
I updated this PR's description to describe the change in detail. I marked this ready for review.

This might require some discussion about the new `utils.config.py` file this PR adds. 

I think it's much better than checking the environment variables directly and cleans up a lot of things. We do need some standardized patterns for enabled experimental code. This `config.py` could be use for other experimental features beyond async. It seems like a good approach, but dirt simple.

However, there might be other ideas to consider.

pwinston:
Doing tiny cleanups in the experimental parts. But nothing substantive has changed. I think in `async-17` the first thing to do will be updating the mini-map live. With that working and correct then we can start rendering only those tiles.

napari/utils/config.py
```diff
@@ -0,0 +1,51 @@
+"""Configuration.
```
sofroniewn, line 1: I like how simple this is. I think fine to use here for now, and when we make decisions on #1183 we can decide how to harmonize

sofroniewn (approved):
Ok, this looks good to me. I think the `napari.utils.config` is fine for now, as noted in a comment above, we might want to look at again after #1183, but we can return to that later and no need to hold things up here for that.

I was also able to load in your example, was cool to see a "timeseries" one and good to see that that level of generality is going to be supported!

napari/utils/config.py
pwinston: The reason I created `config.py` was being we were checking `NAPARI_ASYNC` in many places and I wanted to add `NAPARI_OCTREE` as a 2nd var. But I want to have that one imply async. I did not want people to have to set both.

If we did that with raw env vars everywhere it would be a mess. So `config.py` does it in one centralized place. Minor in a way, but it makes the various places keying off the config cleaner.

But if we move this to some future system as well. Anything that allows flags like these.

PR #1770, opened 2020-10-24
Add warning on failed save

Czaki:
# Description
User should got information about save fail in interface, not in terminal (especially when using bundle app - I got this as response from bilogist). 

## Type of change

# References

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/_qt/qt_viewer.py
+++ b/napari/_qt/qt_viewer.py
@@ -6,13 +6,7 @@ from pathlib import Path
 import numpy as np
 from qtpy.QtCore import QCoreApplication, QSize, Qt
 from qtpy.QtGui import QCursor, QGuiApplication
-from qtpy.QtWidgets import (
-    QFileDialog,
-    QMessageBox,
-    QSplitter,
-    QVBoxLayout,
-    QWidget,
-)
+from qtpy.QtWidgets import QFileDialog, QSplitter, QVBoxLayout, QWidget
 from vispy.scene import SceneCanvas
 from vispy.visuals.transforms import ChainTransform
 
@@ -365,8 +359,7 @@ class QtViewer(QSplitter):
                 '\nor use "Save all layers..."'
             )
         if msg:
-            QMessageBox.warning(self, "Nothing to save", msg, QMessageBox.Ok)
-            return
+            raise IOError("Nothing to save")
 
         filename, _ = QFileDialog.getSaveFileName(
             parent=self,
--- a/napari/_qt/qt_viewer.py
+++ b/napari/_qt/qt_viewer.py
@@ -377,9 +377,15 @@ class QtViewer(QSplitter):
         if filename:
             with warnings.catch_warnings(record=True) as wa:
                 saved = self.viewer.layers.save(filename, selected=selected)
-                error_messages = "\n".join([str(x.message.args[0]) for x in wa])
+                error_messages = "\n".join(
+                    [str(x.message.args[0]) for x in wa]
+                )
             if not saved:
-                QMessageBox().warning(self, "Save failed", f"File {filename} save failed.\n{error_messages}")
+                QMessageBox().warning(
+                    self,
+                    "Save failed",
+                    f"File {filename} save failed.\n{error_messages}",
+                )
 
     def screenshot(self, path=None):
         """Take currently displayed screen and convert to an image array.
--- a/napari/_qt/qt_viewer.py
+++ b/napari/_qt/qt_viewer.py
@@ -1,4 +1,5 @@
 import os.path
+import warnings
 from copy import copy
 from pathlib import Path
 
@@ -372,8 +373,13 @@ class QtViewer(QSplitter):
             caption=f'Save {"selected" if selected else "all"} layers',
             directory=self._last_visited_dir,  # home dir by default
         )
+
         if filename:
-            self.viewer.layers.save(filename, selected=selected)
+            with warnings.catch_warnings(record=True) as wa:
+                saved = self.viewer.layers.save(filename, selected=selected)
+                error_messages = "\n".join([str(x.message.args[0]) for x in wa])
+            if not saved:
+                QMessageBox().warning(self, "Save failed", f"File {filename} save failed.\n{error_messages}")
 
     def screenshot(self, path=None):
         """Take currently displayed screen and convert to an image array.
--- a/napari/_qt/qt_viewer.py
+++ b/napari/_qt/qt_viewer.py
@@ -381,10 +381,8 @@ class QtViewer(QSplitter):
                     [str(x.message.args[0]) for x in wa]
                 )
             if not saved:
-                QMessageBox().warning(
-                    self,
-                    "Save failed",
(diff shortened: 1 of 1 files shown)


sofroniewn:
> User should got information about save fail in interface, not in terminal (especially when using bundle app - I got this as response from bilogist).

I completely agree with this. I know that we've still having problems with our error catching. There was some recent discussion of this here https://github.com/napari/napari/issues/1395, but we definitely need to still do better. cc @ziyangczi @jni 

Are you able @Czaki to leverage any of the error dialog message work done in #1511? I know that box doesn't yet have the ability to show stack traces yet or detailed info, which is also important in the long run too

Czaki:
> I completely agree with this. I know that we've still having problems with our error catching. There was some recent discussion of this here #1395, but we definitely need to still do better. cc @ziyangczi @jni

This is not error catching, but warning catching. There is no error thrown on fail save. 

https://github.com/napari/napari/blob/7b7936b4a485bd5602c232b6cd5ff5548fb6c05e/napari/plugins/io.py#L182-L189


> Are you able @Czaki to leverage any of the error dialog message work done in #1511? I know that box doesn't yet have the ability to show stack traces yet or detailed info, which is also important in the long run too

I will look on it, but if it is this what shown on internal error then I do not like them. It is really not intuitive. And its usage need to change behavior of save function (which breaks backward compatibility).

sofroniewn:
> This is not error catching, but warning catching. There is no error thrown on fail save.

Ah ok, understood.

> I will look on it, but if it is this what shown on internal error then I do not like them. It is really not intuitive. And its usage need to change behavior of save function (which breaks backward compatibility).

Sorry what do you mean "usage need to change behavior of save function" and "which breaks backward compatibility"? Do you mean that going through the plugin architecture is a break in backwards compatibility? I just want to make sure I understand

Czaki:
Save functions from plugins.io and layers return bool with information about succes. So there could exist code which check return value of this functions to check if save pass. Throwing exceptions instead of writing warning may break someone code.

jni:
@Czaki thanks for this fix! I agree that this is a critical issue. However, I don't think this is the fix we want to put in. One of napari's founding principles is to minimise popups. I agree with you that our current error dialogs are not functional enough for common use, but I think this means that improving our error dialogs should be prioritised, so that all these warnings could be put there.

For now, I think this should reuse the error handling machinery. Since there is *nothing* a user can do in response to this dialog, I think it's ok to put it into the small error dialogs, not by throwing errors, but by reusing the dialog code.

Czaki:
> @Czaki thanks for this fix! I agree that this is a critical issue. However, I don't think this is the fix we want to put in. One of napari's founding principles is to minimise popups. I agree with you that our current error dialogs are not functional enough for common use, but I think this means that improving our error dialogs should be prioritised, so that all these warnings could be put there.

Could you argue more about this approach? First time when I meet napari error dialog I ignore it. In comparison to the QMessagegeBox current error message dialog is not grabbing user attention and could be easily ignored if the user does not expect it.

jni:
> Could you argue more about this approach? First time when I meet napari error dialog I ignore it. In comparison to the QMessagegeBox current error message dialog is not grabbing user attention and could be easily ignored if the user does not expect it.

Yeah I think we are disagreeing about whether it's ok to be able to ignore errors or not. 😂 I think it is good for errors to be easily visible, but to not interrupt a workflow. So, as I mentioned before, I think the current error message does need to be improved, a lot, e.g. there could be different "criticality" levels which change the visibility, how long the error box stays up, the ability to recover previously dismissed errors in some kind of log window, etc. But I don't think errors should block the UI completely, which modals do.

I think the exception to that and the *big thing* that we need to be able to do in napari as an application is to match what other desktop applications do: "the viewer contains unsaved data, are you sure you want to quit?" So what I would like is for warnings/errors to be *temporarily* ignorable, but for failed saves to not be ignorable *ultimately*.

Czaki:
What I see. There are two options of usage napari:

1. Development of the new method, Playing with own code. etc. 
    In this scenario, we do not want to interups workflow. But current dialog hides important information, so people plaing with code need to set environment variable to got detailed information. 

2. Some, non programin user, trys to use some tool prepared base on napari. 
   Then such code should be error-free and user should have infdormation that something goes wrong and there is a need to report such problem to author of the tool. 

From my side, the simplest solution to use napari report system is to raise exception instead of opening QMessageBox. (And such a solution could be merged before 0.4.0 release). 

The long time solution is to add two modes of working of error dialog and base on some information decide which errors need to grab attention.

sofroniewn:
I agree with most of what has been said above, and definitely think there can be a difference between where we are at now and where we want to be.

> From my side, the simplest solution to use napari report system is to raise exception instead of opening QMessageBox. (And such a solution could be merged before 0.4.0 release).

Do you want to try this as a temporary work around @Czaki? If that doesn't work we can consider something else.

I will note that in testing this I discovered another place in the code where a QMessageBox has crept in, so we don't have zero right now, but we'll probably want to get rid of that one too! https://github.com/napari/napari/blob/v0.3.8/napari/_qt/qt_viewer.py#L341-L343

Czaki:
Done

Czaki:
But it is still not clear to me why, when such criticall errors occure, you do not like interupting message.

sofroniewn (approved):
Ok, this looks good to me!!

> But it is still not clear to me why, when such criticall errors occure, you do not like interupting message.

I think in the long run we have imagined a system where errors will get caught and logged and you can go back in time and look at them etc. without major interruptions to your workflows. 

I think there's a trend towards less blocking popups and more displays like that in browser based tools and I think that VSCode does that quite nicely sometimes too.

Hopefully like this we'd be able to keep the number of interrupting ones to basically just bare minimum such as you're exiting without saving.

I do wonder though if failing to save is so critical that we should block here too. I was unable to find an example of a tool where you were able to fail to save with just a warning, so while I'd like us to merge this now so we can make the 0.4.0 release, I do see this as something that will get revisited in the future. Thanks for bearing with us!!

sofroniewn:
Just so people can see - here is original behavior in PR

Screen Shot 2020-10-26 at 8 16 57 AM

Here is new behavior

Screen Shot 2020-10-26 at 9 13 53 AM

With original needing to be explicitly dismissed, new fading away on timer

Czaki:
> I think there's a trend towards less blocking popups and more displays like that in browser based tools and I think that VSCode does that quite nicely sometimes too.

Yes. Pop-up is to signal that you need to do some action that you not plan earlier. And the situation, when save, fail is for me such case (a user could modify data later). 


> Hopefully like this we'd be able to keep the number of interrupting ones to basically just bare minimum such as you're exiting without saving.

As mentioned already. Many people like to save the middle steps. 

> I do wonder though if failing to save is so critical that we should block here too. I was unable to find an example of a tool where you were able to fail to save with just a warning, so while I'd like us to merge this now so we can make the 0.4.0 release, I do see this as something that will get revisited in the future. Thanks for bearing with us!!

I can't find an example of another tool where you need to know which extensions are supported on saving operation. All other tools I use have a dropdown list of supported formats.

PR #1771, opened 2020-10-24
Add extent named tuple

### LINKED ISSUE #1697: drop `layer.shape` and make `_extent_*` not private
## 🚀 Refactor
I'd like to drop the `layer.shape` property and make the `_extent_data` and `_extent_world` properties public, but possibly with different names (I can never remember if it is `_extent_data` or `_data_extent`!).

Very open to suggestion for names here @kevinyamauchi @jni

sofroniewn:
# Description
This PR closes #1697 by adding a namedtuple for data and world extent, so you can do `layer.extend.data`. I quite like how it's turned out where the named tuple is really a view into a private attribute. Curious what @jni and others think. I had considered creating an [extent object like this](https://gist.github.com/sofroniewn/c825924fd6a3feef8fbdbb1ff05d0979), but think the namedtuple approach is preferable.

I can add more doc strings if people want too.

## Type of change

### DIFF
--- a/napari/layers/base/base.py
+++ b/napari/layers/base/base.py
@@ -1,5 +1,6 @@
 import warnings
 from abc import ABC, abstractmethod
+from collections import namedtuple
 from contextlib import contextmanager
 from typing import List, Optional
 
@@ -18,6 +19,8 @@ from ..utils.layer_utils import (
     convert_to_uint8,
 )
 from ._base_constants import Blending
+
+Extent = namedtuple('Extent', 'data world step')
 
 
 class Layer(KeymapProvider, ABC):
@@ -561,6 +564,15 @@ class Layer(KeymapProvider, ABC):
         return world_extent
 
     @property
+    def extent(self) -> Extent:
+        """Extent of layer in data and world coordinates."""
+        return Extent(
+            data=self._extent_data,
+            world=self._extent_world,
+            step=abs(self.scale),
+        )
+
+    @property
     def _slice_indices(self):
         """(D, ) array: Slice indices in data coordinates."""
         # clipping plane in world coordinates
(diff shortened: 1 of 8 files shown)


napari/layers/base/base.py
```diff
@@ -602,8 +614,15 @@ def shape(self):
         -------
         shape : tuple
         """
-        # To Do: Deprecate when full world coordinate refactor
-        # is complete
+        warnings.warn(
+            (
+                "The shape parameter is deprecated and will be removed in version 0.4.1."
+                " Instead you should use the extent.data and extent.world parameters"
```
```suggestion
                " Instead you should use the extent.data and extent.world attributes"
```

napari/layers/base/base.py
```diff
@@ -602,8 +614,15 @@ def shape(self):
         -------
         shape : tuple
         """
-        # To Do: Deprecate when full world coordinate refactor
-        # is complete
+        warnings.warn(
+            (
+                "The shape parameter is deprecated and will be removed in version 0.4.1."
```
```suggestion
                "The shape attribute is deprecated and will be removed in version 0.4.1."
```

napari/layers/base/base.py
```diff
@@ -602,8 +614,15 @@ def shape(self):
...
+                " Instead you should use the extent.data and extent.world parameters"
+                " to get the extent of the data in data or world coordinates."
+            ),
+            category=DeprecationWarning,
```
jni: This should (counter-intuitively, I admit) be a FutureWarning, rather than a DeprecationWarning. The latter are not visible by default.

napari/layers/image/image.py
```diff
@@ -495,8 +495,15 @@ def shape(self):
...
         """
-        # To Do: Deprecate when full world coordinate refactor
-        # is complete
+        warnings.warn(
+            (
+                "The shape parameter is deprecated and will be removed in version 0.4.1."
+                " Instead you should use the extent.data and extent.world parameters"
+                " to get the extent of the data in data or world coordinates."
+            ),
+            category=DeprecationWarning,
```
```suggestion
                "The shape attribute is deprecated and will be removed in version 0.4.1."
                " Instead you should use the extent.data and extent.world attributes"
                " to get the extent of the data in data or world coordinates."
            ),
            category=FutureWarning,
```

napari/layers/base/base.py
jni:
```suggestion
            category=FutureWarning,
```

jni (approved):
@sofroniewn I've made minor suggestions, feel free to merge after they've been applied.

jni:
@sofroniewn I've gone ahead and applied the suggestions. I'll let this sit for you to review+merge in the morning!

sofroniewn:
Changes LGTM! Thanks for the quick review @jni - I will merge now

PR #1772, opened 2020-10-24
Fix scaled painting

### LINKED ISSUE #1508: Label editing and creation with scale keyword
## 🐛 Bug

When trying to edit the labels in Napari setting scale attribute in view_image prohibits editing and drawing of the labels.

## To Reproduce

1.  You can open an example image with its labels and then use this code 

`      
scaleZ = 1
scaleX = 0.16
scaleY = 0.16                  
with napari.gui_qt():

                            viewer = napari.view_image(image, name='Raw', rgb=False, scale=[scaleZ, scaleX, scaleY]) 
                            label_layer = viewer.add_labels(labels, name='Label', scale=[scaleZ, scaleX, scaleY]) `
## Expected behavior

Editing and drawing with scaled co ordinates.

## Environment

`
napari: 0.3.6
Python: 3.7.3 (default, Mar 27 2019, 16:54:48)  [Clang 4.0.1 (tags/RELEASE_401/final)]
Qt: 5.14.2
PyQt5: 5.14.2
NumPy: 1.19.1
SciPy: 1.5.2
Dask: 2.21.0
VisPy: 0.6.4

GL version:  2.1 INTEL-12.10.12
MAX_TEXTURE_SIZE: 16384

Plugins:
  - napari-plugin-engine: 0.1.6
  - svg: 0.1.3
`

sofroniewn:
# Description
This will close #1508 by fixing the size of the cursor when painting on scaled labels layers. It's independent of #1763 where we are adding a cursor model, but we might want to revisit after #1763 merges, or I might want to update #1763 after this merges to reduce some of the complexity / make things more clear.

You can see the original problem with the following example

```python
with napari.gui_qt():
    image = np.random.random((5, 128, 128))
    labels = np.zeros((5, 128, 128))
    scale = [1, 0.16, 0.16]
    viewer = napari.view_image(image, name='Raw', rgb=False, scale=scale) 
    label_layer = viewer.add_labels(labels, name='Label', scale=scale)
```

where the cursor size is too big relative to what actually gets painted:

scaled_painting_master

you can see after this PR this has been fixed:

scaled_painting_PR

I've added a fairly crude test too.

Would be great to hear from @kapoorlab if this fixes what let them to report #1508 too. Thanks!!

## Type of change

### DIFF
--- a/napari/layers/labels/labels.py
+++ b/napari/layers/labels/labels.py
@@ -512,6 +512,8 @@ class Labels(Image):
             self.help = 'hold <space> to pan/zoom, click to pick a label'
             self.mouse_drag_callbacks.append(pick)
         elif mode == Mode.PAINT:
+            # Note we have to reset the brush size to tigger an event
+            # This will be fixed in an upcomming PR
             self.brush_size = self.brush_size
             self.cursor = self.brush_shape
             self.interactive = False
@@ -529,6 +531,8 @@ class Labels(Image):
             self.help = 'hold <space> to pan/zoom, click to fill a label'
             self.mouse_drag_callbacks.append(draw)
         elif mode == Mode.ERASE:
+            # Note we have to reset the brush size to tigger an event
+            # This will be fixed in an upcomming PR
             self.brush_size = self.brush_size
             self.cursor = self.brush_shape
             self.interactive = False
--- a/napari/layers/labels/labels.py
+++ b/napari/layers/labels/labels.py
@@ -273,7 +273,12 @@ class Labels(Image):
     @brush_size.setter
     def brush_size(self, brush_size):
         self._brush_size = int(brush_size)
-        self.cursor_size = self._brush_size / self.scale_factor
+        data2world_scale = np.mean(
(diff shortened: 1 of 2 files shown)


napari/layers/labels/labels.py
```diff
@@ -507,7 +512,7 @@ def mode(self, mode: Union[str, Mode]):
             self.help = 'hold <space> to pan/zoom, click to pick a label'
             self.mouse_drag_callbacks.append(pick)
         elif mode == Mode.PAINT:
-            self.cursor_size = self.brush_size / self.scale_factor
+            self.brush_size = self.brush_size
```
jni: This is super confusing to read. At first I thought that we could just delete this line, but seeing the setter above I realise we are using the setter for its side effects.

That's a bad smell, but we can hold our noses for a bit — as long as we add a comment above this line that this is what we are doing!

sofroniewn: ok, comment added, I think I might be able to fix this in #1763 before that merges

jni (approved):
👍!

PR #1774, opened 2020-10-25
async-17: MiniMap WIP

pwinston:
# Description
* MiniMap displays current view bounds and which tiles are in view.
    * New `qt_mini_map.py`, `qt_octree_info.py` GUI files
* Octree bookkeeping changes to support above
    * New `octree_level.py`, `octree_util.py`
    * New `OctreeIntersection` class is what the MiniMap is rendering

# Files

async-17-files-2

### DIFF
--- a/napari/layers/image/experimental/octree_intersection.py
+++ b/napari/layers/image/experimental/octree_intersection.py
@@ -1,0 +1,73 @@
+"""OctreeIntersection class.
+"""
+from typing import Tuple
+
+import numpy as np
+
+from .octree_util import OctreeLevelInfo
+
+# TODO_OCTREE: These types might be a horrible idea but trying it for now.
+Float2 = np.ndarray  # [x, y] dtype=float64 (default type)
+
+
+class OctreeIntersection:
+    # TODO_OCTREE: this class needs a lot of work
+
+    def __init__(self, info: OctreeLevelInfo, data_corners):
+        self.info = info
+        self.data_corners = data_corners
+
+        # TODO_OCTREE: don't split rows/cols?
+        self.rows: Float2 = data_corners[:, 0]
+        self.cols: Float2 = data_corners[:, 1]
+
+        base = self.info.octree_info.base_shape
+
+        self.normalized_rows = np.clip(self.rows / base[0], 0, 1)
+        self.normalized_cols = np.clip(self.cols / base[1], 0, 1)
+
+        self.rows /= self.info.scale
+        self.cols /= self.info.scale
+
+        self.row_range = self.row_range(self.rows)
+        self.col_range = self.column_range(self.cols)
+
+    def tile_range(self, span, num_tiles):
+        """Return tiles indices needed to draw the span."""
+
+        def _clamp(val, min_val, max_val):
+            return max(min(val, max_val), min_val)
+
+        tile_size = self.info.octree_info.tile_size
+
+        span_tiles = [span[0] / tile_size, span[1] / tile_size]
+        clamped = [
+            _clamp(span_tiles[0], 0, num_tiles - 1),
+            _clamp(span_tiles[1], 0, num_tiles - 1) + 1,
+        ]
+
+        # int() truncates which is what we want
+        span_int = [int(x) for x in clamped]
+        return range(*span_int)
+
+    def row_range(self, span: Tuple[float, float]) -> range:
+        """Return row indices which span image coordinates [y0..y1]."""
+        return self.tile_range(span, self.info.tile_shape[0])
+
+    def column_range(self, span: Tuple[float, float]) -> range:
+        """Return column indices which span image coordinates [x0..x1]."""
+        return self.tile_range(span, self.info.tile_shape[1])
+
+    def is_visible(self, row: int, col: int) -> bool:
+        """Return True if the tile [row, col] is in the intersection.
+
+        row : int
+            The row of the tile.
+        col : int
+            The col of the tile.
+        """
+
+        def _inside(value, value_range):
+            return value >= value_range.start and value < value_range.stop
+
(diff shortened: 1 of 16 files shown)


pwinston:
This GIF shows the MiniMap now displays which tiles are visible in the main window. We are still drawing **all** of the tiles in the main window, that has not changed yet, but the next step is we will start drawing only the visible tiles. The MiniMap shows we have that information.

We will start passing`VispyTiledImageLayer` the currently visible chunks every time the camera pans or zoom. It will maintain a "working set" of tiles. It will create new tiles or remove old ones. Tiles that are the same it doesn't need to touch.

Today `VispyTiledImageLayer` creates a new `ImageVisual` for every new chunk. We know drawing a `ImageVisual` for the first time takes a while (up to 20ms?). If this is too slow, David Hoese's idea is that we should write a custom new`TiledImageVisual` that's more efficient. It would be one single visual. It would maintain a texture atlas-like tile cache. It would draw all the tiles with one draw command.

napari-mini-map

pwinston:
Note: The PR Is very unstable right now, with crashes. Not sure what's causing them yet.

sofroniewn:
Wow, I love the minimap!! So cool :-)

pwinston:
Now the minimap shows the actual view bounds. Things are still unstable I think it's from drawing too many new `ImageVisuals` in one frame. Will try to pace it out a few per frame, which we need to do anyway. If that doesn't help might be forced to do the single-visual `TiledImageVisuals` sooner rather than later.

napari-minimap-2

sofroniewn:
Wow, I love the minimap so much, definitely new favourite feature!!

kevinyamauchi:
This is so cool, @pwinston !

pwinston:
I added some more octree info on the right. And in the octree level dropdown you can pick AUTO, which will be the normal mode. But it's not hooked up yet:

napari-octree-3

pwinston:
I marked this ready for review. It's 100% experimental except for a 1 line perf time change to print less digits (print up to microseconds only).

pwinston:
I want to clear the decks for `async-18` because I think it will have some non-experimental changes.

We need`VispyTiledImageLayer` to update itself as the camera moves. This is unlike normal visuals. With a normal visual we throw it into the scene once, and then we let the camera move around and draw that fixed things from different views.

Instead for octree `VispyTiledImageLayer` needs to shuffle tiles in/out as the camera moves. So it's not one fixed thing getting drawn from multiple views. It kind of morphs itself based on where the camera is.

I was thinking just pass the camera into `VispyTiledImageLayer`. Probably into `VispyBaseLayer` because where we create these layers in `create_vispy_visual()` we create them uniformly. All current visuals would ignore the incoming camera, but `VispyTiledImageVisual` could hook up events to it.

But maybe there is some other way to get the camera into `VispyTiledImageLayer` or else have some external thing receive the camera events and then call into `VispyTiledImageLayer`? Not positive what is the best approach.

Is `async-18` I'm passing the camera into just `VispyTiledImageLayer`, other layers unchanged:
https://github.com/pwinston/napari/blob/feature/async-18/napari/_vispy/utils.py
This may or may not be a good idea, but it will get the ball rolling.

pwinston:
Okay I'm moving on to `async-18`.

napari/layers/image/experimental/octree_intersection.py
```diff
@@ -0,0 +1,73 @@
+"""OctreeIntersection class.
+"""
+from typing import Tuple
+
+import numpy as np
+
+from .octree_util import OctreeLevelInfo
+
+# TODO_OCTREE: These types might be a horrible idea but trying it for now.
+Float2 = np.ndarray  # [x, y] dtype=float64 (default type)
```
sofroniewn, line 10: yeah, personally I don't love this and we don't do it in many (or any?) other places, but no need to change right now

sofroniewn (approved):
Ok, this looks fine to me. Thanks for all the "ToDo" notes, helps me get a sense of what is in flux and still needs work. I really love the minimap for debugging/ understanding purposes, makes it very clear whats going on.

Will be exciting when this gets hooked up to real data!!

napari/layers/image/experimental/octree_intersection.py
pwinston: Above is a pseudo-shader convention, they have types like `vec2`, `vec3`, `vec4`, `ivec3`, `uvec3`, etc. But yes probably not appropriate for mainstream napari.

What is our convention is for `ndarray`, comments and/or typing? It does seem like passing around `ndarray` with no indication of the shape or dtype is like passing around `void *` in C++, it's a mystery box.

Should the receiver assert the `ndarray` is the expected shape or type? Or just crash or go wonky if the `ndarray` is wrong? Whatever is napari and/or scipy convention I want to do that. This was just for my own benefit while getting things setup.

sofroniewn: > Whatever is napari and/or scipy convention I want to do that.

I just did some quick poking around. Nothing comprehensive, but gives some rough ideas

In numpy you might see methods with names like `_assert_2d` at the beginning of functions see here https://github.com/numpy/numpy/blob/master/numpy/linalg/linalg.py#L539 and it's definition here https://github.com/numpy/numpy/blob/43683b3256a86659f230dcadbcde1f8020398bfa/numpy/linalg/linalg.py#L193 that will do that sort of dimensionality checking

In PyTorch you can see here an example of computing the length of something and checking it here https://github.com/pytorch/pytorch/blob/master/torch/nn/grad.py#L19-L21

We've also done that to a limited extent within napari already - see here for example, https://github.com/napari/napari/blob/master/napari/layers/shapes/_shapes_models/line.py#L50-L55.

So it seems like the conventions are more along the lines of "receiver assert the ndarray is the expected shape or type"

PR #1780, opened 2020-10-26
Adjust welcome colors of text

sofroniewn:
# Description
This PR follows on from #1721 by fine tuning some of the colors of the text on the welcome visual. Manual inspection of the colors with photoshop shows that the color balance for vispy text is off then rendering text with opacity < 1. It looks like the blending is not happening correctly. We need opacity < 1 though as otherwise there are very strong alias like artifacts at the edge of the text.

I've fine tuned things to improve the color for our default theme. This is clearly a temporary solution, but better than before. I also made the font no longer bold, and made the rendering use the gpu as quality is supposedly higher but that didn't actually improve anything. cc @jni @alisterburt 

Before:
Screen Shot 2020-10-26 at 9 04 01 AM

After:
Screen Shot 2020-10-26 at 9 02 25 AM

## Type of change

### DIFF
--- a/napari/_vispy/vispy_welcome_visual.py
+++ b/napari/_vispy/vispy_welcome_visual.py
@@ -63,6 +63,8 @@ class VispyWelcomeVisual:
             background_color = np.divide(
                 str_to_rgb(darken(self._viewer.palette['background'], 70)), 255
             )
+            # Note this unsual scaling is done to preserve color balance on
+            # rendering by VisPy, which appears to be off when opacity < 1
             text_color = np.multiply(foreground_color, [0.2, 0.45, 0.7])
         else:
             foreground_color = np.divide(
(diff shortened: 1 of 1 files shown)


alisterburt:
this looks great!

sofroniewn:
Note that as we get further along with our design improvements we'll want to make sure that text color conforms to accessibility color contrast recommendations, but we can revisit this when we do a full accessibility audit.

sofroniewn:
I've made text a little brighter

Screen Shot 2020-10-26 at 5 21 39 PM

sofroniewn:
Ok, we've taken some advice from @LCObus and @liaprins-czi from https://github.com/napari/napari/pull/1721#issuecomment-716727745 and https://github.com/napari/napari/pull/1721#issuecomment-716785922 and make a couple more improvements so it now looks as follows

Screen Shot 2020-10-26 at 5 31 53 PM

jni (approved):
Very nice!

PR #1788, opened 2020-10-28
Fix welcome text contrast

### LINKED ISSUE #1787: Text in welcome visual is too dark on non-macOS platforms
## 🐛 Bug

The welcome text is almost unreadable on Windows or Linux:

signal-2020-10-28-132757

IMG_5476

It's fine on mac:

Screen-Shot-2020-10-27-at-7 12 45-PM

@sofroniewn noted that he had to make the text darker for it to appear normal on mac. It appears that this is due to some bug in macOS graphics — the workaround should be applied only on mac.

sofroniewn:
# Description
Attempt to close #1787 by only applying additional scaling on macOS.

In long term we should move welcome visual to be done with Qt widget and not vispy, but that can come later

cc @jni @VolkerH 

## Type of change

### DIFF
--- a/docs/release/release_0_4_0.md
+++ b/docs/release/release_0_4_0.md
@@ -111,6 +111,7 @@ We thank the many contributors who have made this release possible!
 - Fix setup config duplication of dask requirement (#1778)
 - Adjust welcome colors of text (#1780)
 - Fix gray to grays for vispy (#1784)
+- Fix welcome text contrast (#1788)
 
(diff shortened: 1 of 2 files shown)


jni (approved):
@sofroniewn this looks good on Linux now! :tada: contrast is probably still too low from an accessibility standpoint but we can fix later, at least it's not unreadable even for those with good eyesight :joy:

jni:
napari-welcome-pr

PR #1793, opened 2020-10-29
async-18: Basic Quadtree Rendering

pwinston:
# Description

**Basic quadtree rendering:**
1. Given a view, choose an appropriate quadtree level for that view.
2. Render only the chunks (tiles) which are within those view bounds.

You can pan around the image while zoomed in. The tiles are loaded/unloaded as you pan, so it seems like the image exists everywhere. Same thing if you zoom, but while zooming it might switch to a higher or lower level of detail.

Most work is in `vispy_tiled_image_layer.py` and it has been cleaned up to be pretty "real".

**Smaller Changes:**

* Minimap improvements, more accurate.
* Add **Track View** checkbox to `QtRender`.
    * For debugging, uncheck this to "freeze" which tiles are rendered.
    * Can zoom out and see just the rendered tiles.
* Add **Octree Image** checkbox to `QtRender`
    * Can choose between creating a normal `Image` or an `OctreeImage`.
    * This is useful to compare how `Image` and `OctreeImage` render the same thing.
    * You can toggle layer visibility to compare both versions.

# Type of change
- [x] New features in experimental code
- [x] Small changes to regular code detailed below

# Non-experimental changes

All are meant to be backward-compat and harmless to mainstream use:

* **`add_layers_mixin.py`**
   * Simplify how we choose `OctreeImage` or `Image`.
   * Now references `config.create_octree_images` **dynamically**.
   * In `QtRender` widget you can chose whether to create `Image` or `OctreeImage` layers.
* **`qt_viewer.py`**
    * We now pass the camera into `create_vispy_visual`.
* **`_vispy/utils.py`**
    * `create_vispy_visual` now connects each visual's `VispyBaseLayer._on_camera_move`.
    * Only `OctreeImage` reacts to camera moves today, but any layer might want to.
* **`vispy_base_layer.py`**
    * New `_on_camera_move` method which `OctreeImage` uses.
* **`utils/config.py`**
    * New `create_octree_images` config.
    * Although this is toggled at runtime I put it in config because it's part of the async/octree feature toggle. When we remove the feature toggle there will hopefully be no more `OctreeImage` class. Instead the regular `Image` class will render using the octree. So then this config option will go away.

# How has this been tested?
- [x] Manual testing

# Details
* Today we are working with single-scale images only.
    * We compute an octree on the fly when the slice is created.
    * Thus creating an octree-style multi-scale image on the fly:
        * Computing the octree is 0.5 seconds for a 1024 x 1024 image.
        * Over a minute for a 12,000 x 12,000 image (it's 144 times bigger).
        * All the time is downsampling, chunking level 0 is basically free.
* Something about creating or adding `ImageVisuals` leads to crashes.
    * Spent half a day looking into this and it's still a mystery.
    * We now have a reusable pool of `ImageVisuals` which helps, but it can still crash.

# Future work
* Hook the octree up to the `ChunkLoader` so we can do async loading of chunks.
* Hook the octree up to Zarr/Dask data, so we can load pre-tiled images.
    * Probably 99% of real usage will be from files like these.
* Fix crashes related to creating ImageVisuals
    * One solution might be to write a `TiledImageVisual` as David suggested.
    * However this still leaves napari vulnerable to these types of crashes.
    * For example drawing or not drawing the `WelcomeVisual` strongly influences the crash rate!
    * Ideally fix the crashes but then also write `TiledImageVisual` for speed?

# Videos

napari-async-18-rendering

* Video of above GIF but longer: https://youtu.be/KXEQtYpxxyI

napari-octree-image

* Video of above GIF but longer: https://youtu.be/8dbWD1y361k

This is a 12,000 x 12,000 image. It took over a minute to compute the octree, almost 100% of the time was downsampling. This is is why real multi-scale images will generally be in Zarr or similar chunked formats with pre-computed levels of detail.

### DIFF
--- a/napari/_qt/qt_viewer.py
+++ b/napari/_qt/qt_viewer.py
@@ -312,7 +312,9 @@ class QtViewer(QSplitter):
         """
         layers = event.source
         layer = event.item
-        vispy_layer = create_vispy_visual(layer, self.viewer.camera)
+        vispy_layer = create_vispy_visual(layer)
+        self.viewer.camera.events.center.connect(vispy_layer._on_camera_move)
+
         vispy_layer.node.parent = self.view.scene
         vispy_layer.order = len(layers) - 1
         self.layer_to_visual[layer] = vispy_layer
--- a/napari/_qt/qt_viewer.py
+++ b/napari/_qt/qt_viewer.py
@@ -320,7 +320,7 @@ class QtViewer(QSplitter):
         """
         layers = event.source
         layer = event.item
-        vispy_layer = create_vispy_visual(layers, self.viewer.camera)
+        vispy_layer = create_vispy_visual(layer, self.viewer.camera)
         vispy_layer.node.parent = self.view.scene
         vispy_layer.order = len(layers) - 1
         self.layer_to_visual[layer] = vispy_layer
--- a/napari/_qt/qt_viewer.py
+++ b/napari/_qt/qt_viewer.py
@@ -1,17 +1,12 @@
 import os.path
+import warnings
 from copy import copy
 from pathlib import Path
 
 import numpy as np
 from qtpy.QtCore import QCoreApplication, QSize, Qt
 from qtpy.QtGui import QCursor, QGuiApplication
-from qtpy.QtWidgets import (
-    QFileDialog,
-    QMessageBox,
-    QSplitter,
-    QVBoxLayout,
-    QWidget,
-)
+from qtpy.QtWidgets import QFileDialog, QSplitter, QVBoxLayout, QWidget
 from vispy.scene import SceneCanvas
 from vispy.visuals.transforms import ChainTransform
 
@@ -40,6 +35,7 @@ from .._vispy import (  # isort:skip
     VispyAxesVisual,
     VispyCamera,
     VispyScaleBarVisual,
+    VispyWelcomeVisual,
     create_vispy_visual,
 )
 
@@ -60,6 +56,9 @@ class QtViewer(QSplitter):
     ----------
     viewer : napari.components.ViewerModel
         Napari viewer containing the rendered scene, layers, and controls.
+    welcome : bool
+        Flag to show a welcome message when no layers are present in the
+        canvas.
 
     Attributes
     ----------
@@ -95,7 +94,7 @@ class QtViewer(QSplitter):
 
     raw_stylesheet = get_stylesheet()
 
-    def __init__(self, viewer):
+    def __init__(self, viewer, welcome=True):
         # Avoid circular import.
         from .layer_controls import QtLayerControlsContainer
 
@@ -207,6 +206,17 @@ class QtViewer(QSplitter):
         )
         self.canvas.events.resize.connect(self.scale_bar._on_position_change)
 
+        self._show_welcome = welcome
+        if self._show_welcome:
+            self.welcome = VispyWelcomeVisual(
+                self.viewer, parent=self.view, order=-100
+            )
+            self.viewer.events.layers_change.connect(
+                self.welcome._on_visible_change
+            )
+            self.viewer.events.palette.connect(self.welcome._on_palette_change)
+            self.canvas.events.resize.connect(self.welcome._on_canvas_change)
+
         main_widget = QWidget()
         main_layout = QVBoxLayout()
         main_layout.setContentsMargins(10, 22, 10, 2)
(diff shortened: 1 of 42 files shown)


pwinston:
I'm marking this PR ready for review.

Next steps:

1. Investigate the "too many visuals" vispy crash. Even if we implement a single `TiledImageVisual`, these crashes are going to haunt us because they can appear even with relatively small numbers of visuals. I've seen crashes with less than 20 visuals.

2. Start `async-19` and beyond which hopefully will marry the `ChunkLoader` to the octree. Immediately followed by making it so chunked file formats like Zarr are loading using the octree and the `ChunkLoader`. The goal is viewing large remote (high latency) multi-scale images at interactive speeds.

3. Some further out tasks:
    * Replaced the multi-visual `VispyTiledImageLayer` with the single visual `TiledImageVisual` (large).
    * Support 3D and volume rendering (large).

napari/_vispy/utils.py
```diff
@@ -16,17 +18,17 @@
...
-    layer_to_visual = {OctreeImage: VispyTiledImageLayer}
-    layer_to_visual.update(original)
+    # Insert OctreeImage in front so it gets picked over plain Image.
+    new_order = {OctreeImage: VispyTiledImageLayer}
+    new_order.update(layer_to_visual)
+    layer_to_visual = new_order
 
 
-def create_vispy_visual(layer):
+def create_vispy_visual(layer: Layer, camera: Camera) -> VispyBaseLayer:
```
sofroniewn: I'm not sure about this change. I don't necessarily have a huge objection to it, but I probably would have done this here https://github.com/napari/napari/blob/master/napari/_qt/qt_viewer.py#L315-L317 by adding a
```python
vispy_layer = create_vispy_visual(layer)
self.viewer.camera.events.center.connect(vispy_layer._on_camera_move)
```
I think this would achieve the same outcome but without changing this function

Curious about your thoughts on the relative merits of either approach?

napari/_qt/qt_viewer.py
```diff
@@ -312,7 +312,7 @@ def _add_layer(self, event):
         """
         layers = event.source
         layer = event.item
-        vispy_layer = create_vispy_visual(layer)
+        vispy_layer = create_vispy_visual(layer, self.viewer.camera)
```
sofroniewn:
```suggestion
        vispy_layer = create_vispy_visual(layer)
        self.viewer.camera.events.center.connect(vispy_layer._on_camera_move)
```
See comment below, I would have taken this approach. It's not a huge deal, but curious your thoughts on the relative merits

napari/_vispy/vispy_base_layer.py
```diff
@@ -151,3 +151,7 @@ def _reset_base(self):
         self._on_opacity_change()
         self._on_blending_change()
         self._on_matrix_change()
+
+    def _on_camera_move(self, event=None):
+        """Camera was moved."""
+        pass
```
sofroniewn, line 157: I'm surprised you don't also need to hook up `_on_camera_zoom` too? i.e. `camera.events.zoom`

sofroniewn: I guess the zoom knowledge goes through `layer._update_draw`. https://github.com/napari/napari/blob/v0.4.0/napari/_qt/qt_viewer.py#L694-L702

I think this is ok now, but curious how you think about that. In the long run we might want to get rid of that method/ reduce the number of places where this data flows.

Given considerations like "the layer should have no knowledge of the currently viewed slice" it makes sense to not rely on that method more and use the camera where we can

napari/components/add_layers_mixin.py
```diff
@@ -18,27 +18,16 @@
 logger = getLogger(__name__)
 
 
-def _get_async_image_class():
-    """Return layer.Image or OctreeImage."""
-    if config.async_octree:
+def _get_image_class() -> layers.Image:
```
sofroniewn, line 21: changes here look good to me

sofroniewn (approved):
Ok, this looks good to me. I've made one suggestion/ question around connecting the camera event that might be good to get feedback on before merge.

I've also added one question about connecting to the camera.zoom event, which we might want to start using in the future, but that could come in later PRs

napari/_vispy/utils.py
pwinston: Yeah I changed it to that, much better. It was just an artifact of how I arrived at this. First I was passing the camera INTO the VispyLayer. Then I changed to connecting the event and didn't think to pull it out. So this is better.

napari/_qt/qt_viewer.py
pwinston: Yeah I push a change to this way.

napari/_vispy/vispy_base_layer.py
pwinston: Yeah I'd have to think about this more. Zoom seems to mostly change the center I guess. But maybe if you do a perfectly centered zoom we'll need to add zoom!

pwinston:
Updated with the change.

napari/_vispy/utils.py
sofroniewn: ok, nice. thanks!

napari/_vispy/vispy_base_layer.py
sofroniewn: Ha! Yeah, we may need this eventually

sofroniewn:
Ok, requested changes made and tests all passing. Will merge now!

PR #1795, opened 2020-10-30
Update setup.cfg to add aliases for pyqt5 and pyside in extras_require

GenevieveBuckley:
# Description

This PR adds some aliases for the `extras_require` when installing napari. I sometimes find myself doing `pip install napari[pyqt]` which appears to install correctly, but because "pyqt" isn't an option for extras (should be "pyqt5" instead) I end up with no GUI backend installed.

It's not a big problem, just something that might be nice for convenience.

## Type of change

# References

# How has this been tested?

## Final checklist:
- [ ] I have made corresponding changes to the documentation **I haven't done this, but I think that might be confusing. This is really to catch typos/abbreviations when someone types the install command from memory.**
- [ ] ~~I have added tests that prove my fix is effective or that my feature works~~

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -70,8 +70,14 @@ install_requires =
 [options.extras_require]
 pyside2 = 
     PySide2>=5.12.3,<5.15.0
+pyside =  # alias for pyside2
+    %(pyside2)s
 pyqt5 = 
     PyQt5>=5.12.3,<5.15.0
+pyqt =  # alias for pyqt5
+    %(pyqt5)s
+qt =  # alias for pyqt5
+    %(pyqt5)s
 # all will be the full "batteries included" extra.
 # though for now, it's just the GUI backend
 all =


jni (approved):
I like this. I think there were some discussions about how this can be confusing because of the different versions (PySide is a different package from PySide2), but I don't think it's that terrible, since the original packages are not in widespread use anymore.

sofroniewn (approved):
Yeah, I'm fine with this too. I think the potential ambiguity does not out weigh the benefit, the other names relate to Qt4 which we've never started and we can decided what to do with our defaults etc. when Qt6 comes out and if we start supporting that.

Looks like we had some CI cancelations (too many concurrent runs?) so I've just triggered some manual reruns. I'll wait until they all pass and then merge.

Thanks for contribution @GenevieveBuckley!!

sofroniewn:
This PR seems to have many more test runs than our other PRs, it's like the full matrix ran including windows 3.7 and windows 3.8 etc. Am I understanding that right? Is that expected @GenevieveBuckley or am I just confused. cc @jni?

jni:
@sofroniewn I noticed that also and thought maybe something had changed in master, but I see now that it hasn't. 

If I look at the `.cirrus.yml`, I see that those checks are there conditional on `$CIRRUS_PR == ""`. Maybe something weird happened with Cirrus images during this one PR that has since been reverted? :shrug:

jni:
Oh! Another hypothesis (I think probably correct): this PR comes from a branch on this repo, rather than from a fork. @GenevieveBuckley can you avoid doing that in the future? I think it is what GitHub defaults to in the web editor but there might be ways to avoid it.

Issue #1387, opened 2020-06-25
Executed examples in the docs


jni:
## 📚 Documentation
It would be great to be able to use sphinx-gallery together with `nbscreenshot` to run our examples for our documentation and embed screenshots.

The tutorials would also be great to run automatically, although that might be not be feasible since some use large datasets, or demonstrate interactivity.

sofroniewn:
Indeed, see discussion here from @GenevieveBuckley for considerations for our tutorials https://github.com/napari/tutorials/issues/70. I think a good first step though it just getting all our examples in this repo run and included

GenevieveBuckley:
Continuing the discussion from https://github.com/napari/napari/issues/1568 (it seems better placed here, since this issue is not specific to the nbclient library):

https://github.com/napari/napari/issues/1568#issuecomment-716382897
> Exciting news! If we use [runipy](https://github.com/paulgb/runipy)* to execute jupyter notebooks we can finally run them with the `%gui qt` magic. (See the discussion [here](https://github.com/jupyter/nbclient/issues/97#issuecomment-716366259) for more details) 🚀 🚀 🚀
> 
> *Note: runipy is no longer maintained, but it's very small so we could potentially maintain a fork of it in the napari org. I'd say the benefits would be worth it.
> 
> ```
> pip install runipy
> runipy example-notebook.ipynb
> ```
> 
> ```python
> # cell 1
> %gui qt
> 
> # cell 2 - this MUST be in a separate cell after the qt magic above, you can't combine these two into one cell
> import time
> time.sleep(5)
> 
> # cell 3
> import napari
> viewer = napari.Viewer()
> ```

GenevieveBuckley:
Followup from @jni's suggestion at the dev meeting:
I can use `runipy` to run a notebook with napari and then save the executed notebook with outputs (and especially `nbscreenshot` outputs) to a separate file. Yay!

One slight hiccup with this is that it saves an older notebook json format, not the current notebook version 4 format. When you open the output with jupyter it's clever enough to detect and convert this for you, but I'm not sure if sphinx will be able to handle this. So we might need to tweak how the output json is written to file.

GenevieveBuckley:
Interestingly, sometimes I get a strange error (but it's very inconsistent). It doesn't happen every time, but it happens quite a bit. If you re-run the command enough times it ends up working.

Here is me finding this problem, then re-running with exactly the same command and having it work fine.
```
[...]

PR #1796, opened 2020-10-30
Examples tweaks

will-moore:
# Description
A couple of minor additions to the examples. Things that I found by digging in the source code but
I think would benefit others if easier to find.

## Type of change
- [ ] Addition to examples

# References
Source code

# How has this been tested?
- [ ] ran the `custom_mouse_functions.py` to test the new code

## Final checklist:

### DIFF
--- a/examples/custom_mouse_functions.py
+++ b/examples/custom_mouse_functions.py
@@ -50,3 +50,21 @@ with napari.gui_qt():
             msg = f'clicked at {cords} on background which is ignored'
         layer.status = msg
         print(msg)
+
+
+    # Handle click or drag events separately
+    @labels_layer.mouse_drag_callbacks.append
+    def click_drag(layer, event):
+        print('mouse down')
+        dragged = False
+        yield
+        # on move
+        while event.type == 'mouse_move':
+            print(event.pos)
+            dragged = True
+            yield
+        # on release
+        if dragged:
+            print('drag end')
+        else:
+            print('clicked!')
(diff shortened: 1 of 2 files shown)


will-moore:
Also, I wonder if the https://github.com/napari/napari/tree/master/examples page should be mentioned in the tutorials pages. Maybe in the footer under the 'fundamentals' and 'applications' lists? I was going to open a PR to do that but noticed that all the other links are internal 'tutorials' links so maybe there's somewhere else it can go.
It's just that the examples include a lot of useful stuff that and a new user might search for a while before finding it?

Also noticed that https://napari.org/docs/ needs 0.4.0 adding?

sofroniewn:
> Also noticed that https://napari.org/docs/ needs 0.4.0 adding?

Ah nice catch, we really need to improve that page! @kne42 can you add?

> Also, I wonder if the https://github.com/napari/napari/tree/master/examples page should be mentioned in the tutorials pages. Maybe in the footer under the 'fundamentals' and 'applications' lists? I was going to open a PR to do that but noticed that all the other links are internal 'tutorials' links so maybe there's somewhere else it can go.
It's just that the examples include a lot of useful stuff that and a new user might search for a while before finding it?

Yeah we should do that! I thought there might have been, but I wasn't sure. @GenevieveBuckley was recently wondering about whether we might even want to combine the examples and tutorials into one repo. She's been working on some autogeneration functionality in #1387 which would make this incredible.

sofroniewn (approved):
Changes look great. I will merge, but the other points in the comments are worthy of continued discussion. Thanks very much for this @will-moore!!

Issue #855, opened 2020-01-08
Refactor vispy canvas out of qt_viewer.py into own file


sofroniewn:
## 🚀 Feature
After #854 goes in I'd like to refactor the vispy code out of `qt_viewer` into its own file maybe inside the `_vispy` folder or maybe called `QtCanvas` and still in the `_qt` folder.

The goal is to simplify the `QtViewer` code and make the vispy code more separated from the rest of the code.

Curious what @jni @tlambert03 thinks, and how this relates to organization in #823?

GenevieveBuckley:
It looks like this has been addressed by https://github.com/napari/napari/pull/5432 @melonora can you confirm?

melonora:
@GenevieveBuckley I can confirm:)

PR #1798, opened 2020-10-30
Fix ValueError in Layer.coordinates during layer load

pwinston:
# Description

Prevent `ValueError` in `Layer.coordinates` on load under specific conditions. The exception was happening when dragging-n-dropping a 2D time-series file and moving the mouse over the main window while that file is loading. The time-series is needed because the bug is related to things being in a mixed state while it's switching from 2D to 3D coordinates.

# Steps

This was happening all the time with the Octree work. But you can reproduce in master:

1) Start napari with no data.
2) Drop a 2D timeseries, such as a Zarr file `random.random((300, 300, 300))`
3) Shake the mouse rapidly over the main window as soon as you drop the file.

If you can't get the timing right put a `sleep(1)` in `Image.__init__()`.

# Stack

    11:41:03 ERROR Unhandled exception:
    File "/Users/pbw/dev/vispy/vispy/app/backends/_qt.py", line 473, in mouseMoveEvent
        self._vispy_mouse_move(
    File "/Users/pbw/dev/vispy/vispy/app/base.py", line 213, in _vispy_mouse_move
    File "/Users/pbw/dev/vispy/vispy/util/event.py", line 455, in __call__
    File "/Users/pbw/dev/vispy/vispy/util/event.py", line 473, in _invoke_callback
        _handle_exception(self.ignore_callback_errors,
    File "/Users/pbw/dev/vispy/vispy/util/event.py", line 471, in _invoke_callback
        cb(event)
    File "/Users/pbw/dev/napari/napari/_qt/qt_viewer.py", line 625, in on_mouse_move
        self.viewer.cursor.position = self._map_canvas2world(list(event.pos))
    File "/Users/pbw/dev/napari/napari/components/cursor.py", line 55, in position
        self.events.position()
    File "/Users/pbw/dev/napari/napari/utils/events/event.py", line 514, in __call__
    File "/Users/pbw/dev/napari/napari/utils/events/event.py", line 531, in _invoke_callback
        _handle_exception(
    File "/Users/pbw/dev/napari/napari/utils/events/event.py", line 529, in _invoke_callback
        cb(event)
    File "/Users/pbw/dev/napari/napari/components/viewer_model.py", line 394, in _on_cursor_position_change
        layer.position = self.cursor.position
    File "/Users/pbw/dev/napari/napari/layers/base/base.py", line 486, in position
        self._update_value_and_status()
    File "/Users/pbw/dev/napari/napari/layers/base/base.py", line 904, in _update_value_and_status
        self._value = self.get_value()
    File "/Users/pbw/dev/napari/napari/layers/base/base.py", line 866, in get_value
        return self._get_value()
    File "/Users/pbw/dev/napari/napari/layers/image/image.py", line 781, in _get_value
        coord = self.coordinates
    File "/Users/pbw/dev/napari/napari/layers/base/base.py", line 900, in coordinates
        return tuple(self._transforms[1:].simplified.inverse(self.position))
    File "/Users/pbw/dev/napari/napari/utils/transforms/transforms.py", line 328, in __call__
        linear_matrix[
    ValueError: could not broadcast input array from shape (3,3) into shape (2,2)

# Analysis

This stack is from Octree but it's the same in master. The problem was `get_max_texture_sizes()` created a temporary Canvas and closing that Canvas caused `_vispy_close()` to process Qt events. Which is very dangerous of vispy. If it happened to process a mouse event, that might call into our code with things in an inconsistent state because the Layer had not fully loaded yet.

napari-stack

# Solution

1. Move `get_max_texture_sizes()` from `vispy_base_layer.py` to its own new file `_vispy/utils_gl.py`. Getting the texture size is a very generic utility, it's not really related to `VispyBaseLayer`. It can't go into `_vispy/utils.py` because of circular dependencies.

2. Create a helper context object `_opengl_context()` which `get_max_texture_sizes()` now uses. It uses vispy's `get_current_canvas()` to only create a Canvas if needed. This is the first key change: **do not create the invisible Canvas unless needed**.

3. `VispyBaseLayer` still has `MAX_TEXTURE_SIZE_*` variables because people might be depending on these.

4. Turn the hyper-single-purpose `KeyModifierFilterSceneCanvas` into `VispyCanvas` and move it into its own file `vispy_canvas.py`. The new `VispyCanvas` calls `get_max_texture_sizes()` to make sure that happens early while a Canvas exists. This is the second key change: **call `get_max_texture_sizes()` explicitly up front when we have a Canvas**. Later calls will just retrieve the same values because get_max_texture_sizes() is using an `lru_cache`.

The class name `KeyModifierFilterSceneCanvas` implied nothing else could be added to that class. Where now it's obvious we can push functionality from `QtViewer` into`VispyCanvas`. For example the implementation of `QtViewer.screenshot()` could change from:

        img = QImg2array(self.canvas.native.grabFramebuffer())
        if path is not None:
            imsave(path, img)  # scikit-image imsave method
        return img

to

        return self.canvas.screenshot(path)

or remove `QtViewer.screenshot` entirely and have people call `viewer.canvas.screenshot()` directly. There are a number of other things that could be pushed into `VispyCanvas`. Today `QtViewer` is pretty bloated.

# References

The creation of `VispyCanvas` was proposed as part of issue #855.

# Type of change

# How has this been tested?
- [x] Manual testing

## Final checklist:


sofroniewn (approved):
I love it!!! I'm really excited to move more off the `qt_viewer` and into the new canvas class. I've been long excited about #855. I will likely do some of that this weekend!

My only request for this PR is to rename `ViewerCanvas` to `VispyCanvas` and have it live in `vispy_canvas.py`. I think that name will be most consistent with things like `VispyCamera` etc.

@Czaki I don't think this change will break anything for you, but I want to give you aheads up too. I know we were just talking about deprecations and breaking changes on #1799 and I am trying to be mindful of them for you and our other users.

Czaki:
Looks ok.

pwinston:
@sofroniewn Yes `VispyCanvas` is a better name. I changed it.

sofroniewn:
Nice, this has now been updated and tests are passing. Will merge!

Issue #1327, opened 2020-06-02
Expand DockWidget API and add hookspec?


tlambert03:
## 🚀 Feature
In thinking about trying to integrate well-established community projects, like ilastik or deep lab cut, (or at least small modular parts of them)... we are, of course, a long way away from a full functional hook-based plugin architecture that would support all of their UX needs.

I'm wondering if we couldn't expand the dock widget API as a "loose" type of plugin that would immediately meet some needs, while we consider development of stricter hook-based plugins in parallel.  Specifically, when someone runs `viewer.window.add_dock_widget`, we pass back the `QtViewerDockWidget` object... and at the moment, it is a very thin wrapper a plain [`QDockWidget`](https://doc.qt.io/qt-5/qdockwidget.html).  However, it would be pretty easy to establish an API and write up some documentation that says "here's how you access the viewer that your widget is docked in"... and here's how you can use it, and perhaps inject some useful properties into their QWidget class.

We could define a new plugin hookspec that allowed plugins to declare these widgets, along with some metadata that would let us put a new item in the Plugins Menu, and let the end user select it and get that new dockwidget.

It's certainly not as "locked" down, and it does require more from developers (they'd need to pass a `QWidget`... or a function that we run `magicgui` on?), but I think we'd quickly get some very cool components (based on the many cool community dock widget examples we've seen on twitter and image.sc)

## Motivation

Letting advanced developers extend the napari gui with components that are well-beyond the current plugin design.

## Pitch

```python
# the spec

@napari_hook_specification
def napari_provide_dock_widget(): -> dict:
    """Return a dict that can be passed as kwargs to viewer.add_dock_widget"""

# the plugin

class MyWidget(QWidget):
    ...

    # these things would be true once added as a dock widget
    # we could maybe "inject" these properties for plugin developers too
    @property
    def dock_widget(self):
        return self.parent()

    @property
    def viewer(self):
        return self.get_dock_widget().parent()


@napari_hook_implementation
def napari_provide_dock_widget():
    return {
        'widget': MyWidget,
        'name': 'Cool Tracking Plugin',
        'area': 'right',
        'allowed_areas': ['right', 'left'],
        'shortcut': Alt+Shift-T,
    }

```

GenevieveBuckley:
This is exactly what I want to do for some projects of mine, so I'm happy to see things moving in this direction! Can I try working out an initial implementation?

jni:
@tlambert03 Love it! One modification I would suggest is:

```python
[...]

PR #1799, opened 2020-10-30
Deprecation: Make main window private

sofroniewn:
# Description
This PR makes the main window private and exposes `add_dock_widget` on the main viewer. I think we've had a lot of people getting value from the `add_dock_widget` method and so it make sense to expose that now on the Viewer directly. At the same time we've also clean up a lot of our internal models now (like camera, cursor, etc) that I don't think we need/ want users going into the main window anymore, which is still public at `viewer.window`.

I'm also thinking ahead to supporting a `provide_dock_widget` hookspec, as outlined in #1327, and having the `window` private but the `add_dock_widget` will make that much more natural.

I've added a deprecation warning so we can actually remove in `0.4.2`

I'm already working on #1327 so hopefully that will go out in `0.4.1` too, and the gain in functionality there will be well worth it.

Would be great to know from anyone whose been adding dock widgets/ manipulating the Qt window if they see any problems with this -  cc @Czaki @adamltyson @joshmoore @jni 

I also moved a string representation back to the viewer_model to keep the viewer file cleaner, and removed the deprecated update method - cc @AhmetCanSolak 

## Type of change

### DIFF
--- a/napari/_tests/test_viewer.py
+++ b/napari/_tests/test_viewer.py
@@ -1,5 +1,4 @@
 import os
-import time
 
 import numpy as np
 import pytest
@@ -116,44 +115,6 @@ def test_screenshot(make_test_viewer):
     # Take screenshot with the viewer included
     screenshot = viewer.screenshot(canvas_only=False)
     assert screenshot.ndim == 3
-
-
-def test_update(make_test_viewer):
-    data = np.random.random((512, 512))
-    viewer = make_test_viewer()
-    layer = viewer.add_image(data)
-
-    def layer_update(*, update_period, num_updates):
-        # number of times to update
-
-        for k in range(num_updates):
-            time.sleep(update_period)
-
-            dat = np.random.random((512, 512))
-            layer.data = dat
-
-            assert layer.data.all() == dat.all()
-            # if you're looking at this as an example,
-            # it would be best to put a yield statement here...
-            # but we're testing how it handles not having a yield statement
-
-    # NOTE: The closure approach used here has the potential to throw an error:
-    # "RuntimeError: Internal C++ object () already deleted."
-    # if an enclosed object (like the layer here) is deleted in the main thread
-    # and then subsequently called in the other thread.
-    # Previously this error would have been invisible (raised only in the other
-    # thread). But because this can make debugging hard, the new
-    # `create_worker` approach reraises thread errors in the main thread by
-    # default.  To make this test pass, we now need to explicitly use
-    # `_ignore_errors=True`, because the `layer.data = dat` line will throw an
-    # error when called after the main thread is closed.
-    with pytest.warns(DeprecationWarning):
-        viewer.update(
-            layer_update,
-            update_period=0.01,
-            num_updates=100,
-            _ignore_errors=True,
-        )
 
 
 def test_changing_theme(make_test_viewer):
--- a/napari/components/viewer_model.py
+++ b/napari/components/viewer_model.py
@@ -109,6 +109,10 @@ class ViewerModel(AddLayersMixin, KeymapHandler, KeymapProvider):
         self._mouse_drag_gen = {}
         self._mouse_wheel_gen = {}
 
+    def __str__(self):
+        """Simple string representation"""
+        return f'napari.Viewer: {self.title}'
+
     @property
     def palette(self):
         """dict of str: str : Color palette with which to style the viewer.
--- a/napari/viewer.py
+++ b/napari/viewer.py
@@ -204,6 +204,28 @@ class Viewer(ViewerModel):
             shortcut=shortcut,
         )
 
+    def remove_dock_widget(self, widget):
+        """Removes specified dock widget.
+
+        Parameters
+        ----------
+        widget : QWidget | str
+            If widget == 'all', all docked widgets will be removed.
+        """
+        self._window.remove_dock_widget(widget)
+
+    def resize(self, width, height):
+        """Resize the window.
+
+        Parameters
+        ----------
+        width : int
+            Width in logical pixels.
+        height : int
+            Height in logical pixels.
+        """
+        self._window.resize(width, height)
+
     def show(self):
         """Resize, show, and raise the viewer window."""
         self._window.show()
--- a/napari/viewer.py
+++ b/napari/viewer.py
@@ -219,7 +219,3 @@ class Viewer(ViewerModel):
             # https://github.com/napari/napari/issues/1500
             for layer in self.layers:
                 chunk_loader.on_layer_deleted(layer)
-
-    def __str__(self):
-        """Simple string representation"""
-        return f'napari.Viewer: {self.title}'
--- a/napari/viewer.py
+++ b/napari/viewer.py
@@ -113,7 +113,7 @@ class Viewer(ViewerModel):
     def window(self):
         warnings.warn(
             (
-                "The viewer.window parameter is deprecated and will be removed in version 0.4.2."
+                "The viewer.window parameter is deprecated and will be removed in version 0.5.0."
                 " Instead you should use the viewer.add_dock_widget method to extend the GUI."
                 " Directly manipulation of the window is no longer supported."
             ),
--- a/napari/viewer.py
+++ b/napari/viewer.py
@@ -1,14 +1,15 @@
 import platform
 import sys
+import warnings
 from os.path import dirname, join
 
 from qtpy.QtGui import QIcon
-from qtpy.QtWidgets import QApplication
+from qtpy.QtWidgets import QApplication, QWidget
 
 from . import __version__
 from ._qt.qt_main_window import Window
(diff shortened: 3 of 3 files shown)


Czaki:
I'm not enthusiastic about this change. We start talking about the plugin system. I'm not sure if every abstract model is properly exposed from the viewer instance. At least I suggest delaying removing to 0.4.4

joshmoore:
Thanks for the ping, @sofroniewn. Also pinging @will-moore

adamltyson:
For us, I think this is just changing one line of code, so fine by me.

+1 for hookspec

Czaki:
I now noticed that add_docked_widget is a new method. 
There was an idea to change the status to beta. Forcing of change code in during one relase smell like early alpha. I try to keep code compatybile with the older package version. 
But napari is really hard. Near all release force to add some if. Which is really annoing.

sofroniewn:
> I'm not enthusiastic about this change. We start talking about the plugin system. I'm not sure if every abstract model is properly exposed from the viewer instance. At least I suggest delaying removing to 0.4.4

We can delay the remove until later, it would be good though to start collecting cases where people would need to go into `_window`. Do you have any examples?

sofroniewn:
> I now noticed that add_dock_widget is a new method.

Yeah, we're passing through the `window.add_dock_widget` method to expose it to the user. In an upcoming PR I was thinking about allowing this method to take a plugin name. Unclear if we might want to expose the method then as `viewer.experimental.add_dock_widget` @jni 

> There was an idea to change the status to beta. Forcing of change code in during one relase smell like early alpha. I try to keep code compatybile with the older package version.
But napari is really hard. Near all release force to add some if. Which is really annoing.

Sorry about that @Czaki, we do appreciated the work you're putting in to keep your tool up to date and are trying to minimize breaking changes to the public API. I think making window private though will be an important step in the long run though as I think we want to add GUI elements via plugins and not directly manipulating the window

We're definitely still in alpha though, and realistically can see us in this phase for another year.

jni (approved):
I see both points here. @Czaki you did convince me on Zulip that we are in fact still alpha software, so these deprecations are to be expected.

Having said that, @sofroniewn, I don't see any reason why we wouldn't just leave the `.window` property there until 0.5? Doesn't seem like it would cause any real harm or maintenance burden? (If you agree, remember to change the warning message before merge.)

Other than that, I am +100 to putting `add_dock_widget` on the viewer. Thanks @sofroniewn!

sofroniewn:
I've update the deprecation warning to be 0.5.0, happy to have bundle big changes like this for major release.

We might want to think again about using something like towncrier https://github.com/twisted/towncrier to help tracking this stuff and surfacing it to the community. Not sure @jni if you've had more experience there, sounds like it was getting discussed on skimage at some point https://github.com/napari/napari/pull/546#pullrequestreview-295663111

sofroniewn:
@jni one further question actually - should we pass through `remove_dock_widget` onto of `viewer` too? https://github.com/napari/napari/blob/v0.4.0/napari/_qt/qt_main_window.py#L498-L510

We might also want https://github.com/napari/napari/blob/v0.4.0/napari/_qt/qt_main_window.py#L512-L522

Now I'm wondering if we should go the other direction - leave window as public, get rid of the passthrough's that we already have, but make sure everything we don't want exposed on window is now private - which is not the case (for example `QtViewer` there is public).

I still like making window private, and passing through `remove_dock_widget` and `resize` but I can see arguments against too.

It's really

```python
viewer.screenshot
viewer.close
viewer.add_dock_widget
viewer.remove_dock_widget
viewer.show
viewer.resize
viewer.update_console

# vs

viewer.window.screenshot
viewer.window.close
viewer.window.add_dock_widget
viewer.window.remove_dock_widget
viewer.window.show
viewer.window.resize
viewer.window.update_console
```

The only other public method on window right now is `activate`. The public attributes are `qt_viewer`, `main_menu`, `file_menu`, `view_menu`, `window_menu`, `plugins_menu` and `help_menu`. I don't mind having `activate` public or passing through to the viewer, though it could also be renamed `raise`. I'd probably make `qt_viewer` and all the `menu` attributes private if we did leave `viewer.window` exposed

jni:
I prefer to leave things on the viewer. Levels of indirection really hurt discoverability.

sofroniewn:
> I prefer to leave things on the viewer. Levels of indirection really hurt discoverability.

Ok, then I think I should probably move `remove_dock_widget` over too, no?

jni:
Yes, sorry, I left that implicit rather than explicit, against the Zen of Python's advice! :sweat_smile:

sofroniewn:
ok, moved over. thanks!! I think this is done. will let sit for a while in case anyone else has any further comments

sofroniewn:
I've pushed a super early look at a viewer dock widget hook spec in #1800, but we might also want to wait and see how that shapes up before merging this one

tlambert03:
> I'm not enthusiastic about this change.

+1 ... I'm also not enthusiastic about this direction. (edit: I'm not opposed to putting add_dock_widget on `viewer` ... but would not like to see `viewer.window` go private)

jni:
@tlambert03 what is it that you want to access on the window from the public API? Or is it just a principle thing that you think users should have access to it?

sofroniewn:
> I'm not enthusiastic about this change.
>
> +1 ... I'm also not enthusiastic about this direction. (edit: I'm not opposed to putting add_dock_widget on viewer ... but would not like to see viewer.window go private)

ok, good to know! glad we left this open for more discussion

how do people feel about the second half of https://github.com/napari/napari/pull/1799#issuecomment-719870810, in particular making `qt_viewer` and then menubar on window private on?

Alternatively I'm curious if we could do anything to make our instance of window hidden instead of on the viewer itself, maybe a global variable, like the qt app instance. That could actually let us bring together Viewer and ViewerModel back together again which would be nice. We'd have to take the screenshot off the viewer too, but we could make that separately importable like we have done for nbscreenshot.

```python
from napari.utils import screenshot
image = screenshot(viewer)
# and
from napari import Window
window = Window() # where this could be done at any time and returns the window if people want it

from napari import Viewer
viewer = Viewer() # note this will now return a viewer model, but a window was created in the background
```

Thoughts?

tlambert03:
> @tlambert03 what is it that you want to access on the window from the public API? Or is it just a principle thing that you think users should have access to it?

partly the former: I'm thinking about how many times on image.sc I've given someone a tip on how to do something slightly arcane where I needed to use `viewer.window`.  I recognize we can just point them to the private object and put it on our list of things to do to create an API for them...
partly the latter: especially the `QtMainWindow`, (I was actually hoping we'd eventually go the other direction with that, making that easier to access than viewer.window._qt_main_window... since that's such a stable object that has meaning outside of the napari environment).  this is an old debate though... lock it down vs open it up, etc... so I understand that there are good counter arguments.  but what is the impetus here?  Other than "I don't think we want people going in there?" ... has there ever been an issue with that?

Czaki:
> Alternatively I'm curious if we could do anything to make our instance of window hidden instead of on the viewer itself, maybe a global variable, like the qt app instance. That could actually let us bring together Viewer and ViewerModel back together again which would be nice. We'd have to take the screenshot off the viewer too, but we could make that separately importable like we have done for nbscreenshot.

It happens that I use multiple instances of napari viewer. (the result of the algorithm on different files).

joshmoore:
Sorry, all. Took me a bit to get my 🦆s  in a row. The example I mentioned on the euro dev sync was from https://github.com/ome/ome-zarr-py/pull/57#issuecomment-714770222:

Something like:
```
with napari.gui_qt():
    viewer = napari.Viewer()
    # blah blah blah
    for plate_layer in plate_layers:

        @plate_layer.mouse_drag_callbacks.append
        def get_connected_component_shape(layer, event):
```
currently requires shipping a script to users, rather than: `napari --plugin ome.plate_viewer` or even better, having that detected from the `.zarr`. I think that makes the overall goal "be able to bootstrap from a plugin" which perhaps means something that can take the viewer (or more than one, @Czaki :wink:)

On most of the other issues, I'm fairly agnostic especially at the 0.x level. It did occur to me though regarding the "experimental" proposal (which I generally like) that _eventually_ I could see something like this working more like `from future ...`. That would likely imply a longer support lifetime then I think the current `experimental` proposal would guarantee.

tlambert03:
fwiw: the last part of [this comment](https://forum.image.sc/t/magicgui-and-docked-widget-questions/44983/3?u=talley) on the image.sc forum (where someone asked how to resize a dock widget), is a good example of what I meant in https://github.com/napari/napari/pull/1799#issuecomment-720003203
there are lots of things that can still only be done programmatically by accessing the QMainWindow (or other QObjects) and until/unless we offer passthrough methods, I would hope to see napari make it easier, not harder, to get to them

sofroniewn:
There was not a lot of enthusiasm for this change so am closing, if other ideas emerge in the future they can be discussed elsewhere

Issue #1564, opened 2020-08-20
Use the `experimental` namespace to allow us to try out new features


sofroniewn:
## 🚀 Feature
There have been a couple of convos recently in different places in the project where we have wanted to get functionality in for advanced users and developers to try out, but where we might want to make breaking changes more frequently and ultimately might not end up supporting all the functionality that we try out. This could apply to hookspecs, like discussed in #1327, new layer types, new interactivity, or things we havn't even thought of yet. 

One suggestion from @freeman-lab that I really like is we learn from the tensorflow project who use the `experimental` namespace. See [tf.experimental](https://www.tensorflow.org/api_docs/python/tf/experimental?version=nightly) and [tf.data.experimental](https://www.tensorflow.org/api_docs/python/tf/data/experimental?version=nightly) which are clearly indicated as such in the docs.

As noted by @pwinston in discussion in the euro dev meeting, this would now give us a third namespace option, separate from public and the private `_`.

Scikit-image uses the `future` namespace (just one top level instead of individual at every sublevel), see [skimage.future](https://scikit-image.org/docs/stable/api/skimage.future.html) and with the following documentation 

> Functionality with an experimental API. Although you can count on the functions in this package being around in the future, the API may change with any version update and will not follow the skimage two-version deprecation path. Therefore, use the functions herein with care, and do not use them in production code that will depend on updated skimage versions.

I personally prefer the word `experimental` to `future` and I think it describes the napari needs a bit more that `future`, for example we might add an `experimental` hook spec that allows plugin developers lots of freedom but that actually then causes things to frequently crash. We'd then probably want to remove that hook spec entirely rather than preserve it.

These are just early thoughts right now, but curious the opinions of others and any ideas for how we might use this concept both with our plugins and more generally in napari. cc @jni @royerloic

kevinyamauchi:
I am a big fan of adding an "experimental" namespace. A few quick thoughts:
- I think the intention is much more clear than then `_` convention - especially to those that don't spend a lot of time with python.
- I think having an explicitly experimental section gives contributors a lot more freedom to implement and importantly get community feedback on new features much more quickly.
- I also think this ties into the discussions in #1556 , as features in the "experimental" namespace can have a lower bar for test coverage/quality, which again I think helps get things moving more quickly.

In terms of the actual name, I like "experimental", as I think it conveys the right things (i.e., these are things we are trying, they aren't totally solid, and they may change drastically in the future). Of course, I am open to other suggestions.

[...]

PR #1357, opened 2020-06-14
Integrated plugin dialog with install/uninstall & remote discovery

tlambert03:
# Description
As mentioned in https://github.com/napari/napari/pull/1289#issuecomment-640995785, I'm working on a more integrated plugin dialog:

Untitled

Features include:
- finds available plugins by naming convention on PyPI
- finds available plugins on github, even without naming convention (such as `ome-zarr`) based on these conditions (This was the best I could do without a PyPI trove classifier):
   - they have the literal string "napari plugin" in the readme
   - they have a `setup.py` or `setup.cfg` file that declares a "`napari.plugin`" entry_point
   - they are published on PyPI with development status Alpha or greater.
- pip install/uninstall in background process with stdout feedback in the GUI. (paths and environment variables preserved so this would work in a bundle)
- detects outdated plugins and offers update button.
- integrated plugin sorter
- integrated error report (plugins with errors show a clickable warning icon that slides out the report).

(The checkbox next to each plugin will be used to activate/deactivate a plugin without actually uninstalling it)

This requires some changes to napari-plugin-engine that are not yet merged, so if you want to test this locally, you'll need to checkout the [napari-plugin-engine/refactor-discovery](https://github.com/tlambert-forks/napari-plugin-engine/tree/refactor-discovery) branch

## Type of change

- [ ] Bug-fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# References

# How has this been tested?

- [ ] example: the test suite for my feature covers cases x, y, and z
- [ ] example: all tests pass with my change

## Final checklist:
- [ ] My PR is the minimum possible work for the desired functionality
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #1800, opened 2020-10-31
Viewer dock widget hook spec

### LINKED ISSUE #1327: Expand DockWidget API and add hookspec?

sofroniewn:
# Description
This is a very early WIP PR to add a viewer dock widget hook spec and close #1327. The exact API might depend on what we do with #1799. I'll update this as it goes.

## Type of change


sofroniewn:
Ok, this PR is beginning to have some of the key functionality that people might want to see
- (1) A way for plugins to provide a list of dock widgets
- (2) Access to the viewer through those dock widgets
- (3) A way for the user to turn on/ off dock widgets from the API and GUI

All of these are in a pretty early state, and could change and be improved during the course of this PR, but there is now at least a version of all of them

For (1) we have the `napari_experimental_provide_dock_widget` which returns a list of tuples where the first thing in the tuple is a QWidget, and the second is a dictionary of keyword arguments to `window.add_dock_widget`. This seems like a pretty reasonable place to start, but one thing to note is that the kwargs include a keyboard shortcut, which we may or may not want to let plugins do

For (2) we pass the full viewer instance to the `napari_experimental_provide_dock_widget` hookimpl as an arg. This isn't just the ViewerModel it's the full on Viewer so plugins would have access to everything, the window etc and they could add and delete things as they wish. There's been a lot of discussion about this in other place, including using shims, and I don't see this as the finished version, but just an initial attempt that gives full access.

For (3) I've added a drop down menu that gets populated with all the plugins that have this hookimpl and can be used to add and remove the corresponding dock widgets. We might instead want to leverage some of our plugin list/ sorter options, such as present in #1357, but this was quick to do right now. Note after adding a dock widget you can also hide/ show it from the window menu, which is different from removing it, which you have to do back in that plugins dropdown menu.

I've put the word `experimental` inside the hookimpl and some, but not all, of the code in an `experimental` folder. While there's some stuff to indicate the experimental status to the developer, there isn't anything right now to indicate that to the user. I'm very open to ideas here too. See #1564 for some more context.

I also am not using the `historic` arg right now - see comment here https://github.com/napari/napari/issues/1327#issuecomment-638086212, but I'm not sure how that applies if passing the viewer.

Anyway, I consider this a super super early prototype, and there's lots to consider, but I know there are a lot of people in the community excited about this one, so wanted to give it a push along.

Here's a gif of adding/ removing a plugin and then hiding/ showing it too. Sorry the menu bar is getting cut off in my gifs. Looking forward to working on this more!!

plugin_dock_widget

Czaki:
Maybe there should be the third value of tuple. A bool indicating if the plugin is activated by default? 
There are some rare cases for which we would like to have some widget but there is no need to have it visible every napari launch?

sofroniewn:
> Maybe there should be the third value of tuple. A bool indicating if the plugin is activated by default?
There are some rare cases for which we would like to have some widget but there is no need to have it visible every napari launch?

I'm imagining a future where someone might have a hundred napari plugins installed, each providing its own dock widgets, and in that case I don't think you'd want plugins to be able to force themselves to always be visible. Once we have persistent settings in the app we can make it so that if remembers which you had open in the last session.

GenevieveBuckley:
I am very excited about this, thanks @sofroniewn!

I would be willing to help work on a cookiecutter for plugins using the dockwidget hook spec here, similar to the cookiecutter we have for image reader plugins. This capability would be something we would use in our lab.

tlambert03:
> I would be willing to help work on a cookiecutter for plugins using the dockwidget hook spec here, similar to the cookiecutter we have for image reader plugins. This capability would be something we would use in our lab.

I feel like creating multiple cookiecutter repos might be confusing.  That cookiecutter is much more about "scaffolding" than it is about creating an io plugin specifically.  (The io plugin was just an example).  I think you should put _all_ example hook implementations in that one repo, then a user can use it and delete whatever examples they don't want.

tlambert03:
> For (2) we pass the full viewer instance to the napari_experimental_provide_dock_widget hookimpl as an arg. 

@sofroniewn, can you show an example implementation that uses this?
Did you already consider and reject the idea of injecting the viewer into the class as mentioned in #1327 rather than passing it as an argument to the hook implementation?

sofroniewn:
> @sofroniewn, can you show an example implementation that uses this?

```python
@napari_hook_implementation
def napari_experimental_provide_dock_widget(viewer):
    return [(QLabel('hello ' + str(viewer)), {})]
```
This is all I did for the example above, which is no where near anything functional or useful!!!

> Did you already consider and reject the idea of injecting the viewer into the class as mentioned in #1327 rather than passing it as an argument to the hook implementation?

It's definitely not rejected. It seemed very simple just to pass the viewer like this and strictly more general so I thought "why not", but I'd love to know more about why you like the injecting the viewer idea. Apologies if I'm missing something really basic or obvious, I may well be!!

Czaki:
There are any plans to support modifying menu position (like in ImageJ)? Are you thinking that it is a good idea?

tlambert03:
> This is all I did for the example above, 

ah I see... yeah I wasn't intending for the plugin developer to instantiate the widget, but rather to hand us the class itself that _we_ would instantiate when necessary. (so, `napari_provide_dock_widget(): -> Tuple[Type[QWidget], dict]` instead of `napari_provide_dock_widget(viewer): -> Tuple[QWidget, dict]` as mentioned in the earlier thread). One benefit would be that a single plugin could provide multiple widgets without having to instantiate all of them when the hook is called.  Another big benefit is that you can simply call this hook once at import (e.g. "what widgets do you have a available for me?") and you don't need to call it every time you add a new instance of the widget to the viewer, as currently done in `add_plugin_dock_widgets`.  (This maybe also explains why you avoided the `historic` hookspec?)

This has the (nice) effect of leaving the viewer out of the hook signature, but then of course requires that we let them know how to get a hold of it.  I was thinking of one of two options:

1) pass `viewer` to the QWidget upon instantiation: this is nice and straightforward, but basically requires that the QWidget be specifically designed for napari (or that it at least has a catch-all `**kwargs` in the init function)... which is not so nice.
2) do what I was calling "injecting" the viewer, where we just add the viewer as an attribute to the class instance.  This would be a "contract" with the plugin developer that would allow them to access (e.g.) `self.napari_viewer` anywhere in their class, and get whatever viewer instance their widget was embedded in.  (Note: this is basically the same thing that a savvy developer could already do just by traversing the parent tree of their widget after it was added as a dock widget).  The benefit here is that a) it gives them the viewer where they would most likely be wanting to use it anyway... and b) it "de-emphasizes" the transfer of the viewer.  Those who don't need to access it don't need to care about it at all: they don't need it in their hook implementation or their class... it's just invisible to them (unless they want it).

tlambert03:
another thought: if you have them pass a class instead of an instance, you could even get rid of the `dict` in the return signature, and have a (rather django-like) convention of class attributes that would (optionally) provide things like menu names, shortcuts, position in the menu, etc...

```python
@napari_hook_specification(historic=True)
def napari_experimental_provide_dock_widget() -> Union[Type[QWidget], Sequence[Type[QWidget]]]:
    ...


# the plugin

class MyWidget(QWidget):
    napari_shortcut = 'cmd-shift-u'
    napari_menu_name = 'My Special Widget'

@napari_hook_implementation
def napari_experimental_provide_dock_widget():
    return MyWidget # or a list of classes if providing multiple

```

GenevieveBuckley:
> I feel like creating multiple cookiecutter repos might be confusing. That cookiecutter is much more about "scaffolding" than it is about creating an io plugin specifically. (The io plugin was just an example). I think you should put _all_ example hook implementations in that one repo, then a user can use it and delete whatever examples they don't want.

Good points, this sounds like a good approach

sofroniewn:
> but rather to hand us the class itself that we would instantiate when necessary

Ah yes I see, I like this idea, seems much cleaner, and goes well with `historic`.

> but then of course requires that we let them know how to get a hold of it. I was thinking of one of two options

Personally I lean towards the first option - passing viewer to the constructor. If I am building for napari I might find it a little simpler to understand and more explicit than an injection. It might also make it easier for me to be testing things or using my widget independent of the plugin ecosystem. I could imagine for example developing a widget in a jupyter notebook or script for example where I have a constructor take the viewer and then manually add the dock widget myself and then wanting to see how it works. It might be a bit harder for me to replicate the whole injection procedure. Curious if this reasoning makes sense/ how others feel?

> another thought: if you have them pass a class instead of an instance, you could even get rid of the dict in the return signature, and have a (rather django-like) convention of class attributes that would (optionally) provide things like menu names, shortcuts, position in the menu, etc...

ooo I like not having to return the dictionary as well, that seems much cleaner to avoid the whole tuple thing. Class attributes make sense, but just out of curiosity why wouldn't this be possible if they return an instance too?

All in all though, I like how this convo is going

tlambert03:
> It might also make it easier for me to be testing things or using my widget independent of the plugin ecosystem.

This was actually the exact reason why I was suggesting _not_ passing the viewer in the constructor.  There will certainly be many existing `QWidgets` out there that could make for interesting dock widgets that don't actually need access to the napari `viewer` object to function.  By using the constructor approach, we _force_ all of those classes to modify their `__init__` functions to accept a viewer (or at least, as I mentioned, to add a catch-all `**kwargs`).  In either case, their basic widget won't work for napari unless they modify it.

With what we're calling injection, their class is identical as a napari dock widget and independent of the napari ecosystem.  For those developers who _do_ want to access the full viewer object that their widget is embedded into, they don't need to do anything extra they can just directly access it in their class as (e.g.) `self.napari_viewer`...

```python
class MyWidget(QWidget):
    def do_something(self):
        print(self.napari_viewer)  # we (napari) will make this work for them
```

> It might be a bit harder for me to replicate the whole injection procedure.

Are you imagining that the injection procedure is something that the plugin developer does?  They don't need to do anything... `self.napari_viewer` just "magically" appears for them (though, to be safest, they should probably use `if hasattr(self, 'napari_viewer')` to check if their widget is being run inside a napari viewer... or just let it raise an error if their widget _never_ makes sense outside of napari). On our side, we would simply add the viewer attribute to the widget instance when we add it as a dock widget.

tlambert03:
> Class attributes make sense, but just out of curiosity why wouldn't this be possible if they return an instance too?

no, you're right... we could use that approach in either case.

Czaki:
> This was actually the exact reason why I was suggesting _not_ passing the viewer in the constructor. There will certainly be many existing `QWidgets` out there that could make for interesting dock widgets that don't actually need access to the napari `viewer` object to function. By using the constructor approach, we _force_ all of those classes to modify their `__init__` functions to accept a viewer (or at least, as I mentioned, to add a catch-all `**kwargs`). In either case, their basic widget won't work for napari unless they modify it.

Could you propose some example of a useful widget which does not need access to viewer? I try to imagine one.

tlambert03:
Any magicgui widget, for example.  Or anything that doesn't actively need to add a layer to the viewer

Czaki:
> Any magicgui widget, for example. Or anything that doesn't actively need to add a layer to the viewer

And what usefull such widget could do? There is any method of interaction with viewer without viewer object? Of course I could imagine adding clock, calculator etc. But i do not see how such widgets may extends napari viewer possibilities.

sofroniewn:
> Are you imagining that the injection procedure is something that the plugin developer does? They don't need to do anything...

No, I get that's what happens when they have a plugin, I was imagining a developer prototyping something in a notebook or a script and not going through the plugin interface (as they don't have a dedicated repo yet etc) and they just have a dock widget they want to add manually through `window.add_dock_widget`.

One additional question I have here is how are we imagining dealing with PySide2 vs PyQt5 widgets? If a plugin developer has decided to use one of these packages and provided a plugin in only that framework, so not qtpy, and I'm running napari with the other, it won't work right? Is there a way to enforce we only accept things written with qtpy? Apologies if this was discussed in #1327 and I've forgotten.

tlambert03:
> And what usefull such widget could do? There is any method of interaction with viewer without viewer object? 

Yes! a magicgui widget can add things to the viewer simply by nature of its annotations.  It _doesn't need to accept the viewer_ to be able to interact with the viewer.  In that example, there is an existing API that allows them to interact with the viewer.  If we force the class to accept the viewer, then someone would need to effectively redefine their `__init__` function _after_ using the magicgui decorator (just to accept a throwaway viewer argument).  quite ugly. (I'm aware there's a separate magicgui proposal, but that's a different discussion).  Outside of magicgui usage: If you want an example of a widget that someone might use in napari that don't need to modify layers, imagine "dashboard" like things: perhaps something that gives them the status of some folders or some experiment on a remote drive.

I guess I would put the question the other way.  When this discussion started, people were extremely hesitant to pass the viewer at all, why now make the viewer a _non-optional_ part of the spec?  The optional viewer approach seems to solve all problems without requiring developers to make special accommodations for napari unless they want to (which is something we've generally tried to do, right?).

tlambert03:
> One additional question I have here is how are we imagining dealing with PySide2 vs PyQt5 widgets? If a plugin developer has decided to use one of these packages and provided a plugin in only that framework, so not qtpy, and I'm running napari with the other, it won't work right?

the "cleanest" answer is, correct: if a developer wants their end-users to be able to use both frameworks, then they should import from qtpy.  If a plugin has code that states `from PyQt5.QtWidgets import ... ` and the end user doesn't have PyQt5 in their environment it would fail.

the more interesting answer is that we could try to import plugin modules with a modified namespace mapping anything in `{PySide2, PyQt5}.{QtWidgets, QtCore, QtGui}` to their qtpy equivalents.  take, for example, the following module:
```python
# sample.py
from PySide2.QtWidgets import QWidget

def test():
    print(QWidget)
```

If you don't have PySide2 installed, it will obviously fail:
```python
In [1]: import sample
ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-1-c4ef148c3e21> in <module>
----> 1 import sample

~/Dropbox (HMS)/Python/forks/napari/sample.py in <module>
----> 1 from PySide2.QtWidgets import QWidget
      2
      3
      4 def test():
      5     print(QWidget)

ModuleNotFoundError: No module named 'PySide2'
```

However, if you tell python that it should interpret `PySide2` as `qtpy` by modifying sys.modules before importing, then it works:
```python
In [2]: import sys, qtpy

In [3]: sys.modules['PySide2'] = qtpy

In [4]: import sample

In [5]: sample.test()
<class 'PyQt5.QtWidgets.QWidget'>
```

(and we could put that temporary sys.modules modification in a context manager to make sure to undo it after importing).  That said, I definitely think we should start by encouraging plugin devs to use qtpy, and only use this as a fallback if the use case pops up

Czaki:
> One additional question I have here is how are we imagining dealing with PySide2 vs PyQt5 widgets? If a plugin developer has decided to use one of these packages and provided a plugin in only that framework, so not qtpy, and I'm running napari with the other, it won't work right? Is there a way to enforce we only accept things written with qtpy? Apologies if this was discussed in #1327 and I've forgotten.

For bundled version you could prepare dummy PyQt5 package which will transfer import from PyQt5 to import from qtpy. 
I created code which do similar thing:

https://github.com/4DNucleome/PartSeg/blob/master/package/PartSeg/utils/__init__.py

If You think think that it is good approach and something in this code is not clear I could explain.
Solution proposed by @tlambert03 will not work because of `pyqtSignal` and `pyqtSlot` which is possible to handle using my approach. 

But in general I agree with @tlambert03 that plugin creators should use `qtpy`.

tlambert03:
> If You think think that it is good approach and something in this code is not clear I could explain.

interesting. from looking at your code, it looks like you're also saying "wherever you wanted to import name X, import from `PartSegCore` instead".  In mine, I was saying "wherever you want PySide2, use qtpy instead" (which, at first glance, i thought was the same).  Seems like they would _both_ fail if someone tried to import the signal object such as `from PyQt5.QtCore import pyqtSignal`(since that import doesn't exist in qtpy), no?  Would definitely be curious to see a small example showing how you've solved that.

Czaki:
> Yes! a magicgui widget can add things to the viewer simply by nature of its annotations. It _doesn't need to accept the viewer_ to be able to interact with the viewer ...

I do not understand. There are two options. There is some global variable witch holds reference to viewer, which is bad if user would use multiple instances. Or viewer is somehow passed to widget to have reference. 

If there is need to have widget without napari reference there could be used subclass:
```python
class WrapClass(TrueClass):
    def __init__(self, viewer):
		super().__init__()
		self.link_viewer(viewer)
```

We could also use [`inspect.signature`](https://docs.python.org/3/library/inspect.html#inspect.signature) to check if there is viewer argument. 

For me writing code in which I need check every time if `viewer` is connected add additional, obsolete work.

tlambert03:
> I do not understand. There are two options. There is some global variable witch holds reference to viewer, which is bad if user would use multiple instances. Or viewer is somehow passed to widget to have reference.

have a look at the magicgui examples: https://magicgui.readthedocs.io/en/latest/examples/napari_img_math/
note the `image_arithmetic` function, doesn't take viewer anywhere yet adds its result to the viewer.  If you're curious, the logic for finding the current viewer (no global instance is used) is defined in [`napari.utils._magicgui.find_viewer_ancestor `](https://github.com/napari/napari/blob/ffcd827096371ac3693408379aee0621036e6c59/napari/utils/_magicgui.py#L45)

(edit: if you're not familiar with magicgui, this works because the return annotation is a napari layer type)

Czaki:
@tlambert03 

Ok. I see. My bad. 
My code allows printing warning messages. 

And for signal there should be something like 

```
from qtpy import QtCore

QtCore.pyqtSignal = QtCore.Signal
QtCore.pyqtSlot = QtCore.Slot
```

tlambert03:
> And for signal there should be something like
> QtCore.pyqtSignal = QtCore.Signal

yeah, agreed.  both solutions work fine if you redirect all the necessary names.

Czaki:
> have a look at the magicgui examples: [magicgui.readthedocs.io/en/latest/examples/napari_img_math](https://magicgui.readthedocs.io/en/latest/examples/napari_img_math/)

I see this line
```
viewer.layers.events.changed.connect(lambda x: gui.refresh_choices())
```

Which is a connection to the viewer which needs to be provided by the author of the plugin. Which need to be handled by constructor or method of returned Widget.

tlambert03:
> I see this line...

did you miss the whole other part wherein the widget adds a new image to the viewer without a reference to the viewer??

sure... one can _also_ manually make some additional connections (in the case you pointed out, we're making sure that our widget gets notified of new layers that get added... and in fact it's gone in newer versions of magicui) but you could totally remove that line and the core functionality of adding an image to the viewer would still be there, no reference to the viewer.  that was my point... did you see it?

Czaki:
> sure... one can _also_ manually make some additional connections (in the case you pointed out, we're making sure that our widget gets notified of new layers that get added... and in fact it's gone in newer versions of magicui) but you could totally remove that line and the core functionality of adding an image to the viewer would still be there, no reference to the viewer. that was my point... did you see it?

It looks that I should read part of magicui to understand how it works. But if I good understand its use 
https://github.com/napari/napari/blob/ffcd827096371ac3693408379aee0621036e6c59/napari/utils/_magicgui.py#L45 function to find viewer (this function should also contain check of type of `qt_viewer`).

tlambert03:
the "trick" here is more in that `napari.utils._magicgui` module that I linked above than it is in magicgui, but it essentially boils down to this: because napari is using `magicgui.register_type`, if a magicgui function is annotated with a napari layer type as its return value, then napari will take that return value and add it to the viewer (or update the layer if a name is provided).  It finds the viewer by navigating the qt widget parent tree.  the updating happens in [`show_layer_result`](https://github.com/napari/napari/blob/ffcd827096371ac3693408379aee0621036e6c59/napari/utils/_magicgui.py#L114).

but we're really getting way off the point here: this was all just a response to the question "can a `QWidget` ever be useful without explicitly taking a viewer?" and this says "yes ... sometimes".  sometimes because of alternative napari APIs (such as in magicgui), and sometimes because you might just want a widget to be visible, adding contextual information to your "napari dashboard", even if it's not specifically adding a layer.

lastly:
> For me writing code in which I need check every time if viewer is connected add additional, obsolete work.

certainly you wouldn't need to write it _every_ time.  just as you could wrap a class with a modified `__init__` method as you demonstrated above, you could write a property:

```python
@property
def viewer():
    try:
        return self.napari_viewer
	except AttributeError:
        print("this widget is not currently a docked widget")
```

Czaki:
I try this example. 

1. It is intentional that it overwrites the previous layer? For me, it is not intentional that I need to rename layer manually. 
2. Rename of the layer does not trigger an update of the dropdown list (old name visible until adding or remove layer). 

At this moment I'm confused. Only simple widgets could be created using `magicui`. `find_viewer_ancestor` is a function from a private module. 

So either `find_viewer_ancestor` should be public (and hardened) or there should be some advanced logic on side of initialization for viewer which will check the signature of the constructor, the existence of some method, etc to call proper set viewer method? 

Does first option also suggest to use a layzy attribute to wait until widget is docked?

tlambert03:
let's talk magicgui elsewhere... and keep this thread about the viewer dock widget hook API.  ping me if you'd like to discuss magicgui

Czaki:
Only first three lines are about magicui. others are about hook API.

jni:
Ok I'm finally caught up on all the discussion! :sweat_smile: 

So I think I agree philosophically with @tlambert03 that we should not *require* the viewer. @Czaki is right that this limits how much you can do with the widget, but as @tlambert03 pointed out there's still some useful ones, e.g. an OMERO browser or s3 bucket browser or a napari configuration editor or a cluster connection manager or whatever. I'm very much wedded to the idea that people should have to do as little work as possible to make their widgets/functions into napari plugins. We want to feed the impulse that made @cgohlke briefly include a napari plugin within tifffile itself! :joy: So I definitely think the viewer should be optional.

I'm super allergic to the viewer injection option, though. That requires people wanting to use the viewer to write code with an attribute that doesn't exist anywhere else in their code base. If I went in blind and tried to understand what their code is doing, I would be super super confused. (Not to mention PyCharm et al would be confused too!)

I think Python's code introspection capabilities are good enough that we can *check* whether we can pass in the viewer to a dockwidget plugin, and if so, we do it, if not, we don't. Assuming we go with the class return (rather than the instantiated object return), then we can check the `__init__` for:
- does it have a `viewer=` argument? If so, pass the viewer here.
- if not, does it have a `**kwargs`? If so, pass the viewer here.
- if not, does it have a *single*, *required* positional argument? (e.g. `def __init__(self, v):` If so, pass the viewer here. (I could take or leave this last one — I don't think it is too much to ask that if people actually want the viewer, they should call it viewer. But, it's an option.)

I think this should satisfy everyone's requirements?

tlambert03:
> I think Python's code introspection capabilities are good enough that we can check whether we can pass in the viewer to a dockwidget plugin, and if so, we do it, if not, we don't.

Yep, I like that (though maybe it should be "napari_viewer" or should require a type annotation)?

(Don't love the single required positional arg though, I think it should be keyworded)

sofroniewn:
> Yep, I like that (though maybe it should be "napari_viewer" or should require a type annotation)?

Can we provide that type annotation via an import from something like `napari_plugin_engine.types`? I.e. I don't want to force a dependency on `napari` for that.

> (Don't love the single required positional arg though, I think it should be keyworded)

I kinda like the single required positional arg for convenience, but it's certainly not something I need, and something we can always add later if more people want it.

I can also live with `napari_viewer=` as the keyword too.

Sounds like we're pretty close to consensus here which is very exciting!! Thanks for all the great input @tlambert03 @Czaki and @jni - I think a lot of people are going to build a lot of cool stuff using this API!!

tlambert03:
>  Can we provide that type annotation via an import from something like napari_plugin_engine.types? I.e. I don't want to force a dependency on napari for that.

they can use a [Forward reference](https://www.python.org/dev/peps/pep-0484/#forward-references) (just a string) to avoid importing/depending on napari:

```python
class MyWidget(QWidget):
    def __init__(self, viewer: 'napari.viewer.Viewer'):
```

sofroniewn:
> they can use a Forward reference (just a string) to avoid importing/depending on napari:

Ah thanks!! I've always wondered if that was possible, that's great to know. I could probably do that here too https://github.com/napari/napari/pull/1816#issue-513771124 I guess too!!

@tlambert03 if you were available and willing, I'd certainly appreciate it if you took over this PR and implemented the hookspec with signature inspection that @jni described here https://github.com/napari/napari/pull/1800#issuecomment-721504687 and the proper use of `historic` etc.

I'm also curious of your opinion on the dropdown menu for turning the dock-widget on/ off. We could keep that, or integrate something with the dialog from #1357 or the hook spec sorter, or both? I don't think there's a need to get anything perfect there in this PR, but there should at least be something

Czaki:
Some times ago there was a suggestion that it will be nice to have the possibility to implement PartSeg as a plugin. I think a little about this and my current conclusion is that the current proposal does not allow to do this. The mase of the probem is that storing all PartSeg option in one widget will create an illegible interface. And when split on more there is a need to synchronize state beetwen them. As I aexpect using `partial` from `functols` will break `inspect` module. Using global state is also a bad idea. 

But I may also be focused on one apprach and do not see the other one (using special oobject which share states and pipes signals). 

Base on previous talks I think that make possible create a more complicated application as plugin to napari is an important task. I'm not sure if this should be possible ussing the currently discussed hook spec, or create a new one? 

Maybe this thing should be discussed during some meetings?

jni:
Hi @Czaki,

> Base on previous talks I think that make possible create a more complicated application as plugin to napari is an important task. I'm not sure if this should be possible using the currently discussed hook spec, or create a new one?

Could you create a new issue to discuss this, titled something like "Coordination between multiple plugins or plugin widgets", or similar? The Viewer Dock Widget won't solve all developer needs, nor is it expected to, but it solves some and we shouldn't hold it up while discussing more complex plugin development.

tlambert03:
yeah, I'll second @jni's sentiments: this is definitely not intended to be the end-all-be-all way to create a fully-fledged "sub-application" within napari.  (see the first paragraph in #1327 that basically proposed this as a "stop-gap" that would enable some developers while we continue to expand our offerings).

That said... I _can_ imagine ways in which a developer could keep track of all of their classes that had been instantiated as dock widgets, and even coordinate behavior between them using nothing more than this spec (using something like the "special object which shares states and pipes signals" that you mentioned).  So when you create a new issue, please also include some specific examples of what you'd like to be able to achieve.

edit:

> The mase of the problem is that storing all PartSeg option in one widget will create an illegible interface

also just want to point out that once you've got a widget in the viewer (and a reference to that viewer), remember that you can have buttons etc _in that widget_ that can open other windows or widgets.  So one needn't cram the entirety of their application within a single widget to still be able to get to all of the functionality.

tlambert03:
question @sofroniewn: currently, the code here completely deletes the widget when it is toggled off (so a user would not be able to close a widget and reopen it without losing whatever previous work they'd done in that widget).  Is that by design?

sofroniewn:
> question @sofroniewn: currently, the code here completely deletes the widget when it is toggled off (so a user would not be able to close a widget and reopen it without losing whatever previous work they'd done in that widget). Is that by design?

It is, but it's maybe a bit clunky right now. There's a separation between turning on and off the widget (which is a creative/ destructive) process right now and happens through the plugin menubar and showing and hiding the widget (which preserves the previous state) which happens through the window menubar or the little x button in the top corner of the widget.

It maybe that having both of these options is confusing, and we just want to use the plugin menubar but then make turning off equivalent to hiding. I'm fairly open here, but the above was the initial thought process.

dongyaoli:
Thank you for starting this. After looking at this and #1849 , my two cents as a user:

1. I like how viewer is passed to plugin here. My opinion is that it's inevitable to expose the viewer to more sophisticated plugins. As a user and plugin developer I am ok to maintain my custom plugin to adhere to any changes made to the Napari core viewer.
2. I like that in this implementation, the widget is passed as an object instead of a class. I feel like it makes more sense this way and we can separate the instantiation of the widget object with Napari.
3. The current implementation will use module name as the plugin name. I personally prefer approaches similar to #1849 . My intended use case is: we will have a single repository which contains various image processing algorithms. For each algorithm, we would like to enable a Napari plugin (of course the algorithm is not aware of the plugin). So we would need separate name for each plugin. I think a proper way to do this is to have multiple plugins under the same module. I am wondering if the approach in #1849 is the right way to do this.

Thank you and I look forward to more development in this front!

dongyaoli10x:
Thank you for starting this. After looking at this and #1849 , my two cents as a user:

1. I like how viewer is passed to plugin here. My opinion is that it's inevitable to expose the viewer to more sophisticated plugins. As a user and plugin developer I am ok to maintain my custom plugin to adhere to any changes made to the Napari core viewer.
2. I like that in this implementation, the widget is passed as an object instead of a class. I feel like it makes more sense this way and we can separate the instantiation of the widget object with Napari.
3. The current implementation will use module name as the plugin name. I personally prefer approaches similar to #1849 . My intended use case is: we will have a single repository which contains various image processing algorithms. For each algorithm, we would like to enable a Napari plugin (of course the algorithm is not aware of the plugin). So we would need separate name for each plugin. I think a proper way to do this is to have multiple plugins under a QMenu named by the module. I am wondering if the implementation in #1849 is the right way to achieve this. If so, in my fork I will try to implement this.

Thank you and I look forward to more development about dock widgets.

dongyaoli10x:
> 3. ... So we would need separate name for each plugin. I think a proper way to do this is to have multiple plugins under a QMenu named by the module. ...

I think I found the answer to my own question. I figured out that using entry_point in setup.py can specify different plugin name for different dock_widget within a single module, although this won't necessarily collect them under the same menu named by the module.

I did find a small bug. In the dynamic connection of action of the plugin menu in `qt_main_window.py`, the original code at line 438 is: 
```
for imp in hook_caller.get_hookimpls():
    name = imp.plugin_name
    print("add action", name)
    action = QAction(name, parent=self._qt_window, checkable=True, checked=False)
    action.triggered.connect(lambda s : self._toggle_plugin_dock_widget(name, s))
    self._plugin_dock_widget_menu.addAction(action)
```
the `lambda` function line needs to be changed to:
```
action.triggered.connect(lambda s, name=name: self._toggle_plugin_dock_widget(name, s))
```
This is due to the late binding closure. If we don't specify `name`, then it will always be the last plugin name that is added.

I also read here [here](https://stackoverflow.com/a/48501804/3996552) that `lambda` function as a slot could cause widget not being garbage collected. I am not exactly sure but here it probably only affects the main window so we don't care about it.

sofroniewn:
This is being superseded by #1849

sofroniewn:
> I did find a small bug. In the dynamic connection of action of the plugin menu in qt_main_window.py, the original code at line 438 is:

Yes, thanks for catching that @dongyaoli10x, you're right! Much appreciated.

>  I personally prefer approaches similar to #1849 .

I'm glad you prefer that option, I think we will go with something more like that, which is why I just closed this one.

PR #1804, opened 2020-10-31
unpin qt, exclude 5.15.0

### LINKED ISSUE #1312: Pyside 5.15.0 breaking buttons?

tlambert03:
# Description
unpins qt versions by just excluding 5.15.0 (which was a bug in Qt ... not our end)
closes #1312

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -69,11 +69,11 @@ install_requires =
 
 [options.extras_require]
 pyside2 = 
-    PySide2>=5.12.3,<5.15.0
+    PySide2>=5.12.3,!=5.15.0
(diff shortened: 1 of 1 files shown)


sofroniewn (approved):
Looks good to me! Thanks @tlambert03 will merge now

PR #1808, opened 2020-11-01
Simplify viewer and make headless mode easier

sofroniewn:
# Description
This PR is a mostly backwards compatible simplification to our `Viewer` that makes working with napari headless much easier.

The actual API of our instantiated objects doesn't change, but I did change `Window` to now take a `ViewerModel` rather than `QtViewer` (so you can run `Window(viewer)` instead of `Window(QtViewer(viewer))` which is much nicer.

The `Viewer` file has now become much simpler, all the checking of the Qt application, logo setting etc happens inside `Window` which is where it is required. I've left window as public on `Viewer`, so this is unrelated #1799.

The true power of this PR though, is that you can now do much more with the ViewerModel headless and then look at it in a window much later.

For example the following is now possible

```python
# Cell 1
import numpy as np
from napari.components import ViewerModel as Viewer
viewer = Viewer()
viewer.add_image(np.random.random((20, 20)), colormap='red')
viewer.add_points(20*np.random.random((12, 2)), face_color=[0, .2, .5], size=1)
viewer.dims.ndisplay=3

# Cell 2
%gui qt

# Cell 3
from napari._qt import Window
window = Window(viewer)
```
And then you can see the viewer model!!!

You don't end up in an identical place as if you had just done
```python
from napari import Viewer
viewer = Viewer()
viewer.add_image(np.random.random((20, 20)), colormap='red')
viewer.add_points(20*np.random.random((12, 2)), face_color=[0, .2, .5], size=1)
viewer.dims.ndisplay=3
```
As in the first approach you are missing the `viewer.window` property, the methods that get added in `Viewer`, and the keybindings which get added in the main init.

I'd like to continue to move these together, but that might require some more significant changes and so want to do it gradually.

## Type of change

- [x] Refactor

### DIFF
--- a/napari/_qt/__init__.py
+++ b/napari/_qt/__init__.py
@@ -1,5 +1,6 @@
 from ..resources import import_resources
 from .event_loop import gui_qt
+from .qt_main_window import Window
 from .widgets.qt_range_slider import QHRangeSlider, QVRangeSlider
 
 import_resources()
--- a/napari/_qt/qt_main_window.py
+++ b/napari/_qt/qt_main_window.py
@@ -3,6 +3,8 @@ Custom Qt widgets that serve as native objects that the public-facing elements
 wrap.
 """
 import os
+import platform
+import sys
 import time
 
 from qtpy.QtCore import Qt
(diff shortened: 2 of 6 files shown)


sofroniewn:
Just caught a very simple path bug with the perfmon test, fixed now. cc @pwinston 

@Czaki you might be interested in this too, but I don't think it should affect you. This is ready for review @jni. 

I'm working on moving around the keybindings in #1810 but running into some problems there with inheritance and our keybinding system

jni (approved):

jni:
Awesome thank you @sofroniewn! 

If anyone has issues with this please let us know, we can always revert before the next release!

PR #1809, opened 2020-11-01
Fix TransformChain.__init__()

pwinston:
# Description
This PR makes `TransformChain.__init__()` call `Transform.__init__()`. The PR adds a test that shows this was not happening before. The test now passes.

## Type of change

# References
* [Understanding Python MRO - Class search path](https://makina-corpus.com/blog/metier/2014/python-tutorial-understanding-python-mro-class-search-path)
* [Composition over inheritance](https://en.wikipedia.org/wiki/Composition_over_inheritance)

# Comments

I would strongly recommend avoiding multiple inheritance from two or more concrete base classes.  Doing so plunges you into Python's Method Resolution Order (MRO) rules which are complicated, to say the least. Inheriting from Abstract Base Classes (interfaces) is far less dangerous. 

In this case calling `super().__init__` starts Python on a walk through the inheritance graph looking for `__init__` methods. Probably because some of the `__init__` methods we call using `super()` do not in turn call `super().__init__` the search ends without calling `Transform.__init__()`. So we were left with an uninitialized base class.

I'm not aware of any bug this was causing, but it was clearly unintended behavior. Multiple inheritance of concrete base classes create puzzles that are hard to solve even with extensive inspection of the code. I had to use the debugger and insert prints to figure out what was happening. It took a while.

Even expert developers find these situations confusing. Multiple inheritence raises the bar on the skill level required to understand napari's code base, so I think avoiding it when possible is a good idea. Generally there are other designs that meet all the same requirements.

# How has this been tested?
- [x] Added unit test that shows issue and then fixes it.

## Final checklist:

### DIFF
--- a/napari/utils/transforms/_tests/test_transform_chain.py
+++ b/napari/utils/transforms/_tests/test_transform_chain.py
@@ -79,3 +79,11 @@ def test_transform_chain_expanded(Transform):
     new_coord_2 = transform_chain_a(coord)
     new_coord_1 = transform_chain_expandded(coord)
     npt.assert_allclose(new_coord_1, new_coord_2)
+
+
+def test_base_transform_init_is_called():
+    # TransformChain() was not calling Transform.__init__() at one point.
+    # So below would fail with AttributeError: 'TransformChain' object has
+    # no attribute 'name'.
+    chain = TransformChain()
+    assert chain.name is None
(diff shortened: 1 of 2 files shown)


sofroniewn (approved):
LGTM! Thanks for the fix and test. Your points on the complexity of multiple inheritance are well taken. We'll likely be replacing ListModel in #1504, so fine to just fix here.

I'll leave open for a little in case anyone else has any additional comments and then merge

Issue #1535, opened 2020-08-08
Default key bindings of standard layers not working when inherited


JoOkuma:
## 🐛 Bug

The child object of a layer class does not inherit its key bindings.

This bug happens because of the changes in PR #1528.

## To Reproduce

1. Install the latest version of napari.
2. Execute the code provided below.
3. Try to use some key bindings such as hold SPACE to pan or zoom while adding points.

```python3
import napari
from skimage import data


class MyPoints(napari.layers.Points):
    def random_function():
        return 42


with napari.gui_qt():
    viewer = napari.view_image(data.astronaut())

    points = MyPoints()
    viewer.add_layer(points)
```

## Expected behavior

I expected the default key bindings to work normally.

## Environment

```text
napari: 0.3.7.dev12+ge468550.d20200806
Platform: Darwin-18.7.0-x86_64-i386-64bit
Python: 3.7.8 | packaged by conda-forge | (default, Jul 31 2020, 02:37:09) [Clang 10.0.1 ]
Qt: 5.14.2
PySide2: 5.14.2.3
NumPy: 1.19.1
SciPy: 1.5.2
Dask: 2.22.0
VisPy: 0.6.4

GL version: 2.1 ATI-2.11.22
MAX_TEXTURE_SIZE: 16384

Plugins:
- napari_plugin_engine: 0.1.7
- svg: 0.1.3
```

JoOkuma:
I'm not entirely sure if this is expected or not.

However, if the key binding inheritance is not expected, it should not show "hold <space> to pan/zoom" in the bottom right corner when in add/select mode.

sofroniewn:
Hi @JoOkuma, so sorry for taking so long to get back to you on this one. I'm running into this problem again myself actually too. Different context, but same issue. I'm not sure if @jni or @kne42 have any ideas on this one

kne42:
Looking back at the code, this was intended functionality https://github.com/napari/napari/blob/bc3298bb047b00d65133d5e4cf969b62aa65def1/napari/utils/key_bindings.py#L313-L315

IIRC, I originally had it toss out inherited keybindings I think maybe because some of the overwriting might get confusing.

I realize now that this is probably undesirable functionality. To fix this, you would need to traverse down the MRO for `parent` here, adding in each `class_keybindings` where present: https://github.com/napari/napari/blob/bc3298bb047b00d65133d5e4cf969b62aa65def1/napari/utils/key_bindings.py#L361-L363

[...]

PR #1810, opened 2020-11-01
Move viewer keybindings

sofroniewn:
# Description
I'm looking into moving our keybindings off of the `Viewer` and onto the `ViewerModel` as much as possible, this would match the `Layer` objects where the keybindings are. I think in the long run we probably want to allow for importing all our objects without keybindings attached, but for now this will further the objectives of #1808 by bringing our Viewer and ViewerModel closer together.

I'm running into a problem though which appears to be the the same as reported in #1535 which is that the keybindings don't work when inherited. We probably want to fix that in a standalone PR, but raising this now so that we have another example - cc @jni @kne42 

## Type of change


napari/__init__.py
```diff
@@ -57,7 +57,7 @@
 # Note that importing _viewer_key_bindings is needed as the Viewer gets
 # decorated with keybindings during that process, but it is not directly needed
 # by our users and so is deleted below
-from . import _viewer_key_bindings  # noqa: F401
+# from . import _viewer_key_bindings  # noqa: F401
```
sofroniewn: Note for debugging I'm not loading the `Viewer` keybindings, only the `ViewerModel` ones should be functional, but they aren't - same as was reported in #1535

sofroniewn:
I've updated this PR to also fix a regression from #1434 which where we lost the "highlight" coloring of our active dims slider. In theory that could be its own PR and I could pull that out if this PR becomes controversial, but otherwise it's in here for now.

I've moved the fullscreen and toggle play keybindings to the window. All other keybindings are on the `ViewerModel` and there are no more keybindings on the `Viewer` object itself.

I likely still have some tests to fix, and note that this won't work until #1535 is fixed

sofroniewn:
Ok, I have merged the changes from #1838, this is now ready for review @jni. Note that it also includes a fix for the last used slider highlighting.

jni (approved):
Awesome! It's a long PR but so simple in the end! :tada:

PR #1812, opened 2020-11-01
Add NAPARI_EXIT_ON_ERROR

pwinston:
# Description
This is a proposal for adding env var `NAPARI_EXIT_ON_ERROR`. When it's true napari will exit after reporting an exception.

If you use `NAPARI_CATCH_ERRORS=0` without this setting, napari will in some cases stream an endless series of exceptions. This is problematic because often the first one might be the important one, and the following ones are just happening because things are in a broken state.

With `NAPARI_EXIT_ON_ERROR` napari exits when that initial exception happens so you can see it and address it. While developing I mostly want this behavior. The only time I'd want `NAPARI_EXIT_ON_ERROR` off is if there are exceptions I cannot control that I want to ignore, which feels rare.

## Type of change

# How has this been tested?
- [x] Manual testing

## Final checklist:
- [x] I have made corresponding changes to the doc strings

### DIFF
--- a/napari/_qt/exceptions.py
+++ b/napari/_qt/exceptions.py
@@ -47,7 +47,7 @@ class ExceptionHandler(QObject):
 
     def __init__(self, parent=None, *, gui_exceptions=True):
         super().__init__(parent)
-        self.gui_exceptions = (
+        self.gui_exceptions: bool = (
             False if _set_false("NAPARI_CATCH_ERRORS") else gui_exceptions
         )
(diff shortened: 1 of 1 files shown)


sofroniewn (approved):
I have noticed the endless stream of errors too, which makes debugging very hard / impossible in this case. This seems like a very reasonable approach.

I'm approving, but will wait a little to merge in case anyone else wants to weigh in first

jni:
I like this, for the same reasons explained in the PR description. Thank you @pwinston!

jni (approved):

PR #1813, opened 2020-11-01
async-19: Minimap and Test Images

pwinston:
# Description
* New `QtTestImage` dropdown to choose test image: digits, astronaut, coffee, etc.
* Better minimap, now works for non-square images and thin edge tiles.
* `OctreeIntersection` now has a reference to`OctreeLevel`
    * So the minimap can get at the exact tile sizes.
    * Moved the key `get_chunks` routine from `OctreeLevel` to `OctreeIntersection`.

# Note about the minimap

From the code:

        # OCTREE_TODO: Consider this Qt bitmap rendering code just a rough
        # proof-of-concept prototype. A real minimap should probably be a
        # little OpenGL window, not a Qt bitmap. If we used OpenGL the GPU
        # does most of the work and we get better quality and effects, like
        # alpha and anti-aliasing. We do not want to spend a lot of time
        # becoming good at "rendering" into Qt Bitmaps.

# Type of change
- [x] Changes to experimental octree code

# Files
async-19-files

# How has this been tested?
- [x] Manual Testing

### DIFF
--- a/napari/_qt/experimental/render/qt_test_image.py
+++ b/napari/_qt/experimental/render/qt_test_image.py
@@ -58,7 +58,7 @@ class QtTestImageLayout(QVBoxLayout):
         self.octree.setChecked(1)
         self.addWidget(self.octree)
 
-        image_styles = ["Digits", "Astronaut"]
+        image_styles = ["Digits", "Astronaut", "Chelsea", "Coffee"]
         self.style = QComboBox()
         self.style.addItems(image_styles)
         self.addWidget(self.style)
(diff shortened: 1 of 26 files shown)


pwinston:
I'm marking this ready for review.

It's a nice little set of improvements and refactorings. See the PR description.

sofroniewn (approved):
This looks good to me. Note about minimap makes sense too, thanks! I'll merge after all tests pass

napari/_qt/experimental/render/qt_test_image.py
```diff
@@ -2,8 +2,14 @@
...
-import numpy as np
-from qtpy.QtWidgets import QCheckBox, QFrame, QPushButton, QVBoxLayout
+from qtpy.QtWidgets import (
+    QCheckBox,
+    QComboBox,
+    QFrame,
+    QPushButton,
+    QVBoxLayout,
+)
+from skimage import data as skimage_data
```
sofroniewn, line 12: @pwinston i just realized, the main repo does not actually depend on scikit-image, we have it in our test dependencies, so tests will pass with this, but it's not actually required. Maybe you can make them optional in #1834, and my mistake for missing this in review!

PR #1815, opened 2020-11-02
enhance affine example to include scipy.ndimage

sofroniewn:
# Description
Following discussion with @GenevieveBuckley and @m-albert on dask-image affine transform PR at scipy japan - see https://github.com/dask/dask-image/pull/159#issuecomment-720192323 in particular. I've updated this napari affine example to include the result of transforming the image with `scipy.ndimage.affine_transform` - you can see how that example actually does the transform of the pixels. I also add a note about why you need the inverse of the affine too (you can read more about that in the scipy docs too) https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.affine_transform.html

Here in green is what the scipy transformed layer looks like, overlaid on the napari transformed layer in blue

Screen Shot 2020-11-01 at 5 58 14 PM

## Type of change

### DIFF
--- a/examples/affine_transforms.py
+++ b/examples/affine_transforms.py
@@ -3,6 +3,7 @@ Display an image and its corners before and after an affine transform
 """
 import numpy as np
 import napari
+import scipy.ndimage as ndi
 
 # Create a random image
 image = np.random.random((5, 5))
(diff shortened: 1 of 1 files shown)


jni (approved):
Love it!

examples/affine_transforms.py
```diff
@@ -25,7 +26,14 @@
     viewer.add_image(image, colormap='blue', opacity=.5, name='moving', affine=affine)
     viewer.add_points((corners_h @ affine.T)[:, :-1], size=0.5, opacity=.5, face_color=[0, 0, 0.8, 0.8], name='mv corners')
 
-    # Reset the view
-    viewer.reset_view()
+    # Note how the transformed corner points remain at the corners of the transformed image
+
+    # Now add the a regridded version of the image transformed with scipy.ndimage.affine_transform
```
GenevieveBuckley, line 31: I think it's worth explaining *why* you are showing this comparison. Are you making the argument that it's better for people to use the napari affine transformation because the regridding/resampling causes some distortions in the way the data looks? I think that's a valid argument to make.

We should try and be a lot clearer about what take home message we are expecting users to understand from this example.

GenevieveBuckley (approved):
I think we should try and be clearer about what the take home message is from this example, but other than that this looks good to me!

examples/affine_transforms.py
sofroniewn: I like that idea! We can do that when we write up a tutorial on basic usage of the affine

GenevieveBuckley: ...isn't *this* the tutorial on basic affine use?

jni: Not yet! Tutorials live in `napari/tutorials`. *Eventually*, we want to unify them. That hasn't happened yet though.

sofroniewn: Yeah, I was thinking one of our markdown files in the tutorials repo, but agreed I can't wait till we can unify them. It will change everything, making development of tutorials much easier and more fun!!!

Issue #1211, opened 2020-05-06
Plugin parameters


tlambert03:
## 🚀 Feature
As [raised in zulip](https://napari.zulipchat.com/#narrow/stream/212875-general/topic/plugin-params) and elswhere, we need a mechanism to pass arbitrary parameters to plugins (beyond those declared in the hook specification).  Currently, `napari_plugin_engine` (like `pluggy`) does not allow calling hooks with arbitrary kwargs (it's not an error, but they are not passed on) ... but we can probably come up with an acceptable way to open that up.  The main things to decide are 

1) what sort of API do we want to support on the napari side?  Currently, calling `viewer.open(**kwargs)` passes all additional `**kwargs` to the respective `add_*` method.  So we'd need to either add a `plugin_kwargs={}` argument (probably better), or do some clever separation of of kwargs (probably worse).  I also think it would be reasonable to _only_ accept `plugin_kwargs` when also providing a specific `plugin`.

2) haven't thought about the hook calling side as much, but the lines where kwargs not specified in the hook specification are stripped are right [here](https://github.com/napari/napari-plugin-engine/blob/0ea4f144783fa0075a86bd31df15ea47e32355cc/napari_plugin_engine/callers.py#L168-L172)... so that's a place to start.

joshmoore:
My initial (and most pressing) use case is for [ome-zarr-py](https://github.com/ome/ome-zarr-py). `napari https://.../$IMAGE.zarr` works fine. In _some_ cases, however, the necessary [metadata](https://github.com/zarr-developers/zarr-specs/issues/50) is not present in the dataset and I would like to override it. In those cases, I could see using one or all of the following:

```
# Pass the metadata directly.
napari --plugin-metadata='{"multiscales": ...}' https://.../$IMAGE.zarr

# or pass that JSON via a file prefixed with '@"
napari --plugin-metadata=@filepath.json https://.../$IMAGE.zarr

# or explicitly pass a plugin-specific argument.
# Here, the ome-zarr plugin would use the provided URL to load the JSON
napari --plugin-argument=metadata_source=https://... https://...$IMAGE.zarr


```

tlambert03:
thanks.  not that I'm suggesting this as a long-term solution ... but I found myself needing a very [similar thing recently for a plugin I was writing](https://github.com/tlambert03/napari-ndtiffs#overriding-parameters), (when metadata was not present in the dataset) and I wrote a little context manager to be able to do something like this:

```python
viewer = napari.Viewer()
[...]

PR #1816, opened 2020-11-02
Add magicgui hook spec for "functional" plugins

sofroniewn:
# Description
Following discussion with @GenevieveBuckley and @jni at SciPy Japan I've adapted #1800 which was exploring adding a viewer dock widget hookspec to add a magicgui based hookspec for more functional like plugins.

This is super preliminary, and definitely a work in progress, but I wanted to get some proof of concept stuff out here. Note this was originally proposed and discussed as complementary to #1800 here https://github.com/napari/napari/issues/1327#issuecomment-639081999.

This hookspec also doesn't preclude us exploring more of the functional ideas discussed here https://github.com/napari/napari/issues/1211#issuecomment-671106600 or passing of layer tuple data discussed here https://github.com/napari/napari/issues/1272#issuecomment-633106453

I've hooked this up with an example plugin (called `io-test` in the gif below for no good reason than I already had that repo installed) using code from this magicgui example https://magicgui.readthedocs.io/en/latest/examples/napari_parameter_sweep/

The plugin has the following hook implementation
```python
from magicgui._qt.widgets import QDoubleSlider
from napari_plugin_engine import napari_hook_implementation
from napari.layers import Image
import skimage.data
import skimage.filters


@napari_hook_implementation
def napari_experimental_provide_functions():
    args ={
        'auto_call':True,
        'sigma':{"widget_type": QDoubleSlider, "maximum": 6, "fixedWidth": 400},
        'mode': {"choices": ["reflect", "constant", "nearest", "mirror", "wrap"]},
    }

    def gaussian_blur(layer: Image, sigma: float = 1.0, mode="nearest") -> Image:
        """Apply a gaussian blur to ``layer``."""
        if layer:
            return skimage.filters.gaussian(layer.data, sigma=sigma, mode=mode)

    return [(gaussian_blur, args)]
```
Note this example makes a magicgui import for the `QDoubleSlider` and imports the napari layers for Typing purposes.

I'd like us to make those (or equivalent) imports available from `napari_plugin_engine.types` instead so you could just do

```python
from napari_plugin_engine import napari_hook_implementation
from napari_plugin_engine.types import DoubleSlider, Image
```

and the plugin wouldn't have to depend on napari or magicgui at all. I think that should be doable.

Here's a gif of this in action

plugin_magicgui

Note this is all super in progress, this implementation is very crude, but I just wanted to get the ball rolling a little, and thanks in particular to @GenevieveBuckley for some good brainstorming, and @tlambert03 for his beautiful work creating magicgui.

## Type of change


sofroniewn:
Looking at this again a little. I'm wondering if we'd want to expose a public function on the window like `viewer.window.add_dock_widget_from_function` or `viewer.window.add_function_with_magicgui` etc. that that this hookspec would then use. Could be nice, might not be needed. I'd love some more feedback on this PR though.

I'll note that there are still a bunch of open questions on #1849 around dock widget plugins that intersect with this PR

tlambert03:
I like that idea. On the other one, I was trying to make the "plugin" part of it as "un-special" as possible... mostly just around the fact that we'll be putting stuff in the plugins menu (by otherwise trying to make it just another dock widget as far as the code is concerned).  I like a similar idea here, where the "plugin" part is merely just letting devs say "here's a function you might want to use", but then the mechanism of adding it is no different than an end-user might do with their own function

Sorry, still haven't looked at the code though

tlambert03:
I'll also add that there is basically a fully rewritten version of magicgui that I still need to finalize: https://github.com/napari/magicgui/pull/43

It shouldnt change too much of the public API that you'll be using here, but definitely some (it's a breaking change, that is "worth it").  No reason not to proceed anyway, just wanted you to know

sofroniewn:
> I like a similar idea here, where the "plugin" part is merely just letting devs say "here's a function you might want to use", but then the mechanism of adding it is no different than an end-user might do with their own function

Yes exactly, that's the philosophy I was going for!

Given that I might make a new PR that just has the `viewer.window.add_function_with_magicgui` (or whatever we call it part) and then the PR to make that accessible via the plugin interface will really just be like #1849 which really is much about menubars etc. and hopefully we'll be able to reuse for this too

There might also be some "connecting" we can do inside that `add_function_with_magicgui` like the refreshing of the layer list choices etc, that we might want in common if someone is using the plugin interface or adding via the main API. We can discuss this on the new PR though.

> I'll also add that there is basically a fully rewritten version of magicgui that I still need to finalize: napari/magicgui#43

I just had a quick look, seems very exciting, I made a comment over there that I'm happy to help with review if you'd like.

All in all exciting times, there's been a lot of enthusiasm for both this PR and #1849 in our euro time-zone dev meeting

sofroniewn:
@tlambert03 @jni I've updated this with the latest from master. It's still a little rough around the edges, for example right now you have a single option in the menubar that will add all the functions that plugin is providing. Instead I'd like to create a nested menubar so that you can select each function separately.

Thinking also about #1849 and how so much of the menubar functionality is going to be shared, I'm thinking now that we want that nested list to include both the "function widgets" and the "qt dock widget" in the same list.

I might start a new PR which includes these changes and the changes from #1849 in one. I know it combines two PRs into one, but it might just be so much more clear how to name/ design things if I have both in there to work with.

This and other things here are just early ideas though, so feedback and direction very welcome!

haesleinhuepf:
Awesome! I'm happy to see this evolving.

I'd have some questions which came up while I started implementing something against this API in this [plugin prototype implementation](https://github.com/clEsperanto/napari_pyclesperanto_assistant). It appears tricky to make something that's intuitive for the end-user. Happy to discuss those in the next dev-meeting. I just wanted to provide a list of potential issues I hit earlier.

* Is it possible to configure the output layer of a function? E.g. I'd like to configure `contrast_limits`, `blending` or `color_map`. Also `name` etc.

* If my plugin takes a while for computing the new layer, will napari freeze?

* If not, can the user start another plugin while mine is running?

* What if my operation has two input images? How can they be chosen? Would a layer-selection-pulldown also contain the result-image of the current operation? 🙃

* Can my function have two output images? (a bit hypothetical, but possible)

* Are recursive layer updates supported? E.g. when I update an early operation in a chain of operations, will subsequent operations update? This `feature` might have an optional on/off switch as it has the potential to create endless loops 😉

* Is it possible to have a plugin-type that has access to the whole list of layers and related magicguis? E.g. for reading out the sources of the layers (see #2033) and plugins/parameters that created the layer? This would allow code generation from a list of layers... 😃

tlambert03:
> * Is it possible to configure the output layer of a function? E.g. I'd like to configure `contrast_limits`, `blending` or `color_map`. Also `name` etc.
> * Can my function have two output images? (a bit hypothetical, but possible)

The answer to both of these is to annotate your return value not as `Image` but (currently) as `Layer`, and then return a list of layer data tuples (described a while back [on zulip](https://napari.zulipchat.com/#narrow/stream/212875-general/topic/magicgui.20and.20layers/near/215323809)):
```python
@magicgui(auto_call=True)
def gen_layer(mean=1) -> napari.layers.Layer:
    data = np.random.rand(256, 256) * mean
    meta = {
        'name': 'result',
        'contrast_limits': [data.min(), data.max() * 2],
    }
    return [(data, meta)]
```

two caveats:
- as noted on zulip: the current napari logic will only use the metadata on layer creation, not upon updating an existing layer (wherein it only updates the data)
- **big caveat!  don't get used to that annotation!** I've grown to dislike the fact that we used `Image` or `Layer` as return annotations, when in fact one is returning numpy arrays or tuples, etc... (my linter dislikes it even more :joy:).  So https://github.com/napari/napari/issues/2028 proposes that this will change, for instance to:
```python
def my_func() -> napari.types.ImageData: ... # to provide a numpy array as an image
def my_func() -> napari.types.LayerData: ... # to provide a single layer data tuple with meta
def my_func() -> List[napari.types.LayerData]: ... # to provide a multuple layerdata tuples
```

> * If my plugin takes a while for computing the new layer, will napari freeze?

yep, everything in this API is all still single threaded.  At the moment, you'd probably need to use something like [`napari.qt.threading.thread_worker`](https://napari.org/docs/dev/events/threading.html#threading-in-napari-with-thread-worker) inside of your magicgui function.  But we could think about ways to introduce a magic kwarg like `threaded=True` to do that for you?

> * If not, can the user start another plugin while mine is running?

see above... long functions will block _everything_ in the main thread unless [`thread_worker`](https://napari.org/docs/dev/events/threading.html#threading-in-napari-with-thread-worker) is used


> * What if my operation has two input images? How can they be chosen? Would a layer-selection-pulldown also contain the result-image of the current operation? 🙃

It's fine to annotate multiple parameters in your function as `napari.layers.Image`, see the [napari image arithmetic example](https://napari.org/magicgui/examples/napari/napari_img_math.html). Note also in that example, when you click the "execute" button, the result-image _does_ appear in the selection pulldown... so yep, you can do all this.


> * Are recursive layer updates supported? E.g. when I update an early operation in a chain of operations, will subsequent operations update? This `feature` might have an optional on/off switch as it has the potential to create endless loops 😉

We don't currently have any sort of implicit operation chaining API, so that would depend on how you implemented the chaining of operations.  However, magicgui does support [event-based auto-updates](https://napari.org/magicgui/index.html#connecting-events)... so it could work provided that you did something like use `auto_call=True`, _and_ did something like this:

```python
@magicgui(auto_call=True)
def func_a(nx: int = 64, ny: int = 64):
	print("in function a", locals())
	return nx * ny

@magicgui(auto_call=True)
def func_b(num: int, mult=1.0) -> Image:
	return num * mult

def _on_func_a(event):
    func_b.num.value = event.value

func_a.called.connect(_on_func_a)
```

i'll note that in testing this out, the basic idea works fine, but the fact that we're currently doing implicit type coercion in napari when annotating with `napari.layers.Image` (as mentioned in https://github.com/napari/napari/issues/2028) makes this _much_ harder to do.  So that's another reason we should change that.


> * Is it possible to have a plugin-type that has access to the whole list of layers and related magicguis? E.g. for reading out the sources of the layers (see #2033) and plugins/parameters that created the layer? This would allow code generation from a list of layers... 😃

would need some more detail.  You can definitely get all the layers, and you can also query all of the current dock widgets... but yeah, for "provenance" type metadata ala #2033, we'd need to somehow embed that a given layer came _from_ the output of a magicgui widget.  Doable (and very cool!) but not yet there.

haesleinhuepf:
> * as noted on zulip: the current napari logic will only use the metadata on layer creation, not upon updating an existing layer (wherein it only updates the data)

Yes, I've noted that. That's why I have [this repetetive ugly implementation](https://github.com/clEsperanto/napari_pyclesperanto_assistant/blob/master/napari_pyclesperanto_assistant/_operations/_operations.py#L29-L37) in my functions. I need it to update `contrast_limits` all the time, otherwise the user doesn't see anything, when applying gamma-correction as a filter for example.

> But we could think about ways to introduce a magic kwarg like `threaded=True` to do that for you?

I'd love to try this out. In ImageJ this is the default-strategy, which also has downsides.

> the result-image _does_ appear in the selection pulldown... so yep, you can do all this.

Can this behaviour be turned off? I'd like to prevent that input and output layer are the same, if possible.

> However, magicgui does support [event-based auto-updates](https://napari.org/magicgui/index.html#connecting-events)... so it could work provided that you did something like use `auto_call=True`, _and_ did something like this:

Wow! I can't wait to try this out 🤩 

Thanks for your feedback! Great to see that most of my "issues" are already solved or at least on the way :-)

tlambert03:
I've added a more complete example of chaining together functions here: https://github.com/napari/magicgui/blob/master/examples/chaining.py

*edit: updated link*

haesleinhuepf:
> I've added a more complete example of chaining together functions here: https://github.com/tlambert03/magicgui/blob/more-examples/examples/chaining.py

Can you please check that link? Appears like 404 to me 🙃

tlambert03:
whoops!  sorry, merged into master, link is now here: https://github.com/napari/magicgui/blob/master/examples/chaining.py

sofroniewn:
I'm closing this now as these changes and more refinements are included in #2080, thanks for the discussion all

PR #1807, opened 2020-11-01
Allow ability to run all tests on local Windows installs

UmbWill:
Fixes #1455

# Description

## Type of change

- [ ] Bug-fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# References

# How has this been tested?

- [ ] example: the test suite for my feature covers cases x, y, and z
- [ ] example: all tests pass with my change

## Final checklist:
- [ ] My PR is the minimum possible work for the desired functionality
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works

PR #1817, opened 2020-11-02
Try different test for popup skipif mark

jni:
# Description

I am a bit unhappy that we had to use the deprecated string input to pytest.mark.skipif in #1807, so I'm opening this branch to experiment with a few options on the CI.

## Type of change

# References

# How has this been tested?

## Final checklist:

### DIFF
--- a/napari/_tests/test_with_screenshot.py
+++ b/napari/_tests/test_with_screenshot.py
@@ -1,4 +1,6 @@
 import collections
+import os
+import sys
 
 import numpy as np
 import pytest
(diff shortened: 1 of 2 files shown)


jni:
Ok as e734313 shows, indeed switching to a string is sufficient to fix things! The mind boggles. I'm going to try one more thing before submitting a bug report to pytest...

jni:
Boom. Looks like CI is set to `true`, not 1, and through some dark magic this confused pytest. Explicitly checking the CI value fixes things! :tada:

jni:
Ok @napari/core-devs this is ready to merge imho! I'll also raise an issue in pytest because this is weird-as.

sofroniewn (approved):
LGTM! Nice job chasing this down!

jni:
Ok thanks to the help of the pytest devs over at pytest-dev/pytest#7999, I finally figured out that this is why this was happening:

```python
>>> True and 'hello'
'hello'
```

:facepalm: This is the same Python magic that enables the syntax:

```python
var = input_var or DEFAULT
```

which I used to like but now I think it's dangerous hubris. :joy:

PR #1814, opened 2020-11-02
Pre-sort tracks data by ID and time

AbigailMcGovern:
# Description
Currently, inputted tracks vertices must be sorted according to track ID, then time (i.e., in ascending order columns 0 then 1 of the input array). If the data is not sorted, an error is returned. This PR resolves this by pre-sorting the data.

## Type of change

- [x] Bug-fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

# References
This change was discussed by @jni in https://github.com/napari/napari/issues/1662#issuecomment-703180933 and https://github.com/napari/napari/pull/1361#discussion_r497459001.

# How has this been tested?

- [x] Added test for unsorted data
- [x] Track layer tests pass

## Final checklist:
- [x] My PR is the minimum possible work for the desired functionality
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works

Issue #1662, opened 2020-09-30
Tracks performance benchmarks


kevinyamauchi:
## 🚀 Feature
After #1361 is merged, I think we should add airspeed benchmarks for the Tracks layer.

## Motivation

Having performance benchmarks will help us know how many tracks we can reasonably visualize currently and identify parts of the code we can optimize in the future.

## Pitch

Some things to benchmark:

- creation of Tracks layers with varying numbers of tracks
- updating the coloring of tracks with the different coloring modes
- slicing of Tracks layers with varying numbers of tracks

kevinyamauchi:
I can work on this after we merge #1361!

jni:
I'll add links here for various performance suggestions I made on #1361, I think it makes sense to try them together with the benchmarks:

- [ ] Use a track id -> indices map using `scipy.sparse.csr_matrix` to be able to quickly retrieve data given a track id: https://github.com/napari/napari/pull/1361#discussion_r497459001
- [ ] pre-sort full dataset by ID and time using np.lexsort: https://github.com/napari/napari/pull/1361#discussion_r497459001
- [ ] Use an artificial "scale" in the time axis when generating the cKDTree of track points so that track selection with the mouse always finds points in the current frame: https://github.com/napari/napari/pull/1361#discussion_r497482781

quantumjot:
> I think it makes sense to try them together with the benchmarks 

Empirically, I've noticed that the `build_tracks` function can be slow with a large number of tracks. If we enforce/ensure the lexicographic ordering of the data and properties (by ID, time), that would make that function much simpler, and would remove the loop.  

In general, how do you feel about sorting the data in place? I guess it means that data =/= layer.data.

liu-ziyang:
@kevinyamauchi Hi there, we would like to help on adding perf benchmarks to napari, but we are exploring adding monitoring in general in addition to the tracks layer, do you happen to have work in progress on this that we can build on top of? cc @jaimergp

kevinyamauchi:
Hey @potating-potato ! It's great that you're working on perf benchmarks. Unfortunately, I didn't manage to carve out the time to work on this, so I don't have a WIP you can build on top of.

Looking forward to seeing what you come up with!

epompeii:
Would you all like to be able to visualize/share the `asv` results?

If so, I've built a continuous benchmarking tool called Bencher: https://github.com/bencherdev/bencher
It integrates with `asv` to make it easier to visualize and share benchmark results.

PR #1818, opened 2020-11-02
Sort tracks and track properties by ID then t

AbigailMcGovern:
# Description
This PR follows a previous PR, the intention of which was to allow track data to be inputted without being sorted prior (i.e., according to ID then t). Previously, I had neglected the properties data. This PR rectifies this issue by sorting both the tracks and properties data. 

## Type of change

# References
This is a follow up to #1814
This change was discussed by @jni in https://github.com/napari/napari/issues/1662#issuecomment-703180933 and https://github.com/napari/napari/pull/1361#discussion_r497459001.

# How has this been tested?

- [x] Added test to ensure properties are also sorted when inputting unsorted tracks data.
- [x] All tracks layer tests pass. 

## Final checklist:

### DIFF
--- a/napari/layers/tracks/_tests/test_tracks.py
+++ b/napari/layers/tracks/_tests/test_tracks.py
@@ -61,6 +61,7 @@ def test_track_layer_data_flipped():
     """Test data flipped."""
     data = np.zeros((100, 4))
     data[:, 1] = np.arange(100)
+    data[:, 0] = np.arange(100)
     data = np.flip(data, axis=0)
     layer = Tracks(data)
     assert np.all(layer.data == np.flip(data, axis=0))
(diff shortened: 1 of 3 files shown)


jni:
@AbigailMcGovern Thank you!

It looks like you need to set up your napari development environment with some pre-commit hooks to make sure the formatting matches the rest of the library. You can find instructions for how to do this here:

https://napari.org/docs/dev/developers/CONTRIBUTING.html

napari/layers/tracks/_tests/test_tracks.py
```diff
@@ -80,6 +81,18 @@ def test_track_layer_properties(properties):
         np.testing.assert_equal(layer.properties[k], v)
 
 
+@pytest.mark.parametrize("properties", [{}, properties_dict, properties_df])
+def test_track_layer_properties_filpped(properties):
```
jni:
```suggestion
def test_track_layer_properties_flipped(properties):
```

jni (approved):
Other than a minor typo and the missing formatting, this is great! Thank you for including a test right off the bat @AbigailMcGovern!

napari/layers/tracks/_tests/test_tracks.py
AbigailMcGovern: Thanks @jni! All taken care of. I've got to say, those still look identical 😂 🤦‍♀️

sofroniewn (approved):
Ok this looks good to me. Curious if @quantumjot or @kevinyamauchi have any additional thoughts?

I'll leave a open a little longer in case they want to comment and then merge if not

quantumjot:
Looks good to me.  Thanks @AbigailMcGovern!

Issue #1744, opened 2020-10-19
Ability to average over multiple slices along the viewing axis


alisterburt:
## 🚀 Feature
The ability to average multiple slices along viewing direction in 2d viewer

## Motivation
Interpretation of 3D images with low signal to noise ratio often benefits from the on-the-fly projection of multiple slices along the viewing direction. This is often performed in popular cryo-electron tomography software such as 3dmod from the IMOD suite.

## Pitch
A small integer field or integer slider which would allow playing with the projected thickness of a 3D image viewed in 2D.

## Alternatives
Filters applied on 3D images can remove the requirement for this sort of visualisation trickery but tend to be expensive to compute when volumes get large

## Additional context
I see that in the `ChunkKey` object (`napari.components.chunk._request.ChunkKey`) associated with the `ImageSlice` (`napari.layers.image._image_slice.ImageSlice`) indices are already an attribute and that they default to `slice(None, None, None, None)` for x and y in 3d data. I guess this implies that the functionality for slicing the data is already there and just not exposed?

If someone can point me in the right direction I'd be happy to try to get this working with a small example dataset, I'm just not sure where the best place to start would be

sofroniewn:
Welcome @alisterburt! Yes, this would be a very useful feature. If I understand your request correctly, there was some discussion of it here https://github.com/napari/napari/issues/7, which is one of the oldest napari issues still open!

The idea was to use a range slider with two independently movable endpoints (like our current contrast limits slider) to allow people to define a range for a slice instead of just a single point. You could then do a max projection on that slice.

In all we actually imagined 3 modes for the sliders:
- point slice (which is what we have now)
- range (which would do the max project as mentioned above)
- clip (which would clip the data, also possibly useful for 3D viewing).

Probably the biggest question is whether we want to try and leverage the graphics card and do this in a shader or just do it in the cpu (curious what @royerloic and @jni think about this, this has been @royerloic most requested feature for quite some time). Depending on what we choose here the implementation could end up being a little different.

We might also want to consider how to support things like the mixed 2D and 3D rendering that was prototyped in #839 and also non-orthogonal slicing as we decide how to support this. We could also start with a simple cpu approach and move to a gpu approach in time.

The key place in the code where the slice indices are determined right now is https://github.com/napari/napari/blob/master/napari/layers/base/base.py#L760-L816 and https://github.com/napari/napari/blob/master/napari/layers/base/base.py#L551-L583, these would need to be adjusted to handle a range of data.

[...]

PR #1820, opened 2020-11-02
[WIP] Multislice projection

alisterburt:
# Description
This PR is a work in progress implementation of changes discussed in #1744 relating to adding in the ability to project over intervals on an axis rather than point slicing which is done at the moment

@sofroniewn suggested the following approach

> - We might want to add back in the interval and slicing mode enum for our dims model
> - Then on the Qt side have it swap between the scroll bar and the range slider when the slicing mode changed
> - Then in the base layer rather than just use the point, use the point/ interval depending on the dims mode to do the slicing.
> - Then maybe easiest in the vispy_image_visual we can do a cpu based max intensity projection, and send that to the graphics card. Doing the max intensity projection on the gpu would be possible if projecting in one dimension, but not more due to the current vispy API (the library we use for rendering). Maybe that's ok though.
> - We'd probably want to roll back to point based slicing for all non-image layers at first, though it might be nice to support for our other layers later.

Progress up to now...
- `Dims` model now has an `interval` property which initialises properly
- `Dims` model has a `set_interval` method which sets both the interval and the step
- `Dims` model has an event emitter associated with the interval property
- `QtDimsSliderWidget` has an incomplete `_create_interval_slider_widget` method
- `QtDimsSliderWidget` attempts to update both step and range of the `Dims` model in its `_interval_values_changed` method

Any feedback/ideas on basically everything is appreciated - I'm still trying to wrap my head around the codebase and everything Qt related remains a bit mysterious

### DIFF
--- a/napari/_qt/widgets/qt_dims_slider.py
+++ b/napari/_qt/widgets/qt_dims_slider.py
@@ -121,6 +121,16 @@ class QtDimSliderWidget(QWidget):
         """
         self.dims.set_current_step(self.axis, value)
 
+    def _interval_values_changed(self, values):
+        """Interval slider changed to these new values
+        """
+        # Set step of dims to midpoints (rounding and clamping is done in dims.set_current_step)
+        slider_midpoint = sum(values) / len(values)
+        self.dims.set_current_step(self.axis, slider_midpoint)
+
+        # Set interval of dims
+        self.dims.set_interval(self.axis, values)
+
     def _create_range_slider_widget(self):
         """Creates a range slider widget for a given axis."""
         # Set the maximum values of the range slider to be one step less than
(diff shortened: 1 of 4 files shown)


sofroniewn:
> Any feedback/ideas on basically everything is appreciated - I'm still trying to wrap my head around the codebase and everything Qt related remains a bit mysterious

Looks like it's coming along nicely on the Qt/ Dims side - the next big step will be doing the slicing which starts here https://github.com/napari/napari/blob/master/napari/components/viewer_model.py#L316-L319

One thing to think about is that the new tracks layer actually does some Max Intensity Projection like computations in the shader, and we could try and use vispy in that sort of way. Curious what @royerloic would recommend we try first - cpu based or go with gpu based?

royerloic:
Late to this party, two ways this can be done:

i) On the CPU, brute-force numpy.max is probably too slow for large datasets and interactivity. If we were computing average projections, integral images would be the way to go (https://en.wikipedia.org/wiki/Summed-area_table). But we want _max_ projections instead. I just had an idea: you just need to sort the pixels values and keep the indices of the values from highest to lowest. Given an interval [u, v] for which you want the max pixel value, you go through your sorted index list and stop as soon as one of the indices falls within [u, v]. Instead of an integral image you use a permutation(sorting) table that gives you the depth along the axis of each value sorted by decreasing intensity. Very fast for large intervals, slower for smaller ones. Probably not the optimal algorithm, I am sure a better optimal data-structure can be conceived, and maybe someone already has. I am thinking a tree-like structure to acheive log(n) type search performance for the max.  Anyhow, probably not worth the effort given that the real solution is probably below:   

ii) On the GPU that's basically vispy's 3D rendering shader augmented with 'bounding-box' parameters implemented as uniforms. The transform has to be set carefully to have an orthogonal projection and aligned axis. The clearvolume volrender kernel shows how to do the boundary boxes: https://github.com/ClearVolume/clearvolume/blob/master/src/java/clearvolume/renderer/opencl/kernels/VolumeRender.cl
advantage of ding it this way is that we can do slice  projections with arbitrary transforms...

alisterburt:
@royerloic thanks for the input! I had never come across integral images before, very cool stuff. 

The second way is more in line with what @tlambert03 suggested which is the way I've been playing with it, trying to replace the definition of nearpos and farpos with extra snippets - appreciate the link to the clearvol implementation, very useful!

sofroniewn:
@alisterburt just going through some old PRs, a lot has changed since this one started. Do you still want it open? What functionality here has been implemented in other PRs, what is still remaining? Thanks!!

alisterburt:
Thanks @sofroniewn ! As it stands we can achieve this using #3023 although that functionality isn't exposed in the GUI (yet)

This should be closed as it isn't being worked on but if someone were to resurrect it we might think about
- enabling range slicing on the dims slider as discussed above
- on activation of range slicing, `ImageVisual` is swapped out for `VolumeVisual` with plane rendering activated
- plane position, normal direction and thickness used to select data in sliced range

Closing for now!

Issue #1690, opened 2020-10-04
Move grid information off individual layers


sofroniewn:
## Refactor

As part of improvements to our handling of transforms I propose we move `grid` mode off of the individual layers and have all of that done at the level of the viewer

GenevieveBuckley:
This sounds like a good idea. 

What impact do you think this change will have on the current situation? Eg: right now when I select points the coordinates are very different depending on whether grid mode is on or off, so I can't toggle between these views.

One situation I currently find frustrating is that we don't have an easy way to do matched control point selections. We avoided doing this in the [sudoku example](https://github.com/napari/napari/pull/1616#issuecomment-700146465) by using RANSAC, but for a lot of real-world applications I really want to be able to have two images side by side while I pick out the matched control points. Since example use cases are helpful, here's [a non-napari GUI I made for our lab](https://github.com/DeMarcoLab/correlateim), which uses [this python implementation](https://github.com/hofmann-tobias/cpselect) of the MATLAB cpselect function.

sofroniewn:
> What impact do you think this change will have on the current situation? Eg: right now when I select points the coordinates are very different depending on whether grid mode is on or off, so I can't toggle between these views.
>
> One situation I currently find frustrating is that we don't have an easy way to do matched control point selections. 

We could try and get this fixed!! It might be a little tricky without having two canvases, but we can try. Thanks for the links. I do want us to push on the multicanvas too once all of this is done.

From an implementation standpoint, we'd probably need to pass the grid transform to the [base vispy visual](https://github.com/napari/napari/blob/master/napari/_vispy/vispy_base_layer.py) somehow and remove it from the [base layer itself](https://github.com/napari/napari/blob/master/napari/layers/base/base.py#L184-L186).

We could also try pulling all the [grid code out of the viewer_model](https://github.com/napari/napari/blob/master/napari/components/viewer_model.py#L404-L458) and into a little object called `Grid`, so you could do things like `viewer.grid.stride` and `viewer.grid.n_columns` etc.

One big question is then how to handle the camera - that now exists in a gridworld space, but the cursor and be made to work in the world space of the selected layer, irrespective of where in the grid it is. You should then be able to mark up images as you want with points in the correct coordinate system irrespective of where things are in the grid. Am I making sense here or do you think I have it way off? It's definitely tricky.

If you want to start working on this I'm happy to review PRs/ try and offer guidance

GenevieveBuckley:
[...]

PR #1821, opened 2020-11-02
Add a simple grid object

sofroniewn:
# Description
As people seem to like using grid mode and it will probably be around for a while, this PR is beginning to clean it up a little.

It adds a new `Grid` object which is very simple and holds a couple of attributes about the grid. Right now we still just use the stride to put elements in the grid, but we could now add two modes to the grid - one called `Regular` which uses the stride and one called `Positioned` (names to be determined) which would let you set the position of each layer in the grid as you wish. This could alleviate some of the robustness concerns @GenevieveBuckley have expressed about the current grid mode. I probably won't do that in this PR, but it's something that we now have space for.

This PR also does not move the grid information off the layer i.e. it does not close this PR #1690 or deal with any of the tricky camera/ coordinate system stuff. I'd like to do that in a follow-up PR after this one.

I still need to add some more tests, and I want to go back and add deprecation warnings in from of all the public methods that I have pulled. I should be able to preserve all their functionality.

## Type of change

### DIFF
--- a/napari/components/grid.py
+++ b/napari/components/grid.py
@@ -3,8 +3,11 @@ import numpy as np
 from ..utils.events import EmitterGroup
 
 
-class Grid:
-    """Grid mode.
+class GridCanvas:
+    """Grid for canvas.
+
+    Right now the only grid mode that is still inside one canvas with one
+    camera, but future grid modes could support multiple canvases.
 
     Attributes
     ----------
@@ -16,7 +19,7 @@ class Grid:
         Number of rows and columns in the grid. A value of -1 for either or
         both of will be used the row and column numbers will trigger an
         auto calculation of the necessary grid size to appropriately fill
-        all they layers at the appropriate stride.
+        all the layers at the appropriate stride.
     stride : int
         Number of layers to place in each grid square before moving on to
         the next square. The default ordering is to place the most visible
@@ -69,7 +72,7 @@ class Grid:
     def actual_size(self, nlayers=1):
         """Return the actual size of the grid.
 
-        This will return the size paramater, unless one of the row
+        This will return the size parameter, unless one of the row
         or column numbers is -1 in which case it will compute the
         optimal size of the grid given the number of layers and
         current stride.
--- a/napari/components/grid.py
+++ b/napari/components/grid.py
@@ -1,0 +1,140 @@
+import numpy as np
+
+from ..utils.events import EmitterGroup
+
+
+class Grid:
+    """Grid mode.
+
+    Attributes
+    ----------
+    events : EmitterGroup
+        Event emitter group
+    enabled : bool
+        If grid is enabled or not.
+    size : 2-tuple of int
+        Number of rows and columns in the grid. A value of -1 for either or
+        both of will be used the row and column numbers will trigger an
+        auto calculation of the necessary grid size to appropriately fill
+        all they layers at the appropriate stride.
+    stride : int
+        Number of layers to place in each grid square before moving on to
+        the next square. The default ordering is to place the most visible
+        layer in the top left corner of the grid. A negative stride will
+        cause the order in which the layers are placed in the grid to be
+        reversed.
+    """
+
+    def __init__(self, *, size=(-1, -1), stride=1, enabled=False):
+
+        # Events:
+        self.events = EmitterGroup(
+            source=self, auto_connect=True, update=None,
+        )
+
+        self._enabled = enabled
+        self._stride = stride
+        self._size = size
+
+    @property
+    def enabled(self):
+        """bool: If grid is enabled or not."""
+        return self._enabled
+
+    @enabled.setter
+    def enabled(self, enabled):
+        self._enabled = enabled
+        self.events.update()
+
+    @property
+    def size(self):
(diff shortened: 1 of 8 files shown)


sofroniewn:
Deprecation warnings added to all methods/ parameters. We've also removed `viewer.events.grid` which should be replaced by `viewer.grid.events.update`.

I'll work on tests next

sofroniewn:
I've fixed and added tests. I think this is ready for review now. After this merges we can think about addressing #1690 and adding support for additional ways to assign layers to grid cells.

napari/components/grid.py
```diff
@@ -0,0 +1,140 @@
+import numpy as np
+
+from ..utils.events import EmitterGroup
+
+
+class Grid:
```
jni: I don't love just "Grid". GridCanvas? CanvasGrid?

napari/components/grid.py
```diff
@@ -0,0 +1,140 @@
...
+        Number of rows and columns in the grid. A value of -1 for either or
+        both of will be used the row and column numbers will trigger an
+        auto calculation of the necessary grid size to appropriately fill
+        all they layers at the appropriate stride.
+    stride : int
+        Number of layers to place in each grid square before moving on to
+        the next square. The default ordering is to place the most visible
+        layer in the top left corner of the grid. A negative stride will
+        cause the order in which the layers are placed in the grid to be
+        reversed.
```
jni, line 28: OMG I totally forgot about this LOL

napari/components/viewer_model.py
```diff
@@ -152,17 +155,53 @@ def theme(self, theme):
 
     @property
     def grid_size(self):
-        """tuple: Size of grid
-        """
-        return self._grid_size
+        """tuple: Size of grid."""
+        warnings.warn(
+            (
+                "The viewer.grid_size parameter is deprecated and will be removed after version 0.4.2."
```
jni: Maybe we start doing deprecations on minor version changes? ie change to 0.5?

napari/components/viewer_model.py
```diff
@@ -397,11 +436,15 @@ def _on_cursor_position_change(self, event):
...
+        for i, layer in enumerate(self.layers[::-1]):
+            i_row, i_column = self.grid.position(i, len(self.layers))
+            self._subplot(layer, (i_row, i_column))
+
     def grid_view(self, n_row=None, n_column=None, stride=1):
         """Arrange the current layers is a 2D grid.
-
         Default behaviour is to make a square 2D grid.
-
         Parameters
```
jni: Revert removing these lines

napari/components/grid.py
```diff
@@ -0,0 +1,140 @@
...
+
+    @stride.setter
+    def stride(self, stride):
+        self._stride = stride
+        self.events.update()
+
+    def actual_size(self, nlayers=1):
+        """Return the actual size of the grid.
+
+        This will return the size paramater, unless one of the row
```
jni:
```suggestion
        This will return the size parameter, unless one of the row
```

napari/components/grid.py
```diff
@@ -0,0 +1,140 @@
...
+    ----------
+    events : EmitterGroup
+        Event emitter group
+    enabled : bool
+        If grid is enabled or not.
+    size : 2-tuple of int
+        Number of rows and columns in the grid. A value of -1 for either or
+        both of will be used the row and column numbers will trigger an
+        auto calculation of the necessary grid size to appropriately fill
+        all they layers at the appropriate stride.
```
jni:
```suggestion
        all the layers at the appropriate stride.
```

jni (approved):
Awesome, thanks @sofroniewn! I've found a couple of minor typos, and my main objection is that Grid is too generic and we should go for something like GridCanvas, but otherwise I think this is ready to go! As always, thanks for all the tests! :blush:

napari/components/viewer_model.py
sofroniewn: Agreed 0.4.2 is a little fast and we should leave at least two patch releases (which correspond to at least 4 weeks on our ~2 week release cadence) between every deprecation and removal, and longer for more significant API changes, including a minor release. We should adopt a more formal deprecation policy, maybe @jni you can open an issue capturing your thoughts there

napari/components/grid.py
GenevieveBuckley: +1 for GridCanvas

GenevieveBuckley (approved):
LGTM

I also agree with the minor suggestions Juan has made here.

napari/components/grid.py
sofroniewn: changed

napari/components/viewer_model.py
sofroniewn: done

napari/components/grid.py
sofroniewn: done

napari/components/grid.py
sofroniewn: done

sofroniewn:
Thanks for the review @jni @GenevieveBuckley i have made all the requested changes. I'll let this sit till tomorrow in case anyone else has any additional comments and then merge!

Issue #1775, opened 2020-10-25
Potential approach to layer slicing


sofroniewn:
## 🚀 Feature
Building on ideas #1353 and following discussion in the team meeting about how best to support orthoviews and move slicing off of the individual layers, I've been investigating some Python Black Magic and Voodoo.

One nice approach would be to reuse our `Layer` classes themselves so that we wouldn't need to make new sliced objects for each of the layers. It would be nice to have an API like

```python
# when dims change
sliced_layer = full_layer.slice(dims)
# possibly instantiate sliced layer, or treat async
# pass slice_layer to vispy (or vispy already has access) and does update
```

where `sliced_layer` is a new instance of the same `Layer` type as `full_layer` but with sliced values for properties that needed slicing, like `data` or `edge_color` for `Points`, and the same values for properties that didn't need slicing like `opacity` for example.

One question with this approach then is how to synchronize the non-sliced properties of the new `sliced_layer` with the `full_layer`. For example if I then change the opacity on the `full_layer` I shouldn't need to reslice! One solution could be to use events to connect everything, but that could result in a lot of boiler plate code and be very cumbersome, so I've done some experimenting with a new approach! 

After some googling around I saw a recommendation to do something similar to this using a small class which placed the value inside a mutable object, and made this little demo

```python
class MutableProperty:
    def __init__(self, v):
        self.value = [v]
        
    def get(self):
        return self.value[0]
    
    def set(self, v):
        self.value[0] = v

class Layer:
    def __init__(self, d, op):
        self.data = d
        if isinstance(op, MutableProperty):
            self._opactiy = op          
        else:
            self._opactiy = MutableProperty(op)

    @property
    def opactiy(self):
        return self._opactiy.get()

    @opactiy.setter
    def opactiy(self, op):
        self._opactiy.set(op)
    
    def print_value(self):
        print("data: ", self.data, " opactiy: ", self.opactiy)
```

Where I have this very small `MutableProperty` class which I'll then use for my non-sliced property `opacity` in my small example `Layer` class. `data` will be sliced and so doesn't need to change.

I can then do the following

```python
full_layer = Layer('ABCD', 0.8)
slice_layer = Layer(full_layer.data[:2], full_layer._opactiy)
new_layer = Layer(full_layer.data, full_layer.opactiy)
full_layer.print_value()
# data:  ABCD  opactiy:  0.8
slice_layer.print_value()
# data:  AB  opactiy:  0.8
new_layer.print_value()
[...]

PR #1827, opened 2020-11-04
Move grid information off layer

### LINKED ISSUE #1690: Move grid information off individual layers

sofroniewn:
# Description
This PR would move the grid information off the layer and close #1690, as layers shouldn't know where they are in the grid. 

I'd like to address #860 before merge as best we can too. I'll be updating this as we go, this is currently still a work in progress.

## Type of change

### DIFF
--- a/napari/_qt/qt_viewer.py
+++ b/napari/_qt/qt_viewer.py
@@ -238,6 +238,8 @@ class QtViewer(QSplitter):
         self.viewer.layers.events.reordered.connect(self._reorder_layers)
         self.viewer.layers.events.added.connect(self._on_add_layer_change)
         self.viewer.layers.events.removed.connect(self._remove_layer)
+        self.viewer.grid.events.update.connect(self._on_grid_change)
+        self.viewer.layers.events.changed.connect(self._on_grid_change)
         # stop any animations whenever the layers change
         self.viewer.events.layers_change.connect(lambda x: self.dims.stop())
 
@@ -571,6 +573,33 @@ class QtViewer(QSplitter):
         top_left = self._map_canvas2world([0, 0])
         bottom_right = self._map_canvas2world(self.canvas.size)
         return np.array([top_left, bottom_right])
+
+    def _subplot(self, layer, position):
+        """Shift a layer to a specified position in a 2D grid.
+
+        Parameters
+        ----------
+        layer : napari.layers.Layer
+            Layer that is to be moved.
+        position : 2-tuple of int
+            New position of layer in grid.
+        """
+        extent = self.viewer._sliced_extent_world
+        scene_shift = extent[1] - extent[0] + 1
+        translate_2d = np.multiply(scene_shift[-2:], position)
+        translate = [0] * layer.ndim
+        translate[-2:] = translate_2d
+        visual = self.layer_to_visual[layer]
+        # go from NumPy to VisPy axis ordering
+        visual.translate_grid = translate[::-1]
+
+    def _on_grid_change(self, event):
+        """Arrange the current layers is a 2D grid."""
+        for i, layer in enumerate(self.viewer.layers[::-1]):
+            i_row, i_column = self.viewer.grid.position(
+                i, len(self.viewer.layers)
+            )
+            self._subplot(layer, (i_row, i_column))
 
     def on_resize(self, event):
         """Called whenever canvas is resized.
--- a/napari/_qt/qt_viewer.py
+++ b/napari/_qt/qt_viewer.py
@@ -549,10 +549,7 @@ class QtViewer(QSplitter):
             Position in world coordinates, matches the total dimensionality
             of the viewer.
         """
-        nd = self.viewer.dims.ndisplay
-        transform = self.view.camera.transform.inverse
-        mapped_position = transform.map(list(position))[:nd]
-        position_world_slice = mapped_position[::-1]
+        position_world_slice = self.camera.map_canvas_to_worldslice(position)
 
         position_world = copy(self.viewer.dims.point)
         for i, d in enumerate(self.viewer.dims.displayed):
--- a/napari/_qt/qt_viewer.py
+++ b/napari/_qt/qt_viewer.py
@@ -573,16 +573,13 @@ class QtViewer(QSplitter):
 
     def _on_grid_change(self, event):
         """Arrange the current layers is a 2D grid."""
-        extent = self.viewer._sliced_extent_world
-        scene_shift = extent[1] - extent[0] + 1
-        size = tuple(scene_shift[-2:])
-
+        # Place layers in grid in reverse order of layer list
         for i, layer in enumerate(self.viewer.layers[::-1]):
             i_row, i_column = self.viewer.grid.position(
                 i, len(self.viewer.layers)
             )
             visual = self.layer_to_visual[layer]
-            visual.subplot((i_row, i_column), size)
+            visual.subplot((i_row, i_column), self.viewer.grid.size)
 
     def on_resize(self, event):
         """Called whenever canvas is resized.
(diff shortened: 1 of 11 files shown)


sofroniewn:
This is still a work in progress. 3D rendering and multiscale grid need to be fixed

This will though now propagate cursor position back to the layer appropriately.

I don't love what's happening here though, the grid does break what previously were some nice abstractions with `world` coordinates, because now we have `world` and `worldgrid`, but it's sort of unavoidable. Maybe we can make this a little nicer before merge

sofroniewn:
Ok this is now ready for review @jni, I can't say I love love it, as I have fears that I'm trading complexity in one part of the code base for another, particularly around the coordinate handling and offsetting. I do think though that removing `layer.translate_grid` is unambiguously a good thing as we try and simplify our layer models though, so I take that at least as a positive! Any ideas on how to make things simpler would be much appreciated.

After this merges I'd like to fix #1836 and then I think we're ready to continue looking at #1775 and improving the slicing. All in all though I'm pretty happy with our new camera, cursor, and grid objects

napari/_qt/qt_viewer.py
```diff
@@ -697,13 +709,24 @@ def on_draw(self, event):
...
+                    grid_position = self.viewer.grid.position(
+                        i, len(self.viewer.layers)
+                    )
+                    offset = np.multiply(grid_position, self.viewer.grid.size)
+                    corner_pixels[:, self.viewer.dims.displayed[-1]] -= offset[
+                        -1
+                    ]
+                    corner_pixels[:, self.viewer.dims.displayed[-2]] -= offset[
+                        -2
+                    ]
```
jni, line 726: Sometimes I really hate `black`...

jni (changes requested):
@sofroniewn I think what's missing here is a Canvas model, and honestly I don't think this is a good intermediate step to get there. If we want another frontend to this, everything will have to move right back from VisPy and Qt to model files...

What do you think? Happy to discuss on Zulip tomorrow morning my time.

sofroniewn:
Yeah, makes sense, I had my doubts about this too. I might pull out the grid.shape rename as I think it’s better than grid.size, but otherwise this can wait

sofroniewn:
> I might pull out the grid.shape rename as I think it’s better than grid.size, but otherwise this can wait

Ok, I've done this now in #1847, I'm turning this to draft until we get a canvas object or have more ideas

napari/layers/base/base.py
```diff
@@ -463,14 +460,16 @@ def affine(self, affine):
...
-    def translate_grid(self, translate_grid):
-        if np.all(self.translate_grid == translate_grid):
-            return
-        self._transforms['world2grid'].translate = np.array(translate_grid)
-        self.events.translate()
+        warnings.warn(
+            (
+                "The layer.translate_grid attribue is deprecated and will be removed after version 0.4.3."
+                " Layer's no longer no about their position in a grid and this property should no longer"
+                " be used. It now returns a list of 0."
```
jni, line 467:
```suggestion
                "The layer.translate_grid attribute is deprecated and will be removed after version 0.4.3."
                " Layers no longer know about their position in a grid and this property should no longer"
                " be used. It now returns a list of 0."
```

napari/components/grid.py
```diff
@@ -15,10 +15,10 @@ class GridCanvas:
         Event emitter group
     enabled : bool
         If grid is enabled or not.
-    size : 2-tuple of int
+    shape : 2-tuple of int
         Number of rows and columns in the grid. A value of -1 for either or
         both of will be used the row and column numbers will trigger an
-        auto calculation of the necessary grid size to appropriately fill
+        auto calculation of the necessary grid shape to appropriately fill
```
jni, line 21: @sofroniewn this docstring needs an explanation of `.size`, which is still an attribute of the class.

napari/components/viewer_model.py
```diff
@@ -433,14 +433,17 @@ def _update_cursor_size(self, event):
...
-            i_row, i_column = self.grid.position(i, len(self.layers))
-            self._subplot(layer, (i_row, i_column))
+            # Offset the position of the cursor if in grid mode
+            if self.grid.enabled:
+                grid_position = self.grid.position(i, len(self.layers))
+                offset = np.multiply(grid_position, self.grid.size)
+                position = list(self.cursor.position)
+                position[self.dims.displayed[-1]] -= offset[-1]
+                position[self.dims.displayed[-2]] -= offset[-2]
+                layer.position = position
```
jni, line 444: I kind of think that this logic should use a transform (for which we have models) rather than this custom translation. Would that be easy enough to add?

napari/_vispy/vispy_base_layer.py
```diff
@@ -66,14 +71,66 @@ def __init__(self, layer, node):
...
+        Transform used if viewer is in grid mode
+        """
+        # whenever a new parent is set, the transform is reset
+        # to a NullTransform so we reset it here
+        if isinstance(self.node.transform, NullTransform):
+            self.node.transform = ChainTransform(
+                [STTransform(), MatrixTransform()]
+            )
+
+        return self.node.transform.transforms[0]
```
jni, line 99: @sofroniewn to me it's still pretty concerning that the grid transform is not in the layer but it *is* in the VisPy layer...

napari/_tests/test_with_screenshot.py
```diff
@@ -313,9 +314,9 @@ def test_grid_mode(make_test_viewer):
     # return to stack view
     viewer.grid.enabled = False
     assert not viewer.grid.enabled
-    assert viewer.grid.actual_size(6) == (1, 1)
+    assert viewer.grid.actual_shape(6) == (1, 1)
     assert viewer.grid.stride == 1
-    translations = [layer.translate_grid for layer in viewer.layers]
+    translations = [vis.translate_grid[:2][::-1] for vis in visuals]
```
jni, line 319: What is `actual_shape` compared to `shape`? This API might benefit from a rethink...

jni: and `vis.translate_grid[:2][::-1]` is super mysterious here...

jni (approved):
@sofroniewn I'll approve since you mentioned this is blocking, but I still have some real concerns about where this is heading... It's weird that things are on a VisPy layer but not a layer itself...

Other than that, I've left minor comments that should be easy to fix!

sofroniewn:
> @sofroniewn I'll approve since you mentioned this is blocking, but I still have some real concerns about where this is heading... It's weird that things are on a VisPy layer but not a layer itself...

Opps @jni - I requested you to review the wrong PR, I wanted you to look at #1847 which is much more minimal, just changes `size` to `shape`. I share reservations about this one too and want to spend more time thinking about it too see https://github.com/napari/napari/pull/1827#issuecomment-722875601, though I do think the overall goal still makes sense

napari/_qt/qt_viewer.py
Czaki: negate `if` clause and call `continue`. This will reduce indentation

napari/_qt/qt_viewer.py
```diff
@@ -697,13 +709,24 @@ def on_draw(self, event):
         This is triggered from vispy whenever new data is sent to the canvas or
         the camera is moved and is connected in the `QtViewer`.
         """
-        for layer in self.viewer.layers:
+        for i, layer in enumerate(self.viewer.layers[::-1]):
```
Czaki, line 712:
```suggestion
        for i, layer in enumerate(reversed(self.viewer.layers)):
```

sofroniewn:
This is still on hold, we'll have to see how some other refactoring goes, but big picture I do still like this

sofroniewn:
We're going to hold on this until we have a better multicanvas model, so I will close this now but leave the underlying issue open

PR #1831, opened 2020-11-04
add example: visualize 2d+timelapse as 3D space-time images a.k.a. known as "kymographs"

VolkerH:
# Description

This PR adds an educational example, showcasing the possibility to visualize time series as volumes. 
A large part of the PR code is helper code that downloads the example datasets from IDR.

The examples downloaded and displayed are selected phenotypes from the Mitocheck screen, 
which investigated how different genes affect cell division (using siRNA knockdown).

In addition to "just" displaying volumes, the code also demonstrates how to enable grid mode programmatically
(this requires the latest version from master), how to enable display of the coordinate axes and how to set a 
specific camera view.

Running the example should produce this output
image

Note: this evolved from a much more complex example that (in retrospect) lacked focused and was culled down to only retain the 3d kymograph example (just in case you're looking at the commit history).

## Type of change

- Added documentation

# References

Data from the Mitocheck screen is used and downloaded from IDR.
Details are printed to the console.

# How has this been tested?
Run locally.

## Final checklist:

For reviewers: I would be surprised if you didn't find a typo in the text part. Feel free to push corrections.


sofroniewn (approved):
Thanks for the nice example @VolkerH - I think in the long run we'll want this living as a nice runnable notebook with all our tutorials, possibly in that tutorials repo. I think that will be nice because it will make the fetching of the images very explicit, but for now I think it's fine to merge and have here.

Curious if @jni or @GenevieveBuckley have anything further they want to add?

VolkerH:
> Thanks for the nice example @VolkerH - I think in the long run we'll want this living as a nice runnable notebook with all our tutorials, possibly in that tutorials repo.

I had started out doing that, based on the more complex variant initially. 
This may be starting to get off-topic, but since you mentioned it:
Was talking offline with @jni, the current tutorials repo has the drawback of being static markdown. I.e. changes to the code need to be manually incorporated back into the markdown. Notebooks would be better possibly even with the vnc trick to run them in Binder. I also think tools that provide functionality similar to [fastpages](https://github.com/fastai/fastpages) (doesn't have to be that one) would help.

GenevieveBuckley:
> > Thanks for the nice example @VolkerH - I think in the long run we'll want this living as a nice runnable notebook with all our tutorials, possibly in that tutorials repo.
> 
> I had started out doing that, based on the more complex variant initially.
> This may be starting to get off-topic, but since you mentioned it:
> Was talking offline with @jni, the current tutorials repo has the drawback of being static markdown. I.e. changes to the code need to be manually incorporated back into the markdown. Notebooks would be better possibly even with the vnc trick to run them in Binder. I also think tools that provide functionality similar to [fastpages](https://github.com/fastai/fastpages) (doesn't have to be that one) would help.

We're currently trying to move the tutorials into notebooks, see https://github.com/napari/tutorials/pull/95 and linked issues. It's blocked because I can't get the notebooks to execute properly on CI (they execute locally just fine), so we need to work out what Qt stuff needs to be installed on the CI virtual machine first. If you're able to help with any of that, we would likely be able to make the switch to supporting notebooks in the tutorials almost immediately.

examples/3d_kymograph.py
```diff
@@ -0,0 +1,148 @@
...
+
+In the juxtaposed kymographs the reduced branching for the mitotitic 
+phenotypes caused by INCENP, AURKB and KIF11 knockdown compared to 
+TMPRSS11A knockdown is immediately obvious.
+
+Data Source
+===========
+The samples to demonstrate this is downloaded from IDR:
+https://idr.openmicroscopy.org/webclient/?show=screen-1302
+
```
GenevieveBuckley, line 95: It would be good to add how large the dataset is (in GB or similar), and roughly how long you should expect it to take to download (just an estimate).

examples/3d_kymograph.py
```diff
@@ -0,0 +1,148 @@
...
+)
+
+print("------------------------------------")
+print("Downloading sample datasets from IDR")
+print("this may take some time")
+print("------------------------------------")
+for s in samples:
+    print(f"Downloading sample {s['IDRid']}.")
+    print(f"Description: {s['description']}")
+    s["vol"] = np.squeeze(IDR_fetch_image(s["IDRid"]))
```
GenevieveBuckley, line 135: Does this script re-download the data every time it is run?
It would be more convenient if we check whether the data has already been downloaded before, and if not then download it.

GenevieveBuckley:
This is a very cool example, @VolkerH

examples/3d_kymograph.py
jni: @VolkerH and I had a conversation about this offline. My opinion was that it was a bit obnoxious to download something to a not-well-defined place (e.g. current directory). The long term solution is to depend on pooch or similar, but I figure for our examples and for now (pre-tutorials/gallery), re-downloading is ok.

examples/3d_kymograph.py
jni: Agreed. Takes about 2min iirc.

examples/3d_kymograph.py
VolkerH: yes, it does.
This current PR is trimmed down version of [a larger example](https://github.com/VolkerH/napari/blob/regionprops/examples/display_regionprops.py) that includes caching of the downloaded data in a Zarr store on disk.

However, I felt that all of the caching and segmentation code in the larger example obscures the view onto the relevant parts. Even in the current example, a large fraction of the code just deals with downloading the sample data and not with showcasing napari's functionality. I put the larger example on ice for now, but it may be something to revisit for a tutorial. I just don't think the main napari repo would be the right place.

examples/3d_kymograph.py
VolkerH: Good suggestion. I will provide a warning about the download size.
Do you think I should ask for keypress? I.e. "Do you want to go ahead with the download? [Enter to confirm]"

I can't really estimate the download speed as that will depend on the user's internet connection and whatever bandwidth limiting IDR may have in place.
I think the tqdm progress bars should give a reasonable estimate though.

GenevieveBuckley: A keypress confirmation is a good idea, I like that.

> I think the tqdm progress bars should give a reasonable estimate though.

Ok, no need to add a comment about the expected time. Users can always stop the script if the download speed looks super slow for some reason.

examples/3d_kymograph.py
GenevieveBuckley: Those are good points. I guess we should just make it super clear to the user about how it works & our long term plan can be some kind of pooch solution.

VolkerH:
Thanks for your feedback everyone.
I added some information about the approximate download size and a user prompt to confirm the download.

GenevieveBuckley (approved):

PR #1834, opened 2020-11-04
async-20: Better QtTestImage

pwinston:
# Description
The `TEST_IMAGES` dict can now specific that a test image has a settable shape or a fixed shape.

A fixed shape looks like this in the UI:

napari-test-1

A settable shape looks like this:

napari-test-0

## Type of change
- [x] New feature in experimental code

# Files

async-20-files

# Explanation
It might seem unconventional to have this sort of UI inside napari, since you can easily open images. However, I think it will be useful for rendering work. Soon we'll have URL's of remote images, we'll have 3d images, different time series types, etc. I think it's good for someone working on rendering to have easy access to a selection of test images to quickly put rendering through its paces. And so repro cases are easy to report if it can be reproduced on a specific test image.

Repro cases with test images will be much easier to deal with than requiring loading someone's specific and potentially hard to access data.

# How has this been tested?
- [x] Manual testing

## Final checklist:

### DIFF
--- a/napari/_qt/experimental/render/qt_test_image.py
+++ b/napari/_qt/experimental/render/qt_test_image.py
@@ -12,7 +12,6 @@ from qtpy.QtWidgets import (
     QPushButton,
     QVBoxLayout,
 )
-from skimage import data as skimage_data
 
 from ....utils import config
 from .qt_labeled_spin_box import QtLabeledSpinBox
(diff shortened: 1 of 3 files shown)


pwinston:
Updated so if `skimage` is not installed, the `skimage.data` options just don't show up in the dropdown list of test images.

napari/_qt/experimental/render/qt_test_image.py
```diff
@@ -1,35 +1,112 @@
...
+        "factory": lambda image_shape: create_test_image(
+            "0", (16, 16), image_shape
+        ),
+    }
 }
+TEST_IMAGE_DEFAULT = "Digits"
+
+# Add skimage.data images if installed. Napari does not depend on
+# skimage but many developers will have it.
+try:
```
sofroniewn, line 44: Thanks!

sofroniewn (approved):
LGTM! Tests are passing, will merge

PR #1835, opened 2020-11-04
Fix #1811: Move visual creation to end of __init__

pwinston:
# Description
In #1811 this change of moving Vispy visual creating to the of `QtViewer.__init__` greatly reduced crashes when we were creating Nodes() every frame. Crashes were with PyQt5 only, PySide2 did not crash.

However, it does not fix node create related crashes with Octree work.

So it's not a full fix, but it seems to improve stability a lot in #1811. And it seems better anyway to factor out these methods from the very long `QtViewer.__init__`.

## Type of change
- [x] Refactoring to reduce crashes

# References

* See #1811

# How has this been tested?
- [X] Manual testing, test suite, ran some examples.

## Final checklist:


sofroniewn:
I know this is still marked as draft, but I just tried it out and ran into a segfault. If I launch with just `napari` then go to the menubar and display the scale bar I get the following segfault. I guess one could say it's a bit weird to show a scale bar on the welcome visual, but we shouldn't seg fault like this. Note quite sure what's going wrong

```python
(base) czirwc1macos2701:napari nsofroniew$ napari
WARNING: Error drawing visual <vispy.visuals.line.line._GLLineVisual object at 0x7f8b5b3267b8>
15:32:01 WARNING Error drawing visual <vispy.visuals.line.line._GLLineVisual object at 0x7f8b5b3267b8>
/Users/nsofroniew/opt/anaconda3/bin/pythonw: line 3: 93676 Segmentation fault: 11  /Users/nsofroniew/opt/anaconda3/python.app/Contents/MacOS/python "$@"
```

pwinston:
Yeah I'm not sure. That's bad. This PR is definitely not something we have to pursue. It seems to make the "automatic node creator" case way more stable. But that's not a user visible problem, that's a synthetic torture case. And obviously if it introduces a crash that's no good. We could hide the scale bar if no layer exists, but it seems dubious to go down that road at all.

I mentioned to @sofroniewn the best thing might be drop the automatic node creator case as too fringe and artificial. And instead create a PR that restores the way `VispyImageVisual` used to swap between `ImageNode` and `VolumeNode`. Because that seems like legit and realistic usage that should work. And it's not octree related so it's in non-experimental code. So that might be the best test case to try to fix.

Basically `VispyImageLayer` swapping between `ImageNode` and `VolumeNode` seems legit, but it crashes, at least on PyQt5. If someone can find a fix for that, that seems valuable.

sofroniewn:
Ok, I've looked into this again. This does not make the behavior on master worse, I can get a segfault by displaying the just the scalebar and the welcome visual when launching with napari on master. If I remove the text on the welcome visual then there's no more segfault. I will merge this now as is, but it's clear that we still have unresolved issues

sofroniewn (approved):
I've fixed merge conflict. Will merge after tests pass

PR #1837, opened 2020-11-04
async-21: TiledImageVisual with Texture Atlas

pwinston:
# Description
* This PR adds a new texture-atlas-based tiled visual for octree rendering.
    * New classes: `VispyTiledImageLayer`, `TiledImageVisual` and `TextureAtlas2D`.
* I renamed the older one to be `VispyCompoundImageLayer`, which creates one `ImageVisual` per tile.
    * The compound one will probably not be used, but useful for comparison for now.
    * Also good to create the crash in #1811. The crash is using PyQt5 only, not PySide2.
* New `QtRender` GUI  features:
    * Dropdown to select layer type for test image.
    * New green/yellow/red framerate meter.

The new fast tiled rendering:

napari-tiled-zoom

With an 8k x 8k image:

napari-tiled-zoom-2

We show zooming here because panning was fine with the old approach. Panning is much less challenging than zooming. You tend to only get a few new tiles with each pan. But with zooming, there is a pop, and suddenly you need to draw a bunch of new tiles all at once. But even zooming is smooth with the new approach.

# Design

TiledImageVisual

The new `TiledImageVisual` does not know about the whole image. It does not figure out which tiles need to be drawn. Instead, that's done in the View code closer to the Model. The View sends just the tiles to be drawn to the renderer, to the `TiledImageVisual`.

An advantage is the quadtree/octree code can send `TileImageVisual` tiles from different levels. All tiles are the same size in pixels. But those same-sized tiles can be drawn as a mix of bigger or smaller tiles like this:

image

You will get this with async loading because you want to display data as soon as it's available. When zooming in or out, you do not want to wait until the entire new level is available. Some stragglers could hold up the whole level for many seconds. It's better to show what you have as soon as it's ready.

The nice thing is multi-level rendering works without `TiledImageVisual` knowing anything about quadtrees or octrees. Also, a big advantage of this approach is the async/quadtree/octree machinery can be used for non-rendering purposes. Some day we might push more decision making into the rendering layer. Especially if we want to push more onto the GPU. But this is how this PR works. And hopefully, we can stick with it for a while.

# Details

## Atlas-based Rendering?

The older `VispyCompoundImageLayer` creates one `ImageVisual` per tile. This led to big hiccups in framerate since creating `ImageVisuals` is slow. Each new `ImageVisual` requires rebuilding the shader, for one thing. Also this approach causes crashes when using  PyQt5, see #1811. We suspect the crashes were due to scene graph changes, and `TiledImageVisual` does not change the scene graph to add/remove tiles. David Hoese (vispy) recommended the atlas approach. He did something atlas-based for [SIFT](https://github.com/ssec/sift) that we looked at, although it's quite different.

## GUI: Select Image Layer Type

New dropdown to create one of three types of image layers:
1. `Image` layer with the original visuals.
2. `OctreeImage` layer with `VispyCompoundImageLayer` visuals.
3. `OctreeImage` layer with new `VispyTiledImageLayer` visuals.

napari-ui

## GUI: New Framerate Meter

Added an experimental framerate meter at the bottom of the `QtRender` widget. The motivation was the new tiled rendering is way less glitchy and faster, but it's mostly a question of transient slow frames. So this shows your current framerate but also recent slow frames. Which are important to know about.

napari-fps-meter-small

# Type of change
- [x] New features in experimental code
- [x] Necessary changes to in non-experimental code (I will add section below soon.)

# Non-Experimental Changes

Looks like a lot of changes but only `_vispy/vendored/image.py` and `_vispy/vispy_image_layer.py` are meaty.

* `qt_viewer.py` - Add `config` flag to not show welcome. Maybe use command-line flag not config? Env var?
* `qt_layerlist.py` - Comment only
* `napari/_vispy/vendored/image.py` - Small change to `_build_color_transform` so our `VispyTiledImage` can use it without making a copy of it.
* `napari/_vispy/vispy_image_layer.py` - New `ImageLayerNode` class so our `VispyTiledImageLayer` can can pass up its `TiledimageVisual`. So `VispyImageLayer` can have 3 nodes: `Image`, `Volume` or custom from a derived class like `VispyTiledImageLayer`.
* `napari/components/add_layers_mixin.py` - Simplify/update config.
* `napari/layers/image/_image_slice.py` - Comment only.
* `napari/layers/image/image.py` - Comments only.
* `napari/viewer.py` - Comments only

# New Files

GUI:
* `napari/_qt/experimental/render/qt_frame_rate.py`

Visuals:
* `napari/_vispy/experimental/texture_atlas.py`
* `napari/_vispy/experimental/tile_grid.py`
* `napari/_vispy/experimental/tiled_image_visual.py`
* `napari/_vispy/experimental/tile_set.py`
* `napari/_vispy/experimental/vispy_compound_image_layer.py` (but this is just the old one)

# All Files

async-21-files-1

# References

## Texture Atlas
* [Texture Array](https://www.khronos.org/opengl/wiki/Array_Texture) - Khronos Wiki
* [Improve Batching Using Texture Atlases](http://download.nvidia.com/developer/NVTextureSuite/Atlas_Tools/Texture_Atlas_Whitepaper.pdf) - NVIDIA Whitepaper
* [Texture atlases, wrapping and mip mapping](https://0fps.net/2013/07/09/texture-atlases-wrapping-and-mip-mapping/) - 0FPS Blog Post
* [Using Texture Atlas as Texture Array in OpenGL](https://stackoverflow.com/questions/59260533/using-texture-atlas-as-texture-array-in-opengl) - Stack Overflow
* [Array Texture (2D)](https://sites.google.com/site/john87connor/texture-object/tutorial-09-6-array-texture) - Tutorial
* [Satellite Information Familiarization Tool](http://sift.ssec.wisc.edu/) - SIFT
    * Contains a texture atlas and tiled visual written by David Hoese, a lead Vispy maintainer.
    * See the source for [TextureAtlas2d](https://github.com/ssec/sift/blob/master/uwsift/view/texture_atlas.py) and [TileGeolocatedImageVisual](https://github.com/ssec/sift/blob/master/uwsift/view/visuals.py#L246-L254).

## Numpy
* [Numpy's meshgrid Explained](https://stackoverflow.com/questions/36013063/what-is-the-purpose-of-meshgrid-in-python-numpy)
* [Numpy Array Reshaping](https://towardsdatascience.com/reshaping-numpy-arrays-in-python-a-step-by-step-pictorial-tutorial-aed5f471cf0b)
* [Visualizing Strides](https://stackoverflow.com/questions/32034237/how-does-numpys-transpose-method-permute-the-axes-of-an-array)

## Rendering

* [Optimizing Subdivisions in Spatial Data Structures](https://cesium.com/blog/2017/03/30/spatial-subdivision/)
* [Virtual World Sandbox Terrain Enging](https://sandboxdocs.readthedocs.io/en/latest/tutorials/working-with-terrain/)
* [OpenGL Tile Map Rendering](https://gamedev.stackexchange.com/questions/173509/opengl-tile-map-rendering-with-1-quad-for-all-tiles-and-1-texture-atlas)


pwinston:
This is still WIP but the new `VispyTiledImageLayer` with `TiledImageVisual` and `TextureAtlas2D` all work at least the very basics. And it's FAST. But I was having trouble characterizing how slow the older "compound" visual was because steady-state they are both fast, but it's the creating new tiles which is chunky and glitchy in the old system.

So I created a new frame rate meter that I think does a good job of conveying not just the current frame rate but the recent history of slower frames. But in a very compact display not a strip chart or scrolling display. Needs some labels, but already proving useful. It really does give you a flavor of the recent history of frames, not just the most recent.

This is showing the older compound visual:

napari-frame-rate-meter

Also it's logarithmic. I need to calibrate so the numbers are more meaningful, but yellow starts at around 100ms (10Hz) and red starts at around 400ms (2.5Hz). So the red lights are single frames taking over 500ms.

The new tiled visual has FAR fewer slow frames. Fully video comparison coming as soon as color rendering if fixed and some other things. When we can focus on merging this PR.

sofroniewn:
Wow love the frame rate meter, and it's history functionality!

> Also it's logarithmic. I need to calibrate so the numbers are more meaningful, but yellow starts at around 100ms (10Hz) and red starts at around 400ms (2.5Hz). So the red lights are single frames taking over 500ms.

I wouldn't worry about that too much for this PR, I'm happy to merge as is and make an issue for followup PR where we might use a nice colormap of some form

pwinston:
I hand calibrated the framerate meter to be roughly this, from a code comment. 

     # GREEN: up to 31ms (32Hz)
     # YELLOW: up to 312ms (3.1Hz)
     # RED: up to 6000ms (0.16Hz)

So that seems good for now.  Basically green is good, yellow is getting bad fast, and red is pretty awful.

I just experimented and found a log base of 1.35 got kept 30Hz and below in green, which is my main goal. But I shifted less than 16.7ms (60Hz) to be perfect otherwise we wasted too much room on zero to 16.7 which is all basically perfect. Displaying an actual filtered number at some point would be nice, but not worrying about that now.

I just want to fix color (RGB) not showing up, then figure out and writeup what non-experimental stuff changed and decide if it can be improved or cleaned up. Also will fix tests.

pwinston:
Cool, with a small/safe change to our vendored `ImageVisual`, our `TiledImageVisual` can now use `ImageVisual`'s `_build_color_transform` function. So we don't need a copy of it. We still has a mostly copied version of `ImageVisual`'s method `_prepare_draw`, but we might be able to whittle that down.

This was to fix the RGB problem. Which is now fixed.

pwinston:
I just realized this does not support non-power-of-2 images right now. The older `VispyCompoundImageLayer` did, it can create `ImageVisuals` of any size, for edge and corner tiles. But right now `VispyTiledImage` requires all tiles be the same size. We could cheat and just pad them with blank pixels. But the right solution is we accept tiles which are smaller, we pad them, but then we compute the smaller texture coordinates, so draw the exact right size tile with just the exact texels we want.

This might be worth doing in a small `async-22`. And then `async-23` is async/`ChunkLoader`/zarr, hopefully. I haven't needed non-power-of-two because I'm mostly using power-of-two test images. But most real-world data will have edge/corner tiles.

sofroniewn:
I'm trying this out now. If I turn on `NAPARI_OCTREE=1` then astronaut works great for me, but I get a segfault if I try and create a digits example fresh on launch
```python
WARNING: Error drawing visual <vispy.visuals.line.line._GLLineVisual object at 0x7f80cd9f9940>
20:58:33 WARNING Error drawing visual <vispy.visuals.line.line._GLLineVisual object at 0x7f80cd9f9940>
/Users/nsofroniew/opt/anaconda3/bin/pythonw: line 3: 26075 Segmentation fault: 11  /Users/nsofroniew/opt/anaconda3/python.app/Contents/MacOS/python "$@"
```
Not sure if this is related to #1835 or not (I'll try and get to that review soon). Not sure it needs to be fixed now or in the future, but wanted to report.

I'm going to continue with review

napari/_qt/qt_viewer.py
```diff
@@ -198,7 +198,7 @@ def __init__(self, viewer, welcome=True):
         )
         self.canvas.events.resize.connect(self.scale_bar._on_position_change)
 
-        self._show_welcome = welcome
+        self._show_welcome = welcome and config.allow_welcome_visual
```
sofroniewn, line 201: I guess this is fine, in the long run we'll probably want to figure out other ways of doing this (like #1183, but that PR is stalled) but I'm guessing you wanted a quick way to disable this because of all the seg faults etc. we were running into

napari/_vispy/experimental/texture_atlas.py
```diff
@@ -0,0 +1,203 @@
...
+_QUAD = np.array(
+    [[0, 0], [1, 0], [1, 1], [0, 0], [1, 1], [0, 1]], dtype=np.float32,
+)
+
+# TexInfo is returned from TextureAtlas2D.add_tile() so the caller has the
+# texture coordinates to render each tile in the atlas.
+TexInfo = namedtuple('TexInfo', "tile_index tex_coord")
+
+
+class TextureAtlas2D(Texture2D):
```
sofroniewn, line 21: This is very nice, definitely a target for contributing back to vispy at some point!

pwinston:
I marked this ready for review. I think after this `async-22` can be a small PR to support non-power-of-two images (edges and corners in the texture atlas). The real next goal is async/ChunkLoader/zarr, not sure if that can be `async-23` or not, have to see what pops up.

napari/_vispy/experimental/tiled_image_visual.py
```diff
@@ -0,0 +1,354 @@
...
+    scaled_shape = chunk_data.data.shape[:2] * scale
+
+    # Modify XY's into place
+    quad[:, :2] *= scaled_shape
+    quad[:, :2] += chunk_data.pos
+
+    return quad
+
+
+class TiledImageVisual(ImageVisual):
```
sofroniewn, line 78: Looking at this, I'm wondering how we could get this class in a form such that it could go upstream to vispy one day.

The biggest unknown for me right now is how to think about the `ChunkData` https://github.com/napari/napari/blob/master/napari/layers/image/experimental/octree_util.py#L42 class, which although is a very simple isn't part of any existing vispy API.

If using the ChunkData class is very convenient within vispy one approach could be to add a `VispyChunkData` in our `vispy` folder that could go upsteam too, but I think we'd want that to be a "private" class (i.e. you wouldn't need to use it to actually set a chunk, that API would just take the needed keyword arguments and basic numpy types)

We could then still have our own `ChunkData` that got converted to the vispy one our VispyTiledImageLayer class. While that might seem like a lot of overhead, I do think that keeping some separation/ modularity between napari and vispy is important, as well as pushing as much upstream to vispy as we can.

Not something we need to decide now, but just wanted to mention this and see if you had any thoughts on it.

napari/_vispy/vendored/image.py
```diff
@@ -113,8 +113,8 @@
 _c2l = 'float cmap(vec4 color) { return (color.r + color.g + color.b) / 3.; }'
 
 
-def _build_color_transform(data, clim, gamma, cmap):
-    if data.ndim == 2 or data.shape[2] == 1:
+def _build_color_transform(grayscale: bool, clim, gamma, cmap):
```
sofroniewn, line 116: I'm ok with this change, it's in a private method that only gets used once, inside the file. I actually find it much more readable with `grayscale` too. We've got another PR #1853 that touches this too, but in a different way, so I think that's fine too. It's also clear how to adjust #1853 to solve the merge conflict.

We will need to push this up to vispy one day most likely, but I think it should be fine as the method is private

napari/_vispy/vispy_image_layer.py
```diff
@@ -16,11 +17,33 @@
 ]
 
 
-class VispyImageLayer(VispyBaseLayer):
-    def __init__(self, layer):
+class ImageLayerNode:
```
sofroniewn, line 20: I'm fine with this too, very straight forward

sofroniewn (approved):
Great work @pwinston, I'm approving. This is coming along very nicely. The astronaut worked great for me.

I've left some fairly minor comments/ questions - nothing major. I'll let this sit till tomorrow midday, and you can see if there are any more updates you want to make before merge.

napari/_qt/qt_viewer.py
pwinston: In this case, it wasn't the crashes. It was because I wanted to set breakpoints in image visual code, but I didn't want them to hit until my octree images were created. Incredibly minor, yet a developer should have the ability to toggle something like that on or off for a number of reasons.

I wonder if you can push through the worse possible version of #1183 and then learn from that and improve it. Some type of config file is pretty key. Command line args and env vars are not bad, but a config files is arbitrarily big, structured, can be backed up and have multiple versions, etc.

napari/_vispy/experimental/texture_atlas.py
pwinston: I think both `TextureAtlas2D` and `TiledImageVisual` are candidates. But our "tiled image visual" is much less than what David probably had in mind. But it seems useful as part of a bigger thing. A simple/solid building block rather than something too complicated .

napari/_vispy/experimental/tiled_image_visual.py
pwinston: Yeah the weird thing is our model code cannot depend on vispy. Our VispyVisuals can depend on model code. But then vispy itself cannot depend on any of our code at all. It's kind of this weird loop.

The `level_index` should not be in the `ChunkData` that vispy knows about. So then it's just data/pos/scale. So you can just pass those in as 3 variables. And have vispy have a private namedtuple or dataclass for them.

The `level_index` is more complicated. Vispy should not have the concept of a level, that's octree specific. But vispy would need some type of generic identifier so callers can signal if a chunk is really the same one or not. Not really sure how that would look. It could be a hash of the data and the pos/scale, but hashing data is slower than some other things.

TBD basically, not sure. Basically `TiledImageVisual` just needs a way to say "am I already drawing this exact thing or is it new?"

napari/_vispy/vendored/image.py
pwinston: Yeah the next step on the vispy-side would be take all those private methods and put them in a file like `image_util.py` so derived classes can easily call into them without having to call into private methods in `image.py`.

They are only private because no one else needed them. But if derived image-related classes want to use them as well, there's no downside in splitting them out, that I see.

How have to copy and modify `_prepare_draw()` is also not great. But less obvious how to fix that up. But ideally `ImageVisual` can be extended with derived classes in a nice way without duplication or reaching into private methods.

But that's why you have to refactor base classes once you've made some derived classes and seen the pain points.

napari/_vispy/vispy_image_layer.py
pwinston: Yes it feels like some bigger change is afoot there. But this was all I needed. To start I just put `TiledImageVisual` under the empty `VispyBaseLayer.node`. That is, we had a do-nothing `ImageVisual` and then under that our `TiledImageVisual`. But that seemed likely to cause a problem someday. Having `self.node` be the `TiledImageVisual` itself seemed clearly better/cleaner.

pwinston:
I added a design diagram to the PR description above.

If we end up keeping this design we'll want to update the rendering doc with that image.

napari/_vispy/experimental/texture_atlas.py
pwinston: I added a design diagram above that I think explains how our `TiledImageVisual` works compared with maybe what David expected.

napari/_vispy/experimental/tiled_image_visual.py
sofroniewn: Yeah, sounds like we're not too far away, but some details would need to be worked out. We can revisit once we have everything working.

sofroniewn:
Thanks for the replies to my comments. Everything sounds reasonable, so I'm going to merge this now.

